Issue #8, opened 2018-01-30
Build a `oc` wrapper library of sorts


containscafeine:
We need to build a wrapper around the `oc` commands that we will frequently use instead of always execing to `oc`.
Ideal case would be to use https://github.com/openshift/client-go, but AFAIK it does have support for all the `oc` commands, so till happens, we will need a wrapper which we can easily replace later on.

This could very well be a different package or even a project outside which we simply import.

kadel:
can we close this, or do you want to keep it open?

PR #13, opened 2018-01-31
[Work In Progress] First take on OcClient 

kadel:
#8

create pkg/occlient 
this is wrapper around oc command
use NewOcClient to create new OcClient instance

this is more of a prototype , just wanted to test if this approach will work


kadel:
@containscafeine cleaned

Issue #14, opened 2018-01-31
implement `component create`


kadel:
`component create <component_type> [component_name] [--app app_name]  [--binary binary_path] [--git git_url] [--dir directory]`

`component_type` is a type of the application such as php, nodejs, python etc…., or it can be mongodb, mariadb etc…
`component_name` - is an optional parameter, its value defaults to the component_type value, 
Component is created in context of current application by default
`--app` is set than component is created in context of application set by this parameter

PR #18, opened 2018-02-01
Component create and delete

kadel:
#14

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -1,0 +1,66 @@
+package component
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/redhat-developer/ocdev/pkg/application"
+	"github.com/redhat-developer/ocdev/pkg/occlient"
+)
+
+const applicationLabel = "app.ocdev.developers.redhat.com"
+const componentLabel = "component.ocdev.developers.redhat.com"
+
+func CreateFromGit(name string, ctype string, url string) (string, error) {
+	currentAppliction, err := application.GetCurrent()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to get current application")
+	}
+
+	output, err := occlient.NewAppS2I(name, ctype, url, map[string]string{applicationLabel: currentAppliction, componentLabel: name})
+	if err != nil {
+		return "", err
+	}
+
+	return output, nil
+}
+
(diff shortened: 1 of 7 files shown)


kadel:
creating new component from remote git repository should work now.

`ocdev component create nodejs --git https://github.com/openshift/nodejs-ex.git`

pkg/application/application.go
```diff
@@ -52,3 +55,28 @@ func Create(name string) error {
 
 	return nil
 }
+
+// GetCurrentApplication gets currently active application
+// if no application is active, defaultApplication is created
+func GetCurrentApplication() (string, error) {
```
containscafeine: I think should be a part of the config package in config.go, like we have other methods there like, `ApplicationExists()`, `AddApplication`, we can also have `CurrentApplication()`

kadel: I think that if it is operation on application that it should be in the application package. 
Doing it like this, if we move to another way to store application info we won't have to touch anything else than this one function

kadel:
@containscafeine it should be ready to merge.

following should work
```
ocdev component create nodejs --git https://github.com/openshift/nodejs-ex.git
ocdev component create nodejs --dir ./nodejs-ex
ocdev component create nodejs
```

pkg/component/component.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "unable to get current application")
+	}
+
+	output, err := occlient.NewAppS2I(name, ctype, url, map[string]string{applicationLabel: currentAppliction, componentLabel: name})
+	if err != nil {
+		return "", err
+	}
+
+	SetCurrent(name)
```
containscafeine: you maybe forgot to store this to `err` here

pkg/component/component.go
```diff
@@ -0,0 +1,122 @@
...
+	currentAppliction, err := application.GetCurrent()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to get current application")
+	}
+
+	output, err := occlient.NewAppS2IEmpty(name, ctype, map[string]string{applicationLabel: currentAppliction, componentLabel: name})
+	if err != nil {
+		return "", err
+	}
+	SetCurrent(name)
```
containscafeine: same

PR #20, opened 2018-02-01
app application create

containscafeine:
merge after #22

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -16,20 +16,31 @@ package cmd
 
 import (
 	"fmt"
-
+	"github.com/redhat-developer/ocdev/pkg/application"
 	"github.com/spf13/cobra"
+	"os"
 )
 
-// appCmd represents the app command
-var appCmd = &cobra.Command{
-	Use:   "app",
-	Short: "Create Application",
-	Long:  `-`,
+// applicationCmd represents the app command
+var applicationCmd = &cobra.Command{
+	Use:     "application",
+	Short:   "application",
+	Aliases: []string{"app"},
+}
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "create an application",
+	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("app called")
+		if err := application.Create(args[0]); err != nil {
+			fmt.Println(err)
+			os.Exit(-1)
+		}
 	},
 }
 
 func init() {
-	rootCmd.AddCommand(appCmd)
+	applicationCmd.AddCommand(createCmd)
+	rootCmd.AddCommand(applicationCmd)
 }
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -1,0 +1,54 @@
+package application
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/redhat-developer/ocdev/pkg/config"
+	"github.com/redhat-developer/ocdev/pkg/occlient"
+)
+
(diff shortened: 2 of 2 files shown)


cmd/application-create.go
```diff
@@ -0,0 +1,25 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/redhat-developer/ocdev/pkg/ocdev"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+var createCmd = &cobra.Command{
```
kadel: could we have this in app.go together?

containscafeine: I tried, but cobra needs to have an init() function for every new command, so I didn't figure out a way on how to do that :(

kadel: look at this : https://github.com/kadel/ocdev/blob/componentCreate/cmd/component.go

I've  put command and subcommand into one file.

cmd/application-create.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/redhat-developer/ocdev/pkg/application"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "create an application",
+	Run: func(cmd *cobra.Command, args []string) {
+		for _, app := range args {
```
kadel: this looks strange, why go through all arguments? 

We should be creating just one application. `ocdev application create myApp` if more arguments provided, it should error out

pkg/application/application.go
```diff
@@ -0,0 +1,54 @@
...
+// If no project is set, this errors out.
+func Create(name string) error {
+	// Get current project name
+	project, err := occlient.GetCurrentProjectName()
+	if err != nil {
+		return errors.Wrap(err, "unable to get current project's name")
+	}
+
+	// Set default application name if not set
+	if len(name) == 0 {
```
kadel: what is use case for this?
I think this should error out rather than using some default name.

cmd/application-create.go
containscafeine: what's wrong with providing multiple applications in one command?
`ocdev application create wordpress mariadb` will create both the applications.

kadel: I think it is confusing :-(

pkg/application/application.go
kadel: ah, ok 👍

cmd/application-create.go
containscafeine: Hmm, works now, was failing for me before :man_shrugging: 
works now!

pkg/application/application.go
kadel: ok, that is more for when you are creating a component and no application is active.

ok, then let's keep it here, we will use it later

cmd/app.go
```diff
@@ -16,20 +16,32 @@ package cmd
...
+var applicationCmd = &cobra.Command{
+	Use:     "application",
+	Short:   "application",
+	Aliases: []string{"app"},
+}
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "create an application",
 	Run: func(cmd *cobra.Command, args []string) {
```
kadel: Also this should error (show help) when no argument is given.

You can do it by adding `Args: cobra.ExactArgs(1)`
see: https://github.com/spf13/cobra/#positional-and-custom-arguments

containscafeine: Thanks, TIL :)

cmd/application-create.go
containscafeine: fixed

Issue #26, opened 2018-02-05
Auto generate documentation


containscafeine:
Since `ocdev` is a very CLI centric tool, it'd be great if we could somehow generate entire CLI documentation and push it on every merge. This takes away the task of manually writing docs, all we will need to do is to put in good descriptions and usages in every command we have with cobra, and we can generate the docs using that.

Cobra lets us generate markdown, man pages, etc, so we could leverage that.
https://github.com/spf13/cobra#generating-documentation-for-your-command

containscafeine:
I could not find a flat way to implement this such all the command and nested get printed.
Maybe if we need this, we might have to implement this ourself.
We can use something like -
```golang
package cmd

import (
	"github.com/pkg/errors"
	"github.com/spf13/cobra/doc"
)

func GenerateCLIDocs() error {
	if err := doc.GenYamlTree(rootCmd, "/tmp/"); err != nil {
		return errors.Wrap(err, "error generating man from cobra")
	}

	return nil
}
```
and from the YAML that we get, we can pretty print a flat structure.

containscafeine:
This has been implemented to some extent in #27 
Now we need to add a travis check that if some PR is modifying the CLI structure, then README.md needs to be updated with the newer structure.
We need to get the CLI Structure section from README.md and parse it in a script

PR #27, opened 2018-02-05
generate a flat command structure with docs

containscafeine:


containscafeine:
This is a bigger task than I thought it was, so will look at it again later on.
Will fix #26

Issue #23, opened 2018-02-02
implement `application get`


kadel:
`app get [--short]`

`app get` 
Display message: "You are currently working on the XYZ application"

`app get --short`
display just application name nothing else

(use-case 7)

PR #31, opened 2018-02-05
add short flag for `application get`

containscafeine:
This commit implements a `-q` flag for `application get` which
returns only the application name.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -16,10 +16,10 @@ package cmd
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


containscafeine:
fix #23

kadel:
I don't understand why `q`. Like "quiet", but it is not quiet, it still prints something, but only shorter :-)

But I'm ok with this, we can always figure out this later.

containscafeine:
Same as `oc project --short` == `oc project -q`
But yeah, you're correct.

PR #33, opened 2018-02-05
implement `ocdev application list`

containscafeine:


kadel:
```
▶ ./ocdev application list
NAME
project1
project2
```

header `NAME` probably shouldn't be there

PR #93, opened 2018-02-15
README.md - add installation and pre-requisites

kadel:
prerequisite https://github.com/redhat-developer/ocdev/pull/92 
it refers to packages generated by travis

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,21 +4,108 @@
 ## What is ocdev?
 OpenShift Command line for Developers
 
+## Pre-requisites
+- OpenShift version 3.7.0 and up
+
+To use ocdev you need access to an OpenShift instance and OpenShift CLI installed on your local machine (`oc` should be in your $PATH).
+
+
+### OpenShift instance
+You can use [Minishift](https://docs.openshift.org/latest/minishift/index.html) to get a local instance of OpenShift. However ocdev can be used with any instance of OpenShift.
+
+### OpenShift CLI 
+There are different ways to install OpenShift CLI. 
+Please follow [OpenShift documentation](https://docs.openshift.org/latest/cli_reference/get_started_cli.html#installing-the-cli).
+
+
 ## Installation
-You can [download](https://dl.bintray.com/ocdev/ocdev/latest/) latest binaries (build automatically from current master branch).
+You can download latest master builds from [Bintray](https://dl.bintray.com/ocdev/ocdev/latest/) or 
+builds for released versions from [GitHub releases page](https://github.com/redhat-developer/ocdev/releases).
 
-### How to use ocdev as an oc plugin?
-- make sure that ocdev binary exists in your $PATH
-- copy the [plugin.yaml](./plugin.yaml) file to ~/.kube/plugins/ocdev/
-- use the plugin as `oc plugin dev`
+### MacOS
+1. First you need enable `kadel/ocdev` Homebrew Tap:
+    ```sh
+    brew tap kadel/ocdev
+    ```
+2. 
+    - If you want to install latest master build
+    ```sh
+    brew install kadel/ocdev/ocdev -- HEAD
+    ```
+    - If you want to install latest released version
+    ```sh
+    brew install kadel/ocdev/ocdev
+    ```
 
-### Concepts
+### Linux
+#### Debian/Ubuntu and other distributions using deb
+1. First you need to add gpg [public key](https://bintray.com/user/downloadSubjectPublicKey?username=bintray) used to sign repositories.
+    ```sh
+    curl https://bintray.com/user/downloadSubjectPublicKey?username=bintray | apt-key add -
+    ```
+2. Add ocdev repository to your `/etc/apt/sources.list`
+    - If you want to use latest master builds add  `deb https://dl.bintray.com/ocdev/ocdev-deb-dev strech main` repository.
+      ```sh
+      echo "deb https://dl.bintray.com/ocdev/ocdev-deb-releases strech main" | sudo tee -a /etc/apt/sources.list
+      ```
+    - If you want to use latest released version add  `deb https://dl.bintray.com/ocdev/ocdev-deb-releases strech main` repository.
+      ```sh
+      echo "deb https://dl.bintray.com/ocdev/ocdev-deb-dev strech main" | sudo tee -a /etc/apt/sources.list
+      ```
+3. Now you can install `ocdev` and you would install any other package.
+   ```sh
+   yum install ocdev
+   # or 'dnf install ocdev'
+   ```
+
+
+#### Fedora/Centos/RHEL and other distribution using rpm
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,21 +4,105 @@
 ## What is ocdev?
 OpenShift Command line for Developers
 
+## Pre-requisites
+To use ocdev you need access to OpenShift instance and OpenShift CLI installed on your local machine (`oc` command in your $PATH).
```
containscafeine: s/OpenShift CLI/OpenShift client

`oc` command should be preset in $PATH

README.md
```diff
@@ -4,21 +4,105 @@
...
+
+### OpenShift instance
+The easiest way to get started with OpenShift is [Minishift](https://docs.openshift.org/latest/minishift/index.html).
+
```
slemeur: Should we also point that ocdev will work with other OS instances?
Have we tried it on OSO?

README.md
slemeur: Maybe we should also add information about the compatible version of OC ?

kadel: > s/OpenShift CLI/OpenShift client

I've used CLI because the same term is used in official OpenShift documentation 
https://docs.openshift.org/latest/cli_reference/index.html

kadel: I've added information about OpenShift version

README.md
```diff
@@ -4,21 +4,105 @@
 ## What is ocdev?
 OpenShift Command line for Developers
 
+## Pre-requisites
+To use ocdev you need access to OpenShift instance and OpenShift CLI installed on your local machine (`oc` command in your $PATH).
+
+### OpenShift instance
+The easiest way to get started with OpenShift is [Minishift](https://docs.openshift.org/latest/minishift/index.html).
```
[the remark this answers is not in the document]
kadel: thanks @LalatenduMohanty . Replaced that sentence with your suggestion.

README.md
kadel: Yes, I've been testing it with OpenShift Online and it works without any problems.

README.md
```diff
@@ -4,21 +4,108 @@
 ## What is ocdev?
 OpenShift Command line for Developers
 
+## Pre-requisites
+- OpenShift version 3.7 and up
```
LalatenduMohanty: @kadel you mean v3.7.0 right? I think it is better to be more specific.

README.md
```diff
@@ -4,21 +4,108 @@
...
+### OpenShift CLI 
+There are different ways to install OpenShift CLI. 
+Please follow [OpenShift documentation](https://docs.openshift.org/latest/cli_reference/get_started_cli.html#installing-the-cli).
+
+
 ## Installation
-You can [download](https://dl.bintray.com/ocdev/ocdev/latest/) latest binaries (build automatically from current master branch).
+You can download latest master builds from [Bintray](https://dl.bintray.com/ocdev/ocdev/latest/) or 
+builds for released versions from [GitHub releases page](https://github.com/kadel/ocdev/releases).
+
```
LalatenduMohanty: -1 for redirecting to personal repository for releases. I would suggest you to do an alpha release the main repository if you want to mention github release or we can just remove it from the readme for now. We can add it later when the release will be done.

README.md
```diff
@@ -4,21 +4,108 @@
...
+### MacOS
+1. First you need enable `kadel/ocdev` Homebrew Tap:
+    ```sh
+    brew tap kadel/ocdev
+    ```
```
LalatenduMohanty: This also concerns me. The issue is once you mention a location users will start using it and would expect it working every time. So better to have proper brewtab for per-releases i.e. alpha beta etc or do not have it all together for now.

README.md
```diff
@@ -4,21 +4,108 @@
...
+        name=bintray-ocdev-ocdev-rpm-releases
+        baseurl=https://dl.bintray.com/ocdev/ocdev-rpm-releases
+        gpgcheck=0
+        repo_gpgcheck=0
+        enabled=1
+        ```
+        Or you can download it using following command:
+        ```sh
+        sudo wget https://bintray.com/ocdev/ocdev-rpm-releases/rpm -O /etc/yum.repos.d/bintray-ocdev-ocdev-rpm-releases.repo
+        ```
```
LalatenduMohanty: Wondering why you do not use https://copr.fedorainfracloud.org/ for this. This way it is the standard what way for RPMs those are not included in the distribution. You can build it for Fedora (multiple versions) and EL.

README.md
kadel: that is a typo.
the link should be https://github.com/redhat-developer/ocdev/releases
thank for noticing

README.md
kadel: updated

README.md
kadel: > So better to have proper brewtab for per-releases i.e. alpha beta etc or do not have it all together for now.

don't understand this one :-(

README.md
kadel: Because, to be honest creating proper rpm SPEC for go project with vendor sucks :-( and it is hard to automate (especially with covering vendor updates). It would take me multiple days to do it. This was just couple of hours of work and it is working ;-)

I would rather have at least some package than no package at all.

LalatenduMohanty: Fair enough. We can start with what you have.

README.md
LalatenduMohanty: I mean https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask 

We have a cask for alpha, beta releases.

kadel: I'm still not sure if I understand it :-( 

But Cask is only for MacOS gui applications or sw distributed as binaries no?

If you are suggesting creating different taps for alpha/beta/stable.I'm not sure if we need it. I would rather have one channel for released versions and other for master builds.

I don't expect that we will keep releasing beta or alpha releases after first stable release.
Master builds will act as alpha/beta releases and those have a separate channel.

README.md
```diff
@@ -4,21 +4,108 @@
 ## What is ocdev?
 OpenShift Command line for Developers
 
+## Pre-requisites
+- OpenShift version 3.7.0 and up
+
+To use ocdev you need access to OpenShift instance and OpenShift CLI installed on your local machine (`oc` should be in your $PATH).
```
containscafeine: to "an" OpenShift instance

kadel: fixed

kadel:
Thank you for reviews. 
I'm merging it, as this is a big improvement compared to what we have right now. 
It is not perfect, we will continue to work on it and we will keep improving it.

README.md
LalatenduMohanty: @kadel cask is for any applications on macOS. You can define taps for alpha/beta releases, another for master build and another for nightly. IMO that is more than enough. For Minishift we have only beta and GA channel.

Issue #83, opened 2018-02-13
An application should not map to a project


jorgemoralespou:
Looking at the output and source code, it looks like an application is mapped to a project.

```
$ oc get project
NAME                  DISPLAY NAME   STATUS
devconfcz18                          Active
labs                                 Active
samplelabs                           Active
the-lounge-jmorales                  Active

$ ocdev application list
devconfcz18
labs
samplelabs
the-lounge-jmorales
```

As can be inferred from "Use case 14: As a developer, I want to list all applications in my project" an OpenShift project can have multiple applications.

kadel:
mapping applications to project is a temporary simplification because we didn't wont to get blocked on this at the beginning.

kadel:
closed in #94 
there are still some issues with it, separate issues were created to address those #105 #104

PR #94, opened 2018-02-15
Allow multiple applications in one OpenShift project

### LINKED ISSUE #85: Labels for app and component should be the standard
When creating applications on OpenShift/K8S, there's currently one label that is standard "app" and that the OpenSHift UI uses to group components that belong to the same application. OpenShift/K8S are trying to standardize on other labels, (e.g. component), and this should be used, so that the application is properly visualized, and in the future, managed with other tools.

I have created an application called devconfcz18 and then a component called httpd.

This are the labels set on the deployment:
screen shot 2018-02-13 at 17 08 44

And this is what can be seen in the Overview.
screen shot 2018-02-13 at 17 12 35

As can be seen in the second image, OpenShift identifies the application as "httpd" and not as "devconfcz18"

kadel:
Applications are now grouped using labeles.

`app.kubernetes.io/name` is used for application grouping.
`app.kubernetes.io/component-name` is used for component grouping

It is also possible to add additional labels to applications.
Currently, there is  `app` label for OpenShift console.

fixes #85 #83

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -1,17 +1,3 @@
-// Copyright © 2018 NAME HERE <EMAIL ADDRESS>
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 package cmd
 
 import (
@@ -22,8 +8,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const defaultApplication = "app"
-
 // applicationCmd represents the app command
 var applicationCmd = &cobra.Command{
 	Use:     "application",
@@ -31,18 +15,19 @@ var applicationCmd = &cobra.Command{
 	Aliases: []string{"app"},
 }
 
-var createCmd = &cobra.Command{
+var applicationCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: "create an application",
-	Args:  cobra.MaximumNArgs(1),
-	Run: func(cmd *cobra.Command, args []string) {
-		var name string
-		// Set default application name if not set
-		if len(args) == 0 {
-			name = defaultApplication
-		} else {
-			name = args[0]
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
+			return fmt.Errorf("Please provide name for the new application")
 		}
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		// Args validation makes sure that there is exactly one argument
+		name := args[0]
+
 		fmt.Printf("Creating application: %v\n", name)
 		if err := application.Create(name); err != nil {
 			fmt.Println(err)
@@ -97,12 +82,19 @@ var listCmd = &cobra.Command{
 	Short: "lists all the applications",
 	Args:  cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
-		app, err := application.List()
+		apps, err := application.List()
 		if err != nil {
 			fmt.Println(err)
 			os.Exit(-1)
 		}
-		fmt.Print(app)
+		fmt.Printf("ACTIVE   NAME\n")
+		for _, app := range apps {
+			activeMark := " "
+			if app.Active {
+				activeMark = "*"
+			}
+			fmt.Printf("  %s      %s\n", activeMark, app.Name)
+		}
 	},
 }
 
@@ -112,6 +104,6 @@ func init() {
 	applicationCmd.AddCommand(listCmd)
 	applicationCmd.AddCommand(deleteCmd)
 	applicationCmd.AddCommand(getCmd)
-	applicationCmd.AddCommand(createCmd)
+	applicationCmd.AddCommand(applicationCreateCmd)
 	rootCmd.AddCommand(applicationCmd)
 }
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -1,42 +1,124 @@
 package application
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
+
+	"github.com/redhat-developer/ocdev/pkg/config"
 	"github.com/redhat-developer/ocdev/pkg/occlient"
 )
 
-// Create creates a new application and switches to it.
+const defaultApplication = "app"
 
+// ApplicationLabel is label key that is used to group all object that belong to one application
+// It should be save to use just this label to filter application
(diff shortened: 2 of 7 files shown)


pkg/application/application.go
```diff
@@ -1,42 +1,124 @@
...
+	cfg, err := config.New()
 	if err != nil {
-		return "", errors.Wrap(err, "unable to get the active application")
+		return "", errors.Wrap(err, "unable to get active application")
+	}
+
+	app := cfg.GetActiveApplication(project)
+	// if no Application is active use default
+	if app == "" {
+		app = defaultApplication
```
containscafeine: I don't think this should be the behavior. If no application is set, then we should return an empty string. We don't know if the default application was created or not, this might bring in inconsistency in the code because this will be called at a lot of places in the code.
This logic should be outside `GetCurrent()`

pkg/application/application.go
```diff
@@ -1,42 +1,124 @@
...
+	}
+
 	return nil
 }
 
-func List() (string, error) {
-	project, err := occlient.GetProjects()
+// List all application in current project
+func List() ([]ApplicationInfo, error) {
+	// TODO: use project abstaction
```
containscafeine: nit: "abstraction"

pkg/application/application.go
```diff
@@ -1,42 +1,124 @@
...
+	}
+
+	activeApplication, err := GetCurrent()
 	if err != nil {
-		return "", errors.Wrap(err, "unable to list applications")
+		return nil, errors.Wrap(err, "unable to list applications")
 	}
-	return project, nil
+
+	applications := []ApplicationInfo{}
```
containscafeine: `var applications []ApplicationInfo`

cmd/application.go
```diff
@@ -22,27 +8,26 @@ import (
...
-	Args:  cobra.MaximumNArgs(1),
-	Run: func(cmd *cobra.Command, args []string) {
-		var name string
-		// Set default application name if not set
-		if len(args) == 0 {
-			name = defaultApplication
-		} else {
-			name = args[0]
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
```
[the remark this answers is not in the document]
kadel: there is not use-case for using create like this

cmd/application.go
```diff
@@ -97,12 +82,19 @@ var listCmd = &cobra.Command{
...
 		}
-		fmt.Print(app)
+		fmt.Printf("ACTIVE   NAME\n")
+		for _, app := range apps {
+			activeMark := " "
+			if app.Active {
+				activeMark = "*"
+			}
+			fmt.Printf("  %s      %s\n", activeMark, app.Name)
+		}
```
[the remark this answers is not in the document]
kadel: It is not really breaking it.
It is just not showing because there is no component.
Currently, it is not displaying the empty applications.

I can fix that in following PR

pkg/application/application.go
kadel: not really, `var applications []ApplicationInfo` and `applications := []ApplicationInfo{}` means somehting different.
And I want to use `applications := []ApplicationInfo{}` ;-)

`var applications []ApplicationInfo` just declares the variable but doesn't allocate it, after this `applications` will be `nil`.
`applications := []ApplicationInfo{}`  this initializes it to empty  `slice`, after this `applications` is `[]ApplicationInfo{}`

kadel:
> @kadel the behavior is broken right now, will review again once fixed :(

how is it broken?

pkg/application/application.go
kadel: fixed

cmd/application.go
```diff
@@ -22,27 +8,26 @@ import (
...
-	Run: func(cmd *cobra.Command, args []string) {
-		var name string
-		// Set default application name if not set
-		if len(args) == 0 {
-			name = defaultApplication
-		} else {
-			name = args[0]
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
+			return fmt.Errorf("Please provide name for new application")
```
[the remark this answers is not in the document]
kadel: Cobra is doing that. Don't know why :-(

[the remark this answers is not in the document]
kadel: fixed

pkg/application/application.go
kadel: But that will mean that you will have logic for default application in multiple places :-(

You will have to handle default application after every `GetCurrent` call anyway. Why not have it inside? I think that this makes much more sense.

cmd/application.go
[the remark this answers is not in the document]
kadel: fixed in this PR in https://github.com/redhat-developer/ocdev/pull/94/commits/6eea89e6dc5bdec7ec108d60ca496e8ee1d287ed

cmd/application.go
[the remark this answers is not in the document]
containscafeine: yeah, but we should throw the right error, that only supply one name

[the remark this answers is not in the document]
kadel: ah, ok. Now I get it. Ok will fix the message.

kadel:
> ocdev application list does not display all the applications.
> $ ocdev application create node
> Creating application: node
> Switched to application: node
> $ ocdev application create node2
> Creating application: node2
> Switched to application: node2
> $ ocdev application list
> ACTIVE   NAME
>   *      node2

this is not possible to fix :-(

And I'm not sure if we should. Does it make sense to have empty applications that are not active?

kadel:
> $ ocdev application create node
> Creating application: node
> Switched to application: node
> $ ocdev application create node
> Creating application: node
> Switched to application: node

This will be also a bit more complicated. I would rather do it in another PR.

containscafeine:
> Does it make sense to have empty applications that are not active?

I think it's very unintuitive :( When I create an application using `ocdev application create <name>`, then I expect to see it in `ocdev application list`.

pkg/application/application.go
containscafeine: But then this is not the correct information :( The current application is _not_ `defaultApp`, it should be set as `defaultApp` outside this function.
It does not have to be set everywhere, instead like it's currently in master, it's set once in `cmd`, and then we pass the application name to all functions instead of calling `GetCurrent()` everywhere,

kadel:
> I think it's very unintuitive :( When I create an application using ocdev application create <name>, then I expect to see it in ocdev application list.

that is fixed now. (partially)

It doesn't work if you create multiple empty applications. But I think that we can add that later.
More important is to get this in, so we remove 1:1 mapping applications to projects.

We will fix that, and I will work on it as when we get basics in.

containscafeine:
Okay, let's fix that one on priority once this goes in.
Let's fix https://github.com/redhat-developer/ocdev/pull/94/files#r169026875 then this can go in

pkg/application/application.go
kadel: I can't be as it is in the master.
Because following has to be possible.
```
# no application exists or is active
ocdev component create nodejs
```
(use case 19)

Before it was possible because we couldn't have no active application because  application was project.
But not this state is possible.
That means that everytime you call `GetCurrent` you have to also handle state where there is no application.
And this can't lead to error.

containscafeine: okay

Issue #82, opened 2018-02-13
Verify the cluster that you're trying to work with is up


jorgemoralespou:
As the tool relies on "oc" command, it can happen that a developer uses "ocdev" with a context selected in .kube/config of a cluster not accesible. A check should be made, or this error should be properly captured and reported.

```
$ ocdev application list
unable to list applications: unable to get projects: command: [/usr/local/bin/oc get project -o custom-columns=NAME:.metadata.name] failed to run:
Unable to connect to the server: dial tcp 192.168.64.40:8443: i/o timeout
: exit status 1
```

containscafeine:
note to self: we can fix this by trying connecting to server from `oc whoami --show-server` and proceed with the result.

kadel:
https://github.com/redhat-developer/ocdev/issues/125

containscafeine:
Closed in #99

PR #99, opened 2018-02-16
Check if OpenShift server is running

containscafeine:
This commit adds an `IsServerUp()` function which checks if the
endpoint of the server is reachable before firing requests.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -4,13 +4,18 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
+	"net/url"
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
+
+const ocRequestTimeout = 1 * time.Second
 
 // getOcBinary returns full path to oc binary
 // first it looks for env variable KUBECTL_PLUGINS_CALLER (run as oc plugin)
@@ -66,6 +71,9 @@ func runOcComamnd(command *OcCommand) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	if !IsServerUp() {
+		return nil, errors.New("server is down")
+	}
 
 	cmd := exec.Command(ocpath, command.args...)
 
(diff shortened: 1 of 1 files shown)


containscafeine:
Fix #82

pkg/occlient/occlient.go
```diff
@@ -107,6 +118,63 @@ func runOcComamnd(command *OcCommand) ([]byte, error) {
...
 }
 
+func isLoggedIn() bool {
+	ocpath, err := getOcBinary()
+	if err != nil {
+		log.Debug(errors.Wrap(err, "unable to find oc binary"))
+		return false
+	}
+
+	cmd := exec.Command(ocpath, "whoami")
```
kadel: why is this not using `runOcComamnd` ?

pkg/occlient/occlient.go
```diff
@@ -107,6 +118,63 @@ func runOcComamnd(command *OcCommand) ([]byte, error) {
...
+	server := strings.TrimSpace(string(output))
+	u, err := url.Parse(server)
+	if err != nil {
+		log.Debug(errors.Wrap(err, "unable to parse url"))
+		return false
+	}
+
+	var connectionError error
+	for i := 1; i < 3; i++ {
+		_, connectionError = net.DialTimeout("tcp", u.Host, time.Duration(ocRequestTimeout))
```
kadel: You don't need to manually check this like this. 
`oc whoami` is communicating with the sever side so you can use just that.

```
▶ oc whoami
Unable to connect to the server: dial tcp 192.168.64.5:8443: i/o timeout
```

containscafeine: `oc whoami` might also fail for different reasons, e.g. if a user is not logged in i.e. the user is `system:anonymous`, it's going to fail.

kadel: yes, but the message that it outputs is pretty descriptive :-)

containscafeine: I think we want to hide that message, which is why :o

kadel: ok.

pkg/occlient/occlient.go
```diff
@@ -107,6 +118,63 @@ func runOcComamnd(command *OcCommand) ([]byte, error) {
...
+
+	server := strings.TrimSpace(string(output))
+	u, err := url.Parse(server)
+	if err != nil {
+		log.Debug(errors.Wrap(err, "unable to parse url"))
+		return false
+	}
+
+	var connectionError error
+	for i := 1; i < 3; i++ {
```
kadel: I'm not sure if it makes sense to have retries here. Especially if they are directly one after another.

`ocdev` is command line tool that is designed for users to run it manually. I don't think we need retries here.
If there is an error, user will see it, fix it, and run it again.
If it would be daemon or long-running process than it would make sense to have retries.

containscafeine: makes sense, removed the loop

PR #106, opened 2018-02-19
add fpm to travis deploy build

kadel:


kadel:
broke travis deploy job with #92 , this fixes it

Issue #70, opened 2018-02-13
Simple `ocdev` installation/update/uninstallation


slemeur:
### Goal
The goal of this epic is to provide a simple installation flow for `ocdev` as well as update and uninstallation. 
While the installation might not be difficult, we should make it really simple so the users could (ideally) install `ocdev` with a single instruction line. 

#### Getting Started
We need to think about the getting started experience of the end-user with `ocdev` and try to provide the best first experience. This implies a certain requirements on how the user can install `ocdev` as well as a flow in the documentation to help the user to quickly ramp up.

The installation of  `ocdev` is requiring some pre-requisites that we need to document. 

Once those pre-requisites are installed, we should be looking at how we can simplify the installation of `ocdev`. For example, on MacOS, we could be using homebrew packages, on Linux we could install from a single command line `wget -qO- https://github.com/redhat-developer/ocdev/master/blob/install-ocdev.sh | sh` . This should be properly documented too.

After having proceed of the installation, we should guide the user to verify that `ocdev` has been properly installed.

Now that `ocdev` is ready, we should provide a default basic set of first commands, something like:
- `oc login`
- `ocdev application create first-app`
- `ocdev component create nodejs`
- `ocdev ocdev component create --git https://github.com/openshift/nodejs-ex`
- `ocdev component push`

#### Staying up to date
How do we handle the updates of `ocdev` ? Could we be looking at different solutions so that `ocdev` keep itself automatically up to date?
To start, we can just put a warning to notify the user that there is a new version of `ocdev` available, and point to the installation/upgrade documentation.

#### Troubleshooting
What happens when the developer is hitting some issues? Can we provide different `DEBUG` information when interacting with the OpenShift API? 

#### Uninstallation
We need to provide a way for the developer to properly uninstall `ocdev`. 
If `ocdev` was installed with homebrew, we can have an uninstall command. 
We should provide proper set of instructions to delete on Linux and Windows too.

### Documentation
We should be providing documentation for installing `ocdev` to cover:
- Pre-requisities
- Download and Installing: MacOS, Windows, Linux
- Verify your installation
- Getting started
- Update `ocdev`cli
- Troubleshooting
- Uninstallation

### Sub Tasks
- [x] Create Homebrew package - https://github.com/redhat-developer/ocdev/issues/69
- [x] show warning when user is running older ocdev version -  https://github.com/redhat-developer/ocdev/issues/71
- [x] Create installation bash script #74
- [x] add proper documentation #75
- [ ] 
[...]

PR #117, opened 2018-02-21
Add installation script to ocdev

containscafeine:
This commit adds an installation shell script to install ocdev
on Linux machines. The script currently downloads only the latest
ocdev release from GitHub.

related to https://github.com/redhat-developer/ocdev/issues/74

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2,9 +2,14 @@
 set -e
 
 # The version of ocdev to install. Possible values - "master" and "latest"
+# master - builds from git master branch
+# latest - released versions specified by LATEST_VERSION variable
 OCDEV_VERSION="latest"
 
-GITHUB_RELEASES_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
+# Latest released ocdev version
+LATEST_VERSION="v0.0.1"
+
+GITHUB_RELEASES_URL="https://github.com/redhat-developer/ocdev/releases/download/${LATEST_VERSION}"
 BINTRAY_URL="https://dl.bintray.com/ocdev/ocdev/latest"
 
 INSTALLATION_PATH="/usr/local/bin/"
@@ -38,7 +43,7 @@ check_platform() {
         arch="amd64"
     fi
 
-    platform_type="${kernel,,}-$arch"
+    platform_type=$(echo "${kernel}-${arch}" | tr '[:upper:]' '[:lower:]')
 
     if ! echo "# $SUPPORTED_PLATFORMS" | grep "$platform_type" > /dev/null; then
         echo_stderr "
@@ -132,7 +137,7 @@ Aborting now!
     ubuntu|debian)
         echo "# Installing pre-requisites..."
         $PRIVILEGED_EXECUTION "apt-get update"
-        $PRIVILEGED_EXECUTION "apt-get install -y gnupg apt-transport-https"
+        $PRIVILEGED_EXECUTION "apt-get install -y gnupg apt-transport-https curl"
 
         echo "# "Adding GPG public key...
         $PRIVILEGED_EXECUTION "curl -L \"$DEBIAN_GPG_PUBLIC_KEY\" |  apt-key add -"
@@ -184,28 +189,30 @@ Aborting now!
         echo "# Could not identify distribution, proceeding with a binary install..."
 
         BINARY_URL=""
+        TMP_DIR=$(mktemp -d)
         case "$OCDEV_VERSION" in
         master)
             BINARY_URL="$BINTRAY_URL/$platform/ocdev"
             echo "# Downloading ocdev from $BINARY_URL"
-            curl -Lo ocdev "$BINARY_URL"
+            curl -Lo $TMP_DIR/ocdev "$BINARY_URL"
             ;;
         latest)
             BINARY_URL="$GITHUB_RELEASES_URL/ocdev-$platform.gz"
             echo "# Downloading ocdev from $BINARY_URL"
-            curl -Lo ocdev.gz "$BINARY_URL"
+            curl -Lo $TMP_DIR/ocdev.gz "$BINARY_URL"
             echo "# Extracting ocdev.gz"
-            gunzip -d ocdev.gz
+            gunzip -d $TMP_DIR/ocdev.gz
             ;;
         *)
             invalid_ocdev_version_error
         esac
 
         echo "# Setting execute permissions on ocdev"
-        chmod +x ocdev
+        chmod +x $TMP_DIR/ocdev
         echo "# Moving ocdev binary to $INSTALLATION_PATH"
-        $PRIVILEGED_EXECUTION "mv ocdev $INSTALLATION_PATH"
+        $PRIVILEGED_EXECUTION "mv $TMP_DIR/ocdev $INSTALLATION_PATH"
         echo "# ocdev has been successfully installed on your machine"
+        rm -r $TMP_DIR
         ;;
     esac
 }
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,227 @@
+#!/bin/bash
+set -e
+
+# The version of ocdev to install. Possible values - "master" and "latest"
+OCDEV_VERSION="latest"
+
+GITHUB_RELEASES_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
+BINTRAY_URL="https://dl.bintray.com/ocdev/ocdev/latest"
+
+INSTALLATION_PATH="/usr/local/bin/"
+PRIVILEGED_EXECUTION="sh -c"
+
+DEBIAN_GPG_PUBLIC_KEY="https://bintray.com/user/downloadSubjectPublicKey?username=bintray"
+DEBIAN_MASTER_REPOSITORY="https://dl.bintray.com/ocdev/ocdev-deb-dev"
+DEBIAN_LATEST_REPOSITORY="https://dl.bintray.com/ocdev/ocdev-deb-releases"
+
+RPM_MASTER_YUM_REPO="https://bintray.com/ocdev/ocdev-rpm-dev/rpm"
+RPM_LATEST_YUM_REPO="https://bintray.com/ocdev/ocdev-rpm-releases/rpm"
+
+SUPPORTED_PLATFORMS="
+darwin-amd64
+linux-amd64
+linux-arm
+"
+
+echo_stderr ()
+{
+    echo "$@" >&2
+}
+
+command_exists() {
+	command -v "$@" > /dev/null 2>&1
+}
+
+check_platform() {
+    kernel="$(uname -s)"
+    if [ "$(uname -m)" = "x86_64" ]; then
+        arch="amd64"
+    fi
+
+    platform_type="${kernel,,}-$arch"
+
+    if ! echo "# $SUPPORTED_PLATFORMS" | grep "$platform_type" > /dev/null; then
+        echo_stderr "
+# The installer has detected your platform to be $platform_type, which is
+# currently not supported by this installer script.
+
+# Please visit the following URL for detailed installation steps:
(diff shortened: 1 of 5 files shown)


scripts/install.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/sh
+set -e
+
+DOWNLOAD_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
```
kadel: version should be in separate variable so it can be easily updated

scripts/install.sh
```diff
@@ -0,0 +1,92 @@
...
+OCDEV_VERSION="latest"
```
kadel: where is this variable used, and what it is for?

scripts/install.sh
```diff
@@ -0,0 +1,92 @@
...
+    fi
+
+    echo "Downloading ocdev from $BINARY_URL"
+    curl -Lo ocdev.gz $BINARY_URL
+    echo "Extracting ocdev.gz"
+    gunzip -d ocdev.gz
+    echo "Setting execute permissions on ocdev"
+    chmod +x ocdev
+    echo "Moving ocdev binary to $INSTALLATION_PATH"
+    sudo mv ocdev "$INSTALLATION_PATH"
```
kadel: sudo, doesn't have to be installed on some system, this script should check if it is present it at the beginning and error

scripts/install.sh
```diff
@@ -0,0 +1,92 @@
...
+
+    platform="$(check_platform)"
+    echo "Detected platform: $platform"
+    echo
+
+    BINARY_URL="$DOWNLOAD_URL/$BINARY_NAME-$platform.gz"
+
+    if command_exists ocdev; then
+        echo "
+ocdev version \"$(ocdev version)\"is already installed on your system, running this installer script might case issues with your current installation.
```
kadel: missing space before "is"

scripts/install.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/sh
+set -e
+
+DOWNLOAD_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
+OCDEV_VERSION="latest"
+BINARY_NAME="ocdev"
+CANCELLATION_TIMEOUT=15
+INSTALLATION_PATH="/usr/local/bin/"
+
+SUPPORTED_DISTRIBUTIONS="
```
kadel: SUPPORTED_PLATFORMS might be a better name

scripts/install.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/sh
+set -e
+
+DOWNLOAD_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
+OCDEV_VERSION="latest"
+BINARY_NAME="ocdev"
+CANCELLATION_TIMEOUT=15
```
kadel: what does this delay do? can we have a comment explaining it?

kadel:
Just a topic for discussion.
Wouldn't it be better to also detect if it is rpm or deb distribution and install repository and packages for them? copying to `/usr/local/bin` would be fallback for other distros.

kadel:
> on MacOS it also fails ungracefully

you can use `uname -s` to detect if its linux or mac
returns `Darwin` on macos,  `Linux` on linux

scripts/install.sh
containscafeine: it's not used anywhere, removed

scripts/install.sh
containscafeine: done, added logic to check for root, if not, check for sudo, if not check for su, if not fail

scripts/install.sh
containscafeine: fixed

scripts/install.sh
containscafeine: changed

scripts/install.sh
containscafeine: added

containscafeine:
> Wouldn't it be better to also detect if it is rpm or deb distribution and install repository and packages for them? copying to /usr/local/bin would be fallback for other distros.

Correct, we should do that. How about fixing this in the next PR. This is how docker also does it.

> you can use uname -s to detect if its linux or mac

ack, I didn't know that macOS didn't have support for big commands, fixed it.

containscafeine:
> fails completely on ubuntu:

fixed

> on MacOS it also fails ungrasfully

fixed

PTAL

scripts/install.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+set -e
```
kadel: can we please remove that :-(
it is nice for debugging, but it shouldn't be in final script

scripts/install.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo
+
+    BINARY_URL="$DOWNLOAD_URL/$BINARY_NAME-$platform.gz"
+
+    if command_exists ocdev; then
+        echo "
+ocdev version \"$(ocdev version)\" is already installed on your system, running this installer script might case issues with your current installation.
+Press Ctrl+C to abort installation.
+"
+        ( set -x; sleep "$CANCELLATION_TIMEOUT" )
```
kadel: wouldn't it be safer to just exit here?

scripts/install.sh
```diff
@@ -0,0 +1,117 @@
...
+	command -v "$@" > /dev/null 2>&1
+}
+
+check_platform() {
+    kernel="$(uname -s)"
+    if [ "$(uname -m)" = "x86_64" ]; then
+        arch="amd64"
+    fi
+
+    platform_type="${kernel,,}-$arch"
```
kadel: this is a problem on macos as. mac has bash 3.2 but this works only on bash 4.0< 

if you need convert string to lover-case use some other method, like  `tr` (`echo "$a" | tr '[:upper:]' '[:lower:]'`)

kadel:
> Correct, we should do that. How about fixing this in the next PR. This is how docker also does it.

woudln't it be better to do it from the beggining?

I'm little bit afraid that if we pospone this, than we will never do it.

scripts/install.sh
containscafeine: makes sense

scripts/install.sh
containscafeine: do we need to support installation on mac using this script at all?
I don't think this is mentioned in #70, nor did I see other packages like docker (ref: https://get.docker.com/). brew works fine there right?

containscafeine:
> woudln't it be better to do it from the beggining?
I'm little bit afraid that if we pospone this, than we will never do it.

makes sense, will fix

scripts/install.sh
kadel: brew is the same as packages for linux, and we are making this script for it.

This script should be generic and work everywhere. Detect what system it is.
if it is debian based use deb packages, if it is rpm based use rpm packages, if it is other linux copy bin to /usr/local/bin, if it is macos use brew...

containscafeine: Okay. I don't have a mac to test this on :(
How about you add macOS detection once this is merged with deb and rpm?

kadel: on mac  "`uname -s` == Darwin"

for mac first check  if brew is installed (if there is a `brew` command) fail if not, and if there is then run two commands that are in the readme.  

I'll test it and debug/fix if there are problems.

containscafeine:
@kadel made the changes. I've testing on CentOS, Fedora, Ubuntu and Debian latest container images and it works, PTAL!

scripts/install.sh
kadel: ah, sorry, my fault. I thought this is `set -x` 😬

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
...
+        echo_stderr "# 
+ocdev version \"$(ocdev version)\" is already installed on your system, running this installer script might case issues with your current installation. If you want to install ocdev using this script, please remove the current installation of ocdev from you system.
+Aborting now!
+"
+        exit 1
+    fi
+
+    set_privileged_execution
+
+  	distribution=$(get_distribution)
```
kadel: this has weird indentation

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
...
+	command -v "$@" > /dev/null 2>&1
+}
+
+check_platform() {
+    kernel="$(uname -s)"
+    if [ "$(uname -m)" = "x86_64" ]; then
+        arch="amd64"
+    fi
+
+    platform_type="${kernel,,}-$arch"
```
kadel: this line is still failing o macOS,

```
▶ ./install.sh
# Starting ocdev installation...
# Detecting distribution...
./install.sh: line 43: ${kernel,,}-$arch: bad substitution
```
will fix it tomorrow, before we merge this

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
...
+
+        $PRIVILEGED_EXECUTION "apt-get update"
+        $PRIVILEGED_EXECUTION "apt-get install -y ocdev"
+        ;;
+
+    centos|fedora)
+        package_manager=""
+        case "$distribution" in
+        fedora)
+            package_manager="dnf"
```
kadel: is this really true? from what version is `dnf` preinstalled in fedora?

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
...
+DEBIAN_LATEST_REPOSITORY="https://dl.bintray.com/ocdev/ocdev-deb-releases"
+
+RPM_MASTER_YUM_REPO="https://bintray.com/ocdev/ocdev-rpm-dev/rpm"
+RPM_LATEST_YUM_REPO="https://bintray.com/ocdev/ocdev-rpm-releases/rpm"
+
+SUPPORTED_PLATFORMS="
+darwin-amd64
+linux-amd64
+linux-arm
+windows-amd64
```
kadel: should be windows here, if it is not handled by this script?

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/bash
+set -e
+
+OCDEV_VERSION="latest"
```
kadel: this should have commend what it is and what values are possible and

scripts/install.sh
```diff
@@ -0,0 +1,208 @@
...
+GITHUB_RELEASES_URL="https://github.com/redhat-developer/ocdev/releases/download/v0.0.1"
+BINTRAY_URL="https://dl.bintray.com/ocdev/ocdev/latest"
+
+# CANCELLATION_TIMEOUT specifies the number of seconds the script will wait
+# for the user to cancel the execution before continuing on.
+CANCELLATION_TIMEOUT=15
```
kadel: this is not used anywhere

kadel:
We need to put instruction on how to use this script into readme.

scripts/install.sh
containscafeine: fixed

scripts/install.sh
containscafeine: fixed

scripts/install.sh
containscafeine: fixed

scripts/install.sh
containscafeine: removed

scripts/install.sh
containscafeine: added

scripts/install.sh
```diff
@@ -132,7 +137,7 @@ Aborting now!
     ubuntu|debian)
         echo "# Installing pre-requisites..."
         $PRIVILEGED_EXECUTION "apt-get update"
-        $PRIVILEGED_EXECUTION "apt-get install -y gnupg apt-transport-https"
+        $PRIVILEGED_EXECUTION "apt-get install -y gnupg apt-transport-https curl"
```
containscafeine: I think this will fail because curl is run even before this script.

kadel: for debian, where?

kadel:
if you think that tests are not necessary, feel free to remove it from this pr ;-) it is your branch ;-)

PR #120, opened 2018-02-21
 Update user if newer release available

containscafeine:
645f721 (Shubham Minglani, 2 minutes ago)
   Update user if newer release available

   This commit notifies the user if a newer ocdev release is available than
   the release that they have installed right now.

   If there is an error getting the new release, the error is ignored but
   logged, because this should not block program execution.

4a87980 (Shubham Minglani, 3 minutes ago)
   Change VERSION to v0.0.1

   This commit changes the VERSION that is reflected in `ocdev version` to
   v0.0.1 which is in sync with our release versions on GitHub as well as the
   rest of the Kubernetes, OpenShift ecosystem.

083870d (Shubham Minglani, 5 minutes ago)
   vendor add: google/go-github and blang/semver

   This commit updates vendor/ to bundle in the following dependencies:
   - github.com/google/go-github
   - github.com/blang/semver


containscafeine:
The output looks like -
```console
$ ocdev

A newer version of ocdev (version: v0.0.1) is available.

Update using your package manager, or run 

curl https://raw.githubusercontent.com/redhat-developer/ocdev/master/scripts/install.sh | sh

to update manually, or visit https://github.com/redhat-developer/ocdev/releases
```

pkg/util/util.go
```diff
@@ -0,0 +1,41 @@
...
+	"strings"
+)
+
+const (
+	ghorg            = "redhat-developer"
+	ghrepo           = "ocdev"
+	InstallScriptURL = "https://raw.githubusercontent.com/redhat-developer/ocdev/master/scripts/install.sh"
+)
+
+func GetNewerReleaseTag(currentVersion string) (string, bool, error) {
```
kadel: can we have please comments describing what this function is and what it is returning

kadel: shouldn't this be named `GetLatestReleasedTag` or `GetLatestRelease`?

kadel: it might be also better to have a special package for this, rather than putting it into generic `util`.`updater` or `notifier` might be good names

cmd/root.go
```diff
@@ -60,4 +61,16 @@ func init() {
...
 	rootCmd.PersistentFlags().BoolVarP(&GlobalVerbose, "verbose", "v", false, "verbose output")
 
+	newTag, isNew, err := util.GetNewerReleaseTag(VERSION)
+	if err != nil {
+		// The error is intentionally not being handled because we don't want
+		// to stop the execution of the program because of this failure
+		log.Infof("Error checking if newer ocdev release is available: %v", err)
+	}
+	if isNew {
+		fmt.Printf("\nA newer version of ocdev (version: " + fmt.Sprint(newTag) + ") is available.\n\n" +
```
kadel: can we please remove empty lines?  I don't think we need to have extra empty line after each line of text.

But on the other hand, I would add one extra empty line after the end.

kadel:
related to https://github.com/redhat-developer/ocdev/issues/71

pkg/util/util.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"github.com/blang/semver"
+	"github.com/google/go-github/github"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+const (
+	ghorg            = "redhat-developer"
```
kadel: please add comments explaining what those variables are and what they are used for

kadel:
Just noticed one thing: It noticeably slows  `ocdev` execution even on a really fast internet connection.

pkg/util/util.go
containscafeine: - added comments
- renames to `CheckLatestReleaseTag`
- changed package to `notify`
- broke the function down to 2 functions, one to get the latest release tag from GH, another is the exported function
- changed function signature not to return a bool, instead make decisions if the string is empty or not

pkg/util/util.go
containscafeine: done

cmd/root.go
containscafeine: ack, removed

containscafeine:
@kadel updated the PR, and yeah, this slows it down :|
I think we should poll once in the program execution in a non-blocking way, and store it to the config file, and not check for errors or something, and only read from the config file during execution. But yeah, let's do that in another PR.

kadel:
created https://github.com/redhat-developer/ocdev/issues/122 to address issues with current implementation.

containscafeine:
@kadel yeah, I think we can merge it for now, it adds about half a second delay, but should be fine for now

kadel:
> @kadel yeah, I think we can merge it for now, it adds about half a second delay, but should be fine for now


But that delay is not small. :-(
Can you please start working on config option to disable this as soon as this is merged?

PR #128, opened 2018-02-26
Allow setting active application and component

### LINKED ISSUE #68: allow switching active application and component
`ocdev application <name>` -  change active aplication to `<name>` application

`ocdev component <name>` - change active component to `<name>`

kadel:
fixes #68 

Currently implemented as
```
ocdev component set <name>
ocdev application set <name>
```

In use case document this is used without subcommand like this:
```
ocdev application <name>
ocdev component <name>
```
But the problem with not using subcommand is that it puts restrictions on application names. For example, `list` can't be used as application name, because there is `ocdev application list` subcommand.
```
▶ ./ocdev application create list
Creating application: list
Switched to application: list

▶ ./ocdev application list
ACTIVE   NAME
         f
         app
         myapp
  *      list
```

I've chosen `set` as it is short and descriptive. I was also considering `activate` and `use`.

### DIFF
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -1,6 +1,8 @@
 package application
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/redhat-developer/ocdev/pkg/config"
 	"github.com/redhat-developer/ocdev/pkg/occlient"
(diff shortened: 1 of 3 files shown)


kadel:
> set is also creating an application which does not exist right now :(

fixed.
now it shows error:
```
▶ ./ocdev app set ggggg
application ggggg doesn't exists
```

the similar problem is for the component, but that can't be implemented right now because we are missing component.List()

pkg/application/application.go
```diff
@@ -169,6 +171,24 @@ func SetCurrent(name string) error {
...
+	found := false
+	for _, app := range apps {
+
+		if app.Name == name {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return fmt.Errorf("application %s doesn't exists", name)
```
containscafeine: s/exists/exist

containscafeine:
created #143 to track this

kadel:
> LGTM except one typo, @kadel feel free to merge once that's fixed!

fixed

PR #129, opened 2018-02-26
Get latest release info in a non-blocking manner

### LINKED ISSUE #122: Improve update checker
Current implementation noticeably slows ocdev execution even on a really fast internet connection.

We definitely should have an option to disable this check permanently. We could have something like ocdev config set checkForUpdates false and use config file to persist configuration.

We could also perform check only once a day (use config file to persist when check occurred last time)
we can do those in separate PR, but think that we definitely have to add a way to disable it.

Another option could be to perform check as async operation.

---
Update: 2018-07-12

## Summary:
There are two things that are needed.

### First
We need a configuration uption  that will disable update checker.

There should be new commadn `odo config`
```
# disable update notification
odo config set UpdateNotification false

# view current value
odo config get UpdateNotification

UpdateNotification = false
```

Value of `UpdateNotification` is saved in config file.
There is `Config` struct in  `config.go` that represent ODO config file structure. (https://github.com/redhat-developer/odo/blob/4567402e4157834c75afc1e632f7b70dc218751d/pkg/config/config.go#L31)
There should be a new field called `UpdateNotification`.
`odo config set UpdateNotification false` sets value of this field to false, (or true if `odo config set UpdateNotification true` is called)

Update checker should be started only when this configuration field is set to `true`.
Default value if the configuration is not specified is `true`

### Second
We need a better way to check for updates that don't require calling GitHub API.

TBD

containscafeine:
Prior to this commit, the latest release information was fetched
from GitHub in a blocking manner, which means that even for a very
quick command like `ocdev version`, the tool would wait for a few
seconds before outputting the version, because the execution would
be blocked on waiting for the response from GitHub APIs.

This commit changes that and makes the request to GitHub
concurrent and non-blocking. This means that if the request has
not completed but the program execution has ended, then the tool
will exit gracefully without waiting for the request to complete.

However, if the program execution took longer than the request and
the response has been received by the goroutine before the program
execution is over, then, if applicable, the information is
printed.

Fixes #122

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -47,9 +47,20 @@ var rootCmd = &cobra.Command{
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+
+	updateInfo := make(chan string)
+	go getLatestReleaseInfo(updateInfo)
+
 	if err := rootCmd.Execute(); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
+	}
+
+	select {
+	case message := <-updateInfo:
+		fmt.Printf(message)
+	default:
+		log.Debug("Could not get the latest release information in time. Never mind, exiting gracefully :)")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


containscafeine:
non-blocking notifier

containscafeine:
@kadel nice catch, I've fixed it, can you check now?
As for disabling this, sure, we can have it later when we add a config subcommand.

kadel:
it doesn't panic anymore when offline, but it still shows ugly message :-(

```
▶ ./ocdev app get
INFO[0000] Error checking if newer ocdev release is available: unable to get latest release tag: Get https://api.github.com/repos/redhat-developer/ocdev/releases/latest: dial tcp: lookup api.github.com: no such host
The current application is: list
```

kadel:
> As for disabling this, sure, we can have it later when we add a config subcommand.

I think we need this rather sooner than later. I would even say that having the option to disable this completely is even more important than doing it async.

If someone wants to keep specific `ocdev` version this message will annoy him/ her every time `ocdev` is run.

Issue #39, opened 2018-02-06
change cobra cmd variable names


containscafeine:

kadel:
@containscafeine change how?

PR #132, opened 2018-02-28
Cleanup cobra names.

kadel:
Use the same naming pattern for all cobra commands variables.
The patter is camel-cased `<root command name><subcommand name>Cmd`. Example `applicationListCmd`.


kadel:
@containscafeine is this what you had in mind in #39 ?

PR #141, opened 2018-03-02
update commands structure to match it to what is in use cases

### LINKED ISSUE #86: There is no command "ocdev push" but "ocdev component push"
"Use case 24: As a developer, I want to be able to push local directory contents or a binary artifact." defines that:

"As a developer, I want to be able to push local directory contents or an artifact, if I have built a binary artifact that can be directly pushed. The push commands will default to push contents of the local directory, but with a flag the behaviour changes to push a binary artifact."

The push command should push whatever is set, based on the create command mode/flags. By default local directory content, but if the component has been created from a binary, it should try to push the binary.

kadel:
- change `component push` to `push`
  Push is now a root level command as specified in use cases.
   fixes #86
- change all `os.Exit(-1)` to `os.Exit(1)` 
  (exit code should be between 0-255)
-  `ocdev app` and `ocdev component` shows current app/component
    `ocdev application` is the same as `ocdev application get`
    `ocdev component` is the same as `ocdev component get`
- add `ocdev create` as an alias for `ocdev component create`

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -17,6 +17,8 @@ var applicationCmd = &cobra.Command{
 	Use:     "application",
 	Short:   "application",
 	Aliases: []string{"app"},
+	// 'ocdev application' is the same as 'ocdev application get'
+	Run: applicationGetCmd.Run,
 }
 
 var applicationCreateCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


containscafeine:
@kadel LGTM, let's merge once rebased

PR #179, opened 2018-03-05
Fix application create error

containscafeine:
This commit moves the logic of checking if an application exists
outside of SetCurrent()

### DIFF
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -1,8 +1,6 @@
 package application
 
 import (
-	"fmt"
-
 	"github.com/pkg/errors"
 	"github.com/redhat-developer/ocdev/pkg/config"
 	"github.com/redhat-developer/ocdev/pkg/occlient"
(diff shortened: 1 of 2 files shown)


kadel:
👍 thanks for catching that

pkg/application/application.go
```diff
@@ -201,3 +183,17 @@ func SetCurrent(name string) error {
 
 	return nil
 }
+
+func Exists(name string) error {
+	apps, err := List()
+	if err != nil {
+		return errors.Wrap(err, "unable to get active application")
```
kadel: error message is not correct :-(

pkg/application/application.go
```diff
@@ -201,3 +183,17 @@ func SetCurrent(name string) error {
 
 	return nil
 }
+
+func Exists(name string) error {
```
kadel: shouldn't this also return bool?

It is not possible to determine if application doesn't exist or if there were error while checking it

pkg/application/application.go
containscafeine: fixed

pkg/application/application.go
containscafeine: correct, it should return both

containscafeine: fixed

PR #187, opened 2018-03-06
Add init() to package occlient

containscafeine:
This commit adds an init() function to the package occlient, which
does the initializations required for the package like checking
if the server is up, the user is logged in, and most importantly
sets the path for oc binary for the entire package to use.

This is a significant improvement in performance because before
this commit, this logic was a part of the `runOcCommand()`, which
means that every time any part of code called this function, the
`oc` binary was found again in the $PATH and connection to the
server and user being logged in was always checked.

This also reduces the logging output by a lot, which now looks
pretty concise like the following.

```console
$ ocdev application list -v
DEBU[0000] running oc command with arguments: /usr/bin/oc project

DEBU[0000] running oc command with arguments: /usr/bin/oc project -q

DEBU[0000] running oc command with arguments: /usr/bin/oc get all
--selector app.kubernetes.io/name --namespace myproject -o
go-template={{range .items}}{{range $key, $value :=
.metadata.labels}}{{if eq $key "app.kubernetes.io/name"}}
{{$value}},{{end}}{{end}}{{end}}

ACTIVE   NAME
         hasd
         asdas
  *      nodeapp
DEBU[0000] Could not get the latest release information in time. Never mind, exiting gracefully :)
```


pkg/occlient/occlient.go
```diff
@@ -17,6 +17,28 @@ import (
...
+var ocpath string
+
+var isInitialized bool
+
+func initialize() {
+	var err error
+	ocpath, err = getOcBinary()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
kadel: it should return error rather than just exit whole program. 
The same for all the other `os.Exit(1)` in this function

pkg/occlient/occlient.go
```diff
@@ -17,6 +17,28 @@ import (
 
 const ocRequestTimeout = 1 * time.Second
 
+// ocpath stores the path to oc binary
+var ocpath string
+
+var isInitialized bool
+
+func initialize() {
```
kadel: do we need `isInitialized`? Wouldn't be enough just to test `ocpath`?

we can do all this in this `initialize` function.
oc path 
```go
func initialize() error {
// check if this was already initialized
if ocpath == "" {
 return nil
}

// do initialization
// return err if something went wrong
....

```

containscafeine: makes sense, much neater this way, fixed

pkg/occlient/occlient.go
containscafeine: yeah, borrowed it from the previous `init()` implementation since that does not return an error, but yeah, makes sense, fixed now

PR #188, opened 2018-03-06
Add more verbosity to test runs

containscafeine:
This commit adds `-v` to the `go test` commands so we get more
verbose output in runs in tests, which will be easier to debug in
case of failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ prepare-release: cross
 
 .PHONY: test
 test:
-	go test -race $(PKGS)
+	go test -v -race $(PKGS)
 
 # create deb and rpm packages using fpm in ./dist/pkgs/
 # run make cross before this!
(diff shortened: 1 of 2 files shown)


kadel:
Why do you want to add it?
It will just make looking for errors harder.
If test fails it will show the same output as if it would be started with `-v`

containscafeine:
It does not show which exact test failed. I cannot see which test failed in #187

kadel:
> It does not show which exact test failed. I cannot see which test failed in #187

verbose tests is not going to help you in that case.
it failed even before any tests were run on missing `oc`
```
exec: "oc": executable file not found in $PATH
```
You have moved `oc` checking into `init()` init is executed before any other functions in test file.

kadel:
this was sorted out, we don't need to run all tests in verbose mode.

Issue #142, opened 2018-03-02
`ocdev application list` does not show application, but `get` does`


containscafeine:
```console
$ ocdev application list 
ACTIVE   NAME

$ ocdev application get
The current application is: app
```

kadel:
Yeh, this is kind of tricky :-(

We need to define what is the right behavior if there is no application.
So the list is listing existing applications, and it is empty because there is no app.

'get' is showing 'app' because by definition if there is no application the default application should be used, and default application is called 'app'

containscafeine:
@kadel this is why we should separate the setting default logic from GetCurrent() :)

kadel:
> @kadel this is why we should separate the setting default logic from GetCurrent() :)

That is not about that. It is easy to do it. Question is what is the right way to present it!

kadel:
closed in #215

PR #215, opened 2018-03-08
fix application list and get

### LINKED ISSUE #190: ocdev app should return current app, not create it if non existent
On a new cluster, new config, new install.

```
$ ocdev app list
ACTIVE   NAME

$ ocdev app
The current application is: app

$ ocdev app list
ACTIVE   NAME
  *      app
```

The command has created the app, and use it as default current app.

See #152

kadel:
This fixes bug where default application was created when listing apps or
just getting the current app.

- If `ocdev app` is run and there are no applications, a message is shown
  informing user about this fact. Previously it showed that default application is active.
- When `ocdev create ...` is run and there are no application, default
  the application is automatically created and used.

fixes #190 and #142 

TODO:
 - [x] fix tests

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -58,9 +58,13 @@ var applicationGetCmd = &cobra.Command{
 		}
 		if applicationShortFlag {
 			fmt.Print(app)
-		} else {
-			fmt.Printf("The current application is: %v\n", app)
+			return
 		}
+		if app == "" {
+			fmt.Printf("There's no active application.\nYou can create one by running 'ocdev application create <name>'.")
+			return
+		}
+		fmt.Printf("The current application is: %v\n", app)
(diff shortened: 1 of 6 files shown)


cmd/application.go
```diff
@@ -58,9 +58,14 @@ var applicationGetCmd = &cobra.Command{
 		}
 		if applicationShortFlag {
 			fmt.Print(app)
-		} else {
-			fmt.Printf("The current application is: %v\n", app)
+			return
 		}
+		if app == "" {
+			fmt.Printf("There's no active application. You need to create one\n")
```
containscafeine: You can create one by running `ocdev application create <name>`

kadel: updated

PR #235, opened 2018-03-13
Confirm before deleting application

containscafeine:
This commit prompts the user before deleting the application when
`ocdev application delete name` command is run.

Also, a "--force/-f" flag has been introduced which skips this
prompt if set.

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -6,10 +6,12 @@ import (
 
 	"github.com/redhat-developer/ocdev/pkg/application"
 	"github.com/spf13/cobra"
+	"strings"
 )
 
 var (
-	applicationShortFlag bool
+	applicationShortFlag       bool
+	applicationForceDeleteFlag bool
 )
 
 // applicationCmd represents the app command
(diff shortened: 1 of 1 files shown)


kadel:
Can you please link this PR to the issue?

cmd/application.go
```diff
@@ -78,12 +80,27 @@ var applicationDeleteCmd = &cobra.Command{
...
+		if strings.ToLower(confirmDeletion) == "y" {
+			err := application.Delete(appName)
+			if err != nil {
+				fmt.Println(err)
+				os.Exit(1)
+			}
+			fmt.Printf("Deleted application: %s\n", args[0])
+		} else {
+			fmt.Printf("Aborting deletion of application: %v\n", appName)
+			os.Exit(0)
```
kadel: don't use exit here if you are not returning error code, just `return`. Or in this case, there doesn't have to be anything

containscafeine: fixed

containscafeine:
> Can you please link this PR to the issue?

The issue is for component, I'll send a PR for it now

kadel:
> The issue is for component, I'll send a PR for it now

what about this one? https://github.com/redhat-developer/ocdev/issues/203

also related to related to https://github.com/redhat-developer/ocdev/issues/202

containscafeine:
missed that.
I've made the changes, can we merge this?

Issue #232, opened 2018-03-12
refactor occlient pkg


kadel:
create interface for `occlient` so we can mock it in tests

kadel:
no longer relevant as we slowly moving everything to client-go

PR #236, opened 2018-03-13
decouple Create and SetCurrent for app and component

kadel:
fixes https://github.com/redhat-developer/ocdev/issues/180

Also, this will make it easier to write unit tests when #232 is done

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -180,12 +180,7 @@ func (c *ConfigInfo) SetActiveApplication(application string, project string) er
 
 	// if application doesn't exists, add it as Active
 	if !found {
-		c.ActiveApplications = append(c.ActiveApplications,
-			ApplicationInfo{
-				Name:    application,
-				Project: project,
-				Active:  true,
-			})
+		return fmt.Errorf("unable set application %s as active in config, it doesn't exist", application)
 	}
 	// make sure that no other application is active
 	for i, app := range c.ActiveApplications {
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -201,6 +196,36 @@ func (c *ConfigInfo) SetActiveApplication(application string, project string) er
...
+
+	for i, app := range c.ActiveApplications {
+		// if application exists set is as Active
+		if app.Name == application && app.Project == project {
+			c.ActiveApplications[i].Active = true
+			return fmt.Errorf("unable to set add %s application, it already exists in config file", application)
+		}
+	}
+
+	// if application doesn't exists, add it as Active
```
containscafeine: should this comment be here, we are not setting the application as active

pkg/config/config.go
```diff
@@ -201,6 +196,36 @@ func (c *ConfigInfo) SetActiveApplication(application string, project string) er
...
 }
 
+// AddApplication add  new application to the config file
+// Newly create application is NOT going to be se as Active.
+func (c *ConfigInfo) AddApplication(application string, project string) error {
+	if c.ActiveApplications == nil {
+		c.ActiveApplications = []ApplicationInfo{}
+	}
+
+	for i, app := range c.ActiveApplications {
```
containscafeine: this loop sets the application as active. we already have `SetActiveApplication()`, should we have this here?

kadel: Thanks for catching this.
Setting application as active shouldn't be there. This loop is there only to error out if the application already exists.

fixed

pkg/config/config.go
kadel: correct, I  forgot to update that comment.

fixed

PR #237, opened 2018-03-14
Prompt before deleting component

### LINKED ISSUE #200: delete a component should always ask for confirmation
When deleting a component, we need to ask for confirmation to the user. A --force flag could be introduced as an override of this, to automate tasks.

```
$ ocd delete
Do you really want to delete myfrontend in application <name> (y/n)?
```

Subtask of #156

containscafeine:
The user will be prompted before deletion of a component, unless
overriden using --force/-f.

Fixes #200

### DIFF
--- a/cmd/component.go
+++ b/cmd/component.go
@@ -6,17 +6,19 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/redhat-developer/ocdev/pkg/application"
 	"github.com/redhat-developer/ocdev/pkg/component"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
 var (
-	componentBinary    string
-	componentGit       string
-	componentDir       string
-	componentShortFlag bool
-	componentApp       string
+	componentApp             string
+	componentBinary          string
+	componentGit             string
+	componentDir             string
+	componentShortFlag       bool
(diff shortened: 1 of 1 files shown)


cmd/component.go
```diff
@@ -110,15 +111,26 @@ var componentDeleteCmd = &cobra.Command{
...
 
-		// no flag was set, create empty component
-		output, err := component.Delete(componentName)
-		if err != nil {
-			fmt.Println(err)
-			os.Exit(1)
+		if componentForceDeleteFlag {
+			confirmDeletion = "y"
+		} else {
+			fmt.Printf("Are you sure you want to delete the component: %v? [y/N] ", componentName)
```
kadel: It should display from what application the component will be deleted.

kadel:
ping @containscafeine

cmd/component.go
containscafeine: fixed

cmd/component.go
```diff
@@ -110,15 +111,26 @@ var componentDeleteCmd = &cobra.Command{
...
+			fmt.Scanln(&confirmDeletion)
 		}
-		fmt.Println(output)
 
+		if strings.ToLower(confirmDeletion) == "y" {
+			// no flag was set, create empty component
```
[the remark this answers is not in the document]
containscafeine: removed

PR #239, opened 2018-03-15
Added the ocdev project get command

mik-dass:
Added the project get command https://github.com/redhat-developer/ocdev/issues/96

### DIFF
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -1,0 +1,92 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/redhat-developer/ocdev/pkg/project"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+var (
+	projectShortFlag bool
+)
+
+var projectCmd = &cobra.Command{
+	Use:   "project [options]",
+	Short: "project",
+	Run:   projectGetCmd.Run,
+}
+
+var projectSetCmd = &cobra.Command{
+	Use:   "set",
+	Short: "set the current active project",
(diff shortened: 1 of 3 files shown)


mik-dass:
@TomasTomecek I am working on the `ocdev project <project_name> `command currently. I found out that if the user does oc project <project_name> with the current project_name, the command returns 
`Already on project "<project_name>" on server "https://192.168.42.29:8443"`
Should I replicate this on the ocdev tool too?

kadel:
>I am working on the ocdev project <project_name>command currently.

@mik-dass to switch in between project the command should be `ocdev project set`


> I found out that if the user does oc project <project_name> with the current project_name, the command returns
> Already on project "<project_name>" on server "https://192.168.42.29:8443"

Yes, it should show display message like this indicating that you are already using that project.

mik-dass:
@kadel I have added all the commands and I think it's ready for review. Not sure about the codecov thing though. Should I add tests for it?

kadel:
If you can figure out how to add unit tests to this it would be great. But it is not necessary.
Currently, we don't have a easy way to mock `oc` calls :-(  So adding unit tests might be tricky :-(

mik-dass:
@kadel Ok I will look into it :)

kadel:
Sorry about that @mik-dass , but after #245 was merged you code needs small refactor :-(

I'll merge it as soon as it is done, so you don't end up in the same situation with another PR.

mik-dass:
@kadel Done

PR #241, opened 2018-03-16
Use client go

kadel:
So far I was able to replace stuff that we did via `oc new-app` and `oc start-build`.  The rest should be straightforward.


kadel:
This is ready to review.
I've tried to keep changes to `occlient.go`. 

Currently, `occlient.go` is a weird hybrid.  Some parts are handled via executing `oc` others are via `client-go`. 
This PR should go in so we can start building new stuff on top of `client-go.` I'll clean the rest of `occlient.go` in the following PRs.

Issue #35, opened 2018-02-06
implement catalog search


containscafeine:
We can use queries like `oc new-app --search --image-stream=httpd` to implement `ocdev component search <component name> `

containscafeine:
or `ocdev component type search`

kadel:
this should be `ocdev catalog search` as per use-case 9

containscafeine:
fixed in #243

PR #243, opened 2018-03-20
Add ocdev catalog command

### LINKED ISSUE #194: Search all possible components I can deploy
As a developer, I want to be able to list/search all possible components I can use with my application

```
$ ocd catalog search [postgres] [--type=database|java|whatever] 
The following items were found:
(postgresql) PostgreSQL
(postgresql-10.0) PostgreSQL-10.0
(crunchydb) CrunchyDB Clustered (PostgreSQL 9.5)
```
Search has to be able to filter. (Possibly create another issue for this).

Subtask for #154

containscafeine:
Fix #194 #35


containscafeine:
@kadel would love your initial feedback

pkg/occlient/occlient.go
```diff
@@ -380,10 +385,25 @@ func getExposedPorts(image *imagev1.ImageStreamImage) ([]corev1.ContainerPort, e
 	return ports, nil
 }
 
+func GetDefaultBuilderImages() ([]string, error) {
+	return GetImageStreams(openShiftNameSpace)
+}
+
+func GetImageStreams(namespace string) ([]string, error) {
```
kadel: It might be better to call this function `GetImageStreamNames` if it is returning just names.
Also comment would be nice ;-)

pkg/occlient/occlient.go
```diff
@@ -380,10 +385,25 @@ func getExposedPorts(image *imagev1.ImageStreamImage) ([]corev1.ContainerPort, e
 	return ports, nil
 }
 
+func GetDefaultBuilderImages() ([]string, error) {
```
kadel: Comment would be nice ;-)

pkg/occlient/occlient.go
```diff
@@ -49,6 +52,41 @@ var (
 	namespace   string
 )
 
+func init() {
```
kadel: We shouldn't use init for this :-(
We will have to construct struct that will represent openshift client and that will hold all the `client-go` clients

kadel:
a few pointers, but in general I think goes in the right direction

containscafeine:
@kadel made the changes, also added the service catalog support, PTAL!

pkg/occlient/occlient.go
```diff
@@ -793,3 +827,63 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
...
+		if is == name {
+			return true
+		}
+	}
+	return false
+}
+
+// clusterServiceClassExists returns true if the given external name of the
+// cluster service class exists in the cluster, and false otherwise
+func (c *Client) clusterServiceClassExists(name string) bool {
```
containscafeine: we're not using this in this PR, but we will need this when we implement deploying using s2i with this, so maybe we can keep this util function here

pkg/occlient/occlient.go
```diff
@@ -793,3 +827,63 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
...
+
+	for _, class := range classes {
+		classNames = append(classNames, class.Spec.ExternalName)
+	}
+	return classNames, nil
+}
+
+// imageStreamExists returns true if the given image stream exists in the given
+// namespace
+func (c *Client) imageStreamExists(name string, namespace string) bool {
```
containscafeine: we're not using this in this PR, but we will need this when we implement deploying using s2i with this, so maybe we can keep this util function here

pkg/catalog/catalog.go
```diff
@@ -0,0 +1,73 @@
...
+			result = append(result, component)
+		}
+	}
+
+	return result, nil
+}
+
+// getDefaultBuilderImages returns the default builder images available in the
+// openshift namespace
+func getDefaultBuilderImages(client *occlient.Client) ([]string, error) {
```
kadel: Shouldn't this be in `occlient` pkg?

pkg/occlient/occlient.go
```diff
@@ -34,18 +35,23 @@ import (
 	"github.com/openshift/source-to-image/pkg/tar"
 	s2ifs "github.com/openshift/source-to-image/pkg/util/fs"
 
+	"github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
```
kadel: This should be named properly. `v1beta1` is too generic and it will collide with other objects.
maybe `catalogv1beta1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"` ?

kadel:
We have somehow display that this requires service catalog. Probably in the readme and also in help messages.

Another thing is that it displays a lot of stuff that currently can't be deployed (everything from service catalog). We should hide it from output until `component create`can handle that stuff.

containscafeine:
@kadel that means hiding the entire service catalog output

pkg/occlient/occlient.go
containscafeine: changed to scv1beta1

containscafeine:
@kadel commented out service catalog for now, and made the changes, PTAL

kadel:
> @kadel that means hiding the entire service catalog output

Yes,  I doesn't make sense to show it right now when we don't have a way to deploy those thinks

cmd/catalog.go
```diff
@@ -0,0 +1,68 @@
...
+			fmt.Println("The following components can be deployed:")
+			for _, component := range catalogList {
+				fmt.Printf("- %v\n", component)
+			}
+		}
+	},
+}
+
+var catalogSearchCmd = &cobra.Command{
+	Use:   "search",
```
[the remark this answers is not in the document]
containscafeine: added

pkg/catalog/catalog.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/redhat-developer/ocdev/pkg/occlient"
+	log "github.com/sirupsen/logrus"
+)
+
+// List lists all the available component types
+// Currently queries the image streams in openshift namespace
```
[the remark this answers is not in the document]
containscafeine: removed

kadel:
It would be nice to clean up Cobra metadata, so we have nice command line output. 
Something like what is here : https://github.com/redhat-developer/ocdev/pull/247/files#diff-1dcc122e377b3a7b4adf52faf26c0188R31

Otherwise this LGTM

containscafeine:
@kadel added more description to the catalog commands, PTAL

cmd/catalog.go
```diff
@@ -0,0 +1,85 @@
...
+
+var catalogListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List all available component types.",
+	Long:  "List all available component types.",
+	Example: `
+# Get the supported components
+ocdev catalog list
+
+# Deploy the component
```
kadel: There should be only examples for  this command

cmd/catalog.go
```diff
@@ -0,0 +1,85 @@
...
+	Short: "Search component type in catalog",
+	Long: `Search component type in catalog.
+
+This searches for a partial match for the given search term in all the available
+components.
+`,
+	Example: `# Search for a component
+ocdev catalog search pyt
+
+# Deploy the component
```
kadel: just examples for this command

cmd/catalog.go
containscafeine: done

cmd/catalog.go
containscafeine: done

PR #248, opened 2018-03-22
Fix of the app list command

mik-dass:
Fix for https://github.com/redhat-developer/ocdev/issues/244


kadel:
thank you @mik-dass !

Issue #65, opened 2018-02-12
implement `ocdev url` to create routes


containscafeine:
This is needed quite a lot right now, since demoing the tool to someone should not require running any oc commands but right now we need to do an `ocdev expose service <service name>`

kadel:
related to https://github.com/redhat-developer/ocdev/issues/172

containscafeine:
Fixed in #231

PR #252, opened 2018-03-22
Add `ocdev url` command

### LINKED ISSUE #231: Introduce `ocdev url` story
Subtask of #172

containscafeine:
This commit adds the `ocdev url create/delete/list` commands which
can be used to add URLs to components.

Underneath, these commands manage OpenShift route objects to
achieve this functionality.

Fix #231 #65

### DIFF
--- a/cmd/url.go
+++ b/cmd/url.go
@@ -1,0 +1,142 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/redhat-developer/ocdev/pkg/application"
+	"github.com/redhat-developer/ocdev/pkg/component"
+	"github.com/redhat-developer/ocdev/pkg/url"
+	"github.com/spf13/cobra"
+)
+
+var (
+	urlListComponent   string
+	urlListApplication string
+)
+
+var urlCmd = &cobra.Command{
+	Use:   "url",
+	Short: "Expose component to the outside world",
+	Long: `Expose component to the outside world.
+The URLs that are generated using this command, can be used to access the
+deployed components from outside the cluster.
+`,
+}
+
+var urlCreateCmd = &cobra.Command{
+	Use:   "create [component name]",
+	Short: "Create a URL for a component",
+	Long: `Create a URL for a component.
+The created URL can be used to access the specified component from outside the
+OpenShift cluster.
+`,
+	Example: `# Create a URL for the current component.
+ocdev url create
+
+# Create a URL for a specific component
+ocdev url create <component name>
+`,
+	Args: cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+
+		var cmp string
+		switch len(args) {
+		case 0:
+			var err error
+			cmp, err = component.GetCurrent(client)
+			if err != nil {
+				fmt.Println(err)
+				os.Exit(1)
+			}
+		case 1:
+			cmp = args[0]
+		default:
+			fmt.Println("unable to get component")
+			os.Exit(1)
+		}
+
+		fmt.Printf("Adding URL to component: %v\n", cmp)
+		u, err := url.Create(client, cmp)
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		fmt.Printf("URL created for component: %v\n\n"+
+			"%v - %v\n", cmp, u.Name, u.URL)
+	},
+}
+
+var urlDeleteCmd = &cobra.Command{
+	Use:   "delete <URL>",
+	Short: "Delete a URL",
+	Long: `Delete a URL.
+Deleted the given URL, hence making the service inaccessible.
+`,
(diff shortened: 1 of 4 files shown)


cmd/url.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var urlCmd = &cobra.Command{
+	Use:   "url",
+	Short: "expose component to the outside world",
+	Long:  "expose component to the outside world",
+}
+
+var urlCreateCmd = &cobra.Command{
+	Use:   "create",
```
kadel: Use should be `create <component_name>`

cmd/url.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		fmt.Printf("URL %v has been created for component %v\n", u, cmp)
+	},
+}
+
+var urlDeleteCmd = &cobra.Command{
+	Use:   "delete",
```
kadel: `delete <component_name>`

pkg/occlient/occlient.go
```diff
@@ -793,3 +802,26 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 
 	return values, nil
 }
+
+func (c *Client) CreateRoute(service string) (*routev1.Route, error) {
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%v-route", service),
```
kadel: why `-route` suffix? Is that really needed?

cmd/url.go
containscafeine: fixed

containscafeine:
@kadel made the changes, also added the list command, PTAL

cmd/url.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+var urlCmd = &cobra.Command{
+	Use:   "url",
+	Short: "expose component to the outside world",
+	Long:  "expose component to the outside world",
+}
+
+var urlCreateCmd = &cobra.Command{
+	Use:   "create <component name, optional>",
```
kadel: if it is optional it should be `[component name]`
`[component name]`  means optional
`<component name>` means required

pkg/occlient/occlient.go
```diff
@@ -793,3 +802,58 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 
 	return values, nil
 }
+
+// CreateRoute creates a route object for the given service and with the given
+// labels
+func (c *Client) CreateRoute(service string, labels map[string]string) (*routev1.Route, error) {
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   fmt.Sprintf(service),
```
kadel: why `Sprintf` here?

cmd/url.go
```diff
@@ -0,0 +1,129 @@
...
+	Short: "create a URL for a component",
+	Long:  "create a URL for a component",
+	Args:  cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+
+		var cmp string
+		switch len(args) {
```
kadel: It is good practice to have a default case in every switch.

kadel:
Can you please add some description to this PR?
At least link to the issue that this is solving would be nice.

cmd/url.go
containscafeine: fixed

pkg/occlient/occlient.go
containscafeine: fixed

cmd/url.go
containscafeine: added

containscafeine:
@kadel made the changes, also added examples and more description to commands, PTAL

pkg/url/url.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// List lists all
+func List(client *occlient.Client, componentName string, applicationName string) ([]URL, error) {
+
+	var labelSelector string
+	if applicationName != "" {
+		labelSelector = labelSelector + fmt.Sprintf("%v=%v", application.ApplicationLabel, applicationName)
+	}
+	if componentName != "" {
```
kadel: what if i call this function witbh  both `componentName` and `applicationName`?
Than just componentName selector will be used :-(

pkg/url/url.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, errors.Wrap(err, "unable to create route")
+	}
+
+	return &URL{
+		Name: route.Name,
+		URL:  route.Spec.Host,
+	}, nil
+}
+
+// List lists the URLs in an application. The results can further be narrowed
```
kadel: 👍 nice comment 👍 
We should make habit of adding comments like this for every function we write.

containscafeine:
Merging since Travis passed

PR #253, opened 2018-03-22
Updated CLI documentation

cdrage:
Adds capitalization and improves ocdev CLI documentation.

Also removed aliased commands (create and list)

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -18,7 +18,7 @@ var (
 // applicationCmd represents the app command
 var applicationCmd = &cobra.Command{
 	Use:     "application",
-	Short:   "application",
+	Short:   "Perform application operations",
 	Aliases: []string{"app"},
 	// 'ocdev application' is the same as 'ocdev application get'
 	Run: applicationGetCmd.Run,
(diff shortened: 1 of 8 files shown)


kadel:
It shows `create` and `list` as aliased commands even for `ocdev component -h`

```
▶ ./ocdev component -h
Components of an application

Usage:
  ocdev component [flags]
  ocdev component [command]

Available Commands:
  delete      Component delete <component_name>
  get         Component get
  set         Set component as active

Aliased Commands:
  create => component create
  list   => component list
```



I was actually thinking that it might be time to remove aliases. 
Or to be more precise move them to original place to aliases.


As it is defined in use cases we should have just following commands regarding components: 
`ocdev create` - create new componnen
`ocdev list` - list existing components
`ocdev component`  - get current component
`ocdev component <name>` - switch component
`ocdev delete <name>` - delete component

cdrage:
@kadel I agree with removing aliases... seems like it adds too many commands which can just be used elsewhere (instead of `ocdev create`, people can simply refer to `ocdev component create`)

kadel:
> @kadel I agree with removing aliases... seems like it adds too many commands which can just be used elsewhere (instead of ocdev create, people can simply refer to ocdev component create)

I actually meant the opposite.  Aliassed comands should remain as only commands.

Instead of `ocdev component create` we should have just `ocdev create`
instead of   `ocdev component  delete` it should be `ocdev delete`
instead of   `ocdev component get ` it should be `ocdev component`
instead of  `ocdev component list ` it should be `ocdev list`
instead of  `ocdev component  set <name>` it should be `ocdev component <name>`

cdrage:
@kadel Okay, I'll work on that and get those commands removed in favour of the smaller commands :+1:

PR #263, opened 2018-03-23
Error out if component already exists

cdrage:
Uses ImageStreams().Get in order to retrieve the image, if it's able to,
it will error out since the component already exists.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -447,6 +447,16 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 			Labels: labels,
 		},
 	}
+
+	// Check to see if the image already exists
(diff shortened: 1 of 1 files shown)


pkg/occlient/occlient.go
```diff
@@ -447,6 +447,16 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 			Labels: labels,
 		},
 	}
+
+	// Check to see if the image already exists
+	info, err := c.imageClient.ImageStreams(c.namespace).Get(is.ObjectMeta.Name, metav1.GetOptions{})
+	log.Debugf("Checking image: %s", info)
+
+	// If err is nill (errors out if the image does not exist)
+	if err == nil {
```
kadel: What if Get failed for other reason than that ImageStream doesn't exist?

kadel:
If this is supposed to check that component doesn't exist before we try to create it than this should b probably done outside of this this. It might be better to use check component.Exists before we call Create

Issue #209, opened 2018-03-08
change the source of my component to be a remote git repository


jorgemoralespou:
As a developer I want to be able to change the source of a component from local (source or binary being pushed) to remote (git) so that it can be built by the system in an unattended way.

```
$ ocd create php [myfrontend]
<Code for a while>
$ ocd push
<Code for a while>
git init
git remote add github.comafjdalkjfa
git push
$ ocd update [myfrontend] --git=<url>
```

Subtask of #162

Issue #210, opened 2018-03-08
change the source of my component to be a local directory


jorgemoralespou:
As a developer I want to be able to change the source of a component from whatever to local directory so I can work locally

```
$ ocd create php [myfrontend] --git=<url>
$ ocd update [myfrontend] --local
```

Subtask of #162

kadel:
realted to @209

PR #273, opened 2018-03-27
Added component update command

mik-dass:
Added component update command as `ocdev update <component_name_optional> --optional_flags flag_value`

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -1,0 +1,114 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/redhat-developer/ocdev/pkg/component"
+	"github.com/spf13/cobra"
+	"os"
+	"path/filepath"
+)
+
+var updateCmd = &cobra.Command{
+	Use: "update",
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) >= 2 {
+			// TODO: Improve this message
+			return fmt.Errorf("Invalid arguments, maximum 1 arguments possible")
+		}
+		return nil
+	},
+	Example: `  # Change the source of a currently active component to local (use the current directory as a source)
+  ocdev update --local
+
+  # Change the source of the frontend component to local with source in ./frontend directory
+  ocdev update frontend --local ./frontend
+
+  # Change the source of a currently active component to git 
+  ocdev update --git https://github.com/openshift/nodejs-ex.git
+
+  # Change the source of the component named node-ex to git
+  ocdev update node-ex --git https://github.com/openshift/nodejs-ex.git
+	`,
+	Short: "Change the source of a component",
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+
+		checkFlag := 0
+
+		if len(componentBinary) != 0 {
+			checkFlag++
+		}
+		if len(componentGit) != 0 {
+			checkFlag++
+		}
+		if len(componentLocal) != 0 {
+			checkFlag++
+		}
+
+		if checkFlag > 1 {
+			fmt.Println("The source can be either --binary or --local or --git")
+			os.Exit(1)
+		}
+
+		if len(componentBinary) != 0 {
+			fmt.Printf("--binary is not implemented yet\n\n")
+			os.Exit(1)
+		}
+
+		var (
+			componentName string
+			err           error
+		)
+		if len(args) == 0 {
+			componentName, err = component.GetCurrent(client)
+			if err != nil {
+				fmt.Println("Unable to get current component")
+			}
+		} else {
+			componentName = args[0]
+		}
+
+		if len(componentGit) != 0 {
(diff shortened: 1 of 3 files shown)


kadel:
For the record: This PR is trying to fix #209 and #210

cmd/root.go
```diff
@@ -31,6 +33,80 @@ var rootCmd = &cobra.Command{
...
+			}
+		} else {
+			componentName = args[0]
+		}
+
+		if len(componentBinary) != 0 {
+			fmt.Printf("--binary is not implemented yet\n\n")
+			os.Exit(1)
+		}
+
```
kadel: there should be one more check to check that only one flag has been set (only `--git` or `--local` or `--binary`)

pkg/occlient/occlient.go
```diff
@@ -631,6 +631,34 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 
 }
 
+func (c *Client) UpdateComponent(componentName string, projectName string, gitUrl string, annotations map[string]string) error {
```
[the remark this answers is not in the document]
kadel: It might be better to call this function something like `UpdateBuildConfig`.  and change `componentName` to `buildConfigName`. OcClient shouldn't have any references to components or applications. It should be purely about openShift objects.

mik-dass:
@kadel Done

cmd/component.go
```diff
@@ -248,6 +248,91 @@ var componentListCmd = &cobra.Command{
 	},
 }
 
+var componentUpdateCmd = &cobra.Command{
```
kadel: this should be in `update.go` and named `updateCmd`

mik-dass: @kadel Done and fixed

cmd/update.go
```diff
@@ -0,0 +1,102 @@
...
+var updateCmd = &cobra.Command{
+	Use: "update",
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) >= 2 {
+			// TODO: Improve this message
+			return fmt.Errorf("Invalid arguments, maximum 1 arguments possible")
+		}
+		return nil
+	},
+	Short: "update the component",
```
kadel: It might be better to replace this with something like this: "Change the source of a component."

kadel: please add Example
something like this:
```
Example: `  # Change the source of a currently active component to local (use the current directory as a source)
  ocd update --local

  #  Change the source of  frontend component to local with source in ./frontend directory
  ocdev update frontend --local ./frontend

  # Change the source of a currently active component to git 
  ocdev update --git https://github.com/openshift/nodejs-ex.git
	`,
```

mik-dass: @kadel Done

cmd/update.go
```diff
@@ -0,0 +1,114 @@
...
+var updateCmd = &cobra.Command{
+	Use: "update",
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) >= 2 {
+			// TODO: Improve this message
+			return fmt.Errorf("Invalid arguments, maximum 1 arguments possible")
+		}
+		return nil
+	},
+	Example: `# Change the source of a currently active component to local (use the current directory as a source)
```
kadel: please add two spaces after '`' 

    Example: ` . # Change...

cmd/update.go
mik-dass: @kadel Done

cmd/update.go
mik-dass: @kadel Done

kadel: spaces should be before `#` 😬

kadel: check `ocdev update -h` output
```
Examples:
#  Change the source of a currently active component to local (use the current directory as a source)
  ocd update --local

  # Change the source of the frontend component to local with source in ./frontend directory
  ocdev update frontend --local ./frontend

  # Change the source of a currently active component to git
  ocdev update --git https://github.com/openshift/nodejs-ex.git

  # Change the source of the component named node-ex to git
  ocdev update node-ex --git https://github.com/openshift/nodejs-ex.git
```

mik-dass: @kadel Done

cmd/update.go
```diff
@@ -0,0 +1,114 @@
...
+	Use: "update",
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) >= 2 {
+			// TODO: Improve this message
+			return fmt.Errorf("Invalid arguments, maximum 1 arguments possible")
+		}
+		return nil
+	},
+	Example: `  # Change the source of a currently active component to local (use the current directory as a source)
+  ocd update --local
```
surajnarwade: s/ocd/ocdev

mik-dass: @surajnarwade Fixed

PR #275, opened 2018-03-27
Update the CLI structure in README.md

containscafeine:


containscafeine:
@kadel how about adding this to the release script?

PR #276, opened 2018-03-27
Add generate-cli-docs to Makefile

containscafeine:
This commit adds the generate-cli-docs target to prepare-release
target. We can add the output of this to the release notes as
well as in the README.md at the time of the release.


kadel:
I think that `scripts/bump-version.sh` might be better placed for this.
Or maybe just mentioning it in "Making release" docs might be enough. 
I actually look at this document every time I do release.

containscafeine:
@kadel makes sense, added it to contributing.md

PR #277, opened 2018-03-27
Improve error handling in cmd/

containscafeine:
The way errors are currently being handled in `package cmd` is
first the error is printed and then `os.Exit(1)` is called.

This is repeated pattern and makes a code unreadable to some
extent.

This commit extracts the error handling logic to a separate
function and calls that function whenever an error has to be
handled. This improved the readability by quite a bit.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/pkg/errors"
 	"github.com/redhat-developer/ocdev/pkg/notify"
 	"github.com/redhat-developer/ocdev/pkg/occlient"
 	log "github.com/sirupsen/logrus"
@@ -38,10 +39,7 @@ func Execute() {
 	updateInfo := make(chan string)
 	go getLatestReleaseInfo(updateInfo)
 
-	if err := rootCmd.Execute(); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
+	checkError(rootCmd.Execute(), "")
 
 	select {
 	case message := <-updateInfo:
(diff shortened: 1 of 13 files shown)


cmd/root.go
```diff
@@ -79,9 +77,13 @@ func getLatestReleaseInfo(info chan<- string) {
 
 func getOcClient() *occlient.Client {
 	client, err := occlient.New()
+	checkError(err, "")
+	return client
+}
+
+func checkError(err error, context string) {
```
kadel: can we get some comments here.
It might be nice to explain for example what context is.

cmd/root.go
```diff
@@ -79,9 +77,13 @@ func getLatestReleaseInfo(info chan<- string) {
...
 	if err != nil {
-		fmt.Println(err)
+		fmt.Println(errors.Wrap(err, context))
```
kadel: Some time back we were discussing that it would be nice to display just the last Error. Instead of displaying the whole chain of wrapped errors.  Could we do it here?
It would be great if we could display a nice error message for the user. And show whole error message chain in the debug log.

kadel:
This is great. It is much nicer than it was 👍

cmd/root.go
containscafeine: done

cmd/root.go
containscafeine: done

cmd/root.go
```diff
@@ -79,9 +77,22 @@ func getLatestReleaseInfo(info chan<- string) {
...
+}
+
+// checkError prints the cause of the given error and exits the code with an
+// exit code of 1.
+// If the context is provided, then that is printed, if not, then the cause is
+// detected using errors.Cause(err)
+func checkError(err error, context string) {
 	if err != nil {
-		fmt.Println(err)
+		if context == "" {
```
kadel: add `log.Debugf("ERROR: %s", err)`and it is perfect

containscafeine: done

cmd/root.go
```diff
@@ -79,9 +77,23 @@ func getLatestReleaseInfo(info chan<- string) {
...
+}
+
+// checkError prints the cause of the given error and exits the code with an
+// exit code of 1.
+// If the context is provided, then that is printed, if not, then the cause is
+// detected using errors.Cause(err)
+func checkError(err error, context string) {
 	if err != nil {
-		fmt.Println(err)
+		log.Debugf("Checking for error:\n%v", err)
```
kadel: I'm not sure if "Checking for error" is accurate here.
This is in if statement `err!=nil` so when it hits this line we already know that there has been error.

containscafeine: changed

kadel:
LGTM,  it I would like to merge https://github.com/redhat-developer/ocdev/pull/258 first, as there are potential conflicts and it will be easier to fix it here than in https://github.com/redhat-developer/ocdev/pull/258

containscafeine:
rebased, made the changes

surajnarwade:
@containscafeine , run `goimports` otherwise LGTM :)

kadel:
one would assume that one at least runs `make` before pushing PR 😁

PR #279, opened 2018-03-27
Remove all aliases to get/list/create/set/delete commands

cdrage:
This PR removes the `ocdev component X` commands for the following
commands:

```sh
Available Commands:
  application Perform application operations
  catalog     Catalog related operations
  completion  Output shell completion code
  create      Create new component
  delete      Delete existing component
  get         Get currently active component
  help        Help about any command
  list        List all components in the current application
  project     Perform project operations
  push        Push source code to component
  set         Set active component.
  storage     Perform storage operations
  url         Expose component to the outside world
  version     Print the version of ocdev
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -1,6 +1,120 @@
 package cmd
 
-// 'ocdev crate' is just an alias for 'ocdev component create'
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/redhat-developer/ocdev/pkg/component"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	componentBinary string
+	componentGit    string
+	componentLocal  string
+)
+
+var componentCreateCmd = &cobra.Command{
+	Use:   "create <component_type> [component_name] [flags]",
+	Short: "Create new component",
+	Long: `Create new component.
+If component name is not provided, component type value will be used for name.
+	`,
+	Example: `  # Create new nodejs component with the source in current directory. 
+  ocdev create nodejs
+
+  # Create new nodejs component named 'frontend' with the source in './frontend' directory
+  ocdev create nodejs frontend --local ./frontend
+
+  # Create new nodejs component with source from remote git repository.
+  ocdev create nodejs --git https://github.com/openshift/nodejs-ex.git
+	`,
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			// TODO: improve this message. It should say something about requiring component name
+			return fmt.Errorf("At least one argument is required")
+		}
+		if len(args) > 2 {
+			// TODO: Improve this message
+			return fmt.Errorf("Invalid arguments, maximum 2 arguments possible")
+		}
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		log.Debugf("Component create called with args: %#v, flags: binary=%s, git=%s, local=%s", strings.Join(args, " "), componentBinary, componentGit, componentLocal)
+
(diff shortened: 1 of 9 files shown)


cmd/set.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/redhat-developer/ocdev/pkg/component"
+	"github.com/spf13/cobra"
+)
+
+var componentSetCmd = &cobra.Command{
+	Use:   "set",
```
kadel: This command should be called `component` and should be coupled together with `set`

```
# getting current component
ocdev component

# switching current component 
ocdev component <name>
```

if you look `oc project` in works in the same way

kadel: There will be no `set` command. 
`ocdev component <name>` will act as set. If there is no argument, it will just print current application.
Look at `oc project`, it works in the same way.

cdrage: @kadel Ah, that makes sense. Was a bit confused. I'll update the PR.

kadel: no problem, thank ;-)

cdrage:
Hopefully tests pass, but I updated this :+1: I wasn't able to get the e2e tests running locally, so just waiting on TravisCI.

cmd/create.go
```diff
@@ -1,6 +1,121 @@
...
+		// after component is successfully created, set is as active
+		if err := component.SetCurrent(client, componentName); err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+	},
+}
+
 // 'ocdev crate' is just an alias for 'ocdev component create'
```
kadel: we can remove this commend

cmd/delete.go
```diff
@@ -0,0 +1,69 @@
...
+				os.Exit(1)
+			}
+			fmt.Println(output)
+		} else {
+			fmt.Printf("Aborting deletion of component: %v\n", componentName)
+		}
+	},
+}
+
+// 'ocdev list' is just an alias for 'ocdev component list'
```
kadel: remove comment

cmd/create.go
kadel: remove comment

cmd/list.go
```diff
@@ -1,5 +1,40 @@
...
+
+		fmt.Println("You have deployed:")
+		for _, comp := range components {
+			fmt.Printf("%s using the %s component\n", comp.Name, comp.Type)
+		}
+
+	},
+}
+
 // 'ocdev list' is just an alias for 'ocdev component list'
```
kadel: remove comment

cdrage:
@kadel 

Re-based, removed the uneeded comments, and also added the default `ocdev component foobar` set command.

```sh
github.com/redhat-developer/ocdev  remove-aliases ✔                                                                                                                                                                                                                         0m  ⍉
▶ ./ocdev component
The current component is: foobar

github.com/redhat-developer/ocdev  remove-aliases ✔                                                                                                                                                                                                                          0m  
▶ ./ocdev component get
The current component is: foobar

github.com/redhat-developer/ocdev  remove-aliases ✔                                                                                                                                                                                                                          0m  
▶ ./ocdev component set foo
Switched to component: foo

github.com/redhat-developer/ocdev  remove-aliases ✔                                                                                                                                                                                                                          0m  
▶ ./ocdev component foobar
Switched to component: foobar

github.com/redhat-developer/ocdev  remove-aliases ✔                                                                                                                                                                                                                          0m  
▶ ./ocdev component
The current component is: foobar
```

Code for the if statement for the set / get command is here: https://github.com/redhat-developer/ocdev/pull/279/files#diff-1dcc122e377b3a7b4adf52faf26c0188R19

PR #282, opened 2018-03-28
Allow switch to app that is not in config but exists in cluster.

kadel:
There might be a situation where application is not defined in local config but it is present in OpenShift cluster.
This situation can happen for example if user delted config file.
In this case we need to add application back to the the config  before we set is as active.


cdrage:
LGTM

PR #286, opened 2018-03-29
Print error trace when tests fail

ashetty1:


tests/e2e/e2e_test.go
```diff
@@ -36,9 +36,11 @@ func runCmd(cmdS string) string {
 	err := cmd.Run()
 
 	if err != nil {
-		log.Fatal(err)
+		fmt.Println(fmt.Sprint(err) + ": " + stderr.String())
+		os.Exit(1)
```
kadel: wouldn't this prevent executing following tests?
If one test fails we shouldn't stop testing, but rather skip to the next one.

ashetty1: @kadel currently we have most tests in one block. Most tests would need the previous operation to succeed.

PR #290, opened 2018-03-29
Updated documentation for minro spell/grammer check

surajnarwade:


containscafeine:
@surajnarwade did you use any library for this? It would be cool if we could include this in our CI runs as a warning, besides Travis.

PR #292, opened 2018-03-29
Remove unnecessary return from both GetLabels()

containscafeine:
This commit removes the extra return value of type error from the
GetLabels() function in package application and package component.

There was no case under which these functions could have failed,
and having the error as a return was not required.


kadel:
You also have to modify places where these functions were called and remove error checking.

containscafeine:
@kadel fixed

PR #294, opened 2018-03-29
Set current/default application at the start of component create

### LINKED ISSUE #291: Orphaned components should not be created
State - fresh installation of ocdev, no applications, components, etc exist
Step 1 - `ocdev component create nodejs`
Step 2 - `ocdev component list`

This does not return any components.

Basically the logic is broken because the component creation should not even begin binding itself to the application. We need to get the `GetCurrentOrDefault` logic from inside `component.Create()` and move it outside. `component.Create()` should be passed the application name as a parameter.

surajnarwade:
Resolves #291


cmd/component.go
```diff
@@ -212,7 +217,12 @@ var componentSetCmd = &cobra.Command{
 		return nil
 	}, Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		err := component.SetCurrent(client, args[0])
+		currentApplication, err := application.GetCurrentOrDefault(client)
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		err = component.SetCurrent(client, args[0], currentApplication)
```
cdrage: We will probably need to rebase this after #279 is merged in, but since these are two functions which are copied / pasted. Should we put this under a function instead?

Perhaps a `getAndSetApplication()` function?

kadel: > Perhaps a getAndSetApplication() function?

It is getting a current application but setting current component. 
It is better to decouple this logic.

pkg/component/component.go
```diff
@@ -40,14 +40,7 @@ func GetLabels(componentName string, applicationName string, additional bool) ma
...
 
-func CreateFromGit(client *occlient.Client, name string, ctype string, url string) (string, error) {
-	// if current application doesn't exist, create it
-	// this can happen when ocdev is started form clean state
-	// and default application is used (first command run is ocdev create)
-	currentApplication, err := application.GetCurrentOrDefault(client)
-	if err != nil {
-		return "", errors.Wrapf(err, "unable to create git component %s", name)
-	}
+func CreateFromGit(client *occlient.Client, currentApplication string, name string, ctype string, url string) (string, error) {
```
kadel: the argument shouldn't be `currentApplication` but just `applicationName`. 
As in the new context, it doesn't have to be a current application. 
(name can't be just `application` as we already have application package)

pkg/component/component.go
```diff
@@ -74,12 +67,7 @@ func CreateFromGit(client *occlient.Client, name string, ctype string, url strin
 }
 
 // CreateFromDir create new component with source from local directory
-func CreateFromDir(client *occlient.Client, name string, ctype string, dir string) (string, error) {
-	currentApplication, err := application.GetCurrentOrDefault(client)
-	if err != nil {
-		return "", errors.Wrapf(err, "unable to create component %s from local path", name, dir)
-	}
-
+func CreateFromDir(client *occlient.Client, currentApplication string, name string, ctype string, dir string) (string, error) {
```
kadel: again, not `currentApplication` but just `applicationName`

pkg/component/component.go
```diff
@@ -145,19 +133,14 @@ func Delete(client *occlient.Client, name string) (string, error) {
 	return output, nil
 }
 
-func SetCurrent(client *occlient.Client, name string) error {
+func SetCurrent(client *occlient.Client, name string, currentApplication string) error {
```
kadel: s/currentApplication/applicationName

surajnarwade:
@kadel @containscafeine can you have please review this ?

cmd/component.go
```diff
@@ -53,8 +56,19 @@ var componentSetCmd = &cobra.Command{
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		err := component.SetCurrent(client, args[0])
+
+		currentApplication, err := application.GetCurrentOrDefault(client)
 		checkError(err, "")
+		err = component.SetCurrent(client, args[0], currentApplication)
+		if err != nil {
```
containscafeine: replace with `checkError()` #277

containscafeine:
Blocked on #330

pkg/application/application.go
```diff
@@ -197,7 +197,7 @@ func SetCurrent(client *occlient.Client, name string) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to set current application")
 	}
-	if !exists {
+	if !exists && name != "app" {
```
kadel: This check shouldn't be there

kadel:
Isn't this PR the same as https://github.com/redhat-developer/odo/pull/358/files ?
@containscafeine @surajnarwade ?

PR #295, opened 2018-03-29
Move storage to client-go

containscafeine:
This commit refactors the current implementation of `ocdev
storage` command to client-go.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -41,11 +41,17 @@ import (
 	s2ifs "github.com/openshift/source-to-image/pkg/util/fs"
 
 	dockerapiv10 "github.com/openshift/api/image/docker10"
+	"github.com/redhat-developer/ocdev/pkg/util"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
 	ocRequestTimeout   = 1 * time.Second
 	OpenShiftNameSpace = "openshift"
+
+	// The length of the string to be generated for names of resources
+	nameLength = 5
 )
 
 type Client struct {
@@ -817,53 +823,6 @@ func (c *Client) DeleteProject(name string) error {
 	return nil
 }
 
-type VolumeConfig struct {
-	Name             *string
-	Size             *string
-	DeploymentConfig *string
-	Path             *string
-}
-
-type VolumeOpertaions struct {
-	Add    bool
-	Remove bool
-	List   bool
-}
-
-func (c *Client) SetVolumes(config *VolumeConfig, operations *VolumeOpertaions) (string, error) {
-	args := []string{
-		"set",
-		"volumes",
-		"dc", *config.DeploymentConfig,
-		"--type", "pvc",
-	}
-	if config.Name != nil {
-		args = append(args, "--name", *config.Name)
-	}
-	if config.Size != nil {
-		args = append(args, "--claim-size", *config.Size)
-	}
-	if config.Path != nil {
-		args = append(args, "--mount-path", *config.Path)
-	}
-	if operations.Add {
-		args = append(args, "--add")
-	}
-	if operations.Remove {
-		args = append(args, "--remove", "--confirm")
-	}
-	if operations.List {
-		args = append(args, "--all")
-	}
-	output, err := c.runOcComamnd(&OcCommand{
-		args: args,
-	})
-	if err != nil {
-		return "", errors.Wrap(err, "unable to perform volume operations")
-	}
-	return string(output), nil
-}
-
 // GetLabelValues get label values of given label from objects in project that are matching selector
 // returns slice of uniq label values
 func (c *Client) GetLabelValues(project string, label string, selector string) ([]string, error) {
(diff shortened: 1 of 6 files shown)


pkg/util/util.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import "fmt"
+
+func ConvertLabelsToSelector(labels map[string]string) string {
```
kadel: Comment, please ;-)

pkg/occlient/occlient.go
```diff
@@ -51,6 +53,7 @@ const (
 type Client struct {
 	ocpath               string
 	kubeClient           *kubernetes.Clientset
+	coreClient           *coreclient.CoreV1Client
```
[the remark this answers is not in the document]
containscafeine: removed

pkg/occlient/occlient.go
```diff
@@ -982,3 +991,106 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+					corev1.ResourceStorage: quantity,
+				},
+			},
+			AccessModes: []corev1.PersistentVolumeAccessMode{
+				corev1.ReadWriteOnce,
+			},
+		},
+	}
+
+	createdPvc, err := c.coreClient.PersistentVolumeClaims(c.namespace).Create(pvc)
```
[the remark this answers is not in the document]
containscafeine: changed

pkg/occlient/occlient.go
```diff
@@ -33,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
+	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
```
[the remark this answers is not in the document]
containscafeine: removed

pkg/occlient/occlient.go
```diff
@@ -116,6 +119,12 @@ func New() (*Client, error) {
 	}
 	client.routeClient = routeClient
 
+	coreClient, err := coreclient.NewForConfig(config)
```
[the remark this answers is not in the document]
containscafeine: removed

pkg/occlient/occlient.go
```diff
@@ -982,3 +991,106 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+	log.Debugf("Getting DeploymentConfig: %v", dc)
+	deploymentConfig, err := c.appsClient.DeploymentConfigs(c.namespace).Get(dc, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "unable to get DeploymentConfig: %v", dc)
+	}
+
+	log.Debug("Updating DeploymentConfig Object")
+	volumeName := fmt.Sprintf("%v-volume", dc)
+
+	deploymentConfig.Spec.Template.Spec.Volumes = []corev1.Volume{
```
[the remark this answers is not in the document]
containscafeine: yep, I changed it to append later on

pkg/occlient/occlient.go
```diff
@@ -982,3 +991,106 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+			Name: volumeName,
+			VolumeSource: corev1.VolumeSource{
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: pvc,
+				},
+			},
+		},
+	}
+
+	deploymentConfig.Spec.Template.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{
```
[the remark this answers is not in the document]
containscafeine: yep, I changed it to append later on

containscafeine:
@kadel made the changes, and added comments. PTAL.

pkg/occlient/occlient.go
```diff
@@ -1015,3 +973,189 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+// generateVolumeNameFromDC generates a random volume name based on the name
+// of the given Deployment Config
+func generateVolumeNameFromDC(dc string) string {
+	return fmt.Sprintf("%v-%v-volume", dc, util.GenerateRandomString(nameLength))
+}
+
+// AddPVCToDeploymentConfig adds the given PVC to the given Deployment Config
+// at the given path
+func (c *Client) AddPVCToDeploymentConfig(dc *appsv1.DeploymentConfig, pvc string, path string) error {
+	volumeName := generateVolumeNameFromDC(dc.Name)
```
kadel: Why does volume name have random value in it? Volume is inside DC so it can only collide with other volumes for that DC, but not with anything else.
Wouldn't it be better to name it after PVC?

pkg/occlient/occlient.go
```diff
@@ -1015,3 +973,189 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+	_, err := c.appsClient.DeploymentConfigs(c.namespace).Update(dc)
+	if err != nil {
+		return errors.Wrapf(err, "failed to update DeploymentConfig: %v", dc)
+	}
+	return nil
+}
+
+// removeVolumeFromDC removes the volume from the given Deployment Config and
+// returns true. If the given volume is not found, it returns false.
+func removeVolumeFromDC(vol string, dc *appsv1.DeploymentConfig) bool {
```
kadel: we can easily write unit tests for this function

pkg/occlient/occlient.go
```diff
@@ -1015,3 +973,189 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+			dc.Spec.Template.Spec.Volumes = append(dc.Spec.Template.Spec.Volumes[:i], dc.Spec.Template.Spec.Volumes[i+1:]...)
+		}
+	}
+	return found
+}
+
+// removeVolumeMountFromDC removes the volumeMount from all the given containers
+// in the given Deployment Config and return true. If the given volumeMount is
+// not found, it returns false
+func removeVolumeMountFromDC(vm string, dc *appsv1.DeploymentConfig) bool {
```
kadel: we can easily write unit tests for this function

pkg/occlient/occlient.go
```diff
@@ -1015,3 +973,189 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to update the Deployment Config: %v", dc.Name)
+	}
+
+	return nil
+}
+
+// getVolumeFromPVC returns the name of the volume associated with the given
+// PVC in the given Deployment Config
+func (c *Client) getVolumeFromPVC(pvc string, dc *appsv1.DeploymentConfig) (string, error) {
```
kadel: we can easily write unit tests for this function

pkg/occlient/occlient.go
```diff
@@ -1015,3 +973,189 @@ func (c *Client) ListRouteNames(labelSelector string) ([]string, error) {
...
+	for _, volume := range dc.Spec.Template.Spec.Volumes {
```
kadel: There might be multiple volumes that are using the same PVC. Are we thinking about this?
There should be at least comment warning that this situation is theoretically possible.

pkg/storage/storage.go
```diff
@@ -1,39 +1,150 @@
...
+const StorageLabel = "app.kubernetes.io/storage-name"
+
+type StorageInfo struct {
+	Name string
+	Size string
+	Path string
+}
+
+// Add adds storage to given component of given application
+func Add(client *occlient.Client, name string, size string, path string, applicationName string, componentName string) (string, error) {
```
kadel: in all other function, we have opposite order for `applicationName` and `componentName`. `componentName` is before `applicationName` it would be good to keep it consistent.

pkg/storage/storage.go
```diff
@@ -1,39 +1,150 @@
...
+
+	labels := GetLabels(name, componentName, applicationName, false)
```
kadel: when you are creating object, the last parameter for `GetLabels`  should be `true`

kadel:
> @kadel made the changes, and added comments. PTAL.

A few smaller cancers, otherwise this looks good. Great job @containscafeine

pkg/occlient/occlient.go
containscafeine: let's return all the volume names, and error out if more than one was returned

pkg/occlient/occlient.go
containscafeine: makes sense, changed this to derive from PVC

pkg/occlient/occlient.go
containscafeine: made the changes

pkg/storage/storage.go
containscafeine: fixed

pkg/storage/storage.go
containscafeine: fixed

containscafeine:
@kadel made the changes, PTAL. Will cover unit tests in #334

kadel:
found two issues (#338  and #337) when testing this, not sure if they were introduced when migrating to client-go or there were here before.
I don't think those are blockers for this PR. But they should be definitely addressed in the following PR

surajnarwade:
@kadel @containscafeine LGTM, works for me

PR #301, opened 2018-04-04
Fixed component git build log issue

mik-dass:
Issue : https://github.com/redhat-developer/ocdev/issues/264

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -40,22 +40,22 @@ func GetLabels(componentName string, applicationName string, additional bool) ma
 	return labels
 }
 
-func CreateFromGit(client *occlient.Client, name string, ctype string, url string) (string, error) {
+func CreateFromGit(client *occlient.Client, name string, ctype string, url string) error {
 	// if current application doesn't exist, create it
 	// this can happen when ocdev is started form clean state
 	// and default application is used (first command run is ocdev create)
 	currentApplication, err := application.GetCurrentOrDefault(client)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to create git component %s", name)
+		return errors.Wrapf(err, "unable to create git component %s", name)
 	}
 	exists, err := application.Exists(client, currentApplication)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to create git component %s", name)
+		return errors.Wrapf(err, "unable to create git component %s", name)
 	}
 	if !exists {
 		err = application.Create(client, currentApplication)
 		if err != nil {
-			return "", errors.Wrapf(err, "unable to create git component %s", name)
+			return errors.Wrapf(err, "unable to create git component %s", name)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/component/component.go
```diff
@@ -66,28 +66,38 @@ func CreateFromGit(client *occlient.Client, name string, ctype string, url strin
...
+		return errors.Wrapf(err, "unable to create git component %s", name)
 	}
-	return output, nil
+
+	//get the latest build name for following
+	buildName, err := client.GetLatestBuildName(name)
+	if err != nil {
+		errors.Wrap(err, "unable to create from git")
+	}
+	err = client.FollowBuildLog(buildName)
```
kadel: please add the following message before showing logs, so  it is consistent with output when building from a local directory
```
	fmt.Println("please wait, building application...")
```

pkg/component/component.go
```diff
@@ -66,28 +66,38 @@ func CreateFromGit(client *occlient.Client, name string, ctype string, url strin
...
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to create git component %s", name)
+		return errors.Wrapf(err, "unable to create git component %s", name)
 	}
-	return output, nil
+
+	//get the latest build name for following
+	buildName, err := client.GetLatestBuildName(name)
+	if err != nil {
+		errors.Wrap(err, "unable to create from git")
```
containscafeine: you missed to `return` the error here

pkg/component/component.go
containscafeine: I think this should be a part of `NewAppS2I()` instead of being called after that. This should be called right after the BuildConfig has been created in `NewAppS2I()`. We can also add a `stream bool` input parameter to `NewAppS2I()` which will set if the logs should be streamed or not.
Another way could be to break `NewAppS2I()` into smaller functions, one to create DC, another for BC, which can be reused at other places.

Any way, this should be inside `NewAppS2I()`

kadel: I don't think this should go to `NewAppS2I`.
It will just add more logic to that function also because of this difference a between builds from git and from dir.

NewAppS2I should just create objects nothing more

mik-dass: @kadel shouldn't it be building component?

pkg/occlient/occlient.go
```diff
@@ -444,11 +444,11 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
 
 // NewAppS2I create new application using S2I
 // if gitUrl is ""  than it creates binary build otherwise uses gitUrl as buildSource
-func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string) (string, error) {
+func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string, stream bool) error {
```
kadel: If you really want to have the `stream` flag then please add a comment what the stream argument is.

pkg/component/component.go
kadel: It is your code @mik-dass, I'm leaving it up to you how you want to do it.

PR #308, opened 2018-04-04
Add a bit of verbosity to e2e tests

cdrage:
Adds verbosity to the tests as well as command-line-outout:

```sh
▶ make test-e2e
go test github.com/redhat-developer/ocdev/tests/e2e -ginkgo.v
$ git clone https://github.com/openshift/nodejs-ex /tmp/ocdev807324634/nodejs-ex

Running Suite: ocdev test suite
===============================
Random Seed: 1522871416
Will run 9 of 9 specs

Usecase #5 ocdev project
  should create a new project
  /home/wikus/seafile/files/dev/go/src/github.com/redhat-developer/ocdev/tests/e2e/e2e_test.go:120
$ ocdev project create ocdev-1522871416
New project created and now using project : ocdev-1522871416

$ ocdev project get --short
ocdev-1522871416

•
Usecase #5 creating an application when application by the same name doesn't exist
  should create an application
  /home/wikus/seafile/files/dev/go/src/github.com/redhat-developer/ocdev/tests/e2e/e2e_test.go:129
$ ocdev application create usecase5
Creating application: usecase5
Switched to application: usecase5

$ ocdev application get --short
usecase5

•
Usecase #5 creating an application when application by the same name doesn't exist
  should be created within the project
  /home/wikus/seafile/files/dev/go/src/github.com/redhat-developer/ocdev/tests/e2e/e2e_test.go:134
$ ocdev project get --short
ocdev-1522871416

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ test:
 # Run e2e tests
 .PHONY: test-e2e
 test-e2e:
-	go test github.com/redhat-developer/ocdev/tests/e2e
+	go test github.com/redhat-developer/ocdev/tests/e2e -ginkgo.v
 
 # create deb and rpm packages using fpm in ./dist/pkgs/
 # run make cross before this!
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -40,7 +40,7 @@ func runCmd(cmdS string) string {
 		os.Exit(1)
 	}
 
-	fmt.Println(string(out.Bytes()))
+	fmt.Printf("$ %s\n%s\n\n", cmdS, string(out.Bytes()))
 	return string(out.Bytes())
 }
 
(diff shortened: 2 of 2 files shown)


kadel:
To be honest I don't like when tests are showing verbose output for everything.
We should show only verbose output for failed tests. There is no point of showing detailed output for a test that passed, it just makes it hard to find failures.

cdrage:
@kadel The problem is that with Ginko it won't even show a description of the failed test :-1:

kadel:
oh :-( I'm starting to feel like ginko is giving us just problems, instead of making things easier :-(

cdrage:
@kadel We can put a bit of verbosity (but not too much) within: https://github.com/redhat-developer/ocdev/blob/master/tests/e2e/e2e_test.go#L28 which may help with regards to descriptions, etc.

cdrage:
The problem too is that we're using generic names such as "Usecase #5"... which makes no sense, we should use the `Describe()` function more: http://onsi.github.io/ginkgo/#organizing-specs-with-containers-describe-and-context

kadel:
OK, merging this, so at least we get a little bit of context when something fails.

But lets remove verbose flag when https://github.com/redhat-developer/ocdev/issues/316 is fixed

PR #309, opened 2018-04-05
Tests for storage commands

ashetty1:
This commit adds tests for `ocdev storage` operations. Also, the `runCmd` function which runs the ocdev commands will now accept an additional argument of exit status. The exit status value is 0 by default.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -26,6 +26,7 @@ var t = strconv.FormatInt(time.Now().Unix(), 10)
 var projName = fmt.Sprintf("ocdev-%s", t)
 
 func runCmd(cmdS string) string {
+
 	var cmd *exec.Cmd
 	var out, stderr bytes.Buffer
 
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -25,7 +25,8 @@ import (
 var t = strconv.FormatInt(time.Now().Unix(), 10)
 var projName = fmt.Sprintf("ocdev-%s", t)
 
-func runCmd(cmdS string) string {
+func runCmd(cmdS string, exitStatus ...int) string {
```
kadel: what is `existStatus` parameter?
Can we have please have some comments?

tests/e2e/e2e_test.go
```diff
@@ -37,7 +38,9 @@ func runCmd(cmdS string) string {
 
 	if err != nil {
 		fmt.Println(fmt.Sprint(err) + ": " + stderr.String())
-		os.Exit(1)
+		if len(exitStatus) > 0 {
```
kadel: why are we checking the length of `exitStatus`?

ashetty1: I thought making `runCmd` variadic would be good way to deal with exit when there are errors. For most operations, we wouldn't want failing operations exiting test run. With this current, I can specify a non-zero exit code in case I case the test run to error out when a certain operation fails:

`runCmd(<cmd_to_be_run>, 1)` would fail the test run with os.Exit(1) in case the operation fails. If exit code isn't specified, i.e.  `runCmd(<cmd_to_be_run>`), even if the command fails, it would move on to the next test/operation. 

Let me know if this makes sense.

ashetty1: @kadel I have reverted the `runCmd` changes. Will address that issue in another PR with related changes.

PR #317, opened 2018-04-05
Refactor cobra / use cobra.Args arguments

cdrage:
Uses the cobra.Args arguments for setting maximum / range / minimum
arguments required for some commands rather than using a custom checker.

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -27,15 +27,7 @@ var applicationCmd = &cobra.Command{
 var applicationCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: "create an application",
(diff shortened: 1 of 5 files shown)


kadel:
Problem with standard checks is that you can't display a custom message that will guide user what he did wrong, or what arguments he should provide :-(

cdrage:
@kadel Problem is that our current output display messages aren't much different than what cobra displays. Unless we improve the UX a bit. I'll probably open up another issue to improve some of the descriptions (specifically, `ocdev create`).

kadel:
OK, you are right, I looked closer on the messages, and they aren't that much help either :-(

PR #322, opened 2018-04-06
Fixing error-handling and logging of errors in e2e tests

ashetty1:
With this commit, the `runCmd` function has been modified to use GinkgoWriter and gexec.Start.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -5,8 +5,8 @@ package e2e
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	gexec "github.com/onsi/gomega/gexec"
 
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


kadel:
ah, sorry, my fault, shouldn't merge my quick fix PR. I didn't realized that you have already opened this :-(

tests/e2e/e2e_test.go
```diff
@@ -115,191 +92,273 @@ func TestOCdev(t *testing.T) {
...
+				projName, err := runCmd("ocdev project get --short")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(projName).To(ContainSubstring(projName))
 			})
 
-			It("should be able to set an application as current", func() {
-				createApp("usecase5-2")
-				Expect(getApp()).To(Equal("usecase5-2"))
+			It("should be able to create another application", func() {
+				appName, _ := runCmd("ocdev application create usecase5-2")
```
kadel: why is error ignored here?

tests/e2e/e2e_test.go
```diff
@@ -25,51 +25,28 @@ import (
...
-func runCmd(cmdS string) string {
+func runCmd(cmdS string) (string, error) {
+	cmd := exec.Command("/bin/sh", "-c", cmdS)
+	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
 
-	var cmd *exec.Cmd
-	var out, stderr bytes.Buffer
+	// wait for the command execution to complete
+	<-session.Exited
+	Expect(session.ExitCode()).To(Equal(0))
```
kadel: would it make sense to also check `err` (`Expect(err).NotTo(HaveOccurred`) here?
So we don't have to think about checking it after every `runCmd` call.

tests/e2e/e2e_test.go
```diff
@@ -25,51 +25,28 @@ import (
...
+// Function get commands as oc admin
+func runOcAdmin(cmd string) (string, error) {
+	_, err1 := runCmd("oc login -u system:admin > /dev/null")
+	Expect(err1).NotTo(HaveOccurred())
 
-func getProj() string {
-	proj := runCmd("ocdev project get --short")
-	return strings.TrimSpace(proj)
-}
+	cmdOut, err2 := runCmd(cmd)
```
kadel: would it make sense to also check err2 (`Expect(err2).NotTo(HaveOccurred`) here?
So we don't have to think about checking it after every runOcAdmin call.

tests/e2e/e2e_test.go
kadel: If we check err here, then we don't have to even return it. Or am I missing something?

kadel: do we still need that even if `Expect` 'failed'?

ashetty1: @kadel let's leave out `Fail` for now. It is not so crucial as I thought it would be. Let's come back to it when we see the need.

ashetty1:
@kadel rebased and made the changes. Let me know if this looks good.

Issue #207, opened 2018-03-08
sync my local directory sources to an existing component


jorgemoralespou:
I want to be able to code locally and have it sync my changes to my running component and be smart enough to let me know when a restart of the container is required if dependencies have changed. This should work across PHP, Node.JS, Python, Ruby, and Java.

```
$ cd myfrontend
$ ocd create php:5.4
$ ocd push
$ ocd watch [.]  //blocks terminal and watches for file changes

Output:
Watching for changes.  To stop, use control-c
[11:34:12] myfile.php was pushed
[12:04:05] logo.png was pushed
```

Subtask of #160 

There's work in progress to demonstrate how this could work done by @grahamdumpleton and @jorgemoralespou

kadel:
Just to make sure that I understand that correctly.

The build process should be slightly different than regular s2i using BuildConfig.
Application container will be constantly running. 
`ocdev watch` will be continuously watching local source directory and when there is a change in the source it will sync it to the running container. Whan this happens something should trigger reload or restart (but it won't restart container, only the process running given program).


This will mean that we will have to use a special image for this as it will require runtime and also build dependencies.
Before this application will be deployed to production it will require another "clean" build (without build dependencies). This will complicate propagation to production process  :-(

Would it be better to use normal s2i with BuildConfig? If we have good s2i images with support for incremental builds whole rebuild process will be quick. From the beginning, the application will be running in a clean image that will be also used for production deployment.

slemeur:
I would move forward with the normal s2i, which would allow to deliver the user story in a short time.

Once we'll have an initial release we will have time to refine this and use different components.

jorgemoralespou:
@kadel @slemeur yes, we should move forward as this really would require some changes/tweaks to the DeploymentConfig. Ideally (when solved) we will modify the DeploymentConfig to have a supervisor process, that will be injected via an Init container, that will take care of properly reloading the server when appropriate. As mentioned in the initial issue, there's a prototype working, and we would probably sync. The one that's done most of the work and has most of the knowledge from our side is @GrahamDumpleton. He's in Sydney TZ, so we can sync early in the morning any day.

I would defer this for now, or create a follow up task.

jorgemoralespou:
We should not consider watch for scaled workloads of more than 1 pod

kadel:
[...]

PR #325, opened 2018-04-06
Add watch command

kadel:
This is similar to push. But watch blocks terminal and watches
components source directory for changes.

If something changes `oc rsync` is triggered and changes are pushed to the directory where S2I stores applications (`/opt/app-root/src`)

Right now this makes sense only for dynamic languages that don't need reload/restart (like PHP).

related to #207


cmd/watch.go
```diff
@@ -0,0 +1,85 @@
+package cmd
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/redhat-developer/ocdev/pkg/application"
```
surajnarwade: change references from `ocdev` to `odo`

kadel: thx, fixed

surajnarwade:
I tried this PR & worked for me, LGTM

pkg/component/watch.go
```diff
@@ -0,0 +1,166 @@
...
+				changeLock.Unlock()
+			}
+		}
+	}()
+	err = addRecursiveWatch(watcher, dir, ignores)
+	if err != nil {
+		return fmt.Errorf("error watching source path %s: %v", dir, err)
+	}
+
+	delay := 5 * time.Second
```
cdrage: should be a global variable

cmd/watch.go
```diff
@@ -0,0 +1,85 @@
...
+
+		err = component.WatchAndPush(client, componentName, watchPath, stdout)
+		if err != nil {
+			checkError(err, "Error while trying to watch %s", watchPath)
+		}
+	},
+}
+
+func init() {
+	watchCmd.Flags().StringVar(&watchLocal, "local", "", "Use given local directory as a source for component. (It must be a local component)")
```
cdrage: @kadel Should we add another parameter for setting the delay?

pkg/component/watch.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/redhat-developer/odo/pkg/occlient"
+
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
+)
+
+// addRecursiveWatch handles adding watches recursively for the path provided
+// and its subdirectories.  If a non-directory is specified, this call is a no-op.
+// Files matching glob patter defined in ignores will be ignored.
+// Recursive logic from https://github.com/bronze1man/kmg/blob/master/fsnotify/Watcher.go
```
cdrage: Link doesn't work: https://github.com/bronze1man/kmg/blob/master/fsnotify/Watcher.go

pkg/component/watch.go
kadel: why global if it is not used anywhere else?

cmd/watch.go
kadel: I don't want to complicate it right from the beginning. If there is a need for it we can add it later.

pkg/component/watch.go
kadel: updated

cmd/root.go
```diff
@@ -85,13 +85,13 @@ func getOcClient() *occlient.Client {
...
 // detected using errors.Cause(err)
-func checkError(err error, context string) {
+func checkError(err error, context string, a ...interface{}) {
 	if err != nil {
 		log.Debugf("Error:\n%v", err)
 		if context == "" {
 			fmt.Println(errors.Cause(err))
 		} else {
-			fmt.Println(context)
+			fmt.Printf(context, a...)
```
containscafeine: add a newline in the end?

cmd/watch.go
```diff
@@ -0,0 +1,85 @@
...
+
+var watchCmd = &cobra.Command{
+	Use:   "watch [component name]",
+	Short: "Watch for changes, update component on change",
+	Long:  `Watch for changes, update component on change.`,
+	Example: `  # Watch for changes in directory for current component
+  ocdev watch
+
+  # Watch for changes in directory for component called frontend 
+  ocdev watch frontend
```
containscafeine: change ocdev to odo

cmd/root.go
```diff
@@ -85,13 +85,13 @@ func getOcClient() *occlient.Client {
 // exit code of 1.
 // If the context is provided, then that is printed, if not, then the cause is
 // detected using errors.Cause(err)
-func checkError(err error, context string) {
+func checkError(err error, context string, a ...interface{}) {
```
containscafeine: why are we adding `a ...interface{}` in the end? The user can format the string and pass it, right?

cmd/root.go
kadel: added

cmd/root.go
kadel: Yes, for formatting so you can use formatting string with `checkError`  and you don't have to use `fmt.Sprintf` everytime you need a parameter

cmd/watch.go
kadel: fixed

cmd/watch.go
```diff
@@ -0,0 +1,85 @@
...
+		projectName := project.GetCurrent(client)
+		applicationName, err := application.GetCurrent(client)
+		checkError(err, "Unable to get current application.")
+
+		var componentName string
+		if len(args) == 0 {
+			var err error
+			log.Debug("No component name passed, assuming current component")
+			componentName, err = component.GetCurrent(client, applicationName, projectName)
+			if err != nil {
```
containscafeine: we don't need to do `if err != nil` while using `checkError()`

pkg/component/watch.go
```diff
@@ -0,0 +1,204 @@
+package component
```
containscafeine: was just wondering why are we creating a new file, we can add the functions to component/component.go like all the other functions, right?

pkg/component/watch.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/redhat-developer/odo/pkg/occlient"
+	"github.com/redhat-developer/odo/pkg/util"
+
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
+)
+
+// addRecursiveWatch handles adding watches recursively for the path provided
+// and its subdirectories.  If a non-directory is specified, this call is a no-op.
+// Files matching glob patter defined in ignores will be ignored.
```
containscafeine: s/patter/pattern

pkg/component/watch.go
```diff
@@ -0,0 +1,204 @@
...
+			// $ sudo sysctl fs.inotify.max_user_watches=65536
+			// BSD / OSX: "too many open files" issues are ussualy resolved via
+			// $ sysctl variables "kern.maxfiles" and "kern.maxfilesperproc",
+			return fmt.Errorf("error adding watcher for path %s: %v", v, err)
+		}
+	}
+	return nil
+}
+
+// WatchAndPush watched directory dir, if something changes in that directory calls push
```
containscafeine: s/watched/watches

cdrage:
The changes make sense, if we wish to add a paramter later, we can :+1:

pkg/occlient/occlient.go
```diff
@@ -1203,3 +1202,44 @@ func (c *Client) GetOneDeploymentConfigFromSelector(selector string) (*appsv1.De
...
+}
+
+// SyncPath copies local directory to directory in running Pod.
+func (c *Client) SyncPath(localPath string, targetPodName string, targetPath string) (string, error) {
+	// TODO: do this without using 'oc' binary
+	args := []string{
+		"rsync",
+		localPath,
+		fmt.Sprintf("%s:%s", targetPodName, targetPath),
+		"--exclude", ".git",
```
containscafeine: can we also add .swp files to this?

cmd/watch.go
kadel: thx, fixed

pkg/component/watch.go
kadel: fixed

pkg/component/watch.go
kadel: fixed

pkg/occlient/occlient.go
kadel: We could, but then we are starting to ignore files that are editor specific :-( 
I would like to avoid that.  In future, we should probably add an option so users can set it.

pkg/component/watch.go
kadel: We can, I just thought that it this is a nice piece that can be put in the separate file, so we don't end up with everything in one huge file.

I actually don't mind putting it into `component.go` if you think that it is better to have it there.

containscafeine: yeah, I'm okay :man_shrugging:

pkg/occlient/occlient.go
containscafeine: +1

cmd/watch.go
ryanj: What is the use-case for waiting before syncing?  Any downside to removing the delay? Is there any way to remove the delay entirely? 

As a developer, I use "watch" to establish a fast inner dev loop.  Slower tooling is much less useful to me.

kadel: Hi @ryanj 

> What is the use-case for waiting before syncing? Any downside to removing the delay? 

In case of some intensive or batch operation (like a building) there might be a lot of changes, if we would trigger sync immediately after each change it would be ineffective. 

`oc rsync` also uses delayed sync (3 seconds). To be honest I don't remember why I increased it to 5s  😬 

> Is there any way (as a user) to remove the delay entirely?

Not right now. But if there is a real use case for this we could remove delay, and make it configurable via commandline argument

This is just the first implementation mainly to collect feedback. 

> oc rsync already watches for changes. Is another watcher needed?

In future we would like to remove any dependencies on `oc`  binary. Right now it uses `oc rsync` for syncing operation because it was just the quickest way to implement this.

PR #335, opened 2018-04-10
Update desription for ocdev create

cdrage:
Updates the description for ocdev create with a bit more information.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -20,18 +20,25 @@ var (
 var componentCreateCmd = &cobra.Command{
 	Use:   "create <component_type> [component_name] [flags]",
 	Short: "Create new component",
-	Long: `Create new component.
-If component name is not provided, component type value will be used for name.
-	`,
-	Example: `  # Create new nodejs component with the source in current directory. 
+	Long: `Create new component to deploy on OpenShift.
+
+If component name is not provided, component type value will be used for the name.
+
+A full list of component types that can be deployed is available using: 'odo component list'`,
+	Example: `  # Create new Node.js component with the source in current directory. 
   odo create nodejs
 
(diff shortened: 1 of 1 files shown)


cmd/create.go
```diff
@@ -20,18 +20,24 @@ var (
...
 	Short: "Create new component",
-	Long: `Create new component.
-If component name is not provided, component type value will be used for name.
-	`,
-	Example: `  # Create new nodejs component with the source in current directory. 
+	Long: `Create new component to deploy on OpenShift.
+
+If component name is not provided, component type value will be used for the name.
+
+OpenShift built-in types: httpd, nodejs, perl, php, python, ruby, wildfly
```
kadel: I'm not sure if it is a good idea to have the list here. It will be hard to keep this update.
Just referring to `ocdev component list` is I think enough.

cdrage:
Thanks @kadel I've updated it and added one more line with regards to using Python, but that's it. You're right, since OpenShift is constantly adding new containers / build images, it'd be hard to maintain.

cmd/create.go
```diff
@@ -20,18 +20,25 @@ var (
...
 
-  # Create new nodejs component with source from remote git repository.
+  # Create new Node.js component with source from remote git repository.
   ocdev create nodejs --git https://github.com/openshift/nodejs-ex.git
-	`,
+
+  # Create a Ruby component
+  ocdev create ruby
+	
+	# Create a Python component
```
kadel: indentation. There should be two spaces at the beginning to alight it with everything else.

cdrage: stupid vim haha

cdrage: woops vim

cdrage:
Thx! Rebased :+1:

PR #345, opened 2018-04-11
Fixes error when deleting a component

cdrage:
Adds a new function that sets the current application / project to a
blank "activeComponent".

Fixes: https://github.com/redhat-developer/ocdev/issues/251


kadel:
Would it make sense to check if there are other components and set the next component as active? So we would end up with unset active component only when the last component in the application is deleted.

cdrage:
@kadel I was thinking that too.. however, what if there are two other active components? Should it just pick the last used one / end of array?

My thinking was I didn't want the user to randomly do a `ocdev delete` and delete what he/she thought was their previous or blank component.

kadel:
There can be only one active component at the time. But if there is more than one component left in the application after delete, we could just pick one (first component in List) and set that as current. 
I think that that this is better than leaving it in the state where there is no component active. Of course, we have to display a message informing the user what component was set as active.

cdrage:
@kadel Okay, I'll change it to that :+1:

cdrage:
Updated, not that I am using `log.Warnf` rather than `fmt.Println`.

I hope that we are able to convert the rest of all the warnings / println's under this issue: #364

containscafeine:
This must go in before tomorrow's release

pkg/component/component.go
```diff
@@ -59,22 +59,29 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
 }
 
 // Delete whole component
-func Delete(client *occlient.Client, name string, applicationName string, projectName string) (string, error) {
+func Delete(client *occlient.Client, name string) (string, error) {
```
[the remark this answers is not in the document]
containscafeine: also the project

pkg/config/config.go
```diff
@@ -137,6 +138,34 @@ func (c *ConfigInfo) SetActiveComponent(component string, application string, pr
 	return nil
 }
 
+// Sets the active component as blank in the configuration file
+func (c *ConfigInfo) UnsetActiveComponent(application string, project string) error {
```
containscafeine: :+1:

pkg/component/component.go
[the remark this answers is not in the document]
cdrage: okay :+1:

cdrage:
@kadel @containscafeine Updated the PR with your comments, should be good for another review now :+1:

pkg/component/component.go
```diff
@@ -60,21 +60,39 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
+	// If there are components that still exist (len > 0) then set it to the default component..
+	components, err := List(client, applicationName, projectName)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to delete component %s", name)
+		return "", errors.Wrapf(err, "unable to retrieve list of components")
+	}
+
+	// If there's more than one component, set it to the first one..
+	if len(components) > 0 {
+		err = cfg.SetActiveComponent(components[0].Name, applicationName, projectName)
```
kadel: We have to tell the user what application was just set as active.

pkg/component/component.go
```diff
@@ -60,21 +60,39 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
+			log.Warnf("No active component found, setting active component to %s", components[0].Name)
+			return "", errors.Wrapf(err, "unable to set current component %s", name)
+		}
+	} else {
+		// Unset to blank
+		err = cfg.UnsetActiveComponent(applicationName, projectName)
+		if err != nil {
+			return "", errors.Wrapf(err, "error unsetting current component while deleting %s", name)
+		}
+		log.Warnf("No default component has been set")
```
kadel: I don't think this is Warn. It is the normal state. There are just no components left. 
I think that the message should say just that.

pkg/component/component.go
```diff
@@ -60,21 +60,39 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to delete component %s", name)
+		return "", errors.Wrapf(err, "unable to retrieve list of components")
+	}
+
+	// If there's more than one component, set it to the first one..
+	if len(components) > 0 {
+		err = cfg.SetActiveComponent(components[0].Name, applicationName, projectName)
+		if err != nil {
+			log.Warnf("No active component found, setting active component to %s", components[0].Name)
```
kadel: How can we say that active component was set when there was an error in the call setting it as active?

pkg/component/component.go
cdrage: Done! (See next line)

cdrage:
ping @kadel

pkg/component/component.go
```diff
@@ -60,21 +60,41 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
+	components, err := List(client, applicationName, projectName)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to delete component %s", name)
+		return "", errors.Wrapf(err, "unable to retrieve list of components")
+	}
+
+	// If there's more than one component, set it to the first one..
+	if len(components) > 0 {
+		err = cfg.SetActiveComponent(components[0].Name, applicationName, projectName)
+		log.Warnf("No active component found, setting active component to %s", components[0].Name)
```
kadel: This shouldn't be a warning. This is normal operation.
Normal message saying that `Setting active component to %s", components[0].Name.` Is enought.

cdrage:
@kadel I've gone ahead and fixed it!

```sh
github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                58m ⚑  
▶ ./odo list
You have deployed:
nodejs using the nodejs component
nodejs2 using the nodejs component
nodejs3 using the nodejs component

github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                58m ⚑  
▶ ./odo component get
The current component is: nodejs3

github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                58m ⚑  
▶ ./odo delete nodejs2
Are you sure you want to delete nodejs2 from nodeapp? [y/N] y
buildconfig "nodejs2" deleted
imagestream "nodejs2" deleted
deploymentconfig "nodejs2" deleted
service "nodejs2" deleted


github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                58m ⚑  
▶ ./odo component get
The current component is: nodejs3

github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                58m ⚑  
▶ ./odo delete nodejs3
Are you sure you want to delete nodejs3 from nodeapp? [y/N] y
No active component found, setting active component to 'nodejs'
buildconfig "nodejs3" deleted
imagestream "nodejs3" deleted
deploymentconfig "nodejs3" deleted
service "nodejs3" deleted


github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                59m ⚑  
▶ ./odo list
You have deployed:
nodejs using the nodejs component

github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                59m ⚑  
▶ ./odo component get
The current component is: nodejs

github.com/redhat-developer/odo  fix-deleting-error ✗                                                                                                                                                                                                                                                                                                                60m ⚑  
▶ ./odo delete nodejs
Are you sure you want to delete nodejs from nodeapp? [y/N] y
No default component has been set
buildconfig "nodejs" deleted
imagestream "nodejs" deleted
deploymentconfig "nodejs" deleted
service "nodejs" deleted
```

pkg/component/component.go
```diff
@@ -60,21 +60,46 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
+	}
+
+	// We will *only* set a new component if either len(components) is zero, or the
+	// current component matches the one being deleted.
+	if current := cfg.GetActiveComponent(applicationName, projectName); current == name || len(components) == 0 {
+
+		// If there's more than one component, set it to the first one..
+		if len(components) > 0 {
+			err = cfg.SetActiveComponent(components[0].Name, applicationName, projectName)
+			fmt.Printf("No active component found, setting active component to '%s'\n", components[0].Name)
```
kadel: you shouldn't print this in `component.Delete`.
I think that this logic should be in `cmd`, like you had it before.

pkg/component/component.go
```diff
@@ -60,21 +60,46 @@ func CreateFromDir(client *occlient.Client, name string, ctype string, dir strin
...
+				return "", errors.Wrapf(err, "unable to set current component to '%s'", name)
+			}
+		} else {
+			// Unset to blank
+			err = cfg.UnsetActiveComponent(applicationName, projectName)
+			if err != nil {
+				return "", errors.Wrapf(err, "error unsetting current component while deleting %s", name)
+			}
+
+			fmt.Println("No default component has been set")
```
containscafeine: we should not print here

pkg/component/component.go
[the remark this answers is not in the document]
kadel: Do we really need this if there is already the same message `fmt.Printf("Default component set to: %s", currentComponent)` in `cmd`?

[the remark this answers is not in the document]
cdrage: I *just* added that line :+1:

cdrage:
Finallyyyy, there we go. Changes made @kadel @containscafeine

PR #346, opened 2018-04-11
Update the error outout for occlient

cdrage:
Updates the error output with (slightly) better information.


kadel:
👍 thanks @cdrage

PR #352, opened 2018-04-12
Separate pkg for labels

kadel:
- move label handling to separate packages to avoid cyclic imports
 - refactor `component.GetComponentDeploymentConfig` to `occlient.GetOneDeploymentConfigFromSelector`

This avoids importing component in storage pkg, so there is no risk of cyclic imports.

ping @containscafeine and @surajnarwade


surajnarwade:
LGTM

Issue #342, opened 2018-04-11
Add link command 


kadel:
subtask of #171 

example:
link current component with backend component:
``` 
odo link backend
```
```
odo link backend --component frontend
```
Link frontend component to the backend.

What that means is that env variables describing how to connect to backends Service will be injected to frontends DeploymentConfig.

Steps to do this:
Find service for `backend` component.
Get ip and port of that service.
Add following env variables to DeploymentConfig for `frontend` component.
```
COMPONENT_BACKEND_ADDRESS=<some address>
COMPONENT_BACKEND_PORT=<some port>
```

update
`COMPONENT_BACKEND_ADDRESS` should be the hostname of the Service lets use  `COMPONENT_BACKEND_HOST` instad

containscafeine:
Fixed in #354

PR #354, opened 2018-04-12
Add `ocdev link` command

containscafeine:
This commit adds the `ocdev link` command which injects the
connection information from the target component to the source
component


containscafeine:
Fix #342

pkg/component/component.go
```diff
@@ -360,3 +362,91 @@ func Exists(client *occlient.Client, componentName string) (bool, error) {
 	}
 	return false, nil
 }
+
+// GetComponentService returns the Service object associated with the given
+// component.
+// An error is thrown when exactly one Service is not found for the component.
+func GetComponentService(client *occlient.Client, componentName string, applicationName string) (*corev1.Service, error) {
```
kadel: I'm not sure that this function belongs here.  If you really want to keep it here than it definitely shouldn't be public. Otherwise, it is a potential risk for creating cyclic imports like in the `storage` :-(

It might be better to do it similarly like here: https://github.com/redhat-developer/odo/pull/352/commits/f2d5b10ca01fec612f7349e606ba3966ea107574#diff-6025e9b620e9321c03c11bf29e5c4dbdR1190

pkg/component/component.go
```diff
@@ -360,3 +362,91 @@ func Exists(client *occlient.Client, componentName string) (bool, error) {
...
+		return nil, fmt.Errorf("multiple Services exist for the selector: %v. Only one must be present", selector)
+	}
+
+	return &services[0], nil
+}
+
+// LinkInfo contains the information about the link being created between
+// components
+type LinkInfo struct {
+	ComponentName string
```
kadel: Is this source or target component name? 
Why is there only one?
There should be either comment, or better variable name should specify that.

pkg/component/component.go
```diff
@@ -360,3 +362,91 @@ func Exists(client *occlient.Client, componentName string) (bool, error) {
...
+			{
+				Name:  linkPortKey,
+				Value: linkPortValue,
+			},
+		}, sourceDC); err != nil {
+		return nil, errors.Wrapf(err, "unable to add environment variables to Deployment Config %v", sourceDC.Name)
+	}
+
+	return &LinkInfo{
+		ComponentName: sourceComponent,
```
kadel: Why are we returning just sourceComponent?

pkg/component/component.go
containscafeine: moved this similar to the current deployment config logic

pkg/component/component.go
containscafeine: added source and target components

pkg/component/component.go
containscafeine: modified LinkInfo to include both source and target components

cmd/link.go
```diff
@@ -0,0 +1,68 @@
...
+	Short: "Link target component to source component",
+	Long: `Link target component to source component
+If source component is not provided, the link is created to the current active
+component.
+In the linking process, the environment variables containing the connection
+information from target component are injected into the source component and
+printed to STDOUT.
+`,
+	Example: `
+	# Link current component to a component 'nodejs'
```
kadel: There should be just two spaces at the beginning of each line, to make it properly align

cmd/link.go
```diff
@@ -0,0 +1,68 @@
...
+	ocdev link nodejs
```
kadel: Better example would be to do `ocdev link mariadb` as you will be injecting variables about mariadb  to current component

cmd/link.go
```diff
@@ -0,0 +1,68 @@
...
+
+	# Link 'nodejs' component to 'mariadb' component
+	ocdev link nodejs --component mariadb
```
kadel: again this should be the other way around `ocdev link mariadb --component nodejs`

cmd/link.go
containscafeine: fixed

cmd/link.go
containscafeine: fixed

cmd/link.go
containscafeine: fixed

cmd/util.go
```diff
@@ -0,0 +1,18 @@
...
+// getComponent returns the component to be used for the operation. If an input
+// component is specified, then it is returned, if not, the current component
+// is fetched and returned
+func getComponent(client *occlient.Client, inputComponent, applicationName, projectName string) string {
+	if len(inputComponent) == 0 {
+		c, err := component.GetCurrent(client, applicationName, projectName)
+		checkError(err, "Could not get current component")
+		return c
+	}
+	return inputComponent
```
kadel: It would be good to check if the component exists here, so we don't have to do it each time we call this function.

(we need to check it only if `len(inputComponent)!=0`

pkg/component/component.go
```diff
@@ -7,11 +7,15 @@ import (
 	buildv1 "github.com/openshift/api/build/v1"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+
+	"strings"
```
kadel: please group this with other standard library imports

pkg/component/component.go
```diff
@@ -239,3 +243,83 @@ func Exists(client *occlient.Client, componentName, applicationName, projectName
...
+	SourceComponent string
+	TargetComponent string
+	Envs            []string
+}
+
+// Link injects connection information of the target component into the source
+// component as environment variables
+func Link(client *occlient.Client, sourceComponent, targetComponent, applicationName string) (*LinkInfo, error) {
+	// Get Service associated with the target component
+	targetService, err := client.GetOneServiceFromSelector(
```
kadel: my head hurts from trying to parse this :-( 
Would it be better to have separate this? It is hard to parse 3 nested function calls :(

cmd/util.go
containscafeine: added

pkg/component/component.go
containscafeine: fixed

pkg/component/component.go
```diff
@@ -239,3 +243,83 @@ func Exists(client *occlient.Client, componentName, applicationName, projectName
...
+		TargetComponent: targetComponent,
+		Envs: []string{
+			fmt.Sprintf("%v=%v", linkHostKey, linkHostValue),
+			fmt.Sprintf("%v=%v", linkPortKey, linkPortValue),
+		},
+	}, nil
+}
+
+// getPortFromService returns the first port listed in the service
+func getPortFromService(service *corev1.Service) (int32, error) {
```
kadel: Do we really need a separate function that is just one if statement and nothing more? This is used just in one place.

cmd/util.go
```diff
@@ -0,0 +1,26 @@
...
+// getComponent returns the component to be used for the operation. If an input
+// component is specified, then it is returned, if not, the current component
+// is fetched and returned
+func getComponent(client *occlient.Client, inputComponent, applicationName, projectName string) string {
+	if len(inputComponent) == 0 {
+		c, err := component.GetCurrent(client, applicationName, projectName)
+		checkError(err, "Could not get current component")
+		return c
+	}
+	exists, err := component.Exists(client, inputComponent, applicationName, projectName)
```
kadel: Do we need to check that component exists if we got it from `GetCurrent`?
`Exists` is expensive it might be enough to put into `else` block for previous if.

pkg/component/component.go
containscafeine: haha, fixed

pkg/component/component.go
containscafeine: imo it makes the Link() function look cleaner

cmd/util.go
containscafeine: we're not checking if we're getting it from GetCurrent. The if block has a return statement, we don't need an else here

pkg/component/component.go
kadel: IMO it makes it harder to read ;-)

cmd/util.go
kadel: ah, sorry, Didn't notice return there. LGTM than ;-)

PR #355, opened 2018-04-12
Remove unnecessary if condition

containscafeine:
Remove unnecessary if condition


surajnarwade:
@containscafeine rebase it with latest changes `ODO`

PR #357, opened 2018-04-12
Check if component type exists before deploying

containscafeine:
This component first queries the catalog if the given component
type is supported or not. If the component type does not match
any name in catalog list, an error is thrown.


containscafeine:
rebased, merging

PR #358, opened 2018-04-12
Unify application create logic

### LINKED ISSUE #350: Application should not be created in component package
We are calling `application.Create()` a lot of times in  `pkg/component/component.go`. This should not happen. The application creating logic should be outside `package component`, it should probably be called in `package cmd` before calling `component.Create()`. As an effect of this, the application name will be passed as a parameter to `component.Create()`.Z

concaf:
This commit removes the application create logic from component
and url packages. Application is now only created at 2 places -
- when `ocdev application create` is called
- when `ocdev create` is called without creating a new application

This commit also refactors GetCurrentOrGetAndSetDefault() function
to GetCurrentOrGetCreateSetDefault() which now also creates the
default application if there is no current application set

Fix #350

### DIFF
--- a/cmd/component.go
+++ b/cmd/component.go
@@ -2,7 +2,10 @@ package cmd
 
 import (
 	"fmt"
+
+	"github.com/redhat-developer/odo/pkg/application"
 	"github.com/redhat-developer/odo/pkg/component"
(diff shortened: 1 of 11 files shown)


kadel:
At the first glance, the code looks good. But rebase is required before we can test it :-(

Issue #305, opened 2018-04-04
Delete should check earlier if a component exists


cdrage:
We should have a more informative error / error out earlier if a component does not exist:


```
github.com/redhat-developer/ocdev  create-components-same-name ✗                                                                                                                                                                                                                                                                                                    21m ⚑  ⍉
▶ ./ocdev delete thisisatest -f
unable to delete component thisisatest: unable to set  component as active, application myproject in app project doesn't exists
```

containscafeine:
@cdrage now that #277 is in, do you still face this?

containscafeine:
Because #277 prints only the cause of the error, and not the entire error. The entire error is still visible in verbose mode though.

mik-dass:
@cdrage This is the current output 

    [mdas@localhost ocdev]$ ./ocdev delete thisisatest -f
    unable to set  component as active, application blah-project in app project doesn't exists

cdrage:
Still bad error / descriptions ^^ let's keep this open

mik-dass:
I think we should close it as the fix for this has been merged.

containscafeine:
@mik-dass can you link the PR/commit?

kadel:
> @mik-dass can you link the PR/commit?

it is already mentioned above @mik-dass 's comment

containscafeine:
Okay, it's #359, missed it :P

PR #359, opened 2018-04-13
Adds check if a component exists before deleting

mik-dass:
Issue : #305


mik-dass:
@kadel Is the error message good enough?

PR #363, opened 2018-04-13
Don't automatically start build when component is created

### LINKED ISSUE #327: `create --local` should NOT trigger build
When a local component is created the build shouldn't be triggered.
It should just create a component and wait for the user to run `ocdev push` to build component.

If the component is created from remote git than it is ok to start building component right away.

kadel:
fixes #327 

BuildConfig is created without any Triggers.
Build is started explicitly after component from git repo is created.
Build is NOT started if the component is created from a local directory.

Messages informing user what is happening was added.

How output looks like
local component:
```
▶ odo create wildfly
Component 'wildfly' was created.

Component 'wildfly' is now set as active component.
```
git component:
```
▶ odo create nodejs --git https://github.com/openshift/nodejs-ex.git
Component 'nodejs' was created.
Triggering build from https://github.com/openshift/nodejs-ex.git.

Cloning "https://github.com/openshift/nodejs-ex.git" ...
	Commit:	93fc25fd35068aa311f4722e4d7e5187b06edd7b (Merge pull request #170 from dharmit/add-minishift-to-readme)
	Author:	Ben Parees <bparees@users.noreply.github.com>
	Date:	Tue Mar 6 13:34:31 2018 -0500
---> Installing application source ...
---> Installing all dependencies
....
....
....
Pushed 10/10 layers, 100% complete
Push successful

Component 'nodejs' is now set as active component.
```

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -545,15 +545,6 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 					},
 				},
 			},
-			Triggers: []buildv1.BuildTriggerPolicy{
-				{
-					Type: "ConfigChange",
-				},
-				{
-					Type:        "ImageChange",
-					ImageChange: &buildv1.ImageChangeTrigger{},
-				},
-			},
 		},
 	}
 	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -83,22 +83,31 @@ A full list of component types that can be deployed is available using: 'odo com
 		if len(componentGit) != 0 {
 			err := component.CreateFromGit(client, componentName, componentType, componentGit, applicationName)
 			checkError(err, "")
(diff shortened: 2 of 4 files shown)


pkg/occlient/occlient.go
```diff
@@ -545,15 +545,7 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
...
-			Triggers: []buildv1.BuildTriggerPolicy{
-				{
-					Type: "ConfigChange",
-				},
-				{
-					Type:        "ImageChange",
-					ImageChange: &buildv1.ImageChangeTrigger{},
-				},
-			},
+			Triggers: []buildv1.BuildTriggerPolicy{},
```
containscafeine: nit: we can simply drop the `Triggers:` field, we don't need it at all :man_shrugging:

kadel: You are right.

I wanted to leave it here so everyone can quickly see that we are not setting any triggers.

kadel: removed  it ;-)

PR #365, opened 2018-04-13
Adds logrus to existing logging output (fmt.Println)

cdrage:
Changes the fmt.Println output to either Info or Error in each
corresponding file, using logrus as the consistent logger outputter
throughout odo.

Closes https://github.com/redhat-developer/odo/issues/364<Paste>

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+
 	"github.com/redhat-developer/odo/pkg/catalog"
 	"github.com/spf13/cobra"
 )
--- a/cmd/component.go
+++ b/cmd/component.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+
 	"github.com/redhat-developer/odo/pkg/component"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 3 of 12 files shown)


cdrage:
Possible [WIP] since tests (may) fail due to test output / comparison checks!

cdrage:
Ready for review @kadel @containscafeine

surajnarwade:
@cdrage , when I tried to build `odo` from this PR, I got following error

```
go build -o odo ./main.go
# github.com/redhat-developer/odo/cmd
cmd/create.go:67:4: undefined: fmt
cmd/create.go:93:4: undefined: fmt
cmd/create.go:94:4: undefined: fmt
cmd/create.go:103:4: undefined: fmt
cmd/create.go:104:4: undefined: fmt
cmd/create.go:110:4: undefined: fmt
cmd/create.go:111:4: undefined: fmt
cmd/create.go:117:3: undefined: fmt
cmd/update.go:13:2: component redeclared as imported package name
	previous declaration at cmd/update.go:8:2
cmd/update.go:15:2: cobra redeclared as imported package name
	previous declaration at cmd/update.go:10:2
cmd/create.go:117:3: too many errors
```

cdrage:
@surajnarwade Yeah, I need to rebase again.. Thanks :+1:

kadel:
Can we update the format of the messages? 
Having `INFO[0002]`  at the beginning of every message doesn't look nice.
I don't think that there should be any prefix for `Info` messages. Only for `Error: ` and `Warning: `.

What will be rules for using `fmt.Printf` vs `log.Info`?

One small problem with this is that `log.Info` prints to stderr :-(

cdrage:
@kadel I'm going to close this for now as:

1. Too many conflicts
2. We're moving too fast and it'd be better to update the logging everywhere (and also update logrus as the timestamps (INFO[0002]) are still showing.

Issue #296, opened 2018-03-29
Use `app` alias instead of `application`


cdrage:
IMO `app` is much faster / easier to type, should we prioritise / have `ocdev --help` highlight `app` rather that `application`?

piyush1594:
since #367 is merged, let's close this.

PR #367, opened 2018-04-14
Make app the main command and application as alias

piyush1594:
Now when hitting odo command app will be show as command
application is made alias so that users can use both
#296


kadel:
@piyush1594 can you please modify all e2e tests to use `odo app` instead of  `odo application` the same for docs, if it is mentioned anywhere

piyush1594:
@kadel Yaa sure, will do that and soon push. Thanks

surajnarwade:
Tried / Tested, Worked for me, LGTM

Issue #377, opened 2018-04-16
Glide issues (unable to find 'i' in github.com/openshift/api/route/v1)


cdrage:
Anyone else getting these Glide issues when running `glide update --strip-vendor`?

```sh
github.com/redhat-developer/odo  coloured-output ✗                                                                                                                                                                                                                                                                                                                   8m ⚑  ⍉
▶ glide cc
[INFO]  Glide cache has been cleared.

github.com/redhat-developer/odo  coloured-output ✗                                                                                                                                                                                                                                                                                                                    8m ⚑  
▶ glide update --strip-vendor
[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Fetching github.com/blang/semver.
[INFO]  --> Fetching github.com/google/go-github.
[INFO]  --> Fetching github.com/pkg/errors.
[INFO]  --> Fetching github.com/sirupsen/logrus.
[INFO]  --> Fetching github.com/spf13/pflag.
[INFO]  --> Fetching github.com/spf13/cobra.
[INFO]  --> Fetching github.com/openshift/client-go.
[INFO]  --> Fetching github.com/openshift/source-to-image.
[INFO]  --> Fetching github.com/onsi/gomega.
[INFO]  --> Fetching github.com/onsi/ginkgo.
[INFO]  --> Fetching github.com/kubernetes-incubator/service-catalog.
[INFO]  --> Fetching github.com/satori/go.uuid.
[INFO]  --> Fetching github.com/ghodss/yaml.
[INFO]  --> Fetching github.com/fatih/color.
[INFO]  --> Fetching github.com/fsnotify/fsnotify.
[INFO]  --> Setting version for github.com/ghodss/yaml to v1.0.0.
[INFO]  --> Setting version for github.com/sirupsen/logrus to v1.0.5.
[INFO]  --> Setting version for github.com/spf13/pflag to v1.0.0.
[INFO]  --> Setting version for github.com/pkg/errors to v0.8.0.
[INFO]  --> Setting version for github.com/fsnotify/fsnotify to v1.4.7.
[INFO]  --> Setting version for github.com/google/go-github to v15.0.0.
[INFO]  --> Setting version for github.com/fatih/color to v1.6.0.
[INFO]  --> Setting version for github.com/satori/go.uuid to v1.1.0.
[INFO]  --> Setting version for github.com/spf13/cobra to v0.0.1.
[INFO]  --> Setting version for github.com/blang/semver to v3.5.1.
[...]

PR #378, opened 2018-04-16
Add coloured output to OpenShift logs

cdrage:
This adds a slight-yellow to the OpenShift logs when it's streamed in
order to differentiate between OpenShift and `odo` logs (and to add a
bit of colour to the output / improve the UI/UX)


cdrage:
Closes issue: https://github.com/redhat-developer/odo/issues/377 too.

kadel:
> Closes issue: #377 too.

how is this fixing that issue?

glide.yaml
```diff
@@ -16,6 +16,10 @@ import:
   version: v0.0.1
 - package: github.com/spf13/pflag
   version: v1.0.0
+- package: github.com/openshift/api
+- package: github.com/docker/docker
```
kadel: Why docker? We don't import it anywhere.

pkg/occlient/occlient.go
```diff
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
```
kadel: shouldn't this be in glide.yaml?

glide.yaml
cdrage: To further your comment, this is added to fix the Glide issue: https://github.com/redhat-developer/odo/issues/377 OpenShift uses docker/docker

cdrage: This fixes the issue by specifying that we're going to use github.com/openshift/api as well as github.com/docker/docker within Glide. After going through: https://github.com/Masterminds/glide/issues/945 the solution was to simply add it to `glide.yaml` and it'll fix itself.

pkg/occlient/occlient.go
kadel: Yep, that other they might remove it one day. We should have it in glide.yaml if we are explicitly importing it somewhere in our codebase.

glide.yaml
kadel: Can you please add a comment explaining why we have it in glide.yml even tho we are not importing it anywhere?

glide.yaml
```diff
@@ -16,6 +16,12 @@ import:
   version: v0.0.1
 - package: github.com/spf13/pflag
   version: v1.0.0
+- package: github.com/openshift/api
```
kadel: please add the same version as  `client-go` has

pkg/occlient/occlient.go
```diff
@@ -762,6 +764,10 @@ func (c *Client) FollowBuildLog(buildName string) error {
 	}
 	defer rd.Close()
 
+	// Set the colour of the stdout output..
+	color.Set(color.FgYellow)
```
kadel: will this work on Windows. 
Accroding to docs you have to use 
```
fmt.Fprintf(color.Output, "Windows support: %s", color.GreenString("PASS"))
```

to print in color on windows

glide.yaml
```diff
@@ -16,6 +16,12 @@ import:
...
+- package: github.com/fatih/color
+  version: v1.6.0
+- package: github.com/docker/docker
```
kadel: Do we really need it? It worked fine for me without this.

cdrage: I'll remove it and try another version of glide, if it doesn't work I'll add an issue upstream!

pkg/occlient/occlient.go
cdrage: I tried to find an alternative (coloured output by modifying stdout), but since we stream the data, I couldn't do it. 

I think we should just support Mac OS X and Linux for now.

cdrage:
@kadel Updated

Let's not support Windows for now, too cumbersome to implement since we use streaming and I couldn't find any support for stdout modification with our current implementation.

cdrage:
@kadel Could we merge this and set-up a TODO / another issue to support Windows? I wouldn't want to devote too much time debugging / changing the way we do stdout for the streaming logs. I played around a bit implementing this but I was unable to get it to work.

cdrage:
@kadel I got it to compile / work. Literally was just one line! I didn't think it'd be that easy / you can use io.Writer instead of os.Stdout.

Up for one-more review!

cdrage:
@kadel Change is here: https://github.com/redhat-developer/odo/pull/378/commits/e3f9582090e55206e485835cbeea367ec5b65b34#diff-6025e9b620e9321c03c11bf29e5c4dbdR771

kadel:
Tested this on windows and worked fine in `cmd.exe`, but there is no color in PowerShell, but doesn't break anything either, so it is fine.

Issue #381, opened 2018-04-17
README installation instructions are broken


ryanj:
The binary path listed in the README instructions is incorrect:
https://github.com/redhat-developer/odo#installation

Feel free to disregard if a new release of `odo` is due soon, as the README's install URL will likely get patched as a part of that update process.

kadel:
Yep, said this was caused when we renamed the project to ODO.
Release 0.0.4 will be today, it will fix this.

kadel:
fixed in #386

PR #382, opened 2018-04-17
Fixing release URLs in README install guide #381

ryanj:
see #381


surajnarwade:
@ryanj we don't need to update README since we are having release today with new name `odo`

kadel:
thank you @ryanj.
I've just fixed this with the new release.

PR #383, opened 2018-04-17
Fix infinite loop in e2e tests

### LINKED ISSUE #368: Infinite loop in e2e
There is a for loop that can run forever in e2e tests

https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L219

ashetty1:
Fixes #368

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -45,7 +45,7 @@ func pingSvc(url string) {
 	for {
 		select {
 		case <-pingTimeout:
-			Fail("could not ping the specific service in given time: 10 minutes")
+			Fail("could not ping the specific service in given time: 5 minutes")
 
 		case <-tick:
 			httpTimeout := time.Duration(10 * time.Second)
@@ -69,6 +69,35 @@ func pingSvc(url string) {
 
 		if ep {
 			break
+		}
+	}
+}
+
+// waitForCmdOut runs a command until it gets
+// the expected output.
+// It accepts 2 arguments, cmd (command to be run)
+// & expOut (expected output).
+// It times out if the command doesn't fetch the
+// expected output  within the timeout period (1m).
+func waitForCmdOut(cmd string, expOut string) bool {
+
+	pingTimeout := time.After(1 * time.Minute)
+	tick := time.Tick(time.Second)
+
+	for {
+		select {
+		case <-pingTimeout:
+			Fail("Timeout out after 1 minute")
+
+		case <-tick:
+			out, err := exec.Command("/bin/sh", "-c", cmd).Output()
+			if err != nil {
+				Fail(err.Error())
+			}
+
+			if string(out) == expOut {
+				return true
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -73,6 +73,29 @@ func pingSvc(url string) {
...
+		case <-pingTimeout:
+			Fail("timeout out after 5 minutes")
+
+		case <-tick:
+			out, err := exec.Command("/bin/sh", "-c", cmd).Output()
+			if err != nil {
+				Fail(err.Error())
+			}
+
+			outInt, _ := strconv.Atoi(string(out))
```
kadel: why are we converting the output to int?

tests/e2e/e2e_test.go
```diff
@@ -73,6 +73,29 @@ func pingSvc(url string) {
...
+			if outInt > 0 {
```
kadel: what happens when `outInt` is 0 or less?

tests/e2e/e2e_test.go
ashetty1: @kadel I need the output in int since I am doing `wc -l` comparison

tests/e2e/e2e_test.go
ashetty1: @kadel continues in the loop for 5 minutes

tests/e2e/e2e_test.go
kadel: then we need a different name for this function. It is really specific one command, and it won't work for anything that is not ending with `wc`

kadel: :+1: that looks better. 
It might be good to change function name `pingCmd` doesn't really indicate what this function does. 
Also, don't forget to add a comment with a description what this function is for.

tests/e2e/e2e_test.go
```diff
@@ -73,6 +73,29 @@ func pingSvc(url string) {
...
 }
 
+func pingCmd(cmd string) bool {
+	pingTimeout := time.After(5 * time.Minute)
+	tick := time.Tick(time.Second)
+
+	for {
+		select {
+		case <-pingTimeout:
+			Fail("timeout out after 5 minutes")
```
kadel: Shouldn't be this timeout shorter? 5mins is really long time.
How about 1min? Is there a chance that after 1min the command will succeed?

ashetty1: Sure.  Have changed it to 1min for now. ALso, have changed the function name to `waitForCmdOut`

tests/e2e/e2e_test.go
```diff
@@ -73,6 +73,35 @@ func pingSvc(url string) {
 	}
 }
 
+func waitForCmdOut(cmd string, expOut string) bool {
+	// This function runs the command until it \
```
kadel: The comment should be before the function.
Check this out https://blog.golang.org/godoc-documenting-go-code

ashetty1:
@kadel hope this works.

Issue #373, opened 2018-04-16
Switching to a non-existing project


mik-dass:
[mdas@localhost odo]$ ./odo project list
     ACTIVE   NAME
              blah-project
              myproject
       *      reed
     [mdas@localhost odo]$ ./odo project set nodejs 
     Now using project : nodejs

Without creating currently we can switch to a non-existing project

kadel:
the same is true for component https://github.com/redhat-developer/odo/issues/127

containscafeine:
Closed in #384

PR #384, opened 2018-04-17
Fixes the problem of switching to a non-existing project

mik-dass:
Issue : #373


surajnarwade:
Worked for me, LGTM

Issue #127, opened 2018-02-26
`component.SetCurrent(name string)` should error out if setting to component that doesn't exist


kadel:
`component.SetCurrent(name string)` should error out if setting to component that doesn't exis

Currently, it is not erroring out because `component.List()` is not yet implemented (#53).
When we add it, it should be used in `component.SetCurrent(name string)` to verify that component that is requested to be set as current actually exists.

ashetty1:
The issue below because of this, I assume.

```
# ocdev component delete nodejs
buildconfig "nodejs" deleted
imagestream "nodejs" deleted
deploymentconfig "nodejs" deleted
pod "nodejs-1-24d6r" deleted
service "nodejs" deleted

# ocdev component get
The current component is: nodejs

```

kadel:
@ashetty1 not really,
what you are describing is a different issue (#50), that should be fixed once https://github.com/redhat-developer/ocdev/pull/126 is merged

kadel:
`component.List` was implemented, so we can proceed with this

cdrage:
I've assigned this to myself and started the work here: https://github.com/cdrage/ocdev/tree/check-setcurrent-exists I'll push a PR ASAP.

containscafeine:
Hey @cdrage, I think this should be a pretty similar approach as in #384

cdrage:
Closed via: https://github.com/redhat-developer/odo/pull/388/files

PR #388, opened 2018-04-18
Fixes the issue of switching to a non-existing component using set command

mik-dass:
Issue : #127


surajnarwade:
@mik-dass , run gofmt

mik-dass:
@surajnarwade Done

Issue #219, opened 2018-03-09
Configure the component source to be binary file


jorgemoralespou:
When I create a component I want to be able to specify that the source will be a binary file, so when I push, the binary file is pushed.

```
$ od create wildfly --binary=target/ROOT.war
$ od push
# Push the target/ROOT.war
```

cdrage:
@kadel I believe this is implemented now?

```sh
~                                                                                                                                                                                             
▶ ocdev create --help
create new component

Usage:
  ocdev create [flags]

Flags:
      --binary string   binary artifact
      --dir string      local directory as source
      --git string      git source
  -h, --help            help for create

Global Flags:
  -v, --verbose   verbose output
```

kadel:
It is not. Flag is there but it does nothing.

slemeur:
Will need clarify on how this could be achieve in order to be taken in the Milestone-MVP

slemeur:
@kadel : To check the complexity based on the information given by @jorgemoralespou .

kadel:
this is equivalent of running `oc start-build buildname --binary ./path/to/file.war`

kadel:
This will be touching the same parts in code base as #327.
#327 should be fixed before we start working on this.

mik-dass:
@kadel Can we close this?

PR #397, opened 2018-04-19
Adds the binary flag implementation for the create and push command

mik-dass:
Issue : #219

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -37,6 +37,9 @@ A full list of component types that can be deployed is available using: 'odo com
   # Create new Node.js component with source from remote git repository.
   odo create nodejs --git https://github.com/openshift/nodejs-ex.git
 
+  # Create new Wildfly component with binary named sample.war in './downloads' directory
+  odo create wildfly wildly --binary ./downloads/sample.war
+
   # Create a Ruby component
   odo create ruby
 	
@@ -50,12 +53,6 @@ A full list of component types that can be deployed is available using: 'odo com
 		applicationName, err := application.GetCurrentOrGetCreateSetDefault(client)
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
-
-		if len(componentBinary) != 0 {
-			fmt.Printf("--binary is not implemented yet\n\n")
-			cmd.Help()
-			os.Exit(1)
-		}
 
 		checkFlag := 0
 
@@ -89,11 +86,6 @@ A full list of component types that can be deployed is available using: 'odo com
 			componentName = args[1]
 		}
 
-		if len(componentBinary) != 0 {
-			fmt.Printf("--binary is not implemented yet\n\n")
-			os.Exit(1)
-		}
-
 		exists, err = component.Exists(client, componentName, applicationName, projectName)
 		checkError(err, "")
 		if exists {
@@ -112,15 +104,35 @@ A full list of component types that can be deployed is available using: 'odo com
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs(componentLocal)
 			checkError(err, "")
-			err = component.CreateFromDir(client, componentName, componentType, dir, applicationName)
+			err = component.CreateFromPath(client, componentName, componentType, dir, applicationName, "local")
 			checkError(err, "")
 			fmt.Printf("Component '%s' was created.\n", componentName)
 			fmt.Printf("To push source code to the component run 'odo push'\n")
+		} else if len(componentBinary) != 0 {
+			path, err := filepath.Abs(componentBinary)
+			checkError(err, "")
+			fi, err := os.Stat(path)
+			if err != nil {
+				checkError(err, "")
+			}
+			if fi.IsDir() {
+				fmt.Println("Please provide path to binary instead of a directory")
+				os.Exit(1)
+			}
+
+			err = component.CreateFromPath(client, componentName, componentType, path, applicationName, "binary")
+			checkError(err, "")
+
+			err = component.PushLocal(client, componentName, path, true)
+			checkError(err, fmt.Sprintf("failed to push component: %v", componentName))
+
+			fmt.Printf("Component '%s' was created.\n", componentName)
+
 		} else {
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs("./")
 			checkError(err, "")
-			err = component.CreateFromDir(client, componentName, componentType, dir, applicationName)
+			err = component.CreateFromPath(client, componentName, componentType, dir, applicationName, "local")
 			fmt.Printf("Component '%s' was created.\n", componentName)
 			fmt.Printf("To push source code to the component run 'odo push'\n")
 			checkError(err, "")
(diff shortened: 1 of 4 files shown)


cmd/create.go
```diff
@@ -99,15 +89,50 @@ A full list of component types that can be deployed is available using: 'odo com
...
+			}
+			if fi.IsDir() {
+				fmt.Println("Please provide path to binary instead of a directory")
+				os.Exit(1)
+			}
+
+			err = component.CreateFromDir(client, componentName, componentType, componentBinary, applicationName, true)
+			checkError(err, "")
+
+			sourceType, sourcePath, err := component.GetComponentSource(client, componentName, applicationName, projectName)
```
kadel: Why are we getting ComponentSource right after we create it?

pkg/component/component.go
```diff
@@ -44,7 +44,7 @@ func CreateFromGit(client *occlient.Client, name string, ctype string, url strin
 }
 
 // CreateFromDir create new component with source from local directory
-func CreateFromDir(client *occlient.Client, name string, ctype string, dir string, applicationName string) error {
+func CreateFromDir(client *occlient.Client, name string, ctype string, dir string, applicationName string, asFile bool) error {
```
kadel: asFile argument is not used anywhere

pkg/component/component.go
```diff
@@ -135,8 +135,8 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
 }
 
 // PushLocal start new build and push local dir as a source for build
-func PushLocal(client *occlient.Client, componentName string, dir string) error {
-	err := client.StartBinaryBuild(componentName, dir)
+func PushLocal(client *occlient.Client, componentName string, dir string, asFile bool) error {
```
kadel: please add comment explaining what `asFile` argument is

pkg/occlient/occlient.go
```diff
@@ -676,26 +677,37 @@ func (c *Client) GetLatestBuildName(buildConfigName string) (string, error) {
 }
 
 // StartBinaryBuild starts new build and streams dir as source for build
-func (c *Client) StartBinaryBuild(name string, dir string) error {
+func (c *Client) StartBinaryBuild(name string, dir string, asFile bool) error {
```
kadel: please add comment explaining what `asFile` argument is

cmd/create.go
mik-dass: @kadel I wanted to be extra sure about the source but now I have realized that I am checking at the start itself, will fix this

mik-dass:
@kadel `buildv1.BuildSourceBinary` is used for both binary and file or there is some other const for binary?

kadel:
> @kadel buildv1.BuildSourceBinary is used for both binary and file or there is some other const for binary?

sadly no :-( we have to figure out our own way how to recognize those two

pkg/component/component.go
```diff
@@ -21,20 +21,22 @@ import (
...
 
-func CreateFromGit(client *occlient.Client, name string, ctype string, url string, applicationName string) error {
+func CreateFromGit(client *occlient.Client, name string, ctype string, url string, applicationName string, sourceType string) error {
 	labels := componentlabels.GetLabels(name, applicationName, true)
 	// save component type as label
 	labels[componentlabels.ComponentTypeLabel] = ctype
 
 	// save source path as annotation
 	annotations := map[string]string{componentSourceURLAnnotation: url}
+	annotations[componentSourceTypeAnnotation] = sourceType
```
kadel: I think that we can hardcode this as if the `CreateFromGit` is called `sourceType` will always be `git`, isn't it?

pkg/component/component.go
```diff
@@ -44,14 +46,15 @@ func CreateFromGit(client *occlient.Client, name string, ctype string, url strin
 }
 
 // CreateFromDir create new component with source from local directory
-func CreateFromDir(client *occlient.Client, name string, ctype string, dir string, applicationName string) error {
+func CreateFromDir(client *occlient.Client, name string, ctype string, dir string, applicationName string, sourceType string) error {
```
kadel: It might be good to mention in the comment that `sourceType` can be either `local` or `binary` and nothing else.

pkg/component/component.go
```diff
@@ -135,8 +138,9 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
 }
 
 // PushLocal start new build and push local dir as a source for build
-func PushLocal(client *occlient.Client, componentName string, dir string) error {
-	err := client.StartBinaryBuild(componentName, dir)
+// asFile indicates if it is a binary build or not
```
kadel: it is always a "binary build". It would be better to say "binary component".

pkg/occlient/occlient.go
```diff
@@ -678,26 +679,38 @@ func (c *Client) GetLatestBuildName(buildConfigName string) (string, error) {
 }
 
 // StartBinaryBuild starts new build and streams dir as source for build
-func (c *Client) StartBinaryBuild(name string, dir string) error {
+// asFile indicates if it is a binary build or not
```
kadel: You are always starting binary build. Maybe explain that `asFile` indicates that   it will start build with single file specified in `path` parameter, otherwise it assumes that `path` is a directory

pkg/occlient/occlient.go
```diff
@@ -678,26 +679,38 @@ func (c *Client) GetLatestBuildName(buildConfigName string) (string, error) {
...
+func (c *Client) StartBinaryBuild(name string, dir string, asFile bool) error {
```
kadel: `dir` parameter can no longer be called `dir`, as it is not always a directory, but it can be a path to a single file.
 Let's call it `path`

pkg/component/component.go
kadel: `CreateFromDir` is a misleading name for this function, as it is no longer creating component only from dir, but also from file. Also, `dir`  parametr should be called something else as it is not always a directory.

pkg/component/component.go
```diff
@@ -135,8 +138,9 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
 }
 
 // PushLocal start new build and push local dir as a source for build
```
kadel: Please update comment, as it is no longer true. Now it can push the single binary file on top of a directory.

pkg/component/component.go
```diff
@@ -214,7 +218,7 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
 		sourceType = "git"
 		sourcePath = bc.Spec.Source.Git.URI
 	case buildv1.BuildSourceBinary:
-		sourceType = "local"
+		sourceType = bc.ObjectMeta.Annotations[componentSourceTypeAnnotation]
```
kadel: Please update comment for this function (`GetComponentSource`). It says that component source type can be "git" or "local"

pkg/component/component.go
kadel: ping @mik-dass

surajnarwade:
@mik-dass LGTM

surajnarwade:
@kadel @mik-dass it will be good if we add examples in examples directory maybe

surajnarwade:
@mik-dass not in this PR though

PR #399, opened 2018-04-19
Added Check for flags in `create.go`

surajnarwade:
Since there can be only one flag at a time, This PR adds the check for the same.


mik-dass:
I think check for binary should also be included

surajnarwade:
@mik-dass check for binary is happening here, https://github.com/surajnarwade/ocdev/blob/7ae180617297253182276f5f369d87a476df4a58/cmd/create.go#L54

mik-dass:
@surajnarwade But the os.Exit and cmd.Help will be removed in future so I think we can add checkFlag++ in the if block now, otherwise someone removing the exit might forget it and checkFlag won't be incremented

surajnarwade:
@mik-dass , make sense, I have updated PR for the same

mik-dass:
LGTM :)

PR #404, opened 2018-04-20
reformat output of odo list cmd using tabwriter

syamgk:
changes the format of odo list command by using tabwriter for printing the output

fix https://github.com/redhat-developer/odo/issues/256


syamgk:
current output looks like this
```
$odo list
ACTIVE     NAME         TYPE
           frontend     php
*          pysample     python
```

PR #405, opened 2018-04-20
Refactor server version code

containscafeine:
This commit refactors the code to fetch version from the server to
- handle some previously handled errors
- move version info to a struct than an array of strings


piyush1594:
Nice Refactoring

mik-dass:
LGTM :)

PR #406, opened 2018-04-20
Move error checking to checkError()

containscafeine:
This commit refactors some redundant error checking to
checkError()


mik-dass:
LGTM :)

Issue #222, opened 2018-03-09
push command should push contents and start a build


jorgemoralespou:
I want to be able to start a build when **push** command is executed. If the component is configured to  use local or binary content, the content will be uploaded to the build process. If the build is configured with git source, the build will just start.

Depending on how the implementation is done, the build will not be an "OpenShift build" but rather a sync and build with a supervisor process in the container.

Subtask of #169 
Dependencies on #219, #220, #221

slemeur:
This user story is already in place. For the purpose of the MVP release, we are fine. 

If we have time to address the different refinements that we are looking for, that would be wonderful, but I think we should postpone those to after the mvp release. 

Maybe we should be splitting this story, between the user story (already implemented) and the improvements to clarify that here we need to iterate to make the user story better.

LalatenduMohanty:
@kadel We should have added documentation for functionality. I did not see any documentation in the PR. @surajnarwade pointed me to https://github.com/kadel/bootstrap-supervisored-s2i but a normal user can not find out unless documentation is added.

PR #408, opened 2018-04-20
Use Supervisor (Skip BuildConfig process for local components)

kadel:
fixes: #222 
fixes: #207 

requires #397 to work properly. ~~For now there is a ugly hack in `component.GetComponentSource`, that breaks git components.~~

For the quick and iterative way, we can skip the BuildConfig process and build component directly inside running Pod.

When the component is created it uses s2i to inject supervisor process into s2i builder image.
(using [assemble](https://github.com/kadel/bootstrap-supervisored-s2i/blob/master/.s2i/bin/assemble) in  https://github.com/kadel/bootstrap-supervisored-s2i). Then the container is started with  
Supervisor as pid 1 process so it keeps container running. The application itself is run as supervisor service called 'run' (because it  executes s2i run script)
 
Files are synced directly to s2i directory, and then assemble script is run to recompile application in the running pod, and supervisor restarts the application process, without a need to restart the whole container.

TODO:
- [x] more explanation on how this works to this PR ;-)
- [x] run assemble script after each sync
- [x] add supervisor config for `run` script
- [x] send restart signal to the process
- [x] stream assemble script output (currently prints after it finishes)
- [x] update to handle --binary
- [x] fix e2e tests


kadel:
it is ready for review, but don't merge it. I would like to figure out the output streaming before
@cdrage @containscafeine @surajnarwade

pkg/application/application.go
```diff
@@ -97,7 +97,7 @@ func Delete(client *occlient.Client, name string) error {
 	labels := applabels.GetLabels(name, false)
 
 	// delete application from cluster
-	output, err := client.Delete("all", "", labels)
+	output, err := client.Delete("all,pvc", "", labels)
```
cdrage: Just curious, why pvc (shouldn't we only remove pvc via `odo storage delete`?

cdrage: (maybe add comment to code why, I see that we are creating  Pseudo-PVC below..)

pkg/component/component.go
```diff
@@ -57,13 +73,12 @@ func CreateFromPath(client *occlient.Client, name string, ctype string, path str
 	annotations := map[string]string{componentSourceURLAnnotation: sourceURL.String()}
 	annotations[componentSourceTypeAnnotation] = sourceType
 
-	err := client.NewAppS2I(name, ctype, "", labels, annotations)
+	err := client.NewPseudoVPS(name, ctype, labels, annotations)
```
cdrage: VPS? I think you mean PVC?

Maybe "MockPVC" or "FakePVC" would be better since pseudo is hard to type / non-english speakers may be confused.

pkg/component/component.go
```diff
@@ -135,24 +149,74 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
...
-
 }
 
-// PushLocal start new build and push local dir as a source for local or a binary file for a binary build
+// PushLocal push local code to the cluster and trigger build there.
 // asFile indicates if it is a binary component or not
-func PushLocal(client *occlient.Client, componentName string, dir string, asFile bool) error {
-	err := client.StartBinaryBuild(componentName, dir, asFile)
+func PushLocal(client *occlient.Client, componentName string, applicationName string, path string, asFile bool, out io.Writer) error {
+	targetPath := "/opt/app-root/src"
```
cdrage: should it be a constant?

pkg/occlient/occlient.go
```diff
@@ -737,30 +933,85 @@ func (c *Client) StartBinaryBuild(name string, path string, asFile bool) error {
...
+		if e, ok := val.Object.(*buildv1.Build); ok {
+			log.Debugf("Status of %s build is %s", e.Name, e.Status.Phase)
+			switch e.Status.Phase {
+			case buildv1.BuildPhaseComplete:
+				log.Debugf("Build %s completed.", e.Name)
+				return nil
+			case buildv1.BuildPhaseFailed, buildv1.BuildPhaseCancelled, buildv1.BuildPhaseError:
+				return errors.Errorf("build %s status %s", e.Name, e.Status.Phase)
+			}
+		}
```
cdrage: No timeout?

pkg/occlient/occlient.go
```diff
@@ -737,30 +933,85 @@ func (c *Client) StartBinaryBuild(name string, path string, asFile bool) error {
...
+			log.Debugf("Status of %s pod is %s", e.Name, e.Status.Phase)
+			switch e.Status.Phase {
+			case corev1.PodRunning:
+				log.Debugf("Pod %s is running.", e.Name)
+				return e, nil
+			case corev1.PodFailed, corev1.PodUnknown:
+				return nil, errors.Errorf("pod %s status %s", e.Name, e.Status.Phase)
+			}
+		}
+	}
```
cdrage: Add timeout?

pkg/occlient/occlient.go
kadel: So for BuildConfig there is an optional field `completionDeadlineSeconds` that determines the maximal duration of a build. We should probably rely on OpenShift Failing build after the `completionDeadlineSeconds`.  We can't add some arbitrary timeout here, that will be different from what we or user sets in `completionDeadlineSeconds`.
OpenShift will kill build after `completionDeadlineSeconds` If that happens it will result `buildv1.BuildPhaseFailed` state and that is handled.

kadel:
Hi, @cdrage  could you test this and look into why I can't reset font color? I'm lost there :-(

screen shot 2018-04-27 at 16 00 33


The last line is kept yellow, even tho it should be white :-( 
I'm trying to rest the color here:
https://github.com/redhat-developer/odo/pull/408/files#diff-692effa1b0b7a9ed8c1f7bb49a977b9aR194

pkg/component/component.go
```diff
@@ -135,24 +149,81 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
...
+		return errors.Wrap(err, "unable push files to pod")
 	}
+	fmt.Fprintf(out, syncOutput)
+	fmt.Fprintf(out, "Please wait, building component....\n")
+
+	pipeReader, pipeWriter := io.Pipe()
+	go func() {
+		color.Set(color.FgYellow)
+		io.Copy(out, pipeReader)
+		defer color.Unset()
```
cdrage: It looks like you're using this within a `func()` and it's not being set correctly since the Go routine could exit at any time... 

From what I recall, the actual `color.Set` modifies the actual OS terminal rather than the Go binary output. What's happening is that the colour is being set to yellow in the terminal, but it's not being *unset* since the rest of the program is continuing while this Goroutine is running in the background..

Example:

1. PushLocal function starts
2. Goes into the Go routine
3. Colour is set to yellow
4. Rest of the program continues in Yellow
?????? (when the Go routine ends). Finally unset the colour

@kadel

kadel:
Fixed coloring issue. Now ready for another round of review @cdrage @surajnarwade

surajnarwade:
@kadel , works for me :)

surajnarwade:
@kadel , some of the function names that we are calling here are not self explanatory, can you please add single line comment so that it will be useful for future debugging

pkg/occlient/occlient.go
```diff
@@ -1629,21 +1631,42 @@ func (c *Client) GetServerVersion() (*serverInfo, error) {
...
+			TTY:     tty,
+		}, scheme.ParameterCodec)
+
+	config, err := c.kubeConfig.ClientConfig()
+	if err != nil {
+		return errors.Wrapf(err, "unable to get Kubernetes client config")
 	}
 
-	output, err := c.runOcComamnd(&OcCommand{args: args})
+	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
```
surajnarwade: For example, this function name is bit confusing

kadel:
> @kadel , some of the function names that we are calling here are not self explanatory, can you please add single line comment so that it will be useful for future debugging

no problem just point me to those function I'll add comments. ;-) I tried to comment non-obvious thinks but I might miss something.

pkg/occlient/occlient.go
kadel: This function is from client-go library. But I've tried to add comment with explanation.

kadel:
I uploaded binaries to Google Drive for easier testing https://drive.google.com/drive/folders/1BitThBm8I0hPh4JL_orHtE7Dr5Wx5jTf?usp=sharing

pkg/component/component.go
```diff
@@ -135,24 +149,74 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
...
-// PushLocal start new build and push local dir as a source for local or a binary file for a binary build
+// PushLocal push local code to the cluster and trigger build there.
 // asFile indicates if it is a binary component or not
-func PushLocal(client *occlient.Client, componentName string, dir string, asFile bool) error {
-	err := client.StartBinaryBuild(componentName, dir, asFile)
+func PushLocal(client *occlient.Client, componentName string, applicationName string, path string, asFile bool, out io.Writer) error {
+	targetPath := "/opt/app-root/src"
+
+	if !asFile {
+		// We need to make sure that thee is a '/' at the end, otherwise rsync will sync files to wrong directory
```
cdrage: thee -> there

pkg/component/component.go
cdrage: is this fixed?

cdrage:
It works! Tests it and it's awesome. (using the `odo watch` command too).

Side-question, but is it normal to get blank / no build output when first deploying it?

```sh
github.com/redhat-developer/odo  pr_408 ✔                                                                                                                                       4d  ⍉
▶ ./odo create nodejs test --local ~/nodejs-ex
Please wait, creating test component ...
Component 'test' was created.
                                       
Component 'test' is now set as active component.
```

kadel:
> Side-question, but is it normal to get blank / no build output when first deploying it?

Yes, the idea user doesn't care how we create "empty" component. That is just and implementation detail.
But when a user runs `odo push` than it is building his/her application and that is the output that he cares about, as there might be some build error that is in his power to fix.

pkg/component/component.go
kadel: fixed

kadel:
> is this fixed?

yes it is. This is where s2i assembly script expects application source.

cdrage:
Re-base needed, but otherwise code LGTM and implementation works from my side (I built against current master + this PR as well).

kadel:
rebased

kadel:
I don't want to merge this if there are still changes requested :-(
ping @cdrage

kadel:
BTW: this is quite big, and I would like to keep this in multiple commits.

pkg/component/component.go
```diff
@@ -154,7 +154,7 @@ func GetCurrent(client *occlient.Client, applicationName string, projectName str
 // PushLocal push local code to the cluster and trigger build there.
 // asFile indicates if it is a binary component or not
 func PushLocal(client *occlient.Client, componentName string, applicationName string, path string, asFile bool, out io.Writer) error {
-	targetPath := "/opt/app-root/src"
+	const targetPath = "/opt/app-root/src"
```
cdrage: :+1:

cdrage:
Code looks good to me, tests pass as well as using this locally on my machine. :100: all-around from me.

LGTM!

LalatenduMohanty:
@kadel We should have added documentation for this functionality. I did not see any documentation in the PR. @surajnarwade pointed me to https://github.com/kadel/bootstrap-supervisored-s2i but a normal user can not find out unless documentation is added.

kadel:
> @kadel We should have added documentation for this functionality. I did not see any documentation in the PR. @surajnarwade pointed me to https://github.com/kadel/bootstrap-supervisored-s2i but a normal user can not find out unless documentation is added.

I agree that we need documentation. 
I'll add documentation separately in another PR. Basic description of how it works is in this PR description.
At the current state of the project, we don't want to block PR just because of documentation.

PR #413, opened 2018-04-24
Add support for creating component from service catalog

### LINKED ISSUE #266: Add support for service catalog
We have service catalog search implemented in #243, but the current implementation of S2I does not support deploying templates from service catalog. Well, it should.

surajnarwade:
Fixes #266

Signed-off-by: Suraj Narwade <surajnarwade353@gmail.com>

This PR adds the support for creating services from service catalog. for this support, we have used
client from [kubernetes-incubator/service-catalog](https://github.com/kubernetes-incubator/service-catalog).
Currently, we are supporting template service broker only. In future, we will add support for
other brokers as well.

We have added entire new command set for this purpose.

* to check available services in service catalog

```
$ odo service catalog
```

* to create service,

```
$ odo service create mysql-persistent
```

* to delete service,

```
$ odo service delete mysql-persistent
```

* to list the current provisioned services with their status

```
$ odo service list
```

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -7,6 +7,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/redhat-developer/odo/pkg/catalog"
+	svc "github.com/redhat-developer/odo/pkg/service"
 	"github.com/spf13/cobra"
 )
 
@@ -21,15 +22,26 @@ var catalogCmd = &cobra.Command{
 
 var catalogListCmd = &cobra.Command{
 	Use:   "list",
+	Short: "List all available component & service types.",
+	Long:  "List all available component and service types from OpenShift",
+	Example: `  # Get the supported components
+  odo catalog list components
+
+  # Get the supported services from service catalog
+  odo catalog list services
+`,
+}
+
+var catalogListComponentCmd = &cobra.Command{
+	Use:   "components",
 	Short: "List all available component types.",
 	Long:  "List all available component types from OpenShift's Image Builder.",
 	Example: `  # Get the supported components
-  odo catalog list
+  odo catalog list components
 
   # Search for a supported component
-  odo search nodejs
+  odo catalog search nodejs
 `,
-	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		catalogList, err := catalog.List(client)
@@ -50,6 +62,30 @@ var catalogListCmd = &cobra.Command{
 	},
 }
 
+var catalogListServiceCmd = &cobra.Command{
+	Use:   "services",
+	Short: "Lists all the services from service catalog",
+	Long:  "Lists all the services from service catalog",
+	Example: `  # List all services
+  odo catalog list services
+	`,
+	Args: cobra.ExactArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		catalogList, err := svc.ListCatalog(client)
+		checkError(err, "unable to list services because Service Catalog is not enabled in your cluster")
+		switch len(catalogList) {
+		case 0:
+			fmt.Printf("No deployable services found\n")
+		default:
+			fmt.Println("The following services can be deployed:")
+			for _, service := range catalogList {
+				fmt.Printf("- %v\n", service)
+			}
+		}
+	},
+}
+
 var catalogSearchCmd = &cobra.Command{
 	Use:   "search <component name>",
 	Short: "Search component type in catalog",
@@ -59,7 +95,7 @@ This searches for a partial match for the given search term in all the available
 components.
 `,
 	Example: `  # Search for a component
-  odo catalog search pyt
+  odo catalog search python
 	`,
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
@@ -83,7 +119,8 @@ components.
 func init() {
 	catalogCmd.AddCommand(catalogSearchCmd)
 	catalogCmd.AddCommand(catalogListCmd)
-
+	catalogListCmd.AddCommand(catalogListComponentCmd)
+	catalogListCmd.AddCommand(catalogListServiceCmd)
 	// Add a defined annotation in order to appear in the help menu
 	catalogCmd.Annotations = map[string]string{"command": "other"}
 	catalogCmd.SetUsageTemplate(cmdUsageTemplate)
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -40,7 +40,7 @@ var catalogListComponentCmd = &cobra.Command{
   odo catalog list components
 
   # Search for a supported component
-  odo catalog search nodejs
+  odo catalog search components nodejs
 `,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
@@ -68,6 +68,9 @@ var catalogListServiceCmd = &cobra.Command{
 	Long:  "Lists all the services from service catalog",
 	Example: `  # List all services
   odo catalog list services
+
+ # Search for a supported services
+  odo catalog search services mysql
 	`,
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
@@ -87,17 +90,33 @@ var catalogListServiceCmd = &cobra.Command{
 }
 
 var catalogSearchCmd = &cobra.Command{
-	Use:   "search <component name>",
+	Use:   "search",
+	Short: "Search available component & service types.",
+	Long: `Search available component & service types..
+
+This searches for a partial match for the given search term in all the available
+components & services.
+`,
+	Example: `  # Search for a component
+  odo catalog search components python
+
+  # Search for a service
+  odo catalog search service mysql
+	`,
+}
+
+var catalogSearchComponentCmd = &cobra.Command{
+	Use:   "components",
 	Short: "Search component type in catalog",
 	Long: `Search component type in catalog.
 
 This searches for a partial match for the given search term in all the available
 components.
 `,
-	Example: `  # Search for a component
-  odo catalog search python
-	`,
 	Args: cobra.ExactArgs(1),
+	Example: `  # Search for a component
+  odo catalog search components python
+	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		searchTerm := args[0]
@@ -116,11 +135,43 @@ components.
 	},
 }
 
+var catalogSearchServiceCmd = &cobra.Command{
+	Use:   "services",
+	Short: "Search service type in catalog",
+	Long: `Search service type in catalog.
+
+This searches for a partial match for the given search term in all the available
+services from service catalog.
+`,
+	Example: `  # Search for a service
+  odo catalog search services mysql
+	`,
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		searchTerm := args[0]
+		components, err := svc.Search(client, searchTerm)
+		checkError(err, "unable to search for services")
+
+		switch len(components) {
+		case 0:
+			fmt.Printf("No service matched the query: %v\n", searchTerm)
+		default:
+			fmt.Println("The following services were found:")
+			for _, component := range components {
+				fmt.Printf("- %v\n", component)
+			}
+		}
+	},
+}
+
 func init() {
 	catalogCmd.AddCommand(catalogSearchCmd)
 	catalogCmd.AddCommand(catalogListCmd)
 	catalogListCmd.AddCommand(catalogListComponentCmd)
 	catalogListCmd.AddCommand(catalogListServiceCmd)
+	catalogSearchCmd.AddCommand(catalogSearchComponentCmd)
+	catalogSearchCmd.AddCommand(catalogSearchServiceCmd)
 	// Add a defined annotation in order to appear in the help menu
 	catalogCmd.Annotations = map[string]string{"command": "other"}
 	catalogCmd.SetUsageTemplate(cmdUsageTemplate)
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -1,0 +1,186 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/golang/glog"
+	"github.com/redhat-developer/odo/pkg/application"
+	"github.com/redhat-developer/odo/pkg/project"
+	svc "github.com/redhat-developer/odo/pkg/service"
+	"github.com/spf13/cobra"
+)
+
+var serviceForceDeleteFlag bool
+
+// serviceCmd represents the service command
+var serviceCmd = &cobra.Command{
+	Use:   "service",
+	Short: "Perform service catalog operations",
+	Long:  ` Perform service catalog operations, Limited to template service broker only.`,
+	Example: fmt.Sprintf("%s\n%s\n%s",
+		serviceCreateCmd.Example,
+		serviceDeleteCmd.Example,
+		serviceListCmd.Example),
+	Args: cobra.RangeArgs(1, 3),
+}
+
+var serviceCreateCmd = &cobra.Command{
+	Use:   "create <service_type> [service_name]",
+	Short: "Create a new service",
+	Long: `Create a new service from service catalog to deploy on OpenShift.
+
+If service name is not provided, service type value will be used.
+
+A full list of service types that can be deployed is available using: 'odo service catalog'`,
+	Example: `  # Create new mysql-persistent service from service catalog.
+  odo service create mysql-persistent
+	`,
+	Args: cobra.RangeArgs(1, 2),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		applicationName, err := application.GetCurrentOrGetCreateSetDefault(client)
+		checkError(err, "")
+		projectName := project.GetCurrent(client)
+		serviceType := args[0]
+		exists, err := svc.SvcTypeExists(client, serviceType)
+		checkError(err, "unable to create service because Service Catalog is not enabled in your cluster")
+		if !exists {
+			fmt.Printf("Service %v doesn't exist\nRun 'odo service catalog' to see a list of supported services.\n", serviceType)
+			os.Exit(1)
+		}
+		// if only one arg is given, then it is considered as service name and service type both
+		serviceName := args[0]
+		// if two args are given, first is service type and second one is service name
+		if len(args) == 2 {
+			serviceName = args[1]
+		}
+		//validate service name
+		err = validateName(serviceName)
+		checkError(err, "")
+		exists, err = svc.SvcExists(client, serviceName, applicationName, projectName)
+		checkError(err, "")
+		if exists {
+			fmt.Printf("Service with the name %s already exists in the current application.\n", serviceName)
+			os.Exit(1)
+		}
+		err = svc.CreateService(client, serviceName, serviceType, applicationName)
+		checkError(err, "")
+		fmt.Printf("Service '%s' was created.\n", serviceName)
+	},
+}
+
+var serviceDeleteCmd = &cobra.Command{
+	Use:   "delete <service_name>",
+	Short: "Delete an existing service",
+	Long:  "Delete an existing service",
+	Example: `  # Delete service named 'mysql-persistent'
+  odo service delete mysql-persistent
+	`,
+	Args: cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		glog.V(4).Infof("service delete called")
+		glog.V(4).Infof("args: %#v", strings.Join(args, " "))
+		client := getOcClient()
+
+		// Get all necessary names (current application + project)
+		applicationName, err := application.GetCurrent(client)
+		checkError(err, "")
+		projectName := project.GetCurrent(client)
+		serviceName := args[0]
+
+		// Checks to see if the service actually exists
+		exists, err := svc.SvcExists(client, serviceName, applicationName, projectName)
+		checkError(err, "unable to delete service because Service Catalog is not enabled in your cluster")
+		if !exists {
+			fmt.Printf("Service with the name %s does not exist in the current application\n", serviceName)
+			os.Exit(1)
+		}
+
+		var confirmDeletion string
+		if serviceForceDeleteFlag {
+			confirmDeletion = "y"
+		} else {
+			fmt.Printf("Are you sure you want to delete %v from %v? [y/N] ", serviceName, applicationName)
+			fmt.Scanln(&confirmDeletion)
+		}
+
+		if strings.ToLower(confirmDeletion) == "y" {
+			err := svc.DeleteService(client, serviceName, applicationName, projectName)
+			checkError(err, "")
+			fmt.Printf("Service %s from application %s has been deleted\n", serviceName, applicationName)
+
+		} else {
+			fmt.Printf("Aborting deletion of service: %v\n", serviceName)
+		}
+	},
+}
+
+//var serviceCatalogCmd = &cobra.Command{
+//	Use:   "catalog",
+//	Short: "Lists all the services from service catalog",
+//	Long:  "Lists all the services from service catalog",
+//	Example: `  # List all services
+//  odo service catalog
+//	`,
+//	Args: cobra.ExactArgs(0),
+//	Run: func(cmd *cobra.Command, args []string) {
+//		client := getOcClient()
+//		catalogList, err := svc.ListCatalog(client)
+//		checkError(err, "unable to list services because Service Catalog is not enabled in your cluster")
+//		switch len(catalogList) {
+//		case 0:
+//			fmt.Printf("No deployable services found\n")
+//		default:
+//			fmt.Println("The following services can be deployed:")
+//			for _, service := range catalogList {
+//				fmt.Printf("- %v\n", service)
+//			}
+//		}
+//	},
+//}
+
+var serviceListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List all services in the current application",
+	Long:  "List all services in the current application",
+	Example: `  # List all services in the application
+  odo service list
+	`,
+	Args: cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		applicationName, err := application.GetCurrent(client)
+		checkError(err, "")
+		projectName := project.GetCurrent(client)
+		services, err := svc.List(client, applicationName, projectName)
+		checkError(err, "Service Catalog is not enabled in your cluster")
+
+		if len(services) == 0 {
+			fmt.Println("There are no services deployed for this application")
+			return
+		}
+
+		w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
+		fmt.Fprintln(w, "NAME", "\t", "TYPE", "\t", "STATUS")
+		for _, comp := range services {
+			fmt.Fprintln(w, comp.Name, "\t", comp.Type, "\t", comp.Status)
+		}
+		w.Flush()
+	},
+}
+
+func init() {
+	serviceDeleteCmd.Flags().BoolVarP(&serviceForceDeleteFlag, "force", "f", false, "Delete service without prompting")
+
+	// Add a defined annotation in order to appear in the help menu
+	serviceCmd.Annotations = map[string]string{"command": "other"}
+	serviceCmd.SetUsageTemplate(cmdUsageTemplate)
+	serviceCmd.AddCommand(serviceCreateCmd)
+	serviceCmd.AddCommand(serviceDeleteCmd)
+	//serviceCmd.AddCommand(serviceCatalogCmd)
+	serviceCmd.AddCommand(serviceListCmd)
+	rootCmd.AddCommand(serviceCmd)
+}
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -118,30 +118,6 @@ var serviceDeleteCmd = &cobra.Command{
 	},
 }
 
-//var serviceCatalogCmd = &cobra.Command{
-//	Use:   "catalog",
-//	Short: "Lists all the services from service catalog",
-//	Long:  "Lists all the services from service catalog",
-//	Example: `  # List all services
-//  odo service catalog
-//	`,
-//	Args: cobra.ExactArgs(0),
-//	Run: func(cmd *cobra.Command, args []string) {
-//		client := getOcClient()
-//		catalogList, err := svc.ListCatalog(client)
-//		checkError(err, "unable to list services because Service Catalog is not enabled in your cluster")
-//		switch len(catalogList) {
-//		case 0:
-//			fmt.Printf("No deployable services found\n")
-//		default:
-//			fmt.Println("The following services can be deployed:")
-//			for _, service := range catalogList {
-//				fmt.Printf("- %v\n", service)
-//			}
-//		}
-//	},
-//}
-
 var serviceListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List all services in the current application",
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1223,8 +1223,7 @@ func (c *Client) GetBuildConfigFromName(name string, project string) (*buildv1.B
 
 // GetClusterServiceClasses queries the service service catalog to get available clusterServiceClasses
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
-	// TODO: Remove `FieldSelector` from ListOptions when we are confident with Ansible Service Broker
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{FieldSelector: "spec.clusterServiceBrokerName=template-service-broker"})
+	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to list cluster service classes")
 	}
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1129,7 +1129,6 @@ func (c *Client) Delete(labels map[string]string) error {
 		err = c.kubeClient.CoreV1().Services(c.namespace).Delete(svc.Name, &metav1.DeleteOptions{})
 		if err != nil {
 			errorList = append(errorList, "unable to delete service")
-
 		}
 	}
 	// PersistentVolumeClaim
@@ -1148,6 +1147,31 @@ func (c *Client) Delete(labels map[string]string) error {
 
 }
 
+// DeleteServiceInstance takes labels as a input and based on it, deletes respective service instance
+func (c *Client) DeleteServiceInstance(labels map[string]string) error {
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
+	glog.V(4).Infof("Selectors used for deletion: %s", selector)
+	// Service Instance
+	glog.V(4).Infof("Deleting Service Instance")
+	// Listing out serviceInstance because `DeleteCollection` method don't work on serviceInstance
+	svcCatList, err := c.serviceCatalogClient.ServiceInstances(c.namespace).List(metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return errors.Wrap(err, "unable to list service instance")
+	}
+
+	// Iterating over serviceInstance List and deleting one by one
+	for _, svc := range svcCatList.Items {
+		err = c.serviceCatalogClient.ServiceInstances(c.namespace).Delete(svc.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			return errors.Wrap(err, "unable to delete serviceInstace")
+		}
+	}
+
+	return nil
+}
+
+// DeleteProject deletes given project
 func (c *Client) DeleteProject(name string) error {
 	err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})
 	if err != nil {
@@ -1157,8 +1181,9 @@ func (c *Client) DeleteProject(name string) error {
 }
 
 // GetLabelValues get label values of given label from objects in project that are matching selector
-// returns slice of uniq label values
+// returns slice of unique label values
 func (c *Client) GetLabelValues(project string, label string, selector string) ([]string, error) {
+	// List DeploymentConfig according to selectors
 	dcList, err := c.appsClient.DeploymentConfigs(project).List(metav1.ListOptions{LabelSelector: selector})
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to list DeploymentConfigs")
@@ -1171,7 +1196,19 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 			}
 		}
 	}
+
 	return values, nil
+}
+
+// GetServiceInstanceList returns list service instances
+func (c *Client) GetServiceInstanceList(project string, selector string) ([]scv1beta1.ServiceInstance, error) {
+	// List ServiceInstance according to given selectors
+	svcList, err := c.serviceCatalogClient.ServiceInstances(project).List(metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to list ServiceInstances")
+	}
+
+	return svcList.Items, nil
 }
 
 // GetBuildConfigFromName get BuildConfig by its name
@@ -1184,14 +1221,43 @@ func (c *Client) GetBuildConfigFromName(name string, project string) (*buildv1.B
 	return bc, nil
 }
 
-// GetClusterServiceClasses queries the service service catalog to get all the
-// currently available cluster service classes
+// GetClusterServiceClasses queries the service service catalog to get available clusterServiceClasses
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
+	// TODO: Remove `FieldSelector` from ListOptions when we are confident with Ansible Service Broker
+	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{FieldSelector: "spec.clusterServiceBrokerName=template-service-broker"})
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to list cluster service classes")
 	}
 	return classList.Items, nil
+}
+
+// CreateServiceInstance creates service instance from service catalog
+func (c *Client) CreateServiceInstance(componentName string, componentType string, labels map[string]string) error {
+	// Creating Service Instance
+	_, err := c.serviceCatalogClient.ServiceInstances(c.namespace).Create(
+		&scv1beta1.ServiceInstance{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "ServiceInstance",
+				APIVersion: "servicecatalog.k8s.io/v1beta1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Finalizers: []string{"kubernetes-incubator/service-catalog"},
+				Name:       componentName,
+				Namespace:  c.namespace,
+				Labels:     labels,
+			},
+			Spec: scv1beta1.ServiceInstanceSpec{
+				PlanReference: scv1beta1.PlanReference{
+					ClusterServiceClassExternalName: componentType,
+				},
+			},
+			Status: scv1beta1.ServiceInstanceStatus{},
+		})
+
+	if err != nil {
+		return errors.Wrap(err, "unable to create service instance")
+	}
+	return nil
 }
 
 // GetClusterServiceClassExternalNames returns the names of all the cluster service
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1155,16 +1155,16 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
 	// Service Instance
 	glog.V(4).Infof("Deleting Service Instance")
 	// Listing out serviceInstance because `DeleteCollection` method don't work on serviceInstance
-	svcCatList, err := c.serviceCatalogClient.ServiceInstances(c.namespace).List(metav1.ListOptions{LabelSelector: selector})
+	svcCatList, err := c.GetServiceInstanceList(c.namespace, selector)
 	if err != nil {
 		return errors.Wrap(err, "unable to list service instance")
 	}
 
 	// Iterating over serviceInstance List and deleting one by one
-	for _, svc := range svcCatList.Items {
+	for _, svc := range svcCatList {
 		err = c.serviceCatalogClient.ServiceInstances(c.namespace).Delete(svc.Name, &metav1.DeleteOptions{})
 		if err != nil {
-			return errors.Wrap(err, "unable to delete serviceInstace")
+			return errors.Wrap(err, "unable to delete serviceInstance")
 		}
 	}
 
@@ -1201,9 +1201,9 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 }
 
 // GetServiceInstanceList returns list service instances
-func (c *Client) GetServiceInstanceList(project string, selector string) ([]scv1beta1.ServiceInstance, error) {
+func (c *Client) GetServiceInstanceList(namespace string, selector string) ([]scv1beta1.ServiceInstance, error) {
 	// List ServiceInstance according to given selectors
-	svcList, err := c.serviceCatalogClient.ServiceInstances(project).List(metav1.ListOptions{LabelSelector: selector})
+	svcList, err := c.serviceCatalogClient.ServiceInstances(namespace).List(metav1.ListOptions{LabelSelector: selector})
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to list ServiceInstances")
 	}
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	scv1beta1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
 	appsv1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
 	dockerapi "github.com/openshift/api/image/docker10"
@@ -14,6 +15,8 @@ import (
 	projectv1 "github.com/openshift/api/project/v1"
 	routev1 "github.com/openshift/api/route/v1"
 
+	applabels "github.com/redhat-developer/odo/pkg/application/labels"
+	componentlabels "github.com/redhat-developer/odo/pkg/component/labels"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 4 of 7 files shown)


kadel:
please add link to the issue

surajnarwade:
Fixes #266

surajnarwade:
Prerequisite to run/test this PR:

* Run following commands,

```
minishift config set openshift-version v3.7.1

export MINISHIFT_ENABLE_EXPERIMENTAL=y

minishift config set iso-url centos

minishift start --service-catalog
``` 
* Enable ansible-service-broker on minishift, refer https://github.com/minishift/minishift-addons/tree/master/add-ons/ansible-service-broker#deploy-the-ansible-service-broker

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,68 @@ import (
 )
 
 // List lists all the available component types
-func List(client *occlient.Client) ([]string, error) {
+func List(client *occlient.Client) (finalList map[string][]string, err error) {
```
kadel: please don't use named return values. It makes it hard to read :-(

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,68 @@ import (
...
 		}
+
 	}
 
 	return result, nil
 }
 
 // Exists returns true if the given component type is valid, false if not
-func Exists(client *occlient.Client, componentType string) (bool, error) {
+func Exists(client *occlient.Client, componentType string) (bool, error, string) {
```
kadel: Please add a comment explaining what return values are. Especially new returned string

pkg/occlient/occlient.go
```diff
@@ -892,6 +899,21 @@ func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, er
 	return classList.Items, nil
 }
 
+func (c *Client) CreateServiceInstance(componentName string, componentType string, labels map[string]string, applicationName string, projectName string) error {
+	// Patching Template with ODO specific labels
+	labelPatch := fmt.Sprintf(`{"labels": {"app.kubernetes.io/component-name": "%s", "app.kubernetes.io/component-type": "%s", "app.kubernetes.io/name": "%s"}}`, componentName, componentType, applicationName)
+	_, err := c.templateClient.Templates("openshift").Patch(componentType, types.StrategicMergePatchType, []byte(labelPatch))
```
kadel: won't this modify templates saved in `openshift` namespace?

pkg/occlient/occlient.go
```diff
@@ -892,6 +899,21 @@ func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, er
 	return classList.Items, nil
 }
 
+func (c *Client) CreateServiceInstance(componentName string, componentType string, labels map[string]string, applicationName string, projectName string) error {
```
kadel: we are in `occlient` package it should have `componentName` or `componentType` labels

pkg/occlient/occlient.go
```diff
@@ -14,16 +14,18 @@ import (
...
 	buildschema "github.com/openshift/client-go/build/clientset/versioned/scheme"
 	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	projectclientset "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	types "k8s.io/apimachinery/pkg/types"
```
kadel: imports should be logically grouped together

cmd/create.go
```diff
@@ -127,6 +127,10 @@ A full list of component types that can be deployed is available using: 'odo com
 			checkError(err, fmt.Sprintf("failed to push component: %v", componentName))
 
 			fmt.Printf("Component '%s' was created.\n", componentName)
+		} else if len(catalogType) != 0 && catalogType == "template" {
```
kadel: what if ` len(catalogType) != 0` and  `catalogType` is different from `"template"`?

pkg/catalog/catalog.go
surajnarwade: sure

kadel:
It breaks for everything if cluster doesn't have service catalog.
It this case regular s2i should work. 
```
▶ odo create wildfly
User "developer" cannot list clusterserviceclasses.servicecatalog.k8s.io at the cluster scope: User "developer" cannot list all clusterserviceclasses.servicecatalog.k8s.io in the cluster (get clusterserviceclasses.servicecatalog.k8s.io)
```

kadel:
> minishift config set openshift-version v3.7.1

why is 3.7.1 required?

pkg/occlient/occlient.go
surajnarwade: that was auto goimports in goland :(

surajnarwade:
@kadel , service catalog & ansible service broker addons works well with this version

pkg/occlient/occlient.go
kadel: what is service is from some other broker than openshift template broker?

pkg/occlient/occlient.go
```diff
@@ -892,6 +899,21 @@ func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, er
...
 
+func (c *Client) CreateServiceInstance(componentName string, componentType string, labels map[string]string, applicationName string, projectName string) error {
+	// Patching Template with ODO specific labels
+	labelPatch := fmt.Sprintf(`{"labels": {"app.kubernetes.io/component-name": "%s", "app.kubernetes.io/component-type": "%s", "app.kubernetes.io/name": "%s"}}`, componentName, componentType, applicationName)
+	_, err := c.templateClient.Templates("openshift").Patch(componentType, types.StrategicMergePatchType, []byte(labelPatch))
+	if err != nil {
+		return errors.Wrap(err, "unable to patch template")
+	}
+	// Creating Service Instance
+	_, err = c.serviceCatalogClient.ServiceInstances(projectName).Create(&scv1beta1.ServiceInstance{metav1.TypeMeta{Kind: "ServiceInstance", APIVersion: "servicecatalog.k8s.io/v1beta1"}, metav1.ObjectMeta{Finalizers: []string{"kubernetes-incubator/service-catalog"}, Name: componentName, Namespace: "myproject", Labels: labels}, scv1beta1.ServiceInstanceSpec{UserInfo: &scv1beta1.UserInfo{Username: "developer"}, PlanReference: scv1beta1.PlanReference{ClusterServiceClassExternalName: componentType}}, scv1beta1.ServiceInstanceStatus{}})
```
kadel: please add break `ServiceInstance` struct to multiple lines,it is hard to see what is going on there.

pkg/occlient/occlient.go
```diff
@@ -892,6 +899,21 @@ func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, er
 	return classList.Items, nil
 }
 
+func (c *Client) CreateServiceInstance(componentName string, componentType string, labels map[string]string, applicationName string, projectName string) error {
+	// Patching Template with ODO specific labels
+	labelPatch := fmt.Sprintf(`{"labels": {"app.kubernetes.io/component-name": "%s", "app.kubernetes.io/component-type": "%s", "app.kubernetes.io/name": "%s"}}`, componentName, componentType, applicationName)
```
kadel: please break this to multiple lines

pkg/occlient/occlient.go
kadel: You are passing labels there. If you can pass labels, why is the patching needed in the previous step?

cmd/create.go
surajnarwade: other catalog type is `imagestream` so other conditions will handle that

pkg/occlient/occlient.go
surajnarwade: I didn't get second question

pkg/occlient/occlient.go
surajnarwade: oh I am gonna remove this label, I was just playing around it. this label will only apply to `serviceinstance`. But we want label on every resource that will be created for example, deploymentconfig, services, routes

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,69 @@ import (
...
+
 	}
 
 	return result, nil
 }
 
 // Exists returns true if the given component type is valid, false if not
-func Exists(client *occlient.Client, componentType string) (bool, error) {
+// Exists returns catalog type either `imagestream` or `template`
+func Exists(client *occlient.Client, componentType string) (bool, error, string) {
```
kadel: it is a convention in go to have an error as last returned parameter

kadel:
> @kadel , service catalog & ansible service broker addons works well with this version

it doesn't work with anything else?

cmd/create.go
```diff
@@ -127,6 +127,10 @@ A full list of component types that can be deployed is available using: 'odo com
 			checkError(err, fmt.Sprintf("failed to push component: %v", componentName))
 
 			fmt.Printf("Component '%s' was created.\n", componentName)
+		} else if catalogType == "template" {
```
kadel: what is someones does something that is from the catalog and combines it with `--git` or `--local`  argument?
We should probably refactor this big `if` `else` ...  to something nicer, and less prone to errors.

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,69 @@ import (
 )
 
 // List lists all the available component types
-func List(client *occlient.Client) ([]string, error) {
+func List(client *occlient.Client) (map[string][]string, error) {
```
kadel: please add comment explaining what are keys and values int this map

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,69 @@ import (
...
-	//	return nil, errors.Wrap(err, "unable to get cluster service classes")
-	//}
-	//catalogList = append(catalogList, clusterServiceClasses...)
+	clusterServiceClasses, err := client.GetClusterServiceClassExternalNames()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get cluster service classes")
+	}
+	var clusterSvcClass []string
+	for _, csClass := range clusterServiceClasses {
+		if !strings.Contains(csClass, "apb") {
```
kadel: why is `apb` excluded?

pkg/occlient/occlient.go
kadel: The function is called CreateServiceInstance, but is using OpenShift templates.
It is really confusing :-( 
There can be multiple different Service brokers in Service Catalog.

pkg/catalog/catalog.go
surajnarwade: we can not deploy apb directly because there are some parameters which needs to be filled, whereas templates have default values for parameters.
cc @containscafeine

cdrage:
@surajnarwade Status of this PR? Do you need another review or still WIP? :+1:

surajnarwade:
@cdrage this is still WIP, Since it's on less priority, I picked other issues and gave a little break here. we'll work on it now

surajnarwade:
**Review** **Needed** :)

_**Environment_ _needed:**_

```
export MINISHIFT_ENABLE_EXPERIMENTAL=y

minishift config set iso-url centos

minishift start --extra-clusterup-flags "--service-catalog"
```

_**Things to test:**_

* `odo catalog list` (it should list templates as well)

* `odo create <any template>`

for example, `odo create jenkins-persistent myjenkins`

* `odo delete myjenkins`

Since, services from APB (ansible playbook broker) are highly parameterised, as per discussion, we are going with Template service broker only as of now (since it has default values as well). Hence you can see only templates in `odo catalog list` along with image names.

cc @kadel @cdrage @mik-dass @syamgk @ashetty1

surajnarwade:
This PR injects odo specific labels into template from `openshift` namespace. This is not good practice though, but There is no other way, because service catalog picks template only from openshift namespace :(

surajnarwade:
@kadel, Here is brief description of issue that I am facing along with some pre-requisite knowledge:

**ClusterServiceBroker**

* Server that manages service

We have two types of broker in openshift,

_* Template Service Broker_

gives visibility to templates (by default, it loads templates from openshift namespace)

_* Ansible Service Broker_

manages application defined by APBs (Ansible Playbook Bundle), APBs leverage Ansible to create a standard mechanism for automating complex deployments.

As of now, anyone wants to introduce service in openshift, they are following ASB.


**ClusterServiceClass**

* Services offered by brokers

(In our case, we are using clusterserviceclass from template service broker only becuase templates have default values and APBs are paramterised)


**ServiceInstance**

* Provisioned instance of clustersesrviceclass

Now take an example of `jenkins-persistent` template or we can call it as ClusterServiceclass coming from Template Service Broker.

when we do `odo create jenkins-persistent`, it instantiate ServiceInstance which means it deploys it's template. Basically deploymentconfig, services, routes for `jenkins-persistent`.

In odo, we inject labels into every resource. so that it helps us for various operations such as `odo list` for an example.

To inject such labels while creating serviceInstance (when we do odo create) is not possible. because, serviceInstance don't poulate this label into it's children resources. For this process, there is open PR on openshift openshift/origin#19467 (comment).

To mock this, I have edited template with odo specific labels. so that, when we trigger serviceInstance, it deploys that template. Template will populate these labels in children resources.

I hope this is clear.

kadel:
> In odo, we inject labels into every resource. so that it helps us for various operations such as odo list for an example.

We don't need to inject labels into every resource.  We can just add our labels to `ServiceInstance` and modify `odo list` to list for `ServiceInstances` created by odo

surajnarwade:
I think that will work. but If I want to see `odo url list`, it will not work :(

kadel:
> I think that will work. but If I want to see odo url list, it will not work :(
> if that template does url creation

that is ok

surajnarwade:
@kadel , @cdrage , @syamgk , @mik-dass Review Needed :)

_(This will take bit of time)_

**Environment needed:**

* minishift version: minishift v1.19.0+1750702

* Enable ansible service broker addon in minishift, follow this: https://github.com/minishift/minishift-addons/tree/master/add-ons/ansible-service-broker
```
minishift addons enable ansible-service-broker
```
* After enabling addons, follow these steps:
```
export MINISHIFT_ENABLE_EXPERIMENTAL=y

minishift config set iso-url centos

minishift start --extra-clusterup-flags "--service-catalog"
```

**Things to test:**

* odo catalog

```
$ odo catalog list
The following components can be deployed:
- dotnet
- httpd
- nginx
- nodejs
- perl
- php
- python
- ruby
- wildfly
- cakephp-mysql-persistent
- rails-pgsql-persistent
- jenkins-pipeline-example
- mongodb-persistent
- mysql-persistent
- dancer-mysql-persistent
- django-psql-persistent
- nodejs-mongo-persistent
- jenkins-persistent
- mariadb-persistent
- postgresql-persistent
```

* odo create

```
$ odo create jenkins-persistent
Component 'jenkins-persistent' was created.

Component 'jenkins-persistent' is now set as active component.
```

* odo list

```
$ odo list
ACTIVE     NAME                   TYPE
*          jenkins-persistent     jenkins-persistent
```

* odo delete

cdrage:
@surajnarwade 

Here are some simpler instructions (as I had troubles re-creating the development environment):


```sh
# Install the minishift addon
wget https://raw.githubusercontent.com/minishift/minishift-addons/master/add-ons/ansible-service-broker/ansible-service-broker.addon
minishift addons install minishift-addons/add-ons/ansible-service-broker
minishift addons enable ansible-service-broker


# Deploy minishift
MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags "--service-catalog" --iso-url centos --openshift-version 3.9.0
```

surajnarwade:
@cdrage , thanks a lot  for simple instructions.

`rails-pgsql-persistent` worked for me  (it took almost 10 minutes)

```
$ oc get all
NAME                                       REVISION   DESIRED   CURRENT   TRIGGERED BY
deploymentconfigs/postgresql               1          1         1         config,image(postgresql:9.5)
deploymentconfigs/rails-pgsql-persistent   1          1         1         config,image(rails-pgsql-persistent:latest)

NAME                                  TYPE      FROM      LATEST
buildconfigs/rails-pgsql-persistent   Source    Git       1

NAME                              TYPE      FROM          STATUS     STARTED          DURATION
builds/rails-pgsql-persistent-1   Source    Git@98001c4   Complete   10 minutes ago   9m42s

NAME                                  DOCKER REPO                                        TAGS      UPDATED
imagestreams/rails-pgsql-persistent   172.30.1.1:5000/myproject/rails-pgsql-persistent   latest    About a minute ago

NAME                            HOST/PORT                                               PATH      SERVICES                 PORT      TERMINATION   WILDCARD
routes/rails-pgsql-persistent   rails-pgsql-persistent-myproject.192.168.42.69.nip.io             rails-pgsql-persistent   <all>                   None

NAME                                READY     STATUS      RESTARTS   AGE
po/postgresql-1-qxxds               1/1       Running     0          10m
po/rails-pgsql-persistent-1-build   0/1       Completed   0          10m
po/rails-pgsql-persistent-1-zswt2   1/1       Running     0          1m

NAME                          DESIRED   CURRENT   READY     AGE
rc/postgresql-1               1         1         1         10m
rc/rails-pgsql-persistent-1   1         1         1         1m

NAME                         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
svc/postgresql               ClusterIP   172.30.93.159   <none>        5432/TCP   10m
svc/rails-pgsql-persistent   ClusterIP   172.30.84.88    <none>        8080/TCP   10m
```

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,64 @@ import (
...
+	finalList["imagestream"] = catalogList
 	// TODO: uncomment when component create supports template creation
-	//clusterServiceClasses, err := client.GetClusterServiceClassExternalNames()
-	//if err != nil {
-	//	return nil, errors.Wrap(err, "unable to get cluster service classes")
-	//}
-	//catalogList = append(catalogList, clusterServiceClasses...)
+	clusterServiceClasses, err := client.GetClusterServiceClassExternalNames()
+	if err != nil {
+		log.Debug("Unable to get cluster serviceClassExternalName")
```
cdrage: I believe you forgot to add an error out here.. At the moment, this prints out the debug and then continues.

pkg/occlient/occlient.go
```diff
@@ -1096,12 +1096,23 @@ func (c *Client) Delete(labels map[string]string) error {
...
 	// PersistentVolumeClaim
-	log.Debugf("Deleting PersistentVolumeClaims")
+	log.Debug("Deleting PersistentVolumeClaims")
 	err = c.kubeClient.CoreV1().PersistentVolumeClaims(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
 	if err != nil {
 		errorList = append(errorList, "unable to delete volume")
 	}
 
+	// Service Instance
+	log.Debug("Deleting Service Instance")
```
cdrage: Add more comments here of what's happening... It's a lot of long lines / code, we should add commenting around htere of what's happening.

pkg/catalog/catalog.go
surajnarwade: I purposefully did that, because even if service catalog is not there, `odo catalog list` should work.
if there is service catalog but it's not listing the services, user can check debug logs

pkg/occlient/occlient.go
surajnarwade: Added :)

pkg/occlient/occlient.go
```diff
@@ -1150,13 +1173,44 @@ func (c *Client) GetBuildConfig(name string, project string) (*buildv1.BuildConf
 // GetClusterServiceClasses queries the service service catalog to get all the
 // currently available cluster service classes
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
+	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{FieldSelector: "spec.clusterServiceBrokerName=template-service-broker"})
```
kadel: Why is this limited to template service broker?

pkg/occlient/occlient.go
```diff
@@ -1150,13 +1173,44 @@ func (c *Client) GetBuildConfig(name string, project string) (*buildv1.BuildConf
...
+			},
+			metav1.ObjectMeta{
+				Finalizers: []string{"kubernetes-incubator/service-catalog"},
+				Name:       componentName,
+				Namespace:  c.namespace,
+				Labels:     labels,
+			},
+			scv1beta1.ServiceInstanceSpec{
+				UserInfo: &scv1beta1.UserInfo{
+					Username: "developer",
```
kadel: hardcoded username?

pkg/occlient/occlient.go
```diff
@@ -1150,13 +1173,44 @@ func (c *Client) GetBuildConfig(name string, project string) (*buildv1.BuildConf
 // GetClusterServiceClasses queries the service service catalog to get all the
 // currently available cluster service classes
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
```
kadel: Thi is is a bit confusing name and commend. It says that you are getting all the cluster services, but  in listing you limit it to just template service broker

pkg/component/component.go
```diff
@@ -94,9 +94,22 @@ func CreateFromPath(client *occlient.Client, name string, ctype string, path str
 	return nil
 }
 
+// CreateFromTemplate create new component from serviceCatalog
+// we are using template service broker only
+func CreateFromTemplate(client *occlient.Client, componentName string, componentType string, applicationName string) error {
```
kadel: This is confusing, commend and function name says that you are creating it from a template, but in fact you just ServiceInstatnce, that is not limited to just template.

pkg/occlient/occlient.go
surajnarwade: as per my discussion with @containscafeine before starting this PR, we were supposed to add templates only and  Since, services from APB (ansible playbook broker) are highly parameterised, as per discussion, we are going with Template service broker only as of now (since it has default values as well). Hence you can see only templates in odo catalog list along with image names.

pkg/occlient/occlient_test.go
```diff
@@ -448,3 +449,62 @@ func Test_parseImageName(t *testing.T) {
...
+				},
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fkclient, fkclientset := FakeNew()
+
+			//_, err := fkclient.CreatePVC(tt.name, tt.size, tt.labels)
```
ashetty1: Remove the comment?

pkg/occlient/occlient_test.go
```diff
@@ -448,3 +449,62 @@ func Test_parseImageName(t *testing.T) {
...
+			// Check for validating actions performed
+			if len(fkclientset.ServiceCatalogClientSet.Actions()) != 1 && tt.wantErr == false {
+				t.Errorf("expected 1 action in CreateServiceInstace got: %v", fkclientset.ServiceCatalogClientSet.Actions())
+			}
+
+			createdServiceInstance := fkclientset.ServiceCatalogClientSet.Actions()[0].(ktesting.CreateAction).GetObject().(*scv1beta1.ServiceInstance)
+			if !reflect.DeepEqual(createdServiceInstance.Labels, tt.args.labels) {
+				t.Errorf("labels in created serviceInstance is not matching expected labels, expected: %v, got: %v", tt.args.labels, createdServiceInstance.Labels)
+			}
+			if !reflect.DeepEqual(createdServiceInstance.Labels, tt.args.labels) {
```
ashetty1: Are you checking for `createdServiceInstance.Labels` twice?

pkg/occlient/occlient_test.go
```diff
@@ -448,3 +449,62 @@ func Test_parseImageName(t *testing.T) {
 		})
 	}
 }
+
+func TestCreateServiceInstance(t *testing.T) {
+	type args struct {
+		componentName   string
+		componentType   string
+		applicationName string
```
ashetty1: Doesn't appear that `applicationName` is being used anywhere.

pkg/occlient/occlient_test.go
```diff
@@ -448,3 +449,62 @@ func Test_parseImageName(t *testing.T) {
...
+			// Checks for error in positive cases
+			if tt.wantErr == false && (err != nil) {
+				t.Errorf(" client.CreateServiceInstance(componentName,componentType, labels) unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			// Check for validating actions performed
+			if len(fkclientset.ServiceCatalogClientSet.Actions()) != 1 && tt.wantErr == false {
+				t.Errorf("expected 1 action in CreateServiceInstace got: %v", fkclientset.ServiceCatalogClientSet.Actions())
+			}
+
```
ashetty1: The checks below would be necessary only if `err` is `nil`. You could add that check here?

cdrage:
LGTM after the changes, awesome work!!

surajnarwade:
@cdrage good to go, waiting for @kadel  :)

kadel:
hmm, It doesn't work for me :-(

I can list stuff, but I can't create anything from the catalog.

```
▶ odo catalog list
The following components can be deployed:
- dotnet
- httpd
- nginx
- nodejs
- perl
- php
- python
- ruby
- wildfly
- jenkins-persistent
- jenkins-pipeline-example
- mongodb-persistent
- mariadb-persistent
- mysql-persistent
- cakephp-mysql-persistent
- dancer-mysql-persistent
- rails-pgsql-persistent
- postgresql-persistent
- django-psql-persistent
- nodejs-mongo-persistent
```
```
▶ odo create mariadb-persistent
imagestreams.image.openshift.io "mariadb-persistent" not found
```

surajnarwade:
@kadel , can you fetch now and test again, there was small spell mistake before

kadel:
> @kadel , can you fetch now and test again, there was small spell mistake before

Stuff like this should be caught by tests. 
Functionality in this PR was completely broken, and tests were green 😢 

We have to have tests that cover such cases.

cdrage:
@kadel I agree on that! Are we able to add more e2e tests to this?

kadel:
```
▶ odo create mariadb-persistent --binary ff
imagestreams.image.openshift.io "mariadb-persistent" not found
```

There should be a better message when users try to provide `--local` `--binary` or `--git` flag for services from the service catalog.

How about: "This component doesn't support `--git`, `--local` or `--binary` flag."

pkg/catalog/catalog.go
```diff
@@ -9,55 +9,63 @@ import (
...
 		}
+
 	}
 
 	return result, nil
 }
 
 // Exists returns true if the given component type is valid, false if not
-func Exists(client *occlient.Client, componentType string) (bool, error) {
+// Exists returns catalog type either `imageStream` or `serviceCatalog`
```
kadel: Please update this comment, it says "Exits returns" twice. 
The message should clearly describe what this function does and that each return value means

pkg/occlient/occlient.go
```diff
@@ -1157,16 +1180,45 @@ func (c *Client) GetBuildConfig(name string, project string) (*buildv1.BuildConf
 	return bc, nil
 }
 
-// GetClusterServiceClasses queries the service service catalog to get all the
-// currently available cluster service classes
+// GetClusterServiceClasses queries the service service catalog to get available clusterServiceClasses
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
+	// TODO: Remove `FieldSelector` from ListOptions when we are confident with Ansible Service Broker
```
kadel: Why are we not confident with ASB?

pkg/component/component.go
```diff
@@ -94,6 +94,19 @@ func CreateFromPath(client *occlient.Client, name string, ctype string, path str
 	return nil
 }
 
+// CreateFromSvcCatalog create new component from serviceCatalog
+func CreateFromSvcCatalog(client *occlient.Client, componentName string, componentType string, applicationName string) error {
+	labels := componentlabels.GetLabels(componentName, applicationName, true)
+	// save component type as label
+	labels[componentlabels.ComponentTypeLabel] = componentType
+	err := client.CreateServiceInstance(componentName, componentType, labels)
```
kadel: `componentName` should be "namespaced" (prefixed with application name) to allow multiple ServiceCatalog components in different odo applications.

pkg/occlient/occlient.go
surajnarwade: because ASB is highly parameterised and issue talks about templates only

surajnarwade: issue #266

surajnarwade:
**Things to test:**

* odo catalog 

```
$ odo service catalog
The following services can be deployed:
- rails-pgsql-persistent
- jenkins-persistent
- jenkins-pipeline-example
- mongodb-persistent
- mysql-persistent
- cakephp-mysql-persistent
- django-psql-persistent
- nodejs-mongo-persistent
- mariadb-persistent
- postgresql-persistent
- dancer-mysql-persistent
```

* provision a service

```
$ odo service create mysql-persistent
Service 'mysql-persistent' was created.
```

* list provisioned services (provisioning takes a bit of time, so I added status term in list to monitor it)

```
$ odo service list
NAME                 TYPE                 STATUS
mysql-persistent     mysql-persistent     ProvisionedSuccessfully
```

* delete the service

```
$ odo service delete mysql-persistent
Are you sure you want to delete mysql-persistent from myapp? [y/N] y
Service mysql-persistent from application myapp has been deleted
```

@kadel , @cdrage @syamgk @mik-dass

cdrage:
@surajnarwade 

```sh
▶ ./odo service                                                                                                                                                                                                                                                                                                                                                             
See 'odo service -h' for more help and examples
```

Help needs improvement!

I get this when I don't have the ansible service broken addon installed:

```sh
github.com/redhat-developer/odo  pr_413 ✔                                                                                                                                                                                                                                                                                                                              85d
▶ ./odo service list                                     
serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list serviceinstances.servicecatalog.k8s.io in the namespace "myproject": User "developer" cannot list serviceinstances.servicecatalog.k8s.io in project "myproject"
```

We need to inform the user that we require the ansible service as well as 3.9.0 of OpenShift or else `odo service` will not work correctly.

surajnarwade:
HI @cdrage , thanks for the review. 

I updated `odo service`. I never faced issue that you mentioned while `odo service list` :(

cdrage:
@surajnarwade that's because you have the ansible broker enabled :) What my description was that *IF* the user does not have the ansible broker / requirements to deploy a service, we should error out and indicate that they require it within their OpenShift cluster

surajnarwade:
@cdrage , I tried with both (enabling and disabling ASB), I was not able to reproduce the error (Maybe some problem with your minishift setup), @mik-dass  @kadel are you facing similar issue ?

pkg/catalog/svcCatalog/catalog.go
```diff
@@ -0,0 +1,134 @@
+package svcCatalog
```
kadel: shouldn't this be in it `pkg/service/`?

pkg/catalog/catalog.go
```diff
@@ -9,22 +9,14 @@ import (
 )
 
 // List lists all the available component types
+// returned map has key: imageStream and serviceCatalog and value is slice of respective components
```
kadel: this command needs to be updated

pkg/catalog/svcCatalog/catalog.go
surajnarwade: oh yeah

pkg/catalog/svcCatalog/catalog.go
```diff
@@ -0,0 +1,134 @@
...
+	Type   string
+	Status string
+}
+
+// ListCatalog lists all the available service types
+func ListCatalog(client *occlient.Client) ([]string, error) {
+
+	clusterServiceClasses, err := client.GetClusterServiceClassExternalNames()
+	if err != nil {
+		log.Debug("Unable to get cluster serviceClassExternalName")
```
kadel: Shouldn't this return an error?

pkg/catalog/svcCatalog/catalog.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// CreateService create new service from serviceCatalog
+func CreateService(client *occlient.Client, serviceName string, serviceType string, applicationName string) error {
+	labels := componentlabels.GetLabels(serviceName, applicationName, true)
+	// save service type as label
+	labels[componentlabels.ComponentTypeLabel] = serviceType
+	err := client.CreateServiceInstance(serviceName, serviceType, labels)
+	if err != nil {
+		return err
```
kadel: wrap?

pkg/catalog/svcCatalog/catalog.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to retrieve list of services")
+	}
+	return nil
+
+}
+
+// List lists all the deployed services
+func List(client *occlient.Client, applicationName string, projectName string) ([]ServiceInfo, error) {
+	applicationSelector := fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
```
kadel: you should use `ConvertLabelsToSelector` https://github.com/redhat-developer/odo/blob/50ef7cda4a41eae5213e174bb83e7f2b0f7d32ce/pkg/util/util.go#L13

pkg/occlient/occlient.go
```diff
@@ -83,6 +83,11 @@ type Client struct {
 	namespace            string
 }
 
+type SvcCat struct {
```
kadel: Please add comment explaining what this struct is

kadel: Shouldn't this be in `service` package?

kadel: `ServiceStatus` might be a better name

pkg/occlient/occlient.go
```diff
@@ -1152,6 +1180,27 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 			}
 		}
 	}
+
+	return values, nil
+}
+
+// GetLabelValues get label values of given label from objects in project that are matching selector
+// returns slice of struct which contains unique label values and status of service
+func (c *Client) GetLabelValuesForCatalog(project string, label string, selector string) ([]SvcCat, error) {
```
kadel: Maybe `GetServicesStatus` might be a better name?

pkg/occlient/occlient.go
surajnarwade: if I put it in service pkg,

```
go build -o odo ./main.go
import cycle not allowed
package main
	imports github.com/redhat-developer/odo/cmd
	imports github.com/redhat-developer/odo/pkg/application
	imports github.com/redhat-developer/odo/pkg/occlient
	imports github.com/redhat-developer/odo/pkg/service
	imports github.com/redhat-developer/odo/pkg/occlient
```

cmd/service.go
```diff
@@ -0,0 +1,180 @@
...
+If service name is not provided, service type value will be used for the name.
+
+A full list of service types that can be deployed is available using: 'odo service catalog'`,
+	Example: `  # Create new mysql-persistent service from service catalog.
+  odo service create mysql-persistent
+	`,
+	Args: cobra.RangeArgs(1, 2),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		applicationName, err := application.GetCurrentOrGetCreateSetDefault(client)
```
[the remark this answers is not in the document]
surajnarwade: Done

cmd/service.go
```diff
@@ -0,0 +1,180 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var serviceForceDeleteFlag bool
+
+// serviceCmd represents the service command
+var serviceCmd = &cobra.Command{
+	Use:   "service",
+	Short: "Perform service catalog operations",
```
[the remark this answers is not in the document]
surajnarwade: added

pkg/service/service.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return services, nil
+}
+
+// GetServiceType returns type of service in given application and project
+func GetServiceType(client *occlient.Client, serviceName string, applicationName string, projectName string) (string, error) {
+
+	// filter according to service and application name
+	selector := fmt.Sprintf("%s=%s,%s=%s", componentlabels.ComponentLabel, serviceName, applabels.ApplicationLabel, applicationName)
```
kadel: you should use `ConvertLabelsToSelector` https://github.com/redhat-developer/odo/blob/50ef7cda4a41eae5213e174bb83e7f2b0f7d32ce/pkg/util/util.go#L13

pkg/service/service.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	applabels "github.com/redhat-developer/odo/pkg/application/labels"
+	componentlabels "github.com/redhat-developer/odo/pkg/component/labels"
+	"github.com/redhat-developer/odo/pkg/occlient"
+)
+
+// ComponentInfo holds all important information about one component
```
kadel: 👎

pkg/service/service.go
```diff
@@ -0,0 +1,133 @@
...
+			return nil, errors.Wrapf(err, "unable to list services")
+		}
+		services = append(services, ServiceInfo{Name: service.Name, Type: ctype, Status: service.Status})
+	}
+
+	return services, nil
+}
+
+// GetServiceType returns type of service in given application and project
+func GetServiceType(client *occlient.Client, serviceName string, applicationName string, projectName string) (string, error) {
```
kadel: If nothing else then at least this function should have tests.

pkg/service/service.go
```diff
@@ -0,0 +1,133 @@
+package service
```
kadel: not a single test for this whole package :-(

kadel:
Next time, please just add new commits, instead of squashing all changes into one. It will make reviews much easier.

pkg/service/service.go
```diff
@@ -0,0 +1,139 @@
...
+	applicationSelector := fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
+	serviceNames, err := client.GetServiceStatus(projectName, componentlabels.ComponentLabel, applicationSelector)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to list services")
+	}
+
+	services := []ServiceInfo{}
+
+	for _, service := range serviceNames {
+		ctype, err := GetServiceType(client, service.Name, applicationName, projectName)
```
kadel: This function has also `GetServiceStatus` in it. That means that we will get kubernetes api will be called 2 times for each service :-(

pkg/occlient/occlient.go
```diff
@@ -1152,6 +1180,27 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 			}
 		}
 	}
+
+	return values, nil
+}
+
+// GetServiceStatus get label values of given label from objects in project that are matching selector
+// returns slice of struct which contains unique label values and status of service
+func (c *Client) GetServiceStatus(project string, label string, selector string) ([]ServiceStatus, error) {
```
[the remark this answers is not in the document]
surajnarwade: I didn't get this, can you please elaborate with example

cmd/service.go
```diff
@@ -0,0 +1,189 @@
...
+  odo service create mysql-persistent
+	`,
+	Args: cobra.RangeArgs(1, 2),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		applicationName, err := application.GetCurrentOrGetCreateSetDefault(client)
+		checkError(err, "")
+		projectName := project.GetCurrent(client)
+		serviceType := args[0]
+		exists, err := svc.SvcTypeExists(client, serviceType)
```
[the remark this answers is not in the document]
surajnarwade: Checked

cmd/service.go
```diff
@@ -0,0 +1,190 @@
...
+	Short: "Lists all the services from service catalog",
+	Long:  "Lists all the services from service catalog",
+	Example: `  # List all services
+  odo service catalog
+	`,
+	Args: cobra.ExactArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		catalogList, err := svc.ListCatalog(client)
+		checkError(err, "unable to list services")
```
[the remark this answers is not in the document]
surajnarwade: updated

pkg/occlient/occlient_test.go
```diff
@@ -1949,3 +1952,172 @@ func TestCreateNewProject(t *testing.T) {
...
+		}
+
+		if err == nil && !tt.wantErr {
+			if (len(fakeClientSet.ServiceCatalogClientSet.Actions()) != 1) && (tt.wantErr != true) {
+				t.Errorf("expected 1 action in ListRoutes got: %v", fakeClientSet.ServiceCatalogClientSet.Actions())
+			}
+		} else if err == nil && tt.wantErr {
+			t.Errorf("test failed, expected: %s, got %s", "false", "true")
+		} else if err != nil && !tt.wantErr {
+			t.Errorf("test failed, expected: %s, got %s", "no error", "error:"+err.Error())
```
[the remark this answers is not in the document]
surajnarwade: Fixed, I copied it from Listroute test and didn't noticed :(

pkg/occlient/occlient.go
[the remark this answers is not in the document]
surajnarwade: I got this, patch coming soon

pkg/service/service.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to retrieve list of services")
+	}
+	return nil
+
+}
+
+// List lists all the deployed services
+func List(client *occlient.Client, applicationName string, projectName string) ([]ServiceInfo, error) {
+	applicationSelector := fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
```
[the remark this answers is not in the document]
surajnarwade: Updated

kadel:
Test this with latest OpenShift:
```
MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --openshift-version v3.10.0 --extra-clusterup-flags "--enable=*,service-catalog,template-service-broker"
```

surajnarwade:
@kadel somehow this is not working on my machine:

```
MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --openshift-version v3.10.0 --extra-clusterup-flags "--enable=*,service-catalog,template-service-broker"
```
@mik-dass @cdrage @anmolbabu , can you please give it a try

kadel:
> @kadel somehow this is not working on my machine:

why is it not working? Any errors?

cdrage:
@surajnarwade `MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --openshift-version v3.10.0 --extra-clusterup-flags "--enable=*,service-catalog,template-service-broker"` works for me too.

Try using a more up-to-date `minishift` binary maybe?

surajnarwade:
@kadel @cdrage , it is working now :) and this PR also works with latest openshift.
(yesterday it was not working somehow)

surajnarwade:
@kadel , I squashed the commits and updated the commit message as well.
@cdrage your review+merge needed :)

cmd/service.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+var serviceForceDeleteFlag bool
+
+// serviceCmd represents the service command
+var serviceCmd = &cobra.Command{
+	Use:   "service",
+	Short: "Perform service catalog operations",
+	Long: ` Perform service catalog operations (Currently we are limiting to template service broker only
+because, Ansible Service Broker is highly parameterised, whereas Template Service broker comes with templates which has default values )`,
```
cdrage: Long: ` Perform service catalog operations.

Limited to template service broker only. )`,

May be better. The long description seems a bit *too* long. No need to explain *why* we don't support it.

cmd/service.go
```diff
@@ -0,0 +1,190 @@
...
+	Example: fmt.Sprintf("%s\n%s\n%s\n%s",
+		serviceCatalogCmd.Example,
+		serviceCreateCmd.Example,
+		serviceDeleteCmd.Example,
+		serviceListCmd.Example),
+	Args: cobra.RangeArgs(1, 3),
+	Run: func(cmd *cobra.Command, args []string) {
+	},
```
cdrage: why blank? should we error out here instead (or just remove Run completely)

cmd/service.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+var serviceCreateCmd = &cobra.Command{
+	Use:   "create <service_type> [service_name]",
+	Short: "Create a new service",
+	Long: `Create a new service from service catalog to deploy on OpenShift.
+
+If service name is not provided, service type value will be used for the name.
```
cdrage: `If service name is not provided, service type value will be used.`

pkg/occlient/occlient.go
```diff
@@ -1148,6 +1147,29 @@ func (c *Client) Delete(labels map[string]string) error {
...
+	}
+	// Iterating over serviceInstance List and deleting one by one
+	for _, svc := range svcCatList.Items {
+		err = c.serviceCatalogClient.ServiceInstances(c.namespace).Delete(svc.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			return errors.Wrap(err, "unable to delete serviceInstace")
+		}
+	}
+	return nil
+}
```
cdrage: The above function seems a bit too clustered. Could you perhaps add some newlines between separate code sections?

pkg/occlient/occlient.go
```diff
@@ -1148,6 +1147,29 @@ func (c *Client) Delete(labels map[string]string) error {
...
+
+// DeleteProject deletes give project
```
cdrage: `deletes a given project`

pkg/occlient/occlient.go
```diff
@@ -1171,9 +1194,20 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 			}
 		}
 	}
+
 	return values, nil
 }
 
+func (c *Client) GetServiceInstanceList(project string, selector string) ([]scv1beta1.ServiceInstance, error) {
```
cdrage: Need comment!

pkg/occlient/occlient.go
```diff
@@ -1184,16 +1218,45 @@ func (c *Client) GetBuildConfigFromName(name string, project string) (*buildv1.B
...
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
+	// TODO: Remove `FieldSelector` from ListOptions when we are confident with Ansible Service Broker
+	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{FieldSelector: "spec.clusterServiceBrokerName=template-service-broker"})
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to list cluster service classes")
 	}
 	return classList.Items, nil
 }
 
+// CreateServiceInstance create service instance from service catalog
```
cdrage: `creates a service instance`

pkg/service/service.go
```diff
@@ -0,0 +1,105 @@
...
+
+	clusterServiceClasses, err := client.GetClusterServiceClassExternalNames()
+	if err != nil {
+		return []string{}, errors.Wrapf(err, "unable to get cluster serviceClassExternalName")
+	}
+
+	return clusterServiceClasses, nil
+}
+
+// CreateService create new service from serviceCatalog
```
cdrage: `creates a new service`

pkg/service/service.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to list services")
+	}
+
+	var services []ServiceInfo
+	for _, elem := range serviceInstanceList {
+		services = append(services, ServiceInfo{Name: elem.Labels[componentlabels.ComponentLabel], Type: elem.Labels[componentlabels.ComponentTypeLabel], Status: elem.Status.Conditions[0].Reason})
+	}
+
+	return services, nil
```
cdrage: above code needs comments explain what they do. 

For the first code block, explain why you're grabbing labels.

For second code block, something like this would suffice:
`// Iterate through serviceInstanceList and add to service`.

cdrage:
Should we change the list formatting to match `odo catalog list`?

```sh
github.com/redhat-developer/odo  pr_413 ✔                                                                                                                                                                                                                                 107d  ⍉
▶ ./odo service catalog
The following services can be deployed:
- mongodb-persistent
- cakephp-mysql-persistent
- dancer-mysql-persistent
- rails-pgsql-persistent
- nodejs-mongo-persistent
- jenkins-persistent
- jenkins-pipeline-example
- mariadb-persistent
- mysql-persistent
- postgresql-persistent
- django-psql-persistent

github.com/redhat-developer/odo  pr_413 ✔                                                                                                                                                                                                                                  107d
▶ ./odo catalog list
NAME        TAGS
dotnet      2.0,latest
httpd       2.4,latest
nodejs      0.10,4,6,8,latest
ruby        2.0,2.2,2.3,2.4,latest
wildfly     10.0,10.1,8.1,9.0,latest

```

surajnarwade:
@cdrage , I think formatting is fine as of now, once we support multiple brokers, then we can have tabular format, where we mention the service and where(broker) it is coming from

cdrage:
@surajnarwade Isn't there different versions of each service though? It makes sense to add what tags are available for the user to use @kadel @surajnarwade

cdrage:
@surajnarwade 
Yeah, it looks like are multiple versions of the databases, but when I went to create it manually through the UI there's no option for version. You're right. This LGTM.

```sh
▶ oc get is -n openshift
NAME             DOCKER REPO                                TAGS                           UPDATED
dotnet           172.30.1.1:5000/openshift/dotnet           2.0,latest                     8 minutes ago
dotnet-runtime   172.30.1.1:5000/openshift/dotnet-runtime   2.0,latest                     8 minutes ago
httpd            172.30.1.1:5000/openshift/httpd            2.4,latest                     8 minutes ago
jenkins          172.30.1.1:5000/openshift/jenkins          1,2,latest                     8 minutes ago
mariadb          172.30.1.1:5000/openshift/mariadb          10.2,latest,10.1               8 minutes ago
mongodb          172.30.1.1:5000/openshift/mongodb          2.4,2.6,3.2 + 2 more...        8 minutes ago
mysql            172.30.1.1:5000/openshift/mysql            latest,5.5,5.6 + 1 more...     8 minutes ago
nginx            172.30.1.1:5000/openshift/nginx            1.10,1.12,1.8 + 1 more...      8 minutes ago
nodejs           172.30.1.1:5000/openshift/nodejs           6,8,latest + 2 more...         8 minutes ago
perl             172.30.1.1:5000/openshift/perl             latest,5.16,5.20 + 1 more...   8 minutes ago
php              172.30.1.1:5000/openshift/php              7.0,7.1,latest + 2 more...     8 minutes ago
postgresql       172.30.1.1:5000/openshift/postgresql       9.6,latest,9.2 + 2 more...     8 minutes ago
python           172.30.1.1:5000/openshift/python           2.7,3.3,3.4 + 3 more...        8 minutes ago
redis            172.30.1.1:5000/openshift/redis            3.2,latest                     8 minutes ago
ruby             172.30.1.1:5000/openshift/ruby             latest,2.0,2.2 + 2 more...     8 minutes ago
wildfly          172.30.1.1:5000/openshift/wildfly          10.1,8.1,9.0 + 2 more...       8 minutes ago
```

jorgemoralespou:
I have some comments to make to this one :-(

> We have added entire new command set for this purpose.
> to check available services in service catalog
> $ odo service catalog

I think this should still be `odo catalog list`and be able to filter, as these things comes from the catalog. In fact, I don't see `catalog` as being a verb, and we use the form `odo <ITEM> <VERB>` so it doesn't seem to comply with the regular command signature.

I would think we should extend the catalog "item" with more specific actions or command line flags, e.g.:

```
$ odo catalog list-services
$ odo catalog list-components
$ odo catalog list --filter=services,components
$ odo catalog
```

Catalog actions should then be expanded to be able to filter, describe usage/show, and the output should also be enhanced to show if the results are services or components (builders), and potentially in the future, this output could be expanded to differentiate services even further.

Apart from that, I assume we can pass the --component and --application flag to the "service" option, as we can with any other option that odo can take, like component, storage,...

kadel:
It makes sense. 
I would go with `odo catalog list-services` and deprecate `odo catalog list` in favor of `odo catalog list-components`

I don't like `odo catalog list --filter=services,components` as it implies that you would be able to run `odo catalog list` that lists both services and components. This would be confusing because we are using a different command to create service and component.

jorgemoralespou:
@kadel I agree, but it seems maybe too verbose/long. Adding @marekjelen and @gshipley for their thoughts.

Eventually, we will need to filter down/search the catalog for when there are many occurrences or we want to find an occurrence with a particular "condition".

surajnarwade:
I think This output will totally look weird `odo catalog list --filter=services,components`

kadel:
At Cabal call, we agreed that we will go with `odo catalog list-services` and `odo catalog list-components`. Let's keep `odo catalog list` for a while, but add a warning about deprecation to it.

marekjelen:
@kadel why the dashes?

kadel:
> @kadel why the dashes?

So it is one command. But you have also a good point. It can very well be like this:

```
odo catalog list services

odo catalog list components
```

`odo catalog list` should not be a valid command, it should show help explaining that you need to specify what you want to list `services` or `components`

cmd/catalog.go
```diff
@@ -21,15 +22,26 @@ var catalogCmd = &cobra.Command{
...
+	Use:   "components",
 	Short: "List all available component types.",
 	Long:  "List all available component types from OpenShift's Image Builder.",
 	Example: `  # Get the supported components
-  odo catalog list
+  odo catalog list components
 
   # Search for a supported component
-  odo search nodejs
+  odo catalog search nodejs
```
kadel: shouldn't the command be `odo catalog search components` ? How Am I going to define if I'm searching for component or service?

pkg/occlient/occlient.go
kadel: @surajnarwade Can you please add a comment explaining why are we using just template service broker?

cmoulliard:
> odo service catalog

We need perhaps a parameter to specify the catalog to be used as different could be installed : Openshift Templates, Ansible Service Broker or Operator SDK Broker

jorgemoralespou:
> We need perhaps a parameter to specify the catalog to be used as different could be installed : Openshift Templates, Ansible Service Broker or Operator SDK Broker
@cmoulliard why would we want that? This is something the developer don't care, and to be honest, no one cares which broker provides the service.

jorgemoralespou:
@kadel, IMHO catalog list and catalog search should either list all by default or have a default behavior in case you don't specify to list components or services

kadel:
> @kadel, IMHO catalog list and catalog search should either list all by default or have a default behavior in case you don't specify to list components or services

My reason for wanting to separate it is that we are using different commands to create services and components. So I would like listing to be also separate. 
We can also just have a column in the output to indicate if it is service or component. 

@jorgemoralespou Is there a situation where the user wouldn't know if what he/she is searching for is component or service?

pkg/occlient/occlient.go
```diff
@@ -1184,16 +1221,45 @@ func (c *Client) GetBuildConfigFromName(name string, project string) (*buildv1.B
 	return bc, nil
 }
 
-// GetClusterServiceClasses queries the service service catalog to get all the
-// currently available cluster service classes
+// GetClusterServiceClasses queries the service service catalog to get available clusterServiceClasses
 func (c *Client) GetClusterServiceClasses() ([]scv1beta1.ClusterServiceClass, error) {
-	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{})
+	// TODO: Remove `FieldSelector` from ListOptions when we are confident with Ansible Service Broker
+	classList, err := c.serviceCatalogClient.ClusterServiceClasses().List(metav1.ListOptions{FieldSelector: "spec.clusterServiceBrokerName=template-service-broker"})
```
cmoulliard: Why do you hard code this value as different brokers could be installed on Openshift and not only Openshift Template Broker but ASB, ... ?

surajnarwade:
@kadel , I updated PR with your suggestion, please check once

pkg/occlient/occlient.go
surajnarwade: Initially, when I tried with ASB, it was behaving weirdly, so I thought of going with TSB as of now

jorgemoralespou:
> @jorgemoralespou Is there a situation where the user wouldn't know if what he/she is searching for is component or service?
My guess is not once he's read the `odo`documentation and he understands what he's doing, but definitely something that can happen. 
We could probably run this through a usability test with a DX team.

pkg/occlient/occlient.go
surajnarwade: even most of the APBs have `clusterserviceplan`, initially I would like to experiment with TSB and then jump into ASB

cmd/service.go
```diff
@@ -0,0 +1,186 @@
...
+			checkError(err, "")
+			fmt.Printf("Service %s from application %s has been deleted\n", serviceName, applicationName)
+
+		} else {
+			fmt.Printf("Aborting deletion of service: %v\n", serviceName)
+		}
+	},
+}
+
+//var serviceCatalogCmd = &cobra.Command{
```
anmolbabu: Why Commented out?

surajnarwade:
Commands to test:

```
* odo catalog list components

* odo catalog list services

* odo catalog search components

* odo catalog search services

* odo service create <service_name>

* odo service delete <service_name>

* odo service list
```

cc @anmolbabu @cdrage @mik-dass

jorgemoralespou:
I would add to the list of tests the ones for a component, as the possible components that can be used are from the catalog.

```
* odo create <component_type> <component_name>

* odo delete <component_name>

* odo list
```

cmd/catalog.go
```diff
@@ -50,18 +62,61 @@ var catalogListCmd = &cobra.Command{
...
+  odo catalog list services
+
+ # Search for a supported services
+  odo catalog search services mysql
+	`,
+	Args: cobra.ExactArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getOcClient()
+		catalogList, err := svc.ListCatalog(client)
+		checkError(err, "unable to list services because Service Catalog is not enabled in your cluster")
```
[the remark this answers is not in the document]
surajnarwade: our user will not understand openshift error (so we have to simplify it)
if there is network error, process will not reach till this stage :P

pkg/occlient/occlient.go
```diff
@@ -1148,6 +1147,31 @@ func (c *Client) Delete(labels map[string]string) error {
...
 
+// DeleteServiceInstance takes labels as a input and based on it, deletes respective service instance
+func (c *Client) DeleteServiceInstance(labels map[string]string) error {
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
+	glog.V(4).Infof("Selectors used for deletion: %s", selector)
+	// Service Instance
+	glog.V(4).Infof("Deleting Service Instance")
+	// Listing out serviceInstance because `DeleteCollection` method don't work on serviceInstance
+	svcCatList, err := c.serviceCatalogClient.ServiceInstances(c.namespace).List(metav1.ListOptions{LabelSelector: selector})
```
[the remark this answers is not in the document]
surajnarwade: done :+1:

pkg/occlient/occlient.go
```diff
@@ -1171,9 +1196,21 @@ func (c *Client) GetLabelValues(project string, label string, selector string) (
 			}
 		}
 	}
+
 	return values, nil
 }
 
+// GetServiceInstanceList returns list service instances
+func (c *Client) GetServiceInstanceList(project string, selector string) ([]scv1beta1.ServiceInstance, error) {
```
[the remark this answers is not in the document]
surajnarwade: this is fine, I guess

pkg/service/service.go
surajnarwade: added

pkg/occlient/occlient.go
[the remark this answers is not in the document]
surajnarwade: updated

cmd/catalog.go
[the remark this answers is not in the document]
anmolbabu: Hmm so the occlient creation would fail is what you are saying right? So can there be any other error like permissions/authorisation for using service catalog issue ? 

If not please ignore this comment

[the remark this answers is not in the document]
surajnarwade: I think this error message is okay for this situation

jorgemoralespou:
`odo service create mysql-persistent` without the required parameters should yield an error, and user would need to provide them.

Typically every service will require some configuration, so user will need to `odo service describe <service_name>` to get the parameter configuration so that the user can create the service.

In reality this will get horribly complex because:
- When creating a service, one might need to select a plan
- When creating a service, externalName is not what defines the service id to use, but it's just a text that can be repeated multiple times.

This task it's turning more complex than initially envisioned for the way how servicecatalog API works. I would defer the work or think that we should revisit this in the near future with many changes.
cc/ @kadel

surajnarwade:
@jorgemoralespou ,

pkg/service/service.go
surajnarwade: tests already added in `occlient_test.go`

cmoulliard:
> will get horribly complex because

I agree with you. The flow requires that we do :  

- Search about a service and plan
- Extract from the plan the parameters and customize them (-> under MANIFEST or application.yml file maybe)
- (out of scope of odo) Select the correct maven dependency to add. For spring Boot -> starter, cloud connector able to process data of the service
- Create a service instance using parameters customized
- Generate a Secret with the parameters (= binding)
- Update the DeploymentConfig to add the secret as EnvFrom
- Rollout the DeploymentConfig

cmoulliard:
> odo service create -p param1=value1 -p param2=value2

this command will work for service containing string's fields but when json or yaml content is required (as this is the case for the keycloak's [service](https://github.com/snowdrop/k8s-supervisor/issues/54)), then such info should be defined within a config's application file (MANIFEST, odo-application.yaml)

surajnarwade:
@jorgemoralespou  @cmoulliard @cdrage , if you all agree, then @anmolbabu can merge this PR 
(since we got 3 LGTMs :D )

jorgemoralespou:
@surajnarwade I can't vote as I can't read the implementation. To me looks good and as part of the discussion we can create follow up issues.

anmolbabu:
Merging based on previous acks but @surajnarwade please log issue to add documentation and e2e tests

anmolbabu:
As per our discussion, I have also raised https://github.com/redhat-developer/odo/issues/678 please take it up on priority @surajnarwade

pkg/service/service.go
```diff
@@ -0,0 +1,127 @@
...
+
+// List lists all the deployed services
+func List(client *occlient.Client, applicationName string, projectName string) ([]ServiceInfo, error) {
+	labels := map[string]string{
+		applabels.ApplicationLabel: applicationName,
+	}
+	//since, service is associated with application, it consist of application label as well
+	// which we can give as a selector
+	applicationSelector := util.ConvertLabelsToSelector(labels)
+	// get service instance list based on given selector
```
cdrage: add newline / spacing

Issue #416, opened 2018-04-24
Modify `component.Exists` function


cdrage:
Using this issue so I can track a few things that I've been coming across within the code.

The component.Exists function should have applicationName first and then componentName:

```sh
exists, err := component.Exists(client, componentName, applicationName, projectName)
```

Should actually be:

```sh
exists, err := component.Exists(client, applicationName, componentName, projectName)
```

since all other odo functions are in that order (app name and then component name)

kadel:
almost every function in `component` actually follows the rule where the component name is first, and the application name is after that.

```
func Delete(client *occlient.Client, name string, applicationName string, projectName string) 
func GetComponentDesc(client *occlient.Client, cmpnt string, currentApplication string, currentProject string)
func GetComponentType(client *occlient.Client, componentName string, applicationName string, projectName string)
func SetCurrent(client *occlient.Client, name string, applicationName string, projectName string)

```

cdrage:
@kadel Heck, you're right. 

Should we go with component name first then application name then?

kadel:
> Should we go with the component name first then application name then?

yes. I would standardized on the following order: component, application, project.

kadel:
Is this issue still valid? Do we need to fix it this somewhere? Or we can close it?

kadel:
ping @cdrage

cdrage:
We can close this, thanks!

PR #417, opened 2018-04-24
Modify Exists function

cdrage:
Modifies the Exists function so that it's `applicationName,
componentName` similar to other functions within the code-base.


cdrage:
Closes: https://github.com/redhat-developer/odo/issues/416

kadel:
all function in the component package has component name before application. check my comment in #416

PR #423, opened 2018-04-27
Improve odo create command output

cdrage:
Improves the logging a little bit / output so it's clearer to the user.

Closes: https://github.com/redhat-developer/odo/issues/56

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -85,7 +85,9 @@ A full list of component types that can be deployed is available using: 'odo com
 		if len(args) == 2 {
 			componentName = args[1]
 		}
-
+		//validate component name
+		err = validateName(componentName)
+		checkError(err, "")
 		exists, err = component.Exists(client, componentName, applicationName, projectName)
 		checkError(err, "")
 		if exists {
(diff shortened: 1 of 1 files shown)


cmd/create.go
```diff
@@ -120,27 +128,29 @@ A full list of component types that can be deployed is available using: 'odo com
...
 			checkError(err, fmt.Sprintf("failed to push component: %v", componentName))
 
-			fmt.Printf("Component '%s' was created.\n", componentName)
-
 		} else {
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs("./")
 			checkError(err, "")
+
+			fmt.Printf("Building component %s from directory %s", componentName, dir)
```
surajnarwade: add `\n` at the end of line

surajnarwade:
rather than one small comment, LGTM

cmd/create.go
```diff
@@ -93,22 +93,30 @@ A full list of component types that can be deployed is available using: 'odo com
...
 			err = component.RebuildGit(client, componentName)
 			checkError(err, "")
+
 		} else if len(componentLocal) != 0 {
+
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs(componentLocal)
 			checkError(err, "")
+
+			fmt.Printf("Building component %s from local directory %s", componentName, dir)
```
kadel: missing `\n` 
there should be proabbly `\n\n` to make it same as in `--git` component

cmd/create.go
```diff
@@ -120,27 +128,29 @@ A full list of component types that can be deployed is available using: 'odo com
 				os.Exit(1)
 			}
 
+			fmt.Printf("Building component %s from binary %s", componentName, path)
```
kadel: missing `\n\n`

cmd/create.go
kadel: Wouldn't it be better without world "component"? Don't know, just something to think about ;-)

cmd/create.go
```diff
@@ -93,22 +93,30 @@ A full list of component types that can be deployed is available using: 'odo com
 			os.Exit(1)
 		}
 
+		fmt.Printf("Creating component %s from build type %s\n", componentName, componentType)
```
kadel: How about "Creating %s component called %s" ?
So it the resulting string would be something like this:

Creating php component called frontend.
or
Creating nodejs component called nodejs.

cdrage:
@kadel Updated!

cmd/create.go
```diff
@@ -120,27 +128,29 @@ A full list of component types that can be deployed is available using: 'odo com
 				os.Exit(1)
 			}
 
+			fmt.Printf("Building component %s from binary %s\n\n", componentName, path)
```
kadel: This message is different than the similar one for local component (line 112).
We should align it together with the message in line 144.

cdrage: Bit confused here... Lines 112 and 144 are the same / similar?

fmt.Printf("Building component %s from directory %s\n", componentName, dir)
fmt.Printf("Building component %s from binary %s\n\n", componentName, path)

cdrage:
I added an extra `\n` but I'm a little confused what you mean by changing the lines? Both 144 and 112 are the same / very similar.

cmd/create.go
kadel: message on line`112` for the local component doesn't have word "component" in it.
But the others (for binary and local) have. I think they should all have the same wording.

kadel:
fmt.Printf("Building component %s from Git repository: %s\n\n", componentName, componentGit)
fmt.Printf("Building %s from local directory %s\n", componentName, dir)

kadel:
sorry for an accidental close, wrong button 😳

cdrage:
@kadel I see! Updated :+1:

Issue #52, opened 2018-02-07
component create should fail before if directory does not exist


containscafeine:
```console
$ ocdev component create nodejs --dir ./nodejs-ex                                                      
--> Found image cbb7eed (5 days old) in image stream "openshift/nodejs" under tag "6" for "nodejs"                             

    Node.js 6                  
    ---------                  
    Node.js 6 available as docker container is a base platform for building and running various Node.js 6 applications and frameworks. Node.js is a platform built on Chrome's JavaScript runtime for easily building fast, scalable network applications. Node.js uses an event-driven, non-blocking I/O model that makes it lightweight and efficient, perfect for data-intensive real-time applications that run across distributed devices.

    Tags: builder, nodejs, nodejs6                             

    * A source build using binary input will be created        
      * The resulting image will be pushed to image stream "nodejs:latest"                                                     
      * A binary build was created, use 'start-build --from-dir' to trigger a new build                                        
    * This image will be deployed in deployment config "nodejs"                                                                
    * Port 8080/tcp will be load balanced by service "nodejs"  
      * Other containers can access this service through the hostname "nodejs"                                                 

--> Creating resources with label app.ocdev.developers.redhat.com=app,component.ocdev.developers.redhat.com=nodejs ...         
    imagestream "nodejs" created                               
    buildconfig "nodejs" created                               
    deploymentconfig "nodejs" created                          
    service "nodejs" created   
--> Success                    
    Build scheduled, use 'oc logs -f bc/nodejs' to track its progress.                                                         
    Application is not exposed. You can expose services to the outside world by executing one or more of the commands below:   
     'oc expose svc/nodejs'    
    Run 'oc status' to view your app.                          

please wait, building application...                           
failed to execute oc command   
 error: stat /home/concaf/ocdev/src/github.com/redhat-developer/ocdev/nodejs-ex: no such file or directory                    
```

kadel:
related to https://github.com/redhat-developer/ocdev/issues/192

cdrage:
Yeah, looks like this has yet to be implemented, I'll assign this to myself.

```sh
▶ odo create nodejs foobar --local alksdjflasjdfkljsakdljl
Component 'foobar' was created.
To push source code to the component run 'odo push'

Component 'foobar' is now set as active component.
```

kadel:
Yep, closing this. addressed in #425

PR #425, opened 2018-04-30
Adds checks for checking that a directory or a file exists

mik-dass:
Issue : #52

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -104,6 +104,12 @@ A full list of component types that can be deployed is available using: 'odo com
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs(componentLocal)
 			checkError(err, "")
+			fileInfo, err := os.Stat(dir)
+			checkError(err, "")
+			if !fileInfo.IsDir() {
+				fmt.Println("Please provide a path to the directory")
+				os.Exit(1)
+			}
 			err = component.CreateFromPath(client, componentName, componentType, dir, applicationName, "local")
 			checkError(err, "")
 			fmt.Printf("Please wait, creating %s component ...\n", componentName)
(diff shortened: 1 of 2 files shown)


cmd/create.go
```diff
@@ -104,6 +104,12 @@ A full list of component types that can be deployed is available using: 'odo com
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs(componentLocal)
 			checkError(err, "")
+			fi, err := os.Stat(dir)
```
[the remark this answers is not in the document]
mik-dass: ok :+1:

cmd/update.go
```diff
@@ -90,12 +90,24 @@ var updateCmd = &cobra.Command{
...
+				fmt.Println("Please provide path to a directory")
+				os.Exit(1)
+			}
 			err = component.Update(client, componentName, "local", dir)
 			checkError(err, "")
 			fmt.Printf("The component %s was updated successfully\n", componentName)
 		} else if len(componentBinary) != 0 {
 			path, err := filepath.Abs(componentBinary)
 			checkError(err, "")
+			fileInfo, err := os.Stat(path)
```
kadel: It should be possible to create binary component even if binary file doesn't exist #436

mik-dass: What about create.go then?

kadel: for create.go the same. I actually wanted to post my comment for create.go.

for the local component, we can verify that the directory exists.  But for binary component, it is ok, if the binary file doesn't exist at the time when `odo create` or `odo update` is run. that check should be done only for `odo push`

mik-dass: OK :+1:

mik-dass: @kadel What about the isDir thing? Should that be removed too?

kadel:
ping @cdrage

cmd/update.go
```diff
@@ -90,6 +90,12 @@ var updateCmd = &cobra.Command{
 			// we want to use and save absolute path for component
 			dir, err := filepath.Abs(componentLocal)
 			checkError(err, "")
+			fileInfo, err := os.Stat(dir)
+			checkError(err, "")
+			if !fileInfo.IsDir() {
+				fmt.Println("Please provide path to a directory")
```
cdrage: It should be "Please provide a path to the directory"

mik-dass: @cdrage Done

cdrage:
LGTM

PR #429, opened 2018-05-02
Adds DeploymentConfig log functionality to Odo

cdrage:
This adds a new command `odo log` which will display the
DeploymentConfig logs for that particular component.

For example:

```sh
github.com/redhat-developer/odo  add-logs ✔                                                                                                                                                                                                                                                                                                                         15h55m
▶ ./odo list
ACTIVE     NAME        TYPE
*          nodejs3     nodejs

github.com/redhat-developer/odo  add-logs ✔                                                                                                                                                                                                                                                                                                                         15h57m
▶ ./odo log
Environment:
        DEV_MODE=false
        NODE_ENV=production
        DEBUG_PORT=5858
Launching via npm...
npm info it worked if it ends with ok
npm info using npm@3.10.9
npm info using node@v6.11.3
npm info lifecycle nodejs-ex@0.0.1~prestart: nodejs-ex@0.0.1
npm info lifecycle nodejs-ex@0.0.1~start: nodejs-ex@0.0.1

> nodejs-ex@0.0.1 start /opt/app-root/src
> node server.js

Server running on http://0.0.0.0:8080
```

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -1,0 +1,52 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/redhat-developer/odo/pkg/application"
+	"github.com/redhat-developer/odo/pkg/component"
+	"github.com/redhat-developer/odo/pkg/project"
+	"github.com/spf13/cobra"
+)
+
+var logCmd = &cobra.Command{
+	Use:   "log [component_name]",
+	Short: "Retrieve the log for the given component.",
+	Long:  `Retrieve the log for the given component.`,
+	Example: `  # Get the logs for the nodejs component,
+  odo log nodejs
+	`,
+	Args: cobra.RangeArgs(0, 1),
+	Run: func(cmd *cobra.Command, args []string) {
+
+		// Retrieve stdout / io.Writer
+		stdout := os.Stdout
+
+		// Retrieve the client
+		client := getOcClient()
+
+		// Application
+		currentApplication, err := application.GetCurrent(client)
+		checkError(err, "")
+
+		// Project
+		currentProject := project.GetCurrent(client)
+
+		var argComponent string
+
+		if len(args) == 1 {
+			argComponent = args[0]
+		}
+
+		// Retrieve and set the currentComponent
+		currentComponent := getComponent(client, argComponent, currentApplication, currentProject)
+
+		err = component.GetLogs(client, currentComponent, stdout)
+		checkError(err, "Unable to retrieve logs, does your component exist?")
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(logCmd)
+
+}
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -446,3 +446,15 @@ func GetComponentDesc(client *occlient.Client, currentComponent string, currentA
 
 	return componentType, path, componentURL, appStore, nil
 }
+
+// Get Component logs
+func GetLogs(client *occlient.Client, applicationName string, stdout io.Writer) error {
+
+	// Retrieve the logs
+	err := client.DisplayDeploymentConfigLog(applicationName, stdout)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1,6 +1,7 @@
 package occlient
 
 import (
+	"bytes"
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
@@ -1047,6 +1048,43 @@ func (c *Client) FollowBuildLog(buildName string) error {
 
 	if _, err = io.Copy(stdout, rd); err != nil {
 		return errors.Wrapf(err, "error streaming logs for %s", buildName)
+	}
+
+	return nil
+}
+
+// Display DeploymentConfig log to stdout
+func (c *Client) DisplayDeploymentConfigLog(deploymentConfigName string, stdout io.Writer) error {
+
+	// RESTClient call to OpenShift
+	rd, err := c.appsClient.RESTClient().Get().
+		Namespace(c.namespace).
+		Name(deploymentConfigName).
+		Resource("deploymentconfigs").
+		SubResource("log").Stream()
+
+	if rd == nil {
+		return errors.New("unable to retrieve DeploymentConfig from OpenShift, does your component exist?")
+	}
+
+	defer rd.Close()
+
+	if err != nil {
+		return errors.Wrapf(err, "unable get deploymentconfigs log %s", deploymentConfigName)
+	}
+
+	// Copy to buffer (we aren't going to be streaming the logs..)
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, rd)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to copy stream to buffer")
+	}
+
+	// Copy to stdout (in yellow)
+	color.Set(color.FgYellow)
+	defer color.Unset()
+	if _, err = io.Copy(stdout, buf); err != nil {
+		return errors.Wrapf(err, "error copying logs to stdout")
 	}
 
 	return nil
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -226,6 +226,12 @@ var _ = Describe("odoe2e", func() {
 				cmpSet := runCmd("odo component set nodejs")
 				Expect(cmpSet).To(ContainSubstring("nodejs"))
 			})
+
+			It("should be able to retrieve logs", func() {
+				runCmd("odo log")
+				runCmd("odo log nodejs")
+			})
+
 		})
 	})
 


cdrage:
Quite simple (only retrieving DeploymentConfig logs).

But I believe it's a step in the right direction.

What do you think @kadel ?

```sh
github.com/redhat-developer/odo  add-logs ✔                                                                                                                                                                                                                                                                                                                         15h55m  
▶ ./odo list
ACTIVE     NAME        TYPE
*          nodejs3     nodejs

github.com/redhat-developer/odo  add-logs ✔                                                                                                                                                                                                                                                                                                                         15h57m  
▶ ./odo log
Environment: 
        DEV_MODE=false
        NODE_ENV=production
        DEBUG_PORT=5858
Launching via npm...
npm info it worked if it ends with ok
npm info using npm@3.10.9
npm info using node@v6.11.3
npm info lifecycle nodejs-ex@0.0.1~prestart: nodejs-ex@0.0.1
npm info lifecycle nodejs-ex@0.0.1~start: nodejs-ex@0.0.1

> nodejs-ex@0.0.1 start /opt/app-root/src
> node server.js

Server running on http://0.0.0.0:8080
```

pkg/occlient/occlient.go
```diff
@@ -791,6 +795,33 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+	defer rd.Close()
+
+	// Create a new buffer so we can display the entire stream
+	buf := new(bytes.Buffer)
+
+	// Copy to buffer
+	_, err = io.Copy(buf, rd)
+
+	// Print
+	fmt.Println(buf.String())
```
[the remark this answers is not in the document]
cdrage: Okay, no worries. Should we change how to do https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L1048 as well?

cdrage:
Most changes corrected. Should we change https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L1048 as well @kadel with regards to io.Writer? I'll implement it by passing in io.Writer from higher-up in the code.

pkg/occlient/occlient.go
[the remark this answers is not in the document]
kadel: We should, but let's do that in another PR.

cmd/log.go
```diff
@@ -0,0 +1,58 @@
...
+		// Application
+		currentApplication, err := application.GetCurrent(client)
+		checkError(err, "")
+
+		// Project
+		currentProject := project.GetCurrent(client)
+
+		// Retrieve and set the current component if none passed
+		var currentComponent string
+		if len(args) == 0 {
```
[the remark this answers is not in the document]
kadel: yes, but if `inputComponent` is empty than it returns the current component, otherwise it checks if inputComponent exists.

This is exactly what you are doing here with `args[0]` no?

wouldn't this have the same result?
```go
currentComponent := ""
if len(args) == 1 {
  currentComponent = args[1]
}
currentComponent = getComponent(client, currentComponent, currentApplication, currentProject)

 

```

pkg/occlient/occlient.go
```diff
@@ -1020,7 +1021,7 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 }
 
 // FollowBuildLog stream build log to stdout
-func (c *Client) FollowBuildLog(buildName string) error {
+func (c *Client) FollowBuildLog(name string) error {
```
kadel: why was this changed from `buildName` to just `name`?

pkg/occlient/occlient.go
```diff
@@ -1046,8 +1047,35 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
 	if _, err = io.Copy(stdout, rd); err != nil {
-		return errors.Wrapf(err, "error streaming logs for %s", buildName)
+		return errors.Wrapf(err, "error streaming logs for %s", name)
+	}
+
+	return nil
+}
+
+// Display DeploymentConfig log to stdout
+func (c *Client) DisplayDeploymentConfigLog(buildName string) error {
```
kadel: the argument is called `buildName`  but this function accepts DeploymentConfig name as an argument

kadel:
> Most changes corrected. Should we change https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L1048 as well @kadel with regards to io.Writer? I'll implement it by passing in io.Writer from higher-up in the code.

there already is an issue for that ;-) https://github.com/redhat-developer/odo/issues/312

cmd/log.go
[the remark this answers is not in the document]
cdrage: @kadel We're passing in blank for currentComponent.. which feels as if it's a undocumented function since it will return the active component rather than "getting" the component that's being passed in (currentComponent). Does that make sense?

cdrage:
@kadel Updated! I pass in an `io.Writer` now. I retrieve this using `color.Output` in order for us to get coloured output in Windows / macOS / Linux.

I have updated the function as well to write to io.Writer using `io.Copy`. This is all in yellow as well in order to differentiate and indicate to the user that this is raw OpenShift logs.

cdrage:
Let me know what you think regarding `getComponent` and I'll change it to your suggestion.

cmd/log.go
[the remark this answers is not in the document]
kadel: It is documented in a comment for that function.

I agree that it feels weird, but we have this logic in a lot of places, and it actually makes sense to have this abstracted away in one common function. Maybe a different name or small refactoring of that function might help.

pkg/component/component.go
```diff
@@ -446,3 +446,17 @@ func GetComponentDesc(client *occlient.Client, currentComponent string, currentA
 
 	return componentType, path, componentURL, appStore, nil
 }
+
+// Get Component logs
+func GetLogs(client *occlient.Client, applicationName string) (string, error) {
```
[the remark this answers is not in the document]
cdrage: So for io.Writer, should we have this functionality higher-up, in the `cmd` package instead? At the moment, we have the "color" functionlity (using `color.Output`) within component.go rather than up in `cmd`. So perhaps I should move both.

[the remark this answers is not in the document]
kadel: Not sure if I understand that :-(

There are functions that will print output in two colors.
For example `component.Watch`.
It is called from `cmd` package and it takes `io.Writer` as argument.
It writes to `io.Writer` messages in white and also in yellow. So the color can't be set in `cmd` package.

cdrage:
@kadel

I've done the following:

 - Added `os.Stdout` as a variable within `cmd/log.go` so we grab `io.Writer` at the most-highest abstraction level
 - Added `io.Writer` as parameters to both `GetLogs` as well as `DisplayDeploymentConfigLog`

cmd/log.go
```diff
@@ -0,0 +1,62 @@
...
+
+			// Get the current compoment
+			currentComponent = getComponent(client, "", currentApplication, currentProject)
+
+		} else {
+
+			currentComponent = args[0]
```
kadel: we don't need this whole `else` section. This is already taken care of in `getComponent` function

cdrage:
I've also gone ahead and added a small integration test.

cmd/log.go
```diff
@@ -0,0 +1,62 @@
...
+		currentProject := project.GetCurrent(client)
+
+		// Retrieve and set the current component if none passed
+		var currentComponent string
+		if len(args) == 0 {
+
+			// Get the current compoment
+			currentComponent = getComponent(client, "", currentApplication, currentProject)
+
+		} else {
```
kadel: this whole else section is not needed. This logic is already in `getComponent`.

The whole point of using `getComponent` is that you can replace this entire `if else` with  following 5 lines 

```go
currentComponent := ""
if len(args) == 1 {
  currentComponent = args[0]
}
currentComponent = getComponent(client, currentComponent, currentApplication, currentProject)

``

File: pkg/component/component.go [hunk 8]
```diff
@@ -446,3 +446,17 @@ func GetComponentDesc(client *occlient.Client, currentComponent string, currentA
 
 	return componentType, path, componentURL, appStore, nil
 }
+
+// Get Component logs
+func GetLogs(client *occlient.Client, applicationName string, stdout io.Writer) (string, error) {
+
+	// Retrieve the stdout pipe to pass to occlient
```
kadel: is this comment still valid?

cdrage:
@kadel Updated the comment :+1:

tests/e2e/e2e_test.go
```diff
@@ -226,6 +226,12 @@ var _ = Describe("odoe2e", func() {
 				cmpSet := runCmd("odo component set nodejs")
 				Expect(cmpSet).To(ContainSubstring("nodejs"))
 			})
+
+			It("should be able to retrieve logs", func() {
+				runCmd("odo log")
+				runCmd("odo nodejs")
```
kadel: `odo nodejs` what is this command?

tests/e2e/e2e_test.go
```diff
@@ -226,6 +226,12 @@ var _ = Describe("odoe2e", func() {
 				cmpSet := runCmd("odo component set nodejs")
 				Expect(cmpSet).To(ContainSubstring("nodejs"))
 			})
+
+			It("should be able to retrieve logs", func() {
+				runCmd("odo log")
```
kadel: this should probably test if something was returned

cmd/log.go
```diff
@@ -0,0 +1,46 @@
...
+
+		// Application
+		currentApplication, err := application.GetCurrent(client)
+		checkError(err, "")
+
+		// Project
+		currentProject := project.GetCurrent(client)
+
+		// Retrieve and set the currentComponent
+		currentComponent := getComponent(client, args[0], currentApplication, currentProject)
```
kadel: It is failing here if no argument is passed to `odo log` command.

If there is no argument, you have to pass an empty string as `inputComponent` (second) parameter and `getComponent` will return the name of the current component. If `inputComponent`  parameter is non-empty string it will check that `inputComponent`  exists and return it back.

pkg/component/component.go
```diff
@@ -446,3 +446,15 @@ func GetComponentDesc(client *occlient.Client, currentComponent string, currentA
 
 	return componentType, path, componentURL, appStore, nil
 }
+
+// Get Component logs
+func GetLogs(client *occlient.Client, applicationName string, stdout io.Writer) (string, error) {
```
kadel: We don't need to return `string`.

kadel:
There is something wrong with output color, I see some lines in yellow some in white.

screen shot 2018-05-15 at 10 52 36

kadel:
I found out what is happening with the color output.
Looks like wildfly is also using colored output (prints WARN messages in yellow) and it resets our settings.

cdrage:
Updated!

Changes:

 - Fixed tests (tests check for exit codes anyways)
 - I found a nil error bug if there are *no* components that exist at all on OpenShift. I added this here: https://github.com/redhat-developer/odo/pull/429/files#diff-dace9c973da6801d7504aca65cf4da57R45 and within occlient here: https://github.com/redhat-developer/odo/pull/429/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR1066)
 - Check to see if len(args) == 1 when picking a component (https://github.com/redhat-developer/odo/pull/429/files#diff-dace9c973da6801d7504aca65cf4da57R37)

pkg/occlient/occlient.go
```diff
@@ -1052,6 +1053,43 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+
+	// RESTClient call to OpenShift
+	rd, err := c.appsClient.RESTClient().Get().
+		Namespace(c.namespace).
+		Name(deploymentConfigName).
+		Resource("deploymentconfigs").
+		SubResource("log").Stream()
+
+	if rd == nil {
+		return errors.New("Unable to retrieve DeploymentConfig from OpenShift, does your component exist?")
```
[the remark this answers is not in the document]
kadel: We already have nice user-facing message in `cmd`.

cdrage:
@kadel got rid of the Uppercase! :+1:

PR #448, opened 2018-05-14
Porting `GetLabelValues` & `DeleteProject` function to client-go

### LINKED ISSUE #444: port remaining functions from occlient to client-go
We still have a couple of functions that are calling `oc` binary.
 - `Delete`
 - `DeleteProject`
 - `GetLabelValues`

parent: #257 

note:
There two more (`RsyncPath`, `CopyFile`), but those will be a little bit more tricky so there is a sepparate issue for them https://github.com/redhat-developer/odo/issues/443

surajnarwade:
Partially fixes #444, This PR will port `GetLabelValues` & `DeleteProject`
to client-go. so that it will remove dependency of `oc` binary.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1086,9 +1086,7 @@ func (c *Client) Delete(kind string, name string, labels map[string]string) (str
 }
 
 func (c *Client) DeleteProject(name string) error {
-	_, err := c.runOcComamnd(&OcCommand{
-		args: []string{"delete", "project", name},
-	})
+	err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})
 	if err != nil {
 		return errors.Wrap(err, "unable to delete project")
 	}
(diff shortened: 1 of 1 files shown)


pkg/occlient/occlient.go
```diff
@@ -1098,38 +1096,15 @@ func (c *Client) DeleteProject(name string) error {
...
-		"--namespace", project,
-		"-o", "go-template={{range .items}}{{range $key, $value := .metadata.labels}}{{if eq $key \"" + label + "\"}}{{$value}},{{end}}{{end}}{{end}}",
-	}
-
-	output, err := c.runOcComamnd(&OcCommand{args: args})
-	if err != nil {
-		return nil, err
-	}
-
+	appLabel, _ := c.appsClient.DeploymentConfigs(project).List(metav1.ListOptions{LabelSelector: selector})
```
kadel: check for error!

kadel: why is variable called `appLabel`? When it is containing List of DeploymentConfigs.

PR #455, opened 2018-05-15
Porting `Delete` function to client-go

### LINKED ISSUE #444: port remaining functions from occlient to client-go

surajnarwade:
Partially fixes #444, This PR will port Delete function to client-go, so that it will remove dependency of `oc` binary
to client-go. so that it will remove dependency of oc binary.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1089,36 +1089,60 @@ func (c *Client) DisplayDeploymentConfigLog(deploymentConfigName string, stdout 
 	return nil
 }
 
-// Delete calls oc delete
-// kind is always required (can be set to 'all')
-// name can be omitted if labels are set, in that case set name to ''
-// if you want to delete object just by its name set labels to nil
-func (c *Client) Delete(kind string, name string, labels map[string]string) (string, error) {
+// Delete takes labels as a input and based on it, deletes respective resource
+func (c *Client) Delete(labels map[string]string) error {
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
+	log.Debugf("Selectors used for deletion: %s", selector)
 
-	args := []string{
-		"delete",
-		kind,
-	}
-
-	if len(name) > 0 {
-		args = append(args, name)
-	}
-
-	if labels != nil {
-		var labelsString []string
-		for key, value := range labels {
-			labelsString = append(labelsString, fmt.Sprintf("%s=%s", key, value))
-		}
-		args = append(args, "--selector")
-		args = append(args, strings.Join(labelsString, ","))
-	}
-
-	output, err := c.runOcComamnd(&OcCommand{args: args})
+	var errorList []string
+	// Delete DeploymentConfig
+	log.Debug("Deleting DeploymentConfigs")
+	err := c.appsClient.DeploymentConfigs(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
 	if err != nil {
-		return "", err
+		errorList = append(errorList, "unable to delete deploymentconfig")
+	}
+	// Delete Route
+	log.Debug("Deleting Routes")
+	err = c.routeClient.Routes(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		errorList = append(errorList, "unable to delete route")
+	}
+	// Delete BuildConfig
+	log.Debug("Deleting BuildConfigs")
+	err = c.buildClient.BuildConfigs(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		errorList = append(errorList, "unable to delete buildconfig")
+	}
+	// Delete ImageStream
+	log.Debug("Deleting ImageStreams")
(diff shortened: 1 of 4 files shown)


pkg/occlient/occlient.go
```diff
@@ -1056,32 +1056,50 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+	if err != nil {
+		return errors.Wrap(err, "unable to delete build")
+	}
+	// Delete ImageStream
+	err = c.imageClient.ImageStreams(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return errors.Wrap(err, "unable to delete imagestream")
+	}
+	// Delete Services
+	svcList, err := c.kubeClient.CoreV1().Services(c.namespace).List(metav1.ListOptions{LabelSelector: selector})
```
kadel: Do we know why it is not possible to use `DeleteCollection` for Services?

pkg/occlient/occlient.go
```diff
@@ -1056,32 +1056,50 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
 
-	args := []string{
-		"delete",
-		kind,
+	// Delete DeploymentConfig
+	err := c.appsClient.DeploymentConfigs(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return errors.Wrap(err, "unable to delete deploymentconfig")
```
kadel: If for some reason deletion of one Kind fails it should still continue and try to delete the others.

For example, if there was an error during DeploymentConfig deletion it should still continue and try to delete Route .... 

At the end error should be returned with the list all errors that occurred.

pkg/occlient/occlient.go
```diff
@@ -1056,32 +1056,50 @@ func (c *Client) FollowBuildLog(buildName string) error {
 // kind is always required (can be set to 'all')
```
kadel: You have removed `kind` and `name` parameters, you should also update comment!

pkg/occlient/occlient.go
surajnarwade: @kadel , no :(

pkg/occlient/occlient.go
surajnarwade: gotcha

pkg/occlient/occlient.go
```diff
@@ -1052,36 +1052,59 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+	if err != nil {
+		errorList = append(errorList, "unable to delete route")
+	}
+	// Delete BuildConfig
+	err = c.buildClient.BuildConfigs(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		errorList = append(errorList, "unable to delete buildconfig")
 	}
+	// Delete Build
+	err = c.buildClient.Builds(c.namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
```
kadel: We don't need to delete `Builds`. They should be deleted automatically by [garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)

pkg/occlient/occlient.go
```diff
@@ -1052,36 +1052,59 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+// Delete takes labels as a input and based on it, deletes respective resource
+func (c *Client) Delete(labels map[string]string) error {
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
 
-	args := []string{
-		"delete",
-		kind,
+	var errorList []string
+	// Delete DeploymentConfig
```
kadel: It would be good to add some debug messages to see what is happening and progress of deletion.
Like selector that is being used to delete objects, and what type of objects are being deleted.

kadel:
Tested this, and it looks good.
Just a few small comments above.
@surajnarwade  Is this still "work in progress" and "do not merge"

pkg/occlient/occlient.go
surajnarwade: added :)

pkg/occlient/occlient.go
```diff
@@ -1090,36 +1090,60 @@ func (c *Client) DisplayDeploymentConfigLog(deploymentConfigName string, stdout
...
-// Delete calls oc delete
-// kind is always required (can be set to 'all')
-// name can be omitted if labels are set, in that case set name to ''
-// if you want to delete object just by its name set labels to nil
-func (c *Client) Delete(kind string, name string, labels map[string]string) (string, error) {
+// Delete takes labels as a input and based on it, deletes respective resource
+func (c *Client) Delete(labels map[string]string) error {
+	// convert labels to selector
+	selector := util.ConvertLabelsToSelector(labels)
+	log.Debug("Selectors used for deletion: %s", selector)
```
kadel: should be `log.Debugf`

PR #456, opened 2018-05-15
add test for CreateRoute using fakeClient

syamgk:
adding tests using fakeClient for testing functions in occlient pkg making API calls
https://github.com/redhat-developer/odo/issues/395

### DIFF
--- a/pkg/occlient/fakeclient.go
+++ b/pkg/occlient/fakeclient.go
@@ -1,0 +1,32 @@
+package occlient
+
+import (
+	fkappsclientset "github.com/openshift/client-go/apps/clientset/versioned/fake"
+	fkbuildclientset "github.com/openshift/client-go/build/clientset/versioned/fake"
+	fkimageclientset "github.com/openshift/client-go/image/clientset/versioned/fake"
+	fkprojectclientset "github.com/openshift/client-go/project/clientset/versioned/fake"
+	fkrouteClientset "github.com/openshift/client-go/route/clientset/versioned/fake"
+	fkkubernetes "k8s.io/client-go/kubernetes/fake"
+)
+
+// FkClientSet : hold fake ClientSets
+type FkClientSet struct {
+	kubeClientset    *fkkubernetes.Clientset
+	imageClientset   *fkimageclientset.Clientset
+	appsClientset    *fkappsclientset.Clientset
+	buildClientset   *fkbuildclientset.Clientset
+	projectClientset *fkprojectclientset.Clientset
+	routeClientset   *fkrouteClientset.Clientset
+}
+
+// FakeNew : create new fake client
+func FakeNew() (*Client, *FkClientSet) {
+	var client Client
+	var fkclientset FkClientSet
+
+	routeClient := fkrouteClientset.NewSimpleClientset()
+	fkclientset.routeClientset = routeClient
+	client.routeClient = fkclientset.routeClientset.Route()
+
+	return &client, &fkclientset
+}
--- a/pkg/occlient/fakeclient.go
+++ b/pkg/occlient/fakeclient.go
@@ -1,32 +1,24 @@
 package occlient
 
 import (
-	fkappsclientset "github.com/openshift/client-go/apps/clientset/versioned/fake"
-	fkbuildclientset "github.com/openshift/client-go/build/clientset/versioned/fake"
-	fkimageclientset "github.com/openshift/client-go/image/clientset/versioned/fake"
-	fkprojectclientset "github.com/openshift/client-go/project/clientset/versioned/fake"
-	fkrouteClientset "github.com/openshift/client-go/route/clientset/versioned/fake"
-	fkkubernetes "k8s.io/client-go/kubernetes/fake"
+	fakeRouteClientset "github.com/openshift/client-go/route/clientset/versioned/fake"
 )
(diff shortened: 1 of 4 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -6,10 +6,67 @@ import (
...
+				"app.kubernetes.io/component-name": "backend",
+				"app.kubernetes.io/component-type": "golang",
+			},
+			//			want:
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := fkclient.CreateRoute(tt.service, tt.labels)
```
ashetty1: LGTM!

ashetty1: A few things need to be cleaned up to make it look neat, but since it is a WIP, I am not going to add those comments right away. or should I?

pkg/occlient/fakeclient.go
```diff
@@ -0,0 +1,32 @@
...
+	fkappsclientset "github.com/openshift/client-go/apps/clientset/versioned/fake"
+	fkbuildclientset "github.com/openshift/client-go/build/clientset/versioned/fake"
+	fkimageclientset "github.com/openshift/client-go/image/clientset/versioned/fake"
+	fkprojectclientset "github.com/openshift/client-go/project/clientset/versioned/fake"
+	fkrouteClientset "github.com/openshift/client-go/route/clientset/versioned/fake"
+	fkkubernetes "k8s.io/client-go/kubernetes/fake"
+)
+
+// FkClientSet : hold fake ClientSets
+type FkClientSet struct {
```
kadel: this can stay private

pkg/occlient/fakeclient.go
```diff
@@ -0,0 +1,32 @@
...
+	kubeClientset    *fkkubernetes.Clientset
+	imageClientset   *fkimageclientset.Clientset
+	appsClientset    *fkappsclientset.Clientset
```
kadel: why are there clients if they are never initialized? This will just lead to unnecessary errors.

pkg/occlient/occlient_test.go
```diff
@@ -6,10 +6,83 @@ import (
...
 	"testing"
+
+	routev1 "github.com/openshift/api/route/v1"
+	ktesting "k8s.io/client-go/testing"
 )
 
-func TestGetOcBinary(t *testing.T) {
+func TestCreateRoute(t *testing.T) {
+	// initialising the fakeclient
+	fkclient, fkclientset := FakeNew()
```
kadel: this should be initialized for every test case separately

pkg/occlient/fakeclient.go
```diff
@@ -0,0 +1,32 @@
...
+	routeClientset   *fkrouteClientset.Clientset
+}
+
+// FakeNew : create new fake client
+func FakeNew() (*Client, *FkClientSet) {
+	var client Client
+	var fkclientset FkClientSet
+
+	routeClient := fkrouteClientset.NewSimpleClientset()
+	fkclientset.routeClientset = routeClient
```
ashetty1: Could do away with this line? Seems unnecessary. Could be done something like this:

```
routeClient := fkrouteClientset.NewSimpleClientset().Route()
```

pkg/occlient/occlient_test.go
ashetty1: You mean sub-tests? `TestCreateRoute` is a test case, no?

kadel: yes sub-test.
As it is currently written every test case reuses the same client.
That means that state from one test case can effect other.

pkg/occlient/fakeclient.go
kadel: we need ClientSet, so it could be replaced with this
```
fkclientset.routeClientset := fkrouteClientset.NewSimpleClientset()
```

kadel:
Added first test case for `url.Create` to verify that we can pass down fake clients outside of `occlient` package.

kadel:
As I added some changes to this PR it would be nice to have another person to review it.
ping @ashetty1

kadel:
As I added some changes to this PR it would be nice to have at least one another person to review it.
ping @ashetty1, @cdrage, @mik-dass

pkg/occlient/fakeclient.go
```diff
@@ -0,0 +1,24 @@
+package occlient
+
+import (
+	fkrouteClientset "github.com/openshift/client-go/route/clientset/versioned/fake"
```
cdrage: would fakeRouteClientSet be a better name?

pkg/occlient/fakeclient.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// fkClientSet holds fake ClientSets
+// this is returned by FakeNew to access methods of fake client sets
+type FkClientSet struct {
+	RouteClientset *fkrouteClientset.Clientset
+}
+
+// FakeNew create new fake client for testing
```
cdrage: creates a new

pkg/occlient/fakeclient.go
kadel: you are right. I've changed it.

PR #457, opened 2018-05-15
Swap componentName and applicationName in storage

cdrage:
Swaps the applicationName and componentName function parameters for
storage, so it is aligned with all other functions of Odo.


cdrage:
@kadel Good find, I updated this. Looks like we're not testing "description" in end to end tests.

kadel:
> @kadel Good find, I updated this. Looks like we're not testing "odo decsribe" in end to end tests.

yep, we still can't rely on tests. If tests are passing it doesn't mean that something didn't break.

PR #461, opened 2018-05-16
Update CLI information

cdrage:
Updates CLI information / grammar.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -14,22 +14,22 @@ import (
 var updateCmd = &cobra.Command{
 	Use:  "update",
 	Args: cobra.MaximumNArgs(1),
-	Example: `  # Change the source of a currently active component to local (use the current directory as a source)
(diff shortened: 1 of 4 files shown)


cmd/update.go
```diff
@@ -14,22 +14,22 @@ import (
 var updateCmd = &cobra.Command{
 	Use:  "update",
 	Args: cobra.MaximumNArgs(1),
-	Example: `  # Change the source of a currently active component to local (use the current directory as a source)
+	Example: `  # Change the source code of a currently active component to local (use the current directory as a source)
```
kadel: This sound a bit weird to me.
"Change the source code " sounds to me like it is changing the actual source code (lines of code).

I can't comment on the correctness of the previous formulation.
English is not really my forte :-) 

How about "Change the source code path ..." ?

cdrage: @kadel that's a better option :+1: I changed it to that!

cdrage:
Updated. Thanks @kadel

Issue #449, opened 2018-05-14
Don't delete all storage `odo storage delete`


kadel:
currently `odo storage delete` deletes all storage in the currently active component.

I'm not sure if we should allow that. It can be quite a destructive command.
I think that we should always require storage name argument.

```
# delete mystorage from current component
odo storage delete mystorage

# show error - storage name required
odo storage delete
```

kadel:
fixed in #463

PR #463, opened 2018-05-17
This will fix the issue of  deleting all storage

piyush1594:
This will fix the issue of  deleting all storage In spite of that, it will throw error #449


piyush1594:
@kadel Thanks for reviewing. PR updated

PR #469, opened 2018-05-23
 Fixed the formatting of storage list

mik-dass:

This PR fixes the `storage list` command to display the mounted storages on the active component and also the unmounted storages in the application, which can be mounted.

NOTE : https://github.com/redhat-developer/odo/pull/349 has to be merged first for this to work

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -8,6 +8,7 @@ import (
 	"github.com/redhat-developer/odo/pkg/storage"
 	"github.com/spf13/cobra"
 	"os"
+	"text/tabwriter"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


kadel:
This PR should be rebased as  https://github.com/redhat-developer/odo/pull/349  was just merged

cmd/storage.go
```diff
@@ -117,13 +118,38 @@ var storageListCmd = &cobra.Command{
...
 
-		if len(storageList) == 0 {
+		hasMounted := false
+		hasUnmounted := false
+		w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
+		w1 := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
+
+		for _, storage := range storageList {
+			if storage.Path != "" {
+				if !hasMounted {
```
kadel: Why is this condition here? Is just to add header only one time?

cmd/storage.go
```diff
@@ -117,13 +118,38 @@ var storageListCmd = &cobra.Command{
...
+				fmt.Fprintln(w1, storage.Name, "\t", storage.Size)
+			}
+		}
+		if hasMounted {
+			fmt.Printf("The component '%v' has the following storage attached -\n", componentName)
+			w.Flush()
+		} else {
 			fmt.Printf("The component '%v' has no storage attached\n", componentName)
+		}
+
```
kadel: It would be nice to have one empty line between mounted and unmounted section. It will make a nice visual division.

cmd/storage.go
```diff
@@ -71,13 +118,38 @@ var storageListCmd = &cobra.Command{
 		storageList, err := storage.List(client, componentName, applicationName)
 		checkError(err, "Failed to list storage")
 
-		if len(storageList) == 0 {
+		hasMounted := false
+		hasUnmounted := false
+		w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
```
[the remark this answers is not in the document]
mik-dass: @kadel ok :+1:

cmd/storage.go
mik-dass: @kadel You mean a dashed line division?

kadel: no just empty line.
Instead of this:
```
▶ odo storage list
The component 'frontend' has the following storage attached -
NAME     SIZE     PATH
data     100M     /data
The following unmounted storages can be mounted to 'frontend' -
NAME              SIZE
data-frontend     100M
```

it should look like this:
```
▶ odo storage list
The component 'frontend' has the following storage attached -
NAME     SIZE     PATH
data     100M     /data

The following unmounted storages can be mounted to 'frontend' -
NAME              SIZE
data-frontend     100M
```

kadel:
> @kadel Yes it is to show the header only one time

sorry accidentally deleted my previous comment :-|

Wouldn't it be better to move writing header outside of the `for` loop?

kadel:
> @kadel Yes, I thought of that, but I wanted to do the whole list thing in one for loop, and couldn't find a better and simpler way to display the header with the tabwriter and flush thing

how about something like this:
```go
		w1 := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
		//create headers
		fmt.Fprintln(w, "NAME", "\t", "SIZE", "\t", "PATH")
		fmt.Fprintln(w1, "NAME", "\t", "SIZE")

		for _, storage := range storageList {
			if storage.Path != "" {
				hasMounted = true
				fmt.Fprintln(w, storage.Name, "\t", storage.Size, "\t", storage.Path)
			} else {
				hasUnmounted = true
				fmt.Fprintln(w1, storage.Name, "\t", storage.Size)
			}
		}
```

kadel:
> @kadel But that would keep contents in the tabwriter without flushing if hasMounted or hasUnmounted if are false, any way to discard without flushing to the std.output? Will it be garbage collected?

Yes, the garbage collector will collect it once it can't be accessed anymore.

kadel:
It was just a suggestion, if you want to keep it as it is right now it is OK.

mik-dass:
@kadel Ok I will move the headers to the top for simplicity and easy code reading

Issue #484, opened 2018-05-29
Create umount alias for unmount.


kadel:
Linux users have muscle memory for typing `umount`.
Opposite to that Windows users don't have any idea what umount is and unmount will make more sense for them.

We can add `umount` alias to `unmount` command to allow both variations.

But our documentation and examples should use only one variation of this - `unmount`.
Let's make sure that all our examples and docs are using `unmount`.

piyush1594:
We are good to close this. :+1:

PR #486, opened 2018-05-29
Adds an alias umount for unmount

piyush1594:
Adds an alias ` umount` for `unmount `

#484

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -63,7 +63,8 @@ var storageUnmountCmd = &cobra.Command{
   # Unmount storage 'database' from component 'mongodb'
   odo storage unmount database --component mongodb
 `,
-	Args: cobra.ExactArgs(1),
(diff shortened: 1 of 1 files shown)


cdrage:
This will conflict with https://github.com/redhat-developer/odo/pull/481 so maybe leave out the changes to the examples!

cmd/storage.go
```diff
@@ -61,9 +61,9 @@ var storageUnmountCmd = &cobra.Command{
   odo storage unmount dbstorage
 
   # Unmount storage 'database' from component 'mongodb'
-  odo storage unmount database --component mongodb
-`,
-	Args: cobra.ExactArgs(1),
+  odo storage unmount database --component mongodb`,
```
mik-dass: The "," in then end, should be in the next line for consistency with the other commands

PR #487, opened 2018-05-30
v0.0.6

kadel:


kadel:
Only test that failed is "installation test". It is because it is trying to install odo from v0.0.6 tag that will be created after this PR is merged.  This is bug in the test.

PR #492, opened 2018-05-31
Adds the functionality of unmounting storage using the mounted path.

mik-dass:
Issue : https://github.com/redhat-developer/odo/issues/493

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -52,16 +52,22 @@ var storageCreateCmd = &cobra.Command{
 }
 
 var storageUnmountCmd = &cobra.Command{
-	Use:   "unmount [storage name]",
-	Short: "Unmount storage from the current component",
-	Long: `Unmount storage from the current component.
+	Use:   "unmount PATH | STORAGE_NAME",
+	Short: "Unmount storage from the given path or identified by its name, from the current component",
+	Long: `Unmount storage from the given path or identified by its name, from the current component.
 
-  The storage and the contents are not deleted, the storage is only unmounted from the component, and hence is no longer accessible by the component.`,
+  The storage and the contents are not deleted, the storage or storage with the given path, is only unmounted from the component, and hence is no longer accessible by the component.`,
 	Example: `  # Unmount storage 'dbstorage' from current component
   odo storage unmount dbstorage
 
   # Unmount storage 'database' from component 'mongodb'
   odo storage unmount database --component mongodb
+
+  # Unmount storage mounted to path '/data' from current component
+  odo storage unmount /data
+
+  # Unmount storage mounted to path '/data' from component 'mongodb'
+  odo storage unmount /data --component mongodb
 `,
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
@@ -71,12 +77,24 @@ var storageUnmountCmd = &cobra.Command{
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
 
-		storageName := args[0]
-		exists, err := storage.IsMounted(client, storageName, componentName, applicationName)
-		checkError(err, "")
-		if !exists {
-			fmt.Printf("Storage %v does not exist in component %v\n", storageName, componentName)
-			os.Exit(1)
+		var storageName string
+		// checking if the first character in the argument is a "/", indicating a path or not, indicating a storage name
+		if string(args[0][0]) == "/" {
+			path := args[0]
+			storageName, err = storage.GetStorageNameFromMountPath(client, path, componentName, applicationName)
+			checkError(err, "Unable to get storage name from mount path")
+			if storageName == "" {
+				fmt.Printf("No storage is mounted to %s in the component %s\n", path, componentName)
+				os.Exit(1)
+			}
+		} else {
+			storageName = args[0]
+			exists, err := storage.IsMounted(client, storageName, componentName, applicationName)
+			checkError(err, "Unable to check if storage is mounted or not")
+			if !exists {
+				fmt.Printf("Storage %v does not exist in component %v\n", storageName, componentName)
+				os.Exit(1)
+			}
 		}
 
 		err = storage.Unmount(client, storageName, componentName, applicationName)
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -298,3 +298,18 @@ func Mount(client *occlient.Client, path string, storageName string, componentNa
 	}
 	return nil
 }
+
+// Gets the storageName mounted to the given path in the given component and application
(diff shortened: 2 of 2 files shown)


cmd/storage.go
```diff
@@ -71,12 +77,23 @@ var storageUnmountCmd = &cobra.Command{
...
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
 
-		storageName := args[0]
-		exists, err := storage.IsMounted(client, storageName, componentName, applicationName)
-		checkError(err, "")
-		if !exists {
-			fmt.Printf("Storage %v does not exist in component %v\n", storageName, componentName)
-			os.Exit(1)
+		storageName := ""
+		if string(args[0][0]) == "/" {
```
kadel: It might be nice to have command explaining what we are doing here.

cdrage:
What would happen if we pass in: `./` or something that is not valid? Shouldn't we error out to the user?

mik-dass:
@cdrage You mean at the start of the path?

cdrage:
@mik-dass Yes

mik-dass:
@cdrage I think it might be complicated as there are many combinations which are possible

kadel:
> What would happen if we pass in: ./ or something that is not valid (or does not exist)? Shouldn't we error out to the user?

This should be treated as a storage name, and it should fail because no such storage exits.

cmd/storage.go
```diff
@@ -52,16 +52,22 @@ var storageCreateCmd = &cobra.Command{
 }
 
 var storageUnmountCmd = &cobra.Command{
-	Use:   "unmount [storage name]",
-	Short: "Unmount storage from the current component",
-	Long: `Unmount storage from the current component.
+	Use:   "unmount [storage/path name]",
```
kadel: we can update this based on https://github.com/redhat-developer/odo/issues/516
`"PATH | STORAGE_NAME"`

cmd/storage.go
```diff
@@ -52,16 +52,22 @@ var storageCreateCmd = &cobra.Command{
...
+	Short: "Unmount storage or storage with the given path, from the current component",
+	Long: `Unmount storage or storage with the given path, from the current component.
```
kadel: How about this:
Unmount storage from the given path or unmount storage identified by its name.

mik-dass:
@kadel Done

cmd/storage.go
```diff
@@ -52,16 +52,22 @@ var storageCreateCmd = &cobra.Command{
 }
 
 var storageUnmountCmd = &cobra.Command{
-	Use:   "unmount [storage name]",
-	Short: "Unmount storage from the current component",
-	Long: `Unmount storage from the current component.
+	Use:   "unmount [PATH | STORAGE_NAME]",
```
cdrage: this should be lowercase, other sources (such as `create.go`) use lowercase.

`unmount <path | storage_name>` maybe better?

cmd/storage.go
```diff
@@ -71,12 +77,24 @@ var storageUnmountCmd = &cobra.Command{
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
 
-		storageName := args[0]
-		exists, err := storage.IsMounted(client, storageName, componentName, applicationName)
-		checkError(err, "")
-		if !exists {
-			fmt.Printf("Storage %v does not exist in component %v\n", storageName, componentName)
-			os.Exit(1)
+		storageName := ""
```
cdrage: if this starts as blank, I think `var storageName string` is more go-esque

cmd/storage.go
```diff
@@ -71,12 +77,24 @@ var storageUnmountCmd = &cobra.Command{
...
+		// checking if the first character in the argument is a "/", indicating a path or not, indicating a storage name
+		if string(args[0][0]) == "/" {
+			path := args[0]
+			storageName, err = storage.GetStorageNameFromMountPath(client, path, componentName, applicationName)
+			checkError(err, "")
```
cdrage: instead of `""` maybe add explanation (`Unable to get storage name from mount path`) or something similar?

cmd/storage.go
```diff
@@ -71,12 +77,24 @@ var storageUnmountCmd = &cobra.Command{
...
+			if storageName == "" {
+				fmt.Printf("No storage is mounted to %s in the component %s\n", path, componentName)
+				os.Exit(1)
+			}
+		} else {
+			storageName = args[0]
+			exists, err := storage.IsMounted(client, storageName, componentName, applicationName)
+			checkError(err, "")
```
cdrage: same here, maybe add own explanation?

pkg/storage/storage.go
```diff
@@ -298,3 +298,18 @@ func Mount(client *occlient.Client, path string, storageName string, componentNa
 	}
 	return nil
 }
+
+// Gets the storageName mounted to the given path in the given component and application
+// GetStorageNameFromMountPath returns the name of the storage or the error
+func GetStorageNameFromMountPath(client *occlient.Client, path string, componentName string, applicationName string) (string, error) {
+	storages, err := List(client, componentName, applicationName)
+	if err != nil {
+		return "", errors.Wrap(err, "unable to list storages")
```
cdrage: storage not storages

mik-dass: @cdrage But we refer to one PVC as storage

cdrage: @mik-dass In English, we don't use the plural "storages". It's just "unable to list storage" https://en.wiktionary.org/wiki/storage

cmd/storage.go
kadel: @cdrage see https://github.com/redhat-developer/odo/issues/516

there are two options, either uppercase, or lowercase in `<>`

kadel: @mik-dass it can't be in `[]` as it means that it is optional. Please check rules in  #516

pkg/storage/storage.go
mik-dass: @cdrage I know but a single PVC is called storage and all of them together is also called storage. It seemed confusing. So I added the 's' to distinguish. Anyways updating it.

cmd/storage.go
mik-dass: @kadel So between <>?

kadel: As described in docopts docs (link in #516) 

either just capital letters 
```
unmount PATH | STORAGE_NAME
```

or in `<>`
```
unmount <path> | <storage_name>
```

I would use the capital letters, it  seems to be more widly used (heroku and could foundry).
We should pick one and stick with it across all commands in OpenShift-Do.

cdrage:
LGTM

cmd/storage.go
mik-dass: @kadel OK I will changed to the Upper case approach

PR #496, opened 2018-05-31
Adds --follow command to odo log

cdrage:
This adds the --follow or -f command to `odo log` in order to stream logs
as they are being built for DeploymentConfig.

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -9,6 +9,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	logFollow bool
+)
+
 var logCmd = &cobra.Command{
 	Use:   "log [component_name]",
 	Short: "Retrieve the log for the given component.",
@@ -41,12 +45,15 @@ var logCmd = &cobra.Command{
 		// Retrieve and set the currentComponent
 		currentComponent := getComponent(client, argComponent, currentApplication, currentProject)
 
-		err = component.GetLogs(client, currentComponent, stdout)
+		// Retrieve the log
+		err = component.GetLogs(client, currentComponent, logFollow, stdout)
 		checkError(err, "Unable to retrieve logs, does your component exist?")
 	},
 }
 
 func init() {
+	logCmd.Flags().BoolVarP(&logFollow, "follow", "f", false, "Follow logs")
+
 	// Add a defined annotation in order to appear in the help menu
 	logCmd.Annotations = map[string]string{"command": "component"}
 	logCmd.SetUsageTemplate(cmdUsageTemplate)
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -25,6 +25,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	servicecatalogclienset "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset/typed/servicecatalog/v1beta1"
+	appsschema "github.com/openshift/client-go/apps/clientset/versioned/scheme"
 	appsclientset "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	buildschema "github.com/openshift/client-go/build/clientset/versioned/scheme"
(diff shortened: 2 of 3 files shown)


cmd/log.go
```diff
@@ -41,12 +45,13 @@ var logCmd = &cobra.Command{
...
+		// Retrieve the log
+		err = component.GetLogs(client, currentComponent, logWatch, stdout)
 		checkError(err, "Unable to retrieve logs, does your component exist?")
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(logCmd)
-
+	logCmd.Flags().BoolVarP(&logWatch, "watch", "w", false, "Stream logs")
```
kadel: can we use `--follow` / `-f`? 
I think that `-f` is more standard for this kind of things.  `kubect logs -f`, `tail -f` ...

cdrage:
@kadel I'll make the changes! `-f` or `--follow` is indeed a better choice. I was conflicted between --watch or --follow.

pkg/occlient/occlient.go
```diff
@@ -1053,37 +1054,62 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
+
+	// Set standard log options
+	deploymentLogOptions := appsv1.DeploymentLogOptions{Follow: false, Previous: true, NoWait: true}
+
+	// If the log is being followed, set it to follow / don't wait
+	if followLog {
+		// TODO: https://github.com/kubernetes/kubernetes/pull/60696
+		// Unable to set to 0, until we update our vendoring
+		// set to 1 for now.
+		tailLines := int64(1)
```
cdrage: Need your insight on this @kadel 

I'm blocked on issue https://github.com/kubernetes/kubernetes/pull/60696

The commit didn't make it into the `1.10.0` Kubernetes release freeze. It will however be in the `1.11.0` release.

At the moment, the `client-go` vendoring for OpenShift is at `1.10.2` https://github.com/openshift/client-go/blob/master/glide.yaml#L14

So we literally just have to wait until `1.11.0` of Kubernetes is released and go from there. 

The current work-around is to set it to `1`, where it will only output 1 line from the previous log, until `1.11.0` is released.

cdrage:
Need your insight on this @kadel

https://github.com/redhat-developer/odo/pull/496/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR1068

I'm blocked on issue kubernetes/kubernetes#60696

The commit didn't make it into the 1.10.0 Kubernetes release freeze. It will however be in the 1.11.0 release.

At the moment, the client-go vendoring for OpenShift is at 1.10.2 https://github.com/openshift/client-go/blob/master/glide.yaml#L14

So we literally just have to wait until 1.11.0 of Kubernetes is released and go from there.

The current work-around is to set it to 1, where it will only output 1 line from the previous log, until 1.11.0 is released.

kadel:
Setting it to `1` is OK. It might be even better than `0`, the user sees something when the command is run. 

It the future it might be good to have this as a parameter. 

But the current implementation is good for now. 👍

cdrage:
@kadel Awesome :+1: 

Code / review look good to you?

pkg/occlient/occlient.go
```diff
@@ -1053,37 +1054,62 @@ func (c *Client) FollowBuildLog(buildName string) error {
...
 // Display DeploymentConfig log to stdout
-func (c *Client) DisplayDeploymentConfigLog(deploymentConfigName string, stdout io.Writer) error {
+func (c *Client) DisplayDeploymentConfigLog(deploymentConfigName string, followLog bool, stdout io.Writer) error {
+
+	// Set standard log options
+	deploymentLogOptions := appsv1.DeploymentLogOptions{Follow: false, NoWait: true}
+
+	// If the log is being followed, set it to follow / don't wait
+	if followLog {
+		// TODO: https://github.com/kubernetes/kubernetes/pull/60696
```
mik-dass, line 1064: this seems to have been merged a long time back. Is it updated now?

PR #509, opened 2018-06-05
Fixed the problem of storage list when no component is created or selected

mik-dass:
Issue: https://github.com/redhat-developer/odo/issues/498

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -158,8 +158,13 @@ var storageListCmd = &cobra.Command{
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
 
+		if componentName == "" {
+			fmt.Println("No component selected")
+			os.Exit(1)
+		}
+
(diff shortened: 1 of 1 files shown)


cmd/storage.go
```diff
@@ -158,8 +158,13 @@ var storageListCmd = &cobra.Command{
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
 
+		if componentName == "" {
+			fmt.Println("No component selected")
+			os.Exit(1)
+		}
+
 		storageList, err := storage.List(client, componentName, applicationName)
-		checkError(err, "Failed to list storage")
```
kadel: Why has been this message removed?

mik-dass: @kadel It was hiding the underlying error without verbose flag enabled

kadel: That was kind of the point of that.
But ok, lets leave it as it is for now.

PR #515, opened 2018-06-06
Add new examples

cdrage:
Adds a list of different examples (using each language that OpenShift
catalog supports) as well as updates the `create.go` command to list
each and every one of them!


cmd/create.go
```diff
@@ -40,11 +40,14 @@ A full list of component types that can be deployed is available using: 'odo com
   # Create new Wildfly component with binary named sample.war in './downloads' directory
   odo create wildfly wildly --binary ./downloads/sample.war
 
-  # Create a Ruby component
-  odo create ruby
-	
-  # Create a Python component
-  odo create python
+  # List of ready-to-use examples
+  odo create httpd --git https://github.com/openshift/httpd-ex.git
```
kadel: Does it really make sense to list all the possibilities here? 
It looks really polluted. And to be honest is there any benefit of it?

I like the `docs/examples.md` document. But I'm not sure about listing everything here.

Examples here should be there just for users to help them understand the command variations and structure

docs/getting-started.md
```diff
@@ -92,7 +92,7 @@ Switched to application: nodeapp
 First, we'll download the our test application: 
 
 ```console
-$ git clone https://github.com/kadel/nodejs-ex
+$ git clone https://github.com/openshift/nodejs-ex
```
kadel: +1

cmd/create.go
cdrage: We could include maybe one and a link to more examples? What do you think

surajnarwade: something like, `For more examples....`

kadel: > We could include maybe one and a link to more examples? What do you think

That would be much better.

surajnarwade:
LGTM :)

cmd/create.go
```diff
@@ -40,11 +40,10 @@ A full list of component types that can be deployed is available using: 'odo com
   # Create new Wildfly component with binary named sample.war in './downloads' directory
   odo create wildfly wildly --binary ./downloads/sample.war
 
-  # Create a Ruby component
-  odo create ruby
-	
-  # Create a Python component
-  odo create python
+  # List of ready-to-use examples
+	# for more examples, visit: https://github.com/redhat-developer/odo/blob/master/docs/examples.md
```
kadel: wrong indentation

cmd/create.go
```diff
@@ -40,11 +40,10 @@ A full list of component types that can be deployed is available using: 'odo com
...
+  odo create php --git https://github.com/openshift/cakephp-ex.git
```
kadel: Would it make sense to remove this completely and just keep the note above?

cdrage: IMO adding something that is not NodeJS would help with our diversification of languages (showcase that NodeJS is not a first-class citizen)

cdrage:
@kadel Updated!

PR #517, opened 2018-06-07
add unit test for DeletePVC

syamgk:
add unit test for DeletePVC

### DIFF
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -128,6 +128,48 @@ func TestCreatePVC(t *testing.T) {
 	}
 }
 
+func TestDeletePVC(t *testing.T) {
+	tests := []struct {
+		name    string
+		pvcName string
+		wantErr bool
+	}{
+		{
+			name:    "storage 10Gi",
+			pvcName: "postgresql",
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, fakeClientSet := FakeNew()
+
+			fakeClientSet.Kubernetes.PrependReactor("delete", "persistentvolumeclaims", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, nil, nil
+			})
+
+			err := fakeClient.DeletePVC(tt.pvcName)
+
+			//Checks for error in positive cases
+			if !tt.wantErr == (err != nil) {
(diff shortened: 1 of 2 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -8,9 +8,45 @@ import (
...
+			//Checks for error in positive cases
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.DeletePVC(name) unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			// Check for validating actions performed
+			if (len(fakeClientSet.Kubernetes.Actions()) != 1) && (tt.wantErr != true) {
+				t.Errorf("expected 1 action in DeletePVC got: %v", fakeClientSet.Kubernetes.Actions())
+			}
+		})
```
ashetty1: Just check if `fakeClientSet.Kubernetes.Actions()[0].(ktesting.DeleteAction).GetName() returns name

pkg/occlient/occlient_test.go
```diff
@@ -8,9 +8,45 @@ import (
...
 	routev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	ktesting "k8s.io/client-go/testing"
 )
 
+func TestDeletePVC(t *testing.T) {
+	tests := []struct {
+		name    string
+		pvcname string
+		wantErr bool
```
surajnarwade: One more use for failure as well

pkg/occlient/occlient_test.go
syamgk: yep, it did return the `tt.name`,

added checks for validating it.

thanks, @ashetty1

pkg/occlient/occlient_test.go
syamgk: Hi @surajnarwade ,

here in this test case, It is only validating whether the correct function is called with the correct parameter, So adding a failure case is not possible

Also according to the Unit test philosophy, we don't need to test the functionality of library function call eg: in this case `c.kubeClient.CoreV1().PersistentVolumeClaims(c.namespace).Delete(name, nil)`
It's out of our scope of testing because we are in an assumption that the library which we are using is tested by them, and we only need to validate the code which we have written.
This is what my understanding so far

kadel:
hi @syamgk, is this still work in progress?

syamgk:
@kadel its not 
also removed the WIP label

cdrage:
This LGTM! No comments regarding the code

kadel:
need rebase

syamgk:
rebase done, removing label

PR #522, opened 2018-06-11
Change usecase5 to testing app name for e2e tests

cdrage:
Changes the ambiguous "usecase5" name to "testing". Also adds a constant
to the e2e_test.go file.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -118,6 +118,7 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("odoe2e", func() {
+	appTestName := "testing"
 
 	tmpDir, err := ioutil.TempDir("", "odo")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -23,6 +23,7 @@ import (
 
 var t = strconv.FormatInt(time.Now().Unix(), 10)
 var projName = fmt.Sprintf("odo-%s", t)
+var appTestName = "testing"
```
kadel: shouldn't this be defined inside ` var _ = Describe("odoe2e", func() {` ?

cdrage:
@kadel I've updated the tests to indicate the new name in `Describe` as well!

kadel:
> @kadel I've updated the tests to indicate the new name in Describe as well!

I meant something different.

To make it easy to create another test with the different application name we should define `appTestName` inside `Describe` block.

```go
var _ = Describe("odoe2e", func() {
   appTestName := "testapp"
...
...
```

cdrage:
@kadel I see what you mean. Updated it and added the variable to be described under `Describe(`

kadel:
> @kadel I see what you mean. Updated it and added the variable to be described under Describe(
> 

👍  Thank you

PR #526, opened 2018-06-13
Update installation instructions, add binaries

cdrage:
Updates the current installation instuctions. Adds links to the
binaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,21 @@ curl -L https://github.com/redhat-developer/odo/raw/master/scripts/install.sh | 
 #### macOS
 
 ```sh
-sudo curl -L  "https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64.gz" | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo
+# Binary installation
+sudo curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
+
+# Alternative, compressed tarball installation
+sudo sh -c 'curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64.gz | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo'
 ```
 
 #### Linux
 
 ```sh
-sudo curl -L  "https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-linux-amd64.gz" | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo
+# Binary installation
+sudo curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-linux-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
+
+# Alternative, compressed tarball installation
+sudo sh -c 'curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-linux-amd64.gz | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo'
 ```
 
 #### Windows
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -27,8 +27,13 @@ for arch in `ls -1 $BIN_DIR/`;do
         suffix=".exe"
     fi
     source_file=$BIN_DIR/$arch/odo$suffix
-    target_file=$RELEASE_DIR/odo-$arch$suffix.gz
+    target_file=$RELEASE_DIR/odo-$arch$suffix
 
+    # Create a gzip of the binary
     echo "gzipping binary $source_file as $target_file"
-    gzip --keep --to-stdout $source_file > $target_file
+    gzip --keep --to-stdout $source_file > $target_file.gz
+
+    # Move binaries to the release directory as well
+    echo "copying binary $source_file to release directory"
+    cp $source_file $target_file
 done


cdrage:
Ping @kadel 

This should be merged before a release or after editing the current release (by adding the new binaries)

surajnarwade:
Instead of putting version in URL, what if we make link generic (which will fetch latest release always)

For reference:

* https://developer.github.com/v3/repos/releases/#get-the-latest-release
* https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c

Does this make sense ? @cdrage  @kadel

kadel:
> Instead of putting version in URL, what if we make link generic (which will fetch latest release always)
> 
> For reference:
> 
> https://developer.github.com/v3/repos/releases/#get-the-latest-release
> https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
> Does this make sense ? @cdrage @kadel

It is using GitHub API, so there will be problems with rate limiting  (we already have those issues with update chacked - https://github.com/redhat-developer/odo/issues/122#issuecomment-392809205)

scripts/prepare-release.sh
```diff
@@ -29,6 +29,11 @@ for arch in `ls -1 $BIN_DIR/`;do
     source_file=$BIN_DIR/$arch/odo$suffix
     target_file=$RELEASE_DIR/odo-$arch$suffix.gz
 
+    # Create a gzip of the binary
     echo "gzipping binary $source_file as $target_file"
     gzip --keep --to-stdout $source_file > $target_file
+
+    # Move binaries to the release directory as well
+    echo "moving binary $source_file to release directory"
+    mv source_file $RELEASE_DIR/odo-$arch$suffix
```
kadel: you are creating filename again, that means that the logic how to construct filename is in two places.

how about removing `.gz` suffix from `target_file`
```
target_file=$RELEASE_DIR/odo-$arch$suffix
```
updating gzip command to include `.gz`
```
gzip --keep --to-stdout $source_file > $target_file.gz
```

and using the same variable again 
```
mv source_file $RELEASE_DIR/$target_file
```

kadel: Is it safe to move it those files? 
I'm not sure but other scripts that are running after this might expect those binaries to be there.
It is just hunch, I haven't checked that.

cdrage:
@kadel Updated! And using `cp` instead of `mv` for the files

scripts/prepare-release.sh
```diff
@@ -27,8 +27,13 @@ for arch in `ls -1 $BIN_DIR/`;do
...
+    target_file=$RELEASE_DIR/odo-$arch$suffix
 
+    # Create a gzip of the binary
     echo "gzipping binary $source_file as $target_file"
-    gzip --keep --to-stdout $source_file > $target_file
+    gzip --keep --to-stdout $source_file > $target_file.gz
+
+    # Move binaries to the release directory as well
+    echo "moving binary $source_file to release directory"
+    cp source_file $RELEASE_DIR/$target_file
```
kadel: missing `$` for `source_file`

cdrage:
@kadel Updated

README.md
```diff
@@ -39,13 +39,21 @@ curl -L https://github.com/redhat-developer/odo/raw/master/scripts/install.sh |
 #### macOS
 
 ```sh
-sudo curl -L  "https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64.gz" | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo
+# Binary installation
+curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64 -o odo && chmod +x odo && sudo mv odo /usr/local/bin/odo
```
kadel: sorry, noticed this just now.
could we make this shorter?

`sudo curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64 -o  /usr/local/bin/odo && chmod +x /usr/local/bin/odo`

scripts/prepare-release.sh
```diff
@@ -27,8 +27,13 @@ for arch in `ls -1 $BIN_DIR/`;do
...
-    target_file=$RELEASE_DIR/odo-$arch$suffix.gz
+    target_file=$RELEASE_DIR/odo-$arch$suffix
 
+    # Create a gzip of the binary
     echo "gzipping binary $source_file as $target_file"
-    gzip --keep --to-stdout $source_file > $target_file
+    gzip --keep --to-stdout $source_file > $target_file.gz
+
+    # Move binaries to the release directory as well
+    echo "moving binary $source_file to release directory"
```
kadel: `s/moving/copying`

README.md
cdrage: that works

cdrage:
@kadel Updated again.

kadel:
the script is still failing

README.md
```diff
@@ -39,13 +39,21 @@ curl -L https://github.com/redhat-developer/odo/raw/master/scripts/install.sh |
...
+# Alternative, compressed tarball installation
+sudo sh -c 'curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-darwin-amd64.gz | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo'
 ```
 
 #### Linux
 
 ```sh
-sudo curl -L  "https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-linux-amd64.gz" | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo
+# Binary installation
+sudo curl -L https://github.com/redhat-developer/odo/releases/download/v0.0.6/odo-linux-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
```
syamgk: the binary download url says `not found`
i think we have to host the binary first before merging this

cdrage: Yup, that's why the beginning of the PR says that we must push these binaries first either this release or next release

cdrage:
@kadel Fixed. Ready for another review!

PR #530, opened 2018-06-18
fix e2e tests

ashetty1:
Was failing due to references to old application name in some parts of the code.


cdrage:
yeah, woops! LGTM.

PR #532, opened 2018-06-20
fix broken TestDelete

syamgk:
fix broken TestDelete in url_test.go


surajnarwade:
use `tt.args.urlName` instead of `tt.args.name` in line 95

PR #533, opened 2018-06-20
Validate storage path

### LINKED ISSUE #505: If adding storage fails, Odo does not roll-back deployment.
```sh
▶ ./odo storage create pv1 --path /mnt/pv1 --size 5Gi
Added storage pv1 to foobar

▶ ./odo storage list
The component 'foobar' has the following storage attached -
NAME     SIZE     PATH
pv1      5Gi      /mnt/pv1

No unmounted storage exists to mount to 'foobar'

▶ ./odo storage create pv2 --path /mnt/pv1 --size 5Gi
DeploymentConfig "app-foobar" is invalid: spec.template.spec.containers[0].volumeMounts[1].mountPath: Invalid value: "/mnt/pv1": must be unique
```

Steps to re-produce (using the same mount path)
```sh
odo storage create pv1 --path /mnt/pv1 --size 5Gi
odo storage create pv2 --path /mnt/pv1 --size 5Gi
```

Odo will not roll-back the deployment (delete it if it fails with an invalid value..) and will appear in `storage list`:

```sh
▶ odo storage list
The component 'foobar' has the following storage attached -
NAME     SIZE     PATH
pv3      5Gi      /mnt/pv3
pv1      5Gi      /mnt/pv1

The following unmounted storages can be mounted to 'foobar' - 
NAME     SIZE
pv2      5Gi
```

surajnarwade:
This PR will add functionality to validate storage path before creating storage (internally PVC)
Resolves #505


cmd/utils.go
```diff
@@ -85,3 +86,17 @@ func validateName(name string) error {
 	return nil
 
 }
+
+// validateStoragePath will validate storagePath, if there is any existing storage with similar path, it will give an error
+func validateStoragePath(client *occlient.Client, storagePath, componentName, applicationName string) error {
+	storeList, err := storage.List(client, componentName, applicationName)
+	if err != nil {
+		log.Debugf("Unable to list storage")
```
mik-dass: Shouldn't the error be returned and further execution be stopped?

surajnarwade: if there is no existing storage, it will throw an error, so I didn't returned error here

mik-dass: But I think since the function returns the error, it's better to throw the error for later use in the calling function

surajnarwade: Let's say there is no existing storage, it is gonna definitely throw an error. but that's perfectly fine situation. we are just validating here.

mik-dass: But storage.List might return error (e.g due to PVC error), validateStoragePath won't return and stop execution and it will return nil. So checkError() won't catch it and the program will still try to create the PVC.
```
[mdas@localhost odo]$ ./odo storage create backend --path /data --size 100M
Added storage backend to wildfly
[mdas@localhost odo]$ ./odo storage create backend-1 --path /data --size 100M
DeploymentConfig "wildfly-app" is invalid: spec.template.spec.containers[0].volumeMounts[2].mountPath: Invalid value: "/data": must be unique
```
Here I deleted the PVC to replicate the storage.List() PVC error, the error was not caught and the program still tried to make the PVC.

surajnarwade: @mik-dass  Sorry my bad , I was little confused. updating the PR accordingly

mik-dass: No problem, my bad too, I should have explained better at the start. Anyways rest looks good to me :+1:

surajnarwade:
@syamgk @kadel @cdrage review needed :)

cmd/utils.go
```diff
@@ -85,3 +85,17 @@ func validateName(name string) error {
...
+
+// validateStoragePath will validate storagePath, if there is any existing storage with similar path, it will give an error
+func validateStoragePath(client *occlient.Client, storagePath, componentName, applicationName string) error {
+	storeList, err := storage.List(client, componentName, applicationName)
+	if err != nil {
+		return err
+	}
+	for _, store := range storeList {
+		if store.Path == storagePath {
+			return errors.Errorf("%s - storage path must be unique, please use a different path", storagePath)
```
kadel: Not sure if this is a right message.
How about something like "there already is a  storage mounted at %s"

cmd/storage.go
```diff
@@ -44,7 +44,9 @@ var storageCreateCmd = &cobra.Command{
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
-
+		//validate storage path
```
cdrage: add space before validate

cmd/storage.go
```diff
@@ -44,7 +44,9 @@ var storageCreateCmd = &cobra.Command{
...
+		err = validateStoragePath(client, storagePath, componentName, applicationName)
+		checkError(err, "")
 		_, err = storage.Create(client, args[0], storageSize, storagePath, componentName, applicationName)
```
cdrage: I feel as though we should start organizing the code a bit better and add newlines on seperate sections. For instance... The above code talks about validating storage path, while the code below here is regarding Create. Perhaps adding a newline between this code would make things look cleaner.

surajnarwade: thanks @cdrage

surajnarwade:
@kadel @cdrage done with changes

cdrage:
This LGTM, could you add a test for this @surajnarwade ?

surajnarwade:
@cdrage I am confused how can I write test for it since it needs storage list

cmd/cmdutils.go
```diff
@@ -85,3 +85,17 @@ func validateName(name string) error {
 	return nil
 
 }
+
+// validateStoragePath will validate storagePath, if there is any existing storage with similar path, it will give an error
+func validateStoragePath(client *occlient.Client, storagePath, componentName, applicationName string) error {
```
cdrage: test here ^^ Would it be possible to use fakeClient?

kadel:
@surajnarwade what is status of this?

surajnarwade:
@kadel , I am working on tests

kadel:
no tests for `validateStoragePath` :-(

cmd/storage.go
```diff
@@ -52,6 +52,12 @@ var storageCreateCmd = &cobra.Command{
 			storageName = componentName + "-" + util.GenerateRandomString(4)
 		}
 		_, err = storage.Create(client, storageName, storageSize, storagePath, componentName, applicationName)
+
+		// validate storage path
+		err = validateStoragePath(client, storagePath, componentName, applicationName)
+		checkError(err, "")
+
+		_, err = storage.Create(client, args[0], storageSize, storagePath, componentName, applicationName)
```
kadel: create is already there on line 54

cmd/cmdutils.go
surajnarwade: added

pkg/occlient/occlient_test.go
```diff
@@ -2460,3 +2461,69 @@ func TestCreateService(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, fakeClientSet := FakeNew()
+
+			fakeClientSet.AppsClientset.PrependReactor("list", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, &listOfDC, nil
+			})
+			dc, err := fakeClient.GetDeploymentConfigsFromSelector(tt.selector)
+
+			if len(fakeClientSet.AppsClientset.Actions()) != 1 {
+				t.Errorf("test failed")
```
mik-dass: Maybe a better error message like https://github.com/redhat-developer/odo/blob/3f11739e4c29583844222dc2f904dd0d982c8f33/pkg/occlient/occlient_test.go#L1862

pkg/occlient/occlient_test.go
```diff
@@ -2460,3 +2461,69 @@ func TestCreateService(t *testing.T) {
...
+			}
+
+			if tt.wantErr == false && err != nil {
+				t.Errorf("test failed, %#v", dc[0].Labels)
```
mik-dass: Same here something like with expected and current output https://github.com/redhat-developer/odo/blob/3f11739e4c29583844222dc2f904dd0d982c8f33/pkg/occlient/occlient_test.go#L1892

pkg/occlient/occlient_test.go
```diff
@@ -2460,3 +2461,69 @@ func TestCreateService(t *testing.T) {
...
+			}
+
+			for _, dc1 := range dc {
+				if !reflect.DeepEqual(dc1.Labels, tt.label) {
+					t.Errorf("test failed, %#v", dc1.Labels)
```
mik-dass: same here as above

cmd/cmdutils_test.go
```diff
@@ -58,3 +68,133 @@ func Test_validateName(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	client, fakeClientSet := occlient.FakeNew()
+	fakeClientSet.AppsClientset.PrependReactor("list", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
+		return true, &listOfDC, nil
```
mik-dass: Maybe here we can add a check for the selectors https://github.com/redhat-developer/odo/blob/3f11739e4c29583844222dc2f904dd0d982c8f33/pkg/url/url_test.go#L229

cmd/cmdutils_test.go
```diff
@@ -58,3 +68,133 @@ func Test_validateName(t *testing.T) {
...
+	})
+
+	fakeClientSet.Kubernetes.PrependReactor("get", "persistentvolumeclaims", func(action ktesting.Action) (bool, runtime.Object, error) {
+		return true, &pvc, nil
+	})
+
+	fakeClientSet.Kubernetes.PrependReactor("list", "persistentvolumeclaims", func(action ktesting.Action) (bool, runtime.Object, error) {
```
mik-dass: Same here maybe we can add a check for the selectors https://github.com/redhat-developer/odo/blob/3f11739e4c29583844222dc2f904dd0d982c8f33/pkg/url/url_test.go#L229

cmd/cmdutils_test.go
```diff
@@ -58,3 +68,133 @@ func Test_validateName(t *testing.T) {
...
+		},
+	}
+
+	client, fakeClientSet := occlient.FakeNew()
+	fakeClientSet.AppsClientset.PrependReactor("list", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
+		return true, &listOfDC, nil
+	})
+
+	fakeClientSet.Kubernetes.PrependReactor("get", "persistentvolumeclaims", func(action ktesting.Action) (bool, runtime.Object, error) {
+		return true, &pvc, nil
```
mik-dass: Maybe we can add a check here for the pvc name https://github.com/redhat-developer/odo/blob/3f11739e4c29583844222dc2f904dd0d982c8f33/pkg/occlient/occlient_test.go#L1111

pkg/occlient/occlient_test.go
```diff
@@ -2460,3 +2461,69 @@ func TestCreateService(t *testing.T) {
...
+		},
+	}
+
+	// fmt.Println(tests)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, fakeClientSet := FakeNew()
+
+			fakeClientSet.AppsClientset.PrependReactor("list", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, &listOfDC, nil
```
mik-dass: Same here like the other list example above

PR #518, opened 2018-06-07
Fixes odo update from git to binary/local and vice versa

mik-dass:
To achieve this, two functions have been added SetupForSupervisor and CleanupAfterSupervisor in occlient.go. Also some other functions are added in occlient to add and remove initContainers, volumeMounts and volumes to the deploymentConfig. Also updateDcAnnotations in occlient is added to update the annotations in the deploymentConfig.

Issue: https://github.com/redhat-developer/odo/issues/439
Signed-off-by: mdas <mrinald7@gmail.com>

PR #540, opened 2018-06-26
Updating th oc cluster version

mik-dass:
Signed-off-by: mdas mrinald7@gmail.com

This PR updates the OC Cluster from 3.7.2 to 3.9

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Wanted to get started? Follow the instructions below or our [Katacoda tutorial](
 
 ### Requirements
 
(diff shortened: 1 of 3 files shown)


kadel:
Why is this update needed?

mik-dass:
https://github.com/redhat-developer/odo/pull/518 won't run properly on v3.7.2 , so upgrading to the latest version

mik-dass:
The `oc update` implementation in #518 with v3.7.2 was not able to update the deploymentConfigs properly. Things like initContainers were not removed from the deploymentConfig even after the update using the client-go. Cause of the problem is unknown to me but I guess since we are using the latest version of client-go and using a older version of oc binary, so it might be causing the problem.

So for making the current implementation of #518 work, I think we need to update oc version too.

PR #541, opened 2018-06-26
Generate storage name if not provided

### LINKED ISSUE #504: We should be able to create a mount / storage without name being passed.
Similar to functionality with `url create`, we should be able to pass in / use storage without specifying a name (by default, it'll use the name of the current component).

surajnarwade:
Resolves #504

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -9,6 +9,7 @@ import (
 	"github.com/redhat-developer/odo/pkg/application"
 	"github.com/redhat-developer/odo/pkg/project"
 	"github.com/redhat-developer/odo/pkg/storage"
+	"github.com/redhat-developer/odo/pkg/util"
 	"github.com/spf13/cobra"
 )
 
@@ -37,17 +38,22 @@ var storageCreateCmd = &cobra.Command{
 	Example: `  # Create storage of size 1Gb to a component
   odo storage create mystorage --path=/opt/app-root/src/storage/ --size=1Gi
 	`,
-	Args: cobra.ExactArgs(1),
+	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		applicationName, err := application.GetCurrent(client)
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
-
-		_, err = storage.Create(client, args[0], storageSize, storagePath, componentName, applicationName)
+		var storageName string
+		if len(args) != 0 {
+			storageName = args[0]
+		} else {
+			storageName = componentName + "-" + util.GenerateRandomString(4)
+		}
+		_, err = storage.Create(client, storageName, storageSize, storagePath, componentName, applicationName)
 		checkError(err, "")
-		fmt.Printf("Added storage %v to %v\n", args[0], componentName)
+		fmt.Printf("Added storage %v to %v\n", storageName, componentName)
 	},
 }
 


cmd/storage.go
```diff
@@ -37,20 +40,37 @@ var storageCreateCmd = &cobra.Command{
...
-		_, err = storage.Create(client, args[0], storageSize, storagePath, componentName, applicationName)
+		}
+		_, err = storage.Create(client, storageName, storageSize, storagePath, componentName, applicationName)
 		checkError(err, "")
-		fmt.Printf("Added storage %v to %v\n", args[0], componentName)
+		fmt.Printf("Added storage %v to %v\n", storageName, componentName)
 	},
 }
 
+func RandStringRunes() string {
```
mik-dass: There is a already a random string generator in pkg/uitl/util.go https://github.com/redhat-developer/odo/blob/0cc71d145bdc80fad2a9501d27e0904cc17735d5/pkg/util/util.go#L37

Can we reuse it?

cmd/storage.go
```diff
@@ -37,20 +40,37 @@ var storageCreateCmd = &cobra.Command{
...
 	`,
-	Args: cobra.ExactArgs(1),
+	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		applicationName, err := application.GetCurrent(client)
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
+		currentComponentName, err := component.GetCurrent(client, applicationName, projectName)
```
mik-dass: We are retrieving the component name in the last line above, so is this line required?

cmd/storage.go
surajnarwade: thanks @mik-dass , I didn't noticed :(

cmd/storage.go
surajnarwade: yeah, not required :(

mik-dass:
If no component is created or set as active 
```
[mdas@localhost odo]$ ./odo storage create --path /data --size 100M
PersistentVolumeClaim "-oljp-app-pvc" is invalid: [metadata.name: Invalid value: "-oljp-app-pvc": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), metadata.labels: Invalid value: "-oljp": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]
```
Can we also add a check for the component name?

surajnarwade:
@mik-dass , PR for the the check is here #545

mik-dass:
@surajnarwade Thanks https://github.com/redhat-developer/odo/pull/545 is in now.

cmd/storage.go
```diff
@@ -37,17 +38,22 @@ var storageCreateCmd = &cobra.Command{
...
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 		componentName := getComponent(client, storageComponent, applicationName, projectName)
-
-		_, err = storage.Create(client, args[0], storageSize, storagePath, componentName, applicationName)
+		var storageName string
+		if len(args) != 0 {
+			storageName = args[0]
+		} else {
+			storageName = componentName + "-" + util.GenerateRandomString(4)
```
cdrage: @surajnarwade This LGTM. Is this the same way that OpenShift / Kubernetes is generating their random strings for their pods? We should replicate / make this similar.

surajnarwade: Kubernetes, OpenShift way is bit different, Since we already had `GenerateRandomString`. I reused it :)

surajnarwade:
Let's merge this :P

cdrage:
Yeah, this LGTM :+1:

PR #546, opened 2018-06-27
Nit picking code

surajnarwade:


cmd/list.go
```diff
@@ -25,7 +25,7 @@ var componentListCmd = &cobra.Command{
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 		currentComponent, err := component.GetCurrent(client, applicationName, projectName)
-
+		checkError(err, "")
```
kadel: why do you want to remove this check?

pkg/storage/storage.go
```diff
@@ -81,6 +81,9 @@ func Unmount(client *occlient.Client, storageName string, componentName string,
 	}
 
 	pvc, err := client.GetPVCFromName(pvcName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to get pvc from name: %v", pvcName)
```
cdrage: I think we should have it as "PersistentVolumeClaim" instead of just pvc

cdrage:
LGTM!

PR #551, opened 2018-06-28
add test for AddPVCToDeploymentConfig

syamgk:
add test for AddPVCToDeploymentConfig


syamgk:
as of now, this is failing on panic error for 
```
- Test case : dc without container
 -Test case : dc without container
``` 
merge https://github.com/redhat-developer/odo/pull/554 which will add validation for pointers before dereferencing and rerun tests

syamgk:
rebased

pkg/occlient/occlient_test.go
```diff
@@ -15,6 +15,100 @@ import (
...
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, fakeClientset := FakeNew()
+
+			fakeClientset.AppClientset.PrependReactor("update", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
```
cdrage: :+1:

pkg/occlient/occlient_test.go
```diff
@@ -15,6 +15,100 @@ import (
...
+				return true, nil, nil
```
mik-dass: We should return the updated dc instead of nil and verify that the dc was updated with the proper values as done in https://github.com/redhat-developer/odo/blob/e8e223b6edd6f046035ec1ad249448303fbf4b67/pkg/occlient/occlient_test.go#L1337

Also we should check that update was called on the correct dc by verifying against the name of the dc
https://github.com/redhat-developer/odo/blob/e8e223b6edd6f046035ec1ad249448303fbf4b67/pkg/occlient/occlient_test.go#L1323-L1326

syamgk: I agree that it is good to validate the values with which the function is being called but in this specific case there is a random string generation involved here https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L1333
so it won't be feasible.
will add validation for the number of action performed

syamgk: added validation for the presence of mounts and with passing volume names & rebased

now ready for review

pkg/occlient/occlient_test.go
```diff
@@ -99,6 +100,121 @@ func fakePodStatus(status corev1.PodPhase, podName string) *corev1.Pod {
...
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, fakeClientset := FakeNew()
+
+			fakeClientset.AppsClientset.PrependReactor("update", "deploymentconfigs", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, nil, nil
```
mik-dass: Also we should check if the update reactor was called with the correct dc name passed to the function like https://github.com/redhat-developer/odo/blob/e8e223b6edd6f046035ec1ad249448303fbf4b67/pkg/occlient/occlient_test.go#L1323-L1326 because we might end up calling update on the wrong dc.

pkg/occlient/occlient_test.go
```diff
@@ -99,6 +100,121 @@ func fakePodStatus(status corev1.PodPhase, podName string) *corev1.Pod {
...
+					if tt.args.path == updatedDc.Spec.Template.Spec.Containers[0].VolumeMounts[bb].MountPath {
+						found = true
+						if !strings.Contains(updatedDc.Spec.Template.Spec.Containers[0].VolumeMounts[bb].Name, tt.args.pvc) {
+							t.Errorf("pvc name not matching with the specived value got: %v, expected %v", updatedDc.Spec.Template.Spec.Containers[0].VolumeMounts[bb].Name, tt.args.pvc)
+						}
+					}
+				}
+				if found == false {
+					t.Errorf("expected Volume mount path %v not found", tt.args.path)
+				}
```
mik-dass: We should also have a test for this part of the code i.e the volumes, check the volume name as before and that the claimName is the same as desired
https://github.com/redhat-developer/odo/blob/4cd7e795bec7979b74e9856d95c8b18a1c663629/pkg/occlient/occlient.go#L1339-L1346

pkg/occlient/occlient_test.go
```diff
@@ -99,6 +100,121 @@ func fakePodStatus(status corev1.PodPhase, podName string) *corev1.Pod {
...
+
+			// Checks for number of actions performed in positive cases
+			if err == nil {
+				// Check for validating actions performed
+				if (len(fakeClientset.AppsClientset.Actions()) != 1) && (tt.wantErr != true) {
+					t.Errorf("expected 1 action in GetPVCFromName got: %v", fakeClientset.AppsClientset.Actions())
+				}
+				updatedDc := fakeClientset.AppsClientset.Actions()[0].(ktesting.UpdateAction).GetObject().(*appsv1.DeploymentConfig)
+				found := false
+				for bb := range updatedDc.Spec.Template.Spec.Containers[0].VolumeMounts {
```
mik-dass: Can we use reflect.DeepEqual() instead of looping?

pkg/occlient/occlient_test.go
syamgk: added validation for this

syamgk:
@cdrage Pr is updated and ready for review now

PR #558, opened 2018-07-03
Show tags / versions when doing odo catalog list

cdrage:
When using `odo catalog list` the tags will now be listed:

```
github.com/redhat-developer/odo  add-versioning ✗                                                                                                                                                                                                                                                                                                                    7d ⚑  ⍉
▶ ./odo catalog list
NAME        TAGS
dotnet      2.0,latest
httpd       2.4,latest
nodejs      0.10,4,6,8,latest
ruby        2.0,2.2,2.3,2.4,latest
wildfly     10.0,10.1,8.1,9.0,latest

```

### DIFF
--- a/pkg/catalog/catalog_test.go
+++ b/pkg/catalog/catalog_test.go
@@ -1,0 +1,82 @@
+package catalog
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/redhat-developer/odo/pkg/occlient"
+	"github.com/redhat-developer/odo/pkg/testingutil"
+	"k8s.io/apimachinery/pkg/runtime"
+	ktesting "k8s.io/client-go/testing"
+)
+
+func TestList(t *testing.T) {
+	type args struct {
+		name      string
+		namespace string
+		tags      []string
+	}
+	tests := []struct {
+		name     string
+		args     args
+		wantErr  bool
+		wantTags []string
+	}{
+		{
+			name: "Case 1: Valid image output with one tag",
+			args: args{
+				name:      "foobar",
+				namespace: "openshift",
+				tags:      []string{"latest"},
+			},
+			wantErr:  false,
+			wantTags: []string{"latest"},
+		},
+		{
+			name: "Case 2: Valid image output with multiple tags",
+			args: args{
+				name:      "foobar",
+				namespace: "openshift",
+				tags:      []string{"1.0.0", "1.0.1", "0.0.1", "latest"},
+			},
+			wantErr:  false,
+			wantTags: []string{"1.0.0", "1.0.1", "0.0.1", "latest"},
+		},
+		{
+			name: "Case 3: Invalid image output with no tags",
+			args: args{
+				name:      "foobar",
+				namespace: "foo",
+				tags:      []string{},
+			},
+			wantErr:  true,
+			wantTags: []string{},
+		},
+	}
+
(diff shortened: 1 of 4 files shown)


cdrage:
This will probably conflict with your PR @surajnarwade since we're modifying the same sections :+1:

surajnarwade:
yeah @cdrage , this PR looks good to me though :)

cdrage:
Ping @kadel @ashetty1  :)

surajnarwade:
@cdrage , now we dont have conflicts between this and service catalog PR since both are two different commands now

kadel:
Looks good, but there are no tests :-(

cdrage:
Hey @kadel so I started on writing tests and ironically @ashetty1 had already written some! His PR is here: https://github.com/redhat-developer/odo/pull/478

After https://github.com/redhat-developer/odo/pull/478 is merged, I'll be able to add a few tests to cover the listing of tags. I've already began rebasing this PR against https://github.com/redhat-developer/odo/pull/478

surajnarwade:
@cdrage , Since #478 is merged now ? can you please rebase and send so that we can merge this :)

cdrage:
Ready for review! @surajnarwade @kadel @ashetty1 @mik-dass @syamgk

pkg/catalog/catalog_test.go
```diff
@@ -0,0 +1,121 @@
...
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/redhat-developer/odo/pkg/occlient"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ktesting "k8s.io/client-go/testing"
+)
+
+// Function taken from occlient_test.go
+// fakeImageStream gets imagestream for the reactor
+func fakeImageStream(imageName string, namespace string, tags []string) *imagev1.ImageStream {
```
ashetty1: We should move these helper functions to a different file under `occlient` package and just call them from there. WDYT? We could take up that exercise post PR merge too. 

Cc: @kadel

cdrage: Yeah, I was trying to move them to `fakeclient.go` or making them public within `occlient_test.go` (which didn't work).

To be honest, I had to heavily modify these functions to work specifically for `catalog list` (since I'm passing in `tags []string` and modifying it. I would be more keen towards leaving them in the `catalog_test.go` file.

kadel: Those helper function should be moved to something like `testingutil` pkg and merged into just one function, so we can share it between different tests. 
Ideally in this PR :grimacing: If not then we should at least open new Issue so we don't forget to clean this.

cdrage: @kadel No worries, I'll move them to a testingutil pkg :)

cdrage:
@kadel @surajnarwade @syamgk @ashetty1 @syamgk 

Tests have been moved to a different package. Ready for review :+1:

syamgk:
LGTM

pkg/testingutil/imagestreams.go
```diff
@@ -0,0 +1,43 @@
+package testingutil
```
syamgk: wow! this way of separating function looks nice!

Issue #549, opened 2018-06-28
way too long info in PS1


kadel:
When there is not app or/and no component the PS1 shows, too detailed information.
This is suitable for when user run `odo component` but in PS1 we should have something more concise.

```
[There's no active application.
You can create one by running 'odo application create <name>'./No component is set as current]
```
I think that something like `[<no application>/<no component>]` should be enough.

cdrage:
@kadel I'll assign this to myself and fix it. It *should* be reverting to blank instead, may need for me to update the regex.

cdrage:
@kadel 

The problem is that "There's no active application." prompt as well as "No component is set as current" is not returning an error code (this should error out).

The current PS1 implementation checks to see if there is an error code of 0 or 1 outputted and goes from there.

I've implemented this fix (with a more detailed description) here: https://github.com/redhat-developer/odo/pull/572

kadel:
> The problem is that "There's no active application." prompt as well as "No component is set as current" is not returning an error code (this should error out).

It shouldn't error out. The state with no active app or component is not error state.

cdrage:
@kadel I'd disagree that "There's no active application" would be an error rather than a print / prompt. As a compromise, if we add `-q` could we agree to make it error out if no active application / no component was found so that the PS1 script operates correctly based on error-code?

If not, we could always compromise and add regex / string checking to the PS1 instead of checking for error code.

kadel:
> As a compromise, if we add -q could we agree to make it error out if no active application / no component was found so that the PS1 script operates correctly based on error-code?

I don't like it to be honest. No active app or component is not an error state. We shouldn't make  `odo component -q` error out in those cases. Detecting if commands return empty string is easy as checking its return code.

cdrage:
@kadel Alright, no problem. I'll implement this with string checking rather than error checking!

kadel:
@cdrage https://github.com/redhat-developer/odo/issues/550 should be fixed before you do that.

PR #572, opened 2018-07-09
Some outputs should error 1 not error 0

cdrage:
fmt.Printf errors out with checkout code 0. However, there are some
prompts (this component doens't exist, this application doesn't exist)
that error out with 0, they should error with checkout code 1 (using
fmt.Errorf).

This creates incorrect output for PS1 / `odo utils terminal` code as it
creates scenarios where there is a false error code output:

```sh
if [ $? != 0 ]; then
  APP="Unable to retrieve app"
else
  APP=$(echo $APP_GET | sed "s/The current application is: //g")
fi
```

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -81,7 +81,7 @@ var applicationGetCmd = &cobra.Command{
 			return
 		}
 		if app == "" {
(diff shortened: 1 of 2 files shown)


cdrage:
After discussion in #549 I'm going to close this (for now), but I'll still keep this branch just-in-case for reference.

Issue #520, opened 2018-06-11
revamp URL - Allow custom names for URL


kadel:
Url command is a bit odd right now.
It takes a component name as a positional argument, the rest of our commands have `--component` flag for that.
The positional argument should be optional, and it should be the name of the url, if not specified then the component name value should be used. Component for which url will be created should be specified by `--component` as described in https://github.com/redhat-developer/odo/issues/172

Example:
```
# create url for current component, the name of the url will be the same as the component name
$ odo url create 

# create url named myurl for the mycomponent
$ odo url create myurl --component mycomponent
```

TODO:
 - how to avoid host colisions (url names)

PR #582, opened 2018-07-12
Revamps the URL command and adds support for custom names for url.

mik-dass:
Issue: #520 
Signed-off-by: mdas mrinald7@gmail.com


mik-dass:
Closing because the branch name was wrong and instead raised https://github.com/redhat-developer/odo/pull/585

PR #584, opened 2018-07-13
Added check for deletion of current component

surajnarwade:


cmd/delete.go
```diff
@@ -42,6 +42,10 @@ var componentDeleteCmd = &cobra.Command{
 		// else, use the first argument and check to see if it exists
 		if len(args) == 0 {
 			componentName, err = component.GetCurrent(client, applicationName, projectName)
+			if componentName == "" {
```
mik-dass: Can we use the getComponent() function in cmdUtils, I guess it does the same thing

cdrage:
LGTM.

PR #585, opened 2018-07-13
Revamps the URL command and adds support for custom names for url.

mik-dass:
Issue: #520 

This revamps the URL command and adds support for custom names for url. It also adds a component flag to the create command for URL to specify the component. It also changes the behaviour of the odo delete command and now a component flag can be passed to delete the url from or else the current component is used.

### DIFF
--- a/cmd/url.go
+++ b/cmd/url.go
@@ -6,16 +6,15 @@ import (
 	"strings"
 
 	"github.com/redhat-developer/odo/pkg/application"
-	"github.com/redhat-developer/odo/pkg/component"
 	"github.com/redhat-developer/odo/pkg/project"
 	"github.com/redhat-developer/odo/pkg/url"
 	"github.com/spf13/cobra"
 )
 
 var (
-	urlListComponent   string
-	urlListApplication string
-	urlForceDeleteflag bool
+	urlComponent       string
+	urlApplication     string
+	urlForceDeleteFlag bool
 )
 
 var urlCmd = &cobra.Command{
@@ -40,8 +39,11 @@ The created URL can be used to access the specified component from outside the O
 	Example: `  # Create a URL for the current component.
   odo url create
 
-  # Create a URL for a specific component
-  odo url create mycomponent
+  # Create a URL with a specific name
+  odo url create example
+
+  # Create a URL with a specific name for component frontend
+  odo url create example --component frontend
 	`,
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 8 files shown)


mik-dass:
@cdrage Done

cmd/url.go
```diff
@@ -50,24 +53,52 @@ The created URL can be used to access the specified component from outside the O
...
+			app, err = application.GetCurrent(client)
+			checkError(err, "")
+		} else {
+			app = urlListApplication
+		}
+
+		if len(urlListComponent) == 0 {
+			componentName = getComponent(client, "", app, projectName)
+		} else {
+			componentName = urlListComponent
```
kadel: There is no check if this user provided component exists.
Because of that it is possible to create url in not existing components.
```
▶ odo url create myurl1 --component nonexisting
Adding URL to component: nonexisting
URL created for component: nonexisting

myurl1 - http://myurl1-app-odo.dev.rdu2c.fabric8.io
```

cmd/url.go
```diff
@@ -50,24 +53,52 @@ The created URL can be used to access the specified component from outside the O
 		checkError(err, "")
 		projectName := project.GetCurrent(client)
 
+		var app string
 		var componentName string
+		var urlName string
+
+		if urlListApplication == "" {
```
kadel: there is no  flag for `create` commands that uses  `urlListApplication`!

cmd/url.go
```diff
@@ -50,24 +53,52 @@ The created URL can be used to access the specified component from outside the O
...
+			app, err = application.GetCurrent(client)
+			checkError(err, "")
+		} else {
+			app = urlListApplication
+		}
+
+		if len(urlListComponent) == 0 {
```
kadel: why is this called `urlListComponent`? Are we reusing the same variable for flags in `list` and `create`?

mik-dass: yes

kadel: > yes, should I add a new one?

renaming should be fine. The current name is really misleading.

mik-dass:
@kadel Done

cmd/url.go
```diff
@@ -50,24 +53,61 @@ The created URL can be used to access the specified component from outside the O
...
+			checkError(err, "unable to check if the application exists or not")
+			if !exists {
+				fmt.Printf("The application %s does not exists in the project %s\n", urlApplication, projectName)
+				os.Exit(1)
+			}
+			app = urlApplication
+		}
+
+		if len(urlComponent) == 0 {
+			componentName = getComponent(client, "", app, projectName)
```
kadel: just call this as `getComponent(client, urlComponent, app, projectName)`
And you don't have to do any other checks (you can delete this whole `if` statement), existing check is alrady in that function

cmd/url.go
```diff
@@ -83,19 +123,45 @@ var urlDeleteCmd = &cobra.Command{
...
 		// Initialization
 		client := getOcClient()
+		projectName := project.GetCurrent(client)
 		applicationName, err := application.GetCurrent(client)
 		checkError(err, "")
+
+		var componentName string
+
+		if urlComponent == "" {
+			componentName = getComponent(client, "", applicationName, projectName)
```
kadel: the same comments as in above, you can remove this whole if statement, just by providing correct second argument to `getComponent`

cmd/url.go
```diff
@@ -114,17 +180,37 @@ var urlListCmd = &cobra.Command{
...
+			if !exists {
+				fmt.Printf("The application %s does not exists in the project %s\n", urlApplication, projectName)
+				os.Exit(1)
+			}
+			app = urlApplication
+		}
+
+		var componentName string
+		if len(urlComponent) == 0 {
+			componentName = getComponent(client, "", app, projectName)
```
kadel: again ;-)

mik-dass:
@kadel Done

kadel:
@mik-dass Please just add new commits, instead of squashing everything into one. It will make reviews much easier.

cmd/url.go
```diff
@@ -50,24 +52,50 @@ The created URL can be used to access the specified component from outside the O
...
 		}
 
+		exists, err := url.Exists(client, urlName, "", applicationName)
+
+		if exists {
+			fmt.Printf("The url %s already exists in the application: %s\n", urlName, applicationName)
+			os.Exit(1)
+		}
+
+		if urlComponent != "" {
```
kadel: you don't need this, line 71 already handles that

PR #587, opened 2018-07-16
Update PS1

cdrage:
Updates the PS1 output so that "<no application>/<no component>" is
outputted when nothing is detected, rather then the previous
long-string.

This also changes the comparison from error-code to string-checking.

Closes https://github.com/redhat-developer/odo/issues/549

### DIFF
--- a/cmd/terminal.go
+++ b/cmd/terminal.go
@@ -45,21 +45,19 @@ func TerminalGenerate(out io.Writer, cmd *cobra.Command, args []string) error {
 	var PS1 = `
 __odo_ps1() {
 
-    # Get current context
-    COMPONENT_GET=$(odo component get --skip-connection-check | sed "s/The current component is: //g")
-		if [ $? != 0 ]; then
-		COMPONENT="Unable to retrieve component"
-		else
-    COMPONENT=$(echo $COMPONENT_GET | sed "s/The current component is: //g")
+		# Get application context
+    APP=$(odo application get -q --skip-connection-check)
+
+		if [ "$APP" = "" ]; then
+		APP="<no application>"
 		fi
 
-		# Get application context
-    APP_GET=$(odo application get --skip-connection-check | sed "s/The current application is: //g")
-		if [ $? != 0 ]; then
-		APP="Unable to retrieve app"
-		else
-    APP=$(echo $APP_GET | sed "s/The current application is: //g")
-		fi
+    # Get current context
+    COMPONENT=$(odo component get -q --skip-connection-check)
+
+		if [ "$COMPONENT" = "" ]; then
+		COMPONENT="<no component>"
+    fi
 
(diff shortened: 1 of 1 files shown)


cdrage:
Ping @kadel @ashetty1 @surajnarwade

surajnarwade:
Tested, Looks good to me, green signal from me :)

cdrage:
@mik-dass @syamgk Want to test this too?

cmd/terminal.go
```diff
@@ -45,22 +45,20 @@ func TerminalGenerate(out io.Writer, cmd *cobra.Command, args []string) error {
...
-    COMPONENT=$(echo $COMPONENT_GET | sed "s/The current component is: //g")
-		fi
-
 		# Get application context
-    APP_GET=$(odo application get --skip-connection-check | sed "s/The current application is: //g")
-		if [ $? != 0 ]; then
-		APP="Unable to retrieve app"
-		else
-    APP=$(echo $APP_GET | sed "s/The current application is: //g")
+    APP=$(odo application get -q --skip-connection-check)
```
mik-dass: Is the `--skip-connection-check` required? Even if the cluster is down, it's showing me the app/component name

cmd/terminal.go
```diff
@@ -45,22 +45,20 @@ func TerminalGenerate(out io.Writer, cmd *cobra.Command, args []string) error {
...
+
+		if [ "$APP" = "" ]; then
+		APP="<no application>"
 		fi
 
+    # Get current context
+    COMPONENT=$(odo component get -q --skip-connection-check)
```
mik-dass: same as above

kadel:
@cdrage Is this fix for #549? 
If yes, can you please reference that issue in the description?
We are starting to get a lot of PRs and Issues and it is getting hard to keep track what solves what :-(

cmd/terminal.go
cdrage: Yes, it's required. If a connection check is used, each call will take 1+ more seconds.

cmd/terminal.go
cdrage: See other comment

cdrage:
@kadel Done!

cmd/terminal.go
mik-dass: But if the cluster is down, it shows me the last active component/app name

cdrage: @mik-dass That's intentional

kadel:
> Updates the PS1 output so that "/" is

isn't  this description a bit off? From the code it looks like "<no application>/<no component>" will be the output

kadel:
> isn't this description a bit off? From the code it looks like "<no application>/<no component>" will be the output

ok, I just noticed that it is just wrong markdown formatting in the descriptions ;-)  Newer mind ;-)

PR #590, opened 2018-07-19
Enable levelled logging

### LINKED ISSUE #189: ocdev output should follow verbosity rules in oc,kubectl with --log-level
I started a cluster (minishift) that I had stopped for some time, so my credentials were expired. It gave me a bad error, and changed the context to -q.

```
$ ocdev app list
unable to list applications: unable to get current project info: command: [/Users/jmorales/repositories/jorgemoralespou/openshift-evangelists/oc-login.git/oc-origin project] failed to run:
error: You must be logged in to the server (the server has asked for the client to provide credentials (get projects.project.openshift.io -q))
: exit status 1
```

If I want verbose output, I see the flag verbose, which gives a lot of output.

In oc, minishift, kubectl and other go binaries it seems standar to follow logging by levels, with --log-level=(0-10)

```
--loglevel=0: Set the level of log output (0-10)
```

This should then yield by default on such above error:

```
$ ocdev app list
Unable to list applications
error: You must be logged in to the server (the server has asked for the client to provide credentials  (get projects.project.openshift.io -q))
```

anmolbabu:
Aim:
ocdev output should follow verbosity rules in oc,kubectl
Ref: https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-output-verbosity-and-debugging

Task:
To accomplish the above PR, as suggested by:
https://github.com/redhat-developer/odo/issues/189#issuecomment-404124060,
The task will be accomplished in parts as described under:
(a) Add glog as a dependency and add a utility wrapper around it that abstracts
    glog specifics and eases the task of logging.
(b) Switch from usage of github.com/sirupsen/logrus to github.com/golang/glog
(c) Add more log statements with appropriate log levels to indicate, for what user
    chosen verbosity level the statements need to be logged.

Details:

Task (a), above adds a singleton utility which can be intialised with
GlobalVerbosity level and then subsequently, using this one time intialised logger
instance, one can log the messages via 2 ways:
1. Logf
2. Log
Parameters common to both of these are:
severity: It is of type Loglevel and is defined as a type(alias to string) in the utility
          whose value can be any among "INFO", "INFOF", "WARNING", "WARNINGF", "ERROR", "ERRORF"
          where, levels with postfix "F" indicate formatted string logging.
params: It is a map[string]interface{} that can be used to pass any optional parameters to logging
        functions like verbosity level for info level logs. Only those statements that are logged
        at verbosity less than or equal to the GlobalVerbosity level will appear in the log statements.
        Please check https://godoc.org/github.com/golang/glog#V for more details on verbosity.

and the 2 functions Logf and Log differ only in the other 2 params namely:
1. Logf:
        i. format: message with format specifiers
       ii. args: format args
2. Log:
        i. message: string message to be logged.
Accordingly, the PR will be composed of 3 commits

fixes #189


anmolbabu:
!!WORK IN PROGRESS!!
PLEASE DO NOT MERGE

cdrage:
@anmolbabu 

When you create a PR, please add a descriptive message to the commit message as well as the GitHub description. Explaining what is happening, steps to reproduce (if any) and what it fixes.

cdrage:
You can modify your commit message using `git commit --amend`

anmolbabu:
> @anmolbabu
When you create a PR, please add a descriptive message to the commit message as well as the GitHub description. Explaining what is happening, steps to reproduce (if any) and what it fixes.
You can modify your commit message using git commit --amend

Ack

anmolbabu:
!! To be verified !!
@cdrage @kadel Please review

cmd/root.go
```diff
@@ -135,7 +139,12 @@ func getLatestReleaseInfo(info chan<- string) {
 	if err != nil {
 		// The error is intentionally not being handled because we don't want
 		// to stop the execution of the program because of this failure
-		log.Debugf("Error checking if newer odo release is available: %v", err)
+		log.GetInstance().Logf(
```
cdrage: This is kind of ugly, is there a cleaner way of doing this?

kadel: This looks overcomplicated.
We should be able to do just `glog.V(4).Infof(....`

kadel:
Hi @anmolbabu 
this should be much less complicated that what you did :-)

You can really just replace `log "github.com/sirupsen/logrus"` with `"github.com/golang/glog"`
 
Than replace `log.Debugf` with  `glog.V(4).Infof` (a per [Kubectl output verbosity and debugging
](https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-output-verbosity-and-debugging) log level 4 is debug level) and `log.Errorf` with `glog.Errorf`.

Then you will just need to update verbose flag. (remove old `--verbose` flag and replace it with glog's verbose settings (you can find that in documentation or google will help you with that) ;-) )

anmolbabu:
@kadel The reason why I separated out the logger to a utility is, is to minimize the changes to the utility if at any later point in time we decide to move to another logger utility like we moved now..

I personally feel that basic and/or framework level utilities like logger are better exposed via generic interfaces to accommodate the possibilities of changes in their choices..  And separting out the library specifics to a separate location can at any later point in time be easily(least area of influence due to the changes) converted to a pluggable module without which any such change would warrant change to almost every source file..

Cc: @cdrage 

@cdrage While I agree with you that, GetInstance call is probably adding an unintended extra level/layer of calls for every message to be logged and I should probably look at an alternative way of achieving a singleton like behavior that avoids these, like in the links that you shared, I feel most of the parameters, passed to the other parts of the logging call are probably unavoidable(in my opinion),

kadel:
> @kadel The reason why I separated out the logger to a utility is, is to minimize the changes to the utility if at any later point in time we decide to move to another logger utility like we moved now..
> 
> I personally feel that basic and/or framework level utilities like logger are better exposed via generic interfaces to accommodate the possibilities of changes in their choices.. And separting out the library specifics to a separate location can at any later point in time be easily(least area of influence due to the changes) converted to a pluggable module without which any such change would warrant change to almost every source file..

Problem with this is that it creates an unnecessary layer of abstraction that can be limiting. 
Logging is usually simple, and if it changes are needed, a simple search and replaces is usually good enough (like in this case). 

I would also say that if using glog directly is good for kubectl and kuberetes it will be definitely good for us ;)

anmolbabu:

cmd/root.go
```diff
@@ -124,7 +129,7 @@ func init() {
 	// will be global for your application.
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.odo.yaml)")
 
-	rootCmd.PersistentFlags().BoolVarP(&GlobalVerbose, "verbose", "v", false, "Verbose output")
+	rootCmd.PersistentFlags().StringVarP(&GlobalVerbose, "verbose", "v", "", "Verbose output")
```
surajnarwade: description: "log level for V logs"
default value: 0

surajnarwade: can we have `IntVarP` here, instead of taking string input ?

anmolbabu: Need it as a string as in line 105.

But yes, I think making it int and using IntVarP would provide additional advantage of additional pre-validating it as an int and anyway int to string is anyway doable at least additional cost :)

cdrage:
FYI your first commit message is messed up. It just says: `Aim: `. You need to update the title.

cmd/root.go
```diff
@@ -91,16 +93,20 @@ Find more information at https://github.com/redhat-developer/odo`,
...
 	},
 }
 
+func setFlags() {
+	flag.Usage = func() {
+		flag.PrintDefaults()
+		os.Exit(2)
+	}
+	flag.CommandLine.Set("logtostderr", "true")
+	flag.CommandLine.Set("v", strconv.Itoa(GlobalVerbose))
```
kadel: this looks a bit strange, we are setting goflags manualy?

anmolbabu:
@cdrage Done

cmd/root.go
anmolbabu: I Simply followed: 
https://gist.github.com/heatxsink/7221ebe499b0767d4784

and I observed that glog expects those params on cli so found this as a way to acieve it programmatic-ally.

kadel: Yes, but in that example he never sets flags "manualy" ,like this.
What about other glog flags like `-log_dir` or `-vmodule`?

kadel: @anmolbabu Is there a way to allow specification of all glog flags?

Have you tried this approach https://flowerinthenight.com/blog/2017/12/01/golang-cobra-glog ?

kadel: There is also section in pflags docs that might help: https://github.com/spf13/pflag#supporting-go-flags-when-using-pflag

anmolbabu: Ack done this time as per the https://github.com/openshift/origin/blob/master/tools/depcheck/pkg/cmd/cmd.go#L29

cmd/root.go
anmolbabu: @cdrage No the glog uses integer verbosity levels to indicate different levels of verbosity of info

cmd/root.go
anmolbabu: @kadel 
1. GlobalVerbose is not used anywhere now as verbosity setting is 1 time global setting and thereafter each log message carries a verbosity attached inorder for glog to know if the message needs to be logged for the user chosen verbosity..
2. glog doesn't recognize `--verbosity=4` and only understands `-v 4`

Because of 1, I removed GlobalVerbose and let glog handle the verbosity setting directly from cli which means because of 2, `--verbosity=x` can't be used any more only `-v x`  . If we want to preserve even `--verbosity=x` form of flag I'll need to workaround glog by parsing the flag(like we used to do before) and then setting it manually as a flag to glog.

What do you suggest would be good to have?

cmd/root.go
```diff
@@ -124,8 +118,9 @@ func init() {
 	// will be global for your application.
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.odo.yaml)")
 
-	rootCmd.PersistentFlags().BoolVarP(&GlobalVerbose, "verbose", "v", false, "Verbose output")
 	rootCmd.PersistentFlags().BoolVar(&GlobalConnectionCheck, "skip-connection-check", false, "Skip cluster check")
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.CommandLine.Set("logtostderr", "true")
```
kadel: Why are we setting this?
Description for `logtostderr` is `log to standard error instead of files (default false)` It is confusing if the message says that default is false and we are changing it here to true.

anmolbabu: It doesn't work if I remove this line..
Doesn't log anything to console

anmolbabu:
@tkral rebased

cmd/root.go
anmolbabu: Even openshift seems to be doing this as in https://github.com/openshift/origin/blob/master/tools/depcheck/pkg/cmd/cmd.go#L30

cmd/delete.go
```diff
@@ -26,8 +26,8 @@ var componentDeleteCmd = &cobra.Command{
 	`,
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		log.Debugf("component delete called")
-		log.Debugf("args: %#v", strings.Join(args, " "))
+		glog.V(4).Infof("component delete called")
```
cdrage: Should we document this at all anywhere for future devs?

cdrage: Otherwise, this LGTM.

cdrage:
LGTM. @surajnarwade care to have one last look / approve? Then we can merge.

cmd/delete.go
anmolbabu: Done raised https://github.com/redhat-developer/odo/pull/673

PR #593, opened 2018-07-23
Change version checker to use rawURL instead of calling API

syamgk:
Modify getLatestReleaseTag function to
fetch version using raw.github* URL

### DIFF
--- a/pkg/notify/notify.go
+++ b/pkg/notify/notify.go
@@ -1,19 +1,17 @@
 package notify
 
 import (
-	"context"
+	"io/ioutil"
+	"net/http"
 	"strings"
 
 	"github.com/blang/semver"
-	"github.com/google/go-github/github"
 	"github.com/pkg/errors"
 )
 
 const (
-	// project GitHub organization name
-	ghorg = "redhat-developer"
-	// project GitHub repository name
-	ghrepo = "odo"
+	// URL to fetch latest version number
+	VersionFetchURL = "https://raw.githubusercontent.com/redhat-developer/odo/master/build/VERSION"
 	// URL of the installation shell script
 	InstallScriptURL = "https://raw.githubusercontent.com/redhat-developer/odo/master/scripts/install.sh"
 )
@@ -21,15 +19,20 @@ const (
 // getLatestReleaseTag polls odo's upstream GitHub repository to get the
 // tag of the latest release
 func getLatestReleaseTag() (string, error) {
-	client := github.NewClient(nil)
-	release, response, err := client.Repositories.GetLatestRelease(context.Background(), ghorg, ghrepo)
-	if response != nil {
-		defer response.Body.Close()
-	}
+
+	resp, err := http.Get(VersionFetchURL)
 	if err != nil {
 		return "", errors.Wrap(err, "error getting latest release")
 	}
-	return *release.TagName, nil
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "error getting latest release")
+	}
+	version_string := string(body)
+	version_string = strings.TrimSuffix(version_string, "\n")
+
+	return version_string, nil
 }
 
 // CheckLatestReleaseTag returns the latest release tag if a newer latest
--- a/pkg/notify/notify.go
+++ b/pkg/notify/notify.go
@@ -29,10 +29,8 @@ func getLatestReleaseTag() (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "error getting latest release")
 	}
-	version_string := string(body)
-	version_string = strings.TrimSuffix(version_string, "\n")
 
-	return version_string, nil
+	return strings.TrimSuffix(string(body), "\n"), nil
 }
 
 // CheckLatestReleaseTag returns the latest release tag if a newer latest
--- a/pkg/notify/notify_test.go
+++ b/pkg/notify/notify_test.go
@@ -1,0 +1,42 @@
+package notify
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/go-github/github"
+	"reflect"
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


cdrage:
I don't know about this. The API implementation seems a lot cleaner than using raw URL's. APIs would (IMO) be more stable than relying on the raw URL.

How big is this library? I think we should just keep it in.

kadel:
> I don't know about this. The API implementation seems a lot cleaner than using raw URL's. APIs would (IMO) be more stable than relying on the raw URL.
> 
> How big is this library? I think we should just keep it in.

the problem is not about the library. Unauthenticated API calls are pretty heavily limited.  (60 requests per hour per IP) This is a big problem on a corporate network with NAT.

kadel:
> can we remove version from version.go and call that file instead of doing string magic on version.go
> Reference: #122 (comment)

It might be better to have two separate values.  Kepp version.go as it is and create new file called `.latest-release` that will contain just a number of latest released version. Minikube does something similar but they host this [file](https://storage.googleapis.com/minikube/releases.json) in on GCloud and have all history of all releases there. https://github.com/kubernetes/minikube/blob/232080ae0cbcf9cb9a388eb76cc11cf6884e19c0/pkg/minikube/constants/constants.go#L112

Otherwise, there might be a problem when doing a release:
The current release process is:
1) New PR is opened with updated `versions.go`, links and docs
2) PR is merged
3) New git tag is created to mark the version
4) Travis builds binaries and uploads them to the Github release page

if we use the value from `versions.go` than the update checker will report that new version is available directly after step 2. "PR is merged"  But it can take 30mins and more before all checks and tests are done and new release is created with binaries on github release page.

surajnarwade:
I agree with @kadel 
@syamgk instead of doing string magic on `version.go`, use `.release` or whatever file name which consist of only version

syamgk:
@surajnarwade yes that sounds good, am adding a file for tracking the latest version in PR https://github.com/redhat-developer/odo/pull/613

surajnarwade:
@syamgk I can't build binary from source:

```
go build -o odo ./main.go
# github.com/redhat-developer/odo/pkg/notify
pkg/notify/notify.go:34:10: cannot use s[0] (type byte) as type string in return argument
```

pkg/notify/notify.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	// URL to fetch latest version number
-	VersionFetchURL = "https://raw.githubusercontent.com/redhat-developer/odo/master/cmd/version.go"
+	VersionFetchURL = "https://raw.githubusercontent.com/redhat-developer/odo/master/build/VERSION"
```
cdrage: :+1:

surajnarwade:
@syamgk,

```
$ cd odo-example/nodejs-ex/
[snarwade@redsuraj nodejs-ex]$ 
[snarwade@redsuraj nodejs-ex]$ odo create nodejs -v4
I0829 17:26:39.100469   23383 create.go:58] Component create called with args: "nodejs", flags: binary=, git=, local=
I0829 17:26:39.109923   23383 occlient.go:244] Trying to connect to server 192.168.42.169:8443 - %!v(MISSING)
I0829 17:26:39.110299   23383 occlient.go:250] Server https://192.168.42.169:8443 is up
I0829 17:26:39.125867   23383 occlient.go:227] isLoggedIn err:  <nil> 
 output: "developer"
I0829 17:26:39.177402   23383 application.go:171] Setting application app as current.
I0829 17:26:39.199608   23383 catalog.go:127] Found builder images: [{dotnet [2.0 latest]} {httpd [2.4 latest]} {nginx [1.10 1.12 1.8 latest]} {nodejs [0.10 4 6 8 latest]} {perl [5.16 5.20 5.24 latest]} {php [5.5 5.6 7.0 7.1 latest]} {python [2.7 3.3 3.4 3.5 3.6 latest]} {ruby [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I0829 17:26:39.216273   23383 catalog.go:127] Found builder images: [{dotnet [2.0 latest]} {httpd [2.4 latest]} {nginx [1.10 1.12 1.8 latest]} {nodejs [0.10 4 6 8 latest]} {perl [5.16 5.20 5.24 latest]} {php [5.5 5.6 7.0 7.1 latest]} {python [2.7 3.3 3.4 3.5 3.6 latest]} {ruby [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I0829 17:26:39.219208   23383 occlient.go:384] Checking for exact match of builderImage with ImageStream
I0829 17:26:39.229195   23383 occlient.go:395] Found exact image tag match for nodejs:latest
I0829 17:26:39.329618   23383 root.go:133] Error checking if newer odo release is available: unable to make semver from the latest release tag: 0.0.10
```

can you please look into this line:

```
I0829 17:26:39.329618   23383 root.go:133] Error checking if newer odo release is available: unable to make semver from the latest release tag: 0.0.10
```

syamgk:
squashed and ready to merge now!

pkg/notify/notify.go
```diff
@@ -1,35 +1,38 @@
...
+		return "", errors.Wrap(err, "error getting latest release")
 	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return "", errors.Wrap(err, "error getting latest release")
 	}
-	return *release.TagName, nil
+	version_string := string(body)
+	version_string = strings.TrimSuffix(version_string, "\n")
```
cdrage: this can be refactored to:

```go
return strings.TrimSuffix(string(body), "\n"), nil
```

cdrage:
Thanks for writing the test. If you can please re-base to latest master, that'd be great! After all tests have passed, this LGTM / can be merged.

PR #595, opened 2018-07-23
Fix #594: add binary definition to build source

### LINKED ISSUE #594: BuildSource is set to git even when gitUrl is not passed
Looks like PR #574 has introduced a bug in `NewAppS2I` function by not dealing with a scenario where `gitUrl` is empty: https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L539. Because of this change, `Source.Type` value is `git` even when `gitUrl` is empty.

syamgk:
add binary definition to build source
fix issue #594


pkg/occlient/occlient.go
```diff
@@ -537,6 +537,10 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 	}
 
 	// if gitUrl set change buildSource to git and use given repo
+	if gitUrl == "" {
+		return errors.Wrap(err, "unable to create buildSource with empty gitUrl")
```
kadel: What `err` are you wrapping here?

pkg/occlient/occlient.go
```diff
@@ -507,7 +507,7 @@ func getAppRootVolumeName(dcName string) string {
 }
 
 // NewAppS2I create new application using S2I
-// gitUrl is the url of the git repo
+// if gitUrl is ""  then it throws error otherwise uses gitUrl as buildSource
```
cdrage: You may as well move this comment, as you comment on this on line 540.

pkg/occlient/occlient.go
```diff
@@ -537,6 +537,10 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 	}
 
 	// if gitUrl set change buildSource to git and use given repo
```
cdrage: maybe change this comment in this PR.

`if gitUrl is not set, error out.`

syamgk:
rebased

Issue #552, opened 2018-06-28
Add option to allows user to specify port when component is created


kadel:
## Problem statement
Currently when `odo create ` odo inspect builder image and checks what ports are exposed.
This is just a metadata in the container image. 

User can configure its application to listen on a different port, or different port can be used as default in some frameworks. Then the component will not work properly as we create Service that points only to the ports that are specified in the image.

Example of this happening is #468 

## Solution
Add new optional flag to `odo create`. `--port` this flag could be specified multiple times.

Example:
```
odo create nodejs frontend --port 3000 --port 3030
```
The above command should create a component as we do right now, but the Service that is generated will be for ports 3000 and 3030, and there will be no port detection from an image.

If there is no `--port` flag, then the behavior remains the same. Odo will detect ports from the image.
The only thing that should change here, is that we should display a message saying what ports are going to be used.

This should solve #468

mik-dass:
closing #596

PR #596, opened 2018-07-23
Adds option to allow users to specify port when component is created

mik-dass:
Issue : #552 

Adds option to allow users to specify port when component is created and the flag is used as 
```
odo create nodejs --ports 8080/tcp,8081,9100/udp
```
 If the ports flag is not used than the ports, that are exposed by the given image, is used.

It adds a new function getContainerPortsFromString which takes a string array containing the port string e.g 8080/tcp,8081,9090/udp and returns the ContainerPorts or error. If no protocol is passed, tcp is considered as default.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -422,16 +422,25 @@ func getAppRootVolumeName(dcName string) string {
 
 // NewAppS2I create new application using S2I
 // gitUrl is the url of the git repo
-func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string) error {
+// inputPorts is the array containing the string port values
+func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string, inputPorts []string) error {
 
 	imageName, imageTag, _, err := parseImageName(builderImage)
 	if err != nil {
 		return errors.Wrap(err, "unable to create new s2i git build ")
 	}
 
-	containerPorts, err := c.GetExposedPorts(imageName, imageTag)
-	if err != nil {
-		return errors.Wrapf(err, "unable to exposed ports for %s:%s", imageName, imageTag)
+	var containerPorts []corev1.ContainerPort
+	if len(inputPorts) <= 0 {
+		containerPorts, err = c.GetExposedPorts(imageName, imageTag)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get exposed ports for %s:%s", imageName, imageTag)
+		}
+	} else {
+		containerPorts, err = getContainerPortsFromStrings(inputPorts)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get container Ports from %v", inputPorts)
+		}
 	}
 
 	// ObjectMetadata are the same for all generated objects
@@ -568,15 +577,24 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
 // BootstrapSupervisoredS2I uses s2i to inject Supervisor into builder image.
 // Supervisor keeps pod running (runs as pid1), so you it is possible to trigger assembly script inside running pod,
 // and than restart application using Supervisor without need to restart whole container.
-func (c *Client) BootstrapSupervisoredS2I(name string, builderImage string, labels map[string]string, annotations map[string]string) error {
+// inputPorts is the array containing the string port values
+func (c *Client) BootstrapSupervisoredS2I(name string, builderImage string, labels map[string]string, annotations map[string]string, inputPorts []string) error {
 	imageName, imageTag, _, err := parseImageName(builderImage)
 	if err != nil {
 		return errors.Wrap(err, "unable to create new s2i git build ")
 	}
 
-	containerPorts, err := c.GetExposedPorts(imageName, imageTag)
-	if err != nil {
-		return errors.Wrapf(err, "unable to exposed ports for %s:%s", imageName, imageTag)
+	var containerPorts []corev1.ContainerPort
+	if len(inputPorts) <= 0 {
+		containerPorts, err = c.GetExposedPorts(imageName, imageTag)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get exposed ports for %s:%s", imageName, imageTag)
+		}
+	} else {
+		containerPorts, err = getContainerPortsFromStrings(inputPorts)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get container Ports from %v", inputPorts)
+		}
 	}
 
 	// ObjectMetadata are the same for all generated objects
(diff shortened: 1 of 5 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -1437,3 +1438,93 @@ func TestUpdateBuildConfig(t *testing.T) {
...
+		ports          []string
+		containerPorts []corev1.ContainerPort
+		wantErr        bool
+	}{
+		{
+			name:  "with normal port values and normal protocol values in lowercase",
+			ports: []string{"8080/tcp", "9090/udp"},
+			containerPorts: []corev1.ContainerPort{
+				{
+					Name:          fmt.Sprintf("%d-%s", 8080, strings.ToLower(string(corev1.ProtocolTCP))),
```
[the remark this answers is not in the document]
mik-dass: @kadel But I guess this line does generate the string "8080-tcp"

tests/e2e/e2e_test.go
```diff
@@ -273,6 +273,25 @@ var _ = Describe("odoe2e", func() {
 				cmpList := runCmd("odo list")
 				Expect(cmpList).To(ContainSubstring("wildfly"))
 			})
+
+			It("should be able to create git component with required ports", func() {
+				runCmd("odo create nodejs nodejs-git --git https://github.com/openshift/nodejs-ex --ports 8080/tcp,9100/udp")
```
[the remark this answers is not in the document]
mik-dass: I guess it won't make any difference as the local component will also use the supervisor git repo. Also we don't have a component test for a git component, so I added the git component e2e test here

pkg/occlient/occlient_test.go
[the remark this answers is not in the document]
kadel: yes, just plain string "8080-tcp" is simpler and more obvious.
Why complicated it like this?

tests/e2e/e2e_test.go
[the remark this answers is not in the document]
kadel: We should have a separate test for git component. Your test doesn't verify that git component was actually deployed, so it is not useful as a test for git component.

pkg/occlient/occlient_test.go
```diff
@@ -1437,3 +1438,93 @@ func TestUpdateBuildConfig(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			ports, err := getContainerPortsFromStrings(tt.ports)
+			if err == nil && !tt.wantErr {
+				if !reflect.DeepEqual(tt.containerPorts, ports) {
+					t.Errorf("the ports are not matching, expected %#v, got %#v", tt.containerPorts, ports)
+				}
+			} else if err == nil && tt.wantErr {
+				t.Error("test failed expected to get error but no error was returned")
+			} else if err != nil && !tt.wantErr {
+				t.Errorf("test failed, expected: no error, got %s", "error: "+err.Error())
```
kadel: `("test failed, expected: no error, got %v, err")`

pkg/occlient/occlient_test.go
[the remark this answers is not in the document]
mik-dass: I was thinking of generating the string the same way it's generated in the actual function.

tests/e2e/e2e_test.go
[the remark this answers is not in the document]
mik-dass: Should I add a test to check for the git component in this part itself or separate the two? If I add the check here, I think it will be much faster

kadel:
> Should we add some documentation to this (even if it's a small note specifying that we're able to define specific ports?)

Where would you like to document this?

mik-dass:
@cdrage @kadel I have a added an example for the component create process with the ports on top of the cmd create code.

pkg/occlient/occlient_test.go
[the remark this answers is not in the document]
kadel: Why? 
It really doesn't make sense to do it like this, it just makes it harder to read.

tests/e2e/e2e_test.go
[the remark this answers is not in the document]
kadel: it should be separate test

pkg/occlient/occlient.go
```diff
@@ -562,15 +571,24 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
...
 	imageName, imageTag, _, err := parseImageName(builderImage)
 	if err != nil {
 		return errors.Wrap(err, "unable to create new s2i git build ")
 	}
 
-	containerPorts, err := c.GetExposedPorts(imageName, imageTag)
-	if err != nil {
-		return errors.Wrapf(err, "unable to exposed ports for %s:%s", imageName, imageTag)
+	var containerPorts []corev1.ContainerPort
+	if len(inputPorts) <= 0 {
```
cdrage: just add `==` not `<=`

pkg/occlient/occlient.go
```diff
@@ -562,15 +571,24 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
...
+		containerPorts, err = c.GetExposedPorts(imageName, imageTag)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get exposed ports for %s:%s", imageName, imageTag)
+		}
+	} else {
+		containerPorts, err = getContainerPortsFromStrings(inputPorts)
+		if err != nil {
+			return errors.Wrapf(err, "unable to get container Ports from %v", inputPorts)
```
cdrage: `ports`

pkg/occlient/occlient.go
```diff
@@ -1864,3 +1882,43 @@ func (c *Client) UpdatePVCLabels(pvc *corev1.PersistentVolumeClaim, labels map[s
 	}
 	return nil
 }
+
+// getContainerPortsFromStrings generates ContainerPort values from the array of string port values
+// ports is the array containing the string port values
+func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error) {
```
cdrage: :+1:

pkg/occlient/occlient_test.go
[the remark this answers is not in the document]
cdrage: I agree with @kadel could we just change this to "8080-tcp" or some "foobar" style name? No need for fmt.Sprintf

mik-dass:
@cdrage Done

pkg/occlient/occlient_test.go
```diff
@@ -2080,3 +2121,86 @@ func Test_getContainerPortsFromStrings(t *testing.T) {
...
+					Protocol:      corev1.ProtocolTCP,
+				},
+				{
+					Name:          "9100-udp",
+					ContainerPort: 9100,
+					Protocol:      corev1.ProtocolUDP,
+				},
+			},
+		},
+	}
```
ashetty1: We'll need another case with invalid values

ashetty1: We'll need another case with invalid values.

kadel: sadly we can't, as the fake client doesn't do any validations.

mik-dass: @ashetty1 I can't think of anything, can you give an example for a invalid case?

ashetty1: @kadel @mik-dass empty values validation?

mik-dass: @ashetty1 I tried but it didn't work for me

ashetty1: Nevertheless valid case, no?

PR #602, opened 2018-07-25
Run travis test using go 1.10

kadel:


kadel:
Errors should be fixed in https://github.com/redhat-developer/odo/pull/605

cdrage:
Tests failed, odd. Did Go change how tests were done in 1.10?

kadel:
> Tests failed, odd. Did Go change how tests were done in 1.10?

It looks like some additional checks were added to check if type corresponds formatting string.

PR #613, opened 2018-08-02
Add a file to track latest release version

syamgk:
add a file to track latest release version
as mentioned in https://github.com/redhat-developer/odo/pull/593#issuecomment-409173415


scripts/.latest-release
```diff
@@ -0,0 +1 @@
+v0.0.8
```
kadel: v0.0.9

surajnarwade:
@syamgk , as per @kadel suggestion, look into `scripts/bump-version.sh`
we should not manually edit this kind of version files

syamgk:
@kadel  @surajnarwade  as per https://github.com/redhat-developer/odo/pull/593#issuecomment-409173415
 what I understood is that bump-version.sh will change the version before binary becomes available
Is that correct?

kadel:
Good point @syamgk  you are right. 
We will have to update this as the last step in the release process. Can you please update development.md docs and add this as the step in the release process? It might be also worth adding a message/reminder to the bump-version.sh script, I believe that something similar is already there regarding homebrew package.

kadel:
Ah, I didn't notice that it is  `scripts/.latest-release`. Shouldn't it be better if this would be in root directory (`.latest-release`)?

scripts/.latest-release
```diff
@@ -0,0 +1 @@
+v0.0.9
```
cdrage: Maybe instead using a `build` dir within the root similar to https://github.com/kubernetes/kompose/blob/master/build/VERSION which will contain a RELEASE.md on instructions how to release?

@kadel

build/VERSION
```diff
@@ -0,0 +1 @@
+v0.0.9
```
cdrage: need to remove the `v`

cdrage:
2 LGTM's. Merging.

PR #615, opened 2018-08-02
Added the port flag to the URL create command.

mik-dass:
fixes https://github.com/redhat-developer/odo/issues/553

This PR adds the port flag to the URL create command. The 'odo url list' was modified to display the port of the url and the display format was changed to a more tabular form.

Also added function GetComponentServicePortNumbers to pkg/url package to expose the routes exposed by the service.


cmd/url.go
```diff
@@ -87,8 +89,24 @@ The created URL can be used to access the specified component from outside the O
 			os.Exit(1)
 		}
 
+		componentPorts, err := url.GetComponentServicePortNumbers(client, componentName, applicationName)
+		checkError(err, "unable to get component exposed ports")
+
+		var portFound bool
+
+		for _, port := range componentPorts {
+			if urlPort == port {
```
anmolbabu: Is flag corresponding to urlPort mandatory?
Please correct me if I am wrong as I might be missing something

anmolbabu: Is flag corresponding to urlPort mandatory? It seems so from line 103 below(https://github.com/redhat-developer/odo/pull/615/files#diff-626187d5d8d29a6936cc1b5876f7edd9R103)
As I can see from https://github.com/redhat-developer/odo/issues/553 , this is required for cases where a component has multiple ports exposed and to be to correctly choose which port to expose...
In which case, should we have a default no flag required case when there's only 1 port the component has and also accordingly either error out if flag not passed even though there are multiple ports or retain the current round robin choice of port if flag is not passed?
Please correct me if I am wrong as I might be missing something

pkg/url/url_test.go
```diff
@@ -247,3 +266,121 @@ func TestExists(t *testing.T) {
...
+									Port: 9100,
+								},
+							},
+						},
+					},
+				},
+			},
+			wantedPorts: []int{8080, 9100},
+			wantErr:     false,
+		},
```
anmolbabu: Please also add error Case

mik-dass: I don't think that a error case will be possible in this case as the error is thrown by GetServicesFromSelector() which calls List() and we are mocking it through the fake prependReactor

cmd/url.go
mik-dass: Yes I have made the `--port` flag to avoid the round robin problem stated in the issue. But incase of 1 port service component, I think we can retrieve the port exposed and make it the default if no port is passed. @kadel Should I make the port flag completely mandatory?

pkg/url/url_test.go
anmolbabu: Hmm, so there aren't any validations in GetComponentServicePortNumbers.
Ok so sounds good

cmd/url.go
anmolbabu: > Yes I have made the --port flag to avoid the round robin problem stated in the issue.

Yes

> But incase of 1 port service component, I think we can retrieve the port exposed and make it the default if no port is passed

Exactly what I meant, but it doesn't appear to be the case looking at: 
https://github.com/redhat-developer/odo/pull/615/files/b2b39f5a74142e815d8d25b91823faf48a2d93c3#diff-626187d5d8d29a6936cc1b5876f7edd9R217 it appears, that the default port value if not passed is -1(I understand the reason for choice is AFAIK port numbers are positive and hence to be able to differentiate a valid port number from a non-passed port number) which is not checked for to assume a default port(can be the only obvious port in case of single port or a random selection of port in case of multiple ports) as in, if you see line 98, it only validates that the port provided by user as a cli flag, is a valid port from list of component ports but I don't see a place where port not being passed by user is handled

mik-dass: The current PR does not implement the default single port selection as of now. Currently the flag is marked as required. @kadel Should I implement the single port automatic selection or leave the flag required as marked?

kadel: > . @kadel Should I make the port flag completely mandatory?

no flag shouldn't be mandatory if a component is exposing just one port. In that case odo should detect correct port automatically.

mik-dass:
@anmolbabu Done

mik-dass:
@ashetty1 Passing port no is mandatory in case of components which expose two ports while creation. In case of components which expose a single port while creation, the port number is detected and used automatically.

ashetty1:
@mlk-dass then we definitely need unit and e2e tests for cases where port no is not provided.

mik-dass:
@ashetty1 I have already added the unit tests for the new function and also modified a few others, I will add the e2e tests for the no port no provided condition

cmd/url.go
```diff
@@ -87,8 +89,33 @@ The created URL can be used to access the specified component from outside the O
 			os.Exit(1)
 		}
 
+		componentPorts, err := url.GetComponentServicePortNumbers(client, componentName, applicationName)
+		checkError(err, "unable to get component exposed ports")
+
+		var portFound bool
+
+		if urlPort == -1 {
+			if len(componentPorts) > 1 {
```
anmolbabu: Can there be a case where a component exposes no ports?
Ex: An internal heartbeat application that checks states of various other components and only updates state in a db and is not exposed to the external?
In which case, throw an error as there's nothing to expose ?
Just thinking...

anmolbabu: Can there be a case where a component exposes no ports?
Ex: An internal heartbeat application that checks states of various other components and only updates state in a db and is not exposed to the external?
In which case, throw an error as there's nothing to expose ?
Just a thought

tests/e2e/e2e_test.go
```diff
@@ -261,14 +261,18 @@ var _ = Describe("odoe2e", func() {
 
 	Describe("Creating odo url", func() {
 		Context("using odo url", func() {
-			It("should create route", func() {
+			It("should create route with required port", func() {
```
anmolbabu: An e2e test for : Single port component with no port args passed

cmd/url.go
kadel: It makes sense. It should be as little logic in this file as possible. Especially because we can't easily write unit tests for this.

pkg/url/url.go
```diff
@@ -33,7 +35,7 @@ func Delete(client *occlient.Client, urlName string, applicationName string) err
 }
 
 // Create creates a URL
-func Create(client *occlient.Client, urlName string, componentName, applicationName string) (*URL, error) {
+func Create(client *occlient.Client, urlName string, portNumber int, componentName, applicationName string) (*URL, error) {
```
kadel: it would be good to describe what port number is in comment for this function

cmd/url.go
anmolbabu: May be ValidateURLCreate if you are planning to move it to a new function which is ideally good
Otherwise you can also move it to the create function itself as in https://github.com/redhat-developer/odo/pull/615/files/7c24d37c46e48352a24fd11db6653506d739a5a1#r210214808

mik-dass:
@anmolbabu @kadel Done

pkg/url/url.go
```diff
@@ -33,33 +36,57 @@ func Delete(client *occlient.Client, urlName string, applicationName string) err
...
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable to create namespaced name")
 	}
 
-	if urlName == "" {
-		// Namespace the component
-		urlName = serviceName
+	componentPorts, err := GetComponentServicePortNumbers(client, componentName, applicationName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get component exposed ports for component %s", componentName)
```
anmolbabu: unable to get ports for component %s ?

mik-dass: @anmolbabu Done

cmd/url.go
```diff
@@ -36,14 +38,14 @@ var urlCreateCmd = &cobra.Command{
 
 The created URL can be used to access the specified component from outside the OpenShift cluster.
 `,
-	Example: `  # Create a URL for the current component.
-  odo url create
+	Example: `  # Create a URL for the current component with a specific port
+  odo url create --port 8080
```
kadel: we shouldn't have `--port` flag in every example, it looks like it is mandatory argument.

cmd/url.go
```diff
@@ -179,16 +181,24 @@ var urlListCmd = &cobra.Command{
...
 			}
+			tabWriterURL.Flush()
 		}
 	},
 }
 
 func init() {
 	urlCreateCmd.Flags().StringVarP(&urlApplication, "application", "a", "", "create url for application")
 	urlCreateCmd.Flags().StringVarP(&urlComponent, "component", "c", "", "create url for component")
+	urlCreateCmd.Flags().IntVarP(&urlPort, "port", "", -1, "port number for the url of the component")
```
kadel: Help message should explain that this flag is required only if component exposes more than one port. Otherwise, a component port will be used.

pkg/occlient/occlient.go
```diff
@@ -1246,7 +1246,7 @@ func (c *Client) clusterServiceClassExists(name string) bool {
 
 // CreateRoute creates a route object for the given service and with the given
 // labels
-func (c *Client) CreateRoute(name string, serviceName string, labels map[string]string) (*routev1.Route, error) {
+func (c *Client) CreateRoute(name string, serviceName string, portNumber intstr.IntOrString, labels map[string]string) (*routev1.Route, error) {
```
kadel: the comment should explain what `portNumber` argument is.
It would be better to have portNumber as int  and convert it to `IntOrString` inside this function.

pkg/url/url.go
```diff
@@ -33,33 +36,57 @@ func Delete(client *occlient.Client, urlName string, applicationName string) err
 }
 
 // Create creates a URL
-func Create(client *occlient.Client, urlName string, componentName, applicationName string) (*URL, error) {
+// portNumber is the target port number for the route and is -1 in case no port number is specified
```
kadel: it would be better if this mentions that `-1` means that the port will be automatically detected from the component

pkg/url/url.go
```diff
@@ -12,12 +12,15 @@ import (
...
 	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"strings"
 )
 
 type URL struct {
 	Name     string
 	URL      string
 	Protocol string
+	Port     intstr.IntOrString
```
kadel: Are we ever going to store anything else than a port number here? Wouldn't it be better to have this as `int`?

mik-dass:
@kadel Done

PR #616, opened 2018-08-03
FIX - sudo permission issue on mac install

springdo:
Hello team! 
This project looks amazing and I cannot wait to see where it goes! 👍 
Just getting started using the binary install and I ran into this permissions error so I am supplying this PR. 

screen shot 2018-08-03 at 11 31 55

Simple fix for the  issue is `sudo chmod` but it could also be a `sudo sh -c ' curl ...'` if you guys want me to change.


kadel:
Hi Donald, thank you for trying odo, and even more thank you for sending PR   👍

PR #618, opened 2018-08-04
Include imagestreams in current namespace for catalog listing

### LINKED ISSUE #566: Use image builders from your namespace when listing the catalog
`odo catalog list` should show items in my local namespace as well, so if I define a builder in OpenShift Online I can use it.

anmolbabu:
This PR considers imagestream in current namespace along with
imagestreams in openshift namespace for:
1. catalog listing i.e, `odo catalog list` is now expected to show items in local
namespace along with those in openshift namespace
2. component create i.e `odo create <comp_name> --git <git_url>` is expected
to work even if comp_name is present in current namespace but not in openshift
namespace

fixes #566

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -54,10 +54,22 @@ var catalogListComponentCmd = &cobra.Command{
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
 			fmt.Fprintln(w, "NAME", "\t", "TAGS")
 			for _, component := range catalogList {
-				fmt.Fprintln(w, fmt.Sprintf("%s/%s", component.Namespace, component.Name), "\t", strings.Join(component.Tags, ","))
+				component_name := fmt.Sprintf("%s/%s", component.Namespace, component.Name)
+				if component.Namespace == client.GetCurrentProjectName() {
+					for _, comp1 := range catalogList {
+						if comp1.Name == component.Name && component.Namespace != comp1.Namespace {
+							component_name = fmt.Sprintf("%s (*)", component_name)
+						}
+					}
+				}
+				fmt.Fprintln(w, component_name, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
-
+			fmt.Println(
+				`NOTE: There might be catalog items in multiple catalogs.
+Those marked with (*) will take precedence.
+In case you want to use an item from a different catalog, please fully qualify it as $namespace/$catalog_name`,
+			)
 		}
 	},
 }
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -71,11 +71,6 @@ var catalogListComponentCmd = &cobra.Command{
 				fmt.Fprintln(w, componentName, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
-			fmt.Println(
-				`NOTE: There might be items in multiple projects.
-Items from the current project (marked with *) will take precedence.
-If you want to use an item from a different namespace, use the full name as <namespace>/<name>`,
-			)
 		}
 	},
 }
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -52,9 +52,9 @@ var catalogListComponentCmd = &cobra.Command{
 		default:
 			currentProject := client.GetCurrentProjectName()
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(w, "NAME", "\t", "NAMESPACE", "\t", "TAGS")
+			fmt.Fprintln(w, "NAME", "\t", "PROJECT", "\t", "TAGS")
 			for _, component := range catalogList {
-				component_name := component.Name
+				componentName := component.Name
 				if component.Namespace == currentProject {
 					/*
 						If current namespace is same as the current component namespace,
@@ -62,17 +62,17 @@ var catalogListComponentCmd = &cobra.Command{
 						If there exists a component with same name but in different namespaces,
 						mark the one from current namespace with (*)
 					*/
-					for _, comp1 := range catalogList {
-						if comp1.Name == component.Name && component.Namespace != comp1.Namespace {
-							component_name = fmt.Sprintf("%s (*)", component.Name)
+					for _, comp := range catalogList {
+						if comp.Name == component.Name && component.Namespace != comp.Namespace {
+							componentName = fmt.Sprintf("%s (*)", component.Name)
 						}
 					}
 				}
-				fmt.Fprintln(w, component_name, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
+				fmt.Fprintln(w, componentName, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
 			fmt.Println(
-				`NOTE: There might be items in multiple namespaces.
+				`NOTE: There might be items in multiple project.
 Items from the current project (marked with *) will take precedence.
 In case you want to use an item from a different namespace, please fully qualify it as $namespace/$name`,
 			)
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -54,7 +54,7 @@ var catalogListComponentCmd = &cobra.Command{
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
 			fmt.Fprintln(w, "NAME", "\t", "TAGS")
 			for _, component := range catalogList {
-				fmt.Fprintln(w, component.Name, "\t", strings.Join(component.Tags, ","))
+				fmt.Fprintln(w, fmt.Sprintf("%s/%s", component.Namespace, component.Name), "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
 
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -50,25 +50,31 @@ var catalogListComponentCmd = &cobra.Command{
 		case 0:
 			fmt.Printf("No deployable components found\n")
 		default:
-
+			currentProject := client.GetCurrentProjectName()
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(w, "NAME", "\t", "TAGS")
+			fmt.Fprintln(w, "NAME", "\t", "NAMESPACE", "\t", "TAGS")
 			for _, component := range catalogList {
-				component_name := fmt.Sprintf("%s/%s", component.Namespace, component.Name)
-				if component.Namespace == client.GetCurrentProjectName() {
+				component_name := component.Name
+				if component.Namespace == currentProject {
+					/*
+						 If current namespace is same as the current component namespace,
+						 	Loop through every other component,
+								If there exists a component with same name but in different namespaces,
+									 mark the one from current namespace with (*)
+					*/
 					for _, comp1 := range catalogList {
 						if comp1.Name == component.Name && component.Namespace != comp1.Namespace {
-							component_name = fmt.Sprintf("%s (*)", component_name)
+							component_name = fmt.Sprintf("%s (*)", component.Name)
 						}
 					}
 				}
-				fmt.Fprintln(w, component_name, "\t", strings.Join(component.Tags, ","))
+				fmt.Fprintln(w, component_name, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
 			fmt.Println(
-				`NOTE: There might be catalog items in multiple catalogs.
-Those marked with (*) will take precedence.
-In case you want to use an item from a different catalog, please fully qualify it as $namespace/$catalog_name`,
+				`NOTE: There might be items in multiple namespaces.
+Items from current project (marked with *) will take precedence.
+In case you want to use an item from a different namespace, please fully qualify it as $namespace/$name`,
 			)
 		}
 	},
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -72,9 +72,9 @@ var catalogListComponentCmd = &cobra.Command{
 			}
 			w.Flush()
 			fmt.Println(
-				`NOTE: There might be items in multiple project.
+				`NOTE: There might be items in multiple projects.
 Items from the current project (marked with *) will take precedence.
-In case you want to use an item from a different namespace, please fully qualify it as $namespace/$name`,
+If you want to use an item from a different namespace, use the full name as <namespace>/<name>`,
 			)
 		}
 	},
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -57,10 +57,10 @@ var catalogListComponentCmd = &cobra.Command{
 				component_name := component.Name
 				if component.Namespace == currentProject {
 					/*
-						 If current namespace is same as the current component namespace,
-						 	Loop through every other component,
-								If there exists a component with same name but in different namespaces,
-									 mark the one from current namespace with (*)
+						If current namespace is same as the current component namespace,
+						Loop through every other component,
+						If there exists a component with same name but in different namespaces,
+						mark the one from current namespace with (*)
 					*/
 					for _, comp1 := range catalogList {
 						if comp1.Name == component.Name && component.Namespace != comp1.Namespace {
@@ -73,7 +73,7 @@ var catalogListComponentCmd = &cobra.Command{
 			w.Flush()
 			fmt.Println(
 				`NOTE: There might be items in multiple namespaces.
-Items from current project (marked with *) will take precedence.
+Items from the current project (marked with *) will take precedence.
 In case you want to use an item from a different namespace, please fully qualify it as $namespace/$name`,
 			)
 		}
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -85,7 +85,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 		// "Default" values
 		componentImageName := args[0]
 		componentType := args[0]
-		componentName := args[0]
+		componentName := strings.Replace(args[0], "/", "-", -1)
 		componentVersion := "latest"
 
 		// Check if componentType includes ":", if so, then we need to spit it into using versions
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -30,15 +30,8 @@ var componentCreateCmd = &cobra.Command{
 
 If component name is not provided, component type value will be used for the name.
 
-The component type will be chosen by default from current project,
-but if specific version is requested for, it'll be used from either
-* current
-   or
-* openshift
-
-project(with search precendence in same order) where available.
-So, if you would like it to be specifically from openshift project, please fully qualify the
-component type as $project/$component_type:$version.
+By default, builder images will be used from the current namespace.
+You can explicitly supply a namespace by using: odo create namespace/name:version
 If version is not specified by default, latest wil be chosen as the version.
 
 A full list of component types that can be deployed is available using: 'odo catalog list'`,
@@ -54,7 +47,7 @@ A full list of component types that can be deployed is available using: 'odo cat
   # Create new Node.js component with source from remote git repository.
   odo create nodejs --git https://github.com/openshift/nodejs-ex.git
 
-  # Create new Node.js component of version 6 using the builder image specifically from 'openshift' namespace
+  # Create a new Node.js component of version 6 from the 'openshift' namespace
   odo create openshift/nodejs:6 --local /nodejs-ex
 
   # Create new Wildfly component with binary named sample.war in './downloads' directory
@@ -189,13 +182,11 @@ A full list of component types that can be deployed is available using: 'odo cat
 	},
 }
 
-/*
- parseCreateCmdArgs returns
- 1. image name
- 2. component type i.e, builder image name
- 3. component name default value is component type else the user requested component name
- 4. component version which is by default latest else version passed with builder image name
-*/
+// parseCreateCmdArgs returns
+// 1. image name
+// 2. component type i.e, builder image name
+// 3. component name default value is component type else the user requested component name
+// 4. component version which is by default latest else version passed with builder image name
 func parseCreateCmdArgs(args []string) (string, string, string, string) {
 	// We don't have to check it anymore, Args check made sure that args has at least one item
 	// and no more than two
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -30,8 +30,19 @@ var componentCreateCmd = &cobra.Command{
 
 If component name is not provided, component type value will be used for the name.
 
+The component type will be chosen by default from current project,
+but if specific version is requested for, it'll be used from either
+* current
+   or
+* openshift
+
+project(with search precendence in same order) where available.
+So, if you would like it to be specifically from openshift project, please fully qualify the
+component type as $project/$component_type:$version.
+If version is not specified by default, latest wil be chosen as the version.
+
 A full list of component types that can be deployed is available using: 'odo catalog list'`,
-	Example: `  # Create new Node.js component with the source in current directory. 
+	Example: `  # Create new Node.js component with the source in current directory.
   odo create nodejs
 
   # A specific image version may also be specified
@@ -42,6 +53,9 @@ A full list of component types that can be deployed is available using: 'odo cat
 
   # Create new Node.js component with source from remote git repository.
   odo create nodejs --git https://github.com/openshift/nodejs-ex.git
+
+  # Create new Node.js component of version 6 using the builder image specifically from 'openshift' namespace
+  odo create openshift/nodejs:6 --local /nodejs-ex
 
   # Create new Wildfly component with binary named sample.war in './downloads' directory
   odo create wildfly wildly --binary ./downloads/sample.war
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -30,8 +30,7 @@ var componentCreateCmd = &cobra.Command{
 
 If component name is not provided, component type value will be used for the name.
 
-By default, builder images will be used from the current namespace.
-You can explicitly supply a namespace by using: odo create namespace/name:version
+By default, builder images will be used from the current namespace. You can explicitly supply a namespace by using: odo create namespace/name:version
 If version is not specified by default, latest wil be chosen as the version.
 
 A full list of component types that can be deployed is available using: 'odo catalog list'`,
@@ -87,7 +86,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
-		componentImageName, componentType, componentName, componentVersion := parseCreateCmdArgs(args)
+		componentImageName, componentType, componentName, componentVersion := util.ParseCreateCmdArgs(args)
 
 		// Check to see if the catalog type actually exists
 		exists, err := catalog.Exists(client, componentType)
@@ -180,31 +179,6 @@ A full list of component types that can be deployed is available using: 'odo cat
 		checkError(err, "")
 		fmt.Printf("\nComponent '%s' is now set as active component.\n", componentName)
 	},
-}
-
-// parseCreateCmdArgs returns
-// 1. image name
-// 2. component type i.e, builder image name
-// 3. component name default value is component type else the user requested component name
-// 4. component version which is by default latest else version passed with builder image name
-func parseCreateCmdArgs(args []string) (string, string, string, string) {
-	// We don't have to check it anymore, Args check made sure that args has at least one item
-	// and no more than two
-
-	// "Default" values
-	componentImageName := args[0]
-	componentType := args[0]
-	componentName := util.ExtractComponentType(componentType)
-	componentVersion := "latest"
-
-	// Check if componentType includes ":", if so, then we need to spit it into using versions
-	if strings.ContainsAny(componentImageName, ":") {
-		versionSplit := strings.Split(args[0], ":")
-		componentType = versionSplit[0]
-		componentName = util.ExtractComponentType(componentType)
-		componentVersion = versionSplit[1]
-	}
-	return componentImageName, componentType, componentName, componentVersion
 }
 
 func init() {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -80,22 +80,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
-		// We don't have to check it anymore, Args check made sure that args has at least one item
-		// and no more than two
-
-		// "Default" values
-		componentImageName := args[0]
-		componentType := args[0]
-		componentName := util.ExtractComponentType(componentType)
-		componentVersion := "latest"
-
-		// Check if componentType includes ":", if so, then we need to spit it into using versions
-		if strings.ContainsAny(componentImageName, ":") {
-			versionSplit := strings.Split(args[0], ":")
-			componentType = versionSplit[0]
-			componentName = util.ExtractComponentType(componentType)
-			componentVersion = versionSplit[1]
-		}
+		componentImageName, componentType, componentName, componentVersion := parseCreateCmdArgs(args)
 
 		// Check to see if the catalog type actually exists
 		exists, err := catalog.Exists(client, componentType)
@@ -188,6 +173,33 @@ A full list of component types that can be deployed is available using: 'odo cat
 		checkError(err, "")
 		fmt.Printf("\nComponent '%s' is now set as active component.\n", componentName)
 	},
+}
+
+/*
+ parseCreateCmdArgs returns
+ 1. image name
+ 2. component type i.e, builder image name
+ 3. component name default value is component type else the user requested component name
+ 4. component version which is by default latest else version passed with builder image name
+*/
+func parseCreateCmdArgs(args []string) (string, string, string, string) {
+	// We don't have to check it anymore, Args check made sure that args has at least one item
+	// and no more than two
+
+	// "Default" values
+	componentImageName := args[0]
+	componentType := args[0]
+	componentName := util.ExtractComponentType(componentType)
+	componentVersion := "latest"
+
+	// Check if componentType includes ":", if so, then we need to spit it into using versions
+	if strings.ContainsAny(componentImageName, ":") {
+		versionSplit := strings.Split(args[0], ":")
+		componentType = versionSplit[0]
+		componentName = util.ExtractComponentType(componentType)
+		componentVersion = versionSplit[1]
+	}
+	return componentImageName, componentType, componentName, componentVersion
 }
 
 func init() {
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -12,6 +12,7 @@ import (
 	"github.com/redhat-developer/odo/pkg/catalog"
 	"github.com/redhat-developer/odo/pkg/component"
 	"github.com/redhat-developer/odo/pkg/project"
+	"github.com/redhat-developer/odo/pkg/util"
 	"github.com/spf13/cobra"
 )
 
@@ -85,14 +86,14 @@ A full list of component types that can be deployed is available using: 'odo cat
 		// "Default" values
 		componentImageName := args[0]
 		componentType := args[0]
-		componentName := strings.Replace(args[0], "/", "-", -1)
+		componentName := util.ExtractComponentType(componentType)
 		componentVersion := "latest"
 
 		// Check if componentType includes ":", if so, then we need to spit it into using versions
 		if strings.ContainsAny(componentImageName, ":") {
 			versionSplit := strings.Split(args[0], ":")
 			componentType = versionSplit[0]
-			componentName = versionSplit[0]
+			componentName = util.ExtractComponentType(componentType)
 			componentVersion = versionSplit[1]
 		}
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -125,10 +125,10 @@ Note:
 By default builder imagestreams from current project are used for building your application.
 But, if the builder imagestream corresponding to the type of requested component does not exist in the current project,
 builder imagestreams from openshift namespace are also considered. But if you would like to specifically use the default
-builder imagestream, you can fully qualify the component type using $project/$component_type:$version.
+builder imagestream, you can fully qualify the component type using `$project/$component_type:$version`.
 Accordingly following commands can also be used:
-* odo create openshift/nodejs:6 --local=.
-* odo create nodejs:6 --local=.
+* `odo create openshift/nodejs:6 --local=.`
+* `odo create nodejs:6 --local=.`
 
 For more details please refer https://github.com/redhat-developer/odo/blob/master/docs/cli-reference.md#create
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -122,14 +122,8 @@ Component 'nodejs' is now set as active component.
 
 Note:
 
-By default builder imagestreams from current project are used for building your application.
-But, if the builder imagestream corresponding to the type of requested component does not exist in the current project,
-builder imagestreams from openshift namespace are also considered. But if you would like to specifically use the default
-builder imagestream, you can fully qualify the component type using `$project/$component_type:$version`.
-Accordingly following commands can also be used:
-* `odo create openshift/nodejs:6 --local=.`
-* `odo create nodejs:6 --local=.`
-
+By default, builder images will be used from the default OpenShift namespace.
+You can explicitly supply a namespace by using: `odo create opesnshift/nodejs:8`
 For more details please refer https://github.com/redhat-developer/odo/blob/master/docs/cli-reference.md#create
 
 Now that a component is running we'll go ahead and push our initial source code!
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -120,6 +120,18 @@ To push source code to the component run 'odo push'
 Component 'nodejs' is now set as active component.
 ```
 
+Note:
+
+By default builder imagestreams from current project are used for building your application.
+But, if the builder imagestream corresponding to the type of requested component does not exist in the current project,
+builder imagestreams from openshift namespace are also considered. But if you would like to specifically use the default
+builder imagestream, you can fully qualify the component type using $project/$component_type:$version.
+Accordingly following commands can also be used:
+* odo create openshift/nodejs:6 --local=.
+* odo create nodejs:6 --local=.
+
+For more details please refer https://github.com/redhat-developer/odo/blob/master/docs/cli-reference.md#create
+
 Now that a component is running we'll go ahead and push our initial source code!
 
 ```sh
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -1,6 +1,7 @@
 package catalog
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/golang/glog"
@@ -54,7 +55,7 @@ func Exists(client *occlient.Client, componentType string) (bool, error) {
 	}
 
 	for _, supported := range catalogList {
-		if componentType == supported.Name {
+		if componentType == supported.Name || componentType == fmt.Sprintf("%s/%s", supported.Namespace, supported.Name) {
 			return true, nil
 		}
 	}
@@ -72,7 +73,7 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
 
 	// Find the component and then return true if the version has been found
 	for _, supported := range catalogList {
-		if componentType == supported.Name {
+		if componentType == supported.Name || componentType == fmt.Sprintf("%s/%s", supported.Namespace, supported.Name) {
 			// Now check to see if that version matches that components tag
 			for _, tag := range supported.Tags {
 				if componentVersion == tag {
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -103,7 +103,7 @@ func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
 		// We may get the imagestreams from other Namespaces
 		err = errors.Wrapf(e, "unable to get Image Streams from namespace %s", occlient.OpenShiftNameSpace)
 		// log it for debugging purposes
-		log.Debugf("Unable to get Image Streams from namespace %s. Error %s", occlient.OpenShiftNameSpace, e.Error())
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", occlient.OpenShiftNameSpace, e.Error())
 	}
 
 	// Fetch imagestreams from current namespace
@@ -117,7 +117,7 @@ func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
 		// But may be required for debugging purposes
 		err = errors.Wrapf(e, "unable to get Image Streams from namespace %s", currentNamespace)
 		// log it for debugging purposes
-		log.Debugf("Unable to get Image Streams from namespace %s. Error %s", currentNamespace, e.Error())
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", currentNamespace, e.Error())
 	}
 
 	// Resultant imagestreams is list of imagestreams from current and openshift namespaces
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -94,30 +94,34 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
 func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
 	var imageStreams []imagev1.ImageStream
 	currentNamespace := client.GetCurrentProjectName()
-	var err error
 
 	// Fetch imagestreams from default openshift namespace
-	openshiftNSImageStreams, possibleError := client.GetImageStreams(occlient.OpenShiftNameSpace)
-	if possibleError != nil {
+	openshiftNSImageStreams, openshiftNSISFetchError := client.GetImageStreams(occlient.OpenShiftNameSpace)
+	if openshiftNSISFetchError != nil {
 		// Tolerate the error as it might only be a partial failure
 		// We may get the imagestreams from other Namespaces
-		err = errors.Wrapf(possibleError, "unable to get Image Streams from namespace %s", occlient.OpenShiftNameSpace)
+		//err = errors.Wrapf(openshiftNSISFetchError, "unable to get Image Streams from namespace %s", occlient.OpenShiftNameSpace)
 		// log it for debugging purposes
-		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", occlient.OpenShiftNameSpace, possibleError.Error())
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", occlient.OpenShiftNameSpace, openshiftNSISFetchError.Error())
 	}
 
 	// Fetch imagestreams from current namespace
-	currentNSImageStreams, possibleError := client.GetImageStreams(currentNamespace)
-	if possibleError != nil {
-		if err != nil {
-			// Failure to fetch images from any namespace, error out
-			return nil, errors.Wrapf(possibleError, "%s. unable to get Image Streams from namespace %s", err.Error(), currentNamespace)
-		}
-		// It is possible that the current namespace has no imagestreams -- a valid scenario
-		// But may be required for debugging purposes
-		err = errors.Wrapf(possibleError, "unable to get Image Streams from namespace %s", currentNamespace)
+	currentNSImageStreams, currentNSISFetchError := client.GetImageStreams(currentNamespace)
+	// If failure to fetch imagestreams from current namespace, log the failure for debugging purposes
+	if currentNSISFetchError != nil {
+		// Tolerate the error as it is totally a valid scenario to not have any imagestreams in current namespace
 		// log it for debugging purposes
-		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", currentNamespace, possibleError.Error())
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", currentNamespace, currentNSISFetchError.Error())
+	}
+
+	// If failure fetching imagestreams from both namespaces, error out
+	if openshiftNSISFetchError != nil && currentNSISFetchError != nil {
+		return nil, errors.Wrapf(
+			fmt.Errorf("%s.\n%s", openshiftNSISFetchError, currentNSISFetchError),
+			"Failed to fetch imagestreams from both openshift and %s namespaces.\nPlease ensure that a builder imagestream of required version for the component exists in either openshift or %s namespaces",
+			currentNamespace,
+			currentNamespace,
+		)
 	}
 
 	// Resultant imagestreams is list of imagestreams from current and openshift namespaces
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -5,13 +5,15 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
+	imagev1 "github.com/openshift/api/image/v1"
 	"github.com/pkg/errors"
 	"github.com/redhat-developer/odo/pkg/occlient"
 )
 
 type CatalogImage struct {
-	Name string
-	Tags []string
+	Name      string
+	Namespace string
+	Tags      []string
 }
 
 // List lists all the available component types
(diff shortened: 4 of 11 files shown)


anmolbabu:
@surajnarwade I tried grepping for any other usages of `occlient.OpenShiftNameSpace` and also its equivalent "openshift" but didn't find any other the one in this PR, can you please help me find which other areas of code need to undergo changes as per last paragraph of comment in https://github.com/redhat-developer/odo/issues/566#issuecomment-409523030

surajnarwade:
**How to test ?**

* Use this link to create imagestream in current namespace, https://paste.fedoraproject.org/paste/JBld1qYlLozPWqYcFyC32g

* oc create -f filename

* oc get imagestream

```
NAME                      DOCKER REPO                                         TAGS                   UPDATED
nodejs-currentnamespace   172.30.1.1:5000/myproject/nodejs-currentnamespace   0.10,4,6 + 2 more...   
```

* odo catalog list

```
$ odo catalog list
NAME                        TAGS
dotnet                      2.0,latest
httpd                       2.4,latest
nginx                       1.10,1.12,1.8,latest
nodejs                      0.10,4,6,8,latest
perl                        5.16,5.20,5.24,latest
php                         5.5,5.6,7.0,7.1,latest
python                      2.7,3.3,3.4,3.5,3.6,latest
ruby                        2.0,2.2,2.3,2.4,latest
wildfly                     10.0,10.1,8.1,9.0,latest
nodejs-currentnamespace     0.10,4,6,8,latest
```

anmolbabu:
Verified as in :

[redhat@localhost redhat-developer]$ oc project
Using project "myproject" on server "https://127.0.0.1:8443".

[redhat@localhost redhat-developer]$ oc get imagestreams -n myproject
NAME                      DOCKER REPO                                         TAGS                        UPDATED
nodejs-currentnamespace   172.30.1.1:5000/myproject/nodejs-currentnamespace   8,latest,0.10 + 2 more...   

[redhat@localhost redhat-developer]$ odo catalog list
NAME                        TAGS
dotnet                      2.0,latest
httpd                       2.4,latest
nginx                       1.10,1.12,1.8,latest
nodejs                      0.10,4,6,8,latest
perl                        5.16,5.20,5.24,latest
php                         5.5,5.6,7.0,7.1,latest
python                      2.7,3.3,3.4,3.5,3.6,latest
ruby                        2.0,2.2,2.3,2.4,latest
wildfly                     10.0,10.1,8.1,9.0,latest
nodejs-currentnamespace     0.10,4,6,8,latest

anmolbabu:
@surajnarwade can you please revisit the PR I have incorporated your comments..
Please also have a look at https://github.com/redhat-developer/odo/pull/618#issuecomment-410448511

pkg/catalog/catalog_test.go
```diff
@@ -92,45 +92,91 @@ func TestVersionExist(t *testing.T) {
...
 	}
 	tests := []struct {
-		name     string
-		args     args
-		wantErr  bool
-		wantTags []string
+		name    string
+		args    args
+		wantErr bool
+		wantOp  []CatalogImage // Expect 1 duplicate because of 2 calls(1 for current namespace and another for openshift namespace) to client.GetImageStreams per call to list
```
anmolbabu: @kadel Is there a better way of doing this test
i.e, https://github.com/redhat-developer/odo/pull/618/files#diff-1dfbb1796e8d3ca31e26fe9d003bb488R188 for particular parameters instead of whole list?

pkg/catalog/catalog.go
```diff
@@ -87,13 +87,19 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
 func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
 	imageStreams, err := client.GetImageStreams(occlient.OpenShiftNameSpace)
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to get Image Streams")
+		return nil, errors.Wrapf(err, "unable to get Image Streams from %s namespace", occlient.OpenShiftNameSpace)
 	}
 
+	imageStreams2, err := client.GetImageStreams(client.GetCurrentProjectName())
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get Image Streams from %s namespace", imageStreams2)
```
syamgk: this should be `client.GetCurrentProjectName()` instead of `imageStreams2`

pkg/catalog/catalog_test.go
kadel: I think that this is ok, but the duplicate stuff is a bit weird

pkg/catalog/catalog_test.go
```diff
@@ -92,45 +92,91 @@ func TestVersionExist(t *testing.T) {
...
 	type args struct {
-		name      string
-		namespace string
-		tags      []string
+		name               string
+		namespace          string
+		tags               []string
+		currentNSImageName string
+		currentNamespace   string
+		currentNSImageTags []string
```
kadel: what is this variable for?

pkg/catalog/catalog_test.go
```diff
@@ -140,7 +186,7 @@ func TestList(t *testing.T) {
 			// Fake the client with the appropriate arguments
 			client, fakeClientSet := occlient.FakeNew()
 			fakeClientSet.ImageClientset.PrependReactor("list", "imagestreams", func(action ktesting.Action) (bool, runtime.Object, error) {
-				return true, testingutil.FakeImageStreams(tt.args.name, tt.args.namespace, tt.args.tags), nil
+				return true, testingutil.FakeImageStreams(tt.args.name, tt.args.namespace, tt.args.tags, tt.args.currentNSImageName, tt.args.currentNamespace, tt.args.currentNSImageTags), nil
```
kadel: why not just call ` FakeImageStreams` twice? Once to add images to "openshift" namespace, second time to add  them to the other namespace. You don't have to modify ` FakeImageStreams` function at all.

pkg/catalog/catalog.go
```diff
@@ -87,13 +87,19 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
 }
 
 // getDefaultBuilderImages returns the default builder images available in the
-// openshift namespace
+// openshift namespace and current namespace
 func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
```
kadel: This image should also return in what namespace the image is. as there might be two images with the same name in both openshift and current namespace. We also have to figure out what we will do if there is a collision.

surajnarwade:
```
$ odo catalog list
NAME        TAGS
dotnet      2.0,latest
httpd       2.4,latest
nodejs      0.10,4,6,8,latest
ruby        2.0,2.2,2.3,2.4,2.5,latest
wildfly     10.0,10.1,11.0,12.0,8.1,9.0,latest
mynodjs      0.10,4,6,8,latest
$ odo create mynodejs
Invalid component type: mynodejs
Run 'odo catalog list' to see a list of supported components
```

It can list the new imagestream, but can not use it

kadel:
How are we going to solve the duplicity issue? What if the same image/ImageStream is in openshift project and in the current project?

pkg/catalog/catalog.go
anmolbabu: Ack

pkg/catalog/catalog.go
anmolbabu: Ack

The o/p now looks like this: https://gist.github.com/anmolbabu/028a4caf25c009e3c63c82721f8146f3

anmolbabu:
Still facing the issue:

Error resolving ImageStreamTag python:latest in namespace openshift: imagestreams.image.openshift.io "python" not found

Probably an additional param/config needs to be passed to build intialise request

pkg/occlient/occlient.go
```diff
@@ -380,40 +380,122 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
...
+
+	// Try finding imagestream from current namespace
+	// Even if imagestream is present in openshift namespace, it is possible that the required tag is found in current namespace and not in openshift namespace
+	currentNSImageStream, e := c.imageClient.ImageStreams(currentProjectName).Get(imageName, metav1.GetOptions{})
+	if e != nil {
+		if err != nil {
+			return nil, errors.Wrapf(err, "no match found for : %s in namespace %s. error :%v", imageName, currentProjectName, e)
+		}
+		// Ignore the error if there was no error previously as the desired imagestream might be only in openshift namespace
+		err = nil
```
surajnarwade: you dont need this

pkg/catalog/catalog.go
```diff
@@ -65,66 +61,69 @@ func Exists(client *occlient.Client, componentType string) (bool, error) {
...
 	return false, nil
 }
 
 // getDefaultBuilderImages returns the default builder images available in the
-// openshift namespace
-func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
-	imageStreams, err := client.GetImageStreams(occlient.OpenShiftNameSpace)
+// openshift namespace and current namespace
+// Return  map[imageName: string]map[namespace:string][]tags:string
+func getDefaultBuilderImages(client *occlient.Client) (map[string]map[string][]string, error) {
```
kadel: uff this is pretty complicated return type. Wouldn't it be easier to create some nice struct for this?

surajnarwade:
**What we need to do for this PR ?**

* `GetExposedPorts` function should return namespace along with ports as well. This function should also decide which namespace to chose in case of collision (current namespace and openshift namespace consist of same image). Ideally, it should take current namespace, because `Dev` might have some intention behind this to create one.

* In `BootstrapSupervisoredS2I`, use the _namespace_ given by `GetExposedPorts` function

kadel:
> In BootstrapSupervisoredS2I, use the namespace given by GetExposedPorts function here

the same for [NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string, inputPorts []string)](https://sourcegraph.com/github.com/redhat-developer/odo/-/blob/pkg/occlient/occlient.go?utm_source=inline-symbol#L426:18-426:27)

surajnarwade:
@anmolbabu , if there's an issue with creating imagestream, follow this: http://suraj.pro/post/access-minishift-registry/

jorgemoralespou:
@kadel  As how openshift works currently, and I think it might make sense, when there's imagestreams that are duplicated, in openshift namespace and current namespace, we should qualify the image or else, the local namespace should take precendence. Probably would also be good to show in the catalog listing as well:

```
$ odo catalog list
NAME        TAGS
openshift/nodejs      0.10,4,6,8,latest
nodejs      0.10,4,6,8,latest
```

Or add a visual indication to the printed list:
```
$ odo catalog list
NAME        TAGS                     CATALOG
nodejs        0.10,4,6,8,latest    openshift
nodejs (*)   0.10,4,6,8,latest    my-project 

NOTE: There might be catalog items in multiple catalogs. Those marked with (*) will take precedence. In case you want to use an item from a different catalog, please fully qualify it:
e.g. 

$ odo create openshift/python ....

or

$ odo create python ..... --catalog=openshift
```

Then, it would be obvious which one is used. I used the term catalog here in my explanation although it might not be the best term.

pkg/catalog/catalog.go
anmolbabu: Iteration of struct array becomes tedious for any operations like search and delete.
But in a map the access is O(1) for most cases although at the cost of ease of code maintenance.
But it might now boil down to likely max size of list of imagestreams in current NS and openshift NS which looking at default size in my minishift deployment seems very small to bother about time complexities..

So,  I am changing https://github.com/redhat-developer/odo/pull/618/files#diff-4373e1ac6258d2d589cb87e96b1fc994R13 to 

```
type CatalogImage struct {
	Name string
	NamespacedTags map[Namespace: string]Tags: []string
}

```

pkg/catalog/catalog_test.go
anmolbabu: There are 2 calls to the same function(although different params) https://github.com/redhat-developer/odo/pull/618/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR401 and https://github.com/redhat-developer/odo/pull/618/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR391 and the function to mock the call isn't per parameter(as in return x when function called with y as parameter else return z) and the mocking is to api function not its wrapper that is immediately used in our func...

Hence, inorder for the mocking function to be able to return values from both namespaces this was the only way possible because the mocking is not per parameter

pkg/catalog/catalog_test.go
anmolbabu: will try this
but how would this work? the prependReactor doesn't take params as arg right to expect it to return 2 different NS values for 2 different calls based on params passed?

anmolbabu:
ToDo(@anmolbabu) Unit Test + publish results of manual testing

anmolbabu:
Verified as in:
1. Python imagestream available in both openshift namespace and current namespace: https://gist.github.com/anmolbabu/92fdf19c5c16d79809e65e258884314d
2. myprojectnodejs imagestream available in myproject namespace and not in openshift namespace: https://gist.github.com/anmolbabu/78d350df6cfaf5a03208e090678bbbf8
3. wildfly imagestream existing only in openshift namespace: https://gist.github.com/anmolbabu/9e7afc910eea1b79a2c35085d0ede38c

pkg/catalog/catalog_test.go
anmolbabu: @kadel Hey but it worked

anmolbabu:
@kadel @surajnarwade @syamgk Done incorporated all comments. Please review

anmolbabu:
> Or add a visual indication to the printed list:
```
$ odo catalog list
NAME        TAGS                     CATALOG
nodejs        0.10,4,6,8,latest    openshift
nodejs (*)   0.10,4,6,8,latest    my-project 

NOTE: There might be catalog items in multiple catalogs. Those marked with (*) will take precedence. In case you want to use an item from a different catalog, please fully qualify it:
```

@jorgemoralespou Consider the below:

`nodejs` exists both in openshift and current namespace- `myproject` but the tags vary as under:
openshift namespace -- 0.10, 4, 6, 8 and latest
myproject namespace -- 8 and latest

Now user triggers `odo create nodejs:6 --git https://github.com/openshift/nodejs-ex -v` I would expect odo to choose myproject see that the tag doesn't exist then see if it exists in openshift namespace and since it is there, use it; but would the '*' then confuse the user into thinking that the imagestream used would probably be from myproject namespace and not openshift namespace(or the note at the bottom that the myproject namespace takes "precedence" in case of redundancy suffice ?)? Or should it error out as there is an image with same name in top priority namespace -- myproject but matching tag is not available?

Note: The current form of PR has capability to fallback to openshift namespace if required tag is not present in myproject namespace

anmolbabu:
Verified as in :

https://gist.github.com/anmolbabu/a01d04ecf1d32aa476027d2519f6126d -- NodeJS tests from git
https://gist.github.com/anmolbabu/5021c2849a99a4307500c5abd78b1e8b -- wildfly tests from binary
https://gist.github.com/anmolbabu/40765d119ec021c114ff6a443ba57e46 -- python tests from local source

Seems like tests are green with Deploy in progress

@kadel @cdrage @surajnarwade @mik-dass @syamgk Please review

pkg/occlient/occlient.go
anmolbabu: Ack

cmd/catalog.go
```diff
@@ -42,10 +42,22 @@ var catalogListCmd = &cobra.Command{
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
 			fmt.Fprintln(w, "NAME", "\t", "TAGS")
 			for _, component := range catalogList {
-				fmt.Fprintln(w, component.Name, "\t", strings.Join(component.Tags, ","))
+				component_name := fmt.Sprintf("%s/%s", component.Namespace, component.Name)
+				if component.Namespace == client.GetCurrentProjectName() {
```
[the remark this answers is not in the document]
anmolbabu: Ack

cmd/catalog.go
```diff
@@ -42,10 +42,22 @@ var catalogListCmd = &cobra.Command{
...
+					for _, comp1 := range catalogList {
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/catalog/catalog.go
```diff
@@ -89,11 +92,34 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
 // getDefaultBuilderImages returns the default builder images available in the
 // openshift namespace
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/occlient/occlient.go
```diff
@@ -377,43 +391,145 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
...
+		if !isTagInImageStream(*imageStream, imageTag) {
+			return nil, errors.Errorf("image stream %s with tag %s not found in %s namespaces", imageName, imageTag, currentProjectName)
+		}
+	}
+
+	return imageStream, nil
+}
+
+// GetExposedPorts retruns image namespace and list of ContainerPorts that are exposed by given image
+func (c *Client) GetExposedPorts(imageNS string, imageName string, imageTag string) (string, []corev1.ContainerPort, error) {
```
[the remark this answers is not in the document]
anmolbabu: The imagestream is not expected to always be passed by the user.
This Pr has the capability to resolve imagestream based on:
0. User chosen namespace(if explicitly passed)
1. Priority for the imagestream in current namespace.
2. Availability of required tag in the currently considered namespace
For cases 1 and 2, when user has not passed imagestream namespace, the resolved namespace will need to be set into build config for imagestream mapping although, it can be also resolved by exclusive separate call to  GetImageStream, in order to avoid multiple calls to this(GetImageStream) function, I made GetExposedPorts also return namespace

pkg/occlient/occlient.go
```diff
@@ -425,18 +541,23 @@ func getAppRootVolumeName(dcName string) string {
...
-	imageName, imageTag, _, err := ParseImageName(builderImage)
+	imageNS, imageName, imageTag, _, err := ParseImageName(builderImage)
 	if err != nil {
 		return errors.Wrap(err, "unable to parse image name")
 	}
 
 	var containerPorts []corev1.ContainerPort
 	if len(inputPorts) == 0 {
-		containerPorts, err = c.GetExposedPorts(imageName, imageTag)
+		imageNS, containerPorts, err = c.GetExposedPorts(imageNS, imageName, imageTag)
```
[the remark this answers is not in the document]
anmolbabu: Yes because, only in case of user chosen imagestream, it'll be non-empty from ParseImageName, else if user has not chosen Namespace, it needs to be resolved by GetImageStream function.
In case namespace is explicitly passed also, GetImageStream returns the passed namespace itself so, the overwrite is safe

pkg/occlient/occlient.go
```diff
@@ -425,18 +541,23 @@ func getAppRootVolumeName(dcName string) string {
...
 		if err != nil {
 			return errors.Wrapf(err, "unable to get exposed ports for %s:%s", imageName, imageTag)
 		}
 	} else {
+		imageStream, err := c.GetImageStream(imageNS, imageName, imageTag)
+		if err != nil {
+			return errors.Wrapf(err, "unable to create s2i app for %s", name)
+		}
+		imageNS = imageStream.ObjectMeta.Namespace
```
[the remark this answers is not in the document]
anmolbabu: Same as in https://github.com/redhat-developer/odo/pull/618#discussion_r211542412

pkg/occlient/occlient_test.go
```diff
@@ -1946,6 +1965,84 @@ func TestNewAppS2I(t *testing.T) {
...
+			fkclient, fkclientset := FakeNew()
+			fkclient.namespace = "testing"
+
+			fkclientset.ImageClientset.PrependReactor("get", "imagestreams", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, fakeImageStream(tt.imageName, "testing", []string{"latest"}), nil
+			})
+
+			fkclientset.ImageClientset.PrependReactor("get", "imagestreams", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, fakeImageStream(tt.imageName, "openshift", []string{"latest", "3.5"}), nil
+			})
```
anmolbabu: @kadel It looks like the prepend reactor calls here behave randomly and return incorrect images to incorrect namespaces. Can you please help me fix it...

pkg/occlient/occlient.go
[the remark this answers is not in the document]
kadel: It is weird and not really transparent when the function has the same argument that is also returned.
Especially when the function's responsibility is to just return ports. I think that function like this should expect that `imageNS` is always explicitly passed.

pkg/occlient/occlient.go
```diff
@@ -377,43 +391,145 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
...
-		for _, tag := range imageStream.Status.Tags {
-			// look for matching tag
-			if tag.Tag == imageTag {
-				tagFound = true
-				log.Debugf("Found exact image tag match for %s:%s", imageName, imageTag)
-				// ImageStream holds tag history
-				// first item is the latest one
+		errors.Wrapf(err, "error fetching exposed ports for image %s:%s", imageName, imageTag)
+	}
+	imageNS = imageStream.ObjectMeta.Namespace
```
kadel: I just don't understand it :-(
Here you are setting `imageNS` to the namespace of the object that you got at line 478 (`imageStream, err = c.GetImageStream(imageNS, imageName, imageTag)`), in that call you specify `imageNS` so `imageStream.ObjectMeta.Namespace` will already be the same namespace as you `imageNS` because you are requesting ImageStream from `imageNS` namespace.

pkg/occlient/occlient.go
```diff
@@ -377,43 +391,145 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
...
+		// look for matching tag
+		if tag.Tag == imageTag {
+			return true
+		}
+	}
+	return false
+}
+
+// GetImageNS returns the imagestream using image details
+func (c *Client) GetImageStream(imageNS string, imageName string, imageTag string) (*imagev1.ImageStream, error) {
```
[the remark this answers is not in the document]
kadel: Comment for this needs to explain that `imageNS` can be empty and what happens then

pkg/occlient/occlient.go
kadel: OK, I'm starting to understand `c.GetImageStream(imageNS, imageName, imageTag)` works even when `imageNS` is empty. 
To be honest, it is a bit weird :-| We need a lot of comments to make all this logic clear.

pkg/occlient/occlient_test.go
```diff
@@ -717,13 +721,15 @@ func TestParseImageName(t *testing.T) {
 
 	tests := []struct {
 		arg     string
+		want0   string
```
ashetty1: `want0` looks odd. Can we make it `want4`, or change the order.

pkg/occlient/occlient.go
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/occlient/occlient.go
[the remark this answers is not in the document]
anmolbabu: Ack..
But adds a call extra(2 calls everywhere(where GetExposedPorts was used) instead of one previously) :)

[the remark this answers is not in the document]
anmolbabu: Done

pkg/occlient/occlient.go
anmolbabu: Ack

pkg/occlient/occlient_test.go
anmolbabu: Ack

anmolbabu:
@surajnarwade Rebased

anmolbabu:
Verified as in:

1. NodeJS Component created using git url and required version but intelligently chosen imagestream from unspecified intelligently chosen namespace(current namespace in this test): https://gist.github.com/anmolbabu/8acfd835f2fec3f8dbb24398bd8ee958
2. NodeJS component created using git url and fully qualified imagestream(current namespace and version specified): https://gist.github.com/anmolbabu/540a0849c280d7aab3c5d5350e95a6c5
3. NodeJS component created using git url and unqualified(namespace and version) imagestream all defaults chosen(current namespace, and version latest): https://gist.github.com/anmolbabu/2667b71604036202c15f1c40ea3fd6c0
4. NodeJS Component created using local source and required version but intelligently chosen imagestream from unspecified intelligently chosen namespace(current namespace in this test): https://gist.github.com/anmolbabu/224c6797e5b913a5c64a72db1a5cb022
5. NodeJS component created using local source and fully qualified imagestream(current namespace and version specified): https://gist.github.com/anmolbabu/9c29c0ded801875a370dc0753bdfdab8
6. NodeJS component created using local source and unqualified(namespace and version) imagestream all defaults chosen(current namespace, and version latest): https://gist.github.com/anmolbabu/2c21ff404f0344ce883cba022ecf192c
7. Nodejs component created using git url and version qualifed(but not namespace) component and imagestreaam resolved automatically based on passed version to openshift namespace: https://gist.github.com/anmolbabu/eea952dd6863756e068cadecd7e144eb
8. Nodejs component created using local source and version qualifed(but not namespace) component and imagestreaam resolved automatically based on passed version to openshift namespace: : https://gist.github.com/anmolbabu/921f1391a9311b9688695608e459a028
9. Wildfly component created using binary and version qualified from current namespace: https://gist.github.com/anmolbabu/4a0fab66762f8fd940c6d77c4d7d408b
10. Wildfly component created using fully qualified imagestream and binary: https://gist.github.com/anmolbabu/b4e2bb290fa99eef55e94d2b1ef3647e

anmolbabu:
@cdrage  @surajnarwade @syamgk @ashetty1 @mik-dass Please review

pkg/occlient/occlient_test.go
```diff
@@ -1952,6 +1973,227 @@ func TestNewAppS2I(t *testing.T) {
...
+			imageTag:    "3.5",
+			want: []corev1.ContainerPort{
+				{
+					Name:          fmt.Sprintf("%d-%s", 8080, strings.ToLower(string("tcp"))),
+					ContainerPort: 8080,
+					Protocol:      "TCP",
+				},
+			},
+		},
+		/*
```
ashetty1: Reason for commenting this out?

pkg/occlient/occlient_test.go
```diff
@@ -1952,6 +1973,227 @@ func TestNewAppS2I(t *testing.T) {
...
+
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("client.GetExposedPorts = %#v, want %#v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestGetImageStream(t *testing.T) {
+	/*
```
ashetty1: Needs to be removed?

pkg/occlient/occlient_test.go
```diff
@@ -1952,6 +1973,227 @@ func TestNewAppS2I(t *testing.T) {
...
+									From: &corev1.ObjectReference{
+										Kind: "ImageStreamTag",
+										Name: "latest",
+									},
+									Name: "latest",
+								},
+							},
+						},
+					}
+	*/
```
ashetty1: needs to be removed?

pkg/occlient/occlient_test.go
```diff
@@ -1952,6 +1973,227 @@ func TestNewAppS2I(t *testing.T) {
...
+				}
+				if tt.imageNS == "testing" {
+					return true, currentNSIS, nil
+				}
+				if tt.imageNS == "openshift" {
+					return true, openshiftIS, nil
+				}
+				return true, nil, fmt.Errorf("Requested imagestream %s with tag %s not found", tt.imageName, tt.imageTag)
+			})
+
```
ashetty1: Also check for `fkclientset.ImageClientset.Actions`

anmolbabu: Aah commented it out while debugging something.
Will un-comment the test

pkg/occlient/occlient_test.go
anmolbabu: Aah commented it out while debugging something.
Will un-comment the test

pkg/occlient/occlient_test.go
anmolbabu: Yes

pkg/occlient/occlient_test.go
```diff
@@ -1952,6 +1973,227 @@ func TestNewAppS2I(t *testing.T) {
...
+			got := isTagInImageStream(tt.imagestream, tt.imageTag)
+
+			if got != tt.want {
+				t.Errorf("GetImageStream() = %#v, want %#v\n\n", got, tt)
+			}
+		})
+	}
+}
+func fakeImageStreamImage(imageName string, ports []string) *imagev1.ImageStreamImage {
```
ashetty1: Move helper functions to the top above all the test functions.

pkg/occlient/occlient_test.go
anmolbabu: Ack

pkg/occlient/occlient_test.go
anmolbabu: Ack

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
 			for _, component := range catalogList {
-				fmt.Fprintln(w, component.Name, "\t", strings.Join(component.Tags, ","))
+				component_name := component.Name
+				if component.Namespace == currentProject {
+					/*
+						 If current namespace is same as the current component namespace,
+						 	Loop through every other component,
+								If there exists a component with same name but in different namespaces,
+									 mark the one from current namespace with (*)
+					*/
```
cdrage: this comment block is weirdly spaced. your IDE may have done this.

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
+						}
+					}
+				}
+				fmt.Fprintln(w, component_name, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
-
+			fmt.Println(
+				`NOTE: There might be items in multiple namespaces.
+Items from current project (marked with *) will take precedence.
```
cdrage: `Items from the current project`

cmd/create.go
```diff
@@ -85,14 +86,14 @@ A full list of component types that can be deployed is available using: 'odo cat
 		// "Default" values
 		componentImageName := args[0]
 		componentType := args[0]
-		componentName := args[0]
+		componentName := util.ExtractComponentType(componentType)
```
cdrage: Shouldn't this be updated / modified in a different PR?

pkg/catalog/catalog.go
```diff
@@ -54,7 +57,7 @@ func Exists(client *occlient.Client, componentType string) (bool, error) {
 	}
 
 	for _, supported := range catalogList {
-		if componentType == supported.Name {
+		if componentType == supported.Name || componentType == fmt.Sprintf("%s/%s", supported.Namespace, supported.Name) {
```
cdrage: does this mean that we can supply (for example) using: `odo create openshift/nodejs:8.0.0 foobar` and it'd work?

cdrage: if so, this should be documented somewhere

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
-	imageStreams, err := client.GetImageStreams(occlient.OpenShiftNameSpace)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to get Image Streams")
+	var imageStreams []imagev1.ImageStream
+	currentNamespace := client.GetCurrentProjectName()
+	var err error
+
+	// Fetch imagestreams from default openshift namespace
+	openshiftNSImageStreams, e := client.GetImageStreams(occlient.OpenShiftNameSpace)
+	if e != nil {
```
cdrage: i feel as though `e` is a bad name for error, can't we just use `err`?

cmd/create.go
anmolbabu: This PR makes it possible to have component types as the following:
1. "myproject/python:3.5" -- Return python
2. "python:3.5" -- Return python
3. nodejs -- Return nodejs
So, it was essential to correctly identify the component type among other things(version, namespace) so, I added it as part of this PR

pkg/catalog/catalog.go
anmolbabu: > does this mean that we can supply (for example) using: odo create openshift/nodejs:8.0.0 foobar and it'd work?

Yes, where component name will be `foobar` and component is created using `nodejs` builderimagestream of version `8.0.0` picked up from `openshift` namespace.

> if so, this should be documented somewhere

Do you mean documenting in the command for use during cmd help?
Or you suggest putting it in a asciidoc?(If so, I'll raise a bug to track it so that the PR can be unblocked.. Would it be ok?)

pkg/catalog/catalog.go
anmolbabu: Not err because, I am using it as an overarching error in other words to recognize complete failure.
For example, in this case, its possible that a particular imagestream is not found in the namespace that is being searched, in which case, the other possible namespace needs to be searched also to conclude complete failure. So, this means the first failure is not necessarily actual failure. so, to differentiate the 2(complete failure and not/partial failure I needed it to be err and e separated). If I use the same var, the previous error if any, will be overwritten and the previous error will not be available

pkg/catalog/catalog.go
anmolbabu: @cdrage Done as in https://github.com/redhat-developer/odo/pull/673

surajnarwade:
cc @anmolbabu @mik-dass @syamgk please review this PR

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
 		case 0:
 			fmt.Printf("No deployable components found\n")
 		default:
-
+			currentProject := client.GetCurrentProjectName()
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(w, "NAME", "\t", "TAGS")
+			fmt.Fprintln(w, "NAME", "\t", "NAMESPACE", "\t", "TAGS")
```
surajnarwade: we should use `PROJECT` instead of `NAMESPACE`, user may get confused since we have `odo project xxxx` as well

anmolbabu: I have changed in user appearance areas. It seems like we have maintained it as namespace internally in code in a good number of places..

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
 			fmt.Printf("No deployable components found\n")
 		default:
-
+			currentProject := client.GetCurrentProjectName()
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(w, "NAME", "\t", "TAGS")
+			fmt.Fprintln(w, "NAME", "\t", "PROJECT", "\t", "TAGS")
 			for _, component := range catalogList {
-				fmt.Fprintln(w, component.Name, "\t", strings.Join(component.Tags, ","))
+				component_name := component.Name
```
surajnarwade: use camelCase, `componentName`

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
+				if component.Namespace == currentProject {
+					/*
+						If current namespace is same as the current component namespace,
+						Loop through every other component,
+						If there exists a component with same name but in different namespaces,
+						mark the one from current namespace with (*)
+					*/
+					for _, comp1 := range catalogList {
+						if comp1.Name == component.Name && component.Namespace != comp1.Namespace {
+							component_name = fmt.Sprintf("%s (*)", component.Name)
```
surajnarwade: camelCase

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
+						}
+					}
+				}
+				fmt.Fprintln(w, component_name, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
```
surajnarwade: camelCase

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
 			}
 			w.Flush()
-
+			fmt.Println(
+				`NOTE: There might be items in multiple namespaces.
```
surajnarwade: s/namespace/project

cmd/catalog.go
```diff
@@ -38,14 +38,32 @@ var catalogListCmd = &cobra.Command{
...
-				fmt.Fprintln(w, component.Name, "\t", strings.Join(component.Tags, ","))
+				component_name := component.Name
+				if component.Namespace == currentProject {
+					/*
+						If current namespace is same as the current component namespace,
+						Loop through every other component,
+						If there exists a component with same name but in different namespaces,
+						mark the one from current namespace with (*)
+					*/
+					for _, comp1 := range catalogList {
```
surajnarwade: s/comp1/comp

cmd/create.go
surajnarwade: Instead of writing this block twice, create one function which takes input and returns:
* componentImageName
* componentName
* componentType
* componentVersion

if not present, that function should fill whatever default values we have

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
 // getDefaultBuilderImages returns the default builder images available in the
-// openshift namespace
+// openshift and the current namespaces
 func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
-	imageStreams, err := client.GetImageStreams(occlient.OpenShiftNameSpace)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to get Image Streams")
+	var imageStreams []imagev1.ImageStream
+	currentNamespace := client.GetCurrentProjectName()
+	var err error
```
surajnarwade: no need to define error

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
+
+	// Fetch imagestreams from default openshift namespace
+	openshiftNSImageStreams, e := client.GetImageStreams(occlient.OpenShiftNameSpace)
```
surajnarwade: make it err instead of e

surajnarwade: you can reuse err

cmd/create.go
anmolbabu: But what is the gain the 2 blocks will only move to a different function instead of this one right?

pkg/catalog/catalog.go
anmolbabu: No please check https://github.com/redhat-developer/odo/pull/618/files/f397bb7134a87fa371a3616810a3c750186595fa#r213039396

pkg/catalog/catalog.go
anmolbabu: No please check https://github.com/redhat-developer/odo/pull/618/files/f397bb7134a87fa371a3616810a3c750186595fa#r213039396

cmd/catalog.go
anmolbabu: Ack

cmd/catalog.go
anmolbabu: Ack

cmd/catalog.go
anmolbabu: Ack

cmd/catalog.go
anmolbabu: Ack

cmd/catalog.go
anmolbabu: Ack

pkg/catalog/catalog.go
surajnarwade: use something else but not `e` :P

anmolbabu: possibleError seems good? :)

pkg/catalog/catalog.go
cdrage: @anmolbabu Not done as this information needs to be in `cmd/create.go` as well, not just external documentation

cdrage:
See my other comment, we need to update documentation in this PR if you're adding a new feature such as using `odo create namespace/nodejs`

pkg/catalog/catalog.go
anmolbabu: Ok Done

pkg/catalog/catalog_test.go
```diff
@@ -67,6 +67,9 @@ func TestVersionExist(t *testing.T) {
 				return true, testingutil.FakeImageStreams(tt.args.name, tt.args.namespace, tt.args.tags), nil
 			})
 
+			fakeClientSet.ImageClientset.PrependReactor("list", "imagestreams", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, testingutil.FakeImageStreams(tt.args.name, "openshift", tt.args.tags), nil
+			})
```
ashetty1: Is there a reason why we have two `PrependReactor`s defined here? Can't this be made generic?

ashetty1: Check for `fakeClientSet.ImageClientset.Action` missing

pkg/occlient/occlient_test.go
```diff
@@ -1955,6 +2002,179 @@ func TestNewAppS2I(t *testing.T) {
...
+			fkclient, fkclientset := FakeNew()
+			fkclient.namespace = "testing"
+
+			fkclientset.ImageClientset.PrependReactor("get", "imagestreamimages", func(action ktesting.Action) (bool, runtime.Object, error) {
+				return true, fakeImageStreamImage("python", []string{"8080/tcp"}), nil
+			})
+
+			got, err := fkclient.GetExposedPorts(tt.imagestream, tt.imageTag)
+
+			if !tt.wantErr == (err != nil) {
```
ashetty1: Check for actions missing

docs/getting-started.md
```diff
@@ -120,6 +120,18 @@ To push source code to the component run 'odo push'
...
+Note:
+
+By default builder imagestreams from current project are used for building your application.
+But, if the builder imagestream corresponding to the type of requested component does not exist in the current project,
+builder imagestreams from openshift namespace are also considered. But if you would like to specifically use the default
+builder imagestream, you can fully qualify the component type using $project/$component_type:$version.
+Accordingly following commands can also be used:
+* odo create openshift/nodejs:6 --local=.
+* odo create nodejs:6 --local=.
+
```
ashetty1: code block?

pkg/catalog/catalog_test.go
```diff
@@ -132,6 +135,16 @@ func TestList(t *testing.T) {
...
+		{
+			name: "Case 4: Valid image with output tags from current namespace",
+			args: args{
+				name:      "foobar",
+				namespace: "foo",
+				tags:      []string{"1.0.0", "1.0.1", "0.0.1", "latest"},
+			},
+			wantErr:  false,
+			wantTags: []string{"1.0.0", "1.0.1", "0.0.1", "latest"},
+		},
```
ashetty1: Check for Actions missing in the existing code. Would be good to add now.

pkg/catalog/catalog_test.go
anmolbabu: I am really not sure :)
May be added by mistake..
Removing it

anmolbabu: Ack adding

pkg/catalog/catalog_test.go
anmolbabu: Ack

pkg/occlient/occlient_test.go
anmolbabu: Ack

docs/getting-started.md
anmolbabu: Ack

syamgk:
good to go, once anush's comments on the unit test and documentation is added

cdrage:
@syamgk you need to hit "approve" to the suggested changes

cmd/catalog.go
```diff
@@ -50,14 +50,32 @@ var catalogListComponentCmd = &cobra.Command{
...
+							componentName = fmt.Sprintf("%s (*)", component.Name)
+						}
+					}
+				}
+				fmt.Fprintln(w, componentName, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
-
+			fmt.Println(
+				`NOTE: There might be items in multiple project.
```
cdrage: `multiple projects`

cmd/catalog.go
```diff
@@ -50,14 +50,32 @@ var catalogListComponentCmd = &cobra.Command{
...
+Items from the current project (marked with *) will take precedence.
+In case you want to use an item from a different namespace, please fully qualify it as $namespace/$name`,
```
cdrage: `If you want to use an item from a different namespace, use the full name as <NAMESPACE>/<NAME>`

sounds a bit better

cmd/create.go
```diff
@@ -29,8 +30,19 @@ var componentCreateCmd = &cobra.Command{
...
+The component type will be chosen by default from current project,
+but if specific version is requested for, it'll be used from either
+* current
+   or
+* openshift
+
+project(with search precendence in same order) where available.
+So, if you would like it to be specifically from openshift project, please fully qualify the
+component type as $project/$component_type:$version.
+If version is not specified by default, latest wil be chosen as the version.
```
cdrage: This is way too long and you need to shorten it to one or two sentences.

cmd/create.go
```diff
@@ -42,6 +54,9 @@ A full list of component types that can be deployed is available using: 'odo cat
   # Create new Node.js component with source from remote git repository.
   odo create nodejs --git https://github.com/openshift/nodejs-ex.git
 
+  # Create new Node.js component of version 6 using the builder image specifically from 'openshift' namespace
```
cdrage: `Create a new Node.js component`

cdrage: `from the 'openshift' namespace.`

cmd/create.go
```diff
@@ -189,6 +189,33 @@ A full list of component types that can be deployed is available using: 'odo cat
 	},
 }
 
+/*
+ parseCreateCmdArgs returns
+ 1. image name
+ 2. component type i.e, builder image name
+ 3. component name default value is component type else the user requested component name
+ 4. component version which is by default latest else version passed with builder image name
+*/
```
cdrage: you shouldn't use `/*` for multiple line comments with functions. Please change it to `//`

docs/getting-started.md
```diff
@@ -120,6 +120,18 @@ To push source code to the component run 'odo push'
...
+By default builder imagestreams from current project are used for building your application.
+But, if the builder imagestream corresponding to the type of requested component does not exist in the current project,
+builder imagestreams from openshift namespace are also considered. But if you would like to specifically use the default
+builder imagestream, you can fully qualify the component type using `$project/$component_type:$version`.
+Accordingly following commands can also be used:
+* `odo create openshift/nodejs:6 --local=.`
+* `odo create nodejs:6 --local=.`
+
+For more details please refer https://github.com/redhat-developer/odo/blob/master/docs/cli-reference.md#create
```
cdrage: Again, too much. You can shorten this to one or two sentences.

For example:

`By default, builder images will be used from the default OpenShift namespace. You can explicitly supply a namespace by using: `odo create <namespace>/nodejs:8`

pkg/catalog/catalog.go
cdrage: @anmolbabu that link no longer works. you shouldn't have to define error..

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
 func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
-	imageStreams, err := client.GetImageStreams(occlient.OpenShiftNameSpace)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to get Image Streams")
+	var imageStreams []imagev1.ImageStream
+	currentNamespace := client.GetCurrentProjectName()
+	var err error
+
+	// Fetch imagestreams from default openshift namespace
+	openshiftNSImageStreams, possibleError := client.GetImageStreams(occlient.OpenShiftNameSpace)
```
cdrage: possibleError sounds weird. why not just `err`?

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
+	if possibleError != nil {
+		// Tolerate the error as it might only be a partial failure
+		// We may get the imagestreams from other Namespaces
+		err = errors.Wrapf(possibleError, "unable to get Image Streams from namespace %s", occlient.OpenShiftNameSpace)
+		// log it for debugging purposes
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", occlient.OpenShiftNameSpace, possibleError.Error())
 	}
 
+	// Fetch imagestreams from current namespace
+	currentNSImageStreams, possibleError := client.GetImageStreams(currentNamespace)
```
cdrage: again, `possibleError` is unconventional, use `err`

pkg/catalog/catalog.go
```diff
@@ -87,13 +90,40 @@ func VersionExists(client *occlient.Client, componentType string, componentVersi
...
+		if err != nil {
+			// Failure to fetch images from any namespace, error out
+			return nil, errors.Wrapf(possibleError, "%s. unable to get Image Streams from namespace %s", err.Error(), currentNamespace)
+		}
+		// It is possible that the current namespace has no imagestreams -- a valid scenario
+		// But may be required for debugging purposes
+		err = errors.Wrapf(possibleError, "unable to get Image Streams from namespace %s", currentNamespace)
+		// log it for debugging purposes
+		glog.V(4).Infof("Unable to get Image Streams from namespace %s. Error %s", currentNamespace, possibleError.Error())
+	}
```
cdrage: Do you mind refactoring this section to two separate sections? The nested possibleError and err looks a bit weird.

pkg/catalog/catalog_test.go
```diff
@@ -132,6 +136,16 @@ func TestList(t *testing.T) {
 			wantErr:  true,
 			wantTags: []string{},
 		},
+		{
+			name: "Case 4: Valid image with output tags from current namespace",
```
cdrage: `from a different namespace`

pkg/occlient/occlient.go
```diff
@@ -377,42 +391,127 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
 	return names, nil
 }
 
-// GetExposedPorts retruns list of ContainerPorts that are exposed by given image
-func (c *Client) GetExposedPorts(imageName string, imageTag string) ([]corev1.ContainerPort, error) {
+func isTagInImageStream(is imagev1.ImageStream, imageTag string) bool {
```
cdrage: Comments need to be added for this function

pkg/occlient/occlient.go
```diff
@@ -377,42 +391,127 @@ func (c *Client) GetImageStreamsNames(namespace string) ([]string, error) {
...
+	for _, tag := range is.Status.Tags {
+		// look for matching tag
```
cdrage: `for a matching tag`

pkg/util/util.go
```diff
@@ -57,5 +58,22 @@ func NamespaceOpenShiftObject(componentName string, applicationName string) (str
...
+}
+
+// ExtractComponentType returns only component type part from passed component type(default unqualified, fully qualified, versioned, etc...and their combinations) for use as component name
+func ExtractComponentType(namespacedVersionedComponentType string) string {
+	/*
+		Possible types of parameters:
+		1. "myproject/python:3.5" -- Return python
+		2. "python:3.5" -- Return python
+		3. nodejs -- Return nodejs
+	*/
```
cdrage: these comments should be added above (where you describe ExtractComponentType

tests/e2e/e2e_test.go
```diff
@@ -182,12 +183,23 @@ var _ = Describe("odoe2e", func() {
 
 	Describe("creating a component", func() {
 		Context("when application exists", func() {
+			It("should be able to create new imagestream and find it in catalog list", func() {
+				curProj = strings.TrimSuffix(runCmd("oc project -q"), "\n")
+				cmd := fmt.Sprintf("oc create -f https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-centos7.json -n %s", curProj)
```
cdrage: it'd be nice to have this as a const / separate variable in-case it changes in the future. I'm referring to: https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-centos7.json

cdrage: maybe:

```go
testNamespacedImage := "https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-centos7.json"
```

cmd/catalog.go
anmolbabu: Ack

docs/getting-started.md
anmolbabu: Ack

cmd/create.go

cmd/create.go

cmd/create.go

pkg/catalog/catalog.go
anmolbabu: Ok I tried to simplify it this time..
Hope that looks ok

pkg/catalog/catalog.go
anmolbabu: Ok I tried to simplify it this time..
Hope that looks ok

pkg/catalog/catalog.go
anmolbabu: Done

pkg/catalog/catalog_test.go
anmolbabu: Ack

pkg/occlient/occlient.go
anmolbabu: Ack

pkg/occlient/occlient.go
anmolbabu: Ack

pkg/util/util.go
anmolbabu: Ack

tests/e2e/e2e_test.go
anmolbabu: Ack

pkg/occlient/occlient_test.go
```diff
@@ -1955,6 +2002,186 @@ func TestNewAppS2I(t *testing.T) {
...
+				t.Errorf("client.GetExposedPorts(imagestream imageTag) unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if len(fkclientset.ImageClientset.Actions()) != tt.wantActionCnt {
+				t.Errorf("expected 1 ImageClientset.Actions() in GetExposedPorts, got: %v", fkclientset.ImageClientset.Actions())
```
syamgk: this should be``` "expected %d ImageClientset.Actions() in GetExposedPorts, got: %v", tt.wantActionCnt, fkclientset.ImageClientset.Actions()```

pkg/occlient/occlient_test.go
```diff
@@ -1955,6 +2002,186 @@ func TestNewAppS2I(t *testing.T) {
...
+				}
+				if tt.imageNS == "openshift" {
+					return true, openshiftIS, nil
+				}
+				return true, nil, fmt.Errorf("Requested imagestream %s with tag %s not found", tt.imageName, tt.imageTag)
+			})
+
+			got, err := fkclient.GetImageStream(tt.imageNS, tt.imageName, tt.imageTag)
+			if len(fkclientset.ImageClientset.Actions()) != tt.wantActionsCnt {
+				t.Errorf("expected 1 ImageClientset.Actions() in GetImageStream, got %v", fkclientset.ImageClientset.Actions())
```
syamgk: same here

pkg/occlient/occlient_test.go
anmolbabu: Ack

pkg/occlient/occlient_test.go
anmolbabu: Ack

anmolbabu:
@ashetty1 Waiting for all reviews to complete.. Will squash once I get approvals from you and @cdrage

anmolbabu:
Re-Verified as in:

1. NodeJS Component created using git url and required version but intelligently chosen imagestream from unspecified intelligently chosen namespace(current namespace in this test): https://gist.github.com/anmolbabu/dba98896a886d1d32dd3580f3eab40a9
2. NodeJS component created using git url and fully qualified imagestream(current namespace and version specified): https://gist.github.com/anmolbabu/9bbd979fa8e6b1d8236fffb1d621efe5
3. NodeJS component created using git url and unqualified(namespace and version) imagestream all defaults chosen(current namespace, and version latest): https://gist.github.com/anmolbabu/abfef3907a0894de461ab881f5b136f7
4. NodeJS Component created using local source and required version but intelligently chosen imagestream from unspecified intelligently chosen namespace(current namespace in this test): https://gist.github.com/anmolbabu/2947c5120de0ccdad87d8857dd82da0b
5. NodeJS component created using local source and fully qualified imagestream(current namespace and version specified): https://gist.github.com/anmolbabu/0b4d2430e44d38a3f980ebfd45bc005f
6. NodeJS component created using local source and unqualified(namespace and version) imagestream all defaults chosen(current namespace, and version latest): https://gist.github.com/anmolbabu/8f8cfabc2bad50c0903ad5066d413953

cmd/catalog.go
```diff
@@ -50,14 +50,32 @@ var catalogListComponentCmd = &cobra.Command{
...
+							componentName = fmt.Sprintf("%s (*)", component.Name)
+						}
+					}
+				}
+				fmt.Fprintln(w, componentName, "\t", component.Namespace, "\t", strings.Join(component.Tags, ","))
 			}
 			w.Flush()
-
+			fmt.Println(
+				`NOTE: There might be items in multiple projects.
```
cdrage: Honestly, this looks a bit ugly / out of place:

```sh
github.com/redhat-developer/odo  pr_618 ✔                                                                                                                                                                                                                                                                                                                            9h24m  
▶ ./odo catalog list components
nodejs      openshift     0.10,4,6,8,latest
perl        openshift     5.16,5.20,5.24,latest
php         openshift     5.5,5.6,7.0,7.1,latest
python      openshift     2.7,3.3,3.4,3.5,3.6,latest                                                                                                                                                                                                                                                                                                                        
ruby        openshift     2.0,2.2,2.3,2.4,2.5,latest
wildfly     openshift     10.0,10.1,11.0,12.0,8.1,9.0,latest
NOTE: There might be items in multiple projects.
Items from the current project (marked with *) will take precedence.                                                                                                                                                                                                                                                                                                        
If you want to use an item from a different namespace, use the full name as <namespace>/<name>
```

I think that we should remove this note completely.

anmolbabu: Anyway removing it since we already also have it in our docs

cmd/create.go
```diff
@@ -29,8 +30,12 @@ var componentCreateCmd = &cobra.Command{
 
 If component name is not provided, component type value will be used for the name.
 
+By default, builder images will be used from the current namespace.
+You can explicitly supply a namespace by using: odo create namespace/name:version
+If version is not specified by default, latest wil be chosen as the version.
```
cdrage: Make this into one-line, example:

```sh
By default, builder images will be used from the current namespace. You can explicitly supply a namespace by using: odo create namespace/name:version
 ```

Other than that, this PR LGTM! Once you've made the above change, feel free to merge.

File: cmd/create.go [hunk 58]
```diff
@@ -189,6 +182,31 @@ A full list of component types that can be deployed is available using: 'odo cat
 	},
 }
 
+// parseCreateCmdArgs returns
+// 1. image name
+// 2. component type i.e, builder image name
+// 3. component name default value is component type else the user requested component name
+// 4. component version which is by default latest else version passed with builder image name
+func parseCreateCmdArgs(args []string) (string, string, string, string) {
```
cdrage: there are no tests written for this..

cdrage:
1 missing test and 1 comment regarding description, but other than that, LGTM. Once you've written the test, let's merge this in.

cmd/create.go

cdrage:
Thanks for being patient with the added tests! This LGTM! Thanks so much @anmolbabu

PR #622, opened 2018-08-07
SupervisorD Go binary implementation

### LINKED ISSUE #556: Use InitContainer to inject SupervisorD to container.
## Problem
Right new we are using BuildConfig and https://github.com/kadel/bootstrap-supervisored-s2i to inject supervisor to s2i image.
This process is slow and creates unnecessary BuildConfig.

## Solution
Use initContainer and Go implementation of supervisor (https://github.com/ochinchina/supervisord)

InitContainer will copy supervisor binary with config file into the shared volume.
This volume is shared between initContainer and application container.

You can see an example of initContainer and volumes  approach at this small PoC - https://gist.github.com/kadel/35fa7c64c79a845edfe85ada2706019b

1) Create container image
    Image should include:
        - SupervisorD binary and config File. Config file should include one program definition that executes `run.sh` script. Something similar to current [run.conf](https://github.com/kadel/bootstrap-supervisored-s2i/blob/master/configs/supervisord.d/run.conf)
        - utility scripts that we are already using in our current approach  https://github.com/kadel/bootstrap-supervisored-s2i/tree/master/scripts
          Some of the scripts might need to be updated
    Image should be as small as possible
    Push that image somewhere where can be accessed by anyone (for example quay.io/odo/....)

    **Most of this work has already been done at https://github.com/snowdrop/k8s-supervisor/tree/master/supervisord. We should reuse that. It also includes small Go program to generate supervisord config file according to env parameters.**

2) update BootstrapSupervisoredS2I
   - BuildConfig is no longer needed.
   - DeploymentConfig should have new emptyDir volume (supervisor volume) that is mounted to the "application container" and initContainer
   -  modify initContainer
      initContainer should do two things:
        - copy files from /opt/app-root to the appRoot volume using `copy-files-to-volume` script (this is what it is currently doing)
         - copy supervisorD binary and config file to the supervisor volume 

3) make necessary changes to `component.Update`  function to make `odo update` work with new implementation

Change will impact local and binary components, git component will remain the same (with BuildConfig).

cdrage:
This implements a new SupervisorD method for both local and binary
components. Rather than using BuildConfig which takes a considerable
amount of time, this uses multiple initContainer's in order to
successfully deploy source code to OpenShift.

The related code can be found within `pkg/occlient/templates.go`.

This commit also:
  - Removes BuildConfig dependency within `odo update`
  - Refactors `occlient.go` considerably by moving "template" code into
  `templates.go`
  - Refactors multiple functions within occlient.go and component.go to
  use commonObjectMeta more frequently, reducing the amount of
  parameters needed for certain functions.
  - Adds a new constant into the labels with Namespace
  `const ComponentTypeNamespace =
  "app.kubernetes.io/component-namespace"`
  - A new image is used (and provided under the `scripts/supervisord`
  directory) to deploy the supervisorD image
  - New functions such as: PatchCurrentDC which will correctly patch a
  DeploymentConfig but still contain / be consistent with volumes which
  were previously created with `odo create` 

Closes #556


cdrage:
@kadel This is the error I'm getting:

```sh
Volumes:
  shared-data:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
  test-app-s2idata:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  test-app-s2idata
    ReadOnly:   false
  default-token-s8dcz:
    Type:       Secret (a volume populated by a Secret)
    SecretName: default-token-s8dcz
    Optional:   false
QoS Class:      BestEffort
Node-Selectors: <none>
Tolerations:    <none>
Events:
  FirstSeen     LastSeen        Count   From                    SubObjectPath                                   Type            Reason                  Message
  ---------     --------        -----   ----                    -------------                                   --------        ------                  -------
  15s           15s             1       default-scheduler                                                       Normal          Scheduled               Successfully assigned test-app-1-q9nnx to localhost
  15s           15s             1       kubelet, localhost                                                      Normal          SuccessfulMountVolume   MountVolume.SetUp succeeded for volume "shared-data"
  15s           15s             1       kubelet, localhost                                                      Normal          SuccessfulMountVolume   MountVolume.SetUp succeeded for volume "pv0025"
  14s           14s             1       kubelet, localhost                                                      Normal          SuccessfulMountVolume   MountVolume.SetUp succeeded for volume "default-token-s8dcz"
  14s           14s             1       kubelet, localhost      spec.initContainers{supervisord-copy}           Normal          Pulling                 pulling image "docker.io/cdrage/supervisord-test:latest"
  13s           13s             1       kubelet, localhost      spec.initContainers{supervisord-copy}           Normal          Pulled                  Successfully pulled image "docker.io/cdrage/supervisord-test:latest"
  13s           13s             1       kubelet, localhost      spec.initContainers{supervisord-copy}           Normal          Started                 Started container
  13s           13s             1       kubelet, localhost      spec.initContainers{supervisord-copy}           Normal          Created                 Created container
  12s           12s             1       kubelet, localhost      spec.initContainers{copy-files-to-volume}       Normal          Pulling                 pulling image "docker-registry.default.svc:5000/openshift/nodejs:latest"
  12s           12s             1       kubelet, localhost      spec.initContainers{copy-files-to-volume}       Warning         Failed                  Failed to pull image "docker-registry.default.svc:5000/openshift/nodejs:latest": rpc error: code = Unknown desc = Get https://docker-registry.default.svc:5000/v1/_ping: dial tcp: lookup docker-registry.default.svc on 192.168.122.1:53: no such host
  12s           12s             1       kubelet, localhost      spec.initContainers{copy-files-to-volume}       Warning         Failed                  Error: ErrImagePull
  11s           11s             1       kubelet, localhost      spec.initContainers{copy-files-to-volume}       Normal          BackOff                 Back-off pulling image "docker-registry.default.svc:5000/openshift/nodejs:latest"
  11s           11s             1       kubelet, localhost      spec.initContainers{copy-files-to-volume}       Warning         Failed                  Error: ImagePullBackOff

```

The actual supervisord image works correctly, it's the internal template/service ones that do not :(

kadel:
@cdrage you probably have wrong registry address, as you are getting **no such host**  error.
Don't use openshift registry directly like this.

Important part is ImageChangeTrigger, this is what determines what image will be used for container. When deployment is triggered, OpenShift looks for triggers, and if it finds triger with matching container name it looks for the ImageStream that is specified in the trigger, and replaces `image` field in the container with resolved image from the ImageStream.

You can even set `image: " "` and it everything will work properly.

scripts/supervisord/assemble-and-restart
```diff
@@ -0,0 +1,34 @@
...
+
+# run s2i assemble script from source directory if it has been overridden  there
+if [ -f /opt/app-root/src/.s2i/bin/assemble ]; then
+    /opt/app-root/src/.s2i/bin/assemble
+else
+    /usr/libexec/s2i/assemble
+fi
+
+# restart run process
+/var/lib/supervisord/bin/supervisord ctl reload
```
kadel: This just reload the supervisord process, not the application that is supervisord "supervising"

You will probably need to run `ctl stop` and `ctl start`

GrahamDumpleton:
You can set image to image stream name in init container so long as the target image stream has local lookup policy enabled, or annotation has been added on deployment config to cause image name in init container be resolved against image streams.

So image stream either has:

```
                    "spec": {
                        "lookupPolicy": {
                            "local": true
                        }
                    }
```

or deployment/deployment config/stateful set, has:

```
      annotations:
        alpha.image.policy.openshift.io/resolve-names: '*'
```

See:

* https://docs.openshift.com/container-platform/3.9/dev_guide/managing_images.html#using-is-with-k8s

cdrage:
Okay! So here are some instructions on how to test this out before I disappear for 3-days next week.

First, bring up your cluster:

```sh
minishift start 
# or whatever cluster you have
```

With this implementation, `nodejs` is working correctly, so you can deploy with just a `git clone` and then `create`:

```sh
odo create nodejs --local ~/nodejs-ex
```

Because of the supervisord binary implementation, this is deployed immediately.

After the pods have changed to `Running` (you can check pending status using `oc get po` or even just do `odo push -v`, you can push:

```sh
odo push
# or verbose
odo push -v
```

All of this relies on an image that is pushed to Docker Hub with the contents from `script/supervisord`. You can change this by modifying the `cdrage/supervisord-test:latest" variable within the code to your own image.

There's some things that need more testing, but as of right now, nodejs seems to work well.

cdrage:
Anddd here's a video on how to do it:

https://www.youtube.com/watch?v=HoI61g1bDbc

cmoulliard:
I did a quick test using my testing [project](https://github.com/snowdrop/ocp-odo-build-install/blob/master/README.md) 

```
$GOPATH/bin/odo create openjdk18 sb1 --local ./src
Please wait, creating sb1 component ...
Component 'sb1' was created and ports 8778/TCP,8080/TCP,8443/TCP were opened

Component 'sb1' is now set as active component.
```

and when I do the odo push, I get this error

```
$GOPATH/bin/odo push
Pushing changes to component: sb1
Please wait, building component....
+ set -eo pipefail
+ mkdir -p /tmp/src
+ touch /tmp/src/.dummy
++ find /opt/rh/ -name bin -type d
++ grep /root/usr/bin
+ for i in '$(find /opt/rh/ -name '\''bin'\'' -type d | grep '\''/root/usr/bin'\'')'
+ PATH=/opt/rh/rh-maven35/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/s2i
+ PATH=/usr/bin:/opt/rh/rh-maven35/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/s2i
+ HOME=/opt/app-root
+ fix-permissions /opt/app-root/src
/var/lib/supervisord/bin/assemble-and-restart: line 30: fix-permissions: command not found
failed to push component: sb1
```

pkg/occlient/occlient.go
```diff
@@ -740,8 +745,9 @@ func (c *Client) CreateService(commonObjectMeta metav1.ObjectMeta, containerPort
 func addBootstrapInitContainer(dc *appsv1.DeploymentConfig, dcName string) {
 	dc.Spec.Template.Spec.InitContainers = append(dc.Spec.Template.Spec.InitContainers,
 		corev1.Container{
-			Name:  "copy-files-to-volume",
-			Image: dc.Spec.Template.Spec.Containers[0].Image,
+			Name: "copy-files-to-volume",
+			// TODO need to use custom image here (copy files over + combination of supervisord image too)
```
kadel: what custom image?

pkg/occlient/occlient.go
```diff
@@ -755,6 +761,44 @@ func addBootstrapInitContainer(dc *appsv1.DeploymentConfig, dcName string) {
...
+			Args: []string{
+				"-r",
+				"/opt/supervisord",
+				"/var/lib/",
+			},
+			// TODO : The following list should be calculated based on the labels of the S2I image, otherwise, we should be
+			// retrieveing the arguments from either the inside of the container OR from an OpenShift label/annotation
+			Env: []corev1.EnvVar{
+				{
+					Name:  "CMDS",
```
kadel: Where is `$CMDS` variable used? Do we need it?

scripts/supervisord/Dockerfile
```diff
@@ -0,0 +1,30 @@
+#
+# This is an "initContainer" image using the base "source-to-image" OpenShift template
+# in order to apprpriately inject the supervisord binary into the application container.
+#
+# We use the base s2i image in order to use binaries such as "fix-permissions" to
+# add approriate file/folder permissions to nodejs / ruby / python applications.
+# 
+
+FROM centos/s2i-base-centos7
```
kadel: why are we using s2i base image? We don't need any s2i capability in this image, or do we?
We should probably use something really tiny something like busybox, as the time it takes to create new component will mostly depend on the time it takes to pull this image.

scripts/supervisord/assemble-and-restart
```diff
@@ -0,0 +1,42 @@
...
+# We now run the assembly script. If there is a custom one written in the 
+# source files, we use that instead.
+if [ -f /opt/app-root/src/.s2i/bin/assemble ]; then
+    /opt/app-root/src/.s2i/bin/assemble
+else
+    /usr/libexec/s2i/assemble
+fi
+
+# Restart supervisord in order to actualy run the application
+/var/lib/supervisord/bin/supervisord ctl reload
```
kadel: why are we reading supervisor here?

pkg/occlient/occlient.go
```diff
@@ -755,6 +761,44 @@ func addBootstrapInitContainer(dc *appsv1.DeploymentConfig, dcName string) {
...
 }
 
+// addBootstrapSupervisordInitContainer creates an init container that will copy over
+// supervisord to the application image during the start-up procress.
+func addBootstrapSupervisordInitContainer(dc *appsv1.DeploymentConfig, dcName string) {
+
+	// TODO: Use a release process to use an appriate image based on the release (0.1.0, HEAD, etc.)
+	// TODO: Maybe even make this a configurable container (for development at least too), since this will
+	// be pulled publically
+	image := "docker.io/cdrage/supervisord-test:latest"
```
kadel: it might be better use quay.io 
I've already created openshiftdo org there. If you send me your username on quay.io I'll give you rights.

scripts/supervisord/run
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -x
+set -eo pipefail
+
+# HACK. For some reason, *somewhere* $PATH is being overriden and the default
+# "/opt/rh/" is not being added. We instead manually search for the binaries and add it
+# to $PATH.
+for i in $(find /opt/rh/ -name 'bin' -type d | grep '/root/usr/bin'); do
```
kadel: this is weird, we shouldn't need it.

scripts/supervisord/echo.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
kadel: what is this file?

scripts/supervisord/Dockerfile
cmoulliard: This is no needed and could start from busybox as we did here -> https://goo.gl/fbV7GS

scripts/supervisord/assemble-and-restart
```diff
@@ -0,0 +1,42 @@
...
+# REMOVE THIS BEFORE MERGING.
+for i in $(find /opt/rh/ -name 'bin' -type d | grep '/root/usr/bin'); do
+        PATH=$i:$PATH
+done
+
+# Override PATH for the assemble script so that /usr/bin precedes
+# /opt/app-root/bin so wrong Python version isn't found during setup
+# if using Python S2I base image.
+# as per: https://github.com/openshift-evangelists/pseudo-vps-quickstart/blob/master/.s2i/bin/assemble#L7
+PATH=/usr/bin:$PATH
```
kadel: We shouldn't need that as we no longer use Python SupervisorD

scripts/supervisord/assemble-and-restart
```diff
@@ -0,0 +1,42 @@
...
+# S2i fails when /tmp/src/ is empty. 
+# Here we will create an empty folder / dummy file
+mkdir -p /tmp/src
+touch /tmp/src/.dummy
+
+# HACK. For some reason, *somewhere* $PATH is being overriden and the default
+# "/opt/rh/" is not being added. We instead manually search for the binaries and add it
+# to $PATH.
+# REMOVE THIS BEFORE MERGING.
+for i in $(find /opt/rh/ -name 'bin' -type d | grep '/root/usr/bin'); do
```
kadel: this is weird we shouldn't do that.

scripts/supervisord/assemble
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
kadel: Do we need this script?
It looks like we are not using this script anywhere.

scripts/supervisord/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+ENV SUPERVISORD_DIR /opt/supervisord
+
+RUN mkdir -p ${SUPERVISORD_DIR}/conf ${SUPERVISORD_DIR}/bin
+
+ADD supervisor.conf ${SUPERVISORD_DIR}/conf/
```
cmoulliard: Why have you removed the possibility to customize the commands exposed by the supervisord as we did [here](https://github.com/snowdrop/k8s-supervisor/blob/master/supervisord/conf/supervisord.tmpl#L1-L6) and populated using env var and by example this string pf CMDS `run-java:/usr/local/s2i/run;compile-java:/usr/local/s2i/assemble` ?

scripts/supervisord/echo.sh
cmoulliard: It can be removed as it was part of our k8s-supervisor prototype project ;-)

scripts/supervisord/Dockerfile
kadel: I would like from this PR to just migrate what we are doing currently to initContianer approach with Go SupervisorD. And nothing else. So we can keep this PR small or smallish. 
We can work on improvements like this afterward.

scripts/supervisord/assemble
kadel: OK, I can see that this is used in Dockerfile for supervisord image.
But still failing to understand why we would need to do anything with /opt/appr-root in it

scripts/supervisord/Dockerfile
cdrage: @kadel I included using the s2i base image as it has the `fix-permissions` variable.

You are right however, we can use `busybox` or `centos` instead.

scripts/supervisord/run
cdrage: Yeah.. It's a hack right now since I was debugging the $PATH issues. (since we don't copy files over from `/opt/app-root` path.

scripts/supervisord/assemble-and-restart
cdrage: The above comments I added describe why, and that's because of the $PATH issues we were having before. `/usr/bin` must be before everything else in order for the Python S2I to work. I will remove this after I've found the bug.

kadel:
> cdrage deleted the cdrage:implement-supervisord  branch 13 hours ago

?

metacosm:
Why was this closed?

kadel:
I suspect that @cdrage  deleted his branch by mistake.

geoand:
Alarm bells went off, since we are really looking forward to this work :P

cdrage:
Yeah, I didn't close this locally. Odd! It's restored. I'm still working on it.

geoand:
:relieved:

cdrage:
Update:
 - I'm encountering a log of failed storage commands. Looks like having the `share-data` volume with no PVC is messing a few things up. Just going through and bugfixing them.

cmoulliard:
We still have 3 issues to be fixed for this PR. When we do a binary upload, then it fails as the path to copy the binary points to a file `/opt/app-root-src` and not a directory (see #703).
The commands defined to run or assemble the application must be changed too (see #707)

cdrage:
Tests pass (locally at least for me).

Feel free to test / have a look at this implementation! 

It's as simple as:

```sh
git clone https://github.com/openshift/django-ex.git
./odo create python --local django-ex
./odo push
```

pkg/component/component.go
```diff
@@ -367,68 +389,112 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
...
-			return errors.Wrapf(err, "unable to build the component %v", componentName)
+			return errors.Wrapf(err, "unable to delete BuildConfig for %s component", componentName)
 		}
 
-		err = client.SetupForSupervisor(namespacedOpenShiftObject, projectName, annotations, labels)
-		if err != nil {
-			return errors.Wrapf(err, "unable to update DeploymentConfig  for %s component", componentName)
-		}
+		// Update the DeploymentConfig
+		err = client.UpdateDCToSupervisor(commonObjectMeta, componentImageType)
```
golangcibot: ineffectual assignment to `err`

pkg/occlient/occlient.go
```diff
@@ -525,23 +535,27 @@ func getAppRootVolumeName(dcName string) string {
...
 // gitUrl is the url of the git repo
 // inputPorts is the array containing the string port values
-func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labels map[string]string, annotations map[string]string, inputPorts []string) error {
+func (c *Client) NewAppS2I(commonObjectMeta metav1.ObjectMeta, builderImage string, gitUrl string, inputPorts []string) error {
 
+	glog.V(4).Infof("Using BuilderImage: %s", builderImage)
 	imageNS, imageName, imageTag, _, err := ParseImageName(builderImage)
 	if err != nil {
-		return errors.Wrap(err, "unable to parse image name")
+		return errors.Wrap(err, "unable to parse image commonObjectMeta.Name")
```
kadel: this doesn't look right
 search and replace? ;-)

pkg/occlient/occlient.go
```diff
@@ -560,125 +574,59 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitUrl string, labe
...
-				},
-			},
-		},
-	}
-	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)
+	// Retrieve BuildConfig image commonObjectMeta.Name for the DeploymentConfig
+	buildConfig, err := c.GetBuildConfigFromName(commonObjectMeta.Name, c.GetCurrentProjectName())
 	if err != nil {
-		return errors.Wrapf(err, "unable to create BuildConfig for %s", name)
+		return errors.Wrap(err, "unable to retrieve BuildConfig image commonObjectMeta.Name")
```
kadel: this message is also strange

cdrage:
Tests are all :green_heart:

cdrage:
NOTE (for myself): Make sure to keep an eye out on https://github.com/redhat-developer/odo/pull/626 so that `env` variables are persistent across `odo update`

pkg/occlient/occlient.go
```diff
@@ -59,9 +59,19 @@ const (
...
+
+	// Git repository that will be used for bootstraping
+	// TODO remove
 	bootstrapperURI = "https://github.com/kadel/bootstrap-supervisored-s2i"
 	bootstrapperRef = "v0.0.2"
+
+	// Docker image that will be use containing the supervisord binary as well
+	// as the assemble / run / restart scripts
+	// TODO: Move to Quay.io before we merge (discussion release versions, etc.)
+	bootstrapperImage = "quay.io/geoand/supervisord-test:latest"
```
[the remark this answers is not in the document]
kadel: We should also reference image by its hash, or at least by tag. But has would be better ;-)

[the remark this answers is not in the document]
cdrage: @kadel Yeah, I will push the image there :) I just had this here for testing purposes, hence the TODO haha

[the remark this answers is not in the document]
kadel: ah, missed the TODO 😊

pkg/occlient/occlient.go
```diff
@@ -950,6 +763,212 @@ func (c *Client) UpdateBuildConfig(buildConfigName string, projectName string, g
...
+	// Setup PVC
+	_, err = c.CreatePVC(getAppRootVolumeName(commonObjectMeta.Name), "1Gi", commonObjectMeta.Labels)
+	if err != nil {
+		return errors.Wrapf(err, "unable to create PVC for %s", commonObjectMeta.Name)
+	}
+
+	return nil
+}
+
+/*
```
kadel: can we delete this?

pkg/component/component.go
```diff
@@ -123,7 +131,15 @@ func CreateFromPath(client *occlient.Client, name string, componentImageType str
...
-	err = client.BootstrapSupervisoredS2I(namespacedOpenShiftObject, componentImageType, labels, annotations, inputPorts)
+	// Create CommonObjectMeta to be passed in
+	commonObjectMeta := metav1.ObjectMeta{
+		Name:        namespacedOpenShiftObject,
+		Labels:      labels,
+		Annotations: annotations,
+	}
+
+	// Bootstrap the deployment with SupervisorD
+	err = client.BootstrapSupervisoredS2I(commonObjectMeta, componentImageType, inputPorts)
```
anmolbabu: :+1:

pkg/component/component.go
```diff
@@ -367,104 +387,144 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
...
+			return errors.Wrapf(err, "unable to update DeploymentConfig for %s component", componentName)
 		}
+
 	} else {
 		// save source path as annotation
 		// this part is for updates where the source does not change or change from local to binary and vice versa
 
-		var annotations map[string]string
 		if newSourceType == "git" {
-			annotations = map[string]string{componentSourceURLAnnotation: newSource}
```
anmolbabu: Don't we need this anymore?
Bcoz the comment says this can be also be the case "change from local to binary and vice versa" apart from the other possibility "updates where the source does not change".

pkg/component/component.go
```diff
@@ -367,104 +387,144 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
...
-			annotations = map[string]string{componentSourceURLAnnotation: sourceURL}
-			annotations[componentSourceTypeAnnotation] = newSourceType
-			err = client.UpdateBuildConfig(namespacedOpenShiftObject, projectName, "", annotations)
-			if err != nil {
-				return errors.Wrapf(err, "unable to update the build config %v", componentName)
-			}
+			annotations[componentSourceURLAnnotation] = sourceURL
 			err = client.UpdateDCAnnotations(namespacedOpenShiftObject, annotations)
 		}
+
```
anmolbabu: can we move the line `err = client.UpdateDCAnnotations(namespacedOpenShiftObject, annotations)` here? instead of https://github.com/redhat-developer/odo/pull/622/files#diff-692effa1b0b7a9ed8c1f7bb49a977b9aR512 and https://github.com/redhat-developer/odo/pull/622/files#diff-692effa1b0b7a9ed8c1f7bb49a977b9aR525

pkg/occlient/occlient_test.go
```diff
@@ -1611,52 +1611,6 @@ func TestUpdateBuildConfig(t *testing.T) {
 		updatedBuildConfig  buildv1.BuildConfig
 		wantErr             bool
 	}{
-		{
-			name:            "git to local with proper parameters",
```
anmolbabu: Its possible that I am missing something but, Why this test was removed?

pkg/occlient/templates.go
```diff
@@ -0,0 +1,289 @@
...
+			Name: "copy-files-to-volume",
+			// Using custom image from bootstrapperImage variable for the initial initContainer
+			Image: dc.Spec.Template.Spec.Containers[0].Image,
+			Command: []string{
+				"sh",
+				"-c"},
+			// Script required to copy over file information from /opt/app-root
+			// Source https://github.com/jupyter-on-openshift/jupyter-notebooks/blob/master/minimal-notebook/setup-volume.sh
+			Args: []string{`
+SRC=/opt/app-root
```
anmolbabu: Can we move this script to a separate file instead of hard coding here for better maintainability?
Moreover, can the script be copied over instead of being sent as string?

anmolbabu:
I assume "work in progress" label is due to more UTs coming up?

I did very less and basic testing today will do more testing tomorrow...

pkg/component/component.go
```diff
@@ -244,7 +263,8 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 	}()
 
 	err = client.ExecCMDInContainer(pod.Name,
-		[]string{"/opt/app-root/bin/assemble-and-restart.sh"},
+		// We will use the assemble-and-restart script located within the supervisord container we've created
```
surajnarwade: `assemble-and-restart binary`

pkg/occlient/occlient.go
```diff
@@ -59,9 +59,14 @@ const (
...
 	// The length of the string to be generated for names of resources
 	nameLength = 5
-	// git repository that will be used for bootstraping
-	bootstrapperURI = "https://github.com/kadel/bootstrap-supervisored-s2i"
-	bootstrapperRef = "v0.0.2"
+
+	// Docker image that will be use containing the supervisord binary as well
+	// as the assemble / run / restart scripts
+	// TODO: Move to Quay.io before we merge (discussion release versions, etc.)
+	bootstrapperImage = "docker.io/cdrage/supervisord-test:latest"
```
surajnarwade: use quay as per TODO
(unless this codeclimate bot will  :cry: )

anmolbabu:
Verified as in https://gist.github.com/anmolbabu/862f9b8cd4dd767cf6c6adf0400018cb
based on the demo video at https://www.youtube.com/watch?v=HoI61g1bDbc

surajnarwade:
* I created component using `git` as a source:

```
 odo create nodejs mynode --git=https://github.com/sclorg/nodejs-ex
```

* I updated the source to local,
 
```
$ odo update --local=.
The component mynode was updated successfully, please use 'odo push' to push your local changes
```

* I tried to push the code but it crashed,
```
$ odo push -v 4
I1004 14:43:36.136227   29569 occlient.go:264] Trying to connect to server 192.168.42.49:8443
I1004 14:43:36.136875   29569 occlient.go:270] Server https://192.168.42.49:8443 is up
I1004 14:43:36.155102   29569 occlient.go:247] isLoggedIn err:  <nil> 
 output: "developer"
I1004 14:43:36.155383   29569 push.go:45] No component name passed, assuming current component
Pushing changes to component: mynode
I1004 14:43:36.155625   29569 occlient.go:1680] Getting DeploymentConfig: mynode-app
I1004 14:43:36.170104   29569 component.go:382] Source for component mynode is file:///home/snarwade/odo-example/nodejs-ex (local)
I1004 14:43:36.170151   29569 push.go:93] Copying directory /home/snarwade/odo-example/nodejs-ex to pod
I1004 14:43:36.179717   29569 occlient.go:1066] Waiting for deploymentconfig=mynode-app pod
I1004 14:43:36.186904   29569 occlient.go:1081] Status of mynode-app-2-nttb4 pod is Pending
I1004 14:43:36.187002   29569 occlient.go:1081] Status of mynode-app-1-99xrq pod is Running
I1004 14:43:36.187012   29569 occlient.go:1084] Pod mynode-app-1-99xrq is running.
I1004 14:43:36.187088   29569 component.go:249] Copying to pod mynode-app-1-99xrq
Please wait, building component....
rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:247: starting container process caused "exec: \"/var/lib/supervisord/bin/assemble-and-restart\": stat /var/lib/supervisord/bin/assemble-and-restart: no such file or directory"

I1004 14:43:36.404275   29569 root.go:175] Error:
unable to execute assemble script: error while streaming command: command terminated with exit code 126
failed to push component: mynode
```

* After a while, I tried and it worked.

REASON: it takes time to create new deploymentconfig as we change the source, (I think `odo update` should wait till the pod is up and running)

mik-dass:
@cdrage Almost same error as @surajnarwade when following the above scenario, my logs are a bit different. In my case it fails with `read/write on closed pipe`
```
[mrinaldas@localhost odo]$ ./odo push -v 4
I1004 15:21:10.123933   13388 occlient.go:264] Trying to connect to server 127.0.0.1:8443
I1004 15:21:10.124219   13388 occlient.go:270] Server https://127.0.0.1:8443 is up
I1004 15:21:10.135697   13388 occlient.go:247] isLoggedIn err:  <nil> 
 output: "developer"
I1004 15:21:10.135897   13388 push.go:45] No component name passed, assuming current component
Pushing changes to component: nodejs
I1004 15:21:10.136070   13388 occlient.go:1680] Getting DeploymentConfig: nodejs-app
I1004 15:21:10.143926   13388 component.go:382] Source for component nodejs is file:///home/mrinaldas/nodejs-ex (local)
I1004 15:21:10.143965   13388 push.go:93] Copying directory /home/mrinaldas/nodejs-ex to pod
I1004 15:21:10.150515   13388 occlient.go:1066] Waiting for deploymentconfig=nodejs-app pod
I1004 15:21:10.155108   13388 occlient.go:1081] Status of nodejs-app-6-brrv7 pod is Running
I1004 15:21:10.155140   13388 occlient.go:1084] Pod nodejs-app-6-brrv7 is running.
I1004 15:21:10.155178   13388 component.go:249] Copying to pod nodejs-app-6-brrv7
E1004 15:21:10.261045   13388 v2.go:163] io: read/write on closed pipe
I1004 15:21:10.300495   13388 root.go:175] Error:
unable push files to pod: error while streaming command: command terminated with exit code 2
failed to push component: nodejs
```

pkg/component/component.go
kadel: why binary?

pkg/occlient/templates.go
kadel: nope :-( it has to be there we can't put it to the image because the image is a standard s2i image that we don't want to modify.
Maybe in the future, we will figure out how to solve this in the better way, but currently, this is the best we could come up with

pkg/component/component.go
surajnarwade: oh sorry I thought this is binary, my bad

mik-dass:
> That is probably the same as in #647 

I have a similar feeling too. I had posted my findings here https://github.com/redhat-developer/odo/issues/647#issuecomment-418745579

pkg/occlient/occlient.go
```diff
@@ -949,6 +761,160 @@ func (c *Client) UpdateBuildConfig(buildConfigName string, projectName string, g
...
+		if container.Name == name {
+			// Loop through all the volumes
+			for _, volume := range foundCurrentDCContainer.VolumeMounts {
+				// If it's the supervisord volume, ignore it.
+				if volume.Name == supervisordVolumeName {
+					continue
+				} else {
+					dc.Spec.Template.Spec.Containers[index].VolumeMounts = append(dc.Spec.Template.Spec.Containers[index].VolumeMounts, volume)
+				}
+			}
```
anmolbabu: Can we break from here? Outside the inner for but inside if and after the inner for?

pkg/occlient/occlient.go
```diff
@@ -59,9 +59,14 @@ const (
 
 	// The length of the string to be generated for names of resources
 	nameLength = 5
-	// git repository that will be used for bootstraping
-	bootstrapperURI = "https://github.com/kadel/bootstrap-supervisored-s2i"
-	bootstrapperRef = "v0.0.2"
+
+	// Docker image that will be use containing the supervisord binary as well
```
syamgk: Docker image that will be used ?

syamgk:
build locally and deployed nodejs and python app everything went well

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2937,36 @@ func TestGetServiceInstanceList(t *testing.T) {
...
-			client, fakeClientSet := FakeNew()
-			fakeClientSet.ServiceCatalogClientSet.PrependReactor("list", "clusterserviceclasses", func(action ktesting.Action) (bool, runtime.Object, error) {
-				return true, &classList, nil
-			})
-
-			fakeClientSet.ServiceCatalogClientSet.PrependReactor("list", "clusterserviceplans", func(action ktesting.Action) (bool, runtime.Object, error) {
-
-				return true, &planList, nil
-			})
+			fakeClient, fakeClientSet := FakeNew()
```
golangcibot: ineffectual assignment to `fakeClient`

pkg/component/component.go
```diff
@@ -371,104 +391,144 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
...
+
+		// CreateBuildConfig here!
+		glog.V(4).Infof("Creating BuildConfig %s using imageName: %s for updating", namespacedOpenShiftObject, imageName)
+		err = client.CreateBuildConfig(commonObjectMeta, imageName, newSource)
 		if err != nil {
 			return errors.Wrapf(err, "unable to update BuildConfig  for %s component", componentName)
 		}
+
+		// Get the current BuildConfig since we're going to update DeploymentConfig
+		bc, err := client.GetBuildConfigFromName(namespacedOpenShiftObject, projectName)
```
mik-dass: We can retrieve the build Config here from the above create call as suggested in https://github.com/redhat-developer/odo/pull/622/files#r222987807

pkg/occlient/occlient.go
```diff
@@ -559,125 +567,59 @@ func (c *Client) NewAppS2I(name string, builderImage string, gitURL string, labe
...
-							Namespace: imageNS,
-						},
-					},
-				},
-			},
-		},
-	}
-	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)
+	// Retrieve BuildConfig image commonObjectMeta.Name for the DeploymentConfig
+	buildConfig, err := c.GetBuildConfigFromName(commonObjectMeta.Name, c.GetCurrentProjectName())
```
mik-dass: same as https://github.com/redhat-developer/odo/pull/622#discussion_r222988173

pkg/occlient/occlient.go
```diff
@@ -2153,3 +2155,39 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
...
+	if err != nil {
+		return errors.Wrap(err, "unable to retrieve image stream for CreateBuildConfig")
+	}
+	imageNS = imageStream.ObjectMeta.Namespace
+
+	glog.V(4).Infof("Using namespace: %s for the CreateBuildConfig function", imageNS)
+
+	// Use BuildConfig to build the container with Git
+	bc := generateBuildConfig(commonObjectMeta, gitURL, imageName+":"+imageTag, imageNS)
+	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)
```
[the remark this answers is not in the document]
cdrage: @mik-dass Bit confused.

So you want CreateBuildConfig to return `bc` and then use `	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)` more up in the code?

@kadel WDYT?

pkg/component/component.go
cdrage: Nope. We do this higher up in the code.

pkg/component/component.go
cdrage: Can't because we use `build` before we update the annotations. Best to update the annotations before we start building or else it may cause errors if users use `odo create` or `odo update` during an update / build. If that makes sense.

pkg/occlient/occlient_test.go
cdrage: We no longer use `BuildConfig` which this test was testing for. That's why it was removed.

pkg/component/component.go
cdrage: Ahhhh I see what you mean. I'm going to update it.

pkg/component/component.go
anmolbabu: ok so you mean here : https://github.com/redhat-developer/odo/pull/622/files#diff-692effa1b0b7a9ed8c1f7bb49a977b9aR412

pkg/occlient/occlient.go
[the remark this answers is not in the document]
mik-dass: @cdrage No just return the newly created build config 
`bc, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)`
and then use the returned fresh build config for further tasks. That way you won't need to call `GetBuildConfigFromName()` to retrieve it later as suggested in https://github.com/redhat-developer/odo/pull/622/files/beae1afab994d1278829e4e1b9cff3f81ec781fa#r222988173

pkg/component/component.go
anmolbabu: Hmm I see the build is triggered (only) if newSourceType is git

pkg/occlient/occlient_test.go
cdrage: Yes, the use-case is still valid but a different test should be written instead that doesn't check BuildConfig. The current test uses: `		updatedBuildConfig  buildv1.BuildConfig
` as an argument and that's no longer valid with updating from Git to Local

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,79 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+
+			// Error checking PatchCurrentDC
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.PatchCurrentDC() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Client.GetClusterServiceClassExternalNames() = %v, want %v", got, tt.want)
+
+			// Check to see how many actions are being ran
+			if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient.go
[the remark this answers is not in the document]
kadel: @mik-dass your link is now working, so I'm not sure what you are referring to. :-(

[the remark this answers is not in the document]
mik-dass: @kadel That comment seems to be resolved and new changes have been pushed, so the link won't work anymore

pkg/occlient/templates.go
```diff
@@ -0,0 +1,289 @@
...
+DEST=/mnt/app-root
+
+if [ -f $DEST/.delete-volume ]; then
+    rm -rf $DEST
+fi
+ if [ -d $DEST ]; then
+    if [ -f $DEST/.sync-volume ]; then
```
kadel: we can remove all JUPYTER related stuff

scripts/supervisord/assemble-and-restart
```diff
@@ -0,0 +1,19 @@
...
+set -x
+set -eo pipefail
+
+# S2i fails when /tmp/src/ is empty. 
+mkdir -p /tmp/src
+touch /tmp/src/.dummy
+
+# We now run the assembly script. If there is a custom one written in the 
+# source files, we use that instead.
+if [ -f /opt/app-root/src/.s2i/bin/assemble ]; then
```
geoand: I have obviously missed some discussion about it, but could you please let me know what happened to the java s2i stuff? 

Thanks

kadel: @geoand What java s2i stuff?

kadel: @geoand 

> A couple weeks ago I opened a PR against @cdrage 's repo (and said PR got merged) that is the source of this PR, which contained a couple fixes to the bash scripts that made them work with the Java s2i image.
> 
Oh right, I remember now.

> Those changes seem to be gone which is why I am asking :). I am sure that I have just missed some discussion about it, so I just wanted to catch up.

I don't think that you missed something. There was no discussion after we decided that we will use your fix in this PR. 

It is strange that it is not there :-(

geoand: Thanks @kadel !

@cdrage would you like me to open a new PR against your repo if the changes were made unintentionally?

kadel:
I've just added commit https://github.com/redhat-developer/odo/pull/622/commits/c438d0486216509f4ac1db85d9696e7a1fa7c048 that moves a directory from which  supervisord image is build to the separate repository [redhat-developer/odo-supervisord-image](https://github.com/redhat-developer/odo-supervisord-image) and also setup Quay build for it https://quay.io/repository/openshiftdo/supervisord

@cdrage feel free to delete my commit and repo if you wanted to do it in a different way.

scripts/supervisord/assemble-and-restart
geoand: Indeed, those are the changes I had made! Thanks for following up @kadel

pkg/occlient/occlient.go
```diff
@@ -59,9 +59,13 @@ const (
 
 	// The length of the string to be generated for names of resources
 	nameLength = 5
-	// git repository that will be used for bootstraping
-	bootstrapperURI = "https://github.com/kadel/bootstrap-supervisored-s2i"
-	bootstrapperRef = "v0.0.2"
+
+	// Image that will be used containing the supervisord binary and assembly scripts
+	// TODO: change image tag to released version before merging
+	bootstrapperImage = "quay.io/openshiftdo/supervisord:latest"
```
kadel: Do image release before merging this, and then change "latest" to the tagged version.

pkg/occlient/occlient.go
```diff
@@ -59,9 +59,13 @@ const (
 
 	// The length of the string to be generated for names of resources
 	nameLength = 5
-	// git repository that will be used for bootstraping
-	bootstrapperURI = "https://github.com/kadel/bootstrap-supervisored-s2i"
-	bootstrapperRef = "v0.0.2"
+
+	// Image that will be used containing the supervisord binary and assembly scripts
+	// TODO: change image tag to released version before merging
```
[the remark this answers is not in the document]
mik-dass: @cdrage do we need to remove this currently?

pkg/occlient/occlient_test.go
```diff
@@ -2955,3 +2945,175 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Run function UpdateDCToGit
+			err := fakeClient.UpdateDCToGit(metav1.ObjectMeta{Name: tt.args.name}, tt.args.newImage)
+
+			// Error checking UpdateDCToGit
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToGit() unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			// Check to see how many actions are being ran
+			if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2955,3 +2945,285 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			err := fakeClient.UpdateDCToGit(metav1.ObjectMeta{Name: tt.args.name}, tt.args.newImage)
+
+			// Error checking UpdateDCToGit
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToGit() unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if err == nil && !tt.wantErr {
+				// Check to see how many actions are being ran
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2955,3 +2945,285 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			err := fakeClient.UpdateDCToSupervisor(metav1.ObjectMeta{Name: tt.args.name}, tt.args.imageName)
+
+			// Error checking UpdateDCToSupervisor
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToSupervisor() unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			// Check to see how many actions are being ran
+			if err == nil && !tt.wantErr {
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,284 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToGit
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToGit() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("error is here Client.GetClusterServicePlans() = %#v, want %#v", got, tt.want)
+
+			if err == nil && !tt.wantErr {
+				// Check to see how many actions are being ran
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,284 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToSupervisor
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToSupervisor() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Client.GetClusterServiceClassExternalNames() = %v, want %v", got, tt.want)
+
+			// Check to see how many actions are being ran
+			if err == nil && !tt.wantErr {
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,283 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToGit
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToGit() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("error is here Client.GetClusterServicePlans() = %#v, want %#v", got, tt.want)
+
+			if err == nil && !tt.wantErr {
+				// Check to see how many actions are being ran
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,283 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToSupervisor
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToSupervisor() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Client.GetClusterServiceClassExternalNames() = %v, want %v", got, tt.want)
+
+			// Check to see how many actions are being ran
+			if err == nil && !tt.wantErr {
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

geoand:
I also just tested with nodejs and java s2i and both worked :tada:

pkg/occlient/occlient.go
```diff
@@ -2153,3 +2156,39 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
 	}
 	return containerPorts, nil
 }
+
+// CreateBuildConfig creates a buildConfig using the builderImage as well as gitURL.
+func (c *Client) CreateBuildConfig(commonObjectMeta metav1.ObjectMeta, builderImage string, gitURL string) (buildv1.BuildConfig, error) {
```
mik-dass: This needs unit test

pkg/occlient/occlient.go
```diff
@@ -2153,3 +2156,39 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
...
+	_, err = c.buildClient.BuildConfigs(c.namespace).Create(&bc)
+	if err != nil {
+		return buildv1.BuildConfig{}, errors.Wrapf(err, "unable to create BuildConfig for %s", commonObjectMeta.Name)
+	}
+
+	return bc, nil
+}
+
+// findContainer finds the container
+func findContainer(containers []corev1.Container, name string) (corev1.Container, error) {
```
mik-dass: This needs unit test too

pkg/occlient/occlient.go
```diff
@@ -1565,6 +1533,18 @@ func (c *Client) DeletePVC(name string) error {
 	return c.kubeClient.CoreV1().PersistentVolumeClaims(c.namespace).Delete(name, nil)
 }
 
+// DeleteBuildConfig deletes the given BuildConfig by name using CommonObjectMeta..
+func (c *Client) DeleteBuildConfig(commonObjectMeta metav1.ObjectMeta) error {
```
mik-dass: Unit test for this

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,283 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			} else if err != nil && !tt.wantErr {
+				t.Errorf("test failed, expected: no error, got error: %s", err.Error())
 			}
+
 		})
 	}
 }
+
+/*
+Functions to test:
```
mik-dass: Here is a TODO

syamgk:
since the image is being consumed from quay.io(quay.io/openshiftdo/supervisord:0.1.0) we can remove the TODO from commit message as well as PR description

pkg/occlient/occlient.go
kadel: This function is just calling Delete nothing more, I don't think that we need a unit test for this

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,278 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToGit
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToGit() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("error is here Client.GetClusterServicePlans() = %#v, want %#v", got, tt.want)
+
+			if err == nil && !tt.wantErr {
+				// Check to see how many actions are being ran
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient_test.go
```diff
@@ -2953,264 +2946,278 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+			// Error checking UpdateDCToSupervisor
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.UpdateDCToSupervisor() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Client.GetClusterServiceClassExternalNames() = %v, want %v", got, tt.want)
+
+			// Check to see how many actions are being ran
+			if err == nil && !tt.wantErr {
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

pkg/occlient/occlient.go
mik-dass: Just to be sure labels returned from `util.ConvertLabelsSelector()` are not modified before they are passed to the Delete call.

kadel:
Travis is reporting that tests failed, but they actually passed, there is just something wrong with macOS brew package

kadel:
ERR_TIMEOUT - merging ;-)

PR #623, opened 2018-08-07
Retrieve labels via DeploymentConfig not BuildConfig

cdrage:
This:
 - Changes label retrieval for GetComponentSource from usin BuildConfig
 to DeploymentConfig (not all deployments use BuildConfig)
 - Renames GetBuildConfig to GetBuildConfigFromName
 - Adds project param to GetDeploymentConfigFromName function in order
 to be similar / the same as GetBuildConfigFromName


cdrage:
Travis keeps on flacking, but tests pass locally (unit and integration)

PR #626, opened 2018-08-09
Adds the env flag to add the env variables to the component. 

### LINKED ISSUE #501: How to pass env var to the buildConfig to be created when executing the command "odo create" ?
The BuildConfig to be generated should be 

```
    source:
      git:
        uri: 'https://github.com/snowdrop/ocp-odo-build-install.git'
      type: Git
    strategy:
      sourceStrategy:
        from:
          kind: ImageStreamTag
          name: 'openjdk18:latest'
        incremental: true
        env:
        - name: ARTIFACT_COPY_ARGS
          value: "*-exec.jar"
```
but `odo create` generates this one

```
  source:
    git:
      uri: 'https://github.com/snowdrop/ocp-odo-build-install.git'
    type: Git
  strategy:
    sourceStrategy:
      from:
        kind: ImageStreamTag
        name: 'openjdk18:latest'
        namespace: openshift
    type: Source
```

So, How can we pass env var to the buildConfig to be created when executing the command "odo create" ?

mik-dass:
fixes #501 

Incase of git component the env vars are added to both the buildConfig and the deploymentConfig but incase of local/binary, only the deploymentConfig is updated.

The update functionality was changed to add the env vars to the buildConfig incase of git otherwise remove it. A new fucntion GetEnvVarsFromDC() was introduced to retrieve the env vars from the DC.

Function getinputEnvsFromStrings is added to retrive the env vars from the input strings. Also function addEnv is added to add the corev1.EnvVars to the deploymentConfig.


pkg/occlient/occlient.go
```diff
@@ -1750,10 +1786,10 @@ func (c *Client) GetOneServiceFromSelector(selector string) (*corev1.Service, er
...
 }
 
-// AddEnvironmentVariablesToDeploymentConfig adds the given environment
+// UpdateEnvironmentVariablesToDeploymentConfig adds the given environment
 // variables to the only container in the Deployment Config and updates in the
 // cluster
-func (c *Client) AddEnvironmentVariablesToDeploymentConfig(envs []corev1.EnvVar, dc *appsv1.DeploymentConfig) error {
+func (c *Client) UpdateEnvironmentVariablesToDeploymentConfig(envs []corev1.EnvVar, dc *appsv1.DeploymentConfig) error {
 	numContainers := len(dc.Spec.Template.Spec.Containers)
 	if numContainers != 1 {
```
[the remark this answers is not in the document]
mik-dass: yes

pkg/occlient/occlient.go
```diff
@@ -782,12 +807,22 @@ func addBootstrapVolumeMount(dc *appsv1.DeploymentConfig, dcName string) {
 	}
 }
 
+// addEnvVar adds the environmental variables to the containers in the DC
+// dc is the deployment config to be updated
+// inputEnvVars is the array containing the corev1.EnvVar values
+func addEnvVar(dc *appsv1.DeploymentConfig, inputEnvVars []corev1.EnvVar) {
+	for i := range dc.Spec.Template.Spec.Containers {
+		dc.Spec.Template.Spec.Containers[i].Env = inputEnvVars
```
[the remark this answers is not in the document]
kadel: > Also I see from https://github.com/redhat-developer/odo/pull/626/files#diff-6025e9b620e9321c03c11bf29e5c4dbdL517, that the dc created is always with only 1 container why so?

Simply because ODO currently doesn't support pods with multiple containers. There was no need for multi-container pods in use cases that ODO is trying to solve so far. So there is no way for users to create those.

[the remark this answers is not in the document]
kadel: Most of our code simply expect that there is only one container in the DC. But here you are setting env variables for all containers. I think that we should do the same thing as in other places - error out if there is more than one container.  It will make sure that we don't have any unexpected behaviors if we start implementing multi-container solutions in the future (it is better to error out than do something not expected)

pkg/occlient/occlient.go
```diff
@@ -1928,3 +1964,34 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
...
+// GetEnvVarsFromDC retrieves the env vars from the DC
+// dcName is the name of the dc from which the env vars are retrieved
+// projectName is the name of the project
+func (c *Client) GetEnvVarsFromDC(dcName string, projectName string) ([]corev1.EnvVar, error) {
+	dc, err := c.GetDeploymentConfigFromName(dcName, projectName)
+	if err != nil {
+		return nil, errors.Wrap(err, "error occured while retrieving the dc")
+	}
+	if len(dc.Spec.Template.Spec.Containers) > 0 {
+		return dc.Spec.Template.Spec.Containers[0].Env, nil
```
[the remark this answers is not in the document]
kadel: We should error out if there is more than one container. See my other comment https://github.com/redhat-developer/odo/pull/626/files#r211184292

pkg/component/component.go
```diff
@@ -526,7 +533,7 @@ func Link(client *occlient.Client, sourceComponent, targetComponent, application
 		return nil, errors.Wrapf(err, "unable to get Deployment Config for component %v", sourceComponent)
 	}
 
-	if err := client.AddEnvironmentVariablesToDeploymentConfig(
+	if err := client.UpdateEnvironmentVariablesToDeploymentConfig(
 		[]corev1.EnvVar{
```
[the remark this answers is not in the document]
mik-dass: Link internally appends the env vars and then calls update on the DC so I don't think it should be a problem

tests/e2e/e2e_test.go
```diff
@@ -255,6 +271,24 @@ var _ = Describe("odoe2e", func() {
 				protocols := runCmd("oc get services nodejs-git-" + appTestName + " -o go-template='{{range .spec.ports}}{{.protocol}}{{end}}'")
 				Expect(protocols).To(ContainSubstring("TCP"))
 				Expect(protocols).To(ContainSubstring("UDP"))
+
+				// deleting the component
+				runCmd("odo delete -f")
```
ashetty1: Need a check as to whether it is deleting nodejs-git.

pkg/occlient/occlient.go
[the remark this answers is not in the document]
syamgk: @anmolbabu there are many scenarios in which odo is exclusive to oc.
and it is not recommended to create objects using oc and then use odo for managing same
@kadel  will be able to add more insights on this

tests/e2e/e2e_test.go
```diff
@@ -91,6 +91,22 @@ func SourceTest(appTestName string, sourceType string, source string) {
 	Expect(getDc).To(ContainSubstring(source))
 }
 
+// EnvVarTest checks the component container env vars in the build config for git and deployment config for git/binary/local
+// appTestName is the app of the app
+// sourceType is the type of the source of the component i.e git/binary/local
+func EnvVarTest(appTestName string, sourceType string) {
```
ashetty1: Can we make this function generic to accept dc name, key/value as an arguments?

pkg/occlient/occlient_test.go
```diff
@@ -1725,14 +1834,73 @@ func TestNewAppS2I(t *testing.T) {
...
+				},
+				envVars: []string{"key=value", "key1"},
+			},
+			wantedService: map[int32]corev1.Protocol{
+				8080: corev1.ProtocolTCP,
+			},
+			wantErr: true,
+		},
+		{
+			name: "case 5: with valid gitUrl and two env vars (one with two =)",
```
ashetty1: `multiple values for a single key`?

cmd/create.go
```diff
@@ -16,10 +16,11 @@ import (
...
-	componentLocal  string
-	componentPorts  []string
+	componentBinary  string
+	componentGit     string
+	componentLocal   string
+	componentPorts   []string
+	componentEnvVars []string
 )
 
 var componentCreateCmd = &cobra.Command{
```
syamgk: it will be nice to add env example

tests/e2e/e2e_test.go
mik-dass: But we are checking build config too in the function

pkg/occlient/occlient.go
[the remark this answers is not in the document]
kadel: Not everything that is possible to do in `oc` has to be possible to do in `odo`.
For now, we are considering that there is always just one container. As `odo` doesn't have commands that could handle multi-container pods (for now).

pkg/component/component.go
```diff
@@ -414,7 +421,7 @@ func Update(client *occlient.Client, componentName string, applicationName strin
 		annotations := map[string]string{componentSourceURLAnnotation: sourceURL.String()}
 		annotations[componentSourceTypeAnnotation] = newSourceType
 
-		err = client.UpdateBuildConfig(namespacedOpenShiftObject, projectName, "", annotations)
+		err = client.UpdateBuildConfig(namespacedOpenShiftObject, projectName, "", annotations, nil)
```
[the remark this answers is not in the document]
kadel: In case of local or binary component, it doesn't make sense to set env variables for BuildConfig. For those components BuildConfig just injects SupervisorD into new image, and it is not used to build application. This is why it doesn't make sense to add env variables to BC in case of local and binary components.

Also as @mik-dass mentioned, we won't be using BC for SupervisorD injection after Charlie's work is done. BuildConfig will be created only in for git components, and that BC will be regular BC that builds application.

mik-dass:
@ashetty1 @syamgk Done

cmd/create.go
surajnarwade: added here https://github.com/redhat-developer/odo/pull/626/files#diff-452662b1e58d63509f34fc4bd164905bR52

jorgemoralespou:
could we discuss this PR?
This PR should be bound to the [config use case](https://docs.google.com/document/d/1x1Varzv6t067fRxAOmkfCTQtUcHmeBgHg-XkJELESOQ/edit?ts=5b7c5695) and what will be the possible options to inject configuration to a container. 

There's a [set of examples](https://docs.google.com/document/d/1mgkQctKa9hMtpIKM2QHx1gIHrcKPn6GY1CdP2BW0hM0/edit?usp=sharing), related to this use cases, that provide an initial definition of the experience that needs to be refined.

All these are not yet reflected as GH issues but should be.

jorgemoralespou:
@mik-dass Sounds good as current implementation in this PR aligns with what is being proposed. It might require to change once the full proposal for configuration management for components is defined, but at least, will provide a way to configure environment variables at component creation time for the time being.

mik-dass:
@jorgemoralespou The current implementation uses the --env flag. Should I also add a shorthand like -e ?

jorgemoralespou:
@mik-dass I'll open an issue to discuss short vs long flags. For now, leave this PR as it stands.

pkg/occlient/occlient_test.go
```diff
@@ -2726,3 +2905,546 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+				}
+				return true, &tt.returnedDC, nil
+			})
+
+			envVars, err := fakeClient.GetEnvVarsFromDC(tt.dcName, tt.projectName)
+
+			if err == nil && !tt.wantErr {
+				// Check for validating actions performed
+				if (len(fakeClientSet.AppsClientset.Actions()) != 1) && (tt.wantErr != true) {
+					t.Errorf("expected 2 action in GetBuildConfigFromName got: %v", fakeClientSet.AppsClientset.Actions())
```
syamgk: this string should be `expected 1 action in GetEnvVarsFromDC got : %v`

pkg/occlient/occlient_test.go
```diff
@@ -2726,3 +2905,546 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+				if len(tt.args.inputPorts) <= 0 {
+					if len(fkclientset.ImageClientset.Actions()) != 3 {
+						t.Errorf("expected 3 ImageClientset.Actions() in NewAppS2I, got: %v", fkclientset.ImageClientset.Actions())
+					}
+
+					// Check for imagestream objects
+					createdIS = fkclientset.ImageClientset.Actions()[2].(ktesting.CreateAction).GetObject().(*imagev1.ImageStream)
+				} else {
+					if len(fkclientset.ImageClientset.Actions()) != 1 {
+						t.Errorf("expected 3 ImageClientset.Actions() in NewAppS2I, got: %v", fkclientset.ImageClientset.Actions())
```
syamgk: here also mismatch in number of actions

mik-dass:
@syamgk Done and thanks for the suggestions :+1:

pkg/occlient/occlient.go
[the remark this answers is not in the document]
anmolbabu: >@anmolbabu there are many scenarios in which odo is exclusive to oc.
and it is not recommended to create objects using oc and then use odo for managing same
@kadel will be able to add more insights on this

but in that case, whenever we announce public availability, how do we plan to support already existing setups... What would be the migration path from oc to odo and also during the process, would it be nice to suggest users to completely stop using oc client if they want to use odo without issues? How would they get confidence to do so?


>Not everything that is possible to do in oc has to be possible to do in odo.
For now, we are considering that there is always just one container. As odo doesn't have commands that could handle multi-container pods (for now).

Ok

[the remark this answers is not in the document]
mik-dass: @anmolbabu @kadel Let's open a issue for the above mentioned concerns.

cmd/create.go
```diff
@@ -186,6 +190,7 @@ func init() {
 	componentCreateCmd.Flags().StringVar(&componentGit, "git", "", "Use a git repository as the source file for the component")
 	componentCreateCmd.Flags().StringVar(&componentLocal, "local", "", "Use local directory as a source file for the component")
 	componentCreateCmd.Flags().StringSliceVar(&componentPorts, "port", []string{}, "Ports to be used when the component is created")
+	componentCreateCmd.Flags().StringSliceVar(&componentEnvVars, "env", []string{}, "Environmental variables for the component")
```
kadel: Help message should explain how env variables are specified (what is the format)

pkg/occlient/occlient.go
```diff
@@ -1951,10 +1997,10 @@ func (c *Client) GetOneServiceFromSelector(selector string) (*corev1.Service, er
 	return &services[0], nil
 }
 
-// AddEnvironmentVariablesToDeploymentConfig adds the given environment
+// UpdateEnvironmentVariablesToDeploymentConfig adds the given environment
 // variables to the only container in the Deployment Config and updates in the
 // cluster
-func (c *Client) AddEnvironmentVariablesToDeploymentConfig(envs []corev1.EnvVar, dc *appsv1.DeploymentConfig) error {
+func (c *Client) UpdateEnvironmentVariablesToDeploymentConfig(envs []corev1.EnvVar, dc *appsv1.DeploymentConfig) error {
```
kadel: Why was this function renamed?
Using 'Upade' makes it look like env variables in DC are replaced instead of added to the existing one.

cmd/create.go
mik-dass: `Environmental variables for the component, specified like --env Key=Value` Is it good enough?

kadel: It might be better to use something like `VariableName` or something like that indicates that it is name of the variable rather than `Key`

pkg/occlient/occlient_test.go
```diff
@@ -2955,3 +3134,550 @@ func TestGetServiceInstanceList(t *testing.T) {
...
+					return true, nil, fmt.Errorf("get dc called with different name, expected: %s, got %s", tt.dcName, dcName)
+				}
+				return true, &tt.returnedDC, nil
+			})
+
+			envVars, err := fakeClient.GetEnvVarsFromDC(tt.dcName, tt.projectName)
+
+			if err == nil && !tt.wantErr {
+				// Check for validating actions performed
+				if (len(fakeClientSet.AppsClientset.Actions()) != 1) && (tt.wantErr != true) {
```

mik-dass:
@kadel Done

kadel:
```
./pkg/occlient/occlient_test.go
```

mik-dass:
@kadel Rebased and fixed

cdrage:
When doing multiple values with the same key name, it appears duplicated in the resulting deployment config:

Example:

```sh
                                                                                                                                                                              5h57m  ⍉│                                                                                                                                                                                      
▶ ./odo create nodejs f --env foo=bar,foo=bar                                                                                                                                         │                                                                                                                                                                                      
Please wait, creating f component ...            
```

DC output:
```
  template:                                             
    metadata:        
      creationTimestamp: null                       
      labels:                               
        deploymentconfig: f-app                                                                                                                                                                                                                                                                                                                                              
    spec:                                                            
      containers:                            
      - env:               
        - name: foo                   
          value: bar     
        - name: foo                    
          value: bar    
        image: 172.30.1.1:5000/myproject/f-app@sha256:f77c4a931952d8c08c6c054d1cc2eedafdbee017df0adcf9ebf5cf7140d3686c
```

We should check that the user isn't entering multiple keys with the same name as well as making sure that tests cover this scenario too.

@mik-dass

pkg/component/component.go
```diff
@@ -99,7 +100,8 @@ func GetComponentPorts(client *occlient.Client, componentName string, applicatio
 
 // CreateFromPath create new component with source or binary from the given local path
 // sourceType indicates the source type of the component and can be either local or binary
-func CreateFromPath(client *occlient.Client, name string, componentImageType string, path string, applicationName string, sourceType string, inputPorts []string) error {
+// envs is the array containing the string env values
+func CreateFromPath(client *occlient.Client, name string, componentImageType string, path string, applicationName string, sourceType string, inputPorts []string, envVars []string) error {
```
cdrage: These are way too many functions going into CreateFromPath (max 6 for Go is acceptable...)

I've refactored this in #622 but either way, one of us will have to rebase. Just doing a note for this in the future.

mik-dass: Ok if your PR gets in before this, then I will rebase

mik-dass:
@cdrage Done. I have added validation and tests for multiple keys with the same name. Please have a look again.

pkg/occlient/occlient.go
```diff
@@ -2177,11 +2177,19 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
...
 	var inputEnvVars []corev1.EnvVar
+	var keys = make(map[string]int)
 	for _, env := range envVars {
 		splits := strings.Split(env, "=")
 		if len(splits) < 2 || len(splits) > 2 {
 			return nil, errors.New("invalid syntax for env, please specify a key=value pair")
 		}
+		_, ok := keys[splits[0]]
+		if ok {
+			return nil, errors.Errorf("multiple values found for variable: %s", splits[0])
```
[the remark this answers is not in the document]
mik-dass: > multiple keys found for variable

I think it's confusing because of the other word "multiple" used before keys. It seems better before "values" WDYT?

> Multiple values found for key: %s", key

The other error messages start with small letters, so I think it's better to keep this small too. WDYT?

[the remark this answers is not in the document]
kadel: lets not use word key here.  It is variable name and value. Look at it from the users point of view.

tests/e2e/e2e_test.go
```diff
@@ -476,3 +513,325 @@ var _ = Describe("odoe2e", func() {
...
+				appName := runCmd("odo app create " + appTestName)
+				Expect(appName).To(ContainSubstring(appTestName))
+			})
+		})
+	})
+
+	Context("updating the component", func() {
+		It("should be able to create binary component", func() {
+			runCmd("curl -o " + tmpDir + "/sample-binary-testing-1.war " +
+				"https://gist.github.com/mik-dass/f95bd818ddba508ff76a386f8d984909/raw/e5bc575ac8b14ba2b23d66b5cb4873657e1a1489/sample.war")
```
anmolbabu: I was thinking if there can be any alternative to using personal gists?

mik-dass: I tried to look for online war hosted files but didn't find many reliable ones. If you find any online hosted ones, let me know, I will change them to those

pkg/occlient/occlient.go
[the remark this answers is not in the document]
mik-dass: Done

tests/e2e/e2e_test.go
```diff
@@ -21,6 +21,22 @@ import (
...
+
+	if sourceType == "git" {
+		// checking the values of the env vars pairs in bc
+		envVars := runCmd("oc get bc " + resourceName + " -o go-template='{{range .spec.strategy.sourceStrategy.env}}{{.name}}{{.value}}{{end}}'")
+		Expect(envVars).To(Equal("keyvaluekey1value1"))
+	}
+
+	// checking the values of the env vars pairs in dc
+	envVars := runCmd("oc get dc " + resourceName + " -o go-template='{{range .spec.template.spec.containers}}{{range .env}}{{.name}}{{.value}}{{end}}{{end}}'")
+	Expect(envVars).To(Equal("keyvaluekey1value1"))
```
cdrage: This shouldn't be hard-coded right? Shouldn't we pass these "test" variables with parameters?

pkg/occlient/occlient.go
```diff
@@ -719,6 +738,19 @@ func (c *Client) CreateService(commonObjectMeta metav1.ObjectMeta, containerPort
 	return nil
 }
 
+// updateEnvVar updates the environmental variables to the container in the DC
+// dc is the deployment config to be updated
+// inputEnvVars is the array containing the corev1.EnvVar values
+func updateEnvVar(dc *appsv1.DeploymentConfig, inputEnvVars []corev1.EnvVar) error {
```
cdrage: I don't think we're testing this function?

mik-dass: @cdrage Unit test coming up :D

mik-dass:
@cdrage Done. Have added the missing test for updateEnvVar(). Please have a look again.

tests/e2e/e2e_test.go
anmolbabu: Hmm... You can search for the other one ;)
The only reason why we should not depend on external links is, if they ever remove it or if the gists get auto removed due to the expiry of the assigned lifespan for the gist, our tests will start raising alarms :(
Optionally(the best way in my opinion) we could also keep them in our repo itself in a separate directory(may be samples directory for  example) and use it by way of relative path or a git url to the file(the way you want for the test)...
Not sure how high on priority this task would be... but considering the fact that maintaining them in a dir on github is a not so difficult task, we may consider giving it a try... But I would take @kadel or @cdrage opinion before doing this...

anmolbabu: Hmm... You can search for the other one ;)
The only reason why we should not depend on external links is, if they ever remove it or if the gists get auto removed due to the expiry of the assigned lifespan for the gist, our tests will start raising alarms :(
Optionally(the best way in my opinion) we could also keep them in our repo itself in a separate directory(may be samples directory for  example) and use it by way of relative path or a git url to the file(the way you want for the test)...
Not sure how high on priority this task would be... but considering the fact that maintaining them in a dir on github is a not so difficult task, we may consider giving it a try... But I would take @kadel or @cdrage opinion before doing this...

anmolbabu: If you choose to maintain them as part of our repo, consider the discussion on https://github.com/redhat-developer/odo/issues/777

kadel:
@cdrage please check if your comments were addressed, than you can merge it ;-)

mik-dass:
@kadel @cdrage @anmolbabu Rebased with the master.

cmd/create.go
```diff
@@ -194,6 +198,7 @@ func init() {
 	componentCreateCmd.Flags().StringVarP(&componentGit, "git", "g", "", "Use a git repository as the source file for the component")
 	componentCreateCmd.Flags().StringVarP(&componentLocal, "local", "l", "", "Use local directory as a source file for the component")
 	componentCreateCmd.Flags().StringSliceVar(&componentPorts, "port", []string{}, "Ports to be used when the component is created")
+	componentCreateCmd.Flags().StringSliceVar(&componentEnvVars, "env", []string{}, "Environmental variables for the component, specified like --env VariableName=Value")
```
cdrage: small nitpick. don't use "like". Use `For example`.

So in this case:

`Environment variables for the component. For example --env Key=Value`

We also shouldn't use "VariableName", we should keep consistent with using the whole `Key=Value` examples.

pkg/component/component.go
```diff
@@ -48,7 +48,8 @@ func validateSourceType(sourceType string) bool {
 }
 
 // CreateFromGit inputPorts is the array containing the string port values
-func CreateFromGit(client *occlient.Client, name string, componentImageType string, url string, applicationName string, inputPorts []string) error {
+// inputPorts is the array containing the string port values
```
cdrage: you add a comment regarding inputPorts but not about envVars

pkg/component/component.go
```diff
@@ -107,7 +108,7 @@ func GetComponentPorts(client *occlient.Client, componentName string, applicatio
 
 // CreateFromPath create new component with source or binary from the given local path
 // sourceType indicates the source type of the component and can be either local or binary
```
cdrage: add comment regarding envVars

pkg/occlient/occlient.go
cdrage: maybe change the variable name to envVars similar to all the other functions?

pkg/occlient/occlient.go
```diff
@@ -2158,7 +2190,7 @@ func getContainerPortsFromStrings(ports []string) ([]corev1.ContainerPort, error
 }
 
 // CreateBuildConfig creates a buildConfig using the builderImage as well as gitURL.
-func (c *Client) CreateBuildConfig(commonObjectMeta metav1.ObjectMeta, builderImage string, gitURL string) (buildv1.BuildConfig, error) {
+func (c *Client) CreateBuildConfig(commonObjectMeta metav1.ObjectMeta, builderImage string, gitURL string, inputEnvVars []corev1.EnvVar) (buildv1.BuildConfig, error) {
```
cdrage: envVars not inputEnvVars, keep the variable names consistent across functions so dev's aren't confused

cmd/create.go
mik-dass: But @kadel suggested me to use VariableName instead of key https://github.com/redhat-developer/odo/pull/626#discussion_r220176905

cdrage: Ah you're right. Let's do that.

cdrage:
LGTM! Let's wait until tests are green and then we can merge.

mik-dass:
@cdrage Strangely the deploy test didn't run, so I restarted manually. Now, all green :D

PR #633, opened 2018-08-14
Update 'odo component list' to 'odo catalog list' in cli reference doc

### LINKED ISSUE #632: CLI reference document suggests to use non working command 'odo component list' to get available components
[cli-reference.md#L211](https://github.com/redhat-developer/odo/blame/master/docs/cli-reference.md#L211) says 
``` 
A full list of component types that can be deployed is available using: 'odo component list'
```
when executed it prints out
``` shell 
$odo component list
Component list does not exist in the current application
```

Should it be `odo catalog list` instead?

dgolovin:
Fix #632

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -208,7 +208,7 @@ Create a new component to deploy on OpenShift.
 
 If component name is not provided, component type value will be used for the name.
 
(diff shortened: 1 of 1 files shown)


kadel:
Thank you Denis.

PR #642, opened 2018-08-16
rename '--ports' flag for 'odo create' to singular '--port'

kadel:
To make it align with `odo url create` it is better to have flag as singular, even tho multiple ports can be specified.
When `odo create --port 8080 --port 9090` form is used it also makes more sense to have it singular.


mik-dass:
@kadel The e2e tests have to be modified too, travis is failing because of that

PR #650, opened 2018-08-19
Corrected spelling mistakes and some grammatical errors.

acgoliyan:
Made the verbal part grammatically correct and corrected some misspelled words.

### DIFF
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -12,8 +12,8 @@ import (
 	"github.com/redhat-developer/odo/pkg/project"
 )
 
-// getDefaultAppName returns application name to be used as a default name in the case where users doesn't provide a name
-// In future this function should generate name with uniq suffix (app-xy1h), because there might be multiple applications.
+// getDefaultAppName returns application name to be used as a default name in case of user doesn't provide a name.
+// In future this function should generate name with unique suffix (app-xy1h), because there might be multiple applications.
 func getDefaultAppName() string {
 	return "app"
 }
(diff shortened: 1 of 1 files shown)


kadel:
Thank you, @acgoliyan.

pkg/application/application.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/redhat-developer/odo/pkg/project"
 )
 
-// getDefaultAppName returns application name to be used as a default name in the case where users doesn't provide a name
-// In future this function should generate name with uniq suffix (app-xy1h), because there might be multiple applications.
+// getDefaultAppName returns application name to be used as a default name in case of user doesn't provide a name.
```
mik-dass: I guess it should be "... in case the user doesn't provide"

acgoliyan: Thankyou @mik-dass for reviewing. Yes you right, "USERS" should be replaced by "USER".
plus here i made it to "in case of" which means what should be done in case of something might happen(possibility "user doesn't provide file name"). 
here is the reference.
http://englishstudypage.com/grammar/in-case-of-vs-in-the-case-of-in-english/

mik-dass: @acgoliyan Thanks for contributing :+1:  But I was suggesting **in case the user** not **in the case of**. I think **in case** should be used here because http://speakspeak.com/resources/english-grammar-rules/various-grammar-rules/in-case-in-case-of

acgoliyan: @mik-dass Sorry i misunderstood. You are absolutely right "in case" is right to use here.

mik-dass: @acgoliyan No problem thanks again for the contribution :+1:

PR #654, opened 2018-08-20
how odo works documentation

kadel:
Document is at

https://docs.google.com/document/d/1IPrSUw2CVcXpc6RqIwIDC0q9TBheFcUFJ8anfHk3oDE/edit# 

Google docs makes it easier to collaborate on this kind of documents. Anyone can edit it.
Feel free to correct or add anything.

### DIFF
--- a/docs/how-odo-works.adoc
+++ b/docs/how-odo-works.adoc
@@ -1,0 +1,7 @@
+= What happens when you run ODO command
+
+Document is at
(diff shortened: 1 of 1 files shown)


kadel:
> @kadel , why are we using .adoc format ?

AsciiDoc is better for larger documentation as you can do stuff like footnotes indexes etc..

My idea is that we will generate html documentation fopr webpage from this and AsciiDoc will give us more flexibility

surajnarwade:
sounds good to me

Issue #668, opened 2018-08-27
Remove "oc" installation requirement in Getting Started


ALRubinger:
The `oc` tool comes bundled with MiniShift and those one provided there comes with guaranteed version compatibility.  Rev the instructions such that we make note of this, removing the `oc` installation requirement in the process.  Note the ability to install the `oc` binary as an optional step.

cdrage:
Your PR here: https://github.com/ALRubinger/odo/commit/9402be071b3cc96b8c5042e9584114ded1e028e5

Having those instructions for `oc` aren't needed, as `odo` no longer uses the binary and does everything through the OpenShift API.

ALRubinger:
@cdrage Cool; then how does it know which OpenShift instance you're logged into?  By finding the same metadata in the user's home?  And how would the developer log into that instance without oc?

jorgemoralespou:
@ALRubinger `odo login` will be similar to `oc login` and yes, it uses the same metadata. This is not yet merged, though. For the time being, you still need `oc` binary to log into the cluster or manually set the config file.

PR #669, opened 2018-08-27
Issue #668

ALRubinger:
PR for #668 as described in the issue.  Thanks, reviewers!

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -17,7 +17,6 @@ We'll be going over the following steps:
 A few requirements before we proceed!
 
  - A running OpenShift cluster (we recommend using [`minishift`](https://docs.openshift.org/latest/minishift/getting-started/installing.html))
- - `oc` binary ([installation guide here](https://docs.openshift.org/latest/minishift/getting-started/quickstart.html#starting-minishift))
  - `odo` binary ([installation guide here](/README.md#installation))
 
 The quickest way to deploy a development OpenShift cluster is by using [minishift](https://docs.openshift.org/latest/minishift/index.html). Alternatively, `odo` will automatically work with any OpenShift cluster you're currently logged into.
@@ -67,6 +66,16 @@ OpenShift server started.
 The server is accessible via web console at:
     https://192.168.42.10:8443
 ```
+
+Now set your `PATH` to include the version of the `oc` tool bundled with Minishift.  You may already have `oc` installed separately, but this approach releases you from this requirement and additionally ensures that the versions between client and server are aligned.
+
+```console
+$ eval $(minishift oc-env)
+$ which oc
+~/.minishift/cache/oc/v3.10.0/linux/oc
+```
+
+Note that you will have to execute this `eval` command for every command shell you're using as its effects are not global.  Alternatively, you may install the `oc` tool system-wide and put it on the `PATH`, but you will need to ensure version compatibility.  You may take that approach [here](https://docs.openshift.org/latest/minishift/getting-started/quickstart.html#starting-minishift).
 
 Now log into the OpenShift cluster:
 


surajnarwade:
@ALRubinger, user might use `openshift online` or any other `openshift` cluster rather than minishift, so I think, it is okay to have separate oc installation guide. But thing you wrote about setting path is also cool. it just that we should have both.

docs/getting-started.md
```diff
@@ -17,7 +17,6 @@ We'll be going over the following steps:
 A few requirements before we proceed!
 
  - A running OpenShift cluster (we recommend using [`minishift`](https://docs.openshift.org/latest/minishift/getting-started/installing.html))
- - `oc` binary ([installation guide here](https://docs.openshift.org/latest/minishift/getting-started/quickstart.html#starting-minishift))
```
surajnarwade: don't remove this part

docs/getting-started.md
```diff
@@ -68,6 +67,16 @@ The server is accessible via web console at:
     https://192.168.42.10:8443
 ```
 
+Now set your `PATH` to include the version of the `oc` tool bundled with Minishift.  You may already have `oc` installed separately, but this approach releases you from this requirement and additionally ensures that the versions between client and server are aligned.
```
surajnarwade: This LGTM :)

surajnarwade:
cc @cdrage @ALRubinger

docs/getting-started.md
cdrage: why not? We don't require the binary anymore (other than login/logout which is outside of Odo) to communicate with OpenShift.

docs/getting-started.md
```diff
@@ -68,6 +67,16 @@ The server is accessible via web console at:
...
+
+```console
+$ eval $(minishift oc-env)
+$ which oc
+~/.minishift/cache/oc/v3.10.0/linux/oc
```
cdrage: Why do we have to add this / add instructions with regards to minishift here? I thought that it was as simply as using `minishift start`. Adding these instructions makes it seem a lot more complex than it is... Minishift will automatically download the correct `oc`, so why should we do this manually?

docs/getting-started.md
```diff
@@ -68,6 +67,16 @@ The server is accessible via web console at:
...
+```
+
+Note that you will have to execute this `eval` command for every command shell you're using as its effects are not global.  Alternatively, you may install the `oc` tool system-wide and put it on the `PATH`, but you will need to ensure version compatibility.  You may take that approach [here](https://docs.openshift.org/latest/minishift/getting-started/quickstart.html#starting-minishift).
```
cdrage: Still don't know why we should add this.. shouldn't we keep this simple (using just `minishift start`?)

cdrage:
Instead of having documentation, should Odo instead have a built-in functionality that will check the server's version of OpenShift (3.9.0+) and warn the user?

To be honest, as of right now, we no longer require `oc` to talk to OpenShift. The last part is the login/logout functionality. I don't see the need of adding instructions on having the `oc` binary on the developers machine match the version of the cluster that's running.

I don't mind merging this (for the time being) until we have the login/logout functionality completed and we are safe to not rely on `oc` 100%.

docs/getting-started.md
kadel: without `oc` `odo` is useless as the user doesn't' have a way to log in to OpenShift cluster.

We still need `oc` .

surajnarwade:
any update @ALRubinger  @cdrage

docs/getting-started.md
ALRubinger: Am sorry that I am just seeing this notification now, @cdrage.  The reasoning here is that `minishift start` is not the only command a user may execute.  I may use another shell to issue `oc` commands and I need to have the `oc` being used as both available and the correct version.

surajnarwade:
@ALRubinger , feel free to reopen the PR if any further discussion :smile:

PR #671, opened 2018-08-27
Added vendor-update command to makefile

surajnarwade:
This PR adds vendor-update command to Make, so that
we can simply update vendor using `make vendor-update` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,3 +98,8 @@ packages:
 .PHONY: upload-packages
 upload-packages:
 	./scripts/upload-packages.sh
+
+# Update vendoring
+.PHONY: vendor-update
+vendor-update:
+	glide update --strip-vendor


Makefile
```diff
@@ -98,3 +98,8 @@ packages:
 .PHONY: upload-packages
 upload-packages:
 	./scripts/upload-packages.sh
+
+# Update vendoring
+.PHONY: vendor-update
+vendor-update:
+	glide update --strip-vendor
```
cdrage: shouldn't we add `glide update -v` before too?

surajnarwade: According to odo doc, this should be sufficient I guess.
I was thinking of adding `glide-vc --only-code --no-tests` as well 
WDYT ?

cdrage: I don't think we're adding that anymore. Using `glide-vc` and all.

Maybe we should echo after vendor-update saying that the user should `git add .` and do a separate commit? Similar to how `developers.md` describes it.

surajnarwade: using `glide-vc` removes extra clutter which solves https://github.com/redhat-developer/odo/issues/670 as well

cdrage: I believe there's a reason why we don't use glide-vc, @kadel would know once he gets back.

surajnarwade: Got it, so as of now, we can merge this in

cdrage: agreed. this LGTM.

PR #677, opened 2018-08-29
Add CLI-reference related to service catalog

surajnarwade:
Added respective command reference related to service catalog in `cli-reference.md`
under docs section

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,35 +130,45 @@ odo --verbose : Odo (Openshift Do)
         list : Lists all the applications
         set : Set application as active
     catalog : Catalog related operations
-        list : List all available component types.
-        search : Search component type in catalog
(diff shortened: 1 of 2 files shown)


surajnarwade:
Merge only when #413 gets merged

cdrage:
@surajnarwade You do know that we auto generate this from https://github.com/redhat-developer/odo/blob/master/scripts/cli-reference/generate-cli-reference.go right? We may have to close this PR.

surajnarwade:
I was not aware of that :(

surajnarwade:
I used `make generate-cli-structure` to modify `docs/cli-reference.md`  file now, so that new command reference will get added for  `odo service xxxx`

cdrage:
LGTM!

PR #680, opened 2018-08-30
Improve service catalog related stuff from PR 413

surajnarwade:
Since PR 413 is merged, there were few minor comments charlie mentioned later.
Those comments are addressed in this PR.

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -34,13 +34,13 @@ var catalogListCmd = &cobra.Command{
 
 var catalogListComponentCmd = &cobra.Command{
 	Use:   "components",
-	Short: "List all available component types.",
+	Short: "List all components available.",
 	Long:  "List all available component types from OpenShift's Image Builder.",
 	Example: `  # Get the supported components
   odo catalog list components
 
   # Search for a supported component
-  odo catalog search components nodejs
+  odo catalog search component nodejs
 `,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
@@ -77,13 +77,13 @@ var catalogListComponentCmd = &cobra.Command{
 
 var catalogListServiceCmd = &cobra.Command{
 	Use:   "services",
-	Short: "Lists all the services from service catalog",
-	Long:  "Lists all the services from service catalog",
+	Short: "Lists all available services",
+	Long:  "Lists all available services",
 	Example: `  # List all services
   odo catalog list services
 
- # Search for a supported services
-  odo catalog search services mysql
+ # Search for a supported service
+  odo catalog search service mysql
 	`,
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
@@ -111,7 +111,7 @@ This searches for a partial match for the given search term in all the available
 components & services.
 `,
 	Example: `  # Search for a component
-  odo catalog search components python
+  odo catalog search component python
 
   # Search for a service
   odo catalog search service mysql
@@ -119,7 +119,7 @@ components & services.
 }
 
 var catalogSearchComponentCmd = &cobra.Command{
-	Use:   "components",
+	Use:   "component",
 	Short: "Search component type in catalog",
 	Long: `Search component type in catalog.
 
@@ -128,7 +128,7 @@ components.
 `,
 	Args: cobra.ExactArgs(1),
 	Example: `  # Search for a component
-  odo catalog search components python
+  odo catalog search component python
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
@@ -149,7 +149,7 @@ components.
 }
 
 var catalogSearchServiceCmd = &cobra.Command{
-	Use:   "services",
+	Use:   "service",
 	Short: "Search service type in catalog",
 	Long: `Search service type in catalog.
 
@@ -157,7 +157,7 @@ This searches for a partial match for the given search term in all the available
 services from service catalog.
 `,
 	Example: `  # Search for a service
-  odo catalog search services mysql
+  odo catalog search service mysql
 	`,
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -34,7 +34,7 @@ var serviceCreateCmd = &cobra.Command{
 
 If service name is not provided, service type value will be used.
 
-A full list of service types that can be deployed is available using: 'odo service catalog'`,
+A full list of service types that can be deployed are available using: 'odo catalog list services'`,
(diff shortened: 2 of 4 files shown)


anmolbabu:
@surajnarwade I would be glad if you could also resolve https://github.com/redhat-developer/odo/issues/678 as part of this PR

cmd/catalog.go
```diff
@@ -34,7 +34,7 @@ var catalogListCmd = &cobra.Command{
 
 var catalogListComponentCmd = &cobra.Command{
 	Use:   "components",
-	Short: "List all available component types.",
+	Short: "List all components available.",
 	Long:  "List all available component types from OpenShift's Image Builder.",
 	Example: `  # Get the supported components
   odo catalog list components
```
cmoulliard: Is the word `components` appropriate here as the goal is to list the `builder` images available. I propose/suggest then that you use this command -> `List all builder's images available` and `odo catalog list builders`

cmd/catalog.go
```diff
@@ -34,7 +34,7 @@ var catalogListCmd = &cobra.Command{
 
 var catalogListComponentCmd = &cobra.Command{
 	Use:   "components",
-	Short: "List all available component types.",
+	Short: "List all components available.",
```
anmolbabu: I seem to get confused between component and component type
My current understanding is that a component is an application built using a builder imagestream of type same as component type.. which makes me arrive at 
component === app
component type === builder image stream of respective type..
It is possible that I am missing something
Please correct me if I am wrong...

cmd/service.go
```diff
@@ -34,7 +34,7 @@ var serviceCreateCmd = &cobra.Command{
 
 If service name is not provided, service type value will be used.
 
-A full list of service types that can be deployed is available using: 'odo service catalog'`,
+A full list of service types that can be deployed are available using: 'odo catalog list services'`,
 	Example: `  # Create new mysql-persistent service from service catalog.
   odo service create mysql-persistent
```
cmoulliard: To be consistent, the syntax of this command should be 

- creation : `odo catalog service create <name_of_the_service_class>`
- deletion : `odo catalog service delete mysql-persistent`

WDYT ?

cmd/catalog.go
```diff
@@ -64,12 +64,12 @@ var catalogListComponentCmd = &cobra.Command{
...
 	Use:   "services",
-	Short: "Lists all the services from service catalog",
-	Long:  "Lists all the services from service catalog",
+	Short: "Lists all services available",
+	Long:  "Lists all services available",
 	Example: `  # List all services
   odo catalog list services
 
- # Search for a supported services
+ # Search for a supported a service
```
anmolbabu: Search for a supported service :)

cmd/catalog.go
```diff
@@ -64,12 +64,12 @@ var catalogListComponentCmd = &cobra.Command{
 
 var catalogListServiceCmd = &cobra.Command{
 	Use:   "services",
-	Short: "Lists all the services from service catalog",
-	Long:  "Lists all the services from service catalog",
+	Short: "Lists all services available",
+	Long:  "Lists all services available",
```
anmolbabu: Can this be : List all available services ? @cdrage

cmd/catalog.go
```diff
@@ -64,12 +64,12 @@ var catalogListComponentCmd = &cobra.Command{
...
 	Example: `  # List all services
   odo catalog list services
 
- # Search for a supported services
+ # Search for a supported a service
   odo catalog search services mysql
```
cmoulliard: Should we use the plural or singular form ? If the odo's command can do a search to retrieve all the services containing the word passed as parameter, then go ahead to use the plural `services` word otherwise that should be the `singular` -> `odo catalog search service mysql`

cmd/service.go
```diff
@@ -34,7 +34,7 @@ var serviceCreateCmd = &cobra.Command{
 
 If service name is not provided, service type value will be used.
 
-A full list of service types that can be deployed is available using: 'odo service catalog'`,
+A full list of service types that can be deployed are available using: 'odo catalog list services'`,
```
cmoulliard: Do you also envision to list the `plans`, `brokers`, `instances` and `bindings` created ?

`odo catalog list services|plans|brokers|bindings`

cmd/catalog.go
surajnarwade: From user perspective, it is component, right ?

cdrage:
May be good to change the title / name as well to not include my name haha. Maybe "Improve documentation for catalog" or something along those lines.

cmd/service.go
surajnarwade: `odo catalog`  is supposed to only list and search images and services
while `odo create` or `odo delete` will deal with components and
`odo service xxx` will deal with services

cmd/service.go
cmoulliard: 1) Plan detail

You need a command to show the detail of a plan as we can do with `svcat`
We could then use by example : `odo catalog plan <serviceName>/<planName>`

2) Create service
I agree with you for the command to create the service excepted that you must also pass the class name

`odo service create service_name --class class_name --plan plan_name -p parameters`

Remark : As some parameters are json typed, then it will be very difficult to pass them using a one line command and -p, -p ... This is why, we have in our prototype, defined such parameters/info into a yaml config file

3) Mount the secret to component/

To do the binding, then you must also pass the name of the secret

odo link component_name --service service_name --secret secret_name

component_name corresponds to the DeploymentConfig's name
secret_name : secret containing the parameters

Remark : There is a step that you don't convert which is "generating the secret with the parameters of the plan selected. Is it something that you will do automatically during the creation of the service. IF this is what you plan to do, then you must also pass the name of the secret ->

`odo service create service_name --class class_name --plan plan_name --secret secret_name -p parameters`

cmd/service.go
```diff
@@ -76,13 +76,15 @@ var serviceDeleteCmd = &cobra.Command{
...
 	Long:  "Delete an existing service",
-	Example: `  # Delete service named 'mysql-persistent'
+	Example: `  # Delete the service named 'mysql-persistent'
   odo service delete mysql-persistent
 	`,
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		glog.V(4).Infof("service delete called")
+
+		glog.V(4).Infof("service delete called\n")
```
cdrage: similar to my previous comment, can we merge these two debug / info messages?

cmd/catalog.go
surajnarwade: * App is your application which consist of multiple components
* Components is one of the part of your app
* component type is type of component, for example, java or python

cdrage:
This LGTM. Need more reviews however before merging.

cdrage:
@surajnarwade needs another rebase and LGTM.

PR #683, opened 2018-08-30
Refactoring code and added flags `project`, `application` &  `component` 

surajnarwade:
This PR does:

* Adds flags `project`, `application` and `component` flag to respective commands

* added `getProjectName` and `getAppName` reusable functions which looks for `project` or `application` flag first, if not provided then it gives current value by default.

* Provide comments to some of the exported functions

* Refactors the way we provide `namespace` value for methods in `occlient.go`.
(No method will accept namespace as a argument, whenever needed we will use `client.Namespace` value)

* Updated unit tests as per new refactoring

-- edit by @kadel 
https://github.com/redhat-developer/odo/issues/656

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -72,12 +72,12 @@ If no app name is passed, a default app name will be auto-generated.
 		// validate application name
 		err := validateName(appName)
 		checkError(err, "")
-		fmt.Printf("Creating application: %v in project: %v \n", appName, projectName)
+		fmt.Printf("Creating application: %v in project: %v\n", appName, projectName)
 		err = application.Create(client, appName)
 		checkError(err, "")
 		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		fmt.Printf("Switched to application: %v in project\n", appName, projectName)
+		fmt.Printf("Switched to application: %v in project: %v\n", appName, projectName)
 	},
 }
 
@@ -102,7 +102,7 @@ var applicationGetCmd = &cobra.Command{
 			fmt.Printf("There's no active application.\nYou can create one by running 'odo application create <name>'.\n")
 			return
 		}
-		fmt.Printf("The current application is: %v\n", app)
+		fmt.Printf("The current application is: %v in project: %v\n", app, projectName)
 	},
 }
 
@@ -137,14 +137,14 @@ var applicationDeleteCmd = &cobra.Command{
 		if applicationForceDeleteFlag {
 			confirmDeletion = "y"
 		} else {
-			fmt.Printf("Are you sure you want to delete the application: %v? [y/N] ", appName)
+			fmt.Printf("Are you sure you want to delete the application: %v from project: %v? [y/N] ", appName, projectName)
 			fmt.Scanln(&confirmDeletion)
 		}
 
 		if strings.ToLower(confirmDeletion) == "y" {
 			err := application.Delete(client, appName)
 			checkError(err, "")
-			fmt.Printf("Deleted application: %s\n", appName)
+			fmt.Printf("Deleted application: %s from project: %v\n", appName, projectName)
 		} else {
 			fmt.Printf("Aborting deletion of application: %v\n", appName)
 		}
@@ -205,7 +205,7 @@ var applicationSetCmd = &cobra.Command{
 		client := getOcClient()
 		appName := args[0]
 
-		getAndSetNamespace(client)
+		projectName := getAndSetNamespace(client)
 		// error if application does not exist
 		exists, err := application.Exists(client, appName)
 		checkError(err, "unable to check if application exists")
@@ -216,7 +216,7 @@ var applicationSetCmd = &cobra.Command{
 
 		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		fmt.Printf("Switched to application: %v\n", args[0])
+		fmt.Printf("Switched to application: %v in project: %v\n", args[0], projectName)
 	},
 }
 
@@ -237,7 +237,7 @@ var applicationDescribeCmd = &cobra.Command{
 			appName, err = application.GetCurrent(projectName)
 			checkError(err, "")
 			if appName == "" {
-				fmt.Printf("There's no active application in projectName %v", projectName)
+				fmt.Printf("There's no active application in project: %v", projectName)
 				os.Exit(1)
 			}
 		} else {
@@ -246,7 +246,7 @@ var applicationDescribeCmd = &cobra.Command{
 			exists, err := application.Exists(client, appName)
 			checkError(err, "")
 			if !exists {
-				fmt.Printf("Application with the name %s does not exist\n", appName)
+				fmt.Printf("Application with the name %s does not exist in %s \n", appName, projectName)
 				os.Exit(1)
 			}
 		}
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -55,6 +55,7 @@ If no app name is passed, a default app name will be auto-generated.
 	Run: func(cmd *cobra.Command, args []string) {
 		var appName string
 		client := getOcClient()
+		projectName := getAndSetNamespace(client)
 		if len(args) == 1 {
 			// The only arg passed is the app name
 			appName = args[0]
@@ -71,12 +72,9 @@ If no app name is passed, a default app name will be auto-generated.
 		// validate application name
 		err := validateName(appName)
 		checkError(err, "")
-		fmt.Printf("Creating application: %v\n", appName)
-		err = application.Create(client, appName)
-
-		projectName := setNamespace(client)
 		fmt.Printf("Creating application: %v in project: %v \n", appName, projectName)
-
+		err = application.Create(client, appName)
+		checkError(err, "")
 		err = application.SetCurrent(client, appName)
 		checkError(err, "")
 		fmt.Printf("Switched to application: %v in project\n", appName, projectName)
@@ -93,7 +91,7 @@ var applicationGetCmd = &cobra.Command{
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		projectName := setNamespace(client)
+		projectName := getAndSetNamespace(client)
 		app, err := application.GetCurrent(projectName)
 		checkError(err, "")
 		if applicationShortFlag {
@@ -118,7 +116,7 @@ var applicationDeleteCmd = &cobra.Command{
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		projectName := setNamespace(client)
+		projectName := getAndSetNamespace(client)
 		var appName string
 		// If name of the app to be deleted is not passed, consider the current app for deletion
 		if len(args) == 0 {
@@ -129,10 +127,11 @@ var applicationDeleteCmd = &cobra.Command{
 			// If app name passed, consider it for deletion
 			appName = args[0]
 		}
+
 		var confirmDeletion string
 
 		// Print App Information which will be deleted
-		err := printDeleteAppInfo(client, appName, projectName)
+		err := printDeleteAppInfo(client, appName)
 		checkError(err, "")
 
 		if applicationForceDeleteFlag {
@@ -166,7 +165,7 @@ var applicationListCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 
-		projectName := setNamespace(client)
+		projectName := getAndSetNamespace(client)
 		apps, err := application.ListInProject(client)
 		checkError(err, "unable to get list of applications")
 		if len(apps) > 0 {
@@ -206,7 +205,7 @@ var applicationSetCmd = &cobra.Command{
 		client := getOcClient()
 		appName := args[0]
 
-		projectName := setNamespace(client)
+		getAndSetNamespace(client)
 		// error if application does not exist
 		exists, err := application.Exists(client, appName)
 		checkError(err, "unable to check if application exists")
@@ -217,7 +216,7 @@ var applicationSetCmd = &cobra.Command{
 
 		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		fmt.Printf("Switched to application: %v in project: %v \n", args[0], projectName)
+		fmt.Printf("Switched to application: %v\n", args[0])
 	},
 }
 
@@ -232,7 +231,7 @@ var applicationDescribeCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		var appName string
-		projectName := setNamespace(client)
+		projectName := getAndSetNamespace(client)
 		if len(args) == 0 {
 			var err error
 			appName, err = application.GetCurrent(projectName)
@@ -253,7 +252,7 @@ var applicationDescribeCmd = &cobra.Command{
 		}
 
 		// List of Component
-		componentList, err := component.List(client, appName, projectName)
+		componentList, err := component.List(client, appName)
 		checkError(err, "")
 		if len(componentList) == 0 {
 			fmt.Printf("Application %s has no components deployed.\n", appName)
@@ -262,7 +261,7 @@ var applicationDescribeCmd = &cobra.Command{
 		fmt.Printf("Application %s has:\n", appName)
 
 		for _, currentComponent := range componentList {
-			componentType, path, componentURL, appStore, err := component.GetComponentDesc(client, currentComponent.Name, appName, projectName)
+			componentType, path, componentURL, appStore, err := component.GetComponentDesc(client, currentComponent.Name, appName)
 			checkError(err, "")
 			printComponentInfo(currentComponent.Name, componentType, path, componentURL, appStore)
 		}
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -9,14 +9,12 @@ import (
 
 	"github.com/redhat-developer/odo/pkg/application"
 	"github.com/redhat-developer/odo/pkg/component"
-	"github.com/redhat-developer/odo/pkg/project"
 	"github.com/spf13/cobra"
 )
 
 var (
 	applicationShortFlag       bool
 	applicationForceDeleteFlag bool
-	projectName                string
 )
 
 // applicationCmd represents the app command
@@ -55,11 +53,11 @@ If no app name is passed, a default app name will be auto-generated.
 	`,
 	Args: cobra.RangeArgs(0, 1),
 	Run: func(cmd *cobra.Command, args []string) {
-		var name string
+		var appName string
 		client := getOcClient()
 		if len(args) == 1 {
 			// The only arg passed is the app name
-			name = args[0]
+			appName = args[0]
 		} else {
 			// Desired app name is not passed so, generate a new app name
 			// Fetch existing list of apps
@@ -67,18 +65,21 @@ If no app name is passed, a default app name will be auto-generated.
 			checkError(err, "")
 
 			// Generate a random name that's not already in use for the existing apps
-			name, err = application.GetDefaultAppName(apps)
+			appName, err = application.GetDefaultAppName(apps)
 			checkError(err, "")
 		}
 		// validate application name
-		err := validateName(name)
+		err := validateName(appName)
 		checkError(err, "")
-		fmt.Printf("Creating application: %v\n", name)
-		err = application.Create(client, name)
+		fmt.Printf("Creating application: %v\n", appName)
+		err = application.Create(client, appName)
+
+		projectName := setNamespace(client)
+		fmt.Printf("Creating application: %v in project: %v \n", appName, projectName)
+
+		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		err = application.SetCurrent(client, name)
-		checkError(err, "")
-		fmt.Printf("Switched to application: %v\n", name)
+		fmt.Printf("Switched to application: %v in project\n", appName, projectName)
 	},
 }
 
@@ -92,7 +93,8 @@ var applicationGetCmd = &cobra.Command{
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		app, err := application.GetCurrent(client)
+		projectName := setNamespace(client)
+		app, err := application.GetCurrent(projectName)
 		checkError(err, "")
 		if applicationShortFlag {
 			fmt.Print(app)
@@ -116,21 +118,21 @@ var applicationDeleteCmd = &cobra.Command{
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
+		projectName := setNamespace(client)
 		var appName string
 		// If name of the app to be deleted is not passed, consider the current app for deletion
 		if len(args) == 0 {
 			var err error
-			appName, err = application.GetCurrent(client)
+			appName, err = application.GetCurrent(projectName)
 			checkError(err, "")
 		} else {
 			// If app name passed, consider it for deletion
 			appName = args[0]
 		}
 		var confirmDeletion string
-		// Project
-		currentProject := project.GetCurrent(client)
+
 		// Print App Information which will be deleted
-		err := printDeleteAppInfo(client, appName, currentProject)
+		err := printDeleteAppInfo(client, appName, projectName)
 		checkError(err, "")
 
 		if applicationForceDeleteFlag {
@@ -163,17 +165,9 @@ var applicationListCmd = &cobra.Command{
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		if projectName == "" {
-			projectName = project.GetCurrent(client)
-		} else {
-			exist, err := project.Exists(client, projectName)
-			checkError(err, "unable to check if project exists")
-			if !exist {
-				fmt.Printf("Project %s does not exist\n", projectName)
-				os.Exit(1)
-			}
-		}
-		apps, err := application.ListInProject(client, projectName)
+
+		projectName := setNamespace(client)
+		apps, err := application.ListInProject(client)
 		checkError(err, "unable to get list of applications")
 		if len(apps) > 0 {
 			fmt.Printf("The project '%v' has the following applications:\n", projectName)
@@ -211,6 +205,8 @@ var applicationSetCmd = &cobra.Command{
 	}, Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
 		appName := args[0]
+
+		projectName := setNamespace(client)
 		// error if application does not exist
 		exists, err := application.Exists(client, appName)
 		checkError(err, "unable to check if application exists")
@@ -221,7 +217,7 @@ var applicationSetCmd = &cobra.Command{
 
 		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		fmt.Printf("Switched to application: %v\n", args[0])
+		fmt.Printf("Switched to application: %v in project: %v \n", args[0], projectName)
 	},
 }
 
@@ -235,34 +231,38 @@ var applicationDescribeCmd = &cobra.Command{
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		var currentApplication string
+		var appName string
+		projectName := setNamespace(client)
 		if len(args) == 0 {
 			var err error
-			currentApplication, err = application.GetCurrent(client)
+			appName, err = application.GetCurrent(projectName)
 			checkError(err, "")
+			if appName == "" {
+				fmt.Printf("There's no active application in projectName %v", projectName)
+				os.Exit(1)
+			}
 		} else {
-			currentApplication = args[0]
+			appName = args[0]
 			//Check whether application exist or not
-			exists, err := application.Exists(client, currentApplication)
+			exists, err := application.Exists(client, appName)
 			checkError(err, "")
 			if !exists {
-				fmt.Printf("Application with the name %s does not exist\n", currentApplication)
+				fmt.Printf("Application with the name %s does not exist\n", appName)
 				os.Exit(1)
 			}
 		}
-		//Project
-		currentProject := project.GetCurrent(client)
+
 		// List of Component
-		componentList, err := component.List(client, currentApplication, currentProject)
+		componentList, err := component.List(client, appName, projectName)
 		checkError(err, "")
 		if len(componentList) == 0 {
-			fmt.Printf("Application %s has no components deployed.\n", currentApplication)
+			fmt.Printf("Application %s has no components deployed.\n", appName)
 			os.Exit(1)
 		}
-		fmt.Printf("Application %s has:\n", currentApplication)
+		fmt.Printf("Application %s has:\n", appName)
 
 		for _, currentComponent := range componentList {
-			componentType, path, componentURL, appStore, err := component.GetComponentDesc(client, currentComponent.Name, currentApplication, currentProject)
+			componentType, path, componentURL, appStore, err := component.GetComponentDesc(client, currentComponent.Name, appName, projectName)
 			checkError(err, "")
 			printComponentInfo(currentComponent.Name, componentType, path, componentURL, appStore)
 		}
@@ -276,14 +276,20 @@ func init() {
 	// add flags from 'get' to application command
 	applicationCmd.Flags().AddFlagSet(applicationGetCmd.Flags())
 
-	applicationListCmd.Flags().StringVarP(&projectName, "project", "p", "", "List all applications in the specified project")
-
 	applicationCmd.AddCommand(applicationListCmd)
 	applicationCmd.AddCommand(applicationDeleteCmd)
 	applicationCmd.AddCommand(applicationGetCmd)
 	applicationCmd.AddCommand(applicationCreateCmd)
 	applicationCmd.AddCommand(applicationSetCmd)
 	applicationCmd.AddCommand(applicationDescribeCmd)
+
+	//Adding `--project` flag
+	addProjectFlag(applicationListCmd)
+	addProjectFlag(applicationCreateCmd)
+	addProjectFlag(applicationDeleteCmd)
+	addProjectFlag(applicationDescribeCmd)
+	addProjectFlag(applicationSetCmd)
+	addProjectFlag(applicationGetCmd)
 
 	// Add a defined annotation in order to appear in the help menu
 	applicationCmd.Annotations = map[string]string{"command": "other"}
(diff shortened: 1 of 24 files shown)


surajnarwade:
@dgolovin , I updated this PR to be able to run following scenarios, can you please try them out and review this PR, if any condition is left, do let me know

// Application

```
odo app create myapp --project project1

odo app list --project project1

odo app delete myapp --project project1

odo app describe myapp --project1
```

// Component

```
odo list --project project1 --application app1

odo delete nodejs --project project1 --application app1 
```

// Push

```
odo push --project project1 --application app1 --component nodejs
```

// URL

```
odo url create --application app1 --project1 --component nodejs

odo url delete --application app1 --component nodejs --project1

odo url list --application app1 --component nodejs --project1
```

dgolovin:
// Component
odo create nodejs component-name --project project1 --application app1
odo describe component-name --project project1 --application app1
odo log component-name --project project1 --application app1
odo update component-name --project project1 --application app1 --local/--git/--binary path
odo watch component-name --project project1 --application app1

// Storage
odo storage create storage-name --path /path/to --size XXXGi --component component-name  --project project1 --application app1 
odo storage delete storage-name --component component-name  --project project1 --application app1
odo storage list --component component-name  --project project1 --application app1
odo storage mount  --path /path/to --component component-name  --project project1 --application app1
odo storage unmount  /path/to --component component-name  --project project1 --application app1

dgolovin:
Would it be also possible to add short forms for all the switches?
-c --component 
-a --application
-p --project

surajnarwade:
@dgolovin , short forms are confusing sometimes, so I purposefully didn't added :confused:

cmd/cmdutils.go
```diff
@@ -170,3 +172,30 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
 	}
 	fmt.Println("")
 }
+
+// GetAppName returns application name
+func GetAppName(client *occlient.Client) string {
+	project := project.GetCurrent(client)
```
anmolbabu: Should we also consider the `Project` passed instead of only the current project?

anmolbabu: I see from https://gist.github.com/anmolbabu/2077733486031cf6ad92205a56e6275d that the `odo list --project <proj_name>` works only for the current project immaterial of the project passed with --project

surajnarwade:
@kadel , I didn't find any other way to define flags to specific commands. if you notice in `oc` and `kubectl` they also created global flags, when you see `oc -h`, you can't see them because they are clubbed under `oc options`.

I agree with you, For `odo app list` it doesn't make sense to have --component or --application flag, but --project should be there. But having `--component` or `--application` doesn't harm either

kadel:
> @kadel , I didn't find any other way to define flags to specific commands. if you notice in `oc` and `kubectl` they also created global flags, when you see `oc -h`, you can't see them because they are clubbed under `oc options`.
> 
Yes, but everything under `oc options`is a real global flag, that can be used for any command.

> I agree with you, For `odo app list` it doesn't make sense to have --component or --application flag, but --project should be there. But having `--component` or `--application` doesn't harm either

It actually hurts a user experience quite a lot :-|
We can't have flags showing in the help (and not erroring out when they are used) for command where it doesn't make sense.

surajnarwade:
@kadel agree, I am modifying the PR as per your suggestions

surajnarwade:
**Application**

odo app create myapp --project project1

odo app list --project project1

odo app delete myapp --project project1

odo app describe myapp --project1

**Component**

odo create nodejs component-name  --project project1 --application app1

odo describe component-name --project project1 --application app1

odo log component-name --project project1 --application app1

odo list --project project1 --application app1

odo delete component-name --project project1 --application app1 

odo update component-name --project project1 --application app1 --local/--git/--binary path

odo watch component-name --project project1 --application app1

odo push component-name --project project1 --application app1 


**URL**

odo url create --application app1 --project project1 --component nodejs

odo url delete --application app1 --component nodejs --project project1

odo url list --application app1 --component nodejs --project project1

**STORAGE**

odo storage create storage-name --path /path/to --size XXXGi --component component-name --project project1 --application app1

odo storage delete storage-name --component component-name --project project1 --application app1

odo storage list --component component-name --project project1 --application app1

odo storage mount --path /path/to --component component-name --project project1 --application app1

odo storage unmount /path/to --component component-name --project project1 --application app1

pkg/application/application.go
```diff
@@ -207,8 +206,8 @@ func SetCurrent(client *occlient.Client, name string) error {
 	return nil
 }
 
-func Exists(client *occlient.Client, name string) (bool, error) {
-	apps, err := List(client)
+func Exists(client *occlient.Client, project, name string) (bool, error) {
```
kadel: Every function should follow the same order of the arguments: component name, application name, project name.

cmd/application.go
```diff
@@ -55,10 +54,12 @@ var applicationCreateCmd = &cobra.Command{
 		err := validateName(name)
 		checkError(err, "")
 		client := getOcClient()
+		// project
```
cdrage: Better comment description would be good :)

cmd/cmdutils.go
```diff
@@ -170,3 +173,46 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
 	}
 	fmt.Println("")
 }
+
+// GetAppName returns application name
+func GetAppName(client *occlient.Client, project string) string {
```
cdrage: Any way to add tests to this?

cmd/cmdutils.go
```diff
@@ -170,3 +173,46 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+// GetProjectName returns project name
+func GetProjectName(client *occlient.Client) string {
+	if Namespace != "" {
+		_, err := project.Exists(client, Namespace)
+		if err != nil {
+			checkError(err, "")
+		}
+		return Namespace
+	}
+
```
cdrage: Shouldn't we add a "GetCurrent" as well (similar to your GetAppName function?)

cmd/describe.go
```diff
@@ -20,30 +18,45 @@ var describeCmd = &cobra.Command{
...
+		// } else {
+		// 	currentComponent = args[0]
+		// 	//Check whether component exist or not
+		// 	exists, err := component.Exists(client, currentComponent, Application, Project)
+		// 	checkError(err, "")
+		// 	if !exists {
+		// 		fmt.Printf("component with the name %s does not exist\n", currentComponent)
+		// 		os.Exit(1)
+		// 	}
+		// }
```
cdrage: Remove commented out section? I'm assuming this is still WIP this part.

cmd/cmdutils.go
surajnarwade: added on the next line, check this https://github.com/redhat-developer/odo/pull/683/files#diff-28372c336478598898e80dfd05850c30R204

pkg/application/application.go
```diff
@@ -139,83 +137,81 @@ func GetCurrent(client *occlient.Client) (string, error) {
...
 	for _, cfgApp := range cfg.ActiveApplications {
-		if cfgApp.Project == project && cfgApp.Name == name {
+		if cfgApp.Project == projectName && cfgApp.Name == appName {
 			found = true
 			break
 		}
 	}
 	if !found {
-		cfg.AddApplication(name, project)
+		cfg.AddApplication(appName, projectName)
```
golangcibot: Error return value of `cfg.AddApplication` is not checked

cmd/cmdutils.go
```diff
@@ -170,3 +173,48 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
 	}
 	fmt.Println("")
 }
+
+// GetAppName returns application name
```
kadel: returns application name from what?

cmd/cmdutils.go
```diff
@@ -170,3 +173,48 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+	applicationName, err := application.GetCurrent(projectName)
+	checkError(err, "unable to get current application")
+
+	return applicationName
+}
+
+// GetProjectName returns project name
+func getProjectName(client *occlient.Client) string {
+	// Namespace is `--project` flag
+	if Namespace != "" {
```
kadel: using global variable :-( it might be better to pass this as a parameter rather than using it as global variable

cmd/cmdutils.go
```diff
@@ -170,3 +173,48 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+		client.Namespace = Namespace
+		return Namespace
+	}
+	client.Namespace = project.GetCurrent(client)
+	return project.GetCurrent(client)
+
+}
+
+func addProjectFlag(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&Namespace, "project", "p", "", "Project, Defaults to active project")
```
kadel: We need a little bit more descriptive help message.
Also there probably shouldn't be a capital letter after comma

surajnarwade:
cc @kadel , @mik-dass @syamgk @anmolbabu 
PR is ready for review:

HOW TO TEST: https://github.com/redhat-developer/odo/pull/683#issuecomment-424379233

cmd/cmdutils.go
```diff
@@ -170,3 +173,49 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+	}
+	applicationName, err := application.GetCurrent(projectName)
+	checkError(err, "unable to get current application")
+
+	return applicationName
+}
+
+// GetProjectName checks whether project flag is provided,
+// if provided, it validates and returns the project name or returns current project name
+func getProjectName(client *occlient.Client) string {
```
kadel: This function is called `getProjectName`, but also sets the project/namespace in `client`.
Calling this something like setNamespace or something like this might be actually better.

Also, it should explain that in the comment together with saying that it reads `projectFlag` variable as it is not obvious  and it can lead to confusion

cmd/create.go
```diff
@@ -139,6 +145,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 			checkError(err, "")
 			fmt.Printf("Please wait, creating %s component ...\n", componentName)
 			err = component.Build(client, componentName, applicationName, false, true, stdout)
+			fmt.Println("build failed", err)
```
kadel: shouldn't this check if there is an error or not first?
Also, there is a `checkError` function ;-)

cmd/storage.go
```diff
@@ -237,20 +239,20 @@ var storageMountCmd = &cobra.Command{
 
 func init() {
 	storageCreateCmd.Flags().StringVar(&storageSize, "size", "", "Size of storage to add")
-	storageCreateCmd.Flags().StringVar(&storageComponent, "component", "", "Component to add storage to. Defaults to active component.")
+	//storageCreateCmd.Flags().StringVar(&storageComponent, "component", "", "Component to add storage to. Defaults to active component.")
```
kadel: can we remove those comments?

cmd/create.go
surajnarwade: :see_no_evil:

pkg/application/application.go
```diff
@@ -52,7 +49,7 @@ func List(client *occlient.Client) ([]config.ApplicationInfo, error) {
 // Shows also empty applications (empty applications are those that are just
 // mentioned in config file but don't have any object associated with it on cluster).
 func ListInProject(client *occlient.Client, project string) ([]config.ApplicationInfo, error) {
```
kadel: We are already passing `client` that already has namespace set, we shouldn't need pass project here again

pkg/application/application.go
```diff
@@ -94,7 +91,7 @@ func ListInProject(client *occlient.Client, project string) ([]config.Applicatio
 }
 
 // Delete deletes the given application
-func Delete(client *occlient.Client, name string) error {
+func Delete(client *occlient.Client, name, projectName string) error {
```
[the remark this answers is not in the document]
kadel: I found this in many more places

cmd/update.go
```diff
@@ -34,9 +33,9 @@ var updateCmd = &cobra.Command{
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		applicationName, err := application.GetCurrent(client)
-		checkError(err, "")
-		projectName := project.GetCurrent(client)
+
+		projectName := setNamespace(client)
```
golangcibot: ineffectual assignment to `projectName`

kadel:
@surajnarwade please don't amend or squash commits, it i really hard to see what you changes you have done in the PR, and I have to review every

kadel:
sorry, wrong button 😇

cmd/update.go
```diff
@@ -34,9 +33,9 @@ var updateCmd = &cobra.Command{
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		applicationName, err := application.GetCurrent(client)
-		checkError(err, "")
-		projectName := project.GetCurrent(client)
+
+		projectName := getAndSetNamespace(client)
```
golangcibot: ineffectual assignment to `projectName`

dgolovin:
@surajnarwade `odo service *` commands still don't have --project --application options.

surajnarwade:
@dgolovin , there is, can you please check once ?

* odo service create

```
$ odo service create -h
Create a new service from service catalog to deploy on OpenShift.

If service name is not provided, service type value will be used.

A full list of service types that can be deployed are available using: 'odo catalog list services'

Usage:
  odo service create <service_type> [service_name] [flags]

Examples:
  # Create new mysql-persistent service from service catalog.
  odo service create mysql-persistent
	

Flags:
      --application string   Application, defaults to active application
  -h, --help                 help for create
  -p, --project string       Project, defaults to active project

Global Flags:
```

* odo service delete 

```
$ odo service delete -h
Delete an existing service

Usage:
  odo service delete <service_name> [flags]

Examples:
	

Flags:
      --application string   Application, defaults to active application
  -f, --force                Delete service without prompting
  -h, --help                 help for delete
  -p, --project string       Project, defaults to active project

Global Flags:
```

* odo service list

```
$ odo service list -h
List all services in the current application

Usage:
  odo service list [flags]

Examples:
  odo service list
	

Flags:
      --application string   Application, defaults to active application
  -h, --help                 help for list
  -p, --project string       Project, defaults to active project

Global Flags:
```

dgolovin:
@surajnarwade false alarm, sorry. I was running ```odo service --help```.

syamgk:
- [x] **Application**


- [x] odo app create myapp --project project1
- [x] odo app list --project project1
- [x] odo app delete myapp --project project1
- [x] odo app describe myapp --project project1
- [x] odo app set myapp --project project1
- [x] odo app get --project project1
-----------------
- [x] **Component**

- [x] odo create nodejs component-name --project project1 --application app1
- [x] odo describe component-name --project project1 --application app1
- [x] odo log component-name --project project1 --application app1
- [x] odo list --project project1 --application app1
- [x] odo delete component-name --project project1 --application app1
- [x] odo update component-name --project project1 --application app1 --local/--git/--binary path
- [x] odo watch component-name --project project1 --application app1
- [x] odo push component-name --project project1 --application app1
--------------
- [x] **URL** 

- [x] odo url create --application app1 --project project1 --component nodejs
- [x] odo url delete --application app1 --component nodejs --project project1
- [x] odo url list --application app1 --component nodejs --project project1
---------
- [ ] **STORAGE**

- [x] odo storage create storage-name --path /path/to --size XXXGi --component component-name --project project1 --application app1
- [x] odo storage delete storage-name --component component-name --project project1 --application app1
- [x] odo storage list --component component-name --project project1 --application app1
- [x] odo storage mount <storage-name> --path /path/to --component component-name --project project1 --application app1
- [x] odo storage unmount /path/to --component component-name --project project1 --application app1

tested all the above commands 
one suggestion it will be nice to have short flags
one suggestion `--app` for `--application`

cmd/cmdutils.go
```diff
@@ -170,3 +173,51 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+// if not provided, it fetches current namespace and sets it as namespace for further operations
+// getAndSetNamespace also return the project name
+func getAndSetNamespace(client *occlient.Client) string {
+	// projectFlag is `--project` flag
+	if projectFlag != "" {
+		_, err := project.Exists(client, projectFlag)
+		if err != nil {
+			checkError(err, "")
+		}
+		client.Namespace = projectFlag
```
anmolbabu: Any reason why we aren't using SetCurrentProject and only doing a switch in-memory?
Is it to avoid overhead involved in it? But what happens if for example https://github.com/redhat-developer/odo/pull/683/files#diff-cfbbcc0af92db2bb72c86987124fe72cR61 fails? would it not change namespace even for a failed create(although only in-memory)?

cmd/application.go
```diff
@@ -53,17 +51,18 @@ var applicationCreateCmd = &cobra.Command{
...
-		err = application.Create(client, name)
+		getAndSetNamespace(client)
+		fmt.Printf("Creating application: %v\n", appName)
+		err = application.Create(client, appName)
 		checkError(err, "")
-		err = application.SetCurrent(client, name)
+		err = application.SetCurrent(client, appName)
 		checkError(err, "")
-		fmt.Printf("Switched to application: %v\n", name)
+		fmt.Printf("Switched to application: %v\n", appName)
```
anmolbabu: Can we say in which project, if current project is not same as the project in which app was created?

surajnarwade:
@anmolbabu , while rebasing with masters, those `vendor` changes came in my PR, I don't know why :(

cmd/cmdutils.go
surajnarwade: To avoid overhead, we are using `client.Namespace` everywhere instead of confusing with new variable `projectName`

I didn't get your second qs, can you please elaborate ?

> But what happens if for example https://github.com/redhat-developer/odo/pull/683/files#diff-cfbbcc0af92db2bb72c86987124fe72cR61 fails? would it not change namespace even for a failed create(although only in-memory)?.

what is in memory state and the reality state ?

cmd/application.go
surajnarwade: updated :)

cmd/cmdutils.go
```diff
@@ -168,3 +171,51 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
 	fmt.Println("")
 }
+
+// GetAppName returns application name from the provided flag or if flag is not provided, it will return current application name
+func getAppName(client *occlient.Client) string {
+	// applicationFlag is `--application` flag
+	if applicationFlag != "" {
+
+		_, err := application.Exists(client, applicationFlag)
+		if err != nil {
```
mik-dass: we don't need the if statement, checkError() does it automatically

cmd/cmdutils.go
```diff
@@ -168,3 +171,51 @@ func printUnmountedStorage(client *occlient.Client, applicationName string) {
...
+
+// getAndSetNamespace checks whether project flag is provided,
+// if provided, it validates the name and sets it as namespace for further operations
+// if not provided, it fetches current namespace and sets it as namespace for further operations
+// getAndSetNamespace also return the project name
+func getAndSetNamespace(client *occlient.Client) string {
+	// projectFlag is `--project` flag
+	if projectFlag != "" {
+		_, err := project.Exists(client, projectFlag)
+		if err != nil {
```
mik-dass: same as https://github.com/redhat-developer/odo/pull/683/files#r225851025

pkg/application/application.go
```diff
@@ -19,23 +19,20 @@ func getDefaultAppName() string {
 }
 
 // Create a new application
-func Create(client *occlient.Client, applicationName string) error {
-	project := project.GetCurrent(client)
+func Create(client *occlient.Client, appName string) error {
+
+	exists, _ := Exists(client, appName)
```
mik-dass: why are we ignoring the err here? Should we not exit the program if a error occurs?

pkg/application/application.go
```diff
@@ -142,20 +135,17 @@ func GetCurrent(client *occlient.Client) (string, error) {
...
 	// if no Application is active use default
 	if currentApp == "" {
 		// get default application name
 		currentApp = getDefaultAppName()
 		// create if default application does not exist
-		exists, err := Exists(client, currentApp)
-		if err != nil {
-			return "", errors.Wrapf(err, "unable to check if app %v exists", currentApp)
-		}
+		exists, _ := Exists(client, currentApp)
```
mik-dass: Why are we ignoring the error here too? Shouldn't we exit if there is an error here?

mik-dass:
Tested locally and seems to work for me. Just a few comments and questions in the code.

pkg/application/application.go
surajnarwade: we just need to check whether app exist or not, depend on that we are creating the application, that's why we don't care about error here

pkg/application/application.go
surajnarwade: we just need to check whether app exist or not, we don't care about error here

cmd/cmdutils.go
anmolbabu: See for example, if the operation in https://github.com/redhat-developer/odo/pull/683/files#diff-cfbbcc0af92db2bb72c86987124fe72cR61 fails, by then you have set client.Namespace without changing it in config file so, if the operation fails, no change in config file but client.Namespace which is in use has changed so aren't they different was my question?

But it seems, each command is using a new instance of client so this namespace setting would have no impact on them it'll simply be overwritten right? with whatever exists in config...

Ok sounds good...

surajnarwade: awesome :)

kadel:
the vendor updates doesn't look right :-( it shouldn't be there

surajnarwade:
@kadel , somehow during rebase, it came :disappointed:

surajnarwade:
@anmolbabu , @kadel , removed those vendor files, can you please review again ?

anmolbabu:
> @anmolbabu , @kadel , removed those vendor files, can you please review again ?

Ack

anmolbabu:
Also I am finding the below:

```
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
  *      myproject
         project1
[redhat@localhost odo]$ odo app list
The project 'myproject' has the following applications:
ACTIVE     NAME
           odo-exfm
           odo-izxt
           node-xupd
           node-svuf
[redhat@localhost odo]$ odo app set node-svuf
Switched to application: node-svuf in project: myproject
[redhat@localhost odo]$ odo app get
The current application is: node-svuf in project: myproject
[redhat@localhost odo]$ odo app create myapp3 --project project1
Creating application: myapp3 in project: project1
Switched to application: myapp3 in project: project1
[redhat@localhost odo]$ odo project get 
The current project is: myproject
[redhat@localhost odo]$ odo app get 
There's no active application.
You can create one by running 'odo application create <name>'.
```

surajnarwade:
@anmolbabu , you are facing the issue because of https://github.com/redhat-developer/odo/issues/844

anmolbabu:
> @anmolbabu , you are facing the issue because of #844

How is this? I don't have a delete operation in b/w any of the commands in my previous comment or am I missing something

surajnarwade:
@anmolbabu , my bad, I updated the PR, can you please try once again ?

surajnarwade:
> I see edits to
> pkg/project/project.go
> pkg/service/service.go
> pkg/application/application.go
> pkg/component/component.go
> but not to any of their tests(some of them don't have test files at all) but how about those that have? Can we have atleast some basic tests?

methods which we are calling from those files have unit tests already in `occlient_test.go`

I will try to add e2e tests in another PR, once this goes in

anmolbabu:
> @anmolbabu , my bad, I updated the PR, can you please try once again ?

[redhat@localhost odo]$ odo project list
ACTIVE   NAME
         myproject
  *      new-project
         odocmp-1540365259
         odocmp-1540366178
[redhat@localhost odo]$ odo app list
The project 'new-project' has the following applications:
ACTIVE     NAME
           odo-bexi
*          odo-anpn
[redhat@localhost odo]$ odo app list --project myproject
The project 'myproject' has the following applications:
ACTIVE     NAME
           odo-cogh
           odo-fulk
[redhat@localhost odo]$ odo app set odo-cogh --project myproject
Switched to application: odo-cogh in project: myproject
[redhat@localhost odo]$ odo app list
The project 'myproject' has the following applications:
ACTIVE     NAME
*          odo-cogh
           odo-fulk
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
  *      myproject
         new-project
         odocmp-1540365259
         odocmp-1540366178
[redhat@localhost odo]$ odo app list --project new-project
The project 'new-project' has the following applications:
ACTIVE     NAME
           odo-bexi
           odo-anpn

anmolbabu:
@surajnarwade I still seem to be facing the same issue :(

anmolbabu:
@surajnarwade I tried verifying by deleting all resources and clearing odo config file and found the below:

```
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
  *      myproject
[redhat@localhost odo]$ odo app list
There are no applications deployed in the project 'myproject'.
[redhat@localhost odo]$ cat /home/redhat/.kube/odo
activeApplications:
settings:
  updatenotification: true
[redhat@localhost odo]$ odo project create prj1
New project created and now using project : prj1
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
         myproject
  *      prj1
[redhat@localhost odo]$ odo app list
There are no applications deployed in the project 'prj1'.
[redhat@localhost odo]$ odo create nodejs --local /home/redhat/git-srcs/src/github.com/redhat-developer/nodejs-ex -v 4 --project myproject
I1024 20:44:35.047932   21985 create.go:68] Component create called with args: "nodejs", flags: binary=, git=, local=/home/redhat/git-srcs/src/github.com/redhat-developer/nodejs-ex
I1024 20:44:35.055844   21985 occlient.go:265] Trying to connect to server 192.168.42.237:8443
I1024 20:44:35.056541   21985 occlient.go:271] Server https://192.168.42.237:8443 is up
I1024 20:44:35.079247   21985 occlient.go:248] isLoggedIn err:  <nil> 
 output: "developer"
I1024 20:44:35.103282   21985 application.go:208] Setting application odo-onsm as current.
I1024 20:44:35.132851   21985 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 4 6 8 latest]} {perl openshift [5.16 5.20 5.24 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I1024 20:44:35.153706   21985 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 4 6 8 latest]} {perl openshift [5.16 5.20 5.24 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I1024 20:44:35.164164   21985 occlient.go:499] Checking for exact match of builderImage with ImageStream
I1024 20:44:35.164184   21985 occlient.go:509] Found exact image tag match for nodejs:latest
Component 'nodejs-ex-nodejs-drvm' was created and port 8080/TCP was opened
I1024 20:44:35.229348   21985 application.go:208] Setting application odo-onsm as current.

Component 'nodejs-ex-nodejs-drvm' is now set as active component.
I1024 20:44:35.235051   21985 root.go:122] Could not get the latest release information in time. Never mind, exiting gracefully :)
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
  *      myproject
         prj1
[redhat@localhost odo]$ odo app list
The project 'myproject' has the following applications:
ACTIVE     NAME
*          odo-onsm
[redhat@localhost odo]$ odo create nodejs --local /home/redhat/git-srcs/src/github.com/redhat-developer/nodejs-ex -v 4 --project prj1
I1024 20:44:59.481007   22037 create.go:68] Component create called with args: "nodejs", flags: binary=, git=, local=/home/redhat/git-srcs/src/github.com/redhat-developer/nodejs-ex
I1024 20:44:59.485500   22037 occlient.go:265] Trying to connect to server 192.168.42.237:8443
I1024 20:44:59.486164   22037 occlient.go:271] Server https://192.168.42.237:8443 is up
I1024 20:44:59.501321   22037 occlient.go:248] isLoggedIn err:  <nil> 
 output: "developer"
I1024 20:44:59.540400   22037 application.go:208] Setting application odo-llkq as current.
I1024 20:44:59.585369   22037 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 4 6 8 latest]} {perl openshift [5.16 5.20 5.24 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I1024 20:44:59.610971   22037 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 4 6 8 latest]} {perl openshift [5.16 5.20 5.24 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 8.1 9.0 latest]}]
I1024 20:44:59.634148   22037 occlient.go:499] Checking for exact match of builderImage with ImageStream
I1024 20:44:59.634174   22037 occlient.go:509] Found exact image tag match for nodejs:latest
Component 'nodejs-ex-nodejs-qrou' was created and port 8080/TCP was opened
I1024 20:44:59.720857   22037 application.go:208] Setting application odo-llkq as current.

Component 'nodejs-ex-nodejs-qrou' is now set as active component.
I1024 20:44:59.728343   22037 root.go:122] Could not get the latest release information in time. Never mind, exiting gracefully :)
[redhat@localhost odo]$ odo app list
The project 'prj1' has the following applications:
ACTIVE     NAME
*          odo-llkq
[redhat@localhost odo]$ odo project list
ACTIVE   NAME
         myproject
  *      prj1
[redhat@localhost odo]$ odo app list --project myproject
The project 'myproject' has the following applications:
ACTIVE     NAME
           odo-onsm
[redhat@localhost odo]$ cat /home/redhat/.kube/odo
activeApplications:
- active: false
  activeComponent: nodejs-ex-nodejs-drvm
  name: odo-onsm
  project: myproject
- active: true
  activeComponent: nodejs-ex-nodejs-qrou
  name: odo-llkq
  project: prj1
settings:
  updatenotification: true
```

It seems from the above config that active is re-set to false even for a app that is in different project when an app in one project is set as active which seems to cause this issue

surajnarwade:
@anmolbabu , if I am in project p1 and I created app in p2, so shouldn't we set that project as active ?

surajnarwade:
@dgolovin , can you please try out the code and verify all cases that you expect ?

anmolbabu:
> @anmolbabu , if I am in project p1 and I created app in p2, so shouldn't we set that project as active ?

I am not sure :( Sometimes, I feel it might be good to do it but if I think the other way, i.e, why would this(--project flag for example) be required -- It is to enable user to do something in another project while still currently on yet another project(basically without changing the project) I get confused and tend to feel the current/active project after completing the operation should not change... But if the purpose of these flags is to only avoid user having to do 2 steps(firstly, switch the project and then trigger the another action), then the current behavior seems correct. So I think it would be nice to hear from @kadel or @jorgemoralespou on this...

surajnarwade:
@metacosm , can you please review once ?

metacosm:
I think this feature combined with the fact that some commands use arguments that overlap with the new flags leads to confusion. For example, why does `odo app describe foo --project bar` work but not `odo app describe --project bar --app foo`? I will review all the commands in greater details and look at arguments vs. flags.

Regarding switching to project / app based on flag value, I think that the `odo` state shouldn't change when these flags are used. That's the whole point, in my opinion, so that people can interact with entities in a different context without changing the current one.

surajnarwade:
> Regarding switching to project / app based on flag value, I think that the odo state shouldn't change when these flags are used. That's the whole point, in my opinion, so that people can interact with entities in a different context without changing the current one.

cc @dgolovin

kadel:
> I am not sure :( Sometimes, I feel it might be good to do it but if I think the other way, i.e, why would this(--project flag for example) be required -- It is to enable user to do something in another project while still currently on yet another project(basically without changing the project) I get confused and tend to feel the current/active project after completing the operation should not change... But if the purpose of these flags is to only avoid user having to do 2 steps(firstly, switch the project and then trigger the another action), then the current behavior seems correct.
> I also strongly support the argument that decisions should not be taken on behalf of user although suggestions can be given. For example, consider, user does `odo create nodejs --local ./nodejs-ex --project p1` while user's current active project is say `myproject` user has only requested creation of a resource(nodejs component) in currently non-active project(p1). So ideally we should internally do everything it takes to create it and may be prompt to user asking if he would like to switch to the project containing newly created component and do the needful only based on his decision.
> But I am not very sure, So I think it would be nice to hear from @kadel or @jorgemoralespou on this...


What you are saying makes sense.  Let's discuss this in a separate issue/PR. 
This PR is about adding missing flags. It correctly uses the same logic as we have in current odo version.
This PR is already a big one, let's not add even more changes to it.

kadel:
@metacosm 
> I think this feature combined with the fact that some commands use arguments that overlap with the new flags leads to confusion. For example, why does `odo app describe foo --project bar` work but not `odo app describe --project bar --app foo`?

this is what https://github.com/redhat-developer/odo/issues/858 is about. Let's make one change at a time. 
This PR adds missing flags. We need that either way. We can change how we handle flags vs arguments after this gets in.

metacosm:
> > I think this feature combined with the fact that some commands use arguments that overlap with the new flags leads to confusion. For example, why does `odo app describe foo --project bar` work but not `odo app describe --project bar --app foo`?
> 
> this is what #858 is about. Let's make one change at a time.
> This PR adds missing flags. We need that either way. We can change how we handle flags vs arguments after this gets in.

Well, not quite and that's the point of my comment. I would agree with you if the observed behavior was the reverse as the one I described but if we're adding flags, they should work in contexts where they are expected to work, i.e. `odo app describe --project bar --app foo` should work with this PR because `--app foo` should override the default value used by `odo app describe`. 

Then again, this PR could be merged since it's been a long time coming and we could address any remaining issues in subsequent PRs.

anmolbabu:
Merged based on 4 acks on the PR

PR #684, opened 2018-08-30
Add support for odo app list --project name

### LINKED ISSUE #206: list applications in another project
As a Developer, I want to be able to list all of the applications that I currently have in another project.

```
$ ocd app list --project=<name>
You have the following the applications in project [name]

App
MyAwesomeApp (*)
```

Subtask of #159 
Dependency with #205

piyush-garg:
Add e2e regarding the new command
Refactored list function to make code reusable
Use tabwriter to print everything
Fix typos
Fixes #206

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"strings"
 
+	"text/tabwriter"
+
 	"github.com/redhat-developer/odo/pkg/application"
 	"github.com/redhat-developer/odo/pkg/component"
 	"github.com/redhat-developer/odo/pkg/project"
@@ -14,6 +16,7 @@ import (
 var (
 	applicationShortFlag       bool
 	applicationForceDeleteFlag bool
+	projectName                string
 )
 
 // applicationCmd represents the app command
@@ -134,19 +137,39 @@ var applicationListCmd = &cobra.Command{
 	Long:  "List all applications in the current project.",
 	Example: `  # List all applications in the current project 
   odo app list
+
+  # List all applications in the specified project
+  odo app list --project myproject
 	`,
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		apps, err := application.List(client)
-		checkError(err, "")
-		fmt.Printf("ACTIVE   NAME\n")
-		for _, app := range apps {
-			activeMark := " "
-			if app.Active {
-				activeMark = "*"
+		if projectName == "" {
+			projectName = project.GetCurrent(client)
+		} else {
+			exist, err := project.Exists(client, projectName)
+			checkError(err, "unable to check if project exists")
+			if !exist {
+				fmt.Printf("Project %s does not exist\n", projectName)
+				os.Exit(1)
 			}
-			fmt.Printf("  %s      %s\n", activeMark, app.Name)
+		}
+		apps, err := application.ListInProject(client, projectName)
+		checkError(err, "unable to get list of applications")
+		if len(apps) > 0 {
+			fmt.Printf("The project '%v' has the following applications:\n", projectName)
+			tabWriter := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
+			fmt.Fprintln(tabWriter, "ACTIVE", "\t", "NAME")
+			for _, app := range apps {
+				activeMark := " "
+				if app.Active {
+					activeMark = "*"
+				}
+				fmt.Fprintln(tabWriter, activeMark, "\t", app.Name)
+			}
+			tabWriter.Flush()
+		} else {
+			fmt.Printf("There are no applications deployed in the project '%v'.", projectName)
 		}
 	},
 }
@@ -233,6 +256,8 @@ func init() {
 	applicationGetCmd.Flags().BoolVarP(&applicationShortFlag, "short", "q", false, "If true, display only the application name")
 	// add flags from 'get' to application command
 	applicationCmd.Flags().AddFlagSet(applicationGetCmd.Flags())
+
+	applicationListCmd.Flags().StringVarP(&projectName, "project", "p", "", "List all applications in the specified project")
 
 	applicationCmd.AddCommand(applicationListCmd)
 	applicationCmd.AddCommand(applicationDeleteCmd)
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -169,7 +169,7 @@ var applicationListCmd = &cobra.Command{
 			}
 			tabWriter.Flush()
 		} else {
-			fmt.Printf("There are no applications deployed in the project '%v'.", projectName)
+			fmt.Printf("There are no applications deployed in the project '%v'.\n", projectName)
 		}
 	},
 }
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -146,15 +146,19 @@ func Delete(client *occlient.Client, name string, applicationName string, projec
 		return errors.Wrapf(err, "error deleting component %s", name)
 	}
 
-	// Get a list of all active components
+	// Get a list of all components
 	components, err := List(client, applicationName, projectName)
 	if err != nil {
 		return errors.Wrapf(err, "unable to retrieve list of components")
 	}
 
-	// We will *only* set a new component if either len(components) is zero, or the
-	// current component matches the one being deleted.
-	if current := cfg.GetActiveComponent(applicationName, projectName); current == name || len(components) == 0 {
+	// First check is that we want to only update the active component if the component which is getting deleted is the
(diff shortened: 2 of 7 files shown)


piyush1594:
@surajnarwade @mik-dass @anmolbabu @cdrage Can you please review this PR.

@ashetty1 Does e2e test change look good to you.

Thanks

cmd/application.go
```diff
@@ -132,21 +133,40 @@ var applicationListCmd = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		apps, err := application.List(client)
-		checkError(err, "")
-		fmt.Printf("ACTIVE   NAME\n")
-		for _, app := range apps {
-			activeMark := " "
-			if app.Active {
-				activeMark = "*"
+		if projectName == "" {
```
anmolbabu: I think it would be nice to have UTs for all validations..
But, the current practice is that we write UTs for sources in pkg/ but currently not for the cmd/ layer...
So, I feel it would be nice to only have the user facing messages only here but the rest of the things delegated to pkg layer...
 
Also, in the pkg/application/application.go#List I would suggest modifying it to take the form as under:

```

	if projectName != "" {
		exist, err := project.Exists(client, projectName)
		if err!= nil {
			return [], err// may be wrapped error to indicate context instead of just err
		}
		if !exist {
			return [], fmt.Errorf("Requested project does not exist")//Also suggest using odo project list to get list of projects
		}
	} else {
		projectName = project.GetCurrent(client)
	}
	....
	// Normal list code...
```

cmd/application.go
```diff
@@ -132,21 +133,40 @@ var applicationListCmd = &cobra.Command{
...
+			apps, err := application.List(client)
+			checkError(err, "")
+			fmt.Printf("ACTIVE   NAME\n")
+			for _, app := range apps {
+				activeMark := " "
+				if app.Active {
+					activeMark = "*"
+				}
+				fmt.Printf("  %s      %s\n", activeMark, app.Name)
```
anmolbabu: Since you are anyway modifying this section, it would be nice if you can change this to use tabwriter as in https://github.com/redhat-developer/odo/blob/master/cmd/catalog.go#L54-L59

tests/e2e/e2e_test.go
```diff
@@ -180,6 +186,28 @@ var _ = Describe("odoe2e", func() {
...
+			Expect(session).To(ContainSubstring(projName + "-2"))
+		})
+
+		It("should get the project", func() {
+			getProj := runCmd("odo project get --short")
+			Expect(strings.TrimSpace(getProj)).To(Equal(projName + "-2"))
+		})
+
+		It("should be able to list applications in other project", func() {
+			appNames := runCmd("odo app list --project " + projName)
```
anmolbabu: While you are actually currently checking the apps with --project, I feel you are still in the project $projName.
I think it would be more apt to set some other project as active and then trigger this command
So, please move this after https://github.com/redhat-developer/odo/blob/ea6a0cab4c9c5ce27f388f9f994f7e244b671e42/tests/e2e/e2e_test.go#L205-L208

anmolbabu:
Please also note that there is an issue https://github.com/redhat-developer/odo/issues/656 that I feel also includes the issue this PR is trying to solve. So, please discuss with @surajnarwade and let him know that you have resolved this part of the issue(intent is only to avoid duplication of efforts)

cmd/application.go
anmolbabu: where projectName is passed as an arguement to List function

anmolbabu:
Please also add UTs corresponding to the enhancement in the functionality

ashetty1:
Also, we need to have the headers while displaying the list. See this:

```
$ odo app list
ACTIVE   NAME
         ta1
  *      ta2

$ odo app list --project tmp
You have the following applications in project tmp

ta1
ta2

$ odo app list --project odo-1536119387
You have the following applications in project odo-1536119387

a1
a2

$ odo project set odo-1536119387
Switched to project : odo-1536119387

$ odo app list
ACTIVE   NAME
         a1
         a2
```

tests/e2e/e2e_test.go
piyush1594: I am in `projName-2` and executing the `list --project projName` When you create a project it switches to new project same like openshift works. I think we are good here.

piyush1594:
```
$ odo app list
The project 'odo-1536172186' has the following applications:
ACTIVE     NAME
           app
           testing
*          testing-2

$ odo app list --project odo-1536172186
You have the following applications in project odo-1536172186:
NAME
app
testing
```

@ashetty1 Does this looks good to you

tests/e2e/e2e_test.go
anmolbabu: >I am in projName-2 and executing the list --project projName 

aah ok :) I missed that "-2" there.. Thanks :)

>When you create a project it switches to new project same like openshift works. I think we are good here.

Yes .. My eyes missed the presence of "-2" being concatenated in the create line :) 
Thanks for pointing

ashetty1:
@piyush1594 I think the `Active` column should remain which is important to keep the UI consistent, IMHO.

piyush1594:
@ashetty1 If we want `Active` column then code flow will also be nice, and the repetitive code will also get removed.

@cdrage @kadel @jorgemoralespou @surajnarwade Any thoughts on this, I implemented according to the issue description. But if you want like that then awesome. TIA

surajnarwade:
@piyush1594 , Let's have `ACTIVE` column to have consistency :smile:

ashetty1:
This LGTM except this minor bug:
```
$ odo app list --project o1
The project 'o1' has the following applications:
ACTIVE     NAME
           a1
*          a2

$ odo project create o2
New project created and now using project : o2

$ odo app list --project o1
The project 'o1' has the following applications:
ACTIVE     NAME
           a1
*          a2

$ odo app get
There's no active application.

```

Even after creating a new project, it shows the app  from previous project as active. Any way to fix this?

mik-dass:
@ashetty1 @piyush1594  I think we should remove the active column in case of `app list --project temp` completely or maybe we can keep the active column to indicate that this app in project temp is still active. This bug only occurs when the user has no app in the current project, so it might be nice to show the active column in this case. WDYT?

ashetty1:
@mik-dass so unsetting the application when  a new project is created is not possible?

mik-dass:
@ashetty1 No it is possible. I meant that then the **ACTIVE** column will become irrelevant. So it might be better to remove it entirely, WDYT?

ashetty1:
@mik-dass that's fine. Just want the column to stay for the sake of UI consistency

piyush1594:
@ashetty1 Case handled for removing active applications when switching the project.

pkg/component/component.go
```diff
@@ -146,15 +146,17 @@ func Delete(client *occlient.Client, name string, applicationName string, projec
...
 	if err != nil {
 		return errors.Wrapf(err, "unable to retrieve list of components")
 	}
 
 	// We will *only* set a new component if either len(components) is zero, or the
 	// current component matches the one being deleted.
-	if current := cfg.GetActiveComponent(applicationName, projectName); current == name || len(components) == 0 {
+	activeComponent := cfg.GetActiveComponent(applicationName, projectName)
+	activeApplication := cfg.GetActiveApplication(projectName)
+	if (activeComponent == name || len(components) == 0) && (activeApplication == applicationName) {
```
mik-dass: It would be nice to add a comment here explaining the conditions

mik-dass:
@kadel @surajnarwade @cdrage @syamgk @ashetty1 Please review this PR before @piyush1594 closes it again due to inactivity.

pkg/component/component.go
```diff
@@ -146,15 +146,19 @@ func Delete(client *occlient.Client, name string, applicationName string, projec
...
 	if err != nil {
 		return errors.Wrapf(err, "unable to retrieve list of components")
 	}
 
-	// We will *only* set a new component if either len(components) is zero, or the
-	// current component matches the one being deleted.
-	if current := cfg.GetActiveComponent(applicationName, projectName); current == name || len(components) == 0 {
+	// First check is that we want to only update the active component if the component which is getting deleted is the
+	// active component
+	// Second check is that we want to do updation only if it is happening for the active application otherwise we need
```
cdrage: `we want to do an update`

pkg/component/component.go
```diff
@@ -146,15 +146,19 @@ func Delete(client *occlient.Client, name string, applicationName string, projec
...
+	// not to care for the updation of the active component
```
cdrage: `update`

piyush1594:
@cdrage Changes done

mik-dass:
@cdrage Can we merge this?

cdrage:
@mik-dass Implementation still works, code changes LGTM :+1:

PR #701, opened 2018-09-06
Fixes the color issue on Windows for odo push.

### LINKED ISSUE #402: Windows : unknown characters in debug output
There are some strange unknown unicode characters in windows console in odo debug output.

wd0f2ps

mik-dass:
Fixes #402 

Switching to color.output in PushLocal as color.output passed to the function and cast to io.writer is not working on windows. Will solve this in future by moving this more upstream in code at cmd/create.go.

Note : Windows Power Shell doesn't seem to support **FgYellow** color and so the text in not colored in case of power shell but the weird unknown characters are gone. For CMD, the color works fine.

https://superuser.com/questions/1202003/some-git-colors-dont-work-in-powershell-why

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -237,7 +237,9 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 		scanner := bufio.NewScanner(pipeReader)
 		for scanner.Scan() {
 			line := scanner.Text()
-			yellowFprintln(out, line)
+			// color.Output is temporarily used as there is a error when passing in color.Output from cmd/create.go and casting to io.writer in windows
+			// TODO: Fix this in the future, more upstream in the code at cmd/create.go rather than within this function.
+			yellowFprintln(color.Output, line)
 		}
 	}()
 


mik-dass:
@dgolovin Can you please build this PR on windows and mac and check if the issue is solved or not?

dgolovin:
@mik-dass works for me on windows. It was:
image
With fix it is:
image

dgolovin:
@mik-dass old odo and new one built with fix work fine on macOS.

mik-dass:
@dgolovin Thanks for testing :+1: Looks like it's working on windows now

pkg/component/component.go
```diff
@@ -237,7 +237,7 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 		scanner := bufio.NewScanner(pipeReader)
 		for scanner.Scan() {
 			line := scanner.Text()
-			yellowFprintln(out, line)
+			yellowFprintln(color.Output, line)
```
cdrage: :-1: you need to add `color.Output` higher up, we refactored this a while ago and we now pass in `io.Writer` as a parameter.

Have a look at `cmd/create.go` and `cmd/push.go`.

Instead of modifying this line here, you will need to modify it similar to how it's implemented here:

https://github.com/redhat-developer/odo/blob/master/cmd/create.go#L64

By supplying stdout at the beginning and them passing in stdout to the parameter.

mik-dass:
@cdrage I know we pass `color.output` higher up in that code line and pass it as a function parameter, but I think the problem is that color.output gets casted to a **normal** io.writer when passed through the function. That's why it was not working earlier. Even doing `out=color.output` later in the function does not work. The other option is to use PrintfFunc().

mik-dass:
@cdrage The `color.output` works fine in the function itself. But when passed to the other function and cast to io.writer, it fails. So I guess it's a cast error. Not sure though.

cdrage:
@mik-dass Yeah.. it's odd. How about we add a comment / TODO onto the top that we will fix this in the future? It'll as least fix the current colour issue we have with Windows for the time being / temporarily.

mik-dass:
@cdrage I have a added a comment above the line for now.

pkg/component/component.go
```diff
@@ -237,7 +237,8 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 		scanner := bufio.NewScanner(pipeReader)
 		for scanner.Scan() {
 			line := scanner.Text()
-			yellowFprintln(out, line)
+			// switching to color.output as color.output passed to this function and cast to io.writer is not working on windows. Will solve in future.
```
cdrage: You need to add `TODO:` to this so it's searchable in the future.

Have 1 line describing what's happening and another line with `TODO` saying that we will fix this in the future. 

You also need to capitalize `switching`

cdrage: So something like this is acceptable:

```go
// color.Output is temporarily used as there is a type cast error when passing in color.Output from cmd/create.go
// TODO: Fix this in the future more upstream in the code at cmd/create.go rather than within this function.
```

cdrage:
LGTM.

PR #706, opened 2018-09-07
Fix latestReleaseTag() unit test

### LINKED ISSUE #704: Test_getLatestReleaseTage fails on Travis
It seems to occur quite frequently
```
--- FAIL: Test_getLatestReleaseTag (1.36s)
	notify_test.go:18: error getting latest release TagName via API
```

mik-dass:
fixes #704

### DIFF
--- a/pkg/notify/notify_test.go
+++ b/pkg/notify/notify_test.go
@@ -1,30 +1,19 @@
 package notify
 
 import (
-	"context"
 	"fmt"
-	"github.com/google/go-github/github"
-	"reflect"
+	"github.com/blang/semver"
+	"strings"
 	"testing"
 )
 
 func Test_getLatestReleaseTag(t *testing.T) {
-	client := github.NewClient(nil)
-	release, response, err := client.Repositories.GetLatestRelease(context.Background(), "redhat-developer", "odo")
-	if response != nil {
-		defer response.Body.Close()
-	}
-	if err != nil {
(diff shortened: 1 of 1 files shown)


cdrage:
So.... we're exceeding GitHub's API limit:


```sh
	notify_test.go:18: error getting latest release TagName via API, error: GET https://api.github.com/repos/redhat-developer/odo/releases/latest: 403 API rate limit exceeded for 35.184.96.71. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 19m23s]
```

Because we're using the API for the test being written... We need to re-work this test.

@ashetty1

pkg/notify/notify_test.go
```diff
@@ -31,8 +31,8 @@ func Test_getLatestReleaseTag(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			releaseTag, _ := getLatestReleaseTag()
-			if test.success == true {
+			releaseTag, err := getLatestReleaseTag()
```
anmolbabu: Can we mock this func instead using something like monkeypatching golang library and avoid genuine api call altogether? More info @ https://github.com/bouk/monkey

Why an actual get call in UT ?  I feel if you can mock the expected output tht would be more than sufficient in my opinion

surajnarwade:
@syamgk thanks a lot , this LGTM :tada:

cdrage:
This LGTM!

PR #729, opened 2018-09-12
Bump version to 0.0.11

anmolbabu:


anmolbabu:
It looks like `make generate-cli-structure > docs/cli-reference.md` as in https://github.com/redhat-developer/odo/blob/master/docs/development.md#making-a-release doesn't work It completely changes the cli-reference.md to very different one Cc: @cdrage

PR #739, opened 2018-09-14
 Prevent odo fail if `~/.kube` is not present

surajnarwade:


surajnarwade:
This will resolve https://github.com/redhat-developer/odo/issues/738

kadel:
No tests covering this :-(

pkg/config/config.go
```diff
@@ -77,7 +77,15 @@ func New() (*ConfigInfo, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get odo config file")
 	}
-
+	// Check whether directory present or not
+	_, err = os.Stat(filepath.Dir(configFile))
+	if os.IsNotExist(err) {
```
kadel: What if `err` is something else than `IsNotExist`?

PR #648, opened 2018-08-18
Add project delete command

anmolbabu:
This PR adds project delete command.
The core implementation is only wrapped around by this PR
and exposed to the outside as a cli command

fixes #646
Signed-off-by: anmolbabu <anmolbudugutta@gmail.com>

PR #740, opened 2018-09-15
Enhance project delete + Add UTs

### LINKED ISSUE #726: Add unit tests for project delete

anmolbabu:
Delete project enhancements + Unit tests and e2e tests

This PR adds the following:
1. When a project is deleted, the delete project now displays also the
   active project post deletion if there are more projects left post deletion
2. Adds UTs and e2e tests for project delete
3. Handles the case of project list with no projects with a proper message
   indicating that there are no currently available projects and suggests the
   command to create a project instead of displaying an empty table.

fixes #726 #750

### DIFF
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -140,6 +140,7 @@ var projectDeleteCmd = &cobra.Command{
 			fmt.Printf("Aborting deletion of project: %v\n", projectName)
 		}
 
+		fmt.Printf("Deleting project %s...\n(this operation may take some time)\n", projectName)
 		err = project.Delete(client, projectName)
 		if err != nil {
 			checkError(err, "")
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/redhat-developer/odo/pkg/project"
 	"github.com/spf13/cobra"
@@ -124,7 +125,7 @@ var projectDeleteCmd = &cobra.Command{
 		isValidProject, err := project.Exists(client, projectName)
 		checkError(err, "Failed to delete project %s", projectName)
 		if !isValidProject {
-			fmt.Printf("The project %s does not exist. Please check the list of projects using `odo project list`", projectName)
+			fmt.Printf("The project %s does not exist. Please check the list of projects using `odo project list`\n", projectName)
 			os.Exit(1)
 		}
 
@@ -145,6 +146,21 @@ var projectDeleteCmd = &cobra.Command{
 			checkError(err, "")
 		}
 		fmt.Printf("Deleted project : %v\n", projectName)
+
+		// Wait for the delete operation to reflect in the projects list
+		time.Sleep(9 * time.Second)
+
+		// Get Current Project
+		currProject := project.GetCurrent(client)
+
+		// Check if List returns empty, if so, the currProject is showing old currentProject
+		// In openshift, when the project is deleted, it does not reset the current project in kube config file which is used by odo for current project
+		projects, err := project.List(client)
+		checkError(err, "")
+		if len(projects) != 0 {
+			fmt.Printf("%s has been set as active project\n", currProject)
+		}
+
 	},
 }
 
@@ -160,6 +176,10 @@ var projectListCmd = &cobra.Command{
 		client := getOcClient()
 		projects, err := project.List(client)
 		checkError(err, "")
+		if len(projects) == 0 {
+			fmt.Println("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command")
+			return
+		}
 		fmt.Printf("ACTIVE   NAME\n")
 		for _, app := range projects {
 			activeMark := " "
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/redhat-developer/odo/pkg/project"
 	"github.com/spf13/cobra"
@@ -147,9 +146,6 @@ var projectDeleteCmd = &cobra.Command{
 		}
 		fmt.Printf("Deleted project : %v\n", projectName)
 
-		// Wait for the delete operation to reflect in the projects list
-		time.Sleep(9 * time.Second)
-
 		// Get Current Project
 		currProject := project.GetCurrent(client)
 
@@ -159,6 +155,9 @@ var projectDeleteCmd = &cobra.Command{
 		checkError(err, "")
 		if len(projects) != 0 {
 			fmt.Printf("%s has been set as active project\n", currProject)
+		} else {
+			// oc errors out as "error: you do not have rights to view project "$deleted_project"."
+			fmt.Printf("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command\n")
 		}
 
 	},
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -154,10 +154,10 @@ var projectDeleteCmd = &cobra.Command{
 		projects, err := project.List(client)
 		checkError(err, "")
 		if len(projects) != 0 {
-			fmt.Printf("%s has been set as active project\n", currProject)
+			fmt.Printf("%s has been set as the active project\n", currProject)
 		} else {
 			// oc errors out as "error: you do not have rights to view project "$deleted_project"."
-			fmt.Printf("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command\n")
+			fmt.Printf("You are not a member of any projects. You can request a project to be created using the `odo project create <project_name>` command\n")
 		}
 
 	},
@@ -176,7 +176,7 @@ var projectListCmd = &cobra.Command{
 		projects, err := project.List(client)
 		checkError(err, "")
 		if len(projects) == 0 {
-			fmt.Println("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command")
+			fmt.Println("You are not a member of any projects. You can request a project to be created using the `odo project create <project_name>` command")
 			return
 		}
 		fmt.Printf("ACTIVE   NAME\n")
--- a/pkg/occlient/fakeclient.go
+++ b/pkg/occlient/fakeclient.go
@@ -53,5 +53,7 @@ func FakeNew() (*Client, *FakeClientset) {
 	fkclientset.ServiceCatalogClientSet = fakeServiceCatalogClientSet.NewSimpleClientset()
 	client.serviceCatalogClient = fkclientset.ServiceCatalogClientSet.Servicecatalog()
 
+	client.namespace = "testing"
+
 	return &client, &fkclientset
 }
--- a/pkg/occlient/fakeclient.go
+++ b/pkg/occlient/fakeclient.go
@@ -53,7 +53,5 @@ func FakeNew() (*Client, *FakeClientset) {
 	fkclientset.ServiceCatalogClientSet = fakeServiceCatalogClientSet.NewSimpleClientset()
 	client.serviceCatalogClient = fkclientset.ServiceCatalogClientSet.Servicecatalog()
 
-	client.namespace = "testing"
-
 	return &client, &fkclientset
 }
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -301,6 +301,11 @@ func (c *Client) CreateNewProject(name string) error {
 	return nil
 }
 
+// SetCurrentProject sets the passed project as active in kube config file
+// Parameters:
+//	1. project to set as active
+//	2. the instance of client to be used to set the project as active
+// Returns: error. non nil error in case of errors, nil otherwise
 var SetCurrentProject = func(project string, c *Client) error {
 	rawConfig, err := c.kubeConfig.RawConfig()
 	if err != nil {
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
 
 	"github.com/redhat-developer/odo/pkg/util"
 
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1364,7 +1364,7 @@ func (c *Client) DeleteProject(name string) error {
 			}
 		*/
 		if !ok {
-			return fmt.Errorf("Received unexpected signal %+v on project watch channel", val)
+			return fmt.Errorf("received unexpected signal %+v on project watch channel", val)
 		}
 		// So we depend on val.Type as val.Object.Status.Phase is just empty string and not a mapped value constant
 		if prj, ok := val.Object.(*projectv1.Project); ok {
@@ -1376,7 +1376,7 @@ func (c *Client) DeleteProject(name string) error {
 					return nil
 				}
 				if val.Type == watch.Error {
-					return fmt.Errorf("Failed watching the deletion of project %s", name)
+					return fmt.Errorf("failed watching the deletion of project %s", name)
 				}
 			}
 		}
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/runtime"
 	"net"
 	"net/url"
 	"os"
@@ -16,6 +15,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/redhat-developer/odo/pkg/util"
 
@@ -47,7 +48,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/version"
-	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
@@ -85,8 +85,8 @@ type Client struct {
 	serviceCatalogClient servicecatalogclienset.ServicecatalogV1beta1Interface
 	routeClient          routeclientset.RouteV1Interface
 	userClient           userclientset.UserV1Interface
-	kubeConfig           clientcmd.ClientConfig
-	namespace            string
+	KubeConfig           clientcmd.ClientConfig
+	Namespace            string
 }
 
 func New(connectionCheck bool) (*Client, error) {
@@ -95,9 +95,9 @@ func New(connectionCheck bool) (*Client, error) {
 	// initialize client-go clients
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	configOverrides := &clientcmd.ConfigOverrides{}
-	client.kubeConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	client.KubeConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
 
-	config, err := client.kubeConfig.ClientConfig()
+	config, err := client.KubeConfig.ClientConfig()
 	if err != nil {
 		return nil, errors.New(err.Error() + errorMsg)
 	}
@@ -151,11 +151,11 @@ func New(connectionCheck bool) (*Client, error) {
 
 	client.userClient = userClient
 
-	namespace, _, err := client.kubeConfig.Namespace()
+	namespace, _, err := client.KubeConfig.Namespace()
 	if err != nil {
 		return nil, err
 	}
-	client.namespace = namespace
+	client.Namespace = namespace
 
 	// Skip this if connectionCheck is false
 	if !connectionCheck {
@@ -272,7 +272,7 @@ func isServerUp(server string) bool {
 }
 
 func (c *Client) GetCurrentProjectName() string {
-	return c.namespace
+	return c.Namespace
 }
 
 // GetProjectNames return list of existing projects that user has access to.
@@ -302,13 +302,8 @@ func (c *Client) CreateNewProject(name string) error {
 	return nil
 }
 
-// SetCurrentProject sets the passed project as active in kube config file
-// Parameters:
-//	1. project to set as active
-//	2. the instance of client to be used to set the project as active
-// Returns: error. non nil error in case of errors, nil otherwise
-var SetCurrentProject = func(project string, c *Client) error {
-	rawConfig, err := c.kubeConfig.RawConfig()
+func (c *Client) SetCurrentProject(project string) error {
+	rawConfig, err := c.KubeConfig.RawConfig()
 	if err != nil {
 		return errors.Wrapf(err, "unable to switch to %s project", project)
 	}
@@ -319,7 +314,6 @@ var SetCurrentProject = func(project string, c *Client) error {
 	if err != nil {
 		return errors.Wrapf(err, "unable to switch to %s project", project)
 	}
-	c.namespace = project
 	return nil
 }
 
@@ -592,7 +586,7 @@ func (c *Client) NewAppS2I(commonObjectMeta metav1.ObjectMeta, builderImage stri
 	is := imagev1.ImageStream{
 		ObjectMeta: commonObjectMeta,
 	}
-	_, err = c.imageClient.ImageStreams(c.namespace).Create(&is)
+	_, err = c.imageClient.ImageStreams(c.Namespace).Create(&is)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create ImageStream for %s", commonObjectMeta.Name)
 	}
@@ -611,7 +605,7 @@ func (c *Client) NewAppS2I(commonObjectMeta metav1.ObjectMeta, builderImage stri
 	// Generate and create the DeploymentConfig
 	dc := generateGitDeploymentConfig(commonObjectMeta, buildConfig.Spec.Output.To.Name, containerPorts, inputEnvVars)
 
-	_, err = c.appsClient.DeploymentConfigs(c.namespace).Create(&dc)
+	_, err = c.appsClient.DeploymentConfigs(c.Namespace).Create(&dc)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create DeploymentConfig for %s", commonObjectMeta.Name)
 	}
@@ -676,7 +670,7 @@ func (c *Client) BootstrapSupervisoredS2I(commonObjectMeta metav1.ObjectMeta, bu
 	is := imagev1.ImageStream{
 		ObjectMeta: commonObjectMeta,
 	}
-	_, err = c.imageClient.ImageStreams(c.namespace).Create(&is)
+	_, err = c.imageClient.ImageStreams(c.Namespace).Create(&is)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create ImageStream for %s", commonObjectMeta.Name)
 	}
@@ -704,7 +698,7 @@ func (c *Client) BootstrapSupervisoredS2I(commonObjectMeta metav1.ObjectMeta, bu
 		}
 	}
 
-	_, err = c.appsClient.DeploymentConfigs(c.namespace).Create(&dc)
+	_, err = c.appsClient.DeploymentConfigs(c.Namespace).Create(&dc)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create DeploymentConfig for %s", commonObjectMeta.Name)
 	}
@@ -748,7 +742,7 @@ func (c *Client) CreateService(commonObjectMeta metav1.ObjectMeta, containerPort
 			},
 		},
 	}
-	_, err := c.kubeClient.CoreV1().Services(c.namespace).Create(&svc)
+	_, err := c.kubeClient.CoreV1().Services(c.Namespace).Create(&svc)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create Service for %s", commonObjectMeta.Name)
 	}
@@ -795,7 +789,7 @@ func (c *Client) UpdateBuildConfig(buildConfigName string, projectName string, g
 	}
 	buildConfig.Spec.Source = buildSource
 	buildConfig.Annotations = annotations
-	_, err = c.buildClient.BuildConfigs(c.namespace).Update(buildConfig)
+	_, err = c.buildClient.BuildConfigs(c.Namespace).Update(buildConfig)
 	if err != nil {
 		return errors.Wrap(err, "unable to update the component")
 	}
@@ -810,7 +804,7 @@ func (c *Client) UpdateBuildConfig(buildConfigName string, projectName string, g
 func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig) error {
 
 	// Retrieve the current DC
-	currentDC, err := c.GetDeploymentConfigFromName(name, c.namespace)
+	currentDC, err := c.GetDeploymentConfigFromName(name, c.Namespace)
 	if err != nil {
 		return errors.Wrapf(err, "unable to get DeploymentConfig %s", name)
 	}
@@ -864,7 +858,7 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig) error {
 	// Update the current one that's deployed with the new Spec.
 	// despite the "patch" function name, we use update since `.Patch` requires
 	// use to define each and every object we must change. Updating makes it easier.
-	_, err = c.appsClient.DeploymentConfigs(c.namespace).Update(currentDC)
+	_, err = c.appsClient.DeploymentConfigs(c.Namespace).Update(currentDC)
 	if err != nil {
 		return errors.Wrapf(err, "unable to update DeploymentConfig %s", name)
 	}
@@ -882,7 +876,7 @@ func (c *Client) UpdateDCToGit(commonObjectMeta metav1.ObjectMeta, imageName str
 	}
 
 	// Retrieve the current DC in order to obtain what the current inputPorts are..
-	currentDC, err := c.GetDeploymentConfigFromName(commonObjectMeta.Name, c.namespace)
+	currentDC, err := c.GetDeploymentConfigFromName(commonObjectMeta.Name, c.Namespace)
 	if err != nil {
 		return errors.Wrapf(err, "unable to get DeploymentConfig %s", commonObjectMeta.Name)
 	}
@@ -922,7 +916,7 @@ func (c *Client) UpdateDCToSupervisor(commonObjectMeta metav1.ObjectMeta, compon
 	imageNS = imageStream.ObjectMeta.Namespace
 
 	// Retrieve the current DC in order to obtain what the current inputPorts are..
-	currentDC, err := c.GetDeploymentConfigFromName(commonObjectMeta.Name, c.namespace)
+	currentDC, err := c.GetDeploymentConfigFromName(commonObjectMeta.Name, c.Namespace)
 	if err != nil {
 		return errors.Wrapf(err, "unable to get DeploymentConfig %s", commonObjectMeta.Name)
 	}
@@ -969,13 +963,13 @@ func (c *Client) UpdateDCToSupervisor(commonObjectMeta metav1.ObjectMeta, compon
 // dcName is the name of the DeploymentConfig file to be updated
 // annotations contains the annotations for the DeploymentConfig file
 func (c *Client) UpdateDCAnnotations(dcName string, annotations map[string]string) error {
-	dc, err := c.GetDeploymentConfigFromName(dcName, c.namespace)
+	dc, err := c.GetDeploymentConfigFromName(dcName, c.Namespace)
 	if err != nil {
 		return errors.Wrapf(err, "unable to get DeploymentConfig %s", dcName)
 	}
 
 	dc.Annotations = annotations
-	_, err = c.appsClient.DeploymentConfigs(c.namespace).Update(dc)
+	_, err = c.appsClient.DeploymentConfigs(c.Namespace).Update(dc)
 	if err != nil {
 		return errors.Wrapf(err, "unable to uDeploymentConfig config %s", dcName)
 	}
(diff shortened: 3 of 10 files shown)


pkg/project/project_test.go
```diff
@@ -36,6 +34,10 @@ func TestDelete(t *testing.T) {
 
 			// Fake the client with the appropriate arguments
 			client, fakeClientSet := occlient.FakeNew()
+			occlient.SetCurrentProject = func(project string, c *occlient.Client) error {
```
syamgk: nice :+1:

anmolbabu: thanks

pkg/project/project_test.go
```diff
@@ -36,6 +34,10 @@ func TestDelete(t *testing.T) {
...
+				t.Log("Invoked mock SetCurrentProject\n")
```
ashetty1: Might be better to have this message when it is not invoked as a check.

pkg/project/project_test.go
```diff
@@ -36,6 +34,10 @@ func TestDelete(t *testing.T) {
```
ashetty1: Could you tell me how these two cases are different? Don't see anything different.

ashetty1: Need negative cases too.

pkg/project/project_test.go
anmolbabu: Yeah that was a debug print I left accidentally in the PR

pkg/project/project_test.go
anmolbabu: > Could you tell me how these two cases are different? Don't see anything different.

Deleting the only remaining project does not set any other project as active but deleting a project when there are other existing projects does. I think the sentence wasa not conveying the intended meaning fixed it now..

> Need negative cases too.

No mockable/testable negative cases here bcoz, based on https://github.com/redhat-developer/odo/pull/648/files#r216311973, the validations were moved to cmd/project.go that carries no UTs.

pkg/project/project_test.go
```diff
@@ -52,7 +62,10 @@ func TestDelete(t *testing.T) {
 			if !tt.wantErr == (err != nil) {
 				t.Errorf("project Delete() unexpected error %v, wantErr %v", err, tt.wantErr)
 			}
+
+			if tt.wantMockSetCurrentProjectInvoked != mockSetCurrentProjectInvoked {
+				t.Errorf("Expected invocation of SerCurrentProject: %v but actually invoked: %v. Current Project: %v\n", tt.wantMockSetCurrentProjectInvoked, mockSetCurrentProjectInvoked, client.GetCurrentProjectName())
+			}
```
ashetty1: Both values are set by us.  This check makes no sense.

anmolbabu: I agree both values are set by us but I view it as:
The logic to set(mockSetCurrentProjectInvoked) is the logic expected and the tt.wantMockSetCurrentProjectInvoked is the expected value...
And infact it helped me fix a problem in the code :)

pkg/project/project_test.go
```diff
@@ -12,32 +11,42 @@ import (
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 
+			mockSetCurrentProjectInvoked := false
 			// Fake the client with the appropriate arguments
 			client, fakeClientSet := occlient.FakeNew()
+			occlient.SetCurrentProject = func(project string, c *occlient.Client) error {
+				mockSetCurrentProjectInvoked = true
+				return nil
+			}
```
ashetty1: I don't think we need this. We need to test SetCurrentProject separately. Whether a project is set as active or not, we could have that cover under e2e tests.

pkg/project/project_test.go
ashetty1: Negative case: empty project list.

pkg/project/project_test.go
anmolbabu: @ashetty1 Done updated PR accordingly. Please have a look

cdrage:
@anmolbabu This PR actually introduces changes to the main code that's NOT unit tests. 

You'll need to update your title as well as commit to indicate that this changes the behaviour of deleting a project from Odo, tests are included. This is actually a new feature / change.

pkg/project/project_test.go
anmolbabu: Empty project is validated a layer above. So, that case although will not fail here(pkg/project/project.go), is treated as good as a normal case

pkg/project/project.go
```diff
@@ -76,6 +67,15 @@ func Delete(client *occlient.Client, projectName string) error {
 		return errors.Wrap(err, "unable to delete project")
 	}
 
+	// If there will be any projects post the current deletion,
+	// Choose the first project from remainder of the project list to set as current
+	if len(projects) > 0 {
+		currentProject = projects[0].Name
```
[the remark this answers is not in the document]
anmolbabu: Ack added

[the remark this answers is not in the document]
anmolbabu: Hmm but that would add a latency of probably 10-12 seconds.
For some strange reason, delete of a resource takes some time to reflect in the list of resources..
Will add it though

ashetty1:
Hey @anmolbabu, just a suggestion: I don't think using an acronym for unit tests (in this case, UT) is a good idea. It might make it difficult in the future to search for PRs with unit tests, just in case.

pkg/project/project_test.go
syamgk: Unit tests LGTM, I guess the inexistent project validation part is kept under `cmd/project.go` is because of the delete confirmation part.

syamgk:
@anmolbabu as anush said it will be nice to put Unit test instead of UT in the commit its bit cryptic,
apart from that code LGTM!

anmolbabu:
>Hey @anmolbabu, just a suggestion: I don't think using an acronym for unit tests (in this case, UT) is a good idea. It might make it difficult in the future to search for PRs with unit tests, just in case.

@ashetty1 Ack Done

> @anmolbabu as Anush said it will be nice to put Unit test instead of UT in the commit its bit cryptic,
apart from that code LGTM!

@syamgk Ack Done.. I think after I updated the PR, github automatically removed your review could you please have a relook and approve if it looks good?

pkg/project/project.go
```diff
@@ -76,6 +67,16 @@ func Delete(client *occlient.Client, projectName string) error {
 		return errors.Wrap(err, "unable to delete project")
 	}
 
+	SetCurrent(client, "")
```
golangcibot: Error return value of `SetCurrent` is not checked

anmolbabu:

cmd/project.go
```diff
@@ -145,6 +146,21 @@ var projectDeleteCmd = &cobra.Command{
 			checkError(err, "")
 		}
 		fmt.Printf("Deleted project : %v\n", projectName)
+
+		// Wait for the delete operation to reflect in the projects list
+		time.Sleep(9 * time.Second)
```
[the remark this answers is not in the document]
anmolbabu: The `oc` cli itself takes close to 6 seconds ... verified as in `date;oc new-project prj1;date;date;oc delete project prj1; date; echo "Just deleted project";oc projects;sleep 2;date;echo "After 2 secs, the project list is\n";oc projects; sleep 2;date;echo "After 4 secs the projects are:\n";oc projects;sleep 2;date;echo "After 6 secs, the projects are:\n";oc projects;sleep 2;echo "After 8 secs, the projects are :\n";oc projects; sleep 2; echo "After 10 secs, the projects are:\n";oc projects`

cmd/project.go
```diff
@@ -145,6 +146,21 @@ var projectDeleteCmd = &cobra.Command{
...
+
+		// Get Current Project
+		currProject := project.GetCurrent(client)
+
+		// Check if List returns empty, if so, the currProject is showing old currentProject
```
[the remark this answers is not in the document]
anmolbabu: No it is the same even in case of oc client..
Verified using `oc new-project prj4;oc delete project prj4;sleep 6; oc projects;oc project`
@cdrage

cmd/project.go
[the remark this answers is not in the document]
cdrage: @anmolbabu doesn't matter, regardless of how long it takes we should never have `time.Sleep` as the `oc` command can finish in any amount of time. `project.Delete` shouldn't be exiting *until* the project is fully deleted. There should be a ping / check that will continuously loop.

@kadel do you have the same opinion?

cmd/project.go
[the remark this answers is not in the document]
cdrage: 6 seconds isn't a magic number, we should be determining if the project has been deleted through either a loop or simply waiting for `project.Delete` to finish. We should never have `time.Sleep` in the code..

anmolbabu:
@surajnarwade Rebased.. 
ToDo(@anmolbabu) : Incoporate @cdrage comments

surajnarwade:
I found this little bit buggy,

```
$ odo project list
ACTIVE   NAME
         default
         kube-dns
         kube-proxy
         kube-public
         kube-service-catalog
         kube-system
         myproject
  *      newproject
         openshift
         openshift-apiserver
         openshift-automation-service-broker
         openshift-controller-manager
         openshift-core-operators
         openshift-infra
         openshift-node
         openshift-template-service-broker
         openshift-web-console
```
* When I delete the project, it sets the same project name as active
```
[snarwade@redsuraj ~]$ odo project delete newproject
Are you sure you want to delete project newproject? [y/N] y
Deleted project : newproject
newproject has been set as active project
```


* After Deletion:
```
$ odo project list
ACTIVE   NAME
  *      default
         kube-dns
         kube-proxy
         kube-public
         kube-service-catalog
         kube-system
         myproject
         openshift
         openshift-apiserver
         openshift-automation-service-broker
         openshift-controller-manager
         openshift-core-operators
         openshift-infra
         openshift-node
         openshift-template-service-broker
         openshift-web-console
```

anmolbabu:
> I found this little bit buggy,
> 
> ```
> $ odo project list
> ACTIVE   NAME
>          default
>          kube-dns
>          kube-proxy
>          kube-public
>          kube-service-catalog
>          kube-system
>          myproject
>   *      newproject
>          openshift
>          openshift-apiserver
>          openshift-automation-service-broker
>          openshift-controller-manager
>          openshift-core-operators
>          openshift-infra
>          openshift-node
>          openshift-template-service-broker
>          openshift-web-console
> ```
> * When I delete the project, it sets the same project name as active
> 
> ```
> [snarwade@redsuraj ~]$ odo project delete newproject
> Are you sure you want to delete project newproject? [y/N] y
> Deleted project : newproject
> newproject has been set as active project
> ```
> * After Deletion:
> 
> ```
> $ odo project list
> ACTIVE   NAME
>   *      default
>          kube-dns
>          kube-proxy
>          kube-public
>          kube-service-catalog
>          kube-system
>          myproject
>          openshift
>          openshift-apiserver
>          openshift-automation-service-broker
>          openshift-controller-manager
>          openshift-core-operators
>          openshift-infra
>          openshift-node
>          openshift-template-service-broker
>          openshift-web-console
> ```

Aah well not really "sets the same project name as active" but its just that GetCurrentProjectName returns current project name from its local state which was not synced by SetCurrentProject.. So, the bug, anyway fixed it now.. Thanks for pointing :+1:

cmd/project.go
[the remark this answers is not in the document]
anmolbabu: But post the delete of the only remaining project, openshift doesn't seem to unset or set the current/active project as can be seen in:

```
[redhat@localhost odo]$ oc project
error: you do not have rights to view project "myproject".
```

So, we have 2 choices:
1. Set ONLY in our client, the namespace as "" and handle it separately in our code(I feel this can be buggy bcoz occlient namespace on my machine is different from that on some other machine even if we connect to same cluster as it is only local state and not really across machines as its very much within odo and specific to a installation)
2. Current implementation: get list of projects, if length 0 --> No projects available else display currentProject as active project

pkg/project/project_test.go
```diff
@@ -1,9 +1,8 @@
 package project
 
 // ToDo(@anmolbabu) uncomment tests when we have a nicer and cleaner way to stub occlient.go#ModifyConfig
```
syamgk: I guess this line can be removed

anmolbabu: Ack Done

pkg/occlient/occlient_test.go
```diff
@@ -2331,6 +2332,58 @@ func TestStartBuild(t *testing.T) {
 
 }
 
+func fakeProjectStatus(prjStatus corev1.NamespacePhase, prjName string) *projectv1.Project {
```
golangcibot: `fakeProjectStatus` is unused

anmolbabu:
@cdrage @surajnarwade @syamgk Incorporated your comments.. Please revisit it...
@kadel @mik-dass Please review

syamgk:
There is lot of overhead in the implementation of Delete,
this command is taking too much time to delete a project than `oc delete project`

anmolbabu:
> There is a lot of overhead in the implementation of project.Delete as well as in the implementation of projectDeleteCmd,
> this command is taking too much time to delete a project than `oc delete project`

This is bcoz of https://github.com/redhat-developer/odo/pull/740#discussion_r219579069
> 
> I would highly recommend optimizing the number of API calls before merging this.

syamgk:
> > There is a lot of overhead in the implementation of project.Delete as well as in the implementation of projectDeleteCmd,
> > this command is taking too much time to delete a project than `oc delete project`
> 
> @syamgk This(the delay) is bcoz of [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> 
> > I would highly recommend optimizing the number of API calls before merging this.
> 
> @syamgk
> Honestly, there are only the following api calls:
> 
>     1. Exists check -- to see if project requested for delete exists
> 
>     2. Actual deletion:
>        i. List projects: Required to determine the project to set as current project post deletion of the project requested for delete
>        ii. Some non-api config write operations: which sets the project to be deleted as current, without which delete can't happen
>        iii. Delete project api
>        iv. Watch deletion: Wait until the project list reflects actual deletion of project
what is rationale behind doing this(iv)?
>        v. non-api config write operations: which sets the new current project
> 
>     3. non-api get current project already in in-memory config instance
> 
>     4. List project: Required to see if the current project from 3 above is actual current project.. which is required due to, [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r222264813)
> 
> 
> Now, while I felt initially that 1 can be done from the list fetched in 2(i), they were separated based on a comment as in the thread linked to [#648 (comment)](https://github.com/redhat-developer/odo/pull/648#discussion_r216311973) .I feel from the above list, that all the above calls are essential based on my testing... but I might be missing something, please feel free to suggest any possible optimization.

anmolbabu:
> > > There is a lot of overhead in the implementation of project.Delete as well as in the implementation of projectDeleteCmd,
> > > this command is taking too much time to delete a project than `oc delete project`
> > 
> > 
> > @syamgk This(the delay) is bcoz of [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> > > I would highly recommend optimizing the number of API calls before merging this.
> > 
> > 
> > @syamgk
> > Honestly, there are only the following api calls:
> > ```
> > 1. Exists check -- to see if project requested for delete exists
> > 
> > 2. Actual deletion:
> >    i. List projects: Required to determine the project to set as current project post deletion of the project requested for delete
> >    ii. Some non-api config write operations: which sets the project to be deleted as current, without which delete can't happen
> >    iii. Delete project api
> >    iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > ```
> 
> what is rationale behind doing this(2.iv)?
> 

@syamgk https://github.com/redhat-developer/odo/pull/740#discussion_r219579069

> > ```
> >    v. non-api config write operations: which sets the new current project
> > 
> > 3. non-api get current project already in in-memory config instance
> > 
> > 4. List project: Required to see if the current project from 3 above is actual current project.. which is required due to, [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r222264813)
> > ```
> > Now, while I felt initially that 1 can be done from the list fetched in 2(i), they were separated based on a comment as in the thread linked to [#648 (comment)](https://github.com/redhat-developer/odo/pull/648#discussion_r216311973) .I feel from the above list, that all the above calls are essential based on my testing... but I might be missing something, please feel free to suggest any possible optimization.

pkg/project/project.go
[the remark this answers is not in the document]
syamgk: am confused
why does it take 10sec to switch project and print info to terminal?

syamgk:
> > > Honestly, there are only the following api calls:
> > > ```
> > > 1. Exists check -- to see if project requested for delete exists
> > > 
> > > 2. Actual deletion:
> > >    i. List projects: Required to determine the project to set as current project post deletion of the project requested for delete
> > >    ii. Some non-api config write operations: which sets the project to be deleted as current, without which delete can't happen
> > >    iii. Delete project api
> > >    iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > > ```
> > 
> > 
> > what is rationale behind doing this(2.iv)?
> 
> @syamgk [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)

@anmolbabu how is this (watch) related to switching project and printing it to stdout?


> 
> > > ```
> > >    v. non-api config write operations: which sets the new current project
> > > 
> > > 3. non-api get current project already in in-memory config instance
> > > 
> > > 4. List project: Required to see if the current project from 3 above is actual current project.. which is required due to, [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r222264813)
> > > ```

pkg/occlient/occlient.go
```diff
@@ -1306,7 +1313,59 @@ func (c *Client) DeleteProject(name string) error {
...
+	// wait for delete to complete
+	w, err := c.projectClient.Projects().Watch(metav1.ListOptions{
+		FieldSelector: fields.Set{"metadata.name": name}.AsSelector().String(),
+	})
+	if err != nil {
+		return errors.Wrapf(err, "unable to watch project")
+	}
+
+	defer w.Stop()
+	for {
```
syamgk: Can we remove this wait for delete confirmation?

pkg/project/project.go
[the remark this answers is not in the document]
anmolbabu: No I meant:

10 seconds is not for switching project it is for the delete of a project to reflect in project list only after which I show current project.. And its not exactly 10 seconds but around 5-6 seconds so, if we were to show the current project, the command would have to wait/hang for that amount of period(not relying on the time in seconds anymore but watching the project list now... instead of hard coding the wait time interval based on a comment thread as in https://github.com/redhat-developer/odo/pull/740#discussion_r221626756) ...

pkg/occlient/occlient.go
anmolbabu: No, I think that would possibly lead the user into a state of confusion...
Because, the delete takes some time for the project to vanish from the project list and the delay is noticeable.. Now, if we return from delete quickly as soon as we issue the delete api call, and if for some reason the user lists the projects and sees the deleted project still in the list, that would be misleading...

And please note, that the delete of the project involves deletion of all/any resources inside the project which is expected in my opinion to take some time immaterial of whether or not there are any resources within the project(atleast the lookup for presence of any resource has to happen internally..) And whenever the, delete is optimised within the openshift , this watch will also return quickly... optimizing things on our side as well...

WDYT? @syamgk

anmolbabu:
> > > > Honestly, there are only the following api calls:
> > > > ```
> > > > 1. Exists check -- to see if project requested for delete exists
> > > > 
> > > > 2. Actual deletion:
> > > >    i. List projects: Required to determine the project to set as current project post deletion of the project requested for delete
> > > >    ii. Some non-api config write operations: which sets the project to be deleted as current, without which delete can't happen
> > > >    iii. Delete project api
> > > >    iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > > > ```
> > > 
> > > 
> > > what is rationale behind doing this(2.iv)?
> > 
> > 
> > @syamgk [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> 
> @anmolbabu how is this (watch) related to switching project and printing it to stdout?

They are just steps in an ordered sequence :) 
Moreover, consider the case of a single project being deleted, in which case, neither oc client and hence not even Odo reset the current project in the config file(kube config file. It is not maintained at all in the odo config file and I would say this is ideal bcoz, if we maintain it in Odo config file, it would become very specific to the current user and 2 users connecting to the same openshift cluster would still see 2 different versions of this configuration...). I had even tried unsetting the current project but that only sets the current project to `default` project even if that is not available for the current user and when I tried manually editing the kube config file, it only screwed up my config ;) as a blank value for this field renders the config un-interpret-able... Aah, I probably led you into slightly different direction... :) 

Getting back to the topic, in case of a single application being deleted, there is no change in the current active application(also in our code as in https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91->https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103), (and at the time of putting it out to the STDOUT)we will need to fetch the list of projects after delete being fully completed/successful, check if there are no projects in the list and then put up a different message that indicates there are no projects available currently... otherwise, I can show the current project...

@syamgk 

> 
> > > > ```
> > > >    v. non-api config write operations: which sets the new current project
> > > > 
> > > > 3. non-api get current project already in in-memory config instance
> > > > 
> > > > 4. List project: Required to see if the current project from 3 above is actual current project.. which is required due to, [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r222264813)
> > > > ```

cmd/project.go
```diff
@@ -160,6 +160,10 @@ var projectListCmd = &cobra.Command{
 		client := getOcClient()
 		projects, err := project.List(client)
 		checkError(err, "")
+		if len(projects) == 0 {
+			fmt.Println("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command")
+			return
```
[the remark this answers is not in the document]
anmolbabu: thanks :)

anmolbabu:
@cdrage 

>Code LGTM.

>One comment however (will need @surajnarwade or @kadel 's opinion).

>When I'm deleting the project, it's taking about 5 seconds to delete, and there is no prompt / saying that >it's being deleted.

>Should we add a little INFO / log that indicated that "Deleting project from the OpenShift cluster" or >something along those lines? It'd be a good indication on what's happening during the delay.

>```sh
>github.com/redhat-developer/odo  pr_740 ✔                                                                                                                                                                                                                                >3h21m  ⍉
>▶ ./odo project delete myproject -f
>Deleted project : myproject
>You are not a member of any projects. You can request a project to be created with the `odo project >create <project_name>` command
>```

It is printed as in https://github.com/redhat-developer/odo/pull/740/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR1359 but it is a debug msg which will be printed if you use command in verbose mode

pkg/occlient/occlient.go
anmolbabu: @syamgk Please check https://github.com/redhat-developer/odo/pull/740#issuecomment-427768206 which necessiates having to wait until the project is completely deleted...

The only other alternative to not waiting for complete delete, is to not display current project which was the state of the PR earlier but changed in accordance with the comments in order to maintain consistency with the behaviour of delete of other resources...

I feel correctness and completeness is more important than quickness(specially if its a inherent quality of the underlying components).. So, consider a command like component push which is a considerably heavy-weight operation which takes its due time to complete ... But I am open to any suggestions and will not mind changing the PR accordingly... 

@kadel @cdrage @surajnarwade WDYT?

syamgk:
> > > > >    iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > > > > ```
> > > > 
> > > > 
> > > > what is rationale behind doing this(2.iv)?
> > > 
> > > 
> > > @syamgk [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> > 
> > 
> > @anmolbabu how is this (watch) related to switching project and printing it to stdout?
> 
> They are just successive steps in an ordered sequence :)
> Moreover, consider the case of a single project being deleted, in which case, neither oc client and hence not even Odo reset the current project in the config file(kube config file. It is not maintained at all in the odo config file and I would say this is ideal bcoz, if we maintain it in Odo config file, it would become very specific to the current user and 2 users connecting to the same openshift cluster would still see 2 different versions of this configuration...). I had even tried unsetting the current project but that only sets the current project to `default` project even if that is not available for the current user and when I tried manually editing the kube config file, it only screwed up my config ;) as a blank value for this field renders the config un-interpret-able... Aah, I probably led you into slightly different direction... :)
> 
> Getting back to the topic, in case of a single application being deleted, there is no change in the current active application(also in our code as in [https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91->https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103](https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91-%3Ehttps://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103)), (and at the time of putting it out to the STDOUT)we will need to fetch the list of projects after delete being fully completed/successful, check if there are no projects in the list and then put up a different message that indicates there are no projects available currently... otherwise, I can show the current project...
> 

No, we don't actually need to wait and watch the deletion. Deletion of the project can be confirmed in `err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})` . And as we have the list of projects before deleting the specific project we can switch to the available projects. 
Also in future we can a flag for 
1. (default) add confirmations for deletion of each objects
2. (with --force) force delete without confirmation for deletion of objects

anmolbabu:
> > > > > > iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > > > > 
> > > > > 
> > > > > what is rationale behind doing this(2.iv)?
> > > > 
> > > > 
> > > > @syamgk [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> > > 
> > > 
> > > @anmolbabu how is this (watch) related to switching project and printing it to stdout?
> > 
> > 
> > They are just successive steps in an ordered sequence :)
> > Moreover, consider the case of a single project being deleted, in which case, neither oc client and hence not even Odo reset the current project in the config file(kube config file. It is not maintained at all in the odo config file and I would say this is ideal bcoz, if we maintain it in Odo config file, it would become very specific to the current user and 2 users connecting to the same openshift cluster would still see 2 different versions of this configuration...). I had even tried unsetting the current project but that only sets the current project to `default` project even if that is not available for the current user and when I tried manually editing the kube config file, it only screwed up my config ;) as a blank value for this field renders the config un-interpret-able... Aah, I probably led you into slightly different direction... :)
> > Getting back to the topic, in case of a single application being deleted, there is no change in the current active application(also in our code as in [https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91->https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103](https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91-%3Ehttps://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103)), (and at the time of putting it out to the STDOUT)we will need to fetch the list of projects after delete being fully completed/successful, check if there are no projects in the list and then put up a different message that indicates there are no projects available currently... otherwise, I can show the current project...
> 
> No, we don't actually need to wait and watch the deletion. Deletion of the project can be confirmed in `err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})` . And as we have the list of projects before deleting the specific project we can switch to the available projects.

I am not sure if this would be ok to do as it is misleading to show the project as deleted while it still exists for some time(and this some time is a noticebale period of time of the order of 5-6 seconds) after showing it is deleted... I know oc does it... But, that's simply misleading specially considering the fact that it can be used by scripts as you suggested in one of your comments...WhyDYT, not waiting for the delete to fully complete is better? only bcoz it doesn't release the prompt to the user?
But there already are commands that take some time, isn't this one such case at the gain of being apt and complete? But anyway, I might be missing something.. I'll do as you suggest in a separate commit...
We can always look at it again when required...

> Also in future we can a flag for
> 
> 1. (default) add confirmations for deletion of each objects
> 2. (with --force) force delete without confirmation for deletion of objects

syamgk:
> > > > > > > iv. Watch deletion: Wait until the project list reflects actual deletion of project
> > > > > > 
> > > > > > 
> > > > > > what is rationale behind doing this(2.iv)?
> > > > > 
> > > > > 
> > > > > @syamgk [#740 (comment)](https://github.com/redhat-developer/odo/pull/740#discussion_r219579069)
> > > > 
> > > > 
> > > > @anmolbabu how is this (watch) related to switching project and printing it to stdout?
> > > 
> > > 
> > > They are just successive steps in an ordered sequence :)
> > > Moreover, consider the case of a single project being deleted, in which case, neither oc client and hence not even Odo reset the current project in the config file(kube config file. It is not maintained at all in the odo config file and I would say this is ideal bcoz, if we maintain it in Odo config file, it would become very specific to the current user and 2 users connecting to the same openshift cluster would still see 2 different versions of this configuration...). I had even tried unsetting the current project but that only sets the current project to `default` project even if that is not available for the current user and when I tried manually editing the kube config file, it only screwed up my config ;) as a blank value for this field renders the config un-interpret-able... Aah, I probably led you into slightly different direction... :)
> > > Getting back to the topic, in case of a single application being deleted, there is no change in the current active application(also in our code as in [https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91->https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103](https://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004R91-%3Ehttps://github.com/redhat-developer/odo/pull/740/files#diff-249190489286c276602a53b847185004L103)), (and at the time of putting it out to the STDOUT)we will need to fetch the list of projects after delete being fully completed/successful, check if there are no projects in the list and then put up a different message that indicates there are no projects available currently... otherwise, I can show the current project...
> > 
> > 
> > No, we don't actually need to wait and watch the deletion. Deletion of the project can be confirmed in `err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})` . And as we have the list of projects before deleting the specific project we can switch to the available projects.
> 
> @syamgk I am not sure if this would be ok to do as it is misleading to show the project as deleted while it still exists for some time(and this some time is a noticebale period of time of the order of 5-6 seconds) after showing it is deleted... I know oc does it... But, that's simply misleading specially considering the fact that it can be used by scripts as you suggested in one of your comments...WhyDYT, not waiting for the delete to fully complete is better? only bcoz it doesn't release the prompt to the user?

I am used to `oc delete` and my thoughts are from looking from that perspective. 
So my thoughts might be biased. Feel free to proceed with whatever you feel more appropriate 
also lets see what others feel

> But there already are commands that take some time, isn't this one such case at the gain of being apt and complete? But anyway, I might be missing something.. I'll do as you suggest in a separate commit...
> We can always look at it again when required...
> 
> WDYT @kadel @cdrage
> 
> > Also in future we can a flag for
> > 
> > 1. (default) add confirmations for deletion of each objects
> > 2. (with --force) force delete without confirmation for deletion of objects

pkg/occlient/occlient.go
```diff
@@ -1306,7 +1313,59 @@ func (c *Client) DeleteProject(name string) error {
...
+		if !ok {
+			return fmt.Errorf("Received unexpected signal %+v on project watch channel", val)
+		}
+		// So we depend on val.Type as val.Object.Status.Phase is just empty string and not a mapped value constant
+		if prj, ok := val.Object.(*projectv1.Project); ok {
+			glog.V(4).Infof("Status of delete of project %s is %s", name, prj.Status.Phase)
+			switch prj.Status.Phase {
+			//prj.Status.Phase can only be "Terminating" or "Active" or ""
+			case "":
+				if val.Type == watch.Deleted {
```
kadel: You should also check for `Error` type to be sure that you won't end up in the endless loop in case of error.

pkg/occlient/occlient.go
```diff
@@ -297,7 +298,12 @@ func (c *Client) CreateNewProject(name string) error {
 	return nil
 }
 
-func (c *Client) SetCurrentProject(project string) error {
+// SetCurrentProject sets the passed project as active in kube config file
+// Parameters:
+//	1. project to set as active
+//	2. the instance of client to be used to set the project as active
+// Returns: error. non nil error in case of errors, nil otherwise
+var SetCurrentProject = func(project string, c *Client) error {
```
kadel: Why have you done like this? It breaks the pattern that we have been using everywhere :-(

anmolbabu: Only to mute it during testing as in https://github.com/redhat-developer/odo/pull/740/files#diff-c6cce71dcf9031800c52b798ff473d1cR38.

Will try to find a better way to achieve this...

kadel: Could it be done in FakeClient https://github.com/redhat-developer/odo/blob/master/pkg/occlient/fakeclient.go ?

cdrage:
@anmolbabu 

Got a bit lost within the discussion, the feedback I have is that we should:

 - Wait until the project is completely deleted before exiting
 - We shouldn't do this in the background / have time.Sleep anywhere

anmolbabu:
> @anmolbabu
> 
> Got a bit lost within the discussion, the feedback I have is that we should:
> 
> * Wait until the project is completely deleted before exiting
> * We shouldn't do this in the background / have time.Sleep anywhere

@cdrage Ack its currently done(changed as per your comment) that way(wait until the project is deleted using watch on project list and not using time.Sleep)

pkg/project/project.go
```diff
@@ -36,7 +36,7 @@ func SetCurrent(client *occlient.Client, project string) error {
 		return errors.Wrap(err, "unable to unset active component of current project "+project)
 	}
 
-	err = client.SetCurrentProject(project)
+	err = occlient.SetCurrentProject(project, client)
```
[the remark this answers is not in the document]
anmolbabu: Yep, I was trying to mute that function(which involves file operatrions) while unit testing other functions so, I changed the function to this form... I am now(after https://github.com/redhat-developer/odo/pull/740#discussion_r223348197) working on reverting back to its old form and enhancing UT to even mock this function the normal way

pkg/occlient/fakeclient.go
```diff
@@ -53,5 +53,7 @@ func FakeNew() (*Client, *FakeClientset) {
 	fkclientset.ServiceCatalogClientSet = fakeServiceCatalogClientSet.NewSimpleClientset()
 	client.serviceCatalogClient = fkclientset.ServiceCatalogClientSet.Servicecatalog()
 
+	client.namespace = "testing"
```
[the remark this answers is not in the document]
anmolbabu: Ack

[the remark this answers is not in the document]
anmolbabu: It seems the field is un-exported ...
I'll export it

pkg/occlient/occlient.go
anmolbabu: Not FakeClient because it is only file operations..
Now changed it to mock file operations as well

pkg/occlient/occlient.go
anmolbabu: Ack

pkg/occlient/occlient.go
```diff
@@ -83,8 +84,8 @@ type Client struct {
 	serviceCatalogClient servicecatalogclienset.ServicecatalogV1beta1Interface
 	routeClient          routeclientset.RouteV1Interface
 	userClient           userclientset.UserV1Interface
-	kubeConfig           clientcmd.ClientConfig
-	namespace            string
+	KubeConfig           clientcmd.ClientConfig
```
cdrage: Why is this uppercase? It looks like a case of find / replace. You should revert this back to normal. The lowercasing doesn't match with the other variable names.

anmolbabu: Yeah it was required as part of addressing https://github.com/redhat-developer/odo/pull/740#discussion_r225461742 and a similar situation even with mocking config related tests due to SetCurrentProject

anmolbabu:
@kadel @cdrage @surajnarwade @syamgk @mik-dass
* Addressed comments
* Rebased
Please review

cmd/project.go
```diff
@@ -145,6 +145,21 @@ var projectDeleteCmd = &cobra.Command{
...
+
+		// Get Current Project
+		currProject := project.GetCurrent(client)
+
+		// Check if List returns empty, if so, the currProject is showing old currentProject
+		// In openshift, when the project is deleted, it does not reset the current project in kube config file which is used by odo for current project
+		projects, err := project.List(client)
+		checkError(err, "")
+		if len(projects) != 0 {
+			fmt.Printf("%s has been set as active project\n", currProject)
```
cdrage: can't believe I missed this haha.

`has been set as the active project`

cmd/project.go
```diff
@@ -160,6 +175,10 @@ var projectListCmd = &cobra.Command{
 		client := getOcClient()
 		projects, err := project.List(client)
 		checkError(err, "")
+		if len(projects) == 0 {
+			fmt.Println("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command")
```
cdrage: `using the`

cmd/project.go
```diff
@@ -145,6 +145,21 @@ var projectDeleteCmd = &cobra.Command{
...
+
+		// Check if List returns empty, if so, the currProject is showing old currentProject
+		// In openshift, when the project is deleted, it does not reset the current project in kube config file which is used by odo for current project
+		projects, err := project.List(client)
+		checkError(err, "")
+		if len(projects) != 0 {
+			fmt.Printf("%s has been set as active project\n", currProject)
+		} else {
+			// oc errors out as "error: you do not have rights to view project "$deleted_project"."
+			fmt.Printf("You are not a member of any projects. You can request a project to be created with the `odo project create <project_name>` command\n")
```
cdrage: `using the`

cmd/project.go
anmolbabu: Ack :)

cmd/project.go
anmolbabu: Ack

cmd/project.go
anmolbabu: Ack

cdrage:
LGTM! @surajnarwade  / @kadel  @syamgk @mik-dass 

Secondary review?

kadel:
It looks a little weird.

You can see just this for quite a long time
```
▶ odo project delete ff
Are you sure you want to delete project ff? [y/N] y
```
and after waiting you get
```
Deleted project : ff
ff has been set as the active project
```

There should be a message explaining user that it might take some time to delete everything.
Something like this
```
▶ odo project delete ff
Are you sure you want to delete project ff? [y/N] y
Deleting project ff ........ 
(this operation may take some time)
```


I'm not convinced that waiting for project to delete is a best solution. Far easier and elgan solution would be to modify `odo project list` to show  that project mark for deletion.

```
▶ odo project list
ACTIVE   NAME
  *      myproject
         fff        [deletion in process]
```

But as waiting is is already implemented lets go with current implementation.

PR #741, opened 2018-09-15
Provision configurable generation of app and component names making them non-mandatory to be passed

### LINKED ISSUE #216: configurable random name generator for app (and component)
When an app is created, we can provide a name for the app, or if not we will create the app name ourselves. We should provide a "random" generated name for the app, so that I can deploy twice a component as different apps, and I'll be able to easily identify them.

e.g. 

```
cd my source-code-dir
$ od create app
App "app-xyz" has been created
$ od create httpd
httpd component created in app-xyz

$ od create app
App "app-abc" has been created
$ od create httpd
httpd component created in app-abc

$ oc app list
- app-xyz
- app-abc (*)
```

There should be a prefix, which should be configurable, default could be **app**, and then a suffix.
The suffix will be random, and can be a set of random chars within a range (e.g. 4[a-zA-Z]) or can be a value within a list ([a la Docker](https://frightanic.com/computers/docker-default-container-names/))

Subtask of #164 
Dependency on #218

anmolbabu:
Provision configurable generation of app and component names making them non-mandatory to be passed
    
By default(in absence of prefix in config being overriden) app and component
names will be named after the directory from which `odo app create` and
`odo create` components are used along with a random 4 char length string as
unique suffix. However, if the prefix in config is overriden by user, the same
will be used instead of fetching any directory/path names along with the random
4 char length string unique suffix..
This is achieved as follows:
1. For component create, in absence of `prefix` being overriden by user in the
odo config:
   a. If component is created from git, the git url's ending part which
      is the repository name is used for app and compoennt names. For,
      component name, additionally we use component type also.
      case 1: Component creation in absence of an app
        app name: ${repo_name}-${random_4_char_str}
        component name: ${repo_name}-${component_type}-${random_4_char_str}
      case 2: App exists already
        component name: ${repo_name}-${component_type}-${random_4_char_str}
   b. If component is created from local source, the last part of source path,
      which is the source directory, is used for app and compoennt names. For,
      component name, additionally we use component type also.
      case 1: Component creation in absence of an app
        app name: ${src_dir_name}-${random_4_char_str}
        component name: ${src_dir_name}-${component_type}-${random_4_char_str}
      case 2: App exists already
        component name: ${src_dir_name}-${component_type}-${random_4_char_str}
   c. If component is created from binary, the binary name excluding its extension
      is used for app and compoennt names. For component name, additionally we use
      component type also.
      case 1: Component creation in absence of an app
        app name: ${binary_name}-${random_4_char_str}
        component name: ${binary_name}-${component_type}-${random_4_char_str}
      case 2: App exists already
        component name: ${binary_name}-${component_type}-${random_4_char_str}
2. For app create, in absence of `prefix` being overriden by user in the
odo config:
    app name: ${current-dir}-${random_4_char_str}
3. For app create in case of prefix in config being overriden,
    app name: ${prefix}-${random_4_char_str}
4. For component create, in case of a configured prefix,
    app name: ${prefix}-${random_4_char_str}
    component name: ${prefix}-${component_type}-${random_4_char_str}

fixes #216


pkg/config/config.go
```diff
@@ -19,7 +20,8 @@ const (
 // OdoSettings holds all odo specific configurations
 type OdoSettings struct {
 	// Controls if an update notification is shown or not
-	UpdateNotification *bool `json:"updatenotification,omitempty"`
+	UpdateNotification *bool   `json:"updatenotification,omitempty"`
+	Prefix             *string `json:"prefix,omitempty"`
```
syamgk: add a single line comment for the parameter

surajnarwade:
@anmolbabu , if you are want to generate random string, we already have a function:
here,
https://github.com/redhat-developer/odo/blob/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6/pkg/util/util.go#L39

cmd/application.go
```diff
@@ -16,6 +18,8 @@ var (
 	applicationForceDeleteFlag bool
 )
 
+var defaultAppPrefix = "app"
```
surajnarwade: reuse: https://github.com/redhat-developer/odo/blob/master/pkg/application/application.go#L17

surajnarwade:
I think we should modify this: https://github.com/redhat-developer/odo/blob/master/pkg/application/application.go#L141:6

anmolbabu:
> @anmolbabu , if you are want to generate random string, we already have a function:
> here,
> [odo/pkg/util/util.go](https://github.com/redhat-developer/odo/blob/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6/pkg/util/util.go#L39)
> 
> Line 39 in [a7076a6](/redhat-developer/odo/commit/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6)
> 
>  func GenerateRandomString(n int) string {

No but that function is only generating a random string of required length(and I am using this internally in my function for generating a conflict/duplicate resolution as in https://github.com/redhat-developer/odo/pull/741/files#diff-6c6a7fefe37261e2fa8a010cf9e7ae08R106 . As the bug suggests, it is required to have random name be a random real name I just chose names of countries but the library provides many other things which we can as well switch to if we feel that the country names can be misleading or we can even change to docker like names(https://github.com/moby/moby/blob/master/pkg/namesgenerator/names-generator.go) as the bug suggests(but I felt the names can sometimes be a bit awkward combination of adjective and names)...

pkg/config/config.go
anmolbabu: Ack

cmd/application.go
anmolbabu: Ack

anmolbabu:
> I think we should modify this: https://github.com/redhat-developer/odo/blob/master/pkg/application/application.go#L141:6

Done now

pkg/util/util.go
```diff
@@ -45,6 +47,71 @@ func GenerateRandomString(n int) string {
...
+
+// GetRandomName returns a randomly generated name which can be used for naming odo and/or openshift entities
+// prefix: Desired prefix part of the name
+// existList: List to verify that the returned name does not already exist
+// preferredSuffix: Optional suffix if passed, will be checked if prefix-suffix does not exist in existList if not, will add an additional timestamp to make name unique
+// Returns randomname is prefix-suffix, if suffix is passed else generated. Aditionally, if the prefix-suffix is used, it'll be appended with additional 4 char string to take the form prefix-suffix-{a-z A-z}4+
+func GetRandomName(prefix string, existList []string, preferredSuffix string) string {
+	if preferredSuffix == "" {
+		// Generate suffix if not passed, using random country names generated from Pallinder/go-randomdata as suffix
+		preferredSuffix = strings.Replace(
```
mik-dass: We can use strings.NewReplacer() for this task https://golang.org/pkg/strings/#Replacer

pkg/util/util.go
```diff
@@ -45,6 +47,71 @@ func GenerateRandomString(n int) string {
...
+			strings.Replace(strings.Replace(
+				strings.Replace(
+					strings.Replace(
+						strings.ToLower(randomdata.Country(randomdata.FullCountry)),
```
mik-dass: I think we shouldn't use country names, rather `silly names ` in the library would be a better fit

pkg/util/util.go
```diff
@@ -45,6 +47,71 @@ func GenerateRandomString(n int) string {
...
+			-1,
+		)
+	}
+	// name is prefix-suffix
+	name := fmt.Sprintf("%s-%s", prefix, preferredSuffix)
+	// check if generated name is already used in the existList
+	if StringInSlice(name, existList) {
+		prevName := name
+		// keep generating names until generated name is not unique. So, loop terminates when name is unique and hence for condition is false
+		for StringInSlice(prevName, existList) {
```
mik-dass: A max number of retries would a nice addition to prevent it from a infinite loop in worst case scenarios.

pkg/config/config.go
```diff
@@ -20,6 +21,8 @@ const (
 type OdoSettings struct {
 	// Controls if an update notification is shown or not
 	UpdateNotification *bool `json:"updatenotification,omitempty"`
+	// Holds the prefix part of generated random application name
+	Prefix *string `json:"prefix,omitempty"`
```
mik-dass: Would it be a good idea to name this App-Prefix or something like that? I think it would be useful if we later decide to have prefixes for components too, WDYT?

cmd/application.go
```diff
@@ -47,14 +47,31 @@ var applicationCreateCmd = &cobra.Command{
...
+		client := getOcClient()
+		if len(args) == 1 {
+			// The only arg passed is the app name
+			name = args[0]
+		} else {
+			// Desired app name is not passed so, generate a new app name
+			// Fetch existing list of apps
+			apps, err := application.List(client)
+			checkError(err, "")
+			appNames := []string{}
```
mik-dass: Since we are forming this list for searching later only, we can switch to a `map[string]bool`. It will make the searching faster and a lot easier and we won't need the StringInSlice() function too. The search will become a single line like map["app_name"], if it returns `nil`, then the app_name is not used.

pkg/util/util.go
```diff
@@ -45,6 +47,71 @@ func GenerateRandomString(n int) string {
 	return string(b)
 }
 
+// StringInSlice returns if a passed string exists in the passed slice of strings
+func StringInSlice(checkStr string, strSlice []string) bool {
```
mik-dass: We won't need this if we implement the way discussed in https://github.com/redhat-developer/odo/pull/741/files#r219470863

anmolbabu: Same as in https://github.com/redhat-developer/odo/pull/741/files#r219478594 and https://github.com/redhat-developer/odo/pull/741/files#r219478808

pkg/util/util.go
anmolbabu: Nice one :)
Will use it

pkg/util/util.go
anmolbabu: Agreed it would be safe

anmolbabu: will add it

pkg/util/util.go
anmolbabu: Ack.. I think it is also inline with the description in issue...

But, I felt it can generate awkward combination of adjective-noun

I'll change it to silly names for now but would like to hear from @jorgemoralespou and @kadel

pkg/config/config.go
mik-dass: Just a thought to be future proof. In future if we decide to generate component names randomly, in that case having same Prefix for both app and component might be confusing as they start with the same prefix like `blah-peru` and `blah-something` for both app name and component name so might be better to separate them.

pkg/application/application.go
```diff
@@ -10,12 +10,23 @@ import (
...
+	// Get the desired app name prefix from odo config
+	cfg, err := config.New()
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to generate random app name")
+	}
+	// If there's no prefix in config file, use safe default
+	if cfg.OdoSettings.Prefix == nil {
+		cfg.OdoSettings.Prefix = &defaultAppPrefix
+	}
+	appName := util.GetRandomName(*cfg.OdoSettings.Prefix, existingAppNames, "", 3)
```
mik-dass: Would 3 be a good enough number? Maybe 50 for better safety? WDYT?

pkg/util/util.go
```diff
@@ -45,6 +47,61 @@ func GenerateRandomString(n int) string {
...
+		prevName := name
+		trial := 0
+		// keep generating names until generated name is not unique. So, loop terminates when name is unique and hence for condition is false
+		for StringInSlice(prevName, existList) {
+			trial = trial + 1
+			prevName = name
+			// Attempt unique name generation from prefix-suffix by concatenating prefix-suffix withrandom string of length 4
+			prevName = fmt.Sprintf("%s-%s", prevName, GenerateRandomString(4))
+			if trial >= retries {
+				// Avoid infinite loops and fail after passed number of retries
```
mik-dass: Returning an error or printing a debug message might be good idea for better debugging. WDYT?

cmd/application.go
mik-dass: Agreed. But though we have unit tests for StringInSlice(), it will still add the necessity to maintain the function. Also there is the efficiency thing. Anyways I will leave it up to you.

anmolbabu: I would take opinion from @kadel before doing this because I remember receiving a comment from him on my imagestreams PR in a similar scenario where I modified the list, search etc... functions to  use maps instead of slice of objects, but I reverted it to same old state of slice of objects instead of maps with readability as the main reason

pkg/application/application.go
anmolbabu: No I think that's huge :)

Normally I have seen people doing 3 re-tries and giving up but anyway I leave it to @kadel and @jorgemoralespou

pkg/util/util.go
anmolbabu: Aah yes good catch I'll add that

pkg/application/application.go
mik-dass: Maybe 10 for better safety? @kadel @jorgemoralespou WDYT?

cmd/application.go
mik-dass: Another way might be to convert the list to a map inside the GetRandomName() function, which will take one iteration and then performing the search on the map. WDYT?

anmolbabu: I think that sounds good

pkg/config/config.go
anmolbabu: ok so generally, I feel the prefix is a way of identifying who created the resource so immaterial, of the resource type, the prefix is expected to be same in my opinion for a person/group of persons creating the resource

@kadel or @jorgemoralespou can give us a better idea of the use-case for prefix

pkg/application/application_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+func setupEnv(odoconfigfile string) error {
+	err := os.Setenv("ODOCONFIG", odoconfigfile)
+	if err != nil {
+		return errors.Wrap(err, "unable to set ODOCONFIG to odo-test-config")
+	}
+	return nil
+}
+
+func writeTempConfig(config []byte, f *os.File) error {
```
golangcibot: `writeTempConfig` is unused

pkg/application/application_test.go
```diff
@@ -0,0 +1,139 @@
...
+	if config != nil {
+		data, err := yaml.Marshal(&config.Config)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to marshal config %+v", config.Config)
+		}
+		if _, err := configFile.Write(data); err != nil {
+			return nil, errors.Wrapf(err, "unable to write config %+v to mock config file %s", config, configFile)
+		}
+	}
+	setupEnv(configFile.Name())
```
golangcibot: Error return value of `setupEnv` is not checked

pkg/application/application_test.go
```diff
@@ -0,0 +1,139 @@
...
+			odoconfigfile := "odo-test-config"
+
+			configInfo = testingutil.FakeOdoConfig(tt.needPrefix, "")
+
+			configFile, err := setUp(&configInfo, odoconfigfile)
+			if err != nil {
+				t.Errorf("Failed to do required environment setup. Error %v", err)
+			}
+
+			defer cleanupEnv(configFile)
```
golangcibot: Error return value of `cleanupEnv` is not checked

pkg/config/config_test.go
```diff
@@ -972,7 +973,7 @@ func TestSetConfiguration(t *testing.T) {
 			}
 			cfg.Config = tt.existingConfig
 
-			cfg.SetConfiguration(tt.parameter, tt.want)
+			cfg.SetConfiguration(tt.parameter, strconv.FormatBool(tt.want))
```
golangcibot: Error return value of `cfg.SetConfiguration` is not checked

pkg/config/config_test.go
```diff
@@ -4,10 +4,72 @@ import (
...
+	"os/user"
 	"reflect"
+	"strconv"
 	"testing"
+
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
 )
 
+func SetUp(config *ConfigInfo, testFile string) (*os.File, error) {
```
golangcibot: func `SetUp` is unused

pkg/config/config_test.go
```diff
@@ -4,10 +4,72 @@ import (
...
+		}
+		if _, err := configFile.Write(data); err != nil {
+			return nil, errors.Wrapf(err, "unable to write config %+v to mock config file %s", config, configFile.Name())
+		}
+	}
+	return configFile, setupEnv(configFile.Name())
+}
+
+// The invocation of setupTempConfigFile puts the onus of invoking the configCleanUp as well
+func setupTempConfigFile(confFile string) (*os.File, error) {
```
golangcibot: func `setupTempConfigFile` is unused

pkg/config/config_test.go
```diff
@@ -4,10 +4,72 @@ import (
...
+		return nil, err
+	}
+	tmpfile, err := ioutil.TempFile(confFolder, confFile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to create test config file")
+	}
+	return tmpfile, nil
+}
+
+func setupEnv(odoconfigfile string) error {
```
golangcibot: func `setupEnv` is unused

pkg/config/config_test.go
```diff
@@ -4,10 +4,72 @@ import (
...
+	err := os.Setenv("ODOCONFIG", odoconfigfile)
+	if err != nil {
+		return errors.Wrap(err, "unable to set ODOCONFIG to odo-test-config")
+	}
+	return nil
+}
+
+func getConfFolder() (string, error) {
```
golangcibot: func `getConfFolder` is unused

pkg/config/config_test.go
```diff
@@ -4,10 +4,72 @@ import (
...
+		return "", err
+	}
+	dir, err := ioutil.TempDir(currentUser.HomeDir, ".kube")
+	if err != nil {
+		return "", err
+	}
+	return dir, nil
+}
+
+func CleanupEnv(confFile *os.File, t *testing.T) {
```
golangcibot: func `CleanupEnv` is unused

pkg/config/config_test.go
anmolbabu: Used :)

pkg/config/config_test.go
anmolbabu: Used :)

pkg/config/config_test.go
anmolbabu: Used :)

pkg/config/config_test.go
anmolbabu: Used :)

pkg/config/config_test.go
anmolbabu: Used :)

cmd/application.go
```diff
@@ -47,14 +47,31 @@ var applicationCreateCmd = &cobra.Command{
...
+			// Collect the names of existing apps to avoid generated name being already used
+			for _, app := range apps {
```
cdrage: Seems a bit weird that we are using a loop to append appNames and then passing it into `GetDefaultAppName`.

Wouldn't it be better to pass in `apps`, the previous variable you used, into a function that checks to see that the current app doesn't exist? Passing in the struct or whatever parameter `apps` is into another function makes more sense than using this loop.

cmd/create.go
```diff
@@ -86,7 +86,20 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
-		componentImageName, componentType, componentName, componentVersion := util.ParseCreateCmdArgs(args)
+		componentImageName, componentType, defaultComponentName, componentVersion := util.ParseCreateCmdArgs(args)
+
+		componentList, err := component.List(client, applicationName, projectName)
```
cdrage: Add a comment here. Why are we retrieving this?

cmd/create.go
```diff
@@ -86,7 +86,20 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+		checkError(err, "")
+
+		// Get component names
+		var componentNames []string
+		for _, componentInfo := range componentList {
+			componentNames = append(componentNames, componentInfo.Name)
+		}
+
+		// Generate unique name for component
+		componentName, err := component.GetDefaultComponentName(defaultComponentName, componentNames)
```
cdrage: Again, see other comment, I believe that all of this can be shortened into only 4 lines (passing in `apps` or `componentList` into a funciton that checks to see if the new app name exists..)

pkg/application/application.go
```diff
@@ -10,12 +10,28 @@ import (
...
 	"github.com/redhat-developer/odo/pkg/project"
+	"github.com/redhat-developer/odo/pkg/util"
 )
 
-// getDefaultAppName returns application name to be used as a default name in case the user doesn't provide a name.
-// In future this function should generate name with unique suffix (app-xy1h), because there might be multiple applications.
-func getDefaultAppName() string {
-	return "app"
+// GetDefaultAppName returns randomly generated application name with unique configurable prefix suffixed by a randomly generated string which canbe used as a default name in case the user doesn't provide a name.
+func GetDefaultAppName(existingAppNames []string) (string, error) {
```
cdrage: Instead of []string you can use whatever variable `apps` is (I believe it's a struct?)

pkg/application/application.go
```diff
@@ -10,12 +10,28 @@ import (
...
+	var appName string
+	defaultAppPrefix := "app"
+	// Get the desired app name prefix from odo config
```
cdrage: Add newline

pkg/application/application.go
```diff
@@ -10,12 +10,28 @@ import (
...
+	cfg, err := config.New()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to generate random app name")
+	}
+	// If there's no prefix in config file, use safe default
```
cdrage: Add newline to separate code

pkg/testingutil/configs.go
```diff
@@ -0,0 +1,106 @@
...
+	"os/user"
+	"testing"
+
+	"github.com/ghodss/yaml"
+
+	"github.com/pkg/errors"
+	"github.com/redhat-developer/odo/pkg/config"
+)
+
+func getConfFolder() (string, error) {
```
cdrage: missing comments

cmd/application.go
anmolbabu: Ok so you are saying, instead of getting names of apps from []config.ApplicationInfo here, do it inside GetDefaultAppName...  right? Ok

pkg/application/application.go
anmolbabu: Done

pkg/application/application.go
anmolbabu: Ack

pkg/application/application.go
anmolbabu: Ack

cmd/application.go
cdrage: I mean pass in the `apps` variable directly to GetDefaultAppName and iterate through it there.

pkg/testingutil/configs.go
anmolbabu: Ack Done

cmd/application.go
anmolbabu: Ack done.. cahnged it to the same(now)..

Note: https://github.com/redhat-developer/odo/pull/741/files#diff-6c6a7fefe37261e2fa8a010cf9e7ae08R70 is very generic in nature and used for both component name generation and app name generation. So, I am formulating set of names in respective function(GetDefaultAppName in this case) before delegating to https://github.com/redhat-developer/odo/pull/741/files#diff-6c6a7fefe37261e2fa8a010cf9e7ae08R70

cdrage:
This works well:

```sh
github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-swordbold
Switched to application: app-swordbold

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-wolverinebubble
Switched to application: app-wolverinebubble

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-chatterchatter
Switched to application: app-chatterchatter

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-horncypress
Switched to application: app-horncypress

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-houndcalico
Switched to application: app-houndcalico

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-maskmeteor
Switched to application: app-maskmeteor

github.com/redhat-developer/odo  pr_741 ✗                                                                                                                                                                                                                              23h39m ◒  
▶ ./odo app create
Creating application: app-gamblermulberry
Switched to application: app-gamblermulberry
```

My question is, if name is no longer required. Should we do the same with `odo app delete`? I created probably 20 different app's but now I have to manually delete each one of them by name haha.

WDYT @kadel ?

pkg/util/util.go
```diff
@@ -7,6 +7,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
+	randomdata "github.com/Pallinder/go-randomdata"
```
surajnarwade: you can remove `randomdata` alias since package name itself is `randomdata`

surajnarwade: no need to add `alias`

surajnarwade:
I still didn't get it why we need additional library to generate random name instead of https://github.com/redhat-developer/odo/blob/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6/pkg/util/util.go#L39 ?

cc @kadel @cdrage @anmolbabu

anmolbabu:
> I still didn't get it why we need additional library to generate random name instead of
> 
> [odo/pkg/util/util.go](https://github.com/redhat-developer/odo/blob/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6/pkg/util/util.go#L39)
> 
> Line 39 in [a7076a6](/redhat-developer/odo/commit/a7076a66fa7a1d1131ac621f128f0f2497c9f0f6)
> 
>  func GenerateRandomString(n int) string { 
> ?
> cc @kadel @cdrage @anmolbabu

Because "or can be a value within a list (a la Docker)" part of https://github.com/redhat-developer/odo/issues/216#issue-303804643 I thought this would be better than random strings but it seems I missed the last comment(added 7 days ago) by @jorgemoralespou as in https://github.com/redhat-developer/odo/issues/216#issuecomment-425156438

cdrage:
@anmolbabu 

No I don't mean delete all, I mean a `odo app delete -f` command that deletes the current app, similar to how we have it with component, project, etc.

pkg/application/application.go
```diff
@@ -10,12 +10,44 @@ import (
...
+		return "", errors.Wrap(err, "unable to generate random app name")
+	}
+
+	// If there's no prefix in config file or it is equal to $DIR, use safe default which is the name of current directory
+	if cfg.OdoSettings.Prefix == nil || *cfg.OdoSettings.Prefix == config.ConfigPrefixDir {
+		prefix, err := util.GetComponentDir("", util.NONE)
+		if err != nil {
+			return "", errors.Wrap(err, "unable to generate random app name")
+		}
+		appName, err = util.GetRandomName(prefix, appPrefixMaxLen, existingAppNames, appNameMaxRetries)
```
golangcibot: ineffectual assignment to `err`

mik-dass:
The golang Ci is reporting 5 unused functions. Please check if we need them or they can be removed.

anmolbabu:
> @anmolbabu
> 
> No I don't mean delete all, I mean a `odo app delete -f` command that deletes the current app, similar to how we have it with component, project, etc.

@cdrage Ack Done

anmolbabu:

anmolbabu:
> The golang Ci is reporting 5 unused functions. Please check if we need them or they can be removed.

@mik-dass Fixed it

cmd/application.go
```diff
@@ -50,14 +50,27 @@ var applicationCreateCmd = &cobra.Command{
...
+			// The only arg passed is the app name
+			name = args[0]
+		} else {
+			// Desired app name is not passed so, generate a new app name
+			// Fetch existing list of apps
+			apps, err := application.List(client)
+			checkError(err, "")
+
+			// Generate a random name that's not already in use for the existing apps
+			fmt.Println("Now generating random name")
```
surajnarwade: remove this line :)

pkg/application/application.go
```diff
@@ -10,12 +10,47 @@ import (
...
+
+	// Get list of app names
+	for _, app := range existingApps {
+		existingAppNames = append(existingAppNames, app.Name)
+	}
+
+	// Get the desired app name prefix from odo config
+	cfg, err := config.New()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to generate random app name")
```
surajnarwade: `unable to fetch config`

pkg/util/util.go
```diff
@@ -122,3 +234,18 @@ func GenFileUrl(location string, os string) string {
 	}
 	return "file://" + urlPath
 }
+
+// GetDNS1123Name Converts passed string into DNS-1123 string
```
surajnarwade: You may find this function in `k8s.io/apimachinery`, please look for it and call it rather than rewriting

cmd/application.go
anmolbabu: Ack :)

pkg/application/application.go
anmolbabu: Ack

pkg/util/util.go
anmolbabu: @surajnarwade it seems I am unable to find the function you are referring to...
I am able to find a validation function but not the one that formulate given string for me in the DNS-1123 form...
Could you please help me find it

cdrage: @surajnarwade I couldn't find it either

pkg/component/component_test.go
```diff
@@ -89,3 +92,79 @@ func TestGetComponentPorts(t *testing.T) {
...
+			wantRE:             "testing-nodejs-*",
+			needPrefix:         true,
+		},
+		{
+			testName:           "Case: App prefix configured",
+			componentType:      "wildfly",
+			componentPathType:  util.BINARY,
+			componentPath:      "./testing.war",
+			existingComponents: []ComponentInfo{},
+			wantErr:            false,
```
cdrage: we should create a test to find an error.

mik-dass:
@anmolbabu rebase needed

pkg/util/util.go
mik-dass: we don't need this function any more. It's not used anywhere except the test for the function.

pkg/util/util_test.go
```diff
@@ -3,9 +3,228 @@ package util
...
+			r, _ := regexp.Compile(fmt.Sprintf("[a-z]{%d}", tt.strLength))
+			match := r.MatchString(name)
+			if !match {
+				t.Errorf("Randomly generated string %s which does not match regexp %s", name, fmt.Sprintf("[a-z]{%d}", tt.strLength))
+			}
+		})
+	}
+}
+
+func TestStringInSlice(t *testing.T) {
```
mik-dass: We don't need this test as the function is not used anywhere

anmolbabu: Ack Done

pkg/application/application_test.go
```diff
@@ -0,0 +1,77 @@
...
+			existingApps: []config.ApplicationInfo{},
+			wantErr:      false,
+			wantRE:       "testing-*",
+			needPrefix:   true,
+			prefix:       "testing",
+		},
+		{
+			testName:     "Case: App prefix set to $DIR",
+			existingApps: []config.ApplicationInfo{},
+			wantErr:      false,
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/component/component_test.go
anmolbabu: Same as in https://github.com/redhat-developer/odo/pull/741#discussion_r226592953

anmolbabu:
>So when you delete an app, it doesn't select an active app after deletion, is this intended?
```sh
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                               1d
▶ ./odo delete -f
Component odo-nodejs-kwfs from application odo-zbcq has been deleted
Default component set to: odo-nodejs-unjk
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                               1d
▶ ./odo delete -f
Component odo-nodejs-unjk from application odo-zbcq has been deleted
No default component has been set
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                               1d
▶ ./odo delete -f
There is no component set
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                              1d  ⍉
▶ ./odo delete -f
There is no component set
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                              1d  ⍉
▶ ./odo app delete -f
Deleted application: odo-zbcq
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                               1d
▶ ./odo app delete -f
application  doesn't exist
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                              1d  ⍉
▶ ./odo app delete -f
application  doesn't exist
github.com/redhat-developer/odo  pr_741 ✔                                                                                                                                                                                                                                                                                                                              1d  ⍉
▶ ./odo app list
The project 'foo' has the following applications:
ACTIVE     NAME
           odo-uona
           odo-iurb
           odo-ehaa
           odo-kyim
           odo-rfid
           odo-wegm
           odo-ixss
           odo-gesd
```

It seems it didn't exist from before.. Added the functionality but I think adding UTs for it would be beyond the scope of this PR.. Would it be ok if I take this up in a subsequent PR in favour of pushing this PR in? @cdrage @kadel

anmolbabu:
> Please modify `odo utils config view` to display the current prefix value after adding one
```
[mrinaldas@localhost odo]$ odo utils config set Prefix blah
[mrinaldas@localhost odo]$ odo utils config view
UpdateNotification    true
```

@mik-dass Thanks for pointing .. Done now

anmolbabu:
> @anmolbabu rebase needed

Ack done

anmolbabu:

jorgemoralespou:
I think that when you delete a component or an app, another one should not be selected. This is misleading behavior.

jorgemoralespou:
@anmolbabu that is true for project :-D not for component as that concept does not exist in k8s/openshift.
Hence I didn't comment on the other PR you have for project, but on this one for application/component

anmolbabu:
> @anmolbabu that is true for project :-D not for component as that concept does not exist in k8s/openshift.
> Hence I didn't comment on the other PR you have for project, but on this one for application/component

@jorgemoralespou Ok I'll quickly revert the commit that sets the current app.. Thanks for the info

pkg/util/util.go
```diff
@@ -139,3 +143,114 @@ func ConvertKeyValueStringToMap(params []string) map[string]string {
...
+		}
+		// If found to be unique, set name as generated name
+		name = prevName
+	}
+	// return name
+	return name, nil
+}
+
+// ComponentCreateType is an enum to indicate the type of source of component -- local source/binary or git for the generation of app/component names
+type ComponentCreateType string
```
kadel: This and constants below are all related to components. It should probably live in a component package.

It is good practice to have type definitions and constants defined at the beginning of the file to make it easier to find.

pkg/util/util.go
```diff
@@ -139,3 +143,114 @@ func ConvertKeyValueStringToMap(params []string) map[string]string {
...
+	// NONE indicates there's no information about the type of source of the component
+	NONE ComponentCreateType = ""
+)
+
+// GetComponentDir returns source repo name
+// Parameters:
+//		path: git url or source path or binary path
+//		paramType: One of ComponentCreateType as in GIT/SOURCE/BINARY
+// Returns: directory name
+func GetComponentDir(path string, paramType ComponentCreateType) (string, error) {
```
kadel: this is also related to the component, and should be in the component package

pkg/config/config.go
```diff
@@ -14,12 +15,17 @@ import (
 const (
 	configEnvName  = "ODOCONFIG"
 	configFileName = "odo"
+	// ConfigPrefixDir The default value for config `prefix`.
+	// When set to $DIR, the directory name of component dir(or current-dir if not available) will be used as prefix for app and component names
+	ConfigPrefixDir = "$DIR"
```
kadel: can we use something else for the value that doesn't also represent env variable?
This will be tricky to enter on the command line as it will have to be escaped.
Also, it is a little bit misleading. For git and binary component it will be something else than name of the directory.

something like AUTOMATIC might be a more appropriate name. What do you think?

Do we even need this, can we just assume that if the prefix is not set than it is this?

pkg/application/application_test.go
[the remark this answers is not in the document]
kadel: I would say that you can remove `wantError` and associated checks if it is always false, to make code a little bit simpler.

pkg/config/config.go
kadel: I would also use a little bit more specific name than just Prefix. It is too generic.
Ideally, it should be a little bit descriptive and give the user a hint what this configuration means.

anmolbabu: Would AutoGenNamePrefix be good enough?

cmd/config.go
```diff
@@ -50,6 +49,8 @@ UpdateNotification - Controls if an update notification is shown or not (true or
 	Example: `
    # Set a configuration value
    odo utils config set UpdateNotification false
+   odo utils config set Prefix "\$DIR"
+   odo utils config set Prefix app
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/util/util.go
anmolbabu: Ack agreed

pkg/util/util.go
anmolbabu: Ack

pkg/config/config.go
anmolbabu: Ack

pkg/application/application_test.go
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/application/application_test.go
```diff
@@ -0,0 +1,72 @@
...
+			needPrefix:   true,
+			prefix:       "AUTOMATIC",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Log("Running test: ", tt.testName)
+		t.Run(tt.testName, func(t *testing.T) {
+
+			odoConfigFile, kubeConfigFile, err := testingutil.SetUp(
```
golangcibot: ineffectual assignment to `err`

pkg/component/component_test.go
```diff
@@ -89,3 +91,144 @@ func TestGetComponentPorts(t *testing.T) {
...
+			wantRE:             "testing-wildfly-*",
+			needPrefix:         true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Log("Running test: ", tt.testName)
+		t.Run(tt.testName, func(t *testing.T) {
+			odoConfigFile, kubeConfigFile, err := testingutil.SetUp(
```
golangcibot: ineffectual assignment to `err`

pkg/component/component.go
```diff
@@ -25,13 +27,101 @@ import (
 // it can be also file://
 const componentSourceURLAnnotation = "app.kubernetes.io/url"
 const componentSourceTypeAnnotation = "app.kubernetes.io/component-source-type"
+const componentRandomNamePartsMaxLen = 12
+const componentNameMaxRetries = 3
+const componentNameMaxLen = -1
+
+// ComponentCreateType is an enum to indicate the type of source of component -- local source/binary or git for the generation of app/component names
+type ComponentCreateType string
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/config/config.go
```diff
@@ -153,6 +165,14 @@ func (c *ConfigInfo) GetUpdateNotification() bool {
 	return *c.OdoSettings.UpdateNotification
 }
 
+// GetPrefix returns the value of Prefix from config
```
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/component/component.go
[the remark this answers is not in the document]
anmolbabu: Ack

pkg/application/application_test.go
golangcibot: ineffectual assignment to `err`

pkg/component/component_test.go
golangcibot: ineffectual assignment to `err`

pkg/application/application_test.go
anmolbabu: Ack

pkg/component/component_test.go
anmolbabu: Ack

anmolbabu:
@kadel @cdrage @surajnarwade @syamgk @mik-dass Rebased and comments incoporated. Please revisit the PR

pkg/config/config.go
anmolbabu: > Do we even need this, can we just assume that if the prefix is not set than it is this?

@kadel Without a reserved keyword, if user wants to reset to default after configuring it to a non-default value, he would only be left with 1 route => remove the configurable manually from conf file(and then default will be picked up from within code). But, reserving a word takes away a possible value for the configurable.. Not sure what is the best way.. Please suggest

anmolbabu:
>I didn't get an answer for the question here:
https://github.com/redhat-developer/odo/pull/741#discussion_r226688256
Do we really need `AUTOMATIC` value? 
Why not just assume that when it is not set `(==nil)` than it is automatic? Why do we need a special value for this?

@kadel Now replied as in, https://github.com/redhat-developer/odo/pull/741#discussion_r227037274

pkg/config/config.go
kadel: How about the empty string? `odo utils config set NamePrefix ""`

kadel:
ping @mik-dass @cdrage @surajnarwade

mik-dass:
The latest changes don't seem to work for applications but they are working fror components
```
[mrinaldas@localhost odo]$ ./odo app list
The project 'myproject' has the following applications:
ACTIVE     NAME
           app-1
*          app-3
[mrinaldas@localhost odo]$ ./odo create nodejs --local ~/nodejs-ex/
Component 'nodejs-ex-nodejs-guiu' was created and port 8080/TCP was opened

Component 'nodejs-ex-nodejs-guiu' is now set as active component.
```
I am in the nodejs-ex directory and the prefix should be nodejs-ex. Even after adding a NamePrefix, it's not working

```
[mrinaldas@localhost odo]$ ./odo utils config view
UpdateNotification    true
NamePrefix            blah
[mrinaldas@localhost odo]$ ./odo app create app-34
Creating application: app-34
Switched to application: app-34
```

cmd/application.go
```diff
@@ -46,18 +46,34 @@ var applicationCmd = &cobra.Command{
 var applicationCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: "Create an application",
-	Long:  "Create an application",
+	Long: `Create an application.
+If no app name is passed, a default app name will be generated as per $dir-$random-4-char-str or if prefix configured, $prefix-$random-4-char-str
```
cdrage: Seems a bit too long, even when doing `odo create --help` this seems long.

Can you shorten it to just: `If no app name is passed, a default app name will be auto-generated` 

There's no need to tell the user what format it'll be, similar to how OpenShift or `kubectl` doesn't

cmd/application.go
```diff
@@ -98,15 +114,19 @@ var applicationDeleteCmd = &cobra.Command{
...
-			return fmt.Errorf("Please provide application name")
-		}
-		return nil
-	},
+	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		client := getOcClient()
-		appName := args[0]
+		var appName string
+		// If name of app to be deleted is not passed, consider the current app for deletion
```
cdrage: `If the name of the app`

cmd/config.go
```diff
@@ -19,7 +18,8 @@ var configurationCmd = &cobra.Command{
 	Long: `Modifies Odo specific configuration settings within the config file.
 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false)
+NamePrefix - Uses the value of this configurable as a prefix for naming app/component.If set to empty string, the name will be auto generated based on name of component/component source/current directory`,
```
cdrage: Too long. No need for such a long description :(

Maybe shorten to just: `Default prefix is 'odo'. Use this value to set a default name prefix`?

cmd/create.go
```diff
@@ -29,7 +29,10 @@ var componentCreateCmd = &cobra.Command{
 	Short: "Create a new component",
 	Long: `Create a new component to deploy on OpenShift.
 
-If component name is not provided, component type value will be used for the name.
+If component name is not provided, a random name $dir-$component_type-$random-4-char-str, where $dir  will be generated as per the following:
+Create using git url: repo name is $dir
+Create using local path: Directory name of the source in the local path
+Create using binary path: The name of the binary without its extension
```
cdrage: All of this input makes `odo create --help` command *way* too long. We don't need to explain everything.. See other comment, we should only have 1 sentence explaining that the name will be autogenerated.

docs/cli-reference.md
```diff
@@ -216,7 +217,10 @@ Catalog related operations
 
 Create a new component to deploy on OpenShift.
 
-If component name is not provided, component type value will be used for the name.
+If component name is not provided, a random name $dir-$component_type-$random-4-char-str, where $dir  will be generated as per the following:
+Create using git url: repo name is $dir
+Create using local path: Directory name of the source in the local path
+Create using binary path: The name of the binary without its extension
```
cdrage: Again, see other comment

pkg/application/application.go
```diff
@@ -7,15 +7,51 @@ import (
...
-// getDefaultAppName returns application name to be used as a default name in case the user doesn't provide a name.
-// In future this function should generate name with unique suffix (app-xy1h), because there might be multiple applications.
-func getDefaultAppName() string {
-	return "app"
+const (
+	appPrefixMaxLen   = 12
+	appNameMaxRetries = 3
+)
+
+// GetDefaultAppName returns randomly generated application name with unique configurable prefix suffixed by a randomly generated string which canbe used as a default name in case the user doesn't provide a name.
```
cdrage: `can be`

cmd/application.go
anmolbabu: Ack

cmd/application.go
anmolbabu: Ack

cmd/config.go
anmolbabu: No, default prefix is current directory name.
So, changed it to the following:"NamePrefix - Default prefix is current directory name. Use this value to set a default name prefix."

cmd/create.go

pkg/application/application.go
anmolbabu: Ack

docs/cli-reference.md
anmolbabu: Ack

cmd/config.go
cdrage: Still doesn't show that you updated this comment..

Grammar is also: `Default prefix is the current director name. Use this value to set a default name prefix.`

cmd/create.go
```diff
@@ -29,7 +29,7 @@ var componentCreateCmd = &cobra.Command{
 	Short: "Create a new component",
 	Long: `Create a new component to deploy on OpenShift.
 
-If component name is not provided, component type value will be used for the name.
+If component name is not provided, it'll be auto-generated.
```
cdrage: Argh. Sorry this wasn't updated before. But correct would be `If a component name is not provided`

docs/cli-reference.md
```diff
@@ -216,7 +217,7 @@ Catalog related operations
 
 Create a new component to deploy on OpenShift.
 
-If component name is not provided, component type value will be used for the name.
+If component name is not provided, it'll be auto-generated.
```
cdrage: `If a component name`

cmd/create.go
anmolbabu: Done

docs/cli-reference.md
anmolbabu: Ack

anmolbabu:
Merged based on above 4 acks

PR #745, opened 2018-09-17
Fix watch support check

### LINKED ISSUE #746: Watch command fails for component created using local source
User should be able to watch component created using local source and binary.
However, due to a regression [here](https://github.com/redhat-developer/odo/pull/713/files#r218104675), watch for local source is not working and errors out as under

```
▶ odo watch
Watch is supported by binary and local components only and source type of component nodejs is local
```

anmolbabu:
Fix watch support check

This PR fixes a regression which wrongly validated sourcepath against the sourcetype(local)
The PR now checks if sourceType is local or binary and allows watch to be created
 
fixes #746

### DIFF
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -51,7 +51,7 @@ var watchCmd = &cobra.Command{
 		sourceType, sourcePath, err := component.GetComponentSource(client, componentName, applicationName, projectName)
 		checkError(err, "Unable to get source for %s component.", componentName)
 
-		if sourceType != "binary" && sourcePath != "local" {
+		if sourceType != "binary" && sourceType != "local" {
 			fmt.Printf("Watch is supported by binary and local components only and source type of component %s is %s\n", componentName, sourceType)
 			os.Exit(1)
 		}


cdrage:
@kadel @anmolbabu Let's add a test for this too to prevent the error that happened before.

kadel:
@anmolbabu please link every PR to its issue.

kadel:
> @cdrage @kadel Ack
> Currently there seems to be no timeout based command execution in e2e tests.
> But watch is a command that hangs forever until user explicitly terminates the command using SIGINT..
> So, I am looking into a way to add timeout to our runCmd function once that works, will add a e2e for watch that times out after say 15 seconds before pushing one update...

A better way to write test watch would be to do run `odo watch` as separate process. you don't need any timeouts for that.
 - `odo create`
 - `odo push`
 - `odo watch` as separate  process
 -  edit something in the source of the component
 - verify that component was changed and that `odo watch` output reported that given file has been changed
 - kill `odo watch` process

PR #748, opened 2018-09-17
rename ocdev occurances to odo

surajnarwade:


pkg/config/config_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-func TestGetOcDevConfigFile(t *testing.T) {
+func TestGetOdoConfigFile(t *testing.T) {
```
kadel: we can delete this whole function as it does nothing

surajnarwade:
@kadel @syamgk updated the PR with suggestions

PR #757, opened 2018-09-18
add config parameter to modify server connection check timeout

syamgk:
add a parameter for modifying default connect check timeout 

```sh
# Example:
# for setting timeout to 9seconds
$ odo utils config set timeout 9
# for setting timeout to minimum
$ odo utils config set timeout 0
Cannot set timeout to 0 setting it to the minimum value(1sec)
$ odo utils config view 
UpdateNotification    false
Timeout               1


```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -53,10 +53,8 @@ Timeout - Timeout (in seconds) for OpenShift server connection check`,
 	Example: `
    # Set a configuration value
    odo utils config set UpdateNotification false
-   odo utils config set NamePrefix ""
    odo utils config set NamePrefix "app"
    odo utils config set timeout 20
-   odo utils config set timeout 0
 	`,
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) < 2 {
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -46,6 +46,7 @@ var configurationSetCmd = &cobra.Command{
 	Use:   "set",
 	Short: "Set a value in odo config file",
 	Long: `Set an individual value in the Odo configuration file
+
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
 NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -19,15 +19,16 @@ var configurationCmd = &cobra.Command{
 
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
-NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix`,
-	Example: fmt.Sprintf("%s\n%s",
+NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix
+Timeout - Timeout (in seconds) for OpenShift server connection check`,
+	Example: fmt.Sprintf("%s\n%s\n",
 		configurationViewCmd.Example,
 		configurationSetCmd.Example),
 	Aliases: []string{"configuration"},
 	// 'odo utils config' is the same as 'odo utils config --help'
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) >= 1 && args[0] != "view" && args[0] != "set" {
-			return fmt.Errorf("Unknown command, use set or view")
+			return fmt.Errorf(`Unknown command, use "set" or "view"`)
 		}
 		return nil
 	}, Run: func(cmd *cobra.Command, args []string) {
@@ -47,12 +48,16 @@ var configurationSetCmd = &cobra.Command{
 	Long: `Set an individual value in the Odo configuration file
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
-NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.`,
-	Example: `  # Set a configuration value
+NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.
+Timeout - Timeout (in seconds) for OpenShift server connection check`,
+	Example: `
+   # Set a configuration value
    odo utils config set UpdateNotification false
    odo utils config set NamePrefix ""
    odo utils config set NamePrefix "app"
-  `,
+   odo utils config set timeout 20
+   odo utils config set timeout 0
+	`,
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) < 2 {
 			return fmt.Errorf("Please provide a parameter name and value")
@@ -61,6 +66,7 @@ NamePrefix - Default prefix is the current directory name. Use this value to set
 		} else {
 			return nil
 		}
+
 	}, RunE: func(cmd *cobra.Command, args []string) error {
 		cfg, err := config.New()
 		if err != nil {
@@ -87,6 +93,7 @@ var configurationViewCmd = &cobra.Command{
 		fmt.Fprintln(w, "PARAMETER", "\t", "CURRENT_VALUE")
 		fmt.Fprintln(w, "UpdateNotification", "\t", cfg.GetUpdateNotification())
 		fmt.Fprintln(w, "NamePrefix", "\t", cfg.GetNamePrefix())
+		fmt.Fprintln(w, "Timeout", "\t", cfg.GetTimeout())
 		w.Flush()
 	},
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,6 +16,8 @@ import (
 const (
 	configEnvName  = "ODOCONFIG"
 	configFileName = "odo"
+	//DefaultTimeout for openshift server connection check
+	DefaultTimeout = 1
 )
 
 // OdoSettings holds all odo specific configurations
@@ -24,6 +26,8 @@ type OdoSettings struct {
 	UpdateNotification *bool `json:"updatenotification,omitempty"`
 	// Holds the prefix part of generated random application name
 	NamePrefix *string `json:"nameprefix,omitempty"`
+	// Timeout for openshift server connection check
+	Timeout *int `json:"timeout,omitempty"`
 }
 
 // ApplicationInfo holds all important information about one application
@@ -134,20 +138,33 @@ func (c *ConfigInfo) writeToFile() error {
 }
 
 // SetConfiguration modifies Odo configurations in the config file
-// as of now only being used for updatenotification
+// as of now being used for timeout, updatenotification
 func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, err := strconv.Atoi(value)
+		if err != nil {
+			return errors.Wrapf(err, "unable to set %s to %s", parameter, value)
+		}
+		if typedval < 0 {
+			return errors.Errorf("cannot set timeout to less than 0")
+		}
+		c.OdoSettings.Timeout = &typedval
+
 	case "updatenotification":
 		val, err := strconv.ParseBool(strings.ToLower(value))
 		if err != nil {
 			return errors.Wrapf(err, "unable to set %s to %s", parameter, value)
 		}
 		c.OdoSettings.UpdateNotification = &val
+
 	case "nameprefix":
 		c.OdoSettings.NamePrefix = &value
 	default:
 		return errors.Errorf("unknown parameter :'%s' is not a parameter in odo config", parameter)
 	}
+
 	err := c.writeToFile()
 	if err != nil {
 		return errors.Wrapf(err, "unable to set %s", parameter)
@@ -155,7 +172,16 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
 	return nil
 }
 
-// GetupdateNotification returns the value of UpdateNotification from config
+// GetTimeout returns the value of Timeout from config
+func (c *ConfigInfo) GetTimeout() int {
+	// default timeout value is 1
+	if c.OdoSettings.Timeout == nil {
+		return DefaultTimeout
+	}
+	return *c.OdoSettings.Timeout
+}
+
+// GetUpdateNotification returns the value of UpdateNotification from config
 func (c *ConfigInfo) GetUpdateNotification() bool {
 	if c.OdoSettings.UpdateNotification == nil {
 		return true
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -138,7 +138,7 @@ func (c *ConfigInfo) writeToFile() error {
 }
 
 // SetConfiguration modifies Odo configurations in the config file
-// as of now being used for timeout, updatenotification
+// as of now being used for nameprefix, timeout, updatenotification
 func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
 	// processing values according to the parameter names
 	switch parameter {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -140,6 +140,7 @@ func (c *ConfigInfo) writeToFile() error {
 // SetConfiguration modifies Odo configurations in the config file
 // as of now being used for timeout, updatenotification
 func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
+	// processing values according to the parameter names
 	switch parameter {
 
 	case "timeout":
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -920,6 +920,62 @@ func TestDeleteApplication(t *testing.T) {
 	}
 }
 
+func TestGetTimeout(t *testing.T) {
+	tempConfigFile, err := ioutil.TempFile("", "odoconfig")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tempConfigFile.Close()
+	os.Setenv(configEnvName, tempConfigFile.Name())
+	zeroValue := 0
+	nonzeroValue := 5
+	tests := []struct {
+		name           string
+		existingConfig Config
+		want           int
+	}{
+		{
+			name:           "validating value 1 from config in default case",
+			existingConfig: Config{},
+			want:           1,
+		},
+
+		{
+			name: "validating value 0 from config",
+			existingConfig: Config{
+				OdoSettings: OdoSettings{
+					Timeout: &zeroValue,
+				},
+			},
+			want: 0,
+		},
+
+		{
+			name: "validating value 5 from config",
+			existingConfig: Config{
+				OdoSettings: OdoSettings{
+					Timeout: &nonzeroValue,
+				},
+			},
+			want: 5,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg, err := New()
+			if err != nil {
+				t.Error(err)
+			}
+			cfg.Config = tt.existingConfig
+
+			output := cfg.GetTimeout()
+			if output != tt.want {
+				t.Errorf("GetTimeout returned unexpeced value expected \ngot: %d \nexpected: %d\n", output, tt.want)
+			}
+		})
+	}
+}
+
 func TestSetConfiguration(t *testing.T) {
 
 	tempConfigFile, err := ioutil.TempFile("", "odoconfig")
@@ -930,17 +986,21 @@ func TestSetConfiguration(t *testing.T) {
 	os.Setenv(configEnvName, tempConfigFile.Name())
 	trueValue := true
 	falseValue := false
+	zeroValue := 0
 
 	tests := []struct {
 		name           string
 		parameter      string
+		value          string
 		existingConfig Config
-		want           bool
 		wantErr        bool
+		want           interface{}
 	}{
+		// updatenotification
 		{
 			name:           "updatenotification set nil to true",
 			parameter:      "updatenotification",
+			value:          "true",
 			existingConfig: Config{},
 			want:           true,
 			wantErr:        false,
@@ -948,6 +1008,7 @@ func TestSetConfiguration(t *testing.T) {
 		{
 			name:      "updatenotification set true to false",
 			parameter: "updatenotification",
+			value:     "false",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					UpdateNotification: &trueValue,
@@ -959,6 +1020,7 @@ func TestSetConfiguration(t *testing.T) {
 		{
 			name:      "updatenotification set false to true",
 			parameter: "updatenotification",
+			value:     "true",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					UpdateNotification: &falseValue,
@@ -967,8 +1029,66 @@ func TestSetConfiguration(t *testing.T) {
 			want:    true,
 			wantErr: false,
 		},
-	}
 
+		{
+			name:           "updatenotification invalid value",
+			parameter:      "updatenotification",
+			value:          "invalid_value",
+			existingConfig: Config{},
+			wantErr:        true,
+		},
+
+		// timeout
+		{
+			name:      "Timeout set to 5 from 0",
+			parameter: "timeout",
+			value:     "5",
+			existingConfig: Config{
+				OdoSettings: OdoSettings{
+					Timeout: &zeroValue,
+				},
+			},
+			want:    5,
+			wantErr: false,
+		},
+		{
+			name:           "Timeout set to 300",
+			parameter:      "timeout",
+			value:          "300",
+			existingConfig: Config{},
+			want:           300,
+			wantErr:        false,
+		},
+		{
+			name:           "Timeout set to 0",
+			parameter:      "timeout",
+			value:          "0",
+			existingConfig: Config{},
+			want:           0,
+			wantErr:        false,
+		},
+		{
+			name:           "Timeout set to -1",
+			parameter:      "timeout",
+			value:          "-1",
+			existingConfig: Config{},
+			wantErr:        true,
+		},
+		{
+			name:           "Timeout invalid value",
+			parameter:      "timeout",
+			value:          "this",
+			existingConfig: Config{},
+			wantErr:        true,
+		},
+		// invalid parameter
+		{
+			name:           "invalid parameter",
+			parameter:      "invalid_parameter",
+			existingConfig: Config{},
+			wantErr:        true,
+		},
+	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			cfg, err := New()
@@ -977,14 +1097,34 @@ func TestSetConfiguration(t *testing.T) {
 			}
 			cfg.Config = tt.existingConfig
 
-			err = cfg.SetConfiguration(tt.parameter, strconv.FormatBool(tt.want))
+			err = cfg.SetConfiguration(tt.parameter, tt.value)
 
-			if (err != nil) != tt.wantErr {
-				t.Errorf("expected error: %v but got error %v", tt.wantErr, err)
-			}
-			// validating the value after executing Serconfiguration
-			if *cfg.OdoSettings.UpdateNotification != tt.want {
-				t.Errorf("unexpeced value after execution of SetConfiguration expected \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+			if !tt.wantErr && err == nil {
+				// validating the value after executing Serconfiguration
+				// according to component in positive cases
+				switch tt.parameter {
+				case "updatenotification":
+					if *cfg.OdoSettings.UpdateNotification != tt.want {
+						t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+					}
+				case "timeout":
+					if *cfg.OdoSettings.Timeout != tt.want {
+						t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %v \nexpected: %d\n", cfg.OdoSettings.Timeout, tt.want)
+					}
+				}
+			} else if tt.wantErr && err != nil {
+				// negative cases
+				switch tt.parameter {
+				case "updatenotification":
+				case "timeout":
+					typedval, err := strconv.Atoi(tt.value)
+					// if err is found in cases other than value <0 or !ok
+					if !(typedval < 0 || err != nil) {
+						t.Error(err)
+					}
+				}
+			} else {
+				t.Error(err)
 			}
 
 		})
@@ -1046,3 +1186,20 @@ func TestGetupdateNotification(t *testing.T) {
 		})
 	}
 }
+
+//
+//func TestGet(t *testing.T) {
+//
+//}
+//
+//func TestSet(t *testing.T) {
+//
+//}
+//
+//func TestApplicationExists(t *testing.T) {
+//
+//}
+//
+//func TestAddApplication(t *testing.T) {
+//
+//}
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -935,13 +935,13 @@ func TestGetTimeout(t *testing.T) {
 		want           int
 	}{
 		{
-			name:           "validating value 1 from config in default case",
+			name:           "Case 1: validating value 1 from config in default case",
 			existingConfig: Config{},
 			want:           1,
 		},
 
 		{
-			name: "validating value 0 from config",
+			name: "Case 2: validating value 0 from config",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					Timeout: &zeroValue,
@@ -951,7 +951,7 @@ func TestGetTimeout(t *testing.T) {
 		},
 
 		{
-			name: "validating value 5 from config",
+			name: "Case 3: validating value 5 from config",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					Timeout: &nonzeroValue,
@@ -998,7 +998,7 @@ func TestSetConfiguration(t *testing.T) {
 	}{
 		// updatenotification
 		{
-			name:           "updatenotification set nil to true",
+			name:           "Case 1: updatenotification set nil to true",
 			parameter:      "updatenotification",
 			value:          "true",
 			existingConfig: Config{},
@@ -1006,7 +1006,7 @@ func TestSetConfiguration(t *testing.T) {
 			wantErr:        false,
 		},
 		{
-			name:      "updatenotification set true to false",
+			name:      "Case 2: updatenotification set true to false",
 			parameter: "updatenotification",
 			value:     "false",
 			existingConfig: Config{
@@ -1018,7 +1018,7 @@ func TestSetConfiguration(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name:      "updatenotification set false to true",
+			name:      "Case 3: updatenotification set false to true",
 			parameter: "updatenotification",
 			value:     "true",
 			existingConfig: Config{
@@ -1031,7 +1031,7 @@ func TestSetConfiguration(t *testing.T) {
 		},
 
 		{
-			name:           "updatenotification invalid value",
+			name:           "Case 4: updatenotification invalid value",
 			parameter:      "updatenotification",
 			value:          "invalid_value",
 			existingConfig: Config{},
@@ -1040,7 +1040,7 @@ func TestSetConfiguration(t *testing.T) {
 
 		// timeout
 		{
-			name:      "Timeout set to 5 from 0",
+			name:      "Case 5: Timeout set to 5 from 0",
 			parameter: "timeout",
 			value:     "5",
 			existingConfig: Config{
@@ -1052,7 +1052,7 @@ func TestSetConfiguration(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name:           "Timeout set to 300",
+			name:           "Case 6: Timeout set to 300",
 			parameter:      "timeout",
 			value:          "300",
 			existingConfig: Config{},
@@ -1060,7 +1060,7 @@ func TestSetConfiguration(t *testing.T) {
 			wantErr:        false,
 		},
 		{
-			name:           "Timeout set to 0",
+			name:           "Case 7: Timeout set to 0",
 			parameter:      "timeout",
 			value:          "0",
 			existingConfig: Config{},
@@ -1068,14 +1068,14 @@ func TestSetConfiguration(t *testing.T) {
 			wantErr:        false,
 		},
 		{
-			name:           "Timeout set to -1",
+			name:           "Case 8: Timeout set to -1",
 			parameter:      "timeout",
 			value:          "-1",
 			existingConfig: Config{},
 			wantErr:        true,
 		},
 		{
-			name:           "Timeout invalid value",
+			name:           "Case 9: Timeout invalid value",
 			parameter:      "timeout",
 			value:          "this",
 			existingConfig: Config{},
@@ -1083,7 +1083,7 @@ func TestSetConfiguration(t *testing.T) {
 		},
 		// invalid parameter
 		{
-			name:           "invalid parameter",
+			name:           "Case 10: invalid parameter",
 			parameter:      "invalid_parameter",
 			existingConfig: Config{},
 			wantErr:        true,
@@ -1148,12 +1148,12 @@ func TestGetupdateNotification(t *testing.T) {
 		want           bool
 	}{
 		{
-			name:           "updatenotification nil",
+			name:           "Case 1: updatenotification nil",
 			existingConfig: Config{},
 			want:           true,
 		},
 		{
-			name: "updatenotification true",
+			name: "Case 2: updatenotification true",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					UpdateNotification: &trueValue,
@@ -1162,7 +1162,7 @@ func TestGetupdateNotification(t *testing.T) {
 			want: true,
 		},
 		{
-			name: "updatenotification false",
+			name: "Case 3: updatenotification false",
 			existingConfig: Config{
 				OdoSettings: OdoSettings{
 					UpdateNotification: &falseValue,
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1186,20 +1186,3 @@ func TestGetupdateNotification(t *testing.T) {
 		})
 	}
 }
-
-//
-//func TestGet(t *testing.T) {
-//
-//}
-//
-//func TestSet(t *testing.T) {
-//
-//}
-//
-//func TestApplicationExists(t *testing.T) {
-//
-//}
-//
-//func TestAddApplication(t *testing.T) {
-//
-//}
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -19,12 +19,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 
-	"github.com/redhat-developer/odo/pkg/util"
-
 	"github.com/fatih/color"
 	"github.com/golang/glog"
 	dockerapiv10 "github.com/openshift/api/image/docker10"
 	"github.com/pkg/errors"
+	"github.com/redhat-developer/odo/pkg/config"
+	"github.com/redhat-developer/odo/pkg/util"
 
 	servicecatalogclienset "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset/typed/servicecatalog/v1beta1"
 	appsschema "github.com/openshift/client-go/apps/clientset/versioned/scheme"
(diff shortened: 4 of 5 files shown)


pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
 }
 
 // SetConfiguration modifies Odo configurations in the config file
-// as of now only being used for updatenotification
-func (c *ConfigInfo) SetConfiguration(parameter string, value bool) error {
+// as of now being used fot timeout, updatenotification
```
mik-dass: `for timeout`

mik-dass:
@syamgk Please fix the first comment 

>  for setting timeout to 9seconds
> $ odo config set timeout 9

It should be odo utils config

cmd/config.go
```diff
@@ -46,7 +47,8 @@ var configurationSetCmd = &cobra.Command{
 	Short: "Set a value in odo config file",
 	Long: `Set an individual value in the Odo configuration file 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false)
+Timeout            - timeout(in seconds) for openshift server connection check(1 to 99)`,
 	Example: `
    # Set a configuration value
```
mik-dass: It might be a good idea to add an example for the timeout too and separate them

mik-dass:
@syamgk I think it is better to use `Run` instead of `RunE` here https://github.com/redhat-developer/odo/blob/3c89a1b4b0deb62da5bd98ba597aee4af885eff7/cmd/config.go#L62
```
[mrinaldas@localhost odo]$ ./odo utils config set timeot 5
Error: unknown parameter :'timeot' is not a parameter in odo config
Usage:
  odo utils config set [flags]

Examples:

   # Set a configuration value
        

Flags:
  -h, --help   help for set

Global Flags:

unknown parameter :'timeot' is not a parameter in odo config
```
Currently `unknown parameter :'timeot' is not a parameter in odo config` is getting shown twice. I think we should show a simple message instead of the complete help message.

syamgk:
> @syamgk I think it is better to use `Run` instead of `RunE` here
since this part of code has lots of error checks like `cfg, err := config.New()`
I would like to use `RunE`
> 
> [odo/cmd/config.go](https://github.com/redhat-developer/odo/blob/3c89a1b4b0deb62da5bd98ba597aee4af885eff7/cmd/config.go#L62)
> 
> Line 62 in [3c89a1b](/redhat-developer/odo/commit/3c89a1b4b0deb62da5bd98ba597aee4af885eff7)
>  }, RunE: func(cmd *cobra.Command, args []string) error { 
> 
> ```
> [mrinaldas@localhost odo]$ ./odo utils config set timeot 5
> Error: unknown parameter :'timeot' is not a parameter in odo config
> Usage:
>   odo utils config set [flags]
> 
> Examples:
> 
>    # Set a configuration value
>    odo utils config set UpdateNotification false
>         
> 
> Flags:
>   -h, --help   help for set
> 
> Global Flags:
>       --alsologtostderr                  log to standard error as well as files
>       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
>       --log_dir string                   If non-empty, write log files in this directory
>       --logtostderr                      log to standard error instead of files (default false)
>       --skip-connection-check            Skip cluster check
>       --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
>   -v, --v Level                          log level for V logs. Level varies from 0 to 9 (default 0).
>       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
> 
> unknown parameter :'timeot' is not a parameter in odo config
> ```
> 
> Currently `unknown parameter :'timeot' is not a parameter in odo config` is getting shown twice. I think we should show a simple message instead of the complete help message.

mik-dass:
@syamgk Ok it's fine to use RunE then but please remove the complete help message which is displayed incase of invalid parameter.

syamgk:
@mik-dass I changed all messages to printf messages and changed to Run.

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
 		return errors.Wrapf(err, "unable to set %s", parameter)
 	}
 	return nil
 }
 
+// GetTimeout returns the value of Timeout from config
+func (c *ConfigInfo) GetTimeout() int {
+	// minimum timeout value is 1
+	if c.OdoSettings.Timeout == 0 {
+		return 1
```
anmolbabu: Just a curious question, why don’t we set these with defaults in config file ? Is it to avoid the risk of user accidentally deleting it (and then odo fails?)

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
-// as of now only being used for updatenotification
-func (c *ConfigInfo) SetConfiguration(parameter string, value bool) error {
+// as of now being used for timeout, updatenotification
+func (c *ConfigInfo) SetConfiguration(parameter string, value interface{}) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, ok := value.(int)
+		if typedval == 0 {
+			fmt.Println("Cannot set timeout to 0 setting it to the minimum value(1sec)")
```
anmolbabu: I am of the opinion that we should rather error out here instead of taking decision on behalf of the user..

Also, this would be nice to have in the cmd help

pkg/config/config_test.go
```diff
@@ -976,11 +1005,21 @@ func TestSetConfiguration(t *testing.T) {
...
-				t.Errorf("unexpeced value after execution of SetConfiguration expected \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+			// according to component
+			switch tt.parameter {
+			case "updatenotification":
+				if *cfg.OdoSettings.UpdateNotification != tt.want {
+					t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+				}
+			case "timeout":
+				if cfg.OdoSettings.Timeout != tt.want {
+					t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %d \nexpected: %d\n", cfg.OdoSettings.Timeout, tt.want)
```
anmolbabu: No need to type check?

pkg/config/config.go
syamgk: I prefer default value coming from code instead of config file, hence this

anmolbabu: hmm, both have their own advantages and disadvantages.. but I tend to bend in favor of having it in config file probably because of the following reasons:
* it does not involve any changes to the code if the defaults were to ever change or even their values to non-default ones intentionally. Because code changes involve that whole build-release-update cycle
* exposes to the user the list of possible configurables by way of just viewing the config file or even a command that can simply read the config file and dump it to the cmd o/p
* reduces hard coding in the code which is in my opinion very important

but comes with some side-effects ;) like:
* what if user accidentally deletes the config file
* what if user screws up the config file if he manually edits it

Now the above side-effects can be worked around with some overhead(one such way could be don't allow the user to modify the config file directly by creating odo user as part of odo installation give ownership of config file to only odo user but I agree that only makes our current installation complex. But its easily achievable by way of writing an rpm spec and generating rpm out of it(which we don't do) so this brings me to the next question -- why not?) Expose a command to also list config entries in config file.

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
 // SetConfiguration modifies Odo configurations in the config file
-// as of now only being used for updatenotification
-func (c *ConfigInfo) SetConfiguration(parameter string, value bool) error {
+// as of now being used for timeout, updatenotification
+func (c *ConfigInfo) SetConfiguration(parameter string, value interface{}) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, ok := value.(int)
+		if typedval == 0 {
```
anmolbabu: I think we should check error first

Ref: https://play.golang.org/p/YwFcA85ekHW

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
+			fmt.Println("Cannot set timeout to 0, minimum value is 1")
+			return errors.Errorf("type assertion error: unable to set")
```
anmolbabu: I think error should be: "Cannot set timeout to 0, minimum value is 1" instead of typeassertion error

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
-// as of now only being used for updatenotification
-func (c *ConfigInfo) SetConfiguration(parameter string, value bool) error {
+// as of now being used for timeout, updatenotification
+func (c *ConfigInfo) SetConfiguration(parameter string, value interface{}) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, ok := value.(int)
+		if typedval == 0 {
+			fmt.Println("Cannot set timeout to 0, minimum value is 1")
```
anmolbabu: just a debug msg?
we are anyway returning the error can we remove this?

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
+			return errors.Errorf("type assertion error: unable to set")
+		}
+		if ok != true {
+			return errors.Errorf("type assertion error: unable to set %s", parameter)
+		}
+		c.OdoSettings.Timeout = typedval
```
anmolbabu: Anyone that wants config would do config.New which reads a fresh from config file on every call right?
why is this required?

pkg/config/config.go
```diff
@@ -130,21 +132,48 @@ func (c *ConfigInfo) writeToFile() error {
...
 	err := c.writeToFile()
 	if err != nil {
 		return errors.Wrapf(err, "unable to set %s", parameter)
 	}
 	return nil
 }
 
+// GetTimeout returns the value of Timeout from config
+func (c *ConfigInfo) GetTimeout() int {
+	// minimum timeout value is 1
```
anmolbabu: can we say default timeout instead of minimum ?

surajnarwade:
Identation is little bit weird,

```
$ odo utils 
Utilities for completion, terminal commands and modifying Odo configurations

Usage:
  odo utils [command]

Examples:
  # Bash autocompletion support
  source <(odo utils completion bash)

  # Zsh autocompletion support
  source <(odo utils completion zsh)

  # Bash terminal PS1 support
  source <(odo utils terminal bash)

  # Zsh terminal PS1 support
  source <(odo utils terminal zsh)


   # Set UpdateNotification off

   # Set openshift server connection check to 20seconds
   odo utils config set timeout 20
	
   # For viewing the current configuration
   odo utils config view
```

@syamgk, can you please check once ?

pkg/config/config.go
syamgk: in both ways we have to hardcode the minimum value, in code

pkg/config/config.go
syamgk: ack

pkg/config/config.go
syamgk: ack

pkg/config/config.go
syamgk: error will only be returned only if I use `RunE`, but while using `RunE` it will print the entire help msg as well as mentioned https://github.com/redhat-developer/odo/pull/757#issuecomment-423912589

anmolbabu: Sorry not sure, I follow...
We are in a function inside the helper func to the cobra command right which can return anything(even an error) right?
How does RunE come into picture here?

pkg/config/config.go
anmolbabu: how/why in code ?

pkg/config/config_test.go
```diff
@@ -972,11 +1005,21 @@ func TestSetConfiguration(t *testing.T) {
 			}
 			cfg.Config = tt.existingConfig
 
-			cfg.SetConfiguration(tt.parameter, tt.want)
+			cfg.SetConfiguration(tt.parameter, tt.value)
```
golangcibot: Error return value of `cfg.SetConfiguration` is not checked

pkg/config/config.go
```diff
@@ -130,21 +132,47 @@ func (c *ConfigInfo) writeToFile() error {
...
 // SetConfiguration modifies Odo configurations in the config file
-// as of now only being used for updatenotification
-func (c *ConfigInfo) SetConfiguration(parameter string, value bool) error {
+// as of now being used for timeout, updatenotification
+func (c *ConfigInfo) SetConfiguration(parameter string, value interface{}) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, ok := value.(int)
+		if ok != true {
```
golangcibot: should omit comparison to bool constant, can be simplified to !ok

pkg/config/config.go
```diff
@@ -130,21 +132,47 @@ func (c *ConfigInfo) writeToFile() error {
...
+		}
+		if typedval == 0 {
+			return errors.Errorf("cannot set timeout to 0, minimum value is 1")
+		}
+		c.OdoSettings.Timeout = typedval
+
 	case "updatenotification":
-		c.OdoSettings.UpdateNotification = &value
+		typedval, ok := value.(bool)
+		if ok != true {
```
golangcibot: should omit comparison to bool constant, can be simplified to !ok

syamgk:
@surajnarwade  yep there is and now made changes to align them :)

pkg/config/config.go
syamgk: towards the end of this function it writes into the config file and the next time it reads from there

syamgk: also cfg.New() is reading from config file if file exists

syamgk: https://github.com/redhat-developer/odo/blob/master/pkg/config/config.go#L75

syamgk: > Anyone that wants config would do config.New which reads a fresh from config file on every call right?
nope, if config file exist then it reads the existing config file 
https://github.com/redhat-developer/odo/blob/master/pkg/config/config.go#L75
> why is this required?
at the end of this function it gets written into the config file

anmolbabu: > towards the end of this function it writes into the config file and the next time it reads from there

Ok now makes sense

pkg/config/config.go
syamgk: > how/why in code ?

see even if we look into the linux code, the defults are either constants or Macros, 

> 
> Because default config would be in code even if it is a separate file? But think about it this way, the changes to defaults will be localised to one file + any such change will not necessarily warrant an urgent release.. because it can always be worked around by the user by just a change in his version of the config file

I dont know how are we going to alter an already existing default value in a config file,
in fact we can change it easily if the defaults are coming from code

surajnarwade:
Tried, worked for me :tada: 
LGTM :)

pkg/config/config.go
anmolbabu: > > how/why in code ?
> 
> see even if we look into the linux code, the defults are either constants or Macros,
> 

hmm ...

> > Because default config would be in code even if it is a separate file? But think about it this way, the changes to defaults will be localised to one file + any such change will not necessarily warrant an urgent release.. because it can always be worked around by the user by just a change in his version of the config file
> 
> I dont know how are we going to alter an already existing default value in a config file,
> in fact we can change it easily if the defaults are coming from code

Just like you edit any other value in the config file :) How is it any different? (It is only that the config file, in case of my comment being incoporated, will always have an entry immaterial of whether the user chooses to override it or not... and the code is free from any hard-coded default values or atleast concentrated in an expected single source file..) The code just picks the values from the config file without making any assumptions whether the value is default or user chosen ...

Or at the least I would be atleast happy if the defaults we maintain currently alongside the code, be atleast moved to a separate file which would be intended to serve as the single point of updates to any configurable default values and/or constants... without which, as the code matures, its going to be a challenging task, if we maintain the constants in many places in the code...

Feel free to ignore the comment if it is the only one blocking the PR as this is probably not even the intention of the PR... and we can take it up whenever appropriate in the future... I am ok with it...

cmd/config.go
```diff
@@ -19,7 +18,8 @@ var configurationCmd = &cobra.Command{
 	Long: `Modifies Odo specific configuration settings within the config file.
 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false),
+Timeout            - timeout(in seconds) for openshift server connection check(minimum value is 1second)`,
```
kadel: missing space after `1`

pkg/config/config.go
kadel: It would probably make sense to have this as const https://github.com/redhat-developer/odo/blob/d48d812a6bae054b27cf375ab772a0f1801cb1a7/pkg/config/config.go#L14

Would that solve it @anmolbabu ? I'm not sure what config file you mean :-(

cmd/config.go
kadel: Start with Capital letter to match style of other messages

cmd/config.go
```diff
@@ -46,10 +46,14 @@ var configurationSetCmd = &cobra.Command{
 	Short: "Set a value in odo config file",
 	Long: `Set an individual value in the Odo configuration file 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false)
+Timeout            - timeout(in seconds) for openshift server connection check`,
```
kadel: Sart with Capital letter to match style

cmd/config.go
```diff
@@ -46,10 +46,14 @@ var configurationSetCmd = &cobra.Command{
...
 	Example: `
-   # Set a configuration value
-   odo utils config set UpdateNotification false
+  # Set UpdateNotification off
+  odo utils config set UpdateNotification false
+
+  # Set openshift server connection check to 20seconds
```
kadel: space after number

pkg/config/config.go
anmolbabu: @kadel I meant the config file as in $HOME/.kube/odo..

So, timeout is anyway a configurable field and all I meant is just moving any configurable value out of code and into a default config file that is packaged with code but is simply copied to the required config file location may be as part of config init in our code.. That way, the default values for configurables are always isolated to a separate file and restoring them is also not a difficult task... And whenever listing of tunables is required we simply read the config file and put it forth(as the config file would then contain all tunables with wither defaults or user tuned values) and there are not 2 separate lists maintained its just  1 single tunable list... Currently I see that, a config value doesn't appear in config file until user has not tuned it and the defaults come from code regions that can be spread around...

It would atleast be nice to isolate it(default values of tunables and any constants related to them) to constants in separate file.. But, I am ok with anything I just wanted to understand why it can't be isolated to separate file and any configurable value tuned/default appears in that separate file and can be located easily in a single place that affects/alters everywhere else that the constant is/needs to be used...

kadel: > @kadel I meant the config file as in $HOME/.kube/odo..

what if the user is running odo for the first time and there is no config, or if the value is not specified in the config? What value should be used then? Some default value will have to be in the code.

anmolbabu: @kadel The config file with default values(as I propose now) will be copied to the config file location at the very first odo command invocation(or exactly where the config file is created currently it would be)... The only difference is that the separate file with all tunables with their default values, maintained as part of code will be now simply moved/copied to the config directory...

anmolbabu: > > @kadel I meant the config file as in $HOME/.kube/odo..
> 
> what if the user is running odo for the first time and there is no config, or if the value is not specified in the config? What value should be used then? Some default value will have to be in the code.

@kadel, I agree even with my current proposal the config is maintained as part of code base only but my point is only its isolation into a separate file(only to avoid the defaults spreading to more than 1 regions of code) and such that any tunable is always present in config file...

kadel: @anmolbabu so what do you propose?
Moving this to something like a separate package that will just contain constants like https://github.com/redhat-developer/odo/blob/d48d812a6bae054b27cf375ab772a0f1801cb1a7/pkg/config/config.go#L15 and import it here?

It will make those values public, even if they don't have to be :-(

anmolbabu: > @anmolbabu so what do you propose?
> Moving this to something like a separate package that will just contain constants like
> 
> [odo/pkg/config/config.go](https://github.com/redhat-developer/odo/blob/d48d812a6bae054b27cf375ab772a0f1801cb1a7/pkg/config/config.go#L15)
> 
> Line 15 in [d48d812](/redhat-developer/odo/commit/d48d812a6bae054b27cf375ab772a0f1801cb1a7)
> 
>  configEnvName  = "ODOCONFIG" 
> and import it here?
> It will make those values public, even if they don't have to be :-(

@kadel My proposal is to create a file called odo/odo.conf(name only for indication can be anything more appropriate) and have the default config in it with all tunables and with their respective default values... As part of https://github.com/redhat-developer/odo/blob/master/pkg/config/config.go#L75 copy odo/odo.conf to config path($HOME/.kube/odo) if it does not currently exist... and from any region of code apart from odo/odo.conf remove any default value for any config.. 

But I agree its something not really the aim of this PR(and I really don't want to block this PR for this). So, in my opinion, we can track this as part of a different issue based on its priority but all I wanted is, to just bring this up and learn of any reason/s that didn't allow us to do this... You see I love Odo and don't want to stay away from any info/perspective that takes me even an inch closer to Odo... :)

kadel: It makes sense @anmolbabu, can you please open a new issue and describe your proposal there?

anmolbabu: Ack @kadel

tests/e2e/e2e_test.go
```diff
@@ -70,6 +70,10 @@ var _ = Describe("odoe2e", func() {
 			Expect(config).To(ContainSubstring("true"))
 			Expect(config).To(ContainSubstring("UpdateNotification"))
 		})
+		It("should get 1 for timeout by default", func() {
+			config := runCmd("odo utils config view|grep Timeout")
+			Expect(config).To(ContainSubstring("1"))
```
cdrage: We should have the constant here instead of `1`.

tests/e2e/e2e_test.go
```diff
@@ -70,6 +70,10 @@ var _ = Describe("odoe2e", func() {
 			Expect(config).To(ContainSubstring("true"))
 			Expect(config).To(ContainSubstring("UpdateNotification"))
 		})
+		It("should get 1 for timeout by default", func() {
```
cdrage: Should be checking to see if timeout is the same as the constant.

pkg/config/config.go
anmolbabu: From my conversation with @kadel it seems distributing binary is the most common(and convenient) way of distributing go packages and packaging conf text files with binary seems not cleanly supported by golang making my above comments practically not viable at the moment. So, until we find a better way, we probably will need to live with the way it is currently...

mik-dass:
Sorry one last thing to fix
```
[mrinaldas@localhost odo]$ ./odo utils config set timeout -1
Error: unknown shorthand flag: '1' in -1
Usage:
  odo utils config set [flags]

Examples:

  # Set UpdateNotification off

  # Set OpenShift server connection check to 20 seconds
  odo utils config set timeout 20
        

Flags:
  -h, --help   help for set

Global Flags:

unknown shorthand flag: '1' in -1
```
Please take the -1 as a input, instead of a flag, and display a error message like 
```
[mrinaldas@localhost odo]$ oc get pods --request-timeout -1
error: Invalid timeout value. Timeout must be a single integer in seconds, or an integer followed by a corresponding time unit (e.g. 1s | 2m | 3h)
```

syamgk:
> Sorry one last thing to fix
> 
> ```
> [mrinaldas@localhost odo]$ ./odo utils config set timeout -1
> Error: unknown shorthand flag: '1' in -1
> Usage:
>   odo utils config set [flags]
> 
> Examples:
> 
>   # Set UpdateNotification off
>   odo utils config set UpdateNotification false
> 
>   # Set OpenShift server connection check to 20 seconds
>   odo utils config set timeout 20
>         
> 
> Flags:
>   -h, --help   help for set
> 
> Global Flags:
>       --alsologtostderr                  log to standard error as well as files
>       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
>       --log_dir string                   If non-empty, write log files in this directory
>       --logtostderr                      log to standard error instead of files (default false)
>       --skip-connection-check            Skip cluster check
>       --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
>   -v, --v Level                          log level for V logs. Level varies from 0 to 9 (default 0).
>       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
> 
> unknown shorthand flag: '1' in -1
> ```
> 
> Please take the -1 as a input, instead of a flag, and display a error message like
> 
> ```
> [mrinaldas@localhost odo]$ oc get pods --request-timeout -1
> error: Invalid timeout value. Timeout must be a single integer in seconds, or an integer followed by a corresponding time unit (e.g. 1s | 2m | 3h)
> ```

there is an upstream issue https://github.com/spf13/cobra/issues/124
but now it behaves like this 
```
$ ./odo utils config set timeout -- -1
cannot set timeout to less than 1 second

```

pkg/config/config_test.go
```diff
@@ -933,18 +933,21 @@ func TestSetConfiguration(t *testing.T) {
 	tests := []struct {
 		name           string
 		parameter      string
+		value          interface{}
 		existingConfig Config
-		want           bool
+		want           interface{}
```
mik-dass: Why is there no wantErr in this tests? The function seems to throw error in case of case error and others

tests/e2e/e2e_test.go
```diff
@@ -70,6 +71,10 @@ var _ = Describe("odoe2e", func() {
 			Expect(config).To(ContainSubstring("true"))
 			Expect(config).To(ContainSubstring("UpdateNotification"))
 		})
+		It("should be checking to see if timeout is the same as the constant", func() {
+			config := runCmd("odo utils config view|grep Timeout")
```
kadel: you are importing package that is named `config` so this variable conflicts with package name

kadel:
> there is an upstream issue [spf13/cobra#124](https://github.com/spf13/cobra/issues/124)
> but now it behaves like this

from @mik-dass example it looks like `oc` has solution for it, have you checked how they do it?

pkg/config/config_test.go
syamgk: now add -ve cases as well as wantErr

syamgk:
> > there is an upstream issue [spf13/cobra#124](https://github.com/spf13/cobra/issues/124)
> > but now it behaves like this
> 
> from @mik-dass example it looks like `oc` has solution for it, have you checked how they do it?

they are doing it for flag, see even odo has `odo version --v 9`
It is possible if we can do config like `odo utils config set --timeout -1`

pkg/config/config_test.go
```diff
@@ -972,11 +1035,32 @@ func TestSetConfiguration(t *testing.T) {
...
+					if cfg.OdoSettings.Timeout != tt.want {
+						t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %v \nexpected: %d\n", cfg.OdoSettings.Timeout, tt.want)
+					}
+				}
+			} else if tt.wantErr == true && err != nil {
+				// negative cases
+				switch tt.parameter {
+				case "updatenotification":
+				case "timeout":
+					typedval, ok := tt.value.(int)
```
golangcibot: this value of `ok` is never used

pkg/config/config_test.go
```diff
@@ -972,11 +1035,32 @@ func TestSetConfiguration(t *testing.T) {
...
+					if typedval <= 0 || !ok {
```
golangcibot: empty branch

pkg/config/config_test.go
```diff
@@ -972,11 +1035,32 @@ func TestSetConfiguration(t *testing.T) {
 			}
 			cfg.Config = tt.existingConfig
 
-			cfg.SetConfiguration(tt.parameter, tt.want)
+			err = cfg.SetConfiguration(tt.parameter, tt.value)
 
-			// validating the value after executing Serconfiguration
-			if *cfg.OdoSettings.UpdateNotification != tt.want {
-				t.Errorf("unexpeced value after execution of SetConfiguration expected \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+			if tt.wantErr == false && err == nil {
```
golangcibot: should omit comparison to bool constant, can be simplified to !tt.wantErr

pkg/config/config_test.go
```diff
@@ -972,11 +1035,32 @@ func TestSetConfiguration(t *testing.T) {
...
+				case "updatenotification":
+					if *cfg.OdoSettings.UpdateNotification != tt.want {
+						t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %t \nexpected: %t\n", *cfg.OdoSettings.UpdateNotification, tt.want)
+					}
+				case "timeout":
+					if cfg.OdoSettings.Timeout != tt.want {
+						t.Errorf("unexpeced value after execution of SetConfiguration \ngot: %v \nexpected: %d\n", cfg.OdoSettings.Timeout, tt.want)
+					}
+				}
+			} else if tt.wantErr == true && err != nil {
```
golangcibot: should omit comparison to bool constant, can be simplified to tt.wantErr

pkg/config/config_test.go
mik-dass: Maybe a -ve case also for extra surity?

cmd/config.go
```diff
@@ -19,15 +18,16 @@ var configurationCmd = &cobra.Command{
 	Long: `Modifies Odo specific configuration settings within the config file.
 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false),
+Timeout            - Timeout (in seconds) for openshift server connection check (minimum value is 1 second)`,
```
surajnarwade: according to `oc` command line flag,

```
--request-timeout='0': The length of time to wait before giving up on a single server request.
Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means
don't timeout requests.
```

so, we should be able to set  `0` as well so that it `don't timeout requests.`

cmd/config.go
```diff
@@ -19,15 +18,16 @@ var configurationCmd = &cobra.Command{
 	Long: `Modifies Odo specific configuration settings within the config file.
 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false),
+Timeout            - Timeout (in seconds) for openshift server connection check, A value of zero means don't timeout requests`,
```
cdrage: OpenShift

cdrage: Remove the ` A value of zero means don't timeout requests` that doesn't seem needed.

cmd/config.go
```diff
@@ -19,15 +18,16 @@ var configurationCmd = &cobra.Command{
...
 	Example: fmt.Sprintf("%s\n%s\n",
 		configurationViewCmd.Example,
 		configurationSetCmd.Example),
 	Aliases: []string{"configuration"},
 	// 'odo utils config' is the same as 'odo utils config --help'
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) >= 1 && args[0] != "view" && args[0] != "set" {
-			return fmt.Errorf("Unknown command, use set or view")
+			return fmt.Errorf("Unknown command, use \"set\" or \"view\"")
```
cdrage: Instead of having to use `\"`

```sh
return fmt.Errorf(`Unknown command, use "set" or "view"`)
```

cmd/config.go
```diff
@@ -46,29 +46,41 @@ var configurationSetCmd = &cobra.Command{
 	Short: "Set a value in odo config file",
 	Long: `Set an individual value in the Odo configuration file 
 Available Parameters:
-UpdateNotification - Controls if an update notification is shown or not (true or false)`,
+UpdateNotification - Controls if an update notification is shown or not (true or false)
+Timeout            - Timeout (in seconds) for openshift server connection check`,
```
cdrage: This isn't the same output as https://github.com/redhat-developer/odo/pull/757/files#diff-31f9f805fb3e45c114bdaa74f9447759R22

Please update both. Also openshift = OpenShift!

pkg/config/config.go
```diff
@@ -130,21 +134,47 @@ func (c *ConfigInfo) writeToFile() error {
...
+func (c *ConfigInfo) SetConfiguration(parameter string, value interface{}) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, ok := value.(int)
+		if !ok {
+			return errors.Errorf("type assertion error: unable to set %s", parameter)
+		}
+		if typedval < 0 {
+			return errors.Errorf("cannot set timeout to less than 0, A value of zero means don't timeout requests")
```
cdrage: remove `, A value of zero means don't timeout requests` this isn't really needed.

cmd/config.go
syamgk: Moved that information under Examples

pkg/config/config_test.go
mik-dass: Why is this a empty branch? Maybe 

```Suggestion
if typedval <= 0 || ok {
      t.Error(err)
}
```

cmd/config.go
```diff
@@ -47,13 +47,16 @@ var configurationSetCmd = &cobra.Command{
...
-Timeout            - Timeout (in seconds) for openshift server connection check`,
+Timeout            - Timeout (in seconds) for OpenShift server connection check`,
 	Example: `
   # Set UpdateNotification off
   odo utils config set UpdateNotification false
 
   # Set OpenShift server connection check to 20 seconds
   odo utils config set timeout 20
+
+  # Set 0 to don't timeout requests for OpenShift server check
```
cdrage: Weird wording..

Do we really need this example? The examples above seem like enough.

cmd/config.go
```diff
@@ -46,29 +46,44 @@ var configurationSetCmd = &cobra.Command{
...
+  # Set OpenShift server connection check to 20 seconds
+  odo utils config set timeout 20
+
+  # Set 0 to don't timeout requests for OpenShift server check
+  odo utils config set timeout 0
 	`,
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) < 2 {
-			return fmt.Errorf("Please provide a parameter name and value")
+			checkError(fmt.Errorf("Please provide a parameter name and value"), "")
```
cdrage: Why are we using checkError for this? Seems odd in comparison to using return fmt.Errorf ...

pkg/occlient/occlient.go
```diff
@@ -256,8 +257,17 @@ func isServerUp(server string) bool {
 		return false
 	}
 
+	timeout := ocRequestTimeout
```
cdrage: The only instance of `ocRequestTimeout` is being used here.. Why not remove the `ocRequestTimeout` constant completely? Then you can rename `timeout` to `ocRequestTimeout` and use it here.

pkg/config/config_test.go
syamgk: rephrased

cmd/config.go
syamgk: thought it will be nice to mention about 0 and its special behavior in help

pkg/occlient/occlient.go
syamgk: done

pkg/config/config.go
```diff
@@ -130,22 +134,48 @@ func (c *ConfigInfo) writeToFile() error {
...
 	}
 	return nil
 }
 
-// GetupdateNotification returns the value of UpdateNotification from config
+// GetTimeout returns the value of Timeout from config
+func (c *ConfigInfo) GetTimeout() int {
+	// default timeout value is 1
+	if c.OdoSettings.Timeout == 0 {
+		return DefaultTimeout
```
mik-dass: why is it returning 1 when the user sets the timeout to 0?
```
[mrinaldas@localhost odo]$ ./odo utils config set timeout 0
[mrinaldas@localhost odo]$ ./odo utils config view
UpdateNotification    true
Timeout               1
```

syamgk: this how the value has been tackled for in the case of parameter not present in `.kube/odo` file

syamgk: but now since we have 0 for no timeout i should find a better way to do same

syamgk: Changed `OdoSettings.Timeout` to int pointer and now it will distinguish 0, nill and all non zeros.
@mik-dass Try now it will work as expected

mik-dass: :+1:

tests/e2e/e2e_test.go
```diff
@@ -66,9 +67,13 @@ var _ = Describe("odoe2e", func() {
...
-			config := runCmd("odo utils config view")
-			Expect(config).To(ContainSubstring("true"))
-			Expect(config).To(ContainSubstring("UpdateNotification"))
+			config_output := runCmd("odo utils config view")
+			Expect(config_output).To(ContainSubstring("true"))
+			Expect(config_output).To(ContainSubstring("UpdateNotification"))
+		})
+		It("should be checking to see if timeout is the same as the constant", func() {
+			config_output := runCmd("odo utils config view|grep Timeout")
+			Expect(config_output).To(ContainSubstring(fmt.Sprintf("%d", config.DefaultTimeout)))
```
mik-dass: Please add a test for setting both the timeout and updatenotification cases and verifying the values in both the cases.

tests/e2e/e2e_test.go
```diff
@@ -66,9 +67,22 @@ var _ = Describe("odoe2e", func() {
...
+		})
+		It("should be checking to see if config values are the same as the configured ones", func() {
+			runCmd("odo utils config set updatenotification false")
+			runCmd("odo utils config set timeout 5")
+			config_output := runCmd("odo utils config view|grep UpdateNotification")
+			Expect(config_output).To(ContainSubstring("true"))
```
mik-dass: this should be 
```Suggestion
			Expect(config_output).To(ContainSubstring("false"))
```
 The e2e tests are failing because of this

syamgk: ack, fixed

syamgk:
re-based again!

cmd/config.go
syamgk: its also added

syamgk:
resolved conflicts and squashed the commits
its waiting on review @cdrage @surajnarwade @mik-dass

pkg/config/config.go
```diff
@@ -134,28 +138,50 @@ func (c *ConfigInfo) writeToFile() error {
 }
 
 // SetConfiguration modifies Odo configurations in the config file
-// as of now only being used for updatenotification
+// as of now being used for timeout, updatenotification
```
mik-dass: Also and `nameprefix`

pkg/config/config_test.go
```diff
@@ -1046,3 +1186,20 @@ func TestGetupdateNotification(t *testing.T) {
 		})
 	}
 }
+
+//
```
mik-dass: Please clean up these lines

pkg/occlient/occlient.go
```diff
@@ -262,6 +261,15 @@ func isServerUp(server string) bool {
 		return false
 	}
 
+	ocRequestTimeout := config.DefaultTimeout * time.Second
```
mik-dass: we can skip this line as `cfg.GetTimeout()` will return the Default value in the else statement

cmd/config.go
```diff
@@ -19,15 +19,16 @@ var configurationCmd = &cobra.Command{
 
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
-NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix`,
+NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix
+Timeout - Timeout (in seconds) for openshift server connection check`,
```
cdrage: OpenShift

cmd/config.go
```diff
@@ -47,12 +48,15 @@ var configurationSetCmd = &cobra.Command{
 	Long: `Set an individual value in the Odo configuration file
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
-NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.`,
+NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.
+Timeout - Timeout (in seconds) for openshift server connection check`,
```
cdrage: OpenShift

cmd/config.go
```diff
@@ -47,12 +48,15 @@ var configurationSetCmd = &cobra.Command{
...
 	Example: `
    # Set a configuration value
    odo utils config set UpdateNotification false
    odo utils config set NamePrefix ""
```
cdrage: I think this line could be removed to make things shorter

cmd/config.go
```diff
@@ -47,12 +48,15 @@ var configurationSetCmd = &cobra.Command{
...
    odo utils config set NamePrefix "app"
+   odo utils config set timeout 20
+   odo utils config set timeout 0
```
cdrage: same as this line, the line above is a good example, no need to show another example with just a different value

pkg/config/config.go
```diff
@@ -134,28 +138,50 @@ func (c *ConfigInfo) writeToFile() error {
...
 func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
 	switch parameter {
+
+	case "timeout":
+		typedval, err := strconv.Atoi(value)
```
cdrage: add comments on what's happening here :)

pkg/config/config_test.go
```diff
@@ -920,6 +920,62 @@ func TestDeleteApplication(t *testing.T) {
...
+	os.Setenv(configEnvName, tempConfigFile.Name())
+	zeroValue := 0
+	nonzeroValue := 5
+	tests := []struct {
+		name           string
+		existingConfig Config
+		want           int
+	}{
+		{
+			name:           "validating value 1 from config in default case",
```
cdrage: you should have "Case 1, Case 2, etc." for the names, similar format as the other tests we have.

pkg/config/config_test.go
```diff
@@ -967,8 +1029,66 @@ func TestSetConfiguration(t *testing.T) {
...
+			name:           "updatenotification invalid value",
+			parameter:      "updatenotification",
+			value:          "invalid_value",
+			existingConfig: Config{},
+			wantErr:        true,
+		},
+
+		// timeout
+		{
+			name:      "Timeout set to 5 from 0",
```
cdrage: Again, should add "Case 1, Case 2" for names similar to other tests.

tests/e2e/e2e_test.go
```diff
@@ -81,10 +82,23 @@ var _ = Describe("odoe2e", func() {
...
+			Expect(config_output).To(ContainSubstring("UpdateNotification"))
+		})
+		It("should be checking to see if timeout is the same as the constant", func() {
+			config_output := runCmd("odo utils config view|grep Timeout")
+			Expect(config_output).To(ContainSubstring(fmt.Sprintf("%d", config.DefaultTimeout)))
+		})
+		It("should be checking to see if config values are the same as the configured ones", func() {
+			runCmd("odo utils config set updatenotification false")
+			runCmd("odo utils config set timeout 5")
+			config_output := runCmd("odo utils config view|grep UpdateNotification")
```
cdrage: this is unconventional golang naming. The naming should be: `configOutput`.

pkg/occlient/occlient.go
syamgk: No, this is the case `if the config file is not able to read`, it wont fail instead proceed with the default value

syamgk: `unable to read config file` will be logged on odo cmd

mik-dass: OK but please add a comment describing the use of this line so that in future it does not get removed my accident.

cmd/config.go
```diff
@@ -19,15 +19,16 @@ var configurationCmd = &cobra.Command{
 
 Available Parameters:
```
surajnarwade: add new line before this line

cmd/config.go
```diff
@@ -47,12 +48,14 @@ var configurationSetCmd = &cobra.Command{
 	Long: `Set an individual value in the Odo configuration file
 Available Parameters:
```
surajnarwade: this line as well

pkg/config/config.go
surajnarwade: cc @syamgk

surajnarwade:
very small changes needed &&  rest is LGTM :tada:

kadel:
This PR is opened for more than a month :-(
Can we do something with this?

pkg/config/config.go
syamgk: added comment here https://github.com/redhat-developer/odo/pull/757/files#diff-091ace02b49a46f55f2586cfec89a160R143

cmd/config.go
syamgk: added line above

surajnarwade:
@cdrage , @mik-dass please review so that we can merge this

cdrage:
LGTM! Let's merge :+1:

Issue #758, opened 2018-09-18
New line missing for -- short for better UI/UX


mik-dass:
Currently 
```
[mrinaldas@localhost odo]$ ./odo app get --short
app[mrinaldas@localhost odo]$ 
```
```
[mrinaldas@localhost odo]$ ./odo project set --short odo-1
odo-1[mrinaldas@localhost odo]$ 
```
```
[mrinaldas@localhost odo]$ ./odo component get --short
nodejs[mrinaldas@localhost odo]$ 
```
But 
```
[mrinaldas@localhost odo]$ ./odo project get --short
odo-1
[mrinaldas@localhost odo]$
```
So for UI/UX consistency, we need to add a new line for the short flags or maybe remove it from all of them.

PR #759, opened 2018-09-18
Removes a new line from the project get short flag

mik-dass:
fixes : #758

It removes the new line from the short flags for `project get`

### DIFF
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -82,7 +82,7 @@ var projectGetCmd = &cobra.Command{
 		project := project.GetCurrent(client)
 
 		if projectShortFlag {
-			fmt.Println(project)
+			fmt.Print(project)
 		} else {
 			fmt.Printf("The current project is: %v\n", project)
 		}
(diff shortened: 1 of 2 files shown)


mik-dass:
e2e tests need to be modified too

ashetty1:
But is there a specific reason why we need this PR? `--short` flag is normally used within scripts and not users, for which we don't need a newline.

mik-dass:
@ashetty1 I just think it's better to maintain a constant UI/UX throughout the app

ashetty1:
@mik-dass I meant, `--short` is not 'user' interface technically. Anyway, let's see what other say.

mik-dass:
@kadel @cdrage WDYT?

kadel:
Short flags are there to make it easier to write scripts around `odo`.

It would be much better to remove new line than add it.

mik-dass:
@kadel @ashetty1  Done. I have removed the short flag from the project get command and fixed the test, please have a look.

PR #766, opened 2018-09-20
Add provision for running all e2e tests

ashetty1:


Makefile
```diff
@@ -75,9 +75,9 @@ prepare-release: cross
 test:
 	go test -race $(PKGS)
 
-# Run e2e tests
-.PHONY: test-e2e
-test-e2e:
+# Run main e2e tests
```
kadel: it would be good to describe what it "main e2e tests" mean

cdrage:
LGTM :+1:

cdrage:
We might have to merge this in and then modify this after in case @ashetty1 is away :100: 

But thanks for this fix!

anmolbabu:
@kadel can we pull this into master?

cdrage:
Let's merge :+1:

PR #791, opened 2018-09-28
add short flags

### LINKED ISSUE #674: Short vs long flags
Should we support short, long or both type of flags?

e.g:
- `-p` or `--port`
- `-e` or `--env`

Does it make sense to have equivalents to most of the flags, to the ones most used to all, to none?
Depending on the resolution, we should open issues here to track adding these.

@kadel

syamgk:
adding the below short flags to corresponding long flags 
resolves #674
```
-b --binary
-g --git
-l --local
-c --component
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -182,9 +182,9 @@ A full list of component types that can be deployed is available using: 'odo cat
 }
 
 func init() {
-	componentCreateCmd.Flags().StringVar(&componentBinary, "binary", "", "Use a binary as the source file for the component")
-	componentCreateCmd.Flags().StringVar(&componentGit, "git", "", "Use a git repository as the source file for the component")
-	componentCreateCmd.Flags().StringVar(&componentLocal, "local", "", "Use local directory as a source file for the component")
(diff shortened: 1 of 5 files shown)


cmd/create.go
```diff
@@ -182,9 +182,9 @@ A full list of component types that can be deployed is available using: 'odo cat
 }
 
 func init() {
-	componentCreateCmd.Flags().StringVar(&componentBinary, "binary", "", "Use a binary as the source file for the component")
-	componentCreateCmd.Flags().StringVar(&componentGit, "git", "", "Use a git repository as the source file for the component")
-	componentCreateCmd.Flags().StringVar(&componentLocal, "local", "", "Use local directory as a source file for the component")
+	componentCreateCmd.Flags().StringVar(&componentBinary, "binary", "b", "Use a binary as the source file for the component")
```
anmolbabu: Doesn't this mean the default value for `binary` flag is `b`?
Shouldn't we use StringVarP instead?
Please correct me if I am wrong

syamgk:
@kadel removed the changes on --component flag

PR #797, opened 2018-09-30
Fix watch command

### LINKED ISSUE #796: odo watch detects 1 change and subsequently fails to detect more changes
odo watch on a component created from local source, detects 1 change and does not detect subsequent changes

Steps to Reproduce:
1. On Linux, when a file is edited/created using vim editor or even gedit editor and then perform update to any file in watched source
2. Consistently on windows for any file change, subsequent changes are not detected

Reason:
Break  in select as in https://github.com/redhat-developer/odo/blob/master/pkg/component/watch.go#L157, https://github.com/redhat-developer/odo/blob/master/pkg/component/watch.go#L149

anmolbabu:
A break from select case with signal handling causes complete
exit from receiving that particular signal. So avoid break from
select case with signal handling.

fixes #796

### DIFF
--- a/pkg/component/watch.go
+++ b/pkg/component/watch.go
@@ -141,13 +141,16 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
 				if !(event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename) {
 					stat, err := os.Lstat(event.Name)
 					if err != nil {
-						glog.Errorf("Failed getting details of the changed file %s", event.Name)
-						//watchError = errors.Wrap(err, "unable to watch changes")
+						// Some of the editors generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
+						// So, its better to log the error rather than feeding it to error handler via `watchError = errors.Wrap(err, "unable to watch changes")`,
+						// which will terminate the watch
+						glog.Errorf("Failed getting details of the changed file %s. So, ignoring te event", event.Name)
 					}
 					// Some of the editors generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
-					// Avoid pushing such buffer files
+					// So, its better to log the error rather than feeding it to error handler via `watchError = errors.Wrap(err, "unable to watch changes")`,
+					// which will terminate the watch
 					if stat == nil {
-						glog.V(4).Infof("Ignoring event for temp file %s", event.Name)
+						glog.Errorf("Ignoring event for file %s as details about the file couldn't be fetched", event.Name)
 						isIgnoreEvent = true
 					}
 
--- a/pkg/component/watch.go
+++ b/pkg/component/watch.go
@@ -144,7 +144,7 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
 						// Some of the editors like vim and gedit, generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
 						// So, its better to log the error rather than feeding it to error handler via `watchError = errors.Wrap(err, "unable to watch changes")`,
 						// which will terminate the watch
-						glog.Errorf("Failed getting details of the changed file %s. So, ignoring te event", event.Name)
+						glog.Errorf("Failed getting details of the changed file %s. Ignoring the change", event.Name)
 					}
 					// Some of the editors generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
 					// So, its better to log the error rather than feeding it to error handler via `watchError = errors.Wrap(err, "unable to watch changes")`,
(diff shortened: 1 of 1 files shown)


anmolbabu:
Verified on windows and linux

pkg/component/watch.go
```diff
@@ -134,27 +134,29 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
...
+				isIgnoreEvent := false
 				changeLock.Lock()
 				glog.V(4).Infof("filesystem watch event: %s", event)
 
 				if !(event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename) {
 					stat, err := os.Lstat(event.Name)
 					if err != nil {
 						glog.Errorf("Failed getting details of the changed file %s", event.Name)
-						watchError = errors.Wrap(err, "unable to watch changes")
+						//watchError = errors.Wrap(err, "unable to watch changes")
```
kadel: why is this error ignored? If there is a reason for it it should be explained in the comment

pkg/component/watch.go
```diff
@@ -134,27 +134,29 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
...
 					}
 					// Some of the editors generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
 					// Avoid pushing such buffer files
 					if stat == nil {
-						break
+						glog.V(4).Infof("Ignoring event for temp file %s", event.Name)
+						isIgnoreEvent = true
```
kadel: How do you know that it is a temporary file?

pkg/component/watch.go
anmolbabu: The error is ignored because, in case of temporary files created by editors like vi or gedit that get deleted almost immediately and we get file not exists error sometimes. But if we feed this to error then the watch breaks, so I feel its better only to log it and leave it WDYT? @kadel

pkg/component/watch.go
anmolbabu: Because, the file was almost quickly(within interval between the event to handling the event) deleted . But I think I'll change it to say ignoring event for the now non-existant file

anmolbabu: to be more relaistic than making assumptions

pkg/component/watch.go
```diff
@@ -134,27 +134,32 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
...
+				isIgnoreEvent := false
 				changeLock.Lock()
 				glog.V(4).Infof("filesystem watch event: %s", event)
 
 				if !(event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename) {
 					stat, err := os.Lstat(event.Name)
 					if err != nil {
-						glog.Errorf("Failed getting details of the changed file %s", event.Name)
-						watchError = errors.Wrap(err, "unable to watch changes")
+						// Some of the editors generate temporary buffer files during update to the file and deletes it soon after exiting from the editor
```
surajnarwade: can you please mention the editors name as well for which you faced the problem

anmolbabu: Ack Done

anmolbabu:
@surajnarwade @kadel Incoporated your comments.. Please revisit the PR

@cdrage @syamgk @mik-dass Please review the PR

pkg/component/watch.go
```diff
@@ -134,27 +134,32 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
...
+						// So, its better to log the error rather than feeding it to error handler via `watchError = errors.Wrap(err, "unable to watch changes")`,
+						// which will terminate the watch
+						glog.Errorf("Failed getting details of the changed file %s. So, ignoring te event", event.Name)
```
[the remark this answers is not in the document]
cdrage: Actually, remove `So,` `Ignoring the change` sounds a bit better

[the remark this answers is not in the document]
anmolbabu: Ack

anmolbabu:
@kadel @cdrage @surajnarwade All comments incorporated Please revisit the PR

@syamgk @mik-dass Please review

cdrage:
Seems to work :+1: 

After you merge this, can you create another issue indicating that we should have an e2e test created for this?

anmolbabu:
> Seems to work 
> 
> After you merge this, can you create another issue indicating that we should have an e2e test created for this?

There's an issue for the tests already as in https://github.com/redhat-developer/odo/issues/770 and it is targeted for this sprint

anmolbabu:
Thanks @syamgk @cdrage @kadel For reviews

Pushing the PR in as it gets 2 acks :)

pkg/component/watch.go
anmolbabu: Marking the conversation resolved as the comments have been incoporated

anmolbabu:
> Thanks @syamgk @cdrage @kadel For reviews
> 
> Pushing the PR in as it gets 2 acks :)

Aah doesn't seem like I can do it without approval from @kadel and I don't want to misuse my privileges so will patiently wait for @kadel approval :)

Issue #77, opened 2018-02-13
Add component/application/service autocompletion


slemeur:
### Description
When using the commands pointing to a particular component or application, like `ocdev application delete <application_name>`, it would be handy to provide autocompletion.
This way the developer can:
- get the list of all applications / components, in case he does not have them in mind
- fasten the interaction with the commands

cdrage:
Linking this here: https://github.com/spf13/cobra/blob/master/bash_completions.md which was can use in order to generate the autocompletion / use the autocompletion for the list of components / applications. In particular, we'll be using `odo list`

cdrage:
We've gone ahead and implemented this (well, *half*). Autocompletion is complete, however, automatically grabbing the next component / list of components from using tab is not.

PR #801, opened 2018-10-02
Add implementation of bash completion for a few commands

geoand:
This PR implements the requirement in #77 but also goes on step further.
It provides bash completion support to:

* `odo create`
  - By providing the available components (essentially using `odo catalog list components`)
* `odo service create`
  - By providing the available services (essentially using `odo catalog list service`)
* `odo project delete|set`
  - By providing the available projects (essentially using `odo project list`)
* `odo storage delete|mount|unmount`
  - By providing the available storage objects (essentially using `odo storage list`)
* `odo url delete`
  - By providing the available url (essentially using `odo url list`)

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -30,6 +30,12 @@ const (
 		- 'cut' is used in order to remove the leading characters from the service names
 		- 'paste' is then used turn the multiple lines into a single line of names separated by spaces
 
+		'odo project delete|set':
+		Handled by providing of available projects
+		- 'tail' is used in order to drop the first line which is sort of a "header"
+		- 'sed' is used to ensure that we only keep the project name
+		- 'paste' is then used turn the multiple lines into a single line of names separated by spaces
+
 		More information about writing bash completion functions can be found at
 		https://debian-administration.org/article/317/An_introduction_to_bash_completion_part_2 for
 	*/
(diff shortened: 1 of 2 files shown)


geoand:
@surajnarwade  You are welcome!

Should I also create an issue for the service completion and link to it?

surajnarwade:
@geoand just update the same issue https://github.com/redhat-developer/odo/issues/77, that works :+1:

geoand:
@surajnarwade commits updated and added a description to the PR.

Would you like me to add anything else?

surajnarwade:
@geoand thanks, this looks good :)

geoand:
@surajnarwade Cool :)

geoand:
Also, I seem to not be able to remove labels

surajnarwade:
@geoand , we can add auto completion for: (if possible)

* `odo project`
* `odo storage`
*  `odo url`
* etc etc

geoand:
@surajnarwade Wasn't aware of those, I'll lool into them, thanks!

geoand:
@surajnarwade Should be good to go (unless there is some other command you can think of that would benefit from completion).
It seems like I don't have enough Karma to in order to edit labels :P

anmolbabu:
@geoand I was trying to verify this PR on my f27 laptop
but it seems when I key in odo create and hit tab, it only gives me the files in my current folder...
Am I missing something?

mik-dass:
@anmolbabu Same here. @geoand Any steps to follow before?

geoand:
@anmolbabu @mik-dass Here are the steps I used:

```
make bin
sudo cp odo /usr/local/bin/
odo utils completion bash > ~/tmp/odo_completion.sh
source ~/tmp/odo_completion.sh
```

That way the completion is added only to the current terminal session. Also note that the built `odo` binary needs to be on the path (since the bash completion scripts use it), which is why I have copied it into `/usr/local/bin/`.

Please let me know if these steps work for you

anmolbabu:
> @anmolbabu @mik-dass Here are the steps I used:
> 
> ```
> make bin
> sudo cp odo /usr/local/bin/
> odo utils completion bash > ~/tmp/odo_completion.sh
> source ~/tmp/odo_completion.sh
> ```
> That way the completion is added only to the current terminal session. Also note that the built `odo` binary needs to be on the path (since the bash completion scripts use it), which is why I have copied it into `/usr/local/bin/`.
> 
> Please let me know if these steps work for you

@geoand seems to be working now for me but can we wrap these steps up in a make command? and ease the use of this feature

geoand:
@anmolbabu Good idea! What would you like name of the goal to be?

surajnarwade:
@anmolbabu @mik-dass  @geoand ,

you can simply do `  source <(odo utils completion bash)`

or add this line to your bashrc

geoand:
@surajnarwade Good one, thanks! 

Note that there are ways of making this completion "permanent", but I didn't want to do that for testing purposes, which is why I used `source`

anmolbabu:
> @anmolbabu Good idea! What would you like name of the goal to be?

@geoand may be `enable-auto-completion` ?

geoand:
@anmolbabu Thank you for your kind words!

Your suggestion is very very reasonable and makes a ton of sense for a maintenance perspective. 
For the time being I don't have an answer on whether or not it's technically possible (and FWIW I haven't seen an other projects use it), but my hunch it that we could probably figure out a way to make it work.

geoand:
I'm looking into creating the make target you mentioned, should be done soon :)

geoand:
@anmolbabu It doesn't seem like we can `source` from inside a `Makefile` goal, since those goals operate in a new shell.
I am by no means a `make` expert so there might be advanced feature that I have missed here...

anmolbabu:
> @anmolbabu It doesn't seem like we can `source` from inside a `Makefile` goal, since those goals operate in a new shell.
> I am by no means a `make` expert so there might be advanced feature that I have missed here...

Np we can take that up later I don't want to block the PR for it plus its also already there in command help

geoand:
You are welcome @surajnarwade!

mik-dass:
@geoand Thanks for the awesome PR. Tested locally and it seems to work. But just two edge cases that I found
1. odo supports specifying component, app and project using -c/-a/-p. So auto completion fails in cases like 
`odo url delete -c php example-url`

2. If no service/url and others exist, it shows the list of files/ folders in the current folder, which seems confusing and might be better to show nothing.

So will it be possible to resolve these edge cases?

geoand:
@mik-dass Thanks for pointing out these issues!

1. This can probably be accomplished. It will definitely make the scripts more complex, but I am confident that it's doable.
2. I am not sure if this is possible to overcome. This might be the default bash behavior of falling back on the completing using the files in the current directory. I need to look into it

geoand:
@mik-dass I found the cause of 1).
The cause is what `cobra` generates: https://github.com/redhat-developer/odo/blob/v0.0.13/vendor/github.com/spf13/cobra/bash_completions.go#L283-L285

If we could drop `-o default` on those lines, then we would fix the issue. 
But given that there don't appear to be extension points about in Cobra, I am not sure it makes sense to re-implement all the completion function just to add this feature

mik-dass:
@geoand Let's leave it for now. We can take this up at a later stage.

geoand:
Cool

geoand:
Actually I created a [PR](https://github.com/spf13/cobra/pull/757) in cobra that we could use :)

mik-dass:
awesome :+1:

geoand:
Hey @cdrage.

The PR is totally bash specific (as is Cobra's completion for that matter).

As for the absence of of an error you mention, it is similar to what @mik-dass mentioned in his comments if I understand correctly. 
Correctly Cobra limits us in what we can do... So without significant effort on our part (although I have made a small PR to cobra in that direction), we have to live with what Cobra gives us...

cdrage:
@geoand 

Cobra has had zsh integration for a while: https://github.com/spf13/cobra/blob/master/zsh_completions.go

But yeah, we could always revisit ZSH later. Mind adding a small note / TODO that it's bash support only? Maybe a TODO: under zsh to integrate these commands :)

geoand:
Indeed Cobra does have `zsh` like you correctly pointed out, but it's support seems more limited (or at least that is the feeling  I got :) ).

I will of course add the TODO. Where would like me to add it?

cdrage:
@geoand 

Just at the top would be fine! :) Just a `TODO:` comment saying that zsh support for the future, etc. It's just a good tracker for future implementations haha.

geoand:
@cdrage Done :)

geoand:
You're welcome!

I can of course add whatever other resources are missing :)

geoand:
Thanks everyone for reviewing!!!

Issue #800, opened 2018-10-02
Create a proposal for .odo config file


jorgemoralespou:
We need to define the format of a file that can be embedded into the source code repository so that when a user creates a component from a repository with this file, the process will be directed in some way. 
There are some specific things that we know can live along with the source code and will be used by odo. These things are not specific to one's deployment but more to the component deployment itself. e.g. are:

- memory
- cpu
- livenes-probes
- readiness-probes
- container behavior
- container image to use
- files to ignore (for watch)

There are some additional properties that could also be put into this file, which are not really specific to a deployment but could be used as default if no explicit values are specified.

These could be:
- name of the component

We need to create a proper proposal, before implementation that defines the format and use of this file.

jorgemoralespou:
Proposal was merged

PR #803, opened 2018-10-02
[PROPOSAL] Define configuration and format for a local .odo config file

jorgemoralespou:
@kadel This is an initial proposal to define what's described in issue #800. We can discuss in the PR about this and see how the process would work. If it works fine, I'll be happy to keep pushing more proposals.

### DIFF
--- a/docs/proposals/odo-config-file.md
+++ b/docs/proposals/odo-config-file.md
@@ -53,7 +53,7 @@ In this previous example, the user is explicitly instructing `odo` to use 2GB of
 
 ### <a name="configuration-precedence"></a> Configuration precedence
 
-As configuration can exist both localy and globaly or only on one of these scopes, it seems obvious that there needs to be some rule managing the precedence and applicability of the configuration.
+As configuration can exist both locally and globally or only on one of these scopes, it seems obvious that there needs to be some rule managing the precedence and applicability of the configuration.
 
 The most explicit/concrete configuration wins over the less explicit one. In this case, providing a configuration value via a CLI argument always wins over any other existing configuration. If no explicit value is provided, then the configuration defined along with the component's source code wins over any other configuration. In case there is no explicit configuration or configuration for that single component then any global configuration (if existing) will be applied. In any other case, `odo`'s defaults will be used.
 
--- a/docs/proposals/odo-config-file.md
+++ b/docs/proposals/odo-config-file.md
@@ -1,132 +1,222 @@
-# odo config file
+# `odo` config file
 
 ## Abstract
-When developers work on creating an application, there's 2 phases they go though. First is creating the application initial implementation from scratch, and then iterate development on an existing application. While the developer works on the application, he might take some decisions on how the application should behave that should be preserved over time, but that are not source code but specific to how the application should behave in relation to the platform where it's running or in relation to the tool used to develop it.
 
-IDEs often use a configuration file to hold specific configuration for an application, workspace or the whole tool. These files are stored under the same control version system as where the source code lives so that anytime a developer checks out that repository has control on how to configure that tool to work with that source code.
+When developers work on creating an application, there's 2 phases they go through. First phase is creating the application initial implementation from scratch, and then iterate development on an existing application. While the developer works on the application, he might take some decisions on how the application should behave that should be preserved over time, but that are not source code but rather specific to how the application should behave in relation to the platform where it's running or in relation to the tool used to develop it.
+
+IDEs often use a configuration file to hold specific configuration for an application, workspace or the whole tool. These files are stored under the same control version system as where the source code lives, so that anytime a developer checks out that repository has control on how to configure that tool to work with that source code.
+
+Git is another example of a tool that provides configuration, that can be related of every instance of any repository the user has or to a single instance. These are defined as global and local config. Git promoves the fact that local configuration is always more important than global configuration.
 
 ## Motivation
-Provide a way to preserve conscisous decisions the developer has made that are orthogonal to the code but relevant to the tool/platform. Any user that continues working on an existing application don't need to know the decisions the original developer took, as these are preserved along with the source code.
+
+Provide a way to preserve conscisous decisions the developer has made that are orthogonal to the code but relevant to the tool/platform. Any user that continues working on an existing application don't need to know the decisions the original developer took, as these are preserved along with the source code and provides for better reproducibility of the development environment setup.
 
 ## Use Cases
 
 1. As a user, I want to be able to preserve some configuration I've done about my current component along with my source code
 1. As a user, I want to be able to use configuration stored with my source code to help me be more productive
 1. As a user, I want to be able to manage explicit configuration that is saved with my source code
+1. As a user, I want to be able to check out an application source code on different computers and have them producing the same result in relation to the development environment (deployment of the application on the platform)
 
 ## Design overview
-We need to provide a local file, `.odo` (name subject to change) that will hold specific configuration for the source code to instruct how a component should be configured.
 
-### Save user provided explicit flags
-The user should not need to explicitly save the configuration in this file, but the odo CLI should keep the values explicitly used by a user into this file.
+We need to provide a configuration file, `.odo`, that will hold specific configuration that `odo` will use on execution. This configuration might modify how the CLI work or how the component will work on the platform.
 
-- If the file does not exist, it should create it.
-- If the value is already present in the file, it should overwrite it.
+### Scopes of configuration
 
-```
-odo create wildfly example --memory=1GB
+`odo` CLI configuration needs to be preserved in configuration files. As a developer might work on multiple
+components, there is a need to have different scopes for configuration so that a developer don't need to
+explicitly configure each and every component independently.
 
-# This should set value in .odo config file   (memory=2GB)
-```
+#### Local configuration
 
-### Use configuration to "decorate" the deployment
-On component creation, via `odo create ...` the CLI will take information stored in this configuration file to create an actual deployment on OpenShift.
+Every component can be configured independently. We call this local configuration for the component as it only affects the component it references. This configuration will be stored in a file in the root directory of the component's source code. In this way, the configuration file can be added to version control system and preserved along with the application. If any other developer checks out the same source code he will get the same configuration as the original developer.
 
-```
-# Value in .odo config file   (memory=2GB)
-odo create wildfly example
-```
+A good representation of this is a developer that while on the development process realizes his application requires 2 GB of memory to run appropriately, he might decide to save that information in this configuration file. At a later moment, any other developer (or even him in the future) that decides to work on the same application will have this configuration definition along with his source code. `odo` should take into account this configuration definition and materialize when appropriate. The type of configuration, format and applicability will be described later in this document.
 
-The deployment that should be created, would have the following configuration:
+#### Global configuration
 
-```
-apiVersion: v1
-kind: Pod
-metadata:
-spec:
-  containers:
-  - image: xxx
-    resources:
-      limits:
-        memory: 2Gi
-      requests:
-        memory: 2Gi
+When working with multiple components it might be tedious to configure each and every component, or a user wants to have a default configuration value applied to each and every component he works with when no explicit (local) configuration exists. For this purpose, there should be a file, stored in the user's home directory, that should contain this configuration definitions. This configuration is global to every component the user works with.
+
+#### User explicit configuration
+
+Often times a user wants to use different configuration values than the provided in the configuration file. We need to provide a way to allow a developer to specify this new values he want to use. For that, the user needs to explicitly provide he values as CLI arguments to the speficic `odo` command.
+
+An example of this would be:
+
+```bash
+odo create wildfly --memory 2GB
 ```
 
-### Use configuration to "decorate" the CLI behavior
-On component creation, via `odo create ...` the CLI will take information stored in this configuration file to define some values that would typically be defined by the CLI and that are relative to how this works..
+In this previous example, the user is explicitly instructing `odo` to use 2GB of memory when creating a component based on wildfly.
 
-```
-# Value in .odo config file   (app-name=myapp, component-name=mycomp)
-odo create wildfly
-# Would create an application called myapp and component named mycomp using wildfly
-```
+### <a name="configuration-precedence"></a> Configuration precedence
 
+As configuration can exist both localy and globaly or only on one of these scopes, it seems obvious that there needs to be some rule managing the precedence and applicability of the configuration.
+
+The most explicit/concrete configuration wins over the less explicit one. In this case, providing a configuration value via a CLI argument always wins over any other existing configuration. If no explicit value is provided, then the configuration defined along with the component's source code wins over any other configuration. In case there is no explicit configuration or configuration for that single component then any global configuration (if existing) will be applied. In any other case, `odo`'s defaults will be used.
+
+|global|local|CLI flag|applied value|
+|--|--|--|-|
+|--|--|--|odo defaults|
+|Value A|--|--|Value A|
+|Value A|--|Value X|Value X|
+|Value A|Value B|--|Value B|
+|Value A|Value B|Value X|Value X|
+|--|Value B|--|Value B|
+|--|Value B|Value X|Value X|
+|--|-|Value X|Value X|
+
+### Configuration typology
+
+In the scop of this proposal, we refer to configuration in some different ways and it's worth noting what we meant in each case. What types of configuration will be supported and what use case they fullfil.
+
+### `odo` CLI behavior configuration
+
+Like every tool, there is configuration relative to how the tool behaves itself. An example of this would be a timeout that the CLI will apply when issuing commands to the server, or the ability to enable/disable checks for new versions. This configuration will always be global, and stored relative to the user's home directory. By default this configuration file will live in `$HOME/.kube/odo`
+
+The [format](#odo-file-format) of this file is described later in the proposal.
+
+### Application related configuration
+
+Some times what we want to configure is how the application/component will be deployed. Some components will
+require special characteristics, like a specific amount of memory. Some other times, what we want is to instruct `odo` to use an specific value to name our component if none is explicitly provided. This configuration will be stored locally to the component, along with the source code, in an `.odo` configuration file that can be saved into the version control system used, and share by anyone using the same application code.
+
+For some specific values, there will be the possibility to set them globally, as they will be used in that case for every component created, as defined in the [configuration precedence](#configuration-precedence) section. This values will be documented as being `global` configuration values.
 
 ### Developer friendly flags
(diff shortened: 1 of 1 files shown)


cmoulliard:
This proposition looks like what I also propose / describe within these G Slides around the concept about `component | application` crd -> https://docs.google.com/presentation/d/1Z-MlgbcpExhG79RolbFb20iZDiYfmdNfIz9DB71H4Vs/

jorgemoralespou:
@cmoulliard I don't see any relation at all. This is to define a local file to instruct/save/share how the tool will behave when creating a component from source.

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,132 @@
...
+odo utils config list --local                # List all config
+```
+
+## Proposed information to be stored
+We need to define and delimit what information will be stored in this file. 
+
+### Deployment decoration
+Configuration used to define how the deployment would be created:
+
+- mem
```
anmolbabu: Sorry I have not much openshift knowledge, but I would like to know what will be the use of fields like, memory, cpu etc.. after the component is created and if the user/developer has not created this file in his application and we create it based on the parameters passed to `odo create` command?

anmolbabu: Or do we also intend to watch over this configuration for every update of it, a new pod with the new config will be created and the old pod destroyed?

kadel: I suspect that this configuration will be used only when the component is created.

In the future, we might add something to the `odo update` command that will modify existing components, but I think that this proposal is about new components (making it easier to replicate component creation)
Or am I wrong @jorgemoralespou?

kadel:
Just to make sure that I understand it correctly, and that we are looking at it in the same way:

There will be two configs. One "global" (in users home, basically what we currently have in `~/.kube/odo` and second that is local to the project (`.odo` in the project root directory).
I'm going to call first as user's config and the second as project config.

Both configs will allow the same configurations. The highest priority has what is specified using CLI flag. That overrides everything. If the flag is not specified, then the project config will be used, if the value is not set there then the user's config, and if nothing is there than we use default value hardcoded in the code, or fail if the value is required.

And we are talking about component creation (`odo create` command)

docs/proposals/odo-config-file.md
kadel: this is mainly about component creation.
In the future (when there is a need for it), we might add something to the `odo update` command or add the flag to `odo config` that will instruct odo to make changes to the existing components.

jorgemoralespou:
@kadel, no, the idea is that "application related" configuration lives in the "application .odo" file. CLI specific configuration, which is global to how the tool/cli should behave (or some defaults) should be in the "global .odo" config file.

There should not be precedence of one over the other.

When creating a component, if no arguments are provided, the CLI should look for defaults, and it there are defaults, apply them and save them in the "application .odo" file. If there are no defaults, the "application .odo" file should not have config.

e.g.:
```
$HOME/..kube/.odo
----
odoignores=.git
memory=2GB
cpus=1
----

# User is creating a component. No $DIR/.odo file exists.
odo create wildfly example --memory=1GB

# This will create an .odo file with:
----
odoignores=.git                  # From global config
memory=1GB                      # From CLI, overrides default
cpus=1                                # From global config
app-name=app-example    # Autogenerated as no explicit app was created
comp-name=example         # Created via cli
...
----
```

Does it make sense?

kadel:
So `.odo` will allway contain all configuration that was used to create component.
After `odo create` is called all options (even if default value is used) will be specified in the `.odo` file.
So everyone can create component in excatly the same way if they want.

Am I closer?

jorgemoralespou:
We should limit the configuration here stored, but yes, that's the idea.

jorgemoralespou:
Or maybe we can skip adding a value if the default was used. Not sure about this one. Would only know as we keep using it.

kadel:
And running something like `odo utils config set memory 2Gi --local` will not have imidiate effect on running component, it will just modify config file?

jorgemoralespou:
@kadel yes. I would think that the natural way to update a component would be to do `odo update --memory 2GB`.
Maybe, we should avoid that command `odo utils config set memory 2Gi --local`, but I think that this command would be useful to interact with the local `.odo` file without creating a new deployment.

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,132 @@
...
+### Save user provided explicit flags
+The user should not need to explicitly save the configuration in this file, but the odo CLI should keep the values explicitly used by a user into this file.
+
+- If the file does not exist, it should create it.
+- If the value is already present in the file, it should overwrite it.
+
+```
+odo create wildfly example --memory=1GB
+
+# This should set value in .odo config file   (memory=2GB)
```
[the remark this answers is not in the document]
kadel: That is a good question. @jorgemoralespou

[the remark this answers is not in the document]
kadel: What we are missing in the document is description where `.odo` file will be created for different commands


| odo command | locatition for local `.odo` file|
|--|--|
|`odo create` | in the current directory |
|`odo create --local /home/user/mycode` | `/home/user/mycode/.odo`|
| `odo create --binary target/myapp.jar` | current directory or `target/.odo` or something else ? |
| `odo create --git https://example.com/myrepo |  current directory or try to read it remotely from git `https://example.com/myrepo/.odo` (tricky to implement)? |

Could we start with doing this just for local components?

syamgk:
sorry i did not get why this `--local` flag is required?

docs/proposals/odo-config-file.md
syamgk: also, is this `# This should set value in .odo config file   (memory=2GB)` 1GB instead of 2GB?

kadel: yes, that is typo ;-)

kadel:
> sorry i didn't get the relevance of `--local` flag
> is is to explicitly make the feel that its only editing the local file which is in (~/.kube/odo)?

no `--local` means that it is editing config for the component that typically that would be `.odo` file in component root directory /repository

docs/proposals/odo-config-file.md
syamgk: @anmolbabu these are the resource limits for the pod being deployed on openshift cluster.
similar to how much amount of memory and CPU allocated to a VM.

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,132 @@
...
+Currently that command is:
+
+```
+odo utils config [set|get|delete|add] KEY [VALUE] --local
+```
+
+Some examples of this command could be:
+
+```
+odo utils config set memory 2Gi --local      # Set's component memory to 2 GB
```
syamgk: this `--local` flag is a bit confusing?
can we change it to something like global ? 
If you mean `--local` by set the memory value on the local file lies in `$HOME/.kube/odo`

syamgk: got it from @kadel 's previous comment

syamgk:
> > sorry i didn't get the relevance of `--local` flag
> > is is to explicitly make the feel that its only editing the local file which is in (~/.kube/odo)?
> 
> no `--local` means that it is editing config for the component that typically that would be `.odo` file in component root directory /repository

got it, so that means there will be a local as well as a global .odo file

cmoulliard:
Does this config's file will contain the info of the application and associated components/services (one to many relation) as during the iterative processus a user will create for an application X components and potentially pickup Y services from the catalog ? If this is the case, then we need as a k8s's list, a field that we will use to populate this array of components/services

veillard:
I think we need 2 files, one file per repo, which is checked in with the component in git and providing all the defaults for the application/component. And One file in say the user home directory which can contain parameters overridden  locally without changing the global default. E.g. the memory config in the git config might be 8GB, but the user with a limited l;aptop may want to be able to set locally in his own config file that on 6GB should be used for the app.

docs/proposals/odo-config-file.md
anmolbabu: > @anmolbabu these are the resource limits for the application pod being deployed on openshift cluster.
> similar to how much amount of memory and CPU allocated to a VM.

@syamgk I know, I only wanted to understand as @kadel confirms that this is only for creating the components or do we also intend to re-create an already created pod if a local config was created post component creation(in which case we would need to watch for changes in expected config path) with the new config

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+As can be seen, when a scope is ommited, the value will be applied to the local configuration store.
+
+The verbs that add configuration are:
+
+* *create*: Creates a configuration entry with the value provided. If the key exists, this command should error with the corresponding message.
+* *add*: Adds a value to an existing config entry. The type of the entry should be a list. If the key does not exist, or is not a list, this command should error with the corresponding error message.
+
+*NOTE*: When creating a component, no configuration file will be created. If a developer thinks a configuration value is worth being stored in the `.odo` configuration file, he will need to use the previous command to create these files (whether global or local doesn't matter).
```
kadel: This is a big change from the previous version.

I actually liked the fact that it will be automatically created (if it didn't exist).
On the other hand, this is a more expressive way and less magical.

kadel:
This proposal looks great! 

It looks like something went wrong while rebasing or something, as there is a lot of unrelated commits. (I can fix that for you if you want ;-), but then you will probably need to delete your local branch and pull it again)

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+* runtime-image (local)
+  * src-path (local)
+  * deployment-path (local)
+
+### CLI behavior
+
+Configuration used to define how the CLI should behave:
+
+* app-name: (local) Preferred application name if none is defined
+* component-name: (local) Preferred component name if none is defined
```
anmolbabu: name or prefix only? and both(app prefix and component prefix) will be same or different?
I assume it can be only prefix right? Because, name seems to suggest that all apps or components will need to be named in entirety as configured...

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+* odoignore: (local/global) list of files/patterns to ignore by default if no .odoignore file is provided, or when it's cerated. **This configuration can be global, as the developer might want to always ignore specific files (e.g. .git, .odo, ...) when pushing.**
+
+#### <a name="odo-file-format"></a> `odo` file format
+
+The format for both files should be human readable. Whether it's yaml, json or [toml](https://github.com/toml-lang/toml) is not the scope of this proposal to define, as it will be dictated by the engineering team implementing this feature. In any case, it should be possible to:
+
+* Visually interpret the content of a file (vi, cat)
```
anmolbabu: I assume this 

> Visually interpret the content of a file (vi, cat)
> Process the file in an easy way (grep, aqk, sed, jq|yq)

is only for .odoignore file right? Because to me it( the above 2 points ) seems to suggest that the files will be exposed to manual edits...

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+odo utils config add odoignore .odo --local   # Adds a value to odoignore list local config
+odo utils config add odoignore .odo --global  # Adds a value to odoignore list global config
+```
+
+As can be seen, when a scope is ommited, the value will be applied to the local configuration store.
+
+The verbs that add configuration are:
+
+* *create*: Creates a configuration entry with the value provided. If the key exists, this command should error with the corresponding message.
+* *add*: Adds a value to an existing config entry. The type of the entry should be a list. If the key does not exist, or is not a list, this command should error with the corresponding error message.
```
anmolbabu: Not very sure, but can it be append(instead of add), so that it becomes more indicative of appending a value to list?

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+
+* *list*: Lists all configuration. No further argument.
+* *get*: Lists configuration for a specific entry. If the key does not exist this command should error with the corresponding message.
+
+#### Updating configuration
+
+When a user wants to update specific configuration.
+
+```bash
+odo utils config list --global                # List all global config
```
anmolbabu: should this be odo update example here instead?

docs/proposals/odo-config-file.md
kadel: name only. This will be in local .odo config file, it applies only to given component.

docs/proposals/odo-config-file.md
kadel: for all config files. Yes, it should be exposed to manual edits. But it will be suggested not to do that, and use  `odo utils config` commands to manipulate files.

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+* component-name: (local) Preferred component name if none is defined
+* odoignore: (local/global) list of files/patterns to ignore by default if no .odoignore file is provided, or when it's cerated. **This configuration can be global, as the developer might want to always ignore specific files (e.g. .git, .odo, ...) when pushing.**
+
+#### <a name="odo-file-format"></a> `odo` file format
+
+The format for both files should be human readable. Whether it's yaml, json or [toml](https://github.com/toml-lang/toml) is not the scope of this proposal to define, as it will be dictated by the engineering team implementing this feature. In any case, it should be possible to:
+
+* Visually interpret the content of a file (vi, cat)
+* Process the file in an easy way (grep, aqk, sed, jq|yq)
+* Be stored (and visualized) in version control system
```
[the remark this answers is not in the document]
kadel: yes

docs/proposals/odo-config-file.md
kadel: why `odo update`?

anmolbabu: sorry I meant `odo utils config update ...` instead of `odo utils config list` in the example

docs/proposals/odo-config-file.md
anmolbabu: Ok

docs/proposals/odo-config-file.md
jorgemoralespou: @kadel Maybe we can add a --keep flag to keep the config in the configuration file.

docs/proposals/odo-config-file.md
jorgemoralespou: Add is more intuitive, at least in Java that's the ver one would use to add to a list when order don't matter

docs/proposals/odo-config-file.md
jorgemoralespou: Fixed

docs/proposals/odo-config-file.md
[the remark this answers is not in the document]
jorgemoralespou: Fixed

docs/proposals/odo-config-file.md
jorgemoralespou: Fixed

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Developer friendly flags
+
+One of the goals of `odo` is to be easy to use for developers. This forces to translate some configuration possible in the underlying technology (OpenShift/Kubernetes) into a friendlier name that developers could easily understand. An example of this would be the amount of memory a component needs. In OpenShift, this is reflected by limits.memory and requests.memory, and dependening how you set these values influence the QoS of the deployment. In `odo` these should be simplified to `memory`, `min-memory` and `max-memory`. These are terms that any developer will most likely understand without looking at the documentation.
+
+* When `memory` is set, memory requests and limits are the same.
+* `min-memory` should be memory requests.
+* `max-memory` should be memory limits.
+* `memory` is exclusive with `min-memory` or `max-memory`. If set, min-mem and max-mem have the same value as `memory`.
```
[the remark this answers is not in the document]
jorgemoralespou: Fixed

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,223 @@
...
+
+```bash
+odo create wildfly --memory 2GB
+```
+
+In this previous example, the user is explicitly instructing `odo` to use 2GB of memory when creating a component based on wildfly.
+
+### <a name="configuration-precedence"></a> Configuration precedence
+
+As configuration can exist both localy and globaly or only on one of these scopes, it seems obvious that there needs to be some rule managing the precedence and applicability of the configuration.
```
surajnarwade: just a minor spell check,
`locally` and `globally`

jorgemoralespou: Fixed.

jorgemoralespou:
@kadel @surajnarwade @anmolbabu @syamgk I've updated the proposal with what we discussed today in the cabal meeting. 
Let's do one round of feedback and then if everything is ok, merge to take it into L train planning.

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,253 @@
...
+
+`odo` CLI configuration needs to be preserved in configuration files. As a developer might work on multiple
+components, there is a need to have different scopes for configuration so that a developer don't need to
+explicitly configure each and every component independently.
+
+#### Local configuration
+
+Every component can be configured independently. We call this local configuration for the component as it only affects the component it references. This configuration will be stored in a file in `.odo` specific directory located at the root directory of the component's source code. In this way, the configuration file can be added to version control system and preserved along with the application. If any other developer checks out the same source code he will get the same configuration as the original developer.
+
+We propose to have a `.odo` folder in the root of the application's source code directory to store every odo local configurations.
```
syamgk: sorry i might be missing some context here
but why do we need `.odo` folder instead of `.odo` file?
is it because of multiple components under same application? but aren't we assuming a repo per component?_

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,253 @@
...
+
+#### Global configuration
+
+When working with multiple components it might be tedious to configure each and every component, or a user wants to have a default configuration value applied to each and every component he works with when no explicit (local) configuration exists. For this purpose, there should be a file, stored in the user's home directory's odo specific folder, that should contain this configuration definitions. This configuration is global to every component the user works with.
+
+We propose to have a `.odo` folder in the root of the users home directory to store every odo global configurations.
```
syamgk: also do we require `.odo` for global?

docs/proposals/odo-config-file.md
jorgemoralespou: I did send an email around on tuesday and we discuss at cabal. We switched from file to folder so we can have more configuration than what a single file can store. In the future this might be fragments, or k8s descriptors, ...

docs/proposals/odo-config-file.md
jorgemoralespou: Same as before. Not so much currently needed, but this might be a place where maybe we store plugins, or plugin configuration, so yes, better to have a tool folder.

syamgk: sounds good, thanks for clarifying :thumbsup:

docs/proposals/odo-config-file.md
syamgk: Now understood, cool

Issue #675, opened 2018-08-28
R&D - component configuration semantics 


jorgemoralespou:
We need to provide configuration semantics in different scenarios:

- Component creation time. When creating a component, we need to be able to provide configuration for this component in different manners.
  - Just a value as an environment variable
  - A list of values as environment variables
  - A list of values stored in a Configuration object (ConfigMap/Secret)
  - A list of values injected as a file in the component (ConfigMap/Secret) 

- After the fact. When a component exists, we need to be able to provide/update/remove configuration for this component in different manners.
  - Just a value as an environment variable
  - A list of values as environment variables
  - A list of values stored in a Configuration object (ConfigMap/Secret)
  - A list of values injected as a file in the component (ConfigMap/Secret)

We need to define what will be the primitives and possibilities supported. Let's use this issue to define possible options and then come up with a proposal before implementation.

Following is a list of examples that show what needs to be done now, and how this could look like.

### **UC-CF1: Create a component with an environment variable**
* Current state:
```
odo create wildfly frontend
oc set env dc/frontend DEBUG=true                       # This triggers a redeployment
```

* Desired state at creation time: **UC-CF1_1: Create a component with an environment variable**

```
odo create wildfly frontend -e DEBUG=true
```

* Desired state after creation time: **UC-CF1_2: Update a component with an environment variable**

```
odo create wildfly frontend
odo update -e DEBUG=true
```

### **UC-CF2: Create a component with properties from a configmap**

* Current state at creation time:

```
odo create wildfly frontend
oc create configmap my-config --from-file=path/to/ba
oc set env dc/frontend --from=configmap/my-config       # This triggers a redeployment
```

* Desired state: **UC-CF2_1: Create a component with properties from a configmap**

```
odo create wildfly frontend -E path/to/ba 
```

* Desired state after creation time: **UC-CF2_2: Update a component with properties from a configmap**

```
odo create wildfly frontend
odo update -E path/to/ba
```

* Alternate desired state after creation time: **UC-CF2_2: Add configuration to a component with properties from a configmap**

```
odo create wildfly frontend
odo config create path/to/ba
```

### **UC-CF3: Create a component with a configmap as a file**

* Current state:

```
odo create wildfly frontend
oc create configmap my-config --from-file=path/to/ba
oc set volume dc/frontend --add -m /myapp/config --configmap-name=my-config       # This triggers a redeployment
```

* Desired state at creation time: **UC-CF3_1: Create a component with a configmap as a file**

```
[...]

PR #806, opened 2018-10-04
[PROPOSAL] add proposal for odo component configuration

syamgk:
proposing a way for users to deal with environment variables
and configuration files for the applications
deployed on openshift cluster using odo.
fix issue #675

### DIFF
--- a/docs/proposals/odo-config-file.md
+++ b/docs/proposals/odo-config-file.md
@@ -1,7 +1,43 @@
-#### Abstracting config map and Secrets from the user 
+## Abstract
 
-By default values will come from configmaps or secrets, the name of configmap/secret keys will be same as environment variable names.
-configmap name should be unique for each component, so using a combination of appname + component name. eg: <app_name>-<component_name>
+When developers work with applications on openshift there should be a way
+to deal with configuration artifacts using odo.
+
+Implement a way for users to deal with environment variables 
+and configuration files for the applications deployed on openshift cluster 
+using odo. User does not need to be bothered about complex k8s/openshift 
+terminologies like configmaps or secrets,so abstracting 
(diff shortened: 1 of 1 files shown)


docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,56 @@
+#### Abstracting config map and Secrets from the user 
+
```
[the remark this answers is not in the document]
syamgk: added motivation section and ready for review round 2 :)

docs/proposals/odo-config-file.md
```diff
@@ -0,0 +1,97 @@
...
+#### Abstracting config map and secrets from the user 
+
+By default values will come from configmaps or secrets, 
+the name of configmap/secret keys will be same as environment variable names.
+configmap name should be unique for each component, 
+so using a combination of appname + component name. eg: <app_name>-<component_name>
+
+*Problem with this way of implementing:*
+If the user goes into configmap using oc or webUI and modify key names then it becomes messy.
```
anmolbabu: Did we find a solution to this?
We are like any ideal cli tool, as much stateless as possible right?
So, any operation on configmap via Odo is expected to fetch the configmap from openshift api and then permit/error out the operation seeking the then correct values from the user right?
It is very much possible that I do not have the full context, if so, please help me understand what the problem is

syamgk: Here the logic is to keep the `env variable names` same as `config map/secret keys names` and nothing being stored in local machine. So If the user adds a new configmap key or a new environment variable out of our scope then we wont process them as config instead will show a warning.

anmolbabu: Ok

PR #809, opened 2018-10-05
Set up service-catalog on travis

### LINKED ISSUE #725: service catalog on travis CI
Let's have:

* separate script in `scripts` directory for spinning openshift cluster which will have service catalog
* separate job on travis for service catalog related e2e tests

cc @ashetty1

mik-dass:
fixes #725 

This PR setups up service catalog on travis with few examples.

### DIFF
--- a/tests/e2e/e2e.go
+++ b/tests/e2e/e2e.go
@@ -65,18 +65,19 @@ func pingSvc(url string) {
 // waitForCmdOut runs a command until it gets
 // the expected output.
 // It accepts 2 arguments, cmd (command to be run)
-// & expOut (expected output).
+// timeout (the time to wait for the output)
+// check (function with output check logic)
 // It times out if the command doesn't fetch the
-// expected output  within the timeout period (1m).
-func waitForCmdOut(cmd string, expOut string) bool {
+// expected output  within the timeout period.
+func waitForCmdOut(cmd string, timeout int, check func(output string) bool) bool {
 
-	pingTimeout := time.After(1 * time.Minute)
+	pingTimeout := time.After(time.Duration(timeout) * time.Minute)
 	tick := time.Tick(time.Second)
 
 	for {
 		select {
 		case <-pingTimeout:
-			Fail("Timeout out after 1 minute")
+			Fail("Timeout out after " + string(timeout) + " minutes")
 
 		case <-tick:
 			out, err := exec.Command("/bin/sh", "-c", cmd).Output()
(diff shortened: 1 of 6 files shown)


tests/e2e/e2e.go
```diff
@@ -119,3 +119,31 @@ func waitForDeleteCmd(cmd string, object string) bool {
 		}
 	}
 }
+
+// TODO
+func waitForServiceCreateCmd(cmd string, object string, status string) bool {
+
+	pingTimeout := time.After(1 * time.Minute)
+	tick := time.Tick(time.Second)
```
golangcibot: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here

tests/e2e/e2e.go
```diff
@@ -119,3 +119,37 @@ func waitForDeleteCmd(cmd string, object string) bool {
...
 }
+
+// waitForServiceCreateCmd runs a command until it finds
+// the deleted resource to be absent in the list of relevant objects from cmd executed
+// It accepts 3 arguments, cmd (command to be run)
+// object (the created object).
+// & status (the required status of the service)
+// It times out if the command doesn't fetch the
+// expected output  within the timeout period (5m).
+func waitForServiceCreateCmd(cmd string, object string, status string) bool {
```
kadel: This is 3rd function that is almost the same ([waitForCmdOut(cmd string, expOut string)](https://sourcegraph.com/github.com/redhat-developer/odo/-/blob/tests/e2e/e2e.go?utm_source=inline-symbol#L71:6-71:19) [waitForDeleteCmd(cmd string, object string)](https://sourcegraph.com/github.com/redhat-developer/odo/-/blob/tests/e2e/e2e.go?utm_source=inline-symbol#L100:6-100:22) and this one). 

If we can't create one, function that will cover all 3 use cases that it would be good to at least abstract away the common parts like timeout handling and command execution part

tests/e2e/service_test.go
```diff
@@ -0,0 +1,27 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("odoServiceE2e", func() {
+
+	Context("odo component creation", func() {
+		It("should be able to create a service", func() {
+			runCmd("odo service create mysql-persistent")
+			waitForServiceCreateCmd("odo service list | sed 1d", "mysql-persistent", "ProvisionedSuccessfully")
```
kadel: This check should be probably done using `oc` command not `odo`.  Otherwise, it is exactly the same what is the following test doing.

tests/e2e/service_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+		It("should be able to list the service", func() {
+			out := runCmd("odo service list | sed 1d")
+			Expect(out).To(ContainSubstring("mysql-persistent"))
+			Expect(out).To(ContainSubstring("ProvisionedSuccessfully"))
+		})
+
+		It("should be able to delete a service", func() {
+			runCmd("odo service delete mysql-persistent -f")
+			waitForDeleteCmd("odo service list", "mysql-persistent")
```
kadel: The same as for service creation, this should be verified using `oc` command.

tests/e2e/e2e.go
```diff
@@ -65,57 +65,69 @@ func pingSvc(url string) {
...
-func waitForDeleteCmd(cmd string, object string) bool {
+// waitForEqualCmd calls the waitForCmdOut function to wait and check if the output is equal to the given string within 1 min
+// cmd is the command to run
+// expOut is the expected output
+func waitForEqualCmd(cmd string, expOut string) bool {
 
-	pingTimeout := time.After(1 * time.Minute)
-	tick := time.Tick(time.Second)
+	return waitForCmdOut(cmd, 1, func(output string) bool {
+		if output == expOut {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>'

tests/e2e/e2e.go
```diff
@@ -65,57 +65,69 @@ func pingSvc(url string) {
...
+// expOut is the expected output which should not be contained in the output string
+func waitForDeleteCmd(cmd string, object string) bool {
 
-		case <-tick:
-			out, err := exec.Command("/bin/sh", "-c", cmd).Output()
-			if err != nil {
-				Fail(err.Error())
-			}
+	return waitForCmdOut(cmd, 1, func(output string) bool {
+		if !strings.Contains(output, object) {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>'

tests/e2e/e2e.go
```diff
@@ -65,57 +65,69 @@ func pingSvc(url string) {
...
-			if !strings.Contains(string(out), object) {
-				return true
-			}
+// waitForEqualCmd calls the waitForCmdOut function to wait and check if the output is equal to the given string within 5 mins
+// cmd is the command to run
+// expOut is the expected output
+func waitForServiceCreateCmd(cmd string, status string) bool {
+
+	return waitForCmdOut(cmd, 5, func(output string) bool {
+		if output == status {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>'

mik-dass:
@kadel Done

mik-dass:
@syamgk The PR setups up a travis job which setups up the oc environment with service catalog installed. Then run the defined e2e tests. The tests take time so delay with fixed timeout is setup.

cdrage:
Looks like this passes on travis, the code LGTM.

tests/e2e/e2e.go
```diff
@@ -65,57 +65,60 @@ func pingSvc(url string) {
...
+			Fail("Timeout out after " + string(timeout) + " minutes")
 
 		case <-tick:
 			out, err := exec.Command("/bin/sh", "-c", cmd).Output()
 			if err != nil {
 				Fail(err.Error())
 			}
 
-			if string(out) == expOut {
+			if check(string(out)) {
```
anmolbabu: return check(string(out)) ?

tests/e2e/e2e.go
```diff
@@ -65,57 +65,60 @@ func pingSvc(url string) {
...
-// the deleted resource to be absent in the list of relevant objects from cmd executed
-// It accepts 2 arguments, cmd (command to be run)
-// & object (the deleted object).
-// It times out if the command doesn't fetch the
-// expected output  within the timeout period (1m).
-func waitForDeleteCmd(cmd string, object string) bool {
+// waitForEqualCmd calls the waitForCmdOut function to wait and check if the output is equal to the given string within 1 min
+// cmd is the command to run
+// expOut is the expected output
+func waitForEqualCmd(cmd string, expOut string) bool {
```
anmolbabu: wouldn't timeout be required here as a parameter? why go with a default ;)
But would it be better to just implement checkCmdOutputEqual and may be checkCmdOutputContains and the caller just passes these standard func as args? only to avoid many wait funcs...
But it might be possible that, I am missing something

tests/e2e/e2e.go
mik-dass: We don't want to return return false because that would interrupt the test. Either timeout occurs and we exit or the command execution fails or the check condition is met else keep looping.

tests/e2e/e2e.go
mik-dass: 1 minute timeout is a long time to wait according to me so it's better to have a predefined one for now. For services 5 minutes is more than enough, so kept it by default to 5 minutes. Otherwise we would have to pass these values again and again which would be harder to maintain. It would though us customise able  behaviour but for now only a few specific tests are using it

> But would it be better to just implement checkCmdOutputEqual and may be checkCmdOutputContains and the caller just passes these standard func as args? only to avoid many wait funcs

I kept it this way to have customise able behaviour in the future and also have developer friendly names in the main script.

PR #813, opened 2018-10-05
Add proposals Template

kadel:


kadel:
@jorgemoralespou any suggestions that I should add to the template?

kadel:
> Maybe Example?

After the first proposal gets merged I'll link it as an example, it will be this one  https://github.com/redhat-developer/odo/pull/803

anmolbabu:
Merging the PR based on 2 acks and the replies to the queries on the PR

anmolbabu:
> Merging the PR based on 2 acks and the replies to the queries on the PR

Aah it seems @cdrage has a comment so I would wait to see if he is ok with it and consequently merge the PR

cdrage:
Yeah, I don't see any problems with this. This is a good initial template for anyone that wants to submit a proposal. LGTM.

PR #821, opened 2018-10-09
fix #690 : odo version shows server version without logging in

### LINKED ISSUE #690: Why do we suggest/propose to be logged ?
Why the command `odo version` suggest/propose to be logged ? 
Is it because you would like ti display additional info ? If this is not the case, then remove the message `Please log in to the cluster`

```
odo v0.0.10 (1e7221f4)
Please log in to the cluster
```

syamgk:
odo version shows the server version without logging in.
fixes #690

# How to test this?
First login and logout from a cluster, then 
run `odo version` - verify it shows the server version.

disconnect from the network and make the server not reachable from the machine where you are running odo
run `odo version` - verify it shows `Unable to connect to OpenShift cluster, is it down`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,7 +16,7 @@ import (
 
 // Global variables
 var (
-	GlobalConnectionCheck bool
+	GlobalSkipConnectionCheck bool
 )
 
 // Templates
@@ -130,7 +130,7 @@ func init() {
 	// will be global for your application.
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.odo.yaml)")
 
-	rootCmd.PersistentFlags().BoolVar(&GlobalConnectionCheck, "skip-connection-check", false, "Skip cluster check")
+	rootCmd.PersistentFlags().BoolVar(&GlobalSkipConnectionCheck, "skip-connection-check", false, "Skip cluster check")
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.CommandLine.Set("logtostderr", "true")
 
@@ -162,7 +162,7 @@ func getLatestReleaseInfo(info chan<- string) {
 }
 
 func getOcClient() *occlient.Client {
-	client, err := occlient.New(GlobalConnectionCheck)
+	client, err := occlient.New(GlobalSkipConnectionCheck)
 	checkError(err, "")
 	return client
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -40,7 +40,7 @@ var versionCmd = &cobra.Command{
 		fmt.Println("odo " + VERSION + " (" + GITCOMMIT + ")")
 
 		// turning off the flag which checks for login status
-		GlobalConnectionCheck = true
+		GlobalSkipConnectionCheck = true
 		// Lets fetch the info about the server
 		serverInfo, err := getOcClient().GetServerVersion()
 		checkError(err, "")
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -39,6 +39,8 @@ var versionCmd = &cobra.Command{
 
 		fmt.Println("odo " + VERSION + " (" + GITCOMMIT + ")")
 
+		// turning off the flag which checks for login status
+		GlobalConnectionCheck = true
 		// Lets fetch the info about the server
 		serverInfo, err := getOcClient().GetServerVersion()
 		checkError(err, "")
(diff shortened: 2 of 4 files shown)


cmd/version.go
```diff
@@ -39,6 +39,8 @@ var versionCmd = &cobra.Command{
 
 		fmt.Println("odo " + VERSION + " (" + GITCOMMIT + ")")
 
+		// turning off the flag which checks for login status
+		GlobalConnectionCheck = true
```
kadel: I just realized that `GlobalConnectionCheck` is not correctly named.
Shouldn't it be called `GlobalSkipConnectionCheck`? As `true` means that there will be no connection check.

syamgk: done,  renamed

surajnarwade:
I am getting this:

```
✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
12:11 $ odo version
odo v0.0.13 (HEAD)
the server could not find the requested resource
```

cmd/root.go
```diff
@@ -130,7 +130,7 @@ func init() {
 	// will be global for your application.
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.odo.yaml)")
 
-	rootCmd.PersistentFlags().BoolVar(&GlobalConnectionCheck, "skip-connection-check", false, "Skip cluster check")
+	rootCmd.PersistentFlags().BoolVar(&GlobalSkipConnectionCheck, "skip-connection-check", false, "Skip cluster check")
```
anmolbabu: We are hard coding its value in https://github.com/redhat-developer/odo/pull/821/files#diff-10b3cb428c695c798e01998a4fd2938dR43?
Should we instead only default it to true? or remove this flag altogether?

syamgk: > I am getting this:
> 
> ```
> ✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
> 12:11 $ odo version
> odo v0.0.13 (HEAD)
> the server could not find the requested resource
> ```

yep this is because the server from which you recently log out, is no more found/ reachable

syamgk:
> I am getting this:
> 
> ```
> ✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
> 12:11 $ odo version
> odo v0.0.13 (HEAD)
> the server could not find the requested resource
> ```

yep this is because the server from which you recently log out, is no more found/ reachable

cdrage:
@syamgk 

This *not* LGTM.

For: 

```sh
✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
12:11 $ odo version
odo v0.0.13 (HEAD)
the server could not find the requested resource
```

Should still not say "the server could not find the requested resource".

It should only spit out the version and that's it.

syamgk:
> @syamgk
> 
> For:
> 
> ```shell
> ✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
> 12:11 $ odo version
> odo v0.0.13 (HEAD)
> the server could not find the requested resource
> ```
> 
> Should still not say "the server could not find the requested resource".
> 
> It should only spit out the version and that's it.

@cdrage  in this case

syamgk:
> I am getting this:
> 
> ```
> ✔ ~/go/src/github.com/redhat-developer/odo [pr_821 L|⚑ 59] 
> 12:11 $ odo version
> odo v0.0.13 (HEAD)
> the server could not find the requested resource
> ```

hey @surajnarwade do you know how to reproduce this?

cmd/root.go
kadel: a quick search can help you understand why and when this flag was added ;-) https://github.com/redhat-developer/odo/issues/488

kadel: > Should we instead only default it to true? 

No it needs to be enabled  by default, otherwise user will get ugly error whithout much explaining what is happening. If I remember correctly this check if users is logged in or not.

surajnarwade:
working for me now:

```
$ odo version 
odo v0.0.13 (HEAD)

Server: https://api.starter-us-west-2.openshift.com:443
OpenShift: v3.11.16
Kubernetes: v1.11.0+d4cacc0
```

cdrage:
Works for me now! LGTM.

mik-dass:
I don't know if unit tests will be possible or not but please add a e2e test for this scenario :)

syamgk:
@mik-dass added e2e tests which will now validate the versions with `oc version` output

mik-dass:
Merging based on 3 LGTMS

PR #827, opened 2018-10-11
Upload binary to bintray for every PR

kadel:
For every PR build binary is uploaded to bintray to make it easy for anyone to test PR without a need to rebuild manually.

Followup to this will be creating a script that will automatically post a comment to PR with a link to where binary can be downloaded.


kadel:
recreated as https://github.com/redhat-developer/odo/pull/833,
something weird is happening with Travis, it wasn't running test for this

PR #832, opened 2018-10-11
Add a small test for IsVolumeAnEmptyDir

cdrage:
Adds a small test for IsVolumeAnEmptyDir.

Finishes (part) of https://github.com/redhat-developer/odo/issues/828

### DIFF
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -3221,3 +3221,54 @@ func TestUpdateDCToSupervisor(t *testing.T) {
 		})
 	}
 }
+
+func TestIsVolumeAnEmptyDir(t *testing.T) {
+	type args struct {
+		VolumeName string
+		dc         appsv1.DeploymentConfig
+	}
+	tests := []struct {
+		name         string
+		args         args
+		wantEmptyDir bool
+	}{
+		{
+			name: "Case 1 - Check that it is an emptyDir",
+			args: args{
+				VolumeName: supervisordVolumeName,
+				dc:         *fakeDeploymentConfig("foo", "bar"),
+			},
+			wantEmptyDir: true,
+		},
+		{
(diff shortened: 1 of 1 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -3221,3 +3221,49 @@ func TestUpdateDCToSupervisor(t *testing.T) {
...
+func TestIsVolumeAnEmptyDir(t *testing.T) {
+	type args struct {
+		VolumeName string
+		dc         appsv1.DeploymentConfig
+	}
+	tests := []struct {
+		name         string
+		args         args
+		wantEmptyDir bool
+		actions      int
```
kadel: It looks like that `actions` is not used anywhere

pkg/occlient/occlient_test.go
```diff
@@ -3221,3 +3221,49 @@ func TestUpdateDCToSupervisor(t *testing.T) {
...
+		},
+		{
+			name: "Case 2 - Check a non-existent volume",
+			args: args{
+				VolumeName: "foobar",
+				dc:         *fakeDeploymentConfig("foo", "bar"),
+			},
+			wantEmptyDir: false,
+			actions:      1,
+		},
```
kadel: there should probably be one more test case for volume that exists, but it is something else then emptyDir

cdrage:
@kadel 

Updated with the changes you requested (new case added, removes `actions` param)

mik-dass:
@kadel Can we merge this?

kadel:
> @kadel Can we merge this?

sure
I will leave mergin on @cdrage so he can reply to  @anmolbabu's comment

cdrage:
Hey @anmolbabu I had to keep it in the same occlient_test.go file since the supervisorD bootstrap templates are located within the same package.

PR #841, opened 2018-10-17
Updates spacing / newlines on --help messages

cdrage:
Some of the help messages have newlines as well as spacing issues. This
PR corrects it.


cmd/create.go
```diff
@@ -35,7 +35,7 @@ If version is not specified by default, latest wil be chosen as the version.
 
 A full list of component types that can be deployed is available using: 'odo catalog list'`,
 	Example: `  # Create new Node.js component with the source in current directory.
-  odo create nodejs
+	odo create nodejs
```
kadel: this makes it not align, there should be on only two spaces
```
  # Create new Node.js component with the source in current directory.
	odo create nodejs
```
```suggestion
  odo create nodejs
```

cdrage:
@kadel Updated!

PR #848, opened 2018-10-18
Fix intermittent bug of `odo update` failing

cdrage:
This fixes an intermittent bug of `odo update` failing due to Odo being
"too fast" on CI / local machines and not waiting for OpenShift to
update the correct annotation labels.

Closes: https://github.com/redhat-developer/odo/issues/835

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -57,6 +57,7 @@ import (
 )
 
 const (
+	ocUpdateTimeout    = 120 * time.Second
 	OpenShiftNameSpace = "openshift"
 
 	// The length of the string to be generated for names of resources
@@ -69,7 +70,7 @@ const (
 	supervisordVolumeName = "odo-supervisord-shared-data"
 
 	// waitForPodTimeOut controls how long we should wait for a pod before giving up
-	waitForPodTimeOut = 60 * time.Second
+	waitForPodTimeOut = 120 * time.Second
 )
 
 // errorMsg is the message for user when invalid configuration error occurs
@@ -877,6 +878,13 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig) error {
 		return errors.Wrapf(err, "unable to update DeploymentConfig %s", name)
 	}
 
+	// Watch / wait for deploymentconfig to update annotations
+	// importing "component" results in an import loop, so we do *not* use the constants here.
+	_, err = c.WaitAndGetDC(name, "app.kubernetes.io/component-source-type", dc.ObjectMeta.Annotations["app.kubernetes.io/component-source-type"], ocUpdateTimeout)
+	if err != nil {
+		return errors.Wrapf(err, "unable to wait for DeploymentConfig %s to update", name)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


kadel:
> Middle of adding a test for the new function, hence why it's missing, so don't merge just yet 👍 I just want to see that CI passes correctly with this new fix.

Please add "work in progress" or label, or mark it as such in the PR title. So we know when it is ready for review/merge and when more updates to the PR are expected.

pkg/occlient/occlient_test.go
```diff
@@ -3774,3 +3805,77 @@ func Test_updateEnvVar(t *testing.T) {
...
+			_, err := fakeClient.WaitAndGetDC(tt.args.name, tt.args.annotation, tt.args.value, tt.args.timeout)
+
+			// Error checking WaitAndGetDC
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.WaitAndGetDC() unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if err == nil && !tt.wantErr {
+				// Check to see how many actions are being ran
+				if (len(fakeClientSet.AppsClientset.Actions()) != tt.actions) && (tt.wantErr != true) {
```

cdrage:
Ready for review (other than the small TODO comment, need some insight on that!)

@surajnarwade @mik-dass @syamgk @kadel

pkg/occlient/occlient.go
```diff
@@ -1105,6 +1113,49 @@ func (c *Client) WaitForBuildToFinish(buildName string) error {
...
+		return nil, errors.Wrapf(err, "unable to watch dc")
+	}
+
+	timeoutChannel := time.After(timeout)
+	// Keep trying until we're timed out or got a result or got an error
+	for {
+		select {
+
+		// Timout after X amount of seconds
+		case <-timeoutChannel:
```
anmolbabu: why not directly use here:
`case <-time.After(timeout)` instead of https://github.com/redhat-developer/odo/pull/848/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR1129 ?

cdrage: I'm pretty sure it'd be endless if it's that case.. defining `time.After(timeout)` before the loop ensures it exits

anmolbabu: Sounds good it(time.After(timeout)) returns a new channel every-time :)

pkg/occlient/occlient.go
```diff
@@ -1105,6 +1113,49 @@ func (c *Client) WaitForBuildToFinish(buildName string) error {
...
+			return nil, errors.New("Timed out waiting for annotation to update")
+
+		// Each loop we check the result
+		case val, ok := <-w.ResultChan():
```
anmolbabu: Can we also check for error(EventType)?
i.e, val.Type == Error

pkg/occlient/occlient.go
```diff
@@ -1105,6 +1113,49 @@ func (c *Client) WaitForBuildToFinish(buildName string) error {
...
+
+			if !ok {
+				break
```
mik-dass: should we return here?

cdrage: Yeah, you're right.

cdrage: Actually, this is okay. Since we implement the timeout anyways. Something it will return an error when the DC / annotation isn't quite ready yet.

pkg/occlient/occlient.go
```diff
@@ -57,6 +57,8 @@ import (
 )
 
 const (
+	ocRequestTimeout   = 1 * time.Second
```
golangcibot: `ocRequestTimeout` is unused

pkg/occlient/occlient.go
mik-dass: Ok :+1:

pkg/occlient/occlient_test.go
```diff
@@ -4252,3 +4281,72 @@ func Test_findContainer(t *testing.T) {
...
+			go func() {
+				fkWatch.Modify(&tt.args.dc)
+			}()
+			fakeClientSet.AppsClientset.PrependWatchReactor("deploymentconfigs", func(action ktesting.Action) (handled bool, ret watch.Interface, err error) {
+				return true, fkWatch, nil
+			})
+			// Run function WaitAndGetDC
+			_, err := fakeClient.WaitAndGetDC(tt.args.name, tt.args.annotation, tt.args.value, tt.args.timeout)
+			// Error checking WaitAndGetDC
+			if !tt.wantErr == (err != nil) {
```
mik-dass: I guess, this condition is already satisfied in the last two block?

cdrage:
Added the comment! :+1: thanks everyone. Since it's 2 ACK's / LGTM I'll merge after this is green.

mik-dass:
@cdrage I guess we still need to take care of https://github.com/redhat-developer/odo/pull/848#discussion_r229790912 or am I missing something?

kadel:
@cdrage please rebase this with current master

cdrage:
The failing test is a false positive. The quick succession of: `./odo create wildfly wildfly --binary sample.war && ./odo push -v 4` is causing the test to fail.. Despite the included code changes of this PR not affecting / changing anything.

surajnarwade:
@cdrage , I restarted the build to verify the PR
only one test is failing: https://travis-ci.org/redhat-developer/odo/jobs/450912699#L815

cdrage:
I've outlined why this is failing here: https://github.com/redhat-developer/odo/issues/943

PR #859, opened 2018-10-22
Completion in go

metacosm:
Provide a framework for auto-completion in go. Initial implementation only showcases completion on some commands.

In a brief.

How to test changes?
Compile then run `odo --complete`. This will install the completion handler into your shell (bash/zsh/fish) which will then delegate to `odo` itself resolution of completion options. Initially, smart completion is only provided for `odo service create` and `odo service delete`, though "dumb" completion is available for all commands and flags (just not providing any level of filtering on possible values).


main.go
```diff
@@ -1,7 +1,85 @@
...
+	if root.HasAvailableSubCommands() {
+		rootCmp.Sub = make(complete.Commands)
+		for _, c := range root.Commands() {
+			if !c.Hidden {
+				rootCmp.Sub[c.Name()] = createCompletion(c)
+			}
+		}
+	}
+
+	/*if s, ok := root.Annotations[cmd.OdoCompleteArgsAnnotation]; ok {
```
geoand: Should this block be removed?

geoand:
@metacosm Should the description of the PR perhaps be updated to reflect that this is a showcase for a few commands?

FWIW, I believe that this approach is better than what we have with the bash scripts :)

main.go
metacosm: Good point.

surajnarwade:
@metacosm , couple of questions:

* can we detect shell and modify accordingly ?

* it seems you are using some new library for this purpose, but I can't see any changes to `glide.yml`

* how it is different than cobra CLI autocompletion ?

metacosm:
> * can we detect shell and modify accordingly ?

Not sure I understand your question.

> * it seems you are using some new library for this purpose, but I can't see any changes to `glide.yml`

Fixed. Thanks!

> * how it is different than cobra CLI autocompletion ?

cobra requires that you provide shell functions to implement the autocompletion. In this implementation, only a minimal hook in your shell profile is required and then your shell will delegate autocompletion to `odo` itself so autocompletion code can be written in Go and use all the `odo` codebase / functions to perform its work. No more need to maintain shell functions in parallel. I believe this is a vast improvement both in terms of maintainability and ease of development.

surajnarwade:
@metacosm , I assume it's not using completion bash provided by cobra. so we can remove `cmd/completion.go` then

metacosm:
Actually, I'll remove `completion.go` once all the completions covered by that file are re-implemented using this new mechanism. Assuming also that people agree that this new implementation is the way to go as well.

surajnarwade:
Ideally odo help should look like this, (as per master)

```
 $ odo -h
Odo (OpenShift Do) is a CLI tool for running OpenShift applications in a fast and automated matter. Odo reduces the complexity of deployment by adding iterative development without the worry of deploying your source code.

Find more information at https://github.com/redhat-developer/odo

Usage:
  odo [command]

Examples:
  # Creating and deploying a Node.js project
  git clone https://github.com/openshift/nodejs-ex && cd nodejs-ex
  odo create nodejs
  odo push

  # Accessing your Node.js component
  odo url create

Component Commands:
  component   Components of application.
  create      Create a new component
  delete      Delete an existing component
  describe    Describe the given component
  link        Link component to a service
  list        List all components in the current application
  log         Retrieve the log for the given component.
  push        Push source code to a component
  update      Update the source code path of a component
  watch       Watch for changes, update component on change

Other Commands:
  app         Perform application operations
  catalog     Catalog related operations
  project     Perform project operations
  service     Perform service catalog operations
  storage     Perform storage operations
  url         Expose component to the outside world

Utility Commands:
  help        Help about any command
  utils       Utilities for completion, terminal commands and modifying Odo configurations
  version     Print the client version information

Flags:
  -h, --help                             help for odo

Use "odo [command] --help" for more information about a command.
```

as per this PR, it;s showing,

```
$ odo -h
Usage of odo:
  -alsologtostderr
    	log to standard error as well as files
  -complete
    	Install completion for odo command
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -uncomplete
    	Uninstall completion for odo command
  -v value
    	log level for V logs
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
  -y	Don't prompt user for typing 'yes' when installing completion
```

cc @metacosm

surajnarwade:
```
$ odo service create 
cakephp-mysql-persistent  dh-hello-world-apb        dh-mediawiki-apb          dh-pyzip-demo-apb         django-psql-persistent
dancer-mysql-persistent   dh-hello-world-db-apb     dh-mongodb-apb            dh-pyzip-demo-db-apb      jenkins-ephemeral
dh-awx-apb                dh-homeassistant-apb      dh-mssql-apb              dh-rds-postgres-apb       jenkins-pipeline-example
dh-eclipse-che-apb        dh-import-vm-apb          dh-mssql-remote-apb       dh-rhpam-apb              mariadb-persistent
dh-es-apb                 dh-jenkins-apb            dh-mysql-apb              dh-thelounge-apb          mongodb-persistent
dh-etherpad-apb           dh-keycloak-apb           dh-nginx-apb              dh-tiller-apb             mysql-persistent
dh-galera-apb             dh-kibana-apb             dh-postgresql-apb         dh-unifi-controller-apb   nodejs-mongo-persistent
dh-gluster-s3-apb         dh-manageiq-apb           dh-prometheus-apb         dh-vnc-desktop-apb        postgresql-persistent
dh-hastebin-apb           dh-mariadb-apb            dh-proxy-config-apb       dh-wordpress-ha-apb       rails-pgsql-persistent
```

```
$ odo service delete 
postgresql-persistent   rails-pgsql-persistent  
```

Both commands autocompletion working for me on bash :+1:

surajnarwade:
@metacosm ,

```
$ odo --complete
Install completion for odo? y
Installing...
Install failed! 2 errors occurred:
	* already installed in /home/snarwade/.bashrc
	* already installed at /home/snarwade/.config/fish/completions/odo.fish
```

can you please modify this message rather than `failed`

metacosm:
@surajnarwade this is outside of my control (well, unless we want to fork the underlying libray) as this message is issued by the library I use.

metacosm:
> can you please modify this message rather than `failed`

It's outside of my control (barring a change in the underlying library) since the message is issued by the library I use for the completion support.

> * it will be awesome if we check shell type and install accordingly, or at least let user give input,
>   for example,
> 
> ```
> odo  --complete bash
> odo --complete fish
> ```

Not sure that is needed since I think the tool will automatically install the proper command for the proper shell.

kadel:
It looks great. 
The code can look a little little bit intimidating at first look, but it makes sense once you start orient in it.
Writing completions in go will be much better than in bash.

@metacosm what is your plan with this? Merge it and use it as a template for the rest of the commands?

geoand:
If the plan is to break up the commands, I can obviously help out :)

metacosm:
@kadel yes, I think this should be merged and I will incorporate it in the #871 work to see how it works out. I'm not sure we should be creating more completion handlers at this stage, though maybe we could put all of them in a `completion` package so that they can be moved all at once? I will also document the architecture, how it works and explain how one could create a new completion handler. Any idea where that document should live?

PR #865, opened 2018-10-23
Add some simple tests for Java

cdrage:
Adds some starter / simple tests for Java in it's own file.

Doesn't close https://github.com/redhat-developer/odo/issues/777
completely, but it at least adds a start.

### DIFF
--- a/tests/e2e/java_test.go
+++ b/tests/e2e/java_test.go
@@ -1,0 +1,100 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const files = "examples/java/"
+
+var _ = Describe("odoJavaE2e", func() {
+	var t = "java"
+	var projName = fmt.Sprintf("odo-%s", t)
+	var gitRepo = "https://github.com/openshift-evangelists/Wild-West-Backend"
+
+	// Create a separate project for Java
+	Context("create java project", func() {
+		It("should create a new java project", func() {
+			session := runCmd("odo project create " + projName)
+			Expect(session).To(ContainSubstring(projName))
+		})
+	})
+
+	// Test Java
+	Context("odo component creation", func() {
+
+		It("Should be able to deploy the git repo", func() {
+
+			// Deploy the git repo / wildfly example
+			runCmd("odo create wildfly git-test --git " + gitRepo)
+			cmpList := runCmd("odo list")
+			Expect(cmpList).To(ContainSubstring("git-test"))
+
+			// Push changes
+			runCmd("odo push")
+
+			// Create a URL
+			runCmd("odo url create")
+			getRoute := runCmd("odo url list  | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
+			getRoute = strings.TrimSpace(getRoute)
+
+			// Ping said URL
+			waitForEqualCmd("curl -s "+getRoute+" | grep 'Welcome to WildFly' | wc -l | tr -d '\n'", "2", 10)
+
+			// Delete the component
+			runCmd("odo delete git-test -f")
+		})
+
+		It("Should be able to deploy a .jar file", func() {
+			runCmd("odo create wildfly jar-test --binary " + files + "wildwest-1.0.jar")
+			cmpList := runCmd("odo list")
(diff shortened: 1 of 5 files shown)


cdrage:
As you can tell, this is a very *minor* start for the tests.  I'll obviously add some more in a separate PR. 

Just wanted to push this one to see if we're on the same page.

Since Java is one of the first-class citizens, my thoughts were:

 - Adding the binaries to the e2e directory for testing
 - Creating it's own separate file for actual tests / expanding on it.

Do you agree @cmoulliard @kadel ?

tests/e2e/java_test.go
```diff
@@ -0,0 +1,60 @@
...
+			session := runCmd("odo project create " + projName)
+			Expect(session).To(ContainSubstring(projName))
+		})
+	})
+
+	// Test Java
+	Context("odo component creation", func() {
+
+		It("Should be able to deploy a .jar file", func() {
+			runCmd("odo create wildfly jar-test --binary " + files + "wildwest-1.0.jar")
```
surajnarwade: can we download binary if possible ? just saying because it's increasing repo size

cdrage: If we do that, then each time the test is ran it'd waste bandwith :-1: 

What do you think @kadel ? Should we keep the binaries in the repo or download externally?

anmolbabu: +1 for maintaining inside our repo or atleast host them in a reliable trusted place.
We shouldn't ideally rely on external binaries in IMHO anyway I do not think Odo binary size would increase without addition of more go sources(from what I gathered from @kadel during our conversation around why it was a challenge for packaging config file as a separate non-go text file with our source with all default values...)

But we could as well use raw git url to fetch binaries from our own repo as well(If absolutely required)

tests/e2e/java_test.go
```diff
@@ -0,0 +1,60 @@
...
+		})
+
+		It("Should be able to deploy a .war file", func() {
+			runCmd("odo create wildfly war-test --binary " + files + "wildwest-1.0.war")
+			cmpList := runCmd("odo list")
+			Expect(cmpList).To(ContainSubstring("war-test"))
+
+			// Push changes
+			runCmd("odo push")
+
```
kadel: there should be  a check to verify that the application is running correctly

cdrage:
This is a really stupid question..

But with https://github.com/openshift-evangelists/Wild-West-Backend

It *really* shouldn't just show "Welcome to WildFly" correct? Shouldn't it be showing the "game" that's included?

cmoulliard:
You must specify in your curl/http query which endpoint you will call as multiple are available https://goo.gl/SWJwTi @cdrage

cdrage:
@cmoulliard Is there no default "index page" for Wild West?

tests/e2e/java_test.go
```diff
@@ -0,0 +1,100 @@
...
+			cmpList := runCmd("odo list")
+			Expect(cmpList).To(ContainSubstring("git-test"))
+
+			// Push changes
+			runCmd("odo push")
+
+			// Create a URL
+			runCmd("odo url create")
+			getRoute := runCmd("odo url list  | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
+			getRoute = strings.TrimSpace(getRoute)
```
golangcibot: ineffectual assignment to `getRoute`

surajnarwade:
CI passed :tada:

tests/e2e/java_test.go
kadel: let's keep it in our repo, it will eliminate flakes due to timeouts etc..

tests/e2e/java_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+			// Ping said URL
+			//waitForEqualCmd("curl -s "+getRoute+" | grep 'Welcome to WildFly' | wc -l | tr -d '\n'", "2", 10)
```
kadel: why is this commented out?

cdrage:
Making this a **work in progress** due to Java not working correctly at the moment, this gives a *false positive* as you are **NOT** suppose to see the "Welcome to Wildfly" page.. Despite us curl'ing for it.

cdrage:
@geoand If you don't mind picking this up that'd be great :+1:

geoand:
I will do just that :)

geoand:
Opened a new PR: #951

PR #867, opened 2018-10-23
Add test for findContainer

cdrage:
Adds a test for findContainer.

Closes 1/3 of https://github.com/redhat-developer/odo/issues/828

### DIFF
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -3774,3 +3774,116 @@ func Test_updateEnvVar(t *testing.T) {
 		})
 	}
 }
+
+func Test_findContainer(t *testing.T) {
+	type args struct {
+		name       string
+		containers []corev1.Container
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "Case 1 - Find the container",
+			args: args{
+				name: "foo",
+				containers: []corev1.Container{
+					{
+						Name: "foo",
+						VolumeMounts: []corev1.VolumeMount{
+							{
+								MountPath: "/tmp",
+								Name:      "test-pvc",
+							},
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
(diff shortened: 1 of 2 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -3774,3 +3774,122 @@ func Test_updateEnvVar(t *testing.T) {
...
+func Test_findContainer(t *testing.T) {
+	type args struct {
+		name       string
+		containers []corev1.Container
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+		actions int
```
golangcibot: field `actions` is unused

geoand:
The test failure seems unrelated.
I also saw the same exact thing in another PR, looks like that failing test is flacky

surajnarwade:
@geoand , rebase with master, test will pass :smile:  :vulcan_salute:

pkg/occlient/occlient_test.go
```diff
@@ -3774,3 +3774,121 @@ func Test_updateEnvVar(t *testing.T) {
...
+			// Run function findContainer
+			container, err := findContainer(tt.args.containers, tt.args.name)
+
+			// Check that the container matches the name
+			if err == nil && container.Name != tt.args.name {
+				t.Errorf("Wrong container returned, wanted container %v, got %v", tt.args.name, container.Name)
+			}
+
+			// Error checking findContainer
+			if !tt.wantErr == (err != nil) {
```
mik-dass: This condition is already satisfied by the below two conditions, so the error is thrown twice (once by this condition and the other from below
```
--- FAIL: Test_findContainer/Case_1_-_Find_the_container (0.00s)
        occlient_test.go:3883:  client.findContainer() unexpected error <nil>, wantErr true
        occlient_test.go:3887: test failed, expected: false, got true
```

cdrage:
Ready for another review @mik-dass @surajnarwade @sgk

PR #868, opened 2018-10-24
Split env vars only by the first `=` to allow for base64-encoded values

joshainglis:

In order to allow base64 encoded environment variables, only the first encountered `=` is used for the string split. This has the side-effect of reading in `key1=value1=value2` as `{"key1": "value1=value2"}`, but does allow, for example, ssh keys to be passed in as env vars.

No

How to test changes?
make test


mik-dass:
Merging based on 3 ACKS

PR #870, opened 2018-10-24
Remove "cli structure" information from README.md

cdrage:
This information is already available within: https://github.com/redhat-developer/odo/blob/master/docs/cli-reference.md

We remove this from the README.md in favour of the above.


README.md
```diff
@@ -139,65 +139,6 @@ Additional documentation can be found below:
...
-        config : Modifies configuration settings
-            set : Set a value in odo config file
-            view : View current configuration values
-        terminal : Add Odo terminal support to your development environment
-    version : Print the client version information
-    watch : Watch for changes, update component on change
-
-*_autogenerated_
-
+- Service: A service will typically be a database or a "service" a component links / depends on. For example: MariaDB, Jenkins, MySQL
```
surajnarwade: **Service**
(make it bold)

surajnarwade: which is coming from service catalog (service catalog needs to be enabled on cluster to use this feature)

cdrage:
@surajnarwade Updated, thanks!

PR #878, opened 2018-10-26
Add timeout to pod waiting function

### LINKED ISSUE #852: WaitAndGetPod should have a timeout
Note: WaitAndGetPod does not have a timeout, it will run endlessly until the "Pod" is actually running.

See: https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L1109

geoand:
Fix #852

#852

PR #855 also attempted to address the issue, but the tests were failing which lead me to open this PR

How to test changes?
It's difficult to actually verify the timeout is working. You could always reduce it (to like 5 second) and test

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -68,10 +68,10 @@ const (
 
 	// Create a custom name and (hope) that users don't use the *exact* same name in their deployment
 	supervisordVolumeName = "odo-supervisord-shared-data"
-)
 
-// WaitForPodTimeOut controls how long we should wait for a pod before giving up
-var WaitForPodTimeOut = 60 * time.Second
+	// waitForPodTimeOut controls how long we should wait for a pod before giving up
+	waitForPodTimeOut = 60 * time.Second
+)
 
 // errorMsg is the message for user when invalid configuration error occurs
 const errorMsg = `
@@ -1157,8 +1157,8 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 		return val, nil
 	case err := <-watchErrorChannel:
 		return nil, err
-	case <-time.After(WaitForPodTimeOut):
-		return nil, errors.Errorf("waited %s but couldn't find pod matching '%s' sel", WaitForPodTimeOut, selector)
+	case <-time.After(waitForPodTimeOut):
+		return nil, errors.Errorf("waited %s but couldn't find pod matching '%s' sel", waitForPodTimeOut, selector)
 	}
 }
 
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1123,9 +1123,6 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 	podChannel := make(chan *corev1.Pod)
 	watchErrorChannel := make(chan error)
 
-	defer close(podChannel)
-	defer close(watchErrorChannel)
-
 	go func() {
 	loop:
 		for {
@@ -1150,6 +1147,8 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 				break loop
 			}
 		}
+		close(podChannel)
+		close(watchErrorChannel)
 	}()
 
 	select {
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1157,7 +1157,7 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 	case err := <-watchErrorChannel:
 		return nil, err
 	case <-time.After(waitForPodTimeOut):
-		return nil, errors.Errorf("waited %s but couldn't find pod matching '%s' sel", waitForPodTimeOut, selector)
+		return nil, errors.Errorf("waited %s but couldn't find running pod matching selector: '%s'", waitForPodTimeOut, selector)
 	}
 }
 
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1123,6 +1123,9 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 	podChannel := make(chan *corev1.Pod)
 	watchErrorChannel := make(chan error)
 
+	defer close(podChannel)
+	defer close(watchErrorChannel)
+
 	go func() {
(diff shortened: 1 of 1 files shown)


geoand:
@cdrage I am guessing that the timeout should be configurable. 
What do you propose in order to achieve that?

pkg/occlient/occlient.go
```diff
@@ -1116,23 +1119,44 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
...
-				glog.V(4).Infof("Pod %s is running.", e.Name)
-				return e, nil
-			case corev1.PodFailed, corev1.PodUnknown:
-				return nil, errors.Errorf("pod %s status %s", e.Name, e.Status.Phase)
+
+	podChannel := make(chan *corev1.Pod)
+	watchErrorChannel := make(chan error)
+
+	go func() {
+	loop:
```
anmolbabu: I understand the requirement of this is to jump directly out from within the switch and for but because the labels hold the potential to make the code cryptic I personally have a strong dislike for them :(

But feel free to ignore it because labels are also an integral part of code ;) and they deserve equality ;)

pkg/occlient/occlient.go
```diff
@@ -1116,23 +1119,44 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
...
-			glog.V(4).Infof("Status of %s pod is %s", e.Name, e.Status.Phase)
-			switch e.Status.Phase {
-			case corev1.PodRunning:
-				glog.V(4).Infof("Pod %s is running.", e.Name)
-				return e, nil
-			case corev1.PodFailed, corev1.PodUnknown:
-				return nil, errors.Errorf("pod %s status %s", e.Name, e.Status.Phase)
+
+	podChannel := make(chan *corev1.Pod)
+	watchErrorChannel := make(chan error)
```
anmolbabu: Doesn't seem like we are closing these channels? Or I am missing something

pkg/occlient/occlient.go
```diff
@@ -1116,23 +1119,44 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
...
+
+	go func() {
+	loop:
+		for {
+			val, ok := <-w.ResultChan()
```
anmolbabu: I was thinking if the podchannel and watchErrorChannel can be avoided if we move the logic of listening to the w.ResultChan to the select below?

But it seems w.ResultChan will get updates for every update to pod's attributes... and that would cause select to complete :(

pkg/occlient/occlient.go
geoand: I have the same opinion as well... I really don't like the label but I think that the alternative of having various control flags would make things even weirder.

I am totally on the fence with this, don't really know what to do :P

pkg/occlient/occlient.go
geoand: Yes exactly.

anmolbabu:
> @cdrage I am guessing that the timeout should be configurable.
> What do you propose in order to achieve that?

@geoand +1 for it to be configurable with a safe default(if not configured)

pkg/occlient/occlient.go
geoand: Good catch! I will take care of it

geoand:
@anmolbabu How do you propose this value be configurable? Some global flag?

anmolbabu:
> @anmolbabu How do you propose this value be configurable? Some global flag?

@geoand Yes, something like `cmd_timeout` in the configuration file . ..

pkg/occlient/occlient.go
geoand: I have introduced a fix, but I am not 100% sure it's correct since I am still new to go channels. Can you check it out please?

Thanks!

kadel:
> @geoand Yes, something like `CmdTimeout` in the configuration file . ..
> But I would like to take @kadel @jorgemoralespou @cdrage opinion before starting on it

I think that for now it is ok to have this in code.
Lets add is as config option only if it is really needed.

pkg/occlient/occlient.go
```diff
@@ -70,6 +70,9 @@ const (
 	supervisordVolumeName = "odo-supervisord-shared-data"
 )
 
+// WaitForPodTimeOut controls how long we should wait for a pod before giving up
+var WaitForPodTimeOut = 60 * time.Second
```
kadel: I think that it would be better to have this as `const` in block above

geoand: Fixed

geoand:
@anmolbabu @kadel Code updated with the latest fixes to the comments

geoand:
One more review on this would be much appreciated :) 

Thanks

geoand:
@anmolbabu I'll wait until you have verified it :)

geoand:
@cdrage I'll look into it. Does the PR work as expected with more sane timeouts?

cdrage:
@geoand Yes, more sane timeout's it works fine! I'm just trying to purposely break things haha.

geoand:
@cdrage Actually it's great that you uncovered this! 

This is indeed a bug and seems to have to do with the way the channel is being closed. I need to look into this further

geoand:
@cdrage Pushed a fix. Can you please try again?

pkg/occlient/occlient.go
```diff
@@ -1116,23 +1119,46 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
...
+		close(watchErrorChannel)
+	}()
+
+	select {
+	case val := <-podChannel:
+		return val, nil
+	case err := <-watchErrorChannel:
+		return nil, err
+	case <-time.After(waitForPodTimeOut):
+		return nil, errors.Errorf("waited %s but couldn't find pod matching '%s' sel", waitForPodTimeOut, selector)
```
cdrage: The output is a bit odd:

```sh
waited 1s but couldn't find pod matching 'deploymentconfig=wild-west-ba-wildfly-levs-odo-qnvs' sel
```

It adds "sel" at the end?

geoand: I thought I got rid of that... Will fix now!

geoand: Should be fixed now

cdrage:
@geoand Yup! This LGTM.

geoand:
Thanks @cdrage!

PR #879, opened 2018-10-26
Fixed project deletion

surajnarwade:

This will delete project only if `y` is provided and abort for any other letter

https://github.com/redhat-developer/odo/issues/877

Please describe the steps to test the PR

* `odo project delete p1` and give y or any random letter


cmd/project.go
```diff
@@ -138,27 +138,29 @@ var projectDeleteCmd = &cobra.Command{
 
 		if strings.ToLower(confirmDeletion) != "y" {
 			fmt.Printf("Aborting deletion of project: %v\n", projectName)
-		}
+			os.Exit(1)
+		} else {
```
geoand: I would suggest that you drop the `else` since execution of that block will only proceed if condition on `if` doesn't pass (due to your addition of `os.Exit(1)`).
This would also make reviewing the PR easier since the formatting of the line would not change :)

PR #881, opened 2018-10-26
[do-not-merge] Switch to golang/dep with prune packages

vdemeester:
Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

Switching from `glide` to `golang/dep` (which is "the tool of choice" while waiting for "go modules", in go 1.11 and later). This also enables packages pruning, reducing drastically the vendored code.

Nop it's just a proposal, open for discussion — One argument against this PR would be to just switch to `go mod` when ready (and thus removing the need for vendoring altogether).

How to test changes?

`make` :angel:


vdemeester:
@metacosm it would indeed :angel: but this made an hard requirement on `go` 1.11, which… I didn't know if this is something `odo` can do or not. I'll happily switch to `go modules` if maintainers feel like it :wink:

kadel:
modules are still experimental feature even in 1.11. 
We are vendoring OpenShift that uses forked Kubernetes, this can get quite messy. 
Unless we can be pretty sure that it is going to work with OpenShift I would stick with glide.

Issue #880, opened 2018-10-26
Enhance the odo service describe


geoand:
[kind/Enhancement]

## What should odo do differently? Which functionality do you think we should change?

With the inclusion of #756 we now have a lot of information about the parameters of a service. For instance we now know the parameters that can be used when creating the `ServiceInstance`.
 
There is one issue however that has arisen from that implementation: 
The `required` parameters that are shown are not always strictly "required", since most of them have default values. 
For example, the `postgresql_password` service only really required the `postgresql_password` since the rest of the parameters are reported as required, have default values (for example we have `postgresql_user` defaults to `admin`).

If we don't address this, there two implications:
* User's will be mislead into thinking they need to supply all the parameters when in most cases they only need a few of them
* More importantly, this will prohibit us from providing the correct suggestions to users of the service catalog in the proposed interactive mode. Furthermore, knowing which parameters are really required will allow us to fail fast and with the appropriate error message when service create is called without the necessary parameters.

## What are you trying to do?

Improve the experience of the users of the service catalog since currently it's not crystal clear from the user what parameters need to be supplied and what their values look like. 
Addressing this issue will allow us to build better interactive mode and provide better error messages during failed service creation while also improving the describe output.

jorgemoralespou:
@geoand I've seen that you created the issue and then created a PR, but TBH the issue does provide very little information on what is the problem and how you're going to approach the solution, so seems that you went ahead with an implementation before getting any consensus.

I haven't looked into it, and I'm sure it's really good, but I'm raising this topic as I would think that it's better to spend some time defining in the issue how things are going to be laid out, so one knows before the PR if there's any outstanding concern to be discussed.

Not sure if this has been discussed on mattermost or any of our internal calls, but I think that open discussions or at least, openly publish the intents helps. For many people, issues is the only thing they see, if as me, they don't understand code ;-)

Does it make sense? @kadel ???

geoand:
@jorgemoralespou I apologize if I broke the standard process here!

The issue was brought up in the some of the odo discussions we had on BlueJeans last week but I admit that could have done a better done at defining the issue.

The PR itself doesn't change much of the functionality that the user sees, but it instead provides the building blocks to make subsequent improvements to the service catalog handing. 
One reason why I didn't wait for further consensus is that I believed that the problem was best demonstrated by actually showing the fix, which was also quite small. 

I apologize again for straying from the norms here ☺

jorgemoralespou:
[...]

PR #886, opened 2018-10-29
Enhance the odo service describe

geoand:

We now include all the information about parameters of a service plan in the internal data structures.
The information is displayed to the user via `odo catalog describe service ...` but could definitely use an improvement in that area :)

#880 

How to test changes?

Have the service catalog installed in the cluster you are using. Use any service in
 `odo catalog describe service ...` to view the output.

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -233,15 +233,47 @@ This describes the service and the associated plans.
 
 		if len(plans) > 0 {
 			table.Append([]string{"PLANS"})
+
 			for _, plan := range plans {
-				table.Append([]string{".......................", "....................................................."})
+
+				// create the display values for required  and optional parameters
+				requiredWithMandatoryUserInputParameterNames := []string{}
+				requiredWithOptionalUserInputParameterNames := []string{}
+				optionalParameterDisplay := []string{}
+				for _, parameter := range plan.Parameters {
+					if parameter.Required {
+						// until we have a better solution for displaying the plan data (like a separate table perhaps)
+						// this is simplest thing to do
+						if parameter.HasDefaultValue {
+							defaultValueStr := fmt.Sprintf("%v", parameter.DefaultValue)
+							requiredWithOptionalUserInputParameterNames = append(
+								requiredWithOptionalUserInputParameterNames,
+								fmt.Sprintf("%s (default: '%s')", parameter.Name, defaultValueStr))
+						} else {
+							requiredWithMandatoryUserInputParameterNames = append(requiredWithMandatoryUserInputParameterNames, parameter.Name)
+						}
+
+					} else {
+						optionalParameterDisplay = append(optionalParameterDisplay, parameter.Name)
+					}
+				}
+
+				table.Append([]string{"***********************", "*****************************************************"})
+				planLineSeparator := []string{"-----------------", "-----------------"}
 
 				planData := [][]string{
 					{"Name", plan.Name},
+					planLineSeparator,
 					{"Display Name", plan.DisplayName},
+					planLineSeparator,
 					{"Short Description", plan.Description},
-					{"Required Parameters", strings.Join(plan.Required, ", ")},
-					{"Optional Parameters", strings.Join(plan.Optional, ", ")},
+					planLineSeparator,
+					{"Required Params without a default value", strings.Join(requiredWithMandatoryUserInputParameterNames, ", ")},
+					planLineSeparator,
+					{"Required Params with a default value", strings.Join(requiredWithOptionalUserInputParameterNames, ", ")},
+					planLineSeparator,
+					{"Optional Params", strings.Join(optionalParameterDisplay, ", ")},
+					{"", ""},
 				}
 				table.AppendBulk(planData)
 			}
--- a/pkg/service/service_test.go
+++ b/pkg/service/service_test.go
@@ -10,8 +10,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ktesting "k8s.io/client-go/testing"
 	"reflect"
+	"sort"
 	"testing"
 )
+
+// M is an alias for map[string]interface{}
+type M map[string]interface{}
 
 func fakePlanExternalMetaDataRaw() ([][]byte, error) {
 	planExternalMetaData1 := make(map[string]string)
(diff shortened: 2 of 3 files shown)


cmd/catalog.go
```diff
@@ -233,15 +233,37 @@ This describes the service and the associated plans.
...
+				requiredParameterNames := []string{}
+				optionalParameterDisplay := []string{}
+				for _, parameter := range plan.Parameters {
+					if parameter.Required {
+						requiredParameterNames = append(requiredParameterNames, parameter.Name)
+					} else {
+						// until we have a better solution for displaying the plan data (like a separate table perhaps)
+						// this is simplest thing to do
+						defaultValueStr := fmt.Sprintf("%v", parameter.DefaultValue)
+						if len(defaultValueStr) > 0 {
```
metacosm: This causes an issue: `mysql-persistent` has an empty string as default value for several properties (e.g. `MYSQL_PASSWORD`) which appear here as not having a default value. So I think `ServicePlanParameter` should record that information so that more accurate information is displayed.

geoand: Fixed

geoand:
The more I think about it, the more I believe that `required` should only be used as a designation when the user actually needs to provide a value.

metacosm:
> The more I think about it, the more I believe that `required` should only be used as a designation when the user actually needs to provide a value.

I am of the opposite opinion… 😇
I think we should provide accurate information: a property that is required with a default value is still required and, as such, shouldn't be marked as optional. Especially because in some cases, the default values should really be changed (e.g. user / password fields) even if there are default values. How we deal with these properties is different from how we display the information. The displayed information should be accurate. We should somehow highlight required properties without default values (which are technically the only ones that the user *must* provide a value for).

metacosm:
Maybe we shouldn't call them optional or required but rather something like 3 categories: required with required user input, required with optional user input, optional?

geoand:
I don't really like having `required` twice in the same phrase :P

geoand:
@metacosm PR updated, let me know what you think :)

metacosm:
@geoand can we add separators between rows?

geoand:
@metacosm Great idea! Added them

metacosm:
@geoand sorry to be a pain but can you add the separators for the other fields as well? Also, seems like there's a regression back to the state where empty default values result in a property being marked as optional.

geoand:
Yeah, all add them to all fields, makes sense :)

I'll look into the regression you mention and see what I can find

cmd/catalog.go
```diff
@@ -233,15 +233,47 @@ This describes the service and the associated plans.
...
 				planData := [][]string{
 					{"Name", plan.Name},
+					planLineSeparator,
 					{"Display Name", plan.DisplayName},
+					planLineSeparator,
 					{"Short Description", plan.Description},
-					{"Required Parameters", strings.Join(plan.Required, ", ")},
-					{"Optional Parameters", strings.Join(plan.Optional, ", ")},
+					planLineSeparator,
+					{"Required Params with Mandatory User Input", strings.Join(requiredWithMandatoryUserInputParameterNames, ", ")},
```
kadel: Could we just call it "Required Params with a default value"? and "Required Params without a default value"?
To me it sounds a bit better than "Mandatory/Optional User Input"

geoand: I have no strong opinion on this, so I don't mind either way :) @metacosm WDYT?

metacosm: I'm fine with it but the idea was to make it clear(er) which values actually require user input and which don't.

geoand: Fixed

cmd/catalog.go
```diff
@@ -233,15 +233,47 @@ This describes the service and the associated plans.
...
+						} else {
+							requiredWithMandatoryUserInputParameterNames = append(requiredWithMandatoryUserInputParameterNames, parameter.Name)
+						}
+
+					} else {
+						optionalParameterDisplay = append(optionalParameterDisplay, parameter.Name)
+					}
+				}
+
 				table.Append([]string{".......................", "....................................................."})
```
mik-dass: I think we should replace the .... with **** or something distinguishable to denote the start of a new plan. WDYT?

cmd/catalog.go
```diff
@@ -233,15 +233,47 @@ This describes the service and the associated plans.
...
+					planLineSeparator,
 					{"Display Name", plan.DisplayName},
+					planLineSeparator,
 					{"Short Description", plan.Description},
-					{"Required Parameters", strings.Join(plan.Required, ", ")},
-					{"Optional Parameters", strings.Join(plan.Optional, ", ")},
+					planLineSeparator,
+					{"Required Params with a default value", strings.Join(requiredWithMandatoryUserInputParameterNames, ", ")},
+					planLineSeparator,
+					{"Required Params without a default value", strings.Join(requiredWithOptionalUserInputParameterNames, ", ")},
```
mik-dass: Shouldn't it be "Required Params with a default value"?

cmd/catalog.go
```diff
@@ -233,15 +233,47 @@ This describes the service and the associated plans.
...
 				planData := [][]string{
 					{"Name", plan.Name},
+					planLineSeparator,
 					{"Display Name", plan.DisplayName},
+					planLineSeparator,
 					{"Short Description", plan.Description},
-					{"Required Parameters", strings.Join(plan.Required, ", ")},
-					{"Optional Parameters", strings.Join(plan.Optional, ", ")},
+					planLineSeparator,
+					{"Required Params with a default value", strings.Join(requiredWithMandatoryUserInputParameterNames, ", ")},
```
mik-dass: And this ""Required Params without a default value""?

pkg/service/service_test.go
```diff
@@ -167,13 +167,11 @@ func TestGetServiceClassAndPlans(t *testing.T) {
 					Name:        "dev",
 					Description: "this is a example description 1",
 					DisplayName: "plan-name-1",
-					Required:    []string{"PLAN_DATABASE_URI", "PLAN_DATABASE_USERNAME", "PLAN_DATABASE_PASSWORD"},
```
mik-dass: Please modify the tests according to the new changes :)

cmd/catalog.go
geoand: No, that's the previous line :)

cmd/catalog.go
geoand: Let me check this again and get back to you

pkg/service/service_test.go
geoand: I plan on doing that yes :+1:

cmd/catalog.go
mik-dass: I mean currently it shows
```
 Required Params with a default  | mysql_password                                         
  value                          |                                                        
  -----------------              | -----------------                                      
  Required Params without a      | volume_size (default: '1Gi'),                          
  default value                  | mysql_database (default:                               
                                 | 'devel'), mysql_user (default:                         
                                 | 'devel'), mysql_version                                
                                 | (default: '5.7')    
```
volume_size has a default value of 1Gi but it's showing it in the `Required Params without a default value` section  (and the opposite for mysql_password) which is weird. Or am I missing something?

pkg/service/service_test.go
mik-dass: :+1:

cmd/catalog.go
mik-dass: :+1:

cmd/catalog.go
mik-dass: You missed changing this :D

geoand: Indeed :) I will fix it

pkg/service/service_test.go
geoand: Fixed

geoand:
@kadel @mik-dass issues raised during review have been addressed

geoand:
The test failure seems totally unrelated. @kadel @cdrage any idea what might be going on?

kadel:
> The test failure seems totally unrelated. @kadel @cdrage any idea what might be going on?

timeout :-( I've restarted it

geoand:
Thanks @

geoand:
I just saw that I have permissions to restart the build as well :P

Next time I'll do it on my own :)

geoand:
PR has been rebased so it's ready for review :)

kadel:
LGTM, but we will have to think about better formating, it looks a little bit weird.

screenshot 2018-10-30 at 18 16 27

kadel:
uff, golangci is also timeouting :-( But that is not that importing.

geoand:
@kadel Formatting is indeed bad!

I didn't bother with it to tell you the truth since AFAIK we are going to overhaul the output format for the whole project, correct?

Should I go ahead and merge?

WRT the GolangCI timeout, I've been seeing that a lot :(

PR #894, opened 2018-10-30
Vendoring openshift

surajnarwade:

This PR,
* Added openshift as a vendor
* kubernetes projects will point to openshift's forked repos

How it helps ?

* This will help me in `odo login` PR
* since, we are dealing with openshift only, it's good to point to kubernetes repos which are maintained by openshift (because they make changes and maintain in their own org)


kadel:
> What is the purpose of this change? What does it change?
> 
> Was the change discussed in an issue?
> 
> How to test changes?

no description :-(

.wwhrd.yml
```diff
@@ -16,3 +16,9 @@ exceptions:
   - github.com/davecgh/go-spew/spew/...
   # wwhrd currently does not detect ISC which is compatible with Odo so, add it under exceptions to avoid errors due to it being not recognised
   - github.com/olekukonko/tablewriter
+  # gocertifi contains CA bundle and SSL certs which is required for OpenShift
+  # This is OpenShift dependency, ODO do not use this directly
+  - github.com/certifi/gocertifi
```
kadel: what licence is it?

.wwhrd.yml
```diff
@@ -16,3 +16,9 @@ exceptions:
...
+  # go-digest contains common digest package used across the container ecosystem
+  # This is OpenShift dependency, ODO do not use this directly
+  - github.com/opencontainers/go-digest
```
kadel: what license is it?

.wwhrd.yml
surajnarwade: removed since it has MPL 2.0 license

.wwhrd.yml
surajnarwade: removed since it has apache license

metacosm:
What's the point of that change?

surajnarwade:
@metacosm , @kadel , I updated the description :smile:

.wwhrd.yml
```diff
@@ -15,4 +15,4 @@ exceptions:
   # wwhrd currently does not detect ISC which is compatible with Odo so, add it under exceptions to avoid errors due to it being not recognised
   - github.com/davecgh/go-spew/spew/...
   # wwhrd currently does not detect ISC which is compatible with Odo so, add it under exceptions to avoid errors due to it being not recognised
-  - github.com/olekukonko/tablewriter
+  - github.com/olekukonko/tablewriter
```
kadel: ?

glide.yaml
```diff
@@ -40,4 +43,9 @@ import:
   version: d4647c9c7a84d847478d890b816b7d8b62b0b279
 - package: github.com/posener/complete
   version: v1.1.2
+- package: github.com/openshift/origin
```
kadel: shouldn't `github.com/openshift/origin` and `github.com/openshift/client-go` versions match?

glide.yaml
```diff
@@ -40,4 +43,9 @@ import:
...
+  version: v3.11.0
+- package: k8s.io/kubernetes
```
kadel: why is this here?  Glide should be smart and detect shi from openshift import.

glide.yaml
kadel: we are using just one package from origin, why not just import that subpackage?

surajnarwade: @kadel , https://github.com/openshift/client-go/releases don't release regularly, so I took it from the master then :(

surajnarwade: openshift directly uses `master` from `openshift/client-go`
ref: https://github.com/openshift/origin/blob/master/glide.yaml#L42
shall I use the same ?

surajnarwade: I created issue for the same: https://github.com/openshift/client-go/issues/66

kadel: only origin master uses client-go master. If you check glide.yaml for 3.11 you can see that they are using client-go 3.11 https://github.com/openshift/origin/blob/v3.11.0/glide.yaml#L43

surajnarwade:
@kadel updated

cdrage:
@surajnarwade You need to update the commit message too :+1:

surajnarwade:
@cdrage updated with commit message

surajnarwade:
@kadel , can you please review once ?

PR #906, opened 2018-10-31
Open created url with --open

cdrage:
This function adds the --open parameter to `odo url create`.

This will open the created link on the user's default browser.

To test:

```sh
odo create nodejs
odo url create --open
```

Closes https://github.com/redhat-developer/odo/issues/905

### DIFF
--- a/cmd/url.go
+++ b/cmd/url.go
@@ -2,9 +2,11 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/redhat-developer/odo/pkg/odo/util"
 	"os"
 	"strings"
+
+	odoutil "github.com/redhat-developer/odo/pkg/odo/util"
+	"github.com/redhat-developer/odo/pkg/util"
 
 	"text/tabwriter"
 
@@ -14,6 +16,7 @@ import (
 
 var (
 	urlForceDeleteFlag bool
+	urlOpenFlag        bool
 	urlPort            int
 )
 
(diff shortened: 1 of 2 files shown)


cdrage:
False positives on the tests until #848 is fixed.

cmd/url.go
```diff
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/redhat-developer/odo/pkg/odo/util"
 	"os"
 	"strings"
 
+	"github.com/redhat-developer/odo/pkg/odo/util"
+	util2 "github.com/redhat-developer/odo/pkg/util"
```
metacosm: why util2? should we have this renamed instead? ;)

kadel: what? Why do we have two utils??
@metacosm Do I understand correctly that `github.com/redhat-developer/odo/pkg/odo/util`  is  utils for code around cmd?  If yes, than it might be good to import it as `cmdutils` or something like that, @cdrage. We have to have to have better names than just `util` and `util2`, that will be mess.

cdrage: Yeah, I was taking this inspiration from @geoand 's PR :) I'll rename it

cdrage:
@kadel 

I've renamed the "odo" utils to `odotuil` within cmd/utils.go

We should *really* move the util package or rename it to something else. Having multiple `util` packages is a bit confusing.

If you have any suggestions let me know and I'll update this PR.

PR #909, opened 2018-10-31
Add asciinema / demo gif

cdrage:
Adds a demo gif to the README.md that shows us deploying a sample nodejs
application.

Closes https://github.com/redhat-developer/odo/issues/332


cdrage:
False positives on the tests until #848 is fixed.

kadel:
Why don't you want this to be in README? I think that it looks good, it quickly shows what the experience with odo is

geoand:
@kadel For me, just to keep the README source more readable :)

cdrage:
@geoand @metacosm Despite some developers being fine with text, we've got to include some graphics / some form of visual aspect for those who learn visually rather than textually.

This gif / asciinema is being added to satisfy https://github.com/redhat-developer/odo/issues/332 as well as give a developer a quick overview.

geoand:
@cdrage The concern I have (and IIUC is @metacosm's concern as well), is whether or not the asciinema should be embedded in the doc or whether it can just point to an external asciinema URL thus keeping the doc clean.

In any case, it's very minor concern, it's doesn't have to do with the actual content of the asciinema :)

cdrage:
@geoand The reason for the gif was better integration with markdown / GitHub. I did add a preview page / link to the Asciinema but I felt like it took away from the look of the README / markdown documentation. What I could do instead is make the gif an active image/link to the Asciinema page.

geoand:
@cdrage Whatever you think is best :). I just thought it a little odd to have it embedded in the markdown, but I do see the reasoning :)

docs/img/example.cast
```diff
@@ -0,0 +1,282 @@
+{"width": 104, "timestamp": 1541016888, "env": {"TERM": "xterm-256color", "SHELL": "/usr/bin/zsh"}, "height": 25, "version": 2}
```
kadel: What is this file?
Is it used somewhere?

cdrage: It's the original "asciicast" file that I've included for historical purposes / if we need to do any further modifying to it.

README.md
```diff
@@ -19,6 +19,10 @@
 
 OpenShift Do (Odo) is a CLI tool for developers who are writing, building, and deploying applications on OpenShift. With Odo, developers get an opinionated CLI tool that supports fast, iterative development which abstracts away Kubernetes and OpenShift concepts, thus allowing them to focus on what's most important to them: code.
 
+### Demo
+
+![demo](/docs/img/example.gif)
```
surajnarwade: why we are not using asciinema link directly ?

cdrage: Because asciinema won't play automatically within Markdown. You'd have to click the link and press play twice, disrupting the user experience for the README...

metacosm: Personally, I don't mind that much but if we think it really improves thing, let's go ahead with including the gif.

Issue #872, opened 2018-10-25
'odo version` cannot get cluster url for minishift 1.26.0


dgolovin:
[kind/bug]

## What versions of software are you using?

Operating System : Windows 10 Pro
Minishift: v1.26.0+2fb32c8
OpenShift: 3.11.0

Output of `odo version` : 
```
C:\Users\dgolo\.vs-openshift\odo.exe version
odo v0.0.14 (618c6d3)
the server could not find the requested resource
```

## How did you run odo exactly?
From command line

## Expected behavior
`odo version` prints version and cluster url

## Actual behavior
`odo version' prints
`the server could not find the requested resource`
instead of cluster name

## Steps to reproduce the behavior
1. Configure and run minishift 1.26.0
`minishift start`
 
## Any logs, error output, etc?
N/A

## Do you have any idea what may have caused this?
OpenShift 3.10 - 3.11?

## Do you have an idea how to solve the issue?
Nope

geoand:
I was able to reproduce this: `odo version` works fine with OKD 3.10 but not with 3.11

dgolovin:
```
$oc version --loglevel=10
I1101 23:44:38.342914   29867 loader.go:359] Config loaded from file /home/eskimo/.kube/config
I1101 23:44:38.344191   29867 loader.go:359] Config loaded from file /home/eskimo/.kube/config
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO
I1101 23:44:38.344686   29867 round_trippers.go:386] curl -k -v -XGET  -H "Accept: application/json, */*" -H "User-Agent: oc/v1.11.0+d4cacc0 (linux/amd64) kubernetes/d4cacc0" 'https://192.168.42.33:8443/version'
I1101 23:44:38.350720   29867 round_trippers.go:405] GET https://192.168.42.33:8443/version 200 OK in 6 milliseconds
I1101 23:44:38.350736   29867 round_trippers.go:411] Response Headers:
I1101 23:44:38.350740   29867 round_trippers.go:414]     Cache-Control: no-store
I1101 23:44:38.350744   29867 round_trippers.go:414]     Content-Type: application/json
I1101 23:44:38.350747   29867 round_trippers.go:414]     Content-Length: 239
I1101 23:44:38.350750   29867 round_trippers.go:414]     Date: Fri, 02 Nov 2018 06:44:33 GMT
[...]

PR #920, opened 2018-11-02
Change version command to tolerate missing possible /version/openshift

geoand:

Change version command to tolerate missing possible /version/openshift

This is done because a cluster created with "oc cluster up" can in some
cases (as is the case for OKD 3.11) the endpoint is missing completely
and we don't want the tool to fail in such cases (following the example
set by oc)

#872

How to test changes?

Launch Minishift using OKD 3.11.
Doing `odo version` should now yield all the expected output, except the `Openshift:....` line (which is the same behavior you'll see with `oc version` as well).
When applying `odo version` against a proper Openshift cluster, or even one launched by Minishift that uses OKD 3.10, the `Openshift: ....` should be present

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -45,12 +45,18 @@ var versionCmd = &cobra.Command{
 		// Lets fetch the info about the server
 		serverInfo, err := util.GetOcClient().GetServerVersion()
 		util.CheckError(err, "")
+
+		// make sure we only include Openshift info if we actually have it
+		openshiftStr := ""
+		if len(serverInfo.OpenShiftVersion) > 0 {
+			openshiftStr = fmt.Sprintf("OpenShift: %v\n", serverInfo.OpenShiftVersion)
+		}
 		fmt.Printf("\n"+
 			"Server: %v\n"+
-			"OpenShift: %v\n"+
+			"%v"+
 			"Kubernetes: %v\n",
 			serverInfo.Address,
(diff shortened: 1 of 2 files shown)


cmd/version.go
```diff
@@ -45,12 +45,18 @@ var versionCmd = &cobra.Command{
 		// Lets fetch the info about the server
 		serverInfo, err := util.GetOcClient().GetServerVersion()
 		util.CheckError(err, "")
+
+		// make sure we only include openshift info if we actually have it
+		openshiftStr := ""
+		if len(serverInfo.OpenShiftVersion) > 0 {
```
mik-dass: Do we need to check for `serverInfo.OpenShiftVersion != nil`?

geoand: `serverInfo.OpenShiftVersion` is an empty string by default, so we just check that :)

geoand: It only get's set when the actual Openshift version was properly determined from the server

mik-dass: ok :+1:

geoand:
@mik-dass Jobs relaunched :)

geoand:
@syamgk hopefully this problem will go away then :P

mik-dass:
@geoand I guess we wait for a fix :)

geoand:
OK, I'll be waiting :joy:

geoand:
PR was rebased onto master, hopefully the tests will now pass :)

PR #871, opened 2018-10-24
Re-organize code according to kubectl/oc conventions

metacosm:
What is the purpose of this change? What does it change?
This work aims to fix #785. I'm creating this PR while the work is still ongoing so that interested parties can comment and decisions can be taken while code is still being written.

Was the change discussed in an issue?
#785

How to test changes?
If everything goes well, nothing will change externally for `odo` users so tests should work as expected.

PR #931, opened 2018-11-05
Factor out context (project, app, component) handling from commands into NewContextOptions func

metacosm:
This work starts the process of creating common options which will be needed for #871 

No

How to test changes?
`make && make test`. External behavior shouldn't change.


metacosm:
The changes in this PR are not precise enough. A better approach is needed. Closing for now.

Issue #834, opened 2018-10-12
skip-connection-check not processed by odo version


jeffmaury:
I ran odo version `--skip-connection-check` and still got long response time and the following output:

```
> odo version --skip-connection-check
odo v0.0.13 (f92c352)
Get https://192.168.99.101:8443/version/openshift: dial tcp 192.168.99.101:8443: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
```

kadel:
What error did you get?

Everytime when odo starts it checks if it can connect to OpenShift. 
`--skip-connection-check`   was added to allow skip this check when you need to run a command that you know doesn't require a connection to OpenShift ( like getting current component or application) and you don't want odo to complain about not able to reach cluster.

`odo app list` needs to get data from the cluster. So it will still complain about not able to connect to it.  You just won't get nice `Unable to connect to OpenShift cluster, is it down?` but some other ugly error.


What is your use case for `--skip-connection-check` is was originally added just to make bash completion and nice PS1 prompt easier.

syamgk:
In case of `odo version --skip-connection-check` its printing the current odo client version and after that making an HTTP request to get the server version which is failing to get response

jeffmaury:
Then I think this flag is confusing. Also would be nice to allow to get the odo version without cluster exchanges as tooling (VSCode, Developer Studio,...) could use it to check the Odo version without waiting for a costly exchange

surajnarwade:
@jeffmaury , `kubectl` has something similar which returns only client version (without talking to server)

```
kubectl version --client
```

are you talking about something similar ?

jeffmaury:
Yes the idea is for a tooling based on odo to check if the odo has the correct required version so no need to call the server

surajnarwade:
closing this since #935  is merged

PR #935, opened 2018-11-06
Added `--client` flag to `odo version`

surajnarwade:
**What is the purpose of this change? What does it change?**

This PR adds `--client` flag to `odo version` command so that,
to check only client version, we dont need to talk to server.
(Similar to `kubectl`)

**Was the change discussed in an issue?**

fixes issue #834 

**How to test changes?**

```
odo version --client
```


cmd/version.go
```diff
@@ -39,26 +42,28 @@ var versionCmd = &cobra.Command{
...
-		serverInfo, err := util.GetOcClient().GetServerVersion()
-		util.CheckError(err, "")
-		fmt.Printf("\n"+
-			"Server: %v\n"+
-			"OpenShift: %v\n"+
-			"Kubernetes: %v\n",
-			serverInfo.Address,
-			serverInfo.OpenShiftVersion,
-			serverInfo.KubernetesVersion)
+		if clientFlag != true {
```
golangcibot: should omit comparison to bool constant, can be simplified to !clientFlag

kadel: I agree with golangcibot ;-)

geoand: That makes two of us :)

PR #936, opened 2018-11-06
Using curl instead of wget

### LINKED ISSUE #933: Use curl instead of wget in e2e tests.
[kind/Enhancement]
## What should odo do differently? Which functionality do you think we should change?
Use `curl` instead of `wget` in end to end tests.

## What are you trying to do?

`wget` is not installed by default on macOS which means that the tests require an installation step first.

surajnarwade:

 replacing  wget with curl in e2e tests, because wget is not present in macOS

Resolves #933

### DIFF
--- a/tests/e2e/cmp_test.go
+++ b/tests/e2e/cmp_test.go
@@ -82,7 +82,7 @@ var _ = Describe("odoCmpE2e", func() {
 
 	Context("updating the component", func() {
 		It("should be able to create binary component", func() {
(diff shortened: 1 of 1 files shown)


geoand:
@surajnarwade Could you please rebase onto master so the tests will hopefully pass?

metacosm:
@kadel Can we merge this?

kadel:
this is just a small change, and failing test is just a flake
merging

PR #942, opened 2018-11-06
Add synchronization script between docs and website

cdrage:
This adds a script to synchronize any documentation changes between the
GitHub repository and the gh-pages branch / website.


cdrage:
Mergin' this so that we can get synchronization working :+1:

PR #944, opened 2018-11-06
Added the auto completion for components

### LINKED ISSUE #939: Suggestions for component names
parrent: #902 

Suggesting component names:
 -  `odo delete`
 -  `odo component`
 - `odo component set`
 - `odo link`
 -  `odo push`
 -  `odo watch`
 -  + every command that has `--compnent`` flag

mik-dass:
fixes #939

### DIFF
--- a/pkg/odo/util/completion/completion.go
+++ b/pkg/odo/util/completion/completion.go
@@ -160,16 +160,22 @@ func getUserTypedCommands(args complete.Args, command *cobra.Command) []string {
 	return commands
 }
 
+// AddProjectFlag adds a `project` flag to the given cobra command
+// Also adds a completion handler to the flag
 func AddProjectFlag(cmd *cobra.Command) {
 	cmd.Flags().String(genericclioptions.ProjectFlagName, "", "Project, defaults to active project")
 	RegisterCommandFlagHandler(cmd, "project", ProjectNameCompletionHandler)
 }
 
+// AddComponentFlag adds a `component` flag to the given cobra command
+// Also adds a completion handler to the flag
 func AddComponentFlag(cmd *cobra.Command) {
 	cmd.Flags().String(genericclioptions.ComponentFlagName, "", "Component, defaults to active component.")
 	RegisterCommandFlagHandler(cmd, "component", ComponentNameCompletionHandler)
 }
 
+// AddApplicationFlag adds a `app` flag to the given cobra command
+// Also adds a completion handler to the flag
 func AddApplicationFlag(cmd *cobra.Command) {
 	cmd.Flags().String(genericclioptions.ApplicationFlagName, "", "Application, defaults to active application")
 	RegisterCommandFlagHandler(cmd, "app", AppCompletionHandler)
--- a/pkg/odo/util/completion/completion.go
+++ b/pkg/odo/util/completion/completion.go
@@ -159,24 +159,3 @@ func getUserTypedCommands(args complete.Args, command *cobra.Command) []string {
 
 	return commands
 }
-
-// AddProjectFlag adds a `project` flag to the given cobra command
-// Also adds a completion handler to the flag
-func AddProjectFlag(cmd *cobra.Command) {
-	cmd.Flags().String(genericclioptions.ProjectFlagName, "", "Project, defaults to active project")
-	RegisterCommandFlagHandler(cmd, "project", ProjectNameCompletionHandler)
-}
-
-// AddComponentFlag adds a `component` flag to the given cobra command
-// Also adds a completion handler to the flag
-func AddComponentFlag(cmd *cobra.Command) {
-	cmd.Flags().String(genericclioptions.ComponentFlagName, "", "Component, defaults to active component.")
-	RegisterCommandFlagHandler(cmd, "component", ComponentNameCompletionHandler)
-}
-
-// AddApplicationFlag adds a `app` flag to the given cobra command
-// Also adds a completion handler to the flag
-func AddApplicationFlag(cmd *cobra.Command) {
-	cmd.Flags().String(genericclioptions.ApplicationFlagName, "", "Application, defaults to active application")
-	RegisterCommandFlagHandler(cmd, "app", AppCompletionHandler)
-}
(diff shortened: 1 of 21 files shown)


mik-dass:
@metacosm should I wait till #985 is merged?

metacosm:
@mik-dass up to you, if you get this merged first, I’ll adapt my PR to your code! ;)

mik-dass:
@metacosm Actually I need #985 for showing better suggestions based on the flags (it's not working properly with the current code) :D So I wait

mik-dass:
@metacosm Ready for review

pkg/odo/util/completion/completionhandlers.go
```diff
@@ -79,3 +80,22 @@ var ProjectNameCompletionHandler = func(cmd *cobra.Command, args parsedArgs, con
...
+	completions = make([]string, 0)
+	components, err := component.List(context.Client, context.Application)
+	if err != nil {
+		return completions
+	}
+
+	for _, component := range components {
+		// we found the component name in the list which means
+		// that the component name has been already selected by the user so no need to suggest more
+		if val, ok := args.commands[component.Name]; ok && val {
```
metacosm: You only need to do: `if args.commands[component.Name] {` here.

cmd/cmdutils.go
```diff
@@ -157,8 +157,10 @@ func addProjectFlag(cmd *cobra.Command) {
 
 func addComponentFlag(cmd *cobra.Command) {
 	cmd.Flags().String(util.ComponentFlagName, "", "Component, defaults to active component.")
+	completion.RegisterCommandFlagHandler(cmd, "component", completion.ComponentNameCompletionHandler)
 }
 
 func addApplicationFlag(cmd *cobra.Command) {
 	cmd.Flags().String(util.ApplicationFlagName, "", "Application, defaults to active application")
+	completion.RegisterCommandFlagHandler(cmd, "app", completion.AppCompletionHandler)
```
metacosm: That's not strictly related to this PR but is appropriate anyway :)

cmd/cmdutils.go
```diff
@@ -157,8 +157,10 @@ func addProjectFlag(cmd *cobra.Command) {
 
 func addComponentFlag(cmd *cobra.Command) {
 	cmd.Flags().String(util.ComponentFlagName, "", "Component, defaults to active component.")
+	completion.RegisterCommandFlagHandler(cmd, "component", completion.ComponentNameCompletionHandler)
```
metacosm: 👍

mik-dass:
@metacosm Fixed

geoand:
@mik-dass Could you please rebase this so we can re-test? Thank you

mik-dass:
@geoand ok sure

geoand:
Thanks @mik-dass !

geoand:
@mik-dass Cool :)

metacosm:
> @geoand Actually I need some inputs from @metacosm regarding the new changes before I rebase. I don't want to break anything accidentally :)

What do you need, @mik-dass?

mik-dass:
> What do you need, @mik-dass?

I want to remove the addProjectFlag() functions calls and call the genericclioptions stuff directly. But am I missing something?

pkg/odo/util/completion/completionhandlers.go
```diff
@@ -75,7 +76,7 @@ var ProjectNameCompletionHandler = func(cmd *cobra.Command, args parsedArgs, con
 	for _, project := range projects {
 		// we found the project name in the list which means
 		// that the project name has been already selected by the user so no need to suggest more
-		if val, ok := args.commands[project.Name]; ok && val {
+		if args.commands[project.Name] {
```
geoand: Is this guaranteed to work?

If so, can you please apply it to the rest of the handlers as well?

pkg/odo/util/completion/completion.go
```diff
@@ -159,3 +159,18 @@ func getUserTypedCommands(args complete.Args, command *cobra.Command) []string {
 
 	return commands
 }
+
+func AddProjectFlag(cmd *cobra.Command) {
```
geoand: Can you add some comments about this and the following functions as well?
Them being in the completion package makes it a little tricky to immediately understand that they actually add a cobra flag without looking at the source code :)

pkg/odo/util/completion/completionhandlers.go
metacosm: This works if the map is of type `map[string]bool`, which it is :)

geoand: Yeah, just found a relevant SO answer :)

However this change should also be applied to the other functions in this file as well

pkg/odo/util/completion/completion.go
geoand: I agree. However the solution seems better than the alternative :)

metacosm:
I would like to get more opinions on this new re-organization before merging.

metacosm:
So @kadel, @cdrage: what should we do with respect to the design issue?

kadel:
> @metacosm @kadel How about I put those functions in some cmdUtil file?

not sure about that, would that help?


How about adding them to respective cli packages?
`AddProjectFlag` to `github.com/redhat-/developer/odo/pkg/odo/cli/project`
`AddComponentFlag` to `github.com/redhat-/developer/odo/pkg/odo/cli/component`
`AddApplicationFlag` to `github.com/redhat-/developer/odo/pkg/odo/cli/application`

mik-dass:
> How about adding them to respective cli packages?
AddProjectFlag to github.com/redhat-/developer/odo/pkg/odo/cli/project
AddComponentFlag to github.com/redhat-/developer/odo/pkg/odo/cli/component
AddApplicationFlag to github.com/redhat-/developer/odo/pkg/odo/cli/application

Good idea! Done Please have a look :)

metacosm:
@mik-dass @kadel I don't think this new version is a good idea either because this adds even more links between packages that shouldn't be related but if you're fine with it, I can live with it as well :)

mik-dass:
> @mik-dass @kadel I don't think this new version is a good idea either because this adds even more links between packages that shouldn't be related but if you're fine with it, I can live with it as well :)

I agree it's still a bad design but at least it is better than putting those functions in the completion.go file. So let's leave it for now, we can refactor it later on, WDYT? :)

geoand:
Let's just rebase the PR first :P

PR #947, opened 2018-11-07
Remove test execution from appveyor

geoand:

We remove the test execution phase from appveyor
This is done because all we want appveyor to do is produce artifacts
Tests are covered already by Travis.
Furthermore tests can sometimes timeout thus failing appveyor execution, which we can't then seem to restart from the Appveyor UI

No

How to test changes?
Not odo code related, so nothing to test :)


geoand:
@anmolbabu Sure thing :)

I will go ahead and close this then

PR #949, opened 2018-11-07
Factor out creation of contextual information (project, application, component) to make it easier to manage and facilitate refactoring of commands

metacosm:
== What is the purpose of this change? What does it change?
Introduces a new `Context` struct to hold project, application and component information based on specified flags. This should simplify the commands a bit and ultimately, this structure should be passed as an argument to lots of functions which currently take a client, project, application or component name as arguments. This work also targets #871 

== Was the change discussed in an issue?
No

== How to test changes?
External behavior shouldn't change.

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
+	"github.com/redhat-developer/odo/pkg/odo/util"
 	"os"
 	"strings"
 
-	"github.com/redhat-developer/odo/pkg/odo/util"
 	"github.com/redhat-developer/odo/pkg/odo/util/completion"
 
 	"text/tabwriter"
@@ -56,9 +57,11 @@ If no app name is passed, a default app name will be auto-generated.
 	`,
 	Args: cobra.RangeArgs(0, 1),
 	Run: func(cmd *cobra.Command, args []string) {
+		context := genericclioptions.NewContext(cmd)
+		client := context.Client
+		projectName := context.Project
+
 		var appName string
-		client := util.GetOcClient()
-		projectName := util.GetAndSetNamespace(client)
 		if len(args) == 1 {
 			// The only arg passed is the app name
 			appName = args[0]
@@ -79,6 +82,11 @@ If no app name is passed, a default app name will be auto-generated.
 		err = application.Create(client, appName)
 		util.CheckError(err, "")
 		err = application.SetCurrent(client, appName)
+
+		// todo: updating the app name should be done via SetCurrent and passing the Context
+		// not strictly needed here but Context should stay in sync
+		context.Application = appName
+
 		util.CheckError(err, "")
 		fmt.Printf("Switched to application: %v in project: %v\n", appName, projectName)
 	},
@@ -93,10 +101,9 @@ var applicationGetCmd = &cobra.Command{
 	`,
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
-		client := util.GetOcClient()
-		projectName := util.GetAndSetNamespace(client)
-		app, err := application.GetCurrent(projectName)
-		util.CheckError(err, "")
+		context := genericclioptions.NewContext(cmd)
+		projectName := context.Project
+		app := context.Application
 		if applicationShortFlag {
 			fmt.Print(app)
 			return
@@ -118,15 +125,11 @@ var applicationDeleteCmd = &cobra.Command{
 	`,
 	Args: cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		client := util.GetOcClient()
-		projectName := util.GetAndSetNamespace(client)
-		var appName string
-		// If name of the app to be deleted is not passed, consider the current app for deletion
-		if len(args) == 0 {
-			var err error
-			appName, err = application.GetCurrent(projectName)
-			util.CheckError(err, "")
-		} else {
+		context := genericclioptions.NewContext(cmd)
+		client := context.Client
+		projectName := context.Project
+		appName := context.Application
+		if len(args) == 1 {
 			// If app name passed, consider it for deletion
 			appName = args[0]
 		}
@@ -172,9 +175,10 @@ var applicationListCmd = &cobra.Command{
 	`,
 	Args: cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
-		client := util.GetOcClient()
+		context := genericclioptions.NewContext(cmd)
+		client := context.Client
+		projectName := context.Project
 
-		projectName := util.GetAndSetNamespace(client)
 		apps, err := application.ListInProject(client)
 		util.CheckError(err, "unable to get list of applications")
 		if len(apps) > 0 {
@@ -211,11 +215,12 @@ var applicationSetCmd = &cobra.Command{
 		}
 		return nil
 	}, Run: func(cmd *cobra.Command, args []string) {
-		client := util.GetOcClient()
-		appName := args[0]
+		context := genericclioptions.NewContext(cmd)
+		client := context.Client
+		projectName := context.Project
 
-		projectName := util.GetAndSetNamespace(client)
 		// error if application does not exist
+		appName := args[0]
 		exists, err := application.Exists(client, appName)
 		util.CheckError(err, "unable to check if application exists")
 		if !exists {
@@ -226,6 +231,10 @@ var applicationSetCmd = &cobra.Command{
 		err = application.SetCurrent(client, appName)
 		util.CheckError(err, "")
 		fmt.Printf("Switched to application: %v in project: %v\n", args[0], projectName)
+
+		// todo: updating the app name should be done via SetCurrent and passing the Context
+		// not strictly needed here but Context should stay in sync
+		context.Application = appName
 	},
 }
 
@@ -238,13 +247,12 @@ var applicationDescribeCmd = &cobra.Command{
   odo app describe webapp
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
-		client := util.GetOcClient()
-		var appName string
-		projectName := util.GetAndSetNamespace(client)
+		context := genericclioptions.NewContext(cmd)
+		client := context.Client
+		projectName := context.Project
+
+		appName := context.Application
 		if len(args) == 0 {
-			var err error
-			appName, err = application.GetCurrent(projectName)
-			util.CheckError(err, "")
 			if appName == "" {
 				fmt.Printf("There's no active application in project: %v\n", projectName)
 				os.Exit(1)
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -83,7 +83,7 @@ If no app name is passed, a default app name will be auto-generated.
 		util.CheckError(err, "")
 		err = application.SetCurrent(client, appName)
 
-		// todo: updating the app name should be done via SetCurrent and passing the Context
+		// TODO: updating the app name should be done via SetCurrent and passing the Context
 		// not strictly needed here but Context should stay in sync
 		context.Application = appName
 
@@ -232,7 +232,7 @@ var applicationSetCmd = &cobra.Command{
 		util.CheckError(err, "")
 		fmt.Printf("Switched to application: %v in project: %v\n", args[0], projectName)
 
-		// todo: updating the app name should be done via SetCurrent and passing the Context
+		// TODO: updating the app name should be done via SetCurrent and passing the Context
 		// not strictly needed here but Context should stay in sync
 		context.Application = appName
 	},
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -1,0 +1,118 @@
+package genericclioptions
+
+import (
+	"fmt"
+	"github.com/redhat-developer/odo/pkg/application"
+	"github.com/redhat-developer/odo/pkg/component"
+	"github.com/redhat-developer/odo/pkg/occlient"
+	"github.com/redhat-developer/odo/pkg/odo/util"
+	"github.com/redhat-developer/odo/pkg/project"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+// NewContext creates a new Context struct populated with the current state based on flags specified for the provided command
+func NewContext(command *cobra.Command) *Context {
+	flags := command.Flags()
+	skipConnectionCheck, err := flags.GetBool(util.SkipConnectionCheckFlagName)
+	util.CheckError(err, "")
+
+	client, err := occlient.New(skipConnectionCheck)
+	util.CheckError(err, "")
+
+	// project
+	var ns string
+	projectFlag, err := flags.GetString(util.ProjectFlagName)
+	util.CheckError(err, "")
+	if len(projectFlag) > 0 {
+		_, err := project.Exists(client, projectFlag)
+		util.CheckError(err, "")
+		ns = projectFlag
+	} else {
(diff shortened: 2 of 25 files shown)


pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,118 @@
...
+		return o.cmp
+	}
+
+	switch len(optionalComponent) {
+	case 0:
+		if len(o.cmp) == 0 {
+			fmt.Println("No component is set")
+			os.Exit(-1)
+		}
+		break
```
golangcibot: redundant break statement

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,118 @@
...
+	case 1:
+		cmp := optionalComponent[0]
+		// only check the component if we passed a non-empty string, otherwise return the current component set in NewContext
+		if len(cmp) > 0 {
+			o.existsOrExit(cmp)
+			o.cmp = cmp // update context
+		}
+		break
```
golangcibot: redundant break statement

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+	return client(command)
+}
+
+// ClientWithConnectionCheck returns an oc client configured for this command's options but forcing the connection check status
+// to the value of the provided bool, skipping it if true, checking the connection otherwise
+func ClientWithConnectionCheck(command *cobra.Command, skipConnectionCheck bool) *occlient.Client {
+	return client(command, skipConnectionCheck)
+}
+
+//
```
cdrage: Missing comments

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+		util.CheckError(err, "")
+	}
+
+	client, err := occlient.New(skipConnectionCheck)
+	util.CheckError(err, "")
+
+	return client
+}
+
+func newContext(command *cobra.Command, createAppIfNeeded bool) *Context {
```
cdrage: missing comments thoughout the function

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// ClientWithConnectionCheck returns an oc client configured for this command's options but forcing the connection check status
+// to the value of the provided bool, skipping it if true, checking the connection otherwise
+func ClientWithConnectionCheck(command *cobra.Command, skipConnectionCheck bool) *occlient.Client {
+	return client(command, skipConnectionCheck)
+}
+
+//
+func client(command *cobra.Command, shouldSkipConnectionCheck ...bool) *occlient.Client {
```
cdrage: why are we passing in multiple bools? shouldn't it just be `bool` instead of `...bool`?

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+		context.existsOrExit(cmpFlag)
+		cmp = cmpFlag
+	}
+
+	context.cmp = cmp
+
+	return context
+}
+
+func FlagValueIfSet(cmd *cobra.Command, flagName string) string {
```
cdrage: missing comments

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+	return flag
+}
+
+// Context holds contextual information useful to commands such as correctly configured client, target project and application
+// (based on specified flag values) and provides for a way to retrieve a given component given this context
+type Context struct {
+	internalCxt
+}
+
+type internalCxt struct {
```
cdrage: comments

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	default:
+		fmt.Printf("Component function only accepts one optional argument, was given: %v", optionalComponent)
+		os.Exit(-1)
+	}
+
+	return o.cmp
+}
+
+func (o *Context) existsOrExit(cmp string) {
```
cdrage: comments

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,173 @@
...
+	exists, err := component.Exists(o.Client, cmp, o.Application)
+	util.CheckError(err, "")
+	if !exists {
+		fmt.Printf("Component %v does not exist in application %s\n", cmp, o.Application)
+		os.Exit(-1)
+	}
+}
+
+func ignoreButLog(err error) {
```
cdrage: comments

pkg/odo/genericclioptions/context.go
metacosm: The idea here is to be able to determine whether a boolean is passed or not because we cannot rely on zero value for bool (which would be false). Since the method is generic, we need to determine whether or not we want to override the flag value so passing a single bool doesn't work. Using varargs allow a ternary state, making the bool optional so either it's present and its value should always override the flag or it is not, and we take the flag into account. It's not ideal but I preferred that solution as opposed to passing a pointer and check for nil (which didn't really make sense) or passing 2 bools where one would be ignored depending on the value of the first one.

metacosm:
@cdrage updated with comments.

metacosm:
Tests pass!

pkg/odo/genericclioptions/context.go
girishramnani: can we consider a Options like builder pattern?

```
type Option func(client *occlient.Client)

func SkipConnectionCheck(check bool) Option {
	return func(client *occlient.Client) {
		client.skipConnectionCheck = check
	}
}
```
Then the client function will work like this

```
func client(command *cobra.Command, options ...Option) *occlient.Client {
}

```
This can be useful if we want to add more options in the future.

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+		// safeguard: fail if more than one optional string is passed because it would be a programming error
+		fmt.Printf("ComponentAllowingEmpty function only accepts one optional argument, was given: %v", optionalComponent)
+		os.Exit(-1)
+	}
+
+	return o.cmp
+}
+
+// existsOrExit checks if the specified component exists with the given context and exits the app if not.
+func (o *Context) existsOrExit(cmp string) {
```
anmolbabu: checkComponentExistsOrFail ?

metacosm: I thought that `existsOrExit` was pretty explicit but I can change it if needed.

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+// NewContext creates a new Context struct populated with the current state based on flags specified for the provided command
+func NewContext(command *cobra.Command) *Context {
+	return newContext(command, false)
+}
+
+// NewContextCreatingAppIfNeeded creates a new Context struct populated with the current state based on flags specified for the
+// provided command, creating the application if none already exists
+func NewContextCreatingAppIfNeeded(command *cobra.Command) *Context {
```
anmolbabu: NewContextWithAppCreate ?

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+// (based on specified flag values) and provides for a way to retrieve a given component given this context
+type Context struct {
+	internalCxt
+}
+
+// internalCxt holds the actual context values and is not exported so that it cannot be instantiated outside of this package.
+// This ensures that Context objects are always created properly via NewContext factory functions.
+type internalCxt struct {
+	Client      *occlient.Client
+	Project     string
```
anmolbabu: The client already has `Namespace` why project again?

Also, why is fetching project, app and component all 3 always(for every command) required?
Would fetching them on demand for any command be like it used to be before, better?

anmolbabu: I Agree its not much of an overhead though :thinking:

pkg/odo/genericclioptions/context.go
metacosm: I think that `NewContextCreatingAppIfNeeded` is more explicit.

pkg/odo/genericclioptions/context.go
metacosm: Regarding `Namespace` vs. `Project`, I used the `odo` terminology as we're not using `Namespace` in the rest of the code. This also matches the names of the flags that influence that state.

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Component retrieves the optionally specified component or the current one if it is set. If no component is set, exit with
+// an error
+func (o *Context) Component(optionalComponent ...string) string {
+	return o.ComponentAllowingEmpty(false, optionalComponent...)
+}
+
+// ComponentAllowingEmpty retrieves the optionally specified component or the current one if it is set, allowing empty
+// components (instead of exiting with an error) if so specified
+func (o *Context) ComponentAllowingEmpty(allowEmpty bool, optionalComponent ...string) string {
```
anmolbabu: Can we prefix Get to this , to indicate its a getter?

metacosm: Sure, the idea here, though, was to more or less match the `Client`, `Application`, `Project` mold :)

pkg/odo/genericclioptions/context.go
anmolbabu: Hmm so since the Context would now capture the project, can we remove Namespace from within client as it can be ambiguous. We can also take that up later if it makes this a very huge change but presence of 2 similar fields leads to ambiguity right?

pkg/odo/genericclioptions/context.go
metacosm: > can we consider a Options like builder pattern?
> 
> ```
> type Option func(client *occlient.Client)
> 
> func SkipConnectionCheck(check bool) Option {
> 	return func(client *occlient.Client) {
> 		client.skipConnectionCheck = check
> 	}
> }
> ```
> 
> Then the client function will work like this
> 
> ```
> func client(command *cobra.Command, options ...Option) *occlient.Client {
> }
> ```
> 
> This can be useful if we want to add more options in the future.

That's indeed an interesting idea that I hadn't considered. However, I'd rather not speculate on the future at this point and keep this PR as it is. We can revisit this idea at a later point if needed.

pkg/odo/genericclioptions/context.go
metacosm: I agree that we should probably remove it but I think that should be the scope of a different PR. Actually, I think that most functions should take a `Context` as parameter as well where it makes sense.

pkg/odo/genericclioptions/context.go
metacosm: Changed!

pkg/odo/genericclioptions/context.go
kadel: > I agree that we should probably remove it but I think that should be the scope of a different PR. Actually, I think that most functions should take a `Context` as parameter as well where it makes sense.

👍  Definitely, we have to refactor most of the function to take this. But it should be addressed in another PR. Meanwhile, we have to be careful with setting Project/Namespace. @anmolbabu has a good point this duplicity could lead to bugs.
@metacosm Could you please create an issue to make sure all this that we need to do won't get lost?

pkg/odo/genericclioptions/context.go
metacosm: @kadel Done: see #968

pkg/odo/genericclioptions/context.go
metacosm: Done, see #968

cmd/application.go
```diff
@@ -226,6 +231,10 @@ var applicationSetCmd = &cobra.Command{
 		err = application.SetCurrent(client, appName)
 		util.CheckError(err, "")
 		fmt.Printf("Switched to application: %v in project: %v\n", args[0], projectName)
+
+		// todo: updating the app name should be done via SetCurrent and passing the Context
```
cdrage: todo should be TODO.. sorry for small nitpick, but Golang CI picks up on TODO

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+	case 0:
+		var err error
+		skipConnectionCheck, err = command.Flags().GetBool(util.SkipConnectionCheckFlagName)
+		util.CheckError(err, "")
+	case 1:
+		skipConnectionCheck = shouldSkipConnectionCheck[0]
+	default:
+		// safeguard: fail if more than one optional bool is passed because it would be a programming error
+		fmt.Printf("client function only accepts one optional argument, was given: %v", shouldSkipConnectionCheck)
+		os.Exit(-1)
```
cdrage: Shouldn't this be 1 not -1?

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+// ComponentAllowingEmpty retrieves the optionally specified component or the current one if it is set, allowing empty
+// components (instead of exiting with an error) if so specified
+func (o *Context) ComponentAllowingEmpty(allowEmpty bool, optionalComponent ...string) string {
+	switch len(optionalComponent) {
+	case 0:
+		// if we're not specifying a component to resolve, get the current one (resolved in NewContext as cmp)
+		// so nothing to do here unless the calling context doesn't allow no component to be set in which case we exit with error
+		if !allowEmpty && len(o.cmp) == 0 {
+			fmt.Println("No component is set")
+			os.Exit(-1)
```
cdrage: 1 not -1

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+		cmp := optionalComponent[0]
+		// only check the component if we passed a non-empty string, otherwise return the current component set in NewContext
+		if len(cmp) > 0 {
+			o.checkComponentExistsOrFail(cmp)
+			o.cmp = cmp // update context
+		}
+	default:
+		// safeguard: fail if more than one optional string is passed because it would be a programming error
+		fmt.Printf("ComponentAllowingEmpty function only accepts one optional argument, was given: %v", optionalComponent)
+		os.Exit(-1)
```
cdrage: 1 not -1

pkg/odo/genericclioptions/context.go
```diff
@@ -0,0 +1,204 @@
...
+	return o.cmp
+}
+
+// existsOrExit checks if the specified component exists with the given context and exits the app if not.
+func (o *Context) checkComponentExistsOrFail(cmp string) {
+	exists, err := component.Exists(o.Client, cmp, o.Application)
+	util.CheckError(err, "")
+	if !exists {
+		fmt.Printf("Component %v does not exist in application %s\n", cmp, o.Application)
+		os.Exit(-1)
```
cdrage: 1 not -1

cdrage:
Submitted it as comment only, but I've added some suggested changes @metacosm

kadel:
> I believe the exit codes should be `1` not `-1`

yep, we are using `1` as a return value everywhere else.

metacosm:
> > I believe the exit codes should be `1` not `-1`
> 
> yep, we are using `1` as a return value everywhere else.

Should we look into using `errno` values for more semantic error codes? Not for this PR obviously, but more generally speaking.

metacosm:
> Submitted it as comment only, but I've added some suggested changes @metacosm

@cdrage fixed

cdrage:
Changes LGTM.

Let's wait until tests return back green and then we can merge :100:

metacosm:
Awesome, merging based on 3 acks and all checks passing!

kadel:
> Should we look into using `errno` values for more semantic error codes? Not for this PR obviously, but more generally speaking.

are there constants for exit codes in golang stdlib? I couldn't find it.

Issue #950, opened 2018-11-08
odo push all the target directory content and not the binary file


cmoulliard:
## What versions of software are you using?

Operating System : macos

Output of `odo version` : odo v0.0.14 (618c6d3)

## How did you run odo exactly?

When we create a component to push a binary file using this command `odo create redhat-openjdk18-openshift:1.4 my-spring-boot --binary ./target/my-spring-boot-0.0.1-SNAPSHOT.jar` and next we do a `odo push`, then the content of the target directory is pushed

## Expected behavior

Only the binary file should be uploaded within the `/deployments` directory of the dev's pod and not all the files of the target directory

![screenshot 2018-11-08 10 29 56](https://user-images.githubusercontent.com/463790/48189877-7a5b8e00-e341-11e8-8319-73d56173fede.png)

![screenshot 2018-11-08 10 28 22](https://user-images.githubusercontent.com/463790/48189904-8a736d80-e341-11e8-9915-6f0665517aa2.png)

## Investigation

I suspect that the issue is coming from the fact that the tar file which is created and uploaded includes all the files of the directory `/opt/app-root/src` and not the binary file

![screenshot 2018-11-08 10 35 41](https://user-images.githubusercontent.com/463790/48190166-459c0680-e342-11e8-9df4-7a7f943ae27c.png)

geoand:
@cmoulliard I can confirm this issue exists even from the latest `master`. Although the behavior is not correct, it does not cause an problem. The Java s2i image is still perfectly capable of running the jar in the `/deployments` directory.
We should definitely look into fixing this based on your investigation, but I think we can live with it for the moment :)

jorgemoralespou:
@geond I disagree. We need to change this so that push only pushes the binary that is provided. It affects speed and can potentially affect behavior if there are multiple jar files in the target folder.

geoand:
@jorgemoralespou OK :)

metacosm:
@geoand pick it up, man! :)

geoand:
I am looking into this and although I have identified the problem I don't expect to have a fix anytime soon, basically because this is the first time that I am looking into go and kubernetes related IO code :)

geoand:
I created a [fix](https://github.com/redhat-developer/odo/pull/954) that I think works (tested it a few different scenarios and it works).
Please give it a try :)

geoand:
Fixed in: 7379c138

PR #954, opened 2018-11-08
Optimize the case of single file transfer to the supervisord pod

geoand:

This fixes an issue reported #950 which resulted in the whole directory
that contained a binary file being copied to the supevisord pod

#950 

How to test changes?

```
odo push
```

Then looking in `/opt/app-root/src` of the supervisord pod there should only be a single file, the jar file.

For regression testing purposes, also test that a nodejs app where we push an entire source directory still works

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -2483,12 +2483,21 @@ func (c *Client) GetOnePodFromSelector(selector string) (*corev1.Pod, error) {
 // During copying local source components, localPath represent base directory path whereas copyFiles is empty
 // During `odo watch`, localPath represent base directory path whereas copyFiles contains list of changed Files
 func (c *Client) CopyFile(localPath string, targetPodName string, targetPath string, copyFiles []string) error {
+	isSingleFileTransfer := isSingleFileTransfer(copyFiles)
+
 	dest := path.Join(targetPath, filepath.Base(localPath))
 	reader, writer := io.Pipe()
 	// inspired from https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cp.go#L235
 	go func() {
 		defer writer.Close()
-		err := makeTar(localPath, dest, writer, copyFiles)
+
+		var err error
+		if isSingleFileTransfer {
+			onlyFile := copyFiles[0]
+			err = makeTar(onlyFile, targetPath+"/"+path.Base(onlyFile), writer, []string{})
+		} else {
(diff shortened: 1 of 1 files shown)


pkg/occlient/occlient.go
```diff
@@ -2204,7 +2219,10 @@ func (c *Client) CopyFile(localPath string, targetPodName string, targetPath str
 	}()
 
 	// cmdArr will run inside container
-	cmdArr := []string{"tar", "xf", "-", "-C", targetPath, "--strip", "1"}
+	cmdArr := []string{"tar", "xf", "-", "-C", targetPath}
```
jorgemoralespou: Why do we execute an external command instead of using the tar library that go provides? https://golang.org/pkg/archive/tar/

jorgemoralespou:
@geoand Would this support pushing binaries as directories? https://github.com/redhat-developer/odo/issues/854

pkg/occlient/occlient.go
geoand: This command get's executed inside the supervisord container. It's basically needed because we send a stream of bytes to the pod that are "tarred", so inside the "pod" we need to untar them.

Also, I didn't add this code, I just altered the code that was already there to work for our use case :P

geoand:
@jorgemoralespou I wasn't aware of that one.

I need to check, but my hunch is that it won't :)

pkg/occlient/occlient.go
jorgemoralespou: So, we're assuming that container has tar installed? I'm not sure that that's the case in every s2i container (and eventually non s2i).

Maybe we should open an issue to investigate if that can happen.

geoand: @jorgemoralespou That is indeed the case. The code seems to be initially inspired by `kubectl`.

See: https://github.com/kubernetes/kubernetes/blob/040cc281f4849dee05d35e4f2fa2ee36729dea22/pkg/kubectl/cmd/cp/cp.go#L45

geoand: @kadel Can you shine some light here please?

geoand:
@anmolbabu Regarding your first comment, that doesn't work because of what is being passed to `makeTar`.

As for the second comment, are you sure? I don't see how that would work

kadel:
Failing test is one that tests macos installation. all e2e tests are ok.
I think that this can be merged.

geoand:
Sounds good to me!

PR #958, opened 2018-11-08
Correct link documentation

cdrage:
Corrects the link documentation with the correct examples (library /
component source FIRST and then the name..)

### DIFF
--- a/cmd/link.go
+++ b/cmd/link.go
@@ -28,11 +28,11 @@ The source component can then consume the entries of the secret as environment v
 
 For example:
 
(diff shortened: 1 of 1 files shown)


cmd/link.go
```diff
@@ -28,11 +28,11 @@ The source component can then consume the entries of the secret as environment v
 
 For example:
 
-We have created a backend application called 'backend' with port 8080 exposed:
-odo create backend nodejs --port 8080
+We have created a frontend application called 'frontend':
+odo create nodejs frontend nodejs
```
metacosm: This doesn't look right.

PR #959, opened 2018-11-09
Added the project auto completion code

### LINKED ISSUE #940: Suggestions for project names
parrent: #902

Suggesting project names 
 -`odo project delete`
 - + every command that has `--project` flag

mik-dass:
fixes #940

### DIFF
--- a/pkg/odo/util/completion/completion.go
+++ b/pkg/odo/util/completion/completion.go
@@ -5,7 +5,6 @@ import (
 	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
-	"strings"
 )
 
 type completionHandler struct {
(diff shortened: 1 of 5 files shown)


metacosm:
@mik-dass can you please add comments in your code so that it's easier to follow?

mik-dass:
@metacosm I already have added comments inside the functions. Should I explain more in the comments?

metacosm:
@mik-dass I was trying to figure out more the bigger picture of what you were trying to do :)
As a side-note, how do you debug the process?

mik-dass:
@metacosm Ok I will add some more comments :)

> As a side-note, how do you debug the process?

Don't ask :D It's a pain. I don't know how to configure the GoLand debugger for this, so I currently use Print statements. Do you know any better easier way?

pkg/odo/util/completion/completionhandlers.go
```diff
@@ -39,3 +41,57 @@ var ServiceClassCompletionHandler = func(args complete.Args, client *occlient.Cl
...
+			found = true
+			continue
+		}
+		if found {
+			commands = append(commands, arg)
+		}
+	}
+
+	// extract the flags and commands
+	strippedCommands, strippedFlags := getCommandsAndFlags(commands, command)
```
metacosm: Why not make `getCommandsAndFlags` return a map for commands directly?

pkg/odo/util/completion/completionhandlers.go
```diff
@@ -39,3 +41,57 @@ var ServiceClassCompletionHandler = func(args complete.Args, client *occlient.Cl
...
+			completions = append(completions, project.Name)
+			continue
+		}
+		// if the flag suggestion is done
+		if val, ok := strippedFlags["project"]; ok && val != "" {
+			return completions
+		}
+		// we found the project name in the list which means
+		// that the project name has been already selected by the user so no need to suggest more
+		if val, ok := strippedCommandsMap[project.Name]; ok && val {
```
metacosm: You can just do `if strippedCommandsMap[project.Name] {` here I think.

pkg/odo/util/completion/completion.go
```diff
@@ -76,3 +80,50 @@ func GetCommandFlagHandler(command *cobra.Command, flag string) (predictor compl
 	predictor, ok = completionHandlers[getCommandFlagCompletionHandlerKey(command, flag)]
 	return
 }
+
+func getCommandsAndFlags(args []string, c *cobra.Command) ([]string, map[string]string) {
```
metacosm: I believe this method can be implemented in a much simpler way:
```go
func getCommandsAndFlags(args []string, c *cobra.Command) ([]string, map[string]string) {
	if len(args) == 0 {
		return args, nil
	}
	c.ParseFlags(args)
	flags := c.Flags()

	cmds := flags.Args()
	setFlags := make(map[string]string)
	flags.Visit(func(i *flag.Flag) {
		if i.Value.Type() != "bool" {
			setFlags[i.Name]=i.Value.String()
		}
	})

	return cmds, setFlags
}
```

metacosm: As a result, `isBooleanFlag` can also be removed.

pkg/odo/util/completion/completionhandlers.go
```diff
@@ -39,3 +41,57 @@ var ServiceClassCompletionHandler = func(args complete.Args, client *occlient.Cl
...
+
+// ProjectNameCompletionHandler provides project name completion
+var ProjectNameCompletionHandler = func(args complete.Args, client *occlient.Client, command *cobra.Command) (completions []string) {
+	completions = make([]string, 0)
+	projects, err := project.List(client)
+	if err != nil {
+		return completions
+	}
+
+	var commands []string
```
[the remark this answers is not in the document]
mik-dass: If I return here, it would suggest the name of the projects again and again if the user keeps pressing tab again and again.
 
So I want to stop suggesting once the suggestion is complete. So I extract the flags and the commands. Check if the suggestion is already complete. If completed then return nothing else. 

Also for flags, I added a few cases.

pkg/odo/util/completion/completionhandlers.go
mik-dass: I thought of that but thought that list would be useful later. I will change it to map

pkg/odo/util/completion/completion.go
mik-dass: Wow! I didn't notice there was a flags.Args() function for the job. Thanks :+1:  Will try it out

metacosm:
> > As a side-note, how do you debug the process?
> 
> Don't ask :D It's a pain. I don't know how to configure the GoLand debugger for this, so I currently use Print statements. Do you know any better easier way?

Yes, it's a pain. You can set `export COMP_DEBUG=1` on your shell to see some output when you tab but it's not great… I'm not sure there's a way to attach to the go process while it's completing since I assume it's only running the time it needs to output the completions and it's called by the shell… so… :/

mik-dass:
>Yes, it's a pain. You can set export COMP_DEBUG=1 on your shell to see some output when you tab but it's not great… I'm not sure there's a way to attach to the go process while it's completing since I assume it's only running the time it needs to output the completions and it's called by the shell… so… :/

Thanks :+1: I will try the COMP_DEBUG=1 thing. If that doesn't work, will go back to the Print statements :D

pkg/odo/util/completion/completion.go
```diff
@@ -76,3 +80,50 @@ func GetCommandFlagHandler(command *cobra.Command, flag string) (predictor compl
...
+	if len(args) == 0 {
+		return args, nil
+	}
+	c.ParseFlags(args)
```
golangcibot: Error return value of `c.ParseFlags` is not checked

mik-dass:
@kadel @metacosm Rebased and modified the code according to the suggestions. Please have a look.

geoand:
@mik-dass Could you please rebase so the test will pass and we can merge?

Thanks!

mik-dass:
@geoand Done but I would like to wait for @kadel  approval because #962 wasn't working for him. Thanks :)

geoand:
@mik-dass sounds good :)

metacosm:
@mik-dass this PR is not tied to the one on completing file path, though…

mik-dass:
> @mik-dass this PR is not tied to the one on completing file path, though…


@metacosm Ok let's merge then as it seems to be creating a blocker for others

PR #924, opened 2018-11-05
Autodetect path for assemble and run scripts

anmolbabu:
Autodetect path for assemble and run scripts

Odo hardcodes the S2I assemble script path in
`https://github.com/redhat-developer/odo-supervisord-image/blob/master/assemble-and-restart`
But the S2I docs suggest that the S2I scripts need to looked up in the
following order:
1. A script found at the --scripts-url URL
2. A script found in the application source .s2i/bin directory
3. A script found at the default image URL (io.openshift.s2i.scripts-url label)
and ultimately giveup if not found in any of the above paths.

While, 1 is not relevant to us(builds triggered internally by odo), 2 is already
done by odo, 3 is enhanced by this PR instead of using the currently(without this PR)
hard-coded path

This PR fetches the S2I scripts paths from builder image and extracts the protocol and path
and sets them into environment variables in component container for use of it in running
assemble script.

Note: The S2I scripts suggest that the builder image might point to any of the following
paths:

a. image://path_to_scripts_dir - absolute path inside the image
b. file://path_to_scripts_dir - relative or absolute path on the host machine
c. http(s)://path_to_scripts_dir - URL to a directory

Of these, the assemble-and-restart script currently due to:
`https://github.com/redhat-developer/odo-supervisord-image/pull/3` is capable of handling
`image://path_to_scripts_dir` but not `file://` or `http(s)://` and they are expected to be
added over time. At the moment for any of the `file://` or `http(s)://`, the component
create and/or update is expected to fail as it is expected to be even with latest master.

To test:
Using minishift:
1. Start minishift `minishift start`
2. Build the docker image for `odo-supervisord-image` with PR:
        `https://github.com/redhat-developer/odo-supervisord-image/pull/3`.
   To do this follow the below steps:
   i. To pull the PR:
      * Clone `https://github.com/redhat-developer/odo-supervisord-image`
      * Add my git remote: `git remote add anmolbabu git@github.com:anmolbabu/odo-supervisord-image.git`
      * Pull source from the remote `git pull anmolbabu`
      * Checkout my branch `git checkout remotes/anmolbabu/fixes-odo#869`
      * Create new branch for the new source `git checkout -b anmolbabu-fixes-odo#869`
  ii. Get access to minishift docker environment `eval $(minishift docker-env)`
 iii. Build the `odo-supervisord-image` from the above branch may be with a new version(that's not in quay)
      to be sure we are using the image with the required PRs:
[...]

PR #961, opened 2018-11-09
Add end to end tests that test various runtimes with --local

geoand:

This PR adds integration tests for testing how `odo create` and `odo push` behave when dealing with source code (hence the use of `--local`) of the following runtimes: `openjdk`, `nodejs`, `wildfly`, `python` and `dotnet`.

This was not discussed in an issue per se, but it arose out of the need to test PR #924 which overhauls the supevisord implementation.

How to test changes?

The change simply adds new tests.
You can run those new tests yourself by issuing:
`make test-source-e2e`

**Note**

I have used the term `source` throughout the PR, but I am not really sure it conveys the meaning properly so I would be more than willing to change it to anything better :)


tests/e2e/utils.go
```diff
@@ -0,0 +1,7 @@
+package e2e
+
+import "strings"
+
+func determineRouteURL() string {
```
cdrage: Add comments :)

geoand: Updated with comments :)

PR #699, opened 2018-09-06
Add command odo logout

syamgk:
add `odo logout` command

as mentioned in https://github.com/redhat-developer/odo/issues/555

PR #971, opened 2018-11-13
Use context in logout command (fix failing build)

kadel:
Because #699 wasn't properly rebased and up to date with master branch, we ended up with failing build. This fixes that.


cdrage:
LGTM :+1:

Issue #919, opened 2018-11-02
Timeouts for component unit testes on  on appveyor


kadel:
https://ci.appveyor.com/project/odo-bot/odo

There are too many timeouts for component unit test on appveyor.
I suspect that this might be caused by #820. @anmolbabu  can you check it?
If you login to appveyor using your github, you should have permissions to restart jobs etcc... 

```
[00:00:00] Build started
[00:00:03] git config --system core.longpaths true
[00:00:03] git clone -q https://github.com/redhat-developer/odo.git C:\gopath\src\github.com\redhat-developer\odo
[00:00:37] git fetch -q origin +refs/pull/894/merge:
[00:00:43] git checkout -qf FETCH_HEAD
[00:00:51] Running Install scripts
[00:00:51] rmdir c:\go /s /q
[00:00:51] appveyor DownloadFile https://storage.googleapis.com/golang/go%GOVERSION%.windows-amd64.msi
[00:00:54] 
[00:00:54] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...1%
[00:00:54] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...10%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...20%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...30%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...40%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...50%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...60%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...70%
[00:00:55] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...80%
[00:00:56] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...90%
[00:00:56] Downloading go1.11.windows-amd64.msi (116,633,600 bytes)...100%
[00:00:56] msiexec /i go%GOVERSION%.windows-amd64.msi /q
[00:01:33] go version
[00:01:33] go version go1.11 windows/amd64
[00:01:33] go env
[00:01:33] set GOARCH=amd64
[00:01:33] set GOBIN=
[00:01:33] set GOCACHE=C:\Users\appveyor\AppData\Local\go-build
[00:01:33] set GOEXE=.exe
[...]

PR #975, opened 2018-11-14
Fix watch UTs on Appveyor

anmolbabu:
Fix watch UTs on Appveyor

The problem in watch test was that the same channel was fed and listened
to by 2 functions: the UT and the main WatchAndPush function which does
not guarantee which listener gets the message.
So, the fix now is to make 2 separate channels one each for each of the
directions of the communications:
a. watch UT to WatchAndPush: To signal graceful exit without SIGINT to
   terminate WatchAndPush -- ExtChan
b. WatchAndPush to watch UT: To signal the readiness of watch function
   to listen to any file change as of now simulated by the watch
   UT. -- StartChan

fixes: #919

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ prepare-release: cross
 
 .PHONY: test
 test:
-	go test -v -race $(PKGS)
+	go test -race $(PKGS)
 
 # Run main e2e tests
 .PHONY: test-main-e2e
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ prepare-release: cross
 
 .PHONY: test
 test:
-	go test -race $(PKGS)
+	go test -v -race $(PKGS)
 
 # Run main e2e tests
 .PHONY: test-main-e2e
--- a/pkg/component/watch.go
+++ b/pkg/component/watch.go
@@ -109,22 +109,37 @@ func addRecursiveWatch(watcher *fsnotify.Watcher, path string, ignores []string)
 
 var UserRequestedWatchExit = fmt.Errorf("safely exiting from filesystem watch based on user request")
 
+// WatchParameters is designed to hold the controllables and attributes that the watch function works on
+type WatchParameters struct {
+	// Name of component that is to be watched
+	ComponentName string
+	// Name of application, the component is part of
+	ApplicationName string
+	// The path to the source of component(local or binary)
+	Path string
+	// List/Slice of files/folders in component source, the updates to which need not be pushed to component deployed pod
+	FileIgnores []string
+	// Custom function that can be used to push detected changes to remote pod. For more info about what each of the parameters to this function, please refer, pkg/component/component.go#PushLocal
+	WatchHandler func(*occlient.Client, string, string, string, io.Writer, []string) error
+	// This is a channel added to signal readiness of the watch command to the external channel listeners
+	StartChan chan string
+	// This is a channel added to terminate the watch command gracefully without passing SIGINT. "Stop" message on this channel terminates WatchAndPush function
+	ExtChan chan string
+	// Interval of time before pushing changes to remote(component) pod
+	PushDiffDelay int
+}
+
 // WatchAndPush watches path, if something changes in  that path it calls PushLocal
 // ignores .git/* by default
 // inspired by https://github.com/openshift/origin/blob/e785f76194c57bd0e1674c2f2776333e1e0e4e78/pkg/oc/cli/cmd/rsync/rsync.go#L257
 // Parameters:
 //	client: occlient instance
-//	componentName: Name of component that is to be watched
-//	applicationName: Name of application, the component is part of
-//	path: The path to the source of component(local or binary)
 //	out: io Writer instance
-//	ignores: List/Slice of files/folders in component source, the updates to which need not be pushed to component deployed pod
-//	delayInterval: Interval of time before pushing changes to remote(component) pod
-//	extChan: This is a channel added to terminate the watch command gracefully without passing SIGINT. "Stop" message on this channel terminates WatchAndPush function
-//	pushChangesToPod: Custom function that can be used to push detected changes to remote pod. For more info about what each of the parameters to this function, please refer, pkg/component/component.go#PushLocal
-func WatchAndPush(client *occlient.Client, componentName string, applicationName, path string, out io.Writer, ignores []string, delayInterval int, extChan chan string, pushChangesToPod func(*occlient.Client, string, string, string, io.Writer, []string) error) error {
+// 	parameters: WatchParameters
+func WatchAndPush(client *occlient.Client, out io.Writer, parameters WatchParameters) error {
 	// ToDo reduce number of parameters to this function by extracting them into a struct and passing the struct instance instead of passing each of them separately
-	glog.V(4).Infof("starting WatchAndPush, path: %s, component: %s, ignores %s", path, componentName, ignores)
+	// delayInterval int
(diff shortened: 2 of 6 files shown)


anmolbabu:
@kadel @metacosm @cdrage @geoand @surajnarwade @syamgk @mik-dass Please review

cmd/watch.go
```diff
@@ -2,11 +2,12 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
 	"net/url"
 	"os"
 	"runtime"
 
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
+
```
kadel: why empty line here?

Makefile
```diff
@@ -77,7 +77,7 @@ prepare-release: cross
 
 .PHONY: test
 test:
-	go test -race $(PKGS)
+	go test -v -race $(PKGS)
```
kadel: do we really need tests to be verbose?

pkg/component/watch.go
```diff
@@ -109,22 +109,37 @@ func addRecursiveWatch(watcher *fsnotify.Watcher, path string, ignores []string)
...
+	// Name of application, the component is part of
+	ApplicationName string
+	// The path to the source of component(local or binary)
+	Path string
+	// List/Slice of files/folders in component source, the updates to which need not be pushed to component deployed pod
+	FileIgnores []string
+	// Custom function that can be used to push detected changes to remote pod. For more info about what each of the parameters to this function, please refer, pkg/component/component.go#PushLocal
+	WatchHandler func(*occlient.Client, string, string, string, io.Writer, []string) error
+	// This is a channel added to signal readiness of the watch command to the external channel listeners
+	StartChan chan string
```
kadel: could those channels be just a `boolean`?

pkg/component/watch.go
```diff
@@ -109,22 +109,37 @@ func addRecursiveWatch(watcher *fsnotify.Watcher, path string, ignores []string)
 
 var UserRequestedWatchExit = fmt.Errorf("safely exiting from filesystem watch based on user request")
 
+// WatchParameters is designed to hold the controllables and attributes that the watch function works on
+type WatchParameters struct {
```
kadel: Please move type definitions to the beginning of the file

Makefile
anmolbabu: no I have it enabled so that I didn't miss logs for any failures
I'll turn it off once the PR is out of WIP and ready to be in ;)

cmd/watch.go
anmolbabu: That's my editor(vscode) added line :laughing: 
I'll fix it

pkg/component/watch.go
anmolbabu: My bad :disappointed: 
Fixed now

pkg/component/watch.go
anmolbabu: Ack

pkg/component/watch.go
```diff
@@ -142,13 +157,13 @@ func WatchAndPush(client *occlient.Client, componentName string, applicationName
...
-	defer close(extChan)
+	defer close(parameters.ExtChan)
 
 	go func() {
 		for {
 			select {
-			case extMsg := <-extChan:
-				if extMsg == "Stop" {
+			case extMsg := <-parameters.ExtChan:
+				if extMsg == true {
```
golangcibot: should omit comparison to bool constant, can be simplified to extMsg

pkg/component/watch_test.go
```diff
@@ -279,13 +280,15 @@ func TestWatchAndPush(t *testing.T) {
...
 
 			go func() {
+				t.Logf("Starting file simulations \n%+v\n", tt.fileModifications)
 				// Simulating file modifications for watch to observe
 				for {
 					select {
-					case startMsg := <-ExtChan:
-						if startMsg == "Start" {
+					case startMsg := <-StartChan:
+						if startMsg == true {
```
golangcibot: should omit comparison to bool constant, can be simplified to startMsg

tests/e2e/cmp_test.go
```diff
@@ -155,24 +155,30 @@ var _ = Describe("odoCmpE2e", func() {
...
-					FileParent:       "src/main/java/eu/mjelen/katacoda/odo/",
-					FilePath:         "BackendServlet.java",
-					FileType:         testingutil.RegularFile,
-					ModificationType: testingutil.APPEND,
-				}
-				_, err := testingutil.SimulateFileModifications(filepath.Join(tmpDir, "katacoda-odo-backend-1"), fileModification)
-				fmt.Printf("Triggered file modification %+v\n\n", fileModification)
-				if err != nil {
-					fmt.Printf("Failed performing file operation with error %v", err)
+				select {
```
golangcibot: should use a simple channel send/receive instead of select with a single case

.travis.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-cmp-e2e
+        - travis_wait make test-cmp-e2e
```
surajnarwade: sometimes it takes more than 20 min, so it might not help so, I would suggest use,

```
travis_wait 30
```

Ref: https://docs.travis-ci.com/user/common-build-problems/

anmolbabu: ok

kadel: > sometimes it takes more than 20 min, so it might not help so, I would suggest use,

no, it doesn't matter how long the test takes, travis-ci check if the command is outputting messages, if there is no output for more than 10 mins it kills the job, `travis_wait`  increases that to 20. we shouldn't need more than 20. Job can still run for more than 20 mins regardless of this

tests/e2e/e2e.go
```diff
@@ -111,37 +113,36 @@ func pollNonRetCmdStdOutForString(cmdStr string, timeout time.Duration, check fu
...
-	buff := make([]byte, 1048576)
-	tick := time.Tick(1 * time.Second)
+	tick := time.Tick(1 * time.Millisecond)
 
 	if err := cmd.Start(); err != nil {
 		return false, err
 	}
 
+	go func() {
+		cmd.Wait()
```
golangcibot: Error return value of `cmd.Wait` is not checked

surajnarwade:
appveyor is failing :(
codewise LGTM

geoand:
Can we also have the GolangCI issues addressed?

tests/e2e/e2e.go
```diff
@@ -100,48 +101,48 @@ func waitForServiceStatusCmd(cmd string, status string) bool {
...
```
golangcibot: Error return value of `cmd.Wait` is not checked

tests/e2e/cmp_test.go
kadel: golangcibot has a point here ;-)

anmolbabu: There are actually 2 cases not sure what single case means 🤔

anmolbabu: Wondering if it means can continually poll instead of at an interval poll that’s done currently given that the interval is too small?

anmolbabu: @kadel aah I mistook it to some other place...
and thought this is not relevant
fixed it now I put "Work In Progress" for the moment so I can see the tests pass :crossed_fingers: 
and then remove the label :wink:

anmolbabu: @kadel Done looks like the tests are green atleast on my private repo that has this PR

PR #847, opened 2018-10-18
Improve odo link support

geoand:
This PR makes the handling of `odo link` better for the user.
This is done by doing the following:

* Provide a `wait` flag that when enabled wait for the service to be provisioned and the secret to be created, before updating the DeploymentConfig. If the flag is not enabled, the command fails with the appropriate message
* Slightly improve the error messages to give even more information about what is going on
* Provide some "smart" statuses (as part of `odo service list`) for the Services that have been linked and / or bound. This makes it easier for the user to determine the status of the service provisioning

Fixes: #846

PR #976, opened 2018-11-14
Update the service catalog test to take into account the detailed status and improve robustness

geoand:
**What is the purpose of this change? What does it change?**

Robustness is improved by introducing an extra wait step for the
ServiceInstance.
Completeness is improved by updating the list test to take into account
the detailed status. This is needed because of #847 was merged

**Was the change discussed in an issue?**

No

**How to test changes?**

Not part of odo proper, this only affects the e2e tests


kadel:
e2e test are passing, mac brew install is failing, that is known issue, merging

PR #987, opened 2018-11-15
add watch flag for create cmd

syamgk:

Adding `-w` `--watch` flag to `odo create` 

https://github.com/redhat-developer/odo/issues/628

How to test changes?

`$odo create python --git https://github.com/openshift/django-ex.git`
 see if it exit quickly without showing logs

`odo create python -w --git https://github.com/openshift/django-ex.git`
see if this cmd streams the build log and wait for it to finish


cmd/create.go
```diff
@@ -155,8 +156,13 @@ A full list of component types that can be deployed is available using: 'odo cat
 			fmt.Printf("Triggering build from %s.\n\n", componentGit)
 
 			// Git is the only one using BuildConfig since we need to retrieve the git
-			err = component.Build(client, componentName, applicationName, true, true, stdout)
+			err = component.Build(client, componentName, applicationName, componentWatch, componentWatch, stdout)
 			odoutil.CheckError(err, "")
+			// if watch flag is not present
+			if !componentWatch {
+				fmt.Printf("This may take few moments to be ready\n")
+				return
```
mik-dass: I think there is no reason to return here. We should let the control flow and display the component related info at the end. Also I think the above print statement is unnecessary too.

geoand:
Agree with review from @kadel

pkg/component/component.go
```diff
@@ -223,6 +223,11 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
 		return err
 	}
 
+	if params.Watch {
+		selector := fmt.Sprintf("deployment=%s-%s-1", labels["app.kubernetes.io/component-name"], labels["app"])
```
kadel: this is not correct, this will work only for pods from the first deployment!
If deployment was triggered the label will have `-2`, when another deployment is triggered it will `-3`.... 

You can see how to get to the label for the pod from the component name here: https://github.com/redhat-developer/odo/blob/b168907c8430b78b8ae97841e21846f94f3783f3/pkg/component/component.go#L232

tests/e2e/java_test.go
```diff
@@ -94,7 +94,7 @@ var _ = Describe("odoJavaE2e", func() {
 		It("Should be able to deploy a spring boot uberjar file using openjdk", func() {
 			importOpenJDKImage()
 
-			runCmd("odo create openjdk18 sb-jar-test --binary " + javaFiles + "/openjdk/sb.jar")
+			runCmd("odo create openjdk18 sb-jar-test --binary " + javaFiles + "/openjdk/sb.jar -w")
```
kadel: shouldn't there be a test that actually verifies that command waited?

syamgk: ack will add that

syamgk: add validation using the output of create cmd

kadel:
the command should  be `--wait` not `--watch`

pkg/component/component.go
```diff
@@ -245,6 +245,12 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
 		return err
 	}
 
+	if params.Wait {
+		podSelector := fmt.Sprintf("deploymentconfig=%s-%s", labels["app.kubernetes.io/component-name"], labels["app"])
```
kadel: this should use constants from labels package.
And also this function https://github.com/redhat-developer/odo/blob/151760818982a65491425422ac5ad8fb28e44ae2/pkg/util/util.go#L47:6

pkg/odo/cli/component/create.go
```diff
@@ -172,6 +173,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
+		fmt.Printf("Initializing '%s' component ...\n", componentName)
```
cdrage: You should have it like this instead:
`log.Successf("Initializing '%s' component", componentName)`

No newline at the end or the `...`

pkg/component/component.go
cdrage: This would be a good place to add the spinner.

For example, you'd use:

```go
s:= log.Spinner("Waiting for Pod to come up")
defer s.End(false)
```

pkg/component/component.go
```diff
@@ -245,6 +245,12 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
...
+		_, err = client.WaitAndGetPod(podSelector)
+		return err
```
cdrage: why aren't we checking the error and instead just returning it?

We have no error checking here... We should have:

```go
if err != nil {
		return err
}
s.End(true)
```

syamgk:
@cdrage do i need to use any specific shell, terminal to get the spinner working as shown in the gif ?
for me on my mate-terminal, bash its just an underscore and it changes color to green once done, is that the expected behavior? am a bit confused here while testing it locally

cdrage:
@syamgk Have you tried it on any other journal? As long as you have UTF-8 installed with the correct "dots" as shown here: https://github.com/redhat-developer/odo/blob/master/pkg/log/fidget/spinner.go#L33 it will work. I'm not too sure why you are getting underscores. 

Could you push the code and I can test it?

syamgk:
@cdrage I've pushed my changes, it will be nice if you can take a look at this

pkg/odo/cli/component/create.go
```diff
@@ -324,6 +333,7 @@ func NewCmdCreate() *cobra.Command {
 	componentCreateCmd.Flags().StringVar(&cpuMax, "max-cpu", "", "Limit maximum amount of cpu to be allocated to the component. ex. 1")
 	componentCreateCmd.Flags().StringSliceVarP(&componentPorts, "port", "p", []string{}, "Ports to be used when the component is created (ex. 8080,8100/tcp,9100/udp")
 	componentCreateCmd.Flags().StringSliceVar(&componentEnvVars, "env", []string{}, "Environmental variables for the component. For example --env VariableName=Value")
+	componentCreateCmd.Flags().BoolVarP(&componentWait, "wait", "w", false, "If present, wait until the component is ready")
```
amitkrout: ```If present``` in the message is unnecessary overloading the same sense IMO. It is understandable how flags work.

cdrage: I agree, just do: `Wait until the component is ready`

pkg/odo/cli/component/create.go
```diff
@@ -206,7 +208,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 			odoutil.CheckError(err, "")
 
 			// Git is the only one using BuildConfig since we need to retrieve the git
-			err = component.Build(client, componentName, applicationName, true, stdout)
+			err = component.Build(client, componentName, applicationName, componentWait, stdout)
```
cdrage: can you use "wait" as a variable name instead of "componentWait"? Makes more sense.

tests/e2e/java_test.go
```diff
@@ -32,7 +32,8 @@ var _ = Describe("odoJavaE2e", func() {
 		It("Should be able to deploy a git repo that contains a wildfly application", func() {
 
 			// Deploy the git repo / wildfly example
-			runCmd("odo create wildfly javaee-git-test --git " + warGitRepo)
+			cmpCreateLog := runCmd("odo create wildfly javaee-git-test --git " + warGitRepo + " -w")
```
kadel: we should also have some java tests that don't have a wait flag set.

pkg/component/component.go
```diff
@@ -245,6 +245,23 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
...
+		}
+		s := log.Spinner("Waiting for Pod to come up")
+		defer s.End(false)
+		podSelector := fmt.Sprintf("deploymentconfig=%s", selectorLabels)
+		_, err = client.WaitAndGetPod(podSelector)
+		if err != nil {
+			return err
+		}
+		s.End(true)
+		return nil
```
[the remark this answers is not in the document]
syamgk: added comments

tests/e2e/java_test.go
```diff
@@ -50,8 +51,35 @@ var _ = Describe("odoJavaE2e", func() {
...
 		})
 
+		It("Should be able to deploy a git repo that contains a wildfly application without wait flag", func() {
+
+			// Deploy the git repo / wildfly example
+			cmpCreateLog := runCmd("odo create wildfly wo-wait-javaee-git-test --git " + warGitRepo)
+			Expect(cmpCreateLog).Should(ContainSubstring("This may take few moments to be ready"))
+			dcName := runCmd("oc get dc | grep wo-wait-javaee-git-test | cut -f 1 -d ' '")
+			// capturing the logs and waiting for the build to finish
+			buildLog := runCmd("oc logs --version=1 dc/" + dcName)
```
golangcibot: ineffectual assignment to `buildLog` (from `ineffassign`)

tests/e2e/java_test.go
syamgk: https://github.com/redhat-developer/odo/pull/987/files#diff-3db9664e2b8df4cc9a81bb49c3b3303cR57

surajnarwade:
@syamgk , CI is failing :cry:

tests/e2e/java_test.go
```diff
@@ -51,6 +51,32 @@ var _ = Describe("odoJavaE2e", func() {
...
 		})
 
+		It("Should be able to deploy a git repo that contains a wildfly application without wait flag", func() {
+
+			// Deploy the git repo / wildfly example
+			cmpCreateLog := runCmd("odo create wildfly wo-wait-javaee-git-test --git " + warGitRepo)
+			Expect(cmpCreateLog).Should(ContainSubstring("This may take few moments to be ready"))
+			dcName := runCmd("oc get dc | grep wo-wait-javaee-git-test | cut -f 1 -d ' '")
+			// following the logs and waiting for the build to finish
+			_ = runCmd("oc logs --version=1 dc/" + dcName)
```
amitkrout: It won't handle if ```oc logs --version=1 dc/" + dcName``` error out. So try something like 
```
stdout, err = runCmd("oc logs --version=1 dc/" + dcName)
if err != nil {
		return fmt.Errorf("Error: %v", err)
	}
[...]
```
And change accordingly the ```runCmd()``` so that it returns stdout and stderr. For example
```
func runCmd(cmdS string) string, error {
[...]
return string(session.Out.Contents()), err
```

pkg/odo/cli/component/create.go
```diff
@@ -220,8 +222,8 @@ A full list of component types that can be deployed is available using: 'odo cat
 			odoutil.LogErrorAndExit(err, "")
 
 			// Git is the only one using BuildConfig since we need to retrieve the git
-			err = component.Build(client, componentName, applicationName, true, stdout)
-			odoutil.LogErrorAndExit(err, "")
+			err = component.Build(client, componentName, applicationName, wait, stdout)
+			odoutil.CheckError(err, "")
```
amitkrout: @syamgk This line causing the CI failure. Please have a look

tests/e2e/java_test.go
syamgk: error handling for `runCmd` is happening here https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e.go#L23
also if you are suggesting to modify the implementation of `runCmd` I think we can do it in a separate issue

pkg/odo/cli/component/create.go
syamgk: yep thanks, fixed

kadel:
> or should we stream the build logs by default when `-w` is present

`-w` shouldn't change the default behaviour (showing or not showing logs).
If no logs is shown without `-w` that no log should be shown with `-w` unless user provides `--show-logs`

kadel:
@syamgk There has been a lot  of changes in master since you last pushed to this PR. Can you please rebase this to make sure that there  are no problems?

@cdrage @mik-dass Please check this PR again

mik-dass:
> True, I think it will be apt to add another flag which will steam build logs?
or should we stream the build logs by default when -w is present

I believe there is `--show-logs` for that

> this will take way lesser time than waiting for build to finish,
but it will take the time to finish all the API calls which we require for the component creation

It's not waiting for the API calls to finish. It's waiting for the build to finish in case of git components. I just confirmed it. In case of git components with no wait

```
[mrinaldas@localhost odo]$ odo create nodejs --git https://github.com/sclorg/nodejs-ex
 ✓  Checking component
 ✓  Checking component version
 ✓  Initializing 'nodejs-ex-nodejs-ktyb' component
 ✓  Creating component nodejs-ex-nodejs-ktyb
 ✓  Triggering build from git
 ✓  Component 'nodejs-ex-nodejs-ktyb' was created and port 8080/TCP was opened
 ✓  Component 'nodejs-ex-nodejs-ktyb' is now set as active component
This may take few moments to be ready
```
This message `few moments to be ready` seems useless to me. In case of git components, the `--wait` flag doesn't make sense. No matter whether I apply it or not the result is the same. I believe we should change the behaviour and make the shell exit immediately in case of git components too like local/binary. I believe there is a issue for that too. @kadel @cdrage WDYT?

tests/e2e/java_test.go
```diff
@@ -32,7 +32,8 @@ var _ = Describe("odoJavaE2e", func() {
 		It("Should be able to deploy a git repo that contains a wildfly application", func() {
 
 			// Deploy the git repo / wildfly example
-			runCmd("odo create wildfly javaee-git-test --git " + warGitRepo)
+			cmpCreateLog := runCmd("odo create wildfly javaee-git-test --git " + warGitRepo + " -w")
+			Expect(cmpCreateLog).ShouldNot(ContainSubstring("This may take few moments to be ready"))
```
mik-dass: I don't think it's a reliable way of detecting the status. Maybe try `oc get build <build-name>`?

tests/e2e/java_test.go
```diff
@@ -50,8 +51,35 @@ var _ = Describe("odoJavaE2e", func() {
 			runCmd("odo delete javaee-git-test -f")
 		})
 
+		It("Should be able to deploy a git repo that contains a wildfly application without wait flag", func() {
+
+			// Deploy the git repo / wildfly example
+			cmpCreateLog := runCmd("odo create wildfly wo-wait-javaee-git-test --git " + warGitRepo)
+			Expect(cmpCreateLog).Should(ContainSubstring("This may take few moments to be ready"))
```
mik-dass: Same here, Imaybe try `oc get build <build-name>`?

tests/e2e/java_test.go
syamgk: @mik-dass added validation using oc here https://github.com/redhat-developer/odo/pull/987/commits/e534ba62eac87b40728fe60310fe226388be232d

tests/e2e/java_test.go
syamgk: validation using oc here e534ba6

cdrage:
Are you able to fix tests at all? Seems like Java is failing at the moment.

tests/e2e/java_test.go
```diff
@@ -50,8 +56,43 @@ var _ = Describe("odoJavaE2e", func() {
...
+			buildName := runCmd("oc get builds --output='name' | grep wo-wait-javaee-git-test | cut -d \"/\" -f 2")
+			Expect(buildName).To(ContainSubstring("wo-wait-javaee-git-test"))
```
mik-dass: Also please check that the build status is running

tests/e2e/java_test.go
```diff
@@ -50,8 +56,43 @@ var _ = Describe("odoJavaE2e", func() {
...
+
+			// Push changes
+			runCmd("odo push")
+
+			// Create a URL
+			runCmd("odo url create")
+			routeURL := determineRouteURL()
+
+			// Ping said URL
+			waitForEqualCmd("curl -s "+routeURL+" | grep 'Insult' | wc -l | tr -d '\n'", "1", 10)
```
mik-dass: why are we not evaluating the returned bool?

tests/e2e/java_test.go
mik-dass: @syamgk Why was this marked as resolved?

syamgk: i've added the status check in line number 71

syamgk: okey I
 got it the 'oc get build' and build status

mik-dass: :+1:

tests/e2e/java_test.go
```diff
@@ -94,7 +155,12 @@ var _ = Describe("odoJavaE2e", func() {
 		It("Should be able to deploy a spring boot uberjar file using openjdk", func() {
 			importOpenJDKImage()
 
-			runCmd("odo create openjdk18 sb-jar-test --binary " + javaFiles + "/openjdk/sb.jar")
+			cmpCreateLog := runCmd("odo create openjdk18 sb-jar-test --binary " + javaFiles + "/openjdk/sb.jar -w")
+			Expect(cmpCreateLog).ShouldNot(ContainSubstring("This may take few moments to be ready"))
+
+			dcName := runCmd("oc get dc | grep sb-jar-test| cut -f 1 -d ' '")
+			Expect(dcName).To(ContainSubstring("sb-jar-test"))
```
mik-dass: Check here that the build status is complete

syamgk: In this case (--binary) there is no build

mik-dass: Oh OK

kadel:
> @syamgk Would you please have a look into the UTs to fix decrease in coverage.

covering this is by UT going to be tricky as it needs to communicate with a cluster. Mocking this is going to be pretty complicated  :-(
Decreased coverage is there because of this PR add new lines to a pretty complicated function that is not covered by UT :-(
I agree that this should be covered, but that will be a big effort, worth of its separate PR.

PR #991, opened 2018-11-15
Add memory limits to create and also use the resource limits for update

### LINKED ISSUE #907: add memory flags to component creation
[kind/Feature]

We need to add a flag to `odo create` to be able to establish the memory the runtime container will have. Some applications are memory hungry and in some environments the defaults that are assigned when nothing is specified do not work. 

I would think we should add, at a minimum a flag for memory, whether it's `memory` or `mem` that should set request and limits for the container with the runtime in the pod.

Additionally it would be nice that we could set `min-memory` and `max-memory` as flags to set requests on the former and limits on the latter.

`min-memory` and  `max-memory` should have priority over `memory` if all of them are set, although it would be desirable to emit a warning advising the user that in this case `memory` flag is superfluous and could be omitted.

anmolbabu:
Add memory limits to create and also use the resource limits for update

fixes #907


anmolbabu:
ToDo(@anmolbabu):
- [ ] Fix UTs
- [ ] Changes to Patch functions to include previously maintained resource limits
- [ ] Update commit msg

anmolbabu:
@kadel @metacosm @cdrage @geoand @surajnarwade @syamgk @mik-dass Please review

pkg/util/util.go
```diff
@@ -249,3 +258,26 @@ func OpenBrowser(url string) error {
...
 }
+
+// FetchResourceQunatity takes passed min, max and requested resource quantities and returns min and max resource requests
+func FetchResourceQunatity(resourceType corev1.ResourceName, min string, max string, request string) ResourceRequirementInfo {
+	// If minimum and maximum both are passed they carry highest priority
+	// Otherwise, use the request as min and max
+	var minResource resource.Quantity
+	var maxResource resource.Quantity
+	if min != "" {
+		minResource = resource.MustParse(min)
```
girishramnani: Would it be a good idea to return an error here and use `odoutil.CheckError(err, "")` in the `create.go`?

anmolbabu: Ok I think you meant returning an error if a wrong param is passed(vs an error @line 269), in which case the underlying lib raises a panic ... and hence odo would then fail

Regarding, ensuring that atleast one of the flags is set without which, we don't venture into setting this config in the dc, I'll add code handling it

anmolbabu: > Ok I think you meant returning an error if a wrong param is passed like an incorrect unit(vs an error @line 269), in which case the underlying lib raises a panic ... and hence odo would then fail
> 
> Regarding, ensuring that atleast one of the flags is set without which, we don't venture into setting this config in the dc, I'll add code handling it

I just tried to verify the case of wrong units it seems the pod creation fails but not in odo layer :(
@kadel Should we add custom validation of units ?
Like if I pass `--memory 300` instead of `--memory 300Mi` the vendored package("k8s.io/apimachinery/pkg/api/resource") doesn't seem to return error or panic as the func doc suggests

kadel:
I'm a little bit afraid that this will require a little bit of code refactoring.
We can't no longer create the components in the same way as we are doing it right now.

`component.CreateFromPath` and `component.CreateFromGit` are starting to have too many arguments.
I think it is time to refactor those function, especially as we will be adding more and more parameters to this.

one of the possible approaches might be creating a struct that holds all the required info about a component, and then just calling something like  `newcomponent.create()`

What do you think @anmolbabu ?

anmolbabu:
> I'm a little bit afraid that this will require a little bit of code refactoring.
> We can't no longer create the components in the same way as we are doing it right now.
> 
> `component.CreateFromPath` and `component.CreateFromGit` are starting to have too many arguments.
> I think it is time to refactor those function, especially as we will be adding more and more parameters to this.
> 
> one of the possible approaches might be creating a struct that holds all the required info about a component, and then just calling something like `newcomponent.create()`
> 
> What do you think @anmolbabu ?

Sure @kadel I did it for watch in my other PR.. :)
I'll do it here as well

pkg/util/util.go
anmolbabu: This fails although:

```
[redhat@localhost odo]$ odo create nodejs --local ~/git-srcs/src/github.com/redhat-developer/odo-samples/nodejs-ex -v 4 --min-memory 200uu
I1119 19:22:31.134385    3083 create.go:76] Component create called with args: "nodejs", flags: binary=, git=, local=/home/redhat/git-srcs/src/github.com/redhat-developer/odo-samples/nodejs-ex
I1119 19:22:31.137483    3083 occlient.go:365] Trying to connect to server 192.168.42.137:8443
I1119 19:22:31.137784    3083 occlient.go:371] Server https://192.168.42.137:8443 is up
I1119 19:22:31.171986    3083 occlient.go:294] isLoggedIn err:  <nil> 
 output: "developer"
I1119 19:22:31.172229    3083 context.go:202] Ignoring error as it usually means flag wasn't set: flag accessed but not defined: component
I1119 19:22:31.237277    3083 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 10 4 6 8 8-RHOAR latest]} {perl openshift [5.16 5.20 5.24 5.26 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 13.0 8.1 9.0 latest]}]
I1119 19:22:31.258837    3083 catalog.go:161] Found builder images: [{dotnet openshift [2.0 latest]} {httpd openshift [2.4 latest]} {nginx openshift [1.10 1.12 1.8 latest]} {nodejs openshift [0.10 10 4 6 8 8-RHOAR latest]} {perl openshift [5.16 5.20 5.24 5.26 latest]} {php openshift [5.5 5.6 7.0 7.1 latest]} {python openshift [2.7 3.3 3.4 3.5 3.6 latest]} {ruby openshift [2.0 2.2 2.3 2.4 2.5 latest]} {wildfly openshift [10.0 10.1 11.0 12.0 13.0 8.1 9.0 latest]}]
panic: cannot parse '200uu': unable to parse quantity's suffix

goroutine 1 [running]:
github.com/redhat-developer/odo/vendor/k8s.io/apimachinery/pkg/api/resource.MustParse(0x7ffe0ce3001e, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go:133 +0x249
github.com/redhat-developer/odo/pkg/util.FetchResourceQunatity(0x1836baf, 0x6, 0x7ffe0ce3001e, 0x5, 0x0, 0x0, 0x0, 0x0, 0x4)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/pkg/util/util.go:272 +0x306
github.com/redhat-developer/odo/cmd.glob..func22(0x2aabbe0, 0xc000b16150, 0x1, 0x7)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/cmd/create.go:162 +0x748
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).execute(0x2aabbe0, 0xc000b160e0, 0x7, 0x7, 0x2aabbe0, 0xc000b160e0)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:702 +0x2d3
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x2aadbc0, 0x8, 0xc000000300, 0x145ae8e)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:783 +0x2dc
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).Execute(0x2aadbc0, 0x1935ac0, 0xc000122720)
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:736 +0x2b
github.com/redhat-developer/odo/cmd.Execute()
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/cmd/root.go:108 +0xae
main.main()
	/home/redhat/git-srcs/src/github.com/redhat-developer/odo/main.go:40 +0x1d1
```

anmolbabu:
@kadel Done...
Please review
Cc: @cdrage @surajnarwade @metacosm @geoand @syamgk @mik-dass

cmd/create.go
```diff
@@ -91,7 +96,7 @@ A full list of component types that can be deployed is available using: 'odo cat
 		}
 		if len(componentLocal) != 0 {
 			componentPath = componentLocal
-			componentPathType = component.SOURCE
+			componentPathType = component.LOCAL
```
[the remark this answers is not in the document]
anmolbabu: For uniformity :D
Some of the code used component.Source while some used hard-coded "local" ...
Out of laziness I found which was used the least and changed that to majority ;) including changing the hard-coded ones to using constants

cmd/create.go
```diff
@@ -146,11 +151,45 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
+		// If min-memory, max-memory and memory are passed, memory will be ignored as the other 2 have greater precedence.
+		// Emit a message indicating the same
+		if memoryMin != "" && memoryMax != "" {
+			if memory != "" {
```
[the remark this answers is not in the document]
anmolbabu: Ack..
But i wanted the cases to be explicit i.e, memoryMin != "" && memoryMax !="" && memory == "" is not handled here bcoz that'll be handled down the line and also, readability is better when split(in my opinion)..
Anyway I'll fix this here...

cmd/create.go
```diff
@@ -146,11 +151,45 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+				fmt.Printf("%s will be ignored as minimum memory %s and maximum memory %s passed carry greater precedence\n", memory, memoryMin, memoryMax)
```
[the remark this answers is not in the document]

cmd/create.go
```diff
@@ -146,11 +151,45 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+			}
+		}
+		if (memoryMin == "") != (memoryMax == "") {
+			if memory != "" {
```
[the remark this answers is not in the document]
anmolbabu: I think readability will be lost by doing so :disappointed:

cmd/create.go
```diff
@@ -146,11 +151,45 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+				fmt.Printf("Using memory %s for min and max limits.\nIf you want memory min and memory max specifically, please ensure to pass both of them\n", memory)
```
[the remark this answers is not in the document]
anmolbabu: Ack Done

cmd/create.go
```diff
@@ -221,6 +296,9 @@ func init() {
 	componentCreateCmd.Flags().StringVarP(&componentBinary, "binary", "b", "", "Use a binary as the source file for the component")
 	componentCreateCmd.Flags().StringVarP(&componentGit, "git", "g", "", "Use a git repository as the source file for the component")
 	componentCreateCmd.Flags().StringVarP(&componentLocal, "local", "l", "", "Use local directory as a source file for the component")
+	componentCreateCmd.Flags().StringVar(&memory, "memory", "", "Amount of memory to be allocated to the component container. Ex: 100Mi")
+	componentCreateCmd.Flags().StringVar(&memoryMin, "min-memory", "", "Limit minimum amount of memory to be allocated to the component container. Ex: 100Mi")
```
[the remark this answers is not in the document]
anmolbabu: `Ack`

pkg/component/component.go
```diff
@@ -140,12 +152,13 @@ func validateSourceType(sourceType string) bool {
 // CreateFromGit inputPorts is the array containing the string port values
 // inputPorts is the array containing the string port values
 // envVars is the array containing the environment variables
-func CreateFromGit(client *occlient.Client, name string, componentImageType string, url string, applicationName string, inputPorts []string, envVars []string) error {
+// func CreateFromGit(client *occlient.Client, name string, componentImageType string, url string, applicationName string, inputPorts []string, envVars []string, resources []util.ResourceRequirementInfo) error {
```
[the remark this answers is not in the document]
anmolbabu: Aah yes removing it now

pkg/component/component.go
```diff
@@ -200,11 +213,12 @@ func GetComponentPorts(client *occlient.Client, componentName string, applicatio
 // CreateFromPath create new component with source or binary from the given local path
 // sourceType indicates the source type of the component and can be either local or binary
 // envVars is the array containing the environment variables
-func CreateFromPath(client *occlient.Client, name string, componentImageType string, path string, applicationName string, sourceType string, inputPorts []string, envVars []string) error {
-	labels := componentlabels.GetLabels(name, applicationName, true)
+//func CreateFromPath(client *occlient.Client, name string, componentImageType string, path string, applicationName string, sourceType string, inputPorts []string, envVars []string, resources []util.ResourceRequirementInfo) error {
```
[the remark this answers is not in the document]
anmolbabu: Aah yes removing it now

pkg/component/component.go
```diff
@@ -232,7 +246,7 @@ func CreateFromPath(client *occlient.Client, name string, componentImageType str
 	}
 
 	// Bootstrap the deployment with SupervisorD
-	err = client.BootstrapSupervisoredS2I(name, commonObjectMeta, componentImageType, inputPorts, envVars)
+	err = client.BootstrapSupervisoredS2I(params.Name, commonObjectMeta, params.ImageName, params.Ports, params.EnvVars, params.Resources)
```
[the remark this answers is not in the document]
anmolbabu: Ack Done

anmolbabu:
@kadel @metacosm @cdrage @geoand @surajnarwade @syamgk @mik-dass Please review

geoand:
@anmolbabu Hey! Just getting back up to speed after being out for a few days. I assume this is perfectly ready for me to check out? 
As you know I have a larger interest than most in the PR since I will be using it as a base for the cpu work :)

cmd/create.go
```diff
@@ -146,11 +152,42 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+		if memoryQuantity != nil {
+			resourceQuantity = append(resourceQuantity, *memoryQuantity)
+		}
+
 		// Deploy the component with Git
 		if len(componentGit) != 0 {
 
 			// Use Git
-			err := component.CreateFromGit(client, componentName, componentImageName, componentGit, applicationName, componentPorts, componentEnvVars)
+			err := component.CreateFromGit(
```
geoand: This was needed so bad!

pkg/util/util.go
mik-dass: I tried the command
```
odo create nodejs --max-memory 300 --min-memory 100
```
The pod creation failed but the command
```
odo create nodejs --max-memory 300M --min-memory 100
```
seems to work. Weird that passing a unit for one worked for me.

cmd/create.go
```diff
@@ -146,11 +152,42 @@ A full list of component types that can be deployed is available using: 'odo cat
...
 
+		// If min-memory, max-memory and memory are passed, memory will be ignored as the other 2 have greater precedence.
+		// Emit a message indicating the same
+		if memoryMin != "" && memoryMax != "" && memory != "" {
+			fmt.Println("Flag `memory` will be ignored as `min-memory` and `max-memory` have been passed ")
+		}
+		if (memoryMin == "") != (memoryMax == "") && memory != "" {
+			fmt.Printf("Using memory %s for min and max limits.\n", memory)
+		}
+		if (memoryMin == "") != (memoryMax == "") && memory == "" {
```
geoand: Is this a hard requirement? I mean it's definitely good practice to specify both, but is there a reason we have placed a hard requirement on having to specify both min and max fields or none at all?

cmd/create.go
```diff
@@ -146,11 +152,42 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+			fmt.Println("memoryMin should accompany memoryMax or pass memory to use same value for both min and max or try not passing any of them")
+			os.Exit(1)
+		}
+
+		memoryQuantity := util.FetchResourceQunatity(corev1.ResourceMemory, memoryMin, memoryMax, memory)
```
geoand: This is a minor typo, let's fix it before the PR is merged :)

cmd/create.go
```diff
@@ -24,6 +27,9 @@ var (
 	componentLocal   string
 	componentPorts   []string
 	componentEnvVars []string
+	memoryMax        string
+	memoryMin        string
+	memory           string
 )
 
 var componentCreateCmd = &cobra.Command{
```
surajnarwade: Add examples in `Example` section

cmd/create.go
```diff
@@ -221,6 +294,9 @@ func init() {
 	componentCreateCmd.Flags().StringVarP(&componentBinary, "binary", "b", "", "Use a binary as the source file for the component")
 	componentCreateCmd.Flags().StringVarP(&componentGit, "git", "g", "", "Use a git repository as the source file for the component")
 	componentCreateCmd.Flags().StringVarP(&componentLocal, "local", "l", "", "Use local directory as a source file for the component")
+	componentCreateCmd.Flags().StringVar(&memory, "memory", "", "Amount of memory to be allocated to the component container. ex. 100Mi")
```
surajnarwade: remove `container` word, we are assuming our user as noob :baby:

cmd/create.go
```diff
@@ -221,6 +294,9 @@ func init() {
...
+	componentCreateCmd.Flags().StringVar(&memoryMin, "min-memory", "", "Limit minimum amount of memory to be allocated to the component container. ex. 100Mi")
```
surajnarwade: same here

cmd/create.go
```diff
@@ -221,6 +294,9 @@ func init() {
...
+	componentCreateCmd.Flags().StringVar(&memoryMax, "max-memory", "", "Limit maximum amount of memory to be allocated to the component container. ex. 100Mi")
```
surajnarwade: here too

cmd/create.go
anmolbabu: no min and max both or only memory(in which case internally we set min and max both to the same value)
only to avoid confusion
And setting both of them is mandated by openshift based on my experiments

geoand: Understood :)

cmd/create.go

cmd/create.go

cmd/create.go

cmd/create.go

cmd/create.go

cmd/create.go
```diff
@@ -146,11 +152,42 @@ A full list of component types that can be deployed is available using: 'odo cat
...
+		// If min-memory, max-memory and memory are passed, memory will be ignored as the other 2 have greater precedence.
+		// Emit a message indicating the same
+		if memoryMin != "" && memoryMax != "" && memory != "" {
+			fmt.Println("Flag `memory` will be ignored as `min-memory` and `max-memory` have been passed ")
+		}
+		if (memoryMin == "") != (memoryMax == "") && memory != "" {
+			fmt.Printf("Using memory %s for min and max limits.\n", memory)
+		}
+		if (memoryMin == "") != (memoryMax == "") && memory == "" {
+			fmt.Println("memoryMin should accompany memoryMax or pass memory to use same value for both min and max or try not passing any of them")
```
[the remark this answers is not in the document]

pkg/models/component.go
```diff
@@ -0,0 +1,29 @@
+package models
```
kadel: I like this, but shouldn't this be in `component` package?
Everything here is related to the component.

Good go practice is that definitions like this are in `types.go` file in the package that they are related to.

anmolbabu: Yeah but I just did this to avoid cyclic dependencies 
OcClient also has references to this struct which makes occlient depend on package containing this.. and component pkg references occlient

kadel: than this indicates flaw in our design :-|

How about moving this to `occlient`?

kadel: OK, I will answer that myself, that is won't be nice :-)

anmolbabu: Would that be ok ?

kadel: not ideal :-( ideally this should be in `component` and functions in `occlient` should use a different structure. We need to decide for what those arguments are for. 
Is it for functions for components or for functions for `occlient`?

anmolbabu: Ok but I felt the same struct can be used instead of creating one more

cmd/create.go
```diff
@@ -146,11 +156,42 @@ A full list of component types that can be deployed is available using: 'odo cat
 			os.Exit(1)
 		}
 
+		// If min-memory, max-memory and memory are passed, memory will be ignored as the other 2 have greater precedence.
+		// Emit a message indicating the same
+		if memoryMin != "" && memoryMax != "" && memory != "" {
+			fmt.Println("Flag `memory` will be ignored as `min-memory` and `max-memory` have been passed ")
```
cdrage: This seems weird compared to others..

It should be `--memory will be ignored as --min-memory and --max-memory has been passed`

kadel:
> Wouldn't --max-mem and --min-mem be better (smaller + less to type)?

I would say that memory is ok in this case. It is just 3 letters, and to be honest I like full words ;-)

pkg/models/component.go
kadel: Ideally, we should have nice abstraction levels, where the component is one level and occlient is second. component can use lover level abstractions like occlient. And lower-level abstractions shouldn't know anything about higher levels. We shouldn't import anything from component in occlient. We already broke this multiple times, and this is what is causing all the cyclic imports :-(

kadel: Having a separate package with just "models" can be tricky.
To make it work you have to limit imports in that package, ideally, don't have any. You are already using `utils` there, that means that now you can't use anything from `models` in utils.

Having a separate package like this also prevents you from creating receiver methods for those types, because you would have to put those functions to the same package, this would resolve into adding more imports .....

anmolbabu: Hmm so that means https://github.com/redhat-developer/odo/pull/991#discussion_r235377770 makes sense

anmolbabu: @kadel Done I put it finally in occlient

Its actually almost the same set of params that are used from the cmd layer all the way upto occlient with off-course a few of them not used in some of the occlient functions related to create
but overall its almost the same.. So, based on this discussion I felt it makes sense to move it to occlient..

pkg/occlient/occlient.go
```diff
@@ -57,6 +57,32 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+// CreateType is an enum to indicate the type of source of component -- local source/binary or git for the generation of app/component names
+type CreateType string
```
kadel: This is used only in `component` if I'm looking correctly. Why was this moved here?

geoand:
@anmolbabu `make validate` is failing on Travis (and locally when I tested it)

anmolbabu:
Merging based on 2 acks

> @anmolbabu `make validate` is failing on Travis (and locally when I tested it)

Fixed now It seems to be due to a new param added by one of the merged PRs(possibly mine ;) ) but failure is not relevant to master though

PR #1001, opened 2018-11-16
Slim down issue templates a bit

kadel:
Most of the comments in the issue templates were just repeating again what is in the header. We can get rid of them, to make filling issues a bit pleasent experience.


cdrage:
This looks way better. LGTM!

PR #1002, opened 2018-11-16
Move commands to where they should be

### LINKED ISSUE #785: Refactor CLI code to follow kubectl/oc practices
Our CLI code is starting to get quite messy. 
It is time to start working on refactoring it.

The best way would be to follow  Complete-Validate-Run pattern used by kubectl and oc.

Resources that will help with refactoring:
- Kubernetes extracted some of the kubectl stuff to re-usable libraries  https://github.com/kubernetes/cli-runtime 
- OpenShift released nice CLI Hacking Guide https://github.com/openshift/origin/blob/master/docs/cli_hacking_guide.adoc

metacosm:
Move commands to where they should be according to `oc` guidelines.

No, but will help fix #785 via #871 

## How to test changes?
Externally visible behavior shouldn't change.


metacosm:
There's an import cycle that I'm not too sure how to handle and would appreciate discussing the issue.
Basically, I'd like to move the `Add*Flag` functions to `genericclioptions` where I think they would make more sense but doing so adds an import cycle because `genericclioptions` is imported by `completion` which would in turn be imported by `genericclioptions` if we add the `Add*Flag` methods to it (because of the completion handler registration for the flags, which make sense to be done in these functions). As it is, we have many potential import cycles that I tried to step around but this is getting more and more tricky… For example, one reason why I want to move the `Add*Flag` functions (which are currently in `cli` package) is because the code doesn't work as it is, we need to invert the relation between commands and root command where the root command instantiates the child doing something like `NewCmdApplication` to add the child but doing so adds an import cycle (because then, for example, `application` imports `cli` to use the `Add*Flag` functions and `cli` will now import `application`). There's a similar issue with version checking. I'm not sure how to solve these issues, maybe have strict rules regarding the direction of imports (something like parent packages can import children but not the reverse, though that feels wrong as in OOP, you'd do the reverse i.e. a children package is more likely to be more specific and therefore it makes sense that it knows about the general case while the general case (parent package) shouldn't know about the specific implementations). Design-wise, I feel this is very constraining and leads to ugly code (as, right now, the only solution I see is to duplicate the `Add*Flag` function per package to avoid importing them) :(
I've been bashing my head on this issue since last night and this is starting to drive me nuts! :)

girishramnani:
can we consider a package inside `genericclioptions` called `flags` or `flagutils` which holds these functions? would this help removing the cycle?

girishramnani:
The reason I suggest is because the `genericclioptions` holds the `Context` code ( which holds code related to state of the flags ) and I feels that `flags` can be sub module inside it

metacosm:
> can we consider a package inside `genericclioptions` called `flags` or `flagutils` which holds these functions? would this help removing the cycle?

I tried a similar approach which didn't work… but I guess I could try this as well. However, besides this specific issue, it's getting harder and harder not to introduce import cycles… :(

girishramnani:
If that's the case we might have to consider breaking modules more
granularly or probably extract certain conflicting contructs in a sub
package.

metacosm:
That's kind of what this PR is attempting but it's actually easier to avoid cycles by not introducing new packages / modules… The cycles actually occur when you try to reuse code across packages.

kadel:
This is something that I was afraid that it might happen.

Import cycles are a symptom of bad design :-(
We will have to think a little bit more about our code structure.

metacosm:
Indeed, though in this case, I'm not sure how things could be designed differently. Maybe `Context` shouldn't be in `genericclioptions`… The biggest issue, though, is that it doesn't really what the packages are: the facts that it makes sense to have the `Add*Flag` functions also add the completion handlers and that the completion handlers rely on `Context` almost guarantee an import cycle regardless of how things are packaged. The thing that I'm not too sure about is `Context`: I'm not quite sure where it should live. It does make sense to build it when building new commands (either as part of a `PreRun` cobra hook or `Complete` function in the `oc` organization) but it doesn't strictly feel like a generic option per se. I've been racking my brains on this issue but kind of been running in circles. Maybe I need to step back a little bit from it and consider things from a newer perspective…

kadel:
> The biggest issue, though, is that it doesn't really what the packages are: the facts that it makes sense to have the `Add*Flag` functions also add the completion handlers and that the completion handlers rely on `Context` almost guarantee an import cycle regardless of how things are packaged.

Coudn't we just remove the `completion` part of the code from `Add*Flag` and handle that in the `init()` function? I know that it is not ideal. But it might be the easiest solution to move forward, so we don't get stuck on this

metacosm:
@kadel the `init` functions will disappear actually, they will be replaced by `NewCmdXXXX` functions called from the parent (which is how the import cycle actually starts showing up). We could indeed remove the completion code from the `Add*Flag` functions as a temporary measure to make progress, though (which is what I was considering) but that means duplicated code and remembering to add the completion handler where it's needed. So this feels like a bad design to get around a language limitation (there is no such thing as import cycles in Java, for example). That's not to say that the design cannot (or shouldn't) be improved but in this case, this feels like a bad tradeoff. I will proceed like this for now, though, to see how this leads us.

metacosm:
Not sure what's up with AppVeyor…

kadel:
> Not sure what's up with AppVeyor…

we can ignore that for now, test is passing, it is just that we run out of the artifacts storage there :-(

metacosm:
> > Not sure what's up with AppVeyor…
> 
> we can ignore that for now, test is passing, it is just that we run out of the artifacts storage there :-(

Actually, there was also an issue with the PR, the `cross` target wasn't working because it was still assuming the old file organization. :)

cmd/cli-doc/cli-doc.go
```diff
@@ -86,11 +89,74 @@ func referencePrinter(command *cobra.Command, level int) string {
...
+func getFlags(flags *pflag.FlagSet) []string {
+	var f []string
+	flags.VisitAll(func(flag *pflag.Flag) {
+		f = append(f, fmt.Sprintf("--%v", flag.Name))
+	})
+	return f
+}
+
+func flattenFlags(flags []string) string {
+	var flagString string
```
anmolbabu: Can we use https://golang.org/pkg/strings/#Join here?

cmd/cli-doc/cli-doc.go
```diff
@@ -86,11 +89,74 @@ func referencePrinter(command *cobra.Command, level int) string {
...
+		finalCommand = finalCommand + commandPrinter(subcommand, level+1)
+	}
+	return finalCommand
+}
+
 // Generates and returns a markdown-formatted CLI reference page for Odo
-func GenerateCLIReference() string {
-	return referencePrinter(rootCmd, 0)
+func main() {
+	var clidoc = &cobra.Command{
```
anmolbabu: I might be missing something..
But why a cobra command for this?

cmd/cli-doc/cli-doc.go
metacosm: Maybe, I haven't really looked at the code, that command existed before :)

cmd/cli-doc/cli-doc.go
metacosm: To be able to quickly support args since this command now replaces 2 commands and has to deal with args (which wasn't really the case before).

metacosm:
Thank you!

PR #1015, opened 2018-11-20
e2e: add travis_wait for test-cmp-e2e

kadel:
Increase timout to 20 minutes

Travis-ci kills jobs that don't output something for more than 10 mins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-cmp-e2e
+        - travis_wait make test-cmp-e2e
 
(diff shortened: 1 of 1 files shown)


surajnarwade:
can we increase it to 30 min, because I have seen it takes more than 20 min most of the times

cdrage:
Finally passed. This LGTM!

kadel:
> can we increase it to 30 min, because I have seen it takes more than 20 min most of the times

no, it doesn't matter how long the test takes, travis-ci check if the command is outputting messages, if there is no output for more than 10 mins it kills the job, travis_wait increases that to 20. we shouldn't need more than 20. Job can still run for more than 20 mins regardless of this.

PR #1023, opened 2018-11-21
made runFailCmd wait till cmd gets completed

girishramnani:
The `runFailCmd` wasn't waiting for the command to complete due to which the exit code of the Cmd always returned as `-1`. This occurs due to premature stopping.

No Issue

## How to test changes?
try using `runFailCmd` for a failing odo cli combination  eg. `odo create test --local test --git test`


cdrage:
Tests pass, this is the same convention as in: ` func runCmd(cmdS string) string {	`

So this LGTM :+1: I don't see anything wrong here.

PR #1027, opened 2018-11-21
Adds PodTimeout to odo utils config

cdrage:
Adds PodTimeout to odo utils in order to set the length of time to wait
for a Pod to timeout when deploying to OpenShift.

This can be tested with `odo utils config set PodTimeout 30`

Closes https://github.com/redhat-developer/odo/issues/943

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -98,9 +98,6 @@ const (
 	// Create a custom name and (hope) that users don't use the *exact* same name in their deployment
 	supervisordVolumeName = "odo-supervisord-shared-data"
 
-	// waitForPodTimeOut controls how long we should wait for a pod before giving up
-	waitForPodTimeOut = 120 * time.Second
-
 	// ComponentPortAnnotationName annotation is used on the secrets that are created for each exposed port of the component
 	ComponentPortAnnotationName = "component-port"
 
@@ -1561,6 +1558,15 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 	}
 	defer w.Stop()
 
+	// Retrieve the PodTimeout configuration parameter
+	podTimeout := config.DefaultPodTimeout * time.Second
+	cfg, configReadErr := config.New()
+	if configReadErr != nil {
+		glog.V(4).Info(errors.Wrap(configReadErr, "unable to read config file"))
+	} else {
+		podTimeout = time.Duration(time.Duration(cfg.GetPodTimeout()) * time.Second)
+	}
+
 	podChannel := make(chan *corev1.Pod)
 	watchErrorChannel := make(chan error)
 
@@ -1598,8 +1604,8 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 		return val, nil
 	case err := <-watchErrorChannel:
 		return nil, err
-	case <-time.After(waitForPodTimeOut):
-		return nil, errors.Errorf("waited %s but couldn't find running pod matching selector: '%s'", waitForPodTimeOut, selector)
+	case <-time.After(podTimeout):
+		return nil, errors.Errorf("waited %s but couldn't find running pod matching selector: '%s'", podTimeout, selector)
 	}
 }
 
--- a/pkg/odo/cli/utils/config.go
+++ b/pkg/odo/cli/utils/config.go
@@ -13,16 +13,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const availableParameters = `
+Available Parameters:
+UpdateNotification - Controls if an update notification is shown or not (true or false)
+NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix
+Timeout - Timeout (in seconds) for OpenShift server connection check
+PodTimeout - Pod Timeout (in seconds) to wait for an OpenShift Pod to start when deploying a component
+`
+
 // configurationCmd represents the app command
 var configurationCmd = &cobra.Command{
 	Use:   "config",
 	Short: "Modifies configuration settings",
 	Long: `Modifies Odo specific configuration settings within the config file.
-
-Available Parameters:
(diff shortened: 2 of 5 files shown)


pkg/occlient/occlient.go
```diff
@@ -1507,6 +1507,15 @@ func (c *Client) WaitAndGetPod(selector string) (*corev1.Pod, error) {
 	}
 	defer w.Stop()
 
+	// Retrieve the PodTimeout configuration parameter
+	podTimeout := time.Duration(config.DefaultPodTimeout * time.Second)
```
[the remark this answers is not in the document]
cdrage: Yeah, I think that this could be refactored a ton, but at the moment, I thought of implementing this the same way that the current `Timeout` is. 

Ideally, having this separated into a separate function would be better.

I'll work on getting this refactored :+1:

geoand:
It seems like the tests failed and I think that this time it's not a flake

cdrage:
Switching back to WIP. Trying to diagnose a failing test

cdrage:
Ready for review @metacosm @anmolbabu @mik-dass @surajnarwade

tests/e2e/cmp_test.go
```diff
@@ -49,6 +49,11 @@ var _ = Describe("odoCmpE2e", func() {
 
 	Context("odo component creation", func() {
 
+		// This is required as some CI's (such as TravisCI) requires a little bit more time
+		It("should set an appropriate test-configured PodTimeout", func() {
+			runCmd("odo utils config set PodTimeout 120")
```
mik-dass: But 120 is the default value right?

mik-dass:
@cdrage Please remove the WIP label if it's ready for review :)

tests/e2e/cmp_test.go
cdrage: You're right. I'll change it to a different value. 

The reason I set it to "120" again is because I previously set it to 10 and it was causing timeouts / issues..

pkg/odo/cli/utils/config.go
```diff
@@ -22,7 +22,8 @@ var configurationCmd = &cobra.Command{
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
 NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix
-Timeout - Timeout (in seconds) for OpenShift server connection check`,
+Timeout - Timeout (in seconds) for OpenShift server connection check
+PodTimeout - Pod Timeout (in seconds) to wait for an OpenShift Pod to start`,
```
kadel: odo user might not know when this will happen, or even what is a pod.

This should be explained in odo terms or commands that will this effect.

pkg/odo/cli/utils/config.go
```diff
@@ -53,12 +54,14 @@ var configurationSetCmd = &cobra.Command{
 Available Parameters:
 UpdateNotification - Controls if an update notification is shown or not (true or false)
 NamePrefix - Default prefix is the current directory name. Use this value to set a default name prefix.
-Timeout - Timeout (in seconds) for OpenShift server connection check`,
+Timeout - Timeout (in seconds) for OpenShift server connection check
+PodTimeout - Pod Timeout (in seconds) to wait for an OpenShift Pod to start`,
```
kadel: it looks a bit weird that we have in two places, could we put a description of Available Parameters into one place?

tests/e2e/e2e_test.go
```diff
@@ -90,12 +91,15 @@ var _ = Describe("odoe2e", func() {
...
+			runCmd("odo utils config set timeout 31")
+			runCmd("odo utils config set podtimeout 121")
 			configOutput := runCmd("odo utils config view|grep UpdateNotification")
 			Expect(configOutput).To(ContainSubstring("false"))
 			Expect(configOutput).To(ContainSubstring("UpdateNotification"))
 			configOutput = runCmd("odo utils config view|grep Timeout")
-			Expect(configOutput).To(ContainSubstring("5"))
+			Expect(configOutput).To(ContainSubstring("31"))
+			configOutput = runCmd("odo utils config view|grep PodTimeout")
+			Expect(configOutput).To(ContainSubstring("121"))
```
kadel: Shouldn't we have a test that actually verifies that odo is waiting given amount of time?
This just tested that you can read the value back, but not if it is actually used.

cdrage:
@kadel Just writing some simple tests now.

Would you think this is the right way to test it?

1. Set the PodTimeout to something *very* small (2 seconds)
2. Use runFailCmd with `odo push` after creating a component.

kadel:
> @kadel Just writing some simple tests now.
> 
> Would you think this is the right way to test it?
> 
> 1. Set the PodTimeout to something _very_ small (2 seconds)
> 2. Use runFailCmd with `odo push` after creating a component.

you should make sure that the pod never runs by breaking it in some way, for example, nonexisting image.

Tha you need to make sure that ti really waited for two seconds before fail

surajnarwade:
* setting wrong value shows help along with error

```
$ odo utils config set PodTimeout sdfdfd
Error: unable to set podtimeout to sdfdfd: strconv.Atoi: parsing "sdfdfd": invalid syntax
Usage:
  odo utils config set [flags]

Examples:

   # Set a configuration value
   odo utils config set NamePrefix "app"
   odo utils config set Timeout 20
   odo utils config set PodTimeout 20
	

Flags:
  -h, --help   help for set

Global Flags:

 ✗  strconv.Atoi: parsing "sdfdfd": invalid syntax
```

* Code wise LGTM :tada:  , but we should also consider that our user might not aware of terminology `pod` and `podTimeout`

mik-dass:
> Code wise LGTM tada , but we should also consider that our user might not aware of terminology pod and podTimeout

I do agree that the new user with no knowledge of open shift and others, might not be aware of it

cdrage:
I'm going to WIP for this for now until I find time to re-visit this PR.

cdrage:
Going to free up the list of PR's by closing this for now and re-visiting it again at a later time.

PR #1029, opened 2018-11-22
add step to compress before publishing artifacts to appveyor

syamgk:
add step to compress before publishing artifacts to appveyor

to reduce the storage consumption on appveyor

https://github.com/redhat-developer/odo/issues/1025

## How to test changes?
check if artifacts are publishing on appveyor in a compressed form and validate the binaries

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,12 +32,18 @@ test_script:
 build_script:
 - make cross
 
+after_test:
+    - ps : 7z a -tgzip odo-linux-amd64.gz dist\bin\linux-amd64\odo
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -32,12 +32,19 @@ test_script:
 build_script:
 - make cross
 
+after_build:
```
kadel: Wouldn't it be better to run this after tests (`after_tests`)?
Currently, it builds binaries, than gzips it, and then run tests. But gzipping doesn't make sense if tests don't pass.

syamgk: yep, ack!

PR #983, opened 2018-11-14
Add loading progress / revamped UX for logging

cdrage:
Adds a loading progress when deploying containers (in particular,
pushing..)

Revamps the logging so it's more "modern" and cleaner.

PR #1034, opened 2018-11-22
Update frontpage with demo and updated instructions

cdrage:
Updates the website with the Asciinema Demo as well as updated
instructions as we did not previously add `odo push`


cdrage:
Preview can be created by pulling this PR and doing:

```sh
bundle exec jekyll serve .
```

With the correct Ruby Gems installed (jekyll + bundler)

cdrage:
That's okay, I'll update it again once #983 is merged since it'll take a while!

Issue #1037, opened 2018-11-23
`make install` is failing


surajnarwade:
```
$ make install
go install -ldflags="-w -X github.com/redhat-developer/odo/cmd.GITCOMMIT=e6dd7ebc"
can't load package: package github.com/redhat-developer/odo: no Go files in /home/snarwade/go/src/github.com/redhat-developer/odo
make: *** [Makefile:15: install] Error 1
```

kadel:
fix is in https://github.com/redhat-developer/odo/pull/1038

Issue #908, opened 2018-10-31
add cpu flags to component create


jorgemoralespou:
[kind/Feature]

## Which functionality do you think we should add?

We need to add a flag to `odo create` to be able to establish the cpu the runtime container will have. Some applications are cpu hungry and in some environments the defaults that are assigned when nothing is specified do not work. 

I would think we should add, at a minimum a flag for `cpu` that should set request and limits for the container with the runtime in the pod.

Additionally it would be nice that we could set `min-cpu` and `max-cpu` as flags to set requests on the former and limits on the latter.

`min-cpu` and  `max-cpu` should have priority over `cpus` if all of them are set, although it would be desirable to emit a warning advising the user that in this case `cpu` flag is superfluous and could be omitted.

Example
```bash
odo create wildfly my-backend --cpu=2

odo create nodejs my-frontend --min-cpu=0.25 --max-cpu=2 
```

PR #1038, opened 2018-11-23
Add cpu flags to component create

geoand:
In the spirit of what was done for #907, we introduce flags to control the cpu requests and limits.
This work was made super easy by the great implementation that was done by @anmolbabu for #907 - you'll notice that the code changes are very minimal  

Fixes: #908
Fixes: #1037

## How to test changes?
Test the creation of the DC using of `--cpu` or `--min-cpu` / `--max-cpu`.
Also try testing with `--memory` to make sure that both resources are applied properly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ bin:
 
 .PHONY: install
 install:
-	go install ${BUILD_FLAGS}
+	go install ${BUILD_FLAGS} ./cmd/odo/
 
 # run all validation tests
 .PHONY: validate
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -30,6 +30,9 @@ var (
 	memoryMax        string
 	memoryMin        string
 	memory           string
+	cpuMax           string
+	cpuMin           string
+	cpu              string
 )
 
 var componentCreateCmd = &cobra.Command{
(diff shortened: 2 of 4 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -294,13 +296,46 @@ A full list of component types that can be deployed is available using: 'odo cat
 	},
 }
 
+func ensureAndLogProperMemoryUsage() {
```
anmolbabu: :+1: For this function

pkg/util/util.go
```diff
@@ -279,7 +279,7 @@ func FetchResourceQuantity(resourceType corev1.ResourceName, min string, max str
 		maxResource = resource.MustParse(request)
 	}
 	return &ResourceRequirementInfo{
-		ResourceType: corev1.ResourceMemory,
+		ResourceType: resourceType,
```
anmolbabu: :+1: For this one also :)

pkg/odo/cli/component/create.go
geoand: This is being updated :P, meaning I will make one generic function to work in both cases since `code climate` is complaining :)

geoand:
@anmolbabu This should now be ready for review :)

Makefile
```diff
@@ -12,7 +12,7 @@ bin:
 
 .PHONY: install
 install:
-	go install ${BUILD_FLAGS}
+	go install ${BUILD_FLAGS} ./cmd/odo/
```
metacosm: 👍

Issue #1047, opened 2018-11-27
Execution of cli-doc results in a panic when no args are supplied


geoand:
[kind/bug]

## What versions of software are you using?
- Operating System: Linux
- Output of `odo version`: odo v0.0.16 (HEAD)

## How did you run odo exactly?

go run cmd/cli-doc/cli-doc.go

## Actual behavior

```
panic: runtime error: index out of range

goroutine 1 [running]:
main.main.func1(0xc000727200, 0x2ac76c8, 0x0, 0x0)
        /home/gandrian/gowork/src/github.com/redhat-developer/odo/cmd/cli-doc/cli-doc.go:146 +0x211
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).execute(0xc000727200, 0xc0000b01b0, 0x0, 0x0, 0xc000727200, 0xc0000b01b0)
        /home/gandrian/gowork/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:702 +0x2d3
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc000727200, 0x159ce40, 0x2a97601, 0xc000514f30)
        /home/gandrian/gowork/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:783 +0x2dc
github.com/redhat-developer/odo/vendor/github.com/spf13/cobra.(*Command).Execute(0xc000727200, 0xc000514f30, 0xc0000a6058)
        /home/gandrian/gowork/src/github.com/redhat-developer/odo/vendor/github.com/spf13/cobra/command.go:736 +0x2b
main.main()
        /home/gandrian/gowork/src/github.com/redhat-developer/odo/cmd/cli-doc/cli-doc.go:157 +0x12a

```

## Expected behavior

I expected to see a help message on how to use the `cli-doc`

PR #1048, opened 2018-11-27
Fix panic when no args supplied to cli-doc

geoand:
Fixes the bug reported in #1047

fixes #???
#1047

## How to test changes?
`go run cmd/cli-doc/cli-doc.go` should now work instead of resulting in a panic.
Also: `go run cmd/cli-doc/cli-doc.go structure` and `go run cmd/cli-doc/cli-doc.go reference` should continue to work as previously


geoand:
@metacosm LOL :P

PR #1046, opened 2018-11-27
Enhance describe command

surajnarwade:
## What is the purpose of this change? What does it change?
This PR enhances describe command for both application and component.
it also has some ground work for machine readable output work.

## Was the change discussed in an issue?
fixes #853

## How to test changes?

* `odo describe`
* `odo app describe`

PR #1053, opened 2018-11-29
Fix broken master

geoand:
This happened because PR #1046 fell behind master and was not rebased

No

## How to test changes?
`make bin` should now work


cdrage:
LGTM

PR #1059, opened 2018-12-03
Add service catalog test that tests a spring boot application with a postgresql db

geoand:
Add test that tests a spring boot application with a postgresql db. 
This has been requested by various people that would like to be able to test a spring boot application that uses a database, without actually having to know the details of spring boot development

No, but it has been mentioned in team meetings

## How to test changes?
`RUN_ALL_SERVICE_TESTS=true make test-service-e2e` should run the new tests along with the existing service catalog tests

`make test-service-e2` (which is what Travis uses) will should not execute the new tests


geoand:
The reason I would prefer this test not run on Travis is that building a Spring Boot application via s2i on a new cluster would take a long time due to the numerous dependencies that a Spring Boot application contains

tests/e2e/examples/source/openjdk-sb-postgresql/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+    <artifactId>demo</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <name>Spring Boot - CRUD Demo</name>
+    <description>Spring Boot - CRUD Demo</description>
+
+    <properties>
+        <awaitility.version>3.1.0</awaitility.version>
+        <postgresql.version>9.4.1212</postgresql.version>
+        <rest-assured.version>3.1.0</rest-assured.version>
+        <snowdrop-bom.version>1.5.16.SP1</snowdrop-bom.version>
```
metacosm: Why not use `1.5.17`? ;)

geoand: Good catch! I just generated a project from our generator which I guess hasn't been updated

geoand:
I dont mind having a separate file @metacosm, I'm on the fence about what the correct way to go should be

tests/e2e/examples/source/openjdk-sb-postgresql/pom.xml
geoand: Fixed

Issue #1064, opened 2018-12-03
autocompletion for `odo link`


kadel:
`odo link` should have autocompletion for both services and components

When I do `odo link <tab>` it should suggest me list of components and services that I have deployed. And that I can link component to.

geoand:
@kadel I will pick this up and start working on it

PR #1073, opened 2018-12-04
Add completion for odo link

geoand:
This PR adds support for providing completion suggestions for `odo link`.
It also refactors `components.List` to make a single API server call and adds a unit test for it (I also spotted plenty more improvement of such nature that can be made to `component` but didn't attempt them in this PR since I didn't want to change irrelevant code)

Fixes: #1064

## How to test changes?
Assuming that the completion scripts have already been setup and that the binary from this PR is on $PATH, a good test would be the following:

```
odo create nodejs frontend
odo create python backend
odo service create mysql-ephemeral
```

Now `odo link` + [TAB] should yield `fronend` and `mysql-ephemeral` while
`odo link --component frontend` + [TAB] should yield `backend` and `mysql-ephemeral` while

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -491,19 +491,23 @@ func GetComponentType(client *occlient.Client, componentName string, application
 func List(client *occlient.Client, applicationName string) ([]Info, error) {
 
 	applicationSelector := fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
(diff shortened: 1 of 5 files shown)


geoand:
@metacosm That's a great idea! If you open a ticket I can start working on that

metacosm:
> @metacosm That's a great idea! If you open a ticket I can start working on that

Here you go: #1077

geoand:
Merging, thanks!

PR #1074, opened 2018-12-04
fix bump-version.sh (version.go file was moved)

kadel:

fixes script that updates version number before release

this got broken due to moving verision.go file.


cdrage:
LGTM.

PR #1081, opened 2018-12-04
fix path to GITCOMMIT variable to display correct value in 'odo version'

kadel:
`odo version` was always reporting just HEAD because this was not properly updated after the file moved


cdrage:
LGTM.

PR #1082, opened 2018-12-04
Add --show-log parameter

cdrage:
Adds a --show-log paramater to create, update and push.

This will show the build progress.

If --show-log is passed in, it will also disable the "spinner" as it may
conflict / get caught within the build output.

Closes https://github.com/redhat-developer/odo/issues/1058

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -10,7 +10,6 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/fatih/color"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	applabels "github.com/redhat-developer/odo/pkg/application/labels"
@@ -145,7 +144,7 @@ func CreateFromGit(client *occlient.Client, params occlient.CreateArgs) error {
 	labels := componentlabels.GetLabels(params.Name, params.ApplicationName, true)
 
 	// Loading spinner
-	s := log.Spinnerf("Creating component %s", params.Name)
+	s := log.Spinnerf(true, "Creating component %s", params.Name)
 	defer s.End(false)
 
 	// Parse componentImageType before adding to labels
@@ -229,7 +228,7 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
 	labels := componentlabels.GetLabels(params.Name, params.ApplicationName, true)
 
 	// Loading spinner
-	s := log.Spinnerf("Creating component %s", params.Name)
+	s := log.Spinnerf(true, "Creating component %s", params.Name)
 	defer s.End(false)
 
 	// Parse componentImageType before adding to labels
@@ -290,7 +289,7 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
 func Delete(client *occlient.Client, componentName string, applicationName string) error {
 
 	// Loading spinner
-	s := log.Spinnerf("Deleting component %s", componentName)
+	s := log.Spinnerf(true, "Deleting component %s", componentName)
 	defer s.End(false)
 
 	cfg, err := config.New()
@@ -433,7 +432,8 @@ func getS2IPaths(podEnvs []corev1.EnvVar) []string {
 // 	isForcePush indicates if the sources to be updated are due to a push in which case its a full source directory push or only push of identified sources
 // Returns
 //	Error if any
-func PushLocal(client *occlient.Client, componentName string, applicationName string, path string, out io.Writer, files []string, delFiles []string, isForcePush bool) error {
+// 'show' will determine whether or not the log will be shown to the user
+func PushLocal(client *occlient.Client, componentName string, applicationName string, path string, out io.Writer, files []string, delFiles []string, isForcePush bool, show bool) error {
 	glog.V(4).Infof("PushLocal: componentName: %s, applicationName: %s, path: %s, files: %s, delFiles: %s, isForcePush: %+v", componentName, applicationName, path, files, delFiles, isForcePush)
 	// Find DeploymentConfig for component
 	componentLabels := componentlabels.GetLabels(componentName, applicationName, false)
@@ -478,8 +478,7 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 		}
 	}
 
-	// Copy the files to the pod
-	s := log.Spinner("Copying files to component")
+	s := log.Spinner(true, "Copying files to component")
 
 	if !isForcePush {
 		if len(files) == 0 && len(delFiles) == 0 {
@@ -497,23 +496,20 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 	}
 	s.End(true)
 
-	s = log.Spinner("Building component")
+	s = log.Spinner(true, "Building component")
 
-	// use pipes to write output from ExecCMDInContainer in yellow  to 'out' io.Writer
 	pipeReader, pipeWriter := io.Pipe()
 	var cmdOutput string
 	go func() {
-		yellowFprintln := color.New(color.FgYellow).FprintlnFunc()
 		scanner := bufio.NewScanner(pipeReader)
 		for scanner.Scan() {
 			line := scanner.Text()
-			// color.Output is temporarily used as there is a error when passing in color.Output from cmd/create.go and casting to io.writer in windows
-			// TODO: Fix this in the future, more upstream in the code at cmd/create.go rather than within this function.
-			// If we are in debug mode, we should show the output
-			if log.IsDebug() {
-				yellowFprintln(color.Output, line)
+			if log.IsDebug() || show {
+				_, err := fmt.Fprintln(out, line)
+				if err != nil {
+					log.Errorf("Unable to print to stdout: %v", err)
+				}
 			}
-
 			cmdOutput += fmt.Sprintln(line)
 		}
 	}()
@@ -538,10 +534,11 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 // Build component from BuildConfig.
 // If 'wait' is true than it waits for build to successfully complete.
 // If 'wait' is false than this function won't return error even if build failed.
-func Build(client *occlient.Client, componentName string, applicationName string, wait bool, stdout io.Writer) error {
+// 'show' will determine whether or not the log will be shown to the user
+func Build(client *occlient.Client, componentName string, applicationName string, wait bool, show bool, stdout io.Writer) error {
 
-	// Loading spinner
-	s := log.Spinnerf("Triggering build from git")
+	// No loading spinner if we're showing the logging output
+	s := log.Spinnerf(!show, "Triggering build from git")
 	defer s.End(false)
 
 	// Namespace the component
@@ -556,16 +553,16 @@ func Build(client *occlient.Client, componentName string, applicationName string
 	}
 	s.End(true)
 
+	// Now we check to see if we're in debug or "show" mode
 	// Retrieve the Build Log and write to buffer if debug is disabled, else we we output to stdout / debug.
-
 	var b bytes.Buffer
-	if !log.IsDebug() {
+	if !log.IsDebug() && !show {
 		stdout = bufio.NewWriter(&b)
 	}
 
 	if wait {
 
-		s := log.Spinnerf("Waiting for build to finish")
+		s := log.Spinnerf(true, "Waiting for build to finish")
 		defer s.End(false)
 		if err := client.FollowBuildLog(buildName, stdout); err != nil {
 			return errors.Wrapf(err, "unable to follow logs for %s", buildName)
@@ -666,7 +663,8 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
 // newSourceType indicates the type of the new source i.e git/local/binary
 // newSource indicates path of the source directory or binary or the git URL
 // stdout is the io writer for streaming build logs on stdout
-func Update(client *occlient.Client, componentName string, applicationName string, newSourceType string, newSource string, newSourceRef string, stdout io.Writer) error {
+// 'show' will determine whether or not the log will be shown to the user
+func Update(client *occlient.Client, componentName string, applicationName string, newSourceType string, newSource string, newSourceRef string, show bool, stdout io.Writer) error {
 
 	// STEP 1. Create the common Object Meta for updating.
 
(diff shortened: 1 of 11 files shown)


pkg/log/status.go
```diff
@@ -194,18 +194,18 @@ func Askf(format string, a ...interface{}) {
 // Status creates a spinner, sets the prefix then returns it.
 // Remember to use .End(bool) to stop the spin / when you're done.
 // For example: defer s.End(false)
-func Spinner(status string) *Status {
+func Spinner(spin bool, status string) *Status {
```
mik-dass: Maybe add a comment here regarding the spin parameter

pkg/log/status.go
```diff
@@ -194,18 +194,18 @@ func Askf(format string, a ...interface{}) {
...
-	s.Start(status, IsDebug())
+	s.Start(status, IsDebug() || !spin)
 	return s
 }
 
 // Statusf creates a spinner, sets the prefix then returns it.
 // Remember to use .End(bool) to stop the spin / when you're done.
 // For example: defer s.End(false)
-func Spinnerf(format string, a ...interface{}) *Status {
+func Spinnerf(spin bool, format string, a ...interface{}) *Status {
```
mik-dass: Same here, maybe add a comment here regarding the spin parameter

mik-dass:
@cdrage The output for git looks like
```
[mrinaldas@localhost odo]$ ./odo create nodejs nodejs --git https://github.com/sclorg/nodejs-ex
 ✓   Checking component
 ✓   Checking component version
 ✓   Creating component nodejs
 •   Triggering build from git  ...
 ✓   Triggering build from git
 OK  Component 'nodejs' was created and port 8080/TCP was opened
 OK  Component 'nodejs' is now set as active component
```
The spinner is missing when `--show-logs` is missing (I think it should stay for --show-logs is false). Also the message `Triggering build` appears twice. 

I also think --show-log should also work with local and binary components for uniformity. It feels weird to say `--show-log` for local components and see nothing.

geoand:
I agree with @mik-dass, `--show-log` should definitely result in showing the build logs when using `local` or `binary`.

cdrage:
@mik-dass Double output is intended (since the first one is beginning... the second is the OK / checkmark output).

It's a bug that it's not showing, using `--git` locally on my machine did the output, but I'll have to double-check it. I must have did an update and broke it.

However...

--show-log will not work for local or binary when using `odo create` since supervisord / Odo will not actually build it until we use `odo push`. However, it *will* show when using `--git` since Odo / supervisord actually builds it when using `odo create`. 

`--show-log` will work throughout all components however (local, git, binary) when using `odo update` or `odo push`

Does that make sense?

@kadel

geoand:
@cdrage Yeah, I missed that, I was thinking that it wasn't working for `push`

cdrage:
Hey @mik-dass @geoand 

I fixed the bug regarding the spinner now showing when "not" using --show-log.

Can you please test it again?

mik-dass:
> --show-log will not work for local or binary when using odo create since supervisord / Odo will not actually build it until we use odo push. However, it will show when using --git since Odo / supervisord actually builds it when using odo create.

I know but the thing is I find it weird to say `--show-log` and see nothing. I know it's a bad idea but maybe show the supervisorD logs here? Or maybe inform the user via a print statement that `--show-log` only works for git components?

cdrage:
With regards to @mik-dass 's comments, what do you think @kadel ?

It does seem odd to have --show-log for create when it only affects git.. Should we just include it in `odo push` instead and have `--git` in `odo create` default to logging output?

kadel:
> It does seem odd to have --show-log for create when it only affects git.. Should we just include it in `odo push` instead and have `--git` in `odo create` default to logging output?

Wasn't there a ticket for making `odo create --git` behave in the same way as --local and --binary? That means triggering build only with `odo push` and not after the component is created? I think that that should be the correct behavior.   @jorgemoralespou 

This would simplify this, because  than we wouldn't need the --show-logs for create

cdrage:
@mik-dass @anmolbabu @surajnarwade @girishramnani Review time!

amitkrout:
@cdrage Would you please update the pr description template https://github.com/redhat-developer/odo/pull/1082#issue-235940887 as per your final commit.

anmolbabu:
@cdrage Please excuse me if I am missing something but I see logs when I use verbosity 4 for example...
Why an additional flag?

cdrage:
See https://github.com/redhat-developer/odo/issues/1058 on why we need the --show-log flag @anmolbabu 

I've edited the PR, you should see --show-log now when doing push! It seems I implemented it correctly for git deployments but not for local :-1: 

Feel free to try it again!

cdrage:
Ready for review @metacosm @anmolbabu @mik-dass @surajnarwade

mik-dass:
@cdrage please update the comments with the new parameter for https://github.com/redhat-developer/odo/pull/1082#discussion_r238953979 and https://github.com/redhat-developer/odo/pull/1082#discussion_r238954026

mik-dass:
> Thanks @mik-dass ! I've updated the code comments to reflect adding the new parameter.

LGTM :+1:

pkg/component/component.go
```diff
@@ -373,33 +372,32 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
...
+		spin.End(false)
 		return errors.Wrap(err, "unable push files to pod")
 	}
-	s.End(true)
+	spin.End(true)
 
-	s = log.Spinner("Building component")
+	// Create a spinner for the component being built
+	var s *log.Status
+	if show {
```
kadel: why not use `show` in `log.Spinnerf` directly?
`s := log.Spinnerf(!show, "Building component")`

pkg/component/component.go
```diff
@@ -373,33 +372,32 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
...
 		scanner := bufio.NewScanner(pipeReader)
 		for scanner.Scan() {
 			line := scanner.Text()
-			// color.Output is temporarily used as there is a error when passing in color.Output from cmd/create.go and casting to io.writer in windows
-			// TODO: Fix this in the future, more upstream in the code at cmd/create.go rather than within this function.
-			// If we are in debug mode, we should show the output
-			if log.IsDebug() {
-				yellowFprintln(color.Output, line)
+			if log.IsDebug() || show {
+				fmt.Println(line)
```
kadel: shouldn't this write to `out`, instead of directly to stdout?
Why do we have `out` in this function?

pkg/component/component.go
```diff
@@ -423,10 +421,18 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
...
-func Build(client *occlient.Client, componentName string, applicationName string, wait bool, stdout io.Writer) error {
+// 'show' will determine whether or not the log will be shown to the user
+func Build(client *occlient.Client, componentName string, applicationName string, wait bool, show bool, stdout io.Writer) error {
 
+	// No loading spinner if we're showing the logging output
 	// Loading spinner
-	s := log.Spinnerf("Triggering build from git")
+
+	var s *log.Status
+	if show {
```
kadel: why not use `show` in `log.Spinnerf` directly?
`s := log.Spinnerf(!show, "Building component")`

pkg/component/component.go
```diff
@@ -547,7 +553,7 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
 // newSourceType indicates the type of the new source i.e git/local/binary
 // newSource indicates path of the source directory or binary or the git URL
 // stdout is the io writer for streaming build logs on stdout
-func Update(client *occlient.Client, componentName string, applicationName string, newSourceType string, newSource string, newSourceRef string, stdout io.Writer) error {
+func Update(client *occlient.Client, componentName string, applicationName string, newSourceType string, newSource string, newSourceRef string, show bool, stdout io.Writer) error {
```
kadel: please update comment with explanations what `show` is

pkg/odo/cli/component/update.go
```diff
@@ -113,6 +113,7 @@ var updateCmd = &cobra.Command{
 
 // NewCmdUpdate implements the Update odo command
 func NewCmdUpdate() *cobra.Command {
+	updateCmd.Flags().BoolVar(&show, "show-log", false, "If enabled, logs will be shown when built")
```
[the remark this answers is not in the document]
kadel: no we don't need to show example for every parameters

pkg/component/component.go
cdrage: Bit confused, but how it works is that: If show is passed in or debug, it'll print this directly.

We have `cmdOut` in the function purely for when we *dont* use --show or debug. Since if build fails, it'll output the entire build output to the user.

Does that make sense?

cdrage:
I've updated the PR @kadel 

I'm a bit confused with regards to the "out", do you mean the `cmdOutput` parameter? Or you mean that we should pass in `stdout` (since we pass that into the function anyways..)

pkg/odo/cli/component/update.go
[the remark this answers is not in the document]
anmolbabu: Ok the reason I said that is with the failure I faced before while testing this PR, I thought I was passing the flag incorrectly but anyway we found it was a 🕷 in the PR so I 🤔 no need to do that now

pkg/component/component.go
kadel: this function has `our io.Writer` argument. If I'm not mistaken this argument is there  represent `os.Stdout`, you should use that.

kadel: what I mean is to use `fmt.Fprintf(out,`

pkg/odo/cli/component/update.go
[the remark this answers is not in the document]
kadel: Another question is. Does it make sense to have `--show-log` for update, if we don't have it for create?

pkg/component/component.go
cdrage: @kadel Been updated! :+1:

cdrage:
Updated the PR @kadel with the changes. I ended up using Fprintln

surajnarwade:
@cdrage , commit message says `--show-log` should be for create, update and push, but I can see this only for push and update.

kadel:
@cdrage please rebase and update commit message.

amitkrout:
@cdrage Can you please bring some attention over the decreasing coverage in a specific file ```pkg/component/component.go``` https://github.com/redhat-developer/odo/pull/1082#issuecomment-444243099 though we have overall increase in coverage.

kadel:
Now I just realized why are we adding `--show-log` to `update`? It should be there just for `push`. And than in next PR we can add it for `watch` but it doesn't make sense to have it there for `create` and `update` nothing should be build during those operations

cdrage:
@kadel 

It makes sense to add it to `update` for components which are pushed using `--git`. It does make sense for `push`, but for `create` and `update`, it should be there for the components deployed with `git`.

kadel:
> It makes sense to add it to `update` for components which are pushed using `--git`. It does make sense for `push`, but for `create` and `update`, it should be there for the components deployed with `git`.

not really. `odo create` should never build a component (even for git). I know that it does that right now, but we have to change that. The same for `update`.

commands should behave the same regardless of what type of the component is.

cdrage:
Okay, I'll remove this from all other command line options except for `push`

kadel:
@cdrage please rebase :-|

cdrage:
I've rebased :+1: just waiting for tests to pass.

kadel:
Uff, log looks really ugly 😞 

` ◐  Building component` is there inside build log :-(

This should be displayed only at the beggining of the log.



```
[INFO] --- maven-resources-plugin:3.0.2:resources (default-resources) @ spring-boot-rest-http ---
 ◒  Building component[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 2 resources
[INFO]
[INFO] --- fabric8-maven-plugin:3.5.40:resource (fmp) @ spring-boot-rest-http ---
 ◐  Building component[INFO]
[INFO] --- maven-compiler-plugin:3.7.0:compile (default-compile) @ spring-boot-rest-http ---
 ◓  Building component[INFO] Changes detected - recompiling the module!
[INFO] Compiling 3 source files to /tmp/src/target/classes
 ◒  Building component[INFO]
[INFO] --- maven-resources-plugin:3.0.2:testResources (default-testResources) @ spring-boot-rest-http ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 2 resources
[INFO]
[INFO] --- maven-compiler-plugin:3.7.0:testCompile (default-testCompile) @ spring-boot-rest-http ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 3 source files to /tmp/src/target/test-classes
 ◓  Building component[INFO]
[INFO] --- maven-surefire-plugin:2.20:test (default-test) @ spring-boot-rest-http ---
[INFO] Tests are skipped.
[INFO]
[INFO] --- maven-jar-plugin:3.0.2:jar (default-jar) @ spring-boot-rest-http ---
 ◑  Building component[INFO] Building jar: /tmp/src/target/spring-boot-rest-http-1.5.17-5-SNAPSHOT.jar
[INFO]
[INFO] --- spring-boot-maven-plugin:1.5.17.RELEASE:repackage (default) @ spring-boot-rest-http ---
 ◐  Building component[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 4.418 s
[INFO] Finished at: 2019-01-31T14:53:15Z
 ◓  Building component[INFO] Final Memory: 60M/623M
[INFO] ------------------------------------------------------------------------
Copying Maven artifacts from /tmp/src/target to /deployments ...
Running: cp *.jar /deployments
 ◑  Building component... done
+ '[' '!' -z /home/jboss ']'
+ '[' -n /opt/app-root/src-backup ']'
+ '[' /tmp '!=' /home/jboss ']'
+ '[' /deployments '!=' /home/jboss ']'
+ rsync -rlO /opt/app-root/src-backup/src/. /tmp/src/
 ◒  Building component+ '[' -n /deployments ']'
+ '[' '!' -d /opt/app-root/deployment-backup ']'
+ rsync -rlO /deployments/. /opt/app-root/deployment-backup/
 ◒  Building component+ /var/lib/supervisord/bin/supervisord ctl stop run
 ◐  Building componentrun: stopped
+ /var/lib/supervisord/bin/supervisord ctl start run
 ◓  Building componentrun: started
 ✓  Building component
 ✓  Changes successfully pushed to component: spring-boot-redhat-openjdk18-openshift-nvmd
```

cdrage:
@kadel What the heck hahah.

Looks like the most recent rebase has messed everything up, and now I have even *more* rebasing / conflicts now that we've merged @anmolbabu 's PR.

I'm going to change this to WIP again and work on it again on Monday (since I'm traveling Thursday evening / Friday morning) and finally get this done. :100

PR #1101, opened 2018-12-06
[WIP] Refractored the project code file to use structs

### LINKED ISSUE #1087: Creating structs and related functions for Odo entities
Odo entities should be represented as structs and they should have related functions like CREATE, DELETE, GET, LIST. This will help in machine readable output and also help in reducing the number of parameters passed to functions.

I would like to work on it and will create a simple sample PR for discussing design patterns

mik-dass:
fixes #1087 

TODO
* Modify the tests accordingly

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -162,3 +162,147 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
 	}
 	return false, errors.Errorf(" %v project does not exist", projectName)
 }
+
+type Project struct {
+	Name   string
+	Active bool
+	Client *occlient.Client
+}
+
+func (p *Project) Create() error {
+	err := p.Client.CreateNewProject(p.Name)
+	if err != nil {
+		return errors.Wrap(err, "unable to create new project")
+	}
+	return nil
+}
+
+func (p *Project) Exists() (bool, error) {
+	project, err := p.Client.GetProject(p.Name)
+	if err != nil {
+		return false, errors.Wrap(err, "unable to get the project")
+	}
+	if project == nil {
+		return false, nil
+	}
+	return true, nil
+}
+
+func (p *Project) SetActive() error {
+	currentProject := GetCurrent(p.Client)
+
+	cfg, err := config.New()
+	if err != nil {
+		return errors.Wrap(err, "unable to access config file")
+	}
+	err = cfg.UnsetActiveApplication(currentProject)
+	if err != nil {
+		return errors.Wrap(err, "unable to unset active application of current project "+p.Name)
+	}
+	err = cfg.UnsetActiveComponent(currentProject)
+	if err != nil {
+		return errors.Wrap(err, "unable to unset active component of current project "+p.Name)
+	}
+
+	err = p.Client.SetCurrentProject(p.Name)
+	if err != nil {
+		return errors.Wrap(err, "unable to set current project to"+p.Name)
+	}
+	return nil
+}
+
+func (p *Project) Delete() (string, error) {
+	// Loading spinner
+	s := log.Spinnerf("Deleting project %s", p.Name)
+	defer s.End(false)
+
+	currentProject := GetCurrent(p.Client)
+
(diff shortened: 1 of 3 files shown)


pkg/project/project.go
```diff
@@ -162,3 +162,148 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
...
+}
+
+func (p *Project) Exists() (bool, error) {
+	project, err := p.Client.GetProject(p.Name)
+	if err != nil {
+		return false, errors.Wrap(err, "unable to get the project")
+	}
+	if project == nil {
+		return false, nil
+	} else {
```
surajnarwade: we don't need `else` here

cdrage:
Hey @mik-dass what's the status on this? Is this currently being worked on, or blocked on something else?

mik-dass:
> Hey @mik-dass what's the status on this? Is this currently being worked on, or blocked on something else?

@cdrage I just rebased the PR. I am waiting for initial onions on the design/patterns of the structs and the related functions because I will have to rewrite all of the unit test code acoordingly. @kadel @girishramnani @geoand @metacosm Please have a look and provide your initial thoughts. Thanks :)

pkg/project/project.go
mik-dass: Fixed

pkg/project/project.go
```diff
@@ -162,3 +162,147 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
...
+		return "", errors.Wrap(err, "unable to delete project")
+	}
+
+	// delete from config
+	cfg, err := config.New()
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to delete project from config file")
+	}
+
+	err = cfg.DeleteProject(p.Name)
```
girishramnani: why delete twice?

girishramnani: If I am not wrong shouldn't we delete the project first from the cfg and then create a new config?

girishramnani:
@mik-dass I have been following @surajnarwade work on machine readable structs and correct me if I am wrong but would it reduce our work if the follow the `ObjectMeta`,`TypeMeta` and `Spec` attributes so in the future suraj doesn't need to re-write this?

girishramnani:
adding @amitkrout to get approvals on UTs and e2e tests

pkg/project/project.go
```diff
@@ -162,3 +162,147 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
 	}
 	return false, errors.Errorf(" %v project does not exist", projectName)
 }
+
+type Project struct {
+	Name   string
+	Active bool
+	Client *occlient.Client
```
anmolbabu: Since, client is not really a property of project it doesn't deserve to be in the project struct in my opinion

pkg/project/project.go
anmolbabu: Ok 2 things here:

1. Update of config file is a local file(created by odo at the moment) update
    a. No n/w access
    b. possible failure cause: disk failure
2. Deletion of project is a request to the cluster
    a. N/w failure cutting off access to the (possibly remote) cluster
    b. incorrect parameters to the request -- Unlikely, considering they need to be pre-validated by odo before proceeding to here...

Considering the above, if we update the config first and then the delete request to the cluster, I feel it would lead us to inconsistent behaviour... Whereas, if you make the delete request and then updation to the config file fails due to say a local disk failure, there is hardly any recovery mechanism until the disk failure is fixed so in essence, this inconsistency can hardly be met AFAIK but the other common case above is dealt with

So I tend to favor the approach of deleting the project first and then trying an update on config file

pkg/project/project.go
mik-dass: I agree that the client shouldn't be here. But I can't think of any other way to wrap the client and it is needed so that we don't pass the client in every function call. @girishramnani @anmolbabu @metacosm Please suggest a better design

mik-dass:
closing this for now, will revisit later

PR #1110, opened 2018-12-07
Modify structs for machine readable output

surajnarwade:
fixes #???

### DIFF
--- a/pkg/application/types.go
+++ b/pkg/application/types.go
@@ -1,0 +1,31 @@
+package application
+
+import (
+	"github.com/redhat-developer/odo/pkg/component"
+	"github.com/redhat-developer/odo/pkg/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Application
+type App struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              AppSeco `json:"spec,omitempty"`
+}
+
+type AppSeco struct {
+	Components []string
+}
+
+// AppSpec holds all information about application
+type AppSpec struct {
+	Name       string                    `json:"applicationName,omitempty"`
+	Components []component.ComponentSpec `json:"components,omitempty"`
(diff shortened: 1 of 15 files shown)


pkg/application/types.go
```diff
@@ -0,0 +1,26 @@
...
+// Application
+type App struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              AppSpec `json:"spec,omitempty"`
+}
+
+// AppSpec holds all information about application
+type AppSpec struct {
+	Name       string                    `json:"applicationName,omitempty"`
```
kadel: `metav1.ObjectMeta` already have `name` field what would be the difference between the `name` and `spec.name`?

pkg/application/types.go
```diff
@@ -0,0 +1,26 @@
...
+	Components []component.ComponentSpec `json:"components,omitempty"`
```
kadel: Wouldn't be enough to just to reference a component by some identification (like a name) rather than embedding whole component spec?

metacosm:
Would it make sense to use https://github.com/snowdrop/component-api for this?

pkg/odo/cli/application/application.go
```diff
@@ -195,17 +197,38 @@ var applicationListCmd = &cobra.Command{
...
-			for _, app := range apps {
-				activeMark := " "
-				if app.Active {
-					activeMark = "*"
+
+			if outputFlag == "json" {
+				var appo []application.App
+				for _, app := range apps {
+					a, _ := AppMachineOutput(client, app.Name)
+					appo = append(appo, a)
```
golangcibot: this result of append is never used, except maybe in other appends (from `megacheck`)

surajnarwade:
@metacosm , what's component-api, is it focussed on ODO or snowdrop specific ?

kadel:
Can you please update type definitions as per discussions in https://github.com/redhat-developer/odo/issues/1062 ?

PR #1113, opened 2018-12-07
Missing Deploying an Application using Odo link

amitkrout:
Fixing the missing link. ```Deploying an Application using Odo``` should redirect to the proper place

Actually it is a minor change, not really need any discussion i guess. The problem is with the link ```Deploying an Application using Odo``` in [README.md](https://github.com/redhat-developer/odo/blob/master/README.md) and is not redirected to the place where should actually go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 - [What is Odo?](#odo)
 - [Features](#features)
 - [Setup and Installation](#setup-and-installation)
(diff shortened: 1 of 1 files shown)


amitkrout:
One more thing i want to add here is that ```CLI Structure``` nowhere described in the README doc. Is it intentional ? If no then i will remove the link as it not solving any purpose

Issue #1099, opened 2018-12-05
odo watch pushes files to the wrong directory


kadel:
a good way to reproduce this is with  https://github.com/openshift/nodejs-ex

```
git clone  https://github.com/openshift/nodejs-ex
odo create nodejs mynode
odo push
odo url create
odo watch
```
now  change something in views/index.html

after watch triggers, refresh you won't see your changes in the webpage

The reason is that if you check `/opt/opt-root/src/` you will see that `index.html` is there instead of `/opt/opt-root/src/views`  where it should be. Basically, it ignores where the file is and always puts it into the document root folder.

`odo push` works correctly

kadel:
This might be nodejs specific. I wasn't able to reproduce this with java

geoand:
I just tried it with Java and was not able to reproduce it either

kadel:
Starting to think  that using `io.openshift.s2i.destination` was not a good idea :-( we might go back to pushing files directly to `/opt/app-root/src`

geoand:
We have more e2e tests now, so if we do that we should be able to see if anything breaks. 
Should perhaps add watch e2e tests as well?

geoand:
Maybe this is just a bug in the NodeJS s2i image that we should work around?

anmolbabu:
Yeah I tend to agree with @geoand with may be a bug on nodejs s2i image repo

kadel:
> Given that the current approach(relying on labels) is a perfectly legal approach as per s2i docs

not really, as we are not copying the whole application, but only a single files, that is something that s2i was never designed to do

I supsect that if you try python image, you will see the same problem. 
I also think that we haven't this problem before we started using this label.

Both are just my working theroies that I haven't verifed yet and I'm not sure if I get change to try it today :-(

anmolbabu:
@kadel I'll try it post lunch

kadel:
@anmolbabu  have you been able to determine why this affects python and nodejs but not java based images?

kadel:
@jorgemoralespou @jankleinert @sspeiche  Please be aware of this bug, when you are demoing nodejs apps. `odo push` works but `odo watch` can break the app if you edit file that is subdirectory and not in the root folder of the project :-(

kadel:
@geoand Is it possible that this was introduced with changes in https://github.com/redhat-developer/odo/pull/954 ?

geoand:
[...]

PR #1115, opened 2018-12-07
Limit single file transfer to top-level files

geoand:
This fixes an issue where watch was failing when attempting to push a single file change that was not a top-level file

Fixes: #1099 

## How to test changes?
The problematic scenario should now work

git clone  https://github.com/openshift/nodejs-ex
odo create nodejs mynode
odo push
odo url create
odo watch

now change something in views/index.html

The original jar pushing scenario should also work

odo push


geoand:
This is not a holistic solution to the problem as @anmolbabu has pointed out, but I believe it to be adequate for the time being (given the severity of the issue)

PR #1116, opened 2018-12-07
Limit single file transfer to top-level files

geoand:
This fixes an issue where watch was failing when attempting to push a single file change that was not a top-level file

Fixes: #1099 

## How to test changes?
The problematic scenario should now work

git clone  https://github.com/openshift/nodejs-ex
odo create nodejs mynode
odo push
odo url create
odo watch

now change something in views/index.html

The original jar pushing scenario should also work

odo push


pkg/occlient/occlient.go
```diff
@@ -2522,18 +2522,29 @@ func (c *Client) GetOnePodFromSelector(selector string) (*corev1.Pod, error) {
 func (c *Client) CopyFile(localPath string, targetPodName string, targetPath string, copyFiles []string) error {
 	isSingleFileTransfer := isSingleFileTransfer(copyFiles)
 
-	dest := path.Join(targetPath, filepath.Base(localPath))
+	var makeTarDestPath string
+	var containerDirectory string
+	var onlyFile string
+	if isSingleFileTransfer {
```
anmolbabu: Sorry but I tend to feel against this :(
Ideally in my opinion, recursivetar itself should be able to handle case of single files vs multiple files vs whole src directory :laughing: 
I am trying it out lets see where it goes ...
Parallely, I'll verify your PR once so we have either this or the other by today EOD

geoand: I feel like I've already spent too much time on this, so I'll let you handle the `recursivetar` refactoring if you so like :)
FWIW, if you do get it work, I'm all for it

anmolbabu: @geoand It seems like I ended up doing the fix mostly like yours(except special handling of single file transfers-- I believe so ;) need to check again to be honest)

Can you please have a look at https://github.com/redhat-developer/odo/pull/1114/files
I am ok with anything you say about the PR we can discuss and take a call about your PR and mine
Because honestly, the recursiveTar was correct and had to do nothing about single vs multiple files and the changes remained confined to the same functions that you are touching ;)

geoand: You PR definitely looks cleaner than mine :) 
Let me check and see if everything works for me as well

PR #1126, opened 2018-12-11
Improve ui code organization to make it more useful and reusable

metacosm:
Make the ui code reusable (at least, the parts that can be) so that commands can make use of some of these functions. Builds on top of #857.

No.

## How to test changes?
No visible changes.


pkg/odo/util/validation.go
```diff
@@ -22,3 +24,82 @@ func ValidateName(name string) error {
...
+	Type                 string
+	AdditionalValidators []string
+}
+
+// AsValidatable allows avoiding type casts in client code
+func (v Validatable) AsValidatable() Validatable {
+	return v
+}
+
+// Validator is a function that validates that the provided interface is conform to expectations or return an error
```
geoand: ` that the provided interface is conform to` -> ` that the provided interface conforms to`

pkg/odo/util/validation.go
```diff
@@ -22,3 +24,82 @@ func ValidateName(name string) error {
...
+type Validator func(interface{}) error
+
+// always validates
+var nilValidator = func(ans interface{}) error { return nil }
+
+var validators = make(map[string]Validator)
```
geoand: Can we a comment on what this does?

pkg/service/service.go
```diff
@@ -45,8 +46,20 @@ type ServicePlanParameter struct {
 	Description     string
 	HasDefaultValue bool
 	Default         string
-	Type            string
-	Required        bool
+	odoutil.Validatable
+}
+
+// NewServicePlanParameter creates a new ServicePlanParameter instance with the specified state
+func NewServicePlanParameter(name, typeName, defaultValue string, required bool) ServicePlanParameter {
```
geoand: Actually I don't think this should be here, since it's only used for tests - so why not add it to one of the test util files?

metacosm: This point is debatable. This function could be used outside of tests, even though it's not currently, especially since with these changes, you need to properly initiate the `Validatable` part of the `ServicePlanParameter`.

geoand: I doesn't feel right to me, but I see your point

pkg/odo/util/validation.go
metacosm: Removed it with new commit as this didn't work very well. Trying to explain what it was doing to add a comment made me realize that it was awkward… :)

pkg/odo/util/validation.go
metacosm: Fixed. Thanks.

pkg/odo/util/validation.go
geoand: Cool, I'll check again

pkg/service/service.go
```diff
@@ -45,8 +46,20 @@ type ServicePlanParameter struct {
 	Description     string
 	HasDefaultValue bool
 	Default         string
-	Type            string
-	Required        bool
+	validation.Validatable
```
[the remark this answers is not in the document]
metacosm: Good point. I don't think it would but I'm going to check.

glide.yaml
```diff
@@ -53,3 +53,4 @@ import:
   version: v3.11.0
   subpackages:
   - pkg/oc/cli/login
+
```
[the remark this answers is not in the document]
metacosm: Done.

glide.yaml
```diff
@@ -52,4 +52,4 @@ import:
 - package: github.com/openshift/origin
   version: v3.11.0
   subpackages:
-  - pkg/oc/cli/login
+  - pkg/oc/cli/login
```
surajnarwade: since you updated `glide.yaml`, did you run `make vendor-update` after that ?
ideally there should be `glide.lock` and updated `vendor` directory as well

metacosm: Hmm, that's a whitespace. No dependencies were changed.

pkg/odo/cli/service/create.go
```diff
@@ -119,6 +120,11 @@ func (o *ServiceCreateOptions) validateServiceName(i interface{}) (err error) {
 	return
 }
 
+var validateNameFn = func(i interface{}) error {
```
kadel: Why not use the normal function?

glide.yaml
metacosm: Also, since many dependencies are not at a fixed version, randomly running `make vendor-update` will almost certainly result in changes in dependencies even without a change in `glide.yaml`…

kadel: every dependency should be fixed to specific version. Only one that is currently not is github.com/hinshun/vt10x

surajnarwade: okay cool

pkg/odo/cli/service/create.go
surajnarwade: I am bit confused here, can you please add comments here ?

pkg/odo/util/validation/validatable.go
```diff
@@ -0,0 +1,13 @@
+package validation
+
+// Validatable represents a common ancestor for validatable parameters
+type Validatable struct {
```
surajnarwade: if this is gonna used by service only, should we move this under `pkg/odo/service` ?

kadel: Right now it is used by service only, but that is because service create is the only command that has an interactive mode.
Once we add it into more commands it will be used in more places.

Am I right @metacosm ?

metacosm: > Right now it is used by service only, but that is because service create is the only command that has an interactive mode.
> Once we add it into more commands it will be used in more places.
> 
> Am I right @metacosm ?

Yes, that's the point.

pkg/odo/cli/service/create.go
metacosm: Indeed, let me check.

metacosm: Looks like rebasing errors… downside of having a PR opened for so long. :(

metacosm: Fixed.

pkg/odo/util/validation/validatable.go
```diff
@@ -0,0 +1,13 @@
...
+	Required             bool        `json:"required,omitempty"`
```
anmolbabu: Can you please add some comments around what each of these mean?
But awesome code I must say... remembered by Java days :)

pkg/odo/util/validation/validatable.go
```diff
@@ -0,0 +1,13 @@
...
+	Type                 string      `json:"type"`
+	AdditionalValidators []Validator `json:"-"`
+}
+
+// AsValidatable allows avoiding type casts in client code
```
anmolbabu: Can you explain how this can be helpful?
As in, at the moment I see the func returning the object its invoked upon

pkg/odo/util/validation/validation.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// ValidateName will do validation of application & component names according to DNS (RFC 1123) rules
+// Criteria for valid name in kubernetes: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md
+func ValidateName(name string) error {
```
anmolbabu: was thinking if this can be one of the validators
but I somehow have strong dislike(may be bcoz I have always got myself strangled in type assertions and conversions) for golang interface{} but where necessary they have to be used :)

anmolbabu: with go, I have always more enjoyed things being more typed than generic may be bcoz I felt that was the go principles but I might be wrong :)

anmolbabu: > was thinking if this can be one of the validators

I see its done below

anmolbabu:
Code LGTM apart from some code comments request
Will verify it tomorrow

pkg/odo/util/validation/validatable.go
metacosm: This just helps using Validatable in client code to avoid casting. See https://github.com/redhat-developer/odo/pull/1126/files#diff-6750ef04ff7fb19ec7f108ffcc9df23aR298: without this function, we would need a cast on the call site and that would be a lot less elegant :)

vasudevanrao:
Nice to follow this thread

anmolbabu:
I apologise for not being enough confident to push this huge PR especially when it touches areas in service catalog so I believe my ack would gain more weight may be with atleast one more ack

PR #1133, opened 2018-12-12
Fix #1132: Add debug target.

### LINKED ISSUE #1132: Add a debug make target
[kind/Enhancement]
Add a make debug target that removes the `-w` flag so that debuggers can be attached to a running `odo`.

## Why is this needed? 
Not needed but useful while developing on `odo`.

metacosm:
Adds a `debug` target to Makefile.

fixes #1132 

## How to test changes?
`make debug` then try to attach a debugger to the running `odo` instance. Without this fix, you'd get an error similar to: `could not attach to pid 90236: decoding dwarf section info at offset 0x0: too short`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
 PROJECT := github.com/redhat-developer/odo
 GITCOMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
 PKGS := $(shell go list  ./... | grep -v $(PROJECT)/vendor)
-BUILD_FLAGS := -ldflags="-w -X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)"
-DEBUG_BUILD_FLAGS := -ldflags="-X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)"
+COMMON_FLAGS := -X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)
+BUILD_FLAGS := -ldflags="-w $(COMMON_FLAGS)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,10 +2,15 @@ PROJECT := github.com/redhat-developer/odo
 GITCOMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
 PKGS := $(shell go list  ./... | grep -v $(PROJECT)/vendor)
 BUILD_FLAGS := -ldflags="-w -X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)"
+DEBUG_BUILD_FLAGS := -ldflags="-X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)"
```
kadel: It would be better to separate common flags to the variable and than use it to define build flags and debug flags, so we don't have everything defined twice.

something like this
```
COMMON_LDFLAGS := -X $(PROJECT)/pkg/odo/cli/version.GITCOMMIT=$(GITCOMMIT)
BUILD_FLAGS := -ldflags=" -w $(LDFLAGS)"
DEBUG_FLAGS := -ldflags="$(LDFLAGS)"
```

metacosm: Done, thank you!

geoand: Good stuff :wink: 

Us java folks are learning these new (or should I say old) tricks!

metacosm:
Tests pass plus the change shouldn't even impact anything.

geoand:
I'll just restart Travis to get rid of the damn flake

cdrage:
LGTM.

surajnarwade:
Merging this based on 3 :heavy_check_mark:  and @kadel's concern is being addressed :smile:

PR #1134, opened 2018-12-13
Fix potential race condition on odo link/unlink

geoand:
Fixes a subtle race condition that can occur during `odo link` or `odo unlink`.
The race condition could (and sometimes does in our e2e tests) occur if we were adding a link while the
deployment was still being rolled out.
With this change, we now use Patch instead of Update ensuring that
no write conflicts occur.

No, but I have seen the problem occur a few times in the integration tests

## How to test changes? 
`odo link` and `odo unlink` should continue to work as before.
Furthermore, the `link` integration tests should now not exhibit flaky behavior (unless Travis fails to launch Openshift)

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/types"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/occlient/occlient.go
```diff
@@ -2100,19 +2097,19 @@ func (c *Client) updateDCOfComponent(componentName, applicationName string, dcUp
...
 		}
-	} else {
-		return errors.Wrapf(err, "dcUpdater was not properly set - this is definitely an implementation issue")
-	}
 
-	// update the DeploymentConfig with the secret
-	_, err = c.appsClient.DeploymentConfigs(c.Namespace).Update(dc)
-	if err != nil {
-		return errors.Wrapf(err, "DeploymentConfig not updated %s", dc.Name)
+		// update the DeploymentConfig with the secret
```
cdrage: update comment :)

pkg/occlient/occlient.go
```diff
@@ -2100,19 +2097,19 @@ func (c *Client) updateDCOfComponent(componentName, applicationName string, dcUp
...
+		_, err = c.appsClient.DeploymentConfigs(c.Namespace).Patch(dcName, types.JSONPatchType, []byte(patch))
+		if err != nil {
+			return errors.Wrapf(err, "DeploymentConfig not updated %s", dc.Name)
```
cdrage: `not patched` ?

pkg/occlient/occlient.go
```diff
@@ -2100,19 +2097,19 @@ func (c *Client) updateDCOfComponent(componentName, applicationName string, dcUp
...
+		}
+	} else {
+		return errors.Wrapf(err, "dcPatch was not properly set - this is definitely an implementation issue")
```
cdrage: error message seems kind of odd, remove `definitely an implemention issue` maybe?

pkg/occlient/occlient.go
geoand: LOL

geoand: Fixed

pkg/occlient/occlient.go
geoand: Fixed

pkg/occlient/occlient.go
geoand: Fixed

geoand:
Travis is stuck, so I will restart the job

amitkrout:
> Travis is stuck, so I will restart the job
@geoand How you are restating a stuck job in Travis? Are you doing force push ?

geoand:
> > Travis is stuck, so I will restart the job
> > @geoand How you are restating a stuck job in Travis? Are you doing force push ?

You can restart it from the Travis UI, but in this case I did force push since I wanted to rebase onto the latest master (and due to the force push Travis started again automatically)

PR #1139, opened 2018-12-14
added check for componentName before pushing

### LINKED ISSUE #1016: Error message for a missing component was clearer in 0.14 than in 0.15
[kind/bug]

- Operating System: Windows
- Output of `odo version`:

```
odo v0.0.15 (2f7ed497)

Server: https://172.18.93.44:8443
OpenShift: v3.10.45
Kubernetes: v1.10.0+b81c8f8
```

from command-line

## Actual behavior

odo push --local c:\ws\workingDirectoryForTest\fuse-rest-http-booster
Pushing changes to component:
unable to get component source

## Expected behavior

something more explicit like in odo 0.14:

odo push --local c:\ws\workingDirectoryForTest\fuse-rest-http-booster
No component is set as active.
Use 'odo component set <component name> to set and existing component as active or call this command with component name as and argument.

girishramnani:
added a check for componentName being set before pushing

fixes #1016 

## How to test changes?
remove the `~/.kube/odo` from your local computer then go to any local component source, try `odo push --local .`

### DIFF
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -51,6 +51,13 @@ var pushCmd = &cobra.Command{
 		}
 		componentName := context.Component(argComponent)
 
+		// if the componentName is blank then there is not active component set
+		if len(componentName) == 0 {
+			log.Error(`No component is set as active.
+Use 'odo component set' to set and existing component as active or call this command with component name as and argument.`)
+			os.Exit(1)
+		}
+
 		log.Namef("Pushing changes to component: %v", componentName)
 
 		sourceType, sourcePath, err := component.GetComponentSource(client, componentName, applicationName)
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -53,8 +53,9 @@ var pushCmd = &cobra.Command{
 
 		// if the componentName is blank then there is no active component set
 		if len(componentName) == 0 {
-			log.Error(`No component is set as active.
-Use 'odo component set' to set and existing component as active or call this command with component name as an argument.`)
+			log.Error("No component is set as active.\n" +
+				"Use 'odo component set' to set and existing component as active \n" +
+				"or call this command with component name as an argument.")
 			os.Exit(1)
 		}
 
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -53,9 +53,7 @@ var pushCmd = &cobra.Command{
 
 		// if the componentName is blank then there is no active component set
 		if len(componentName) == 0 {
-			log.Error("No component is set as active.\n" +
-				"Use 'odo component set' to set and existing component as active \n" +
-				"or call this command with component name as an argument.")
+			log.Error("No component is set as active. Use 'odo component set' to set an active component.")
 			os.Exit(1)
 		}
 
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -51,10 +51,10 @@ var pushCmd = &cobra.Command{
 		}
 		componentName := context.Component(argComponent)
 
-		// if the componentName is blank then there is not active component set
+		// if the componentName is blank then there is no active component set
 		if len(componentName) == 0 {
 			log.Error(`No component is set as active.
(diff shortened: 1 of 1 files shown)


pkg/odo/cli/component/push.go
```diff
@@ -51,6 +51,13 @@ var pushCmd = &cobra.Command{
 		}
 		componentName := context.Component(argComponent)
 
+		// if the componentName is blank then there is not active component set
+		if len(componentName) == 0 {
+			log.Error(`No component is set as active.
```
mik-dass: Why not use "" and `\n` for break-line instead?

pkg/odo/cli/component/push.go
```diff
@@ -51,6 +51,13 @@ var pushCmd = &cobra.Command{
 		}
 		componentName := context.Component(argComponent)
 
+		// if the componentName is blank then there is not active component set
```
mik-dass: `there is no active component set`

pkg/odo/cli/component/push.go
amitkrout: Make sense however TBH I feel this is too lengthy as a reported error message from odo end. 

Can it be improvised like
```
No active component is available, to activate a component please use `odo component set`.
```
I am not confident after the or - ```or call this command with component name as and argument.````. This means you are passing the argument and of course it is no more blank. So this is unnecessary IMO. Correct me f I am wrong ?

girishramnani: this message was mentioned in the issue as something which was present in odo's older version so I just went with it

pkg/odo/cli/component/push.go
girishramnani: ack

pkg/odo/cli/component/push.go
girishramnani: @mik-dass are you talking about the condition? `len(componentName) == 0`?

girishramnani:
I don't think we can write a unit test for this one @amitkrout .

pkg/odo/cli/component/push.go
amitkrout: BTW there is typo in the message ```and argument```. Replaced with ```an argument```

amitkrout:
@girishramnani Why you think so ? Might be i am missing some context here

pkg/odo/cli/component/push.go
mik-dass: @girishramnani No I meant the message itself. It seemed too lengthy and for new-line, please use `\n`

girishramnani:
I actually meant it would make more sense to write an e2e test for this one

pkg/odo/cli/component/push.go
girishramnani: @amitkrout resolved it

mik-dass: Cool LGTM from myside

pkg/odo/cli/component/push.go
```diff
@@ -51,6 +51,14 @@ var pushCmd = &cobra.Command{
 		}
 		componentName := context.Component(argComponent)
 
+		// if the componentName is blank then there is no active component set
+		if len(componentName) == 0 {
+			log.Error("No component is set as active.\n" +
+				"Use 'odo component set' to set and existing component as active \n" +
+				"or call this command with component name as an argument.")
```
cdrage: This is too long, I'd always try to avoid doing multi-line error messages. Just simply log error `No component is set as active. Use 'odo component set' to set an active component`

cdrage: Yeah, I'd avoid multi-line as that messes up any other output :+1: Highly suggest a one-liner :)

girishramnani: ack and resolved

cdrage:
LGTM :+1:

cdrage:
If you want to try to make an e2e test that'd be great :+1: Otherwise, this code LGTM.

amitkrout:
@girishramnani you merged the pr without addressing @cdrage's https://github.com/redhat-developer/odo/pull/1139#issuecomment-448212414 ;)

cdrage:
@girishramnani I don't think it's touch to add, just a quick check when no component is present. Could you open up an issue for this @girishramnani to keep track / have to add an e2e test?

PR #1143, opened 2018-12-17
Add newlines for help menu, modify odo service create help

cdrage:
`ktemplates` removes all extra newlines. Example output has been
modified to use `\n\n`.

Before:

```sh
Examples:
  # Create new postgresql service from service catalog using dev plan and name my-postgresql-db.
  odo service create dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret
  odo service list
```

After:

```sh
Examples:
  # Create new postgresql service from service catalog using dev plan and name my-postgresql-db.
  odo service create dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret

  odo service list
```

`odo service create --help` has been modified and truncated into a
simpler long description.

### DIFF
--- a/pkg/odo/cli/service/create.go
+++ b/pkg/odo/cli/service/create.go
@@ -21,12 +21,14 @@ var (
     # Create new postgresql service from service catalog using dev plan and name my-postgresql-db.
     %[1]s dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret`)
 
+	createShortDesc = `Create a new service from service catalog using the plan defined and deploy it on OpenShift.`
+
 	createLongDesc = ktemplates.LongDesc(`
 Create a new service from service catalog using the plan defined and deploy it on OpenShift.
-If service name is not provided, service type value will be used. The plan to be used must be passed along the service type
-using this convention <service_type>/<plan>. The parameters to configure the service are passed as a list of key=value pairs.
-The list of the parameters and their type is defined according to the plan selected.
-A full list of service types that can be deployed are available using: 'odo catalog list services'`)
+
+A --plan must be passed along with the service type. Parameters to configure the service are passed as key=value pairs.
+
+For a full list of service types, use: 'odo catalog list services'`)
 )
 
 // ServiceCreateOptions encapsulates the options for the odo service create command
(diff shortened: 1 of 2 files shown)


pkg/odo/cli/service/create.go
```diff
@@ -21,12 +21,14 @@ var (
     # Create new postgresql service from service catalog using dev plan and name my-postgresql-db.
     %[1]s dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret`)
 
+	createShortDesc = `Create a new service from service catalog using the plan defined and deploy it on OpenShift.`
```
amitkrout: +1

pkg/odo/cli/service/service.go
```diff
@@ -27,7 +26,7 @@ func NewCmdService(name, fullName string) *cobra.Command {
 		Use:   name,
 		Short: "Perform service catalog operations",
 		Long:  serviceLongDesc,
-		Example: fmt.Sprintf("%s\n%s\n%s",
+		Example: fmt.Sprintf("%s\n\n%s\n\n%s",
```
amitkrout: +1

metacosm:
Hmm, actually, this change is weird to me. The idea of using `ktemplates` is to have consistent formatting between commands. So if we need to add extra `\n` to get the formatting we want, maybe we shouldn't be using `ktemplates` at all.

geoand:
@metacosm Do we prefer we revert?

metacosm:
> @metacosm Do we prefer we revert?

It's not up to me but I'd revert, yes, as I think the fix is not to introduce more `\n` but rather fix on the root cause (i.e. we don't like the formatting provided by `ktemplates`). Maybe we need to create our own templates instead of `ktemplates`?

geoand:
@kadel ?

amitkrout:
> Hmm, actually, this change is weird to me. The idea of using ktemplates is to have consistent formatting between commands. So if we need to add extra \n to get the formatting we want, maybe we shouldn't be using ktemplates at all.

I didn't know this. Thanks for the technical side details

cdrage:
@metacosm We don't have to revert for now, but we can at least open up an issue with regards to `ktemplates`.  I've opened up issue https://github.com/redhat-developer/odo/issues/1148 for this

PR #1146, opened 2018-12-18
Use variable to run independent test scenario

### LINKED ISSUE #1138: Use variable through test terminal to run independent test feature
[kind/Enhancement]

Make a generic  ```PHONY``` statement which can handle any test feature individually just by passing through variable name like ```ODO_TEST_FEATURE```. For example pass the variable in the terminal like 
```
make integration ODO_TEST_FEATURE=odoJavaE2e
```
and in the Makefile only one ```.PHONY``` can handle whatever the argument we passed. 
```
# Run integration test
.PHONY: integration
integration: ODO_TEST_FEATURE = ""
integration:
ifdef TIMEOUT
	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus=$(ODO_TEST_FEATURE) -ginkgo.succinct -timeout $(TIMEOUT)
else
	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus=$(ODO_TEST_FEATURE) -ginkgo.succinct -timeout 15m
endif
```

## Why is this needed? 
To remove unnecessary ```.PHONY``` statement for individual test in ```Makefile``` and make it more dynamic for developer to test their feature file just by passing it through a variable.

amitkrout:

Make the test suite more user friendly/independent and reduce the ```.PHONY``` load on Makefile
fixes #1138 

## How to test changes?

```
make test-e2e ODO_TEST_SCENARIO="test-feature"
```
In ```test-feature``` you can use the available spec. For example 
```
make test-e2e ODO_TEST_SCENARIO="odo-main-e2e"
make test-e2e ODO_TEST_SCENARIO="odo-java-e2e"
make test-e2e ODO_TEST_SCENARIO="odo-source-e2e"
[...]
```
**NOTE:** No use of variable ```ODO_TEST_SCENARIO``` will run the entire test suite. For example 
```
make test-e2e
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-main-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odo-main-e2e"
 
     # Run component e2e tests
     - <<: *base-test
@@ -51,7 +51,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - travis_wait make test-cmp-e2e
+        - travis_wait make test-e2e ODO_TEST_SCENARIO="odo-cmp-e2e"
 
     # Run java e2e tests
     - <<: *base-test
@@ -62,7 +62,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-java-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odo-java-e2e"
 
     # Run source e2e tests
     - <<: *base-test
@@ -73,7 +73,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-source-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odo-source-e2e"
 
     # Run service-catalog e2e tests
     - <<: *base-test
@@ -84,7 +84,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-service-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odo-service-e2e"
 
     # Run link e2e tests
     - <<: *base-test
@@ -95,7 +95,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-link-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odo-link-e2e"
 
     # test installation script on linux
     # - stage: test
--- a/Makefile
+++ b/Makefile
@@ -79,67 +79,14 @@ prepare-release: cross
 test:
 	go test -race $(PKGS)
 
-# Run main e2e tests
-.PHONY: test-main-e2e
-test-main-e2e:
-ifdef TIMEOUT
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoe2e" -ginkgo.succinct -timeout $(TIMEOUT)
-else
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoe2e" -ginkgo.succinct
-endif
-
-# Run component e2e tests
-.PHONY: test-cmp-e2e
-test-cmp-e2e:
-ifdef TIMEOUT
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoCmpE2e" -ginkgo.succinct -timeout $(TIMEOUT)
-else
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoCmpE2e" -ginkgo.succinct -timeout 15m
-endif
-
-# Run java e2e tests
-.PHONY: test-java-e2e
-test-java-e2e:
-ifdef TIMEOUT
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         - make bin
         - sudo cp odo /usr/bin
         - oc login -u developer
-        - make test-main-e2e
+        - make test-e2e ODO_TEST_SCENARIO="odoe2e"
```
geoand: Perhaps we should call this `odoMaine2e`?

amitkrout: hmm...Looks precise, will update it.
Thanks for the feedback :)

geoand: yw :)

metacosm: `odoMaine2e` has weird capitalization. I'd rather have `odomaine2e` or `odo-main-e2e` instead.

geoand: Then `odoMainE2e` like the rest of the changes use?

docs/development.md
```diff
@@ -182,20 +182,20 @@ Requirements:
...
 # The entire suite
 make test-e2e
 
 # Just the main tests
-make test-main-e2e
+make test-e2e ODO_TEST_SCENARIO="odoe2e"
 
 # Just component tests
-make test-cmp-e2e
+make test-e2e ODO_TEST_SCENARIO="test-cmp-e2e"
```
metacosm: Passed value doesn't match actual used value: it should be `odoCmpE2e`.

docs/development.md
```diff
@@ -182,20 +182,20 @@ Requirements:
...
 
 # Just service catalog tests
-make test-service-e2e
+make test-e2e ODO_TEST_SCENARIO="test-service-e2e"
```
metacosm: Passed value doesn't match actual used value: it should be `odoServiceE2e`.

.travis.yml
metacosm: That feels weird as well, but at least, it's consistent :)

geoand: I don't mind either way

docs/development.md
amitkrout: It was completely out of my mind. Thanks for pointing it out

amitkrout: Done

docs/development.md
amitkrout: Done

Makefile
```diff
@@ -79,67 +79,14 @@ prepare-release: cross
...
-ifdef TIMEOUT
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoLinkE2e" -ginkgo.succinct -timeout $(TIMEOUT)
-else
-	go test -v github.com/redhat-developer/odo/tests/e2e --ginkgo.focus="odoLinkE2e" -ginkgo.succinct
-endif
-
 # Run all e2e tests
 .PHONY: test-e2e
+test-e2e: ODO_TEST_SCENARIO=""
 test-e2e:
```
geoand: I am by no means a `Makefile` expert so I have to ask, is this correct?
I mean does it make sense to have 

```
test-e2e: ODO_TEST_SCENARIO=""
test-e2e:
```

?

cdrage:
I find that `make` is really fast and reducing the "load" on PHONY is neglible. To be honest, I haven't seen any other project do tests like this (using ENV variables..)

With this change, users will no longer be able to do auto-completion of all the "test" scenarios that's built-in to their terminals / OS with "make".

For example, I can quick press "tab" and get all the options for `make test-` when I enter it in the terminal.

Introducing this would make (at least my development life) more difficult as I'd have to edit the env variable with a lot more keystrokes than just simple tabbing.

WDYT @surajnarwade @kadel

amitkrout:
> With this change, users will no longer be able to do auto-completion of all the "test" scenarios that's built-in to their terminals / OS with "make".
For example, I can quick press "tab" and get all the options for make test- when I enter it in the terminal.
Introducing this would make (at least my development life) more difficult as I'd have to edit the env variable with a lot more keystrokes than just simple tabbing.

@cdrage Your concern is absolutely right. However my idea is to achieve the code reusability I mean same .PHONY will serve the purpose by passing any test spec through the ENV variable.  And also I understand it will be little hard to switch for anybody those are familiar with terminal tab completion. Let's see how others see it.

Makefile
amitkrout: Me neither ;)
Syntax seems to be correct as it does not throw any make error. I can check if this can be done in a better way

geoand:
> > With this change, users will no longer be able to do auto-completion of all the "test" scenarios that's built-in to their terminals / OS with "make".
> > For example, I can quick press "tab" and get all the options for make test- when I enter it in the terminal.
> > Introducing this would make (at least my development life) more difficult as I'd have to edit the env variable with a lot more keystrokes than just simple tabbing.
> 
> @cdrage Your concern is absolutely right. However my idea is to achieve the code reusability I mean same .PHONY will serve the purpose by passing any test spec through the ENV variable. And also I understand it will be little hard to switch for anybody those are familiar with terminal tab completion. Let's see how others see it.

I am personally fine with either way, but I would definitely like odo to follow conventions that go developers are used to ( that said, me not being a regular go developer, I don't know what those conventions are :) )

geoand:
@cdrage thanks for providing the data! In that case, I also vote **against** the change

amitkrout:
Got more vote against the pr. 

I am closing the pr as of now. IMO as test package grows there might be a hard requirement of this implementation in future to manage the test run smartly. Let see how quick we can reach that point ;)

Issue #1152, opened 2018-12-18
Support Context Dir Like oc new-app


sherl0cks:
[kind/Enhancement]

## Which functionality do you think we should update/improve?
`oc create --git` needs a` --context-dir` flag like we have in  `oc new-app`

## Why is this needed? 
Allows one to create build configs that build from a subdirectory, so we can use repos like https://github.com/jboss-openshift/openshift-quickstarts

girishramnani:
currently we don't have plans to implement multi-component in a single directory. Closing this issue for now but if someone feels this feature is needed then please re-open this issue.

Issue #839, opened 2018-10-16
Doc - Add a section to explain how to install a java, spring boot application


cmoulliard:
The existing's doc explains how to play with `nodejs`'s [application](https://github.com/redhat-developer/odo#deploying-a-nodejs-application-using-odo) but not how to install a `Java | Spring Boot application`

Tasks :  

- [ ] Enhance the [Getting started](docs/getting-started.md) page to include a java's section
- [ ] Update the [Examples](docs/examples.md) page to include the java's image

Preeticp:
TBD if it needs docs

kadel:
docs now have examples based on java and springboot, for example muliti-component example https://docs.openshift.com/container-platform/4.3/cli_reference/openshift_developer_cli/creating-a-multicomponent-application-with-odo.html

/close

PR #1153, opened 2018-12-19
adding example java docs

sherl0cks:
fixes #???
#839, but does not completely resolve. #839 suggests that the getting started page include java. This makes sense, but is a more involved rewrite. I would suggest the team 1) split 839 and 2) use the example from the katacoda to rewrite the getting started docs for java and node

## How to test changes?
It's a doc change.

A couple notes:

- due to #1152, I cannot use any of the red hat repos for java examples as they need --context-dir support. Thus, I'm using Spring Pet Clinic here. 
- I think it would be good to refactor this whole page to include all commands needed for a working example (e.g. `odo push` & `odo url create`). Creating the component is only one part of the job and makes the docs a bit confusing.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -91,7 +91,7 @@ Build and run WildFly applications on CentOS 7. For more information about using
 
 ### java 
 
-Java can be used to deploy binary artifact
+Java can be used to deploy binary artifact, for example:
 
 ```sh
   git clone https://github.com/spring-projects/spring-petclinic.git
@@ -109,4 +109,3 @@ WildFly can deploy a binary application.
   wget -O example.war 'https://github.com/appuio/hello-world-war/blob/master/repo/ch/appuio/hello-world-war/1.0.0/hello-world-war-1.0.0.war?raw=true'
   odo create wildfly --binary example.war
 ```
-
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,6 +1,6 @@
 # Examples
 
-Odo is compatible with any language or runtime listed within OpenShift's catalog of component types. To create url for component, see the [cli reference for `odo url`](cli-reference.md#url).
+Odo is compatible with any language or runtime listed within OpenShift's catalog of component types. In order to access the component over the web, you can create a URL using `odo url create`.
 
 This can be found by using `odo catalog list components`.
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,6 +1,6 @@
 # Examples
 
-Odo is compatible with any language or runtime listed within OpenShift's catalog of component types.
+Odo is compatible with any language or runtime listed within OpenShift's catalog of component types. To create url for component, see the [cli reference for `odo url`](cli-reference.md#url).
 
 This can be found by using `odo catalog list components`.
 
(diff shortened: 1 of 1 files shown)


geoand:
@cdrage Weren't you working on something similar?

docs/examples.md
```diff
@@ -1,6 +1,6 @@
 # Examples
 
-Odo is compatible with any language or runtime listed within OpenShift's catalog of component types.
+Odo is compatible with any language or runtime listed within OpenShift's catalog of component types. To create url for component, see the [cli reference for `odo url`](cli-reference.md#url).
```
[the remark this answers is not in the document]
sherl0cks: How about now?

sherl0cks:
@cdrage where can I follow along the component / database docs? I need them personally and would also like to contribute back.

sherl0cks:
This thread also makes me thing we should look at #1155

geoand:
@sherl0cks although it's definitely *not* documentation, we have various integration tests that use java.
 
I think the one that @cdrage is planning his docs update on is [this](https://github.com/redhat-developer/odo/tree/master/tests/e2e/examples/source/openjdk-sb-postgresql) and the code that uses it can be found [here](https://github.com/redhat-developer/odo/blob/master/tests/e2e/service_test.go#L40)

docs/examples.md
```diff
@@ -80,6 +89,18 @@ Build and run WildFly applications on CentOS 7. For more information about using
...
 
+### java 
+
+Java can be used to deploy binary artifact
+
+```sh
+  git clone https://github.com/spring-projects/spring-petclinic.git
+  cd spring-petclinic
+  mvn package
+  odo create java test3 --binary target/*.jar
```
metacosm: Not sure you want to push **all** jars all the time. Usually, you only push the uber-jar…

sherl0cks: Generally speaking, when packaging as an uber jar, you will only get a single jar in target - the uber jar. If you had code the name of that, simply things like maven version changes can break automation

metacosm: Generally true, yes but still risky :)

sherl0cks: OK - how about `odo create java test3 --binary target/spring-petclinic-*.jar` ? I'm trying to avoid spring reving the version and breaking the example.

sherl0cks:
bump

surajnarwade:
cc @cdrage @metacosm , need your :green_heart:

docs/examples.md
```diff
@@ -80,6 +89,18 @@ Build and run WildFly applications on CentOS 7. For more information about using
 
 ## Binary example
 
+### java 
+
+Java can be used to deploy binary artifact
```
cdrage: `Java may be used to deploy a binary artifact, for example:`

Or something similar ^^

docs/examples.md
```diff
@@ -88,3 +109,4 @@ WildFly can deploy a binary application.
   wget -O example.war 'https://github.com/appuio/hello-world-war/blob/master/repo/ch/appuio/hello-world-war/1.0.0/hello-world-war-1.0.0.war?raw=true'
   odo create wildfly --binary example.war
 ```
+
```
cdrage: newline added by mistake?

sherl0cks:
@cdrage done.

May I suggest the team look at a more [optimistic merging](http://hintjens.com/blog:106) approach? A one month cycle time for a small docs with changes for newlines isn't the greatest new contributor experience. I realize the holidays got in there, but still something to consider.

cdrage:
LGTM!

cdrage:
@sherl0cks Yeah, we need to improve on merging. We talked about this at the face-to-face meeting we had last week and we're going to be creating a newly updated DOC for merging / PR review procedures.

I send the link to the team too! http://hintjens.com/blog:106

PR #1161, opened 2018-12-20
Event bus + plugin architecture proof of concept

metacosm:
Add an event bus to be able to have decoupled extensions implemented as listeners of events produced at different stages of command execution. Commands would thus be able to push events during different stages of their execution (e.g. pre-run, post-complete, post-validate, post-run). The bus will allow interested third parties to register listeners to react to events and possibly interrupt the command processing based on this (by returning a specific error type, which will allow the bus to abort the command processing and notify listeners that might have already performed some process based on the event to revert their work).

#1173 

## How to test changes?
In progress.


pkg/odo/events/events.go
```diff
@@ -0,0 +1,212 @@
...
+	err := eventBus.DispatchEvent(Event{
+		Name:    EventNameFrom(cmd),
+		Type:    eventType,
+		Payload: payload,
+	})
+
+	return err
+}
+
+type tracer struct{}
```
golangcibot: type `tracer` is unused (from `megacheck`)

surajnarwade:
@metacosm , can you please also create the issue regarding the same and provide more context or any specific use case for this ?

girishramnani:
This PR as of now needs alot of work due to the major changes in odo. It would be better if a new PR using the current master is created at a later point when plugins become priority. 

Hence closing the PR

PR #1168, opened 2018-12-28
Support for global and local odo-config files

girishramnani:

- Breaks the config into global and local configs.
- Adds the `--global` flag to the `odo utils config` command

partly  https://github.com/redhat-developer/odo/issues/1092

## How to test changes?

```

odo utils config set --global timeout 5
odo utils config set componentType java
odo utils config view --global
odo utils config view
odo utils config unset componentType java
odo utils config unset -f componentType java
```


pkg/config/config.go
```diff
@@ -42,78 +57,117 @@ type ApplicationInfo struct {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get odo config file")
+	}
+	// Check whether directory and file are not present if they aren't then create them
+	if err = createIfNotExists(configFile); err != nil {
+		return nil, err
+	}
+	c := GlobalConfigInfo{}
+	c.Filename = configFile
+	get(c.GlobalConfig, c.Filename)
```
golangcibot: Error return value of `get` is not checked (from `errcheck`)

pkg/config/config.go
```diff
@@ -42,78 +57,117 @@ type ApplicationInfo struct {
...
 	}
-	// Check whether directory present or not
-	_, err = os.Stat(filepath.Dir(configFile))
+	// Check whether directory and file are not present if they aren't then create them
+	if err = createIfNotExists(configFile); err != nil {
+		return nil, err
+	}
+	c := LocalConfigInfo{}
+	c.Filename = configFile
+	get(c.LocalConfig, c.Filename)
```
golangcibot: Error return value of `get` is not checked (from `errcheck`)

pkg/odo/cli/utils/config.go
```diff
@@ -9,10 +9,14 @@ import (
 	"github.com/pkg/errors"
 	"github.com/redhat-developer/odo/pkg/config"
 	"github.com/redhat-developer/odo/pkg/log"
+	"github.com/redhat-developer/odo/pkg/odo/util"
 	odoutil "github.com/redhat-developer/odo/pkg/odo/util"
 	"github.com/spf13/cobra"
 )
 
+var configGlobalFlag bool
```
golangcibot: `configGlobalFlag` is unused (from `deadcode`)

pkg/odo/cli/utils/config.go
```diff
@@ -9,10 +9,14 @@ import (
...
+var configLocalFlag bool
```
golangcibot: `configLocalFlag` is unused (from `deadcode`)

geoand:
I have not followed the config discussion at all, so I don't think I'm qualified to review this

jorgemoralespou:
@girishramnani Did some basic tests with your binary, and these are some comments:

On empty configuration (~/.odo/config-odo is empty), when I do

```
$ odo utils config view --global
UpdateNotification    true
NamePrefix
Timeout               1
ComponentType
```

Are these default values? Can this be saved in the config-odo file by default somehow?

When I do:

```
$ odo utils config view
PARAMETER        CURRENT_VALUE
ComponentType
```

In this case, I don't see those default values. Why? Because they are not explicitly set as local?

When I do:

```
$ odo utils config set --global Timeout 20

$ cat ~/.odo/odo-config
activeApplications: null
component_settings: {}
settings:
  timeout: 20
```

- I don't get any output that the configuration was applied. It would be nice a user friendly message
- Why is the parameter in the file `timeout` and the parameter I provide in the CLI `Timeout`

When I do:

```
$ odo utils config set Timeout 20
Error: unknown parameter :'timeout' is not a parameter in odo config
```

How do I know which parameters are global and local? Are you going to update the help every time?

When I do:
```
$ odo utils config set ComponentType 20
$ cat .odo/odo-config
component_settings:
  componenttype: "20"
```

- I don't get any output that the configuration was applied. It would be nice a user friendly message
- Why is the parameter in the file `timeout` and the parameter I provide in the CLI `Timeout`

When I do again:

```
$ odo utils config set ComponentType java

$ cat .odo/odo-config
component_settings:
  componenttype: java
```

The value is changed. I don't get any output, and if I remember correctly, you should use `update` and not `set` as the value already exists, and the user needs to acknowledge explicitly that he's changing a config value. https://github.com/redhat-developer/odo/blob/master/docs/proposals/odo-config-file.md#updating-configuration

girishramnani:
> On empty configuration (~/.odo/config-odo is empty), when I do
> 
> ```
> $ odo utils config view --global
> PARAMETER             CURRENT_VALUE
> UpdateNotification    true
> NamePrefix
> Timeout               1
> ComponentType
> ```
> Are these default values? Can this be saved in the config-odo file by default somehow?

Yes, these are default value. Sure I can do that

> When I do:
> 
> ```
> $ odo utils config view
> PARAMETER        CURRENT_VALUE
> ComponentType
> ```
> In this case, I don't see those default values. Why? Because they are not explicitly set as local?

because the local odo config will only hold `ComponentSettings` like `ComponentType` and in future `MinMemory` etc. hence the global config will not be shown.



> When I do:
> 
> ```
> $ odo utils config set --global Timeout 20
> 
> $ cat ~/.odo/odo-config
> activeApplications: null
> component_settings: {}
> settings:
>   timeout: 20
> ```
> * I don't get any output that the configuration was applied. It would be nice a user friendly message
We can add a user friendly message :+1: . 
> * Why is the parameter in the file `timeout` and the parameter I provide in the CLI `Timeout`
because we lower case all the config params. So I followed the same.


> When I do:
> 
> ```
> $ odo utils config set Timeout 20
> Error: unknown parameter :'timeout' is not a parameter in odo config
> ```
> How do I know which parameters are global and local? Are you going to update the help every time?

We would need to add the explicitly in the help and yes would update the help every time. Unless you want to autogenerate the help for config?

> When I do again:
> 
> ```
> $ odo utils config set ComponentType java
> 
> $ cat .odo/odo-config
> component_settings:
>   componenttype: java
> ```
> The value is changed. I don't get any output, and if I remember correctly, you should use `update` and not `set` as the value already exists, and the user needs to acknowledge explicitly that he's changing a config value. https://github.com/redhat-developer/odo/blob/master/docs/proposals/odo-config-file.md#updating-configuration

But currently I think we can `set` an already set `Timeout` and other configs so not sure if we should break the behavior. 
But if we do then should I error out saying that component type has been already set? assuming this is only for `ComponentSettings` attributes.

jorgemoralespou:
> > On empty configuration (~/.odo/config-odo is empty), when I do
> > ```
> > $ odo utils config view --global
> > PARAMETER             CURRENT_VALUE
> > UpdateNotification    true
> > NamePrefix
> > Timeout               1
> > ComponentType
> > ```
> > Are these default values? Can this be saved in the config-odo file by default somehow?
> 
> Yes, these are default value. Sure I can do that
> 

@kadel WDYT? Or we can just show in a way which are coming from the user's config file and which are coming from the tool's default. Also, I don't think `ComponentType` should be in the global config.

> > When I do:
> > ```
> > $ odo utils config view
> > PARAMETER        CURRENT_VALUE
> > ComponentType
> > ```
> > In this case, I don't see those default values. Why? Because they are not explicitly set as local?
> 
> because the local odo config will only hold `ComponentSettings` like `ComponentType` and in future `MinMemory` etc. hence the global config will not be shown.
> 

Sounds good.

> > When I do:
> > ```
> > $ odo utils config set --global Timeout 20
> > 
> > $ cat ~/.odo/odo-config
> > activeApplications: null
> > component_settings: {}
> > settings:
> >   timeout: 20
> > ```
> > 
> > * I don't get any output that the configuration was applied. It would be nice a user friendly message
> 
> We can add a user friendly message 👍 .
> 

Let's do it, as user will want information of what and how things happened. Optionally, when makes sense, we should give TIPS as we discussed in F2F, if there's a common next action.

> > * Why is the parameter in the file `timeout` and the parameter I provide in the CLI `Timeout`
> 
> because we lower case all the config params. So I followed the same.
> 

I don't think we should do this. We should keep casing in the parameters the same as in the config file. Also, I think we should use camelCase, which means `timeout` and  `updateNotification` and `componentType`. @kadel WDYT? 

> > When I do:
> > ```
> > $ odo utils config set Timeout 20
> > Error: unknown parameter :'timeout' is not a parameter in odo config
> > ```
> > How do I know which parameters are global and local? Are you going to update the help every time?
> 
> We would need to add the explicitly in the help and yes would update the help every time. Unless you want to autogenerate the help for config?
> 

We should probably need a flag/command to know what things can be set as parameters and whether they are global or local or both.  @kadel WDYT?

> > When I do again:
> > ```
> > $ odo utils config set ComponentType java
> > 
> > $ cat .odo/odo-config
> > component_settings:
> >   componenttype: java
> > ```
> > The value is changed. I don't get any output, and if I remember correctly, you should use `update` and not `set` as the value already exists, and the user needs to acknowledge explicitly that he's changing a config value. https://github.com/redhat-developer/odo/blob/master/docs/proposals/odo-config-file.md#updating-configuration
> 
> But currently I think we can `set` an already set `Timeout` and other configs so not sure if we should break the behavior.

Yeah, as anyways this is breaking as using different config file. This is what is in the proposal and I would stick to it. @kadel WDYT?

> But if we do then should I error out saying that component type has been already set? assuming this is only for `ComponentSettings` attributes.

Yes, and give the right command, e.g.

```
$ odo utils config set componentType java
That configuration option already exists with value: `nodejs`
To update it use: `odo utils config update componentType java
```

pkg/config/config.go
```diff
@@ -14,12 +14,18 @@ import (
 )
 
 const (
-	configEnvName  = "ODOCONFIG"
-	configFileName = "odo"
+	globalConfigEnvName = "GLOBALODOCONFIG"
+	localConfigEnvName  = "LOCALODOCONFIG"
+	configFileName      = "odo-config"
```
kadel: it might be better to have file with extension ¨odo-config.yaml` to show what format we are using

girishramnani: ack

girishramnani:
- we have considered using `camelCase` for config
- we will not show default configs when doing `odo utils config view` instead show the defaults in description of `set` ( maybe )
- implement `config set` with the workflow mentioned by @kadel 
e.g.
```
$ odo utils config set Timeout 20
Local config was successfully updated.

$ odo utils config set Timeout 30
Timeout is already set. Current value is 30.
Do you want to override it in the local config? y/n


$ odo utils config set -f Timeout 30
Local config was successfully updated.

```

Need to discuss with @jorgemoralespou 
- Global vs Local config e.g. should we have `ComponentType` in global config

jorgemoralespou:
> But, shouldn't it be possible to have every global parameter also as local?
I remember that we had this discussion, but forgot what was the conclusion, sorry :-(

I don't think that every parameter makes sense as global or local as there are some that need to be only in one scope. Maybe we need to expand the implementation to support the scope as local, global, both.

e.g. UpdateNotification should only be global, componentName and componentType should only be local.

As per the update, we already do this with component (create, update) but I can't be disagreeing all the time ;-)

kadel:
I had a short discussion with @jorgemoralespou.

Regardless where we put `ComponentType`, we will be adding more local and global options in near future, and the code needs to be prepared that there might be config options that will be only local or only global.

jorgemoralespou:
> I don't think odo itself should have a local .odo folder, should it?

Where would you store local config? odo-config file needs to be in a local .odo folder, as potentially more files relevant to odo will eventually end up there

girishramnani:
Yes but this odo project shouldn't have a .odo folder, odo project itself
doesn't have a component in it, does it?

jorgemoralespou:
> Yes but this odo project shouldn't have a .odo folder, odo project itself doesn't have a component in it, does it?

Not sure what you mean with this `odo project`? I'm happy to jump into a BJN chat and move this forward :-D Ping me on mattermost

pkg/config/config.go
```diff
@@ -163,19 +241,73 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
...
+	if val.Kind() == reflect.Ptr {
+		return val.Elem().Interface(), true
+	}
+
+	return val.Interface(), true
+}
+
+func caseInsensitive(parameter string) func(word string) bool {
+	return func(word string) bool {
+		return strings.ToLower(parameter) == strings.ToLower(word)
```
golangcibot: should use strings.EqualFold(a, b) instead of strings.ToLower(a) == strings.ToLower(b) (from `staticcheck`)

jorgemoralespou:
@girishramnani 

- ISSUE

When I do:
```
odo utils config view
PARAMETER        CURRENT_VALUE
ComponentType
```

But the local file is empty.
```
cat .odo/odo-config.yaml
```

- ISSUE
  - OK: When I do any action, the .odo folder and .odo/odo-config.yaml file is created.
  - NOOK: But this happens in any directory where I execute odo command. IMHO this dir/file should only be created on `odo create` command as otherwise there's no component. @kadel WDYT?

- ISSUE: There's no way to see all configuration at once, you need to do either nothing (local) or --global to see the configuration, but you can see both at same time. I think a `--all` would probably make sense, and in this case, you would also need to print out the scope where the config exist.

- ISSUE: I know I've said this before, but it needs to be consistent what you provide via CLI to what's written in the file:
```
$ odo utils config set ComponentType java
Local config was successfully updated.
 
$ cat .odo/odo-config.yaml
componentSettings:
  componentType: java

$ odo utils config view
PARAMETER        CURRENT_VALUE
ComponentType    java
```
In this example, you set `ComponentType` but the file has `componentType` although it prints correctly. It's not keen to manual edits :-(

- ISSUE: I guess it's not part yet of this PR, but once I have set the componentType locally, I can not just create a component with `odo create`

- ISSUE: Flags that I provide to `odo create` are not persisted. I guess again, this is not part of this PR.

- OK: Overriding a value prints out a message and confirmation.
- OK: Overriding with confirmation `-f` just overwrites the value.

- ISSUE: A config can not be deleted. `odo utils config delete ... --> Unknown command, use "set" or "view"`

- BUG: Help does not properly align example messages:
```
   # Set a global configuration value
   odo utils config set --global UpdateNotification false
   odo utils config set --global NamePrefix "app"
   odo utils config set --global Timeout 20

   # Set a local configuration value
   odo utils config set ComponentType java

  # For viewing the local configuration
   odo utils config view
```

- OK: Completion works fine

kadel:
> * NOOK: But this happens in any directory where I execute odo command. IMHO this dir/file should only be created on `odo create` command as otherwise there's no component. @kadel WDYT?

+1. The config file should be created only when once you run `odo create` or `odo utils config set --local`


> * ISSUE: There's no way to see all configuration at once, you need to do either nothing (local) or --global to see the configuration, but you can see both at same time. I think a `--all` would probably make sense, and in this case, you would also need to print out the scope where the config exist.

Good point. 
But I would prefer to add this in next PR and not in this one. We would like to progress in small PRs, one step at a time. 

> * ISSUE: I know I've said this before, but it needs to be consistent what you provide via CLI to what's written in the file:

+1 . What is displayed in `odo utils config view` has to match what is in config file and it has to be unified. We can accept any casing for `odo utils config set` command, but it has to be saved in normalized form.
```
$ odo utils config set tImEoUt 10

$ odo util config view
PARAMETER CURRENT_VALUE 
Timeout 10

$ cat .odo/odo-config.yaml 
Timeout: 10
```

> * ISSUE: I guess it's not part yet of this PR, but once I have set the componentType locally, I can not just create a component with `odo create`
> * ISSUE: Flags that I provide to `odo create` are not persisted. I guess again, this is not part of this PR.

Yes this should be added in the followup PRs.


> * ISSUE: A config can not be deleted. `odo utils config delete ... --> Unknown command, use "set" or "view"`

+1 we need delete. Completly forgot about that.

cmd/odo/odo.go
```diff
@@ -53,7 +54,7 @@ func main() {
 	// Call commands
 	// checking the value of updatenotification in config
 	// before proceeding with fetching the latest version
-	cfg, err := config.New()
+	cfg, err := config.NewGlobalConfig()
```
anmolbabu: Why is this gloabal config?
Any package outside the pkg/config should in my opinion not refer to either global or local config explicitly in my opinion.
even for commands that request update of a configurable to local or global config explicitly, I would say pass on the states of the respective flags/defaults as applicable onto a single config update function in the pkg/config package and let no outside packages treat them separately? 

Basically external packages shouldn't do in my opinion, NewGlobalConfig or NewLocalConfig may be just NewConfig and pass on everything collected from user down to pkg/config and let all intelligence of config to the pkg/config

girishramnani: I can just revert this to `New()` and essentially that proxies to `NewGlobalConfig()` for now ( that will change in the next PR )

girishramnani:
The rebase would require me to change my code to work with the `Complete-Validate-Run` for config. So working on that now

pkg/config/config.go
```diff
@@ -423,8 +567,14 @@ var (
...
 	}
+
+	supportedLocalParameterDescriptions = map[string]string{
+		ComponentType: ComponentTypeDescription,
+	}
 	// set-like map to quickly check if a parameter is supported
-	lowerCaseParameters = getLowerCaseParameters()
+	lowerCaseParameters = getLowerCaseParameters(GetSupportedParameters())
+
+	lowerCaseLocalParameters = getLowerCaseParameters(GetLocallySupportedParameters())
```
golangcibot: `lowerCaseLocalParameters` is unused (from `deadcode`)

pkg/odo/cli/utils/config/delete.go
```diff
@@ -0,0 +1,116 @@
...
+   # Delete a configuration value in the local config
+   %[1]s %[5]s
+	`)
+)
+
+// DeleteOptions encapsulates the options for the command
+type DeleteOptions struct {
+	paramName        string
+	configGlobalFlag bool
+	configForceFlag  bool
```
golangcibot: `configForceFlag` is unused (from `structcheck`)

golangcibot: `configForceFlag` is unused (from `structcheck`)

kadel:
capitalization is still different in the file and what is shown on the command line 

```
▶ odo utils config view
PARAMETER        CURRENT_VALUE
ComponentType    fff

▶ cat .odo/odo-config.yaml
componentSettings:
  componentType: fff

```

docs/cli-reference.md
```diff
@@ -607,18 +607,25 @@ The URLs that are generated using this command, can be used to access the deploy
...
 
 
-   # Set a configuration value
-   odo utils config set UpdateNotification false
-   odo utils config set NamePrefix "app"
-   odo utils config set Timeout 20
+   # Set a global configuration value
+   odo utils config set --global UpdateNotification false
+   odo utils config set --global NamePrefix "app"
+   odo utils config set --global timeout 20
```
kadel: Why timeout starts with lowercase, but everything else with upper?

pkg/config/config.go
```diff
@@ -2,35 +2,52 @@ package config
...
+	SetConfiguration(parameter string, value string) error
+	GetConfiguration(parameter string) (interface{}, bool)
+	DeleteConfiguration(parameter string) error
+}
+
 // OdoSettings holds all odo specific configurations
 type OdoSettings struct {
 	// Controls if an update notification is shown or not
-	UpdateNotification *bool `json:"updatenotification,omitempty"`
+	UpdateNotification *bool `json:"updateNotification,omitempty"`
```
kadel: Maybe we can have everything "CamelCase" instead of "camelCase" ("UpdateNotification")

pkg/config/config.go
```diff
@@ -43,107 +60,174 @@ type ApplicationInfo struct {
...
 }
 
-// ConfigInfo records information about odo's configuration
-type ConfigInfo struct {
-	Filename string
-	Config
+// GlobalConfigInfo wraps the global config and provides helpers to
+// serialize it.
+type GlobalConfigInfo struct {
+	Filename     string `json:"fileName,omitempty" yaml:"fileName,omitempty"`
```
kadel: why we have `yaml` tag here, but not anywhere else?
`yaml` marshall looks for `json` tags, we shouldn't this info twice.

pkg/config/config.go
```diff
@@ -43,107 +60,174 @@ type ApplicationInfo struct {
...
+	GlobalConfig `json:",omitempty" yaml:",omitempty"`
```
kadel: no name in the json tag?

pkg/config/config.go
girishramnani: ack

pkg/config/config.go
girishramnani: yes because its embedded

pkg/config/config.go
girishramnani: ack

docs/cli-reference.md
girishramnani: ack

pkg/config/config.go
kadel: Does it require json tag than?

girishramnani: for omitempty part yes

docs/cli-reference.md
girishramnani: need to generate the cli-reference

girishramnani: resolved

pkg/config/config.go
girishramnani: resolved

pkg/config/config.go
girishramnani: resolved

pkg/config/config.go
```diff
@@ -2,148 +2,232 @@ package config
...
+
+	// The builder image to use
+	ComponentType *string `json:"ComponentType,omitempty"`
 }
 
 // ApplicationInfo holds all important information about one application
 type ApplicationInfo struct {
 	// name of the application
-	Name string `json:"name"`
+	Name string `json:"Name"`
```
kadel: 👍

girishramnani: just so the whole file follows the convention of `CamelCase`

pkg/config/config.go
girishramnani: resolved

pkg/config/config.go
girishramnani: resolved

pkg/config/config.go
```diff
@@ -169,18 +252,147 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
...
 	}
 	return nil
 }
 
+// DeleteConfiguration delete Odo configurations in the global config file
+// as of now being used for nameprefix, timeout, updatenotification
+func (c *GlobalConfigInfo) DeleteConfiguration(parameter string) error {
+	if p, ok := asSupportedParameter(parameter); ok {
+		// processing values according to the parameter names
+		switch p {
```
kadel: this needs to have todo, to update it to use reflect

kadel: you have nice, generic `getConfiguration` why not do the same here?

kadel: If you don't want to do it now, that it needs to have TODO

pkg/config/config.go
```diff
@@ -169,18 +252,147 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
...
+
+func caseInsensitive(parameter string) func(word string) bool {
+	return func(word string) bool {
+		return strings.EqualFold(word, parameter)
+	}
+}
+
+// SetConfiguration sets the common config settings like component type, min memory
+// max memory etc.
+func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (err error) {
```
kadel: TODO for using reflect

pkg/config/config.go
```diff
@@ -169,18 +252,147 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
...
+// GetConfiguration uses reflection to get the parameter from the localconfig struct, currently
+// it only searches the componentSettings
+func (lci *LocalConfigInfo) GetConfiguration(parameter string) (interface{}, bool) {
+	return getConfiguration(lci.ComponentSettings, parameter)
+}
+
+// DeleteConfiguration is used to delete config from local odo config
+func (lci *LocalConfigInfo) DeleteConfiguration(parameter string) error {
+	if parameter, ok := asLocallySupportedParameter(parameter); ok {
+		switch parameter {
```
kadel: the same as for DeleteConfiguration for GlobalConfig

pkg/config/config.go
```diff
@@ -169,18 +252,147 @@ func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
...
+		if allNil(lci.ComponentSettings) {
+			lci.ComponentSettings = nil
+		}
+		return writeToFile(lci.LocalConfig, lci.Filename)
+	}
+	return errors.Errorf("unknown parameter :'%s' is not a parameter in local odo config", parameter)
+
+}
+
+func allNil(info interface{}) bool {
```
kadel: it would be nice to have comment what it is doing

jorgemoralespou:
Also,
Running `odo utils config view` in a directory without `.odo` directory, yields result:

```
odo utils config view
PARAMETER        CURRENT_VALUE
ComponentType
```

girishramnani:
@cdrage you need to run `odo utils config view --global ` to view global configs like `Timeout`.
Also the `ComponentType` is not a global config so it can be only set locally

girishramnani:
@jorgemoralespou that's the intended behaviour

girishramnani:
> Also,
> Running `odo utils config view` in a directory without `.odo` directory, yields result:
> 
> ```
> odo utils config view
> PARAMETER        CURRENT_VALUE
> ComponentType
> ```

This is the intended behaviour @jorgemoralespou

girishramnani:
> code looks mostly good.
> Just one change in behaviour, and user interaction.
> 
> I think that it would be better to have "odo utils config unset" instead of "odo utils config delete".
> Delete should ask for confirmation similar similar to calling set on already set variable.
> 
> Also the message for calling double delet(unset) and calling it on nonexistin varible should be different
> 
> ```
> ▶ odo utils config delete --global ff
>  ✗  config already unset, cannot delete an unset configuration
> ```
> Should be `ff is not a valid configuration variable` or something like this.

Then do you need a `-f` to bypass user conformation?

cdrage:
@girishramnani Are you able to update your PR description and commit to remove the non-valid commands? That's where I got the `--global ComponentType` command from.

girishramnani:
@cdrage done

jorgemoralespou:
> > Also,
> > Running `odo utils config view` in a directory without `.odo` directory, yields result:
> > ```
> > odo utils config view
> > PARAMETER        CURRENT_VALUE
> > ComponentType
> > ```
> 
> This is the intended behaviour @jorgemoralespou

What's the rationale?

girishramnani:
> > > Also,
> > > Running `odo utils config view` in a directory without `.odo` directory, yields result:
> > > ```
> > > odo utils config view
> > > PARAMETER        CURRENT_VALUE
> > > ComponentType
> > > ```
> > 
> > 
> > This is the intended behaviour @jorgemoralespou
> 
> What's the rationale?

to show the user the possible values they can configure

girishramnani:
I am open to suggestions, what do you feel is better here @jorgemoralespou ?

jorgemoralespou:
IMHO, it doesn't make sense to show the possible options if you're not in
the context of an .odo project as, what would you do with it?
Do we think a use will set this up before doing `odo create`. While
possible seems not the most plausible scenario.

girishramnani:
So should I show "cannot read the local odo configuration" or "cannot find
local odo configuration" as error?

kadel:
> > code looks mostly good.
> > Just one change in behaviour, and user interaction.
> > I think that it would be better to have "odo utils config unset" instead of "odo utils config delete".
> > Delete should ask for confirmation similar similar to calling set on already set variable.
> > Also the message for calling double delet(unset) and calling it on nonexistin varible should be different
> > ```
> > ▶ odo utils config delete --global ff
> >  ✗  config already unset, cannot delete an unset configuration
> > ```
> > Should be `ff is not a valid configuration variable` or something like this.
> 
> Then do you need a `-f` to bypass user conformation?

yes.
Can we please also rename command to  `unset`?

kadel:
> IMHO, it doesn't make sense to show the possible options if you're not in the context of an .odo project as, what would you do with it? Do we think a use will set this up before doing `odo create`. While possible seems not the most plausible scenario.

This is a valid concern, but we are not there yet. `odo create` is currently not creating config file :-(
This is something that will be done in the next PR. 

A lot that is done in this PR will change based on the new workflow. But this PR if the first one that establishes the local config, the others will build on this.

girishramnani:
@kadel done with the unset and force flag

tests/e2e/e2e_test.go
```diff
@@ -115,17 +115,17 @@ var _ = Describe("odoe2e", func() {
 			Expect(configOutput).To(ContainSubstring("ComponentType"))
 		})
 
-		It("should allow deleting a set config locally", func() {
+		FIt("should allow unsetting a config locally", func() {
```
kadel: Shouldn't this remain as `It`?

amitkrout: Should be ```It```. I guess it is a typo, However surprisingly not caught in CI. It could be a literal in ```GonkGo```.

tests/e2e/e2e_test.go
```diff
@@ -89,26 +89,48 @@ var _ = Describe("odoe2e", func() {
...
+
+		FIt("should allow unsetting a config locally", func() {
+			runCmd("odo utils config set componenttype java")
+			configOutput := runCmd("odo utils config unset -f componentType")
+			Expect(configOutput).To(ContainSubstring("Local config was successfully updated."))
+			configOutput = runCmd("odo utils config view|grep ComponentType")
+			Expect(configOutput).NotTo(ContainSubstring("java"))
+		})
+
+		FIt("should allow unsetting a config globally", func() {
```
amitkrout: Same as above

tests/e2e/e2e_test.go
kadel: `Fit` is valid - it is Focused Spec - https://onsi.github.io/ginkgo/#focused-specs

kadel:
just `It` vs `FIt` issue, otherwise LGTM

tests/e2e/e2e_test.go
amitkrout: Ahh...Right
@girishramnani What you want to achieve here. I guess you are using it for your debugging propose, right ?

girishramnani: yeah

girishramnani: resolved

tests/e2e/e2e_test.go
girishramnani: resolved

anmolbabu:
Merging based on 4 acks

anmolbabu:
Oops I am sorry will be able to merge only after ack from @cdrage

geoand:
> Oops I am sorry will be able to merge only after ack from @cdrage

I don' think it makes sense to wait for an ack when the PR already has 4.
Moreover, Charlie will be on PTO soon and might not get a change to review it again

girishramnani:
yeah waiting for @kadel to take a decision on this

anmolbabu:
> > Oops I am sorry will be able to merge only after ack from @cdrage
> 
> I don' think it makes sense to wait for an ack when the PR already has 4.
> Moreover, Charlie will be on PTO soon and might not get a change to review it again

Yeah but I think the merge button is disabled for me(it'll be until all requested changes are addressed and acked by reviewer requesting change.) In my opinion, may be only those with admin privilege can merge it or, if change requesting review is dismissed, probably only then merge will be legal.. :)

But I would let @kadel take a call as @girishramnani mentions

Issue #1174, opened 2019-01-06
Improve link error message with --port usage


sspeiche:
[kind/bug]

## Actual behavior
`odo link backend --component frontend`
 ✗  Multiple secrets exist for component backend. 
Please select one of the following ports: '8080,8443,8778'

## Expected behavior
I would have expected instruction on how to specify the ports, such as:

Please select one of the following ports: '8080,8443,8778' by supplying the `--port` option and rerun the command.

PR #1176, opened 2019-01-07
Improve error message when a port is needed for odo link

geoand:
See $SUBJECT

Fixes: #1174

## How to test changes?
Create a java backend component (which exposes multiple ports).
Create a frontend component.
Try to link the frontend to the backend. The error should now be what #1174 describes

### DIFF
--- a/pkg/secret/secret.go
+++ b/pkg/secret/secret.go
@@ -33,7 +33,7 @@ Please delete the component and recreate it using 'odo create'`, componentName)
 		if len(secrets) == 1 {
 			return secrets[0].Name, nil
 		}
-		return "", fmt.Errorf("Multiple secrets exist for component %s. "+
+		return "", fmt.Errorf("Unable to properly link to component %s. "+
 			"Please select one of the following ports: '%s' "+
 			"by supplying the --port option and rerun the command", componentName, strings.Join(availablePorts(secrets), ","))
 	}
@@ -44,7 +44,7 @@ Please delete the component and recreate it using 'odo create'`, componentName)
 			return secret.Name, nil
 		}
 	}
-	return "", fmt.Errorf("None of the secrets that exist for component %s match port %s. "+
+	return "", fmt.Errorf("Unable to properly link to component %s. "+
 		"Please select one of the following ports: '%s' "+
 		"by supplying the --port option and rerun the command", componentName, port, strings.Join(availablePorts(secrets), ","))
 }
--- a/pkg/secret/secret.go
+++ b/pkg/secret/secret.go
@@ -34,7 +34,8 @@ Please delete the component and recreate it using 'odo create'`, componentName)
 			return secrets[0].Name, nil
 		}
 		return "", fmt.Errorf(`Multiple secrets exist for component %s. 
-Please select one of the following ports: '%s'`, componentName, strings.Join(availablePorts(secrets), ","))
+Please select one of the following ports: '%s' 
+by supplying the --port option and rerun the command`, componentName, strings.Join(availablePorts(secrets), ","))
 	}
 
 	// search each secret to see which port is corresponds to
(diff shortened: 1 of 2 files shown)


girishramnani:
@geoand https://github.com/redhat-developer/odo/blob/6055931a8ea917d93306351f7878e0ce06f604e0/pkg/secret/secret.go#L47 might also need to be changed

geoand:
> @geoand
> 
> [odo/pkg/secret/secret.go](https://github.com/redhat-developer/odo/blob/6055931a8ea917d93306351f7878e0ce06f604e0/pkg/secret/secret.go#L47)
> 
> Line 47 in [6055931](/redhat-developer/odo/commit/6055931a8ea917d93306351f7878e0ce06f604e0)
>  Please select one of the following ports: '%s'`, componentName, port, strings.Join(availablePorts(secrets), ",")) 
> might also need to be changed

You're absolutely right, fixing now

girishramnani:
a suggestion on the message going above 80 characters in one line, can we move a part to the next line ( not add a \n in the actual message ) just new line in the code.

geoand:
> a suggestion on the message going above 80 characters in one line, can we move a part to the next line ( not add a \n in the actual message ) just new line in the code.

Makes sense, fixing now

pkg/secret/secret.go
```diff
@@ -34,7 +34,8 @@ Please delete the component and recreate it using 'odo create'`, componentName)
 			return secrets[0].Name, nil
 		}
 		return "", fmt.Errorf(`Multiple secrets exist for component %s. 
-Please select one of the following ports: '%s'`, componentName, strings.Join(availablePorts(secrets), ","))
+Please select one of the following ports: '%s' 
+by supplying the --port option and rerun the command`, componentName, strings.Join(availablePorts(secrets), ","))
```
girishramnani: as you are using back ticks `` ` ``, the new line is actually not ignored and the `by supplying ...` prints in the next line.
Might wanna consider
 ```
"stuff " +
"next line" 
```

Comment (geoand) [reply pkg/secret/secret.go]: Hm... The message I am getting in the terminal  is the following:

```
None of the secrets that exist for component backend match port 9324. 
Please select one of the following ports: '8080' 
by supplying the --port option and rerun the command
```

 which looks correct. What am I missing here?

Comment (girishramnani) [reply pkg/secret/secret.go]: the wrapping of the message should be taken care by the terminal, try to resize the terminal once. This is what it looks like to me
image

Comment (girishramnani) [reply pkg/secret/secret.go]: so we shouldn't add a new line in the message but new line in the code so its more readable.

File: pkg/secret/secret.go [hunk 2]
```diff
@@ -33,8 +33,9 @@ Please delete the component and recreate it using 'odo create'`, componentName)
 		if len(secrets) == 1 {
 			return secrets[0].Name, nil
 		}
-		return "", fmt.Errorf(`Multiple secrets exist for component %s. 
-Please select one of the following ports: '%s'`, componentName, strings.Join(availablePorts(secrets), ","))
+		return "", fmt.Errorf("Multiple secrets exist for component %s. "+
+			"Please select one of the following ports: '%s' "+
```
mik-dass: Maybe something like
```
Please rerun the command by supplying the --port option with one of the following ports : `%s`
```
???

pkg/secret/secret.go
```diff
@@ -43,9 +44,9 @@ Please select one of the following ports: '%s'`, componentName, strings.Join(ava
 			return secret.Name, nil
 		}
 	}
-	return "", fmt.Errorf(`None of the secrets that exist for component %s match port %s. 
-Please select one of the following ports: '%s'`, componentName, port, strings.Join(availablePorts(secrets), ","))
-
+	return "", fmt.Errorf("None of the secrets that exist for component %s match port %s. "+
+		"Please select one of the following ports: '%s' "+
```
mik-dass: Same as above

pkg/secret/secret.go
geoand: Well, I just followed what was added in the open issue :)

mik-dass: That was just a suggestion :)

geoand: I don't mind either way of course :P

mik-dass: Ofcourse the decision is upto you :D

geoand: Let's go with this for the time being and if anyone complains we can always change it again

mik-dass: ok :+1:

geoand:
@girishramnani One of the new integration tests seems to be flaky. Mind taking a look please?

girishramnani:
Sure

geoand:
Thanks!

PR #1181, opened 2019-01-08
added check in tags for services in catalog

girishramnani:
This change checks for `hidden` tag in the services and hides them.

partly https://github.com/redhat-developer/odo/issues/1180

## How to test changes?
`odo catalog list services`
`odo catalog search service <service>`

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -19,7 +19,6 @@ type CatalogImage struct {
 
 // List lists all the available component types
 func List(client *occlient.Client) ([]CatalogImage, error) {
-
 	catalogList, err := getDefaultBuilderImages(client)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get image streams")
@@ -138,14 +137,13 @@ func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
 	// Resultant imagestreams is list of imagestreams from current and openshift namespaces
 	imageStreams = append(imageStreams, openshiftNSImageStreams...)
 	imageStreams = append(imageStreams, currentNSImageStreams...)
-
 	var builderImages []CatalogImage
 
 	// Get builder images from the available imagestreams
 	for _, imageStream := range imageStreams {
 		var allTags []string
 		buildImage := false
-
+		hidden := false
 		for _, tag := range imageStream.Spec.Tags {
 
 			allTags = append(allTags, tag.Name)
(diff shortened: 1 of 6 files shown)


pkg/occlient/occlient.go
```diff
@@ -1936,16 +1938,18 @@ func (c *Client) GetClusterServiceClass(serviceName string) (*scv1beta1.ClusterS
 		FieldSelector: fields.OneTermEqualSelector("spec.externalName", serviceName).String(),
 	}
 	searchResults, err := c.serviceCatalogClient.ClusterServiceClasses().List(opts)
+
+	items := filterHiddenClasses(searchResults.Items)
```
geoand: To be honest, I don't think this should be done. 
From what I understand, `hidden` is meant only to hide things when all services are listed, not when obtaining info in other ways

pkg/catalog/catalog.go
```diff
@@ -138,22 +137,18 @@ func getDefaultBuilderImages(client *occlient.Client) ([]CatalogImage, error) {
...
-
 	var builderImages []CatalogImage
 
 	// Get builder images from the available imagestreams
 	for _, imageStream := range imageStreams {
 		var allTags []string
 		buildImage := false
-
 		for _, tag := range imageStream.Spec.Tags {
-
```
cdrage: @girishramnani you you need to revert these changes, you've deleted the newlines

pkg/occlient/utils.go
```diff
@@ -0,0 +1,23 @@
+package occlient
+
+import scv1beta1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+
+func isHidden(tags []string) bool {
```
metacosm: I'd actually make this function a little more generic: `hasTag(tags []string, tag string)` instead of making it specific to `hidden`. Asking selfishly because I actually need such a function ;)

girishramnani: ack

geoand:
We could have an integration test, using one of the `sso` services that should be hidden (don't remember which one it is, @jorgemoralespou should now)

tests/e2e/service_test.go
```diff
@@ -81,6 +83,14 @@ var _ = Describe("odoServiceE2e", func() {
...
 			})
 		})
+
+		Context("odo hides a hidden service in service catalog", func() {
+			It("not show a hidden service in the catalog", func() {
+				runCmd("oc apply -f https://github.com/openshift/library/raw/master/official/sso/templates/sso72-https.json -n openshift")
+				output := runFailCmd("odo catalog search service sso72-https")
+				Expect(output).To(ContainSubstring("No service matched the query: sso72-https"))
+			})
+		})
```
amitkrout: +1

girishramnani:
Merging

PR #1182, opened 2019-01-09
Error output should to go stderr

cdrage:
Error output should be outputted to stderr rather than stdout.

### DIFF
--- a/tests/e2e/cmp_test.go
+++ b/tests/e2e/cmp_test.go
@@ -55,7 +55,8 @@ var _ = Describe("odoCmpE2e", func() {
 		})
 
 		It("should show an error when ref flag is provided with sources except git", func() {
-			output := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
+			output, _ := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
+
 			Expect(output).To(ContainSubstring("The --ref flag is only valid for --git flag"))
 		})
 
--- a/tests/e2e/e2e.go
+++ b/tests/e2e/e2e.go
@@ -27,8 +27,8 @@ func runCmd(cmdS string) string {
 }
 
 // runFailCmd runs a failing command
-// and returns the stdout
-func runFailCmd(cmdS string) string {
+// and returns the stdout and stderr
+func runFailCmd(cmdS string) (string, error) {
 	cmd := exec.Command("/bin/sh", "-c", cmdS)
 	fmt.Fprintf(GinkgoWriter, "Running command: %s\n", cmdS)
 	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
@@ -38,7 +38,7 @@ func runFailCmd(cmdS string) string {
 	Expect(session.ExitCode()).To(Equal(1))
 	Expect(err).NotTo(HaveOccurred())
 
-	return string(session.Out.Contents())
+	return string(session.Out.Contents()), err
 }
 
 // waitForCmdOut runs a command until it gets
--- a/pkg/log/status.go
+++ b/pkg/log/status.go
@@ -162,13 +162,13 @@ func Successf(format string, a ...interface{}) {
 // Errorf will output in an appropriate "progress" manner
 func Errorf(format string, a ...interface{}) {
 	red := color.New(color.FgRed).SprintFunc()
-	fmt.Printf(" %s%s%s\n", red("✗"), suffixSpacing, fmt.Sprintf(format, a...))
+	fmt.Fprintf(os.Stderr, " %s%s%s\n", red("✗"), suffixSpacing, fmt.Sprintf(format, a...))
 }
 
 // Error will output in an appropriate "progress" manner
 func Error(a ...interface{}) {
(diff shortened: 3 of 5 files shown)


geoand:
@cdrage does the test failure seem relevant?

mik-dass:
@cdrage The test is reproducible locally. The error thrown at https://github.com/redhat-developer/odo/blob/712d227336f97e0341d74d308e09250bd120ea62/tests/e2e/cmp_test.go#L58 seems to be outputted to stderr.

amitkrout:
> @cdrage @mik-dass doesn't seem to be outputted to stderr.

It won't be because the function ```runFailCmd``` only returns stdout of type string. This could be helpful
```
// runFailCmd runs a failing command
// and returns the stdout and stderr
func runFailCmd(cmdS string) string, error {
	cmd := exec.Command("/bin/sh", "-c", cmdS)
	fmt.Fprintf(GinkgoWriter, "Running command: %s\n", cmdS)
	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)

	// wait for the command execution to complete
	<-session.Exited
	Expect(session.ExitCode()).To(Equal(1))
	Expect(err).NotTo(HaveOccurred())

	return string(session.Out.Contents()), err
}
```
Then call this as 
```
output, err := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t)) 
```

tests/e2e/cmp_test.go
```diff
@@ -55,7 +55,8 @@ var _ = Describe("odoCmpE2e", func() {
 		})
 
 		It("should show an error when ref flag is provided with sources except git", func() {
-			output := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
+			output, err := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

amitkrout:
@cdrage Can you please have a look why UT coverage is down by some percentage. I can see those files are mentioned by the bot https://github.com/redhat-developer/odo/pull/1182#issuecomment-454162844 is nowhere updated in the pr.

tests/e2e/e2e.go
```diff
@@ -38,7 +38,7 @@ func runFailCmd(cmdS string) string {
 	Expect(session.ExitCode()).To(Equal(1))
 	Expect(err).NotTo(HaveOccurred())
```
amitkrout: I do not think ```Expect(err).NotTo(HaveOccurred())``` no more valid because err is returned through ```return string(session.Out.Contents()), err``` anyway.

tests/e2e/cmp_test.go
```diff
@@ -55,7 +55,8 @@ var _ = Describe("odoCmpE2e", func() {
 		})
 
 		It("should show an error when ref flag is provided with sources except git", func() {
-			output := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
+			output, _ := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
+
 			Expect(output).To(ContainSubstring("The --ref flag is only valid for --git flag"))
```
amitkrout: Here return of the command is captured in stderr not in output causing CI failure.  

Try something like 
_, stderr := runFailCmd(fmt.Sprintf("odo create nodejs cmp-git-%s --local test --ref test", t))
Expect(fmt.Fprintf(stderr)).To(ContainSubstring("The --ref flag is only valid for --git flag"))

tests/e2e/e2e.go
amitkrout: @cdrage Don't you think these two line are contradicting
```
Expect(session.ExitCode()).To(Equal(1))
Expect(err).NotTo(HaveOccurred()) 
```
I am surprised how it is working then. Ping @girishramnani

girishramnani: actually the gexec.Start's `err` is the error returned by `Cmd.Start()` https://github.com/onsi/gomega/blob/master/gexec/session.go#L91-L99. This doesn't represent the error corresponding to how the command pans out, its related to the error which could occur while starting the command. One of the error returned while starting the command is PathError (https://golang.org/pkg/os/#StartProcess) . `StartProcess` is used by `cmd.Start()` to start the process.

girishramnani: so essentially @amitkrout you can read those statements as 
- The outcome of the command when it exits should be error ( exit code = 1 )
- The command should be executed ( there shouldn't be any error like odo not found ) by the `err` not occurring part

girishramnani: Hence its not contradicting @amitkrout :-)

tests/e2e/cmp_test.go
amitkrout: As pr @girishramnani comments https://github.com/redhat-developer/odo/pull/1182/files#r248625658 and https://github.com/redhat-developer/odo/pull/1182/files#r248626040 the above comment will not resolve the failure i guess

cdrage:
I'm going to close this for now since @amitkrout will be taking over

amitkrout:
#1205

PR #1211, opened 2019-01-17
Adds color support for Windows

cdrage:
This fixes the issue of ASCII characters not appearing correctly for
Windows users.

Changes with this PR:
- Odo will detect if Windows is used and use a non-Unicode spinner.
- Odo will also use "X" and "V" to indicate success and unsuccessful on Windows since "✓" is Unicode.

Closes issue: https://github.com/redhat-developer/odo/issues/1072


cdrage:
I'm unable to test this against Windows due to `minishift` issues running in a nest-virtualization environment. If someone can, that'd be great!

cdrage:
This is ready for a quick test / review. If someone on a Windows machine could test this, that'd be great. I'm having issues setting up Minishift in a nested virtualization environment so it'll have to wait for me to test this locally.

geoand:
I don't have a Windows machine unfortunately 😣

pkg/log/status.go
```diff
@@ -26,9 +26,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/fatih/color"
```
metacosm: Can we remove this lib, then?

cdrage: We still need this for Linux color support.

vendor/gopkg.in/AlecAivazis/survey.v1/terminal/LICENSE.txt
```diff
@@ -1,22 +1,22 @@
...
-included in all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
-OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
-WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+Copyright (c) 2014 Takashi Kokubun
```
anmolbabu: by the way what is the change in this file?
Github seems to show this as a modified file

cdrage: This was done after the re-vendor.

cdrage:
@anmolbabu Funny enough, the use of two libraries isn't a problem. Ironically, we already vendor both (hence why glide update only changed the licensing file...) since we import the OpenShift library!

girishramnani:
so the tests are failing but these are my tests on windows 

### Git shell on windows 

- component create- OK
image

@cdrage the colors look fine on git shell on windows

### windows CMD

- component create - ISSUE
image

still the error persists

### Powershell

- component create - ISSUE
image

@cdrage the color now seems to work but the cross and tick characters are not being rendered correctly and ascii code is also being shown

odo used for the latest artifacts of this PR from apveyor
- odo version 
```
odo v0.0.18 (e1ccaa49)

Server: https://api.engint.openshift.com:443
OpenShift: v3.11.43
Kubernetes: v1.11.0+d4cacc0

```

cdrage:
@girishramnani Thanks! I was only able to test this on command-prompt at the time (SSH'ing into a Windows machine). But this helps immensely. I'm going to figure out why it won't output it correctly..

kadel:
@cdrage you can download official  Windows 10 for dev and testing for free - https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

cdrage:
Hey @girishramnani and @kadel 

I made these changes to the PR:
- Odo will detect if Windows is used and use a non-Unicode spinner.
- Odo will also use "X" and "V" to indicate success and unsuccessful on Windows since "✓" is Unicode.

Unfortunately we have to use ASCII characters to indicate success and unsuccessful on Windows since CMD and Powershell don't suppose Unicode (most of the time.... unless the user has it enabled / installed).

I couldn't find any other version of success (using "V") at the moment that would be a good indicator: http://www.asciitable.com

I've seen "V" used before in other programs wayyyyy back when I was in highschool, but it's tricky to find examples.

If you guys have any other suggestions let me know.

Here is proof that it works now in CMD + PowerShell and outputs the colors correctly and Non-Unicode:

windows

cdrage:
There are three TODO notes with regards to adding tests. Unfortunately this is Windows-specific and I'm unable to write them until we get our Windows CI up-and-running. However... I believe the functions are simple enough to be without tests.

cdrage:
@girishramnani @surajnarwade @syamgk 

Could I get a review please? Thanks!

kadel:
I can see that it is still broken  :-(

screenshot 2019-03-07 at 09 37 33

@cdrage On your screenshot, I can see that you are only testing error messages. That is not a good verification that it is working ok. You have to verify this on the full workflow!

cdrage:
So I left a few comments, however, I've fixed the previous issue that @kadel  was encountering.

Unfortunately there's current no direct method to determine if Odo is being ran within a terminal or not within Windows. Although it's safe to assume that it is *always* in a terminal considering you're required to use one to run `odo.exe` (unlike Linux behaviour..)

@girishramnani @surajnarwade @syamgk

Feel free to try this again! Spinners + colour + output will all work.

kadel:
> Unfortunately there's current no direct method to determine if Odo is being ran within a terminal or not within Windows. Although it's safe to assume that it is _always_ in a terminal considering you're required to use one to run `odo.exe` (unlike Linux behaviour..)


I'm a little bit confused here.
What do you mean by "is being run within a terminal or within Windows"? You can easily detect if it is run on Windows or not.

kadel:
@cdrage can you please just add new commits to PR instead of amending to the old one? It makes reviewing much harder :-(

cdrage:
@kadel Yeah, you can easily detect if it's running in Windows. But you cannot detect if the user double-clicked "odo.exe" or if they are using it in the terminal. Similar to how we detect if odo is being used in terminal vs piped into another command or `odo create nodejs > foobar.txt` and thus *not* use the spinner (or else it'd spam the spinner output in the log file)

kadel:
> @kadel Yeah, you can easily detect if it's running in Windows. But you cannot detect if the user double-clicked "odo.exe" or if they are using it in the terminal.

I wouldn't worry about that right now.

> EDIT: It doesn't affect anything, it's just something that I noticed. It's assumed that whenever odo.exe is ran within Windows that it'll always be within a terminal window.

👍 I wouldn't consider anything else as a valid use case. It is either execute by user in terminal, or by other tool (like vscode)

pkg/log/status.go
```diff
@@ -147,57 +154,57 @@ func (s *Status) End(success bool) {
...
 	green := color.New(color.FgGreen).SprintFunc()
-	fmt.Printf("%s%s%s%s\n", prefixSpacing, green("✓"), suffixSpacing, fmt.Sprintf(format, a...))
+	fmt.Fprintf(GetStdout(), "%s%s%s%s\n", prefixSpacing, green(getSuccessString()), suffixSpacing, fmt.Sprintf(format, a...))
 }
 
 // Errorf will output in an appropriate "progress" manner
 func Errorf(format string, a ...interface{}) {
 	red := color.New(color.FgRed).SprintFunc()
-	fmt.Fprintf(os.Stderr, " %s%s%s\n", red("✗"), suffixSpacing, fmt.Sprintf(format, a...))
+	fmt.Fprintf(GetStdout(), " %s%s%s\n", red(getErrString()), suffixSpacing, fmt.Sprintf(format, a...))
```
mohammedzee1000: Would this not mean that error would not go to stderr but into stdout?

cdrage: Good catch, you're right. I'll fix that.

kadel:
/hold

remove the hold once @mohammedzee1000  comment is addressed

cdrage:
/hold cancel

@mohammedzee1000 's comment has been addressed.

PR #1215, opened 2019-01-18
Remove GetServiceClass

metacosm:
Remove unused and functionality duplicating function.


pkg/occlient/occlient.go
```diff
@@ -2117,26 +2117,13 @@ func (c *Client) patchDCOfComponent(componentName, applicationName string, dcPat
 	return nil
 }
 
-// Service struct holds the servicename and it's corresponding list of plans
+// Service struct holds the service name and its corresponding list of plans
```
mik-dass: Today I learned possessive noun :D Thanks :+1:

PR #1223, opened 2019-01-23
use Recreate deployment strategy

### LINKED ISSUE #1141: Change deployment strategy to Recreate
[kind/Enhancement]
We need to provide a way to configure odo's generated deployments to have a nodeselector (or similar) so that they are always scheduled to the same node, or else, to configure pods with `Recreate` strategy instead of `Rollout` so that they can work in cloud deployments where only block storage is available.

## Why is this needed? 
Any action that requires a redeployment (e.g. link, config,... ) can potentially leave the pod in `Pending` state. 

This can easily be tested in OpenShift online.

kadel:
odo will generate DeplymentConfig with Recreate strategy instead of Rollout. 
More info an be found at #1141

fixes #1141


surajnarwade:
merging based on 2 acks and :green_heart:  CI  :tada:

PR #1227, opened 2019-01-23
Remove 'time' from watch tests

cdrage:
Removes 'time' from the watch tests as these tests fail locally with zsh
terminals.

Closes issue https://github.com/redhat-developer/odo/issues/1226


geoand:
Any idea why the tests are failing?

girishramnani:
Maybe this is the reason https://unix.stackexchange.com/questions/478216/zsh-time-doesnt-work-with-source

cdrage:
@girishramnani I think we should remove it for now and then add the Golang timed-based statistics after

PR #1234, opened 2019-01-26
Fix improper version refactoring missed during review: Run takes no arguments.

metacosm:
See $SUBJECT.

No

## How to test changes?
No externally visible changes.

### DIFF
--- a/pkg/odo/cli/version/version.go
+++ b/pkg/odo/cli/version/version.go
@@ -2,6 +2,7 @@ package version
 
 import (
 	"fmt"
+	"github.com/redhat-developer/odo/pkg/occlient"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/odo/cli/version/version.go
```diff
@@ -46,6 +49,13 @@ func NewVersionOptions() *VersionOptions {
 
 // Complete completes VersionOptions after they have been created
 func (o *VersionOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	if !o.clientFlag {
+		// Let's fetch the info about the server
+		o.serverInfo, err = genericclioptions.ClientWithConnectionCheck(cmd, true).GetServerVersion()
```
geoand: Won't this fail if the connection to the server can't be made?

With the code in `master`, `odo version` does not fail when the connection to the server cannot be made. It just doesn't print the server info

metacosm: Hmm, I didn't think of that, let me check.

PR #1247, opened 2019-01-29
Update build version to 0.0.18

cdrage:


cdrage:
Shouldn't of this been merged before your homebrew release @kadel ? Are we able to merge this now?

kadel:
> Shouldn't of this been merged before your homebrew release @kadel ? Are we able to merge this now?

homebrew has not been merged yet. It shouldn't have any effect on it, as this file is only used by update checker in odo. Or at least I think that it is used only there.

Issue #1248, opened 2019-01-29
odo version fails if  ~/.kube/config doesn't exist


kadel:
[kind/bug]

## What versions of software are you using?
- Operating System:
- Output of `odo version`: error :-(
(running from current master 31e898e60ca8a7b99c04122d16a70ddc78c688e5)

## How did you run odo exactly?
`rm -r ~/.kube/config`
`odo version`

## Actual behavior
```
 ✗  invalid configuration: no configuration has been provided
Please login to your server:

odo login https://mycluster.mydomain.com
```

## Expected behavior
show version

PR #1249, opened 2019-01-29
use 'odo version --client' in sh scripts

kadel:
'odo version --client' will show just odo version without attempting to
connect to a cluster

`odo version` is trying to check cluster version. Problem is that if there is no KUBE_CONFIG it fails with an ugly error (reported in #1248). 

`odo version` has `--client` flag that skips the cluster version and shows just odo version.
Odo version is all that our release and install scripts need.

workaround for #1248

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -99,7 +99,7 @@ install_odo() {
 
     if command_exists odo; then
         echo_stderr "# 
-odo version \"$(odo version)\" is already installed on your system, running this installer script might case issues with your current installation. If you want to install odo using this script, please remove the current installation of odo from you system.
+odo version \"$(odo version --client)\" is already installed on your system, running this installer script might case issues with your current installation. If you want to install odo using this script, please remove the current installation of odo from you system.
 Aborting now!
 "
         exit 1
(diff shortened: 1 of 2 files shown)


geoand:
@kadel shouldn't this be changed as well? 
https://github.com/redhat-developer/odo/blob/31e898e60ca8a7b99c04122d16a70ddc78c688e5/scripts/create-packages.sh#L17

kadel:
> @kadel shouldn't this be changed as well?
> [odo/scripts/create-packages.sh](https://github.com/redhat-developer/odo/blob/31e898e60ca8a7b99c04122d16a70ddc78c688e5/scripts/create-packages.sh#L17)
> 
> Line 17 in [31e898e](/redhat-developer/odo/commit/31e898e60ca8a7b99c04122d16a70ddc78c688e5)
> 
>  bin_version=$(${BIN_DIR}/linux-amd64/odo version | head -1 | sed "s/^odo \(.*\) (.*)$/\1/")

You are right, it should. Thank you for catching that

geoand:
> > @kadel shouldn't this be changed as well?
> > [odo/scripts/create-packages.sh](https://github.com/redhat-developer/odo/blob/31e898e60ca8a7b99c04122d16a70ddc78c688e5/scripts/create-packages.sh#L17)
> > Line 17 in [31e898e](/redhat-developer/odo/commit/31e898e60ca8a7b99c04122d16a70ddc78c688e5)
> > bin_version=$(${BIN_DIR}/linux-amd64/odo version | head -1 | sed "s/^odo (._) (._)$/\1/")
> 
> You are right, it should. Thank you for catching that

You're welcome :)

Issue #1203, opened 2019-01-14
services don't show as part of an application


jorgemoralespou:
[kind/bug]
`odo app describe` don't show services as part of an application:

```
$ odo list
ACTIVE     NAME        TYPE
*          backend     java

$ odo service list
NAME                       TYPE                       STATUS
sso72-mysql-persistent     sso72-mysql-persistent     ProvisionedAndBound

$ odo app describe
Application Name: wwgame has 1 components:
--------------------------------------
Component Name: backend
Type: java
Source: file:///Users/jmorales/repositories/jorgemoralespou/odo/WW-Game/Wild-West-Backend/target/wildwest-1.0.jar
--------------------------------------
```

PR #1252, opened 2019-01-30
Add services to odo app describe

geoand:
See $SUBJECT

Fixes: #1203

## How to test changes?
Create various component and/or services. 
Make sure that `odo app describe` shows the appropriate output

### DIFF
--- a/tests/e2e/link_test.go
+++ b/tests/e2e/link_test.go
@@ -63,7 +63,7 @@ var _ = Describe("odoLinkE2e", func() {
 			})
 		})
 
-		It("app describe should show the service", func() {
+		It("app describe should show the mysql service", func() {
 			describeOutput := runCmd("odo app describe")
 
 			// ensure that the output contains the service
--- a/tests/e2e/link_test.go
+++ b/tests/e2e/link_test.go
@@ -63,6 +63,13 @@ var _ = Describe("odoLinkE2e", func() {
 			})
 		})
 
+		It("app describe should show the service", func() {
+			describeOutput := runCmd("odo app describe")
+
+			// ensure that the output contains the service
+			Expect(describeOutput).To(ContainSubstring("mysql-persistent"))
+		})
+
 		It("should link backend to service", func() {
 			runCmd("odo link mysql-persistent -w --component backend")
 
--- a/pkg/odo/cli/application/application.go
+++ b/pkg/odo/cli/application/application.go
@@ -2,6 +2,7 @@ package application
 
 import (
 	"fmt"
+	"github.com/redhat-developer/odo/pkg/service"
 	"os"
 	"strings"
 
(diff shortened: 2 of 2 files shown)


tests/e2e/link_test.go
```diff
@@ -63,6 +63,13 @@ var _ = Describe("odoLinkE2e", func() {
 			})
 		})
 
+		It("app describe should show the mysql service", func() {
```
amitkrout: LGTM. UTs won’t be a good idea here. Only user facing scenario (Integration test) would be enough jot verify a description change. 

However i want to test the pr just to see the output format

geoand:
> @amitkrout That is a very good catch!
> 
> I don't know what the actual problem is, but it's unrelated to this PR - you should see the same output if you test `master`.
> I propose that you open a new issue, assign it to me and I will check it out as part of a new PR. Does that sound good?

Actually @amitkrout , hold on a sec, let check something

geoand:
> > Actually @amitkrout , hold on a sec, let check something
> 
> Sure

Can you post the steps you followed to reproduce the problem?
I can't seem to reproduce it either with the PR or with master

amitkrout:
@geoand Steps to reproduce https://github.com/redhat-developer/odo/pull/1252#pullrequestreview-198048365

1. Download minishift v1.30.1
2. start minishift
```
$ MINISHIFT_ENABLE_EXPERIMENTAL=y ./minishift start --extra-clusterup-flags "--enable=*,service-catalog,automation-service-broker,template-service-broker"
```
3. Fetch the pr and run make bin
4. ```./odo create nodejs frontend```
5. ```./odo create python backend```
6. ```./odo link backend --component frontend```
7. ```./odo service create mysql-persistent```
8. ```./odo link mysql-persistent -w --component backend```
9. ```./minishift console```
10. Create the external route for both ```frontend``` and ```backend```
11. ```./odo app describe```

Please correct me if i diverted the usecase in a wrong way

geoand:
> @geoand Steps to reproduce [#1252 (review)](https://github.com/redhat-developer/odo/pull/1252#pullrequestreview-198048365) comment
> 
> 1. Download minishift v1.30.0
> 2. start minishift
> 
> ```
> $ MINISHIFT_ENABLE_EXPERIMENTAL=y ./minishift start --extra-clusterup-flags "--enable=*,service-catalog,automation-service-broker,template-service-broker"
> ```
> 1. Fetch the pr and run make bin
> 2. `./odo create nodejs frontend`
> 3. `./odo create python backend`
> 4. `./odo link backend --component frontend`
> 5. `./odo service create mysql-persistent`
> 6. `./odo link mysql-persistent -w --component backend`
> 7. `./minishift console`
> 8. Create the external route for both `frontend` and `backend`
> 9. `./odo app describe`
> 
> Please correct me if i diverted the usecase in a wrong way

OK, so I see what the problem is.
You need to create the URLs using  `odo url create` instead of doing it from the console.
Can you please check?

amitkrout:
> You need to create the URLs using odo url create instead of doing it from the console.
Can you please check?

Right, This is the issue. 
However manually i created external route to use port 8080 and i think this should be captured by ```./odo app describe```

screen shot 2019-01-30 at 7 17 18 pm

geoand:
> > You need to create the URLs using odo url create instead of doing it from the console.
> > Can you please check?
> 
> Right, This is the issue.
> However manually i created external route to use port 8080 and i think this should be captured by `./odo app describe`
> 
> screen shot 2019-01-30 at 7 17 18 pm

That makes sense (although I don't it's easy since we use specific labels that aren't added by the console).
Can you open another issue for this and ping Jorge to get his opionion?

Thanks @amitkrout

amitkrout:
> Can you open another issue for this and ping Jorge to get his opionion?

Sure, will do that

geoand:
> LGTM
> Will discuss the observation [#1252 (comment)](https://github.com/redhat-developer/odo/pull/1252#issuecomment-458950426) in a separate issue.

Thanks!

kadel:
> @geoand it looks like that the `0` is shown in case of nodejs because there is not external port mapped. So for that we shouldn't show `exposed via 0` to the user. I would suggest to resolve that in this PR itself if that is OK. This could be confusing for the user.

If I understand this correctly, this issue is in the current master, this is not something that was introduced by this PR. If that is true then it shouldn't be resolved in this PR. Let's open a new issue to report this as a bug, and we can take care of it in separate PR.

geoand:
> > @geoand it looks like that the `0` is shown in case of nodejs because there is not external port mapped. So for that we shouldn't show `exposed via 0` to the user. I would suggest to resolve that in this PR itself if that is OK. This could be confusing for the user.
> 
> If I understand this correctly, this issue is in the current master, this is not something that was introduced by this PR. If that is true then it shouldn't be resolved in this PR. Let's open a new issue to report this as a bug, and we can take care of it in separate PR.

This only happens when the URL is created **outside** of odo, so it has nothing to do with PR and I would argue that it probably doesn't even need to be handled by odo in general

metacosm:
I move that we should merge this PR, then.

kadel:
> This only happens when the URL is created **outside** of odo, so it has nothing to do with PR and I would argue that it probably doesn't even need to be handled by odo in general

I think that we can detect port from Route even if it is not created by odo. But it is not a priority right now. This is why I would like to create an Issue for it and then we can address this later, or decide that we don't want to do that. 

Let's discuss this elsewhere.

geoand:
> > This only happens when the URL is created **outside** of odo, so it has nothing to do with PR and I would argue that it probably doesn't even need to be handled by odo in general
> 
> I think that we can detect port from Route even if it is not created by odo. But it is not a priority right now. This is why I would like to create an Issue for it and then we can address this later, or decide that we don't want to do that.
> 
> Let's discuss this elsewhere.

@amitkrout Have you already created the issue?

amitkrout:
> @amitkrout Have you already created the issue?

Not yet, busy with some other stuff. @geoand Please create the issue, Later i will put the test content if needed

girishramnani:
ok LGTM

amitkrout:
@geoand Created https://github.com/redhat-developer/odo/issues/1256

geoand:
Thanks @amitkrout ! Was going to do that in a while

Issue #1129, opened 2018-12-12
Use regular expression for string operation in testsuite


amitkrout:
[kind/Enhancement]

## Which functionality do you think we should update/improve?
string operation like sed, awk should be replaced with inbuilt go package ```re```.  

## Why is this needed? 

Just to make sure the testsuite should run on all the platforms.

amitkrout:
Test file to cover

- [ ]  tests/e2e/component.go

- [ ] tests/e2e/e2e_test.go

- [ ] tests/e2e/java_test.go

- [ ]  tests/e2e/service_test.go

- [ ]  tests/e2e/source_test.go

- [ ]  tests/e2e/watch_test.go

PR #1264, opened 2019-01-31
Refactor utils command to follow Complete/Validate/Run pattern

### LINKED ISSUE #1194: Refactor utils command to follow complete-validate-run pattern
more info in #1142

metacosm:
fixes #1194 

## How to test changes?
Play with `odo utils` commands.


pkg/config/config.go
```diff
@@ -395,3 +398,43 @@ func (c *ConfigInfo) DeleteProject(projectName string) error {
 	}
 	return nil
 }
+
+var (
+	// records information on supported parameters
+	supportedParameterDescriptions = map[string]string{
+		"UpdateNotification": "Controls if an update notification is shown or not (true or false)",
```
anmolbabu: Can this be configurables be string constants?
I see them being referred in more than 1 places which might be prone to bugs
So, this then converts to map[string_alias]string

pkg/config/config.go
```diff
@@ -141,29 +142,31 @@ func (c *ConfigInfo) writeToFile() error {
 // SetConfiguration modifies Odo configurations in the config file
 // as of now being used for nameprefix, timeout, updatenotification
 func (c *ConfigInfo) SetConfiguration(parameter string, value string) error {
-	// processing values according to the parameter names
-	switch parameter {
+	if p, ok := asSupportedParameter(parameter); ok {
```
anmolbabu: I tend to feel config or configurable would be apt instead of parameter?

anmolbabu: but from what i see in git diff view, its been like that for some time now in which case please ignore my comment

metacosm: I actually have more ideas on how to refactor things here but I don't think it was appropriate to do these changes in this PR (I've deleted code that I spent a couple of hours on with some of these changes I had in mind and decided against pushing further).

pkg/config/config.go
metacosm: Let me see what I can do.

metacosm: Extracted constants.

tests/e2e/e2e_test.go
```diff
@@ -80,22 +80,24 @@ var _ = Describe("odoe2e", func() {
...
 			configOutput := runCmd("odo utils config view")
 			Expect(configOutput).To(ContainSubstring("true"))
-			Expect(configOutput).To(ContainSubstring("UpdateNotification"))
+			Expect(configOutput).To(ContainSubstring(config.UpdateNotificationSetting))
+			Expect(configOutput).To(ContainSubstring(config.NamePrefixSetting))
+			Expect(configOutput).To(ContainSubstring(config.TimeoutSetting))
 		})
 		It("should be checking to see if timeout is the same as the constant", func() {
-			configOutput := runCmd("odo utils config view|grep Timeout")
+			configOutput := runCmd("odo utils config view|grep " + config.TimeoutSetting)
```
amitkrout: Avoid using terminal specific command like ```grep```. regular expression would be an alternate choice and can be run on any terminal.

metacosm: I didn't create these tests, they've been like this for a while, not going to change them now.

amitkrout: Ok, will track this change in #1129 then.

metacosm:
> Major command `odo utils config set ...` is missed as part of this pr. Would you please add scenario for `odo utils config set ...` in integration test too.
> 
> ```
> $ ./odo utils -h
> Utilities for terminal commands and modifying Odo configurations
> 
> Usage:
>   odo utils [command]
> 
> […]
> 
>    # Set a configuration value
>    odo utils config set UpdateNotification false
>    odo utils config set NamePrefix "app"
>    odo utils config set Timeout 20
> 	
>   # For viewing the current configuration
>    odo utils config view
>   
> […]
> ```

Actually, `set` is covered: https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L92-L99

PR #1274, opened 2019-02-04
Doc: create instances from any service class

sbose78:
Fix in-line/cmd docs to reflect that we can add any service that is visible in `svcat get classes -o=json` ( ie, any ClusterServiceClass ), and not just those from the Template Service Broker.

### DIFF
--- a/pkg/odo/cli/service/service.go
+++ b/pkg/odo/cli/service/service.go
@@ -14,7 +14,7 @@ import (
 // RecommendedCommandName is the recommended service command name
 const RecommendedCommandName = "service"
 
-var serviceLongDesc = ktemplates.LongDesc(`Perform service catalog operations, limited to template service broker and OpenShift Ansible Broker only.`)
+var serviceLongDesc = ktemplates.LongDesc(`Perform service catalog operations`)
 
 // NewCmdService implements the odo service command
 func NewCmdService(name, fullName string) *cobra.Command {


geoand:
Currently we have only tested with Template Service Broker and Openshift Ansible Broker (aka Ansible Service Broker, Automation Broker) which is why the text is worded that way.
That's not to say that we don't expect other brokers to work as well, but we would like to be on the safe side in terms of what we claim

kadel:
Isn't this a little bit confusing. 
Odo user will don't even know what Service Broker is, and he/she doesn't care about it.

`odo catalog list services` will list all ClusterServiceClasses, but our help says that we support only two service brokers?

geoand:
> Isn't this a little bit confusing.
> Odo user will don't even know what Service Broker is, and he/she doesn't care about it.
> 
> `odo catalog list services` will list all ClusterServiceClasses, but our help says that we support only two service brokers?

Indeed. Let's just change it and if someone complains later, we can see what happens with other brokers

amitkrout:
Thanks for the pr. 

**It would be great if add the steps to reproduce the issue and what changes (preferable the output) after applying this patch**

@sbose78 FYI, as per our pr merge protocol UTs coverage should not be decreased with any fresh patch. Good to see that the coverage in the ```pkg/odo/cli/service/service.go``` looks fine (unaffected) for a message change. However in rest packages there are some decreases in coverage resulting final decrease of 0.12% which @sbose78 should not worry about IMO.

@kadel If this train of decreasing coverage continues then it will be very hard to achieve the code quality in later stage. So if we dedicate few capacity to write UTs for a complete spring in very recent future could be helpful IMO. WDYT ?

geoand:
> Thanks for the pr.
> 
> **It would be great if you add the steps to reproduce the issue and what changes (preferable the output) after applying this patch**
> 
> @sbose78 FYI, as per our pr merge protocol UTs coverage should not be decreased with any fresh patch. Good to see that the coverage in the `pkg/odo/cli/service/service.go` looks fine (unaffected) for a message change. However in rest packages there are some decreases in coverage resulting final decrease of 0.12% which @sbose78 should not worry about IMO.
> 
> @kadel If this train of decreasing coverage continues then it will be very hard to achieve the code quality in later stage. So if we dedicate few capacity to write UTs for a complete sprint in very recent future could be helpful IMO. WDYT ?

The fact that test coverage changes in a PR that is simply a doc change should be worrying IMHO....

amitkrout:
@geoand Right, for doc changes ```pkg/odo/cli/service/service.go``` coverage won't be affected.

geoand:
> @geoand Right, for doc changes `pkg/odo/cli/service/service.go` coverage won't be affected.

Yes, but the coverage is indicated to be decreasing in the PR... Which to me suggest that the metric is totally flawed

amitkrout:
> but the coverage is indicated to be decreasing in the PR... Which to me suggest that the metric is totally flawed

Absolutely right, This need to be investigated :)

amitkrout:
screen shot 2019-02-05 at 1 17 14 pm

File ```pkg/odo/cli/service/service.go``` is modified for a doc change in this pr. Coverage looks unaffected as per the image which is obvious, but no idea why some other files are getting involved in coverage matrix.

amitkrout:
In most of the cases i find ```pkg/component/watch.go``` is the culprit file...haha...

geoand:
> In most of the cases i find `pkg/component/watch.go` is the culprit file...haha...

I don't even look at the coverage anymore since I know it's faulty :)

PR #1276, opened 2019-02-04
Update imported packages

cdrage:
Updates imported packages according to:

```sh
github.com/redhat-developer/odo  update-go-pkg ✔                                                                                                                              3h37m
▶ goimports -w pkg
```

This avoids conflicts / constant rebasing for users using `goimports`.

### DIFF
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -3,16 +3,15 @@ package occlient
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/redhat-developer/odo/pkg/testingutil"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/pkg/errors"
-
 	applabels "github.com/redhat-developer/odo/pkg/application/labels"
 	componentlabels "github.com/redhat-developer/odo/pkg/component/labels"
+	"github.com/redhat-developer/odo/pkg/testingutil"
 	"github.com/redhat-developer/odo/pkg/util"
 
 	// api resources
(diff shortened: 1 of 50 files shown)


pkg/occlient/occlient_test.go
```diff
@@ -749,7 +750,7 @@ func TestCreateRoute(t *testing.T) {
 			service:    "mailserver",
 			portNumber: intstr.FromInt(8080),
 			labels: map[string]string{
-				"SLA":                              "High",
```
metacosm: Isn't that going against what `go fmt` prescribes?

cdrage: Yeah, I just realized that, it seems that `goimports` isn't following standard gofmt conventions...

cdrage:
@metacosm Issues have been addressed, I updated the `goimports` too and it works correctly now.

pkg/occlient/occlient_test.go
```diff
@@ -3,12 +3,13 @@ package occlient
...
 	"encoding/json"
 	"fmt"
-	"github.com/redhat-developer/odo/pkg/testingutil"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/redhat-developer/odo/pkg/testingutil"
+
```
kadel: why is this here? should it be grouped together wiht 
```
applabels "github.com/redhat-developer/odo/pkg/application/labels"
componentlabels "github.com/redhat-developer/odo/pkg/component/labels"
"github.com/redhat-developer/odo/pkg/util"
```

cdrage: No idea to be honest. Just what `gofmt` does. If I modify it manually it should work :+1:

cdrage:
@kadel updated with your suggested changes, should be good now :+1:

PR #1279, opened 2019-02-05
Fixing #1278: Changing cleanup to run as needed

### LINKED ISSUE #1278: login e2e tests are incorrect / dangerous
login e2e tests currently attempt to delete all projects. This might not work and is quite dangerous if the tests are run by an unsuspecting user on a cluster with elevated rights. Moreover, it might be difficult to ensure that the state is appropriate (i.e. no projects) when the tests are launched so end to end testing of the different login scenarios might be difficult/impossible.

mohammedzee1000:
See title
 - Project deletion does not happen now when it is not needed
 - Project deletion, when run only targets specific projects

fixes #1278 

### DIFF
--- a/tests/e2e/login_test.go
+++ b/tests/e2e/login_test.go
@@ -8,88 +8,74 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const baseOdoProjectDelete = "odo project delete -f "
-
 var _ = Describe("odoLoginE2e", func() {
 	// user related constants
-	const loginTestUser = "testdeveloper"
-	const loginTestUserPassword = "testdeveloper"
-	const odoTestProject1 = "testproject1"
-
-	// Comand related constants
-	const baseOdoLoginCommand = "odo login"
-	const baseOdoProjectCreate = "odo project create"
-	const ocWhoamiCommand = "oc whoami"
-	const ocTokenCommand = "oc whoami -t"
+	const loginTestBaseUser = "testdeveloper"
+	const loginTestUserPassword = "developer"
 
 	// variables to be used in test
 	var session string
-	var backToCurrentUserCommand string
-	var testUserLoginCommand string
-	var testUserLoginCommandWithToken string
-	var testUserLoginFailCommandWithToken string
-	var testUserCreateProject1Command string
+	var testUserToken string
+	var testUser string
 
 	Describe("Check for successful login and logout", func() {
 		Context("Initialize", func() {
 			It("Should initialize some variables", func() {
-				// Save currently logged in users token, so we can get back to that context after being done
-				t := runCmd(ocTokenCommand)
-				backToCurrentUserCommand = fmt.Sprintf("%s -t %s", baseOdoLoginCommand, t)
-				testUserLoginCommand = fmt.Sprintf("%s -u %s -p %s", baseOdoLoginCommand, loginTestUser, loginTestUserPassword)
-				testUserCreateProject1Command = fmt.Sprintf("%s %s", baseOdoProjectCreate, odoTestProject1)
-				testUserLoginFailCommandWithToken = fmt.Sprintf("%s -t verybadtoken", baseOdoLoginCommand)
+				// Logout of current user to ensure state
+				runCmd("oc logout")
 			})
 		})
 
 		Context("Run login tests with no active projects, having default is also considered as not having active project", func() {
 			AfterEach(func() {
-				runCmd(backToCurrentUserCommand)
+				// Log out of whoever is logged in
+				runCmd("oc logout")
 			})
 
 			It("Should login successfully with username and password without any projects with appropriate message", func() {
-				session = runCmd(testUserLoginCommand)
+				session = runCmd(fmt.Sprintf("odo login -u %s -p %s", loginTestBaseUser, loginTestUserPassword))
 				Expect(session).To(ContainSubstring("Login successful"))
 				Expect(session).To(ContainSubstring("You don't have any projects. You can try to create a new project, by running"))
 				Expect(session).To(ContainSubstring("odo project create <project-name>"))
-				session = runCmd(ocWhoamiCommand)
-				Expect(session).To(ContainSubstring(loginTestUser))
-				token := runCmd(ocTokenCommand)
+				session = runCmd("oc whoami")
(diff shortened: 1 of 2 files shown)


tests/e2e/login_test.go
```diff
@@ -17,7 +19,6 @@ var _ = Describe("odoLoginE2e", func() {
 	// Comand related constants
 	const baseOdoLoginCommand = "odo login"
 	const baseOdoProjectCreate = "odo project create"
-	const baseOdoProjectDelete = "odo project delete -f "
 	const ocWhoamiCommand = "oc whoami"
```
metacosm: I feel the tests would read better without these constants and variables (i.e. if the commands were inlined in the code directly).

tests/e2e/login_test.go
```diff
@@ -41,18 +42,8 @@ var _ = Describe("odoLoginE2e", func() {
...
-				var waitOut bool
-				for _, p := range projects {
-					if len(p) > 0 {
-						waitOut = waitForCmdOut(fmt.Sprintf("%s %s", baseOdoProjectDelete, p), 10, func(out string) bool {
-							return strings.Contains(out, fmt.Sprintf("Deleted project : %s", p))
-						})
-						Expect(waitOut).To(BeTrue())
-					}
-				}
 				runCmd(backToCurrentUserCommand)
```
metacosm: What happens if no user was initially logged in? Wouldn't it be better to log the current user out? (Not sure, just asking)

mohammedzee1000: BeforeSuite actually ensures that.  It gets current project so a user has to be logged in, with an active project Infact

tests/e2e/login_test.go
mohammedzee1000: I can remove them no problem.

I did that because the same command was being used in multiple places and it would make it easier to update the command,  should anything change.

metacosm: Yep, I understand but I think it'd be better to be more explicit in this case… :)

tests/e2e/login_test.go
metacosm: Tests shouldn't really assume an initial state, otherwise they are not independent… but yes, it's probably better to leave the state how the tests found it so maybe, remember the previous user to be able revert to it but do a logout after that, so that the login tests can work without a user being logged in?

tests/e2e/login_test.go
mohammedzee1000: Updated locally, pushing shortly

tests/e2e/login_test.go
mohammedzee1000: Will do

tests/e2e/login_test.go
```diff
@@ -10,89 +10,83 @@ import (
...
+				session = runCmd("oc whoami")
+				Expect(session).To(ContainSubstring(loginTestBaseUser))
 			})
 
 			It("Should fail login on invalid token with appropriate message", func() {
-				session = runFailCmd(testUserLoginFailCommandWithToken, 1)
+				session = runFailCmd("odo login -t verybadtoken", 1)
 				Expect(session).To(ContainSubstring("The token provided is invalid or expired"))
-				runCmd(testUserLoginCommand)
+				runCmd(fmt.Sprintf("oc login -t %s", testUserToken))
```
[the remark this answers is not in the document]
mohammedzee1000: The actual tests checks for invalid login (with invalid token), and cleanup always does logout, this is to ensure a user is logged in and logout does not fail

tests/e2e/login_test.go
```diff
@@ -59,28 +60,33 @@ var _ = Describe("odoLoginE2e", func() {
 			})
 		})
 
-		Context("Run login tests with single active project", func() {
+		Context("Run login tests with single active project with username and password", func() {
 			AfterEach(func() {
-				runCmd("oc logout")
+				cleanUpAfterProjects([]string{odoTestProjectForSingleProject1})
```
metacosm: 👍

mohammedzee1000:
Rebased on master

tests/e2e/login_test.go
```diff
@@ -10,89 +10,89 @@ import (
...
-				testUserLoginCommand = fmt.Sprintf("%s -u %s -p %s", baseOdoLoginCommand, loginTestUser, loginTestUserPassword)
-				testUserCreateProject1Command = fmt.Sprintf("%s %s", baseOdoProjectCreate, odoTestProject1)
-				testUserLoginFailCommandWithToken = fmt.Sprintf("%s -t verybadtoken", baseOdoLoginCommand)
+				// Logout of current user to ensure state
+				runCmd("oc logout")
 			})
 		})
 
-		Context("Run login tests", func() {
+		Context("Run login tests with no active projects, having default is also considered as not having active project", func() {
```
geoand: Great test name!

mohammedzee1000:
> Looks like logging out after each test doesn't work properly… :(

Issues due to overlap of deletion are fixed and this should take care of the above failures. Waiting for CI to go green :)

tests/e2e/login_test.go
```diff
@@ -8,91 +8,96 @@ import (
...
+				Expect(session).To(ContainSubstring("Login successful"))
+				Expect(session).To(ContainSubstring(odoTestProjectForSingleProject1))
+				session = runCmd("oc whoami")
+				Expect(session).To(ContainSubstring(loginTestUserForSingleProject1))
 			})
 		})
 	})
 })
+
+func cleanUpAfterProjects(projects []string) {
```
amitkrout: Can you keep function ```cleanUpAfterProjects``` in e2e.go file just to make sure the consistency of the test structure.

tests/e2e/login_test.go
```diff
@@ -8,91 +8,96 @@ import (
...
+	for _, p := range projects {
+		deleteProject(p)
+	}
+	// log back in as original user
+	runCmd(fmt.Sprintf("oc login --token %s", currentUserToken))
+}
+
+func deleteProject(project string) {
```
amitkrout: Same as above comment

tests/e2e/login_test.go
mohammedzee1000: Done

tests/e2e/login_test.go
mohammedzee1000: Done

PR #1282, opened 2019-02-05
Fix #779: move component commands under component.

### LINKED ISSUE #779: Command language consistency?
Right now, `component` actions are triggered directly (`odo create`, `odo watch`) and there is some value to this model as components could be considered the core entity that `odo` deals with. However, since other commands follow the `odo <entity> <action> <options>` language, this makes for an inconsistent UI and reduces learnability/discoverability when compared to a UI where components actions would also follow the `odo component create` grammar.
Wouldn't it make sense to unify the grammar for all commands?

metacosm:
Make `component` commands sub-commands of `component` command! ^_^
Top-level commands are kept for compatibility reason since I don't think there's an obvious way to create aliases to sub-commands.

fixes #779 

## How to test changes?
Check that `odo component <command>` works as expected, i.e. exactly as `odo <command>`.


metacosm:
There are no aliases, the original commands are still wired so there really are no changes there.

PR #1285, opened 2019-02-05
Clean-up after command refactoring

metacosm:
Fixes:
- documentation issues
- missed global variable usage
- use command name constants where possible

No.

## How to test changes?
Everything should still work as expected. Only change that could have an impact is on `storage mount` where global variable use was replaced by options field.

### DIFF
--- a/pkg/odo/cli/project/project.go
+++ b/pkg/odo/cli/project/project.go
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// RecommendedCommandName is the recommended service command name
+// RecommendedCommandName is the recommended project command name
 const RecommendedCommandName = "project"
 
 // NewCmdProject implements the project odo command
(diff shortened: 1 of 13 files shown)


pkg/odo/cli/project/project.go
```diff
@@ -35,7 +35,7 @@ func NewCmdProject(name, fullName string) *cobra.Command {
 		// 'odo project' is the same as 'odo project get'
 		// 'odo project <project_name>' is the same as 'odo project set <project_name>'
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) > 0 && args[0] != "get" && args[0] != "set" {
+			if len(args) > 0 && args[0] != getRecommendedCommandName && args[0] != setRecommendedCommandName {
```
geoand: Perhaps this should be changed to `RecommendedGetCommandName` like it's been done in `component.go`? I mean it doesn't need to be visible, but lets at least keep the same naming convention

metacosm: Actually, I will change the component names instead since all the other commands are using the `<command>RecommendedCommandName` format but I will do that in  #1282 instead.

geoand: :+1:

PR #1287, opened 2019-02-06
Do not use closed writer when pushing files to pod.

### LINKED ISSUE #1286: `E0205 14:23:32.599302   16744 v2.go:163] io: read/write on closed pipe` error when call `odo push` for local component
[kind/bug]

- Operating System: Windows 10 Pro 
- Output of `odo version`: 0.0.18

``` shell
$ minishift start
$git clone https://github.com/openshift-evangelists/Wild-West-Frontend
$cd Wild-West-Frontend
$odo create nodejs:latest wild --local  .\Wild-West-Frontend
$odo push
```

## Actual behavior
``` shell
Pushing changes to component: wild 
 •  Waiting for pod to start  ...
✓  Waiting for pod to start
•  Copying files to component  ...
E0205 14:23:32.599302   16744 v2.go:163] io: read/write on closed pipe
✓  Copying files to component
•  Building component  ...
✓  Building component
✓  Changes successfully pushed to component: wild 
```

## Expected behavior

``` shell
Pushing changes to component: wild 
 •  Waiting for pod to start  ...
✓  Waiting for pod to start
•  Copying files to component  ...
✓  Copying files to component
•  Building component  ...
✓  Building component
✓  Changes successfully pushed to component: wild 
```

N/A

Duplicate of #647 odo push failed.

dgolovin:
Get rid of `io: read/write on closed pipe` error.

fixes #1286.

## How to test changes?

See issue #1286. description.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -2739,7 +2739,7 @@ func (c *Client) CopyFile(localPath string, targetPodName string, targetPath str
 
 	// cmdArr will run inside container
 	cmdArr := []string{"tar", "xf", "-", "-C", targetPath, "--strip", "1"}
-	err := c.ExecCMDInContainer(targetPodName, cmdArr, writer, writer, reader, false)
+	err := c.ExecCMDInContainer(targetPodName, cmdArr, nil, nil, reader, false)
 	if err != nil {
 		return err
 	}


metacosm:
> I can't really comment on this so I leave it to others to verify :)

Same here… :(

pkg/occlient/occlient.go
```diff
@@ -2739,7 +2739,7 @@ func (c *Client) CopyFile(localPath string, targetPodName string, targetPath str
 
 	// cmdArr will run inside container
 	cmdArr := []string{"tar", "xf", "-", "-C", targetPath, "--strip", "1"}
-	err := c.ExecCMDInContainer(targetPodName, cmdArr, writer, writer, reader, false)
+	err := c.ExecCMDInContainer(targetPodName, cmdArr, bytes.NewBuffer([]byte{}), bytes.NewBuffer([]byte{}), reader, false)
```
anmolbabu: can we move the statement in line 2729 instead to may be outside the go func that is proably to the line after 2726
?

kadel: @anmolbabu You can't do that. If you move defer outside of the go routine into `CopyFile` you will end up in an infinite loop. 
`reader` and `writer` are connected with Pipe. If you move defer to the `CopyFile` it will close the writer side only when `CopyFile` function is finished, and this function will finish only after `ExecCMDInContainer` is done. `ExecCMDInContainer` reads from the reader side of the Pipe, and it will finish only after the writer side is closed. In this case it would be never :-)

kadel: What @dgolovin did is correct. `writer` shouldn't be used inside  `ExecCMDInContainer`.

Instead if passing empty buffer, we can just use `nil`, `ExecCMDInContainer` will handle that correctly.

kadel: Good catch @dgolovin

dgolovin: @kadel `nil` was also working, but I used [this line 206](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cp/cp.go#L206) as an example. should I update it to `nil`?

dgolovin: @kadel have no idea why  `login test` is failing though.

kadel: > @kadel `nil` was also working, but I used [this line 206](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cp/cp.go#L206) as an example. should I update it to `nil`?

They allocate a new buffer because it is used later on (it is returned from that function).
We never read that output, it is safe to use `nil` here, to clearly mark that that output is discarded. 


> @kadel have no idea why `login test` is failing though.

I see all test green now. I tried `odo login` and works. Maybe it was time timeout in tests :-(

dgolovin: I've updated PR to use `nil` instead of empty buffers.

geoand:
CI is very problematic today. Rebasing onto master probably won't help.
I'm hoping #1300 will fix the CI issues

geoand:
@kadel Have your comments been addressed? Should we merge this?

kadel:
> @kadel Have your comments been addressed? Should we merge this?

yes, this looks good

PR #1290, opened 2019-02-06
adding cico script and minishift bring up script

### LINKED ISSUE #952: Move e2e tests to ci.centos.org
We are starting to have quite a few issues with our e2e tests. They are slow and there is a lot of timeouts.

As a first step, let's move all our existing e2e tests from travis-ci to ci.centos.org.

syamgk:

Adding ci.centos jenkinsfile and script to bring up minishift cluster
fixes #952 

## How to test changes?

validate the scripts by compairing with the cico-test branch ones

This will currently do builds only on changes to master, since we are blocked on how to fetch the pr number from the webhook post on jenkins side and consume on jenkinsfile (.cico.pipeline)

### DIFF
--- a/.cico.pipeline
+++ b/.cico.pipeline
@@ -39,7 +39,7 @@ node('odo') {
                     onmyduffynode '''git clone --single-branch --branch master --depth 1 "https://github.com/redhat-developer/odo.git" ${CLONE_DIR} && 
                     cd  ${CLONE_DIR} && ls -la'''
                 }
-    
+   
                 stage('Build binary') {
                         onmyduffynode '''export PATH=$PATH:/usr/local/go/bin:/root/go/bin &&
                         go version &&
@@ -60,6 +60,14 @@ node('odo') {
                         make test |& tee unit_test.log'''
                 }
 
+                stage('Bringing up minishift') {
+                        onmyduffynode '''cd ${CLONE_DIR} &&
+                        echo "----------starting-cluster-provision-----------" &&
+                        ./scripts/minishift-setup.sh &&
+                        echo "----------finished-provisioning-cluster-----------"
+                        '''
+                }
+ 
                 timeout(20) {
                     stage('Main e2e Test') {
                         onmyduffynode '''export PATH=$PATH:/usr/local/go/bin &&
@@ -68,13 +76,9 @@ node('odo') {
                         cp odo /usr/local/go/bin &&
                         git branch && 
                         rpm --query centos-release &&
-                        echo "----------starting-cluster-provision-----------" &&
-                        ./scripts/minishift-setup.sh &&
-                        echo "----------finished-provisioning-cluster-----------" &&
                         odo version &&
                         oc whoami &&
-                        make test-main-e2e &&
-                        minishift delete -f 
+                        make test-main-e2e
                         '''
                     }
                 }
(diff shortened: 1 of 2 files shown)


.cico.pipeline
```diff
@@ -0,0 +1,200 @@
...
+                timeout(20) {
+                    stage('Main e2e Test') {
+                        onmyduffynode '''export PATH=$PATH:/usr/local/go/bin &&
+                        cd ${CLONE_DIR} &&
+                        make &&
+                        cp odo /usr/local/go/bin &&
+                        git branch && 
+                        rpm --query centos-release &&
+                        echo "----------starting-cluster-provision-----------" &&
+                        ./scripts/minishift-setup.sh &&
```
kadel: Wouldn't it be better to have to use the same OpenShift instance for all the test?
Right now we are creating and deleting cluster for each stage, and it takes a lot of time.

metacosm:
You might want to look into the CodeClimate issues as well: https://github.com/redhat-developer/odo/pull/1290#issuecomment-460951980

.cico.pipeline
syamgk: done!

cdrage:
@syamgk Do you have links to the current instance being built so we can check the logs / how long it takes each build?

amitkrout:
@cdrage https://github.com/redhat-developer/odo/issues/952#issuecomment-460186400

syamgk:
right now java-e2e tests are failing on ci.co, I'm investigating it
also, saw this pr from george https://github.com/redhat-developer/odo/pull/1300

so please wait for a while before reviewing this one

geoand:
> right now java-e2e tests are failing on ci.co, I'm investigating it
> also, saw this pr from @geoand #1300
> 
> so please wait for a while before reviewing this one

Yeah, any e2e test that relies on building a java app from source could be potentially slow due to the large number of dependencies involved

syamgk:
> > right now java-e2e tests are failing on ci.co, I'm investigating it
> > also, saw this pr from @geoand #1300
> > so please wait for a while before reviewing this one
> 
> Yeah, any e2e test that relies on building a java app from source could be potentially slow due to the large number of dependencies involved

thanks for the heads up!

syamgk:
Since we are planning to bring up the cluster once and run all the test suits on the same cluster.
And bringing up the cluster with `service-catalog` on ci.co.
There is one issue with one test in the `test-main-e2e`, which is for validating the scenario[1] if `service-catalog` doesn't exist on the cluster. So disabling that suite on ci.co for now.
so disabling this suite for now.

[1] https://github.com/redhat-developer/odo/pull/1280/files

kadel:
Lets slim this down a little bit.
Running all tests in series takes more than 1h. 

Merging this won't be much help. 
Right now we just need to verify that everything is working, for that we can run the only subset of the tests, we can comment out all e2e tests except one (maybe main e2e) to speed this up. 

Then we can figure out how to run tests in parallel against single OpenShift instance and uncomment the rest.

scripts/minishift-setup.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
+OPENSHIFT_CLIENT_BINARY_URL=${OPENSHIFT_CLIENT_BINARY_URL:-'https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz'}
```
amitkrout: The client binary itself bundle within minishift binary and can be available in ```MINISHIFT_HOME``` dir once you run ```minishift start```. Use the available binary instead of downloading it to avoid more time consumption for preparing the setup.

cdrage:
@syamgk Since @mohammedzee1000 is moving us to the Prow / OpenShift CI is it still viable to continue with this PR and push these CI changes?

kadel:
> @syamgk Since @mohammedzee1000 is moving us to the Prow / OpenShift CI is it still viable to continue with this PR and push these CI changes?

I don't think so. I think that we will close this PR.

PR #1293, opened 2019-02-06
Implement interactive mode for odo component create

### LINKED ISSUE #1272: Create interactive mode for odo create
[kind/Enhancement]

`odo create` is very commonly used and could benefit from an interactive mode.

## Why is this needed? 

A general consensus (see #1242) is that component create should be the command to have an interactive mode

geoand:
See $SUBJECT

fixes #1272 

## How to test changes?
Use `odo create` and make sure everything goes as expected


pkg/odo/cli/component/create.go
```diff
@@ -192,25 +205,61 @@ func (co *CreateOptions) setResourceLimits() {
...
+		selectedSourcePath := ""
+		if selectedSourceType == occlient.LOCAL {
+			currentDirectory, err := util.GetAbsPath("./")
+			if err != nil {
+				return err
+			}
+			selectedSourcePath = ui.EnterLocalInputTypePath(currentDirectory)
+		} else if selectedSourceType == occlient.BINARY {
+			selectedSourcePath = ui.EnterBinaryInputTypePath()
+		} else if selectedSourceType == occlient.GIT {
```
golangcibot: empty branch (from `staticcheck`)

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,108 @@
...
+		Message: "Which component type would wish to create",
+		Options: getComponentTypeNameCandidates(options),
+	}
+	err := survey.AskOne(prompt, &componentType, survey.Required)
+	ui.HandleError(err)
+	return componentType
+}
+
+func getComponentTypeNameCandidates(options []catalog.CatalogImage) []string {
+	result := make([]string, len(options), len(options))
```
golangcibot: should use make([]string, len(options)) instead (from `gosimple`)

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,108 @@
...
+			sort.Strings(option.NonHiddenTags)
+			return option.NonHiddenTags
+		}
+	}
+	glog.V(4).Infof("Selected component type %s was not part of the catalog images", selectedComponentType)
+	return []string{}
+}
+
+func SelectSourceType(sourceTypes []occlient.CreateType) occlient.CreateType {
+	options := make([]string, len(sourceTypes), len(sourceTypes))
```
golangcibot: should use make([]string, len(sourceTypes)) instead (from `gosimple`)

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,212 @@
...
+func SelectYesNo(message string) bool {
+	var componentType string
+	prompt := &survey.Select{
+		Message: message,
+		Options: []string{"Yes", "No"},
+	}
+	err := survey.AskOne(prompt, &componentType, survey.Required)
+	ui.HandleError(err)
+
+	if componentType == "Yes" {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

geoand:
So now I have added what I think is all the features needed.

I would like some feedback from @jorgemoralespou @kadel @metacosm before polishing the PR

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	sort.Strings(result)
+	return result
+}
+
+// SelectImageTagInteractively lets the user to select a specific tag for the previously selected builder image in a prompt
+func SelectImageTagInteractively(options []catalog.CatalogImage, selectedComponentType string) string {
+	var tag string
+	prompt := &survey.Select{
+		Message: fmt.Sprintf("Which version of type %s  would you wish to create", selectedComponentType),
```
metacosm: Wording is a little awkward here. Also, extra space after `%s`

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,212 @@
...
+		Message: "What git ref (branch, tag, commit) would you wish to use",
+		Default: defaultRef,
+	}
+	err := survey.AskOne(prompt, &path, survey.Required)
+	ui.HandleError(err)
+	return path
+}
+
+// SelectYesNo displays a given message and asks the user to respond with yes or no and returns true of the answer is Yes
+func SelectYesNo(message string) bool {
```
metacosm: There's a component that does that: `survey.Confirm`. See https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/service/ui/ui.go#L262-L265

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,212 @@
...
+		result[i] = strings.TrimSpace(port)
+	}
+	return result
+}
+
+// EnterPorts allows the user to specify the ports to be used in a prompt
+func EnterPorts() []string {
+	var portsStr string
+	prompt := &survey.Input{
+		Message: "Enter the ports you wish to set (for example: 8080,8100/tcp,9100/udp)",
```
metacosm: You could use the `Help` here for the example (though it'd be less immediately obvious to users).

pkg/odo/cli/component/create.go
```diff
@@ -192,25 +208,90 @@ func (co *CreateOptions) setResourceLimits() {
...
+				}
+				cpuQuantity := util.FetchResourceQuantity(corev1.ResourceCPU, cpuMin, cpuMax, "")
+				if cpuQuantity != nil {
+					resourceQuantity = append(resourceQuantity, *cpuQuantity)
+				}
+				co.CreateArgs.Resources = resourceQuantity
+			}
 
-	co.setResourceLimits()
+			co.CreateArgs.Wait = ui.SelectYesNo("Would you wish to wait until the component is fully ready after after creation")
```
metacosm: I think you're missing an assignment to `co.wait` as well here since I still get the `this might take a few moments` message even if I asked to wait.

pkg/odo/cli/component/ui/ui.go
geoand: I'll check it out, thanks!

pkg/odo/cli/component/ui/ui.go
geoand: I'll see what it looks like, thanks

pkg/odo/cli/component/create.go
geoand: Indeed yes

geoand:
> Looks pretty good apart from the comments I added. I'm not too sure about the flow (i.e. I'd like to explore the multi-select option for optional values) and probably tie the values to the validation infrastructure where possible but I think that could happen in subsequent PRs.

Yeah, I would prefer to explore better user flows in a subsequent PR :)

pkg/odo/cli/component/ui/ui.go
geoand: I tried it and I actually think that the current approach would be more useful for users

pkg/odo/cli/component/ui/ui.go
geoand: Used it, thanks

pkg/odo/cli/component/create.go
geoand: Fixed

pkg/odo/cli/component/create.go
```diff
@@ -251,12 +251,12 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
 
-		advancedOptionsNeeded := ui.SelectYesNo("Do you wish to set advanced options")
+		advancedOptionsNeeded := ui.Proceed("Do you wish to set advanced options")
 		if advancedOptionsNeeded {
 			co.CreateArgs.Ports = ui.EnterPorts()
 			co.CreateArgs.EnvVars = ui.EnterEnvVars()
 
-			resourceLimitsNeeded := ui.SelectYesNo("Do you wish to set resource limits")
+			resourceLimitsNeeded := ui.Proceed("Do you wish to set resource limits")
 			if resourceLimitsNeeded {
```
metacosm: Nit-picking but how about: `if ui.Proceed…` then now? :)

pkg/odo/cli/component/ui/ui.go
```diff
@@ -134,17 +134,16 @@ func enterGitRef(defaultRef string) string {
 	return path
 }
 
-// SelectYesNo displays a given message and asks the user to respond with yes or no and returns true of the answer is Yes
-func SelectYesNo(message string) bool {
-	var componentType string
-	prompt := &survey.Select{
+// Proceed displays a given message and asks the user if they want to proceed
+func Proceed(message string) bool {
```
metacosm: 👍

pkg/odo/cli/component/create.go
geoand: Not sure I follow :(

geoand: Sorry yeah, now I get get it - I'll fix it :)

geoand: Fixed

geoand:
@metacosm do you mind if I rebase onto the current master, since my master is a little behind and doesn't contain the e2e  fixes that should take care of the flakes?

pkg/odo/cli/component/ui/ui.go
metacosm: "Which version of '%s' component type do you wish to use"?

metacosm: Did you remove the extra space?

metacosm:
> @metacosm do you mind if I rebase onto the current master, since my master is a little behind and doesn't contain the e2e fixes that should take care of the flakes?

No, go ahead.

pkg/odo/cli/component/ui/ui.go
geoand: Nope, forgot to do that. I will do that now and also use the message you propose

geoand: Fixed

geoand:
> > @metacosm do you mind if I rebase onto the current master, since my master is a little behind and doesn't contain the e2e fixes that should take care of the flakes?
> 
> No, go ahead.

All comments should now be addressed.
The PR was also rebased onto the latest master

geoand:
> The change made to re-organize component commands under `component` combined with the changes in allowed arg numbers cause an issue. I need to run more tests simply with components but using this PR `odo component create` fails because `odo` tries to switch to the `create` component.

Any pointers on how to fix it?

metacosm:
> Any pointers on how to fix it?

Not yet but the issue lies with: https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L26-L32 and/or https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L63-L68 which actually doesn't look right… :(

geoand:
> > Any pointers on how to fix it?
> 
> Not yet but the issue lies with: https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L26-L32 and/or https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L63-L68 which actually doesn't look right… :(

Ah yes, that does look fishy...
Perhaps for the time being we should revert the change that brought the commands under component?

metacosm:
> > > Any pointers on how to fix it?
> > 
> > 
> > Not yet but the issue lies with: https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L26-L32 and/or https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L63-L68 which actually doesn't look right… :(
> 
> Ah yes, that does look fishy...
> Perhaps for the time being we should revert the change that brought the commands under component?

The issue is that I'm not sure how to fit the `oc` command pattern with the commands that do implied `get/set` commands… I think I have a fix for the issue, though, but it's a little ugly, just replace the `Run:` function of the `component` command by:
```go
cmd, args, e := cmd.Find(args)
odoutil.LogErrorAndExit(e, "")
cmd.Run(cmd, args)
```

geoand:
> > > > Any pointers on how to fix it?
> > > 
> > > 
> > > Not yet but the issue lies with: https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L26-L32 and/or https://github.com/redhat-developer/odo/blob/master/pkg/odo/cli/component/component.go#L63-L68 which actually doesn't look right… :(
> > 
> > 
> > Ah yes, that does look fishy...
> > Perhaps for the time being we should revert the change that brought the commands under component?
> 
> The issue is that I'm not sure how to fit the `oc` command pattern with the commands that do implied `get/set` commands… I think I have a fix for the issue, though, but it's a little ugly, just replace the `Run:` function of the `component` command by:
> 
> ```go
> cmd, args, e := cmd.Find(args)
> odoutil.LogErrorAndExit(e, "")
> cmd.Run(cmd, args)
> ```

Do you want me to add that as well?

metacosm:
Looks like it only required a rebase.

jorgemoralespou:
@geoand Looks good. Some comments:

- When selecting path for local, if path does not exist, it will not fail, but at provisioning it will fail (`stat /temp: no such file or directory`)
- When giving a name for the component, if there is already a component with that name, should check (if possible), as it'll fail at the end of the process (while provisioning, otherwise `component with name backend already exists in application app-pobe`)
- Seems as port and ENV are more important than limits. Not sure why, but I'm ok, just asking if there is a rationale.
- It doesn't check port format, it fails at ptovisioning time (`strconv.ParseInt: parsing "uuu": invalid syntax`)
- Environments doesn't check for format (you can input DEBUG, instead of DEBUG=xyz) and it works, but then fails at provisioning (`invalid syntax for env, please specify a VariableName=Value pair`)
- I would ask if same or different limits for minimum and maximum, as you're assuming a user would prefer to input these, instead of the default --memory which is min=max. Another option is to set the default for max, the value the user inputed for min, as he can just `click`
- It didn't ask for an app, so it created a new app, instead of using the active one (that already existed in my case)
- If not setting advanced options, it doesn't ask you if you want to wait for component creation.
- When waiting for pod to start, message should not say pod, but component. `Waiting for component to start`

geoand:
I'll try to address those comments tomorrow

geoand:
So the basic theme that needs to be added is validation - that is fair. I'll wrap that up today so you can try it again @jorgemoralespou

geoand:
> @geoand Looks good. Some comments:
> 

I pushed various updates to this PR @jorgemoralespou to address your concerns. 
Replies inline.

>     * When selecting path for local, if path does not exist, it will not fail, but at provisioning it will fail (`stat /temp: no such file or directory`)
> 

Fixed

>     * When giving a name for the component, if there is already a component with that name, should check (if possible), as it'll fail at the end of the process (while provisioning, otherwise `component with name backend already exists in application app-pobe`)
> 

Fixed

>     * Seems as port and ENV are more important than limits. Not sure why, but I'm ok, just asking if there is a rationale.
> 

No reason in particular, we can of course do whatever you like.

>     * It doesn't check port format, it fails at ptovisioning time (`strconv.ParseInt: parsing "uuu": invalid syntax`)
> 

Fixed

>     * Environments doesn't check for format (you can input DEBUG, instead of DEBUG=xyz) and it works, but then fails at provisioning (`invalid syntax for env, please specify a VariableName=Value pair`)
> 

Fixed

>     * I would ask if same or different limits for minimum and maximum, as you're assuming a user would prefer to input these, instead of the default --memory which is min=max. Another option is to set the default for max, the value the user inputed for min, as he can just `click`
> 

I opted for the second option. We now first ask for the max, and the default min is set to that value so the user can just press enter and continue

>     * It didn't ask for an app, so it created a new app, instead of using the active one (that already existed in my case)
> 

I wasn't able to reproduce this

>     * If not setting advanced options, it doesn't ask you if you want to wait for component creation.
> 

Fixed

>     * When waiting for pod to start, message should not say pod, but component. `Waiting for component to start`

Fixed

geoand:
This should now be considered feature complete, please review and let me know what you think

jorgemoralespou:
@geoand Still some minor things:

- message when failing validation for port is not nice: `X Sorry, your reply was invalid: invalid port number uuuu: strconv.ParseInt: parsing "uuuu": invalid syntax`. Should be only: `X Sorry, your reply was invalid: invalid port number uuuu`
- Validating limits does not validate properly: `100U` is not a valid limit for cpus. Same for memory. I get a panic: `panic: cannot parse '100U': quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'`
- Validating limits happen after setting the 4 values.

geoand:
@jorgemoralespou Currently we are not doing any limit validation during the input phase and frankly I don't even know how to do it - meaning I don't know what is accepted by Kubernetes.

I will proceed to fix the error message during port validation

geoand:
Port validation error message fixed

geoand:
> Tested it locally and works like a charm but I wish choosing the component local path provided either an autocompletion or may be some interactive way to choose the directory(but may be the choice of which path to start from is the difficulty)

I was wishing for the same thing, unfortunately it doesn't seem possible... :(

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,228 @@
...
+	result := make([]string, len(options))
+	for i, option := range options {
+		result[i] = option.Name
+	}
+	sort.Strings(result)
+	return result
+}
+
+// SelectImageTagInteractively lets the user to select a specific tag for the previously selected builder image in a prompt
+func SelectImageTagInteractively(options []catalog.CatalogImage, selectedComponentType string) string {
```
metacosm: I'd remove the `Interactively` part for this function since you haven't used it in the other functions.

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,228 @@
...
+	ui.HandleError(err)
+
+	return response
+}
+
+// EnterPorts allows the user to specify the ports to be used in a prompt
+func EnterPorts() []string {
+	var portsStr string
+	prompt := &survey.Input{
+		Message: "Enter the ports you wish to set (for example: 8080,8100/tcp,9100/udp)",
```
metacosm: You should add here that you can just press enter / not input anything if you want to bypass this setup.

pkg/odo/cli/component/ui/ui.go
```diff
@@ -0,0 +1,228 @@
...
+	ui.HandleError(err)
+
+	return util.GetSplitValuesFromStr(portsStr)
+}
+
+// EnterEnvVars allows the user to specify the environment variables to be used in a prompt
+func EnterEnvVars() []string {
+	var envVarsStr string
+	prompt := &survey.Input{
+		Message: "Enter the environment variables you would like to set (for example: MY_TYPE=backed,PROFILE=dev)",
```
metacosm: Same as for ports.

pkg/odo/util/validation/validators_test.go
```diff
@@ -54,3 +56,45 @@ func TestNameValidator(t *testing.T) {
...
+	err = PathValidator("/tmp/1ewfvjnfkhvhbf")
+	if err == nil {
+		t.Error("path validator should return an error when the path does not exist")
+	}
+}
+
+func TestPortValidator(t *testing.T) {
+	err := PortsValidator("8080,9090/udp")
+	if err != nil {
+		t.Errorf("port validator should have accepted an correct port declaration, but got: %v instead", err)
```
metacosm: an -> a

pkg/util/util.go
```diff
@@ -436,3 +436,56 @@ func GetSortedKeys(mapping map[string]string) []string {
...
 }
+
+//returns a slice containing the split string, using ',' as a separator
+func GetSplitValuesFromStr(inputStr string) []string {
+	if len(inputStr) == 0 {
+		return []string{}
+	}
+
+	result := strings.Split(inputStr, ",")
+	for i, port := range result {
```
metacosm: I'd rename that `port` variable to something else ;)

geoand: LOL, isn't port the best possible name?

pkg/odo/util/validation/validators_test.go
geoand: Fixed

pkg/odo/cli/component/ui/ui.go
geoand: Fixed

pkg/odo/cli/component/ui/ui.go
geoand: Fixed

geoand:
@jorgemoralespou OK to merge?

metacosm:
Let's address any remaining issues in subsequent PRs.

geoand:
SGTM

kadel:
> Let's address any remaining issues in subsequent PRs.

do we have an issue to track remaining issues?

geoand:
> > Let's address any remaining issues in subsequent PRs.
> 
> do we have an issue to track remaining issues?

We don't have any remaining issues at this point. Just whatever @jorgemoralespou uncovers :)

PR #1306, opened 2019-02-07
Add Runnable interface and use it to introduce generic Run function.

metacosm:
Most commands all use the same sequence of function that is now factored out.

No, but this change removes duplicated code and adds the opportunity to modify all commands Run functionality at once.

## How to test changes?
No changes in behaviour should occur.


kadel:
2 acks, 0 changes requests,  green test -> merge

kadel:
@mik-dass 
> LGTM

Don't be afraind to hit that magical button ;-)

PR #1307, opened 2019-02-08
Issue #1205 Update test suite to handle stderr

### LINKED ISSUE #1205: Update test suite function runCmd to handle stderr
[kind/Enhancement]

```func runCmd()``` in test suite

## Why is this needed? 
Right now in test suite function ```runCmd``` doesn't handle/capture the reported error.

### PR DESCRIPTION
**Pull request by amitkrout**

Error will be reported in ```stderr```
fixes #1205

## How to test changes?

make test*
Check travis for test execution reference

### DIFF
--- a/tests/e2e/e2e.go
+++ b/tests/e2e/e2e.go
@@ -19,36 +19,40 @@ func generateTimeBasedName(prefix string) string {
 }
 
 func getActiveElementFromCommandOutput(command string) string {
-	result := runCmd(command + " | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
+	result := runCmdShouldPass(command + " | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
 	return strings.TrimSpace(result)
 }
 
-func runCmd(cmdS string) string {
+// cmdRunner runs a command
+// and returns the stdout, stderr and exitcode
+func cmdRunner(cmdS string) (string, string, int) {
 	cmd := exec.Command("/bin/sh", "-c", cmdS)
 	fmt.Fprintf(GinkgoWriter, "Running command: %s\n", cmdS)
 	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
 
 	// wait for the command execution to complete
 	<-session.Exited
-	Expect(session.ExitCode()).To(Equal(0))
-	Expect(err).NotTo(HaveOccurred())
 
-	return string(session.Out.Contents())
+	Expect(err).NotTo(HaveOccurred())
+	return string(session.Out.Contents()), string(session.Err.Contents()), session.ExitCode()
 }
 
-// runFailCmd runs a failing command
-// and returns the stdout
-func runFailCmd(cmdS string, exitCode int) string {
-	cmd := exec.Command("/bin/sh", "-c", cmdS)
-	fmt.Fprintf(GinkgoWriter, "Running command: %s\n", cmdS)
-	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+// runCmdShouldPass runs a command
+// and returns stdout if passes,
+// error out otherwise
+func runCmdShouldPass(cmd string) string {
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: tests/e2e/e2e.go [hunk 1]
```diff
@@ -19,25 +19,19 @@ func runCmd(cmdS string) string {
...
+	if err != nil {
+		return err.Error()
+	}
 
-	return string(session.Out.Contents())
+	commandFails := (session.ExitCode() != 0 || len(string(session.Err.Contents())) != 0)
+	if commandFails == true {
+		return string(session.Err.Contents())
+	}
+	if commandFails == false {
```
golangcibot: should omit comparison to bool constant, can be simplified to !commandFails (from `gosimple`)

amitkrout:
Debugging the CI failure locally

amitkrout:
@metacosm Does it make sense now ? PTAL

tests/e2e/e2e.go
```diff
@@ -19,36 +19,40 @@ func generateTimeBasedName(prefix string) string {
...
-	session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
-
-	// wait for the command execution to complete
-	<-session.Exited
-	Expect(session.ExitCode()).To(Equal(exitCode))
-	Expect(err).NotTo(HaveOccurred())
+// runCmdShouldPass runs a command
+// and returns stdout if passes,
+// error out otherwise
+func runCmdShouldPass(cmd string) string {
```
metacosm: Looks better now. However, I'd rather keep the function names as they were before so this function should be called `runCmd`.

tests/e2e/e2e.go
```diff
@@ -19,36 +19,40 @@ func generateTimeBasedName(prefix string) string {
...
+	stdout, _, exitcode := runCmd(cmd)
+	Expect(exitcode).To(Equal(0))
+	return stdout
+}
 
-	return string(session.Out.Contents())
+// runCmdShouldFail runs a command
+// and returns stderr if fails,
+// error out otherwise
+func runCmdShouldFail(cmd string) string {
```
metacosm: I'm OK with this function being renamed this way.

tests/e2e/e2e.go
```diff
@@ -19,36 +19,40 @@ func generateTimeBasedName(prefix string) string {
...
 func getActiveElementFromCommandOutput(command string) string {
-	result := runCmd(command + " | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
+	result := runCmdShouldPass(command + " | sed -n '1!p' | awk 'FNR==2 { print $2 }'")
 	return strings.TrimSpace(result)
 }
 
-func runCmd(cmdS string) string {
+// runCmd runs a command
+// and returns the stdout, stderr and exitcode
+func runCmd(cmdS string) (string, string, int) {
```
metacosm: To minimize impact on client code (and potential errors by developers being used to the old `runCmd` function, this new function should be renamed differently (perhaps `internalRunCmd`?) and the name `runCmd` should be kept to cover the original cases. Does that make sense?

tests/e2e/e2e.go
amitkrout: > However, I'd rather keep the function names as they were before so this function should be called ```runCmd```.

Right, even I had the similar thought. But later I thought that the name itself can be made in more talkative way like ```cmdRunShouldPass```. Moreover names ```cmdRunShouldPass``` and ```cmdRunShouldFail``` looks more expressive over ```runCmd``` and ```runCmdShouldFail```

tests/e2e/e2e.go
amitkrout: The real impact on test code will not hurt us a lot IMO unless there is a major functionality change. We can ignore the impact for changing function name and of course handling of ```stderr``` in test suite. The current name change in the long run will bring some good test experience.

metacosm: OK, the important thing was that the old name shouldn't be used for a function that's different since it might cause people to use it improperly thinking it was still as before.

PR #1308, opened 2019-02-08
add circleci file

syamgk:
This pr will add circleci config for building odo in circle ci

 #1309

## How to test changes?
proof-read the config and check if the builds are successful for this pr on circle-ci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,121 @@
+version: 2
+
+# this circle ci config is Inspired by minishift's scripts
+# https://github.com/minishift/minishift/blob/cfc599cffdc41d31b207f7cfd40190be6b86b858/.circleci/config.yml
+# and an orb defined at https://github.com/snowdrop/circleci-orbs for provisioning the oc cluster
+
+jobs:
+  build:
+    machine: true
+    environment:
+      GOVERSION: 1.11.4
+    working_directory: ~/go/src/github.com/redhat-developer/odo
+
+    steps:
+    - checkout
+    - run:
+        name: Settingup machine with golang
+        command: |
+            echo 'export GOPATH=$HOME/go' >> $BASH_ENV
+            echo 'export PATH=$GOPATH/bin:/usr/local/go/bin/:$PATH' >> $BASH_ENV
+            sudo rm -rf /usr/local/go
+            curl https://storage.googleapis.com/golang/go${GOVERSION}.linux-amd64.tar.gz | sudo tar -C /usr/local -xz
+            go env
+            mkdir -p $GOPATH/bin
+
+    - run: 
+        name: Build Binary
+        command: |
+            go version
+            env
+            cd $GOPATH/src/github.com/redhat-developer/odo
+            make goget-tools
+            make
+            sudo cp odo /usr/bin
+    #    
+    #    - run: 
+    #        name: oc cluster up
+    #        command: |
+    #            pwd
+    #            cd $GOPATH/src/github.com/redhat-developer/odo
+    #            pwd
+    #            ls
+    #            ./scripts/oc-cluster.sh service-catalog
+    #
+    - run:
+        name: "Download, install okd, kubectl tools, launch okd and grant cluster admin role for the admin user"
+        command: |
+            kube_version=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,95 @@
...
+            go version
+            env
+            cd $GOPATH/src/github.com/redhat-developer/odo
+            make goget-tools
+            make
+            sudo cp odo /usr/bin
+    
+    - run:
+        name: "Download, install okd, kubectl tools, launch okd and grant cluster admin role for the admin user"
+        command: |
```
kadel: Why we can't use the same script as we have for travis? 
./scripts/oc-cluster.sh

syamgk: I consumed an orb which georgies shared(the one they are using here[1] ) during the initial runs in my fork and it successfully ran all tests which do not require `service-catalog`.  
so pushed it here for testing, now started a test using `oc-cluster.sh` script and added all the test suite.

[1] https://github.com/snowdrop/spring-boot-bom/blob/sb-1.5.x/.circleci/config.yml#L7
[2] https://circleci.com/orbs/registry/orb/snowdrop/openshift#commands-launch_openshift_311

syamgk: it seems like there are errors[1] while trying with `scripts/oc-cluster.sh` 
so for now we can run test suites which doesnot require service-catalog on circlieCI.

[1] https://circleci.com/gh/syamgk/odo/55?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

geoand:
@syamgk IIUC you only launch Openshift once right? And then subsequently proceed to execute e2e tests  in succession.

syamgk:
> @syamgk IIUC you only launch Openshift once right? And then subsequently proceed to execute e2e tests in succession.

yes and without `service-catalog` using the scripts found here[1], also same on ci.centos
there is one issue with this on ci.co https://github.com/redhat-developer/odo/pull/1290#issuecomment-461806195. 

not using orbs here since we have to enable it for the org on github.
[1] https://circleci.com/orbs/registry/orb/snowdrop/openshift#commands-launch_openshift_311

geoand:
> > @syamgk IIUC you only launch Openshift once right? And then subsequently proceed to execute e2e tests in succession.
> 
> yes and without `service-catalog` using the scripts found here[1], also same on ci.centos
> there is one issue with this on ci.co [#1290 (comment)](https://github.com/redhat-developer/odo/pull/1290#issuecomment-461806195).
> 
> not using orbs here since we have to enable it for the org on github.
> [1] https://circleci.com/orbs/registry/orb/snowdrop/openshift#commands-launch_openshift_311

Sounds good to me. Makes things a lot faster!

.circleci/config.yml
geoand: I'm confused, why not reuse the orb? If you don't, then it probably makes sense to just use the script within odo to launch the cluster

geoand: Just saw that you are not using the orb because it requires enabling them. IMHO that shouldn't be a problem :)

metacosm: I'm confused as well…

syamgk: @metacosm @geoand , here this its using the same scripts which is on that orb openshift_311.
since the other script `oc-cluster.sh` is erroring as mentionend here https://github.com/redhat-developer/odo/pull/1308/files#r255052209

geoand: OK, but why not use the orb itself? It's just a matter of configuring the org and/or repo settings right?

metacosm:
I'm no expert on CircleCI so I'd say let's merge this and see what happens… We should have a parallel PR that activates only the tests that are not covered by CircleCI in Travis so that the workload is split between both.

geoand:
I would still prefer to use the orb as to have no extra code, but if it's not possible, then I'm obviously fine with that.

Other than the orb, this looks good to me

metacosm:
See #1313.

kadel:
It is not ideal, this means that we have to mantain two separate ways for spinining  up openshift (one for travis, one for circle-ci :-( ) 

We should ether enable orbs or modify our script to work on both.

kadel:
created separate issue.
We can merge this and see how it works.

geoand:
I still think however that someone with admin permissions in `redhat-developer` needs to properly configure the project

Issue #1144, opened 2018-12-17
The use-case of `--wait` and having it enabled by default with `odo service create`


cdrage:
This is just to spur discussion, but whenever we push a component, by default, Odo will "wait" and the spinner will spin until it's completed.

This is not the case however with `odo service create`. Instead, we output:

```sh
▶ ./odo service create dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret

 ✓  Service 'my-postgresql-db' was created
Progress of the provisioning will not be reported and might take a long time.
You can see the current status by executing 'odo service list'
```

This is completely different than `odo create` or `odo push`.

I believe we should unify / have similar interactions with both commands. 

My suggestions:

 - By default, we should show the status of provisioning and have Odo wait until the service is deployed. Last time I deployed a database it took less than 30 seconds.
 - Show provisioning output by default when deploying (same as the above suggestion)
 - Have --wait an alias / the same as --show for the two commands, we should either have --wait OR --show for both commands, but not both.

ping @geoand @metacosm @kadel

geoand:
Currently we have a wait flag on odo link (which is off by default).

I think the reason we didn't add wait to odo service create was that it
could take a long time without being an actual blocker in itself for the
user (since having a service on it's own probably won't be of much use to
odo users).
Link is the actual blocker and that's where I prefer we made odo wait by
default.

What do you all think?

jorgemoralespou:
@cdrage I can show you many services that takes minutes to deploy :-( so I'm against having the CLI waiting for the service to be created. Think that it's way different deploying a service implemented as a template than as an APB. Also, think that in many cases, the real provisioning action happens asynchronously, so even if the broker/catalog tells you it's done, it's not, and that's dependant on the service implementation.

I'd rather have the `--wait` be an optional flag that the user passes whenever he knows is going to be fast. Then we have `odo service list` or we can implement `odo service status` to get information about a state of a specific service, but definitely not block the user for the service to be created.

I'm pretty sure that while the service is provisioned, the user will make use of the time for some other stuff (maybe creating a service or pushing some code).

> By default, we should show the status of provisioning and have Odo wait until the service is deployed. Last time I deployed a database it took less than 30 seconds.

I would definitely not do this.

>Show provisioning output by default when deploying (same as the above suggestion)

Have you looked at deploying services from the TSB as well as AB? As I said, this behave very differently, and the output you can get is quite different. What would you show here as output?

[...]

Issue #1063, opened 2018-12-03
Add spinner for service create


cdrage:
Quick issue, we should add spinners / logging for service create

kadel:
I'm not sure if it makes sense to have a spinner for `service create`. It is not waiting for service to be created.

```
▶ odo service create mysql-persistent
Service 'mysql-persistent' was created.
Progress of the provisioning will not be reported and might take a long time.
You can see the current status by executing 'odo service list'
```



```
✓ Service 'mysql-persistent' was created.

Progress of the provisioning will not be reported and might take a long time.
You can see the current status by executing 'odo service list'
```

geoand:
I had a PR that added the wait flag but I think we closed it because it was very out of date.
That said I can easily create a new one and a spinner to it.
What do you think?

jorgemoralespou:
I think it makes sense, but @kadel is the master :-D

geoand:
Just give me the word @kadel and I will get on it 😁

geoand:
Actually it seems like I had added the `--wait` flag for service linking, but not for service create. I can do that in the similar fashion if we think it makes sense

kadel:
`odo service create` shouldn't wait by default.
But adding `--wait` flag makes sense.

geoand:
I'll take this since it's related to #1144

kadel:
closed in #1321

PR #1321, opened 2019-02-12
Add wait flag to service create

geoand:
Add wait flag to service create and also use spinners to display progress

Fixes: #1144, #1063

## How to test changes?
Both
`odo service create mongodb-ephemeral --plan default -p MONGODB_VERSION=3.2`

and 

`odo service create mongodb-ephemeral --plan default -p MONGODB_VERSION=3.2 -w` should work with the latter waiting for the service to come up fully

### DIFF
--- a/tests/e2e/service_test.go
+++ b/tests/e2e/service_test.go
@@ -13,7 +13,7 @@ var _ = Describe("odoServiceE2e", func() {
 
 	Context("odo service creation", func() {
 		It("should be able to create a service", func() {
-			runCmd("odo service create mysql-persistent")
+			runCmd("odo service create mysql-persistent -w")
 			waitForCmdOut("oc get serviceinstance -o name", 1, func(output string) bool {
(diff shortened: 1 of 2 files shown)


tests/e2e/service_test.go
```diff
@@ -13,7 +13,7 @@ var _ = Describe("odoServiceE2e", func() {
 
 	Context("odo service creation", func() {
 		It("should be able to create a service", func() {
-			runCmd("odo service create mysql-persistent")
+			runCmd("odo service create mysql-persistent -w")
```
mik-dass: Maybe one service without the wait flag?

geoand: Good point, but we already have tests that don't use the flag, they just don't show up in the diff :wink:

mik-dass: OK found it. Missed those tests in the first look up. Thanks :+1:

geoand: Np :)

PR #1325, opened 2019-02-13
added component settings and used reflection for deletion

girishramnani:

added component settings and used reflection for deletion

partly  https://github.com/redhat-developer/odo/issues/1294

## How to test changes?

```
odo utils config set componentName test
odo utils config set minCPU 0.5
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -278,7 +278,7 @@ func (c *GlobalConfigInfo) DeleteConfiguration(parameter string) error {
 	if p, ok := asSupportedParameter(parameter); ok {
 		// processing values according to the parameter names
 
-		if err := deleteConfiguration(c, p); err != nil {
+		if err := deleteConfiguration(&c.OdoSettings, p); err != nil {
 			return err
 		}
 	} else {
@@ -383,7 +383,7 @@ func (lci *LocalConfigInfo) DeleteConfiguration(parameter string) error {
 			lci.ComponentSettings.MinCPU = nil
 			lci.ComponentSettings.MaxCPU = nil
 		}
-		if err := deleteConfiguration(lci, parameter); err != nil {
+		if err := deleteConfiguration(&lci.ComponentSettings, parameter); err != nil {
 			return err
 		}
 		return writeToFile(lci.LocalConfig, lci.Filename)
@@ -400,9 +400,9 @@ func deleteConfiguration(info interface{}, parameter string) error {
 	}
 	val := imm.FieldByNameFunc(caseInsensitive(parameter))
 	if !val.IsValid() {
-		return errors.Errorf("unknown parameter :'%s' is not a parameter in local odo config", parameter)
-
+		return fmt.Errorf("unknown parameter :'%s' is not a parameter in odo config", parameter)
 	}
+
 	if val.CanSet() {
 		val.Set(reflect.Zero(val.Type()))
 		return nil
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -341,6 +341,9 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (er
 			lci.ComponentSettings.MinMemory = &value
 		case "maxmemory":
 			lci.ComponentSettings.MaxMemory = &value
+		case "memory":
+			lci.ComponentSettings.MaxMemory = &value
+			lci.ComponentSettings.MinMemory = &value
 		case "ignore":
 			val, err := strconv.ParseBool(strings.ToLower(value))
 			if err != nil {
@@ -367,24 +370,37 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (er
 // it only searches the componentSettings
 func (lci *LocalConfigInfo) GetConfiguration(parameter string) (interface{}, bool) {
 
-	if strings.ToLower(parameter) == "cpu" {
+	switch strings.ToLower(parameter) {
+	case "cpu":
 		if lci.ComponentSettings.MinCPU == nil {
-			return lci.ComponentSettings.MinCPU, true
+			return nil, true
 		}
 		return *lci.ComponentSettings.MinCPU, true
+	case "memory":
+		if lci.ComponentSettings.MinMemory == nil {
+			return nil, true
+		}
+		return *lci.ComponentSettings.MinMemory, true
 	}
+
 	return getConfiguration(lci.ComponentSettings, parameter)
 }
 
 // DeleteConfiguration is used to delete config from local odo config
 func (lci *LocalConfigInfo) DeleteConfiguration(parameter string) error {
 	if parameter, ok := asLocallySupportedParameter(parameter); ok {
-		if parameter == "cpu" {
+
+		switch parameter {
+		case "cpu":
 			lci.ComponentSettings.MinCPU = nil
 			lci.ComponentSettings.MaxCPU = nil
-		}
-		if err := deleteConfiguration(&lci.ComponentSettings, parameter); err != nil {
-			return err
+		case "memory":
+			lci.ComponentSettings.MinMemory = nil
+			lci.ComponentSettings.MaxMemory = nil
+		default:
+			if err := deleteConfiguration(&lci.ComponentSettings, parameter); err != nil {
+				return err
+			}
 		}
 		return writeToFile(lci.LocalConfig, lci.Filename)
 	}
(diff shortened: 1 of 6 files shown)


geoand:
I'm not going to review it because I can't comment on the use of reflection, sorry

pkg/config/config.go
```diff
@@ -610,6 +662,34 @@ const (
 	ComponentType = "ComponentType"
 	// ComponentTypeDescription is human-readable description of the componentType setting
 	ComponentTypeDescription = "The type of component"
+	// ComponentName is the name of the setting controlling the component name
+	ComponentName = "ComponentName"
+	// ComponentNameDescription is human-readable description of the componentType setting
```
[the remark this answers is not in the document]
metacosm: It should be noted that otherwise, things look good! :)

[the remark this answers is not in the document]
girishramnani: addressed

tests/e2e/e2e_test.go
```diff
@@ -105,19 +105,69 @@ var _ = Describe("odoe2e", func() {
...
-		})
-
-		It("should allow unsetting a config locally", func() {
-			runCmd("odo utils config set componenttype java")
-			configOutput := runCmd("odo utils config unset -f componentType")
-			Expect(configOutput).To(ContainSubstring("Local config was successfully updated."))
-			configOutput = runCmd("odo utils config view|grep ComponentType")
-			Expect(configOutput).NotTo(ContainSubstring("java"))
+
+		FIt("should be checking to see if local config values are the same as the configured ones", func() {
```
amitkrout: Do not forget to replace ```Fit``` with ```It```

tests/e2e/e2e_test.go
```diff
@@ -105,19 +105,69 @@ var _ = Describe("odoe2e", func() {
...
+			for _, testCase := range cases {
+				runCmd(fmt.Sprintf("odo utils config set %s %s", testCase.paramName, testCase.paramValue))
+				configOutput := runCmd(fmt.Sprintf("odo utils config view|grep %v", testCase.paramName))
+				Expect(configOutput).To(ContainSubstring(testCase.paramValue))
+				Expect(configOutput).To(ContainSubstring(testCase.paramName))
+			}
+
+		})
+
+		FIt("should allow unsetting a config locally", func() {
```
amitkrout: https://github.com/redhat-developer/odo/pull/1325#discussion_r256703040

pkg/config/config.go
```diff
@@ -340,25 +366,51 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (er
...
+			return err
 		}
-
 		return writeToFile(lci.LocalConfig, lci.Filename)
 	}
 	return errors.Errorf("unknown parameter :'%s' is not a parameter in local odo config", parameter)
 
 }
 
+func deleteConfiguration(info interface{}, parameter string) error {
```
amitkrout: Codecov reports this function is missing UTs. Could you please bring some attention here.

How to check missing coverage
1. Go to the coverage report (for example - https://github.com/redhat-developer/odo/pull/1325#issuecomment-463237149)
2. Click on hyperlink ```pkg/config/config.go``` in ```Impacted Files``` table 
3. Then it will be easy to visualize the misses through color coding.

girishramnani:
> It would be great if you isolate the `odo util config ... ` test in a separate file something like `cmd_util_config_test.go`. Later on it will be helpful for those who only want to enhance/execute `odo util config ...` test.

The package structure is going to change alot hence I am refraining from such changes @amitkrout .

amitkrout:
> The package structure is going to change alot hence I am refraining from such changes @amitkrout .

@girishramnani I think you misunderstood my concern with UTs. Correct me if i am wrong.

My intention is to make a separate integration test file that only keeps the util config test. Does it make sense ?

girishramnani:
I am ok with writing UTs, I am concerned about the extraction of some of
the functions because the package will break into two parts in the next PR.
The two parts being
`odo preferences ` and `odo config`

girishramnani:
so the global config will be moved to `odo preference` and local to `odo config` and that which change the package location and structure

amitkrout:
@girishramnani ```cmdRun``` has been deprecated. You can use ```cmdRunShouldPass``` and counter part ```cmdRunShouldFail``` with same experience.

girishramnani:
Got it

pkg/config/config.go
girishramnani: added UTs @amitkrout

tests/e2e/e2e_test.go
girishramnani: resolved

pkg/config/config.go
surajnarwade: don't use `interface{}` as a type, info is of type `ComponentSettings` I guess

anmolbabu: This can be any config one func for everything that is config and needs to be deleted

pkg/config/config.go
```diff
@@ -340,25 +366,51 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (er
 // GetConfiguration uses reflection to get the parameter from the localconfig struct, currently
 // it only searches the componentSettings
 func (lci *LocalConfigInfo) GetConfiguration(parameter string) (interface{}, bool) {
+
+	if strings.ToLower(parameter) == "cpu" {
```
anmolbabu: why specific handling only for cpu?

Please add it as a comment also please add a note as to what the bool return value stands for
I would be happy if you use a python like structured way of adding function comments for example:

```
// FuncName func does abc
// Parameters:
//    a: a is bla bla
//    b: b is bla bla
// Returns:
//    q: bla bla
//     r: bla bla
```

pkg/config/config_test.go
```diff
@@ -1295,6 +1295,221 @@ func TestSetConfiguration(t *testing.T) {
 	}
 }
 
+func TestSetLocalConfiguration(t *testing.T) {
+
+	tempConfigFile, err := ioutil.TempFile("", "odoconfig")
```
anmolbabu: Can you use testingutil/fileutils.go ?

pkg/config/config.go
girishramnani: @surajnarwade we can use the method to delete any setting from a flat struct and hence the reflection.

pkg/config/config.go
girishramnani: Because setting `cpu` would mean that `mincpu` and `maxcpu` will get set to the same value. It's not a separate parameter but a shortcut.

pkg/config/config.go
```diff
@@ -46,6 +46,19 @@ type ComponentSettings struct {
 
 	// The builder image to use
 	ComponentType *string `json:"ComponentType,omitempty"`
+
+	ComponentName *string `json:"ComponentName,omitempty"`
```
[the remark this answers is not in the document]
girishramnani: The proposal mentioned it be componentName, we can even consider `Type` instead of `ComponentType` if we plan to change the `ComponentName`.

pkg/config/config.go
anmolbabu: why not memory ?

anmolbabu: also where is maxcpu set?

pkg/config/config.go
```diff
@@ -330,7 +335,28 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value string) (er
 		switch parameter {
```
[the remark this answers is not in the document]
girishramnani: It becomes a bit complex if we have to do some operations with the underlying type before assignment. e.g. a `bool` needs to be `parsed`. But I can try that in the next PR :+1: .

pkg/config/config_test.go
girishramnani: I went with uniformity here as all the other tests did this I also went for it

pkg/config/config.go
girishramnani: The issue didn't mention `memory` and here is where the maxcpu is set https://github.com/redhat-developer/odo/blob/69d944f8c6ce5af1c307f847b57f1547254691ea/pkg/config/config.go#L355

pkg/config/config.go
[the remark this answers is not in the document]
girishramnani: But one thing I realized is that as we are using reflection here the way to get this parameter will also change to `odo utils config get name` and that can get confusing for a user @anmolbabu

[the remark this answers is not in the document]
anmolbabu: ok

pkg/config/config.go
girishramnani: added the memory parameter

anmolbabu:
Lets wait for the travis to be green after which @girishramnani feel free to merge this PR

anmolbabu:
Merging based on 5 acks and green travis

PR #1332, opened 2019-02-14
Added .odoignore ignore feature to the odo push command.

### LINKED ISSUE #1036: Support for .odoignore file
[kind/Feature]

We need to support having an `.odoignore` file in the code directory so that when the users do `odo watch` and `odo push` these files are not monitored or pushed to the server.

Initially the creation and management of this file should be manual, but once #803 is done, the initial configuration for this file should come from global configuration tuneable by the user.

The content of this file should support similar syntax to `.gitignore` and should behave the same way.

## Why is this needed?
To improve on push and watch behavior and speed.

mik-dass:
fixes #1036 

Also --ignore flag was also added to the odo push command.

GetAbsGlobExps() was added converting relative glob expressions into absolute ones. Function signature of IsGlobExpMatch() was changed to work without the directory name of the component.

Signed-off-by: mik-dass <mrinald7@gmail.com>

## How to test changes?
1. Trigger `odo push` and check if the files in matching in .gitignore/.odoignore are pushed into the pod or not
2. Trigger `odo push` with the `--ignore` flag and check if the files are pushed into the pod or not


girishramnani:
can we have an e2e test for this @mik-dass ?

amitkrout:
> can we have an e2e test for this @mik-dass ?

@girishramnani Good point, Now a days some issue are filed on latest release version, those were working on previous release version. And these failure are not captured due the absence of integration test.

To avoid regression failure, good to have test in integration test. @mik-dass Make sure that you place integration test to verify this pr.

mik-dass:
@amitkrout Added a e2e test, please have a look :)

tests/e2e/e2e_test.go
```diff
@@ -297,8 +297,23 @@ var _ = Describe("odoe2e", func() {
 				curProj = strings.TrimSuffix(curProj, "\n")
 				// Sleep until status tags and their annotations are created
 				time.Sleep(15 * time.Second)
+
+				runCmdShouldPass("echo -e \".git\ntests/\nREADME.md\" >> " + tmpDir + "/nodejs-ex/.odoignore")
```
amitkrout: As per the PR description i really don't understand what you are doing here. Also can you please update the pr description with proper outcome 

Copied from pr description
```
* Trigger `odo push` and check if the files in matching in .gitignore/.odoignore are pushed into the pod or not

* Trigger `odo push` with the `--ignore` flag and check if the files are pushed into the pod or not
```

Are you sure the command ```"echo -e \".git\ntests/\nREADME.md\" >> " + tmpDir + "/nodejs-ex/.odoignore"``` reporting the right info you are looking for

This is what i am seeing in my bash terminal
```
$ echo -e \".git\ntests/\nREADME.md\" >> nodejs-ex/.odoignore
$ cd nodejs-ex/ && cat .odoignore
".gitntests/nREADME.md"
```

NOTE: Use go library instead of terminal specific command.

amitkrout: @mik-dass Please fix the test code as discussed ;)

mik-dass: ok on it

mik-dass:
@amitkrout Fixed, please have a look again

tests/e2e/e2e_test.go
```diff
@@ -297,8 +297,42 @@ var _ = Describe("odoe2e", func() {
...
+				podName := runCmdShouldPass("oc get pods | grep nodejs | awk '{print $1}' | tr -d '\n'")
+
+				// verify that the views folder got pushed
+				runCmdShouldPass("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep views")
+
+				// verify that the tests was not pushed
+				runCmdShouldFail("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep tests")
+
+				// verify that the README.md file was not pushed
+				runCmdShouldFail("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep README.md")
```
amitkrout: Then delete the freshly created ```.odoignore``` file. So that the next scenario ```It("should create a component and push using the --ignore flag", func() {}``` will get a clean environment to execute command ```runCmdShouldPass("odo push --ignore tests/,README.md")```. 

This is how we can confirm push with ```--ignore``` flag

mik-dass: Yeah I noticed that. About to send a fix :)

tests/e2e/e2e_test.go
```diff
@@ -297,8 +297,42 @@ var _ = Describe("odoe2e", func() {
...
+			})
+
+			It("should create a component and push using the --ignore flag", func() {
+				runCmdShouldPass("odo create " + curProj + "/nodejs push-example --local " + tmpDir + "/nodejs-ex")
```
amitkrout: Use a proper name like ```push-with-ignore```

tests/e2e/e2e_test.go
```diff
@@ -288,7 +288,7 @@ var _ = Describe("odoe2e", func() {
 				Expect(cmpList).To(ContainSubstring(curProj))
 			})
 
-			It("should create a named component", func() {
+			It("should create a named component, add a .odoignore file and push", func() {
```
amitkrout: ```should create a named component, add a .odoignore file and push``` -> ```Should create and push a named component along with a .odoignore file```

@mik-dass Does it make sense ?

mik-dass:
@amitkrout Fixed

tests/e2e/e2e_test.go
```diff
@@ -297,8 +298,49 @@ var _ = Describe("odoe2e", func() {
...
+				runCmdShouldPass("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep views")
+
+				// verify that the tests was not pushed
+				runCmdShouldFail("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep tests")
+
+				// verify that the README.md file was not pushed
+				runCmdShouldFail("oc exec " + podName + " -- ls -lai /opt/app-root/src | grep README.md")
+
+				// remove the .odoignore file
+				if os.Remove(ignoreFilePath) != nil {
```
amitkrout: @mik-dass  Cool!!!
Write the file remove function in util.go and call that function in the test file.

Advantage of doing this
1. Test file will be clean and easy to understand 
2. Same function can be used by multiple test file

kadel: Why?  This is just a single function call, it will look exactly the same.

tests/e2e/e2e_test.go
```diff
@@ -288,7 +289,7 @@ var _ = Describe("odoe2e", func() {
 				Expect(cmpList).To(ContainSubstring(curProj))
 			})
 
-			It("should create a named component", func() {
+			It("should create a named component, add a .odoignore file and then push the contents from the source excluding the contents in the .odoignore file", func() {
```
amitkrout: Too lengthy. Why you are not adding the description as i suggested one https://github.com/redhat-developer/odo/pull/1332#discussion_r259273409 or similar

tests/e2e/e2e_test.go
amitkrout: My concern is to make test file easy to understand i mean in a talkative way

For example 
1. 
```
status := removeFile(ignoreFilePath)
Expect(status).Should(BeTrue())
```
2.
Or you can write the same code in more understandable ginkgo way
```
Expect(os.Remove(ignoreFilePath)).To(BeNil())
```

Example 2 would be more reasonable IMO as a single function call

mik-dass:
@amitkrout Done. Please have a look again

PR #1334, opened 2019-02-14
Use ui.Proceed instead of log.AskF

metacosm:
Adding interactive mode to commands has brought along a library that provides UI elements to ask questions to users. It makes sense to use these elements where we ask the user for confirmation. This simplifies client code and makes sure the experience is consistent across commands.

No.

## How to test changes?
Any of the `delete` (`app`, `component`, etc.) commands that ask for confirmation should still work as before, though the UI might be a little different.

### DIFF
--- a/pkg/odo/cli/ui/common.go
+++ b/pkg/odo/cli/ui/common.go
@@ -24,3 +24,21 @@ func HandleError(err error) {
 func GetValidatorFor(prop validation.Validatable) survey.Validator {
 	return survey.Validator(validation.GetValidatorFor(prop))
 }
+
+// Proceed displays a given message and asks the user if they want to proceed using the optionally specified Stdio instance (useful
+// for testing purposes)
+func Proceed(message string, stdio ...terminal.Stdio) bool {
+	var response bool
+	prompt := &survey.Confirm{
+		Message: message,
+	}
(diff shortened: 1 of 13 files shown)


geoand:
Has the handling of the service properties also changed?

metacosm:
> Has the handling of the service properties also changed?

Not sure I understand the question…

geoand:
> > Has the handling of the service properties also changed?
> 
> Not sure I understand the question…

Never mind, I got mixed up when looking at the changes to the service parameter UI handling - they make sense now

pkg/odo/cli/ui/common.go
```diff
@@ -24,3 +24,21 @@ func HandleError(err error) {
 func GetValidatorFor(prop validation.Validatable) survey.Validator {
 	return survey.Validator(validation.GetValidatorFor(prop))
 }
+
+// Proceed displays a given message and asks the user if they want to proceed using the optionally specified Stdio instance (useful
+// for testing purposes)
+func Proceed(message string, stdio ...terminal.Stdio) bool {
```
geoand: :+1:

girishramnani:
@metacosm works fine locally, merging

Issue #1173, opened 2019-01-04
`odo` extension architecture


metacosm:
[kind/Feature]

## Which functionality do you think we should add?
Adding an extension mechanism for `odo` would allow interested third-parties to extend the tool to suit their own needs without needing features to be added to `odo` itself. Extensions would work as plugins that would be dynamically loaded when the tool starts and communicate with `odo` via an event system over an event bus to be able to have decoupled extensions implemented as listeners of events produced at different stages of command execution. Commands would thus be able to push events during different stages of their execution (e.g. pre-run, post-complete, post-validate, post-run). The bus will allow interested third parties to register listeners to react to events and possibly interrupt the command processing based on this (by returning a specific error type, which will allow the bus to abort the command processing and notify listeners that might have already performed some process based on the event to revert their work).

## Why is this needed?
Not strictly needed but would be a nice to have feature. This could, in particular, allow runtimes to customize `odo`'s behavior to account for their specificity or make it easier to implement a scaffolding feature to more easily bootstrap projects. Another idea could be to use a listener to play the commands the user triggers on different clusters than the one the user is logged to…

metacosm:
Golang issue regarding Windows support: https://github.com/golang/go/issues/19282

johnfriz:
Hi Folks. Just wondering what the plan in to develop the plugin concept further? Is there any follow up work planned at this time?

PR #1335, opened 2019-02-14
PoC: Initial implementation of a very basic kubectl-style plugin system.

metacosm:
See what's involved in supporting a `kubectl`-style of plugins. Note that this currently doesn't attempt to do sub-commands as `kubectl` does. Currently, `odo` will only load plugins defined as binaries located in `~/.odo/plugins` directory and named `<name>.odo.plugin` where name is the name of the command `odo` will use for the plugin. So, to implement a `scaffold` command in `odo`, you could write a binary that does this, put it in `~/.odo/plugins` and name it `scaffold.odo.plugin`. Since `odo` doesn't know any such command, it will attempt to see if it knows a plugin by that name when you run `odo scaffold`. All arguments are passed. Any standard output from the called binary is output directly. Any output to `stderr` is interpreted as an error.

Not exactly, but see #1173 

## How to test changes?
Put any binary in `~/.odo/plugins` with the appropriate name format and go crazy! Hint: it's possible to have `odo` call itself as a plugin.

### DIFF
--- a/cmd/odo/odo.go
+++ b/cmd/odo/odo.go
@@ -13,10 +13,10 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
+	"syscall"
 )
 
 const pluginSuffix = ".odo.plugin"
@@ -32,13 +32,7 @@ type pluginImpl struct {
 }
 
 func (p pluginImpl) execute(args []string) error {
-	cmd := exec.Command(p.path, args...)
-	out, err := cmd.Output()
-	if err != nil {
-		return err
-	}
-	fmt.Println(string(out))
-	return nil
+	return syscall.Exec(p.path, append([]string{p.path}, args...), os.Environ())
 }
 
 func main() {
--- a/cmd/odo/odo.go
+++ b/cmd/odo/odo.go
@@ -3,8 +3,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"os"
-
 	"github.com/golang/glog"
 	"github.com/posener/complete"
 	"github.com/redhat-developer/odo/pkg/config"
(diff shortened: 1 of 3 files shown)


cmd/odo/odo.go
```diff
@@ -14,9 +12,38 @@ import (
...
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"strings"
 )
 
+const pluginSuffix = ".odo.plugin"
+
+var plugins = make(map[string]plugin, 10)
```
kadel: I guess this is just for PoC. Or we can't have more than 10 plugins? :-)

metacosm: This is the initial capacity, isn't it? Maps are growing on demand past this initial capacity, no?

metacosm:
You can try this PR with our `scaffold` command, see https://github.com/snowdrop/scaffold-command.

jorgemoralespou:
Nice.
Conventions on kubectl and oc are that it finds binaries in the path that starts with `kubectl-` or `oc-` but I like better having a plugin directory for it. Still not sure if the kubectl/oc way should be supported.

Plugin works fine, and shows help, but:

- `odo --help´ does not show the plugins as commands
- If I do completion it does not get completion for plugins as well

Apart from that, looks nice.

metacosm:
> Nice.
> Conventions on kubectl and oc are that it finds binaries in the path that starts with `kubectl-` or `oc-` but I like better having a plugin directory for it. Still not sure if the kubectl/oc way should be supported.
> 
> Plugin works fine, and shows help, but:
> 
>     * `odo --help´ does not show the plugins as commands
> 
>     * If I do completion it does not get completion for plugins as well

Yep, I was going to work on these aspects next. Not too sure about help but maybe using a `odo plugin list` command?

metacosm:
@jorgemoralespou give it another try, plugins should now appear in help and be completable.

girishramnani:
This PR as of now needs alot of work due to the major changes in odo. It would be better if a new PR using the current master is created at a later point when plugins become priority. 

Hence closing the PR

odra:
Hello, I would be highly interested in this - is it stil being worked on?

kadel:
> Hello, sorry for jumping late here but I would be highly interested in this - is it stil being worked on?

It is not being worked on currently. But the plugin system is still on our roadmap, and we will definitely get back to this.

johnfriz:
@kadel / @girishramnani - any idea on current timelines / priority for this? Are there any docs / specs beyond this spike (and perhaps https://github.com/openshift/odo/pull/1235)

PR #1338, opened 2019-02-15
Added suitable auto app switching in case of application deletion

### LINKED ISSUE #1331: Active context does not switch after deleting the active application
[kind/bug]

- Operating System: Fedora

1. Clone https://github.com/redhat-developer/odo
2. make bin
3. ./odo login -u developer -p password
4. ./odo project create testproject
6. ./odo app create test1
7. ./odo app create test2
8. ./odo app list 
```
$ odo app list
The project 'testproject' has the following applications:
ACTIVE     NAME
           test1
*          test2
```
9. ./odo app delete test2
10. ./odo app list
```
$ odo app list
The project 'testproject' has the following applications:
ACTIVE     NAME
           test1
```
## Actual behavior
Active context does not switch to other available app

## Expected behavior
Active context should switch
```
$ odo app list
The project 'testproject' has the following applications:
ACTIVE     NAME
*          test1
```

mik-dass:
fixes #1331 

How to test:
1. Create many apps
2. Delete the active one and check app switching

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -623,20 +623,34 @@ func (c *GlobalConfigInfo) DeleteApplication(application string, project string)
 		c.ActiveApplications = []ApplicationInfo{}
 	}
 
-	found := false
+	foundAt := -1
+	nextFoundAt := -1
+	isDeletedAppActive := false
 	for i, app := range c.ActiveApplications {
-		// if application exists set is as Active
+		// if application exists, save the index for deletion later and check if it is the active application
 		if app.Name == application && app.Project == project {
-			// remove current item from array
-			c.ActiveApplications = append(c.ActiveApplications[:i], c.ActiveApplications[i+1:]...)
-			found = true
+			isDeletedAppActive = app.Active
+			foundAt = i
+		}
+
+		// find the first other app in the same project
+		if app.Name != application && app.Project == project && nextFoundAt == -1 {
+			nextFoundAt = i
 		}
 	}
 
-	if !found {
+	if foundAt == -1 {
 		return fmt.Errorf("application %s doesn't exist", application)
 
 	}
+
+	// if the deleted app is the active application then set the next app in the project as active
+	if nextFoundAt != -1 && isDeletedAppActive {
+		c.ActiveApplications[nextFoundAt].Active = true
+	}
+
+	// remove current item from array with the found index of the item
+	c.ActiveApplications = append(c.ActiveApplications[:foundAt], c.ActiveApplications[foundAt+1:]...)
 
 	err := writeToFile(c.GlobalConfig, c.Filename)
 	if err != nil {
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -5,17 +5,16 @@ package e2e
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/redhat-developer/odo/pkg/config"
+	"log"
 	"os"
+	"time"
 
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


amitkrout:
Ping @geoand @metacosm @girishramnani for review.

@mik-dass Did you check it locally ? i am asking because no integration test is tracking the change. Would be great if you add the integration test for it or else create a follow up issue for integration test.

amitkrout:
@mik-dass Please address the issue #1340 in this pr too. Could be a similar kind of fix i guess.

mik-dass:
> i am asking because no integration test is tracking the change

Ok. But I am confused in which file to add this integration test. WDYT?

kadel:
> @mik-dass Please address the issue #1340 in this pr too. Could be a similar kind of fix i guess.

let's do this in separate PR.

Small PRs that are quick and easy to merge, that is our motto ;-)

pkg/config/config.go
```diff
@@ -555,21 +555,35 @@ func (c *GlobalConfigInfo) DeleteApplication(application string, project string)
...
 	}
 
-	if !found {
+	if foundAt == -1 {
 		return fmt.Errorf("application %s doesn't exist", application)
 
 	}
 
+	// if the deleted app is the active application then set the next app in the project as active
+	if nextFoundAt != -1 && isDeletedAppActive {
```
mohammedzee1000: +1 this should ensure that if application being deleted is the last application, then no active application is set

mik-dass:
> Can we have a test to ensure that if last application is being deleted, then there is no active application set.

Sure I will add that

syamgk:
@mik-dass for adding e2e test https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L70 this will be the right place 
apart from that it looks good to me

mik-dass:
> @mik-dass for adding e2e test https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L70 this will be the right place
apart from that it looks good to me

Thanks, I will have a look ASAP :)

tests/e2e/utils.go
```diff
@@ -74,3 +75,18 @@ func cleanUpAfterProjects(projects []string) {
 		odoDeleteProject(p)
 	}
 }
+
+// getActiveApplication returns the active application in the project
+// returns empty string if no active application is present in the project
+func getActiveApplication() string {
```
amitkrout: Lovely

tests/e2e/e2e_test.go
```diff
@@ -768,6 +767,38 @@ var _ = Describe("odoe2e", func() {
...
+
+			// delete in the active project
+			// the current app should switch
+			runCmdShouldPass("odo app delete -f")
+			Expect(getActiveApplication()).To(Equal("app-4"))
+
+			// deleting the last app in the active project
+			runCmdShouldPass("odo app delete -f")
+			Expect(getActiveApplication()).To(Equal(""))
+
```
amitkrout: You can add few steps to check in project ```-auto-0``` app-3 is still the active app.

For example
```
odo project -auto-0
Expect(getActiveApplication()).To(Equal("app-3"))
```

tests/e2e/utils.go
mik-dass: What are the changes needed here?

amitkrout: haha...
It went to wrong place ;)

tests/e2e/e2e_test.go
mik-dass: I have created a issue https://github.com/redhat-developer/odo/issues/1363 to track of it

amitkrout: Thanks for the issue.

PR #1339, opened 2019-02-15
Rationalize the use of --wait for link/unlink

### LINKED ISSUE #1337: wait flag does not work when linking components
When doing `odo link backend --wait` and `backend` is a component (instead of a service), then `wait` does not have any effect

geoand:
We now:
* Re-purpose `wait` to enforce waiting for the pod being updated to be
in running state (meaning that is can be immediately used by
other tools)
* Introduce `wait-for-target` that would do what wait does currently

Fixes #1337 

## How to test changes?
When using link/unlink with the `-w` flag should no wait until the component has been redeployed.
If the flag is not specified, the command returns immediately.

The old `-w` flag is now `--wait-for-target`. Use it when linking to a service to have odo wait until the service has been fully provisioned.

### DIFF
--- a/pkg/odo/cli/component/unlink.go
+++ b/pkg/odo/cli/component/unlink.go
@@ -82,7 +82,7 @@ func NewCmdUnlink(name, fullName string) *cobra.Command {
 	}
 
 	unlinkCmd.PersistentFlags().StringVar(&o.port, "port", "", "Port of the backend to which to unlink")
-	unlinkCmd.PersistentFlags().BoolVarP(&o.wait, "wait", "w", false, "If enabled the link will return only when the component is fully running after the link is created")
+	unlinkCmd.PersistentFlags().BoolVarP(&o.wait, "wait", "w", false, "If enabled the link will return only when the component is fully running after the link is deleted")
 
 	// Add a defined annotation in order to appear in the help menu
 	unlinkCmd.Annotations = map[string]string{"command": "component"}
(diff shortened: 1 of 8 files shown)


girishramnani:
@geoand do we need more documentation on this, change some examples maybe?

pkg/odo/cli/component/unlink.go
```diff
@@ -82,6 +82,7 @@ func NewCmdUnlink(name, fullName string) *cobra.Command {
 	}
 
 	unlinkCmd.PersistentFlags().StringVar(&o.port, "port", "", "Port of the backend to which to unlink")
+	unlinkCmd.PersistentFlags().BoolVarP(&o.wait, "wait", "w", false, "If enabled the link will return only when the component is fully running after the link is created")
```
kadel: Shouldn't this say: "If enabled the link will return only when the component is fully running after the link is **deleted**"?

geoand: Good catch! Fixed

geoand:
> @geoand do we need more documentation on this, change some examples maybe?

I don't really think so. The help info on the command should be good enough - plus now the `wait` flag is consistent with the rest of odo

kadel:
> > @geoand do we need more documentation on this, change some examples maybe?
> 
> I don't really think so. The help info on the command should be good enough - plus now the `wait` flag is consistent with the rest of odo

👍 I think that help message for the flag is clear and it should be enought

PR #1344, opened 2019-02-19
Fix typo.

metacosm:
See $SUBJECT

Nope.

## How to test changes?
Just a typo… but if you really want to test, you will need to deploy the example and look at it. :)


tests/e2e/examples/source/openjdk-sb-postgresql/src/main/resources/static/index.html
```diff
@@ -115,7 +115,7 @@
 <body ng-app="FruitManagement" ng-controller="FruitManagementController">
 
 <div class="container">
-    <h1>Spring Boot - CRUD DEmo</h1>
+    <h1>Spring Boot - CRUD Demo</h1>
```
amitkrout: +1

PR #1347, opened 2019-02-20
Modify components according to new structs

### LINKED ISSUE #1291: Modify Component to use new structs
- [ ] Add missing fields
- [ ] Update all functions that refer to Component struct to new structs (machine readable)

surajnarwade:
Refactor `Component` portion to use new structs throughtout and remove old Component struct and all it's usage
fixes #1291

## How to test changes?
CI should be :green_heart:  and test all Component related commands

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -28,22 +28,10 @@ import (
 // componentSourceURLAnnotation is an source url from which component was build
 // it can be also file://
 const componentSourceURLAnnotation = "app.kubernetes.io/url"
-const componentSourceTypeAnnotation = "app.kubernetes.io/component-source-type"
+const ComponentSourceTypeAnnotation = "app.kubernetes.io/component-source-type"
 const componentRandomNamePartsMaxLen = 12
 const componentNameMaxRetries = 3
 const componentNameMaxLen = -1
-
-// Description holds all information about component
-type Description struct {
-	ComponentName      string
-	ComponentImageType string
-	Path               string
-	URLs               urlpkg.UrlList
-	Env                []corev1.EnvVar
-	Storage            storage.StorageList
-	LinkedComponents   map[string][]string
-	LinkedServices     []string
-}
 
 // GetComponentDir returns source repo name
 // Parameters:
@@ -75,13 +63,13 @@ func GetComponentDir(path string, paramType occlient.CreateType) (string, error)
 // GetDefaultComponentName generates a unique component name
 // Parameters: desired default component name(w/o prefix) and slice of existing component names
 // Returns: Unique component name and error if any
-func GetDefaultComponentName(componentPath string, componentPathType occlient.CreateType, componentType string, existingComponentList []Description) (string, error) {
+func GetDefaultComponentName(componentPath string, componentPathType occlient.CreateType, componentType string, existingComponentList ComponentList) (string, error) {
 	var prefix string
 
 	// Get component names from component list
 	var existingComponentNames []string
-	for _, componentInfo := range existingComponentList {
-		existingComponentNames = append(existingComponentNames, componentInfo.ComponentName)
+	for _, component := range existingComponentList.Items {
+		existingComponentNames = append(existingComponentNames, component.Name)
 	}
 
 	// Fetch config
@@ -154,7 +142,7 @@ func CreateFromGit(client *occlient.Client, params occlient.CreateArgs) error {
 
 	// save source path as annotation
 	annotations := map[string]string{componentSourceURLAnnotation: params.SourcePath}
-	annotations[componentSourceTypeAnnotation] = "git"
+	annotations[ComponentSourceTypeAnnotation] = "git"
 
 	// Namespace the component
 	namespacedOpenShiftObject, err := util.NamespaceOpenShiftObject(params.Name, params.ApplicationName)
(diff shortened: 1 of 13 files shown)


girishramnani:
Is this WIP @surajnarwade

girishramnani:
Some of the UTs are failing @surajnarwade

pkg/component/component.go
```diff
@@ -880,17 +878,18 @@ func GetComponentDesc(client *occlient.Client, componentName string, application
...
-		LinkedComponents:   linkedComponents,
-		LinkedServices:     linkedServices,
-	}
-	return componentDesc, nil
+	currCompo, _ := GetCurrent(applicationName, projectName)
+
+	component = getMachineReadableFormat(componentName, componentType)
+	component.Spec.Source = path
+	component.Spec.URL = urls
+	component.Spec.Storage = storage
```
girishramnani: Shouldn't these also be set inside `getMachineReadableFormat` or some other method?

surajnarwade: so then method will need to provide too much inputs and codeclimate will cry :(

pkg/odo/cli/application/application.go
```diff
@@ -76,19 +77,20 @@ func printDeleteAppInfo(client *occlient.Client, appName string, projectName str
...
 
-		if len(componentDesc.URLs.Items) != 0 {
+		if len(componentDesc.Spec.URL) != 0 {
 			fmt.Println("  Externally exposed URLs will be removed")
 		}
 
-		for _, store := range componentDesc.Storage.Items {
-			fmt.Println("  Storage", store.Name, "of size", store.Spec.Size, "will be removed")
+		for _, store := range componentDesc.Spec.Storage {
+			storo := storage.GetStorage(client, store, currentComponent.Name, appName)
```
mik-dass: I guess we can use the ListMounted() method in the pkg/storage package and then iterate over the storage. It will save us extra API calls. Or am I missing something?

pkg/odo/util/cmdutils.go
```diff
@@ -65,45 +67,46 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
...
 	}
 	// URL
-	if componentDesc.URLs.Items != nil {
+	if componentDesc.Spec.URL != nil {
 		fmt.Println("\nURLs")
-		for _, componentUrl := range componentDesc.URLs.Items {
-
-			fmt.Printf(" - %v exposed via %v\n", url.GetURLString(componentUrl.Spec.Protocol, componentUrl.Spec.URL), componentUrl.Spec.Port)
+		for _, componentURL := range componentDesc.Spec.URL {
+			url := urlPkg.GetURL(client, componentURL, currentComponentName, applicationName)
```
mik-dass: I guess we can use list() directly and save some api calls

pkg/odo/cli/application/application.go
surajnarwade: it's one and the same, instead of writing there, I created separate function which can be reused in future. For example, this function can serve like GET request.

pkg/odo/util/cmdutils.go
surajnarwade: it's one and the same, instead of writing there, I created separate function which can be reused in future. For example, this function can serve like GET request.

pkg/odo/cli/application/application.go
```diff
@@ -76,18 +77,19 @@ func printDeleteAppInfo(client *occlient.Client, appName string, projectName str
...
-		log.Info("Component", currentComponent.ComponentName, "will be deleted.")
+		log.Info("Component", currentComponent.Name, "will be deleted.")
 
-		if len(componentDesc.URLs.Items) != 0 {
+		if len(componentDesc.Spec.URL) != 0 {
 			fmt.Println("  Externally exposed URLs will be removed")
 		}
-
-		for _, store := range componentDesc.Storage.Items {
+		storages, err := storage.List(client, currentComponent.Name, appName)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/component/component.go
girishramnani: Cool

pkg/odo/cli/application/application.go
surajnarwade: solved

pkg/url/url.go
```diff
@@ -17,6 +17,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// GetURL returns URL defination for given URL name
+func GetURL(urlName string, urls UrlList) Url {
```
kadel: This should be called just `Get`

pkg/storage/storage.go
```diff
@@ -16,6 +16,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+// GetStorage returns Storage defination for given Storage name
+func GetStorage(storageName string, storages StorageList) Storage {
```
kadel: This should be called just `Get`

kadel: This looks weird. 
It might be better to have `StorageList` as a receiver. What do you think?  `func (sl *StorageList) Get(name string) Storage`

kadel:
Just a few small pointers, in general this looks good

pkg/storage/storage.go
surajnarwade: looks good to me

PR #1348, opened 2019-02-20
Avoid use of odo core package in testsuite

amitkrout:

Odo core package should not be the part of test suite as it brings dependency in test and more importantly user specific scenario will be compromised. 

No

## How to test changes?

make test-main-e2e

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"testing"
 	"time"
-
-	"github.com/redhat-developer/odo/pkg/config"
 )
 
 // TODO: A neater way to provide odo path. Currently we assume \
@@ -86,11 +84,11 @@ var _ = Describe("odoe2e", func() {
 	})
 
 	Context("odo utils config", func() {
-		It("should get blank for updatenotification by default globally as its not set", func() {
+		It("should get globally availaible default parameters", func() {
 			configOutput := runCmdShouldPass("odo utils config view --global")
-			Expect(configOutput).To(ContainSubstring(config.UpdateNotificationSetting))
-			Expect(configOutput).To(ContainSubstring(config.NamePrefixSetting))
-			Expect(configOutput).To(ContainSubstring(config.TimeoutSetting))
+			Expect(configOutput).To(ContainSubstring("UpdateNotification"))
+			Expect(configOutput).To(ContainSubstring("NamePrefix"))
+			Expect(configOutput).To(ContainSubstring("Timeout"))
 		})
 		It("should be checking to see if timeout is shown as blank globally as its not set", func() {
 			configOutput := runCmdShouldPass("odo utils config view --global|grep Timeout")


metacosm:
could you elaborate why importing constants compromising user scenarios?

kadel:
I'm not sure about this. If we change those constants and forgot to change it in the tests, it will start to fail. Is this something that we want?  To me, it looks like adding an unnecessary burden on updating tests.

metacosm:
> I'm not sure about this. If we change those constants and forgot to change it in the tests, it will start to fail. Is this something that we want? To me, it looks like adding an unnecessary burden on updating tests.

That's why I'd like to know why using constants is considered bad…

amitkrout:
> could you elaborate why importing constants compromising user scenarios?

@metacosm Consider this
```
$ odo utils config view --global
UpdateNotification    false
NamePrefix            
Timeout               
$ 
```
As a user i will always look for user facing parameter values like seen above ```UpdateNotification```, ```UpdateNotification``` because i am running my stuff on odo binary and will not care about what is there in the odo code, So the integration test should be done with user prospective and code level verification should be done in unit test. 

So if we are verifying the result using the core package then we are completely ignoring the real output of ```odo utils config view --global```.

amitkrout:
> If we change those constants and forgot to change it in the tests, it will start to fail. Is this something that we want? To me, it looks like adding an unnecessary burden on updating tests.

@kadel Yes, This is how we should test user facing parameters. We suppose to do verify the final output in integration test as its run on the odo binary, not the constant being used in the care package.

metacosm:
> > If we change those constants and forgot to change it in the tests, it will start to fail. Is this something that we want? To me, it looks like adding an unnecessary burden on updating tests.
> 
> @kadel Yes, This is how we should test user facing parameters. We suppose to do verify the final output in integration test as its run on the odo binary, not the constant being used in the care package.

I don't agree. The behavior we're testing is that the command shows the available parameters. The parameter names are only relevant if we want to check that the "API" (i.e. externally visible behavior) doesn't change. Are the names of the settings supposed to be stable? If yes, maybe that should be a separate test.

amitkrout:
> > > If we change those constants and forgot to change it in the tests, it will start to fail. Is this something that we want? To me, it looks like adding an unnecessary burden on updating tests.
> > 
> > 
> > @kadel Yes, This is how we should test user facing parameters. We suppose to do verify the final output in integration test as its run on the odo binary, not the constant being used in the care package.
> 
> I don't agree. The behavior we're testing is that the command shows the available parameters. The parameter names are only relevant if we want to check that the "API" (i.e. externally visible behavior) doesn't change. Are the names of the settings supposed to be stable? If yes, maybe that should be a separate test.

True, But how to confirm that it really reflects on the terminal. IMO by directly verifying terminal parameters we can confirm that the changing setting names also reflect in the terminal output too.

girishramnani:
@amitkrout I feel this could potentially increase our work and defeats the purpose of having these constants in the first place

amitkrout:
> @amitkrout I feel this could potentially increase our work and defeats the purpose of having these constants in the first place

@girishramnani Are these constant name keeps changing on each time we run utils config command, No right ? Except in scenario ```It("should get blank for updatenotification by default globally as its not set", func()
{}``` none of the other scenario use same ```config.<parameter>``` validation even on ```odo utils config view``` operation in e2e_test.go file. 

Check  - https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L99-L171

amitkrout:
ping @metacosm @kadel @girishramnani

kadel:
> > could you elaborate why importing constants compromising user scenarios?
> 
> @metacosm Consider this
> 
> ```
> $ odo utils config view --global
> PARAMETER             CURRENT_VALUE
> UpdateNotification    false
> NamePrefix            
> Timeout               
> $ 
> ```
> As a user i will always look for user facing parameter values like seen above `UpdateNotification`, `UpdateNotification` because i am running my stuff on odo binary and will not care about what is there in the odo code, So the integration test should be done with user prospective and code level verification should be done in unit test.
> 
> So if we are verifying the result using the core package then we are completely ignoring the real output of `odo utils config view --global`.

In any other case, I would agree with you, you are right. But this case is a little bit different. We are testing that the user sees the config parameters. I think that in this case it make sense to use constants.

But in the next test we are using strings instead of contanants, so not sure if we are not overthinknig it. :-D 

I'm not going to block this PR, neitehr I'm going to approve it 😇 If you get two approvals feel free to merge ;-)

amitkrout:
> > > could you elaborate why importing constants compromising user scenarios?
> > 
> > 
> > @metacosm Consider this
> > ```
> > $ odo utils config view --global
> > PARAMETER             CURRENT_VALUE
> > UpdateNotification    false
> > NamePrefix            
> > Timeout               
> > $ 
> > ```
> > As a user i will always look for user facing parameter values like seen above `UpdateNotification`, `UpdateNotification` because i am running my stuff on odo binary and will not care about what is there in the odo code, So the integration test should be done with user prospective and code level verification should be done in unit test.
> > So if we are verifying the result using the core package then we are completely ignoring the real output of `odo utils config view --global`.
> 
> In any other case, I would agree with you, you are right. But this case is a little bit different. We are testing that the user sees the config parameters. I think that in this case it make sense to use constants.
> 
> But in the next test we are using strings instead of contanants, so not sure if we are not overthinknig it. :-D
> 
> I'm not going to block this PR, neitehr I'm going to approve it 😇 If you get two approvals feel free to merge ;-)

No issue :)

Let see how other team member see it

amitkrout:
ping @mik-dass @geoand @surajnarwade @mohammedzee1000

geoand:
I don't agree with the proposed changes. @metacosm and @kadel have articulated very well and I agree with them.

cdrage:
I don't believe in the required changes either. Just created more workload for us. Having us use the imported constants is viable. 

There's 5 of us that disagree. I'm going to close this PR.

amitkrout:
> I don't believe in the required changes either. Creates more workload for us in order to maintain the test suite. Having us use the imported constants is viable (for now!)

e2e tests are written to validate the Odo binary not the source code. So IMO it is very bad idea to import the packages from code which are directly affecting the user facing output. 

For example:
In the source you made a config variable typo (let ```Timeoute```) then this won't be caught in the test suite because you are accepting whatever the value you passed but not validating the real output.  

I think you missed this code - 
Check - https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e_test.go#L99-L171

In the above link there is no use of imported core package to do the same job. Then really i do not understand where is point of maintenance issue arises.

> There's 5 of us that disagree. I'm going to close this PR (no harsh feelings!). We can re-open this / re-visit this if there are any issues in the future.

This is a necessary change atm i guess, So think it again

PR #1358, opened 2019-02-22
Fix #1356: make sure that paths are absolute

### LINKED ISSUE #1356: odo create in interactive mode doesn't support relative paths
[kind/bug]

In odo latest, when you run in interactive mode and set the binary location as a relative path, the path set in the annotation is wrong, as it's written as absolute.

Using target/parkasmap-web.jar in interactive I get this annotation in the container:

```
app.kubernetes.io/url: 'file://target/parksmap-web.jar'
```

Using the same path in non-interactive, I get the following annotation:

```
app.kubernetes.io/url: 'file:///Users/jmorales/hack_and_sangria/frontend/target/parksmap-web.jar'
```

cc/ @geoand

metacosm:
`odo create` might issue improper paths when creating components.

fixes #1356 

## How to test changes?
`odo create` in both interactive and non-interactive mode should work properly.


metacosm:
@jorgemoralespou let us know if it works fine now.

Issue #1352, opened 2019-02-21
make get command available for odo config


amitkrout:
[kind/Enhancement]

## Which functionality do you think we should update/improve?
```odo config get <PARAMETER>``` should return the ```CURRENT_VALUE``` assigned to it

## Why is this needed? 
Better user experience and ofcourse it makes the validation easy for a particular ```PARAMETER``` in the test file.

amitkrout:
ping @girishramnani , Please share your thoughts. 

cc_ @kadel

girishramnani:
@amitkrout as part of the config changes, the `odo utils config` will be moved to `odo config` so maybe there we can consider the `get`

girishramnani:
@tkral should decide if this is needed

kadel:
there is `odo config view` that lists all the configuration values. I think that this is enough for now. 
Let's wait if there is a real need `get` before adding it.

girishramnani:
closing this issue as we haven't seen any request for this feature so far. If someone feels this feature is need then they can re-open this issue

PR #1324, opened 2019-02-13
Use of re package for string operation in testsuite

amitkrout:
## Was the change discussed in an issue?
fixes #1129

## How to test changes?
CI should go green or run test locally one by one through ```make test-*```

PR #1362, opened 2019-02-25
broke out the preferences and config

girishramnani:

Breaks down whole config into the preferences and config
partly https://github.com/redhat-developer/odo/issues/1294

## How to test changes?

```
odo preference set nameprefix test
odo config set minmemory 500MB
odo config unset cpu
odo preference unset nameprefix

```


pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+		return errors.Wrap(err, "unable to write configuration file")
+	}
+	return nil
+}
+
+// GetActiveComponent if no component is set as current returns empty string
+func (c *GlobalConfigInfo) GetActiveComponent(application string, project string) string {
+	if c.ActiveApplications != nil {
+		for _, app := range c.ActiveApplications {
+			if app.Project == project && app.Name == application && app.Active == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `app.Active` (from `gosimple`)

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	return ""
+}
+
+// GetActiveApplication get currently active application for given project
+// if no application is active return empty string
+func (c *GlobalConfigInfo) GetActiveApplication(project string) string {
+	if c.ActiveApplications != nil {
+		for _, app := range c.ActiveApplications {
+			if app.Project == project && app.Active == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `app.Active` (from `gosimple`)

metacosm:
Please fix the CodeClimate doc issues. I'll review after rebase.

tests/e2e/e2e_test.go
```diff
@@ -85,24 +85,24 @@ var _ = Describe("odoe2e", func() {
...
+			configOutput := runCmdShouldPass("odo preference view |grep Timeout")
 			Expect(configOutput).To(ContainSubstring(fmt.Sprintf("Timeout")))
 		})
 		It("should be checking to see if global config values are the same as the configured ones", func() {
-			runCmdShouldPass("odo utils config set --global updatenotification false")
-			runCmdShouldPass("odo utils config set --global timeout 5")
-			configOutput := runCmdShouldPass("odo utils config view --global |grep UpdateNotification")
+			runCmdShouldPass("odo preference set updatenotification false")
+			runCmdShouldPass("odo preference set timeout 5")
+			configOutput := runCmdShouldPass("odo preference view |grep UpdateNotification")
```
amitkrout: To avoid terminal specific command ```grep```. Use function like

```
func findLocalConfigValueOfGivenKey(key string) string {

	stdOut, _, _ := cmdRunner("odo config view")
	re := regexp.MustCompile(key + `.+`)
	odoConfigKeyValue := re.FindString(stdOut)
	if odoConfigKeyValue == "" {
		return fmt.Sprintf("%s not found", key)
	}
	trimKeyValue := strings.TrimSpace(odoConfigKeyValue)
	if strings.Compare(key, trimKeyValue) != 0 {
		return strings.TrimSpace(strings.SplitN(trimKeyValue, " ", 2)[1])
	}
	return ""
}

func findGlobalConfigValueOfGivenKey(key string) string {
	stdOut, _, _ := cmdRunner("odo preference view")
	re := regexp.MustCompile(key + `.+`)
	odoConfigKeyValue := re.FindString(stdOut)
	if odoConfigKeyValue == "" {
		return fmt.Sprintf("%s not found", key)
	}
	trimKeyValue := strings.TrimSpace(odoConfigKeyValue)
	if strings.Compare(key, trimKeyValue) != 0 {
		return strings.TrimSpace(strings.SplitN(trimKeyValue, " ", 2)[1])
	}
	return ""
}
```

amitkrout: Add this function in utils.go file in testsuite and call it in the test file

amitkrout:
@girishramnani Fix reported issue in [1] if reasonable.

[1] https://github.com/redhat-developer/odo/pull/1362#issuecomment-466928932

amitkrout:
@girishramnani Are you planning to add #1352 in this pr ?

tests/e2e/e2e_test.go
kadel: This should be probably handled in different PR.

kadel:
> @girishramnani Are you planning to add #1352 in this pr ?

you can use `odo config view` and `odo preferences view` for now. We can add `get` later in another PR

tests/e2e/e2e_test.go
amitkrout: No issue, Anyway i am handling it in a separate pr #1324

amitkrout:
> > @girishramnani Are you planning to add #1352 in this pr ?
> 
> you can use `odo config view` and `odo preferences view` for now. We can add `get` later in another PR

Ok, Works for me

pkg/odo/cli/preference/set.go
```diff
@@ -0,0 +1,109 @@
...
+			util.LogErrorAndExit(fmt.Errorf("'%s' is not a parameter in the odo config", o.paramName), "")
+		}
+	}
+
+	err = cfg.SetConfiguration(strings.ToLower(o.paramName), o.paramValue)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Preference was successfully updated.")
```
metacosm: Shouldn't we use `log` here rather than `fmt`?

pkg/odo/cli/preference/set.go
```diff
@@ -0,0 +1,109 @@
...
+// Run contains the logic for the command
+func (o *SetOptions) Run() (err error) {
+
+	cfg, err := preference.New()
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to set preference")
+	}
+
+	if !o.configForceFlag {
```
metacosm: This should probably be in `Validate` rather than `Run` actually.

pkg/odo/cli/preference/unset.go
```diff
@@ -0,0 +1,108 @@
...
+// Run contains the logic for the command
+func (o *UnsetOptions) Run() (err error) {
+
+	cfg, err := preference.New()
+
+	if err != nil {
+		return errors.Wrapf(err, "")
+	}
+
+	if value, ok := cfg.GetConfiguration(o.paramName); ok && (value != nil) {
```
metacosm: I would put this code in `Validate` and use the same logic as used in `Set` to make things simpler to understand (since the logic is fundamentally the same) and the logic here is more difficult to follow, in my opinion.

pkg/odo/cli/preference/unset.go
```diff
@@ -0,0 +1,108 @@
...
+// NewCmdUnset implements the preference unset odo command
+func NewCmdUnset(name, fullName string) *cobra.Command {
+	o := NewUnsetOptions()
+	preferenceUnsetCmd := &cobra.Command{
+		Use:     name,
+		Short:   "Unset a value in odo preference file",
+		Long:    fmt.Sprintf(unsetLongDesc, preference.FormatSupportedParameters()),
+		Example: fmt.Sprintf(fmt.Sprint("\n", unsetExample), fullName, preference.UpdateNotificationSetting, preference.NamePrefixSetting, preference.TimeoutSetting),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
```
metacosm: Why not use `cobra.MaximumNArgs(1)` here instead of a custom function (though the custom function provides a better error message)?

pkg/odo/cli/preference/unset.go
```diff
@@ -0,0 +1,108 @@
...
+			} else if len(args) > 1 {
+				return fmt.Errorf("only one parameter is allowed")
+			} else {
+				return nil
+			}
+
+		}, Run: func(cmd *cobra.Command, args []string) {
+			util.LogErrorAndExit(o.Complete(name, cmd, args), "")
+			util.LogErrorAndExit(o.Validate(), "")
+			util.LogErrorAndExit(o.Run(), "")
```
metacosm: Use `genericclioptions.GenericRun(o, cmd, args)` here instead.

pkg/odo/cli/preference/unset.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+	preferenceUnsetCmd.Flags().BoolVarP(&o.preferenceForceFlag, "force", "f", false, "Dont ask for confirmation, directly move forward")
```
metacosm: I'd rather have a more explicit explanation, related to the specific command: "Don't ask for confirmation, unsetting the preference directly" or something similar.

pkg/odo/cli/preference/view.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/spf13/cobra"
+	ktemplates "k8s.io/kubernetes/pkg/kubectl/cmd/templates"
+)
+
+const viewCommandName = "view"
+
+var viewExample = ktemplates.Examples(`# For viewing the current local preference
+   %[1]s
+
+   For viewing the current global preference
```
metacosm: Missing `#`.

pkg/odo/cli/utils/utils.go
```diff
@@ -15,19 +13,17 @@ const RecommendedCommandName = "utils"
 // NewCmdUtils implements the utils odo command
 func NewCmdUtils(name, fullName string) *cobra.Command {
 	terminalCmd := NewCmdTerminal(terminalCommandName, util.GetFullName(fullName, terminalCommandName))
-	configurationCmd := config.NewCmdConfiguration(config.RecommendedCommandName, util.GetFullName(fullName, config.RecommendedCommandName))
 	utilsCmd := &cobra.Command{
 		Use:   name,
 		Short: "Utilities for terminal commands and modifying Odo configurations",
 		Long:  `Utilities for terminal commands and modifying Odo configurations`,
```
metacosm: This needs to be changed. Also, maybe it would make sense to remove the `utils` command altogether if it only has one sub-command?

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
+package preference
```
metacosm: Maybe `preferences` instead?

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+		TimeoutSetting:            TimeoutSettingDescription,
+	}
+
+	// set-like map to quickly check if a parameter is supported
+	lowerCaseParameters = util.GetLowerCaseParameters(GetSupportedParameters())
+)
+
+// GlobalConfigInfo wraps the global config and provides helpers to
+// serialize it.
+type GlobalConfigInfo struct {
```
metacosm: Wouldn't it make sense to rename this to `PreferencesInfo`?

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+	Active bool `yaml:"Active"`
+	// name of the openshift project this application belongs to
+	Project string `yaml:"Project"`
+	// last active component for  this application
+	ActiveComponent string `yaml:"ActiveComponent"`
+}
+
+// GlobalConfig stores all the config related to odo, its the superset of
+// local config.
+type GlobalConfig struct {
```
metacosm: Rename this to `Preferences`?

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	err := util.WriteToFile(c.GlobalConfig, c.Filename)
+	if err != nil {
+		return errors.Wrap(err, "unable to set current application")
+	}
+	return nil
+}
+
+// AddApplication add  new application to the config file
```
metacosm: add -> adds (plus extra space to remove)

pkg/preference/preference.go
```diff
@@ -0,0 +1,436 @@
...
+// Newly create application is NOT going to be se as Active.
```
metacosm: create -> created

pkg/util/config_util.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// GetFromFile unmarshals a struct from a odo config file
+func GetFromFile(c interface{}, filename string) error {
+	configData, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return errors.Wrapf(err, "unable to read file %v", filename)
+	}
+
+	err = yaml.Unmarshal(configData, c)
```
metacosm: Looks like there is an asymmetry between Get and Write that I don't like. Doesn't this assume that `c` is a pointer here? While, in `WriteToFile`, `c` is assumed to be a `struct` since we take its address? If we keep the asymmetry shouldn't we check that `c` is indeed of the proper type (pointer or not) at least?

pkg/preference/preference.go
girishramnani: We want to keep stuff singular

pkg/preference/preference.go
girishramnani: Least amount of change, we will again change all this hence not changed it

pkg/odo/cli/config/config.go
```diff
@@ -38,6 +38,7 @@ func NewCmdConfiguration(name, fullName string) *cobra.Command {
 	configurationCmd.AddCommand(configurationViewCmd, configurationSetCmd)
 	configurationCmd.AddCommand(configurationUnsetCmd)
 	configurationCmd.SetUsageTemplate(util.CmdUsageTemplate)
+	configurationCmd.Annotations = map[string]string{"command": "component"}
```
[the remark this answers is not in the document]
kadel: yes, it manges component configuration

pkg/application/application_test.go
```diff
@@ -5,34 +5,34 @@ import (
...
 		wantRE       string
 		needPrefix   bool
 		prefix       string
 	}{
 		{
 			testName:     "Case: App prefix not configured",
-			existingApps: []config.ApplicationInfo{},
-			wantRE:       "app-*",
+			existingApps: []preference.ApplicationInfo{},
+			wantRE:       "application-*",
```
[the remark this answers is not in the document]
kadel: good catch, this is weird, the default application name should be `app`

pkg/preference/preference.go
kadel: Why we will change all this again?

girishramnani: I was going for least amount of change, I can change this if you feel?

metacosm: @kadel we don't have to but if global config is named preferences now, it makes sense to change the associated types as well…

pkg/odo/cli/preference/unset.go
girishramnani: assuming that could be the reason, do you want me to change it?

pkg/odo/cli/utils/utils.go
girishramnani: @kadel would love your opinion here

pkg/odo/cli/config/unset.go
```diff
@@ -21,29 +22,23 @@ var (
...
+   %[1]s %[2]s
+   %[1]s %[3]s 
+   %[1]s %[4]s  
+   %[1]s %[5]s 
    %[1]s %[6]s 
    %[1]s %[7]s  
-   %[1]s %[8]s 
-   %[1]s %[9]s 
+   %[1]s %[8]s  
+   %[1]s %[9]s
```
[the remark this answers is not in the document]
girishramnani: resolved

pkg/odo/cli/preference/unset.go
girishramnani: resolved

pkg/odo/cli/preference/view.go
girishramnani: resolved

pkg/preference/preference.go
girishramnani: resolved

pkg/util/config_util.go
girishramnani: resolved, now both assume pointers

pkg/preference/preference.go
girishramnani: resolved

pkg/odo/cli/preference/unset.go
girishramnani: resolved

pkg/odo/cli/preference/set.go
girishramnani: resolved

pkg/odo/cli/utils/utils.go
kadel: I would keep it for now. We can remove it later.

girishramnani:
@tkral trying to resolve this failure https://travis-ci.org/redhat-developer/odo/jobs/499683537

pkg/preference/preference.go
kadel: I would prefer to change it now. It doesn't really make sense to rename it only in some places and leave the rest. We we should remove all references to global config and use just preference

pkg/application/application_test.go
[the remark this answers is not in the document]
girishramnani: from my assumption of the case bellow this https://github.com/redhat-developer/odo/pull/1362/files#diff-3378c0b9455d2a824a2de1831c18cef4L35, I thought this is an old test which was failing and hence I updated it.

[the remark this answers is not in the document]
girishramnani: I thought that prefix = "" and prefix = nil have same behavior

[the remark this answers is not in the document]
kadel: > I thought that `prefix = ""` and `prefix = nil` have same behavior

how was it working before? I'm confused now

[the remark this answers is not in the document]
kadel: "" and `nil` is not the same, and it should't be theraetd as the same. "" means that it was explicitly set to empty string, the `nil` means that it was not set and the default value should be used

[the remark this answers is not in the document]
kadel: hmm, by according this they should behave the same https://github.com/redhat-developer/odo/blob/f8a4df2557710809cd603fce33ca0c16ad693ec3/pkg/application/application.go#L39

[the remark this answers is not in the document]
girishramnani: Yeah hence the change @tkral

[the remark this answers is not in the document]
girishramnani: I said that the behavior for them of the application is same not that they are same @tkral

tests/e2e/e2e_test.go
girishramnani: resolving this

pkg/preference/preference.go
girishramnani: renamed to prefrenceInfo

pkg/odo/cli/preference/unset.go
girishramnani: keeping it the same as of now because it gives a better error message

pkg/application/application_test.go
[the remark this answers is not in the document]
girishramnani: @kadel the `prefix` comes from the https://github.com/redhat-developer/odo/blob/a9b145798ae10971b974c3af88bdd827a612dbe7/pkg/application/application.go#L40 which depends on the current folder directory name which is `application` hence the prefix being `application-`

pkg/odo/cli/preference/set.go
girishramnani: but I am a bit confused how would we short circuit the validate in case of 
```
	if value, ok := cfg.GetConfiguration(o.paramName); ok && (value != nil) {
			fmt.Printf("%v is already set. Current value is %v.\n", o.paramName, value)
			if !ui.Proceed("Do you want to override it in the config") {
				fmt.Println("Aborted by the user.")
				return nil
			}
		} else if !ok {
			util.LogErrorAndExit(fmt.Errorf("'%s' is not a parameter in the odo config", o.paramName), "")
```
`as aborted by the user` is technically not an error?

pkg/application/application_test.go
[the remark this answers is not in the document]
girishramnani: https://play.golang.org/p/ZxqzoH95dW0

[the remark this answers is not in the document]
girishramnani: the tests for `app-` weren't failing, it would be my mistake. reverting this change

pkg/odo/cli/preference/unset.go
girishramnani: used the same logic as set but still having the same query as https://github.com/redhat-developer/odo/pull/1362#discussion_r

girishramnani:
Can we do that in a separate PR, this PR is blocking people and as it's
already huge. Let's add more stuff in a different PR. @kadel

kadel:
> Can we do that in a separate PR, this PR is blocking people and as it's already huge. Let's add more stuff in a different PR. @kadel

As long as all existing test was updated and no test was removed I'm ok with adding more test in separate PR. @girishramnani Can you please open issues as a reminder that we need to do this later?

girishramnani:
Opened an issue for the tests @kadel @amitkrout  https://github.com/redhat-developer/odo/issues/1414

pkg/odo/cli/preference/set.go
kadel: @metacosm Not sure about that. Validate should just make sure that input is valid no?
If I understand this correctly, here @girishramnani is just acting based on a provided flag, but not doing any validation.

metacosm: While it is validation code (checking that the param indeed exists), there is indeed no way to bypass the rest of the action in the context of `Validate` so 👍 to stay in `Run`. However, you should use `log` and not `fmt` for output, @girishramnani.

girishramnani: Should I do that for all fmt.Println for consistency? E.g. https://github.com/redhat-developer/odo/pull/1362/files#diff-b533ce52cc14b27228695f599794930bR84

girishramnani:
@metacosm done, changed the fmt occurences to log for output

PR #1385, opened 2019-02-27
[WIP] Fixes the problem of auto switching active app when switching project

### LINKED ISSUE #1363: Active app doesn't switch when switching project

`odo project delete`

## Why is this needed? 

When the project is delete and odo sets an existing project as active, one of the apps from the project should be set as active automatically

mik-dass:
fixes #1363 

## How to test changes?
1. Create multiple projects
2. Delete a project and check if the first app in the switched project is set as active or not
3. Set a project as active and check if the first app in it is set as active or not


mik-dass:
@girishramnani yeah since app activeness depends upon the current directory, closing this

PR #1404, opened 2019-03-01
Added json output for storage create

### LINKED ISSUE #1379: machine readable output: odo storage create
`odo storage create -o json` should return JSON formatted output describing the newly created storage.

mik-dass:
fixes #1379 
## How to test changes?

run `odo storage` ommand with `-o json`

### DIFF
--- a/pkg/odo/cli/storage/create.go
+++ b/pkg/odo/cli/storage/create.go
@@ -1,7 +1,6 @@
 package storage
 
 import (
-	"encoding/json"
 	"fmt"
 	"github.com/redhat-developer/odo/pkg/log"
 	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
@@ -29,7 +28,6 @@ type StorageCreateOptions struct {
 	storageName string
 	storageSize string
 	storagePath string
-	outputFlag  string
 	*genericclioptions.Context
 }
 
@@ -51,6 +49,10 @@ func (o *StorageCreateOptions) Complete(name string, cmd *cobra.Command, args []
 
 // Validate validates the StorageCreateOptions based on completed values
 func (o *StorageCreateOptions) Validate() (err error) {
+	// check the machine readable format
+	if !util.CheckOutputFlag(o.OutputFlag) {
+		return fmt.Errorf("given output format %s is not supported", o.OutputFlag)
+	}
 	// validate storage path
 	return validateStoragePath(o.Client, o.storageName, o.Component(), o.Application)
 }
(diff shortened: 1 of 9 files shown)


pkg/odo/cli/storage/create.go
```diff
@@ -55,9 +57,17 @@ func (o *StorageCreateOptions) Validate() (err error) {
 
 // Run contains the logic for the odo storage create command
 func (o *StorageCreateOptions) Run() (err error) {
-	_, err = storage.Create(o.Client, o.storageName, o.storageSize, o.storagePath, o.Component(), o.Application)
+	storageResult, err := storage.Create(o.Client, o.storageName, o.storageSize, o.storagePath, o.Component(), o.Application)
 	if err == nil {
```
kadel: what is there was an error? No error output?

mik-dass: It will call and return the error to `util.LogErrorAndExit(o.Run(), "")` in GenericRun() in runnable.go

So it will log the error and and exit

pkg/odo/cli/storage/create.go
```diff
@@ -55,9 +57,17 @@ func (o *StorageCreateOptions) Validate() (err error) {
...
-		log.Successf("Added storage %v to %v", o.storageName, o.Component())
+		if o.outputFlag == "json" {
+			out, err := json.Marshal(storageResult)
+			if err != nil {
+				return err
+			}
+			fmt.Println(string(out))
+		} else {
```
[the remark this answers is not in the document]
mik-dass: The if statement will be abstracted in https://github.com/redhat-developer/odo/pull/1395. Should I wait?

cdrage:
Hey @surajnarwade could we get your PR / code here merged in: https://github.com/redhat-developer/odo/pull/1395/files#diff-6c6a7fefe37261e2fa8a010cf9e7ae08R537 so that @mik-dass can use your validation code with this PR?

pkg/storage/types.go
```diff
@@ -29,4 +27,6 @@ type StorageList struct {
 // StorageStatus is status of storage
 type StorageStatus struct {
 	Mounted bool `json:"mounted"`
+	// if path is empty, it indicates that the storage is not mounted in any component
```
kadel: Do we still need `Mounted` than?

mik-dass: I don't think of any case where it will be required since path will indicate the state. I leave the decision up to you. Should I remove it?

kadel: Yes, if it doesn't make sense, please remove it.

mik-dass:
@kadel Fixed

PR #1406, opened 2019-03-01
Do not error out if a service instance status cannot be determined

metacosm:
Mark it as `Unknown` instead.

No.

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -186,10 +186,15 @@ func List(client *occlient.Client, applicationName string) ([]ServiceInfo, error
 	// Iterate through serviceInstanceList and add to service
 	for _, elem := range serviceInstanceList {
 		conditions := elem.Status.Conditions
+		var status string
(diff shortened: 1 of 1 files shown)


pkg/service/service.go
```diff
@@ -264,21 +269,6 @@ func updateStatusIfMatchingDeploymentExists(dcs []appsv1.DeploymentConfig, secre
 	}
 }
 
-// GetSvcByType returns the matching (by type) service or nil of there are no matches
-func GetSvcByType(client *occlient.Client, serviceType string) (*occlient.Service, error) {
```
girishramnani: +1 this function is not used anywhere

amitkrout:
@metacosm Can you please add the steps for how to test

kadel:
Appveyor test passed, we just exceeded artifact storage quota :-(

girishramnani:
@kadel  can we clean up the old artifacts?

PR #1435, opened 2019-03-06
Remove terminal specific command from e2e watch test file

amitkrout:

Update the test file to run it on all oc supported platform including windows
fixes Part of #1129

## How to test changes?

make test-watch-e2e

### DIFF
--- a/tests/e2e/utils.go
+++ b/tests/e2e/utils.go
@@ -126,3 +126,24 @@ func matchResponseSubString(url, match string, retry, sleep int) bool {
 	fmt.Printf("Could not get the match string \"%s\" in %d seconds\n", match, i)
 	return false
 }
+
+// Creates a directory in the path specified. It takes one argument
+// dirNameFullpath (string) - Directory name along with the path
+func createDir(dirNameFullpath string) error {
+	return os.MkdirAll(dirNameFullpath, 0777)
+}
+
+// Creates a file in a specified path. It takes two argument
+// filePath (string) - The path where you want to create the file
+// fileName (string) - Name of the file
+func createFileAtPath(filePath string, fileName string) error {
+	_, err := os.Stat(filePath + "/" + fileName)
+	if os.IsNotExist(err) {
+		file, err := os.Create(filePath + "/" + fileName)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+	}
+	return nil
+}
--- a/tests/e2e/watch_test.go
+++ b/tests/e2e/watch_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


tests/e2e/utils.go
```diff
@@ -126,3 +126,24 @@ func matchResponseSubString(url, match string, retry, sleep int) bool {
...
+// dirNameFullpath (string) - Directory name along with the path
+func createDir(dirNameFullpath string) error {
+	return os.MkdirAll(dirNameFullpath, 0777)
+}
+
+// Creates a file in a specified path. It takes two argument
+// filePath (string) - The path where you want to create the file
+// fileName (string) - Name of the file
+func createFileAtPath(filePath string, fileName string) error {
+	_, err := os.Stat(filePath + "/" + fileName)
```
kadel: instead of concatenating strings with `/` you should use `filepath.Join`, this will handle path properly on different platforms like windows

tests/e2e/utils.go
```diff
@@ -126,3 +126,24 @@ func matchResponseSubString(url, match string, retry, sleep int) bool {
...
+	if os.IsNotExist(err) {
+		file, err := os.Create(filePath + "/" + fileName)
```
kadel: maybe save the full path to a variable so you don't have to create it twice?

tests/e2e/watch_test.go
```diff
@@ -57,16 +58,18 @@ var _ = Describe("odoWatchE2e", func() {
...
 					)
 					fmt.Println("Received signal, starting file modification simulation")
-					runCmdShouldPass("mkdir -p " + tmpDir + "/nodejs-ex" + "/'.a b'")
-					runCmdShouldPass("mkdir -p " + tmpDir + "/nodejs-ex" + "/'a b'")
-					runCmdShouldPass("touch " + tmpDir + "/nodejs-ex" + "/'a b.txt'")
-
-					runCmdShouldPass(fmt.Sprintf("mkdir -p %s/nodejs-ex/tests/sample-tests", tmpDir))
-					runCmdShouldPass(fmt.Sprintf("touch %s/nodejs-ex/tests/sample-tests/test_1.js", tmpDir))
+					fmt.Println("Received signal, starting file modification simulation")
+					Expect(createDir(tmpDir + "/nodejs-ex" + "/.a")).To(BeNil())
```
kadel: instead of concatenating strings with `/` you should use `filepath.Join`, this will handle path properly on different platforms like windows

tests/e2e/utils.go
amitkrout: Thanks for the comment. I will consider the suggestion for sure, however i am getting a panic out message when trying locally 
```
[...]
panic: 
Your test failed.
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.
[...]
```

tests/e2e/watch_test.go
amitkrout: true

tests/e2e/utils.go
kadel: this is realted to `filepath.Join`?

kadel: Have you checked https://onsi.github.io/ginkgo/#marking-specs-as-failed ?

amitkrout:
Closing this as invalid for now. Will be fixed along with new odo work flow watch flag implementation.

PR #1446, opened 2019-03-11
Implement new odo workflow for: odo watch

cdrage:
Implements `odo watch` by using the local configuration file.

To use these changes, simply use: `odo watch` in the same directory that
you've deployed, or `odo watch --context ~/nodejs-ex`.

This builds upon: https://github.com/redhat-developer/odo/pull/1360

Closes https://github.com/redhat-developer/odo/issues/1419

### DIFF
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -1,8 +1,6 @@
 package component
 
 import (
-	"github.com/openshift/odo/pkg/config"
-	"github.com/openshift/odo/pkg/occlient"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -36,13 +34,4 @@ type ComponentStatus struct {
 	Active           bool                `json:"active"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
-}
-
-// ComponentOptions are generic options that are used between commands such as watch and push
-type ComponentOptions struct {
-	SourceType       config.SrcType
-	SourcePath       string
-	LocalConfig      *config.LocalConfigInfo
-	ComponentContext string
-	Client           *occlient.Client
 }
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -1,6 +1,8 @@
 package component
 
 import (
+	"github.com/openshift/odo/pkg/config"
+	"github.com/openshift/odo/pkg/occlient"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -34,4 +36,13 @@ type ComponentStatus struct {
 	Active           bool                `json:"active"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
+}
+
+// ComponentOptions are generic options that are used between commands such as watch and push
+type ComponentOptions struct {
+	SourceType       config.SrcType
+	SourcePath       string
+	LocalConfig      *config.LocalConfigInfo
+	ComponentContext string
+	Client           *occlient.Client
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -526,10 +526,10 @@ func GetSrcType(ctStr string) (SrcType, error) {
 	}
 }
 
-// CorrectSourcePath corrects the current sourcePath depending on local or binary configuration,
+// GetOSSourcePath corrects the current sourcePath depending on local or binary configuration,
 // if Git has been passed, we simply return the source location from LocalConfig
 // this will get the correct source path whether on Windows, macOS or Linux.
-func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
+func GetOSSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
 
 	cmpName := localConfig.GetName()
 	sourceType := localConfig.GetSourceType()
@@ -564,7 +564,7 @@ func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
 			return "", fmt.Errorf("Component %s has invalid source path %s", cmpName, u.Scheme)
 		}
 
-		return util.ReadFilePath(u, runtime.GOOS), nil
+		return sourcePath, nil
 
 	}
 	return "", fmt.Errorf("SourceType for correcting a SourcePath must be binary or local")
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -526,7 +526,8 @@ func GetSrcType(ctStr string) (SrcType, error) {
 	}
 }
 
-// correctSourcePath corrects the current sourcePath depending on local or binary configuration
+// CorrectSourcePath corrects the current sourcePath depending on local or binary configuration
+// this will get the correct source path whether on Windows, macOS or Linux.
 func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
 
 	cmpName := localConfig.GetName()
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -531,6 +531,10 @@ func GetSrcType(ctStr string) (SrcType, error) {
 // this will get the correct source path whether on Windows, macOS or Linux.
 func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 
+	// TODO:
+	// Always piped to "fromslash" so it's correct for the OS..
+	//
+
 	cmpName := lci.GetName()
 	sourceType := lci.GetSourceType()
 	sourceLocation := lci.GetSourceLocation()
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -526,7 +526,8 @@ func GetSrcType(ctStr string) (SrcType, error) {
 	}
 }
 
-// CorrectSourcePath corrects the current sourcePath depending on local or binary configuration
+// CorrectSourcePath corrects the current sourcePath depending on local or binary configuration,
+// if Git has been passed, we simply return the source location from LocalConfig
 // this will get the correct source path whether on Windows, macOS or Linux.
 func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
 
@@ -534,31 +535,37 @@ func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
 	sourceType := localConfig.GetSourceType()
 	sourceLocation := localConfig.GetSourceLocation()
 
-	// We use "ToSlash" here to make sure that deliminators are in a slash format
 	if sourceLocation == "" {
 		return "", fmt.Errorf("Blank source location provided")
 	}
 
-	sourcePath := filepath.ToSlash(localConfig.GetSourceLocation())
+	// If we are passing in a GIT source type, we make no changes.
+	// if it's binary or local, we will correct the source path.
+	if sourceType == GIT {
+		glog.V(4).Info("Git source type detected, not correcting SourcePath location")
+		return sourceLocation, nil
+	} else if sourceType == BINARY || sourceType == LOCAL {
 
-	if sourceType == BINARY || sourceType == LOCAL {
+		// We use "ToSlash" here to make sure that deliminators are in a slash format
+		sourcePath := filepath.ToSlash(localConfig.GetSourceLocation())
 		u, err := url.Parse(sourcePath)
 		if err != nil {
 			return "", errors.Wrapf(err, "unable to parse source %s from component %s", sourcePath, cmpName)
 		}
 
-		// See https://github.com/golang/go/issues/13276
-		// url/parse is unable to parse Windows file paths.. so here we detect if there is one character in the Scheme
-		// Ex. C or Z drive, etc.. as well as check to see if we're on Windows.
-		// else
-		// we check the schema normally (on Linux / macOS)
 		if len(u.Scheme) == 1 && runtime.GOOS == "windows" {
+			// See https://github.com/golang/go/issues/13276
+			// url/parse is unable to parse Windows file paths.. so here we detect if there is one character in the Scheme
+			// Ex. C or Z drive, etc.. as well as check to see if we're on Windows.
+			// else
+			// we check the schema normally (on Linux / macOS)
 			glog.V(4).Info("We're using Windows here and unfortunately net/url doesn't support parsing.. we're going to skip this")
 		} else if u.Scheme != "" && u.Scheme != "file" {
 			return "", fmt.Errorf("Component %s has invalid source path %s", cmpName, u.Scheme)
 		}
 
 		return util.ReadFilePath(u, runtime.GOOS), nil
+
 	}
 	return "", fmt.Errorf("SourceType for correcting a SourcePath must be binary or local")
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,11 +2,14 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
 	"github.com/openshift/odo/pkg/util"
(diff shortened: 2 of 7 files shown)


cdrage:
A lot of TODOs:

- [ ] Have to watch for https://github.com/redhat-developer/odo/pull/1360 to be merged
- [ ] Tests to be added
- [ ] CLI validation (similar to @anmolbabu 's work on `odo push` validation) also needs to be added

P.S: I'm unable to open a draft through CLI for git :( so this is WIP / draft!

cdrage:
This is ready for review / testing / trying it out!

To test:

```sh
odo create nodejs --context ~/nodejs-ex
odo push --context ~/node-js
odo watch --context ~/node-js
```

Same with git (use `--git https://github.com/openshift/nodejs-ex`) as well as `--binary`.

Examples below:

```sh
# Binary
▶ odo watch
Waiting for something to change in /home/wikus/example.war
File /home/wikus/example.war changed
Pushing files...
 ✓  Waiting for component to start
 ✓  Copying files to component
 ✓  Building component
Waiting for something to change in /home/wikus/example.war

# Local
▶ odo watch --context ~/nodejs-ex 
Waiting for something to change in /home/wikus/nodejs-ex
File /home/wikus/nodejs-ex/views/index.html changed
File /home/wikus/nodejs-ex/views/index.html~ changed
Pushing files...
Waiting for something to change in /home/wikus/nodejs-ex

```

cdrage:
ping @anmolbabu @amitkrout @surajnarwade @girishramnani

cdrage:
@amitkrout I already had an e2e_test.go created, but since we're refactoring e2e tests I won't be able to add it until #1535  is merged

pkg/odo/cli/utils/utils.go
```diff
@@ -36,3 +42,63 @@ func VisitCommands(cmd *cobra.Command, f func(*cobra.Command)) {
...
+}
+
+// correctSourcePath corrects the current sourcePath depending on local or binary configuration
+func correctSourcePath(localConfig *config.LocalConfigInfo) (path string, err error) {
+
+	cmpName := localConfig.GetName()
+	sourceType := localConfig.GetSourceType()
+	sourcePath := localConfig.GetSourceLocation()
+
+	if sourceType == config.BINARY || sourceType == config.LOCAL {
```
anmolbabu: Can the lines 79 to 88 be moved instead to pkg/config/config.go#GetSourceLocation instead ?

pkg/odo/cli/utils/utils.go
```diff
@@ -36,3 +42,63 @@ func VisitCommands(cmd *cobra.Command, f func(*cobra.Command)) {
...
+// LocalConfigInfo is a struct that will contain the LocalConfig as well as SourcePath
+// the reasoning behind SourcePath is due to the fact that we must correct what the path is
+// on different platforms (Linux, Windows, etc.) as well as get the absolute path of the component.
+type LocalConfigInfo struct {
+	LocalConfig *config.LocalConfigInfo
+	SourcePath  string
+}
+
+// RetrieveLocalConfigInfo retrieves the local configuration
+func RetrieveLocalConfigInfo(componentContext string) (configInfo LocalConfigInfo, err error) {
```
anmolbabu: Can we move this into pkg/odo/genericclioptions/context.go ?
Essentially make localconfig a member field of Context and resolve it in pkg/odo/genericclioptions/context.go#newContext probably after removing project, app, component names from Context struct and resolving based on context flag in cobra.Command and hence the resolution will be there for every command ?

pkg/component/types.go
```diff
@@ -35,3 +37,12 @@ type ComponentStatus struct {
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
 }
+
+// ComponentOptions are generic options that are used between commands such as watch and push
+type ComponentOptions struct {
```
anmolbabu: :+1:  from me for this

pkg/odo/cli/utils/utils.go
cdrage: I don't think so. The only reasoning is that if we add it to GetSourceLocation:

```sh
// GetSourceLocation returns the sourcelocation, returns default if nil
func (lc *LocalConfig) GetSourceLocation() string {
  if lc.componentSettings.SourceLocation == nil {
    return ""
  }
  return *lc.componentSettings.SourceLocation
}
```

It would impact the general functionality of it. It would no longer "get" the value, but the function would interpolate something different...

cdrage:
PR has been updated. Please review again @amitkrout @syamgk @girishramnani @surajnarwade

surajnarwade:
@cdrage , just a small enhancement which can be taken care in other PR also, I created issue for the same here https://github.com/openshift/odo/issues/1546

kadel:
> tested, works for me 🎉

@surajnarwade please use `/lgtm`

amitkrout:
> @amitkrout I already had an e2e_test.go created, but since we're refactoring e2e tests I won't be able to add it until #1535 is merged

Pr #1535 is merged, So please start adding watch test in new structure format.

pkg/odo/genericclioptions/context.go
```diff
@@ -257,3 +260,63 @@ func ignoreButLog(err error) {
 		glog.V(4).Infof("Ignoring error as it usually means flag wasn't set: %v", err)
 	}
 }
+
+// LocalConfigInfo is a struct that will contain the LocalConfig as well as SourcePath
+// the reasoning behind SourcePath is due to the fact that we must correct what the path is
+// on different platforms (Linux, Windows, etc.) as well as get the absolute path of the component.
+type LocalConfigInfo struct {
```
kadel: This looks a little bit weird to me. Here you are defining another `LocalConfigInfo` with `SourcePath` that overlaps with what is already in `config.LocalConfigInfo` as `SourceLocation`, only in a slightly different format.

Shouldn't we store the SourceLocation  in url format, to make all the conversion easier?

pkg/odo/genericclioptions/context.go
```diff
@@ -257,3 +260,63 @@ func ignoreButLog(err error) {
 		glog.V(4).Infof("Ignoring error as it usually means flag wasn't set: %v", err)
 	}
 }
+
+// LocalConfigInfo is a struct that will contain the LocalConfig as well as SourcePath
```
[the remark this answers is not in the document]
cdrage: Are you sure we should be inserting it into here:

```sh
// GetSourceLocation returns the sourcelocation, returns default if nil                                                               
func (lc *LocalConfig) GetSourceLocation() string {                                                                                   
  if lc.componentSettings.SourceLocation == nil {                                                                                     
    return ""                                                                                                                         
  }
  return *lc.componentSettings.SourceLocation
}
```

This would change the functionality with regards to how the other `Get()` functions work..if you're okay with that I'll go forward with pushing it to: `GetSourceLocation()`

pkg/odo/genericclioptions/context.go
cdrage: I agree.. We could do url format, for example `git:https..blahblah`, `local:` `binary:` etc?

kadel: I think that a standard url fromat might be better, so for local an binary it would be `file:///path/` or `file:///path/to/file.jar`, and for git `https://github.com/ff/f/` or `git@github.com:user/ff`.

But not sure how `file://` works with a relative path. We will need to do research.

I would like to use something that is a standard rather than creating our own way.

kadel: it looks like we can't use a schema with the relative file path.


I think that the best would be to just store path for the local and binary component as a relative path with forwarding slashes `/`, Go has functions that deals with OS-specific paths (`/` vs `\`). 
We could use [filepath.FromSlash](https://golang.org/pkg/path/filepath/#FromSlash) and [filepath.ToSlash](https://golang.org/pkg/path/filepath/#ToSlash). ToSlash to convert path to `/` separated, and save it like this. Than before using it call `FromSlash` to convert it to `\`  or `/`  based on the OS.

cdrage:
@kadel @girishramnani @anmolbabu @surajnarwade @amitkrout 

Tests as well as the fix for Windows has been implemented here.. The fix for Windows in particular is here:

https://github.com/openshift/odo/pull/1446/files#diff-091ace02b49a46f55f2586cfec89a160R554

Please try this out on Windows! 

The test scenarios are there for Linux, unfortunately no Windows tests yet... :(

cdrage:
Will close: https://github.com/openshift/odo/issues/1543 as I've tested this on Windows and it works well.

surajnarwade:
:heavy_check_mark:  works for me on linux :penguin:  
/lgtm

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
...
+		}
+
+		if len(u.Scheme) == 1 && runtime.GOOS == "windows" {
+			// See https://github.com/golang/go/issues/13276
+			// url/parse is unable to parse Windows file paths.. so here we detect if there is one character in the Scheme
+			// Ex. C or Z drive, etc.. as well as check to see if we're on Windows.
+			// else
+			// we check the schema normally (on Linux / macOS)
+			glog.V(4).Info("We're using Windows here and unfortunately net/url doesn't support parsing.. we're going to skip this")
+		} else if u.Scheme != "" && u.Scheme != "file" {
```
kadel: not sure about this, do we even need this check?

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
 		return NONE, fmt.Errorf("Unsupported component source type: %s", ctStr)
 	}
 }
+
+// CorrectSourcePath corrects the current sourcePath depending on local or binary configuration,
+// if Git has been passed, we simply return the source location from LocalConfig
+// this will get the correct source path whether on Windows, macOS or Linux.
+func CorrectSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
```
kadel: Can we call this function something else? It sounds like we are correcting something that is broken :-(
Also to align it with the others wouldn't it be better to have LocalConfigInfo as a receiver?

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
...
+		// We use "ToSlash" here to make sure that deliminators are in a slash format
+		sourcePath := filepath.ToSlash(localConfig.GetSourceLocation())
+		u, err := url.Parse(sourcePath)
+		if err != nil {
+			return "", errors.Wrapf(err, "unable to parse source %s from component %s", sourcePath, cmpName)
+		}
+
+		if len(u.Scheme) == 1 && runtime.GOOS == "windows" {
+			// See https://github.com/golang/go/issues/13276
+			// url/parse is unable to parse Windows file paths.. so here we detect if there is one character in the Scheme
```
kadel: why don't you try to parse it before it is converted to OS specific path?

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
...
+
+	// If we are passing in a GIT source type, we make no changes.
+	// if it's binary or local, we will correct the source path.
+	if sourceType == GIT {
+		glog.V(4).Info("Git source type detected, not correcting SourcePath location")
+		return sourceLocation, nil
+	} else if sourceType == BINARY || sourceType == LOCAL {
+
+		// We use "ToSlash" here to make sure that deliminators are in a slash format
+		sourcePath := filepath.ToSlash(localConfig.GetSourceLocation())
```
kadel: Shouldn't it be `FromSlash`?

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
...
+			// url/parse is unable to parse Windows file paths.. so here we detect if there is one character in the Scheme
+			// Ex. C or Z drive, etc.. as well as check to see if we're on Windows.
+			// else
+			// we check the schema normally (on Linux / macOS)
+			glog.V(4).Info("We're using Windows here and unfortunately net/url doesn't support parsing.. we're going to skip this")
+		} else if u.Scheme != "" && u.Scheme != "file" {
+			return "", fmt.Errorf("Component %s has invalid source path %s", cmpName, u.Scheme)
+		}
+
+		return util.ReadFilePath(u, runtime.GOOS), nil
```
kadel: We shouldn't needs this function anymore `FromSlash` and `ToSlash` should handle everything

pkg/config/config.go
cdrage: That makes more sense :+1:

pkg/config/config.go
cdrage: I tried it with FromSlash and it does not work with Windows. I have a feeling it's because Go / Windows works by default with forward slashes. Using "ToSlash" will work regardless since if the user passes in C:\foo\bar it'll convert to C:/foo/bar anyways.

pkg/config/config.go
cdrage: Yes, it's nice to have this check in to make sure that it's been parsed correctly and it's a file not an https location, this would be a form of validation check for `local` and `binary` to make sure that a URL hasn't been passed in.

pkg/config/config.go
cdrage: you're right :+1: I've changed this.

cdrage:
@kadel addressed your comments as well as why we keep some checks (we need the schema + file check for validation).

pkg/odo/cli/component/push.go
```diff
@@ -42,59 +40,53 @@ const PushRecommendedCommandName = "push"
...
 type PushOptions struct {
-	ignores          []string
+	ignores []string
+	show    bool
+
 	sourceType       config.SrcType
 	sourcePath       string
-	localConfig      *config.LocalConfigInfo
 	componentContext string
+	client           *occlient.Client
```
anmolbabu: I htink this can be safely removed
bcoz context would have a client

pkg/config/config.go
kadel: I thought that this function is used for reading.
Full windows path like `C:\foo\bar` should never get into the config file

pkg/odo/genericclioptions/context.go
```diff
@@ -12,11 +12,20 @@ import (
...
+	"github.com/openshift/odo/pkg/util"
 )
 
+// ComponentOptions are generic options that are used between commands such as watch and push
+type ComponentOptions struct {
+	SourceType       config.SrcType
+	SourcePath       string
+	LocalConfig      *config.LocalConfigInfo
+	ComponentContext string
+	Client           *occlient.Client
```
anmolbabu: why client here again?
Almost every command generates context which already has client

pkg/odo/genericclioptions/context.go
```diff
@@ -12,11 +12,20 @@ import (
...
 	"github.com/openshift/odo/pkg/occlient"
-	"github.com/openshift/odo/pkg/odo/util"
+	odoutil "github.com/openshift/odo/pkg/odo/util"
 	"github.com/openshift/odo/pkg/project"
-	pkgUtil "github.com/openshift/odo/pkg/util"
+	"github.com/openshift/odo/pkg/util"
 )
 
+// ComponentOptions are generic options that are used between commands such as watch and push
+type ComponentOptions struct {
```
anmolbabu: I don't see it used anywhere in this PR in the git diff view shown here
And why have Sourcetype, SourcePath etc here ?
Is there any more processing to these than what can be fetched from LocalConfig?

pkg/odo/cli/component/push.go
cdrage: We need this for `isPrjExists, err := project.Exists(po.client, prjName)` so it has to remain here..

pkg/config/config.go
```diff
@@ -522,3 +525,47 @@ func GetSrcType(ctStr string) (SrcType, error) {
 		return NONE, fmt.Errorf("Unsupported component source type: %s", ctStr)
 	}
 }
+
+// GetOSSourcePath corrects the current sourcePath depending on local or binary configuration,
+// if Git has been passed, we simply return the source location from LocalConfig
+// this will get the correct source path whether on Windows, macOS or Linux.
+func GetOSSourcePath(localConfig *LocalConfigInfo) (path string, err error) {
```
kadel: Can we align it with the rest of Get functions and use LocalConfig as a receiver?
```go
func (lc *LocalConfig) GetOSSourcePath()
```

cdrage: Done.

amitkrout:
> > @amitkrout I already had an e2e_test.go created, but since we're refactoring e2e tests I won't be able to add it until #1535 is merged
> 
> Pr #1535 is merged, So please start adding watch test in new structure format.

@cdrage Can you pleas add the e2e test scenario in old format if you are not familiar enough with new test structure.

amitkrout:
@cdrage Please have a look at CI failures

cdrage:
Okay, will do @amitkrout  :+1:

cdrage:
@amitkrout @kadel 

So I've updated this PR again, if you could have another look at it...

@kadel I was thinking of instead using a separate PR to fix the default behaviour that we talked about with regards to `local` and `binary` components. Essentially, saving the path as `./` instead of the absolute path in the configuration..

As of right now, this PR works wonderfully with Windows and fixes the issues with `odo watch`.

A few things however:
 - e2e tests are proving to be very difficult as I'm uncommenting all the `odo watch` e2e tests and encountering wayyyyyy too many errors, could we open up an issue and I write some e2e tests after #1559  is merged in?
 - I'll open up another issue with regards to changing the default behaviour with `./` absolute directory being saved instead of the actual directory.

WDYT @kadel ?

Please check out the new updated function too! ^^

pkg/config/config.go
```diff
@@ -522,3 +524,34 @@ func GetSrcType(ctStr string) (SrcType, error) {
...
+	sourceLocation := lci.GetSourceLocation()
+
+	if sourceLocation == "" {
+		return "", fmt.Errorf("Blank source location provided")
+	}
+
+	if sourceType == GIT {
+		glog.V(4).Info("Git source type detected, not correcting SourcePath location")
+		return sourceLocation, nil
+	}
```
kadel: 👍  I like that you didn't put `else` here ❤️

amitkrout:
> e2e tests are proving to be very difficult as I'm uncommenting all the odo watch e2e tests and encountering wayyyyyy too many errors, could we open up an issue and I write some e2e tests after #1559 is merged in?

@cdrage No issue, please create an follow up issue for watch test only. Anyway the minimal watch test is handled by pr #1559 for beta release.

PR #1449, opened 2019-03-12
added env variables in component config

girishramnani:
Added env variables in the component settings of local odo config.

Note - using the env variables in workflow is not in the scope of this PR, its in the next one once this gets merged

partly  https://github.com/redhat-developer/odo/issues/1390

## How to test changes?

```
odo config set --env PORT=4000
odo config set  --env DB_USERNAME=postgres,DB_HOSTNAME=postgres
odo config set  --env KAFKA_HOST=kafka --env KAFKA_PORT=6639

# Unset a env variable in the local config
odo config unset --env PORT
odo config unset --env DB_USERNAME,DB_HOSTNAME
odo config unset --env KAFKA_HOST --env KAFKA_PORT

# should show the env variables
odo config view 

```

### DIFF
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -47,14 +47,6 @@ func NewEnvVarListFromSlice(envList []string) (EnvVarList, error) {
 
 }
 
-// MergeEnvVarList merges the two provided envVarList
-func MergeEnvVarList(evl EnvVarList, otherEvl EnvVarList) EnvVarList {
-	for _, envVar := range otherEvl {
-		evl = append(evl, envVar)
-	}
-	return evl
-}
-
 // RemoveEnvVarsFromList removes the env variables based on the keys provided
 // and returns a new EnvVarList
 func RemoveEnvVarsFromList(envVarList EnvVarList, keys []string) EnvVarList {
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -1,0 +1,71 @@
+package config
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/redhat-developer/odo/pkg/util"
+)
+
+// EnvVar represents an enviroment variable
+type EnvVar struct {
+	Name  string `yaml:"Name"`
+	Value string `yaml:"Value"`
+}
+
+// EnvVarList represents a list of environment variables
+type EnvVarList []*EnvVar
+
+// NewEnvVarFromString takes a string of format "name=value" and returns an Env
+// variable struct
+func NewEnvVarFromString(envStr string) (*EnvVar, error) {
+	envList := strings.Split(envStr, "=")
+	// if there is not = in the string
+	if len(envList) < 2 {
+		return nil, errors.New("invalid environment variable format")
+	}
+
+	return &EnvVar{
+		Name:  strings.TrimSpace(envList[0]),
+		Value: strings.TrimSpace(envList[1]),
+	}, nil
+}
+
+// NewEnvVarListFromSlice takes multiple env variables with format
+// "name=value" and returns an EnvVarList
+func NewEnvVarListFromSlice(envList []string) (EnvVarList, error) {
+	var envVarList EnvVarList
+	for _, envStr := range envList {
+		envVar, err := NewEnvVarFromString(envStr)
+		if err != nil {
+			return nil, err
+		}
+		envVarList = append(envVarList, envVar)
+	}
+
+	return envVarList, nil
+
+}
+
+// MergeEnvVarList merges the two provided envVarList
+func MergeEnvVarList(evl EnvVarList, otherEvl EnvVarList) EnvVarList {
+	for _, envVar := range otherEvl {
+		evl = append(evl, envVar)
+	}
+	return evl
+}
+
+// RemoveEnvVarsFromList removes the env variables based on the keys provided
+// and returns a new EnvVarList
+func RemoveEnvVarsFromList(envVarList EnvVarList, keys []string) EnvVarList {
+	newEnvVarList := EnvVarList{}
+	for _, envVar := range envVarList {
+		// if the env is in the keys we skip it
+		if util.In(keys, envVar.Name) {
+			continue
+		}
+
+		newEnvVarList = append(newEnvVarList, envVar)
+	}
+	return newEnvVarList
+}
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -23,7 +23,7 @@ func (evl EnvVarList) Merge(other EnvVarList) EnvVarList {
 	var dedupNewEvl EnvVarList
 	newEvl := append(evl, other...)
 	uniqueMap := make(map[string]string)
-	// last value will be kept in case o
+	// last value will be kept in case of duplicate env vars
 	for _, envVar := range newEvl {
 		uniqueMap[envVar.Name] = envVar.Value
 	}
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -16,6 +16,29 @@ type EnvVar struct {
 // EnvVarList represents a list of environment variables
 type EnvVarList []EnvVar
 
+// Merge merges the other EnvVarlist with keeping last value for duplicate EnvVars
+// and returns a new EnvVarList
+func (evl EnvVarList) Merge(other EnvVarList) EnvVarList {
+
+	var dedupNewEvl EnvVarList
+	newEvl := append(evl, other...)
+	uniqueMap := make(map[string]string)
+	// last value will be kept in case o
+	for _, envVar := range newEvl {
+		uniqueMap[envVar.Name] = envVar.Value
+	}
+
+	for key, value := range uniqueMap {
+		dedupNewEvl = append(dedupNewEvl, EnvVar{
+			Name:  key,
+			Value: value,
+		})
+	}
+
+	return dedupNewEvl
+
+}
+
 // NewEnvVarFromString takes a string of format "name=value" and returns an Env
 // variable struct
 func NewEnvVarFromString(envStr string) (EnvVar, error) {
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"strings"
 
-	"github.com/redhat-developer/odo/pkg/util"
+	"github.com/openshift/odo/pkg/util"
 )
 
 // EnvVar represents an enviroment variable
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -14,18 +14,18 @@ type EnvVar struct {
 }
 
 // EnvVarList represents a list of environment variables
-type EnvVarList []*EnvVar
+type EnvVarList []EnvVar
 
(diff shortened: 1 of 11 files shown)


pkg/config/env_var.go
```diff
@@ -0,0 +1,71 @@
...
+		envVarList = append(envVarList, envVar)
+	}
+
+	return envVarList, nil
+
+}
+
+// MergeEnvVarList merges the two provided envVarList
+func MergeEnvVarList(evl EnvVarList, otherEvl EnvVarList) EnvVarList {
+	for _, envVar := range otherEvl {
```
golangcibot: S1011: should replace loop with `evl = append(evl, otherEvl...)` (from `gosimple`)

girishramnani:
@amitkrout added e2e tests

tests/e2e/e2e_test.go
```diff
@@ -276,6 +276,32 @@ var _ = Describe("odoe2e", func() {
 			timeoutValue := getPreferenceValue("Timeout")
 			Expect(timeoutValue).To(BeEmpty())
 		})
+
+		It("should set env variables", func() {
+			runCmdShouldPass("odo config set --env PORT=4000")
+			configValue := runCmdShouldPass("odo config view")
+			Expect(configValue).To(ContainSubstring("PORT"))
+			runCmdShouldPass("odo config set --env SECRET_KEY=R2lyaXNoIFJhbW5hbmkgaXMgdGhlIGJlc3Q=")
```
amitkrout: $ echo R2lyaXNoIFJhbW5hbmkgaXMgdGhlIGJlc3Q= | base64 --decode
Girish Ramnani is the best

kadel:
> /approve

self-approving? :-D

girishramnani:
@tkral wasn't this happening by itself before? self approving the PR

girishramnani:
> Bit confused why int are in quotes but not string variables?
> 
> ```shell
> github.com/redhat-developer/odo  pr_1449 ✔                                                                                                                                                                                                                                   2d
> ▶ cat .odo/config.yaml 
> kind: LocalConfig
> apiversion: odo.openshift.io/v1alpha1
> ComponentSettings:
>   Envs:
>   - Name: KAFKA_HOST
>     Value: kafka
>   - Name: KAFKA_PORT
>     Value: "6639"
>   - Name: DB_USERNAME
>     Value: postgres
>   - Name: DB_HOSTNAME
>     Value: postgres
>   - Name: PORT
>     Value: "4000"
> ```

this is the yaml compiler implementation detail we cannot change how string that look like int are encoded in yaml

pkg/odo/cli/config/set.go
```diff
@@ -30,6 +32,11 @@ var (
    %[1]s %[8]s 0.5 
    %[1]s %[9]s 2 
    %[1]s %[10]s 1 
+
+   # Set a env variable in the local config
+   %[1]s %[2]s --env PORT=4000
+   %[1]s %[3]s --env DB_USERNAME=postgres,DB_HOSTNAME=postgres
+   %[1]s %[4]s --env KAFKA_HOST=kafka --env KAFKA_PORT=6639
```
cdrage: This is incorrect...

See below output from help:

```sh
  # Set a env variable in the local config
  odo config set Type --env PORT=4000
  odo config set Name --env DB_USERNAME=postgres,DB_HOSTNAME=postgres
  odo config set MinMemory --env KAFKA_HOST=kafka --env KAFKA_PORT=6639

  # Unset a configuration value in the local config
  odo config unset Type
  odo config unset Name
  odo config unset MinMemory
  odo config unset MaxMemory
  odo config unset Memory
  odo config unset Ignore
  odo config unset MinCPU
  odo config unset MaxCPU
  odo config unset CPU

  # Unset a env variable in the local config
  odo config unset Type --env PORT
  odo config unset Name --env DB_USERNAME,DB_HOSTNAME
  odo config unset MinMemory --env KAFKA_HOST --env KAFKA_PORT
```

tests/e2e/e2e_test.go
```diff
@@ -276,6 +276,32 @@ var _ = Describe("odoe2e", func() {
...
+			configValue = runCmdShouldPass("odo config view")
```
amitkrout: Do more specific  search. For example use ```Value := getConfigValue(KEY)``` instead of ```configValue = runCmdShouldPass("odo config view")```. Also add test steps to verify https://github.com/openshift/odo/pull/1449#pullrequestreview-219098881

pkg/odo/cli/config/set.go
kadel: Do we really need to show all three options? It makes the help look convoluted :-(

pkg/odo/cli/config/unset.go
```diff
@@ -32,13 +33,19 @@ var (
    %[1]s %[8]s 
    %[1]s %[9]s 
    %[1]s %[10]s
+
+   # Unset a env variable in the local config
+   %[1]s %[2]s --env PORT
+   %[1]s %[3]s --env DB_USERNAME,DB_HOSTNAME
+   %[1]s %[4]s --env KAFKA_HOST --env KAFKA_PORT
```
kadel: Do we really need to show all three options? It makes the help look convoluted :-(

kadel:
duplicate values shoud not be allowed 

```
github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
▶ odo config set --env PORT=4000

Environment variables were successfully updated.

github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
▶ odo config set --env PORT=4000

Environment variables were successfully updated.

github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Envs:
  - Name: PORT
    Value: "4000"
  - Name: PORT
    Value: "4000"
```

pkg/odo/cli/config/set.go
girishramnani: so what do you suggest @kadel

tests/e2e/e2e_test.go
girishramnani: done

pkg/odo/cli/config/unset.go
girishramnani: so only 1 is fine then

girishramnani:
> duplicate values shoud not be allowed
> 
> ```
> github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
> ▶ odo config set --env PORT=4000
> 
> Environment variables were successfully updated.
> 
> github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
> ▶ odo config set --env PORT=4000
> 
> Environment variables were successfully updated.
> 
> github.com/openshift/odo  feature/component_env ✗                                                                                          16h35m ◒
> ▶ cat .odo/config.yaml
> kind: LocalConfig
> apiversion: odo.openshift.io/v1alpha1
> ComponentSettings:
>   Envs:
>   - Name: PORT
>     Value: "4000"
>   - Name: PORT
>     Value: "4000"
> ```

resolved

girishramnani:
blocked on  https://github.com/openshift/odo/pull/1517

pkg/odo/cli/config/unset.go
girishramnani: kept only one example

pkg/odo/cli/config/set.go
girishramnani: kept only one example

girishramnani: resolve the incorrect output too @cdrage

pkg/config/env_var.go
```diff
@@ -0,0 +1,86 @@
...
+type EnvVarList []EnvVar
+
+// Merge merges the other EnvVarlist with keeping last value for duplicate EnvVars
+// and returns a new EnvVarList
+func (evl EnvVarList) Merge(other EnvVarList) EnvVarList {
+
+	var dedupNewEvl EnvVarList
+	newEvl := append(evl, other...)
+	uniqueMap := make(map[string]string)
+	// last value will be kept in case o
```
cdrage: missing words on the end of comment I think?

girishramnani: resolved

anmolbabu:
So, aren't planning to use/set/apply the env vars in config onto the component as part of this PR ?

girishramnani:
That is after this one, smaller and faster PRs. And as we r changing the
test structure and the config set of env requires a good e2e coverage in
comparison to this part, I would prefer that consuming the env is done in
the next PR

mohammedzee1000:
> @tkral wasn't this happening by itself before? approving the PR

Its much safer to have that disabled. Luckily you cant lgtm your own pr

pkg/util/util.go
```diff
@@ -78,6 +78,16 @@ func GenerateRandomString(n int) string {
 	return string(b)
 }
 
+// In checks if the value is in the array
+func In(arr []string, value string) bool {
```
anmolbabu: Just a thought, since this is in common utils file, should we call this IsStrInSlice instead?

girishramnani:
I assumed that we still had that workflow and so followed it. My bad :sweat_smile:

pkg/util/util.go
girishramnani: The argument does makes sure that its a string slice, do we wanna make it more verbose?

girishramnani:
> Also should we avoid setting/overriding any env that's set by odo like in https://github.com/openshift/odo/blob/master/pkg/occlient/occlient.go#L107-L154
> I know they are not exposed to the user by config view so if user is setting/overriding it, it might be a intended action but just a thought should we stop user from setting any env with name beginning from ODO and reserve env names starting from ODO to use by ODO internally only? I am confused about priority of such a check though...

Sure I can record this for the next PR. But also would like @kadel view on this

pkg/util/util.go
anmolbabu: Very vague ;)

What if I want to add a function that checks the similar for a different data type you have reserved `In` for strings here ;) But feel free to ignore the comment

kadel:
> > Also should we avoid setting/overriding any env that's set by odo like in https://github.com/openshift/odo/blob/master/pkg/occlient/occlient.go#L107-L154
> > I know they are not exposed to the user by config view so if user is setting/overriding it, it might be a intended action but just a thought should we stop user from setting any env with name beginning from ODO and reserve env names starting from ODO to use by ODO internally only? I am confused about priority of such a check though...
> 
> Sure I can record this for the next PR. But also would like @kadel view on this

It is really good point. We should have a check like this, but it can be addressed later. Please create separate issue for it

girishramnani:
opened an issue to track this https://github.com/openshift/odo/issues/1530

pkg/util/util.go
girishramnani: I can change this in the next PR? this will delay the merge by 50min. I will document this in a follow up issue

girishramnani: tracked https://github.com/openshift/odo/issues/1530

amitkrout:
prow glitch - ```level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"```

/lgtm

PR #1451, opened 2019-03-13
Fixing username as per e2e_test file

amitkrout:

To run e2e_test on OpenShift 4.0. Just update the user name which is being used in e2e_test.go as a login user.

No

## How to test changes?

As of now make sure the user name "odonoprojectattemptscreate" should be present in e2e_test.go test file.


amitkrout:
@mohammedzee1000 Can you please do a review as its blocking debugging of some failure in e2e_test scenario on OpenShift 4.0. 

Won't take much time. I just updated a user name

PR #1360, opened 2019-02-23
Apply ComponentSettings as part of component push

anmolbabu:
~~fixes #1198~~
Signed-off-by: anmolbabu <anmolbudugutta@gmail.com>

## Was the change discussed in an issue?
fixes #1366

## How to test changes?

Everything related to this PR that will be addressed separately after this gets merged is recorded as a issue with "new workflow" label.
https://github.com/redhat-developer/odo/labels/new%20workflow

PR #1453, opened 2019-03-13
local config improvements

girishramnani:
Use of a proxyLocalConfig to hide the ComponentSettings and provide getters for all local config parameters.

No issue

## How to test changes?
check if all `odo config` commands still work


pkg/config/config.go
```diff
@@ -110,40 +152,66 @@ func NewLocalConfig() LocalConfig {
...
+		case "app":
+			lci.componentSettings.App = &strValue
+		case "project":
+			lci.componentSettings.Project = &strValue
+		case "sourcetype":
+			cmpSourceType, err := util.GetCreateType(strValue)
+			if err != nil {
+				return errors.Wrapf(err, "unable to set %s to %s", parameter, strValue)
+			}
+			value = string(cmpSourceType)
```
golangcibot: ineffectual assignment to `value` (from `ineffassign`)

pkg/config/config.go
```diff
@@ -32,6 +32,22 @@ type ComponentSettings struct {
...
+	Path *string `yaml:"Path,omitempty"`
+
+	// Ref is component source git ref but can be levaraged for more in future
+	Ref *string `yaml:"Ref,omitempty"`
+
+	// Type is type of component source: git/local/binary
+	SourceType *string `yaml:"SourceType,omitempty"`
+
+	// Ports is a slice of ports to be exposed when a component is created
+	Ports *string `yaml:"Ports,omitempty"`
```
anmolbabu: slice like we discussed on call with @kadel ?

pkg/config/config.go
```diff
@@ -32,6 +32,22 @@ type ComponentSettings struct {
...
 	ComponentType *string `yaml:"ComponentType,omitempty"`
 
+	// Path is path to binary in current/context dir
+	Path *string `yaml:"Path,omitempty"`
+
+	// Ref is component source git ref but can be levaraged for more in future
+	Ref *string `yaml:"Ref,omitempty"`
+
+	// Type is type of component source: git/local/binary
+	SourceType *string `yaml:"SourceType,omitempty"`
```
anmolbabu: also, @kadel was of the opinion that this could be of type https://github.com/redhat-developer/odo/blob/master/pkg/occlient/occlient.go#L67

pkg/config/config.go
```diff
@@ -32,6 +32,22 @@ type ComponentSettings struct {
...
+
+	// Ports is a slice of ports to be exposed when a component is created
+	Ports *string `yaml:"Ports,omitempty"`
+
+	App *string `yaml:"App,omitempty"`
```
kadel: full `Application!` would be better

pkg/config/config.go
anmolbabu: aah i think i see the magic below

pkg/config/config.go
```diff
@@ -32,6 +32,22 @@ type ComponentSettings struct {
...
+
+	Project *string `yaml:"Project,omitempty"`
+
 	ComponentName *string `yaml:"ComponentName,omitempty"`
```
kadel: Can we call this just `Name`?

pkg/config/config.go
```diff
@@ -32,6 +32,22 @@ type ComponentSettings struct {
 	// The builder image to use
 	ComponentType *string `yaml:"ComponentType,omitempty"`
```
kadel: just `Type` might be better

pkg/config/config.go
```diff
@@ -48,6 +64,13 @@ type ComponentSettings struct {
 
 // LocalConfig holds all the config relavent to a specific Component.
 type LocalConfig struct {
+	metav1.TypeMeta   `yaml:",inline"`
+	componentSettings ComponentSettings `yaml:"ComponentSettings,omitempty"`
+}
+
+// ProxyLocalConfig holds all the parameter that local config does but exposes all
+// of it, used for serialization.
+type ProxyLocalConfig struct {
```
kadel: can we make`ProxyLocalConfig` itself unexported?

pkg/config/config.go
girishramnani: done

pkg/config/config.go
```diff
@@ -110,40 +152,69 @@ func NewLocalConfig() LocalConfig {
 	}
 }
 
+// NewProxyLocalConfig creates an empty ProxyLocalConfig struct with typeMeta populated
+func NewProxyLocalConfig() ProxyLocalConfig {
```
kadel: can we unexport this, as it shouldn't be used outside of this package

pkg/config/config.go
```diff
@@ -155,18 +226,18 @@ func (lci *LocalConfigInfo) GetConfiguration(parameter string) (interface{}, boo
 
 	switch strings.ToLower(parameter) {
```
kadel: Can we remove `GetConfiguration` function? It is returning `interface{}` :-( We have getters for every field

pkg/config/config.go
girishramnani: @anmolbabu there is a `CreateType` in utils package as well

pkg/config/config.go
```diff
@@ -48,6 +64,13 @@ type ComponentSettings struct {
 
 // LocalConfig holds all the config relavent to a specific Component.
 type LocalConfig struct {
+	metav1.TypeMeta   `yaml:",inline"`
```
kadel: do we even need `TypeMeta` here? It is the internal thing that shouldn't be never touched outside this package

girishramnani: we shouldn't diverge how proxyconfig and this config looks, lets keep them same

pkg/config/config.go
girishramnani: need `GetConfiguration` for `odo config view`

pkg/config/config.go
kadel: why?

pkg/config/config.go
kadel: Now we are handling default values in two places :-( It will lead to inconsistencies. 

Getters for each field should be the only source of truth for the config values.

kadel: You are using `GetComponentSettings` not `GetConfiguration` in `view.go`

pkg/util/util.go
```diff
@@ -554,3 +554,31 @@ func MachineOutput(outputFlag string, resource interface{}) (string, error) {
 
 	return string(out), err
 }
+
+// CreateType is an enum to indicate the type of source of component -- local source/binary or git for the generation of app/component names
+type CreateType string
```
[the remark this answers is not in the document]
girishramnani: this is what @anmolbabu has done in his PR

pkg/config/config.go
girishramnani: done

pkg/config/config.go
girishramnani: done

pkg/config/config.go
girishramnani: done

pkg/config/config.go
girishramnani: done

pkg/config/config.go
girishramnani: done

pkg/util/util.go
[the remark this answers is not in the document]
girishramnani: currently I have `CreateType` in both `occlient` and `util` package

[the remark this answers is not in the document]
girishramnani: figured it out, resolved

pkg/config/config.go
girishramnani: done resolved, changed sourceType to create type

pkg/config/config.go
```diff
@@ -110,40 +154,68 @@ func NewLocalConfig() LocalConfig {
 	}
 }
 
+// newProxyLocalConfig creates an empty proxyLocalConfig struct with typeMeta populated
+func newProxyLocalConfig() proxyLocalConfig {
+	lc := NewLocalConfig()
```
anmolbabu: can you add a way to pass a config file path(that's not the current directory basically context use case) which will be used here?

girishramnani: there is alot of work related to this in your PR, maybe you can rebase and do that, If I do it then alot of the work would have to be moved here.

girishramnani: also seems out of scope

anmolbabu: ok

girishramnani: resolving

pkg/util/util.go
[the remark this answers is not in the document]
anmolbabu: no I moved it from occlient to util it could be in config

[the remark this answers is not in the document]
anmolbabu: so with this I think its in 2 places now :) and 2 different structs?

pkg/config/config.go
```diff
@@ -175,29 +243,136 @@ func (lci *LocalConfigInfo) DeleteConfiguration(parameter string) error {
...
+// GetRef returns the ref, returns default if nil
+func (lc *LocalConfig) GetRef() string {
+	if lc.componentSettings.Ref == nil {
+		return ""
+	}
+	return *lc.componentSettings.Ref
+}
+
+// GetSourceType returns the source type, returns default if nil
+func (lc *LocalConfig) GetSourceType() string {
```
anmolbabu: Can this return CreateType or SourceType or what ever it is please?

girishramnani: you are looking at outdated code @anmolbabu

pkg/util/util.go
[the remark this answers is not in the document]
anmolbabu: I'll move it to config in my PR bcoz its used currently in multiple places not just in config

pkg/config/config.go
```diff
@@ -30,9 +32,25 @@ type Info interface {
...
+	Path *string `yaml:"Path,omitempty"`
+
+	// Ref is component source git ref but can be levaraged for more in future
+	Ref *string `yaml:"Ref,omitempty"`
+
+	// Type is type of component source: git/local/binary
+	SourceType *occlient.CreateType `yaml:"SourceType,omitempty"`
+
+	// Ports is a slice of ports to be exposed when a component is created
+	Ports *[]string `yaml:"Ports,omitempty"`
```
kadel: shouldn't be port `int32`?

pkg/config/config.go
```diff
@@ -30,9 +32,25 @@ type Info interface {
...
 
 	// The builder image to use
-	ComponentType *string `yaml:"ComponentType,omitempty"`
+	Type *string `yaml:"Type,omitempty"`
+
+	// Path is path to binary in current/context dir
+	Path *string `yaml:"Path,omitempty"`
+
+	// Ref is component source git ref but can be levaraged for more in future
+	Ref *string `yaml:"Ref,omitempty"`
```
kadel: here we are storing `Ref` but where would be store url to the git repository?

anmolbabu: In Path

pkg/config/config.go
anmolbabu: Ports I think are specified as for ex: 8080/TCP

pkg/occlient/occlient.go
```diff
@@ -3241,3 +3241,17 @@ func (c *Client) PropagateDeletes(targetPodName string, delSrcRelPaths []string,
 	}
 	return err
 }
+
+// GetCreateType returns enum equivalent of passed component source type or error if unsupported type passed
+func GetCreateType(ctStr string) (CreateType, error) {
```
kadel: why this can' be in config pkg? It should be probably called `SourceType` to be in pair what we have in the config

pkg/config/config.go
kadel: then it needs to be explained in the comment. It is not obvious.

pkg/config/config.go
kadel: Again, the comment doesn't say that. Also in that case, `Path` is not a good name for it. If it can be git url we have to find a different name for it.

pkg/occlient/occlient.go
anmolbabu: I am anyway already moving it as part of my PR I'll do it in #1360 if its ok bcoz altering The CreateType location will have wide impact on this PR and I am anyway touching it in #1360

pkg/config/config.go
anmolbabu: Would SourceLocation be better name? @kadel

kadel: > Would SourceLocation be better name? @kadel

that is better

pkg/config/config.go
girishramnani: so should I add that in the comment?

pkg/occlient/occlient.go
girishramnani: so can this be resolved?

pkg/config/config.go
kadel: yes, if you are touching some code, and you see something wrong fix it right away. Otherwise, it might be never fixed.

anmolbabu: yep

pkg/occlient/occlient.go
anmolbabu: Yeah it will be resolved in #1360 
As long as this PR doesn't suffer cyclic import ;)

pkg/config/config.go
girishramnani: so 2 approaches -
- we remove this typemeta as the proxylocalconfig already has it
- we make this typemeta unexported

girishramnani: @kadel which one do you think works best?

pkg/config/config.go
girishramnani: so just to confirm change `Path` to `SourceLocation`?

kadel: `SourceLocation` still sounds a bit weird, but I can't think of anything better right now. It still makes more sense than just `Path`. Also please explain what this is used for in the comment.

girishramnani: @anmolbabu what is `SourceLocation` used for?

pkg/config/config.go
girishramnani: done, took 2nd approach

pkg/config/config.go
girishramnani: added comment

pkg/config/config.go
girishramnani: added comment

pkg/occlient/occlient.go
girishramnani: resolving this

pkg/config/config.go
girishramnani: And renamed the field

girishramnani:
@kadel @anmolbabu resolved all the comments

pkg/config/config.go
kadel: > Also please explain what this is used for in the comment

As mentioned in https://github.com/redhat-developer/odo/pull/1453/files#r265439578 `SourceLocation` can also bit git url, comment needs to say that

pkg/config/config.go
```diff
@@ -48,6 +67,13 @@ type ComponentSettings struct {
 
 // LocalConfig holds all the config relavent to a specific Component.
 type LocalConfig struct {
+	typeMeta          metav1.TypeMeta   `yaml:",inline"`
```
kadel: Now `LocalConfig` and `proxyLocalConfig` have complety different structures.

`proxyLocalConfig` has `metav1.TypeMeta` embeded, and `LocalConfig` has it under the `typMeta`

```
LocalConfig {
  typeMeta: {
      kind: "LocalConfig",
      apiversion: "odo.openshift.io/v1alpha1"
  },
  componentSettings {
      Name: "mycomponent",
       .....
  }
}
```
vs 
```
proxyLocalConfig {
  kind: "LocalConfig",
  apiversion: "odo.openshift.io/v1alpha1"
  componentSettings {
      Name: "mycomponent",
       .....
  }
}
```

Is that what you wanted?

girishramnani: to hide the `typemeta` that's the only way @kadel

pkg/config/config.go
girishramnani: I am using `GetConfiguration` in `set` and `unset` and I don't think we can replace that.

pkg/config/config.go
girishramnani: and that's what I said in one of the comments that both `localconfig` and `proxylocalconfig` need to have same structure which will not be possible if we hide the `typemeta`

pkg/config/config.go
girishramnani: added @kadel

pkg/config/config.go
kadel: But we could still remove it, no?? 
What is the reason you want to keep it there?

girishramnani: https://github.com/redhat-developer/odo/pull/1453#discussion_r265453192

Took the 2nd approach

girishramnani: I want the localconfig to be the primary source of truth, proxyconfig is just a transient struct

girishramnani: And even the proxyconfig gets the type meta from localconfig, so in my opinion we should keep it. But I will keep the final decision to you

anmolbabu:
Please address @kadel comments
Rest LGTM

pkg/config/config.go
kadel: You are using it just to check if the config is set or not, but not for anything else. You can create `IsSet` function for that that will be much simpler.
This will have a huge maintenance cost, to manage two different ways to read the config and making sure that you handle default values etc.. in the same way.

girishramnani: I get it but its not just about checking if the value is set, if the value is set, we show the user
`the value is already set. The current value being %v`. 
I will not be able to show the current value if we change this to `IsSet`. We can consider a reflection mechanism in the future where the `GetConfiguration` instead of directly accessing the attributes uses the getters ( via reflection ). This way the maintain ability stays.

kadel: To be honest, I would rather stop showing what the current value of the config is than having to manage the same logic in two places.

kadel:
:-(

girishramnani:
Now that's interesting

cdrage:
Wat. Any clue what happened @mohammedzee1000 ? Did we discover an OpenShift bug? Haha

PR #1455, opened 2019-03-14
Removing login and json tests from circleci

mohammedzee1000:
These tests now run on prow and are not needed to be in circle ci as well

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,14 +74,6 @@ jobs:
         command: |
             cd ${HOME}/go/src/github.com/redhat-developer/odo
             make test-main-e2e
-
-    - run: 
-        name: Login e2e test
-        command: |
-            cd ${HOME}/go/src/github.com/redhat-developer/odo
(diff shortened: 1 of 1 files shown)


syamgk:
So this means we have enough confidence in Openshift CI and the tests on prow are dependable at the moment ?

mohammedzee1000:
> So this means we have enough confidence in Openshift CI and the tests on prow are dependable at the moment ?

yes they are they are running on most PRs successfully.

surajnarwade:
@mohammedzee1000 , can you please also create issue for the same and link it to this PR ?

amitkrout:
> > So this means we have enough confidence in Openshift CI and the tests on prow are dependable at the moment ?
> 
> yes they are they are running on most PRs successfully.

Right, After fix of #1460 test on prow seems more stable than earlier.

mohammedzee1000:
@surajnarwade makes sense, will do

mohammedzee1000:
We no longer need this
/close

PR #1457, opened 2019-03-14
remove GetConfiguration from config and preference

girishramnani:
Removes the `GetConfiguration` method and adds `IsSet` method. Uses `IsSet` for checking if the config or preference is set or not.

No issue

## How to test changes?
check if `odo config` and `odo preference` work correctly


kadel:
/lgtm

/hold
waiting for the second review

mohammedzee1000:
/hold
let ci go green
/refresh

PR #1465, opened 2019-03-15
[WIP] Add documentation outlining machine readable output

cdrage:
Adds a document that outlines all possible examples / options for each
command available.

Closes https://github.com/redhat-developer/odo/issues/1464

### DIFF
--- a/docs/machine-output.md
+++ b/docs/machine-output.md
@@ -1,0 +1,125 @@
+# Machine Readable Output
+
+This document outlines all the machine readable output options and examples.
+
+`$ odo url create -o json | jq`
+
+```json
+{
+  "kind": "url",
+  "apiVersion": "odo.openshift.io/v1alpha1",
+  "metadata": {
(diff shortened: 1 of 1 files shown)


cdrage:
Note, this is a WIP as I'm probably missing a lot more commands, although some are being deprecated, for example `odo app create` since we're removing app, as well as `odo app list`..

docs/machine-output.md
```diff
@@ -0,0 +1,125 @@
+# Machine Readable Output
+
+This document outlines all the machine readable output options and examples.
+
+`$ odo url create -o json | jq`
```
mik-dass: I think we should mention before that the `jq` tool is required for pretty formatting

surajnarwade: +1

cdrage: Done :+1:

cdrage:
I'm going to close this for now :+1: will revisit this in the future.

PR #1494, opened 2019-03-20
Remove demo file

cdrage:
Removes the demo folder / file


girishramnani:
merging this as there is no code change and it doesn't impact any tests either

Issue #1470, opened 2019-03-18
Stability of our e2e tests


kadel:
I'm seeing a lot of failed tests even on the PRs that are not changing anything in the source code. 

We need to make sure that we can count on our e2e tests.

Some examples:
 - https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/redhat-developer_odo/1467/pull-ci-redhat-developer-odo-master-e2e-core/116
- https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/redhat-developer_odo/1462/pull-ci-redhat-developer-odo-master-e2e-core/110

cdrage:
Another issue from PR: https://github.com/redhat-developer/odo/pull/1275

Looks to be a timeout to the cluster: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/redhat-developer_odo/1275/pull-ci-redhat-developer-odo-master-e2e-core/160

amitkrout:
Some suspicious clusterup/timeout failure in prow - https://coreos.slack.com/archives/CEKNRGF25/p1553001003940200 asked by @mohammedzee1000

amitkrout:
openshift-install times out creating infrastructure resources - https://bugzilla.redhat.com/show_bug.cgi?id=1690069

amitkrout:
Unless verbose is enabled for each test on prow It's hard to root cause the failure. I have observed in recent time these failure are more often consistent. Some failures are from CI side and some test failure, could be the side effect of unstable prow setup which causes the odo test failures. 

So I would recommend enable the test verbose to crack down the odo test failure on prow setup as on aws the chance of reproduce the same failure is less. It also helps on aws too if any test failure is introduced. Issue from OpenShift CI side @mohammedzee1000 is taking care anyway.

amitkrout:
```
Running command: odo create odo-1553168711/nodejs push-odoignore-flag-example --local /var/folders/c9/fb55pf_568dfd1fzj9w1v4n00000gn/T/odo880818179/nodejs-ex
 •  Checking component  ...
 ✓  Checking component
 •  Checking component version  ...
 ✓  Checking component version
 ✓  Initializing 'push-odoignore-flag-example' component
 •  Creating component push-odoignore-flag-example  ...
 ✓  Creating component push-odoignore-flag-example
 ✓  Component 'push-odoignore-flag-example' was created and port 8080/TCP was opened
[...]

PR #1499, opened 2019-03-21
[Do not merge. Only for analysis]Enable test verbose log

amitkrout:

Due to inconsistency test behaviour on OpenShift CI, i am enabling test verbose log to identify the cause of test failure. 

Test those are depend on service catalog are not part of prow setup. 

Analysis of #1470

## How to test changes?
More details log in test result

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,54 +93,54 @@ test:
 .PHONY: test-main-e2e
 test-main-e2e:
 ifdef TIMEOUT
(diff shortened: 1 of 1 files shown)


mohammedzee1000:
This should be OK for now,  but ideally this should be an option like timeout

amitkrout:
Fixed in #1517

PR #1505, opened 2019-03-21
redhat-developer/odo -> openshift/odo

kadel:


mohammedzee1000:
Awesome CI all green :)

PR #1508, opened 2019-03-21
Group TravisCI jobs

kadel:
group our e2e test into groups, to cut down cluster setup time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,6 @@ jobs:
       script:
         - make bin
         - make validate
-        - make test
-
-    - <<: *base-test
-      name: "Test coverage"
-      script:
-        - make bin
-        - make validate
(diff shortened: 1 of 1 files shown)


kadel:
ping @amitkrout @mohammedzee1000

.travis.yml
```diff
@@ -75,69 +50,40 @@ jobs:
```
amitkrout: Can you add 
```
- travis_wait make test-source-e2e
- travis_wait make test-json-format-output
```
Can be added here, so that we have three test file run per each cluster up

kadel: not right now. It will break things, as tests are not cleaning up properly, and there is a mess with projects

amitkrout: Got it. Thanks

PR #1516, opened 2019-03-25
Scripts to run on OpenShift CI

kadel:
This abstracts all commands that are run on OpenShift CI into separate scripts in this repository. 
This will allow us to modify those scripts without touching openshift/release repo.

Idea is that jobs will just run `make openshiftci-presubmit` and `make openshiftci-master`.

https://github.com/openshift/release/blob/0f4bf0f581103c14f54dd6eee548c63c3455ab36/ci-operator/config/openshift/odo/openshift-odo-master.yaml

https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/odo/openshift-odo-master-presubmits.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,5 +206,9 @@ vendor-update:
 	glide update --strip-vendor
 
 .PHONY: openshiftci-presubmit-e2e
-openshiftci-presubmit:
+openshiftci-presubmit-e2e:
 	./scripts/openshiftci-presubmit-e2e.sh
+
+.PHONY: openshiftci-presubmit-unittests
+openshiftci-presubmit-unittests:
+	./scripts/openshiftci-presubmit-unittests.sh
--- a/Makefile
+++ b/Makefile
@@ -204,3 +204,7 @@ upload-packages:
 .PHONY: vendor-update
 vendor-update:
 	glide update --strip-vendor
+
+.PHONY: openshiftci-presubmit-e2e
+openshiftci-presubmit:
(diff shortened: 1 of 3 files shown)


surajnarwade:
Travis was failing, restarted the job

kadel:
ping @mohammedzee1000

girishramnani:
/ hold

Makefile
```diff
@@ -204,3 +204,7 @@ upload-packages:
 .PHONY: vendor-update
 vendor-update:
 	glide update --strip-vendor
+
+.PHONY: openshiftci-presubmit-e2e
```
girishramnani: shouldn't this be `openshiftci-presubmit`? @kadel

Makefile
```diff
@@ -204,3 +204,7 @@ upload-packages:
...
+openshiftci-presubmit:
```
girishramnani: or this can be `openshiftci-presubmit-e2e`

Makefile
kadel: we could have more than one presubmit job.
Other will be named something like `openshiftci-presubmit-integration`

Makefile
kadel: ah, yes, that is my mistake

Makefile
kadel: sorry, the next line is wrong

kadel:
/hold cancel

@girishramnani thx, for the review, the Makefile should be ok now

PR #1518, opened 2019-03-26
Fixing error when user lists component in directory without one

### LINKED ISSUE #1485: `odo component list` in a non component directory
> @anmolbabu when running the `odo component list` from a non component directory I get an error which is not so helpful
> 
> ```
> [gramnani@girish Github]$ odo component list
>  ✗  namespacing: component name cannot be blank
> ```
> maybe something like
> `the current directory doesn't represent a odo component` or something?

we can fix all those issues separately.
Every command that needs to read information about a component or application should exit whit nice error.

_Originally posted by @girishramnani in https://github.com/redhat-developer/odo/pull/1360#issuecomment-474776233

mohammedzee1000:

See title
 - Adding fuction that allows to check if a config file exists
 - Updating component list to check existence of odo config and fail
   if it does not exist
 - Adding regression tests

fixes #1485 

## How to test changes?
```
$ cd nodejs-ex
$ pwd
/redacted//nodejs-ex
$ odo component list
ACTIVE     NAME            TYPE
*          nodejs-vypi     nodejs
$ cd someotherdir
$ odo component list
 ✗  the current directory does not represent an odo component

```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -95,28 +95,14 @@ func getLocalConfigFile(cfgDir string) (string, error) {
 	return filepath.Join(cfgDir, ".odo", configFileName), nil
 }
 
-// LocalConfigExists checks if a config file exists.
-// If not it returns false
-func LocalConfigExists(cfgDir string) (bool, error) {
-	localConfig, err := getLocalConfigFile(cfgDir)
-	if err != nil {
-		return false, err
-	}
-	_, err = os.Stat(localConfig)
-	if os.IsNotExist(err) {
-		return false, nil
-	}
-	return true, nil
-}
-
 // New returns the localConfigInfo
 func New() (*LocalConfigInfo, error) {
-	return NewLocalConfigInfo("")
+	return NewLocalConfigInfo("", false)
 }
 
 // NewLocalConfigInfo gets the LocalConfigInfo from local config file and creates the local config file in case it's
 // not present then it
-func NewLocalConfigInfo(cfgDir string) (*LocalConfigInfo, error) {
+func NewLocalConfigInfo(cfgDir string, errorNoLocalConfig bool) (*LocalConfigInfo, error) {
 	configFile, err := getLocalConfigFile(cfgDir)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get odo config file")
@@ -128,6 +114,9 @@ func NewLocalConfigInfo(cfgDir string) (*LocalConfigInfo, error) {
 
 	// if the config file doesn't exist then we dont worry about it and return
 	if _, err = os.Stat(configFile); os.IsNotExist(err) {
+		if errorNoLocalConfig {
+			return nil, fmt.Errorf("the current directory does not represent an odo component.\nMaybe use 'odo create' to create component here or switch to directory with a component")
+		}
 		return &c, nil
 	}
 	err = getFromFile(&c.LocalConfig, c.Filename)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -95,6 +95,20 @@ func getLocalConfigFile(cfgDir string) (string, error) {
 	return filepath.Join(cfgDir, ".odo", configFileName), nil
 }
 
+// LocalConfigExists checks if a config file exists.
+// If not it returns false
+func LocalConfigExists(cfgDir string) (bool, error) {
+	localConfig, err := getLocalConfigFile(cfgDir)
+	if err != nil {
+		return false, err
+	}
+	_, err = os.Stat(localConfig)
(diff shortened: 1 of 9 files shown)


tests/e2e/component.go
```diff
@@ -61,6 +61,18 @@ func componentTests(componentCmdPrefix string) {
...
 	})
 
+	Context("Regression : listing component outside of component directory should fail", func() {
+		It("creates a component from local context, tries to list components from outside and fails", func() {
+			componentName := generateTimeBasedName("nodejs-ex")
+			runCmdShouldPass(componentCmdPrefix + " create nodejs " + componentName + " --git https://github.com/openshift/nodejs-ex")
+			session := runCmdShouldFail("odo component list")
+			Expect(session).To(ContainSubstring("doesn't represent"))
+			Expect(session).To(ContainSubstring("odo component"))
+			// clean up
```
girishramnani: would be a good idea to have a single `ContainSubstring` which has the whole message

pkg/odo/cli/component/list.go
```diff
@@ -46,11 +47,19 @@ func (lo *ListOptions) Complete(name string, cmd *cobra.Command, args []string)
 
 // Validate validates the list parameters
 func (lo *ListOptions) Validate() (err error) {
+	configExists, err := odoconfig.LocalConfigExists("")
+	if err != nil {
+		return err
+	}
+	if !configExists {
+		return fmt.Errorf("the current directory does not represent an odo component")
```
girishramnani: would be great to have give a hint to the developer related to what they should do next like -
```
the current directory .... component
use "odo create" to create a component in the directory # maybe a better phrasing
```

tests/e2e/component.go
```diff
@@ -61,6 +62,20 @@ func componentTests(componentCmdPrefix string) {
 		})
 	})
 
+	Context("Regression : listing component outside of component directory should fail", func() {
+		It("creates a component from local context, tries to list components from outside and fails", func() {
+			componentName := generateTimeBasedName("nodejs-ex")
+			dirName := generateTimeBasedName("context_dir")
+			runCmdShouldPass(fmt.Sprint("mkdir ", dirName))
+			runCmdShouldPass(componentCmdPrefix + " create nodejs " + componentName + " --git https://github.com/openshift/nodejs-ex --context " + dirName)
+			session := runCmdShouldFail("odo component list")
```
girishramnani: this will not fail as the code above this line generates a `.odo` folder. this is the reason for the cryptic 
```
0
not equal to
0
```
in the tests

pkg/odo/cli/component/list.go
mohammedzee1000: Addressed, pushing shortly

tests/e2e/component.go
mohammedzee1000: Trying something different lets see

tests/e2e/component.go
mohammedzee1000: Addressed, pushing shortly

pkg/config/config.go
```diff
@@ -95,6 +95,20 @@ func getLocalConfigFile(cfgDir string) (string, error) {
 	return filepath.Join(cfgDir, ".odo", configFileName), nil
 }
 
+// LocalConfigExists checks if a config file exists.
+// If not it returns false
+func LocalConfigExists(cfgDir string) (bool, error) {
```
anmolbabu: I was thinking if it is better instead to add a bool flag to indicate if NewLocalConfigInfo and/or New should return an error if config does not exist it might be ok to even have it as a optional arguement(variable length args) with the current behaviour of those functions as default ?

tests/e2e/component.go
anmolbabu: ok I might be missing something but why does it not(never) fail

The line above takes `--context` but not component list

mohammedzee1000: Yea it is cryptic indeed. Anyway it should fail now. I am renaming odo directory before and renaming it back after

pkg/config/config.go
mohammedzee1000: Possible maybe. But having a specialized function that does not create but rather just checks if it exists helps. Overloading one function with too much functionality could become a problem. Also, this might end up being a fix for now as component list is not yet tied to `--context`, so both these cases will need to be covered ie it checks for `--context` as well.

anmolbabu: Well IMO, this function is actually replicating initial logic from NewLocalConfigInfo except 1 conditional handling which can be easily controlled by a bool.

mohammedzee1000: Alright i will take a look :)

mohammedzee1000: Updated

Issue #1447, opened 2019-03-12
new odo workflow: odo component create and odo config behaviour


kadel:
if `.odo/config.yaml` doesn't exist:
   -  `odo component create` - will create new `.odo/config.yaml`
   -  `odo config` - fails when there is no local config

 
if `.odo/config.yaml` already exist:
  - `odo component create` - fails with error message explaining that component aready exist
  - `odo config`  - ok, adds/modifies/deletes values in config file

PR #1526, opened 2019-03-27
config behavior according to new workflow

girishramnani:
now the `odo config` cannot create a config, only `odo create` can. making `odo config` compliant to the intended workflow

fixes https://github.com/openshift/odo/issues/1447

## How to test changes?
- `odo create` twice in one directory will fail
- `odo config set`, `odo config view` and `odo config unset` will fail in a non-component directory


pkg/config/config.go
```diff
@@ -126,6 +126,17 @@ func NewLocalConfigInfo(cfgDir string, errorNoLocalConfig bool) (*LocalConfigInf
 	return &c, nil
 }
 
+// LocalConfigFileExists checks if the local config file exists or not. an error while getting the
+// file name is also considered as file not existing
+func LocalConfigFileExists(cfgDir string) bool {
```
cdrage: I am refactoring some code in my https://github.com/openshift/odo/pull/1446 PR

We may have to move this code into utils as I have a feeling that this should be applied to most, if not all commands before execution. Checking to see if the config file exists.

girishramnani: There is a flag in the `NewLocalConfigInfo` which will give u an error in case file doesn't exist which can be used everywhere else. I can maybe think of a way to reuse that.

girishramnani:
the feature implemented in this PR is implemented in a more general manner in this PR https://github.com/openshift/odo/pull/1532

girishramnani:
so closing this PR

girishramnani:
Reopening the PR because the https://github.com/openshift/odo/pull/1532 PR is not adding the `not allow odo create in a already created component directory `

girishramnani:
> When running interactive mode it fails too late.
> 
> ```
> ▶ odo component create
> ? Which component type would wish to create nginx
> ? Which version of 'nginx' component type would you wish to create 1.10
> ? Location of context component, relative to '/Users/tkral/tmp/odo-examples/nodejs-ex' /Users/tkral/tmp/odo-examples/nodejs-ex
> ? Which input type would wish to use for the component local
> ? How would you wish to name the new component nginx-urjp
> ? Which application do you want the commponent to be associated with nodejs-ex-wzxy
> ? Which project go you want the component to be created in tkral
> ? Do you wish to set advanced options No
>  ✗  this directory already contains a component
> ```
> It needs to fail before the first question

then I need to add the check before `Complete` phase as thats where interactive mode occurs @kadel

pkg/odo/cli/component/create.go
```diff
@@ -235,6 +235,13 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 	if err != nil {
 		return errors.Wrap(err, "failed intiating local config")
 	}
+
+	// check to see if config file exists or not, if it does that
+	// means we shouldn't allow the user to override the current component
+	if co.localConfigInfo.ConfigFileExists() {
+		return errors.New("this directory already contains a component")
```
girishramnani: This check is in the `Complete` phase because of https://github.com/openshift/odo/pull/1526#issuecomment-481162707

girishramnani:
added test for this @kadel

girishramnani:
This paves the way for more tests being added to the new files.

girishramnani:
> 1 minor comment in the interest of having a concise test suite, can we add the test as part of an existing e2e instead of a new e2e ? Just a thought but otherwise LGTM
> 
> Also lets wait for the gate-keepers to signal green

The tests are failing, resolving the issues

kadel:
> LGTM! :) no changes to code and awesome job on the tests.

please use `/lgtm`

kadel:
> ### [@girishramnani](/girishramnani) [girishramnani](/girishramnani) requested a review from [kadel](/kadel) [15 hours ago](#event-2274885662)

approved long time ago. Someone else needs to have a look.


@girishramnani can you please rebase this PR?

Issue #1486, opened 2019-03-20
odo app list from non component directory


kadel:
to add to @surajnarwade point 
```
$ odo app list
```
This command returns the applications in my project `myproject` from any folder. I am not sure if that should happen @anmolbabu

_Originally posted by @girishramnani in https://github.com/redhat-developer/odo/pull/1360#issuecomment-474749146_

`odo app list`
 - in a directory with `.odo` - list all application in the project defined in the config file
 - in a director without `.odo` - show error that there is no component configuration 

`odo app list --context /my/project`
  - similar to `odo app list`, but uses the context dir instead of the current directory

`odo app list --project`
 - show all applications in given project - doesn't check for the `.odo` folder
 - this command can be run from any directory regardless of `.odo` presence

kadel:
@mohammedzee1000 I have updated the description

mohammedzee1000:
@kadel  i don't think app list implements `--context` so that might be outside purview of this at the moment

Issue #1520, opened 2019-03-26
new workflow:  odo component commands


kadel:
Each `odo component` command has optional `--context` argument.
That defines what is the component that the action shoudl be performed on.
By default context is `./` 

####  create 
creates component config (definition) in the context directory

#### delete 
flags : --force --context
deletes component from the cluster (undeploy), configuration in intacted

####  describe 
flags:  --context : Describe the given component
show all information about the current component, including information if it is deployed or not

#### link
flags:  --app --project --wait --context
Link current component to a service or component
do live change, storing in config file will depend on whether service is stored in the config or not

TODO: `--app` and `--project`

#### unlink
flags:  --app --port --project --context
Unlink current component from a service or component
TODO: `--app` and `--project`

#### list
flags:  --app --project --all --context
List all components in the current application  (the same application as the current component is)
`--application` flag can be specified to list components in the different application  (if not specified defaults to the application defined in the context)
`--project` flag can be specified to list components in the different project (if not specified defaults to the project defined in the context)

if we have following config
```
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Application: testing
  Project: myproj
  Name: php
```

`odo component list` - list all components in `testing` app in `myproj`
`odo component list --application  myapp` - list all components in `myapp` app in `myproj`
`odo component list --project  secondproj` - list all components in `testing` app in `secondproj`
`odo component list --application  myapp --project  secondproj`- list all components in `myapp` app in `secondproj`. This command is only `odo component list` that be run outside of the directory  with `.odo` o

All the above commands can be run with `--context` to use config file in different directory than the current one (expect the last one)

#### log|logs 
Retrieve the log for the given component
flags:   --app --follow --project --context

`--project` and `--app`  have the same behavior like in `list` command

#### push 
Push source code to a component
flags:  --show-log --context

show-log implementation in #1082 

#### watch 
Watch for changes, update component on change
flags:  --delay --ignore --show-log --context

#### update 
TODO:

mohammedzee1000:
I personally think context resolution should happen in context.go. With commands not having to worry about it. They should instead complain if they dont have enough information. 

For eg, If user has specified a context with no active component then context does is set but without active component and when user tries to perform any op that needs component while not providing --component flag. Then it complains asking user to set/create a component first

kadel:
> Then it complains asking user to set/create a component first or provide `--component`

[...]

PR #1532, opened 2019-03-28
Short circuiting invalid context file in context.go

mohammedzee1000:
- Does above
 - Removes old fix for component list as this covers that
   more easily
 - Does some breakup of functions of context.go along the way.

#1486 

## How to test changes?

Go into a directory where there is no .odo file and try `odo app list` or `odo component list`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,11 +2,13 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
 	"github.com/openshift/odo/pkg/util"
@@ -526,4 +528,35 @@ func GetSrcType(ctStr string) (SrcType, error) {
 	default:
 		return NONE, fmt.Errorf("Unsupported component source type: %s", ctStr)
 	}
+}
+
+// GetOSSourcePath corrects the current sourcePath depending on local or binary configuration,
+// if Git has been passed, we simply return the source location from LocalConfig
+// this will get the correct source path whether on Windows, macOS or Linux.
+func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
+
+	sourceType := lci.GetSourceType()
+	sourceLocation := lci.GetSourceLocation()
+
+	if sourceLocation == "" {
+		return "", fmt.Errorf("Blank source location provided")
+	}
+
+	if sourceType == GIT {
+		glog.V(4).Info("Git source type detected, not correcting SourcePath location")
+		return sourceLocation, nil
+	}
+
+	// Validation check if the user passes in a URL despite us being LOCAL or BINARY
+	u, err := url.Parse(sourceLocation)
+	if err != nil || (u.Scheme == "https" || u.Scheme == "http") {
+		return "", fmt.Errorf("URL: %s passed even though source type is: %s", sourceLocation, sourceType)
+	}
+
+	// Always piped to "fromslash" so it's correct for the OS..
+	// after retrieving the sourceLocation we will covert it to the
+	// correct source path depending on the OS.
+	sourceOSPath := filepath.FromSlash(lci.GetSourceLocation())
+
+	return sourceOSPath, nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,13 +2,11 @@ package config
 
 import (
 	"fmt"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 
-	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
 	"github.com/openshift/odo/pkg/util"
@@ -79,8 +77,9 @@ type proxyLocalConfig struct {
 // LocalConfigInfo wraps the local config and provides helpers to
 // serialize it.
 type LocalConfigInfo struct {
-	Filename    string `yaml:"FileName,omitempty"`
-	LocalConfig `yaml:",omitempty"`
+	Filename         string `yaml:"FileName,omitempty"`
+	LocalConfig      `yaml:",omitempty"`
+	configFileExists bool
 }
 
 func getLocalConfigFile(cfgDir string) (string, error) {
@@ -101,26 +100,25 @@ func getLocalConfigFile(cfgDir string) (string, error) {
 
 // New returns the localConfigInfo
 func New() (*LocalConfigInfo, error) {
-	return NewLocalConfigInfo("", false)
+	return NewLocalConfigInfo("")
 }
 
 // NewLocalConfigInfo gets the LocalConfigInfo from local config file and creates the local config file in case it's
 // not present then it
-func NewLocalConfigInfo(cfgDir string, errorNoLocalConfig bool) (*LocalConfigInfo, error) {
+func NewLocalConfigInfo(cfgDir string) (*LocalConfigInfo, error) {
 	configFile, err := getLocalConfigFile(cfgDir)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get odo config file")
 	}
 	c := LocalConfigInfo{
-		LocalConfig: NewLocalConfig(),
-		Filename:    configFile,
+		LocalConfig:      NewLocalConfig(),
+		Filename:         configFile,
+		configFileExists: true,
 	}
 
 	// if the config file doesn't exist then we dont worry about it and return
 	if _, err = os.Stat(configFile); os.IsNotExist(err) {
-		if errorNoLocalConfig {
-			return nil, fmt.Errorf("the current directory does not represent an odo component.\nMaybe use 'odo create' to create component here or switch to directory with a component")
-		}
+		c.configFileExists = false
 		return &c, nil
 	}
 	err = getFromFile(&c.LocalConfig, c.Filename)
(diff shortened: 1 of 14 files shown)


pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,42 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
 	}
 }
 
+func checkValidConfig(command *cobra.Command, configFileName string) {
```
girishramnani: error should be returned

pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,42 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
...
+	allowedFirstChildCommandList := []string{"project", "catalog", "login", "logout"}
+
+	if command.HasParent() {
+		rootCommand := command.Root()
+		// if command is create operation or push operation just allow it
+		if command.Name() == "create" || command.Name() == "push" {
+			return
+		}
+		// Find the first child of the command. At end of this, c will be the first child
+		p := command.Parent()
```
girishramnani: extracting this out in a function like maybe `getFirstChildOfCommand()` will make this more understandable. just a suggestion

pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,42 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
 	}
 }
 
+func checkValidConfig(command *cobra.Command, configFileName string) {
+	_, err1 := config.NewLocalConfigInfo(configFileName, true)
+	allowedFirstChildCommandList := []string{"project", "catalog", "login", "logout"}
```
girishramnani: maybe add a comment which explains that these commands dont need config and hence can run from directories which do not have a `.odo`

pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,42 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
 	}
 }
 
+func checkValidConfig(command *cobra.Command, configFileName string) {
+	_, err1 := config.NewLocalConfigInfo(configFileName, true)
```
girishramnani: why call this `err1`?

pkg/odo/genericclioptions/context.go
```diff
@@ -147,6 +169,46 @@ func newContext(command *cobra.Command, createAppIfNeeded bool) *Context {
...
 		}
 	}
+	return app
+}
+
+// resolveComponent resolves component
+func resolveComponent(command *cobra.Command, configFileName string, context *Context) string {
+	var cmp string
+	lci, err := config.NewLocalConfigInfo(configFileName, false)
+	util.LogErrorAndExit(err, "could not get component settings from config file")
```
girishramnani: I would like all the logging to be doing by the `GenericRun` and these function can just return error. just a suggestion. maybe @kadel can confirm this

mohammedzee1000: Agreed. But that will require a bigger refactor as this is pattern in most of the file. Maybe an issue to track it

pkg/odo/genericclioptions/context.go
mohammedzee1000: Done

pkg/odo/genericclioptions/context.go
mohammedzee1000: Added

mohammedzee1000:
Will add more tests once people are ok with flow

pkg/odo/genericclioptions/context.go
```diff
@@ -147,6 +186,46 @@ func newContext(command *cobra.Command, createAppIfNeeded bool) *Context {
...
+	}
+	return cmp
+}
+
+// newContext creates a new context based on the command flags, creating missing app when requested
+func newContext(command *cobra.Command, createAppIfNeeded bool) *Context {
+	client := client(command)
+
+	// resolve project
+	ns := resolveProject(command, client)
```
anmolbabu: Can this be done below after the config resolution?

pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,59 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
...
+			c = p
+			p = c.Parent()
+		}
+	}
+	return nil
+}
+
+func checkValidConfig(command *cobra.Command, configFileName string) {
+	_, err := config.NewLocalConfigInfo(configFileName, true)
+	// List of allowed firstchild command
```
anmolbabu: err ignored?

pkg/odo/genericclioptions/context.go
```diff
@@ -72,6 +72,59 @@ func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, er
...
+			// Traverse backwards making current command as the parent and parent as the grandparent
+			c = p
+			p = c.Parent()
+		}
+	}
+	return nil
+}
+
+func checkValidConfig(command *cobra.Command, configFileName string) {
+	_, err := config.NewLocalConfigInfo(configFileName, true)
```
anmolbabu: NewLocalConfigInfo actually reads from file everytime its invoked
I believe the aim of this PR is also to decrease the number of times its called which is the reason I believe why we are moving config resolution to context
In my opinion component settings is parallel to context except for only one field in context which is client so should Context struct also be made to contain LocalConfigInfo? and remove project, component name and app from context bcoz they are there in LocalConfigInfo

pkg/odo/genericclioptions/context.go
mohammedzee1000: Not really. It gets caught as the config is read again later with 2nd parameter set to false. This is specifically meant to fail if file is not found

pkg/odo/genericclioptions/context.go
anmolbabu: in which case there probably is no need to resolve each of project, component and app in my opinion bcoz now the single source of truth in my opinion should be config

mohammedzee1000: Unless overridden by respective parameters :)

anmolbabu: https://github.com/openshift/odo/issues/1520#issuecomment-477943538

No flags related to component or app needs to be passed to any command other than component create, if passed they'll be ignored or will error out

mohammedzee1000: Cool then

pkg/odo/genericclioptions/context.go
mohammedzee1000: done

pkg/odo/genericclioptions/context.go
mohammedzee1000: Handled

pkg/odo/genericclioptions/context.go
mohammedzee1000: Updated. i really feel like removing those flags right now. Already have it in my git stash :)

pkg/odo/genericclioptions/context.go
mohammedzee1000: Did the best i could for now, it is returned and handled by caller. But will require bigger refactor

mohammedzee1000:
@cdrage it seems to be like it already is as far as context.go is concerned. What is important is shifting to making use of it

mohammedzee1000:
/cc @anmolbabu @girishramnani 
Anything else?

pkg/odo/genericclioptions/context.go
```diff
@@ -147,6 +211,47 @@ func newContext(command *cobra.Command, createAppIfNeeded bool) *Context {
...
+	lci, err := getValidConfig(command)
+	if err != nil {
+		util.LogErrorAndExit(err, "")
+	}
+
+	// resolve project
+	ns := resolveProject(command, client, lci)
+
+	// resolve application
+	app := resolveApp(command, createAppIfNeeded, lci)
```
anmolbabu: I don't see the importance of this bcoz it's only going to be component create that will not pass these flagsas far as I can think of but probably defer it to post new workflow implementation for all commands after which we can probably revisit this

mohammedzee1000:
> Could we have e2e tests with the new framework? Suggestions @kadel

I am not sure if we want to do this right now. Lets setup separate prs to move all the tests over properly

pkg/odo/genericclioptions/context.go
mohammedzee1000: Yes we can revisit this after. Its more about making things clear and concise. It also makes this function much smaller which is what is favoured, especially in go

amitkrout:
> > Could we have e2e tests with the new framework? Suggestions @kadel
> 
> I am not sure if we want to do this right now. Lets setup separate prs to move all the tests over properly

Anyway we have stable e2e test and only effort will be migrate them to new structure. So separate pr would be fine.

mohammedzee1000:
Interesting fail
/retest

mohammedzee1000:
Cluster problem
/retest

pkg/config/config.go
```diff
@@ -77,8 +77,9 @@ type proxyLocalConfig struct {
 // LocalConfigInfo wraps the local config and provides helpers to
 // serialize it.
 type LocalConfigInfo struct {
-	Filename    string `yaml:"FileName,omitempty"`
-	LocalConfig `yaml:",omitempty"`
+	Filename         string `yaml:"FileName,omitempty"`
+	LocalConfig      `yaml:",omitempty"`
+	ConfigFileExists bool
```
anmolbabu: Needs to be an un-exported field exposed if required only be a getter and no setter needs to be there
as this ideally needs to be set only by New... atleast for now

mohammedzee1000: Makes sense, updating

mohammedzee1000:
there was an actual issue in the test
/cc @girishramnani

Issue #1131, opened 2018-12-12
Write command runner for oc and odo in test package


amitkrout:
[kind/Enhancement]

## Which functionality do you think we should update/improve?
The current implementation of running command in the [test suite](https://github.com/redhat-developer/odo/blob/master/tests/e2e/e2e.go#L17) looks premature. Moreover this testsuite won’t work in windows environment. So creating a generic command runner for both would be helpful

## Why is this needed? 
To run testsuite on all supported platform

girishramnani:
we should consider after beta

amitkrout:
Pr #1612 has the fix of ```oc``` and ```odo``` command runner.

amitkrout:
So closing it as it has been fix via #1559.

PR #1535, opened 2019-03-28
Clean e2e test example

kadel:
This is an example of how to write e2e tests using ginkgo that perform automatic setup and teardown.

To run this do 
```
go test -v github.com/openshift/odo/tests/e2e -ginkgo.focus="Example of a clean test"  -ginkgo.slowSpecThreshold=120 -ginkgo.randomizeAllSpecs
``` 
` -ginkgo.randomizeAllSpecs` runs all test in random order.  This will make sure that we don't create dependencies between tests specs (`It(` functions) even accidntaly, as the order of in which tests are run can't be prediceted.

You can also run tests in parallel:
```
ginkgo -nodes=2 -focus="Example of a clean test" slowSpecThreshold=120 -randomizeAllSpecs  tests/e2e/
```
`nodes` sets number of parallel executions

Note that you have to use `ginkgo` (`go get github.com/onsi/ginkgo/ginkgo` than `$GOPATH/bin/ginkgo`) to tests in parallel

I had to updated Ginkgo as the version that we were using doesn't support `JustAfterEach`


kadel:
ping @amitkrout @girishramnani

amitkrout:
Thanks for the pr in a quick time. Tomorrow will have a look

girishramnani:
great thinking of using the latest version of ginkgo :+1:

tests/e2e/helper/helper_run.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+)
+
+// and returns the stdout, stderr and exitcode
+func cmdRunner(cmdS string) (string, string, int) {
+	//TODO: this needs to be os independent
```
girishramnani: lets document this in an issue so that @amitkrout can work on making this os independent

girishramnani:
I am not seeing any helpers related to `waitForCmdOut` @kadel are we planning to add those or use the one in the utils?

tests/e2e/helper/helper_run.go
amitkrout: @girishramnani We have issue for it #1131

tests/e2e/cleantest_test.go
```diff
@@ -0,0 +1,145 @@
...
+})
+
+// AfterSuite is run before whole test suite
+// there can be only one AfterSuite
+var _ = AfterSuite(func() {
+	if originalDir != "" {
+		helper.Chdir(originalDir)
+	}
+	if originalProject != "" {
+		helper.OcSwitchProject(originalProject)
```
mohammedzee1000: Could this not cause an overlap? if oc active project is depended on. What happens if one of the parallel runs gets here first and sets a different active project, while another is working with a different project (in a single cluster context)

glide.yaml
```diff
@@ -26,9 +26,9 @@ import:
   repo:    https://github.com/openshift/kubernetes-api.git
   version: origin-3.11-kubernetes-1.11.1
 - package: github.com/onsi/gomega
-  version: v1.3.0
+  version: v1.5.0
 - package: github.com/onsi/ginkgo
-  version: v1.4.0
+  version: v1.8.0
```
mohammedzee1000: +1

tests/e2e/helper/helper_oc.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	//. "github.com/onsi/gomega"
+)
+
+// CreateRandProject create new project with random name (10 letters)
+// without writing to the config file (without switching project)
+func OcCreateRandProject() string {
```
mohammedzee1000: +1

tests/e2e/e2e_test.go
```diff
@@ -44,11 +44,11 @@ func TestOdo(t *testing.T) {
 	RunSpecs(t, "odo test suite")
```
amitkrout: Remove the suite function in a separate file to isolate it form the actual test

For example: 

Create a file ```odo_suite_test.go```
```
package e2e

import (
	"testing"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
)

func TestOdo(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "odo test suite")
}
```

amitkrout:
> I am not seeing any helpers related to `waitForCmdOut` @kadel are we planning to add those or use the one in the utils?

@girishramnani I think its just a templates for our reference to write the test. Those functions are in utils can be moved under helper package in the subsequent improvement of this pr template.

amitkrout:
Test went fine locally and the interesting part is that with four node parallel execution the total run time reduces to half compare to serial execution of the same template.

Test structure looks good to me.

tests/e2e/cleantest_test.go
kadel: good point. 
It looks like we could use `SynchronizedBeforeSuite` and `SynchronizedAfterSuite`

https://onsi.github.io/ginkgo/#managing-singleton-external-processes-in-parallel-test-suites

tests/e2e/e2e_test.go
kadel: Agree, but I think that this goes a little bit beyond creating an example of how to write tests.
This should be part of the cleaning up effort .

kadel:
/retest
aws error

tests/e2e/e2e_test.go
kadel: This PR is intended to provide just a clean direction and template how we can structure our tests.
This should be part of a cleanup effort in a separate PR.

tests/e2e/cleantest_test.go
kadel: Fixed by removing those function all together, as I realized that there is a better way to do this

tests/e2e/cleantest_test.go
```diff
@@ -0,0 +1,145 @@
...
+	rand.Seed(time.Now().UTC().UnixNano())
+
+	// this needs to bedeclared separately
+	// if := is used in the next line it won't save originalDir to global variable
+	var err error
+	originalDir, err = os.Getwd()
+	Expect(err).NotTo(HaveOccurred())
+
+	originalProject = helper.OcCurrentProject()
+})
```
cdrage: If these will be ran for each and every test, would it make sense to try to put as much information as possible into the utility functions / helper files?

amitkrout:
@kadel Rebase and fix the formatting issue 
```
$ make validate
./scripts/check-gofmt.sh
./tests/e2e/template_cleantest_test.go
make: *** [gofmt] Error 1
```

kadel:
ping @amitkrout @girishramnani

PR #1540, opened 2019-04-01
Fixes the issue of storage unmounting during odo push

### LINKED ISSUE #1539: Storage unmounting during odo push
```
odo create nodejs
odo push 
odo storage create <name> --path <path> --size <size>
odo storage list
odo push
odo storage list
```

The storage gets unmounted during the push.

mik-dass:
fixes #1539

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -1267,7 +1267,7 @@ func (c *Client) UpdateBuildConfig(buildConfigName string, gitURL string, annota
 }
 
 // Define a function that is meant to update a DC in place
-type dcStructUpdater func(dc *appsv1.DeploymentConfig, existingDC *appsv1.DeploymentConfig, existingCmpContainer corev1.Container) error
+type dcStructUpdater func(dc *appsv1.DeploymentConfig) error
 type dcRollOutWait func(*appsv1.DeploymentConfig, int64) bool
 
 // PatchCurrentDC "patches" the current DeploymentConfig with a new one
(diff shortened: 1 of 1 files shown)


pkg/occlient/occlient.go
```diff
@@ -1287,6 +1287,9 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatc
 		}
 	}
 
+	// copy the any remaining volumes and volume mounts
+	copyVolumesAndVolumeMounts(dc, currentDC, existingCmpContainer)
```
kadel: Why is this called again? I is already inside the `prePatchDCHandler` that is called a few lines before this.

mik-dass: `prePatchDCHandler()` above is only called during local to git updates

kadel: Wouldn't it be better to remove it from prePatchHandler and call just call it here?

mik-dass:
@kadel Done

kadel:
ping @anmolbabu @girishramnani @surajnarwade

cdrage:
@mik-dass the code looks good, however, are you able to add a test (even e2e?)

amitkrout:
> > Copied from issue description
> > ```
> > odo create nodejs
> > odo push 
> > odo storage create <name> --path <path> --size <size>
> > odo storage list
> > odo push
> > odo storage list
> > ```
> > @mik-dass Such scenario is commented in our existing test due to new odo workflow . Can you please add a test scenario for this.
> 
> I don't think that this issue is supposed to fix this scenario. There is #1547

If this is the case then i will be happy to see the scenario coverage in #1547

mik-dass:
> @mik-dass the code looks good, however, are you able to add a test (even e2e?)

I will add the e2e test as part of #1547 :)

Issue #1493, opened 2019-03-20
push triggers two deployments


kadel:
```
mp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ odo create java
 ✓  Checking component
 ✓  Checking component version
Please use `odo push` command to create the component with source deployed

tmp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: java
  SourceLocation: /Users/tkral/tmp/odo-examples/fuse-rest-http-booster
  SourceType: local
  Application: fuse-rest-ht-gslp
  Project: myproject
  Name: java-sxmf

tmp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ odo push
 ✓  Checking component
 ✓  Checking component version
 ✓  Creating java component with name java-sxmf
 ✓  Initializing 'java-sxmf' component
 ✓  Creating component java-sxmf
 ✓  Successfully created component java-sxmf
 ✓  Pushing changes to component: java-sxmf of type local
 ✓  Waiting for component to start
 ✓  Copying files to component
 ✓  Building component
 ✓  Changes successfully pushed to component: java-sxmf


tmp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ oc get deploymentconfig
NAME                          REVISION   DESIRED   CURRENT   TRIGGERED BY
java-sxmf-fuse-rest-ht-gslp   1          1         1         config,image(java:latest)

tmp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ odo push
 ✓  Checking component
 ✓  Checking component version
 ✓  Applying component settings to component: java-sxmf
 ✓  The component java-sxmf was updated successfully
 ✓  Successfully created component with name: java-sxmf
 ✓  Pushing changes to component: java-sxmf of type local
 ✓  Waiting for component to start
 ✓  Copying files to component
 ◐  Building component
 ◑  Building component
 ◒  Building componentoc get deploymentconfig
 ✓  Building component
 ✓  Changes successfully pushed to component: java-sxmf


tmp/odo-examples/fuse-rest-http-booster  07d1c28 ✗                                                    67d ◒
▶ oc get deploymentconfig
[...]

PR #1541, opened 2019-04-01
Fix issue of 2 deployments on one odo push

cdrage:
Fixes the issues of two DeploymentConfig revisions happening on one `odo
push`.

OpenShift will automatically propagate the actual image using the
TriggerChange and thus having it blank is viable / okay.

Closes https://github.com/openshift/odo/issues/1493

note from @kadel:
This is why it works: https://github.com/openshift/origin/blob/f5f540d6503b3978a242117fb94b229135201df3/pkg/apps/controller/deploymentconfig/deploymentconfig_controller.go#L146

### DIFF
--- a/pkg/occlient/templates.go
+++ b/pkg/occlient/templates.go
@@ -105,7 +105,7 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
 					// The application container
 					Containers: []corev1.Container{
 						{
-							Image: fmt.Sprintf("%s/%s:%s", commonImageMeta.Namespace, commonImageMeta.Name, commonImageMeta.Tag),
+							Image: " ",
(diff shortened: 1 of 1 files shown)


cdrage:
Special thanks to @kadel 

I'm not joking.. but this *actually* fixes it... Ridiculous, isn't it? 

I did however encounter a bug while diagnosing this which I've opened here: https://github.com/openshift/odo/issues/1542

This will go into this PR (I've marked this PR as work-in-progress as I'll fix the bug here too)

mohammedzee1000:
Interesting

kadel:
> LGTM, works for me

@surajnarwade  please use `/lgtm`

kadel:
> LGTM

@mik-dass 
please use `/lgtm`

PR #1551, opened 2019-04-02
Converting Readme from markdown to adoc

Preeticp:
This PR converts the Readme from a markdown to an Asciidoc file


Preeticp:
@cdrage please take a look at this PR. I have converted the readme file from .md to .adoc.
I have added a TOC to this since the document is long. However the OpenShift icon does not render if I get the TOC to work. On the other hand if I get the Image to work the TOC rendering goes for a toss. If it's an either/or I would prefer to keep the TOC for easy navigation through the doc. Do you have any suggestions wrt the icon?

cdrage:
@Preeticp My only suggestions with the icon is perhaps having it at the very top of the document? It'd be nice to have the OpenShift logo there.

Preeticp:
@cdrage PTAL, I have added the OpenShift logo and have the TOC rendered too. I had to reduce the logo size. If I used the previous size, the 'OpenShift' name below the logo almost seemed to be the same size as the OpenShift in the heading, which made it seem as if we had written OpenShift twice. Also I see too much white space above the title.

See below for what I describe above:
Screenshot from 2019-04-03 11-14-30

I have tried multiple permutations and combinations for this, and this seems the best for now. Let me know your thoughts.
This is how it renders now: 
Screenshot from 2019-04-03 11-30-09

amitkrout:
@Preeticp Are you also planning to apply the same adoc format to other .md file ? I am asking because imo doc format should be in sync in the whole odo repo

Preeticp:
Yes @amit absolutely, I will be converting all the .md files to adoc.

Meanwhile, on another note, @cdrage, team, I had expected changes in the Getting Started as per my merged PR, to be reflected on the website here: https://openshiftdo.org/getting-started/ but they don't.

Can you please let me know if there is any additional step required for the website to render updated docs.

cdrage:
@Preeticp thanks. Looks like the bot has no access.. I'll have to fix it: https://travis-ci.org/openshift/odo/jobs/514686848

README.adoc
```diff
@@ -0,0 +1,203 @@
+[id="readme"]
+= odo - Developer-focused CLI for OpenShift image:/docs/img/openshift.png[float="right",align="center"]
```
cdrage: So I found a fix, you can just add raw HTML to here in order to render this correctly..

See my branch here: https://github.com/cdrage/odo/tree/preview

Essentially, remove what you have now for image and at the top of the document add:

```
+++
<img align="right" role="right" width="180" src="/docs/img/openshift.png?raw=true"/>
+++
```

Preeticp: I will try that @cdrage but I find the TOC not rendering in your branch too. Throughout my attempts, I have been able to either have the Image or the TOC, finally I reached what I see on this PR.
The TOC syntax is peculiar in that it needs to be right below the top most lines. As in there should be no line breaks between the content on the first line and the TOC syntax. Only then it renders correctly.

Preeticp: Tried, image looks perfect, though slightly big. But TOC is gone.
See here for what I described above: http://discuss.asciidoctor.org/toc-not-working-td5427.html#a5428

README.adoc
```diff
@@ -0,0 +1,205 @@
++++
+<img align="right" role="right" width="180" src="/docs/img/openshift.png?raw=true"/>
```
cdrage: Hey @Preeticp I've looked at your doc and I believe you did too big of a compression to the image... It's really really blurry? Perhaps keep the original file and just re-size. 

No idea what happened, but looks like modified the image and now it's super pixelated.

cdrage: See the file here: https://github.com/openshift/odo/blob/934b5412fc8a5f6aec1cea89c166fc3a91c0b0f7/README.adoc

Preeticp:
You are right, replaced with original image.
But I still have 2 concerns, we do not have the TOC rendered and the image seems too big, as in it is reaching way down to the overview itself. 
If you are fine with the above 2 issues then we can go ahead and merge this in. We can always fine tune later. For now this seems alright.

cdrage:
@Preeticp I don't think the TOC is very necessary on such a small README.md doc, we could work on fixing the issue later.

Please add a commit to delete the .md file and then we can merge this in :+1: thanks so much for being so patient.

Preeticp:
@cdrage  @kadel Closing this PR as I am unable to resolve conflict in a branch which has deleted the file I am supposed to resolve the conflict in. I have raised a new PR with all the changes at #1566

PR #1559, opened 2019-04-04
e2e test: core scenario for beta

kadel:
The new clean e2e scenario, that tries to simulate basic flow.
This can be a replacement for the "old" template.

This will be required to pass before we release beta version.

Watch test is missing, I'll add that in another PR.

### DIFF
--- a/tests/e2escenarios/core_beta_flow.go
+++ b/tests/e2escenarios/core_beta_flow.go
@@ -1,0 +1,169 @@
+package e2escenarios
+
+import (
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+
+	"github.com/openshift/odo/tests/e2escenarios/helper"
+)
+
+var _ = Describe("Core beta flow", func() {
+	//new clean project and context for each test
+	var project string
+	var context string
+
+	//  current directory and project (before eny test is run) so it can restored  after all testing is done
+	var originalDir string
+	var originalProject string
+
+	var oc helper.OcRunner
+	// path to odo binary
+	var odo string
+
+	BeforeEach(func() {
+		// Set default timeout for Eventually assertions
+		// commands like odo push, might take a long time
+		SetDefaultEventuallyTimeout(10 * time.Minute)
+
+		// initialize oc runner
+		// right now it uses oc binary, but we should convert it to client-go
+		oc = helper.NewOcRunner("oc")
+		odo = "odo"
+
+		project = oc.CreateRandProject()
+		context = helper.CreateNewContext()
+	})
+
+	AfterEach(func() {
+		oc.DeleteProject(project)
+		helper.DeleteDir(context)
+	})
+
+	Context("when project from KUBECONFIG is used", func() {
+		// Set active project for each test spec
+		JustBeforeEach(func() {
+			originalProject = oc.GetCurrentProject()
+			// WARNING: this project switching makes it impossible to run this in parallel
+			// it should set different KUBECONFIG before witching project
+			oc.SwitchProject(project)
+		})
+		// go back to original project after each test
+		JustAfterEach(func() {
+			oc.SwitchProject(originalProject)
+		})
+
+		// abstract main test to the function, to allow running the same test in a different context (slightly different arguments)
+		TestBasicCreateConfigPush := func(extraArgs ...string) {
+			createSession := helper.CmdShouldPass(odo, append([]string{"component", "create", "java", "mycomponent", "--app", "myapp"}, extraArgs...)...)
+			// output of the commands should point user to running "odo push"
+			Eventually(createSession).Should(gbytes.Say("odo push"))
+			configFile := filepath.Join(context, ".odo", "config.yaml")
+			Expect(configFile).To(BeARegularFile())
+			helper.FileShouldContainSubstring(configFile, "Name: mycomponent")
+			helper.FileShouldContainSubstring(configFile, "Type: java")
+			helper.FileShouldContainSubstring(configFile, "Application: myapp")
+			helper.FileShouldContainSubstring(configFile, "SourceType: local")
+			// SourcePath should be relative
+			//helper.FileShouldContainSubstring(configFile, "SourceLocation: .")
+			helper.FileShouldContainSubstring(configFile, "Project: "+project)
+
+			configSession := helper.CmdShouldPass(odo, append([]string{"config", "set", "--env", "FOO=bar"}, extraArgs...)...)
+			// output of the commands should point user to running "odo push"
+			// currently failing
+			Eventually(configSession).Should(gbytes.Say("odo push"))
+			helper.FileShouldContainSubstring(configFile, "Name: FOO")
+			helper.FileShouldContainSubstring(configFile, "Value: bar")
+
+			urlCreateSession := helper.CmdShouldPass(odo, append([]string{"url", "create", "--port", "8080"}, extraArgs...)...)
+			// output of the commands should point user to running "odo push"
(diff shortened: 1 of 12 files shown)


cdrage:
`template_cleantest_test.go` was deleted, is this normal?

I saw you create the new "flow", should we have another template? 

I'm wondering this since I'm currently implementing an e2e test for PR https://github.com/openshift/odo/pull/1446

kadel:
> `template_cleantest_test.go` was deleted, is this normal?

yes, to be honest, there was some mistakes in that template that could lead people to a wrong path :-(
I think that having a concrete  useful working test that can be used as a example is better than "half backed" template that doesn't  test anything properly

> 
> I saw you create the new "flow", should we have another template?

why another template?

> 
> I'm wondering this since I'm currently implementing an e2e test for PR #1446

This should be independent of any PRs. It is just generic testing scenario.

kadel:
/hold cancel
ping @amitkrout  @mohammedzee1000

amitkrout:
I will take a look and verify

surajnarwade:
nodejs flake, restarted the travis

kadel:
ping @surajnarwade @amitkrout all test green

kadel:
> It’s a good idea to keep e2e scenario test file in a separate dir, however I can see more improvement can be done to avoid duplicate helper/source file in tests dir. This improvement can be done later after beta release in a proper way.

My idea was to start with separating scenarios (real e2e test) and our current e2e tests (more component level/integration tests)

I thought that in future we would rename `tests/e2e` to `tests/integration` and `tests/e2escenarios`  to `tests/e2e`.  But it is something for longer discussion it is clear that we need a better way to organize tests, this is just one of the options before we figure out the right way

amitkrout:
@girishramnani @mohammedzee1000 Can please add ```lgtm``` if looks good to you  review 

> > It’s a good idea to keep e2e scenario test file in a separate dir, however I can see more improvement can be done to avoid duplicate helper/source file in tests dir. This improvement can be done later after beta release in a proper way.
> 
> My idea was to start with separating scenarios (real e2e test) and our current e2e tests (more component level/integration tests)
> 
> I thought that in future we would rename `tests/e2e` to `tests/integration` and `tests/e2escenarios` to `tests/e2e`. But it is something for longer discussion it is clear that we need a better way to organize tests, this is just one of the options before we figure out the right way

Right, I am also thinking the same way. For example we can follow something like 
```
Amits-MacBook-Pro:tests amit$ tree -L 1
.
├── OWNERS
├── e2e-scenario
├── helper
├── integration
├── test-src
└── test-template

5 directories, 1 file
Amits-MacBook-Pro:tests amit$
```

amitkrout:
> @girishramnani @mohammedzee1000 Can please add `lgtm` if looks good to you review
> 
> > > It’s a good idea to keep e2e scenario test file in a separate dir, however I can see more improvement can be done to avoid duplicate helper/source file in tests dir. This improvement can be done later after beta release in a proper way.
> > 
> > 
> > My idea was to start with separating scenarios (real e2e test) and our current e2e tests (more component level/integration tests)
> > I thought that in future we would rename `tests/e2e` to `tests/integration` and `tests/e2escenarios` to `tests/e2e`. But it is something for longer discussion it is clear that we need a better way to organize tests, this is just one of the options before we figure out the right way
> 
> Right, I am also thinking the same way. For example we can follow something like
> 
> ```
> Amits-MacBook-Pro:tests amit$ tree -L 1
> .
> ├── OWNERS
> ├── e2e-scenario
> ├── helper
> ├── integration
> ├── test-src
> └── test-template
> 
> 5 directories, 1 file
> Amits-MacBook-Pro:tests amit$
> ```

Created #1581

kadel:
> @kadel Unfortunately i see failures locally :(
> 
> Seems `make test-e2e-scenarios` is not a part of travis and prow CI run. Can you please add the test to both the CI

yes, it the test will fail as it is testing stuff that were not fixed yet.

Can't add it to CI until this is merged.

amitkrout:
> > @kadel Unfortunately i see failures locally :(
> > Seems `make test-e2e-scenarios` is not a part of travis and prow CI run. Can you please add the test to both the CI
> 
> yes, it the test will fail as it is testing stuff that were not fixed yet.
> 
> Can't add it to CI until this is merged.

Ahh...Thank for the info. 
I would say let wait for all the pr for beta is merged first then you can enable this in CI.

tests/e2escenarios/core_beta_flow.go
```diff
@@ -0,0 +1,169 @@
...
+			JustAfterEach(func() {
+				oc.SwitchProject(originalProject)
+			})
+
+			It("odo component should fail if there already is .odo dir", func() {
+				helper.CmdShouldPass("odo", "component", "create", "nodejs", "--context", context)
+				helper.CmdShouldFail("odo", "component", "create", "nodejs", "--context", context)
+			})
+
+			It("odo config should fail if there is no .odo dir", func() {
```
surajnarwade: this test should be put before previous test because previous test will create `.odo` dir

amitkrout:
> > I would say let wait for all the pr for beta is merged first then you can enable this in CI.
> 
> Why? How are we going to know when this is passing if it is not run?
> Would it be a problem if this is running before everything is in?
> Also merging this is not going to affect anything as the new job needs to be added in openshift/release

My assumption is based on local test failure and the failure reason is as you explained above https://github.com/openshift/odo/pull/1559#issuecomment-481212494. So I thought if the e2e for beta is failing then there is no reason to merge it to master.

From the above set of question I can assume your thought behind this 
```
1.  Let e2e test for beta ready in the master
2. Once the all the required pr for beta is in then just enable the e2e test job 
3. Then e2e test will test and verify the beta bits for release
```

Hope my guess is right ;)

Considering your assumption I am approving the pr
/approve

tests/e2escenarios/core_beta_flow.go
```diff
@@ -0,0 +1,169 @@
...
+			JustAfterEach(func() {
+				helper.Chdir(originalDir)
+			})
+
+			It("'odo component' should fail if there already is .odo dir", func() {
+				helper.CmdShouldPass("odo", "component", "create", "nodejs")
+				helper.CmdShouldFail("odo", "component", "create", "nodejs")
+			})
+
+			It("'odo config' should fail if there is no .odo dir", func() {
```
surajnarwade: this should be put before previous test

surajnarwade: ideally command will run successfully because previous test will create `.odo`

girishramnani:
\lgtm

surajnarwade:
@kadel , ignore the review, I was bit confused but CI seems to be :green_heart:

kadel:
> > > I would say let wait for all the pr for beta is merged first then you can enable this in CI.
> > 
> > 
> > Why? How are we going to know when this is passing if it is not run?
> > Would it be a problem if this is running before everything is in?
> > Also merging this is not going to affect anything as the new job needs to be added in openshift/release
> 
> My assumption is based on local test failure and the failure reason is as you explained above [#1559 (comment)](https://github.com/openshift/odo/pull/1559#issuecomment-481212494). So I thought if the e2e for beta is failing then there is no reason to merge it to master.
> 
> From the above set of question I can assume your thought behind this
> 
> ```
> 1.  Let e2e test for beta ready in the master
> 2. Once the all the required pr for beta is in then just enable the e2e test job 
> 3. Then e2e test will test and verify the beta bits for release
> ```
> Hope my guess is right ;)
> 
> Considering your assumption I am approving the pr
> /approve

yes, sorry if it was not clear. I thought that we discussed this at the blj call that we have when I opened this.

tests/e2escenarios/core_beta_flow.go
kadel: Each `It` is an independent test case. There can't be any dependencies on the order in which the `It` sections are run.

PR #1568, opened 2019-04-05
Development guide review and adoc conversion

Preeticp:
This PR reviews, modifies and converts the Development guide from .md to .adoc

### DIFF
--- a/docs/development.adoc
+++ b/docs/development.adoc
@@ -3,7 +3,7 @@
 :toc-title:
 :toclevels: 1
 
-  toc::[]
+toc::[]
 
 == Submitting a pull request(PR)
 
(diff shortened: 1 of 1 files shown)


Preeticp:
@cdrage @kadel please review and merge if you find it alright.

Preeticp:
As per discussions with Robert (rkratky), to keep it consistent with documentation standards, `odo` willl be referenced as OpenShift Do in the headings and titles for all docs. In the current case retaining odo-bot as is as the bot is called odo-bot.

cdrage:
@Preeticp Any way to not get the "note" sections in tables? I think instead it should be bolded like so:


**Note:** blah blah blah

Preeticp:
@cdrage the syntax to be used as per doc conventions and Asciidoc syntax is 'NOTE:' 
As for rendering we should try to render it as the OC docs do it, so that it is consistent. I will discuss it with @rkratky and see if we can tweak the css to render it without tables. For now the syntax used is correct and I would suggest we let it be as is.

PR #1577, opened 2019-04-08
openshiftci-presubmit-unittests: get gox before attemting make cross

kadel:
- get gox before running `make cross`
- remove arm from build targets (no one is using it right now, and it is just slowing down builds), it can be added back if needed


amitkrout:
Minor changes. So
/approve
/lgtm

PR #1583, opened 2019-04-09
re-arranged the contexts in template test

girishramnani:
re-arranged the `Context` in the clean template test so please can refer it.

There should be two high level contexts
- when component is in the current directory
- when --context is used

No issue

### DIFF
--- a/tests/e2e/template_cleantest_test.go
+++ b/tests/e2e/template_cleantest_test.go
@@ -87,18 +87,6 @@ var _ = Describe("Example of a clean test", func() {
 
 	var _ = Context("when --context is used", func() {
 
-		// we will be testing components that are created from the current directory
-		// switch to the clean context dir before each test
-		var _ = JustBeforeEach(func() {
-			originalDir = helper.Getwd()
-			helper.Chdir(context)
-		})
-
-		// go back to original directory after each test
-		var _ = JustAfterEach(func() {
-			helper.Chdir(originalDir)
-		})
-
 		var _ = Context("when project from KUBECONFIG is used", func() {
 			// Set active project for each test spec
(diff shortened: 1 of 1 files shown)


tests/e2e/template_cleantest_test.go
```diff
@@ -83,39 +83,52 @@ var _ = Describe("Example of a clean test", func() {
...
-					helper.CmdShouldPass("odo push")
-					//TODO: verify resources on cluster
-				})
+	})
+
+	var _ = Context("when --context is used", func() {
+
+		// we will be testing components that are created from the current directory
+		// switch to the clean context dir before each test
+		var _ = JustBeforeEach(func() {
```
kadel: This shouldn't be here. If you are calling commands with `--context` flag you shouldn't to the directory, otherwise, there is no point of specifying `--context`

girishramnani: ack, I felt so too.

kadel:
To be honest, the template has a lot of problems. 
I would rather use what I did in https://github.com/openshift/odo/pull/1559 as a template

tests/e2e/template_cleantest_test.go
girishramnani: done

girishramnani:
@kadel while that scenario gets in, I feel we should enhance the template too

girishramnani:
After this current change, this current template is quite understandable in my opinion @kadel

kadel:
> After this current change, this current template is quite understandable in my opinion @kadel

It has a lot of other problem like us of Linux specific programs (for example `/bin/sh`) :-(

girishramnani:
Closing this PR then @kadel

PR #1600, opened 2019-04-10
inform user to run odo push after odo config

kadel:
this also fixes a few minor issues in e2e beta scenario
The test is passing now. After this is merged I'll open PR on openshift/release to run this tests

### DIFF
--- a/tests/e2escenarios/core_beta_flow.go
+++ b/tests/e2escenarios/core_beta_flow.go
@@ -6,7 +6,6 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/gbytes"
 
 	"github.com/openshift/odo/tests/e2escenarios/helper"
 )
@@ -60,7 +59,7 @@ var _ = Describe("Core beta flow", func() {
 		TestBasicCreateConfigPush := func(extraArgs ...string) {
 			createSession := helper.CmdShouldPass(odo, append([]string{"component", "create", "java", "mycomponent", "--app", "myapp"}, extraArgs...)...)
 			// output of the commands should point user to running "odo push"
-			Eventually(createSession).Should(gbytes.Say("odo push"))
+			Expect(createSession).Should(ContainSubstring("odo push"))
 			configFile := filepath.Join(context, ".odo", "config.yaml")
 			Expect(configFile).To(BeARegularFile())
 			helper.FileShouldContainSubstring(configFile, "Name: mycomponent")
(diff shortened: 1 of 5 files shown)


kadel:
/retest 
```
level=error msg="\t* aws_lb.api_internal: timeout while waiting for state to become 'active' (timeout: 10m0s)"
```

surajnarwade:
@kadel , please rebase so that travis will be happy :green_heart:  :smile_cat:

tests/e2escenarios/core_beta_flow.go
```diff
@@ -60,7 +59,7 @@ var _ = Describe("Core beta flow", func() {
 		TestBasicCreateConfigPush := func(extraArgs ...string) {
 			createSession := helper.CmdShouldPass(odo, append([]string{"component", "create", "java", "mycomponent", "--app", "myapp"}, extraArgs...)...)
 			// output of the commands should point user to running "odo push"
-			Eventually(createSession).Should(gbytes.Say("odo push"))
+			Expect(createSession).Should(ContainSubstring("odo push"))
```
amitkrout: +1, i prefer ```Expect``` and ```ContainSubstring``` over ```Eventually``` and ```gbytes.Say```

amitkrout:
> this also fixes a few minor issues in e2e beta scenario
The test is passing now. After this is merged I'll open PR on openshift/release to run this tests

Cool!!! Then why you are not adding it in travis CI in this pr.

tests/e2escenarios/core_beta_flow.go
kadel: > +1, i prefer `Expect` and `ContainSubstring` over `Eventually` and `gbytes.Say`

Yes, this is better here, but both have different use ;-) 
Of you would test larger stream of output, it would be better to use `Eventually` with `Say` as it would exist with the first occurrence and you wouldn't need to wait for the whole output. Here we have to get the whole output as we need multiple assertions on one output .

kadel:
> Cool!!! Then why you are not adding it in travis CI in this pr.

not sure. Should we run this also on Travis? It will add to more time to already overloaded TravisCI tests

amitkrout:
> > Cool!!! Then why you are not adding it in travis CI in this pr.
> 
> not sure. Should we run this also on Travis? It will add to more time to already overloaded TravisCI tests

 It is understandable as part of test load sharing is concern between ```TravisCI``` and ```OpenShiftCI```. I have no issue with sharing test between CIs.

tests/e2escenarios/core_beta_flow.go
amitkrout: > > +1, i prefer `Expect` and `ContainSubstring` over `Eventually` and `gbytes.Say`
> 
> Yes, this is better here, but both have different use ;-)
> Of you would test larger stream of output, it would be better to use `Eventually` with `Say` as it would exist with the first occurrence and you wouldn't need to wait for the whole output. Here we have to get the whole output as we need multiple assertions on one output .

Thanks, really helpful explanation

kadel:
/retest

```
 stderr: ✗  Failed to create component with name ref-test-1555314459. Please use `odo config view` to view settings used to create component. Error: image stream ruby with tag latest not found in openshift and project-1555314459 namespaces
```

PR #1608, opened 2019-04-11
Logic fix for OpenShift depoymentconfig triggers

cdrage:
This changes the logic to see if the latest image is greather than the
desired image due to rare cases of OpenShift updating the
deploymentconfig quickly in succession when using an ImageTrigger.


kadel:
Looks good. 
We don't need any special test for this. Verification will be if it fixes our current CI failures ;-)

/aprove

Issue #1576, opened 2019-04-08
odo push doesn't remove configuration from DC


kadel:
```
▶ odo create java --port 8080 --port 8088
▶ odo config set memory 2Gi
Local config was successfully updated.

▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: java
  SourceLocation: /Users/tkral/tmp/odo-examples/spring-boot-http-booster
  SourceType: local
  Ports:
  - "8080"
  - "8088"
  Application: spring-boot-zhrc
  Project: hyycxtlhgj
  Name: java-wygx
  MinMemory: 2Gi
  MaxMemory: 2Gi


▶ odo push
 ✓  Checking component
 ✓  Checking component version
 ✓  Creating java component with name java-wygx
 ✓  Initializing 'java-wygx' component
 ✓  Creating component java-wygx
 ✓  Successfully created component java-wygx
 ✓  Pushing changes to component: java-wygx of type local
 ✓  Waiting for component to start
 ✓  Copying files to component
 ✓  Building component
 ✓  Changes successfully pushed to component: java-wygx



▶ oc get pod java-wygx-spring-boot-zhrc-1-snb96 -o yaml
apiVersion: v1
kind: Pod
....
...
    
    resources:
      limits:
        memory: 2Gi
      requests:
        memory: 2Gi
 ....
....
```

This setting config is fine.
But problem is when I want to remove something.

```
▶ odo config unset memory
? Do you want to unset memory in the config Yes
Local config was successfully updated.

▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: java
  SourceLocation: /Users/tkral/tmp/odo-examples/spring-boot-http-booster
  SourceType: local
  Ports:
  - "8080"
  - "8088"
  Application: spring-boot-zhrc
  Project: hyycxtlhgj
  Name: java-wygx

▶ odo push
 ✓  Checking component
 ✓  Checking component version
 ✓  Applying component settings to component: java-wygx
 ✓  The component java-wygx was updated successfully
 ✓  Successfully created component with name: java-wygx
 ✓  Pushing changes to component: java-wygx of type local
 ✓  Waiting for component to start
 ✓  Copying files to component
 ✓  Building component
 ✓  Changes successfully pushed to component: java-wygx


[...]

PR #1614, opened 2019-04-15
resolved resource limit not reflecting on unsetting

girishramnani:
resolved resource limit not reflecting on unsetting.

fixes https://github.com/openshift/odo/issues/1576

## How to test changes?

```
odo create java --port 8080 --port 8088
odo config set memory 2Gi
odo push
oc get pod <pod-name> -o yaml # should have memory
odo config unset memory
oc get pod <pod-name> -o yaml # should not have memory
```


kadel:
/hold
no tests :-(

kadel:
@girishramnani how are tests going? We just need one simple test that verifies that the value is unset from the DC. It doesn' have to be the whole e2e scenario just one simple integration test case

kadel:
👍 

/lgtm

amitkrout:
Test looks awesome to me. 👍

Issue #1503, opened 2019-03-21
odo project delete fails after first delete


anmolbabu:
[kind/bug]

## What versions of software are you using?
- Operating System: Fedora 29
- Output of `odo version`: master with #1360 

## How did you run odo exactly?
odo project create odo2-1553171170
odo project create prj4

odo project delete prj4
odo project delete odo2-1553171170

## Actual behavior
[anmolbabu@localhost odo]$ odo project delete prj4
? Are you sure you want to delete project prj4 Yes
 ✓  Deleting project prj4
Deleted project : prj4

[anmolbabu@localhost odo]$ odo project delete odo2-1553171045
 ✗  deploymentconfigs.apps.openshift.io is forbidden: User "developer" cannot list deploymentconfigs.apps.openshift.io in the namespace "prj4": no RBAC policy matched

## Expected behavior

Both deletes should go through

girishramnani:
picking this up @kadel

syamgk:
Note: after the deletion of project it's not switching the active project
and while doing the second project delete it's trying to query the same project instead of the name passed via the cli cmd

girishramnani:
If I remember correctly the switching project part was already removed with
anmol's PR

PR #1618, opened 2019-04-15
fix 1503: odo project delete 2nd iteration

syamgk:

After the deletion of project it's not switching the active project,
this patch will modify the active project in runtime after confirming the existence of given project name.

#1503

## How to test changes?

Validate if the scenario mentioned in this issue description works properly
https://github.com/openshift/odo/issues/1503

### DIFF
--- a/pkg/odo/cli/project/delete.go
+++ b/pkg/odo/cli/project/delete.go
@@ -61,6 +61,8 @@ func (pdo *ProjectDeleteOptions) Validate() (err error) {
 
 // Run runs the project delete command
 func (pdo *ProjectDeleteOptions) Run() (err error) {
+
+	pdo.Context.Client.Namespace = pdo.projectName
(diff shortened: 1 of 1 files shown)


pkg/odo/cli/project/delete.go
```diff
@@ -61,6 +61,8 @@ func (pdo *ProjectDeleteOptions) Validate() (err error) {
 
 // Run runs the project delete command
 func (pdo *ProjectDeleteOptions) Run() (err error) {
+
+	pdo.Context.Client.Namespace = pdo.projectName
```
girishramnani: should this happen as part of context initialization?

syamgk: no, its not required during context initialization

girishramnani: ack

girishramnani:
closing this PR as we need a different approach for this fix

PR #1622, opened 2019-04-15
Improve project list

### LINKED ISSUE #1620: slow project.Exist is making whole odo slow
The whole odo is slow because of how we check if project exist.
Instead of using one call to the api server it list all the projects in the cluster.

And because of how `project.List` is implemented it also lists all DC in each project.

If `project.Exists("myproject")` is called  and user has 10 projects,this simple check will result in  10 API calls :-(

surajnarwade:

Improves `project.Exist` and `project.List`

fixes #1620

## How to test changes?

tests should be fast

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -1,9 +1,9 @@
 package project
 
 import (
+	"github.com/openshift/odo/pkg/application"
 	"github.com/pkg/errors"
 
-	"github.com/openshift/odo/pkg/application"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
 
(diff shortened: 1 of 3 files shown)


pkg/project/project.go
```diff
@@ -49,7 +49,7 @@ func Delete(client *occlient.Client, projectName string) error {
 	return nil
 }
 
-func List(client *occlient.Client) (ProjectList, error) {
+func DescribeProjects(client *occlient.Client) (ProjectList, error) {
 	currentProject := client.GetCurrentProjectName()
 	allProjects, err := client.GetProjectNames()
```
mohammedzee1000: You could just reuse List here

kadel: Is it the same?

The real question is do we need both `GetProjectNames` and `List`. `List` just wraps the error nothing else

surajnarwade: one and the same

surajnarwade: * `GetProjectNames` is in occlient.go though

surajnarwade: updated the code, please review again

kadel: > * `GetProjectNames` is in occlient.go though

ah, didn't noticed that they are in a different package 😇 , in that case, it might be ok to have both 😇 

But this is also ok ;-)

Issue #1623, opened 2019-04-15
odo component fails when anyuid is set


kadel:
### UPDATE 2019-04-17
see https://github.com/openshift/odo/issues/1623#issuecomment-484110785

### Original issue description
I'm unable to create component due to failure in one of the iniContainer (copy-files-to-volume)

~~The strange part is that this is not consistent it is happening only sometimes.~~
Update: Now I'm getting this every time on OpenShift 4.0 cluster

```
▶ odo push
 ✓  Checking component
 ✓  Checking component version
 ✓  Creating java component with name java-bayi
 ✓  Initializing 'java-bayi' component
 ✓  Creating component java-bayi
 ✓  Successfully created component java-bayi
 ✓  Applying component settings to component: java-bayi
 ✓  Checking URL java-bayi-8080
 ✓  Successfully created URL for component: java-bayi
 ✓  http://java-bayi-8080-spring-boot-kvcm-default.apps.tkral.devcluster.openshift.com
 ✓  The component java-bayi was updated successfully
 ✓  Successfully updated component with name: java-bayi
 ✓  Pushing changes to component: java-bayi of type binary
 ✗  Waiting for component to start
 ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=java-bayi-spring-boot-kvcm'


▶ oc get pods
NAME                                  READY   STATUS                  RESTARTS   AGE
java-bayi-spring-boot-kvcm-1-cz86r    0/1     Init:CrashLoopBackOff   4          2m34s
java-bayi-spring-boot-kvcm-1-deploy   1/1     Running                 0          2m44s


▶ oc logs java-bayi-spring-boot-kvcm-1-cz86r -c copy-files-to-volume
mkdir: cannot create directory '/mnt/app-root.setup-volume': Permission denied
tar: /opt/app-root: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
mv: cannot stat '/mnt/app-root.setup-volume': No such file or directory

```

I'm able to observe this only on OpenShift 4.0.
Haven't noticed this error last week :-(

kadel:
inside initContainer:
```
[jboss@java-bayi-spring-boot-kvcm-2-kmz7v ~]$ mount | grep mnt
/dev/xvdbx on /mnt type ext4 (rw,relatime,seclabel)

[...]

PR #1626, opened 2019-04-15
Adds security context / SYS_ADMIN privileges to bootstrap container

cdrage:
There are multiple issues in getting volume support / PVC to work inside
of an OpenShift container, such as: https://bugzilla.redhat.com/show_bug.cgi?id=1622670

This PR adds SYS_ADMIN privileges to the boostrap volume copy container, allowing the
container to mount the /tmp folder correctly and thus copy over the files.

During the transition from OpenShift 3.11 to 4.0, SELinux settings (look) to have changed,
I have yet to confirm this however..

Closes https://github.com/openshift/odo/issues/1623

### DIFF
--- a/pkg/occlient/templates.go
+++ b/pkg/occlient/templates.go
@@ -323,11 +323,15 @@ func generateBuildConfig(commonObjectMeta metav1.ObjectMeta, gitURL, gitRef, ima
 // dc is the deployment config to be updated
 // dcName is the name of the deployment config
 func addBootstrapVolumeCopyInitContainer(dc *appsv1.DeploymentConfig, dcName string) {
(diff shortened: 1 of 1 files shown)


cdrage:
ping @kadel

cdrage:
Let's see if CI passes :+1: this worked on my 3.11 cluster as well as the 4.0 cluster that @kadel gave me access to!

kadel:
/hold

I don't think that this is a good solution. I suspect that this decreases the security of the cluster, and this won't be possible to do on all clusters.

kadel:
I'm going to close this one, as this is not an acceptable solution. #1623 was updated with more inside to what is happening

PR #1629, opened 2019-04-16
Improving format of commands in the main guides.

Preeticp:
This PR modifies the format of commands in the main guides, Readme, Development, Getting started and Installation and examples.
It also has a change in command in the Readme on line 145 & 146 based on input from @surajnarwade 

I will raise another PR for the rest of the docs.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -114,7 +114,7 @@ $ minishift start
 . Log into the OpenShift cluster.
 +
 ----
-$ odo login --username developer --password developer
+$ odo login -u developer -p developer
 ----
 . Create an application. An application in `odo` is an umbrella
 under which you add other components.
--- a/README.adoc
+++ b/README.adoc
@@ -114,7 +114,7 @@ $ minishift start
 . Log into the OpenShift cluster.
 +
 ----
-$ odo login -u developer -p developer
+$ odo login --username developer --password developer
 ----
 . Create an application. An application in `odo` is an umbrella
 under which you add other components.
@@ -142,8 +142,8 @@ $ odo push
 . Create a URL to access the application and visit it to test it.
 +
 ----
-$ odo url create
-$ curl nodejs-myproject.192.168.42.147.nip.io
+$ odo url create --port <application port>
+$ curl <generated URL>
 ----
 
 For more in-depth information and advanced use-cases such as adding storage to a component or linking components, see the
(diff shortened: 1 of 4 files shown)


Preeticp:
@cdrage, @surajnarwade please review

README.adoc
```diff
@@ -114,7 +114,7 @@ $ minishift start
 . Log into the OpenShift cluster.
 +
 ----
-$ odo login -u developer -p developer
+$ odo login --username developer --password developer
```
cdrage: I don't like how we're using `--username` and `--password` here. Could we stick with `-u`? WDYT @kadel

README.adoc
```diff
@@ -142,8 +142,8 @@ $ odo push
 . Create a URL to access the application and visit it to test it.
 +
 ----
-$ odo url create
-$ curl nodejs-myproject.192.168.42.147.nip.io
+$ odo url create --port <application port>
```
cdrage: port is automatically generated when you do `odo url create` and specifying `--port` should only be used when there are multiple ports to be opened (which `odo` warns the user about...)

Preeticp: This is as per input from @surajnarwade Suraj can you please confirm on this?

README.adoc
Preeticp: Well, expanding options in the commands is just to enable first time users to understand quickly. If this is too obvious we can change it back. Let me know what you think.

surajnarwade: `-u` and `-p` is quite obivious, even noob will understand, so we should change it back
I agree with @cdrage

README.adoc
surajnarwade: @cdrage ,as of now, `--port` is mandatory, we will try to remove in future but as of now it is mandatory.
because when we do `odo url create`, we just make entry to localconfig, we dont talk to cluster, hence we can not detect the port.

kadel:
```
Warning: oc apply should be used on resource created by either oc create --save-config or oc apply
oauth.config.openshift.io/cluster configured
error: You must be logged in to the server (Unauthorized)
```
/retest

Preeticp:
@kadel  could you provide more context to the above comment? Is it to do with the changes made in this PR?

README.adoc
Preeticp: Reverted

kadel:
/test e2e-core

```
 Failed to login as developer
make[1]: *** [configure-installer-tests-cluster] Error 1
```

Preeticp:
@kadel I have rebased my PR post Suraj's PR introducing a conflict. I think this is good to go, can we merge it?

PR #1636, opened 2019-04-23
[DO NOT MERGE] Enhance openshift 4 configuration script

surajnarwade:
fixes #???

### DIFF
--- a/scripts/configure-installer-tests-cluster.sh
+++ b/scripts/configure-installer-tests-cluster.sh
@@ -85,7 +85,7 @@ EOF
 # Login as developer and check for stable server
 for i in {1..40}; do
     # Try logging in as developer
-    oc login -u developer -p $USERPASS &> /dev/null
+    oc login -u developer -p $USERPASS
     if [ $? -eq 0 ]; then
         # If login succeeds, assume success
 	    OC_STABLE_LOGIN="true"
@@ -115,7 +115,7 @@ fi
 oc whoami
 
 # Setup project
-=======
 oc new-project myproject
+oc status
 sleep 4
 oc version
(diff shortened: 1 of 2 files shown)


kadel:
/approve cancel 

It looks like `oc get secret htpasswd-secret -n openshift-config` is returning exit code `1`

surajnarwade:
@kadel , I removed `/dev/null` temporarily to see the error

scripts/configure-installer-tests-cluster.sh
```diff
@@ -92,6 +94,11 @@ if [ -z $OC_LOGIN_SUCCESS ]; then
     exit 1
 fi
 
+sleep 3
```
mohammedzee1000: +1 Guess it makes sense here

scripts/configure-installer-tests-cluster.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 # Setup to find nessasary data from cluster setup
 ## Constants
+set -x
```
mohammedzee1000: +1

scripts/configure-installer-tests-cluster.sh
```diff
@@ -79,7 +81,7 @@ EOF
 
 # Login as developer and setup project
 for i in {1..40}; do
-    oc login -u developer -p $USERPASS &> /dev/null
+    oc login -u developer -p $USERPASS
```
mohammedzee1000: Hmm @kadel do we want to reenable the output of these commands

mohammedzee1000: I guess i was reviewing in parallel so didint see comment by @surajnarwade

scripts/configure-installer-tests-cluster.sh
kadel: how this is going to help?

kadel:
bump @surajnarwade

PR #1637, opened 2019-04-23
Updated getting started guide

surajnarwade:
few updates about URL in getting started guide

### DIFF
--- a/docs/getting-started.adoc
+++ b/docs/getting-started.adoc
@@ -54,19 +54,19 @@ $ odo push
 ----
 +
 Your component is now deployed to OpenShift.
-. This creates the URL and adds an entry in the local configuration file:
+. Create a URL and add an entry in the local configuration file as follows:
 +
 ----
 $ odo url create --port 8080
 ----
 +
-. Push the changes which will create a URL on cluster:
+. Push the changes. This creates a URL on the cluster.
 +
 ----
 $ odo push
 ----
 +
-. List the URLs to check the desired URL for component.
+. List the URLs to check the desired URL for the component.
 +
 ----
 $ odo url list
--- a/docs/getting-started.adoc
+++ b/docs/getting-started.adoc
@@ -54,13 +54,29 @@ $ odo push
 ----
 +
 Your component is now deployed to OpenShift.
-. To access the component, create an OpenShift route and view your
-deployed application using the generated URL:
+. To access the component, create an URL:
 +
 ----
-$ odo url create --open
+$ odo url create --port 8080
 ----
-
++
+. Previous command will make entry in config file, to create URL on cluster:
++
+---
+$ odo push
+---
++
+. List the URL to check desired URL for component.
++
+---
+$ odo url list
+---
++
+. Curl the URL to check the view your deployed application using the generated URL.
+---
+$ curl <URL>
+---
++
 .  Edit your code and push the changes to the component:
 ..  Edit one of the layout files within the Node.js directory.
 +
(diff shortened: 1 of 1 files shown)


docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
...
+. To access the component, create an URL:
 +
 ----
-$ odo url create --open
+$ odo url create --port 8080
 ----
-
++
+. Previous command will make entry in config file, to create URL on cluster:
++
```
Preeticp: There seems to be some mix up here. Has an explanation for the previous step been mixed up with the next instruction step. In that case we will have to split it as mentioned here and in the next command.
Suggest: This also adds an entry in the configuration file. 
Please remove the period (it creates a numbered step) at the beginning of the sentence, as it is not an user action/next step. It simply explains what happens in the previous step.

docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
...
+---
+$ odo push
+---
++
+. List the URL to check desired URL for component.
++
```
Preeticp: the desired URL for the component
Does `$ odo url list` list out all the URLs, if yes then replace the first URL with URLs

docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
...
+---
+$ odo url list
+---
++
+. Curl the URL to check the view your deployed application using the generated URL.
```
Preeticp: Delete 'Curl the URL to check the' . 'View your deployed...' or Check your deployed...' should be enough.

docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
 ----
 +
 Your component is now deployed to OpenShift.
-. To access the component, create an OpenShift route and view your
-deployed application using the generated URL:
+. To access the component, create an URL:
```
Preeticp: a URL

docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
...
 ----
-$ odo url create --open
+$ odo url create --port 8080
 ----
-
++
+. Previous command will make entry in config file, to create URL on cluster:
++
+---
+$ odo push
```
Preeticp: What does Push do here? Does it create the URL on the cluster? If so the the numbered step should be either 'create the URL on the cluster' or maybe 'Push the URL to the cluster' since we already created the URL earlier.

docs/getting-started.adoc
```diff
@@ -54,13 +54,29 @@ $ odo push
...
 +
 ----
-$ odo url create --open
+$ odo url create --port 8080
 ----
-
++
+. Previous command will make entry in config file, to create URL on cluster:
++
+---
```
Preeticp: There should be four hyphens ---- here and on lines 67. 71, 73, 76, 78

docs/getting-started.adoc
```diff
@@ -54,13 +54,30 @@ $ odo push
 ----
 +
 Your component is now deployed to OpenShift.
-. To access the component, create an OpenShift route and view your
-deployed application using the generated URL:
+. This creates the URL and adds an entry in the local configuration file:
```
Preeticp: Create a URL and add an entry in the local configuration file as follows:

docs/getting-started.adoc
Preeticp: the component

Preeticp:
@surajnarwade lgtm from my side :)

kadel:
/retest
```
Failed to login as developer
```

PR #1644, opened 2019-04-23
Adds the odo website to the master branch

cdrage:
This adds the odo website to the master branch of odo.

From here on out, we'll modify any site changes (layout and such) in the
root directory, while the `gh-pages` branch will contain the built / raw
html.


cdrage:
This is a step in the right direction @rkratky

If someone could merge this ASAP that'd be great. After it's merged I can update the synchronization script.

site/pkg/build/.coverprofile
```diff
@@ -0,0 +1,89 @@
+mode: atomic
+github.com/kedgeproject/kedge/pkg/build/build.go:43.79,46.16 2 0
```
surajnarwade: kedge :question:

kadel: good catch.
This file shouldn't be there. It is unit test coverage

cdrage: wow that's an old file. thanks! removed.

kadel:
```
level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
```
/retest

PR #1681, opened 2019-05-02
Update labels to match app-label recommendation

### LINKED ISSUE #599: Adhere to new labels created for tool standarization in k8s
There's been an effort on the kubernetes community that has defined some labels for tooling to be able to work with applications.
We should investigate if these fit and if we can incorporate them:

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

kadel:
https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc

updates labels that odo uses to match guideline at https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc

This will allow seeing odo components in topology view in devconsole

fixes #599 

**WARNING:**
This PR makes incompatible changes. When merged components that are already deployed on the cluster with the older odo version will no longer be visible or display show incorrectly in odo output.


girishramnani:
@kadel do we want to consider a migration plan for the people using beta?

kadel:
/hold

some clarification is needed. See discussion in #599

kadel:
the current state of labels in this PR

| use | currently use label | new label based on [guidelines](https://github.com/gorkem/app-labels) |
|-|-|-|
| application name | app.kubernetes.io/name | app.kubernetes.io/part-of |
| component name | app.kubernetes.io/component-name |app.kubernetes.io/instance |
| component type (image builder e.g. java, nodejs) | app.kubernetes.io/component-type | app.kubernetes.io/name |
| component version (builder image version e.g. latest, 0.17) | app.kubernetes.io/component-version |  app.openshift.io/runtime-version |
| URI for the source code under version control (only if component is created from git repo) | app.kubernetes.io/url | app.openshift.io/vcs-uri |

kadel:
closing this PR for now, until we decide on the correct set of labels

Issue #1682, opened 2019-05-02
push triggers 4 deployments for git components


mik-dass:
[kind/bug]

`odo push` triggers 4 deployments and 2 builds for git components

## Actual behavior
4 deployments and 2 builds

## Expected behavior
1 deployment and 1 build should be triggered

## Any logs, error output, etc?

```
[mrinaldas@localhost /]$ oc get dc
NAME              REVISION   DESIRED   CURRENT   TRIGGERED BY
nodejs-dnuu-app   2          1         1         config,image(nodejs-dnuu-app:latest)
[mrinaldas@localhost /]$ oc get dc
NAME              REVISION   DESIRED   CURRENT   TRIGGERED BY
nodejs-dnuu-app   6          1         1         config,image(nodejs-dnuu-app:latest)
```

amitkrout:
@mik-dass i am closing thsi issue as it has been addressed in pr #1684.

mik-dass:
https://github.com/openshift/odo/pull/1684 doesn't resolve this issue

/reopen

PR #1684, opened 2019-05-02
Deploy fix

### LINKED ISSUE #1633: prevent unnecessary deployment triggers
Even if config has not been changed new deployment is triggered because odo changes the order of env variables in DC definition.

Screenshot 2019-04-18 at 17 39 06

mik-dass:
fixes #1633 

This PR prevents the unnecessary roll out of new deployment config.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -984,9 +984,19 @@ func uniqueAppendOrOverwriteEnvVars(existingEnvs []corev1.EnvVar, envVars ...cor
 		mapExistingEnvs[newEnvVar.Name] = newEnvVar
 	}
 
-	// Convert map to slice
-	for _, envVar := range mapExistingEnvs {
-		retVal = append(retVal, envVar)
+	// append the values to the final slice
+	// don't loop because we need them in order
+	for _, envVar := range existingEnvs {
+		if val, ok := mapExistingEnvs[envVar.Name]; ok {
+			retVal = append(retVal, val)
+			delete(mapExistingEnvs, envVar.Name)
+		}
+	}
+
+	for _, newEnvVar := range envVars {
+		if val, ok := mapExistingEnvs[newEnvVar.Name]; ok {
+			retVal = append(retVal, val)
+		}
 	}
 
 	return retVal
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -23,7 +23,6 @@ import (
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/preference"
 	"github.com/openshift/odo/pkg/util"
-
 	// api clientsets
 	servicecatalogclienset "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset/typed/servicecatalog/v1beta1"
 	appsschema "github.com/openshift/client-go/apps/clientset/versioned/scheme"
@@ -1288,7 +1287,7 @@ type dcRollOutWait func(*appsv1.DeploymentConfig, int64) bool
 // if prePatchDCHandler is specified (meaning not nil), then it's applied
 // as the last action before the actual call to the Kubernetes API thus giving us the chance
 // to perform arbitrary updates to a DC before it's finalized for patching
-func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatchDCHandler dcStructUpdater, waitCond dcRollOutWait, currentDC *appsv1.DeploymentConfig, existingCmpContainer corev1.Container) error {
+func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatchDCHandler dcStructUpdater, waitCond dcRollOutWait, currentDC *appsv1.DeploymentConfig, existingCmpContainer corev1.Container, waitForDc bool) error {
 
 	modifiedDC := *currentDC
 
@@ -1319,8 +1318,24 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatc
 		return errors.Wrapf(err, "unable to update DeploymentConfig %s", name)
 	}
 
+	// till we find a better solution for git
+	// it is better, not to follow the logs as the build will anyways trigger a new deployment later
+	if !waitForDc {
+		return nil
+	}
+
+	// we check after the update that the template in the earlier and the new dc are same or not
+	// if they are same, we don't wait as new deployment won't run and we will wait till timeout
+	// inspired from https://github.com/openshift/origin/blob/bb1b9b5223dd37e63790d99095eec04bfd52b848/pkg/apps/controller/deploymentconfig/deploymentconfig_controller.go#L609
 	if reflect.DeepEqual(updatedDc.Spec.Template, currentDC.Spec.Template) {
 		return nil
+	} else {
+		currentDCBytes, err := json.Marshal(currentDC.Spec.Template)
+		updatedDCBytes, err := json.Marshal(updatedDc.Spec.Template)
+		if err != nil {
+			return errors.Wrapf(err, "unable to unmarshal dc")
+		}
+		glog.V(4).Infof("updatedDc Spec.Template: %v, not matching currentDc Spec.Template: %v", string(updatedDCBytes), string(currentDCBytes))
 	}
 
 	// We use the currentDC + 1 for the next revision.. We do NOT use the updated DC (see above code)
@@ -1405,11 +1420,8 @@ func (c *Client) UpdateDCToGit(ucp UpdateComponentParams, isDeleteSupervisordVol
 	}
 
 	var dc appsv1.DeploymentConfig
-	if isDeleteSupervisordVolumes {
-		dc = generateGitDeploymentConfig(ucp.CommonObjectMeta, ucp.ImageMeta.Name, ucp.ImageMeta.Ports, ucp.EnvVars, &ucp.ResourceLimits)
-	} else {
-		dc = updateGitDeploymentConfig(ucp.ExistingDC.DeepCopy(), ucp.CommonObjectMeta, ucp.ImageMeta.Name, ucp.ImageMeta.Ports, ucp.EnvVars, &ucp.ResourceLimits)
-	}
+
+	dc = generateGitDeploymentConfig(ucp.CommonObjectMeta, ucp.ImageMeta.Name, ucp.ImageMeta.Ports, ucp.EnvVars, &ucp.ResourceLimits)
 
 	if isDeleteSupervisordVolumes {
 		// Patch the current DC
@@ -1420,7 +1432,18 @@ func (c *Client) UpdateDCToGit(ucp UpdateComponentParams, isDeleteSupervisordVol
 			ucp.DcRollOutWaitCond,
 			ucp.ExistingDC,
 			existingCmpContainer,
+			true,
 		)
+
+		if err != nil {
+			return errors.Wrapf(err, "unable to update the current DeploymentConfig %s", ucp.CommonObjectMeta.Name)
+		}
+
+		// Cleanup after the supervisor
+		err = c.DeletePVC(getAppRootVolumeName(ucp.CommonObjectMeta.Name))
+		if err != nil {
+			return errors.Wrapf(err, "unable to delete S2I data PVC from %s", ucp.CommonObjectMeta.Name)
+		}
 	} else {
 		err = c.PatchCurrentDC(
 			ucp.CommonObjectMeta.Name,
@@ -1429,18 +1452,12 @@ func (c *Client) UpdateDCToGit(ucp UpdateComponentParams, isDeleteSupervisordVol
 			ucp.DcRollOutWaitCond,
 			ucp.ExistingDC,
 			existingCmpContainer,
+			false,
 		)
 	}
+
 	if err != nil {
 		return errors.Wrapf(err, "unable to update the current DeploymentConfig %s", ucp.CommonObjectMeta.Name)
-	}
-
-	if isDeleteSupervisordVolumes {
-		// Cleanup after the supervisor
-		err = c.DeletePVC(getAppRootVolumeName(ucp.CommonObjectMeta.Name))
-		if err != nil {
-			return errors.Wrapf(err, "unable to delete S2I data PVC from %s", ucp.CommonObjectMeta.Name)
-		}
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


pkg/occlient/occlient.go
```diff
@@ -984,9 +985,19 @@ func uniqueAppendOrOverwriteEnvVars(existingEnvs []corev1.EnvVar, envVars ...cor
...
 
-	// Convert map to slice
-	for _, envVar := range mapExistingEnvs {
-		retVal = append(retVal, envVar)
+	// append the values to the final slice
+	// don't loop because we need them in order
+	for _, envVar := range existingEnvs {
+		if val, ok := mapExistingEnvs[envVar.Name]; ok {
+			retVal = append(retVal, val)
+			delete(mapExistingEnvs, envVar.Name)
```
girishramnani: would it be a better idea to just sort them instead of doing this?

pkg/occlient/templates.go
```diff
@@ -139,6 +139,51 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
 	return dc
 }
 
+func updateSupervisorDeploymentConfig(existingDc *appsv1.DeploymentConfig, commonObjectMeta metav1.ObjectMeta, commonImageMeta CommonImageMeta,
+	envVar []corev1.EnvVar, envFrom []corev1.EnvFromSource, resourceRequirements *corev1.ResourceRequirements) appsv1.DeploymentConfig {
```
girishramnani: too many function args

pkg/occlient/templates.go
```diff
@@ -275,6 +320,44 @@ func generateGitDeploymentConfig(commonObjectMeta metav1.ObjectMeta, image strin
 	return dc
 }
 
+func updateGitDeploymentConfig(existingDc *appsv1.DeploymentConfig, commonObjectMeta metav1.ObjectMeta, image string, containerPorts []corev1.ContainerPort, envVars []corev1.EnvVar, resourceRequirements *corev1.ResourceRequirements) appsv1.DeploymentConfig {
```
girishramnani: too many function args

pkg/occlient/occlient.go
kadel: The order of the env variables is important if you start sorting, you will change the order.
This function is called `appendOrOvewrite` it shouldn't do any side effects like sorting

kadel:
/test e2e-scenarios

```
error: You must be logged in to the server (the server has asked for the client to provide credentials)
```

pkg/occlient/occlient.go
```diff
@@ -1290,22 +1303,26 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatc
...
 	// Update the current one that's deployed with the new Spec.
 	// despite the "patch" function name, we use update since `.Patch` requires
 	// use to define each and every object we must change. Updating makes it easier.
-	_, err := c.appsClient.DeploymentConfigs(c.Namespace).Update(currentDC)
+	updatedDc, err := c.appsClient.DeploymentConfigs(c.Namespace).Update(&modifiedDC)
 	if err != nil {
 		return errors.Wrapf(err, "unable to update DeploymentConfig %s", name)
 	}
 
+	if reflect.DeepEqual(updatedDc.Spec.Template, currentDC.Spec.Template) {
```
kadel: It might be worth to add a comment explaining what lead to this and why are we doing this. I'm sure that we will forget it after some time ;-)

pkg/occlient/occlient.go
```diff
@@ -1387,7 +1404,12 @@ func (c *Client) UpdateDCToGit(ucp UpdateComponentParams, isDeleteSupervisordVol
 		return errors.New("UpdateDCToGit imageName cannot be blank")
 	}
 
-	dc := generateGitDeploymentConfig(ucp.CommonObjectMeta, ucp.ImageMeta.Name, ucp.ImageMeta.Ports, ucp.EnvVars, &ucp.ResourceLimits)
+	var dc appsv1.DeploymentConfig
+	if isDeleteSupervisordVolumes {
```
kadel: the same `if` condition is below, why not combine it?

pkg/occlient/occlient.go
```diff
@@ -1502,20 +1524,30 @@ func (c *Client) UpdateDCToSupervisor(ucp UpdateComponentParams, isToLocal bool,
...
-	// Generate the SupervisorD Config
-	dc := generateSupervisordDeploymentConfig(
-		ucp.CommonObjectMeta,
-		ucp.ImageMeta,
-		inputEnvs,
-		cmpContainer.EnvFrom,
-		&ucp.ResourceLimits,
-	)
+	var dc appsv1.DeploymentConfig
+	if isCreatePVC {
```
kadel: again, the same if the statement is below

pkg/occlient/utils.go
```diff
@@ -84,3 +84,20 @@ func IsDCRolledOut(config *appsv1.DeploymentConfig, desiredRevision int64) bool
 	}
 	return false
 }
+
+func hasUpdatedImages(oldDC *appsv1.DeploymentConfig, newDC *appsv1.DeploymentConfig) bool {
```
kadel: it looks like this function is never used

pkg/occlient/templates.go
```diff
@@ -139,6 +139,51 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
 	return dc
 }
 
+func updateSupervisorDeploymentConfig(existingDc *appsv1.DeploymentConfig, commonObjectMeta metav1.ObjectMeta, commonImageMeta CommonImageMeta,
+	envVar []corev1.EnvVar, envFrom []corev1.EnvFromSource, resourceRequirements *corev1.ResourceRequirements) appsv1.DeploymentConfig {
+
+	dc := *existingDc
+
+	dc.ObjectMeta = commonObjectMeta
+	dc.Spec.Strategy.Type = appsv1.DeploymentStrategyTypeRecreate
```
kadel: this should be already set, we should overwrite it again

pkg/occlient/templates.go
```diff
@@ -139,6 +139,51 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+		"deploymentconfig": commonObjectMeta.Name,
+	}
+
+	dc.Spec.Template.ObjectMeta = metav1.ObjectMeta{
+		Labels: map[string]string{
+			"deploymentconfig": commonObjectMeta.Name,
+		},
+		// https://github.com/openshift/odo/pull/622#issuecomment-413410736
+		Annotations: map[string]string{
+			"alpha.image.policy.openshift.io/resolve-names": "*",
```
kadel: this looks like a duplication of what is already in `occlient.generateSupervisordDeploymentConfig` (templates.go)

pkg/occlient/templates.go
```diff
@@ -139,6 +139,51 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+		},
+	}
+
+	for i := range dc.Spec.Template.Spec.Containers {
```
kadel: all the containers would be the same?

pkg/occlient/templates.go
```diff
@@ -139,6 +139,51 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+
+	dc.ObjectMeta = commonObjectMeta
+	dc.Spec.Strategy.Type = appsv1.DeploymentStrategyTypeRecreate
+
+	dc.Spec.Selector = map[string]string{
+		"deploymentconfig": commonObjectMeta.Name,
+	}
+
+	dc.Spec.Template.ObjectMeta = metav1.ObjectMeta{
+		Labels: map[string]string{
```
kadel: what about already existing labels?

pkg/occlient/templates.go
mik-dass: We have only one container support so I guess we can keep them same for now

mik-dass:
@girishramnani @kadel Fixed. Please have a look again

mik-dass:
> https://travis-ci.org/openshift/odo/jobs/529622044

Let me find a fix for this, marking this as WIP again

mik-dass:
The unit test failed looking into it

kadel:
@mik-dass 

```
./pkg/occlient/occlient_test.go
```

kadel:
/retest

```
error: You must be logged in to the server (the server has asked for the client to provide credentials)

```

pkg/occlient/occlient.go
```diff
@@ -1290,22 +1302,42 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatc
...
+	// inspired from https://github.com/openshift/origin/blob/bb1b9b5223dd37e63790d99095eec04bfd52b848/pkg/apps/controller/deploymentconfig/deploymentconfig_controller.go#L609
+	if reflect.DeepEqual(updatedDc.Spec.Template, currentDC.Spec.Template) {
+		return nil
+	} else {
+		currentDCBytes, err := json.Marshal(currentDC.Spec.Template)
+		updatedDCBytes, err := json.Marshal(updatedDc.Spec.Template)
+		if err != nil {
+			return errors.Wrapf(err, "unable to unmarshal dc")
+		}
+		glog.V(4).Infof("updatedDc Spec.Template: %v, not matching currentDc Spec.Template: %v", string(updatedDCBytes), string(currentDCBytes))
```
kadel: it would be good to add a  little bit more info to that message, like what it means if this is not matching. If I would read just this message in logs and didn't know how this is working I would think that this means that something went wrong

pkg/occlient/templates.go
kadel: then why not directly go for the first container? This will make it really confusing when someones add another container. Going just for the first one would be much safer

mik-dass:
@kadel Fixed. Also I have made changes for the git update strategy. I guess we shouldn't wait for the deployment as of now because the build after it will anyways trigger a new deployment. We can revisit it in https://github.com/openshift/odo/issues/1682

pkg/occlient/templates.go
```diff
@@ -22,6 +22,23 @@ type CommonImageMeta struct {
...
+type SupervisorDUpdateParams struct {
+	existingDc           *appsv1.DeploymentConfig
+	commonObjectMeta     metav1.ObjectMeta
+	commonImageMeta      CommonImageMeta
+	envVar               []corev1.EnvVar
+	envFrom              []corev1.EnvFromSource
+	resourceRequirements *corev1.ResourceRequirements
+}
+
+type GitDeploymentConfigUpdateParams struct {
```
kadel: looks like this is not used anywhere

pkg/occlient/occlient.go
```diff
@@ -1290,22 +1302,42 @@ func (c *Client) PatchCurrentDC(name string, dc appsv1.DeploymentConfig, prePatc
...
 		return errors.Wrapf(err, "unable to update DeploymentConfig %s", name)
 	}
 
+	// till we find a better solution for git
+	// it is better, not to follow the logs as the build will anyways trigger a new deployment later
+	if !waitForDc {
+		return nil
+	}
+
+	// we check after the update that the template in the earlier and the new dc are same or not
```
kadel: 👍

mik-dass:
@kadel Fixed

pkg/occlient/templates.go
```diff
@@ -139,6 +148,35 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+			}
+		}
+		if containerIndex != -1 {
+			dc.Spec.Template.Spec.Containers[containerIndex].Resources = *updateParams.resourceRequirements
+		}
+	}
+
+	return dc
+}
+
```
cdrage: Can you please add comments to the above function of what's happening what why it's needed?

pkg/occlient/templates.go
```diff
@@ -139,6 +148,35 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+	dc.ObjectMeta = updateParams.commonObjectMeta
+
+	if len(dc.Spec.Template.Spec.Containers) > 0 {
+		dc.Spec.Template.Spec.Containers[0].Name = updateParams.commonObjectMeta.Name
+		dc.Spec.Template.Spec.Containers[0].Ports = updateParams.commonImageMeta.Ports
+		dc.Spec.Template.Spec.Containers[0].Env = updateParams.envVar
+		dc.Spec.Template.Spec.Containers[0].EnvFrom = updateParams.envFrom
+	}
+
+	containerIndex := -1
```
cdrage: Never really seen a `-1` been used.. Mind explaining in a comment why?

pkg/occlient/occlient.go
```diff
@@ -1502,27 +1541,38 @@ func (c *Client) UpdateDCToSupervisor(ucp UpdateComponentParams, isToLocal bool,
...
-	// Generate the SupervisorD Config
-	dc := generateSupervisordDeploymentConfig(
-		ucp.CommonObjectMeta,
-		ucp.ImageMeta,
-		inputEnvs,
-		cmpContainer.EnvFrom,
-		&ucp.ResourceLimits,
-	)
+	var dc appsv1.DeploymentConfig
+	if isCreatePVC {
```
cdrage: what happens if `isCreatePVC` isn't true? Would it be correct in saying that supervisord won't be added at all?

pkg/occlient/templates.go
```diff
@@ -139,6 +148,35 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
...
+	if updateParams.resourceRequirements != nil {
+		for index, container := range dc.Spec.Template.Spec.Containers {
```
kadel: similarly as done above, we don't need to loop over all containers

pkg/occlient/occlient.go
kadel: If I'm not mistaken it is an indicator that we are updating from git component to local or binary component and that PVC to hold source code/binary needs to be created. Am I right @mik-dass?

Not a good name variable, but that is something that is probably outside of the scope of this PR.

mik-dass: Yes it is used to indicate if a new supervisorD PVC should be created during the update or not e.g during the update from local/binary to git component. I have added a comment regarding this variable.

pkg/occlient/templates.go
mik-dass: I used it as the container index cannot be -1 and since we can store the container index in the same variable. But I am removing this logic due to https://github.com/openshift/odo/pull/1684#discussion_r282535213

mik-dass:
@cdrage @kadel Fixed

girishramnani:
maybe an integration test would be helpful here if possible @mik-dass

mik-dass:
> maybe an integration test would be helpful here if possible @mik-dass

I have already a added a unit test for the same :)

pkg/occlient/occlient.go
```diff
@@ -1502,27 +1542,38 @@ func (c *Client) UpdateDCToSupervisor(ucp UpdateComponentParams, isToLocal bool,
...
```
cdrage: We should rename this function before the PR gets merged in. :+1:

mik-dass:
@cdrage Fixed

surajnarwade:
```
level=warning msg="Found override for ReleaseImage. Please be warned, this is not advised"
level=info msg="Consuming \"Install Config\" from target directory"
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: Error applying plan:"
level=error
level=error msg="2 errors occurred:"
level=error msg="\t* module.vpc.aws_route_table_association.route_net[1]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.route_net.1: timeout while waiting for state to become 'success' (timeout: 5m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_internet_gateway.igw: 1 error occurred:"
level=error msg="\t* aws_internet_gateway.igw: error attaching EC2 Internet Gateway (igw-05564ebe4b83158c0): timeout while waiting for state to become 'success' (timeout: 2m0s)"
level=error
level=error
level=error
level=error
level=error
level=error msg="Terraform does not automatically rollback in the face of errors."
level=error msg="Instead, your Terraform state file has been partially updated with"
level=error msg="any resources that successfully completed. Please address the error"
level=error msg="above and apply again to incrementally change your infrastructure."
level=error
level=error
```

/retest

PR #1697, opened 2019-05-06
Print appropriate parameter name when trying to set its value again

### LINKED ISSUE #1675: Set preference second time shows incorrect message
```
$ odo preference set Timeout 20
Global preference was successfully updated.

$ odo preference set Timeout 30
? %v is already set. Do you want to override it in the config Yes
Global preference was successfully updated.
```

Note the value of the message after the second run. `? %v is already set`

dharmit:

Prints the name of the parameter whose value the user is trying to set again

fixes #1675


kadel:
```
error: could not run steps: step e2e-scenarios failed: could not wait for template instance to be ready: could not determine if template instance was ready: failed to create objects: object is being deleted: pods "e2e-scenarios" already exists
```
/retest

Issue #1587, opened 2019-04-10
'odo list' command should provide option to show not deployed components in local filesystem along with deployed ones


dgolovin:
[kind/Feature]

## Which functionality do you think we should add?
In odo up to v0.0.20 new component is always deployed to the cluster. That means after successful 'odo create'  command 'odo list' would always list created component. With new workflow that would not work that way anymore, because new component created only in local file-system and not deployed to the cluster.

Would be good to support that in new odo without change, by saving new not deployed component's location in ~/.odo configuration and then inlude them in `odo list` or at least provide search for component in the specified local file system folder(s) with recursion support.

## Why is this needed?
To allow developers and tools see list of all locally created not deployed yet components.

## Workaround 
Devs or tools would have to iterate over local file-system folder and use `odo describe --context folder/path` to test if there component in specified `folder/path`.

kadel:
How about something like `odo component discover `
This command will search given path, and list all components, based on presence of `.odo` dir that it can find in given path


```
odo component discover --path Documents/MyProjects

Name                 Type    Path
myfirstcompnent      java    Documents/MyProjects/MyJavaProject
mysecontcomp         nodejs  Documents/MyProject/MyNodeProject
```

and also with `-o json`

That will be list of all info from the .odo/config.yaml in json format
```json
{
   "kind":"List",
   "apiVersion":"odo.openshift.io/v1alpha1",
   "items":[
      {
         "kind":"LocalConfig",
         "apiversion":"odo.openshift.io/v1alpha1",
         "ComponentSettings":{
            "Type":"java",
            "SourceLocation":"/Users/user/Documents/MyJavaProject",
            "SourceType":"local",
            "Application":"myapp",
            "Project":"myproject",
            "Name":"myfirstcomponent"
         }
      },
      {
         "kind":"LocalConfig",
         "apiversion":"odo.openshift.io/v1alpha1",
         "ComponentSettings":{
            "Type":"nodejs",
            "SourceLocation":"/Users/user/Documents/MyNodeProject",
            "SourceType":"local",
            "Application":"myapp",
            "Project":"myproject",
            "Name":"mysecontcomp"
         }
      }
   ]
}
```

kadel:
Instead of a new command, we could also introduce a new flag to `odo component list`.

That might be better, but it might require to rework how `odo component list` is defined right now. Currently, it is used to list components inside the given application.



Operations that `odo component list` should support:
 - list deployed components in the given application and project
[...]

PR #1710, opened 2019-05-09
resolved the consecutive project deletion issue

girishramnani:
The issue was that the after a project is created it is set as current. But someone deletes the project, the `printDeleteProjectInfo` function takes the consumes the project from the KUBECONFIG rather then 
 the one provided as command line arg.

The resolution was as follows -

1 - set the latest project which we are deleting as current in the file ( so we point to the most recent project ) even though the user has to change the project any way in the future

2 - the SetCurrentProject doesn't actually set the runtime project of the client so I made sure it does, I made this change in the low level function of `occlient` i.e. SetCurrentProject instead of cli's Run() because I felt that SetCurrentProject should do both set the runtime and file project and hence mitigate some future misunderstanding.

fixes https://github.com/openshift/odo/issues/1503

## How to test changes?
```
odo project create odo2-1553171170
odo project create prj4

odo project delete -f prj4
odo project delete -f odo2-1553171170
```

### DIFF
--- a/tests/integration/generic_test.go
+++ b/tests/integration/generic_test.go
@@ -155,6 +155,39 @@ var _ = Describe("odo generic", func() {
 		})
 	})
 
+	Context("project deletion", func() {
+		var originalDir string
+
+		JustBeforeEach(func() {
+			context = helper.CreateNewContext()
+			originalDir = helper.Getwd()
+			helper.Chdir(context)
+		})
+
+		JustAfterEach(func() {
+			helper.Chdir(originalDir)
+			os.RemoveAll(context)
+		})
+		It("be able to delete two project one after the other", func() {
+			project1 := helper.CreateRandProject()
+			project2 := helper.CreateRandProject()
+
+			helper.DeleteProject(project2)
+			helper.DeleteProject(project1)
+		})
+
+		It("be able to delete three project one after the other in opposite order", func() {
+			project1 := helper.CreateRandProject()
+			project2 := helper.CreateRandProject()
+			project3 := helper.CreateRandProject()
+
+			helper.DeleteProject(project1)
+			helper.DeleteProject(project2)
+			helper.DeleteProject(project3)
+
+		})
+	})
+
 	Context("validate odo version cmd with other major components version", func() {
 		It("should show the version of odo major components", func() {
 			odoVersion := helper.CmdShouldPass("odo", "version")
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -519,6 +519,9 @@ func (c *Client) SetCurrentProject(projectName string) error {
 	if err != nil {
 		return errors.Wrapf(err, "unable to switch to %s project", projectName)
 	}
+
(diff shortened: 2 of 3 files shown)


tests/integration/generic_test.go
```diff
@@ -155,6 +155,39 @@ var _ = Describe("odo generic", func() {
 		})
 	})
 
+	Context("project deletion", func() {
+		var originalDir string
+
+		JustBeforeEach(func() {
+			context = helper.CreateNewContext()
+			originalDir = helper.Getwd()
```
girishramnani: project doesn't support `--context` and hence went with the `Chdir` mechanism

girishramnani:
Login failure
/retest

cdrage:
I'm assuming these actions are normal?

```sh
github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h24m
▶ ./odo project create foo
 ✓  New project created and now using project : foo

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h24m
▶ ./odo project create foo2
 ✓  New project created and now using project : foo2

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h24m
▶ ./odo project

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                    4h24m  ⍉
▶ ./odo create nodejs --context ~/nodejs-ex
 ✗  this directory already contains a component


github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                    4h24m  ⍉
▶ ./odo create nodejs --context ~/nodejs-ex
 ✓  Checking component
 ✓  Checking component version

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h24m
▶ ./odo project delete foo -f
 ✓  Deleting project foo
Deleted project : foo

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h25m
▶ oc get

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                    4h25m  ⍉
▶ ./odo list     
 ✗  the current directory does not represent an odo component.
Maybe use 'odo create' to create component here or switch to directory with a component

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                    4h25m  ⍉
▶ ./odo list --context ~/nodejs-ex 
 ✗  There are no components deployed.

github.com/openshift/odo  pr_1710 ✔                                                                                                                                                                                                     4h25m
▶ ./odo project list
ACTIVE     NAME
           foo2
```

In order words if I:

1. Create 2 projects (foo and foo2)
2. Create a component (in foo2)
3. Delete the first project (foo)
4. Shouldn't I be able to do `odo list` no problem?

kadel:
> * Shouldn't I be able to do `odo list` no problem?

Not really. `odo list` is listing only components deployed on the cluster. If you haven't run `odo push` it can't show you anything. https://github.com/openshift/odo/issues/1587 should add support for listing undeployed components.

girishramnani:
@cdrage I dont find any issue with the command output

Issue #1630, opened 2019-04-16
.gitignore and .odoignore not working correctly under Windows


cdrage:
[kind/bug]

## What versions of software are you using?
- Operating System: Windows
- Output of `odo version`: master

## How did you run odo exactly?

Using `odo watch` after deploying a component

## Actual behavior

.odoignore and .gitignore isn't working correctly.

## Expected behavior

To ignore files correctly

## Any logs, error output, etc?

See: https://github.com/openshift/odo/pull/1601 after findings implementing the relative path PR, it looks like ignoring files hasn't worked on Windows for a while..

girishramnani:
@kadel is this only confined to `odo watch` or `odo push` too

cdrage:
@girishramnani It's happening in both watch and push I believe. 100% watch as that is what I've been testing.

cdrage:
IMO should we create an additional document instead of having it only in the getting-started guide for .odoignore and .gitignore?

https://github.com/openshift/odo/blob/c27e25e93a6a940b434444e8743f1d9219235aaa/docs/getting-started.adoc

Ping @boczkowska

PR #1705, opened 2019-05-08
Fixes invalid paths when using rsync with Windows

cdrage:
This PR fixes the invalid paths that are being passed into OpenShift due
to using `FromSlash` that defaults to whatever backslash / forwardslash
the OS uses.

This changes it so that `ToSlash` is used for transfering as *all*
containers in OpenShift are Linux-based and thus forward-slashes are
required.

Also added:
 - Checks to see if a directory is empty and error out.

To test this PR out:
 - Have a Windows-based OS
 - `git clone https://github.com/openshift/nodejs-ex`
 - `odo create nodejs --context ~/nodejs-ex`
 - `odo push -v 4 --context ~/nodejs-ex`
 - `odo watch -v 4 --context ~/nodejs-ex`

Closes issue https://github.com/openshift/odo/issues/1429

Issue #1702, opened 2019-05-07
nodejs-ex fails to copy local files into pod during odo push


theRemix:
[kind/bug]

## What versions of software are you using?
- Operating System: Windows 10 Pro
- Output of `odo version`:
    -  odo v1.0.0-beta1 (c226b008)
        Kubernetes: v1.11.0+d4cacc0

## How did you run odo exactly?

1. `minishift start`
1. `odo login`
1. `git clone https://github.com/openshift/nodejs-ex`
1. `cd nodejs-ex`
1. `odo create nodejs`
1. `odo push`

## Actual behavior

`odo push` fails, files from source directory do not get copied into pod

## Expected behavior

source files are copied from nodejs-ex directory into odo generated pod

## Any logs, error output, etc?

Error: 
```
 odo push
 V  Checking component
 V  Checking component version
 V  Creating nodejs component with name nodejs-xccu
 V  Initializing 'nodejs-xccu' component
 V  Creating component nodejs-xccu
 V  Successfully created component nodejs-xccu
 V  Applying component settings to component: nodejs-xccu
 V  The component nodejs-xccu was updated successfully
 V  Successfully updated component with name: nodejs-xccu
 V  Pushing changes to component: nodejs-xccu of type local
 V  Waiting for component to start
 V  Copying files to component
 <  Building component X  Unable to build files
+ set -eo pipefail
+ '[' '!' -z /opt/app-root/src ']'
+ '[' /tmp '!=' /opt/app-root/src ']'
+ '[' -n /opt/app-root/src-backup ']'
+ '[' '!' -d /opt/app-root/src-backup ']'
+ mkdir -p /opt/app-root/src-backup/src
+ rsync -rlO /tmp/src/. /opt/app-root/src-backup/src/
+ b_IFS='
'
+ b_OFS=
+ OIFS='
'
+ IFS='
'
++ ls -A /tmp/src/
+ for file in '`ls -A ${ODO_S2I_SRC_BIN_PATH}/src/`'
+ rm -fr /opt/app-root/src/.dummy
+ OIFS=
+ IFS='
'
+ '[' -n bucharestgold/centos7-s2i-nodejs ']'
+ '[' bucharestgold/centos7-s2i-nodejs == redhat-openjdk-18/openjdk18-openshift ']'
+ '[' -f /tmp/src/.s2i/bin/assemble ']'
+ '[' -n /usr/libexec/s2i ']'
+ rm -rf /opt/app-root/src/.git
[...]

PR #1711, opened 2019-05-09
Fixes ignore behaviour with Windows

cdrage:
This fixes the error occuring with ignore on Windows due to
back slash being used and not forward slash.

With the way that "Glob" programming works as well as the
library that we use, we cannot use back slashes.

Closes https://github.com/openshift/odo/issues/1630

### DIFF
--- a/pkg/component/watch.go
+++ b/pkg/component/watch.go
@@ -46,6 +46,7 @@ type WatchParameters struct {
 // path is the path of the file or the directory
 // ignores contains the glob rules for matching
 func addRecursiveWatch(watcher *fsnotify.Watcher, path string, ignores []string) error {
+
 	file, err := os.Stat(path)
 	if err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


cdrage:
Notes:

 - https://github.com/openshift/odo/pull/1705 must be merged in before this

The debug output is kind of ugly. Not too sure if we want to change that (some output is backward slashes, the glob matching output is forward slash...)

Proof that it works:

Screenshot from 2019-05-09 15-06-42

kadel:
Hmm, it doesn't seem to work for me :-(

after running `odo push` I don't see any files on the container side :-(

cdrage:
@kadel That's really odd.. What example have you been using? I've been using Python and Node.JS examples

kadel:
> @kadel That's really odd.. What example have you been using? I've been using Python and Node.JS examples

nodejs and one java

kadel:
> @kadel That's really odd.. What example have you been using? I've been using Python and Node.JS examples

I'm probably hitting https://github.com/openshift/odo/issues/1702 bug every time, so I'm not able to verify this :-/

Issue #1654, opened 2019-04-26
odo component delete doesn't actually delete the component


jankleinert:
[kind/bug]

## What versions of software are you using?
- Operating System: macOS
- Output of `odo version`: v1.0.0-beta1

## How did you run odo exactly?
```
odo create nodejs backend
odo push
odo url create backend --port 8080
odo push
odo component delete backend (y, to delete URL)
```

See output below.

## Actual behavior
As you can see from the output below, running odo component delete backend says it's deleted, but the config is still there and I cannot create a new component in that directory.

```
$  odo component get
The current component is: backend
 $  cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: nodejs
  SourceLocation: ./
  SourceType: local
  Application: app
  Project: odo-workshop-h6djc
  Name: backend
  Url:
  - Name: backend
    Port: 8080
 $ odo component delete backend
This component has following urls that will be deleted with component
URL named backend with host backend-app-odo-workshop-h6djc.apps.jkleinert-e293.openshiftworkshop.com h
aving protocol http at port 8080
? Are you sure you want to delete backend from app? Yes
 ✓  Deleting component backend
 ✓  Component backend from application app has been deleted
$ odo component get
The current component is: backend
 $ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: nodejs
  SourceLocation: ./
  SourceType: local
  Application: app
  Project: odo-workshop-h6djc
  Name: backend
  Url:
  - Name: backend
    Port: 8080
  $  odo create nodejs backend2
 ✗  this directory already contains a component
```

## Expected behavior
I expected odo component delete to remove the component configuration. I cannot find a way to do that, meaning I'd have to delete .odo/config.yaml if I want to get rid of my component and create a new one.

cdrage:
Yeah, I ran into the exact same issue too.

Any chance you can pick this up @girishramnani ?

jorgemoralespou:
I think this should be the default behavior, and to delete the components configuration files locally we should do one of either:

- `odo delete --all` (this or some other similar flag makes sense to not introduce a new command)
- `odo destroy` (this makes sense since odo create is the action to create this)

I think we should see which option is best, but not delete local configuration.

jankleinert:
If the default, intended behavior is to delete it from the cluster but not
[...]

PR #1731, opened 2019-05-15
added --all flag for component deletion and test for it

girishramnani:

Added a `--all` flag for `odo component delete` which deletes the component from the cluster as well as removes the local config file and directory.

fixes https://github.com/openshift/odo/issues/1654

## How to test changes?
```
odo create nodejs 
odo push
odo delete --all
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -234,6 +234,11 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value interface{}
 
 }
 
+// DeleteConfigDir Deletes the config directory with the config file
+func (lci *LocalConfigInfo) DeleteConfigDir() error {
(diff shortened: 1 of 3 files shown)


girishramnani:
> Git commit messages could have been better. At the moment, they are vague at best.
> 
> Should `odo delete` prompt for deletion of the config? Shouldn't the user be prompted for it only when they do `odo delete --all`?
> 
> ```shell
> $ odo delete
> ? Are you sure you want to delete python-xwup from app? Yes
>  ✓  Deleting component python-xwup
>  ✓  Component python-xwup from application app has been deleted
> ? Are you sure you want to delete local config for python-xwup? No
>  ✗  Aborting deletion of config for component: python-xwup
> ```

resolved this issue @dharmit

Issue #1647, opened 2019-04-24
remove `odo component get` command


kadel:
It serves no purpose now.
It was used to get the currently active component.

If someone wants to get info about the component they can use `odo component describe`

cdrage:
I'll take this up :+1:

PR #1734, opened 2019-05-15
Hide "component get" command

cdrage:
This PR:
 - Hides the component get command. We do not remove it as `odo
 component get -q` is required for bash/zsh autocompletion support.
 - Updates the "show sub commands" to exclude hidden flags

Closes https://github.com/openshift/odo/issues/1647


girishramnani:
login failure
/retest

PR #1735, opened 2019-05-16
Format and synchronize README and installation page

### LINKED ISSUE #1670: Installation document should be similar to the README doc
[kind/Documentation]

## What did you find missing in the documentation?

https://github.com/openshift/odo/blob/master/docs/installation.adoc and https://github.com/openshift/odo/blob/master/README.adoc should be similar / almost indistinguishable. The only difference between the two should be that the installation doc contains all the "alternative" ways of installation such as by OS (CentOS / Ubuntu, etc.).

## What is the relevance of it?

Both docs should be similar, right now, they're too different.

dharmit:

On README page:
- Removed steps for binary and tarball installation
- Formated the installation section

On Installation page:
- Added bulleted list for quick access to sections
- Added more section headings to support above
- Changed URLs to use latest release so that we don't have to change them for every release. Ref: https://help.github.com/en/articles/linking-to-releases
fixes #1670

## How to test changes?

Take a look at
- https://github.com/dharmit/odo/blob/fix-1670/README.adoc
- https://github.com/dharmit/odo/blob/fix-1670/docs/installation.adoc

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -42,7 +42,9 @@ OpenShift 3.10.0 or above is required.
 
 === Procedure:
 
-==== OS-independent automated installation for Mac and Linux systems
+==== Mac and Linux
+
+===== OS-independent automated installation for Mac and Linux systems
 
 Use this link:./scripts/install.sh[bash script] to quickly install `odo`. It will automatically detect your operating system and install `odo` accordingly.
 
@@ -50,13 +52,13 @@ Use this link:./scripts/install.sh[bash script] to quickly install `odo`. It wil
 $ curl -L https://github.com/openshift/odo/raw/master/scripts/install.sh | bash
 ----
 
-==== Binary installation for Mac OS
+===== Binary installation for Mac OS
 
 ----
 # curl -L https://github.com/openshift/odo/releases/latest/download/odo-darwin-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
 ----
 
-==== Binary installation for Linux
+===== Binary installation for Linux
 
 ----
 # curl -L https://github.com/openshift/odo/releases/latest/download/odo-linux-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
@@ -64,7 +66,7 @@ $ curl -L https://github.com/openshift/odo/raw/master/scripts/install.sh | bash
 
 For more installation options please refer the link:docs/installation.adoc[installation page].
 
-=== Windows
+==== Windows
 
 In order to correctly use `odo` you must download it and add it
 to your `PATH` environment variable:
--- a/README.adoc
+++ b/README.adoc
@@ -42,12 +42,24 @@ OpenShift 3.10.0 or above is required.
 
 === Procedure:
 
-==== OS-independent automated installation for Mac and *nix systems
+==== OS-independent automated installation for Mac and Linux systems
 
 Use this link:./scripts/install.sh[bash script] to quickly install `odo`. It will automatically detect your operating system and install `odo` accordingly.
 
 ----
 $ curl -L https://github.com/openshift/odo/raw/master/scripts/install.sh | bash
+----
+
+==== Binary installation for Mac OS
+
+----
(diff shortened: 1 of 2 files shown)


cdrage:
IMO I think we should keep

```sh
# curl -L https://github.com/openshift/odo/releases/latest/download/odo-linux-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
```

Installation method on the front-page as some people are very OCD / don't like the fact that we're running a random bash script with `sudo`...

Having the choice to download the binary would be ideal :+1:

README.adoc
```diff
@@ -36,60 +36,34 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
...
-----
-# curl -L https://github.com/openshift/odo/releases/download/v1.0.0-beta1/odo-linux-amd64 -o /usr/local/bin/odo && chmod +x /usr/local/bin/odo
-----
-
-*Tarball installation:*
-
-----
-# sh -c 'curl -L https://github.com/openshift/odo/releases/download/v1.0.0-beta1/odo-linux-amd64.gz | gzip -d > /usr/local/bin/odo; chmod +x /usr/local/bin/odo'
-----
+For more installation options please refer the link:docs/installation.adoc[installation page].
```
Preeticp: This sentence should be deleted from here. 
The structure should be:
== Installing ODO
=== OS-independent...
=== Windows
In the concluding statement for the installation section we should add a sentence saying:
For detailed information and more installation options see the link:docs/installation.adoc[installation guide].

docs/installation.adoc
```diff
@@ -3,13 +3,37 @@
...
-
+* The latest `odo` releases are available on link:https://github.com/openshift/odo/releases/latest[Github].
+
+[[prerequisites]]
+== Prerequisites
+
+OpenShift 3.10.0 or above is required.
+
+[[procedure]]
+== Procedure
```
Preeticp: I would skip the title Procedure here, since it is a list of operating systems with a single step in each. Procedure would imply a list of steps to be followed consequentially. Which is why I added Installing `odo`.
How about deleting the sentence on line 15 and adding it as a title: Installation methods for various operating systems
I suggest deleting that line anyway as it repeats information from line 3

docs/installation.adoc
```diff
@@ -125,7 +202,7 @@ The following example demonstrates how to set up a path variable. Your binaries
 . Select *Path* from the *Variable* section and click *Edit*.
 . Click *New* and type `C:\go-bin` into the field or click *Browse* and select the directory, and click *OK*.
 
-==== Setting a PATH variable for Windows 10
+==== Setting the PATH variable for Windows 10
```
Preeticp: `PATH` with back ticks?

dharmit:
@cdrage  +1. I hesitated while keeping *only* the bash script method on the front page. I'll add binary install method for Mac & *nix platforms.

README.adoc
dharmit: If I'm a Linux user, I'd probably not bother to read what's mentioned in the Windows section. Doing what you are suggesting would put the detailed installation steps for *all* OS at the end of Windows section. I'd suggest we put it in both the "Mac & *nix" as well as "Windows" section.

docs/installation.adoc
dharmit: Ack, makes sense!

docs/installation.adoc
dharmit: Ack, thanks for the catch. :)

README.adoc
Preeticp: Sounds good

README.adoc
```diff
@@ -36,60 +36,34 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
...
 
 === Windows
```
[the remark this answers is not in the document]
dharmit: MIssed this one earlier. It's fixed now. Thanks for the catch!

amitkrout:
```
[odo]  ✗  Unable to connect to OpenShift cluster, is it down?
```
/test e2e-scenarios
/refresh

PR #1747, opened 2019-05-20
resolved development docs and bump-version.sh script

girishramnani:
Currently the Readme.adoc doesn't actually show any version for odo but we still try to update that using the bump-version script and also it has been mentioned in the dev docs, so changed that.
Also resolved the extension of the `cli-reference` from `.md` to `.adoc`

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -21,9 +21,6 @@ check_version(){
     echo ""
 }
 
-echo "* Bumping version in README.adoc"
-sed -i "s/v[0-9]*\.[0-9]*\.[0-9]*\(?:-\w+\)\?/${NEW_VERSION}/g" README.adoc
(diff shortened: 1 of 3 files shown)


scripts/bump-version.sh
```diff
@@ -21,9 +21,6 @@ check_version(){
     echo ""
 }
 
-echo "* Bumping version in README.adoc"
```
amitkrout: Why you removed these three line ? Are they not useful atm ?

kadel: Yes, they are not useful anymore, as the version number is not mentioned anywhere in the README. anymore

girishramnani: we don't have any version information in the readme right now

amitkrout: Thanks for the info

PR #1758, opened 2019-05-22
Use local config info for viewing odo config

### LINKED ISSUE #1745: "odo config view" doesn't work when OpenShift cluster is not reachable
[kind/bug]

- Operating System: CentOS 7.6, Fedora 30
- Output of `odo version`:
```
$ odo version
odo v1.0.0-beta1 (3ff77cb)
```

```bash
$ odo config view
COMPONENT SETTINGS
Type              python
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             
Name              python-ccqz
MinMemory         
MaxMemory         
Ignore            
MinCPU            
MaxCPU            

$ minishift stop
Stopping the OpenShift cluster...
Cluster stopped.

$ odo config view
 ✗  Unable to connect to OpenShift cluster, is it down?
```

## Actual behavior
`odo config view` fails when OpenShift cluster is not reachable.

## Expected behavior
Shouldn't viewing the config be possible in spite of cluster being down since the config values are already present in `.odo/config.yaml`?

dharmit:

Using `genericclioptions.Context` required creation of an oc client
connection to the OpenShift server which was not necessary for viewing
the config stored in local config file at `.odo/config.yaml`

fixes #1745 

## How to test changes?

1. Stop the minishift instance used by `odo`: `minishift stop`
2. Navigate to a directory containing the `.odo/config.yaml` file or don't if you'd like to view the global config stored under `~/.odo/config.yaml`
3. Execute `odo config view`

### DIFF
--- a/pkg/odo/cli/config/view.go
+++ b/pkg/odo/cli/config/view.go
@@ -33,6 +33,11 @@ func NewViewOptions() *ViewOptions {
 
 // Complete completes ViewOptions after they've been created
 func (o *ViewOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	cfg, err := config.NewLocalConfigInfo(o.contextDir)
+	if err != nil {
+		return err
+	}
+	o.lci = cfg
 	return
 }
 
@@ -43,13 +48,9 @@ func (o *ViewOptions) Validate() (err error) {
 
 // Run contains the logic for the command
 func (o *ViewOptions) Run() (err error) {
-	cfg, err := config.NewLocalConfigInfo(o.contextDir)
-	if err != nil {
-		return err
-	}
-	cs := cfg.GetComponentSettings()
+	cs := o.lci.GetComponentSettings()
 	w := tabwriter.NewWriter(os.Stdout, 5, 2, 2, ' ', tabwriter.TabIndent)
-	envVarList := cfg.GetEnvVars()
+	envVarList := o.lci.GetEnvVars()
 	if len(envVarList) != 0 {
 		fmt.Fprintln(w, "ENVIRONMENT VARIABLES")
 		fmt.Fprintln(w, "------------------------------------------------")
--- a/pkg/odo/cli/config/view.go
+++ b/pkg/odo/cli/config/view.go
@@ -23,7 +23,7 @@ var viewExample = ktemplates.Examples(`# For viewing the current local configura
 // ViewOptions encapsulates the options for the command
 type ViewOptions struct {
 	contextDir string
-	context    *genericclioptions.Context
+	lci        *config.LocalConfigInfo
 }
 
 // NewViewOptions creates a new ViewOptions instance
@@ -33,7 +33,6 @@ func NewViewOptions() *ViewOptions {
 
 // Complete completes ViewOptions after they've been created
 func (o *ViewOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	o.context = genericclioptions.NewContext(cmd)
 	return
 }
 
--- a/tests/integration/config_test.go
+++ b/tests/integration/config_test.go
@@ -209,5 +209,16 @@ var _ = Describe("odo config test", func() {
 			Expect(configValue).To(Not(ContainSubstring(("PORT"))))
 			Expect(configValue).To(Not(ContainSubstring(("SECRET_KEY"))))
 		})
+
+		It("should list config without logging into the OpenShift cluster", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
+			helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
+			kubeconfigOld := os.Getenv("KUBECONFIG")
+			os.Setenv("KUBECONFIG", "")
+			configValue := helper.CmdShouldPass("odo", "config", "view")
+			Expect(configValue).To(ContainSubstring("hello"))
+			Expect(configValue).To(ContainSubstring("world"))
+			os.Setenv("KUBECONFIG", kubeconfigOld)
+		})
 	})
 })
--- a/tests/integration/config_test.go
+++ b/tests/integration/config_test.go
@@ -214,7 +214,7 @@ var _ = Describe("odo config test", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
 			helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
 			kubeconfigOld := os.Getenv("KUBECONFIG")
-			os.Setenv("KUBECONFIG", "")
+			os.Setenv("KUBECONFIG", "/no/such/path")
 			configValue := helper.CmdShouldPass("odo", "config", "view")
 			Expect(configValue).To(ContainSubstring("hello"))
 			Expect(configValue).To(ContainSubstring("world"))


dharmit:
Thanks @girishramnani for the help on this! :+1:

pkg/odo/cli/config/view.go
```diff
@@ -33,7 +33,6 @@ func NewViewOptions() *ViewOptions {
 
 // Complete completes ViewOptions after they've been created
 func (o *ViewOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	o.context = genericclioptions.NewContext(cmd)
```
girishramnani: you would have to initialize the `LocalConfigInfo` in the Complete stage

girishramnani:
@dharmit might wanna throw in an integration test

dharmit:
I took a look at the test suite and had some discussion around adding test for this particular PR. @kadel suggested to configure `KUBECONFIG` with a value that's incorrect and see if `odo config view` gives expected result.

I see that `KUBECONFIG` is set at: https://github.com/openshift/odo/blob/b6cb9852d1e4cc08f16a2868a94d78d2963df0e1/scripts/configure-installer-tests-cluster.sh#L15 
and test is at: https://github.com/openshift/odo/blob/b6cb9852d1e4cc08f16a2868a94d78d2963df0e1/tests/integration/config_test.go#L208

I'm not sure as to where to set a temporarily invalid value for `KUBECONFIG`. Should it be done at: https://github.com/openshift/odo/blob/b6cb9852d1e4cc08f16a2868a94d78d2963df0e1/Makefile#L154
by doing something like:
```
KUBECONFIG=/no/such/file go test -v github.com/openshift/odo/tests/integration --ginkgo.focus="odo config test" -ginkgo.slowSpecThreshold=$(SLOW_SPEC_THRESHOLD) -ginkgo.v -timeout $(TIMEOUT)
```
or should it be done in `tests/integration/config_test.go` file before the check `configValue := helper.CmdShouldPass("odo", "config", "view")`? The reason I'm not sure as to where to add things is that, I think, if `KUBECONFIG` is not set to correct value as set in `scripts/configure-installer-tests-cluster.sh`, it might cause other tests to fail.

I'm new to both Makefile and testing in Go. So, inputs are more than welcome!

@girishramnani @amitkrout

girishramnani:
@dharmit you might not need to over complicate this. You could just provide the env variable inline by doing something like this `KUBECONFIG="" odo config view`

girishramnani:
This will make the `KUBECONFIG` invalid only for that command and hence would not need any clean up.
`helper.CmdShouldPass("KUBECONFIG=\"\","odo", "config", "view")` should probably do the trick

kadel:
> This will make the `KUBECONFIG` invalid only for that command and hence would not need any clean up.
> `helper.CmdShouldPass("KUBECONFIG=\"\"","odo", "config", "view")` should probably do the trick

I don't think that this will work. I'm afraid that this requires shell (like bash, zsh, sh etc...). 

Also, be aware that we want to run our test on windows.

kadel:
You could use `os.Setenv`  to set the variable in `BeforeEach` or `JustBeforeEach`. This should work everywhere. (Don't forget to backup the original value and then restore it in `AfterEach` or `JustAfterEach`)

dharmit:
Shouldn't we do the storing and restoring of `KUBECONFIG` value just before `configValue := helper.CmdShouldPass("odo", "config", "view")`? I think that `BeforeEach`, `JustBeforeEach` would be executed before `It` block so it will impact other checks within `It` block as well, no?

kadel:
> Shouldn't we do the storing and restoring of `KUBECONFIG` value just before `configValue := helper.CmdShouldPass("odo", "config", "view")`? I think that `BeforeEach`, `JustBeforeEach` would be executed before `It` block so it will impact other checks within `It` block as well, no?

yes that is true. It depends on how your test looks like. Please be aware that `It` block represents one test. Each `It` test should test one thing, and it should be independent of other `it` blocks.

tests/integration/config_test.go
```diff
@@ -209,5 +209,16 @@ var _ = Describe("odo config test", func() {
 			Expect(configValue).To(Not(ContainSubstring(("PORT"))))
 			Expect(configValue).To(Not(ContainSubstring(("SECRET_KEY"))))
 		})
+
+		It("should list config without logging into the OpenShift cluster", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
+			helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
+			kubeconfigOld := os.Getenv("KUBECONFIG")
```
amitkrout: Does it help on travis CI ?, i guess no because the scenario is not actually being tested on a down cluster in travis. Correct me if i am wrong cc_ @kadel @dharmit 

In travis CI ```os.Getenv("KUBECONFIG")``` will return an empty string because no such ```KUBECONFIG``` is set before or after setting up ```oc cluster up```. Then how it is used for travis CI

If this is true  then can we rewrite it in this way 
```
helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
val, ok := os.LookupEnv("CI")
if ok && val == "openshift" {
	kubeconfigOld := os.Getenv("KUBECONFIG")
       os.Setenv("KUBECONFIG", "")
       configValue := helper.CmdShouldPass("odo", "config", "view")
       os.Setenv("KUBECONFIG", kubeconfigOld)
} else {
       helper.CmdShouldPass("odo", "logout")
       configValue := helper.CmdShouldPass("odo", "config", "view")
       helper.CmdShouldPass("odo", "login")
}
Expect(configValue).To(ContainSubstring("hello"))
Expect(configValue).To(ContainSubstring("world"))
```

kadel: > Does it help on travis CI ?, i guess no because the scenario is not actually being tested on a down cluster in travis. Correct me if i am wrong cc_ @kadel @dharmit
> 
> In travis CI `os.Getenv("KUBECONFIG")` will return an empty string because no such `KUBECONFIG` is set before or after setting up `oc cluster up`. Then how it is used for travis CI

That is OK, why it would be a problem if it returns an empty string?

tests/integration/config_test.go
```diff
@@ -209,5 +209,16 @@ var _ = Describe("odo config test", func() {
 			Expect(configValue).To(Not(ContainSubstring(("PORT"))))
 			Expect(configValue).To(Not(ContainSubstring(("SECRET_KEY"))))
 		})
+
+		It("should list config without logging into the OpenShift cluster", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
+			helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
```
kadel: I think that even `odo config set` should work without logging to the cluster

dharmit: It doesn't at the moment. Requires login to the OpenShift cluster. Would that mean that config value gets changed in `.odo/config.yml` and requires connection to a cluster only when doing `odo push --config`? Also, should that be a separate issue and PR of its own?

kadel: yes, to all of it ;-)

tests/integration/config_test.go
amitkrout: > The issue that I see here is on the next line - os.Setenv("KUBECONFIG", ""), @dharmit you need to set it to some path (preferably not existing), not to an empty string. Empty string means that the default path will be used usually .kube/config

I was not sure about the use of empty ```KUBECONFIG```. @kadel Thanks for the clarification.

tests/integration/config_test.go
kadel: Can you please open an issue for it, if not already?

tests/integration/config_test.go
```diff
@@ -214,7 +214,7 @@ var _ = Describe("odo config test", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project)
 			helper.CmdShouldPass("odo", "config", "set", "--env", "hello=world")
 			kubeconfigOld := os.Getenv("KUBECONFIG")
-			os.Setenv("KUBECONFIG", "")
+			os.Setenv("KUBECONFIG", "/no/such/path")
```
kadel: 👍

tests/integration/config_test.go
dharmit: Same here. Was a TIL for me. :smile:

tests/integration/config_test.go
dharmit: Done https://github.com/openshift/odo/issues/1791

amitkrout:
/approve

Issue #1676, opened 2019-04-30
odo create behavior with 'local'


dharmit:
[kind/bug]

## What versions of software are you using?
- Operating System: CentOS Linux release 7.6.1810 (Core)
- Output of `odo version`:
```
$ odo version
odo v1.0.0-beta1 (c226b008)

Server: https://192.168.42.98:8443
Kubernetes: v1.11.0+d4cacc0
```
## How did you run odo exactly?
In the current working directory, I have some code that I would like to run on OpenShift cluster. I did:
```
$ odo component create                                                                                                                         
? Which component type would wish to create python                                                                                                            
? Which version of 'python' component type would you wish to create 3.6                                                                                       
? Location of context component, relative to '/root/django_basic' ./                                                                                          
? Which input type would wish to use for the component local                                                                                                  
? How would you wish to name the new component python-ejpp                                                                                                    
? Which application do you want the commponent to be associated with app                                                                                      
? Which project go you want the component to be created in myproject                                                                                          
? Do you wish to set advanced options No                                                                                                                      
 ✓  Checking component                                                                                                                                        
 ✓  Checking component version                                                                                                                                
Please use `odo push` command to create the component with source deployed                                                                                    
$ odo config view                                                                                                                              
COMPONENT SETTINGS                                                                                                                                            
------------------------------------------------                                                                                                              
PARAMETER         CURRENT_VALUE                                                                                                                               
Type              python:3.6                                                                                                                                  
Application       app                                                                                                                                         
Project           myproject                                                                                                                                   
SourceType        local                                                                                                                                       
Ref                                                                                                                                                           
SourceLocation    /root/django_basic                                                                                                                          
Ports                                                                                                                                                         
Name              python-ejpp                                                                                                                                 
MinMemory                                                                                                                                                     
MaxMemory                                                                                                                                                     
Ignore                                                                                                                                                        
MinCPU
MaxCPU
```

And then did `odo push`. But the container doesn't seem to have any source code:
```
(app-root) sh-4.2$ ls -a
.  ..  .dummy  .pki  .rnd
(app-root) sh-4.2$
```

But if I do just `odo create python` from the same directory, it works. Looking at the config output, I see that `SourceLocation` is set differently in the two cases.
```
$ odo config view
COMPONENT SETTINGS
------------------------------------------------
PARAMETER         CURRENT_VALUE
Type              python
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             
Name              python-srnf
[...]

PR #1764, opened 2019-05-22
Fixes bug of path directories not working correctly with interactive UI

cdrage:
Fixes the path bug when using `odo create` and specifying either:
  - A path outside of the location of the command being ran
  - ./
  - Blank input

This also changes the order of `odo create`. We ask for the type of
deployment (binary, local or git) FIRST and then the directory.

Closes https://github.com/openshift/odo/issues/1676

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -335,8 +335,6 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 			co.componentSettings.SourceLocation = &sourceLocation
 
-			//glog.V(4).Infof("Selected source path: %s, context %s", *(co.componentSettings.SourceLocation), co.componentContext)
-
 		}
 
 		defaultComponentName, err := createDefaultComponentName(co.Context, selectedComponentType, selectedSourceType, selectedSourcePath)
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


surajnarwade:
> This also changes the order of odo create. We ask for the type of
deployment (binary, local or git) FIRST and then the directory.

does this mean, it is mandatory to provide `--binary` or `--git` flag ?

cdrage:
Ready for review!

@girishramnani @mohammedzee1000 @surajnarwade 

To test:

```sh
cd nodejs-ex
odo create
```

Then go through the options of using "local".

girishramnani:
Would it be possible to have test for this, UT/E2E?

cdrage:
@girishramnani 

Would have to create a new test suite. At the moment there are no integration or unit tests for interactive mode.

pkg/odo/cli/component/create.go
```diff
@@ -270,42 +288,55 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+			co.componentContext = ui.EnterInputTypePath("path", currentDirectory, currentDirectory)
+
+			// Get the correct source location
+			sourceLocation, err := getSourceLocation(co.componentContext, currentDirectory)
+			if err != nil {
+				return errors.Wrapf(err, "unable to get source location")
 			}
+			co.componentSettings.SourceLocation = &sourceLocation
+
+			//glog.V(4).Infof("Selected source path: %s, context %s", *(co.componentSettings.SourceLocation), co.componentContext)
```
mik-dass: Comment found

cdrage:
@mik-dass Updated, can you LGTM / approve?

PR #1765, opened 2019-05-24
create the preference file lazily

girishramnani:

recently there was confusion around the global preference file `~/.odo/config.yaml`. some of it was related to why it was present if it was empty. So to reduce this confusion the preference file is created lazily now instead of when NewPreferenceInfo is called.

No issue

## How to test changes?

```
odo preference view # will not create a file
odo preference set Timeout 5000 # will create the file
```

### DIFF
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -117,14 +117,16 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to get odo config file")
 	}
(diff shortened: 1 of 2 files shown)


cdrage:
@girishramnani I feel as though we should also fix the name (use preference.yaml instead) as well as make sure that the file isn't created at times when using `odo create`.

Or should we do this in a separate PR?

girishramnani:
A separate PR @cdrage because that is a breaking change which needs to be communicated and discussed. 
Issue to track this https://github.com/openshift/odo/issues/1766

Issue #1646, opened 2019-04-24
Running odo commands outside of the component directory


kadel:
Improve commands so they can be executed outside of the component context or they provide more information on how to use them.

**This applies to all commands:**
If the command is run "outside the component" and no `--project` flag is set, use project that is set in `KUBECONFIG`. That means that `--project` flag will be optional even if the command is run outside component context.

#### `odo app list`
right now:
```
▶ odo app list
 ✗  the current directory does not represent an odo component.
Maybe use 'odo create' to create component here or switch to a directory with a component
```

Better would be to use the project from `KUBECONFIG` if there is no component in the current directory and no `--project` flag was specified.

```
▶ odo app list

Applications deployed in "myproject" project:
NAME
firstapp
secondapp

```

#### `odo app describe`
Currently the same as `odo app list`
This would be better

```
▶ odo app describe
 ✗  the current directory does not represent an odo component.
Please specify the application name and project. 
HINT: See 'odo app describe --help' for more information.
```

#### `odo app delete`
```
▶ odo app delete
 ✗  the current directory does not represent an odo component.
Please specify the application name.
HINT: See 'odo app delete --help' for more information.
```

- `odo app delete myapp` is valid and it should try to delete the application from the current project
- `odo app delete myapp --project` is valid and it should try to delete the application from the current project
- `odo app delete` is valid only if it is run in the component context, and it will try to delete the whole application where the component is

#### `odo component list`
instead of this:
```
▶ odo component list
 ✗  the current directory does not represent an odo component.
Maybe use 'odo create' to create component here or switch to a directory with a component
```
something like this would be better
```
▶ odo component list
 ✗  the current directory does not represent an odo component.
Please specify project and application to list components deployed in a given project and application.
HINT: See 'odo component list --help' for more information.
```

 `odo component list --app myapp`  is a valid command even outside of the context. It should list components inside the application in default project as set in `KUBECONFIG`

#### `odo component describe`
Currently, the same situation as with `odo component list`
Instead, do this?
```
▶ odo component describe 
 ✗  the current directory does not represent an odo component.
Please specify the name of the deployed component, application, and project.
HINT: See 'odo component describe --help' for more information.
```
 `odo component describe mycomponent --app myapp`  is a valid command even outside of the context. 

#### `odo component delete`
```
▶ odo component delete
 ✗  the current directory does not represent an odo component.
Please specify the name of the deployed component, application, and project.
HINT: See 'odo component delete --help' for more information.
```

[...]

PR #1771, opened 2019-05-27
Added path flag and respective logic

### LINKED ISSUE #1587: 'odo list' command should provide option to show not deployed components in local filesystem along with deployed ones

surajnarwade:
fixes #1587

## How to test changes?

```
[snarwade@redsuraj github.com]$ odo list --context openshift/nodejs-ex --path openshift
NAME            TYPE       SOURCE     STATE            CONTEXT
php-esgq        php        ./         not deployed     /home/snarwade/go/src/github.com/openshift/cakephp-ex
httpd-nesr      httpd      ./         not deployed     /home/snarwade/go/src/github.com/openshift/httpd-ex
nodejs-rtsc     nodejs     ./         deployed         /home/snarwade/go/src/github.com/openshift/nodejs-ex
ruby-kept       ruby       ./         deployed         /home/snarwade/go/src/github.com/openshift/ruby-ex
```

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -908,31 +908,35 @@ func List(client *occlient.Client, applicationName string) (ComponentList, error
 }
 
 // ListIfPathGiven lists all available component in given path directory
-func ListIfPathGiven(client *occlient.Client, path string) (ComponentList, error) {
+func ListIfPathGiven(client *occlient.Client, paths []string) (ComponentList, error) {
 	var components []Component
-	err := filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
-		if strings.Contains(f.Name(), ".odo") {
-			data, err := config.NewLocalConfigInfo(filepath.Dir(path))
-			if err != nil {
-				return err
+	var err error
+	for _, path := range paths {
+		err = filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
+			if strings.Contains(f.Name(), ".odo") {
+				data, err := config.NewLocalConfigInfo(filepath.Dir(path))
+				if err != nil {
+					return err
+				}
+				exist, err := Exists(client, data.GetName(), data.GetApplication())
+				if err != nil {
+					return err
+				}
+				con, _ := filepath.Abs(filepath.Dir(path))
+				a := getMachineReadableFormat(data.GetName(), data.GetType())
+				a.Spec.Source = data.GetSourceLocation()
+				a.Status.Context = con
+				state := "Not Pushed"
+				if exist {
+					state = "Pushed"
+				}
+				a.Status.State = state
+				components = append(components, a)
 			}
-			exist, err := Exists(client, data.GetName(), data.GetApplication())
-			if err != nil {
-				return err
-			}
-			con, _ := filepath.Abs(filepath.Dir(path))
-			a := getMachineReadableFormat(data.GetName(), data.GetType())
-			a.Spec.Source = data.GetSourceLocation()
-			a.Status.Context = con
-			state := "Not Pushed"
-			if exist {
-				state = "Pushed"
-			}
-			a.Status.State = state
-			components = append(components, a)
-		}
-		return nil
-	})
+			return nil
+		})
+
+	}
 	return getMachineReadableFormatForList(components), err
 }
 
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -906,6 +906,31 @@ func List(client *occlient.Client, applicationName string) (ComponentList, error
 	return compoList, nil
 }
 
+// ListIfPathGiven lists all available component in given path directory
+func ListIfPathGiven(path string, client *occlient.Client) (ComponentList, error) {
+	var components []Component
+	err := filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
+		if strings.Contains(f.Name(), ".odo") {
+			data, err := config.NewLocalConfigInfo(filepath.Dir(path))
+			if err != nil {
+				return err
+			}
+			exist, err := Exists(client, data.GetName(), data.GetApplication())
+			if err != nil {
+				return err
+			}
+			con, _ := filepath.Abs(filepath.Dir(path))
+			a := getMachineReadableFormat(data.GetName(), data.GetType())
+			a.Status.Context = con
+			a.Status.State = exist
+			a.Spec.Source = data.GetSourceLocation()
+			components = append(components, a)
+		}
+		return nil
+	})
+	return getMachineReadableFormatForList(components), err
+}
+
 // GetComponentSource what source type given component uses
 // The first returned string is component source type ("git" or "local" or "binary")
 // The second returned string is a source (url to git repository or local path or path to binary)
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -898,6 +898,7 @@ func List(client *occlient.Client, applicationName string) (ComponentList, error
 		if err != nil {
 			return ComponentList{}, errors.Wrap(err, "Unable to get component")
 		}
+		component.Status.State = "Pushed"
 		components = append(components, component)
 
 	}
@@ -921,9 +922,13 @@ func ListIfPathGiven(path string, client *occlient.Client) (ComponentList, error
 			}
 			con, _ := filepath.Abs(filepath.Dir(path))
 			a := getMachineReadableFormat(data.GetName(), data.GetType())
-			a.Status.Context = con
-			a.Status.State = exist
 			a.Spec.Source = data.GetSourceLocation()
+			a.Status.Context = con
+			state := "Not Pushed"
+			if exist {
+				state = "Pushed"
+			}
+			a.Status.State = state
 			components = append(components, a)
 		}
 		return nil
@@ -1301,6 +1306,7 @@ func GetComponent(client *occlient.Client, componentName string, applicationName
 	component.Spec.Env = filteredEnv
 	component.Status.LinkedComponents = linkedComponents
 	component.Status.LinkedServices = linkedServices
+	component.Status.State = "Pushed"
 
 	return component, nil
 }
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -908,7 +908,7 @@ func List(client *occlient.Client, applicationName string) (ComponentList, error
 }
 
 // ListIfPathGiven lists all available component in given path directory
-func ListIfPathGiven(path string, client *occlient.Client) (ComponentList, error) {
+func ListIfPathGiven(client *occlient.Client, path string) (ComponentList, error) {
 	var components []Component
 	err := filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
 		if strings.Contains(f.Name(), ".odo") {
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -913,7 +913,7 @@ func ListIfPathGiven(client *occlient.Client, paths []string) (ComponentList, er
 	var err error
 	for _, path := range paths {
 		err = filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
-			if strings.Contains(f.Name(), ".odo") {
+			if f != nil && strings.Contains(f.Name(), ".odo") {
 				data, err := config.NewLocalConfigInfo(filepath.Dir(path))
 				if err != nil {
 					return err
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -32,7 +32,7 @@ type ComponentList struct {
 // ComponentStatus is Status of components
 type ComponentStatus struct {
 	Context          string              `json:"context,omitempty"`
-	State            bool                `json:"state"`
+	State            string              `json:"state"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
 }
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -33,7 +33,6 @@ type ComponentList struct {
 type ComponentStatus struct {
 	Context          string              `json:"context,omitempty"`
 	State            bool                `json:"state,omitempty"`
-	Active           bool                `json:"active"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
 }
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -31,6 +31,8 @@ type ComponentList struct {
 
 // ComponentStatus is Status of components
 type ComponentStatus struct {
+	Context          string              `json:"context,omitempty"`
+	State            bool                `json:"state,omitempty"`
 	Active           bool                `json:"active"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -32,7 +32,7 @@ type ComponentList struct {
 // ComponentStatus is Status of components
 type ComponentStatus struct {
 	Context          string              `json:"context,omitempty"`
-	State            bool                `json:"state,omitempty"`
+	State            bool                `json:"state"`
 	LinkedComponents map[string][]string `json:"linkedComponents,omitempty"`
 	LinkedServices   []string            `json:"linkedServices,omitempty"`
 }
--- a/pkg/odo/cli/component/list.go
+++ b/pkg/odo/cli/component/list.go
@@ -101,16 +101,21 @@ func (lo *ListOptions) Run() (err error) {
 		if err != nil {
 			return err
 		}
-
+		activeMark := " "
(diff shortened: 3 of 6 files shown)


surajnarwade:
@kadel , @dgolovin please try this and check output and provide feedback (though this WIP PR)
(I havn't looked at sanity checks yet :see_no_evil: )

cdrage:
This is **awesome** and a welcomed change, however, could we align it so that `odo list` shows similarly to `odo list --path` ?

```sh
▶ odo list
ACTIVE     NAME            TYPE
*          nodejs-trym     nodejs

~/nodejs-ex  master ✗                                                                                                                                                                                                                                                                                                                                               223d ◒  
▶ odo list --path ~/nodejs-ex
NAME            TYPE       SOURCE     STATE        CONTEXT
nodejs-trym     nodejs     ./         deployed     /home/wikus/nodejs-ex
```

IMO, I think that having State as well as Context in the normal `odo list` would be beneficial.

What about the `ACTIVE` column too?

Side-note: Deployed I think should be capitalized.

surajnarwade:
@cdrage thanks for the feedback, I added `STATE` for normal `odo list` which will be always `Deployed`, I am confused how to get `CONTEXT` field for normal `odo list`
cc @kadel

cdrage:
@surajnarwade Looks awesome:

```sh
▶ odo list
ACTIVE     NAME                    TYPE       SOURCE                                     STATE
           example-dotnet-sjco     dotnet     file:///tmp/binary/example.war             Deployed
           example-java-fiov       java       file://example.war                         Deployed
           foobar                  nodejs     file://./                                  Deployed
           java-fwwm               java       file://./                                  Deployed
           nodejs-nuvb             nodejs     file://./                                  Deployed
*          nodejs-ynrj             nodejs     https://github.com/openshift/nodejs-ex     Deployed
```

As per https://en.wikipedia.org/wiki/File_URI_scheme the file:// scheme is correct.

I don't see anything wrong with this and I like the implementation.

Another issue I found.

If you create a component and don't deploy it, it doesn't show up as "Undeployed" or "Not deployed"..

Ex:

```sh
odo create nodejs --git https://github.com/openshift/nodejs-ex
odo list
```

surajnarwade:
@cdrage, as of now, I worked on only `--path` functionality as you can see.

`odo list`  (without --path) gathers component info only by talking to cluster, so it not aware of `not Deployed` components from config file which it is using

cdrage:
@surajnarwade Ah, we should then remove the deployed column then if it's not detecting it, or can you add the functionality?

mik-dass:
I think we don't need this active column anymore 

```
[mrinaldas@localhost occlient]$ odo list --path ~/ --context ~/nodejs-ex/
ACTIVE     NAME            TYPE       SOURCE                                  STATE            CONTEXT
           nodejs-knib     nodejs     https://github.com/sclorg/nodejs-ex     not Deployed     /home/mrinaldas/nodejs-3.5
*          nodejs-eakr     nodejs     ./                                      not Deployed     /home/mrinaldas/nodejs-ex
```

pkg/odo/cli/component/list.go
```diff
@@ -52,6 +57,65 @@ func (lo *ListOptions) Validate() (err error) {
 
 // Run has the logic to perform the required actions as part of command
 func (lo *ListOptions) Run() (err error) {
+	var ap []component.Component
+	if lo.pathFlag != "" {
+		err := filepath.Walk(lo.pathFlag, func(path string, f os.FileInfo, err error) error {
```
kadel: Can we put this in some separate function? This `if` statement is getting huge.

surajnarwade: sure

surajnarwade:
@kadel , WIP :smile:

kadel:
> @kadel , WIP 😄

NP, just making sure that it doesn't get lost ;-)

surajnarwade:
@kadel , that will be possible but it will take a time to crawl, probably will not be the good UX, we can add if user asks for it ? WDYT ?

kadel:
> @kadel , that will be possible but it will take a time to crawl, probably will not be the good UX, we can add if user asks for it ? WDYT ?

The user that we are writing this feature for already asked for it ;-) https://github.com/openshift/odo/issues/1587#issuecomment-497820283

dgolovin:
@kadel json without project and application should work, because first we build cluster/project/app hierarchy. Then request list of components for each application using `odo list --project projectName --app appName`, that means we know project and app names for listed components.

surajnarwade:
@dgolovin , can you please test this PR and check whether it is working as per your expectation ?

pkg/component/types.go
```diff
@@ -31,7 +31,8 @@ type ComponentList struct {
 
 // ComponentStatus is Status of components
 type ComponentStatus struct {
-	Active           bool                `json:"active"`
+	Context          string              `json:"context,omitempty"`
+	State            bool                `json:"state"`
```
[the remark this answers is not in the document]
surajnarwade: ok :+1:

pkg/component/component.go
```diff
@@ -900,6 +901,31 @@ func List(client *occlient.Client, applicationName string) (ComponentList, error
 	return compoList, nil
 }
 
+// ListIfPathGiven lists all available component in given path directory
+func ListIfPathGiven(path string, client *occlient.Client) (ComponentList, error) {
```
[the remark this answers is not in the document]
surajnarwade: :+1:  for client place,
/me bad at naming, any suggestions for the name ?

dgolovin:
I have these issues:
1. `odo list --path` always required to be running from existing component's context folder. What if I don't know what folder has a component and just want to search components for specific location.
2. `odo list --path ./not/existing/path` fails with stacktrace 
image
3. `odo list --path location1:location2:location3` fails with error similar to shown above

surajnarwade:
> odo list --path always required to be running from existing component's context folder. What if I don't know what folder has a component and just want to search components for specific location.

@dgolovin this will resolved by upcoming @mik-dass  PR

/me looking into issue 2 and 3.

surajnarwade:
> odo list --path ./not/existing/path fails with stacktrace

I am not able to reproduce this, can you please provide context also

> odo list --path location1:location2:location3 fails with error similar to shown above

for this there are possible solutions:

1. `odo list --path a --path b`
2. `odo list --path a:b`
3. `odo list --path a,b`

what do you prefer ? @dgolovin , @kadel

kadel:
> > odo list --path ./not/existing/path fails with stacktrace
> 
> I am not able to reproduce this, can you please provide context also

@surajnarwade maybe it is windows specific?

> 
> > odo list --path location1:location2:location3 fails with error similar to shown above
> 
> for this there are possible solutions:
> 
> 1. `odo list --path a --path b`
> 2. `odo list --path a:b`
> 3. `odo list --path a,b`
> 
> what do you prefer ? @dgolovin , @kadel

1. this is how cobra does it when you use `StringSliceVar` or `StringSliceVarP`.

surajnarwade:
> this is how cobra does it when you use StringSliceVar or StringSliceVarP.

right, I am following the same, will send soon

surajnarwade:
@dgolovin , @kadel , I updated the PR with `--path a --path b` feature, can you please check and test it ?

kadel:
> > odo list --path ./not/existing/path fails with stacktrace
> 
> I am not able to reproduce this, can you please provide context also

@surajnarwade From the stacktrace it looks like `f` can  be `nil`  for some files (in `filepath.Walk(path, func(path string, f os.FileInfo, err error)` )

surajnarwade:
@kadel , yeah
/me adding case for the same

dgolovin:
@kadel @surajnarwade Using `--path /path1 --path /path2 --path ...`  would require a extra typing, but should work.

I would go with `--path path1:path2:path3` option because it is less typing. It is also 7 characters less per folder and would let handle more folders in one command considering limited command length ~8Kb on latest Windows.
This option seems to be very easy to implement with [SplitList](https://golang.org/pkg/path/filepath/#example_SplitList) function, which would support Windows/Unix file lists transparently.

For extreme cases feeding file list through a file would be required, but that should not be needed soon ;)

surajnarwade:
@dgolovin ack

mik-dass:
It's not working if two path are mentioned and in both `~` is used for representing the home folder.

```
[mrinaldas@localhost ~]$ odo list --context /home/mrinaldas/nodejs-3.5/ --path ~/nodejs-ex/:~/prac/java
NAME            TYPE       SOURCE     STATE          CONTEXT
nodejs-rpqm     nodejs     ./         Not Pushed     /home/mrinaldas/nodejs-ex
```
In the above example, I have a odo component folder in `~/prac/java`

By further digging into the code, I discovered that the `~` for the second argument is not resolved to the home directory.

```
[mrinaldas@localhost ~]$ odo list --context /home/mrinaldas/nodejs-3.5/ --path ~/nodejs-ex/:~/prac/java
/home/mrinaldas/nodejs-ex
~/prac/java
NAME            TYPE       SOURCE     STATE          CONTEXT
nodejs-rpqm     nodejs     ./         Not Pushed     /home/mrinaldas/nodejs-ex
```

kadel:
> @kadel @surajnarwade Using `--path /path1 --path /path2 --path ...` would require a extra typing, but should work.
> 
> I would go with `--path path1:path2:path3` option because it is less typing. It is also 7 characters less per folder and would let handle more folders in one command considering limited command length ~8Kb on latest Windows.
> This option seems to be very easy to implement with [SplitList](https://golang.org/pkg/path/filepath/#example_SplitList) function, which would support Windows/Unix file lists transparently.
> 
> For extreme cases feeding file list through a file would be required, but that should not be needed soon ;)

How many paths do you plan to provide? It doesn't look right to me if you need to provide more than 2 or 3.

surajnarwade:
@mik-dass , I found the problem, go doesn't handle `~`  properly, so we can use `https://github.com/mitchellh/go-homedir` to expand the path if `~` is given

kadel:
I've tried this PR with https://github.com/openshift/odo/pull/1792, and I  still can't run this outside of the odo component directory :-/

```
▶ odo component list  --path .
 ✗  the current directory does not represent an odo component.
Maybe use 'odo create' to create component here or switch to directory with a component
```

@surajnarwade this is slightly different than what is described in https://github.com/openshift/odo/issues/1646. You need to make sure `--path` doesn't require a component directory.

dgolovin:
> How many paths do you plan to provide? It doesn't look right to me if you need to provide more than 2 or 3.

@akadel I would provide as many as added to vscode workspace for now because I really like the opportunity to do it in one odo call. We can also do iteration with one call per added workspace folder. Or even add a check for command length, if it is 
over the max we would bundle as much as we can in one command and do it in several calls.

surajnarwade:
@kadel , https://github.com/openshift/odo/pull/1792 will handle running component list command without requiring component directory

kadel:
> @kadel , #1792 will handle running component list command without requiring component directory

Not in a way this PR requires.

pkg/odo/genericclioptions/context.go
```diff
@@ -118,9 +118,11 @@ func getValidConfig(command *cobra.Command) (*config.LocalConfigInfo, error) {
 		// Gather nessasary info
 		p := command.Parent()
 		r := command.Root()
+		fmt.Println("root is", r.Name(), "parent is", p.Name())
```
kadel: I suspect that this is debugging output left accidentally

pkg/odo/genericclioptions/context.go
```diff
@@ -118,9 +118,11 @@ func getValidConfig(command *cobra.Command) (*config.LocalConfigInfo, error) {
...
 		afs := FlagValueIfSet(command, ApplicationFlagName)
 		// Find the first child of the command. As some groups are allowed even with non existent config
 		fcc := getFirstChildOfCommand(command)
+		fmt.Println("fcc is", fcc.Name(), "command name is", command.Name())
```
kadel: I suspect that this is debugging output left accidentally

pkg/odo/genericclioptions/context.go
surajnarwade: yeah :see_no_evil:

pkg/odo/genericclioptions/context.go
surajnarwade: yeah :see_no_evil:

surajnarwade:
@kadel , test case which I added don't use context flag anyway, so I think it should satisfy the case

kadel:
> @kadel , test case which I added don't use context flag anyway, so I think it should satisfy the case

nope. It doesn't test the situation when you run the command from the directory without a component.
This needs to be tested. Basically, there needs to be a test that will fail without your last commit.

surajnarwade:
@kadel ack 
I will check with @amitkrout and add this test

kadel:
> @kadel ack
> I will check with @amitkrout and add this test

just add one test that calls `odo component list --path` from a directory that doesn't have a component (no .odo directory), doesn't have to be much more complicated than this

surajnarwade:
@kadel , can you please review again :smile_cat:

tests/integration/json_test.go
```diff
@@ -81,9 +81,11 @@ var _ = Describe("odojsonoutput", func() {
 			Expect(desiredSrorageList).Should(MatchJSON(actualSrorageList))
 
 			// odo list -o json --path .
-			pwd := helper.CmdShouldPass("pwd")
+			pwd := helper.Getwd()
 			desired := fmt.Sprintf(`{"kind":"List","apiVersion":"odo.openshift.io/v1alpha1","metadata":{},"items":[{"kind":"Component","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"nodejs","creationTimestamp":null},"spec":{"type":"nodejs","source":"https://github.com/openshift/nodejs-ex"},"status":{"context":"%s","state":"Pushed"}}]}`, strings.TrimSpace(pwd))
-			actual := helper.CmdShouldPass("odo", "list", "-o", "json", "--path", ".")
+			helper.Chdir("/tmp")
```
kadel: this is not going to work on windows

kadel: you can use `helper.CreateNewContext()` instead of this. Don't forget to delete the dir

surajnarwade: ah okay

surajnarwade: done

tests/integration/json_test.go
```diff
@@ -83,9 +83,11 @@ var _ = Describe("odojsonoutput", func() {
 			// odo list -o json --path .
 			pwd := helper.Getwd()
 			desired := fmt.Sprintf(`{"kind":"List","apiVersion":"odo.openshift.io/v1alpha1","metadata":{},"items":[{"kind":"Component","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"nodejs","creationTimestamp":null},"spec":{"type":"nodejs","source":"https://github.com/openshift/nodejs-ex"},"status":{"context":"%s","state":"Pushed"}}]}`, strings.TrimSpace(pwd))
-			helper.Chdir("/tmp")
+			tmpDir := helper.CreateNewContext()
+			helper.Chdir(tmpDir)
 			actual := helper.CmdShouldPass("odo", "list", "-o", "json", "--path", pwd)
 			helper.Chdir(pwd)
+			helper.DeleteDir(tmpDir)
```
kadel: if the `CmdShouldPass` fails the `tmpDir` is not going to be deleted :-(

kadel: call `defer helper.DeleteDir(tmpDir)` right after `tmpDir := helper.CreateNewContext()`

pkg/odo/cli/component/list.go
```diff
@@ -4,18 +4,19 @@ import (
...
-	"github.com/spf13/cobra"
-
 	"github.com/openshift/odo/pkg/component"
 	"github.com/openshift/odo/pkg/log"
 	appCmd "github.com/openshift/odo/pkg/odo/cli/application"
 	projectCmd "github.com/openshift/odo/pkg/odo/cli/project"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	odoutil "github.com/openshift/odo/pkg/odo/util"
+	"github.com/openshift/odo/pkg/odo/util/completion"
+	"github.com/pkg/errors"
```
kadel: why was `"github.com/pkg/errors"` and `"github.com/spf13/cobra"` moved here?

surajnarwade: just IDE things :disappointed:

tests/integration/json_test.go
surajnarwade: updated

pkg/odo/cli/component/list.go
surajnarwade: updated

amitkrout:
/lgtm

Issue #1766, opened 2019-05-24
cannot create component in home directory


girishramnani:
As there is a preference file present in the home directory with the name path as the local config file `.odo/config.yaml`, people cannot create components in their home directory ( which could be something someone does )

girishramnani:
we can consider changing the name of the preference file to `.odo/preference.yaml` instead to make sure people can create components in home directories

cdrage:
:+1: on this, this would be related to the previous issue I had: https://github.com/openshift/odo/issues/1729

:+1: on renaming it to `preference.yaml`

amitkrout:
@girishramnani Can you please assign label to it

girishramnani:
@jorgemoralespou could you please provide you suggestion on these two questions -
1 - renaming the global preference to `preference.yaml` or `preferences.yaml`
2 - do we provide automatic migration for users i.e. rename the `~/.odo/config.yaml` to the new file?

jorgemoralespou:
@girishramnani I would go ahead, although bc I honestly hope no one does that, but makes sense.
For question #2 I would just add it to the release notes as I also don't think many people is yet using the preferences.

PR #1784, opened 2019-06-03
Renamed global preference file to preference.yaml from config.yaml

girishramnani:
The global preference's default name changes to preference.yaml instead of config.yaml. As this file is present in the ~/.odo folder it will not colide with the config.yaml if someone chooses to create a component in home dir.

fixes https://github.com/openshift/odo/issues/1766

## How to test changes?

```
odo preference set Timeout 5 # this should create a ~/.odo/preference.yaml instead of ~/.odo/config.yaml
```

### NOTE - Add this to the ChangeLog as this is a breaking change

### DIFF
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	globalConfigEnvName  = "GLOBALODOCONFIG"
-	configFileName       = "config.yaml"
+	configFileName       = "preference.yaml"
 	preferenceKind       = "Preference"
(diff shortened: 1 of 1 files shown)


girishramnani:
No all the tests should work @cdrage

kadel:
@girishramnani  We are not doing ensuring the backward compatibility after all?

girishramnani:
@kadel according to the discussion in the issue, @jorgemoralespou said that having a changelog entry would be enough

kadel:
@girishramnani can you please add how will the Release change entry look like?

Issue #1707, opened 2019-05-09
odo project create stuck inconsistently


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: macOS
- Output of `odo version`: master

## How did you run odo exactly?
```
$odo login -u <username> -p <password>
$ odo project create test1 -w
 ◐  Waiting for project to come up
```
## Actual behavior
```
$odo login -u <username> -p <password>
✓  Login successful.

You have access to the following projects and can switch between them with 'odo project <project-name>':

  * default
    kube-public
    kube-system
    myproject
[...]

$ odo project create test1 -w
 ◐  Waiting for project to come up
```
## Expected behavior
Should not stuck

amitkrout:
@kadel @mohammedzee1000 Can you please check on your aws cluster

oc i am using:
```
$ oc version
Client Version: version.Info{Major:"4", Minor:"1+", GitVersion:"v4.1.0+21f1917-94-dirty", GitCommit:"21f1917", GitTreeState:"dirty", BuildDate:"2019-05-08T04:30:13Z", GoVersion:"go1.11.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.4+21f1917", GitCommit:"21f1917", GitTreeState:"clean", BuildDate:"2019-05-08T04:27:15Z", GoVersion:"go1.11.8", Compiler:"gc", Platform:"linux/amd64"}
```

amitkrout:
I can also see the same issue on 3.11 too
```
------------------------------
odoCmpSubE2e Test odo push with --source and --config flags when using project flag(--project) and current directory 
  create local nodejs component and push source and code separately
  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/component.go:134
Created dir: /tmp/332880534
Creating a new project: gkhmzeodob
Running odo with args: [project create gkhmzeodob -w] 
[odo]  •  Waiting for project to come up  ...
Setting current dir to: 
Deleting project: kjddnulrrn
Running odo with args: [project delete kjddnulrrn -f] 
[odo]  ✗  The project kjddnulrrn does not exist. Please check the list of projects using `odo project list`
• Failure in Spec Setup (JustBeforeEach) [600.356 seconds]
odoCmpSubE2e
/home/travis/gopath/src/github.com/openshift/odo/tests/integration/cmp_sub_test.go:13
  Test odo push with --source and --config flags
[...]

Issue #1782, opened 2019-05-31
Update from local to git fails


mik-dass:
```
[odo] Successfully pushed //image-registry.openshift-image-registry.svc:5000/jtkpphcors/cmp-git-testing:latest@sha256:08292070fe9a9c0adc7a7d58c45a8d34f54544cbc13d4ef4bc93b7fd6aacf2a6
[odo] Push successful
[odo] I0531 07:30:38.719650   19049 occlient.go:1752] Waiting for cmp-git-testing-1  build to finish
[odo] I0531 07:30:38.944928   19049 occlient.go:1767] Status of cmp-git-testing-1 build is Complete
[odo] I0531 07:30:38.944966   19049 occlient.go:1770] Build cmp-git-testing-1 completed.
[odo] I0531 07:30:38.945099   19049 component.go:1074] Updating the DeploymentConfig cmp-git-testing image to cmp-git-testing:latest
[odo]  ✓  Waiting for build to finish
[odo]  •  Applying component settings to component: cmp-git  ...
[odo] I0531 07:30:38.973846   19049 occlient.go:3266] Deployment cmp-git-testing started.
[odo] I0531 07:30:39.009024   19049 occlient.go:3278] Deployment cmp-git-testing for DeploymentConfig cmp-git-testing triggered.
[odo] I0531 07:30:39.412991   19049 utils.go:72] Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
[odo] I0531 07:30:39.426097   19049 utils.go:72] Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
[odo] I0531 07:30:47.226928   19049 utils.go:72] Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
[odo] I0531 07:30:53.214208   19049 utils.go:72] Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
[odo] I0531 07:30:53.226263   19049 utils.go:80] Waiting for rollout to finish: 0 of 1 updated replicas are available...
[odo] I0531 07:35:39.228848   19049 cmdutils.go:26] Error:
[odo] unable to update DeploymentConfig image for cmp-git component: unable to update the current DeploymentConfig cmp-git-testing: unable to wait for DeploymentConfig cmp-git-testing to update: Timed out waiting for annotation to update
[odo]  ✗  Failed to update config to component deployed
[odo]  ✗  Applying component settings to component: cmp-git
```
It seems to be a flake. It was introduced #1740

[...]

PR #1785, opened 2019-06-03
fix infinite wait in CreateNewProject

kadel:
fixes situation when odo waited forever for a project that was already created

fixes https://github.com/openshift/odo/issues/1707

## How to test changes?
No easy way to test the fix. 
The only way to that I know how to reproduce the previous bug was by using the debugger. 
I had to comment out the `ProjectRequests().Create` call and then create the project manually right before stepping to water creation.

Creating watcher before the project we make sure that we watcher gets the create even in the go channel.

Due to the nature of the bug, it is almost impossible to write a test to cover this situation :-/

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/preference"
 	"github.com/openshift/odo/pkg/util"
+
 	// api clientsets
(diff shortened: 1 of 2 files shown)


mik-dass:
```
 ✗  Applying component settings to component: cmp-git
[odo] I0603 15:48:49.997841   18462 cmdutils.go:26] Error:
[odo] unable to update DeploymentConfig image for cmp-git component: unable to update the current DeploymentConfig cmp-git-testing: unable to start deployment: unable to instantiate Deployment for cmp-git-testing: replicationcontrollers "cmp-git-testing-2" not found
[odo]  ✗  Failed to update config to component deployed
```
/retest

@kadel Let's comment out the update test for now. I will fix it in https://github.com/openshift/odo/issues/1782

PR #1798, opened 2019-06-10
Validate config values (#1549)

### LINKED ISSUE #1549: Validate config values
`odo config set` needs to check if the input value is valid
This can't be allowed:

```
▶ odo config set memory asdf
? memory is already set. Do you want to override it in the config Yes
Local config was successfully updated.

tmp/odo-examples/nodejs-ex  master ✗                                                                                        241d ⚑ ◒
▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: nodejs
  SourceLocation: /Users/tkral/tmp/odo-examples/nodejs-ex
  SourceType: local
  Application: nodejs-ex-wzxy
  Project: tkral
  Name: mynode
  MinMemory: asdf
  MaxMemory: asdf

```

Sensible validation needs to be added to each config option.
Some of the validation: 
 - Correct value for memory and cpu (with units check)
    - memory and cpu should be validated using the same rules as Kuberentes uses (for example support both `M` and `Mi` - https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory)
 - length check for name, project, and application
 - ...

drichtarik:
Checks if CPU parameter is able to be parsed to float64
Checks if Ports parameter compiles with a regex: `[0-9]{1,5}\/(TCP|UDP)`
Checks if Name parameter is longer than 0
Checks Memory, MaxMemory, MinMemory values if they comply to github.com/openshift/odo/pkg/util Parse method. Memory can't be set to 0 as it would set MaxMemory to 0 which does not make sense to me. 

fixes #1549 

## How to test changes?
The commit also provides tests.


drichtarik:
I would like to use [this](https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/libopenstorage/openstorage/pkg/units/units.go) to check the MaxMemory and MinMemory parameters. Thoughts?

drichtarik:
Added Ports check. Regex checks for one to five numbers at the beginning + dash + either TCP or UDP strings.

girishramnani:
@drichtarik tests are failing

girishramnani:
also if you feel that this is ready to review then please remove the `work-in-progress` and move it from Draft PR to actual PR @drichtarik

amitkrout:
@drichtarik test are failing due to the mismatch value of memory in the config file. The memory value is converted to bytes (for example 100M -> 104857600 bytes) whatever the value is passed through the set command. Probably you need to look into that conversion logic.

I think the conversion is controllable/fixed thorough our odo code as the value is not coming from os Level.

drichtarik:
@amitkrout I hope I fixed it correctly.

pkg/config/config.go
```diff
@@ -207,15 +212,41 @@ func (lci *LocalConfigInfo) SetConfiguration(parameter string, value interface{}
 		case "sourcelocation":
 			lci.componentSettings.SourceLocation = &strValue
 		case "ports":
-			arrValue := strings.Split(strValue, ",")
+			arrValue := util.GetSplitValuesFromStr(strValue)
+			fmt.Println(arrValue)
+			for _, value := range arrValue {
+				if !regexp.MustCompile("[0-9]{1,5}\\/(TCP|UDP)").MatchString(value) {
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

girishramnani:
/hold
till GA

pkg/config/config.go
drichtarik: @girishramnani Do I have to address this? The tests have passed.

girishramnani: yes, I would suggest resolving this would be better

dharmit:
It should not be possible to set a `Type` that's not listed in the output of `odo catalog list components` (refer the `Name` column):
```sh
$ odo config set Type noschtype
? Type is already set. Do you want to override it in the config Yes
Local config was successfully updated.
Run `odo push --config` command to apply changes to the cluster.

$ odo catalog list components
Odo Supported OpenShift Components:
NAME       PROJECT       TAGS
nodejs     openshift     10,8,8-RHOAR,latest
java       myproject     latest

Odo Unsupported OpenShift Components:
nodejs      openshift     6
perl        openshift     5.24,5.26,latest
php         openshift     7.0,7.1,latest
python      openshift     2.7,3.5,3.6,latest
ruby        openshift     2.3,2.4,2.5,latest
wildfly     openshift     10.0,10.1,11.0,12.0,13.0,8.1,9.0,latest
```

I think we could change the error response here to indicate that the user has specified invalid value:
```sh
$ odo config set MinMemory -f 50L
 ✗  Unit parse error: invalid fields [50 50  ]

$ odo config set MaxMemory -f 50L
 ✗  Unit parse error: invalid fields [50 50  ]

$ odo config set Memory -f 50L
 ✗  Unit parse error: invalid fields [50 50  
```

mohammedzee1000:
@drichtarik can you rebase please

girishramnani:
can you rebase this PR @drichtarik ?

girishramnani:
closing this PR due to inactivity.

Issue #1473, opened 2019-03-18
run e2e tests in parallel on OpenShift CI


kadel:
Right now e2e tests are run in series on OpenShift CI.
As we add more and more tests it gets slower.

It is a good time to pause and figure out how we can run our tests in parallel before adding more.

It would be great if we could run all tests against the same cluster, to cut down time to spin up new clusters.

girishramnani:
all the tests that not depend on service catalog should be run on openshift CI

kadel:
GinkGo can run tests in parallel -  https://onsi.github.io/ginkgo/#parallel-specs

This would be ideal for us. as we can execute all our tests in one pod.
We just need to make sure that our tests are written in a way where one spec doesn't depend on the other.


ping @amitkrout

amitkrout:
This awesome!!!

Thanks for the reference. Will have a look

amitkrout:
> This would be ideal for us. as we can execute all our tests in one pod.
We just need to make sure that our tests are written in a way where one spec doesn't depend on the other.

As of now i don't think test scenarios will be overlapped to other test file scenario because each test are running in their own namespace. For example: java test running in ```odo-java``` name space. So the only task will be a implementation of GonkGo ```Parallel Specs``` in our make file.

Will check and test locally. If test looks fine then will send the pr

kadel:
@amitkrout problem is that structure of our tests is wrong.
As I understand it GinkGo can run specs in parallel. But that means that each Spec (`It(`) has to be independent,  But that is not true in our e2e. This needs to be fixed first

amitkrout:
> @amitkrout problem is that structure of our tests is wrong.
> As I understand it GinkGo can run specs in parallel. But that means that each Spec (`It(`) has to be independent, But that is not true in our e2e. This needs to be fixed first

Forgot to update the issue with my findings which I got few days back. Sorry for the delay

IMHO in our case it will be very difficult to to isolate and maintain each spec i mean the It(…, func() {} clause. Even there will be a high chance of race condition arises in operation like creating/deleting and possible in intermediate operation. Also isolation increases the duplication of some steps in each spec.

There is an open issue on how to run parallel spec/testsuite in smart way for environment like us - https://github.com/onsi/ginkgo/issues/280. 

But yes I can give a try on single test file with currently ginkgo supported parallel spec run.

kadel:
[...]

PR #1804, opened 2019-06-11
Run app command test in parallel

amitkrout:
fixes - part of #1473

## How to test changes?

make test-cmd-app

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,7 @@ jobs:
         - oc login -u developer
         - travis_wait make test-watch-e2e
         - travis_wait make test-storage-e2e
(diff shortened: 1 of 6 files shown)


amitkrout:
@kadel Please review the ```cmd_app_test.go``` file where i covered only the ```odd app ...``` command test.
@surajnarwade Also need review from your end because i removed some app specific machine readable output scenario from ```json_format``` test and added those here. Comment those missing scenario if any
@mik-dass I used separate context for each spec to achieve maximum isolation. Does it make sense

amitkrout:
#1832 
/test integration

PR #1802, opened 2019-06-10
Rename install.sh to installer.sh and Add Uninstall

danielhelfand:
This pull request is the same as #1762. I am opening a new pull request to address an issue I had with rebasing #1762. 

Closes #1733

PR #1814, opened 2019-06-13
Fixes links to the install script

mik-dass:
after #1802 we need to update the link to the install script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ jobs:
     #   install:
     #     - true
     #   script:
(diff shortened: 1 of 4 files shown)


kadel:
there is a couple more palace where it needs to be updated
 - docs/development.adoc - in Release guide
 - README.adoc -  There is a link to this bash script in Installing section in a sentence "Use this **bash script** to quickly install odo. It will automatically detect your operating system and install odo accordingly."

mik-dass:
@kadel Done

PR #1829, opened 2019-06-17
Run preference and config command test in parallel

amitkrout:
fixes - part of #1473

## How to test changes?

make test-odo-config

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ test-odo-login-e2e:
 # Run config tests
 .PHONY: test-odo-config
 test-odo-config:
-	go test -v github.com/openshift/odo/tests/integration --ginkgo.focus="odo config test" -ginkgo.slowSpecThreshold=$(SLOW_SPEC_THRESHOLD) -ginkgo.v -timeout $(TIMEOUT)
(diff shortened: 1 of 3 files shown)


amitkrout:
```
✗  Unable to connect to OpenShift cluster, is it down?
```
/test integration

Makefile
```diff
@@ -167,7 +167,7 @@ test-odo-login-e2e:
 # Run config tests
 .PHONY: test-odo-config
 test-odo-config:
-	go test -v github.com/openshift/odo/tests/integration --ginkgo.focus="odo config test" -ginkgo.slowSpecThreshold=$(SLOW_SPEC_THRESHOLD) -ginkgo.v -timeout $(TIMEOUT)
+	ginkgo -v -nodes=$(TEST_EXEC_NODES) -focus="odo config test" slowSpecThreshold=$(SLOW_SPEC_THRESHOLD) -randomizeAllSpecs  tests/integration/ -timeout $(TIMEOUT)
```
cdrage: any reason why we're switching to using the `ginkgo` command instead of `go test`? that means that we have to make sure that the CI or anything local has the command already downloaded.

surajnarwade: @cdrage , `-nodes` flag is not supported in `go test` but it is required to run tests in parallel

Issue #1848, opened 2019-06-21
odo project delete: Confusing message about ServiceCatalog


kadel:
When deleting project there is a confusing message about ServiceCatalog not being available.

```
▶ odo project create asdf
 ✓  New project created and now using project : asdf

▶ odo project delete asdf
 ⚠  Unable to access Service Catalog instances, may not be enabled on cluster
? Are you sure you want to delete project asdf Yes
 ✓  Deleting project asdf [6s]
Deleted project : asdf
```

There should be no such message

cdrage:
This was implemented here: https://github.com/openshift/odo/pull/1809

What should we do with regards to the error message? When we perform `application.List` we now also check ServiceInstances.

kadel:
> We could drop this into verbose / hidden output.

Yes, both are ok, up to you. 

The message is pretty confusing especially for the user that doesn't have service catalog enabled on their cluster and never used or deployed any Service, no suddenly they see some message complaining about Service Catalog when deleting a project.

PR #1853, opened 2019-06-24
Drop warning message for Service Catalog

cdrage:
Removes the warning that Service Catalog is not enabled since some
OpenShift clusters may not have them enabled.

Closes https://github.com/openshift/odo/issues/1848

### DIFF
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -6,7 +6,6 @@ import (
 
 	applabels "github.com/openshift/odo/pkg/application/labels"
 	"github.com/openshift/odo/pkg/component"
-	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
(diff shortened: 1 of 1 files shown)


pkg/application/application.go
```diff
@@ -45,7 +44,6 @@ func ListInProject(client *occlient.Client) ([]string, error) {
 	serviceInstanceAppNames, err := client.GetServiceInstanceLabelValues(applabels.ApplicationLabel, applabels.ApplicationLabel)
 	if err != nil {
 		glog.V(4).Infof("Unable to list Service Catalog instances: %s", err)
-		log.Warning("Unable to access Service Catalog instances, may not be enabled on cluster")
```
girishramnani: maybe make it a `glog.V(4).Infof`?

cdrage: see above code... line 46

PR #1635, opened 2019-04-22
add json output for `project create`

syamgk:
add -o json flag for project create
which will print output in machine-readable format

## What is the purpose of this change? What does it change?
adds -o json flag for project  create cmd
## Was the change discussed in an issue?
fixes #???
https://github.com/openshift/odo/issues/1381

## How to test changes?
run `odo project create <project name> -o json` 
and validate the json formatted output

Issue #1381, opened 2019-02-27
machine readable output: odo project create


kadel:
`odo project create -o json` should return JSON formatted output describing the newly created project.

girishramnani:
```
$ odo project create girishramnani -o json
{
    "kind": "Project",
    "apiVersion": "odo.openshift.io/v1alpha1",
    "metadata": {
        "name": "girishramnani",
        "namespace": "girishramnani",
        "creationTimestamp": null
    },
    "message": "Project 'girishramnani' is ready for use"
}
```
`-o` seems to be implemented for `odo project create`.
Hence closing

PR #1854, opened 2019-06-24
Add machine readable output for odo project create

cdrage:
- Adds machine readable output for `odo project create`
- Added test

To test:

```sh
odo project create foobar -o json
```

Closes https://github.com/openshift/odo/pull/1635
Closes https://github.com/openshift/odo/issues/1381

### DIFF
--- a/pkg/odo/cli/project/create.go
+++ b/pkg/odo/cli/project/create.go
@@ -1,6 +1,7 @@
 package project
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/openshift/odo/pkg/log"
@@ -32,6 +33,9 @@ type ProjectCreateOptions struct {
 
 	// generic context options common to all commands
 	*genericclioptions.Context
+
+	// machine-readable output
+	outputFlag string
 }
 
 // NewProjectCreateOptions creates a ProjectCreateOptions instance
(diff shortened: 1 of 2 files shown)


tests/integration/json_test.go
```diff
@@ -30,6 +30,15 @@ var _ = Describe("odojsonoutput", func() {
 		helper.DeleteDir(tmpDir)
 	})
 
+	Context("odo machine create project json output", func() {
+		// odo project create newprojectjson -o json
+		It("should be able to create project and show output in json format", func() {
+			actual := helper.CmdShouldPass("odo project create newprojectjson -o json")
```
dharmit: Should this be `helper.CmdShouldPass("odo", "project", "create", "newprojectjson", "-o", "json")`?

pkg/odo/cli/project/create.go
```diff
@@ -53,6 +57,29 @@ func (pco *ProjectCreateOptions) Validate() (err error) {
 
 // Run runs the project create command
 func (pco *ProjectCreateOptions) Run() (err error) {
+
+	// Machine readable output
+	if pco.OutputFlag == "json" {
+
+		// Create the project
+		err = project.Create(pco.Client, pco.projectName, false)
```
girishramnani: what if someone wants to wait and get a machine readable output as well?

pkg/odo/cli/project/create.go
```diff
@@ -53,6 +57,29 @@ func (pco *ProjectCreateOptions) Validate() (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		// Get the proper machine readable format
+		output := project.GetMachineReadableFormat(pco.projectName, false, nil)
```
girishramnani: I would suggest to move this to end of the function i.e. after the `pco.wait` and and `if` condition for this with `else` printing the standard user message.

pkg/odo/cli/project/create.go
cdrage: What do you mean? This method is done the exact same way in `odo app list` as well in terms of output, etc. We should create the project within the `if` statement rather than outside.

pkg/odo/cli/project/create.go
cdrage: We can't. Output needs to have absolutely no information from `wait` in terms of the the spinner.. so that's why we put it before.

tests/integration/json_test.go
cdrage: Yeah, you're right :) Updated!

cdrage:
Closing in favour of https://github.com/openshift/odo/pull/1916

Issue #1847, opened 2019-06-20
odo component commands are confusing before first push


alanconway:
[kind/bug]

## What versions of software are you using?
- Operating System: 
Linux grommit 5.1.11-300.fc30.x86_64 #1 SMP Mon Jun 17 19:33:15 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- Output of `odo version`:
odo v1.0.0-beta3 (HEAD)

Server: https://192.168.42.249:8443
Kubernetes: v1.11.0+d4cacc0

## Actual behavior

I've put the "expected" comments in-line with the actual odo commands and output:

```
[aconway@grommit nodejs-ex (master *%=)]$ odo create nodejs
 ✓  Validating component [118ms]
Please use `odo push` command to create the component with source deployed
[aconway@grommit nodejs-ex (master *%=)]$ odo component describe
 ✗  component nodejs-kedz does not exist
```
I just created the component so it makes no sense to say it does not exist. I suggest:
```
Component Name: nodejs-kedz
Status: not deployed, use `odo push`
```
Also: `create` should tell me the name of the component it created "nodejs-kedz" as I have no way of knowing that and it is important if there are ever multiple components that I know what I just created.

```
[aconway@grommit nodejs-ex (master *%=)]$ odo component list
 ✗  There are no components deployed.
```
The command is "component list" not "component list-deployed" - it should list the component that I just created because...
```
[aconway@grommit nodejs-ex (master *%=)]$ odo component create
 ✗  this directory already contains a component
```
"list" just told me there are no components, but now I'm told I can't create one because there already is one. Clearly the component *does* exist, it is just in an un-deployed state. `list` should not hide it from me.

kadel:
/bug
/priority medium
/milestone 1.0


### Fixes:

#### `odo component describe`
Needs to work for nondeployed components.
The output needs to contain the information that the component is not pushed, and how to push it.


#### `odo component list`
I agree with you @alanconway it is a little bit confusing. But this one is a little bit trickier, we can't really list all non-deployed components as the only place how to find them would be going through the whole user's filesystem.
But we can at least include a component in the current directory (pushed or not) in the list.

cc @jorgemoralespou

jorgemoralespou:
`odo component list` definition is: List components in the current application, so I would understand the behavior to be:

- Give details about the current component (if no args provided or if -app is provided and matches the app for the current component). Also details whether this component is deployed or not (as status).
[...]

PR #1861, opened 2019-06-25
Add notes with regards to no component deployed when using odo describe

cdrage:
Adds a disclaimer to use `odo push` to push the component if there is no
component deployed.

Closes https://github.com/openshift/odo/issues/1847

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -635,7 +635,7 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 	componentContext := strings.Trim(filepath.Dir(lci.Filename), ".odo")
 
 	if sourceLocation == "" {
-		return "", fmt.Errorf("Blank source location provided")
+		return "", fmt.Errorf("Blank source location, does the .odo directory exist?")
 	}
 
 	if sourceType == GIT {
--- a/pkg/odo/cli/component/describe.go
+++ b/pkg/odo/cli/component/describe.go
@@ -53,7 +53,7 @@ func (do *DescribeOptions) Validate() (err error) {
 		return err
 	}
 	if !existsInCluster {
-		return fmt.Errorf("component %s not found in the OpenShift cluster, use `odo push` to deploy the component", do.componentName)
+		return fmt.Errorf("component %s not pushed to OpenShift cluster, use `odo push` to deploy the component", do.componentName)
 	}
 
 	return odoutil.CheckOutputFlag(do.outputFlag)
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -635,7 +635,7 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 	componentContext := strings.Trim(filepath.Dir(lci.Filename), ".odo")
 
 	if sourceLocation == "" {
-		return "", fmt.Errorf("Blank source location provided")
+		return "", fmt.Errorf("Blank source location, does the .odo directory exist?")
```
girishramnani: `check if .odo directory exists?` <- sounds better?

girishramnani:
/approve
If you feel the comment makes sense then you can do the change else its good to go.

pkg/config/config.go
cdrage: IMO no, but it conveys the same. I don't mind either way. `check if the .odo directory exists` sounds like too many words :(

surajnarwade: I liked this one `does the .odo directory exist`

pkg/odo/cli/component/describe.go
```diff
@@ -48,12 +48,12 @@ func (do *DescribeOptions) Validate() (err error) {
...
 
-	isExists, err := component.Exists(do.Context.Client, do.componentName, do.Context.Application)
+	existsInCluster, err := component.Exists(do.Context.Client, do.componentName, do.Context.Application)
 	if err != nil {
 		return err
 	}
-	if !isExists {
-		return fmt.Errorf("component %s does not exist", do.componentName)
+	if !existsInCluster {
+		return fmt.Errorf("component %s not pushed to OpenShift cluster, use `odo push` to deploy the component", do.componentName)
```
surajnarwade: :+1:

mik-dass: @cdrage Maybe `hasn't been pushed to a OpenShift cluster`?

mik-dass:
/hold
I may be wrong, but please have a look at https://github.com/openshift/odo/pull/1861#discussion_r299827504

pkg/odo/cli/component/describe.go
mik-dass: Also `a openshift cluster` or maybe `any openshift cluster`?

cdrage: "any" or "hasn't" doesn't make sense :( I am going by the issue: https://github.com/openshift/odo/issues/1847

mik-dass: @kadel @jorgemoralespou  TBH the grammar seems out of place to me. Also according to https://github.com/openshift/odo/issues/1847#issuecomment-505934386, `the openhsift` is missing :)

cdrage:
@mik-dass updated. if you could have a look and cancel the hold that'd be great :) I changed the grammar, I missed the `the`.

PR #1867, opened 2019-06-27
Removed active field for application

### LINKED ISSUE #1866: Remove active field from application resource
as of **listApplications** from  https://github.com/openshift/odo/issues/1739, we don't need `active` (status.Active) field anymore in Application

surajnarwade:

Remove `Status.Active` field
Fixes #1866

## How to test changes?

`odo app list -o json` will not show active field

### DIFF
--- a/pkg/application/types.go
+++ b/pkg/application/types.go
@@ -8,8 +8,7 @@ import (
 type App struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              AppSpec   `json:"spec,omitempty"`
-	Status            AppStatus `json:"status,omitempty"`
+	Spec              AppSpec `json:"spec,omitempty"`
 }
 
 // AppSpec is list of components present in given application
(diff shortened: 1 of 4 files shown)


pkg/application/types.go
```diff
@@ -8,8 +8,7 @@ import (
 type App struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              AppSpec   `json:"spec,omitempty"`
-	Status            AppStatus `json:"status,omitempty"`
```
cdrage: Hey so ironically I was working on this at the same time...


You don't have to remove *any* of this. Actually, I would encourage that we still keep it since we may need it for the future (listing all applictions and their status..)

To not have it appear when you unmarshal, change `status,omitempty` to `-`

For example:

```go
Status            AppStatus `json:"-"`
```

surajnarwade: gotcha

kadel: An active application is always the one that context points to. Otherwise, there is no active application, and it needs to be specified explicitly.

With new workflow `active` /`non-active` for applications doesn't make any sense anymore.


We are not using Status field currently for anything else, we can remove it.



Not sure why `Spec` was removed. We are not using it anymore? Then we should probably also remove `AppSpec` type. Or if we do than keep `Spec` as it is.

surajnarwade: @kadel , we didn't removed the `Spec`
thanks for clarification :tada:

surajnarwade:
@cdrage @mik-dass ready for review / lgtm / approve

Issue #1739, opened 2019-05-16
Review JSON output for all odo commands


kadel:
One more acceptance criteria would be to

- [ ] JSON Output for devfile needs to have long term design as we cannot break it after v2

- [ ] review existing commands as mentioned below

List of all commands that vscode adapter uses - https://github.com/redhat-developer/vscode-openshift-tools/blob/8b9baa56a77d02a54a37fd4532a3d1948ed53a01/src/odo.ts#L36

### listProjects
vscode plugin executes:  `oc get project -o jsonpath="{range .items[?(.status.phase == \\"Active\\" )]}{.metadata.name}{\\"\\n\\"}{end}"'`

`odo project list -o json`

*Note: (we don't need/have --project flag for odo project list )*

Question: do we need this flag to retrieve specific project (equivalent can be `odo project get`)

```json
{
    "kind": "List",
    "apiVersion": "odo.openshift.io/v1alpha1",
    "metadata": {},
    "items": [
        {
            "kind": "Project",
            "apiVersion": "odo.openshift.io/v1alpha1",
            "metadata": {
                "name": "myproject",
            },
            "spec": {
                "apps": [
                    "myapp",
                ]
            },
           "status": {
              "active": false
            }
        }
    ]
}
```
TODO: 
 - ~~remove `status.active` field~~ 

### listApplications
vscode plugin executes: `odo application list --project ${project} -o json`

```json
{
    "kind": "List",
    "apiVersion": "odo.openshift.io/v1alpha1",
    "items": [
        {
            "kind": "Application",
            "apiVersion": "odo.openshift.io/v1alpha1",
            "metadata": {
                "name": "app",
                "namespace": "myproject"
            },
            "spec": {
                "components": [
                    "java-jpxy",
                    "nodejs-ngrl",
                    "nodejs-yvnq"
                ]
            }
        }
    ]
}
```
TODO: 
 - remove `status.active` field 
- Issue: https://github.com/openshift/odo/issues/1866

### deleteProject
vscode plugin executes: `odo project delete ${name} -f`

`odo project delete ${name} -f -o json`
On success:
 - exitcode 0, and no other output on stdout or stderr.

On failure:
 - exitcode >0, json output on stderr
```json
{
    "kind": "Error",
    "apiVersion": "odo.openshift.io/v1alpha1",
    "metadata": {
        "namespace": "myproject",
    },
    "message": "Unable to delete project myproject"
}
```

TODO:
- Add `-o json` 

[...]

PR #1870, opened 2019-06-27
Remove status from machine output

cdrage:
Removes the status for machine readable output,

See below example:

```sh
github.com/openshift/odo  remove-status-active-from-application ✔                                                                                                                                                                             11h25m  ⍉
▶ make bin && ./odo project list -o json | jq
go build -ldflags="-w -X github.com/openshift/odo/pkg/odo/cli/version.GITCOMMIT=54b9a583" cmd/odo/odo.go
{
  "kind": "List",
  "metadata": {},
  "items": [
    {
      "kind": "Project",
      "metadata": {
        "name": "foobar",
        "creationTimestamp": null
      },
      "spec": {
        "apps": [
          "app"
        ]
      }
    }
  ]
}

```

See issue: https://github.com/openshift/odo/issues/1739

### DIFF
--- a/pkg/project/types.go
+++ b/pkg/project/types.go
@@ -8,7 +8,7 @@ type Project struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	Spec              ProjectSpec   `json:"spec,omitempty"`
-	Status            ProjectStatus `json:"status,omitempty"`
+	Status            ProjectStatus `json:"-"`
 }
 
 type ProjectSpec struct {


dharmit:
@cdrage maybe I'm missing something here but, how's having `status` in the output a problem? Doesn't it help determine which project is active when the field is present in the output?

mik-dass:
> @cdrage maybe I'm missing something here but, how's having status in the output a problem?

@cdrage Maybe I am missing something too but why is the status field a problem?

cdrage:
@kadel Can chime in, but the status is un-needed for machine readable output. 

It is the same as https://github.com/openshift/odo/pull/1867

See issue: https://github.com/openshift/odo/issues/1739

cdrage:
@kadel After a discussion with Tomas, I'm closing this PR (we dont need it for `odo app list`, but we DO need it for `odo project list`).

dharmit:
You mean this is needed for `odo app list` but NOT for `odo project list`?

cdrage:
@dharmit Yes. We need it for project, but NOT app list.

Issue #1879, opened 2019-07-02
Speed up file copies for odo push


alanconway:
## Which functionality do you think we should update/improve?

Speed of file copies in odo push. Odo file copies are significantly slower than direct rsync or tar via ssh.

## Why is this needed? 

Reduce inner loop time for developers

alanconway:
PR #1880 doesn't improve the user experience, but does improve the internal structure and (hidden) performance of the code. Next step: reduce the overhead in creating the remote "tar" command. A possible solution is to start a single "bash" command at the beginning of an odo session and re-use that for all remote commands that need to occur in sequence.

alanconway:
#1886 will re-use a command connection. After that is implemented, this code should be re-tested to see if it has real benefit before it is comitted.

PR #1849, opened 2019-06-21
Implementation of a walker for smart odo push

mik-dass:
fixes #1788 

How to test:

- run `odo push`
- a `odo-file-index.json` will be generate with all the fileNames, sizes and lastModifiedDates
- rerun and check if the build gets started or not
- use `-f` flag to force build

PR #1880, opened 2019-07-02
Speed up file copies in odo push

alanconway:

Original intent was to speed up odo push file copies:

- Use io.CopyBuffer with a single buffer in place of io.Copy (allocates on every call)
- Clean up filename manipulations (ToSlash, Clean) to happen only once per file
- Separate directory traversal from tar creation: simplicity and reuse
- Pre-compile Glob patterns
- Added benchmarks to measure progress

In isolation the new CopyFiles is about 20% faster.

Sadly in odo, it turns out that the big overhead is the time it takes to connect and start the remote tar receiver process. This PR is worthwhile anyway because it better factors the code for future improvement, and the (hidden) performance improvement may yet surface if we can reduce the connection overhead.

Issue #1879 - not marking fixed yet as there's no apparent speedup for the user.

## How to test changes?

This change is a refactoring, so all existing tests should continue to pass.


alanconway:
Will re-work this change to include the improvements of  #1849.  Ignore until that is done.

alanconway:
Updated as follows:

- walk filesystem once only: check for mods and deleted files, and send actions.
- start walking before WaitAndGetPod but don't send files till ready.
- start sending remote files concurrently with finishing the walk.

@mik-dass the incremental send looks equivalent to your #2030  but I redid the file walker to avoid multiple passes over the file system. Please review the code for anything I missed. Also it should pass your new tests for incremental update.

With these changes the file send takes about 150ms for nodejs-size project (600K) and 500ms when I add the contents of odo/pkg (2.4M)  At this size walking the filesystem is the bottleneck so theres not much difference between -f and an empty copy.

Increased up to 22m a -f copy takes about 6s, an empty or small incremental copy about 1s, so now you can see the difference. Feels like our file system walk is slow but thats a lot better than it was :)

girishramnani:
@alanconway could u please go over the tests, all of them are failing

alanconway:
Missing dependency caused the test failures, I got rid of it as it wasn't really required.

alanconway:
This is odd - all these are unit tests that pass fine on my laptop. Unfortunately I can't fix this until I get back from Vacation on the 30th :(

mik-dass:
> @mik-dass the incremental send looks equivalent to your #2030 but I redid the file walker to avoid multiple passes over the file system. Please review the code for anything I missed. Also it should pass your new tests for incremental update.

Cool I will have a look at the PR asap

pkg/component/component.go
```diff
@@ -640,6 +641,21 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
 		return errors.New(fmt.Sprintf("Directory / file %s is empty", path))
 	}
 
+	// FIXME(alanconway) earlier setup?
```
mik-dass: TODO?

pkg/component/component.go
```diff
@@ -640,6 +641,21 @@ func PushLocal(client *occlient.Client, componentName string, applicationName st
...
+
+	// Start reading local files concurrently, no changes yet.
+	globExps = append(globExps, ".odo", ".git", ".odoignore", ".gitignore")
```
mik-dass: we are already adding ignore rules here https://github.com/openshift/odo/blob/a6970c43592670ea7a1d59627166a168a3fd03a6/pkg/util/util.go#L423

Maybe add these rules there itself?

pkg/component/watch.go
```diff
@@ -134,8 +134,6 @@ var ErrUserRequestedWatchExit = fmt.Errorf("safely exiting from filesystem watch
 //	out: io Writer instance
 // 	parameters: WatchParameters
 func WatchAndPush(client *occlient.Client, out io.Writer, parameters WatchParameters) error {
-	// ToDo reduce number of parameters to this function by extracting them into a struct and passing the struct instance instead of passing each of them separately
```
mik-dass: TODO found

pkg/push/fileset.go
```diff
@@ -0,0 +1,87 @@
...
+	fs = FileSet{Root: normalize(root)}
+	fs.rootSlash = fs.Root
+	if !strings.HasSuffix(fs.rootSlash, string(filepath.Separator)) {
+		fs.rootSlash += string(filepath.Separator)
+	}
+	joined := make([]string, len(ignore))
+	for i, p := range ignore {
+		joined[i] = fs.join(p)
+	}
+	matchAll := fmt.Sprintf("{%s}", strings.Join(joined, ","))
```
mik-dass: Some comments here as to why we are joining the strings and how it will work.

alanconway:
I believe the performance issues are now largely addressed by other changes. 
If we need to revive any of the ideas from this PR, we can open a new issue.

Issue #1822, opened 2019-06-14
Wrong command is shown in UX


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: All supported
- Output of `odo version`: Using the master branch just before the beta3 cut

## How did you run odo exactly?
```
$ odo delete nodejs-inwf -f
 [...]
---
A newer version of odo (version: v1.0.0-beta3) is available.
[...]
If you wish to disable the update notifications, you can disable it by running
'odo config set UpdateNotification false'
```
## Actual behavior
```
[...]
If you wish to disable the update notifications, you can disable it by running
'odo config set UpdateNotification false'
[...]
```
## Expected behavior
If you wish to disable the update notifications, you can disable it by running
**'odo preference set UpdateNotification false'**

PR #1887, opened 2019-07-08
changed config to preference in the new version output

girishramnani:

fixes the incorrect output when a new odo version is available
fixes https://github.com/openshift/odo/issues/1822

## How to test changes?
download an older version of release from odo's release and that should show the change

### DIFF
--- a/pkg/odo/cli/version/version.go
+++ b/pkg/odo/cli/version/version.go
@@ -132,6 +132,6 @@ func GetLatestReleaseInfo(info chan<- string) {
 			"to update manually, or visit https://github.com/openshift/odo/releases\n" +
 			"---\n" +
 			"If you wish to disable the update notifications, you can disable it by running\n" +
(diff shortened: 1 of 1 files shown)


amitkrout:
#1889 
/test integration

pkg/odo/cli/version/version.go
```diff
@@ -132,6 +132,6 @@ func GetLatestReleaseInfo(info chan<- string) {
 			"to update manually, or visit https://github.com/openshift/odo/releases\n" +
 			"---\n" +
 			"If you wish to disable the update notifications, you can disable it by running\n" +
```
dharmit: Why not a simpler message like:
```
You can disable the update notifications by running 'odo preference set UpdateNotification false'
```
instead?

girishramnani: I actually prefer the original message 😅

PR #1892, opened 2019-07-10
Machine readable output changes (global changes + error output)

cdrage:
This PR:
  - Adds `-o json` as a global parameter
  - Makes sure that NO logging is done when using `-o json` with the
  exception of error out / success output via json

### DIFF
--- a/pkg/log/status.go
+++ b/pkg/log/status.go
@@ -110,8 +110,9 @@ func IsTerminal(w io.Writer) bool {
 
 // Start starts a new phase of the status, if attached to a terminal
 // there will be a loading spinner with this status
-func (s *Status) Start(status string, debug bool) {
+func (s *Status) Start(status string, debug bool, json bool) {
 	s.End(true)
+
 	// set new status
 	isTerm := IsTerminal(s.writer)
 	s.status = status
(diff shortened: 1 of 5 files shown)


pkg/machineoutput/types.go
```diff
@@ -0,0 +1,64 @@
+package machineoutput
+
+import (
+	"encoding/json"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const errorKind = "Error"
```
surajnarwade: also present in `pkg/odo/util/cmdutils.go`

pkg/machineoutput/types.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// We print and then exit making sure we exit with an error code..
+	fmt.Println(string(out))
+
+	return nil
+}
+
+// Return success result to fmt.Println
+func SuccessMachineOutput(namespace string, outputMessage string, kind string) error {
```
surajnarwade: also present in `pkg/odo/util/cmdutils.go`

cdrage:
Close in favour of #1898

Issue #1895, opened 2019-07-10
Config set creates local config file


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: Supported platform
- Output of `odo version`: master

## How did you run odo exactly?
```
$ mkdir -p /Users/amit/Context
$ odo config set memory 2Gi --context /Users/amit/Context/
```

## Actual behavior
```
$ odo config set memory 2Gi --context /Users/amit/Context/
Local config was successfully updated.
Run `odo push --config` command to apply changes to the cluster.

$ cat /Users/amit/Context/.odo/config.yaml 
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  MinMemory: 2Gi
  MaxMemory: 2Gi
Amits-MacBook-Pro:odo amit$

```
## Expected behavior
Should report error

amitkrout:
ping @kadel @girishramnani

kadel:
@amitkrout yes, this is a bug. If local config file doesn't exist `odo config set` should fail, not create it.


I'm a decresing priority a little bit. This is not preventing the use of odo, it is just inconvenience that might confuse people.

/priority Medium
/remove-priority High

amitkrout:
> I'm a decreasing priority a little bit. 

@kadel This is a clear case of regression of spec [1] and [2]. In the mentioned spec i am continuously hitting the issue locally. But surprisingly in CI only spec [2] is hit with the issue - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/1843/pull-ci-openshift-odo-master-e2e-scenarios/1046.  

There is something very weird happening. So i would say lets not decrease the priority as it breaks the existing feature and lets propose a fix in the current sprint. Till then i can comment out those affected test spec.

[1] https://github.com/openshift/odo/blob/master/tests/e2escenarios/core_beta_flow.go#L111
[2] https://github.com/openshift/odo/blob/master/tests/e2escenarios/core_beta_flow.go#L128

> This is not preventing the use of odo, it is just inconvenience that might confuse people.

Right, but i am concern about the existing test cases those are broken.

girishramnani:
I remember implementing this, a recent change might have removed the `erroring out`

amitkrout:
> There is something very weird happening. So i would say lets not decrease the priority as it breaks the existing feature and lets propose a fix in the current sprint. Till then i can comment out those affected test spec.

pr #1897 sent

PR #1900, opened 2019-07-11
check if the config file exists for odo config

girishramnani:
Added checks for checking if config file exists before doing `odo config set, view or unset`. also uncommented the failing test

fixes https://github.com/openshift/odo/issues/1895

## How to test changes?
the below commands will fail in a non-component dir
```
odo config set memory 2Gi
odo config unset memory
odo config view
```

### DIFF
--- a/pkg/odo/cli/config/set.go
+++ b/pkg/odo/cli/config/set.go
@@ -5,11 +5,11 @@ import (
 	"strings"
 
 	"github.com/openshift/odo/pkg/odo/cli/ui"
+	"github.com/pkg/errors"
 
 	"github.com/openshift/odo/pkg/config"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
-	"github.com/pkg/errors"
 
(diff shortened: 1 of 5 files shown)


amitkrout:
```
[odo] Pushing to component nodejs of type local
[odo]  •  Waiting for component to start  ...
[odo]  ✗  Waiting for component to start [4m]
[odo]  ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=nodejs-app'
```
/test integration

amitkrout:
> ```
> [odo] Pushing to component nodejs of type local
> [odo]  •  Waiting for component to start  ...
> [odo]  ✗  Waiting for component to start [4m]
> [odo]  ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=nodejs-app'
> ```
> 
> /test integration

@girishramnani Do you have any idea why this error is reported ? What i suspect is that due to test load (Four test spec at a time) cluster becomes slow. 

Additional info - i have observed the same failure most of time in spec contains ```odo push``` just after url create.

Issue #1805, opened 2019-06-11
optimize component.Exists


kadel:
The function is currently using `component.List` which can be expensive. 

This function needs to be optimized so it doesn't require to list all components. The check should be done using just one single api call.

surajnarwade:
as of now `Exists` function uses List, but instead we can use `client.GetDeploymentConfigFromName`, if it returns the output (name of deploymentconfig) that means given component exists.

I would like to work on this :tada:

PR #1905, opened 2019-07-12
optimize component.exists function

surajnarwade:

fasten the `component.Exists` function

fixes https://github.com/openshift/odo/issues/1805

## How to test changes?

`Validating Component` step will be bit faster :smile:


amitkrout:
> /approve
> Great 👍 On my local cluster, the time taken was reduced from 5 sec to ~250 ms!!!

Did observe almost the same time difference on my local cluster. 
/lgtm

Issue #1803, opened 2019-06-11
Inconsistency in odo command exit code


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: All supported platform
- Output of `odo version`: master

## How did you run odo exactly?
```
$ odo app describe test
 ✗  Application test has no components or services deployed.
Amits-MacBook-Pro:odo amit$ echo $?
0
Amits-MacBook-Pro:odo amit$ odo app delete test -f
 ✗  test app does not exists
Amits-MacBook-Pro:odo amit$ echo $?
1
```

## Actual behavior
Inconsistent exit code

## Expected behavior
Exit code should be consistent through out the odo command/sub-command execution. For example:

On a fresh cluster ```odo app describe test``` returns 0 (graceful exit) where as ```odo app delete test -f``` returns ```1```.

In both the case either it should be 0 (graceful exit) or non-zero otherwise.

amitkrout:
ping @kadel

kadel:
If commands results in error, return code needs to be non-zero.

amitkrout:
@girishramnani Found some weird behaviour observed which i believe is not acceptable for a cli tool. Due to the issue #1899 unfortunately it was not reported. When i am fixing the issue via #1902 this is what i am hitting

```
$ odo app describe test --project myproject
 ✗  Application test has no components or services deployed.

$ echo $?
0

$ odo app describe test --project myproject > out 2>error

$ cat out 

$ cat error 
 ✗  Application test has no components or services deployed.
```

**Analysis** 

Actual - For a graceful exit, message is redirected to stderr chanel
Expected - For a graceful exit, message should be redirected to stdout chanel

Due to this inconsistency in behaviour pr #1902 is failing. I am increasing the priority to bring more attention to this issue. For now i am commenting the assertion statment to make CI pass.

Issue #1832, opened 2019-06-18
[Flake] push fails followed by push --config and then URL creation


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: Supported platform
- Output of `odo version`: master

## How did you run odo exactly?
Check ```Any logs, error output, etc?``` tab

## Actual behavior
Push fails 

## Expected behavior
Push should pass

## Any logs, error output, etc?
```
Running odo with args: [create php testcmp --app e2e-xyzk --project nylolzxmxq --git https://github.com/appuio/example-php-sti-helloworld] 
[odo]  •  Validating component  ...
[odo]  ✓  Validating component [2562047h47m16.854775807s]
[odo] Please use `odo push` command to create the component with source deployed
Running odo with args: [config set Ports 8080/TCP] 
[odo] Local config was successfully updated.
[odo] Run `odo push --config` command to apply changes to the cluster.
Running odo with args: [push --config] 
[odo] Validation
[odo]  •  Validating component  ...
[odo]  ✓  Validating component [2562047h47m16.854775807s]
[odo]  •  Checking component  ...
[odo]  ✓  Checking component [2562047h47m16.854775807s]
[odo] 
[odo] Configuration changes
[odo]  •  Creating component  ...
[odo]  ✓  Initializing component
[odo]  •  Triggering build from git  ...
[odo]  ✓  Triggering build from git [2562047h47m16.854775807s]
[odo]  •  Waiting for build to finish  ...
[odo]  ✓  Waiting for build to finish [2562047h47m16.854775807s]
[odo]  •  Deploying component testcmp  ...
[odo]  ✓  Deploying component testcmp [2562047h47m16.854775807s]
[odo]  ✓  Creating component [2562047h47m16.854775807s]
[odo]  •  Applying configuration  ...
[odo]  ✓  Applying configuration [2562047h47m16.854775807s]
Running odo with args: [url create myurl --port 8080] 
[odo]  ✓  URL myurl created
[odo] 
[odo] Run `odo push` to apply URL: myurl
Running odo with args: [push] 
[odo] Validation
[odo]  •  Validating component  ...
[odo]  ✓  Validating component [2562047h47m16.854775807s]
[odo]  •  Checking component  ...
[odo]  ✓  Checking component [2562047h47m16.854775807s]
[odo] 
[odo] Configuration changes
[odo]  •  Applying configuration  ...
[odo]  •  Triggering build from git  ...
[...]

PR #1919, opened 2019-07-18
having no components in an app shouldn't be an error

girishramnani:
having no components in an app shouldn't be an error

fixes https://github.com/openshift/odo/issues/1803

## How to test changes?
```
odo app describe test --project myproject > out 2>error
```
check `out` file, now it should have the `Application test has no components or services deployed.` message


amitkrout:
#1832 
/test integration

Issue #1846, opened 2019-06-20
Binary component push fails with context flag


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: I tested it on macOS
- Output of `odo version`: master

## How did you run odo exactly?
```
$ odo create wildfly javaee-war-test --project myproject --binary tests/examples/binary/java/wildfly/ROOT.war --context tests/examples/binary/java/wildfly/ ✓  Validating component [112ms]
Please use `odo push` command to create the component with source deployed


$ odo push --context tests/examples/binary/java/wildfly/
Validation
 ✓  Validating component [86ms]
 ✓  Checking component [3ms]

Configuration changes
 ✓  Initializing component
 ✓  Creating component [365ms]
 ✓  Applying configuration [0s]

Pushing to component javaee-war-test of type binary
 ✗  open /Users/amit/go/src/github.com/openshift/odo/tests/examples/binary/java/wildfly/tests/examples/binary/java/wildfly: no such file or directory
```

## Actual behavior
Push fails with context flag

## Expected behavior
Should push binary component 

## Any logs, error output, etc?
Check ```How did you run odo exactly?``` section

amitkrout:
ping @kadel

cdrage:
I can investigate this. I have a feeling it has to do with my changes to `context` and paths...

/assign cdrage

girishramnani:
so essentials whats happening here is that the `componentContext` and `SourceLocation` are joined to get the sourceOsPath.
In this case both are `tests/examples/binary/java/wildfly/ROOT.war` and `tests/examples/binary/java/wildfly` which get appended together generating `tests/examples/binary/java/wildfly/tests/examples/binary/java/wildfly` which doesn't exist

The approach to solve this should be that the `SourceLocation` should be saved as a path relative to `componentContext` because in the end we are joining them both here 
https://github.com/openshift/odo/blob/a15c0615151ed3435c11cb427b9f5a071d800737/pkg/config/config.go#L655

I also feel that there will be multiple places this can be reproduced - not just here

cc @amitkrout @kadel

girishramnani:
https://github.com/openshift/odo/issues/1922#issuecomment-512885702

Issue #1922, opened 2019-07-18
switching directory between odo create and odo push fails


girishramnani:
As the sourceLocation is stored relative to the directory `odo create` is run it, when the directory is switched to run `odo push` it fails.
```
[gramnani@girish integration]$ odo create java sb-jar-test --project jxxjiebvvn --binary ../examples/binary/java/openjdk/sb.jar --context /tmp/294104207
 ✓  Validating component [4s]
Please use `odo push` command to create the component with source deployed

[gramnani@girish integration]$ cat /tmp/294104207/.odo/config.yaml 
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: java
  SourceLocation: ../examples/binary/java/openjdk/sb.jar
  SourceType: binary
  Ports:
  - 8080/TCP
  - 8443/TCP
  - 8778/TCP
  Application: app
  Project: jxxjiebvvn
  Name: sb-jar-test
--> [gramnani@girish integration]$ cd ..
[gramnani@girish tests]$ odo push --context /tmp/294104207
 ✓  Creating project jxxjiebvvn
 ✓  Successfully created project jxxjiebvvn
Validation
 ✓  Validating component [5s]
 ✓  Checking component [312ms]

Configuration changes
 ✓  Initializing component
 ✓  Creating component [3s]
 ✓  Applying configuration [41320ns]

Pushing to component sb-jar-test of type binary
 ✗  open /tmp/examples/binary/java/openjdk: no such file or directory
 ```

**Comment by girishramnani** [comment]
This is happening because we store the `SourceLocation` relative to the currentDirectory where is running. This is working currently because the tests run the `odo create` and `odo push` from the same directory making currentDirectory same in both cases but in the real world that wouldn't be the case.
Hence we should make sure that `SourceLocation` is set relative to the `context` dir and hence it wont be dependent on where the user runs `odo` from.
This certainly has a down side that l`SourceLocation` of the component is case of `--local` and `--binary` will always need to be relative to `componentContext` due to this, commands like
```
odo create java sb-jar-test --project jxxjiebvvn --binary ../examples/binary/java/openjdk/sb.jar --context /tmp/294104207
```
will fail. i.e. creations in which binary location is not a child or relative to context dir will fail.
cc @kadel @jorgemoralespou @cdrage

jorgemoralespou:
the source location has to always be sub directory of the context if binary or source. That has to be a requirement, otherwise too many problems can arise.

[...]

PR #1923, opened 2019-07-18
use currentDir in when getting source path

girishramnani:
This is an alternate approach to solving https://github.com/openshift/odo/issues/1846. In this approach we join `currentWorkingDirectory` to `SourceLocation` which works because we actually store the `SourceLocation` related to the `currentWorkingDirectory` - here https://github.com/openshift/odo/blob/ad79434c2276eb3166766178c869849c86cd8e57/pkg/odo/cli/component/create.go#L316 and here https://github.com/openshift/odo/blob/ad79434c2276eb3166766178c869849c86cd8e57/pkg/odo/cli/component/create.go#L335

fixes https://github.com/openshift/odo/issues/1846

## How to test changes?

```
odo create wildfly javaee-war-test --project myproject --binary tests/examples/binary/java/wildfly/ROOT.war --context tests/examples/binary/java/wildfly/
odo push --context tests/examples/binary/java/wildfly/
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -630,10 +630,6 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 	sourceType := lci.GetSourceType()
 	sourceLocation := lci.GetSourceLocation()
 
-	// Get the component context folder
-	// ".odo" is removed as lci.Filename will always return the '.odo' folder.. we don't need that!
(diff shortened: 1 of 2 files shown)


girishramnani:
Problem with this approach is that we are embracing this issue https://github.com/openshift/odo/issues/1922 as it gives us the benefit to have `binary` or `local` location totally different from the `context` dir

girishramnani:
We decided to go other route

Issue #1825, opened 2019-06-14
push fails if context refer to directory having name only in numbers


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: All supported platform
- Output of `odo version`: master

## How did you run odo exactly?
```
$ odo create php --context /Users/amit/1234/
$ odo push --context /Users/amit/1234/
```
## Actual behavior
local config.yaml
```
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: php
  SourceLocation: ./
  SourceType: local
  Application: app
  Project: ifvxbfgkrp
  Name: 1234-php-ykle
```

## Expected behavior

local config.yaml

kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: php
  SourceLocation: ./
  SourceType: local
  Application: app
  Project: ifvxbfgkrp
  **Name: php-ykle**

Do not prefix context dir name in component name
## Any logs, error output, etc?
```
$ odo create php --context /Users/amit/1234/
 ✓  Validating component [106ms]
Please use `odo push` command to create the component with source deployed

$ odo push --context /Users/amit/1234/
Validation
 ✓  Validating component [213ms]
 ✓  Checking component [264ms]

Configuration changes
 ✓  Initializing component
 ◐  Creating component ✗  Failed to create component with name 1234-php-ykle. Please use `odo config view` to view settings used to create component. Error: Service "1234-php-ykle-app" is invalid: metadata.name: Invalid value: "1234-php-ykle-app": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
unable to create Service for 1234-php-ykle-app
[...]
```

kadel:
This is inconsistent behavior that needs to be fixed.


Running `odo create` commands using the same directory `test` results in a different naming schema if `--context` flag is specifed with path.

```
▶ odo create nodejs --context /Users/tkral/tmp/odo-examples/test/
▶ cat .odo/config.yaml | grep Name
  Name: test-nodejs-sinu

▶ cd ..
▶ odo create nodejs --context test/
▶ cat .odo/config.yaml | grep Name
  Name: test-nodejs-tpqh

▶ cd test
▶ odo create nodejs --context .
[...]

PR #1933, opened 2019-07-21
consistent component name and resolved failing numeric dir push

girishramnani:
- Changed the ordering of prefix and component type when creating a default component name.
- Use of absolute path for source path when generating the default component name when source type is local

fixes https://github.com/openshift/odo/issues/1825

## How to test changes?
```
mkdir nodejs-test
odo create nodejs --context ./nodejs-test
cd nodejs-test
odo create nodejs --context .
```
should give the same component name

also
```
mkdir 1234
odo create nodejs --context ./1234
cd 1234
npm init --yes  #just so it doesn't fail on package.json not found error
odo push 
```
shouldn't fail due to `a DNS-1035 label must consist of lower case alphanumeric characters or '-'`

### DIFF
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -1,6 +1,8 @@
 package integration
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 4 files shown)


tests/integration/component.go
```diff
@@ -496,4 +510,26 @@ func componentTests(args ...string) {
 			helper.Chdir(originalDir)
 		})
 	})
+
+	Context("Creating component with numeric named context", func() {
+		var context string
+		JustBeforeEach(func() {
+			var err error
+			ts := time.Now().UnixNano()
+			context, err = ioutil.TempDir("", fmt.Sprint(ts))
```
amitkrout: Instead of ```fmt.Sprint(ts)``` can you pass something simple like ```1234```.

tests/integration/component.go
```diff
@@ -496,4 +510,26 @@ func componentTests(args ...string) {
 			helper.Chdir(originalDir)
 		})
 	})
+
+	Context("Creating component with numeric named context", func() {
+		var context string
```
amitkrout: Already declared as global. You can remove it from here

girishramnani: There is a reason for this here, as the test run parallel - this could get changed being a different context

girishramnani: Hence a different variable then the global one

tests/integration/component.go
girishramnani: ts is a int64 so cannot directly pass it and We should avoid using constants as the next test which requires similar naming the developer wouldn't remember there is `1234` used somewhere and hence they shouldn't use that ( parallel running test could clash )

tests/integration/component.go
amitkrout: Right

girishramnani: Resolving

tests/integration/component.go
amitkrout: I initiated the discussion because ```.UnixNano()``` naming looks bit counter intuitive to me. Anyway thanks for the details.

/lgtm

PR #1958, opened 2019-07-29
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in
golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile.

### DIFF
--- a/tests/helper/helper_http.go
+++ b/tests/helper/helper_http.go
@@ -21,6 +21,8 @@ func HttpWaitFor(url string, match string, maxRetry int, interval int) {
 	for i := 0; i < maxRetry; i++ {
 		fmt.Fprintf(GinkgoWriter, "try %d of %d\n", i, maxRetry)
 
+		// #nosec
+		// gosec:G107 -> This is safe since it's just used for testing.
 		resp, err := http.Get(url)
 		if err != nil {
 			Expect(err).NotTo(HaveOccurred())
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ install:
 
 # run all validation tests
 .PHONY: validate
-validate: gofmt check-vendor vet validate-vendor-licenses #lint
+validate: gofmt check-vendor vet validate-vendor-licenses sec #lint
 
 .PHONY: gofmt
 gofmt:
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,11 @@ lint:
 vet:
 	go vet $(PKGS)
 
+.PHONY: sec
+sec:
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
+	gosec -severity medium -confidence medium -exclude G304,G204 -quiet  ./...
+
 .PHONY: clean
 clean:
 	@rm -rf $(FILES)
@@ -67,6 +72,7 @@ goget-tools:
 	go get -u github.com/mitchellh/gox
 	go get github.com/frapposelli/wwhrd
 	go get -u github.com/onsi/ginkgo/ginkgo
+	go get -u github.com/securego/gosec/cmd/gosec
 
 # Run unit tests and collect coverage
 .PHONY: test-coverage
--- a/pkg/util/config_util.go
+++ b/pkg/util/config_util.go
@@ -16,7 +16,7 @@ import (
 func CreateIfNotExists(configFile string) error {
 	_, err := os.Stat(filepath.Dir(configFile))
 	if os.IsNotExist(err) {
-		err = os.MkdirAll(filepath.Dir(configFile), 0755)
+		err = os.MkdirAll(filepath.Dir(configFile), 0750)
 		if err != nil {
 			return errors.Wrap(err, "unable to create directory")
 		}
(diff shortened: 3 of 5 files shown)


mik-dass:
@JAORMX @kadel Should we run this check on CI?

JAORMX:
@mik-dass that would be good! There are some warnings that the tool throws out (nothing serious), so I wanted folks to try it out before starting to discuss those.

kadel:
@mik-dass, yes that will be next step.
But we should first try to fix existing errors.  
Some will be easy to fix: G301 and G302
G204 can be added to ignore list, as we actually want to do this.  Do you agree@JAORMX?
Not sure why G107 is a problem.



```
▶ make sec
/Users/tkral/Code/goenv/odo/bin/gosec
gosec -severity medium -confidence medium -exclude G304 -quiet  ./...
Results:


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/pkg/util/config_util.go:19] - G301: Expect directory permissions to be 0750 or less (Confidence: HIGH, Severity: MEDIUM)
  > os.MkdirAll(filepath.Dir(configFile), 0755)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/tests/helper/helper_http.go:24] - G107: Potential HTTP request made with variable url (Confidence: MEDIUM, Severity: MEDIUM)
  > http.Get(url)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/tests/helper/helper_filesystem.go:143] - G302: Expect file permissions to be 0600 or less (Confidence: HIGH, Severity: MEDIUM)
  > os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0644)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/tests/helper/helper_filesystem.go:47] - G301: Expect directory permissions to be 0750 or less (Confidence: HIGH, Severity: MEDIUM)
  > os.MkdirAll(dir, 0755)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/tests/integration/component.go:521] - G301: Expect directory permissions to be 0750 or less (Confidence: HIGH, Severity: MEDIUM)
  > os.Mkdir(context, 0755)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/tests/helper/helper_run.go:20] - G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command(program, args...)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/pkg/util/util.go:333] - G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command("open", url)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/pkg/util/util.go:331] - G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command("rundll32", "url.dll,FileProtocolHandler", url)


[/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/pkg/util/util.go:329] - G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command("xdg-open", url)


Summary:
   Files: 131
   Lines: 22333
   Nosec: 0
  Issues: 9

make: *** [sec] Error 1
```

JAORMX:
@kadel G107 tries to warn folks about doing http.Get calls with values that might be provided by the user. Since in some instances it could be used in malicious ways. In this case it can be safely ignored since this was for test code.

Perhaps we should limit the coverage of this tool to only check the "cmd" and "pkg" packages, instead of the current call, which attempts to find everything in the repo (except the vendor folder).

JAORMX:
@kadel @mik-dass should I address the warnings as part of this PR? or should that be done on another one?

mik-dass:
> @kadel @mik-dass should I address the warnings as part of this PR? or should that be done on another one?

@JAORMX  I would say let's fix them in this PR itself or at least make an issue for a future follow up. @kadel WDYT?

JAORMX:
@kadel @mik-dass I'll address the warnings as part of this PR

tests/helper/helper_http.go
```diff
@@ -21,6 +21,8 @@ func HttpWaitFor(url string, match string, maxRetry int, interval int) {
 	for i := 0; i < maxRetry; i++ {
 		fmt.Fprintf(GinkgoWriter, "try %d of %d\n", i, maxRetry)
 
+		// #nosec
+		// gosec:G107 -> This is safe since it's just used for testing.
```
amitkrout:
```
G107: Url provided to HTTP request as taint input
```
@JAORMX I think the the comment you added is not what it actually for. It is passing through the ```sec``` check because the ```http.Get(url)``` doesn't use url as an argument of a sub process call. 

So the comment does not make sense here.

JAORMX: Right; my intention with the comment was to suggest that we can safely ignore the warning because the call is just used for the test suite.

amitkrout: Understood, Please remove those two lines of comment to avoid the confusion
Thnaks

JAORMX: @amitkrout // #nosec is needed because otherwise, the gosec tool that I'm introducing as part of this PR will generate a warning. The line after it is explaining why it's safe to ignore that security warning.

amitkrout: Hmm, Good point.
Lets keep it then

PR #1959, opened 2019-07-29
Update labels to match gorkem/app-labels recommendations

### LINKED ISSUE #599: Adhere to new labels created for tool standarization in k8s

kadel:
This updates labels to match  https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc and also to match how those labels are used in OpenShift devconsole.
This is to ensure that components created by odo are correctly displayed in devconsole topology view.

**This breaks compatibility with older odo versions!** 
Release notes will need to include a warning about this.

Components that are already pushed to the cluster using older odo versions will not work. User needs to delete a component from the cluster and push it again.

fixes #599

### DIFF
--- a/pkg/url/labels/labels.go
+++ b/pkg/url/labels/labels.go
@@ -6,7 +6,7 @@ import (
 
 // URLLabel is the label key that is applied to all url resources
 // that are created
-const URLLabel = "app.kubernetes.io/url-name"
+const URLLabel = "app.openshift.io/vcs-uri-name"
 
 // GetLabels gets the labels to be applied to the given url besides the
 // component labels and application labels.
--- a/pkg/url/labels/labels.go
+++ b/pkg/url/labels/labels.go
@@ -6,7 +6,7 @@ import (
 
 // URLLabel is the label key that is applied to all url resources
 // that are created
-const URLLabel = "app.openshift.io/vcs-uri-name"
+const URLLabel = "odo.openshift.io/url-name"
 
 // GetLabels gets the labels to be applied to the given url besides the
 // component labels and application labels.
--- a/pkg/url/url_test.go
+++ b/pkg/url/url_test.go
@@ -43,9 +43,9 @@ func TestCreate(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "nodejs-app",
 					Labels: map[string]string{
-						"app.kubernetes.io/part-of":     "app",
-						"app.kubernetes.io/instance":    "nodejs",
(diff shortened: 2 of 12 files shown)


pkg/url/labels/labels.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // URLLabel is the label key that is applied to all url resources
 // that are created
-const URLLabel = "app.kubernetes.io/url-name"
+const URLLabel = "app.openshift.io/vcs-uri-name"
```
mik-dass: I don't see any label change for URL in https://github.com/openshift/odo/issues/599#issuecomment-488610781

pkg/url/url_test.go
```diff
@@ -43,9 +43,9 @@ func TestCreate(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "nodejs-app",
 					Labels: map[string]string{
-						"app.kubernetes.io/name":           "app",
-						"app.kubernetes.io/component-name": "nodejs",
-						"app.kubernetes.io/url-name":       "nodejs",
+						"app.kubernetes.io/part-of":     "app",
+						"app.kubernetes.io/instance":    "nodejs",
+						"app.openshift.io/vcs-uri-name": "nodejs",
```
mik-dass: same as above in this file

pkg/url/labels/labels.go
kadel: Good cache. 
This should not have been updated. This label is for odo internal use.

This should probably be something like `odo.openshift.io/url-name` to indicate that it is odo specific. I'll update it

kadel:
The current  view in the developer console.
Screenshot 2019-07-31 at 14 23 11


Not sure why backend has the OpenShift logo. It is Java component the DeploymentConfig has the following labels:
```
  labels:
    app: wildwest
    app.kubernetes.io/instance: backend
    app.kubernetes.io/name: java
    app.kubernetes.io/part-of: wildwest
    app.openshift.io/runtime-version: "8"
```

@sbose78  do you know why might that be?

sbose78:
@kadel I tried doing an `Import from Git` with the Java s2i image and I see the same. So we can ignore that.

kadel:
ping @cdrage

girishramnani:
looks good to me.
/lgtm

PR #1916, opened 2019-07-17
Adds `odo create project -o json` output

cdrage:
Adds machine readable output for odo create project.

To test:

```sh
odo create project -o json
```

PR #1898, opened 2019-07-10
Make --output a global parameter, disable logging, fix bug

cdrage:
This PR:
  - Makes `-o` or `--output` a global parameter
  - Disables logging completely when performing `-o json`, including
  using verbosity `-v`
  - Fixes a minor bug within some commands where `-o json` wasn't
  detected corretly due to a local flag being used `outputFlag` rather
  than `OutputFlag`

PR #1960, opened 2019-07-29
Add odo project delete -o json

cdrage:
This PR:

 - Adds `odo project delete -o json`
 - Builds upon the global parameter and `odo project create -o json`
 commits


cdrage:
/hold

This PR builds upon #1898 and #1916 and thus requires both to be merged before anything :)

pkg/odo/cli/project/delete.go
```diff
@@ -53,30 +54,53 @@ func (pdo *ProjectDeleteOptions) Complete(name string, cmd *cobra.Command, args
 func (pdo *ProjectDeleteOptions) Validate() (err error) {
 	// Validate existence of the project to be deleted
 	isValidProject, err := project.Exists(pdo.Context.Client, pdo.projectName)
-	if !isValidProject {
+	errorMessage := fmt.Sprintf("The project %s does not exist. Please check the list of projects using `odo project list`", pdo.projectName)
+
+	if log.IsJSON() && !isValidProject {
+		project.MachineReadableErrorOutput(pdo.projectName, errors.New(errorMessage))
+	} else if !isValidProject {
 		return fmt.Errorf("The project %s does not exist. Please check the list of projects using `odo project list`", pdo.projectName)
```
dharmit: Shouldn't we just reuse `errorMessage` here?

pkg/project/project.go
```diff
@@ -82,26 +87,81 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
...
 			Applications: apps,
 		},
 		Status: ProjectStatus{
-
 			Active: isActive,
 		},
 	}
 }
 
+// MachineReadableSuccessOutput ...
```
dharmit: Is the godoc empty on purpose or you missed adding it?

pkg/project/project.go
```diff
@@ -82,26 +87,81 @@ func Exists(client *occlient.Client, projectName string) (bool, error) {
...
+		},
+		Message: message,
+	}
+
+	printableOutput, err := json.Marshal(machineOutput)
+
+	// If we error out... there's no way to output it (since we disable logging when using -o json)
+	if err != nil {
+		fmt.Printf("Unable to unmarshal JSON: %s", err.Error())
+		os.Exit(1)
```
dharmit: Exit code for failure to create a project and for failure to generate the machine readable output is same. How would someone who is using the API determine what the exact cause of error?

tests/integration/json_test.go
```diff
@@ -34,6 +34,22 @@ var _ = Describe("odojsonoutput", func() {
...
+		})
+
+		// odo project create newprojectjson -o json (x2)
+		It("should fail saying that there is already an existing project", func() {
+			actual := helper.CmdShouldFail("odo", "project", "create", "newprojectjson", "-o", "json")
+			desired := `{"kind":"Error","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"newprojectjson","creationTimestamp":null},"message":"unable to create new project: unable to create new project newprojectjson: project.project.openshift.io \"newprojectjson\" already exists"}`
+			Expect(desired).Should(MatchJSON(actual))
+		})
+	})
+
```
dharmit: Similar tests for project deletion would also be helpful, no? One test that deletes an existing project and another where we try to delete a project that does not exist.

pkg/project/project.go
cdrage: Ideally this should never happen, but unfortunately we do hide all non-json related output. What would happen upstream is that VSCode will be unable to marshal the JSON output anyways (it will output "Unable to unmarshal json... from Odo..").

I honestly couldn't think of a better solution.

dharmit: Yes, I kind of assumed that the message (in this case `"Unable to unmarshal JSON`) would be a differentiator, but it still feels a bit skewed. I might sound pedantic here but if we have more such ambiguities w.r.t exit codes, it might come back to bite us since our APIs are going to be crucial for projects like Adapters. Correct me if I'm wrong.

amitkrout:
@cdrage Are these CI failure related to your change ? Can you please check

cdrage:
@amitkrout Most likely. This PR is still WIP / need to rebase again.

cdrage:
Too many changes / rebasing, I've opened up a new PR: https://github.com/openshift/odo/pull/2037

PR #1966, opened 2019-07-30
Add newline spacing to odo app --help

cdrage:
Adds newline spacing to `odo app --help` due to not having `\n\n` within
the help example (spacing was only 1 line)

```sh
Examples:
  # Delete the application
  odo app delete myapp

  # Describe 'webapp' application,
  odo app describe webapp

  # List all applications in the current project
  odo app list

  # List all applications in the specified project
  odo app list --project myproject
```

### DIFF
--- a/pkg/odo/cli/application/application.go
+++ b/pkg/odo/cli/application/application.go
@@ -32,7 +32,7 @@ func NewCmdApplication(name, fullName string) *cobra.Command {
 		Use:   name,
 		Short: "Perform application operations",
 		Long:  `Performs application operations related to your OpenShift project.`,
-		Example: fmt.Sprintf("%s\n%s\n%s",
+		Example: fmt.Sprintf("%s\n\n%s\n\n%s",
 			delete.Example,
 			describe.Example,
 			list.Example),


pkg/odo/cli/application/application.go
```diff
@@ -32,7 +32,7 @@ func NewCmdApplication(name, fullName string) *cobra.Command {
 		Use:   name,
 		Short: "Perform application operations",
 		Long:  `Performs application operations related to your OpenShift project.`,
-		Example: fmt.Sprintf("%s\n%s\n%s",
+		Example: fmt.Sprintf("%s\n\n%s\n\n%s",
```
amitkrout: The other command help like catalog, config and also rest of the command help follows the single spacing format. Even look and feel point of view single space in string looks like an ideal format.

TBH i am not in the favor of this change

amitkrout: > The other command help like [catalog](https://github.com/openshift/odo/blob/master/pkg/odo/cli/catalog/catalog.go#L26), [config](https://github.com/openshift/odo/blob/master/pkg/odo/cli/config/config.go#L30) and also rest of the command help doesn't follow two new line spacing spacing format. Even UX look and feel point of view single new line spacing looks an ideal format to me.
> 
> TBH this change doesn't make sense to me unless there is a hard requirement to do it cc_ @surajnarwade @mik-dass

Ahh... 
After testing the pr i can see the difference. I see the rest of cli command help, new line spacing is implemented in a bit different way i mean in the sub command file. 

I am fine with the current change.

cdrage: Yup. It's implemented differently. We'll have to refactor this later haha.

amitkrout:
ping @mik-dass

pkg/odo/cli/application/application.go
girishramnani: Can we please have an issue to track that @cdrage 🙂

PR #1982, opened 2019-08-05
resolved a lint error and handled some ignored errors in file_indexer

girishramnani:
resolved a lint error and handled some ignored errors in file_indexer

No issue

## How to test changes?
just check if the index file is created correctly

### DIFF
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -2,11 +2,12 @@ package util
 
 import (
 	"encoding/json"
-	"github.com/golang/glog"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/golang/glog"
 )
 
 const fileIndexDirectory = ".odo"
@@ -26,14 +27,10 @@ func read(filePath string) (map[string]FileData, error) {
 		return map[string]FileData{}, nil
 	}
 
-	jsonFileRead, err := os.Open(filePath)
-	// if we os.Open returns an error then handle it
+	byteValue, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		return nil, err
 	}
-	defer jsonFileRead.Close()
-
-	byteValue, _ := ioutil.ReadAll(jsonFileRead)
 	// unmarshals the byte values and fill up the file read map
 	err = json.Unmarshal(byteValue, &fileReadMap)
 	if err != nil {
@@ -126,7 +123,7 @@ func Run(directory string, ignoreRules []string) (filesChanged []string, filesDe
 	}
 
 	// find files which are deleted/renamed
-	for fileName, _ := range readData {
+	for fileName := range readData {
 		if _, ok := filesMap[fileName]; !ok {
 			glog.V(4).Infof("file deleted: %s", fileName)
 			filesDeleted = append(filesDeleted, fileName)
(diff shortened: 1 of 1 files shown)


amitkrout:
> just check if the index file is created correctly

@girishramnani Which index file and how ?

dharmit:
> > just check if the index file is created correctly
> 
> @girishramnani Which index file and how ?

+1. Have the same query

pkg/util/file_indexer.go
```diff
@@ -149,17 +146,9 @@ func Run(directory string, ignoreRules []string) (filesChanged []string, filesDe
 // filepath is the location of the file to which it is supposed to be written
 func write(filePath string, writeMap map[string]FileData) error {
 	jsonData, err := json.Marshal(writeMap)
```
girishramnani: two `err` one after the other, but error checked in the end. which means this error will be ignore

pkg/util/file_indexer.go
```diff
@@ -126,7 +123,7 @@ func Run(directory string, ignoreRules []string) (filesChanged []string, filesDe
 	}
 
 	// find files which are deleted/renamed
-	for fileName, _ := range readData {
```
girishramnani: lint: second arg doesn't need to be ignored. we can just eliminate it

pkg/util/file_indexer.go
```diff
@@ -26,14 +27,10 @@ func read(filePath string) (map[string]FileData, error) {
...
 
-	jsonFileRead, err := os.Open(filePath)
-	// if we os.Open returns an error then handle it
+	byteValue, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		return nil, err
 	}
-	defer jsonFileRead.Close()
-
-	byteValue, _ := ioutil.ReadAll(jsonFileRead)
```
girishramnani: fix: ignored the error here

girishramnani:
@amitkrout @dharmit `.odo/odo-file-index.json`, introducted as part of the smart odo push https://github.com/openshift/odo/pull/1849

girishramnani:
@amitkrout I think it would be good idea to add that as part of a separate PR. I can record this in an issue. But if you feel we need it as part of this PR then I can surely add it.

amitkrout:
> I think it would be good idea to add that as part of a separate PR.

Works for me. Please create a separate issue and assign to me. I will take care of that

/approve

pkg/util/file_indexer.go
```diff
@@ -26,14 +27,10 @@ func read(filePath string) (map[string]FileData, error) {
 		return map[string]FileData{}, nil
 	}
 
-	jsonFileRead, err := os.Open(filePath)
-	// if we os.Open returns an error then handle it
+	byteValue, err := ioutil.ReadFile(filePath)
```
mik-dass: please add some comments explaining the benefit of this approach as compared to the earlier one

girishramnani: This is just DRYing the code

mik-dass: But some comments will benefit in the future

girishramnani: there is nothing different or better about this approach other than that its less code then before :sweat_smile:

girishramnani: and at other places like this https://github.com/openshift/odo/pull/1982#discussion_r310565470. I have provided the reasoning

Issue #822, opened 2018-10-09
Provide debug/port-forward capabilities for the deployment


jorgemoralespou:
odo components should enable whatever debug capabilities exist in the image and when the component is configured as local/binary, these capabilities should be enabled. odo should also provide a way to express how to connect to the remote container via port-forward.

As the connection cannot always be opened, even if the component is activating a port-forward by default, there should be a mechanism to re establish the port forwarding capabilities.

I can see various options:
- Always enable debug capabilities by default, and provide a way to specify the local port to forward the remote port via a flag
- Provide a flag to enable debug mode and open the port forwarding
- Provide a specific command for enabling debug mode and open the port forward. This requires a redeployment, hence I would not prefer this option.
- Provide a mechanism in the supervisor to activate the debug mode.

I would love to gather some ideas on how debugging should work.

gorkem:
I see this as 3 steps. 

1. Enabling the debug mode on the runtime(JavaVM, node.js etc). This usually ends up with a port opened for debuggers to attach. Different runtimes have different behaviour here. For instance node.js allows opening the debug port while it is running while JVM requires a restart. I feel that it is OK if odo start these runtimes with debug always on. It makes the debugging more agile for things like JVM compared to restarting them all the time.

2. The second step is port forwarding. In this case debugger just needs to tell `odo` which local port it will use and `odo` and `odo` should setup the port forwarding from local port to the container's debug port. 

3. This is where debugger connects to local port and profit!! We will probably write special launch configurations to vscode and Eclipse IDE to invoke `odo` to do this but these are outside the scope of `odo`.

Preeticp:
@boczkowska this is going to affect docs for the 4.2 release, please add a parallel issue to our docs board.
@kadel @sspeiche:
1. We need to have a label "Needs Docs" to identify issues which need docs. So far we tracked issues needing docs using Type/Docs or Kind/Docs but those would be inappropriate for these kind of issues.
2. Also, though I am a member of the OpenShift org I cannot add the above labels, neither create a new one. This makes it difficult for docs team to track such issues. Please let me know if I could get the rights to assign labels and assignees so that the docs team is able to assign issues and labels. Thank you.

kadel:
I can see two paths here:
1.  as @gorkem and @jorgemoralespou described we could enable debug mode by default
2. Enable it only when user requests. This will probably require restarting the application for most of the languages.


[...]

PR #1995, opened 2019-08-07
Proposal document: odo debug

kadel:
Proposal for `odo debug` related to #822

### DIFF
--- a/docs/proposals/odo-debug.md
+++ b/docs/proposals/odo-debug.md
@@ -1,0 +1,82 @@
+# odo debug support proposal
+
+Table of Contents:
+- [odo debug support proposal](#odo-debug-support-proposal)
+  - [Design overview](#design-overview)
+    - [New odo commands](#new-odo-commands)
+      - [`odo debug start`](#odo-debug-start)
+      - [`odo debug stop`](#odo-debug-stop)
+      - [`odo debug status`](#odo-debug-status)
+    - [`odo debug port-forward LOCAL_PORT:REMOTE_PORT`](#odo-debug-port-forward-localportremoteport)
+- [Languages support](#languages-support)
+  - [NodeJS](#nodejs)
+    - [Problem 1](#problem-1)
+      - [Solution 1](#solution-1)
+      - [Solution 2](#solution-2)
+  - [Java](#java)
+
+## Design overview
+
+The debug mode inside the container is controlled by `DEBUG_MODE` and `DEBUG_PORT` env variable.
+- if `DEBUG_MODE == true` than the code should be executed with remote debugging capabilities enabled.
+- `DEBUG_PORT` is optional and controls the port of the remote debugger.
+
+Odo sets up port-forwarding to allow a debugger to connect to the running process. 
+Debugger client will always connect to `localhost:$DEBUG_PORT`
+### New odo commands
+#### `odo debug start`
+Enables debug mode for the component. It does this by adding `DEBUG_MODE = true` env variable.
+
+After adding env variables the component (Deployment Config) is restarted.
+Once the pod is in running state `odo debug start` executes `odo debug port-forward $DEBUG_PORT:$DEBUG_PORT`.
+This process executed on the background and PID of the process is recorded in `~/.odo/port-forward.pid` file.
+
+
+Flags:
+- `--port`: Optional. Controls value of `DEBUG_PORT` env variable
+- `--context`: Optional. The same as in other commands (see `odo push --context` for example)
+- `--no-port-forward`: Optional. Just enable debugging but don't start forwarding port. User can use `odo debug port-forward` manually or use other mechanisms.
+
+#### `odo debug stop`
+Stop port-forward process if running and remove `DEBUG_PORT` and `DEBUG_MODE` env variables from Deployment Config.
+
+First, terminate and cleanup port-forwarding.
+If `~/.odo/port-forward.pid` exists and process with PID is running terminate the process and remove the file.
+If the process is not running just remove the file (warn the user).
+If the file doesn't exist it is ok, as `odo debug start --no-port-forward` might have been used.
+
+Second, remove `DEBUG_MODE` and `DEBUG_PORT` env variables from Deployment Config and wait for the pod to be the ready state after a restart.
+
+
+#### `odo debug status`
+Informs the user if the debug mode is enabled for the component and if the port-forward process is running or not.
+
+
+### `odo debug port-forward LOCAL_PORT:REMOTE_PORT`
+Forwards `LOCAL_PORT` to container port `REMOTE_PORT`.
+This blocks the terminal. It will be used for `odo debug start`
+As long as this command is running the port forward is enable (same as `oc port-forward`).
+
+
+
+# Languages support
+
+## NodeJS
+
+To run the NodeJS application in the debug mode it has to start with `--inspect` (`node --inspect [host:port] script.js`) flag.
+
+### Problem 1 
+The command requires the entry point file. So odo has to know what is the `.js` file.
+#### Solution 1
+We can assume some default like `server.js`, and allow users to overwrite this using env variable.
+#### Solution 2
+User will provide their own debug command in `package.json`, the same way s2i builder image already assumes that there is a `start` script in `package.json`
+```
+ "scripts": {
+ "start": "node server.js",
+ "debug": "node --inspect server.js"
+ }
+```
+
+## Java
+
--- a/docs/proposals/odo-debug.md
+++ b/docs/proposals/odo-debug.md
@@ -1,10 +1,14 @@
 # odo debug support proposal
+odo debug command is meant for debugging code running as odo component.
+First implementation in odo will just support Java and NodeJS components.
+
 
 Table of Contents:
 - [odo debug support proposal](#odo-debug-support-proposal)
   - [Design overview](#design-overview)
     - [New odo commands](#new-odo-commands)
       - [`odo debug start`](#odo-debug-start)
+    - [port-forward.json format](#port-forwardjson-format)
       - [`odo debug stop`](#odo-debug-stop)
       - [`odo debug status`](#odo-debug-status)
     - [`odo debug port-forward LOCAL_PORT:REMOTE_PORT`](#odo-debug-port-forward-localportremoteport)
@@ -29,7 +33,21 @@ Enables debug mode for the component. It does this by adding `DEBUG_MODE = true`
 
 After adding environment variables the component (Deployment Config) is restarted.
 Once the pod is in running state `odo debug start` executes `odo debug port-forward $DEBUG_PORT:$DEBUG_PORT`.
-This process is executed in the background and PID of the process is recorded in `~/.odo/port-forward.pid` file.
+This process is executed in the background and PID of the process is recorded in `~/.odo/port-forward.json` file.
+
+### port-forward.json format
+```json
+[
+  {
+     "directory": "/path/to/component/",
+     "port-forward-pid": 344
+  },
+  {
+     "directory": "/path/to/another/component/",
+     "port-forward-pid": 332
+  }
+]
+```
 
 
 Flags:
@@ -41,7 +59,7 @@ Flags:
 Stop port-forward process if running and remove `DEBUG_PORT` and `DEBUG_MODE` environment variables from Deployment Config.
 
 First, terminate and cleanup port-forwarding.
-If `~/.odo/port-forward.pid` exists and process with PID is running terminate the process and remove the file.
+If `~/.odo/port-forward.json` exists and process with PID is running terminate the process and remove the file.
 If the process is not running just remove the file (warn the user).
 If the file doesn't exist it is ok, as `odo debug start --no-port-forward` might have been used.
 
--- a/docs/proposals/odo-debug.md
+++ b/docs/proposals/odo-debug.md
@@ -6,10 +6,14 @@ First implementation in odo will just support Java and NodeJS components.
 Table of Contents:
 - [odo debug support proposal](#odo-debug-support-proposal)
   - [Design overview](#design-overview)
-    - [New odo commands](#new-odo-commands)
-      - [`odo debug info`](#odo-debug-info)
-    - [`odo debug port-forward LOCAL_PORT:REMOTE_PORT`](#odo-debug-port-forward-localportremoteport)
-      - [Open Question](#open-question)
+  - [new `DebugPort` component setting](#new-debugport-component-setting)
+  - [New odo commands](#new-odo-commands)
+    - [`odo debug port-forward`](#odo-debug-port-forward)
+      - [`--local-port` flag](#local-port-flag)
+      - [Examples](#examples)
+        - [The `DebugPort` is set in the LocalConfig:](#the-debugport-is-set-in-the-localconfig)
+        - [If the `DebugPort` is NOT set in the LocalConfig:](#if-the-debugport-is-not-set-in-the-localconfig)
+    - [`odo debug info`](#odo-debug-info)
 - [Languages support](#languages-support)
   - [NodeJS](#nodejs)
   - [Java](#java)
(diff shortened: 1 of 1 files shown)


docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+Enables debug mode for the component. It does this by adding `DEBUG_MODE = true` env variable.
+
+After adding env variables the component (Deployment Config) is restarted.
+Once the pod is in running state `odo debug start` executes `odo debug port-forward $DEBUG_PORT:$DEBUG_PORT`.
+This process executed on the background and PID of the process is recorded in `~/.odo/port-forward.pid` file.
+
+
+Flags:
+- `--port`: Optional. Controls value of `DEBUG_PORT` env variable
+- `--context`: Optional. The same as in other commands (see `odo push --context` for example)
```
dharmit: I assume this flag would help one debug a component from outside the component directory. Takes me back to the `~/.odo/port-forward.pid` vs. `/path/to/component/.odo/port-forward.pid` question because if my assumption is correct, this means that `odo` could start multiple components in debug mode if the developer wishes so.

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+- `--no-port-forward`: Optional. Just enable debugging but don't start forwarding port. User can use `odo debug port-forward` manually or use other mechanisms.
+
+#### `odo debug stop`
+Stop port-forward process if running and remove `DEBUG_PORT` and `DEBUG_MODE` env variables from Deployment Config.
+
+First, terminate and cleanup port-forwarding.
+If `~/.odo/port-forward.pid` exists and process with PID is running terminate the process and remove the file.
+If the process is not running just remove the file (warn the user).
+If the file doesn't exist it is ok, as `odo debug start --no-port-forward` might have been used.
```
dharmit: I guess I'm asking same question in a different way: What if `odo debug` was started with port-forwarding but the file failed to create for some reason?

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+
+Second, remove `DEBUG_MODE` and `DEBUG_PORT` env variables from Deployment Config and wait for the pod to be the ready state after a restart.
```
dharmit: `s/to be the ready/to be in ready/`

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+#### `odo debug status`
+Informs the user if the debug mode is enabled for the component and if the port-forward process is running or not.
+
+
+### `odo debug port-forward LOCAL_PORT:REMOTE_PORT`
+Forwards `LOCAL_PORT` to container port `REMOTE_PORT`.
+This blocks the terminal. It will be used for `odo debug start`
```
dharmit: Not sure I understand this. When one executes `odo debug port-forward 8080:8080`, it will block the terminal and doing a `Ctrl+C` would stop the port-forwarding. But what does it mean when we say `It will be used for "odo debug start"`. Does it mean that with `odo debug port-forward 8080:8080` done, I can open a different terminal and do `odo debug start` and it will pick-up that localhost's port 8080 is mapped to components 8080 port?

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+- [Languages support](#languages-support)
+  - [NodeJS](#nodejs)
+    - [Problem 1](#problem-1)
+      - [Solution 1](#solution-1)
+      - [Solution 2](#solution-2)
+  - [Java](#java)
+
+## Design overview
+
+The debug mode inside the container is controlled by `DEBUG_MODE` and `DEBUG_PORT` env variable.
```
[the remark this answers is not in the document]
dharmit: Actually, `s/env/environment` needs to be done at quite a few places across the file. :)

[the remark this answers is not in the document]
kadel: fixed

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
+# odo debug support proposal
+
+Table of Contents:
+- [odo debug support proposal](#odo-debug-support-proposal)
```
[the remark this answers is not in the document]
kadel: Not sure what you mean :-( 
Something like general reasoning why we need this command?

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+    - [Problem 1](#problem-1)
+      - [Solution 1](#solution-1)
+      - [Solution 2](#solution-2)
+  - [Java](#java)
+
+## Design overview
+
+The debug mode inside the container is controlled by `DEBUG_MODE` and `DEBUG_PORT` env variable.
+- if `DEBUG_MODE == true` than the code should be executed with remote debugging capabilities enabled.
+- `DEBUG_PORT` is optional and controls the port of the remote debugger.
```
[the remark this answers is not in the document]
kadel: nope, the application will still start on port 8080, or how it is configured.

`DEBUG_PORT` controls additional port that will be opened for the debugger connections


for nodejs the run command will look something like this:
```
nodejs --inspect ${DEBUG_PORT} server.js
```
for java
```
java -Xdebug -Xrunjdwp:transport=dt_socket,address=${DEBUG_PORT},server=y,suspend=y
```

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,82 @@
...
+
+Odo sets up port-forwarding to allow a debugger to connect to the running process. 
+Debugger client will always connect to `localhost:$DEBUG_PORT`
+### New odo commands
+#### `odo debug start`
+Enables debug mode for the component. It does this by adding `DEBUG_MODE = true` env variable.
+
+After adding env variables the component (Deployment Config) is restarted.
+Once the pod is in running state `odo debug start` executes `odo debug port-forward $DEBUG_PORT:$DEBUG_PORT`.
+This process executed on the background and PID of the process is recorded in `~/.odo/port-forward.pid` file.
```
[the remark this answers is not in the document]
kadel: > Could the creation of `~/.odo/port-forward.pid` fail under any circumstances?
> * If yes, what would `odo` do?

fail



> Is this information really going to be stored in `~/.odo/port-forward.pid` file (global configuration directory) or in a component directory `/path/to/component/.odo/port-forward.pid`?
> * If it's going to be in `~/.odo/port-forward.pid` indeed, wouldn't that limit the developer to debug only one app at a time? What if they want to start another component in debug mode?

good question. Yes, it would limit it to debugging just one application at a time.
Problem with storing it in components local .odo dir is that users will tend to commit this into the source code :-( And this is not something that should be persistent, it is runtime information.

Maybe storing pids in the global configuration in json format?
```json
[
  {
     "directory": "/path/to/component/",
     "port-forward-pid": 344
  },
  {
     "directory": "/path/to/another/component/",
     "port-forward-pid": 332
  }
]
```

docs/proposals/odo-debug.md
kadel: already replied in above

docs/proposals/odo-debug.md
kadel: https://github.com/openshift/odo/pull/1995#discussion_r312413621

docs/proposals/odo-debug.md
kadel: fixed

docs/proposals/odo-debug.md
kadel: The sentence is wrong 😇  It is meant to be "It will be used *in* `odo debug start`".
What I mean by that that the `start` command will actually make use of `port-forward` command. It will execute it as a separate background process and record its pid. `stop` command will then kill `port-forward` process based on recorded pid.

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: Yes, something like what it's meant to do and what it's not meant to do.

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: OK so besides running the app on 8080, port 9090 in the container will also be exposed and listening for some kind of connections. And the debug client (IDE in case of Adapters) will connect to this 9090 port. Is that understanding correct?

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: Makes sense to me. Thanks for the clarification. :+1:

docs/proposals/odo-debug.md
dharmit: I'm still confused, I guess.  `odo debug port-forward LP:RP` won't be used by a user but by `odo debug start`. What if a user executes it on the terminal? Does a user need to use it through a terminal? If no, should we have a separate command `odo debug port-forward LP:RP` or have that logic in `odo debug start` itself?

I could be asking absolutely dumb stuff because I'm confused. :laughing:

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: For the first question: so we intend to have the operation of enabling port-forwarding and creation/modification of the file as an atomic action, right?

docs/proposals/odo-debug.md
kadel: > I'm still confused, I guess. `odo debug port-forward LP:RP` won't be used by a user but by `odo debug start`.

yes, a normal user won't use `port-forward` command.

>  What if a user executes it on the terminal? Does a user need to use it through a terminal?
 
Doesn't need to, and most likely won't use this command thought terminal. But there might be corner cases where it can be handy to allow it.

>If no, should we have a separate command odo debug port-forward LP:RP or have that logic in odo debug start itself?

`odo debug start` shouldn't block terminal. It will need to either spaw new process or bring itself to the background. Spawning new process seems an easier and more reliable solution to me, especially when we have to deal with windows. 

> I could be asking absolutely dumb stuff because I'm confused. 😆

You are asking the right questions. I should have included more info on why I thought about this design.

I had two use cases in mind:

**standard use** - local and remote ports are the same
this is what I expect will be the most common, 99.9% will use just this one command.
```
odo debug start --port 9090
```

**advance use** - local and remote ports are different (local 9999 forwarded to remote 9090)
this is a little bit more complicated (two commands are required), but most users won't ever need this
```
odo debug start --port 9090 --no-port-forward
odo debug port-forward 9999:9090
```

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
kadel: Allow remote debugging of the java and nodejs components.

^ Something like this?

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
kadel: Yes, that is correct.

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
kadel: Yes. from the outside ti should look like atomic operation.
If any of the file operations fails, the `port-forward` process needs to be killed, and error reported.


Implementation `odo debug start` could be something like this:

- check if `port-forward-pids.json` has a pid record for the component
- if yes and the process with given PID exists, stop and show a message that the port forwarding is already running
- if yes and the process is not running, show a warning that the process might fail and this will execute the new one
- set environment variables
- wait for deployment
- execute `odo debug port-forward`, get its pid
- record pid to the port-forward-pids.json (create file if doesn't exist)
- if any of the file operations fails, kill `odo debug port-forward` process and error out.

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: I was suggesting more along the lines of clarifying that `odo debug` is meant for debugging the components being deployed through `odo` and not debugging `odo` itself. For the latter, running `odo` with `-v` flag is the right choice. I was suggesting this because `debug` and `verbose` are often used interchangeably in practice and it could cause confusion to the end user.

> Allow remote debugging of the java and nodejs components.

Yeah. And a list of components that have debug support and are planned to be added, maybe. 

Does that make sense?

docs/proposals/odo-debug.md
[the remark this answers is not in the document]
dharmit: @kadel this sound awesome to me! Thanks for the detailed write-up. It's going to be pretty easy for whoever picks up this piece to code.

docs/proposals/odo-debug.md
dharmit: > > What if a user executes it on the terminal? Does a user need to use it through a terminal?
> 
> Doesn't need to, and most likely won't use this command thought terminal. But there might be corner cases where it can be handy to allow it.

Agree with corner use cases point. I can't think of one right away but that's the definition of a corner use-case. :joy:

> > If no, should we have a separate command odo debug port-forward LP:RP or have that logic in odo debug start itself?
> 
> `odo debug start` shouldn't block terminal. It will need to either spaw new process or bring itself to the background. Spawning new process seems an easier and more reliable solution to me, especially when we have to deal with windows.

Not a technical question: And having a unified approach for all platforms would be a good idea for maintainability, right?  Asking for my understanding.

> I had two use cases in mind:
> 
> **standard use** - local and remote ports are the same
> this is what I expect will be the most common, 99.9% will use just this one command.
> 
> ```
> odo debug start --port 9090
> ```

So if a user only does `odo debug start --port 9090`, we will automagically map localhost's port 9090 to container's port 9090 on which the deployed component is listening? I think we should document this default behaviour. I couldn't find it in this proposal. I think it was sort of assumed in this statement of proposal: 
```
Once the pod is in running state `odo debug start` executes `odo debug port-forward $DEBUG_PORT:$DEBUG_PORT`.
```

> **advance use** - local and remote ports are different (local 9999 forwarded to remote 9090)
> this is a little bit more complicated (two commands are required), but most users won't ever need this
> 
> ```
> odo debug start --port 9090 --no-port-forward
> odo debug port-forward 9999:9090
> ```
:+1:

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,100 @@
...
+  - [Java](#java)
+
+## Design overview
+
+The debug mode inside the container is controlled by `DEBUG_MODE` and `DEBUG_PORT` environment variables.
+- if `DEBUG_MODE == true` than the code should be executed with remote debugging capabilities enabled.
+- `DEBUG_PORT` is optional and controls the port of the remote debugger.
+
+Odo sets up port-forwarding to allow a debugger to connect to the running process. 
+Debugger client will always connect to `localhost:$DEBUG_PORT`
```
cdrage: I'm assuming we'll use a const / default port for odo communication or is this based on each language / image?

docs/proposals/odo-debug.md
```diff
@@ -0,0 +1,100 @@
...
+#### Solution 2
+User will provide their own debug command in `package.json`, the same way s2i builder image already assumes that there is a `start` script in `package.json`
+```
+ "scripts": {
+ "start": "node server.js",
+ "debug": "node --inspect server.js"
+ }
+```
+
+## Java
```
girishramnani: for java we would probably need to run the jvm with `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n` ( this works for oracle java not sure about openjdk )

girishramnani: I was able to find docs for jdb till java 8 not sure if its support after that or not https://linux.die.net/man/1/jdb-java-1.7.0-openjdk

docs/proposals/odo-debug.md
kadel: yes, we will use some default port

jorgemoralespou:
Why debug is not the default mode when running odo from local or binary? Why would I need to restart my pod to connect my debugger to it, if I'm developing it will most likely be one of the things I want to do.
Is there a good reason not to have this the default, And let the `debug` command be the one that opens the port-forward as only action? This would simplify the workflow of odo.

Quarkus, when run with `mvn quarkus:dev` already starts in debugging mode. Why? Because it's the development phase of a Quarkus app, the one that the developer will interact with. Once he's done, he'll build the native app, and hence no debugging will be started by default. This is smart and this is the right thing to do when developing.

odo should take a similar approach, IMHO.

kadel:
> Why debug is not the default mode when running odo from local or binary? Why would I need to restart my pod to connect my debugger to it, if I'm developing it will most likely be one of the things I want to do.
> Is there a good reason not to have this the default, And let the `debug` command be the one that opens the port-forward as only action? This would simplify the workflow of odo.
> 
> Quarkus, when run with `mvn quarkus:dev` already starts in debugging mode. Why? Because it's the development phase of a Quarkus app, the one that the developer will interact with. Once he's done, he'll build the native app, and hence no debugging will be started by default. This is smart and this is the right thing to do when developing. [Blog on Quarkus debugging with CRStudio](https://developers.redhat.com/blog/2019/05/09/create-your-first-quarkus-project-with-eclipse-ide-red-hat-codeready-studio/)
> 
> odo should take a similar approach, IMHO.
> 
> You can ask developers of EAP, QUarkus, Springboot, node.js, ... for their opinion.

I wanted to do it also like this at first, but then I found some issues with nodejs.

It will put additional requirements. User will have to have proper `debug` script in `packages.json` or will have to provide additional arguments to specify the main `.js` file.

https://github.com/openshift/odo/blob/6f54061853ed8384c7617d5fcd72f8b896b495f7/docs/proposals/odo-debug.md#problem-1   

This will affect even users that won't use the debug.

jorgemoralespou:
I don't really get it, since the nodeshift images have a DEV_MODE that does run in debug by default (which is the obvious thing to do) and not sure there's any special requirement.
https://github.com/nodeshift/centos7-s2i-nodejs/blob/master/s2i/run#L28

Maybe the nodejs experts (@lholmquist, @lance) can help. Do you guys know of a reason why odo should not start in DEV_MODE style for node.js applications? Any implication on the user side's application/code so that `--inspect` can work?

kadel:
> I don't really get it, since the nodeshift images have a DEV_MODE that does run in debug by default (which is the obvious thing to do) and not sure there's any special requirement.
> https://github.com/nodeshift/centos7-s2i-nodejs/blob/master/s2i/run#L28

even DEV_MODE there assumes that the main file is called `server.js` if the project uses something else it won't work.

> 
> Maybe the nodejs experts (@lholmquist, @lance) can help. Do you guys know of a reason why odo should not start in DEV_MODE style for node.js applications? Any implication on the user side's application/code so that `--inspect` can work?

lholmquist:
> even DEV_MODE there assumes that the main file is called server.js if the project uses something else it won't work.

that is not necessarily true.  If you look at this starter app,  https://github.com/nodeshift-starters/nodejs-rest-http, there is no server.js and this works using the DEV_MODE env var.  It all depends on what you specify as your main in your package.json

> Maybe the nodejs experts (@lholmquist, @lance) can help. Do you guys know of a reason why odo should not start in DEV_MODE style for node.js applications? Any implication on the user side's application/code so that --inspect can work?

For the nodeshift images,  we are using nodemon in DEV_MODE to start the node.js process which allows the node process to restart if you make changes to a file in the Pod.  We use `npx` so we don't need to have nodemon installed prior

For the `--inspect` flag to properly work with nodemon, we did need to have a main file property in our package.json, which is recommended anyway, and is created as part of an `npm init`, so i don't see that being an issue.  i don't believe you even need to have a start script for this to work

This is the GH issue were we talked about making sure nodemon knew how to use `--inspect` https://github.com/nodeshift/centos7-s2i-nodejs/issues/33

Just for completeness, i also wrote a blog post about debugging with chrome dev-tools:
https://developers.redhat.com/blog/2018/05/15/debug-your-node-js-application-on-openshift-with-chrome-devtools/

Hopefully some of those words i wrote can help you guys out :)

kadel:
> > even DEV_MODE there assumes that the main file is called server.js if the project uses something else it won't work.
> 
> that is not necessarily true. If you look at this starter app, https://github.com/nodeshift-starters/nodejs-rest-http, there is no server.js and this works using the DEV_MODE env var. It all depends on what you specify as your main in your package.json
> 
> > Maybe the nodejs experts (@lholmquist, @lance) can help. Do you guys know of a reason why odo should not start in DEV_MODE style for node.js applications? Any implication on the user side's application/code so that --inspect can work?
> 
> For the nodeshift images, we are using nodemon in DEV_MODE to start the node.js process which allows the node process to restart if you make changes to a file in the Pod. We use `npx` so we don't need to have nodemon installed prior

One issue that I found when using `npx nodemon` is that it takes a looooog time to start even when it is already installed. 5 seconds is not an exception, it looks like that `npx` is checking something each time it is executed. Is there a way how to speed this up?

> 
> For the `--inspect` flag to properly work with nodemon, we did need to have a main file property in our package.json, which is recommended anyway, and is created as part of an `npm init`, so i don't see that being an issue. i don't believe you even need to have a start script for this to work
> 

That is nice, so it means that we can just execute `nodemon --inspect` inside the source directory without worrying about anything else.


> This is the GH issue were we talked about making sure nodemon knew how to use `--inspect` [nodeshift/centos7-s2i-nodejs#33](https://github.com/nodeshift/centos7-s2i-nodejs/issues/33)
> 
> Just for completeness, i also wrote a blog post about debugging with chrome dev-tools:
> https://developers.redhat.com/blog/2018/05/15/debug-your-node-js-application-on-openshift-with-chrome-devtools/
> 
> Hopefully some of those words i wrote can help you guys out :)

It helps a lot. Thank you!

lholmquist:
the thing with npx is that it will check to see if you have the command(nodemon, in this case) installed already,  if not, it goes and gets it and stores it temporarily.  I think if the users app also had nodemon as a dev dependcy, then it would be faster.

girishramnani:
we should have a global preference to disable default debug mode in case there are failures

girishramnani:
After a good amount of discussion with @kadel this is final proposal for debug -

- we will only have have one command in the debug context and that would be `odo debug port-forward`. 
- `odo debug start` and `odo debug stop` wouldn't be implemented because that has a dis-advantage of developer needing to re-deploy the app with dev-mode ( by which time they might not be able to reproduce the error or well it takes time )
- debug mode will always be on which means that developer just needs to port-forward the debug port using the `odo debug port-forward` command.
- in case there is an issue with debug mode - we will provide a global preference called `DebugMode` which will be set to true and developer can set it to `false` and redeploy the component to disable debugging. ( this can be considered a fail safe in case debug mode breaks )

There are multiple advantage of this approach over having `odo debug start & stop` which explicitly allow you start and stop. some of the advantages -

1 - `odo debug start` felt like an alias for `set dev_MODE and then odo create` essentially an extra entrypoint for the developer which feels confusing
2 - reduced complexity
3 - simpler UX

girishramnani:
> > * debug mode will always be `on`, which means that developer just needs to port-forward the debug port using the `odo debug port-forward` command.
> > * in case there is an issue with debug mode - we will provide a global preference called `DebugMode` which will be set to true and developer can set it to `false` and redeploy the component to disable debugging. ( this can be considered a fail safe in case debug mode breaks )
> 
> What does "debug mode will always be `on`" mean? When you create the component for inner loop you'll set the debug mode at that point?
It will always be set which means that the supervisord image will run the component with debug flags from the get go. 
> I guess this assumes that an existing running deployment, can't be port forwarded until someone puts in debug mode and . what does that experience look like?
This was the case with `odo debug start and stop` in the new proposal, if the dev finds an issue they just do `odo debug port-forward` and get the debugger attached locally, no need to re-deploy

girishramnani:
@dgolovin please check this amendment to the proposal https://github.com/openshift/odo/pull/1995#issuecomment-525259554

jorgemoralespou:
> I guess this assumes that an existing running deployment, can't be port forwarded until someone puts in debug mode and . what does that experience look like?

An existing deployment will not have been started with the `DEBUG` flags (in case of Java) or `DEV_MODE` in case of node.js, or.... so it will not be debugabble in any way. 
Currently any application that is created via `oc new-app` or the likes will not have been enabled for debugging, so it shouldn't really make a difference.

odo just assumes that the application it `manages` are created for the inner loop (specially if they are from binary or local).

girishramnani:
if we are only going to have `port-forward` then do we need to call it via `odo debug port-forward`?
Instead why not `odo port-forward`?

jorgemoralespou:
There will not only be port-forward. As discussed on our call with @kadel on Tuesday, there is probably also room for `odo debug info` that will give you information about the debug port configured, whether debug is enabled or not (as there should be a preference), etc...
Having a namespace gives you room for extension, and also makes clear that port-forward is meant for debugging as by default will port-forward to the debug port and not the application listening port.

kadel:
@jorgemoralespou @girishramnani @dharmit I've updated the proposal to match the latest discussion. Can you please have another look?

PR #1997, opened 2019-08-07
Adds "odo service create --context" feature

### LINKED ISSUE #1872: Missing context flag in service create command
[kind/bug]

- Operating System: All supported platform

```
$ odo service -h
Perform service catalog operations

Usage:
  odo service [flags]
  odo service [command]

Examples:
  # Create new postgresql service from service catalog using dev plan and name my-postgresql-db.
  odo service create dh-postgresql-apb my-postgresql-db --plan dev -p postgresql_user=luke -p postgresql_password=secret

  odo service list

Available Commands:
  create      Create a new service from service catalog using the plan defined and deploy it on OpenShift.
  delete      Delete an existing service
  list        List all services in the current application

Flags:
  -h, --help   Help for service

Global Flags:
      --skip-connection-check   Skip cluster check
  -v, --v Level                 Log level for V logs. Level varies from 0 to 9 (default 0).
      --vmodule moduleSpec      Comma-separated list of pattern=N settings for file-filtered logging

Use "odo service [command] --help" for more information about a command.
```
## Actual behavior
flag support is not available

## Expected behavior
Support should be added

dharmit:

This PR enables the users to create a service with `odo service create --context /path/to/component/dir` command. This allows service creation from anywhere in the system.

fixes #1872 

## How to test changes?

- Build the binary
- Using [nodejs-ex](https://github.com/sclorg/nodejs-ex) as an example, perform following:
  ```sh
  $ git clone https://github.com/sclorg/nodejs-ex /tmp/nodejs-ex
  $ cd /tmp/nodejs-ex
  $ odo create nodejs
  $ cd # this takes you to home dir
  $ odo service create dh-postgresql-apb --plan dev -p postgresql_user=luke -p postgresql_password=secret -p postgresql_database=my_data -p postgresql_version=9.6 --context /tmp/nodejs-ex

  # Use the --context from /tmp/nodejs-ex/ to list the service
  $ odo service list --context /tmp/nodejs-ex
  
  # Use the --context from /tmp/nodejs-ex/ to delete the service
  $ odo service delete --context /tmp/nodejs-ex
  
  # Bonus points for testing with --app and --project flag along with --context
  # I think the project should exist on the cluster. If app doesn't exist, odo will create it.
  $ odo service create dh-prometheus-apb --plan ephemeral --context /tmp/nodejs-ex --app myapp --project myproject
  ```

### DIFF
--- a/tests/integration/servicecatalog/service_test.go
+++ b/tests/integration/servicecatalog/service_test.go
@@ -17,6 +17,7 @@ var _ = Describe("odoServiceE2e", func() {
 	var project string
 	var context string
 	var app string
+	var serviceName string
 
 	//  current directory and project (before eny test is run) so it can restored  after all testing is done
 	var originalDir string
@@ -107,6 +108,7 @@ var _ = Describe("odoServiceE2e", func() {
 			project = helper.CreateRandProject()
 			context = helper.CreateNewContext()
 			app = helper.RandString(7)
+			serviceName = "odo-postgres-service"
 			originalDir = helper.Getwd()
 			helper.Chdir(context)
 		})
@@ -116,24 +118,30 @@ var _ = Describe("odoServiceE2e", func() {
 			helper.Chdir(originalDir)
 		})
 
-		It("should be able to create service using a given value for --context", func() {
+		It("should be able to create, list and delete a service using a given value for --context", func() {
 			// create a component by copying the example
 			helper.CopyExample(filepath.Join("source", "nodejs"), context)
 			helper.CmdShouldPass("odo", "create", "nodejs", "--app", app, "--project", project)
 
 			// cd to the originalDir to create service using --context
 			helper.Chdir(originalDir)
-			helper.CmdShouldPass("odo", "service", "create", "dh-prometheus-apb", "--plan", "ephemeral",
+			helper.CmdShouldPass("odo", "service", "create", "dh-postgresql-apb", "--plan", "dev",
+				"-p", "postgresql_user=luke", "-p", "postgresql_password=secret",
+				"-p", "postgresql_database=my_data", "-p", "postgresql_version=9.6", serviceName,
 				"--context", context,
 			)
 
-			// now check if listing the service using app & project used above passes
+			// now check if listing the service using --context works
 			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
 			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
-				return strings.Contains(output, "dh-prometheus-apb")
+				return strings.Contains(output, serviceName)
 			})
-			stdOut := helper.CmdShouldPass("odo", "service", "list", "--app", app, "--project", project)
-			Expect(stdOut).To(ContainSubstring("dh-prometheus-apb"))
+			stdOut := helper.CmdShouldPass("odo", "service", "list", "--context", context)
+			Expect(stdOut).To(ContainSubstring(serviceName))
+
+			// now check if deleting the service using --context works
+			stdOut = helper.CmdShouldPass("odo", "service", "delete", "--context", context)
+			Expect(stdOut).To(ContainSubstring(serviceName))
 		})
 
 		It("should be able to list services in a given app and project combination", func() {
--- a/tests/integration/servicecatalog/service_test.go
+++ b/tests/integration/servicecatalog/service_test.go
@@ -140,8 +140,20 @@ var _ = Describe("odoServiceE2e", func() {
 			Expect(stdOut).To(ContainSubstring(serviceName))
 
 			// now check if deleting the service using --context works
-			stdOut = helper.CmdShouldPass("odo", "service", "delete", "--context", context)
+			stdOut = helper.CmdShouldPass("odo", "service", "delete", "-f", "serviceName", "--context", context)
 			Expect(stdOut).To(ContainSubstring(serviceName))
+		})
+
+		It("should fail to create a service when given directory doesn't contain a context", func() {
+			// cd to the originalDir to create service using --context
+			helper.Chdir(originalDir)
+			stdOut := helper.CmdShouldFail("odo", "service", "create", "dh-postgresql-apb", "--plan", "dev",
+				"-p", "postgresql_user=luke", "-p", "postgresql_password=secret",
+				"-p", "postgresql_database=my_data", "-p", "postgresql_version=9.6", serviceName,
+				"--context", context,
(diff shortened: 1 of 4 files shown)


dharmit:
I'll add tests for failure scenario. Missed adding that in the initial PR.

dharmit:
> I'll add tests for failure scenario. Missed adding that in the initial PR.

Added this as well as added `--context` flag to `odo service list` and `odo service delete` commands as that seemed like a logical thing to do since we're adding the flag to `odo service create` command.

surajnarwade:
code wise LGTM

I will wait for CI and then approve it :smile:

tests/integration/servicecatalog/service_test.go
```diff
@@ -107,6 +108,7 @@ var _ = Describe("odoServiceE2e", func() {
 			project = helper.CreateRandProject()
```
amitkrout: To avoid preference conflict in parallel run, use format of context and project create
```
context = helper.CreateNewContext()
project = helper.CreateRandProject()
```

Also in ```JsutAfterEach``` unset the env var 
```
```

tests/integration/servicecatalog/service_test.go
```diff
@@ -115,6 +117,45 @@ var _ = Describe("odoServiceE2e", func() {
 			helper.DeleteDir(context)
 			helper.Chdir(originalDir)
 		})
+
+		It("should be able to create, list and delete a service using a given value for --context", func() {
+			// create a component by copying the example
+			helper.CopyExample(filepath.Join("source", "nodejs"), context)
+			helper.CmdShouldPass("odo", "create", "nodejs", "--app", app, "--project", project)
```
amitkrout: We have already one ```nodejs``` component create in this test file. Can you try with some other component 
```
$ odo catalog list components
nodejs      openshift     10,6,8,8-RHOAR,latest
perl        openshift     5.24,5.26,latest
php         openshift     7.0,7.1,latest
python      openshift     2.7,3.5,3.6,latest
ruby        openshift     2.3,2.4,2.5,latest
wildfly     openshift     10.0,10.1,11.0,12.0,13.0,8.1,9.0,latest
```

tests/integration/servicecatalog/service_test.go
```diff
@@ -115,6 +117,45 @@ var _ = Describe("odoServiceE2e", func() {
...
+
+			// now check if deleting the service using --context works
+			stdOut = helper.CmdShouldPass("odo", "service", "delete", "-f", "serviceName", "--context", context)
+			Expect(stdOut).To(ContainSubstring(serviceName))
+		})
+
+		It("should fail to create a service when given directory doesn't contain a context", func() {
+			// cd to the originalDir to create service using --context
+			helper.Chdir(originalDir)
+			stdOut := helper.CmdShouldFail("odo", "service", "create", "dh-postgresql-apb", "--plan", "dev",
```
amitkrout: Pleas add ```SetDefaultConsistentlyDuration(30 * time.Second)``` in ```BeforeEach(func(){}``` block because ```CmdShouldFail``` has been implemented in a different way, otherwise later it will be flaky

See http://onsi.github.io/gomega/#making-asynchronous-assertions for more details.

tests/integration/servicecatalog/service_test.go
```diff
@@ -115,6 +117,45 @@ var _ = Describe("odoServiceE2e", func() {
...
+			// now check if listing the service using --context works
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, serviceName)
+			})
+			stdOut := helper.CmdShouldPass("odo", "service", "list", "--context", context)
+			Expect(stdOut).To(ContainSubstring(serviceName))
+
+			// now check if deleting the service using --context works
+			stdOut = helper.CmdShouldPass("odo", "service", "delete", "-f", "serviceName", "--context", context)
```
amitkrout: ```"serviceName" -> serviceName```

cdrage:
/retest
(travis failed due to a flake)

tests/integration/servicecatalog/service_test.go
dharmit: Thanks for the catch!

dharmit:
Made changes as per @amitkrout's review feedback.

@kadel @girishramnani @mohammedzee1000 @cdrage can I please have a review on this? I have already demo'd this and have made numerous requests in the past on slack for reviewing this!

amitkrout:
Travis is failing 
/hold

tests/integration/servicecatalog/service_test.go
```diff
@@ -104,17 +105,60 @@ var _ = Describe("odoServiceE2e", func() {
...
+
+			// now check if deleting the service using --context works
+			stdOut = helper.CmdShouldPass("odo", "service", "delete", "-f", serviceName, "--context", context)
+			Expect(stdOut).To(ContainSubstring(serviceName))
+		})
+
+		It("should fail to create a service when given directory doesn't contain a context", func() {
+			// cd to the originalDir to create service using --context
+			helper.Chdir(originalDir)
+			stdOut := helper.CmdShouldFail("odo", "service", "create", "dh-postgresql-apb", "--plan", "dev",
```
amitkrout: Hmm, it seems the command returns exit code ```0```. So the command actually does not fail.

@dharmit use ```CmdShouldPass``` instead.

dharmit: I think the right thing to do would be remove the `"--context", context,` appearing a couple of lines below the one you pointed because as we can see from the `It(should fail to create a service when given directory doesn't contain a context")`, this test block is supposed to fail when working without a context.

amitkrout: Right, Have you tested it locally, i mean the same steps in the terminal ?

I doubt it will give a none zero exit code. Anyway lets see

amitkrout: > I think the right thing to do would be remove the "--context", context,

In this scenario there is no component in the context dir. So i feel even after removing the context flag won't work as per the expectation.

dharmit: I've not checked tests locally for this instance. For some reason the tests were not getting started and I took the lazy approach of pushing it to a PR. Let me check this locally because what you say about removing the context flag also makes sense.

dharmit: #2051 is a result of this discussion. Marking this conversation resolved.

Issue #1975, opened 2019-08-01
Odo index file stores absolute filepath for files


girishramnani:
[kind/bug]

## What versions of software are you using?
- Operating System: Fedora
- Output of `odo version`:

## How did you run odo exactly?
```
odo create java
odo push
```
Check the `.odo/odo-file-index.json`, it will have files stored as absolute path which would 
- if someone checks in the index file and another user gets that repo then the first push will always be a full build
- makes the project less portable. If a developer moves the project around, the index breaks 
- unnecessary prefix `/Users/<user>` for each filepath. Contributes to size of the index file

girishramnani:
Opened a draft pr for people to try and also so we could discuss the best way to handle the breaking change

girishramnani:
Maybe we can even version the index file? 
We can use the same base struct as we use that we use for json output?

kadel:
> Maybe we can even version the index file?
> We can use the same base struct as we use that we use for json output?

One thing to keep in mind is that this file can be always recomputed. It doesn't hold any important information. Even if the structure changes, we can delete the file and recreate it. The only input that this will have is that it will trigger push even if files were not changed.

dharmit:
I have a question which is not exactly related to the absolute vs. relative path discussion.

Is the index file going to be tracked in a git repo? If it is, are we assuming that two developers working on same project will also work on same OpenShift cluster? Because if they work on different OpenShift clusters, wouldn't the json file make odo believe that some files are already pushed to the cluster when they actually aren't?

Am I thinking correctly or am I missing something? :thinking:

PR #2003, opened 2019-08-12
using relative filepaths in file indexer and added versioned file index

girishramnani:

storing relative filepaths in the file index and also storing the file index in a versioned struct called `FileIndex`
Note - This is a breaking change, this will break all the existing file maps. and I think users should remove the older file maps to transition ( or there could be some unexpected behavior )
fixes https://github.com/openshift/odo/issues/1975

## How to test changes?
```
odo create
odo push # check the file map to have relative paths
```

### DIFF
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -59,11 +59,13 @@ func resolveFilePath(directory string) (string, error) {
 	return directory, nil
 }
 
-// Run walks the given directory and finds the files which have changed and which were deleted/renamed
+// RunIndexer walks the given directory and finds the files which have changed and which were deleted/renamed
 // it reads the odo index file from the .odo folder
 // if no such file is present, it means it's the first time the folder is being walked and thus returns a empty list
 // after the walk, it stores the list of walked files with some information in a odo index file in the .odo folder
-func Run(directory string, ignoreRules []string) (filesChanged []string, filesDeleted []string, err error) {
+// The filemap stores the values as "relative filepath" => FileData but it the filesChanged and filesDeleted are absolute paths
+// to the files
+func RunIndexer(directory string, ignoreRules []string) (filesChanged []string, filesDeleted []string, err error) {
 	resolvedPath, err := resolveFilePath(directory)
 	if err != nil {
 		return filesChanged, filesDeleted, err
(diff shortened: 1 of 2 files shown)


pkg/util/file_indexer.go
```diff
@@ -8,35 +8,59 @@ import (
...
 	byteValue, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		return nil, err
 	}
 	// unmarshals the byte values and fill up the file read map
-	err = json.Unmarshal(byteValue, &fileReadMap)
+	err = json.Unmarshal(byteValue, &fi)
 	if err != nil {
-		return nil, err
+		// This is added here for backward compatibility because
```
girishramnani: Something to be aware of.
cc @kadel

pkg/util/file_indexer.go
```diff
@@ -8,35 +8,59 @@ import (
...
+
+// NewFileIndex returns a fileIndex
+func NewFileIndex() *FileIndex {
+
+	return &FileIndex{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "FileIndex",
+			APIVersion: "v1",
+		},
+		FileMap: make(map[string]FileData),
```
mik-dass: Maybe something better than `FileMap` like `FileObjects` or just `Files`

pkg/util/file_indexer.go
```diff
@@ -8,35 +8,59 @@ import (
...
+	FileMap map[string]FileData
+}
+
+// NewFileIndex returns a fileIndex
+func NewFileIndex() *FileIndex {
+
+	return &FileIndex{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "FileIndex",
+			APIVersion: "v1",
```
mik-dass: @kadel Should we go with semantic version?

pkg/util/file_indexer.go
```diff
@@ -8,35 +8,59 @@ import (
...
+	metav1.TypeMeta
+	FileMap map[string]FileData
+}
+
+// NewFileIndex returns a fileIndex
+func NewFileIndex() *FileIndex {
+
+	return &FileIndex{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "FileIndex",
```
mik-dass: OdoFileIndex

pkg/util/file_indexer.go
```diff
@@ -8,35 +8,59 @@ import (
...
 
 	"github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const fileIndexDirectory = ".odo"
 const fileIndexName = "odo-file-index.json"
 
+// FileIndex holds the file index used for storing local file state change
+type FileIndex struct {
```
mik-dass: OdoFileIndex

kadel: Why prepend it with `Odo`? Isn't it a little bit redundant?

pkg/util/file_indexer.go
kadel: I think that this is ok.
We can use the same versioning schema that Kuberentes uses - https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning 

> Stable level:
The version name is vX where X is an integer.

pkg/util/file_indexer.go
girishramnani: Yeah as its not going to be used by something outside odo, I do we need it?

pkg/util/file_indexer.go
girishramnani: ack

pkg/util/file_indexer.go
girishramnani: The reason for not adding `Odo` as because the `Kind` for preference is `Preference` and not `OdoPreference` hence wanted to keep things uniform

pkg/util/file_indexer.go
mik-dass: > Yeah as its not going to be used by something outside odo

Since the struct isn't useful outside odo, just wanted it in the prefix to indicate that it's for odo's internal use only

girishramnani: with that logic wouldn't we prefix every single struct in our project with `Odo`?

mik-dass: ok makes sense

Issue #1981, opened 2019-08-05
[Flake] Failed to update config to git component deployed


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: OpenShift CI
- Output of `odo version`: pr run

## How did you run odo exactly?
Run pr (on two test node) in OpenShift CI. I have seen this even on single test node.  

## Actual behavior
Flakey

## Expected behavior
Should not be flakey 

## Any logs, error output, etc?

Reference  - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/1920/pull-ci-openshift-odo-master-integration/1126

amitkrout:
Same error i can reproduce it on my local setup on the cluster of 3.x that i have been using since 2 days, however on a fresh cluster the chance of hitting the issue is very less (1 out of 10).

Same observation i can find on 4.* cluster on aws too.

amitkrout:
Some more same failure in CI - https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_odo/2001/pull-ci-openshift-odo-master-integration/1241/build-log.txt

amitkrout:
Reproduced in CI - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/1954/pull-ci-openshift-odo-master-integration/1061
```
odoURLIntegration Listing urls 
  should list appropriate URLs and push message
  /go/src/github.com/openshift/odo/tests/integration/url_test.go:41
 Created dir: /tmp/475069174
Creating a new project: qllqkgdnzd
Running odo with args: [project create qllqkgdnzd -w -v4] 
[odo] I0730 05:19:17.632597   31253 preference.go:116] The configFile is /tmp/475069174/config.yaml
[odo] I0730 05:19:17.632680   31253 occlient.go:486] Trying to connect to server api.ci-op-mvjt0q5q-f09f4.origin-ci-int-aws.dev.rhcloud.com:6443
[odo] I0730 05:19:17.651839   31253 occlient.go:493] Server https://api.ci-op-mvjt0q5q-f09f4.origin-ci-int-aws.dev.rhcloud.com:6443 is up
[odo] I0730 05:19:17.754536   31253 occlient.go:416] isLoggedIn err:  <nil> 
[odo]  output: "developer"
[...]

PR #2016, opened 2019-08-13
[WIP] Testing 1981

mik-dass:
just for reproducing #1981 and finding the reason

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ TIMEOUT ?= 7200s
 # TEST_EXEC_NODES=1, otherwise by default the specs are run in parallel on 4 ginkgo test node.
 # NOTE: Any TEST_EXEC_NODES value greater than one runs the spec in parallel
 # on the same number of ginkgo test nodes.
-TEST_EXEC_NODES ?= 2
+TEST_EXEC_NODES ?= 4
 
 # Slow spec threshold for ginkgo tests. After this time (in second), ginkgo marks test as slow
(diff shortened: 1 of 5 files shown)


mik-dass:
```
✗  Waiting for build to finish [30s]
[odo]  ✗  Failed to create component with name cmp-git. Please use `odo config view` to view settings used to create component. Error: unable to wait for build cmp-git-testing-1 to run: timed out waiting for the condition
[odo] unable get build log cmp-git-testing-1
[odo] github.com/openshift/odo/pkg/occlient.(*Client).FollowBuildLog
[odo] 	/go/src/github.com/openshift/odo/pkg/occlient/occlient.go:1855
[odo] github.com/openshift/odo/pkg/component.Build
[odo] 	/go/src/github.com/openshift/odo/pkg/component/component.go:794
[odo] github.com/openshift/odo/pkg/component.CreateComponent
```
/retest

mik-dass:
@kadel @amitkrout After some more debugging, it seems that the replication which is supposed to be created after the update call is not yet created when calling instantiate. Thus we get the 404 replication controller not found error.

mik-dass:
@kadel It seems the `automatic: false` set in https://github.com/openshift/odo/blob/bb301712ae6b67f922a2c4ae70ab5ab143b29db1/pkg/occlient/templates.go#L290 as part of the image triggers for git components doesn't work and thus a new replication controller and a new deployments was getting triggered.

mik-dass:
since #1981 is closed, closing this

PR #2025, opened 2019-08-16
Grooming formatting of the docs

boczkowska:
Small fixes in formatting of the docs to improve readability.

fixes #???


cdrage:
For `cli-reference.adoc` we actually use a script to generate it:

https://github.com/openshift/odo/blob/master/cmd/cli-doc/cli-doc.go

The formatting is a bit off.. so we'll have to edit it there.

boczkowska:
@cdrage thanks! Got it, will file an issue then.

PR #2053, opened 2019-08-28
Documentation on uninstallation procedure

### LINKED ISSUE #1733: Uninstall Script
[kind/Feature]

As far as I can tell, there is no uninstall script available for `odo` at this time. It would be helpful to have an OS-independent script similar to the install script. 

Uninstall `odo` in an OS-independent manner.

## Why is this needed?
Give users the ability to uninstall `odo`. My use case is that I would like to remove an old version from an environment I am working in.

boczkowska:
Added instructions on how to uninstall odo 

fixes #1733

### DIFF
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -207,3 +207,27 @@ You can edit `Environment Variables` faster using search as follows:
 
 . Left click on *Search* and type `env` or `environment`.
 . Select *Edit environment variables for your account* and follow steps 5 and 6 listed above.
+
+==== Uninstalling odo from your system
+
+To uninstall `odo` run the script `installer.sh` which is located in link:https://github.com/openshift/odo[`odo` repository on GitHub].
+
+. Run the script
+
+** If you don't have the scipt locally on your machine:
++
+----
+$ curl -L https://github.com/openshift/odo/raw/master/scripts/installer.sh | bash -s -- --uninstall
+----
+
+** If you have the scipt locally on your machine:
++
+----
+$ sh installer.sh --uninstall
+----
+
+. Delete local `odo` files: 
++
+----
+$ rm -rf ~/.odo
+----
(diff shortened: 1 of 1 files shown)


docs/installation.adoc
```diff
@@ -207,3 +207,27 @@ You can edit `Environment Variables` faster using search as follows:
...
+** If you have the scipt locally on your machine:
++
+----
+$ sh installer.sh --uninstall
+----
+
+. Delete local `odo` files: 
++
+----
+$ rm -rf ~/.odo
```
mik-dass: This will remove the `.odo` directory from the home directory only. Maybe add `example` in the above line

kadel: @mik-dass not sure what you mean. It is supposed to remove it from the home directory only.

kadel:
/retest

the error was the same as in https://github.com/openshift/odo/issues/1981

docs/installation.adoc
boczkowska: @mik-dass I'm not sure what you mean by adding `example`. The procedure is two-step: 1) uninstall odo w/ script, 2) remove whatever local files are left there in home directory.

mik-dass: @kadel @boczkowska Is this example to remove the odo preference folder? Or is to remove all the `.odo` folder from sub directories? I am kind of confused by the heading `Delete local odo files`

boczkowska: @mik-dass Reworded slightly. Tell me if that makes more sense now. :)

kadel: > @kadel @boczkowska Is this example to remove the odo preference folder? Or is to remove all the `.odo` folder from sub directories? I am kind of confused by the heading `Delete local odo files`

@mik-dass 
odo preference files. Uninstall process shouldn't remove `.odo` folder from the component directories. That is part of the component, and `odo component delete` should handle that.

mik-dass: > odo preference files. Uninstall process shouldn't remove .odo folder from the component directories. That is part of the component, and odo component delete should handle that.

Exactly, that was my concern. Now the rewording seems better.

Issue #1357, opened 2019-02-22
Machine output in case of catalog list operation


lordofthejars:
[kind/Feature]

## Which functionality do you think we should add?

`odo catalog list` command should be able to produce the output in machine format too.

## Why is this needed?

Some integrators might need this information to create some guards in case of the catalog does not contain a specific element. For example, we could create a guard that makes a test to not be executed if openjdk s2i is not in the catalog.

surajnarwade:
TODO: 

* `odo catalog list components -o json`
* `odo catalog list services -o json`

lordofthejars:
Ok I am going to work on that

cdrage:
Closed by mistake. Still need #2067 merged in

PR #2066, opened 2019-08-30
Adds `odo catalog list services -o json`

cdrage:
Adds the ability to list catalog services and output to json.

The format is as follows:

```json
{
  "kind": "ServiceList",
  "metadata": {
    "creationTimestamp": null
  },
  "items": [
    {
      "name": "cakephp-mysql-persistent",
      "hidden": false,
      "planList": [
        "default"
      ]
    },
  ]
}
```

Closes https://github.com/openshift/odo/issues/1357

### DIFF
--- a/tests/integration/servicecatalog/cmd_service_test.go
+++ b/tests/integration/servicecatalog/cmd_service_test.go
@@ -37,6 +37,14 @@ var _ = Describe("odo service command tests", func() {
 		})
 	})
 
+	Context("checking machine readable output for service catalog", func() {
+		It("should succeed listing catalog components", func() {
+			// Since service catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			output := helper.CmdShouldPass("odo", "catalog", "list", "services", "-o", "json")
+			Expect(output).To(ContainSubstring("CatalogListServices"))
+		})
+	})
+
 	Context("create service with Env non-interactively", func() {
 		JustBeforeEach(func() {
 			project = helper.CreateRandProject()
--- a/pkg/machineoutput/types.go
+++ b/pkg/machineoutput/types.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/occlient"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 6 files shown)


cdrage:
Ready for reviewing! @dharmit @kadel @girishramnani 

Unfortunately for tests, I can't test against service catalog... any ideas @amitkrout ?

pkg/service/service.go
```diff
@@ -397,3 +399,21 @@ func isRequired(required []string, name string) bool {
 	}
 	return false
 }
+
+// MachineReadableSuccessOutput prints out machine readable output
+func MachineReadableSuccessOutput(services []occlient.Service) {
```
dharmit: Since the function is dealing with machine output, wouldn't it be helpful to keep it under the `machineoutput` package?

cdrage: No, as each output will be different.. Or else we'll run into an issue where in the machineout package, we will have multiple functions with different names :(

For example, It'd probably be named ServiceCatalogList or something along those lines, IMO doesn't make sense. But I could be wrong.

cdrage:
You have a good point on putting the functions under `machineoutout` package instead, thanks @dharmit !

Ready for another review :+1:

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -37,6 +37,13 @@ var _ = Describe("odo service command tests", func() {
 		})
 	})
 
+	Context("checking machine readable output for service catalog", func() {
```
amitkrout: +1, This is a good place to validate instead of creating a separate test file.

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -37,6 +37,13 @@ var _ = Describe("odo service command tests", func() {
...
+		It("should succeed listing catalog components", func() {
+			// Since service catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			helper.CmdShouldPass("odo", "catalog", "list", "services", "-o", "json")
```
amitkrout: Instead of matching the whole json can you match the basic key value pair like ```"kind": "ServiceList"``` and atleast one ```"items": [``` you mentioned in pr description.

amitkrout:
> Create a separate test file `cmd_catalog_test.go` under integration package

https://github.com/openshift/odo/pull/2066#discussion_r321587777

cdrage:
Updated based on your comments @amitkrout

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -37,6 +37,14 @@ var _ = Describe("odo service command tests", func() {
 		})
 	})
 
+	Context("checking machine readable output for service catalog", func() {
+		It("should succeed listing catalog components", func() {
+			// Since service catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			output := helper.CmdShouldPass("odo", "catalog", "list", "services", "-o", "json")
+			Expect(output).To(ContainSubstring("CatalogListServices"))
```
amitkrout: +1

pkg/service/service.go
dharmit: I agree with the function name point you raised. To purely counter that point, we have adopted similar approach for auto-completion where everything lives under `pkg/odo/util/completion/completionhandlers.go`.

Besides that, I think, there's scope of avoiding duplication of code by putting it under `machineoutput` package. For example, we have renamed `Success` and `Error` structs in that package to `GenericSuccess` and `GenericError` respectively.

@kadel can you please share your thoughts?

girishramnani:
Looks good
/lgtm

PR #2067, opened 2019-08-30
Add `odo catalog list components -o json`

cdrage:
Adds json output for listing catalog components. For example:

```sh
odo catalog list components -o json
```

Example:

```json
{
  "kind": "ComponentTypeList",
  "metadata": {
    "creationTimestamp": null
  },
  "items": [
    {
      "kind": "ComponentType",
      "metadata": {
        "name": "dotnet",
        "creationTimestamp": null
      },
      "spec": {
        "namespace": "openshift",
        "allTags": [
          "2.0",
          "latest"
        ],
        "nonHiddenTags": [
          "2.0",
          "latest"
        ],
        "imageStreamRef": {
...

          }
        }
      }
    },
  ],
}
```

Closes https://github.com/openshift/odo/issues/1357

### DIFF
--- a/pkg/catalog/types.go
+++ b/pkg/catalog/types.go
@@ -1,0 +1,28 @@
+package catalog
+
+import (
+	imagev1 "github.com/openshift/api/image/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Catalog ...
+type Catalog struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              CatalogSpec `json:"spec,omitempty"`
+}
+
+// CatalogSpec ...
+type CatalogSpec struct {
+	Namespace      string              `json:"namespace"`
+	AllTags        []string            `json:"allTags"`
+	NonHiddenTags  []string            `json:"nonHiddenTags"`
+	ImageStreamRef imagev1.ImageStream `json:"imageStreamRef"`
+}
+
+// CatalogImageList ...
+type CatalogImageList struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Items             []Catalog `json:"items"`
+}
--- a/pkg/machineoutput/types.go
+++ b/pkg/machineoutput/types.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -118,6 +118,13 @@ func componentTests(args ...string) {
 			Expect(cmpAllList).To(ContainSubstring("cmp-git"))
 			helper.CmdShouldPass("odo", append(args, "delete", "cmp-git", "-f")...)
 		})
+
+		It("should succeed listing catalog components", func() {
+
+			// Since components catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			output := helper.CmdShouldPass("odo", "catalog", "list", "components", "-o", "json")
+			Expect(output).To(ContainSubstring("CatalogList"))
+		})
 	})
 
 	Context("Test odo push with --source and --config flags", func() {
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -8,30 +8,28 @@ import (
 	imagev1 "github.com/openshift/api/image/v1"
 	"github.com/openshift/odo/pkg/occlient"
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type CatalogImage struct {
-	Name          string
-	Namespace     string
-	AllTags       []string
-	NonHiddenTags []string
-
-	// this is used to get metadata like image URL for each tag present in the image stream
-	imageStreamRef imagev1.ImageStream
-}
-
 // List lists all the available component types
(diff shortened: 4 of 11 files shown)


cdrage:
This builds commit from this PR: https://github.com/openshift/odo/pull/2066

So https://github.com/openshift/odo/pull/2066 will need to be merged in first :+1:

pkg/service/service.go
```diff
@@ -397,3 +399,21 @@ func isRequired(required []string, name string) bool {
 	}
 	return false
 }
+
+// MachineReadableSuccessOutput prints out machine readable output
+func MachineReadableSuccessOutput(services []occlient.Service) {
```
dharmit: Besides the same point that I mentioned in https://github.com/openshift/odo/pull/2066#pullrequestreview-283387331, it looks like `MachineReadableSuccessOutput` in both the PRs has more code common between them than uncommon. Is it possible to have single function for these?

cdrage: With the way that yaml unmarshal's, I don't think so :(

cdrage:
Thanks for commenting @dharmit I've updated it based upon your suggestions.

tests/integration/component.go
```diff
@@ -118,6 +118,12 @@ func componentTests(args ...string) {
 			Expect(cmpAllList).To(ContainSubstring("cmp-git"))
 			helper.CmdShouldPass("odo", append(args, "delete", "cmp-git", "-f")...)
 		})
+
+		It("should succeed listing catalog components", func() {
```
amitkrout: +1, No need for a separate test file.

tests/integration/component.go
```diff
@@ -118,6 +118,12 @@ func componentTests(args ...string) {
...
+
+			// Since components catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			helper.CmdShouldPass("odo", "catalog", "list", "components", "-o", "json")
```
amitkrout: Can you match it with some minimum output if the json output is big.

amitkrout: Or It can be helpful - https://github.com/Benjamintf1/unmarshalledmatchers. 

its already vendored in our repo, you just need to import it.

tests/integration/component.go
```diff
@@ -118,6 +118,13 @@ func componentTests(args ...string) {
 			Expect(cmpAllList).To(ContainSubstring("cmp-git"))
 			helper.CmdShouldPass("odo", append(args, "delete", "cmp-git", "-f")...)
 		})
+
+		It("should succeed listing catalog components", func() {
+
+			// Since components catalog is constantly changing, we simply check to see if this command passes.. rather than checking the JSON each time.
+			output := helper.CmdShouldPass("odo", "catalog", "list", "components", "-o", "json")
+			Expect(output).To(ContainSubstring("CatalogListComponents"))
```
amitkrout: +1

amitkrout:
```
[odo]  ✗  Unable to connect to OpenShift cluster, is it down?
```
/test integration

amitkrout:
Integration test looks good to me, however i seek review from @dharmit too.

pkg/machineoutput/types.go
```diff
@@ -54,6 +56,24 @@ func OutputError(machineOutput interface{}) {
 	}
 }
 
+// CatalogListComponents `odo catalog list components -o json` standard machine readable output
+func CatalogListComponents(images []catalog.CatalogImage) {
```
kadel: This is not good approach :-( 
Catalog should have similar structure like `components`  - see https://github.com/openshift/odo/blob/master/pkg/component/types.go
or `url` https://github.com/openshift/odo/blob/master/pkg/url/types.go

You need to create `types.go`  move `CatalogImage` to it and  create `CatalogImageList` struct and modify `catalog.List`  to return it.

cdrage:
Updated this implementation, let me know if this is okay and then I can refactor what I've already done in machineoutput for `odo catalog list services`...

Also, please check if the `Kind` is correct. `CatalogListServices` or if you want me to use `CatalogList`

pkg/catalog/types.go
```diff
@@ -0,0 +1,18 @@
+package catalog
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type CatalogImage struct {
+	Name          string   `json:"name"`
```
kadel: - Missing `TypeMeta` and `ObjectMeta`.
- `AllTags` and `NonHiddentTags` should be in a separate `CatalogImageSpec` struct  and in `Spec` field.
- no need for `Namespace` as it is already is in `ObjectMeta`

cdrage:
Ready for another review, example ofthe output is below:


```json
{
  "kind": "ComponentTypeList",
  "metadata": {
    "creationTimestamp": null
  },
  "items": [
    {
      "kind": "ComponentType",
      "metadata": {
        "name": "dotnet",
        "creationTimestamp": null
      },
      "spec": {
        "namespace": "openshift",
        "allTags": [
          "2.0",
          "latest"
        ],
        "nonHiddenTags": [
          "2.0",
          "latest"
        ],
        "imageStreamRef": {
...

          }
        }
      }
    },
  ],
}
```

cdrage:
ping @amitkrout @dharmit @kadel

pkg/catalog/types.go
```diff
@@ -0,0 +1,28 @@
...
+// Catalog ...
+type Catalog struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              CatalogSpec `json:"spec,omitempty"`
+}
+
+// CatalogSpec ...
+type CatalogSpec struct {
+	Namespace      string              `json:"namespace"`
```
kadel: namespace field is already in `metav1.ObjectMeta` (`metadata`) we should use that one instead of creating a new duplicate field.

cdrage: Heck. Completely forgot about this part.

cdrage:
Closing in favour of https://github.com/openshift/odo/pull/2133

PR #2078, opened 2019-09-04
Removes connection check with OpenShift cluster

### LINKED ISSUE #1588: connection checks
Commands that don't require cluster should work even "offline"

We need to remove the connection check that we have there for every command. And use it only for the commands that actually require to communicate with the cluster.

Commands that require connection:
 - `odo push`
 - `odo component delete`
 - `odo service *`
 - `odo link *`
 - `odo project *`
 - `odo app delete`
 - `odo login`
 - `odo logout`

I might be missing something

dharmit:

Removes connection check from the workflow

related to #1588 

## How to test changes?

Do `odo logout` and see if things that should work without an active connection with OpenShift cluster still work.

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -221,7 +221,7 @@ func getBootstrapperImage() string {
 }
 
 // New creates a new client
-func New(skipConnectionCheck bool) (*Client, error) {
+func New() (*Client, error) {
 	var client Client
 
 	// initialize client-go clients
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -289,15 +289,6 @@ func New(skipConnectionCheck bool) (*Client, error) {
 	}
 	client.Namespace = namespace
 
-	// if we're not skipping the connection check, check the connection :)
-	if !skipConnectionCheck {
-		if !isServerUp(config.Host) {
-			return nil, errors.New("Unable to connect to OpenShift cluster, is it down?")
-		}
-		if !client.isLoggedIn() {
-			return nil, errors.New("Please log in to the cluster")
-		}
-	}
 	return &client, nil
 }
 
--- a/pkg/odo/cli/cli.go
+++ b/pkg/odo/cli/cli.go
@@ -18,7 +18,6 @@ import (
 	"github.com/openshift/odo/pkg/odo/cli/url"
 	"github.com/openshift/odo/pkg/odo/cli/utils"
 	"github.com/openshift/odo/pkg/odo/cli/version"
-	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	"github.com/openshift/odo/pkg/odo/util"
 	odoutil "github.com/openshift/odo/pkg/odo/util"
 	"github.com/spf13/cobra"
@@ -86,8 +85,6 @@ func NewCmdOdo(name, fullName string) *cobra.Command {
 	// Cobra supports persistent flags, which, if defined here,
 	// will be global for your application.
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.odo.yaml)")
-
-	rootCmd.PersistentFlags().Bool(genericclioptions.SkipConnectionCheckFlagName, false, "Skip cluster check")
 
 	// Add the machine readable output flag to all commands
 	// We use "flag" in order to make this accessible throughtout ALL of odo, rather than the
--- a/pkg/odo/cli/version/version.go
+++ b/pkg/odo/cli/version/version.go
@@ -52,7 +52,7 @@ func NewVersionOptions() *VersionOptions {
 func (o *VersionOptions) Complete(name string, cmd *cobra.Command, args []string) error {
 	if !o.clientFlag {
 		// Let's fetch the info about the server, ignoring errors
-		client, err := occlient.New(true)
+		client, err := occlient.New()
 		if err == nil {
 			o.serverInfo, _ = client.GetServerVersion()
 		}
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -40,9 +40,7 @@ func ClientWithConnectionCheck(command *cobra.Command, skipConnectionCheck bool)
 	return client(command)
 }
 
-// client creates an oc client based on the command flags, overriding the skip connection check flag with the optionally
-// specified shouldSkipConnectionCheck boolean.
-// We use varargs to denote the optional status of that boolean.
+// client creates an oc client based on the command flags
 func client(command *cobra.Command) *occlient.Client {
 	client, err := occlient.New()
 	util.LogErrorAndExit(err, "")
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -37,28 +37,14 @@ func Client(command *cobra.Command) *occlient.Client {
 // ClientWithConnectionCheck returns an oc client configured for this command's options but forcing the connection check status
 // to the value of the provided bool, skipping it if true, checking the connection otherwise
 func ClientWithConnectionCheck(command *cobra.Command, skipConnectionCheck bool) *occlient.Client {
-	return client(command, skipConnectionCheck)
+	return client(command)
 }
 
 // client creates an oc client based on the command flags, overriding the skip connection check flag with the optionally
 // specified shouldSkipConnectionCheck boolean.
 // We use varargs to denote the optional status of that boolean.
-func client(command *cobra.Command, shouldSkipConnectionCheck ...bool) *occlient.Client {
-	var skipConnectionCheck bool
-	switch len(shouldSkipConnectionCheck) {
-	case 0:
-		var err error
-		skipConnectionCheck, err = command.Flags().GetBool(SkipConnectionCheckFlagName)
-		util.LogErrorAndExit(err, "")
-	case 1:
-		skipConnectionCheck = shouldSkipConnectionCheck[0]
-	default:
-		// safeguard: fail if more than one optional bool is passed because it would be a programming error
-		log.Errorf("client function only accepts one optional argument, was given: %v", shouldSkipConnectionCheck)
-		os.Exit(1)
-	}
-
-	client, err := occlient.New(skipConnectionCheck)
+func client(command *cobra.Command) *occlient.Client {
+	client, err := occlient.New()
 	util.LogErrorAndExit(err, "")
 
 	return client
--- a/pkg/odo/genericclioptions/context_flags.go
+++ b/pkg/odo/genericclioptions/context_flags.go
@@ -1,8 +1,6 @@
 package genericclioptions
 
 const (
-	// SkipConnectionCheckFlagName is the name of the global flag used to skip connection check in the client
-	SkipConnectionCheckFlagName = "skip-connection-check"
 	// ProjectFlagName is the name of the flag allowing a user to specify which project to operate on
 	ProjectFlagName = "project"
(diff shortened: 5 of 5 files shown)


girishramnani:
Can we add a odo login logout test for this?

pkg/occlient/occlient.go
```diff
@@ -289,15 +289,6 @@ func New(skipConnectionCheck bool) (*Client, error) {
 	}
 	client.Namespace = namespace
 
-	// if we're not skipping the connection check, check the connection :)
-	if !skipConnectionCheck {
```
girishramnani: Do we need this skipConnectionCheck variable anymore? We removed the check from where ever we can and now if its there at any place that would be a necessity.

girishramnani: So can we remove this variable al all together?

dharmit:
@amitkrout 

> @dharmit These are the command i abstracted from the issue
> 
>     * odo component create *
> 
>     * odo url create *
> 
>     * odo config *
> 
>     * odo preference *
> 
>     * odo utils *
> 
>     * odo version
>
> Do you think documentation of these commands are helpful in someway ? 

I don't understand what you mean by "abstracted" here. I think these are the commands that use `skipConnectionCheck` in some way in the code, and as @girishramnani mentioned in https://github.com/openshift/odo/pull/2078#pullrequestreview-284035727, I believe I should remove its references from the code.

> and global flag `--skip-connection-check` should not be available for these command help, right ?

I think we're removing `--skip-connection-check` completely. If this PR gets merged, by default, we are not going to check for the connection in our code. So we need to remove it from everywhere/anywhere. I'll do that. Thanks for the catch. :+1:

amitkrout:
> I don't understand what you mean by "abstracted" here. I think these are the commands that use skipConnectionCheck in some way in the code, and as @girishramnani mentioned in #2078 (review), I believe I should remove its references from the code.

I mean abstracted form issue #1588 discussion.

mik-dass:
Will it be possible to remove this check for specific commands?

dharmit:
> Will it be possible to remove this check for only specific commands?

Which commands do you think require a connection check before being executed? Also, I believe, based on the comments and discussions I read/heard, that we'll be handling connection error at a later stage when the command fails instead of spending additional time doing it.

@kadel maybe you have some comments to add here?

pkg/occlient/occlient.go
dharmit: Addressed. PTAL.

amitkrout:
> Can we add a odo login logout test for this?

No, but can be handled like @dharmit has done it before - https://github.com/openshift/odo/blob/master/tests/integration/cmd_pref_config_test.go#L267

dharmit:
> > Can we add a odo login logout test for this?
> 
> No, but can be handled like @dharmit has done it before - https://github.com/openshift/odo/blob/master/tests/integration/cmd_pref_config_test.go#L267

Testing `odo login` after removing connection checks doesn't make a lot of sense to me because `odo login` code doesn't seem to make use of `occlient.New()` function as far as I can understand. And testing `odo logout` is not recommended because the tests are executed paralelly and doing `odo logout` could cause the subsequent tests to fail (thanks @amitkrout for explaining this).

Besides, we're changing the way `odo` works at a very fundamental level and, IMHO, CI passing all the tests is the only indicator of things being OK in this situation. Pretty much every command makes a call to `genericclioptions.NewContext` which then calls `genericclioptions.newContext` which is where the call to `occlient.New` is buried into. And if the integration tests are working OK, that's an indicator that `skipConnectionCheck` didn't cause any adverse effect.

Correct me if I am wrong. @girishramnani @kadel @amitkrout

amitkrout:
> > > Can we add a odo login logout test for this?
> > 
> > 
> > No, but can be handled like @dharmit has done it before - https://github.com/openshift/odo/blob/master/tests/integration/cmd_pref_config_test.go#L267
> 
> Testing `odo login` after removing connection checks doesn't make a lot of sense to me because `odo login` code doesn't seem to make use of `occlient.New()` function as far as I can understand. And testing `odo logout` is not recommended because the tests are executed paralelly and doing `odo logout` could cause the subsequent tests to fail (thanks @amitkrout for explaining this).
> 
> Besides, we're changing the way `odo` works at a very fundamental level and, IMHO, CI passing all the tests is the only indicator of things being OK in this situation. Pretty much every command makes a call to `genericclioptions.NewContext` which then calls `genericclioptions.newContext` which is where the call to `occlient.New` is buried into. And if the integration tests are working OK, that's an indicator that `skipConnectionCheck` didn't cause any adverse effect.
> 
> Correct me if I am wrong. @girishramnani @kadel @amitkrout

@dhrmit thanks for the clarification.
On the basis of existing test scenario, TBH even i don't think we need any integration test because the connection check code is removed (instead of improve) to improve the command performance. 

There is a counter intuitive way to verify it just by checking the verbose log that should not contain server connection info, but i don't recommend it.

kadel:
> > Will it be possible to remove this check for only specific commands?
> 
> Which commands do you think require a connection check before being executed? Also, I believe, based on the comments and discussions I read/heard, that we'll be handling connection error at a later stage when the command fails instead of spending additional time doing it.
> 
> @kadel maybe you have some comments to add here?

Lets completely remove the connection checks. If needed we can always add it back on per-command basis.

kadel:
@dharmit Don't forget to remove `--skip-connection-check` global flag.

mik-dass:
> Which commands do you think require a connection check before being executed? Also, I believe, based on the comments and discussions I read/heard, that we'll be handling connection error at a later stage when the command fails instead of spending additional time doing it.

https://github.com/openshift/odo/issues/1588#issuecomment-505322381 I think we need connection checks for `odo component create`. And how will the error be handled at a later?

amitkrout:
> > Which commands do you think require a connection check before being executed? Also, I believe, based on the comments and discussions I read/heard, that we'll be handling connection error at a later stage when the command fails instead of spending additional time doing it.
> 
> [#1588 (comment)](https://github.com/openshift/odo/issues/1588#issuecomment-505322381) I think we need connection checks for `odo component create`. And how will the error be handled at a later?

For example
```
$ odo component create pytho -v 4
I0906 13:53:59.233942   60705 preference.go:118] The path for preference file is /Users/amit/.odo/preference.yaml
I0906 13:53:59.567310   60705 cmdutils.go:48] Error:
no match found for : pytho in namespace default: imagestreams.image.openshift.io "pytho" is forbidden: User "amitkrout" cannot get resource "imagestreams" in API group "image.openshift.io" in the namespace "default"
.no match found for : pytho in namespace openshift: imagestreams.image.openshift.io "pytho" not found
 ✗  imagestreams.image.openshift.io "pytho" not found
```

kadel:
We definitely don't need a global connection check.
But we also need it for some commands. We will have to find some elegant way.

amitkrout:
@kadel @dharmit Connection check will not be the part of ```odo push``` flow as per the code change because the whole check is removed. 

Correct me if i am wrong.

amitkrout:
> @kadel @dharmit Connection check will not be the part of `odo push` flow as per the code change because the whole check is removed.
> 
> Correct me if i am wrong.

@dharmit I just check ```odo push -v 4``` and ```odo catalog list component -v 4``` using this pr. None of these command uses connection check.

dharmit:
```
---
Container test exited with code 1, reason Error
---
Another process exited
---
```
/test e2e-scenarios

dharmit:
@amitkrout yes, that's the idea here. If you try to do `odo push` or `odo catalog list components` after doing a `odo logout`, that's when you'll see the error. I see an error like below:
```
$ odo catalog list components -v4
I0906 18:07:52.646662    3912 catalog.go:81] Unable to get Image Streams from namespace openshift. Error unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "openshift": no RBAC policy matched
I0906 18:07:52.647365    3912 catalog.go:90] Unable to get Image Streams from namespace myproject. Error unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "myproject": no RBAC policy matched
I0906 18:07:52.647395    3912 cmdutils.go:48] Error:
unable to get image streams: Failed to fetch imagestreams from both openshift and myproject namespaces.
Please ensure that a builder imagestream of required version for the component exists in either openshift or myproject namespaces: unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "openshift": no RBAC policy matched.
unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "myproject": no RBAC policy matched
 ✗  unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "openshift": no RBAC policy matched.
unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "myproject": no RBAC policy matched

$ odo catalog list components
 ✗  unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "openshift": no RBAC policy matched.
unable to list imagestreams: imagestreams.image.openshift.io is forbidden: User "system:anonymous" cannot list imagestreams.image.openshift.io in the namespace "myproject": no RBAC policy matched
```
And for `odo push`:
```sh
$ odo push
 ✓  Creating project myproject
 ✗  Failed creating project myproject
 ✗  unknown (get projects.project.openshift.io)
```
We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.

amitkrout:
```
$ odo -h
(OpenShift Do) odo is a CLI tool for running OpenShift applications in a fast and automated matter. Reducing the complexity of deployment, odo adds iterative development without the worry of deploying your source code. 

Find more information at https://github.com/openshift/odo

[...]
Flags:
      --complete             Install completion for odo command
  -o, --o string             Specify output format, supported format: json (default "json")
      --uncomplete           Uninstall completion for odo command
  -y, --y                    Don't prompt user for typing 'yes' when installing completion

Use "odo [command] --help" for more information about a command.
```
/approve

amitkrout:
> We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.

@kadel WDYT ?

kadel:
> We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.

This can be solved in another PR. Let's keep the issue open once this gets merged.

amitkrout:
> > We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.
> 
> This can be solved in another PR. Let's keep the issue open once this gets merged.

/lgtm

amitkrout:
> > > We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.
> > 
> > 
> > This can be solved in another PR. Let's keep the issue open once this gets merged.
> 
> /lgtm

@kadel @mohammedzee1000 @cdrage @girishramnani @mik-dass I am unable to add ```lgtm``` label because my access has been revoked since 7hrs. If you guys want you can add the ```/lgtm``` label on behalf of me or else you can go for a round of review.

amitkrout:
> > > > We could argue if handling these failures in more understandable way should be done as a part of this PR or a different one.
> > > 
> > > 
> > > This can be solved in another PR. Let's keep the issue open once this gets merged.
> > 
> > 
> > /lgtm
> 
> @kadel @mohammedzee1000 @cdrage @girishramnani @mik-dass I am unable to add `lgtm` label because my access has been revoked since 7hrs. If you guys want you can add the `/lgtm` label on behalf of me or else you can go for a round of review.

/lgtm

PR #2086, opened 2019-09-04
Makes update information italic text

cdrage:
Changes the update information to display at italic in order to not get
mixed in with the other output.

Also remove the quotes for `odo preference set UpdateNotification false`
so it's easier to copy-and-paste from the terminal.

To test:

 1. Modify pkg/odo/cli/version/version.go VERSION var to "v1.0.0-beta4"
 or any version previous
 2. Compile
 3. Run a standard create command `odo create nodejs`

### DIFF
--- a/cmd/odo/odo.go
+++ b/cmd/odo/odo.go
@@ -2,10 +2,10 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
 
 	"github.com/golang/glog"
+	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/odo/cli"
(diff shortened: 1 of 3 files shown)


cdrage:
Demo:

Screenshot from 2019-09-04 14-25-58

kadel:
@cdrage have you tested this on Windows?

girishramnani:
@cdrage waiting on windows result. Once that's sorted will lgtm

cdrage:
Italic does not work on Windows which is intentional. With the library, it instead uses bold / a different colour in PowerShell / CMD.

See below:
Screenshot from 2019-09-06 15-17-52

ping @kadel / @dharmit / @amitkrout

PR #2091, opened 2019-09-05
Removes overlapping messages for odo git components

### LINKED ISSUE #1971: Overlapping messages when pushing git component
There are overlapping messages when running `odo push` on git component.

"Creating component" and "Waiting for build to finish" 

git-component

And when everything is finished "Creating component" message is there twice

Screenshot 2019-08-01 at 13 14 31

mik-dass:
fixes #1971 

How to test:

1. Create a git component and push twice
2. Try all update scenarios e.g git to local 

No overlapping should happen in the logs

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -161,17 +161,6 @@ func CreateFromGit(client *occlient.Client, params occlient.CreateArgs) error {
 		return errors.Wrapf(err, "unable to create git component %s", namespacedOpenShiftObject)
 	}
 
-	// Trigger build
-	if err = Build(client, params.Name, params.ApplicationName, params.Wait, params.StdOut, false); err != nil {
-		return errors.Wrapf(err, "failed to build component with args %+v", params)
-	}
-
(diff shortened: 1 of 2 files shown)


pkg/component/component.go
```diff
@@ -398,6 +390,19 @@ func CreateComponent(client *occlient.Client, componentConfig config.LocalConfig
 		); err != nil {
 			return errors.Wrapf(err, "failed to create component with args %+v", createArgs)
 		}
+
+		s.End(true)
+
+		// Trigger build
```
cdrage: Curious why we are moving the below code into this section rather than keeping it within`CreateWithGit`?

mik-dass: According to https://github.com/openshift/odo/issues/1971#issuecomment-521586286, build and deploy should happen outside the CreateFromGit function, so I removed it from there

cdrage: Makes sense. Thanks.

PR #2096, opened 2019-09-05
Remove newline from odo create output

cdrage:
Removes an extra newline that was added to the output.

`log.Infof` already adds a newline..

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -496,7 +496,7 @@ func (co *CreateOptions) Run() (err error) {
 			return errors.Wrapf(err, "failed to push the changes")
 		}
 	} else {
-		log.Infof("Please use `odo push` command to create the component with source deployed\n")
+		log.Infof("Please use `odo push` command to create the component with source deployed")
 	}
 	return
 }


pkg/odo/cli/component/create.go
```diff
@@ -496,7 +496,7 @@ func (co *CreateOptions) Run() (err error) {
 			return errors.Wrapf(err, "failed to push the changes")
 		}
 	} else {
-		log.Infof("Please use `odo push` command to create the component with source deployed\n")
+		log.Infof("Please use `odo push` command to create the component with source deployed")
```
dharmit: Comparing function names with those in `fmt` package, my understanding is that `Printf` requires one to explicitly add a new-line character if they wish to while `Println` adds a new-line character by itself. Although the main difference, IIUC, is that `Printf` allows one to format the output while `Println` doesn't. In this case, IMHO, it is the behaviour of `log.Infof` that we should change instead of the change we are making here. WDYT?

amitkrout:
```
```
/retest

pkg/odo/cli/component/create.go
cdrage: Our own logging adds newline out of convenience :)

dharmit: Yep, I understand that. But I think we could/should follow the convention of `Printf` vs. `Println`. Obviously I'm not going to hold this PR for that reason. :joy:

PR #2104, opened 2019-09-09
Refactor machinereadable output

cdrage:
Puts the code from project delete / create into `machineoutput/types.go`

### DIFF
--- a/pkg/odo/cli/project/create.go
+++ b/pkg/odo/cli/project/create.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/machineoutput"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	"github.com/openshift/odo/pkg/project"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


pkg/odo/cli/project/delete.go
```diff
@@ -83,7 +84,7 @@ func (pdo *ProjectDeleteOptions) Run() (err error) {
 		}
 
 		if log.IsJSON() {
-			project.MachineReadableSuccessOutput(pdo.projectName, successMessage)
+			machineoutput.ProjectSuccess(pdo.projectName, successMessage)
```
kadel: If the project was successfully deleted, we should not return it.

pkg/odo/cli/project/create.go
```diff
@@ -81,7 +82,7 @@ func (pco *ProjectCreateOptions) Run() (err error) {
 
 	// If -o json has been passed, let's output the approriate json output.
 	if log.IsJSON() {
-		project.MachineReadableSuccessOutput(pco.projectName, successMessage)
+		machineoutput.ProjectSuccess(pco.projectName, successMessage)
```
kadel: If the project creation was scessfull we should return `Kind` Project as defined in https://github.com/openshift/odo/blob/master/pkg/project/types.go#L7 representing the project that was just created

kadel: The best way to do this would be to modify `project.Create` to return the newly created project. The underlying call in `client-go` is doing exactly that, we just need to convert it to our `Project` and return it

cdrage:
/hold

Changes are being made to the API

PR #2112, opened 2019-09-10
Remove creating project info

cdrage:
There is enough info in verbose, we shouldn't always be outputting that
a project is being created.


kadel:
small fix

/lgtm
/approve

PR #2114, opened 2019-09-11
[WIP] Improve component tests

### LINKED ISSUE #2256: Improve component test spec
[kind/Enhancement]

Test spec for component test

## Why is this needed? 
To add more test coverage

amitkrout:

Includes more test scenario and of course smart use of all supported components flag
fixes #2256

## How to test changes?

make test-cmd-cmp

### DIFF
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -92,7 +92,7 @@ func componentTests(args ...string) {
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


amitkrout:
It is really hard to comprise all the improvement task in a single pr, so as per @kadel suggested i am split big task into multiple small tasks so that we can maximize the improvement/coverage. Details will be added in the issue #2256.

Issue #2073, opened 2019-09-03
Painfully slow `odo project list`


kadel:
`oc get projects` takes under the second, but `odo project list` is 13 seconds :-(

```
▶ gtime -f  "%es" oc get projects
NAME                                                    DISPLAY NAME   STATUS
default                                                                Active
kube-node-lease                                                        Active
kube-public                                                            Active
kube-system                                                            Active
oibwrltsgf                                                             Active
openshift                                                              Active
openshift-apiserver                                                    Active
openshift-apiserver-operator                                           Active
openshift-authentication                                               Active
openshift-authentication-operator                                      Active
openshift-cloud-credential-operator                                    Active
openshift-cluster-machine-approver                                     Active
openshift-cluster-node-tuning-operator                                 Active
openshift-cluster-samples-operator                                     Active
openshift-cluster-storage-operator                                     Active
openshift-cluster-version                                              Active
openshift-config                                                       Active
openshift-config-managed                                               Active
openshift-console                                                      Active
openshift-console-operator                                             Active
openshift-controller-manager                                           Active
openshift-controller-manager-operator                                  Active
openshift-dns                                                          Active
openshift-dns-operator                                                 Active
openshift-etcd                                                         Active
openshift-image-registry                                               Active
openshift-infra                                                        Active
openshift-ingress                                                      Active
openshift-ingress-operator                                             Active
openshift-insights                                                     Active
openshift-kube-apiserver                                               Active
openshift-kube-apiserver-operator                                      Active
openshift-kube-controller-manager                                      Active
openshift-kube-controller-manager-operator                             Active
openshift-kube-scheduler                                               Active
openshift-kube-scheduler-operator                                      Active
openshift-machine-api                                                  Active
openshift-machine-config-operator                                      Active
openshift-marketplace                                                  Active
openshift-monitoring                                                   Active
openshift-multus                                                       Active
openshift-network-operator                                             Active
openshift-node                                                         Active
openshift-operator-lifecycle-manager                                   Active
openshift-operators                                                    Active
openshift-sdn                                                          Active
openshift-service-ca                                                   Active
openshift-service-ca-operator                                          Active
openshift-service-catalog-apiserver-operator                           Active
openshift-service-catalog-controller-manager-operator                  Active
rgzcsmmdch                                                             Active
uhrnluzzwm                                                             Active
0.73s

▶ gtime -f  "%es" odo project list
ACTIVE     NAME
           default
           kube-node-lease
           kube-public
           kube-system
[...]

PR #2123, opened 2019-09-12
Add development doc on machine readable output

cdrage:
Adds a development doc on how to create correct machine readable output
with a working example on how to implement it.

### DIFF
--- a/docs/development.adoc
+++ b/docs/development.adoc
@@ -448,6 +448,61 @@ in https://github.com/kadel/homebrew-odo[`kadel/homebrew-odo`].
 . Confirm the binaries are available in the GitHub release page.
 . Create a PR and update the file `build/VERSION` with the  latest version number.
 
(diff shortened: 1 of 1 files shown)


docs/development.adoc
```diff
@@ -448,6 +448,61 @@ in https://github.com/kadel/homebrew-odo[`kadel/homebrew-odo`].
 . Confirm the binaries are available in the GitHub release page.
 . Create a PR and update the file `build/VERSION` with the  latest version number.
 
+== Writing machine readable output code
+
+Here are some tips to consider when writing machine-readable output code.
+
+- Match similar Kubernetes / OpenShift API structures
+- Put as much information as possible within `Spec`
```
mik-dass: `Put as much relevant information as possible within Spec as needed by the other teams` to prevent #2073

PR #2130, opened 2019-09-16
Replace dump-init with go-init

mohammedzee1000:

This PR makes use of go-init instead of dumb-init to avoid having to maintain our own rpm for dumb-init

## How to test changes?

All standard workflows should pass

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -95,7 +95,7 @@ const (
 
 	// Default Image that will be used containing the supervisord binary and assembly scripts
 	// use getBoostrapperImage() function instead of this variable
-	defaultBootstrapperImage = "quay.io/openshiftdo/init:0.10.0"
+	defaultBootstrapperImage = "quay.io/openshiftdo/init:0.11.0"
 
 	// ENV variable to overwrite image used to bootstrap SupervisorD in S2I builder Image
 	bootstrapperImageEnvName = "ODO_BOOTSTRAPPER_IMAGE"
--- a/pkg/occlient/templates.go
+++ b/pkg/occlient/templates.go
@@ -83,15 +83,13 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
 							Ports: commonImageMeta.Ports,
 							// Run the actual supervisord binary that has been mounted into the container
 							Command: []string{
-								"/opt/odo/bin/dumb-init",
-								"--",
+								"/bin/bash",
+								"-c",
 							},
 							// Using the appropriate configuration file that contains the "run" script for the component.
 							// either from: /usr/libexec/s2i/assemble or /opt/app-root/src/.s2i/bin/assemble
 							Args: []string{
-								"/opt/odo/bin/supervisord",
-								"-c",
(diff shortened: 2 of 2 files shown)


kadel:
for reference: changes on odo-init-image - https://github.com/openshift/odo-init-image/pull/35

pkg/occlient/templates.go
```diff
@@ -83,15 +83,13 @@ func generateSupervisordDeploymentConfig(commonObjectMeta metav1.ObjectMeta, com
 							Ports: commonImageMeta.Ports,
 							// Run the actual supervisord binary that has been mounted into the container
 							Command: []string{
-								"/opt/odo/bin/dumb-init",
-								"--",
+								"/bin/bash",
```
kadel: we probably don't need `bash` here. Executing just `/opt/odo/bin/go-init` should be enough.

mohammedzee1000: Fixed

pkg/occlient/occlient.go
```diff
@@ -95,7 +95,7 @@ const (
 
 	// Default Image that will be used containing the supervisord binary and assembly scripts
 	// use getBoostrapperImage() function instead of this variable
-	defaultBootstrapperImage = "quay.io/openshiftdo/init:0.10.0"
+	defaultBootstrapperImage = "quay.io/mohammedzee1000/odo-init-image:go-init"
```
kadel: Just so we don' forget. This needs to point to  `quay.io/openshiftdo/init:0.11.0` once the https://github.com/openshift/odo-init-image/pull/35 and new version of the image is released.

mohammedzee1000: Yep, that is why i specifically marked that change in a separate commit, so that it can be dropped :)

kadel: `quay.io/openshiftdo/init:0.11.0` is built. @mohammedzee1000 please update it to this image

mohammedzee1000: updated

kadel:
```
[odo] error: build error: After retrying 2 times, Pull image still failed due to error: while pulling "docker://image-registry.openshift-image-registry.svc:5000/openshift/php@sha256:64a0c4fb363357761a9f5550eecda1476d7f737e5f1d0863919d1f60985811ba" as "image-registry.openshift-image-registry.svc:5000/openshift/php@sha256:64a0c4fb363357761a9f5550eecda1476d7f737e5f1d0863919d1f60985811ba": Error initializing source docker://image-registry.openshift-image-registry.svc:5000/openshift/php@sha256:64a0c4fb363357761a9f5550eecda1476d7f737e5f1d0863919d1f60985811ba: pinging docker registry returned: Get https://image-registry.openshift-image-registry.svc:5000/v2/: dial tcp 172.30.102.228:5000: connect: no route to host
```
/retest

mohammedzee1000:
s2i related error
```
[odo] subprocess exited with status 255
[odo] subprocess exited with status 255
[odo] error: build error: error building at STEP "RUN /usr/libexec/s2i/assemble": exit status 255
```
/retest

girishramnani:
looks good to me
/lgtm

Issue #2132, opened 2019-09-16
We shouldn't require an odo component / context for `odo catalog search`


cdrage:
[kind/bug]

## How did you run odo exactly?

```sh
github.com/openshift/odo  refactor-catalog-list-services ✗                                                                                                                 4h17m ⚑  ⍉
▶ ./odo catalog search component rails                      
 ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component
```

## Actual behavior

Returning with "current directory doesn't represet an odo component"

## Expected behavior

Shouldn't have to provide a component or context

## Any logs, error output, etc?

See above!

PR #2134, opened 2019-09-16
Remove context check for search

cdrage:
Removes the check for search.

To test:

```sh
odo catalog search component ruby
```

Closes https://github.com/openshift/odo/issues/2132


cdrage:
@mik-dass @girishramnani Test has been added here: https://github.com/openshift/odo/pull/2135 that will cover both the context and actual searching (it does the search without having a component created).

Let's merge this PR in first as it's needed for the other one :)

PR #2146, opened 2019-09-19
Improve look and feel of integration test doc

prietyc123:

Improve the look and feel of integration test portion of the document
No, However when I was going through the unit and integration test doc I found these minor improvements.

## How to test changes?

Verify it in the doc.


prietyc123:
Ping @boczkowska

PR #2185, opened 2019-09-26
Fixing no such file or directory issue

amitkrout:
## What is the purpose of this change? What does it change?

Fixing ```no such file or directory``` issue for scenario https://github.com/openshift/odo/blob/master/tests/integration/component.go#L579
## Was the change discussed in an issue?
fixes #2184 

## How to test changes?

make test-cmd-cmp

PR #2190, opened 2019-09-26
log stdout and stderr of tar command

kadel:
- logs stdout and stderr of the tar command
- logs the DeploymentConfig definition

This will make it easier to debug issues during the file transfer. And issues with DeploymentConfig.


amitkrout:
#2185 will fix the CI failure

PR #2198, opened 2019-09-27
Adds message to inform user to do odo push for storage changes to tak…

### LINKED ISSUE #2195: odo storage commands should mention the need of odo push for changes to take effect
[kind/bug]

- Operating System: all

1. Created storage using `odo storage create` 
2. Deleted storage using `odo storage delete`

## Actual behavior
1. In case of `odo storage create`, the storage doesn't show up on remote pod.
2. In case of `odo storage delete`, the storage doesn't get deleted from remote pod.

## Expected behavior
`odo storage create` and `odo storage delete` commands should inform the user about the need of `odo push` for the changes to take effect on remote pod.

dharmit:
…e effect

$Subject

fixes #2195

## How to test changes?

1. Create a storage using `odo storage create`. Upon successful creation, a message to do `odo push` should show up.
2. Delete a storage using `odo storage delete`. Upon successful deletion, a message to do `odo push` should show up.

### DIFF
--- a/pkg/odo/cli/storage/delete.go
+++ b/pkg/odo/cli/storage/delete.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"fmt"
+
 	"github.com/openshift/odo/pkg/config"
 	"github.com/openshift/odo/pkg/log"
 	appCmd "github.com/openshift/odo/pkg/odo/cli/application"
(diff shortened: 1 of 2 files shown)


pkg/odo/cli/storage/delete.go
```diff
@@ -77,6 +78,7 @@ func (o *StorageDeleteOptions) Run() (err error) {
 		}
 
 		log.Infof("Deleted storage %v from %v", o.storageName, o.localConfig.GetName())
+		log.Infof("Please use `odo push` command to delete the storage from the component")
```
mik-dass: This might look a bit confusing maybe something like 
`Please use `odo push` command to delete the storage from the cluster`?

dharmit: Makes sense. Thanks for the simpler recommendation!

PR #2209, opened 2019-09-30
fix --binary  flag when --context is not set

### LINKED ISSUE #1969: --binary flag doesn't work without --context flag while creating component
[kind/bug]

- Operating System: Supported

```
$ odo component create wildfly:latest wildfly --binary /Users/amit/go/src/github.com/openshift/odo/tests/examples/binary/java/wildfly/ROOT.war --app app1 --project myproject
 ✗  Rel: can't make /Users/amit/go/src/github.com/openshift/odo/tests/examples/binary/java/wildfly/ROOT.war relative to
```

## Actual behavior
See how to run section

## Expected behavior
--binary flag should not depend on --context flag

**EDIT:** --binary flag should not always depend on --context flag. For example creating a binary component in the working dir should not explicitly look for context flag. cc_ @kadel @girishramnani

kadel:
fixes #1969

## How to test changes?
```
cd /component/path
odo component create java:8 --binary /component/path/target/java.jar
```
should succeed

Previously it was failing. This is a valid command, as binary is inside the context.

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -490,10 +490,10 @@ func ValidateComponentCreateRequest(client *occlient.Client, componentSettings c
 	// Check to see if the catalog type actually exists
 	exists, err := catalog.ComponentExists(client, componentType, componentVersion)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to check component of type %s", componentType)
+		return errors.Wrapf(err, "failed to check component of type %s", componentType)
 	}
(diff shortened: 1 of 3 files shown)


dharmit:
@kadel I think we need to mention in the code (`odo create -h`?) and/or documentation about what's pointed out in https://github.com/openshift/odo/issues/1969#issuecomment-521230274, no?

PR #2211, opened 2019-09-30
Check that --binary is in --context folder

kadel:
odo create command should verify that binary is inside the context folder.

how to test:
- When the binary file is outside of the context folder the command should fail.
  - `odo create --binary ../file.jar`
  - `odo create --binary target/file.jar --context my-component`

- When the binary is in the context folder everything should be ok.
  - `odo create --binary target/file.jar`
  - `odo create --binary ../mycomponent/file.jar --context ../mycomponent`

fixes https://github.com/openshift/odo/issues/2266

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -471,13 +471,9 @@ func CheckComponentMandatoryParams(componentSettings config.ComponentSettings) e
 }
 
 // ValidateComponentCreateRequest validates request for component creation and returns errors if any
-// Parameters:
-//	componentSettings: Component settings
-//	isCmpExistsCheck: boolean to indicate whether or not error out if component with same name already exists
 // Returns:
 //	errors if any
-func ValidateComponentCreateRequest(client *occlient.Client, componentSettings config.ComponentSettings, isCmpExists bool, isCmpExistsCheck bool) (err error) {
-
+func ValidateComponentCreateRequest(client *occlient.Client, componentSettings config.ComponentSettings, contextDir string) (err error) {
 	// Check the mandatory parameters first
 	err = CheckComponentMandatoryParams(componentSettings)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/component/component.go
```diff
@@ -502,13 +498,6 @@ func ValidateComponentCreateRequest(client *occlient.Client, componentSettings c
 		return errors.Wrapf(err, "failed to check component of name %s", *componentSettings.Name)
 	}
 
-	// If component does not exist, create it
-	if isCmpExistsCheck {
-		if isCmpExists {
```
kadel: the function was always called with `isCmpExistsCheck = false`, so this was never executed, probably some relict from before

tests/integration/component.go
```diff
@@ -205,6 +205,29 @@ func componentTests(args ...string) {
 				project, "--binary", filepath.Join(context, "sb.jar"))
 		})
 
+		It("binary component should fail when --binary is not in --context folder", func() {
+			oc.ImportJavaIS(project)
+			helper.CopyExample(filepath.Join("binary", "java", "openjdk"), context)
+
+			newContext := helper.CreateNewContext()
+			defer helper.DeleteDir(newContext)
```
amitkrout: @kadel Does this line will be executed if in case line ```Expect(output).To(ContainSubstring("inside of the context directory"))``` fails ? If yes then we are fine, otherwise a context dir will be left behind each time.

kadel: Yes, defer is executed when the function where this is called ends. Even if tests end with en error when this `It` block ends the defer function will be called.

kadel:
/retest 

```
level=error msg="Error: Error waiting for instance (i-0ce7953e72aafea91) to become ready: timeout while waiting for state to become 'running' (last state: 'pending', timeout: 10m0s)"

```

cdrage:
Let's retest again and get this merged.

/test v4.2-integration

PR #2213, opened 2019-09-30
Updates environment variable output

cdrage:
Bolds the info message for outputting the environment variables as well
as changes the grammar.

### DIFF
--- a/pkg/odo/cli/component/common_link.go
+++ b/pkg/odo/cli/component/common_link.go
@@ -122,14 +122,36 @@ func (o *commonLinkOptions) run() (err error) {
 		return err
 	}
 
-	if o.operationName == "link" {
-		fmt.Printf("Following environment variables were added to %s component:\n", o.Component())
+	if len(secret.Data) == 0 {
+		log.Infof("There are no secret environment variables to expose within the %s service", o.suppliedName)
 	} else {
-		fmt.Printf("Following environment variables were removed from %s component:\n", o.Component())
-	}
+		if o.operationName == "link" {
+			log.Infof("The below secret environment variables were added to the '%s' component:\n", o.Component())
+		} else {
+			log.Infof("The below secret environment variables were removed from the '%s' component:\n", o.Component())
+		}
 
-	for i := range secret.Data {
-		fmt.Printf("- %v\n", i)
+		// Output the environment variables
+		for i := range secret.Data {
+			fmt.Printf("· %v\n", i)
+		}
+
+		// Retrieve the first variable to use as an example.
+		// Have to use a range to access the map
+		var exampleEnv string
+		for i := range secret.Data {
+			exampleEnv = i
+			break
+		}
+
+		// Output what to do next if first linking...
+		// TODO: Retrieve the podName to add to the example?
+		if o.operationName == "link" {
+			log.Infof(`
+You can now access the environment variables from within the component pod, for example:
+oc rsc <pod-name>
+echo $%s`, exampleEnv)
+		}
 	}
 
 	if o.wait {
--- a/pkg/odo/cli/component/common_link.go
+++ b/pkg/odo/cli/component/common_link.go
@@ -145,12 +145,10 @@ func (o *commonLinkOptions) run() (err error) {
 		}
 
 		// Output what to do next if first linking...
-		// TODO: Retrieve the podName to add to the example?
 		if o.operationName == "link" {
 			log.Infof(`
 You can now access the environment variables from within the component pod, for example:
-oc rsc <pod-name>
-echo $%s`, exampleEnv)
+$%s is now available as a variable within component %s`, exampleEnv, o.Component())
 		}
 	}
 
--- a/tests/integration/servicecatalog/cmd_link_unlink_test.go
+++ b/tests/integration/servicecatalog/cmd_link_unlink_test.go
@@ -199,13 +199,13 @@ var _ = Describe("odo link and unlink command tests", func() {
 
 			// tests for linking a component to a component
 			stdOut := helper.CmdShouldPass("odo", "link", "component2", "--context", context1)
-			Expect(stdOut).To(ContainSubstring("Following environment variables were added"))
+			Expect(stdOut).To(ContainSubstring("The below secret environment variables were added"))
 			Expect(stdOut).To(ContainSubstring("COMPONENT_COMPONENT2_HOST"))
 			Expect(stdOut).To(ContainSubstring("COMPONENT_COMPONENT2_PORT"))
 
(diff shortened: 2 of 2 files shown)


cdrage:
Thanks @dharmit :) fixed!

cdrage:
Changed the output a bit, let me know what you think @kadel @girishramnani 

```sh
~/nodejs-ex  master ✗                                                                                                                                                                                                                                                                                                                                              351d ◒  ⍉
▶ odo unlink dh-mysql-apb                                       
 ✓  Service dh-mysql-apb has been successfully unlinked from the component frontend

The below secret environment variables were removed from the 'frontend' component:

· DB_USER
· DB_HOST
· DB_NAME
· DB_PASSWORD
· DB_PORT
· DB_TYPE

~/nodejs-ex  master ✗                                                                                                                                                                                                                                                                                                                                               351d ◒  
▶ odo link dh-mysql-apb  
 ✓  Service dh-mysql-apb has been successfully linked from the component frontend

The below secret environment variables were added to the 'frontend' component:

· DB_PASSWORD
· DB_PORT
· DB_TYPE
· DB_USER
· DB_HOST
· DB_NAME

You can now access the environment variables from within the component pod, for example:
oc rsc <pod-name>
echo $DB_NAME
```

dharmit:
>```
>You can now access the environment variables from within the component pod, for example:
>oc rsc <pod-name>
>echo $DB_NAME
>```

IMHO, this will be confusing to the users who're not familiar with OpenShift terminology. I think we're trying to abstract most things about OpenShift/oc via odo. A user who's not aware of what a `pod` means, would wonder what's going on, no? For a user using odo via VS Code adapters, we will have to choose different message (in json output, I guess) because they're likely not aware of what odo or oc is.

Besides that, `oc rsc <pod-name>` is a command that one has to execute from the development system, right? The statement `You can now access the environment variables from within the component pod, for example` made me think that it's a command that could be executed from inside the pod. But it could be just me.

cdrage:
@dharmit Yeah, I thought the same thing too, but wasn't too sure. I'll remove that output.

When we merge in https://github.com/openshift/odo/pull/2233 we can simply say use `odo describe` to see what environment variables are available.

cdrage:
It will now look like this:

```sh
You can now access the environment variables from within the component pod, for example:
$DB_NAME is now available within component dh-mysql-apb
```

kadel:
> ```shell
> ▶ odo link dh-mysql-apb  
>  ✓  Service dh-mysql-apb has been successfully linked from the component frontend
> 
> The below secret environment variables were added to the 'frontend' component:
> 
> · DB_PASSWORD
> · DB_PORT
> · DB_TYPE
> · DB_USER
> · DB_HOST
> · DB_NAME
> 
> You can now access the environment variables from within the component pod, for example:
> $DB_NAME is now available within component dh-mysql-apb
> ```

The last statement is not true. ` $DB_NAME is now available within component dh-mysql-apb` it should be ` $DB_NAME is now available within component nodejs-something`.
You are linking nodejs component with mysql. So the environment variables are added to the nodejs component not mysql


/hold

pkg/odo/cli/component/common_link.go
```diff
@@ -122,14 +122,35 @@ func (o *commonLinkOptions) run() (err error) {
...
+
+		// Output the environment variables
+		for i := range secret.Data {
+			fmt.Printf("· %v\n", i)
+		}
+
+		// Retrieve the first variable to use as an example.
+		// Have to use a range to access the map
+		var exampleEnv string
+		for i := range secret.Data {
```
girishramnani: why use a range? why not just `secret.Data[0]` it? I already see `secret.Data == 0` check which means `Data` array should have atleast one element.

pkg/odo/cli/component/common_link.go
```diff
@@ -122,14 +122,35 @@ func (o *commonLinkOptions) run() (err error) {
...
+		}
 
-	for i := range secret.Data {
-		fmt.Printf("- %v\n", i)
+		// Output what to do next if first linking...
+		// TODO: Retrieve the podName to add to the example?
+		if o.operationName == "link" {
+			log.Infof(`
+You can now access the environment variables from within the component pod, for example:
+$%s is now available as a variable within component %s`, exampleEnv, o.Component())
```
girishramnani: @kadel the example might be incorrect, but the code mentions `o.Component()`

girishramnani:
Th

> > ```shell
> > ▶ odo link dh-mysql-apb  
> >  ✓  Service dh-mysql-apb has been successfully linked from the component frontend
> > 
> > The below secret environment variables were added to the 'frontend' component:
> > 
> > · DB_PASSWORD
> > · DB_PORT
> > · DB_TYPE
> > · DB_USER
> > · DB_HOST
> > · DB_NAME
> > 
> > You can now access the environment variables from within the component pod, for example:
> > $DB_NAME is now available within component dh-mysql-apb
> > ```
> 
> The last statement is not true. ` $DB_NAME is now available within component dh-mysql-apb` it should be ` $DB_NAME is now available within component nodejs-something`.
> You are linking nodejs component with mysql. So the environment variables are added to the nodejs component not mysql
> 
> /hold

@cdrage the issue mentioned by tomas seems to be happening because the first argument is taken as component name so you might have to get the component name from the localConfigInfo or a better option might be to make `o.Component()` populate with the correct value.

pkg/odo/cli/component/common_link.go
cdrage: secret.Data is a map not an array, and that's why you have to use a range to get the first variable.

girishramnani: Hehe, felt confusing.

cdrage:
@kadel @girishramnani 

The output was actually incorrect.. some copy-and-paste foo must of messed it up, but here is the correct output:

```sh
▶ odo link mongodb-persistent                                      
 ✓  Service mongodb-persistent has been successfully linked from the component nodejs-nodejs-ex-eapl

The below secret environment variables were added to the 'nodejs-nodejs-ex-eapl' component:

· admin_password
· database_name
· password
· uri
· username

You can now access the environment variables from within the component pod, for example:
$username is now available as a variable within component nodejs-nodejs-ex-eapl
```

girishramnani:
From the code, it seemed that component name was populated from the first arg, I could be wrong

cdrage:
@girishramnani See line: https://github.com/openshift/odo/blob/083bf1852627793927304f24a61cf21b3011f6ca/pkg/odo/cli/component/common_link.go#L118 previously used in the code for `o.Component()`.

Was simply a bad copy and paste from me. The code is correct :)

PR #2218, opened 2019-10-01
Bump odo version to 1.0.0-beta6 in "build/VERSION" file

dharmit:

$subject


dharmit:
/test v4.1-e2e-scenarios

```
 [odo]  ✓  Triggering build from git [80ms]
[odo]  •  Waiting for build to finish  ...
[odo]  ✗  Waiting for build to finish [30s]
[odo]  ✗  Failed to create component with name wo-wait-javaee-git-test. Please use `odo config view` to view settings used to create component. Error: unable to wait for build wo-wait-javaee-git-test-app-1 to run: timed out waiting for the conditio
```

amitkrout:
```
[odo]  ✗  Waiting for build to finish [30s]
[odo]  ✗  Failed to create component with name wo-wait-javaee-git-test. Please use `odo config view` to view settings used to create component. Error: unable to wait for build wo-wait-javaee-git-test-app-1 to run: timed out waiting for the condition
```

@dharmit I have a doubt here. 
We have already release beta6 version and now the build VERSION file is changed to beta6. So don't you think the bits we released is not the proper beta6 then ? may be some metadata which depends on this file is not updated in the release bits.

dharmit:
I don't think that would be the reason behind the time out error here. @kadel WDYT?

amitkrout:
> I don't think that would be the reason behind the time out error here. @kadel WDYT?

@dharmit Yup i understand, That was just a failure (flake) message which i pasted above of my comment. Can you please address which i asked after that

PR #2221, opened 2019-10-01
Fixes project wait flag

### LINKED ISSUE #2117: Project create doesn't wait properly
https://travis-ci.com/openshift/odo/jobs/233964101#L2793

```
Running odo with args [odo project list]
[odo] ACTIVE     NAME
[odo] *          kdiadkbatt
[odo]            myproject
Running odo with args [odo project list]
[odo] ACTIVE     NAME
[odo]            kbcwtnzrez
[odo] *          kdiadkbatt
[odo]            myproject
```
I had applied a sleep operation of 1 sec in between the two list operations and the create was performed with `-w` flag

mik-dass:
fixes #2117 

How to test:
Check if the `odo project create <project-name> -w` works and it waits properly for the project to be ready.

### DIFF
--- a/tests/integration/cmd_project_test.go
+++ b/tests/integration/cmd_project_test.go
@@ -41,7 +41,6 @@ var _ = Describe("odo project command tests", func() {
 	Context("when running project command app parameter in directory that doesn't contain .odo config directory", func() {
 		It("should successfully execute list along with machine readable output", func() {
 			helper.CmdRunner("odo", "project", "list")
-			time.Sleep(1 * time.Second)
 			listOutput := helper.CmdShouldPass("odo", "project", "list")
 			Expect(listOutput).To(ContainSubstring(project))
 
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ test-cmd-service:
 # Run odo project command tests
 .PHONY: test-cmd-project
 test-cmd-project:
-	ginkgo $(GINKGO_FLAGS) -focus="odo project command tests" tests/integration/
+	ginkgo $(GINKGO_FLAGS_SERIAL) -focus="odo project command tests" tests/integration/project/
 
 # Run odo app command tests
 .PHONY: test-cmd-app
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -523,8 +523,9 @@ func (c *Client) CreateNewProject(projectName string, wait bool) error {
 	// If watched is created after the project it can lead to situation when the project is created before the watcher.
 	// When this happens, it gets stuck waiting for event that already happened.
 	var watcher watch.Interface
+	var err error
 	if wait {
-		watcher, err := c.projectClient.Projects().Watch(metav1.ListOptions{
+		watcher, err = c.projectClient.Projects().Watch(metav1.ListOptions{
 			FieldSelector: fields.Set{"metadata.name": projectName}.AsSelector().String(),
 		})
 		if err != nil {
@@ -538,7 +539,7 @@ func (c *Client) CreateNewProject(projectName string, wait bool) error {
 			Name: projectName,
 		},
 	}
-	_, err := c.projectClient.ProjectRequests().Create(projectRequest)
+	_, err = c.projectClient.ProjectRequests().Create(projectRequest)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create new project %s", projectName)
 	}
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -2304,7 +2304,7 @@ func TestCreateNewProject(t *testing.T) {
 						},
 					})
 				}()
-				fkclientset.ProjClientset.AddWatchReactor("projects", func(action ktesting.Action) (handled bool, ret watch.Interface, err error) {
+				fkclientset.ProjClientset.PrependWatchReactor("projects", func(action ktesting.Action) (handled bool, ret watch.Interface, err error) {
 					if len(tt.projName) == 0 {
 						return true, nil, fmt.Errorf("error watching project")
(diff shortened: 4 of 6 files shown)


tests/integration/cmd_project_test.go
```diff
@@ -41,13 +41,12 @@ var _ = Describe("odo project command tests", func() {
...
 		It("should successfully execute list along with machine readable output", func() {
 			helper.CmdRunner("odo", "project", "list")
-			time.Sleep(1 * time.Second)
 			listOutput := helper.CmdShouldPass("odo", "project", "list")
 			Expect(listOutput).To(ContainSubstring(project))
 
 			// project deletion doesn't happen immediately, so we test subset of the string
 			listOutputJson := helper.CmdShouldPass("odo", "project", "list", "-o", "json")
-			Expect(listOutputJson).To(ContainSubstring(`{"kind":"Project","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"` + project + `","namespace":"` + project + `","creationTimestamp":null},"spec":{},"status":{"active":false}}`))
+			Expect(listOutputJson).To(ContainSubstring(`{"kind":"Project","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"` + project + `","namespace":"` + project + `","creationTimestamp":null},"spec":{},"status":{"active":true}}`))
```
amitkrout: In parallel run validating the projects status ```"status":{"active":true}}``` or ```"status":{"active":true}}``` could lead to a flaky behavior.  Instead of matching the full json can you match some major key value pair.

For now see how this is being done in one of our test cases - https://github.com/openshift/odo/blob/master/tests/integration/component.go#L197

tests/integration/cmd_project_test.go
```diff
@@ -41,13 +41,12 @@ var _ = Describe("odo project command tests", func() {
 	Context("when running project command app parameter in directory that doesn't contain .odo config directory", func() {
 		It("should successfully execute list along with machine readable output", func() {
 			helper.CmdRunner("odo", "project", "list")
-			time.Sleep(1 * time.Second)
```
amitkrout: Cool!!!

amitkrout: Also remove ```helper.CmdRunner("odo", "project", "list")```. In the next line it is being duplicated.

mik-dass:
```
✗  Waiting for component to start [4m]
[odo]  ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=mycomponent-myapp'
```
/retest

mik-dass:
@amitkrout Done

tests/integration/cmd_project_test.go
```diff
@@ -40,14 +40,12 @@ var _ = Describe("odo project command tests", func() {
...
 		It("should successfully execute list along with machine readable output", func() {
-			helper.CmdRunner("odo", "project", "list")
-			time.Sleep(1 * time.Second)
 			listOutput := helper.CmdShouldPass("odo", "project", "list")
 			Expect(listOutput).To(ContainSubstring(project))
 
 			// project deletion doesn't happen immediately, so we test subset of the string
 			listOutputJson := helper.CmdShouldPass("odo", "project", "list", "-o", "json")
-			Expect(listOutputJson).To(ContainSubstring(`{"kind":"Project","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"` + project + `","namespace":"` + project + `","creationTimestamp":null},"spec":{},"status":{"active":false}}`))
+			Expect(listOutputJson).To(ContainSubstring(`{"kind":"Project","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"` + project + `","namespace":"` + project + `","creationTimestamp":null},"spec":{},`))
```
cdrage: Shouldn't this be status: active? Why was it removed? This is obvious that the test is failing.. In this test scenario, you can modify the `beforeEach` to add `-w` to wait for the project to come up.

amitkrout: > Shouldn't this be status: active? Why was it removed? This is obvious that the test is failing..

@cdrage https://github.com/openshift/odo/pull/2221#discussion_r330991328

Think a case where you have more scenarios are running for example ```make test-integration``` then in two node you cannot be sure that the current scenario project is active throughout the scenario. May be in the same time some different project scenario is created and become active.

> In this test scenario, you can modify the beforeEach to add -w to wait for the project to come up

In the test script we have already made this behavior as default to avoid flakes. see - https://github.com/openshift/odo/blob/master/tests/helper/odo_utils.go#L79. The failure was happening due to the step ```CmdRunner(...)``` which does not wait for command output.

mik-dass: As @amitkrout pointed due to parallel tests we can't be sure of the activeness of the test throughout the scenario so I removed the status. @cdrage @kadel WDYT?

cdrage: Ah, but that defeats the whole purpose of the test though, to make sure that it's the active project.

amitkrout: yup, there is an option of sequential run. In a long run it would be hard to even manage ```get``` and ```set``` command in parallel run. @mik-dass Please do the change to run it on single node test.

For reference check - https://github.com/openshift/odo/tree/master/tests/integration/loginlogout.

mik-dass:
@cdrage @amitkrout Done please have a look

amitkrout:
> @cdrage @amitkrout Done please have a look

Looks good to me.

mik-dass:
```
[oc] Error in configuration: 
[oc] * context was not found for specified context: yfkoarskib/api-ci-op-r1xy5kvm-1b2f4-origin-ci-int-aws-dev-rhcloud-com:6443/developer
[oc] * cluster has no server defined
```
/retest

PR #2228, opened 2019-10-02
Machine output for `odo app list` should be `Kind: List`

cdrage:
Machine readable output for ApplicationList Kind should actually just be
`Kind`.


amitkrout:
Test has been updated
/lgtm

PR #2230, opened 2019-10-02
ComponentTypeList should be List in machine output

cdrage:
For `odo catalog list components` machine-readable output for `Kind`
should simply be `List`

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -26,7 +26,7 @@ func ListComponents(client *occlient.Client) (ComponentTypeList, error) {
 
 	return ComponentTypeList{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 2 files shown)


dharmit:
@cdrage don't we have tests that check for the json output of `odo catalog list -o json`? Should we not have it?

cc @amitkrout

amitkrout:
> @cdrage don't we have tests that check for the json output of `odo catalog list -o json`? Should we not have it?
> 
> cc @amitkrout
> 
> UPDATE: We do have tests for ServiceTypeList #2229. I think we should add for ComponentTypeList as well. 😄

Test has been updated for this pr - https://github.com/openshift/odo/pull/2230/files#diff-f0f2e54d32d2c1ab57020c3a3dd9a3fbR197. 

/approve
/lgtm

PR #2231, opened 2019-10-02
Updates `odo describe` output

cdrage:
Updates the `odo describe` output to be a bit more intuitive / better
looking.

### DIFF
--- a/pkg/log/status.go
+++ b/pkg/log/status.go
@@ -249,6 +249,16 @@ func Infof(format string, a ...interface{}) {
 	}
 }
 
+// Describef will print out the first variable as BOLD and then the second not..
+// this is intended to be used with `odo describe` and other outputs that list
+// a lot of information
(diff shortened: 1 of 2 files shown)


cdrage:
See the comparison below:


### New

Screenshot from 2019-10-02 15-43-01


### Old

Screenshot from 2019-10-02 15-44-15

girishramnani:
Looks the same on windows?

girishramnani:
as a follow up show we change the output of `app describe` and `catalog describe` to the same **Bold**  headers?

cdrage:
@girishramnani Yes, we should change it in `app describe` and `catalog describe` too. But that'll be for a different PR. We should get this one merged in first.

girishramnani:
Cool,
/approve

girishramnani:
An issue to track the follow up task would be great @cdrage

Issue #2232, opened 2019-10-02
`odo describe` should also show secrets / linked variables from other components and services


cdrage:
[kind/Enhancement]

## Which functionality do you think we should update/improve?

When we use `odo describe`:

```sh
Component Name: nodejs-nodejs-ex-cstb
Type: nodejs
Source: file://./
Environment Variables:
 · foo=bar
 · bar=foo
 · DEBUG_PORT=5858
Storage:
 · mystorage of size 1Gi mounted to /opt/app-root/src/storage/
URLs:
 · http://foobar-app-myproject.192.168.42.79.nip.io exposed via 8080
Linked Components:
 · nodejs2 - Port(s): 8080
Linked Services:
 · mongodb-persistent
 · dh-mysql-apb
```

We should also link what secrets have been passed in via `Linked Services`

## Why is this needed? 

More information / easier to add / edit environment variables for front-end applications.

cdrage:
PR open here: https://github.com/openshift/odo/pull/2233

PR #2233, opened 2019-10-02
Add secrets to `odo describe`

cdrage:
Adds the "secrets" or environment variables that have been passed into
the component by using the `odo link` feature when connecting to a
database.

Builds upon PR: https://github.com/openshift/odo/pull/2231
Closes: https://github.com/openshift/odo/issues/2232

See the below example of how it now looks like:

```sh
Component Name: nodejs-nodejs-ex-cstb
Type: nodejs
Source: file://./
Environment Variables:
 · foo=bar
 · bar=foo
 · DEBUG_PORT=5858
Storage:
 · mystorage of size 1Gi mounted to /opt/app-root/src/storage/
URLs:
 · http://foobar-app-myproject.192.168.42.79.nip.io exposed via 8080
Linked Components:
 · nodejs2 - Port(s): 8080
Linked Services:
 · mongodb-persistent
   Environment Variables:
    · uri
    · username
    · admin_password
    · database_name
    · password
 · dh-mysql-apb
   Environment Variables:
    · DB_HOST
    · DB_NAME
    · DB_PASSWORD
    · DB_PORT
    · DB_TYPE
    · DB_USER
```

### DIFF
--- a/pkg/odo/util/cmdutils.go
+++ b/pkg/odo/util/cmdutils.go
@@ -71,7 +71,7 @@ func CheckOutputFlag(outputFlag string) error {
 }
 
 // PrintComponentInfo prints Component Information like path, URL & storage
-func PrintComponentInfo(client *occlient.Client, currentComponentName string, componentDesc component.Component, applicationName string) {
+func PrintComponentInfo(client *occlient.Client, currentComponentName string, componentDesc component.Component, applicationName string, project string) {
 	localConfig, err := config.New()
 	if err != nil {
 		LogErrorAndExit(err, "")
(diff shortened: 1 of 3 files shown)


girishramnani:
As secrets being **secret** , should we actually print them?

cdrage:
@girishramnani They're just the environment variables. We don't print what they are.

pkg/odo/util/cmdutils.go
```diff
@@ -162,7 +162,24 @@ func PrintComponentInfo(client *occlient.Client, currentComponentName string, co
...
+
+			// Let's also get the secrets / environment variables that are being passed in.. (if there are any)
+			secrets, err := client.GetSecret(linkedService, project)
+			LogErrorAndExit(err, "")
+
+			if len(secrets.Data) > 0 {
+				// Iterate through the secrets to throw in a string
+				var secretOutput string
+				for i := range secrets.Data {
+					secretOutput += fmt.Sprintf("    · %v\n", i)
```
girishramnani: should we use `\t` here instead of providing spaces?

cdrage: Nope, spaces will keep it consistent in output on Windows + Linux.

dharmit: Wouldn't using `\t` keep the output consistent and predictable? :thinking:

girishramnani: Yeah, that was my thinking

girishramnani: But we can merge it

PR #2238, opened 2019-10-03
Removes app,project and component flag from storage commands

### LINKED ISSUE #2203: "odo storage create" with "--app" and "--project" flags is broken
[kind/bug]

- Operating System: all

Created storage with `--app` and `--project` flags where the values of these flags were different from the values in component directory. Also, these values were invalid because the `app` and `project` with those names don't exist.

```sh
$ odo config view
COMPONENT SETTINGS
Type              nodejs
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP
Name              nodejs-nodejs-ex-vaiv
MinMemory         
MaxMemory         
DebugPort         
Ignore            
MinCPU            
MaxCPU

# we now create storage that doesn't belong to the app/project in above config
$ odo storage create invalidstorage \
> --path=/opt/newstorage \
> --size=1Gi \
> --app newapp \
> --project newproject
 ✓  Added storage invalidstorage to nodejs-nodejs-ex-vaiv

$ odo push # works just fine.

$ odo storage list
The component 'nodejs-nodejs-ex-vaiv' has the following storage attached:
NAME               SIZE     PATH
invalidstorage     1Gi      /opt/newstorage

# inside the remote pod
sh-4.2$ ls /opt
app-root  newstorage  odo  rh

```

## Actual behavior
1. `odo storage create` command reports no error
2. `odo push` creates a storage for the component

## Expected behavior
`odo  storage create` should report some error indicating that the `app`/`project` are invalid. Creating the `app`/`project` upon `odo push` after `odo storage create` doesn't make sense because a storage is created for a component.

mik-dass:
fixes #2203 

How to test:
Try the app,project and component flags with all storage commands. ODO will complain of unknown flags for them.

### DIFF
--- a/pkg/odo/cli/storage/create.go
+++ b/pkg/odo/cli/storage/create.go
@@ -12,10 +12,6 @@ import (
 	"github.com/openshift/odo/pkg/util"
 	"github.com/spf13/cobra"
 	ktemplates "k8s.io/kubernetes/pkg/kubectl/cmd/templates"
-
-	appCmd "github.com/openshift/odo/pkg/odo/cli/application"
-	componentCmd "github.com/openshift/odo/pkg/odo/cli/component"
-	projectCmd "github.com/openshift/odo/pkg/odo/cli/project"
 )
 
 const createRecommendedCommandName = "create"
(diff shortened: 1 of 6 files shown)


amitkrout:
#2243

Issue #2243, opened 2019-10-04
Removing --app, --project, --component from subcommands that depends on component dir


amitkrout:
Copied from our last cabal discussion 

- we should remove --project, --app, --component flags from url and storage

- we need to check with vscode adapter that they are not using it

- lets check if there are more commands where we need to remove those flags

amitkrout:
- [ ] Storage command - #2238 

- [ ] Url command -

amitkrout:
ping @kadel @dgolovin

girishramnani:
Intrested in doing the url one, if someone hasn't started 
cc @mik-dass

mik-dass:
> Intrested in doing the url one, if someone hasn't started
cc @mik-dass

@girishramnani There is already a PR open for it https://github.com/openshift/odo/pull/2247

girishramnani:
I opened it :smile:

mik-dass:
> I opened it smile

Yeah I know, thanks for that :D

kadel:
We need a list of the commands from which the flags should be removed!

kadel:
> * we need to check with vscode adapter that they are not using it
> *

did someone check this?

girishramnani:
The only commands that have the above flags is link, unlink and service stuff

girishramnani:
ping @dgolovin 

> * we need to check with vscode adapter that they are not using it

kadel:
There is `odo.ts` file inside the vscode plug source that contains all the odo commands execute by plugin in place.

You can look there.

I just did a quick check. `url` commands look clean. But there is a `odo storage` command using all 3 of them  https://github.com/redhat-developer/vscode-openshift-tools/blob/8388b83e750c686b773fed91e6275da425675576/src/odo.ts#L139

girishramnani:
and the `odo storage` PR already got merged in, which means if we do a release the vscode plugin would break? @kadel

kadel:
> and the `odo storage` PR #2238 already got merged in, which means if we do a release the vscode plugin would break? @kadel

yes!

girishramnani:
then we should probably tell them :sweat_smile:

mohitsuman:
We will update the `odo storage` command to use `--context`. This goes for beta-7 release, right ?

girishramnani:
yes

girishramnani:
@amitkrout I went through the code and it seems these are the only places from where we need to remove --project, --app and --component commands.
Closing this issue as we have already 
[...]

Issue #2204, opened 2019-09-30
"odo url create" with "--app" and "--project" flags is broken


dharmit:
Creating a URL using `--app` and `--project` flags from a component directory creates a URL for the component in current working directory.
```sh
$ odo config view
COMPONENT SETTINGS
------------------------------------------------
PARAMETER         CURRENT_VALUE
Type              nodejs
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP
Name              nodejs-nodejs-ex-vaiv
MinMemory         
MaxMemory         
DebugPort         
Ignore            
MinCPU            
MaxCPU 

$ odo url create myurl --app nosuchapp --project nosuchproject
 ✓  URL myurl created for component: nodejs-nodejs-ex-vaiv

To create URL on the OpenShift Cluster, please use `odo push`

$ odo url list
Found the following URLs for component nodejs-nodejs-ex-vaiv in application app:
NAME      STATE          URL     PORT
myurl     Not Pushed     ://     8080

There are local changes. Please run 'odo push'.
$ odo push 
# push was successful and url got created for the component in pwd
```

_Originally posted by @dharmit in https://github.com/openshift/odo/issues/2203#issuecomment-536456785_

amitkrout:
#2247

amitkrout:
closed via #2247 
/close

PR #2247, opened 2019-10-04
remove --app, --component and --project from odo url

girishramnani:
see $SUBJECT

part of https://github.com/openshift/odo/issues/2243

## How to test changes?
```
odo url create --help
```
shouldn't have the flags

### DIFF
--- a/pkg/odo/cli/url/create.go
+++ b/pkg/odo/cli/url/create.go
@@ -32,9 +32,6 @@ var (
   
 	# Create a URL with a specific name and port
 	%[1]s example --port 8080
-  
-	# Create a URL with a specific name and port for component frontend
-	%[1]s example --port 8080 --component frontend
 	  `)
 )
 
--- a/pkg/odo/cli/url/url.go
+++ b/pkg/odo/cli/url/url.go
@@ -3,9 +3,6 @@ package url
 import (
 	"fmt"
 
-	appCmd "github.com/openshift/odo/pkg/odo/cli/application"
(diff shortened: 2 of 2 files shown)


kadel:
Shouldn't we also remove all the logic that is using those flags?

girishramnani:
Sure, But I think --component flag is still used in link and unlink and service catalog uses the other two

girishramnani:
Interestingly the logic related to those flags has been already removed

girishramnani:
so after going through the code, I figured out the reason. The `--component`, `--app` and `--project` flags are populated when we initialise a `NewContext`.
https://github.com/openshift/odo/blob/4eee157aa9eb5d660d79b516a7e18328b6728d46/pkg/odo/genericclioptions/context.go#L254-L277

This code is common among all commands and as `odo link, unlink and service` still use these flags we cannot remove that code.
@kadel

kadel:
> so after going through the code, I figured out the reason. The `--component`, `--app` and `--project` flags are populated when we initialise a `NewContext`.
> https://github.com/openshift/odo/blob/4eee157aa9eb5d660d79b516a7e18328b6728d46/pkg/odo/genericclioptions/context.go#L254-L277
> 
> This code is common among all commands and as `odo link, unlink and service` still use these flags we cannot remove that code.
> @kadel

ok, that makes sense

girishramnani:
This PR removes flags which might affect other projects dependent on odo so added a label so the release manager adds this PR in the release notes

amitkrout:
#2204

Issue #2139, opened 2019-09-18
"odo config unset" doesn't unset an environment variable but says it did


dharmit:
[kind/bug]

## What versions of software are you using?
- Operating System: all
- Output of `odo version`: `master`

## How did you run odo exactly?
1. Set an env var using `odo config set --env`. 
2. Unset it using `odo config unset --env foo=bar`. 
3. Unset it using `odo config unset --env foo` (don't mention the value of `foo` this time).
Verify output of `odo config view` at the end of each step
```sh
$ odo config set --env foo=bar
Environment variables were successfully updated.
Run `odo push --config` command to apply changes to the cluster.

$ odo config view
ENVIRONMENT VARIABLES
------------------------------------------------
NAME    VALUE
foo     bar

COMPONENT SETTINGS
------------------------------------------------
PARAMETER         CURRENT_VALUE
Type              nodejs
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP
Name              nodejs-nodejs-ex-ulzb
MinMemory         
MaxMemory         
Ignore            
MinCPU            
MaxCPU           

$ odo config unset --env foo=bar
Environment variables were successfully updated.

$ odo config view
ENVIRONMENT VARIABLES
------------------------------------------------
NAME    VALUE
foo     bar        <------------------- This still shows!

COMPONENT SETTINGS
------------------------------------------------
PARAMETER         CURRENT_VALUE
Type              nodejs
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP
Name              nodejs-nodejs-ex-ulzb
MinMemory         
MaxMemory         
Ignore            
MinCPU            
MaxCPU            

$ odo config unset --env foo
Environment variables were successfully updated.

$ odo config view
COMPONENT SETTINGS
------------------------------------------------
PARAMETER         CURRENT_VALUE
Type              nodejs
Application       app
Project           myproject
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP
Name              nodejs-nodejs-ex-ulzb
MinMemory         
MaxMemory         
Ignore            
MinCPU            
MaxCPU            
```

## Actual behavior
`odo` says that it has successfully unset the env var but it actually hasn't.

## Expected behavior
Either `odo config unset` should complain that it can't handle the input given by user or unset the value as user would expect

amitkrout:
[...]

PR #2271, opened 2019-10-10
Adds ability to use `odo config unset --env foo=bar`

cdrage:
This PR:
- Adds ability to use `foo=bar` when unsetting an environment variable
- Tests have been added

To test:
```sh
git clone https://github.com/openshift/nodejs-ex && cd ~/nodejs-ex
odo create
odo push
odo config set --env foo=bar
odo push  # optional
odo config view
odo config unset --env foo=bar
odo push  # optional
odo config view
```

Closes https://github.com/openshift/odo/issues/2139

### DIFF
--- a/tests/integration/cmd_pref_config_test.go
+++ b/tests/integration/cmd_pref_config_test.go
@@ -248,6 +248,7 @@ var _ = Describe("odo preference and config command tests", func() {
 			os.Unsetenv("GLOBALODOCONFIG")
 			helper.DeleteDir(context)
 		})
+
 		It("should set and unset env variables", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project, "--context", context)
 			helper.CmdShouldPass("odo", "config", "set", "--env", "PORT=4000", "--env", "PORT=1234", "--context", context)
@@ -262,6 +263,23 @@ var _ = Describe("odo preference and config command tests", func() {
 			Expect(configValue).To(Not(ContainSubstring(("PORT"))))
 			Expect(configValue).To(Not(ContainSubstring(("SECRET_KEY"))))
 		})
+
+		It("should be able to unset an environment variable with foo=bar", func() {
+
+			// Create the project
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project, "--context", context)
+
+			// Test out passing in with an equals sign
+			helper.CmdShouldPass("odo", "config", "set", "--env", "FOOBAR=foobar", "--context", context)
+			output := helper.CmdShouldPass("odo", "config", "view", "--context", context)
+			Expect(output).To(ContainSubstring("FOOBAR"))
+
+			// Check that we can remove it
+			helper.CmdShouldPass("odo", "config", "unset", "--env", "FOOBAR=foobar", "--context", context)
+			output = helper.CmdShouldPass("odo", "config", "view", "--context", context)
+			Expect(output).NotTo(ContainSubstring("FOOBAR"))
+		})
+
 	})
 
 	Context("when viewing local config without logging into the OpenShift cluster", func() {
--- a/pkg/config/env_var.go
+++ b/pkg/config/env_var.go
@@ -85,6 +85,8 @@ func NewEnvVarListFromSlice(envList []string) (EnvVarList, error) {
 // RemoveEnvVarsFromList removes the env variables based on the keys provided
 // and returns a new EnvVarList
 func RemoveEnvVarsFromList(envVarList EnvVarList, keys []string) EnvVarList {
+	removeEqualsFromEnvList(&keys)
+
 	newEnvVarList := EnvVarList{}
 	for _, envVar := range envVarList {
 		// if the env is in the keys we skip it
@@ -95,4 +97,16 @@ func RemoveEnvVarsFromList(envVarList EnvVarList, keys []string) EnvVarList {
 		newEnvVarList = append(newEnvVarList, envVar)
 	}
 	return newEnvVarList
+}
+
+// removeEqualsFromEnvList removes the equal sign from the list of environment variables
+// if (for example) foo=bar was passed in. Since we're only interested in the actual variable name
+// we don't care if foo=bar=foobar has been passed in.. Only the first value.
+func removeEqualsFromEnvList(keys *[]string) {
+
+	// Iterate through, if key contains "=", split it.
+	for index, key := range *keys {
+		(*keys)[index] = strings.Split(key, "=")[0]
+	}
+
 }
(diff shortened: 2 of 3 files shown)


dharmit:
/test v4.2-benchmark

```
2019/10/10 15:35:09 Container setup in pod benchmark failed, exit code 1, reason Error
Another process exited
```

dharmit:
/test v4.1-integration

```
[91m[1m[Fail] [0m[90modo component command tests [0m[0mwhen running odo push multiple times, check for existence of environment variables [0m[91m[1m[It] should should retain the same environment variable on multiple push [0m
[37m/go/src/github.com/openshift/odo/tests/helper/helper_run.go:33[0m
```

@amitkrout this seems to be failing in tests not touched by this PR. Is it a known flake?

amitkrout:
This PR does not solve the purpose 

```
$ odo component create python
 ✓  Validating component [494ms]

$ odo config set --env PORT=4000
Environment variables were successfully updated.
Run `odo push --config` command to apply changes to the cluster. //// Local config updated.

$ odo config unset --env PORT=anything
Environment variables were successfully updated.      //// It removes the env PORT from the local config
```
Ideally it should report error 
/hold

dharmit:
@cdrage upon unset'ing the env vars, it should still print the message for `odo push` that's being printed when setting an env var, no?

```sh
$ odo config set --env PORT=8080
Environment variables were successfully updated.
Run `odo push --config` command to apply changes to the cluster.

$ odo config unset --env PORT=9090
Environment variables were successfully updated.
```

amitkrout:
> /test v4.1-integration
> 
> ```
> �[91m�[1m[Fail] �[0m�[90modo component command tests �[0m�[0mwhen running odo push multiple times, check for existence of environment variables �[0m�[91m�[1m[It] should should retain the same environment variable on multiple push �[0m
> �[37m/go/src/github.com/openshift/odo/tests/helper/helper_run.go:33�[0m
> ```
> 
> @amitkrout this seems to be failing in tests not touched by this PR. Is it a known flake?

It could be a flake. From this log it's hard to confirm the same. Here you just pasted the ginkgo produced output. The actual failure is 
```
[odo] Configuration changes
[odo]  •  Retrieving component data  ...
[odo]  ✓  Retrieving component data [80ms]
[odo]  •  Applying configuration  ...
[odo]  ✗  Applying configuration [5m]
[odo]  ✗  Failed To Update Config To Component Deployed
```
This is being tracked in #1981

cdrage:
This PR does solve it's purpose... The edge case is simply the user had put in `--unset FOO=BAR` instead of `--unset FOO`.

It shouldn't matter what the value is being passed, only the key.

```sh
$ odo config set --env PORT=4000
Environment variables were successfully updated.
Run `odo push --config` command to apply changes to the cluster. //// Local config updated.

$ odo config unset --env PORT=anything
```

Would be correct.

What do you think @kadel ?

tests/integration/cmd_pref_config_test.go
```diff
@@ -262,6 +263,23 @@ var _ = Describe("odo preference and config command tests", func() {
...
+			// Create the project
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", project, "--context", context)
+
+			// Test out passing in with an equals sign
+			helper.CmdShouldPass("odo", "config", "set", "--env", "FOOBAR=foobar", "--context", context)
+			output := helper.CmdShouldPass("odo", "config", "view", "--context", context)
+			Expect(output).To(ContainSubstring("FOOBAR"))
+
+			// Check that we can remove it
+			helper.CmdShouldPass("odo", "config", "unset", "--env", "FOOBAR=foobar", "--context", context)
```
girishramnani: I dont think someone needs to provide the value of the env var as well to unset it? something like `odo config unset --env FOOBAR` should also work

tests/integration/cmd_pref_config_test.go
```diff
@@ -262,6 +263,23 @@ var _ = Describe("odo preference and config command tests", func() {
 			Expect(configValue).To(Not(ContainSubstring(("PORT"))))
 			Expect(configValue).To(Not(ContainSubstring(("SECRET_KEY"))))
 		})
+
+		It("should be able to unset an environment variable with foo=bar", func() {
```
girishramnani: can we have more complex cases? like setting multiple env vars and then unset multiple env vars ( all, subset )?

girishramnani:
> This PR does solve it's purpose... The edge case is simply the user had put in `--unset FOO=BAR` instead of `--unset FOO`.
> 
> It shouldn't matter what the value is being passed, only the key.
> 
> ```shell
> $ odo config set --env PORT=4000
> Environment variables were successfully updated.
> Run `odo push --config` command to apply changes to the cluster. //// Local config updated.
> 
> $ odo config unset --env PORT=anything
> ```
> 
> Would be correct.
> 
> What do you think @kadel ?

I think would be confusing for the user and they should be able to unset an env var without specifying the value part

dharmit:
@girishramnani 

> I think would be confusing for the user and they should be able to unset an env var without specifying the value part as well

This works fine already. Doing `odo config unset --env foo` is going to work as expected. However, it's the case when someone provides a value `foo=bar` while trying to unset the env var that we're trying to address. I know most one doesn't provide a value when unsetting an env var but, at the moment, when someone does `odo config unset --env foo=bar`, `odo` does nothing but prints on the CLI that env var were updated.

kadel:
> This PR does solve it's purpose... The edge case is simply the user had put in `--unset FOO=BAR` instead of `--unset FOO`.
> 
> It shouldn't matter what the value is being passed, only the key.
> 
> ```shell
> $ odo config set --env PORT=4000
> Environment variables were successfully updated.
> Run `odo push --config` command to apply changes to the cluster. //// Local config updated.
> 
> $ odo config unset --env PORT=anything
> ```
> 
> Would be correct.
> 
> What do you think @kadel ?

`odo config unset --env PORT` definitely needs to work.

`odo config unset --env PORT=sometnig` can either work, but if it does the value should probably match, otherwise it might be a little bit confusing. 
Or it can simply fail. Which I think is better. As the value `--env` should be defined as a variable name (`odo config unset --env <var>`, and in the example above `PORT=sometnig` is not a valid variable name

dharmit:
> Or it can simply fail. Which I think is better.

I would have personally advocated for this. But, unfortunately, we didn't come to a decision on #2139 before this PR. :(

tests/integration/cmd_pref_config_test.go
cdrage: That's the point of the test. To test to see if equals still works ex. `FOOBAR=qwerty` would unset FOOBAR.

cdrage:
I'll update this to error out of someone provides `=` ...

/label do-not-merge/work-in-progress

cdrage:
I have replaced this PR with this one: https://github.com/openshift/odo/pull/2293

Issue #2138, opened 2019-09-17
odo watch should fail with error message for git based components


dharmit:
[kind/bug]

## What versions of software are you using?
- Operating System: all
- Output of `odo version`: `master`

## How did you run odo exactly?
1. Created and pushed a component based on remote git repo
2. Executed `odo watch`

## Actual behavior
```sh
 $ odo watch
 ✗  stat https://github.com/dharmit/django-ex: no such file or directory
```

## Expected behavior
It should fail with a message like `Cannot watch git based components`
```sh
$ odo watch
 ✗  odo cannot watch git based components. Please run `odo push` to push changes
```

PR #2280, opened 2019-10-12
show error when using odo watch on git components

girishramnani:
Currently due to the check of source path existence using `os.Stat` being present before `sourceType` validation - `odo watch` showed `directory doesn't exist` error instead of `watch not supported for git ...` kind.
So resolved that and added a test.

fixes https://github.com/openshift/odo/issues/2138

## How to test changes?
```
odo watch # should show an error for kind "Watch is supported by binary and local components only...."
```

### DIFF
--- a/pkg/odo/cli/component/watch.go
+++ b/pkg/odo/cli/component/watch.go
@@ -86,16 +86,16 @@ func (wo *WatchOptions) Complete(name string, cmd *cobra.Command, args []string)
 // Validate validates the watch parameters
 func (wo *WatchOptions) Validate() (err error) {
 
-	// Validate component path existence and accessibility permissions for odo
-	if _, err := os.Stat(wo.sourcePath); err != nil {
-		return errors.Wrapf(err, "Cannot watch %s", wo.sourcePath)
-	}
-
 	// Validate source of component is either local source or binary path until git watch is supported
 	if wo.sourceType != "binary" && wo.sourceType != "local" {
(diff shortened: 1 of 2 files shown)


pkg/odo/cli/component/watch.go
```diff
@@ -86,18 +86,18 @@ func (wo *WatchOptions) Complete(name string, cmd *cobra.Command, args []string)
...
-
 	// Validate source of component is either local source or binary path until git watch is supported
 	if wo.sourceType != "binary" && wo.sourceType != "local" {
 		return fmt.Errorf("Watch is supported by binary and local components only and source type of component %s is %s",
 			wo.localConfig.GetName(),
 			wo.sourceType)
 	}
 
+	// Validate component path existence and accessibility permissions for odo
+	if _, err := os.Stat(wo.sourcePath); err != nil {
```
dharmit: Just thinking out loud here. Can we not have a simpler condition that throws this error if the `sourceType` is `git`? If not, then I would recommend that we add more comments here about when the said error could hit us and why we're using this way to handle it.

girishramnani: @dharmit the line
https://github.com/openshift/odo/pull/2280/files/0a1625b32e96ffa0ba608170b113ec37bbae6baf#diff-f36b01078d99cd02f0a8239f8d2553ebR90 is doing the check for `sourceType != git` validation. This check is if the user tries to watch a path that doesn't exist or/ we dont have permissions to watch it.

girishramnani:
3rd time the charm?

PR #2275, opened 2019-10-11
Fails if user tries to unset invalid environment variable

dharmit:
## What is the purpose of this change? What does it change?

At the moment, there's no check for the scenario where a user tries to unset an environment variable that doesn't exist in the configuration. This PR aims to fix that and fails the execution.

## Was the change discussed in an issue?
fixes #2274 

## How to test changes?

* Create a component
* Try to unset an environment variable that doesn't exist in component's config: `odo config unset --env nosuchvar`.

PR #2293, opened 2019-10-16
Error out when there is a missing environment variable

cdrage:
This PR errors out when there is a missing environment variable or an
invalid environment variable has been passed.

To test:

```sh
odo config set --env foo=bar
odo config unset --env testvariable # will error
odo config unset --env foo=bar # will error
odo config unset --env foo # will pass
```

Closes https://github.com/openshift/odo/issues/2139

### DIFF
--- a/pkg/config/env_var_test.go
+++ b/pkg/config/env_var_test.go
@@ -131,12 +131,15 @@ func TestNewEnvVarListFromSlice(t *testing.T) {
 }
 
 func TestRemoveEnvVarsFromList(t *testing.T) {
-	cases := []struct {
+	tests := []struct {
+		name       string
 		envVarList EnvVarList
 		expected   EnvVarList
 		keys       []string
+		wantErr    bool
 	}{
 		{
+			name: "Case 1: Check removing one environment variable",
 			envVarList: EnvVarList{
 				EnvVar{
 					Name:  "foo",
@@ -153,9 +156,11 @@ func TestRemoveEnvVarsFromList(t *testing.T) {
 					Value: "bar",
 				},
 			},
-			keys: []string{"fizz"},
+			keys:    []string{"fizz"},
+			wantErr: false,
 		},
 		{
+			name: "Case 2: Check removing two environment variables",
 			envVarList: EnvVarList{
 				EnvVar{
 					Name:  "foo",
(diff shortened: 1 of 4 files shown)


girishramnani:
according to the PR https://github.com/openshift/odo/pull/2271, shouldn't `odo config unset --env foo=bar` work? or we decided against it?

dharmit:
@cdrage I think there's another PR open for this already #2275. :wink:

cdrage:
Duplicate PR: https://github.com/openshift/odo/pull/2275

Going to:

/hold

@dharmit if you can copy over the tests I wrote here to your PR that'd be great! The env tests are still using the old format for test writing.

pkg/config/env_var_test.go
```diff
@@ -192,29 +199,68 @@ func TestRemoveEnvVarsFromList(t *testing.T) {
...
 					Name:  "foo",
 					Value: "bar",
 				},
+				EnvVar{
+					Name:  "hi",
+					Value: "hello",
+				},
+			},
+			expected: EnvVarList{},
+			keys:     []string{"foo=bar", "hi"},
```
dharmit: Shouldn't this pass because although the env var `foo=bar` doesn't exist, `hi` does exist.

pkg/config/env_var_test.go
```diff
@@ -192,29 +199,68 @@ func TestRemoveEnvVarsFromList(t *testing.T) {
...
 				},
+				EnvVar{
+					Name:  "hi",
+					Value: "hello",
+				},
 			},
 			expected: EnvVarList{},
-			keys:     []string{"foo"},
+			keys:     []string{""},
+			wantErr:  true,
```
dharmit: Why are we expecting an error here? Isn't this similar to case 3 where we don't pass any keys?

pkg/config/env_var_test.go
cdrage: No, because we fail at foo=bar immediately before returning the whole list. We shouldn't return a pass when one of the parameters is incorrect.

pkg/config/env_var_test.go
cdrage: We want an error because there are no keys passed

dharmit: But this looks similar to Case 3 where we're not having `keys` defined and we don't expect an error there.

Issue #1940, opened 2019-07-23
Push fails with error "unable to wait for build: timed out waiting for the condition"


amitkrout:
[kind/bug]

## What versions of software are you using?
- Operating System: OpenShift CI
- Output of `odo version`: master

## How did you run odo exactly?
OpenShift CI failure. 

## Actual behavior
```
[odo]  ✗  Waiting for build to finish [30s]
[odo]  ✗  Failed to create component with name ref-test. Please use `odo config view` to view settings used to create component. Error: unable to wait for build ref-test-app-1 to run: timed out waiting for the condition 
```

## Expected behavior
Should not error out

PR #2305, opened 2019-10-21
Adding odo file index json to gitignore file

### LINKED ISSUE #2223: add odo-file-index.json entry into .gitignore
When a component is created `odo component create` the command should automatically add `./.odo/odo-file-index.json` to the `.gitignore` file.

If `.gitignore` doesn't exist it should be created even if the current directory is not a git repo as the user might be creating git repo later after the component is created.

prietyc123:

Adding odo-file-index.json into .gitignore file
fixes #2223 

## How to test changes?

make test-cmd-push

### DIFF
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -2,9 +2,12 @@ package util
 
 import (
 	"encoding/json"
+
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -175,6 +178,41 @@ func RunIndexer(directory string, ignoreRules []string) (filesChanged []string, 
 			return filesChanged, filesDeleted, err
 		}
 	}
+
+	// Check .gitignore file exists or not
+	// Add odo-file-index.json entry to .gitignore
+
+	_, err = os.Stat(filepath.Join(filepath.Dir(directory), ".gitignore"))
+
+	if err == nil {
+		file, err := os.OpenFile(filepath.Join(filepath.Dir(directory), ".gitignore"), os.O_APPEND|os.O_RDWR,0600)
+		if err != nil {
+			log.Fatalf("failed opening file: %s", err)
+		}
+		defer file.Close()
+
+		data, err := ioutil.ReadFile(filepath.Join(filepath.Dir(directory), ".gitignore"))
+		if err != nil {
+			log.Panicf("failed reading data from file: %s", err)
+		}else {
+			s := string(data)
+			if strings.Contains(s, resolvedPath){}else {
+				if _, err = file.WriteString("\n" + resolvedPath); err != nil {
+					log.Panicf("failed writing string to file: %s", err)
+				}
+			}
+		}
+	}else if os.IsNotExist(err){
+		file, err := os.OpenFile(filepath.Join(filepath.Dir(directory), ".gitignore"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			log.Fatalf("failed opening file: %s", err)
+		}
+		defer file.Close()
+
+		if _, err = file.WriteString("\n" + resolvedPath); err != nil {
+			log.Panicf("failed writing string to file: %s", err)
+		}
+	}else{}
 
 	return filesChanged, filesDeleted, nil
 }
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -2,12 +2,13 @@ package util
 
 import (
 	"encoding/json"
-	"github.com/pkg/errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/golang/glog"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -115,30 +116,22 @@ func addOdoFileIndex(ignoreFile string) error {
 	return nil
 }
 
-// Check .gitignore file exists or not
-func checkGitIgnoreFile(directory string) error {
+// Check .gitignore file exists or not, if not then create it
+func checkGitIgnoreFile(directory string) (string, error) {
 
 	ignoreFile, err := gitIgnoreFilePath(directory)
 	// err checks the existence of .gitignore and then creates if does not exists
 	_, err = os.Stat(ignoreFile)
-	// As the directory already exists it means it already has one file i.e .odo file in it so ignore err
-	files, _ := ioutil.ReadDir(filepath.Join(directory))
 
-	if len(files) > 1 {
+	if err != nil {
+		file, err := os.OpenFile(ignoreFile, os.O_WRONLY|os.O_CREATE, 0600)
 		if err != nil {
-			file, err := os.OpenFile(ignoreFile, os.O_WRONLY|os.O_CREATE, 0600)
-			if err != nil {
-				return errors.Wrap(err, "failed to create .gitignore file")
-			}
-			file.Close()
+			return ignoreFile, errors.Wrap(err, "failed to create .gitignore file")
 		}
-		err = addOdoFileIndex(ignoreFile)
-		if err != nil {
-			return err
-		}
+		file.Close()
 	}
 
-	return nil
+	return ignoreFile, nil
 }
 
 // RunIndexer walks the given directory and finds the files which have changed and which were deleted/renamed
@@ -154,7 +147,13 @@ func RunIndexer(directory string, ignoreRules []string) (filesChanged []string, 
 	}
 
 	// check for .gitignore file and add odo-file-index.json to .gitignore
-	err = checkGitIgnoreFile(directory)
+	ignoreFile, err := checkGitIgnoreFile(directory)
+	if err != nil {
+		return filesChanged, filesDeleted, err
+	}
+
+	// add odo-file-index.json path to .gitignore
+	err = addOdoFileIndex(ignoreFile)
 	if err != nil {
 		return filesChanged, filesDeleted, err
 	}
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -127,10 +127,8 @@ func checkGitIgnoreFile(directory string) error {
 		if err != nil {
 			return errors.Wrap(err, "failed reading data from file")
 		} else {
-			s := string(data)
 			// check whether .odo/odo-file-index.json is already in the .gitignore file
-			if strings.Contains(s, ".odo/odo-file-index.json") {
-			} else {
+			if !strings.Contains(string(data), ".odo/odo-file-index.json") {
 				if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
 					return errors.Wrap(err, "failed writing string to file")
 				}
@@ -152,7 +150,6 @@ func checkGitIgnoreFile(directory string) error {
 				return errors.Wrap(err, "failed writing string to file")
 			}
 		}
-	} else {
 	}
 	return nil
 }
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -119,7 +119,9 @@ func addOdoFileIndex(ignoreFile string) error {
 func checkGitIgnoreFile(directory string) error {
 
 	ignoreFile, err := gitIgnoreFilePath(directory)
+	// err checks the existence of .gitignore and then creates if does not exists
 	_, err = os.Stat(ignoreFile)
+	// As the directory already exists it means it already has one file i.e .odo file in it so ignore err
 	files, _ := ioutil.ReadDir(filepath.Join(directory))
 
 	if len(files) > 1 {
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -88,7 +88,7 @@ func resolveFilePath(directory string) (string, error) {
 
 // gitignoreFilePath gives the filepath of the .gitignore file in the context
 func gitIgnoreFilePath(directory string) (string, error) {
-	_, err := os.Stat(filepath.Join(directory))
+	_, err := os.Stat(directory)
 	if err != nil {
 		return "", err
 	}
@@ -110,7 +110,7 @@ func addOdoFileIndex(gitIgnoreFile string) error {
 	// check whether .odo/odo-file-index.json is already in the .gitignore file
 	if !strings.Contains(string(data), filepath.Join(fileIndexDirectory, fileIndexName)) {
 		if _, err := file.WriteString("\n" + filepath.Join(fileIndexDirectory, fileIndexName)); err != nil {
-			return errors.Wrap(err, "failed to Add odo-file-index.json to .gitignore file")
+			return errors.Wrapf(err, "failed to Add %v to .gitignore file", fileIndexName)
 		}
 	}
 	return nil
@@ -124,9 +124,7 @@ func checkGitIgnoreFile(directory string) (string, error) {
 		return "", err
 	}
 	// err checks the existence of .gitignore and then creates if does not exists
-	_, err = os.Stat(gitIgnoreFile)
-
-	if err != nil {
+	if _, err := os.Stat(gitIgnoreFile); os.IsNotExist(err) {
 		file, err := os.OpenFile(gitIgnoreFile, os.O_WRONLY|os.O_CREATE, 0600)
 		if err != nil {
 			return gitIgnoreFile, errors.Wrap(err, "failed to create .gitignore file")
--- a/pkg/util/file_indexer.go
+++ b/pkg/util/file_indexer.go
@@ -2,8 +2,7 @@ package util
 
 import (
 	"encoding/json"
-	"log"
-
+	"github.com/pkg/errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -87,19 +86,20 @@ func resolveFilePath(directory string) (string, error) {
 }
 
 // gitignoreFilePath gives the filepath of the .gitignore file in the contex
-func gitignoreFilePath(directory string) (string, error) {
-	directoryFi, err := os.Stat(filepath.Join(directory))
+func gitIgnoreFilePath(directory string) (string, error) {
+	directoryFile, err := os.Stat(filepath.Join(directory))
 	if err != nil {
 		return "", err
 	}
 
-	switch mode := directoryFi.Mode(); {
+	switch mode := directoryFile.Mode(); {
 	case mode.IsDir():
-		// do directory stuff
+		// whether the mode is a directory stuff
+		// mode directory bit is set
 		return filepath.Join(directory, ".gitignore"), nil
 	case mode.IsRegular():
-		// do file stuff
-		// for binary files
+		// whether the mode is a file stuff
+		// No mode type is set
 		return filepath.Join(filepath.Dir(directory), ".gitignore"), nil
 	}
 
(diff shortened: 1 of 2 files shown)


amitkrout:
@prietyc123 Fix the go formatting failure.
```
$ make validate
./scripts/check-gofmt.sh
./pkg/util/file_indexer.go
make: *** [gofmt] Error 1
The command "make validate" exited with 2.
```

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+	return directory, nil
+}
+
+// Check .gitignore file exists or not
+// Add odo-file-index.json entry to .gitignore
+func checkGitignoreFile(directory string) {
+
+	ignoreFile, err := gitignoreFilePath(directory)
+	if err != nil {
+		log.Fatalf("file does not exist: %s", err)
```
cdrage: you shouldn't be using `log.Fatalf`, but instead be returning `err` in the function so that it errors up throughout the code until it his logging.

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+	if err == nil {
+		file, err := os.OpenFile(ignoreFile, os.O_APPEND|os.O_RDWR, 0600)
+		if err != nil {
+			log.Fatalf("failed opening file: %s", err)
+		}
+		defer file.Close()
+
+		data, err := ioutil.ReadFile(ignoreFile)
+		if err != nil {
+			log.Panicf("failed reading data from file: %s", err)
```
cdrage: same here.. shouldn't be using log, get rid of all of it.

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+func checkGitignoreFile(directory string) {
+
+	ignoreFile, err := gitignoreFilePath(directory)
+	if err != nil {
+		log.Fatalf("file does not exist: %s", err)
+	}
+
+	_, err = os.Stat(ignoreFile)
+
+	if err == nil {
```
cdrage: using `err == nill` and then putting code within here is unorthodox. why exactly are we continuing after erroring here: https://github.com/openshift/odo/pull/2305/files#diff-ded22a7f93f540164727d9088a72def4R118 ?

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+		file, err := os.OpenFile(ignoreFile, os.O_APPEND|os.O_RDWR, 0600)
```
cdrage: have you tested this on windows?

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+				log.Fatalf("failed opening file: %s", err)
+			}
+			defer file.Close()
+
+			if _, err = file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
+				log.Panicf("failed writing string to file: %s", err)
+			}
+		}
+	} else {
+	}
```
cdrage: what

prietyc123:
> Unfortunately there are too many nested if statements, specifically:
> 
> ```shell
> 	} else if os.IsNotExist(err) {
> 		files, err := ioutil.ReadDir(filepath.Join(directory))
> 		if err != nil {
> 			return errors.Wrap(err, "No such directory")
> 		}
> 		if len(files) > 1 {
> 			file, err := os.OpenFile(ignoreFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
> 			if err != nil {
> 				return errors.Wrap(err, "failed opening file")
> 			}
> 			defer file.Close()
> 
> 			if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
> 				return errors.Wrap(err, "failed writing string to file")
> 			}
> 		}
> ```
> 
> Which is unorthodox as we rely on erroring out of the function in order to continue.
> 
> Is there any way for you to do any peer programming with @amitkrout or @dharmit in the office? Go is pretty difficult to grasp especially with the way it handles errors :(

Hmm, Let me find out some better way to do this. Anyway I will discuss with @amitkrout or @dharmit whoever is available.

prietyc123:
@cdrage Can you please do next round of review

pkg/util/file_indexer.go
dharmit: This indeed looks unusual. In Go, we generally:
1. have a check for `if err != nil` and do something (mostly do some cleanup if required and throw the error back to caller or exit the program)
2. then, in the same function that has `if err != nil` check, we have more code that's supposed to be executed if there was no error.

In pseodo-code it might look something like below:
```go
func myAwesomeFunction(abc string) error {
    ignoreFile, err := gitIgnoreFilePath(directory)

    if err != nil {
        // do what needs to be done if there's an error
    }
    // now write the code that's supposed to be executed if there's no error in the step above.
}

File: pkg/util/file_indexer.go [hunk 6]
```diff
@@ -83,6 +85,75 @@ func resolveFilePath(directory string) (string, error) {
...
+		file, err := os.OpenFile(ignoreFile, os.O_APPEND|os.O_RDWR, 0600)
+		if err != nil {
+			return errors.Wrap(err, "failed opening file")
+		}
+		defer file.Close()
+
+		data, err := ioutil.ReadFile(ignoreFile)
+		if err != nil {
+			return errors.Wrap(err, "failed reading data from file")
+		} else {
```
dharmit: Like I mentioned in previous comment, in Go, we generally have a check for error and then the code that is supposed to be executed if there was no error. So in this case, we could simplify it to:
```go
data, err := ioutil.ReadFile(ignoreFile)
if err != nil {
	return errors.Wrap(err, "failed reading data from file")
}
// check whether .odo/odo-file-index.json is already in the .gitignore file
if !strings.Contains(string(data), ".odo/odo-file-index.json") {
	if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
			return errors.Wrap(err, "failed writing string to file")
	}
}
```

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,75 @@ func resolveFilePath(directory string) (string, error) {
...
+		}
+	} else if os.IsNotExist(err) {
+		files, err := ioutil.ReadDir(filepath.Join(directory))
+		if err != nil {
+			return errors.Wrap(err, "No such directory")
+		}
+		if len(files) > 1 {
+			file, err := os.OpenFile(ignoreFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+			if err != nil {
+				return errors.Wrap(err, "failed opening file")
```
dharmit: Which file are we talking about here?

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,75 @@ func resolveFilePath(directory string) (string, error) {
...
+			}
+			defer file.Close()
+
+			if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
+				return errors.Wrap(err, "failed writing string to file")
```
dharmit: Which string? Which file? We could rephrase this to something like `Failed to add odo-file-index.json to gitignore`. That's just a suggestion. You could, of course, change it to something else that conveys the user what *exactly* failed.

pkg/util/file_indexer.go
prietyc123: No not yet. Will test and update the result here.

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
 	return directory, nil
 }
 
+// gitignoreFilePath gives the filepath of the .gitignore file in the contex
```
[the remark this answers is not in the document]
prietyc123: Fixed the typo in the latest change.

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,77 @@ func resolveFilePath(directory string) (string, error) {
...
+func gitignoreFilePath(directory string) (string, error) {
```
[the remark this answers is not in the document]
prietyc123: Corrected the naming :-)

pkg/util/file_indexer.go
prietyc123: Thanks @dharmit for pseudo-code.
I have updated all the changes. Can you please review once again and let me know if any modification is required further.

pkg/util/file_indexer.go
prietyc123: For better understanding to the user I have done the modification in my latest change update.

pkg/util/file_indexer.go
```diff
@@ -95,6 +151,12 @@ func RunIndexer(directory string, ignoreRules []string) (filesChanged []string,
 		return filesChanged, filesDeleted, err
 	}
 
+	// check for .gitignore file and add odo-file-index.json to .gitignore
```
dharmit: I'm a bit sceptical about putting this here. This piece of code would run each time a user does `odo push`. Should this be a part of the push action or create, that is, `odo create`? I feel it should be in `odo create`. WDYT @kadel ?

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,60 @@ func resolveFilePath(directory string) (string, error) {
...
+		}
+	}
+	return nil
+}
+
+// Check .gitignore file exists or not
+func checkGitIgnoreFile(directory string) error {
+
+	ignoreFile, err := gitIgnoreFilePath(directory)
+	_, err = os.Stat(ignoreFile)
```
dharmit: Since we're doing `_, err = os.Stat(ignoreFile)`, I'm assuming that we're not interested in the `FileInfo` returned by `os.Stat`. In that case, why do we have a call to it here? I'm confused. Or am I looking at it the wrong way?

We're also not doing anything with the `err` here.

amitkrout:
> /test v4.3-integration

wait till pr #2362 merge in

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,60 @@ func resolveFilePath(directory string) (string, error) {
...
+	files, _ := ioutil.ReadDir(filepath.Join(directory))
```
dharmit: Why are we ignoring the error returned by `ioutil.ReadDir`?

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,60 @@ func resolveFilePath(directory string) (string, error) {
...
+
+	if len(files) > 1 {
+		if err != nil {
```
dharmit: Which `err` this checking for? I think it's the one in the line `_, err = os.Stat(ignoreFile)`. What's the point in checking for that err here?

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,60 @@ func resolveFilePath(directory string) (string, error) {
...
+			}
+			file.Close()
+		}
+		err = addOdoFileIndex(ignoreFile)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
dharmit: IIUC, `checkGitIgnoreFile` is doing multiple things:
1. Checks if the `.gitignore` file is present.
2. If it's not present, it creates the file.
3. Adds entry for `.odo/odo-file-index.json` in the `.gitignore` file.

I think we should separate the 3rd step into a separate function of its own. I get it that we've created a separate function `addOdoFileIndex` but calling it within `checkGitIgnoreFile` still defies the purpose of this function. We could add it to `file_indexer.go` after the call to `checkGitIgnoreFile`.

tests/integration/cmd_push_test.go
```diff
@@ -298,6 +298,19 @@ var _ = Describe("odo push command tests", func() {
 		})
 	})
 
+	Context("when .gitignore file exists or not", func() {
+		It("should create and push the contents of a named component and include odo-file-index.json path to .gitignore file to exclude the contents", func() {
+			helper.CmdShouldPass("git", "clone", "https://github.com/openshift/nodejs-ex", context+"/nodejs-ex")
```
dharmit: Please use `helper.CopyExample` like in https://github.com/openshift/odo/blob/master/tests/integration/servicecatalog/cmd_link_unlink_test.go#L66.

pkg/util/file_indexer.go
prietyc123: @dharmit IMO this should be in ```odo push``` because ```odo-file-index.json``` will always creates after ```odo push``` and ofcourse the code operation comes into picture after creation of ```odo-file-index.json```. Lets wait how @kadel see it.

pkg/util/file_indexer.go
prietyc123: Yes here we are just checking the existence of ```.gitignore``` path and if it does not exists the we are using ```err ``` to verify and create .gitignore if doesn't.

pkg/util/file_indexer.go
prietyc123: if ```filepath.join(directory)``` exists it means it will already have .odo file so there is no need of error handling.

pkg/util/file_indexer.go
prietyc123: https://github.com/openshift/odo/pull/2305#discussion_r342974192

pkg/util/file_indexer.go
prietyc123: @kadel @girishramnani @mik-dass WDYT? Can you please address if you have any suggestion for this.

tests/integration/cmd_push_test.go
prietyc123: I will add such new case in my test script.

pkg/util/file_indexer.go
mik-dass: @prietyc123 I feel @dharmit is right here. This function should just check for the existence of the .gitignore file.Only step 1 should be performed here.

pkg/util/file_indexer.go
kadel: the error checking should immediately follow `Stat` call. You should either move `if err!=nil` right after this, or move `os.Stat` down to where you are checking the error.

pkg/util/file_indexer.go
kadel: both make sense. It depends when we are adding it to the `.gitignore`. If we add `odo-file-index.json` to `.gitignore` when component (localConfig) is created that all the function should be in `create` package. If we are adding it to `.gitignore` when the `odo-file-index.json` file is created for the first time it should be in the package that creates `odo-file-index.json` file

pkg/util/file_indexer.go
kadel: better be safe ;-) If there is no need to explicitly ignore the error you should always check it.

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,62 @@ func resolveFilePath(directory string) (string, error) {
...
+}
+
+// Check .gitignore file exists or not
+func checkGitIgnoreFile(directory string) error {
+
+	ignoreFile, err := gitIgnoreFilePath(directory)
+	// err checks the existence of .gitignore and then creates if does not exists
+	_, err = os.Stat(ignoreFile)
+	// As the directory already exists it means it already has one file i.e .odo file in it so ignore err
```
kadel: This is not true. the directory can exist and be empty

pkg/util/file_indexer.go
```diff
@@ -83,6 +85,62 @@ func resolveFilePath(directory string) (string, error) {
...
+	files, _ := ioutil.ReadDir(filepath.Join(directory))
```
kadel: Why do you need to know what files are in the directory?

pkg/util/file_indexer.go
kadel: @dharmit +1

pkg/util/file_indexer.go
kadel: It is really not necessary to run this more than once. And there is no harm in adding it to `.gitignore` even if the file doesn't exist yet. 
I think it makes sense to do in `odo create` when the component gets initialized and `config.yaml` is created.


We can have it in fileindex also, but than we have to make sure that it is getting called only once when the `odo-file-index.json` is being created, definitely not every time the file is updated.

tests/integration/cmd_push_test.go
kadel: @prietyc123 this is just adding unnecessary test case and it is taking longer. Using just `CopyExample` is ok

prietyc123:
Seems to be a platform issue
/retest

amitkrout:
```
Submitted failure event to sentry (id=90bf0a60491a4877b897daa1c1ce62cf)
error: could not run steps: step root failed: could not wait for build: the build root failed with reason PushImageToRegistryFailed: Failed to push the image to the registry.
```
/retest

amitkrout:
@prietyc123 can you please fix this
```
$ make golint
golangci-lint run ./... --timeout 5m
pkg/util/file_indexer.go:122:14: ineffectual assignment to `err` (ineffassign)
	ignoreFile, err := gitIgnoreFilePath(directory)
	            ^
make: *** [golint] Error 1
The command "make golint" exited with 2.
```

pkg/util/file_indexer.go
prietyc123: @dharmit I have modified the content addressed by you :-)

prietyc123:
> @prietyc123 can you please fix this
> 
> ```
> $ make golint
> golangci-lint run ./... --timeout 5m
> pkg/util/file_indexer.go:122:14: ineffectual assignment to `err` (ineffassign)
> 	ignoreFile, err := gitIgnoreFilePath(directory)
> 	            ^
> make: *** [golint] Error 1
> The command "make golint" exited with 2.
> ```

Yes. I have fixed it .

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,57 @@ func resolveFilePath(directory string) (string, error) {
...
+	}
+	defer file.Close()
+
+	if data, err = ioutil.ReadFile(ignoreFile); err != nil {
+		return errors.Wrap(err, "failed reading data from .gitignore file")
+	}
+	// check whether .odo/odo-file-index.json is already in the .gitignore file
+	if !strings.Contains(string(data), ".odo/odo-file-index.json") {
+		if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
+			return errors.Wrap(err, "failed to Add odo-file-index.json to .gitignore file")
```
dharmit: Instead of hard-coding the `.odo/odo-file-index.json` value in above three lines, we could use the constants `fileIndexDirectory` & `fileIndexName` defined in this same file so that if the filename changes in future, we don't have to make changes at multiple places.

pkg/util/file_indexer.go
```diff
@@ -96,20 +96,20 @@ func gitIgnoreFilePath(directory string) (string, error) {
...
-	if data, err = ioutil.ReadFile(ignoreFile); err != nil {
+	if data, err = ioutil.ReadFile(gitIgnoreFile); err != nil {
 		return errors.Wrap(err, "failed reading data from .gitignore file")
 	}
 	// check whether .odo/odo-file-index.json is already in the .gitignore file
-	if !strings.Contains(string(data), ".odo/odo-file-index.json") {
-		if _, err := file.WriteString("\n" + ".odo/odo-file-index.json"); err != nil {
+	if !strings.Contains(string(data), filepath.Join(fileIndexDirectory, fileIndexName)) {
+		if _, err := file.WriteString("\n" + filepath.Join(fileIndexDirectory, fileIndexName)); err != nil {
 			return errors.Wrap(err, "failed to Add odo-file-index.json to .gitignore file")
```
dharmit: You mean `"failed to add %v to .gitignore file", fileIndexName`? :smile:

prietyc123: Yes :-D

pkg/util/file_indexer.go
```diff
@@ -83,6 +86,57 @@ func resolveFilePath(directory string) (string, error) {
...
+func checkGitIgnoreFile(directory string) (string, error) {
+
+	gitIgnoreFile, err := gitIgnoreFilePath(directory)
+	if err != nil {
+		return "", err
+	}
+	// err checks the existence of .gitignore and then creates if does not exists
+	_, err = os.Stat(gitIgnoreFile)
+
+	if err != nil {
```
mik-dass: The err can be other type too and thus may not indicate that the file doesn't exist. Please check the err to know the type like below

```
if _, err := os.Stat(filePath); os.IsNotExist(err) {
           // logic here
	}
```

mik-dass:
@prietyc123 If possible, also maybe add some unit tests for those functions?

tests/integration/cmd_push_test.go
```diff
@@ -318,7 +316,30 @@ var _ = Describe("odo push command tests", func() {
...
+
+	Context("when .gitignore file exists", func() {
+		It("should create and push the contents of a named component and include odo-file-index.json path to .gitignore file to exclude the contents", func() {
+			helper.CmdShouldPass("git", "clone", "https://github.com/openshift/nodejs-ex", context+"/nodejs-ex")
+			helper.CmdShouldPass("odo", "component", "create", "nodejs", cmpName, "--project", project, "--context", context+"/nodejs-ex", "--app", appName)
+
+			// push and include the odo-file-index.json path to .gitignore file
+			helper.CmdShouldPass("odo", "push", "--context", filepath.Join(context, "nodejs-ex"))
+			ignoreFilePath := filepath.Join(context, "nodejs-ex", ".gitignore")
+			helper.FileShouldContainSubstring(ignoreFilePath, filepath.Join(".odo", "odo-file-index.json"))
```
amitkrout: Then you need to verify inside the pod that ```odo-file-index.json``` is not pushed.

tests/integration/cmd_push_test.go
```diff
@@ -318,7 +316,30 @@ var _ = Describe("odo push command tests", func() {
...
+
+	Context("when .gitignore file does not exist", func() {
+		It("should create and push the contents of a named component and also create .gitignore then include odo-file-index.json path to .gitignore file to exclude the contents", func() {
+			helper.CopyExample(filepath.Join("source", "nodejs"), context)
+			helper.CmdShouldPass("odo", "component", "create", "nodejs", cmpName, "--project", project, "--context", context, "--app", appName)
 
+			// push and include the odo-file-index.json path to .gitignore file
+			helper.CmdShouldPass("odo", "push", "--context", filepath.Join(context))
+			ignoreFilePath := filepath.Join(context, ".gitignore")
+			helper.FileShouldContainSubstring(ignoreFilePath, filepath.Join(".odo", "odo-file-index.json"))
```
amitkrout: https://github.com/openshift/odo/pull/2305/files#r346914510

tests/integration/cmd_push_test.go
prietyc123: IMO as it is related to gitignore so while git push it should not be pushed. I think its nowhere related to pod and hence IMO it will not be the used case.

tests/integration/cmd_push_test.go
prietyc123: https://github.com/openshift/odo/pull/2305#discussion_r347314598

prietyc123:
#1940
/test v4.1-e2e-scenarios

tests/integration/cmd_push_test.go
amitkrout: ok

prietyc123:
> @prietyc123 If possible, also maybe add some unit tests for those functions?

https://github.com/openshift/odo/issues/2404

tests/integration/cmd_push_test.go
amitkrout: /lgtm

PR #2312, opened 2019-10-22
removed some un-necessary cleanups from the integration tests

girishramnani:
see $SUBJECT

no issue

## How to test changes?

the tests should pass

### DIFF
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -88,11 +88,6 @@ func componentTests(args ...string) {
 			applicationName := helper.GetConfigValue("Application")
 			Expect(applicationName).To(Equal("app"))
 
-			// clean up
-			helper.CmdShouldPass("odo", "app", "delete", "app", "-f")
-			helper.CmdShouldFail("odo", "app", "delete", "app", "-f")
-			helper.CmdShouldFail("odo", "component", "delete", componentName, "-f")
-
 		})
 
 		It("should create default named component when passed same context differently", func() {
@@ -190,7 +185,6 @@ func componentTests(args ...string) {
 			Expect(desiredCompListJSON).Should(MatchJSON(actualCompListJSON))
 			cmpAllList := helper.CmdShouldPass("odo", append(args, "list", "--all")...)
 			Expect(cmpAllList).To(ContainSubstring("cmp-git"))
-			helper.CmdShouldPass("odo", append(args, "delete", "cmp-git", "-f")...)
 		})
 
 		It("should list the component when it is not pushed", func() {
@@ -198,7 +192,6 @@ func componentTests(args ...string) {
 			cmpList := helper.CmdShouldPass("odo", append(args, "list", "--context", context)...)
 			Expect(cmpList).To(ContainSubstring("cmp-git"))
 			Expect(cmpList).To(ContainSubstring("Not Pushed"))
-			helper.CmdShouldPass("odo", append(args, "delete", "-f", "--all", "--context", context)...)
 		})
 
 		It("should list the component in the same app when one is pushed and the other one is not pushed", func() {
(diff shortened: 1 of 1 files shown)


tests/integration/component.go
```diff
@@ -88,11 +88,6 @@ func componentTests(args ...string) {
 			applicationName := helper.GetConfigValue("Application")
 			Expect(applicationName).To(Equal("app"))
 
-			// clean up
-			helper.CmdShouldPass("odo", "app", "delete", "app", "-f")
-			helper.CmdShouldFail("odo", "app", "delete", "app", "-f")
```
amitkrout: This is a verification steps that you can not remove same app multiple time

tests/integration/component.go
```diff
@@ -88,11 +88,6 @@ func componentTests(args ...string) {
...
-			helper.CmdShouldFail("odo", "component", "delete", componentName, "-f")
```
amitkrout: This is also a verification steps of component delete should fail if the application itself is not present

tests/integration/component.go
```diff
@@ -190,15 +185,13 @@ func componentTests(args ...string) {
 			Expect(desiredCompListJSON).Should(MatchJSON(actualCompListJSON))
 			cmpAllList := helper.CmdShouldPass("odo", append(args, "list", "--all")...)
 			Expect(cmpAllList).To(ContainSubstring("cmp-git"))
-			helper.CmdShouldPass("odo", append(args, "delete", "cmp-git", "-f")...)
```
amitkrout: If the test spec is responsible to create the component then the same component should be deleted as part of the test pec. Because just to make sure certain operation after component create should not affect the component delete.

tests/integration/component.go
```diff
@@ -190,15 +185,13 @@ func componentTests(args ...string) {
...
 		})
 
 		It("should list the component when it is not pushed", func() {
 			helper.CmdShouldPass("odo", append(args, "create", "nodejs", "cmp-git", "--project", project, "--git", "https://github.com/openshift/nodejs-ex", "--min-memory", "100Mi", "--max-memory", "300Mi", "--min-cpu", "0.1", "--max-cpu", "2", "--context", context, "--app", "testing")...)
 			cmpList := helper.CmdShouldPass("odo", append(args, "list", "--context", context)...)
 			Expect(cmpList).To(ContainSubstring("cmp-git"))
 			Expect(cmpList).To(ContainSubstring("Not Pushed"))
-			helper.CmdShouldPass("odo", append(args, "delete", "-f", "--all", "--context", context)...)
```
amitkrout: https://github.com/openshift/odo/pull/2312/files#r337863354, however more validation steps can be added here. For example 
```
helper.CmdShouldFail("odo", append(args, "delete", "-f")
helper.CmdShouldPass("odo", append(args, "delete", "-f", "--all", "--context", context)...)
Add a steps that should verify there should not be any odo local config context
```

tests/integration/component.go
```diff
@@ -215,10 +208,9 @@ func componentTests(args ...string) {
 			Expect(cmpList).To(ContainSubstring("cmp-git-2"))
 			Expect(cmpList).To(ContainSubstring("Not Pushed"))
 			Expect(cmpList).To(ContainSubstring("Pushed"))
-
-			helper.CmdShouldPass("odo", append(args, "delete", "-f", "--all", "--context", context)...)
```
amitkrout: It's a verification step of how component delete works with ```--all``` flag when you have multiple component in same app but in different context. This is a useful user scenario, so it can not be removed.

Add some verification step to confirm component delete with ```--all``` flag works even in this scenario

girishramnani:
closing this PR as it seems the cleanup is validation

PR #2325, opened 2019-10-25
Modifies the PR commit template

cdrage:
**What kind of PR is this?**
/kind docs

We need this for a few reasons:
  - Our current template has too much "fluff", needing to remove
  multiple lines of comments just to commit..
  - If you use the template from the command-line using `git commit` you
  are actually unable use markdown headers (`#` is a comment in Git...)
  - We should have a way to use `/kind`
  - This template follows the way Kubernetes does it:
  https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md
  but simpler

Fixes #?

View the doc :) The above example is what the new "template" looks like.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,21 @@
-## What is the purpose of this change? What does it change?
-<!-- Describe the changes here, as detailed as needed. -->
+**What kind of PR is this?**
+<!--
+DELETE the kind(s) which are not applicable before opening the PR.
+-->
 
-## Was the change discussed in an issue?
-fixes #???
-<!-- Please do Link issues here. -->
+/kind api-change
+/kind bug
+/kind cleanup
+/kind code refactorting
+/kind docs
+/kind enhancement
+/kind feature
+/kind flake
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,21 @@
...
-fixes #???
-<!-- Please do Link issues here. -->
+/kind api-change
+/kind bug
+/kind cleanup
+/kind code refactorting
+/kind docs
+/kind enhancement
+/kind feature
+/kind flake
```
cdrage: Note: The reason why we do NOT put these in the commented portion, like so:

```sh
<!--
/kind cleanup
/kind code refactorting
/kind docs
/kind enhancement
/kind feature
/kind flake
-->
```

Is because @openshift-ci-robot will *still* apply the label despite it being inside of a comment..

Hence why there is a comment that says to delete what you dont need :)

cdrage:
Ignore @openshift-ci-robot 's massive labeling :) It was doing some funky stuff while I edited the commit message. This works well now / I've tested it!

Ping @dharmit @girishramnani

PR #2334, opened 2019-10-30
change "from" to "to" for odo link

danielhelfand:
**What kind of PR is this?**
/kind cleanup

Changes successful `odo link` message from 

`Component backend has been successfully linked from the component frontend`

to

`Component backend has been successfully linked to the component frontend`

This reads a lot more clearly since information is being shared from the first component as part of `odo link`. The current implementation makes it sound as if information is coming from the component specified via the `--component` flag or via the context being worked in.

Looking for advice on whatever `operationName` options are available that could potentially not work with this message. Link was the only option considered here.

None

### DIFF
--- a/pkg/odo/cli/component/common_link.go
+++ b/pkg/odo/cli/component/common_link.go
@@ -115,7 +115,14 @@ func (o *commonLinkOptions) run() (err error) {
 		return err
 	}
 
-	log.Successf("%s %s has been successfully %sed from the component %s\n", linkType, o.suppliedName, o.operationName, o.Component())
+	switch o.operationName {
+	case "link":
+		log.Successf("%s %s has been successfully linked to the component %s\n", linkType, o.suppliedName, o.Component())
+	case "unlink":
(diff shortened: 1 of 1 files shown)


kadel:
> Looking for advice on whatever `operationName` options are available that could potentially not work with this message. Link was the only option considered here.

`operationName` can have two possible values `link` or `unlink` based on the command that is called `odo link` or `odo unlink`.

It was probably done like this to to safe some coding, but now this is just complicating the messaging :-(  
```
Component backend has been successfully unlinked to the component frontend
```
Now this doesn't look right to me.  But the original message in `odo link` was not good either. 


We might have to split the messages into two independent messages. 
Maybe something like this?
```go
switch operationName {
  case "link":
    log.Successf("%s %s has been successfully linked to the component %s\n", linkType, o.suppliedName, o.Component())
  case "unlink":
    log.Successf("%s %s has been successfully unlinked from the component %s\n", linkType, o.suppliedName, o.Component())
  default:
    return fmt.Errorf("unknwon operation %s", operationName)
}

```

kadel:
/approve

@danielhelfand thank you for helping us making odo much more understandable.

PR #2330, opened 2019-10-29
Fixes wait for project creation

mik-dass:
fixes #2117 
fixes #2284 

How to test:

- Run `odo project create <name> -w` and check if the command waits properly.
- `make test-cmd-project` should pass.

PR #2099, opened 2019-09-06
Removes apps from project specs and speeds up the query

mik-dass:
fixes #2073 

How to test:
1. Execute `project list` and `project list -o json` and compare speeds with the master branch

PR #2009, opened 2019-08-12
change the release extension from .gz to .tar.gz

girishramnani:
## What is the purpose of this change? What does it change?
see $SUBJECT.

## Was the change discussed in an issue?
fixes #1994 

## How to test changes?
`make prepare-release` should create

PR #2356, opened 2019-11-05
Commenting project command test flake spec

amitkrout:
**What kind of PR is this?**

/kind flake

There is already an ongoing pr #2330 that @mik-dass is working on to fix this flake. For time being to avoid the pain of same flake in other pr i just commented out the test. 

Fixes - None, For time being i just commented the test that causing the flake in most of the pr 

CI should pass without throwing the same flake

### DIFF
--- a/tests/integration/project/cmd_project_test.go
+++ b/tests/integration/project/cmd_project_test.go
@@ -38,7 +38,7 @@ var _ = Describe("odo project command tests", func() {
 		})
 	})
 
-	Context("when running project command app parameter in directory that doesn't contain .odo config directory", func() {
+	/*Context("when running project command app parameter in directory that doesn't contain .odo config directory", func() {
 		It("should successfully execute list along with machine readable output", func() {
 			listOutput := helper.CmdShouldPass("odo", "project", "list")
 			Expect(listOutput).To(ContainSubstring(project))
@@ -50,5 +50,5 @@ var _ = Describe("odo project command tests", func() {
 			Expect(err).Should(BeNil())
 			Expect(listOutputJSON).To(ContainSubstring(expected))
 		})
-	})
+	})*/
 })


kadel:
Do we know when this started failing and what changes introduced it?

amitkrout:
> Do we know when this started failing and what changes introduced it?

@kadel The flake was introduced after pr #2221. cc_ @mik-dass

kadel:
> > Do we know when this started failing and what changes introduced it?
> 
> @kadel The flake was introduced after pr #2221. cc_ @mik-dass

So it is there for exactly one month? Why do you want to comment it now?

amitkrout:
> > > Do we know when this started failing and what changes introduced it?
> > 
> > 
> > @kadel The flake was introduced after pr #2221. cc_ @mik-dass
> 
> So it is there for exactly one month? Why do you want to comment it now?

Because the appearance of the flake is more frequent. Moreover @mik-dass has WIP pr #2330 to fix this and we don't know when it will be fixed as per the last triage call.

girishramnani:
This PR also has the https://github.com/openshift/odo/issues/2357 4.3 intg failures @amitkrout

girishramnani:
@amitkrout the error seems different this time
```
[odo] unable to create new project: unable to create new project mwjiohfdco: projectrequests.project.openshift.io is forbidden: User "system:serviceaccount:ci-op-xb0s2vyf:default" cannot create projectrequests.project.openshift.io at the cluster scope: no RBAC policy matched
[odo]  ✗  projectrequests.project.openshift.io is forbidden: User "system:serviceaccount:ci-op-xb0s2vyf:default" cannot create projectrequests.project.openshift.io at the cluster scope: no RBAC policy matched
[odo]  ✗  Waiting for project to come up [2ms]```

girishramnani:
@mik-dass's PR makes the project -w command wait till its active, this error seems to be related to not having enough permissions to create one

mik-dass:
> Do we know when this started failing and what changes introduced it?

@amitkrout @kadel The error was introduced in #2099. It introduced a project test file and the since `project list` was failing back then also, the PR used sleep of 1 second. 

> @kadel The flake was introduced after pr #2221. cc_ @mik-dass

#2221 just tried to solve it, but couldn't completely. Maybe we should just introduce a sleep of 1 second as done in #2009 ?

kadel:
> #2221 just tried to solve it, but couldn't eliminate it completely. Maybe we should just introduce a sleep of 1 second as done in #2009 ?

If we don't have a solution right now, adding `sleep` and keeping the test is a much better solution than just commenting it out.

amitkrout:
> > #2221 just tried to solve it, but couldn't eliminate it completely. Maybe we should just introduce a sleep of 1 second as done in #2009 ?
> 
> If we don't have a solution right now, adding `sleep` and keeping the test is a much better solution than just commenting it out.

Ok, then i will add a sleep of 1 sec as it was before. @mik-dass make sure that you will remove the sleep in your fix.

Issue #2357, opened 2019-11-05
Tests failing on 4.3 openshift cluster


girishramnani:
[kind/bug]

## What versions of software are you using?
- Operating System: CI
- Output of `odo version`: odo master

We are seeing similar failures like these
```
Running oc with args [oc get dc nodejs-push-atonce-nodejs-push-test --namespace fleanejubz]
[oc] NAME                                  REVISION   DESIRED   CURRENT   TRIGGERED BY
[oc] nodejs-push-atonce-nodejs-push-test   1          1         1         config,image(nodejs:latest)
Running oc with args [oc get pods --namespace fleanejubz --selector=deploymentconfig=nodejs-push-atonce-nodejs-push-test -o jsonpath='{.items[0].metadata.name}']
[oc] 'nodejs-push-atonce-nodejs-push-test-1-fxgcz'Running oc with args [oc exec nodejs-push-atonce-nodejs-push-test-1-fxgcz --namespace fleanejubz -c nodejs-push-atonce-nodejs-push-test -- ls -la /tmp/src/package.json]
[oc] ls: cannot access /tmp/src/package.json: No such file or directory
```
where it seems that we are trying to find the `package.json` in `tmp/src/package.json`. This could have been changed in openshift 4.3

Below are the PRs where this failure is occuring

- Update PR template	                                    https://github.com/openshift/odo/pull/2344 (4.3 intg)
- Refactor template                                           https://github.com/openshift/odo/pull/2340  (4.3 intg)
- renamed Url to URL to resolve lint warning and consistency  https://github.com/openshift/odo/pull/2336 (4.3 intg)
- Add italic font when describing what to do next https://github.com/openshift/odo/pull/2348 (4.3 intg)

girishramnani:
there could be something different with the s2i paths in 4.3 because failure is due to oc exec not finding the package at /tmp/src

girishramnani:
This is occurring because the `tmp/src`'s is being moved the opt/app-root/src. https://github.com/sclorg/s2i-nodejs-container/blob/master/10/s2i/bin/assemble#L36

we need to resolve the tests to look into deployment dir

girishramnani:
verified that the backup is present in the backup dir
```
bash-4.2$ cd /opt/app-root/src-backup/
bash-4.2$ ls
src
bash-4.2$ cd src/
.dummy        .odo/         package.json  server.js     
```

girishramnani:
this means everything should work as expected but we need to just resolve the tests

Issue #2359, opened 2019-11-05
Nodejs images are not showing as supported on 4.3 cluster


girishramnani:
[kind/bug]

## What versions of software are you using?
- Operating System:
- Output of `odo version`: odo master

## How did you run odo exactly?
odo create nodejs

## Actual behavior
warning shows up

## Expected behavior
No warning

## Any logs, error output, etc?
```
[odo] Warning: nodejs is not fully supported by odo, and it is not guaranteed to work.
```

kadel:
/priority high

We need to figure out what changed, how the new images differ from the old one.

amitkrout:
@kadel @girishramnani There is no latest tag for nodejs component 
```
$ odo catalog list components
Odo Supported OpenShift Components:
NAME       PROJECT       TAGS
java       openshift     11,8,latest
nodejs     openshift     8,8-RHOAR
```

girishramnani:
yes, this image is `rhscl/nodejs-10-rhel7` latest, we need to add `rhscl/nodejs-10-rhel7` in the supported list

girishramnani:
Debug mode seems to be broken on the nodejs latest version on openshift 4.3

girishramnani:
so we need to resolve it

cdrage:
@girishramnani Is there any output / logging as to why debug mode isn't working for OpenShift 4.3 or an issue created against the S2I images?

girishramnani:
This failure has to be handled on our side @cdrage . I can open another issue scoping the failing debug mode if we want to not include that in this issue's scope.

kadel:
Should be fixed as part of https://github.com/openshift/odo-init-image/pull/43/files

kanchwala-yusuf:
This issue has been fixed by the PR https://github.com/openshift/odo/pull/2427

**Testing results:**
**CRC version:**
```zsh
➜  ~ crc version
crc version: 1.3.0-dev+b0d660e
OpenShift version: 4.3.0-0.nightly-2019-11-25-153929 (embedded in binary)
```

**odo-init-image:** "registry.access.redhat.com/openshiftdo/odo-init-image-rhel7:1.0.1"

**assemble-and-restart script correctly identifying the language:**
```zsh
++ /opt/odo/bin/getlanguage
+ IMAGE_LANG=nodejs
+ '[' -f /tmp/.s2i/bin/run ']'
+ '[' '!' -z nodejs ']'
```

**node.js application starting correctly with debugger:**
```
[...]

PR #2362, opened 2019-11-05
change checks of ODO_SRC_PATH/src in tests and a temporary fix for project list flake

girishramnani:
**What kind of PR is this?**

/kind bug
Due to recent change in the s2i image for nodejs in openshift 4.3. The `/tmp/src` directory doesn't hold the code anymore because it gets moved. 
We know this happens for some images and we have a workaround in the odo-init-image. But our test for nodejs assumes that source is present in `/tmp/src`, due to which they are failing on openshift 4.3.

Fixes https://github.com/openshift/odo/issues/2357

tests should pass on 4.3 cluster

### DIFF
--- a/tests/integration/cmd_debug_test.go
+++ b/tests/integration/cmd_debug_test.go
@@ -34,10 +34,10 @@ var _ = Describe("odo debug command tests", func() {
 		os.Unsetenv("GLOBALODOCONFIG")
 	})
 
-	Context("odo debug on a nodejs component", func() {
+	Context("odo debug on a nodejs:8 component", func() {
 		It("should expect a ws connection when tried to connect on different debug port locally and remotely", func() {
 			helper.CopyExample(filepath.Join("source", "nodejs"), context)
-			helper.CmdShouldPass("odo", "component", "create", "nodejs", "--project", project, "--context", context)
+			helper.CmdShouldPass("odo", "component", "create", "nodejs:8", "--project", project, "--context", context)
 			helper.CmdShouldPass("odo", "config", "set", "--force", "DebugPort", "9292", "--context", context)
 			dbgPort := helper.GetConfigValueWithContext("DebugPort", context)
 			Expect(dbgPort).To(Equal("9292"))
@@ -53,7 +53,7 @@ var _ = Describe("odo debug command tests", func() {
 
 		It("should expect a ws connection when tried to connect on default debug port locally", func() {
 			helper.CopyExample(filepath.Join("source", "nodejs"), context)
-			helper.CmdShouldPass("odo", "component", "create", "nodejs", "--project", project, "--context", context)
+			helper.CmdShouldPass("odo", "component", "create", "nodejs:8", "--project", project, "--context", context)
 			helper.CmdShouldPass("odo", "push", "--context", context)
 			go func() {
 				helper.CmdShouldRunWithTimeout(60*time.Second, "odo", "debug", "port-forward", "--context", context)
--- a/tests/integration/cmd_push_test.go
+++ b/tests/integration/cmd_push_test.go
@@ -197,7 +197,7 @@ var _ = Describe("odo push command tests", func() {
 			Expect(output).To(Not(ContainSubstring("No file changes detected, skipping build")))
 		})
 
-		FIt("should push only the modified files", func() {
+		It("should push only the modified files", func() {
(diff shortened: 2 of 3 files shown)


girishramnani:
so this issue https://github.com/openshift/odo/issues/2359 is failing on this PR

girishramnani:
Debug mode seems to be failing

dharmit:
I have a silly question. How do we know which tests need to be changed to use `nodejs:8` instead of `nodejs`? I mean, in spite of the latest commit pushed to address the test failures, there are still various places in the code where we're using `nodejs`, no? @girishramnani @kadel

girishramnani:
> I have a silly question. How do we know which tests need to be changed to use `nodejs:8` instead of `nodejs`? I mean, in spite of the latest commit pushed to address the test failures, there are still various places in the code where we're using `nodejs`, no? @girishramnani @kadel

in the tests where we changed to `nodejs:8` the internals are being tested.

tests/integration/cmd_push_test.go
```diff
@@ -197,72 +197,78 @@ var _ = Describe("odo push command tests", func() {
...
-				func(cmdOp string, err error) bool {
-					modifiedCatServerFile = cmdOp
-					return true
-				},
-			)
-
-			Expect(modifiedCatViewFile).NotTo(Equal(earlierCatViewFile))
-			Expect(modifiedCatServerFile).To(Equal(earlierCatServerFile))
-		})
+		// these tests are failing with stat output being different
```
kadel: Why is stat output different? We shouldn't comment out tests just because they are failing.
/hold

girishramnani: resolved @kadel

tests/integration/cmd_push_test.go
```diff
@@ -197,9 +197,9 @@ var _ = Describe("odo push command tests", func() {
 			Expect(output).To(Not(ContainSubstring("No file changes detected, skipping build")))
 		})
 
-		It("should push only the modified files", func() {
+		FIt("should push only the modified files", func() {
```
kadel: Focused test.

girishramnani: resolved

tests/integration/cmd_debug_test.go
```diff
@@ -34,10 +34,10 @@ var _ = Describe("odo debug command tests", func() {
 		os.Unsetenv("GLOBALODOCONFIG")
 	})
 
-	Context("odo debug on a nodejs component", func() {
+	Context("odo debug on a nodejs:8 component", func() {
```
kadel: why was this changed to `8`?

girishramnani: Because the latest nodejs image is not working with debug mode and hence also it's now shown as fully supported

girishramnani: We need to resolve the debug mode issue for latest node in a separate PR

girishramnani: Issue for the broken debug mode https://github.com/openshift/odo/issues/2359#issuecomment-550284845

girishramnani: We also have an issue which mentions that we should run test across all the image versions. So we have recorded everything that has to be done after this fix

tests/integration/cmd_push_test.go
```diff
@@ -55,27 +55,27 @@ var _ = Describe("odo push command tests", func() {
 		})
 
 		It("should be able to create a file, push, delete, then push again propagating the deletions", func() {
-			helper.CmdShouldPass("git", "clone", "https://github.com/openshift/ruby-ex", context+"/ruby-ex")
-			helper.CmdShouldPass("odo", "component", "create", "ruby", cmpName, "--project", project, "--context", context+"/ruby-ex", "--app", appName)
+			helper.CmdShouldPass("git", "clone", "https://github.com/openshift/nodejs-ex", context+"/nodejs-ex")
```
cdrage: Why are we changing it from ruby to nodejs? Is it because we don't "officially" support ruby (yet?)

girishramnani: Yes and some variables like `DEPLOYMENT_DIR` are not populated in ruby s2i images

amitkrout:
Review comments are addressed. Thanks @girishramnani 
/approve

PR #2367, opened 2019-11-06
Fixes watch unit

### LINKED ISSUE #2365: fix watch unit test
The current watch unit test doesn't work as CompDirStructure is a new variable at https://github.com/openshift/odo/blob/adf3006360abc5cb707d9a9f442669989272f4e5/pkg/component/watch_test.go#L218 and not the global one https://github.com/openshift/odo/blob/adf3006360abc5cb707d9a9f442669989272f4e5/pkg/component/watch_test.go#L75
and thus https://github.com/openshift/odo/blob/adf3006360abc5cb707d9a9f442669989272f4e5/pkg/component/watch_test.go#L72 is always empty and the mockLocalPush always throws a error and since we ignore the error at https://github.com/openshift/odo/blob/adf3006360abc5cb707d9a9f442669989272f4e5/pkg/component/watch.go#L308 the tests always pass.

Also it doesn't test operations like file delete, folder create and delete properly.

mik-dass:
**What kind of PR is this?**

/kind bug

Fixes the watch unit test and adds more 

Fixes #2365 

run `make test` and check if it passes

### DIFF
--- a/pkg/component/watch_test.go
+++ b/pkg/component/watch_test.go
@@ -6,6 +6,11 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


amitkrout:
As of now it looks good to me. Recently @dharmit and @girishramnani has spent good amount of time on researching UTs cases. So i would recommend atleast anyone of them to have a look into it.

pkg/component/watch_test.go
```diff
@@ -68,31 +73,86 @@ func setUpF8AnalyticsComponentSrc(componentName string, requiredFilePaths []test
...
 // Mock PushLocal to collect changed files and compare against expected changed files
 func mockPushLocal(client *occlient.Client, componentName string, applicationName string, path string, out io.Writer, files []string, delFiles []string, isPushForce bool, globExps []string, show bool) error {
-	for _, gotChangedFile := range files {
+	muLock.Lock()
+	defer muLock.Unlock()
+	if componentName != mockPush.componentName || applicationName != mockPush.applicationName || isPushForce != mockPush.isForcePush || show != mockPush.show {
+		fmt.Printf("some of the push parameters are different, wanted: %v, got: %v", mockPush, []string{
+			componentName, applicationName, "isPushForce:" + strconv.FormatBool(isPushForce), "show:" + strconv.FormatBool(show),
+		})
+		os.Exit(1)
```
kadel: Shouldn't this return error, rather than calling Exit?

mik-dass: we are ignoring the error later on in the code, so the test passes if I don't exit

mik-dass:
```
Container setup exited with code 1, reason Error
```
/retest

Issue #2093, opened 2019-09-05
Compatibility table for hot reloading support / image support


cdrage:
[kind/Documentation]

## What did you find missing in the documentation?

We should have a "matrix" table that lists what images are compability, whether they support hot reloading or not.

As well as:
 - What images support binaries (Java)
 - What images to use when using certain frameworks (Django, Ruby on Rails, etc.)
 - Compatibility support for Python 2.7 or 3.0 modules and what images support what.

## What is the relevance of it?

More openness as to what's available and compatible.

kadel:
The only images that we officially support are Java and NodeJS, and only those that are based on the images specified [here](https://github.com/openshift/odo-init-image/blob/master/language-scripts/image-mappings.json). Everything else is just as-is, we don't provide any guarantees there.

girishramnani:
The PR was merged and hence we can close this issue.

PR #2374, opened 2019-11-07
Fix README.adoc text about supported versions

### LINKED ISSUE #2178: Incorrect information about supported languages in readme
Readme says "Supports multiple languages and frameworks such as Node.js, Java, Ruby, Perl, PHP, and Python." But odo officially supports only Java and NodeJs, the other languages are without any guarantees. 

Not sure what the right wording should be.

pmacko1:
**What kind of PR is this?**

/kind docs

Fixes #2178 #2093

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -29,13 +29,12 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
 * Official support for Node.js and Java components.
 * Partial compatibility with languages and frameworks such as Ruby, Perl, PHP, and Python. 
 * Detects changes to local code and deploys it to the cluster automatically, giving instant feedback to validate changes in real time.
-* Lists all the available components and services from the OpenShift cluster.
+* Lists all the available components and services from the {product-title} cluster.
 
-=== Supported languages and corresponding container images
-Run `odo catalog list components` to list the supported languages and corresponding container images present on your cluster.
-
+[id="odo-supported-languages-and-images"]
 === Officially supported languages and corresponding container images
 
+This conversation was marked as resolved by pmacko1
 .Supported container images
 [options="header"]
 |====
@@ -48,14 +47,24 @@ Run `odo catalog list components` to list the supported languages and correspond
 |`rhscl/nodejs-10-rhel7`|
 |====
 
-.Listing available container images
-. Log in to the OpenShift cluster with `odo`:
+[id="odo-listing-available-images"]
+==== Listing available container images
+This conversation was marked as resolved by pmacko1
+
+[NOTE]
+====
+The list of available container images is sourced from the cluster's internal container registry and external registries associated with the cluster.
+====
+
+To list the available components and associated container images for your cluster:
+
+. Log in to the {product-title} cluster with {odo-title}:
 +
 ----
 $ odo login -u developer -p developer
 ----
 
-. List the languages and corresponding container images available for your cluster with `odo catalog list components`.
+. List the available {odo-title} supported and unsupported components and corresponding container images:
 +
 ----------------------------------------------------
 $ odo catalog list components
@@ -63,14 +72,13 @@ Odo Supported OpenShift Components:
 NAME        PROJECT      TAGS      
 java       openshift     8,latest
 nodejs     openshift     10,8,8-RHOAR,latest
-
 Odo Unsupported OpenShift Components:
 NAME                      PROJECT       TAGS
 dotnet                    openshift     1.0,1.1,2.1,2.2,latest
 fuse7-eap-openshift       openshift     1.3
 ----------------------------------------------------
 +
-The command lists the components and the languages in two sections - supported and unsupported by Red Hat. `TAGS` column represents the image version, for example `10` represents `rhoar-nodejs/nodejs-10`.
+The `TAGS` column represents the available image versions, for example, `10` represents the `rhoar-nodejs/nodejs-10` container image.
 
 [[official-documentation]]
 == Official documentation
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -24,11 +24,12 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
...
 `odo` is designed to be simple and concise with the following key features:
 
-* Simple syntax and design centered around concepts familiar to developers, such as project, application, and component.
+* Simple syntax and design centered around concepts familiar to developers, such as projects, applications, and components.
 * Completely client based. No server is required within the OpenShift cluster for deployment.
-* Supports multiple languages and frameworks such as Node.js, Java, Ruby, Perl, PHP, and Python.
-* Detects changes to local code and deploys it to the cluster automatically, giving instant feedback to validate changes in real-time.
-* Lists all available components and services from the OpenShift cluster.
+* Official full support of Node.js and Java components.
+* Partial compatibility with many languages and frameworks such as Ruby, Perl, PHP, and Python. Run odo catalog list components to see the list of supported and unsupported components.
```
boczkowska: Run odo catalog list components to see -> Run `odo catalog list components` to see

boczkowska: And remove "and unsupported components".

boczkowska:
@pmacko1 you can refer for tomas' https://github.com/openshift/odo/issues/2093#issuecomment-528777633 to get a list of supported images.

README.adoc
Preeticp: Run odo catalog list components to see... needs to e removed from this conceptual module. It should be part of a procedural module.

kadel:
> @kadel I am not sure but I vaguely remember a discussion/issue that talked about being more precise than just saying that odo supports Java because Java could mean a lot of things. By things I mean frameworks, versions, etc. not the coffee, of course. 😉

Technically we support any generic Java (OpenJDK) projects

README.adoc
pmacko1: according to @kadel `odo catalog list components` can have different outputs depending on the cluster's configuration. I'd remove this command altogether from here and find a new place to put it

pmacko1:
@kadel @amitkrout Hello guys, can I ask you for a review, please?

Thanks!

pmacko1:
@boczkowska or @Preeticp can I ask one of you for a review, please? :)

Preeticp:
@cdrage, @amitkrout  can you please take a look at this PR. 

Questions: 
1. Considering that we are adding a section telling the users about the supported languages and images right under the Key features sections, do we still want to add these points in the key features:
"- Official full support of Node.js and Java components.
 - Partial compatibility with languages and frameworks such as Ruby, Perl, PHP, and Python.
"
2. Also, is the first point technically correct? I remember Steve S pointing out that we do not support all the Java images. So would 'full support' imply the right thing to the user? Would it be safer if we simply remove these 2 points and just have the Supported languages and images section as done in this PR.

amitkrout:
> @cdrage, @amitkrout can you please take a look at this PR.
> 
> Question: Also, is the first point technically correct? I remember Steve S pointing out that we do not support all the Java images. So would 'full support' imply the right thing to the user? Would it be safer if we simply remove these 2 points and just have the Supported languages and images section as done in this PR.

AFAIK java image stream contains these supported container image like ```openjdk/openjdk-11-rhel7, ...``` and these container images are used when a java component is created and pushed. If some body wants to apply his/her won java image stream, then i am not sure in which category it falls in ```odo catalog list components```. More importantly we can not guaranty customize java or nodejs image stream having container image not same as we test on.

IMO instead of writing ```... full support ...``` i would recommend just remove ```full``` from the sentence.  @cdrage WDYT ?

pmacko1:
WIP, as the description of the command is not correct. I received an update from Tomas.

README.adoc
```diff
@@ -24,11 +24,26 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
...
-* Lists all available components and services from the OpenShift cluster.
+* Official full support of Node.js and Java components.
+* Partial compatibility with languages and frameworks such as Ruby, Perl, PHP, and Python. 
+* Detects changes to local code and deploys it to the cluster automatically, giving instant feedback to validate changes in real time.
+* Lists all the available components and services from the OpenShift cluster.
+
+=== Supported languages and corresponding container images
+Run `odo catalog list components` to list the supported languages and corresponding container images present on your cluster.
+
+.Supported container images
```
cdrage: Maybe `Officially supported container images` ?

README.adoc
```diff
@@ -24,11 +24,26 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
 
 `odo` is designed to be simple and concise with the following key features:
 
-* Simple syntax and design centered around concepts familiar to developers, such as project, application, and component.
+* Simple syntax and design centered around concepts familiar to developers, such as projects, applications, and components.
 * Completely client based. No server is required within the OpenShift cluster for deployment.
-* Supports multiple languages and frameworks such as Node.js, Java, Ruby, Perl, PHP, and Python.
-* Detects changes to local code and deploys it to the cluster automatically, giving instant feedback to validate changes in real-time.
-* Lists all available components and services from the OpenShift cluster.
+* Official full support of Node.js and Java components.
```
cdrage: remove the `full` ?

README.adoc
```diff
@@ -24,11 +24,26 @@ Existing tools such as `oc` are more operations-focused and require a deep-under
...
+
+.Supported container images
+[options="header"]
+|====
+|Node.js|Java
+|`centos/nodejs-8-centos7`|`redhat-openjdk-18/openjdk18-openshift`
+|`rhoar-nodejs/nodejs-8`|`openjdk/openjdk-11-rhel8`
+|`rhoar-nodejs/nodejs-10`|`openjdk/openjdk-11-rhel7`
+|`bucharestgold/centos7-s2i-nodejs`|
+|`rhscl/nodejs-8-rhel7`|
```
amitkrout: Also include image ```rhscl/nodejs-10-rhel7```.

Reference - https://github.com/openshift/odo-init-image/pull/43/commits/0ce478aefc8f46813ec0b94187e537ec5d0a40b4#diff-23c2abb47c564635a127893860b3cb08R6

README.adoc
pmacko1: might be a little redundant, maybe. We have "compatibility" vs "support" there now.

On a second thought, we might want to include unsupported languages there as a new table as well. But explicitly state, that no support/compatiblity is guaranteed. WDYT?

README.adoc
pmacko1: yep

README.adoc
pmacko1: :+1:

pmacko1:
@amitkrout would `Official support of Node.js and Java components based on Red Hat images.` be better worded?

PR #2384, opened 2019-11-08
Build and package for all RHEL architectures

yselkowitz:
**What kind of PR is this?**
/kind enhancement

OpenShift is currently available for X86-64 and IBM Power LE, and IBM Z is coming.  IBM customers will expect to be able to download and run odo on their systems (particularly on Z where admins generally prefer homogeneous environments).  oc is also available on all RHEL architectures:

https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.2/

The same should apply to odo, particularly as download links thereto are provided from the console.

A successful scratch build has been done internally.

### DIFF
--- a/rpms/openshift-odo.spec
+++ b/rpms/openshift-odo.spec
@@ -19,8 +19,6 @@ Summary:        %{product_name} client odo CLI binary
 License:        ASL 2.0
 URL:            https://github.com/openshift/odo/tree/%{odo_cli_version}
 
-ExclusiveArch:  x86_64
-
 Source0:        %{source_tar}
 BuildRequires:  gcc
 BuildRequires:  golang >= %{golang_version}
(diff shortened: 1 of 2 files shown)


rpms/openshift-odo.spec
```diff
@@ -32,24 +30,31 @@ OpenShift Do (odo) is a fast, iterative, and straightforward CLI tool for develo
 
 %prep
 %setup -q -n %{source_dir}
+%patch0 -p1
```
mohammedzee1000: Where exactly is this patch coming from?

yselkowitz: Sorry, that was left over from earlier testing, fixed.

mohammedzee1000:
Looking good
/approve

kadel:
Not sure if the alternative platform makes sense. odo is a developer tool, meant to be run on a developer's laptop. I don't expect anyone to run odo on IBM Z system

PR #2391, opened 2019-11-11
Update doc since Codewind project already used odo

GeekArthur:

**What kind of PR is this?**

/kind docs

Currently Codewind already supported OpenShift projects and can use odo as build engine to build projects, so we can add Codewind as `projects using odo` on official odo website.

Issue in Codewind repository: https://github.com/eclipse/codewind/issues/754


mohammedzee1000:
This looks good. Also, minor docs change
/approve
/lgtm

Issue #1884, opened 2019-07-04
Implement `--wait` flag for `odo project delete`


surajnarwade:
[kind/Enhancement]

## Which functionality do you think we should update/improve?

Add `--wait` flag in `odo project delete` similar to `odo project create`

## Why is this needed? 

Reference: `waitForProjectToBeGone` from https://github.com/openshift/odo/issues/1739

as of now, vscode plugin relies on,

```
oc wait project/${project} --for delete
```

surajnarwade:
cc @kadel @cdrage @dgolovin

cdrage:
ping @girishramnani

cdrage:
Se my current implementation: #2397 

I'm having major troubles mitigating ONE niche case and that's essentially when you run `odo project create` IMMEDIATELY after `odo project delete` with the same project name.

Here's essentially the TLDR;

- `--wait` works. And works well
- if you do project create with the same name as the project you just deleted, it *may* fail as your cluster is still propagating changes..

The project is deleted asynchronously by OpenShift. We cannot control this. OpenShift will delete all the pods + project + etc when it wants. No amount of code can fix this. My implementation #2397 even waits for **all** pods to be deleted, but to no avail. You must wait until the project is *completely* closed by OpenShift.

This is actually a "feature" of OpenShift and no amount of code would fix it. For example, if you were to create a watch for Project or even Pods (from the deleted project) it'd fail. OpenShift says it's deleted.. `oc get projects` also shows blank... but when you do `oc create project foobar`, it'll error out!

Here's an explanation of what happens by Will Gordon @ Red Hat from https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o

```
Projects are deleted asynchronously after you send the delete command. So it's possible that the deletion just hasn't been reconciled yet. It should happen within a minute or so, so try again.

Also, please be aware that in a multitenant environment, like OpenShift Online, you are prevented from creating a project with the same name as any other project in the cluster, even if it's not your own. So if you can't create the project, it's possible that someone has already created a project with the same name.
```

So for the niche case of:

```sh
odo project delete foobar --wait
# create a few components here
odo project create foobar
```

We cannot support it. You will have to literally wait for your cluster to automatically delete the project. In the case of `minishift` or `crc` or something local, this is done almost instantly. But when you are dealing with multi-tenant environments such as OpenShift Online or our internal test cluster, you'll just have to wait for OpenShift to finish. 

[...]

PR #2397, opened 2019-11-12
Add --wait to project delete

cdrage:
**What kind of PR is this?**

/kind feature
/kind enhancement

Adds `--wait` parameter. Ex. `odo project delete myproject --wait`

Fixes https://github.com/openshift/odo/issues/1884

```sh
odo project create foobar
odo project delete foobar --wait
```

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -103,6 +103,9 @@ const (
 	// timeout for getting the default service account
 	getDefaultServiceAccTimeout = 1 * time.Minute
 
+	// timeout for waiting for project deletion
+	waitForProjectDeletionTimeOut = 3 * time.Minute
+
 	// The length of the string to be generated for names of resources
 	nameLength = 5
 
@@ -2214,67 +2217,102 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
 }
 
 // DeleteProject deletes given project
-func (c *Client) DeleteProject(name string) error {
-	err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})
+//
+// NOTE:
+// There is a very specific edge case that may happen during project deletion when deleting a project and then immediately creating another.
+// Unfortunately, despite the watch interface, we cannot safely determine if the project is 100% deleted. See this link:
+// https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o
+// Will Gordon (Engineer @ Red Hat) describes the issue:
+//
+// "Projects are deleted asynchronously after you send the delete command. So it's possible that the deletion just hasn't been reconciled yet. It should happen within a minute or so, so try again.
+// Also, please be aware that in a multitenant environment, like OpenShift Online, you are prevented from creating a project with the same name as any other project in the cluster, even if it's not your own. So if you can't create the project, it's possible that someone has already created a project with the same name."
+func (c *Client) DeleteProject(name string, wait bool) error {
+
+	// Instantiate watcher for our "wait" function
+	var watcher watch.Interface
+	var err error
+
+	// If --wait has been passed, we will wait for the project to fully be deleted
+	if wait {
+		watcher, err = c.projectClient.Projects().Watch(metav1.ListOptions{
+			FieldSelector: fields.Set{"metadata.name": name}.AsSelector().String(),
+		})
+		if err != nil {
+			return errors.Wrapf(err, "unable to watch project")
+		}
+		defer watcher.Stop()
+	}
+
+	// Delete the project
+	err = c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})
 	if err != nil {
 		return errors.Wrap(err, "unable to delete project")
 	}
 
-	// wait for delete to complete
-	w, err := c.projectClient.Projects().Watch(metav1.ListOptions{
-		FieldSelector: fields.Set{"metadata.name": name}.AsSelector().String(),
-	})
-	if err != nil {
-		return errors.Wrapf(err, "unable to watch project")
+	// If watcher has been created (wait was passed) we will create a go routine and actually **wait**
+	// until *EVERYTHING* is successfully deleted.
+	if watcher != nil {
+
+		// Project channel
+		// Watch error channel
+		projectChannel := make(chan *projectv1.Project)
+		watchErrorChannel := make(chan error)
+
+		// Create a go routine to run in the background
+		go func() {
+
+			for {
+
+				// If watch unexpected has been closed..
+				val, ok := <-watcher.ResultChan()
+				if !ok {
+					//return fmt.Errorf("received unexpected signal %+v on project watch channel", val)
+					watchErrorChannel <- errors.Errorf("watch channel was closed unexpectedly: %+v", val)
+					break
+				}
+
+				// So we depend on val.Type as val.Object.Status.Phase is just empty string and not a mapped value constant
+				if projectStatus, ok := val.Object.(*projectv1.Project); ok {
+
+					klog.V(4).Infof("Status of delete of project %s is '%s'", name, projectStatus.Status.Phase)
+
+					switch projectStatus.Status.Phase {
+					//projectStatus.Status.Phase can only be "Terminating" or "Active" or ""
+					case "":
+						if val.Type == watch.Deleted {
+							projectChannel <- projectStatus
+							break
+						}
+						if val.Type == watch.Error {
+							watchErrorChannel <- errors.Errorf("failed watching the deletion of project %s", name)
+							break
+						}
+					}
+
+				} else {
+					watchErrorChannel <- errors.New("unable to convert event object to Project")
+					break
+				}
+
+			}
+			close(projectChannel)
+			close(watchErrorChannel)
+		}()
+
+		select {
+		case val := <-projectChannel:
+			klog.V(4).Infof("Deletion information for project: %+v", val)
+			return nil
+		case err := <-watchErrorChannel:
+			return err
+		case <-time.After(waitForProjectDeletionTimeOut):
+			return errors.Errorf("waited %s but couldn't delete project %s in time", waitForProjectDeletionTimeOut, name)
+		}
+
 	}
 
-	defer w.Stop()
-	for {
-		val, ok := <-w.ResultChan()
-		// When marked for deletion... val looks like:
-		/*
-			val: {
-				Type:MODIFIED
-				Object:&Project{
-					ObjectMeta:k8s_io_apimachinery_pkg_apis_meta_v1.ObjectMeta{...},
-					Spec:ProjectSpec{...},
-					Status:ProjectStatus{
-						Phase:Terminating,
-					},
-				}
-			}
-		*/
-		// Post deletion val will look like:
-		/*
-			val: {
-				Type:DELETED
-				Object:&Project{
-					ObjectMeta:k8s_io_apimachinery_pkg_apis_meta_v1.ObjectMeta{...},
-					Spec:ProjectSpec{...},
-					Status:ProjectStatus{
-						Phase:,
-					},
-				}
-			}
-		*/
-		if !ok {
-			return fmt.Errorf("received unexpected signal %+v on project watch channel", val)
-		}
-		// So we depend on val.Type as val.Object.Status.Phase is just empty string and not a mapped value constant
-		if prj, ok := val.Object.(*projectv1.Project); ok {
-			klog.V(4).Infof("Status of delete of project %s is %s", name, prj.Status.Phase)
-			switch prj.Status.Phase {
-			//prj.Status.Phase can only be "Terminating" or "Active" or ""
-			case "":
-				if val.Type == watch.Deleted {
-					return nil
-				}
-				if val.Type == watch.Error {
-					return fmt.Errorf("failed watching the deletion of project %s", name)
-				}
-			}
-		}
-	}
+	// Return nil since we don't bother checking for the watcher..
+	return nil
 }
 
 // GetDeploymentConfigLabelValues get label values of given label from objects in project that are matching selector
--- a/pkg/odo/cli/project/delete.go
+++ b/pkg/odo/cli/project/delete.go
@@ -35,6 +35,10 @@ type ProjectDeleteOptions struct {
 
 	// generic context options common to all commands
 	*genericclioptions.Context
+
+	// wait is a boolean value to choose if we wait or not for
+	// our project to be deleted
+	wait bool
 }
 
 // NewProjectDeleteOptions creates a ProjectDeleteOptions instance
@@ -62,6 +66,9 @@ func (pdo *ProjectDeleteOptions) Validate() (err error) {
 // Run runs the project delete command
 func (pdo *ProjectDeleteOptions) Run() (err error) {
 
+	// Create the "spinner"
+	s := &log.Status{}
+
 	// This to set the project in the file and runtime
 	err = project.SetCurrent(pdo.Context.Client, pdo.projectName)
 	if err != nil {
@@ -77,15 +84,23 @@ func (pdo *ProjectDeleteOptions) Run() (err error) {
 	if log.IsJSON() || (pdo.projectForceDeleteFlag || ui.Proceed(fmt.Sprintf("Are you sure you want to delete project %v", pdo.projectName))) {
 		successMessage := fmt.Sprintf("Deleted project : %v", pdo.projectName)
 
-		err := project.Delete(pdo.Context.Client, pdo.projectName)
+		// If the --wait parameter has been passed, we add a spinner..
+		if pdo.wait {
+			s = log.Spinner("Waiting for project to be deleted")
+			defer s.End(false)
+		}
+
+		err := project.Delete(pdo.Context.Client, pdo.projectName, pdo.wait)
 		if err != nil {
 			return err
 		}
+		s.End(true)
 
 		if log.IsJSON() {
 			project.MachineReadableSuccessOutput(pdo.projectName, successMessage)
 		} else {
 			log.Success(successMessage)
+			log.Warning("Warning! Projects are deleted from the cluster asynchronously. Odo does its best to delete the project. Due to multi-tenant clusters, the project may still exist on a different node.")
 		}
 		return nil
 	}
@@ -109,6 +124,7 @@ func NewCmdProjectDelete(name, fullName string) *cobra.Command {
 		},
 	}
 
+	projectDeleteCmd.Flags().BoolVarP(&o.wait, "wait", "w", false, "Wait until the project has been completely deleted")
 	projectDeleteCmd.Flags().BoolVarP(&o.projectForceDeleteFlag, "force", "f", false, "Delete project without prompting")
 
 	return projectDeleteCmd
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -4,7 +4,6 @@ import (
 	"github.com/openshift/odo/pkg/machineoutput"
 	"github.com/pkg/errors"
 
-	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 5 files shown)


tests/integration/project/cmd_project_test.go
```diff
@@ -38,6 +39,40 @@ var _ = Describe("odo project command tests", func() {
...
+		})
+
+		It("--wait should work with deleting a project", func() {
+
+			// Create the project
+			helper.CmdShouldPass("odo", "project", "create", projectName)
+
+			// Delete with --wait
+			output := helper.CmdShouldPass("odo", "project", "delete", projectName, "-f", "--wait")
+			Expect(output).To(ContainSubstring("Waiting for project to be deleted"))
```
mik-dass: We should also list the projects and test that the project deletion was successful.

tests/integration/project/cmd_project_test.go
```diff
@@ -38,6 +39,40 @@ var _ = Describe("odo project command tests", func() {
...
+			projectName = helper.RandString(6)
+		})
+
+		// odo project delete foobar -o json
+		It("should be able to delete project and show output in json format", func() {
+			helper.CmdShouldPass("odo", "project", "create", projectName, "-o", "json")
+
+			actual := helper.CmdShouldPass("odo", "project", "delete", projectName, "-o", "json")
+			desired := fmt.Sprintf(`{"kind":"Project","apiVersion":"odo.openshift.io/v1alpha1","metadata":{"name":"%s","namespace":"%s","creationTimestamp":null},"message":"Deleted project : %s"}`, projectName, projectName, projectName)
+			Expect(desired).Should(MatchJSON(actual))
```
mik-dass: same as above

pkg/odo/cli/project/delete.go
```diff
@@ -109,6 +123,7 @@ func NewCmdProjectDelete(name, fullName string) *cobra.Command {
 		},
 	}
 
+	projectDeleteCmd.Flags().BoolVarP(&o.wait, "wait", "w", false, "Wait until the project completely deleted")
```
mik-dass: `Wait until the project has been completely deleted`

dharmit:
Haven't reviewed the code yet. Was checking out the functionality first and stumbled upon this:
```sh

$ odo project delete -w myproject
This project contains the following applications, which will be deleted
Application app
This application has following components that will be deleted
component named nodejs-nodejs-ex-xyuo
? Are you sure you want to delete project myproject Yes
 ✓  Waiting for project to be deleted [6s]
 ✓  Deleted project : myproject

$ odo project list
ACTIVE     NAME
           newproject

$ odo project create myproject
 ✗  project.project.openshift.io "myproject" already exists

# Tried again after some time
$ odo project create myproject
 ✓  Project 'myproject' is ready for use
 ✓  New project created and now using project: myproject
```
Why doesn't it let me create `myproject` immediately if we waited for it to get deleted? :thinking:

cdrage:
@dharmit Yeah, something is weird is happening here... I'm going to investigate it, seems like our watch code doesn't work correctly!

/hold

Thanks

tests/integration/project/cmd_project_test.go
cdrage: This is testing for using odo project delete without `--wait`, so we don't bother checking `odo project list` or `oc get projects` here.

cdrage:
/hold cancel

This is ready for review! Unit and integration tests have been added.

Ping @girishramnani @amitkrout @dharmit

pkg/occlient/occlient.go
```diff
@@ -2006,67 +2009,103 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
...
+	// until *EVERYTHING* is successfully deleted.
+	if watcher != nil {
+
+		// Project channel
+		// Watch error channel
+		projectChannel := make(chan *projectv1.Project)
+		watchErrorChannel := make(chan error)
+
+		// Create a go routine to run in the background
+		go func() {
```
mik-dass: Why are we using a separate go routine here?

pkg/occlient/occlient.go
```diff
@@ -2006,67 +2009,103 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
...
+
+		loop:
```
mik-dass: Why are we using a label here? Since it's a single loop, why not break normally?

pkg/occlient/occlient.go
```diff
@@ -2006,67 +2009,103 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
...
+						}
+					}
+
+				} else {
+					watchErrorChannel <- errors.New("unable to convert event object to Project")
+					break loop
 				}
+
 			}
+			close(projectChannel)
```
mik-dass: Maybe close it right after creation like `defer close(projectChannel)`?

pkg/occlient/occlient.go
cdrage: We should be doing this in the background and then using `select`, no other explanation really, but this was previously not implemented correctly (it's really bad to have an endless for loop without a break for timeout)

pkg/occlient/occlient.go
cdrage: Removing this :+1: you're right.

pkg/occlient/occlient.go
cdrage: IMO, I like keeping the close statements at the end instead of the beginning (especially since this just for the go routine).

cdrage:
Looks like a failed unit test from my updating the tests. Fixing :+1: but feel free to still review.

tests/integration/project/cmd_project_test.go
```diff
@@ -38,6 +39,44 @@ var _ = Describe("odo project command tests", func() {
...
+
+			// Create the project
+			helper.CmdShouldPass("odo", "project", "create", projectName)
+
+			// Delete with --wait
+			output := helper.CmdShouldPass("odo", "project", "delete", projectName, "-f", "--wait")
+			Expect(output).To(ContainSubstring("Waiting for project to be deleted"))
+
+			// Check that it's NOT there..
+			output = helper.CmdShouldPass("odo", "project", "list")
```
mik-dass: This seems to run into a flake due to the explanation given in the note of the DeleteProject() function
https://travis-ci.com/openshift/odo/jobs/259258747#L3847

I guess we may need to add a sleep timeout?

pkg/occlient/occlient.go
mik-dass: But we can also add a timeout with a non go routine approach like https://github.com/openshift/odo/blob/adf3006360abc5cb707d9a9f442669989272f4e5/pkg/occlient/occlient.go#L1715-L1722

pkg/occlient/occlient.go
mik-dass: https://groups.google.com/forum/#!topic/golang-nuts/Hj7-HV-W_iU
I went through the above discussion. If a panic happens in the go routine, we wouldn't reach the close statements and thus the channels won't close.

pkg/occlient/occlient.go
cdrage: The above solution is similar to how we have it with PodTimeout thus we have it in it's own separate go routine. No, we don't have to have a go routine, but I find that this is more elegant / less CPU usage when using an endless loop as a blocker.

pkg/occlient/occlient.go
cdrage: Okay, this is how we currently have it for the PodTimeout go routine that's been written. Should we change this in both locations?

mik-dass: I would suggest a different PR but I leave the decision up to you

tests/integration/project/cmd_project_test.go
```diff
@@ -47,6 +48,44 @@ var _ = Describe("odo project command tests", func() {
...
+		It("--wait should work with deleting a project", func() {
+
+			// Create the project
+			helper.CmdShouldPass("odo", "project", "create", projectName)
+
+			// Delete with --wait
+			output := helper.CmdShouldPass("odo", "project", "delete", projectName, "-f", "--wait")
+			Expect(output).To(ContainSubstring("Waiting for project to be deleted"))
+
+			// Check that it's NOT there..
```
mik-dass: comments found

cdrage:
More terraform flakes. Re-running test (doing a rebase anyways).

mik-dass:
```
Switching to project : wqbmnxxswf
Running oc with args [oc project wqbmnxxswf]
[oc] error: You are not a member of project "wqbmnxxswf".
[oc] Your projects are:
[oc] * eahmcdjvon
[oc] * fojpnnxkeg
[oc] * mukfineelg
[oc] * myproject
• Failure in Spec Teardown (AfterEach) [64.046 seconds]
Basic benchmark
/go/src/github.com/openshift/odo/tests/benchmark/basic_benchmark.go:13
  Simple NodeJS component [AfterEach]
```
/retest

cdrage:
Terraform / AWS errors (again)

/retest

mik-dass:
```
✗  image stream wildfly with tag latest not found in openshift and qjpghikobp namespaces
Deleting project: qjpghikobp
```
/retest

cdrage:
```sh
2019/12/13 18:57:50 Container test in pod integration-e2e-benchmark failed, exit code 2, reason Error
2019/12/13 19:04:13 Copied 176.18Mi of artifacts from integration-e2e-benchmark to /logs/artifacts/integration-e2e-benchmark
2019/12/13 19:04:13 Releasing lease for "aws-quota-slice"
2019/12/13 19:07:07 No custom metadata found and prow metadata already exists. Not updating the metadata.
2019/12/13 19:07:09 Ran for 1h51m53s
2019/12/13 19:07:09 Submitted failure event to sentry (id=30db77c4f7414948a694035578ef0c17)
error: could not run steps: step integration-e2e-benchmark failed: template pod "integration-e2e-benchmark" failed: the pod ci-op-0qyjtsgl/integration-e2e-benchmark failed after 1h41m20s (failed containers: test): ContainerFailed one or more containers exited
```

¯\_(ツ)_/¯

pkg/occlient/occlient.go
kadel: > Maybe close it right after creation like `defer close(projectChannel)`?

+1

cdrage:
Seriously terraform :(

/retest

kadel:
What is the situation with this PR?

girishramnani:
a project delete cleanup is failing
```
Running odo with args [odo project delete dpjrlpomrg -f]
[odo]  ✗  invalid configuration: [context was not found for specified context: adbrupdkcr/127-0-0-1:8443/developer, cluster has no server defined]
```
not sure if its a flake

cdrage:
Here is the current update on this PR and honestly, I'm unsure on how to proceed...

The consensus and according to: https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o is that honestly, we cannot safely determine if a project is *ever* deleted within OpenShift since it's deleted asynchronously.

For example, in a multi-tenant environment, after issuing `odo project delete foobar --wait`, even *with* the watcher confirming it's deleted: https://github.com/openshift/odo/pull/2397/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR2101 the project can potentially *still* exist in the cluster.

So even with this PR in, it won't fix the issue.

I'm not saying this because I'm giving up, but this could be one of the few issues we have to mark as WONTFIX because we cannot safely determine if a project has been deleted.

cdrage:
ping @girishramnani @mik-dass @kadel  (see above comment)

mik-dass:
> I'm not saying this because I'm giving up, but this could be one of the few issues we have to mark as WONTFIX because we cannot safely determine if a project has been deleted.

:+1:

cdrage:
> /retest

See above comment, retesting won't work :rofl: 

Need everyone's opinion @kadel @amitkrout on how to move forward with this. See above comment(s).

kadel:
> Here is the current update on this PR and honestly, I'm unsure on how to proceed...
> 
> The consensus and according to: https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o is that honestly, we cannot safely determine if a project is _ever_ deleted within OpenShift since it's deleted asynchronously.
> 
> For example, in a multi-tenant environment, after issuing `odo project delete foobar --wait`, even _with_ the watcher confirming it's deleted: https://github.com/openshift/odo/pull/2397/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR2101 the project can potentially _still_ exist in the cluster.
> 
> So even with this PR in, it won't fix the issue.
> 
> I'm not saying this because I'm giving up, but this could be one of the few issues we have to mark as WONTFIX because we cannot safely determine if a project has been deleted.

Let's do as much as we can on odo side (using watch). And then leave the rest on the user to handle it, as we can't do anything more than that.

I would also put a warning at the end of the `odo project delete foobar --wait` command output.
Someting like this, but better ;-):
```
Warning: Project might still exist because project deletion is handled asynchronously on the cluster side.
```

cdrage:
> > Here is the current update on this PR and honestly, I'm unsure on how to proceed...
> > The consensus and according to: https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o is that honestly, we cannot safely determine if a project is _ever_ deleted within OpenShift since it's deleted asynchronously.
> > For example, in a multi-tenant environment, after issuing `odo project delete foobar --wait`, even _with_ the watcher confirming it's deleted: https://github.com/openshift/odo/pull/2397/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR2101 the project can potentially _still_ exist in the cluster.
> > So even with this PR in, it won't fix the issue.
> > I'm not saying this because I'm giving up, but this could be one of the few issues we have to mark as WONTFIX because we cannot safely determine if a project has been deleted.
> 
> Let's do as much as we can on odo side (using watch). And then leave the rest on the user to handle it, as we can't do anything more than that.
> 
> I would also put a warning at the end of the `odo project delete foobar --wait` command output.
> Someting like this, but better ;-):
> 
> ```
> Warning: Project might still exist because project deletion is handled asynchronously on the cluster side.
> ```

That's a good point. We could add a warning..

The problem is that we're not going to have any integration tests, is that okay?

mik-dass:
> The problem is that we're not going to have any integration tests, is that okay?

Maybe use sleep or the helper function helper.WaitForCmdOut()?

cdrage:
@mik-dass 

Even with sleep, it won't work.. We'd have to essentially spam `odo project create` until it doesn't error out. It's a pain.

I'm going to modify this PR.

mik-dass:
> Even with sleep, it won't work.. We'd have to essentially spam odo project create until it doesn't error out. It's a pain.

I am OK with it. @amitkrout WDYT?

cdrage:
:( Travis passes, but not benchmark..

/retest
(again)

kadel:
Where we are on this?

cdrage:
@kadel 

CI flakes causing none of the benchmark tests to pass:

```sh
Ginkgo ran 1 suite in 2m27.664546656s
Test Suite Failed
make: *** [test-benchmark] Error 1
2020/04/14 18:23:09 Container test in pod integration-e2e-benchmark failed, exit code 2, reason Error
2020/04/14 18:33:20 Copied 183.35MB of artifacts from integration-e2e-benchmark to /logs/artifacts/integration-e2e-benchmark
2020/04/14 18:33:20 Releasing lease for "aws-quota-slice"
2020/04/14 18:33:23 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/04/14 18:33:24 Ran for 2h22m29s
2020/04/14 18:33:24 Submitted failure event to sentry (id=b7a7022e7b354b20aa3ac4ef2661e6ea)
error: could not run steps: step integration-e2e-benchmark failed: template pod "integration-e2e-benchmark" failed: the pod ci-op-svkmgjvt/integration-e2e-benchmark failed after 2h5m6s (failed containers: test): ContainerFailed one or more containers exited
```

girishramnani:
any update on this @cdrage ?

cdrage:
Ayyyy, all the tests pass.

Going to rebase and this should be good to go for another set of reviews!

pkg/occlient/occlient.go
```diff
@@ -102,6 +102,9 @@ const (
 	// timeout for getting the default service account
 	getDefaultServiceAccTimeout = 1 * time.Minute
 
+	// timeout for waiting for project deletion
+	waitForProjectDeletionTimeOut = 3 * time.Minute
```
dharmit, line 106: Can user modify this timeout by updating odo preferences? If no, should we allow to do it?

cc @kadel @girishramnani

pkg/occlient/occlient.go
```diff
@@ -2206,67 +2209,102 @@ func (c *Client) DeleteServiceInstance(labels map[string]string) error {
...
-func (c *Client) DeleteProject(name string) error {
-	err := c.projectClient.Projects().Delete(name, &metav1.DeleteOptions{})
-	if err != nil {
-		return errors.Wrap(err, "unable to delete project")
+//
+// NOTE:
+// There is a very specific edge case that may happen during project deletion when deleting a project and then immediately creating another.
+// Unfortunately, despite the watch interface, we cannot safely determine if the project is 100% deleted. See this link:
+// https://stackoverflow.com/questions/48208001/deleted-openshift-online-pro-project-has-left-a-trace-so-cannot-create-project-o
+// Will Gordon (Engineer @ RedHat) describes the issue:
```
dharmit, line 2217: Couldn't resist this one. :wink: `s/RedHat/Red Hat/`

pkg/project/project.go
```diff
@@ -43,18 +42,16 @@ func Create(client *occlient.Client, projectName string, wait bool) error {
 }
 
 // Delete deletes the project with name projectName and returns errors if any
-func Delete(client *occlient.Client, projectName string) error {
-	// Loading spinner
-	s := log.Spinnerf("Deleting project %s", projectName)
-	defer s.End(false)
+func Delete(client *occlient.Client, projectName string, wait bool) error {
+	if projectName == "" {
```
dharmit, line 46: I think the cobra package handles this condition for us already. Is this check going to be helpful to IDE plugins folks? If not, we could remove it, I guess.
```sh
$ odo project delete --wait
Error: accepts 1 arg(s), received 0
Usage:
  odo project delete [flags]
```

tests/integration/project/cmd_project_test.go
dharmit: +1. We should check if this worked as expected.

pkg/occlient/occlient.go
cdrage: Heck, you're right. We can add this.

tests/integration/project/cmd_project_test.go
mik-dass: @cdrage Maybe wait for the required conditions to be met using `WaitForCmdOut`

pkg/occlient/occlient.go
mik-dass: @cdrage Can you please create a issue regarding this discussion?

cdrage:
*Sigh* another rebase. Hopefully tests pass :+1:

dharmit:
@cdrage I think the points raised by me & @mik-dass in the review have not been addressed yet. Or am I missing something here?

cdrage:
> @cdrage I think the points raised by me & @mik-dass in the review have not been addressed yet. Or am I missing something here?

You're right. Not yet resolved. I'll ping again for another review when ready.

tests/integration/project/cmd_project_test.go
cdrage: @mik-dass @dharmit Unfortunately I don't think we can with the whole project deletion being asynchronous...

cdrage:
Pushed the changes, can you please /lgtm @dharmit ?

pkg/occlient/occlient.go
dharmit: Opened a user story https://github.com/openshift/odo/issues/3199.

dharmit:
> Pushed the changes, can you please /lgtm @dharmit ?

Can you please provide a response to https://github.com/openshift/odo/pull/2397#discussion_r420575744?

pkg/project/project.go
cdrage: Actually, this is for nil pointer checking / when using the test suite. 

Yes, Cobra covers it. But this was added in case the function was used somewhere else / testing. So we don't encounter any nil point errors.

dharmit:
/hold

@cdrage I think you'll need to change the value of `apiVersion` we expect in result. Please take a look at the error in the tests:
```
 • Failure [3.505 seconds]
odo project command tests
/go/src/github.com/openshift/odo/tests/integration/project/cmd_project_test.go:15
  Delete the project with flag -o json
  /go/src/github.com/openshift/odo/tests/integration/project/cmd_project_test.go:71
    should be able to delete project and show output in json format [It]
    /go/src/github.com/openshift/odo/tests/integration/project/cmd_project_test.go:78
    Expected
      
        <string>: {
          "kind": "Project",
          "metadata": {
            "name": "dvxmph",
            "namespace": "dvxmph",
            "creationTimestamp": null
          },
          "message": "Deleted project : dvxmph"
        }
    to match JSON of
        <string>: {
          "kind": "Project",
          "metadata": {
            "name": "dvxmph",
            "namespace": "dvxmph",
            "creationTimestamp": null
          },
          "message": "Deleted project : dvxmph"
        }
        
    
    first mismatched key: "apiVersion"
    /go/src/github.com/openshift/odo/tests/integration/project/cmd_project_test.go:83
```

dharmit:
Removed the lgtm label since tests would keep getting triggered automatically otherwise.

PR #2402, opened 2019-11-19
E2E test for every supported images

### LINKED ISSUE #2364: e2e test for every supported image
We should create an e2e test that validates that the images that we claim are supported work.

We can start with something simpler covering just the core flow of editing files (no storage or services):
 - odo create
 - odo push 
 - validate
 - odo url create
 - odo push 
 - validate
 - edit source
 - odo push
 - validate
 - odo watch
   - edit source
   - validate
 - stop odo watch
 - odo delete 
 - validate

list of supported images https://github.com/openshift/odo/blob/425d77670264efc502e02aed28a5665546bfefd8/pkg/catalog/catalog.go#L282-L290

/area testing
/priority medium

prietyc123:
**What kind of PR is this?**

/kind test

Tests the e2e scenarios of supported images.

Fixes #2364 

```make test-e2e-images```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ test-e2e-java:
 test-e2e-source:
 	ginkgo $(GINKGO_FLAGS) -focus="odo source e2e tests" tests/e2escenarios/
 
-# Run image e2e tests
+# Run supported images e2e tests
 .PHONY: test-e2e-images
 test-e2e-images:
 	ginkgo $(GINKGO_FLAGS) -focus="odo supported images e2e tests" tests/e2escenarios/
--- a/Makefile
+++ b/Makefile
@@ -228,6 +228,11 @@ test-e2e-java:
 test-e2e-source:
 	ginkgo $(GINKGO_FLAGS) -focus="odo source e2e tests" tests/e2escenarios/
 
+# Run image e2e tests
+.PHONY: test-e2e-images
+test-e2e-images:
+	ginkgo $(GINKGO_FLAGS) -focus="odo supported images e2e tests" tests/e2escenarios/
+
 # Run all e2e test scenarios
 .PHONY: test-e2e-all
 test-e2e-all:
--- a/tests/e2escenarios/e2e_images_test.go
+++ b/tests/e2escenarios/e2e_images_test.go
@@ -39,25 +39,25 @@ var _ = Describe("odo supported images e2e tests", func() {
 		go func() {
 			startMsg := <-startSimulationCh
 			if startMsg {
-				err := os.MkdirAll(context+"/.abc", 0755)
+				err := os.MkdirAll(filepath.Join(context, ".abc"), 0755)
 				if err != nil {
 					panic(err)
 				}
-				err = os.MkdirAll(context+"/abcd", 0755)
+				err = os.MkdirAll(filepath.Join(context, "abcd"), 0755)
 				if err != nil {
 					panic(err)
 				}
-				_, err = os.Create(context + "/a.txt")
+				_, err = os.Create(filepath.Join(context, "a.txt"))
 				if err != nil {
 					panic(err)
 				}
 
-				helper.DeleteDir(context + "/abcd")
+				helper.DeleteDir(filepath.Join(context, "abcd"))
 
 				if srcType == "openjdk" {
-					helper.ReplaceString(filepath.Join(context+"/src/main/java/MessageProducer.java"), "Hello", "Hello odo")
+					helper.ReplaceString(filepath.Join(context, "src", "main", "java", "MessageProducer.java"), "Hello", "Hello odo")
 				} else {
-					helper.ReplaceString(filepath.Join(context+"/server.js"), "Hello", "Hello odo")
+					helper.ReplaceString(filepath.Join(context, "server.js"), "Hello", "Hello odo")
 				}
 			}
 		}()
@@ -72,7 +72,7 @@ var _ = Describe("odo supported images e2e tests", func() {
 					podName := oc.GetRunningPodNameOfComp(srcType+"-app", project)
 					envs := oc.GetEnvs(srcType+"-app", appName, project)
 					dir := envs["ODO_S2I_SRC_BIN_PATH"]
-					stdOut := oc.ExecListDir(podName, project, dir+"/src")
+					stdOut := oc.ExecListDir(podName, project, filepath.Join(dir, "src"))
 					Expect(stdOut).To(ContainSubstring(("a.txt")))
 					Expect(stdOut).To(Not(ContainSubstring("abcd")))
 				}
@@ -116,11 +116,11 @@ var _ = Describe("odo supported images e2e tests", func() {
 
 		// edit source and validate
 		if srcType == "openjdk" {
-			helper.ReplaceString(filepath.Join(context+"/src/main/java/MessageProducer.java"), "Hello", "Hello Java UPDATED")
+			helper.ReplaceString(filepath.Join(context, "src", "main", "java", "MessageProducer.java"), "Hello", "Hello Java UPDATED")
 			helper.CmdShouldPass("odo", "push", "--context", context)
 			helper.HttpWaitFor(routeURL, "Hello Java UPDATED", 90, 1)
 		} else {
-			helper.ReplaceString(filepath.Join(context+"/server.js"), "Hello", "Hello nodejs UPDATED")
+			helper.ReplaceString(filepath.Join(context, "server.js"), "Hello", "Hello nodejs UPDATED")
 			helper.CmdShouldPass("odo", "push", "--context", context)
 			helper.HttpWaitFor(routeURL, "Hello nodejs UPDATED", 90, 1)
 		}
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by prietyc123** [comment]
@kadel It seems prow does not handle the authentication token for registry.redhat.io for image ```openjdk-11-rhel8``` .

I have created an upstream issue https://github.com/openshift/release/issues/6056 to get an idea of fixing it more efficiently.

**Comment by prietyc123** [comment]
> Add the same e2e-image test target to travis configuration

Done

**Comment by prietyc123** [comment]
Initiated conversation regarding authetication token https://coreos.slack.com/archives/CBN38N3MW/p1574868814277000?thread_ts=1574868757.276700&cid=CBN38N3MW

**Comment by prietyc123** [comment]
> @kadel It seems prow does not handle the authentication token for registry.redhat.io for image `openjdk-11-rhel8` .
> 
> I have created an upstream issue [openshift/release#6056](https://github.com/openshift/release/issues/6056) to get an idea of fixing it more efficiently.

Still looking into how to handle secret-id for registry.redhat.io for image `openjdk-11-rhel8` .

**Comment by prietyc123** [comment]
https://github.com/openshift/odo/pull/2402/commits/fc7e7f805b2f38f40b5bfa61cb54b92dab9daf16 

/hold

**Comment by amitkrout** [comment]
> @amitkrout can you please mount [openshift/release#6341](https://github.com/openshift/release/pull/6341) in the job file
> 
> [fc7e7f8](https://github.com/openshift/odo/commit/fc7e7f805b2f38f40b5bfa61cb54b92dab9daf16)
> 
> /hold

@prietyc123  Till now i did not get any response on my OpenShift release pr https://github.com/openshift/release/pull/6764. So at the moment let's remove the test spec which are failing for specific registry and will create a separate issue to track there.

@prietyc123 Open this pr for review after removing the failure case

**Comment by prietyc123** [comment]
> > @amitkrout can you please mount [openshift/release#6341](https://github.com/openshift/release/pull/6341) in the job file
> > [fc7e7f8](https://github.com/openshift/odo/commit/fc7e7f805b2f38f40b5bfa61cb54b92dab9daf16)
> > /hold
> 
> @prietyc123 Till now i did not get any response on my OpenShift release pr [openshift/release#6764](https://github.com/openshift/release/pull/6764). So at the moment let's remove the test spec which are failing for specific registry and will create a separate issue to track there.
> 
> @prietyc123 Open this pr for review after removing the failure case

Issue created #2537

File: Makefile [hunk 1]
```diff
@@ -228,6 +228,11 @@ test-e2e-java:
 test-e2e-source:
 	ginkgo $(GINKGO_FLAGS) -focus="odo source e2e tests" tests/e2escenarios/
 
+# Run image e2e tests
```
amitkrout: ```Run image e2e tests``` -> ```Run supported images e2e tests```

amitkrout: @prietyc123 ^

tests/e2escenarios/e2e_images_test.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+				_, err = os.Create(context + "/a.txt")
+				if err != nil {
+					panic(err)
+				}
+
+				helper.DeleteDir(context + "/abcd")
+
+				if srcType == "openjdk" {
+					helper.ReplaceString(filepath.Join(context+"/src/main/java/MessageProducer.java"), "Hello", "Hello odo")
```
amitkrout: @prietyc123 Use ```filepath.Join(context,"src","main","java","MessageProducer.java")``` instead and use the same format in other occurrences too.

tests/e2escenarios/e2e_images_test.go
```diff
@@ -0,0 +1,178 @@
...
+				} else {
+					helper.ReplaceString(filepath.Join(context+"/server.js"), "Hello", "Hello odo")
```
girishramnani: you should use `filepath.Join(context,"server.js", "Hello", "Hello odo")` here

girishramnani:
There are alot of places where platform dependent `/` have been added ( just in the code added as part of the PR ), can we please change that with `filepath.Join`. As to doing this change for all of the test suite, that can be done as part of different PR.

tests/helper/helper_oc.go
```diff
@@ -274,6 +275,13 @@ func (oc *OcRunner) checkForImageStream(name string, tag string) bool {
 	return false
 }
 
+// ImportSupportedImage import all the supported images
+func (oc *OcRunner) ImportSupportedImage(registry, image, cmpType, project string) {
```
amitkrout: @prietyc123 The name looks like it only works for support images but actually it does not. As per the args passed, It imports any image type into the namespace. So instead of making the function name specific you can use name something like ```ImportImageTypeIntoProject``` instead.

tests/helper/helper_oc.go
```diff
@@ -274,6 +275,13 @@ func (oc *OcRunner) checkForImageStream(name string, tag string) bool {
 	return false
 }
 
+// ImportSupportedImage import all the supported images
```
amitkrout: This is too short to explain what this command does. You can describe like
```
// ImportSupportedImage import any images types into the namespace. It takes four arguments such as
resgistry - Registry path of the image
image - Namespace/imagename (tag is optional)
cmpType - Component name (tag is optional)
project - Namespace
```

prietyc123:
> There are alot of places where platform dependent `/` have been added ( just in the code added as part of the PR ), can we please change that with `filepath.Join`. As to doing this change for all of the test suite, that can be done as part of different PR.

Yes, this kind of issues are already addressed in PR https://github.com/openshift/odo/pull/2524

tests/helper/helper_generic.go
```diff
@@ -81,3 +83,52 @@ func Unindented(jsonStr string) (string, error) {
 	}
 	return string(obj), err
 }
+
+// WatchNonRetCmdStdOut run odo watch and get the cmdSTDOUT output into buffer.
```
amitkrout: You can make the comment line more talkative i guess :)

tests/helper/helper_generic.go
```diff
@@ -81,3 +83,52 @@ func Unindented(jsonStr string) (string, error) {
...
+func WatchNonRetCmdStdOut(cmdStr string, timeout time.Duration, check func(output string) bool, startSimulationCh chan bool, startIndicatorFunc func(output string) bool) (bool, error) {
```
amitkrout: +1, looks good

Makefile
prietyc123: > `Run image e2e tests` -> `Run supported images e2e tests`

I would say if we add unsupported image e2e test in future then IMO `Run image e2e tests` is the perfect comment. WDYT?

girishramnani:
> > There are alot of places where platform dependent `/` have been added ( just in the code added as part of the PR ), can we please change that with `filepath.Join`. As to doing this change for all of the test suite, that can be done as part of different PR.
> 
> Yes, this kind of issues are already addressed in PR #2524

in the code added here I can still see `/` so can you please address those too

Makefile
amitkrout: Right, i could not realize in that prospect. ```Run image e2e tests``` would be more genetic. So we can keep this

amitkrout: My mistake, Unfortunately i interpreted your comment wrongly.

This is a target in the make file which can only execute the supported images. So it should be specific of what it does ```Run supported images e2e tests``` . As part of unsupported image target is concern, we can create a separate target anyway.

tests/helper/helper_oc.go
prietyc123: > @prietyc123 The name looks like it only works for support images but actually it does not. As per the args passed, It imports any image type into the namespace. So instead of making the function name specific you can use name something like `ImportImageTypeIntoProject` instead.

How about `ImportImageFromRegistry` ?

tests/helper/helper_generic.go
prietyc123: Done

tests/helper/helper_oc.go
amitkrout: > How about ImportImageFromRegistry ?

This makes more sense

girishramnani:
I found something interesting about ginkgo that it supports table driven tests https://onsi.github.io/ginkgo/#table-driven-tests . So could you consider using that for these @prietyc123 . 
cc @kadel

amitkrout:
> I found something interesting about ginkgo that it supports table driven tests https://onsi.github.io/ginkgo/#table-driven-tests . So could you consider using that for these @prietyc123 .
> cc @kadel

@girishramnani Nice reference to go with tabular formatted test. However me and @prietyc123 had a quick discussion around it and found we have also few more places in out test script where tabular test format can be applied. For example config file test.

IMO it would be great if we create a code refactor issue against the test script and fix that in the later sprint. cc_ @prietyc123 @kadel 

+1, Test script looks good to me.

girishramnani:
If the next step would be to use the tabular version then why not consider doing it as part of this PR?

kadel:
How big change it would be to convert this into a tabular version?
If it is just a few hours it makes sense to do it here. 

If it is a bigger refactor I would be ok with doing it in another PR together with bigger systematic change. As this PR is otherwise ready.

amitkrout:
> If the next step would be to use the tabular version then why not consider doing it as part of this PR?

@girishramnani No atm because this is not only the place where we are going to do the change. Tabula format may affects more than one test files. I would say let's do the proper cost analysis i mean the total number of test files which are being affected and the time consume to implement the tabular format.

> How big change it would be to convert this into a tabular version? If it is just a few hours it makes sense to do it here. If it is a bigger refactor I would be ok with doing it in another PR together with bigger systematic change. As this PR is otherwise ready.

@kadel Right now i am not sure, however me and @prietyc123 had a quick discuss on the tabular format and we found few places. May be tabular format affect more file than what we found in the quick discussion. So this need a proper analysis to find out files affected and cost.

I have created a separate issue to track the ginkgo tabular format implementation - https://github.com/openshift/odo/issues/2590

amitkrout:
> > If the next step would be to use the tabular version then why not consider doing it as part of this PR?
> 
> @girishramnani No atm because this is not only the place where we are going to do the change. Tabula format may affects more than one test files. I would say let's do the proper cost analysis i mean the total number of test files which are being affected and the time consume to implement the tabular format.
> 
> > How big change it would be to convert this into a tabular version? If it is just a few hours it makes sense to do it here. If it is a bigger refactor I would be ok with doing it in another PR together with bigger systematic change. As this PR is otherwise ready.
> 
> @kadel Right now i am not sure, however me and @prietyc123 had a quick discuss on the tabular format and we found few places. May be tabular format affect more file than what we found in the quick discussion. So this need a proper analysis to find out files affected and cost.
> 
> I have created a separate issue to track the ginkgo tabular format implementation - #2590

/lgtm

PR #2441, opened 2019-12-09
Unit test add odo-file-index.json to gitignore file

### LINKED ISSUE #2404: Unit Test for Adding odo file index json to gitignore file
/kind bug

https://github.com/openshift/odo/pull/2305#issuecomment-554246063

prietyc123:
/kind test

Add unit test for odo-file-index.json into gitignore file

Fixes #2404 

```make test```

### CODE REVIEW

**Comment by cdrage** [comment]
@prietyc123 Needs rebase. Is this ready for review?

**Comment by prietyc123** [comment]
> @prietyc123 Needs rebase. Is this ready for review?

Rebase is done. Yes its ready for review.

File: pkg/util/file_indexer_test.go [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+	// create a context directory on fake fs
+	contextDir, err := fs.TempDir(os.TempDir(), "context")
+	if err != nil {
+		t.Error(err)
+	}
+
+	gitignorePath := filepath.Join(contextDir, ".gitignore")
+
+	tests := []struct {
+		testName        string
```
dharmit: We use `name` in the unit tests I have come across.

pkg/util/file_indexer_test.go
```diff
@@ -0,0 +1,129 @@
...
+				t.Errorf("addOdoFileIndex unexpected error %v, wantErr %v", err, tt.wantErr)
+			}
+
+		})
+	}
+}
+
+func mockDirectoryInfo(create bool, contextDir string, fs filesystem.Filesystem) error {
+
+	if !create {
```
dharmit: I'm a little confused. This condition will be satisfied when `create` is set to `false`. So when it's `false`, we're creating the file/directory and when it's `true`, we're not doing anything of the sort. This is by design, right?

Just trying to affirm my understanding. Correct me if I'm wrong.

pkg/util/file_indexer_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// create a context directory on fake fs
+	contextDir, err := fs.TempDir(os.TempDir(), "context")
+	if err != nil {
+		t.Error(err)
+	}
+
+	gitignorePath := filepath.Join(contextDir, ".gitignore")
+
+	tests := []struct {
+		testName        string
```
dharmit: Same as earlier. We use `name`.

pkg/util/file_indexer_test.go
prietyc123: Correct @dharmit 

In the test case where .gitignore directory does not exist, create will be set to true as we don't want that directory to exist.

dharmit: It feels counter-intuitive to me. In the scenario that you explained, why not have `create` set to `false` since we already have what we need? Since you have used the word "directory" twice, aren't we talking about `.gitignore` file here? I'm feeling lost here. :disappointed:

pkg/util/file_indexer_test.go
prietyc123: I think we use both ```name``` as well ```testName```. As I came cross in https://github.com/openshift/odo/blob/master/pkg/util/util_test.go#L22 

We have used both in our unit test. Let me know your views on this.

dharmit: Haha. Fine by me. :smile:

prietyc123: Cool 👍

pkg/util/file_indexer_test.go
prietyc123: https://github.com/openshift/odo/pull/2441/files#r358119130

pkg/util/file_indexer_test.go
prietyc123: "directory" is the kind of context directory you can say where the component will get created. That directory we need to pass to the functions. 

Again in second unit test https://github.com/openshift/odo/pull/2441/files#diff-0292f6b604e75b79cef052987afa7fb5R72 we should be having .gitignore file as already checked from https://github.com/openshift/odo/pull/2441/files#diff-0292f6b604e75b79cef052987afa7fb5R11

dharmit: In the earlier comment, you wrote:
> In the test case where .gitignore directory does not exist

so I thought you're talking about a directory called `.gitinore`. Hence the question.

I'm still confused as to why we're creating a directory when `create` bool is `false`. :confused:

prietyc123: > In the earlier comment, you wrote:
> 
> > In the test case where .gitignore directory does not exist
> 
> so I thought you're talking about a directory called `.gitinore`. Hence the question.
> 
> I'm still confused as to why we're creating a directory when `create` bool is `false`. 😕

Because whatever false directory we are creating will get created in RAM and once that unit test is complete it will delete the directory and everything.

As for https://github.com/openshift/odo/pull/2441/files#diff-0292f6b604e75b79cef052987afa7fb5R72 this unit test .gitignore file should be there to add odo-file-index into .gitignore.

Hence ```create``` bool is set to false, so that we can have .gitignore file there in the directory.

prietyc123: > so I thought you're talking about a directory called `.gitinore`. Hence the question.
There are two different unit test, one for checking ```.gitignore``` exists or not. If not then create one. Second is to add ```odo-file-index.json``` into ```.gitignore```.

If we can link these two unit test then ```create``` bool will be set to ```true```. So is there any way we can do that as I am not aware of that. Please do suggest if there is any to return ```.gitignore``` path from the first unit test I mentioned above. If not then we have have to create ```.gitignore``` directory as we are using fake filesystem with ```afero```.

kadel:
@girishramnani is probably the right person to review changes in fakefs pkg
/cc @girishramnani

cdrage:
/lgtm

Looks good to me!

PR #2442, opened 2019-12-09
cleanup OWNERS file

kadel:
- Removed OWNERS file from `pkg/*` we are not that big to need to create separate areas like this (for now).
- Relying mostly on top-level OWNERS file will make sure that review requests are distributed more equally.
- Kept OWNERS for tests and scripts, as those contain highly specific things.

/cc @cdrage @girishramnani 

there is also related change in prow config https://github.com/openshift/release/pull/6198


mohammedzee1000:
Awesome stuff, I was wondering when we would truly adopt this :)

kadel:
> Awesome stuff, I was wondering when we would truly adopt this :)

We can start right now. There are still some issues left to solve around github permissions and labels, but that shouldn't prevent us from adopting this for reviews

PR #2446, opened 2019-12-11
Update pull request template

kadel:
- Remove /kinds that don't make sense for PR
- Possible kinds are as quotes to prevent labeling with everything when someone doesn't read this.


kadel:
small fix -> self approve ;-)
/approve

Issue #2436, opened 2019-12-06
Implement a kclient package for interacting with Kubernetes clusters


johnmcollier:
As we work to make odo more Kube-generic and have it function on non-OpenShift clusters, we'll need a library to interact with those clusters, much like how `occlient` functions today. This would be a standalone library from `occlient` that would interface with `client-go` to interact with resources on a Kubernetes cluster. 

- [ ] Determine which client-go library to use
    - Right now odo is pulling down OpenShift's fork of client-go for Kube 1.11 and is still needed for certain OpenShift packages it seems, so we can't get rid of it
    - Dual maintenance of multiple client-go's seems messy at best, so probably best to just keep using the OpenShift fork of client-go (it _should_ work on other Kubes for what we need)
- [ ] Implement a kclient library
    - See what / if any functions can be moved over from `occlient`
    - Best to keep things simple and modularized, rather than having a huge single file like there is for `occlient.go`
- [ ] Implement tests for kclient

johnmcollier:
Per the earlier odo-dev call this week, I can take a look at this.

kadel:
I would add to this is that the main point why we want to implement kclient is to use it as we start implementing devfile support.

I wouldn't start with moving functions from `occlient`. Devfile flow will be quite different from what we are doing right now with s2i. I would rather start with a clean slate, make it as you say, simpler and modularized and add functions only as we need them for devfile support.

rajivnathan:
We can start real small with kclient, maybe just these 2 functions:
WaitAndGetPod
ExecCMDInContainer

Once we have the file created and the plumbing in place we can add to it as needed.

johnmcollier:
@rajivnathan @kadel That sounds like a good approach. I'll get to work on it.

Here's what I'm thinking in terms of structure. Separate out kclient into subpackages for different resources (i.e. interacting with pods, services, etc):
```
pkg/kclient
├── pods
|   ├── pods.go
|   └── pods_test.go
├── services
|   ├── services.go
|   └── services_test.go
├── kclient.go
```

rajivnathan:
@girishramnani @kadel Should this one have the devfile label as well? Do we need a kubernetes label as well or is just devfile label good enough? Just want to make sure I track the right issues.

PR #2458, opened 2019-12-13
Implement a stub kclient package for odo

johnmcollier:

**What type of PR is this?**

/kind feature

This PR implements a stub kclient package for use in odo (particularly for use with the upcoming Devfile integration), allowing odo to interact with non-OpenShift clusters (as occlient requires an OpenShift cluster). For now, the forked `k8s.io/client-go` is used, as it will work fine on non-OpenShift clusters, but we should consider pulling down the upstream version when possible.

The following files are added to the new `kclient` package:
- kclient.go - For creating and initializing a Kube client with client-go
- pods.go - For interacting with pods. Has just two functions in it right now: `WaitAndGetPod` and `ExecCMDInContainer`. 
- pods_test.go and fakeclient.go: For running tests against the `kclient` package

Fixes https://github.com/openshift/odo/issues/2436

1) `cd pkg/kclient`
2) `go test`
```
johns-mbp-3:odo johncollier$ cd pkg/kclient/
johns-mbp-3:kclient johncollier$ go test
 ✓  Waiting for component to start [225986ns]
 ✗  Waiting for component to start [106133ns]
 ✗  Waiting for component to start [111110ns]
PASS
ok  	github.com/openshift/odo/pkg/kclient	0.118s
```


pkg/kclient/pods.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	// waitForPodTimeOut controls how long we should wait for a pod before giving up
+	waitForPodTimeOut = 240 * time.Second
+)
+
+// WaitAndGetPod block and waits until pod matching selector is in in Running state
```
rajivnathan: Looks like the comment for this function is outdated because it doesn't just match `Running` phase since desiredPhase was added? Also, `in in`.

pkg/kclient/pods.go
```diff
@@ -0,0 +1,118 @@
...
+// desiredPhase cannot be PodFailed or PodUnknown
+func (c *Client) WaitAndGetPod(watchOptions metav1.ListOptions, desiredPhase corev1.PodPhase, waitMessage string) (*corev1.Pod, error) {
```
rajivnathan: Should we accept multiple desired phases? eg. Running or Succeeded?

pkg/kclient/kclient.go
```diff
@@ -0,0 +1,53 @@
...
+// Client is a collection of fields used for client configuration and interaction
+type Client struct {
+	KubeClient       kubernetes.Interface
+	KubeConfig       clientcmd.ClientConfig
+	KubeClientConfig *rest.Config
+	Namespace        string
+}
+
+// New creates a new client
+func New(skipConnectionCheck bool) (*Client, error) {
```
kadel: Do we need `skipConnectionCheck`? It is not used anywhere in the function.

pkg/kclient/pods_test.go
```diff
@@ -0,0 +1,85 @@
...
+			timeout := int64(10)
+			watchOptions := metav1.ListOptions{
+				LabelSelector:  podSelector,
+				TimeoutSeconds: &timeout,
+			}
+			pod, err := fkclient.WaitAndGetPod(watchOptions, corev1.PodRunning, "Waiting for component to start")
+
+			if !tt.wantErr == (err != nil) {
+				t.Errorf(" client.WaitAndGetPod(string) unexpected error %v, wantErr %v", err, tt.wantErr)
+				return
```
dharmit: We can do `t.Fatal` if we want the execution to stop at this point. I mean instead of doing `t.Errorf` and `return`, we could use just `t.Fatal`.

pkg/kclient/kclient.go
```diff
@@ -0,0 +1,53 @@
...
+	var client Client
+
+	// initialize client-go clients
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	client.KubeConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+
+	config, err := client.KubeConfig.ClientConfig()
+	if err != nil {
+		return nil, errors.New(err.Error() + errorMsg)
```
girishramnani: we have an `errors.Wrap` function in utils which can be used here

pkg/kclient/pods.go
```diff
@@ -0,0 +1,118 @@
...
+	podChannel := make(chan *corev1.Pod)
+	watchErrorChannel := make(chan error)
+
+	go func() {
+	loop:
+		for {
+			val, ok := <-w.ResultChan()
+			if !ok {
+				watchErrorChannel <- errors.New("watch channel was closed")
+				break loop
```
girishramnani: can we `return` here instead of breaking?

pkg/kclient/pods.go
```diff
@@ -0,0 +1,118 @@
...
+				case corev1.PodFailed, corev1.PodUnknown:
+					watchErrorChannel <- errors.Errorf("pod %s status %s", e.Name, e.Status.Phase)
+					break loop
+				}
+			} else {
+				watchErrorChannel <- errors.New("unable to convert event object to Pod")
+				break loop
+			}
+		}
+		close(podChannel)
```
girishramnani: `defer close` this?

pkg/kclient/pods.go
johnmcollier: Updated the comment

pkg/kclient/kclient.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// errorMsg is the message for user when invalid configuration error occurs
+const errorMsg = `
```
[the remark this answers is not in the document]
johnmcollier: That's definitely an interesting approach, and at that article definitely has some merits. But I'm wondering if we want to implement it as part of our `errors` package we have or maybe a subpackage in kclient if we want to isolate things (and to prevent having to update a ton of existing odo code)

My concern is around ensuring we can use the existing functions in our `errors` package, and defining a custom error type might cause problems with currently.

pkg/kclient/kclient.go
```diff
@@ -0,0 +1,53 @@
...
+	// initialize client-go clients
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	client.KubeConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+
+	config, err := client.KubeConfig.ClientConfig()
+	if err != nil {
+		return nil, errors.New(err.Error() + errorMsg)
+	}
+	client.KubeClientConfig = config
```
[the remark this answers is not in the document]
johnmcollier: No specific pattern, just a habit of how I learned Go 🙂.

Given that we're checking the `err` value (and erroring out where applicable), it probably makes sense to assign client.KubeClientConfig (and others) directly.

pkg/kclient/kclient.go
johnmcollier: Thanks, updated!

pkg/kclient/pods.go
johnmcollier: Sure. I've updated the goroutine to use `returns` instead of `breaks` in the loop.

pkg/kclient/pods.go
johnmcollier: Updated to `defer close` both channels.

pkg/kclient/pods_test.go
johnmcollier: Ah, good to know about `t.Fatal`, I didn't realize the testing package had that too. I've updated the test to use it.

pkg/kclient/kclient.go
johnmcollier: Ah, good catch @kadel, that can be removed as it's not being used right now (and the kclient in our kdo-poc branch never used it)

pkg/kclient/pods.go
johnmcollier: Since the corresponding function in `occlient` only accepts one desiredPhase, I'd prefer to leave it as-is for now.

pkg/kclient/pods.go
johnmcollier: Updated the comment to reflect the current status of the function.

pkg/kclient/kclient.go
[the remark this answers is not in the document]
dharmit: Thanks for that explanation. :smile:

johnmcollier:
@dharmit @kadel @girishramnani I've updated the PR to address your review comments, please review at your convenience, thanks!

kadel:
This is a good start. 👍 
/lgtm

It maybe takes a little bit too much a similar approach o our current `occlient` package. Which I think doesn't have a good structure. I wouldn't be afraid to do big changes. But we can do that as we start using `kclient` in new code.

PR #2484, opened 2019-12-19
Update changelog script

cdrage:
**What type of PR is this?**

> /kind bug

Fix up the changelog script by commenting out the summary section since
we do not (yet) summarize releases in issues.

Need to comment this out so the next release goes out smoothly.

N/A

N/A


cdrage:
Force merging this in since we need this in before the release :+1:

Issue #2208, opened 2019-09-30
odo catalog list components -o json is missing the supported flag


jeffmaury:
[kind/bug]

## What versions of software are you using?
- Operating System: Win10 Pro
- Output of `odo version`:

```
odo v1.0.0-beta6 (0cdcc908)      
                                 
Server: https://172.18.84.8:8443 
Kubernetes: v1.11.0+d4cacc0
```      

## How did you run odo exactly?

`odo catalog list components -o json`

## Actual behavior

JSON output list stacks but the supported flag is missing

## Expected behavior

Add a new property to each stack that exposes the supported nature

girishramnani:
@jeffmaury I can introduce a new list called `supportedTags` which will list all the supported tags ( subset of `nonHiddenTags` )

PR #2486, opened 2019-12-24
add supportedTags list in the odo catalog list component json output

girishramnani:
**What type of PR is this?**

/kind feature

adds `supportedTags` list in the `odo catalog list components`'s machine readable output.
The `supportedTags` is not removed from the out if it's empty, this is done to make the output consistent for all `ComponentType`.

Fixes https://github.com/openshift/odo/issues/2208

```
odo catalog list components -o json
```
should have the `supportedTags` and it should be consistent with `odo catalog list components`

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -225,6 +225,7 @@ func getDefaultBuilderImages(client *occlient.Client) ([]ComponentType, error) {
 // SliceSupportedTags splits the tags in to fully supported and unsupported tags
 func SliceSupportedTags(component ComponentType) ([]string, []string) {
 
+	// this makes sure that json marshal shows these lists as [] instead of null
 	supTag, unSupTag := []string{}, []string{}
 	tagMap := createImageTagMap(component.Spec.ImageStreamRef.Spec.Tags)
 	for _, tag := range component.Spec.NonHiddenTags {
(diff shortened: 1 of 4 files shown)


prietyc123:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/2486/pull-ci-openshift-odo-master-v4.1-integration-e2e-benchmark/245#1:build-log.txt%3A498

girishramnani:
What would I unit test :thinking: ?

girishramnani:
The integration test already checks for the presence of `supportedTags`. We cannot check for more in that integration test because supported components change from cluster to cluster. 
The `SliceSupportedTags` has been tested, the machine readable output has been tested as well. So not entirely sure when more we can unit test specific to this change?

dharmit:
> The integration test already checks for the presence of `supportedTags`. We cannot check for more in that integration test because supported components change from cluster to cluster.
> The `SliceSupportedTags` has been tested, the machine readable output has been tested as well. So not entirely sure what more we can unit test specific to this change?

Neither am I sure what exactly the test could look like. But I guess it would be something about checking the presence of `supportedTags` in the json output, right? My intention in asking for a unit test was that if we can add a reliable test there, we can remove integration test since unit tests are faster.

If you say that it's not possible/required to add a unit test here, fine by me. :smile:

girishramnani:
the check for presence for `supportedTags` is already present 
https://github.com/openshift/odo/pull/2486/files#diff-f0f2e54d32d2c1ab57020c3a3dd9a3fbR238

dharmit:
Ack. Since you're doing this in an existing spec, my nitpicking doesn't make sense here. :smile: 

/lgtm

girishramnani:
self approving as other approvers are on leave 
/approve

PR #2491, opened 2019-12-31
Delete services belonging to an application during 'odo app delete'

### LINKED ISSUE #2113: `odo app delete` removes application components but not services
[kind/bug]
- Operating System: Windows
- Output of `odo version`: 
```
> odo version
odo v1.0.0-beta5 (4d1d5071)
```

```
> odo.exe create nodejs:latest nodejs1 --git https://github.com/dgolovin/nodejs-ex --ref HEAD --app app --project myproject

> odo.exe push

> odo.exe service create postgresql-persistent --plan default psql --app app --project myproject -w

> odo list
APP     NAME        TYPE       SOURCE                                    STATE
app     nodejs1     nodejs     https://github.com/dgolovin/nodejs-ex     Pushed

> odo service list
NAME     TYPE                      STATUS
psql     postgresql-persistent     ProvisionedAndBound

> odo app delete app -f
This application has following components that will be deleted
component named nodejs1
This application has following service that will be deleted
service named psql of type postgresql-persistent
Deleted application: app from project: myproject

> odo list
 X  There are no components deployed.

> odo service list
NAME     TYPE                      STATUS
psql     postgresql-persistent     ProvisionedAndBound
```
## Actual behavior
Services have not been deprovisioned

## Expected behavior
Services have been deprovisioned

N/A

dharmit:
**What type of PR is this?**

/kind bug

This PR deletes the service (ServiceInstance) created from Service Catalog when user tries to delete an application (`odo app delete <app-name>`).

When there are no components in the application but only service(s), it also prints the list of services that would be deleted when doing `odo app delete <app-name>`. This list of services was not getting printed unless there was a component in `<app-name>` marked for deletion.

Fixes #2113 

1. Create a service from Service Catalog `odo service create`
2. Make sure the service shows up in `odo service list --app app`
3. Now delete the app `odo app delete app -f`
4. Make sure that no services exist. (Check on OpenShift web console)

### DIFF
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -7,6 +7,7 @@ import (
 	applabels "github.com/openshift/odo/pkg/application/labels"
 	"github.com/openshift/odo/pkg/component"
 	"github.com/openshift/odo/pkg/occlient"
+	"github.com/openshift/odo/pkg/service"
 	"github.com/openshift/odo/pkg/util"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -71,8 +72,22 @@ func Delete(client *occlient.Client, name string) error {
 
 	labels := applabels.GetLabels(name, false)
 
+	// first delete the services (ServiceInstance in OpenShift terminology)
+	// belonging to the app
+	svcList, err := service.List(client, name)
+	if err != nil {
+		return errors.Wrapf(err, "unable to delete the application %s due to failure in listing service(s) in the application", name)
+	}
+
+	for _, svc := range svcList.Items {
+		err = service.DeleteServiceAndUnlinkComponents(client, svc.Name, name)
+		if err != nil {
+			return errors.Wrapf(err, "unable to delete the application %s due to failure in deleting service(s) in the application", name)
+		}
+	}
+
 	// delete application from cluster
-	err := client.Delete(labels)
+	err = client.Delete(labels)
 	if err != nil {
 		return errors.Wrapf(err, "unable to delete application %s", name)
 	}
--- a/pkg/application/application.go
+++ b/pkg/application/application.go
@@ -76,16 +76,16 @@ func Delete(client *occlient.Client, name string) error {
 	// belonging to the app
 	svcList, err := service.List(client, name)
 	if err != nil {
-		return errors.Wrapf(err, "unable to delete the application %s due to failure in listing service(s) in the application", name)
-	}
-
-	for _, svc := range svcList.Items {
-		err = service.DeleteServiceAndUnlinkComponents(client, svc.Name, name)
-		if err != nil {
-			return errors.Wrapf(err, "unable to delete the application %s due to failure in deleting service(s) in the application", name)
+		// error is returned when there's no Service Catalog enabled in the service
+		glog.V(4).Infof("Service catalog is not enabled in the cluster, skipping service deletion")
+	} else {
+		for _, svc := range svcList.Items {
+			err = service.DeleteServiceAndUnlinkComponents(client, svc.Name, name)
+			if err != nil {
+				return errors.Wrapf(err, "unable to delete the application %s due to failure in deleting service(s) in the application", name)
+			}
 		}
 	}
-
 	// delete application from cluster
 	err = client.Delete(labels)
 	if err != nil {
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -2398,8 +2398,6 @@ func (c *Client) CreateServiceBinding(bindingName string, namespace string, labe
 				Labels:    labels,
 			},
 			Spec: scv1beta1.ServiceBindingSpec{
-				//ExternalID: UUID,
-
 				InstanceRef: scv1beta1.LocalObjectReference{
 					Name: bindingName,
 				},
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -2379,7 +2379,7 @@ func (c *Client) CreateServiceInstance(serviceName string, serviceType string, s
 	}
 
 	// Create the secret containing the parameters of the plan selected.
-	err = c.CreateServiceBinding(serviceName, c.Namespace)
+	err = c.CreateServiceBinding(serviceName, c.Namespace, labels)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create the secret %s for the service instance", serviceName)
 	}
@@ -2389,12 +2389,13 @@ func (c *Client) CreateServiceInstance(serviceName string, serviceType string, s
 
 // CreateServiceBinding creates a ServiceBinding (essentially a secret) within the namespace of the
 // service instance created using the service's parameters.
-func (c *Client) CreateServiceBinding(bindingName string, namespace string) error {
+func (c *Client) CreateServiceBinding(bindingName string, namespace string, labels map[string]string) error {
 	_, err := c.serviceCatalogClient.ServiceBindings(namespace).Create(
 		&scv1beta1.ServiceBinding{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      bindingName,
 				Namespace: namespace,
+				Labels:    labels,
 			},
 			Spec: scv1beta1.ServiceBindingSpec{
 				//ExternalID: UUID,
--- a/pkg/occlient/occlient_test.go
+++ b/pkg/occlient/occlient_test.go
@@ -1487,12 +1487,14 @@ func TestCreateServiceBinding(t *testing.T) {
 		name        string
 		bindingNS   string
 		bindingName string
+		labels      map[string]string
 		wantErr     bool
 	}{
 		{
 			name:        "Case: Valid request for creating a secret",
 			bindingNS:   "",
 			bindingName: "foo",
+			labels:      map[string]string{"app": "app"},
 			wantErr:     false,
 		},
 	}
@@ -1500,7 +1502,7 @@ func TestCreateServiceBinding(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			fakeClient, fakeClientSet := FakeNew()
 
-			err := fakeClient.CreateServiceBinding(tt.bindingName, tt.bindingNS)
+			err := fakeClient.CreateServiceBinding(tt.bindingName, tt.bindingNS, tt.labels)
 
 			if err == nil && !tt.wantErr {
 				if len(fakeClientSet.ServiceCatalogClientSet.Actions()) != 1 {
--- a/tests/integration/servicecatalog/cmd_service_test.go
+++ b/tests/integration/servicecatalog/cmd_service_test.go
@@ -454,8 +454,8 @@ var _ = Describe("odo service command tests", func() {
 
 			ocArgs = []string{"get", "serviceinstances"}
 			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
-				return strings.Contains(output, "Deprovisioning")
-			})
+				return strings.Contains(output, "No resources found")
+			}, true)
 		})
 	})
 })
--- a/tests/integration/servicecatalog/cmd_service_test.go
+++ b/tests/integration/servicecatalog/cmd_service_test.go
@@ -1,7 +1,6 @@
 package integration
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -452,10 +451,10 @@ var _ = Describe("odo service command tests", func() {
 			})
 
 			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
-			odoArgs := []string{"service", "list", "--app", app}
-			helper.WaitForCmdOut("odo", odoArgs, 1, false, func(output string) bool {
-				fmt.Println(output)
-				return strings.Contains(output, "There are no services deployed for this application")
+
+			ocArgs = []string{"get", "serviceinstances"}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "Deprovisioning")
 			})
 		})
 	})
--- a/tests/integration/servicecatalog/cmd_service_test.go
+++ b/tests/integration/servicecatalog/cmd_service_test.go
@@ -1,6 +1,7 @@
 package integration
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -433,6 +434,29 @@ var _ = Describe("odo service command tests", func() {
 
 			helper.CmdShouldPass("odo", "catalog", "describe", "service", "dh-es-apb")
 			helper.CmdShouldPass("odo", "catalog", "describe", "service", "dh-import-vm-apb")
+		})
+	})
+
+	Context("When the application is deleted", func() {
+		JustBeforeEach(func() {
+			preSetup()
+			app = helper.RandString(6)
+		})
+		It("should delete the service(s) in the application as well", func() {
+			helper.CmdShouldPass("odo", "service", "create", "--app", app, "-w", "dh-postgresql-apb", "--project", project, "--plan", "dev",
+				"-p", "postgresql_user=luke", "-p", "postgresql_password=secret",
+				"-p", "postgresql_database=my_data", "-p", "postgresql_version=9.6")
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+			odoArgs := []string{"service", "list", "--app", app}
+			helper.WaitForCmdOut("odo", odoArgs, 1, false, func(output string) bool {
+				fmt.Println(output)
+				return strings.Contains(output, "There are no services deployed for this application")
+			})
 		})
 	})
 })
--- a/pkg/application/labels/labels.go
+++ b/pkg/application/labels/labels.go
@@ -22,7 +22,7 @@ const OdoVersion = "app.kubernetes.io/managed-by-version"
 // GetLabels return labels that identifies given application
 // additional labels are used only when creating object
 // if you are creating something use additional=true
-// if you need labels to filter component than use additional=false
+// if you need labels to filter component then use additional=false
 func GetLabels(application string, additional bool) map[string]string {
 	labels := map[string]string{
 		ApplicationLabel: application,
--- a/pkg/component/labels/labels.go
+++ b/pkg/component/labels/labels.go
@@ -16,7 +16,7 @@ const ComponentTypeVersion = "app.openshift.io/runtime-version"
 // GetLabels return labels that should be applied to every object for given component in active application
 // additional labels are used only for creating object
 // if you are creating something use additional=true
-// if you need labels to filter component that use additional=false
+// if you need labels to filter component then use additional=false
 func GetLabels(componentName string, applicationName string, additional bool) map[string]string {
 	labels := applabels.GetLabels(applicationName, additional)
 	labels[ComponentLabel] = componentName
--- a/pkg/odo/cli/application/application.go
+++ b/pkg/odo/cli/application/application.go
@@ -92,19 +92,19 @@ func printDeleteAppInfo(client *occlient.Client, appName string, projectName str
 				}
 			}
 		}
-		// List services that will be removed
-		serviceList, err := service.List(client, appName)
-		if err != nil {
-			log.Info("No services / could not get services")
-			glog.V(4).Info(err.Error())
-		}
-		if len(serviceList.Items) != 0 {
-			log.Info("This application has following service that will be deleted")
-			for _, ser := range serviceList.Items {
-				log.Info("service named", ser.ObjectMeta.Name, "of type", ser.Spec.Type)
-			}
-		}
-
 	}
+	// List services that will be removed
+	serviceList, err := service.List(client, appName)
+	if err != nil {
+		log.Info("No services / could not get services")
+		glog.V(4).Info(err.Error())
+	}
+	if len(serviceList.Items) != 0 {
+		log.Info("This application has following service(s) that will be deleted")
+		for _, ser := range serviceList.Items {
+			log.Info("service named", ser.ObjectMeta.Name, "of type", ser.Spec.Type)
+		}
+	}
+
 	return nil
 }
--- a/tests/helper/helper_generic.go
+++ b/tests/helper/helper_generic.go
@@ -11,6 +11,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
 )
 
 func init() {
@@ -36,7 +37,7 @@ func RandString(n int) string {
 // check (function with output check logic)
 // It times out if the command doesn't fetch the
 // expected output  within the timeout period.
-func WaitForCmdOut(program string, args []string, timeout int, errOnFail bool, check func(output string) bool) bool {
+func WaitForCmdOut(program string, args []string, timeout int, errOnFail bool, check func(output string) bool, includeStdErr ...bool) bool {
 	pingTimeout := time.After(time.Duration(timeout) * time.Minute)
 	// this is a test package so time.Tick() is acceptable
 	// nolint
@@ -47,8 +48,16 @@ func WaitForCmdOut(program string, args []string, timeout int, errOnFail bool, c
 			Fail(fmt.Sprintf("Timeout out after %v minutes", timeout))
 
 		case <-tick:
-			stdOut := CmdShouldPass(program, args...)
-			if check(strings.TrimSpace(string(stdOut))) {
+			session := CmdRunner(program, args...)
+			Eventually(session).Should(gexec.Exit(0), runningCmd(session.Command))
+			session.Wait()
+			output := string(session.Out.Contents())
+
+			if len(includeStdErr) > 0 && includeStdErr[0] {
+				output += "\n"
+				output += string(session.Err.Contents())
+			}
+			if check(strings.TrimSpace(string(output))) {
 				return true
 			}
 		}


pkg/occlient/occlient.go
```diff
@@ -2212,12 +2212,13 @@ func (c *Client) CreateServiceInstance(serviceName string, serviceType string, s
...
+func (c *Client) CreateServiceBinding(bindingName string, namespace string, labels map[string]string) error {
 	_, err := c.serviceCatalogClient.ServiceBindings(namespace).Create(
 		&scv1beta1.ServiceBinding{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      bindingName,
 				Namespace: namespace,
+				Labels:    labels,
 			},
 			Spec: scv1beta1.ServiceBindingSpec{
 				//ExternalID: UUID,
```
prietyc123: IMO this particular line is having no significance over here, So it can be removed.

dharmit: It's not having a major significance. I've added it so that a service binding has a label for the app. It will help us have same labels for a service instance and service bindings. Hence, I'd prefer to keep it.

girishramnani: she is talking about this commented like `//ExternalID: UUID,`

dharmit:
```
[odo]  ✗  serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list resource "serviceinstances" in API group "servicecatalog.k8s.io" in the namespace "lcnolchtql"
```

This error might show up when there are permission issue, but it also shows up when Service Catalog is not enabled in the cluster. I tried to delete the app on a minishift cluster that didn't have Service Catalog setup:

```sh
# Before enabling the Service Catalog
$ odo app delete app -f
This application has following components that will be deleted
component named node
No services / could not get services
 ✗  serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list serviceinstances.servicecatalog.k8s.io in the namespace "myproject": no RBAC policy matched

$ odo service list --project myproject
 ✗  Service catalog is not enabled within your cluster: unable to list services: unable to list ServiceInstances: serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list serviceinstances.servicecatalog.k8s.io in the namespace "myproject": no RBAC policy matched

# After enabling the Service Catalog
$ odo app delete app -f
This application has following components that will be deleted
component named node
Deleted application: app from project: myproject
```

I remember having a conversation where the reason for not enabling Service Catalog in the tests was to keep the test cluster less beefy. Let me know what you recommend in this situation. @kadel @amitkrout

girishramnani:
> ```
> [odo]  ✗  serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list resource "serviceinstances" in API group "servicecatalog.k8s.io" in the namespace "lcnolchtql"
> ```
> 
> This error might show up when there are permission issue, but it also shows up when Service Catalog is not enabled in the cluster. I tried to delete the app on a minishift cluster that didn't have Service Catalog setup:
> 
> ```shell
> # Before enabling the Service Catalog
> $ odo app delete app -f
> This application has following components that will be deleted
> component named node
> No services / could not get services
>  ✗  serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list serviceinstances.servicecatalog.k8s.io in the namespace "myproject": no RBAC policy matched
> 
> $ odo service list --project myproject
>  ✗  Service catalog is not enabled within your cluster: unable to list services: unable to list ServiceInstances: serviceinstances.servicecatalog.k8s.io is forbidden: User "developer" cannot list serviceinstances.servicecatalog.k8s.io in the namespace "myproject": no RBAC policy matched
> 
> # After enabling the Service Catalog
> $ odo app delete app -f
> This application has following components that will be deleted
> component named node
> Deleted application: app from project: myproject
> ```
> 
> I remember having a conversation where the reason for not enabling Service Catalog in the tests was to keep the test cluster less beefy. Let me know what you recommend in this situation. @kadel @amitkrout

we shouldn't fail the deletion of app if the service catalog doesn't exist.

pkg/occlient/occlient.go
dharmit: Done

pkg/application/application.go
```diff
@@ -71,8 +72,22 @@ func Delete(client *occlient.Client, name string) error {
 
 	labels := applabels.GetLabels(name, false)
 
+	// first delete the services (ServiceInstance in OpenShift terminology)
+	// belonging to the app
+	svcList, err := service.List(client, name)
+	if err != nil {
+		// error is returned when there's no Service Catalog enabled in the service
+		glog.V(4).Infof("Service catalog is not enabled in the cluster, skipping service deletion")
```
dharmit: I'm assuming this based on https://github.com/openshift/odo/blob/974de1fd1285ce8680e7597ba09cf17345e0504d/pkg/odo/cli/service/list.go#L59-L62 in which `List` function is one among more functions which are generalized by this log message

dharmit:
> we shouldn't fail the deletion of app if the service catalog doesn't exist.

Of course we shouldn't. And that's not what were doing. The problem was only in the CI. If you tried in your dev environment with an OpenShift cluster that doesn't have Service Catalog, it wouldn't have failed. It would have just exited silently and successfully!

Anyway, I get your point and I have made the changes by referring to other parts of the code base. Most CI checks look happy now. :man_dancing:

pkg/occlient/occlient_test.go
```diff
@@ -1398,20 +1398,22 @@ func TestCreateServiceBinding(t *testing.T) {
 		name        string
 		bindingNS   string
 		bindingName string
+		labels      map[string]string
 		wantErr     bool
 	}{
 		{
 			name:        "Case: Valid request for creating a secret",
```
prietyc123: I think we should have some more test case for this unit test like we should have the test case for ```invalid request ``` also. What will happen if we pass ```empty labels``` or ```different binding name``` other than ```foo```. I think we should consider these test case also. It's just a thought, WDYT?

dharmit: I looked at it as you suggested. But we're getting these values from the function that calls `CreateServiceBinding`. About the invalid request point, it's valid but `CreateServiceBinding` uses function in the OpenShift API to create the Service Binding. So adding a test for it would be futile (assuming the tests already exist in OpenShift source code.)

pkg/occlient/occlient.go
```diff
@@ -2212,15 +2212,15 @@ func (c *Client) CreateServiceInstance(serviceName string, serviceType string, s
 
 // CreateServiceBinding creates a ServiceBinding (essentially a secret) within the namespace of the
 // service instance created using the service's parameters.
-func (c *Client) CreateServiceBinding(bindingName string, namespace string) error {
+func (c *Client) CreateServiceBinding(bindingName string, namespace string, labels map[string]string) error {
```
mik-dass: Shouldn't this throw an error when the bindingName or namespace are empty strings?

dharmit: I think so. I think it should throw an error for empty value of `namespace` as well.

mik-dass: Yeah I think so. It shouldn't be a empty value right?

pkg/occlient/occlient_test.go
prietyc123: IMO we should also add some invalid test case to be safe. Depending on Openshift source code tests won't be good idea i guess. :-)

pkg/occlient/occlient.go
prietyc123: > I think so. I think it should throw an error for empty value of `namespace` as well.

He is talking about ```bindingName```. I am also thinking about ```labels```, shouldn't this will throw error for empty labels?

dharmit: So, just for the sake of it, I changed the values of `bindingName` and `namespace` to `""` before the call to create a ServiceBinding. And it failed with below errors:
```sh
# when namespace is ""
odo service create dh-postgresql-apb psql-1 --plan dev -p postgresql_database=mydb -p postgresql_password=password -p postgresql_user=luke -p postgresql_version=10
Deploying service psql-1 of type: dh-postgresql-apb
 ✗  Deploying service [369ms]
 ✗  an empty namespace may not be set during creation

# when bindingNs is ""
$ odo service create dh-postgresql-apb psql-1 --plan dev -p postgresql_database=mydb -p postgresql_password=password -p postgresql_user=luke -p postgresql_version=10
Deploying service psql-1 of type: dh-postgresql-apb
 ✗  Deploying service [564ms]
 ✗  ServiceBinding.servicecatalog.k8s.io "" is invalid: [metadata.name: Required value: name or generateName is required, spec.instanceRef.name: Invalid value: "": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '
-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.secretName: Invalid value: "": a DNS-1123 subdomai
n must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]
```

What's surprising in both cases is that in spite of the error, service gets deployed and provisioned:
```sh
$ odo service list --app app
NAME       TYPE                  PLAN     STATUS
psql-1     dh-postgresql-apb     dev      ProvisionedSuccessfully
```

But, again in both the cases, the service deletion fails since it can't find the corresponding service binding:
```sh
$ odo service delete psql-1 --app=app -f
 ✗  unable to delete service psql-1:
unable to delete serviceBinding: servicebindings.servicecatalog.k8s.io "psql-1" not found
```

So one needs to do `oc delete` to delete the ServiceInstance.

dharmit: > I am also thinking about labels, shouldn't this will throw error for empty labels?

Empty labels are fine because they are not mandatory to create an API object. I tried it out btw.

dharmit:
Breaks the test at:
```
--- FAIL: TestCreateServiceInstance/Create_service_instance (0.00s)
        occlient_test.go:2562:  client.CreateServiceInstance(serviceName,serviceType, labels) unexpected error unable to create the secret jenkins for the service instance: Found nil value for bindingName or namesapce; need values for both, wantErr false
```

girishramnani:
> namesapce

typo :slightly_smiling_face: @dharmit

pkg/occlient/occlient.go
mik-dass: > So, just for the sake of it, I changed the values of bindingName and namespace to "" before the call to create a ServiceBinding. And it failed with below errors:

So I guess bindingName and namespace shouldn't be empty and thus we should have a check for that in this function to make sure that we don't face the described error.

dharmit:
> > namesapce
> 
> typo slightly_smiling_face @dharmit

:joy: 

I'm anyway considering taking back the check for `bindingName` and `namesapce`. Because it's causing a different test to fail. Also, in spite of empty value for `bindingName` in the unit tests, it just worked earlier. Plus this PR is not touching anything related to it.

pkg/occlient/occlient_test.go
girishramnani: > I think we should have some more test case for this unit test like we should have the test case for `invalid request ` also. What will happen if we pass `empty labels` or `different binding name` other than `foo`. I think we should consider these test case also. It's just a thought, WDYT?

The function we are testing here just calls the openshift api server with the arguments passed to it. The cases that you have mentioned would have been already taken care of by openshift.

It's not about scope but actually who is responsible of the test. once the occlient handed over the data all the validation and functionality come under the scope of openshift. and so does testing those pieces.

pkg/occlient/occlient.go
dharmit: Opened #2493 to handle this. Is that OK @mik-dass? Those lines are unrelated to this PR and I'd like to keep it that way.

mik-dass: ok :+1:

dharmit:
```
```

dharmit:
/hold do-not-merge

Adding integration test for this. I was under an impression that we don't have service catalog tests in the CI.

pkg/occlient/occlient.go
kadel: I just added a [comment](https://github.com/openshift/odo/issues/2493#issuecomment-572023557) on #2493  related to this.

dharmit:
Travis is failing due to:
```sh
Created dir: /tmp/863945455
Creating a new project: ltwieroyye
Running odo with args [odo project create ltwieroyye -w -v4]
[odo]  •  Waiting for project to come up  ...
[odo] I0109 14:34:54.285469   31887 occlient.go:537] Project ltwieroyye now exists
[odo]  ✓  Waiting for project to come up [844ms]
[odo]  ✓  Project 'ltwieroyye' is ready for use
[odo]  ✓  New project created and now using project: ltwieroyye
[odo] I0109 14:34:54.293428   31887 odo.go:77] Could not get the latest release information in time. Never mind, exiting gracefully :)
Created dir: /tmp/526972802
Created dir: /tmp/454825721
Current working dir: /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog
Running odo with args [odo service create --app oahjjh -w dh-postgresql-apb --project ltwieroyye --plan dev -p postgresql_user=luke -p postgresql_password=secret -p postgresql_database=my_data -p postgresql_version=9.6]
[odo] Deploying service dh-postgresql-apb of type: dh-postgresql-apb
[odo]  •  Deploying service  ...
[odo]  ✓  Deploying service [34ms]
[odo]  •  Waiting for service to come up  ...
[odo]  ✓  Waiting for service to come up [1m]
[odo]  ✓  Service 'dh-postgresql-apb' is ready for use
[odo] Optionally, link dh-postgresql-apb to your component by running: 'odo link <component-name>'
Running oc with args [oc get serviceinstance -o name -n ltwieroyye]
[oc] serviceinstance.servicecatalog.k8s.io/dh-postgresql-apb
Running odo with args [odo app delete oahjjh -f]
[odo]  ✗  oahjjh app does not exists
Setting current dir to: /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog
Deleting project: ltwieroyye
Running odo with args [odo project delete ltwieroyye -f]
[odo] This project contains the following applications, which will be deleted
[odo] Application oahjjh
[odo] This application has following service that will be deleted
[odo] service named dh-postgresql-apb of type dh-postgresql-apb
[odo]  •  Deleting project ltwieroyye  ...
[odo]  ✓  Deleting project ltwieroyye [6s]
[odo]  ✓  Deleted project : ltwieroyye
Deleting dir: /tmp/863945455
Deleting dir: /tmp/526972802
Deleting dir: /tmp/454825721
• Failure [71.751 seconds]
odo service command tests
/home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:16
  When the application is deleted
  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:421
    should delete the service(s) in the application as well [It]
    /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:429
    No future change is possible.  Bailing out early after 0.505s.
    Running odo with args [odo app delete oahjjh -f]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0
    /home/travis/gopath/src/github.com/openshift/odo/tests/helper/helper_run.go:33
```

which indicates that the app we're trying to delete doesn't exist on the cluster. However, on my local minishift setup, it fails with:

```sh
NAME                  TYPE                  PLAN     STATUS      
dh-postgresql-apb     dh-postgresql-apb     dev      Deprovisioning               
Running odo with args [odo service list --app ogrssx]                                                                  
[odo] NAME                  TYPE                  PLAN     STATUS  
[odo] dh-postgresql-apb     dh-postgresql-apb     dev      Deprovisioning                                              
NAME                  TYPE                  PLAN     STATUS                                                                                                                                                                                   
dh-postgresql-apb     dh-postgresql-apb     dev      Deprovisioning      
Running odo with args [odo service list --app ogrssx]                                                                  
[odo]  ✗  There are no services deployed for this application                                         
Setting current dir to: /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog                                                                                                                                          
Deleting project: yxavlqckrg                                                                                           
Running odo with args [odo project delete yxavlqckrg -f]                                                               
[odo]  •  Deleting project yxavlqckrg  ...                                                                             
[odo]  ✓  Deleting project yxavlqckrg [6s]                                                                             
[odo]  ✓  Deleted project : yxavlqckrg               
Deleting dir: /tmp/190614664                                                                                           
Deleting dir: /tmp/516509255                                                                                           
Deleting dir: /tmp/798073594                                                                                           
                                                                                                                       
• Failure [116.506 seconds]                          
odo service command tests                                                                                              
/home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:16      
  When the application is deleted                                                                                      
  /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:421
    should delete the service(s) in the application as well [It]                                                       
    /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:429
                                                                                                                       
    No future change is possible.  Bailing out early after 0.344s.
    Running odo with args [odo service list --app ogrssx]                                                              
    Expected                                         
        <int>: 1                                                                                                       
    to match exit code:                                                                                                
        <int>: 0                                                                                                       
                                                                                                                       
    /home/dshah/go/src/github.com/openshift/odo/tests/helper/helper_run.go:33                                          
                                                                                                                       
                                                                                                                       
Summarizing 1 Failure:                                                                                                 
                                                                                                                       
[Fail] odo service command tests When the application is deleted [It] should delete the service(s) in the application as well                                                                                                                 
/home/dshah/go/src/github.com/openshift/odo/tests/helper/helper_run.go:33
                                                                                                                       
Ran 1 of 22 Specs in 116.506 seconds                                                                                   
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 21 Skipped
--- FAIL: TestServicecatalog (116.51s)                                                                                 
FAIL
```
In the tests I've called `helper.WaitForCmdOut` such that failure of the command doesn't cause the test spec to fail. Yet, it's failing. 

@amitkrout can you help me understand what's going wrong here? :disappointed:

dharmit:
> In the tests I've called helper.WaitForCmdOut such that failure of the command doesn't cause the test spec to fail. Yet, it's failing.
>
> @amitkrout can you help me understand what's going wrong here? 

Never mind. I was wrong in expecting things to work if the underlying command fails. `WaitForCmdOut` wraps `CmdShouldPass` so I figured a different approach. Added some sleep before checking the output of `odo service list`.

Hopefully the tests will pass this time. :smile:

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -416,4 +416,31 @@ var _ = Describe("odo service command tests", func() {
...
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+			// This timeout is needed because it takes some time to delete the
+			// application and the components/services that belong to it on the
+			// cluster
+			time.Sleep(60 * time.Second)
```
mik-dass: Why not ping the server constantly and check if the service list is deleted or not? This will pause the execution for 60 secs even if the deletion occurred after 30 secs.

dharmit: That's what I did in this commit https://github.com/openshift/odo/pull/2491/commits/efe3eb0f6edcbe041311c697b36253ed091b49dd.

But that command expects the command to exit successfully. It's a wrapper on top of `CmdShouldPass` whereas in this case, we are expecting `odo service list` command to fail.
Hence I went with this approach.

mik-dass: Why not create a wrapper on top of `CmdShouldFail`? Or use the use the oc command above and keep checking till the output doesn't contain `dh-postgresql-apb`?

dharmit: > Why not create a wrapper on top of CmdShouldFail? 

That's doable but I didn't do it since it's just one check. Had I come across more such scenarios, it would have been more worth adding it, IMO.

> Or use the use the oc command above and keep checking till the output doesn't contain dh-postgresql-apb?

That's what I was checking for earlier. But it exits with non-zero return code and causes the test to fail since it's a wrapper on top of `CmdShouldPass`.

mik-dass: > That's what I was checking for earlier. But it exits with non-zero return code and causes the test to fail since it's a wrapper on top of CmdShouldPass.

I think that shouldn't use CmdShouldPass or CmdShouldFail. It should just get the ouput of the command. @amitkrout Can you please have a look?

amitkrout: @mik-dass Thanks for pointing me. will have a look for sure

dharmit: @amitkrout do you have any comments on this? I'd really like this PR to be able to move forward even if that means having to make some changes from my side.

amitkrout: Sorry, i was fully packed up with other test stuff.

@dharmit I would suggest lets not encourage ```sleep()``` in our test code unless it is unavoidable.

Can you please explore if you can change the test code as per @mik-dass suggestion ?

dharmit: While I agree with what @mik-dass has suggested, if we do it in this PR, I will have to make changes at all those places in the code where `helper.WaitForCmdOut` is being called:
```sh
$ grep -r "helper.WaitForCmdOut" tests
tests/integration/servicecatalog/cmd_link_unlink_test.go:                       helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_link_unlink_test.go:                       helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_link_unlink_test.go:                       helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("odo", odoArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
tests/integration/servicecatalog/cmd_service_test.go:                   helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
```
I'd rather open a new issue based on this conversation which can then be picked up and fixed!

amitkrout: @dharmit You can, However your test script itself is not completed till you are not handling the sleep(). IMO this not a good test practice to write unhealthy script in the current test change and fix it in the next followup pr. 

Then the better approach would be w.r.t. the situation you are facing - First fix the dependency in a separate pr and merge. Then rebase the changes to this pr. WDYT ?

dharmit: Okay. Let's open a new issue based on this discussion, have team's thoughts on it and open a PR for it. Till that new issue is fixed, we can mark this PR as blocked on that issue. Does that sound OK?

kadel:
@dharmit this needs rebase :-(

dharmit:
@kadel  Actually this needs a PR before it can be reviewed again. Refer https://github.com/openshift/odo/pull/2491#discussion_r365660644 :disappointed:

tests/integration/servicecatalog/cmd_service_test.go
girishramnani: opened a PR which adds a test helper that does what you have mentioned @dharmit https://github.com/openshift/odo/pull/2676

kadel: Why not just use `helper.WaitForCmdOut` with `oc` command to check for `dh-postgresql-apb` service?

Just PR is stale for 2 months just because of this one thing?

dharmit: > Why not just use helper.WaitForCmdOut with oc command to check for dh-postgresql-apb service?
oc get serviceinstance

Because, IIRC (since it's 2 months), `WaitForCmdOut` doesn't work if the cmd fails. We're trying to execute a command that's expected to fail.

> Is this PR stale for 2 months just because of this one thing?

That's my bad. I got into OperatorHub piece and left my sight from this one. Just discussed on a call. I'll address it this week.

kadel: > Because, IIRC (since it's 2 months), `WaitForCmdOut` doesn't work if the cmd fails. We're trying to execute a command that's expected to fail.

Using `oc get serviceinstance` would work with `WaitForCmdOut`

dharmit: I've used `oc get serviceinstance` with `WaitForCmdOut` and pushed the changes. When I tried to run this test locally, it would not succeed in spite of having found the sub-string that we're looking for in the output of the command. I've pushed it however to get feedback from the CI.

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -435,4 +435,27 @@ var _ = Describe("odo service command tests", func() {
...
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+
+			ocArgs = []string{"get", "serviceinstances"}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "[oc] No resources found.")
```
kadel: This doesn't look right, I should be without `[oc]`. `[oc]` Is something that is added by our logging, but it is not part of normal `oc` output

```suggestion
				return strings.Contains(output, "No resources found.")
```

kadel: This doesn't look right, I should be without `[oc]`. `[oc]` Is something that is added by our logging, but it is not part of normal `oc` output

```suggestion
				return strings.Contains(output, "No resources found.")
```

dharmit: I added that to "exactly" match what we are seeing in the output. It doesn't work with or without that. :confused:

dharmit:
Failing at a different place than what I'm seeing while running tests locally:
```sh
Failure [23.111 seconds]
odo service command tests
/home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:15
  When the application is deleted
  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:439
    should delete the service(s) in the application as well [It]
    /home/travis/gopath/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:444
    No future change is possible.  Bailing out early after 0.243s.
    Running odo with args [odo app delete ogffqx -f]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0
    /home/travis/gopath/src/github.com/openshift/odo/tests/helper/helper_run.go:34
```
Maybe a flake? Triggering again for now.

/retest

dharmit:
@kadel @amitkrout the travis check is failing with same error as it fails for me in my local setup and I don't understand why that's the case or what I'm doing wrong. https://travis-ci.com/github/openshift/odo/jobs/298915517#L733-L909

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -435,4 +435,27 @@ var _ = Describe("odo service command tests", func() {
...
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+
+			ocArgs = []string{"get", "serviceinstances"}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "No resources found.")
```
girishramnani: @dharmit can you please print the `output` content here? I think `No resources found.` might be printed on stderr stream which `WaitForCmdout` doesn't check

dharmit: Sure, let me check.

girishramnani: confirmed 
```
Running oc with args [oc get serviceinstances]
[oc] No resources found.
GIRISH output=   # its empty here
```

girishramnani: confirming that `No resources found.` is showing in the err stream

dharmit:
Travis failure:
```
• Failure [31.691 seconds]
odo component command tests
/home/travis/gopath/src/github.com/openshift/odo/tests/integration/cmd_cmp_test.go:13
  when component is in the current directory and --project flag is used
  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/component.go:475
    creates a local python component, pushes it and then deletes it using --all flag [It]
    /home/travis/gopath/src/github.com/openshift/odo/tests/integration/component.go:510
    No future change is possible.  Bailing out early after 22.515s.
    Running odo with args [odo push --context /tmp/148843195]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0
    /home/travis/gopath/src/github.com/openshift/odo/tests/helper/helper_run.go:34
```
4.2 integration failure:
```
Ginkgo ran 1 suite in 1h16m49.696326188s
Test Suite Failed
2020/03/19 11:47:56 Container test in pod integration-e2e-benchmark failed, exit code 2, reason Error
[Fail] odo preference and config command tests when using --now with config command [It] should successfully set and unset variables 
```
I'm guessing these are flake.

/test v4.2-integration-e2e-benchmark

dharmit:
/hold cancel

ready for another review @mik-dass @adisky @cdrage @kanchwala-yusuf

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -435,4 +435,27 @@ var _ = Describe("odo service command tests", func() {
...
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+
+			ocArgs = []string{"get", "serviceinstances"}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "Deprovisioning")
```
amitkrout: ```Deprovisioning``` literally seems the process is still going on. Mab be ```Deprovisioned``` would be the best fit for this. But it seems we don't have control over the ```oc``` output. So @dharmit Can you please confirm does the substring ```Deprovisioning``` means all the services link to the application are deleted.

dharmit: `Deprovisioning` indeed means that the process of deletion of service is still going on. Once the service is deleted, the output of `oc get serviceinstances` becomes something like `No resources found...` since there are no service instances in the project. It won't print `Deprovisioned` for deleted service.

dharmit: I'm trying to see if doing `ocArgs = []string{"get", "serviceinstances", "2>&1"}` helps. So far, it doesn't seem to be working on my test setup.

dharmit: Well, trying to redirect stderr to stdout is not working:
```sh
Running oc with args [oc get serviceinstances 2>&1]
[oc] Error from server (NotFound): serviceinstances.servicecatalog.k8s.io "2>&1" not found

Failure [20.698 seconds]
odo service command tests
/home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:15
  When the application is deleted
  /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:439
    should delete the service(s) in the application as well [It]
    /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:444

    No future change is possible.  Bailing out early after 0.111s.
    Running oc with args [oc get serviceinstances 2>&1]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0

    /home/dshah/go/src/github.com/openshift/odo/tests/helper/helper_run.go:34
```
Doesn't work when I use `odo` command either:
```sh
[odo]  ✗  unknown command "2>&1" for "odo service list"

• Failure [19.730 seconds]
odo service command tests
/home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:15
  When the application is deleted
  /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:439
    should delete the service(s) in the application as well [It]
    /home/dshah/go/src/github.com/openshift/odo/tests/integration/servicecatalog/cmd_service_test.go:444

    No future change is possible.  Bailing out early after 0.040s.
    Running odo with args [odo service list --app kushnf 2>&1]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0

    /home/dshah/go/src/github.com/openshift/odo/tests/helper/helper_run.go:34
```

Looks like using shell way of redirecting stderr to stdout doesn't work well from Go code. :disappointed:

girishramnani:
one quick fix here would be to augment the `WaitForCmdOut` with an extra var-arg `includeStderr`
```
func WaitForCmdOut(program string, args []string, timeout int, errOnFail bool, check func(output string) bool, includeStderr ...bool) bool {
	pingTimeout := time.After(time.Duration(timeout) * time.Minute)
	// this is a test package so time.Tick() is acceptable
	// nolint
	tick := time.Tick(time.Second)
	for {
		select {
		case <-pingTimeout:
			Fail(fmt.Sprintf("Timeout out after %v minutes", timeout))

		case <-tick:
			session := CmdRunner(program, args...)
			Eventually(session).Should(gexec.Exit(0), runningCmd(session.Command))
			session.Wait()
			output := string(session.Out.Contents())

			if len(includeStderr) > 0 && includeStderr[0] {
				output += "\n" // just so we can separate stuff
				output += string(session.Err.Contents())
			}
			if check(strings.TrimSpace(string(output))) {
				return true
			}
		}
	}
}

```

so now the WaitForCmdOut would be like
```
helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
	return strings.Contains(output, "No resources found.")
}, true) // <- observe the extra true here. and this doesn't break any of existing code because 
           //`includeStderr` is a var-args parameter.
```

I know this might not be the most documentation friendly approach but soon we need to redo the helpers ( all of them ) to something nice which covers all our cases and are clean to use. Until that day this looks ok.

pkg/application/application.go
cdrage: nitpick, it's `Service Catalog`

tests/integration/servicecatalog/cmd_service_test.go
```diff
@@ -435,4 +435,27 @@ var _ = Describe("odo service command tests", func() {
...
+				"-p", "postgresql_user=luke", "-p", "postgresql_password=secret",
+				"-p", "postgresql_database=my_data", "-p", "postgresql_version=9.6")
+			ocArgs := []string{"get", "serviceinstance", "-o", "name", "-n", project}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "dh-postgresql-apb")
+			})
+
+			helper.CmdShouldPass("odo", "app", "delete", app, "-f")
+
+			ocArgs = []string{"get", "serviceinstances"}
```
cdrage: This LGTM!

Issue #2464, opened 2019-12-16
provide admin commands to install operators and one-off setup 


girishramnani:
[kind/Feature]

## Which functionality do you think we should add?
To start **using** `Operators` on openshift, an administrator needs to install operators first. We can provide the feature to interactively install these operators easily ( only possible for an admin user ).
Also install necessary operator like Service Binding Operator as part of a one-off `setup` command.

dharmit:
AFAIK, our users are mostly  going to have CRC on their laptops or a OCP 4.x cluster managed by the admins. So, should we just mention this in documentation instead of having commands in odo to install operators? Do we want to have installation of Operators as a part of odo tool? Or do we want to focus on spinning up service from installed operators? WDYT @kadel @girishramnani ?

kadel:
We should focus on spinning up services from installed operators.
Similarly, how we are not installing ServiceCatalog Brokers because that is cluster admin's job we don't need to spend much time on providing commands for installing Operators, as this is usually cluster admin's job.
This might change in the future, as there are some plans to allow user accounts to install certain verified operators.
This is still valid concern, but it is not priority right now.

dharmit:
@kadel @girishramnani I think we're in consensus to not do admin activities like installing operators using odo. Should we close this?

Issue #2461, opened 2019-12-16
Validation of cluster's capabilities to run operators and listing services


girishramnani:
[kind/Feature]

## Which functionality do you think we should add?
Checking the cluster version and validating the capabilities for running operators.
Checking if cluster has 
- some minimum `OperatorSource`s present to be useful
- has the service binding operator installed
- check the version of openshift

Listing all the possible operators and services that a user can create is also part of the this issue

## Why is this needed?
We want the operator backed services and service catalog backed services to co-exists. which means a user could run both a service-catalog service and operator using odo.
To allow this initially we need to check 
- possibility to run service-catalog services ( this becomes a necessary check for openshift 4+ )
- possibility to run operator-backed services

kadel:
> has the service binding operator installed

Is ServiceBinging operator going to be a hard requirement?

> check the version of openshift

It should not depend on OpenShift version.  The important part is the presence of the OperatorHub.



I haven't seen mentioned listing of available operators in any other issues, so I assume that it should be part of this one.

 - When the cluster has both, ServiceCatalog and OperatorHub the listing should contain services that can be created by either of the methods (provisioner)
```
▶ odo catalog list services
NAME                                                  PROVISIONER
mariadb-ephemeral                                     ServiceCatalog
mongodb-ephemeral                                     ServiceCatalog
mysql-ephemeral                                       ServiceCatalog
mongodb-enterprise.v1.2.4                             OperatorHub
mariadb-operator-0.1.3-6                              OperatorHub
```

- When the cluster has just OperatorHub enabled the `odo catalog list services` should list all available services that can be deployed using operators, ServiceCatalog is silently ignored.
```
▶ odo catalog list services
NAME                                                  PROVISIONER
mongodb-enterprise.v1.2.4                             OperatorHub
mariadb-operator-0.1.3-6                              OperatorHub
```
- When the cluster has just ServiceCatalog enabled the `odo catalog list services` should list all available services that can be deployed using ServiceCatalog, OperatorHub is silently ignored.
```
▶ odo catalog list services
NAME                                                  PROVISIONER
mariadb-ephemeral                                     ServiceCatalog
mongodb-ephemeral                                     ServiceCatalog
mysql-ephemeral                                       ServiceCatalog
```

girishramnani:
> Is ServiceBinging operator going to be a hard requirement?

I mean if you just want to start the operators and dont wanna link it with components then its not required

girishramnani:
[...]

Issue #2161, opened 2019-09-23
List & Provision Services from the in-cluster installed operators (Dev Catalog)


sbose78:
[kind/Enhancement]

## Which functionality do you think we should update/improve?
`odo catalog list services`

## Why is this needed? 
Support in-cluster Operator Catalog as the primary way to look-up and provision services, as an alternative to Service Catalog.

It should behave as demonstrated here https://asciinema.org/a/261224  

## Challenges
1. By default, none of the operators in the OLM Catalog are installed. `odo catalog list services` could potentially be empty - unless we silently install the operator if not installed before usage.
2. Creating a service would entail creating a CR that would be managed by the OLM operator. There's no good way to know which is the primary CR associated with a Database operator for example.

## Implementation Approaches & Existing Work
* Appears no existing CLI tools exist
* (from Evan, OLM lead) 

> Mostly this will be vendoring in OLM's apis. We are looking to place that in `operator-framework/api`, but for now they're in `operator-framework/operator-lifecycle-manager`. The dynamic forms in console are driven by the metadata available on the ClusterServiceVersion and the schemas of the CRDs owned by the operators.

* (from Rob S, OLM PM) 
Some brainstorming...
> 1. What type of service do you want to create?
> - api
> - database
> - ...
> 2. What type of database?
> [...list of CR types from OLM with that category...]
> 
> [...grab descriptions from CSV...]
> 3. Enter required values
> [...pull these from the schema with default values...]
> 4. Provide a name for your database
> foobar
> 5. Wait for the service to deploy?
> Yes
> 6. Creating the following resources:
> [...match resource tab from Console...]

## User Stories

- [ ] Validating the cluster for operator and service catalog capability and odo service list https://github.com/openshift/odo/issues/2461
- [ ] odo service describe for operators + services https://github.com/openshift/odo/issues/2615
- [ ] Instantiating operators using CRDs https://github.com/openshift/odo/issues/2462
- [x] status command for operators https://github.com/openshift/odo/issues/2479
- [ ] Use the service binding operator to link operator and a component https://github.com/openshift/odo/issues/2463
- [x] enhance the odo service delete command to allow deletion of operators https://github.com/openshift/odo/issues/2792

kadel:
What is the best way to list installed operators on the cluster?

sbose78:
oc get csvs

tmds:
Is `odo service` unusable until this gets resolved?

kadel:
[...]

PR #2498, opened 2020-01-07
Proposal for Operator Hub integration with odo

dharmit:
**What type of PR is this?**

/kind design
/kind feature

Proposes a plan for Operator Hub integration with odo

NA

Easier way to navigate the markdown: https://github.com/dharmit/odo/blob/operator-hub-proposal/docs/proposals/operator-hub.md

### DIFF
--- a/docs/proposals/operator-hub.md
+++ b/docs/proposals/operator-hub.md
@@ -1,0 +1,197 @@
+# Operator Hub integration
+odo currently supports creation, listing and deletion of services created from
+the Service Catalog. With newer versions of OCP, Service Catalog is being
+deprecated in favour of Operator Hub.
+
+This proposal discusses adding similar support for Operator Hub as is currently
+available for Service Catalog. Idea is to be able to create, list and delete
+the services (basically Custom Resources in Operator framework terminology) on
+an OpenShift or Kubernetes cluster. Considering the ongoing `kclient` and DevFile
+efforts, we intend to add support for Kubernetes from day 1. However, there
+will be certain prerequisites that need to be satisfied in case of Kubernetes.
+
+Table of Contents
+- [Which Operators do we support](#which-operators-do-we-support)
+- [Installing Operators](#installing-operators)
+- [List the services/operators](#list-the-servicesoperators)
+  - [Both Service Catalog and Operator Hub are enabled](#both-service-catalog-and-operator-hub-are-enabled)
+  - [Only Service Catalog is enabled](#only-service-catalog-is-enabled)
+  - [Only Operator Hub is enabled](#only-operator-hub-is-enabled)
+  - [Kubernetes](#kubernetes)
+- [Link components with operands](#link-components-with-operands)
+- [Listing the services](#listing-the-services)
+- [Describe deployed services](#describe-deployed-services)
+- [Delete the service](#delete-the-service)
+- [Testing on Kubernetes](#testing-on-Kubernetes)
+- [Questions](#questions)
+- [GitHub issues related to this task](#github-issues-related-to-this-task)
+
+## Which Operators do we support
+Operator Hub is not the only place to find/list Operators that could be
+installed on a OpenShift/k8s cluster. There’s also KUDO. However, to start
+things off, we must focus on a limited number of Operators and iterate from
+there. Hence, we will focus only on OperatorHub (kind of obvious 😉).
+
+There are three main categories of Operators as described in the OpenShift
+docs. We could focus on one of those categories or select specific Operators
+from each of the categories.
+
+From a development perspective, we might just take a CRD definition and throw
+it to the Kubernetes API. The controller running in background would take care
+of spinning up the service. So, it wouldn’t matter a lot to differentiate
+between what we support and what we don’t. However, limiting our scope to a few
+Operators will keep us from being overwhelmed.
+
+## Installing Operators
+There’s an open issue about adding capability in odo to install Operators and
+installing “Service Binding Operator” (which helps bind applications with
+Operator-backed services) as a one-off command.
+
+As far as creating services from Service Catalog is concerned, we expect the
+users to enable Service Catalog in their Minishift or OpenShift cluster before
+being able to see any services in `odo catalog list services` output. Why should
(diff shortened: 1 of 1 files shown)


docs/proposals/operator-hub.md
```diff
@@ -0,0 +1,197 @@
...
+- [Link components with operands](#link-components-with-operands)
+- [Listing the services](#listing-the-services)
+- [Describe deployed services](#describe-deployed-services)
+- [Delete the service](#delete-the-service)
+- [Testing on Kubernetes](#testing-on-Kubernetes)
+- [Questions](#questions)
+- [GitHub issues related to this task](#github-issues-related-to-this-task)
+
+## Which Operators do we support
+Operator Hub is not the only place to find/list Operators that could be
```
kadel: The more precise way would be to say that our goal is to use Operator Lifecycle Manager (OLM)

docs/proposals/operator-hub.md
```diff
@@ -0,0 +1,197 @@
...
+of spinning up the service. So, it wouldn’t matter a lot to differentiate
+between what we support and what we don’t. However, limiting our scope to a few
+Operators will keep us from being overwhelmed.
+
+## Installing Operators
+There’s an open issue about adding capability in odo to install Operators and
+installing “Service Binding Operator” (which helps bind applications with
+Operator-backed services) as a one-off command.
+
+As far as creating services from Service Catalog is concerned, we expect the
```
kadel: I would explicitly say that odo will just  "instantiate" operands and that it is expected that someone with the cluster-admin will enable (install) operators first.

docs/proposals/operator-hub.md
```diff
@@ -0,0 +1,197 @@
...
+being able to see any services in `odo catalog list services` output. Why should
+this be different for Operators? We should mention in the documentation about
+using `kube:admin` user for CRC and/or contacting administrator to install
+required Operators in OpenShift cluster.
+
+As for using Service Binding Operator, we also should explore the possibility
+of linking services to components without using the Operator.
+
+## List the services/operators
+Taking this from Tomas’s comment on the issue. Added column for available
```
kadel: it would be good to include a link to the mentioned comment https://github.com/openshift/odo/issues/2461#issuecomment-566577064

docs/proposals/operator-hub.md
```diff
@@ -0,0 +1,197 @@
...
+could augment this to have the information about the services created from the
+Operators. To make a distinction between the source of the services in the
+cluster, we could add a column to the output just like we plan to do for
+listing available/installed operators via `odo catalog list services`.
+
+```sh
+$ odo services list
+NAME                          PLANS                    PROVISIONER
+mongodb-enterprise.v1.2.4     prod                     OperatorHub
+mariadb-operator-0.1.3-6      ephemeral,persistent     ServiceCatalog
```
kadel: The command lists service instances. Each service instance uses only one plan that was specified during the service creation

docs/proposals/operator-hub.md
```diff
@@ -0,0 +1,197 @@
...
+3. Decide what should be shown to the user as the output on CLI
+4. Implement the command for both ServiceCatalog and OperatorHub
+
+## Delete the service
+[GitHub issue](https://github.com/openshift/odo/issues/2481)
+
+At the moment, we use `odo service delete <service-name>` to delete the service
+deployed in OpenShift cluster. We should be able to delete the service
+deployed using Operator Hub with the same command.
+Testing on Kubernetes
```
kadel: Is this supposed to be under the `Delete the service` headline?
It looks like ti might be copy/paste error.

docs/proposals/operator-hub.md
sspeiche: +1

sspeiche:
I'm confused by this proposal, why are we putting admin tasks into the developer client?

Is this proposal linked to https://github.com/openshift/odo/issues/2161 ?

docs/proposals/operator-hub.md
sspeiche: Agree with @kadel an admin needs to do something outside of `odo` to make the operator/CRD available.

dharmit: I agree with your concern @kadel & @sspeiche, but if you read the paragraph fully, you'll see that your concern is exactly what is proposed. :wink: There are more words because we have #2464 which talks about letting developers do "few" admin tasks but I don't agree with that notion/idea. Let me know if this is still unclear/vague/lengthy and I'll change it.

docs/proposals/operator-hub.md
dharmit: Yes, this section is about listing the service instances or the services/operands created using Operators. I pasted the wrong output.

dharmit:
> One big part that I'm missing there is actual service creation.
odo service create We don't have to define it right now, but there should be a section with at least TODO note as this is the most important part

What an irony on my part! :disappointed: Thanks for this note. At this moment, we don't have a clear idea about how `odo service create` command will look for Operator Hub. But it could look something like:
```sh
$ odo service create <operator-name> <service-name> --crd <crd-name> -p parameter1=value1 -p parameter2=value2
```

Adding to TODO. Thanks for catching this!

docs/proposals/operator-hub.md
dharmit: OLM itself is an operator that helps manage lifecycle of the Operators. It seemed like going into the implementation detail so I didn't mention much in this doc. I'll add more info about it. :+1:

dharmit:
> I'm confused by this proposal, why are we putting admin tasks into the developer client?

@sspeiche we surely don't want the developers to be concerned with admin tasks. Things like installing Operators and Operator Lifecycle Manager (for Kubernetes since it's pre-installed in OCP/crc) should be admin's concern. Can you let me know which part of the document feels like we're putting admin tasks into developer client?

> Is this proposal linked to #2161 ?

Yes. I think we could very well close that issue once we have reached a quorum on this proposal. WDYT @kadel ?

kadel:
/approve
I think that we can merge this. The basic idea is described well. 
We should be adding more details as progress.

PR #2504, opened 2020-01-10
Bump release version 1.0.2 -> 1.0.3

mohammedzee1000:
Preparing to release v1.0.3
**What type of PR is this?**

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind code-refactoring

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -16,7 +16,7 @@ LABEL com.redhat.component=atomic-openshift-odo-cli-artifacts-container \
     summary="This image contains the Linux, Mac and Windows version of odo"
 
 # Change version as needed. Note no "-" is allowed
-LABEL version=1.0.2
+LABEL version=1.0.3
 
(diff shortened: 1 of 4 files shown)


mohammedzee1000:
> Don't we need to change the version info in `build/VERSION` file?
> 
we do it after the release.
> And what about changing image reference in `pkg/occlient/occlient.go` in the variable `defaultBootstrapperImage`? Do we do that after updating the init image?
Already updated

kadel:
> Don't we need to change the version info in `build/VERSION` file?

That should be done only after release. This file is used by update notifier and we don't want for users to be notified about the new version before it is available to download.



> 
> And what about changing image reference in `pkg/occlient/occlient.go` in the variable `defaultBootstrapperImage`? Do we do that after updating the init image?

that was already updated. image has its own versioning and it doesn't have to be updated with each release

dharmit:
Thanks for the answers @mohammedzee1000 @kadel 

/lgtm

PR #2520, opened 2020-01-20
add devfile proposal

### LINKED ISSUE #2290: devfile support - implementation plan
implementation plan https://docs.google.com/document/d/16OUCIvuoTPuJfqS32RLNDgJS1ZJ-sY3NDR9ZMP6wQH4/edit#

kadel:
**What type of PR is this?**

> /kind design

High-level design and description of [devfile](https://github.com/redhat-developer/devfile/) support in odo.

This is just https://docs.google.com/document/d/16OUCIvuoTPuJfqS32RLNDgJS1ZJ-sY3NDR9ZMP6wQH4/edit# converted into markdwon

Fixes #2290

### DIFF
--- a/docs/proposals/devfile.md
+++ b/docs/proposals/devfile.md
@@ -1,0 +1,64 @@
+
+# Devfile support in odo
+
+To better integrate odo with other developer tooling we need to align on one common definition of a developer workspace and application lifecycle.
+
+[Eclipse Che](https://github.com/eclipse/che) project already created and started using [Devfile](https://github.com/redhat-developer/devfile). Devfile is a YAML file that defines developer workspace in Che environment. It is structured in such a way that can be used also in odo.
+
+Implementing Devfile support in odo will create a great benefit for users, as it will allow them to easily switch tools with no additional configuration. Adding support for new languages will be also much easier with devfiles, as it will be only a matter of using the right devfile template.
+
+## Implementation plan
+
+### Consuming current version of Devfile in odo
+
+1. Create a golang library for parsing and reading data from Devfile
+    * Devfile is versioned file so the design of a library needs to be able to handle and support multiple Devfile versions at the same time
+
+        **Parser high level design**
+        * The parser would read the provided devfile.
+        * From the devfile, parser would look for the API Version of the devfile and validate whether the API version in supported in odo.
+        * The devfile schema is based on <http://json-schema.org/draft-07/schema#>
+        * Hence, the parser would validate the integrity of the provided devfile with the devfile json schema of the respective API version.
+        * Furthermore, parser would try to map devfile sections (projects, components, commands etc) to their respective Golang structs.
+        * The idea is to have same Golang structs for the devfile sections across different API version contingent to the condition that the sections themselves are compatible across the API versions..
+        * This is very similar to the way Golang structs are handled across different GVK versions in kubernetes. [https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32](https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32)
+        * The implementation details for this with respect to odo are being worked upon and once confirmed would be updated here.
+
+2. Implement experimental support for reading current version of a Devfile and bootstrap odo components from a Devfile.
+    * If there is a devfile.yml and no .odo/config.yaml `odo push` will use Devfile.yaml and create necessary resources in Kubernetes cluster.
+    * Before devfile has support for category of a commands (application lifecycle commands like run, build, test etc..) we could just add flags to odo push that will allow to specify the name of the command that should be used for build and run, as currently those are only two commands that odo cares about. \
+For [springboot petstore example](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/java-web-spring/devfile.yaml) it would look like this: `odo push --run-command="maven build" --build-command="run webapp"`.
+This will be only temporary measure until we can mark commands directly in devfile.
+    * The best approach would probably be to use “fat” pods. Component will consist of a single Pod typically with two containers - “build” container and “run” container. Both containers will share a volume with the source code (emptyDir or PVC?)
+    * **Build container** will use the image (dockerImage) based on a component field in build command definition in devfile. The main command should be something that just sleeps indefinitely.  Only ad-hoc commands (build command) will be executed in this container.
+    * **Run container** will use the image  (dockerImage) based on a component field in run command definition in devfile. The main command will be the process that can control other processes. We could reuse Go implementation of the SupervisorD as it is currently used in odo s2i flow. \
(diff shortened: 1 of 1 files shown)


docs/proposals/devfile.md
```diff
@@ -0,0 +1,64 @@
...
+        * This is very similar to the way Golang structs are handled across different GVK versions in kubernetes. [https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32](https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32)
+        * The implementation details for this with respect to odo are being worked upon and once confirmed would be updated here.
+
+2. Implement experimental support for reading current version of a Devfile and bootstrap odo components from a Devfile.
+    * If there is a devfile.yml and no .odo/config.yaml `odo push` will use Devfile.yaml and create necessary resources in Kubernetes cluster.
+    * Before devfile has support for category of a commands (application lifecycle commands like run, build, test etc..) we could just add flags to odo push that will allow to specify the name of the command that should be used for build and run, as currently those are only two commands that odo cares about. \
```
dharmit: If it's not too much of a hassle, we could try adding support for category commands instead of adding flags to `odo push` because making changes later will also mean that plugins dependant on odo would have to change the way things are done. This can cause more breakage in those tools than we can initially estimate.

docs/proposals/devfile.md
```diff
@@ -0,0 +1,64 @@
...
+        * Furthermore, parser would try to map devfile sections (projects, components, commands etc) to their respective Golang structs.
+        * The idea is to have same Golang structs for the devfile sections across different API version contingent to the condition that the sections themselves are compatible across the API versions..
+        * This is very similar to the way Golang structs are handled across different GVK versions in kubernetes. [https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32](https://github.com/kubernetes/kubernetes/blob/ded2ff39c381b122951dd103dafb291755791921/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L32)
+        * The implementation details for this with respect to odo are being worked upon and once confirmed would be updated here.
+
+2. Implement experimental support for reading current version of a Devfile and bootstrap odo components from a Devfile.
+    * If there is a devfile.yml and no .odo/config.yaml `odo push` will use Devfile.yaml and create necessary resources in Kubernetes cluster.
```
dharmit: What about the case when both `devfile.yml` and `.odo/config.yml` exist?

docs/proposals/devfile.md
```diff
@@ -0,0 +1,64 @@
...
+
+* Add a command (like build, run)
+* Url/Ingres/Routes support
+* Storage support
+* Linking multiple components (applications)
+* Bootstrapping services from OperatorHub and linking it to the component (application)
+
+### Implement odo operations on top of a Devfile
+
+* Implement commands that will modify Devfile.yaml. Uses shouldn’t need to touch Devfile.yaml, everything should be handled by odo commands \
```
dharmit: `s/Uses shouldn't/Users shouldn't`

docs/proposals/devfile.md
```diff
@@ -0,0 +1,64 @@
...
+For example:
+  * odo create url - instead of modifing .odo/config.yaml it will add all necessary information into the Devfile.yaml
+  * odo delete url - will delete url information from Devfile.yaml
```
dharmit: Nitpick but for the sake of correctness, it's `odo url craete` and `odo url delete`.

docs/proposals/devfile.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. Create a golang library for parsing and reading data from Devfile
+    * Devfile is versioned file so the design of a library needs to be able to handle and support multiple Devfile versions at the same time
+
+        **Parser high level design**
+        * The parser would read the provided devfile.
+        * From the devfile, parser would look for the API Version of the devfile and validate whether the API version in supported in odo.
+        * The devfile schema is based on <http://json-schema.org/draft-07/schema#>
+        * Hence, the parser would validate the integrity of the provided devfile with the devfile json schema of the respective API version.
+        * Furthermore, parser would try to map devfile sections (projects, components, commands etc) to their respective Golang structs.
```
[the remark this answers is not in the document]
kadel: Not sure what you mean by that. We define the odo component as a building block of a bigger application (microservice). In that sense, it will be still the same component

docs/proposals/devfile.md
kadel: Any changes to devfile need to be consulted with Che team and it Che needs to support those also.
Devfile implementation will have the status of an experiment (probably hidden behind feature flag) until we have a stable interface.

docs/proposals/devfile.md
kadel: There will be an experimental feature flag (https://github.com/openshift/odo/issues/2536) if enabled and both files exists, the devfile.yaml will have higher priority. If experimental is not enabled it will ignore the devfile.yaml file

docs/proposals/devfile.md
kadel: fixed

docs/proposals/devfile.md
kadel: fixed

docs/proposals/devfile.md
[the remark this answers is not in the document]
dharmit: I mean that, AFAIK, the word `component` means different things to odo and DevFile. Are we adding any information about that here or do we plan to do that in the docs?

docs/proposals/devfile.md
dharmit: I guess there's some misunderstanding here. I understand that changes to DevFile need to be discussed with Che team. But I'm not proposing a change to DevFile here.

I'm just asking to add support for category commands to odo right away instead of having flags for it on a temporary basis. So, odo should be able to understand category commands in the DevFile (if they are already a part of current version of DevFile we're developing against.)

docs/proposals/devfile.md
dharmit: Ack, thanks.

docs/proposals/devfile.md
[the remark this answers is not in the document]
kadel: There are discussions about renaming this in Devfile. So hopefully it will get renamed in future devfile spec revision.
If not that we will have to address this in docs.
I don't think it needs to be mentioned in this doc.

docs/proposals/devfile.md
kadel: That is the problem. There is no such thing as category of commands in the current devfile.
The goal is to make odo work with the current devfiles version.
I don't see any other option than adding the flags until devfile has proper support for this.

We could try to do some smart autodetection based on the command name. Like if the command name exactly matches "build" we assume that is built command and the same for "run".  But there is nothing in devfile spec that dictates how the name should look like, it is just free form string. That was originally meant for displaying in the UI for the user to select one.
Even if we try to do this smart detection we will need the flags to allow user overwrite it if is wrong.

We can discuss this in later stages.

girishramnani:
any update on this PR?

Issue #2248, opened 2019-10-04
Deleted files are not remove from Java component


kadel:
Java assemble script copies the source files somewhere where outside the destination directory `/tmp/src`, where it does the build.

```
odo create java
odo push
# delete a file that is important and that should break the build
rm something-important
odo push

#everything is ok, but it shouldn't be
```

amitkrout:
@kadel Hard to understand the scenario in the description area you mentioned. Can you please elaborate it

kadel:
Did you mean this part?

```
odo create java
odo push
# delete a file that is important and that should break the build
rm something-important
odo push

#everything is ok, but it shouldn't be
```

amitkrout:
> Did you mean this part?
> 
> ```
> odo create java
> odo push
> # delete a file that is important and that should break the build
> rm something-important
> odo push
> 
> #everything is ok, but it shouldn't be
> ```

Got it
Then it should be same case with nodejs component i guess

amitkrout:
> > Did you mean this part?
> > ```
> > odo create java
> > odo push
> > # delete a file that is important and that should break the build
> > rm something-important
> > odo push
> > 
> > #everything is ok, but it shouldn't be
> > ```
> 
> Got it
> Then it should be same case with nodejs component i guess

@kadel Tested with url creation for nodejs component and it seems to work fine
Nodejs source - https://github.com/openshift/odo/tree/master/tests/examples/source/nodejs

amitkrout:
> > > Did you mean this part?
> > > ```
> > > odo create java
> > > odo push
> > > # delete a file that is important and that should break the build
> > > rm something-important
> > > odo push
> > > 
> > > #everything is ok, but it shouldn't be
> > > ```
> > 
> > 
> > Got it
> > Then it should be same case with nodejs component i guess
> 
> @kadel Tested with url creation for nodejs component and it seems to work fine
> Nodejs source - https://github.com/openshift/odo/tree/master/tests/examples/source/nodejs

Tested on java source and I can confirm this breaks for java component

kadel:
> > Did you mean this part?
> > ```
> > odo create java
> > odo push
> > # delete a file that is important and that should break the build
> > rm something-important
> > odo push
> > 
> > #everything is ok, but it shouldn't be
> > ```
> 
> Got it
> Then it should be same case with nodejs component i guess

nodejs should be fine (it will broke as it should), it is not copying source code to another directory, everything is happening inside the `/tmp/src`

[...]

PR #2522, opened 2020-01-21
add test for java component to verify reflection of deleted files

girishramnani:
**What type of PR is this?**
/kind test

Add a test to verify for java component that local file deletions get reflected on the component as well. 

Fixes https://github.com/openshift/odo/issues/2248

The tests should pass

### DIFF
--- a/tests/helper/helper_oc.go
+++ b/tests/helper/helper_oc.go
@@ -126,7 +126,6 @@ func (oc *OcRunner) CheckCmdOpInRemoteCmpPod(cmpName string, appName string, prj
 	podName := strings.Replace(outPodName, "'", "", -1)
 	session := CmdRunner(oc.path, append([]string{"exec", podName, "--namespace", prjName,
 		"-c", cmpDCName, "--"}, cmd...)...)
-	Eventually(session).Should(gexec.Exit())
 	stdOut := string(session.Wait().Out.Contents())
 	stdErr := string(session.Wait().Err.Contents())
 	if stdErr != "" {
--- a/tests/helper/helper_oc.go
+++ b/tests/helper/helper_oc.go
@@ -126,7 +126,7 @@ func (oc *OcRunner) CheckCmdOpInRemoteCmpPod(cmpName string, appName string, prj
 	podName := strings.Replace(outPodName, "'", "", -1)
 	session := CmdRunner(oc.path, append([]string{"exec", podName, "--namespace", prjName,
 		"-c", cmpDCName, "--"}, cmd...)...)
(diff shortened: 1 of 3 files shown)


tests/helper/helper_oc.go
```diff
@@ -126,7 +126,7 @@ func (oc *OcRunner) CheckCmdOpInRemoteCmpPod(cmpName string, appName string, prj
 	podName := strings.Replace(outPodName, "'", "", -1)
 	session := CmdRunner(oc.path, append([]string{"exec", podName, "--namespace", prjName,
 		"-c", cmpDCName, "--"}, cmd...)...)
-	Eventually(session).Should(gexec.Exit(0))
+	Eventually(session).Should(gexec.Exit())
```
girishramnani: This is an important change because when we run a failing command like `listing a file that doesn't exist` the session exits with a non-zero exit code. This check wouldn't help us verifying such failing commands and hence we need to remove it.

prietyc123: As we are not validating the exit returned code then I don't think line `Eventually(session).Should(gexec.Exit())` is required over here.

`stdOut` and `stdErr` will not be affected anyway.

girishramnani: actually if you read the docs, the checking of `returned code` is optional. This statement confirms that the session exits successfully. so its still necessary.

prietyc123: @girishramnani Right, but when this return code is not being validated and in the very next line we are waiting for the `stderr` and `stdout` which itself confirms that session has completed. So, I think we don't need the statement. I might be wrong but it would be great if you check locally after removing whatever I am suggesting.

girishramnani: Removed

girishramnani:
```
[odo]  ✗  Waiting for build to finish [30s]
[odo]  ✗  Failed to create component with name cmp-git. Please use `odo config view` to view settings used to create component. Error: failed to build component with args {Name:cmp-git SourcePath:https://github.com/openshift/nodejs-ex SourceRef: SourceType:git ImageName:nodejs EnvVars:[DEBUG_PORT=5858] Ports:[8080/TCP] Resources:&ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{},} ApplicationName:testing Wait:true StorageToBeMounted:map[] StdOut:0xc00000e018}: unable to follow logs for cmp-git-testing-1: unable get build log cmp-git-testing-1: unable to wait for build cmp-git-testing-1 to run: timed out waiting for the condition
```

prietyc123:
Looks good to me. Can somebody have a look from code change prospective? 

/lgtm

PR #2552, opened 2020-01-31
Including testingutil package in test coverage report

amitkrout:
**What type of PR is this?**

/kind bug

missing testingutil test coverage report 

Fixes #? NA

make test-coverage should include ```testingutil``` pkg and it should be reflected in travis job that generates the test coverage report

### DIFF
--- a/scripts/generate-coverage.sh
+++ b/scripts/generate-coverage.sh
@@ -6,7 +6,7 @@
 set -e
 echo "" > coverage.txt
 go test -i -race ./cmd/odo
-for d in $(go list ./... | grep -v vendor | grep -v tests | grep -v testingutil); do
+for d in $(go list ./... | grep -v vendor | grep -v tests); do
     # For watch related tests, race check causes issue so disabling them here as race is already tested in other tests when used with `-coverprofile=profile.out`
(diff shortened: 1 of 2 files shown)


scripts/generate-coverage.sh
```diff
@@ -6,7 +6,7 @@
 set -e
 echo "" > coverage.txt
 go test -i -race ./cmd/odo
-for d in $(go list ./... | grep -v vendor | grep -v tests | grep -v testingutil); do
```
kadel: Isn't this going to do quite the opposite what title of this PR says?
`grep -v` does invert match. So it actually excludes `testingutil`

amitkrout: My plan is not to exclude ```testingutil```. I just remove the last grep so that pkg ```testingutil``` will be appear in the go list.

Screen Shot 2020-01-31 at 6 50 20 PM

kadel: Oh, sorry. My bad. I was reading the diff in the other way (like you added the `grep -v testingutil`).
Not I see that you are removing it 😇 

I'll blame it on Friday :-D 
Sorry

kadel:
small change
/approve
/lgtm

PR #2554, opened 2020-01-31
Update User Story Issue template

kadel:


kadel:
Using admin powers to merge this as it is small not code related change.

PR #2103, opened 2019-09-09
Remove / modifies fmt.Errorf for util.LogErrorAndExit

cdrage:
Removes / modifies fmt.Errorf for util.LogErrorAndExit

There have been cases were `odo [COMMAND] -o json` has been blank. This
is due to log.Errorf or fmt.Errorf rather than util.LogErrorAndExit

This modifies that output so errors are properly returned (using
`return`) rather than exiting with logging and then returning a "blank"
`return` statement.

PR #2564, opened 2020-02-04
Odo Push Show Error to User from Openshift

adisky:
**What type of PR is this?**

> /kind api-change
 /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind code-refactoring

This commit updates, odo push output to show exact error to user from openshift, so that user can know the cause of error and updates local config accordingly. e.g. When a user creates a storage 
with some **invalid name** ,  it does not out show user the cause of error.
```
[adisky@localhost odo]$ odo storage create storageFive --context ~/my_components/nodejs-ex/ --path /data5 --size 1Gi
 ✓  Added storage storageFive to nodejs-nodejs-ex-cqli

[adisky@localhost odo]$ odo push
 ✗  Blank source location, does the .odo directory exist?
[adisky@localhost odo]$ odo push --context ~/my_components/nodejs-ex/
Validation
 ✓  Checking component [12ms]

Configuration changes
 ✗  Retrieving component data [9ms]
 ✗  Failed To Update Config To Component Deployed
```

Fixes #?

create storage with some invalid specs, In the output you will get the exact cause of the error.

### DIFF
--- a/pkg/odo/cli/component/common_push.go
+++ b/pkg/odo/cli/component/common_push.go
@@ -82,7 +82,7 @@ func (cpo *CommonPushOptions) createCmpIfNotExistsAndApplyCmpConfig(stdout io.Wr
 	// Apply config
 	err := component.ApplyConfig(cpo.Context.Client, *cpo.localConfigInfo, stdout, cpo.doesComponentExist)
 	if err != nil {
-		odoutil.LogErrorAndExit(err, "Failed to update config to component deployed")
+		odoutil.LogErrorAndExit(err, "Failed to update config to component deployed.")
 	}
 
 	return nil
--- a/pkg/odo/util/cmdutils.go
+++ b/pkg/odo/util/cmdutils.go
@@ -49,7 +49,8 @@ func LogErrorAndExit(err error, context string, a ...interface{}) {
 			if context == "" {
 				log.Error(errors.Cause(err))
 			} else {
-				log.Errorf(strings.Title(context), a...)
+				printstring := fmt.Sprintf("%s%s", strings.Title(context), "\nError: %v")
+				log.Errorf(printstring, err)
 			}
 		}
 


pkg/odo/cli/component/common_push.go
```diff
@@ -75,14 +74,18 @@ func (cpo *CommonPushOptions) createCmpIfNotExistsAndApplyCmpConfig(stdout io.Wr
...
 	}
 
 	// Apply config
 	err := component.ApplyConfig(cpo.Context.Client, *cpo.localConfigInfo, stdout, cpo.doesComponentExist)
 	if err != nil {
-		odoutil.LogErrorAndExit(err, "Failed to update config to component deployed")
+		log.Errorf("Failed to update config to component deployed. Error: %v",
+			err,
+		)
+		os.Exit(1)
```
mik-dass: we don't use exit in this package. We have helper functions like `LogErrorAndExit` for that

adisky: @mik-dass Please check here https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/common_push.go#L73 in the same file it is being used, so i have also used it.
in `LogErrorAndExit` it is not printing error in case context given, so either i need to change it or do not pass context here.

adisky: https://github.com/openshift/odo/blob/master/pkg/odo/util/cmdutils.go#L52

mik-dass: > @mik-dass Please check here https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/common_push.go#L73 in the same file it is being used, so i have also used it.
 
Then we need to remove it from there too in a separate PR. Here's a discussion regarding that https://github.com/openshift/odo/pull/2103#issuecomment-532690129

> in LogErrorAndExit it is not printing error in case context given, so either i need to change it or do not pass context here.

I guess we can avoid the context in this case as the current one isn't clear enough.

adisky: @mik-dass Thanks for the reference, I will update the PR.

PR #2572, opened 2020-02-06
Use env variable for unit test verbose mode

amitkrout:
**What type of PR is this?**

> /kind feature

Enabling verbose mode for unit test run. Helps to debugging the unit test failure 

Fixes #? NA

UNIT_TEST_VERBOSE_MODE=-v make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,10 @@ SLOW_SPEC_THRESHOLD := 120
 # To enable verbosity export or set env GINKGO_VERBOSE_MODE like "GINKGO_VERBOSE_MODE=-v"
 GINKGO_VERBOSE_MODE ?=
 
-# Env variable UNIT_TEST_VERBOSE_MODE is used to get control over enabling go test
-# verbose mode for unit test run. By default go test verbosity is not enabled.
-# To enable verbosity export or set env UNIT_TEST_VERBOSE_MODE like "UNIT_TEST_VERBOSE_MODE=-v"
-UNIT_TEST_VERBOSE_MODE ?=
+# Env variable UNIT_TEST_ARGS is used to get control over enabling test flags
+# along with go test. To enable verbosity export or set env UNIT_TEST_ARGS like "UNIT_TEST_ARGS=-v"
+
+UNIT_TEST_ARGS ?=
 
 GINKGO_FLAGS_ALL = $(GINKGO_VERBOSE_MODE) -randomizeAllSpecs -slowSpecThreshold=$(SLOW_SPEC_THRESHOLD) -timeout $(TIMEOUT)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,11 @@ SLOW_SPEC_THRESHOLD := 120
 # To enable verbosity export or set env GINKGO_VERBOSE_MODE like "GINKGO_VERBOSE_MODE=-v"
 GINKGO_VERBOSE_MODE ?=
 
+# Env variable UNIT_TEST_VERBOSE_MODE is used to get control over enabling go test
+# verbose mode for unit test run. By default go test verbosity is not enabled.
+# To enable verbosity export or set env UNIT_TEST_VERBOSE_MODE like "UNIT_TEST_VERBOSE_MODE=-v"
+UNIT_TEST_VERBOSE_MODE ?=
```
girishramnani: it should either be `UNIT_TEST_VERBOSE_MODE=true` or `UNIT_TEST_ARGS=-v`

girishramnani:
I would suggest that we use either `UNIT_TEST_VERBOSE_MODE=true` or `UNIT_TEST_ARGS=-v` because UNIT_TEST_VERBOSE_MODE=-v feels a bit confusing

Makefile
amitkrout: @girishramnani 
> UNIT_TEST_VERBOSE_MODE=true

Looks good, but it needs few more unnecessary lines in make file to implement this. Unnecessary because we have already a ```-v``` flag which enables unit test verbose in golang. For example ```go test -v```. Here just need to use the ```-v``` flag through an env var.

> UNIT_TEST_ARGS=-v

Sounds good, but it is not specific around which args it holds. So imo ```UNIT_TEST_VERBOSE_MODE ?=``` is not a bad choice ;)

amitkrout:
> I would suggest that we use either `UNIT_TEST_VERBOSE_MODE=true` or `UNIT_TEST_ARGS=-v` because UNIT_TEST_VERBOSE_MODE=-v feels a bit confusing

@kadel @girishramnani https://github.com/openshift/odo/pull/2572/files#r377582865

Makefile
kadel: > > UNIT_TEST_ARGS=-v
> 
> Sounds good, but it is not specific around which args it holds. So imo `UNIT_TEST_VERBOSE_MODE ?=` is not a bad choice ;)

That is exactly the intention. This variable can hold any flags for `go test` not just `-v`. So something like `UNIT_TEST_ARGS` makes more sense. 
If you want to limit it to just verbose mode than you should make it `UNIT_TEST_VERBOSE_MODE=true` and add those few additional lines.


Btw `GINKGO_VERBOSE_MODE` has the same issue ;-)

girishramnani: couldn't have said it better myself :+1: @kadel

amitkrout: @kadel Awesome explanation. Review comments are addressed.

prietyc123:
Looks good to me 
/lgtm

PR #2598, opened 2020-02-12
Remove devfile push cmd

### LINKED ISSUE #2538: Remove push-devfile sub command from odo
/kind bug
The `push-devfile` sub command has been exposed temporarily for experimental purposes to support the development of devfile support in odo (https://github.com/openshift/odo/issues/2467).

This command should be removed once the feature for adding EXPERIMENTAL flag has been successfully implemented (https://github.com/openshift/odo/issues/2536)

kanchwala-yusuf:
**What type of PR is this?**
> /kind feature

This PR removes the `push-devfile` command and replaces it's behavior by adding the `--devfile` flag to the `odo push` command itself. 

Fixes #2538 

The experimental mode needs to be enabled and the `--devfile` flag needs to be explicitly set for the devfile feature to work.

### DIFF
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -28,10 +28,8 @@ var pushCmdExample = ktemplates.Examples(`  # Push source code to the current co
 %[1]s my-component --context ~/mycode
   `)
 
-const (
-	// PushRecommendedCommandName is the recommended push command name
-	PushRecommendedCommandName = "push"
-)
+// PushRecommendedCommandName is the recommended push command name
+const PushRecommendedCommandName = "push"
 
 // PushOptions encapsulates options that push command uses
 type PushOptions struct {
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -27,8 +27,13 @@ var pushCmdExample = ktemplates.Examples(`  # Push source code to the current co
 %[1]s my-component --context ~/mycode
   `)
 
-// PushRecommendedCommandName is the recommended push command name
-const PushRecommendedCommandName = "push"
+const (
+	// PushRecommendedCommandName is the recommended push command name
+	PushRecommendedCommandName = "push"
+
+	// devfile flag name
+	devfileFlagName = "devfile"
+)
 
 // PushOptions encapsulates options that push command uses
 type PushOptions struct {
@@ -36,6 +41,9 @@ type PushOptions struct {
 
 	// devfile path
 	devfilePath string
+
+	// indicates if --devfile flag has been explicitly set on cli
+	devfileFlag bool
 }
 
 // NewPushOptions returns new instance of PushOptions
@@ -49,7 +57,13 @@ func NewPushOptions() *PushOptions {
 // Complete completes push args
 func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
 
+	// if experimental mode is enabled, check if --devfile flag is explicitly set
 	if experimental.IsExperimentalModeEnabled() {
+		if cmd.Flags().Changed(devfileFlagName) {
+			po.devfileFlag = true
+		} else {
+			po.devfileFlag = false
+		}
 		return nil
 	}
 
@@ -116,7 +130,7 @@ func (po *PushOptions) Validate() (err error) {
 // Run has the logic to perform the required actions as part of command
 func (po *PushOptions) Run() (err error) {
 	// if experimental mode is enabled, use devfile push
-	if experimental.IsExperimentalModeEnabled() {
+	if experimental.IsExperimentalModeEnabled() && po.devfileFlag && po.devfilePath != "" {
 		// devfile push
 		return po.DevfilePush()
 	} else {
@@ -149,7 +163,7 @@ func NewCmdPush(name, fullName string) *cobra.Command {
 
 	// enable devfile flag if experimental mode is enabled
 	if experimental.IsExperimentalModeEnabled() {
-		pushCmd.Flags().StringVar(&po.devfilePath, "devfile", "./devfile.yaml", "Path to a devfile.yaml")
+		pushCmd.Flags().StringVar(&po.devfilePath, devfileFlagName, "", "Path to a devfile.yaml")
 	}
 
 	pushCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	"github.com/openshift/odo/pkg/odo/util/completion"
+	"github.com/openshift/odo/pkg/odo/util/experimental"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
@@ -32,6 +33,9 @@ const PushRecommendedCommandName = "push"
 // PushOptions encapsulates options that push command uses
 type PushOptions struct {
 	*CommonPushOptions
+
+	// devfile path
+	devfilePath string
 }
 
 // NewPushOptions returns new instance of PushOptions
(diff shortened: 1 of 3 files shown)


elsony:
Do we need the `--devfile` flag to decide whether we use the devfile path for `odo push`?  It means that the user will need to specify `--devfile` on every `odo push` command.  Will it be simpler if we detect the existence of the devfile in the project to switch to use the devfile code path for `odo push`?

kanchwala-yusuf:
> Do we need the `--devfile` flag to decide whether we use the devfile path for `odo push`? It means that the user will need to specify `--devfile` on every `odo push` command. Will it be simpler if we detect the existence of the devfile in the project to switch to use the devfile code path for `odo push` when EXPERIMENTAL flag is set to true?

@elsony,
While the devfile feature is in experimental mode, wouldn't it make sense to provide the odo user with flexibility to choose between the orthodox `odo push` behavior and the new devfile feature? The `--devfile` precisely does that. If we automatically detect the presence of `devfile.yaml`, even though the experimental mode is enabled, there is no way for the user to switch back to the old `odo push` behavior. The only way is to disable the experimental mode. 

We may add other experimental features which are not related to devfiles (operator hub integration for example), in such a case we end up using devfiles even for these experimental features.

Your suggestion would definitely be helpful, when the devfile feature is not in experimental mode anymore and devfile is accepted as the default standard for configuring odo.

cc: @kadel

kadel:
> > Do we need the `--devfile` flag to decide whether we use the devfile path for `odo push`? It means that the user will need to specify `--devfile` on every `odo push` command. Will it be simpler if we detect the existence of the devfile in the project to switch to use the devfile code path for `odo push` when EXPERIMENTAL flag is set to true?
> 
> @elsony,
> While the devfile feature is in experimental mode, wouldn't it make sense to provide the odo user with flexibility to choose between the orthodox `odo push` behavior and the new devfile feature? The `--devfile` precisely does that. If we automatically detect the presence of `devfile.yaml`, even though the experimental mode is enabled, there is no way for the user to switch back to the old `odo push` behavior. The only way is to disable the experimental mode.
> 
> We may add other experimental features which are not related to devfiles (operator hub integration for example), in such a case we end up using devfiles even for these experimental features.
> 
> Your suggestion would definitely be helpful, when the devfile feature is not in experimental mode anymore and devfile is accepted as the default standard for configuring odo.
> 
> cc: @kadel

I was actually also thinking that if experimental is enabled and there is devfile.yaml in the directory `odo push` should automatically pick it up.
One of the reasons why `ODO_EXPERIMENTAL` can be useful is to test odo behavior before we make it default. And I think that this is how it should work.



EXPERIMENTAL | devfile.yaml present | .odo/config.yaml present | odo push behavior | odo push --devfile devfile.yaml
-- | -- | -- | -- | --
TRUE | yes | yes | use devfile.yaml | use devfile.yaml
TRUE | yes | no | use devfile.yaml | use devfile.yaml
TRUE | no | yes | use .odo/config.yaml | use devfile.yaml
TRUE | no | no | error out (not existing component) | error out (not existing component)
FALSE | yes | yes | use .odo/config.yaml | error out (not existing flag --devfile)
FALSE | yes | no | use .odo/config.yaml | error out (not existing flag --devfile)
FALSE | no | yes | use .odo/config.yaml | error out (not existing flag --devfile)
FALSE | no | no | error out (not existing component) | error out (not existing flag --devfile)

`--devfile` flag is still needed as it will force the use of devfile even if someone names it differently. For example `ODO_EXPERIMENTAL=true odo push --devfile mydevfile.yaml`

kanchwala-yusuf:
> > > Do we need the `--devfile` flag to decide whether we use the devfile path for `odo push`? It means that the user will need to specify `--devfile` on every `odo push` command. Will it be simpler if we detect the existence of the devfile in the project to switch to use the devfile code path for `odo push` when EXPERIMENTAL flag is set to true?
> > 
> > 
> > @elsony,
> > While the devfile feature is in experimental mode, wouldn't it make sense to provide the odo user with flexibility to choose between the orthodox `odo push` behavior and the new devfile feature? The `--devfile` precisely does that. If we automatically detect the presence of `devfile.yaml`, even though the experimental mode is enabled, there is no way for the user to switch back to the old `odo push` behavior. The only way is to disable the experimental mode.
> > We may add other experimental features which are not related to devfiles (operator hub integration for example), in such a case we end up using devfiles even for these experimental features.
> > Your suggestion would definitely be helpful, when the devfile feature is not in experimental mode anymore and devfile is accepted as the default standard for configuring odo.
> > cc: @kadel
> 
> I was actually also thinking that if experimental is enabled and there is devfile.yaml in the directory `odo push` should automatically pick it up.
> One of the reasons why `ODO_EXPERIMENTAL` can be useful is to test odo behavior before we make it default. And I think that this is how it should work.
> 
> EXPERIMENTAL	devfile.yaml present	.odo/config.yaml present	odo push behavior	odo push --devfile devfile.yaml
> TRUE	yes	yes	use devfile.yaml	use devfile.yaml
> TRUE	yes	no	use devfile.yaml	use devfile.yaml
> TRUE	no	yes	use .odo/config.yaml	use devfile.yaml
> TRUE	no	no	error out (not existing component)	error out (not existing component)
> FALSE	yes	yes	use .odo/config.yaml	error out (not existing flag --devfile)
> FALSE	yes	no	use .odo/config.yaml	error out (not existing flag --devfile)
> FALSE	no	yes	use .odo/config.yaml	error out (not existing flag --devfile)
> FALSE	no	no	error out (not existing component)	error out (not existing flag --devfile)
> `--devfile` flag is still needed as it will force the use of devfile even if someone names it differently. For example `ODO_EXPERIMENTAL=true odo push --devfile mydevfile.yaml`

@kadel, 
Wouldn't it make sense to implement the above matrix when the devfile feature implementation is complete? Currently, only the devfile parser has been implemented. The complete `odo push` workflow for devfile is still under development.

We can implement the above matrix, once the devfile feature implementation is complete and then gradually move the devfile feature out of the experimental mode.

Another aspect here is that with the above matrix, wouldn't it be difficult for us to add any other feature into experimental mode, since we would be defaulting to `devfile` feature?

kanchwala-yusuf:
After having a chat with @kadel, it seems that we want the features in experimental mode to work just the way we would want them to work when these features are removed out of the experimental mode. So, I would make the necessary changes to accommodate for the suggestion given by @kadel and @elsony .

amitkrout:
> How to test changes / Special notes to the reviewer:
The experimental mode needs to be enabled and the --devfile flag needs to be explicitly set for the devfile feature to work.

@kanchwala-yusuf As a user i want to visualize the change done by this pr, so can you please elaborate it or may be the steps would really be helpful.

dharmit:
> After having a chat with @kadel, it seems that we want the features in experimental mode to work just the way we would want them to work when these features are removed out of the experimental mode. So, I would make the necessary changes to accommodate for the suggestion given by @kadel and @elsony .

Do we think this will have any impact on the work-flow of Adapters that use our tool? If yes, have we discussed the changes they need to make?

pkg/odo/cli/component/push.go
```diff
@@ -32,6 +34,9 @@ const PushRecommendedCommandName = "push"
 // PushOptions encapsulates options that push command uses
 type PushOptions struct {
 	*CommonPushOptions
+
+	// devfile path
+	devfilePath string
```
dharmit: Why not make this a part of the `CommonPushOptions` struct in `common_push.go`?

kadel:
> Do we think this will have any impact on the work-flow of Adapters that use our tool? If yes, have we discussed the changes they need to make?

Not at this point. This is one of the reasons why we want to hide it under an experimental flag, at least for now.

pkg/odo/cli/component/push.go
kanchwala-yusuf: Because the changes in this PR do not affect the common push code path in `common_push.go`. 

In my understanding, common push has code which is common to the various `odo push` code paths whereas in the case of devfile there is common code at all.

dharmit: I think `odo push` is commong across different use cases, i.e., local sourcetype or git sourcetype or binary sourcetype. It behaves according to the underlying sourcetype. Considering DevFile is going to replace odo config (`.odo/config.yaml`) at some point, IMO, it would make more sense to make `devfilePath string` a part of `CommonPushOptions`.

Having said that, I agree that it's probably OK to keep it as-is right now. I hope we don't forget/overlook when we make it a default behaviour. :wink:

Issue #2473, opened 2019-12-17
Devfile components sync support


rajivnathan:
Related to https://github.com/openshift/odo/issues/2470

- [ ] Implement source syncing to containers within a fat pod.
- [ ] Ignore files in .gitignore
- [ ] Sync delta changes instead of entire source

johnmcollier:
As discussed on the odo dev meeting today:
- We'll keep the existing watch and sync technology, it works fine as-is, and should work well with Devfile projects
- Current sync function can only sync to one container
- As Devfile projects may deploy fat / multi-container pods, we may need to sync the source code to multiple containers, so need to modify the sync function to support that
    - However, we don't want to sync to **every** container
- We'll use the `mountsources` field in the devfile to determine which containers need access to the source code, and have a single shared volume with the source that those containers would mount
    - Means we only need to sync once

johnmcollier:
Updating the plan for Devfile syncing:
- When creating the deployment for devfiles, create an emptyDir volume and mount it under `/projects`
    - `/projects` is the default mount path for projects in Che with devfiles
- If a component has `mountSources: true` set in it, we'll mount the emptyDir volume to it
- When syncing, we'll sync the source code to the first container we found that had set `mountSources: true`.
    - No need to sync across the other containers since they will be sharing the same volume.

PR #2626, opened 2020-02-24
Move file sync to its own package

johnmcollier:

**What type of PR is this?**

/kind feature
/kind code-refactoring

In order to implement file syncing for devfile components, we need to make the file sync functions that exist in occlient, available to kclient. The file sync functionality only needs a very small modification to work with devfile components (adding the option to specify a container name in `ExecCmdInContainer`). As such, we'd likely be duplicating a lot of code if we had the sync code in both occlient **and** kclient. 

So, this PR moves the file sync functionality from `occlient` into its own package, `sync`. I then created an interface (`SyncClient`) that both `occlient` and `kclient` implement, allowing the sync function to be called from both "legacy" odo (with `occlient`) and the "devfile" parts of odo (with `kclient`)

This PR does the following:
- Creates an interface, `SyncClient`, with one function: `ExecCMDInContainer` that both `occlient` and `kclient implement
- Moves `CopyFile()` (and the functions it depends on) from the `occlient` package to sync.go in the `sync` package.
    - Updates `odo push` and `odo watch` so that they call this function for file syncing
- Modifies `occlient`'s ExecCMDInContainer function to take in a container name as a parameter (required to implement `SyncClient`).
    - If a blank/no container name is passed in, it defaults to the first container.

https://github.com/openshift/odo/issues/2473 (partially)


johnmcollier:
The tests appear to be failing with the following RBAC errors from the OpenShift CI:
```
 [odo] E0224 18:55:57.750852   12628 glog.go:119] err: error while streaming command: pods "backend-app-1-ch87l" is forbidden: User "developer" cannot create resource "pods/exec" in API group "" in the namespace "vxgcyuwihe"
[odo]  ✗  Syncing files to the component [340ms]
[odo]  ✗  pods "backend-app-1-ch87l" is forbidden: User "developer" cannot create resource "pods/exec" in API group "" in the namespace "vxgcyuwihe"
```

johnmcollier:
Fixed the above issue, it was occurring when `odo push` was called and the current Kube namespace differs from the namespace/project specified in the config.yaml.

johnmcollier:
Updating the PR title to reflect changes:
- Moved file sync into its own package, `sync`, which has an interface `SyncClient` that both kclient and occlient implement.
- No longer adding `kclient` to the CLI context with "legacy" odo components
    - We can always do this later, but for now it's not needed and makes for a riskier code change

The alternative to this PR is https://github.com/openshift/odo/pull/2643

johnmcollier:
@kanchwala-yusuf Thanks! Agreed that this is the better approach. So I've closed https://github.com/openshift/odo/pull/2643

Regarding unit tests for the sync package, I had been thinking about it, but I couldn't find any existing tests for those file sync in the occlient package, and since CopyFile requires an already running pod, I didn't think it was something that could be easily unit-testable with the mock Kube-api. 

What are your thoughts?

kadel:
I agree that those functions are important for odo functionality and they should be covered. But to be completely fair to @johnmcollier this PR just moves uncovered functions to a different package. It is not adding anything new. I'm not sure if it is fair to require unit tests for functions that were just moved. 
/approve

johnmcollier:
@kadel I'll go open an issue so that we can get at least _some_ unittest coverage on the file sync functionality

kanchwala-yusuf:
Looks good to me.
/lgtm

PR #2589, opened 2020-02-11
Create deployment based on Devfile

rajivnathan:
**What type of PR is this?**
/kind feature

## What does this PR do / why we need it
Implements Deployment creation when the `push` command is used.

## List of changes
- Added convenience functions to devfile parser package to get devfile components
- Updated container generator to accept resource requirements (eg. memoryLimit)
- Implemented utility functions for converting environment variables and resource requirements from Devfile types to kube api types
- Removed service account property from pod spec for now because it is optional and we don't have a way to determine whether a non-default service account should be used at this point in time
- Created a devfile command context that can be applicable for any devfile-related commands, it currently only contains the component name
- Created a platform adapter interface that currently only defines a Start() function
- Implemented a Kubernetes platform adapter that creates a Deployment with a fat pod representing the devfile components
- Created a common platform package that contains functions common to all platforms
- Created helper functions for creating a platform adapter based on a Devfile
- Updated existing tests
- Updated component name generation to ensure the name satisfies DNS-1123 rules
- New tests
- Rebased on master and updated to work with https://github.com/openshift/odo/pull/2598

**Which issue(s) this PR fixes**:

Fixes #2471

## How to test changes / Special notes to the reviewer**:

**Automated tests**
Run the kclient tests:
In `odo/pkg/kclient` run `go test ./... -v`

Expected results:
```
=== RUN   TestCreateDeployment
=== RUN   TestCreateDeployment/Case:_Valid_deployment_name
=== RUN   TestCreateDeployment/Case:_Invalid_deployment_name
--- PASS: TestCreateDeployment (0.00s)
    --- PASS: TestCreateDeployment/Case:_Valid_deployment_name (0.00s)
    --- PASS: TestCreateDeployment/Case:_Invalid_deployment_name (0.00s)
=== RUN   TestUpdateDeployment
=== RUN   TestUpdateDeployment/Case:_Valid_deployment_name
=== RUN   TestUpdateDeployment/Case:_Invalid_deployment_name
--- PASS: TestUpdateDeployment (0.00s)
    --- PASS: TestUpdateDeployment/Case:_Valid_deployment_name (0.00s)
    --- PASS: TestUpdateDeployment/Case:_Invalid_deployment_name (0.00s)
=== RUN   TestGenerateContainer
=== RUN   TestGenerateContainer/#00
=== RUN   TestGenerateContainer/container1
--- PASS: TestGenerateContainer (0.00s)
    --- PASS: TestGenerateContainer/#00 (0.00s)
[...]

PR #2629, opened 2020-02-25
Removing odo specific checks from devfile parser

### LINKED ISSUE #2600: Validation with default Devfile's does not work.
/kind bug

<!--

Welcome! - We kindly ask you to:

  1. Fill out the issue template below 
  2. Use the Google group if you have a question rather than a bug or feature request.

The group is at: https://groups.google.com/forum/#!forum/odo-users

Thanks for understanding, and for contributing to the project!

-->

Tried using: https://github.com/redhat-developer/devfile/blob/master/getting-started/go/devfile.yaml

against master `odo`.

## Actual behavior

```
odo push-devfile
~/foobar                                                                                                                                                                                                                                                                                                                                                                    
▶ odo push-devfile            
 ✗  invalid devfile schema. errors :
- components.0: Must validate "then" as "if" was valid
- components.0: Must validate "then" as "if" was valid
- components.0.endpoints.0.attributes.public: Invalid type. Expected: string, given: boolean
- components.0: Must validate all the schemas (allOf)
- components.2: Must validate "then" as "if" was valid
- components.2: Must validate "then" as "if" was valid
- components.2.id: Does not match pattern '[a-z0-9_\-.]+/[a-z0-9_\-.]+/[a-z0-9_\-.]+$'
- components.2: Must validate all the schemas (allOf)
- components.3: Must validate "then" as "if" was valid
- components.3: Must validate "then" as "if" was valid
- components.3.id: Does not match pattern '[a-z0-9_\-.]+/[a-z0-9_\-.]+/[a-z0-9_\-.]+$'
- components.3: Must validate all the schemas (allOf)
- commands.1.actions.0: Must validate one and only one schema (oneOf)
- commands.1.actions.0: command is required
- commands.1.actions.0: Additional property timeout is not allowed
- commands.1.actions: Array must have at most 1 items
```


## Expected behavior

Parser working correctly and THEN erroring out with no `build` or `push` available.


See above

@kanchwala-yusuf

### LINKED ISSUE #2655: Projects field in devfiles isn't optional for odo
/kind bug

odo built from master branch

1. Ran `odo preference set experimental true`
2. Downloaded this [sample devfile](https://raw.githubusercontent.com/johnmcollier/devfiles/master/odo/liberty/devfile.yaml)
3. Ran `odo push` and was presented with the following error:
```
johns-mbp-3:liberty-ex johncollier$ odo push
 ✗  no projects present in devfile
```

Since the devfile spec allows the `projects` field to be optional (and since odo needs to support projects locally on disk), odo should support devfiles that don't have any projects under `projects` in a devfile.

## Actual behavior
Odo push fails:
```
johns-mbp-3:liberty-ex johncollier$ odo push
 ✗  no projects present in devfile
```

## Expected behavior
Odo push succeeds without a project:
```
johns-mbp-3:nodejs-ex johncollier$ odo push 
 •  Push devfile component nodejs-ex  
...
```

johns-mbp-3:nodejs-ex johncollier$ odo push
 ✗  no projects present in devfile

If I add a sample project to the devfile, everything runs fine.

kanchwala-yusuf:
**What type of PR is this?**
> /kind bug

This PR removes the odo  specific validations from the devfile parser. The odo specific validations would be the responsibility of the calling function. The reason for this change is that devfile parser needs to be independent of odo, it should work for any valid devfile. 

Fixes #2600 
Fixes #2655

Devfile validations for any valid devfile should work fine.


kanchwala-yusuf:
@rajivnathan , I think that the changes in this PR may affect your PR #2589 . Can you please check?

pkg/devfile/parse.go
```diff
@@ -39,12 +39,6 @@ func Parse(path string) (d DevfileObj, err error) {
 		return d, errors.Wrapf(err, "failed to decode devfile content")
 	}
 
-	// Validate devfile data
-	err = d.Data.Validate()
```
rajivnathan: Which packages are considered independent of odo? Is it `pkg/devfile/parser` and `pkg/devfile/versions` only or were you thinking more? If it's only those two then can we move the existing validation files to a new package like `pkg/devfile/validation` instead of removing them? I'm thinking for odo we'd want the `pkg/devfile/Parse` function to still validate here but just call the validation from an odo package.

Perhaps we should consider moving the parser packages to a top-level folder in pkg as well? Maybe `pkg/parser` or `pkg/devfile-parser` so it will make it easier to separate it out later?

Issue #2602, opened 2020-02-13
Issues with localConfigInfo in common_push.go


cdrage:
/kind bug

## How did you run odo exactly?

So... for some reason, `LocalConfigInfo` is imported twice into `component/create.go` here:

https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/create.go#L48

Which then imports it twice in `common_push.go`...

https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/common_push.go#L31

```go
type CommonPushOptions struct {
	ignores []string
	show    bool

	sourceType       config.SrcType
	sourcePath       string
	componentContext string
	LocalConfigInfo  *config.LocalConfigInfo

	pushConfig         bool
	pushSource         bool
	forceBuild         bool
	doesComponentExist bool

	*genericclioptions.Context
}
```

Because in `*genericclioptions.Context` we also import LocalConfigInfo...

https://github.com/openshift/odo/blob/master/pkg/odo/genericclioptions/context.go#L314

Causing a conflict when implementing: https://github.com/openshift/odo/pull/2532

Because LocalConfigInfo ends up being blank..

This needs to be fixed in order to no longer have any conflicting LocalConfigInfo's / null pointers.

## Actual behavior

Incorrect LocalConfigInfo being returned

## Expected behavior

Correct LocalConfigInfo being returned.

We should also refactor the code that's being used in common_push.go as it conflicts with code that's also in `component/describe.go` which uses a different way to retrieve the local context / LocalConfigInfo.

## Any logs, error output, etc?

Running https://github.com/openshift/odo/pull/2532 pulls in incorrect LocalConfgInfo.

cdrage:
Ping @mik-dass  and @girishramnani I believe you two are the best suited for the task. I tried some simple debugging but found I would fix one bug, cause 99 others haha

cdrage:
Setting priority as high as this is affecting us adding new features (`odo create -o json`) and this has already as well as can cause more issues in the future.

cdrage:
ping @mik-dass @girishramnani

girishramnani:
and it seems this is happening at multiple places -

- Describe uses ComponentOptions which has context ( which has localconfiginfo ) https://github.com/openshift/odo/blob/de1aa2bcb44e8f1106d49454952c718b814e9fe4/pkg/odo/cli/component/describe.go#L31
[...]

PR #2631, opened 2020-02-25
removed duplicate LocalConfigInfo and redundent file reads

girishramnani:
**What type of PR is this?**

 /kind cleanup
/kind code-refactoring

We have `LocalConfigInfo` as part of `Context` but we still are initialising it multiple times and even at places have duplicate declarations. So removing those.

Fixes https://github.com/openshift/odo/issues/2602

Tests should pass and @cdrage you should probably be unblocked.

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -269,7 +269,9 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		co.interactive = true
 	}
 
-	co.LocalConfigInfo, err = config.NewLocalConfigInfo(co.componentContext)
+	// this populates the LocalConfigInfo as well
+	co.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
+
 	if err != nil {
 		return errors.Wrap(err, "failed intiating local config")
 	}
--- a/pkg/odo/cli/component/watch.go
+++ b/pkg/odo/cli/component/watch.go
@@ -42,7 +42,6 @@ type WatchOptions struct {
 	sourcePath       string
 	componentContext string
 	client           *occlient.Client
-	localConfig      *config.LocalConfigInfo
 
 	*genericclioptions.Context
 }
@@ -54,16 +53,13 @@ func NewWatchOptions() *WatchOptions {
 
 // Complete completes watch args
 func (wo *WatchOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	// Set the correct context
+	wo.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
+
 	wo.client = genericclioptions.Client(cmd)
 
-	// Retrieve configuration
-	conf, err := config.NewLocalConfigInfo(wo.componentContext)
-	if err != nil {
-		return errors.Wrap(err, "unable to retrieve configuration information")
-	}
-
 	// Set the necessary values within WatchOptions
-	wo.localConfig = conf
+	conf := wo.Context.LocalConfigInfo
 	wo.sourceType = conf.LocalConfig.GetSourceType()
 
 	// Get SourceLocation here...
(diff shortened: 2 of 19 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -269,7 +269,9 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		co.interactive = true
 	}
 
-	co.LocalConfigInfo, err = config.NewLocalConfigInfo(co.componentContext)
+	// this populates the LocalConfigInfo as well
+	co.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
```
girishramnani: Point to think: we have two ways to init context, one which creates the app i.e. `NewContextCreatingAppIfNeeded`  other that doesn't `NewContext`. Now I am creating the app if needed.
Now some would argue that - post new workflow, all cluster interactions should happen on `odo push` only and I agree to a certain degree. But here an API call moves to the `create` portion which would remove that 250ms from push and its barely noticeable for create. So felt like we can keep this optimisation?

girishramnani:
This PR kind of shows that we have forgotten that there is a `LocalConfigInfo` in the `Context` struct :laughing:

girishramnani:
added @cdrage to validate that he is unblocked and others to review.

The original reviewers assigned to this PR were on PTO or part of IBM ( who are working on kclient and might not have time review this PR )

pkg/odo/cli/component/create.go
dharmit: IMO, it's an optimization if we're reducing from 250ms to something less. But by moving this from `odo push` to `odo create`, we're simply doing this in a different place.

IMO, if it's agreed (TBH, I'm not aware of the new workflow you've mentioned) to keep cluster interactions to `odo push`, then moving this `odo create` is disobeying it for no apparent reason (is 250ms so noticeable in `odo push`?).  I'm guessing there's a context behind taking a decision to put certain things at certain place. 

This is just a personal opinion. I'm sure others would have their own. :wink:

pkg/odo/cli/component/watch.go
```diff
@@ -54,16 +53,13 @@ func NewWatchOptions() *WatchOptions {
 
 // Complete completes watch args
 func (wo *WatchOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	wo.client = genericclioptions.Client(cmd)
+	// Set the correct context
+	wo.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
```
dharmit: A bit lost here. Why do we need to create a context in `odo watch`? Isn't `odo watch` supposed to work if and only if user is working from a valid component directory or has specified CLI parameters that help deduce a valid context?

pkg/odo/genericclioptions/context.go
```diff
@@ -30,6 +30,26 @@ func NewContextCreatingAppIfNeeded(command *cobra.Command) *Context {
 	return newContext(command, true, false)
 }
 
+// NewConfigContext is a special kind of context which only contains local configuration, other information is not retrived
+//  from the cluster. This is useful for commands which don't want to connect to cluster.
+func NewConfigContext(command *cobra.Command) *Context {
```
dharmit: Should we name this `NewLocalConfigContext` or `NewLocalConfig` since it's specifically about local configuration?

pkg/odo/cli/component/watch.go
girishramnani: context contains the `LocalConfigInfo` hence every command that requires a valid component dir would need context.

pkg/odo/genericclioptions/context.go
girishramnani: NewLocalConfigContext sounds better

pkg/odo/cli/component/create.go
girishramnani: resolved

pkg/odo/genericclioptions/context.go
girishramnani: resolved

girishramnani:
all comments resolved, please review

PR #2645, opened 2020-02-27
Fix odo catalog describe service command to not require odo config file

### LINKED ISSUE #2605: odo catalog commands without existing component
# User Story
- As a user, I want to list items from `odo catalog` without creating a component first.

## Acceptance Criteria
- `odo catalog *` commands should not require component (localConfig) to list, describe or search services and component types
- if any of `odo catalog *`  commands need project, the current project from `KUBECONFIG` should be used.

/kind user-story

Feature request: https://github.com/openshift/odo/issues/2111

kanchwala-yusuf:
**What type of PR is this?**
> /kind bug

Currently, the odo configuration file is required for describing catalog services. This is not the desired behavior. odo configuration file should not be necessary for describing catalog services. This PR fixes that issue.

Fixes #2605 

Execution of `odo catalog describe service <serivce-name>` command should work without requiring the `.odo/config.yml` file.

### DIFF
--- a/pkg/odo/cli/catalog/describe/service.go
+++ b/pkg/odo/cli/catalog/describe/service.go
@@ -41,7 +41,7 @@ func NewDescribeServiceOptions() *DescribeServiceOptions {
 
 // Complete completes DescribeServiceOptions after they've been created
 func (o *DescribeServiceOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	o.Context = genericclioptions.NewContext(cmd)
+	o.Context = genericclioptions.NewContext(cmd, true)
 	o.serviceName = args[0]
 
 	return
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -20,8 +20,12 @@ import (
 const DefaultAppName = "app"
 
 // NewContext creates a new Context struct populated with the current state based on flags specified for the provided command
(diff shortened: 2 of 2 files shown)


adisky:
@kanchwala-yusuf Thanks for the PR!! It works well,  after checking your fix i was wondering how other commands are ignoring local config, as `ignoreMissingConfiguration` is being not used by any other commands, i found the logic for other commands here 
https://github.com/openshift/odo/blob/master/pkg/odo/genericclioptions/context.go#L129
I am not sure which way is correct, but i am fine with your changes as well, infact it seems to be more cleaner approach.
Looks good to me!!

kanchwala-yusuf:
> @kanchwala-yusuf Thanks for the PR!! It works well, after checking your fix i was wondering how other commands are ignoring local config, as ignoreMissingConfiguration is being not used by any other commands, i found the logic for other commands here
> https://github.com/openshift/odo/blob/master/pkg/odo/genericclioptions/context.go#L129
> I am not sure which way is correct, but i am fine with your changes as well, infact it seems to be more cleaner approach.
> Looks good to me!!

@adisky , You are actually right about the way other commands handle it. Just thought of a different approach, I thought it is much simpler if `ignoreMissingConfiguration` is passed than figuring it out on the fly.

May be we can have others review it as well and give an opinion. I am open to suggestions and reviews.

PR #2649, opened 2020-02-27
run validate with unit tests on OpenShift CI

kadel:
Currently, the validations are run only on travis-ci. 
The problem is that Travis-CI status is not checked by merge-bot, so the PR can be merged even if the tests are failing.
This PR makes sure that the validations are also run on Prow, so we don't merge code that, for example not correctly formated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,14 +16,11 @@ jobs:
       go: "1.13.1"
       install:
         - make goget-tools
-        # binary will be $(go env GOPATH)/bin/golangci-lint
-        - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
       script:
         - export PATH="$PATH:$GOPATH/bin"
         - make bin
         - make validate
         - make test-coverage
-        - make golint
       after_success:
         # submit coverage.txt to codecov.io
         - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -7,9 +7,16 @@ set -x
 
 export ARTIFACTS_DIR="/tmp/artifacts"
 export CUSTOM_HOMEDIR=$ARTIFACTS_DIR
+export PATH=$PATH:$GOPATH/bin
+# set localtion for golangci-lint cache
```
mik-dass: location

scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -7,9 +7,16 @@ set -x
...
+# otherwise /.cache is used, and it fails on permission denied
+export GOLANGCI_LINT_CACHE="/tmp/.cache"
 
+make goget-tools
+make validate
+make golint
```
mik-dass: According to https://github.com/openshift/odo/pull/2649/files#diff-b67911656ef5d18c4ae36cb6741b7965R61, won't make validate run make golint too?

.travis.yml
```diff
@@ -16,14 +16,11 @@ jobs:
       go: "1.13.1"
       install:
         - make goget-tools
-        # binary will be $(go env GOPATH)/bin/golangci-lint
-        - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
       script:
         - export PATH="$PATH:$GOPATH/bin"
         - make bin
         - make validate
```
amitkrout: Remove ```make validate``` too.

kadel: I would actually keep it here. It is fairly quick. And there is no harm in running it also here

scripts/openshiftci-presubmit-unittests.sh
kadel: good catch. 
Fixed.

Issue #2577, opened 2020-02-07
Add ability to switch context between local and kube (odo context)


johnmcollier:
## User Story
As a user, I want to be able to switch contexts between using odo for Kubernetes development with devfiles, and local development with devfiles (with Docker as the container runtime). An `odo context <context>` command can enable this, where `<context>` is either `local` or `kube`.

## Acceptance Criteria 
- [ ] It should allow users to easily switch between contexts when developing with devfile projects
- [ ] Users shouldn't have to re-run `odo create ...` after switching contexts
- [ ] It should be optional. If it isn't run, the default context will be Kube/OpenShift.
- [ ] It should add a field to the config.yaml to note the context: `context: <context>`

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

/kind user-story

gorkem:
What do we expect the context switch to do? Asking because I always imagined the local support would be a flag to certain commands such as `push` or `watch`

johnmcollier:
Our line of thinking was that it would enable users to activate support for "local" development without having to add `--local` to all of their commands. The context switch would just add some metadata to the odo config.yaml. Alternatively, if we wanted to keep `odo context` as an optional command for local support, we could also add support for `--local` flags as well (if `--local` is used, it would override the current/default context)

Longer-term, we were also thinking `odo context` could be used as a way to quickly switch between Kubernetes contexts for multiple Kube clusters with odo (like https://github.com/ahmetb/kubectx).

elsony:
The other point regarding to adding `odo context` command is that when user starts to use the local scenario, most of the commands that they call will be all executing on the local.  It will pretty rare that the user will call one command on Kube and another command on local and doing constant context switch.  Having the `odo context` command will save the user from having to pass the `--local` flag to all the commands that they are calling.

girishramnani:
- we already have `context` being used for component directory all through out odo commands. So I would suggest  `target` parameter for this feature.
- we have `sourceType` that can be set to `git`, `binary` or `local` so please consider something else for local development. maybe consider `--docker` or `--oci` or `--container`? 
- we can avoid implementing `odo context` by providing `--target` flag to `odo create` which would add `Target: <<local or kube>>` in the config.yaml. We wouldn't need to provide this flag to all commands as they can pick this information from the local config.
[...]

PR #2656, opened 2020-02-27
Implement PushTarget preference

johnmcollier:
**What type of PR is this?**

/kind feature

This PR implements the `PushTarget` setting as described in https://github.com/openshift/odo/issues/2577.

The push target for odo can be set in two ways:
   - `odo preference set pushtarget <target>`, which will add the following to the preference.yaml file:
      ```
      OdoSettings:
        PushTarget: docker
      ```
      or
       ```
      OdoSettings:
        PushTarget: kube
      ```
   - Setting the `ODO_PUSHTARGET=<target>` environment variable, which will override the value that's set in the preference.yaml file
    

As this functionality is currently a no-op (pending the implementation of odo push for docker), I've left it out of any of the descriptions in `odo preference` or `odo preference set/unset` to avoid confusing users. 

This PR adds the following:
- New `PushTarget` preference
   - Valid values are `kube` and `docker`.
   - PreferenceInfo.GetPushTarget() returns `kube` if no field is set in the preferences
- New `pushtarget` package in `pkg/odo/util` to help determining if the push target is set to docker (via `IsPushTargetDocker()`)
   - Checks the environment variable first, and if the environment variable is not set, it will check the preference.yaml field for `PushTarget`
- Unit tests
   - Unit tests for the new preference `PushTarget`
      - Tests both valid and invalid values
   - Unit tests for the new `pushtarget` package
      - Verifies that `IsPushTargetDocker` returns the proper values based on the values of `ODO_PUSHTARGET` and the `PushTarget` preference

Fixes https://github.com/openshift/odo/issues/2577

1. Build my branch and build odo
2. Run `make test` to verify the tests I added in the `preferences` and `pushtarget` packages pass
3. Run `odo preference set pushtarget <target>` to set the push target in the preferences.yaml file (valid values are docker and kube). Invalid values should fail.
4. Run `odo preference unset pushtarget` to verify that the setting gets removed from preference.yaml


pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,45 @@
...
+	// and will override the setting in the preferences file.
+
+	OdoPushTarget = "ODO_PUSHTARGET"
+)
+
+// IsPushTargetDocker checks if the push target preference has been set to docker
+func IsPushTargetDocker() bool {
+
+	// ODO's push target can be told to use docker by:
+	//		- setting an env variable "ODO_PUSHTARGET=docker" or
```
mik-dass: if the user has set both but to opposite values, the env variable will be the first choice?

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/preference"
+)
+
+// env variables
+const (
+	// Setting this env to `docker` will enable pushing to docker containers
+	// and will override the setting in the preferences file.
+
+	OdoPushTarget = "ODO_PUSHTARGET"
```
mik-dass: maybe `ODO_PUSH_TARGET`?

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,45 @@
...
+	// Check "ODO_PUSHTARGET" env variable.
+	pushTargetEnv, _ = os.LookupEnv(OdoPushTarget)
+	if pushTargetEnv == "docker" || pushTargetEnv == "kube" {
+		return pushTargetEnv == "docker"
+	}
+
+	// Fetch odo preferences and check if pushtarget is set
+	cfg, err := preference.New()
+	if err != nil {
+		log.Errorf("failed to read odo preferences config. err: '%v'\n", err)
```
mik-dass: maybe use glog here and also return false after logging. Then we won't need the else block and the variable pushTargetPreference.

pkg/odo/util/pushtarget/pushtarget_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/openshift/odo/pkg/preference"
+)
+
+func TestIsDockerPushTargetEnabled(t *testing.T) {
```
mik-dass: Please change the name to the corresponding function it tests `PushTargetDocker`.

pkg/odo/util/pushtarget/pushtarget_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	const (
+		pushTargetSetting string = "PushTarget"
+	)
+
+	// temp config file
+	tempConfigFile, err := ioutil.TempFile("", "odoconfig")
```
mik-dass: maybe use the fake filesystem here?

pkg/odo/util/pushtarget/pushtarget_test.go
```diff
@@ -0,0 +1,114 @@
...
+			setConfig: "kube",
+			want:      true,
+		},
+		{
+			name:      "override pushtarget prefrence with kube pushtarget env",
+			setEnv:    "kube",
+			setConfig: "docker",
+			want:      false,
+		},
+	}
```
mik-dass: please add a case where the env is neither `kube` nor `docker`

pkg/odo/util/pushtarget/pushtarget_test.go
```diff
@@ -0,0 +1,114 @@
...
+			if tt.setEnv != "" {
+				err := os.Setenv(OdoPushTarget, tt.setEnv)
+				if err != nil {
+					t.Errorf("failed to set env %s. err: '%v'", OdoPushTarget, err)
+				}
+				defer os.Unsetenv(OdoPushTarget)
+			}
+
+			// create new preference file
+			cfg, err := preference.NewPreferenceInfo()
```
mik-dass: maybe use the fake filesystem here too?

pkg/preference/preference.go
```diff
@@ -189,7 +199,7 @@ func (c *PreferenceInfo) SetConfiguration(parameter string, value string) error
 				return errors.Wrapf(err, "unable to set %s to %s", parameter, value)
 			}
 			if typedval < 0 {
-				return errors.Errorf("cannot set timeout to less than 0")
```
mik-dass: why was this line removed?

pkg/preference/preference.go
```diff
@@ -294,6 +310,16 @@ func (c *PreferenceInfo) GetExperimental() bool {
 	return *c.OdoSettings.Experimental
 }
 
+// GetPushTarget returns the value of PushTarget from config
```
mik-dass: `from preference`

pkg/odo/util/pushtarget/pushtarget_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// test table
+	tests := []struct {
+		name      string
+		setEnv    string
+		setConfig string
+		want      bool
+	}{
+		{
+			name:      "no pushtarget setting set",
```
mik-dass: maybe add some case numbering?

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,45 @@
...
+// env variables
+const (
+	// Setting this env to `docker` will enable pushing to docker containers
+	// and will override the setting in the preferences file.
+
+	OdoPushTarget = "ODO_PUSHTARGET"
+)
+
+// IsPushTargetDocker checks if the push target preference has been set to docker
+func IsPushTargetDocker() bool {
```
mik-dass: Maybe pass along a filesystem. That way we can use the filesystem in our tests and speed it up.

mik-dass:
Works well :+1:  Just added some suggestions and fixes.

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,45 @@
...
+	//		- setting odo preference using "odo preference set pushtarget docker"
+
+	var (
+		pushTargetEnv        string
+		pushTargetPreference string
+	)
+
+	// Check "ODO_PUSHTARGET" env variable.
+	pushTargetEnv, _ = os.LookupEnv(OdoPushTarget)
+	if pushTargetEnv == "docker" || pushTargetEnv == "kube" {
```
kadel: If I set `ODO_PUSHTARGET=foo" than it will get silently ignored and value will be taken from the preference file. 
Wouldn't it be better to fail if there is a nonsense value?

pkg/preference/preference.go
kadel: even linter is complaining ;-)
```
golangci-lint run ./... --timeout 5m
pkg/preference/preference.go:201:4: SA9003: empty branch (staticcheck)
			if typedval < 0 {
			^
make: *** [golint] Error 1
```

mik-dass:
Also @johnmcollier Please add some integration tests

pkg/odo/util/pushtarget/pushtarget.go
johnmcollier: @mik-dass How would the filesystem be passed in? `preference.New()` (which `IsPushTargetDocker` calls) doesn't take in any args, so the preferences package would need to be refactored quite a bit to support `filesystem.Filesystem`

pkg/odo/util/pushtarget/pushtarget.go
johnmcollier: Yeah, my thinking was to have the environment var override what was set in the preferences file (perhaps the user wants to use `ODO_PUSH_TARGET=docker` for a one off `odo push`, without wanting to change the global preferences.

pkg/preference/preference.go
johnmcollier: Yikes, that's embarrassing, I'm not sure how I missed that or how it happened. But I've fixed it either way.

pkg/preference/preference.go
johnmcollier: Good catch! I've updated the other functions from preferences that were referencing config as well.

pkg/odo/util/pushtarget/pushtarget_test.go
johnmcollier: Good idea, done!

pkg/odo/util/pushtarget/pushtarget.go
johnmcollier: That seems like a reasonable approach 👍 , I've updated the code to log an error and erturb false if the environment var has a nonsense value (and included a new test to catch that)

johnmcollier:
Thanks for reviewing @mik-dass @kadel! I've updated the PR with your requested changes and integration tests.

The following integration tests were added:
- Set pushtarget to docker
- Set pushtarget to Kube
- Unset pushtarget
- Setting pushtarget to invalid value

pkg/odo/util/pushtarget/pushtarget.go
johnmcollier: Good idea, updated.

pkg/odo/util/pushtarget/pushtarget_test.go
johnmcollier: Done, updated!

pkg/odo/util/pushtarget/pushtarget_test.go
johnmcollier: Updated.

pkg/odo/util/pushtarget/pushtarget.go
johnmcollier: Good idea, I used glog for both of the errors, and made sure to return false in both cases.

pkg/odo/util/pushtarget/pushtarget.go
mik-dass: Yeah you're right. Sorry about this, I didn't realize we don't have support of fake filesystems for preferences.

pkg/odo/util/pushtarget/pushtarget.go
mik-dass: > Wouldn't it be better to fail if there is a nonsense value?

Or maybe continue and look into the preference file instead of failing?

johnmcollier:
More flakiness, different test fails each time 😢 

johnmcollier:

8th time's the charm?

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	// Fetch odo preferences and check if pushtarget is set
+	cfg, err := preference.New()
+	if err != nil {
+		glog.Errorf("failed to read odo preferences config. err: '%v'\n", err)
+		return false
+	}
+
+	return cfg.GetPushTarget() == "docker"
```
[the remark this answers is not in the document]
johnmcollier: I would say constants are fine, but I don't think we need a custom type for them

pkg/odo/util/pushtarget/pushtarget.go
```diff
@@ -0,0 +1,47 @@
...
+// env variables
+const (
+	// Setting this env to `docker` will enable pushing to docker containers
+	// and will override the setting in the preferences file.
+
+	OdoPushTarget = "ODO_PUSH_TARGET"
+)
+
+// IsPushTargetDocker checks if the push target preference has been set to docker
+func IsPushTargetDocker() bool {
```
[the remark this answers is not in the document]
johnmcollier: IMO, we should keep it as is, and add other functions to the pushtarget package as needed later.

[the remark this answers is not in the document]
rajivnathan: Ok, I was more just thinking about it from an adapter perspective. This way you'd call IsPushTargetDocker() to decide between docker/kube adapter. With getPushTarget() you'd be able to get it and then switch based on the type which I think is cleaner but you're right doesn't matter much right now.

PR #2500, opened 2020-01-08
Add devfile parser

kanchwala-yusuf:
**What type of PR is this?**
> /kind feature

**What does does this PR do / why we need it**:
This PR implements the basic components devfile parser.

Which issue(s) this PR fixes**:

Fixes #2477

PR #2667, opened 2020-03-03
Fake filesystem changes to devfile parser

### LINKED ISSUE #2553: PR #2500 Followup
- [x] https://github.com/openshift/odo/pull/2500#discussion_r372890591 let's use the fake fs package here and write the stuff in memory instead of the disk. That will speed up the testing.
- [x] https://github.com/openshift/odo/pull/2500#discussion_r372891177 why not use the Fs Filesystem which is part of the DevfileCtx? What will help us with unit tests using the fake system.
- [x] https://github.com/openshift/odo/pull/2500#discussion_r372891839 We can pass the fake filesystem here in DevfileCtx
- [ ]  https://github.com/openshift/odo/pull/2500#discussion_r372895747 since the description says run command not present, should this be build app?
- [x] https://github.com/openshift/odo/pull/2500#discussion_r372906855 can we convert these tests to the tabular format we currently use in odo?

/assign @kanchwala-yusuf 
/kind devfile
/label state/Ready

kanchwala-yusuf:
**What type of PR is this?**
> /kind feature

Accommodating review comments from #2500 

Fixes #2553 

Devfile related actions should work without any errors.

### DIFF
--- a/pkg/devfile/parser/content_test.go
+++ b/pkg/devfile/parser/content_test.go
@@ -1,9 +1,10 @@
 package parser
 
 import (
-	"io/ioutil"
 	"os"
 	"testing"
+
+	"github.com/openshift/odo/pkg/testingutil/filesystem"
 )
 
 func TestSetDevfileContent(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/devfile/parser/content_test.go
```diff
@@ -39,8 +40,12 @@ func TestSetDevfileContent(t *testing.T) {
 	t.Run("valid file", func(t *testing.T) {
 
 		var (
-			tempDevfile = createTempDevfile(t, validJsonRawContent100())
-			d           = DevfileCtx{absPath: tempDevfile.Name()}
+			fakeFs      = filesystem.NewFakeFs()
```
mik-dass: shouldn't this test be converted to table format?

girishramnani: +1 on this

PR #2674, opened 2020-03-04
Removes the annotation 'app.openshift.io/vcs-uri' for local and binary components

### LINKED ISSUE #2175: Invalid URL annotation generated with binary component
[kind/bug]

- Operating System: Win10
- Output of `odo version`:

```
odo v1.0.0-beta5 (4d1d5071)     
                                
Server: https://172.18.84.5:8443
Kubernetes: v1.11.0+d4cacc0
```     

`odo create java:8 bincomp --binary target\springboot-rest2-1.0.0-SNAPSHOT.jar --project bin --app binapp`

## Actual behavior

The generated annotation is:

`app.openshift.io/vcs-uri: 'file:///target/springboot-rest2-1.0.0-SNAPSHOT.jar'`
## Expected behavior

Should be:

`app.openshift.io/vcs-uri: 'file:target/springboot-rest2-1.0.0-SNAPSHOT.jar'`

That would allow tooling (VSCode, IntelliJ) to use this annotation to import a component from the deployment.

mik-dass:
**What type of PR is this?**

/kind cleanup

Removes the annotation 'app.openshift.io/vcs-uri' for local and binary components. It also removes the 'SOURCE' column from the 'odo list' command.

Fixes #2175 

- create a local component and check that the above mentioned annotation is not present in the DC.
- create a git component and check that the above mentioned annotation is present in the DC.
- check the source code and check if the annotation is used for only git components.
- run `odo list` with combination of all the flags (not with json flag) and check that the `SOURCE` column is not present.
- run the above steps with the json flag and check if the source is displayed only for the git components.

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -1020,7 +1020,7 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
 	}
 
 	var sourcePath string
-	if sourceType == "git" {
+	if sourceType == string(config.GIT) {
 		sourcePath = deploymentConfig.ObjectMeta.Annotations[componentSourceURLAnnotation]
 	}
 
@@ -1073,7 +1073,7 @@ func Update(client *occlient.Client, componentConfig config.LocalConfigInfo, new
 
 	// Create annotations
 	annotations := make(map[string]string)
-	if newSourceType == "git" {
+	if newSourceType == config.GIT {
 		annotations[componentSourceURLAnnotation] = newSource
 	}
 	annotations[ComponentSourceTypeAnnotation] = string(newSourceType)
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -979,7 +979,6 @@ func ListIfPathGiven(client *occlient.Client, paths []string) (ComponentList, er
 				a := getMachineReadableFormat(data.GetName(), data.GetType())
 				a.Namespace = data.GetProject()
 				a.Spec.App = data.GetApplication()
-				a.Spec.Source = data.GetSourceLocation()
 				a.Spec.Ports = data.GetPorts()
 				a.Status.Context = con
 				state := "Not Pushed"
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -30,7 +30,6 @@ import (
 )
 
 // componentSourceURLAnnotation is an source url from which component was build
-// it can be also file://
 const componentSourceURLAnnotation = "app.openshift.io/vcs-uri"
 const ComponentSourceTypeAnnotation = "app.kubernetes.io/component-source-type"
 const componentRandomNamePartsMaxLen = 12
@@ -223,8 +222,7 @@ func CreateFromPath(client *occlient.Client, params occlient.CreateArgs) error {
 	labels[componentlabels.ComponentTypeVersion] = imageTag
 
 	// save source path as annotation
-	sourceURL := util.GenFileURL(params.SourcePath)
-	annotations := map[string]string{componentSourceURLAnnotation: sourceURL}
+	annotations := map[string]string{}
 	annotations[ComponentSourceTypeAnnotation] = string(params.SourceType)
 
 	// Namespace the component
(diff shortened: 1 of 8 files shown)


pkg/component/component.go
```diff
@@ -1015,13 +1013,17 @@ func GetComponentSource(client *occlient.Client, componentName string, applicati
...
 
-	sourcePath := deploymentConfig.ObjectMeta.Annotations[componentSourceURLAnnotation]
 	sourceType := deploymentConfig.ObjectMeta.Annotations[ComponentSourceTypeAnnotation]
 
 	if !validateSourceType(sourceType) {
 		return "", "", fmt.Errorf("unsupported component source type %s", sourceType)
 	}
 
+	var sourcePath string
+	if sourceType == "git" {
```
adisky: could we use `config.GIT`  here?

pkg/component/component.go
```diff
@@ -1070,7 +1072,10 @@ func Update(client *occlient.Client, componentConfig config.LocalConfigInfo, new
 	}
 
 	// Create annotations
-	annotations := map[string]string{componentSourceURLAnnotation: newSource}
+	annotations := make(map[string]string)
+	if newSourceType == "git" {
```
adisky: ditto

pkg/component/component.go
mik-dass: fixed

pkg/component/component.go
mik-dass: fixed

pkg/odo/cli/component/list.go
```diff
@@ -70,9 +70,9 @@ func (lo *ListOptions) Run() (err error) {
 			machineoutput.OutputSuccess(components)
 		} else {
 			w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(w, "APP", "\t", "NAME", "\t", "PROJECT", "\t", "TYPE", "\t", "SOURCE", "\t", "STATE", "\t", "CONTEXT")
```
girishramnani: why do we need to remove the source from display?

girishramnani:
The scope of the issue doesn't describe removing the `source` column from `odo component list`, can't we still fetch that information and show it?

pkg/odo/cli/component/list.go
mik-dass: Because otherwise the source would be empty for local and binary components for commands such as `odo list --path <some_path>`.

mik-dass:
> The scope of the issue doesn't describe removing the source column from odo component list, can't we still fetch that information and show it?

For commands such as `odo list --path <some_path>` we already have a context field for displaying the path of the component directory. And for commands such as `odo list --app <app_name> --project <project_name>`, the path fetched will be empty for local and binary components. Also since the information might be fetched from the online cluster, it might be irrelevant for the current user. So we decided to remove it completely as displaying the column for just git components might be confusing for the user.

kadel:
for `odo list -o json` is still showing `"source": "file://./"`
+1 for keeping the "source" field there, but shouldn't we just show  ``"source": ""` for local components?

mik-dass:
>  but for the local and binary component it should probably be empty (``"source": ""). Showing always "file://./"` can be quite misleading.

But the problem is we use omitempty for json output in component type. So currently it is getting removed from the json output.

girishramnani:
this change removes some fields from the json output, we should inform adapter team for this. also added `release-note` tag so we don't forget to add this in the change logs.

cdrage:
Added blocked label, please check with the adapter team that this does not break anything @mik-dass thanks!

mik-dass:
> Added blocked label, please check with the adapter team that this does not break anything @mik-dass thanks!

@cdrage @girishramnani Already asked by @kadel on the issue https://github.com/openshift/odo/issues/2175#issuecomment-578648924. But we didn't manage to get any response from them.

amitkrout:
@mik-dass Test changes looks good to me. Thanks

PR #2678, opened 2020-03-04
Add machine output documentation

cdrage:
**What type of PR is this?**

> /kind documentation

Adds a table and references to machine output documentation

N/A

View the doc

### DIFF
--- a/docs/dev/machine-output.adoc
+++ b/docs/dev/machine-output.adoc
@@ -1,10 +1,173 @@
-# Machine Readable Output
+= Machine readable output
 
 This document outlines all the machine readable output options and examples.
 
-`odo app describe -o json | jq`
- 
-```json
+See the below table for a list of all possible machine readable output commands:
+
+|===
+| Command | Subcommand | Implemented?
+
+| application
+| delete
+| N
+
+|
+| <<application-describe,describe>>
+| Y
+
+|
+| <<application-list,list>>
+| Y
+
+| catalog
+| describe
+| N
+
+|
+| search
+| N
+
+|
+| <<catalog-list-components,list components>>
+| Y
+
+|
+| <<catalog-list-services,list services>>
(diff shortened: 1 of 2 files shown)


docs/dev/machine-output.adoc
```diff
@@ -1,10 +1,173 @@
...
+| catalog
+| describe
+| N
+
+|
+| search
+| N
+
+|
+| <<catalog-list-components,list components>>
```
Preeticp: This link and the ones on lines 35, 39, 111, 115, 131, 139 do not work. Most probably cause they are not  listed below, are we planning to add those sections here? Is that why we have these links?

docs/dev/machine-output.adoc
```diff
@@ -133,12 +268,12 @@ This document outlines all the machine readable output options and examples.
     }
   ]
 }
-```
+----
 
+== component describe
```
Preeticp: Component describe and Component list are described twice.

docs/dev/machine-output.adoc
```diff
@@ -1,10 +1,173 @@
-# Machine Readable Output
+= Machine readable output
 
 This document outlines all the machine readable output options and examples.
 
-`odo app describe -o json | jq`
- 
-```json
+See the below table for a list of all possible machine readable output commands:
```
Preeticp: @cdrage, iiuc, only those which have machine readable output implemented have the links? Do you think it makes sense to have the links to all of those which have sections outlined below?

cdrage: Yup. Only the ones implemented have the links.

docs/dev/machine-output.adoc
cdrage: Yes, that's intentional, just wanted to see if this format is fine :+1:

Issue #2673, opened 2020-03-03
Create integration test suites for devfile support


johnmcollier:
Currently there's no integration test suites for the devfile support in odo. While this isn't much of an issue right now, when there isn't much in terms of devfile support (at this moment, just creating a deployment), as we add more features and as the devfile-mode becomes more complete, this will become a bigger and bigger issue. Additionally, adding integration tests for devfile support will require more work the longer that we delay it (as more and more functionality needs to be tested).

So, we should create integration test suites for odo's devfile support sooner rather than later.

- [ ] Create new test suites under `tests/integration` for the devfile/experimental features in odo
    - Need to add some sample devfiles and projects under `tests/examples`
- [ ] Get the integration tests running in Travis
    - Update the makefile with the new test suites 
    - Update the .travis.yml to run those tests
- [ ] Get the integration tests running in OpenShift CI
    - @kadel What needs to be done for this step? For Travis it's pretty clear, but I'm not sure how/where to tell the OpenShift CI which tests to run.

/area testing
/area devfile

johnmcollier:
I took a crack at writing integration test suites for devfiles (mostly just `odo push` and syncing) and hit https://github.com/openshift/odo/issues/2684 as a blocker.

johnmcollier:
My sync PR does the following:
- Adds a `—namespace` flag to `odo push`, meaning #2684 is no longer a blocker for integration tests (though we should still fix #2684)
- Adds integration tests for devfile push and syncing.

Once my sync PR has been delivered, this issue can be closed out.

johnmcollier:
This can be closed out - looks like my PR didn't close it when it was merged

Issue #2661, opened 2020-03-02
odo push doesn't wait for running pod


kadel:
/kind bug
/area devfile
/priority medium

`odo push` should wait for the Pod to be in running state before exits.
Currently, it does that only for first invocation (when resources are created). But any additional `odo push` doesn't wait.

```
export ODO_EXPERIMENTAL=true

▶ odo push ; kubectl get pods
 •  Push devfile component odo-examples  ...
 ✓  Waiting for component to start [9s]
 ✓  Push devfile component odo-examples [10s]
NAME                           READY   STATUS    RESTARTS   AGE
odo-examples-8cb546c7b-sbmxd   1/1     Running   0          11s

# modify devfile.yaml

▶ odo push ; kubectl get pods
 •  Push devfile component odo-examples  ...
 ✓  Waiting for component to start [174ms]
 ✓  Push devfile component odo-examples [1s]
NAME                           READY   STATUS        RESTARTS   AGE
odo-examples-8cb546c7b-sbmxd   1/1     Terminating   0          71s
```

Also if edits to devfile.yaml caused Pod to get into CrashLoop state users doesn't know about it

rajivnathan:
Spoke to John about this, he mentioned his sync PR may resolve this so I'll try to reproduce this once https://github.com/openshift/odo/pull/2681 is in to see if it's still an issue.

rajivnathan:
@kadel has found a nice solution to this problem. He has a PR incoming so I'll unassign myself.

Issue #2652, opened 2020-02-27
Add unit tests for file sync code


johnmcollier:
While implementing https://github.com/openshift/odo/pull/2626, I saw that our file sync code (`CopyFile()` and the functions it depends on) didn't have any unit test coverage and only our integration tests were providing test coverage.

As this is such a critical piece of code for one of odo's core functionalities, we should add unit tests to it and ensure that there is proper test coverage

- [ ] Figure out if we can test `sync.CopyFile` with the mock Kube API
    - This might be tricky as it requires an already running pod to exec to, and I'm not sure if the test/mock Kube API can handle that
- [ ] Add unit tests for `checkFileExist`, `makeTar` and `recursiveTar` functions
    - These should be relatively straightforward to test and don't depend on the Kube API.

/area testing

PR #2681, opened 2020-03-05
File sync for devfiles

### LINKED ISSUE #2473: Devfile components sync support

johnmcollier:
**What type of PR is this?**

/kind feature

This PR implements file syncing for devfile components, using the kubernetes client library. The push code is based on the existing push/sync code used in classic odo, with only a few small tweaks required.

Overall this PR does the following:
- It updates the Devfile deployment code to create a shared emptyDir volume (called `odo-projects`) and mount it under `/projects` in all of the devfile components that set `mountSources: true`
    - This is in line with what Che does for mounting project source
- Creates a `Push()` function in the devfile adapters to handle file syncing
    - If syncing to the pod for the first time (or if `--force` is set), it will sync all of files in the source directory using the `sync` package
    - If syncing to an already running component:
       - it syncs only the files that changed (by checking odo-file-index.json)
       - If any files were deleted from the source directory, it deletes from the remote pod.
- Creates a `DoesComponentExist()` function in the devfile adapters
    - Returns true if a component with the specified label already exists
    - Wrapper for `utils.ComponentExists`
- Moved `IsEmpty` into the `pkg/utils` package from `pkg/component`
    - Needed for both odo "classic" and "devfile" components, so needed to move it to a neutral location.

- Tests
    - Unit tests for new functions
         - Can't really unit test Push(), as it can't be mocked with the fake Kube client (e.g. requires exec'ing to a running pod)
    - Integration tests to cover push and sync scenarios

Fixes #2473, https://github.com/openshift/odo/issues/2673

1. Build my branch and run `make`
2. Run `make test` to verify unit tests pass
3. Enable experimental mode
4. Run `odo push` verify that the deployment starts and files get synced
5. Change/add/remove files, verify the proper changes get reflected on the container.


johnmcollier:
Updated to include:
`-namespace` flag on odo push
    - Unrelated to file syncing, but was a requirement to implement integration tests
- integration tests 🎉 
    - tests to verify both odo push and syncing works
    - Adds two simple devfiles, both nodejs. One single container, one multi container
    - Just a few simple tests right now as there isn’t a complete end to end scenario.

.travis.yml
```diff
@@ -98,6 +98,7 @@ jobs:
         - travis_wait make test-cmd-app
         - travis_wait make test-cmd-push
         - travis_wait make test-cmd-project
+        - travis_wait make test-cmd-devfile-push
```
amitkrout: +1,
FYI - Travis job runs test on 3.11 cluster

Makefile
```diff
@@ -190,6 +190,11 @@ test-cmd-pref-config:
 test-cmd-push:
 	ginkgo $(GINKGO_FLAGS) -focus="odo push command tests" tests/integration/
 
+# Run odo push command tests
```
amitkrout: ```Run odo push command tests``` -> ```Run odo push devfile command tests```

.travis.yml
amitkrout: Update the [name:](https://github.com/openshift/odo/pull/2681/files#diff-354f30a63fb0907d4ad57269548329e3L90) field accordingly. For example 
```
name: "watch, storage, app, project, push and devfile push command integration tests"
```

tests/integration/cmd_devfile_push_test.go
```diff
@@ -0,0 +1,156 @@
...
+	oc = helper.NewOcRunner("oc")
+
+	// This is run after every Spec (It)
+	var _ = BeforeEach(func() {
+		SetDefaultEventuallyTimeout(10 * time.Minute)
+		project = helper.CreateRandProject()
+		context = helper.CreateNewContext()
+		currentWorkingDirectory = helper.Getwd()
+		helper.Chdir(context)
+		os.Setenv("ODO_EXPERIMENTAL", "true")
```
amitkrout: Do not use it like this because odo test spec (It) runs on two ginkgo test node in parallel, so it may cause race condition. To avoid such situation we have a env variable which can control preference path.

You can use 
```
```
and in the spec i mean in the It block you can update the preference value. For example
```
```

tests/integration/cmd_devfile_push_test.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+
+	// Clean up after the test
+	// This is run after every Spec (It)
+	var _ = AfterEach(func() {
+		helper.DeleteProject(project)
+		helper.DeleteDir(context)
+		helper.Chdir(currentWorkingDirectory)
+		os.Unsetenv("ODO_EXPERIMENTAL")
```
amitkrout: As per https://github.com/openshift/odo/pull/2681/files#r388746348 use ```os.Unsetenv("GLOBALODOCONFIG")``` instead.

tests/integration/cmd_devfile_push_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+		It("Check that odo push works with a devfile", func() {
+			project = helper.CreateRandProject()
+
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs"), context)
+
+			output := helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--namespace", project)
+			Expect(output).To(ContainSubstring("✓  Waiting for component to start"))
+			Expect(output).To(ContainSubstring("✓  Push devfile component"))
+			Expect(output).To(ContainSubstring("✓  Changes successfully pushed to component"))
```
amitkrout: One assertion that confirms the devfile push is successful would be enough. So  ```Expect(output).To(ContainSubstring("Changes successfully pushed to component"))``` would be the proper fit imo. Do not to pass character like ```✓``` in assert as it may behave differently on terminal like PowerShell.

tests/integration/cmd_devfile_push_test.go
```diff
@@ -0,0 +1,156 @@
...
+		currentWorkingDirectory = helper.Getwd()
+		helper.Chdir(context)
+		os.Setenv("ODO_EXPERIMENTAL", "true")
+	})
+
+	// Clean up after the test
+	// This is run after every Spec (It)
+	var _ = AfterEach(func() {
+		helper.DeleteProject(project)
+		helper.DeleteDir(context)
```
amitkrout: It will throw error on windows platform. Rearranging the line with the below one will fix the problem

```
helper.Chdir(currentWorkingDirectory)
helper.DeleteDir(context)
```

tests/integration/cmd_devfile_push_test.go
```diff
@@ -0,0 +1,156 @@
...
+		It("should not build when no changes are detected in the directory and build when a file change is detected", func() {
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs"), context)
+
+			helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--namespace", project)
+			output := helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--namespace", project)
+
+			Expect(output).To(ContainSubstring("No file changes detected, skipping build"))
+			helper.ReplaceString(filepath.Join(context, "server.js"), "node listening on", "UPDATED!")
+
+			helper.CmdShouldPass("odo", "push", "--context", context+"/nodejs-ex", "--namespace", project)
```
amitkrout: Use ```filepath.Join(context, "nodejs-ex")``` instead of ```context+"/nodejs-ex"```.

tests/integration/cmd_devfile_push_test.go
johnmcollier: I was wondering about the tests running in parallel and whether or not that would cause problems, good to know! I'll update my PR to include that.

.travis.yml
johnmcollier: Good catch, updated!

Makefile
johnmcollier: Thanks, updated!

tests/integration/cmd_devfile_push_test.go
johnmcollier: Ah, right good point.

tests/integration/cmd_devfile_push_test.go
johnmcollier: Ahh, that's a good point about Windows. Updated.

tests/integration/cmd_devfile_push_test.go
johnmcollier: Updated, good catch!

johnmcollier:
@amitkrout Thanks for reviewing! I've addressed your review comments in a new commit.

pkg/devfile/adapters/common/interface.go
```diff
@@ -1,6 +1,10 @@
 package common
 
+import "io"
+
 // ComponentAdapter defines the functions that platform-specific adapters must implement
 type ComponentAdapter interface {
 	Start() error
+	Push(path string, out io.Writer, files []string, delFiles []string, isForcePush bool, globExps []string, show bool) error
```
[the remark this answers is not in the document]
rajivnathan: Actually, do we need both Start() and Push()? Why can't one function handle both? ie. start if the component doesn't already exist and then sync.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -62,11 +68,91 @@ func (a Adapter) Start() (err error) {
...
+}
+
+// DoesComponentExist returns true if a component with the specified name exists, false otherwise
+func (a Adapter) DoesComponentExist(cmpName string) bool {
+	return utils.ComponentExists(a.Client, cmpName)
+}
+
+// getFirstContainerWithSourceVolume returns the first container that set mountSources: true
+// If no container was found, that means there's no container to sync to, so return an error
+func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, error) {
```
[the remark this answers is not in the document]
johnmcollier: Yup, sounds good to me

pkg/devfile/adapters/common/interface.go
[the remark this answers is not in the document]
johnmcollier: We could merge the two, but my thinking for keeping them separate was to keep things separate and to avoid having one big monolithic start function

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -62,11 +68,91 @@ func (a Adapter) Start() (err error) {
...
+		for _, vol := range c.VolumeMounts {
+			if vol.Name == kclient.OdoSourceVolume {
+				return c.Name, nil
+			}
+		}
+	}
 
-	_, err = a.Client.WaitAndGetPod(watchOptions, corev1.PodRunning, "Waiting for component to start")
-	return err
+	return "", fmt.Errorf("No containers specified mountSources: true")
```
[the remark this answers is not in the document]
johnmcollier: Makes sense to me.

pkg/devfile/adapters/helper.go
```diff
@@ -17,14 +17,32 @@ func NewPlatformAdapter(componentName string, devObj devfile.DevfileObj) (Platfo
 
 	// Only the kubernetes adapter is implemented at the moment
 	// When there are others this function should be updated to retrieve the correct adapter for the desired platform target
-	return createKubernetesAdapter(adapterContext)
+	return createKubernetesAdapter(adapterContext, "")
 }
 
-func createKubernetesAdapter(adapterContext common.AdapterContext) (PlatformAdapter, error) {
+// NewPlatformAdapter returns a Devfile adapter for the targeted platform
+func NewPlatformAdapterWithNamespace(componentName string, devObj devfile.DevfileObj, namespace string) (PlatformAdapter, error) {
```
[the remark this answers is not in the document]
johnmcollier: Ooh, that's a better approach, I like it. I'll go take a look.

johnmcollier:
@rajivnathan Thanks for the review! I've addressed your review comments. I've also moved the `FilterIgnores` function into the `utils` package (and written unit tests for it!), as the refactoring of `Push()` into `Start()` required it.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -28,7 +36,100 @@ type Adapter struct {
 }
 
 // Start updates the component if a matching component exists or creates one if it doesn't exist
-func (a Adapter) Start() (err error) {
+// Once the component has started, it will sync the source code to it.
+func (a Adapter) Start(path string, out io.Writer, ignoredFiles []string, forceBuild bool, globExps []string, show bool) (err error) {
```
rajivnathan: `out` and `show` don't appear to be used, can we remove them?

pkg/devfile/adapters/interface.go
```diff
@@ -1,5 +1,7 @@
 package adapters
 
+import "io"
+
 type PlatformAdapter interface {
-	Start() error
+	Start(path string, out io.Writer, ignoredFiles []string, forceBuild bool, globExps []string, show bool) error
```
rajivnathan: I think we discussed renaming `Start` to `Push`. Should we still go ahead with that?

pkg/devfile/adapters/kubernetes/component/adapter.go
johnmcollier: They are used, in the call to `pushLocal`

pkg/devfile/adapters/interface.go
johnmcollier: Done

pkg/devfile/adapters/kubernetes/component/adapter.go
johnmcollier: Done

kadel:
@johnmcollier 
All existing devfiles are expecting that the source code will be in `/projects/<project-name>`.
I've noticed that you are pushing files to `/projects/`. 

This will break every existing devfile there is :-(
For example, neither of this will work:
 - https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/java-web-spring/devfile.yaml
 - https://github.com/elsony/devfile-registry/blob/master/devfiles/springBoot/devfile.yaml

I know that there is no good solution for us, as we can't work with multiple projects in devfile right now.

What if we do this:
- If there is exactly one project defined in devfile, odo pushes code to `/projects/<project-name>/` 
- In all other cases (no project, or multiple projects) it pushes the source code directly to `/projects`.

This will make most of the devfiles usable with odo.

rajivnathan:
@kadel I agree we need to figure out where source should be synced in a way that works for most devfiles. I've just finished trying out John's latest changes and things work pretty well so maybe we can open a separate issue to handle deciding where files should be synced? We can then go through the use cases in a bit more detail and updating the code to change where to sync should be pretty simple.

johnmcollier:
@kadel @rajivnathan The changes to support pushing to `/projects/<projectName>` when exactly one project is defined was pretty straightforward, so I've pushed up a commit that does the following:
- Before syncing the source code, odo checks how many projects were defined in the devfile
     - If exactly one, it sets the "sync folder" to `/projects/<projectName>`, ensures `/projects/<projectName>` exists in the container and pushes the files to that folder
     - If either no projects are defined, or multiple projects are defined, odo syncs to `/projects`

As Rajiv notes, we'll probably need to do some more more work later to properly iron out support for multiple projects, but this is a fine start at the very least.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -246,6 +248,22 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 	s := log.Spinner("Syncing files to the component")
 	defer s.End(false)
 
+	// If there's only one project defined in the devfile, sync to `/projects/project-name`, otherwise sync to /projects
+	projects := a.Devfile.Data.GetProjects()
+	fmt.Println(len(projects))
```
rajivnathan: Looks like an errant println slipped through

johnmcollier: That was an errant debugging line that slipped through, removed.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -117,11 +211,105 @@ func (a Adapter) Create() (err error) {
 		return err
 	}
 
-	podSelector := fmt.Sprintf("component=%s", componentName)
+	return nil
+}
+
+// Push syncs source code from the user's disk to the component
+func (a Adapter) pushLocal(path string, files []string, delFiles []string, isForcePush bool, globExps []string) error {
```
girishramnani: this function is doing alot of things, lets try to extract certain portions to make unit testing better. 
two suggestions would be be -
1 - https://github.com/openshift/odo/pull/2681/files#diff-825a1a88d75e789899c58e89a79eb767R254-R264 can be extracted into a function
2 - https://github.com/openshift/odo/pull/2681/files#diff-825a1a88d75e789899c58e89a79eb767R268-R277 can be extracted into a function

Also if the function has more then 5 args please try to create a struct to pass on data or consider moving it as part of the method's struct if it doesn't change through out the lifecycle of the struct. 
e.g. path and ignoredFiles fields might not change throughout the life of `Adapter` so maybe it can be places as part of the Adapter's struct? - this is just an example.

johnmcollier: @girishramnani I'm not sure moving those sections into separate functions would make unit testing easier - they both involve exec'ing into an already running container and manipulating the files that exist there, and that can't be mocked with the fake kube client.

girishramnani: we can mock it without the fake kube client and it will definitely make unit testing pushLocal easier because when you would write test for pushLocal you would mock the https://github.com/openshift/odo/pull/2681/files#diff-825a1a88d75e789899c58e89a79eb767R254-R264 and https://github.com/openshift/odo/pull/2681/files#diff-825a1a88d75e789899c58e89a79eb767R268-R277 functions.

johnmcollier: I think what makes it different is that we're interacting with the kube client different than  we do for other resources, we're interacting directly with the rest client, not calling an existing `exec` function:
```
	req := c.KubeClient.CoreV1().RESTClient().
		Post().
		Namespace(c.Namespace).
		Resource("pods").
		Name(podName).
		SubResource("exec").
		VersionedParams(&podExecOptions, scheme.ParameterCodec)
```
and
```
	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
	if err != nil {
		return errors.Wrapf(err, "unable execute command via SPDY")
	}
	// initialize the transport of the standard shell streams
	err = exec.Stream(remotecommand.StreamOptions{
		Stdin:  stdin,
		Stdout: stdout,
		Stderr: stderr,
		Tty:    tty,
	})
```

(see https://github.com/openshift/odo/blob/master/pkg/kclient/pods.go#L93)

johnmcollier: As a compromise, rather than putting those sections with the exec into separate functions, what about moving everything _before_ the exec into their own functions (maybe `getCommandToCreateFolder` and `getCommandToDelete`)? Then we could still write unit

(Note when moving `ExecCMDInContainer` over from occlient, I couldn't find any references in the existing odo code that had unit tests with it)

girishramnani: > (Note when moving `ExecCMDInContainer` over from occlient, I couldn't find any references in the existing odo code that had unit tests for it, or unit tests for functions that called it, so other parts of odo have had this problem too.)

and that is the reason that we should not do this mistake when we have the opportunity to re-write/replace this functionality

girishramnani: > As a compromise, rather than putting those sections with the exec into separate functions, what about moving everything _before_ the exec into their own functions (maybe `getCommandToCreateFolder` and `getCommandToDelete`)? Then we could still write unit tests.

I keep that upto you, I just wanted to break this function down into smaller pieces to make it more unit testable

johnmcollier: @girishramnani I'll go with approach for now (and see if there are any other pieces we can break up). I'll also address your comment regarding the number of parameters.

Since unit tests for the `ExecCmdInContainer` function will require more work / intevestigation / design, I'll address it as part of https://github.com/openshift/odo/issues/2652 (which is also dealing with unit-testing exec)

kadel:
```
▶ odo push
 •  Push devfile component spring-petclinic  ...
 ✓  Checking file changes for pushing [12ms]
 ✓  Waiting for component to start [121ms]
 ✓  Syncing files to the component [2s]
 ✓  Push devfile component spring-petclinic [3s]

▶ odo push
 •  Push devfile component spring-petclinic  ...
 ✓  Checking file changes for pushing [8ms]
 ✓  Waiting for component to start [114ms]
 ✗  Syncing files to the component [30s]
 ✗  Failed to start component with name spring-petclinic.
Error: Failed to create the component: Failed to sync to component with name spring-petclinic: error while streaming command: Internal error occurred: error executing command in container: container is not created or running
```

For the second push, I changed the environment variable in devfile.
I guess it's because of this https://github.com/openshift/odo/issues/2661

johnmcollier:
@kadel Yup, looks like it

kadel:
> @kadel Yup, looks like it, since changing the environment variable would have restarted the deployment.

Push command needs to detect this and wait, till the pod is ready again. 
This can be addressed in a separate PR, but we need to fix it.

johnmcollier:
Separated out pushLocal into 3 more functions and added unit tests for them. Going to refactor ExecCMDInContainer in a separate PR to make it:
- Take fewer params (probably a struct of some kind in the sync package). Will also update pushlocal and push to take fewer params too.
- Make it more unit-testable.

I'll refactor both ExecCMDInContainer in kclient and occlient (as the syncclient interface requires clients to implement this function)

elsony:
Confirmed with @johnmcollier this sync support do incremental sync to only sync the changed file.

PR #2694, opened 2020-03-10
Add support for url describe command

jichenjc:
**What type of PR is this?**

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake
> /kind code-refactoring

implemented this: I certainly can do further change on the output format, just submit 
this for some comments

```
[root@xyimgr77 frontend]# ../odo url list
Found the following URLs for component ubi-frontend-gmhz in application app:
NAME                       STATE          URL     PORT
ubi-frontend-gmhz-8080     Not Pushed     ://     8080

[root@xyimgr77 frontend]# ../odo url describe
 ✗  url  not found
[root@xyimgr77 frontend]# ../odo url describe abc
 ✗  url abc not found
[root@xyimgr77 frontend]# ../odo url describe ubi-frontend-gmhz-8080
NAME                       STATE          URL     PORT
ubi-frontend-gmhz-8080     Not Pushed     ://     8080

[root@xyimgr77 frontend]#

```

Fixes #?

### DIFF
--- a/pkg/odo/cli/url/describe.go
+++ b/pkg/odo/cli/url/describe.go
@@ -1,0 +1,107 @@
+package url
+
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/openshift/odo/pkg/config"
+	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/machineoutput"
+	"github.com/openshift/odo/pkg/odo/genericclioptions"
+	"github.com/openshift/odo/pkg/odo/util"
+	"github.com/openshift/odo/pkg/odo/util/completion"
+	"github.com/openshift/odo/pkg/url"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	ktemplates "k8s.io/kubernetes/pkg/kubectl/util/templates"
+)
+
+// DescribeRecommendedCommandName is the recommended describe command name
+const describeRecommendedCommandName = "describe"
+
+var describeExample = ktemplates.Examples(`  # Describe a URL
+%[1]s myurl
+`)
+
+// URLListOptions encapsulates the options for the odo url list command
+type URLDescribeOptions struct {
+	localConfigInfo  *config.LocalConfigInfo
+	componentContext string
+	url              string
+	*genericclioptions.Context
+}
+
+// NewURLDescribeOptions creates a new URLCreateOptions instance
+func NewURLDescribeOptions() *URLDescribeOptions {
+	return &URLDescribeOptions{&config.LocalConfigInfo{}, "", "", &genericclioptions.Context{}}
+}
+
+// Complete completes URLDescribeOptions after they've been Listed
+func (o *URLDescribeOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.Context = genericclioptions.NewContext(cmd)
+	o.localConfigInfo, err = config.NewLocalConfigInfo(o.componentContext)
+	if err != nil {
+		return errors.Wrap(err, "failed initialize local config")
+	}
+	o.url = args[0]
+	return
+}
+
+// Validate validates the URLDescribeOptions based on completed values
+func (o *URLDescribeOptions) Validate() (err error) {
+	return util.CheckOutputFlag(o.OutputFlag)
+}
+
+// Run contains the logic for the odo url list command
+func (o *URLDescribeOptions) Run() (err error) {
+
+	u, err := url.Get(o.Client, o.localConfigInfo, o.url, o.Application)
+	if err != nil {
+		return err
+	}
+
+	if log.IsJSON() {
+		machineoutput.OutputSuccess(u)
+	} else {
+
+		tabWriterURL := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
+		fmt.Fprintln(tabWriterURL, "NAME", "\t", "STATE", "\t", "URL", "\t", "PORT")
+
+		// are there changes between local and cluster states?
+		outOfSync := false
+		fmt.Fprintln(tabWriterURL, u.Name, "\t", u.Status.State, "\t", url.GetURLString(u.Spec.Protocol, u.Spec.Host), "\t", u.Spec.Port)
+		if u.Status.State != url.StateTypePushed {
+			outOfSync = true
+		}
+		tabWriterURL.Flush()
+		if outOfSync {
+			fmt.Fprintf(os.Stdout, "\n")
+			fmt.Fprintf(os.Stdout, "There are local changes. Please run 'odo push'.\n")
+		}
+	}
+
+	return
+}
+
+// NewCmdURLDescribe implements the odo url describe command.
+func NewCmdURLDescribe(name, fullName string) *cobra.Command {
+	o := NewURLDescribeOptions()
+	urlDescribeCmd := &cobra.Command{
+		Use:         name + " [url name]",
+		Short:       "Describe a URL",
+		Long:        `Describe a URL`,
+		Example:     fmt.Sprintf(describeExample, fullName),
+		Args:        cobra.ExactArgs(1),
+		Annotations: map[string]string{"machineoutput": "json", "command": "url"},
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(o, cmd, args)
+		},
+	}
+	urlDescribeCmd.SetUsageTemplate(util.CmdUsageTemplate)
+	genericclioptions.AddContextFlag(urlDescribeCmd, &o.componentContext)
+	completion.RegisterCommandHandler(urlDescribeCmd, completion.URLCompletionHandler)
+	completion.RegisterCommandFlagHandler(urlDescribeCmd, "context", completion.FileCompletionHandler)
+
+	return urlDescribeCmd
+}
--- a/pkg/url/url.go
+++ b/pkg/url/url.go
@@ -19,7 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// Get returns URL defination for given URL name
+// Get returns URL definition for given URL name
 func (urls URLList) Get(urlName string) URL {
 	for _, url := range urls.Items {
 		if url.Name == urlName {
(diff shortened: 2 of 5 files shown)


mik-dass:
@jichenjc Thanks for the PR. Please edit the description and link the issue with this PR. Also please add some steps for testing the PR.

pkg/url/url.go
```diff
@@ -20,13 +20,13 @@ import (
 )
 
 // Get returns URL defination for given URL name
```
mik-dass: can you please change this to definition

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,112 @@
...
+func (o *URLDescribeOptions) Run() (err error) {
+
+	urls, err := url.List(o.Client, o.localConfigInfo, o.Component(), o.Application)
+	if err != nil {
+		return err
+	}
+	u, err := urls.Get(o.url)
+	if err != nil {
+		return err
+	}
```
mik-dass: Since `url describe` can take only one URL name as a parameter input, I don't think we need url.List() here. We have faced slowness because of listing when we didn't need it. Please write a generic Get function for URL and use the client to make a get call for a route.

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,112 @@
...
+func NewURLDescribeOptions() *URLDescribeOptions {
+	return &URLDescribeOptions{&config.LocalConfigInfo{}, "", "", &genericclioptions.Context{}}
+}
+
+// Complete completes URLDescribeOptions after they've been Listed
+func (o *URLDescribeOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.Context = genericclioptions.NewContext(cmd)
+	o.localConfigInfo, err = config.NewLocalConfigInfo(o.componentContext)
+	if err != nil {
+		return errors.Wrap(err, "failed intiating local config")
```
mik-dass: `initiating`

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,112 @@
...
+		Example:     fmt.Sprintf(describeExample, fullName),
+		Args:        cobra.RangeArgs(0, 1),
+		Annotations: map[string]string{"machineoutput": "json", "command": "url"},
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(o, cmd, args)
+		},
+	}
+	urlDescribeCmd.SetUsageTemplate(util.CmdUsageTemplate)
+	genericclioptions.AddContextFlag(urlDescribeCmd, &o.componentContext)
+	completion.RegisterCommandFlagHandler(urlDescribeCmd, "context", completion.FileCompletionHandler)
```
mik-dass: also add a command handler like we have in `url delete`

pkg/odo/util/cmdutils.go
```diff
@@ -136,7 +136,7 @@ func PrintComponentInfo(client *occlient.Client, currentComponentName string, co
 
 			// Gather the output
 			for _, componentURL := range componentDesc.Spec.URL {
-				url := urls.Get(componentURL)
+				url, _ := urls.Get(componentURL)
```
mik-dass: please don't ignore the error. If the function returns a error, this might cause a panic as the returned URL will be empty.

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,112 @@
...
+
+// NewCmdURLDescribe implements the odo url describe command.
+func NewCmdURLDescribe(name, fullName string) *cobra.Command {
+	o := NewURLDescribeOptions()
+	urlDescribeCmd := &cobra.Command{
+		Use:         name,
+		Short:       "Describe a URL",
+		Long:        `Describe a URL`,
+		Example:     fmt.Sprintf(describeExample, fullName),
+		Args:        cobra.RangeArgs(0, 1),
```
mik-dass: since `url describe` needs a parameter, this should be `cobra.ExactArgs(1)`

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Complete completes URLDescribeOptions after they've been Listed
+func (o *URLDescribeOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.Context = genericclioptions.NewContext(cmd)
+	o.localConfigInfo, err = config.NewLocalConfigInfo(o.componentContext)
+	if err != nil {
+		return errors.Wrap(err, "failed intiating local config")
+	}
+	if len(args) == 1 {
```
mik-dass: if we implement https://github.com/openshift/odo/pull/2694/files#r391425980, we won't need this check

mik-dass:
Also please add some integration test for this feature

pkg/odo/cli/url/describe.go
jichenjc: will update, thx

pkg/odo/cli/url/describe.go
jichenjc: ok, will do

pkg/odo/cli/url/describe.go
jichenjc: got it , will double check how to do this

pkg/odo/cli/url/describe.go
jichenjc: will check that

pkg/url/url.go
jichenjc: ok

pkg/odo/util/cmdutils.go
jichenjc: got it , thanks

pkg/odo/cli/url/describe.go
jichenjc: @mik-dass 
um.. after double check , seems we have something like 
clusterURL := getMachineReadableFormat(r)
which means we can't search in etcd directly, we need get back the data then do some transform and 
this of course need list ,

I can change the function to Get() the compare in the new function but still need list
so don't know how much improvement we get can... so, still a O(n) search , not O(1)

mik-dass: No you don't need a list. First write getRoute() function in occlient.go like https://github.com/openshift/odo/blob/master/pkg/occlient/occlient.go#L484 which returns the found route. Then write a Get() (don't use the existing one, it requires a list) function, call the getRoute() in occlient.go and transform the returned route to machine readable format using getMachineReadableFormat().

jichenjc: ok, get your point, I will try the way you suggested, thanks!

pkg/url/url.go
```diff
@@ -30,6 +30,45 @@ func (urls URLList) Get(urlName string) URL {
...
 }
 
+// Get returns URL definition for given URL name
+func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName string, applicationName string) (URL, error) {
+	remoteUrlName, err := util.NamespaceOpenShiftObject(urlName, applicationName)
+	if err != nil {
+		return URL{}, errors.Wrapf(err, "unable to create namespaced name")
+	}
+
+	// Check whethter remote already created the route
```
mik-dass: whether

pkg/url/url.go
```diff
@@ -30,6 +30,45 @@ func (urls URLList) Get(urlName string) URL {
...
+			if remoteExist {
+				localURL.Status.State = StateTypePushed
+			} else {
+				localURL.Status.State = StateTypeNotPushed
+			}
+			return localURL, nil
+		}
+	}
+
+	if remoteExist {
```
mik-dass: also check that the error value from above is nil otherwise the below code will crash

pkg/url/url.go
```diff
@@ -30,6 +30,45 @@ func (urls URLList) Get(urlName string) URL {
...
+		// Remote exist, but not in local, so it's deleted status
+		clusterURL := getMachineReadableFormat(*route)
+		clusterURL.Status.State = StateTypeLocallyDeleted
+		return clusterURL, nil
+	}
+
+	// can't find the in local and remote
+	return URL{}, errors.New("provided url not exist")
```
mik-dass: does not exist

tests/integration/cmd_url_test.go
```diff
@@ -88,6 +88,21 @@ var _ = Describe("odo url command tests", func() {
 		})
 	})
 
+	Context("Describing urls", func() {
+		It("should describe appropriate URLs and push message", func() {
```
mik-dass: please add some more cases with the various pushed states or maybe even add them here itself

pkg/url/url.go
```diff
@@ -30,6 +30,45 @@ func (urls URLList) Get(urlName string) URL {
...
+	}
+
+	if remoteExist {
+		// Remote exist, but not in local, so it's deleted status
+		clusterURL := getMachineReadableFormat(*route)
+		clusterURL.Status.State = StateTypeLocallyDeleted
+		return clusterURL, nil
+	}
+
+	// can't find the in local and remote
```
mik-dass: find the URL

pkg/odo/cli/url/describe.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return
+}
+
+// NewCmdURLDescribe implements the odo url describe command.
+func NewCmdURLDescribe(name, fullName string) *cobra.Command {
+	o := NewURLDescribeOptions()
+	urlDescribeCmd := &cobra.Command{
+		Use:         name,
```
mik-dass: name + " [url name]"

jichenjc: ok, will do

pkg/url/url.go
jichenjc: ok

pkg/url/url.go
jichenjc: oops, yes

pkg/url/url.go
jichenjc: ok

tests/integration/cmd_url_test.go
jichenjc: ok

kadel:
This might not be that useful right now, as it is not providing any additional information compared to `odo url list` but I can see that in the future odo can provide more detailed information about url.

👍 
/approve

amitkrout:
/lgtm

Issue #2697, opened 2020-03-10
CheckForExistence in helper_oc.go only checks for routes


johnmcollier:
I was poking through the integration test code and saw that `CheckForExistence` in `test/helper/helper_oc.go` doesn't actually work as intended:
https://github.com/openshift/odo/blob/9b3c652fdf9b155273c06909165fa5dc48d9e858/tests/helper/helper_oc.go#L434-L439

It hardcodes the resource type to `routes` rather than the resource type that was passed in, so all of these checks in `test/integration/component.go` don't function as intended: 

https://github.com/openshift/odo/blob/9b3c652fdf9b155273c06909165fa5dc48d9e858/tests/integration/component.go#L849-L853

/kind bug
/area testing
/assign

PR #2698, opened 2020-03-10
Update CheckForExistence to get right resource

johnmcollier:

**What type of PR is this?**

/kind cleanup

I was poking through the integration test code and saw that `CheckForExistence` in `test/helper/helper_oc.go` doesn't actually work as intended:
https://github.com/openshift/odo/blob/9b3c652fdf9b155273c06909165fa5dc48d9e858/tests/helper/helper_oc.go#L434-L439

This PR updates `CheckForExistence` so that it uses the passed in resource type, rather than just routes.

Fixes https://github.com/openshift/odo/issues/2697

### DIFF
--- a/tests/helper/helper_oc.go
+++ b/tests/helper/helper_oc.go
@@ -432,7 +432,7 @@ func (oc *OcRunner) WaitForDCRollout(dcName string, project string, timeout time
 
 // CheckForExistence checks if the given resource exists on the cluster
 func (oc *OcRunner) CheckForExistence(resourceName, namespace string) {
-	session := CmdRunner(oc.path, "get", "routes", "--namespace", namespace)
+	session := CmdRunner(oc.path, "get", resourceName, "--namespace", namespace)
 	Eventually(session).Should(gexec.Exit(0))
 	output := string(session.Wait().Out.Contents())
 	Expect(output).To(ContainSubstring(""))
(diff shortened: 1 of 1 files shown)


tests/helper/helper_oc.go
```diff
@@ -430,9 +430,9 @@ func (oc *OcRunner) WaitForDCRollout(dcName string, project string, timeout time
...
 
-// CheckForExistence checks if the given resource exists on the cluster
-func (oc *OcRunner) CheckForExistence(resourceName, namespace string) {
-	session := CmdRunner(oc.path, "get", "routes", "--namespace", namespace)
+// CheckForExistence checks if the given resource type exists on the cluster
+func (oc *OcRunner) CheckForExistence(resourceType, namespace string) {
+	session := CmdRunner(oc.path, "get", resourceType, "--namespace", namespace)
 	Eventually(session).Should(gexec.Exit(0))
 	output := string(session.Wait().Out.Contents())
 	Expect(output).To(ContainSubstring(""))
```
adisky: This would still not work as expected, as `oc get <resource-type>`  returns `No resources found in <namespaceName> namespace`. It never returns empty string.

johnmcollier: I'm wondering if we can just remove that second check? If `oc get <resource-type>` returns  `No resources found in <namespaceName> namespace`, it will also return a non-zero exit code, meaning the `Eventually(session).Should(gexec.Exit(0))` would fail.

Issue #2684, opened 2020-03-05
[Devfile] odo needs to be storing namespace in local env.yaml file


johnmcollier:
/kind bug

## What versions of software are you using?

**Operating System:**
macOS
**Output of `odo version`:**
odo from master

## Actual behavior
I've been looking at writing integration tests for odo's devfile support and was hitting a bizarre issue: tests would intermittently fail because it was trying to retrieve Kube resources or do an odo push in a different namespace than what was expected, thus causing a number of errors

What was happening is that the integration tests would be running in parallel, and standing up their own namespace/project for each test. Since `odo push --devfile` always just uses the current active namespace (rather than one specified by the user with a `--namespace` flag or the namespace where the component was pushed to), this would mean a number of issues could occur:
    - `odo push` fails because the namespace is being terminated
    - `odo push` pushes to the wrong namespace
    - Unable to retrieve pods because the namespace we did the `odo push` in, is no longer the active namespace

We should modify `odo push` for devfiles to do the following:
- Add a `--namespace` flag to the command, allowing the user to specify the namespace to be used
    - Store the namespace in `env.yaml`
- If no value was passed in with the flag, check If a namespace exists in env.yaml and if so, use it
- Use the current active namespace if no namespace is set in either `--namespace` or in the env.yaml

## Expected behavior
`odo push` doesn't always use the current active namespace

kadel:
> We should modify `odo push` for devfiles to do the following:
> 
> * Add a `--namespace` flag to the command, allowing the user to specify the namespace to be used
>   
>   * Store the namespace in `env.yaml`
> * If no value was passed in with the flag, check If a namespace exists in env.yaml and if so, use it
> * Use the current active namespace if no namespace is set in either `--namespace` or in the env.yaml

+1


A little bit of context:
We haven't had this problem before, as for "LocalConfig" components (based on s2i) odo is storing namespace in LocalConfg in `.odo/config.yaml` with other values. But this is not a good solution as it mixes application definitions and environment-specific config in one file.
What you are proposing makes sense. 


But for your testing problem, there is a solution even with the current implementation. It should be possible to just copy the current `$KUBECONFIG` file into a new location and set `KUBECONFIG` to this copy in  `BeforeSuite` function https://onsi.github.io/ginkgo/#parallel-specs.
Putting it here just for a record. Not saying that we need to do this. We will need `--namespace` flag anyway

GeekArthur:
[...]

Issue #2557, opened 2020-01-31
Create component support for odo devfile scenario


elsony:
[kind/Feature]
## Which functionality do you think we should add?
Create component support for odo devfile scenario:

- Provide ability to filter devfiles that can be supported by odo devfile support in a given Che devfile repository
- When the experimental flag is enabled (see https://github.com/openshift/odo/issues/2536),  "odo catalog list components" command will also include the list of components (on top of the traditional odo components) that are provided by the Che devfile repository from above
- User can call the "odo component create" to create the devfile style of the component, it will result in a copy of the devfile (from the Che devfile repo) for that particular component
- With the existence of the devfile on the project, all further odo operations, e.g. odo push, will be using the devfile support

## Why is this needed?
The IDP definition of the devfile style applications are defined in a Che devfile repository.  The Che devfile repository will contain a mix of devfile for regular Che usage as well as the ones that can provide IDP support for odo.  We'll need a way to filter the devfile repo to build the list of components that odo supports.  Combined with the odo create component support on the devfile style components, the user will be able to build and run their application via the devfile style component support.

elsony:
kind/feature
area/devfile

elsony:
kind/feature

GeekArthur:
@elsony I have two questions need to confirm:
1. Is this the Che devfile repository/registry location: https://github.com/eclipse/che-devfile-registry/tree/master/devfiles?
2. Should this issue also include pushing odo supported devfile (I think the odo supported devfile should have more customized fields and they are not fully decided yet) to Che devfile repository/registry?

kadel:
> 2\. Should this issue also include pushing odo supported devfile (I think the odo supported devfile should have more customized fields and they are not fully decided yet) to Che devfile repository/registry?

There should not be any special odo supported devfile. The goal is to have any Che supported devfile to be also odo supported devfile.




> 1. Is this the Che devfile repository/registry location: https://github.com/eclipse/che-devfile-registry/tree/master/devfiles?

That is the source for devfile registry. This devfile registry is supposed to run inside the cluster.
For example, for public che.redhat.com the registry is at https://che-devfile-registry.openshift.io//devfiles/index.json

elsony:
> There should not be any special odo supported devfile. The goal is to have any Che supported devfile to be also odo supported devfile.

[...]

PR #2699, opened 2020-03-10
Devfile support for odo create

GeekArthur:

**What type of PR is this?**
/kind feature

This PR implements devfile support for `odo create`, basically the `odo create` command does two things:
- Download supported `devfile.yaml` from predefined registries to user's local file system
- Create configuration file (file path: `./.odo/env/env.yaml`) and store component metadata (eg. component name, component namespace) into the the configuration file

It covers three cases below:
- `odo create <component type>`
- `odo create <component type> <component name>`
- `odo create <component type> <component name> --project <namespace>`

Note: Supported devfile component should satisfy the following conditions:
- Devfile has `dockerimage` as component type
- Devfile has `alias`
- Devfile has `devRun` command
- Devfile has `devBuild` command

Fixes https://github.com/openshift/odo/issues/2557
Fixes https://github.com/openshift/odo/issues/2684

Unit tests have been done in this PR. All devfile support integration tests can start once [integration test suites PR](https://github.com/openshift/odo/issues/2673) merges in.


pkg/envinfo/envinfo.go
```diff
@@ -135,6 +138,11 @@ func newProxyEnvInfo() proxyEnvInfo {
 func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}) (err error) {
 	if parameter, ok := asLocallySupportedParameter(parameter); ok {
 		switch parameter {
+		case "create":
```
yangcao77: Why we use a single parameter for `Name`, `Type` and `Namespace`? 
Should these 3 be separate parameters?
Putting multiple settings into a signle parameter would cause error for some of the functions. e.g. `isSet()` returns `util.IsSet(esi.componentSettings, parameter)`.

GeekArthur: Single parameter can help assign all env fields related with `odo create` with one function call of `SetConfiguration` instead of multiple function calls of `SetConfiguration`, so using single parameter can:
- Save the number of function calls
- To make it clear that we get `Name`, `Type` and `Namespace` together from `odo create` and assign them together

Given the `SetConfiguration` is just for feeding info into `componentSettings` struct and write to local file system, so there shouldn't be any affect for single parameter.

Regarding the affect of `isSet()`, from the screenshot below we can see `isSet()` is only called by test cases, there is no code in product calls that function. (If that's the case I would suggest to put `isSet()` in `envinfo_test.go` instead of `envinfo.go`)
Screen Shot 2020-03-11 at 2 40 20 PM

So `isSet()` is just a testing function to test if the `componentSettings` struct is set properly, using single parameter (set all fields together) shouldn't affect the setting of `componentSettings` struct. However, `util.IsSet(esi.componentSettings, parameter)` is the wrapper of [reflect package](https://golang.org/pkg/reflect/) which helps checking the field's kind, value, etc. The `util.IsSet(esi.componentSettings, parameter)` function itself is lack of checking `string` field, so I add the changes here https://github.com/openshift/odo/pull/2699/files#diff-975cb50bfa08efda085ba94fac1c40c9R81 and no errors will be caused.

yangcao77: Understood. 
The `isSet()` function allows check of the parameter existence with env.yaml file, it is not a test function specific, even it's currently only being called in it's unit test.

GeekArthur: Make senses, we can keep it in `envinfo.go` for future usage.

pkg/odo/cli/component/create.go
```diff
@@ -263,8 +282,50 @@ func (co *CreateOptions) setResourceLimits() error {
...
+		return errors.New("env.yaml already exists")
+	}
+
+	return nil
+}
+
 // Complete completes create args
 func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	if experimental.IsExperimentalModeEnabled() {
+		co.devfileMetadata.componentType = args[0]
```
johnmcollier: This line is before the section that checks the number of args passed in to odo:

https://github.com/openshift/odo/blob/4852786c14ff643c490dbdd1ea31d6b699d1590f/pkg/odo/cli/component/create.go#L268-L270

If you run just `odo create`  without any args you'll get a panic from Go:
```
johns-mbp-3:odo johncollier$ ./odo create
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/openshift/odo/pkg/odo/cli/component.(*CreateOptions).Complete(0xc0001a23c0, 0xc00073c240, 0x6, 0xc000a02fc0, 0x3e2f790, 0x0, 0x0, 0x2, 0x279f0c7)
	/Users/johncollier/go/src/github.com/openshift/odo/pkg/odo/cli/component/create.go:301 +0x1b5b
```

We should:
- Only reference the args after we've validated the number of args passed in
- Support interactive mode if possible, as odo1 also supported it

pkg/odo/cli/component/create.go
```diff
@@ -555,6 +647,10 @@ func NewCmdCreate(name, fullName string) *cobra.Command {
 	componentCreateCmd.Flags().StringSliceVarP(&co.componentPorts, "port", "p", []string{}, "Ports to be used when the component is created (ex. 8080,8100/tcp,9100/udp)")
 	componentCreateCmd.Flags().StringSliceVar(&co.componentEnvVars, "env", []string{}, "Environmental variables for the component. For example --env VariableName=Value")
 
+	if experimental.IsExperimentalModeEnabled() {
```
johnmcollier: Wondering if we should gate the `projectCmd.AddProjectFlag(componentCreateCmd)` line farther down for when the experimental mode is disabled. We won't need both `--project` and `--namespace` when working with devfile components

pkg/odo/cli/component/create.go
```diff
@@ -263,8 +282,50 @@ func (co *CreateOptions) setResourceLimits() error {
...
 }
 
+// Validate if devfile component already exists
+func validateDevfileComponentCreateRequest() error {
+	if util.CheckPathExists(DevfilePath) {
+		return errors.New("Devfile.yaml already exists")
+	}
+
+	if util.CheckPathExists(EnvFilePath) {
+		return errors.New("env.yaml already exists")
```
johnmcollier: What happens if the devfile.yaml is missing but there is already an env.yaml?

johnmcollier:
Another thought I've had: should `odo create` also be pulling down any associated projects in the devfile?

E.g. if a devfile has:
```
projects:
  -
    name: springbootproject
    source:
      type: git
      location: "https://github.com/spring-projects/spring-petclinic"
```

Should odo also be git cloning the project when we do `odo create`?

pkg/odo/cli/component/create.go
maysunfaisal: Also, how do we handle component name if user does not pass it in? I can do `odo create openLiberty` and no component name is set... we should fix that as well.

We should also check for the size of the component name, since it is going to create kube resources.

You can have a look at function https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/devfile.go#L42 for trimming component name

I think this PR should also update the same line - https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/devfile.go#L42, here we should read it from env.yaml, and you can test odo create & odo push with the same PR :D

pkg/odo/cli/component/create.go
```diff
@@ -477,6 +552,23 @@ func (co *CreateOptions) Validate() (err error) {
 
 // Run has the logic to perform the required actions as part of command
 func (co *CreateOptions) Run() (err error) {
+	if experimental.IsExperimentalModeEnabled() {
+		if co.devfileMetadata.devfileSupport {
```
maysunfaisal: nit: can be a single if statement and reduce indent clutters

pkg/envinfo/envinfo.go
```diff
@@ -135,6 +138,11 @@ func newProxyEnvInfo() proxyEnvInfo {
...
+			createValue := value.(ComponentSettings)
+			esi.componentSettings.Type = createValue.Type
+			esi.componentSettings.Name = createValue.Name
```
maysunfaisal: we should have a GetName() func like GetURL() to read the component name for usage in devfile.go(see my other review comment)

pkg/odo/cli/component/create.go
maysunfaisal: I think we should not overwrite devfile if its present in the local disk. If user devfile is present, we should continue with that!

pkg/odo/cli/component/create.go
GeekArthur: Good catch, I would go for the first option. I discussed the interactive mode work with Elson before, given finally we need to integrate odo2 with Codewind, CLI interactive mode doesn't have that much value for us and I believe our IDE can handle all interactive stuff properly.

pkg/odo/cli/component/create.go
GeekArthur: My original thinking was we should error out if any file is already exists. I guess your point is if devfile.yaml is missing but there is already an env.yaml we should pull down the devfile.yaml according to the existing env.yaml? If that's your point we also need to consider if there is a conflict between the new arguments that user pass in via `odo create` and the existing env.yaml. So follow the existing design as odo1 and keep it simple, we only generate two files if they are not exist, otherwise we error out.

But I think I still need to modify the code to cover all cases and throw proper errors, I think I should do:

Case | Current version | Modified version
------------|------------ | -------------
Both are missing | Create both files | Create both files
Only devfile.yaml is already exists | Throw error that devfile.yaml already exists | Throw error that devfile.yaml already exists
Only env.yaml is already exists | Throw error that env.yaml already exists | Throw error that env.yaml already exists
Both are already exist | Throw error that devfile.yaml already exists | **Throw errors that both files are already exist**

pkg/odo/cli/component/create.go
GeekArthur: I think the bottom line is we don't want to break the existing `s2i` support, so I am worry about gating `projectCmd.AddProjectFlag(componentCreateCmd)` may cause that kind of issue. I agree it's kind of redundant but it's not too bad to let them live together currently to avoid breaking changes. Also for `--namespace` flag we already had the specific description to make it only applies to `devfile` since it's generic kube (I think there is no `Project` definition in generic kube):
```
-n, --namespace string    Create devfile component under specific namespace
```

GeekArthur:
> Nice changes Jingfu. I've left some review comments. Can you add some integration tests too? 

Since your changes don't rely on odo push, you won't need my changes in #2681.
Sure, I can add the integration tests for `odo create`.

> Additionally, if I do odo create <component> where <component> is a non existent devfile component, it tells me:
johns-mbp-3:odo johncollier$ ./odo create fakedevfile
 ✗  imagestreams.image.openshift.io "fakedevfile" not found
We shouldn't be telling them that an image stream with that name didn't exist, but rather a devfile component with that name didn't exist

That's the current expected behavior and I agree the error message a little bit misleading. I discussed this with Elson, basically the design is we wanna support both `devfile` and `s2i` but `devfile` has higher priority, that means if we can't find the component type in `devfile` we still keep checking if we can find that in `s2i`, that why you hit error from `s2i` instead of `devfile`. That's being said, I can add error message for `devfile` but don't terminate the code. So for the case that component don't exist in both `devfile` and `s2i` the error message would be something like:

```
Checking if the component type "fakedevfile" is supported by devfile
X Component type "fakedevfile" is not supported by devfile

Checking if the component type "fakedevfile" is supported by s2i
X imagestreams.image.openshift.io "fakedevfile" not found 
```

GeekArthur:
> Another thought I've had: I'm thinking `odo create`should also be pulling down any associated projects in the devfile?
> 
> E.g. if a devfile has:
> 
> ```
> projects:
>   -
>     name: springbootproject
>     source:
>       type: git
>       location: "https://github.com/spring-projects/spring-petclinic"
> ```
> 
> Then odo would git clone this project (and any other projects in the devfile) during `odo create`. This would be in-line with how Che treats projects in devfiles (it clones the projects when the workspaces is first created)

If we follow the existing odo design, `odo create` shouldn't pull down the project source/template automatically according to the `devfile.yaml`. I think the problem is that user would have issue to build their own project if we help user pull down the project automatically, and I think this may results in the affection of `bind` case once we integrate with Codewind and cause some conflict since currently IDE/Portal handles pulling down project source/template.

pkg/odo/cli/component/create.go
GeekArthur: @maysunfaisal I remember last time Rajiv said `odo push` can pick up the component type as default component name if no component name is set, but to make it clear, I can make the env.yaml has the default component name if user doesn't specify.

If we already had the util function for sanitizing, I can directly apply it to `odo create` instead of `odo push` so that we don't need double sanitization.

That's all good suggestions!

pkg/envinfo/envinfo.go
GeekArthur: @maysunfaisal Answer in your other review, should handle the two comments together

pkg/odo/cli/component/create.go
GeekArthur: Yeah, I know the logic can be optimized, the reason I leave all `experimental.IsExperimentalModeEnabled()` as independent condition is because it's better for readability (easy to separate experimental feature) and easy to remove since finally we don't need that condition.

pkg/odo/cli/component/create.go
maysunfaisal: I think we should
- create component name in `odo create`
- if not provided during `odo create`, create one maybe like `devfile.go` `os.Getwd()` and store in `env.yaml`
- make sure component length is good, because we will create deployments, volume, url with component (check `devfile.go`)
- `odo push` should do `GetName()` from `envinf.go`. If not present push should error out. 

So `odo push` does not do component name creation, it just reads it from `env.yaml` and `odo create` does everything for the component name. I think current odo non-experimental is also like that? `odo create` creates component name in `config.yaml`

GeekArthur: @maysunfaisal Yes, that's what I am saying above, which `env.yaml` should have the component name field (either from user via `odo create` or the default component name we define) and we can reuse the sanitizing util function from `odo push` to sanitize (eg. string length, special character, etc). If we handle the component name on `odo create`, definitely that part is not necessary on `odo push`.

But for the default component name I would suggest we follow the same design as current odo which uses component type as default component name instead of using project directory path which the current `odo push` uses by `os.Getwd()` function. Reason is component type is less complicated than project directory path so that we can't reduce the burden of sanitizing.

The current odo handles all component generation configurations (config.yaml) on `odo create`, and `odo push` implements those configurations. We do the same for devfile support, only difference is generation configurations are devfile.yaml + env.yaml.

pkg/odo/cli/component/create.go
GeekArthur: @maysunfaisal Since we don't if user wants to overwrite or still use the existing devfile, so I would say we can leave the option to user and let user decide which way is better, so that's why I just directly return the error that devfile is already exists (same for existing odo, odo won't overwrite user's previous configuration).

GeekArthur:
> > Nice changes Jingfu. I've left some review comments. Can you add some integration tests too?
> > Since your changes don't rely on odo push, you won't need my changes in #2681.
> 
> Sure, I can add the integration tests for `odo create`.
> 
> > Additionally, if I do odo create  where  is a non existent devfile component, it tells me:
> > ```
> > johns-mbp-3:odo johncollier$ ./odo create fakedevfile
> > ✗  imagestreams.image.openshift.io "fakedevfile" not found
> > ```
> > 
> > 
> > We shouldn't be telling them that an image stream with that name didn't exist, but rather a devfile component with that name didn't exist
> 
> That's the current expected behavior and I agree the error message a little bit misleading. I discussed this with Elson, basically the design is we wanna support both `devfile` and `s2i` but `devfile` has higher priority, that means if we can't find the component type in `devfile` we still keep checking if we can find that in `s2i`, that why you hit error from `s2i` instead of `devfile`. That's being said, I can add error message for `devfile` but don't terminate the code. So for the case that component don't exist in both `devfile` and `s2i` the error message would be something like:
> 
> ```
> Checking if the component type "fakedevfile" is supported by devfile
> X Component type "fakedevfile" is not supported by devfile
> 
> Checking if the component type "fakedevfile" is supported by s2i
> X imagestreams.image.openshift.io "fakedevfile" not found 
> ```

I just change my mind, since we don't use `s2i` support, I think for the time being we can only error the errors related with devfile support out.

pkg/odo/cli/component/create.go
GeekArthur: @johnmcollier But in the meantime, I can take a look the interactive mode.

pkg/odo/cli/component/create.go
GeekArthur: According to the latest design, we share `--project` flag for s2i and devfile

pkg/envinfo/envinfo_test.go
```diff
@@ -23,6 +23,7 @@ func TestSetEnvInfo(t *testing.T) {
 	os.Setenv(envInfoEnvName, tempEnvFile.Name())
 	testURL := EnvInfoURL{Name: "testURL", ClusterHost: "1.2.3.4.nip.io", TLSSecret: "testTLSSecret"}
 	invalidParam := "invalidParameter"
+	testCreate := ComponentSettings{Type: "componentType", Name: "componentName", Namespace: "namespace"}
```
kadel:
```suggestion
	testCreate := ComponentSettings{Name: "componentName", Namespace: "namespace"}
```

kadel:
```suggestion
	testCreate := ComponentSettings{Name: "componentName", Namespace: "namespace"}
```

GeekArthur: Yeah, I already made the changes but they were in my local branch, will commit them with my integration tests.

GeekArthur: Yeah, I already made the changes but they were in my local branch, will commit them with my integration tests.

GeekArthur:
All comments are addressed, integration and unit tests are added. @johnmcollier @yangcao77 @maysunfaisal @kadel Please review again, thanks!

tests/integration/cmd_devfile_create_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	Context("When executing odo create with devfile component type argument and --project flag", func() {
+		It("should successfully create the devfile component", func() {
+			componentNamespace := helper.RandString(6)
+			helper.CmdShouldPass("odo", "create", "openLiberty", "--project", componentNamespace)
+		})
+	})
+
+	Context("When executing odo create with devfile component name that contains unsupported character", func() {
+		It("", func() {
```
amitkrout: ```"" -> "should fail with proper error message"```

tests/integration/cmd_devfile_create_test.go
```diff
@@ -0,0 +1,74 @@
...
+			componentName := "BAD@123"
+			helper.CmdShouldFail("odo", "create", "openLiberty", componentName)
```
amitkrout: Assert the message substring just for verification

tests/integration/cmd_devfile_create_test.go
```diff
@@ -0,0 +1,74 @@
...
+		})
+	})
+
+	Context("When executing odo create with devfile component name that contains all numeric values", func() {
+		It("", func() {
```
amitkrout: https://github.com/openshift/odo/pull/2699/files#r395980052

tests/integration/cmd_devfile_create_test.go
```diff
@@ -0,0 +1,74 @@
...
+			componentName := "123456"
+			helper.CmdShouldFail("odo", "create", "openLiberty", componentName)
+		})
+	})
+
+	Context("When executing odo create with devfile component name that contains more than 63 characters ", func() {
+		It("", func() {
```
amitkrout: https://github.com/openshift/odo/pull/2699/files#r395980052

adisky:
@GeekArthur Since Integration tests are not here, could please add manual steps/example to test this PR.

GeekArthur:
> @GeekArthur Since Integration tests are not here, could please add manual steps/example to test this PR.

@adisky Thanks! I will add my integration tests to the corresponding `.travis.yml` and `Makefile` to enable the integration tests

pkg/odo/cli/component/create.go
```diff
@@ -456,6 +579,27 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 
 // Validate validates the create parameters
 func (co *CreateOptions) Validate() (err error) {
+	if experimental.IsExperimentalModeEnabled() {
+		if co.devfileMetadata.devfileSupport {
+			// Validate if the devfile component that user wants to create already exists
+			spinner := log.Spinner("Validating component")
+			defer spinner.End(false)
+
+			if util.CheckPathExists(DevfilePath) || util.CheckPathExists(EnvFilePath) {
```
adisky: we need to check this before running `catalogDevfileList, err := catalog.ListDevfileComponents()`, currently if you run `odo create openLiberty` and then run `odo create openLiberty` in the same folder it is taking too much time to return `"Devfile.yaml or env.yaml already exists`, I think this should be the first validation.

pkg/odo/cli/component/create.go
```diff
@@ -265,6 +285,109 @@ func (co *CreateOptions) setResourceLimits() error {
 
 // Complete completes create args
 func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	if experimental.IsExperimentalModeEnabled() {
```
adisky: just a thought should we make separate  function `func (co *CreateOptions) completeDevfile(...)` and do like this here
```
completeDevfile(...)
}
```

adisky:
@GeekArthur I tried this out, Works well. Looks good to me.

tests/integration/cmd_devfile_create_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// This is run after every Spec (It)
+	var _ = BeforeEach(func() {
+		SetDefaultEventuallyTimeout(10 * time.Minute)
+		SetDefaultConsistentlyDuration(30 * time.Second)
+		project = helper.CreateRandProject()
+		context = helper.CreateNewContext()
+		originalDir = helper.Getwd()
+		helper.Chdir(context)
+		helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
```
johnmcollier: We shouldn't be setting and unsetting experimental mode in the `BeforeEach` / `AfterEach` sections, as the integration tests run concurrently and this could cause issues for other integration tests that _may_ be running at the same time as your's.

See https://github.com/openshift/odo/blob/master/tests/integration/cmd_devfile_push_test.go#L33 and https://github.com/openshift/odo/blob/master/tests/integration/cmd_devfile_push_test.go#L49 for how we need to set experimental mode on devfile tests

pkg/odo/cli/component/create.go
GeekArthur: By design after we implement devfile support for odo, odo needs to support both devfile and s2i components, but devfile component has higher priority.

`catalogDevfileList, err := catalog.ListDevfileComponents()` helps us check devfile registry so that we can validate if the component is supported devfile component, if it's not supported devfile component we won't validate `Devfile.yaml or env.yaml already exists` and we go to the existing s2i code path. That's why we call `catalogDevfileList, err := catalog.ListDevfileComponents()` before `Devfile.yaml or env.yaml already exists`

pkg/odo/cli/component/create.go
GeekArthur: By design after we implement devfile support for odo, odo needs to support both devfile and s2i components, but devfile component has higher priority. 

So the experimental flag will be removed finally, currently we don't separate the generic `Complete`, `Validate` and `Run` functions for all devfile support for odo features (eg odo create, odo push, odo url). I think the reason is it would be better for refactoring once we remove the experimental flag.

PR #2733, opened 2020-03-19
Extend with plugins

bigkevmcd:
**What type of PR is this?**

> /kind feature

It allows for external binaries to be used as plugins.

for example:

```shell
$ odo promote --from a --to
```
This would look for an executable called `odo-promote` on the path, and execute that, passing the parameters through.

Write the following to an executable script on your PATH called "odo-testing" (perhaps in ~/bin if you have that on your PATH).

```shell
#!/bin/sh
MESSAGE=$1
shift
ARGS=$@
echo "this is executing this command ${MESSAGE} with args ${ARGS}"
```
This should trigger executing your shell script

```shell
$ odo testing mytest and more`
this is executing this command mytest with args with more
```


amitkrout:
@bigkevmcd Thanks for the pr. I really don't understand what is the usecase and the acceptance criteria of this pr. May be you can add a document for it with more details.

Can you please create an issue and link to this pr. I am saying this because i feel we need to discuss properly and validate the thoughts behind it among the team. ping @girishramnani @kadel

kadel:
> @bigkevmcd Thanks for the pr. I really don't understand what is the usecase and the acceptance criteria of this pr. May be you can add a document for it with more details.
> 
> Can you please create an issue and link to this pr. I am saying this because i feel we need to discuss properly and validate the thoughts behind it among the team. ping @girishramnani @kadel

I had a brief discussion about this with @bigkevmcd on Slack.
It is probably the simplest plugin mechanism. It will allow others to prototype new odo commands. 

I would treat is as a hidden easter egg that should be used only for demos and experiments.

Before making this available for anyone to use, I would like to have some kind of plugin management and distribution system (maybe something like  https://krew.sigs.k8s.io/)

/approve

sbose78:
+1

amitkrout:
@bigkevmcd UT is failing at line - https://travis-ci.com/github/openshift/odo/jobs/300352024#L155. PTAL
/hold

kadel:
ping @bigkevmcd

pkg/odo/cli/plugins/plugin_handler_test.go
```diff
@@ -0,0 +1,66 @@
...
+var sampleScript = []byte(`
+#!/bin/sh
+echo 'hello'
+`)
+
+func TestDefaultHandlerLookup(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip()
+	}
+	tempDir, cleanup := makeTempDir(t)
```
girishramnani, line 21: Ideally unit tests shouldn’t interact with external entities such as a filesystem which are subject to change. Can you please use the mocked filesystem we have inplace?

girishramnani:
/hold

Waiting for this to get resolved https://github.com/openshift/odo/pull/2733#discussion_r405229775

pkg/odo/cli/plugins/plugin_handler_test.go
amitkrout: Right, @girishramnani Can you please add some reference where we have already done it. Reference would really be helpful.

girishramnani: - https://github.com/openshift/odo/blob/49eae2d4dd57d09c6180e7d6cf70df5edd70cd4a/pkg/config/config_test.go#L446
- https://github.com/openshift/odo/blob/5e9c163aac521fae608f15486bd3eb0e73c7d1c6/pkg/debug/info_test.go#L38

girishramnani: - https://github.com/openshift/odo/blob/a93f4dae81282475a6e58bf025411f516329afcb/pkg/devfile/parser/content_test.go#L10

bigkevmcd: This looks like I'd need to reimplement https://golang.org/pkg/os/exec/#LookPath on top of afero.

In other words, replace https://github.com/golang/go/blob/master/src/os/exec/lp_unix.go#L34 with a version that worked on top of afero, purely for test purposes.

I'm not sure of the value of this?

girishramnani: Then maybe consider writing an integration test?

sspeiche:
Does this work take the place of https://github.com/openshift/odo/issues/1173 or it is separate?

sspeiche:
See also https://github.com/openshift/odo/issues/1173

kadel:
> Does this work take the place of #1173 or it is separate?

It os part of that.
I look at this as a way to validate this extension mechanism.
For now, gitops experiment should be the only one who uses that to validate that this is a useful way to do odo plugins. Once validated we should create some sort of plugin manager (something like Krew), before we announce it for wider adoption.

kadel:
@girishramnani can you please re-check if your concerns were addressed

Issue #2741, opened 2020-03-20
Projects are not synced to the devfile clonePath


rajivnathan:
/kind bug
/area devfile

## What versions of software are you using?
Using odo experimental support for devfiles by building from source

## Description of the problem
Devfile sync doesn't currently handle the `clonePath` field that's used by some devfiles. eg. https://github.com/eclipse/che-devfile-registry/blob/5a41b4b0c85893a6ee99059fcceab35b8a71780c/devfiles/go/devfile.yaml#L11

## Actual behavior
The source is synced to `/projects`

## Expected behavior
The source should be synced to `/projects/src/github.com/golang/example/` which is what Che does.

## Any logs, error output, etc?
No errors for the sync, but errors result when trying to use commands because they're expecting the source to be under the clonePath directory.

johnmcollier:
/assign

This should be an easy one to fix, we'll adapt how we handle syncing when a project is defined in the devfile:
- If `clonePath` is set, we'll sync to `/projects/<clonePath>`
- If `clonePath` is not set, we'll sync to `/projects/<projectName>`, as before.

rajivnathan:
There also seems to be a `sparseCheckoutDir` property. See https://github.com/eclipse/che-devfile-registry/blob/5a41b4b0c85893a6ee99059fcceab35b8a71780c/devfiles/quarkus/devfile.yaml#L11

There are more properties for `source`, see https://github.com/redhat-developer/devfile/blob/master/docs/devfile.md#projects

johnmcollier:
As discussed on Slack, we won't need `sparseCheckoutDir` for file syncing, as it just determines _which_ folder in the Git repo to pull down.

We will need it however for template/resource creation, so I've added it as an item in https://github.com/openshift/odo/issues/2582

kadel:
> This should be an easy one to fix, we'll adapt how we handle syncing when a project is defined in the devfile:
> 
> * If `clonePath` is set, we'll sync to `/projects/<clonePath>`
> * If `clonePath` is not set, we'll sync to `/projects/<projectName>`, as before.

It looks a bit weird, but I don't think that we can't do much better than this as we currently don't support multiple projects in devfile anyway


so +1

kadel:
> * If `clonePath` is set, we'll sync to `/projects/<clonePath>`

This actually doesn't look right. I think that if the `clonePath` is set you should use it as it is.
It should be `/<clonePath>`  not `/projects/<clonePath>`

Issue #2582, opened 2020-02-07
Add ability to create application resource/template from stack


johnmcollier:
## User Story
As a user, I want to be able to use odo to point to a devfile/stack that has its own application template associated with it, and have the template generated on my disk. This will be similar to the `appsody init` command that we're trying to replicate in odo.

## Acceptance Criteria 
- [ ] It should clone/download project/application resources on my disk from a template associated with a stack/devfile
     - If `sparseCheckoutDir` is set, need to pull down the specified folder from the git repo, rather than the entire repo
- [ ] resource creation will only be triggered when:

1. experimental flag has been enabled

1. the component is a devfile component

1. there is a project defined in the devfile of the given component

1. `odo component create` command called with a `--downloadSource` flag

- [ ] if a devfile contains a command called `devInit`, the devInit command will be executed at the end of the `odo component create` to do any init setup that the stack needs
- [ ] the `devInit` command should function similarly (if not equivalently) to the `appsody init` command

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

CC @neeraj-laad

/kind user-story
/area devfile

elsony:
Added more details on the requirement on the description

elsony:
Based on the Sprint plan discussion, we need to figure out a better name than `--createResource` to clarify that the creation/cloning of resource is on the local file system instead of creating Kube resources.

EnriqueL8:
@upLukeWeston and I added some notes on this issue [here](https://docs.google.com/document/d/1vTWDRV4sfJPn_O3DNDZPd7O4ZXCRjVeXKjsYZuziPkE/edit?usp=sharing)
FYI: @johnmcollier @elsony

GeekArthur:
FYI, this issue https://github.com/openshift/odo/issues/2850 description includes the rough design for template support, we can discuss and finalize the design later then just use one issue for tracking purpose.

elsony:
This item is done with the remaining issues have separate items open

PR #2747, opened 2020-03-23
If clonePath is set, sync to it

johnmcollier:

**What type of PR is this?**

/kind bug

This PR updates the sync code to sync to the devfile's clonePath folder if it's set. Certain devfiles, such as the official Go devfile, rely on this field being set properly (see https://github.com/eclipse/che-devfile-registry/blob/5a41b4b0c85893a6ee99059fcceab35b8a71780c/devfiles/go/devfile.yaml#L46 for example)

Fixes https://github.com/openshift/odo/issues/2741

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -343,7 +343,8 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, e
 }
 
 // getSyncFolder returns the folder that we need to sync the source files to
-// If there's exactly one project defined in the devfile, return `/projects/<projectName>`
+// If there's exactly one project defined in the devfile, and clonePath isn't set return `/projects/<projectName>`
+// If there's exactly one project, and clonePath is set, return `/projects/<clonePath>`
 // Otherwise (zero projects or many), return `/projects`
 func getSyncFolder(projects []versionsCommon.DevfileProject) string {
 	if len(projects) == 1 {
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -343,10 +343,16 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, e
 }
 
 // getSyncFolder returns the folder that we need to sync the source files to
-// If there's exactly one project defined in the devfile, return `/projects/<projectName`
+// If there's exactly one project defined in the devfile, return `/projects/<projectName>`
 // Otherwise (zero projects or many), return `/projects`
 func getSyncFolder(projects []versionsCommon.DevfileProject) string {
 	if len(projects) == 1 {
+		project := projects[0]
+		// If the clonepath is set to a value, set it to be the sync folder
+		// As some devfiles rely on the code being synced to the folder in the clonepath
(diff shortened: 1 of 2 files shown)


amitkrout:
@johnmcollier Can you please add steps for how to validate this change with an example. Later we can think upon the possibility of integration test script.

johnmcollier:
@amitkrout I've added instructions on how to validate.

How to validate:
1. Download devfile with a project that uses the `clonePath` field (try https://raw.githubusercontent.com/eclipse/che-devfile-registry/5a41b4b0c85893a6ee99059fcceab35b8a71780c/devfiles/go/devfile.yaml)
2. Pull down the project specified in the devfile and put in the same folder as the devfile (as https://github.com/openshift/odo/issues/2582 hasn't been implemented)
3. Build my branch
4. Run `odo push`
5. Verify the project gets synced to `/projects/<clonePath>` in the remote container

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -343,10 +343,17 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, e
...
+// If there's exactly one project defined in the devfile, and clonePath isn't set return `/projects/<projectName>`
+// If there's exactly one project, and clonePath is set, return `/projects/<clonePath>`
 // Otherwise (zero projects or many), return `/projects`
 func getSyncFolder(projects []versionsCommon.DevfileProject) string {
 	if len(projects) == 1 {
+		project := projects[0]
+		// If the clonepath is set to a value, set it to be the sync folder
+		// As some devfiles rely on the code being synced to the folder in the clonepath
+		if project.ClonePath != nil {
+			return filepath.ToSlash(filepath.Join(kclient.OdoSourceVolumeMount, *project.ClonePath))
```
kadel: The  should be a check to prevent escaping `/projects` using `..`
see: https://github.com/redhat-developer/devfile/blob/master/docs/devfile.md#clonepath

johnmcollier: Hmm, good catch. 

How should we handle that? Che throws an error if the clonePath starts with either `/` or `..` (https://github.com/eclipse/che/blob/420d4ea34cf50e5835027016211f16b02eba8a54/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/devfile/convert/ProjectConverter.java#L109). We should probably do the same.

johnmcollier:
@kadel Thanks for the review! I've updated the PR with changes to only allow relative clonePaths and to not allow `..` in the clonePath.

pkg/devfile/adapters/kubernetes/component/adapter_test.go
```diff
@@ -205,59 +205,126 @@ func TestGetFirstContainerWithSourceVolume(t *testing.T) {
 }
 
 func TestGetSyncFolder(t *testing.T) {
-	projectPaths := []string{"/some/path", "/another/path"}
 	projectNames := []string{"some-name", "another-name"}
 	projectRepos := []string{"https://github.com/some/repo.git", "https://github.com/another/repo.git"}
+	projectClonePath := "src/github.com/golang/example/"
+	invalidClonePaths := []string{"/var", "../var", "pkg/../../var"}
```
kadel: 👍

PR #2749, opened 2020-03-23
List devfile component without connecting cluster if experimental flag is enabled

### LINKED ISSUE #2706: [Devfile] odo catalog can't list components on non-OpenShift clusters
/kind bug

odo from master

**Operating System:**
macOS

1. Started Minikube
2. Enabled experimental mode
3. Ran `odo catalog list components`

odo reported the following error:
```
johns-mbp-3:odo johncollier$ ./odo catalog list components
 ✗  unable to list imagestreams: the server could not find the requested resource (get imagestreams.image.openshift.io).
unable to list imagestreams: the server could not find the requested resource (get imagestreams.image.openshift.io)
```

Now, this is happening since I'm not running on OpenShift and thus there are no imagestreams that can be retrieved. Since odo2 and its devfile support is Kube-generic, it should at least return the list of devfile components when running on a non-OpenShift cluster.

/kind bug
/area devfile

### LINKED ISSUE #2740: `odo catalog list components` should work without cluster
It should be possible to list devfile components even if odo is not connected to the cluster

```
 export ODO_EXPERIMENTAL=true
▶ odo catalog list components
 ✗  unable to list imagestreams: Get "<url>": dial tcp <IP>:8443: connect: connection refused.
unable to list imagestreams: Get "<url>": dial tcp <IP>:8443: connect: connection refused
```

Instead of failing, odo can just show a warning that it can't connect to the cluster, and show just devfile components.

/kind bug
/priority low
/area devfile
/area catalog

GeekArthur:

**What type of PR is this?**
/kind bug

If the experimental flag is enabled, user should be able to list devfile components even if user doesn't connect to OpenShift cluster. Instead of failing the `odo catalog list components`, we output the following warning to user:
```
odo catalog list components                                                                                  ✔  17:00:33 
 ⚠  Please log in to an OpenShift cluster to list OpenShift/s2i components

Odo Supported Devfile Components:
NAME                 DESCRIPTION
java-spring-boot     Spring Boot® using IBM Java
openLiberty          Open Liberty microservice in Java
```

Fixes #2740 
Fixes #2706 

### DIFF
--- a/pkg/odo/cli/catalog/list/components.go
+++ b/pkg/odo/cli/catalog/list/components.go
@@ -43,7 +43,11 @@ func (o *ListComponentsOptions) Complete(name string, cmd *cobra.Command, args [
 	o.Context = genericclioptions.NewContext(cmd)
 	o.catalogList, err = catalog.ListComponents(o.Client)
 	if err != nil {
-		return err
+		if experimental.IsExperimentalModeEnabled() {
+			log.Warning("Please log in to an OpenShift cluster to list OpenShift/s2i components")
+		} else {
+			return err
+		}
 	}
 
 	if experimental.IsExperimentalModeEnabled() {
--- a/pkg/odo/cli/catalog/list/components.go
+++ b/pkg/odo/cli/catalog/list/components.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"text/tabwriter"
 
+	"github.com/golang/glog"
 	"github.com/openshift/odo/pkg/catalog"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/machineoutput"
@@ -44,7 +45,7 @@ func (o *ListComponentsOptions) Complete(name string, cmd *cobra.Command, args [
 	o.catalogList, err = catalog.ListComponents(o.Client)
 	if err != nil {
 		if experimental.IsExperimentalModeEnabled() {
-			log.Warning("Please log in to an OpenShift cluster to list OpenShift/s2i components")
+			glog.V(4).Info("Please log in to an OpenShift cluster to list OpenShift/s2i components")
 		} else {
 			return err
 		}


GeekArthur:
@kadel Please review this PR, thanks!

pkg/odo/cli/catalog/list/components.go
```diff
@@ -43,7 +43,11 @@ func (o *ListComponentsOptions) Complete(name string, cmd *cobra.Command, args [
 	o.Context = genericclioptions.NewContext(cmd)
 	o.catalogList, err = catalog.ListComponents(o.Client)
 	if err != nil {
-		return err
+		if experimental.IsExperimentalModeEnabled() {
+			log.Warning("Please log in to an OpenShift cluster to list OpenShift/s2i components")
```
johnmcollier: I have some concerns about _always_ displaying that warning message when on non-OpenShift Kube clusters (or with no Kube cluster).

Maybe this is more of question: Would it be better for us to just hide the error message if experimental mode is enabled? Or leave it as you have it?

johnmcollier: Rather than pulling in the Go log library, can you use `glog` and log the message with an appropriate log level? LIke we've done elsewhere in odo.

cdrage: I agree with @johnmcollier I don't like always having that warning message.

GeekArthur: @cdrage @johnmcollier I totally agree with you guys, actually I think both ways are not perfect, and I think this is because by design we want to support both devfile and s2i together.

Always display warning message is annoying for devfile user since they don't actually need that message.

Hide warning message is not user friendly for s2i user since they still need the warning message for debugging purpose.

I think to make it better, I can use `glog` V logging style as John suggests so that devfile user won't see this warning message and s2i user can debug via -v flag.

GeekArthur: New to `glog`, I can switch to use that, and this is a good point. I find odo uses various logging packages (eg glog, log, and some of them directly use fmt) and I don't find any docs mentions odo uses them for different purpose, make the logging packages consistent are good for both development and use cases (this is the only docs for logging that I find: https://github.com/openshift/odo/blob/5eb0ff5e73ad76322724e0cc9ef3beb9f016c72e/docs/dev/logging.adoc). I can bring this up in Slack channel.

GeekArthur:
@cdrage @johnmcollier Comments addressed, please view again, thanks!

kadel:
comments addressed, and this is a small change

/approve
/lgtm

Issue #2761, opened 2020-03-25
Odo url should support both devfile and s2i components


GeekArthur:
/kind bug

## What versions of software are you using?

**Operating System:**
```
MacOS Catalina
```

**Output of `odo version`:**
```
odo v1.1.1 (5ee77dd23)

Server: https://jing-okd-master.rtp.raleigh.ibm.com:8443
OpenShift: v3.11.0+3a34b96-217
Kubernetes: v1.11.0+d4cacc0
```

## Bug description
Here is the output from our OpenShift CI integration tests
```
[odo]  •  Checking if the specified component type is supported devfile component type  ...
[odo]  ✗  Checking if the specified component type is supported devfile component type [48716ns]
[odo] 
[odo] Please run 'odo catalog list components' for a list of supported devfile component types
[odo]  •  Validating component  ...
[odo] 
 ✓  Validating component [162ms]
[odo] 
[odo] Please use `odo push` command to create the component with source deployed
Running odo with args [odo url create --port 8080]
[odo]  ✗  cluster host must be provided in order to create ingress
```
From the output we can clearly see the `odo url create --port 8080` doesn't work if creating s2i component with experimental mode is enabled. By design when experimental mode is enabled we should support both devfile and s2i components but devfile component has higher priority, currently both `odo create` and `odo push` already implemented the design, `odo url` should implement the design as well.

kadel:
TODO for me: Add more information 
/triage needs-information

GeekArthur:
@kadel The bug is `odo url create --port 8080` fails with experimental mode enabled given user uses `odo create` to create a s2i component beforehand. This issue should be fixed by PR #2762.
FYI @yangcao77

mik-dass:
@GeekArthur So the steps to reproduce this are
```
odo preference set experimental false -f
odo create nodejs
odo preference set experimental true -f
odo url create example --port 8080
```
Are these the right steps?

GeekArthur:
@mik-dass Yup, that's right steps, or even more simple
```
odo preference set experimental true -f
odo create python
odo url create example --port 8080  //fails here 
```
Because currently devfile component doesn't support python, you can create s2i python component directly with experimental mode enabled to save one step to switch experimental mode

mik-dass:
> Because currently devfile component doesn't support python, you can create s2i python component directly with experimental mode enabled

But currently while creating a python component with experimental mode enabled, it panics.

```
mrinaldas@localhost nodejs]$ odo create pyhton
[...]

Issue #2592, opened 2020-02-11
Url create support for devfile ingress


elsony:
Related to #2470

## User Story
As a user, I want to expose my service over ingress using either secure or non-secure connection.

## Acceptance Criteria 
- [ ] Example:$ odo url create --port 8081 --ingressDomain xx
- [ ] Secure example:$ odo url create --port 8081 --secure --ingressDomain xx
- [ ] It should enable HTTPS when the secure parameter has been specified
- [ ] create .odo/env folder to store environment specific information, e.g. ingressDomain

/kind user-story
/area devfile

girishramnani:
There is a correlated issue of occlient here https://github.com/openshift/odo/issues/2561
cc @mik-dass

yangcao77:
When `Experimental=true`, the devfile path has been chosen. `udo url` would create an ingress instead of Openshift route. 

`odo url create <ingressDomain>--port 8081 --secure true `
The optional flags `port`  and `secure`  work the same as current existing odo url. 
`<ingressDomain>` is required to be provided. 


If `secure=true`,  a self-signed certificate will be created and a tls secret & ingress will be created using the certificate.

yangcao77:
Here are the detailed information on the https support: 
- The certificate will use x509 with rsa key.
- `SerialNumber` will use `time.Now()` and encoded to make sure of the uniqueness. 
- Livetime of the certificate will be 1 year. 
- All ingresses of a single component will share the same ssl certificate & tls secret. 
- The tls secret will not be deleted by `odo url delete`. It will only be deleted when the component is deleted.

yangcao77:
```
stephaniesmbp2:springtest1 stephanie$ kubectl get ingress
NAME                          HOSTS                               ADDRESS   PORTS     AGE
spring-springtest1-gldv-8080-app    spring-springtest1-gldv-8080.9.42.26.107.nip.io              80, 443   9d
stephaniesmbp2:springtest1 stephanie$ kubectl get ingress spring-springtest1-gldv-app -o yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  creationTimestamp: "2020-02-04T20:22:56Z"
  generation: 1
  labels:
    app.kubernetes.io/component-name: spring-springtest1-gldv
    app.kubernetes.io/name: app
    app.kubernetes.io/url-name: spring-springtest1-gldv
  name: spring-springtest1-gldv-8080-app
  namespace: che
  resourceVersion: "30805111"
[...]

PR #2735, opened 2020-03-19
Devfile command execution

maysunfaisal:
Signed-off-by: Maysun J Faisal <maysun.j.faisal@ibm.com>

**What type of PR is this?**
/kind feature

**What does does this PR do / why we need it**:
This PR:
- executes `devBuild` and `devRun` commands from the devfile
- bootstraps supervisord to the devfile run component container from an init container
- makes supervisord the entrypoint of the devfile run component container if no entrypoint is mentioned in the devfile
- executes the `devRun` command with supervisord
- introduces two new push flags `build-command` & `run-command` which can be used instead of `devBuild` and `devRun`
- unit tests

**Which issue(s) this PR fixes**:

Fixes #2654

**How to test changes / Special notes to the reviewer**:

Until openshift/odo-init-image#52  is merged, pls set `ODO_BOOTSTRAPPER_IMAGE=jeevandroid/odo-init-image` for testing with the openshift/odo-init-image#52 changes.

`odo push`
`odo push --show-log`
`odo push --build-command="some build command" --run-command="some run command"`

Issue #2727, opened 2020-03-17
Add option to create Route instead of Ingress


kadel:
/kind user-story

## User Story
As an OpenShift and odo user, I want to create Route instead of k8s Ingress  So that benefit from some of the Route's additional features,  because my organization is using only Routes.

## Acceptance Criteria 
- [ ] `odo push` should detect the cluster type and create a route instead of ingress for devfile components on openshift clusters.
- [ ] `odo push` should create ingress instead of route if this was requested by the user by using `--ingress` flag.

/kind user-story
/area devfile
/area openshift
/priority high

kadel:
> * [ ]  It should be possible to specify that Route should be used instead of Ingress when the url is created. (`odo url create --ingress`, or maybe `odo url create --openshift`?)
> *

This information should be saved in `env.yaml` as whatever to use Route or Ingress is environment-specific setting.

elsony:
What happens if someone specifies `odo url create --openshift` but their target context is not OpenShift, should it fail when doing `odo push`?

kadel:
> What happens if someone specifies `odo url create --openshift` but their target context is not OpenShift, should it fail when doing `odo push`?

Yes, it should fail

kadel:
> [<img alt="@girishramnani" width="20" height="20" src="https://avatars0.githubusercontent.com/u/6551988?s=60&amp;u=93eaf1f99f7a22757db5bad83759b6cf8fd40007&amp;v=4">](/girishramnani) > [girishramnani](/girishramnani) added the [status/blocked](/openshift/odo/labels/status%2Fblocked) label [22 hours ago](#event-3188088780)

blocked don what?

mik-dass:
blocked on #2762 

`odo url create --openshift` sounds better to me as we are planning to move towards devfile completely later on.

dharmit:
I feel that many users won't have detailed idea about kubernetes using ingress and OpenShift using routes. I understand that a few users use OpenShift only as a Kubernetes and would like to use ingress instead of routes. But does that comprise majority of the odo users that use OpenShift? I doubt that.

I think we should do something like this:
* odo should automagically create ingress if working on kubernetes cluster
* odo should automagically create route if working on OpenShift cluster
* If user wants to create ingress on OpenShift cluster, they have to specify that with a flag

Besides that `--ingress` sounds better than `--openshift` since it's ingress that the user's intending to create when they do `odo url create`.

adisky:
[...]

PR #2762, opened 2020-03-26
Odo url Ingress support

### LINKED ISSUE #2657: Unit tests required for pkg/kclient/ingress.go & secrets.go
While creating the base function for ingress support in kclient, (as part of https://github.com/openshift/odo/pull/2636), the unit test for pkg/kclient/ingress.go & secrets.go is missing in that PR.

The functions might be changed, and more functions might be added when implementing the feature of ingress support for `odo url create/list/delete` when devfile is provided. (https://github.com/openshift/odo/issues/2592)

Creating this issue to track the absence of unit tests for `ingress.go` and `secrets.go` under kclient pkg. 

/area testing

### LINKED ISSUE #2719: Simplifying URL parameters and not using camelCase (ex. tlsSecret being tls-secret)
[kind/Feature]

I'm a bit late to the party, but I saw https://github.com/openshift/odo/pull/2636 be merged in a while ago.

Compared to all of our other parameters, (`--max-memory`, `--min-memory`, `--min-cpu`, etc.), the commands for URL are unconvential with the use of camel-case.

Other tools such as `oc` and `kubectl` do *not* use camelCase for their parameters.

I propose that we change the implementation of https://github.com/openshift/odo/pull/2636 to using dashes for the variable names.

Ex. tlsSecret will become tls-secret 

## Why is this needed?

Coincide with all the other parameters :)

### LINKED ISSUE #2721: ClusterHost is an incorrect parameter / term for url creation
/kind feature

`--clusterHost` should be `--host` for `odo url create`.

## Why is this needed?

ClusterHost parameter for url creation is an incorrect parameter term.

The term **Cluster Host** in kubectl and oc is confusing because Cluster Host actually means Kubernetes / OpenShift "host".

It is confusing because we are specifying a domain name for url creation.. ex: `odo url create --clusterHost example.domain.com`

We should be using the correct term which would actually be `host`, which is exactly what OpenShift uses in the yaml creation for Route.

For example, the current code that is implemented looks like this:

```go
				Host: ingressParam.IngressDomain,
```

yangcao77:
**What type of PR is this?**

/kind feature

This PR implements `odo url create/list/delete/describe` support for devfile

**create**
`odo url create <urlName> --host <clusterIP>` would create an ingress which domain is `<urlName>.<clusterIP>`
`--port` flag is required if the devfile has more than one ports exposed. If only one port is exposed, the flag is optional. 
`--secured` flag is to specify if it is a secured URL. 
`--tls-secret` flag is to specify if the user has his own tls secret want to bind with the ingress

Here are the detailed information on the https support:
- The certificate will use x509 with rsa key.
- `SerialNumber` will use time.Now() and encoded to make sure of the uniqueness.
- All ingresses of a single component will share the same ssl certificate & tls secret. If secure=true will check if there is a valid tls secret existing for the component, if exists, use the existing one; if not, create a new one.
- The tls secret will not be deleted by odo url delete. It will only be deleted when the component is deleted.

This PR contains unit tests for ingress.go, secret.go and integration test for odo url

Fixes #2721 
Fixes #2719 
Fixes #2657 
Implemented #2592 

**Automation test**
`make test-cmd-devfile-url`
Expect result: 
```
ginkgo  -randomizeAllSpecs -slowSpecThreshold=120 -timeout 7200s -nodes=2 -focus="odo devfile url command tests" tests/integration/
Running Suite: Integration Suite
================================
Random Seed: 1585200121 - Will randomize all specs
Will run 169 specs

Running in parallel across 2 nodes

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSS•SSSSSSSSSSSSSSSSSSSSSSSSSSSSS
Ran 4 of 169 Specs in 100.660 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 165 Skipped
```

**Manual test**
git clone https://github.com/rajivnathan/openLibertyDevfile

```
odo create openLiberty openlibertydevfile 
odo url create testurl --port 9090 --host 1.2.3.4.nip.io --secure
odo push
```
Expected: 
```
 •  Push devfile component openlibertydevfile  ...

Applying URL changes
 ✓  URL testurl: https://testurl.1.2.3.4.nip.io created
 ✓  Checking files for pushing [4ms]
 ✓  Waiting for component to start [4s]
 ✓  Syncing files to the component [3s]
 ✓  Push devfile component openlibertydevfile [10s]
```

check the ingress existence:
```
$ kubectl get ingress
NAME                          HOSTS                              ADDRESS   PORTS     AGE
testurl                       testurl.1.2.3.4.nip.io                       80, 443   42s
```

check the tle secret has been created:
```
$ kubectl get secrets
NAME                                                           TYPE                                  DATA   AGE
openlibertydevfile-tlssecret                                   kubernetes.io/tls                     2      1m
```

list the url:
```
$ odo url list
Found the following URLs for component openlibertydevfile
NAME        URL                                PORT
testurl     https://testurl.1.2.3.4.nip.io     9090
```

```
$ odo url list -o json
{
    "kind": "List",
    "apiVersion": "udo.udo.io/v1alpha1",
    "metadata": {},
    "items": [
        {
            "kind": "Ingress",
            "apiVersion": "extensions/v1beta1",
            "metadata": {
                "name": "testurl",
                "creationTimestamp": null
            },
            "spec": {
                "tls": [
                    {
                        "hosts": [
                            "testurl.1.2.3.4.nip.io"
                        ],
                        "secretName": "openlibertydevfile-tlssecret"
                    }
                ],
                "rules": [
                    {
                        "host": "testurl.1.2.3.4.nip.io",
                        "http": {
                            "paths": [
                                {
                                    "path": "/",
                                    "backend": {
                                        "serviceName": "openlibertydevfile",
                                        "servicePort": 9090
                                    }
                                }
                            ]
                        }
                    }
                ]
            },
            "status": {
                "loadBalancer": {}
            }
        }
    ]
}
```

describe the url:
```
$ odo url describe testurl
NAME        URL                                PORT
testurl     https://testurl.1.2.3.4.nip.io     9090
```

delete the url: 
```
$ odo url delete testurl
? Are you sure you want to delete the url testurl Yes
 ✓  URL testurl removed from the config file

To delete URL on the OpenShift Cluster, please use `odo push`
```
check the url info has been deleted from `.odo/env/env.yaml`
```
$ cat .odo/env/env.yaml 
ComponentSettings:
  Url: []
```
run `odo push`
```
e$ odo push
 •  Push devfile component openlibertydevfile  ...

Applying URL changes
 ✓  URL testurl successfully deleted
 ✓  Checking file changes for pushing [3ms]
 ✓  Push devfile component openlibertydevfile [1s]
```
check the ingress has been deleted. but the tls secret still exist

run `odo url list` again:
```
$ odo url list
 ✗  no URLs found for component openlibertydevfile
```


mik-dass:
@yangcao77 Please add some steps/instructions on how to test this PR in the description.

amitkrout:
@yangcao77 Can you please check golint target failure at - https://travis-ci.com/github/openshift/odo/jobs/304804234#L501

.travis.yml
```diff
@@ -86,6 +86,7 @@ jobs:
         - odo login -u developer
         - travis_wait make test-cmd-pref-config
         - travis_wait make test-cmd-url
+         - travis_wait make test-cmd-devfile-url
```
amitkrout: Please fix the indentation

GeekArthur:
Fixes https://github.com/openshift/odo/issues/2761

pkg/component/component.go
```diff
@@ -525,36 +530,39 @@ func ValidateComponentCreateRequest(client *occlient.Client, componentSettings c
...
+//  cmpExist: true if components exists in the cluster
 // Returns:
 //	err: Errors if any else nil
-func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo, stdout io.Writer, cmpExist bool) (err error) {
-
-	// if component exist then only call the update function
-	if cmpExist {
-		if err = Update(client, componentConfig, componentConfig.GetSourceLocation(), stdout); err != nil {
-			return err
+func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool) (err error) {
```
mik-dass: We need to refactor this function. The URL list function makes a list call in checkIfURLChangesWillBeMade(), then in ApplyConfigCreateURL() (it calls Exists internally which calls list) and then in applyConfigDeleteURL(). Since list can be very expensive call, we should call list once and compute the URLs to be added/deleted and then proceed with the respective actions.

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+		if err != nil {
+			return err
+		}
+		ingressList, err := urlpkg.ListPushedIngress(kClient, componentName)
+		if err != nil {
+			return err
+		}
+		localURLList := envSpecificInfo.GetURL()
+		for _, i := range ingressList.Items {
+			if !checkIfURLPresentInConfig(checkIfURLPresentInConfigParam{envinfoURL: localURLList, url: i.Name}) {
```
mik-dass: Since we are comparing against the name of the URLs, we don't need this function checkIfURLPresentInConfig(). Please create a map from localURLList and then check if the current url name is present in the map or not. That way we can avoid the extra loops in checkIfURLPresentInConfig().

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+		componentName, err := GetComponentName()
+		if err != nil {
+			return err
+		}
+		ingressList, err := urlpkg.ListPushedIngress(kClient, componentName)
+		if err != nil {
+			return err
+		}
+		localURLList := envSpecificInfo.GetURL()
+		for _, i := range ingressList.Items {
```
mik-dass: please use a more descriptive variable name than `i`

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+			}
+		}
+		return nil
+	} else {
+		urlList, err := urlpkg.ListPushed(client, componentConfig.GetName(), componentConfig.GetApplication())
+		if err != nil {
+			return err
+		}
+		localURLList := componentConfig.GetURL()
+		for _, u := range urlList.Items {
```
mik-dass: please use a more descriptive name than `u`

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+			if !checkIfURLPresentInConfig(checkIfURLPresentInConfigParam{localURL: localURLList, url: u.Name}) {
```
mik-dass: same as https://github.com/openshift/odo/pull/2762/files#r399131010

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+		return nil
 	}
-	return nil
 }
 
-func checkIfURLPresentInConfig(localURL []config.ConfigURL, url string) bool {
-	for _, u := range localURL {
-		if u.Name == url {
-			return true
+func checkIfURLPresentInConfig(checkIfURLPresentInConfigParam checkIfURLPresentInConfigParam) bool {
```
mik-dass: https://github.com/openshift/odo/pull/2762/files#r399131010 we can avoid this

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+			}
 		}
 	}
+
 	return false
 }
 
-// ApplyConfigCreateURL applies url config onto component
-func ApplyConfigCreateURL(client *occlient.Client, componentConfig config.LocalConfigInfo) error {
+type checkIfURLPresentInConfigParam struct {
```
mik-dass: https://github.com/openshift/odo/pull/2762/files#r399131010 we can avoid this

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
 		if err != nil {
-			return errors.Wrapf(err, "unable to check url")
+			return errors.Wrapf(err, "unable to get componentName")
 		}
-		if exist {
-			log.Successf("URL %s already exists", urlo.Name)
-		} else {
-			host, err := urlpkg.Create(client, urlo.Name, urlo.Port, urlo.Secure, componentConfig.GetName(), componentConfig.GetApplication())
+		for _, urlo := range urls {
+			exist, err := urlpkg.Exists(client, kClient, urlo.Name, componentName, "")
```
mik-dass: Since Exists calls ListPushed, this might end up calling list multiple times. I would suggest call ListPushed once before the function and once use the output here.

pkg/component/component.go
```diff
@@ -564,51 +572,110 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
+				if err != nil {
+					return errors.Wrapf(err, "unable to create url")
+				}
+				log.Successf("URL %s: %s created", urlo.Name, host)
+			}
+		}
+	} else {
+		urls := componentConfig.GetURL()
+		for _, urlo := range urls {
+			exist, err := urlpkg.Exists(client, kClient, urlo.Name, componentConfig.GetName(), componentConfig.GetApplication())
```
mik-dass: same as https://github.com/openshift/odo/pull/2762/files#r399140382

pkg/component/component.go
```diff
@@ -1462,19 +1529,47 @@ func getStorageFromConfig(localConfig *config.LocalConfigInfo) storage.StorageLi
 	return storageList
 }
 
+// GetComponentName gets the component name from current directory name
+func GetComponentName() (string, error) {
```
mik-dass: we already have a similar function at https://github.com/openshift/odo/blob/a8a5eca39b80abf693bbe1e5e98880f9f31bf34a/pkg/odo/cli/component/devfile.go#L87.

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// initialising the fakeclient
+			fkclient, fkclientset := FakeNew()
+			fkclient.Namespace = "default"
+
+			objectMeta := CreateObjectMeta(tt.ingressName, "default", nil, nil)
+
+			fkclientset.Kubernetes.PrependReactor("create", "ingresses", func(action ktesting.Action) (bool, runtime.Object, error) {
```
mik-dass: we don't need a prependReactor in case of create. To replicate the scenario of empty ingress name, check and throw the error from the create function itself. That way we can save a API call.

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+	extensionsv1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	ktesting "k8s.io/client-go/testing"
+)
+
+func TestCreateIngress(t *testing.T) {
+
+	tests := []struct {
```
mik-dass: since the function takes label selector as an input, the struct should at least contain the below fields

```
type args struct {
		labelSelector string
}
tests := []struct {
                args    args
		wantIngress   []extensionsv1.Ingress
		wantErr bool
}
```
the ingressName should be not be used and should instead be a part of ingress as the length of the returned list can be greater than 1

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+					}
+				}
+			}
+
+		})
+	}
+}
+
+func TestListIngresses(t *testing.T) {
+	tests := []struct {
```
mik-dass: since the function takes label selector as an input, the struct should at least contain the below fields

```
type args struct {
		labelSelector string
}
tests := []struct {
                args    args
		wantIngress   []extensionsv1.Ingress
		wantErr bool
}
```
the ingressName should be not be used and should instead be a part of ingress as the length of the returned list can be greater than 1

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// initialising the fakeclient
+			fkclient, fkclientset := FakeNew()
+			fkclient.Namespace = "default"
+
+			fkclientset.Kubernetes.PrependReactor("list", "ingresses", func(action ktesting.Action) (bool, runtime.Object, error) {
+				if tt.ingressName == "" {
```
mik-dass: this is not a valid check as the ingressName is not the input to the list function. The check should validate the labelSelector like 
```
if tt.args.labelSelector == action.(ktesting.ListAction).GetListRestrictions().Labels {
					return true, nil, errors.Errorf("selectors are different")
}
```

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+							ObjectMeta: metav1.ObjectMeta{
+								Name: tt.ingressName,
+							},
+						},
+					},
+				}
+				return true, &ingress, nil
+			})
+
+			ingresses, err := fkclient.ListIngresses("")
```
mik-dass: the input shouldn't be `""`. Once https://github.com/openshift/odo/pull/2762/files#r399159723 is implemneted, the input should be `tt.args.labelSelector`

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+		ingressName string
+		wantErr     bool
+	}{
+		{
+			name:        "Case: Valid ingress name",
+			ingressName: "testIngress",
+			wantErr:     false,
+		},
+		{
+			name:        "Case: Invalid ingress name",
```
mik-dass: add a case when more than two ingresses are returned

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+func TestDeleteIngress(t *testing.T) {
+
+	tests := []struct {
+		name        string
+		ingressName string
+		wantErr     bool
+	}{
+		{
+			name:        "delet test",
```
mik-dass: `delete`

pkg/kclient/ingress_test.go
```diff
@@ -0,0 +1,236 @@
...
+				}
+				ingress := extensionsv1.Ingress{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: tt.ingressName,
+					},
+				}
+				return true, &ingress, nil
+			})
+
+			ingress, err := fkclient.GetIngress("")
```
mik-dass: this should be GetIngress(tt.ingressName)

pkg/kclient/secrets.go
```diff
@@ -15,8 +13,7 @@ import (
 // portNumber is the target port of the ingress
 func (c *Client) CreateTLSSecret(tlsCertificate []byte, tlsPrivKey []byte, componentName string, applicationName string, portNumber int) (*corev1.Secret, error) {
 	labels := componentlabels.GetLabels(componentName, applicationName, true)
-	portAsString := fmt.Sprintf("%v", portNumber)
```
mik-dass: why is port number removed? If it not required anymore please remove the portNumber from the function signature and also the function comment

pkg/kclient/secrets_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// initialising the fakeclient
+			fkclient, fkclientset := FakeNew()
+			fkclient.Namespace = "default"
+
+			fkclientset.Kubernetes.PrependReactor("create", "secrets", func(action ktesting.Action) (bool, runtime.Object, error) {
+				if tt.secretName == "" {
```
mik-dass: we can implement this empty check in the function itself. That way we save a API call and remove the create prependReactor entirely.

pkg/odo/cli/component/common_push.go
```diff
@@ -185,6 +184,12 @@ func (cpo *CommonPushOptions) Push() (err error) {
 	cmpName := cpo.LocalConfigInfo.GetName()
 	appName := cpo.LocalConfigInfo.GetApplication()
 
+	if cpo.Context == nil {
+		fmt.Println("normal push Context is nil!!!!!")
+	} else {
+		fmt.Println("hihihihi    " + cpo.Context.Client.Namespace)
```
mik-dass: is this even required :D

pkg/odo/cli/component/common_push.go
```diff
@@ -185,6 +184,12 @@ func (cpo *CommonPushOptions) Push() (err error) {
 	cmpName := cpo.LocalConfigInfo.GetName()
 	appName := cpo.LocalConfigInfo.GetApplication()
 
+	if cpo.Context == nil {
+		fmt.Println("normal push Context is nil!!!!!")
```
mik-dass: please use glog for logging

pkg/odo/cli/url/create.go
```diff
@@ -60,17 +64,51 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
...
+		var postList []string
+		for _, c := range containers {
+			if len(c.Ports) != 0 {
+				compWithEndpoint++
+				for _, port := range c.Ports {
+					postList = append(postList, strconv.FormatInt(int64(port.ContainerPort), 10))
+				}
+			}
+			if compWithEndpoint > 1 {
+				return fmt.Errorf("Devfile should only have 1 component contains endpoint")
```
mik-dass: `only have 1 component containing a endpoint`

pkg/odo/cli/url/create.go
```diff
@@ -60,17 +64,51 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
...
+			}
+		}
+		if compWithEndpoint == 0 {
+			return fmt.Errorf("No valid component with endpoint found in the devfile")
```
mik-dass: an endpoint

pkg/odo/cli/url/create.go
```diff
@@ -60,17 +64,51 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
...
+		}
+		componentName, _ := component.GetComponentName()
```
mik-dass: please don't ignore the err here

pkg/odo/cli/url/create.go
```diff
@@ -142,21 +180,26 @@ func (o *URLCreateOptions) Validate() (err error) {
...
 	} else {
 		err = o.LocalConfigInfo.SetConfiguration("url", config.ConfigURL{Name: o.urlName, Port: o.componentPort, Secure: o.secureURL})
 	}
 	if err != nil {
 		return errors.Wrapf(err, "failed to persist the component settings to config file")
 	}
-	if experimental.IsExperimentalModeEnabled() {
-		log.Successf("URL created for component: %v, cluster host: %v", o.Component(), o.clusterHost)
+	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
+		componentName, _ := component.GetComponentName()
```
mik-dass: please don't ignore the error

pkg/odo/genericclioptions/context.go
```diff
@@ -110,6 +120,25 @@ func getFirstChildOfCommand(command *cobra.Command) *cobra.Command {
 	return nil
 }
 
+func getValidEnvinfo(command *cobra.Command) (*envinfo.EnvSpecificInfo, error) {
+	// Get details from the local config file
```
mik-dass: from the env folder

pkg/odo/cli/url/delete.go
```diff
@@ -38,69 +40,97 @@ func NewURLDeleteOptions() *URLDeleteOptions {
...
-		if o.now {
-			err = o.Push()
+		if experimental.IsExperimentalModeEnabled() {
+			err = o.EnvSpecificInfo.DeleteURL(o.urlName)
 			if err != nil {
-				return errors.Wrap(err, "failed to push changes")
+				return err
 			}
-		} else {
+			log.Successf("URL %s removed from the config file", o.urlName)
```
mik-dass: from the env file

pkg/url/url.go
```diff
@@ -191,11 +283,40 @@ func List(client *occlient.Client, localConfig *config.LocalConfigInfo, componen
 	return urlList, nil
 }
 
-func getProtocol(route routev1.Route) string {
-	if route.Spec.TLS != nil {
-		return "https"
+// ListIngress returns all Ingress URLs for given component.
+func ListIngress(client *kclient.Client, envSpecificInfo *envinfo.EnvSpecificInfo, componentName string) (iextensionsv1.IngressList, error) {
```
mik-dass: Why are there two list functions with different names? They seem to be doing the same thing and envSpecificInfo is unused in this function.

pkg/url/url.go
```diff
@@ -191,11 +283,40 @@ func List(client *occlient.Client, localConfig *config.LocalConfigInfo, componen
...
+	return urlList, nil
+}
+
+// GetProtocol returns the protocol string
+func GetProtocol(route routev1.Route, ingress iextensionsv1.Ingress) string {
+	if experimental.IsExperimentalModeEnabled() {
+		if ingress.Spec.TLS != nil {
+			return "https"
+		}
+		return "http"
```
mik-dass: since "http" is returned in the else block too, put it at the end of the function

pkg/url/url.go
```diff
@@ -216,25 +337,42 @@ func ConvertConfigURL(configURL config.ConfigURL) URL {
...
 // Exists checks if the url exists in the component or not
 // urlName is the name of the url for checking
 // componentName is the name of the component to which the url's existence is checked
 // applicationName is the name of the application to which the url's existence is checked
-func Exists(client *occlient.Client, urlName string, componentName string, applicationName string) (bool, error) {
-	urls, err := ListPushed(client, componentName, applicationName)
-	if err != nil {
-		return false, errors.Wrap(err, "unable to list the urls")
-	}
+func Exists(client *occlient.Client, kClient *kclient.Client, urlName string, componentName string, applicationName string) (bool, error) {
```
mik-dass: Please refactor the function and use the get functions instead of the List ones.

pkg/url/url.go
```diff
@@ -279,12 +417,35 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
 	return portNumber, fmt.Errorf("given port %d is not exposed on given component, available ports are: %s", portNumber, strings.Trim(strings.Replace(fmt.Sprint(componentPorts), " ", ",", -1), "[]"))
 }
 
+// // GetComponentServicePortNumbers returns the port numbers exposed by the service of the component
```
mik-dass: comments found

tests/helper/helper_oc.go
```diff
@@ -457,3 +457,11 @@ func (oc *OcRunner) CheckForExistence(resourceType, namespace string) {
 	output := string(session.Wait().Out.Contents())
 	Expect(strings.ToLower(output)).To(ContainSubstring(""))
 }
+
+// GetIngress gets ingress list on the cluster
```
mik-dass: comments found

tests/integration/cmd_devfile_url_test.go
```diff
@@ -0,0 +1,126 @@
...
+		helper.DeleteDir(context)
+		os.Unsetenv("GLOBALODOCONFIG")
+	})
+
+	Context("Listing urls", func() {
+		It("should list url after push", func() {
+			var stdout string
+			url1 := helper.RandString(5)
+			host := helper.RandString(5) + ".com"
+			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
```
mik-dass: can we move this in the beforeEach?

pkg/odo/cli/url/describe.go
```diff
@@ -55,29 +57,45 @@ func (o *URLDescribeOptions) Validate() (err error) {
 
 // Run contains the logic for the odo url list command
 func (o *URLDescribeOptions) Run() (err error) {
+	if experimental.IsExperimentalModeEnabled() {
+		u, err := url.GetIngress(o.KClient, o.EnvSpecificInfo, o.url)
```
mik-dass: don't we need state sync like below here?

cdrage:
First all, awesome description of the PR, really really goes into detail which is awesome.

I'll do a review shortly, but what do you think about changing the `--secured` parameter? For some reason, it seems out of place. What about `--https` or `--tls` instead?

What do you think @kadel @girishramnani @dharmit 

I need someone else's opinion on this.

pkg/odo/cli/url/describe.go
yangcao77: This PR is for the basic usage of odo orl devfile support. I've noticed the missing of the `state`. I've create another issue to track that: https://github.com/openshift/odo/issues/2787

pkg/url/url.go
yangcao77: I was planning to use `ListIngress` to list all local ingresses and pushed ingresses, also add state to each ingress found. 
Then found out it might won't be a small change, need to rethink and rewrite some code. 
Delete the `ListIngress` for now. Added a `TODO` comment
The work for it will be in https://github.com/openshift/odo/issues/2787

pkg/odo/cli/url/create.go
yangcao77: was a workaround. Now getting the name from envinfo.

pkg/odo/cli/url/create.go
yangcao77: was a workaround. Now getting the name from envinfo.

pkg/url/url.go
yangcao77: updated.

pkg/odo/cli/url/create.go
yangcao77: updated.

pkg/odo/cli/url/create.go

pkg/odo/cli/component/common_push.go
yangcao77: Was a debug message. Removed.

pkg/odo/cli/component/common_push.go
yangcao77: Was a debug message. Removed.

pkg/kclient/secrets_test.go
yangcao77: updated

pkg/kclient/secrets.go
yangcao77: updated

pkg/kclient/ingress_test.go
yangcao77: updated the testcase.

pkg/kclient/ingress_test.go
yangcao77: updated

kadel:
> I'll do a review shortly, but what do you think about changing the `--secured` parameter? For some reason, it seems out of place. What about `--https` or `--tls` instead?

To be honest, I don't know :-) I don't have anything against `--secured`, but `--https` or `--tls` also makes sense. One small benefit in something more generic like `--secured` can be that it doesn't specify any protocol, so in future, it can be used for something else. But not sure if there will ever by something else :-)

/retest

kadel:
On OpenShift cluster this creates both Ingress and Route with the same host so it doesn't work due to conflicts in hostnames. 
Only Ingress or Route should be created for URL, but not both. 

For now, we can stick just with Ingress when url is created for Devfile component.
Routes can be handled later as described in this issue https://github.com/openshift/odo/issues/2727

pkg/component/component.go
yangcao77: IMO, it's better to call `Exist` to list current pushed ingresses/routes everytime before creation. I don't know if concurrent creation is possible; if it is, my concern is that.
What if two users connecting to the same cluster and same namespace, with same component (Let's say they cloned from their git repo), has same `urlName` but with different url definition in their `env.yaml`, try to create url almost at the same time? get a URL list earlier and pass into the function before the creation loop would have an issue.

pkg/url/url.go
mik-dass: OK :+1:

pkg/component/component.go
mik-dass: > What if two users connecting to the same cluster and same namespace, with same component (Let's say they cloned from their git repo), has same urlName but with different url definition in their env.yaml, try to create url almost at the same time?

In such a race condition, I would expect the creation of later one to fail.

> get a URL list earlier and pass into the function before the creation loop would have an issue. Since a cluster can establish multiple connections by different users, it's not guaranteed there won't be any identical urlName being created in between.

I don't think identical urlNames will be a problem as the creation of one of them would simply fail in such a case. Or am I missing something?

Plus this is a extremely rare use case scenario. And even if it occurs, the cluster should handle this race condition by itself.

pkg/odo/cli/url/describe.go
mik-dass: OK :+1:

pkg/component/component.go
```diff
@@ -564,51 +572,80 @@ func ApplyConfig(client *occlient.Client, componentConfig config.LocalConfigInfo
...
-	localURLList := componentConfig.GetURL()
-	for _, u := range urlList.Items {
-		if !checkIfURLPresentInConfig(localURLList, u.Name) {
-			err = urlpkg.Delete(client, u.Name, componentConfig.GetApplication())
-			if err != nil {
-				return err
+func applyConfigDeleteURL(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo) (err error) {
+	if experimental.IsExperimentalModeEnabled() {
+		componentName := envSpecificInfo.GetName()
+		ingressList, err := urlpkg.ListPushedIngress(kClient, componentName)
```
mik-dass, line 578: I don't think we need it. We can do something like what you did in ApplyConfigCreateURL(). Just pass the `pushedURLMap` here and check against it. Or am I missing something?

pkg/component/component.go
```diff
@@ -1464,20 +1501,41 @@ func getStorageFromConfig(localConfig *config.LocalConfigInfo) storage.StorageLi
...
+		urlList, err := urlpkg.ListPushedIngress(kClient, componentName)
+		if err != nil {
+			return false, nil, err
+		}
 
-	urlList, err := urlpkg.ListPushed(client, componentConfig.GetName(), componentConfig.GetApplication())
-	if err != nil {
-		return false, err
-	}
+		// If config has URL(s) (since we check) or if the cluster has URL's but
```
mik-dass, line 1512: config -> env

pkg/component/component.go
```diff
@@ -1464,20 +1501,41 @@ func getStorageFromConfig(localConfig *config.LocalConfigInfo) storage.StorageLi
...
+		// componentConfig does not (deleting)
```
mik-dass, line 1513: componentEnv

pkg/odo/cli/component/create.go
```diff
@@ -585,8 +585,8 @@ func (co *CreateOptions) Validate() (err error) {
 			spinner := log.Spinner("Validating component")
 			defer spinner.End(false)
 
-			if util.CheckPathExists(DevfilePath) || util.CheckPathExists(EnvFilePath) {
```
mik-dass, line 588: shouldn't we have a check for devfile too? Or am I missing something?

pkg/odo/cli/component/push.go
```diff
@@ -35,7 +37,7 @@ type PushOptions struct {
 	*CommonPushOptions
 
 	// devfile path
-	devfilePath string
+	DevfilePath string
```
mik-dass, line 40: why was this converted to a public variable?

pkg/odo/cli/url/create.go
```diff
@@ -182,10 +224,11 @@ func NewCmdURLCreate(name, fullName string) *cobra.Command {
...
 	urlCreateCmd.Flags().IntVarP(&o.urlPort, "port", "", -1, "port number for the url of the component, required in case of components which expose more than one service port")
 	urlCreateCmd.Flags().BoolVarP(&o.secureURL, "secure", "", false, "creates a secure https url")
-	// if experimental mode is enabled, add more flags to support ingress creation
+	// if experimental mode is enabled, add more flags to support ingress creation based on devfile
 	if experimental.IsExperimentalModeEnabled() {
-		urlCreateCmd.Flags().StringVar(&o.tlsSecret, "tlsSecret", "", "tls secret name for the url of the component if the user bring his own tls secret")
-		urlCreateCmd.Flags().StringVar(&o.clusterHost, "clusterHost", "", "Cluster host for this URL")
+		urlCreateCmd.Flags().StringVar(&o.tlsSecret, "tls-secret", "", "tls secret name for the url of the component if the user bring his own tls secret")
+		urlCreateCmd.Flags().StringVarP(&o.host, "host", "", "", "Cluster ip for this URL")
+		urlCreateCmd.Flags().StringVar(&o.DevfilePath, "devfile", "./devfile.yaml", "Path to a devfile.yaml")
```
mik-dass, line 231: This seems to use the Devfile path from https://github.com/openshift/odo/pull/2762/files#diff-6e3a2602ffd394e9a73d4fbf1388be8dR40. I think we should have it in URLCreateOptions itself.

pkg/kclient/secrets_test.go
mik-dass: you forgot to update this

kadel:
> @kadel I think that's how ingress works on Openshift cluster.
> Even if you do it through command line `kubectl apply -f your-ingress.yaml`. It would create a route at the same time. It's also mentioned in openshift blog (https://blog.openshift.com/kubernetes-ingress-vs-openshift-route/)
> 
> ```
> If you list the ingress and route objects in the project, you’d see that there is a route object generated for the ingress object that we created in order to expose the service and is kept in sync by the ingress controller.
> ```
> 
> If I'm understanding this correctly, it has nothing to do with this PR, that's just how ingress works in Openshift, the route created is to serve the ingress, and won't have any conflicts.


You are right. Sorry. I should not do rush reviews at evening 😇 
Ingress created by odo is for some reason now working for me on OpenShift :-/ 
I'll try to debug it it might be something on my side.

kadel:
I figured out why it is not working for me.

I was testing this on the component that was already created and pushed using code from https://github.com/openshift/odo/pull/2735
It looks like the code form PR #2735 is not respecting the `Name` field from `env.yaml` so it created `Service` with a different name. When I executed `url create` and `push` with this PR it created a new `Service` matching `Name` in `env.yaml`. 

This is an issue with PR #2735 (it is not up to date with the current master), not with this PR

Sorry for confusion

pkg/odo/cli/component/push.go
yangcao77: Because if there is a `now` flag provided, we need to do run `devfilePush`, which requires existance of a devfilePath in the `PushOptions`.

pkg/odo/cli/url/create.go
yangcao77: I have raised up the same concern during our team meeting, but the decision is that we do not save the devfilePath into `env.yaml` during create. 
Since the default devfile path should be `./devfile.yaml`, if the user brings their own, it is also encourage to put it under the project directory. They need to specify the path if they put that file elsewhere.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -70,6 +70,8 @@ var _ = Describe("odo devfile push command tests", func() {
 			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs"), context)
+			cmpName := helper.RandString(6)
+			helper.CmdShouldPass("odo", "create", "nodejs", "--context", context, "--namespace", namespace, cmpName)
```
GeekArthur, line 74: This command shouldn't pass based on the design https://github.com/openshift/odo/issues/2789 we discussed before.

```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
```
If you want to specify the component type, you have to delete the existing devfile.yaml as component type is used for downloading the devfile.yaml. Or you can run odo create without component type if you want to leave the devfile.yaml

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -88,14 +90,15 @@ var _ = Describe("odo devfile push command tests", func() {
...
```
GeekArthur, line 94: This command shouldn't pass based on the design https://github.com/openshift/odo/issues/2789 we discussed before.

```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
```
If you want to specify the component type, you have to delete the existing devfile.yaml as component type is used for downloading the devfile.yaml. Or you can run odo create without component type if you want to leave the devfile.yaml

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -104,6 +107,8 @@ var _ = Describe("odo devfile push command tests", func() {
 
 			// Create a new file that we plan on deleting later...
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs"), context)
+			cmpName := helper.RandString(6)
+			helper.CmdShouldPass("odo", "create", "nodejs", "--context", context, "--namespace", namespace, cmpName)
```
GeekArthur, line 111: This command shouldn't pass based on the design https://github.com/openshift/odo/issues/2789 we discussed before.

```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
```
If you want to specify the component type, you have to delete the existing devfile.yaml as component type is used for downloading the devfile.yaml. Or you can run odo create without component type if you want to leave the devfile.yaml

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -143,11 +145,10 @@ var _ = Describe("odo devfile push command tests", func() {
 			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs-multicontainer"), context)
+			cmpName := helper.RandString(6)
+			helper.CmdShouldPass("odo", "create", "nodejs", "--context", context, "--namespace", namespace, cmpName)
```
GeekArthur, line 149: This command shouldn't pass based on the design https://github.com/openshift/odo/issues/2789 we discussed before.

```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
```
If you want to specify the component type, you have to delete the existing devfile.yaml as component type is used for downloading the devfile.yaml. Or you can run odo create without component type if you want to leave the devfile.yaml

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -183,6 +184,8 @@ var _ = Describe("odo devfile push command tests", func() {
...
```
GeekArthur, line 188: This command shouldn't pass based on the design https://github.com/openshift/odo/issues/2789 we discussed before.

```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
```
If you want to specify the component type, you have to delete the existing devfile.yaml as component type is used for downloading the devfile.yaml. Or you can run odo create without component type if you want to leave the devfile.yaml

pkg/odo/cli/component/create.go
```diff
@@ -585,8 +585,8 @@ func (co *CreateOptions) Validate() (err error) {
...
-				return errors.New("Devfile.yaml or env.yaml already exists")
+			if util.CheckPathExists(EnvFilePath) {
```
GeekArthur, line 588: I still think if we remove this check we should handle the following two cases properly at the same time:
```
If devfile.yaml is present
Case 1: odo create <component type> -> Throw errors and tell user to delete the existing devfile.yaml then run odo create again
Case 2: odo create -> Use the existing devfile.yaml
```
Just simply remove this check can cause more issues such as for the current code, if user has nodejs devfile.yaml in the workspace directory but user runs `odo create maven`, it can still pass but that's not expected

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: This PR is not going to handle the changes to the create command. The create validation should be handled by your PR.

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: same as: https://github.com/openshift/odo/pull/2762#discussion_r401864713

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: same as: https://github.com/openshift/odo/pull/2762#discussion_r401864713

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: same as: https://github.com/openshift/odo/pull/2762#discussion_r401864713

pkg/odo/cli/component/create.go
yangcao77: same as: https://github.com/openshift/odo/pull/2762#discussion_r401864713

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: same as: https://github.com/openshift/odo/pull/2762#discussion_r401864713

tests/integration/devfile/cmd_devfile_push_test.go
GeekArthur: I'm not sure you get my point, It doesn't matter whose PR handles this issue, it matters your test case is WRONG here based on the design (I mean your test case should detect the errors instead of letting it pass), if you decides to remove the devfile.yaml existence check here https://github.com/openshift/odo/pull/2762/files#diff-58fe314b4e3f9120402717ab8c2b3049R588, you have to handle the corresponding behaviors correctly. Or you shouldn't remove that and wait for my PR to handle that, it totally depends on you.

The key point is with you changes, it can causes the regression of this use case below:
```
If user has nodejs devfile.yaml in the workspace directory but user runs odo create maven, it can still pass but that's not expected
```

pkg/odo/cli/component/create.go
GeekArthur: Actually... this is another new comment... not same as above test cases comments.... 🤦‍♂ It's about one use case this PR may break...and you should test that use case

tests/integration/devfile/cmd_devfile_push_test.go
GeekArthur: Also keep in mind by design currently there are no required/mandatory arguments, we have interactive mode handles the use case.

yangcao77: As we discussed in the meeting, the validation should be handled in a separate PR. 
The `componentType` is a required argument in master branch. Since the validation is missing, you cannot expect the test case to meet the behavior like that. 
But I don't mind take that issue and doing the whole validation in this PR.  You can assign the new issue to me, I will do the entire validation in the PR.

pkg/odo/cli/component/create.go
yangcao77: Actually I remember I should only handle Case2, and also the issue you created says that: https://github.com/openshift/odo/issues/2789
In addition, do think all the validation should be done within the same PR. 
If you want me to handle case1 listed in the PR, I don't mind added that in the PR. But I think it makes more sense to also cover `If devfile not exist` case as well. 

Please close your PR, and I will checkout your changes.

yangcao77: Also the new use case test should be under `devfile-create` test. not the `devfile-push` or `devfile-url` I have modified in the PR.  And should be added within the same PR of devfile existence validation in `odo create`

Again, I don't mind work on that as well.

tests/integration/devfile/cmd_devfile_push_test.go
GeekArthur: We do have interactive mode, which means even for today's master branch component type is also not required (same design as odo1), you can give a try. 

But anyway I already opened a PR https://github.com/openshift/odo/pull/2798 to cover all cases of https://github.com/openshift/odo/issues/2789, and already asked @kadel review that, you can talk with him if you want to take over that

pkg/odo/cli/component/create.go
GeekArthur: This is my PR https://github.com/openshift/odo/pull/2798, if you want to take over that you can talk with @kadel to avoid confusion since I already asked him review and he may have more comments against the issue/PR

pkg/odo/cli/url/create.go
yangcao77: I posted this concern again in the teem, and waiting for other's thoughts. 
I still think saving devfilePath in `env.yaml` during creation makes more sense. Since create cmd is required before other commands being executed. So for any following command (push, url create, etc.), we can directly get it from the env.yaml file. 

If the user wants to use another devfilePath(i.e. change the value of devfilePath), we can also allow those commands which accept the `--devfile` to overwrite the value within `env.yaml`.
@elsony @kadel @johnmcollier Any thought?

(But anyways, if we are going to do that, I would prefer to have a separate issue and handle that in a separate PR. I would say leave the `DevfilePath` exposed as it is for now.)

pkg/kclient/secrets_test.go
yangcao77: Missed this one. 
Actually the `secretName` would never be empty, since it's `componentName + "-tlssecret"`. So I checked for invalid componentName instead.

tests/integration/devfile/cmd_devfile_push_test.go
yangcao77: Just tried, if would fail without specifying any componentType:
```
 ✗  Checking if the specified component type is supported devfile component type [99789ns]

Please run 'odo catalog list components' for a list of supported devfile component types
```

pkg/odo/cli/component/create.go
yangcao77: Since you have already have all that implemented, I think it's better to keep it in your PR. 
You should also add the user case test in `devfile-create`

It makes more sense to leave this PR as it is, and we can modify the test code to match the expected behavior in no matter this PR or your PR. (As long as one of our PR merges, the test code would fail, we need to modify the test anyways.)

mik-dass:
@yangcao77 `odo url create` seems to be failing with `--now` flag

```
[mrinaldas@localhost nodejs]$ odo url create example --host com  --now
 ✗  Checking component [33749ns]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x78 pc=0x168902b]

goroutine 1 [running]:
github.com/openshift/odo/pkg/config.(*LocalConfig).GetName(...)
        /home/mrinaldas/go/src/github.com/openshift/odo/pkg/config/config.go:470
github.com/openshift/odo/pkg/odo/cli/component.(*CommonPushOptions).ValidateComponentCreate(0xc00058f490, 0x0, 0x0)
        /home/mrinaldas/go/src/github.com/openshift/odo/pkg/odo/cli/component/common_push.go:89 +0x9b
github.com/openshift/odo/pkg/odo/cli/url.(*URLCreateOptions).Validate(0xc000410d80, 0x0, 0x0)
        /home/mrinaldas/go/src/github.com/openshift/odo/pkg/odo/cli/url/create.go:172 +0x109
github.com/openshift/odo/pkg/odo/genericclioptions.GenericRun(0x1f59860, 0xc000410d80, 0xc00040d200, 0xc00005d5c0, 0x1, 0x4)
```
/hold

pkg/odo/cli/component/push.go
mik-dass: OK

pkg/odo/cli/url/create.go
mik-dass: OK

yangcao77:
@mik-dass Thanks for pointing this out!
I've pushed the fix for the `now` flag, and also added an integration test for it.
```
$ ./odo url create example --host com  --now
CORRECT!!!!!
 ✓  URL created for component: springtest1, cluster host: com
 •  Push devfile component springtest1  ...

Applying URL changes
 ✓  URL example: http://example.com created
 ✓  Checking file changes for pushing [10ms]
 ✓  Waiting for component to start [119ms]
```

mik-dass:
> @mik-dass Thanks for pointing this out!
I've pushed the fix for the now flag, and also added an integration test for it.

Thanks for the immediate fix :+1:  But it seems `odo url delete` is missing the `--now` implementation.

kadel:
> Thanks for the immediate fix 👍
> 
> Sorry I didn't notice it the last time, but it seems `odo url delete` is missing the `--now` implementation.

I think that we can address this in separate PR, this one is already getting quite huge

mik-dass:
> I think that we can address this in separate PR, this one is already getting quite huge

OK :+1:

mik-dass:
```
2020/04/02 07:14:08 Ran for 7m27s
2020/04/02 07:14:08 Submitted failure event to sentry (id=cc425f624edf4ac389bdcb4bd9add35c)
error: could not run steps: step integration-e2e-benchmark failed: could not wait for template instance to be ready: failed to create objects: rolebindings.authorization.openshift.io "v4.3-integration-e2e-benchmark-image-puller" already exists
pods "v4.3-integration-e2e-benchmark" already exists
```
/retest

yangcao77:
@mik-dass 
Issue created for `now` flag in `url delete`: https://github.com/openshift/odo/issues/2804
Please help unhold this PR.

yangcao77:
@mik-dass Can you help add back the lgtm label?
I  modified the test and pushed a new commit. Trying to fix the flaky test.

Result on my local: 
```
$ make test-cmd-devfile-url
ginkgo  -randomizeAllSpecs -slowSpecThreshold=120 -timeout 7200s -nodes=2 -focus="odo devfile url command tests" tests/integration/devfile/
Running Suite: Devfile Suite
============================
Random Seed: 1585885726 - Will randomize all specs
Will run 23 specs

Running in parallel across 2 nodes

S••SSSSSSSS••SSSSSSSSS•
Ran 5 of 23 Specs in 165.446 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 18 Skipped
```

tests/helper/helper_generic.go
```diff
@@ -143,3 +143,25 @@ func WatchNonRetCmdStdOut(cmdStr string, timeout time.Duration, check func(outpu
 		}
 	}
 }
+
+// WaitTillUrlexist wait for the given url to be up on the cluster
+func WaitTillUrlexist(namespace, url string, timeoutMinutes int) bool {
```
mik-dass, line 148: We don't need this function. We already have a similar generic function https://github.com/openshift/odo/blob/3587b9eac2501ffc29d54207368af591424cb0be/tests/helper/helper_generic.go#L40

mik-dass:
> I just unhold it. Thought I did not have permission to do that. :D

Lol! I was almost completely sure that you have the permission, so I left it up to you to unhold :D 

Also I just noticed that there is no integration test case for `odo url delete`. 

And deleting the deployment doesn't delete the ingress

```
[mrinaldas@localhost nodejs]$ odo push
 •  Push devfile component nodejs  ...

Applying URL changes
 ✓  URL example-1: http://example-1.1.2.3.4 created
 ✓  Checking files for pushing [990318ns]
 ✓  Waiting for component to start [2s]
 ✓  Syncing files to the component [6s]
 ✓  Push devfile component nodejs [17s]
[mrinaldas@localhost nodejs]$ oc get ingress
NAME        HOSTS               ADDRESS   PORTS     AGE
example-1   example-1.1.2.3.4             80        22s
[mrinaldas@localhost nodejs]$ oc delete deployments --all
deployment.extensions "nodejs" deleted
[mrinaldas@localhost nodejs]$ oc get ingress
NAME        HOSTS               ADDRESS   PORTS     AGE
example-1   example-1.1.2.3.4             80        41s
```
This is because the owner reference is not set for the ingress
```
[mrinaldas@localhost nodejs]$ oc get ingress -o json
{
    "apiVersion": "v1",
    "items": [
        {
            "apiVersion": "extensions/v1beta1",
            "kind": "Ingress",
            "metadata": {
                "creationTimestamp": "2020-04-03T05:45:23Z",
                "generation": 1,
                "labels": {
                    "app": "",
                    "app.kubernetes.io/instance": "nodejs",
                    "app.kubernetes.io/managed-by": "odo",
                    "app.kubernetes.io/managed-by-version": "v1.1.1",
                    "app.kubernetes.io/part-of": "",
                    "odo.openshift.io/url-name": "example-1"
                },
                "name": "example-1",
                "namespace": "bla",
                "resourceVersion": "654793819",
                "selfLink": "/apis/extensions/v1beta1/namespaces/bla/ingresses/example-1",
                "uid": "4ff1fad1-756e-11ea-b49f-0e13a02d8ebd"
            },
            "spec": {
                "rules": [
                    {
                        "host": "example-1.1.2.3.4",
                        "http": {
                            "paths": [
                                {
                                    "backend": {
                                        "serviceName": "nodejs",
                                        "servicePort": 9090
                                    },
                                    "path": "/"
                                }
                            ]
                        }
                    }
                ]
            },
            "status": {
                "loadBalancer": {}
            }
        }
    ],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```
Since this PR is already huge, please make a issue and let's address it in a separate PR.

yangcao77:
@mik-dass pushed a new commit to use the existing function to check for output.
bug created for the missing owner reference: https://github.com/openshift/odo/issues/2808

mik-dass:
@mik-dass pushed a new commit to use the existing function to check for output.

https://travis-ci.com/github/openshift/odo/jobs/313372317#L801 You forgot to add the namespace while listing the URLs

yangcao77:
> > @mik-dass pushed a new commit to use the existing function to check for output.
> 
> https://travis-ci.com/github/openshift/odo/jobs/313372317#L801
> 
> I think the error is because you forgot to add the context while listing the URLs.

I saw the namespace mismatch in the log. However, I think `odo url list` never supports `--namespace`/`--project` option.

mik-dass:
> I saw the namespace mismatch in the log. However, I think odo url list never supports --namespace/--project option. Not sure why for odo components, there are no namespace mismatch in the parallel runs

I think that is because we are using the context flag while executing every command. Thus the project/namespace gets automatically read from the context.

yangcao77:
@mik-dass I finally found out what caused the failure. 
It is because this issue: https://github.com/openshift/odo/issues/2796
`oco create` now takes `--namespace`, but does not set the value. So I switched to use `--project`.
Another failure is because `DevfilePush` requires a namespace, when using the `--now` flag to push after url creation, the namespace is empty, and would use the current namespace, which incorrect due to the parallel runs. So I get the namespace from `env.yaml`. 
Also added a `resolveNamespace()` function when create context, so if namespace is not provided, it will get it from `env.yaml`, and set `kclient.Client.Namespace` to use the correct namespace. 
 
These changes resolved the namespace mismatch in parallel runs,  and the test finally passed. (https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/2762/pull-ci-openshift-odo-master-v4.1-integration-e2e-benchmark/1540)

mik-dass:
```
@mik-dass I finally found out what caused the failure.
It is because this issue: #2796
```
OK :+1:  let's get this in once the CI passes.

PR #2786, opened 2020-03-31
Adds disclaimer that experimental mode is enabled

cdrage:
**What type of PR is this?**

/kind documentation

Adds a disclaimer that experimental mode is enabled when using `odo
create` or `odo push`.

See asciinema:

[asciicast](https://asciinema.org/a/A6ITI1wNQL9hahhLsCz4DT6wQ)

N/A

```sh
odo create nodejs
odo push
```

### DIFF
--- a/pkg/log/status.go
+++ b/pkg/log/status.go
@@ -197,6 +197,14 @@ func Swarningf(format string, a ...interface{}) string {
 	return fmt.Sprintf(" %s%s%s", yellow(getWarningString()), suffixSpacing, fmt.Sprintf(format, a...))
 }
 
+// Experimental will output in an appropriate "progress" manner
+func Experimental(a ...interface{}) {
+	yellow := color.New(color.FgYellow).SprintFunc()
+	if !IsJSON() {
+		fmt.Fprintf(GetStdout(), "%s\n", yellow(fmt.Sprintln(a...)))
+	}
+}
+
 // Warning will output in an appropriate "progress" manner
 func Warning(a ...interface{}) {
 	yellow := color.New(color.FgYellow).SprintFunc()
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -285,7 +285,12 @@ func (co *CreateOptions) setResourceLimits() error {
 
 // Complete completes create args
 func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+
 	if experimental.IsExperimentalModeEnabled() {
+
+		// Add a disclaimer that we are in *experimental mode*
+		log.Experimental("Experimental mode is enabled, use at your own risk")
+
 		if len(args) == 0 {
 			co.interactive = true
 		}
@@ -361,10 +366,13 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		co.devfileMetadata.componentName = strings.ToLower(componentName)
 		co.devfileMetadata.componentNamespace = componentNamespace
 
+		// Categorize the sections
+		log.Info("Validation")
+
 		// Since we need to support both devfile and s2i, so we have to check if the component type is
 		// supported by devfile, if it is supported we return, but if it is not supported we still need to
 		// run all codes related with s2i
-		spinner := log.Spinner("Checking if the specified component type is supported devfile component type")
+		spinner := log.Spinner("Checking Devfile compatibility")
 
 		for _, devfileComponent := range catalogDevfileList.Items {
 			if co.devfileMetadata.componentType == devfileComponent.Name && devfileComponent.Support {
@@ -579,10 +587,11 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 
 // Validate validates the create parameters
 func (co *CreateOptions) Validate() (err error) {
+
 	if experimental.IsExperimentalModeEnabled() {
 		if co.devfileMetadata.devfileSupport {
 			// Validate if the devfile component that user wants to create already exists
-			spinner := log.Spinner("Validating component")
+			spinner := log.Spinner("Validating Devfile")
 			defer spinner.End(false)
 
 			if util.CheckPathExists(EnvFilePath) {
@@ -599,6 +608,8 @@ func (co *CreateOptions) Validate() (err error) {
 			return nil
 		}
 	}
+
+	log.Info("Validation")
 
 	supported, err := catalog.IsComponentTypeSupported(co.Context.Client, *co.componentSettings.Type)
 	if err != nil {
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -65,6 +65,25 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(output).To(ContainSubstring(cmpName))
 		})
 
+		It("Check that the experimental warning appears for create and push", func() {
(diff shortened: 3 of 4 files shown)


amitkrout:
Platform issue 
/retest

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -44,6 +44,17 @@ var _ = Describe("odo devfile push command tests", func() {
...
 
+		It("Check that the experimental warning appears for create and push", func() {
+			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
+			helper.CopyExample(filepath.Join("source", "nodejs"), context)
+
+			// Check that it will contain the experimental mode output
+			experimentalOutputMsg := "Experimental mode is enabled, use at your own risk"
+			Expect(helper.CmdShouldPass("odo", "create", "nodejs")).To(ContainSubstring(experimentalOutputMsg))
+			Expect(helper.CmdShouldPass("odo", "push")).To(ContainSubstring(experimentalOutputMsg))
+
```
mik-dass, line 55: also a check that when experimental is turned off, this string is gone

pkg/log/status.go
```diff
@@ -197,6 +197,14 @@ func Swarningf(format string, a ...interface{}) string {
 	return fmt.Sprintf(" %s%s%s", yellow(getWarningString()), suffixSpacing, fmt.Sprintf(format, a...))
 }
 
+// Warning will output in an appropriate "progress" manner
+func Experimental(a ...interface{}) {
```
mik-dass, line 201: can't we just use some of the other functions for this?

mohammedzee1000:
@cdrage rebase please and please address mrinal's comments. Otherwise looks good to me

pkg/log/status.go
```diff
@@ -197,6 +197,14 @@ func Swarningf(format string, a ...interface{}) string {
 	return fmt.Sprintf(" %s%s%s", yellow(getWarningString()), suffixSpacing, fmt.Sprintf(format, a...))
 }
 
+// Warning will output in an appropriate "progress" manner
```
mik-dass, line 200: warning->Experimental

pkg/log/status.go
cdrage: Ah no, because that'll add the whole warning symbol: ⚠️ on it. Which I believe we shouldn't need / it looks ugly.

tests/integration/devfile/cmd_devfile_push_test.go
cdrage: Ah yes, I'll add that.

cdrage:
@mik-dass your comments have been addressed :+1:

pkg/log/status.go
mik-dass: TBH I think the warning symbol would look appropriate

mik-dass:
> @mik-dass your comments have been addressed +1


you forgot https://github.com/openshift/odo/pull/2786/files#r401546048 :D

pkg/log/status.go
cdrage: The reason I made this function separate is so we do not have the indenting and the warning symbol, for example:

```sh
                                                                                                                                                                                      
Validation                                                                                                                                                                            
 ✓  Checking Devfile compatibility [43501ns]                                                                                                                                          
 ✓  Validating Devfile [38035ns]     
```

vs

```sh
 ⚠️ Experimental mode is enabled, use at your own risk                                                                                                                                                                                                                                                                                                                          
Validation                                                                                                                                                                            
 ✓  Checking Devfile compatibility [43501ns]                                                                                                                                          
 ✓  Validating Devfile [38035ns]     
```

IMO the bottom one looks kind of odd without the newline. So I went with the first one (separate function).

mik-dass: Yeah the newline makes more sense, but can we still keep the warning symbol. Looks cool and approriate TBH.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -65,6 +65,27 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(output).To(ContainSubstring(componentName))
 		})
 
+		It("Check that the experimental warning appears for create and push", func() {
+			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
+			helper.CopyExample(filepath.Join("source", "nodejs"), context)
+
+			// Check that it will contain the experimental mode output
+			experimentalOutputMsg := "Experimental mode is enabled, use at your own risk"
```
amitkrout, line 73: It seems like we are exposing experimental feature putting original feature at risk. AFAIK experimental feature is 
some functionality/feature added to the original product to get more feedback from the user side and ofcoures there is a trade-off if it breaks but it should never have bad impact on the original product functionality/feature.

The message ```Experimental mode is enabled, use at your own risk``` exactly opposite what i mentioned above. Please counter my arguments 

Ping @cdrage @mik-dass @girishramnani

mik-dass: @amitkrout This message just informs the user that experimental mode is enabled and somethings may work differently. So if a user activates this mode accidentally, he gets a warning and thus can choose to stop the process.

amitkrout: Cool!!!
@mik-dass Actually recently i came across a issue #2813, so thought the message is for original feature too. You can find certain scenario where your justification contradicts. For example 

```
$ ODO_EXPERIMENTAL=true odo create nodejs --project test321 --context ../tests/examples/source/nodejs/
```

I am creating a regular nodejs component with experimental enabled.

mik-dass: > You can find certain scenario where your justification contradicts.

Within the code itself, something might work differently for experimental mode. We are just informing the user that something might go wrong if the mode is activated, not that it will. It might pass without failure too.

For example `odo push` might push using the devfile if enabled or will create a normal nodejs component if disabled.

The warning will inform him straight away that push might work differently since he is in experimental mode and certain features might be missing.

pkg/odo/cli/component/create.go
```diff
@@ -285,7 +285,12 @@ func (co *CreateOptions) setResourceLimits() error {
 
 // Complete completes create args
 func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+
 	if experimental.IsExperimentalModeEnabled() {
+
+		// Add a disclaimer that we are in *experimental mode*
+		log.Experimental("Experimental mode is enabled, use at your own risk")
```
johnmcollier, line 292: I agree with the general idea, but I'm not sure if I'm _crazy_ about always showing the message for every `odo create` and `odo push` with experimental mode enabled.

I'm not sure how feasible this would be (you tell me 😉 ), but what about only showing it for the first (or first `n`) `odo` commands used by the user once it's enabled? Or maybe only showing it when `odo preference set experimental true` is called?

cdrage: Ah, I don't think that'd be feasible for the first `n` commands.

But in terms of the output, I think it's viable since honestly, with me switching from `set experimental true` from false to true all the time testing PR's, I sometimes end up pushing a devfile instead of what I think is a regular component, since there was no output from `odo create`.

Perhaps we can output it only for `odo create` and that's it? What do you think?

johnmcollier: Yeah, having it display during `odo create` seems reasonable to me 👍

cdrage: AWesome! Updated.

PR #2788, opened 2020-03-31
Integrate odo create with odo push

### LINKED ISSUE #2734: Integrate odo create with odo push
/kind feature

Currently `odo push` still uses project source directory name as component name, `odo push` needs to get the component name from `env.yaml` (user can specify the component name by running `odo create` and the component name metadata can be stored in `env.yaml`) once this PR https://github.com/openshift/odo/pull/2699 merges in, so that the whole use case of `odo create` + `odo push` can be finished. 

Specific tasks are:
- Use the function to get component name from `env.yaml` instead of project source directory name
- Update integration tests in `cmd_devfile_push_test.go`, currently the tests copy nodejs project source (devfile.yaml + nodejs template) to temporary folder then test `odo push`. Instead of doing that, we can copy the OpenLiberty/nodejs template to temporary folder and use `odo create` to generate corresponding devfile.yaml then test `odo push`.

## Why is this needed?
This is necessary as the integration work is part of devfile support for odo feature.

GeekArthur:

**What type of PR is this?**
/kind feature

This PR is to let `odo push` uses the component name in env.yaml instead of using the root directory name as component name. With this PR, the life cycle of `odo create` + `odo push` is enabled, below is the example:

```
odo create maven                                                                                                       
 ✓  Checking if the specified component type is supported devfile component type [231090ns]
 ✓  Validating component [159410ns]

odo push                                                                                                            
 •  Push devfile component maven  ...
 ✓  Checking files for pushing [2s]
 ✓  Waiting for component to start [45ms]
 ✓  Syncing files to the component [4m]
 ✓  Push devfile component maven [4m]
```

Fixes #2734 


GeekArthur:
Close the PR since @yangcao77 's PR should cover this.

Issue #2642, opened 2020-02-26
Flaky tests for --now flag


kadel:
```
c] 'Running odo with args [odo config unset --now --env hello --context /tmp/363231666]
[odo]  •  Checking component  ...
[odo]  ✓  Checking component [66ms]
[odo]  ✓  Environment variables were successfully updated
[odo] 
[odo] Configuration changes
[odo]  •  Retrieving component data  ...
[odo]  ✓  Retrieving component data [88ms]
[odo]  •  Applying configuration  ...
[odo]  ✗  Applying configuration [86ms]
[odo]  ✗  Failed To Update Config To Component Deployed.
[odo] Error: unable to update DeploymentConfig for nodejs component: unable to update the current DeploymentConfig nodejs-app: unable to update DeploymentConfig nodejs-app: Operation cannot be fulfilled on deploymentconfigs.apps.openshift.io "nodejs-app": the object has been modified; please apply your changes to the latest version and try again
Deleting dir: /tmp/363231666
Deleting project: usaszqseld
Running odo with args [odo project delete usaszqseld -f]
[odo] This project contains the following applications, which will be deleted
[odo] Application app
[odo] This application has following components that will be deleted
[odo] component named nodejs
[odo] No services / could not get services
[odo]  •  Deleting project usaszqseld  ...
[odo]  ✓  Deleting project usaszqseld [6s]
[odo]  ✓  Deleted project : usaszqseld
• Failure [9.242 seconds]
odo preference and config command tests
/go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:20
  when using --now with config command
  /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:310
    should successfully set and unset variables [It]
    /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:322
    No future change is possible.  Bailing out early after 0.427s.
```

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/2637/pull-ci-openshift-odo-master-v4.3-integration-e2e-benchmark/728#1:build-log.txt%3A395

/priority high
/kind flake
/assign @mohammedzee1000 
/area  testing

amitkrout:
```
[odo]  ✗  Applying configuration [86ms]
[odo]  ✗  Failed To Update Config To Component Deployed.
[...]

PR #2798, opened 2020-04-01
odo create support existing devfile.yaml

### LINKED ISSUE #2789: odo create support existing devfile.yaml
/kind user-story
/area devfile

## User Story
As a user I want to use my own devfile.yaml instead of using the devfile.yaml from devfile registry, so we come up with a new design for supporting existing devfile.yaml

odo create always creates env.yaml

If devfile.yaml is not present
Case 1: `odo create <component type>` -> Download devfile from devfile egistry ✅ 
Case 2: `odo create` -> ~Throw errors and tell user to specify component type with running odo create <component type>~
UPDATE: I check the current code, the interactive mode already covered this case ✅ 

If devfile.yaml is present
Case 1: `odo create <component type>` -> Throw errors and tell user to delete the existing devfile.yaml then run odo create <component type> again 🚧 
Case 2: `odo create` -> Use the existing devfile.yaml 🚧  This case should be implemented by @yangcao77 's PR or I can implement this in my PR, either way works for me

GeekArthur:

**What type of PR is this?**
/kind feature

This PR improves the `odo create` to make it support the following use cases:

If devfile.yaml is not present
Case 1: `odo create <component type>` -> odo downloads devfile from devfile registry and creates env.yaml
Case 2: `odo create` -> User can specify component type via interactive mode, then odo downloads devfile from devfile egistry and creates env.yaml

If devfile.yaml is present
Case 1: `odo create <component type>` -> odo throws errors and tells user to delete the existing devfile.yaml then run odo create again
Case 2: `odo create` -> odo directly uses the existing devfile.yaml and creates env.yaml, user can specify component name via interactive mode

Fixes #2789 

N/A


pkg/odo/cli/component/create.go
```diff
@@ -319,7 +319,11 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 					supDevfileCatalogList = append(supDevfileCatalogList, devfileComponent)
 				}
 			}
-			componentType = ui.SelectDevfileComponentType(supDevfileCatalogList)
+
+			// devfile.yaml is not present, user has to specify the component type
+			if !util.CheckPathExists(DevfilePath) {
```
adisky, line 324: can we move this check before listing from registry, as for the case devfile is present it might save one call.

pkg/odo/cli/component/create.go
```diff
@@ -338,6 +342,10 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			// Component type: Get from full command's first argument (mandatory)
```
adisky, line 342: move the comments at the right place, above `componentType = args[0]`

pkg/odo/cli/component/create.go
```diff
@@ -319,7 +319,11 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 					supDevfileCatalogList = append(supDevfileCatalogList, devfileComponent)
 				}
 			}
-			componentType = ui.SelectDevfileComponentType(supDevfileCatalogList)
+
```
adisky, line 322: as [1] is the first check, if a `<component-type` not present and `devfile.yaml` is present, it is default going into interactive, IIUC it should create from existing `devfile.yaml`
[1]
```
if len(args) == 0 {
  co.interactive = true
	}
```

kadel:
`odo  create` should not create a devfile.yaml if there is already "old style" odo component.


In the following scenario, the second `odo create` should fail with the message that the current directory already contains a component


````
▶ export ODO_EXPERIMENTAL=false

▶ odo create nodejs
 ✓  Validating component [389ms]

▶ export ODO_EXPERIMENTAL=true

▶ odo create nodejs
 ✓  Checking if the specified component type is supported devfile component type [122381ns]
 ✓  Validating component [85245ns]

```

pkg/odo/cli/component/create.go
GeekArthur: Nice suggestion! I can do to improve the performance.

pkg/odo/cli/component/create.go
GeekArthur: Yes, it should create the component using the existing `devfile.yaml`, but user is still able to specify the `<component name>` via interactive mode, that's why it falls down to interactive mode

pkg/odo/cli/component/create.go
GeekArthur: Yup, I can refactor the comments for both interactive mode and direct mode

GeekArthur:
@kadel @adisky Comments addressed, please review again, thanks!

kadel:
/approve

@GeekArthur it needs rebase

GeekArthur:
@kadel @amitkrout comments addressed, please review again, thanks!

dharmit:
#2642 
```
 • Failure [14.541 seconds]
odo preference and config command tests
/go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:20
  when using --now with config command
  /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:342
    should successfully set and unset variables [It]
[Fail] odo preference and config command tests when using --now with config command [It] should successfully set and unset variables 
Ran 168 of 168 Specs in 3941.357 seconds
FAIL! -- 167 Passed | 1 Failed | 0 Pending | 0 Skipped
Ginkgo ran 1 suite in 1h5m49.131778396s
Test Suite Failed
make: *** [test-integration] Error 1 
```
/test v4.3-integration-e2e-benchmark

GeekArthur:
@amitkrout Integration tests are added, can you please review this PR again? Thanks

GeekArthur:
@kadel Rebasing is done, Please also review this, thanks!

maysunfaisal:
Re-adding the lgtm after rebase
/lgtm

maysunfaisal:
re-add lgtm after rebase
/lgtm

Issue #2579, opened 2020-02-07
Odo push should deploy local Docker containers when pushtarget is set to Docker


johnmcollier:
## User Story
As a user, I want to be able to create components in odo (using `odo create component`), and be able to deploy the same set of devfiles locally, as I would on Kube. This also means `odo create component` should also be identical, [regardless of the context that's being targeted](https://github.com/openshift/odo/issues/2577).

## Acceptance Criteria 
- [x] Devfiles that work on Kube with Odo, should also work locally with odo.
    - The `odo create component` command should function the same as it would on Kube
- [x] `odo push` should deploy containers on the users system via Docker
    - Each container in the devfile corresponds to a container on Docker
- [x] Unit and integration tests
    - Integration tests to only run on Travis CI

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

/kind user-story

kadel:
I don't think all this was implemented in #2806
/reopen

johnmcollier:
@kadel Initially this story covered sync and exec as well, but @elsony and I made the decision to split that functionality off into https://github.com/openshift/odo/issues/2717, so my PR that was merged covered everything that was in scope for this issue (deploying containers when pushtarget is set, and integration test platform for odo-docker). Can this issue be closed?

Issue #2717, opened 2020-03-12
Devfile sync and exec support for local docker containers


johnmcollier:
## User Story
As a user, I want to be able to sync my code changes to a Docker container running locally on my laptop. Once my code changes are synced, odo should build the source and run the application. Subsequent `odo push`'s should only sync the changed files (and remove the deleted files), then proceed to rebuild and re-run the application.

(As discussed with @elsony and team, we're breaking up https://github.com/openshift/odo/issues/2579 into two issues)

## Acceptance Criteria 
- [x] It should re-use the `sync` package in odo, and make `lclient` / `dockerclient` implement the `SyncClient` interface
- [x] It should only sync the changed files (and remove any deleted files)
- [x] Container exec for build/run should follow the same model as described in https://github.com/openshift/odo/issues/2654
- [x] Unit and tests and integration tests need to be written
   - Integration tests to only run on Travis CI

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

/kind user-story
/area devfile
/area local

kadel:
What about volumes created based on the `volumes` definition in devfile.yaml component?
I can see that with #2922 they are not deleted.

```
▶ odo push
 •  Push devfile component backend  ...
 ✓  Pulling image quay.io/eclipse/che-java11-maven:nightly [5s]
 ✓  Starting container for quay.io/eclipse/che-java11-maven:nightly [382ms]
 ✓  Pulling image busybox [3s]
 ✓  Starting container for busybox [496ms]
 ✓  Push devfile component backend [9s]
 ✓  Changes successfully pushed to component

tmp/odo-examples/demo-backend  master ✗                                                                                                                                          14d ⚑ ◒
▶ docker volume ls
DRIVER              VOLUME NAME
local               m2-backend-nwpm
local               odo-project-source-backend-bqfz

tmp/odo-examples/demo-backend  master ✗                                                                                                                                          14d ⚑ ◒
▶ odo delete
? Are you sure you want to delete the devfile component: backend? Yes
 ✓  Deleting devfile component backend [398ms]
 ✓  Successfully deleted component

tmp/odo-examples/demo-backend  master ✗                                                                                                                                          14d ⚑ ◒
▶ docker volume ls
DRIVER              VOLUME NAME
local               m2-backend-nwpm
```

elsony:
Feature done

Issue #2581, opened 2020-02-07
Use 'odo delete` to delete a component that's running locally with odo


johnmcollier:
## User Story
As a user, I will need to be able to delete the applications I've deployed and developed with locally (on Docker), just as I can on Kube right now. `odo delete <component>` should remove all of the running containers for that application, alongside any volumes or other resources that were created.

## Acceptance Criteria 
- [ ] It should clean up all of the resources deployed for that component + the component's config
- [ ] It should use the same command as on Kube
- [ ] When delete, only remove the container and leave the images around

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

/kind user-story

jgwest:
First PR has merged. Since the first PR does not include support for deleting non-source volumes (the non-source volumes PR has not been merged at time of first PR) I will be posting a second PR shortly to complete this user story.

This issue should remain open until second PR has merged.

elsony:
Code merged, feature done

Issue #2580, opened 2020-02-07
Use 'odo create url' to expose an application running locally


johnmcollier:
## User Story
As a user, I want to be able to expose my applications/containers when using odo for local development. Instead of exposing a route or ingress, it should assign an external port to the application container. `odo push` would then restart that application container, with the specified port used as the exposed port.

## Acceptance Criteria 
- [ ] It should assign an exposed port to the application container
- [ ] It should allow users to specify a specific exposed port on the `odo url create` command by specifying a `--exposed-port` parameter, otherwise a randomly assigned 5 digit port will be used.
- [ ] Exposed port number will be stored in the env.yaml file
- [ ] If port already exists in env.yaml, running the `odo url create` command will reassign the port
- [ ] `odo push` will still be used to make the changes take effect on the app container

## Links 
- Parent epic: https://github.com/openshift/odo/issues/2576

/kind user-story

PR #2806, opened 2020-04-02
Implement a basic, first pass `odo push` with Docker support

johnmcollier:
**What type of PR is this?**
/kind feature

This PR implements a basic, first pass implementation of `odo push` with local Docker containers when the PushTarget preference is set to Docker. Currently it's limited to just deploying the containers and setting up the shared source volume. This will unblock folks who are assigned issues that are dependent on this functionality (such as https://github.com/openshift/odo/issues/2717, https://github.com/openshift/odo/issues/2580, https://github.com/openshift/odo/issues/2581)

So, what does the PR do in all?

- Creates a Docker platform adapter for interacting with devfile components in Docker
    - The Docker adapter is used only if `experimental` is true **and** `PushTarget` is set to Docker
- Creates a Docker component adapter, which does the following
    - `Push` pulls images and starts containers based on what was specified in the devfile. Each container is labeled with `component=<odo-componentName>` and `alias=<devfile-component-alias>` that provide an easy way to group and find the pods associated with an odo component
    - `DoesComponentExist` returns true if there's at least **one** container running that's labeled with `component=<odo-componentName>`

- Unit tests
- Integration tests

Fixes https://github.com/openshift/odo/issues/2579

0) Have Docker installed and running on your system

1) Build my branch

2) Set **both** experimental mode to true and PushTarget to Docker
    - `odo preference set experimental true`
    - `odo preference set pushtarget docker`

3) Create a devfile component
   - `odo create openLiberty`
4) `odo push`


amitkrout:
> Integration tests coming in a later commit to this branch. Needed to get this PR up first before experimenting with Travis

👍 
@johnmcollier  It won't be much complex to run the test on travis. As we have ```oc cluster up``` [configured](https://github.com/openshift/odo/blob/master/scripts/oc-cluster.sh) to run for each job on travis which uses default running docker instances as a pre requisite. So you just need to write the test scenario, write the make target for it then call it in .travis.yaml. 

As part of running the same test on 4.* is concerned there we need to figure out how to do it (Possibly we need to configure docker instances in the test [container](https://github.com/openshift/odo/blob/master/openshift-ci/build-root/Dockerfile)). I would suggest create a separate issue for running the same test against 4.* cluster and assign it to @prietyc123.

johnmcollier:
Updated the PR with integration tests 😃 

For now they're just running in Travis, I'll open an issue later today to track running them elsewhere.

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+		// Check to see if this component is already running and if so, update it
+		// If component isn't running, re-create it, as it either may be new, or crashed.
+		containers, err := a.Client.GetContainersByComponentAndAlias(componentName, *comp.Alias)
+		if err != nil {
+			return errors.Wrapf(err, "unable to list containers for component %s", componentName)
+		}
+		if len(containers) == 0 {
+			// Container doesn't exist, so need to pull its image (to be safe) and start a new container
+			s := log.Spinner("Pulling image " + *comp.Image)
+			defer s.End(false)
```
girishramnani, line 133: I understand this part that if this loop returns before reaching https://github.com/openshift/odo/pull/2806/files#diff-00b1316479a0f694654a918c0915c476R139. The spinner will end in failure but this is a bit confusing, so can we please move it inside the err != nil if statement?

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+
+			err = a.Client.PullImage(*comp.Image)
+			if err != nil {
+				return errors.Wrapf(err, "Unable to pull %s image", *comp.Image)
+			}
+			s.End(true)
+
+			// Start the container
+			err = a.startContainer(componentName, projectVolumeName, comp)
```
girishramnani, line 142: Shouldn’t starting a container also have a spinner of its own?

girishramnani: Just thinking from UX perspective

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+					return errors.Wrapf(err, "Unable to start container for devfile component %s", *comp.Alias)
+				}
+				glog.V(3).Infof("Successfully created container %s for component %s", *comp.Image, componentName)
+			}
+		}
+	}
+	return nil
+}
+
+func (a Adapter) startContainer(componentName string, projectVolumeName string, comp versionsCommon.DevfileComponent) error {
```
maysunfaisal, line 178: Can this be moved to docker/utils/utils.go to be consistent with kubernetes package?

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+	containerLabels := map[string]string{
+		"component": componentName,
+		"alias":     *comp.Alias,
+	}
+
+	containerConfig := a.Client.GenerateContainerConfig(*comp.Image, comp.Command, comp.Args, envVars, containerLabels)
+	return containerConfig
+}
+
+func addProjectVolumeToComp(projectVolumeName string, hostConfig *container.HostConfig) *container.HostConfig {
```
maysunfaisal, line 210: Can this be moved to docker/utils/utils.go to be consistent with kubernetes package?

pkg/lclient/containers.go
```diff
@@ -33,10 +48,9 @@ func (dc *Client) GetContainerList() ([]types.Container, error) {
 // containerConfig - configurations for the container itself (image name, command, ports, etc) (if needed)
 // hostConfig - configurations related to the host (volume mounts, exposed ports, etc) (if needed)
 // networkingConfig - endpoints to expose (if needed)
-// containerName - name to give to the container
 // Returns an error if the container couldn't be started.
-func (dc *Client) StartContainer(containerConfig *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, containerName string) error {
-	resp, err := dc.Client.ContainerCreate(dc.Context, containerConfig, hostConfig, networkingConfig, containerName)
+func (dc *Client) StartContainer(containerConfig *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig) error {
+	resp, err := dc.Client.ContainerCreate(dc.Context, containerConfig, hostConfig, networkingConfig, "")
```
maysunfaisal, line 53: any reason why Container Name is empty string? 🤔

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return nil
+}
+
+func (a Adapter) updateComponent() (err error) {
+	glog.V(3).Info("The component already exists, attempting to update it")
+	componentName := a.ComponentName
+
+	// Get the project source volume
+	volumeLabels := map[string]string{
```
maysunfaisal, line 105: Can we have a getter function in utils that returns the volume label, given the component name, since we're using it create and update functions

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+		s.End(true)
+
+		// Start the container
+		err = a.startContainer(componentName, projectVolumeName, comp)
+		if err != nil {
+			return errors.Wrapf(err, "Unable to start container for devfile component %s", *comp.Alias)
+		}
+		glog.V(3).Infof("Successfully created container %s for component %s", *comp.Image, componentName)
+	}
+	glog.V(3).Infof("Successfully created all containers for component %s", componentName)
```
maysunfaisal, line 95: Can we common out this code, as update component's len(containers)=0 case is also doing the same thing..

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+			s.End(true)
+
+			// Start the container
+			err = a.startContainer(componentName, projectVolumeName, comp)
+			if err != nil {
+				return errors.Wrapf(err, "Unable to start container for devfile component %s", *comp.Alias)
+			}
+			glog.V(3).Infof("Successfully created container %s for component %s", *comp.Image, componentName)
+		} else {
```
maysunfaisal, line 147: So we create one container for every component, but what if for some weird edge scenario out of odo's control, there is more than one container with the same labels?

Can we say if len > 1, error out and this else condition be len == 1?

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+	s := log.Spinner("Starting container for " + *comp.Image)
+	defer s.End(false)
+	err := a.Client.StartContainer(&containerConfig, &hostConfig, nil)
+	if err != nil {
+		return err
+	}
+	s.End(true)
+	return nil
+}
+func (a Adapter) getContainerConfig(componentName string, comp versionsCommon.DevfileComponent) container.Config {
```
maysunfaisal, line 197: Nit: can we rename this function to something else, I was confused for a moment between this and lclient.GetContainerConfig which gets it from an active container and this function generates and returns the config..

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+			// Get the associated container config from the container ID
+			containerConfig, err := a.Client.GetContainerConfig(containerID)
+			if err != nil {
+				return errors.Wrapf(err, "unable to get the container config for component %s", componentName)
+			}
+
+			// See if the container needs to be updated
+			if utils.DoesContainerNeedUpdating(comp, containerConfig) {
+				glog.V(3).Info("Updating the component")
+				// Remove the container
```
maysunfaisal, line 160: Can we have a spinner here? 🤔

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		s.End(true)
+
+		// Start the container
+		err = a.startContainer(componentName, projectVolumeName, comp)
+		if err != nil {
+			return errors.Wrapf(err, "Unable to start container for devfile component %s", *comp.Alias)
+		}
+		glog.V(3).Infof("Successfully created container %s for component %s", *comp.Image, componentName)
+	}
```
maysunfaisal, line 94: Can this be moved to a func in docker/utils/utils.go to be consistent with kubernetes package, which does all the heavy handling like kubernetes component? Same for the update component function..

pkg/lclient/containers.go
johnmcollier: Good question. As we use labels to identify and retrieve containers for odo-docker, we don't care too much about the container name. By setting it to be a blank string, it allows Docker to assign a randomly generated name to it instead and avoid any potential conflicts

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Yeah, makes sense to me.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Yeah, probably makes sense to have a spinner.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Haha, fair enough. How about `getConfigFromContainer`?

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Yeah, we can error out if len > 1

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Makes sense to me 👍

pkg/lclient/containers.go
maysunfaisal: yeah i thought so too, that is what you were going for hmm

pkg/lclient/containers.go
```diff
@@ -48,3 +62,44 @@ func (dc *Client) StartContainer(containerConfig *container.Config, hostConfig *
...
+		Image:      image,
+		Entrypoint: entrypoint,
+		Cmd:        cmd,
+		Env:        envVars,
+		Labels:     labels,
+	}
+	return containerConfig
+}
+
+func (dc *Client) GenerateHostConfig(isPrivileged bool, publishPorts bool) container.HostConfig {
```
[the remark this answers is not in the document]
johnmcollier: Yeah, that makes sense.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Makes sense to me 👍

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Yeah, sure

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: `generateAndGetContainerConfig`? since this function generates and gets the container config

johnmcollier: Works for me.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: `pullAndStartComponent` in utils.go perhaps?

maysunfaisal: `pullAndStartContainer` instead of component?

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Actually, this should stay in adapter.go since it relies on the Adapter struct defined in this file (and it's equivalent to `createOrUpdateComponent` in `kubernetes/component/adapter.go`)

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Actually, this should stay in adapter.go since it relies on the Adapter struct defined in this file (it's also equivalent to code in createOrUpdateComponent in kubernetes/component/adapter.go)

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: It does have a spinner, the startContainer function invokes it.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Ended up moving these functions into their own file (utils.go) in the `component` package, like what was done for the storage adapter.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Done

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Done

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Done

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Done

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Done

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Updated.

pkg/lclient/containers.go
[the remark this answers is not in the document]
johnmcollier: Updated.

pkg/devfile/adapters/helper.go
```diff
@@ -19,11 +22,16 @@ func NewPlatformAdapter(componentName string, devObj devfile.DevfileObj, platfor
 
 	// Only the kubernetes adapter is implemented at the moment
 	// When there are others this function should be updated to retrieve the correct adapter for the desired platform target
```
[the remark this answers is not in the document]
johnmcollier: Haha, good catch. Fixed.

pkg/devfile/adapters/docker/component/adapter.go
johnmcollier: Sure, done.

pkg/lclient/storage.go
```diff
@@ -0,0 +1,41 @@
...
+	volumeTypes "github.com/docker/docker/api/types/volume"
+	"github.com/pkg/errors"
+)
+
+// CreateVolume creates a Docker volume with the given labels and the default Docker storage driver
+func (dc *Client) CreateVolume(labels map[string]string) (types.Volume, error) {
+	volume, err := dc.Client.VolumeCreate(dc.Context, volumeTypes.VolumeCreateBody{
+		Driver: DockerStorageDriver,
+		Labels: labels,
+	})
```
maysunfaisal, line 17: Do we also want to keep the volume name unique by not passing the name and only relying on labels?
```
Maysuns-MacBook-Pro:odo maysun$ d volume ls
DRIVER              VOLUME NAME
local               b8e4b86d235b571354b3e2080b1eb07f05cc5234746cc1f8ca0b182fd228d9f5
local               codewind_cw-workspace
Maysuns-MacBook-Pro:odo maysun$ d volume inspect b8e4b86d235b571354b3e2080b1eb07f05cc5234746cc1f8ca0b182fd228d9f5
[
    {
        "CreatedAt": "2020-04-10T17:02:20Z",
        "Driver": "local",
        "Labels": {
            "component": "myspring",
            "type": "projects"
        },
        "Mountpoint": "/var/lib/docker/volumes/b8e4b86d235b571354b3e2080b1eb07f05cc5234746cc1f8ca0b182fd228d9f5/_data",
        "Name": "b8e4b86d235b571354b3e2080b1eb07f05cc5234746cc1f8ca0b182fd228d9f5",
        "Options": null,
        "Scope": "local"
    }
]
```

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			// See if the container needs to be updated
+			if utils.DoesContainerNeedUpdating(comp, containerConfig) {
+				s := log.Spinner("Updating the component " + *comp.Alias)
+				defer s.End(false)
+				// Remove the container
+				err := a.Client.RemoveContainer(containerID)
+				if err != nil {
+					return errors.Wrapf(err, "Unable to remove container %s for component %s", containerID, *comp.Alias)
```
maysunfaisal, line 96: do we want to `s.End(false)` here? I checked some of the Kubernetes adapter code, and we dont do it in some of the `if err != nil {` blocks as well, wonder what the UX ramifications are 🤔

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -0,0 +1,167 @@
...
+				}
+
+				// Start the container
+				err = a.startContainer(componentName, projectVolumeName, comp)
+				if err != nil {
+					return errors.Wrapf(err, "Unable to start container for devfile component %s", *comp.Alias)
```
maysunfaisal, line 102: Same comment about `s.End(false)`

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -0,0 +1,167 @@
...
+	if comp.MountSources {
+		utils.AddProjectVolumeToComp(projectVolumeName, &hostConfig)
+	}
+
+	// Create the docker container
+	s := log.Spinner("Starting container for " + *comp.Image)
+	defer s.End(false)
+	err := a.Client.StartContainer(&containerConfig, &hostConfig, nil)
+	if err != nil {
+		return err
```
maysunfaisal, line 150: Same comment about `s.End(false)`

tests/integration/docker/cmd_docker_devfile_push_test.go
```diff
@@ -0,0 +1,81 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("odo docker devfile push command tests", func() {
+	var context string
+	var currentWorkingDirectory string
+	var cmpName string
+
+	// TODO: all oc commands in all devfile related test should get replaced by kubectl
+	// TODO: to goal is not to use "oc"
```
maysunfaisal, line 20: can you pls remove these comments

tests/helper/helper_docker.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/onsi/gomega/gexec"
+)
+
+type DockerRunner struct {
+	// path to docker binary
+	path string
+}
+
+// NewDockerRunner initializes new DockerRunner
+func NewDockerRunner(ocPath string) DockerRunner {
```
maysunfaisal, line 18: `ocPath` to `dockerPath`

tests/integration/docker/cmd_docker_devfile_push_test.go
```diff
@@ -0,0 +1,81 @@
+package docker
```
maysunfaisal, line 1: how do you feel about all devfile tests being devfile dir? and then separated out to `tests/integration/devfile/docker` and `tests/integration/devfile/kube`

tests/integration/docker/cmd_docker_devfile_push_test.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var _ = Describe("odo docker devfile push command tests", func() {
+	var context string
+	var currentWorkingDirectory string
+	var cmpName string
+
+	// TODO: all oc commands in all devfile related test should get replaced by kubectl
+	// TODO: to goal is not to use "oc"
+	dockerClient := helper.NewDockerRunner("docker")
```
maysunfaisal, line 21: So this is just a question, was docker already installed on the CI or did you have to do something?

pkg/devfile/adapters/docker/component/utils.go
johnmcollier: It's not necessary. The `defer s.End(false)` when we initialize the spinner will make the spinner end with a failure if the function exits at any point before we call `s.End(true)`

pkg/devfile/adapters/docker/component/utils.go
johnmcollier: See my reply to your other comment.

pkg/devfile/adapters/docker/component/utils.go
johnmcollier: See my reply to your other comment.

pkg/lclient/storage.go
johnmcollier: We should still rely on labels to be safe, as we can let Docker deal with any potential naming conflicts.

tests/integration/docker/cmd_docker_devfile_push_test.go
johnmcollier: Docker was already installed on travis where these tests are running. Docker isn't available on the OpenShift CI, so the tests don't run there.

tests/integration/docker/cmd_docker_devfile_push_test.go
johnmcollier: Fixed.

tests/integration/docker/cmd_docker_devfile_push_test.go
johnmcollier: Makes sense, fixed.

tests/helper/helper_docker.go
johnmcollier: Fixed.

maysunfaisal:
changes look good
/lgtm

tests/integration/docker/cmd_docker_devfile_push_test.go
maysunfaisal: So would you know, how to check the list of tests running on the CI like
`ci/prow/v4.1-integration-e2e-benchmark`, etc.

johnmcollier: The integration tests that run on the OpenShift CI are defined here: https://github.com/openshift/odo/blob/master/scripts/openshiftci-presubmit-all-tests.sh

johnmcollier:
Had to fix a merge conflict

maysunfaisal:
adding back lgtm after rebase
/lgtm

kadel:
> As part of running the same test on 4.* is concerned there we need to figure out how to do it (Possibly we need to configure docker instances in the test [container](https://github.com/openshift/odo/blob/master/openshift-ci/build-root/Dockerfile)). I would suggest create a separate issue for running the same test against 4.* cluster and assign it to @prietyc123.

Why we would want to run this against clusters. No cluster is used in this case, it is just using Docker daemon

PR #2739, opened 2020-03-20
Add ability to create new service backed by an operator

dharmit:
**What type of PR is this?**
/kind feature

**What does does this PR do / why we need it**:
This PR creates operator backed services. 

**Which issue(s) this PR fixes**:
Fixes a part of #2613 

**How to test changes / Special notes to the reviewer**:
* On a CRC or OCP 4.x cluster, create a project using `developer` user: `odo project create myproject`
* Log into the OpenShift Web Console as `kubeadmin` user and install etcd operator in the `myproject` namespace.
* Go to the cli and confirm that the operator is installed in `myproject` namespace
   ```sh
   $ export ODO_EXPERIMENTAL=true
   $ odo login https://api.crc.testing:6443 -u developer -p developer
   $ odo catalog list services
   Operators available through Operator Hub
   NAME                       CRDs
   etcdoperator.v0.9.4        EtcdCluster, EtcdBackup, EtcdRestore
   ```
* Create the operator backed service:
  ```sh
  odo service create etcdoperator.v0.9.4 --crd EtcdCluster
  Deploying service etcdoperator.v0.9.4 of type: etcdoperator.v0.9.4
  example
   ✓  Deploying service [6ms]
   ✓  Service 'etcdoperator.v0.9.4' was created

  Progress of the provisioning will not be reported and might take a long time
  You can see the current status by executing 'odo service list'
  Optionally, link etcdoperator.v0.9.4 to your component by running: 'odo link <component-name>'
  ```
* Verify that pods have come up (`odo` doesn't support listing operator backed services yet so use `oc` CLI for this):
  ```sh
  $ oc get pods
  NAME                               READY   STATUS    RESTARTS   AGE
  etcd-operator-59c7bf9dfd-sdtl2     3/3     Running   0          20m
  example-8jgjmhq4m2                 1/1     Running   0          31s
  example-d6p2l2s7f8                 1/1     Running   0          103s
  example-rrnm4sz7rs                 1/1     Running   0          55s
  ```

Issue #2723, opened 2020-03-16
Instantiate an operator backed service using yaml file


dharmit:
## User Story
As a user I want to edit the example (`alm-example`) used to spin up an operator backed service. This is required because:
- `alm-examples` used by odo to spin up an operator backed service are not always 100% functional in that, they contain template data which can't actually spin up a service.
- I'd like to change certain parameters in the CR before spinning up an actual service

## Acceptance Criteria 
- [ ] odo must add a flag (eg: `--dry-run`) which would show the example that will eventually be used to spin up the service. (#2810)
- [ ] user should be able to provide a yaml file similar to aforementioned example that can then be used to spin up the service.

  This will be similar to `oc create -f example.yaml`

## Links 
- Feature Request: #2161 
- Related issue: #2613 (first point: ` odo should create a new instance of an operator in my project using odo service create command`)

/kind user-story

kadel:
What will be the command?
Are you thinking about something like `odo service create -f etcd.yaml` ? Or some other approach?

dharmit:
#2810 addresses the first part of this issue's acceptance criteria.

dharmit:
@kadel I'm thinking of making user specify the operator name and crd name as is the case right now, and specify file name after that. Something like:
```
$ odo service create etcdoperator.v0.9.4 --crd EtcdCluster -f etcd.yaml
```
This long command will help us do some validation so that a user doesn't erroneously or purposefully try to create something else by doing `odo service create -f etcd.yaml`. They could create something else if we don't keep any validation because we're simply passing the unstructured data to the APIs.

What do you think about this approach?

kadel:
Not sure about that. The commands above looks way too long and complicated :-( It is also full of duplicate values, everything that is specified is also in the yaml file :-( 

We can still do some validation even without forcing users to provide duplicated parameters.

Would it be possible to simply parse the yaml file, get `Kind` and check if it matches one of the `CRD`s provided by already installed operators?

dharmit:
Yes. That sounds doable. Agree with your points about long and complicated. Let's go with it.

Just one clarification however. The `odo service create` command will accept a path to some yaml file only when experimental mode is enabled.

PR #2810, opened 2020-04-03
Print yaml definition of CR when asked for dry-run

dharmit:
**What type of PR is this?**

/kind feature

When user does `odo service create <operator-name> --crd <crd-name> --dry-run`,
it will print only the CR's yaml definition that will be used to bring up the
service.

This is needed because yaml definitions provided in the alm-examples are not
always complete. Sometimes they contain template data which needs to be
modified to be able to start the service. After this PR, we will have another
PR which will let the user specify file containing a yaml definition to bring
up the service. That's where this will yaml output in this PR will be really
helpful.

Addresses first acceptance criteria of #2723.

1. Make sure you have an operator installed on your cluster.
2. Enable experimental mode `export ODO_EXPERIMENTAL=true`
3. Execute following command:
    ```sh
    $ odo service create etcdoperator.v0.9.4 --crd EtcdCluster --dry-run
    Following definition will be used to create EtcdCluster:

    apiVersion: etcd.database.coreos.com/v1beta2
    kind: EtcdCluster
    metadata:
      name: example
    spec:
      size: 3
      version: 3.2.13
    ```

**NOTE to reviewer & QE**: The output generated by above command is likely to
change when the operator is upgrade by its developers. Hence, the integration
test for this PR only checks for the presence of the line `Following definition
will be used to create`.

**NOTE to approver**: This PR is rebased on top of #2739 and hence that should
go in first.

### DIFF
--- a/pkg/odo/cli/service/create.go
+++ b/pkg/odo/cli/service/create.go
@@ -326,7 +326,6 @@ func (o *ServiceCreateOptions) Run() (err error) {
 				return err
 			}
 
-			log.Infof("Following definition will be used to create %s:\n", o.CustomResource)
 			log.Info(string(yamlCR))
 
 			return nil
--- a/pkg/odo/cli/service/create.go
+++ b/pkg/odo/cli/service/create.go
@@ -12,6 +12,7 @@ import (
 	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/pkg/errors"
 
+	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	scv1beta1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
 	"github.com/openshift/odo/pkg/log"
@@ -87,6 +88,8 @@ type ServiceCreateOptions struct {
 	version string
 	// Resource of the GVR
 	resource string
+	// If set to true, DryRun prints the yaml that will create the service
+	DryRun bool
 }
 
 // NewServiceCreateOptions creates a new ServiceCreateOptions instance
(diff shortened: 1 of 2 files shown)


tests/integration/operatorhub/cmd_service_test.go
```diff
@@ -27,4 +30,40 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
...
+			// Look for pod with example name because that's the name etcd will give to the pods.
+			etcdPod := regexp.MustCompile(`example-.[a-z0-9]*`).FindString(pods)
+
+			ocArgs := []string{"get", "pods", etcdPod, "-o", "template=\"{{.status.phase}}\"", "-n", CI_OPERATOR_HUB_PROJECT}
+			helper.WaitForCmdOut("oc", ocArgs, 1, true, func(output string) bool {
+				return strings.Contains(output, "Running")
+			})
+
+			// Delete the pods created. This should idealy be done by `odo
+			// service delete` but that's implemented for operator backed
```
amitkrout, line 52: that's -> that's not

pkg/odo/cli/service/create.go
```diff
@@ -190,6 +210,31 @@ func (o *ServiceCreateOptions) Validate() (err error) {
...
+		}
+
+		var almExamples []map[string]interface{}
+		err = json.Unmarshal([]byte(csv.Annotations["alm-examples"]), &almExamples)
+		if err != nil {
+			return errors.Wrap(err, "unable to unmarshal alm-examples")
+		}
+
+		for _, example := range almExamples {
+			if example["kind"].(string) == o.Crd {
```
girishramnani, line 230: if the Crd is doesn't match here then odo will assume it as a service catalog request, so `odo service create  etcdOperator --crdEtdCluster --dry-run` would not show error and instead try to create service catalog stuff If i am not wrong.

dharmit: Since you brought up similar point in a different PR (https://github.com/openshift/odo/pull/2739#pullrequestreview-387155338), this has been addressed.

dharmit:
> @dharmit No UTs will be added because of [#2739 (comment)](https://github.com/openshift/odo/pull/2739#issuecomment-607660629) including the the changes you made for cli packages `pkg/odo/cli/service/create.go` too, right ?

Yes, that's right. The data is so dynamic that it's not possible to add much in terms of unit tests. But I'm really willing to discuss adding more integration tests if you or anyone thinks that we can add some scenario.

tests/integration/operatorhub/cmd_service_test.go
dharmit: Thanks for the catch! Fixed.

adisky:
Looks good to me, just one suggestion, we should have some difference in formatting of message and yaml  
```
$ odo service create nfd.4.3.10-202003311428 --crd NodeFeatureDiscovery --dry-run
Following definition will be used to create NodeFeatureDiscovery:

apiVersion: nfd.openshift.io/v1alpha1
kind: NodeFeatureDiscovery
metadata:
  name: nfd-master-server
spec:
  namespace: openshift-nfd

```

tests/integration/operatorhub/cmd_service_test.go
```diff
@@ -49,10 +49,20 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
...
 		})
 	})
 
+	Context("When using dry-run option to create operator backed service", func() {
+		It("should only output the definition of the CR that will be used to start service", func() {
+			// First let's grab the etcd operator's name from "odo catalog list services" output
+			operators := helper.CmdShouldPass("odo", "catalog", "list", "services")
+			etcdOperator := regexp.MustCompile(`etcdoperator\.*[a-z][0-9]\.[0-9]\.[0-9]`).FindString(operators)
+
+			helper.CmdShouldPass("odo", "service", "create", etcdOperator, "--crd", "EtcdCluster", "--dry-run")
```
adisky, line 64: can we verify that output contains `apiVersion` `Kind` strings? as the now this `Following definition will be used to create.` is removed

dharmit: Makes sense to me. Done.

Issue #2807, opened 2020-04-02
odo push deletes project files after making a change to devfile


maysunfaisal:
/kind bug

## What versions of software are you using?

**Operating System:** 

**Output of `odo version`:** master

## How did you run odo exactly?

I noticed this when writing integration tests for devfile exec and confirmed this on the master branch without any of the exec code. Also seeing this with the integration test suite on the master branch.

- `odo set experimental true`
- `odo push`
- change devfile like ENV for a container
- `odo push`

## Actual behavior
All the project files disappear from the container after the second push 😢 

## Expected behavior
Project files should be there! 😄 

##### after first push
```
Maysuns-MacBook-Pro:T maysun$ oc exec -it spring-devfile-66ffc6549-skn2d bash -c tools
root@spring-devfile-66ffc6549-skn2d:/# cd /projects/springbootproject/
root@spring-devfile-66ffc6549-skn2d:/projects/springbootproject# ls -la
total 55372
drwxr-xr-x. 5 root root       231 Apr  2 22:40 .
drwxrwxrwx. 3 root root        31 Apr  2 22:40 ..
-rw-r--r--. 1  501 staff      213 Oct 21 19:00 .cw-settings
-rw-r--r--. 1  501 staff       74 Mar 13 16:23 .gitignore
drwxr-xr-x. 2 root root        25 Apr  2 22:40 .vscode
-rw-r--r--. 1  501 staff      237 Oct 21 19:00 Dockerfile
-rw-r--r--. 1  501 staff     2409 Oct 21 19:00 Dockerfile-build
-rw-r--r--. 1  501 staff      325 Oct 21 19:00 Dockerfile-tools
-rw-r--r--. 1  501 staff       97 Oct 21 19:00 Jenkinsfile
-rw-r--r--. 1  501 staff     1814 Oct 21 19:00 README.md
drwxr-xr-x. 3 root root        32 Apr  2 22:40 chart
-rw-r--r--. 1  501 staff     1770 Apr  2 22:34 devfile.yaml
-rwxr-xr-x. 1  501 staff 56662652 Mar 31 19:56 odo
-rw-r--r--. 1  501 staff     3011 Oct 21 19:00 pom.xml
drwxr-xr-x. 4 root root        30 Apr  2 22:40 src
```

##### after second push
```
Maysuns-MacBook-Pro:T maysun$ oc exec -it spring-devfile-57484d9b8f-8kq2x bash -c tools
[...]

PR #2811, opened 2020-04-03
Fix sync when the pod gets re-created

johnmcollier:
**What type of PR is this?**
/kind bug

This PR fixes the issue seen in https://github.com/openshift/odo/issues/2807, where if something in the devfile changed that resulted in a new pod, the source code would get lost (as the emptyDir volume went away). Odo wouldn't detect this and wouldn't do a full sync, only a sync of the files that changed on the user's disk.

To fix this, if the component already exists, this PR compares the pod name before and after the call to `a.createOrUpdateComponent`. If the pod names are different, the pod has been re-created and a full sync is needed.

Fixes https://github.com/openshift/odo/issues/2807

1) Build my branch
2) Create a devfile component
3) Run `odo push`
4) Make a change in the devfile to cause a new pod to get created
5) Run `odo push` again, verify the entire source code is where it should be on the container(s)

@maysunfaisal is working on integration tests for devfile exec that will have this scenario covered

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -314,10 +314,12 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 		return errors.New(fmt.Sprintf("Directory / file %s is empty", path))
 	}
 
+	podSelector := fmt.Sprintf("component=%s", a.ComponentName)
+
 	// Wait for Pod to be in running state otherwise we can't sync data to it.
 	pod, err := a.waitAndGetComponentPod()
 	if err != nil {
-		return errors.Wrapf(err, "unable to get pod for component %s", a.ComponentName)
+		return errors.Wrapf(err, "error retrieve pod %s for component %s", podSelector, a.ComponentName)
 	}
 
 	// Find at least one pod with the source volume mounted, error out if none can be found
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -46,6 +46,17 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	deletedFiles := []string{}
 	changedFiles := []string{}
 	isForcePush := false
+	podChanged := false
+	var podName string
+
+	// If the component already exists, retrieve the pod's name before it's potentially updated
+	if componentExists {
+		pod, err := a.waitAndGetComponentPod()
+		if err != nil {
+			return errors.Wrapf(err, "unable to get pod for component %s", a.ComponentName)
+		}
+		podName = pod.GetName()
+	}
 
 	err = a.createOrUpdateComponent(componentExists)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -295,13 +317,11 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 	}
 
 	podSelector := fmt.Sprintf("component=%s", a.ComponentName)
-	watchOptions := metav1.ListOptions{
-		LabelSelector: podSelector,
-	}
+
 	// Wait for Pod to be in running state otherwise we can't sync data to it.
-	pod, err := a.Client.WaitAndGetPod(watchOptions, corev1.PodRunning, "Waiting for component to start")
+	pod, err := a.waitAndGetComponentPod(false)
```
maysunfaisal, line 322: Can we just use the pod retrieved above after createOrUpdateComponent and rollout check https://github.com/openshift/odo/pull/2811/files#diff-825a1a88d75e789899c58e89a79eb767R73? 

Because that is essentially the new pod, we dont necessarily need to retrieve it again here. You can probably skip the if condition `if componentExists` and initialize `podName := ""` and you wont need this retrieval here.

johnmcollier: The object returned from `WaitForDeploymentRollout` is a Deployment which we can't get the pod name from directly, unfortunately.

maysunfaisal:
@johnmcollier https://github.com/openshift/odo/pull/2735 was merged, you will need to rebase the PR

amitkrout:
> @maysunfaisal is working on integration tests for devfile exec that will have this scenario covered

@johnmcollier ok, make sense. @maysunfaisal do we have an open issue for this? or are you are planning to send the test pr directly ? ATM i would say let create an open issue for integration test for devfile exec and mark this pr test may be in the task list of devfile exec test overage, so that we won't forget to add this scenario. WDYT ?

kadel:
> To fix this, if the component already exists, this PR compares the pod name before and after the call to `a.createOrUpdateComponent`. If the pod names are different, the pod has been re-created and a full sync is needed.

As long as we are using `Recreate` strategy using the pod names should work. 
But wouldn't it be safer to use `metadata.generation` to detect if there was an update?

/approve

maysunfaisal:
@amitkrout I have opened an issue here https://github.com/openshift/odo/issues/2818

johnmcollier:
@kadel Good question! I was finding that the `metadata.generation` field was still getting updated, even if nothing in the `spec` field of the deployment was changing, so it wasn't viable.

(On a side note: I'm note sure _why_ the generation field would change when nothing in the spec changed, but that's what I observed)

```
Johns-MacBook-Pro-3:openLiberty johncollier$ kubectl get deploy openliberty -o jsonpath='{.metadata.generation}{"\n"}'
6
Johns-MacBook-Pro-3:openLiberty johncollier$ odo push
 •  Push devfile component openliberty  ...
 ✓  Waiting for component to start [44ms]
 ✓  Checking file changes for pushing [3ms]
 ✓  Push devfile component openliberty [509ms]
Johns-MacBook-Pro-3:openLiberty johncollier$ kubectl get deploy openliberty -o jsonpath='{.metadata.generation}{"\n"}'
7
```

kadel:
we need to cover this scenario with integration test

maysunfaisal:
fixes look good, thx for the change
/lgtm

johnmcollier:
@kadel Yeah, agreed. Maysun's been working on devfile exec tests and already has this scenario covered in his tests, so that's why I left it out of this PR.

maysunfaisal:
yep thats how this issue was discovered.

PR #2817, opened 2020-04-06
Adds secure field to url list table ouput and json output.

### LINKED ISSUE #2775: 'odo url list'/'odo url list -o json' does not include information to distinguish secure and not secure URLs
/kind bug

**Operating System:** 
Windows 10
**Output of `odo version`:** 
odo v1.1.1 (02fbd3d08)

I have a local component not pushed to the cluster with two urls. One is secured and another is not.

## Actual behavior
```
> odo.exe url list
Found the following URLs for component nodejs-ex-git in application app:
NAME     STATE          URL     PORT
url2     Not Pushed     ://     8080
url1     Not Pushed     ://     8080

```

## Expected behavior
There should be information to distinguish, secured and not secured URLs.

N/A

mik-dass:
**What type of PR is this?**

/kind feature

Adds information about the url being a secure one or not in `odo url list` normal and json output.

Fixes #2775 

- create a normal odo component with secure url and execute `odo url list` and also json version of it.
- create a devfile odo component with secure url and execute `odo url list` and also json version of it.

The outputs should contain information about the url being a secure one or not.

### DIFF
--- a/pkg/odo/cli/url/list.go
+++ b/pkg/odo/cli/url/list.go
@@ -80,7 +80,7 @@ func (o *URLListOptions) Run() (err error) {
 
 			log.Infof("Found the following URLs for component %v", componentName)
 			tabWriterURL := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
-			fmt.Fprintln(tabWriterURL, "NAME", "\t", "URL", "\t", "PORT")
+			fmt.Fprintln(tabWriterURL, "NAME", "\t", "URL", "\t", "PORT", "\t", "SECURE")
 
 			// are there changes between local and cluster states?
 			outOfSync := false
@@ -88,7 +88,7 @@ func (o *URLListOptions) Run() (err error) {
 				var present bool
 				for _, u := range urls.Items {
 					if i.Name == u.Name {
-						fmt.Fprintln(tabWriterURL, u.Name, "\t", url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host), "\t", u.Spec.Rules[0].IngressRuleValue.HTTP.Paths[0].Backend.ServicePort.IntVal)
+						fmt.Fprintln(tabWriterURL, u.Name, "\t", url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host), "\t", u.Spec.Rules[0].IngressRuleValue.HTTP.Paths[0].Backend.ServicePort.IntVal, "\t", u.Spec.TLS != nil)
 						present = true
 					}
 				}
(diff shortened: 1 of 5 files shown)


pkg/odo/cli/url/list.go
```diff
@@ -76,15 +76,15 @@ func (o *URLListOptions) Run() (err error) {
...
+			fmt.Fprintln(tabWriterURL, "NAME", "\t", "URL", "\t", "PORT", "\t", "SECURE")
 
 			// are there changes between local and cluster states?
 			outOfSync := false
 			for _, i := range localUrls {
 				var present bool
 				for _, u := range urls.Items {
 					if i.Name == u.Name {
-						fmt.Fprintln(tabWriterURL, u.Name, "\t", url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host), "\t", u.Spec.Rules[0].IngressRuleValue.HTTP.Paths[0].Backend.ServicePort.IntVal)
+						fmt.Fprintln(tabWriterURL, u.Name, "\t", url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host), "\t", u.Spec.Rules[0].IngressRuleValue.HTTP.Paths[0].Backend.ServicePort.IntVal, "\t", u.Spec.TLS != nil)
```
girishramnani, line 87: can we make this more readable? just top of my head maybe extract `url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host)` into a helper?

mik-dass: > url.GetURLString(url.GetProtocol(routev1.Route{}, u), "", u.Spec.Rules[0].Host)

But url.GetURLString() is already a helper function.

girishramnani: I am ok if you feel we cannot make this more clean.

mik-dass: I tried certain things but it was making things a bit complicated and weird for me.

dharmit:
For a URL that's not set to be `secure`, is it expected that we don't see anything in json output?
```sh
$ odo url list
Found the following URLs for component nodejs-nodejs-ex-zdgz in application app:
NAME     STATE      URL                                            PORT     SECURE
url1     Pushed     http://url1-app-myproject.apps-crc.testing     8080     false

$ odo url list -o json
{
    "kind": "List",
    "metadata": {},
    "items": [
        {
            "kind": "url",
            "metadata": {
                "name": "url1",
                "creationTimestamp": null
            },
            "spec": {
                "host": "url1-app-myproject.apps-crc.testing",
                "protocol": "http",
                "port": 8080
            },
            "status": {
                "state": "Pushed"
            }
        }
    ]
}
```

mik-dass:
> For a URL that's not set to be secure, is it expected that we don't see anything in json output?

We are using ``json:"secure,omitempty"`. So it gets removed during json formatting.

dharmit:
I'd vote to keep it. Explicit is better than implicit. :wink: 

Let's wait for others to chime in. :smile:

mik-dass:
> I'd vote to keep it. Explicit is better than implicit. wink

Added back. Please have a look again :)

amitkrout:
> We are using json:"secure,omitempty". So it gets removed during json formatting when it is false. @dharmit @kadel @girishramnani Should I keep it?

Test looks good to me. I won't mind if you guys finalize the concern and apply lgtm label.

PR #2830, opened 2020-04-07
Update Go version and base image on Openshift CI

prietyc123:
**What type of PR is this?**
```

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind code-refactoring
```

Updates the base image of go on  openshift CI

Fixes NA

It should pull go-1.13+ image on Openshift CI


prietyc123:
Team I have updated the go version to 1.13 in OpenSHift CI to be in sync with our travis build env. So in case you are using any version less than this I would suggest let's migrate to 1.13 to be in sync actual dev env with CI setup.

Issue #2782, opened 2020-03-31
Warning message gets printed multiple times


mik-dass:
/kind bug

## How did you run odo exactly?

`odo push`

## Actual behavior

```
[mrinaldas@localhost nodejs-ex]$ odo push
Validation
 ✓  Checking component [1s]

Configuration changes
 ✓  Initializing component
 ✓  Creating component [5s]

Pushing to component nodejs-nodejs-ex-nyyg of type local
 ✓  Checking files for pushing [715424ns]
 ◐  Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◓  Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◑  Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◒  Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◐  Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◓  
^C
```
`Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ◓  ` gets printed again and again.

## Expected behavior

`Waiting for component to start [⚠ WARNING x5: BackOff, use `-v` for more inf ` to be printed once

cdrage:
@mik-dass Is your terminal really small? This is what may be causing the issue.. I may have to shorten the info / output)

mik-dass:
> @mik-dass Is your terminal display really small? This is what may be causing the issue.. I may have to shorten the info / output). Can you re-run it with your terminal display wider?

yes on a wider terminal, the duplication seems to be gone.

cdrage:
@mik-dass Yup. That's normal. It's the terminal doing it's thing with regards to the spinner. 

I do realize that the warning message is actually quite long, I'm going to push a PR to reduce the size to hopefully we don't have this occur again on smaller terminals.

mik-dass:
> I do realize that the warning message is actually quite long, I'm going to push a PR to reduce the size to hopefully we don't have this occur again on smaller terminals.

OK :+1:

PR #2836, opened 2020-04-07
Shortens warning output for event information

cdrage:
**What type of PR is this?**

/kind code-refactoring

Updates the warning output to have less information (remove the part
about using `-v` for more information).

This is needed in order to reduce the width of text so we don't run into
any issues with output being resized due to small terminals.

Fixes https://github.com/openshift/odo/issues/2782

Run it :)

```sh

odo create node js
odo preference set PushTimeout

watch -n 1 oc delete pvc --all

odo push -f
```


girishramnani:
XS PR so merging
/lgtm

PR #2844, opened 2020-04-08
Update documentation links from 4.2 to 4.3

dharmit:
**What type of PR is this?**
/kind documentation

We need it to keep the docs updated.

NA

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -106,6 +106,7 @@ The `TAGS` column represents the available image versions, for example, `10` rep
 == Official documentation
 
 * link:https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/installing-odo.html[Installing odo]
+* link:https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/odo-architecture.html[Odo architecture]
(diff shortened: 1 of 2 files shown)


dharmit:
@Preeticp @boczkowska @cdrage I've used `latest` in the URL so that it always takes the users to latest version of OpenShift documentation. Let me know if this is not a good idea.

Preeticp:
That's the way we want to do it @dharmit. Thank you!
LGTM

Just an observation and not a part of this PR: I think the odo architecture doc (https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/odo-architecture.html) is an important doc to link to from the Readme. WDYT?

dharmit:
Thanks, @Preeticp. And I agree with making architecture doc a part of the README. Since this PR is really small, I pushed a commit to add it just after the link to installing odo. If you think it'd make sense to have it a different place, please let me know.

Preeticp:
LGTM, thank you!

amitkrout:
prow issue
/retest

amitkrout:
> LGTM, thank you!

/lgtm

PR #2872, opened 2020-04-13
Fix documentation so that it would proper render

dharmit:
**What type of PR is this?**
/kind documentation
[skip ci]

We need it so that bullet points and sections of document would properly render.

NA

Compare the rendered document against current master and proposed PR. This can be done by clicking "Display the rich diff" when reviewing the PR.


prietyc123:
flake tracked on issue : https://github.com/openshift/odo/issues/2878

/retest

PR #2873, opened 2020-04-13
Add a blank to align with other command

jichenjc:
**What type of PR is this?**

 /kind cleanup

before the change the command looks like following, to align with other command's format
```
Examples:
  # Create a URL with a specific name by automatically detecting the port used by the component
  odo url create example
  # Create a URL for the current component with a specific port
  odo url create --port 8080

```

Fixes #?


dharmit:
@jichenjc Thanks for opening this PR. Coincidentally I opened a PR (#2875) for `odo url create` that, among other things, also addresses the fix you've proposed in this PR.

Would it be OK if we close this PR since my PR has proposed fix as well?

jichenjc:
sure, thanks for the reminder~

Issue #2646, opened 2020-02-27
panic: runtime error: slice bounds out of range, when running `odo update --git`


kadel:
```
▶ odo component create java:8 --project myproject
 ✓  Validating component [58ms]

Please use `odo push` command to create the component with source deployed

▶ odo push
Validation
 ✓  Checking component [102ms]

Configuration changes
 ✓  Initializing component
 ✓  Creating component [387ms]

Pushing to component java-spring-boot-pgbf of type local
 ✓  Checking files for pushing [4ms]
 ✓  Waiting for component to start [1m]
 ✓  Syncing files to the component [621ms]
 ✓  Building component [1m]
 ✓  Changes successfully pushed to component


▶ cat .odo/config.yaml
kind: LocalConfig
apiversion: odo.openshift.io/v1alpha1
ComponentSettings:
  Type: java:8
  SourceLocation: ./
  SourceType: local
  Ports:
  - 8443/TCP
  - 8778/TCP
  - 8080/TCP
  Application: app
  Project: myproject
  Name: java-spring-boot-pgbf

▶ odo url create --port 8080
 ✓  URL java-spring-boot-pgbf-8080 created for component: java-spring-boot-pgbf

To create URL on the OpenShift Cluster, please use `odo push`

▶ odo push
Validation
 ✓  Checking component [106ms]

Configuration changes
 ✓  Retrieving component data [126ms]
 ✓  Applying configuration [150ms]

Applying URL changes
 ✓  URL java-spring-boot-pgbf-8080: http://java-spring-boot-pgbf-8080-app-myproject.apps.tkral.devcluster.openshift.com created

Pushing to component java-spring-boot-pgbf of type local
 ✓  Checking file changes for pushing [6ms]
 ✓  No file changes detected, skipping build. Use the '-f' flag to force the build.

▶ odo update --git https://github.com/kadel/spring-boot-http-booster.git

Configuration changes
 ✓  Retrieving component data [310ms]
 ✓  Triggering build from git [110ms]
 ✓  Waiting for build to finish [2m]
 ✗  Applying configuration [68012ns]
panic: runtime error: slice bounds out of range [2:1]

goroutine 1 [running]:
github.com/openshift/odo/pkg/occlient.removeVolumeMountsFromDC(0xc000a76600, 0x21, 0xc0005ca800, 0x1)
	/Users/tkral/Code/goenv/odo/src/github.com/openshift/odo/pkg/occlient/volumes.go:143 +0x479
[...]

PR #2881, opened 2020-04-14
Remedy panic slice-out-bounds error when using `odo update`

cdrage:
**What type of PR is this?**

/kind bug

This PR:

- Fixes the common issue of:
`panic: runtime error: slice bounds out of range [2:1]` when going from
`odo create --local` to `odo update --git`
- Adds way more tests for privatized functions
- Fixes a potential issue with removing volume mounts with

Why?

What was happening was that when removing a volume, if the length of volumes was only **ONE** then an `out of bounds` error occurred. This code checks to see if the length is 1 as well as performing the correct array slicing.

Fixes https://github.com/openshift/odo/issues/2646

To try and recreate the error (which isn't happening anymore):

```sh
odo create nodejs
odo url create
odo push

odo update --git https://github.com/openshift/nodejs-ex
```

### DIFF
--- a/pkg/occlient/volumes.go
+++ b/pkg/occlient/volumes.go
@@ -120,31 +120,86 @@ func (c *Client) getVolumeNamesFromPVC(pvc string, dc *appsv1.DeploymentConfig) 
 
 // removeVolumeFromDC removes the volume from the given Deployment Config and
 // returns true. If the given volume is not found, it returns false.
-func removeVolumeFromDC(vol string, dc *appsv1.DeploymentConfig) bool {
+func removeVolumeFromDC(vol string, dc *appsv1.DeploymentConfig) error {
+
+	// Error out immediately if there are zero volumes to begin with
+	if len(dc.Spec.Template.Spec.Volumes) == 0 {
+		return errors.New("there are *no* volumes in this DeploymentConfig to remove")
+	}
+
 	found := false
-	for i, volume := range dc.Spec.Template.Spec.Volumes {
-		if volume.Name == vol {
-			found = true
-			dc.Spec.Template.Spec.Volumes = append(dc.Spec.Template.Spec.Volumes[:i], dc.Spec.Template.Spec.Volumes[i+1:]...)
+
+	// If for some reason there is only one volume, let's slice the array to zero length
+	// or else you will get a "runtime error: slice bounds of of range [2:1] error
+	if len(dc.Spec.Template.Spec.Volumes) == 1 && vol == dc.Spec.Template.Spec.Volumes[0].Name {
+		// Mark as found and slice to zero length
+		found = true
+		dc.Spec.Template.Spec.Volumes = dc.Spec.Template.Spec.Volumes[:0]
+	} else {
+
+		for i, volume := range dc.Spec.Template.Spec.Volumes {
+
+			// If we find a match
+			if volume.Name == vol {
+				found = true
+
+				// Copy (it takes longer, but maintains volume order)
+				copy(dc.Spec.Template.Spec.Volumes[i:], dc.Spec.Template.Spec.Volumes[i+1:])
+				dc.Spec.Template.Spec.Volumes = dc.Spec.Template.Spec.Volumes[:len(dc.Spec.Template.Spec.Volumes)-1]
+
+				break
+			}
+
 		}
 	}
-	return found
+
+	if found {
+		return nil
+	}
+
+	return fmt.Errorf("unable to find volume '%s' within DeploymentConfig '%s'", vol, dc.ObjectMeta.Name)
 }
 
 // removeVolumeMountsFromDC removes the volumeMounts from all the given containers
 // in the given Deployment Config and return true. If any of the volumeMount with the name
 // is not found, it returns false.
-func removeVolumeMountsFromDC(vm string, dc *appsv1.DeploymentConfig) bool {
+func removeVolumeMountsFromDC(volumeMount string, dc *appsv1.DeploymentConfig) error {
+
+	if len(dc.Spec.Template.Spec.Containers) == 0 {
+		return errors.New("something went wrong, there are *no* containers available to iterate through")
+	}
+
 	found := false
+
 	for i, container := range dc.Spec.Template.Spec.Containers {
-		for j, volumeMount := range container.VolumeMounts {
-			if volumeMount.Name == vm {
-				found = true
-				dc.Spec.Template.Spec.Containers[i].VolumeMounts = append(dc.Spec.Template.Spec.Containers[i].VolumeMounts[:j], dc.Spec.Template.Spec.Containers[i].VolumeMounts[j+1:]...)
+
+		if len(dc.Spec.Template.Spec.Containers[i].VolumeMounts) == 1 && dc.Spec.Template.Spec.Containers[i].VolumeMounts[0].Name == volumeMount {
+			// Mark as found and slice to zero length
+			found = true
+			dc.Spec.Template.Spec.Containers[i].VolumeMounts = dc.Spec.Template.Spec.Containers[i].VolumeMounts[:0]
+		} else {
+
+			for j, volMount := range container.VolumeMounts {
+
+				// If we find a match
+				if volMount.Name == volumeMount {
+					found = true
+
+					// Copy (it takes longer, but maintains volume mount order)
+					copy(dc.Spec.Template.Spec.Containers[i].VolumeMounts[j:], dc.Spec.Template.Spec.Containers[i].VolumeMounts[j+1:])
+					dc.Spec.Template.Spec.Containers[i].VolumeMounts = dc.Spec.Template.Spec.Containers[i].VolumeMounts[:len(dc.Spec.Template.Spec.Containers[i].VolumeMounts)-1]
+
+					break
+				}
 			}
 		}
 	}
-	return found
+
+	if found {
+		return nil
+	}
+
+	return fmt.Errorf("unable to find volume mount '%s'", volumeMount)
 }
 
 // generateVolumeNameFromPVC generates a random volume name based on the name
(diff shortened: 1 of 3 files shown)


girishramnani:
@cdrage can you share what was the cause of the bug and how you resolved it?

cdrage:
> @cdrage can you share what was the cause of the bug and how you resolved it?

It's explained in the code, but essentially, we were getting an out of bounds error for an array length that only had 1 volume in it. :)

I'll update the PR

dharmit:
I did following:
```
$ odo update --git https://github.com/sclorg/nodejs-ex

Configuration changes
 ✓  Retrieving component data [84ms]
 ✓  Triggering build from git [70ms]
 ✓  Waiting for build to finish [59s]
 ✓  Applying configuration [24s]

Pushing to component nodejs-nodejs-ex-hyhk of type git
 ✓  The component nodejs-nodejs-ex-hyhk was updated successfully

$ odo url create url
 ✓  URL url created for component: nodejs-nodejs-ex-hyhk

To create URL on the OpenShift Cluster, please use `odo push`

$ odo push
```
and that re-triggered the build. Is this expected? There's no change whatsoever to the underlying code.

cdrage:
> I did following:
> 
> ```
> $ odo update --git https://github.com/sclorg/nodejs-ex
> 
> Configuration changes
>  ✓  Retrieving component data [84ms]
>  ✓  Triggering build from git [70ms]
>  ✓  Waiting for build to finish [59s]
>  ✓  Applying configuration [24s]
> 
> Pushing to component nodejs-nodejs-ex-hyhk of type git
>  ✓  Changes successfully pushed to component
>  ✓  The component nodejs-nodejs-ex-hyhk was updated successfully
> 
> $ odo url create url
>  ✓  URL url created for component: nodejs-nodejs-ex-hyhk
> 
> To create URL on the OpenShift Cluster, please use `odo push`
> 
> $ odo push
> ```
> 
> and that re-triggered the build. Is this expected? There's no change whatsoever to the underlying code.

Yup. If you read https://github.com/openshift/odo/issues/2646 it was a flake that was causing issues.

cdrage:
CI flake:

```sh
Installing from release default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com/ci-op-p99v29v6/release@sha256:35c4b1b9c63eb3f1b472ced9b6524ac7b93de1df6a8b16cda09527c9f427ec33
AWS region: us-west-1 (zones: us-west-1a us-west-1b)
level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Bootstrap Ignition Config\": failed to fetch dependency of \"Common Manifests\": failed to generate asset \"DNS Config\": getting public zone for \"origin-ci-int-aws.dev.rhcloud.com\": listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400, request id: 3824e7c1-4556-4d43-9a84-598287d6fd93"
```

pkg/occlient/volumes.go
```diff
@@ -120,31 +120,84 @@ func (c *Client) getVolumeNamesFromPVC(pvc string, dc *appsv1.DeploymentConfig)
...
+
+		for i, volume := range dc.Spec.Template.Spec.Volumes {
+
+			// If we find a match
+			if volume.Name == vol {
+				found = true
+
+				// Copy (it takes longer, but maintains volume order)
+				copy(dc.Spec.Template.Spec.Volumes[i:], dc.Spec.Template.Spec.Volumes[i+1:])
+				dc.Spec.Template.Spec.Volumes = dc.Spec.Template.Spec.Volumes[:len(dc.Spec.Template.Spec.Volumes)-1]
```
girishramnani, line 148: shouldn't we break here once we have found the volume?

pkg/occlient/volumes.go
```diff
@@ -120,31 +120,84 @@ func (c *Client) getVolumeNamesFromPVC(pvc string, dc *appsv1.DeploymentConfig)
...
+		return errors.New("there are *no* volumes in this DeploymentConfig to remove")
+	}
+
 	found := false
-	for i, volume := range dc.Spec.Template.Spec.Volumes {
-		if volume.Name == vol {
-			found = true
-			dc.Spec.Template.Spec.Volumes = append(dc.Spec.Template.Spec.Volumes[:i], dc.Spec.Template.Spec.Volumes[i+1:]...)
+
+	// If for some reason there is only one volume, let's slice the array to zero length
```
girishramnani, line 132: The reason is because you are editing the list that you are iterating over and there is no `break` statement after we found a volume.
e.g.
the volumes list `dc.Spec.Template.Spec.Volumes ` initially has 2 volumes. then when a volume is found with the matching name - `dc.Spec.Template.Spec.Volumes` shinks by 1 length but we don't break the loop there so if the next volume also has the same name 
```
dc.Spec.Template.Spec.Volumes = append(dc.Spec.Template.Spec.Volumes[:i], dc.Spec.Template.Spec.Volumes[i+1:]...)
```
will fail.

So I would suggest maybe consider just adding a `break` statement in the older code and see if that works @cdrage

girishramnani: and this issue is present here
https://github.com/openshift/odo/blob/5cb97bcd7dbfbc243e8405857fadea268c4e3db2/pkg/occlient/volumes.go#L171
https://github.com/openshift/odo/blob/5cb97bcd7dbfbc243e8405857fadea268c4e3db2/pkg/occlient/volumes.go#L143
and here
https://github.com/openshift/odo/blob/5cb97bcd7dbfbc243e8405857fadea268c4e3db2/pkg/occlient/volumes.go#L128
just from the first glance

pkg/occlient/volumes.go
```diff
@@ -120,31 +120,84 @@ func (c *Client) getVolumeNamesFromPVC(pvc string, dc *appsv1.DeploymentConfig)
...
 	for i, container := range dc.Spec.Template.Spec.Containers {
-		for j, volumeMount := range container.VolumeMounts {
-			if volumeMount.Name == vm {
-				found = true
-				dc.Spec.Template.Spec.Containers[i].VolumeMounts = append(dc.Spec.Template.Spec.Containers[i].VolumeMounts[:j], dc.Spec.Template.Spec.Containers[i].VolumeMounts[j+1:]...)
+
+		if len(dc.Spec.Template.Spec.Containers[i].VolumeMounts) == 1 && dc.Spec.Template.Spec.Containers[i].VolumeMounts[0].Name == volumeMount {
+			// Mark as found and slice to zero length
+			found = true
+			dc.Spec.Template.Spec.Containers[i].VolumeMounts = dc.Spec.Template.Spec.Containers[i].VolumeMounts[:0]
```
girishramnani, line 177: same as above, just a break would solve this.

pkg/occlient/volumes.go
cdrage: Yup. That's intentional. That is where the issue was.

The comment that "if for some reason there is only 1 volume" is that sometimes when editing the deployment config there is actually only *one* volume available rather than two (s2i bootstrap image & the volume for source code).

So that's the reason for the if statement. Although you are right, there should be a break in the for loop.

pkg/occlient/volumes.go
cdrage: Yup, added one!

pkg/occlient/volumes.go
cdrage: added

cdrage:
Friendly reminder that this is ready for another review! :)

mohammedzee1000:
@cdrage can you rebase your PR?

dharmit:
@cdrage I did a bit of back and forth while testing the functionality. And this is probably not related to your PR but an inherent bug maybe. 
1. Created a nodejs-ex component of `local` type and pushed it.
2. Did `odo update` as you've mentioned in the PR description.
3. Again did `odo update -l ./` and `odo push`. Things work well so far.
4. Did step 2 again and it failed with below error:
    ```sh
    odo update --git https://github.com/openshift/nodejs-ex                  
    
    Configuration changes
     ✗  Retrieving component data [64ms]
     ✗  Failed To Update Config To Component Deployed.
    Error: unable to update BuildConfig  for nodejs-nodejs-ex-pzhw component: unable to create BuildConfig for nodejs-nodejs-ex-pzhw-app: buildconfigs.build.openshift.io "nodejs-nodejs-ex-pzhw-app" already exists
    ```


I faced this after I made the comment https://github.com/openshift/odo/issues/2646#issuecomment-625038884. In that comment, I've mentioned similar error when `odo update` times out.

Code looks good to me on this PR, but I guess this is a bug for `odo update` that we might need to fix. WDYT?

mohammedzee1000:
Interesting, changing multiple times fails o_O

mohammedzee1000:
ah yea this should happen as buildconfig was already created once the first time. We should basically check for existing build config and remove it or just retrigger a build

cdrage:
> @cdrage I did a bit of back and forth while testing the functionality. And this is probably not related to your PR but an inherent bug maybe.
> 
> 1. Created a nodejs-ex component of `local` type and pushed it.
> 2. Did `odo update` as you've mentioned in the PR description.
> 3. Again did `odo update -l ./` and `odo push`. Things work well so far.
> 4. Did step 2 again and it failed with below error:
>    ```shell
>    odo update --git https://github.com/openshift/nodejs-ex                  
>    
>    Configuration changes
>     ✗  Retrieving component data [64ms]
>     ✗  Failed To Update Config To Component Deployed.
>    Error: unable to update BuildConfig  for nodejs-nodejs-ex-pzhw component: unable to create BuildConfig for nodejs-nodejs-ex-pzhw-app: buildconfigs.build.openshift.io "nodejs-nodejs-ex-pzhw-app" already exists
>    ```
> 
> I faced this after I made the comment [#2646 (comment)](https://github.com/openshift/odo/issues/2646#issuecomment-625038884). In that comment, I've mentioned similar error when `odo update` times out.
> 
> Code looks good to me on this PR, but I guess this is a bug for `odo update` that we might need to fix. WDYT?

Looks like separate issue :( Can you open up another GitHub issue?

dharmit:
@cdrage yep! I've opened #3114. Please rebase this so that we can lgtm it.

PR #2885, opened 2020-04-15
Adds route support of URLs for openshift clusters

mik-dass:
**What type of PR is this?**

/kind feature
/kind code-refactoring

**What does does this PR do / why we need it**:

It adds the ability of creation of routes for openshift clusters even when experimental mode is on.

**Which issue(s) this PR fixes**:

Fixes #2727 
Fixes #2816 

**How to test changes / Special notes to the reviewer**:

- execute `odo url create` on a openshift cluster (with experimental mode off) and a route should be created
- execute `odo url create` on a openshift cluster (with experimental mode on) and a route should be created
- execute `odo url create --ingress` on a openshift cluster (with experimental mode on) and a ingress should be created
- execute `odo url create` on a kubernetes cluster (with experimental mode on) and a ingress should be created

PR #2906, opened 2020-04-16
local docker support for url creation and deletion

### LINKED ISSUE #2580: Use 'odo create url' to expose an application running locally

yangcao77:
**What type of PR is this?**
/kind feature

> /kind documentation
> [skip ci]

This PR implemented local docker support for url creation and deletion. 
`--exposed-port` allows users to specify a specific exposed port on the `odo url create` command, otherwise a randomly assigned port will be used.
The exposed port will be saved in `env.yaml` file.
rerun `odo url create` would reassign an exposed port for the component. 
If there are more than 1 ports in devfile, user would need to specify `--port` parameter during url creation as well. 

Fixes #2580 

```
odo preference set pushtarget docker
odo create openLiberty
odo push
odo url create --port 9080 --now
```

run `docker ps`, expected an exposed port assigned to `9080`
```
CONTAINER ID        IMAGE                                       COMMAND                  CREATED             STATUS              PORTS                                                  NAMES
c15b2f10b17c        jeevandroid/open-liberty-dev:latest         "/opt/ol/helpers/run…"   4 seconds ago       Up 3 seconds        9443/tcp, 127.0.0.1:63947->9080/tcp                    goofy_bell
```

then run `odo url  delete openliberty-9080` and `odo push`
expected no exposed port bound with 9080 for that container. 

run `odo url create` again for the same port, should override the existing exposed port number:
```
$  odo url create --port 9080 --now
? Port 9080 already has an exposed port 63947 set for it. Do you want to override the exposed port Yes
 ✓  URL lopenliberty-9080 created for component: openliberty with exposed port: 59940
 •  Push devfile component openliberty  ...
 ◐  Updating the component devruntime
Applying URL configuration
 ✓  URL 127.0.0.1:59940 created

 ✓  Starting container for jeevandroid/open-liberty-dev:latest [474ms]
 ✓  Updating the component devruntime [749ms]
 ✓  Push devfile component openliberty [785ms]
```

Tested it worked on Windows, Linux and MacOS:

Windows: 
image

Linux:
image

MacOS: 
image


pkg/url/url.go
```diff
@@ -396,6 +398,36 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
...
+		for {
+			randomPort := rand.Intn(endPort-startPort) + startPort
+			listener, err := net.Listen("tcp", ":"+strconv.Itoa(randomPort))
+			if err == nil {
+				port = randomPort
+				listener.Close()
+				break
+			}
+		}
+		return port, nil
```
mik-dass, line 419: We already have a function in util for finding in utils package `HttpGetFreePort()`. Please use that here.

pkg/url/url.go
```diff
@@ -396,6 +398,36 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
 	return portNumber, fmt.Errorf("given port %d is not exposed on given component, available ports are: %s", portNumber, strings.Trim(strings.Replace(fmt.Sprint(componentPorts), " ", ",", -1), "[]"))
 }
 
+// GetValidExposedPortNumber checks if the given exposed port number is a valid port or not
+// if exposed port is not provided, a random 5 digit valid port will be generated and returned
+func GetValidExposedPortNumber(componentName string, exposedPort int) (int, error) {
```
mik-dass, line 403: componentName is not used in the function. Also this function should be in the util package.

yangcao77: Deleted `componentName` from the function parameter. 

IMO, the function is to valid the user provided exposed port, which should only be used for `odo url create`, so still should be in `url.go`

pkg/url/url.go
yangcao77: Thanks for pointing this out! Updated.

pkg/url/url.go
mik-dass: OK

pkg/url/url.go
mik-dass: OK cool :+1:

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -135,8 +140,11 @@ func (a Adapter) pullAndStartContainer(componentName string, projectVolumeName s
 
 func (a Adapter) startContainer(componentName string, projectVolumeName string, comp versionsCommon.DevfileComponent) error {
 	containerConfig := a.generateAndGetContainerConfig(componentName, comp)
-	hostConfig := container.HostConfig{}
 
+	hostConfig, err := a.generateAndGetHostConfig()
```
johnmcollier, line 144: 👍

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +174,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+					return nat.PortMap{}, err
+				}
+				portmap[port] = []nat.PortBinding{
+					nat.PortBinding{
+						HostIP:   "127.0.0.1",
+						HostPort: strconv.Itoa(element.ExposedPort),
+					},
+				}
+				log.Info("\nApplying URL configuration")
+				log.Successf("URL https://127.0.0.1:%v created\n", element.ExposedPort)
```
johnmcollier, line 215: We shouldn't hard code https here in case a non-secure port was exposed.

yangcao77: How can we tell if it's http or https then?

johnmcollier: Hmm, good point. Let's leave the scheme out for now then.

cdrage:
One big thing with both the previous docker implementation and this is that there is *zero* documentation.

Is there any way possible that yourself or someone on the IBM team with experience with docs can add a little document outlining the features as well as what's beeen implemented?

To be honest, we have to hunt through random PR's and issues in order to find out exactly what's been implemented and how it works.

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -10,9 +14,12 @@ import (
 	adaptersCommon "github.com/openshift/odo/pkg/devfile/adapters/common"
 	"github.com/openshift/odo/pkg/devfile/adapters/docker/utils"
 	versionsCommon "github.com/openshift/odo/pkg/devfile/parser/data/common"
+	"github.com/openshift/odo/pkg/envinfo"
 	"github.com/openshift/odo/pkg/log"
 )
 
+var LocalhostIP = "127.0.0.1"
```
cdrage, line 21: Have you tested this on both Windows, macOS and Linux platforms? I have a feeling that there may be an issue with one of the OS'

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+		return container.HostConfig{}, err
+	}
+	hostConfig := container.HostConfig{}
+	if len(portMap) > 0 {
+		hostConfig = a.Client.GenerateHostConfig(false, false, portMap)
+	}
+	return hostConfig, nil
+}
+
+func getPortMap() (nat.PortMap, error) {
```
cdrage, line 192: Need comments. Entire function has no comments on what's happening.

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
 	containerConfig := a.Client.GenerateContainerConfig(*comp.Image, comp.Command, comp.Args, envVars, containerLabels)
 	return containerConfig
 }
+
+func (a Adapter) generateAndGetHostConfig() (container.HostConfig, error) {
```
cdrage, line 180: Need more comments

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+				if err != nil {
+					return nat.PortMap{}, err
+				}
+				portmap[port] = []nat.PortBinding{
+					nat.PortBinding{
+						HostIP:   LocalhostIP,
+						HostPort: strconv.Itoa(element.ExposedPort),
+					},
+				}
+				log.Info("\nApplying URL configuration")
```
cdrage, line 216: `\n` shouldn't be here.

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+				log.Successf("URL %v:%v created\n", LocalhostIP, element.ExposedPort)
```
cdrage, line 217: Same as above, `\n` shouldn't be used in logs

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,49 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+			}
+		}
+		return portmap, nil
+	} else {
```
cdrage, line 221: don't need else here, you can put this as the last return statement in the function.

pkg/odo/cli/url/create.go
```diff
@@ -118,10 +122,18 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
...
-			o.urlName = url.GetURLName(componentName, o.componentPort)
+		if pushtarget.IsPushTargetDocker() {
+			o.urlName = "local-" + url.GetURLName(componentName, o.componentPort)
+			o.exposedPort, err = url.GetValidExposedPortNumber(o.exposedPort)
+			if err != nil {
+				return err
+			}
 		} else {
-			o.urlName = args[0]
+			if len(args) == 0 {
```
cdrage, line 132: no need for `else` statement.. you can do:

```go
		if pushtarget.IsPushTargetDocker() {
			o.urlName = "local-" + url.GetURLName(componentName, o.componentPort)
			o.exposedPort, err = url.GetValidExposedPortNumber(o.exposedPort)
			if err != nil {
				return err
			}
		} else if len(args) == 0 {
				o.urlName = url.GetURLName(componentName, o.componentPort)
		} else {
				o.urlName = args[0]
	       }
```

pkg/odo/cli/url/create.go
```diff
@@ -157,17 +169,18 @@ func (o *URLCreateOptions) Validate() (err error) {
...
-		for _, localURL := range o.EnvSpecificInfo.GetURL() {
-			curIngressDomain := fmt.Sprintf("%v.%v", o.urlName, o.host)
-			ingressDomainEnv := fmt.Sprintf("%v.%v", localURL.Name, localURL.Host)
-			if curIngressDomain == ingressDomainEnv {
-				return fmt.Errorf("the url %s already exists in the application: %s", curIngressDomain, o.Application)
+		if !pushtarget.IsPushTargetDocker() {
+			if len(o.host) <= 0 {
+				return fmt.Errorf("host must be provided in order to create ingress")
+			}
+			for _, localURL := range o.EnvSpecificInfo.GetURL() {
```
cdrage, line 176: no need for this to be inside the if statement, you can change the above to:

```go
		if !pushtarget.IsPushTargetDocker() && if len(o.host) <= 0 {
				return fmt.Errorf("host must be provided in order to create ingress")
		}
			
                  for _, localURL := range o.EnvSpecificInfo.GetURL() {
				curIngressDomain := fmt.Sprintf("%v.%v", o.urlName, o.host)
				ingressDomainEnv := fmt.Sprintf("%v.%v", localURL.Name, localURL.Host)
				if curIngressDomain == ingressDomainEnv {
					return fmt.Errorf("the url %s already exists", curIngressDomain)
			}
		}
```

which is more idiomatic

yangcao77: The ingress domain is only used if the `pushtarget` is not docker. So the for loop has to be within the if statement.

pkg/devfile/adapters/docker/component/utils.go
yangcao77: I followed the format of the log statement here: https://github.com/openshift/odo/blob/0df81788d9adf28e49c9193162310f4d66b4ef5a/pkg/component/component.go#L559

but since for local docker case, it's always create a new container, so it's not applying "changes", so I modified the statement to be `applying configuration`

pkg/devfile/adapters/docker/component/utils.go
yangcao77: removed this `\n`

pkg/devfile/adapters/docker/component/utils.go
yangcao77: removed

pkg/odo/cli/url/create.go

pkg/devfile/adapters/docker/component/utils.go
yangcao77: added comment for the function

pkg/devfile/adapters/docker/component/utils.go
yangcao77: added comment for the function

pkg/devfile/adapters/docker/component/utils.go
yangcao77: Tested on Windows, macoS and Linux. All worked as expected.

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -165,3 +176,51 @@ func (a Adapter) generateAndGetContainerConfig(componentName string, comp versio
...
+	hostConfig := container.HostConfig{}
+	if len(portMap) > 0 {
+		hostConfig = a.Client.GenerateHostConfig(false, false, portMap)
+	}
+	return hostConfig, nil
+}
+
+func getPortMap() (nat.PortMap, error) {
+	// Convert the exposed and internal port pairs saved in env.yaml file to PortMap
+	dir, err := os.Getwd()
```
mik-dass, line 195: why are we using os.Getwd()? Won't it be inaccurate when we support context flag?

pkg/url/url.go
```diff
@@ -396,6 +397,27 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
...
+	if exposedPort == -1 {
+		freePort, err := util.HttpGetFreePort()
+		if err != nil {
+			return -1, err
+		}
+		return freePort, nil
+	} else {
+		// check if the given port is available
+		listener, err := net.Listen("tcp", ":"+strconv.Itoa(exposedPort))
+		if err != nil {
```
mik-dass, line 413: `defer listener.Close()` right after creating it

pkg/url/url.go
```diff
@@ -396,6 +397,27 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
...
+			return -1, errors.Wrapf(err, "given port %d is not available, please choose another port", exposedPort)
+		}
+		listener.Close()
```
mik-dass, line 416: if error is thrown, it won't be reached and thus the listener won't be closed. Thus https://github.com/openshift/odo/pull/2906/files#r412672504

pkg/envinfo/envinfo.go
```diff
@@ -38,6 +38,8 @@ type EnvInfoURL struct {
 	Host string `yaml:"host,omitempty"`
 	// TLS secret name to create ingress to provide a secure URL
 	TLSSecret string `yaml:"TLSSecret,omitempty"`
+	// Exposed port number for docker container, required for local senario
```
mik-dass, line 41: scenarios

pkg/odo/cli/url/create.go
```diff
@@ -118,7 +132,13 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
 			return err
 		}
 
-		if len(args) == 0 {
+		if pushtarget.IsPushTargetDocker() {
```
mik-dass, line 135: please also add a check that the exposed port flag is not used when the push target is not docker

pkg/odo/cli/url/create.go
```diff
@@ -157,17 +177,18 @@ func (o *URLCreateOptions) Validate() (err error) {
...
-		if len(o.host) <= 0 {
-			return fmt.Errorf("host must be provided in order to create ingress")
-		}
-		for _, localURL := range o.EnvSpecificInfo.GetURL() {
-			curIngressDomain := fmt.Sprintf("%v.%v", o.urlName, o.host)
-			ingressDomainEnv := fmt.Sprintf("%v.%v", localURL.Name, localURL.Host)
-			if curIngressDomain == ingressDomainEnv {
-				return fmt.Errorf("the url %s already exists in the application: %s", curIngressDomain, o.Application)
+		if !pushtarget.IsPushTargetDocker() {
+			if len(o.host) <= 0 {
```
mik-dass, line 181: please also add a check that the exposed port flag is not used when the push target is not docker

pkg/url/url.go
```diff
@@ -396,6 +397,27 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
 	return portNumber, fmt.Errorf("given port %d is not exposed on given component, available ports are: %s", portNumber, strings.Trim(strings.Replace(fmt.Sprint(componentPorts), " ", ",", -1), "[]"))
 }
 
+// GetValidExposedPortNumber checks if the given exposed port number is a valid port or not
+// if exposed port is not provided, a random 5 digit valid port will be generated and returned
```
mik-dass, line 401: HttpGetFreePort() can assign a port which is not of 5 digits too

pkg/devfile/adapters/docker/component/utils_test.go
```diff
@@ -235,3 +238,113 @@ func TestStartContainer(t *testing.T) {
...
+
+			esi, err := envinfo.NewEnvSpecificInfo("")
+			if err != nil {
+				t.Error(err)
+			}
+			for _, element := range tt.urlValue {
+				err = esi.SetConfiguration("URL", element)
+				if err != nil {
+					t.Error(err)
+				}
```
mik-dass, line 326: This will be a write operation. Instead you can just do 

```
settings:=esi.GetComponentSettings()
settings.URL=&tt.urlValue
```

pkg/odo/cli/url/create.go
```diff
@@ -241,12 +285,19 @@ func NewCmdURLCreate(name, fullName string) *cobra.Command {
...
 	urlCreateCmd.Flags().BoolVarP(&o.secureURL, "secure", "", false, "creates a secure https url")
-	// if experimental mode is enabled, add more flags to support ingress creation based on devfile
+	// if experimental mode is enabled, add more flags to support ingress creation or docker application based on devfile
 	if experimental.IsExperimentalModeEnabled() {
-		urlCreateCmd.Flags().StringVar(&o.tlsSecret, "tls-secret", "", "tls secret name for the url of the component if the user bring his own tls secret")
-		urlCreateCmd.Flags().StringVarP(&o.host, "host", "", "", "Cluster ip for this URL")
+		if pushtarget.IsPushTargetDocker() {
+			urlCreateCmd.Flags().IntVarP(&o.exposedPort, "exposed-port", "", -1, "an external port to the application container")
+			urlCreateCmd.Flags().BoolVarP(&o.forceFlag, "force", "f", false, "Don't ask for confirmation, assign an exposed port directly")
+			urlCreateCmd.Args = cobra.MaximumNArgs(0)
```
mik-dass, line 293: Why can't we name the url? 
```
[mrinaldas@localhost nodejs-ex]$ odo url create example
Error: accepts at most 0 arg(s), received 1
Usage:
  odo url create [url name] [flags]
```
What am I missing? Please add a comment regarding that here.

pkg/devfile/adapters/docker/component/utils_test.go
yangcao77: It has to write to the file. Since `getPortMap()` will read from the env.yaml file.

johnmcollier:
@yangcao77 Can you rebase this PR and address merge conflicts?

pkg/devfile/adapters/docker/component/utils_test.go
mik-dass: OK :+1:

pkg/devfile/adapters/docker/component/utils.go
yangcao77: Good point! I think the same concern applies on https://github.com/openshift/odo/blob/5a26a56426b5d4e214c0817c2820350d9b862e0a/pkg/odo/cli/component/devfile.go#L117 and https://github.com/openshift/odo/blob/5a26a56426b5d4e214c0817c2820350d9b862e0a/pkg/odo/cli/component/devfile.go#L131
We need to change to use context to get the envinfo after context is supported in experimental mode.  I've added comments in the code, and also created an issue to track:https://github.com/openshift/odo/issues/2959

pkg/odo/cli/url/create.go
yangcao77: fixed.

pkg/url/url.go

pkg/envinfo/envinfo.go
yangcao77: updated

pkg/url/url.go
yangcao77: updated

yangcao77:
@mik-dass @cdrage I've pushed new commits based on the code review. Can you help review again?

pkg/devfile/adapters/docker/component/utils.go
mik-dass: Thanks for creating the issue :+1:

pkg/url/url.go
```diff
@@ -396,6 +397,27 @@ func GetValidPortNumber(componentName string, portNumber int, portList []string)
...
+			return -1, err
+		}
+		return freePort, nil
+	} else {
+		// check if the given port is available
+		listener, err := net.Listen("tcp", ":"+strconv.Itoa(exposedPort))
+		if err != nil {
+			return -1, errors.Wrapf(err, "given port %d is not available, please choose another port", exposedPort)
+		}
+		defer listener.Close()
```
mik-dass, line 416: This needs to be before the if statement

kadel:
@yangcao77 It looks like this PR pics up a lot of commits from @johnmcollier that were already merged as part of another PR  https://github.com/openshift/odo/pull/2806/commits

pkg/odo/cli/url/create.go
```diff
@@ -241,12 +286,18 @@ func NewCmdURLCreate(name, fullName string) *cobra.Command {
 	}
 	urlCreateCmd.Flags().IntVarP(&o.urlPort, "port", "", -1, "port number for the url of the component, required in case of components which expose more than one service port")
 	urlCreateCmd.Flags().BoolVarP(&o.secureURL, "secure", "", false, "creates a secure https url")
-	// if experimental mode is enabled, add more flags to support ingress creation based on devfile
+	// if experimental mode is enabled, add more flags to support ingress creation or docker application based on devfile
 	if experimental.IsExperimentalModeEnabled() {
-		urlCreateCmd.Flags().StringVar(&o.tlsSecret, "tls-secret", "", "tls secret name for the url of the component if the user bring his own tls secret")
-		urlCreateCmd.Flags().StringVarP(&o.host, "host", "", "", "Cluster ip for this URL")
+		if pushtarget.IsPushTargetDocker() {
+			urlCreateCmd.Flags().IntVarP(&o.exposedPort, "exposed-port", "", -1, "an external port to the application container")
```
kadel, line 292: How is this flag going to work for K8s and OpenShift?
If this is valid only for local than the description should mention that.

kadel: We could also consider hiding flags based on `pushtarget` settings.

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -187,6 +194,10 @@ func (a Adapter) startContainer(mounts []mount.Mount, projectVolumeName string,
 		Mounts: mounts,
 	}
 
+	hostConfig, err := a.generateAndGetHostConfig()
```
johnmcollier, line 197: You're overwriting the hostconfig added in the earlier step, which adds the volume mounts to the host config, causing the volumes defined in the devfile to not get mounted to the container. I think we need to rework this to either:
- Pass in the `mounts` object to `generateAndGetHostConfig`, or:
- Move the line that sets the mounts down to below this line.

pkg/odo/cli/url/create.go
yangcao77: this flag is only added if `pushtarget.IsPushTargetDocker()`. It won't be valid for kube cases.

pkg/devfile/adapters/docker/component/utils.go
yangcao77: updated

yangcao77:
@kadel I have combined all my changes into one single commit to clean up the git history. 
Now this PR only include my commits.

pkg/odo/cli/url/create.go
cdrage: ah, that makes sense.

pkg/odo/cli/url/create.go
```diff
@@ -197,7 +219,25 @@ func (o *URLCreateOptions) Validate() (err error) {
...
 func (o *URLCreateOptions) Run() (err error) {
 	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
-		err = o.EnvSpecificInfo.SetConfiguration("url", envinfo.EnvInfoURL{Name: o.urlName, Port: o.componentPort, Host: o.host, Secure: o.secureURL, TLSSecret: o.tlsSecret})
+		if pushtarget.IsPushTargetDocker() {
+			for _, localURL := range o.EnvSpecificInfo.GetURL() {
+				if o.urlPort == localURL.Port && &localURL.ExposedPort != nil {
+					if !o.forceFlag {
+						if !ui.Proceed(fmt.Sprintf("Port %v already has an exposed port %v set for it. Do you want to override the exposed port", localURL.Port, localURL.ExposedPort)) {
+							log.Info("Aborted by the user")
+							return nil
```
cdrage, line 228: If the previous statement was Aborted by the user, shouldn't this be returning an error?

yangcao77: IMO, if the port is already bound with an exposed port, this operation is just set a new exposed port. (similar to the set operation: https://github.com/openshift/odo/blob/f52328b9d4c996ba2c0647b6fb0bb91900f8a2bd/pkg/odo/cli/config/set.go#L125)
If user select not to overwrite, then the user knows there won't be any changes to the current env.yaml file. So it should not be an error.

johnmcollier:
It seems I need to be connected to a Kube cluster for `odo url create` to work on Docker. I was testing out the changes with my vpn off (and my kube context set to a cluster behind the firewall), I kept getting the following error:
```
Johns-MacBook-Pro-3:springboot johncollier$ odo url create
 ✗  listen tcp: lookup localhost: no such host
```

I logged in to the vpn and it worked:
```
Johns-MacBook-Pro-3:springboot johncollier$ odo url create
 ✓  URL java-spring-boot-8080 created for component: java-spring-boot with exposed port: 51193

To create the URL on the cluster, please use `odo push`
```

johnmcollier:
@yangcao77 I think I found another issue:

I pulled your branch into master and was finding that `odo push`'s after running `odo url create` wasn't restarting the container and exposing the port. So, I think you need to update the `DoesContainerNeedUpdating` function to see if any ports need to be exposed for that container (see https://github.com/openshift/odo/blob/e28aaf7468cf6869ac2002f398964f505d968a19/pkg/devfile/adapters/docker/component/utils.go#L145 and https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/docker/utils/utils.go#L38)


The reason we didn't see this before is that my [volume PR for docker](https://github.com/openshift/odo/pull/2911) fixed a bug where container's were always getting restarted with each `odo push`, now they only get restarted if needed.

yangcao77:
@johnmcollier I have a fix for the issue you found. Modified `DoesContainerNeedUpdating` to check for devfile endpoints update, also checks for exposedPort difference between `HostConfig` of existing container and value in env.yaml file. 
Also added a couple new integration test cases for those changes. 

Please help review again.

mik-dass:
> @mik-dass @cdrage I've pushed new commits based on the code review. Can you help review again?

LGTM. 

@yangcao77 Please address the comments by @johnmcollier .

pkg/odo/cli/url/create.go
kadel: I missed the `if` statement, or it was not there before?

kadel:
/hold 


We need properly handle the following situations

### URL created for Docker but pushed to Kube
```
▶ odo preference set pushtarget docker
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated

▶ odo url create
 ✓  URL backend-8080 created for component: backend with exposed port: 49213

To create the URL on the cluster, please use `odo push`


▶ odo preference set pushtarget kube
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated

▶ odo push
 •  Push devfile component backend  ...
 ✓  Waiting for component to start [24s]

Applying URL changes
 ✗  Failed To Update Config To Component Deployed.
Error: unable to create url: unable to create ingress: error creating ingress: Ingress.extensions "backend-8080" is invalid: spec.rules[0].host: Invalid value: "backend-8080.": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a
```

The error message is confusing, it doesn't say anything useful. It should detect what happened, and tell user what to do.
This PR https://github.com/openshift/odo/pull/2885 is adding URL `Kind` to `env.yaml` which might be useful here.



### URL created for Kube but pushed to Docker

```
▶ odo url create  --host apps.testocp4x.psiodo.net
 ✓  URL backend-8080.apps.testocp4x.psiodo.net created for component: backend

To create the URL on the cluster, please use `odo push`

▶ odo preference set pushtarget docker
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated

▶ odo push
 •  Push devfile component backend  ...
 ✓  Pulling image quay.io/eclipse/che-java11-maven:nightly [2s]
 ✓  Starting container for quay.io/eclipse/che-java11-maven:nightly [595ms]
 ✓  Push devfile component backend [2s]
```

Url was created but it was silently ignored :-/

 

### secure for docker
```
▶ odo url create -h

      --secure             Creates a secure https url
```

How is `--secure` going to work on docker?

kadel:
The issues that I mentioned in the previous comment doesn't have to be a blocker for this.
I've created https://github.com/openshift/odo/issues/2977 to chapter this. We can follow up on that later.

/hold cancel
/approve

yangcao77:
@kadel I've addressed the issue in this PR. 

**URL created for docker but push to kube** 

```
$ odo url create --port 8080
 ✓  URL local-java-spring-boot-8080 created for component: java-spring-boot with exposed port: 61981

$ cat .odo/env/env.yaml 
ComponentSettings:
  Name: java-spring-boot
  Url:
  - Name: local-java-spring-boot-8080
    Port: 8080
    ExposedPort: 61981
$ odo preference set pushtarget kube
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated
$ odo url create local-java-spring-boot-8080 --host 1.2.3.4.com --port 8080
 ✗  the url local-java-spring-boot-8080 already exists with a different push target. Please recreate one using different urlName
```

**URL created for kube but push to docker** 
```
$ odo url create testurl1 --host 1.2.3.4.com --port 9090
 ✓  URL testurl1.1.2.3.4.com created for component: java-spring-boot

e$ cat .odo/env/env.yaml 
ComponentSettings:
  Name: java-spring-boot
  Url:
  - Name: local-java-spring-boot-8080
    Port: 8080
    ExposedPort: 61981
  - Name: testurl1
    Port: 9090
    host: 1.2.3.4.com
$ odo preference set pushtarget docker
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated
$ odo url create testurl1 --port 9090
 ✗  the url testurl1 already exists with a different push target. Please rerun the command using a different urlName.

```


Also remove `--secure` flag if the push target is docker.

PR #2907, opened 2020-04-16
Adding init command capability to devfiles

### LINKED ISSUE #2672: devfile init command support
## User Story
As a devfile provider, I want to define some component initialization code so that the init code will be run as part of the odo create component command during component creation.

## Background
The existing Appsody CLI has an Appsody init command which will do extra initialization work in order to setup the environment required for developing the specific component.  This only applies to certain Appsody stacks. This feature is to allow us to provide the same level of function in odo devfile scenario.

## Acceptance Criteria 
- [ ] It should execute the `devInit` command when creating a component to do all the necessary setup for the given component.
- [ ] This `devInit` command will only be run if the following conditions are meet:

1. `Experimental` flag has been set to true

1. The component that is being created is the devfile style of component
 
1. The devfile of the component contains a command called `devInit`

## Links 
- Parent Epic: #2576

/kind user-story

upLukeWeston:
**What type of PR is this?**
 /kind feature

Implements the use of `devinit` commands to be ran on an `odo push` when giving a `devfile`.

Fixes #2672

Added unit and integrations testing for this feature.


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +425,108 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-				if show {
-					s = log.SpinnerNoSpin("Executing " + command.Name + " command " + fmt.Sprintf("%q", *action.Command))
+	commandOrder := []CommandNames{
+		CommandNames{defaultName: string(common.DefaultDevfileInitCommand), adapterName: a.devfileInitCmd},
+		CommandNames{defaultName: string(common.DefaultDevfileBuildCommand), adapterName: a.devfileBuildCmd},
+		CommandNames{defaultName: string(common.DefaultDevfileRunCommand), adapterName: a.devfileRunCmd},
+	}
+	for i, currentCommand := range commandOrder {
+		for _, command := range pushDevfileCommands {
+			if command.Name == currentCommand.defaultName || command.Name == currentCommand.adapterName {
```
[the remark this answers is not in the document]
upLukeWeston: We no longer need the `buildRequired` bool anymore because we verify that the commands are required in `ValidateAndGetPushDevfileCommands` - an `init` can't be provided without a `build`, so we don't add it to the array of commands, and like the old code we currently make the assumption that if only one command is given it is `run`. This means the only possible combinations of `pushDevfileCommands` are: [init build run], [build run], and [run]. Since this function is internal and only gets called once after we retrieve the commands, it should be fine. However, if this function would be called in a different scenario, an additional check would be required.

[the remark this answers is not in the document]
upLukeWeston: On a related note: should we error out if the `pushDevfileCommands` slice is empty?

pkg/devfile/adapters/common/command.go
```diff
@@ -134,20 +147,39 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
...
 	buildCommand, buildCmdErr := GetBuildCommand(data, devfileBuildCmd)
-
 	if reflect.DeepEqual(emptyCommand, buildCommand) && buildCmdErr == nil {
 		// If there was no build command specified through odo push and no default build command in the devfile, default validate to true since the build command is optional
 		isBuildCommandValid = true
 		glog.V(3).Infof("No build command was provided")
+
+		// If init command was provided, and no build command provide log a message
+		if !reflect.DeepEqual(emptyCommand, initCommand) && initCmdErr == nil {
+			glog.V(3).Infof("Not adding init command provided, since build command wasn't provided.")
```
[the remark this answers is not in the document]
maysunfaisal: Think we should error out, since you suggested in the other thread on this PR that init command should be accompanied by a build command

pkg/devfile/adapters/kubernetes/component/adapter.go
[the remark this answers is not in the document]
maysunfaisal: the validate function should handle that, you can try it out with your PR to double confirm

pkg/devfile/adapters/common/command.go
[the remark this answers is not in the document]
upLukeWeston: Okay, no problem!

pkg/devfile/adapters/kubernetes/component/adapter.go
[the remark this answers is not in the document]
upLukeWeston: You're right, it does look like the Validate function checks that the run command exists, I can add an additional check in the  `execDevfile` function too for good measure though.

pkg/devfile/adapters/common/command.go
```diff
@@ -134,20 +147,43 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
...
+	isBuildCmdEmpty := reflect.DeepEqual(emptyCommand, buildCommand)
+	if isBuildCmdEmpty && buildCmdErr == nil {
 		// If there was no build command specified through odo push and no default build command in the devfile, default validate to true since the build command is optional
 		isBuildCommandValid = true
 		glog.V(3).Infof("No build command was provided")
-	} else if !reflect.DeepEqual(emptyCommand, buildCommand) && buildCmdErr == nil {
+
+		// If init command was provided, and no build command provide log a message
+		if !isInitCmdEmpty && initCmdErr == nil {
+			return []common.DevfileCommand{}, errors.New(fmt.Sprint("Devfile init command provided without accompanied build command."))
```
kadel, line 176: Why the init command requires the build command?

upLukeWeston:
@kadel good catch! We've changed it to only run when the component is created.

pkg/devfile/adapters/common/command.go
```diff
@@ -107,6 +107,19 @@ func validateAction(data data.DevfileData, action common.DevfileCommandAction) (
...
+func GetInitCommand(data data.DevfileData, devfileInitCmd string) (initCommand common.DevfileCommand, err error) {
+	if devfileInitCmd != "" {
+		// a init command was specified so if it is not found then it is an error
+		initCommand, err = getCommand(data, devfileInitCmd, true)
+	} else {
+		// a init command was not specified so if it is not found then it is not an error
+		initCommand, err = getCommand(data, string(DefaultDevfileInitCommand), false)
+	}
+
+	return
```
cdrage, line 120: you don't check for `err` and you return just `return`.

You should be doing `return err`

pkg/devfile/adapters/common/command.go
```diff
@@ -134,17 +147,32 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
...
-	isBuildCommandValid, isRunCommandValid := false, false
+	isInitCommandValid, isBuildCommandValid, isRunCommandValid := false, false, false
+
+	initCommand, initCmdErr := GetInitCommand(data, devfileInitCmd)
+
+	isInitCmdEmpty := reflect.DeepEqual(emptyCommand, initCommand)
+	if isInitCmdEmpty && initCmdErr == nil {
+		// If there was no init command specified through odo push and no default init command in the devfile, default validate to true since the init command is optional
+		isInitCommandValid = true
+		glog.V(3).Infof("No init command was provided")
```
cdrage, line 160: shouldn't be using `v(3)`

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-	if len(pushDevfileCommands) == 1 {
-		// if there is one command, it is the mandatory run command. No need to build.
-		buildRequired = false
-	} else if len(pushDevfileCommands) == 2 {
-		// if there are two commands, it is the optional build command and the mandatory run command, set buildRequired to true
-		buildRequired = true
-	} else {
-		return fmt.Errorf("error executing devfile commands - there should be at least 1 command or at most 2 commands, currently there are %v commands", len(pushDevfileCommands))
+	if len(pushDevfileCommands) == 0 {
+		return errors.New(fmt.Sprint("error executing devfile commands - there should be at least 1 command."))
```
cdrage, line 435: no periods after error message

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
 	return pod, nil
 }
 
-// Push syncs source code from the user's disk to the component
-func (a Adapter) execDevfile(pushDevfileCommands []versionsCommon.DevfileCommand, componentExists, show bool, podName string, containers []corev1.Container) (err error) {
-	var buildRequired bool
+// Executes Devfile Commands
```
cdrage, line 430: you removed/ changed the comments. Could you go into more detail what actually happens in the function?

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+		CommandNames{defaultName: string(common.DefaultDevfileBuildCommand), adapterName: a.devfileBuildCmd},
+		CommandNames{defaultName: string(common.DefaultDevfileRunCommand), adapterName: a.devfileRunCmd},
+	)
 
-				if show {
-					s = log.SpinnerNoSpin("Executing " + command.Name + " command " + fmt.Sprintf("%q", *action.Command))
+	for i, currentCommand := range commandOrder {
+		for _, command := range pushDevfileCommands {
+			if command.Name == currentCommand.defaultName || command.Name == currentCommand.adapterName {
+				if i < len(commandOrder)-1 {
```
cdrage, line 457: This seems kind of odd to be honest. What if we reach a `slice out of bounds error` if the len is 0?

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-				}
-
-				defer s.End(false)
-
-				err = exec.ExecuteCommand(&a.Client, podName, *action.Component, cmdArr, show)
-				if err != nil {
-					s.End(false)
-					return err
+					// Last command is "Run"
+					glog.V(3).Infof("Executing devfile command %v", command.Name)
```
cdrage, line 465: 4

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+						// Start the supervisord if the odo component does not exist
+						if !componentExists {
+							err = a.InitRunContainerSupervisord(*action.Component, podName, containers)
+							if err != nil {
+								return
+							}
+						}
+
+						// Exec the supervisord ctl stop and start for the devrun program
+						type devRunExecutable struct {
```
cdrage, line 479: structs should be defined at the top of the file / public struct. Would make this function a bit more tidy.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+							command []string
+						}
+						devRunExecs := []devRunExecutable{
+							{
+								command: []string{common.SupervisordBinaryPath, "ctl", "stop", "all"},
```
cdrage, line 484: what's `ctl`? comment? Shouldn't we make this a `const`?

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+							},
+							{
+								command: []string{common.SupervisordBinaryPath, "ctl", "start", string(common.DefaultDevfileRunCommand)},
+							},
+						}
+
+						s = log.Spinner("Executing " + command.Name + " command " + fmt.Sprintf("%q", *action.Command))
```
cdrage, line 491: Instead of `+` each string. Why not put the entire thing in a `fmt.Sprintf`?

Ex: `log.Spinner(fmt.Sprintf("Executing %s %s, etc, etc.", command.Name, *action.Command),` etc.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-					return err
+					// Last command is "Run"
+					glog.V(3).Infof("Executing devfile command %v", command.Name)
+
+					for _, action := range command.Actions {
+
+						// Check if the devfile run component containers have supervisord as the entrypoint.
+						// Start the supervisord if the odo component does not exist
+						if !componentExists {
+							err = a.InitRunContainerSupervisord(*action.Component, podName, containers)
```
cdrage, line 472: We should check that `*action.Component` isn't nil or else we'll get some race errors.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-				// Start the supervisord if the odo component does not exist
-				if !componentExists {
-					err = a.InitRunContainerSupervisord(*action.Component, podName, containers)
-					if err != nil {
-						return
-					}
-				}
+func (a Adapter) executeDevfileCommand(command versionsCommon.DevfileCommand, show bool, podName string) error {
+	var s *log.Status
+	glog.V(3).Infof("Executing devfile command %v", command.Name)
```
cdrage, line 513: 4

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
 
-				// Check if the devfile run component containers have supervisord as the entrypoint.
-				// Start the supervisord if the odo component does not exist
-				if !componentExists {
-					err = a.InitRunContainerSupervisord(*action.Component, podName, containers)
-					if err != nil {
-						return
-					}
-				}
+func (a Adapter) executeDevfileCommand(command versionsCommon.DevfileCommand, show bool, podName string) error {
```
cdrage, line 511: need comments :)

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-					},
-					{
-						command: []string{common.SupervisordBinaryPath, "ctl", "start", string(common.DefaultDevfileRunCommand)},
-					},
-				}
+	for _, action := range command.Actions {
+		// Change to the workdir and execute the command
+		var cmdArr []string
+		if action.Workdir != nil {
+			cmdArr = []string{"/bin/sh", "-c", "cd " + *action.Workdir + " && " + *action.Command}
```
cdrage, line 519: `/bin/sh` should be a const.. See the code you deleted and how we used `common.SupervisordBinaryPath`

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -423,102 +427,117 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+		} else {
+			cmdArr = []string{"/bin/sh", "-c", *action.Command}
+		}
 
-				s = log.Spinner("Executing " + command.Name + " command " + fmt.Sprintf("%q", *action.Command))
-				defer s.End(false)
+		if show {
+			s = log.SpinnerNoSpin("Executing " + command.Name + " command " + fmt.Sprintf("%q", *action.Command))
```
cdrage, line 525: use `fmt.Sprintf` for everything, much cleaner.

pkg/devfile/adapters/common/command_test.go
```diff
@@ -38,7 +38,35 @@ func TestGetCommand(t *testing.T) {
 					Type:      &validCommandType,
 				},
 			},
-			isCommandRequired: []bool{false, true},
+			isCommandRequired: []bool{false, false, true},
+			wantErr:           false,
+		},
+		{
+			name:              "Case: Valid devfile with devinit and devbuild",
```
cdrage, line 45: Case 1, Case 2, etc.

pkg/devfile/adapters/common/command_test.go
```diff
@@ -38,7 +38,35 @@ func TestGetCommand(t *testing.T) {
...
+					Component: &components[0],
+					Workdir:   &workDir[0],
+					Type:      &validCommandType,
+				},
+			},
+			isCommandRequired: []bool{false, false, true},
+			wantErr:           false,
+		},
+		{
+			name:              "Case: Valid devfile with devinit and devrun",
```
cdrage, line 59: Case 1, Case 2, etc.

pkg/devfile/adapters/common/command.go
EnriqueL8: Since this function uses a "named" return value of `(initCommand common.DevfileCommand, err error)`, the function will return those two variables. Within the function, these two variables will be set by either calls to the `getCommand`. We will be simply returning to same values back after they have been set.  https://golang.org/ref/spec#Return_statements

pkg/devfile/adapters/common/command.go
EnriqueL8: Is there a guideline for using different log levels?

pkg/devfile/adapters/kubernetes/component/adapter.go
EnriqueL8: If the `len(commandOrder)` is zero, the for loop will never run which means this code will not get executed. Here we are just checking that it's not the last command.

pkg/devfile/adapters/common/command.go
johnmcollier: I believe odo mostly uses `v(4)`

upLukeWeston: Okay, thanks! The existing code used `v(3)`, so we just copied the standard there. I'll update them now though.

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: Already being done during push by validate func https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/common/command.go#L85

@EnriqueL8 tho, just confirm this by removing a command's component before doing odo push

tests/examples/source/devfiles/nodejs/devfile-with-volumes.yaml
```diff
@@ -33,6 +33,12 @@ components:
       - name: myvol2
         containerPath: /data2
 commands:
+  - name: devinit
```
adisky, line 36: As I know, we are following `devInit`, `devBuild`, `devRun` in yamls.

pkg/devfile/adapters/common/command.go
```diff
@@ -134,17 +147,32 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
 // ValidateAndGetPushDevfileCommands validates the build and the run command,
 // if provided through odo push or else checks the devfile for devBuild and devRun.
 // It returns the build and run commands if its validated successfully, error otherwise.
-func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileBuildCmd, devfileRunCmd string) (pushDevfileCommands []common.DevfileCommand, err error) {
+func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, devfileBuildCmd, devfileRunCmd string) (pushDevfileCommands []common.DevfileCommand, err error) {
 	var emptyCommand common.DevfileCommand
-	isBuildCommandValid, isRunCommandValid := false, false
+	isInitCommandValid, isBuildCommandValid, isRunCommandValid := false, false, false
+
+	initCommand, initCmdErr := GetInitCommand(data, devfileInitCmd)
```
girishramnani, line 154: can we please add check for `initCmdErr != nil`? and can't we just call it `err`?

pkg/devfile/adapters/common/command.go
```diff
@@ -107,6 +107,19 @@ func validateAction(data data.DevfileData, action common.DevfileCommandAction) (
 	return
 }
 
+// GetInitCommand iterates through the components in the devfile and returns the init command
+func GetInitCommand(data data.DevfileData, devfileInitCmd string) (initCommand common.DevfileCommand, err error) {
+	if devfileInitCmd != "" {
+		// a init command was specified so if it is not found then it is an error
+		initCommand, err = getCommand(data, devfileInitCmd, true)
```
girishramnani, line 114: you can directly return the `getCommand(...)` instead of assigning it to anything.
And once you do that golint will complain about not having an `else` clause as if `returns`. so please do that change too.

girishramnani: also if possible can we change this for  `GetBuildCommand` and `GetRunCommand`, those are not part of this PR but it would be great to have some consistency :slightly_smiling_face: 
ref - https://github.com/openshift/odo/blob/51134ae769c00463a584df6c79887b01572d43c2/pkg/devfile/adapters/common/command.go#L124

pkg/devfile/adapters/common/command.go
```diff
@@ -134,17 +147,32 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
...
+		// If there was no init command specified through odo push and no default init command in the devfile, default validate to true since the init command is optional
+		isInitCommandValid = true
+		glog.V(3).Infof("No init command was provided")
+	} else if !isInitCmdEmpty && initCmdErr == nil {
+		isInitCommandValid = true
+		pushDevfileCommands = append(pushDevfileCommands, initCommand)
+		glog.V(3).Infof("Init command: %v", initCommand.Name)
+	}
 
 	buildCommand, buildCmdErr := GetBuildCommand(data, devfileBuildCmd)
```
girishramnani, line 167: can we please add check for `buildCmdErr`? this would simplify the `isBuildCmdEmpty && buildCmdErr == nil` statement and make it more readable.

pkg/devfile/adapters/common/command.go
girishramnani: this will simplify `isInitCmdEmpty && initCmdErr == nil` statement below

upLukeWeston: Currently it checks if `initCmdErr != nil` on line 191, because the errors are only amended if any of the commands are invalid, to match the existing flow.

pkg/devfile/adapters/common/command.go
```diff
@@ -134,17 +147,32 @@ func GetRunCommand(data data.DevfileData, devfileRunCmd string) (runCommand comm
 // ValidateAndGetPushDevfileCommands validates the build and the run command,
 // if provided through odo push or else checks the devfile for devBuild and devRun.
 // It returns the build and run commands if its validated successfully, error otherwise.
-func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileBuildCmd, devfileRunCmd string) (pushDevfileCommands []common.DevfileCommand, err error) {
+func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, devfileBuildCmd, devfileRunCmd string) (pushDevfileCommands []common.DevfileCommand, err error) {
```
girishramnani, line 150: I would suggest refactoring all the `Get*Command` to return a `*common.DevfileCommand` as we depend so highly on them to not be empty, checking if the command is `nil` or not will remove most of `reflect` code and increase readability.

girishramnani: not sure if it should be done as part of this PR, I would settle for the quickest way.

pkg/devfile/adapters/common/command.go
```diff
@@ -158,8 +186,11 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileBuildCmd, d
 	}
 
 	// If either command had a problem, return an empty list of commands and an error
-	if !isBuildCommandValid || !isRunCommandValid {
+	if !isInitCommandValid || !isBuildCommandValid || !isRunCommandValid {
 		commandErrors := ""
+		if initCmdErr != nil {
+			commandErrors += initCmdErr.Error()
```
girishramnani, line 192: should we add a new line between each error so they are more readable? or maybe consider `errors.Wrap`?

pkg/devfile/adapters/common/command.go
EnriqueL8: We did notice that `reflect.DeepEqual` introduced more complexity. That is a good idea to solve it, but I would recommend doing it in another PR for all the commands.

pkg/devfile/adapters/common/command.go
upLukeWeston: Same as the `initCmdErr`, we check if `buildCmdErr` is nil further down in the flow. It would be possible to refactor the code to accommodate this though, to make it a bit cleaner.

girishramnani:
/approve
Please create and link the issue for the follow up work.

pkg/odo/cli/component/push.go
```diff
@@ -40,10 +40,10 @@ type PushOptions struct {
 	DevfilePath string
 
 	// devfile commands
+	devfileInitCommand  string
```
maysunfaisal, line 43: This command was added to push.go and it is also accepting a custom command in function `GetInitCommand()` but `odo push` does not have a new flag to pass the custom init command from the command line.. Users can also push with a devfile without needing to always have a `devinit` command.. This need to be fixed before being merged.

EnriqueL8:
@maysunfaisal changes made

johnmcollier:
odo v1.2 is now out (and PR #2922 merged)

/hold cancel

johnmcollier:
@upLukeWeston Please rebase on master

johnmcollier:
@upLukeWeston Can you rebase your PR once again?

Issue #2769, opened 2020-03-27
Devfile odo push should use --project instead of --namespace


rajivnathan:
/kind bug
/area devfile

## Description

`odo create` uses the `--project` option to specify a project/namespace but `odo push` uses `--namespace`. Push should be updated to use `--project` as previously agreed upon.

kadel:
We need to be a little bit careful about this. 
It will be a big breaking change!
It needs to be coordinated with https://github.com/redhat-developer/intellij-openshift-connector and https://github.com/redhat-developer/vscode-openshift-tools (cc: @dgolovin and @jeffmaury )

Some kind of slow depreciation might be good here.

For example, in the first stage, we could just hide `--project` from help output but it would still work.

Or I would maybe even consider keeping both flags as for OpenShift users the `--project` might make more sense and `--namespace` might be just confusing.

rajivnathan:
@kadel IIRC what was previously discussed was to just support `--project` and add support for `--namespace` later which is what this issue's description was saying. Although, I may not have worded it that well. :) Supporting both is also fine but just wanted to clarify I was not suggesting we switch to use --namespace.

kadel:
> @kadel IIRC what was previously discussed was to just support `--project` and add support for `--namespace` later which is what this issue's description was saying. Although, I may not have worded it that well. :) Supporting both is also fine but just wanted to clarify I was not suggesting we switch to use --namespace.


Oh sorry, I didn't read the description carefully and assumed that this is about switching from `--project` to `--namespace`


Yes, for now, let's just use `--project` everywhere to be consistent.

We can figure out later what to do with namespace

PR #2909, opened 2020-04-16
[Devfile] Remove namespace flag from odo commands

johnmcollier:
**What type of PR is this?**

/kind cleanup
/area devfile

As mentioned earlier, we decided to use only the `--project` flag in odo for devfile components and to not use `--namespace` at this moment. This PR removes the `--namespace` flag that was added to certain odo commands for use with devfile components and uses `--project` instead. For commands that didn't have the `--project` flag added to it, the flag was added. 

Integration tests updated to use `--project` only as well.

Fixes https://github.com/openshift/odo/issues/2769

- Run the unit and integration tests
- `odo create --project <project>` -> Verify namespace/project in env.yaml is correct
- `odo push --project <project>` -> should push to the proper namespace
- `odo delete --project <project>` -> should delete from the proper namespace
- `odo watch --project <project>` -> should push changes to the proper namespace


johnmcollier:
/retest 

More OpenShift CI flakiness

pkg/odo/cli/component/push.go
```diff
@@ -65,6 +66,13 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string)
 		}
 		po.EnvSpecificInfo = envinfo
 		po.Context = genericclioptions.NewDevfileContext(cmd)
+
+		if cmd.Flags().Changed("project") {
```
mik-dass, line 70: I think this needs to be moved here https://github.com/openshift/odo/blob/6740e5177ae11db734f7d3d303c4f689015e3b91/pkg/odo/genericclioptions/context.go#L343. We are repeating the same steps in all the above files.

johnmcollier: @mik-dass You're right, good catch!

johnmcollier: Actually, we don't need to move any code to `resolveNamespace`, just update it to retrieve the namespace from the `--project` flag.

So long as we initialize the devfile context in each command that needs it, we can retrieve the namespace via `kclient.Namespace`

johnmcollier:
More CI infra flakes 😢 

/retest

mik-dass:
@johnmcollier needs rebase

johnmcollier:
/retest

Giving the builds a kick due to the gosec issues earlier

johnmcollier:
@mik-dass I've rebased and addressed your review comment (retrieving namespace from the getNamespace function)

johnmcollier:
There may be a potential merge conflict with this and https://github.com/openshift/odo/issues/2922, so I'm holding this so that it doesn't slip in to this afternoon's release and delay 2922. I'll cancel the hold after the release.

/hold

maysunfaisal:
tests were updated! Thx
/lgtm

PR #2917, opened 2020-04-17
--context support for odo create

### LINKED ISSUE #2824: "--context" support for odo create
/kind user-story

## User Story
As a user, I want to create devfile component from outside the source code directory using the context flag.

## Acceptance Criteria 
- [ ] `odo create` for devfiles should support context and should work outside of the component directory (eg `odo create nodejs --context source-dir/`)

## Links 
- Related issue: https://github.com/openshift/odo/issues/2813

/kind user-story
/area devfile

GeekArthur:

/kind feature
/area devfile

Add `--context` support for devfile component creation to allow user specify the component context

Fixes #2824 

`odo create <component type> --context <context path>`


tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -82,6 +84,21 @@ var _ = Describe("odo devfile create command tests", func() {
 		})
 	})
 
+	Context("When executing odo create with devfile component type argument and --context flag", func() {
+		It("should successfully create the devfile componet in the context", func() {
+			newContext := path.Join(context, "newContext")
+			devfilePath := filepath.Join(newContext, "devfile.yaml")
+			envFilePath := filepath.Join(newContext, ".odo/env/env.yaml")
+			helper.MakeDir(newContext)
```
amitkrout, line 92: @GeekArthur +1, however you need to add the cleanup steps for the dir ```newContext``` at the end of this spec. For example you just need to add ```helper.DeleteDir(newContext)``` after line number 98.

pkg/odo/cli/component/create.go
```diff
@@ -313,6 +315,19 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+		if cmd.Flags().Changed("context") {
+			context, err := cmd.Flags().GetString("context")
+			if err != nil {
+				return err
+			}
+
+			DevfilePath = filepath.Join(context, "devfile.yaml")
+			EnvFilePath = filepath.Join(context, ".odo/env/env.yaml")
+			co.componentContext = context
+			co.CommonPushOptions.componentContext = context
```
girishramnani, line 328: we shouldn't have multiple contexts assignments here

girishramnani: and can we use `genericclioptions.AddContextFlag` helper here instead?

tests/integration/devfile/cmd_devfile_create_test.go
GeekArthur: @amitkrout Actually `newContext` is subdir of `context` because of its assignment here `newContext := path.Join(context, "newContext")`, and we clean up the `context` dir directly here https://github.com/openshift/odo/pull/2917/files#diff-5291a6a42f2148d4c3e48ea4b73cff23L33 in `AfterEach` statement, which means `newContext` dir cleanup will be covered by `context` dir cleanup. That being said, I can still add `newContext` cleanup explicitly for readability.

pkg/odo/cli/component/create.go
GeekArthur: @girishramnani we can't use `genericclioptions.AddContextFlag ` here because we already used that in cobra command creation section and use here again will cause re-define cobra command issue. But I refactor the code with my latest commit to avoid multiple contexts assignments

GeekArthur:
@amitkrout @girishramnani Comments addressed, please review again, thanks!

pkg/odo/cli/component/create.go
```diff
@@ -71,11 +71,11 @@ const CreateRecommendedCommandName = "create"
...
 
-// DevfilePath is the default path of devfile.yaml
-const DevfilePath = "./devfile.yaml"
+// DevfilePath is the path of devfile.yaml, the default path is "./devfile.yaml"
+var DevfilePath = filepath.Join(LocalDirectoryDefaultLocation, "devfile.yaml")
 
-// EnvFilePath is the default path of env.yaml for devfile component
-const EnvFilePath = "./.odo/env/env.yaml"
+// EnvFilePath is the path of env.yaml for devfile component, the defult path is "./.odo/env/env.yaml"
+var EnvFilePath = filepath.Join(LocalDirectoryDefaultLocation, ".odo/env/env.yaml")
```
johnmcollier, line 78: This just got me thinking, are any of these directories safe to use on Windows, as we're using Unix paths.

@jgwest thoughts?

GeekArthur: From the `filepath` docs https://golang.org/pkg/path/filepath/#Join, I think `filepath.Join` can handle the path conversion to the target OS

pkg/odo/cli/component/create.go
```diff
@@ -71,11 +71,11 @@ const CreateRecommendedCommandName = "create"
 // since the application will always be in the same directory as `.odo`, we will always set this as: ./
 const LocalDirectoryDefaultLocation = "./"
 
-// DevfilePath is the default path of devfile.yaml
-const DevfilePath = "./devfile.yaml"
+// DevfilePath is the path of devfile.yaml, the default path is "./devfile.yaml"
+var DevfilePath = filepath.Join(LocalDirectoryDefaultLocation, "devfile.yaml")
```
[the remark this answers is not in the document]
GeekArthur: IIRC some functions still need that, but this could become local variable in my `--devfile` support PR since those functions can always use `./devfile.yaml` as devfile path once `--devfile` support is implemented, I can address that later.

pkg/odo/cli/component/create.go
jgwest: Yah, As Jingfu mentioned, the `filepath` package works like Java's `File` class and allows forward slash on Windows (with the API translating as needed behind the scenes.)

pkg/odo/cli/component/create.go
cdrage, line 75: `devfile.yaml` should still be a const, please move it back to a const and then use `DevfilePath` as a var?

pkg/odo/cli/component/create.go
cdrage, line 78: see other comment, `.odo/env/env.yaml` should be a const

pkg/odo/cli/component/create.go
```diff
@@ -313,6 +315,13 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			defaultComponentNamespace = client.Namespace
 		}
 
+		// Configure the context
+		if len(co.componentContext) != 0 {
+			DevfilePath = filepath.Join(co.componentContext, "devfile.yaml")
+			EnvFilePath = filepath.Join(co.componentContext, ".odo/env/env.yaml")
```
cdrage, line 321: `devfile.yaml` should be a const

same as `.odo/env/env.yaml`

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -82,6 +84,22 @@ var _ = Describe("odo devfile create command tests", func() {
 		})
 	})
 
+	Context("When executing odo create with devfile component type argument and --context flag", func() {
+		It("should successfully create the devfile componet in the context", func() {
+			newContext := path.Join(context, "newContext")
+			devfilePath := filepath.Join(newContext, "devfile.yaml")
```
cdrage, line 90: const

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -82,6 +84,22 @@ var _ = Describe("odo devfile create command tests", func() {
...
+			envFilePath := filepath.Join(newContext, ".odo/env/env.yaml")
```
cdrage, line 91: const

PR #2921, opened 2020-04-18
Enable docker devfile push test in OpenShift CI

amitkrout:
**What type of PR is this?**

/kind feature

Enable docker devfile push test in OpenShift CI

Fixes NA

docker devfile push test should pass in OpenShift CI

### DIFF
--- a/openshift-ci/build-root/Dockerfile
+++ b/openshift-ci/build-root/Dockerfile
@@ -2,6 +2,7 @@
 
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
 
-RUN yum -y install make wget gcc git httpd-tools
+RUN yum -y install make wget gcc git httpd-tools docker
+RUN sudo systemctl start docker
 
 RUN mkdir -p /tmp/secret
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -17,6 +17,7 @@ export CUSTOM_HOMEDIR=$ARTIFACTS_DIR
 # Integration tests
 make test-integration
 make test-integration-devfile
+make test-cmd-docker-devfile-push
 make test-cmd-login-logout
 make test-cmd-project
 make test-operator-hub


amitkrout:
https://github.com/openshift/odo/pull/2806#issuecomment-609381442 cc_  @johnmcollier

girishramnani:
@amitkrout there is a legitimate failures on openshift CI. can you please debug this?
cc @maysunfaisal just in case you need help

johnmcollier:
The tests are failing because there isn't a Docker socket available on the OpenShift CI:
```

[odo]  ✗  Failed to start component with name nflpne.
[odo] Error: Failed to create the component: unable to create or update component: Unable to create project source volume for component nflpne: error creating docker volume: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Removing locally running Docker images with label component=nflpneListing locally running Docker images with label component=nflpneRunning docker with args [docker ps -q --filter label=component=nflpne]
```

In order to run the Docker tests there, we'd need to find a way to set up a remote docker daemon (maybe running in another container?) and get odo/docker to talk to it instead of `/var/run/docker.sock`

But I'm wondering, do we need the Docker tests to run on the OpenShift CI? Since they won't be interacting with a Kube environment, we shouldn't need coverage on different OCPs. The Travis CI should be sufficient for test coverage for odo with Docker.

kadel:
> Enable docker devfile push test in OpenShift CI

Can we actually do this? Test on Prow run inside Pod, I'm not sure if you will be able to access docker daemon inside container on OpenShift

kadel:
> But I'm wondering, do we need the Docker tests to run on the OpenShift CI? Since they won't be interacting with a Kube environment, we shouldn't need coverage on different OCPs. The Travis CI should be sufficient for test coverage for odo with Docker.

+1

amitkrout:
@johnmcollier @kadel I just opend the pr just to have more discussoin on it 

> The tests are failing because there isn't a Docker socket available on the OpenShift CI:
[odo]  ✗  Failed to start component with name nflpne.
[odo] Error: Failed to create the component: unable to create or update component: Unable to create project source volume for component nflpne: error creating docker volume: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Removing locally running Docker images with label component=nflpneListing locally running Docker images with label component=nflpneRunning docker with args [docker ps -q --filter label=component=nflpne]

@johnmcollier The error is coming because of docker is not started in the test container. Here is the log details - https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_odo/2921/pull-ci-openshift-odo-master-v4.2-integration-e2e-benchmark/2026/artifacts/build-logs/root.log. 

Ideally the docked demon should start as per the change i made https://github.com/openshift/odo/pull/2921/files#diff-01f4ec918b634e024905add2f91a58abR5-R6.   

> But I'm wondering, do we need the Docker tests to run on the OpenShift CI? Since they won't be interacting with a Kube environment, we shouldn't need coverage on different OCPs. The Travis CI should be sufficient for test coverage for odo with Docker.

I miss interpret the whole test scenario, actually the line https://github.com/openshift/odo/blob/master/.travis.yml#L150 force me to test it against 4.* too. Now understood that for docker push target, no OpenShift cluster is needed. Thanks @johnmcollier.

So instead of oc version i think we should add coverage for docker version. I am creating a separate issue https://github.com/openshift/odo/issues/2995 for the same.

> Can we actually do this? Test on Prow run inside Pod, I'm not sure if you will be able to access docker daemon inside container on OpenShift

@kadel I think there should not be any issues as we are triggering the test inside a [container](https://github.com/openshift/odo/blob/master/openshift-ci/build-root/Dockerfile). We just need to setup a docker environment inside that container. Anyway it is will add no sense as per @johnmcollier statement.

Issue #2619, opened 2020-02-17
JSON output for operators available in the cluster


dharmit:
## User Story
As a consumer of odo (for example: Adapters), I want to get `json` output for list of OperatorHub services that would help us present the information to end-users (for example: those using IDE plugins)

## Acceptance Criteria 
- [ ] `odo catalog list services -o json` should list operators from OperatorHub along with the services that it currently lists from Service Catalog 

## Links 
- Feature Request: #2461 
- Related issue: 

/kind user-story

girishramnani:
we have all the methods to output operators as part of work done in this PR https://github.com/openshift/odo/pull/2593/files#diff-5e27f5585eee6d00b3adcddf59b96261R67-R75

so in terms of implementation I have to -
- [ ] build the json output of the csvs
- [ ] handle `EXPERIMENTAL_MODE` in json output

So the struct of `ClusterServiceVersionList` looks like this
```
type ClusterServiceVersionList struct {
	metav1.TypeMeta `json:",inline"`
	metav1.ListMeta `json:"metadata"`

	Items []ClusterServiceVersion `json:"items"`
}
```

Where each Item has a `Spec`
```
type ClusterServiceVersionSpec struct {
	InstallStrategy           NamedInstallStrategy      `json:"install"`
	Version                   version.OperatorVersion   `json:"version,omitempty"`
	Maturity                  string                    `json:"maturity,omitempty"`
	CustomResourceDefinitions CustomResourceDefinitions `json:"customresourcedefinitions,omitempty"`
	APIServiceDefinitions     APIServiceDefinitions     `json:"apiservicedefinitions,omitempty"`
	NativeAPIs                []metav1.GroupVersionKind `json:"nativeAPIs,omitempty"`
	MinKubeVersion            string                    `json:"minKubeVersion,omitempty"`
	DisplayName               string                    `json:"displayName"`
	Description               string                    `json:"description,omitempty"`
	Keywords                  []string                  `json:"keywords,omitempty"`
	Maintainers               []Maintainer              `json:"maintainers,omitempty"`
	Provider                  AppLink                   `json:"provider,omitempty"`
	Links                     []AppLink                 `json:"links,omitempty"`
	Icon                      []Icon                    `json:"icon,omitempty"`

	// InstallModes specify supported installation types
	// +optional
	InstallModes []InstallMode `json:"installModes,omitempty"`

	// The name of a CSV this one replaces. Should match the `metadata.Name` field of the old CSV.
	// +optional
[...]

PR #2926, opened 2020-04-20
add json output for listing operator backed services

girishramnani:
**What type of PR is this?**

/kind feature

> /kind documentation
> [skip ci]

add json output for operator backed services.

Fixes https://github.com/openshift/odo/issues/2619

Run CRC, install some operators then 

```
odo catalog list services -o json
```

### DIFF
--- a/pkg/machineoutput/services.go
+++ b/pkg/machineoutput/services.go
@@ -1,0 +1,25 @@
+package machineoutput
+
+import (
+	"github.com/openshift/odo/pkg/catalog"
+	olm "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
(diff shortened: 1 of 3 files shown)


girishramnani:
Now I am not sure if we need test for this because, the only thing we would be testing is assigning values to the struct, rest all has already been tested (by human readable output pieces)

dharmit:
Should we not add some integration tests to this?

girishramnani:
I am not sure if we need integration tests for this, technically no code/scenerio would be tested other then a struct and json marshaller

cdrage:
> Should we not add some integration tests to this?

I agree that we should still add tests :+1: 

/hold

girishramnani:
added tests
/hold cancel

PR #2940, opened 2020-04-20
Dynamic registry support

### LINKED ISSUE #2635: Support dynamic devfile registry
## User Story
As a user I want to use devfile from different registries to build and deploy devfile component so that odo should have a mechanism to let user specify the registry for registry flexibility.

## Acceptance Criteria 
- [ ] By default we automatically detect the registry on user's cluster to check if user has the private registry hosed on the cluster
- [ ] If user's cluster doesn't host private registry we can let user specify the registry. The registry that user specify can be:
- Public registry hosted by Che or somewhere else
- Private registry hosted by user (on user's cluster or GitHub) 

Design documentation: https://docs.google.com/document/d/10a7xm5ynJa7asrY63xxjICmCwCRCUXTdO31bGb57-1E/edit?usp=sharing

## Links 
- Related issue: 
https://github.com/openshift/odo/issues/2608
https://github.com/openshift/odo/issues/2557

/kind user-story
/area devfile

GeekArthur:

**What type of PR is this?**
/kind feature

Currently we only support static registry, which means user only can create component with devfile that is hosted on our two predefined registries:
Che devfile registry: https://github.com/eclipse/che-devfile-registry/tree/master/devfiles
Private devfile registry: https://github.com/elsony/devfile-registry/tree/master/devfiles  

This PR is to implement dynamic registry support, with this PR users can easily configure their own registries with the following operations:
- Add registry to the registry list
- List registry in the registry list
- Delete registry from the registry list
- Update registry in the registry list
- List component with corresponding registry
- Create component that is hosted by specific registry

Fixes #2635 

```
# Add registry to the registry list
odo registry add <registry name> <registry URL>

# List registry in the registry list
odo registry list

# Delete registry from the registry list
odo registry delete <registry name>

# Update registry in the registry list
odo registry update <registry name> <registry URL>

# List component with corresponding registry
odo catalog list components

# Create component that is hosted by specific registry
odo create <component type> --registry <registry name>
```

Special notes: 
- For testing purpose, you can use my personal registry https://raw.githubusercontent.com/GeekArthur/registry/master (eg. `odo registry add MyRegistry https://raw.githubusercontent.com/GeekArthur/registry/master`)


GeekArthur:
I will add unit tests and integration tests in my following commit.

GeekArthur:
> We also need to handle gracefully if the user already have a preference.yaml file but have no registry entry, i.e. a preference file created in previous version of odo. In that case, we can just default back to the default list of repos (instead of empty); otherwise, the user will need to workaround the case to manually add the default list of repos before they can get `odo catalog list components` to work again after they update to the latest version of odo.

Yup, my latest commit will handle the following two cases regarding new odo version migration and add default registries for user:
- User doesn't have `preference.yaml` file
- User has `preference.yaml` file but doesn't use dynamic registry feature before

groeges:
If we are adding the capability to add registries, should we still be providing a default that points to a personal git repository (ie Elson's)?

I can see how having this temporarily to provide the newer registry index.json might have been a good thing but with the new functionality I don't think having a personal git repository listed in the code is OK.

GeekArthur:
> If we are adding the capability to add registries, should we still be providing a default that points to a personal git repository (ie Elson's)?
> 
> I can see how having this temporarily to provide the newer registry index.json might have been a good thing but with the new functionality I don't think having a personal git repository listed in the code is OK.

The original idea regarding default registry was to let it host on personal git repository for tech preview as it's easier to manage and manipulate for both testing and development. Given we have plan for support devfile 2.0, so one option is we can do the default registry migration with devfile 2.0 support together. I can bring this up in our status meeting and discuss: 
- Where we should host default registry publicly
- Do we need to create corresponding account or we can use the existing one
- How long it takes to do the default registry migration

GeekArthur:
@groeges I open an issue to track the default registry migration: https://github.com/openshift/odo/issues/2961

pkg/catalog/catalog.go
```diff
@@ -8,17 +8,38 @@ import (
...
 	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// DevfileRegistries contains the links of all devfile registries
-var DevfileRegistries = []string{
-	"https://raw.githubusercontent.com/elsony/devfile-registry/master",
-	"https://che-devfile-registry.openshift.io/",
+// GetDevfileRegistries gets devfile registries from preference file
+func GetDevfileRegistries(registryNameList []string) (map[string]string, error) {
```
mik-dass, line 21: please add a comment why `registryNameList` is used in the function

pkg/catalog/catalog.go
```diff
@@ -8,17 +8,38 @@ import (
...
+	cfg, err := preference.New()
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.OdoSettings.RegistryList != nil {
+		for _, registry := range *cfg.OdoSettings.RegistryList {
+			if len(registryNameList) != 0 && registryNameList[0] == registry.Name {
+				devfileRegistries[registry.Name] = registry.URL
+				break
```
mik-dass, line 33: please add a comment why are we breaking here instead returning all the values

pkg/odo/cli/registry/remove.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return err
+	}
+
+	log.Info("Successfully removed registry")
+	return nil
+}
+
+// NewCmdRemove implements the "odo registry remove" command
+func NewCmdRemove(name, fullName string) *cobra.Command {
```
mik-dass, line 72: this command should be called `delete` instead of `remove` to be consistent with other odo commands

mik-dass:
@GeekArthur @kadel Should we display the registries on `odo preference view` also? Maybe just the name of the registry?

pkg/odo/cli/registry/remove.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Run contains the logic for "odo registry remove" command
+func (o *RemoveOptions) Run() (err error) {
+
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Wrapf(err, "Unable to remove registry")
+	}
+
+	err = cfg.RegistryHandler(o.operation, o.registryName, o.registryURL)
```
mik-dass, line 62: maybe ask for the user's confirmation  before deletion like we do for other delete commands
https://github.com/openshift/odo/blob/master/pkg/odo/cli/url/delete.go#L111

pkg/odo/cli/registry/update.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Run contains the logic for "odo registry update" command
+func (o *UpdateOptions) Run() (err error) {
+
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Wrapf(err, "Unable to update registry")
+	}
+
+	err = cfg.RegistryHandler(o.operation, o.registryName, o.registryURL)
```
mik-dass, line 62: maybe ask for the user's confirmation like we do for other command commands
https://github.com/openshift/odo/blob/master/pkg/odo/cli/url/delete.go#L111

pkg/catalog/catalog.go
GeekArthur: I update the logic in new commit and add corresponding comments

pkg/catalog/catalog.go
GeekArthur: I update the logic in new commit and add corresponding comments

pkg/odo/cli/registry/remove.go
GeekArthur: Done

GeekArthur:
> @GeekArthur @kadel Should we display the registries on `odo preference view` also? Maybe just the name of the registry?

Given we already displayed the full registry information (registry name + registry URL) by running `odo registry list`, I don't think it's necessary to display partial registry information (registry name) in `odo preference view`. From user perspective, it's also good for user to use one command set (odo registry) to handle all registry related operations to reduce user's confusion.

pkg/odo/cli/registry/remove.go
GeekArthur: Done

pkg/odo/cli/registry/update.go
GeekArthur: Done

GeekArthur:
@mik-dass Thanks for reviewing the PR! Comments addressed, please review again when you get a chance.

pkg/odo/cli/registry/update.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/openshift/odo/pkg/odo/cli/ui"
+	"github.com/openshift/odo/pkg/odo/genericclioptions"
+	"github.com/openshift/odo/pkg/preference"
+)
+
+const updateCommandName = "update"
+
+// "odo registry update" command description and examples
+var (
+	updateLongDesc = ktemplates.LongDesc(`Update devfile registry`)
```
dharmit, line 23: We should clarify here that it helps to update only the URL of the devfile registry. It can't help change the name of the registry. Or does it?

pkg/odo/cli/registry/update.go
```diff
@@ -0,0 +1,97 @@
...
+func (o *UpdateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.operation = "update"
+	o.registryName = args[0]
+	o.registryURL = args[1]
+	return
+}
+
+// Validate validates the UpdateOptions based on completed values
+func (o *UpdateOptions) Validate() (err error) {
+	return
```
dharmit, line 53: Again, IMO, there should be URL validation here like I've requested for `add.go`.

pkg/preference/preference.go
```diff
@@ -117,6 +117,15 @@ type OdoSettings struct {
 
 	// PushTarget for telling odo which platform to push to (either kube or docker)
 	PushTarget *string `yaml:"PushTarget,omitempty"`
+
+	// RegistryList for telling odo to connect to all the registries in the registry list
+	RegistryList *[]Registry `yaml:"RegistryList,omitempty"`
+}
+
+// Registry includes the registry metadata
+type Registry struct {
```
dharmit, line 126: Thinking out loud here. Does it add value or make sense to name this as `DevfileRegistry`? Just in case we decide to support registry of something else in the future? Or if devfile v2 might have a registry that might need a different struct of itself? WDYT @kadel @elsony

dharmit: If we do so, it will percolate to various other places in the code like `RegistryList`.

pkg/odo/cli/registry/update.go
dharmit: IMO, we should not ask for a confirmation. The validation step should be able to tell the user that the registry they've specified doesn't exist:
```
$ odo registry update myreg22 google.com
? Are you sure you want to update registry myreg22 Yes
 ✗  Failed to update registry: registry myreg22 doesn't exist
```

pkg/odo/cli/registry/add.go
```diff
@@ -0,0 +1,86 @@
...
+func (o *AddOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.operation = "add"
+	o.registryName = args[0]
+	o.registryURL = args[1]
+	return
+}
+
+// Validate validates the AddOptions based on completed values
+func (o *AddOptions) Validate() (err error) {
+	return
```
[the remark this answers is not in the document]
dharmit: Should we also have a validation that the user's trying to add a valid registry?
```sh
$ odo registry list
NAME       URL
myreg2     example.com

$ odo catalog list components
 ⚠  Registry myreg2 is not set up properly with error: Unable to download the devfile index.json from example.com/devfiles/index.json: Get example.com/devfiles/index.json: unsupported protocol scheme ""
 ✗  no deployable components found
```

pkg/odo/cli/registry/update.go
dharmit: As pointed out for `odo registry add`, should we check that the registry has a valid `index.json` file so that user doesn't erroneously enter an invalid devfile registry?

pkg/odo/cli/registry/remove.go
mik-dass: As https://github.com/openshift/odo/pull/2940#pullrequestreview-402369076 pointed out, we should update all the references of remove to delete. Even the file should be renamed to delete to be consistent with other odo commands.

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
kadel: > Should we also add validation to ensure that user is not mistakenly adding a registry with different name but same URL. For example, I could manage to do something like below:
> 
> ```shell
> $ odo registry list
> NAME                       URL
> CheDevfileRegistry         https://che-devfile-registry.openshift.io/
> DefaultDevfileRegistry     https://raw.githubusercontent.com/elsony/devfile-registry/master
> myreg                      https://che-devfile-registry.openshift.io/
> myreg2                     https://che-devfile-registry.openshift.io/
> ```

This could be a perfectly valid use-case. It should be possible to have the same registry URL under different names

pkg/odo/cli/registry/update.go
kadel: > IMO, we should not ask for a confirmation. The validation step should be able to tell the user that the registry they've specified doesn't exist:
> 
> ```
> $ odo registry update myreg22 google.com
> ? Are you sure you want to update registry myreg22 Yes
>  ✗  Failed to update registry: registry myreg22 doesn't exist
> ```

+1 This validation needs to happen first before we ask for confirmation

kadel: > As pointed out for `odo registry add`, should we check that the registry has a valid `index.json` file so that user doesn't erroneously enter an invalid devfile registry?

Early validation like this has also its downsides. It means that if there is a  temporary network error, or for example, you want a registry that is inside the corporate network and you need VPN to access it you won't be able to do it unless it is accessible in a time of running `odo registry add`. Not saying that this is a necessary bad thing, but just something to consider

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
kadel: > Should we also have a validation that the user's trying to add a valid registry?
> 
> ```shell
> $ odo registry list
> NAME       URL
> myreg2     example.com
> 
> $ odo catalog list components
>  ⚠  Registry myreg2 is not set up properly with error: Unable to download the devfile index.json from example.com/devfiles/index.json: Get example.com/devfiles/index.json: unsupported protocol scheme ""
>  ✗  no deployable components found
> ```

Early validation in `odo registry add` has also its downsides. It means that if there is a temporary network error, or for example, you want a registry that is inside the corporate network and you need VPN to access it you won't be able to do it unless it is accessible in a time of running odo registry add. Not saying that this is a necessarily bad thing, but just something to consider

`⚠  Registry myregistry is not set up properly... ` is currently a warning that doesn't make the whole common fail. This is a good thing 👍 We don't want to make odo fail when there is a connection issue with just one registry that might not even be used in the command.


```
▶ odo registry add myregistry fff
New registry was successfully added

▶ odo registry list
NAME                       URL
CheDevfileRegistry         https://che-devfile-registry.openshift.io/
myregistry                 fff

▶ odo catalog list components
 ⚠  Registry myregistry is not set up properly with error: Unable to download the devfile index.json from fff/devfiles/index.json: Get fff/devfiles/index.json: unsupported protocol scheme ""
Odo OpenShift Components:
NAME        PROJECT       TAGS                                        SUPPORTED
nodejs      openshift     10,8,8-RHOAR,latest                         YES
dotnet      openshift     2.0,latest                                  NO
httpd       openshift     2.4,latest                                  NO
nginx       openshift     1.10,1.12,1.8,latest                        NO
nodejs      openshift     6                                           NO
perl        openshift     5.24,5.26,latest                            NO
php         openshift     7.0,7.1,latest                              NO
python      openshift     2.7,3.5,3.6,latest                          NO
ruby        openshift     2.3,2.4,2.5,latest                          NO
wildfly     openshift     10.0,10.1,11.0,12.0,13.0,8.1,9.0,latest     NO

Odo Devfile Components:
NAME                 DESCRIPTION                           REGISTRY                   SUPPORTED
maven                Upstream Maven and OpenJDK 11         DefaultDevfileRegistry     YES
nodejs               Stack with NodeJS 10                  DefaultDevfileRegistry     YES
openLiberty          Open Liberty microservice in Java     DefaultDevfileRegistry     YES
java-spring-boot     Spring Boot® using Java               DefaultDevfileRegistry     Y
```

pkg/odo/cli/registry/update.go
dharmit: I agree with the downside about network glitch. But what benefit does a user have by simply adding a registry behind the VPN if they're not going to be able to use it without being connected to VPN?

I feel it's worth adding validation here and in the `add` command to avoid unnecessarily failures later. Plus, the tools built on top of odo would able to just use it and hence the messaging remains the same across all the tools (IDE plugins) because it comes from odo!

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
GeekArthur: @dharmit We do have validation logic to validate user specify registry, basically we validate the following things:
- Whether the registry URL itself is reachable or not
- Whether the index file format is correct or not as the registry URL points to the index file
- Whether the devfile format is correct or not as the index file points to devfile

But we don't do the early validation as Tomas mentions above, we validate when users **use** those registries (eg. odo create, odo catalog), it would efficient we can do the registry validation with getting values from registry together so that we can save one generic "iteration" from time complexity perspective.

pkg/odo/cli/registry/update.go
GeekArthur: I can update the message to make it more clear.

pkg/odo/cli/registry/update.go
GeekArthur: Regarding validation: https://github.com/openshift/odo/pull/2940#discussion_r417400387

GeekArthur: I would say we absolutely need the validation, but we need to consider where to add the validation, IMO that depends on the complexity of the validation itself and how long the validation takes. Given the registry validation is complicated and take pretty long time + we still goes over the same "iteration" when we use the registry + network glitch Tomas mentions, it would be better we add the validation on using the registry instead of configuring registry, it won't affect user perspective as finally users should use the registry and we gate there.

pkg/preference/preference.go
GeekArthur: Registry is the fundamental struct for any kind of registry as it only includes `Name` and `URL` fields, for future support we can use "inheritance", I know Go doesn't support "inheritance" but we still can use "composition" to achieve that.

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,44 @@
...
+	registryAddCmd := NewCmdAdd(addCommandName, util.GetFullName(fullName, addCommandName))
+	registryListCmd := NewCmdList(listCommandName, util.GetFullName(fullName, listCommandName))
+	registryUpdateCmd := NewCmdUpdate(updateCommandName, util.GetFullName(fullName, updateCommandName))
+	registryRemoveCmd := NewCmdRemove(removeCommandName, util.GetFullName(fullName, removeCommandName))
+
+	registryCmd := &cobra.Command{
+		Use:   name,
+		Short: registryDesc,
+		Long:  registryDesc,
+		Example: fmt.Sprintf("%s\n\n%s\n\n%s\n\n%s",
```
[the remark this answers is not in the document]
GeekArthur: Yup, I can add that.

pkg/odo/cli/registry/update.go
kadel: We can add some early validation like checking if the URL format actually makes sense to avoid errors like @dharmit mentioned in https://github.com/openshift/odo/pull/2940#discussion_r417075309
Later we assume that the registry URL is http/https anyway, this check would make sense here. But I would avoid any additional checks that require a network, like if URL is actually pointing to the valid registry for reasons I described above. 

I would model this after Linux package managers. They allow you to add repo even if it is not accessible (you can do `yum-config-manager --add-repo` or `add-apt-repository` with URL that is not accessible, they just validate url format) but later they show unreachable repo warnings when the repo is not accessible (during `yum update` `apt update`).

GeekArthur:
> @GeekArthur Besides the requests in the review, could we please rename ` pkg/odo/cli/registry/remove.go` to ` pkg/odo/cli/registry/delete.go` so that it's in sync with other parts of the code? Could we also do a `sed s/Remove/Delete/g` (just giving an example, not requesting to run this exact command) in that file so that resulting `delete.go` doesn't have something like `RemoveOptions`, `removeExample`, etc. but has `DeleteOptions`, `deleteExample` so on and so forth?

Yes, I can fix that, regarding development process, it would great next time we can point the CLI design change out before we finalize the design in the issue https://github.com/openshift/odo/issues/2635 (eg. `remove` design is there for a long time) as it can make the development process more efficient.

pkg/odo/cli/registry/update.go
GeekArthur: @kadel That's a good point! I think separate the validation is best option for us, we can both validate format error on registry configuring and validate registry error on registry usage, and I think it benefits for use case and product performance.

GeekArthur: > IMO, we should not ask for a confirmation. The validation step should be able to tell the user that the registry they've specified doesn't exist:
> 
> ```
> $ odo registry update myreg22 google.com
> ? Are you sure you want to update registry myreg22 Yes
>  ✗  Failed to update registry: registry myreg22 doesn't exist
> ```

Good catch! The confirmation should happens after the validation.

GeekArthur:
@kadel @dharmit FYI, I am gonna use this regex `^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$` for registry URL format validation.

GeekArthur:
@kadel @dharmit Comments addressed, please review again, thanks!

kadel:
> @kadel @dharmit FYI, I am gonna use this regex `^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$` for registry URL format validation.

oh, isn't there something already in Go std library? Maybe just trying to parse it using `url.Parse` and then, for example, checking if it has supported `Schema` and `Host` is not empty?

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
dharmit: @kadel 
> `⚠ Registry myregistry is not set up properly... ` is currently a warning that doesn't make the whole common fail. This is a good thing +1 We don't want to make odo fail when there is a connection issue with just one registry that might not even be used in the command.

@GeekArthur 
> But we don't do the early validation as Tomas mentions above, we validate when users use those registries (eg. odo create, odo catalog), it would efficient we can do the registry validation with getting values from registry together so that we can save one generic "iteration" from time complexity perspective.

Makes sense to me. :+1:

pkg/odo/cli/registry/update.go
dharmit: > @kadel That's a good point! I think separate the validation is best option for us, we can both validate format error on registry configuring and validate registry error on registry usage, and I think it benefits for both use case and product performance.

+1. Separate validations makes lot of sense. :+1:

pkg/preference/preference.go
dharmit: Yep. Agree. :smile:

GeekArthur:
> > @kadel @dharmit FYI, I am gonna use this regex `^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$` for registry URL format validation.
> 
> oh, isn't there something already in Go std library? Maybe just trying to parse it using `url.Parse` and then, for example, checking if it has supported `Schema` and `Host` is not empty?

Good to know built-in library support that, made the changes in my latest commit.

GeekArthur:
@kadel @dharmit All tests are green, please review/approve the PR again, thanks

GeekArthur:
@kadel @dharmit PR has been rebased, can you please add `lgtm` if it looks good to you? Thanks

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if c.OdoSettings.Experimental != nil && *c.OdoSettings.Experimental {
+		if c.OdoSettings.RegistryList == nil {
+			// Handle user has preference file but doesn't use dynamic registry before
+			defaultRegistryList := []Registry{
+				{
+					Name: "CheDevfileRegistry",
```
cdrage, line 201: const

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+					URL:  "https://che-devfile-registry.openshift.io/",
+				},
+				{
+					Name: "DefaultDevfileRegistry",
+					URL:  "https://raw.githubusercontent.com/elsony/devfile-registry/master",
```
cdrage, line 206: const

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+	if c.OdoSettings.Experimental != nil && *c.OdoSettings.Experimental {
+		if c.OdoSettings.RegistryList == nil {
+			// Handle user has preference file but doesn't use dynamic registry before
+			defaultRegistryList := []Registry{
+				{
+					Name: "CheDevfileRegistry",
+					URL:  "https://che-devfile-registry.openshift.io/",
+				},
+				{
+					Name: "DefaultDevfileRegistry",
```
cdrage, line 205: const

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
 		return nil, err
 	}
+
+	if c.OdoSettings.Experimental != nil && *c.OdoSettings.Experimental {
+		if c.OdoSettings.RegistryList == nil {
+			// Handle user has preference file but doesn't use dynamic registry before
+			defaultRegistryList := []Registry{
+				{
+					Name: "CheDevfileRegistry",
+					URL:  "https://che-devfile-registry.openshift.io/",
```
cdrage, line 202: const

pkg/preference/preference.go
cdrage: All below needs to be a const. Maybe it could be a const of []Registry and then here with `defaultRegistryList` you can append to it?

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	c.OdoSettings.RegistryList = &registryList
+	err = util.WriteToFile(&c.Preference, c.Filename)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to write the configuration of %s operation to preference file", operation)
```
cdrage, line 253: remove capitalization

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+			Name: registryName,
+			URL:  registryURL,
+		}
+		registryList = append(registryList, registry)
+
+	case "update":
+		return nil, errors.Errorf("Failed to update registry: registry %s doesn't exist", registryName)
+
+	case "delete":
+		return nil, errors.Errorf("Failed to delete registry: registry %s doesn't exist", registryName)
```
cdrage, line 273: remove capitalization

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+
+	case "add":
+		registry := Registry{
+			Name: registryName,
+			URL:  registryURL,
+		}
+		registryList = append(registryList, registry)
+
+	case "update":
+		return nil, errors.Errorf("Failed to update registry: registry %s doesn't exist", registryName)
```
cdrage, line 270: remove capitalization

pkg/preference/preference.go
```diff
@@ -180,9 +192,124 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
...
+	}
+
+	return registryList, nil
+}
+
+func handleWithRegistryExist(index int, registryList []Registry, operation string, registryName string, registryURL string, forceFlag bool) ([]Registry, error) {
+	switch operation {
+
+	case "add":
+		return nil, errors.Errorf("Failed to add registry: registry %s already exists", registryName)
```
cdrage, line 283: remove capitalization

pkg/odo/cli/catalog/list/components.go
```diff
@@ -57,10 +57,14 @@ func (o *ListComponentsOptions) Complete(name string, cmd *cobra.Command, args [
 	}
 
 	if experimental.IsExperimentalModeEnabled() {
-		o.catalogDevfileList, err = catalog.ListDevfileComponents()
+		o.catalogDevfileList, err = catalog.ListDevfileComponents("")
```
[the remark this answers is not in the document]
GeekArthur: This argument is for specifying registry name and ideally this argument should be optional so that we list devfile components from all registries without passing in specific registry name, but Go doesn't support optional argument so here we pass empty string `""` means there is no specific registry name.

pkg/odo/cli/catalog/list/components.go
```diff
@@ -57,10 +57,14 @@ func (o *ListComponentsOptions) Complete(name string, cmd *cobra.Command, args [
...
 		if err != nil {
 			return err
 		}
+
+		if o.catalogDevfileList.DevfileRegistries == nil {
+			log.Warning("Please run 'odo registry add <registry name> <registry URL>' to add registry for listing devfile components\n")
```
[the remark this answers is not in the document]
GeekArthur: The reason is `odo catalog list components` list both devfile component and s2i component, we don't want the `error` blocks user from viewing the s2i components, so we use `warn` instead.

The use case we cover is that user is on experimental mode without specifying any registry, but user only wants to list/create s2i component, we should support that and only warns user in case user wants to list/create devfile component later.

pkg/odo/cli/component/create.go
```diff
@@ -329,10 +329,13 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			co.CommonPushOptions.componentContext = co.componentContext
 		}
 
-		catalogDevfileList, err := catalog.ListDevfileComponents()
+		catalogDevfileList, err := catalog.ListDevfileComponents(co.devfileMetadata.devfileRegistry.Name)
 		if err != nil {
 			return err
 		}
+		if catalogDevfileList.DevfileRegistries == nil {
+			log.Warning("Please run `odo registry add <registry name> <registry URL>` to add registry then create devfile components\n")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/component/create.go
```diff
@@ -437,18 +440,22 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
+		registrySpinner := log.Spinnerf("Creating devfile component from registry: %s", co.devfileMetadata.devfileRegistry.Name)
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/component/create.go
```diff
@@ -422,7 +425,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		}
 
 		// Categorize the sections
-		log.Info("Validation")
+		log.Info("Validation (devfile component)")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/component/create.go
```diff
@@ -672,7 +679,7 @@ func (co *CreateOptions) Validate() (err error) {
 		}
 	}
 
-	log.Info("Validation")
+	log.Info("Validation (s2i component)")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/registry/add.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+const addCommandName = "add"
+
+// "odo registry add" command description and examples
+var (
+	addLongDesc = ktemplates.LongDesc(`Add devfile registry`)
+
+	addExample = ktemplates.Examples(`# Add devfile registry
+	%[1]s cheregistry https://che-devfile-registry.openshift.io/
```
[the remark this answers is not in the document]
GeekArthur: Yes, that's just user-defined registry name, it's should be one word. Below is the full help page of `odo registry add`

```
odo registry add --help                                                                                                                                                   ✔  15:09:50 
Add devfile registry

Usage:
  odo registry add <registry name> <registry URL> [flags]

Examples:
  # Add devfile registry
  odo registry add cheregistry https://che-devfile-registry.openshift.io/

Flags:
  -h, --help   Help for add

Additional Flags:
```

pkg/odo/cli/registry/add.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return
+}
+
+// Run contains the logic for "odo registry add" command
+func (o *AddOptions) Run() (err error) {
+
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Wrap(err, "Unable to add registry")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/registry/add.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	err = cfg.RegistryHandler(o.operation, o.registryName, o.registryURL, o.forceFlag)
+	if err != nil {
+		return err
+	}
+
+	log.Info("New registry was successfully added")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/odo/cli/registry/delete.go
```diff
@@ -0,0 +1,86 @@
...
+// Validate validates the DeleteOptions based on completed values
+func (o *DeleteOptions) Validate() (err error) {
+	return
+}
+
+// Run contains the logic for "odo registry delete" command
+func (o *DeleteOptions) Run() (err error) {
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Wrap(err, "Unable to delete registry")
```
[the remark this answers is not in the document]
GeekArthur: Done in my latest commit.

pkg/preference/preference.go
GeekArthur: In think in Go, constants can only be scalar values which are character, string, boolean, or numeric values. If we use constant here we should define multiple different constants for `Name` and `URL` for different registries instead of assigning values to the share field directly given those values are only used in this code block, IMO converting to constant here is not good for readability and maintainability. Let me know what you think.

pkg/preference/preference.go
GeekArthur: https://github.com/openshift/odo/pull/2940/#discussion_r419687048

pkg/preference/preference.go
GeekArthur: https://github.com/openshift/odo/pull/2940/#discussion_r419687048

pkg/preference/preference.go
GeekArthur: https://github.com/openshift/odo/pull/2940/#discussion_r419687048

GeekArthur:
@cdrage Comments addressed, please review again, thanks

.travis.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       # scenario of docker devfile url testing needs only Kube config file. So the test has been
       # added here just to make sure docker devfile url command test gets a proper kube config file.
       # without creating a separate OpenShift cluster.
-      name: "devfile catalog, create, push, delete and docker devfile url command integration tests"
+      name: "devfile catalog, create, push, delete, registry and docker devfile url command integration tests"
```
johnmcollier, line 133: I think we should split off the devfile integration tests in Travis like so:
- Integration tests that require Kube (odo create, push, delete, url for Kube)
- Integration tests that require Docker (create, push, delete, url for Docker)
- Integration tests that don't require a specific runtime (catalog and registry)

I'm not sure if we should lumping in six separate integration tests scenarios into one travis job, especially since some (catalog and registry) aren't specific to this Docker job

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
johnmcollier: Could we add another example that's pointing to an index.json on github?

tests/integration/devfile/cmd_devfile_registry_test.go
```diff
@@ -0,0 +1,91 @@
+package devfile
```
johnmcollier, line 1: I think we also need to test cases where you run `odo create` from devfiles hosted in devfile registries that are added using the `registry` command. And similarly, that you can't `odo create` a component from a devfile registry that's been removed.

pkg/odo/cli/catalog/list/components.go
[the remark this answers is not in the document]
cdrage: Yes, Go does't support optional arguments for a reason. We shouldn't have to pass in `""` into the function. I'd advise that we make this into a separate function with no passing in and create another function called `FilterDevfileComponents` so you can pass in a string. That's just IMO.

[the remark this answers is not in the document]
GeekArthur: Actually they share the same function body only difference is the optional function argument, I would say creating another function would cause duplication and function inconsistency. `ioutil.TempFile()` uses the same way to handle optional argument case for default filepath with proper comment, I think we can do the same thing.

pkg/odo/cli/registry/add.go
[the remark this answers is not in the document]
GeekArthur: Yes, we can add that.

.travis.yml
GeekArthur: I think @amitkrout  already did some categorizing integration tests work:
- All k8s devfile integration tests (with and without k8s cluster) are in this job, seems like he uses `odo login/logout` to differentiate them
- All docker devfile integration tests are here: https://github.com/openshift/odo/blob/master/.travis.yml#L164 (`docker devfile url` is a special case in k8s devfile integration tests, reason is in his comments above)

GeekArthur: Also I find another job for running k8s cluster only: https://github.com/openshift/odo/blob/master/.travis.yml#L172, the one we added before is running on OpenShift cluster. I think we should open a separate issue only for categorizing tests properly, and leave comments there.

johnmcollier: Yes, we have them separated or docker and k8s, but I don't think the `registry` test suite that you're adding belongs in the docker devfile test job for travis

pkg/preference/preference.go
johnmcollier: Can we make the values in the default registry list constants instead?
So `CheDevfileRegistry`, `https://che-devfile-registry.openshift.io`, `DefaultDevfileRegistry`, `https://raw.githubusercontent.com/elsony/devfile-registry/master`. Having them be constants will make it easier to update the URLs when we need to in the future.

.travis.yml
GeekArthur: You are right, I can create a new job to separate test suites that only needs odo runtime, for now the test suites would be `odo catalog`, `odo registry` and `odo create`.

In the mean time, more comments regarding categorizing tests can be discussed under this issue: https://github.com/openshift/odo/issues/3111

pkg/preference/preference.go
GeekArthur: That's what I mention above, we have to define multiple different constants for `Name` and `URL` given only this code block needs the those default values, but I can make the change if you think those values are useful in the future.

.travis.yml
GeekArthur: But I am gonna use separate PR to cover the tests separation as we need to change both travis config file and test structure to remove stuff (eg. namespace) that force us to use cluster inside each test case. So for the time being I will leave `registry` test suite in k8s job and do all the test separation work together in another PR.

GeekArthur:
@kadel @dharmit @johnmcollier @cdrage @mik-dass All comments are addressed, currently this PR is on "dangling" state and I believe it belongs to sprint 183. I think this PR is good to go and only thing left is to pass the flaky tests. Can you guys please review and add `/lgtm` if it looks good to you?

PR #2941, opened 2020-04-21
warn --app and --all-apps coexist

jichenjc:
/kind bug
--app and --all-apps exist should report error

`odo list --all-apps --app sdaf`

There's a edge case that I can run --all-apps in a folder has .odo subfolder
so --app will be set .. thus we can't prevent run --all-apps in this folder so we have to warn
user about this action but not block it

Fixes #?

### DIFF
--- a/pkg/odo/cli/component/list.go
+++ b/pkg/odo/cli/component/list.go
@@ -64,6 +64,10 @@ func (lo *ListOptions) Validate() (err error) {
 
 	var project, app string
 
+	if len(lo.Application) != 0 && lo.allAppsFlag {
+		klog.V(4).Infof("either --app and --all-apps both provided or provided --all-apps in a folder has app, use --all-apps anyway")
+	}
+
 	if !util.CheckKubeConfigExist() {
 		project = lo.LocalConfigInfo.GetProject()
 		app = lo.LocalConfigInfo.GetApplication()


girishramnani:
This requires fixing of an integration test as well
```
Running odo with args [odo component list --all-apps]
862[odo]  ✗  Please specify the application name and project name
```

girishramnani:
@jichenjc can you please resolve the integration tests?

jichenjc:
yes, I will fix it soon~

kadel:
@jichenjc I've added `[WIP]` to indicate that this PR still needs some work. Feel free to remove it from the title once the issues are resolved.

pkg/odo/cli/component/list.go
```diff
@@ -64,6 +64,10 @@ func (lo *ListOptions) Validate() (err error) {
 
 	var project, app string
 
+	if len(lo.Application) != 0 && lo.allAppsFlag {
+		return odoutil.ThrowContextError()
```
girishramnani, line 68: I would prefer an error like "--app and --all-apps cannot be provided together" or something more UX enhancing

jichenjc:
```
[jichen@jjtest2 odo]$ dist/bin/linux-amd64/odo list --all-apps --app sdaf
 ✗  --app and --all-apps cannot be provided together
```

updated version result

PR #2966, opened 2020-04-23
Fix `odo component describe <component-name>` for not pushed components 

### LINKED ISSUE #2689: describe command failed to find component in list output
/kind bug

**Operating System:**

**Output of `odo version`:**
master compiled version 
```
[root@xyimgr77 backend]# ../odo component list
APP     NAME                  PROJECT       TYPE        SOURCE        STATE
app     ubi-frontend-bdjz     openshift     ubi         file://./     Pushed
app     ubi-backend-qmuj      openshift     ubi:1.0     file://./     Not Pushed
[root@xyimgr77 backend]# ../odo component describe
Component Name: ubi-backend-qmuj
Type: ubi:1.0
Source: file://./
[root@xyimgr77 backend]# ../odo component describe ubi-frontend-bdjz
Component Name: ubi-frontend-bdjz
Type: ubi
Source: file://./
Environment Variables:
 · DEBUG_PORT=5858
[root@xyimgr77 backend]# ../odo component describe ubi-backend-qmuj
 ✗  Component ubi-backend-qmuj does not exist in application app
```

adisky:
**What type of PR is this?**

 /kind bug

Fixes #2689 

$ `odo create nodejs component1`
`odo describe <component-name>` should work for not pushed components, provided context is given or running from odo component directory. previously `odo describe` was working but `odo describe <component-name>` was failing.
$ `odo describe component1`

### DIFF
--- a/pkg/odo/cli/component/delete.go
+++ b/pkg/odo/cli/component/delete.go
@@ -92,6 +92,10 @@ func (do *DeleteOptions) Validate() (err error) {
 	}
 	if !do.isCmpExists {
 		log.Errorf("Component %s does not exist on the cluster", do.ComponentOptions.componentName)
+		// If request is to delete non existing component without all flag, exit with exit code 1
+		if !do.componentDeleteAllFlag {
+			os.Exit(1)
+		}
 	}
 	return
 }
@@ -179,11 +183,6 @@ func (do *DeleteOptions) Run() (err error) {
 		} else {
 			return fmt.Errorf("Aborting deletion of config for component: %s", do.componentName)
 		}
-	}
-
-	// If request is to delete non existing component exit with exit code 1
-	if !do.isCmpExists && !do.componentDeleteAllFlag {
-		os.Exit(1)
 	}
 
 	return
--- a/pkg/odo/cli/component/delete.go
+++ b/pkg/odo/cli/component/delete.go
@@ -2,6 +2,7 @@ package component
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/openshift/odo/pkg/envinfo"
@@ -178,6 +179,11 @@ func (do *DeleteOptions) Run() (err error) {
 		} else {
 			return fmt.Errorf("Aborting deletion of config for component: %s", do.componentName)
 		}
+	}
+
+	// If request is to delete non existing component exit with exit code 1
+	if !do.isCmpExists && !do.componentDeleteAllFlag {
+		os.Exit(1)
 	}
 
 	return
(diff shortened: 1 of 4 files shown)


tests/integration/component.go
```diff
@@ -250,6 +250,9 @@ func componentTests(args ...string) {
 			expected, err := helper.Unindented(`{"kind": "Component","apiVersion": "odo.openshift.io/v1alpha1","metadata": {"name": "cmp-git","namespace": "` + project + `","creationTimestamp": null},"spec":{"app": "testing","type":"nodejs","source": "https://github.com/openshift/nodejs-ex","sourceType": "git","url": ["url-1"],"storage": ["storage-1"],"ports": ["8080/TCP"]},"status": {"state": "Not Pushed"}}`)
 			Expect(err).Should(BeNil())
 			Expect(cmpDescribeJSON).To(Equal(expected))
+
+			// odo should describe not pushed component if component name is given.
+			helper.CmdShouldPass("odo", append(args, "describe", "cmp-git", "--context", context)...)
```
amitkrout, line 255: Needs validation, may be one substring match would be enough.

adisky: done.

pkg/odo/cli/component/delete.go
```diff
@@ -180,6 +181,11 @@ func (do *DeleteOptions) Run() (err error) {
 		}
 	}
 
+	// If request is to delete non existing component exit with exit code 1
+	if !do.isCmpExists && !do.componentDeleteAllFlag {
+		os.Exit(1)
```
girishramnani, line 186: should we just exit? shouldn't we show an error here?

adisky: Thanks for pointing, Updated

pkg/odo/cli/component/delete.go
```diff
@@ -180,6 +181,12 @@ func (do *DeleteOptions) Run() (err error) {
 		}
 	}
 
+	// If request is to delete non existing component exit with exit code 1
+	if !do.isCmpExists && !do.componentDeleteAllFlag {
```
dharmit, line 185: Is this a part of the issue we're trying to fix? Besides that, if I try to delete the component that's not pushed I already see this error:
```sh
$ odo delete
 ✗  Component node does not exist on the cluster
```
So, I'm kind of confused as to why we're adding this here.

dharmit:
Besides PR review, I'd request that we have better git commit messages because in current sprint, we've got [three bugs that are related to `odo describe`](https://github.com/openshift/odo/projects/41?card_filter_query=describe). If we all have git commit message like `Fixes odo describe`, it's not going to be helpful, no?

pkg/odo/cli/component/delete.go
adisky: Actually I have removed this case to fix the issue, which was causing one integration test failure for `odo delete`, To fix that i added that. https://github.com/openshift/odo/blob/master/pkg/odo/genericclioptions/context.go#L532

adisky:
> Besides PR review, I'd request that we have better git commit messages because in current sprint, we've got [three bugs that are related to `odo describe`](https://github.com/openshift/odo/projects/41?card_filter_query=describe). If we all have git commit message like `Fixes odo describe`, it's not going to be helpful, no?

Agreed on that, would update the commit message.

pkg/odo/cli/component/delete.go
dharmit: The change makes sense. Thanks! :+1:

PR #2969, opened 2020-04-23
Adding check for minimum go version

mohammedzee1000:

**What type of PR is this?**
/kind feature

Instead of seeing random messages due to lower version mismatch, you will know exactly what the problem is when building

```
❯ make bin
go run scripts/check-go-version.go
Golang version 1.13.9 checked successfully to be match constraint >= 1.13
go build -ldflags="-w -X github.com/openshift/odo/pkg/version.GITCOMMIT=d4d7571ba" cmd/odo/odo.go
```
If version is incorrect, the script will exit 1 with message `Golang version 1.12.9 does not match the constraint >= 1.13, please install correct version`

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -84,3 +84,5 @@ import:
   version: v30.1.0
 - package: github.com/golang/mock
   version: v1.4.3
+- package: github.com/hashicorp/go-version
+  version: v1.2.0
--- a/Makefile
+++ b/Makefile
@@ -44,32 +44,36 @@ GINKGO_FLAGS=$(GINKGO_FLAGS_ALL) -nodes=$(TEST_EXEC_NODES)
 
 default: bin
 
+.PHONY: common
+common:
+	go run scripts/check-go-version.go
+
 .PHONY: debug
-debug:
+debug: common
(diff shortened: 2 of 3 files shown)


dharmit:
Is there an issue where this was discussed before opening a PR? It's a feature so I'm guessing a user-story or some discussion must have happened. Can you please link it to the PR?

glide.yaml
```diff
@@ -83,4 +83,6 @@ import:
 - package: github.com/google/go-github/github
   version: v30.1.0
 - package: github.com/golang/mock
-  version: v1.4.3
\ No newline at end of file
+  version: v1.4.3
+- package: github.com/hashicorp/go-version
```
mik-dass, line 87: It seems you have added a new library but it's not in the vendor directory. Thus the builds are failing.

mohammedzee1000: yes, i will add do that

girishramnani:
I would definitely prefer to have this in the docs instead of pulling a library to run a script  , so can we have an issue for this where the team discusses the possibilities and finalise @mohammedzee1000 ?
/hold

PR #2976, opened 2020-04-23
Create a non-colliding random string generator for namespace generation and port dev file push tests to use it

jgwest:
**What type of PR is this?**
/kind feature

Create a non-colliding random string generator:
- The current `CreateRandNamespace` (in `kubernetes_utils.go`) will return namespace values that collide across multiple processes (causing `[kubectl] Error from server (AlreadyExists): namespaces "xvqjwfscgf" already exists`). This fixes that problem.

See parent issue for details: https://github.com/openshift/odo/issues/3028

To run the tests that affected by this PR"
```
export KUBERNETES=true
make test-cmd-devfile-push
```


johnmcollier:
@jgwest FYI, it looks like kubectl is not installed on the OpenShift CI:
```
     Unexpected error:
      
        <*exec.Error | 0xc0001a1f60>: {
            Name: "kubectl",
            Err: {
                s: "executable file not found in $PATH",
            },
        }
        exec: "kubectl": executable file not found in $PATH
    occurred
    /go/src/github.com/openshift/odo/tests/helper/helper_run.go:27
```

You'd probably need to add it to the Dockerfile [here](https://github.com/openshift/odo/blob/master/openshift-ci/build-root/Dockerfile) if you wanted to kubectl runner to work there.

girishramnani:
can we have an issue to discuss this feature @jgwest ?

jgwest:
Sure @girishramnani, I have opened https://github.com/openshift/odo/issues/3028.

tests/helper/helper_generic.go
```diff
@@ -162,6 +183,35 @@ func WatchNonRetCmdStdOut(cmdStr string, timeout time.Duration, check func(outpu
 	}
 }
 
+var randMutex = &sync.Mutex{}
```
girishramnani, line 186: Can we not use global mutex? Consider having a struct with the mutex and the object it protects?

jgwest:
Closing as the issue has been handled by https://github.com/openshift/odo/pull/3358 by switching everything to use crypto rand.

Issue #2979, opened 2020-04-23
Reduce the number of calls to pushtarget.IsPushTargetDocker()


johnmcollier:
/kind cleanup
/area devfile
/area local

A lot of the code that is going in for the Docker pushtarget support in odo is making a number of calls to the `pushtarget.IsPushTargetDocker()` function. Each call of that function involves needing to read the `preference.yaml` from disk and retrieving the pushtarget value from it

We should try and reduce the number of unnecessary calls to that function and reduce how many times we need to access the filesystem. Rather than calling that function each time we need to check the pushtarget, maybe it makes sense to call the function once (maybe when we initialize the context for a command), and store the value in the `genericclioptions`.

A similar approach was done for storing and accessing the Kube namespace when deploying components on Kube

GeekArthur:
Regarding this optimization, one thing I can think of is we need to ensure getting the updated pushTarget value via fs call as I believe preference.yaml is the only place that stores user’s updated pushTarget value, but we can reduce the number of fs calls for the successive functions/operations within the command

PR #2981, opened 2020-04-23
Reduce the number of calls to get experimental/pushtarget from preferences

johnmcollier:

**What type of PR is this?**

/kind cleanup
/area devfile
/area local

This PR reduces the number of calls to the `preference.PushTarget` function and stores it in the context in `genericclioptions`, making the setting available to those that have it.

Fixes https://github.com/openshift/odo/issues/2979


johnmcollier:
Tests are failing because https://github.com/openshift/odo/pull/2917#discussion_r412262923 isn't in yet.

mik-dass:
> Tests are failing because #2917 isn't in yet.

@johnmcollier since #2917 is in, please rebase the PR

girishramnani:
any update on this?

johnmcollier:
Rebased and updated the PR. Also added the experimental mode setting into the context (along with pushtarget).

Note: There were some cases where we couldn't avoid looking up the experimental/pushtarget setting in the preferences file, such as when the CLI context hasn't been initialized or where it's not available, so I left those calls in.

johnmcollier:
Needs to be rebased again, sad 🙁

johnmcollier:
There's some issues with RBAC that's causing the tests to fail, I've narrowed down the cause, but a proper fix will take some re-organizing.

Gonna hold this PR til next sprint, so I can prioritize the other issues I'm working on

/hold

girishramnani:
@johnmcollier any update on this PR?

PR #2986, opened 2020-04-24
Should be using glog.V(4) not glog.V(3)

cdrage:
**What type of PR is this?**

/kind code-refactoring

No idea what happened, but some of the code switched to using
`glog.V(3)` for verbose output. We should only be using `V(0)` or
`V(4)`.

N/A

N/A


cdrage:
Closing for now so we can discuss: https://github.com/openshift/odo/issues/3007

Issue #2550, opened 2020-01-30
Event notification support for build and application status for IDE integration for devfile scenarios


elsony:
## Which functionality do you think we should add?

Add a new mechanism to provide event notification on build status and application status change to reflect the build and run status of the application:

- The build and run status will be separated as two different types of notifications since those are two independent operations.

- The main goal of providing the build and run status is to allow facilitate IDE integration with the CLI.  In a typical IDE support, they may have custom views to display individual application build and run status of a given project.  This allows the user to have a quick snapshot view on the projects that they are working on to provide a richer developer experience on application development.

- Some examples of the build status are building, build success, build failed. Some examples of application status are starting, started, stopping, stopped.

- Individual build and run events can have more detailed messages to provide more detailed notification event, e.g. "building" can provide a detailed message on "Compiling application" or "Packaging application", etc.

- Consumer of the notification events can registry to listen to the events and those events will be sent out during the build process.

## Why is this needed?
The existing odo CLI relying on the user to take a look at the output of the build in text format to find out the progress of the build and application.  This works well in a pure CLI scenario but when used in an IDE integration, it does not provide enough fine grained details to provide a richer IDE developer experience, e.g. IDE won't know when the application build is completed and application is ready for serving that it can start opening the homepage of the application to the user (if there is one) or when they can start shelling into containers if the user needs to, etc.

Checklist for the items to be implement can be found in here:
https://github.com/openshift/odo/issues/2550#issuecomment-635556448

kadel:
CLI user and also IDE plugins can already know when the build is finished as the build is triggered by `odo push` (or `odo watch`) and both commands are synchronous. They report the status and progress.
When `odo push` command is finished the build is finished.

We currently don't have a command that would report if the application that was built already started or not but that could be easily resolved implementing something like `odo component status`.

elsony:
For the push status and progress, I think it is currently feeding to the user as log style output to the user on the odo command.  If IDEs wants to provider richer integration, e.g. have some kind of custom view to show the list of running application as well as detailed status of the build, e.g. compiling application, starting container, etc., they'll need to default back to text styled output.  Providing a proper notification event system that allows tools provider to listen to individual events and display notifications in those custom views will allow them to provide better developer experience on their application.  

If we rely on the odo component status, the tools provider will need to periodically calling the command to find out status on applications. Having an event notification system will allow them to listen for changes to update the view as the app status changes.

jgwest:
# ODO build and application status notification 

[...]

PR #3003, opened 2020-04-27
Push support: Event notification support for build and application status for IDE integration for devfile scenarios

jgwest:
**What type of PR is this?**
/kind feature

Short answer: This PR enables JSON events to be emitted, on dev file push, when the `-o json` parameter is passed.

Long answer: Design proposal is [here](https://github.com/openshift/odo/blob/parser-v2/docs/proposals/event-notification.md).

**Parent issue**:
https://github.com/openshift/odo/issues/2550

```
# Set docker target OR ensure a valid Kube context is set
export ODO_EXPERIMENTAL=true

git clone https://github.com/odo-devfiles/nodejs-ex.git
cd nodejs-ex
odo create nodejs
odo push -f -o json
```

**Sample output**:
```
{"devFileCommandExecutionBegin":{"commandId":"devbuild","componentName":"runtime","commandLine":"npm install","groupKind":"build","timestamp":"1591336117.632920"}}
{"logText":{"text":"npm WARN nodejs-sample@1.0.0 No repository field.","stream":"stderr","timestamp":"1591336119.140467"}}
{"logText":{"text":"","stream":"stderr","timestamp":"1591336119.147436"}}
{"logText":{"text":"audited 50 packages in 0.892s","stream":"stdout","timestamp":"1591336119.435678"}}
{"logText":{"text":"found 0 vulnerabilities","stream":"stdout","timestamp":"1591336119.484515"}}
{"logText":{"text":"","stream":"stdout","timestamp":"1591336119.491524"}}
{"devFileCommandExecutionComplete":{"commandId":"devbuild","componentName":"runtime","commandLine":"npm install","groupKind":"build","timestamp":"1591336119.497485"}}
{"devFileCommandExecutionBegin":{"commandId":"devrun","componentName":"runtime","commandLine":"nodemon app.js","groupKind":"run","timestamp":"1591336119.505459"}}
{"logText":{"text":"devrun                                     OK","stream":"stdout","timestamp":"1591336120.799012"}}
{"logText":{"text":"debugrun                                   OK","stream":"stdout","timestamp":"1591336120.804983"}}
{"devFileCommandExecutionComplete":{"commandId":"devrun","componentName":"runtime","commandLine":"nodemon app.js","groupKind":"run","timestamp":"1591336120.810968"}}
{"devFileCommandExecutionBegin":{"commandId":"devrun","componentName":"runtime","commandLine":"nodemon app.js","groupKind":"run","timestamp":"1591336120.816952"}}
{"logText":{"text":"devrun: started","stream":"stdout","timestamp":"1591336122.115481"}}
{"devFileCommandExecutionComplete":{"commandId":"devrun","componentName":"runtime","commandLine":"nodemon app.js","groupKind":"run","timestamp":"1591336122.121467"}}
```

### DIFF
--- a/pkg/devfile/adapters/docker/component/utils.go
+++ b/pkg/devfile/adapters/docker/component/utils.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift/odo/pkg/exec"
 	"github.com/openshift/odo/pkg/lclient"
 	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/machineoutput"
 )
 
 const (
@@ -311,10 +312,14 @@ func getPortMap(context string, endpoints []versionsCommon.Endpoint, show bool) 
 // Executes all the commands from the devfile in order: init and build - which are both optional, and a compulsary run.
 // Init only runs once when the component is created.
 func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists, show bool, containers []types.Container) (err error) {
+
 	// If nothing has been passed, then the devfile is missing the required run command
 	if len(commandsMap) == 0 {
 		return errors.New(fmt.Sprint("error executing devfile commands - there should be at least 1 command"))
 	}
+
+	stdoutWriter := a.machineEventLogger.CreateContainerOutputWriter(false)
+	stderrWriter := a.machineEventLogger.CreateContainerOutputWriter(true)
 
 	// Only add runinit to the expected commands if the component doesn't already exist
 	// This would be the case when first running the container
@@ -325,7 +330,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 
 			containerID := utils.GetContainerIDForAlias(containers, command.Exec.Component)
 			compInfo := common.ComponentInfo{ContainerName: containerID}
-			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, stdoutWriter, stderrWriter, a.machineEventLogger)
 			if err != nil {
 				return err
 			}
@@ -337,7 +342,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	if ok {
 		containerID := utils.GetContainerIDForAlias(containers, command.Exec.Component)
 		compInfo := common.ComponentInfo{ContainerName: containerID}
-		err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+		err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, stdoutWriter, stderrWriter, a.machineEventLogger)
 		if err != nil {
 			return err
 		}
@@ -353,6 +358,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 		if !componentExists {
 			err = a.initRunContainerSupervisord(command.Exec.Component, containers)
 			if err != nil {
+				a.machineEventLogger.ReportError(err, machineoutput.TimestampNow())
 				return
 			}
 		}
@@ -361,10 +367,10 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 		compInfo := common.ComponentInfo{ContainerName: containerID}
 		if componentExists && !common.IsRestartRequired(command) {
 			klog.V(4).Info("restart:false, Not restarting DevRun Command")
-			err = exec.ExecuteDevfileRunActionWithoutRestart(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+			err = exec.ExecuteDevfileRunActionWithoutRestart(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, stdoutWriter, stderrWriter, a.machineEventLogger)
 			return
 		}
-		err = exec.ExecuteDevfileRunAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+		err = exec.ExecuteDevfileRunAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, stdoutWriter, stderrWriter, a.machineEventLogger)
 	}
 
 	return
@@ -379,7 +385,7 @@ func (a Adapter) initRunContainerSupervisord(component string, containers []type
 			compInfo := common.ComponentInfo{
 				ContainerName: container.ID,
 			}
-			err = exec.ExecuteCommand(&a.Client, compInfo, command, true)
+			err = exec.ExecuteCommand(&a.Client, compInfo, command, true, nil, nil)
 		}
 	}
 
--- a/pkg/devfile/adapters/docker/component/utils.go
+++ b/pkg/devfile/adapters/docker/component/utils.go
@@ -318,9 +318,6 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 		return errors.New(fmt.Sprint("error executing devfile commands - there should be at least 1 command"))
 	}
 
-	stdoutWriter := a.machineEventLogger.CreateContainerOutputWriter(false)
-	stderrWriter := a.machineEventLogger.CreateContainerOutputWriter(true)
-
 	// Only add runinit to the expected commands if the component doesn't already exist
 	// This would be the case when first running the container
 	if !componentExists {
@@ -330,7 +327,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 
 			containerID := utils.GetContainerIDForAlias(containers, command.Exec.Component)
 			compInfo := common.ComponentInfo{ContainerName: containerID}
-			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, stdoutWriter, stderrWriter, a.machineEventLogger)
+			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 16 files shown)


pkg/machineoutput/types_event_logging.go
```diff
@@ -0,0 +1,86 @@
...
+type DevFileCommandExecutionBegin struct {
+	CommandID     string `json:"commandId"`
+	ComponentName string `json:"componentName"`
+	CommandLine   string `json:"commandLine"`
+	GroupKind     string `json:"groupKind"`
+	AbstractLogEvent
+}
+
+// DevFileCommandExecutionComplete is the JSON event that is emitted when a dev file command completes execution.
+type DevFileCommandExecutionComplete struct {
```
johnmcollier, line 40: To avoid duplication could we do something like:
```
type DevFileCommandExecutionComplete struct {
    DevFileCommandExecutionBegin
    Error         string `json:"error,omitempty"`
}
```

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -359,7 +361,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 
 			containerID := utils.GetContainerIDForAlias(containers, command.Exec.Component)
 			compInfo := common.ComponentInfo{ContainerName: containerID}
-			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
 			if err != nil {
 				return err
```
johnmcollier, line 366: Should the machineEventLogger return an error here? Since the regular error output will be swallowed

pkg/machineoutput/types.go
```diff
@@ -30,6 +31,26 @@ type GenericSuccess struct {
...
+// OutputSuccessUnindented outputs a "successful" machine-readable output format in unindented json
+func OutputSuccessUnindented(machineOutput interface{}) {
+	printableOutput, err := json.Marshal(machineOutput)
+
+	unindentedMutex.Lock()
+	defer unindentedMutex.Unlock()
+
+	// If we error out... there's no way to output it (since we disable logging when using -o json)
+	if err != nil {
+		fmt.Fprintf(log.GetStderr(), "Unable to unmarshal JSON: %s\n", err.Error())
```
johnmcollier, line 48: Is the `Unable to unmarshal JSON` error message necessary? Wouldn't that be contained in the error returned from `json.Marshal`?

johnmcollier: Also, could we use `log.Errorf`?

pkg/machineoutput/types.go
```diff
@@ -30,6 +31,26 @@ type GenericSuccess struct {
...
+	} else {
+		fmt.Fprintf(log.GetStdout(), "%s\n", string(printableOutput))
```
johnmcollier, line 50: `log.Infof` instead?

pkg/machineoutput/event_logging.go
```diff
@@ -0,0 +1,258 @@
...
+	} else if entry.GetType() == TypeDevFileCommandExecutionComplete {
+		return entry.(*DevFileCommandExecutionComplete).CommandID
+	} else {
+		return ""
+	}
+
+}
+
+// FindNextEntryByType locates the next entry of a given type within a slice. Currently used for test purposes only.
+func FindNextEntryByType(initialIndex int, typeToFind MachineEventLogEntryType, entries []MachineEventLogEntry) (MachineEventLogEntry, int) {
```
johnmcollier, line 244: Since this is only used for test purposes at the moment, should it be moved into `tests/integration`?

pkg/machineoutput/event_logging.go
```diff
@@ -0,0 +1,258 @@
...
+	TypeDevFileCommandExecutionComplete MachineEventLogEntryType = 1
+	// TypeLogText is the entry type for that event.
+	TypeLogText MachineEventLogEntryType = 2
+	// TypeReportError is the entry type for that event.
+	TypeReportError MachineEventLogEntryType = 3
+)
+
+// GetCommandName returns a command if the MLE supports that field (otherwise empty string is returned).
+// Currently used for test purposes only.
+func GetCommandName(entry MachineEventLogEntry) string {
```
johnmcollier, line 231: Since this is only used for test purposes at the moment, should it be moved into tests/integration?

pkg/machineoutput/event_logging.go
```diff
@@ -0,0 +1,258 @@
...
+			Error:            errorVal.Error(),
+			AbstractLogEvent: AbstractLogEvent{Timestamp: timestamp},
+		},
+	}
+
+	OutputSuccessUnindented(json)
+}
+
+// GetEntry will return the JSON event parsed from a single line of '-o json' machine readable console output.
+func (w MachineEventWrapper) GetEntry() (MachineEventLogEntry, error) {
```
johnmcollier, line 176: Unit tests for `GetEntry`?

pkg/machineoutput/types_event_logging.go
```diff
@@ -0,0 +1,86 @@
...
+	"io"
+)
+
+// MachineEventLoggingClient is an interface which is used by consuming code to output machine-readable
+// event JSON to the console. Both no-op and non-no-op implementations of this interface exist.
+type MachineEventLoggingClient interface {
+
+	// These functions output the corresponding eponymous JSON event to the console
+
+	DevFileCommandExecutionBegin(commandID string, componentName string, commandLine string, groupKind string, timestamp string)
```
johnmcollier, line 13: Nit: Could we rename this and `DevFileCommandExecutionComplete` to something else as they share the same name as the struct that they both return?

pkg/exec/exec.go
```diff
@@ -17,14 +17,43 @@ type ExecClient interface {
 }
 
 // ExecuteCommand executes the given command in the pod's container
-func ExecuteCommand(client ExecClient, compInfo common.ComponentInfo, command []string, show bool) (err error) {
-	reader, writer := io.Pipe()
+func ExecuteCommand(client ExecClient, compInfo common.ComponentInfo, command []string, show bool, consoleOutputStdout *io.PipeWriter, consoleOutputStderr *io.PipeWriter) (err error) {
```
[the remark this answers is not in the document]
jgwest: This is definitely something to consider: of course this is definitely down to individual opinion (a 'heuristic'), but to me this doesn't yet cross that line of 'this should be merged into a struct of params', though it is definitely very close to that line. I had considered merging the stderr/stdout into a struct, since they are semantically related, but it didn't feel like enough of an improvement (6 params -> 5) to be worth doing, and likewise none of the other parameters appeared to be semantically related enough to combine.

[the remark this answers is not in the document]
johnmcollier: Makes sense.

pkg/devfile/adapters/docker/component/utils.go
jgwest: This is covered by the called `exec.ExecuteDevfile*` functions: in all cases the error will logged as part of the DevFileCommandExecutionComplete in these functions ([for example](https://github.com/jgwest/odo/blob/30058c589ec3faa80dbcb72b3bde5c58eb9a3268/pkg/exec/devfile.go#L47)). `ReportError` in the machine logger does exist for this purpose, but it would be a double-print of the same error in this case.

pkg/machineoutput/types.go
jgwest: These lines are copied directly from the OutputSuccess/OutputError functions that were already in that file, but re: `log.Errorf`, can't use it since output from all log.* functions is disabled (throughout the code) when `-o JSON` is specified. This is part of the original machine output design, where such an output would break the pretty-printed JSON output used by consuming tools.

pkg/machineoutput/event_logging.go
jgwest: Since this function is strongly coupled to the individual struct implementations, I'd prefer to locate it alongside those implementations rather than move it to a separate package: when a developer is adding new JSON log event outputs they need to update this function as well, and moving it to a different package decreases the likelihood they will do so (though since it's only used for test implementations, it will presumably be discovered during a test run).

pkg/machineoutput/event_logging.go
jgwest: This is likewise only used for test purposes (I will add a comment indicating this), and so it likely doesn't make sense to unit test it (and likewise since the implementation is so simple).

pkg/machineoutput/types_event_logging.go
jgwest: I did this intentionally, but am open to better ideas: I considered prepending `Output` to each of the method names, but I liked the API simplicity of the name of the method corresponding to the JSON event that it emits.

Was there a specific use case you were concerned about, or just a general concern about it being non-idiomatic Go? One thing I would note is that the structs are never referenced outside of the `machineoutput` package (though I need still to capitalize them to allow the JSON (un)marshalling to work), so I don't think there is a concern about confusion over which is referred to.

pkg/machineoutput/event_logging.go
johnmcollier: Yeah, a comment indicating it's just for test purposes is good for me

jgwest:
> Also, it seems that at least some errors from the CLI are getting swallowed.
Confirmed, and fixed.

maysunfaisal:
John is on PTO, and @jgwest said he had to rebase the PR. Adding lgtm back
/lgtm

jgwest:
Hey @dharmit, re: CLI doc for this feature, currently this PR includes the following line (only displayed when experimental mode is enabled):
```
-o, --o string                 Specify output format, supported format: json
```
(This message is the same one used for the other machine readable operations with the codebase.)

Were you thinking this should be changed? (I can take a look but I'm not sure I could change this message ONLY for push, without affecting the other odo commands it is used for)

I could ALSO include an example under `Examples:`, is that the kind of thing you were thinking? Something along these lines:
```
Examples:
  # Output JSON events corresponding to devfile command execution and log text
  odo push -o json
```

dharmit:
> Hey @dharmit, re: CLI doc for this feature, currently this PR includes the following line (only displayed when experimental mode is enabled):
> 
> ```
> -o, --o string                 Specify output format, supported format: json
> ```

Ah, I didn't know that this was added as a part of this PR. Thanks for adding this.

> I could ALSO include an example under `Examples:`, is that the kind of thing you were thinking? Something along these lines:
> 
> ```
> Examples:
>   # Output JSON events corresponding to devfile command execution and log text
>   odo push -o json
> ```

Yes, this is what I was actually requesting for with a bit more info. But I'm not sure what more info we could add beyond what you've already mentioned in the comment. Sample output is too long to be shown in the CLI doc. I'll let you decide that. :+1: 

I'm also adding a label for the docs so that they know that this one needs a proper mention in the odo docs. But that would be web based docs, not the CLI one that we're discussing here.

jgwest:
Thanks @dharmit, I've added the new example, here's how it looks:

#### odo push -h - Experimental mode enabled
```
Examples:
(...)

  # Push source code with custom devfile commands using --build-command and --run-command for experimental mode
  odo push --build-command="mybuild" --run-command="myrun"

  # Output JSON events corresponding to devfile command execution and log text
  odo push -o json

Flags:
(...)
```

When experimental mode is not enabled, this example is not displayed, and the output is unchanged:

#### odo push -h - Experimental mode off
```
Examples:
(...)

  # Push source code with custom devfile commands using --build-command and --run-command for experimental mode
  odo push --build-command="mybuild" --run-command="myrun"

Flags:
(...)
```

dharmit:
Thank you @jgwest!

/approve

PR #3011, opened 2020-04-27
[Don't Merge] Run test on 4 ginkgo test node

amitkrout:
**What type of PR is this?**
/kind feature

Run test on 4 test node and we should not merge this pr until all the issues it encounter has been addressed. 

Actually running test on 4 node will reduce the CI time upto 30 min. i have verified this change in time with 4 different 4.* cluster running on AWS.

Fixes NA

CI should run on 4 test node


girishramnani:
I would like to discuss this with the team, before going back and forth 2 and 4 nodes. because the issue is increase in flakes which can only be quantified with continuously running on all test on 2 nodes and then 4 nodes. So I would suggest running the nightly job with 4 nodes and checking the flakes vs 2 nodes - finalising the approach based on that.

amitkrout:
This will be the part of Nightly run as discussed in cabal.

PR #3022, opened 2020-04-27
Make odo buildable on Go 1.12

kadel:
This should make odo buildable on Go ~~1.11.~~ 1.12.

- pining  github.com/hashicorp/go-multierror to v1.0.0 (https://github.com/hashicorp/go-multierror/issues/36)
- ~~using `user.Current()` instead of `os.UserHomeDir()` (`os.UserHomeDir()` was added in 1.12)~~
- ~~using `strings.Replace(,,,-1)` instead of `strings.ReplaceAll`  (`strings.ReplaceAll` was added in 1.12)~~


mik-dass:
@amitkrout @kadel I think we should change the go version on Travis too.

kadel:
> @amitkrout @kadel I think we should change the go version on Travis too.

I would rather go other way, if possible. Changing everything to 1.13

mik-dass:
> I would rather go other way, if possible. Changing everything to 1.13

But I think it would be good to have a backwards compatibility check on the CI so that if we use some new breaking code, it would be detected early

mohammedzee1000:
> > I would rather go other way, if possible. Changing everything to 1.13
> 
> But I think it would be good to have a backwards compatibility check on the CI so that if we use some new breaking code, it would be detected early

The problem here is though that the version was not updated across the board. Whenever we move up versions, we should ensure internal build system can handle it too. I would add that as a PR review guideline infact.

We only looked at CI system and our machines which is not the whole picture. Our releases happen on internal build systems after all

kadel:
It looks like we can't currently switch to 1.13 due to limitation in the internal build system, So let's make sure that we use 1.12 everywhere

mohammedzee1000:
@kadel for some reason, there appears to be failure happening in travis ci :confused: But otherwise its gone green everywhere else

amitkrout:
> @kadel for some reason, there appears to be failure happening in travis ci 😕 But otherwise its gone green everywhere else. It seems to be failing devfile stuff

Looks like a test script failure. Looking into it

amitkrout:
> > @kadel for some reason, there appears to be failure happening in travis ci 😕 But otherwise its gone green everywhere else. It seems to be failing devfile stuff
> 
> Looks like a test script failure. Looking into it

It's purely a race condition of ```os.Getenv("KUBECONFIG")``` and ```os.Setenv("KUBECONFIG")``` in parallel test run. For time being we can make the test to run in serial and later we can fix it for the parallel test run.

kadel:
> It's purely a race condition of `os.Getenv("KUBECONFIG")` and `os.Setenv("KUBECONFIG")` in parallel test run. For time being we can make the test to run in serial and later we can fix it for the parallel test run.

lets fix it another PR. We don't have to block on this

amitkrout:
> > It's purely a race condition of `os.Getenv("KUBECONFIG")` and `os.Setenv("KUBECONFIG")` in parallel test run. For time being we can make the test to run in serial and later we can fix it for the parallel test run.
> 
> lets fix it another PR. We don't have to block on this

Created https://github.com/openshift/odo/issues/3032 to track it

mohammedzee1000:
It says it is `retesting jobs` but I don't see any jobs in progress
`retest` and `refresh` are not appearing to fix it. This is delaying the release of v1.2.0 more than necessary. The only failure is related to test script failure which we are going to fix separately
Merging manually
cc @girishramnani @kadel

PR #3030, opened 2020-04-28
use just odo instead of odo (OpenShift Do)

kadel:


kadel:
Closing in favor of https://github.com/openshift/odo/pull/3031
/close

rpms/openshift-odo.spec
```diff
@@ -26,7 +26,7 @@ Provides:       %{package_name}
 Obsoletes:      %{package_name}
 
 %description
-OpenShift Do (odo) is a fast, iterative, and straightforward CLI tool for developers who write, build, and deploy applications on OpenShift.
+odo is a fast, iterative, and straightforward CLI tool for developers who write, build, and deploy applications on OpenShift.
```
johnmcollier, line 29: What about removing the `on OpenShift` part too? I'm torn because the non-OpenShift functionality is gated behind experimental mode, so odo's primary use case is still OpenShift.

Something like `write, build, and deploy cloud-native applications`?

Issue #2752, opened 2020-03-24
devfile integration tests should run against Kubernetes cluster


kadel:
Everything related to devfiles should work with Kubernetes as well as OpenShift clusters.

Integration tests need to be executed against Kubernetes as well as OpenShift cluster

/area devfile
/kind test
/area testing
/priority high

johnmcollier:
@kadel I've been thinking about the same. As a stopgap, I was thinking about bringing up Minikube in Travis, and running the devfile tests against it.

kadel:
> @kadel I've been thinking about the same. As both a stopgap and the most free/easiest solution, I was thinking about bringing up Minikube in Travis, and running the devfile tests against it.

+1 
`minikube start --vm-driver=none` should work well on Travis.

kadel:
The tricky part will be nicely structuring our tests as we want to keep testing also on OpenShift.

We have 2 groups of tests:
- tests that need to be executed with both Kubernetes and OpenShift (everything around devfile, except when Route is involved)
- tests that can be executed only with OpenShift cluster (legacy s2i components, and devfile component when Route is created)

prietyc123:
We have raised a concern in test-platform whether we can get a kubernetes cluster using openshift CI/prow. However I have gone through some of the documents to setup the kubernetes cluster on travis CI. 
There are certain configuration w.r.t Kubernetes which needs to be taken care like writing some bash script to start minikube for example : https://github.com/bitnami/kubernetes-travis/tree/minimal-minikube-branch/scripts .
As part of running test script is concerned, tests may need some restructure w.r.t to ```oc``` and ```kubectl``` commands, it can be handled smartly through env var IMO. @mohammedzee1000 @amitkrout

girishramnani:
I would suggest to start on travis and not consider having a kubernetes cluster on openshift.

amitkrout:
@prietyc123 refer https://minikube.sigs.k8s.io/docs/start/linux/ for how to install minikube on travis (Ubuntu VM). While installing make sure that you are using ```--vm-driver=none``` . if travis does not support configuration for --vm-driver=none (However there is an open issue https://github.com/kubernetes/minikube/issues/2704) then use  ``--vm-driver=docker``` in your travis configuration script.

prietyc123:
> I would suggest to start on travis and not consider having a kubernetes cluster on openshift.

[...]

PR #2965, opened 2020-04-23
Fix parallel test run failure for devfile integration test 

prietyc123:
**What type of PR is this?**

/kind bug

**What does does this PR do / why we need it**:

It supports parallel test run on kubernetes cluster.

**Which issue(s) this PR fixes**:

Fixes #2954 

**How to test changes / Special notes to the reviewer**:
Kubernetes job should pass on travis.

PR #2913, opened 2020-04-17
devfile delete Integration test on kubernetes cluster

prietyc123:
**What type of PR is this?**

/kind test

**What does does this PR do / why we need it**:

Runs the devfile delete integration test on kubernetes cluster.

**Which issue(s) this PR fixes**:

Fixes - Part of #2752

**How to test changes / Special notes to the reviewer**:

The test should pass on travis with ```make test-cmd-devfile-delete``` command.

PR #3041, opened 2020-04-29
devfile push Integration test on kubernetes cluster

prietyc123:
What type of PR is this?

/kind test

What does does this PR do / why we need it:

Runs the devfile delete integration test on kubernetes cluster.

Which issue(s) this PR fixes:

Fixes - Part of #2752

How to test changes / Special notes to the reviewer:

```make test-cmd-devfile-push``` should run on kubernetes cluster in travis CI.

### CODE REVIEW

**Comment by prietyc123** [comment]
Blocked because of https://github.com/openshift/odo/issues/2877#issuecomment-621117106

**Comment by prietyc123** [comment]
Issue has been tracked in https://github.com/openshift/odo/issues/3096

/retest

**Comment by prietyc123** [comment]
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.2-integration-e2e/75#1:build-log.txt%3A748

/retest

**Comment by amitkrout** [comment]
I was following the changes since you started working on it. Looks good to me 👍 
/lgtm

File: scripts/oc-cluster.sh [hunk 1]
```diff
@@ -8,8 +8,8 @@ OPENSHIFT_CLIENT_BINARY_URL=${OPENSHIFT_CLIENT_BINARY_URL:-'https://github.com/o
 
 sudo service docker stop
 
-sudo sed -i -e 's/sock/sock --insecure-registry 172.30.0.0\/16/' /etc/default/docker
-sudo cat /etc/default/docker
+sudo sed -i -e 's/"mtu": 1460/"mtu": 1460, "insecure-registries": ["172.30.0.0\/16"]/' /etc/docker/daemon.json
```
mohammedzee1000, line 11: daemon.json +1

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -17,32 +17,46 @@ var _ = Describe("odo devfile push command tests", func() {
...
-		helper.Chdir(context)
-
 		os.Setenv("GLOBALODOCONFIG", filepath.Join(context, "config.yaml"))
 
 		// Devfile push requires experimental mode to be set
 		helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
+
+		if os.Getenv("KUBERNETES") == "true" {
+			homeDir := helper.GetUserHomeDir()
+			kubeConfigFile := helper.CopyKubeConfigFile(filepath.Join(homeDir, ".kube", "config"), filepath.Join(context, "config"))
```
kadel, line 41: Why is kubeconfig copied only for Kubernetes?

scripts/oc-cluster.sh
kadel: the same change is already done in https://github.com/openshift/odo/pull/2913

prietyc123: As both the pr needs that change and we don't have this on master, So I was just checking if it works. Once it goes in I will rebase the pr and we are good to go.

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: Comment https://github.com/openshift/odo/pull/2965#discussion_r413744746 elaborates the reason in a proper manner.

kadel: Yes, I understand that. But the same problem will be with `helper.CreateRandProject` that calls `odo project create` that also modifies kubeconfig. Or am I missing something?

prietyc123: Actually the issue was with running the test in parallel on kubernetes cluster. By default the project is not active when it is created through kubectl command on kubernetes cluster, a .lock file is created while set-context is used to make it active. In parallel run this .lock file causes the issue of ```[1] [kubectl] error: open (path)/k3s.yaml.lock: The file exists.``` as described in https://github.com/openshift/odo/pull/2965#pullrequestreview-399016104.

On the other hand in case of openshift cluster such thing does not happen when it creates the project and others using oc command. Yes ofcourse it modifies the kubeconfig but does not create a .lock file I guess.

amitkrout:
Rebased
/lgtm

tests/integration/devfile/cmd_devfile_push_test.go
kadel: > On the other hand in case of openshift cluster such thing does not happen when it creates the project and others using oc command. Yes ofcourse it modifies the kubeconfig but does not create a .lock file I guess. Does it make sense?

`helper.CreateRandProject` is not using `oc`, but `odo` (don't know why I think that it should use `oc`).

But this doesn't mean that the problem with the parallel run is not there with OpenShift.
If  `helper.CreateRandProject` gets called multiple times at the same time, they will try to modify the same KUBECONFIG, and you will have the same problem as with `kubectl`. And in this situation, it is even more dangerous because odo doesn't use locking to prevent race condition when editing the KUBECONFIG

prietyc123: > `helper.CreateRandProject` is not using `oc`, but `odo` (don't know why I think that it should use `oc`).

AFAIU `odo` is being used to create project because in all the test scenarios we are validating `odo project create` functionality while creating the project, so we should not care about how `oc new-project create` works in `odo` project. Ofcourse there are few functionality which we are borrowing from `oc` as there is no such options that `odo` command provides.

> But this doesn't mean that the problem with the parallel run is not there with OpenShift.
> If `helper.CreateRandProject` gets called multiple times at the same time, they will try to modify the same KUBECONFIG, and you will have the same problem as with `kubectl`. And in this situation, it is even more dangerous because odo doesn't use locking to prevent race condition when editing the KUBECONFIG

ok, however when I tested the same script on 4 test nodes I didn't see any failure from openshift side rather I could be able to see the ```.lock``` file error from kubernetes cluster side. 
I am not sure how these ```KUBECONFIG``` files are being handled during the creation of  random project against openshift and kubernetes cluster behind the scene respectively. Could you please add some more content like how it works, what call it makes etc... during project creation?
I will make the changes as per your suggestions.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -17,32 +17,46 @@ var _ = Describe("odo devfile push command tests", func() {
...
 	var projectDir = "/projectDir"
 	var sourcePath = "/projects/nodejs-web-app"
+	var clusterType helper.CliRunner
 
-	// TODO: all oc commands in all devfile related test should get replaced by kubectl
-	// TODO: to goal is not to use "oc"
-	oc := helper.NewOcRunner("oc")
+	// Using program commmand according to cluter type in devfile
+	if os.Getenv("KUBERNETES") == "true" {
+		clusterType = helper.NewKubectlRunner("kubectl")
```
girishramnani, line 24: Can we have a better variable name for this? clusterType doesn’t seem appropriate for a runner.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -17,32 +17,46 @@ var _ = Describe("odo devfile push command tests", func() {
 	var namespace, context, cmpName, currentWorkingDirectory, projectDirPath string
 	var projectDir = "/projectDir"
 	var sourcePath = "/projects/nodejs-web-app"
+	var clusterType helper.CliRunner
 
-	// TODO: all oc commands in all devfile related test should get replaced by kubectl
-	// TODO: to goal is not to use "oc"
-	oc := helper.NewOcRunner("oc")
+	// Using program commmand according to cluter type in devfile
+	if os.Getenv("KUBERNETES") == "true" {
```
girishramnani, line 23: This condition should be extracted into a helper. Something like getCliRunner

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -17,32 +17,46 @@ var _ = Describe("odo devfile push command tests", func() {
...
+		currentWorkingDirectory = helper.Getwd()
+		cmpName = helper.RandString(6)
+		helper.Chdir(context)
 	})
 
 	// Clean up after the test
 	// This is run after every Spec (It)
 	var _ = AfterEach(func() {
-		helper.DeleteProject(namespace)
+		if os.Getenv("KUBERNETES") == "true" {
```
girishramnani, line 54: This condition should also be extracted into a function.

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: > This condition should be extracted into a helper. Something like getCliRunner

This is just a env variable I am fetching it over here. It has nothing to deal with like the name you suggest `getCliRunner`.

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: > Can we have a better variable name for this? clusterType doesn’t seem appropriate for a runner.

@girishramnani I am open to a better name. Can you please suggest one?

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: https://github.com/openshift/odo/pull/3041#discussion_r421446577

tests/integration/devfile/cmd_devfile_push_test.go
kadel: > ok, however when I tested the same script on 4 test nodes I didn't see any failure from openshift side rather I could be able to see the `.lock` file error from kubernetes cluster side.

Yes, because that problem is just hidden due to the fact that there is no locking mechanism to prevent two processes to modify the same file at the same time. But that doesn't mean that it is not happening. 
If you would have two test specs running at the same time and they would be executing a  command that uses active project/namespsce from KUBECONFIG one of them would be using the wrong one.

> Could you please add some more content like how it works, what call it makes etc... during project creation?

`odo project create`? It is not doing anything special. Just creates the project (the equivalent of `oc create project <name>`) and then sets it as active `odo project <name>`.
The setting is as active is where it can collide because it modifies KUBECONFIG.
So if the same KUBECONFIG is being used and the process tries to set active project at the same time. What is going to be written to KUBECONFIG?

amitkrout: > If you would have two test specs running at the same time and they would be executing a command that uses active project/namespsce from KUBECONFIG one of them would be using the wrong one.
> odo project create? It is not doing anything special. Just creates the project (the equivalent of oc create project <name>) and then sets it as active odo project <name>. The setting is as active is where it can collide because it modifies KUBECONFIG.

@kadel hmm, but i wonder even on running test in parallel on 4 node against OpenShift cluster none of the time it throws any error related to KUBECONFIG or error like it throws against Kubernetes cluster. Is it a bug on OpenShift ? or OpenShift handles it in abetter way unlike Kubernetes ?

prietyc123: > If you would have two test specs running at the same time and they would be executing a command that uses active project/namespsce from KUBECONFIG one of them would be using the wrong one.

@kadel I did a quick experiment to confirm the weird behavior but unfortunately none of the time I am able to hit the issue of KUBECONFIG collision against OpenShift cluster on 8 test node. 
Command executed :
```make test-cmd-project``` at least 5 times.
Experimented on:
```
func CreateRandProject() string {
  projectName := RandString(10)
  fmt.Fprintf(GinkgoWriter, "Creating a new project: %s\n", projectName)
  // session := CmdShouldPass("odo", "project", "create", projectName, "-w", "-v4")
  session := CmdShouldPass("oc", "new-project", projectName)
  // Expect(session).To(ContainSubstring("New project created"))
  Expect(session).To(ContainSubstring(projectName))
  return projectName
}
var _ = BeforeEach(func() {
    SetDefaultEventuallyTimeout(10 * time.Minute)
    SetDefaultConsistentlyDuration(30 * time.Second)
    context = helper.CreateNewContext()
    project = helper.CreateRandProject()
  })
  // Clean up after the test
  // This is run after every Spec (It)
  var _ = AfterEach(func() {
    helper.DeleteProject(project)
    helper.DeleteDir(context)
  })
  Context("Machine readable output tests", func() {
    It("Help for odo project list should contain machine output", func() {
      output := helper.CmdShouldPass("odo", "project", "list", "--help")
      Expect(output).To(ContainSubstring("Specify output format, supported format: json"))
    })
    It("Help for odo project list should contain machine output", func() {
      output := helper.CmdShouldPass("odo", "project", "list", "--help")
      Expect(output).To(ContainSubstring("Specify output format, supported format: json"))
    })
    [similar spec 6 more times]
  })
```
Can we file an issue on OpenShift to confirm the behavior before going for a major changes in our repo ? cc_ @amitkrout

File: tests/integration/devfile/cmd_devfile_push_test.go (hunk 3 above)
Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go]: done

**Comment by prietyc123** [comment]
prow failure

/retest

File: tests/integration/devfile/cmd_devfile_push_test.go (hunk 2 above)
Comment (amitkrout) [reply tests/integration/devfile/cmd_devfile_push_test.go:39]: Right, Lets check if there is any issue reported on ````OKD``` or ```oc```. Otherwise i would say lets drop a mail to ```aos-devel``` to get a clarity on this.

Comment (kadel) [reply tests/integration/devfile/cmd_devfile_push_test.go:39]: > Can we file an issue on OpenShift to confirm the behavior before going for a major changes in our repo ? cc_ @amitkrout

It is not issued with OpenShift itself. I'm saying that the same problem as we have with namespace collisions in Kubernetes tests is also present in our OpenShift tests. The only reason why we don't see it is that in most of our tests we use `--project` flag.

We simply should treat Openshift and Kubernetes the same way. If we are doing something for Kubernetes it should be also done for OpenShift tests.

File: tests/integration/devfile/cmd_devfile_push_test.go (hunk 5 above)
Comment (kadel) [reply tests/integration/devfile/cmd_devfile_push_test.go:52]: Why don't we create something like  `helper.DeleteNamespaceOrProject` (not a good name, this is just for illustration) to avoid this whole `if` statement?

in `OcRunner` this function would be implemented in a way that it deletes Project.
in `KubectlRunner` this function would be implemented in a way that it deletes Namespace

The same could be don for `CreateRandProject`/`CreateRandNamespace`

File: tests/integration/devfile/cmd_devfile_push_test.go (hunk 2 above)
Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go:39]: > The only reason why we don't see it is that in most of our tests we use --project flag.

Ahh...Now i understood
It's the ```--project``` flag which is saving us from the failure. using oc command it was passing because I was using freshly created project for each spec in script i pasted above to test it with oc which @amitkrout somehow missed it.
Actually i am in the favor of test script in sync for both the cluster but i was little confused why it was even not failing with oc.
Thanks for the info. Will update it ASAP.

Comment (kadel) [reply tests/integration/devfile/cmd_devfile_push_test.go:39]: > It's the `--project` flag which is saving us from the failure. using oc command it was passing because I was using freshly created project for each spec in script i pasted above to test it with oc which @amitkrout somehow missed it.

yes, exactly.  And you never know when someone adds test that doesn't use `--project` flag. When that happens we would start getting flakes due to wrong project being used.

File: tests/integration/devfile/cmd_devfile_push_test.go (hunk 5 above)
Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go:52]: > Why don't we create something like `helper.DeleteNamespaceOrProject` (not a good name, this is just for illustration) and make it part of `CliRunner` interface to avoid this whole `if` statement?
> 
> in `OcRunner` this function would be implemented in a way that it deletes Project.
> in `KubectlRunner` this function would be implemented in a way that it deletes Namespace
> 
> The same could be done for `CreateRandProject`/`CreateRandNamespace`

I will check and figure it out.

Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go:52]: https://github.com/openshift/odo/issues/2752#issuecomment-627977945

Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go:52]: > Why don't we create something like helper.DeleteNamespaceOrProject (not a good name, this is just for illustration) and make it part of CliRunner interface to avoid this whole if statement?

@kadel Yes creating `helper.DeleteNamespaceProject` and `helper.CreateRanNamespaceProject` make sense. However I would say the tests which are only running for openshift let it use `helper.CreateRanProject` and `helper.DeleteProject` because replacing it with `helper.CreateRanNamespaceProject` won't justify the running cluster whereas we can implement `helper.DeleteNamespaceProject` and `helper.CreateRanNamespaceProject` for devfile related tests only using `CliRunner`.

IMO changing the project create/delete function name will only justify devfile related test. So lets keep it as it is for the tests only running on openshift and implement the changes for the devfile tests.

Comment (prietyc123) [reply tests/integration/devfile/cmd_devfile_push_test.go]: Addressed the requested change in 73c4a24

**Comment by prietyc123** [comment]
Prow failure https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.4-integration-e2e/397#1:build-log.txt%3A40

/retest

**Comment by prietyc123** [comment]
Seeing failure on prow : https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.2-integration-e2e/408

@amitkrout @mohammedzee1000 As per my change OpenShift CI is failing because we are triggering the test from the test container `registry.svc.ci.openshift.org/openshift/release:golang-1.13` where we do not have `KUBECONFIG`. Can you please suggest how to fix it.

File: tests/integration/devfile/cmd_devfile_catalog_test.go [hunk 6]
```diff
@@ -14,32 +14,32 @@ var _ = Describe("odo devfile catalog command tests", func() {
...
 		helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
-		if os.Getenv("KUBERNETES") == "true" {
-			homeDir := helper.GetUserHomeDir()
-			kubeConfigFile := helper.CopyKubeConfigFile(filepath.Join(homeDir, ".kube", "config"), filepath.Join(context, "config"))
-			project = helper.CreateRandNamespace(kubeConfigFile)
-		} else {
-			project = helper.CreateRandProject()
-		}
+		homeDir := helper.GetUserHomeDir()
+		helper.CopyKubeConfigFile(filepath.Join(homeDir, ".kube", "config"), filepath.Join(context, "config"))
```
kadel, line 33: You should not just assume that the kubeconfig will be in a `$HOME/.kube/config`.
Kubeconfig can be elsewhere. You first need to check if `$KUBECONFIG` if set.
If `$KUBECONFIG` is already set use that, if not use the `$HOME/.kube/config`.

kadel:
> @amitkrout @mohammedzee1000 As per my change OpenShift CI is failing because we are triggering the test from the test container `registry.svc.ci.openshift.org/openshift/release:golang-1.13` where we do not have `KUBECONFIG`. Can you please suggest how to fix it.

The problem is quite opposite. the `$KUBECONFIG` is set there - https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.2-integration-e2e/408#1:build-log.txt%3A43

adisky:
@priti You can check this function, its check for existence though but consider all  cases. https://github.com/openshift/odo/blob/99f1a2f84fef58c2aafbce9a9bc44e47f7e63fe8/pkg/util/util.go#L976

tests/integration/devfile/cmd_devfile_catalog_test.go
prietyc123: > You should not just assume that the kubeconfig will be in a $HOME/.kube/config.
Kubeconfig can be elsewhere. You first need to check if $KUBECONFIG if set.
If $KUBECONFIG is already set use that, if not use the $HOME/.kube/config.

@kadel Yes, `KUBECONFIG` can be somewhere and we need to check if $KUBECONFIG is set. But in `helper.CopyKubeConfigFile` we are again setting up `KUBECONFIG` env var with the temporary config context which mean if we use `$KUBECONFIG` as an arg to `helper.CopyKubeConfigFile` there might occur some racing condition in parallel test run.

Like, If we are running tests on two nodes and first node sets `KUBECONFIG` to some temporary config file and second node comes up and passes `$KUBECONFIG` which can pick up the the previously set file. This will again ruin everything.

IMO we should handle `$HOME/.kube/config` and `$HOME/admin.kubeconfig` according to the CI type. I mean on OpenShift CI or travis CI.

prietyc123:
@kadel @amitkrout @mohammedzee1000 After copying the `[KUBECONFIG](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.2-integration-e2e/440#1:build-log.txt%3A44)` file to temporary config I am hitting a weird [failure](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.2-integration-e2e/440#1:build-log.txt%3A507 ) on prow. Though the changes works fine on travis CI.

Not sure why is it happening on openshift CI. Can you please suggest what is the issue about and how to fix it?

tests/integration/devfile/cmd_devfile_catalog_test.go
kadel: > Like, If we are running tests on two nodes and first node sets `KUBECONFIG` to some temporary config file and second node comes up and passes `$KUBECONFIG` which can pick up the the previously set file. This will again ruin everything.

Doesn't each node use a separate environment? The second node should not be able to pick up the previously set file.

tests/helper/helper_generic.go
```diff
@@ -167,3 +168,13 @@ func GetUserHomeDir() string {
 	Expect(err).NotTo(HaveOccurred())
 	return homeDir
 }
+
+// LocalKubeconfigSet checks for the running CI and sets the KUBECONFIG to the temporary config directory respectively
+func LocalKubeconfigSet(context string) {
+	homeDir := GetUserHomeDir()
+	if os.Getenv("CI") == "openshift" {
+		copyKubeConfigFile(filepath.Join("/tmp", "admin.kubeconfig"), filepath.Join(context, "config"))
```
kadel, line 176: This doesn't look right at all. You should not use admin.kubeconfig.
Use the `$KUBECONFIG` if it is set. the same for Kubernetes

kadel:
> Not sure why is it happening on openshift CI. Can you please suggest what is the issue about and how to fix it?

it is happening because you are using wrong KUBECONFIG, you need to use the one that `$KUBECONFIG` points to

tests/helper/helper_generic.go
```diff
@@ -167,3 +168,13 @@ func GetUserHomeDir() string {
 	Expect(err).NotTo(HaveOccurred())
 	return homeDir
 }
+
+// LocalKubeconfigSet checks for the running CI and sets the KUBECONFIG to the temporary config directory respectively
+func LocalKubeconfigSet(context string) {
+	homeDir := GetUserHomeDir()
+	if os.Getenv("CI") == "openshift" {
```
kadel, line 175: can you try something  like this?
```go
	originalKubeCfg := os.Getenv("KUBECONFIG")
	if originalKubeCfg == "" {
		originalKubeCfg = filepath.Join(homeDir, ".kube", "config")
	}
	copyKubeConfigFile(originalKubeCfg, filepath.Join(context, "config"))
```

tests/integration/devfile/cmd_devfile_catalog_test.go
prietyc123: > Doesn't each node use a separate environment? The second node should not be able to pick up the previously set file.

Actually previously I tried using `$KUBECONFIG` by passing `os.Getenv("KUBECONFIG")` as an arg to `copyKubeConfigFile()` but there I found the racing condition with parallel tests run as mentioned in [comment](https://github.com/openshift/odo/pull/3041#discussion_r425198685) and it was picking up the previously set temporary file and error says `no such file exists`

AFAIU It takes the existing `$KUBECONFIG`.

tests/helper/helper_generic.go
prietyc123: > can you try something like this?
> 
> ```go
> 	originalKubeCfg := os.Getenv("KUBECONFIG")
> 	if originalKubeCfg == "" {
> 		originalKubeCfg = filepath.Join(homeDir, ".kube", "config")
> 	}
> 	copyKubeConfigFile(originalKubeCfg, filepath.Join(context, "config"))
> ```

It is already there in the function https://github.com/openshift/odo/blob/99f1a2f84fef58c2aafbce9a9bc44e47f7e63fe8/pkg/util/util.go#L976 which I am using [here](https://github.com/openshift/odo/pull/3041/files#diff-318517042700080474192fc7dc04e456R33)

tests/integration/devfile/cmd_devfile_catalog_test.go
prietyc123: Used this scenario here https://github.com/openshift/odo/commit/7eef087768f567eae9ab887e5327b5c3e472c50f#diff-d3d0e9ea582e16c9f0d4edabc01c6247R39 and got the failure `no such file exists` and then changed the approach after researching a lot.

amitkrout:
@prietyc123 Can you please revert the implementation you made for syncing the test code to run against prow. BTW test code sync will be the part of #3203. 

Let unblock the this pr and yours other devfile prs too. cc_ @kadel

prietyc123:
> Can you please revert the implementation you made for syncing the test code to run against prow.

I have reverted my changed and skipped setting `KUBECONFIG` to a temporary config file on prow due to https://github.com/openshift/odo/issues/3203

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -16,31 +16,40 @@ import (
...
+		helper.Chdir(context)
 	})
 
 	// Clean up after the test
 	// This is run after every Spec (It)
 	var _ = AfterEach(func() {
-		helper.DeleteProject(namespace)
+		cliRunner.DeleteNamespaceProject(namespace)
+		if os.Getenv("KUBERNETES") == "true" {
+			os.Unsetenv("KUBECONFIG")
```
kadel, line 51: This unset might be causing the problems that you have on OpenShift CI.
OpenShift CI sets custom KUBECONFIG, you change it in your `LocalKubeconfigSet` but then instead of restoring its original value, you just unset it.

prietyc123:
prow failure

/retest

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: Yes, This was the main reason it seems. As per our discussion I have updated the change related to `KUBECONFIG`.

Thanks for the input 👍

prietyc123:
External issue https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3041/pull-ci-openshift-odo-master-v4.4-integration-e2e/865#1:build-log.txt%3A573

/retest

tests/integration/devfile/cmd_devfile_push_test.go
girishramnani: I don’t think we are worried about over crowding the helper file 🙂

girishramnani: I see +3 instances of this condition, that’s 15 lines of code we can reduce by adding a small function, I hope that’s a good enough trade off to have to function in the helper file?

prietyc123: @girishramnani Yes, we should have it in a function, exactly same we are already doing in pr  [here](https://github.com/openshift/odo/pull/3221/files#diff-c5e2055f719fdb07a45e8abeac3b112eR207). I don't see any point adding it in this pr.

girishramnani: Think about it this way, in between merging that PR and your PR there will be a time difference where who so ever adds a new test file will use this same if condition instead of the helper function. And then there will be another PR after a while which would clean that up - if we remember that is.
So please do the change even if it’s redundant. It’s a good cleanup.

prietyc123: ok, Anyway the [pr](https://github.com/openshift/odo/pull/3221) will cleanup it again. As part of helper file over crowding is concern, i will think later how to split the file on the basis of function we use there if volume of helper increases.

prietyc123: Done

amitkrout:
https://github.com/openshift/odo/issues/3122
/test v4.3-integration-e2e

Issue #3000, opened 2020-04-27
Allow a project to be specified using the downloadSource flag


Kamran64:
/kind user-story

## User Story
As a developer, I'd like to be able to specify a specific project (by `name` field) to be downloaded from the `--downloadSource` flag on `odo create`. Currently, using the flag only downloads the first project in the devfile.

## Acceptance Criteria 
- [ ] Add an optional argument to `--downloadSource` which downloads the project specified by the argument. The argument will be the `name` field of the project definition e.g. `nodejs-web-app`
- [ ] If no argument is supplied to `--downloadSource`, default to the first project as we have in todays implementation.
- [ ] If a project name doesn't match a project in the devfile or there are multiple projects with the same name, we should fail.
- [ ] Resource creation will only be triggered if:

1. The experimental flag has been enabled
2. The component is a devfile component
3. There is a project defined in the devfile of the given component
4. odo create command called with a --downloadSource flag

## Usage
- Set odo to experimental mode

- Run an `odo create <component-name> --downloadSource` and optionally specify a project using `--downloadSource=<project-name>`.

- Resource creation will only work if there is a project listed in the devfile,

## Links 
- Related issue: https://github.com/openshift/odo/issues/2582

/kind user-story

Kamran64:
A couple of thoughts after looking at implementing this yesterday:

1. The `--downloadSource` flag is currently a bool value. If we change it to hold a string, then it becomes mandatory that a value is entered after the flag is invoked which defeats the purpose of acceptance criteria 2.

2. The other option is to keep the flag as a bool but check for a non flag argument passed after the flag (similar to how the component name is passed). However, if we do this: how do we differentiate between whether a component name is passed or project to be downloaded since both arguments are optional.
    - E.g. if a user passes `odo create nodejs --downloadSource my-component` how do we know if they're trying to specify a project after downloadSource or if they're just trying to name their component?
    - One solution could be that we make it mandatory that if a component name is being provided, then it must come before `--downloadSource` and we assume that the arg after (if any) is the project to be downloaded.

@elsony @johnmcollier - What are your thoughts on this?

Kamran64:
Other solutions considered are:
- Add a new flag to specify the project to be downloaded - However this may not be nice for the user experience (might be confusing to differentiate between `--downloadSource` and this new flag.)
- Keep `--downloadSource` as it is and if there is more than one project in the devfile, trigger interactive mode to allow the user to choose the project. This won't work in cases where project creation is automated though.

[...]

Issue #2194, opened 2019-09-27
odo Interactive mode test script implementation


amitkrout:
Add integration test for odo interactive mode usecase

mohammedzee1000:
It appears we may not need any 3rd party library as this can be done with stdlib os.exec.
Here is an example:

```bash
$ cat test.sh 
#!/usr/bin/env bash

echo "Enter something"
read TEST
echo "You entered $TEST"

$  ./test.sh 
Enter something
hai  
You entered hai

$  cat prcommunicate.go 
package main

import (
	"bufio"
	"fmt"
	"log"
	"os/exec"
)

func CreateInteractiveCommand(name string, args ...string) (*exec.Cmd, *bufio.Reader, *bufio.Writer, error) {
	c := exec.Command(name, args...)
	si, err := c.StdinPipe()
	if err != nil {
		return nil, nil, nil, err
	}
	so, err := c.StdoutPipe()
	if err != nil {
		return nil, nil, nil, err
	}
	out := bufio.NewReader(so)
	in := bufio.NewWriter(si)
	return c, out, in, nil
}

func main() {
	var outdata string
	cmd, out, in, err := CreateInteractiveCommand("./test.sh")
	if err != nil {
		log.Fatal("Error initializing command : ", err)
	}
	err = cmd.Start()
	if err != nil {
		log.Fatal("Error starting command : ", err)
	}

	//Actual test begins
	outdata, err = out.ReadString('\n')
	if err != nil {
		log.Fatal("unable to read ")
	}
	fmt.Println("Actual First output : ", outdata, "Expected FIrst Output : Enter something\\n")
	if outdata != "Enter something\n" {
		log.Fatal("Failed")
	}
	_, err = in.Write([]byte("hello\n"))
	if err != nil {fmt.Println("Actual First output : ", outdata, "Expected FIrst Output : Enter something\n")
		log.Fatal("failed to write : ")
	}
	err = in.Flush()
	if err != nil {
		log.Fatal("Failed to flush")
	}
	outdata, err = out.ReadString('\n')
	if err != nil {
		log.Fatal("unable to read ")
	}
	fmt.Println("Actual second output : ", outdata, "Expected second Output : You entered hello\\n")
	if outdata != "You entered hello\n" {
		log.Fatal("Failed")
	}
	//out.Read(outdata)
}

$ go run prcommunicate.go
Actual First output :  Enter something
 Expected FIrst Output : Enter something\n
Actual second output :  You entered hello
[...]

PR #3044, opened 2020-04-29
Allow a project to be specified in --downloadSource

Kamran64:
**What type of PR is this?**
/kind feature

This PR will enable the user to specify which project they want to download when using the `--downloadSource` flag. If no project is specified when using the flag, we download the first one in the list if there is only one, or resort to interactive mode if there are multiple.

Due to a limitation with the `NoOptDefVal` attribute in `flags` the format must be `--downloadSource=<project-name>` or just `--downloadSource`. See: https://github.com/spf13/pflag/issues/134

Fixes: https://github.com/openshift/odo/issues/3000

-  Set odo to experimental mode
- Run an `odo create nodejs --downloadSource` and optionally specify a project using `--downloadSource=<project-name>`.

- Since the currently supported devfiles only have one project, the tests don't cover the interactive mode portion (although I'm not sure how that could be tested anyway)


mik-dass:
ok-to-test

kadel:
> If no project is specified when using the flag, we download the first one in the list if there is only one, or resort to interactive mode if there are multiple.

I don't think that this is a good experience. From the begging of odo we have a philosophy that if you want to run interactive mode you do it from start, that means that you just do `odo create` to invoke full interactive mode.

When we start invoking interactive mode partially (just for some parameters and only if some conditions are met) it becomes really hard to create scripts that use odo.

I would prefer odo to show an error when you do `odo create --downloadSource` and the devfile has multiple projects. Other option would be to always download the first, and in case devfile has multiple projects include a warning to inform the user that odo is automatically picking the first one

kadel:
> Due to a limitation with the `NoOptDefVal` attribute in `flags` the format must be `--downloadSource=<project-name>` or just `--downloadSource`. See: [spf13/pflag#134](https://github.com/spf13/pflag/issues/134)

hmm :-( is making it inconsistent with other flags :-(

pkg/odo/cli/component/create.go
```diff
@@ -899,7 +927,8 @@ func NewCmdCreate(name, fullName string) *cobra.Command {
 	componentCreateCmd.Flags().StringSliceVar(&co.componentEnvVars, "env", []string{}, "Environmental variables for the component. For example --env VariableName=Value")
 
 	if experimental.IsExperimentalModeEnabled() {
-		componentCreateCmd.Flags().BoolVar(&co.devfileMetadata.downloadSource, "downloadSource", false, "Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource")
+		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.downloadSource, "downloadSource", "", "Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource")
+		componentCreateCmd.Flags().Lookup("downloadSource").NoOptDefVal = "no-project-passed-to-flag" //Default value to pass to the flag if one is not specified.
```
kadel, line 931: Would it be possible to use an empty string (`""`) instead of `"no-project-passed-to-flag"`?

Kamran64:
@kadel - I agree with the first point, will push a commit that downloads the first project if there are multiple whilst warning the user of this.

On your second point: I'm not too sure of any feasible solutions to get around this until its fixed in the package.. We could:
- Make it mandatory that an argument is passed to the flag meaning we don't need to use `NoOptDefVal` but this may not be ideal either.

pkg/odo/cli/component/create.go
Kamran64: I tried that and it complains that no args have been passed to the flag. I think it requires some non empty string here :(

kadel:
I was thinking about it, It would solve the problem, but I'm afraid
that it will create bad UX :-(
I assume that in most of the cases, there will be just one project in
a devfile. It will be pretty annoying to force users to always specify
project name especially when we don't have an easy way to list
projects from Devfiles via cli :-(

pkg/odo/cli/component/create.go
kadel: hmm :-( ok, then lets stick with `no-project-passed-to-flag`

kadel:
> @kadel - I agree with the first point, will push a commit that downloads the first project if there are multiple whilst warning the user of this.
> 
> On your second point: I'm not too sure of any feasible solutions to get around this until its fixed in the package.. We could:
> 
> * Make it mandatory that an argument is passed to the flag meaning we don't need to use `NoOptDefVal` but this may not be ideal either.

I was thinking about it, It would solve the problem, but I'm afraid
that it will create bad UX :-(
I assume that in most of the cases, there will be just one project in
a devfile. It will be pretty annoying to force users to always specify
project name especially when we don't have an easy way to list
projects from Devfiles via cli :-(

girishramnani:
Is the project mentioned here different from `odo project` project? if so then this definitely creates confusion. because we would have openshift project, kubernetes project (namespace) and then devfile project.

neeraj-laad:
> Is the project mentioned here different from `odo project` project? if so then this definitely creates confusion. because we would have openshift project, kubernetes project (namespace) and then devfile project.

These are starter projects (specified in the Devfile). Typically there will be one, but there could be multiple projects.

- no project specified for download - download the first one (optionally, show a warning if there are multiple projects defined in the devfile)
- if project specified for download - download the specified one

Would love to get some feedback/suggestions on how to achieve this with best UX

kadel:
> These are starter projects (specified in the Devfile). Typically there will be one, but there could be multiple projects.
> 
> * no project specified for download - download the first one (optionally, show a warning if there are multiple projects defined in the devfile)

+1
I like this.

@Kamran64 I really think that odo should not jump into an interactive mode like this.

> * if project specified for download - download the specified one
>

Kamran64:
@neeraj-laad @kadel - I agree this flow makes more sense. Thanks!

Kamran64:
@girishramnani - I can see this point too. This is supposed to allow the user to specify a devfile project. We will have to clarify this in docs/help text.

GeekArthur:
> > Is the project mentioned here different from `odo project` project? if so then this definitely creates confusion. because we would have openshift project, kubernetes project (namespace) and then devfile project.
> 
> These are starter projects (specified in the Devfile). Typically there will be one, but there could be multiple projects.
> 
>     * no project specified for download - download the first one (optionally, show a warning if there are multiple projects defined in the devfile)
> 
>     * if project specified for download - download the specified one
> 
> 
> Would love to get some feedback/suggestions on how to achieve this with best UX

I think we have terminology confusion here regarding the terminology "project", with this new feature, we have two definitions for "project" now:
- OpenShift project: Similar to `namespace` in k8s world
- Devfile project: It's the project `template` that is hosted on GitHub or other source host

IMO, we should **either** make the terminology more clear in both our code (eg. help page, comment, etc) and documentation **or**  we choose to use another terminology such as `template`, otherwise it may cause more confusions for OpenShift/Red Hat users as they assume "project" is same as "namespace" which is used for isolating generic k8s resource.

GeekArthur:
> > Due to a limitation with the `NoOptDefVal` attribute in `flags` the format must be `--downloadSource=<project-name>` or just `--downloadSource`. See: [spf13/pflag#134](https://github.com/spf13/pflag/issues/134)
> 
> hmm :-( is making it inconsistent with other flags :-(

I agree with @kadel, it's inconsistent with other flags if I run `odo create --help` 
```
odo create --help                                                                                                                                                 ✔  11:19:39 
Create a configuration describing a component.

 If a component name is not provided, it'll be auto-generated.

 A full list of component types that can be deployed is available using: 'odo catalog list'

 By default, builder images (component type) will be used from the current namespace. You can explicitly supply a namespace by using: odo create namespace/name:version If version is not specified by default, latest will be chosen as the version.

Usage:
  odo create <component_type> [component_name] [flags]

Examples:
  # Create new Node.js component with the source in current directory.
  odo create nodejs

  # Create new Node.js component named 'frontend' with the source in './frontend' directory
  odo create nodejs frontend --context ./frontend

  # Create new Java component with binary named sample.jar in './target' directory
  odo create java:8  --binary target/sample.jar

  # Create new Node.js component with source from remote git repository

  # Create new Node.js component with custom ports, additional environment variables and memory and cpu limits
  odo create nodejs --port 8080,8100/tcp,9100/udp --env key=value,key1=value1 --memory 4Gi --cpu 2

Flags:
      --app string                                            Application, defaults to active application
  -b, --binary string                                         Create a binary file component component using given artifact. Works only with Java components. File needs to be in the context directory.
      --context string                                        Use given context directory as a source for component settings
      --cpu string                                            Amount of cpu to be allocated to the component. ex. 100m or 0.1 (sets min-cpu and max-cpu to this value)
      --downloadSource string[="no-project-passed-to-flag"]   Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource/--downloadSource=<devfile-project>
      --env stringSlice                                       Environmental variables for the component. For example --env VariableName=Value
  -g, --git string                                            Create a git component using this repository.
  -h, --help                                                  Help for create
      --max-cpu string                                        Limit maximum amount of cpu to be allocated to the component. ex. 1
      --max-memory string                                     Limit maximum amount of memory to be allocated to the component. ex. 100Mi
      --memory string                                         Amount of memory to be allocated to the component. ex. 100Mi (sets min-memory and max-memory to this value)
      --min-cpu string                                        Limit minimum amount of cpu to be allocated to the component. ex. 100m
      --min-memory string                                     Limit minimum amount of memory to be allocated to the component. ex. 100Mi
      --now                                                   Push changes to the cluster immediately
  -p, --port stringSlice                                      Ports to be used when the component is created (ex. 8080,8100/tcp,9100/udp)
      --project string                                        Project, defaults to active project
  -r, --ref string                                            Use a specific ref e.g. commit, branch or tag of the git repository

Additional Flags:
  -o, --o string             Specify output format, supported format: json (default "json")
```

Especially this help message `--downloadSource string[="no-project-passed-to-flag"]   Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource/--downloadSource=<devfile-project>` it's a little confused here `--downloadSource [project name]` VS `--downloadSource=[project name]` (I know part of help message is generated automatically from `cobra` package and it's hard resolve the combination issue)

GeekArthur:
Another comment from UX perspective for this use case:
```
We select the first project if one hasn't been specified in --downloadSource. If there is more than one project in the devfile we still download the first but also warn the user that multiple exist.
```

I think we should return the project name directly to user in warning message instead of simply return we are downloading the `first` project, so that user doesn't need to go back to check the devfile again for confirming which project is downloading.
```
 ⚠  There are multiple projects in this devfile but none have been specified in --downloadSource. Downloading the first.
```

Kamran64:
@GeekArthur - ah I didn't realise the default value would be displayed in the help text. Maybe we can change the name to something clearer that indicates a devfile project has to be passed after the `=` e.g. `devfile-project-name`.

GeekArthur:
Regarding integration test, even though we remove the sub-interactive mode this feature still need to be tested for full-interactive mode as users can use their own devfile with multiple/single projects, given currently we don't have tests for interactive mode, we have to wait for this issue https://github.com/openshift/odo/issues/2194, @amitkrout any progress for that issue? It would be great we can finish that as soon as possible.

Kamran64:
@GeekArthur - Are you suggesting we implement interactive mode to some level in this PR?

GeekArthur:
> @GeekArthur - Are you suggesting we implement interactive mode to some level in this PR?

Not adding extra interactive mode to this PR, just adding auto tests to test the changes with existing interactive mode to ensure no regression, but that can't be implemented for now due to https://github.com/openshift/odo/pull/3044#issuecomment-622448636, for this PR I manually test it, so we all good.

Kamran64:
Understood, just wanted to be sure. I changed the default string so hopefully the help text is a little bit more clear now. As for the `=` problem, still no solution in sight 😞

kadel:
> I think we have terminology confusion here regarding the terminology "project", with this new feature, we have two definitions for "project" now:
> 
> * OpenShift project: Similar to `namespace` in k8s world
> * Devfile project: It's the project `template` that is hosted on GitHub or other source host
> 
> IMO, we should **either** make the terminology more clear in both our code (eg. help page, comment, etc) and documentation **or** we choose to use another terminology such as `template`, otherwise it may cause more confusions for OpenShift/Red Hat users as they assume "project" is same as "namespace" which is used for isolating generic k8s resource.

I agree that the terminology is confusing :-(
But I can't think of a good solution right now :-(
If we use something like "template" then it will be confusing for anyone familiar with devfile, as there is no template in devfile :-(

pkg/odo/cli/component/create.go
```diff
@@ -899,7 +914,8 @@ func NewCmdCreate(name, fullName string) *cobra.Command {
 	componentCreateCmd.Flags().StringSliceVar(&co.componentEnvVars, "env", []string{}, "Environmental variables for the component. For example --env VariableName=Value")
 
 	if experimental.IsExperimentalModeEnabled() {
-		componentCreateCmd.Flags().BoolVar(&co.devfileMetadata.downloadSource, "downloadSource", false, "Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource")
+		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.downloadSource, "downloadSource", "", "Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource/--downloadSource=<devfile-project>")
```
kadel, line 917: There is a special section for command examples. See https://github.com/openshift/odo/blob/ba8f12df0232a6f12b4206c5211651935af838ad/pkg/odo/cli/component/create.go#L100

Can you put the example showing how to use the flag there? Having it in a flag description makes it really long and looks quite ugly and overwhelming in the terminal.

pkg/odo/cli/component/create.go
```diff
@@ -706,11 +708,24 @@ func (co *CreateOptions) downloadProject() error {
 		return errors.Errorf("No project found in devfile component.")
 	}
 
-	if nOfProjects > 1 {
-		log.Warning("Only downloading first project from list.")
-	}
+	if nOfProjects == 1 && projectPassed == "devfile-project-name" {
```
johnmcollier, line 711: Can you make `devfile-project-name` (and all instances of it) a const?

Kamran64:
@johnmcollier @kadel Done :)

adisky:
Looks good to me, i believe the documentation has been updated in this PR.
https://github.com/openshift/odo/pull/3163

PR #3056, opened 2020-04-30
Add instructions how to do commit signing

cdrage:
**What type of PR is this?**

/kind documentation
[skip ci]

Adds documentation with regards to creating a pull request / commit
signing. It is good practice to get into the habit of signing each
commit.

N/A

N/A

### DIFF
--- a/docs/dev/development.adoc
+++ b/docs/dev/development.adoc
@@ -42,6 +42,7 @@ A PR must include:
 
  * Descriptive context that outlines what has been changed and why
  * A link to the active or open issue it fixes (if applicable)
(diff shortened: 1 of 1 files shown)


docs/dev/development.adoc
```diff
@@ -42,6 +42,7 @@ A PR must include:
 
  * Descriptive context that outlines what has been changed and why
  * A link to the active or open issue it fixes (if applicable)
+ * (OPTIONAL) A signed commit is good practice. You may sign your commit using `git commit -s` or `git commit --amend --no-edit -s` to a previously created commit
```
johnmcollier, line 45: 👍

kadel: Wouldn't it be more accurate to use "sign off" rather than just "sign"?  To make it at least a little bit clear that this is not GPG signing (`-S`)?

Issue #2931, opened 2020-04-20
Run periodic job on OpenShift CI


amitkrout:
/kind feature

## Which functionality do you think we should add?
Currently we have high volume of test running for each pr which consumes almost 3hrs. So the plan here is to distribute the test in different types run. For example in nightly run all the test script will run periodically on master branch over a period of time which we should not bother about. The pr test will be validated against integration test only. 

Also there is one more benefit to add the nightly run on master branch is to catch the flake if there is any to avoid unpleasant experience while running pr test.

kadel:
/remove-area release-eng
/triage needs-informion

What kind of jobs or test will run nightly?

amitkrout:
> /remove-area release-eng
> /remove-kind feature
> /kind test
> /triage needs-information
> 
> What kind of jobs or test will run nightly?

Updated the description

mohammedzee1000:
Template: https://steps.svc.ci.openshift.org/help/ci-operator#periodic

amitkrout:
Task need to be done

- [ ] Separate integration test https://github.com/openshift/odo/pull/3066
- [ ] Create periodic job and update regular test name with the new test name on OpenShift CI https://github.com/openshift/release/pull/8707

kadel:
How are we going to handle the situations when nightly job fails? 
Is there going to be email notification when that happens? Who is going to be responsible for fixing it?

amitkrout:
> How are we going to handle the situations when nightly job fails? Is there going to be email notification when that happens?

We will follow the regular process like we have been doing, file the issue on master and fix it. BTW this is a duration gap of 6hrs periodic job, i mean in every 6 hrs the test (integration+e2e) job will be triggered. Ideally there should be an option of email notification in the job configuration. I need to figure out the way of only getting the e mail notification when the job fails otherwise no email notification is needed. Till then we need to consistently monitor the test results at - https://prow.svc.ci.openshift.org/?repo=openshift%2Fodo  

> Who is going to be responsible for fixing it?

Once the issue is figured out, the package owner will be the first choice for fixing the issue or anybody can volunteer and fix it.

kadel:
> EDIT: As per my slack communication with platform team - OpenShift CI does not support email notifications but they **do support slack alerts** for now. We can opt for slack alerts then.
> 
> >

It would be great if we can enable slack notification in a separate channel in  public odo slack

kadel:
What jobs will be executed as periodic and what jobs will run as PR tests?

[...]

PR #3066, opened 2020-05-03
Integration test only for pr validation

amitkrout:
**What type of PR is this?**
/kind cleanup

Currently we have high volume of test running for each pr which consumes almost 3hrs. So the plan here is to distribute the test in different types run. For example in nightly run all the test script will run periodically on master branch over a period of time. The pr test will be validated against integration test only.

Fixes - part of #2931

Each pr will be validated against integration test only on OpenShift CI

### DIFF
--- a/scripts/openshiftci-presubmit-integration-tests.sh
+++ b/scripts/openshiftci-presubmit-integration-tests.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+# fail if some commands fails
+set -e
+# show commands
+set -x
+
+export CI="openshift"
+make configure-installer-tests-cluster
+make bin
+mkdir -p $GOPATH/bin
+go get -u github.com/onsi/ginkgo/ginkgo
+export PATH="$PATH:$(pwd):$GOPATH/bin"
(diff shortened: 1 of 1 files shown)


amitkrout:
Once this pr goes in, i will update https://github.com/openshift/release/pull/8707 to make this changes reflected in OpenShift CI.

scripts/openshiftci-presubmit-integration-tests.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# fail if some commands fails
+set -e
+# show commands
+set -x
+
+export CI="openshift"
+make configure-installer-tests-cluster
+make bin
+mkdir -p $GOPATH/bin
```
mohammedzee1000, line 11: should the `make bin` not happen after the bin directory is created?

prietyc123: It does not matter.
Actually the ```$GOPATH/bin``` is created to keep the ginkgo binary within the bin dir. ```make bin``` just create the odo binary in the odo clone dir.

prietyc123:
Good to see separate integration test for only pr validation.

/lgtm

Issue #2688, opened 2020-03-06
odo config support for settings in devfile.yaml


elsony:
## User Story
As a developer, I want to configure my configuration settings of devfile style components in the devfile.yaml so that I can change the settings after the initial creation.

## Background
The storing of the configuration settings for the devfile style components is currently stored in the `devfile.yaml` that follows the devfile specification. However, the user currently cannot edit the settings in using the command line after the devfile in the project has been created as part of other commands. 

### for "LocalConfig" components (s2i)
- [x] `odo config` should work  as currently does

#### odo config
`odo config` is doing all operations on `devfile.yaml`, should not touch any other file.
Devfile can have multiple containers defined.  `odo config` should by default apply the configuration to all containers in devfile.

- [x] `odo config set Name` - set `metadata.name` field i in devfile.yaml
- [x] `odo config set --env FOO=bar` - add `FOO` env variable with `bar` value to all containers defined in devfile
- [x] `odo config set Ports 8080/TCP` - add new endpoint to each container in devfile
- [x] `odo config set Memory` - set `memoryLimit` for all containers in devfile.yaml
- [x] `odo config unset/view` should  work with new and updated fields 
- [x] config options in `odo confg` should have new updated description to match its meaning in devfile spec
- [x] `odo config set DebugPort` - we can **remove this option** it in favor of `odo config set --env DEBUG_PORT=12345`
- [x] `odo config set MaxCPU` - **remove this command** - devfile v2 currently doesn't have cpu limits
- [x] `odo config set MaxMemory` - **remove this command** - devfile v2 currently have only memory limit
- [x] `odo config set CPU` - **remove this command** - devfile v2 currently doesn't have cpu limits
- [x] `odo config set MinCPU` - **remove this command** - devfile v2 currently doesn't have cpu limits
- [x] `odo config set MinMemory` - **remove this command** - devfile v2 currently have only memory limit
- [x] `odo config set Ref ` - **remove this command**  - not valid for devfile
- [x] `odo config set SourceType ` - **remove this command**   - not valid for devfile
- [x] `odo config set Type ` - **remove this command**   - not valid for devfile
- [x] `odo config set SourceLocation ` - **remove this command**   - not valid for devfile

The [devfile reference](https://devfile.github.io/devfile/_attachments/api-reference.html)

## Links 
- Related issue: #2592, #3401

/kind user-story
/area devfile

cdrage:
[...]

PR #3075, opened 2020-05-04
Add devfile command flags to watch

### LINKED ISSUE #2819: Add devfile command flags to odo watch
/kind user-story

[EDITED] (Check Edit history for previous User Story Acceptance Criteria)

## User Story
We discussed when working on the PR https://github.com/openshift/odo/pull/3075, that saving commands for push flags, was not the way to go.

So, we've introduced flags `build-command`, `run-command` for odo watch

## Acceptance Criteria 
- [ ] odo watch should be able to execute with `build-command`
- [ ] odo watch should be able to execute with `run-command`

## Links 
- Feature Request: 
- Related issue: https://github.com/openshift/odo/issues/2654

/kind user-story
/area devfile

maysunfaisal:

**What type of PR is this?**
/kind feature

This PR:
- ~saves non default devfile commands `odo push --build-command="mybuild" --run-command="myrun"  --init-command="myinit" ` in `.odo/env/env.yaml`~
- ~Future push reads the non default commands from `.odo/env/env.yaml` and  executes them without having to explicitly mention them, eg `odo push` after step 1 will execute `mybuild` instead of `devbuild`~
- ~`odo watch` can also now execute the non default push commands~
- ~Any future push with non custom commands, eg `odo push --build-command="b"`, overwrites the previously saved push command in `.odo/env/env.yaml`~
- updates odo watch with devfile command flags as per https://github.com/openshift/odo/pull/3075#issuecomment-628069683, updated issue #2688 for saving devfile commands
- introduces new integration test for odo watch
- helper functions for saving devfile commands to `env.yaml` that was written as part of the initial PR before pivoting to new logic

Fixes #2819 

- Run  watch with some custom commands `odo watch --build-command="mybuild" --run-command="myrun"  --init-command="myinit" `

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -191,3 +191,4 @@ jobs:
         - sudo cp odo /usr/bin
         - export KUBERNETES=true
         - travis_wait make test-cmd-devfile-catalog
+        - travis_wait make test-cmd-devfile-watch
--- a/.travis.yml
+++ b/.travis.yml
@@ -191,4 +191,3 @@ jobs:
         - sudo cp odo /usr/bin
         - export KUBERNETES=true
         - travis_wait make test-cmd-devfile-catalog
-        - travis_wait make test-cmd-devfile-watch
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,7 +162,7 @@ jobs:
     
     - <<: *base-test
       stage: test
-      name: "docker devfile push, watch and delete command integration tests"
+      name: "docker devfile push, watch, catalog and delete command integration tests"
       script:
         - make bin
         - sudo cp odo /usr/bin
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,11 +162,12 @@ jobs:
     
     - <<: *base-test
       stage: test
-      name: "docker devfile push, url, catalog, delete command integration tests"
+      name: "docker devfile push, watch and delete command integration tests"
       script:
         - make bin
         - sudo cp odo /usr/bin
         - travis_wait make test-cmd-docker-devfile-push
+        - travis_wait make test-cmd-docker-devfile-watch
         - travis_wait make test-cmd-docker-devfile-catalog
         - travis_wait make test-cmd-docker-devfile-delete
         - travis_wait make test-cmd-docker-devfile-url
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -14,25 +14,23 @@ import (
 	"github.com/openshift/odo/pkg/util"
 )
 
-const (
-	envInfoEnvName  = "ENVINFO"
-	envInfoFileName = "env.yaml"
-)
-
 // ComponentSettings holds all component related information
 type ComponentSettings struct {
-	Name      string        `yaml:"Name,omitempty"`
-	Namespace string        `yaml:"Namespace,omitempty"`
-	URL       *[]EnvInfoURL `yaml:"Url,omitempty"`
+	Name        string              `yaml:"Name,omitempty"`
+	Namespace   string              `yaml:"Namespace,omitempty"`
+	URL         *[]EnvInfoURL       `yaml:"Url,omitempty"`
+	PushCommand *EnvInfoPushCommand `yaml:"PushCommand,omitempty"`
 }
 
 // URLKind is an enum to indicate the type of the URL i.e ingress/route
 type URLKind string
 
 const (
-	INGRESS URLKind = "ingress"
-	ROUTE   URLKind = "route"
-	DOCKER  URLKind = "docker"
+	DOCKER          URLKind = "docker"
+	INGRESS         URLKind = "ingress"
+	ROUTE           URLKind = "route"
+	envInfoEnvName          = "ENVINFO"
+	envInfoFileName         = "env.yaml"
 )
 
 // EnvInfoURL holds URL related information
@@ -44,7 +42,7 @@ type EnvInfoURL struct {
 	// Indicates if the URL should be a secure https one
 	Secure bool `yaml:"Secure,omitempty"`
 	// Cluster host
-	Host string `yaml:"host,omitempty"`
+	Host string `yaml:"Host,omitempty"`
 	// TLS secret name to create ingress to provide a secure URL
 	TLSSecret string `yaml:"TLSSecret,omitempty"`
 	// Exposed port number for docker container, required for local scenarios
@@ -53,7 +51,14 @@ type EnvInfoURL struct {
 	Kind URLKind `yaml:"Kind,omitempty"`
 }
 
-// EnvInfo holds all the env specific infomation relavent to a specific Component.
+// EnvInfoPushCommand holds the devfile push commands for the component
+type EnvInfoPushCommand struct {
+	Init  string `yaml:"Init,omitempty"`
+	Build string `yaml:"Build,omitempty"`
+	Run   string `yaml:"Run,omitempty"`
+}
+
+// EnvInfo holds all the env specific information relavent to a specific Component.
 type EnvInfo struct {
 	componentSettings ComponentSettings `yaml:"ComponentSettings,omitempty"`
 }
@@ -161,6 +166,9 @@ func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}
 			} else {
 				esi.componentSettings.URL = &[]EnvInfoURL{urlValue}
 			}
+		case "push":
+			pushCommandValue := value.(EnvInfoPushCommand)
+			esi.componentSettings.PushCommand = &pushCommandValue
 		}
 
 		return esi.writeToFile()
@@ -266,6 +274,14 @@ func (ei *EnvInfo) GetURL() []EnvInfoURL {
 	return *ei.componentSettings.URL
 }
 
+// GetPushCommand returns the EnvInfoPushCommand, returns default if nil
+func (ei *EnvInfo) GetPushCommand() EnvInfoPushCommand {
+	if ei.componentSettings.PushCommand == nil {
+		return EnvInfoPushCommand{}
+	}
+	return *ei.componentSettings.PushCommand
+}
+
 // GetName returns the component name
 func (ei *EnvInfo) GetName() string {
 	if ei.componentSettings.Name == "" {
@@ -287,16 +303,21 @@ const (
 	Create = "CREATE"
 	// CreateDescription is the description of Create parameter
 	CreateDescription = "Create parameter is the action to write devfile metadata to env.yaml"
-	// URL
+	// URL parameter
 	URL = "URL"
 	// URLDescription is the description of URL
 	URLDescription = "URL to access the component"
+	// Push parameter
+	Push = "PUSH"
+	// PushDescription is the description of URL
+	PushDescription = "Push parameter is the action to write devfile commands to env.yaml"
 )
 
 var (
 	supportedLocalParameterDescriptions = map[string]string{
 		Create: CreateDescription,
 		URL:    URLDescription,
+		Push:   PushDescription,
 	}
 
 	lowerCaseLocalParameters = util.GetLowerCaseParameters(GetLocallySupportedParameters())
--- a/pkg/envinfo/envinfo_test.go
+++ b/pkg/envinfo/envinfo_test.go
@@ -234,8 +234,80 @@ func TestDeleteURLFromMultipleURLs(t *testing.T) {
 
 }
 
+func TestGetPushCommand(t *testing.T) {
+	tempEnvFile, err := ioutil.TempFile("", "odoenvinfo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tempEnvFile.Close()
+	os.Setenv(envInfoEnvName, tempEnvFile.Name())
+
+	tests := []struct {
+		name            string
+		existingEnvInfo EnvInfo
+		wantPushCommand EnvInfoPushCommand
+	}{
+		{
+			name: "Case 1: Init, Build & Run commands present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Init:  "myinit",
+						Build: "mybuild",
+						Run:   "myrun",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Init:  "myinit",
+				Run:   "myrun",
+				Build: "mybuild",
+			},
+		},
+		{
+			name: "Case 2: Build & Run commands present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Build: "mybuild",
+						Run:   "myrun",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Run:   "myrun",
+				Build: "mybuild",
+			},
+		},
+		{
+			name: "Case 3: No commands present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			esi, err := NewEnvSpecificInfo("")
+			if err != nil {
+				t.Error(err)
+			}
+			esi.EnvInfo = tt.existingEnvInfo
+			pushCommand := esi.GetPushCommand()
+
+			if !reflect.DeepEqual(tt.wantPushCommand, pushCommand) {
+				t.Errorf("TestGetPushCommand error: push commands mismatch, expected: %v got: %v", tt.wantPushCommand, pushCommand)
+			}
+		})
+	}
+
+}
+
 func TestLowerCaseParameterForLocalParameters(t *testing.T) {
-	expected := map[string]bool{"create": true, "url": true}
+	expected := map[string]bool{"create": true, "push": true, "url": true}
 	actual := util.GetLowerCaseParameters(GetLocallySupportedParameters())
 	if !reflect.DeepEqual(expected, actual) {
 		t.Errorf("expected '%v', got '%v'", expected, actual)
--- a/pkg/envinfo/envinfo_test.go
+++ b/pkg/envinfo/envinfo_test.go
@@ -281,7 +281,76 @@ func TestGetPushCommand(t *testing.T) {
 			},
 		},
 		{
-			name: "Case 3: No commands present",
+			name: "Case 3: Build & Init commands present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Build: "mybuild",
+						Init:  "myinit",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Init:  "myinit",
+				Build: "mybuild",
+			},
+		},
+		{
+			name: "Case 4: Init & Run commands present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Init: "myinit",
+						Run:  "myrun",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Run:  "myrun",
+				Init: "myinit",
+			},
+		},
+		{
+			name: "Case 5: Build command present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Build: "mybuild",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Build: "mybuild",
+			},
+		},
+		{
+			name: "Case 6: Run command present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Run: "myrun",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Run: "myrun",
+			},
+		},
+		{
+			name: "Case 7: Init command present",
+			existingEnvInfo: EnvInfo{
+				componentSettings: ComponentSettings{
+					PushCommand: &EnvInfoPushCommand{
+						Init: "myinit",
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Init: "myinit",
+			},
+		},
+		{
+			name: "Case 8: No commands present",
 			existingEnvInfo: EnvInfo{
 				componentSettings: ComponentSettings{
 					PushCommand: &EnvInfoPushCommand{},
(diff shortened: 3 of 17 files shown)


maysunfaisal:
@johnmcollier @mik-dass pls help review the PR changes for watch and devfile commands!

pkg/envinfo/envinfo.go
```diff
@@ -44,7 +42,7 @@ type EnvInfoURL struct {
 	// Indicates if the URL should be a secure https one
 	Secure bool `yaml:"Secure,omitempty"`
 	// Cluster host
-	Host string `yaml:"host,omitempty"`
+	Host string `yaml:"Host,omitempty"`
```
johnmcollier, line 45: Does changing `host` to `Host` break anything (it shouldn't, but we should double check)

maysunfaisal: I tested kube and it was fine, but i will double confirm it again

mik-dass:
Can we have some integration tests for both push and watch? 

Also should we, instead on saving the commands during push, do it via some commands like `odo env set`? What am I missing here?

maysunfaisal:
@mik-dass @kadel that would require to execute one more command `odo env set` before doing `odo push`. Personally I'm more with the flow of saving the commands passed during `odo push --build-command="xyz" --run-command="abc"`; so that we can just do `odo push`/`odo watch` next onwards with the save commands.

As for the integration tests, how do we test `odo watch`? Seems like there is no integration tests for `odo watch` even for odo 1(non experimental) other than bad cases

mik-dass:
> As for the integration tests, how do we test odo watch? Seems like there are no integration tests for odo watch even for odo 1(non experimental) other than the bad cases (EDIT: I see e2e test have odo watch, so let me see if i can reuse it and have some integration test)

Sorry I forget that we don't have integration tests for watch. Let's skip it for now.

> @mik-dass @kadel that would require to execute one more command odo env set before doing odo push. Personally I'm more with the flow of saving the commands passed during odo push --build-command="xyz" --run-command="abc"; so that we can just do odo push/odo watch next onwards with the save commands.

I think if the user wants to do `odo push --build-command="xyz" --run-command="abc"` only for one time (maybe for debugging), he will have to rerun the command with the previous configuration to go back to the previous state. I think this will remove the user's ability to perform one time actions with the push command. Or am I missing something?

maysunfaisal:
@mik-dass yes, your understanding is correct.

Case 1: `devbuild` and `devrun` commands:
I can just do `odo push` and `odo watch` everytime to execute these commands

Case 2: custom `xyz` and `abc` commands:
I have to `odo push --build-command="xyz" --run-command="abc"` but currently there is no way for `odo watch` to push with these commands because we dont have these flags. So i think it simplifies if watch was simply `odo watch` instead of `odo watch --build-command="xyz" --run-command="abc"`

This was the idea behind saving the non-default commands; so once you run `odo push --build-command="xyz" --run-command="abc"`; you can just do `odo push` & `odo watch` to execute `xyz` and `abc` commands.

But yes, if you want to switch back to some other commands, you will have to re-run push with them: `odo push --build-command="devbuild"`, will switch to `devbuild` and still have `abc` for the run command; and you can still do `odo push` & `odo watch` with these new commands.

The idea was
1. Minimize changes to odo watch without introducing and new flags
2. Dont want to mention `build-command` or `run-command` everytime, if i am going to be pushing with them for some time. If I want to default back or change, run push again with new command.

mik-dass:
> So, to summarize: Minimize changes to odo watch without introducing any new flags for UX
Dont want to mention build-command or run-command everytime, if i am going to be pushing with them for some time. If I want to default back or change, run push again with new command.

I agree with those points :+1:  But my only concern is that a user will lose the ability to execute push with those parameters only once without the need to roll back again. The options I can think of

- A user defaults to these options using `odo env set`.
OR
- A warning message while performing this action with push about the consequences and documentation regarding the flag usage as otherwise it might confuse him/her.

maysunfaisal:
@mik-dass not sure I follow, when you say uses will lose the ability to use those parameters.. If users want they can still run push with those flags

1. `odo push --build-command="xyz"`
2. `odo push` is the same as running as `odo push --build-command="xyz"` at this point. Users can opt to use either of the way..

mik-dass:
> @mik-dass not sure I follow, when you say uses will lose the ability to use those parameters.. If users want they can still run push with those flags

I mean the user will have to use the flags again to go back to the previous state.

johnmcollier:
+1 on displaying an info message whenever the default has changed as a result of `build-command`, `run-command`, or `init-command`

maysunfaisal:
@mik-dass @johnmcollier when would you like to display the info msg:

1. whenever `--build-command`/`--run-command`/`--init-command` is used during push
2. whenever non-default commands are executed?

With option 1; it will display only whenever user executes `odo push --build-command`, not thereafter with `odo push`

With option 2; it will display the info whenever a non-default command is executed; irrespective of  `odo push --build-command`/`odo push`

johnmcollier:
Option 1 sounds good to me, I think we only need to display it when the stored/default commands in the env.yaml change

mik-dass:
> With option 1; it will display only whenever user executes odo push --build-command, not thereafter with odo push

This seems better to me :+1: Let's display it only when the stored or default commands are changed in the env.yaml.

maysunfaisal:
@mik-dass @johnmcollier I have updated the PR to now display a msg whenever the default or stored commands in env changes:

Screen Shot 2020-05-11 at 4 25 53 PM

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +189,81 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+			return strings.Contains(output, "Waiting for something to change")
+		})
+
+	Expect(success).To(Equal(true))
+	Expect(err).To(BeNil())
+}
+
+func getContainerExecListdir(runner interface{}, platform, cmpName, project, sourcePath string) string {
+	var stdOut string
+	// if !reflect.DeepEqual(runner.(helper.OcRunner), helper.OcRunner{}) {
```
mik-dass, line 255: comment found

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +189,81 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+// this was taken from e2e_images_test.go and modified for devfile
+func OdoWatch(cmpName, project, context string, stringsToBeMatched []string, runner interface{}, platform string) {
+
+	startSimulationCh := make(chan bool)
+	go func() {
+		startMsg := <-startSimulationCh
+		if startMsg {
+			err := os.MkdirAll(filepath.Join(context, ".abc"), 0750)
+			if err != nil {
+				panic(err)
```
mik-dass, line 203: I think we shouldn't panic as it will cause the test suite to panic and no subsequent tests will run. You can use gingkoRecover() to recover.

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +189,81 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+	Expect(success).To(Equal(true))
+	Expect(err).To(BeNil())
+}
+
+func getContainerExecListdir(runner interface{}, platform, cmpName, project, sourcePath string) string {
+	var stdOut string
+	// if !reflect.DeepEqual(runner.(helper.OcRunner), helper.OcRunner{}) {
+
+	if platform == "kube" {
+		ocRunner := runner.(helper.OcRunner)
```
mik-dass, line 258: we don't need this conversion

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -86,4 +96,44 @@ var _ = Describe("odo devfile watch command tests", func() {
 			Expect(output).To(ContainSubstring("Error: unknown flag: --devfile"))
 		})
 	})
+
+	Context("when executing odo watch after odo push", func() {
+		It("should listen for file changes", func() {
+			// Devfile push requires experimental mode to be set
+			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
```
mik-dass, line 103: We should move this to AfterEach()

pkg/odo/cli/component/push.go
```diff
@@ -63,13 +69,41 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string)
...
+			devfileCommandChanged = true
+		}
+
+		// only if the devfile command changed, update env.yaml and display the information
+		if devfileCommandChanged {
+			err = po.EnvSpecificInfo.SetConfiguration("push", pushCommand)
+			if err != nil {
+				return errors.Wrap(err, "unable to set push configuration for env.yaml")
+			}
+			log.Italicf("The default devfile command(s) has changed for%s", pushCommands)
```
mik-dass, line 104: I don't think we need the (s) as we the pointing the exact command

pkg/envinfo/envinfo.go
```diff
@@ -161,6 +166,9 @@ func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}
 			} else {
 				esi.componentSettings.URL = &[]EnvInfoURL{urlValue}
 			}
+		case "push":
```
mik-dass, line 169: please add a test case for this in `TestSetEnvInfo`

tests/integration/devfile/utils/utils.go
maysunfaisal: yep true but since we need it for docker runner; i thought I'd keep the conversion to keep it consistent aesthetically

tests/integration/devfile/cmd_devfile_watch_test.go
maysunfaisal: I did initially have it in `BeforeEach` but a test in watch, check for non-experimental mode  `when executing odo watch with devfile flag without experimental mode`.

I'll move the set experimental true to `BeforeEach` and reset to false for the above test

tests/integration/devfile/utils/utils.go
maysunfaisal: I'll just use `Expect(err).To(BeNil())` like the other test cases in odo

tests/integration/devfile/utils/utils.go
maysunfaisal: removed

pkg/odo/cli/component/push.go
maysunfaisal: updated

kadel:
> @mik-dass @kadel that would require to execute one more command `odo env set` before doing `odo push`. Personally I'm more with the flow of saving the commands passed during `odo push --build-command="xyz" --run-command="abc"`; so that we can just do `odo push`/`odo watch` next onwards with the save commands.

I agree with @mik-dass saving it when running `odo push` is an antipattern.
As a user, I wouldn't expect that. Flags are there to make one-time change not to persist it.

If we want to safe non-default commands it should be using `odo config` command. Or I could also see this in `odo create`. But `odo push` doesn't really make much sense.

maysunfaisal:
@kadel if `build-command` and `run-command` were intended to run as a one time, how would we expect users to run `odo watch` with these commands?

maysunfaisal:
@kadel Alternative as suggested by @johnmcollier, perhaps the cleanest way is to just add the flags to `odo watch`.

This PR was raised because `odo watch` did not have a way to push with custom commands - see issue OP discussion https://github.com/openshift/odo/issues/2819

kadel:
> @kadel Alternative as suggested by @johnmcollier, perhaps the cleanest way is to just add the flags to `odo watch`.
> 
> This PR was raised because `odo watch` did not have a way to push with custom commands - see issue OP discussion #2819

That would make sense.  `odo watch` should have the same flags as `odo push`.

The idea about saving alternative command names to env.yaml is good.
It just that it should not happen in `odo push` or `odo watch` but rather it should be separate command  something like `odo config set runCommand`

tests/integration/devfile/docker/cmd_docker_devfile_watch_test.go
```diff
@@ -0,0 +1,82 @@
...
+		dockerClient.StopContainers(label)
+
+		helper.Chdir(currentWorkingDirectory)
+		helper.DeleteDir(context)
+		os.Unsetenv("GLOBALODOCONFIG")
+	})
+
+	Context("when executing odo watch after odo push", func() {
+		It("should listen for file changes", func() {
+			helper.CmdShouldPass("git", "clone", "https://github.com/che-samples/web-nodejs-sample.git", projectDirPath)
```
maysunfaisal, line 50: I'll remove the git clone, just noticed this now

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +189,81 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
 	Expect(output).To(ContainSubstring("if not running"))
 
 }
+
+// OdoWatch creates files, dir in the context and watches for the changes to be pushed
+// this was taken from e2e_images_test.go and modified for devfile
+func OdoWatch(cmpName, project, context string, stringsToBeMatched []string, runner interface{}, platform string) {
```
prietyc123, line 195: How about using the OdoWatch function from the e2e_images_test.go itself by doing some modification over there. Like you can pass dynamic args to the function OdoWatch and include some checks if it is related to devfile.
As I can see you are replacing the app.js I think that can be handled smartly and yes you need to do modifications to some few places.

But this will avoid duplicacy of the code and ofcourse its maintainance also.

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -86,4 +96,44 @@ var _ = Describe("odo devfile watch command tests", func() {
...
+
+			// odo watch and validate
+			utils.OdoWatch(cmpName, namespace, projectDirPath, []string{"Executing devbuild command", "Executing devrun command"}, oc, "kube")
+		})
+	})
+
+	Context("when executing odo watch after odo push with custom commands", func() {
+		It("should listen for file changes", func() {
+			// Devfile push requires experimental mode to be set
+			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
```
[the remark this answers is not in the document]
maysunfaisal: done

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -9,13 +9,20 @@ import (
...
 var _ = Describe("odo devfile watch command tests", func() {
 	var namespace string
 	var context string
+	var cmpName string
+	var projectDirPath string
 	var currentWorkingDirectory string
 
+	// TODO: all oc commands in all devfile related test should get replaced by kubectl
+	// TODO: to goal is not to use "oc"
+	oc := helper.NewOcRunner("oc")
```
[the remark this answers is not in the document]
maysunfaisal: Done for `oc` only.

The `CliRunner` interface will not work for `dockerRunner` as of today, since most of those extracted `CliRunner` functions are not implemented for `dockerRunner`. Opened an issue to clean this up - https://github.com/openshift/odo/issues/3225

tests/integration/devfile/utils/utils.go
maysunfaisal: done!

maysunfaisal:
@mik-dass @johnmcollier @prietyc123 hello, could you pls help re-review the PR? I've update the PR with @kadel's feedback - we're now doing `odo watch` with the flags. The PR also has addressed the other review feedback..

pkg/envinfo/envinfo.go
maysunfaisal: done

kadel:
tested, looks good
/approve

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -9,13 +9,19 @@ import (
...
 
 var _ = Describe("odo devfile watch command tests", func() {
 	var namespace string
 	var context string
+	var cmpName string
 	var currentWorkingDirectory string
 
+	// TODO: all oc commands in all devfile related test should get replaced by kubectl
+	// TODO: to goal is not to use "oc"
+	oc := helper.NewCliRunner("oc")
```
prietyc123, line 23: Please run the changes for `kubectl` also as we are running devfile watch on minikube.

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -72,18 +77,59 @@ var _ = Describe("odo devfile watch command tests", func() {
...
+			output := helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--project", namespace)
+			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
+
+			watchFlag := ""
+			odoV2Watch := utils.OdoV2Watch{
+				CmpName:            cmpName,
+				StringsToBeMatched: []string{"Executing devbuild command", "Executing devrun command"},
+			}
+			// odo watch and validate
+			utils.OdoWatch(utils.OdoV1Watch{}, odoV2Watch, namespace, context, watchFlag, oc, "kube")
```
prietyc123, line 112: Runner should be `kubectl` and `oc` both. Please make the change throughout the test.

prietyc123:
I am holding this pr because some changes still needs to be added. Can you please go through my review comments and address it.

/hold

pkg/envinfo/envinfo_test.go
```diff
@@ -234,8 +235,80 @@ func TestDeleteURLFromMultipleURLs(t *testing.T) {
...
+					},
+				},
+			},
+			wantPushCommand: EnvInfoPushCommand{
+				Run:   "myrun",
+				Build: "mybuild",
+			},
+		},
+		{
+			name: "Case 3: No commands present",
```
prietyc123, line 284: Can't we also think of a case where Build is only present ?

tests/e2escenarios/e2e_images_test.go
```diff
@@ -125,8 +65,14 @@ var _ = Describe("odo supported images e2e tests", func() {
...
 
+		watchFlag := ""
+		odoV1Watch := utils.OdoV1Watch{
+			SrcType:  srcType,
+			RouteURL: routeURL,
+			AppName:  appName,
+		}
 		// odo watch and validate
-		OdoWatch(srcType, routeURL, project, appName, context)
+		utils.OdoWatch(odoV1Watch, utils.OdoV2Watch{}, project, context, watchFlag, oc, "kube")
```
prietyc123, line 75: +1

tests/helper/helper_cli.go
```diff
@@ -13,3 +13,16 @@ type CliRunner interface {
 	WaitAndCheckForExistence(resourceType, namespace string, timeoutMinutes int) bool
 	GetServices(namespace string) string
 }
+
+// NewCliRunner initializes new NewCliRunner with the appropriate platform
+func NewCliRunner(cli string) CliRunner {
+	var clirunner CliRunner
+	switch cli {
+	case "oc":
```
prietyc123, line 21: `kubectlRunner` should also be considered as case.

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +190,132 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+	} else if platform == "docker" {
+		dockerRunner := runner.(helper.DockerRunner)
+		containers := dockerRunner.GetRunningContainersByCompAlias(odoV2Watch.CmpName, "runtime")
+		Expect(len(containers)).To(Equal(1))
+		stdOut = dockerRunner.ExecContainer(containers[0], "ls -la "+sourcePath)
+	}
+
+	Expect(stdOut).To(ContainSubstring(("a.txt")))
+	Expect(stdOut).To(ContainSubstring((".abc")))
+	Expect(stdOut).To(Not(ContainSubstring(("abcd"))))
```
prietyc123, line 318: How  if we check for `stdout` in a single line like https://github.com/openshift/odo/blob/master/tests/helper/helper_generic.go#L73

tests/helper/helper_cli.go
```diff
@@ -13,3 +13,16 @@ type CliRunner interface {
 	WaitAndCheckForExistence(resourceType, namespace string, timeoutMinutes int) bool
 	GetServices(namespace string) string
 }
+
+// NewCliRunner initializes new NewCliRunner with the appropriate platform
+func NewCliRunner(cli string) CliRunner {
```
prietyc123, line 18: I don't think this is require because for kubernetes we do have `KUBERNETES` env var set. Also `OcRunner`  and `KubectRunner` can be used accordingly. For ex - 

```
var cliRunner helper.CliRunner

	// Using program commmand according to cliRunner in devfile
	if os.Getenv("KUBERNETES") == "true" {
		cliRunner = helper.NewKubectlRunner("kubectl")
	} else {
		cliRunner = helper.NewOcRunner("oc")
	}
```

So, I am not sure why we need this.

.travis.yml
```diff
@@ -189,4 +190,3 @@ jobs:
         - sudo cp odo /usr/bin
         - export KUBERNETES=true
         - travis_wait make test-cmd-devfile-catalog
-        - travis_wait make test-cmd-devfile-watch
```
amitkrout, line 192: This make target runs on Kubernetes cluster. It seems unfortunately it has been deleted, no ?

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +190,132 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+		getMemoryLimit := runner.(helper.OcRunner).MaxMemory(odoV1Watch.SrcType+"-app", odoV1Watch.AppName, project)
+		Expect(getMemoryLimit).To(ContainSubstring("700Mi"))
+		getMemoryRequest := runner.(helper.OcRunner).MinMemory(odoV1Watch.SrcType+"-app", odoV1Watch.AppName, project)
+		Expect(getMemoryRequest).To(ContainSubstring("400Mi"))
+	}
+}
+
+func getContainerExecListdir(odoV1Watch OdoV1Watch, odoV2Watch OdoV2Watch, runner interface{}, platform, project string, isDevfileTest bool) string {
+	var stdOut string
+	sourcePath := "/projects/nodejs-web-app"
```
amitkrout, line 296: You can directly use this path in ```ExecListDir```. This is unnecessary creation of a variable.

tests/integration/devfile/utils/utils.go
```diff
@@ -186,3 +190,132 @@ func ExecWithRestartAttribute(projectDirPath, cmpName, namespace string) {
...
+
+	if platform == "kube" {
```
amitkrout, line 298: Can you please use switch statement instead. Actually switch works much faster than equivalent if-else ladder. It is more readable. It is more manageable for having higher level of indentation than if. For example

```
switch platform {
	case "kube":
		// run the steps
		switch testType {
			case "devfile":
				// run the steps for devfile
			default:
				// run the steps for regular command test
		}
	case "docker":
		// run the steps
	default:
		fmt.Errorf("Platform %s is not supported", platform)
	}
```

tests/helper/helper_cli.go
maysunfaisal: I think its cleaner to have `cliRunner := NewCliRunner("kube")` & `cliRunner := NewCliRunner("docker")`

and in `NewCliRunner` func, do the checks you mentioned for either `oc` and `kubectl`; instead of having a if-else every time in the test we want to use `kubectl`.

But I'll make your suggested changes!

tests/helper/helper_cli.go
maysunfaisal: this function is being removed like you suggested here https://github.com/openshift/odo/pull/3075#discussion_r429518183

Issue #3077, opened 2020-05-05
Upgrade k8s dependencies


bigkevmcd:
/kind feature

## Which functionality do you think we should add?

Updated dependencies to allow supporting newer versions of Tekton (k8s.io dependencies), preferably [v0.16.5](https://github.com/tektoncd/pipeline/blob/master/go.mod#L63).

## Why is this needed?

We're driving integration with Tekton, and they've recently updated to newer k8s.io dependencies, which the current dependencies (from the glide.yaml `version: origin-4.1-kubernetes-1.13.4`) are not binary compatible.

PR #3081, opened 2020-05-05
Update openshift to 4.4. and k8s to 1.17

kadel:
**What type of PR is this?**
/kind dependency-change

- updates OpenShift libraries to 4.4 and k8s libs to 4.4
- This is required to be able to add additional functionality that is not available in the previous versions
- replaced glog with klog everywhere (https://github.com/kubernetes/klog) glog is not actively maintained anymore

Fixes https://github.com/openshift/odo/issues/3077

**How to test changes / Special notes**
just build and run odo, it should work as before


cmd/odo/odo.go
```diff
@@ -77,7 +79,7 @@ func main() {
 		case message := <-updateInfo:
 			log.Italic(message)
 		default:
-			glog.V(4).Info("Could not get the latest release information in time. Never mind, exiting gracefully :)")
+			klog.V(4).Info("Could not get the latest release information in time. Never mind, exiting gracefully :)")
```
mik-dass, line 82: why are we moving to klog instead of glog?

kadel: klog is kubernetes fork of glog. Glog is not maintained anymore.

https://github.com/kubernetes/klog

mik-dass:
https://travis-ci.com/github/openshift/odo/jobs/328172115#L723

It seems to fail because we are not providing the two required flags

mik-dass:
@kadel I can't even load the 30k+ file changes on the browser. Retried a couple of times :D

kadel:
> https://travis-ci.com/github/openshift/odo/jobs/328172115#L723
> 
> It seems to fail because we are not providing the two required flags

The new version of Cobra changed the capitalization of the first letter in the error message. This is why the test was failing

cmd/odo/odo.go
mik-dass: :+1:

mik-dass:
> The new version of Cobra changed the capitalization of the first letter in the error message. This is why the test was failing

OK :+1: I guess I missed it.

PR #3084, opened 2020-05-05
Update web

kadel:
**What type of PR is this?**
/kind documentation
[skip ci]

- update homepage to remove OpenShift Do mention
- re-sync docs (run scripts/sync.sh)


cdrage:
/lgtm
/approve

I've created PR: https://github.com/openshift/odo/pull/3085 to fix the CSS.

There is a bit of problems with outputting `code highlighting` within "notes" and "important" div's, but we can fix that later.

PR #3087, opened 2020-05-05
Update site

cdrage:
Remove the div's (for now) since there was an issue with codeblocks

/kind documentation


cdrage:
Ping @kadel I'm going to create a Docker file for doing the sync, as unfortunately we have to use an older version of Pandoc to convert the "notes" correctly.

Issue #2744, opened 2020-03-23
context support for odo push


mik-dass:
/kind feature

### User Story

As a user, I want to trigger ODO commands from outside the source code directory using the context flag.
 
### Acceptance criteria

- [ ] ODO push for devfiles should support context and should work outside of the component directory e.g `odo push --devfile test/devfile.yaml --context source-dir/`

## Why is this needed?

The user won't have to switch directories to push components.

kadel:
This needs to be converted into user story

yangcao77:
When working on context support, this issue should be take into consideration: https://github.com/openshift/odo/issues/2959

PR #3065, opened 2020-05-02
Remove devfile integration test dependency on git clone

maysunfaisal:
**What type of PR is this?**
/kind cleanup
/kind flake

**What does does this PR do / why we need it**:
- Removes external dependency on git clone for integration tests, PR now has sample project in the examples dir, and copies it for performing the integration test
- update .gitignore to list DS_Store

**Which issue(s) this PR fixes**:

#3020

**How to test changes / Special notes to the reviewer**:
make <any of the devfile test>

PR #3090, opened 2020-05-05
Support context in `odo push` commands

johnmcollier:

**What type of PR is this?**

/kind feature
/area devfile

This PR adds support for the `--context` flag to `odo push` commands. Before this PR, `--context` didn't do anything, and the devfile and env info file would only be read from the current directory odo is run from.

I've also added an integration test case for the `--context` flag.

Fixes https://github.com/openshift/odo/issues/2744

- Run integration tests
- Manual test:
    1) Create a devfile component
    2) Change to a different directory
    3) Run `odo push --context <dir-containing-component>`

### DIFF
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -100,7 +100,7 @@ var _ = Describe("odo devfile push command tests", func() {
 
 			// update devfile and push again
 			helper.ReplaceString("devfile.yaml", "name: FOO", "name: BAR")
-			helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--project", namespace, "--context", projectDirPath)
+			helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--context", projectDirPath)
 		})
 
 	})
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -95,7 +95,7 @@ var _ = Describe("odo devfile push command tests", func() {
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs"), projectDirPath)
 
-			output := helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--project", namespace, "--context", projectDirPath)
+			output := helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--context", projectDirPath)
 			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
 
 			// update devfile and push again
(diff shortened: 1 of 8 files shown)


tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -88,6 +88,21 @@ var _ = Describe("odo devfile push command tests", func() {
 			helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--project", namespace)
 		})
 
+		It("checks that odo push works outside of the context directory", func() {
+			helper.CmdShouldPass("git", "clone", "https://github.com/che-samples/web-nodejs-sample.git", projectDirPath)
```
johnmcollier, line 92: I left the `git clone` in since https://github.com/openshift/odo/pull/3065 isn't merged yet.

If https://github.com/openshift/odo/pull/3065 goes in before this PR is merged, I'll update the PR to copy the project instead of cloning it.

tests/integration/devfile/cmd_devfile_watch_test.go
```diff
@@ -63,9 +63,10 @@ var _ = Describe("odo devfile watch command tests", func() {
 			// Devfile push requires experimental mode to be set
 			helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
 			cmpName := helper.RandString(6)
-			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, "--context", context, cmpName)
+			projectDir := helper.CreateNewContext()
```
mik-dass, line 66: I think we don't need this. It's already created in the before each.

johnmcollier: @mik-dass I changed it because we were using the context flag, but staying in the same directory, so it wasn't really testing anything. I changed it so that we'd run those commands outside of the context, making sure `--context` works.

mik-dass: I would suggest removing https://github.com/openshift/odo/pull/3090/files#diff-fab0bc73a25f467aa206fd31b58a4c19R34 and changing into the context directory in each test spec or changing to the working directory `currentWorkingDirectory` here.

johnmcollier: @mik-dass Fair point. I've updated the test so that it changes into `currentWorkingDirectory` before running the `odo watch` command

mik-dass:
you will also need to change the sourcePath for watch too
```
[mrinaldas@localhost ~]$ odo watch --context /home/mrinaldas/che-nodejs
I0507 20:56:48.418615   24957 cert_rotation.go:137] Starting client certificate rotation controller
/home/mrinaldas    <-------------------------- here
 ✗  error watching source path /home/mrinaldas: open /home/mrinaldas/openshift.local.clusterup/etcd/member: permission denied
```
Putting a log statement I was able to verify that the context was not changed while running watch and thus it failed. Push works fine though.

johnmcollier:
@mik-dass Ah, you're right! Good catch about odo watch, not sure why I didn't see that before.

Easy fix, I'll push a change up, thanks!

johnmcollier:
@mik-dass Fixed!

Issue #3002, opened 2020-04-27
Download github projects in devfile at a specific revision


EnriqueL8:
/kind user-story
/area devfile

## User Story
As a developer, I want to get the correct revision for a sample project. If the devfile specifies a branch, commit or a tag for a project, I want to be able to clone that project at that specific revision. The fields used to specify a revision for a project are:
- Branch
- CommitId
- Tag
- StartPoint

## Acceptance Criteria 
- [ ] It should download the project at the specific revision if the `--downloadSource` flag is used. 
- [ ] It should fail if multiple revisions are present for that project 
- [ ] If no revision is specified for the project, it should default to master.

1. The experimental flag has been enabled
2. The component is a devfile component
3. There is a project defined in the devfile of the given component
4. odo create command called with a --downloadSource flag

## Links 
- Related issue:  #2902 

/kind user-story

PR #3100, opened 2020-05-06
Extend download source to support revisions

EnriqueL8:
**What type of PR is this?**
/kind feature

This PR adds the ability to download a sample project from the devfile at the specified revision such as Commit, Tag, StartPoint and Branch

Fixes: #3002 

Since we don't want external dependencies in these tests, I had to remove in `TestGithubZipURL` any tests that call the `GithubAPI` to retrieve the URL used to download the zip for that repository.  

The other suggestion would be to add integration tests with 4 different devfiles (branch, tag, commitId, startPoint), log out the URL to the console in verbose mode and check that the URL is there in the tests. I also cannot test that the correct revision is pull down by executing`git log` commands in the folder where the project was downloaded because the zip from Github doesn't include the `.git` folder. Would appreciate any thoughts on this.


pkg/util/util.go
```diff
@@ -785,11 +787,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
 		repo = strings.TrimSuffix(repo, ".git")
 	}
 
-	// TODO: pass branch or tag from devfile
-	branch := "master"
+	// Extract reference checkout for project
+	references := make(map[string]string)
```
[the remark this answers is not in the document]
EnriqueL8: It could be a struct but the checks will become a bit more complicated. A series of if statements to check each value, but can be done.

pkg/util/util.go
```diff
@@ -785,11 +787,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
...
+
+	if project.Source.Tag != nil && *project.Source.Tag != "" {
+		references["Tag"] = *project.Source.Tag
+	}
+
+	if project.Source.StartPoint != nil && *project.Source.StartPoint != "" {
+		references["StartPoint"] = *project.Source.StartPoint
+	}
+
+	if len(references) > 1 {
```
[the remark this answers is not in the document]
EnriqueL8: You could have a branch and commitID as per the devfiles, but a commit is universal in a repo which means that one commit can be in multiple branches. The github api accepts only one reference. Maybe we should not throw an error if branch and commitID and use the commitID as the reference. But we still throw an error in the other cases?

pkg/util/util.go
```diff
@@ -785,11 +787,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
...
+
+	var ref string
+	if len(references) == 1 {
+		// Only way to get the first element of the map
+		for _, value := range references {
+			ref = value
+		}
+	} else {
+		// Default to master if reference is not set
+		ref = "master"
```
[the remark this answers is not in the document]
EnriqueL8: Sure

pkg/util/util.go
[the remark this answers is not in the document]
johnmcollier: Okay, that's fair. If github only accepts one value as a reference, then I think we should throw an error, to prevent any confusion on the user's end.

pkg/util/util.go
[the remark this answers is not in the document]
johnmcollier: If it'll make it more complicated, that's fair. If you don't want to implement a struct, could you make `Branch`, `CommitId`, `Tag` and `StartPoint` constants?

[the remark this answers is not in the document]
EnriqueL8: Yeah, good idea

kadel:
I don't know why I haven't noticed that before. But this whole implementation is pretty specific to GitHub. What if someone is using something else?

johnmcollier:
@kadel I think due to issues with the go-git API, the initial PR for the `--downloadSource` flag (https://github.com/openshift/odo/pull/2902) was scoped to just github, and `odo` will error if a non-GitHub link is provided in the URL. @EnriqueL8 can confirm.

However, I don't see an issue open for adding support for any Git repo, so I've opened https://github.com/openshift/odo/issues/3147 (and we should have this looked at soon IMO)

EnriqueL8:
Yes, we had some issues with the `go-git` package and glide. Hopefully once we move to go modules, we can work on that. Thanks @johnmcollier for opening that issue. We can look at working on it in the next sprint.

pkg/util/util.go
```diff
@@ -785,11 +793,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
 		repo = strings.TrimSuffix(repo, ".git")
 	}
 
-	// TODO: pass branch or tag from devfile
-	branch := "master"
+	// Extract reference checkout for project
+	references := make(map[string]string)
```
cdrage, line 797: Why not a struct instead to store this information? Or just using `project.Source.`?

pkg/util/util.go
```diff
@@ -785,11 +793,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
...
+
+	if project.Source.Tag != nil && *project.Source.Tag != "" {
+		references[githubTagKey] = *project.Source.Tag
+	}
+
+	if project.Source.StartPoint != nil && *project.Source.StartPoint != "" {
+		references[githubStartPointKey] = *project.Source.StartPoint
+	}
+
+	if len(references) > 1 {
```
cdrage, line 814: I don't believe this is a good way of checking to see if there are any references? See below comment with regards to using a struct instead.

pkg/util/util.go
```diff
@@ -785,11 +793,46 @@ func GetGitHubZipURL(repoURL string) (string, error) {
...
+		errMsg := ""
+		for key, value := range references {
+			errMsg += fmt.Sprintf("%s specified with value %s\n", key, value)
+		}
+
+		return url, errors.Errorf("More than one source reference specified. The following were specified:\n%s", errMsg)
```
cdrage, line 820: errors shouldn't be capitalized

pkg/util/util.go
EnriqueL8: See: https://github.com/openshift/odo/pull/3100#discussion_r420882602 for the struct. We would need to have a few if statements, and probably some boolean to track with permutation is set to then print a nice message for the user. Maybe we can use reflect? That would give us a map of field to value, which is what I created manually. Not sure what is best. By `project.Source.` do you mean creating a var, and setting it depending on the which field is set in `project.Source`?  I did think of something like
```
var ref string
var errMsg
if project.Source.branch != nil {
   ref = project.Source.branch
}

if project.Source.tag != nil {
   if ref != nil {
       ref = project.Source.tag
   } else {
      errMsg += "tag also provided"
      // Issue with error message, and not knowing what ref was set with. Either tag or branch
   }
}
```
Please let me know if I am missing something!

pkg/util/util_test.go
```diff
@@ -1571,49 +1573,69 @@ func TestIsValidProjectDir(t *testing.T) {
...
+					Location: "https://github.com/che-samples//archive/master",
+				},
+			},
+			expectedErrors: []string{"Invalid GitHub URL: repo cannot be empty. Expecting 'https://github.com/<owner>/<repo>'"},
 		},
 		{
-			name:          "Case 5: Valid SSH Github URL",
-			zipURL:        "git@github.com:che-samples/web-nodejs-sample.git",
-			expectedError: "",
+			name: "Case 4: Invalid HTTPS Github URL with tag and commit",
```
prietyc123, line 1612: I can see only the negative test cases has been covered. Can you please add some more unit test case by considering positive scenarios also?

pkg/util/util.go
prietyc123: > I don't believe this is a good way of checking to see if there are any references?

+1

pkg/util/util.go
```diff
@@ -45,6 +46,12 @@ const maxAllowedNamespacedStringLength = 63 - len("-s2idata") - 1
 // note for mocking purpose ONLY
 var customHomeDir = os.Getenv("CUSTOM_HOMEDIR")
 
+const defaultGithubRef = "master"
```
johnmcollier, line 49: Nit: Can we wrap these in a constant block?

I.e.:
```
const (
    defaultGithubRef = "master"
    ...
)
```

pkg/util/util_test.go
EnriqueL8: I had those originally but since we don't want any external dependencies on test cases I cannot test the positive scenarios. The function calls the GithubAPI to retrieve the zip url.

pkg/util/util.go
EnriqueL8: @prietyc123 Any thoughts on: https://github.com/openshift/odo/pull/3100#discussion_r423143935 ?

pkg/util/util.go
EnriqueL8: I just pushed another commit changing the implementation to booleans, instead of using a map. Please let me know what you think @cdrage  @prietyc123

pkg/util/util_test.go
prietyc123: I am not sure if we can mock those `GithubAPI`, Anway I am seeking suggestion from @mik-dass  @girishramnani  @kadel  on how to make it happen.  @EnriqueL8  please feel free to add your suggestion too.

pkg/util/util.go
prietyc123: >I just pushed another commit changing the implementation to booleans, instead of using a map. Please let me know what you think @cdrage @prietyc123

I think too many nested if else statement making it very complicated and clumsy. IMO taking `reflect` approach would be a way simpler than this.
You can verify the `Branch` and `tags` also using switch statement.

pkg/util/util.go
prietyc123: > @prietyc123 Any thoughts on: #3100 (comment) ?

https://github.com/openshift/odo/pull/3100#discussion_r428086370

pkg/util/util.go
EnriqueL8: I think the reflect way, would be the same output as the reference map since with reflect we will just be extracting the values of each field and populating a type.

pkg/util/util_test.go
mik-dass: I think we can mock the APIs using httptest package. Here's how they did the mocking upstream https://github.com/google/go-github/blob/master/github/repos_contents_test.go#L372 and here's how they setup up the test server https://github.com/google/go-github/blob/master/scrape/scrape_test.go#L14.

EnriqueL8: I think that is a good idea. I would need to change the function definition to accept the client instance used to query the server. I think a bigger question comes to mind, which is by adding these test we will be testing the go-github package and my function just calls that api and returns. So what benefit would we gain from testing here, over it being already tested in the package?

mik-dass: Agreed. I think we need to find a way to mock the library. Maybe this discussion might help https://github.com/google/go-github/issues/113

EnriqueL8: @mik-dass Thanks for the link to that thread! I tried to implement the mocking solution they propose in https://github.com/gaia-pipeline/gaia/blob/69b69b1186c484a55133f58ff0919151baa79dfd/workers/pipeline/git.go#L211 by doing: 

```
type GithubRepoService interface {
	GetArchiveLink(ctx context.Context, owner, repo string, archiveformat archiveFormat, opts *github.RepositoryContentGetOptions, followRedirects bool)(*url.URL, *github.Response, error)
}
```
Unfortunately the type `archiveFormat` in the function definition is internal, so this interface cannot be used for both the Github Client Repo struct and the mocked version similar to the previous implementation.  I am not sure there is way to solve it without asking the library to make that definition  public. I tried putting a `string` type, because the actual type is defined as `type archiveFormat string` but that didn't work.

mik-dass: OK let's leave it for now.

EnriqueL8:
To Summarize the discussions, we have two options in the implementation:
- Using a map such as: https://github.com/openshift/odo/pull/3100/commits/5038374d74ddf04aa2fccd4fbcc359ebc248a125
- Using booleans and if statements such as: https://github.com/openshift/odo/pull/3100/commits/9e09c6f9db86ec5bac05a925a3a6241317e7378f

Please let me know what you think

adisky:
/hold
holding it for now, as it might conflict with devfile v2 PR

adisky:
removing the hold, as support for devfile V2 got merged
/hold cancel

pkg/util/util.go
```diff
@@ -796,11 +798,21 @@ func GetGitHubZipURL(repoURL string) (string, error) {
...
-	// TODO: pass branch or tag from devfile
-	branch := "master"
+	var ref string
+	if branch != "" && startPoint == "" {
+		ref = branch
+	} else if branch == "" && startPoint != "" {
+		ref = startPoint
+	} else if branch != "" && startPoint != "" {
+		return url, errors.Errorf("Branch %s and StartPoint %s specified as project reference, please only specify one", branch, startPoint)
+	} else {
```
adisky, line 808: nit:
```
if branch != "" && startPoint != ""  {
   return err
} else if branch != "" {
   ref = branch
} else if startPoint != "" {
    ref = branch
}  else {
   ref = master 
}
```

EnriqueL8:
@adisky Rebased and made change.

PR #3102, opened 2020-05-06
[WIP] Running tests on 4 nodes and gather failure

prietyc123:
**What type of PR is this?**

/kind bug

Observe all the failures occurs while running the test on 4 nodes

Fixes NA


amitkrout:
This will be the part of out periodic run https://github.com/openshift/odo/issues/2931#issuecomment-623064940 as discussed in out first cabal discussion of this sprint. In the periodic run we can increase the no of node to find out the flaky behavior in more consistent manner as it the test will be verified in each 6 hrs. 
/close

Issue #3074, opened 2020-05-04
[Docker] `odo url create` is broken if there's no Kube config


johnmcollier:
/kind bug
/area devfile
/area local

If there's no Kube config on the user's system, `odo url create` for a local odo component will fail with the following error:
```
$ odo url create
 ✗  invalid configuration: no configuration has been provided
Please login to your server:

odo login https://mycluster.mydomain.com
```

We shouldn't require a kube config to use odo's Docker functionality.

johnmcollier:
@mik-dass This seems to be related to your changes in https://github.com/openshift/odo/pull/2885, because we're always initializing a OpenShift config now (which is causing the error), rather than when only the pushtarget is not set to Docker.

kadel:
Run on this issue when I got disconnected from OpenShift cluster 
```
▶ odo preference set pushtarget docker
? pushtarget is already set. Do you want to override it in the config Yes
Global preference was successfully updated

tmp/odo-examples/demo-backend  master ✗                                                                                                                                                           33d ⚑ ◒
▶ odo url create --port 8080 -v 4
I0512 19:56:20.779057   17095 preference.go:165] The path for preference file is /Users/tkral/.odo/preference.yaml
I0512 19:56:20.779733   17095 preference.go:165] The path for preference file is /Users/tkral/.odo/preference.yaml
 ✗  Get "https://api.testocp4x.psiodo.net:6443/apis/route.openshift.io/v1?timeout=32s": dial tcp 10.0.109.148:6443: i/o timeout
```

I would say that this is a pretty important issue
/priority high

PR #3116, opened 2020-05-07
Fix URL create on Docker with no Kube config

johnmcollier:

**What type of PR is this?**

/kind bug
/area devfile
/area local

This PR fixes an issue seen when using odo with Docker and no kube config was set up on the user's machine, and `odo url create` would fail. Rather than always initializing an oc client on `odo url create` commands, it will only initialize it if the pushtarget is unset, or set to Kube.

Fixes https://github.com/openshift/odo/issues/3074

1. Move `~/.kube` to another folder if one exists on your machine
2. Set pushtarget to Docker
3. Run `odo create`, and `odo url create`. It should run fine
4. Run `odo push`, exposed port should get created

### DIFF
--- a/pkg/odo/cli/url/create.go
+++ b/pkg/odo/cli/url/create.go
@@ -100,25 +100,27 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
 		o.Context = genericclioptions.NewContext(cmd)
 	}
 
-	o.Client = genericclioptions.Client(cmd)
-
-	routeSupported, err := o.Client.IsRouteSupported()
-	if err != nil {
-		return err
-	}
-	if routeSupported {
-		o.isRouteSupported = true
-	}
-
 	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
-		if o.wantIngress || (!o.isRouteSupported) {
-			o.urlType = envinfo.INGRESS
-		} else {
-			o.urlType = envinfo.ROUTE
-		}
+		if !pushtarget.IsPushTargetDocker() {
+			o.Client = genericclioptions.Client(cmd)
 
-		if o.tlsSecret != "" && (!o.wantIngress || !o.secureURL) {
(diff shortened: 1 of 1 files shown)


johnmcollier:
@mik-dass Do you mind taking a look when you get the chance? This PR affects your route changes in odo

pkg/odo/cli/url/create.go
```diff
@@ -100,25 +100,27 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
 		o.Context = genericclioptions.NewContext(cmd)
 	}
 
-	o.Client = genericclioptions.Client(cmd)
+	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
```
adisky, line 103: IMO, there is no need of additional if 
`if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) && !pushtarget.IsPushTargetDocker()`

johnmcollier: @adisky It's required because there's code in that block farther down that's required when the pushtarget is set to Docker, and using `if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) && !pushtarget.IsPushTargetDocker()` would prevent it from executing.

pkg/odo/cli/url/create.go
```diff
@@ -100,25 +100,27 @@ func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []stri
...
-		return err
-	}
-	if routeSupported {
-		o.isRouteSupported = true
-	}
+			routeSupported, err := o.Client.IsRouteSupported()
+			if err != nil {
+				return err
+			}
+			if routeSupported {
```
girishramnani, line 111: this is redundant, please directly assign `o.isRouteSupported, err :=  := o.Client.IsRouteSupported()`

johnmcollier: @girishramnani Fixed

PR #3112, opened 2020-05-06
Performance improvement for odo catalog list components when using devfiles

metacosm:
**What type of PR is this?**
> /kind feature

**What does does this PR do / why we need it**:
Performance went from more than 4.5 seconds to less than 1.8 seconds on my system.
Also allowed me to fix #3105 as we now also fetch devfiles for completion, which was
previously prohibitively slow. Note that more performance gains could be achieved by caching locally the devfile registry metadata but this current change was a lower-hanging fruit… :)

**Which issue(s) this PR fixes**:

Fixes #3105
Fixes #2967 
Fixes #3123 

**How to test changes / Special notes to the reviewer**:
Just run `odo catalog list components` without the fix then with the fix. One noticeable change, though, is that the order in which devfiles are presented is now random.

PR #3118, opened 2020-05-07
Add --devfile flag support

### LINKED ISSUE #2862: "--devfile" support for odo create
/kind user-story

## User Story
As a user I want to use my own devfile instead of the devfile from registry, to achieve that I can use `--devfile` to specify my own devfile location 

## Acceptance Criteria 
1. Features
- [x] It should download/copy devfile from the specific location to user's working directory
- [x] --devfile flag value should support both http(s) path and file system path
- [x] Remove --devfile flag for all other commands (eg. `odo url`, `odo push`, `odo delete`, `odo watch`)
- [x] New design of existing devfile support, if user decides to use existing devfile (devfile exists in user's working directory or user specifies the existing devfile via `--devfile` flag) the first argument will be the component name instead of component type, and user can pass in the component name directly in direct mode

2. Handling errors
- [x] Validate `--devfile` flag value (eg. error out if the devfile path doesn't exist)
- [x] If an existing devfile already exists in the user's working directory, show an error to avoid overriding user's existing devfile

/kind user-story

GeekArthur:

**What type of PR is this?**
/kind feature
/area devfile

This PR implements the `--devfile` flag support with the following details:

1. Features
- [x] When `--devfile` is set, odo should download/copy devfile from the specific location to user's working directory
- [x] The value of `--devfile` should support both http(s) path and file system path
- [x] Remove `--devfile` for all other commands (eg. `odo url`, `odo push`, `odo delete`, `odo watch`), which means `--devfile` only be used by `odo create` and all other commands can directly use the devile in user's working directory instead of specifying the devfile path separately
- [x] New design of existing devfile support: if user decides to use existing devfile (devfile exists in user's working directory or user specifies the existing devfile by `--devfile`) the first argument will be the component name instead of component type, and user can pass in the component name directly in direct mode rather than interactive mode

2. Handling errors
- [x] Validate existence and correctness of `--devfile` value
- [x] If a devfile already exists in the user's working directory, error out if user still specify **another** existing devfile by `--devfile`
- [x] `--devfile` and `--registry` should be mutually exclusive flags

Fixes #2862 

Run 
- `odo create [component name] --devfile <your own devfile path>`


pkg/odo/cli/component/create.go
```diff
@@ -772,23 +779,30 @@ func (co *CreateOptions) Run() (err error) {
 	if experimental.IsExperimentalModeEnabled() {
 		// Download devfile.yaml file and create env.yaml file
 		if co.devfileMetadata.devfileSupport {
+			if !util.CheckPathExists(DevfilePath) && len(co.devfileMetadata.devfilePath) != 0 {
+				err = util.CopyFile(co.devfileMetadata.devfilePath, DevfilePath)
+				if err != nil {
+					return errors.Wrap(err, "failed to copy user specify devfile path")
```
johnmcollier, line 785: Should be: `failed to copy devfile <devfile> to <dir>` or something similar.

pkg/odo/cli/component/create.go
```diff
@@ -342,16 +343,18 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			// Interactive mode
 			// Get component type, name and namespace from user's choice via interactive mode
 
-			// devfile.yaml is not present, user has to specify the component type
+			// devfile.yaml is not present and --devfile is not specified, user has to specify the component type
 			// Component type: We provide supported devfile component list then let you choose
-			if !util.CheckPathExists(DevfilePath) {
+			if !util.CheckPathExists(DevfilePath) && len(co.devfileMetadata.devfilePath) == 0 {
```
johnmcollier, line 348: We can safely compare strings in go, so `co.devfileMetadata.devfilePath == ""` should be fine (and might make the code a little easier to read)

pkg/odo/cli/component/create.go
```diff
@@ -900,6 +914,7 @@ func NewCmdCreate(name, fullName string) *cobra.Command {
 
 	if experimental.IsExperimentalModeEnabled() {
 		componentCreateCmd.Flags().BoolVar(&co.devfileMetadata.downloadSource, "downloadSource", false, "Download sample project from devfile. (ex. odo component create <component_type> [component_name] --downloadSource")
+		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.devfilePath, "devfile", "", "Specify user's own devfile path")
```
johnmcollier, line 917: Help text should be something like `Path to a devfile.yaml`

dharmit:
Should this validation not happen before we ask the user for the details? I think this is because of the `co.Interactive` in `Complete` function while the validation happening in `Validate`.

If it makes sense, we can create an issue out of it and discuss it there.

```sh
$ odo create --devfile /tmp/devfile.yaml

? What do you wish to name the new devfile component nothing
? What namespace do you want the devfile component to be created in newproject
 ✗  Validating devfile component [69997ns]
 ✗  This workspace directory already contains a devfile component
```

dharmit:
OK, this gets interesting. Continuing from my above comment:
```sh
$ odo create --devfile /tmp/devfile.yaml

? What do you wish to name the new devfile component nothing
? What namespace do you want the devfile component to be created in newproject
 ✗  Validating devfile component [84901ns]
 ✗  This workspace directory already contains a devfile component

$ odo delete --all -f

$ tree .odo
.odo
└── env
    └── env.yaml

1 directory, 1 file

$ ls devfile.yaml
ls: cannot access 'devfile.yaml': No such file or directory
```

So, for `odo create`, existence of `env.yaml` is an indicator that `pwd` contains a component. But for `odo delete`, it isn't. Am I looking at this the right way or is my understanding flawed? @GeekArthur @johnmcollier @kadel

dharmit:
This one's got me confused. Why does it take us 9 seconds to tell the user that a `devfile.yaml` already exists in `pwd`? Shouldn't that be faster? :thinking: 
```sh
$ time odo create --devfile /tmp/devfile.yaml

 ✗  This directory already contains a devfile.yaml, you can't specify devfile path via --devfile
odo create --devfile /tmp/devfile.yaml  0.15s user 0.02s system 1% cpu 9.276 total
```

pkg/odo/cli/component/create.go
```diff
@@ -65,6 +65,7 @@ type DevfileMetadata struct {
 	devfileLink        string
 	devfileRegistry    string
 	downloadSource     bool
+	devfilePath        string
```
adisky, line 68: should we add comments here for each structs fields?? as sometimes it makes difficult to understand the code .
e.g. `util.CheckPathExists(DevfilePath) && len(co.devfileMetadata.devfilePath)`

adisky:
> This one's got me confused. Why does it take us 9 seconds to tell the user that a `devfile.yaml` already exists in `pwd`? Shouldn't that be faster? 

I have raised this point earlier too, actually call to  `Listdevfilecomponents` is the cause of the delay, if possible all required error checks should be made before this call. https://github.com/openshift/odo/blob/9411894b1b50ea0caf7f09e7810848e6997030ef/pkg/odo/cli/component/create.go#L332

dharmit:
> I have raised this point earlier too, actually call to `Listdevfilecomponents` is the cause of the delay, if possible all required error checks should be made before this call.

In that case, I guess #3112 should help address the concerns in terms of seconds. But yes I do feel like some checks should be made sooner rather than later.

pkg/odo/cli/component/create.go
GeekArthur: This is like `tab` VS `space`, firstly I am pretty sure there is no performance difference from Go compiler perspective between `len(x) == 0` and `x == ""` as Go compiler treats them the same. Seems like the discussion is pretty hot on stackoverflow as well https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go.

I am fine for both, but the reason I choose `len(x) == 0` is consistency as in this file we have ~21 places use `len(x) == 0`

GeekArthur:
> OK, this gets interesting. Continuing from my above comment:
> 
> ```shell
> $ odo create --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
> ? What do you wish to name the new devfile component nothing
> ? What namespace do you want the devfile component to be created in newproject
>  ✗  Validating devfile component [84901ns]
>  ✗  This workspace directory already contains a devfile component
> 
> $ odo delete --all -f
>  ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component
> 
> $ tree .odo
> .odo
> └── env
>     └── env.yaml
> 
> 1 directory, 1 file
> 
> $ ls devfile.yaml
> ls: cannot access 'devfile.yaml': No such file or directory
> ```
> 
> So, for `odo create`, existence of `env.yaml` is an indicator that `pwd` contains a component. But for `odo delete`, it isn't. Am I looking at this the right way or is my understanding flawed? @GeekArthur @johnmcollier @kadel
> 
> **EDIT**: I'm not suggesting we fix it in this PR. It's a separate issue and PR of it's own, provided my understanding is correct.

I think `env.yaml` is the only indicator that we can use to decide if the devfile component is present or not and `odo delete` should check that indicator and clean that up with user's confirmation. It's similar to s2i component, I think only difference is devfile component uses `env.yaml` and s2i component uses `config.yaml`.

GeekArthur:
> Should this validation not happen before we ask the user for the details? I think this is because of the `co.Interactive` in `Complete` function while the validation happening in `Validate`.
> 
> If it makes sense, we can create an issue out of it and discuss it there.
> 
> ```shell
> $ odo create --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
> ? What do you wish to name the new devfile component nothing
> ? What namespace do you want the devfile component to be created in newproject
>  ✗  Validating devfile component [69997ns]
>  ✗  This workspace directory already contains a devfile component
> ```

I am not very clear about this comment, we have validations in both Complete() and Validate(), because we need to validate if the devfile component is supported by us, if not we continue to create s2i component, but all validation happens after we get input values from user.

GeekArthur:
> This one's got me confused. Why does it take us 9 seconds to tell the user that a `devfile.yaml` already exists in `pwd`? Shouldn't that be faster? 🤔
> 
> ```shell
> $ time odo create --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
>  ✗  This directory already contains a devfile.yaml, you can't specify devfile path via --devfile
> odo create --devfile /tmp/devfile.yaml  0.15s user 0.02s system 1% cpu 9.276 total
> ```
> 
> **EDIT 1**: Same for specifying component type along with `--devfile` flag:
> 
> ```shell
> $ time odo create nodejs --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
>  ✗  Component type and --devfile can not be specified together
> odo create nodejs --devfile /tmp/devfile.yaml  0.18s user 0.01s system 1% cpu 14.590 total
> ```
> 
> I understand that the behaviour I've reported here and in previous comments doesn't indicate any bug. But this is sending a bad user experience, IMO.
> 
> **EDIT 2**: If I specify an invalid file to the `--devfile` flag, it's not throwing an error right away but taking 10+ seconds. See below:
> 
> ```shell
> $ time odo create nodejs --devfile /tmp/nosuchfile
> Experimental mode is enabled, use at your own risk
> 
>  ✗  Component type and --devfile can not be specified together
> odo create nodejs --devfile /tmp/nosuchfile  0.16s user 0.03s system 1% cpu 11.891 total
> 
> $ time odo create --devfile /tmp/nosuchfile      
> Experimental mode is enabled, use at your own risk
> 
> ? What do you wish to name the new devfile component nothing
> ? What namespace do you want the devfile component to be created in newproject
>  ✓  Validating devfile component [48332ns]
>  ✗  stat /tmp/nosuchfile: no such file or directory
> odo create --devfile /tmp/nosuchfile  0.18s user 0.03s system 1% cpu 15.457 total
> ```
> 
> I'm sorry but I'm going to hold this PR till the UX issues are addressed. @kadel @elsony @girishramnani @johnmcollier if you think these issues are OK to live with, feel free to remove the hold.
> 
> /hold

Thanks for testing those error handling! Although my personal system doesn't have the delay issue, I still need to fix this to tolerate all systems. The root cause of all scenarios can be boiled down to calling this function `Listdevfilecomponents()`, I can do some logic change to improve the performance to cover all scenarios that you test.

pkg/odo/cli/component/create.go
GeekArthur: Regarding `DevfilePath` VS `co.devfileMetadata.devfilePath` that you point out, actually the difference of the two variables are self-explanatory from _the way we access them_. `DevfilePath` is exported global variable, but  `co.devfileMetadata.devfilePath` is class field (I know Go doesn't have class but struct + receiver can implement the OOP in Go), given that we can clearly know `co.devfileMetadata.devfilePath` belongs to each individual component and `DevfilePath` is the default one that can be shared by all components. (It's like belongs to Class VS Instance with/without static keyword in Java).

I don't think add comments to all struct fields is a good approach as most of them are self-explanatory and even today the s2i component doesn't do that either. That being said, I think the best plan is to check which field can cause confusion and I can add comment to those kind of fields.

GeekArthur:
> > This one's got me confused. Why does it take us 9 seconds to tell the user that a `devfile.yaml` already exists in `pwd`? Shouldn't that be faster?
> 
> I have raised this point earlier too, actually call to `Listdevfilecomponents` is the cause of the delay, if possible all required error checks should be made before this call.
> 
> https://github.com/openshift/odo/blob/9411894b1b50ea0caf7f09e7810848e6997030ef/pkg/odo/cli/component/create.go#L332

I think I reply this under this comment https://github.com/openshift/odo/pull/2699#discussion_r398213652 before. But I rethink the structure, I think it's fine we check check the existence for both s2i and devfile components, it's not necessary to wait for devfile support check.

GeekArthur:
> This one's got me confused. Why does it take us 9 seconds to tell the user that a `devfile.yaml` already exists in `pwd`? Shouldn't that be faster? 🤔
> 
> ```shell
> $ time odo create --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
>  ✗  This directory already contains a devfile.yaml, you can't specify devfile path via --devfile
> odo create --devfile /tmp/devfile.yaml  0.15s user 0.02s system 1% cpu 9.276 total
> ```
> 
> **EDIT 1**: Same for specifying component type along with `--devfile` flag:
> 
> ```shell
> $ time odo create nodejs --devfile /tmp/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
>  ✗  Component type and --devfile can not be specified together
> odo create nodejs --devfile /tmp/devfile.yaml  0.18s user 0.01s system 1% cpu 14.590 total
> ```
> 
> I understand that the behaviour I've reported here and in previous comments doesn't indicate any bug. But this is sending a bad user experience, IMO.
> 
> **EDIT 2**: If I specify an invalid file to the `--devfile` flag, it's not throwing an error right away but taking 10+ seconds. See below:
> 
> ```shell
> $ time odo create nodejs --devfile /tmp/nosuchfile
> Experimental mode is enabled, use at your own risk
> 
>  ✗  Component type and --devfile can not be specified together
> odo create nodejs --devfile /tmp/nosuchfile  0.16s user 0.03s system 1% cpu 11.891 total
> 
> $ time odo create --devfile /tmp/nosuchfile      
> Experimental mode is enabled, use at your own risk
> 
> ? What do you wish to name the new devfile component nothing
> ? What namespace do you want the devfile component to be created in newproject
>  ✓  Validating devfile component [48332ns]
>  ✗  stat /tmp/nosuchfile: no such file or directory
> odo create --devfile /tmp/nosuchfile  0.18s user 0.03s system 1% cpu 15.457 total
> ```
> 
> I'm sorry but I'm going to hold this PR till the UX issues are addressed. @kadel @elsony @girishramnani @johnmcollier if you think these issues are OK to live with, feel free to remove the hold.
> 
> /hold

@dharmit My latest commit improves the performance of error handling, you can check again on your system

Case 1: Component is present, **From ~9s to ~0.6s**
```
time ( repeat 10; odo create --devfile /Users/jingfuwang/wang/devfile.yaml)                                  ✔  3s   17:32:49 

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component

 ✗  This directory already contains a component
( repeat 10; do; odo create --devfile /Users/jingfuwang/wang/devfile.yaml; ; )  0.69s user 0.16s system 128% cpu 0.652 total
```

Case 2: Component type conflict, **From ~15s to ~0.6**
```
time ( repeat 10; odo create nodejs --devfile /fakefile)                                                           1 ✘  17:31:59 

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together

 ✗  Component type and --devfile can not be specified together
( repeat 10; do; odo create nodejs --devfile /fakefile; done; )  0.73s user 0.18s system 131% cpu 0.690 total
```

Case 3: Fake devfile path, **From ~15s to 0.6s**
```
time ( repeat 10; odo create --devfile /fakefile)                                                                  1 ✘  17:32:26 

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory

 ✗  stat /fakefile: no such file or directory
( repeat 10; do; odo create --devfile /fakefile; done; )  0.72s user 0.16s system 130% cpu 0.675 total
```

dharmit:
> > OK, this gets interesting. Continuing from my above comment:
> > ```shell
> > $ odo create --devfile /tmp/devfile.yaml
> > Experimental mode is enabled, use at your own risk
> > 
> > ? What do you wish to name the new devfile component nothing
> > ? What namespace do you want the devfile component to be created in newproject
> >  ✗  Validating devfile component [84901ns]
> >  ✗  This workspace directory already contains a devfile component
> > 
> > $ odo delete --all -f
> >  ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component
> > 
> > $ tree .odo
> > .odo
> > └── env
> >     └── env.yaml
> > 
> > 1 directory, 1 file
> > 
> > $ ls devfile.yaml
> > ls: cannot access 'devfile.yaml': No such file or directory
> > ```
> > 
> > 
> > So, for `odo create`, existence of `env.yaml` is an indicator that `pwd` contains a component. But for `odo delete`, it isn't. Am I looking at this the right way or is my understanding flawed? @GeekArthur @johnmcollier @kadel
> > **EDIT**: I'm not suggesting we fix it in this PR. It's a separate issue and PR of it's own, provided my understanding is correct.
> 
> I think `env.yaml` is the only indicator that we can use to decide if the devfile component is present or not and `odo delete` should check that indicator and clean that up with user's confirmation. It's similar to s2i component, I think only difference is devfile component uses `env.yaml` and s2i component uses `config.yaml`.

But this is confusing. I'm guessing this is by design.

pkg/odo/cli/component/create.go
kadel: It is about code readability, reading `o.devfileMetadata.devfilePath == ""` you understand right away what it is doing `len(co.devfileMetadata.devfilePath) == 0` now there if a function call, it it forces you to think a little bit what the author wanted from this line.

kadel:
This implementation of ` --devfile` flag for `odo create` is completely out of the place. 
The same flag is used in other commands but it means something else. 
In `odo push` `--devfile` specify devfile that should be used instead of reading it from the current directory. 
Using this logic the `odo create --devfile /path/mydevfile.yaml` should create devfile in `/path/mydevfile.yaml` not copy it form this path to the current directory.

Another issue I have with this is why we even need this?  I wouldn't a user just copy the file? What benefit there is using odo command for simply comping devfile from one path to other?

pkg/odo/cli/component/create.go
adisky: Its not about the class variable, instance variable, if you look good code at first glance https://github.com/openshift/odo/blob/master/pkg/odo/cli/component/create.go#L82, it suggests `DevfilePath` is path of `devfile.yaml`, and `co.devfileMetadata.devfilePath` is also a path of `devfile.yaml`. One is user specified one, another is the default one, that is not clear by variable names, ideally the default one should be renamed to `DefaultDevfilePath`

An Agreed S2I components also does not have sufficient commenting, resulting on duplicate field values some times. e.g. Unless I read the code below , I dont know what `devfileLink` field is used for. sometimes struct fields are self explanatory to us(who writes the code), not to others.

adisky:
> This implementation of ` --devfile` flag for `odo create` is completely out of the place.
> The same flag is used in other commands but it means something else.
> In `odo push` `--devfile` specify devfile that should be used instead of reading it from the current directory.
> Using this logic the `odo create --devfile /path/mydevfile.yaml` should create devfile in `/path/mydevfile.yaml` not copy it form this path to the current directory.
> 
> Another issue I have with this is why we even need this? I wouldn't a user just copy the file? What benefit there is using odo command for simply comping devfile from one path to other?

+1, the `odo push` --devfile flag, confused me too.

GeekArthur:
> This implementation of ` --devfile` flag for `odo create` is completely out of the place.
> The same flag is used in other commands but it means something else.
> In `odo push` `--devfile` specify devfile that should be used instead of reading it from the current directory.
> Using this logic the `odo create --devfile /path/mydevfile.yaml` should create devfile in `/path/mydevfile.yaml` not copy it form this path to the current directory.

Actually all `--devfile` flag means the same thing, they all mean use the devfile that is specified via `--devfile` flag (user doesn't need to pass that flag any more), _But_ they do have difference and let me clarify this, `odo push --devfile`use the user specify devfile directly, but `odo create --devfile` copy the user specify devfile to user's working directory then let all the subsequent commands (eg. odo push, odo url) to use that user specify devfile, the plan is we will only keep `--devfile` for `odo create` and depreciate the `--devfile` for all other commands so that all other commands can assume the user specify devfile is present under user's working directory.

> Another issue I have with this is why we even need this? I wouldn't a user just copy the file? What benefit there is using odo command for simply comping devfile from one path to other?

From my understanding, the value of this feature is to make other commands' lives easier as they don't need `--devfile`. To be honest, I am the person who request this feature (but I do create the issue and implement this feature), @yangcao77 requests this and she may have more points regarding the value.

yangcao77:
As we discussed in our meeting, if the user wants to bring their own devfile, they can do `odo create --devfile`, which would copy the devfile content to the project context directory. 

And for other commands, `--devfile` should be removed. As we would assume there is an existing devfile under the context directory. If not, then it's not a valid odo component for devfile. 

So I think in this PR, @kadel and @adisky are correct, `--devfile` for other commands does not make sense, you need to remove the `--devfile` flag from other commands.

GeekArthur:
@kadel I think all explanations and implementations are above, please let me know what your thoughts/comments then we can decide how we proceed this (Drop/Implement the feature).

GeekArthur:
> > > OK, this gets interesting. Continuing from my above comment:
> > > ```shell
> > > $ odo create --devfile /tmp/devfile.yaml
> > > Experimental mode is enabled, use at your own risk
> > > 
> > > ? What do you wish to name the new devfile component nothing
> > > ? What namespace do you want the devfile component to be created in newproject
> > >  ✗  Validating devfile component [84901ns]
> > >  ✗  This workspace directory already contains a devfile component
> > > 
> > > $ odo delete --all -f
> > >  ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component
> > > 
> > > $ tree .odo
> > > .odo
> > > └── env
> > >     └── env.yaml
> > > 
> > > 1 directory, 1 file
> > > 
> > > $ ls devfile.yaml
> > > ls: cannot access 'devfile.yaml': No such file or directory
> > > ```
> > > 
> > > 
> > > So, for `odo create`, existence of `env.yaml` is an indicator that `pwd` contains a component. But for `odo delete`, it isn't. Am I looking at this the right way or is my understanding flawed? @GeekArthur @johnmcollier @kadel
> > > **EDIT**: I'm not suggesting we fix it in this PR. It's a separate issue and PR of it's own, provided my understanding is correct.
> > 
> > 
> > I think `env.yaml` is the only indicator that we can use to decide if the devfile component is present or not and `odo delete` should check that indicator and clean that up with user's confirmation. It's similar to s2i component, I think only difference is devfile component uses `env.yaml` and s2i component uses `config.yaml`.
> 
> But this is confusing. I'm guessing this is by design.

I am not sure originally who/how to implement the `odo delete`, but I don't think that's correct behavior based on the output, and `odo delete` should align with `odo create`, you can open an issue to track this.

dharmit:
> @dharmit I don't know originally who/how to implement the `odo delete`, but I don't think that's correct behavior based on the output, and `odo delete` should align with `odo create`, you can open an issue to track this (how does `odo delete` handle `env.yaml`).

#3155

kadel:
> As we discussed in our meeting, if the user wants to bring their own devfile, they can do `odo create --devfile`, which would copy the devfile content to the project context directory.
> 
> And for other commands, `--devfile` should be removed. As we would assume there is an existing devfile under the context directory. If not, then it's not a valid odo component for devfile.

+1

If we remove `--devfile` flag from other commands this would make a lot of sense. It would make it nice and clean and much more understandable.
I would add one more requirement to this: Make it also work with HTTP(s). Being able to do `odo create --devfile https://raw.githubusercontent.com/eclipse/che-devfile-registry/master/devfiles/nodejs/devfile.yaml` would be awesome. It is going to be much more useful than with just a local file.

pkg/odo/cli/component/create.go
```diff
@@ -342,16 +340,34 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+				if err != nil {
+					return err
+				}
+
+				if file.IsDir() {
+					return errors.Errorf("%s exists but it's not file", co.devfileMetadata.devfilePath)
+				}
+			}
+
+			if !util.CheckPathExists(DevfilePath) && len(co.devfileMetadata.devfilePath) == 0 {
```
kadel, line 354:
```suggestion
			if !util.CheckPathExists(DevfilePath) && co.devfileMetadata.devfilePath == "" {
```

pkg/odo/cli/component/create.go
kadel: > I don't think add comments to all struct fields is a good approach as most of them are self-explanatory and even today the s2i component doesn't do that either.

This is one of the many thinks that is wrong with old odo codebase handling s2i components. This should not be an excuse for not properly comment code.

pkg/odo/cli/component/create.go
GeekArthur: Okay, I will make change to use `x == ""` style instead.

pkg/odo/cli/component/create.go
GeekArthur: No problem, rename variable and comments will be added on my next commit.

pkg/odo/cli/component/create.go
GeekArthur: Yep, I will use `x == ""` style to handle the empty comparison instead if `len(x) == 0`

GeekArthur:
Thanks for everyone's comments and suggestions! I update the PR and issue description to add more requirements, you guys can feel free to add more.

adisky:
@GeekArthur I know this has been discussed above, But I still have a concern on this, I am not able to specify component name also, this is forcing me to go in interactive mode.
```
$ odo create mycomp --devfile https://raw.githubusercontent.com/elsony/devfile-registry/master/devfiles/nodejs/devfile.yaml

 ✗  Component type and --devfile can not be specified together
```

kadel:
If I specify a `--devfile` flag it should not enter interactive mode.


```
▶ odo create --devfile https://raw.githubusercontent.com/kadel/demo-backend/master/devfile.yaml

? What do you wish to name the new devfile component
```

If we need a name for the component we could autogenerate it based on the current directory name.

GeekArthur:
> @GeekArthur I know this has been discussed above, But I still have a concern on this, I am not able to specify component name also, this is forcing me to go in interactive mode.
> 
> ```
> $ odo create mycomp --devfile https://raw.githubusercontent.com/elsony/devfile-registry/master/devfiles/nodejs/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
>  ✗  Component type and --devfile can not be specified together
> ```

@adisky This is a good point and I was also thinking about this before as force user to interactive mode is not user friendly from UX perspective given user may need a way to pass in the component name explicitly with the use case you mention above.

I think the generic question is "how does user specify the component name without specifying component type in direct mode?" and I can think of two use cases related with the question:
**Use case 1**. User wants to use existing devfile in user's working directory (either PWD or user specify directory via `--context`) -> user can't specify component name in direct mode
**Use case 2**. User wants to use the devfile that is specified in `--devfile` -> user can't specify component name in direct mode

I think the root cause of the problem is the design of the `odo create` usage, currently `odo create` usage is:
```
odo create [component type] [component name] [flags]
```
**For use case 1**: If user wants to use the existing devfile, user can't specify component type as component type may have conflict with the existing devfile and currently we don't have a clear way to check the component type from devfile itself (We may or may not have that field in devfile 2.0), given that if user still wants to specify component name, the component name argument will become the first argument which odo thinks the component name is the component type then cause issue
**For use case 2**: Similar reason as use case 1, only difference is that there is conflict between component type and devfile from `--devfile` (Use case 1 is conflict between component type and existing devfile in user's working directory)

@kadel 's approach below can solve the problem "force user to interactive mode", but it still doesn't let user specify the component name, it forces user to use the static component name which is the directory name instead, which the concern of UX is still there
> If I specify a `--devfile` flag it should not enter interactive mode.
> 
> ```
> ▶ odo create --devfile https://raw.githubusercontent.com/kadel/demo-backend/master/devfile.yaml
> Experimental mode is enabled, use at your own risk
> 
> ? What do you wish to name the new devfile component
> ```
> 
> If we need a name for the component we could autogenerate it based on the current directory name.

Back to the original question "how does user specify the component name without specifying component type in direct mode?" I think we have the following options:
1. Add component type (eg. springboot, nodejs, quarkus) to devfile to check if the component type matches with existing devfile in user's working directory/devfile from --devfile (downside: user has to always specify component type for the two use cases above as we need that for validation)
2. Use directory name as component name (downside: user can't specify component name for the two use cases above)
3. Let user specify component name as `flag` instead of `argument` then we don't have argument order issue such as second argument becomes first argument if no first argument. Then component type will become the only indicator that differentiate between interactive mode and direct mode (downside: we may have many changes on both devfile and s2i components)

That's all I can think of regarding this UX issue, @adisky @kadel please let me know your comments and thoughts, or probably you have greater idea to address that 🤔

johnmcollier:
Can we just prevent the user from specifying a component type as an arg if the `--devfile` flag is set? 

That is, if `--devfile` is set, then the argument after `odo create` becomes the component name, rather than the component type. e.g.:

```
$ odo create mycomp --devfile devfile.yaml
```
Would create a component from the devfile.yaml with name `mycomp`

GeekArthur:
> If `--devfile` is set, and the user passed in an arg to `odo create`, can we just interpret that as the component name, rather than component type?
> 
> That is, if `--devfile` is set, then the argument after `odo create` becomes the component name, rather than the component type. e.g.:
> 
> ```
> $ odo create mycomp --devfile devfile.yaml
> ```
> 
> Would create a component from the devfile.yaml with name `mycomp`
> 
> and
> 
> ```
> $ odo create nodejs mycomp
> ```
> 
> would still create a nodejs component named `mycomp`

It's easy to implement, and we need to apply the same idea to existing devfile in user's working directory case as well. For UI, we need to do two additional things:
1. Locally validate the number of arguments for the two use cases we talk above instead of using Cobra package to do global validation and throw proper errors out if user still pass in two arguments with `--devfile` or existing devfile
2. Carefully modify the `odo create` help page to resolve the confusion for user as we have different argument meaning for different use case

The biggest concern is this change may break the user behavior consistency as user may get confused about "when should I pass what argument?", because IMHO let user constantly check the help page or memorize the difference of usage is not that ideal, but this approach can be option 4 I think.

johnmcollier:
IMO, I don't think the risk of user confusion is that high, provided we give clear examples in the help page for `odo create` and error out when `odo create <comp-type> <comp-name> --devfile <devfile>` is given. 

If the user's passing in the `--devfile` flag, they're likely aware that they don't need to pass in the component type to `odo create`.

yangcao77:
+1 with that. We should allow user to specify a component name during creation, no matter whether there is an existing devfile.

docs/dev/experimental-mode.adoc
```diff
@@ -105,7 +105,7 @@ Applying URL changes
  ✓  Changes successfully pushed to component
 ```
 
-By default, `odo push` executes `devbuild` and `devrun` devfile commands. However, custom commands can be also be provided to `odo push` via flags `--build-command` & `--run-command`. `odo push` also provides a `--devfile` flag to execute push with a `devfile.yaml` in a custom path. These flags are only available in the experimental mode.
+By default, `odo push` executes `devbuild` and `devrun` devfile commands. However, custom commands can be also be provided to `odo push` via flags `--build-command` & `--run-command`. These flags are only available in the experimental mode.
```
kadel, line 108: 👍

adisky:
I think to make it simple we could do this
`odo create  mycomp --devfile devfile.yaml`  creates a component name with `mycomp`
`odo create mycomp nodejs` --devfile devfile.yaml errors out for specifying two args with --devfile
`odo create nodejs --devfile devfile.yaml` creates a component with name `nodejs`
`odo create --devfile devfile.yaml` generate name from current dir/ or from devfile metadata.

kadel:
> I think to make it simple we could do this
> `odo create mycomp --devfile devfile.yaml` creates a component name with `mycomp`
> `odo create mycomp nodejs` --devfile devfile.yaml errors out for specifying two args with --devfile
> `odo create nodejs --devfile devfile.yaml` creates a component with name `nodejs`
> `odo create --devfile devfile.yaml` generate name from current dir/ or from devfile metadata.

As I understand this is aligned with what @johnmcollier proposes.
Currently, this might be the best solution that would have least impact on the current CLI design.

To be honest I think that what we originally did with  `odo create` command where we have two positional arguments and one is optional is not a good approach. `--name` flag would be much better.
I will start a discussion if it makes sense to start introducing breaking changes before we release odo v2, and analyse how big impact it would have. 
Meanwhile, I would go with what is described above.

GeekArthur:
> > I think to make it simple we could do this
> > `odo create mycomp --devfile devfile.yaml` creates a component name with `mycomp`
> > `odo create mycomp nodejs` --devfile devfile.yaml errors out for specifying two args with --devfile
> > `odo create nodejs --devfile devfile.yaml` creates a component with name `nodejs`
> > `odo create --devfile devfile.yaml` generate name from current dir/ or from devfile metadata.
> 
> As I understand this is aligned with what @johnmcollier proposes.
> Currently, this might be the best solution that would have least impact on the current CLI design.
> 
> To be honest I think that what we originally did with `odo create` command where we have two positional arguments and one is optional is not a good approach. `--name` flag would be much better.

Yeah, I agree with you, let user pass in component name as flag (eg. `--name` or something similar) would be better and that's the option 3 of my proposal above https://github.com/openshift/odo/pull/3118#issuecomment-628094380

> I will start a discussion if it makes sense to start introducing breaking changes before we release odo v2, and analyse how big impact it would have.
> Meanwhile, I would go with what is described above.

When we discuss/redesign the alternatives, IMO for CLI design we should deeply think the type of parameter that user passes in, which is `Argument` vs `Flag`, especially when we choose to use `optional argument` (From my experience, most argument is mandatory). I believe this rule can apply to all new features/commands.

If there is no objection here, I would proceed the following proposal for both `--devfile` case and `existing devfile in user's working directory` case, so final proposal would be:
```
Devfile doesn't exist in user's working directory:
- odo create -> interactive mode
- odo create <component type> <component name> -> successfully create component
- odo create --devfile <path to devfile> -> successfully create component + use component directory name as component name
- odo create <component name> --devfile <path to devfile> -> successfully create component
- odo create <component type> <component name> --devfile <path to devfile> -> error out

Devfile exists in user's working directory:
- odo create -> successfully create component + use component directory name as component name
- odo create <component name> -> successfully create component
- odo create <component type> <component name> -> error out
- odo create --devfile <path to devfile> -> error out
- odo create <component name> --devfile <path to devfile> -> error out
```

@elsony Can you please review this proposal as well?

johnmcollier:
@GeekArthur One small modification:

If the user specifies the devfile in current directory with the devfile flag (e.g. `odo create --devfile devfile.yaml`), can we not error out?

To me, to create a component from a devfile in my current directory, `odo create --devfile devfile.yaml` seems more intuitive than relying on iteractive mode with just `odo create`

GeekArthur:
> @GeekArthur One small modification:
> 
> If the user specifies the devfile in current directory with the devfile flag (e.g. `odo create --devfile devfile.yaml`), can we not error out? It might not be clear to the user that `odo create` needs to be used in that case, so it might be good to support both `odo create` and `odo create --devfile devfile.yaml`

Currently the error out means **odo throws errors and tells user to delete the existing devfile.yaml then run the same odo create command again**, overwrite prompting at that point is a good idea too

johnmcollier:
But if I'm creating a component from a devfile in the current directory (and I don't want to use interactive mode for whatever reason), then I should be able to just do `odo create --devfile devfile.yaml`. In that case, we're not going to be overwriting any devfile.

johnmcollier:
@GeekArthur What I mean is, if the devfile we pass in to `odo create --devfile` is the devfile in the current directory, then I don't think we need to error out, since we're not going to be overwriting any devfiles

GeekArthur:
> @GeekArthur What I mean is, if the devfile we pass in to `odo create --devfile` is the devfile in the current directory, then I don't think we need to error out, since we're not going to be overwriting any devfiles

+1. overwriting for that is not necessary, so the case would be when devfile exists in user's working directory
- `--devfile` points to the existing devfile in user's working directory -> successfully create component silently
- `--devfile` points to other devfile -> overwriting prompt

GeekArthur:
Based on the discussion above, feature design is finalized 🚀 
/remove-triage needs-information
/triage ready

pkg/util/util_test.go
```diff
@@ -1652,3 +1652,56 @@ func TestValidateURL(t *testing.T) {
...
+		wantErr bool
+	}{
+		{
+			name:    "Case 1: Copy successfully",
+			srcPath: srcPath,
+			dstPath: dstPath,
+			wantErr: false,
+		},
+		{
+			name:    "Case 2: Source doesn't exist",
```
prietyc123, line 1687: I think `Case 2: Invalid source path` justifies the case concern more. no?

pkg/util/util.go
```diff
@@ -1004,3 +1004,49 @@ func ValidateURL(sourceURL string) error {
...
+
+	if file.IsDir() {
+		return errors.Errorf("%s exists but it's not file", filePath)
+	}
+
+	return nil
+}
+
+// CopyFile copies file from source path to destination path
+func CopyFile(srcPath string, dstPath string) error {
```
prietyc123, line 1024: Duplicacy of the function `CopyFile`. We already have the same function in our code https://github.com/openshift/odo/blob/master/tests/helper/helper_filesystem.go#L144

pkg/util/util_test.go
```diff
@@ -1652,3 +1652,56 @@ func TestValidateURL(t *testing.T) {
...
+			srcPath: fakePath,
+			dstPath: dstPath,
+			wantErr: true,
```
prietyc123, line 1690: Can you please also specify a case where we are providing invalid destination path?

prietyc123:
@GeekArthur Pr needs rebase

pkg/util/util.go
prietyc123: It is just a file copy paste. So without duplicating It, we can use the same function in integration test code for sure. Can you please update the integration test code accordingly

pkg/util/util_test.go
GeekArthur: Done

pkg/util/util_test.go
GeekArthur: That could work, done.

pkg/util/util.go
GeekArthur: Done

GeekArthur:
@kadel @adisky @johnmcollier @dharmit @prietyc123 Comments are addressed, please review again when you get a chance. Also new PR description has been updated https://github.com/openshift/odo/pull/3118#issue-414863141.

pkg/odo/cli/component/create.go
```diff
@@ -98,6 +105,10 @@ By default, builder images (component type) will be used from the current namesp
 If version is not specified by default, latest will be chosen as the version.`)
 
 var createExample = ktemplates.Examples(`  # Create new Node.js component with the source in current directory.
+
+Note: When you use odo with experimental mode enabled and create devfile component, if you want to use existing devfile the first argument will be the component name
```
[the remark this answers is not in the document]
GeekArthur: Yes, I can add that.

pkg/odo/cli/component/create.go
```diff
@@ -308,15 +319,36 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		// Add a disclaimer that we are in *experimental mode*
 		log.Experimental("Experimental mode is enabled, use at your own risk")
 
-		if util.CheckPathExists(ConfigFilePath) {
+		if util.CheckPathExists(ConfigFilePath) || util.CheckPathExists(EnvFilePath) {
```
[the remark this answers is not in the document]
GeekArthur: My commit of handling the two edge cases above will include the fix for this as well.

pkg/odo/cli/component/create.go
```diff
@@ -308,15 +319,36 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+		if co.devfileMetadata.devfilePath != "" {
+			fileErr := util.ValidateFile(co.devfileMetadata.devfilePath)
+			urlErr := util.ValidateURL(co.devfileMetadata.devfilePath)
+			if fileErr != nil && urlErr != nil {
+				return errors.Errorf("The devfile path you specify is invalid with either file error \"%v\" or url error \"%v\"", fileErr, urlErr)
+			}
+		}
+
+		// Configure the context
+		if co.componentContext != "" {
```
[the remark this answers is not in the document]
GeekArthur: Yup, that's also my plan for the fix.

pkg/odo/cli/component/create.go
```diff
@@ -308,15 +319,36 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+			DevfilePath = filepath.Join(co.componentContext, devFile)
+			EnvFilePath = filepath.Join(co.componentContext, envFile)
+			co.CommonPushOptions.componentContext = co.componentContext
+		}
+
+		if util.CheckPathExists(DevfilePath) && co.devfileMetadata.devfilePath != "" && !util.PathEqual(DevfilePath, co.devfileMetadata.devfilePath) {
```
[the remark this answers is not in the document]
GeekArthur: Agree.

pkg/odo/cli/component/create.go
```diff
@@ -446,9 +474,10 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
-		registrySpinner := log.Spinnerf("Creating a devfile component from registry: %s", co.devfileMetadata.devfileRegistry.Name)
-
 		if co.devfileMetadata.devfileSupport {
+			registrySpinner := log.Spinnerf("Creating a devfile component from registry: %s", co.devfileMetadata.devfileRegistry.Name)
```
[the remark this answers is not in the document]
GeekArthur: Let me try to add more separate the code path and add validations accordingly.

pkg/odo/cli/component/create.go
```diff
@@ -803,25 +827,47 @@ func (co *CreateOptions) downloadProject(projectPassed string) error {
 // Run has the logic to perform the required actions as part of command
 func (co *CreateOptions) Run() (err error) {
 	if experimental.IsExperimentalModeEnabled() {
-		// Download devfile.yaml file and create env.yaml file
 		if co.devfileMetadata.devfileSupport {
+			// Use existing devfile directly from --devfile flag
+			if co.devfileMetadata.devfilePath != "" {
+				if util.ValidateURL(co.devfileMetadata.devfilePath) == nil {
```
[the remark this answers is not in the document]
GeekArthur: Create another field like `devfilePathType` would resolve that.

pkg/odo/cli/component/create.go
```diff
@@ -803,25 +827,47 @@ func (co *CreateOptions) downloadProject(projectPassed string) error {
...
+					}
+				}
+			}
+
 			if !util.CheckPathExists(DevfilePath) {
+				// Download devfile from registry
 				err := util.DownloadFile(co.devfileMetadata.devfileRegistry.URL+co.devfileMetadata.devfileLink, DevfilePath)
 				if err != nil {
-					return errors.Wrap(err, "Faile to download devfile.yaml for devfile component")
+					return errors.Wrap(err, "failed to download devfile for devfile component")
```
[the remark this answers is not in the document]
GeekArthur: That makes sense.

pkg/odo/cli/component/create.go
```diff
@@ -935,6 +981,7 @@ func NewCmdCreate(name, fullName string) *cobra.Command {
 		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.downloadSource, "downloadSource", "", "Download sample project from devfile.")
 		componentCreateCmd.Flags().Lookup("downloadSource").NoOptDefVal = defaultProjectName //Default value to pass to the flag if one is not specified.
 		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.devfileRegistry.Name, "registry", "", "Create devfile component from specific registry")
+		componentCreateCmd.Flags().StringVar(&co.devfileMetadata.devfilePath, "devfile", "", "Path to the user specify devfile")
```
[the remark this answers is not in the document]
GeekArthur: Yes, that would be handled by my commit of handling two edge cases above.

pkg/odo/cli/component/delete.go
```diff
@@ -57,7 +57,7 @@ func NewDeleteOptions() *DeleteOptions {
 
 // Complete completes log args
 func (do *DeleteOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	do.devfilePath = filepath.Join(do.componentContext, do.devfilePath)
+	do.devfilePath = filepath.Join(do.componentContext, DevfilePath)
```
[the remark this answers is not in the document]
GeekArthur: I think the change `do.devfilePath = filepath.Join(do.componentContext, DevfilePath)` already includes the context support, isn't it?

`filepath.Join(do.componentContext, DevfilePath)` == `The component context + DevfilePath` I think.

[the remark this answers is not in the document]
johnmcollier: @GeekArthur Yes haha, I can't believe I missed that 😆

pkg/odo/cli/url/create.go
```diff
@@ -92,6 +92,8 @@ func NewURLCreateOptions() *URLCreateOptions {
 
 // Complete completes URLCreateOptions after they've been Created
 func (o *URLCreateOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	o.DevfilePath = clicomponent.DevfilePath
```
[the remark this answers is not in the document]
GeekArthur: In `push.go` we use `po.DevfilePath = filepath.Join(po.componentContext, po.DevfilePath)` for context support, similar with `odo delete`, but I think `odo url` uses different way to do the context support as there is no `componentContext` in URLCreateOptions struct before and previously `odo url` just needs take the path from `--devfile`, the change in this PR is to let `o.DevfilePath` points to the global exported variable `DevfilePath` instead of the path from `--devfile`, and the context support should be handled implicitly in the following context functions.

I test the `odo url` context support with this PR, looks like it works for me and we don't need to explicitly handle the context support
```
odo create nodejs --context ./odo-test                                                                                                                                             ✔  17:11:42 

Validation
 ✓  Checking devfile compatibility [62731ns]
 ✓  Creating a devfile component from registry: Test [47635ns]
 ✓  Validating devfile component [269039ns]

    ~/wang  odo url create test --context ./odo-test                                                                                                                                           ✔  17:12:03 
 ✓  URL test created for component: nodejs

    ~/wang  cd odo-test                                                                                                                                                                        ✔  17:12:15 
    ~/wang/odo-test  cat ./.odo/env/env.yaml                                                                                                                                                   ✔  17:12:20 
ComponentSettings:
  Name: nodejs
  Namespace: default
  Url:
  - Name: test
    Port: 3000
    Kind: route
```

adisky:
/hold 
holding this PR for now, as it might create conflict with devfile V2 PR.

dharmit:
@johnmcollier @GeekArthur I'm removing lgtm to avoid retest triggers by the bot. The PR needs a rebase.

Issue #3015, opened 2020-04-27
Devfile Adapter Cleanup/Refactor


maysunfaisal:
/kind user-story

## User Story
Some cleanups to be tracked, that were raised during https://github.com/openshift/odo/pull/2922

## Acceptance Criteria 
- [x] Should clean up build command. Instead of appending, they should be separated out by commas https://github.com/openshift/odo/blob/master/pkg/exec/devfile.go#L20 (N/A - reason https://github.com/openshift/odo/issues/3015#issuecomment-649761051)
- [x] Refactor `docker/utils/utils.go StartBootstrapSupervisordInitContainer()` since it is very similar to `docker/component/utils.go startComponent()` https://github.com/openshift/odo/issues/3015#issuecomment-655701217
- [x] `sync/adapter.go SyncFiles()` has a lot of variables, see if they can be cleaned up further nicely with a struct https://github.com/openshift/odo/issues/3015#issuecomment-655718429
- [x] Refactor the project vol creation code, during component update if the project vol is not present, create it and sync the project https://github.com/openshift/odo/pull/3129
- [ ] Check if kube and docker adapter's `execDevfile()` can be refactored more cleanly
- [x] Put up a heading during initializating of supervisord if `--show-log` is absent, it takes some time to pull the docker image down, and the push output does not show anything at the time https://github.com/openshift/odo/issues/3033
- [x] During component update on kube, show the url link when mentioning it, currently just says url already exist (N/A - reason https://github.com/openshift/odo/issues/3015#issuecomment-649761051)
- [x] Update Supervisord volume name to be unique for each volume https://github.com/openshift/odo/pull/3120

## Links 
- Feature Request: 
- Related issue: 

/kind user-story

girishramnani:
I would suggest we do the cleanup once we have devfilev2

girishramnani:
we need to break this issue into two parts, devfile dependent and independent tasks.

maysunfaisal:
I went through the above checklist, and looks like none of the items are hard devfile dependent. The only possible items that MAYBE affected are:
`Refactor docker/utils/utils.go StartBootstrapSupervisordInitContainer() since it is very similar to docker/component/utils.go startComponent()`
` Check if kube and docker adapter's execDevfile() can be refactored more cleanly`

Would need to go through the individual devfile v2 feature request to see if they would be truly affected!

maysunfaisal:
Updates:
- not changing the url code from the checklist because it shares code path with odo v1
[...]

PR #3129, opened 2020-05-08
Fix Docker Supervisord Vol Initialization

### LINKED ISSUE #3034: [Docker] Hard to recover if supervisord volume is empty/corrupted
/kind bug
/area devfile
/area local

I've found that if the user exits odo while the supervisord image is being pulled, it can leave the user's system in a bad state that's hard to recover from:

```
Johns-MacBook-Pro-3:nodejs johncollier$ odo push

Validation
 ✓  Validating the devfile [55787ns]
^C
Johns-MacBook-Pro-3:nodejs johncollier$ odo push

Validation
 ✓  Validating the devfile [163221ns]

Creating Docker resources for component nodejs
 ✓  Pulling image quay.io/eclipse/che-nodejs10-ubi:nightly [2s]
 ✗  Starting container for quay.io/eclipse/che-nodejs10-ubi:nightly [505ms]
 ✗  Failed to start component with name nodejs.
Error: Failed to create the component: unable to create or update component: unable to pull and start container runtime for component nodejs: Unable to start container for devfile component runtime: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/opt/odo/bin/supervisord\": stat /opt/odo/bin/supervisord: no such file or directory": unknown
```

The volume exists, but doesn't have the supervisord binary on it. It's not really clear to the user _how_ to recover from this. For me, I had to do the following:
1) Restart docker
2) Run `docker system prune`
3) Delete the supervisord volume
4) Delete the supervisord image

(Alternatively a single docker restart + `docker system prune -a` would've worked too)

This happens because we create the supervisord volume before we pull the image. So, when `odo push` is run again, we check only for the existence of the volume, not that the volume actually contains the binary. I think we need to add a check that confirms that the supervisord volume has the binary on it, and if not, re-pull the image and re-create the volume.

maysunfaisal:

**What type of PR is this?**
/kind bug
/kind cleanup

Docker adapter uses one supervisord volume for all the odo components. However, if the supervisord volume is not initialized properly(for example, when someone presses Ctrl-C during odo push); future odo push is going to fail because supervisord may not have been copied to the volume

This PR:
- copies the odo-init image to the supervisord volume when component does not exist
- cleans up init container after vol has been initialized
- refactor component project vol logic
- spins up a supervisord vol per odo component instead of one for all odo components
- delete the supervisord vol on odo delete
- updated componentExist logic for docker adapter
- clean up and new unit tests

Fixes #3034 
Partially Addresses #3015

1. `odo push` and press Ctrl-C soon
2. Make sure there is a supervisord volume `docker volume ls`
3. `odo push` should succeed now

### DIFF
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -37,17 +37,21 @@ type Adapter struct {
 	devfileBuildCmd           string
 	devfileRunCmd             string
 	supervisordVolumeName     string
+	projectVolumeName         string
 }
 
 // Push updates the component if a matching component exists or creates one if it doesn't exist
 func (a Adapter) Push(parameters common.PushParameters) (err error) {
-	componentExists := utils.ComponentExists(a.Client, a.ComponentName)
+	componentExists, err := utils.ComponentExists(a.Client, a.Devfile.Data, a.ComponentName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
+	}
 
 	// Process the volumes defined in the devfile
 	a.componentAliasToVolumes = common.GetVolumes(a.Devfile)
 	a.uniqueStorage, a.volumeNameToDockerVolName, err = storage.ProcessVolumes(&a.Client, a.ComponentName, a.componentAliasToVolumes)
 	if err != nil {
-		return errors.Wrapf(err, "Unable to process volumes for component %s", a.ComponentName)
+		return errors.Wrapf(err, "unable to process volumes for component %s", a.ComponentName)
 	}
 
 	a.devfileInitCmd = parameters.DevfileInitCmd
@@ -64,19 +68,14 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	}
 	s.End(true)
 
-	// Get the supervisord volume
-	supervisordLabels := utils.GetSupervisordVolumeLabels()
-	supervisordVolumes, err := a.Client.GetVolumesByLabel(supervisordLabels)
+	a.supervisordVolumeName, err = a.createAndInitSupervisordVolumeIfReqd(componentExists)
 	if err != nil {
-		return errors.Wrapf(err, "unable to retrieve supervisord volume for component %s", a.ComponentName)
+		return errors.Wrapf(err, "unable to create supervisord volume for component %s", a.ComponentName)
 	}
-	if len(supervisordVolumes) == 0 {
-		a.supervisordVolumeName, err = utils.CreateAndInitSupervisordVolume(a.Client)
-		if err != nil {
-			return errors.Wrapf(err, "unable to create supervisord volume for component %s", a.ComponentName)
-		}
-	} else {
-		a.supervisordVolumeName = supervisordVolumes[0].Name
+
+	a.projectVolumeName, err = a.createProjectVolumeIfReqd()
+	if err != nil {
+		return errors.Wrapf(err, "unable to determine the project source volume for component %s", a.ComponentName)
 	}
 
 	if componentExists {
@@ -131,7 +130,8 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 
 // DoesComponentExist returns true if a component with the specified name exists, false otherwise
 func (a Adapter) DoesComponentExist(cmpName string) bool {
-	return utils.ComponentExists(a.Client, cmpName)
+	componentExists, _ := utils.ComponentExists(a.Client, a.Devfile.Data, cmpName)
+	return componentExists
 }
 
 // getFirstContainerWithSourceVolume returns the first container that set mountSources: true
@@ -199,10 +199,10 @@ func (a Adapter) Delete(labels map[string]string) error {
 
 			if snVal := vol.Labels["storage-name"]; len(strings.TrimSpace(snVal)) > 0 {
 				vols = append(vols, vol)
-			} else {
-				if typeVal := vol.Labels["type"]; typeVal == "projects" {
-					vols = append(vols, vol)
-				}
+			} else if typeVal := vol.Labels["type"]; typeVal == utils.ProjectsVolume {
+				vols = append(vols, vol)
+			} else if typeVal := vol.Labels["type"]; typeVal == utils.SupervisordVolume {
+				vols = append(vols, vol)
 			}
 		}
 	}
--- a/pkg/devfile/adapters/docker/component/utils.go
+++ b/pkg/devfile/adapters/docker/component/utils.go
@@ -519,39 +519,37 @@ func (a Adapter) startBootstrapSupervisordInitContainer(supervisordVolumeName st
 	return nil
 }
 
-// updateComponentWithSupervisord updates the devfile component's
+// UpdateComponentWithSupervisord updates the devfile component's
 // 1. command and args with supervisord, if absent
 // 2. env with ODO_COMMAND_RUN and ODO_COMMAND_RUN_WORKING_DIR, if absent
 func updateComponentWithSupervisord(comp *versionsCommon.DevfileComponent, runCommand versionsCommon.DevfileCommand, supervisordVolumeName string, hostConfig *container.HostConfig) {
 
 	// Mount the supervisord volume for the run command container
-	for _, action := range runCommand.Actions {
-		if *action.Component == *comp.Alias {
-			utils.AddVolumeToContainer(supervisordVolumeName, common.SupervisordMountPath, hostConfig)
+	if runCommand.Exec.Component == comp.Container.Name {
+		utils.AddVolumeToContainer(supervisordVolumeName, common.SupervisordMountPath, hostConfig)
 
-			if len(comp.Command) == 0 && len(comp.Args) == 0 {
-				klog.V(4).Infof("Updating container %v entrypoint with supervisord", *comp.Alias)
-				comp.Command = append(comp.Command, common.SupervisordBinaryPath)
-				comp.Args = append(comp.Args, "-c", common.SupervisordConfFile)
-			}
+		if len(comp.Container.Command) == 0 && len(comp.Container.Args) == 0 {
+			klog.V(4).Infof("Updating container %v entrypoint with supervisord", comp.Container.Name)
+			comp.Container.Command = append(comp.Container.Command, common.SupervisordBinaryPath)
+			comp.Container.Args = append(comp.Container.Args, "-c", common.SupervisordConfFile)
+		}
 
-			if !common.IsEnvPresent(comp.Env, common.EnvOdoCommandRun) {
-				envName := common.EnvOdoCommandRun
-				envValue := *action.Command
-				comp.Env = append(comp.Env, versionsCommon.DockerimageEnv{
-					Name:  &envName,
-					Value: &envValue,
-				})
-			}
+		if !common.IsEnvPresent(comp.Container.Env, common.EnvOdoCommandRun) {
+			envName := common.EnvOdoCommandRun
+			envValue := runCommand.Exec.CommandLine
+			comp.Container.Env = append(comp.Container.Env, versionsCommon.Env{
+				Name:  envName,
+				Value: envValue,
+			})
+		}
 
-			if !common.IsEnvPresent(comp.Env, common.EnvOdoCommandRunWorkingDir) && action.Workdir != nil {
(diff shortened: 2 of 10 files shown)


pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -25,36 +25,17 @@ import (
...
-		}
-		_, err := a.Client.CreateVolume(projectVolumeName, projectVolumeLabels)
-		if err != nil {
-			return errors.Wrapf(err, "Unable to create project source volume for component %s", componentName)
-		}
-	} else if len(projectVols) == 1 {
-		projectVolumeName = projectVols[0].Name
-	} else if len(projectVols) > 1 {
-		return errors.Wrapf(err, "Error, multiple source volumes found for component %s", componentName)
+		return errors.Wrapf(err, "Unable to determine the project source volume for component %s", componentName)
```
johnmcollier, line 38: We shouldn't capitalize the beginning of error messages

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -95,18 +76,10 @@ func (a Adapter) updateComponent() (componentExists bool, err error) {
...
-	projectVols, err := a.Client.GetVolumesByLabel(volumeLabels)
+	projectVolumeName, err := utils.CreateAndGetProjectVolume(a.Client, componentName)
 	if err != nil {
-		return componentExists, errors.Wrapf(err, "Unable to retrieve source volume for component "+componentName)
-	}
-	if len(projectVols) == 0 {
-		return componentExists, fmt.Errorf("Unable to find source volume for component %s", componentName)
-	} else if len(projectVols) > 1 {
-		return componentExists, errors.Wrapf(err, "Error, multiple source volumes found for component %s", componentName)
+		return componentExists, errors.Wrapf(err, "Unable to determine the project source volume for component %s", componentName)
```
johnmcollier, line 81: error message should be lower case

johnmcollier:
@maysunfaisal Do we need to copy the supervisord binary on every odo push, or just the odo pushes that require container start/restart?

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -61,19 +61,9 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	}
 	s.End(true)
 
-	// Get the supervisord volume
-	supervisordLabels := utils.GetSupervisordVolumeLabels()
-	supervisordVolumes, err := a.Client.GetVolumesByLabel(supervisordLabels)
+	a.supervisordVolumeName, err = utils.CreateAndInitSupervisordVolume(a.Client)
```
johnmcollier, line 64: Couple questions around always initializing the supervisord volume:
- Can we change it so that it's only on odo pushes that require a container start/restart? If the container is running and we just need to sync the code + exec some command, we shouldn't need to re-init the volume, no?
- Does running this on every odo push impact performance?

pkg/devfile/adapters/docker/utils/utils.go
```diff
@@ -233,6 +234,37 @@ func UpdateComponentWithSupervisord(comp *common.DevfileComponent, runCommand co
 	}
 }
 
+// CreateAndGetProjectVolume creates a project volume if absent and returns the
+// name of the created project volume
+func CreateAndGetProjectVolume(client lclient.Client, componentName string) (string, error) {
```
johnmcollier, line 239: Nit: Maybe we should change the name since we're not always creating a volume? `CreateProjectVolumeIfNeeded?` Or something similar

johnmcollier: Additionally, the only place where this function is called is in `docker/adapters/component/utils.go`, could we move this function there and also make it unexported?

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -25,36 +25,17 @@ import (
...
 func (a Adapter) createComponent() (err error) {
 	componentName := a.ComponentName
 
 	log.Infof("\nCreating Docker resources for component %s", a.ComponentName)
 
-	// Get or create the project source volume
-	var projectVolumeName string
-	projectVolumeLabels := utils.GetProjectVolumeLabels(componentName)
-	projectVols, err := a.Client.GetVolumesByLabel(projectVolumeLabels)
+	projectVolumeName, err := utils.CreateAndGetProjectVolume(a.Client, componentName)
```
johnmcollier, line 36: I'm wondering (and want your thoughts on this), if `CreateAndGetProjectVolume` returns an already existing volume (potentially with old/outdated/incomplete source code on it) during first-time push, should it overwrite the contents on the volume?

pkg/devfile/adapters/docker/utils/utils.go
```diff
@@ -246,14 +278,23 @@ func CreateAndInitSupervisordVolume(client lclient.Client) (string, error) {
...
+		return "", errors.Wrapf(err, "unable to retrieve supervisord volume for component")
+	}
+
+	if len(supervisordVolumes) == 0 {
+		_, err := client.CreateVolume(supervisordVolumeName, supervisordLabels)
+		if err != nil {
+			return "", errors.Wrapf(err, "unable to create supervisord volume for component")
+		}
+	} else {
+		supervisordVolumeName = supervisordVolumes[0].Name
```
johnmcollier, line 292: What happens if we overwrite the supervisord binary on the volume while another process is copying it to the container?

I guess, would it be better to have one supervisord volume per component, rather than for all of docker?

pkg/devfile/adapters/docker/utils/utils.go
```diff
@@ -297,13 +338,18 @@ func StartBootstrapSupervisordInitContainer(client lclient.Client, supervisordVo
...
-	err = client.StartContainer(&containerConfig, &hostConfig, nil)
+	containerID, err := client.StartContainer(&containerConfig, &hostConfig, nil)
 	if err != nil {
 		return err
 	}
 	if log.IsDebug() {
 		s.End(true)
 	}
 
+	err = client.RemoveContainer(containerID)
```
johnmcollier, line 349: What's the reasoning behind removing the supervisord container?

kadel:
Instead of initializing every time could we create some kind of indicator that would tell odo if the initialization was completed or not?
For example, we could just add a file to the volume that would contain the timestamp and odo version. This file would be created as the last step after everything has been copied to the volume.
Odo version can be later used to safely update the content of the supervisord volume if it changes in future odo versions.

pkg/devfile/adapters/docker/utils/utils.go
```diff
@@ -233,6 +234,37 @@ func UpdateComponentWithSupervisord(comp *common.DevfileComponent, runCommand co
...
+		projectVolumeName = projectVols[0].Name
+	} else if len(projectVols) > 1 {
+		return "", errors.New(fmt.Sprintf("multiple source volumes found for component %s", componentName))
+	}
+
+	return projectVolumeName, nil
+}
+
 // CreateAndInitSupervisordVolume creates the supervisord volume and initializes
 // it with supervisord bootstrap image - assembly files and supervisord binary
```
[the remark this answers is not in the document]
maysunfaisal: done

pkg/devfile/adapters/docker/component/utils.go
maysunfaisal: done

pkg/devfile/adapters/docker/component/utils.go
maysunfaisal: done

pkg/devfile/adapters/docker/utils/utils.go
maysunfaisal: do you reckon i should move the supervisord stuff and make it unexported too since its only being used in component utils? 🤔

pkg/devfile/adapters/docker/component/utils.go
maysunfaisal: so if its a first time push, i.e.; `componentExists` is false, we can force sync if a volume already exists. Yeah i like the idea, it sounds like we should do it

maysunfaisal: it looks like our current master code already covers this.

If its first time push, componentExists = false, we sync. Doesn't matter if we are reusing volume.

if its update component, componentExists = true, we sync 
- if any of the container from devfile component is missing
- if any of the container needs to be updated
- if containers does not require update, files changed, indexer handles it.

So looks like i might not need to do anything.

pkg/devfile/adapters/docker/utils/utils.go
maysunfaisal: Everytime we init, the container is left behind, thought its exited but you can see a bunch of init containers when you run `docker ps -a`

pkg/devfile/adapters/docker/utils/utils.go
maysunfaisal: I was trying to test this, I had the same concerns when writing this PR. I was trying to do simultaneous `odo push` at the same time. But honestly, I couldn't hit the issue.

But I guess we can move it a supervisord volume per odo component, to be safe!

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: Regd point 2, it takes only 1s when we init volume every time.

maysunfaisal:
@kadel we're creating the supervisord volume with supervisord version label.. so if a new version of supervisord was released, a new volume would be created. - https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/docker/utils/utils.go#L165

As for the timestamp idea, yeah i was thinking maybe we can create a flag in the volume that says if supervisord vol was init or not. But, it looks like if we run init supervisord volume everytime, it takes about 1s for the operation to complete. So I am debating if its worth doing it the other way since currently the supervisord init container logic is same for all odo v1, odo v2 kube and odo v2 docker ie; cp from odo-init image container to supervisord volume(in docker) or pvc(in kubernetes)

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: addressed in review commit

pkg/devfile/adapters/docker/utils/utils.go
maysunfaisal: addressed in review commit

maysunfaisal:
@johnmcollier @kadel PR is up for re-review, thx

pkg/devfile/adapters/docker/utils/utils.go
```diff
@@ -240,37 +252,82 @@ func UpdateComponentWithSupervisord(comp *common.DevfileComponent, runCommand co
...
-	s.End(true)
 
 	return supervisordVolumeName, nil
 }
 
 // StartBootstrapSupervisordInitContainer pulls the supervisord bootstrap image, mounts the supervisord
 // volume, starts the bootstrap container and initializes the supervisord volume via its entrypoint
-func StartBootstrapSupervisordInitContainer(client lclient.Client, supervisordVolumeName string) error {
-	supervisordLabels := GetSupervisordVolumeLabels()
+func StartBootstrapSupervisordInitContainer(client lclient.Client, componentName, supervisordVolumeName string) error {
```
johnmcollier, line 329: Function can be unexported

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -256,7 +219,7 @@ func (a Adapter) startComponent(mounts []mount.Mount, projectVolumeName string,
 	// Create the docker container
 	s := log.Spinner("Starting container for " + *comp.Image)
 	defer s.End(false)
-	err = a.Client.StartContainer(&containerConfig, &hostConfig, nil)
+	_, err = a.Client.StartContainer(&containerConfig, &hostConfig, nil)
```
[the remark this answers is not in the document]
maysunfaisal: Nope, we just return..

We only had to update this func for container id, so that we removed the supervisord init container after starting it, so that it doesn't clutter up `docker ps -a`

maysunfaisal:
@johnmcollier @kadel Hi 👋 , could you pls take a look at this PR and re-review? I've updated the PR and addressed all the review comments!

adisky:
removing the hold, as support for devfile V2 got merged
/hold cancel

adisky:
adding hold, to stop retest by bot
/hold

Issue #2959, opened 2020-04-22
Should use context to get appropriate envinfo instead of using os.Getwd()


yangcao77:
/kind bug

Since context is currently not supported in experimental mode, we use `os.Getwd()` to get the env.yaml
https://github.com/openshift/odo/blob/5a26a56426b5d4e214c0817c2820350d9b862e0a/pkg/odo/cli/component/devfile.go#L116-L117
https://github.com/openshift/odo/blob/5a26a56426b5d4e214c0817c2820350d9b862e0a/pkg/odo/cli/component/devfile.go#L130-L131
https://github.com/openshift/odo/blob/10826cc3f43c74ebadd917c3c53e9b40d6853aa6/pkg/devfile/adapters/docker/component/utils.go#L292

However, after context is supported, the project might be under a different directory; we need to switch to use context instead of using the current directory.

This issue is now blocked by experimental mode context support.

girishramnani:
this PR resolves the above issue https://github.com/openshift/odo/pull/3134

mohammedzee1000:
It looks to me like this problem was fixed already in all the places.
`os.getcwd()` is now only used when context="", but that should probably be handled while reading context, value. Will sent pr for that shortly

girishramnani:
`getComponentName` we need to remove this function

mohammedzee1000:
I figured this out while working on the now flag issue. Created local branch and have started looking at it

Issue #2835, opened 2020-04-07
Support listing devfile v2 components in odo


johnmcollier:
/kind user-story
/area devfile

## User Story
As a user, I want to be able to list the running devfile components on my system (Either Kube cluster or Docker) and see which devfile they belong to, using `odo list`.

## Acceptance Criteria 
- [x] `odo list` should be able to list all of the running devfile components in a given namespace
- [x] `odo list` should specify which devfile each component is part of (e.g. `openLiberty`, `java-springboot`, etc)
- [x] If running on an OpenShift cluster with both s2i and devfile components, it needs to support listing both

/kind user-story

kadel:
> * [ ]  `odo list` should be able to list all of the running devfile components in a given namespace
> *

Currently `odo list` filters the component based on the application (using `app` or `app.kubernetes.io/part-of` label).  Would we want to do the same?

Each component belongs to the application, and if you do `odo list` it lists components in the current application.

dharmit:
As for the possible state of the component, please refer https://github.com/openshift/odo/issues/2773#issuecomment-610922457.

adisky:
IMO, for devfiles `odo list` should list components from cluster only in the current namespace, `odo describe` should describe component in the current context folder from cluster/local config both. Keeping `odo list` to list down components from cluster would keep it simple and easy to understand.

girishramnani:
Do we need to implement the —all flag and —path flag to support compatibility for vscode?
That would definitely increase the complexity and we need to consider a situation where there are components in path which are deployed on kube using devfile and s2i on openshift.

adisky:
Not sure about the `--path` flag but the flag `--all-apps`  lists ` List all components from all applications for the current set project`.
The first question is do we want to have `app` concept for devfiles?

girishramnani:
we dont have the `app` concept for devfiles, that is one thing we need to address as part of this PR as well. might be a good idea to add that as part of the AC.

girishramnani:
- we would be able to fetch all the components by going over all the deployments ( maybe using a label selector ) searching for `component` label https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/kubernetes/component/adapter.go#L199, might also be a good idea to have `managed-by` label in place like we used to have for openshift components.
- change the odo list output to have `cluster type` as an extra field which could for now have `openshift`, `kubernetes` and `docker/local` value? or have a different table to each of these types like we have for operator services and catalog services.

adisky:
[...]

PR #3134, opened 2020-05-11
Listing created devfile component support for kubernetes and major cleanup

girishramnani:
**What type of PR is this?**

 /kind feature

Adding support for `odo list` showing created devfile components. 
Currently the scope of this PR is not to support `--path` and docker runtime. Support for those flags will be provided as a follow up PR.

This PR contains some major cleanup that 
- removes duplicate env.yaml reads
- optimisations like not listing the devfile catalog when there is a devfile.yaml already present
- redundant checks
- general cleanup

partly https://github.com/openshift/odo/issues/2835
fixes https://github.com/openshift/odo/issues/2959

Create some devfile components
```
odo list --project <project>
```
for all apps
```
odo list --project <project> --all-apps
```

Then create some s2i components, do
```
odo list --project <project>
```
confirm that both s2i and devfile components are displayed

### DIFF
--- a/pkg/odo/cli/component/list.go
+++ b/pkg/odo/cli/component/list.go
@@ -37,6 +37,7 @@ type ListOptions struct {
 	pathFlag         string
 	allAppsFlag      bool
 	componentContext string
+	IsOpenshift      bool
 	*genericclioptions.Context
 }
 
@@ -51,6 +52,8 @@ func (lo *ListOptions) Complete(name string, cmd *cobra.Command, args []string) 
 	if experimental.IsExperimentalModeEnabled() {
 		// Add a disclaimer that we are in *experimental mode*
 		log.Experimental("Experimental mode is enabled, use at your own risk")
+
+		lo.Context = genericclioptions.NewDevfileContext(cmd)
 
 	} else {
 		if util.CheckKubeConfigExist() {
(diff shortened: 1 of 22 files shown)


girishramnani:
will implement integration tests soon, but in the meanwhile if people can review the code that would be great.

girishramnani:
ok so now implemented integration tests

pkg/odo/cli/component/list.go
```diff
@@ -47,21 +56,45 @@ func NewListOptions() *ListOptions {
...
+			klog.V(4).Infof("New Config Context")
+			lo.Context = genericclioptions.NewConfigContext(cmd)
+
+		}
+	}
+
+	lo.Client = genericclioptions.Client(cmd)
+	dcSupport, err := lo.Client.IsDeploymentConfigSupported()
 
+	if err != nil {
```
adisky, line 79: i think it is safe to assume `dcSupport=false` here rather than returning error, because on a kubernetes cluster it might not return `404`.  Not very sure on this, but faced this problem in one of my pr and @mik-dass also ignored it here
https://github.com/openshift/odo/blob/master/pkg/component/component.go#L577

girishramnani: DC is not present on kubernetes for sure.

girishramnani: yeah, fair point though, this check shouldn't fail `odo`. and error would mean we just say `dcSupport` is not there.

adisky: Yup DC is not supported on k8s for sure, but using this (we cannot avoid this)`const ClusterVersionGroup = "apps.openshift.io"` for discovery client on kubernetes cluster, we wont't get `kerrors.IsNotFound(err)`, so if we return error here, it will always on k8s cluster. So IMO it is safe to assume `false` or  check for that error too here 
https://github.com/openshift/odo/pull/3134/files#diff-6025e9b620e9321c03c11bf29e5c4dbdR3287

dharmit:
@girishramnani can you please rebase this PR?

girishramnani:
will create a new PR as this is quite old

PR #3157, opened 2020-05-12
Did some cleanup in the devfile code

girishramnani:
**What type of PR is this?**
 /kind cleanup

Cleans up some devfile code

All tests should pass

### DIFF
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -162,10 +162,6 @@ func newProxyEnvInfo() proxyEnvInfo {
 func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}) (err error) {
 	if parameter, ok := asLocallySupportedParameter(parameter); ok {
 		switch parameter {
-		case "create":
-			createValue := value.(ComponentSettings)
-			esi.componentSettings.Name = createValue.Name
-			esi.componentSettings.Namespace = createValue.Namespace
 		case "url":
 			urlValue := value.(EnvInfoURL)
 			if esi.componentSettings.URL != nil {
@@ -296,10 +292,6 @@ func (ei *EnvInfo) GetApplication() string {
 }
 
 const (
-	// Create parameter
-	Create = "CREATE"
-	// CreateDescription is the description of Create parameter
-	CreateDescription = "Create parameter is the action to write devfile metadata to env.yaml"
 	// URL
 	URL = "URL"
 	// URLDescription is the description of URL
@@ -308,8 +300,7 @@ const (
 
 var (
 	supportedLocalParameterDescriptions = map[string]string{
-		Create: CreateDescription,
-		URL:    URLDescription,
+		URL: URLDescription,
 	}
 
 	lowerCaseLocalParameters = util.GetLowerCaseParameters(GetLocallySupportedParameters())
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -19,23 +19,12 @@ const (
 	envInfoFileName = "env.yaml"
 )
 
-// LocalConfigProvider is an interface which all local config providers need to implement
-// currently for openshift there is localConfigInfo and for devfile its EnvInfo.
-// The reason this interface is declared here instead of config package is because
-// some day local config would get deprecated and hence to keep the interfaces in the new package
-type LocalConfigProvider interface {
-	GetApplication() string
-}
-
 // ComponentSettings holds all component related information
 type ComponentSettings struct {
 	Name      string `yaml:"Name,omitempty"`
 	Namespace string `yaml:"Namespace,omitempty"`
 
-	// AppName is the application name. Application is a virtual concept present in odo used
-	// for grouping of components. A namespace can contain multiple applications
-	AppName string        `yaml:"AppName,omitempty" json:"AppName,omitempty"`
-	URL     *[]EnvInfoURL `yaml:"Url,omitempty"`
+	URL *[]EnvInfoURL `yaml:"Url,omitempty"`
 }
 
 // URLKind is an enum to indicate the type of the URL i.e ingress/route
@@ -170,8 +159,7 @@ func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}
 			} else {
 				esi.componentSettings.URL = &[]EnvInfoURL{urlValue}
 			}
-		case "appname":
-			esi.componentSettings.AppName = parameter
+
 		}
 
 		return esi.writeToFile()
@@ -285,11 +273,6 @@ func (ei *EnvInfo) GetName() string {
 // GetNamespace returns component namespace
 func (ei *EnvInfo) GetNamespace() string {
 	return ei.componentSettings.Namespace
-}
-
-// GetApplication returns the application name
-func (ei *EnvInfo) GetApplication() string {
-	return ei.componentSettings.AppName
 }
 
 const (
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -127,6 +127,7 @@ func newEnvSpecificInfo(envDir string, fs filesystem.Filesystem) (*EnvSpecificIn
 	// if the env.yaml file doesn't exist then we dont worry about it and return
 	if _, err = e.fs.Stat(envInfoFile); os.IsNotExist(err) {
 		e.envinfoFileExists = false
+
 		return &e, nil
 	}
 
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -21,9 +21,13 @@ const (
 
 // ComponentSettings holds all component related information
 type ComponentSettings struct {
-	Name      string        `yaml:"Name,omitempty"`
-	Namespace string        `yaml:"Namespace,omitempty"`
-	URL       *[]EnvInfoURL `yaml:"Url,omitempty"`
+	Name      string `yaml:"Name,omitempty"`
+	Namespace string `yaml:"Namespace,omitempty"`
+
+	// AppName is the application name. Application is a virtual concept present in odo used
+	// for grouping of components. A namespace can contain multiple applications
+	AppName string        `yaml:"AppName,omitempty" json:"AppName,omitempty"`
+	URL     *[]EnvInfoURL `yaml:"Url,omitempty"`
 }
 
 // URLKind is an enum to indicate the type of the URL i.e ingress/route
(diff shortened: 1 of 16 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -347,15 +352,13 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 
 			// devfile.yaml is not present, user has to specify the component type
 			// Component type: We provide supported devfile component list then let you choose
-			if !util.CheckPathExists(DevfilePath) {
```
girishramnani, line 350: this check is not needed because we are already checking it above.

pkg/odo/cli/component/create.go
```diff
@@ -412,18 +411,6 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		co.devfileMetadata.componentName = strings.ToLower(componentName)
 		co.devfileMetadata.componentNamespace = strings.ToLower(componentNamespace)
 
-		// If devfile.yaml is present, we don't need to download the devfile.yaml later
-		if util.CheckPathExists(DevfilePath) {
```
girishramnani, line 416: this check is not needed because we are already checking it above.

pkg/odo/cli/component/create.go
```diff
@@ -777,23 +769,22 @@ func (co *CreateOptions) downloadProject() error {
 // Run has the logic to perform the required actions as part of command
 func (co *CreateOptions) Run() (err error) {
 	if experimental.IsExperimentalModeEnabled() {
-		// Download devfile.yaml file and create env.yaml file
+		// Download devfile.yaml file and create env.yaml file only if its not already present
 		if co.devfileMetadata.devfileSupport {
-			if !util.CheckPathExists(DevfilePath) {
```
girishramnani, line 782: this check is not needed because we are already checking it above.

pkg/odo/cli/component/create.go
```diff
@@ -656,6 +643,11 @@ func (co *CreateOptions) Validate() (err error) {
 			spinner := log.Spinner("Validating devfile component")
 			defer spinner.End(false)
 
+			// This checks for the presence of s2i style component
+			if util.CheckPathExists(ConfigFilePath) {
+				return errors.New("This directory already contains a component")
```
johnmcollier, line 648: Should we clarify that this is an s2i style component rather than devfile component?

pkg/envinfo/envinfo.go
```diff
@@ -150,17 +152,14 @@ func newProxyEnvInfo() proxyEnvInfo {
 func (esi *EnvSpecificInfo) SetConfiguration(parameter string, value interface{}) (err error) {
 	if parameter, ok := asLocallySupportedParameter(parameter); ok {
 		switch parameter {
-		case "create":
```
johnmcollier, line 153: Why are we removing the switch the create here?

pkg/odo/genericclioptions/context.go
```diff
@@ -468,16 +468,18 @@ func newDevfileContext(command *cobra.Command) *Context {
 
 	envInfo, err := getValidEnvinfo(command)
 	if err != nil {
-		util.LogErrorAndExit(err, "")
+		util.LogErrorAndExit(err, "unable to retrieve configuration information")
```
johnmcollier, line 471: Should we clarify which configuration info we couldn't retrieve?

pkg/envinfo/envinfo.go
GeekArthur: @johnmcollier That's because he switches to use `co.EnvSpecificInfo.SetComponentSettings` to set up the parameters for `odo create` below instead of `co.EnvSpecificInfo.SetConfiguration`

GeekArthur: @girishramnani Can we switch to use `co.EnvSpecificInfo.SetComponentSettings` as well for `odo url` given the argument of that function is `ComponentSettings` struct which includes all metadata of `odo url`, so that we can make the parameter setup consistent rather than using separate function for parameter setup.

pkg/odo/cli/component/create.go
GeekArthur: Can we move `util.CheckPathExists(ConfigFilePath)` to the top with the one `util.CheckPathExists(EnvFilePath)` below. Then the code block would be:
```
if util.CheckPathExists(ConfigFilePath) || util.CheckPathExists(EnvFilePath) {
    return errors.New("This workspace directory already contains a devfile component")
}
```
It can validate both devfile and s2i components and we don't need to clarify.

```

Comment (GeekArthur) [reply pkg/odo/cli/component/create.go:648]: Actually I already did this cleanup in my PR https://github.com/openshift/odo/pull/3118/files as well, I can handle that if you don't mind.

File: pkg/odo/cli/component/create.go [hunk 7]
```diff
@@ -329,6 +326,14 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			co.CommonPushOptions.componentContext = co.componentContext
 		}
 
+		co.Context = genericclioptions.NewDevfileContext(cmd)
+
+		// One exception of a validation present in Complete code because, this is an optimisation
+		// i.e. if the devfile is present locally then we dont need to list the devfile catalog
+		if util.CheckPathExists(DevfilePath) {
```
GeekArthur, line 333: I don't think we can do the global check for devfile existence here, and we have to check separately in interactive mode and direct mode, reason is:
- Interactive mode: we can use devfile directly if devfile is present rather than error out, use can still specify the component name via interactive mode
- Direct mode: we need to error out as component type that user specify may have conflict with the existing devfile

pkg/odo/cli/component/create.go
GeekArthur: This is needed because:
- If devfile is present we **only** let user specify component name and namespace
- If devfile is not present, we let user specify component type, name and namespace
If we use the above checking, if devfile is present it will error our and return instead of falling down to interactive mode

pkg/odo/cli/component/create.go
GeekArthur: This is needed because:

I don't think we can do the global check for devfile existence here, and we have to check separately in interactive mode and direct mode, reason is:
- Interactive mode: we can use devfile directly if devfile is present rather than error out, use can still specify the component name via interactive mode
- Direct mode: we need to error out as component type that user specify may have conflict with the existing devfile

pkg/odo/cli/component/create.go
GeekArthur: This is needed as interactive mode uses devfile directly, remove it can make interactive mode download devfile from registry, which causes component type conflict issue.

GeekArthur:
@girishramnani Thanks for the cleanup PR, overall looks good, I leave some comments regarding logic issue for supporting both interactive mode and direct mode. Actually I feel like we may do many similar cleanup/improvements for `create.go` file, my PR is here: https://github.com/openshift/odo/pull/3118/files 😄

pkg/odo/cli/component/create.go
girishramnani: ok

pkg/odo/genericclioptions/context.go
girishramnani: we cannot as this error should be related to file read, so its all or nothing.

pkg/envinfo/envinfo.go
girishramnani: actually we have in the past preferred to be targeted, so if you are changing/adding a url then use the interface of url. so would it be ok if we keep that as is.

girishramnani:
@GeekArthur @johnmcollier please take a look

pkg/odo/cli/component/create.go
girishramnani: done

pkg/odo/cli/component/create.go
girishramnani: done

pkg/odo/cli/component/create.go
girishramnani: done

pkg/envinfo/envinfo.go
GeekArthur: I think all odo commands share the same mechanism (`envinfo.go`) to configure the `env.yaml` file so it would be better to centralize and make the mechanism generic, but it's fine, we can keep as it is for future url change.

GeekArthur:
Another suggestion is please do some smoke tests for the interactive mode as currently odo doesn't have interactive auto tests enabled, it's a test gap until we done this issue https://github.com/openshift/odo/issues/2194

girishramnani:
> Another suggestion is please do some smoke tests for the interactive mode as currently odo doesn't have interactive auto tests enabled for both s2i and devfile components, it's a test gap until we done this issue #2194

can we do that as part of a different issue?

GeekArthur:
> > Another suggestion is please do some smoke tests for the interactive mode as currently odo doesn't have interactive auto tests enabled for both s2i and devfile components, it's a test gap until we done this issue #2194
> 
> can we do that as part of a different issue?

For the interactive mode test automation we definitely need another PR to cover that, but we still need to do smoke tests **manually** only against this PR for interactive mode to avoid regression, so just do some manual smokes to see if interactive mode still works would be fine. 😄

girishramnani:
same test failing on both
```
should successfully create the devfile component [It]  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_create_test.go:76
```
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3157/pull-ci-openshift-odo-master-v4.2-integration-e2e/637#1:build-log.txt%3A558

https://travis-ci.com/github/openshift/odo/jobs/333571414#L531

@amitkrout is this something we are aware of? or do you feel it was introduced in this PR?

amitkrout:
> same test failing on both
> 
> ```
> should successfully create the devfile component [It]  /home/travis/gopath/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_create_test.go:76
> ```
> 
> https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3157/pull-ci-openshift-odo-master-v4.2-integration-e2e/637#1:build-log.txt%3A558
> 
> https://travis-ci.com/github/openshift/odo/jobs/333571414#L531
> 
> @amitkrout is this something we are aware of? or do you feel it was introduced in this PR?

@girishramnani i tried the same test locally on master by increasing test node by 4 just to see the namespace collision but none of the time it occurred.  Even before none of odo open prs hit this failure. This is freshly introduce may be due to some issue with the code cleanup itself.

girishramnani:
will open a fresh one

Issue #3076, opened 2020-05-05
Conflicts in odo and Devfile terminology 


kadel:
With moving everything towards Devfiles some of the odo terminology clashes with Devfile terminology.

One of the biggest clashes is the use of the term **component**.
 -  in odo "component" is part of the application (for example one microservice that is part of a bigger application) 
 - in Devfile "component" is part of the workspace (for example a build container or an IDE plugin)

This should be resolved before we release odo v2 with stable Devfile support. 

We have two options:
 1) Change on odo side. This will include big changes in most of the odo commands, making it backward incompatible.
2) Rename components in devfile spec to something else

/kind discussion
/kind epic
/priority high

dharmit:
> in odo "component" is part of the application (for example one microservice that is part of a bigger application)

IMO, if possible, we could use this opportunity to change it on odo side. Personally, in my head, even though I have an idea about what a component (DC+pods+service+url) means in odo, reading its definition still leaves me confused. :confused: 

Plus, application means yet another thing in OpenShift (a label that can be assigned to different objects, IIUC) and here in the definition you mention, it means something else.

kadel:
This might be a good opportunity to even more match odo terminology with OpenShift web console.
One big downside is that this will make odo v2 completely incompatible with the previous version. Which will make it more complicated to adopt odo v2 for IDE plugins

dharmit:
By matching the terminology with OpenShift web console, we don't mean to use the lingo of OpenShift, right? I think a major goal of odo is to abstract k8s/OpenShift for the users so that they don't have to understand those. So, using the web console terminology is fine as long as we don't make things confusing for users.

I think this is a great topic for odo contributor's call.

kadel:
> By matching the terminology with OpenShift web console, we don't mean to use the lingo of OpenShift, right? I think a major goal of odo is to abstract k8s/OpenShift for the users so that they don't have to understand those. So, using the web console terminology is fine as long as we don't make things confusing for users.
> 
> I think this is a great topic for odo contributor's call.

Sorry, I should have been more specific. I meant one specific part of the OpenSHift web console - Topology view in Developer console

gorkem:
I think, the above odo definition of component is a subset of devfile definition and `odo component` command is not able to create everything in the the full definition on the devfile. I do not think `odo component` command should actually be used to create builders and side-car tools that are useful for development time only either. I am OK if it only deals with a subset of devfile and we let "experts" to modify additional components in devfile only

[...]

Issue #2756, opened 2020-03-25
Add `odo deploy` command for initial transition from inner to outer loop


bbrowning:
/kind feature

## Which functionality do you think we should add?

Today odo helps a developer with their inner loop workflow. Once the developer is ready to deploy that application, they need to go outside of odo to make that happen. I propose odo have an `odo deploy` command to allow an initial transition from inner loop to outer loop.

The result of an `odo deploy` command should be the component gets deployed to the cluster in a configuration suitable to receive production traffic.

## Why is this needed?

This is meant primarily for new user experience, demos, and such as a developer learns how to deploy applications to the cluster. Best practice for real production usage would still be integration with pipelines (#1238) or integration with gitops workflows (#2617). There's a gap between `odo push` for in-cluster inner loop and pipelines or gitops workflows that this intends to fill.

## How should it be added?

As a strawman proposal, I'd suggest bringing this in following the recent devfile support by having a convention for metadata in the devfile that odo can understand to deploy an application. This would allow devfile-aware pipelines to reuse that same devfile metadata to do the actual deployment at the output of those pipelines.

What the exact metadata in the devfile should look like needs to be decided. I believe the existing devfile 2.0 spec covers the need here, but we'll need to verify that.

## User stories

- [x]  #5109
- [x]  #5110

sspeiche:
This is the "same" related to this discussion https://github.com/openshift/odo/issues/1229

girishramnani:
We are planning a proposal issue which would break this epic down https://github.com/openshift/odo/issues/2703

lance:
In addition to pipelines, I would like to see another devfile command supported by odo. Currently, if I understand correctly, odo supports two commands in the devfile: `devBuild` and `devRun`. Could there be a supported command that will deploy your image without the supervisor and possibly using a different image component?

kadel:
> Could there be a supported command that will deploy your image without the supervisor and possibly using a different image component?

Yes, this is what the deploy command should do. 
"deployed to the cluster in a configuration suitable to receive production traffic." means that the supervisor won't be present and different image can be optionally used (for example this image should be completely stripped form any development tools, and should be as small as possible).

Not sure how we exactly we will achieve this but defining another command like `prodRun` or `prodBuild` could be one way of doing this.

serenamarie125:
Moved out to 2.2 milestone due to dependencies on devfile 2.2

dharmit:
[...]

PR #3177, opened 2020-05-13
Design proposal: Event notification support for build and application status for IDE integration for devfile scenarios (#2550)

jgwest:
**What type of PR is this?**
/kind documentation
[skip ci]

Parent issue: https://github.com/openshift/odo/issues/2550

**Proposal as rendered by GitHub markdown**:
https://github.com/jgwest/odo/blob/doc-2550-proposal/docs/proposals/event-notification.md

### DIFF
--- a/docs/proposals/event-notification.md
+++ b/docs/proposals/event-notification.md
@@ -148,11 +148,11 @@ Fundamentally, this proposal needs to find a solution to this scenario:
 
 1) IDE creates a URL (calls `odo url create`) and pushes the user's code (calls `odo push`)
 2) To get the status of that app, the IDE runs `odo component status -o json` to start the long-running odo process. The status command then helpfully reports the pod/url/supervisord container status, which allows the IDE to determine when the app process is up.
-3) [some time passes]
-4) IDE creates a new URL (or performs some other action that invalidates the existing `odo status` stte, such as `odo component delete`) by calling `odo url create`.
-5) The long-running `odo status` process is still running, but somehow needs to know about the new URL from step 4 (or other events)
+3) *[some time passes]*
+4) IDE creates a new URL (or performs some other action that invalidates the existing `odo status` state, such as `odo component delete`) by calling `odo url create`.
+5) The long-running `odo status` process is still running, but somehow needs to know about the new URL from step 4 (or other events).
 
-Thus, in some way that existing long-running `odo status` process needs to be informed of the new event (a new url event, a component deletion event, etc). Since these events are generated across independent OS processes, this requires some form of [IPC](https://en.wikipedia.org/wiki/Inter-process_communication). 
+Thus, in some way, that existing long-running `odo status` process needs to be informed of the new event (a new url event, a component deletion event, etc). Since these events are generated across independent OS processes, this requires some form of [IPC](https://en.wikipedia.org/wiki/Inter-process_communication). 
 
 ### Some options in how to communicate these data across independent odo processes (in ascending order of complexity)
 
@@ -162,16 +162,17 @@ This is the solution I propose in this proposal, and is included for contrast.
 
 Since the IDE has a lifecycle that is greater than each of the individual calls to `odo`, and the IDE is directly and solely responsible for calling odo (when the user is interacting with the IDE), it is a good fit to ensure the state of `odo component status` is up-to-date and consistent.
 
-But option this is by no means a perfect solution: 
-- this does introduce complexity on the IDE side, as the IDE needs to keep track of which `odo` processes are running for each component, and it needs to know when/how to respond to actions (delete/url create/etc). But since the IDE is a monolithic process, this is at least straightforward (I mocked up the algorithm that the IDE will use in each case, which I can share if useful.)
-- this introduces complexity for EVERY new IDE/consuming tool that uses this mechanism; rather than solving it once in ODO, it needs to be solved X times for X IDEs.
+But this option is by no means a perfect solution: 
+- This does introduce complexity on the IDE side, as the IDE needs to keep track of which `odo` processes are running for each component, and it needs to know when/how to respond to actions (delete/url create/etc). But since the IDE is a monolithic process, this is at least straightforward (I mocked up the algorithm that the IDE will use in each case, which I can share if useful.)
+- This introduces complexity for EVERY new IDE/consuming tool that uses this mechanism; rather than solving it once in ODO, it needs to be solved X times for X IDEs.
+- Requires multiple concurrent long-running odo processes per odo-managed component
 
 #### 2) `odo component status` could monitor files under the `.odo` directory in order to detect changes; for example, if a new URL is added to `.odo/env/env.yaml`, `odo component status` would detect that and update the URLs it is checking
 
 This sounds simple, but is surprisingly difficult:
-- No way to detect a delete operation just by watching `.odo` directory: At present, `odo delete` does not delete/change any of the files under `.odo`
-- Partial file writes/atomicity/file locking: How to ensure that when `odo component status` reads a file that it has been fully written by the consuming process? One way is to use file locks, but that  means using using/testing each supported platform's file locking mechanisms. Then need to implement a cross-process polling mechanisms.
-- Or, need to implement a cross-platform filewatching mechanism: We need a way to watch the `.odo` directory and respond to I/O events to the files, either by modification. 
+- No way to detect a delete operation just by watching `.odo` directory: at present, `odo delete` does not delete/change any of the files under `.odo`
+- Partial file writes/atomicity/file locking: How to ensure that when `odo component status` reads a file that it has been fully written by the producing process? One way is to use file locks, but that means using/testing each supported platform's file locking mechanisms. Then need to implement a cross-process polling mechanism.
+- Or, need to implement a cross-platform [filewatching mechanism](https://github.com/fsnotify/fsnotify): We need a way to watch the `.odo` directory and respond to I/O events to the files, either by modification. 
 - Windows: Unlike other supported platforms, Windows has a number of quirky file-system behaviours that need to be individually handled. The most relevant one here is that Windows will not let you delete/modify a file in one process if another process is holding it open (we have been bitten by this a number of times in Codewind)
 - Need to support all filesystems: some filesystems have different file write/locking/atomicity guarantees for various operations.
 
@@ -189,7 +190,7 @@ As one example of this, we could create a new odo process/command (`odo status -
 
 
 Drawbacks:
-- Odo's code currently assuming that commands are a short-lived, mostly single-threaded, and compartmentalized; switching to a server would fundamentally alter this presumption of existing code
+- Odo's code currently assumes that commands are short-lived, mostly single-threaded, and compartmentalized; switching to a server would fundamentally alter this presumption of existing code
 - Much more complex to implement versus other options: requires changing the architecture of the odo tool into a multithreaded client-server model, meaning many more moving parts, and [the perils of distributed computing](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing).
 - Most be cross-platform; IPC mechanisms/behaviour are VERY platform-specific, so we probably need to use TCP/IP sockets. 
 - But, if using HTTP/S over TCP/IP socket, we need to secure endpoints; just listening on localhost [is not necessarily enough to ensure local-only access](https://bugs.chromium.org/p/project-zero/issues/detail?id=1524).
@@ -199,7 +200,7 @@ Variants on this idea: 1) a new odo daemon/LSP-style server process that was res
 
 ### Proposed option vs options 2/3
 
-Hopefully the inherent complexity in options 2-3 is clear; I can further flesh out these options if not. Likewise, if you all have another fourth option, I'm all ears.
+Hopefully I have sucessfully conveyed the inherent complexity of options 2-3... if not I can further flesh out these options. Likewise, if you all have another fourth option, I'm all ears.
 
 In any case, we certainly COULD go with any of these more sophisticated solutions: there are no technical impediments to doing so, and speaking personally it would be interesting/fun to implement them :). 
 
--- a/docs/proposals/event-notification.md
+++ b/docs/proposals/event-notification.md
@@ -6,7 +6,7 @@ With this proposal and [it's related issue](https://github.com/openshift/odo/iss
 
 In short, the proposal is:
 - New flag for push command, `odo push -o json`: Outputs JSON events that correspond to what devfile actions/commands are being executed by push.
-- New odo command `odo component status -o json`: Sends an HTTP/S request to the application URL, and checks the container/pod status every X seconds (and/or a watch, for Kubernetes case). The result of both is output as JSON to be used by external tools to determine if the application is running.
+- New odo command `odo component status -o json`: Calls `supervisord ctl status` to check the running container processes, checks the container/pod status every X seconds (and/or a watch, for Kubernetes case), and sends an HTTP/S request to the application URL. The result of those is output as JSON to be used by external tools to determine if the application is running.
 - A standardized markup format for communicating detailed build status, something like: `#devfile-status# {"buildStatus":"Compiling application"}`, to be optionally included in devfile scripts that wish to provide detailed build status to consuming tools.
 
 ## Terminology
@@ -17,7 +17,7 @@ With this issue, we are looking at adding odo support for allowing external tool
 
 **Detailed build status**: An indication of which step in the build process the build is in. For example: are we 'Compiling application', or are we 'Running unit tests'?
 
-**App status**: Determine if an application is running, using container status (for both local and Kubernetes, various status: containercreating, started, restarting, etc), or whether an HTTP/S request to the root application URL returns an HTTP response with any status code.
+**App status**: Determine if an application is running, using container status (for both local and Kubernetes, various status: containercreating, started, restarting, etc), `supervisord ctl status`, or whether an HTTP/S request to the root application URL returns an HTTP response with any status code.
 
 **Detailed application status**: 
 - While app status works well for standalone application frameworks (Go, Node Express, Spring Boot), it works less well for full server runtimes such as Java EE application servers like OpenLiberty/WildFly that may begin responding to Web requests before the user's deployed WAR/EAR application has finished startup. 
@@ -25,7 +25,7 @@ With this issue, we are looking at adding odo support for allowing external tool
 - Fortunately, in these cases the IDE/consuming tool can use the console logs (from `odo log`) from the runtime container to determine a more detailed application status.
 - For example, OpenLiberty (as an example of an application-server-style container) prints a specific code when an application is starting  `CWWKZ0018I: Starting application {0}.`, and another when it has started. `CWWKZ0001I: Application {0} started in {1} seconds.`
 - Odo itself should NOT know anything about these specific application codes; knowing how these translate into a detailed application status would be the responsibility of the IDE/consuming tool. Odo's role here is only to provide the console output log. 
-- In the future, we could add these codes into the devfile to give Odo itself some agency over determining the detailed application status, but for this proposal I am leaving this responsibility with the consuming tool.
+- In the future, we could add these codes into the devfile to give Odo itself some agency over determining the detailed application status, but for this proposal responsibility is left with the consuming tool.
 
 **Devfile writer**: A devfile writer may be a runtime developer (for example, a Red-Hatter working on WildFly or an IBMer working on OpenLibery) creating a devfile for their organization's runtime (for example 'OpenLiberty w/ Maven' dev file), or an application developer creating/customizing a dev file for use with their own application. In either case, the devfile writer must be familiar with the semantics of both odo and the devfile.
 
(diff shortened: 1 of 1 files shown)


dharmit:
> /ok-to-test

Why are we explicitly testing a docs PR? :(

amitkrout:
> > /ok-to-test
> 
> Why are we explicitly testing a docs PR? :(

The pr for skipping the test on OpenShift CI is not in yet. Need to create a pr in release repo. I will do that. Till then please have patience ;)

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,141 @@
+# ODO build and application status notification 
+
+## Summary
+
+With this proposal and [it's related issue](https://github.com/openshift/odo/issues/2550) we examine how to consume build/run event output from odo in order to allow external tools (such as IDEs) to determine the application/build status of an odo-managed application.
+
+In short, the proposal is:
+- New flag for push command, `odo push -o json`: Outputs JSON events that correspond to what devfile actions/commands are being executed by push.
+- New odo command `odo component status -o json`: Pings the application URLs, and checks the container/pod status every X seconds (and/or a watch, for Kubernetes case). The result of both is output as JSON to be used by external tools to determine if the application is running.
```
kadel, line 9: What if application doesn't have any URLs?
Also, what would be considered as pinging url? Just checking if the port is open or sending HTTP commands like GET or HEAD?

Wouldn't it be enough to just check if the application process is running?
That could be done easily, as it will run as a supervisord service. There could be additional information in the status message like if the URL as alive or not (using HTTP HEAD), but the primary source for the status would be if the process  is running or not.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,141 @@
...
+- A standardized markup format for communicating detailed build status, something like: `#devfile-status# {"buildStatus":"Compiling application"}`, to be optionally included in devfile scripts that wish to provide detailed build status to consuming tools.
```
kadel, line 10: 👍 simple and powerful solution. I like it

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Terminology
+
+With this issue, we are looking at adding odo support for allowing external tools to gather build status and application status. We further divide both statuses into detailed and non-detailed, with detailed principally being related to statuses that can be determine by looking at container logs.
+
+**Build status**: A simple build status indicating whether the build is running (y/n), and whether the last build succeeded or failed. This can be determined based on whether odo is running a dev file action/command, and the process error code (0 = success, non-0 = failed) of that action/command.
+
+**Detailed build status**: An indication of which step in the build process the build is in. For example: are we 'Compiling application', or are we 'Running unit tests'?
+
+**App status**: Determine if an application is running, using container status (for both local and Kubernetes, various status: containercreating, started, restarting, etc), or whether an HTTP/S request to the root application URL returns an HTTP response with any status code.
```
kadel, line 20: Due to the use of the SupervisorD the application might not be running or falling and the container would be still running. It needs to check the application process to determine if the application is running or not.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Detailed application status**: 
+- While app status works well for standalone application frameworks (Go, Node Express, Spring Boot), it works less well for full server runtimes such as Java EE application servers like OpenLiberty/WildFly that may begin responding to Web requests before the user's deployed WAR/EAR application has finished startup. 
+- Since these application servers are built to serve multiple concurrently-deployed applications, it is more difficult to determine the status of any specific application running on them. The lifecycle of the application server differs from the lifecycle of the application running inside the application server. 
+- Fortunately, in these cases we can use the console logs (from `odo log`) from the runtime container to determine a more detailed application status.
+- For example, OpenLiberty (as an example of an application-server-style container) prints a specific code when an application is starting  `CWWKZ0018I: Starting application {0}.`, and another when it has started. `CWWKZ0001I: Application {0} started in {1} seconds.`
```
kadel, line 26: How would odo know what log line means what? For example that `CWWKZ0018I: Starting application {0}.` means that the application is starting initialization. This is framework-specific knowledge and as such, it should be somewhere in devfile or in a Stack but not inside odo.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,141 @@
...
+(Exact details for which fields are included with events are TBD)
+
+In addition, `odo push -o json` should return a non-zero error code if one of the actions returned a non-zero error code, otherwise zero is returned.
+
+### `odo push -o json` example output
+
+This is what an `odo push -o json` command invocation would look like:
+```
+odo push -o json
+{ "devFileCommandExecutionBegun": { "commandName" : "build", "timestamp" : "(UTC unix epoch seconds.microseconds)" } }
```
kadel, line 58: Each line has a different JSON structure (different keys), this might make parsing in some languages (like Go) a little bit trickier. Not a huge problem, but just something to be aware of.

docs/proposals/event-notification.md
jgwest: Agreed: the use of container status here is an imperfect representation of the app status for that exact reason. It's still a useful data point (eg if the container ISN'T running. then the app is obviously not running), but if the container IS running it doesn't fully/necessarily imply the app is running (due to SupervisorD).

I like the idea of checking the application process itself, let me check the feasibility of that...

docs/proposals/event-notification.md
jgwest: Yep, we definitely don't want ODO to know anything about these codes/strings, it looks like the language I used in the proposal is imprecise re: who would use these data. 

I have updated the proposal to clarify that it is the consuming tool's (for example, IDE) responsibility to know what these codes mean.. I added these lines:
> Odo itself should NOT know anything about these specific application codes; knowing how these translate into a detailed application status would be the responsibility of the IDE/consuming tool. Odo's role here is only to provide the console output log. 
>
> In the future, we could add these codes into the devfile to give Odo itself some agency over determining the detailed application status, but for this initial proposal I am leaving this responsibility with the consuming tool.
```

File: docs/proposals/event-notification.md (hunk 5 above)
Comment (jgwest) [reply docs/proposals/event-notification.md:60]: This is true; here's my draft PR of part of this feature (the PR has gone through zero review, so is subject to change) that contains the go structs used to marshal/unmarshall these entries: https://github.com/openshift/odo/blob/e8040649268c0e7624c5662e364fdb8607837b41/pkg/machineoutput/types_event_logging.go#L24

File: docs/proposals/event-notification.md (hunk 1 above)
Comment (jgwest) [reply docs/proposals/event-notification.md:9]: Good questions:

If the user is developing a web application, often the first metric of "is my application running?" is "can I see it in my browser when I refresh?". I attempt to capture this idea with this part of the proposal.

The algorithm I had considered would be:
- For each odo-managed url:
  - issue a GET/HEAD request to the app domain root, for example a component 'my-app' might be pinged at https://my-app.my-cluster-domain.com:8080
  - `status` command to output a success to the console if ANY HTTP response (2xx/3xx/4xx/5xx) code is received, since what we are testing is whether the app is connectable (a failure would be no response before timeout).

Now this raises a few good question: 
- Is it always safe (eg [safe](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Safe_methods), non-destructive, immutable, idempotent) to issue GET/HEAD requests to the root of an any odo-managed application without the user's explicit acknowledgement? I had assumed yes (eg low risk), but am definitely open to further discussion here. 
- Will Kubernetes ingress controllers ALWAYS respond with an HTTP response, even if sent to an invalid cluster URL? If so, this would make this check not especially useful for the Kubernetes case...
- Could we just remove this entire algorithm from the proposal and leave it up to the IDE to ping the URL if appropriate? After all, the IDE does have full knowledge over which URLs are exposed via odo... at the expense of slightly more work on the IDE-side.

As for the case where no application URLs are exposed (and there are many such applications): if no odo-managed URLs are defined, the algorithm will be a no-op, and no URL success/failure output will be output.

RE: checking if the application process is running, I like the idea and is definitely preferable if feasible, I will expand on this further [on the other comment where you mentioned this](https://github.com/openshift/odo/pull/3177#discussion_r425430516).

File: docs/proposals/event-notification.md (hunk 4 above)
Comment (kadel) [reply docs/proposals/event-notification.md:26]: > Yah, we definitely don't want odo to know anything about these codes/strings; it looks like the language I used in the proposal is imprecise re: who would use these data.
> 
> I have updated the proposal to clarify that it is the _consuming tool_'s responsibility (for example, IDE) to know what these codes mean.. I added these lines:

Maybe it was just my bad implementation of this. What you added makes this perfectly clear, thx.

File: docs/proposals/event-notification.md [hunk 6]
```diff
@@ -0,0 +1,143 @@
...
+{ "devFileActionExecutionBegun" : { "commandName" : "build", "actionName" : "run-build-script", "timestamp" : "(...)" } }
+{ "logText" : { "text:" "one line of text received\\n another line of text received", "timestamp" : "(...)" } }
+{ "devFileActionExecutionComplete" : { "errorCode" : 1, ( same as above )} }
+{ "logText" : { "text": (... ), "timestamp" : "(...)" } } # Log text is interleaved with events 
+{ "devFileCommandExecutionComplete": { "success" : false, (same as above) } }
+
+(Exact details on event name, and JSON format are TBD; feedback welcome!)
+```
+
+These events allow an external odo-consuming tool to determine the build status of an application (build succeeded, build failed, build not running).
```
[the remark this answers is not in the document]
jgwest: Since these are just JSON-formatted strings that are  output to stdout, the consuming tool is free to ignore whichever events it wishes (if event == 'type I don't want' then skip). I don't think there is value add to allowing any additional customization of which events are output by odo.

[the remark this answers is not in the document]
dharmit: Makes sense :+1:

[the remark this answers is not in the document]
jgwest: > Thinking out loud: can we make use of k8s health checks for this and/or other status info that we intend to come up with?

K8s health checks appear not to be defined in either odo devfile v1 or v2, so this would be something to consider if/when these are added there. (And these would not apply if odo was running with docker target)

docs/proposals/event-notification.md
jgwest: K8s health checks appear not to be defined in either odo devfile v1 or v2 (correct me if I'm wrong, I didn't see it), so this would be something to consider if/when these are added there. (And these would not apply if odo was running with docker target)

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+Ideally, we would like for odo to provide consuming tools with all 3 sets of data. Thus, as proposed:
+- 1 and 2 are handled by a new `odo component status -o json` command, described here.
+- 3 is handled by the existing unmodified `odo log --follow` command.
+
+The new proposed `odo component status -o json` command will:
+- Be a *long-running* command that will continue outputing status until it is aborted by the parent process.
+- Every X seconds, ping the URLs/routes of the application as they existed when the command was first executed. Output the result as a JSON string.
+- Every X seconds (or using a Kubernetes watch, where appropriate), check the container status for the application, based on the application data that was present when the command was first issued. Output the result as a JSON string.
+
+**Note**: This command will NOT, by design, respond to route/application changes that occur during or after it is first invoked. It is up to consuming tools to ensure that the `odo component status` command is stopped/restarted as needed.
```
dharmit, line 110: If we do not design it to respond to changes that occur in route/app since first run of `odo component status`, this command would report failure in accessing the route/app, right? For example:
- I pushed the component and executed `odo component status -o json`.
- Now, 
  - I deleted the URL and created another one instead, or
  - I change the port (less likely than URL deletion+creation scenario), or
  - I feel like adding another endpoint to the application `/status` that displays the status of my app.
- If the user has setup some action to be performed in case of an URL/app being inaccessible, wouldn't this give a false positive?

This might be irrelevant right now, but it might become more relevant once #2756 starts moving ahead.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+
+**Note**: This command will NOT, by design, respond to route/application changes that occur during or after it is first invoked. It is up to consuming tools to ensure that the `odo component status` command is stopped/restarted as needed. 
+  - For example, if the user tells the IDE to delete their application with the IDE UI, the IDE will call `odo delete (component)`; at the same time, the IDE should also abort any existing `odo component status` commands that are running (as these are no longer guaranteed to return a valid status now that the application itself no longer exists). `odo component status` will not automatically abort when the application is deleted (because it has no reliable way to detect this in all cases).
+  - Another example: if the IDE adds a new URL via `odo url create [...]`, any existing `odo component status` commands that are running should be aborted, as these commands would still only be checking the URLs that existed when the command was first invoked (eg there is intentionally no cross-process notification mechanism for created/updated/deleted URLs implemented as part of this command.)
```
dharmit, line 112: This answers my previous long comment about a hypothetical scenario about change in URL/app to a good extent. However, I think the last point in that comment referring to `odo deploy` still stands. Keeping the onus on IDE, or any other tooling that's consuming odo stream, to stop the `odo component status` process sounds like a deal-breaker (apologies, couldn't find a milder word).

Instead, if `odo component status -o json` command can be designed to be smart enough to handle these changes in URL/app, it might be a pain for us initially but, the users would find it to be the most useful command once they decide to deploy their application using odo. Something like "odo Day 2". :wink:

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+(...)
+echo "#devfile-status# {'buildStatus':'Compiling application'}
+mvn compile
+echo "#devfile-status# {'buildStatus':'Running unit tests'}
+mvn test
+```
+
+This 'detailed build status' markup text is *entirely optional*: if this markup is not present, the odo tool can still determine build succeeded/failed and build running/not-running using the other `odo push -o json` JSON events. 
+
+### App status notification via `odo component status -o json` and `odo log --follow`
```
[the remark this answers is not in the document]
jgwest: Agreed! Now, what I'm about to say may be a distinction without a useful different, but:  'App status' is definitely an imperfect term for the reasons you described, but I needed a term that also refers to whether the application's OS process is running inside the container or not. In contrast, the term component status is more general; it would refer to whether the pod/container is running, but not (necessarily) to whether the app is running.

These two terms are mostly the same, except in the following cases:
- User's applications fails to start, or dies after starting: In this case the component status is up (eg the pod/container are still running, the odo init volume is mapped in, etc), but the app is down.
- During a devfile build, the app running in the container may be temporarily stopped in order to allow the build to step to deploy new files; in this scenario the app would be down but the component would be up.

In any case, like I said, using app status is problematic, but I'll think a bit more if there's a better term I can use here to refer to both ideas, so that there is no confusion with the odo application/component model terminology 👍 .

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+
+In general, within the execution context that odo operates, there are a few ways for us to determine the application status:
+1) Can the application be pinged at its exposed URL?
```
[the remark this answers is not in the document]
jgwest: Good catch: Every case where I wrote 'ping' in the document, I actually meant 'send an HTTP/S request'... since 'ping' is is ambiguous language for this idea I've removed the term and replaced it with better HTTP/S-specific language.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+This 'detailed build status' markup text is *entirely optional*: if this markup is not present, the odo tool can still determine build succeeded/failed and build running/not-running using the other `odo push -o json` JSON events. 
+
+### App status notification via `odo component status -o json` and `odo log --follow`
+
+In general, within the execution context that odo operates, there are a few ways for us to determine the application status:
+1) Can the application be pinged at its exposed URL? 
+2) What state is the container in? (running/container creating/restarting/etc -- different statuses between local and Kube but same general idea)
+3) In the application log, specific hardcoded text strings can be searched for (for example, OpenLiberty outputs defined status codes to its log to indicate that an app started.)
```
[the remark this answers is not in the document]
jgwest: @kadel had the same concern in [a different part of the proposal](https://github.com/openshift/odo/pull/3177#discussion_r425101177). Odo definitely shouldn't contain/use any hardcoded text, so this proposal leaves it up to the IDEs to do this (at least for the scope of this proposal). In the future we may wish to add new devfile elements to allow odo to handle this.

As I did with the section @kadel referenced, I've improved the language in the above section to remove ambiguity over who (eg it is the IDE/consuming tool, not odo) would be responsible for using these logs to determine application status.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,143 @@
...
+
+Ideally, we would like for odo to provide consuming tools with all 3 sets of data. Thus, as proposed:
+- 1 and 2 are handled by a new `odo component status -o json` command, described here.
+- 3 is handled by the existing unmodified `odo log --follow` command.
+
+The new proposed `odo component status -o json` command will:
+- Be a *long-running* command that will continue outputing status until it is aborted by the parent process.
```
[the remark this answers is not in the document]
jgwest: Good point, and actually, in an earlier version of this design proposal this was called `odo component status -o json --follow`, but then I simplified it to remove `--follow`.

Would you prefer something like `odo component status -o json --follow`? This definitely makes the long-running behaviour of the command more clear.

The other item it sounds like you may be proposing in your above comment is combining the `odo log` and `odo component status` commands? I had considered this, but ultimately realized there is a case where the lifecycle of the two commands should differ: when the user adds a new URL to an existing component, the `odo component status` process needs to be restarted (eg to pick up the new URL), but the `odo log --follow` process does not need to be restarted (because the container/pod remain the same).

docs/proposals/event-notification.md
jgwest: Yep, the scenario you describe is exactly the problem with concurrent modification of the application while the status is running, and I included this line to indicate this `Note: This command will NOT, by design, respond to route/application changes that occur during or after it is first invoked. It is up to consuming tools to ensure that the odo component status command is stopped/restarted as needed.` 

How to handle this exact problem is a fundamental architectural constraint in the problem space of this proposal: we need a way to get always up-to-date status, and always up-to-date logs, but we also need the ability to deploy changes to the application. And we need to do this without turning the odo tools into either a server, an LSP-style process, or a multi-process application that needs to do IPC between various running instances. I'll go into more details in your next comment where you have added additional details.

docs/proposals/event-notification.md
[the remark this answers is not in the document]
dharmit: How about the term "developer application" when we're talking about the application inside the pod/container that a dev cares about? Two words, I know, but unambiguous to a good extent, IMO.

However, broader discussion around the terminology is happening at #3076 & #3190.

docs/proposals/event-notification.md
jgwest: Re: the 'deal-breaker' terminology, no worries 😄. Hopefully I can convince you otherwise... I have added [a new section to the proposal where I examine some of the other solutions I considered](https://github.com/jgwest/odo/blob/doc-2550-proposal/docs/proposals/event-notification.md#other-solutions-considered).

See that new section for details, but in short, the option I propose here -- requiring the IDE to restart the odo status process-- is admittedly not a perfect solution, but is less complex (and better meets our other requirements) than the alternatives.

Certainly we certainly COULD go with any of the other more sophisticated options: there are no technical impediments to doing so, and speaking personally it would be interesting/fun to implement them 😃 . 

BUT, I suspect this is a case where the perfect is the enemy of the good: it would take an order of magnitude greater time/effort to implement a more sophisticated solution, and in the mean time (with IDEs champing at the bit to add support for odo dev files) the existing requirement to provide app/build status doesn't go away: we would still need a short-term stop-gap measure which would probably look a lot like this proposal anyways.

Finally, re: Day 2, this option definitely does not in any way tie our hands in implementing a more complex solution in the future if/when they become necessary for 'day two' odo usage. This proposal is strictly scoped to current requirements: the `odo component status` command is ONLY designed to be invoked by IDEs (or other similar tools) that want to provide detailed app/build status to a user through a UI. It is not designed for users to invoke themselves (for example, from deployment scripts). 

But in a future day 2 proposal we could definitely look at using the building blocks discussed here to build command behaviour like that, ideally as part of a larger deployment story.

docs/proposals/event-notification.md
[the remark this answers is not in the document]
dharmit: Makes sense. Thanks for your efforts in explaining it in further detail. :+1:

docs/proposals/event-notification.md
kadel: Not a bad idea, but I'm afraid that using k8s health checks might have unwanted consequences. I wouldn't want to use LivelinessProbe as once it starts failing Kubernetes will restart the Pod. Using ReadinessProbe might work, but if ReadinessProbe starts to fail it Kubernetes will stop sending traffic to the Service, It might not be what you want when developing an application.

docs/proposals/event-notification.md
dharmit: > See that new section for details, but in short, the option I propose here -- requiring the IDE to restart the odo status process -- is admittedly not a perfect solution, but is less complex (and better meets our other requirements) than the alternatives.

Agreed.

> BUT, I suspect this is a case where the perfect is the enemy of the good:

+1. I have the negative trait of perfectionism. Doesn't mean I don't write buggy code - I do that all the time. :wink: But when I imagined the day 2 scenario and the "onus on IDE" part, I think it was my perfectionist point of view at play.

> Finally, re: Day 2, this option definitely does not in any way tie our hands in implementing a more complex solution in the future if/when they become necessary for 'day two' odo usage.

Most important thing is that it leaves the room open for further development around the ideas I had while still having a functionality that can be useful to the users (here, IDE plugins) fairly quickly.

Among the proposed solutions, the one about `fsnotify` makes more sense than client-server architecture. Of course they both come with their own pros & cons but, that's for another day when we plan to handle the complex stuff.

Thanks for this discussion and sharing more ideas! :+1:

docs/proposals/event-notification.md
dharmit: I think we agree upon this being a problem since its the end-user who has to perform an action. And as you said elsewhere, this leads to a scenario of the problem being solved at many places instead of being solved once in odo. But that could be taken up in the next phase, I guess.

docs/proposals/event-notification.md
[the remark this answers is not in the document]
dharmit: > Would you prefer something like `odo component status -o json --follow`? This definitely makes the long-running behaviour of the command more clear.

I agree that this makes the long-running behaviour more obvious to the user. However, I was thinking of `odo log --follow -o json` but your later point about `odo log --follow` being something that shouldn't need a restart makes sense as well and. So, I guess, `odo component status -o json --follow` would be more appropriate.

docs/proposals/event-notification.md
dharmit: @kadel that leaves us with startup probes which, from the documentation, sound like what might fit in. However, it would only make sense to use those if integrating those won't be a problem. Otherwise, coming up with our own solution for that shouldn't be a hassle, I guess.

docs/proposals/event-notification.md
kadel: > BUT, I suspect this is a case where the perfect is the enemy of the good: it would take an order of magnitude greater time/effort to implement a more sophisticated solution, and in the mean time (with IDEs champing at the bit to add support for odo dev files) the existing requirement to provide app/build status doesn't go away: we would still need a short-term stop-gap measure which would probably look a lot like this proposal anyways.

+1

I think that what is currently being proposed where tools need to restart `odo component status` after the component was modified is ok. If we are talking about IDE plugins, they will safely know when the `odo component status` restart will be required.

docs/proposals/event-notification.md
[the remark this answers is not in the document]
kadel: > However, broader discussion around the terminology is happening at #3076 & #3190.

I would keep the discussion about terminology separate in  #3076. We will need to change and redefine some of the terms used in odo across the odo all commands

docs/proposals/event-notification.md
jgwest: A followup: as discussed earlier in the thread, `supervisord ctl status` can be used to verify that long-running processes inside the container are running as expected.

I have prototyped it and it works well, so I've updated the proposal to include these data as another event type produced by `component status`.

PR #3184, opened 2020-05-15
Use "dev" environment instead of "cicd" to make binding name

wtam2018:
Binding names are constructed in this format <env>-<service>-bindnig but the <env> should be the environment that the service is in (not the cicd environment).  Otherwise, we cannot have two distinct bindings for services with the same service name in different environments.

In this PR, we also remove a dead function.


wtam2018:
Not targetted to master branch.

Issue #2937, opened 2020-04-20
Specific working directory defined in the command


elsony:
/kind user-story

## User Story
As a stack creator, I want to specify the working directory in a command in the devfile so that the command is executed in the given working directory in the container.

## Acceptance Criteria 
- [ ] It should support the workingDir definition as specified to the devfile 2.0 spec: https://github.com/che-incubator/devworkspace-api/issues/22

## Links 
- Related issue: https://github.com/che-incubator/devworkspace-api/issues/22

/kind user-story
/area devfile

kadel:
/status blocked
we need to have a parser implementation first #2918

maysunfaisal:
This seems to be covered by PR https://github.com/openshift/odo/pull/3216

yangcao77:
This issue should be closed. It's covered by https://github.com/openshift/odo/pull/3194

elsony:
Item completed

elsony:
Reopen since it turns out the code is on parser-v2 branch but not master.

yangcao77:
```
commands:
 - exec:
     commandLine: "./buildSchema.sh"
     workingDir: "~/schemaScripts"
     component: build-tools
     alias: buildSchema
```
Devfile v2 have the workingDir defined under `commands.exec`.  Any commandLine provided will be run under the workingDir defined. 
```
cmdArr = []string{adaptersCommon.ShellExecutable, "-c", "cd " + exec.WorkingDir  + " && " + *action.Command}
```

yangcao77:
/triage cancel

Verified on both Kube and Docker with master branch,  commands can be run under the `workingDir` provided. 

Since the parser v2 branch has been merged in to master. This item can be safely closed.

elsony:
Code merged. Closing.

PR #3194, opened 2020-05-18
V2 command

mik-dass:
**What type of PR is this?**
/kind code-refactoring

Fixes command.go for parser v2

**Note: for branch parser-v2**


yangcao77:
Fixes #2937

PR #3196, opened 2020-05-18
Update Command Logic to use groups

adisky:
**What type of PR is this?**

/kind code-refactoring

update common adapter to use groups

**NOTE : PR for parser-v2 branch**

### DIFF
--- a/pkg/devfile/adapters/common/command.go
+++ b/pkg/devfile/adapters/common/command.go
@@ -10,18 +10,37 @@ import (
 )
 
 // GetCommand iterates through the devfile commands and returns the associated devfile command
-func getCommand(data data.DevfileData, commandName string, required bool) (supportedCommand common.DevfileCommand, err error) {
+func getCommand(data data.DevfileData, commandName string, groupType common.DevfileCommandGroupType, required bool) (supportedCommand common.DevfileCommand, err error) {
+
 	for _, command := range data.GetCommands() {
-		if command.Exec.Id == commandName {
+		// validate command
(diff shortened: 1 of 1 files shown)


pkg/devfile/adapters/common/command.go
```diff
@@ -10,30 +10,49 @@ import (
...
-			err = validateCommand(data, command)
+		// if not command specified via flag, default command has the highest priority
+		if command.Exec.Group.Kind == groupType && command.Exec.Group.IsDefault {
+			supportedCommand = command
+			break
+		}
+
+		// return the first command found for the matching type.
+		if command.Exec.Group.Kind == groupType {
+			supportedCommand = command
```
mik-dass, line 40: should we return after this?

pkg/devfile/adapters/common/command.go
```diff
@@ -10,30 +10,49 @@ import (
...
 
-	// The command was not found
-	msg := fmt.Sprintf("The command \"%v\" was not found in the devfile", commandName)
-	if required {
-		// Not found and required, return an error
-		err = fmt.Errorf(msg)
-	} else {
-		// Not found and optional, so just log it
-		klog.V(3).Info(msg)
+	if commandName != "" && !found {
```
mik-dass, line 46: do we need this check? please add a comment here

pkg/devfile/adapters/common/command.go
adisky: Done

PR #3206, opened 2020-05-19
Update Docker and Kubernetes adapter to use groups

adisky:
**What type of PR is this?**

/kind code-refactoring

Update adapters to use commands maps, updated logic for common commands.go

Fixes #?

### DIFF
--- a/pkg/devfile/adapters/common/command.go
+++ b/pkg/devfile/adapters/common/command.go
@@ -26,6 +26,23 @@ func getCommand(data data.DevfileData, commandName string, groupType common.Devf
 		if commandName != "" {
 
 			if command.Exec.Id == commandName {
+
+				if supportedCommand.Exec.Group.Kind == "" {
+					// Devfile V1 for commands passed from flags
+					// Group type is not updated during conversion
+					command.Exec.Group.Kind = groupType
+				}
+
+				// we have found the command with name, its groupType Should match to the flag
+				// e.g --build-command "mybuild"
+				// exec:
+				//   id: mybuild
+				// group:
+				//   kind: build
+				if command.Exec.Group.Kind != groupType {
+					return supportedCommand, fmt.Errorf("mismatched type, command %s is of type %v groupType in devfile", commandName, groupType)
+
+				}
 				supportedCommand = command
 				return supportedCommand, nil
 			}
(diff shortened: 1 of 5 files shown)


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -305,76 +305,64 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
+	// Get Build Command
+	command, ok := commandsMap[versionsCommon.BuildCommandGroupType]
+	if ok {
+		compInfo.ContainerName = command.Exec.Component
+		err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show)
+		if err != nil {
+			return err
+		}
+
+		command, ok := commandsMap[versionsCommon.RunCommandGroupType]
```
mik-dass, line 343: why is the run command run only if a build command is present?

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -344,77 +344,61 @@ func getPortMap(context string, endpoints []*versionsCommon.Endpoint, show bool)
...
-
 	// Only add runinit to the expected commands if the component doesn't already exist
 	// This would be the case when first running the container
 	if !componentExists {
-		commandOrder = append(commandOrder, common.CommandNames{DefaultName: string(common.DefaultDevfileInitCommand), AdapterName: a.devfileInitCmd})
+		// Get Init Command
+		command, ok := commandsMap[versionsCommon.InitCommandGroupType]
+		if ok {
+
+			containerID := utils.GetContainerIDForAlias(containers, command.Exec.Component)
```
mik-dass, line 360: Should we rename this function? Because Alias is not there in v2 and this seems confusing to me.

pkg/devfile/adapters/common/types.go
```diff
@@ -52,3 +53,9 @@ type ComponentInfo struct {
 	PodName       string
 	ContainerName string
 }
+
+type CommandsMap map[common.DevfileCommandGroupType]common.DevfileCommand
```
mik-dass, line 57: Some comments regarding the variable

pkg/devfile/adapters/docker/component/utils.go
adisky: yes, we should rename, not doing it as a part of this PR.

pkg/devfile/adapters/kubernetes/component/adapter.go
adisky: my mistake, updated

pkg/devfile/adapters/docker/component/utils.go
mik-dass: OK

pkg/devfile/adapters/common/types.go
```diff
@@ -52,3 +53,11 @@ type ComponentInfo struct {
 	PodName       string
 	ContainerName string
 }
+
+// PushCommandsMap stores the commands to be executed as per there types.
```
mik-dass, line 57: their types

pkg/devfile/adapters/common/command.go
```diff
@@ -84,7 +101,7 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
 
 	isActionValid := false
 	for _, component := range components {
-		if command.Exec.Component == component.Container.Name && isComponentSupported(component) {
+		if command.Exec.Component == component.Container.Name {
```
mik-dass, line 104: don't we need the isComponentSupported check? It seems to be modified below

adisky: this is a redundant check, we are already doing the same check in  `GetSupportedComponents`

https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/common/utils.go#L101

Issue #2939, opened 2020-04-20
Support container component type in devfile 2.0


elsony:
/kind user-story

## User Story
As a stack creator, I want to use the new container component type in a devfile 2.0 to define the image used by a given component so that my command can execute is the given container.

## Acceptance Criteria 
- [ ] It should follow the specification as defined in https://github.com/che-incubator/devworkspace-api/issues/8 to define the container component in devfile 2.0.

## Links 
- Related issue: https://github.com/che-incubator/devworkspace-api/issues/8

/kind user-story
/area devfile

GeekArthur:
`dockerimage` component type depreciates, will check whether devfile 2.0 parser https://github.com/openshift/odo/pull/3188 handles `container` component type properly.

GeekArthur:
From quick glance, PR #3188 already supports `container` component type in both product https://github.com/openshift/odo/pull/3188/files#diff-8e874d12aadb58a96431fbf88d9fd3a2R122 and tests https://github.com/openshift/odo/pull/3188/files#diff-d16cd992949b0e25e1b445933b9f7ac1R9, I think this issue is done but let's wait for the devfile v2 is fully support so that we can do more testing then close the issue.

GeekArthur:
We need to do more testing, we may need to do some changes for `odo create` and `odo catalog` to fully support `container` component type.

kadel:
> From quick glance, PR #3188 already supports `container` component type in both product https://github.com/openshift/odo/pull/3188/files#diff-8e874d12aadb58a96431fbf88d9fd3a2R122 and tests https://github.com/openshift/odo/pull/3188/files#diff-d16cd992949b0e25e1b445933b9f7ac1R9, I think this issue is done but let's wait for the devfile v2 is fully support so that we can do more testing then close the issue.

I'm assigning this to @adisky, to track this correctly,  as she is working on initial implementations in parser-v2 branch.
if there is something left after initial implementation we can create new issues, or resign this.
/assign @adisky

adisky:
@elsony @kadel `container` component type is implemented with devfile V2 PR,
I can see some fields are not supported by odo, one is `sourceMapping` and other is `endpoints configuration`, support for them has to be done in this issue or we need a new issue?
https://github.com/openshift/odo/blob/master/pkg/devfile/parser/data/2.0.0/types.go#L232
```
type Container struct {
	Endpoints []*Endpoint `json:"endpoints,omitempty"`
[...]

Issue #2938, opened 2020-04-20
Support group in devfile commands (application lifecycle)


elsony:
/kind user-story

## User Story
As a stack creator, I want to define the group that a given command in a devfile belongs to so that odo know what default command to when it build, run, test and debug.

## Acceptance Criteria 
- [ ] It should follow the corresponding settings specified in the devfile 2.0 specification: https://github.com/che-incubator/devworkspace-api/issues/27
- [ ] It should support setting default for each group (see spec above)
- [ ] It should replace the existing `devBuild` and `devRun` commands on the push logic
- [ ] It should modify the existing `odo catalog list components` logic acceptance criteria to use the group info instead of the odo specific `devBuild` and `devRun` commands

## Links 
- Related issue: https://github.com/che-incubator/devworkspace-api/issues/27

/kind user-story
/area devfile

kadel:
this is blocked on  https://github.com/openshift/odo/issues/2918

/stats blocked

kadel:
> * [ ]  It should replace the existing `devBuild` and `devRun` commands on the push logic

@elsony What about devfile v1 support. Are we not going to support it anymore?

maysunfaisal:
Some observations for this work to be completed:
- If there are multiple group commands without a default, it should err out
- there needs to be one run group command defined, otherwise it should err out
- build group command optional like devfile v1

elsony:
> Some observations for this work to be completed:
> 
>     * If there are multiple group commands without a default, it should err out
> 
>     * there needs to be one run group command defined, otherwise it should err out
> 
>     * build group command optional like devfile v1

I would suggest to modify slightly to clarify:
1. If there are multiple group commands without a default, it should err out. This check is only done for the groups that odo consumes and the check is done when doing the corresponding execution, e.g. check on `debug` group only if running `odo push` with `--debug`
2. there needs to be at least one run group command defined, otherwise it should err out during `odo push`. Rule for multiple run command in point #1 still applies.
3. build group command optional like devfile v1

maysunfaisal:
@elsony what do you think about this issue here regarding command groups https://github.com/openshift/odo/pull/3216#discussion_r433471771

kadel:
> * [ ]  It should modify the existing `odo catalog list components` logic acceptance criteria to use the group info instead of the odo specific `devBuild` and `devRun` commands

I think that we should list all devfiles in the registry regardless of what command groups are there. 
[...]

Issue #3142, opened 2020-05-11
Migrate the devfile v1 Tests to devfile  v2


adisky:
/kind test

Followup issue for PR https://github.com/openshift/odo/pull/3216

prietyc123:
Thanks @adisky  for sharing the doc [1] for ```devfile v2 implementation plan```. AFAIU we are just adding a compatibility matrix of v2 for devfile support, I mean both v1 and v2 will be supported for devfile implementation and in later state v1 will be deprecated, right ?

However TBH I am clueless what test plan you are looking for it whereas we have the test for devfile in place. So can you please be more specific of what else test you are looking for. 

[1] https://docs.google.com/document/d/11BVhP1gViu5C3IGYQpiulX1lUkSGOG8p-lOJq4I5mUs/edit#heading=h.bw3vmyuddut1

prietyc123:
I am not sure about keeping this issue in this sprint. As I have gone through the planned execution for V2 parser and it seems more challenging that it might take time, may be the whole sprint. So, keeping this issue in this sprint doesn't make sense to me.

Even I feel if we assign this task to the owner of the V2 parser this will be done parallely without consuming additional time that someone may take in getting the context. This will release some bandwidth for other tasks. So, I am unassigning myself.

/unassign @prietyc123

prietyc123:
> I am not sure about keeping this issue in this sprint. As I have gone through the planned execution for V2 parser and it seems more challenging that it might take time, may be the whole sprint. So, keeping this issue in this sprint doesn't make sense to me.
> 
> Even I feel if we assign this task to the owner of the V2 parser this will be done parallely without consuming additional time that someone may take in getting the context. This will release some bandwidth for other tasks. So, I am unassigning myself.

ping @adisky

Issue #2933, opened 2020-04-20
Support Devfile 2.0 format on devfile components


elsony:
Add support to devfile 2.0 format when reading the devfile.yaml. Details of the devfile 2.0 support can be found in https://github.com/che-incubator/devworkspace-api/issues/15.

This item includes:
MVP:
- [x] Devfile 2.0 parser (#2918)
- [x] devfile v2 support for odo push (k8s and OpenShift) (#3140)
- [x] Specify the source mapping path for component containers (#2935) 
- [x] Container initialization (#2936)
- [x] Specific working directory in container (#2937)
- [x] Command group (application lifecycle) (#2938)
- [x] Renaming dockerimage component type (#2939)

MVP +:
- [x] Support composite commands (#3338)
- [x] Add parent support (devfile 2.0) to devfile (#2919 )
- [x] Allow the size of a component's volume to be specified in a Devfile (#3407)
- [x] Allow volumes to be shared across components (#3407)
- [x] Environment variables for a specific command (#3406)

/area devfile

girishramnani:
epics shouldn't be assigned `required-for-v2` labels

kadel:
/close
all done

PR #3216, opened 2020-05-20
Add Support for Devfile V2

### LINKED ISSUE #3140: devfile v2 support for odo push (k8s and OpenShift)
/kind user-story

## User Story
As an odo user, I want to be able to use devfile.yaml defined using v2 spec

## Acceptance Criteria 
- [x] `odo push` should autodetect devfile.yaml version
- [x] `odo push` It should create Deployment with containers and Services based on the devfile.yaml v2 containers definition
- [x] `odo push` should create URL (Route or Ingress) based on devfile v2 definition
- [x] `odo push` should create storage for storage defined in devfile v2

above is relevant for `odo preference set pushtarget kube`

some examples of devfile v2 https://github.com/sbose78/stack-registry/tree/sbose78-patch-1/stacks

## Links 
https://github.com/openshift/odo/issues/3145 handles this for Docker (local pustarget)

/kind user-story
/area devfilev2
/area devfile
/priority high

### LINKED ISSUE #3145: devfile v2 support for odo push (Docker)
/kind user-story

## User Story
As an odo user, I want to be able to use devfile.yaml defined using v2 spec

## Acceptance Criteria 
- [x] `odo push` should autodetect devfile.yaml version
- [x] `odo push` It should create and strat  containers  based on the devfile.yaml v2 containers definition
- [x] `odo push` should create URL (expose port) based on devfile v2 definition
- [x] `odo push` should create local (docker) storage for storage defined in devfile v2

above is relevant for `odo preference set pushtarget docker`

## Links 
https://github.com/openshift/odo/issues/3140 handles this for k8s and OpenShift

/kind user-story
/area devfilev2
/area devfile
/area local
/priority high

adisky:
**What type of PR is this?**
/kind feature

1. Add Devfile V2 Json Schema GO structures, Parser.
2. Update Common devfile structures to accommodate v2 types.
3. Updated adapters package to work with v2 devfiles.
4. Added converter from v2 to v1, to support v1 devfiles
5. Updated all unit and Integration tests.
6. Added v2 example devfiles

Detailed Information on implementation can be found in this doc
https://docs.google.com/document/d/11BVhP1gViu5C3IGYQpiulX1lUkSGOG8p-lOJq4I5mUs/edit#heading=h.bw3vmyuddut1

Fixes  #3145 
Fixes #3140 
#2939 #2938 

```
$ odo preference set experimental true
$ odo create nodejs <comp-name>
$  wget https://raw.githubusercontent.com/openshift/odo/parser-v2/tests/examples/source/devfilesV2/nodejs/devfile.yaml
$ `odo push`
```

Build binaries can be downloaded from unit prow job - https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_odo/3216/pull-ci-openshift-odo-master-unit/4685/artifacts/unit/dist/bin/


johnmcollier:
@adisky Can you also verify with the devfiles in https://github.com/elsony/devfile2-registry?

I tried with them, and the following devfiles failed:
- openLiberty
- java-spring-boot
- nodejs

All due to the same issue:
```
Executing devfile commands for component oltest
 ◐  Executing devBuild command "/artifacts/bin/build.sh" ✗  
Unable to exec command [/bin/sh -c cd /projects/openLiberty && /artifacts/bin/build.sh]: 
```

 It seems when `starterProject` is set in the devfile, instead of syncing to `/projects/<starterProjectName>`, it just syncs to `/projects`, which breaks the devfiles that rely on that.

kadel:
> It seems when `starterProject` is set in the devfile, instead of creating and syncing to `/projects/<starterProjectName>`, it just syncs to `/projects`, which breaks the devfiles that rely on that.

I'm not sure that this is how it should work.
If devfile has multiple starter projects each will have a different name, so it will end up in a different directory. How would you then setup the  `workingDir` dir for commands if it depends on what start project user chooses?

johnmcollier:
@kadel If there's one project set in the devfile, we decided to sync to `/projects/projectName` (see https://github.com/openshift/odo/pull/2681#issuecomment-597783079). If multiple projects are set in the devfile, it still only syncs to `/project`.

Wouldn't we need to do the same for `starterProjects` in devfile v2?

johnmcollier:
@adisky brought up on slack that `starterProjects` isn't yet present in the devfile v2 schema (see https://devfile.github.io/website/devfile/_attachments/api-reference.html), though it has been agreed to in https://github.com/devfile/kubernetes-api/issues/42.

As such, the rest of the expected functionality of `starterProjects` isn't present in this PR (user picking a starter project when running `odo create` in interactive mode, for example). So, maybe we open a story to add support for `starterProject` elements once it's been added to the v2 schema?

adisky:
> @adisky brought up on slack that `starterProjects` isn't yet present in the devfile v2 schema (see https://devfile.github.io/website/devfile/_attachments/api-reference.html), though it has been agreed to in [devfile/kubernetes-api#42](https://github.com/devfile/kubernetes-api/issues/42).
> 
> As such, the rest of the expected functionality of `starterProjects` isn't present in this PR (user picking a starter project when running `odo create` in interactive mode, for example). So, maybe we open a story to add support for `starterProject` elements once it's been added to the v2 schema?

@johnmcollier We can open an issue now also, what about the devfiles in v2 registry are you going to update them until `starterProjects` is implemented?

johnmcollier:
@adisky I'll check with Elson to see if we can update them to add/use the `projects` field (he may have wanted sample devfiles to showcase the `starterProjects` element). But sounds like we can at least work around the issues at the devfile level.

adisky:
> @adisky I'll check with Elson to see if we can update them to add/use the `projects` field (he may have wanted sample devfiles to showcase the `starterProjects` element). But sounds like we can at least work around the issues at the devfile level.

Sure, also I have logged separate issue to add devfile v2 registry.
https://github.com/openshift/odo/issues/3249

johnmcollier:
@adisky @kadel I've opened https://github.com/openshift/odo/issues/3263 to handle `starterProjects` separately.

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
-	if *action.Type != common.DevfileCommandTypeExec {
-		return fmt.Errorf("Actions must be of type \"exec\"")
+	if command.Exec == nil {
+		return fmt.Errorf("Command must be of type \"exec\"")
 	}
 
 	// component must be specified
-	if action.Component == nil || *action.Component == "" {
-		return fmt.Errorf("Actions must reference a component")
+	if command.Exec.Component == "" {
```
johnmcollier, line 103: I'm wondering if we should handle this as part of the devfile validator, rather than here. Then it would save a bunch of time if the user was using a devfile with invalid commands

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
+		return fmt.Errorf("Exec commands must reference a component")
 	}
 
 	// must specify a command
-	if action.Command == nil || *action.Command == "" {
-		return fmt.Errorf("Actions must have a command")
+	if command.Exec.CommandLine == "" {
```
johnmcollier, line 108: Same as before, wonder if we should move this into the devfile validator.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -305,81 +305,64 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
-						if componentExists && !common.IsRestartRequired(command) {
-							klog.V(4).Infof("restart:false, Not restarting DevRun Command")
-							err = exec.ExecuteDevfileRunActionWithoutRestart(&a.Client, action, command.Name, compInfo, show)
-							return
-						}
-
-						err = exec.ExecuteDevfileRunAction(&a.Client, action, command.Name, compInfo, show)
-					}
-				}
+		// Get Init Command
```
maysunfaisal, line 321: I have a few questions here with which you might help

1. Do we want to have three separate if blocks and execute the commands in order; or do we want to loop the map and use the previous idea of having a predefined order and executing it?
2. What can we do about refactoring code between Kubernetes Adapter and Docker Adapter execDevfile() since they're pretty much the same. See my comment here https://github.com/openshift/odo/issues/3015#issuecomment-634071947

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -344,83 +344,61 @@ func getPortMap(context string, endpoints []versionsCommon.DockerimageEndpoint,
 
 // Executes all the commands from the devfile in order: init and build - which are both optional, and a compulsary run.
 // Init only runs once when the component is created.
-func (a Adapter) execDevfile(pushDevfileCommands []versionsCommon.DevfileCommand, componentExists, show bool, containers []types.Container) (err error) {
+func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists, show bool, containers []types.Container) (err error) {
```
maysunfaisal, line 347: Pls see my comment on execDevfile() for the Kubernetes Adapter

pkg/devfile/parser/data/common/types.go
```diff
@@ -1,210 +1,449 @@
...
 
-	// The name of the devfile. Workspaces created from devfile, will inherit this
-	// name
-	Name *string `yaml:"name,omitempty" json:"name,omitempty"`
+	// Optional semver-compatible version
+	Version string `json:"version,omitempty"`
 }
 
-// Description of the projects, containing names and sources locations
+// CommandsItems
```
maysunfaisal, line 58: Could you please update the descriptions for all the exported items in this file?

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
 	"github.com/openshift/odo/pkg/devfile/parser/data"
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 	"k8s.io/klog"
-
-	"github.com/pkg/errors"
 )
 
 // GetCommand iterates through the devfile commands and returns the associated devfile command
-func getCommand(data data.DevfileData, commandName string, required bool) (supportedCommand common.DevfileCommand, err error) {
+func getCommand(data data.DevfileData, commandName string, groupType common.DevfileCommandGroupType) (supportedCommand common.DevfileCommand, err error) {
```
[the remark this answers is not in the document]
adisky: This is done to be in sync from here https://github.com/devfile/kubernetes-api/blob/master/pkg/apis/workspaces/v1alpha1/commands.go#L20, In future we might utilize these package directly.

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
+
 	for _, command := range data.GetCommands() {
-		if command.Name == commandName {
 
-			// Get the supported actions
-			supportedCommandActions, err := getSupportedCommandActions(data, command)
+		command = updateGroupforCustomCommand(commandName, groupType, command)
```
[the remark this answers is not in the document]
adisky: Actually, group for custom commands has to be inserted before `validateCommand`, otherwise it would fail with `group not found` . So it cannot bet inside `if commandName != ""`

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
+				}
+
+				// we have found the command with name, its groupType Should match to the flag
+				// e.g --build-command "mybuild"
+				// exec:
+				//   id: mybuild
+				// group:
+				//   kind: build
+				if command.Exec.Group.Kind != groupType {
+					return supportedCommand, fmt.Errorf("mismatched type, command %s is of type %v groupType in devfile", commandName, groupType)
```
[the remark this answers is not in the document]
adisky: here, i agree to mention `kind`, as this error would be shown to user.

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
-		// Not found and required, return an error
-		err = fmt.Errorf(msg)
-	} else {
-		// Not found and optional, so just log it
-		klog.V(3).Info(msg)
-	}
-
-	return
-}
+	if commandName == "" {
```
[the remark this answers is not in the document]
adisky: not sure about this, would try to do that

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
+		// if default command is not found return the first command found for the matching type.
+		for _, command := range data.GetCommands() {
```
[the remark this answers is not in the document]
adisky: sure, would update it.

pkg/devfile/adapters/common/command.go
adisky: @johnmcollier I also think validation should happen after parsing, but refactoring validation logic is not part of this PR :)
you can raise new issue for that.

pkg/devfile/adapters/common/types.go
```diff
@@ -52,3 +53,11 @@ type ComponentInfo struct {
 	PodName       string
 	ContainerName string
 }
+
+// PushCommandsMap stores the commands to be executed as per there types.
+type PushCommandsMap map[common.DevfileCommandGroupType]common.DevfileCommand
```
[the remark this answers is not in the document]
adisky: It has single command of a kind, but it contain multiple commands, so i think ` PushCommandsMap` is fine.

pkg/devfile/adapters/common/utils.go
```diff
@@ -81,8 +81,12 @@ type CommandNames struct {
 }
 
 func isComponentSupported(component common.DevfileComponent) bool {
-	// Currently odo only uses devfile components of type dockerimage, since most of the Che registry devfiles use it
-	return component.Type == common.DevfileComponentTypeDockerimage
+	// Currently odo only uses devfile components of type container, since most of the Che registry devfiles use it
+	if component.Container != nil {
+		klog.V(3).Infof("Found component \"%v\" with name \"%v\"\n", common.ContainerComponentType, component.Container.Name)
```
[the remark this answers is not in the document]
adisky: After parsing V1, `DevfileComponentTypeDockerimage` is converted to type `Container`

pkg/devfile/adapters/common/command.go
adisky: replied above

pkg/devfile/parser/data/common/types.go
```diff
@@ -1,210 +1,449 @@
...
+	VscodeLaunchCommandType DevfileCommandType = "VscodeLaunch"
+	CompositeCommandType    DevfileCommandType = "Composite"
+	CustomCommandType       DevfileCommandType = "Custom"
 )
 
-// ----------- Devfile Schema ---------- //
-type Attributes map[string]string
+// CommandGroupType describes the kind of command group.
+// +kubebuilder:validation:Enum=build;run;test;debug
+type DevfileCommandGroupType string
```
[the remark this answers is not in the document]
adisky: This is done to be in sync from here https://github.com/devfile/kubernetes-api/blob/master/pkg/apis/workspaces/v1alpha1/commands.go#L20, In future we might utilize these package directly.

pkg/devfile/adapters/kubernetes/component/adapter.go
adisky: > I have a few questions here with which you might help
> 
> 1. Do we want to have three separate if blocks and execute the commands in order; or do we want to loop the map and use the previous idea of having a predefined order and execute the commands?

Predefined order is not required as we have `kind` associated with each command, for now lets leave it like this, It has to be updated with `events` support.

> 2. What can we do about refactoring code between Kubernetes Adapter and Docker Adapter execDevfile() since they're pretty much the same. See my comment here [#3015 (comment)](https://github.com/openshift/odo/issues/3015#issuecomment-634071947)

Agreed, that can be refactored.

pkg/devfile/parser/data/common/types.go
```diff
@@ -1,210 +1,449 @@
...
 
-// Description of the projects, containing names and sources locations
+// CommandsItems
+type DevfileCommand struct {
+
+	// Exec command
+	Exec *Exec `json:"exec,omitempty"`
+
+	// Type of workspace command
+	Type DevfileCommandType `json:"type,omitempty"`
```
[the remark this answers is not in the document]
adisky: I have not added it, it is from schema
https://github.com/devfile/kubernetes-api/blob/master/pkg/apis/workspaces/v1alpha1/commands.go#L67
Actually for simplicity we have removed other types of command, so `DevfileCommandType` signifies which type of command it contains.  
Agreed, as currently we have only `exec`  so it makes no sense here.

adisky:
@johnmcollier regarding integration tests, as i know  @mik-dass discussed with @kadel, We would not write additional tests for `devfile V2` instead we would migrate all V1 tests to V2.

cc @mik-dass @kadel

pkg/devfile/adapters/common/command.go
johnmcollier: @adisky Fair enough 😄

pkg/devfile/adapters/common/utils.go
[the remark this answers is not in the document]
adisky: @johnmcollier Log level is same is before.

pkg/devfile/adapters/common/utils.go
```diff
@@ -112,24 +115,24 @@ func GetVolumes(devfileObj devfileParser.DevfileObj) map[string][]DevfileVolume
...
-		if comp.Volumes != nil {
-			for _, volume := range comp.Volumes {
+		if len(comp.Container.VolumeMounts) != 0 {
+			for _, volume := range comp.Container.VolumeMounts {
 				vol := DevfileVolume{
 					Name:          volume.Name,
-					ContainerPath: volume.ContainerPath,
-					Size:          &size,
+					ContainerPath: volume.Path,
+					Size:          size,
```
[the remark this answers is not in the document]
adisky: Thanks for getting our attention here, actually `VolumeMounts` and `Volume` are different in V2, and `Volume` as size not `VolumeMounts`. currently our code is handling `volumeMounts` only has `Volume` is new in devfile V2. We need to file a new issue to support `Volumes`.
https://docs.google.com/document/d/1GRGGIxQP3vidsdE3yAFgC0zHqHvgqdeDGQ5EtROXjMI/edit#heading=h.fg1yldouktd2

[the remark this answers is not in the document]
adisky: it is already included in this issue  
https://github.com/openshift/odo/issues/2933, but separate issue could also be logged.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -305,81 +305,64 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
...
 		return errors.New(fmt.Sprint("error executing devfile commands - there should be at least 1 command"))
 	}
 
-	commandOrder := []common.CommandNames{}
+	compInfo := common.ComponentInfo{
+		PodName: podName,
+	}
 
 	// Only add runinit to the expected commands if the component doesn't already exist
 	// This would be the case when first running the container
```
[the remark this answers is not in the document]
adisky: Done.

pkg/devfile/parser/data/common/types.go
adisky: Done.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -305,81 +305,64 @@ func (a Adapter) waitAndGetComponentPod(hideSpinner bool) (*corev1.Pod, error) {
 
 // Executes all the commands from the devfile in order: init and build - which are both optional, and a compulsary run.
```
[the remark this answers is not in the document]
adisky: Done.

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,153 @@ import (
...
+		}
 	}
 
 	return
 }
 
-// validateAction validates the given action
-// 1. action has to be of type exec
+// validateCommand validates the given command
+// 1. command has to be of type exec
```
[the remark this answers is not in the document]
adisky: Done.

kadel:
I've done some basic tests with both devfile v1 and v2, haven't found any big issues so far.
We still have a lot of work to make most of the stuff from v2 work, but from what I can see odo with this we have v2 working on the same level as v1.

/approve

adisky:
Issue logged for tests https://github.com/openshift/odo/issues/3142
Issue for throwing error in case multiple commands exist of same kind with none default https://github.com/openshift/odo/issues/3271

pkg/devfile/adapters/common/command.go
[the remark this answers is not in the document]
maysunfaisal: is the custom command update func call here for v1?

pkg/devfile/adapters/common/utils.go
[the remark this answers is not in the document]
maysunfaisal: ok thx for the heads up, we can do this with the issue when going through that epic

Yes volume mounts only has the name and path. Whereas volume is now a component, and a size can be mentioned along with it/

pkg/devfile/adapters/common/command.go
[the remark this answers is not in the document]
maysunfaisal: opened new issue https://github.com/devfile/kubernetes-api/issues/59

pkg/devfile/parser/data/common/types.go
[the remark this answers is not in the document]
maysunfaisal: alright, thx for the explanation. Do we need the `Type`, maybe it can be cleaned up later?

pkg/devfile/validate/components.go
```diff
@@ -19,17 +20,17 @@ func ValidateComponents(components []common.DevfileComponent) error {
 		return fmt.Errorf(ErrorNoComponents)
 	}
 
-	// Check wether component of type dockerimage is present
-	isDockerImageComponentPresent := false
+	// Check weather component of type container  is present
```
maysunfaisal, line 23: whether* typo, can be cleaned up later.

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,211 @@
+# ODO build and application status notification 
+
+## Summary
+
+With this proposal and [it's related issue](https://github.com/openshift/odo/issues/2550) we examine how to consume build/run event output from odo in order to allow external tools (such as IDEs) to determine the application/build status of an odo-managed application.
```
cdrage, line 5: its

docs/proposals/event-notification.md
```diff
@@ -0,0 +1,211 @@
...
+
+In short, the proposal is:
+- New flag for push command, `odo push -o json`: Outputs JSON events that correspond to what devfile actions/commands are being executed by push.
+- New odo command `odo component status -o json`: Calls `supervisord ctl status` to check the running container processes, checks the container/pod status every X seconds (and/or a watch, for Kubernetes case), and sends an HTTP/S request to the application URL. The result of those is output as JSON to be used by external tools to determine if the application is running.
```
cdrage, line 9: Hmm. Haven't heard of this before. @kadel thoughts? Have there been other discussions elsewhere?

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
+				}
+
+				// we have found the command with name, its groupType Should match to the flag
+				// e.g --build-command "mybuild"
+				// exec:
+				//   id: mybuild
+				// group:
+				//   kind: build
+				if command.Exec.Group.Kind != groupType {
+					return supportedCommand, fmt.Errorf("mismatched group kind, command %s is of group kind %v groupType in devfile", commandName, groupType)
```
cdrage, line 48: So the previous error output was like this: `The command \"%v\" was found but some of its actions are not supported` which was very useful to the user with regards to what went wrong. The output here is quite ambiguous. Can you elaborate more as to why the error is happening and why it's mismatched?

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
+	// if any command specified via flag is not found in devfile then it is an error.
+	if commandName != "" {
+		err = fmt.Errorf("The command \"%v\" is not found in the devfile", commandName)
+	} else {
+		msg := fmt.Sprintf("The command type \"%v\" is not found in the devfile", groupType)
+		// if run command is not found in devfile then it is an error
+		if groupType == common.RunCommandGroupType {
+			err = fmt.Errorf(msg)
+		} else {
+			klog.V(3).Info(msg)
```
cdrage, line 85: we should be using `4`

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
 // 3. command should be present
-func validateAction(data data.DevfileData, action common.DevfileCommandAction) (err error) {
+// 4. command must have group
+func validateCommand(data data.DevfileData, command common.DevfileCommand) (err error) {
 
 	// type must be exec
-	if *action.Type != common.DevfileCommandTypeExec {
-		return fmt.Errorf("Actions must be of type \"exec\"")
+	if command.Exec == nil {
+		return fmt.Errorf("Command must be of type \"exec\"")
```
cdrage, line 102: errors cannot be capitalized

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
 	}
 
 	// component must be specified
-	if action.Component == nil || *action.Component == "" {
-		return fmt.Errorf("Actions must reference a component")
+	if command.Exec.Component == "" {
+		return fmt.Errorf("Exec commands must reference a component")
```
cdrage, line 107: errors cannot be capitalized

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
 	}
 
 	// must specify a command
-	if action.Command == nil || *action.Command == "" {
-		return fmt.Errorf("Actions must have a command")
+	if command.Exec.CommandLine == "" {
+		return fmt.Errorf("Exec commands must have a command")
```
cdrage, line 112: errors cannot be capitalized

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
+	}
+
+	if command.Exec.Group == nil {
+		return fmt.Errorf("Exec commands must have group")
```
cdrage, line 116: errors cannot be capitalized

pkg/devfile/adapters/common/command.go
```diff
@@ -152,8 +168,8 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
 		klog.V(3).Infof("No init command was provided")
 	} else if !isInitCmdEmpty && initCmdErr == nil {
 		isInitCommandValid = true
-		pushDevfileCommands = append(pushDevfileCommands, initCommand)
-		klog.V(3).Infof("Init command: %v", initCommand.Name)
+		commandMap[common.InitCommandGroupType] = initCommand
+		klog.V(3).Infof("Init command: %v", initCommand.Exec.Id)
```
cdrage, line 172: 4

pkg/devfile/adapters/common/command.go
```diff
@@ -165,15 +181,15 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
 		klog.V(3).Infof("No build command was provided")
 	} else if !reflect.DeepEqual(emptyCommand, buildCommand) && buildCmdErr == nil {
 		isBuildCommandValid = true
-		pushDevfileCommands = append(pushDevfileCommands, buildCommand)
-		klog.V(3).Infof("Build command: %v", buildCommand.Name)
+		commandMap[common.BuildCommandGroupType] = buildCommand
+		klog.V(3).Infof("Build command: %v", buildCommand.Exec.Id)
```
cdrage, line 185: 4

pkg/devfile/adapters/common/command.go
```diff
@@ -165,15 +181,15 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
...
 	}
 
 	runCommand, runCmdErr := GetRunCommand(data, devfileRunCmd)
 	if runCmdErr == nil && !reflect.DeepEqual(emptyCommand, runCommand) {
-		pushDevfileCommands = append(pushDevfileCommands, runCommand)
 		isRunCommandValid = true
-		klog.V(3).Infof("Run command: %v", runCommand.Name)
+		commandMap[common.RunCommandGroupType] = runCommand
+		klog.V(3).Infof("Run command: %v", runCommand.Exec.Id)
```
cdrage, line 192: 4

pkg/devfile/adapters/common/command_test.go
```diff
@@ -10,389 +10,235 @@ import (
...
+						t.Errorf("TestGetCommand error: command names do not match expected: %v actual: %v", tt.retCommandName, cmd.Exec.Id)
+					}
 				}
 			}
 		})
 	}
 
 }
 
-func TestGetSupportedCommandActions(t *testing.T) {
```
cdrage, line 228: why was all of this removed?

pkg/devfile/adapters/common/command_test.go
```diff
@@ -486,52 +339,60 @@ func TestGetBuildCommand(t *testing.T) {
...
+					Group:       &versionsCommon.Group{Kind: buildGroup, IsDefault: true},
 				},
 			},
 			wantErr: false,
 		},
 		{
-			name:        "Case: Custom Build Command",
-			commandName: "customcommand",
-			commandActions: []common.DevfileCommandAction{
+			name:        "Case: Build Command passed through the odo flag",
```
cdrage, line 366: Although the previous tests has incorrect names, can you update the name of the cases in this PR? Ex. Each should be numbered Case 1, Case 2, etc in the tests.

pkg/devfile/adapters/common/command_test.go
```diff
@@ -671,84 +571,114 @@ func TestValidateAndGetPushDevfileCommands(t *testing.T) {
...
+			},
+			numberOfCommands: 1,
+			wantErr:          false,
 		},
 		{
-			name:               "Case: Missing Init Command with provided Build and Run Command",
-			initCommand:        emptyString,
-			buildCommand:       "customcommand",
-			runCommand:         "customcommand",
+			name:         "Case: Missing Init Command with provided Build and Run Command",
```
cdrage, line 631: Case 1, Case 2, etc.

pkg/devfile/adapters/common/types.go
```diff
@@ -14,9 +15,9 @@ type AdapterContext struct {
 
 // DevfileVolume is a struct for Devfile volume that is common to all the adapters
 type DevfileVolume struct {
-	Name          *string
-	ContainerPath *string
-	Size          *string
+	Name          string
```
cdrage, line 18: YESSS :+1: no more pointers.

pkg/devfile/adapters/common/utils.go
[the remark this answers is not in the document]
cdrage: Please change to 4, may be good to do it in this PR :+1:

pkg/devfile/adapters/common/utils_test.go
```diff
@@ -8,75 +8,56 @@ import (
...
+			component:            []versionsCommon.DevfileComponent{{Kubernetes: &versionsCommon.Kubernetes{}}},
 			expectedMatchesCount: 0,
 		},
+
 		{
-			name:                 "Case: Valid devfile with wrong component type (Openshift)",
-			componentType:        versionsCommon.DevfileComponentTypeOpenshift,
-			alias:                []string{"alias1", "alias2"},
-			expectedMatchesCount: 0,
+			name:                 "Case: Valid devfile with correct component type (Container)",
```
cdrage, line 38: Case 2, Case 3, etc.

pkg/devfile/adapters/docker/storage/utils.go
```diff
@@ -109,23 +109,23 @@ func ProcessVolumes(client *lclient.Client, componentName string, componentAlias
...
 		for _, vol := range volumes {
-			if _, ok := processedVolumes[*vol.Name]; !ok {
-				processedVolumes[*vol.Name] = true
+			if _, ok := processedVolumes[vol.Name]; !ok {
+				processedVolumes[vol.Name] = true
 
 				// Generate the volume Names
-				klog.V(3).Infof("Generating Docker volumes name for %v", *vol.Name)
-				generatedDockerVolName, err := GenerateVolName(*vol.Name, componentName)
+				klog.V(3).Infof("Generating Docker volumes name for %v", vol.Name)
```
cdrage, line 116: 4?

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -6,35 +6,216 @@ import (
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 )
 
-// GetComponents returns the slice of DevfileComponent objects parsed from the Devfile
+func (d *Devfile100) GetMetadata() common.DevfileMetadata {
+	// No GenerateName field in V2
```
cdrage, line 10: This part is confusing. There is no GenerateName in V2. But there *is* in V1. So we would need to still return GenerateName here as this is the `1.0.0` parser (hence the `1.0.0` dir). So we'd most likely need to include it.

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -6,35 +6,216 @@ import (
...
+	for _, v := range d.Components {
+		comps = append(comps, convertV1ComponentToCommon(v))
+	}
+	return comps
 }
 
 // GetAliasedComponents returns the slice of DevfileComponent objects that each have an alias
 func (d *Devfile100) GetAliasedComponents() []common.DevfileComponent {
+	// TODO(adi): we might not need this for V2 as name is a required field now.
+	var comps []common.DevfileComponent
```
cdrage, line 29: Again, see other comment. These sections are all about Version 1. So why are we making changes / breaking backwards compatibility here?

johnmcollier:
@cdrage Design proposal review was done in https://github.com/openshift/odo/pull/3177 (which was merged into the `parser-v2` branch)

pkg/devfile/parser/data/1.0.0/components.go
maysunfaisal: @cdrage Aditi can explain in more detail, but I think she had to make changes here to use the new common structs for both v1 and v2; so that v1 doesn't break

cdrage:
> @cdrage Design proposal review was done in #3177 (which was merged into the `parser-v2` branch)

Ah, missed that. Thanks!

pkg/devfile/adapters/common/command.go
```diff
@@ -7,140 +7,156 @@ import (
...
 	}
 
 	// must specify a command
-	if action.Command == nil || *action.Command == "" {
-		return fmt.Errorf("Actions must have a command")
+	if command.Exec.CommandLine == "" {
+		return fmt.Errorf("Exec commands must have a command")
+	}
+
+	if command.Exec.Group == nil {
```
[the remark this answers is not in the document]
elsony: We only fail the validation during `odo push` for a devfile that does not include at least one run command. Having said that, if the user specify `--run-command`, I agree that validation is not needed any more.  The reason is that the user already specify a run command so our `odo push` operation will work fine.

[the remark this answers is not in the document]
adisky: @maysunfaisal yes correct, if user is telling via odo flags what build and run commands are, we need not to fail for `groups`, that is why groups are inserted for custom commands,we need to know the group for commands passed via flag too as thats how we are executing it.

adisky:
> Err... the proposal, shouldn't it be in a separate PR?

@cdrage That PR mistakenly merged to `parser-v2` branch

pkg/devfile/adapters/common/command.go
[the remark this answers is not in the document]
adisky: @maysunfaisal we are updating group for commands specified by odo flag `--build-command`, `--run-command`, not for all commands. `updateGroupforCustomCommand`, we can remove the validation for group but group needs to be inserted for custom command, this is internally only so that user does not need to specify it in devfile.

pkg/devfile/adapters/common/command.go
adisky: @cdrage This is completely different error, let me explain it here
consider the below command specified in devfile
```
exec:
   id: "cmd-one"
   commandLine: "mvn compile"
   group:
        Kind: build
```
 now a user does odo push
`odo push --run-command cmd-one`
In this case it would error out, as command group specified in odo flag does not match specified in devfile

@maysunfaisal @cdrage  Would fix the strings in the err message and make it more intuitive to user.

pkg/devfile/parser/data/1.0.0/components.go
adisky: To add `generateName` field here, i need to add it common types. currently we have updated our code not to use any v1 specific fields,  `generateName` is currently used by `odo catalog list components` that has a parser of its own, so adding it here would of no use.

pkg/devfile/parser/data/1.0.0/components.go
adisky: @cdrage all v1 integration tests are passing, so it is the first sign we are not breaking anything.
secondly it is just a refactoring comment to remove this method from interface when we remove V1, as all components are aliased(they have name) for v2. agreed though comment is not clear.

pkg/devfile/adapters/common/command_test.go
adisky: actions are not present in devfile V2

adisky:
@maysunfaisal I think there is lot of confusion  about updating and validating groups, clearing it here

1. Is validation for group required `command.Exec.Group == nil `, I think no it can be removed
2. is updating of group required? we are updating groups only for commands specified by odo flags `--build-command`, `--run-command`, just to designate there group from flag value. 

case: 1 group not specified in devfile, no failure 
```
exec:
  id: "myrun"
  commandLine: "/run/server.sh"
$ odo push --run-command "myrun"
```

case: 2 group specified in devfile, no failure
```
exec:
  id: "myrun"
  commandLine: "/run/server.sh"
  group:
    kind: run
$ odo push --run-command "myrun"
```
case: 3 mismatched group, failure
```
exec:
  id: "myrun"
  commandLine: "mvn compile"
  group:
    kind: build
$ odo push --run-command "myrun"
```

pkg/devfile/adapters/common/command.go
adisky: @maysunfaisal @cdrage regarding the formatting of error in message, it is present in v1 too, the source of error is not from here, it needs to be resolved with another PR..
```
$ odo push

Validation
 ✗  Validating the devfile [35201ns]
 ✗  Failed to start component with name nodejs.
Error: Failed to create the component: failed to validate devfile build and run commands: 
The command "devbuild" was found but its actions are not supported: Problem with command "devbuild" action #1: Actions must be of type "exec"%!!(MISSING)(EXTRA string=
)
```

adisky: https://github.com/openshift/odo/issues/3280

docs/proposals/event-notification.md
kadel: This proposal was discussed in this PR https://github.com/openshift/odo/pull/3177
The PR was by mistake opened against `parser-v2` branch, so this is why it shows up here.

pkg/devfile/parser/data/1.0.0/components.go
girishramnani: the `odo catalog list components` uses `generateName` to figure out the component type i.e. nodejs. how would that change for v2 then?

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -24,17 +40,157 @@ func (d *Devfile100) GetAliasedComponents() []common.DevfileComponent {
...
+		if *action.Type == DevfileCommandTypeExec {
+			exec = common.Exec{
+				Attributes:  c.Attributes,
+				CommandLine: *action.Command,
+				Component:   *action.Component,
+				Group:       getGroup(c.Name),
+				Id:          c.Name,
+				WorkingDir:  action.Workdir,
+				// Env:
+				// Label:
```
girishramnani, line 97: shouldn't we remove these?

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -24,17 +40,157 @@ func (d *Devfile100) GetAliasedComponents() []common.DevfileComponent {
...
+	}
+
+	// TODO: Previewurl
+	return common.DevfileCommand{
+		//TODO(adi): Type
+		Exec: &exec,
+	}
+}
+
+func convertV1ComponentToCommon(c Component) (d common.DevfileComponent) {
```
girishramnani: I dont think the named argument is needed here for return value?

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -24,17 +40,157 @@ func (d *Devfile100) GetAliasedComponents() []common.DevfileComponent {
...
+	var volumes []*common.VolumeMount
+	for _, v := range c.ComponentDockerimage.Volumes {
+		volumes = append(volumes, convertV1VolumeToCommon(v))
+	}
+
+	container := common.Container{
+		Name:         *c.Alias,
+		Endpoints:    endpoints,
+		Env:          envs,
+		Image:        *c.ComponentDockerimage.Image,
```
girishramnani: null checks needed here?

pkg/devfile/validate/components.go
```diff
@@ -2,13 +2,14 @@ package validate
...
 
 // Errors
 var (
-	ErrorNoComponents           = "no components present"
-	ErrorNoDockerImageComponent = fmt.Sprintf("odo requires atleast one component of type '%s' in devfile", common.DevfileComponentTypeDockerimage)
+	ErrorNoComponents         = "no components present"
+	ErrorNoContainerComponent = fmt.Sprintf("odo requires atleast one component of type '%s' in devfile", common.ContainerComponentType)
 )
 
 // ValidateComponents validates all the devfile components
```
girishramnani, line 15: can we make this function lowercase? it seems to not be used anywhere outside this package.

pkg/devfile/parser/context/apiVersion.go
```diff
@@ -19,19 +19,34 @@ func (d *DevfileCtx) SetDevfileAPIVersion() error {
...
-	if !ok {
-		return fmt.Errorf("apiVersion not present in devfile")
-	}
+	var apiVer string
+
+	// Get "apiVersion" value from map for devfile V1
+	apiVersion, okApi := r["apiVersion"]
+
+	// Get "schemaVersion" value from map for devfile V2
+	schemaVersion, okSchema := r["schemaVersion"]
```
girishramnani, line 28: what happens if both are provided, apiVersion and schemaVersion?

adisky: there is no field `apiVersion` for V2 and no field `schemaVersion` for V1, so they cannot be provided together, json validation would fail for such case.

pkg/devfile/parser/data/1.0.0/components.go
```diff
@@ -6,35 +6,222 @@ import (
...
+
+	component = common.DevfileComponent{Container: &container}
+
+	return component
+}
+
+func convertV1EndpointsToCommon(e DockerimageEndpoint) common.Endpoint {
+	return common.Endpoint{
+		// Attributes:
+		// Configuration:
```
girishramnani, line 147: should we remove these commented attributes?

pkg/devfile/parser/data/1.0.0/components.go
adisky: thats what i said, `odo catalog list components` has parser of its own, so no affect to add it here

girishramnani:
small nitpicks, rest the code is great :+1:

pkg/devfile/parser/data/1.0.0/components.go
adisky: you are reviewing an previous commit, `string pointers` has been removed at many places.

girishramnani:
@adisky the `pkg/devfile/validate` package in my opinion might not be needed and possibly can just be a `validate.go` file inside the `pkg/devfile/adapters` or somewhere.

adisky:
> @adisky the `pkg/devfile/validate` package in my opinion might not be needed and possibly can just be a `validate.go` file inside the `pkg/devfile/adapters` or somewhere. something to consider in a followup

agreed it is not needed, it was in my mind too, I would log a refactoring issue tomorrow.

pkg/devfile/parser/data/1.0.0/components.go
cdrage: Honestly I still think we should add GenerateName here... This is fairly confusing as this is the v1.0.0 parser folder but we are implementing v2 features.

pkg/devfile/parser/data/1.0.0/components.go
cdrage: Then we should rename the folder / all mentions of v1.0.0 to something else in this folder as to be honest, we are implementing v2 features where v1 is..

cdrage:
I still don't like how we have a folder named 1.0.0 which we have 2.0.0 features implemented..

At the moment, I don't mind merging this as well once tests pass. But we will 100% need to refactor this later on so that it's less confusing on implementation.

/lgtm

adisky:
> I still don't like how we have a folder named 1.0.0 which we have 2.0.0 features implemented..
> 
> At the moment, I don't mind merging this as well once tests pass. But we will 100% need to refactor this later on so that it's less confusing on implementation.
> 
> /lgtm

@cdrage Please read the doc about v2 implementation plan, Hope it makes things clear to you!!
https://docs.google.com/document/d/11BVhP1gViu5C3IGYQpiulX1lUkSGOG8p-lOJq4I5mUs/edit#

PR #3238, opened 2020-05-22
replace secured with secure

### LINKED ISSUE #3237: incorrect sample in url create
/kind bug

the sample has a type, should be `secure` instead of `secured`
```
ODO_BOOTSTRAPPER_IMAGE=quay.io/zhengxiaomei123/odo4z:1.1.3-s390x ~/go/src/github.com/openshift/odo/dist/bin/linux-s390x/odo  url create --secured
Error: unknown flag: --secured
Usage:
  odo url create [url name] [flags]

Examples:
  # Create a URL with a specific host by automatically detecting the port used by the component (using CRC as an exampple)
  odo url create example  --host apps-crc.testing

  # Create a URL with a specific name and host (using CRC as an example)
  odo url create example --host apps-crc.testing

  # Create a URL for the current component with a specific port and host (using CRC as an example)
  odo url create --port 8080 --host apps-crc.testing

  # Create a URL of ingress kind for the current component with a host (using CRC as an example)
  odo url create --host apps-crc.testing --ingress

  # Create a secured URL for the current component with a specific host (using CRC as an example)
  odo url create --host apps-crc.testing --secured
```

jichenjc:
A typo existing in sample file

/kind bug

[skip ci]
no need to CI as it's modify sample output and the real test cases for `secure` is correct using
no need to add additionl test

Fixes #3237 

### DIFF
--- a/pkg/odo/cli/url/create.go
+++ b/pkg/odo/cli/url/create.go
@@ -53,8 +53,8 @@ var (
 	# Create a URL of ingress kind for the current component with a host (using CRC as an example)
 	%[1]s --host apps-crc.testing --ingress
 
-	# Create a secured URL for the current component with a specific host (using CRC as an example)
-	%[1]s --host apps-crc.testing --secured
+	# Create a secure URL for the current component with a specific host (using CRC as an example)
+	%[1]s --host apps-crc.testing --secure
 	  `)
 
 	urlCreateExampleDocker = ktemplates.Examples(`  # Create a URL with a specific name by automatically detecting the port used by the component


pkg/odo/cli/url/create.go
```diff
@@ -54,7 +54,7 @@ var (
 	%[1]s --host apps-crc.testing --ingress
 
 	# Create a secured URL for the current component with a specific host (using CRC as an example)
```
prietyc123, line 56: if you can replace in comment comment also.

kadel:
/retest
small fix
/lgmt
/approve

pkg/odo/cli/url/create.go
jichenjc: updated, thanks for reminder

PR #3240, opened 2020-05-22
Fixes unit tests

mik-dass:
**What type of PR is this?**

/kind code-refactoring

Fixes uni tests got parser v2

### DIFF
--- a/pkg/devfile/adapters/docker/component/adapter_test.go
+++ b/pkg/devfile/adapters/docker/component/adapter_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/golang/mock/gomock"
 	adaptersCommon "github.com/openshift/odo/pkg/devfile/adapters/common"
 	devfileParser "github.com/openshift/odo/pkg/devfile/parser"
-	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 	versionsCommon "github.com/openshift/odo/pkg/devfile/parser/data/common"
 	"github.com/openshift/odo/pkg/lclient"
 	"github.com/openshift/odo/pkg/testingutil"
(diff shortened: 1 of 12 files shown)


pkg/kclient/deployments_test.go
```diff
@@ -18,7 +18,7 @@ import (
 // createFakeDeployment creates a fake deployment with the given pod name and labels
 func createFakeDeployment(fkclient *Client, fkclientset *FakeClientset, podName string, labels map[string]string) (*appsv1.Deployment, error) {
 	fakeUID := types.UID("12345")
-	container := GenerateContainer("container1", "image1", true, []string{"tail"}, []string{"-f", "/dev/null"}, []corev1.EnvVar{}, corev1.ResourceRequirements{}, []corev1.ContainerPort{})
```
adisky, line 21: we need to keep args,commands that would be added in v2

pkg/kclient/deployments_test.go
```diff
@@ -100,7 +100,7 @@ func TestCreateDeployment(t *testing.T) {
 
 func TestUpdateDeployment(t *testing.T) {
 
-	container := GenerateContainer("container1", "image1", true, []string{"tail"}, []string{"-f", "/dev/null"}, []corev1.EnvVar{}, corev1.ResourceRequirements{}, []corev1.ContainerPort{})
```
adisky, line 103: ditto

pkg/kclient/generators_test.go
```diff
@@ -21,8 +21,6 @@ func TestGenerateContainer(t *testing.T) {
 		name         string
 		image        string
 		isPrivileged bool
-		command      []string
```
adisky, line 24: ditto, same for all places below.

Issue #2978, opened 2020-04-23
odo create interactive mode asks for namespace when the push target is set to docker


yangcao77:
/kind bug
/area devfile

## What versions of software are you using?

**Operating System:**
macos
**Output of `odo version`:**
master branch
## How did you run odo exactly?
enable docker mode by setting the pushtarget
run `odo create`

## Actual behavior
```
$ odo preference set pushtarget docker -f
Global preference was successfully updated
$ odo create
Experimental mode is enabled, use at your own risk
? Which devfile component type do you wish to create openLiberty
? What do you wish to name the new devfile component openLiberty
? What namespace do you want the devfile component to be created in 
```

## Expected behavior
Should not ask for namespace since namespace is only applicable for kube

girishramnani:
@johnmcollier are you still working on this?

PR #3248, opened 2020-05-26
Don't prompt for namespace if pushtarget is Docker

johnmcollier:

**What type of PR is this?**

/kind bug

This PR removes the prompt for the user to enter a namespace when running `odo create` with the pushtarget set to Docker. Since no namespaces are used in odo's Docker support, we don't need to prompt for it.

Fixes https://github.com/openshift/odo/issues/2978

1. Set pushtarget to Docker
2. `odo create`
You should not be prompted to enter a namespace


GeekArthur:
/lgtm
🚀

mik-dass:
```
[Fail] odo service command tests for OperatorHub When operator backed services are created [It] should list the services if they exist 
```
/retest

PR #3259, opened 2020-05-28
Add validation for parameters passed with  command

### LINKED ISSUE #3036: "odo config set" doesn't validate port values and thus "odo describe" panics
/kind bug

**Operating System:** all

**Output of `odo version`:** master

https://github.com/openshift/odo/pull/3008#discussion_r416561531
```sh
$ odo config set ports 909,808 -f
 ✓  Local config successfully updated

Run `odo push --config` command to apply changes to the cluster

$ odo describe 
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/openshift/odo/pkg/component.getPortsProtocolMapping(0xc0006136c0, 0x2, 0x2, 0x0)
        /home/mrinaldas/go/src/github.com/openshift/odo/pkg/component/component.go:991 +0x177
github.com/openshift/odo/pkg/component.GetComponentFromConfig(0xc000ba7500, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/mrinaldas/go/src/github.com/openshift/odo/pkg/component/component.go:962 +0x6d6
github.
```

## Actual behavior
`odo describe` panics because `odo config set` doesn't perform a check for ports provided by the user

## Expected behavior
odo should validate that provided value for ports follow `<port>/<protocol>` format and the values are valid.

dev-gaur:
kind bug

Adds Validation on parameters passed with `odo config set ` commands

Fixes #3036

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -200,7 +200,7 @@ func newProxyLocalConfig() proxyLocalConfig {
 func (lci *LocalConfigInfo) SetConfiguration(parameter string, value interface{}) (err error) {
 	// getting the second arg makes sure that this never panics
 	strValue, _ := value.(string)
-	if parameter, ok := asLocallySupportedParameter(parameter); ok {
+	if parameter, ok := AsLocallySupportedParameter(parameter); ok {
 		switch parameter {
 		case "type":
 			lci.componentSettings.Type = &strValue
@@ -326,7 +326,7 @@ func (lci *LocalConfigInfo) ConfigFileExists() bool {
 
 // DeleteConfiguration is used to delete config from local odo config
 func (lci *LocalConfigInfo) DeleteConfiguration(parameter string) error {
-	if parameter, ok := asLocallySupportedParameter(parameter); ok {
+	if parameter, ok := AsLocallySupportedParameter(parameter); ok {
 
 		switch parameter {
 		case "cpu":
@@ -362,7 +362,7 @@ func (lci *LocalConfigInfo) DeleteURL(parameter string) error {
 // parameter is the name of the config parameter
 // value is the value to be deleted
 func (lci *LocalConfigInfo) DeleteFromConfigurationList(parameter string, value string) error {
-	if parameter, ok := asLocallySupportedParameter(parameter); ok {
+	if parameter, ok := AsLocallySupportedParameter(parameter); ok {
 		switch parameter {
 		case "storage":
 			for i, storage := range lci.GetStorage() {
@@ -653,7 +653,8 @@ func FormatLocallySupportedParameters() (result string) {
 	return "\nAvailable Local Parameters:\n" + result
 }
 
-func asLocallySupportedParameter(param string) (string, bool) {
+// AsLocallySupportedParameter returns the parameter in lower case and a boolean indicating if it is a supported parameter
+func AsLocallySupportedParameter(param string) (string, bool) {
 	lower := strings.ToLower(param)
 	return lower, lowerCaseLocalParameters[lower]
 }
--- a/pkg/odo/cli/config/set.go
+++ b/pkg/odo/cli/config/set.go
@@ -9,6 +9,7 @@ import (
 	clicomponent "github.com/openshift/odo/pkg/odo/cli/component"
 	"github.com/openshift/odo/pkg/odo/cli/ui"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
+	"github.com/openshift/odo/pkg/odo/util/validation"
 	"github.com/pkg/errors"
 
 	"github.com/spf13/cobra"
@@ -62,6 +63,7 @@ func (o *SetOptions) Complete(name string, cmd *cobra.Command, args []string) (e
 		o.paramName = args[0]
 		o.paramValue = args[1]
 	}
+
 	// we initialize the context irrespective of --now flag being provided
 	if o.now {
 		o.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
@@ -152,6 +154,33 @@ func (o *SetOptions) Run() (err error) {
 	return nil
 }
 
+func isValidArgumentList(args []string) error {
+
+	if len(args) < 2 {
+		return fmt.Errorf("please provide a parameter name and value")
+	} else if len(args) > 2 {
+		return fmt.Errorf("only one value per parameter is allowed")
+	}
+
+	var err error
+	param, ok := config.AsLocallySupportedParameter(args[0])
+
+	if !ok {
+		err = errors.Errorf("the provided parameter is not supported, %v", args[0])
+	}
+
+	switch param {
+	case "ports", "debugport":
+		err = validation.PortsValidator(args[1])
+	}
+
+	if err != nil {
+		err = errors.Errorf("validation failed for the provided arguments, %v", err)
+	}
+
+	return err
+}
+
 // NewCmdSet implements the config set odo command
 func NewCmdSet(name, fullName string) *cobra.Command {
 	o := NewSetOptions()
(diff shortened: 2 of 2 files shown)


dharmit:
Couple of things:
```sh
$ odo config set ports "8080/TCP/UDP" -f
E0529 11:36:57.513401    6790 config.go:690] Validation failed for the provided arguments, Error: unable to parse the port string 8080/TCP/UDP
 ✗  unknown parameter :'' is not a parameter in local odo config
```
IMO, we should simply print the error message. Details about error number, line where it occurred, etc. should not be shown to the user.

Also, why is it showing the error `unknown parameter :'' is not a parameter in local odo config`? I'm little confused there. `Error: unable to parse the port string 8080/TCP/UDP` looks good, however.

IMO, below should fail because I've not provided TCP or UDP to port 9090:
```sh
$ odo config set ports "8080/TCP,9090" -f
 ✓  Local config successfully updated

Run `odo push --config` command to apply changes to the cluster

$ odo config view
COMPONENT SETTINGS
Type              nodejs
Application       app
Project           default
SourceType        local
Ref               
SourceLocation    ./
Ports             8080/TCP,9090
Name              nodejs-nodejs-ex-wcxj
MinMemory         
MaxMemory         
DebugPort         
Ignore            
MinCPU            
MaxCPU
```
Another thing from code point of view - please add a comment describing what `IsValidLArgumentList` is doing.

dev-gaur:
@dharmit 

```
odo config set ports "8080/TCP,9090" -f
```
is not failing because the port validator function accepts it.

I'm using existing validation utilities so that the validation behavior remains uniform. Should I create a different validator ?


https://github.com/openshift/odo/blob/6bd0d73dc67e014e34f7e92a35dab26734163afc/pkg/odo/util/validation/validators.go#L60

dharmit:
> Should I create a different validator ?
 
Nope. My bad. Should have read the comments that said it's a valid value.

girishramnani:
If I remember correctly not providing <PORT>TCP/UDP defaults to TCP I think in most tools

dharmit:
> If I remember correctly not providing the protocol in `<PORT>TCP/UDP` defaults to TCP I think in most tools

Behind the scenes, it does. I tried creating a URL for port without protocol info in `config.yaml` and it works. It doesn't show up in `odo config view`, however. Not an issue, IMO.

dev-gaur:
> > If I remember correctly not providing the protocol in `<PORT>TCP/UDP` defaults to TCP I think in most tools
> 
> Behind the scenes, it does. I tried creating a URL for port without protocol info in `config.yaml` and it works. It doesn't show up in `odo config view`, however. Not an issue, IMO.

I think would help with respect to clarity for the user to specify that in `odo config view`

dharmit:
> I think it would help to be clear with the user if we specify that in `odo config view`

Or, it could be mentioned in the CLI documentation or actual documentation. If we'd want to make it clear in the output of `odo config view`, we would have to modify the validation function.

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+
+	if !ok {
+		return errors.Errorf("the parameter provided : %v, is not present in the supported parameters list.", args[0])
+	}
+	var err error
+	switch param {
+	case "ports", "debugport":
+		err = validation.PortsValidator(args[1])
+	default:
+		if len(args) < 2 || len(args[1]) == 0 {
```
girishramnani, line 751: Shouldn’t this check be at the top?

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+
+	param, ok := asLocallySupportedParameter(args[0])
+
+	if !ok {
+		return errors.Errorf("the parameter provided : %v, is not present in the supported parameters list.", args[0])
+	}
+	var err error
+	switch param {
+	case "ports", "debugport":
+		err = validation.PortsValidator(args[1])
```
girishramnani, line 749: What would happen if I provided `odo config set ports` without any value for the ports? Probably would panic due to `args[1]`

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+// IsValidArgumentList validates the argument list for `odo config` command
+func IsValidArgumentList(args []string) error {
+	if len(args) == 0 {
+		return nil
+	}
+
+	param, ok := asLocallySupportedParameter(args[0])
+
+	if !ok {
+		return errors.Errorf("the parameter provided : %v, is not present in the supported parameters list.", args[0])
```
girishramnani, line 744: There is an error string convention that error don’t start with capitals or end with a ### fullstop

pkg/config/config.go
dev-gaur: There's a check for this https://github.com/openshift/odo/pull/3259/files#diff-091ace02b49a46f55f2586cfec89a160R737

pkg/config/config.go
```diff
@@ -730,3 +731,30 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 
 	return sourceOSPath, err
 }
+
+// IsValidArgumentList validates the argument list for `odo config` command
+func IsValidArgumentList(args []string) error {
```
metacosm, line 736: This function should actually be merged with the `Args` function that is passed to the command struct when it's created since otherwise the check on args length will never occur (as it's done in that function). Actually, there would be merits to check the args there instead of in `Complete`…

girishramnani:
@dev-gaur CI failing

girishramnani:
CI seem to fail but code is good hence, please fix the tests if need be.
/approve

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+	}
+
+	var err error
+	param, ok := asLocallySupportedParameter(args[0])
+
+	if !ok {
+		err = errors.Errorf("the parameter provided : %v, is not supported", args[0])
+	}
+
+	switch param {
```
amitkrout, line 751: +100, I feel happy when i see switch statement instead of multiple if else

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
 
 	return sourceOSPath, err
 }
+
+// IsValidArgumentList validates the argument list for `odo config` command
+func IsValidArgumentList(args []string) error {
```
metacosm, line 736: This function should not be in that file as it only deals with command-related code, not config-related functionality per se. My advice would be to put it with the set command and unexport it.

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+		return fmt.Errorf("please provide a parameter name and value")
+	} else if len(args) > 2 {
+		return fmt.Errorf("only one value per parameter is allowed")
+	}
+
+	var err error
+	param, ok := asLocallySupportedParameter(args[0])
+
+	if !ok {
+		err = errors.Errorf("the parameter provided : %v, is not supported", args[0])
```
metacosm, line 748: The wording is awkward. See my previous comment on this topic.

pkg/config/config.go
```diff
@@ -730,3 +731,31 @@ func (lci *LocalConfigInfo) GetOSSourcePath() (path string, err error) {
...
+func IsValidArgumentList(args []string) error {
+
+	if len(args) < 2 {
+		return fmt.Errorf("please provide a parameter name and value")
+	} else if len(args) > 2 {
+		return fmt.Errorf("only one value per parameter is allowed")
+	}
+
+	var err error
+	param, ok := asLocallySupportedParameter(args[0])
```
metacosm, line 745: Moving the enclosing function to the set command will require exporting this function.

girishramnani:
@dev-gaur any update on the comments made by chris?

Issue #3128, opened 2020-05-08
Better error when incorrect component type has been specified


elsony:
/kind bug
## What versions of software are you using?
1.2

**Operating System:**
mac

**Output of `odo version`:**

## How did you run odo exactly?
When I specified a wrong component type name during on create in the devfile scenario, the error message that comes back does not clearly point out on what may have caused the problem.

## Actual behavior
```
eyuentp2:springboot-ex eyuen$ odo create java-sprint-boot myspring
Experimental mode is enabled, use at your own risk

Validation
 ✗  Checking devfile compatibility [42454ns]

Please run 'odo catalog list components' for a list of supported devfile component types
 ✗  imagestreams.image.openshift.io "java-sprint-boot" not found
eyuentp2:springboot-ex eyuen$ odo create java-spring-boot myspring
```

## Expected behavior
It should return a more direct error that indicates the component type does not exists.

## Any logs, error output, etc?
See actual behaviour section

/area devfile

johnmcollier:
So the root issue of this is that we always fall back to checking for an s2i image stream of that name, if a devfile component wasn't found. And if no s2i imagestream is found, we output the above error message, which may be consufin

We discussed on an earlier team meeting and decided to change the error message that gets printed when experimental mode is enabled and no devfile component or image stream was found:

Before
```
 ✗  imagestreams.image.openshift.io "java-sprint-boot" not found
```

After
```
 ✗  component "java-sprint-boot" not found
```

If experimental mode is not enabled, the original error message would still show up.

@kadel @elsony Any concerns with that?

PR #3265, opened 2020-05-28
Don't show imagestreams error on experimental mode

johnmcollier:

**What type of PR is this?**

/kind cleanup

When a user enters an invalid component name to `odo create` when experimental mode is enabled on OpenShift, they see this error:
```
 ✗  imagestreams.image.openshift.io "dsfdsfd" not found
```

This happens because on OpenShift (with odo's experimental mode enabled), if no valid devfile component with the given name is found, we drop down to checking for s2i/imagestream components instead, rather than erroring out.

This error can be confusing if the user doesn't know what imagestreams are / was only intending to create from a devfile component. So instead, I modified the error message that's returned when an image stream can't be found to display the following when experimental mode is enabled:
```
 ✗  component "dsfdsfd" not found
```

This PR does not affect odo without experimental mode enabled, so the regular imagestreams error message is shown.

Fixes https://github.com/openshift/odo/issues/3128

1. Log in to an OpenShift cluster and enable odo's experimental mode
2. Run `odo create <some-fake-component>`
3. Error message should say `component "<some-fake-component>" not found`

### DIFF
--- a/pkg/occlient/occlient.go
+++ b/pkg/occlient/occlient.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/odo/pkg/config"
 	"github.com/openshift/odo/pkg/devfile/adapters/common"
 	"github.com/openshift/odo/pkg/log"
+	"github.com/openshift/odo/pkg/odo/util/experimental"
 	"github.com/openshift/odo/pkg/preference"
 	"github.com/openshift/odo/pkg/util"
 
@@ -374,6 +375,9 @@ func (c *Client) GetPortsFromBuilderImage(componentType string) ([]string, error
 	}
 	imageStream, err := c.GetImageStream(imageNS, imageName, imageTag)
 	if err != nil {
+		if experimental.IsExperimentalModeEnabled() {
+			return []string{}, fmt.Errorf("component \"%s\" not found", componentType)
+		}
 		return []string{}, err
 	}
 	imageStreamImage, err := c.GetImageStreamImage(imageStream, imageTag)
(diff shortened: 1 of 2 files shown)


johnmcollier:
@kadel Any concerns with this approach? I'm open to alternatives 😄 

It's a little hacky, but I wanted to avoid breaking s2i-support when experimental mode is enabled and still maintain a _somewhat_ meaningful error message for both s2i and devfile components

pkg/occlient/occlient.go
```diff
@@ -753,6 +754,9 @@ func (c *Client) GetImageStream(imageNS string, imageName string, imageTag strin
 		}
 		if e != nil && err != nil {
 			// Imagestream not found in openshift and current namespaces
+			if experimental.IsExperimentalModeEnabled() {
+				return nil, fmt.Errorf("component \"%s\" not found", imageName)
```
dharmit, line 758: You could use `%q` to avoid escaping the `"` character.

johnmcollier: @dharmit good to know about %q, updated 🙂

dharmit:
```sh
$ odo create fakecomp 

Validation
 ✗  Checking devfile compatibility [19888ns]
 ✗  Creating a devfile component from registry:  [13492ns]

Please run `odo catalog list components` for a list of supported devfile component types
 ✗  component "fakecomp" not found
```
@johnmcollier should we start the validation stage for a component that's not found in the devfile registry? Shouldn't we handle it similarly for devfile component as well? Maybe even print only the last part, i.e.,:
```sh
$ odo create fakecomp 

 ✗  component "fakecomp" not found
```
By no means am I suggesting to address it in this PR. We can discuss this in a separate issue. The code you've added looks good to me! But, if it makes sense, I would open a separate issue to discuss the validation part.

/lgtm

PR #3282, opened 2020-06-02
Address review comments for devfile V2 PR: Part 2

adisky:
**What type of PR is this?**

 /kind code-refactoring

Fixes #?

**NOTE: PR for parser-v2 branch**

### DIFF
--- a/pkg/devfile/adapters/common/command.go
+++ b/pkg/devfile/adapters/common/command.go
@@ -42,11 +42,10 @@ func getCommand(data data.DevfileData, commandName string, groupType common.Devf
 				// e.g --build-command "mybuild"
 				// exec:
 				//   id: mybuild
-				// group:
-				//   kind: build
+				//   group:
+				//     kind: build
 				if command.Exec.Group.Kind != groupType {
-					return supportedCommand, fmt.Errorf("mismatched group kind, command %s is of group kind %v groupType in devfile", commandName, groupType)
-
+					return supportedCommand, fmt.Errorf("command group mismatched, command %s is of group %v in devfile.yaml", commandName, command.Exec.Group.Kind)
 				}
 				supportedCommand = command
 				return supportedCommand, nil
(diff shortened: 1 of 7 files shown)


pkg/devfile/adapters/common/command.go
```diff
@@ -30,54 +28,54 @@ func getCommand(data data.DevfileData, commandName string, groupType common.Devf
...
 
-		// if not command specified via flag, default command has the highest priority
+		// if no command specified via flag, default command has the highest priority
 		// We need to scan all the commands to find default command
-		if command.Exec.Group.Kind == groupType && command.Exec.Group.IsDefault {
-			supportedCommand = command
-			return supportedCommand, nil
+		// exec.Group is a pointer, to avoid null pointer
+		if command.Exec.Group != nil {
+			if command.Exec.Group.Kind == groupType && command.Exec.Group.IsDefault {
```
mik-dass, line 55: why not join the two lines using `&&`?

pkg/devfile/adapters/common/command.go
```diff
@@ -30,54 +28,54 @@ func getCommand(data data.DevfileData, commandName string, groupType common.Devf
...
 	}
 
 	if commandName == "" {
 		// if default command is not found return the first command found for the matching type.
 		for _, command := range commands {
-			if command.Exec.Group.Kind == groupType {
-				supportedCommand = command
-				return supportedCommand, nil
+			if command.Exec.Group != nil {
+				if command.Exec.Group.Kind == groupType {
```
mik-dass, line 66: why not join using `&&`?

mik-dass:
Some tests seem to be failing https://travis-ci.com/github/openshift/odo/jobs/342528540#L428

pkg/devfile/adapters/common/command.go
adisky: if group is `nil` and we try to retrieve `kind`, it would give `nil pointer dereference `

mik-dass: I guess it will short circuit if group is `nil` and it won't try to retrieve `Kind`. https://stackoverflow.com/questions/45698298/does-go-excecute-all-conditions-in-if-statement

adisky: ya you are right, good to know, but do you think it is good to change? as someday may forget to preserve the order.

mik-dass: You are right that somebody might forget to preserve the order, but we can have proper unit tests to prevent that scenario.

Issue #3280, opened 2020-06-02
Wrong error formatting in devfile


adisky:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**

## How did you run odo exactly?

```
$ odo push

Validation
 ✗  Validating the devfile [35201ns]
 ✗  Failed to start component with name nodejs.
Error: Failed to create the component: failed to validate devfile build and run commands: 
The command "devbuild" was found but its actions are not supported: Problem with command "devbuild" action #1: Actions must be of type "exec"%!!(MISSING)(EXTRA string=
)
```

## Actual behavior
if an invalid action is specified, the error is `Actions must be of type "exec"%!!(MISSING)(EXTRA string=)`, the error formatting is wrong, same applies to all other validation errors.

## Expected behavior
error should be `actions must be of type exec`

maysunfaisal:
can be closed, fixed by https://github.com/openshift/odo/pull/3291

Issue #3271, opened 2020-06-01
Throw Error, If multiple commands of same kind specified in devfile v2 and none is default


adisky:
/kind feature

## Which functionality do you think we should add?
Currently for devfile V2, we are allowing users to specify multiple commands of same kind, and selecting the first command from the list, if none of them is default.

We need to throw error for following case _"Multiple build command specified in devfile with no default"_

```
commands:
  - exec:
      id: build the app
      component: tools
      commandLine: "/artifacts/bin/build-container-full.sh"
      workingDir: /projects/springbootproject
      group:
        kind: build
  - exec:
      id: build
      component: runtime
      commandLine: "mvn compile"
      workingDir: /
      group:
        kind: build
```

girishramnani:
- multiple build commands is a valid use case
- but having no default build command should fail either way.

maysunfaisal:
I think this can be covered by issue https://github.com/openshift/odo/issues/2938
since I am working on #2938, I can take this issue
/assign

maysunfaisal:
closing, fixed by https://github.com/openshift/odo/pull/3291

Issue #3249, opened 2020-05-27
Improve performance for `odo catalog list components` .


adisky:
/kind feature
 
currently we are parsing all devfiles  from all registries during `odo catalog list components`,  We should list devfiles from index.json

adisky:
for V1 we are parsing devfile from each registry to this struct https://github.com/openshift/odo/blob/master/pkg/catalog/types.go#L45
and validating for supported component here
https://github.com/openshift/odo/blob/master/pkg/catalog/types.go#L45

IIUC We are parsing and validating each devfile from each registry before listing, We are parsing the devfile for two things
- Name
- Supported

is parsing each devfile really required?  

We can get name from `index.json` of a devfile registry.

We can decide supported on the basis of devfile registry
**odo specific registry** : Ideally if a devfile has been added to an odo specific registry it is supported, no need to validate with each call of `odo catalog list components`. The devfile should be validated and tested before adding it to the registry. So for odo specific registries(registry by odo team) we could default them to `true`.

**other registries** : other registries could be added by user, or some other open registries.
devfiles in those are not tested by odo, so we should not list them as `supported` and hence they can be defaulted to `false`

cc @kadel @elsony

kadel:
> is parsing each devfile really required?

We should avoid parsing each devfile from the registry. It is not a good long term solution and it will definitely create problems for even moderately large devfile registries.

elsony:
The `odo specific registry` is a point-in-time thing.  Our eventual goal is share the same devfile registry as Che (and other dev tools).  In that case, we don't have a registry that we know for sure we support all devfiles in there.  That leave us a couple of options:
1. Continue to scan for all devfiles in a repo: as @kadel  mentioned, it may causes performance problem on large registry.
2. Not scanning any at all and just assume all devfiles are supported. We just error out during the execution if someone tries to run do `odo push` on some devfiles that we cannot support, e.g. without any run command: I don't think this will provide a good user experience since it discover quite late on the cycle.
3. Inject some info into the index.json, e.g. an `executable` or `dev` tag, that indicates those devfiles contain the minimum that we can support.  We basically move the scanning from real time on all devfiles to one time on building the index.json. Once we have that, we can easily filter out just based on info within the index.json.

I think #3 is a good trade off to inject a build step on the repo for which we already have scripts to generate the index.json in our repo.

kadel:
> I think #3 is a good trade-off to inject a build step on the repo for which we already have scripts to generate the index.json in our repo.

[...]

PR #3291, opened 2020-06-03
Devfile v2 command group impl + Odo catalog list components

### LINKED ISSUE #2938: Support group in devfile commands (application lifecycle)

maysunfaisal:

**What type of PR is this?**
/kind feature

- Implements complete feature of devfile v2 spec for command groups cdc998972f657dcbc5333fcfdbeef487581e95b3
- Refactored and cleaned up code around devfile parser and validate to be consistent across odo commands a82ca625b691343cc37824e9a938669ccb545e24
- Update odo catalog to list components from new default devfile v2 registry 13d67fcef528742e7c592f09d33d0e68169fb7dc
- Update odo catalog list components to only read index.json entries for name instead of parsing the entire devfile. We also no longer show the `SUPPORTED` column for devfile catalog, see discussion at https://github.com/openshift/odo/issues/3249#issuecomment-641350713 065f19b3fc1f3e08cc1b146af239bc01559a10c9 9da66e99d07ae82b2679216ae466130c71996bdb
- Unit tests 065f19b3fc1f3e08cc1b146af239bc01559a10c9
- Fix odo catalog & odo create bugs 66deec8bfc8db2d38a1d036551d34d3ab78eb84f
- integration tests 6d2a0aa2c011649e84ad2f5ad91363d86b7e48ca

Fixes #2938 #3271 #3249 #3280 #3436

- `odo catalog list components` (now lists new v2 devfiles)
```
$ odo catalog list components
Odo OpenShift Components:
NAME              PROJECT       TAGS                        SUPPORTED
java              openshift     11,8,latest                 YES
nodejs            openshift     10-SCL,8,8-RHOAR,latest     YES
dotnet            openshift     2.1,2.2,3.0,latest          NO
golang            openshift     1.11.5,latest               NO
httpd             
Odo Devfile Components:
NAME            DESCRIPTION                            REGISTRY
maven           Upstream Maven and OpenJDK 11          DefaultDevfileRegistry
nodejs          Stack with NodeJS 10                   DefaultDevfileRegistry
openLiberty     Open Liberty microservice in Java      DefaultDevfileRegistry
quarkus         Upstream Quarkus with Java+GraalVM     DefaultDevfileRegistry
springBoot      Spring Boot® using Java                DefaultDevfileRegistry
```
- `odo create springBoot mondayspringboot`
- `odo url create`
- `odo push`

To test out the various push scenarios:

1. commands with multiple default will error out
2. commands with multiple default but with `odo push --build-command --run-command` should succeed
3. push without a run group should error out
4. push with flags using the wrong group should error out like `odo push --build-command="a run command of group run"`


maysunfaisal:
👋 

@adisky could you pls help review the command group code?

@GeekArthur could you pls help review the catalog and create code?

thx!

pkg/devfile/adapters/common/command.go
```diff
@@ -13,66 +13,70 @@ import (
...
-		if command.Exec.Group != nil && command.Exec.Group.Kind == groupType && command.Exec.Group.IsDefault {
-			supportedCommand = command
-			return supportedCommand, nil
+		if command.Exec.Group != nil && command.Exec.Group.Kind == groupType {
+			if command.Exec.Group.IsDefault {
+				// We need to scan all the commands to find default command
+				return command, validateCommand(data, command)
+			} else if reflect.DeepEqual(firstMatchedCommand, common.DevfileCommand{}) {
+				// store the first matched command in case there is no default command
+				firstMatchedCommand = command
```
adisky, line 63: should we rename it to `onlyCommand` or `singleCommand`? IIUC we are throwing error if more than one command of a grouptype is present and none of them is default and if only one command is present without any default value we are allowing it.  So it should be that one command

pkg/devfile/adapters/common/command.go
```diff
@@ -13,66 +13,70 @@ import (
...
 	if commandName == "" {
 		// if default command is not found return the first command found for the matching type.
-		for _, command := range commands {
-
-			if command.Exec.Group != nil && command.Exec.Group.Kind == groupType {
-				supportedCommand = command
-				return supportedCommand, nil
-			}
+		if !reflect.DeepEqual(firstMatchedCommand, common.DevfileCommand{}) {
+			return firstMatchedCommand, validateCommand(data, firstMatchedCommand)
```
adisky, line 71: IMO this is not required now, this block could be removed and we can return from above `else if` block

pkg/devfile/adapters/common/utils.go
```diff
@@ -109,6 +109,18 @@ func GetSupportedComponents(data data.DevfileData) []common.DevfileComponent {
 	return components
 }
 
+func getCommandsForGroup(data data.DevfileData, groupType common.DevfileCommandGroupType) []common.DevfileCommand {
+	var commands []common.DevfileCommand
```
adisky, line 113: nit: s/getCommandsForGroup/getCommandsByGroup

pkg/devfile/parser/data/2.0.0/components.go
```diff
@@ -11,7 +13,14 @@ func (d *Devfile200) GetComponents() []common.DevfileComponent {
 
 // GetCommands returns the slice of DevfileCommand objects parsed from the Devfile
 func (d *Devfile200) GetCommands() []common.DevfileCommand {
-	return d.Commands
+	var commands []common.DevfileCommand
+
+	for _, command := range d.Commands {
+		command.Exec.Id = strings.ToLower(command.Exec.Id)
```
adisky, line 19: why we need this? Please add comment

pkg/devfile/adapters/common/command.go
```diff
@@ -84,6 +88,33 @@ func getCommand(data data.DevfileData, commandName string, groupType common.Devf
...
+			}
+		}
+	} else {
+		// if there is only one command, it is the default command for the group
+		defaultCommandCount = 1
+	}
+
+	if defaultCommandCount != 1 {
+		return fmt.Errorf("there should be one default command for command group %v", groupType)
+	}
```
adisky, line 113: `there should be only one default command for command group`

`there should be one default command for command group` signifying that no default command is present.

adisky:
@maysunfaisal Overall looks good to me, tested for command scenarios working well.
Just left some nit's.

pkg/devfile/adapters/common/command.go
maysunfaisal: yeah we can rename it.

Actually there is a case where multiple defaults can be present ie when we use `odo push --build-command="mybuild1"`. `mybuild1` and `mybuild2` can be default or have no defaults at all and this is fine since we're explicitly using a flag and saying what command to run

pkg/devfile/adapters/common/command.go
maysunfaisal: updated

pkg/devfile/adapters/common/utils.go
maysunfaisal: updated

pkg/devfile/parser/data/2.0.0/components.go
maysunfaisal: updated

pkg/devfile/adapters/common/command.go
maysunfaisal: actually this condition checks for both missing and multiple defaults

pkg/devfile/adapters/common/command.go
maysunfaisal: Actually we need this. Prior to this PR, we used two for loops to get the default command and the first command of group type.

But now we're checking for both in the same for loop..

For example if a devfile has two build commands in order
- devbuild1 
- devbuild2 {default:true}

if we removed the if condition and returned in `else if`, it will incorrectly return devbuild1 even tho in the next iteration there is a default command.

But with this if block it solves the two cases ie multiple commands of same group in any order and also one command of a group kind.

pkg/catalog/catalog.go
```diff
@@ -188,43 +119,18 @@ func ListDevfileComponents(registryName string) (DevfileComponentTypeList, error
 		return *catalogDevfileList, err
 	}
 
-	// 1. Load each devfile concurrently from the previously retrieved devfile index entries
-	// 2. Populate devfile components with devfile data
-	// 3. Add devfile component types to the catalog devfile list
-	retrieveDevfiles := util.NewConcurrentTasks(len(registryIndices))
-	devfileMutex := &sync.Mutex{}
+	// Add stacks to the catalog devfile list
 	for _, index := range registryIndices {
```
GeekArthur, line 123: If we remove the code block of downloading the devfile for getting information of devfile support criterias, we don't need this for loop either. `DevfileComponentType` population step can be done inside function `getDevfileIndexEntries` when we download the index file. With this for loop removal, we can increase the performance a lot as we don't need to iterate through the all index entries again.

pkg/odo/cli/component/create.go
```diff
@@ -382,41 +382,40 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			defaultComponentNamespace = client.Namespace
 		}
 
-		var catalogDevfileList catalog.DevfileComponentTypeList
 		var componentType string
 		var componentName string
 		var componentNamespace string
+		isDevfileRegistryPresent := true // defaulted to true since odo ships with a default registry set
+
+		// Component type: We provide devfile component list to let user choose
```
GeekArthur, line 390: We can't make the function `catalog.ListDevfileComponents` global because we have the case where creating devfile component from existing devfile, that case doesn't need this function so that make this function global can reduce the performance. This function only use by creating devfile component from registry (either interactive or direct mode)

pkg/odo/cli/component/create.go
```diff
@@ -475,48 +470,49 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
-		// Set devfileMetadata struct
-		co.devfileMetadata.componentType = componentType
-		co.devfileMetadata.componentName = strings.ToLower(componentName)
-		co.devfileMetadata.componentNamespace = strings.ToLower(componentNamespace)
+		if isDevfileRegistryPresent {
```
GeekArthur, line 473: This `isDevfileRegistryPresent` flag should only apply to the case where creating devfile component from registry, which means if the registry list is empty user still can use the existing devfile to create devfile component. So `isDevfileRegistryPresent` flag should be used here https://github.com/openshift/odo/pull/3291/files#diff-58fe314b4e3f9120402717ab8c2b3049L512 instead of the whole code block.

GeekArthur:
The PR looks good, request some changes for performance improvement as `odo catalog` and `odo create` get involve in the http request/response (network is always an important factor that we should consider for performance).

Also leave couple of comments below:
1. Regarding `odo catalog` and `odo create`, do we consider backwards compatibility? If yes we should modify this PR as this PR breaks the `odo catalog list components` support for devfile 1.0 registry. If no we should provide migration instruction/step because users may hit the issue where the new odo release doesn't work on their devfile 1.0 registry.
2. If we decide to remove the preliminary check for devfile support criteria (devfile has run command and build command etc), IMO we should add the check somewhere back before `odo push` in the same PR, otherwise the simple devfile incompatibility error will out until user run `odo push`, that's not a good UX I think.

pkg/catalog/catalog.go
```diff
@@ -188,43 +119,18 @@ func ListDevfileComponents(registryName string) (DevfileComponentTypeList, error
...
 		// Load the devfile
```
GeekArthur, line 124: Nit: we can remove this comment as we don't download/load the devfile.

pkg/devfile/adapters/common/command.go
adisky: IMO, we could return different error for  `defaultCommandCount=0` & `defaultCommandCount >1`
or 
`there should be exactly one default command for command group `

pkg/devfile/adapters/common/command.go
adisky: but that case we are already covering above if commandName != "" , IIUC  that case would not fallback here, also can we update the comment 	`// store the first matched command in case there is no default command` to `return the only remaining command for the group`

adisky: Not necessary to do in this PR, but should we separate the flag command and devfile command function to better code readability.
`getCommandFromFlags`
`getCommandfromDevfile`
and in the `GetRunCommand`, `GetBuildCommand` functions includes the check for command name?
IMO this one(what i had written also) is difficult to understand.

kadel:
> 1. Regarding `odo catalog` and `odo create`, do we consider backwards compatibility? If yes we should modify this PR as this PR breaks the `odo catalog list components` support for devfile 1.0 registry. If no we should provide migration instruction/step because users may hit the issue where the new odo release doesn't work on their devfile 1.0 registry.

I don't think that we need to deal with backward compatibility in this case. All this is still done in the experimental mode. The main reason to introduce experimental mode was to allow us to experiment and break thinks if needed without dealing with backward compatibility issues. 
If we think that someone already has its own devfile v1 registries we can document migration steps, but I wouldn't do more than that. We will drop whole devfile v1 support before devfile goes out of experimental anyway.


> 2\. If we decide to remove the preliminary check for devfile support criteria (devfile has run command and build command etc), IMO we should add the check somewhere back before `odo push` in the same PR, otherwise, the simple devfile incompatibility error will out until user run `odo push`, that's not a good UX I think.

I don't think that we need to check for support criteria before `odo push`.
With this PR we are just using our own devfile registry that is under our full control so we can make sure that all devfiles in our registry are valid.
If someone adds their own registry it is their responsibility.

pkg/devfile/adapters/common/command.go
maysunfaisal: > but that case we are already covering above if commandName != "" , IIUC that case would not fallback here, also can we update the comment // store the first matched command in case there is no default command to return the only remaining command for the group

you're right, i have updated the comment to 
```
// return the only remaining command for the group if there is no default command
// NOTE: we return outside the for loop since the next iteration can have a default command
```

> Not necessary to do in this PR, but should we separate the flag command and devfile command function for better code readability?

yeah i think it makes sense to have different funcs and separate out their logic for better readability. I think anyone else reading it will be confused in the first few attempts.

pkg/devfile/adapters/common/command.go
maysunfaisal: Yeah makes sense, updated the err statements based on the default command count

pkg/devfile/adapters/common/command.go
maysunfaisal: I've updated the getCommand() func. Broken down into getCommandFromFlag() and getCommandfromDevfile() and moved their specific logic to this func.

getCommand() does the check for commandName and splits into either of the above two function calls. This way we can just use getCommand() from anywhere instead of doing this check everywhere when we need a command.

adisky: @maysunfaisal Thanks for doing this, now it is much easier to understand :)

adisky:
looks good to me, leaving lgtm for @GeekArthur

pkg/catalog/catalog.go
maysunfaisal: good catch, updated

pkg/odo/cli/component/create.go
maysunfaisal: updated

pkg/odo/cli/component/create.go
maysunfaisal: actually it doesnt matter because `isDevfileRegistryPresent` is true by default but i moved it because local devfile is not devfile registry. So later when ppl comes to read the code they dont get confused,

But I had to move the header `Validation` to inside the if blocks to keep interactive mode clean.

pkg/catalog/catalog.go
maysunfaisal: updated

maysunfaisal:
@GeekArthur addressed your PR feedback, can be reviewed again, thx.

GeekArthur:
> @GeekArthur addressed your PR feedback, can be reviewed again, thx.
> 
> Also fixed another bug in interactive mode for Docker - it asked for namespace [0ae5d5f#diff-58fe314b4e3f9120402717ab8c2b3049R417](https://github.com/openshift/odo/commit/0ae5d5fef6209f035e2316cf2691e1f96aac0fe0#diff-58fe314b4e3f9120402717ab8c2b3049R417).

Shouldn't the issue of prompt namespace for docker case fix by this PR https://github.com/openshift/odo/pull/3248 for this issue https://github.com/openshift/odo/issues/2978?

maysunfaisal:
> Shouldn't the issue of prompt namespace for docker pushTarget fix by this PR #3248 for this issue #2978?

seems like it got removed between PRs @GeekArthur

GeekArthur:
> > Shouldn't the issue of prompt namespace for docker pushTarget fix by this PR #3248 for this issue #2978?
> 
> seems like it got removed between PRs @GeekArthur

I think it's not the first time we hit this kind of issue, the root cause is the PR is not rebased properly before merging, to fix this we can:
1. Rebase properly before merging.
2. Increase the test coverage to make the test stronger, I believe we didn't catch that fix is removed by another PR is because currently we don't have interactive mode test.

Also thanks for the changes, I can review the PR today or tomorrow.

pkg/odo/cli/component/create.go
```diff
@@ -480,10 +482,10 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		co.devfileMetadata.componentName = strings.ToLower(componentName)
 		co.devfileMetadata.componentNamespace = strings.ToLower(componentNamespace)
 
-		// Categorize the sections
-		log.Info("Validation")
-
 		if util.CheckPathExists(DevfilePath) || co.devfileMetadata.devfilePath.value != "" {
+			// Categorize the sections
+			log.Info("Validation")
```
GeekArthur, line 487: Why we add this message in each code block? If the two blocks are mutual exclusive then we should add the message globally, I test both direct mode and interactive mode it works fine for me.

pkg/odo/cli/component/create.go
```diff
@@ -509,14 +511,19 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 
 			return nil
-		} else {
+		}
+
+		if isDevfileRegistryPresent {
+			// Categorize the sections
+			log.Info("Validation")
```
GeekArthur, line 518: Why we add this message in each code block, if the two blocks are mutual exclusive then we should add the message globally, I test both direct mode and interactive mode it works fine for me.

adisky:
@maysunfaisal This would also get fixed by your PR?
https://github.com/openshift/odo/issues/3280

pkg/odo/cli/component/create.go
maysunfaisal: discussed with Jingfu, more info here https://github.com/openshift/odo/pull/3291#discussion_r442324400

pkg/odo/cli/component/create.go
maysunfaisal: discussed with Jingfu, more info here https://github.com/openshift/odo/pull/3291#discussion_r442324400

maysunfaisal:
> @maysunfaisal This would also get fixed by your PR?
> #3280

@adisky yes this has been fixed with this PR. Just moved the formatting around to fix it - https://github.com/openshift/odo/pull/3291/files#diff-ca003cd73cab471abb4c9aca10266a96R231

pkg/odo/cli/component/create.go
```diff
@@ -454,12 +454,14 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 					componentName = args[0]
 				}
 
-				// Get available devfile components for checking devfile compatibility
+				// Component type: We provide devfile component list to let user choose
```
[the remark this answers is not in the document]
maysunfaisal: updated

pkg/catalog/catalog.go
```diff
@@ -81,72 +73,21 @@ func getDevfileIndexEntries(registry Registry) ([]DevfileIndexEntry, error) {
 		return nil, errors.Wrapf(err, "Unable to unmarshal the devfile index.json from %s", indexLink)
 	}
 
+	var catalogRegistryDevfiles []DevfileComponentType
+
 	for i := range devfileIndex {
```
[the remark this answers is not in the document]
maysunfaisal: updated

pkg/catalog/catalog.go
```diff
@@ -68,7 +60,7 @@ func GetDevfileRegistries(registryName string) (map[string]Registry, error) {
 const indexPath = "/devfiles/index.json"
 
 // getDevfileIndexEntries retrieves the devfile entries associated with the specified registry
-func getDevfileIndexEntries(registry Registry) ([]DevfileIndexEntry, error) {
+func getDevfileIndexEntries(registry Registry) ([]DevfileComponentType, error) {
```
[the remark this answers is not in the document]
maysunfaisal: updated

adisky:
@maysunfaisal please update the PR description to close this bug, when this PR gets merged

GeekArthur:
Add `lgtm` back after rebase
/lgtm

johnmcollier:
@maysunfaisal Trying out your branch, it seems interactive mode is broken (but with a different issue than https://github.com/openshift/odo/issues/3436):
Screen Shot 2020-06-26 at 5 53 44 PM

maysunfaisal:
@johnmcollier you need to delete `~/.odo`, it is still pointing to v1 devfiles. v2 registry devfile index has changed and has the name and hence i think its showing empty names.

johnmcollier:
@maysunfaisal Ah, good catch. Yeah, that fixed things.

We should maybe send a note out on #odo-dev then, if `~/.odo` needs to be deleted

Issue #2935, opened 2020-04-20
Specify the source mapping path for component containers


elsony:
/kind user-story

## User Story
As a stack creator, I want to customize the source mapping path on the containers defined in the devfile. 

## Acceptance Criteria 
- [ ] It should support consuming source mapping path defined in the corresponding devfile 2.0 item: https://github.com/che-incubator/devworkspace-api/issues/17

## Links 
- Related issue: https://github.com/che-incubator/devworkspace-api/issues/17

/kind user-story
/area devfile

kadel:
/status blocked
#2918 we need devfile v2 parser first

PR #3295, opened 2020-06-03
Add support for SourceMapping field in devfiles

johnmcollier:
**What type of PR is this?**
/kind feature
/area devfileV2

This PR adds support for the new `SourceMapping` field in devfile v2. When it is set, odo will mount the project source volume to that folder, rather than `/projects` or `/projects/<projectName>`.

Fixes https://github.com/openshift/odo/issues/2935

Use this sample V2 devfile and verify that it syncs to `/test`:
```
schemaVersion: 2.0.0
metadata:
  name: nodejs
  version: 1.0.0
projects:
  - name: nodejs-web-app
    git:
      location: "https://github.com/odo-devfiles/nodejs-ex.git"
components:
  - container:
      name: runtime
      image: quay.io/eclipse/che-nodejs10-ubi:nightly
      memoryLimit: 1024Mi
      mountSources: true
      sourceMapping: /test
      endpoints:
        - name: "nodejs"
          configuration:
            discoverable: false
            public: true
            protocol: tcp
            scheme: http
          targetPort: 3000
commands:
  - exec:
      id: devBuild
      component: runtime
      commandLine: npm install
      workingDir: /test/app
      group:
        kind: build
        isDefault: true
  - exec:
      id: devRun
      component: runtime
      commandLine: nodemon app.js
      workingDir: /test/app
      group:
        kind: run
        isDefault: true
```

I've also updated the odo push integration tests for Kube and Docker to test this scenario.


johnmcollier:
Removing WIP tag now that integration tests have been added

tests/examples/source/devfilesV2/nodejs/devfileSourceMapping.yaml
```diff
@@ -0,0 +1,40 @@
+schemaVersion: 2.0.0
+metadata:
+  name: nodejs
```
prietyc123, line 3: I can see only devfileSourceMapping.yaml for `nodejs` component type. So i think this pr only handles for `nodejs` type and later other supported devfile component type will be added, right ?

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -83,6 +83,31 @@ var _ = Describe("odo devfile push command tests", func() {
 			helper.CmdShouldPass("odo", "push", "--devfile", "devfile.yaml", "--project", namespace)
 		})
 
+		It("checks that odo push works with a devfile with sourcemapping set", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, cmpName)
```
prietyc123, line 87: I am not confused why we are only looking for nodejs only? Why not `springboot` ?

tests/examples/source/devfilesV2/nodejs/devfileSourceMapping.yaml
johnmcollier: The sample devfile I used is nodejs, but the implementation of `sourceMapping` isn't specific to any devfile or runtime, so other devfiles don't need to be added, as it would be testing the same functionality.

tests/integration/devfile/cmd_devfile_push_test.go
johnmcollier: This is required since `--devfile` isn't supported in `odo create` (see https://github.com/openshift/odo/pull/3118), so this is a necessary step for all tests for devfiles currently.

maysunfaisal:
readd lgtm after rebase
/lgtm

maysunfaisal:
readd lgtm after fixing tests
/lgtm

johnmcollier:
Ready for approval review

/assign @dharmit

dharmit:
@johnmcollier things work as per the PR description. However, I have a concern that might be relevant for this PR or a different one or entirely irrelevant.

After a successful `odo push`, should the user be able to exec directly to the directory provided by `sourceMapping` field instead of being exec'd into `/projects`?

```sh
# what I see
$ kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
/projects

# what I expect as a user
kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
/test
```

Of course, there could be a valid reason to not do so and I'm fine by it.

cc @elsony @kadel

adisky:
> @johnmcollier things work as per the PR description. However, I have a concern that might be relevant for this PR or a different one or entirely irrelevant.
> 
> After a successful `odo push`, should the user be able to exec directly to the directory provided by `sourceMapping` field instead of being exec'd into `/projects`?
> 
> ```shell
> # what I see
> $ kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
> /projects
> 
> # what I expect as a user
> $ kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
> /test
> ```
> 
> Of course, there could be a valid reason to not do so and I'm fine by it.
> 
> cc @elsony @kadel

@dharmit The directory for exec commands is set while creating the image, for most che images it is set to /projects.
https://docs.docker.com/engine/reference/commandline/exec/. so when we `kubectl exec` it defaults to the `WORKDIR` while creating the image.I think we cannot change it, for docker we can pass  `--workdir` while docker exec, for kubectl i guess it is not there.

girishramnani:
> > @johnmcollier things work as per the PR description. However, I have a concern that might be relevant for this PR or a different one or entirely irrelevant.
> > After a successful `odo push`, should the user be able to exec directly to the directory provided by `sourceMapping` field instead of being exec'd into `/projects`?
> > ```shell
> > # what I see
> > $ kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
> > /projects
> > 
> > # what I expect as a user
> > $ kubectl exec nodejs-7d6bb4dc94-c84xh -c runtime -- pwd
> > /test
> > ```
> > 
> > 
> > Of course, there could be a valid reason to not do so and I'm fine by it.
> > cc @elsony @kadel
> 
> @dharmit The directory for exec commands is set while creating the image, for most che images it is set to /projects.
> https://docs.docker.com/engine/reference/commandline/exec/. so when we do `kubectl exec` it defaults to the `WORKDIR` set while creating the image.I think we cannot change it, for docker we can pass `--workdir` while docker exec, for kubectl i guess it is not there.

I feel this could be done in a follow up PR is need be.
cc @dharmit

girishramnani:
@dharmit opened a new issue for this discussion https://github.com/openshift/odo/issues/3472

girishramnani:
/lgtm 
As the code is fine and tests got resolved too

PR #3305, opened 2020-06-04
url list routes and ingress 

### LINKED ISSUE #2985: List routes kind URLs in odo list
/kind bug

Once https://github.com/openshift/odo/pull/2885 is in, we should list route based URLs in `odo list`.

blocked on https://github.com/openshift/odo/pull/2885

### LINKED ISSUE #3182: URL describe under experimental mode should work for both routes and ingresses
/kind bug

This bug is due to the new change for adding support for routes under experimental mode. 
The current logic for `list` and `describe` only search for ingresses if experimental mode is turned on. An expected bug, need to modify the logic. 
  
With experimental mode set, create a route UR
Run `url describe <route url>`

## Actual behavior
`url describe <route url>` complains the URL does not exist

## Expected behavior
`url describe` should be able to look for routes as well in experimental mode

### LINKED ISSUE #3274: odo describe doesn't print secure and URL properly
/kind bug

## Actual behavior

```
[mrinaldas@localhost che-nodejs]$ odo url create example
[mrinaldas@localhost che-nodejs]$ odo push
[mrinaldas@localhost che-nodejs]$ odo url describe example
NAME        STATE      URL     PORT
example     Pushed     ://     8080
```

## Expected behavior

`odo describe` should display information about the URL being secure or not. Also after push, the proper URL should be displayed instead of `://`

### LINKED ISSUE #3304: secured odo url get re-created during push even with no changes to it
/kind bug
/area url

create a secured url and run `odo push`
run `odo push` again

## Actual behavior
The url get re-created even with no changes to it
```
Applying URL changes
 ✓  URL ingressurl2 successfully deleted
 ✓  URL ingressurl2: https://ingressurl2.1.2.3.4.com created
```

## Expected behavior
```
Applying URL changes
```

This is because the `secure` value is missing when combining routes and ingresses into a single list:
https://github.com/openshift/odo/blob/ddc2a204674e0e1ef0749802e5e6b3c5ee04abfd/pkg/url/url.go#L867-L872
https://github.com/openshift/odo/blob/ddc2a204674e0e1ef0749802e5e6b3c5ee04abfd/pkg/url/url.go#L883-L887

Which result in the spec difference, so the code think the config has been changed. 
```
val.Specs: {ingressurl2.1.2.3.4.com  3000 true ingress  0}
urlSpec.Spec: {ingressurl2.1.2.3.4.com  3000 false ingress  0} ✓
```

yangcao77:
Signed-off-by: Stephanie <stephanie.cao@ibm.com>

**What type of PR is this?**
/kind bug
/area url
/area devfile

Since experimental mode now supports both ingress and route as url, we should show both types of URL created under experimental mode when running `odo url list`
Also `odo url describe` should be able to get the route URL information

```
$ odo url list
Found the following URLs for component nodejsproj
NAME            STATE               URL                                          PORT     SECURE     KIND
ingressurl2     Pushed              https://ingressurl2.1.2.3.4.com              3000     true       ingress
testroute1      Locally Deleted     http://testroute1-che.9.42.26.107.nip.io     3000     false      route
routeurl1       Pushed              http://routeurl1-che.9.42.26.107.nip.io      3000     false      route
```

```
$ odo url describe routeurl2
NAME          STATE      URL                                          PORT     SECURE     KIND
routeurl1     Pushed     http://routeurl1-che.9.42.26.107.nip.io      3000     false       route
```

Fixes #2985
Fixes #3304
Fixes #3274 
Fixes #3182

Automation test:
Simply run ` make test-cmd-devfile-url`

Manual test:
Enable experimental mode `odo preference set experimental true`
And create both route URL and ingress URL
```
odo url create route1
odo url create ingress1 --host 1.2.3.4.com --ingress
```

Run `url list` and `url describe` to check the route URL information is displayed correctly.


pkg/url/url.go
```diff
@@ -85,39 +86,53 @@ func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName s
...
-	ingress, err := kClient.GetIngress(urlName)
-	if kerrors.IsNotFound(err) {
+	ingress, getIngressErr := kClient.GetIngress(urlName)
+	// Check whether remote already created the route
+	route, getRouteErr := client.GetRoute(urlName)
+	if kerrors.IsNotFound(getIngressErr) && kerrors.IsNotFound(getRouteErr) {
 		remoteExist = false
-	} else if err != nil {
-		return URL{}, errors.Wrap(err, "unable to get ingress on cluster")
+	} else if getIngressErr != nil && getRouteErr != nil {
```
mik-dass, line 98: Shouldn't this be a `or` condition? If either of the them is an actual error, then we should immediately error out rather than proceeding forward.

pkg/url/url.go
```diff
@@ -85,39 +86,53 @@ func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName s
...
-func GetIngress(kClient *kclient.Client, envSpecificInfo *envinfo.EnvSpecificInfo, urlName string, componentName string) (URL, error) {
+// GetIngressOrRoute returns ingress/route spec for given URL name
+func GetIngressOrRoute(client *occlient.Client, kClient *kclient.Client, envSpecificInfo *envinfo.EnvSpecificInfo, urlName string, componentName string) (URL, error) {
 	remoteExist := true
 	// Check whether remote already created the ingress
-	ingress, err := kClient.GetIngress(urlName)
-	if kerrors.IsNotFound(err) {
+	ingress, getIngressErr := kClient.GetIngress(urlName)
+	// Check whether remote already created the route
+	route, getRouteErr := client.GetRoute(urlName)
```
mik-dass, line 95: We need to check if route resource type is available on the cluster before this action.

mik-dass, line 95: Do we need to check for the route if the url is already found after the ingress get action?

pkg/testingutil/routes.go
```diff
@@ -31,6 +32,9 @@ func GetSingleRoute(urlName string, port int, componentName, applicationName str
 				applabels.OdoVersion:           version.VERSION,
 				labels.URLLabel:                urlName,
 			},
+			Annotations: map[string]string{
+				"openshift.io/host.generated": "true",
+			},
```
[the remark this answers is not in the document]
mik-dass: We don't need this. For OpenShift clusters, creation of an ingress also creates a route. So the list of routes, also contains the routes belonging to these ingresses. But for these routes, the owner reference is set to their belonging ingresses. So we can use the owner reference to distinguish these routes. We are already doing this in the ListPushed function
https://github.com/openshift/odo/blob/8faf7e5c998344938524ef6970d3dbe3bec58c6f/pkg/url/url.go#L357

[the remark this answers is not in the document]
yangcao77: You are right, I checked the route created for ingress, the `OwnerReference[0].Kind` is set to `Ingress`. I have updated the PR to use the OwnerReference instead of the annotation.

pkg/url/url.go
```diff
@@ -484,9 +531,23 @@ func ListIngressURL(client *kclient.Client, envSpecificInfo *envinfo.EnvSpecific
...
+			// URL is in both local env file and cluster
+			routeURL.Status.State = StateTypePushed
+			urls = append(urls, routeURL)
+		} else {
+			// URL is on the cluster but not in local env file
+			routeURL.Status.State = StateTypeLocallyDeleted
+			urls = append(urls, routeURL)
+		}
+	}
+
```
mik-dass, line 546: These actions seem to be similar to the ingressMap. Can we merge the two into a single map?

tests/integration/cmd_url_test.go
```diff
@@ -121,14 +121,16 @@ var _ = Describe("odo url command tests", func() {
...
+			desiredURLDescribeJSON := fmt.Sprintf(`{"kind":"url","apiVersion":"odo.dev/v1alpha1","metadata":{ "name": "myurl","creationTimestamp": null},"spec":{"port": 8080,"secure": true,"kind": "route"},"status": {"state": "Not Pushed"}}`)
 			Expect(desiredURLDescribeJSON).Should(MatchJSON(actualURLDescribeJSON))
 
 			helper.CmdShouldPass("odo", "push", "--context", context)
 
 			// odo url describe -o json
 			actualURLDescribeJSON = helper.CmdShouldPass("odo", "url", "describe", "myurl", "-o", "json", "--context", context)
-			desiredURLDescribeJSON = fmt.Sprintf(`{"kind":"url","apiVersion":"odo.dev/v1alpha1","metadata":{ "name": "myurl","creationTimestamp": null},"spec":{ "port": 8080,"secure": true},"status": {"state": "Pushed"}}`)
+			fullURLPath := helper.DetermineRouteURL(context)
+			pathNoHTTP := strings.Split(fullURLPath, "//")[1]
```
mik-dass, line 132: can you please add a comment regarding this?

yangcao77: Sucre, comment added. 
It's to get the route URL, and actually this is exactly same code as 
https://github.com/openshift/odo/blob/3154f6010d9dc5f7dba66be0a81ddbd1c530023f/tests/integration/cmd_url_test.go#L152
and 
https://github.com/openshift/odo/blob/3154f6010d9dc5f7dba66be0a81ddbd1c530023f/tests/integration/cmd_url_test.go#L169-L170

pkg/url/url.go
yangcao77: you are right. Updated.

tests/integration/cmd_url_test.go
mik-dass: Thanks :+1:

mik-dass:
It seems on minikube, `describe` for a non pushed URL doesn't report the state properly.
```
[mrinaldas@localhost che-nodejs]$ odo url create example --host com
 ✓  URL example created for component: nodejs

[mrinaldas@localhost che-nodejs]$ odo url describe example
NAME        STATE     URL                    PORT     SECURE     KIND
example               http://example.com     3000     false      ingress
```

pkg/url/url.go
```diff
@@ -85,39 +86,68 @@ func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName s
...
+	var getRouteErr error
 	// Check whether remote already created the ingress
-	ingress, err := kClient.GetIngress(urlName)
-	if kerrors.IsNotFound(err) {
+	ingress, getIngressErr := kClient.GetIngress(urlName)
+	if kerrors.IsNotFound(getIngressErr) && routeSupported {
+		// Check whether remote already created the route
+		route, getRouteErr = client.GetRoute(urlName)
+	}
+	if kerrors.IsNotFound(getIngressErr) && kerrors.IsNotFound(getRouteErr) {
```
mik-dass, line 101: This will fail when routeSupported is false as `getRouteErr` is nil

mik-dass:
Also please add some integration tests for checking `secure:false` and `secure:true` in `odo url describe` for normal non JSON output.

pkg/url/url.go
yangcao77: Thanks for catching this bug. 
Updated the condition, and also add unit tests for describing ingresses if route is not supported.

mik-dass: Cool :+1:

pkg/testingutil/routes.go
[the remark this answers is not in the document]
mik-dass: OK :+1:

pkg/url/url.go
```diff
@@ -85,39 +86,53 @@ func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName s
...
-	if kerrors.IsNotFound(err) {
+	ingress, getIngressErr := kClient.GetIngress(urlName)
+	// Check whether remote already created the route
+	route, getRouteErr := client.GetRoute(urlName)
+	if kerrors.IsNotFound(getIngressErr) && kerrors.IsNotFound(getRouteErr) {
 		remoteExist = false
-	} else if err != nil {
-		return URL{}, errors.Wrap(err, "unable to get ingress on cluster")
+	} else if getIngressErr != nil && getRouteErr != nil {
+		return URL{}, errors.Wrap(getIngressErr, "unable to get URL on cluster")
```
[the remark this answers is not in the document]
mik-dass: :+1:

pkg/util/util.go
```diff
@@ -84,7 +84,7 @@ func GenerateRandomString(n int) string {
 	rand.Seed(time.Now().UnixNano())
 	b := make([]rune, n)
 	for i := range b {
-		b[i] = letterRunes[rand.Intn(len(letterRunes))]
+		b[i] = letterRunes[rand.Intn(len(letterRunes))] // #nosec G404
```
mik-dass, line 87: once https://github.com/openshift/odo/pull/3358 is in you probably will have to rebase

yangcao77: #3358 has merged, and I've rebased from master

mik-dass: Cool :+1:

Issue #2828, opened 2020-04-06
Unsafe data race in ExecuteCommand in exec.go


jgwest:
/kind bug

In `ExecuteCommand(...)` of `pkg/exec/exec.go` ([link](https://github.com/openshift/odo/blob/master/pkg/exec/exec.go#L39)), if an error occurs when executing a command within a container, the `cmdOutput += fmt.Sprintln(line)` may write to `cmdOutput` at the same point that the `log.ErrorF(...)` is reading from it.

Introduced with [this commit](https://github.com/openshift/odo/commit/5622eb92ba3ece4d42ddd8c4bb00cf6f0b8c8ef0) in 2018.

Race-instrumented build output for the errant lines:
```
WARNING: DATA RACE
Read at 0x00c0009d2f90 by main goroutine:
  github.com/openshift/odo/pkg/exec.ExecuteCommand()
      C:/Users/JONATHANWest/go/src/github.com/openshift/odo/pkg/exec/exec.go:44 +0x4fd
  github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component.Adapter.execDevfile()
      C:/Users/JONATHANWest/go/src/github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component/adapter.go:475 +0x407
  github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component.Adapter.Push()
      C:/Users/JONATHANWest/go/src/github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component/adapter.go:159 +0x89b
  github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component.(*Adapter).Push()
      <autogenerated>:1 +0x12a
  github.com/openshift/odo/pkg/devfile/adapters/kubernetes.Adapter.Push()
      C:/Users/JONATHANWest/go/src/github.com/openshift/odo/pkg/devfile/adapters/kubernetes/adapter.go:32 +0xa6
  github.com/openshift/odo/pkg/devfile/adapters/kubernetes.(*Adapter).Push()
      <autogenerated>:1 +0xd8
  github.com/openshift/odo/pkg/odo/cli/component.(*PushOptions).DevfilePush()
      C:/Users/JONATHANWest/go/src/github.com/openshift/odo/pkg/odo/cli/component/devfile.go:97 +0xab1
  github.com/openshift/odo/pkg/odo/cli/component.(*PushOptions).Run()
[...]

PR #3310, opened 2020-06-05
Fix unsafe data race in ExecuteCommand

cdrage:
**What type of PR is this?**

/kind bug

Fixes a potential flake with regards to when writing to cmdOutput at the
same time that log.ErrorF(...) is reading from it.

Closes https://github.com/openshift/odo/issues/2828

N/A, honestly not too sure

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
 	"k8s.io/klog"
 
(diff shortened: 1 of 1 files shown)


cdrage:
I know there's no tests in this PR. I'm working on it, but this will most likely have to go in without tests since it's so hard to capture StdOut correctly in the golang testing suite.

pkg/exec/exec.go
```diff
@@ -21,7 +26,7 @@ func ExecuteCommand(client ExecClient, compInfo common.ComponentInfo, command []
 	reader, writer := io.Pipe()
 	var cmdOutput string
```
girishramnani, line 27: The conventions for mutexes are that they are kept above the things they are guarding. Now you can create a struct called `SafeString` which holds the mutex and the string or as the mutex is not accessed externally just keep it above `cmdOutput`.

Issue #3307, opened 2020-06-05
component watch stuck over multiple file changes in a single shoot


amitkrout:
/kind bug

## What versions of software are you using?

**Operating System:**
CI
**Output of `odo version`:**
master
## How did you run odo exactly?
https://github.com/openshift/odo/issues/3307#issuecomment-649487799

## Actual behavior
watch stuck  

## Expected behavior
watch should push the changes

amitkrout:
It seems the issue is with pushing files steps 
```
$ odo watch nodejs-app --context tests/examples/source/nodejs/
[...]
Pushing files...
 ✓  Waiting for component to start [629ms]
 ✓  Syncing files to the component [4s]
 ✓  Building component [5s]
[...]
```
I think it stuck in any one of these steps may be in the step ```Waiting for component to start``` or ```Building component```. I need to figure out if this is the case then why the function [WatchNonRetCmdStdOut](https://github.com/openshift/odo/blob/master/tests/helper/helper_generic.go#L117) does not through any error. 

Cluster load increases with consistently pushing and build files due to the multiple changes (file/dir create , delete and update in server.js file) in the context dir causing the issue. Just to confirm it i am going to keep the server.js file change in the first place for watch command.

dharmit:
To be split into two. @amitkrout to open another issue where we'll need dev inputs.

kadel:
@amitkrout Can you please provide steps how it can be reproduced outside of our tests?

amitkrout:
> @amitkrout Can you please provide steps how it can be reproduced outside of our tests?

@kadel 
Steps to reproduce:
```
$ odo project create test123

$ oc --request-timeout 5m import-image nodejs --namespace=test123 --from=registry.access.redhat.com/rhscl/nodejs-10-rhel7:latest --confirm

$ oc annotate istag/nodejs:latest --namespace=test123 tags=builder --overwrite

$ odo create nodejs nodejs-app --project test123 --context tests/examples/source/nodejs/

$ odo push --context tests/examples/source/nodejs/

$ odo url create --port 8080 --context tests/examples/source/nodejs/

$ odo push --context tests/examples/source/nodejs/

$ curl -ik http://nodejs-app-8080-app-test123.apps.gamore46installer25thjune.devcluster.openshift.com
// Hello world from node.js! 

// Update the message in the file https://github.com/openshift/odo/blob/master/tests/examples/source/nodejs/server.js#L12

$ odo push --context tests/examples/source/nodejs/

[...]

PR #3315, opened 2020-06-07
Added few supported container images to e2e image test

amitkrout:
**What type of PR is this?**
/kind feature

Added few supported container images to e2e image test. See list supported container images in https://github.com/openshift/odo-init-image/blob/master/language-scripts/image-mappings.json 

Fixes NA, I am sending the pr directly.

```make test-e2e-images```

### DIFF
--- a/tests/e2escenarios/e2e_images_test.go
+++ b/tests/e2escenarios/e2e_images_test.go
@@ -1,3 +1,5 @@
+// This test file verifies all the supported container images listed in the
+// file https://github.com/openshift/odo-init-image/blob/master/language-scripts/image-mappings.json
 package e2escenarios
 
 import (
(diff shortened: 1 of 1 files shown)


amitkrout:
prow issue
/retest

amitkrout:
> Maybe it'd be more interesting to retrieve the list, parse it and check images instead of harcoding the references and iterating over them manually?

@metacosm This is a good approach but currently we are hardcoding it because we do not have such command or sub command that can retrieve the supported container images for odo from the concerned registry. @girishramnani @kadel Can you please share your thoughts ?

metacosm:
> > Maybe it'd be more interesting to retrieve the list, parse it and check images instead of harcoding the references and iterating over them manually?
> 
> @metacosm This is a good approach but currently we are hardcoding it because we do not have such command or sub command that can retrieve the supported container images for odo from the concerned registry. @girishramnani @kadel Can you please share your thoughts ?

My point, though, was that the test would retrieve the list outside of odo and then tests the images from that list, not rely on an odo command, which, by essence, might be bugged and shouldn't be relied on in a test anyway. Though, to be fair, the list retrieval code could also be bugged… Test the tests! 😉

girishramnani:
I would find it interesting if we can parse the json output of `odo catalog list components` and then use each of the versions programmatically for testing. This would help in keeping a better api contract with json output users that “whatever we have given you as json output - all those images work”.

This will also allow us to track and detect an unsupported image that lands on the catalog.

girishramnani:
But not sure this should be done as part of this PR

amitkrout:
> My point, though, was that the test would retrieve the list outside of odo and then tests the images from that list, not rely on an odo command, which, by essence, might be bugged and shouldn't be relied on in a test anyway. Though, to be fair, the list retrieval code could also be bugged… Test the tests! 😉

@metacosm Then all the time we need a human effort to monitor the script and as well as odo repo for any changes with respect to supported images. That is why i was looking for a odo specific command that tells us the supported images. As @girishramnani mentioned https://github.com/openshift/odo/pull/3315#issuecomment-641838982 we can parse the catalog list for supported images to avoid human monitoring.

> I would find it interesting if we can parse the json output of odo catalog list components and then use each of the versions programmatically for testing. This would help in keeping a better api contract with json output users that “whatever we have given you as json output - all those images work”.
This will also allow us to track and detect an unsupported image that lands on the catalog.

+1, @girishramnani Here i can see the list of supported container images - https://github.com/openshift/odo-init-image/blob/master/language-scripts/image-mappings.json but in ```odo catalog list components -o json``` i don't see these container images are tagged in the output list. BTW am i looking into the right place for supported container images ?

Question - When we run ```odo component create nodejs -- context <context>``` and then ```odo push -- context <context>``` then which nodejs container image is used to build the nodejs app ? 

> But not sure this should be done as part of this PR

Yes, this will be part of a separate effort.

girishramnani:
Seems like a good cabal topic @amitkrout to discuss not hard coding the images. 
For now this looks ok
/approve

girishramnani:
Do open an issue to track the "scripted approach to testing images"

prietyc123:
> When we run odo component create nodejs -- context <context> and then odo push -- context <context> then which nodejs container image is used to build the nodejs app ?

AFAIK it will by default look into `registry.access.redhat.com` registry and IMO it should build the app from there itself. If it does not find the specified image then only it will go for `docker hub` or any. 

> Yes, this will be part of a separate effort.

Yes, I really like the approach @girishramnani has suggested. This will reduce the manual efforts. And ofcourse it will require more effort and we should do it in different pr. But @amitkrout Can you please create an issue? So that we don't miss it.

amitkrout:
https://github.com/openshift/odo/issues/3307
/test v4.3-integration-e2e

amitkrout:
> @amitkrout Can you please create an issue? So that we don't miss it.

created https://github.com/openshift/odo/issues/3388

Issue #3316, opened 2020-06-08
CI tests failing due to Open Liberty devfile being renamed


johnmcollier:
/kind bug
/area testing

The Open Liberty was recently updated, and had its named changed from `openLiberty` to `java-openliberty`, which has caused a number of our integration tests to fail (as `odo create openLiberty` no longer works). 

The integration tests will need to be updated to check for `java-openliberty` instead

Issue #2641, opened 2020-02-26
Prow test are failing due to nodejs-8 image


kadel:
Prow test are failing due to nodejs-8 image!
 

ci/prow/v4.2-integration-e2e-benchmark 
ci/prow/v4.3-integration-e2e-benchmark 

```
• Failure [397.878 seconds]
odo supported images e2e tests
/go/src/github.com/openshift/odo/tests/e2escenarios/e2e_images_test.go:14
  odo supported images deployment
  /go/src/github.com/openshift/odo/tests/e2escenarios/e2e_images_test.go:137
    Should be able to verify the nodejs-8-centos7 image [It]
    /go/src/github.com/openshift/odo/tests/e2escenarios/e2e_images_test.go:173
    Timeout out after � minutes
    /go/src/github.com/openshift/odo/tests/helper/helper_generic.go:122
------------------------------
```

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/2637/pull-ci-openshift-odo-master-v4.2-integration-e2e-benchmark/684

/priority critical
/area testing
/kind failing-test

amitkrout:
@prietyc123 @kadel I have tested the same scenario on 4.2 cluster more than 5 times to have atleast one occurrence of failure but none of the the time it hit. 

From the logs it is clear that curl command could not hit the route in the specified time limit. Not sure if this is some kind of glitch from network side or from the nodejs application side itself. @kadel Until we are not sure on the issue behaviour till then can we add flake in the title.

kadel:
Yesterday I saw almost all PRs failing on this.
/kind flake
/priority high
/remove-priority critical
/remove-kind failing-test

kadel:
This is getting really really annoying :-/ 

Another occurrence in 

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_odo/2629/pull-ci-openshift-odo-master-v4.1-integration-e2e-benchmark/754

this time for centos7-s2i-nodejs:10.x image

prietyc123:
@amitkrout I am able to reproduce the error with ```nodejs-10:latest``` and ```centos7-s2i-nodejs:10.x``` images on 4.2 cluster. 

Curl command could not be able to hit for both the images. Need to figure out the reason behind this.

prietyc123:
[...]

PR #3318, opened 2020-06-08
Fix failing CI tests due to Open Liberty devfile rename

johnmcollier:
**What type of PR is this?**
/kind bug
/area testing

Updates odo's devfile integration tests to use `java-openliberty` instead of `openLiberty`, as the name of the Open Liberty devfile stack was recently renamed.

I'm putting this change up so that our PRs are unblocked on CI failures. I'm not crazy about an external dependency (a devfile registry in this case) being able to completely break our tests, but given that we need to test `odo create` with devfiles in a registry, I'm not sure what's the best approach here.

Fixes https://github.com/openshift/odo/issues/3316

Run the integration tests that I updated.


amitkrout:
https://github.com/openshift/odo/issues/2641
/retest

Issue #2893, opened 2020-04-15
Support secured devfile registries / index.jsons in odo


johnmcollier:
/kind user-story
/area devfile

## User Story
Currently, odo supports devfile registries that are running behind publicly signed certificates and don't require any authentication. As a user, I want to be able to use odo with devfile registries / index.jsons that may be behind self-signed certs and/or require authentication, via a username/password or personal access token.

## Acceptance Criteria 
- [ ] odo should be able to work with devfile registries / indexes that use self-signed certificates (such as ones running behind a corporate firewall), using a flag (such as `--insecure`)
- [ ] odo should allow the user to enter credentials to access a secured devfile registry if needed
     - Examples being an index.json on a private github repo, Github enterprise, etc

CC @elsony @deboer-tim

kadel:
> * odo should be able to work with devfile registries / indexes that use self-signed certificates (such as ones running behind a corporate firewall)

Self-signed certificates are insecure. By default, odo needs to fail if self-signed or otherwise invalid or untrusted certificate is used.
If there is a need to support self-signed certificates, we need to make it optional and odo needs to  always show a big warning that the certificate is self-signed or not checked.

> odo should allow the user to enter credentials to access a secured devfile registry if needed
> * Examples being an index.json on a private github repo, Github enterprise, etc

we should define how devfile registry will look like first. Is it going to be always a git repository? Simple HTTP(S) server serving static files? Or shoudl we define some API for it?

deboer-tim:
I agreed that self-signed certs are by definition insecure and should fail by default, but I do hit them regularly enough in dev systems. IMHO there should be an option to accept them, e.g. similar to "curl --insecure".

I also agree we should discuss where/how the registry can be hosted, and what the client requirements are. This should be a discussion along with other devfile registry clients so I've just opened an issue to discuss: https://github.com/che-incubator/devworkspace-api/issues/39.

We have an internal security team that is recommending support for several scenarios. There's a meeting later today, we can try to distill this and provide more specific examples of the environments that we think it should support.

adisky:
The current picture for a devfile registry is not much clear, We could assume that in any form, we need to provide authentication parameters to odo, 

As with the current code odo access registry when it downloads `devfile.yaml`, in `odo create` or while `odo catalog list components`. Two approach i can think of is
[...]

PR #3329, opened 2020-06-09
Design proposal: secure registry support

GeekArthur:

**What type of PR is this?**
/kind documentation
[skip ci]

Parent issue: https://github.com/openshift/odo/issues/2893

### DIFF
--- a/docs/proposals/secure-registry.md
+++ b/docs/proposals/secure-registry.md
@@ -66,6 +66,10 @@ Work flow to access secure registry:
 
     If multiple secure registries share the same credential, `odo registry delete <registry name>` will delete the credential from keyring instance once the last secure registry using that credential has been deleted.
 
+Create devfile component from secure registry:
+
+When downloading devfile from secure registry, we validate if the credential is valid by adding token to the request header and checking the response.
+
 ## Related issues
 - Dynamic registry support: https://github.com/openshift/odo/pull/2940
 - Performance improvement for `odo catalog list components`: https://github.com/openshift/odo/pull/3112
--- a/docs/proposals/secure-registry.md
+++ b/docs/proposals/secure-registry.md
@@ -40,7 +40,9 @@ Work flow to access secure registry:
 
     Currently we have `odo registry add <registry name> <registry URL>` and `odo registry update <registry name> <registry URL>` to add and update registry accordingly. Regarding the CLI design, we can implement the following CLI design to support collecting credential:
     - `odo registry add <registry name> <registry URL> --token <token>`
-    -  `odo registry update <registry name> <registry URL> --token <token>`
+    - `odo registry add <registry name> <registry URL> --user <user> --password <password>`
+    - `odo registry update <registry name> <registry URL> --token <token>`
+    - `odo registry update <registry name> <registry URL> --user <user> --password <password>`
 
 2. Store Credential
 
--- a/docs/proposals/secure-registry.md
+++ b/docs/proposals/secure-registry.md
@@ -1,0 +1,71 @@
+# Secure Registry Support
+
+Table of contents
+- [Problem Statement](#problem-statement)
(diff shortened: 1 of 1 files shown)


GeekArthur:
> Looks good. It looks like you have addressed my early comments on the design from our previous conversation already. One thing that I think is missing is slightly beyond the catalog list, when someone is doing `odo create` on a component that belongs to a secure registry, we'll need to make sure we handle the component creation properly (including download of the devfile from the secure repo) using same secure setting.

Thanks for reviewing! I add the `Create devfile component from secure registry` section to the design proposal.

docs/proposals/secure-registry.md
```diff
@@ -0,0 +1,75 @@
...
+
+Context:
+1. Given GitHub is going to depreciate basic authentication with username/password https://developer.github.com/changes/2020-02-14-deprecating-password-auth/, we have to only support personal access token authentication method for GitHub-hosted registry scenario.
+2. For cluster-hosted registry, the registry architecture would be creating a NGINX server to host the secure registry, then create a ingress gateway for the NGNIX server to let client access.
+
+Work flow to access secure registry:
+1. Collect Credential
+
+    Currently we have `odo registry add <registry name> <registry URL>` and `odo registry update <registry name> <registry URL>` to add and update registry accordingly. Regarding the CLI design, we can implement the following CLI design to support collecting credential:
+    - `odo registry add <registry name> <registry URL> --token <token>`
```
kadel, line 42: What would be a command for adding registry with basic-auth?

I can see two options:

`odo registry add <registry name> <registry URL> --user <user> --password <password>`
or 
`odo registry add <registry name> <registry URL> --basic-auth <user>:<password>`

GeekArthur: I like the first option as it aligns with `kubectl`, `docker`, `oc` and `odo`, I am gonna update the design proposal.

GeekArthur:
@elsony @kadel Comments addressed, please review again, thanks

gorkem:
Also consider https://github.com/devfile/kubernetes-api/issues/82

GeekArthur:
> Also consider [devfile/kubernetes-api#82](https://github.com/devfile/kubernetes-api/issues/82). I would hold a decision until it is concluded on the devfile side. If we decide on the OCI based registry it is possible that some (all?) of the concerns on this document will be addressed.

@gorkem Thanks for sharing the OCI-based registry architecture design, this design proposal only focuses on secure registry support, more specifically it only addresses concerns of authentication and authorization stages for existing supported registry and it's not related with the registry design. We can open separate issue and create another design proposal for the registry design to adopt the architecture of OCI-based registry.

docs/proposals/secure-registry.md
```diff
@@ -0,0 +1,77 @@
...
+    req.Header.Add("Authorization", bearer)
+    ```
+
+4. Delete Credential
+
+    If multiple secure registries share the same credential, `odo registry delete <registry name>` will delete the credential from keyring instance once the last secure registry using that credential has been deleted.
+
+Create devfile component from secure registry:
+
+When downloading devfile from secure registry, we validate if the credential is valid by adding token to the request header and checking the response.
```
adisky, line 73: we are not doing any validation on `odo registry add <registry name> <registry URL> --user <user> --password <password>`? Just thinking out here, shouldn't we validate credentials and url here? than doing at the time of `odo create`?

GeekArthur: Regarding arguments/flags validation, the design is that we divide the validation into two stages:
Early validation stage: Only validate URL basic format, happens when user adds registry URL
Validation stage: Validate URL accessibility, happens when user uses registry URL

That design is based on `yum-config-manager`, design proposal with details under this thread: https://github.com/openshift/odo/pull/2940#discussion_r417440866

PR #3331, opened 2020-06-09
refactoring: remove redundant PlatformAdapter interface

### LINKED ISSUE #3333: PlatformAdapter interface seems redundant and should be removed
/kind bug

**Operating System:** N/A

**Output of `odo version`:** N/A

N/A

## Actual behavior
N/A

## Expected behavior
N/A

N/A

metacosm:
**What type of PR is this?**
/kind cleanup

Remove seemingly redundant PlatformAdapter interface

Fixes #3333

Behavior shouldn't be impacted.


girishramnani:
@metacosm please open an issue and discuss in the public slack channel before opening PRs if possible 🙂

metacosm:
I wanted to check the tests actually before proposing this as an actual change, hence the draft status… 😉

kadel:
> I wanted to check the tests actually before proposing this as an actual change, hence the draft status… 😉

Test are not automatically executed on Draft PRs ;-) You need to manually trigger them

/test all

girishramnani:
I would prefer to use the interface as that enforces a layer of abstraction and someday allows someone to mock it for unit tests

metacosm:
> I would prefer to use the interface as that enforces a layer of abstraction and someday allows someone to mock it for unit tests

Can you please explain? There's already such an interface: `ComponentAdapter`.

pkg/devfile/adapters/helper.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/openshift/odo/pkg/odo/util/pushtarget"
 )
 
-// NewPlatformAdapter returns a Devfile adapter for the targeted platform
-func NewPlatformAdapter(componentName string, context string, devObj devfileParser.DevfileObj, platformContext interface{}) (PlatformAdapter, error) {
+// NewComponentAdapter returns a Devfile adapter for the targeted platform
```
adisky, line 15: can we also rename the `ComponentAdapter` interface to `PlatformAdapter`?
https://github.com/openshift/odo/blob/7da08b69bab0cc6e172c4e1a39ac570566f26d1d/pkg/devfile/adapters/common/interface.go#L4

IMO  as this interface is intended for different platform(docker/k8s), `PlatformAdapter` sounds better, and component term is already too overloaded in odo.

metacosm: Agreed, I just kept the most used interface and its name but it's probably indeed better to rename it and maybe move it somewhere where it makes more sense.

metacosm: Actually, after taking another look at the code, `ComponentAdapter` is appropriate because its API deals with component and adapts the underlying implementation to the component functionality so I think it's OK to keep `ComponentAdapter`.

girishramnani: but there is a component adapter present here https://github.com/openshift/odo/blob/5f71123d4332e47116b8c8112e76f91ec73f8833/pkg/devfile/adapters/kubernetes/component/adapter.go#L29
its not called componentAdapter but its inside component/adapter.go wouldn't that be confusing?

metacosm: There's also one here: https://github.com/openshift/odo/blob/5f71123d4332e47116b8c8112e76f91ec73f8833/pkg/devfile/adapters/docker/component/adapter.go#L21… What's your point? I didn't do an indepth refactoring here, just a quick win.

girishramnani:
/approve
/lgtm
looks good and is a small change - just refactors an interface

Issue #3297, opened 2020-06-03
`odo catalog list components` fails if registry is not set up properly 


GeekArthur:
/kind bug

## What versions of software are you using?

**Operating System:**
MacOS

**Output of `odo version`:**
Latest master

## How did you run odo exactly?
```
odo registry add Fake https://fake                                                                                      ✔  14:29:21 
New registry successfully added
odo catalog list components                                                                                             ✔  14:29:35 
 ⚠  Registry  is not set up properly with error: <nil>
 ✗  Get "https://fake/devfiles/index.json": dial tcp: lookup fake: no such host
```

## Actual behavior
`odo catalog list components` fails if registry is not set up properly.

## Expected behavior
We should use warning instead of error, because we need to support the case where user has multiple registries but only part of the registries is not set up properly, for that case we shouldn't block user from listing registry that is set up properly and listing s2i components. 

## Any logs, error output, etc?
N/A

## Special Notes:
This regression starts from this PR https://github.com/openshift/odo/pull/3112 gets merged, we need to do the following things to fix the issue:
1. Use warning instead of error for registry that is not set up properly
2. Display the registry name properly when throwing warning
3. Add tests to cover the case to avoid regression in the future

johnmcollier:
I think there's a couple things that need to be done as part of this:
1) `odo catalog list` shouldn't error out if one of the registries is invalid or cannot be reached. A warning should be displayed instead
2) `odo registry add` should not allow users to add invalid registries.

GeekArthur:
> 1. odo catalog list shouldn't error out if one of the registries is invalid or cannot be reached. A warning should be displayed instead

Do you mean `odo registry list` instead of `odo catalog list`? Because `odo catalog list` issue is already under this issue description. If you mean `odo registry list`, I think that error is unlikely as `odo registry add` already validates for the error so that the error won't be moved to `odo registry list`

> odo registry add should not allow users to add invalid registries.

We already have that validation in existing code base

johnmcollier:
> Do you mean odo registry list instead of odo catalog list? Because odo catalog list issue is already under this issue description. If you mean odo registry list, I think that error is unlikely as odo registry add already validates for the error so that the error won't be moved to odo registry list

Yes, I mean `odo catalog list`

> We already have that validation in existing code base

We don't validate that the URL we're adding is valid or that the URL + `devfiles/index.json` is valid. I think an error should occur for the following (which it does not do today)
```
odo registry add Fake https://fake
```

[...]

Issue #3167, opened 2020-05-13
`odo create <component-name> --registry` falls back to S2I component, if invalid registry used


adisky:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:master

## How did you run odo exactly?
```
$ odo create maven --registry blbla
Experimental mode is enabled, use at your own risk

 ⚠  Registry blbla is not set up properly with error: Unable to unmarshal the devfile index.json from https://raw.githubusercontent.com/blbla/devfiles/index.json: invalid character ':' after top-level value
Validation
 ✗  Checking devfile compatibility [102517ns]
 ✗  Creating a devfile component from registry: blbla [141367ns]

Please run `odo catalog list components` for a list of supported devfile component types
 ✗  imagestreams.image.openshift.io "maven" not found
```

## Actual behavior
mentioned above

## Expected behavior: odo should exits, as soon as an invalid registry is specified via flag, It is should not go ahead for further validation
```
$ odo create maven --registry blbla
Experimental mode is enabled, use at your own risk

 x Registry blbla is not set up properly with error: Unable to unmarshal the devfile index.json from https://raw.githubusercontent.com/blbla/devfiles/index.json: invalid character ':' after top-level value
Validation

```

GeekArthur:
+1, we should handle the invalid registry error properly, basically better error handling would be:
1. If registry doesn't exist (eg. pass in fake registry) -> directly throw error to user and won't continue to run both devfile and s2i code (given currently s2i doesn't have registry support)
2. If registry exists but component type doesn't exist under the specified registry -> throw properly error to let user know the component type doesn't exist instead of using `Checking devfile compatibility` as that error may make users confused (users may think root cause is devfile doesn't exist under the specified registry OR root cause is devfile format/content is wrong)

Regarding another topic which is fully support s2i registry, that will be done under this issue: https://github.com/openshift/odo/issues/3156

GeekArthur:
> If registry doesn't exist (eg. pass in fake registry) -> directly throw error to user and won't continue to run both devfile and s2i code (given currently s2i doesn't have registry support)

Add validation code to validate `--registry` value, the validation code should create a function/code block to get available registries from preference file then compare with the `--registry` value to validate registry existence and correctness

[...]

PR #3341, opened 2020-06-12
Fix multiple registry related issues

GeekArthur:

**What type of PR is this?**
/kind bug
/area devfile

This PR fixes the following issues related with registry
1. Only throw warning for invalid registry and display registry name properly
2. Improve `--registry` flag validation
3. Improve component type existence validation when creating devfile component from registry
4. Add GitHub raw URL conversion support so that user can directly add GirHub regular URL (can be reachable) instead of GitHub raw URL as registry URL
5. Update corresponding help and usage

Fixes https://github.com/openshift/odo/issues/3297
Fixes https://github.com/openshift/odo/issues/3167
Fixes https://github.com/openshift/odo/issues/3382

1. Only throw warning for invalid registry and display registry name properly
   1. Run `odo registry add fake http://fake`
   2. Run `odo catalog list components` can list valid registry and throw warning for invalid registry
2. Improve `--registry` flag validation
   1. Run `odo create nodejs --registry fake` can throw proper error directly
3. Improve component type existence validation when creating devfile component from registry
   1. Run `odo create nodejs` can validate if the component type exists in the registry

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -73,7 +73,7 @@ func convertURL(URL string) (string, error) {
 	if strings.Contains(url.Host, "github") && !strings.Contains(url.Host, "raw") {
 		// Convert path part of the URL
 		URLSlice := strings.Split(URL, "/")
-		if URLSlice[len(URLSlice)-2] == "tree" {
+		if len(URLSlice) > 2 && URLSlice[len(URLSlice)-2] == "tree" {
 			// GitHub raw URL doesn't have "tree" structure in the URL, need to remove it
 			URL = strings.Replace(URL, "/tree", "", 1)
 		} else {
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -25,12 +25,6 @@ const (
 	apiVersion = "odo.dev/v1alpha1"
 )
 
-// DevfileRegistries contains the links of all devfile registries
-var DevfileRegistries = []string{
-	"https://raw.githubusercontent.com/elsony/devfile-registry/master",
-	"https://che-devfile-registry.openshift.io/",
-}
-
 // GetDevfileRegistries gets devfile registries from preference file,
 // if registry name is specified return the specific registry, otherwise return all registries
 func GetDevfileRegistries(registryName string) (map[string]Registry, error) {
@@ -76,7 +70,7 @@ func convertURL(URL string) (string, error) {
 		return "", err
 	}
 
-	if strings.Contains(url.Host, "github") {
+	if strings.Contains(url.Host, "github") && !strings.Contains(url.Host, "raw") {
 		// Convert path part of the URL
 		URLSlice := strings.Split(URL, "/")
 		if URLSlice[len(URLSlice)-2] == "tree" {
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -3,10 +3,11 @@ package catalog
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/openshift/odo/pkg/preference"
 	"sort"
 	"strings"
 	"sync"
+
+	"github.com/openshift/odo/pkg/preference"
 
 	imagev1 "github.com/openshift/api/image/v1"
 	"github.com/openshift/odo/pkg/devfile/adapters/common"
@@ -171,10 +172,9 @@ func ListDevfileComponents(registryName string) (DevfileComponentTypeList, error
 		// Load the devfile registry index.json
 		registry := reg // needed to prevent the lambda from capturing the value
 		retrieveRegistryIndices.Add(util.ConcurrentTask{ToRun: func(errChannel chan error) {
-			indexEntries, e := getDevfileIndexEntries(registry)
-			if e != nil {
-				log.Warningf("Registry %s is not set up properly with error: %v", registryName, err)
-				errChannel <- e
+			indexEntries, err := getDevfileIndexEntries(registry)
+			if err != nil {
+				log.Warningf("Registry %s is not set up properly with error: %v", registry.Name, err)
 				return
 			}
 
@@ -203,7 +203,6 @@ func ListDevfileComponents(registryName string) (DevfileComponentTypeList, error
 			devfile, err := GetDevfile(link)
 			if err != nil {
 				log.Warningf("Registry %s is not set up properly with error: %v", devfileIndex.Registry.Name, err)
-				errChannel <- err
 				return
 			}
 
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -3,6 +3,7 @@ package catalog
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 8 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -359,9 +359,16 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 
 		// Validate user specify registry
 		if co.devfileMetadata.devfileRegistry.Name != "" {
-			// TODO: We should add more validations here to validate registry existence and correctness
 			if co.devfileMetadata.devfilePath.value != "" {
-				return errors.New("You can specify registry via --registry if you want to use the devfile that is specified via --devfile")
+				return errors.New("You can't download devfile from registry by specifying --registry if you want to use the devfile that is specified via --devfile")
```
johnmcollier, line 363: Nit: I think the old error message was more clear, the new one is a little too long

pkg/odo/cli/component/create.go
```diff
@@ -359,9 +359,16 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+			}
+
+			registryList, err := catalog.GetDevfileRegistries(co.devfileMetadata.devfileRegistry.Name)
+			if err != nil {
+				return err
```
johnmcollier, line 368: Wrap this error with a message

pkg/catalog/catalog.go
```diff
@@ -172,10 +172,9 @@ func ListDevfileComponents(registryName string) (DevfileComponentTypeList, error
 		// Load the devfile registry index.json
 		registry := reg // needed to prevent the lambda from capturing the value
 		retrieveRegistryIndices.Add(util.ConcurrentTask{ToRun: func(errChannel chan error) {
-			indexEntries, e := getDevfileIndexEntries(registry)
-			if e != nil {
-				log.Warningf("Registry %s is not set up properly with error: %v", registryName, err)
-				errChannel <- e
```
johnmcollier, line 178: 👍

pkg/odo/cli/component/create.go
```diff
@@ -359,9 +359,16 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+			}
+			if len(registryList) == 0 {
+				return errors.Errorf("Registry %s doesn't exist, please specify the valid registry via --registry", co.devfileMetadata.devfileRegistry.Name)
```
johnmcollier, line 371: Nit: Should be `please specify a valid registry`

tests/integration/devfile/cmd_devfile_catalog_test.go
```diff
@@ -92,4 +92,18 @@ var _ = Describe("odo devfile catalog command tests", func() {
 			helper.MatchAllInOutput(output, wantOutput)
 		})
 	})
+
+	Context("When executing catalog list components with registry that is not set up properly", func() {
```
johnmcollier, line 96: Can you also check for the warning message that's logged?

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -121,10 +121,15 @@ var _ = Describe("odo devfile create command tests", func() {
 	})
 
 	Context("When executing odo create with devfile component type argument and --registry flag", func() {
-		It("should successfully create the devfile component", func() {
+		It("should successfully create the devfile component if specified registry is valid", func() {
 			componentRegistry := "DefaultDevfileRegistry"
 			helper.CmdShouldPass("odo", "create", "java-openliberty", "--registry", componentRegistry)
 		})
+
+		It("should fail to create the devfile component if specified registry is invalid", func() {
```
johnmcollier, line 129: Can you also validate the log message that you added?

pkg/odo/cli/component/create.go
GeekArthur: No problem, I can revert it back, I think the biggest problem for the old error message is it's missing "not". 

`You can specify registry` -> `You can't specify registry` because `--registry` and `--devfile` are mutual exclusive

tests/integration/devfile/cmd_devfile_catalog_test.go
GeekArthur: Yeah, that's a good question, I tried that before but failed. From CLI perspective we can see the warning message and I think both warning and regular messages are in `stdout`, but from testing perspective inside `CmdShouldPass` function, I find it wraps `ginkgo` and `gomega` functions for returning command output and warning message won't be returned (only regular message is returned), which means we can't catch the warning messages from test cases, I suspect in test cases warning message redirects to other stream, that's why I use this way for testing (to test invalid registry won't block listing valid registry).

tests/integration/devfile/cmd_devfile_create_test.go
GeekArthur: Yep, I can do that.

pkg/odo/cli/component/create.go
GeekArthur: Sure, will add that.

adisky:
@GeekArthur one more issue i have seen, not related to this PR. Since it is fixing multiple issue i thought to comment here, I could log a separate issue if you want.
When we run, the url listed ` https://raw.githubusercontent.com/elsony/devfile-registry/master`, returns `400: Invalid Request`.
```
$ odo registry list
NAME                       URL
CheDevfileRegistry         https://che-devfile-registry.openshift.io
```
 IMO we should show user a url that is reachable, I know we add the url in the same format.
`  odo registry add CheRegistryFromGitHub https://raw.githubusercontent.com/eclipse/che-devfile-registry/master
`
but shouldn't we directly take github repo url and convert it into raw url?

GeekArthur:
> @GeekArthur one more issue i have seen, not related to this PR. Since it is fixing multiple registry issue i thought to comment here, I could log a separate issue if you want.
> When we run, the url listed ` https://raw.githubusercontent.com/elsony/devfile-registry/master`, returns `400: Invalid Request`.
> 
> ```
> $ odo registry list
> NAME                       URL
> CheDevfileRegistry         https://che-devfile-registry.openshift.io
> DefaultDevfileRegistry     https://raw.githubusercontent.com/elsony/devfile-registry/master
> ```
> 
> IMO we should show user a url that is reachable, I know we add the url in the same format.
> ` odo registry add CheRegistryFromGitHub https://raw.githubusercontent.com/eclipse/che-devfile-registry/master`
> but shouldn't we directly take github repo url and convert it into raw url?

I think that's doable and we should consider two cases below as the host part of the raw URL is different:
- GitHub public: convert from `github.com` to `raw.githubusercontent.com`
- GitHub enterprise: convert from `github.<enterprise>.com` to `raw.github.<enterprise>.com`

@adisky Can you open a separate to track this?

tests/integration/devfile/cmd_devfile_catalog_test.go
```diff
@@ -92,4 +92,18 @@ var _ = Describe("odo devfile catalog command tests", func() {
...
+		It("should list components from valid registry", func() {
+			helper.CmdShouldPass("odo", "registry", "add", "fake", "http://fake")
+			output := helper.CmdShouldPass("odo", "catalog", "list", "components")
+			wantOutput := []string{
```
prietyc123, line 100: How about doing this in a single line like https://github.com/openshift/odo/blob/master/tests/integration/devfile/cmd_devfile_push_test.go#L197 ?

tests/integration/devfile/docker/cmd_docker_devfile_catalog_test.go
```diff
@@ -64,4 +64,18 @@ var _ = Describe("odo docker devfile catalog command tests", func() {
 			helper.MatchAllInOutput(output, wantOutput)
 		})
 	})
+
+	Context("When executing catalog list components with registry that is not set up properly", func() {
+		It("should list components from valid registry", func() {
+			helper.CmdShouldPass("odo", "registry", "add", "fake", "http://fake")
+			output := helper.CmdShouldPass("odo", "catalog", "list", "components")
+			wantOutput := []string{
```
prietyc123, line 72: https://github.com/openshift/odo/pull/3341#discussion_r441962518

adisky:
@GeekArthur I had opened the issue https://github.com/openshift/odo/issues/3382

tests/integration/devfile/docker/cmd_docker_devfile_catalog_test.go
GeekArthur: Done

tests/integration/devfile/cmd_devfile_catalog_test.go
GeekArthur: Done

pkg/catalog/catalog.go
```diff
@@ -65,11 +60,49 @@ func GetDevfileRegistries(registryName string) (map[string]Registry, error) {
...
+func convertURL(URL string) (string, error) {
+	url, err := url.Parse(URL)
+	if err != nil {
+		return "", err
+	}
+
+	if strings.Contains(url.Host, "github") && !strings.Contains(url.Host, "raw") {
+		// Convert path part of the URL
+		URLSlice := strings.Split(URL, "/")
+		if URLSlice[len(URLSlice)-2] == "tree" {
```
girishramnani, line 76: Please check if the length is more then 2 so that the subtraction doesn’t yield a negative number

GeekArthur: I can add that validation but that case is unlikely because the format of regular GitHub URL is `<protocol>://<host>/<user>/<repository>`, it may or may not have branch part to append the URL, so definitely the length of the slice is more than 2.

PR #3365, opened 2020-06-16
fix: properly check that odo version contains the server URL

### LINKED ISSUE #3342: No odo version verification is done, only information parttern is matched
/kind bug

**Output of `odo version`:**
master

## Actual behavior
```odo version``` test spec https://github.com/openshift/odo/blob/master/tests/integration/generic_test.go#L266 only checks the server pattern URL but does not validate the actual server URL and other information too. 

## Expected behavior
Server URL and other information in ```odo version``` string output should be verified with ```oc version``` output.

Born from discussion - https://github.com/openshift/odo/issues/3325#issuecomment-643594506

metacosm:
Fixes #3342

**What type of PR is this?**
/kind bug

See related issue.

Fixes #3342

### DIFF
--- a/tests/helper/helper_oc.go
+++ b/tests/helper/helper_oc.go
@@ -50,6 +50,20 @@ func (oc OcRunner) GetCurrentProject() string {
 	return ""
 }
 
+// GetCurrentServerURL retrieves the URL of the server we're currently connected to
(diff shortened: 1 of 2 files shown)


amitkrout:
prow issue
/retest

Issue #3300, opened 2020-06-04
Support OpenShift build and deploy support in odo


neeraj-laad:
/kind feature

## Which functionality do you think we should add?
Add a new command (verb) to build a production-like/slim container image for their project and deploy the built image on the target platform.

- `odo deploy` could provide developers with a way to build a container image for their application and deploy it on a target Kubernetes deployment using the build/deploy guidance provided by the stack.
- This is not meant to replace GitOps / pipeline based deployments for your controlled environments (test, stage, production), but it will complement them and can be a good way to assure the developer that their application can be built and deployed successfully. Both the deployment routes must honour the build and deployment information provided by the application stack (devfile).

The design proposal for this feature is here: https://github.com/openshift/odo/pull/3368

Early design thoughts were captured here:  https://docs.google.com/document/d/1uaFOmutO_bWMv78tPTrdCUuFEBMfz9GADyVInySDsOI
Note: This google doc is now obsolete as we have a forma design proposal.

## Why is this needed?
odo provides a great way to develop containerized applications on Docker and Kubernetes platforms (including OpenShift) using applications stacks based on Devfiles. However, this is limited to development inner-loop for a project and there is no support for outer-loop - build a slim/production container image for your application and deploy the build container. 

It would be very useful for application developers to be able to build a container image and perform a Kubernetes deployment for their project. This information could be provided by the devfile to avoid developers having to worry about these aspects.

neeraj-laad:
@kadel @elsony @girishramnani Would be very useful to get your view on this proposal/design.

sbose78:
I did plan to have both 'odo deploy' and 'odo pipelines' co-exist. So this is well-aligned!

"odo build" is for deploying the code as a 'real' workload.
"odo pipelines" is for multi-stage & multi-task activities which would eventually be "saved" in a Git repo and managed out of it.


An ideal solution should look like this:

"odo deploy" should only build an image, push it to a registry and deploy it, with/without use of Tekton objects. 

* On OpenShift, If the cluster has BuildConfig, use BuildConfig ( supports Dockerfile, s2i with detected language ). Doesn't need any devfile guidance.

* On Kubernetes/OpenShift, Use [Build v2](https://github.com/redhat-developer/build/) / Tekton APIs if present ( supports Dockerfile, Buildpacks, s2i  )

* Else, use plain pod/container steps in a Deployment as recommended in the doc.

Regarding guidance,
* If there's devfile guidance, use it!
[...]

PR #3368, opened 2020-06-17
Proposal for odo deploy

neeraj-laad:
**What type of PR is this?**
/kind design
[skip ci]

This is a design proposal for adding limited outer-loop capabilities to odo. Additional information non this is captured here:  #3300 

Relates to: #3300 

This is a design proposal, once the design is agreed upon we can submit a PR for the implementation. We might start making some progress on a prototype based on this proposal and then adjust as the design evolves.

### DIFF
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -126,9 +126,9 @@ spec:
 
 ## Future evolution
 
-Devfile 2.1.0 should broaden the scope for the outer-loop support in devfiles. For example: 
-- Support multiple build strategies - buildah, s2i, build v3 etc.
-- Support multiple deployment strategies - k8s manifest, native service, pod spec etc.
-- Any referenced assets should be immutable to ensure reproducible builds/deployments.
+- Devfile 2.1.0 should broaden the scope for the outer-loop support in devfiles. For example:
+    - Support multiple build strategies - buildah, s2i, build v3 etc.
+    - Support multiple deployment strategies - k8s manifest, native service, pod spec etc.
+    - Any referenced assets should be immutable to ensure reproducible builds/deployments.
 
-When adding support for devfile 2.1.0, odo will implement several build/deployment strategies, and if the strategy is not supported by odo then display a meaningful error message to the user.
+- If a devfile does not provide deployment manifest, odo can perhaps create a manifest in the way it does for inner-loop. This will mean devfile creators do not need to provide a deployment manifest if they do not care so much about the deployment aspect.
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -58,7 +58,7 @@ This command will build a container image for the application and deploy it on t
 Flags:
  - `--tag`: The tag to be used for the built application container image - `<registry>/<org>/<name>:<tag>` (mandatory).
  - `--service-account`: The service account for running privileged containers and push access to the image registry.
- - `--credentials`: The credentials needed to push the image to the container image registry (mandatory).
+ - `--credentials`: The credentials needed to push the image to the container image registry (optional).
 
 #### odo deploy delete
 This command will delete any resources created by odo deploy.
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -17,14 +17,14 @@ It would be useful to start the design/development of a simpler version of `odo 
 
 It is not meant to replace GitOps/pipeline-based deployments for governed environments (test, stage, production). However, both `odo deploy` and `odo pipelines` must honour the build and deployment information provided by the application stack (devfile).
 
-## User Flow
+## User flow
 
 This command will allow a user to perform inner-loop and then test the outer-loop, so the application is truly ready for checking into git and for pipelines to take over. Here's how a typical application development flow might look like: 
 
 User flow: 
 1. `odo create <component> <mycomponent>` - This initializes odo component in the current directory.
 1. Edit the project source code to develop the application.
-1. `odo URL create` - This stores URL information (host, port etc.) for accessing the application on the cluster (if not done already).
+1. `odo url create` - This stores URL information (host, port etc.) for accessing the application on the cluster (if not done already).
 1. `odo push` - This runs the application source code using inner-loop instructions from devfile.
 1. Validate the running application is working as intended.
 1. Iterate over steps 2 and beyond (as needed). 
@@ -34,7 +34,7 @@ User flow:
 1. Optionally, run `odo deploy delete` to clean up resources created with `odo deploy`.
 1. Push your code to Git - ready for sharing with your team and for CI/CD pipelines to take over.
 
-## User Stories
+## User stories
 
 ### Initial build and deploy support to odo - https://github.com/openshift/odo/issues/3300
 
@@ -47,18 +47,17 @@ This deployment is equivalent to a development version of your production and wi
 
 ### High-level design:
 
-#### Technical considerations:
+#### Pre-requisites:
 - The implementation would be under the experimental flag.
 - Only supported for Devfile v2.0.0 components.
 - Only supported for Kubernetes/ OpenShift targets.
-- `odo deploy` will not make any changes to the user's project.
 
 #### odo deploy 
 This command will build a container image for the application and deploy it on the target Kubernetes environment. 
 
 Flags:
  - `--tag`: The tag to be used for the built application container image - `<registry>/<org>/<name>:<tag>` (mandatory).
- - `service-account`: The service account for running privileged containers and push access to the image registry.
+ - `--service-account`: The service account for running privileged containers and push access to the image registry.
  - `--credentials`: The credentials needed to push the image to the container image registry (mandatory).
 
 #### odo deploy delete
@@ -77,12 +76,10 @@ attributes:
     deployment-manifest: <URI>
 ```
 
-This should not need any change to the devfile v2 schema or the parser.
-
 ### Dockerfile
 This could be any valid dockerfile.
 
-### Deployment Manifest
+### Deployment manifest
 The deployment manifest could be templated to help with replacing key bits of information:
 - PROJECT_NAME
 - CONTAINER_IMAGE
@@ -124,13 +121,12 @@ This command will perform the following actions:
 - Delete the build container and the pod.
 
 #### Deploy
-- Delete any existing deployments (if the user runs `odo deploy` multiple times).
 - Fetch the deployment manifest using URI in the attributes of the devfile. and make it available in the container.
 - Replace templated text in the deployment manifest with relevant values:
     - PROJECT_NAME: name of odo project
     - CONTAINER_IMAGE: `tag` for the built image
     - PORT: URL information in env.yaml
-- Apply the new deployment manifest.
+- Apply the new deployment manifest create/update the application deployment.
 - Save the deployment manifest in `.odo` folder.
 - Provide the user with a URL for accessing the deployed application. 
 
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -68,12 +68,12 @@ This command will delete any resources created by odo deploy.
 
 ### Devfile
 
-For the initial implementation, we could use devfile v2.0.0 and capture basic outer-loop information as `metadata` on the devfile. `odo` could look for  specific keys, while other tools like Che could ignore them.
+For the initial implementation, we could use devfile v2.0.0 and capture basic outer-loop information as `attributes` on the devfile. `odo` could look for  specific keys, while other tools like Che could ignore them.
 
 For example: 
 ```
-metadata:
- dockerfile: <URI>
+attributes:
+ build-dockerfile: <URI>
  deployment-manifest: <URI>
 ```
 
@@ -109,7 +109,7 @@ spec:
 This command will perform the following actions:
 
 #### Input validation
-- Check if the devfile  is v2.0.0 and that it specifies the expected outer-loop metadata. 
+- Check if the devfile  is v2.0.0 and that it specifies the expected outer-loop attributes. 
     - If not provided, display a meaningful error message.
 - Validate all arguments passed by the user. 
     - If argument values are invalid, display a meaningful error message.
@@ -118,14 +118,14 @@ This command will perform the following actions:
 - Start a new pod with a build container (privileged container with `buildah:stable` image).
 - Mount the image registry credentials passed by the user.
 - Copy the application source code into the build container.
-- Fetch the dockerfile using URI in the metadata of the devfile and make it available in the container.
+- Fetch the dockerfile using URI in the attributes of the devfile and make it available in the container.
 - Execute `buildah bud` command using src code, dockerfile and tag information.
 - Push the built image using the tag and credentials.
 - Delete the build container and the pod.
 
 #### Deploy
 - Delete any existing deployments (if the user runs `odo deploy` multiple times).
-- Fetch the deployment manifest using URI in the metadata of the devfile and make it available in the container.
+- Fetch the deployment manifest using URI in the attributes of the devfile. and make it available in the container.
 - Replace templated text in the deployment manifest with relevant values:
     - PROJECT_NAME: name of odo project
     - CONTAINER_IMAGE: `tag` for the built image
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -73,8 +73,8 @@ For the initial implementation, we could use devfile v2.0.0 and capture basic ou
 For example: 
 ```
 attributes:
- build-dockerfile: <URI>
- deployment-manifest: <URI>
+    build-dockerfile: <URI>
+    deployment-manifest: <URI>
 ```
 
 This should not need any change to the devfile v2 schema or the parser.
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -41,7 +41,7 @@ User flow:
 ## Design overview
 `odo deploy` could provide developers with a way to build a container image for their application and deploy it on a target Kubernetes deployment using the build/deploy guidance provided by the devfile.
 
-This deployment is equivalent to a development version of your production and will be using the namespace, service binding, URL information from the inner-loop. This will ensure that it is not seen as a way to deploy real workloads in production.
+This deployment is equivalent to a development version of your production and will be using the namespace and URL information from the inner-loop. This will ensure that it is not seen as a way to deploy real workloads in production.
 
 `odo deploy delete` could provide developers with a way to clean up any existing deployment of the application. 
 
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -86,6 +86,7 @@ The deployment manifest could be templated to help with replacing key bits of in
 - PORT
 
 For example: 
+A deployment based on Runtime component operator (https://operatorhub.io/operator/) might look like this: 
 ```
 apiVersion: app.stacks/v1beta1
 kind: RuntimeComponent
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -85,7 +85,7 @@ This could be any valid dockerfile.
 
 ### Deployment manifest
 The deployment manifest could be templated to help with replacing key bits of information:
-- PROJECT_NAME
+- COMPONENT_NAME
 - CONTAINER_IMAGE
 - PORT
 
@@ -110,7 +110,7 @@ This command will perform the following actions:
     - Use BuildConfig to build and push a container image by using:
         - the source code in the user's workspace
         - dockerfile specified by the devfile
-        - tag generated based on project and internal registry details
+        - tag generated based on component and internal registry details
 - If the cluster does not support `BuildConfig`, switch to Kaniko:
     - Use Kaniko to build and push a container image by using:
         - the source code in the user's workspace
@@ -123,7 +123,7 @@ This command will perform the following actions:
 - Delete existing deployment, (if invoked with --force flag)
 - Fetch the deployment manifest using URI in the metadata of the devfile.
 - Replace templated text in the deployment manifest with relevant values:
-    - PROJECT_NAME: name of odo project
+    - COMPONENT_NAME: name of odo component/microservice
     - CONTAINER_IMAGE: `tag` for the built image
     - PORT: URL information in env.yaml
 - Apply the new deployment manifest create/update the application deployment.
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -76,8 +76,8 @@ For the initial implementation, we could use devfile v2.0.0 and capture basic ou
 For example: 
 ```
 metadata:
-    build-dockerfile: <URI>
-    deployment-manifest: <URI>
+    alpha.build-dockerfile: <URI>
+    alpha.deployment-manifest: <URI>
 ```
 
 ### Dockerfile
--- a/docs/proposals/odo-deploy.md
+++ b/docs/proposals/odo-deploy.md
@@ -7,7 +7,7 @@ There is an existing proposal for adding on outer-loop information (including su
 https://github.com/devfile/kubernetes-api/issues/49 
 
 It would be useful to start the design/development of a simpler version of `odo deploy` with devfile 2.0.0 that covers:
-- single build strategy - Dockerfile built using `buildah`
+- single build strategy - Dockerfile built using `BuildConfig` or `Kaniko`.
 - single deployment strategy - Kubernetes manifest deployed using `kubectl`.
 
 ## Motivation
@@ -19,7 +19,7 @@ It is not meant to replace GitOps/pipeline-based deployments for governed enviro
 
 ## User flow
 
-This command will allow a user to perform inner-loop and then test the outer-loop, so the application is truly ready for checking into git and for pipelines to take over. Here's how a typical application development flow might look like: 
+This command will allow a user to test the transition from inner-loop t outer-loop, so the application is truly ready for pipelines to take over. Here's how a typical application development flow might look like: 
 
 User flow: 
 1. `odo create <component> <mycomponent>` - This initializes odo component in the current directory.
@@ -50,15 +50,19 @@ This deployment is equivalent to a development version of your production and wi
 #### Pre-requisites:
 - The implementation would be under the experimental flag.
 - Only supported for Devfile v2.0.0 components.
-- Only supported for Kubernetes/ OpenShift targets.
+- Only supported for Kubernetes/OpenShift targets.
 
 #### odo deploy 
 This command will build a container image for the application and deploy it on the target Kubernetes environment. 
 
 Flags:
- - `--tag`: The tag to be used for the built application container image - `<registry>/<org>/<name>:<tag>` (mandatory).
- - `--service-account`: The service account for running privileged containers and push access to the image registry.
- - `--credentials`: The credentials needed to push the image to the container image registry (optional).
+ - `--tag`: The tag to be used for the built application container image - `<registry>/<org>/<name>:<tag>` (optional)
+    - Openshift: the tag can be generated based on project details and internal image registry can be used
+    - Kubernetes: the user must provide the fully qualified registry and tag details.
+ - `--credentials`: The credentials needed to push the image to the container image registry (optional)
+    - Openshift: default builder account for BuildConfig is used
+    - Kubernetes: the user must provide the credentials
+ - `--force`: Delete the existing deployment (if present) and re-create a new deployment.
 
 #### odo deploy delete
 This command will delete any resources created by odo deploy.
@@ -85,23 +89,12 @@ The deployment manifest could be templated to help with replacing key bits of in
 - CONTAINER_IMAGE
 - PORT
 
-For example: 
-A deployment based on Runtime component operator (https://operatorhub.io/operator/) might look like this: 
-```
-apiVersion: app.stacks/v1beta1
-kind: RuntimeComponent
-metadata:
-    name: PROJECT_NAME
-spec:
-    applicationImage: CONTAINER_IMAGE 
-    service:
-        type: ClusterIP
-        port: PORT
-    expose: true
-    storage:
-        size: 2Gi
-        mountPath: "/logs"
-```
+Examples: 
+- Standard Kubernetes resources (deployment, service, route): https://github.com/groeges/devfile-registry/blob/master/devfiles/nodejs/deploy_deployment.yaml
+
+- Operator based resources e.g. Runtime component operator (https://operatorhub.io/operator/): https://github.com/groeges/devfile-registry/blob/master/devfiles/nodejs/deploy_runtimecomponent.yaml
+
+- Knative service: https://github.com/groeges/devfile-registry/blob/master/devfiles/nodejs/deploy_knative.yaml
 
 ### odo deploy
 This command will perform the following actions:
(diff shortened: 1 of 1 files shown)


girishramnani:
Some questions -

- Would the `odo log` work with the deployed slim container?
- Would the `odo exec` command be allowed to run commands on the deployed prod container?
- What kind of kubernetes resource would the Image be deployed as? Like a low level `Pod` resource or a replicated `Deployment` or something else?
- Would we allow configure health checks (liveness probes)?

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+This command will allow a user to perform inner-loop and then test the outer-loop, so the application is truly ready for checking into git and for pipelines to take over. Here's how a typical application development flow might look like: 
+
+User flow: 
+1. `odo create <component> <mycomponent>` - This initializes odo component in the current directory.
+1. Edit the project source code to develop the application.
+1. `odo URL create` - This stores URL information (host, port etc.) for accessing the application on the cluster (if not done already).
+1. `odo push` - This runs the application source code using inner-loop instructions from devfile.
+1. Validate the running application is working as intended.
+1. Iterate over steps 2 and beyond (as needed). 
+1. `odo deploy` - This would build a new clean image and deploy it to the target cluster using outer-loop instructions from devfile and user-provided arguments.
```
kadel, line 31: Would this be deploying the application "on the side" to the inner-loop so after this stage is complete the application will be running in two instances one in inner-loop mode and the other in outer-loop mode? Or is this more like a transition and the inner-loop run will be deleted?

I'm assuming that it is the first case as that would be more helpful, but it will create additional challenges like preventing name collisions when deployed to the same namespace.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+- Only supported for Devfile v2.0.0 components.
+- Only supported for Kubernetes/ OpenShift targets.
+- `odo deploy` will not make any changes to the user's project.
+
+#### odo deploy 
+This command will build a container image for the application and deploy it on the target Kubernetes environment. 
+
+Flags:
+ - `--tag`: The tag to be used for the built application container image - `<registry>/<org>/<name>:<tag>` (mandatory).
+ - `service-account`: The service account for running privileged containers and push access to the image registry.
```
kadel, line 61: Are the privileged containers going to be required? Most of OpenShift clusters won't allow privileged containers.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+ - `--credentials`: The credentials needed to push the image to the container image registry (mandatory).
```
kadel, line 62: I'm not sure if credentials should be mandatory. It might not be required when pushing to the internal OpenShift registry from inside the cluster. I'll need to check this.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+This command will perform the following actions:
+
+#### Input validation
+- Check if the devfile  is v2.0.0 and that it specifies the expected outer-loop attributes. 
+    - If not provided, display a meaningful error message.
+- Validate all arguments passed by the user. 
+    - If argument values are invalid, display a meaningful error message.
+
+#### Build
+- Start a new pod with a build container (privileged container with `buildah:stable` image).
```
kadel, line 118: As stated above, requiring a privileged container will make this almost impossible to run on OpenShift.

docs/proposals/odo-deploy.md
dharmit, line 61: Is this going to be mandatory? Asking because there's no `--` so it seems more like a positional argument to me. When we talk about `service-account`, aren't we diving into the OpenShift/k8s terminology that we want the odo users to be shielded from?

kadel:
> * Would the `odo log` work with the deployed slim container?

yes, but we can address this later.

> * Would the `odo exec` command be allowed to run commands on the deployed prod container?

yes, but we can address this later.

> * What kind of kubernetes resource would the Image be deployed as? Like a low level `Pod` resource or a replicated `Deployment` or something else?

That will be complete under the devfile creator control. `deployment-manifest:` could point to anything. It can be Pod, Deployment, Knative Service, or CR, or even multiple resources-

> * Would we allow configure health checks (liveness probes)?

Same as above, it will depend on what will be defined in `deployment-manifest` should be able to add there almost anything they want.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+- PROJECT_NAME
+- CONTAINER_IMAGE
+- PORT
+
+For example: 
+```
+apiVersion: app.stacks/v1beta1
+kind: RuntimeComponent
+metadata:
+    name: PROJECT_NAME
```
dharmit, line 96: `metadata.name` generally (correct me if I'm wrong) indicates name of the resulting object that's created on k8s. In this case `RuntimeComponent`. So is the `RuntimeComponent` going to be named same as `PROJECT_NAME`? Does that mean only one `RuntimeComponent` per project?

**DISCLAIMER**: I don't know what a `RuntimeComponent` is. Asking because this looks like anomaly or maybe misinterpretation on my part.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+- Start a new pod with a build container (privileged container with `buildah:stable` image).
+- Mount the image registry credentials passed by the user.
+- Copy the application source code into the build container.
+- Fetch the dockerfile using URI in the attributes of the devfile and make it available in the container.
+- Execute `buildah bud` command using src code, dockerfile and tag information.
+- Push the built image using the tag and credentials.
+- Delete the build container and the pod.
+
+#### Deploy
+- Delete any existing deployments (if the user runs `odo deploy` multiple times).
```
dharmit, line 127: Probably an implementation level point, but I think the behaviour should be similar to what `odo push` and `odo push -f` do. Presently, you need to do `odo push -f` if you'd like to delete the component pushed by an earlier run of `odo push`?

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+- Fetch the deployment manifest using URI in the attributes of the devfile. and make it available in the container.
+- Replace templated text in the deployment manifest with relevant values:
+    - PROJECT_NAME: name of odo project
```
dharmit, line 130: I think this brings me back to the point I made earlier about `metadata.name`. IMO, `PROJECT_NAME` can't be in the `metadata.name`. Maybe, `metadata.generateName`. But that depends on whether my understanding in the previous comment is correct.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+    - CONTAINER_IMAGE: `tag` for the built image
+    - PORT: URL information in env.yaml
+- Apply the new deployment manifest.
+- Save the deployment manifest in `.odo` folder.
+- Provide the user with a URL for accessing the deployed application.
```
dharmit, line 135: `odo deploy` automagically creates a URL without having to use `odo url create`? Where does it get the info to create URL from? Sorry if I'm missing something from devfile side of the discussion on this.

docs/proposals/odo-deploy.md
groeges: Yes, this would be the first case with there being 2 deployments, the inner-loop and then the outer-loop, each being independent of each other.  We would be looking at appending a specific value  to the names of the created objects, ie `-deploy` to ensure there are no name clashes with the inner-loop.  The value to be appended is still under discussion.

docs/proposals/odo-deploy.md
groeges: My understanding is that a privileged container is required to build and push the image to the internal registry. If there are ways this can be done without using a privileged container then please let us know (all or testing failed without using privileged!!)

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+
+There is an existing proposal for adding on outer-loop information (including support for multiple strategies for building and deploying the projects) to Devfile 2.1.0: 
+https://github.com/devfile/kubernetes-api/issues/49 
+
+It would be useful to start the design/development of a simpler version of `odo deploy` with devfile 2.0.0 that covers:
+- single build strategy - Dockerfile built using `buildah`
+- single deployment strategy - Kubernetes manifest deployed using `kubectl`.
+
+## Motivation
+`odo` is limited to development inner-loop for a project and there is no support for outer-loop - build a slim/production container image for your application and deploy the build container. It would be very useful for developers to be able to try inner-loop and then transition over to the outer-loop using odo. The outer-loop information could be provided by the application stack (devfile) to avoid developers having to worry about these aspects.
```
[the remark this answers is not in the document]
neeraj-laad: I don't think `odo deploy` should be the way for devs to deploy their apps as real workloads on production environments. Rather, this is the way they can test that their app is ready for outer-loop based deployments which will be typically done by pipelines.  Hence, the proposal is to use same Kubernetes/openshift target that they are using for development. 

If we really wanted to allow them yo specify a different target namespace or context information. I would prefer to avoid that though.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+https://github.com/devfile/kubernetes-api/issues/49 
+
+It would be useful to start the design/development of a simpler version of `odo deploy` with devfile 2.0.0 that covers:
+- single build strategy - Dockerfile built using `buildah`
+- single deployment strategy - Kubernetes manifest deployed using `kubectl`.
+
+## Motivation
+`odo` is limited to development inner-loop for a project and there is no support for outer-loop - build a slim/production container image for your application and deploy the build container. It would be very useful for developers to be able to try inner-loop and then transition over to the outer-loop using odo. The outer-loop information could be provided by the application stack (devfile) to avoid developers having to worry about these aspects. 
+
+`odo deploy` can be a good way to assure the developer that their application can be built and deployed successfully using the build/deploy guidance that comes from devfile.
```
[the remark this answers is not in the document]
neeraj-laad: You are right `odo push` also runs the app on k8s. During development, we use a base runtime image and sync the user's application code and run it. We do not create a container image for the application and deploy that application image. 

Typically for outer loop, you will build a container image for the application without any dev tools etc. and then deploy that built image. This can be significantly smaller and different to what we use in development. 

The proposal is allowing users to test that this flow will work based on info from devfile.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+
+It is not meant to replace GitOps/pipeline-based deployments for governed environments (test, stage, production). However, both `odo deploy` and `odo pipelines` must honour the build and deployment information provided by the application stack (devfile).
+
+## User Flow
+
+This command will allow a user to perform inner-loop and then test the outer-loop, so the application is truly ready for checking into git and for pipelines to take over. Here's how a typical application development flow might look like:
```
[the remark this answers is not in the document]
neeraj-laad: `odo push` is great and is all a developer needs for the development aspect or the inner loop capabilities. Though, there are certain aspects that might be different when a really CI pipeline builds the app into a container image and deploy it.

Lets take example of a typical node application.
- During development one typically uses `node` base image while in production people want to use node:slim image as it does not container npm, or any other dev tools.
- During development `npm install` is run to get all the dependencies, but for production builds we would typically run `npm install --production` which does not install any dev-time dependencies.

So the overall container that you usually get for a final production build can be different that what we typically use during development. 

That said even the deployment itself could be different. I might want to deploy my app using an operator for my runtime, or deploy is as a Knative service. All of these aspects are not rightly covered by the `odo push` as that is more about developing the app.

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,152 @@
...
+1. Edit the project source code to develop the application.
+1. `odo URL create` - This stores URL information (host, port etc.) for accessing the application on the cluster (if not done already).
+1. `odo push` - This runs the application source code using inner-loop instructions from devfile.
+1. Validate the running application is working as intended.
+1. Iterate over steps 2 and beyond (as needed). 
+1. `odo deploy` - This would build a new clean image and deploy it to the target cluster using outer-loop instructions from devfile and user-provided arguments.
+1. Validate the deployed application is working as intended.
+1. Iterate over steps 2 and beyond (as needed).
+1. Optionally, run `odo deploy delete` to clean up resources created with `odo deploy`.
+1. Push your code to Git - ready for sharing with your team and for CI/CD pipelines to take over.
```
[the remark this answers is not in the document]
neeraj-laad: I have to admit I agree with your view on inner/outer loop. :)

That said `odo deploy` is trying to bridge that transition from inner to outer loop, and assure that the CI pipeline that gets triggered from developer's `git commit/push` will work and deploy the application, exactly how the dev wanted. 

In my view: 
- `odo deploy` - test the outer loop.
- `odo pileines ...` - real outer loop.

docs/proposals/odo-deploy.md
neeraj-laad: You are right, they should only be needed for generic k8s, and not needed for OpenShift. I'll update and reflect that.

docs/proposals/odo-deploy.md
neeraj-laad: Apologies, I should have made this clearer. I chose an operator based deployment for the example as it shows the diff fro ma typical k8s deployment that we do during inner loop. 

https://operatorhub.io/operator/runtime-component-operator is the operator I used in this example. But, you could imagine devfile creator creating a Knative service, or other types if they wish to.

docs/proposals/odo-deploy.md
neeraj-laad: Would like to get your thoughts on this as we are not experts on OpenShift. Can we detect we are replying to openshift and use the default builder service account? I guess that has all the permissions needed.

docs/proposals/odo-deploy.md
neeraj-laad: Sure, I'm not an expert on this either. But, the point is that the devfile creator will be able to provide a manifest with these placeholders and application developers don't need to worry about it.

docs/proposals/odo-deploy.md
neeraj-laad: The initial idea was to expect user has done `odo url create` during development so we know key information about the PORT etc. Rest of the bits about route should be handled/provided by the deployment manifest. 

In the example of runtime component operator listed above, the operator reads the CR, and then creates necessary routes etc. If that is not the case then the route should be part of the deployment manifest too with placeholder values for app specific info.

@groeges  please correct me if I got this wrong.

neeraj-laad: The initial idea was to expect user has done `odo url create` during development so we know key information about the PORT etc. Rest of the bits about route should be handled/provided by the deployment manifest. 

In the example of runtime component operator listed above, the operator reads the CR, and then creates necessary routes etc. If that is not the case then the route should be part of the deployment manifest too with placeholder values for app specific info.

@groeges  please correct me if I got this wrong.

neeraj-laad:
@girishramnani 

> * Would the `odo log` work with the deployed slim container?
I have to admit I had not thought about this up till now, so based on current thinking `No`. It would be good to not intermingle the commands for inner-loop/outer-loop too much. But we can put this as a future piece if we think it might be useful. 

> * Would the `odo exec` command be allowed to run commands on the deployed prod container?
I would think `No` - If this is a production-like container, I do not see why you would need to do this. Can you think of a reason why this would be useful?

> * What kind of kubernetes resource would the Image be deployed as? Like a low level `Pod` resource or a replicated `Deployment` or something else?
We would like to support any Kubernetes resource. So devfile creator can decide what makes most sense. standard k8s deployment, knative service, operator custom resource etc. 

> * Would we allow configure health checks (liveness probes)?
My current thought is that we will not do anything over and above what the devfile creator specifies. If they want specific health/liveness probes they need to provide those in the deployment manifest. or maybe consider this as a future piece.

docs/proposals/odo-deploy.md
neeraj-laad: Sorry my mistake. I have fixed this in a later push.

Regarding terminology, I agree we should think of better names. I'm happy if someone what's to suggest a better name for the flags etc.

docs/proposals/odo-deploy.md
neeraj-laad: Agree. It seems like a reasonable thing to do and keep them both consistent. I'll update the proposal with this.

docs/proposals/odo-deploy.md
groeges: Yes, the proposal was that the `odo url create` had already been done and the `odo deploy` would use that info to substitute the relevant PORT info into the manifest (if these template variables had been included). If the manifest wasn't configured to produce a Route/Endpoint, ie explicitly in the manifest or via an Operator then one would not be automatically created and no URL would be returned.   

This is one of the benefits of testing the deployment using `odo deploy` in order to test that the application is suitable to be deployed using the manifest etc that will be used in full outer-loop support with pipelines etc.

docs/proposals/odo-deploy.md
kadel: Don't want to go too deep into the implementation details. But we need to be careful about this. Modifying resource names might have consequences. For example, k8s `Service` name maps to a name in internal k8s DNS. If we change its name without the user being awaited other services might not be able to connect to the deployed app.

docs/proposals/odo-deploy.md
kadel: > Can we detect we are replying to openshift and use the default builder service account?

I think that even builder service account won't be able to run privileged containers.


This is a big deal. We need to figure out a way to do build on a default OpenShift configuration without requiring privileged containers.

docs/proposals/odo-deploy.md
sbose78: @neeraj-laad yea, let's use a more commonly known `Kind` here please :)

docs/proposals/odo-deploy.md
sbose78: @neeraj-laad @kadel Maybe go for a Kaniko build, it's easy to get non-privileged builds working ?

This is a tried and tested container spec from our `[buildv2](https://github.com/redhat-developer/build/)` project

```
      name: step-build-and-push
      image: gcr.io/kaniko-project/executor:v0.23.0
      workingdir: /workspace/source
      securityContext:
        runAsUser: 0
        capabilities:
          add:
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - SETGID
            - SETUID
      env:
        - name: DOCKER_CONFIG
          value: /tekton/home/.docker
        - name: AWS_ACCESS_KEY_ID
          value: NOT_SET
        - name: AWS_SECRET_KEY
          value: NOT_SET
      command:
        - /kaniko/executor
      args:
        - --skip-tls-verify=true
        - --dockerfile=$(build.dockerfile)
        - --context=/workspace/source/$(build.source.contextDir)
        - --destination=$(build.output.image)
```

docs/proposals/odo-deploy.md
EnriqueL8: We need priviledged access to run the `buildah` container, I remember @sbose78 mentioned that we can run these containers without priviledged access? In generic Kube, we don't need priviledged.

sbose78: * It's a bit tricky to get buildah to do builds without privileges : check out this thread https://github.com/redhat-developer/build/issues/134#issuecomment-619494883 , if we can get to this, awesome.

* Else, see my comment below https://github.com/openshift/odo/pull/3368#discussion_r442193330

sbose78:
@neeraj-laad Could you please address in your proposal why we aren't going to use the BuildConfig Dockerfile strategy in OpenShift ? 

@girishramnani @kadel @dharmit 

Would it be messy if we had a couple of `if-else` blocks where 
* if openshift, use buildconfig and push to imagestream
* if non-openshift, use plain pods, and push to quay.io-ish registry with credentials.

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
sbose78: @dharmit More like ..

1. Inner loop: I can run stuff in a container, wohoo!
2. Outer loop : Let's see if I can build and image out of it and run it like a conventional workload :)
3. Multi-stage outer loop / gitops : We break out of the dev env, and manage this workload across multiple clusters/stages/namespaces in Git. 

:)

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
wtam2018: It sounds like `odo deploy` is not intended to make use of any outer loop information in devfile if and when it becomes available.

kadel:
> Would it be messy if we had a couple of `if-else` blocks where
> 
> * if openshift, use buildconfig and push to imagestream
> * if non-openshift, use plain pods, and push to quay.io-ish registry with credentials.

This would be ok. We already doing something similar for URLs where odo creates Routes when working with OpenShift and Ingress when working with Kubernetes

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
dharmit: > Rather, this is the way they can test that their app is ready for outer-loop based deployments which will be typically done by pipelines. Hence, the proposal is to use same Kubernetes/openshift target that they are using for development.

Makes a lot of sense to me now. I know that we're working on adding `odo pipelines` support but didn't know about the whole idea. Thanks for explaining, Neeraj. :+1:

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
dharmit: Makes sense. Thank you. :+1:

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
dharmit: Perfect. Makes sense. :+1:

docs/proposals/odo-deploy.md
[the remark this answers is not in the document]
dharmit: Thanks for taking time to explain this, folks. After making a few comments, I had realized that I was stuck. :)

docs/proposals/odo-deploy.md
dharmit: Thinking out loud here.

A simpler alternative could be `user` which `odo` should be smart enough to map to a service-account/user account. Documentation could explain things in depth and it's a good (preferred?) place to use some k8s/openshift lingo.

docs/proposals/odo-deploy.md
dharmit: I'm not pointing to `RuntimeComponent` kind here. I'm trying to point that if we use `PROJECT_NAME` (which I assume is the name of a k8s/ocp namespace), it would probably lead to odo being able to create only one `RuntimeComponent` per project/namespace. There could be multiple applications a user is developing within the same k8s/ocp namespace, no?

Apologies if this falls into nitpick category. I felt like commenting because I understood it to be a template/default manifest that we would create when user's interested in going beyond `odo push` to `odo deploy`.

docs/proposals/odo-deploy.md
dharmit: This is how `odo push` works currently:
```sh
$ odo push                                                                                                                                                                                    
                                                                                                                                                                                              
Validation                                                                                                                                                                                    
 ✓  Validating the devfile [24381ns]                                                                                                                                                          
                                                                                                                                                                                              
Applying URL changes                                                                                                                                                                          
                                                                                                                                                                                              
Syncing to component nodejs                                                                                                                                                                   
 ✓  Checking file changes for pushing [556622ns]                                                                                                                                              
 ✓  Syncing files to the component [209ms]                                                                                                                                                    
                                                                                                                                                                                              
 ✓  Executing devBuild command "npm install" [2s]                                                                                                                                             
 ✓  Executing devRun command "nodemon app.js" [1s]                                                                                                                                            
                                                                                                                                                                                              
$ odo push

Validation
 ✓  Validating the devfile [21294ns]

 ✓  Waiting for component to start [5ms]

Applying URL changes

Syncing to component nodejs
 ✓  Checking file changes for pushing [258076ns]

```

Key part being:
```
Syncing to component nodejs
 ✓  Checking file changes for pushing [258076ns]
```
Now, I'm not sure if `odo deploy`'s going to, or even required to, check for any file changes between two runs of `odo deploy`. As you pointed, keeping behaviour consistent would be helpful, IMO, if file changes matter to `odo deploy` as they currently do to `odo push`.

docs/proposals/odo-deploy.md
dharmit: Makes sense. Thanks for the explanation @neeraj-laad & @groeges :+1:

docs/proposals/odo-deploy.md
neeraj-laad: Have updated the proposal to use BuildConfig / Kaniko to remove need for privileged access.

docs/proposals/odo-deploy.md
neeraj-laad: Have updated the proposal to go for BuildConfig/Kaniko based builds

docs/proposals/odo-deploy.md
neeraj-laad: @dharmit Thanks! I like your suggestion. Further discussions with @kadel and @sbose78 meant that we have removed this argument and will use BuildConfig's default account or Kaniko (which des not need a service account. One less argument.. :)

docs/proposals/odo-deploy.md
neeraj-laad: Have updated the proposal to include multiple examples, standard Kubernetes deployment, knative service, operator based deployments. I hope this makes it clearer.

docs/proposals/odo-deploy.md
neeraj-laad: Have updated the proposal to just create/update the deployment by default.. and when used with -f, it will delete and then re-create.

docs/proposals/odo-deploy.md
neeraj-laad: If we foresee multiple projects from same user with same name which need to be deployed, we might have to add a  random suffix to make them unique. Is that what you are thinking of?

neeraj-laad:
@sbose78 @kadel I have updated to proposal to use BuildConfig if present on cluster or use kaniko if not. 

This will mean the first PR could provide OpenShift support using BuildConfig.  and a follow on PR could add support for generic Kubernetes with Kaniko.

docs/proposals/odo-deploy.md
dharmit: @neeraj-laad I think it's safe to leave this part for implementation detail. I'm giving an example below only to explain the scenario that I'm concerned about. I'm sure that the dev who's implementing this will have this scenario covered or would encounter it sooner rather than later. :wink: 

The "random suffix" part is what I'm thinking of for a scenario like below:
1. I have an OpenShift/k8s project called `myproject`.
1. I am an enterprise developer working on a microservice application that's got:
    - nodejs frontend
    - java backend
    - postgresql database
1. To have a working end product, I need to deploy all the above three in same OpenShift/k8s namespace.
1. In the context of odo, there's a unique component for nodejs and java part of above architecture while postgresql is a service.
1. If the manifest will have `metadata.name` value set as `PROJECT_NAME` for both nodejs and java components, there will be a conflict in creating a k8s resource and we'll be able to create only one resource with the value of `PROJECT_NAME` value. So, it might help to have a random suffix (`metadata.generateName` can be handy.)

neeraj-laad: Thanks for explaining this @dharmit 

With PROJCT_NAME, I was referring to the name of this specific component / micro service (node-frontend, or java-backend etc. I think the terminology might be causing the confusion.

Is there a specific name in odo for each micro-service/component the user is developing? Is COMPONENT_NAME a better term?

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,148 @@
...
+### Detailed design: 
+
+### Devfile
+
+For the initial implementation, we could use devfile v2.0.0 and capture basic outer-loop information as `metadata` on the devfile. `odo` could look for  specific keys, while other tools like Che could ignore them.
+
+For example: 
+```
+metadata:
+    build-dockerfile: <URI>
```
kadel, line 79: Would it make sense to highlight the fact that this is probably just a temporary location that will be replaced with proper fields in future devfile spec versions?
Kubernetes API does something similar where they use annotations containing `alpha` group before they settle down on the proper placement.

For example, we could do something like this:
```yaml
metadata:
  alpha.build-dockerfile
  alpha.deployment-manifest
```

docs/proposals/odo-deploy.md
dharmit: Ah, now it makes sense. Yes, "COMPONENT_NAME" is the term on odo side of the things.

docs/proposals/odo-deploy.md
neeraj-laad: Makes sense. I'll update the proposal to reflect this.

docs/proposals/odo-deploy.md
neeraj-laad: I'll update the PR to reflect that.. Apologies for the confusion.

kadel:
/lgtm

overriding CI, we don't need tests for this PR
/override ci/prow/v4.5-integration-e2e 
/override ci/prow/v4.4-integration-e2e 
/override ci/prow/v4.3-integration-e2e 
/override ci/prow/v4.2-integration-e2e

jaideepr97:
Hi folks, based on my understanding of this proposal, I had a few questions (apologies for any trivial/previously answered questions) :-
1. odo deploy flags mentions multiple ways to handle flag inputs (registry credentials etc) depending on the type of target( k8/OS)  - is the target specified in the devfile or supplied in some other way?

2. I understand that as of now odo deploy is intended to only be used for deployments in the dev environment. However, I was wondering, given enough time to mature through successive improvements, if there was any likelihood of configuring existing pipelines to leverage odo deploy internally to deploy applications to prod and other environments down the line 

A couple peripheral questions that also seemed relevant to me : 

3. Is the intention of having a devfile that the user would not have to worry about any of the deployment details at all? Because as per my understanding the devfile points to variable resources like the deployment manifest and source URL which would probably be specific to an application and can't just be picked up by odo deploy from a stock devfile for a particular language/framework

4. Has the structure of the devfile 2.0 been fixed or is it still being debated? The link for the documentation (https://devfile.github.io/website/) provided on the kubernetes-api github page seems to be broken (https://github.com/devfile/kubernetes-api)

Appreciate any inputs!

neeraj-laad:
@jaideepr97 Sorry missed this message as the PR got merged. 

> 1. odo deploy flags mentions multiple ways to handle flag inputs (registry credentials etc) depending on the type of target ( k8s/OS)  - is the target specified in the devfile or supplied in some other way?

The intent was to sue a fully qualified image name (registry, repo, name, version etc.) to be supplied as the `--tag` argument and corresponding credentials to be supplied as `credentials`. That said the first implementation will be using Openshift BuildConfig so we will not need tag and credentials. for that case.

> 2. I understand that as of now odo deploy is intended to only be used for deployments in the dev environment. However, I was wondering, given enough time to mature through successive improvements, if there was any likelihood of configuring existing pipelines to leverage odo deploy internally to deploy applications to prod and other environments at some point down the line

It can be though, using a developer CLI tool in a pipeline was a bit odd. Instead the `odo pipelines init/bootstrap` will use the same code to provide appropriate build/deploy artifacts that the pipelines can use. This will allow the pipelines to be generic and not have to rely on odo.
 
> 3. Is the intention of having a devfile that the user would not have to worry about any of the deployment details at all? Because as per my understanding the devfile points to variable resources like the deployment manifest and source URL which would probably be specific to an application and can't just be picked up by odo deploy from a stock devfile for a particular language/framework

The intent is that the devfile can provide default dockerfile/deployment manifests that will work with standard apps for that technology/framework. If the developer wants to override this, they do have the capability to do so by utilising standard inheritance and override capabilities that will be available to devfiles.

> 2. Has the structure of the devfile 2.0 been fixed or is it still being debated? The link for the documentation (https://devfile.github.io/website/) provided on the kubernetes-api github page seems to be broken (https://github.com/devfile/kubernetes-api)

I would think this would be temporary. It is close to getting finalised but this feature was suggesting no changes to devfile spec. As a proper syntax and first class support will be added as part of Devfile 2.1.0

docs/proposals/odo-deploy.md
```diff
@@ -0,0 +1,148 @@
...
+
+## User flow
+
+This command will allow a user to test the transition from inner-loop t outer-loop, so the application is truly ready for pipelines to take over. Here's how a typical application development flow might look like: 
+
+User flow: 
+1. `odo create <component> <mycomponent>` - This initializes odo component in the current directory.
+1. Edit the project source code to develop the application.
+1. `odo url create` - This stores URL information (host, port etc.) for accessing the application on the cluster (if not done already).
+1. `odo push` - This runs the application source code using inner-loop instructions from devfile.
```
sbose78, line 28: @EnriqueL8 Is `odo push` needed to get to `odo deploy`, I guess No ? :)

PR #3400, opened 2020-06-22
Odo test cmd for devfile 2.0

### LINKED ISSUE #3284: Support application testing defined in the devfile v2
/kind user-story

## User Story
As a user, I want to test my application using the test defined in a devfile.  Devfile 2.0 supports defining commands belongs to the `test` group (https://github.com/devfile/kubernetes-api/issues/27). The test command will be used for running automated testing on the application.

## Acceptance Criteria 
- [ ] `odo test` will run test command (command belongs to the `test` group) specified in the command that belongs to the group `test` defined in the devfile
- [ ] odo uses this logic to decide which test command to run on `odo test`:
    - if only one test command is define, use that command regardless of whether that command with isDefault is set to true or not.
    - if more than one test command, execute the one with `isDefault` set to true
    -`odo test --test-command <command name>` will execute the command with the given command name regardless of a default test command is defined or not.
   - if more than one test command but no default is defined, error out if `--test-command` is not use.

## Links 

/kind user-story
/area devfile

yangcao77:
**What type of PR is this?**

/kind user-story
/area devfileV2

`odo test` will run test command specified in the command that belongs to the group test defined in the devfile

odo uses this logic to decide which test command to run on odo test:

- if only one test command is define, use that command regardless of whether that command with isDefault is set to true or not.
- if more than one test command, execute the one with isDefault set to true
- `odo test --test-command <command name>` will execute the command with the given command name regardless of a default test command is defined or not.
- if more than one test command but no default is defined(or if multiple default commands are defined), error out if --test-command is not use.

Fixes #3284 

Automation:
run `make test-cmd-devfile-test` and `make test-cmd-docker-devfile-test`

Manual:
(The following shows docker scenario, kube scenario should be similar)
Use the devfile.yaml within this PR: `devfile-with-testgroup.yaml`

run `odo create` and `odo push`. 
run `odo test`
`docker exec -it <containerID> ls -la /projects/nodejs-web-app/` should see test1 directory exist

run ` odo test --test-command test2`
`docker exec -it <containerID> ls -la /projects/nodejs-web-app/` should see test2 directory exist


yangcao77:
Mark as a draft pr since this PR is blocked by https://github.com/openshift/odo/pull/3291
Copied some functions from #3291 in order to make it work. Need to rebase after #3291 gets merged.

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -140,6 +141,27 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+func (a Adapter) Test(testcmd versionsCommon.DevfileCommand, show bool) (err error) {
```
mik-dass, line 144: Please add some comments for the function

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -161,6 +161,25 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+// Test runs the devfile test command
+func (a Adapter) Test(testcmd versionsCommon.DevfileCommand, show bool) (err error) {
+	componentExists := utils.ComponentExists(a.Client, a.ComponentName)
+	if !componentExists {
+		return fmt.Errorf("component does not exist, a valid component is required to run 'odo test'")
+	}
```
mik-dass, line 169: Do we really need this call since we are getting the pod just after it? Won't that be enough?

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,121 @@
...
+	return to.RunTestCommand()
+}
+
+// NewCmdTest implements the odo tets command
+func NewCmdTest(name, fullName string) *cobra.Command {
+	to := NewTestOptions()
+	testCmd := &cobra.Command{
+		Use:     name,
+		Short:   "Run test command defined in devfile",
+		Long:    "Run test command defined in devfile",
```
mik-dass, line 103: `run the`

vendor/k8s.io/kubectl/go.sum
```diff
@@ -323,9 +323,12 @@ gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
 gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
 honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.0-20190106161140-3f1c8253044a/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+k8s.io/api v0.17.1 h1:i46MidoDOE9tvQ0TTEYggf3ka/pziP1+tHI/GFVeJao=
 k8s.io/api v0.17.1/go.mod h1:zxiAc5y8Ngn4fmhWUtSxuUlkfz1ixT7j9wESokELzOg=
+k8s.io/apimachinery v0.17.1 h1:zUjS3szTxoUjTDYNvdFkYt2uMEXLcthcbp+7uZvWhYM=
 k8s.io/apimachinery v0.17.1/go.mod h1:b9qmWdKlLuU9EBh+06BtLcSf/Mu89rWL33naRxs1uZg=
 k8s.io/cli-runtime v0.17.1/go.mod h1:e5847Iy85W9uWH3rZofXTG/9nOUyGKGTVnObYF7zSik=
+k8s.io/client-go v0.17.1 h1:LbbuZ5tI7OYx4et5DfRFcJuoojvpYO0c7vps2rgJsHY=
```
mik-dass, line 331: I think these changes need to be removed

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,121 @@
...
+// NewTestOptions creates a new TestOptions instance
+func NewTestOptions() *TestOptions {
+	return &TestOptions{}
+}
+
+// Complete completes TestOptions after they've been created
+func (to *TestOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	if !experimental.IsExperimentalModeEnabled() {
+		return fmt.Errorf("'odo test' is only supported under experimental mode. Run 'odo preference set experimental true' to enable experimental mode. ")
+	}
```
mik-dass, line 56: Since https://github.com/openshift/odo/pull/3400/files#diff-ae6461ee1175c7152132051efcd0bbecR205 is already under experimental mode, I think we can remove this check.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -361,7 +380,7 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	command, ok := commandsMap[versionsCommon.BuildCommandGroupType]
 	if ok {
 		compInfo.ContainerName = command.Exec.Component
-		err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
+		err = exec.ExecuteDevfileBuildOrTestAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
```
mik-dass, line 383: Maybe rename to something more simpler as even the init command implementation seems to also use this function.

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: This checks if the deployment exist; and we should show different error message if there is a problem when getting pod.

pkg/devfile/adapters/docker/component/adapter.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

vendor/k8s.io/kubectl/go.sum
yangcao77: removed

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,121 @@
...
+	to.devfilePath = filepath.Join(to.componentContext, DevfilePath)
+	envInfo, err := envinfo.NewEnvSpecificInfo(to.componentContext)
+	if err != nil {
+		return errors.Wrap(err, "unable to retrieve configuration information")
+	}
+	to.EnvSpecificInfo = envInfo
```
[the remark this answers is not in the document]
yangcao77: Good point. Actually the assignment in context.go has been included in this commit made by Girish 4 days ago: https://github.com/openshift/odo/commit/917da80ea8e0fae2b8f8ea440e779012e22c87d3
simply removed the above code

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: renamed to be `ExecuteDevfileCommands`
updated the comment for the function, `ExecuteDevfileCommands executes the devfile init, build and test command actions`

pkg/exec/devfile.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ExecuteDevfileBuildAction executes the devfile build command action
-func ExecuteDevfileBuildAction(client ExecClient, exec common.Exec, commandName string, compInfo adaptersCommon.ComponentInfo, show bool, machineEventLogger machineoutput.MachineEventLoggingClient) error {
+// ExecuteDevfileCommands executes the devfile init, build and test command actions
+func ExecuteDevfileCommands(client ExecClient, exec common.Exec, commandName string, compInfo adaptersCommon.ComponentInfo, show bool, machineEventLogger machineoutput.MachineEventLoggingClient) error {
```
maysunfaisal, line 15: probably rename this `ExecuteDevfileCommandSynchronously` to
1. differentiate from `ExecuteDevfileRunAction`
2. init, build and test are all synchronous/blocking execution

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,109 @@
...
+// Validate validates the TestOptions based on completed values
+func (to *TestOptions) Validate() (err error) {
+	devObj, err := devfileParser.ParseAndValidate(to.devfilePath)
+	if err != nil {
+		return errors.Wrap(err, "fail to parse devfile")
+	}
+	if reflect.DeepEqual(devObj.Ctx.GetApiVersion(), "1.0.0") {
+		return fmt.Errorf("'odo test' is not supported in devfile 1.0.0")
+	}
+	to.testCommand, err = adaptercommon.GetTestCommand(devObj.Data, to.commandName)
```
maysunfaisal, line 70: I would say get the test command in the adapter and not in the cli code, to be consistent with rest of odo since all the devfile cli is doing the other way ie just pass the command name string and get it in the adapter code

pkg/devfile/adapters/common/interface.go
```diff
@@ -1,10 +1,15 @@
...
+import (
+	versionsCommon "github.com/openshift/odo/pkg/devfile/parser/data/common"
+)
+
 // ComponentAdapter defines the functions that platform-specific adapters must implement
 type ComponentAdapter interface {
 	Push(parameters PushParameters) error
 	DoesComponentExist(cmpName string) bool
 	Delete(labels map[string]string) error
+	Test(testcmd versionsCommon.DevfileCommand, show bool) error
```
maysunfaisal, line 12: I think when we wrote the interface, there was a conscious effort to make it as low level as possible. So we should not use other structs like `versionsCommon.DevfileCommand`. It should pass the command name as `string` and pls check my other related comment on pkg/odo/cli/component/test.go

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -408,6 +427,18 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
...
+func (a Adapter) execTestCmd(testcmd versionsCommon.DevfileCommand, podName string, containers []corev1.Container, show bool) (err error) {
+	compInfo := common.ComponentInfo{
+		PodName: podName,
+	}
+	compInfo.ContainerName = testcmd.Exec.Component
+	err = exec.ExecuteDevfileCommands(&a.Client, *testcmd.Exec, testcmd.Exec.Id, compInfo, show, a.machineEventLogger)
+	if err != nil {
+		return err
+	}
+	return
```
maysunfaisal, line 439: you can just do `return err` here, no need for the if block since its returning right after

pkg/odo/cli/component/devfile.go
```diff
@@ -181,6 +181,36 @@ func (do *DeleteOptions) DevfileComponentDelete() error {
...
+			Namespace: to.KClient.Namespace,
+		}
+		platformContext = kc
+	}
+
+	devfileHandler, err := adapters.NewComponentAdapter(componentName, to.componentContext, devObj, platformContext)
+	if err != nil {
+		return err
+	}
+	return devfileHandler.Test(to.testCommand, true)
```
maysunfaisal, line 211: we should not hardcode `show` to `true`, get it from the test cli show flag

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,109 @@
...
+		Args:    cobra.MaximumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(to, cmd, args)
+		},
+	}
+
+	// Add a defined annotation in order to appear in the help menu
+	testCmd.Annotations = map[string]string{"command": "main"}
+	testCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
+	testCmd.Flags().StringVar(&to.commandName, "test-command", "", "command name to run")
```
maysunfaisal, line 102: add a `show` flag to test cli. So when the user executes with `--show`, we show the output, otherwise we should not show the output

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,109 @@
...
+	return &TestOptions{}
+}
+
+// Complete completes TestOptions after they've been created
+func (to *TestOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	to.devfilePath = filepath.Join(to.componentContext, DevfilePath)
+	to.Context = genericclioptions.NewDevfileContext(cmd)
+	if !pushtarget.IsPushTargetDocker() {
+		// The namespace was retrieved from the --project flag (or from the kube client if not set) and stored in kclient when initalizing the context
+		to.namespace = to.KClient.Namespace
```
maysunfaisal, line 55: Dont think this is required. In `RunTestCommand()`, you're getting the namespace from `Namespace: to.KClient.Namespace` https://github.com/openshift/odo/pull/3400/files#diff-13ec3fd846d38b18f541990dc88bf6baR202

pkg/devfile/adapters/kubernetes/component/adapter.go
mik-dass: But on a slow cluster both calls will cause performance problems. Since without the pod, it won't run anyways, just check for the pod and error out.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -162,6 +162,25 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+// Test runs the devfile test command
+func (a Adapter) Test(testcmd versionsCommon.DevfileCommand, show bool) (err error) {
+	componentExists := utils.ComponentExists(a.Client, a.ComponentName)
+	if !componentExists {
+		return fmt.Errorf("component does not exist, a valid component is required to run 'odo test'")
+	}
+	pod, err := a.waitAndGetComponentPod(true)
```
mik-dass, line 171: Is there any reason to wait? Might cause performance issues. Instead maybe use https://github.com/openshift/odo/blob/b1d8462e7bce01e41e24e2e853d4203e32d91408/pkg/kclient/pods.go#L153

yangcao77: The pod has to be started in order to run the test command. 

Actually, if you search in our code base, `GetPodUsingComponentName` is never being used. This might because of the same reason, it does not guarantee the pod is in a healthy state.

pkg/devfile/adapters/common/interface.go
yangcao77: Makes sense. Updated

pkg/odo/cli/component/test.go
yangcao77: Updated

pkg/exec/devfile.go
yangcao77: Updated

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: Updated

pkg/odo/cli/component/devfile.go
yangcao77: Updated

pkg/odo/cli/component/test.go
yangcao77: Updated

pkg/odo/cli/component/test.go
yangcao77: Updated

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: Since this PR does both docker and kube; docker logic cannot be applied with the above logic since an odo component in docker can have multiple containers. So if a devfile has 2 components, that translates to 2 containers in docker. But if one of the containers is down; we cannot say a component exist with at least one container in our example.

I can honestly see where your argument is but i think even `odo watch` and `odo push` have these logic which can be refactored more optimally.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -408,6 +432,15 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	return
 }
 
+func (a Adapter) execTestCmd(testcmd versionsCommon.DevfileCommand, podName string, containers []corev1.Container, show bool) (err error) {
```
maysunfaisal, line 435: `testcmd` should be camel case

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -141,6 +141,33 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+// Test runs the devfile test command
+func (a Adapter) Test(testcmd string, show bool) (err error) {
```
maysunfaisal, line 145: unit tests for `Test()`

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -374,6 +374,16 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	return
 }
 
+func (a Adapter) execTestCmd(testcmd versionsCommon.DevfileCommand, containers []types.Container, show bool) (err error) {
```
maysunfaisal, line 377: unit tests for `execTestCmd`

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -162,6 +162,30 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+// Test runs the devfile test command
+func (a Adapter) Test(testcmd string, show bool) (err error) {
```
maysunfaisal, line 166: unit tests for `Test()`

pkg/odo/cli/component/devfile.go
```diff
@@ -181,6 +181,36 @@ func (do *DeleteOptions) DevfileComponentDelete() error {
 	return nil
 }
 
+// RunTestCommand runs the specific command in devfile
```
maysunfaisal, line 184: runs the test* command

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,96 @@
...
+
+// RecommendedCommandName is the recommended command name
+const TestRecommendedCommandName = "test"
+
+// TestOptions encapsulates the options for the odo command
+type TestOptions struct {
+	commandName      string
+	componentContext string
+	devfilePath      string
+	// testCommand      common.DevfileCommand
```
maysunfaisal, line 27: remove commented code

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,96 @@
...
+func (to *TestOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	to.devfilePath = filepath.Join(to.componentContext, DevfilePath)
+	to.Context = genericclioptions.NewDevfileContext(cmd)
+
+	return
+}
+
+// Validate validates the TestOptions based on completed values
+func (to *TestOptions) Validate() (err error) {
+	devObj, err := devfileParser.ParseAndValidate(to.devfilePath)
```
maysunfaisal, line 56: Since we're parsing and validating here, is there a way to pass this `devObj` to `RunTestCommand()` rather than getting it again in `RunTestCommand()`

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,96 @@
...
+		Args:    cobra.MaximumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(to, cmd, args)
+		},
+	}
+
+	// Add a defined annotation in order to appear in the help menu
+	testCmd.Annotations = map[string]string{"command": "main"}
+	testCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
+	testCmd.Flags().StringVar(&to.commandName, "test-command", "", "command name to run")
```
maysunfaisal, line 88: perhaps update the flag description to be consistent with push cli `"Devfile Run Command to execute"`

tests/integration/devfile/cmd_devfile_test_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			output := helper.CmdShouldPass("odo", "test", "--context", context)
+			helper.MatchAllInOutput(output, []string{"Executing test1 command", "mkdir test1"})
+
+			podName := cliRunner.GetRunningPodNameByComponent(cmpName, namespace)
+			output = cliRunner.ExecListDir(podName, namespace, sourcePath)
+			Expect(output).To(ContainSubstring("test1"))
+		})
+
+		It("Should run test command successfully with test-command specified", func() {
```
maysunfaisal, line 98: Can you also add a case where you're specifying test command with an invalid command like:
1. wrong name(ie the command is not present)
2. a build/run command(--test-command="some build command")

```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo test --test-command="devrun"

Executing devfile test command for component testm
 ✗  command group mismatched, command devrun is of group run in devfile.yaml
Maysuns-MacBook-Pro:springboot-ex maysun$ 
Maysuns-MacBook-Pro:springboot-ex maysun$ 
Maysuns-MacBook-Pro:springboot-ex maysun$ 
Maysuns-MacBook-Pro:springboot-ex maysun$ odo test --test-command="some wrong command"

Executing devfile test command for component testm
 ✗  the command "some wrong command" is not found in the devfile
```

Probably add the two cases for docker as well

tests/integration/devfile/cmd_devfile_test_test.go
```diff
@@ -0,0 +1,114 @@
...
+			helper.CmdShouldPass("odo", "push", "--context", context)
+			output := helper.CmdShouldFail("odo", "test", "--context", context)
+
+			Expect(output).To(ContainSubstring("the command group of kind \"test\" is not found in the devfile"))
+		})
+	})
+
+	Context("Should run test command successfully", func() {
+
+		It("Should run test command successfully with only one default specified", func() {
```
maysunfaisal, line 83: Can you also add cases where there are multiple default test commands or no default commands. Probably add the use case for docker as well.

pkg/devfile/adapters/common/command.go
```diff
@@ -279,3 +285,24 @@ func ValidateAndGetDebugDevfileCommands(data data.DevfileData, devfileDebugCmd s
...
+		klog.V(4).Infof("Test command: %v", testCommand.Exec.Id)
+	}
+
+	if !isTestCommandValid {
+		commandErrors := ""
+		if testCmdErr != nil {
+			commandErrors += testCmdErr.Error()
+		}
+		return common.DevfileCommand{}, fmt.Errorf(commandErrors)
+	}
```
maysunfaisal, line 305: if there is no test cmd, but i run `odo test`; it throws an empty err:
```
$ odo test

Executing devfile test command for component testm
 ✗  

```

I think the debug validation is also broken

tests/integration/devfile/cmd_devfile_test_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		It("should show error if no test group is defined", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--context", context, cmpName)
+
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile.yaml"), filepath.Join(context, "devfile.yaml"))
+			helper.CmdShouldPass("odo", "push", "--context", context)
+			output := helper.CmdShouldFail("odo", "test", "--context", context)
+
+			Expect(output).To(ContainSubstring("the command group of kind \"test\" is not found in the devfile"))
```
maysunfaisal, line 77: not sure how this is passing since for me it does not show the error:
```
$ odo test

Executing devfile test command for component testm
 ✗  
```

pls see related comment in command.go file

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return
+}
+
+// Run contains the logic for the odo command
+func (to *TestOptions) Run() (err error) {
+	return to.RunTestCommand()
+}
+
+// NewCmdTest implements the odo tets command
```
maysunfaisal, line 71: test*

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -141,6 +141,33 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "error while retrieving container for odo component %s", a.ComponentName)
+	}
+
+	testCommand, err := common.ValidateAndGetTestDevfileCommands(a.Devfile.Data, testcmd)
+	if err != nil {
+		return errors.Wrap(err, "failed to validate devfile test command")
+	}
+
+	err = a.execTestCmd(testCommand, containers, show)
```
maysunfaisal, line 164: no header in docker

```
$ odo test
 ✓  Executing test2 command "/artifacts/bin/build-container-full.sh" [8s]
```

pkg/devfile/adapters/kubernetes/component/adapter.go
mik-dass: > But if one of the containers is down; we cannot say a component exist with at least one container in our example.

For docker we can write it separately in the docker adapter. Logic in the Kubernetes should be independent of that.

pkg/devfile/adapters/kubernetes/component/adapter.go
mik-dass: > Actually, if you search in our code base, GetPodUsingComponentName is never being used

It has been used here https://github.com/openshift/odo/blob/e367c334c2e6a15f584821c34af4659bb2658a30/pkg/debug/portforward.go#L55 and a occlient version was used here https://github.com/openshift/odo/blob/e367c334c2e6a15f584821c34af4659bb2658a30/pkg/debug/portforward.go#L65.

> it does not guarantee the pod is in a healthy state.

@yangcao77 Even waitAndGetComponentPod() doesn't guarantee that. It might be in a `crashLoopBackOff` state and thus this command will have to wait till timeout and without spinners, it will feel unresponsive. Thus it's better to exit immediately for a better and faster user experience.

pkg/devfile/adapters/common/command.go
yangcao77: This is broken because of this
https://github.com/openshift/odo/blob/cd858a7c3fd897034173819d7d4bdc03dbe2e99d/pkg/devfile/adapters/common/command.go#L61-L66

test and debug type should also be included in the if statement

tests/integration/devfile/cmd_devfile_test_test.go
yangcao77: I doesn't work for me either, not sure how this test passing.  But after I include testgroup in the if statement as mentioned in: https://github.com/openshift/odo/pull/3400#discussion_r451670873


It works for me now. 
```
$ odo test

Executing devfile test command for component nodejs
 ✗  the command group of kind "test" is not found in the devfile
```

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: updated the Kubernetes logic

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: updated

pkg/devfile/adapters/common/command.go
```diff
@@ -279,3 +285,24 @@ func ValidateAndGetDebugDevfileCommands(data data.DevfileData, devfileDebugCmd s
 
 	return debugCommand, nil
 }
+
+// ValidateAndGetTestDevfileCommands validates the test command
+func ValidateAndGetTestDevfileCommands(data data.DevfileData, devfileTestCmd string) (testCommand common.DevfileCommand, err error) {
```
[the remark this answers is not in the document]
yangcao77: Added unit test

pkg/devfile/adapters/common/interface.go
```diff
@@ -5,6 +5,7 @@ type ComponentAdapter interface {
 	Push(parameters PushParameters) error
 	DoesComponentExist(cmpName string) bool
 	Delete(labels map[string]string) error
+	Test(testcmd string, show bool) error
```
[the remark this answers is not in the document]
yangcao77: updated to `testCmd`

pkg/devfile/adapters/docker/adapter.go
```diff
@@ -42,3 +42,8 @@ func (d Adapter) DoesComponentExist(cmpName string) bool {
 func (d Adapter) Delete(labels map[string]string) error {
 	return d.componentAdapter.Delete(labels)
 }
+
+// Test runs devfile test command
+func (d Adapter) Test(testcmd string, show bool) error {
```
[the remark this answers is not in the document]
yangcao77: updated to `testCmd`

pkg/devfile/adapters/docker/component/adapter.go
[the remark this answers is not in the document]
yangcao77: updated to testCmd

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -374,6 +374,16 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
...
+	containerID := utils.GetContainerIDForAlias(containers, testcmd.Exec.Component)
+	compInfo := common.ComponentInfo{ContainerName: containerID}
+	err = exec.ExecuteDevfileCommandSynchronously(&a.Client, *testcmd.Exec, testcmd.Exec.Id, compInfo, show, a.machineEventLogger)
+	if err != nil {
+		return err
+	}
+	return
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/adapters/kubernetes/adapter.go
```diff
@@ -52,3 +52,8 @@ func (k Adapter) Delete(labels map[string]string) error {
 
 	return nil
 }
+
+// Test runs the devfile test command
+func (k Adapter) Test(testcmd string, show bool) error {
```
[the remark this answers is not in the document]
yangcao77: updated to testCmd

pkg/devfile/adapters/kubernetes/component/adapter.go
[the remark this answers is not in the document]
yangcao77: updated to testCmd

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: updated to testCmd

pkg/devfile/adapters/docker/component/adapter.go
yangcao77: added unit test

pkg/devfile/adapters/docker/component/utils.go
yangcao77: added unit test

pkg/odo/cli/component/devfile.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

tests/integration/devfile/cmd_devfile_test_test.go
yangcao77: added integration tests

tests/integration/devfile/cmd_devfile_test_test.go
yangcao77: added integration tests

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/devfile/adapters/docker/component/adapter.go
yangcao77: updated

pkg/devfile/adapters/docker/component/adapter_test.go
```diff
@@ -121,6 +121,153 @@ func TestPush(t *testing.T) {
 
 }
 
+func TestTest(t *testing.T) {
```
maysunfaisal, line 124: 😆 probably name it `TestDockerTest`

pkg/devfile/adapters/docker/component/utils_test.go
```diff
@@ -591,6 +591,107 @@ func TestExecDevfile(t *testing.T) {
...
+					Component:   component,
+					Group: &versionsCommon.Group{
+						Kind: versionsCommon.TestCommandGroupType,
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name:   "Case 3: No devfile test commands should result in an err",
```
maysunfaisal, line 642: Case 2*, Case 3*

tests/integration/devfile/docker/cmd_docker_devfile_test_test.go
```diff
@@ -0,0 +1,189 @@
...
+			helper.CmdShouldPass("odo", "create", "nodejs", "--context", context, cmpName)
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile-with-testgroup.yaml"), filepath.Join(context, "devfile.yaml"))
+			helper.ReplaceString("devfile.yaml", "isDefault: true", "")
+			helper.CmdShouldPass("odo", "push", "--context", context)
+			output := helper.CmdShouldFail("odo", "test", "--context", context)
+			Expect(output).To(ContainSubstring("there should be exactly one default command for command group test, currently there is no default command"))
+		})
+
+		It("should show error if devfile has multiple default test command", func() {
```
maysunfaisal, line 115: duplicate test as above?

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -163,6 +163,38 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
...
+func (a Adapter) Test(testCmd string, show bool) (err error) {
+	// componentExists := utils.ComponentExists(a.Client, a.ComponentName)
+	// if !componentExists {
+	// 	return fmt.Errorf("component does not exist, a valid component is required to run 'odo test'")
+	// }
+	// pod, err := a.waitAndGetComponentPod(true)
+	// if err != nil {
+	// 	return errors.Wrapf(err, "unable to get pod for component %s", a.ComponentName)
+	// }
+
```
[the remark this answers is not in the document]
maysunfaisal: Remember to update the integration test since its relying on the err msg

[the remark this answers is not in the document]
yangcao77: The message`no Pod was found for the selector` in from `GetOnePodFromSelector` which is called by `GetPodUsingComponentName`. I prefer not update the existing code, since it was originally approved.

tests/integration/devfile/docker/cmd_docker_devfile_test_test.go
yangcao77: The above one should be with no default command. Need to rename the test. 
Thanks for catching this

pkg/devfile/adapters/kubernetes/component/adapter.go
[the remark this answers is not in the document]
yangcao77: I tried it does not wrap the err as expect. It because the error returned from  `GetPodUsingComponentName` is built using `fmt.Errorf`.

And I tried the following, it only shows `<some format string1>` in the terminal. It might be a general issue with our code & golang. 
 ```
errors.Wrapf(fmt.Errorf(<some format string1> ), <some format string2>)
```

Since the error message from `GetPodUsingComponentName` is clear enough, adding an additional error wrap is not meaningful; I will simply return err in this case.

File: tests/integration/devfile/docker/cmd_docker_devfile_test_test.go (hunk 37 above)
Comment (yangcao77) [reply tests/integration/devfile/docker/cmd_docker_devfile_test_test.go:115]: updated

File: pkg/devfile/adapters/docker/component/utils_test.go (hunk 36 above)
Comment (yangcao77) [reply pkg/devfile/adapters/docker/component/utils_test.go]: updated

File: pkg/devfile/adapters/docker/component/adapter_test.go (hunk 35 above)
Comment (yangcao77) [reply pkg/devfile/adapters/docker/component/adapter_test.go]: updated

File: pkg/devfile/adapters/kubernetes/component/adapter.go (hunk 38 above)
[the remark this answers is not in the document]
Comment (maysunfaisal) [reply pkg/devfile/adapters/kubernetes/component/adapter.go]: Ok, but it looks weird to me that the error is very generic. We should at least wrap it like`failed to run 'odo test': no Pod was found for the selector: component=xyz`

File: pkg/devfile/adapters/kubernetes/component/adapter.go [hunk 39]
```diff
@@ -186,6 +177,7 @@ func (a Adapter) Test(testCmd string, show bool) (err error) {
 
 	testCommand, err := common.ValidateAndGetTestDevfileCommands(a.Devfile.Data, testCmd)
 	if err != nil {
+		err.Error()
```
maysunfaisal, line 184: also not sure the point of this over here

yangcao77: oops. This was added by mistake, thought I was editing another branch. 
Removed.

pkg/devfile/adapters/kubernetes/component/adapter.go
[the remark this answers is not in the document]
yangcao77: IMO, after executing the command, the error message is shown with a red X, it's clear enough that `odo test` fail to run, so wrap wrap it like `failed to run 'odo test': no Pod was found for the selector: component=xyz` is meaningless. 

How about, `No valid Pod can be found to run 'odo test': no Pod was found for the selector: component=xyz`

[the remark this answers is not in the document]
maysunfaisal: sounds good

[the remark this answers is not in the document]
yangcao77: Or wrap as Mrinal suggest: `component is not running: <error message from GetPodUsingComponentName>`?
Which one do you think is more appropriate?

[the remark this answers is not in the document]
maysunfaisal: `no valid pod can be found to run 'odo test'` since like you said the func that you're using can return 0 pod or more than 1 pod. So in the case of the latter, the component may be running but odo doesn't know which pod to use. So we should stick with what you suggested earlier

.travis.yml
```diff
@@ -138,6 +138,7 @@ jobs:
         - travis_wait make test-cmd-devfile-watch
         - travis_wait make test-cmd-devfile-delete
         - travis_wait make test-cmd-devfile-registry
+        - travis_wait make test-cmd-devfile-test
```
mik-dass, line 141: please update the name of the test to `test`
/hold

pkg/devfile/adapters/common/command.go
```diff
@@ -59,7 +59,7 @@ func getCommandFromDevfile(data data.DevfileData, groupType common.DevfileComman
 
 	msg := fmt.Sprintf("the command group of kind \"%v\" is not found in the devfile", groupType)
 	// if run command is not found in devfile then it is an error
-	if groupType == common.RunCommandGroupType {
+	if groupType == common.RunCommandGroupType || groupType == common.TestCommandGroupType {
```
mik-dass, line 62: please update the comment

pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -374,6 +374,13 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	return
 }
 
+func (a Adapter) execTestCmd(testCmd versionsCommon.DevfileCommand, containers []types.Container, show bool) (err error) {
```
mik-dass, line 377: comments here

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -167,6 +167,29 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	return nil
 }
 
+// Test runs the devfile test command
+func (a Adapter) Test(testCmd string, show bool) (err error) {
+	pod, err := a.Client.GetPodUsingComponentName(a.ComponentName)
+	if err != nil {
+		return fmt.Errorf("no valid pod can be found to run 'odo test': %w", err)
```
mik-dass, line 174: We shouldn't say `no valid pod can be found to run` as we couldn't determine it due to the error. This should be more like `error occurred while getting the pod...`

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -413,6 +436,15 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	return
 }
 
+func (a Adapter) execTestCmd(testCmd versionsCommon.DevfileCommand, podName string, containers []corev1.Container, show bool) (err error) {
```
mik-dass, line 439: comments

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,97 @@
...
+	return to.RunTestCommand()
+}
+
+// NewCmdTest implements the odo test command
+func NewCmdTest(name, fullName string) *cobra.Command {
+	to := NewTestOptions()
+	testCmd := &cobra.Command{
+		Use:     name,
+		Short:   "Run the test command defined in devfile",
+		Long:    "Run the test command defined in devfile",
```
mik-dass, line 78: `defined in the devfile`

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,97 @@
...
+		Example: fmt.Sprintf(testExample, fullName),
+		Args:    cobra.MaximumNArgs(1),
```
mik-dass, line 80: Why are we accepting a argument? It doesn't seem to anything with it in the code.
```
[mrinaldas@localhost nodejs-ex]$ odo test garbage
I0714 10:58:53.346770   16885 cert_rotation.go:137] Starting client certificate rotation controller

Executing devfile test command for component nodejs
 ✓  Executing test command "npm test" [265ms]
[mrinaldas@localhost nodejs-ex]$ odo test blhaaddaskjha
I0714 10:58:56.330496   16908 cert_rotation.go:137] Starting client certificate rotation controller

Executing devfile test command for component nodejs
 ✓  Executing test command "npm test" [269ms]
```

pkg/odo/cli/component/test.go
```diff
@@ -0,0 +1,97 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(to, cmd, args)
+		},
+	}
+
+	// Add a defined annotation in order to appear in the help menu
+	testCmd.Annotations = map[string]string{"command": "main"}
+	testCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
+	testCmd.Flags().StringVar(&to.commandName, "test-command", "", "Devfile Test Command to execute")
+	testCmd.Flags().BoolVar(&to.show, "show-log", false, "If enabled, logs will be shown when running test command")
```
mik-dass, line 90: `running the test command `

pkg/devfile/adapters/common/command_test.go
```diff
@@ -1313,6 +1403,91 @@ func TestValidateAndGetPushDevfileCommands(t *testing.T) {
...
+			wantErr:       false,
+		},
+		{
+			name:          "Case: provided test Command",
+			testCommand:   "customtestcommand",
+			componentType: versionsCommon.ContainerComponentType,
+			wantErr:       false,
+		},
+		{
+			name:          "Case: invalid debug Command",
```
mik-dass, line 1454: invalid test command?

pkg/devfile/adapters/common/command.go
```diff
@@ -279,3 +285,24 @@ func ValidateAndGetDebugDevfileCommands(data data.DevfileData, devfileDebugCmd s
...
+	testCommand, testCmdErr := GetTestCommand(data, devfileTestCmd)
+	if testCmdErr == nil && !reflect.DeepEqual(emptyCommand, testCommand) {
+		isTestCommandValid = true
+		klog.V(4).Infof("Test command: %v", testCommand.Exec.Id)
+	}
+
+	if !isTestCommandValid {
+		var commandErrors string
+		if testCmdErr != nil {
+			commandErrors = testCmdErr.Error()
```
mik-dass, line 302: I think this should be `commandErrors += testCmdErr.Error()` like the others above

pkg/odo/cli/component/test.go
yangcao77: The argument is the specific command name we want to run if we don't want to run the default one. 
And if the command does not exist, we run the default command.

pkg/devfile/adapters/common/command.go
yangcao77: Actually for test command, it only has testCmdErr, so we can use `=`
we use `+=` in the above due to push executes initCmd, runCmd and buildCmd, so need to concatenate the errors.

.travis.yml
yangcao77: I don't quite understand this comment.

pkg/odo/cli/component/test.go
yangcao77: You are right. This should be 0 args

.travis.yml
mik-dass: I meant updating the name of the job like https://github.com/openshift/odo/pull/3400/files#diff-354f30a63fb0907d4ad57269548329e3R161 below

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: updated

pkg/devfile/adapters/kubernetes/component/adapter.go
yangcao77: updated

pkg/devfile/adapters/docker/component/utils.go
yangcao77: updated

pkg/devfile/adapters/common/command_test.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

pkg/odo/cli/component/test.go
yangcao77: updated

.travis.yml
yangcao77: updated

cdrage:
/hold

Honestly, we haven't had much discussion with regards to this command in the issue: https://github.com/openshift/odo/issues/3284

Perhaps we can make 100% sure we want this command before we merge?

maysunfaisal:
re-adding lgtm after rebase
/lgtm

yangcao77:
@cdrage Do you still have any questions or concerns that you want to discuss? 
Can I get an approve label for this PR?

PR #3405, opened 2020-06-23
Update odo delete for no component and devfile.yaml

### LINKED ISSUE #3104: Erasing a local devfile is not enough to reset odo
/kind bug

**Operating System:** macOS 10.15.4

**Output of `odo version`:**
odo v1.2.0 (e1232ad1a)

Server: <redacted>
Kubernetes: v1.11.0+d4cacc0

1. odo component create maven
2. Realize that I actually wanted another component type
3. odo component create java-spring-boot
4. Get an error: ✗  This directory already contains a devfile.yaml, please delete it and run the component creation command again
5. So follow the instructions and delete `devfile.yaml`
6. odo component create java-spring-boot:
```

Validation
 ✓  Checking devfile compatibility [14ms]
 ✗  Validating devfile component [114030ns]
 ✗  This workspace directory already contains a devfile component
```

## Actual behavior
See above.

## Expected behavior
Be able to create my component the second time around without issue. At least, provide a better error message that would allow a user to fix the issue.

maysunfaisal:

**What type of PR is this?**
/kind bug

- `odo delete`/`odo delete -a` does not err out when there is no component to delete. Just shows a message saying there is no component present
- `odo delete -a` now also prompts if user wants to delete the devfile.yaml

Fixes #3104 

See https://github.com/openshift/odo/pull/3405#issuecomment-651913789 for detailed case scenarios

### DIFF
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -203,8 +203,8 @@ func (a Adapter) Delete(labels map[string]string) error {
 	componentContainer := a.Client.GetContainersByComponent(componentName, containers)
 
 	if len(componentContainer) == 0 {
-		log.Infof("Component %s does not exist", componentName)
-		spinner.End(true)
+		spinner.End(false)
+		log.Warningf("Component %s does not exist", componentName)
 		return nil
 	}
 
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -200,7 +200,8 @@ func (a Adapter) Delete(labels map[string]string) error {
 	componentContainer := a.Client.GetContainersByComponent(componentName, containers)
 
 	if len(componentContainer) == 0 {
-		return errors.Errorf("the component %s doesn't exist", a.ComponentName)
+		log.Italicf("Component %s does not exist", componentName)
+		return nil
 	}
 
 	allVolumes, err := a.Client.GetVolumes()
@@ -223,6 +224,9 @@ func (a Adapter) Delete(labels map[string]string) error {
 			}
 		}
 	}
+
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
+	defer spinner.End(false)
 
 	// A unique list of volumes to delete; map key is volume name.
 	volumesToDelete := map[string]string{}
@@ -272,6 +276,9 @@ func (a Adapter) Delete(labels map[string]string) error {
 			return errors.Wrapf(err, "Unable to remove volume %s of component %s", name, componentName)
 		}
 	}
+
+	spinner.End(true)
+	log.Successf("Successfully deleted component")
 
 	return nil
 
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -146,9 +146,9 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 }
 
 // DoesComponentExist returns true if a component with the specified name exists, false otherwise
-func (a Adapter) DoesComponentExist(cmpName string) bool {
-	componentExists, _ := utils.ComponentExists(a.Client, a.Devfile.Data, cmpName)
-	return componentExists
+func (a Adapter) DoesComponentExist(cmpName string) (bool, error) {
+	componentExists, err := utils.ComponentExists(a.Client, a.Devfile.Data, cmpName)
+	return componentExists, err
 }
 
 // getFirstContainerWithSourceVolume returns the first container that set mountSources: true
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -175,6 +175,9 @@ func (a Adapter) Delete(labels map[string]string) error {
 		return errors.New("unable to delete component without a component label")
 	}
 
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
+	defer spinner.End(false)
+
 	containers, err := a.Client.GetContainerList()
 	if err != nil {
 		return errors.Wrap(err, "unable to retrieve container list for delete operation")
@@ -200,7 +203,8 @@ func (a Adapter) Delete(labels map[string]string) error {
 	componentContainer := a.Client.GetContainersByComponent(componentName, containers)
 
 	if len(componentContainer) == 0 {
-		log.Italicf("Component %s does not exist", componentName)
+		log.Infof("Component %s does not exist", componentName)
+		spinner.End(true)
 		return nil
 	}
 
@@ -224,9 +228,6 @@ func (a Adapter) Delete(labels map[string]string) error {
 			}
 		}
 	}
-
-	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
-	defer spinner.End(false)
 
 	// A unique list of volumes to delete; map key is volume name.
 	volumesToDelete := map[string]string{}
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -175,7 +175,7 @@ func (a Adapter) Delete(labels map[string]string) error {
 		return errors.New("unable to delete component without a component label")
 	}
 
-	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
+	spinner := log.Spinnerf("Deleting devfile component %s", componentName)
 	defer spinner.End(false)
 
 	containers, err := a.Client.GetContainerList()
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -458,16 +458,16 @@ func (a Adapter) Delete(labels map[string]string) error {
 	if kerrors.IsForbidden(err) {
 		klog.V(4).Infof("Resource for %s forbidden", a.ComponentName)
 		// log the error if it failed to determine if the component exists due to insufficient RBACs
-		log.Infof("%s: %v", errorMsg, err)
-		spinner.End(true)
+		spinner.End(false)
+		log.Warningf("%s: %v", errorMsg, err)
 		return nil
 	} else if err != nil {
 		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
 	}
 
 	if !componentExists {
-		log.Infof(errorMsg)
-		spinner.End(true)
+		spinner.End(false)
+		log.Warningf(errorMsg)
 		return nil
 	}
 
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift/odo/pkg/machineoutput"
 	odoutil "github.com/openshift/odo/pkg/odo/util"
 	"github.com/openshift/odo/pkg/sync"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // New instantiantes a component adapter
@@ -451,17 +452,21 @@ func (a Adapter) Delete(labels map[string]string) error {
 	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
 	defer spinner.End(false)
 
+	errorMsg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+
 	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
-	if !componentExists || err != nil {
-		msg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+	if kerrors.IsForbidden(err) {
+		klog.V(4).Infof("Resource for %s forbidden", a.ComponentName)
+		// log the error if it failed to determine if the component exists due to insufficient RBACs
+		log.Infof("%s: %v", errorMsg, err)
+		spinner.End(true)
+		return nil
+	} else if err != nil {
+		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
+	}
 
-		// log the error if it failed to determine if the component exists
-		if err != nil {
-			msg = fmt.Sprintf("%s: %v", msg, err)
-		}
-
-		log.Infof(msg)
-
+	if !componentExists {
+		log.Infof(errorMsg)
 		spinner.End(true)
 		return nil
 	}
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -60,7 +60,10 @@ type Adapter struct {
 // Push updates the component if a matching component exists or creates one if it doesn't exist
 // Once the component has started, it will sync the source code to it.
 func (a Adapter) Push(parameters common.PushParameters) (err error) {
-	componentExists := utils.ComponentExists(a.Client, a.ComponentName)
+	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
+	}
 
 	a.devfileInitCmd = parameters.DevfileInitCmd
 	a.devfileBuildCmd = parameters.DevfileBuildCmd
@@ -165,7 +168,7 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 
 // DoesComponentExist returns true if a component with the specified name exists, false otherwise
 func (a Adapter) DoesComponentExist(cmpName string) (bool, error) {
-	return utils.ComponentExists(a.Client, cmpName), nil
+	return utils.ComponentExists(a.Client, cmpName)
 }
 
 func (a Adapter) createOrUpdateComponent(componentExists bool) (err error) {
@@ -445,11 +448,26 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
 
 // Delete deletes the component
 func (a Adapter) Delete(labels map[string]string) error {
-	if !utils.ComponentExists(a.Client, a.ComponentName) {
-		return errors.Errorf("the component %s doesn't exist on the cluster", a.ComponentName)
+	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
+	}
+	if !componentExists {
+		log.Italicf("Component %s does not exist", a.ComponentName)
+		return nil
 	}
 
-	return a.Client.DeleteDeployment(labels)
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
+	defer spinner.End(false)
+
+	err = a.Client.DeleteDeployment(labels)
+	if err != nil {
+		return err
+	}
+
+	spinner.End(true)
+	log.Successf("Successfully deleted component")
+	return nil
 }
 
 // Log returns log from component
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -164,8 +164,8 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 }
 
 // DoesComponentExist returns true if a component with the specified name exists, false otherwise
-func (a Adapter) DoesComponentExist(cmpName string) bool {
-	return utils.ComponentExists(a.Client, cmpName)
+func (a Adapter) DoesComponentExist(cmpName string) (bool, error) {
+	return utils.ComponentExists(a.Client, cmpName), nil
 }
 
 func (a Adapter) createOrUpdateComponent(componentExists bool) (err error) {
@@ -252,7 +252,7 @@ func (a Adapter) createOrUpdateComponent(componentExists bool) (err error) {
 	klog.V(4).Infof("Creating deployment %v", deploymentSpec.Template.GetName())
 	klog.V(4).Infof("The component name is %v", componentName)
 
-	if utils.ComponentExists(a.Client, componentName) {
+	if componentExists {
 		// If the component already exists, get the resource version of the deploy before updating
 		klog.V(4).Info("The component already exists, attempting to update it")
 		deployment, err := a.Client.UpdateDeployment(*deploymentSpec)
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -449,17 +449,15 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
 
 // Delete deletes the component
 func (a Adapter) Delete(labels map[string]string) error {
-	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
+	spinner := log.Spinnerf("Deleting devfile component %s", a.ComponentName)
 	defer spinner.End(false)
-
-	errorMsg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
 
 	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
 	if kerrors.IsForbidden(err) {
 		klog.V(4).Infof("Resource for %s forbidden", a.ComponentName)
 		// log the error if it failed to determine if the component exists due to insufficient RBACs
 		spinner.End(false)
-		log.Warningf("%s: %v", errorMsg, err)
+		log.Warningf("%v", err)
 		return nil
 	} else if err != nil {
 		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
@@ -467,7 +465,7 @@ func (a Adapter) Delete(labels map[string]string) error {
 
 	if !componentExists {
 		spinner.End(false)
-		log.Warningf(errorMsg)
+		log.Warningf("Component %s does not exist", a.ComponentName)
 		return nil
 	}
 
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -448,17 +448,18 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
 
 // Delete deletes the component
 func (a Adapter) Delete(labels map[string]string) error {
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
+	defer spinner.End(false)
+
 	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
 	if err != nil {
 		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
 	}
 	if !componentExists {
-		log.Italicf("Component %s does not exist", a.ComponentName)
+		log.Infof("Component %s does not exist", a.ComponentName)
+		spinner.End(true)
 		return nil
 	}
-
-	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
-	defer spinner.End(false)
 
 	err = a.Client.DeleteDeployment(labels)
 	if err != nil {
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -452,11 +452,16 @@ func (a Adapter) Delete(labels map[string]string) error {
 	defer spinner.End(false)
 
 	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
-	if err != nil {
-		return errors.Wrapf(err, "unable to determine if component %s exists", a.ComponentName)
-	}
-	if !componentExists {
-		log.Infof("Component %s does not exist", a.ComponentName)
+	if !componentExists || err != nil {
+		msg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+
+		// log the error if it failed to determine if the component exists
+		if err != nil {
+			msg = fmt.Sprintf("%s: %v", msg, err)
+		}
+
+		log.Infof(msg)
+
 		spinner.End(true)
 		return nil
 	}
--- a/pkg/devfile/adapters/kubernetes/component/adapter_test.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter_test.go
@@ -302,7 +302,7 @@ func TestDoesComponentExist(t *testing.T) {
 			}
 
 			// Verify that a component with the specified name exists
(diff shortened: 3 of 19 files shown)


maysunfaisal:
@mik-dass @johnmcollier could you pls take a look and help me review this? thx!

kadel:
`odo delete` still fails when the `namespace` where components used to be no longer exists.

```
▶ odo project create test1
 ✓  Project 'test1' is ready for use
 ✓  New project created and now using project: test1

▶ odo create maven

Validation
 ✓  Checking devfile existence [49487ns]
 ✓  Checking devfile compatibility [157525ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [163298ns]
 ✓  Validating devfile component [518605ns]

tmp/odo-examples/demo-backend  master ✗                                                                                                                                    76d ⚑ ◒
▶ odo push

Validation
 ✓  Validating the devfile [83687ns]

Creating Kubernetes resources for component maven
 ✓  Waiting for component to start [1m]

Applying URL changes

Syncing to component maven
 ✓  Checking files for pushing [5ms]
 ✓  Syncing files to the component [6s]

Executing devfile commands for component maven
 ✓  Executing devbuild command "mvn package" [41s]
 ✓  Executing devrun command "java -jar target/*.jar" [3s]

Pushing devfile component maven

▶ odo project delete test1
? Are you sure you want to delete project test1 Yes
 ✓  Deleted project : test1
 ⚠  Warning! Projects are deleted from the cluster asynchronously. Odo does its best to delete the project. Due to multi-tenant clusters, the project may still exist on a different node.

▶ odo create maven

▶ odo delete
 ✗  You don't have permission to create or set namespace 'test1' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)

```

maysunfaisal:
@kadel isn't that expected if you delete the namespace/project? odo push/odo url fails in that case too since there is no active namespace until you switch to a namespace that is available:

before deleting the namespace/project:
```
$ oc projects
You have access to the following projects and can switch between them with 'oc project <projectname>':

    default
    kube-node-lease
    kube-public
    kube-system
  * myodo
    openshift

```

delete the namespace/project
```
$ odo project delete myodo
? Are you sure you want to delete project myodo Yes
 ✓  Deleted project : myodo
 ⚠  Warning! Projects are deleted from the cluster asynchronously. Odo does its best to delete the project. Due to multi-tenant clusters, the project may still exist on a different node.
```

after deleting the namespace/project, there is no active namespace/project:
```
$ oc projects
You have access to the following projects and can switch between them with 'oc project <projectname>':

default
kab
kube-node-lease
kube-public
kube-system
openshift
```

odo push or odo url fails:
```
$ odo url create testspr1 --host apps.xxxxx
 ✗  You don't have permission to create or set namespace 'myodo' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)



$ odo push

Validation
 ✓  Validating the devfile [86787ns]

Creating Kubernetes resources for component testspr2
 ✗  Failed to start component with name testspr2. Error: Failed to create the component: unable to create or update component: unable to create Deployment testspr2: namespaces "myodo" not found
```

odo delete fails as expected like the other commands
```
$ odo delete
 ✗  You don't have permission to create or set namespace 'myodo' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)
```

kadel:
> **TL;DR** - Isnt that expected? if you delete the namespace/project, then you dont have an active namespace/project and you need to switch to one before you run any odo commands for kube
> 
> @kadel isn't that expected if you delete the namespace/project? odo push/odo url fails in that case too since there is no active namespace until you switch to a namespace that is available:


`odo push` and `odo url` doesn't use active namespace as set in KUBECONFIG (what you can see in `oc project`).
They should use a namespace that is set in `.odo/env/env.yaml`.

I view this as a variation of a situation when you try to run `odo delete` before pushing. `odo delete` shouldn't exit with error if stuff was already deleted. 

I'm not 100% set on this, but this is what I would probably expect as a user.
If you think that erroring out in a situation when someone tries to run `odo delete` on a component where the whole namespace was already deleted is what you would expect we can keep this.

(BTW: I had copy/paste error in my previous post, the scenario was running `odo delete` right after `odo project delete` (no `odo create` in between)

maysunfaisal:
@kadel 

to avoid any confusion, yes we can agree that odo looks at env.yaml for the project and not the active project in kubeconfig. I was just using the default behaviour when explaining things.

here is the odo delete when no component exists in the project specified in env.yaml:
```
$ odo delete
? Are you sure you want to delete the devfile component: testspr4? Yes
Component testspr4 does not exist
```

here is my env.yaml:
```
$ cat .odo/env/env.yaml 
ComponentSettings:
  Name: testspr4
  Namespace: faisal
  Url:
  - Name: testspr4
    Port: 8080
    Host: xxxxx
    Kind: route
```

if i delete the project specified in the env.yaml and try odo commands:
```
$ odo create java-spring-boot testspr4 --project faisal

Validation
 ✓  Checking devfile existence [46322ns]
 ✓  Checking devfile compatibility [91399ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [119227ns]
 ✓  Validating devfile component [281644ns]

$ odo url create testspr4 --host apps.maysun2.os.fyre.ibm.com
 ✗  You don't have permission to create or set namespace 'faisal' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)


$ odo push

Validation
 ✓  Validating the devfile [87033ns]

Creating Kubernetes resources for component testspr4
 ✗  Failed to start component with name testspr4. Error: Failed to create the component: unable to create or update component: unable to create Deployment testspr4: namespaces "faisal" not found


$ odo delete
 ✗  You don't have permission to create or set namespace 'faisal' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)
```

so, odo url, odo push and odo delete does fail with the same errror when the project is deleted. But as discussed with @kadel , this does not allow `odo delete -a` to delete `.odo` and `devfile.yaml`

pkg/devfile/adapters/kubernetes/component/adapter_test.go
```diff
@@ -266,7 +277,7 @@ func TestDoesComponentExist(t *testing.T) {
 			}
 
 			// Verify that a component with the specified name exists
-			componentExists := componentAdapter.DoesComponentExist(tt.getComponentName)
+			componentExists, _ := componentAdapter.DoesComponentExist(tt.getComponentName)
```
mik-dass, line 280: we should check the error here.

pkg/odo/cli/component/devfile.go
```diff
@@ -169,6 +169,12 @@ func (do *DeleteOptions) DevfileComponentDelete() error {
 		return err
 	}
 
+	componentExists, _ := devfileHandler.DoesComponentExist(componentName)
+	if !componentExists {
+		log.Italicf("Component %s does not exist", componentName)
+		return nil
+	}
```
mik-dass, line 176: We are already doing the existence check in the kubernetes adapter https://github.com/openshift/odo/blob/edc7c336973c3978d3e8cadfee99e1c3dff2b91c/pkg/devfile/adapters/kubernetes/component/adapter.go#L445. Also we shouldn't we ignoring this error and should return it.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -162,8 +162,8 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 }
 
 // DoesComponentExist returns true if a component with the specified name exists, false otherwise
-func (a Adapter) DoesComponentExist(cmpName string) bool {
-	return utils.ComponentExists(a.Client, cmpName)
+func (a Adapter) DoesComponentExist(cmpName string) (bool, error) {
+	return utils.ComponentExists(a.Client, cmpName), nil
```
mik-dass, line 166: I think `utils.ComponentExists()` should throw the underlying error too.

maysunfaisal: kube's `utils.ComponentExists()` client API is a special case. It throws an error if there is no deployment(component). So its going to be difficult to distinguish between an actual error and when there is no deployment. So thats the reason I haven't changed kube's `utils.ComponentExists()`

pkg/devfile/adapters/kubernetes/component/adapter_test.go
maysunfaisal: explicitly returning nil like I mentioned here https://github.com/openshift/odo/pull/3405/files#r446364162

pkg/devfile/adapters/kubernetes/component/adapter.go
johnmcollier: But why return an error at all here for `func (a Adapter) DoesComponentExist` then if we're always returning nil?

pkg/odo/cli/component/devfile.go
maysunfaisal: We cant use the err here.
1. we're explicitly passing nil from kube, see comment https://github.com/openshift/odo/pull/3405/files#r446364162
2. docker's `utils.ComponentExists()` is a special case since docker does not have a deployment and has multiple individual containers - see code https://github.com/openshift/odo/blob/master/pkg/devfile/adapters/docker/utils/utils.go#L43. So if for example, len(containers) != len(devfile components), we return an error acc to the code but odo delete should be able to delete those remaining/hanging containers and this is the reason we ignore err here

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: @johnmcollier  because its an interface and docker is returning special err codes since docker has multiple containers and `utils.ComponentExists()` is different for docker. So `odo delete` should be able to handle the hanging containers see explanation here https://github.com/openshift/odo/pull/3405#discussion_r446366050

maysunfaisal: Maybe I can dig around more again for the kube client api and see if we can make it uniform. Open to ideas to handle consistency between docker client and kube client for the interface.

johnmcollier: Hmm, yeah. Good point

maysunfaisal: I think I can make Kube and Docker `utils.ComponentExists()` consistent

replacing 

```
deployment, err := c.KubeClient.AppsV1().Deployments(c.Namespace).Get(name, metav1.GetOptions{})
```

with 

```
label := "component=" + name
deploymentList, err := c.KubeClient.AppsV1().Deployments(c.Namespace).List(metav1.ListOptions{
	LabelSelector: label,
})
```

mik-dass: > kube's utils.ComponentExists() client API is a special case. It throws an error if there is no deployment(component). So its going to be difficult to distinguish between an actual error and when there is no deployment and thus it returns only a bool.

We can use the kerrors package to distinguish between the errors https://github.com/openshift/odo/blob/9a470629475bf6339dec637ee5ea8e326188a7fd/pkg/occlient/occlient.go#L3334

pkg/odo/cli/component/devfile.go
mik-dass: > we're explicitly passing nil from kube, see comment #3405 (comment)

https://github.com/openshift/odo/pull/3405#discussion_r446839977 will solve this problem

> So if for example, len(containers) != len(devfile components), we return an error acc to the code but odo delete should be able to delete those remaining/hanging containers and this is the reason we ignore err here

Then let's move this check from move here into the delete handler. That way it won't affect the kube part and also it will be consistent with the kubernetes delete handler which has the exist check in it https://github.com/openshift/odo/blob/edc7c336973c3978d3e8cadfee99e1c3dff2b91c/pkg/devfile/adapters/kubernetes/component/adapter.go#L445

pkg/devfile/adapters/kubernetes/component/adapter_test.go
mik-dass: https://github.com/openshift/odo/pull/3405#discussion_r446839977 will solve this problem

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: Thanks, I will use that

pkg/odo/cli/component/devfile.go
maysunfaisal: Moved over the logic to the adapter check. This also resulted in moving over the spinners to the individual adapters to keep things clean from UX pov.

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -219,6 +220,9 @@ func (a Adapter) Delete(labels map[string]string) error {
 		}
 	}
 
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
+	defer spinner.End(false)
+
```
mik-dass, line 225: This seems to be repeated again for the kubernetes adapter. Can we move it before the call?

pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -195,7 +195,8 @@ func (a Adapter) Delete(labels map[string]string) error {
 	componentContainer := a.Client.GetContainersByComponent(componentName, containers)
 
 	if len(componentContainer) == 0 {
-		return errors.Errorf("the component %s doesn't exist", a.ComponentName)
+		log.Italicf("Component %s does not exist", componentName)
```
mik-dass, line 198: Any reason for turning the error message into Italic? I guess we normally start error messages with a small letter.

pkg/odo/cli/component/devfile.go
```diff
@@ -169,16 +168,7 @@ func (do *DeleteOptions) DevfileComponentDelete() error {
...
-	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", componentName))
-	defer spinner.End(false)
-
-	err = devfileHandler.Delete(labels)
-	if err != nil {
-		return err
-	}
-	spinner.End(true)
-	log.Successf("Successfully deleted component")
-	return nil
```
mik-dass, line 181: These actions seems to be repeated in the delete call for both kube and docker. Any reason for removing it from here?

pkg/odo/genericclioptions/context.go
```diff
@@ -99,7 +99,7 @@ func kClient(command *cobra.Command) *kclient.Client {
 // checkProjectCreateOrDeleteOnlyOnInvalidNamespace errors out if user is trying to create or delete something other than project
 // errFormatforCommand must contain one %s
 func checkProjectCreateOrDeleteOnlyOnInvalidNamespace(command *cobra.Command, errFormatForCommand string) {
-	if command.HasParent() && command.Parent().Name() != "project" && (command.Name() == "create" || command.Name() == "delete") {
+	if command.HasParent() && command.Parent().Name() != "project" && (command.Name() == "create" || (command.Name() == "delete" && !command.Flags().Changed("all"))) {
```
mik-dass, line 102: can you please add comment regarding the change

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -17,9 +17,9 @@ import (
 )
 
 // ComponentExists checks whether a deployment by the given name exists
-func ComponentExists(client kclient.Client, name string) bool {
-	_, err := client.GetDeploymentByName(name)
-	return err == nil
+func ComponentExists(client kclient.Client, name string) (bool, error) {
+	deployment, err := client.GetDeploymentByName(name)
+	return len(deployment.GetName()) != 0, err
```
mik-dass, line 22: if deployment is not found, won't it be nil?

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: We do not want to error out on `odo delete` if the component is not there. We just want to print the information to the user. Any logger besides italic you think is better suited?

pkg/odo/cli/component/devfile.go
maysunfaisal: I think its a bad UX if we're not deleting anything(when there is no component on `odo delete`) but showing `Deleting devfile component testsp11`/`Successfully deleted component` spinners.

Thats why I had the component exist check in this file initially, rather than the individual adapters. But since we decided to keep the component check on delete in the individual adapters, I think we should only show spinners when we're actually deleting the component.

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: See https://github.com/openshift/odo/pull/3405#discussion_r447820362

pkg/odo/genericclioptions/context.go
maysunfaisal: coincidentally added the comment on my new commit before reading this :D

mik-dass: Cool :+1:

pkg/devfile/adapters/kubernetes/utils/utils.go
maysunfaisal: This is a weird behaviour from kube client api. It ALWAYS returns a deployment object even if there is no deployment. For example when the deployment is not found, kube api returns 
```
&Deployment{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:DeploymentSpec{Replicas:nil,Selector:nil,Template:{{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] []}},Strategy:DeploymentStrategy{Type:,RollingUpdate:nil,},MinReadySeconds:0,RevisionHistoryLimit:nil,Paused:false,ProgressDeadlineSeconds:nil,},Status:DeploymentStatus{ObservedGeneration:0,Replicas:0,UpdatedReplicas:0,AvailableReplicas:0,UnavailableReplicas:0,Conditions:[]DeploymentCondition{},ReadyReplicas:0,CollisionCount:nil,},} err: <nil>
```

I cant check for `nil` or `reflect.DeepEqual` an empty deployment struct. I found that when the deployment is not there, the name length is always 0. 

Any other suggestions welcome! But I think alternative would be to replace `GetDeploymentByName` logic with the `List` API like I suggested earlier here https://github.com/openshift/odo/pull/3405#discussion_r446437850 and try getting the deployment with labels.

maysunfaisal:
### Delete component successfully
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete
? Are you sure you want to delete the devfile component: testsp11? Yes
 ✓  Deleting devfile component testsp11 [118ms]

```

### Try deleting when component is already deleted (does not err out)
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete
? Are you sure you want to delete the devfile component: testsp11? Yes
Component testsp11 does not exist
```

### -a flag now prompts devfile delete
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete -a
? Are you sure you want to delete the devfile component: testsp11? Yes
Component testsp11 does not exist
? Are you sure you want to delete env folder? Yes
 ✓  Successfully deleted env file
? Are you sure you want to delete devfile.yaml? Yes
 ✓  Successfully deleted devfile.yaml file

```


### Try odo delete when the namespace is not there (errors out)
```
Maysuns-MacBook-Pro:springboot-ex maysun$ cat .odo/env/env.yaml 
ComponentSettings:
  Name: testsp11
  Namespace: garbage


Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete
 ✗  You don't have permission to create or set namespace 'garbage' or the namespace doesn't exist. Please create or set a different namespace
	%!(EXTRA string=odo)


```


### When the namespace is not there, odo delete -a should let me delete the local config
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete -a
? Are you sure you want to delete the devfile component: testsp11? Yes
Component testsp11 does not exist
? Are you sure you want to delete env folder? Yes
 ✓  Successfully deleted env file
? Are you sure you want to delete devfile.yaml? Yes
 ✓  Successfully deleted devfile.yaml file
```


### When the namespace is not there and if the project flag is used (it should err out)
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete --project garbage
 ✗  namespaces "garbage" not found

```


### When the namespace is not there and if the project and all flags are used, it should let me delete local config
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete --project garbage -a
? Are you sure you want to delete the devfile component: testsp11? Yes
Component testsp11 does not exist
? Are you sure you want to delete env folder? Yes
 ✓  Successfully deleted env file
? Are you sure you want to delete devfile.yaml? Yes
 ✓  Successfully deleted devfile.yaml file
```

pkg/odo/cli/component/devfile.go
maysunfaisal: check https://github.com/openshift/odo/pull/3405#issuecomment-651913789 for the `odo delete` output from this PR

pkg/devfile/adapters/docker/component/adapter.go
mik-dass: I think Infof() is better suited for this task

pkg/odo/cli/component/devfile.go
mik-dass: But the problem is now the exist call is not covered by the spiners. Thus on a slow connection/cluster, that call might look unresponsive. Too many spiners is also a bad design. I think we can consider the exist call as a part of the delete action as it was called due to the delete process.

pkg/devfile/adapters/kubernetes/utils/utils.go
mik-dass: Maybe let's move the error checking logic from GetDeploymentByName() to here and leave that unchanged. I feels that GetDeploymentByName throws the error so other functions can modify the error checking according to their needs.
```
	if kerrors.IsNotFound(err) {
		klog.V(4).Infof("Deployment %s not found", name)
		return false, nil
	} else if err!=nil {
		return true, err
	}
	return deployment!=nil, err
```

pkg/odo/cli/component/devfile.go
maysunfaisal: updated, added the spinners to include the component check in their individual adapters

pkg/devfile/adapters/docker/component/adapter.go
maysunfaisal: updated

pkg/devfile/adapters/kubernetes/utils/utils.go
maysunfaisal: true makes much more sense, updated

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -444,9 +447,30 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
...
+	defer spinner.End(false)
+
+	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
+	if !componentExists || err != nil {
+		msg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+
+		// log the error if it failed to determine if the component exists
+		if err != nil {
+			msg = fmt.Sprintf("%s: %v", msg, err)
+		}
```
mik-dass, line 460: We shouldn't say `Component %s does not exist` in case of an error asit might cause confusion.

adisky: agreed here, also it looks complicated. Its better to simply return `err` here

maysunfaisal: @mik-dass ok, i will try to just log the err then w/o saying component does not exist

@adisky Cant return the err here. This was asked by Tomas, we shouldn't return an err when no component exist:
1. if the kube user is `admin` and the namespace is deleted; the kube api tells there is no component present.
2. if the kube user is `developer` and the namespace is deleted; the kube api complains about not having sufficient RBAC policies to get deployments, and returns an error

So this was the reason why the err msg is being logged instead of being returned

adisky: @maysunfaisal can we check the error code here? and wrap error accordingly?

maysunfaisal: @adisky hmm let me see if i can clean it up further with `kerror`

maysunfaisal: Updated

So I have now cleaned it up a bit more. I am specifically handling Forbidden errors on delete asides the Not Found errors as well. I have also put unit tests and made sure integration tests on CI for OpenShift and Minikube works with these specific errors!

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -444,9 +448,34 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
...
+	defer spinner.End(false)
+
+	errorMsg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+
+	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
+	if kerrors.IsForbidden(err) {
+		klog.V(4).Infof("Resource for %s forbidden", a.ComponentName)
+		// log the error if it failed to determine if the component exists due to insufficient RBACs
+		log.Infof("%s: %v", errorMsg, err)
+		spinner.End(true)
```
mik-dass, line 461: since this is an error, shouldn't this be false?

maysunfaisal: I had it as `true` because `Forbidden` is a special case when no namespace exist which is similar to no component exist from delete's perspective and we dont want to show a cross check mark on `odo delete`

maysunfaisal: @kadel do you have any recommendation on whether we should show a ✗ or ✓; when 
1. component doesnt exist
2. namespace doesn't exist

In both instances, we're logging information on the screen.

mik-dass: I think namespace error can be considered a genuine error as this will inform the user that he is not logged in the cluster.

maysunfaisal: I've updated the spinners and the `log.Infof` to `log.Warningf` since I chatted with Tomas and we wanted a spinner with warning symbol. But looks like odo spinner either does a ✗ or ✓, not ⚠. And looks like the spinner.WarningStatus is broken:
```
✗  Deleting devfile component testm [2ms] [this is a warning spinner example]
```

it doesn't print `[this is a warning spinner example]` in yellow or have a ⚠ suffix :( we need to fix this on master

So I went with :
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete --project ga -af 
 ✗  Deleting devfile component testm [1ms]
 ⚠  Component testm does not exist: deployments.apps "testm" is forbidden: User "developer" cannot get resource "deployments" in API group "apps" in the namespace "ga"
 ✓  Successfully deleted env file
 ✓  Successfully deleted devfile.yaml file

Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete -af
 ✗  Deleting devfile component testm [3ms]
 ⚠  Component testm does not exist
 ✓  Successfully deleted env file
 ✓  Successfully deleted devfile.yaml file
```

@kadel @mik-dass lmk, if this looks alright!

kadel: > So the current PR does this (end spinner as false and print a warning msg):
> 
> ```
> Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete --project ga -af 
>  ✗  Deleting devfile component testm [1ms]
>  ⚠  Component testm does not exist: deployments.apps "testm" is forbidden: User "developer" cannot get resource "deployments" in API group "apps" in the namespace "ga"
>  ✓  Successfully deleted env file
>  ✓  Successfully deleted devfile.yaml file
> 
> Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete -af
>  ✗  Deleting devfile component testm [3ms]
>  ⚠  Component testm does not exist
>  ✓  Successfully deleted env file
>  ✓  Successfully deleted devfile.yaml file
> ```

this looks good

/approce

kadel: /approve

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -444,9 +448,34 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
 
 // Delete deletes the component
 func (a Adapter) Delete(labels map[string]string) error {
-	if !utils.ComponentExists(a.Client, a.ComponentName) {
-		return errors.Errorf("the component %s doesn't exist on the cluster", a.ComponentName)
+	spinner := log.Spinner(fmt.Sprintf("Deleting devfile component %s", a.ComponentName))
```
mik-dass, line 451: We can use Spinnerf() directly.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -444,9 +448,34 @@ func getFirstContainerWithSourceVolume(containers []corev1.Container) (string, s
...
+	defer spinner.End(false)
+
+	errorMsg := fmt.Sprintf("Component %s does not exist", a.ComponentName)
+
+	componentExists, err := utils.ComponentExists(a.Client, a.ComponentName)
+	if kerrors.IsForbidden(err) {
+		klog.V(4).Infof("Resource for %s forbidden", a.ComponentName)
+		// log the error if it failed to determine if the component exists due to insufficient RBACs
+		spinner.End(false)
+		log.Warningf("%s: %v", errorMsg, err)
```
mik-dass, line 461: Since it's a forbidden error, I think we can skip the `Component %s does not exist` addition to error message.

pkg/devfile/adapters/kubernetes/component/adapter.go
mik-dass: > Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete --project ga -af 
 ✗  Deleting devfile component testm [1ms]
 ⚠  Component testm does not exist: deployments.apps "testm" is forbidden: User "developer" cannot get resource "deployments" in API group "apps" in the namespace "ga"
 ✓  Successfully deleted env file
 ✓  Successfully deleted devfile.yaml file

Maysuns-MacBook-Pro:springboot-ex maysun$ odo delete -af
 ✗  Deleting devfile component testm [3ms]
 ⚠  Component testm does not exist
 ✓  Successfully deleted env file
 ✓  Successfully deleted devfile.yaml file

This looks good :+1:

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: good catch, updated

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: updated

maysunfaisal:
@mik-dass PR updated for feedback, thx

pkg/devfile/adapters/kubernetes/component/adapter.go
maysunfaisal: Thx, will keep it as is 👍

Issue #2662, opened 2020-03-02
odo push  trigger redeploy even if there are no changes in devfile


kadel:
/kind bug
/area devfile
/priority medium

If there are no changes to Devfile.yaml that would resolve in changes in Deploment the Deployment should not be triggered.

Current behaviour
```
export ODO_EXPERIMENTAL=true
# running push for the first time
▶ odo push ; kubectl get pods
 •  Push devfile component odo-examples  ...
 ✓  Waiting for component to start [9s]
 ✓  Push devfile component odo-examples [10s]
NAME                           READY   STATUS    RESTARTS   AGE
odo-examples-8cb546c7b-sbmxd   1/1     Running   0          11s


# just triggering push again without any changes

▶ odo push ; kubectl get pods
 •  Push devfile component odo-examples  ...
 ✓  Waiting for component to start [174ms]
 ✓  Push devfile component odo-examples [1s]
NAME                           READY   STATUS        RESTARTS   AGE
odo-examples-8cb546c7b-sbmxd   1/1     Terminating   0          71s
```

rajivnathan:
@cdrage I was wondering why I couldn't reproduce it but others could. We figured out that it has to do with ordering of things in the deployment spec when there is more than a single item in an array. In my case I had a devfile with only a single volume but others were reproducing the problem with devfiles that had two volumes.

For example, I cannot reproduce this problem with this devfile:
```
apiVersion: 1.0.0
metadata:
  generateName: openLiberty
attributes:
  odo.autorestart: "false"
projects:
- name: openLiberty
  source:
    type: git
    location: https://github.com/rajivnathan/openLiberty.git
components:
- type: chePlugin
  id: redhat/java/latest
  memoryLimit: 1512Mi
- alias: devruntime
  mountSources: true
  type: dockerimage
  image: jeevandroid/open-liberty-dev:latest
  command: ['tail']
  args: ['-f', '/dev/null']
  memoryLimit: 768Mi
  env:
  - name: test15
    value: test4
  volumes:
  - name: m2
    containerPath: /home/user/.m2
  endpoints:
  - name: 9090/tcp
    port: 9090
commands:
- actions:
  - command: /artifacts/bin/build.sh
    component: devruntime
    type: exec
    workdir: /projects/openLiberty
  name: DevBuild
- actions:
  - command: /artifacts/bin/run.sh
    component: devruntime
    type: exec
    workdir: /projects/openLiberty
  name: DevRun
- actions:
[...]

PR #3409, opened 2020-06-23
Fix bug that triggers redeploy even if there are no changes in the file

cdrage:
**What type of PR is this?**

/kind bug

If there are no changes to your code, we shouldn't have to redeploy.
There was an issue with maps being passed back in the volumes functionss
which was causing different orders of volumes.

When the volumes were passed back, they were causing unexpected redeploys
when deploying to Kubernetes / OpenShift.

Fixes https://github.com/openshift/odo/issues/2662

```sh
odo create java-spring-boot --downloadSource
odo push
odo push # try again and make sure that it does not redeploy with no
changes
```

### DIFF
--- a/pkg/devfile/adapters/docker/component/utils.go
+++ b/pkg/devfile/adapters/docker/component/utils.go
@@ -49,14 +49,24 @@ func (a Adapter) createComponent() (err error) {
 	// Loop over each component and start a container for it
 	for _, comp := range supportedComponents {
 		var dockerVolumeMounts []mount.Mount
-		for _, vol := range a.componentAliasToVolumes[comp.Container.Name] {
 
-			volMount := mount.Mount{
-				Type:   mount.TypeVolume,
-				Source: a.volumeNameToDockerVolName[vol.Name],
-				Target: vol.ContainerPath,
+		// Let's find the container / component alias!
+		for _, vol := range a.componentAliasToVolumes {
+
+			// If we find the container, we add all the volumes we'll be mounting.
+			if vol.ComponentAlias == comp.Container.Name {
+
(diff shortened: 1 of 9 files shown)


pkg/devfile/adapters/docker/component/utils.go
```diff
@@ -49,14 +49,24 @@ func (a Adapter) createComponent() (err error) {
...
 	for _, comp := range supportedComponents {
 		var dockerVolumeMounts []mount.Mount
-		for _, vol := range a.componentAliasToVolumes[comp.Container.Name] {
 
-			volMount := mount.Mount{
-				Type:   mount.TypeVolume,
-				Source: a.volumeNameToDockerVolName[vol.Name],
-				Target: vol.ContainerPath,
+		// Let's find the container / component alias!
+		for _, vol := range a.componentAliasToVolumes {
```
girishramnani, line 54: Seems like this and https://github.com/openshift/odo/pull/3409/files#diff-ec2e16874aae4e6d1f0ff0deefac8e04R106 are duplicates, can extract this into function and DRY the code?

girishramnani:
So the gist of this PR is that a Map of volumes is changed into a list of struct so that the random ordering behaviour of maps can be avoided

girishramnani:
Agreed on the philosophy so will approve it now but do Address the comment.
/approve

prietyc123:
Everything looks fine to me but please address @girishramnani comments.

girishramnani:
This PR has diverged alot, should we open a new one later?

girishramnani:
I dont think we would be to able to rebase this now - being in the state that it is. Probably open a fresh one @cdrage

Issue #2936, opened 2020-04-20
Container initiailization support


elsony:
/kind user-story

## User Story
As a stack creator, I want to initialize my the container that I defined in the devfile 2.0.  
Devfile 2.0 providers ability to execute commands based on lifecycle events.

## Acceptance Criteria 
- [ ] It should follow the specification as defined in https://github.com/che-incubator/devworkspace-api/issues/32 to support execution of the commands based on the events.
   -  [ ] implement `preStart`according to [lifecycle_events.md](https://github.com/openshift/odo/blob/master/docs/proposals/lifecycle_events.md)
   -  [x] implement `postStart` according to [lifecycle_events.md](https://github.com/openshift/odo/blob/master/docs/proposals/lifecycle_events.md) -  done in https://github.com/openshift/odo/pull/3533
   -  [ ] implement `preStop`according to [lifecycle_events.md](https://github.com/openshift/odo/blob/master/docs/proposals/lifecycle_events.md)
   -  [ ] implement `postStop`according to [lifecycle_events.md](https://github.com/openshift/odo/blob/master/docs/proposals/lifecycle_events.md)

- [x] It should replace the existing `devInit` command mechanism based on the devfile 1.0 specification.

## Links 
- Related issue: https://github.com/che-incubator/devworkspace-api/issues/32

/kind user-story
/area devfile

kadel:
/status blocked
#2918 we need devfile v2 parser first

upLukeWeston:
After discussing potential solutions (with @johnmcollier, @elsony, @EnriqueL8) we came to an agreement for the time being on how each of the lifecycle bindings could be used and what would trigger them. 

Proposed event timeline:
- Pre Start  (in initContainer) e.g:
    - Component Starts (Pod)
    - Containers Creating 
- devinit becomes PostStart E.g Parent Pom/maven install
- Build and Run commands
- Odo delete component triggers these both events:
    - PreStop before killing the component
    - PostStop after the pod/component killed: e.g Cleanup work

There are still some unanswered questions regarding differentiating commands between multiple projects within the same devfile, and the applicability of lifecycle bindings because of their intended functionalities being Workspace specific in Che.
Ongoing discussions and more details can be found on [this google doc](https://docs.google.com/document/d/1Xw8CDaGJXincggtObOeeYrgr-LrHRkWKknY-VUeXyDk/edit?usp=sharing).

kadel:
I'm afraid that it might be too early to start implementing this.
[...]

PR #3413, opened 2020-06-24
Lifecycle events proposal

upLukeWeston:
**What type of PR is this?**
/kind design
[skip ci]

This is a design proposal for adding support for devfile lifecycle events.

Relates to: #2936 

This is a design proposal, once the design is agreed upon we can submit a PR for the implementation.


docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+ - These are commands that you’d want to run before the main containers are created. 
+
+    **Note: We need some good use cases to understand when the user would run commands as preStart opposed to postStart.**
+
+**Component Initialization**
+ - After the preStart event has completed, the containers specified in the Devfile are initialised and created if the component doesn’t already exist.
+
+**postStart**
+ - If the component is newly created, the postStart events are executed sequentially within the containers given in the Devfile. 
+ - These commands could all be in the same container, or all in different ones.
```
elsony, line 47: I am not sure if this makes sense. Each command defines the container that it is supposed to be run on. So which container that the command is run on is determined by the command itself.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+ - The preStart command(s) that are specified in the Devfile will be translated into entry points for their specified containers, and added to the pod spec as init containers. 
+ - These will only run on the first odo push, or when doing a force push (odo push -f). 
+ - These are commands that you’d want to run before the main containers are created. 
+
+    **Note: We need some good use cases to understand when the user would run commands as preStart opposed to postStart.**
+
+**Component Initialization**
+ - After the preStart event has completed, the containers specified in the Devfile are initialised and created if the component doesn’t already exist.
+
+**postStart**
```
elsony, line 45: 1. Consider to add some implementation details on this design, e.g. before starting a given container in odo, we scan for all the events `postStart` events defined in the devfile and find out all the commands that uses the same container. This is the logic to identify the list of `postStart` event that is needed to run for a given container.
2. Add error handling. What if the `postStart` command failed to execute, how do we handle it and how do we surface that error to the user.
3. The type of container (i.e. the command group that links to a given container) may affect the behviour, e.g. the `postStart` only gets executed in the first `odo push` in most cases for build containers (containers used by the build command) since build containers will stay up and running after the first push. However, for `run` containers, the container may or may not get restarted depending on the `restart` flag. So the `postStart` may gets executed again for `run` containers. What about containers used by `debug` and `test` group of commands? We may want to clarify the behaviour in this proposal.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+ - If the component is newly created, the postStart events are executed sequentially within the containers given in the Devfile. 
+ - These commands could all be in the same container, or all in different ones. 
+
+**Command Execution**
+ - After postStart, we’d run the usual build/run/test/debug commands (depending on what sort of odo push parameters the user has provided) as usual.
+
+The reason we have chosen to use init containers for preStart rather than to be consistent, is because we don’t think there would be any difference in the preStart and postStart stages if they both ran in the same containers. If we were instead initialising the containers, running preStart in those containers, and then postStart in those containers, what would be different to running them all in preStart, or all in postStart? At least with init containers for preStart, there is a definitive difference between the two, and therefore reason to include them both as separate events.
```
elsony, line 52: If we ignore the `preStart` and just execute the `postStart`, we won't have this problem.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+### The flow for **odo delete** including the *preStop* and *postStop* lifecycle events will be as follows:
+**preStop**
+ - Exec the specified command(s) in their respective containers before deleting the deployment and any clean up begins.
+
+**Clean up resources**
+ - Clean up the pod and deployment etc. (as done today)
+
+**postStop**
```
elsony, line 63: 1. Similar to the `preStart`, I don't think it makes too much sense for odo to execute `postStop`. The `postStop` command has the container defined in the command itself.  If we try to execute this, we'll ended up bringing up the same container as the one that we just stopped. I think this event makes more sense in Che case since it is talking about `postStop` of workspace vs a given component in odo.
2. Add error handling as mentioned in `postStart` comment

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+ - Execute the command specified by postStop
+ - Would we be spinning up a new container exclusively for the postStop commands?
+ - Would the command need to run on the host instead for local clean up? 
+
+
+
+### Conclusions:
+ - We think that the most important event is the postStart because it has a clear, reasonable use case within odo’s flow.
```
elsony, line 71: I think both `postStart` and `preStop` are userful.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+**postStop**
+ - Execute the command specified by postStop
+ - Would we be spinning up a new container exclusively for the postStop commands?
+ - Would the command need to run on the host instead for local clean up? 
+
+
+
+### Conclusions:
+ - We think that the most important event is the postStart because it has a clear, reasonable use case within odo’s flow. 
+ - preStart and preStop could potentially be useful, but in a much more niche range. It would be important to clearly document the execution order/process for these events, because it would likely cause confusion.
```
elsony, line 72: I agree `preStart` will cause confusion but I don't think `preStop` will cause confusion at all.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
+# Support execution of the commands based on lifecycle events.
+
+## Abstract
+Add support for the lifecycle events which can be defined in 2.0.0 devfiles.
+
+Our proposed solution is to perform the preStart and postStart events as a part of the **odo push** command, and the preStop and Poststop events as a part of the **odo delete** command.
```
adisky, line 6: nit: s/Poststop/postStop

docs/proposals/lifecycle_events.md
adisky, line 45: could we add yaml definitions of events here, as how it would refer components and commands.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+ - These will only run on the first odo push, or when doing a force push (odo push -f). 
+ - These are commands that you’d want to run before the main containers are created. 
+
+    **Note: We need some good use cases to understand when the user would run commands as preStart opposed to postStart.**
+
+**Component Initialization**
+ - After the preStart event has completed, the containers specified in the Devfile are initialised and created if the component doesn’t already exist.
+
+**postStart**
+ - If the component is newly created, the postStart events are executed sequentially within the containers given in the Devfile.
```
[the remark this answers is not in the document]
upLukeWeston: Yes. For example with the following devfile content:
```
...
commands:
  - exec:
      id: firstPostStartCmd
      commandLine: echo I am the first PostStart
      component: tools
      workingDir: ${CHE_PROJECTS_ROOT}/
      group:
        kind: init
commands:
  - exec:
      id: secondPostStartCmd
      commandLine: echo I am the second PostStart
      component: tools
      workingDir: ${CHE_PROJECTS_ROOT}/
      group:
        kind: init
...
events:
  postStart:
   - "firstPostStartCmd"
   - "secondPostStartCmd"
```
We would execute them in that order - first, then second. Presumably waiting for the previous one to complete before executing the next.

docs/proposals/lifecycle_events.md
upLukeWeston: Yeah you're right. This wasn't meant to come across as a statement about the postStarts behaviour, rather it was just acknowledging how the commands would be executed.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,77 @@
...
+## Design overview
+
+The lifecycle events that have been implemented in Devfile 2.0.0 are generic. So their effect/meaning might be slightly different when applying them to a workspace (Che) or a single project (odo). However, this proposal will aim to provide a similar user experience within the scope of a single project.
+
+Our proposal is that the preStart and postStart events are designed to run as a part of **odo push**, and that the preStop and postStop events run as a part of **odo delete**.
+
+
+
+### The flow for **odo push** including the *preStart* and *postStart* lifecycle events will be as follows:
+ - preStart - Devfile command gets translated to entrypoint for specified init container - one time only
```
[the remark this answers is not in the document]
neeraj-laad: I agree, form the use cases we are aware of, `preStart` seems of no use to odo. So, we can skip supporting that event for now and add it at a later point when we have a Clea use case. 

@upLukeWeston Perhaps we should add a `Future work` section and put this in there?

docs/proposals/lifecycle_events.md
elsony: I think point#1 and point#3 have not been addressed yet.

upLukeWeston: I understand the premise of your point, but I have been struggling to think of any specific use cases where this would be a driving factor.
We (@EnriqueL8, @neeraj-laad, and I) have discussed the potential to be more granular in the solution by identifying which container specifically we are initialising, and whether or not there is a postStart command associated with it. This way we could better utilise the postStart commands and use it in cases where only come of the containers are being initialised.


A potential flow could be to check for postStart commands associated with a specific container when inititialising it, storing those commands in an Object(?) and then executing only those postStart commands when the containers have all finished initialising, in the order that they are defined in within the devfile. However, this isn't necessary for the basic implementation of the postStart event. There could be room for such a feature though, and a case could be presented to argue the validity of changing the flow of `odo push` in order to accommodate such cases in a future implementation.

I will update the proposal to reflect this though, as it was definitely something we hadn't considered.

neeraj-laad:
@elsony Does the latest update address your review comments?

elsony:
> @elsony Does the latest update address your review comments?

@neeraj-laad See comment https://github.com/openshift/odo/pull/3413#discussion_r447198157. I don't think those has been addressed yet.

elsony:
Looking at it again, the requested info have been added the info on line 135

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Add support for `postStart` commands: [container initialization support](https://github.com/openshift/odo/issues/2936)
+- Add support for `preStop` commands: 
+
+## Design overview
+Some of the events might not be useful for `odo` to adopt. We should only support the ones that have a clear use-case, and then add more as other use-cases emerge.
+
+For odo, the most important events are the `postStart`, and the `preStop` because they have clear, reasonable use case within odo’s flow. There is no clear use case within odo for `preStart` and `postStop` today.
+
+- `preStart` - Do not support in odo
```
kadel, line 22: We should support this odo. This can be easily implemented as an initContainer.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,128 @@
...
+- `postStart` - support executing these commands as part of component initialization
+- `preStop` - support executing these commands as part of the component delete
+- `postStop` - Do not support in odo
```
kadel, line 25: I think that we can just simply execute this after `preStop` commands.

docs/proposals/lifecycle_events.md
```diff
@@ -0,0 +1,128 @@
...
+- Commands associated with a lifecycle binding should provide a mechanism to figure out when they have completed. That means that the command should be terminating or having a readiness probe.
+
+- Commands associated with the same lifecycle binding do not guarantee to be executed sequentially or in any order. To run commands in a given order a user should use a composite.
+
+The flow for initializing and deleting components should be updated to add execution of commands bound to relevant lifecycle event while initializing and destroying any containers.
+
+### Initialize containers
+- Identify which containers have been created as part of creating/updating the component deployment
+  - Initial implementation can assume containers are only initialised during the component creation (first push). We can improve this in future to allow execution of postStart commands at a more granular level.
+- Iterate over the list of `postStart` commands and identify the ones that are associated with newly started containers
```
kadel, line 37: This should probably define where exactly in initialization process the `postStart` commands should be executed. 
`postStart` commands should be executed after the code is synced to the container and before the `build` command (or any other commands are executed)

docs/proposals/lifecycle_events.md
kadel, line 25: I think that this can be supported in odo. It can simply execute after `preStop` commands.

docs/proposals/lifecycle_events.md
kadel, line 22: We should support this odo. This can be easily implemented as an initContainer.

docs/proposals/lifecycle_events.md
neeraj-laad: We can support `postStart`, but it has no real value as it will be same as `preStop`. We have listed this as a futures as it might be better to support the ones where there is a clear use case and then build up.

@kadel are you recommending that we support both and do the same thing for them effectively? I feel  that might be more confusing for devfile creators.

docs/proposals/lifecycle_events.md
neeraj-laad: Yup, that was the thought, will update the PR.

docs/proposals/lifecycle_events.md
neeraj-laad: @kadel Do you have a clear use case in mind for preStart, that can't be achieved by postStart? 

We did consider init containers - convert devfile command into a command string and pass it as the command for the init container. But, weren't sure it would make sense if the command was going to use the same build/run container. So left it out as a future piece of work.

kadel: > @kadel Do you have a clear use case in mind for preStart, that can't be achieved by postStart?

Currently, I don't. 
But this is how it was defined in devfile. We should implement it. Supporting only half the events in odo is going to just confuse everyone who will try to create devfiles.
If there are no technical reasons not to implement it we should implement it.

docs/proposals/lifecycle_events.md
kadel: > We have listed this as a futures as it might be better to support the ones where there is a clear use case and then build up.

It already got into devfile spec, so someone probably has a use case for it. 


> @kadel are you recommending that we support both and do the same thing for them effectively? It might be ok for completeness but I feel that might be more confusing for devfile creators.

I don't see any problems with this.
But I see a problem when there is devfile that uses this event and someone tries to use it with odo. Would we error out and say that this devfile is unsupported? Or just show a warning and never execute the commands? Both are pretty bad options :-(

docs/proposals/lifecycle_events.md
elsony: It's just not clear to me how we can handle cleanly in odo.  If we try to support this and someone specify a command that uses the build/run container for the `preStart`, how do you even handle that? This command is meant to be called before those container get started. In order to execute those command, we need to start the container.

docs/proposals/lifecycle_events.md
elsony: It's not clear on how we want to implement this. If someone specify a `postStop` command using the build/run container, we'll be starting up the exact containers that we just stopped.

neeraj-laad: I see your point on partial implementation. However, if we consider stop is equivalent to deleting the component and containers, the flow would be:
```
- preStop
- stop (delete component and all containers)
- postStop
```

The issue now is that by the time we have done the stop/delete operation, we don't have any existing deployment where we can execute the the `postStop` commands, and I don't think we want to create another one and then delete it for this.

If I change this to what you are proposing:

```
- preStop
- postStop
- stop (delete component and all containers)
```

Then we really are changing the meaning of postStop and I would be concerned about that. 

I feel the real use case for odo could potentially come from actions that need to be take on "local" target. As in you want to clean up some temporary files from the developer's workspace on the host. That could be a good fit, but the way to execute them will be different. than running them on the cluster, and I'm not sure if odo supports that today.

Implementing this will be trivial as we just need to parse few more commands, but I'm trying to get clarity on use case before we go down this path.

docs/proposals/lifecycle_events.md
kadel: `preStart` needs to be executed before the workspace pod (in odo case the component) is started. This is exactly what `initContainer` does. Che is coming to implement this in the same way. Check discussion in https://github.com/devfile/kubernetes-api/issues/32

docs/proposals/lifecycle_events.md
kadel: > It's not clear on how we want to implement this. If someone specify a `postStop` command using the build/run container, we'll be starting up the exact containers that we just stopped.

yes, and if they define it like this it is what they wanted.

kadel: > I see your point on partial implementation. However, if we consider stop is equivalent to deleting the component and containers, the flow would be:
> 
> ```
> - preStop
> - stop (delete component and all containers)
> - postStop
> ```

That is true. This can be fairly easily implemented. We can just create standalone Pod or Job after the component is deleted.

neeraj-laad: I've tried to re-read : https://github.com/devfile/kubernetes-api/issues/32 and found something that I had missed earlier: 

`Implementation of preStart commands may need execution of a workspace pod init container. That means that the component associated to the preStart command should not be part of the workspace pod. The same applies to postStop bindings.`

There is a clear requirement that container used for preStart and postStart commands will not overlap with containers of the main workspace pod. This makes is a bit clearer and perhaps we can implement postStop as: 

- create a new deployment for running just these postStop command
- run the postStop command
- delete the deployment

or change the meaning slightly and:
- run the commands in the same component after preStop but before deleting the component and all containers

docs/proposals/lifecycle_events.md
neeraj-laad: I've tried to re-read : devfile/kubernetes-api#32 and found something that I had missed earlier:

Implementation of preStart commands may need execution of a workspace pod init container. That means that the component associated to the preStart command should not be part of the workspace pod. The same applies to postStop bindings.

There is a clear requirement that container used for preStart and postStart commands will not overlap with containers of the main workspace pod. This makes is a bit clearer and perhaps we can implement preStart as init container or take an approach similar to postStop and run them as normal container commands before we run any postStart commands. 

If there is no overlap in containers, there should not really be any difference in the outcome and it might be simpler to implement without init containers as we have all that we need for that already.

docs/proposals/lifecycle_events.md
EnriqueL8: What would we be deleting with `PostStop` when starting a new Job or Pod in the cluster? a PVC? a database? In these scenarios, we don't need to create a resource instead we would be deleting a resource. Example: a logging sidecar receiving information from the runtime container. Then it makes sense to delete the runtime container, before deleteing the logging sidecar.
Even though, I still think this step makes more sense in a Che workspace, and maybe in the local docker support were we have create volumes locally or configuration files on the host.

neeraj-laad: I'm think these commands will more be like stop the database and back up data to a volume. vs simply deleting a container.

docs/proposals/lifecycle_events.md
neeraj-laad: > It's just not clear to me how we can handle cleanly in odo. If we try to support this and someone specify a command that uses the build/run container for the `preStart`, how do you even handle that? This command is meant to be called before those container get started. In order to execute those command, we need to start the container.

@elsony Please see my comment above. This is something that I had missed too, but the spec clearly states that there should not be any overlap of containers between main workspace containers (build, run, test etc.) and preStart, postStop.

neeraj-laad: I think we should go with preStart => initContainers and only run them when the component is bieng created.

docs/proposals/lifecycle_events.md
neeraj-laad: We can update the proposal to use a standalone pod for postStart commands and then delete the pod. all of this will be done as part of `odo delete`.

I'll reach out to the devfiles team to get more clarity on the use case they had in mind too.

docs/proposals/lifecycle_events.md
kadel: > when the component is bieng created.

This part might be tricky. I'm not sure if it will be possible. The initContainer will be executed everytime Deployment is triggered. Which might be right, or not :-( depending on what the command is doing.

neeraj-laad: I was thinking that we could add init containers to the deployment.yaml only for the first push. and consecutive push, we can generate the deployment without init containers. But these can be looked at in more details during implementation phase as each will be implemented under separate issue.

For now we can update the proposal to include:

`preStart` - execute commands via init containers
`postStop` - execute commands via standalone pod or job and delete it once complete


@elsony @kadel would this make sense.

neeraj-laad:
@kadel Is there any other changes needed on this proposal? I believe we have addressed all the comments discussed above.

Issue #3436, opened 2020-06-26
odo create interactive mode is broken for devfiles


johnmcollier:
/kind bug
/area devfile

## What versions of software are you using?

**Operating System:**
macOS

**Output of `odo version`:**
master

It seems that `odo create`'s interactive mode is broken, and gives the following error:
```
Johns-MacBook-Pro-3:nodejs johncollier$ odo create
Experimental mode is enabled, use at your own risk

? Which devfile component type do you wish to create nodejs
? What do you wish to name the new devfile component nodejs
? What namespace do you want the devfile component to be created in odo
Validation
 ✗  Checking devfile existence [52460ns]
 ✗  Checking devfile compatibility [31039ns]
 ⚠  Devfile component type nodejs is not supported, please run `odo catalog list components` for a list of supported devfile component types
 ✗  component type not found
```

It complains that the devfile component isn't supported (even though it is).

maysunfaisal:
can be closed, fixed by https://github.com/openshift/odo/pull/3291

PR #3437, opened 2020-06-26
Fix `odo create` interactive mode

johnmcollier:

**What type of PR is this?**
/kind bug

This PR fixes interactive mode for devfiles. It seems a recent change that was delivered used a short assignment operator (:=) to assign a value to the `catalogDevfileList` variable, which caused a local-scoped version of it to get created in the if-block. This meant the value of `catalogDevfileList` couldn't be seen outside of the if-statement (despite the variable being declared outside it)

Fixes https://github.com/openshift/odo/issues/3436

`odo create`

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -398,7 +398,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			// Interactive mode
 
 			// Component type: We provide supported devfile component list to let user choose
-			catalogDevfileList, err := catalog.ListDevfileComponents(co.devfileMetadata.devfileRegistry.Name)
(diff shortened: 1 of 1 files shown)


GeekArthur:
Good catch, I can review and approve this once #3291 merges in.

maysunfaisal:
@johnmcollier this is actually getting fixed with #3291

johnmcollier:
@maysunfaisal Ah, good catch. But, interactive mode is broken on your PR it seems:
Screen Shot 2020-06-26 at 5 53 44 PM

johnmcollier:
@maysunfaisal I've updated the PR description for https://github.com/openshift/odo/pull/3291 to include #3436

maysunfaisal:
@johnmcollier yeah I should have opened issues earlier to keep track. There was one more issue with interactive mode that I am forgetting but its fixed by #3291

PR #3453, opened 2020-06-30
added deprecation and error messages for 'odo update`

### LINKED ISSUE #3396: deprecation warning for odo update
## Background:
The original intention for  `odo update`  was to provide a path forward for outer-loop by switching to git component.
We don't need that anymore as there is work in progress for proper outer-loop (`odo deploy` `odo pipelines`)

## Acceptance criteria 
- [ ] When `odo update` command is executed in a directory with devfile component it should display error: 
  ```
  'odo update' command is not available for Devfile based components.
  ```
- [ ] When `odo update` is executed with LocalConfig component it should display the following warning on as the first line of the output: 
  ```
  WARNING: 'odo update' command will be removed in the future odo version.
  You should be using `odo config` command instead.
  example:
    odo config set SourceType git
    odo config set SourceLocation https://github.com/example/example
  ```

/kind user-story
/priority high
/area update

dev-gaur:
/kind deprecation

added deprecation and error messages for `odo update`

Fixes #3396 

run `odo update` for 1. s2i components (expect a deprecation warning)
 2. devfile component (expect an error)

Its a Draft right now

### DIFF
--- a/pkg/odo/cli/component/update.go
+++ b/pkg/odo/cli/component/update.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
@@ -17,6 +18,8 @@ import (
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	odoutil "github.com/openshift/odo/pkg/odo/util"
 	"github.com/openshift/odo/pkg/odo/util/completion"
+	"github.com/openshift/odo/pkg/odo/util/experimental"
+	"github.com/openshift/odo/pkg/util"
 
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
 )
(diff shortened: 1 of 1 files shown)


girishramnani:
adding WIP in the label as this is still draft PR

girishramnani:
@dev-gaur please make this as non-draft PR by clicking the `Ready for review` button

pkg/odo/cli/component/update.go
```diff
@@ -44,6 +49,14 @@ var updateCmdExample = ktemplates.Examples(`  # Change the source code path of c
 	  %[1]s --binary ./downloads/sample.war
 		`)
 
+const deprecationWarning string = "WARNING: 'odo update' command will be removed in the future odo version.\n" +
+	"You should be using `odo config` command instead.\n" +
```
adisky, line 53: I think we should add the same warning  in `odo update --help` 
Also can we change the color of message to yellow?

girishramnani:
Please address @adisky , everything else good.
/approve

pkg/odo/cli/component/update.go
dev-gaur: image


changes made.

Issue #3440, opened 2020-06-29
odo watch does not bother about component name


amitkrout:
/kind bug

## What versions of software are you using?

**Operating System:**
supported
**Output of `odo version`:**
master
## How did you run odo exactly?
```
$ odo project create test123

$ oc --request-timeout 5m import-image nodejs --namespace=test123 --from=registry.access.redhat.com/rhscl/nodejs-10-rhel7:latest --confirm

$ oc annotate istag/nodejs:latest --namespace=test123 tags=builder --overwrite

$ odo create nodejs nodejs-app --project test123 --context tests/examples/source/nodejs/

$ odo push --context tests/examples/source/nodejs/

go to another terminal and run the watch command with a random argument

$ odo watch randomname --context tests/examples/source/nodejs/
Waiting for something to change in /Users/amit/go/src/github.com/openshift/odo/tests/examples/source/nodejs
```

## Actual behavior
it does not bother about the ```randomname``` argument

## Expected behavior
should throw error something like ```✗  accepts at most 1 arg(s), received 2```

girishramnani:
so should it bother about `randomname` or not?

also `randomname` is the first argument so why should it complain
`odo watch randomname randomname2` complains about 
```
✗ accepts at most 1 arg(s), received 2
```
So I am a bit confused. can you please clarify the issue description @amitkrout

amitkrout:
> so should it bother about randomname or not?

If you see may steps to reproduce, i have created a component of name ```nodejs-app```. In the next step i am passing a none existing component name along with watch and --context flag. Ideally it should report ```randomname``` is not present in the cluster, but does not care the component name instead it picks the correct component name whatever passed through the context flag.

> also randomname is the first argument so why should it complain

Yes, its the first argument, but not the correct component name

> odo watch randomname randomname2 complains about
✗ accepts at most 1 arg(s), received 2

Yup, this is right. I have updated the description. Please revisit.

amitkrout:
@girishramnani Few more inputs may be you need to address through your pr #3457

In regular mode 
```
$ odo watch -h
Watch for changes, update component on change. Watch doesn't provide support for git component.
Usage:
  odo watch [flags]

```
In experimental mode
```
$ odo watch -h
Watch for changes, update component on change. Watch doesn't provide support for git component.
Usage:
  odo watch [component name] [flags]
```

In experimental mode odo watch takes ```component_name``` as an argument where as in regular mode there is no place for ```component_name```.

PR #3457, opened 2020-06-30
update odo watch to remove [component name] arg also resolved bugs in odo watch test utils

girishramnani:
**What type of PR is this?**

 /kind cleanup

`odo watch` commands allows a `[component name]` as 1st arg which is not being used.

Fixes https://github.com/openshift/odo/issues/3440

`odo watch -h` now shouldn't show the `[component name]`

### DIFF
--- a/tests/helper/helper_generic.go
+++ b/tests/helper/helper_generic.go
@@ -108,8 +108,18 @@ func ExtractSubString(output, start, end string) string {
 func WatchNonRetCmdStdOut(cmdStr string, timeout time.Duration, check func(output string) bool, startSimulationCh chan bool, startIndicatorFunc func(output string) bool) (bool, error) {
 	var cmd *exec.Cmd
 	var buf bytes.Buffer
-
-	cmdStrParts := strings.Split(cmdStr, " ")
+	var errBuf bytes.Buffer
+	var cmdStrParts []string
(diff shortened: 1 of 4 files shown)


girishramnani:
@dgolovin please take a look at this change. This removes the `<component name>` from odo watch

tests/helper/helper_generic.go
```diff
@@ -119,6 +129,7 @@ func WatchNonRetCmdStdOut(cmdStr string, timeout time.Duration, check func(outpu
 		cmd = exec.Command(cmdName)
 	}
 	cmd.Stdout = &buf
+	cmd.Stderr = &errBuf
```
girishramnani, line 132: @prietyc123 we never ever ignore the stderr in a subprocess execution. any error in the `odo watch` command would just stall the tests for 5mins

girishramnani: cc @prietyc123

amitkrout:
@girishramnani PTAL- https://github.com/openshift/odo/issues/3440#issuecomment-652846425

girishramnani:
Already addressed @amitkrout

girishramnani:
I don’t see anywhere that the component name was actually used in experimental mode as well @amitkrout

PR #3464, opened 2020-07-01
Building redistributable only on x86_64 and updating golang version to 1.13

mohammedzee1000:
We dont need redistributable package building
on multiple architectures as it is happening right
now (case in point internal scratch build)

Eg:
  - openshift-odo-redistributable-1.2.3-1.el8.s390x.rpm
  - openshift-odo-redistributable-1.2.3-1.el8.ppc64le.rpm
  - openshift-odo-redistributable-1.2.3-1.el8.x86_64.rpm

We only need last one as individual rpms for specific architectures
is fine but not the redistributable ones.

Internal build system now supports golang 1.13

**What type of PR is this?**

/kind cleanup

See above

None, this is cleanup in preparation of new build system

You can double check by building using golang 1.13 for any breakages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jobs:
     - &base-test
       stage: test
       go_import_path: github.com/openshift/odo
-      go: "1.12.x"
+      go: "1.13.x"
       install:
         - make goget-tools
       script:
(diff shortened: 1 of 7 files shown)


girishramnani:
Actually
/approve cancel
We need Z arch because there is a team dependent on that

mohammedzee1000:
@girishramnani that will happen automatically in build system .The redistributable package will be build in x86_64 just not in all other architectures as there is no point having redistributable package on 3 different archs

Issue #3517, opened 2020-07-08
Update "Understanding odo" documentation to correspond to officially supported Devfiles


cdrage:
/kind documentation

## What mistake did you find / what is missing in the documentation?

See:

https://odo.dev/docs/understanding-odo/#officially-supported-languages-and-corresponding-container-images

and the corresponding documentation "downstream"

https://github.com/openshift/openshift-docs/blob/ad655a7d0a3b69ab89f7516c593130a6ebbdbc43/cli_reference/developer_cli_odo/understanding-odo.adoc

We must create out own "understanding odo" documentation that corresponds to Devfile rather than S2I. 

## What is the relevance of it?

Needed for Devfile.

PR #3519, opened 2020-07-08
Add "Understanding odo" Documentation

cdrage:
**What type of PR is this?**

/kind documentation
[skip ci]

Since we are doing a future Devfile release of odo that uses Devfile, we
require our "Understanding odo" guide to be located upstream rather than
downstream (from github.com/openshift/openshift-docs).

This adds that documentation.

Closes https://github.com/openshift/odo/issues/3517

N/A

### DIFF
--- a/docs/public/understanding-odo.adoc
+++ b/docs/public/understanding-odo.adoc
@@ -1,0 +1,92 @@
+= Understanding `odo`
+
+`odo` is a CLI tool for creating applications on OpenShift and Kubernetes. `odo` allows developers to concentrate on creating applications without the need to administer a cluster itself.
+Creating deployment configurations, build configurations, service routes and other OpenShift or Kubernetes elements are all automated by `odo`.
+
+Existing tools such as `oc` are more operations-focused and require a deep understanding of Kubernetes and OpenShift concepts. `odo` abstracts away complex Kubernetes and OpenShift concepts allowing developers to focus on what is most important to them: code.
+
+[id="odo-key-features"]
+== Key features
+
+`odo` is designed to be simple and concise with the following key features:
+
+* Simple syntax and design centered around concepts familiar to developers, such as projects, applications, and components.
+* Completely client based. No additional server other than OpenShift is required for deployment.
+* Official support for Node.js and Java components.
+* Detects changes to local code and deploys it to the cluster automatically, giving instant feedback to validate changes in real time.
+* Lists all the available components and services from the cluster.
+
+== Core concepts
+
+Project::
+A project is your source code, tests, and libraries organized in a separate single unit.
+Application::
+An application is a program designed for end users. An application consists of multiple microservices or components that work individually to build the entire application.
+Examples of applications: a video game, a media player, a web browser.
+Component::
+A component is a set of Kubernetes resources which host code or data. Each component can be run and deployed separately.
+Examples of components: Node.js, Perl, PHP, Python, Ruby.
+Service::
+A service is software that your component links to or depends on.
+Examples of services: MariaDB, Jenkins, MySQL.
+Devfile::
+A portable file responsible for your entire reproducable development environment.
+
+[id="odo-supported-devfiles"]
+=== Official Devfiles
+
+Devfiles describe your development environment link. link:https://odo.dev/docs/deploying-a-devfile-using-odo/[Click here for more information on Devfile.]
+
+All official example Devfiles are hosted on the link:https://github.com/odo-devfiles/registry[registry].
+
+.List of Devfiles which are officially supported by odo
+[options="header"]
+|===
+|Language | Devfile Name | Description | Devfile Source
+
+| Java
+| java-maven
+| Upstream Maven and OpenJDK 11
+| link:https://github.com/odo-devfiles/registry/blob/master/devfiles/java-maven/devfile.yaml[java-maven/devfile.yaml]
+
+| Java
+| java-openliberty
+| Open Liberty microservice in Java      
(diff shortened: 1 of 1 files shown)


docs/public/understanding-odo.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+== Core concepts
+
+Project::
+A project is your source code, tests, and libraries organized in a separate single unit.
+Application::
+An application is a program designed for end users. An application consists of multiple microservices or components that work individually to build the entire application.
+Examples of applications: a video game, a media player, a web browser.
+Component::
+A component is a set of Kubernetes resources which host code or data. Each component can be run and deployed separately.
```
adisky, line 27: `A component is as set of Kubernetes resources` I find it little confusing,.
may be something like this
A component is a part of application or application itself that could be atomically deployed on kubernetes.
e.g a web application can have front-end and back-end components

docs/public/understanding-odo.adoc
```diff
@@ -0,0 +1,92 @@
...
+A service is software that your component links to or depends on.
+Examples of services: MariaDB, Jenkins, MySQL.
+Devfile::
+A portable file responsible for your entire reproducable development environment.
+
+[id="odo-supported-devfiles"]
+=== Official Devfiles
+
+Devfiles describe your development environment link. link:https://odo.dev/docs/deploying-a-devfile-using-odo/[Click here for more information on Devfile.]
+
```
adisky, line 39: for odo context, devfiles more of defines the stack required to run the application

docs/public/understanding-odo.adoc
```diff
@@ -0,0 +1,92 @@
...
+All official example Devfiles are hosted on the link:https://github.com/odo-devfiles/registry[registry].
+
+.List of Devfiles which are officially supported by odo
+[options="header"]
+|===
+|Language | Registry Name | Description | Source
+
```
adisky, line 46: here source signifies yaml devfiles, and we are referring starter projects source code.

docs/public/understanding-odo.adoc
```diff
@@ -0,0 +1,92 @@
...
+| Java
+| java-maven
```
adisky, line 48: `java-maven` is devfile name not registry name

docs/public/understanding-odo.adoc
adisky: We can update table as
|Language | Devfile Name | Description | Source | Starter Projects

cdrage: We describe above that all the Devfiles are located in the registry, I don't see why we have to add source + starter projects to here, that's just my opinion

docs/public/understanding-odo.adoc
cdrage: Hi Aditi, that part was approved by the OpenShift docs team, I'm unsure if we should update this to say "stack"

docs/public/understanding-odo.adoc
adisky: not saying to add starter projects, but in `source` you are referring to starter projects. it should refer devfile.yaml
similarly what you are saying as registry name is devfile name.

docs/public/understanding-odo.adoc
cdrage: This was copy-and-pasted from the OpenShift docs which are already approved. To be honest, I don't find it that confusing since a component does describe Kubernetes resources.

docs/public/understanding-odo.adoc
cdrage: Updated! Thanks

docs/public/understanding-odo.adoc
cdrage: Updated! Thanks

cdrage:
Updated, thanks @adisky

docs/public/understanding-odo.adoc
adisky: no hard opinion here, its fine.

docs/public/understanding-odo.adoc
adisky: no hard opinion here, its fine.

docs/public/understanding-odo.adoc
```diff
@@ -0,0 +1,92 @@
...
+| Upstream Maven and OpenJDK 11
+| link:https://github.com/odo-devfiles/maven-ex[odo-devfiles/maven-ex]
+
+| Java
+| java-openliberty
+| Open Liberty microservice in Java      
+| link:https://github.com/odo-devfiles/openliberty-ex[odo-devfiles/openliberty-ex]
```
adisky, line 55: May be i am missing something, I am not getting why we want to refer example source here?
shouldn't we refer devfile.yaml?
 https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml

cdrage: See: https://github.com/openshift/odo/pull/3519/files#diff-37da5e46febfdc71c368cd55ff0d6a2fR40

I guess we can add both a link to the Devfile.yaml as well as the starter code?

adisky: yes, i have seen that, we are already pointing to registry. but what is the need of pointing starter projects? we are already able to download with `--starter` flag.
IMO with all the devfiles details (name, registry, description) user expected to see its contents, not source code for starter projects.

cdrage: Thanks! I've updated the table with your suggestions. You're right, we should include a direct link to the devfile rather than the starter project.

cdrage:
Going to merge this in for now so we can start syncing with the current documentation (this will not appear on the website just yet)

Issue #3338, opened 2020-06-11
Support composite commands


elsony:
/kind user-story

## User Story
As a stack creator, I want to define composite commands to chain multiple commands together when running in odo.

## Acceptance Criteria 
- [ ] It should follow the corresponding settings specified in the devfile 2.0 specification: https://github.com/devfile/kubernetes-api/issues/18
- [ ] It should support running composite command for all command groups that odo supports, e.g. build, run, etc.
- [ ] If parallel flag is set to true, the commands should be executed sequentially based on the order specified in the composite command; otherwise, the command will be executed in parallel.

This item will only cover the composite command for the build, not including the run command.

## Links 
- Related issue: https://github.com/devfile/kubernetes-api/issues/18

/kind user-story
/area devfile
/area devfilev2

johnmcollier:
Sample Devfile 2.0 with composite commands:
```
schemaVersion: 2.0.0
metadata:
  name: openLiberty
  version: 1.0.0
  description: Java application stack using Open Liberty runtime
projects:
  - name: user-app
    git:
      location: 'https://github.com/odo-devfiles/openliberty-ex.git'
      branch: master
components:
  - chePlugin:
      id: redhat/java/latest
      memoryLimit: 1512Mi
  - container:
      name: appsodyrun
      image: 'ajymau/java-openliberty-dev:latest'
      memoryLimit: 1024Mi
      mountSources: true
      endpoints:
        - name: 9080/tcp
          configuration:
            discoverable: false
            public: true
            protocol: tcp
            scheme: http
          targetPort: 9080
      env:
        - name: MODE2
          value: TEST2
        - name: myprop2
          value: myval2
commands:
  - exec:
      id: build
      component: appsodyrun
      commandLine: mvn -DhotTests=true -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository liberty:dev
      workingDir: /projects/user-app
      attributes:
        restart: "false"
      group:
        kind: run
        isDefault: true
   - exec:
       id: run
       commandLine: "./run.sh"
       component: tooling
       group: 
          kind: run
   - composite:
       id: buildAndRun
       label: Build and Run
       commands:
         - build
         - run
       parallel: false
       group: 
          kind: run
```

Based on the second point, it sounds like this PR depends on support for command groups in devfiles (https://github.com/openshift/odo/issues/2938)

johnmcollier:
Some questions on composite commands that I couldn't find answers for in the schema:
[...]

PR #3551, opened 2020-07-13
Add support for composite commands on Kubernetes

johnmcollier:
**What type of PR is this?**
/kind feature

This PR implements support for composite commands on Kubernetes, [as defined in the Devfile 2.0 schema](https://devfile.github.io/devfile/_attachments/api-reference.html). Support for running parallel composite commands is also included.

Because of the nature of the supervisord integration for long-running run commands, composite run commands are not supported in this PR and a separate story will be opened for it. 

In all, this PR does the following:
- Updates the devfile validation to consider composite commands to be valid
- Unit testing for the updated devfile validation
- Updates the devfile exec code for Kubernetes to allow composite commands to be executed
    - Uses `util.ConcurrentTask` to execute the parallel commands
    - Supports nested composite commands
- Integration tests for both non-parallel and parallel composite commands

Fixes https://github.com/openshift/odo/issues/3338

- Run devfile push integration tests
- Or test with the following devfile:
```
schemaVersion: 2.0.0
metadata:
  name: nodejs
  version: 1.0.0
projects:
  - name: nodejs-starter
    git:
      location: "https://github.com/odo-devfiles/nodejs-ex.git"
components:
  - container:
      name: runtime
      image: registry.access.redhat.com/ubi8/nodejs-12:1-36
      memoryLimit: 1024Mi
      mountSources: true
      endpoints:
        - name: http-3000
          targetPort: 3000
          configuration:
            protocol: tcp
            scheme: http
            type: terminal
commands:
  - exec:
      id: install
      component: runtime
      commandLine: npm install
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
      group:
        kind: build
        isDefault: false
  - exec:
      id: echo
      component: runtime
      commandLine: echo hi
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
      group:
        kind: build
        isDefault: false
  - exec:
      id: run
      component: runtime
      commandLine: npm start
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
      group:
        kind: run
        isDefault: true
  - composite:
         id: buildAndEcho
         label: Build and Echo
         commands:
           - echo
           - install
         parallel: false
         group: 
            kind: build
            isDefault: true
```


pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +179,35 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
 	return
 }
 
+// validateCompositeCommand checks that the specified composite command is valid
+func validateCompositeCommand(data data.DevfileData, compositeCommand *common.Composite) error {
+	if compositeCommand.Group.Kind == common.RunCommandGroupType {
+		return fmt.Errorf("compoosite commands cannot belong to the run grouop")
```
mik-dass, line 185: `composite`. Also I think the message should be `composite commands of run Kind are not supported currently`.

pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +179,35 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
...
+	}
+
+	// Loop over the commands and validate that each command points to a command that's in the devfile
+	for _, command := range compositeCommand.Commands {
+		if command == compositeCommand.Id {
+			return fmt.Errorf("the composite command %q cannot reference itself", compositeCommand.Id)
+		}
+
+		if !hasCommand(data, command) {
```
mik-dass, line 194: Maybe store in a map and search for faster searching

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -349,10 +352,19 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 		// Get Init Command
 		command, ok := commandsMap[versionsCommon.InitCommandGroupType]
 		if ok {
-			compInfo.ContainerName = command.Exec.Component
-			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
-			if err != nil {
-				return err
+			if command.Composite != nil {
+				fmt.Println("Composite detected")
+				//err = a.execCompositeCommand(command.Composite, compInfo, show, podName)
```
mik-dass, line 357: comment found

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -349,10 +352,19 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 		// Get Init Command
 		command, ok := commandsMap[versionsCommon.InitCommandGroupType]
 		if ok {
-			compInfo.ContainerName = command.Exec.Component
-			err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
-			if err != nil {
-				return err
+			if command.Composite != nil {
+				fmt.Println("Composite detected")
```
mik-dass, line 356: this should be a log statement

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -362,10 +374,18 @@ func (a Adapter) execDevfile(commandsMap common.PushCommandsMap, componentExists
 	// Get Build Command
 	command, ok := commandsMap[versionsCommon.BuildCommandGroupType]
 	if ok {
-		compInfo.ContainerName = command.Exec.Component
-		err = exec.ExecuteDevfileBuildAction(&a.Client, *command.Exec, command.Exec.Id, compInfo, show, a.machineEventLogger)
-		if err != nil {
-			return err
+		if command.Composite != nil {
+			//err = a.execCompositeCommand(command.Composite, compInfo, show, podName)
```
mik-dass, line 378: comment found

tests/examples/source/devfiles/nodejs/devfileCompositeCommandsParallel.yaml
```diff
@@ -0,0 +1,56 @@
...
+      component: runtime
+      commandLine: npm install
+      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
+      group:
+        kind: build
+        isDefault: false
+  - exec:
+      id: echo
+      component: runtime
+      commandLine: echo hi
```
mik-dass, line 34: maybe a folder creation operation and after execution, we check if the folder was created or not

tests/examples/source/devfiles/nodejs/devfileCompositeCommands.yaml
```diff
@@ -0,0 +1,56 @@
...
+      component: runtime
+      commandLine: npm install
+      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
+      group:
+        kind: build
+        isDefault: false
+  - exec:
+      id: echo
+      component: runtime
+      commandLine: echo hi
```
mik-dass, line 34: same as https://github.com/openshift/odo/pull/3551/files#r454981491

pkg/devfile/adapters/common/command.go
johnmcollier: Ah, good idea! I did that for the other instances where I had to search for a command, I should do the same here.

pkg/devfile/adapters/kubernetes/component/adapter.go
johnmcollier: The println can be removed too, will fix that 😄

pkg/devfile/adapters/kubernetes/component/adapter.go
johnmcollier: Debugging statement that I accidentally left in, can be removed 👍

tests/examples/source/devfiles/nodejs/devfileCompositeCommandsParallel.yaml
johnmcollier: Sure, that sounds good 👍

pkg/devfile/adapters/common/command.go
johnmcollier: 👍

johnmcollier:
@mik-dass Thanks for reviewing! I've gone and addressed your review comments 🙂

tests/examples/source/devfiles/nodejs/devfileCompositeCommands.yaml
johnmcollier: 👍

johnmcollier:
Flaky CI

/retest

pkg/devfile/adapters/common/command.go
```diff
@@ -39,8 +40,15 @@ func getCommandFromDevfile(data data.DevfileData, groupType common.DevfileComman
 	}
 
 	for _, command := range commands {
-		// exec.Group is a pointer, to avoid null pointer
-		if command.Exec.Group != nil && command.Exec.Group.Kind == groupType {
+		if command.Composite != nil && command.Composite.Group != nil && command.Composite.Group.Kind == groupType {
```
maysunfaisal, line 43: Can the if and else if be combined? since their blocks are doing the same thing and its only the condition thats different

pkg/devfile/adapters/common/command.go
```diff
@@ -109,7 +117,9 @@ func validateCommandsForGroup(data data.DevfileData, groupType common.DevfileCom
 
 	if len(commands) > 1 {
 		for _, command := range commands {
-			if command.Exec.Group.IsDefault {
+			if command.Composite != nil && command.Composite.Group.IsDefault {
```
maysunfaisal, line 120: we're validating command defaults here by group kind, so i think even this if and else if can be combined for a single condition..

pkg/devfile/adapters/common/command.go
```diff
@@ -133,9 +143,14 @@ func validateCommandsForGroup(data data.DevfileData, groupType common.DevfileCom
 // 4. command must have group
 func validateCommand(data data.DevfileData, command common.DevfileCommand) (err error) {
 
-	// type must be exec
-	if command.Exec == nil {
-		return fmt.Errorf("command must be of type \"exec\"")
+	// type must be exec or composite
+	if command.Exec == nil && command.Composite == nil {
```
maysunfaisal, line 147: Pls update the function description to include composite command as well

pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +179,38 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
...
+	if compositeCommand.Group.Kind == common.RunCommandGroupType {
+		return fmt.Errorf("composite commands of run Kind are not supported currently")
+	}
+
+	commandsMap := GetCommandsMap(data.GetCommands())
+
+	// Loop over the commands and validate that each command points to a command that's in the devfile
+	for _, command := range compositeCommand.Commands {
+		if command == compositeCommand.Id {
+			return fmt.Errorf("the composite command %q cannot reference itself", compositeCommand.Id)
```
maysunfaisal, line 193: 👍

pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +179,38 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
...
+		}
+
+		_, ok := commandsMap[command]
+		if !ok {
+			return fmt.Errorf("the command %q does not exist in the devfile", command)
```
maysunfaisal, line 198: Can we update this error msg to say the command mentioned in the composite command does not exist in the devfile?

pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +179,38 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
...
+		}
+	}
+	return nil
+}
+
+// hasCommand returns true if the devfile contains the specified command
+func hasCommand(data data.DevfileData, command string) bool {
```
maysunfaisal, line 205: I assume you were using this before switching to map, looking at the previous review comments?
Since you left this here, there is a chance someone else may use it instead of the map. Can we move the map logic to this function, so that anyone can use it?

pkg/devfile/adapters/common/command.go
```diff
@@ -206,7 +253,13 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
 	} else if !isInitCmdEmpty && initCmdErr == nil {
 		isInitCommandValid = true
 		commandMap[common.InitCommandGroupType] = initCommand
-		klog.V(4).Infof("Init command: %v", initCommand.Exec.Id)
+		var commandID string
+		if initCommand.Composite != nil {
```
maysunfaisal, line 257: you can use your helper func `GetID()`?

pkg/devfile/adapters/common/command.go
```diff
@@ -219,14 +272,26 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
 	} else if !reflect.DeepEqual(emptyCommand, buildCommand) && buildCmdErr == nil {
 		isBuildCommandValid = true
 		commandMap[common.BuildCommandGroupType] = buildCommand
-		klog.V(4).Infof("Build command: %v", buildCommand.Exec.Id)
+		var commandID string
+		if buildCommand.Composite != nil {
```
maysunfaisal, line 276: you can use your helper func `GetID()`?

pkg/devfile/adapters/common/command.go
```diff
@@ -219,14 +272,26 @@ func ValidateAndGetPushDevfileCommands(data data.DevfileData, devfileInitCmd, de
...
+		klog.V(4).Infof("Build command: %v", commandID)
 	}
 
 	runCommand, runCmdErr := GetRunCommand(data, devfileRunCmd)
 	if runCmdErr == nil && !reflect.DeepEqual(emptyCommand, runCommand) {
 		isRunCommandValid = true
 		commandMap[common.RunCommandGroupType] = runCommand
-		klog.V(4).Infof("Run command: %v", runCommand.Exec.Id)
+		var commandID string
+		if runCommand.Composite != nil {
```
maysunfaisal, line 289: you can use your helper func `GetID()`?

pkg/devfile/adapters/common/command.go
maysunfaisal: Can we add some comp cmd unit test cases for this func?

pkg/devfile/adapters/common/command.go
```diff
@@ -73,7 +81,7 @@ func getCommandFromFlag(data data.DevfileData, groupType common.DevfileCommandGr
 	commands := data.GetCommands()
 
 	for _, command := range commands {
-		if command.Exec.Id == commandName {
+		if command.GetID() == commandName {
```
maysunfaisal, line 84: Can we add some comp cmd unit test cases for this func?

pkg/devfile/adapters/common/command.go
maysunfaisal: Can we add some comp cmd unit test cases for this func?

pkg/devfile/adapters/common/command.go
maysunfaisal: Can we add some comp cmd unit test cases for this func?

pkg/devfile/adapters/common/utils.go
```diff
@@ -128,6 +128,8 @@ func getCommandsByGroup(data data.DevfileData, groupType common.DevfileCommandGr
 	for _, command := range data.GetCommands() {
 		if command.Exec != nil && command.Exec.Group != nil && command.Exec.Group.Kind == groupType {
 			commands = append(commands, command)
+		} else if command.Composite != nil && command.Composite.Group != nil && command.Composite.Group.Kind == groupType {
```
maysunfaisal, line 131: if and else-if condition can be combined here too?

pkg/devfile/adapters/common/utils.go
```diff
@@ -191,3 +193,14 @@ func IsRestartRequired(command common.DevfileCommand) bool {
 
 	return restart
 }
+
+// GetCommandsMap returns a mapping of all of devfile command names to their corresponding DevfileCommand struct
+// Allowing us to easily retrieve the DevfileCommand of any command listed in a composite command
+func GetCommandsMap(commands []common.DevfileCommand) map[string]common.DevfileCommand {
```
maysunfaisal, line 199: probably a unit test for this func since its exported

adisky:
@johnmcollier few questions i have

1.  if we have a build command  like `echo2` not specified in composite command, would it be executed? if yes then after composite command or before?
```
- exec:
      id: echo
      component: runtime
      commandLine: echo hi
- exec:
      id: echo2
      component: runtime
      commandLine: echo hello
- exec:
      id: install
      component: runtime
      commandLine: npm install
  - composite:
         id: buildAndEcho
         label: Build and Echo
         commands:
           - echo
           - install
         parallel: false
         group: 
            kind: build
            isDefault: true
```
2. what is the purpose of specifying group in composite commands, if we already have group specified in command?
3. what would happen if group defined in composite command differ from command reference?
4. can composite commands be of mix type(build, run) ?

pkg/devfile/adapters/common/command.go
adisky, line 43: I believe we are overpopulating this function with complex logic,  @johnmcollier is it possible to separate out implementation for exec and composite commands?

johnmcollier: Yeah, if I had a helper function to the Command type to receive the Kind, it can be simplified

pkg/devfile/adapters/common/command.go
johnmcollier: What's funny is that these code blocks is what made me implement `GetID` as a helper function, but then I forgot to update these 😆

johnmcollier:
@adisky Good questions

1. `echo2` would only be executed in that scenario if it belonged to one of the groups (run, build, init, debug, test) *and* was the specified command for that group (either by being marked default, or passed in as a flag). In your scenario, it doesn't belong to a group, and wouldn't be executed

2. In the case of composite commands, we only care about the group that the composite command belongs to and not the group that its subcommands belong to. It's valid for a composite command in one group to reference commands in other groups.

3. See 2. It's valid for a composite command to reference commands in other groups.

4. Yes

adisky:
> @adisky Good questions
> 
> 1. `echo2` would only be executed in that scenario if it belonged to one of the groups (run, build, init, debug, test) _and_ was the specified command for that group (either by being marked default, or passed in as a flag). In your scenario, it doesn't belong to a group, and wouldn't be executed

if i specify group an default both
```
- exec:
      id: echo2
      component: runtime
      commandLine: echo hello
      group:
        kind: build
        isDefault: true
```
then when it would get executed before or after composite?

johnmcollier:
@adisky It's not valid for there to be more than one default command per group, in which case a validation error would be thrown.

adisky:
> @adisky It's not valid for there to be more than one default command per group, in which case a validation error would be thrown.

yea, forgot that got it know, Thanks :)

pkg/devfile/adapters/common/command.go
adisky: @johnmcollier can we split this from `getCommand` to `getExecCommandFromDevfile`  and `getCompositeCommandFromDevfile`

johnmcollier:
@maysunfaisal I've addressed your review comments, please re-review when you get the chance, thanks!

pkg/devfile/adapters/common/command.go
johnmcollier: @adisky Using the command helper functions I wrote, I was able to significantly simplify this function

pkg/devfile/adapters/common/utils.go
johnmcollier: Done

pkg/devfile/adapters/common/utils.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

pkg/devfile/adapters/common/command.go
johnmcollier: Done

johnmcollier: So I was able to avoid needing to create two separate functions 🙂

adisky: ya, it looks good now :), sorry i hadn't seen revised commit

pkg/exec/devfile.go
```diff
@@ -222,6 +224,61 @@ func ExecuteDevfileDebugActionWithoutRestart(client ExecClient, exec common.Exec
...
+		// Execute the commands in order
+		for _, command := range composite.Commands {
+			if devfileCommand, ok := commandsMap[strings.ToLower(command)]; ok {
+				err = execCommandFromComposite(client, devfileCommand, commandsMap, compInfo, show, machineEventLogger)
+				if err != nil {
+					return fmt.Errorf("command execution failed: %v", err)
+				}
+			} else {
+				// Devfile validation should have caught a missing command earlier, but should include error handling here as well
+				return fmt.Errorf("command %q not found in devfile", command)
```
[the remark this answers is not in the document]
maysunfaisal: same in parallel block

pkg/exec/devfile.go
```diff
@@ -222,6 +224,61 @@ func ExecuteDevfileDebugActionWithoutRestart(client ExecClient, exec common.Exec
...
+// The composite command may reference exec commands, composite commands, or both
+func ExecuteCompositeDevfileAction(client ExecClient, composite common.Composite, commandsMap map[string]common.DevfileCommand, compInfo adaptersCommon.ComponentInfo, show bool, machineEventLogger machineoutput.MachineEventLoggingClient) (err error) {
+	if composite.Parallel {
+		// Loop over each command and execute it in parallel
+		commandExecs := util.NewConcurrentTasks(len(composite.Commands))
+		for _, command := range composite.Commands {
+			cmd := command // needed to prevent the lambda from capturing the value
+			if devfileCommand, ok := commandsMap[strings.ToLower(cmd)]; ok {
+				commandExecs.Add(util.ConcurrentTask{ToRun: func(errChannel chan error) {
+					err := execCommandFromComposite(client, devfileCommand, commandsMap, compInfo, show, machineEventLogger)
```
maysunfaisal, line 237: Just trying to think of all the edge cases, can we successfully exec in the same pod in parallel? Do we need mutex?

pkg/exec/devfile.go
```diff
@@ -222,6 +224,61 @@ func ExecuteDevfileDebugActionWithoutRestart(client ExecClient, exec common.Exec
...
+					}
+				}})
+			} else {
+				return fmt.Errorf("command %q not found in devfile", command)
+			}
+		}
+
+		err := commandExecs.Run()
+		if err != nil {
+			return errors.Wrap(err, "command exec failed")
```
maysunfaisal, line 249: Can we be more explicit here like, parallel execution of comp command `<name>` failed

pkg/devfile/parser/data/common/types.go
```diff
@@ -158,6 +161,27 @@ type Exec struct {
 	WorkingDir string `json:"workingDir,omitempty"`
 }
 
+// Composite command containing a list of commands to execute in a component container
+type Composite struct {
```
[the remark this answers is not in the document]
johnmcollier: Manually updated

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -133,7 +133,7 @@ func UpdateContainersWithSupervisord(devfileObj devfileParser.DevfileObj, contai
 	for i := range containers {
 		container := &containers[i]
 		// Check if the container belongs to a run command component
-		if container.Name == runCommand.Exec.Component {
+		if runCommand.Exec != nil && container.Name == runCommand.Exec.Component {
```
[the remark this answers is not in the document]
johnmcollier: Hmm, not sure why that's in there, it may be from when I initially looked at implementing composite run commands. Anyhow, it can safely be removed.

pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// GetGroup returns the group the command belongs to
+func (dc DevfileCommand) GetGroup() *Group {
+	if dc.Composite != nil {
+		return dc.Composite.Group
+	}
+
+	return dc.Exec.Group
+}
```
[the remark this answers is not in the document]
johnmcollier: Can't hurt to check

pkg/devfile/parser/data/common/types.go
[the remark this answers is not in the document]
johnmcollier: (Based on the schema)

pkg/exec/devfile.go
```diff
@@ -222,6 +224,61 @@ func ExecuteDevfileDebugActionWithoutRestart(client ExecClient, exec common.Exec
...
+		}
+
+	} else {
+		// Execute the commands in order
+		for _, command := range composite.Commands {
+			if devfileCommand, ok := commandsMap[strings.ToLower(command)]; ok {
```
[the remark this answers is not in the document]
johnmcollier: It's needed because when the command IDs are parsed by odo, they're made lower case, but because the command list in a composite command is just a simple list of strings, we need to make the conversion.

Without the ToLower, the check in the map will fail

pkg/exec/devfile.go
```diff
@@ -222,6 +224,61 @@ func ExecuteDevfileDebugActionWithoutRestart(client ExecClient, exec common.Exec
...
+		}
+	}
+
+	return nil
+}
+
+// execCommandFromComposite takes a command in a composite command and executes it.
+func execCommandFromComposite(client ExecClient, command common.DevfileCommand, commandsMap map[string]common.DevfileCommand, compInfo adaptersCommon.ComponentInfo, show bool, machineEventLogger machineoutput.MachineEventLoggingClient) (err error) {
+	if command.Composite != nil {
+		err = ExecuteCompositeDevfileAction(client, *command.Composite, commandsMap, compInfo, show, machineEventLogger)
```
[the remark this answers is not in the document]
johnmcollier: Yup! It's a bit of edge case definitely, but totally valid!

pkg/exec/devfile.go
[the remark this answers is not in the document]
johnmcollier: Sure

pkg/exec/devfile.go
johnmcollier: Yeah, we can safely exec in the same pod in parallel.

pkg/exec/devfile.go
johnmcollier: Good point

johnmcollier:
@maysunfaisal Review comments addressed 👍 . Also caught an edge case when passing in a composite command with no group failed due to nil pointer reference, so added some nil pointer checks to fix that

pkg/exec/devfile.go
johnmcollier: Granted, it may not be the best idea to have two commands exec'ing in parallel and accessing the same resource, but that's up to the devfile writer and is hard to guard against from our side.

pkg/exec/devfile_test.go
```diff
@@ -0,0 +1,202 @@
...
+			composite: common.Composite{
+				Id:       commands[2],
+				Commands: []string{commands[0], "fake-command"},
+				Parallel: true,
+			},
+			execClient: fakeExecClient,
+			wantErr:    true,
+		},
+		{
+			name: "Case 7: Nested composite commands",
```
maysunfaisal, line 170: probably add another case for nested parallel composite commands

pkg/devfile/parser/data/common/command_helper.go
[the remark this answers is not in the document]
johnmcollier: Added a nil check and corresponding test case for `GetGroup`, ended up removing `GetKind` since it wasn't needed.

pkg/exec/devfile_test.go
johnmcollier: Done 👍

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -131,6 +131,56 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(statErr).ToNot(HaveOccurred())
 		})
 
+		It("checks that odo push works with a devfile with composite commands", func() {
```
maysunfaisal, line 134: I think we should also have the other combinations in integration tests:
- composite command A referencing another composite command B
- composite command A parallel true/false and composite command B true/false

also check for a composite command failure like changing its group to run and referencing an invalid command

johnmcollier: I don't think we need (or should add that much) coverage, especially given the first 5 scenarios (your first point, then the 4 in your second) are largely the same

How about:
- composite command A, referencing another composite command. One is parallel, the other isn't
- composite command validation failure

maysunfaisal: 👍

johnmcollier: Pushed up 3 more integrations tests:
- composite command A, referencing another composite command. One is parallel, the other isn't
- composite command referencing invalid command
- composite command in run group

pkg/exec/devfile.go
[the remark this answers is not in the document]
maysunfaisal: ok, gotcha.

But thinking out loud, do we want to update it in the parser itself? So that anyone who is using it in the future dont have to do it everytime they use it like and can be done once in the parser directly? 🤔

pkg/exec/devfile_test.go
```diff
@@ -0,0 +1,226 @@
...
+			composite: common.Composite{
+				Id:       commands[3],
+				Commands: []string{commands[0], commands[2]},
+				Parallel: false,
+			},
+			execClient: fakeExecClient,
+			wantErr:    false,
+		},
+		{
+			name: "Case 7: Nested composite commands",
```
maysunfaisal, line 194: Update case number and name

pkg/exec/devfile.go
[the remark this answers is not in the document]
johnmcollier: I'm not sure there's a way to directly parse an array of strings as lower case, I'd leave as-is for now

pkg/devfile/adapters/common/command.go
```diff
@@ -39,10 +40,9 @@ func getCommandFromDevfile(data data.DevfileData, groupType common.DevfileComman
 	}
 
 	for _, command := range commands {
-		// exec.Group is a pointer, to avoid null pointer
-		if command.Exec.Group != nil && command.Exec.Group.Kind == groupType {
-			if command.Exec.Group.IsDefault {
-				// We need to scan all the commands to find default command
+		cmdGroup := command.GetGroup()
+		if cmdGroup != nil && cmdGroup.Kind == groupType {
+			if command.GetGroup().IsDefault {
```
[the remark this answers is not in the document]
johnmcollier: Good catch, fixed.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -131,6 +131,104 @@ var _ = Describe("odo devfile push command tests", func() {
...
+		})
+
+		It("should throw a validation error for composite run commands", func() {
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, cmpName)
+
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfileCompositeRun.yaml"), filepath.Join(context, "devfile.yaml"))
+
+			// Verify odo push failed
+			output := helper.CmdShouldFail("odo", "push", "--context", context)
```
maysunfaisal, line 217: in this same test case, can we also use one of the valid devfiles, and do with the flag `run-command=<composite command>`

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -131,6 +131,104 @@ var _ = Describe("odo devfile push command tests", func() {
...
+				[]string{"stat", "/projects/testfolder"},
+				func(cmdOp string, err error) bool {
+					statErr = err
+					return true
+				},
+			)
+			Expect(statErr).ToNot(HaveOccurred())
+		})
+
+		It("checks that odo push works with a devfile with parallel composite commands", func() {
```
maysunfaisal, line 159: lets use a `--build-command` flag in this test case and remove the build group from the composite command in the devfile.yaml

tests/integration/devfile/cmd_devfile_push_test.go
johnmcollier: Is this needed? The run command validation happens in the same spot, regardless of whether or not it's passed in via the flag.

maysunfaisal:
Found another bug, looks like the spinner messes up with parallel execution.

johnmcollier:
Fixed. Set `show` to false when exec'ing a parallel action:
```
 •  Executing install command "npm install"  ...
 •  Executing mkdir command "mkdir /projects/testfolder"  ...
 ✓  Executing mkdir command "mkdir /projects/testfolder" [251ms]
```

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -131,6 +131,104 @@ var _ = Describe("odo devfile push command tests", func() {
...
+			Expect(output).To(ContainSubstring("Executing mkdir command"))
+
+			// Verify the command executed successfully
+			var statErr error
+			podName := cliRunner.GetRunningPodNameByComponent(cmpName, namespace)
+			cliRunner.CheckCmdOpInRemoteDevfilePod(
+				podName,
+				"runtime",
+				namespace,
+				[]string{"stat", "/projects/testfolder"},
```
prietyc123, line 150: This will ends up with error on windows. Please use something like https://github.com/openshift/odo/blob/master/tests/integration/cmd_push_test.go#L289

prietyc123: Can you please also run your tests on windows once

johnmcollier: @prietyc123 why would that fail on Windows? That’s verifying the path of a folder in a remote Linux container, so Unix paths need to be used here.

adisky:
> Fixed. Set spinner to no spin when exec'ing a parallel action:
> 
> ```
> Executing devfile commands for component nodejs
>  •  Executing install command "npm install"  ...
>  •  Executing mkdir command "mkdir /projects/testfolder"  ...
>  ✓  Executing mkdir command "mkdir /projects/testfolder" [251ms]
> ```

It is printing two times for parallel true
```
Executing devfile commands for component check-composite
 •  Executing install command "npm install"  ...
 •  Executing echo2 command "echo hellow"  ...
 •  Executing echo3 command "echo hellohii"  ...
 •  Executing echo1 command "echo hi"  ...
 ✓  Executing echo3 command "echo hellohii" [138ms]
 ✓  Executing echo1 command "echo hi" [147ms]
 ✓  Executing echo2 command "echo hellow" [175ms]
 ✓  Executing install command "npm install" [2s]
 ✓  Executing run command "npm start" [2s]

```

adisky:
@johnmcollier tested PR, works well.
I have concern why we are allowing composite command to refer itself? it is causing execution in forever loop, and for parallel true It failed after sometime with `socket: too many open files`
`

johnmcollier:
@adisky can you share your devfile where the composure command referred itself?

I’m allowing composite commands to refer to other composite commands, but not themselves However, this doesn’t stop a composite command A from referencing a composite B that then references composite command A. This is because it would’ve been much too complicated and expensive to fully evaluate the command tree before execution starts.

adisky:
> @adisky can you share your devfile where the composure command referred itself?
> 
> I’m allowing composite commands to refer to other composite commands, but not themselves However, this doesn’t stop a composite command A from referencing a composite B that then references composite command A. This is because it would’ve been much too complicated and expensive to fully evaluate the command tree before execution starts.

```
commands:
  - exec:
      id: install
      component: runtime
      commandLine: npm install
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
  - exec:
      id: echo1
      component: runtime
      commandLine: echo hi
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
  - exec:
      id: run
      component: runtime
      commandLine: npm start
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
      group:
        kind: run
        isDefault: true
  - composite:
         id: buildAndEcho
         label: Build and Echo
         commands:
           - buildAndEcho
           - echo1
           - install
           - echo2
           - echo3
         parallel: false
         group:
           kind: build
           isDefault: true
  - exec:
      id: echo2
      component: runtime
      commandLine: echo hello
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
  - exec:
      id: echo3
      component: runtime
      commandLine: echo hellohii
      workingDir: ${CHE_PROJECTS_ROOT}/nodejs-starter
```

johnmcollier:
@adisky Thanks! I'm able to reproduce with your devfile 👍 

I've pushed a fix up that fixes it for me

pkg/devfile/adapters/common/command.go
```diff
@@ -164,6 +170,28 @@ func validateCommand(data data.DevfileData, command common.DevfileCommand) (err
...
+	if compositeCommand.Group != nil && compositeCommand.Group.Kind == common.RunCommandGroupType {
+		return fmt.Errorf("composite commands of run Kind are not supported currently")
+	}
+
+	commandsMap := GetCommandsMap(data.GetCommands())
+
+	// Loop over the commands and validate that each command points to a command that's in the devfile
+	for _, command := range compositeCommand.Commands {
+		if command == compositeCommand.Id {
+			return fmt.Errorf("the composite command %q cannot reference itself", compositeCommand.Id)
```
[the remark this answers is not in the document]
johnmcollier: Good find. It ended up being a string comparison issue with non-lower case commands (like `buildAndEcho`). This is because the `compositeCommand.Id` value will be lower case, but the value of `command` in the subcommand array may not. So we have to call `strings.ToLower()`

tests/integration/devfile/cmd_devfile_push_test.go
johnmcollier: @prietyc123 The tests I added run fine on Windows without modifciation

johnmcollier:
Had to rebase due to merge conflict

johnmcollier:
Found a simple solution for preventing indirect self-reference of composite commands, so won't be opening an issue

johnmcollier:
Output from running just the integration tests I added on Windows
```
ginkgo  -randomizeAllSpecs -slowSpecThreshold=120 -timeout 7200s -nodes=2 -focus="odo devfile push command tests" tests/integration/devfile/
[2] Running Suite: Devfile Suite
[2] ============================
[2] Random Seed: 1595271871 - Will randomize all specs
[2] Parallel test node 2/2.
[2]
[1] Running Suite: Devfile Suite
[1] ============================
[1] Random Seed: 1595271871 - Will randomize all specs
[1] Parallel test node 1/2.
[1]
[2] SS+SSS+SSSSSSSSSSSS++SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS+S
[2] Ran 5 of 59 Specs in 129.692 seconds
[2] SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 54 Skipped
[2] PASS
[1] S+SSSSSSSS+SSSSSSSSSSSSSSSSSSSSSSSSS+
[1] Ran 3 of 37 Specs in 155.610 seconds
[1] SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 34 Skipped
[1] PASS
Ginkgo ran 1 suite in 2m41.7264814s
Test Suite Passed
```

maysunfaisal:
/lgtm
thx for fixing the indirect referencing issue

adisky:
@johnmcollier looks good to me, 
can events refer to composite commands? if so PR for post start event is merged recently, either your PR needs an update or could be followup.

johnmcollier:
@adisky when I rebased yesterday I updated to support composite postStart commands 👍

pkg/devfile/parser/data/common/types.go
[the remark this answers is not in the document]
metacosm: We really should create a shared lib to handle types, de-/serialization of devfile data.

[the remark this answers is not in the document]
johnmcollier: @metacosm Agreed. There's some on-going discussion in https://github.com/openshift/odo/issues/3602 too

Issue #3547, opened 2020-07-13
odo push with --source and --config is failing more often on multi stage infrastructure


prietyc123:
/kind bug

## What versions of software are you using?

**Operating System:**
All supported

**Output of `odo version`:**
master

## How did you run odo exactly?

Running tests on openshift release repo pr https://github.com/openshift/release/pull/9431

## Actual behavior

```
[...]
Running oc with args [oc get dc nodejs-nodejs-push-context-test --namespace gcxdajfudq]
[oc] I0712 16:14:14.434158   10620 request.go:621] Throttling request took 1.026026133s, request: GET:https://api.ci-op-8lkn26zg-3c947.origin-ci-int-aws.dev.rhcloud.com:6443/apis/scheduling.k8s.io/v1beta1?timeout=32s
[oc] NAME                              REVISION   DESIRED   CURRENT   TRIGGERED BY
[oc] nodejs-nodejs-push-context-test   1          1         0         config,image(nodejs:latest)
Running odo with args [odo push --source --context /tmp/909511552]
[odo] I0712 16:14:23.368848   10650 preference.go:182] The path for preference file is /tmp/909511552/config.yaml
[odo] I0712 16:14:23.464672   10650 common_push.go:164] SourceLocation: ./
[odo] I0712 16:14:23.464703   10650 common_push.go:172] Source Path: /tmp/909511552
[odo] I0712 16:14:23.494778   10650 preference.go:182] The path for preference file is /tmp/909511552/config.yaml
[odo] Validation
[...]
[odo] run: stopped
[odo] + /opt/odo/bin/supervisord ctl start run
[odo] run: started
[odo] 
 ✓  Building component [8s]
[odo]  ✓  Changes successfully pushed to component
[odo] I0712 16:15:05.095681   10650 odo.go:72] Could not get the latest release information in time. Never mind, exiting gracefully :)
Running oc with args [oc get dc nodejs-nodejs-push-context-test --namespace gcxdajfudq]
[oc] I0712 16:15:06.584766   10677 request.go:621] Throttling request took 1.045373096s, request: GET:https://api.ci-op-8lkn26zg-3c947.origin-ci-int-aws.dev.rhcloud.com:6443/apis/storage.k8s.io/v1beta1?timeout=32s
[oc] NAME                              REVISION   DESIRED   CURRENT   TRIGGERED BY
[oc] nodejs-nodejs-push-context-test   1          1         1         config,image(nodejs:latest)
[...]

PR #3555, opened 2020-07-14
debugging push test failure on multistage test infra

prietyc123:
**What type of PR is this?**

/kind bug

This is needed to debug test failure on 4.5 multistage cluster #3547 . I just added 3 debug statement to checkout the output and may be that helps me to find out the reason for failure.

Fixes NA

NA. This is only for debug purpose. Later stage I will remove it in a clean up pr.


prietyc123:
ping @amitkrout

PR #3575, opened 2020-07-15
Proposal for transitioning from inner loop to outerloop

sbose78:
**What type of PR is this?**

/kind design

/kind documentation
[skip ci]

Proposal for supporting outerloop in odo

"As a user of a tool which understands devfiles, 
I wish to build one or more images from my source code and deploy them on a Kubernetes cluster using the outer-loop guidance from a relevant devfile."

Fixes #?

### DIFF
--- a/docs/proposals/innerloop-to-outerloop.md
+++ b/docs/proposals/innerloop-to-outerloop.md
@@ -1,0 +1,138 @@
+# Switching from inner-loop to outer-loop
+
+## Definitions
+
+
+
+### Inner loop
+The Inner loop stage is the development stage where the developer creates an application and it goes through the interation of coding, debugging, testing, locally on the workstation.
+
+### Outer loop validation
+The Outer loop validation stage is the development stage where the developer creates a platform-native build artifact ready to be deployed and distributed on Kubernetes. In case of the Kubernetes platform, a developer would build one or more images from source and deploy them as a Kubernetes Deployment, OpenShift DeploymentConfig or a Knative Service.
+
+### Outer loop Gitops
+The outer loop Gitops stage is the development stage where the developer has validated outerloop by having a built image deployed, and now wishes to deploy the app across multiple environments/stages using CI/CD
+
+
+## End-to-end user experience
+
+
+> As a user of a tool which understands devfiles, 
+I wish to build one or more images from my source code and deploy them on a Kubernetes cluster using the outer-loop guidance from a relevant devfile.
+
+
+The user validates her code in the hot-reload mode in the cluster using: 
+
+```
+odo push
+```
+
+The user validates the Kubernetes-style image-based deployment 
+
+```
+odo deploy
+```
+
+The user chooses to generate deployment manifests, which she would be using to manage Git-driven deployments across multiple stages using CI/CD pipelines
+
+
+```
+odo pipelines bootstrap
+```
+
+## Outer-loop guidance in the devfile
+
+### Build guidance
+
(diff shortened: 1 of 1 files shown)


docs/proposals/innerloop-to-outerloop.md
```diff
@@ -0,0 +1,138 @@
...
+
+```
+odo deploy
+```
+
+The user chooses to generate deployment manifests, which she would be using to manage Git-driven deployments across multiple stages using CI/CD pipelines
+
+
+```
+odo pipelines bootstrap
```
kadel, line 40: Based on description above, should this be called `odo gitops bootstrap`?

docs/proposals/innerloop-to-outerloop.md
```diff
@@ -0,0 +1,138 @@
...
+* generate [Tekton Task yaml manifests](https://github.com/tektoncd/catalog/blob/master/task/buildah/0.1/buildah.yaml) which does a Dockerfile build, or 
+* generate an equivalent [k8s-build manifest](https://github.com/redhat-developer/build/blob/master/samples/build/build_buildah_cr.yaml), orchestrated by Tekton Pipelines.
+* directly use a Tekton Task referenced in the `devfile`.
+
+#### Deploy
+Similarly, the `odo pipelines` command would generate 
+* The necessary 'yaml' manifests to deploy the built image using the guidance from the devfile.
+* The Argo CRs that represent apps that would be reconciled by Argo on a given cluster.
+
+The generated manifests may optionally be consumed outside the context of `odo` using plain old `kubectl` and `kustomize`. Overtime, the user may choose to modify them and commit them
```
kadel, line 95: If the manifests are **optionally** consumed outside of the context of `odo`. Does that mean that there would be a way to consume them in `odo`?

docs/proposals/innerloop-to-outerloop.md
```diff
@@ -0,0 +1,138 @@
...
+
+Example, the current proposal to add a devfile component called `dockerfile` looks like: 
+
+
+```
+schemaVersion: 2.1.0
+metadata:
+  name: nodejs
+  version: 1.0.0
+  alpha.deployment-manifest: "https://raw.githubusercontent.com/groeges/devfile-registry/master/devfiles/nodejs/deploy_deployment.yaml"
```
kadel, line 110: If we are talking about devfile v2.1 shouldn't there be "first-class" support for deployment manifests?

docs/proposals/innerloop-to-outerloop.md
kadel: Not entirely sure if it would make sense, but would it be possible to break this command into two parts?
One that generates generic stuff that is not application-specific like ReleBindings TektonTemplates etc..
The other that generates manifests that are specific for a given application (devfile).

docs/proposals/innerloop-to-outerloop.md
sbose78: I gotta remove this, copy-pasted by mistake. Thanks for pointing out.

docs/proposals/innerloop-to-outerloop.md
sbose78: `odo` would have the capability to parse the structure and makes changes to it, locally. Example, when someone is adding a new app to the same gitops structure.

girishramnani:
@sbose78 any update on this?

kadel:
We need to first agree on how to express outer loop in devfile. 
There is a lot of questions and uncertainty about it.
 
I'm not sure if the discussion should happen here on in https://github.com/devfile/kubernetes-api/issues/49

sbose78:
We've taken out of flow to the new CLI called Gitops Application Manager.

PR #3576, opened 2020-07-15
Changing format for the section level as expected

prietyc123:
**What type of PR is this?**

/kind documentation
[skip ci]

Changes the doc format as expected. After document title i.e level 0 we should use level 1. 

Ref : https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/

Fixes NA

👀

### DIFF
--- a/docs/public/debugging-using-devfile.adoc
+++ b/docs/public/debugging-using-devfile.adoc
@@ -52,7 +52,7 @@ commands:
 - Next we port forward a local port for debugging using `odo debug port-forward`. The default local port used for debugging is 5858. If 5858 is occupied, odo will automatically pick up a local port. We can also specify the local port using, `odo debug port-forward --local-port 5858`
 - Next we need to attach the debugger to the local port. Here's a guide to do it for VS Code : https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging[Remote Debugging]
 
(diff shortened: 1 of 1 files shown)


docs/public/debugging-using-devfile.adoc
```diff
@@ -52,7 +52,7 @@ commands:
 - Next we port forward a local port for debugging using `odo debug port-forward`. The default local port used for debugging is 5858. If 5858 is occupied, odo will automatically pick up a local port. We can also specify the local port using, `odo debug port-forward --local-port 5858`
 - Next we need to attach the debugger to the local port. Here's a guide to do it for VS Code : https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging[Remote Debugging]
 
-### Check if a debugging session is running
+## Check if a debugging session is running
```
amitkrout, line 55: @prietyc123 Can you please more details why we need this ?

prietyc123: It is pointing to some warning error like 

image

and this is just because after level0 we have directly used level2 which I have also mentioned in description.

prietyc123:
But anyways I can see this is unnecessarily changing the look and feel of doc.

image

We can ignore the warning https://github.com/openshift/odo/pull/3576#discussion_r454821493

/close

PR #3588, opened 2020-07-16
Clean ups

metacosm:
**What type of PR is this?**
/kind cleanup

Removes some redundant / unneeded code. Clean up.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -411,26 +411,24 @@ func (lci *LocalConfigInfo) writeToFile() error {
 
 // GetType returns type of component (builder image name) in the config
 func (lc *LocalConfig) GetType() string {
-	if lc.componentSettings.Type == nil {
+	return getStringOrEmpty(lc.componentSettings.Type)
+}
+
+func getStringOrEmpty(ptr *string) string {
+	if ptr == nil {
 		return ""
 	}
-	return *lc.componentSettings.Type
+	return *ptr
 }
 
 // GetSourceLocation returns the sourcelocation, returns default if nil
 func (lc *LocalConfig) GetSourceLocation() string {
-	if lc.componentSettings.SourceLocation == nil {
-		return ""
-	}
-	return *lc.componentSettings.SourceLocation
+	return getStringOrEmpty(lc.componentSettings.SourceLocation)
 }
 
 // GetRef returns the ref, returns default if nil
 func (lc *LocalConfig) GetRef() string {
-	if lc.componentSettings.Ref == nil {
-		return ""
-	}
-	return *lc.componentSettings.Ref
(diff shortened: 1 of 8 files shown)


pkg/devfile/validate/validate.go
```diff
@@ -14,16 +13,13 @@ import (
 func ValidateDevfileData(data interface{}) error {
 	var components []common.DevfileComponent
 
-	typeData := reflect.TypeOf(data)
-
-	if typeData == reflect.TypeOf(&v100.Devfile100{}) {
-		d := data.(*v100.Devfile100)
+	switch d := data.(type) {
+	case *v100.Devfile100:
```
dev-gaur, line 17: this can be adjusted in a single case clause
```suggestion
	case *v100.Devfile100, *v200.Devfile200:
	         components := d.GetComponents( )
	
```

pkg/devfile/parser/context/context.go
```diff
@@ -46,12 +44,6 @@ func (d *DevfileCtx) populateDevfile() (err error) {
 		return err
 	}
 
-	// Check if the apiVersion is supported
-	if !d.IsApiVersionSupported() {
```
mik-dass, line 50: The function `IsApiVersionSupported()` is now not used anywhere. Should we remove it?

pkg/devfile/validate/validate.go
metacosm: I tried that, this won't work because go != java  and `Devfile100` & `Devfile200` don't share a common interface (which might actually be something worth exploring provided we actually still care about devfile 1.0). I would have done this change if it was possible to do so… 😄

pkg/devfile/parser/context/context.go
metacosm: I kept it just in case we wanted to use it but I certainly can remove it.

mik-dass: Ok let's keep it for now

pkg/config/config.go
```diff
@@ -411,26 +411,24 @@ func (lci *LocalConfigInfo) writeToFile() error {
 
 // GetType returns type of component (builder image name) in the config
 func (lc *LocalConfig) GetType() string {
-	if lc.componentSettings.Type == nil {
+	return getStringOrEmpty(lc.componentSettings.Type)
+}
+
+func getStringOrEmpty(ptr *string) string {
```
mik-dass, line 417: I think we should move this to a utils packages as this can helpful at other places too like preference.go etc too.

metacosm: Done.

pkg/devfile/validate/validate.go
dev-gaur: oh I see.

my bad :sweat_smile:

PR #3597, opened 2020-07-17
Adds runMode to env for debugging with odo watch.

### LINKED ISSUE #3480: odo watch always dismisses debug mode
/kind bug

**Operating System:**
MacOS 

**Output of `odo version`:**
built from source.
odo v1.2.3 (33a73e351)
 

I am using https://github.com/gorkem/odo-devfile/tree/master/python-example to do python debugging 

```sh
odo create
odo push --debug
## At this point, application is running and I can attach my debugger
odo watch
```
Modify `requirements.py` and observe the following output 

```sh
❯ odo watch
Waiting for something to change in ...../odo-devfile/python-example
File ...../odo-devfile/python-example/requirements.txt changed
Pushing files...

Validation
 ✓  Validating the devfile [51740ns]

Creating Kubernetes resources for component python-example
 ✓  Waiting for component to start [12s]

Applying URL changes
 ✓  URL python-example-8000 successfully deleted
 ✓  URL python-example-8000: http://python-example-8000-XXXXXX.com created

Syncing to component python-example
 ✓  Syncing files to the component [382ms]

Executing devfile commands for component python-example
 ✓  Executing py-build command "pip install --user -r requirements.txt" [17s]
 ✓  Executing py-run command "python /projects/app.py" [2s] <----------------------------
Waiting for something to change in ..../odo-devfile/python-example
```
Notice that on the marked line `odo` is reverting the debug command to run command as a result of a change when I am doing a watch. I expect `odo` to preserve my last choice which is debug in this case.

mik-dass:
/kind bug

If runMode is debug, odo watch will trigger push with debug settings else it will trigger normal  push.

Fixes #3480 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

- Execute `odo push --debug`
- Execute `odo watch` and check if it detects debug mode for pushing files.
- Revert to `odo push` and check if it uses normal mode for pushing files.

### DIFF
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -31,8 +31,8 @@ type ComponentSettings struct {
 type RUNMode string
 
 const (
-	RunMODE   RUNMode = "run"
-	DebugMODE RUNMode = "debug"
+	Run   RUNMode = "run"
+	Debug RUNMode = "debug"
 )
 
 // URLKind is an enum to indicate the type of the URL i.e ingress/route
@@ -49,7 +49,7 @@ const (
 	DefaultDebugPort = 5858
 
 	// DefaultRunMode is the default run mode of the component
-	DefaultRunMode = "run"
+	DefaultRunMode = Run
 )
 
 // EnvInfoURL holds URL related information
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -23,7 +23,17 @@ type ComponentSettings struct {
 
 	// DebugPort controls the port used by the pod to run the debugging agent on
 	DebugPort *int `yaml:"DebugPort,omitempty"`
+
+	// RunMode indicates the mode of run used for a successful push
+	RunMode *RUNMode `yaml:"RunMode,omitempty"`
 }
+
+type RUNMode string
+
+const (
+	RunMODE   RUNMode = "run"
+	DebugMODE RUNMode = "debug"
+)
 
 // URLKind is an enum to indicate the type of the URL i.e ingress/route
 type URLKind string
@@ -37,6 +47,9 @@ const (
 
 	// DefaultDebugPort is the default port used for debugging on remote pod
 	DefaultDebugPort = 5858
+
+	// DefaultRunMode is the default run mode of the component
+	DefaultRunMode = "run"
 )
 
 // EnvInfoURL holds URL related information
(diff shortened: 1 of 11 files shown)


pkg/envinfo/envinfo.go
```diff
@@ -23,8 +23,18 @@ type ComponentSettings struct {
...
+	// RunMode indicates the mode of run used for a successful push
+	RunMode *RUNMode `yaml:"RunMode,omitempty"`
 }
 
+type RUNMode string
+
+const (
+	RunMODE   RUNMode = "run"
+	DebugMODE RUNMode = "debug"
+)
```
adisky, line 36: it could be simply
```
type RunMode string
const (
     Run       RunMode =  "run"
     Debug   RunMode = "run"
)
```

pkg/envinfo/envinfo.go
```diff
@@ -37,6 +47,9 @@ const (
 
 	// DefaultDebugPort is the default port used for debugging on remote pod
 	DefaultDebugPort = 5858
+
+	// DefaultRunMode is the default run mode of the component
+	DefaultRunMode = "run"
```
adisky, line 52: use the constant defined above

mik-dass:
@adisky Fixed

pkg/envinfo/fake.go
```diff
@@ -0,0 +1,7 @@
+package envinfo
+
+func GetFakeEnvInfo(settings ComponentSettings) *EnvInfo {
```
prietyc123, line 3: Can you please add comment what does this function do?

mik-dass:
@prietyc123 Fixed

pkg/envinfo/fake.go
girishramnani: we keep fake stuff in _test.go files as they are for testing perposes if I am not wrong?

girishramnani:
what about all the `--now` flag cases? would those all set the run mode to `Push`?

pkg/envinfo/fake.go
mik-dass: I tried that but doesn't work. I think we cannot use functions from _test.go in the main code. Also found this article https://stackoverflow.com/questions/59064256/is-it-possible-to-call-a-test-func-from-another-file-to-start-the-testing

mik-dass:
> what about all the --now flag cases? would those all set the run mode to Push?

Fixed :+1:

girishramnani:
> > what about all the --now flag cases? would those all set the run mode to Push?
> 
> Fixed 👍
can we have intg test for this as well?

mik-dass:
> can we have intg test for this as well?

Already added in the URL `--now` scenario.

PR #3611, opened 2020-07-21
Fix post start event logged message

adisky:
**What type of PR is this?**

 /kind bug

post start event message gets printed as pre-start and printed also when there are no events.
```
Syncing to component nodejs
 ✓  Checking files for pushing [749597ns]
 ✓  Syncing files to the component [183ms]

Executing preStart lifecycle event commands for component nodejs
```

This PR updates from `preStart` to `postStart` and prints the message only when post start event present in devfile

Fixes #

**PR acceptance criteria**:

- [ ] Unit test  No need

- [ ] Integration test No need

- [ ] Documentation  No need

run `odo push`, no event message gets printed if there are no events in devfile and correct event type is printed.

### DIFF
--- a/pkg/devfile/adapters/docker/component/adapter.go
+++ b/pkg/devfile/adapters/docker/component/adapter.go
@@ -137,9 +137,10 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 
 	// PostStart events from the devfile will only be executed when the component
 	// didn't previously exist
-	if !componentExists {
-		log.Infof("\nExecuting preStart lifecycle event commands for component %s", a.ComponentName)
-		err = a.execDevfileEvent(a.Devfile.Data.GetEvents().PostStart, containers)
+	postStartEvents := a.Devfile.Data.GetEvents().PostStart
+	if !componentExists && len(postStartEvents) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/devfile/adapters/docker/component/adapter.go
```diff
@@ -137,9 +137,10 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 
 	// PostStart events from the devfile will only be executed when the component
 	// didn't previously exist
-	if !componentExists {
-		log.Infof("\nExecuting preStart lifecycle event commands for component %s", a.ComponentName)
-		err = a.execDevfileEvent(a.Devfile.Data.GetEvents().PostStart, containers)
+	postStartEvents := a.Devfile.Data.GetEvents().PostStart
+	if !componentExists && len(postStartEvents) > 0 {
+		log.Infof("\nExecuting postStart event commands for component %s", a.ComponentName)
```
girishramnani, line 142: you could have just pushed the log into the execDevfileEvent -> after the `len(events) > 0`.
But this is ok too I guess
/approve

adisky: `execDevfileEvent` just execute events, it does not know type of event, either the function needs to be updated to receive the type of event or this way

PR #3619, opened 2020-07-22
Sync operation within the Push command does not remove deleted/renamed files

### LINKED ISSUE #3460: Sync operation within the Push command does not remove deleted/renamed files when -f flag is set
/kind bug

**Operating System:** Windows

**Output of `odo version`:** 
Built from a branch that is a couple days behind `master`.

It appears sync isn't removing deleted/renamed files inside the container if the `-f` flag is used:

Run these steps with a recent version of odo:
```
# Set a Kubernetes context
# Ensure experimental mode is enabled
git clone https://github.com/odo-devfiles/springboot-ex.git
cd springboot-ex
odo create java-springboot
odo push -v 5
mv pom.xml pom.xml.old
odo push -f -v 5      # <============ This line must be -f in order to reproduce
# maven build should fail here, but doesn't
```

In the final step, you will see that even though pom.xml no longer exists, the Maven build still completes successfully. The build should FAIL since the pom.xml should no longer exist in the container. Instead it is passing, even though`pom.xml` was renamed at the source, it was not renamed/removed on the target container.

If you look inside the container after running those steps, you will notice both pom.xml and pom.xml.old still exist:
```
root@java-spring-boot-7fddf48df-2j82m:/projects/springbootproject# ls -l
total 24
-rw-rw-rw- 1 root root 1814 Jun 30 05:08 README.md
-rw-rw-rw- 1 root root 1216 Jun 30 05:08 devfile.yaml
-rw-rw-rw- 1 root root 3011 Jun 30 05:08 pom.xml
-rw-rw-rw- 1 root root 3011 Jun 30 05:08 pom.xml.old
drwxr-xr-x 4 root root 4096 Jun 30 05:08 src
drwxr-xr-x 6 root root 4096 Jun 30 05:10 target
```

jgwest:
**What type of PR is this?**
/kind bug

- This PR ensures that if `forcePush` is set during a sync operation, that the sync directory is cleared before updated files are copied to the sync directory. 
- This PR adds a test that verifies the correct behaviour described in the original issue.

Fixes #3460

**PR acceptance criteria**:

- [ ] Unit test 

- [X] Integration test 

- [ ] Documentation 

See issue #3460 for step-by-step reproduction steps.

### DIFF
--- a/pkg/devfile/adapters/common/exec.go
+++ b/pkg/devfile/adapters/common/exec.go
@@ -42,7 +42,7 @@ func ExecuteCommand(client ExecClient, compInfo ComponentInfo, command []string,
 		// It is safe to read from cmdOutput here, as the goroutines are guaranteed to have terminated at this point.
 		klog.V(4).Infof("ExecuteCommand returned an an err: %v. for command '%v'. output: %v", err, command, cmdOutput)
 
-		return errors.Wrapf(err, "Unable to exec command %v: \n%v", command, cmdOutput)
+		return errors.Wrapf(err, "unable to exec command %v: \n%v", command, cmdOutput)
 	}
 
 	return
--- a/pkg/devfile/adapters/common/exec.go
+++ b/pkg/devfile/adapters/common/exec.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/klog"
 
 	"github.com/openshift/odo/pkg/log"
+	"github.com/pkg/errors"
 )
 
 // ExecClient  is a wrapper around ExecCMDInContainer which executes a command in a specific container of a pod.
@@ -39,9 +40,9 @@ func ExecuteCommand(client ExecClient, compInfo ComponentInfo, command []string,
 
 	if err != nil {
 		// It is safe to read from cmdOutput here, as the goroutines are guaranteed to have terminated at this point.
-		log.Errorf("\nUnable to exec command %v: \n%v", command, cmdOutput)
+		klog.V(4).Infof("ExecuteCommand returned an an err: %v. for command '%v'. output: %v", err, command, cmdOutput)
 
-		return err
+		return errors.Wrapf(err, "Unable to exec command %v: \n%v", command, cmdOutput)
 	}
 
 	return
--- a/pkg/devfile/adapters/common/exec.go
+++ b/pkg/devfile/adapters/common/exec.go
@@ -81,3 +81,40 @@ func startReaderGoroutine(reader io.Reader, show bool, cmdOutput *string, consol
 	return result
 
 }
+
+// CreateConsoleOutputWriterAndChannel is a utility function that returns a pipeWriter and a channel;
+// any strings written to that PipeWriter will be output to the channel (as lines) when the
+// writer closes. This is used to retrieve the stdout/stderr output from the container exec commands.
+//
+// The io.PipeWriter can be passed to ExecuteCommand(...) above, in order to receive the full
+// stderr/stdout output from the process.
+// See calling functions of CreateConsoleOutputWriterAndChannel for examples of usage.
+func CreateConsoleOutputWriterAndChannel() (*io.PipeWriter, chan []string) {
+	reader, writer := io.Pipe()
+
+	closeChannel := make(chan []string)
+
+	go func() {
+
+		consoleContents := []string{}
+
+		bufReader := bufio.NewReader(reader)
+		for {
+			line, _, err := bufReader.ReadLine()
+			if err != nil {
+				if err != io.EOF {
+					klog.V(4).Infof("Unexpected error on reading container output reader: %v", err)
+				}
+
+				break
+			}
+
+			consoleContents = append(consoleContents, string(line))
+		}
+
+		// Output the final console contents to the channel
+		closeChannel <- consoleContents
+	}()
+
+	return writer, closeChannel
+}
--- a/pkg/sync/adapter.go
+++ b/pkg/sync/adapter.go
@@ -109,14 +109,6 @@ func (a Adapter) SyncFiles(syncParameters common.SyncParameters) (isPushRequired
 			}
 		}
 
-		if isForcePush {
-			//reset the index
-			err = util.DeleteIndexFile(pushParameters.Path)
-			if err != nil {
-				return false, errors.Wrap(err, "unable to reset the index file")
-			}
-
-		}
 		// run the indexer and find the modified/added/deleted/renamed files
 		ret, err = util.RunIndexer(pushParameters.Path, absIgnoreRules)
 		s.End(true)
@@ -143,7 +135,7 @@ func (a Adapter) SyncFiles(syncParameters common.SyncParameters) (isPushRequired
 		changedFiles = filesChangedFiltered
 		klog.V(4).Infof("List of files changed: +%v", changedFiles)
 
-		if len(filesChangedFiltered) == 0 && len(filesDeletedFiltered) == 0 {
+		if len(filesChangedFiltered) == 0 && len(filesDeletedFiltered) == 0 && !isForcePush {
 			// no file was modified/added/deleted/renamed, thus return without synching files
 			log.Success("No file changes detected, skipping build. Use the '-f' flag to force the build.")
 			return false, nil
@@ -213,41 +205,6 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 			// nothing to push
 			s.End(true)
 			return nil
-		}
-	}
-
-	if isForcePush {
-		// If isForcePush is set, then the entire local directory contents will be copied over by CopyFile(...), thus
-		// we first need to clear the remote target folder.
-
-		err = common.ExecuteCommand(a.Client, compInfo, []string{"rm", "-rf", syncFolder}, false, nil, nil)
-		if err != nil {
-			// This command will return a non-zero error code if 'syncFolder' cannot be removed; this occurs
-			// if the folder is owned by a different container user (for example, if it is a source volume mount) or
-			// if we do not have sufficient permissions to remove it.
-			//
-			// However, for this function, we do not want 'syncFolder' itself to be removed, we just want the
-			// contents inside of it removed.
-			//
-			// So, to verify that the files inside that directory are actually removed, we will call
-			// isDirectoryEmpty below.
-			klog.V(4).Infof("error on deleting sync folder, but this is most likely an expected error. error: %v", err)
-		}
-
-		// (Re)create the folder
-		err = common.ExecuteCommand(a.Client, compInfo, getCmdToCreateSyncFolder(syncFolder), false, nil, nil)
-		if err != nil {
-			return err
-		}
-
-		// The folder contents should be empty after we have cleared it; return an error if it's not
-		if isEmpty, err := isRemoteFolderEmpty(a, compInfo, syncFolder); err != nil || !isEmpty {
-
-			if err != nil {
-				return err
-			}
-			return errors.Errorf("remote directory '%s' should be empty, but is not", syncFolder)
-
 		}
 	}
 
--- a/pkg/sync/adapter.go
+++ b/pkg/sync/adapter.go
@@ -215,6 +215,33 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 		}
 	}
 
+	if isForcePush {
+		if syncFolder != kclient.OdoSourceVolumeMount {
+			err = exec.ExecuteCommand(a.Client, compInfo, []string{"rm", "-rf", syncFolder}, true, nil, nil)
+			if err != nil {
+				return err
+			}
+			err = exec.ExecuteCommand(a.Client, compInfo, []string{"mkdir", "-p", syncFolder}, false, nil, nil)
+			if err != nil {
+				return err
+			}
+
+		} else {
+
+			cmdRmFiles := []string{"sh", "-c", "rm -rf '" + syncFolder + "'/*"}
+			err = exec.ExecuteCommand(a.Client, compInfo, cmdRmFiles, false, nil, nil)
+			if err != nil {
+				klog.V(4).Infof("error on deleting sync folder, but this is most likely an expected error. cmd: %v  error: %v", cmdRmFiles, err)
+			}
+
+			cmdRmDotFiles := []string{"sh", "-c", "rm -rf '" + syncFolder + "'/.*"}
+			err = exec.ExecuteCommand(a.Client, compInfo, cmdRmDotFiles, false, nil, nil)
+			if err != nil {
+				klog.V(4).Infof("error on deleting sync folder, but this is most likely an expected error. cmd: %v  error: %v", cmdRmFiles, err)
+			}
+		}
+	}
+
 	if isForcePush || len(files) > 0 {
 		klog.V(4).Infof("Copying files %s to pod", strings.Join(files, " "))
 		err = CopyFile(a.Client, path, compInfo, syncFolder, files, globExps)
--- a/pkg/sync/adapter.go
+++ b/pkg/sync/adapter.go
@@ -220,7 +220,7 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 		// If isForcePush is set, then the entire local directory contents will be copied over by CopyFile(...), thus
 		// we first need to clear the remote target folder.
 
-		err = exec.ExecuteCommand(a.Client, compInfo, []string{"rm", "-rf", syncFolder}, true, nil, nil)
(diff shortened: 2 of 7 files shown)


pkg/sync/adapter.go
```diff
@@ -197,6 +197,35 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
...
+		// 	err = exec.ExecuteCommand(a.Client, compInfo, cmdRmFiles, false, nil, nil)
+		// 	if err != nil {
+		// 		klog.V(4).Infof("error on deleting sync folder, but this is most likely an expected error. cmd: %v  error: %v", cmdRmFiles, err)
+		// 	}
+
+		// 	cmdRmDotFiles := []string{"sh", "-c", "rm -rf '" + syncFolder + "'/.*"}
+		// 	err = exec.ExecuteCommand(a.Client, compInfo, cmdRmDotFiles, false, nil, nil)
+		// 	if err != nil {
+		// 		klog.V(4).Infof("error on deleting sync folder, but this is most likely an expected error. cmd: %v  error: %v", cmdRmFiles, err)
+		// 	}
```
mik-dass, line 225: commented code

pkg/sync/adapter.go
```diff
@@ -197,6 +197,35 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 		}
 	}
 
+	if isForcePush {
+		// if syncFolder != kclient.OdoSourceVolumeMount {
```
mik-dass, line 201: comment found

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -462,4 +465,39 @@ var _ = Describe("odo devfile push command tests", func() {
 		})
 	})
 
+	Context("Verify devfile push works", func() {
+
+		It("should mooooooo", func() {
```
mik-dass, line 470: Better description please :D

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -462,4 +465,39 @@ var _ = Describe("odo devfile push command tests", func() {
...
+			helper.CmdShouldPass("odo", "create", "java-springboot", "--project", namespace, cmpName)
+			helper.CopyExample(filepath.Join("source", "devfiles", "springboot", "project"), context)
+
+			output := helper.CmdShouldPass("odo", "push", "--namespace", namespace)
+			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
+
+			err := os.Rename(filepath.Join(context, "pom.xml"), filepath.Join(context, "pom.xml.renamed"))
+			Expect(err).NotTo(HaveOccurred())
+
+			// session := helper.CmdRunner("odo", "push", "-v", "5", "--namespace", namespace)
```
mik-dass, line 481: comment found

jgwest:
Thanks for reviewing @mik-dass! When I posted the initial PR I had been up late last night working on this; I hit push on my commits (such as they were), and then went immediately to bed 😄 . In any case, I've cleaned them up and updated the PR description above.

dharmit:
@jgwest what I'm observing is different from what's mentioned in #3460. Upon doing `mv pom.xml pom.xml.old` and `odo push -f`, I don't see the `pom.xml` any more in the pod.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -552,3 +574,13 @@ var _ = Describe("odo devfile push command tests", func() {
...
+
+// Wait for the session stdout output to contain a particular string
+func waitForOutputToContain(substring string, session *gexec.Session) {
+
+	Eventually(func() string {
+		contents := string(session.Out.Contents())
+		return contents
+	}, 180, 10).Should(ContainSubstring(substring))
+
+}
```
mik-dass, line 586: this should be in the /tests/helper package

pkg/sync/adapter.go
```diff
@@ -197,6 +197,23 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
...
 	}
 
+	if isForcePush {
+		// If isForcePush is set, then the entire local directory contents will be copied over by CopyFile(...), thus
+		// we first need to clear the remote target folder.
+
+		err = exec.ExecuteCommand(a.Client, compInfo, []string{"rm", "-rf", syncFolder}, true, nil, nil)
+		if err != nil {
+			// This command may fail if the parent directory permissions are restrictive; since this doesn't
+			// affect our ability to delete the files _within_ the directory, we merely log it here.
```
mik-dass, line 207: >  ability to delete the files _within_ the directory

can you please elaborate this line?

mik-dass:
> Thanks for reviewing @mik-dass! When I posted the initial PR I had been up late last night working on this; I hit push on my commits (such as they were), and then went immediately to bed smile . In any case, I've cleaned them up and updated the PR description above and everything is ready for review.

No problem :D  The code seems to work for me :+1: I have left some commands. Please have a look.

pkg/sync/adapter.go
```diff
@@ -197,6 +197,23 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
 		}
 	}
 
+	if isForcePush {
+		// If isForcePush is set, then the entire local directory contents will be copied over by CopyFile(...), thus
+		// we first need to clear the remote target folder.
+
+		err = exec.ExecuteCommand(a.Client, compInfo, []string{"rm", "-rf", syncFolder}, true, nil, nil)
```
mik-dass, line 204: maybe rm and mkdir in one call?

jgwest:
Hi @dharmit, interesting... I ran the [reproduction steps](https://github.com/openshift/odo/issues/3460.) against latest `master` and can definitely still reproduce, here's a gist of how it looks:
https://gist.github.com/jgwest/8dca5875a4a5cf034da2bf5a8986ba5b
(In this log I reset branch to `origin/master`, run `git logs` to show the commit log to at `master`, rebuild, then follow the steps of issue https://github.com/openshift/odo/issues/3460.)

At line 1581. you can see that the call to `odo push -f -v 5` is a passing build, when it should be a failing build. 

If you have any more suggestions on how I can reproduce the behaviour you are seeing, let me know!

(Ran on commit b3cf59550634294a49cf5a91124031c4ae9cebcb, on Linux, against generic Kubernetes)

tests/integration/devfile/cmd_devfile_push_test.go
jgwest: Done.

pkg/sync/adapter.go
```diff
@@ -210,6 +245,34 @@ func (a Adapter) pushLocal(path string, files []string, delFiles []string, isFor
...
 }
 
+// isRemoteFolderEmpty returns true if the 'syncFolder' path in the pod contains no files/folders, false otherwise
+func isRemoteFolderEmpty(a Adapter, compInfo common.ComponentInfo, syncFolder string) (bool, error) {
+
+	stdoutWriter, stdoutOutputChan := exec.CreateConsoleOutputWriterAndChannel()
+	stderrWriter, stderrOutputChan := exec.CreateConsoleOutputWriterAndChannel()
+
+	err := exec.ExecuteCommand(a.Client, compInfo, []string{"ls", "-a", syncFolder}, false, stdoutWriter, stderrWriter)
+	if err != nil {
```
amitkrout, line 255: Can be done in a more talkative way and more importantly we can make err scope at block level instead of the function level. For example

```
if err := exec.ExecuteCommand(
		a.Client,
		compInfo,
		[]string{"ls", "-a", syncFolder},
		false,
		stdoutWriter,
		stderrWriter,
); err != nil {
		return false, err
}
```

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -409,6 +409,27 @@ var _ = Describe("odo devfile push command tests", func() {
...
+			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
+
+			// 2) Rename the pom.xml, which should cause the build to fail if sync is working as expected
+			err := os.Rename(filepath.Join(context, "pom.xml"), filepath.Join(context, "pom.xml.renamed"))
+			Expect(err).NotTo(HaveOccurred())
+
+			// 3) Ensure that the build fails due to missing 'pom.xml', which ensures that the sync operation
+			// correctly renamed pom.xml to pom.xml.renamed.
+			session := helper.CmdRunner("odo", "push", "-v", "5", "-f", "--namespace", namespace)
+			helper.WaitForOutputToContain("Non-readable POM", 180, 10, session)
```
amitkrout, line 429: The wait is done on ```session.Out.Contents()``` however i was expecting the wait should be done on ```session.Err.Contents()``` instead as it fails. Is it by design ```session := helper.CmdRunner("odo", "push", "-v", "5", "-f", "--namespace", namespace)``` always returns to stdout ?

jgwest: Hi @amitkrout , yes this is by design: the `Non-readable POM` message is coming from `mvn` command executed from  [here](https://github.com/openshift/odo/blob/3f3d8b970e9fa3693e319dd27a6adeb84a5ac1bc/pkg/exec/exec.go#L64) which is stdout-only.

pkg/sync/adapter.go
jgwest: Hi @mik-dass, they are intentionally kept separate as the `rm -rf` may fail (which is fine), but the `mkdir` should not fail. 

I've updated that logic in the function since you reviewed it, take a look and see what you think. It is now:
1) Do `rm -rf (sync folder)` to clear out the directory (log any errors, but continue to step 2)
2) Do `mkdir -p (sync folder)` to ensure the directory exists (this should never fail, so err is fatal to this function)
3) Verify that the `(sync folder)` exists, and contains no files: this checks that both step 1 and step 2 worked as expected.

pkg/sync/adapter.go
jgwest: Done.

pkg/sync/adapter.go
jgwest: I've changed the logic and updated the comment, let me know what you think now, [see my comment here](https://github.com/openshift/odo/pull/3619#discussion_r460817847).

tests/integration/devfile/cmd_devfile_push_test.go
amitkrout: Cool!!!
Thanks

dharmit:
@jgwest maven build indeed passes for me (but it shold fail, as you mentioned, since there's no `pom.xml`). But what's different from what you're observing is that I don't see `pom.xml` inside the container:
```sh
$ kubectl exec `kubectl get pods -o name | grep springboot` -c tools -- sh -c "ls -al /projects/springbootproject"
total 16
drwxr-sr-x. 4 1000570000 1000570000  105 Jul 28 12:12 .
drwxrwsrwx. 3 root       1000570000   31 Jul 28 12:03 ..
-rw-r--r--. 1 1000570000 1000570000   74 Jul  3 11:35 .gitignore
-rw-r--r--. 1 1000570000 1000570000 1814 Jul  3 11:31 README.md
-rw-r--r--. 1 1000570000 1000570000 1241 Jul 23 06:51 devfile.yaml
-rw-r--r--. 1 1000570000 1000570000 3011 Jul  3 11:31 pom.xml.old
drwxr-sr-x. 4 1000570000 1000570000   30 Jul 28 12:03 src
drwxr-sr-x. 6 1000570000 1000570000  163 Jul 28 12:07 target
```

jgwest:
@dharmit 
Using `master` I've got both, after running the reproduction steps, I've got both:
```
root@java-springboot-7f4c9bb6cf-v6xlf:/projects/springbootproject# ls -l
total 24
-rw-rw-rw- 1 root root 1814 Jul 28 15:59 README.md
-rw-rw-rw- 1 root root 1241 Jul 28 15:59 devfile.yaml
-rw-rw-rw- 1 root root 3011 Jul 28 15:59 pom.xml
-rw-rw-rw- 1 root root 3011 Jul 28 15:59 pom.xml.old
drwxr-xr-x 4 root root 4096 Jul 28 16:01 src
drwxr-xr-x 6 root root 4096 Jul 28 16:04 target
root@java-springboot-7f4c9bb6cf-v6xlf:/projects/springbootproject#
```
I wouldn't think it would be possible for the Maven build to pass if the original `pom.xml` didn't exist.

dharmit:
> I wouldn't think it would be possible for the Maven build to pass if the original `pom.xml` didn't exist.

Exactly. I wouldn't think so either. But it did for me when I was testing against OCP. I've been trying to reproduce this since a couple of days but `odo push` errors out with:
```
 •  Waiting for component to start  ...
I0731 06:24:55.573530   16901 deployments.go:89] Waiting for deployment "java-springboot" rollout to finish: 0 of 1 updated replicas are available...
I0731 06:24:55.573549   16901 deployments.go:96] Waiting for deployment spec update to be observed...
 ✗  Waiting for component to start [5m]
 ✗  Failed to start component with name java-springboot. Error: Failed to create the component: error while waiting for deployment rollout: timeout while waiting for java-springboot deployment roll out
```

I'm gonna let this one pass since functionality has already been approved.

pkg/exec/exec.go
```diff
@@ -82,3 +82,40 @@ func startReaderGoroutine(reader io.Reader, show bool, cmdOutput *string, consol
...
 }
+
+// CreateConsoleOutputWriterAndChannel is a utility function that returns a pipeWriter and a channel;
+// any strings written to that PipeWriter will be output to the channel (as lines) when the
+// writer closes. This is used to retrieve the stdout/stderr output from the container exec commands.
+//
+// The io.PipeWriter can be passed to ExecuteCommand(...) above, in order to receive the full
+// stderr/stdout output from the process.
+// See calling functions of CreateConsoleOutputWriterAndChannel for examples of usage.
+func CreateConsoleOutputWriterAndChannel() (*io.PipeWriter, chan []string) {
```
cdrage, line 93: Sorry but we should have a unit test for this command :(

maysunfaisal:
@jgwest messaged you privately but posting here for discussion I guess. I am seeing errors with rm -rf project dir:
```
Maysuns-MacBook-Pro:nodejs-ex maysun$ odo push -f

Validation
 ✓  Validating the devfile [94028ns]

Creating Kubernetes resources for component madridnode
 ✓  Waiting for component to start [102ms]

Applying URL changes

Syncing to component madridnode
 ✓  Checking file changes for pushing [9ms]
 ◐  Syncing files to the component ✗  
Unable to exec command [rm -rf /projects]: 
rm: cannot remove '/projects': Permission denied

 ✓  Syncing files to the component [5s]

Executing devfile commands for component madridnode
 ✓  Executing devbuild command "npm install" [12s]
 ✓  Executing devrun command "npm start" [4s]

Pushing devfile component madridnode
```

when i exec into the container to try it out manually
```
bash-4.4$ whoami
whoami: cannot find name for user ID 10001
bash-4.4$ ls -la /projects
total 0
drwxrwxrwx. 2 root root  6 Aug  6 20:17 .
drwxr-xr-x. 1 root root 29 Aug  6 20:11 ..
bash-4.4$ rm -rf /projects
rm: cannot remove '/projects': Permission denied
```

I am on OCP 4.3 which is why i am user `10001` since Openshift launches container with non-root as userid

However, the other thing i communicated with you was about removing files not mentioned in `.odo/odo-file-index.json` like the use case where a preStart event may write to the mount dir of the container via an init container. This file operations are not in the index json file and when I do `odo push -f` with this PR, the files are deleted. Any thoughts? 🤔

pkg/exec/exec.go
jgwest: Done.

pkg/devfile/adapters/common/exec.go
```diff
@@ -39,9 +40,9 @@ func ExecuteCommand(client ExecClient, compInfo ComponentInfo, command []string,
 
 	if err != nil {
 		// It is safe to read from cmdOutput here, as the goroutines are guaranteed to have terminated at this point.
-		log.Errorf("\nUnable to exec command %v: \n%v", command, cmdOutput)
+		klog.V(4).Infof("ExecuteCommand returned an an err: %v. for command '%v'. output: %v", err, command, cmdOutput)
 
-		return err
+		return errors.Wrapf(err, "Unable to exec command %v: \n%v", command, cmdOutput)
```
cdrage, line 45: nitpick, but errors should not be capitalized or use newlines.

cdrage: See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

jgwest: @cdrage : Done... nitpicking is always good, thanks for continuing to keep us honest on good error format 😃

Issue #3385, opened 2020-06-18
Improve the test run time on PRs


girishramnani:
- we are considering running all tests on later stages of the PRs
- we are considering running the travis CI and latest version openshift tests in the PR and rest in the periodic job
- ~~we are considering running the e2e tests ( not integration tests ) in the periodic jobs and omit them from the PR tests altogether~~ - this was done in https://github.com/openshift/odo/issues/2931

girishramnani:
cc @metacosm

metacosm:
This should be worked on as quickly as possible, in my opinion as it would help speed up the rest of the development.

amitkrout:
> we are considering running all tests on later stages of the PRs

@metacosm This deserves a long description, so that we can analyze the requirement and ofcourse we need to check it (if complicated) with test platform team as our pr run is controlled by the prow bot.

> we are considering running the travis CI and latest version openshift tests in the PR and rest in the periodic job

Just want to understand, what is the motive here ? Test time optimization ? it won't help much because CI takes almost the same time for spinning up and running test against each version of cluster. If we are looking for AWS resource optimization then yes, number of AWS resource will be reduced.

> we are considering running the e2e tests ( not integration tests ) in the periodic jobs and omit them from the PR tests altogether

+1, but we will keep both test type until we make sure that our test does not hit flakes.

metacosm:
The motivation is to move faster PRs faster from submitted to merged as well as reduce resources usage. 

Right now, lots of time/resources is spent on running tests that are not needed: integration tests should only be run when a PR is approved and considered ready to be merged by reviewers. There is no point in running all the tests while the review process is not finished and only unit tests / smoke tests should be run during that period. 

This would have the following benefits:
- reduce resource usage by not running tests that are not needed
- free resources for PRs that require these tests to be run (i.e. the ones that we attempt to merge)
- reduce the noise in the PR review loop by removing test output that no one cares about
- hopefully keep people engaged in PRs instead of letting them unattended for too long because some tests are failing 

Another aspect of this is obviously the unbelievably high occurrence of flakes, which should also be addressed. If a test cannot reliably run then it is useless because if it fails, you don't know if it's positive signal of an issue with the code and if it passes, you're not sure it somehow worked when maybe it shouldn't have…

amitkrout:
@metacosm  So basically we are looking for flow like

[...]

PR #3625, opened 2020-07-23
Clean up and customisation for delete tests

prietyc123:
**What type of PR is this?**

/kind cleanup
/kind code-refactoring

Customising the existing tests and clean up unnecessary steps for devfile as well as regular delete test. 

Fixes - Part of #3385

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

`make test-cmd-devfile-delete` should run successfully.

### DIFF
--- a/tests/helper/helper_kubectl.go
+++ b/tests/helper/helper_kubectl.go
@@ -145,14 +145,3 @@ func (kubectl KubectlRunner) GetEnvsDevFileDeployment(componentName string, proj
 	}
 	return mapOutput
 }
-
-// DeleteLocalConfig helps user to delete local config files with flags
-func (kubectl KubectlRunner) DeleteLocalConfig(args ...string) {
-	CmdShouldFail("odo", args...)
-	output := CmdShouldPass("odo", append(args, "-af")...)
-	expectedOutput := []string{
-		"Successfully deleted env file",
-		"Successfully deleted devfile.yaml file",
-	}
-	MatchAllInOutput(output, expectedOutput)
-}
--- a/tests/helper/helper_kubectl.go
+++ b/tests/helper/helper_kubectl.go
@@ -145,3 +145,14 @@ func (kubectl KubectlRunner) GetEnvsDevFileDeployment(componentName string, proj
 	}
 	return mapOutput
 }
+
+// DeleteLocalConfig helps user to delete local config files with flags
+func (kubectl KubectlRunner) DeleteLocalConfig(args ...string) {
+	CmdShouldFail("odo", args...)
+	output := CmdShouldPass("odo", append(args, "-af")...)
+	expectedOutput := []string{
+		"Successfully deleted env file",
+		"Successfully deleted devfile.yaml file",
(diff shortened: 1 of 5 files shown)


amitkrout:
Awesome!!!
/approve

prietyc123:
Prow failure

/retest

tests/integration/devfile/cmd_devfile_delete_test.go
```diff
@@ -12,8 +11,19 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+// deleteLocalConfig helps user to delete local config files with flags
+func deleteLocalConfig(args ...string) {
+	helper.CmdShouldFail("odo", args...)
```
dharmit, line 16: It's a bit unclear as to why this should fail. To get the context, I scrolled down to where the function is called. IMO, it makes more sense to limit the scope of this function to the context where it's being called. That would still get the task done and avoid any confusion for someone who's looking at using this later.

prietyc123: > It's a bit unclear as to why this should fail.

oops... my bad. I should have just add a single comment so that it doesn't confuse everyone.

> To get the context, I scrolled down to where the function is called. IMO, it makes more sense to limit the scope of this function to the context where it's being called. That would still get the task done and avoid any confusion for someone who's looking at using this later.

Yes, I agree the earlier format will also get the task done. But from cleanup and customisation point of view and keeping the context also in mind, I would say this approach would be better if we add a single line of comment why the `CmdShouldFail` or Pass. And going forward someone will just use this function according to their need. Also this function has reduced number of lines from the test scripts.

prietyc123: BTW this function ```deleteLocalConfig``` validate both pass and fail condition in a single function block.

amitkrout: @prietyc123 Can you please move it to helper package for re usability purpose

tests/helper/helper_kubectl.go
```diff
@@ -145,3 +145,14 @@ func (kubectl KubectlRunner) GetEnvsDevFileDeployment(componentName string, proj
...
+// DeleteLocalConfig helps user to delete local config files with flags
+func (kubectl KubectlRunner) DeleteLocalConfig(args ...string) {
+	CmdShouldFail("odo", args...)
+	output := CmdShouldPass("odo", append(args, "-af")...)
+	expectedOutput := []string{
+		"Successfully deleted env file",
+		"Successfully deleted devfile.yaml file",
+	}
+	MatchAllInOutput(output, expectedOutput)
+}
```
mik-dass, line 158: This is a odo specific function and the oc runner version is exactly same as this one. This should be in the helper file.

prietyc123: The best appropriate location I could look for is `utils` file. So moved it there.

PR #3627, opened 2020-07-23
Update devfile schema

adisky:
**What type of PR is this?**

/kind feature

Update devfile schema to unblock some issues.

New in Schema: Apply, endpoint changes, starter project, description of volume mounts are new changes in the schema.

This PR updates the schema and go structs here pkg/devfile/parser/data/2.0.0/types.go, we are not using go structs from here Directly we consume from pkg/devfile/parser/data/common/types.go, issue stakeholders can update corresponding struct  accordingly.

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Tests should pass

### DIFF
--- a/pkg/devfile/parser/data/2.0.0/types.go
+++ b/pkg/devfile/parser/data/2.0.0/types.go
@@ -185,8 +185,8 @@ type Endpoint struct {
 	Protocol string `json:"protocol,omitempty"`
 
 	// Describes whether the endpoint should be secured and protected by some authentication process
-	Secure     bool `json:"secure,omitempty"`
-	TargetPort int  `json:"targetPort,omitempty"`
+	Secure     bool  `json:"secure,omitempty"`
+	TargetPort int32 `json:"targetPort,omitempty"`
 }
 
 // Env is the key value pair representing an Environment variable
--- a/pkg/devfile/parser/data/2.0.0/types.go
+++ b/pkg/devfile/parser/data/2.0.0/types.go
@@ -185,8 +185,8 @@ type Endpoint struct {
 	Protocol string `json:"protocol,omitempty"`
 
 	// Describes whether the endpoint should be secured and protected by some authentication process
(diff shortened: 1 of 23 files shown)


pkg/devfile/parser/data/2.0.0/types.go
```diff
@@ -138,10 +153,40 @@ type Container struct {
...
+	// - `wss`: Endpoint will have `wss` traffic, typically on a TCP connection.
+	// - `tcp`: Endpoint will have traffic on a TCP connection, without specifying an application protocol.
+	// - `udp`: Endpoint will have traffic on an UDP connection, without specifying an application protocol.
+	//
+	// Default value is `http`
+	Protocol string `json:"protocol,omitempty"`
+
+	// Describes whether the endpoint should be secured and protected by some authentication process
+	Secure     bool `json:"secure,omitempty"`
+	TargetPort int  `json:"targetPort,omitempty"`
```
yangcao77, line 189: The `TargetPort` is required in the definition: https://devfile.github.io/devfile/_attachments/api-reference.html
The omitempty should be removed

yangcao77: Also the type should be `int32`, since the HostPort and ContainerPort defined in Kube Core v1 are all type of `int32`. https://godoc.org/k8s.io/api/core/v1#ContainerPort

adisky: @yangcao77 updated port to int32,
 https://golang.org/pkg/encoding/json/#Marshal these tags are used while marshaling the data, it is just there to skip writing targetPort while converting it to json/yaml
if you do not specify targetPort in your devfile, schema validation would fail as it is `required`. you can test yourself

adisky:
Disabled parent devfile tests 
https://github.com/openshift/odo/issues/3638

Issue #3581, opened 2020-07-15
Support PROJECTS_ROOT env variable in odo


johnmcollier:
/kind user-story

## User Story
As a user, I want to be able to configure the sync folder using the `$PROJECTS_ROOT` environment variable in my terminal, as per the Devfile 2.0 spec. 

## Acceptance Criteria 
- [ ] It should replace the existing `$CHE_PROJECTS_ROOT` variable used in Odo, and function the same.
- [ ] The sample odo devfiles in github.com/odo-devfiles should be updated accordingly
- [ ] Odo's test automation should be updated accordingly

## Links 
- Related issue: https://github.com/openshift/odo/issues/3494

/kind user-story
/area devfile

johnmcollier:
As a side note, I'll be handling this with the other changes in https://github.com/openshift/odo/issues/3494, but thought I'd open a separate story since this change is not directly related to the changes in https://github.com/openshift/odo/issues/3494.

elsony:
This one is done on Sprint 187

Issue #3494, opened 2020-07-06
Make the sync directory more consistent


johnmcollier:
Currently, odo uses the following logic for determining what folder to sync to:
1) If a single project is defined in the devfile, it syncs to `/projects/<projectName>`
2) If no projects are defined, or if there are multiple projects defined, it syncs to `/projects`
3) (Most recently) If the `sourceMapping` field is set, it syncs to the folder specified.

The logic for 1 and 2 were in place because we wanted to maintain compatibility with Che devfiles initially, a number of which relied on the assumption that the code for a component would be under `/projects/<projectName>`. However, as it turned out, there would be a number of blockers that would prevent Che devfiles from working with odo. Additionally, this was before we had implemented support for source mapping, allowing the user to configure where the source gets synced to.

The downside to this current approach is that the expected sync directory isn't consistent for somebody writing a devfile, and as they add or remove projects from the devfile, the sync directory will change, potentially breaking the commands in their devfile each time. 

@elsony @maysunfaisal and I chatted last week about this, and decided that long term, it would be better if odo always synced to `/projects`, unless the sourceMapping field is set, in which case, it will sync to the sourceMapping folder. This will ensure the sync directory is much more consistent and not subject to unwritten rules in odo.

kadel:
> @elsony @maysunfaisal and I chatted last week about this, and decided that long term, it would be better if odo always synced to `/projects`, unless the sourceMapping field is set, in which case, it will sync to the sourceMapping folder. This will ensure the sync directory is much more consistent and not subject to unwritten rules in odo.

+1 with one exception. If `sourceMapping` is not set it should check for `PROJECTS_ROOT` env variable and use that, and only if the `PROJECTS_ROOT` is not set then use `/projects`

This is my understanding based on how it is defined here: https://github.com/devfile/kubernetes-api/blob/80da7f73abaf7572c2bfb5c6dd1c1a696d2380b1/pkg/apis/workspaces/v1alpha1/containerComponent.go#L42-L46

maysunfaisal:
I think we should also consider the possibility of `clonePath` in projects.. If we're going to eventually support this property, we should take it into consideration with this issue 

https://devfile.github.io/devfile/_attachments/api-reference.html

PR #3630, opened 2020-07-23
Add proposal for updating the file sync location

johnmcollier:

**What type of PR is this?**
/kind design

Adds a design proposal doc for modifying the file sync location functionality in odo, as we decided on last Thursday's meeting to do that first. 

See https://github.com/openshift/odo/issues/3494 and https://github.com/openshift/odo/issues/3581 for some background

N/A 

N/A

### DIFF
--- a/docs/proposals/file-sync-location.md
+++ b/docs/proposals/file-sync-location.md
@@ -8,7 +8,7 @@ Currently, odo uses the following logic to determine where to sync source code t
 1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there
 
 2) If `sourceMapping` is unset, odo will do one of two things:
-   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>
+   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>`
    b) If no projects are defined, or there are multiple projects defined, odo will sync to `/projects`
 
 The downside to this current approach is that the expected sync directory isn't consistent for somebody writing a devfile, and as they add or remove projects from the devfile, the sync directory will change, potentially breaking the commands in their devfile each time.
--- a/docs/proposals/file-sync-location.md
+++ b/docs/proposals/file-sync-location.md
@@ -1,0 +1,26 @@
+
+# File Sync Location for Devfile Components in Odo
+
+## Background 
+
+Currently, odo uses the following logic to determine where to sync source code to a devfile component:
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there
+
+2) If `sourceMapping` is unset, odo will do one of two things:
+   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>
+   b) If no projects are defined, or there are multiple projects defined, odo will sync to `/projects`
+
+The downside to this current approach is that the expected sync directory isn't consistent for somebody writing a devfile, and as they add or remove projects from the devfile, the sync directory will change, potentially breaking the commands in their devfile each time.
+
+## Implementation plan
+
+To make the syncing directory more consistent, odo will do the following (in the order given):
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there. The `PROJECTS_ROOT` environment variable in each container will be set to this value (as-is today).
+
+2) If the `PROJECTS_ROOT` environment variable is set on the user's machine, odo will sync to there, *and* set it as an environment variable in each remore container
+
+3) If neither `sourceMapping` or `PROJECTS_ROOT` are set, odo will mount and sync the project source code to `/projects`
+
+The devfiles in [github.com/odo-devfiles](github.com/odo-devfiles) will also need to be updated, as each devfile assumes the source code was synced to `/projects/<projectName>`. Additionally, any test devfiles that rely on the old sync logic will also need to be updated.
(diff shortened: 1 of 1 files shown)


docs/proposals/file-sync-location.md
```diff
@@ -0,0 +1,26 @@
...
+
+The downside to this current approach is that the expected sync directory isn't consistent for somebody writing a devfile, and as they add or remove projects from the devfile, the sync directory will change, potentially breaking the commands in their devfile each time.
+
+## Implementation plan
+
+To make the syncing directory more consistent, odo will do the following (in the order given):
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there. The `PROJECTS_ROOT` environment variable in each container will be set to this value (as-is today).
+
+2) If the `PROJECTS_ROOT` environment variable is set on the user's machine, odo will sync to there, *and* set it as an environment variable in each remore container
```
dharmit, line 22: It checks if the environment variable is set on user's development system? Wouldn't it ease things if we provide an interface to set this through `odo env` command instead?

docs/proposals/file-sync-location.md
```diff
@@ -0,0 +1,26 @@
...
+# File Sync Location for Devfile Components in Odo
+
+## Background 
+
+Currently, odo uses the following logic to determine where to sync source code to a devfile component:
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there
+
+2) If `sourceMapping` is unset, odo will do one of two things:
+   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>
```
dharmit, line 11:
```suggestion
   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>`
```

docs/proposals/file-sync-location.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Implementation plan
+
+To make the syncing directory more consistent, odo will do the following (in the order given):
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there. The `PROJECTS_ROOT` environment variable in each container will be set to this value (as-is today).
+
+2) If the `PROJECTS_ROOT` environment variable is set on the user's machine, odo will sync to there, *and* set it as an environment variable in each remore container
+
+3) If neither `sourceMapping` or `PROJECTS_ROOT` are set, odo will mount and sync the project source code to `/projects`
```
dharmit, line 24: What if both `sourceMapping` and `PROJECTS_ROOT` are set by the user on purpose or by mistake?

docs/proposals/file-sync-location.md
johnmcollier: Good find 😅

docs/proposals/file-sync-location.md
johnmcollier: Good question. The devfile mentions has the followings (under sourceMapping):
>  sourceMapping: string
    Optional specification of the path in the container where project sources should be transferred/mounted when mountSources is true. When omitted, the value of the PROJECTS_ROOT environment variable is used.

So, the spec doesn't explicitly mention when both `sourceMapping` and `PROJECTS_ROOT` are set, but based off of what it does say, it seems `sourceMapping` would still take priority if both are set?

docs/proposals/file-sync-location.md
johnmcollier: We could certainly provide that as an option, but according to the devfile spec (under `sourceMapping`), it's expected that `PROJECTS_ROOT` can be overridden via an environment variable

docs/proposals/file-sync-location.md
kadel: > So, the spec doesn't explicitly mention when both `sourceMapping` and `PROJECTS_ROOT` are set, but based off of what it does say, it seems `sourceMapping` would still take priority if both are set?

+1

I think that description on devfile spec implies that `sourceMapping` takes priority.

docs/proposals/file-sync-location.md
kadel: This looks a little bit weird to me. Are we really want to map local env variables to the container env variables?

The way I was imagining this was that the `PROJECT_ROOT` env variable needs to be set in devfile.yaml in the container section. The benefit of using this over `soruceMapping` is that can have scripts inside container that are using this variable.

docs/proposals/file-sync-location.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Background 
+
+Currently, odo uses the following logic to determine where to sync source code to a devfile component:
+
+1) If the `sourceMapping` field is set, odo will mount the source volume under to the specified folder and sync the source code there
+
+2) If `sourceMapping` is unset, odo will do one of two things:
+   a) If a single project is defined in the devfile, odo will sync to `/projects/<projectName>`
+   b) If no projects are defined, or there are multiple projects defined, odo will sync to `/projects`
```
kadel, line 12: This problem cloud be solved with "local" components (https://github.com/devfile/kubernetes-api/issues/80) If devfile have support for local component, we could require from devfile to always have at least one project

docs/proposals/file-sync-location.md
johnmcollier: Re-reading:
```
 Optional specification of the path in the container where project sources should be transferred/mounted when mountSources is true. When omitted, the value of the PROJECTS_ROOT environment variable is used.
```

I suppose it is referring to the value of `PROJECTS_ROOT` in the container where the source is being mounted, in which case, it could be modified by setting it as an env var in the container.

I can update the doc accordingly.

johnmcollier: @kadel Updated the doc accordingly.

docs/proposals/file-sync-location.md
johnmcollier: Good to know about.

docs/proposals/file-sync-location.md
elsony: +1 on sourceMapping has priority

docs/proposals/file-sync-location.md
kadel: I was wrong :-/ Local component is something else than I thought :-/

PR #3647, opened 2020-07-27
add Kubernetes namespace support

### LINKED ISSUE #2899: odo project on Kubernetes
/kind user-story

## User Story
As an odo user, I want to be able to use `odo project *` commands regardless of what type of the cluster I'm using (Kubernetes or OpenShift`

## Acceptance Criteria 
- [x] `odo project create` should create Namespace if odo is connected to Kubernetes cluster, and Project if connected to OpenShift
- [x] `odo project delete` should delete Namespace if odo is connected to Kubernetes cluster, and Project if connected to OpenShift
- [x] `odo project get`should get the currently active Namespace from `KUBECONFIG` 
- [x] `odo project list` should list all user's Namespaces if odo is connected to Kubernetes cluster, and all users's Projects if connected to OpenShift
- [x] `odo project set` should set the currently active Namespace from `KUBECONFIG` 

## Links 
- https://github.com/openshift/odo/issues/2846#issuecomment-611064771
- https://github.com/openshift/odo/issues/2769

## Notes
odo should always use the project as an abstraction on top of Kubernetes Namespace and OpenShift Project. From the odo user's point of view, it should not matter what cluster they are using, the commands should remain the same.

/kind user-story
/area project
/area kubernetes
/area openshift
/priority high

kadel:
Add support for Namespace when odo is used with Kubernetes cluster

Fixes #2899 

**PR acceptance criteria**:

- [x] Unit test 
- [x] Integration test 
  - I just enabled existing tests for `odo project` command on Kubernetes
- [ ] Documentation 

run any of the `odo project *` commands against the Kubernetes cluster, they should work the same way as with OpenShift cluster


pkg/occlient/occlient.go
```diff
@@ -3281,6 +3315,11 @@ func (c *Client) IsRouteSupported() (bool, error) {
 	return c.isResourceSupported("route.openshift.io", "v1", "routes")
 }
 
+// IsProjectSupported checks if Project resource type is present on the cluster
+func (c *Client) IsProjectSupported() (bool, error) {
+	return c.isResourceSupported("project.openshift.io/v1", "v1", "projects")
```
girishramnani, line 3320: please remove `/v1` from here. this would fail with `v1` component not found

vendor/k8s.io/client-go/discovery/discovery_client_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
-	"github.com/googleapis/gnostic/OpenAPIv2"
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
```
mik-dass, line 30: Is this vendor changes required?

pkg/project/project_test.go
```diff
@@ -103,6 +167,63 @@ func TestCreate(t *testing.T) {
...
+				return true, fkWatch2, nil
+			})
+
+			client.SetDiscoveryInterface(fakeDiscoveryWithNamespace)
+
+			// The function we are testing
+			err := Create(client, tt.projectName, true)
+
+			// Checks for error in positive cases
+			if !tt.wantErr == (err != nil) {
```
mik-dass, line 214: can we reuse some parts from the above test e.g error checking, reactors etc?

kadel: I tried that, but the problem is that almost every line is different due to Namespace vs Project calls

mik-dass:
```
[odo]  ✗  Get https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/devfile.yaml: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```
/retest

pkg/project/project_test.go
mik-dass: OK

pkg/occlient/namespace.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/klog"
+)
+
+// CreateNamespace creates new namespace
+func (c *Client) CreateNamespace(name string) (*corev1.Namespace, error) {
```
mik-dass, line 15: I think the namespace functions related things should we moved to the kClient package as we have been adding the kubernetes related things like ingress etc there.

kadel: problem is that when I do that I will have to add pass whole `kclient` to all function that deal with project :-/

mik-dass: How about instead of passing the whole occlient or the kclient, we pass the context? The devfile context can contain both the clients and we can use them accordingly.

kadel: I can do that. But wouldn't it be weird if we use a struct that is from `genericclioptions` for all project functions?

kadel: I've tried that in the latest commit (a18c72cb03a00ca548feac60de287d148b56a3f4) take a look

mik-dass: I agree it would be weird to use this struct, but it might create confusion if we keep the namespace parts in occlient. In future we can try to find a way to make a common abstracted client.

mik-dass:
```
  * could not run steps: execution cancelled
  * could not run steps: step integration-e2e failed: "integration-e2e" pre steps failed: ["integration-e2e" pod "integration-e2e-ipi-conf" failed: context canceled
Link to step on registry info site: https://steps.ci.openshift.org/reference/ipi-conf
Link to job on registry info site: https://steps.ci.openshift.org/job?
```
/retest

pkg/occlient/namespace.go
kadel: So what do you suggest, should I remove the last commit and keep functions in `occlient` or should I use context? I have both options ready :-)

mik-dass: Let's keep it here in kclient package for now.

Issue #3615, opened 2020-07-21
url list and describe need update to display url defined in devfile endpoint entry


yangcao77:
/area devfile
/area url

After https://github.com/openshift/odo/issues/3515 is implemented, Ingress/Route can be created via devfile endpoint definition during odo push. We need to update the logic of `url list` and `url describe` to properly display the url defined in endpoint entry.

## Acceptance Criteria 
- [ ] `url list` should display info of url defined in devfile endpoint entry, also should display it's state
- [ ] `url describe` should display info of url defined in devfile endpoint entry, also should display it's state

kadel:
There is a related issue https://github.com/openshift/odo/issues/3718.

It might be better to address https://github.com/openshift/odo/issues/3718 before this one

elsony:
Blocked by #3718

kadel:
Wouldn't this be covered in https://github.com/openshift/odo/issues/3718 ?

yangcao77:
Yes. This one will be covered in the same PR of #3718.

PR #3665, opened 2020-07-31
create url during push with infomation defined in devfile endpoint

### LINKED ISSUE #3515: Implement container endpoints from devfile v2 spec
We need to implement the `endpoints` section from devfile v2 container definition.

### Acceptance criteria
 - [ ] `odo push`  against OpenShift should create resources based on described mappings with Route
 - [ ]  `odo push` against Kubernetes will create Services based on described mappings. If Ingress needs to be created we have problem with missing host information. In this case, odo should print a warning explaining what is the problem, and how user can create URL (example command should be automatically constructed with all flags based on endpoint definition) example: "Devfile defined exposed port 8080. In order to make this port accessible via URL you need to run `odo url create --port 8080 --path /api --host <yourIngressHost>`"
- [ ] Update devfiles in https://github.com/odo-devfiles/registry to include detailed configuration when needed (especially `exposure` )

https://devfile.github.io/devfile/_attachments/api-reference.html

### Mapping to Kubernetes Resource:
- `exposure:`
   - `public` - create Route/Ingress and Service (default)
   - `internal` - create only Service
   - `none` - don't create Service nor Route/Ingress
- `protocol:`
   - `http` - Protocol for port in Service will be `tcp`, 
   - `https` - Protocol for port in Service will be `tcp`, Ingress/Route (when `exposure: public` is set should be securet  with SSL (the same way that odo dues when `odo url create --secure` is executed 
   - `ws` - Protocol for port in Service will be `tcp`, 
   - `wss` - Protocol for port in Service will be `tcp`,  Ingress/Route (when `exposure: public` is set should be securet  with SSL (the same way that odo dues when `odo url create --secure` is executed 
   - `tcp` - Service port will be `tcp`, 
   - `udp` - Protocol for port in Service will be `udp`, 

### example of endpoint definition:
```yaml
endpoints:
 - name: web
    path: /api
    exposure: public
    protocol: http
    secure: false
    targetPort: 8080
```

Based on this definition odo should create:
- Service for container port 8080
- Ingress or Route for Service on port 8080 with `path: /api`

### LINKED ISSUE #3663: UDP support for devfile component services and URLs
/area devfile
/area url

After #3515 is implemented, Service/Ingress/Route can be created via devfile endpoint definition during odo push. 
The endpoint entry contains `protocol`, which would use either `TCP` or `UDP` as connection type. 

Since services of clusterIP type cannot support both TCP and UDP in a single service, we need to update our currently logic to support UDP connection. 

## Acceptance Criteria 
- [ ] For a single devfile component, create a service for all UDP ports( suggested service name <componentName>-UDP)
- [ ] For a single devfile component, create a service for all TCP ports( suggested service name <componentName>-TCP)
- [ ] Ingress/Route should use the corresponding service based on the target port. 
- [ ] Should delete the old ingress/route, and create a new one during `odo push` if the connection type(UDP/TCP) has been changed.

yangcao77:
**What type of PR is this?**
/area devfile
/area url

This PR implements the endpoints entry from devfile v2 container definition.
`odo push` creates url without running `odo url create` with the information specified in devfile endpoint.
 
odo push against OpenShift creates resources based on described mappings with Route
odo push against Kubernetes will create Services based on described mappings. If host information is missing in env.yaml a warning message will be shown.

This PR only implements the url creation. 
For url list and describe, need to rewrite the listing logic,  issue https://github.com/openshift/odo/issues/3615 is created to track.
For UDP service support, since service of type clusterIP cannot support both TCP and UDP. Our current logic needs to be change, will work with a separate issue: https://github.com/openshift/odo/issues/3663

`--secure` flag is removed  from `odo url create` to avoid conflict between `env.yaml` and `devfile.yaml`. We will always use `secure` defined in `devfile.yaml`.  The `secure` field will be completely removed within the list/describe issue (#3615 ). 

Added `path` in route & ingress creation. 

For `exposure` and `protocol` defined in devfile, follows behavior defined in devfile 2.0 schema: https://devfile.github.io/devfile/_attachments/api-reference.html

For same port, if `urlName` in env.yaml is different from `devfile.yaml`, will use the name from env.yaml; since env.yaml is considered as a "local" configuration, user runs `odo create` cli to intentionally set the value, therefore env.yaml has precedence over devfile.yaml 

Also implement https://github.com/openshift/odo/issues/3718#issuecomment-670648628, for URLs for devfile components, the ingress and routes created will be `urlName-componentName`. It won't affect the name saved in env.yaml file, also `url list`, `url describe` and `url, delete` will still use `urlName`, so user won't get confused. 

Fixes #3515 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

`make test-cmd-devfile-url`

Manual test:

Against Openshift cluster:
run `odo create` with the `devfile-with-multiple-endpoints.yaml` in this pr

directlry run `odo push`
should see 1 service with port 8080, 8090, 9090 created. 
should see 3 routes got created.
run `oc get route <routename> -o yaml`, should see route `8090-tcp` with path `/testpath`
 should see route `9090-tcp` is secured

run `odo url create testurl1 --port 8080 --host com --ingress`
modify `devfile`, remove endpoint entry with port=8080 (or can change the port, or can add `exposure: internal`)and run `odo push`. Should see error that port 8080 is not exposed. 

Modify the devfile endpoint entry, add `exposure: none` to one of the endpoints. 
`odo push` should see service does not contain the corresponding port


pkg/occlient/occlient_test.go
```diff
@@ -628,6 +644,11 @@ func TestCreateRoute(t *testing.T) {
 				if createdRoute.Spec.TLS != nil {
 					t.Errorf("tls config is set for a non secure url")
 				}
+				if tt.path == "" && createdRoute.Spec.Path != "/" {
+					t.Errorf("expect path: /, but get path: %v", createdRoute.Spec.Path)
+				} else if tt.path != "" && createdRoute.Spec.Path != tt.path {
+					t.Errorf("expect path: %v, but get path: %v", tt.path, createdRoute.Spec.Path)
```
mik-dass, line 650: I think you meant `got path`

pkg/url/url.go
```diff
@@ -880,13 +884,34 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
 		urls := parameters.EnvURLS
 		for _, url := range urls {
 			if url.Kind != envinfo.DOCKER {
+				if parameters.EndpointMap == nil {
+					klog.V(4).Infof("No Endpoint entry defined in devfile.")
+					return nil
+				}
+				exist := false
+				var endpoint parsercommon.Endpoint
+				if parameters.EndpointMap != nil {
```
mik-dass, line 893: We don't need this check. The condition is checked above.

mik-dass:
> --secure flag is removed from odo url create to avoid conflict between env.yaml and devfile.yaml. We will always use secure defined in devfile.yaml. The secure field will be completely removed within the list/describe issue

@yangcao77 @kadel I don't think we should remove the secure flag. If there is no endpoint in the devfile and the user wants to create a secure URL, he/she will have no option.

tests/integration/devfile/cmd_devfile_url_test.go
```diff
@@ -140,19 +129,14 @@ var _ = Describe("odo devfile url command tests", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, componentName)
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
-			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile.yaml"), filepath.Join(context, "devfile.yaml"))
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile-with-multiple-endpoints.yaml"), filepath.Join(context, "devfile.yaml"))
 
-			helper.CmdShouldPass("odo", "url", "create", url1, "--port", "3000", "--host", host, "--secure", "--ingress")
+			helper.CmdShouldPass("odo", "url", "create", url1, "--port", "9090", "--host", host, "--ingress")
```
prietyc123, line 134: Do we really need this spec? I can see secure url is created and verified in https://github.com/openshift/odo/pull/3665/files#diff-5b34c0d9e5819031b99fc98f27ed1690R100 spec. I would say lets add `odo url list` also in that spec. No need to create a separate spec for this

yangcao77:
> > --secure flag is removed from odo url create to avoid conflict between env.yaml and devfile.yaml. We will always use secure defined in devfile.yaml. The secure field will be completely removed within the list/describe issue
> 
> @yangcao77 @kadel I don't think we should remove the secure flag. If there is no endpoint in the devfile and the user wants to create a secure URL, he/she will have no option.
> 
> We should just error out in `odo url create` when a user tries to create a secure URL which is already defined in the devfile with the opposite settings.

@mik-dass The case you mentioned "no endpoint, but user want to create a secure URL" is invalid. 
In order to create a URL, the port has to be exposed in endpoint entry in devfile. 

One reason I want to remove the `--secure` flag, instead of showing error if there's any conflict, is because devfile now contains other fields `path`/`protocol`/`exposure`, we do not have those in our cli. If user wants to change the value of it, they need to modify devfile.  IMO, it will be more consistent and makes more sense to use devfile for all those settings. 
The only thing we really require is `kind` and `host`(for ingress only)
See Tomas's comment: https://github.com/openshift/odo/issues/3515#issuecomment-666304200

mik-dass:
> @mik-dass The case you mentioned "no endpoint, but user want to create a secure URL" is invalid.
In order to create a URL, the port has to be exposed in endpoint entry in devfile. Otherwise, you will get an error.

Sorry missed that. But what if the user wants to create a different URL and make it secure when the devfile one is not? The PR seems to turn all URLs secure or none secure based on the devfile value.

> One reason I want to remove the --secure flag, instead of showing error if there's any conflict, is because devfile now contains other fields path/protocol/exposure, we do not have those in our cli. If user wants to change the value of those settings, they need to modify devfile. IMO, it will be more consistent and makes more sense to use devfile for all those settings.

I guess the basic idea behind odo was that the user doesn't need to touch any yaml and I feel this is moving away from that and making the CLI a bit limiting.

> See Tomas's comment: #3515 (comment)

Looking at the comment, I see that `odo url create` is still a valid command and the push will just create a extra URL from values defined in the devfile. It doesn't mention anything about removing the flag.

yangcao77:
> But what if the user wants to create a different URL and make it secure when the devfile one is not? The PR seems to turn all URLs secure or not based on the devfile value.

User can modify the devfile.yaml to change the secure value of the url. 

> I guess the basic idea behind odo was that the user doesn't need to touch any yaml and I feel this is moving away from that and making the CLI a bit limiting. 
The devfile endpoint entry is a little it different. First of all, in order to run `odo url create` the port has to be defined in `TargetPort`, and it has to be exposed, i.e. `exposure=public`. The devfile.yaml is limiting the url cli already, no matter how we deal with the secure flag.

mik-dass:
> User can modify the devfile.yaml to change the secure value of the url.

So on the same port, one URL cannot be secure and another cannot be non secure.

> The devfile endpoint entry is a little it different. First of all, in order to run odo url create the port has to be defined in TargetPort; in addition, it has to be exposed, i.e. exposure=public. The devfile.yaml is limiting the url cli already, no matter how we deal with the secure flag.

@kadel @yangcao77 Then I think we need to discuss and come up with a better design.

mik-dass:
> @kadel @yangcao77 Then I think we need to discuss and come up with a better design.

Me and @kadel had a discussion about the URL design. We agreed to move the URLs to the devfile itself as keeping it in two places will create confusion. So `odo url create` will create a URL in the devfile and parts of the data that cannot be fitted in it e.g host etc will exists in the env.yaml against the URL name.

Regarding the conflict, we agreed to comment about the flag for now. Once `odo url create` starts putting the URLs in the devfile as endpoints, we will add it back again. The flag then will just mark the endpoints as `secure` in the devfile and push will interpret it correctly. The other commands e.g `odo url delete` will work accordingly.

@yangcao77 WDYT?

girishramnani:
@mik-dass we just confirmed this now in the meeting

mik-dass:
> @mik-dass we just confirmed this now in the meeting

Great :+1: Let's create a issue to keep track of it.

tests/integration/devfile/cmd_devfile_url_test.go
```diff
@@ -116,11 +105,11 @@ var _ = Describe("odo devfile url command tests", func() {
...
 			helper.CmdShouldPass("odo", "push")
-			helper.CmdShouldPass("odo", "url", "create", url2, "--port", "3000", "--host", host, "--ingress")
+			helper.CmdShouldPass("odo", "url", "create", url2, "--port", "8080", "--host", host, "--ingress")
```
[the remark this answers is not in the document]
yangcao77: This is a limitation we currently have, and we have discussed about that in today's meeting. it's gonna be a big change to support that, so https://github.com/openshift/odo/issues/3710 is created to track that

pkg/component/component.go
```diff
@@ -564,9 +565,10 @@ func ensureAndLogProperResourceUsage(resourceMin, resourceMax *string, resourceN
 //	componentConfig: Component configuration
 //	envSpecificInfo: Component environment specific information, available if uses devfile
 //  cmpExist: true if components exists in the cluster
+//  endpointMap: value is devfile endpoint entry, key is the TargetPort for each enpoint entry
 // Returns:
 //	err: Errors if any else nil
-func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool) (err error) {
+func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool, endpointMap map[int32]parsercommon.Endpoint) (err error) {
```
cdrage, line 571: So I've ran into issues before with regards to using maps as a parameter before as it's usually not in the correct order. 

Can you pass in an array of structs or something instead? Endpoint's would be different each time.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -297,14 +302,23 @@ func (a Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSpe
...
 	})
+
 	var containerPorts []corev1.ContainerPort
+
 	for _, c := range deploymentSpec.Template.Spec.Containers {
-		if len(containerPorts) == 0 {
-			containerPorts = c.Ports
-		} else {
+		// No need to check
+		if reflect.DeepEqual(a.Devfile.Ctx.GetApiVersion(), "1.0.0") {
```
cdrage, line 307: `1.0.0` should be a const

tests/integration/devfile/cmd_devfile_url_test.go
```diff
@@ -70,19 +70,8 @@ var _ = Describe("odo devfile url command tests", func() {
...
-				}
-				return false
-			})
-			helper.CmdShouldPass("odo", "url", "delete", url1, "-f")
-			helper.CmdShouldPass("odo", "push", "--project", namespace)
-
-			stdout = helper.CmdShouldFail("odo", "url", "list")
-			Expect(stdout).To(ContainSubstring("no URLs found"))
+			stdout = helper.CmdShouldPass("odo", "push", "--project", namespace)
+			Expect(stdout).Should(ContainSubstring(url1 + "." + host))
```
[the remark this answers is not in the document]
yangcao77: As mentioned in the PR description, the url listing is not going to work for URLs defined in devfile.  I'm currently working on #3615 for url list and describe
```
For url list and describe, need to rewrite the listing logic, issue #3615 is created to track.
```

pkg/occlient/occlient.go
```diff
@@ -2605,8 +2605,13 @@ func (c *Client) GetAllClusterServicePlans() ([]scv1beta1.ClusterServicePlan, er
 // CreateRoute creates a route object for the given service and with the given labels
 // serviceName is the name of the service for the target reference
 // portNumber is the target port of the route
+// path is the path of the endpoint URL
 // secureURL indicates if the route is a secure one or not
-func (c *Client) CreateRoute(name string, serviceName string, portNumber intstr.IntOrString, labels map[string]string, secureURL bool, ownerReference metav1.OwnerReference) (*routev1.Route, error) {
+func (c *Client) CreateRoute(name string, serviceName string, portNumber intstr.IntOrString, labels map[string]string, secureURL bool, path string, ownerReference metav1.OwnerReference) (*routev1.Route, error) {
+	routePath := "/"
```
cdrage, line 2579: Bit confused why we use `/` for this?

cdrage: Should we have a comment explaining why we use `/` for route. Also, did you check to see with regards to `\` or backslashes? Any niche cases / other possible input issues?

pkg/occlient/occlient_test.go
```diff
@@ -602,6 +603,21 @@ func TestCreateRoute(t *testing.T) {
 			existingDC: *fakeDeploymentConfig("blog", "", nil, nil, t),
 			secureURL:  true,
 		},
+
+		{
+			name:       "Case : specify a path",
```
cdrage, line 608: Weird case name. I think you forgot to specify a number? For ex. `Case 1: Specify a path` or something?

tests/integration/devfile/cmd_devfile_url_test.go
yangcao77: same for https://github.com/openshift/odo/pull/3665#discussion_r466633933

pkg/odo/cli/component/common_push.go
```diff
@@ -127,7 +128,7 @@ func (cpo *CommonPushOptions) createCmpIfNotExistsAndApplyCmpConfig(stdout io.Wr
 		}
 	}
 	// Apply config
-	err := component.ApplyConfig(cpo.Context.Client, nil, *cpo.LocalConfigInfo, envinfo.EnvSpecificInfo{}, stdout, cpo.doesComponentExist)
+	err := component.ApplyConfig(cpo.Context.Client, nil, *cpo.LocalConfigInfo, envinfo.EnvSpecificInfo{}, stdout, cpo.doesComponentExist, map[int32]parsercommon.Endpoint{})
```
cdrage, line 131: See other comment, may be better to pass in array rather than map as Endpoint{} will be unordered

tests/integration/devfile/cmd_devfile_url_test.go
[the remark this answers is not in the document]
yangcao77: This is a limitation we always have for devfile components, it won't break any functions, so would have lower priority than #3615 and #3663

pkg/url/url.go

pkg/occlient/occlient_test.go
yangcao77: updated

pkg/occlient/occlient.go
yangcao77: `/` is the default path for routes and ingresses. 
e.g. if path is `/`, for URL `http://1.2.3.4/`  and URL `http://1.2.3.4` will be treated the same.

pkg/occlient/occlient_test.go
yangcao77: Non of the test cases in `TestCreateRoute` has a case number defined, I'm following the format

tests/integration/devfile/cmd_devfile_url_test.go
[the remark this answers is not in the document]
mik-dass: > This is a limitation we currently have, and we have discussed about that in today's meeting. it's gonna be a big change to support that, so #3710 is created to track that

It is possible to have the same port for multiple URLs currenty https://github.com/openshift/odo/issues/3710#issuecomment-670303026

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -111,6 +112,27 @@ func GetContainers(devfileObj devfileParser.DevfileObj) ([]corev1.Container, err
...
+func GetEndpoints(data data.DevfileData) (map[int32]common.Endpoint, error) {
+	endpointsMap := make(map[int32]common.Endpoint)
+
+	for _, comp := range adaptersCommon.GetDevfileContainerComponents(data) {
+		// Currently type container is the only devfile component that odo supports
+		if comp.Container.Endpoints != nil {
+			for _, endpoint := range comp.Container.Endpoints {
+				// TargetPort is a required entry for an Endpoint
+				// Devfile should not contains multiple identical TargetPorts, since all containers are inside one pod
+				if _, keyexist := endpointsMap[endpoint.TargetPort]; keyexist {
```
mik-dass, line 125: Why are we using the `targetPort` as the key? Why not use the endpoint name?

pkg/url/url.go
```diff
@@ -880,13 +884,30 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
+				if !exist || endpoint.Exposure == "none" || endpoint.Exposure == "internal" {
+					return fmt.Errorf("port %v defined in env.yaml file for URL %v is not exposed in devfile Endpoint entry", url.Port, url.Name)
+				}
+				secure := false
+				if endpoint.Secure || endpoint.Protocol == "https" || endpoint.Protocol == "wss" {
+					secure = true
+				}
+				path := "/"
+				if endpoint.Path != "" {
+					path = endpoint.Path
```
mik-dass, line 901: This will convert all the URLs on the same port as secure or not secure and based on the one defined in the devfile. It will also use the same path for all the URLs. I think the key to the endpoint map should be the endpoint name.

pkg/url/url.go
```diff
@@ -899,11 +920,55 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
+				continue
+			}
+			exist := false
+			for _, envURL := range urlLOCAL {
+				if envURL.Spec.Port == int(port) {
+					exist = true
+					break
+				}
+			}
+			if !exist {
```
mik-dass, line 944: Why are we not creating the endpoint URL if the port is used by a env URL?

pkg/url/url.go
yangcao77: I understand your concern. 
When we only have env.yaml to store URL settings, we expect the devfile not to contain multiple identical ports. It is because the only value we care about in devfile.yaml endpoint entries was `Name` and `TargetPort`. We use these two value to define containers in deployment spec and exposed port in service spec, that is why we expect for a single devfile, there should not exist multiple `Targetport`. 


However, with devfile 2.0, the URL is a little bit tricky now since both env.yaml and endpoints in devfile.yaml contains URL settings and are able to create URL during push. 
If we want to support duplicate port in devfile.yaml in this PR, we encontereed more issues. For example, if multiple endpoints with same `TargetPort` defined in devfile, but with different secure value/path/expose defined, when user call `odo url create` with a url name not defined in devfile, which endpoint entry should it map with?



@girishramnani @kadel @elsony we discussed this issue in last Thursday's meeting, and agreed it cannot be simply done in this PR, and requires more consideration, so we just continue to only allow one single `TargetPort` in this PR. 

We need https://github.com/openshift/odo/issues/3710 and https://github.com/openshift/odo/issues/3718 to address this issue.

pkg/url/url.go
yangcao77: It is because of this https://github.com/openshift/odo/pull/3665#discussion_r468094439

As described is the PR description, 
```
For same port, if urlName in env.yaml is different from devfile.yaml, will use the name from env.yaml; since env.yaml is considered as a "local" configuration, user runs odo create cli to intentionally set the value, therefore env.yaml has precedence over devfile.yaml
```

We have discussed the issue in last Thursday's meeting, for this PR, it is simply to support url creation with endpoint settings defined in devfile. Due to the awkward situation we are in, with both env.yaml and devfile.yaml; we want to implement https://github.com/openshift/odo/issues/3718 before we make more changes on URL creation logics.

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: I totally agree with you that using endpoint name as the key makes more sense, if we has https://github.com/openshift/odo/issues/3718 implemented; which means for every entry in env.yaml we are going to have an exact match in devfile.yaml, then we can use the urlName as a key. 

For now, as described in https://github.com/openshift/odo/pull/3665#discussion_r468094439, we currently does not support duplication port in devfile, using `targetport` as a key is also unique.

yangcao77: I have considered about this, so see the second point in issue description when I create this issue. 
https://github.com/openshift/odo/issues/3710#issue-674544229

```
update the logic of url creation. (i.e. any maps use port number as key should use URL name instead)
```

tests/integration/cmd_url_test.go
```diff
@@ -131,7 +131,7 @@ var _ = Describe("odo url command tests", func() {
 			// get the route URL
 			fullURLPath := helper.DetermineRouteURL(context)
 			pathNoHTTP := strings.Split(fullURLPath, "//")[1]
-			desiredURLDescribeJSON = fmt.Sprintf(`{"kind":"url","apiVersion":"odo.dev/v1alpha1","metadata":{ "name": "myurl","creationTimestamp": null},"spec":{"host":"%s","protocol": "https","port": 8080,"secure": true,"kind": "route"},"status": {"state": "Pushed"}}`, pathNoHTTP)
+			desiredURLDescribeJSON = fmt.Sprintf(`{"kind":"url","apiVersion":"odo.dev/v1alpha1","metadata":{ "name": "myurl","creationTimestamp": null},"spec":{"host":"%s","protocol": "https","port": 8080,"secure": true, "path": "/", "kind": "route"},"status": {"state": "Pushed"}}`, pathNoHTTP)
```
amitkrout, line 134: I doubt the ```"path": "/"``` will be same on windows platform.

kadel: > I doubt the "path": "/" will be same on windows platform.

This is a path part of an URL so it will be the same everywhere.

pkg/component/component.go
kadel: @cdrage Why would this be an issue? The order should not matter here.

pkg/occlient/occlient_test.go
kadel: Cases don't need to have a number. More important is to have good description.

tests/integration/devfile/cmd_devfile_url_test.go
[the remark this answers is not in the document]
amitkrout: @yangcao77 Does ```odo url delete url1 -f``` is supported for URLs defined in devfile ? if yes you can add these validation at the end of the spec.

pkg/devfile/adapters/kubernetes/utils/utils.go
kadel: @mik-dass Do you think that it would make sense to address this in separate PR as part of addressing https://github.com/openshift/odo/issues/3710?

pkg/url/url.go
kadel: @mik-dass Do you think that it would make sense to address this in separate PR as part of addressing #3718 ?

tests/integration/devfile/cmd_devfile_url_test.go
```diff
@@ -70,19 +70,8 @@ var _ = Describe("odo devfile url command tests", func() {
...
 			helper.CmdShouldPass("odo", "url", "create", url1, "--port", "3000", "--host", host, "--ingress")
-			helper.CmdShouldPass("odo", "push")
-			helper.WaitForCmdOut("odo", []string{"url", "list"}, 1, false, func(output string) bool {
-				if strings.Contains(output, url1) {
-					Expect(output).Should(ContainSubstring(url1 + "." + host))
-					return true
-				}
-				return false
-			})
-			helper.CmdShouldPass("odo", "url", "delete", url1, "-f")
```
amitkrout, line 81: @yangcao77 Does ```odo url delete url1 -f``` is supported for URLs defined in devfile ? if yes you can add these validation at the end of the spec.

tests/integration/devfile/cmd_devfile_url_test.go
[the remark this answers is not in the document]
kadel: > It is possible to have the same port for multiple URLs currenty [#3710 (comment)](https://github.com/openshift/odo/issues/3710#issuecomment-670303026)

but not for devfile components

tests/integration/devfile/cmd_devfile_url_test.go
kadel: @amitkrout It works for manually created URLs, but not for automatically created URLs from endpoint definition in devfile. 
`odo url delete -f` is tested many times in this file. I would say that it is ok to skip it in this test. The point of this test is to make sure that URL is listed  in the `odo push` output

amitkrout: @kadel 
> It works for manually created URLs, but not for automatically created URLs from endpoint definition in devfile.

Then i assume that it is same issue like url listing which does not work for URLs defined in devfile. 

> odo url delete -f is tested many times in this file. I would say that it is ok to skip it in this test. The point of this test is to make sure that URL is listed in the odo push output

My concern is why the existing steps is removed without any reason like #3615, may be the steps has been added initially for certain reason. Its ok for me if the same functionalities is verified in in other spec 👍

yangcao77: Yes. The same reason like url listing and describe for URLs defined in devfile. 
It is expected to resolve in issue https://github.com/openshift/odo/issues/3718
Since we want to move the url creation and deletion into devfile.yaml.

pkg/url/url.go
mik-dass: > For example, if multiple endpoints with same TargetPort defined in devfile, but with different secure value/path/expose defined, when user call odo url create with a url name not defined in devfile, which endpoint entry should it map with?

As you said, we don't support multiple endpoints with the same TargetPort currently in this PR, so it will fail at the time of validation. My concern is why are we deciding the path, secure etc on the basis of the endpoint with the same target port. We currently don't have a flag for path, so it might confuse the user why the URL he/she created is using a different path. I think we should leave the secure, path etc values  for user created URLs to their defaults e.g secure:false, path:/ etc for now.

> @girishramnani @kadel @elsony we discussed this issue in last Thursday's meeting, and agreed it cannot be simply done in this PR, and requires more consideration, so we just continue to only allow one single TargetPort in this PR.

I agree we should do it in a different PR.

pkg/devfile/adapters/kubernetes/utils/utils.go
mik-dass: I am confused. If the user does `odo push` and the URL couldn't be created due to missing host information etc, what are we expecting/asking the user to do? 
- Create a URL with the exact name as defined in the endpoint?
- Create any URL on the same port and we will map it to the endpoint?

I like the first option better as it reduces a lot of confusion like https://github.com/openshift/odo/pull/3665#discussion_r468605247

> @mik-dass Do you think that it would make sense to address this in separate PR as part of addressing #3710?

@kadel Please have a look.

pkg/url/url.go
mik-dass: So this was implemented to support the second case in mentioned here https://github.com/openshift/odo/pull/3665#discussion_r468624308?

pkg/url/url.go
yangcao77: As I said, I totally understand your concern, how we handle the URL creation is not the best attempt, that is why we need #3718 . And this PR blocks #3718 since it implements the URL creation using devfile.
If we want to implement #3718 as the next PR, does it worth the effort to make this change? As we eventually would have all info saved in devfile.

pkg/url/url.go
yangcao77: Yes. If we want to implement the first option, then this means we only allow user to create url using name defined in devfile.yaml, which bring more limitation. Which means, it breaks the current feature, user can no longer create one route and one ingress at the same time using the same port, coz only one URL can be create to match the endpoint name. 
Almost all the integration test need to be changed because of that change, and we need to change those tests back when we implement #3718. 

Some of your suggestions can be made as part of the PR and also is with good reason; the question is, do we want to spend those effort knowing that we are going to work on #3718 as soon as this PR get merged.

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: See comment https://github.com/openshift/odo/pull/3665#discussion_r468640987

pkg/url/url.go
mik-dass: >  we only allow user to create url using name defined in devfile.yaml, which brings more limitation

I agree but the other case brings confusion described in https://github.com/openshift/odo/pull/3665#discussion_r468605247.

> Which means, it breaks the current feature. ( For example, user can no longer create one route and one ingress at the same time using the same port, coz only one URL can be created to match the endpoint name. )

I didn't get you. How does it break the current feature? They can use `odo url create` and create a route and ingress on the same port if they want. What am I missing?

pkg/url/url.go
mik-dass: > If we want to implement #3718 as the next PR, does it worth the effort to make this change? As we eventually would have all info saved in devfile.

I know we need this to move forward. But if https://github.com/openshift/odo/issues/3718 doesn't get in time, the migration later might become difficult.

mik-dass: Anyways leave it, let's fix them in #3718.

pkg/url/url.go
mik-dass: Anyways leave it. Let's fix them in #3718.

pkg/devfile/adapters/kubernetes/utils/utils.go
mik-dass: https://github.com/openshift/odo/pull/3665#discussion_r468658887

Anyways leave it for now. Let's fix it in #3718 .

pkg/url/url.go
kadel: > I know we need this to move forward. But if #3718 doesn't get in time, the migration code later to move env URLs to devfile might become tricky.

Agreed. We need to make #3718 a high priority. We should not release 2.0 GA without it

pkg/url/url.go
```diff
@@ -346,6 +350,8 @@ func Create(client *occlient.Client, kClient *kclient.Client, parameters CreateP
 
 			ownerReference = occlient.GenerateOwnerReference(dc)
 		} else {
+			// to avoid error due to deplicate ingress name defined in different devfile components
```
mik-dass, line 353: `duplicate`

mik-dass, line 353: did you mean `duplicate`?

yangcao77:
@mik-dass Can you help review the PR?  #3718 has some dependency on this PR.  Thanks.

pkg/url/url.go
```diff
@@ -963,7 +1034,13 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
 				continue
 			}
 			// delete the url
-			err := Delete(client, kClient, urlName, parameters.ApplicationName, urlSpec.Spec.Kind)
+			deleteURLName := urlName
+			if parameters.IsExperimentalModeEnabled && kClient != nil {
+				// route/ingress name is defined as <urlName>-<componentName>
+				// to avoid error due to deplicate ingress name defined in different devfile components
```
mik-dass, line 1040: did you mean `duplicate`?

pkg/url/url.go
```diff
@@ -981,7 +1058,11 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
 			if urlInfo.Spec.Kind == envinfo.INGRESS && kClient == nil {
 				continue
 			}
-
+			// if parameters.IsExperimentalModeEnabled && kClient != nil {
+			// 	// route/ingress name is defined as <urlName>-<componentName>
+			// 	// to avoid error due to deplicate ingress name defined in different devfile components
```
mik-dass, line 1063: `duplicate`?

pkg/url/url.go
```diff
@@ -899,11 +924,55 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
+						Spec: URLSpec{
+							Port:   int(port),
+							Secure: secure,
+							Kind:   envinfo.ROUTE,
+							Path:   path,
+						},
+					}
+				} else {
+					// display warning since Host info is missing
+					log.Warningf("Unable to create ingress, missing host information for Endpoint %v, please check instructions on URL creation (refer `odo url create --help`)\n", endpoint.Name)
```
mik-dass, line 970: this should be printed after `Applying URL change`

pkg/url/url.go
```diff
@@ -308,10 +311,11 @@ func Create(client *occlient.Client, kClient *kclient.Client, parameters CreateP
 
 		}
 
-		ingressParam := kclient.IngressParameter{ServiceName: serviceName, IngressDomain: ingressDomain, PortNumber: intstr.FromInt(parameters.portNumber), TLSSecretName: parameters.secretName}
+		ingressParam := kclient.IngressParameter{ServiceName: serviceName, IngressDomain: ingressDomain, PortNumber: intstr.FromInt(parameters.portNumber), TLSSecretName: parameters.secretName, Path: parameters.path}
 		ingressSpec := kclient.GenerateIngressSpec(ingressParam)
 		objectMeta := kclient.CreateObjectMeta(parameters.componentName, kClient.Namespace, labels, nil)
-		objectMeta.Name = parameters.urlName
+		// to avoid error due to deplicate ingress name defined in different devfile components
```
mik-dass, line 317: `duplicate`?

pkg/odo/cli/url/describe.go
```diff
@@ -114,6 +114,9 @@ func (o *URLDescribeOptions) Run() (err error) {
 			if err != nil {
 				return err
 			}
+			// route/ingress name is defined as <urlName>-<componentName>
+			// to avoid error due to deplicate ingress name defined in different devfile components
```
mik-dass, line 118: `duplicate`?

mik-dass:
```
    Expected
        <string>: 
        Validation
         •  Validating the devfile  ...
 ✓  Validating the devfile [157947ns]
        
        Creating Kubernetes resources for component dkokdc
         •  Waiting for component to start  ...
 ✓  Waiting for component to start [32s]
        
        Applying URL changes
        
    to contain substring
        <string>: Changes successfully pushed to component
```
@yangcao77 Failure on travis. Please have a look.

pkg/odo/cli/url/describe.go
yangcao77: updated

pkg/url/url.go

pkg/url/url.go
yangcao77: updated

pkg/url/url.go
yangcao77: These were comments I forgot to delete. Removed.

pkg/url/url.go

yangcao77:
@mik-dass 
Found out the travis failure was due to the namespace is empty when trying to create route for devfile component. 
It is because we use devfile endpoint entry to create route when envinfo does not contain any namespace information. Modified to code to use kClient's namespace when creating temporary oclient in order to support route URL for devfile component. 
The failure should be fixed now.

mik-dass:
```
[odo] I0812 15:35:43.854797   16203 util.go:734] HTTPGetRequest: http://fake/devfiles/index.json
[odo] I0812 15:35:43.854797   16203 util.go:734] HTTPGetRequest: https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/index.json
[odo] I0812 15:35:43.854843   16203 util.go:755] Response will be cached in /tmp/odohttpcache for 15m0s
[odo] I0812 15:35:43.854843   16203 util.go:755] Response will be cached in /tmp/odohttpcache for 15m0s
[odo] I0812 15:35:43.854968   16203 util.go:768] Cached response used.
[odo]  ⚠  Registry DefaultDevfileRegistry is not set up properly with error: Unable to unmarshal the devfile index.json from https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/index.json: invalid character ']' after top-level value, please check the registry URL and credential (refer `odo registry update --help`)
[odo] 
[odo]  ⚠  Registry fake is not set up properly with error: Unable to download the devfile index.json from http://fake/devfiles/index.json: Get http://fake/devfiles/index.json: dial tcp: lookup fake on 169.254.169.254:53: no such host, please check the registry URL and credential (refer `odo registry update --help`)
[odo] 
[odo]  ✗  no deployable components found
Deleting project: oszpmyjpvv
Running kubectl with args [kubectl delete namespaces oszpmyjpvv]
[kubectl] namespace "oszpmyjpvv" deleted
Setting current dir to: /home/travis/gopath/src/github.com/openshift/odo/tests/integration/devfile
Deleting dir: /tmp/803438858
```
Looks like a catalog error on Travis. Restarting the job.

mik-dass:
```
[mrinaldas@localhost project]$ odo push

Validation
 ✓  Validating the devfile [142438ns]

 ✓  Waiting for component to start [5s]

Applying URL changes
 ✓  URL http-3000: http://http-3000-nodejs-blah.b6ff.rh-idev.openshiftapps.com/ created

Syncing to component nodejs
 ✓  Checking files for pushing [7ms]
 ✓  Syncing files to the component [10s]

 ✓  Executing install command "npm install", if not running [8s]
 ✓  Executing run command "npm start", if not running [4s]

[mrinaldas@localhost project]$ odo url create url
 ✓  URL url created for component: nodejs

[mrinaldas@localhost project]$ odo push

Validation
 ✓  Validating the devfile [163008ns]

 ✓  Waiting for component to start [616ms]

Applying URL changes
 ✓  URL http-3000 successfully deleted
 ✓  URL url: http://url-nodejs-blah.b6ff.rh-idev.openshiftapps.com/ created

Syncing to component nodejs
 ✓  Checking file changes for pushing [765340ns]

[mrinaldas@localhost project]$ odo push
```
it seems if we create a new URL, the devfile one is deleted.

yangcao77:
@mik-dass It is because these two URLs are using same port, and we have discussed about it in this thread: https://github.com/openshift/odo/pull/3665#discussion_r466806641

PR #3681, opened 2020-08-03
Caching registry responses to fix client.Timeout issue

kadel:
**What type of PR is this?**

/kind bug
/kind flake

Test are often failing on 
```
[odo]  ⚠  Registry DefaultDevfileRegistry is not set up properly with error: Unable to download the devfile index.json from https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/index.json: Get https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/index.json: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers), please check the registry URL and credential (refer `odo registry update --help`)

```
and 
```
[odo]  ✗  Get https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/devfile.yaml: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

This PR adds caching of responses coming from Devfile registry.
There is a new preference option `RegistryCacheTime`  that controls how for how long (in minutes) odo will cache responses from devfile registry.
The default cache time is 15min. `odo preference set registrycachetime 0` can be used to disable caching.

This drastically reduces the number of requests against devfile registries and it helps with timeouts that were caused by our CI hitting GitHub requests limits (60 req/min from single IP)

Fixes https://github.com/openshift/odo/issues/3684

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

 run command requesting information from the registry with verbose logging
 for example `odo catalog list components -v 4`

it output would see 
```
I0805 10:16:52.949943    1758 util.go:734] HTTPGetRequest: https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/index.json
I0805 10:16:52.950176    1758 util.go:744] Response will be cached in /var/folders/0l/m50bc_nj27lf8pbd_sdv_7h40000gn/T/odohttpcache for 15m0s
```
Which indicates that where the cases directory is that odo is using.
In the directory, you should see files with cached responses, you even can check its content.

If you run the catalog command again you will see the following line in the output indicating that the cached response was used
```
I0805 10:16:52.950448    1758 util.go:755] Using cached response.
```

By default, the responses are cached for 15min. You can reduce cache time to 1min
```
odo preference set RegistryCacheTime 1
```
Now if you run `odo catalog list components -v 4` again, and it was more than 1min since the first run, you should see the following message in the log
```
I0804 19:52:42.303479   82783 httpcache.go:21] Removing cache file 68cfbf2eeed643a845f1fc27856b9c74, because it is older than 1m0s
```

unset preference to get back to default
```
odo preference unset RegistryCacheTime
```


pkg/odo/cli/component/create.go
```diff
@@ -923,7 +924,7 @@ func (co *CreateOptions) Run() (err error) {
 						},
 						Filepath: DevfilePath,
 					}
-					err = util.DownloadFile(params)
+					err = util.DownloadFile(params, 0)
```
girishramnani, line 927: can we make this `0` a var-args or a functional parameter or make a new function called `DownloadFileWithCache(a,b)` which takes the second param because at many places its being set to `0` so would be better that the standard function has `DownloadFile` ( this wouldn't have the `cacheFor` param ) and a function `DownloadFileWithCache` that takes the `cacheFor`.

kadel: Wouldn't that just complicate things?

kadel: @girishramnani someting like this 21d5f510aec836ad5f2cc395864cb097e25bf2bf?

pkg/util/util.go
```diff
@@ -708,7 +712,8 @@ func GetRemoteFilesMarkedForDeletion(delSrcRelPaths []string, remoteFolder strin
 }
 
 // HTTPGetRequest gets resource contents given URL and token (if applicable)
-func HTTPGetRequest(request HTTPRequestParams) ([]byte, error) {
+// cacheFor determines how long the response should be cached (in minutes), 0 for caching
```
mik-dass, line 715: `0 for no caching`

pkg/util/util.go
```diff
@@ -719,19 +724,37 @@ func HTTPGetRequest(request HTTPRequestParams) ([]byte, error) {
...
 			ResponseHeaderTimeout: ResponseHeaderTimeout,
 		},
 		Timeout: HTTPRequestTimeout,
 	}
+
+	klog.V(4).Infof("HTTPGetRequest: %s", req.URL.String())
+
+	if cacheFor > 0 {
+		httpCacheTime := time.Duration(cacheFor) * time.Minute
+		_ = os.Mkdir(httpCacheDir, 0750)
```
mik-dass, line 738: we should handle the error

pkg/util/util.go
```diff
@@ -1017,10 +1040,11 @@ func Unzip(src, dest, pathToUnzip string) ([]string, error) {
 	return filenames, nil
 }
 
-// DownloadFile downloads the file to the filepath given URL and token (if applicable)
-func DownloadFile(params DownloadParams) error {
+// DownloadFileWithCache downloads the file to the filepath given URL and token (if applicable)
+// cacheFor determines how long the response should be cached (in minutes), 0 for caching
```
mik-dass, line 1044: `0 for no caching`?

pkg/util/util.go
kadel: oh, forgot to update the comment

pkg/preference/preference.go
```diff
@@ -367,6 +380,16 @@ func (c *PreferenceInfo) SetConfiguration(parameter string, value string) error
 			}
 			c.OdoSettings.PushTimeout = &typedval
 
+		case "registrycachetime":
```
mik-dass, line 383: please add a test case for this in the test function of SetConfiguration.

pkg/util/util.go
kadel: fixd

pkg/preference/preference.go
kadel: added

pkg/preference/preference.go
```diff
@@ -141,6 +151,9 @@ type OdoSettings struct {
 
 	// RegistryList for telling odo to connect to all the registries in the registry list
 	RegistryList *[]Registry `yaml:"RegistryList,omitempty"`
+
+	// RegistryCacheTime how long odo should cache information from registry
+	RegistryCacheTime *int `yaml:"RegistryCacheTime,omitempty"`
```
mik-dass, line 156: we need to also update the help message
```
# Set a preference value in the global preference
  odo preference set UpdateNotification false
  odo preference set NamePrefix "app"
  odo preference set Timeout 20
  odo preference set BuildTimeout 40
  odo preference set PushTimeout 30
  odo preference set Experimental true
  odo preference set PushTarget docker
```

kadel: this is just an example, I don't think that we need to show example for all preference.

If you scroll up to `Available Parameters` section you will see description for `RegistryCacheTime` option

mik-dass: Sorry didn't notice the `Available Parameters` section.

pkg/odo/cli/component/create.go
girishramnani: > Wouldn't that just complicate things?

Would complicate the code but make the interface cleaner for whoever uses it

pkg/preference/preference.go
kadel: > Sorry didn't notice the `Available Parameters` section.

I had the same problem first time :-) This is why I was thinking about reducing the number of examples that we have there :-)

kadel:
/hold 
I have test this on windows

kadel:
```
{"component":"entrypoint","file":"prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2020-08-05T12:03:31Z"}

```
/retest

adisky:
@kadel Tried this out works well, just a query shouldn't we delete the cache dir when cache is disabled? 
and how about using the ~/.odo folder for cache dir  `~/.odo/cache/`?

kadel:
> /hold
> I haven to test this on windows

tested on windows, works as expected 

/hold cancel

kadel:
> @kadel Tried this out works well, just a query shouldn't we delete the cache dir when cache is disabled?

We can change this behavior later if needed. Don't want to get too deep into this. We need badly some caching.


> and how about using the `~/.odo` folder for cache dir `~/.odo/cache/`?

I was thinking about ~/.odo, but then I went with temporary. Because I think it is better for this, it is just a temporary cache, so it is ok if it gets lost, and also I feel this is more an implementation detail that uses doesn't need to know about.

PR #3705, opened 2020-08-05
Make Devfile the default deployment mechanism

### LINKED ISSUE #3550: Move the devfile feature set out of the experimental mode
/kind user-story

## User Story
As a  user I should be able to use the devfile features without having experimental mode enabled. The decision between devfile and s2i will be made based on which config file is present locally.

## Acceptance Criteria 
- [ ] It should use devfile by default
- [ ] It should allow user to use devfile without experimental mode
- [ ] resolve broken tests
## Links 
- Feature Request: 
- Related issue: 

/kind user-story

cdrage:
**What type of PR is this?**

/kind feature

Makes Devfile the default deployment mechanism, removing S2I in favour
of Devfile deployment.

Closes #3550 https://github.com/openshift/odo/issues/3842

Run:

```sh
odo preference set experimental false
odo create --starter nodejs
odo push
```

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -25,7 +25,6 @@ import (
 	parsercommon "github.com/openshift/odo/pkg/devfile/parser/data/common"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
-	"github.com/openshift/odo/pkg/odo/util/experimental"
 	"github.com/openshift/odo/pkg/odo/util/validation"
 	"github.com/openshift/odo/pkg/preference"
 	"github.com/openshift/odo/pkg/storage"
@@ -565,10 +564,10 @@ func ensureAndLogProperResourceUsage(resourceMin, resourceMax *string, resourceN
 //	envSpecificInfo: Component environment specific information, available if uses devfile
 //  cmpExist: true if components exists in the cluster
 //  endpointMap: value is devfile endpoint entry, key is the TargetPort for each enpoint entry
+//  isS2I: Legacy option. Set as true if you want to use the old S2I method as it differentiates slightly.
 // Returns:
 //	err: Errors if any else nil
-func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool, containerComponents []parsercommon.DevfileComponent) (err error) {
-	isExperimentalModeEnabled := experimental.IsExperimentalModeEnabled()
+func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool, containerComponents []parsercommon.DevfileComponent, isS2I bool) (err error) {
 
 	if client == nil {
 		var err error
@@ -580,7 +579,7 @@ func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConf
 		client.Namespace = kClient.Namespace
 	}
 
-	if !isExperimentalModeEnabled {
+	if isS2I {
 		// if component exist then only call the update function
 		if cmpExist {
 			if err = Update(client, componentConfig, componentConfig.GetSourceLocation(), stdout); err != nil {
@@ -591,7 +590,7 @@ func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConf
 
 	var componentName string
 	var applicationName string
-	if !isExperimentalModeEnabled || kClient == nil {
+	if isS2I || kClient == nil {
 		componentName = componentConfig.GetName()
 		applicationName = componentConfig.GetApplication()
 	} else {
@@ -605,13 +604,13 @@ func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConf
 	}
 
 	return urlpkg.Push(client, kClient, urlpkg.PushParameters{
-		ComponentName:             componentName,
-		ApplicationName:           applicationName,
-		ConfigURLs:                componentConfig.GetURL(),
-		EnvURLS:                   envSpecificInfo.GetURL(),
-		IsRouteSupported:          isRouteSupported,
-		IsExperimentalModeEnabled: isExperimentalModeEnabled,
-		ContainerComponents:       containerComponents,
+		ComponentName:       componentName,
+		ApplicationName:     applicationName,
+		ConfigURLs:          componentConfig.GetURL(),
+		EnvURLS:             envSpecificInfo.GetURL(),
+		IsRouteSupported:    isRouteSupported,
+		ContainerComponents: containerComponents,
+		IsS2I:               isS2I,
 	})
 }
 
--- a/pkg/odo/cli/cli.go
+++ b/pkg/odo/cli/cli.go
@@ -205,9 +205,6 @@ func odoRootCmd(name, fullName string) *cobra.Command {
 		debug.NewCmdDebug(debug.RecommendedCommandName, util.GetFullName(fullName, debug.RecommendedCommandName)),
 		registry.NewCmdRegistry(registry.RecommendedCommandName, util.GetFullName(fullName, registry.RecommendedCommandName)),
 		component.NewCmdTest(component.TestRecommendedCommandName, util.GetFullName(fullName, component.TestRecommendedCommandName)),
-	)
-
-	rootCmd.AddCommand(
 		env.NewCmdEnv(env.RecommendedCommandName, util.GetFullName(fullName, env.RecommendedCommandName)),
 	)
 
--- a/pkg/odo/cli/cli.go
+++ b/pkg/odo/cli/cli.go
@@ -25,7 +25,6 @@ import (
 	"github.com/openshift/odo/pkg/odo/cli/version"
 	"github.com/openshift/odo/pkg/odo/util"
 	odoutil "github.com/openshift/odo/pkg/odo/util"
-	"github.com/openshift/odo/pkg/odo/util/experimental"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -204,15 +203,13 @@ func odoRootCmd(name, fullName string) *cobra.Command {
 		config.NewCmdConfiguration(config.RecommendedCommandName, util.GetFullName(fullName, config.RecommendedCommandName)),
 		preference.NewCmdPreference(preference.RecommendedCommandName, util.GetFullName(fullName, preference.RecommendedCommandName)),
 		debug.NewCmdDebug(debug.RecommendedCommandName, util.GetFullName(fullName, debug.RecommendedCommandName)),
+		registry.NewCmdRegistry(registry.RecommendedCommandName, util.GetFullName(fullName, registry.RecommendedCommandName)),
+		component.NewCmdTest(component.TestRecommendedCommandName, util.GetFullName(fullName, component.TestRecommendedCommandName)),
 	)
 
-	if experimental.IsExperimentalModeEnabled() {
-		rootCmd.AddCommand(
-			registry.NewCmdRegistry(registry.RecommendedCommandName, util.GetFullName(fullName, registry.RecommendedCommandName)),
-			component.NewCmdTest(component.TestRecommendedCommandName, util.GetFullName(fullName, component.TestRecommendedCommandName)),
-			env.NewCmdEnv(env.RecommendedCommandName, util.GetFullName(fullName, env.RecommendedCommandName)),
-		)
-	}
+	rootCmd.AddCommand(
+		env.NewCmdEnv(env.RecommendedCommandName, util.GetFullName(fullName, env.RecommendedCommandName)),
+	)
 
 	odoutil.VisitCommands(rootCmd, reconfigureCmdWithSubcmd)
 
--- a/pkg/odo/cli/component/common_link.go
+++ b/pkg/odo/cli/component/common_link.go
@@ -12,7 +12,6 @@ import (
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
-	"github.com/openshift/odo/pkg/odo/util/experimental"
 	"github.com/openshift/odo/pkg/secret"
 	svc "github.com/openshift/odo/pkg/service"
 	"github.com/openshift/odo/pkg/util"
@@ -40,6 +39,8 @@ type commonLinkOptions struct {
 	secretName       string
 	isTargetAService bool
 
+	devfilePath string
+
 	suppliedName  string
 	operation     func(secretName, componentName, applicationName string) error
 	operationName string
@@ -57,12 +58,13 @@ func newCommonLinkOptions() *commonLinkOptions {
 
 // Complete completes LinkOptions after they've been created
 func (o *commonLinkOptions) complete(name string, cmd *cobra.Command, args []string) (err error) {
+
 	o.operationName = name
 
 	suppliedName := args[0]
 	o.suppliedName = suppliedName
 
-	if experimental.IsExperimentalModeEnabled() {
+	if util.CheckPathExists(o.devfilePath) {
 		o.Context = genericclioptions.NewDevfileContext(cmd)
 
 		oclient, err := occlient.New()
@@ -168,7 +170,7 @@ func (o *commonLinkOptions) complete(name string, cmd *cobra.Command, args []str
 
 func (o *commonLinkOptions) validate(wait bool) (err error) {
 
-	if experimental.IsExperimentalModeEnabled() {
+	if util.CheckPathExists(o.devfilePath) {
 		// let's validate if the service exists
 		svcFullName := strings.Join([]string{o.serviceType, o.serviceName}, "/")
 		svcExists, err := svc.OperatorSvcExists(o.KClient, svcFullName)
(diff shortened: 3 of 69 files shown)


amitkrout:
I will check it with macOS and windows platform.

tests/integration/cmd_link_unlink_test.go
```diff
@@ -41,14 +41,16 @@ var _ = Describe("odo link and unlink command tests", func() {
 	Context("when running help for link command", func() {
 		It("should display the help", func() {
 			appHelp := helper.CmdShouldPass("odo", "link", "-h")
-			Expect(appHelp).To(ContainSubstring("Link component to a service or component"))
+			// Check for -vmodule moduleSpec output which is in additional flags
+			Expect(appHelp).To(ContainSubstring("--vmodule moduleSpec"))
```
prietyc123, line 45: Little confused with `--vmodule moduleSpec` part. Could you please help me understand why are we verifying the same part for both link and [unlink](https://github.com/openshift/odo/pull/3705/files#diff-93c99b01e2a566d8d27400ffb6e2ee51R53).

I think this is a common part which is reflected everywhere, I mean `odo push -h` will also have this. Correct me if I am wrong. I would suggest lets verify for specific substring for `odo link`

tests/integration/devfile/cmd_devfile_catalog_test.go
```diff
@@ -24,13 +24,19 @@ var _ = Describe("odo devfile catalog command tests", func() {
...
-		helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
 
 		originalKubeconfig = os.Getenv("KUBECONFIG")
 		helper.LocalKubeconfigSet(context)
 		project = cliRunner.CreateRandNamespaceProject()
 		currentWorkingDirectory = helper.Getwd()
 		helper.Chdir(context)
+
+		// !!
+		// REMOVE THIS BEFORE MERGING!
```
prietyc123, line 35: According to the comment could you please remove this

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -44,48 +43,51 @@ var _ = Describe("odo devfile create command tests", func() {
 		os.Unsetenv("GLOBALODOCONFIG")
 	})
 
-	Context("Enabling experimental preference should show a disclaimer", func() {
-		It("checks that the experimental warning appears for create", func() {
-			helper.CopyExample(filepath.Join("source", "nodejs"), context)
+	// These tests no longer matter
```
prietyc123, line 46: Can we remove this here itself if it does not matter.

prietyc123 (changes requested):

tests/integration/cmd_link_unlink_test.go
cdrage: Hey @prietyc123 we can use anything in the example. The only purpose of this test it to make sure that `--help` works. Since `--vmodule moduleSpec` is in pretty much every `--help` output. I used that instead.

pkg/odo/cli/component/push.go
```diff
@@ -176,7 +175,7 @@ func (po *PushOptions) Validate() (err error) {
 	// If the experimental flag is set and devfile is present, then we do *not* validate
 	// TODO: We need to clean this section up a bit.. We should also validate Devfile here
 	// too.
-	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(po.DevfilePath) {
```
johnmcollier, line 179: I'm wondering if we need an additional check here for when the pushtarget is set to Docker, but experimental mode isn't set, and so we error out? Since we're keeping Docker support as experimental.

Some commands, like `odo url create` require both to be set, but others like `odo push` and `odo delete` don't.

pkg/odo/cli/component/create.go
```diff
@@ -307,7 +294,7 @@ func (co *CreateOptions) checkConflictingFlags() (err error) {
 
 func (co *CreateOptions) checkConflictingS2IFlags() error {
 	if !co.forceS2i {
-		errorString := "flag --%s, requires --s2i flag to be set, when in experimental mode."
+		errorString := "flag --%s, requires --s2i flag to be set, when deploying S2I (Source-to-Image)."
```
johnmcollier, line 297: Nit: Should this be `when deploying S2I (Source-to-Image) components`?

pkg/odo/cli/component/common_link.go
```diff
@@ -40,6 +39,8 @@ type commonLinkOptions struct {
 	secretName       string
 	isTargetAService bool
 
+	devfilePath string
```
johnmcollier, line 42: Rather than adding `devfilePath` to each struct for the commands that are missing it or need it, could we add it to something common like `genericclioptions.Context`?

Doing so might require slightly retooling the initilization of the contexts, however.

pkg/component/component.go
```diff
@@ -591,7 +590,7 @@ func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConf
 
 	var componentName string
 	var applicationName string
-	if !isExperimentalModeEnabled || kClient == nil {
+	if isS2I || kClient == nil {
```
johnmcollier, line 593: Nit: Is there a scenario here where `isS2I` could be false, but `kClient` could be nil? (Basically, I'm wondering if we need to check for both, or if just checking for the value of `isS2I` would suffice)

pkg/component/component.go
```diff
@@ -565,10 +564,10 @@ func ensureAndLogProperResourceUsage(resourceMin, resourceMax *string, resourceN
 //	envSpecificInfo: Component environment specific information, available if uses devfile
 //  cmpExist: true if components exists in the cluster
 //  endpointMap: value is devfile endpoint entry, key is the TargetPort for each enpoint entry
+//  isS2I: Legacy option. Set as true if you want to use the old S2I method as it differentiates slightly.
 // Returns:
 //	err: Errors if any else nil
-func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool, endpointMap map[int32]parsercommon.Endpoint) (err error) {
-	isExperimentalModeEnabled := experimental.IsExperimentalModeEnabled()
+func ApplyConfig(client *occlient.Client, kClient *kclient.Client, componentConfig config.LocalConfigInfo, envSpecificInfo envinfo.EnvSpecificInfo, stdout io.Writer, cmpExist bool, endpointMap map[int32]parsercommon.Endpoint, isS2I bool) (err error) {
```
johnmcollier, line 570: `ApplyConfig` has 8 parameters, would it make sense to roll the parameters up in to a struct or just leave as-is?

We're only calling `ApplyConfig` twice, so honestly I'm fine either way, tbh.

tests/integration/cmd_link_unlink_test.go
prietyc123: AFAIU here we are trying to get the examples/help for the `odo link` commands like `odo link <service> ...` . Wouldn't it be the case like somehow `odo link -h` gives you the default `odo -h` description/examples/help then even though the help message points to the wrong one it will pass, right?

.travis.yml
```diff
@@ -63,139 +60,703 @@ jobs:
...
         - ./scripts/oc-cluster.sh
         - make configure-supported-311-is
         - make bin
         - sudo cp odo /usr/bin
         - odo login -u developer
         - travis_wait make test-generic
+        - odo logout
+
+
+    - <<: *base-test
```
prietyc123, line 74: Not sure why we are separating the CI jobs again. @cdrage I think we have already discussed it in pr comment https://github.com/openshift/odo/pull/3776#issuecomment-677433898 . 

Is there any other factor you are considering which may be I am not aware of. Why I am saying this because we are always worried about CI timings and separating jobs increasing it to more than an hour.
image

prietyc123 (changes requested):

pkg/odo/cli/cli.go
```diff
@@ -204,15 +203,13 @@ func odoRootCmd(name, fullName string) *cobra.Command {
...
 
-	if experimental.IsExperimentalModeEnabled() {
-		rootCmd.AddCommand(
-			registry.NewCmdRegistry(registry.RecommendedCommandName, util.GetFullName(fullName, registry.RecommendedCommandName)),
-			component.NewCmdTest(component.TestRecommendedCommandName, util.GetFullName(fullName, component.TestRecommendedCommandName)),
-			env.NewCmdEnv(env.RecommendedCommandName, util.GetFullName(fullName, env.RecommendedCommandName)),
-		)
-	}
+	rootCmd.AddCommand(
+		env.NewCmdEnv(env.RecommendedCommandName, util.GetFullName(fullName, env.RecommendedCommandName)),
```
mik-dass, line 211: we can directly add it after `component.NewCmdTest()` above

pkg/odo/cli/component/exec.go
```diff
@@ -64,7 +64,7 @@ Please provide a command to execute, odo exec -- <command to be execute>`)
 	eo.devfilePath = filepath.Join(eo.componentContext, devFile)
 
 	// if experimental mode is enabled and devfile is present
-	if experimental.IsExperimentalModeEnabled() {
+	if util.CheckPathExists(eo.devfilePath) {
```
mik-dass, line 67: we should inform the user that this command doesn't work with s2i as the error message might be confusing
```
[mrinaldas@localhost ~]$ odo exec --context ~/project  -- ls 
 ✗  open /home/mrinaldas/project/devfile.yaml: no such file or directory
```

pkg/odo/cli/component/link.go
```diff
@@ -115,7 +118,8 @@ func (o *LinkOptions) Validate() (err error) {
 		return err
 	}
 
-	if experimental.IsExperimentalModeEnabled() {
+	// Wat?
```
mik-dass, line 121: I don't think this is required :D

pkg/odo/cli/debug/portforward.go
```diff
@@ -166,7 +165,7 @@ func (o PortForwardOptions) Run() error {
 		return err
 	}
 
-	return o.PortForwarder.ForwardPorts(o.PortPair, o.StopChannel, o.ReadyChannel, o.isExperimental)
+	return o.PortForwarder.ForwardPorts(o.PortPair, o.StopChannel, o.ReadyChannel, util.CheckPathExists(o.devfilePath))
```
mik-dass, line 168: maybe store `util.CheckPathExists(o.devfilePath)` in a variable and pass here

mik-dass:
`odo test` doesn't work with s2i components and we can inform the user with error message that it is not supported for s2i components. The current error might be confusing
```
[mrinaldas@localhost ~]$ odo test --context ~/project
```

adisky:
@cdrage Mostly works for me, tried out various scenarios with `create`, `delete`, `list`, `push`, `log`, `url`, `storage`
some nitpicks that could be resolved in followup PR or an issue could be logged

flags `token` `starter` `registry`  if used with  s2i components during `odo create nodejs --s2i` does not give error.
flags  `build-command` `run-command` `debug-command`  if used with  s2i components during `odo push` does not give error.

`odo url create --ingress` does not error out for s2i components and created a route url

some issues with url and `odo list`  which were already identified and assigned.

prietyc123 (changes requested):
@cdrage I have run this pr on windows and its really great that we came out of Experimental so far. But I suspect the way service catalog has been handled effecting operatorhub services which leads it to fail on all the platform it seems. So far I have observed the failures on linux and windows.
```
> odo preference set Experimental true
> odo project create fqllucxoxm
 V  Project 'fqllucxoxm' is ready for use
 V  New project created and now using project: fqllucxoxm

> oc get csv
NAME                                     DISPLAY                           VERSION             REPLACES                              PHASE
etcdoperator.v0.9.4-clusterwide          etcd                              0.9.4-clusterwide                                         Succeeded
percona-server-mongodb-operator.v1.4.0   Percona Server MongoDB Operator   1.4.0                                                     Succeeded
service-binding-operator.v0.1.1-364      Service Binding Operator          0.1.1-364           service-binding-operator.v0.1.1-354   Succeeded

> odo service create etcdoperator.v0.9.4-clusterwide/EtcdCluster --dry-run
 X  unable to search classes by name (clusterserviceclasses.servicecatalog.k8s.io is forbidden: User "developer" cannot list resource "clusterserviceclasses" in API group "servicecatalog.k8s.io" at the cluster scope)
```
and lots of failure like this which will be observed more broardly in log https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/3705/pull-ci-openshift-odo-master-v4.5-integration-e2e/1301589250180911104

.travis.yml
cdrage: @prietyc123 Yes, I know. The problem is that we keep having Travis flakes. Most likely because of being out of memory int he job, as we are running too many things at once in the Travis VM's. This travis modification is temporary in order for me to diagnose it easier.

cdrage:
> @cdrage Mostly works for me, tried out various scenarios with `create`, `delete`, `list`, `push`, `log`, `url`, `storage`
> some nitpicks that could be resolved in followup PR or an issue could be logged
> 
> flags `token` `starter` `registry` if used with s2i components during `odo create nodejs --s2i` does not give error.
> flags `build-command` `run-command` `debug-command` if used with s2i components during `odo push` does not give error.

Unfortunately these flags should of been covered already with enabling experimental mode. We will have to implement this in a different PR.

> 
> `odo url create --ingress` does not error out for s2i components and created a route url
> 
> some issues i had seen with url and `odo list` which were already identified and assigned.

amitkrout:
I have covered the almost all test target on macOS except service catalog test. Two failure i hit on macOS [1] looks like a test script failure and [2] i need to analyze it further. 

As part of operator tests concerned, i am hitting the same error as mentioned above by @prietyc123 https://github.com/openshift/odo/pull/3705#pullrequestreview-482632213

[1] https://github.com/openshift/odo/issues/3797
[2] https://github.com/openshift/odo/issues/3897

pkg/odo/cli/component/exec.go
cdrage: Ah, you're right. I will add a test case for this.

pkg/odo/cli/component/common_link.go
cdrage: Yeah, we can make this change in a later PR as it'd thorooughly mess up my constant rebasing :( I do know that Chris has done refactoring he'd like to make in terms of Devfile / context usage.

pkg/component/component.go
cdrage: I don't think so as I'm simply copy-and-pasting / replacing each isExperimentalModeEnabled.

pkg/odo/cli/debug/portforward.go
cdrage: Since we're only retrieving this variable once, I think it'd be okay to pass in util.CheckPathExists to ForwardPorts since ForwardPorts is using the parameter `isS2i` which requires true / false.

mik-dass: Actually we are retrieving this twice. First at complete() and then here. Maybe store in a variable like isDevfile once like in the other files and pass it to the function.

pkg/odo/cli/component/common_link.go
johnmcollier: Ah good point, we can handle that either as part of the context refactoring or later on 👍

pkg/component/component.go
johnmcollier: Yup, understood

cdrage:
@dharmit any way you could investigate / give some insight on the operator hub failures?

cdrage:
Thanks @amitkrout @prietyc123 @mik-dass @johnmcollier for reviewing.

I have updated this PR again with this commit: https://github.com/openshift/odo/pull/3705/commits/9d9510a73435387c15a6a55f5d0d5ec3f52d5be7

I have implemented the following changes:
 - Applied Mrinal's changes in moving the adding the test command https://github.com/openshift/odo/pull/3705/commits/9d9510a73435387c15a6a55f5d0d5ec3f52d5be7#diff-ae6461ee1175c7152132051efcd0bbecR208
- Minor grammar / wording changes
- Error out when running `odo exec` with s2i
- @johnmcollier 's changes to registry: https://github.com/openshift/odo/pull/3705/commits/9d9510a73435387c15a6a55f5d0d5ec3f52d5be7#diff-0cd3824178637db7a863b63f797b00d4R60
- Added test scenario for trying to run `odo exec` with an s2i component (it should fail)

prietyc123:
@cdrage Seems like `odo list -o json` not working properly as reported in https://github.com/openshift/odo/issues/3901 .

.travis.yml
amitkrout: I guess there should not be any memory outage because as per the travis documentation [1] the memory allocated to ```dist: xenial``` is ```7.5 GB``` and as part of parallel process is concerned we can run 4 thread max [2] without degrading the travis performance. 

[1] https://docs.travis-ci.com/user/reference/overview/
[2] https://docs.travis-ci.com/user/common-build-problems/#parallel-processes

johnmcollier:
Looks good from my end (minus the conflic

johnmcollier:
Gah, sorry, did not mean to close this!

Looks good from my end

cdrage:
NOTES BEFORE MERGING:

 - `travis.yml` modifications MUST be removed
- https://github.com/openshift/odo/pull/3883 has been integrated into this PR. If this is merged, it will also merge in https://github.com/openshift/odo/pull/3883

cdrage:
NOTES BEFORE MERGING: 

 - `travis.yml` modifications MUST be removed
- https://github.com/openshift/odo/pull/3883 has been integrated into this PR. If this is merged, it will also merge in https://github.com/openshift/odo/pull/3883

dgolovin:
> EDIT: Nevermind, I see that it was tested against the alpha branch, I'll fix it.

@kadel  @cdrage please fix #3901 and #3550 in next release. Those two slows us down to integrate devfile support in IDE extensions. Now we can create components, but IDEs cannot see them because they are not listed in 'odo list',  --path option is not supported to discover them in local file system and 'odo describe' is not supported too.

cdrage:
Re-adding /lgtm since I had to rebase and disable OperatorHub tests.

PR #3715, opened 2020-08-07
Skiping s2iItem check on kubernetes cluster

prietyc123:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind failing-test
/kind code-refactoring

**What does does this PR do / why we need it**:

PR skin s2i image check for kubernetes clusters due to the fact that imagestreams are not there with k8s.
Also code clean up.

**Which issue(s) this PR fixes**:

Fixes #3713 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

**How to test changes / Special notes to the reviewer**:

`make test-cmd-devfile-catalog` should run on both openshift and kubernetes cluster

PR #3717, opened 2020-08-07
Clean up

metacosm:
**What type of PR is this?**
/kind cleanup

Make DevfileData.GetCommands return a map since that's how it's most often used.

N/A

Behavior should be unchanged


pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -11,6 +13,19 @@ func (dc DevfileCommand) GetID() string {
 	return ""
 }
 
+// SetIDToLower converts the command's id to lower case for more efficient processing and returns the new id
+func (dc *DevfileCommand) SetIDToLower() string {
```
johnmcollier, line 17: 👍

pkg/testingutil/devfile.go
```diff
@@ -108,7 +105,21 @@ func (d TestDevfileData) AddComponents(components []common.DevfileComponent) err
...
+	if d.Commands == nil {
+		d.Commands = make(map[string]common.DevfileCommand, 7)
+	}
+
+	for _, command := range commands {
+		id := command.GetID()
+		if _, ok := d.Commands[id]; !ok {
+			d.Commands[id] = command
+		} else {
+			panic("command with id '" + id + "' already exists in this TestDevfileData")
```
johnmcollier, line 118: Why panic here instead of returning a new error?

pkg/devfile/adapters/common/utils_test.go
```diff
@@ -566,14 +566,14 @@ func TestGetCommandsMap(t *testing.T) {
 	for _, tt := range tests {
```
johnmcollier, line 566: Super minor nit: Could we rename the test function from `TestGetCommandsMap` to `TestGetCommands`?

pkg/devfile/parser/data/2.0.0/components.go
```diff
@@ -126,47 +126,41 @@ func (d *Devfile200) UpdateComponent(component common.DevfileComponent) {
...
-			return &common.AlreadyExistError{Name: command.Exec.Id, Field: "command"}
+			return &common.AlreadyExistError{Name: id, Field: "command"}
 		}
 	}
 	return nil
 }
 
 // UpdateCommand updates the command with the given id
 func (d *Devfile200) UpdateCommand(command common.DevfileCommand) {
+	id := strings.ToLower(command.Exec.Id)
```
johnmcollier, line 161: Why not call `command.GetID()` here? Otherwise, should we add a nil check for `command.Exec` and `command.Composite`?

prietyc123:
@metacosm Can you please rebase the pr as travis blocker https://github.com/openshift/odo/pull/3719 and https://github.com/openshift/odo/pull/3715 got merged

pkg/testingutil/devfile.go
metacosm: Because it's not a recoverable error? Trying to add a command with an already existing id in the test data should fail. Is panicking causing undesired side-effects?

pkg/devfile/adapters/common/utils_test.go
metacosm: Sure.

pkg/devfile/parser/data/2.0.0/components.go
metacosm: Agreed, I just ported the existing code, as-is, if I remember correctly.

metacosm: More precisely, the previous code used `command.Exec.Id` to identify which command to update. I guess this function doesn't take into account compound commands. I did wonder about it but I left the behavior unchanged because I'm just refactoring things here, not fixing bugs, though I could certainly change the behavior if we agree that's what we should be doing here. Wdyt, @johnmcollier @mik-dass ?

pkg/devfile/adapters/common/command_test.go
```diff
@@ -2004,7 +2005,9 @@ func TestValidateAndGetTestDevfileCommands(t *testing.T) {
 }
 
 func getExecCommand(id string, group common.DevfileCommandGroupType) versionsCommon.Exec {
-
+	if len(id) == 0 {
+		id, _ = util.GetRandomName("cmd", 10, []string{}, 2)
```
mik-dass, line 2009: we need to handle the error

pkg/testingutil/devfile.go
mik-dass: I think it will cause the test suite to stop and the other tests won't run. Better to return and let the caller handle it.

pkg/devfile/parser/data/2.0.0/components.go
mik-dass: Sure. This function was written before composite commands was implemented. So parent support won't work with composite commands. I will open a new issue for that.

pkg/devfile/adapters/common/command_test.go
metacosm: How exactly do you propose to handle the error? If `GetRandomName` fails, what should be done, especially considering that it can only result in an error if the given string array is not empty?

mik-dass: I looked into the function again and you are right that it can only throw an error when the string array is not empty. I think we can directly use `fmt.Sprintf("%s-%s", "cmd", GenerateRandomString(10))` to avoid the unnecessary inner function calls and checks.

mik-dass:
```
--- FAIL: TestPrintSupportedParameters (0.00s)
    --- FAIL: TestPrintSupportedParameters/Case_1:_Test_print_supported_set_parameters (0.00s)
        env_test.go:49: Got Available parameters:
              DebugPort: Use this value to set component debug port
              Name: Use this value to set component name
              Namespace: Use this value to set component namespace, want Available parameters:
              Name: Use this value to set component name
              Namespace: Use this value to set component namespace
              DebugPort: Use this value to set component debug port
FAIL
```/retest

**Comment by mik-dass** [comment]
```
--- FAIL: TestPrintSupportedParameters (0.00s)
    --- FAIL: TestPrintSupportedParameters/Case_1:_Test_print_supported_set_parameters (0.00s)
        env_test.go:49: Got Available parameters:
              DebugPort: Use this value to set component debug port
              Name: Use this value to set component name
              Namespace: Use this value to set component namespace, want Available parameters:
              Name: Use this value to set component name
              Namespace: Use this value to set component namespace
              DebugPort: Use this value to set component debug port
FAIL
```
/retest

metacosm:
> ```
> --- FAIL: TestPrintSupportedParameters (0.00s)
>     --- FAIL: TestPrintSupportedParameters/Case_1:_Test_print_supported_set_parameters (0.00s)
>         env_test.go:49: Got Available parameters:
>               DebugPort: Use this value to set component debug port
>               Name: Use this value to set component name
>               Namespace: Use this value to set component namespace, want Available parameters:
>               Name: Use this value to set component name
>               Namespace: Use this value to set component namespace
>               DebugPort: Use this value to set component debug port
> FAIL
> ```
> 
> /retest

This looks weird because my understanding of `DeepEqual` on maps should be ordering-independent, while the test, seems to randomly test because of, I assume, ordering… which it shouldn't… So I'm not sure what's going on here.

metacosm:
Actually, that's not even what's going on since `DeepEqual` is actually checking `bool`s and not maps…

metacosm:
> ```
> --- FAIL: TestPrintSupportedParameters (0.00s)
>     --- FAIL: TestPrintSupportedParameters/Case_1:_Test_print_supported_set_parameters (0.00s)
>         env_test.go:49: Got Available parameters:
>               DebugPort: Use this value to set component debug port
>               Name: Use this value to set component name
>               Namespace: Use this value to set component namespace, want Available parameters:
>               Name: Use this value to set component name
>               Namespace: Use this value to set component namespace
>               DebugPort: Use this value to set component debug port
> FAIL
> ```
> 
> /retest

This test is actually wrong and unstable: `printSupportedParameters` iterates over a map which doesn't preserve ordering so expecting a given output based on the supposed map ordering is wrong. This will at best pass once in a while, not always.

mik-dass:
> This test is actually wrong and unstable: `printSupportedParameters` iterates over a map which doesn't preserve ordering so expecting a given output based on the supposed map ordering is wrong. This will at best pass once in a while, not always.

Yes. I have raised a PR for it https://github.com/openshift/odo/pull/3760

metacosm:
@mik-dass @kadel can I get the approve / lgtm labels back, please?

kadel:
/hold
There is a lot of test failures due to timeouts. We should verify if this is CI issue or something that might get introduced in this PR.

Issue #2540, opened 2020-01-29
Run e2e and integration test on windows and macOS host


prietyc123:
[kind/Feature]

## Which functionality do you think we should add?

We can borrow macOS host on travis and windows from appveyor then only we can start our tests on these hosts.

## Why is this needed?

Not yet implemented in CI.

 - [ ] Integration tests
     - [ ] Mac OS https://github.com/redhat-developer/odo/issues/5301
     - [x] Windows https://github.com/redhat-developer/odo/issues/5300
 - [ ] e2e tests
     - [ ] Mac OS https://github.com/redhat-developer/odo/issues/5301
     - [x] Windows https://github.com/redhat-developer/odo/issues/5300

amitkrout:
@mohammedzee1000 Can we please add those inputs that we discussed yesterday and before. @prietyc123 FYI...

mohammedzee1000:
We can use travis.rb which is the official cli for travis ci or we can directly use the API.

The main drawbacks of travis.rb is we cannot create job manually, just restarting existing ones and also while we can set envs, we cannot do it per build

So i propose:

- configure travis job which looks for env and if not present fails
- prow job starts and brings up cluster
- prow job passes on cluster url to travis job with env and restarts it

mohammedzee1000:
https://github.com/travis-ci/travis.rb
The problematic bit where dirty reads can occur is https://github.com/travis-ci/travis.rb#env and we should plan around that.

For example after cluster is brought up, the prow job goes into a loop, where it injects env say `CLUSTER_URL=10;example.com` (where 10 is PR no) and restarts the job, then waits for build to become a success. If any builds fail then it restarts the same for say 3 times after which it is assumed as failed forever (basically full retrigger required).

Travis job on its part reads `CLUSTER_URL="10;example.com"` reads pr no and if it does not match, it fails. Otherwise, it reads the cluster URL and stores it in different env and resumes tests

@amitkrout @girishramnani  wdyt

mohammedzee1000:
https://github.com/travis-ci/travis.rb/issues/717

mohammedzee1000:
```
❯ travis env set FOO BAR
[+] setting environment variable $FOO
❯ travis env list
# environment variables for mohammedzee1000/odo
FOO=[secure]


```

mohammedzee1000:
Steve said:
Prow provides these env vars, so you could get the PR number by:
[...]

PR #3732, opened 2020-08-11
Docker test container for travis login

amitkrout:
**What type of PR is this?**
/kind Feature

we are in process of getting the odo-bot GitHub token injected into prow. So creating a test container from where all these short of operation like travis cli install, travis login... etc will be launched 

Fixes : part of #2540 

**PR acceptance criteria**:

Should not break master

```
$ docker run -it registry.svc.ci.openshift.org/openshift/release:golang-1.13
$ yum -y install sudo make wget gcc git httpd-tools ruby ruby-devel gcc-c++ patch \
    readline readline-devel zlib zlib-devel libffi-devel openssl-devel make bzip2 \
    autoconf automake libtool bison sqlite-devel
$ sudo curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
$ sleep 5
$ sudo curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -
$ curl -L get.rvm.io | bash -s stable
$ source /etc/profile.d/rvm.sh
$ rvm reload
$ rvm requirements run
$ rvm install 2.7
$ gem install travis --pre --no-document
```

### DIFF
--- a/openshift-ci/build-root/Dockerfile.travislogin
+++ b/openshift-ci/build-root/Dockerfile.travislogin
@@ -1,0 +1,23 @@
+# Dockerfile to bootstrap build and test in openshift-ci
+
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
+
+RUN yum -y install sudo make wget gcc git httpd-tools ruby ruby-devel gcc-c++ patch \
+    readline readline-devel zlib zlib-devel libffi-devel openssl-devel make bzip2 \
+    autoconf automake libtool bison sqlite-devel
+
+RUN sudo curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
+RUN sleep 5
+RUN sudo curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -
+
+RUN curl -L get.rvm.io | bash -s stable
+
+RUN source /etc/profile.d/rvm.sh \
+    rvm reload
+
+RUN rvm requirements run
+RUN rvm install 2.7
+
+RUN ruby --version
+
+RUN gem install travis --pre --no-document


openshift-ci/build-root/Dockerfile.travislogin
```diff
@@ -0,0 +1,23 @@
+# Dockerfile to bootstrap build and test in openshift-ci
+
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
```
prietyc123, line 3: Instead of fetching image stream again, I would say let's implement it as https://github.com/openshift/odo/blob/master/openshift-ci/build-root/source-image/Dockerfile#L1

amitkrout: We can not create a separate test container with out pulling the image. BTW the travis test container has certain steps to follow and i do not pollute the regular test container with the travis steps.

Ofcourse as per the requirement i may need https://github.com/openshift/odo/blob/master/openshift-ci/build-root/source-image/Dockerfile#L1 in my corresponding release job pr for copying the oc binary.

prietyc123: > BTW the travis test container has certain steps to follow and i do not pollute the regular test container with the travis steps.

make sense 👍 

But I would say once you get the overall workflow to do this, please ensure if we can have only one dockerfile for build. Other than that I am good with the changes.

/lgtm

amitkrout: Corresponding ci pr created - https://github.com/openshift/release/pull/10942

amitkrout: > But I would say once you get the overall workflow to do this, please ensure if we can have only one dockerfile for build.

@prietyc123 AFAIU It is not even possible because if i am installing packages in the source image for example https://github.com/openshift/odo/blob/master/openshift-ci/build-root/source-image/Dockerfile#L1 then it won't be propagated to the main build container. As per my requirement i do not need binary copy paste through the docker file, i need to install the packages for travis login in the main test container.

prietyc123:
> New changes are detected. LGTM label has been removed.

/lgtm

mohammedzee1000 (approved):
looking good
/approve

amitkrout:
Please do not restart travis job. I am just experimenting the travis status is captured by prow bot or not

PR #3740, opened 2020-08-11
Stop other commands in case of restart false

### LINKED ISSUE #3738: Devfile mode: the run command process isn't killed, before starting the debug command process, if 'restart' attribute is set as false
Pre requisite knowledge :  

the **restart** attribute, as illustrated [here](https://github.com/odo-devfiles/registry/blob/78f685b7476791d9e45a0c9d6bbb8398b93d0216/devfiles/java-quarkus/devfile.yaml#L36)  , enables (when set as true) or disbles (when set as false) the retriggering of a project's build cycle inside the deployed container when `odo push` is used and newly added or edited files are pushed inside the running app container.

#### Case when `restart` attribute to set as true

Using a nodejs project while using the nodejs devfile from the default registry

https://github.com/odo-devfiles/registry/blob/78f685b7476791d9e45a0c9d6bbb8398b93d0216/devfiles/nodejs/devfile.yaml
 
if you see the logs here, it is ensured that the running commands are forcefully killed before the debug command is started. 
image

#### Case when `restart` attribute to set as false

Using a quarkus project while using this devfile.

```
--- 
  commands: 
    - 
      exec: 
        commandLine: "mvn compile"
        component: tools
        id: init-compile
        workingDir: $PROJECTS_ROOT
    - 
      exec:
        attributes:
          restart: "false"
        commandLine: "mvn quarkus:dev"
        component: tools
        group: 
          isDefault: true
          kind: run
        id: dev-run
        workingDir: $PROJECTS_ROOT
    - 
      exec:
        commandLine: "mvn quarkus:dev -Ddebug=${DEBUG_PORT}"
        component: tools
        group: 
          isDefault: true
          kind: debug
        id: dev-debug
        workingDir: $PROJECTS_ROOT
  components: 
    - 
      container: 
        endpoints: 
          - 
            name: 8080/http
            targetPort: 8080
        image: "quay.io/eclipse/che-quarkus:nightly"
        memoryLimit: 1512Mi
        mountSources: true
        name: tools
        volumeMounts: 
          - 
            name: m2
            path: /home/user/.m2
    - 
      volume: 
        name: m2
        size: 3Gi
  events: 
    postStart: 
      - init-compile
  metadata: 
    name: java-quarkus
    version: "1.0.0"
    website: "https://quarkus.io"
  projects: 
    - 
      git: 
        location: "https://github.com/odo-devfiles/quarkus-ex"
      name: quarkus-ex
  schemaVersion: "2.0.0"  
```

if you see the logs here, the process started by the run command is not killed before starting the debug command process. 
image (3)

This leads to an error as the port occupied from the run command is not released for the debug command to start properly.

BUT

When I set the restart attribute as true, the ongoing processes are killed before the debug command starts. Check the logs...

image

adisky:
stop debug command, if we are running run command and vice versa

**What type of PR is this?**
 /kind bug

to restart any command in supervisord the way is to stop it and start it, in case of `restart:false` we are omitting the stop supervisord command step. It creates a problem when user switch modes during `odo push` (run/debug). This PR updates in case of `restart:false` we stop other command, like if it is `odo push --debug`,  we stop run command. 

Fixes #3738 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/devfile/adapters/common/command_supervisor.go
+++ b/pkg/devfile/adapters/common/command_supervisor.go
@@ -32,12 +32,9 @@ func newSupervisorInitCommand(command common.DevfileCommand, adapter commandExec
 }
 
 // newSupervisorStopCommand creates a command implementation that stops the specified command via the supervisor
-func newSupervisorStopCommand(command common.DevfileCommand, cmd string, executor commandExecutor) (command, error) {
-	if cmd == "" {
-		cmd = "all"
-	}
-	cmdLine := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", cmd}
-	if stop, err := newOverriddenSimpleCommand(command, executor, cmdLine); err == nil {
+func newSupervisorStopCommand(command common.DevfileCommand, executor commandExecutor) (command, error) {
+	cmd := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", "all"}
+	if stop, err := newOverriddenSimpleCommand(command, executor, cmd); err == nil {
 		// use empty spinner message to avoid showing it altogether
 		stop.msg = ""
 		return stop, err
--- a/pkg/devfile/adapters/common/command_supervisor.go
+++ b/pkg/devfile/adapters/common/command_supervisor.go
@@ -32,9 +32,12 @@ func newSupervisorInitCommand(command common.DevfileCommand, adapter commandExec
 }
 
 // newSupervisorStopCommand creates a command implementation that stops the specified command via the supervisor
-func newSupervisorStopCommand(command common.DevfileCommand, executor commandExecutor) (command, error) {
-	cmd := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", "all"}
-	if stop, err := newOverriddenSimpleCommand(command, executor, cmd); err == nil {
+func newSupervisorStopCommand(command common.DevfileCommand, cmd string, executor commandExecutor) (command, error) {
+	if cmd == "" {
+		cmd = "all"
+	}
+	cmdLine := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", cmd}
+	if stop, err := newOverriddenSimpleCommand(command, executor, cmdLine); err == nil {
 		// use empty spinner message to avoid showing it altogether
 		stop.msg = ""
 		return stop, err
--- a/pkg/devfile/adapters/common/generic.go
+++ b/pkg/devfile/adapters/common/generic.go
@@ -1,13 +1,14 @@
 package common
 
 import (
+	"io"
+	"strings"
+
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/machineoutput"
 	"github.com/pkg/errors"
-	"io"
 	"k8s.io/klog"
-	"strings"
 )
 
 // ComponentInfoFactory defines a type for a function which creates a ComponentInfo based on the information provided by the specified DevfileCommand.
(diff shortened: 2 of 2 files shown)


pkg/devfile/adapters/common/command_supervisor.go
```diff
@@ -32,9 +32,12 @@ func newSupervisorInitCommand(command common.DevfileCommand, adapter commandExec
 }
 
 // newSupervisorStopCommand creates a command implementation that stops the specified command via the supervisor
-func newSupervisorStopCommand(command common.DevfileCommand, executor commandExecutor) (command, error) {
-	cmd := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", "all"}
-	if stop, err := newOverriddenSimpleCommand(command, executor, cmd); err == nil {
+func newSupervisorStopCommand(command common.DevfileCommand, cmd string, executor commandExecutor) (command, error) {
```
dev-gaur, line 35: The parameter naming for this method can be improved. first parameter is "command" and the second one is "cmd", which is just short for "command". Will be better to address this and add a small description about the function parameters in the function description comment above.

pkg/devfile/adapters/common/generic.go
```diff
@@ -140,13 +141,34 @@ func (a GenericAdapter) ExecDevfile(commandsMap PushCommandsMap, componentExists
...
-			if cmd, err := newSupervisorStopCommand(command, a); cmd != nil {
+			if cmd, err := newSupervisorStopCommand(command, "", a); cmd != nil {
 				if err != nil {
 					return err
 				}
 				commands = append(commands, cmd)
 			}
 		} else {
+			// if there is switch of commands from run<->, we need to stop the other command
+			var otherCommand common.DevfileCommand
```
mik-dass, line 152: I think we can just check against the env's `runMode` and detect the switch between the commands.

pkg/devfile/adapters/common/command_supervisor.go
```diff
@@ -32,9 +32,12 @@ func newSupervisorInitCommand(command common.DevfileCommand, adapter commandExec
...
+	if cmd == "" {
+		cmd = "all"
+	}
+	cmdLine := []string{SupervisordBinaryPath, SupervisordCtlSubCommand, "stop", cmd}
```
mik-dass, line 39: Why not stop all and restart?

pkg/devfile/adapters/common/generic.go
dev-gaur: What if the user tweaked the env file not knowing what they are doing?

mik-dass: We don't recommend the user to tweak odo related files. Even if the user changes the namespace to a non existent namespace, it will fail too. If the user is tweaking the env file, the failure his/her responsibility.

pkg/devfile/adapters/common/command_supervisor.go
adisky: we do not want to stop all, if  `restart:false`, 
`odo push`  
`odo push` // do not stop run command here 
`odo push --debug` // stop run command here
`odo push`// stop debug command here

pkg/devfile/adapters/common/generic.go
adisky: @mik-dass thanks for your suggestion, updated.

pkg/devfile/adapters/common/generic.go
```diff
@@ -137,15 +145,27 @@ func (a GenericAdapter) ExecDevfile(commandsMap PushCommandsMap, componentExists
...
-		if componentExists && IsRestartRequired(command) {
+		if componentExists && restart {
 			klog.V(4).Infof("restart:true, restarting %s", defaultCmd)
 			if cmd, err := newSupervisorStopCommand(command, a); cmd != nil {
 				if err != nil {
 					return err
 				}
 				commands = append(commands, cmd)
 			}
+		} else if componentExists && previousMode != currentMode && !restart {
```
mik-dass, line 159: I think we can simplify the condition
```
if componentExists {
   if restart || previousMode != currentMode {
     // stop supervisord
  }
}

adisky: Done

pkg/devfile/adapters/common/command_supervisor.go
adisky: now no need

PR #3760, opened 2020-08-17
Updates test for printSupportedParameters to fix flake

### LINKED ISSUE #3755: UTs fails within env folder
/kind flake

**Operating System:**
Windows, Linux

**Output of `odo version`:**
master

make test

## Actual behavior
Observed in travis CI
```
?   	github.com/openshift/odo/pkg/odo/cli/component/ui	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/config	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/debug	[no test files]
--- FAIL: TestPrintSupportedParameters (0.00s)
    --- FAIL: TestPrintSupportedParameters/Case_1:_Test_print_supported_set_parameters (0.00s)
        env_test.go:49: Got Available parameters:
              Namespace: Use this value to set component namespace
              DebugPort: Use this value to set component debug port
              Name: Use this value to set component name, want Available parameters:
              Name: Use this value to set component name
              Namespace: Use this value to set component namespace
              DebugPort: Use this value to set component debug port
FAIL
FAIL	github.com/openshift/odo/pkg/odo/cli/env	0.333s
?   	github.com/openshift/odo/pkg/odo/cli/login	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/logout	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/plugins	[no test files]
```
## Expected behavior
No such flake should appear

mik-dass:
**What type of PR is this?**

/kind failing-test

The printing order in printSupportedParameters() is dependent on a for loop over a map which is random in nature. Thus the unit test should not check for a complete match.

Fixes #3755

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Unit test for printSupportedParameters() should pass without flakes.

### DIFF
--- a/pkg/odo/cli/env/env_test.go
+++ b/pkg/odo/cli/env/env_test.go
@@ -45,8 +45,20 @@ func TestPrintSupportedParameters(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := strings.TrimSpace(printSupportedParameters(tt.supportedParameters))
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Got %s, want %s", got, tt.want)
+
+			gotStrings := strings.Split(got, "\n")
+			wantStrings := strings.Split(tt.want, "\n")
+
+			gotStringMap := make(map[string]bool)
+			for _, gotString := range gotStrings {
+				gotStringMap[gotString] = true
+			}
+
+			for _, wantString := range wantStrings {
+				_, found := gotStringMap[wantString]
+				if !found {
+					t.Errorf("String %s not found in output %s", wantString, got)
+				}
 			}
 		})
 	}
--- a/pkg/odo/cli/env/env_test.go
+++ b/pkg/odo/cli/env/env_test.go
@@ -2,6 +2,7 @@ package env
 
 import (
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 )
@@ -49,16 +50,11 @@ func TestPrintSupportedParameters(t *testing.T) {
 			gotStrings := strings.Split(got, "\n")
 			wantStrings := strings.Split(tt.want, "\n")
 
-			gotStringMap := make(map[string]bool)
-			for _, gotString := range gotStrings {
-				gotStringMap[gotString] = true
-			}
+			sort.Strings(gotStrings)
+			sort.Strings(wantStrings)
 
-			for _, wantString := range wantStrings {
-				_, found := gotStringMap[wantString]
-				if !found {
-					t.Errorf("String %s not found in output %s", wantString, got)
-				}
+			if !reflect.DeepEqual(wantStrings, gotStrings) {
+				t.Errorf("\nGot: %s\nWant: %s", got, tt.want)
 			}
 		})
 	}


pkg/odo/cli/env/env_test.go
```diff
@@ -45,8 +45,20 @@ func TestPrintSupportedParameters(t *testing.T) {
...
+
+			gotStrings := strings.Split(got, "\n")
+			wantStrings := strings.Split(tt.want, "\n")
+
+			gotStringMap := make(map[string]bool)
+			for _, gotString := range gotStrings {
+				gotStringMap[gotString] = true
+			}
+
+			for _, wantString := range wantStrings {
```
metacosm, line 57: Did you try to use `DeepEqual` instead? Because, technically, you should also check that there is no extra strings in `gotStrings`, which would be wrong (i.e. there is a parameter in `gotStrings` that is not supposed to be there, which this doesn't check.

pkg/odo/cli/env/env_test.go
```diff
@@ -45,8 +45,20 @@ func TestPrintSupportedParameters(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := strings.TrimSpace(printSupportedParameters(tt.supportedParameters))
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Got %s, want %s", got, tt.want)
+
```
metacosm, line 48: Here's how I would implement it, which I think is more correct:

```suggestion
			gotSet, gotString := getAsSetAndRecomposedString(got)
			wantedSet, wantedString := getAsSetAndRecomposedString(tt.want)
			if !reflect.DeepEqual(wantedSet, gotSet) {
				t.Errorf("Got %s, want %s", gotString, wantedString)
			}
		})
	}
}

func getAsSetAndRecomposedString(output string) (map[string]bool, string) {
	gotStrings := strings.Split(output, "\n")
	gotSet := make(map[string]bool, len(gotStrings))
	for _, s := range gotStrings {
		gotSet[s] = true
	}
	return gotSet, strings.Join(gotStrings, "\n")
}
```

pkg/odo/cli/env/env_test.go
mik-dass: Thanks :+1:  I missed that. I have updated the logic to use DeepEqual and sorting.

pkg/odo/cli/env/env_test.go
mik-dass: I have updated the logic to use DeepEqual and sorting instead.

pkg/odo/cli/env/env_test.go
```diff
@@ -45,8 +46,15 @@ func TestPrintSupportedParameters(t *testing.T) {
...
+			gotStrings := strings.Split(got, "\n")
+			wantStrings := strings.Split(tt.want, "\n")
+
+			sort.Strings(gotStrings)
```
metacosm, line 53: That works, too. 👍 
And I was actually thinking that maybe that's what `printSupportedParameters` should do (sort by parameter name) as well to get consistent output.

mik-dass: I thought about that. But I didn't want to modify the actual function. So left it unchanged for now.

PR #3762, opened 2020-08-17
Remove unnecessary devfile flag and update tests

### LINKED ISSUE #3661: Remove --devfile parameter from non `odo create` commands.
/kind feature

See discussion: https://openshiftdo.slack.com/archives/CQK6HE78S/p1596118006155100

We have 4 locations where `--devfile` is used outside of `odo create`. `--devfile` should not be located on these commands:

- odo exec
- odo url delete
- odo debug port-forward
- odo debug info

## Why is this needed?

There is no need to pass in `--devfile` if the default name is always `devfile.yaml` in the working directory.

maysunfaisal:

**What type of PR is this?**

/kind bug
/kind cleanup

- only `odo create` needs the devfile flag to copy contents from the specified flag to project dir/context dir devfile.yaml
- devfile is always devfile.yaml, so no point in having a devfile flag for non create commands
- commands listed in the issue do not require --devfile flag
- there were some bugs when using --context and --devfile flags for the command listed, they have been fixed
- updated integration tests

Fixes #3661 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- Run the commands with the devfile flag; it should err out
- Run the commands with the context flag, it should work

### DIFF
--- a/pkg/odo/cli/component/exec.go
+++ b/pkg/odo/cli/component/exec.go
@@ -61,7 +61,8 @@ Please provide a command to execute, odo exec -- <command to be execute>`)
 		return fmt.Errorf("no parameter is expected for the command")
 	}
 
-	eo.devfilePath = filepath.Join(eo.componentContext, eo.devfilePath)
+	eo.devfilePath = filepath.Join(eo.componentContext, devFile)
+
 	// if experimental mode is enabled and devfile is present
 	if experimental.IsExperimentalModeEnabled() {
 		eo.componentOptions.Context = genericclioptions.NewDevfileContext(cmd)
@@ -99,8 +100,6 @@ func NewCmdExec(name, fullName string) *cobra.Command {
 			genericclioptions.GenericRun(o, cmd, args)
 		},
 	}
-
-	execCmd.Flags().StringVar(&o.devfilePath, "devfile", "./devfile.yaml", "Path to a devfile.yaml")
 
 	execCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
 	completion.RegisterCommandHandler(execCmd, completion.ComponentNameCompletionHandler)
--- a/pkg/odo/cli/debug/info.go
+++ b/pkg/odo/cli/debug/info.go
@@ -2,26 +2,25 @@ package debug
 
 import (
 	"fmt"
+
 	"github.com/openshift/odo/pkg/debug"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/machineoutput"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
 	"github.com/openshift/odo/pkg/odo/util/experimental"
-	"github.com/openshift/odo/pkg/util"
 	"github.com/spf13/cobra"
 	k8sgenclioptions "k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/kubectl/pkg/util/templates"
 )
 
-// PortForwardOptions contains all the options for running the port-forward cli command.
+// InfoOptions contains all the options for running the info cli command.
 type InfoOptions struct {
 	componentName   string
 	applicationName string
 	Namespace       string
 	PortForwarder   *debug.DefaultPortForwarder
 	*genericclioptions.Context
-	contextDir  string
-	DevfilePath string
+	contextDir string
 }
 
 var (
@@ -46,7 +45,7 @@ func NewInfoOptions() *InfoOptions {
 
 // Complete completes all the required options for port-forward cmd.
 func (o *InfoOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
+	if experimental.IsExperimentalModeEnabled() {
 		o.Context = genericclioptions.NewDevfileContext(cmd)
 
 		// a small shortcut
@@ -104,9 +103,6 @@ func NewCmdInfo(name, fullName string) *cobra.Command {
 		},
 	}
 	genericclioptions.AddContextFlag(cmd, &opts.contextDir)
-	if experimental.IsExperimentalModeEnabled() {
-		cmd.Flags().StringVar(&opts.DevfilePath, "devfile", "./devfile.yaml", "Path to a devfile.yaml")
-	}
 
 	return cmd
 }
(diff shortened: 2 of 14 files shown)


maysunfaisal:
@cdrage @GeekArthur can you help review this PR? Thx

pkg/odo/cli/url/delete.go
```diff
@@ -41,7 +40,7 @@ func NewURLDeleteOptions() *URLDeleteOptions {
 // Complete completes URLDeleteOptions after they've been Deleted
 func (o *URLDeleteOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
 
-	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
```
johnmcollier, line 44: I think we should still check for the existence of a devfile in the current context. Otherwise, this'll break `odo delete` for s2i components when experimental mode is enabled, fyi (currently, if experimental mode is enabled, we fall back to s2i support for commands when there's no devfile).

pkg/odo/cli/component/exec.go
```diff
@@ -61,7 +61,8 @@ Please provide a command to execute, odo exec -- <command to be execute>`)
 		return fmt.Errorf("no parameter is expected for the command")
 	}
 
-	eo.devfilePath = filepath.Join(eo.componentContext, eo.devfilePath)
+	eo.devfilePath = filepath.Join(eo.componentContext, devFile)
```
mik-dass, line 64: I think we should be using `DevfilePath` from create.go

pkg/odo/cli/debug/info.go
```diff
@@ -46,7 +45,7 @@ func NewInfoOptions() *InfoOptions {
 
 // Complete completes all the required options for port-forward cmd.
 func (o *InfoOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	if experimental.IsExperimentalModeEnabled() && util.CheckPathExists(o.DevfilePath) {
```
mik-dass, line 49: Why was the other check removed? This will trigger `NewDevfileContext` even for s2i components.
```
[mrinaldas@localhost project]$ odo create nodejs --s2i

Validation
 ✓  Validating component [2s]

[mrinaldas@localhost project]$ odo debug info
```

pkg/odo/cli/debug/portforward.go
```diff
@@ -76,7 +76,7 @@ func (o *PortForwardOptions) Complete(name string, cmd *cobra.Command, args []st
 
 	o.isExperimental = experimental.IsExperimentalModeEnabled()
 
-	if o.isExperimental && util.CheckPathExists(o.DevfilePath) {
```
mik-dass, line 79: why was the other check removed?

pkg/odo/cli/url/delete.go
mik-dass, line 44: removing this check will make `odo url delete` not work for s2i components
```
[mrinaldas@localhost project]$ odo create nodejs --s2i

Validation
 ✓  Validating component [1s]

[mrinaldas@localhost project]$ odo url create example
 ✓  URL example created for component: nodejs-project-vwui

[mrinaldas@localhost project]$ odo url delete example
```

tests/integration/devfile/cmd_devfile_debug_test.go
```diff
@@ -67,20 +67,20 @@ var _ = Describe("odo devfile debug command tests", func() {
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), projectDirPath)
 			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile-with-debugrun.yaml"), filepath.Join(projectDirPath, "devfile-with-debugrun.yaml"))
 			helper.RenameFile("devfile-with-debugrun.yaml", "devfile.yaml")
-			helper.CmdShouldPass("odo", "push", "--debug")
```
mik-dass, line 70: since we are using context, we should also remove `helper.Chdir(context)` from beforeEach

pkg/odo/cli/debug/info.go
maysunfaisal: I didnt realize we had a `s2i` flag for experimental, I was wondering what the point of the check was. Thanks for showing it out, will update it!

pkg/odo/cli/component/exec.go
maysunfaisal: yeah i thought about it, but in create.go
`var DevfilePath = filepath.Join(LocalDirectoryDefaultLocation, devFile)`

and its only changing in Complete():
```
if co.componentContext != "" {
			DevfilePath = filepath.Join(co.componentContext, devFile)
```

which is correct since odo commands can run with any context during their execution.

pkg/odo/cli/debug/portforward.go
maysunfaisal: updated

pkg/odo/cli/url/delete.go
maysunfaisal: updated

tests/integration/devfile/cmd_devfile_debug_test.go
maysunfaisal: it doesn't matter because the context path here is the full path to the component but nevertheless i have updated it.

mik-dass: > it doesn't matter because the context path here is the full path to the component but nevertheless if i remove helper.Chdir(context), I need to update a bunch of statements because we're copying devfiles from the repo to the context.

We have had problems before where the context flag wasn't working but it was never detected because we were changing into the context directory. The current directory might be used even though the context was passed.

>  We should be updating all the tests where we use context to not chdir to the dir if we want to do it!

Yes, we need this behavior to change and need a issue to track it.

maysunfaisal: I have opened an issue here https://github.com/openshift/odo/issues/3801 and yes ideally we should not do it

pkg/odo/cli/component/exec.go
mik-dass: OK

tests/integration/devfile/cmd_devfile_debug_test.go
mik-dass: Thanks 👍

mik-dass (approved):
/lgtm

Issue #3578, opened 2020-07-15
Move container image used in springboot devfile to some odo owned image repository


adisky:
The image we are using in java-springboot devfile is in @maysunfaisal 's personal image repository, it should be moved to some odo owned repository
https://github.com/odo-devfiles/registry/blob/master/devfiles/java-springboot/devfile.yaml#L12

adisky:
Got image pull error whole day
```
  Normal   Started           5m26s                kubelet, crc-45nsk-master-0  Started container copy-supervisord
  Warning  Failed            118s                 kubelet, crc-45nsk-master-0  Failed to pull image "maysunfaisal/springbootbuild": rpc error: code = Unknown desc = Error writing blob: error storing blob to file "/var/tmp/storage760563944/4": read tcp 192.168.130.11:46146->104.18.124.25:443: read: connection reset by peer
  Normal   Pulling           47s (x2 over 5m26s)  kubelet, crc-45nsk-master-0  Pulling image "maysunfaisal/springbootbuild"
  Warning  Failed            47s                  kubelet, crc-45nsk-master-0  Error: ErrImagePull
  Warning  Failed            47s                  kubelet, crc-45nsk-master-0  Failed to pull image "maysunfaisal/springbootruntime": rpc error: code = Unknown desc = Error writing blob: error storing blob to file "/var/tmp/storage1863
```

girishramnani:
I agree that we need to move from personal image registries but the reason for that is so that we can productise them ( control changes to them ). 
Not sure if being in mayson's personal registry would give you this image pull error?

adisky:
> I agree that we need to move from personal image registries but the reason for that is so that we can productise them ( control changes to them ).
> Not sure if being in mayson's personal registry would give you this image pull error?

I thought maysun might restricted access to his repo :smile:, I just wanted to point here with personal repositories if you get any error you cannot check the reason behind it. I get image pull error from quay also but then you get the status it is down

maysunfaisal:
AFAIK my dockerhub springboot image is not restricted.

I can give you the dockerfiles to build and push to a registry of your choosing, or you can just pull my image, tag and push to the registry of your choosing 🤔

adisky:
> > I agree that we need to move from personal image registries but the reason for that is so that we can productise them ( control changes to them ).
> > Not sure if being in mayson's personal registry would give you this image pull error?
> 
[...]

Issue #2889, opened 2020-04-15
odo push gets succeded even if Devfile build command fails


adisky:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**

## How did you run odo exactly?
```
$ odo create java-spring-boot
..........

$ odo push  -f 
 •  Push devfile component java-spring-boot  ...

Applying URL changes
 ✓  URL java-spring-boot-8080 already exists
 ✓  Waiting for component to start [1ms]
 ✓  Syncing files to the component [135ms]
 ✓  Executing devbuild command "/artifacts/bin/build-container-full.sh" [1s]
 ✓  Executing devrun command "/artifacts/bin/start-server.sh" [132ms]
 ✓  Push devfile component java-spring-boot [1s]
 ✓  Changes successfully pushed to component
```

## Actual behavior
build command failed inside the container, still odo shows it as successfull
```
1000540000@java-spring-boot-648c9ffd6b-l7hjm:/artifacts/bin$ ./build-container-full.sh 
Wed Apr 15 12:05:41 UTC 2020
Started - Full build using container folders
listing project src
total 8
drwxr-xr-x. 2 root root  60 Apr  9 22:38 .
drwxr-xr-x. 3 root root  17 Apr  9 22:38 ..
-rwxr-xr-x. 1 root root 527 Apr  9 22:38 build-container-full.sh
-rwxr-xr-x. 1 root root 593 Apr  9 21:12 start-server.sh
Wed Apr 15 12:05:41 UTC 2020
running full maven build
POM file ./pom.xml specified with the -f/--file command line argument does not exist
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-readable POM /artifacts/bin/./pom.xml: /artifacts/bin/./pom.xml (No such file or directory) @ 
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project  (/artifacts/bin/./pom.xml) has 1 error
[ERROR]     Non-readable POM /artifacts/bin/./pom.xml: /artifacts/bin/./pom.xml (No such file or directory)
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
```
## Expected behavior
odo push, should show some info about failed devbuild, devrun command.

johnmcollier:
[...]

PR #3799, opened 2020-08-20
Update springboot devfile

adisky:
**What type of PR is this?**

 /kind code-refactoring

Fixes:
partially fixes: #3578

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/examples/source/devfiles/springboot/devfile.yaml
+++ b/tests/examples/source/devfiles/springboot/devfile.yaml
@@ -6,18 +6,18 @@ starterProjects:
   - name: springbootproject
     git:
       remotes:
-        origin: "https://github.com/maysunfaisal/springboot.git"
+        origin: "https://github.com/odo-devfiles/springboot-ex.git"
 components:
   - name: tools
     container:
-      image: maysunfaisal/springbootbuild
+      image: quay.io/eclipse/che-java11-maven:nightly
       memoryLimit: 768Mi
       command: ['tail']
       args: [ '-f', '/dev/null']
       mountSources: true
   - name: runtime
     container:
-      image: maysunfaisal/springbootruntime
+      image: quay.io/eclipse/che-java11-maven:nightly
       memoryLimit: 768Mi
       endpoints:
         - name: "8080/tcp"
@@ -25,23 +25,24 @@ components:
       volumeMounts:
         - name: springbootpvc
           path: /data
-      mountSources: false
+      mountSources: true
   - name: springbootpvc
-    volume: {}
+    volume:
+      size: 3Gi
 commands:
   - id: defaultBuild
     exec:
       component: tools
-      commandLine: mvn install
+      commandLine: "mvn clean -Dmaven.repo.local=/data/cache/.m2/repository package -Dmaven.test.skip=true"
       workingDir: /projects
       group:
         kind: build
-        isDefault: true
   - id: defaultRun
     exec:
       component: runtime
-      commandLine: "java -jar target/*.jar"
+      commandLine: "mvn -Dmaven.repo.local=/data/cache/.m2/repository spring-boot:run"
       workingDir: /projects
       group:
         kind: run
         isDefault: true
+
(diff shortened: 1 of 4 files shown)


girishramnani:
/approve
Haven’t checked if there are other instances of files needing update

adisky:
/hold
wanted the registry one to get reviewed first, would update this one accordingly
https://github.com/odo-devfiles/registry/pull/38/files

prietyc123:
@adisky As per the comment https://github.com/openshift/odo/issues/2889#issuecomment-679132826 , this pr might fix issue https://github.com/openshift/odo/issues/2889 . Can you please confirm on this. If yes then please add the issue as fixes in description.

tests/examples/source/devfiles/springboot/devfile.yaml
```diff
@@ -5,47 +5,37 @@ metadata:
...
       endpoints:
         - name: "8080/tcp"
           targetPort: 8080
-      volumeMounts:
-        - name: springbootpvc
-          path: /data
-      mountSources: false
-  - volume:
-      name: springbootpvc
+      mountSources: true
```
maysunfaisal, line 24: lets keep this is as false, so that one container has the source while the other doesnt

tests/examples/source/devfiles/springboot/devfile.yaml
```diff
@@ -5,47 +5,37 @@ metadata:
...
 commands:
   - exec:
       id: defaultBuild
       component: tools
-      commandLine: "/artifacts/bin/build-container-full.sh"
+      commandLine: mvn install
```
maysunfaisal, line 29: are you planning on having these commands or keeping ti similar to your devfile PR?

maysunfaisal (commented):
Screen Shot 2020-09-10 at 5 19 53 PM

We have a few more files that reference the old image. Maybe we can even delete the v1 devfiles altogether!

tests/examples/source/devfiles/springboot/devfile.yaml
adisky: there are lots of issues if we do ` mountSources: false`, we need to copy the jar files from build to run container, tried doing that (check this devfile https://gist.github.com/adisky/d690abc464f65f864baacd47f29635f6) but faced lots of issue related to file permission.

tests/examples/source/devfiles/springboot/devfile.yaml
adisky: kept same a registry

adisky:
> We have a few more files that reference the old image. Maybe we can even delete the v1 devfiles altogether!

yes we can do that, let's do this with this PR https://github.com/openshift/odo/pull/3898

tests/examples/source/devfiles/springboot/devfile.yaml
maysunfaisal: I suspect its because you're on an OpenShift cluster? OpenShift starts container as a non-root user and your container's default user does not have permission to write to pvc dirs.

maysunfaisal: The devfile is good but the reason why the previous Spring devfile didnt utilize the project mount was because it is an emptyDir vol, and if for any reason the pod restarts; the contents are lost and that would include the artifacts/jar which runs the Spring server https://kubernetes.io/docs/concepts/storage/volumes/#emptydir This may mean, we would have to do a mvn build again to get the application up and running.

If we're however storing the artifacts/jar in a pvc; pvc contents are always present as long as the deployment is present. If for any reason, the pod restarts; the application would be back up and running.

However, since this a test devfile and we're really testing the two component container logic; it should be ok! Manually tried it out as well. 

/lgtm

adisky: completely agree that, but it is in general problem with all devfiles, there is separate issue to figure out to save the code/binaries in pvc instead of empty dir.

Issue #3750, opened 2020-08-12
Help text gets printed twice for invalid odo commands


johnmcollier:
/kind bug

## What versions of software are you using?

**Operating System:**
macOS

**Output of `odo version`:**
master

## How did you run odo exactly?
`odo <some-fake-command>`

## Actual behavior
```
Johns-MacBook-Pro-3:odo johncollier$ odo sdfsdf
odo is a CLI tool for running OpenShift applications in a fast and automated manner.
Reducing the complexity of deployment, odo adds iterative development without the worry of deploying your source code.

Find more information at https://github.com/openshift/odo

Usage:
  odo [flags]
  odo [command]

Examples:
  # Creating and deploying a Node.js project
  git clone https://github.com/openshift/nodejs-ex && cd nodejs-ex
  odo create nodejs
  odo push
  
  # Accessing your Node.js component
  odo url create

Commands:
  app         Perform application operations (delete, describe, list)
  catalog     Catalog related operations (describe, list, search)
  component   Manage components (create, delete, describe, exec, link, list, log, push, test, unlink, update, watch)
  config      Change or view configuration (set, unset, view)
  debug       Debug commands (info, port-forward)
  preference  Modifies preference settings (set, unset, view)
  project     Perform project operations (create, delete, get, list, set)
  registry    Configure devfile registry (add, delete, list, update)
  service     Perform service catalog operations (create, delete, list)
  storage     Perform storage operations (create, delete, list)
  test        Run the test command defined in the devfile
  url         Expose component to the outside world (create, delete, describe, list)

Utility Commands:
  help        Help about any command
  login       Login to cluster
  logout      Log out of the current OpenShift session
  utils       Utilities for terminal commands and modifying odo configurations (convert-to-devfile, terminal)
  version     Print the client version information

Component Shortcuts:
  create      Create a new component 
  delete      Delete component 
  describe    Describe component 
  exec        Executes a command inside the component 
  link        Link component to a service or component 
  list        List all components in the current application 
  log         Retrieve the log for the given component 
  push        Push source code to a component 
  unlink      Unlink component to a service or component 
  update      Update the source code path of a component 
  watch       Watch for changes, update component on change. Watch doesn't support git components. 

Flags:
      --complete             Install completion for odo command
  -h, --help                 Help for odo
      --uncomplete           Uninstall completion for odo command
  -v, --v Level              Number for the log level verbosity. Level varies from 0 to 9 (default 0).
[...]

PR #3805, opened 2020-08-21
remove showing help twice for incorrect commands

girishramnani:
**What type of PR is this?**

> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind failing-test
> /kind feature
> /kind flake
> /kind code-refactoring
>
> /kind documentation
> [skip ci]

see $SUBJECT

Fixes https://github.com/openshift/odo/issues/3750

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

```
odo hello 
```
should only print help once

### DIFF
--- a/pkg/odo/cli/cli.go
+++ b/pkg/odo/cli/cli.go
@@ -272,6 +272,5 @@ func ShowHelp(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	_ = cmd.Help()
 	return fmt.Errorf("Invalid command - see available commands/subcommands above")
 }
--- a/tests/integration/generic_test.go
+++ b/tests/integration/generic_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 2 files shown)


pkg/odo/cli/cli.go
```diff
@@ -272,6 +272,5 @@ func ShowHelp(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	_ = cmd.Help()
```
mik-dass, line 275: I think we can have a integration test where we check for only one occurence of `Invalid command - see available commands/subcommands above`

amitkrout:
@girishramnani Can you please attach the screenshot or text output of before and after change. I am asking because it's a UX change, so output before and after would really be helpful.

pkg/odo/cli/cli.go
amitkrout: Yes, may be it will be helpful to avoid future regression.

girishramnani: added the test - I couldn't use `Invalid command - see available commands/subcommands above` as that shows up twice already and hence went for a unique string

mik-dass (approved):
/approve

PR #3814, opened 2020-08-23
fixed incorrect logging order

dev-gaur:
**What type of PR is this?**
/kind code-refactoring

I found this bit of logging order a bit disinforming. 

"\nPushing devfile component %s" should be logged before the actually `Pushing` logic starts. not after that.

### DIFF
--- a/pkg/odo/cli/component/devfile.go
+++ b/pkg/odo/cli/component/devfile.go
@@ -118,15 +118,16 @@ func (po *PushOptions) devfilePushInner() (err error) {
 
 	warnIfURLSInvalid(po.EnvSpecificInfo.GetURL())
 
+	log.Infof("\nPushing devfile component %s", componentName)
+
 	// Start or update the component
 	err = devfileHandler.Push(pushParams)
 	if err != nil {
-		err = errors.Errorf("Failed to start component with name %s. Error: %v",
+		err = errors.Errorf("Failed to push component with name %s. Error: %v",
 			componentName,
 			err,
 		)
 	} else {
-		log.Infof("\nPushing devfile component %s", componentName)
 		log.Success("Changes successfully pushed to component")
 	}
 


prietyc123:
looks good to me.

/lgtm

dharmit:
/hold

@dev-gaur the part of code being moved here is specific to `else` block. What's being proposed as a part of this PR is to log it for both `if` and `else` blocks. I get it that it's just some logging so not a big deal. But, IMO, it's best to do `git blame` and discuss this quickly with the original contributor.

dev-gaur:
> /hold
> 
> @dev-gaur the part of code being moved here is specific to `else` block.

^ Which was wrong in the first place, right?

dharmit:
> > /hold
> > @dev-gaur the part of code being moved here is specific to `else` block.
> 
> ^ Which was wrong in the first place, right?

TBH, I'm not sure. :disappointed:

@jgwest can you PTAL since it looks like you added the code that @dev-gaur thinks needs a change.

adisky:
@dev-gaur I agree with your concern but with your PR  the line` ✓  Changes successfully pushed to component`  is looking misplaced. i believe the log `Pushing devfile component nodejs` is to denote success not process. 

``` 
$ odo push

Validation
 ✓  Validating the devfile [46035ns]

 ✓  Waiting for component to start [4ms]

Applying URL changes

Syncing to component nodejs
 ✓  Checking file changes for pushing [523446ns]
```

```
$ odo push
Validation
 ✓  Validating the devfile [50844ns]

Applying URL changes

Syncing to component nodejs
 ✓  Checking file changes for pushing [526513ns]
 ✓  Syncing files to the component [109ms]

 ✓  Executing install command "npm install", if not running [3s]
 ✓  Executing run command "npm start", if not running [1s]

```

jgwest:
> @jgwest can you PTAL since it looks like you added the code that @dev-gaur thinks needs a change.

Re: the specific ordering of these blocks, it wasn't I that added/changed them, I preserved the behaviour of that code as I found it (https://github.com/openshift/odo/commit/e28812a34d4a2443d1e816141c401e3562a5c6ae)

dharmit:
@dev-gaur https://github.com/openshift/odo/pull/3814#issuecomment-679975107 makes me think that the change you've proposed is likely not required. Unless there's some reasoning to take this further, we can close this PR. WDYT?

dev-gaur:
@adisky thanks for pointing that out.. I didn't realise this small a change would disrupt the logging order on that scale.

@dharmit sure.. if this is trivial then we can close this.

PR #3835, opened 2020-08-26
consolidated some integration tests

girishramnani:
**What type of PR is this?**

> /kind cleanup
see $SUBJECT

No issue

**PR acceptance criteria**:

- [ ] Unit test 

- [X] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -65,20 +65,11 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(output).To(ContainSubstring("Executing devfile commands for component " + name))
 		})
 
-		It("should be able to push based on name passed", func() {
-			name := "springboot"
-			helper.CopyExample(filepath.Join("source", "devfiles", "springboot", "project"), context)
-			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "springboot", "devfile.yaml"), filepath.Join(context, "devfile.yaml"))
-
-			output := helper.CmdShouldPass("odo", "push", "--namespace", namespace, name)
-			Expect(output).To(ContainSubstring("Executing devfile commands for component " + name))
-		})
-
 	})
 
 	Context("Verify devfile push works", func() {
 
-		It("should have no errors when no endpoints within the devfile, should create a service when devfile has endpoints", func() {
+		It("should have no errors when no endpoints within the devfile, should create a service when devfile has endpoints and multiple pushes with changes in devfile", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, cmpName)
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
@@ -93,22 +84,14 @@ var _ = Describe("odo devfile push command tests", func() {
 			output = helper.CmdShouldPass("odo", "push", "--project", namespace)
 
 			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
-			output = cliRunner.GetServices(namespace)
-			Expect(output).To(ContainSubstring(cmpName))
-		})
-
-		It("checks that odo push works with a devfile", func() {
-			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace, cmpName)
-
-			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), context)
-			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile.yaml"), filepath.Join(context, "devfile.yaml"))
-
-			output := helper.CmdShouldPass("odo", "push", "--project", namespace)
-			Expect(output).To(ContainSubstring("Changes successfully pushed to component"))
+			Expect(output).To(ContainSubstring("Executing devfile commands for component " + cmpName))
 
 			// update devfile and push again
 			helper.ReplaceString("devfile.yaml", "name: FOO", "name: BAR")
 			helper.CmdShouldPass("odo", "push", "--project", namespace)
+
+			output = cliRunner.GetServices(namespace)
+			Expect(output).To(ContainSubstring(cmpName))
 		})
 
 		It("checks that odo push works with a devfile with sourcemapping set", func() {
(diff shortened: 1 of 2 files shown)


tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -65,20 +65,11 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(output).To(ContainSubstring("Executing devfile commands for component " + name))
 		})
 
-		It("should be able to push based on name passed", func() {
```
prietyc123, line 68: Why are we removing this spec?

prietyc123 (changes requested):

tests/integration/devfile/cmd_devfile_test_test.go
```diff
@@ -44,15 +44,21 @@ var _ = Describe("odo devfile test command tests", func() {
 
 	Context("Should show proper errors", func() {
 
-		It("should show error if component is not pushed", func() {
+		It("should show error if component is not pushed or specify non-exist command or use a command from wrong group", func() {
```
prietyc123, line 47: +1

prietyc123 (commented):
Is it possible to fix all the devfile integration test with this? Or else if you take it in a separate pr also works for me.

tests/integration/devfile/cmd_devfile_push_test.go
girishramnani: because the check that this test is doing is done in the test bellow it https://github.com/openshift/odo/pull/3835/files#diff-c127bd883f89dd3a7abff6183ad9921dR87

girishramnani:
I am gradually going to improve the test and hence would open separate PR

tests/integration/devfile/cmd_devfile_push_test.go
prietyc123: This is completely different scenario as compared to the one you mentioned. The spec you removed has a `name` field passing along with `odo push` command, which is nowhere verified throughout the spec AFAIK.

Can you please revert it back or adjust with your modified steps (may be it will be difficult).

girishramnani: It’s not completely different, I will add the name in odo push

prietyc123: Either way works for me.

tests/integration/devfile/cmd_devfile_test_test.go
```diff
@@ -44,15 +44,22 @@ var _ = Describe("odo devfile test command tests", func() {
 
 	Context("Should show proper errors", func() {
 
-		It("should show error if component is not pushed", func() {
+		// used ";" as consolidating symbol as this spec covers multiple scenerios
+		It("should show error if component is not pushed; should error out if a non-existent command or a command from wrong group is specified", func() {
```
prietyc123, line 48: `It` clause can be acts as a `then` clause in ginkgo. Of Course you can add multiple statement by using any separator like ```;```, ```/``` whatever but it will be a misuse of delimiter in a ```It``` description I believe. 
Moreover `It` clause should be a simple string which should explain what is the spec stand for and TBH adding a ```;``` looks like we are implementing our programming skills in writing a simple string 🙂 . Instead we can add the multiple statement in a single like.
```
Should show error if component is not pushed or specified commands either does not exist or belongs to wrong group
```
Can’t we ?

prietyc123 (changes requested):

tests/integration/devfile/cmd_devfile_test_test.go
girishramnani: consolidation is signified with `;` and I discussed this with @dharmit . And I agree that `;` makes it more readable at this point.

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -90,25 +82,17 @@ var _ = Describe("odo devfile push command tests", func() {
 			Expect(output).NotTo(ContainSubstring(cmpName))
 
 			helper.RenameFile("devfile-old.yaml", "devfile.yaml")
-			output = helper.CmdShouldPass("odo", "push", "--project", namespace)
+			output = helper.CmdShouldPass("odo", "push", "--project", namespace, cmpName)
```
prietyc123, line 85: What if we don't pass `cmpName` with `odo push`. Does it mean that won't work? No right? So I would say lets keep it as simple as it was before and revert https://github.com/openshift/odo/pull/3835/files#diff-c127bd883f89dd3a7abff6183ad9921dL68 back. TBH its complicating the current spec and one more reason is, I can see only one spec in this file which verifies `sprinboot` devfile.yaml works well here - https://github.com/openshift/odo/pull/3835/files#diff-c127bd883f89dd3a7abff6183ad9921dL68 

Removing this spec doesn't make sense to me which verifies two thing:
- push works fine with `springboot devfile.yaml`
- We can also do `odo push` along with `name` field.

prietyc123 (changes requested):

tests/integration/devfile/cmd_devfile_push_test.go
girishramnani: but wouldn't this spec also verify its working of `springboot` and I have kept the non-component name push in this spec as well
https://github.com/openshift/odo/pull/3835/files#diff-c127bd883f89dd3a7abff6183ad9921dR92
So I am trying to understand why do you feel its complicating, its just consolidating a very small verification of push with name with other specs which seems like a good refactor to me.

tests/integration/devfile/cmd_devfile_test_test.go
prietyc123: Make sense. However I have gone through the whole ginkgo doc and nowhere it has been mentioned that we can use delimiter like `;` for separating multiple scenario in a single spec description. Also it is nowhere written we can not use this, afterall its a string message. As we are introducing `;` for the first time, I feel its better to document it in test guidelines https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc of our reference.

prietyc123:
Thanks @girishramnani for addressing the requested change. Changes looks good to me.

/lgtm

tests/integration/devfile/cmd_devfile_test_test.go
prietyc123: @girishramnani As per https://github.com/openshift/odo/pull/3835#discussion_r478213783 could you please add `;` description in the test guideline.

prietyc123: If you want i can add this change in a separate pr in the test guideline.

Issue #3801, opened 2020-08-20
Clean up context in devfile integration tests


maysunfaisal:
Currently most of the devfile integration tests that use context flag `--context`, change to the context directory in `BeforeEach` even though we're using the full path in `--context`

Ideally, we should not switch to the context directory when using `--context`. Clean up all devfile integration tests for the flag.

/area devfile
/kind cleanup

dharmit:
Doesn't look like we addressed this. Reopining the issue.

valaparthvi:
I would like to work on this.
/assign me

valaparthvi:
Currently the following devfile integration tests need this cleanup -
```
tests/integration/devfile/cmd_devfile_debug_test.go
tests/integration/devfile/cmd_devfile_delete_test.go
tests/integration/devfile/cmd_devfile_describe_test.go
tests/integration/devfile/cmd_devfile_log_test.go
tests/integration/devfile/cmd_devfile_push_test.go
tests/integration/devfile/cmd_devfile_test_test.go
tests/integration/devfile/cmd_devfile_url_test.go
tests/integration/devfile/cmd_devfile_watch_test.go
```

valaparthvi:
Scope of this issue for sprint 199 - 
[ ] 1. Fix the tests mentioned in https://github.com/openshift/odo/issues/3801#issuecomment-779113525. Some of the tests are a quick fix, some might require a little more work.
[ ] 2. Identify any other tests that might need cleanup.

_Might be a 3-day task._

valaparthvi:
How do we want to go about this? Do we want to remove the `--context` flag wherever it is not used to test the context or remove the `Chdir` in `BeforeEach` and modify all tests to work with the changes? I am not sure if doing either affects the test coverage.

@dharmit @girishramnani @mik-dass thoughts?

mik-dass:
> How do we want to go about this? Do we want to remove the --context flag wherever it is not used to test the context or remove the Chdir in BeforeEach and modify all tests to work with the changes? I am not sure if doing either affects the test coverage.

We should remove the Chdir in BeforeEach wherever possible as that would help us figure out if the context flag works for the command properly or not.

PR #3843, opened 2020-08-26
implement odo describe for devfile

### LINKED ISSUE #3644: implement odo describe for devfile
`odo describe` should work with the devfile component including the JSON output.

The output format should be same for both types of components. 

/priority high
/area devfile

metacosm:
**What type of PR is this?**
/kind bug

Make `describe` work with devfiles.

Fixes #3644 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


girishramnani:
@metacosm can we please focus on implementing the `odo describe` feature and then open a follow up for cleanup as it is a time critical task

metacosm:
> @metacosm can we please focus on implementing the `odo describe` feature and then open a follow up for cleanup as it is a time critical task

This is not a clean up. These are steps leading to the implementation as the code is a mess and I'm not about to duplicate the whole code that deals with s2i components just to get it to work on devfiles as well.

pkg/component/component_full_description.go
```diff
@@ -131,15 +128,23 @@ func NewComponentFullDescriptionFromClientAndLocalConfig(client *occlient.Client
 
 	cfd.fillEmptyFields(componentDesc, componentName, applicationName, projectName)
 
-	err = cfd.loadURLSFromClientAndLocalConfig(client, localConfigInfo, componentName, applicationName)
+	var urls urlpkg.URLList
+	if envInfo != nil {
+		routeSupported, _ := client.IsRouteSupported()
```
mik-dass, line 133: please handle the error here

mik-dass:
Please add a integration test using `odo describe` for devfiles.

pkg/envinfo/envinfo.go
```diff
@@ -122,20 +132,20 @@ type EnvInfoLink struct {
 
 // getEnvInfoFile first checks for the ENVINFO variable
 // then we check for directory and eventually the file (which we return as a string)
-func getEnvInfoFile(envDir string) (string, error) {
+func getEnvInfoFile(envDir string) (string, string, error) {
 	if env, ok := os.LookupEnv(envInfoEnvName); ok {
-		return env, nil
+		return env, filepath.Join(env, "..", "..", "..", "devfile.yaml"), nil
```
mik-dass, line 137: why not directly set the devfilePath in newEnvSpecificInfo() using envDir there?

metacosm: Because then I would need to duplicate the logic in that function to compute the starting point.

mik-dass: I didn't get what you mean by starting point. In most cases, `envDir` is the context directory. We can just join the default devfilePath i.e `./devfile.yaml` to it. Or am I missing something?

metacosm: I may be missing something but the starting point from which the devfile path is computed depends on several things. Look at the code for this function and you'll see what I mean.

metacosm: That said, I've seen different ways of dealing with this in different parts of the code so… 🤷

pkg/envinfo/envinfo.go
```diff
@@ -280,8 +291,8 @@ func (esi *EnvSpecificInfo) IsSet(parameter string) bool {
 	return util.IsSet(esi.componentSettings, parameter)
 }
 
-// EnvInfoFileExists if the envinfo file exists or not
-func (esi *EnvSpecificInfo) EnvInfoFileExists() bool {
+// Exists returns whether the envinfo file exists or not
+func (esi *EnvSpecificInfo) Exists() bool {
```
prietyc123, line 295: Shouldn't `EnvInfoFileExists` makes more sense than `Exists`

prietyc123 (changes requested):

pkg/envinfo/envinfo.go
metacosm: No because it's part of the `LocalConfigProvider` interface which is also implemented by `LocalConfigInfo`.

kadel: This is correct. I don't like the long names that are duplicating package or struct names that we have in some places.

metacosm:
1 test failing and not always the same one… seems like a flake to me.

dgolovin:
@metacosm 'odo describe' reports component as not pushed, when it actually is

```
denis@denis-ThinkPad-P50:~/p/test-ws/nodejs-ex$ ~/go/src/github.com/openshift/odo/dist/bin/linux-amd64/odo list

Devfile Components: 
APP      NAME          PROJECT        TYPE       STATE
app1     nodejs-ex     project156     nodejs     Pushed

Openshift Components: 
APP      NAME           PROJECT        TYPE       SOURCETYPE     STATE
app1     component1     project156     nodejs     local          Pushed
denis@denis-ThinkPad-P50:~/p/test-ws/nodejs-ex$ ~/go/src/github.com/openshift/odo/dist/bin/linux-amd64/odo describe -o json
{
	"kind": "Component",
	"metadata": {
		"name": "nodejs-ex",
		"namespace": "project156",
		"creationTimestamp": null
	},
	"spec": {
		"app": "app1",
		"type": "devfile",
		"urls": {
			"kind": "List",
			"metadata": {},
			"items": null
		},
		"storages": {
			"kind": "List",
			"metadata": {},
			"items": null
		},
		"ports": [
			"5858"
		]
	},
	"status": {
		"state": "Not Pushed"
	}
}
```

pkg/envinfo/envinfo.go
girishramnani, line 137: lets not do this

metacosm: Indeed, let's see if I can find a better way.

mik-dass: I checked the function again. You are right, my suggestion won't work in case of the env specified envDir.

pkg/component/component_full_description.go
```diff
@@ -44,36 +47,44 @@ func (cfd *ComponentFullDescription) copyFromComponentDesc(component *Component)
...
+func (cfd *ComponentFullDescription) loadStoragesFromClientAndLocalConfig(client *occlient.Client, kClient *kclient.Client, envinfo *envinfo.EnvSpecificInfo, localConfigInfo *config.LocalConfigInfo, componentName string, applicationName string, componentDesc *Component) error {
 	var storages storage.StorageList
 	var err error
+
+	isDevfile := envinfo != nil
+	var devfile devfileParser.DevfileObj
+	if isDevfile {
+		devfile, err = devfileParser.Parse(envinfo.GetDevfilePath())
+		if err != nil {
+			panic(err)
```
johnmcollier, line 60: Return an error here rather than panicking

johnmcollier:
So I can run `odo describe` on an unpushed devfile component, but it fails for me after pushing:
Before:
```
Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
Component Name: nodejs
Type: devfile: nodejs
```

After (on OpenShift):
```
Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
 ✗  deploymentconfigs.apps.openshift.io "nodejs-app" not found
```

After (on Kube):
```
Johns-MacBook-Pro-3:do-node johncollier$ odo describe
 ✗  the server could not find the requested resource (get deploymentconfigs.apps.openshift.io)
```

metacosm:
> So I can run `odo describe` on an unpushed devfile component, but it fails for me after pushing:
> Before:
> 
> ```
> Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> Component Name: nodejs
> Type: devfile: nodejs
> ```
> 
> After (on OpenShift):
> 
> ```
> Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
>  ✗  deploymentconfigs.apps.openshift.io "nodejs-app" not found
> ```
> 
> After (on Kube):
> 
> ```
> Johns-MacBook-Pro-3:do-node johncollier$ odo describe
>  ✗  the server could not find the requested resource (get deploymentconfigs.apps.openshift.io)
> ```

Yep, just realized this myself. The architecture is horrendous and makes it difficult to support all possible cases…

pkg/component/component.go
```diff
@@ -913,40 +914,73 @@ func List(client *occlient.Client, applicationName string, localConfigInfo *conf
...
+		return component, nil
+	}
+	return Component{}, nil
+}
+
+func GetComponentFromDevfile(info *envinfo.EnvSpecificInfo) Component {
+	if info.Exists() {
+		devfile, err := devfileParser.Parse(info.GetDevfilePath())
+		if err != nil {
+			panic(err)
```
johnmcollier, line 946: Rather than panicking here, can you update `GetComponentFromDevfile` to also return an error value and return an error here?

pkg/occlient/occlient.go
```diff
@@ -3030,6 +3030,10 @@ func (c *Client) BuildPortForwardReq(podName string) *rest.Request {
 		SubResource("portforward")
 }
 
+func (c *Client) GetKubeClient() kubernetes.Interface {
```
johnmcollier, line 3033: 👍

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -0,0 +1,71 @@
+package devfile
```
johnmcollier, line 1: Can you add a rule to the makefile so that the tests in here can be run individually?

Something like
```
.PHONY: test-cmd-devfile-describe
test-cmd-devfile-describe:
	ginkgo $(GINKGO_FLAGS) -focus="odo devfile describe command tests" tests/integration/devfile/
```

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -0,0 +1,71 @@
...
+		cliRunner.DeleteNamespaceProject(namespace)
+		helper.Chdir(currentWorkingDirectory)
+		err := os.Setenv("KUBECONFIG", originalKubeconfig)
+		Expect(err).NotTo(HaveOccurred())
+		helper.DeleteDir(context)
+		os.Unsetenv("GLOBALODOCONFIG")
+	})
+
+	Context("When executing odo describe", func() {
+		It("should describe the component when it is not pushed", func() {
```
johnmcollier, line 44: Can you add a test case for when it is pushed?

johnmcollier (changes requested):
Left some review comments. Some good refactoring that I like too 👍 

Also, thoughts on changing the output slightly from:
```
Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
Component Name: mynode
Type: devfile: nodejs
```

To either:
```
Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
Component Name: mynode
Deployment mode: devfile
Component type: nodejs
```
or 
```
Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
Component Name: mynode
Type: nodejs (devfile)
```

metacosm:
> Left some review comments. Some good refactoring that I like too 👍
> 
> Also, thoughts on changing the output slightly from:
> 
> ```
> Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> Component Name: mynode
> Type: devfile: nodejs
> ```
> 
> To either:
> 
> ```
> Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> Component Name: mynode
> Deployment mode: devfile
> Component type: nodejs
> ```
> 
> or
> 
> ```
> Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> Component Name: mynode
> Type: nodejs (devfile)
> ```

I don't know about this point. I asked on Slack about it but didn't get an answer… /cc @kadel @girishramnani

kadel:
> > Left some review comments. Some good refactoring that I like too 👍
> > Also, thoughts on changing the output slightly from:
> > ```
> > Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> > Component Name: mynode
> > Type: devfile: nodejs
> > ```
> > 
> > 
> > To either:
> > ```
> > Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> > Component Name: mynode
> > Deployment mode: devfile
> > Component type: nodejs
> > ```
> > 
> > 
> > or
> > ```
> > Johns-MacBook-Pro-3:node-devfile johncollier$ odo describe
> > Component Name: mynode
> > Type: nodejs (devfile)
> > ```
> 
> I don't know about this point. I asked on Slack about it but didn't get an answer… /cc @kadel @girishramnani

Does the user need to know if it is a devfile component or not? I would consider completely skipping that information from the normal output. Otherwise I would go with  `Deployment mode: devfile` option

metacosm:
> Does the user need to know if it is a devfile component or not? I would consider completely skipping that information from the normal output. Otherwise I would go with `Deployment mode: devfile` option

So just: `Type: nodejs` without mention of `devfile`?

kadel:
> @kadel So just: `Type: nodejs` without mention of `devfile`?

yes. I would go with just that.

tests/integration/devfile/cmd_devfile_describe_test.go
metacosm: Done.

tests/integration/devfile/cmd_devfile_describe_test.go
metacosm: Done.

pkg/component/component.go
```diff
@@ -913,40 +914,73 @@ func List(client *occlient.Client, applicationName string, localConfigInfo *conf
...
+				}
+			}
+		}
+
+		return component
+	}
+	return Component{}
+}
+
+func getComponentFrom(info envinfo.LocalConfigProvider, componentType string) Component {
```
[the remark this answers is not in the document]
metacosm: The something are the parameters :smile:

pkg/component/component.go
```diff
@@ -822,30 +817,16 @@ func Deploy(client *occlient.Client, params occlient.CreateArgs, desiredRevision
 	return nil
 }
 
-// GetComponentType returns type of component in given application and project
-func GetComponentType(client *occlient.Client, componentName string, applicationName string) (string, error) {
-
-	// filter according to component and application name
-	selector := fmt.Sprintf("%s=%s,%s=%s", componentlabels.ComponentLabel, componentName, applabels.ApplicationLabel, applicationName)
-	componentImageTypes, err := client.GetDeploymentConfigLabelValues(componentlabels.ComponentTypeLabel, selector)
+func GetComponentNames(client *occlient.Client, applicationName string) ([]string, error) {
```
mik-dass, line 820: Some comments regarding the function

pkg/component/pushed_component.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return "", fmt.Errorf("%s component doesn't provide a type label", component.GetName())
+}
+
+func GetPushedComponents(c *occlient.Client, applicationName string) (map[string]PushedComponent, error) {
+	applicationSelector := fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
+	dcList, err := c.GetDeploymentConfigsFromSelector(applicationSelector)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			dList, err := c.GetKubeClient().ListDeployments(applicationSelector)
```
mik-dass, line 216: can we run these two network operations in parallel?

pkg/component/pushed_component.go
```diff
@@ -0,0 +1,276 @@
...
+func GetPushedComponent(c *occlient.Client, componentName, applicationName string) (PushedComponent, error) {
+	d, err := c.GetKubeClient().GetDeploymentByName(componentName)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			// if it's not found, check if there's a deploymentconfig
+			deploymentName, err := util.NamespaceOpenShiftObject(componentName, applicationName)
+			if err != nil {
+				return nil, err
+			}
+			dc, err := c.GetDeploymentConfigFromName(deploymentName)
```
mik-dass, line 264: can we run these client operations in parallel?

pkg/component/pushed_component.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func getType(component provider) (string, error) {
+	if componentType, ok := component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
+		return componentType, nil
+	}
+	return "", fmt.Errorf("%s component doesn't provide a type label", component.GetName())
+}
+
+func GetPushedComponents(c *occlient.Client, applicationName string) (map[string]PushedComponent, error) {
```
mik-dass, line 211: Any reason why we are returning a list instead of a map?

mik-dass: Please add some comment for the function

pkg/component/pushed_component.go
metacosm: We could but I'd rather not complicate things at this point.

pkg/component/pushed_component.go
metacosm: Why not return a map?

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		SetDefaultEventuallyTimeout(10 * time.Minute)
+		context = helper.CreateNewContext()
+		os.Setenv("GLOBALODOCONFIG", filepath.Join(context, "config.yaml"))
+		helper.CmdShouldPass("odo", "preference", "set", "Experimental", "true")
+		originalKubeconfig = os.Getenv("KUBECONFIG")
+		helper.LocalKubeconfigSet(context)
+		namespace = cliRunner.CreateRandNamespaceProject()
+		currentWorkingDirectory = helper.Getwd()
+		helper.Chdir(context)
```
mik-dass, line 33: Let's not change the directory

pkg/component/pushed_component.go
```diff
@@ -0,0 +1,276 @@
...
+		provider:    p,
+	}
+	if err := comp.retrieveURLsAndRoutes(c); err != nil {
+		return nil, err
+	}
+	return comp, nil
+}
+
+// GetPushedComponent returns an abstraction over the cluster representation of the component
+func GetPushedComponent(c *occlient.Client, componentName, applicationName string) (PushedComponent, error) {
```
mik-dass, line 255: Please add unit tests for this and some of the above functions

pkg/component/pushed_component.go
girishramnani: ordering becomes predictable

tests/integration/devfile/cmd_devfile_describe_test.go
metacosm: I merely ported the existing test to devfile so sure, I could do that but then why was it allowed for the existing test?

pkg/component/pushed_component.go
metacosm: These should already be covered by the existing tests, I think as this function is only meant to hide where the component data comes from. The external behavior should already be tested.

pkg/component/pushed_component.go
mik-dass: Because this function is only called by `GetComponentNames()` which loops over the map keys to get the names. But we can loop over the list of components and use `<pushed-component>.GetName()` to get the name of each component too. Unless we use the keys to find individual items, a map usage feels weird to me as we are looping over all the items.

tests/integration/devfile/cmd_devfile_describe_test.go
mik-dass: We currently have a issue to remove those https://github.com/openshift/odo/issues/3801.

pkg/component/pushed_component.go
mik-dass: But for s2i, it will always be one extra unnecessary call before the correct one.

mik-dass:
```
[mrinaldas@localhost project]$ odo url create example --port 3000 --host com
 ✓  URL example created for component: nodejs

[mrinaldas@localhost project]$ odo describe Component Name: nodejs
Type: nodejs
URLs:
 · URL named example will be exposed via 5858
[mrinaldas@localhost project]$ odo storage create --path /data
 ✓  Added storage nodejs-uzoy to nodejs

[mrinaldas@localhost project]$ odo describe Component Name: nodejs
Type: nodejs
Storage:
 · nodejs-uzoy of size 1Gi mounted to /data
URLs:
 · URL named example will be exposed via 5858
```
it seems to show the wrong port for non pushed components. After push, the port is displayed correctly.

metacosm:
/retest


> ```
> [mrinaldas@localhost project]$ odo url create example --port 3000 --host com
>  ✓  URL example created for component: nodejs
> 
> To apply the URL configuration changes, please use `odo push`
> [mrinaldas@localhost project]$ odo describe Component Name: nodejs
> Type: nodejs
> URLs:
>  · URL named example will be exposed via 5858
> [mrinaldas@localhost project]$ odo storage create --path /data
>  ✓  Added storage nodejs-uzoy to nodejs
> 
> Please use `odo push` command to make the storage accessible to the component
> [mrinaldas@localhost project]$ odo describe Component Name: nodejs
> Type: nodejs
> Storage:
>  · nodejs-uzoy of size 1Gi mounted to /data
> URLs:
>  · URL named example will be exposed via 5858
> ```
> 
> it seems to show the wrong port for non pushed components. After push, the port is displayed correctly.
This behavior is due to https://github.com/metacosm/odo/blob/05e992c1512a42cfa085805e3a08c70961003779/pkg/component/component_full_description.go#L222-L236, which I didn't touch. I'm not sure why this is done this way, actually…

mik-dass:
> @mik-dass This behavior is due to https://github.com/metacosm/odo/blob/05e992c1512a42cfa085805e3a08c70961003779/pkg/component/component_full_description.go#L222-L236, which I didn't touch. I'm not sure why this is done this way, actually… and I'm therefore not sure what the fix should be. I'm tired of changing something to fix an issue just to break something else.

I had a look again and it seems we are using the debugPort for the ports in case of the devfile https://github.com/openshift/odo/pull/3843/files#diff-692effa1b0b7a9ed8c1f7bb49a977b9aR952 and there is no `GetPorts()` in case of devfile. We need to use `devfile.ToRepresentation()` to get the ports from the devfile.

pkg/component/pushed_component.go
metacosm: Sure, but there is no intrinsic order here which returning a list would imply. `GetComponentNames` should indeed return a sorted list, which I will do.

johnmcollier (approved):
Changes look fine to me, don't see any glaring issues. Tested manually too on OCP and Kube, works great. 

Gonna approve this now if there's no concerns, let's get this one in 👍 

/lgtm

kadel:
There is something horribly wrong with our setup scripts.
This is why tests are failing :-( 
```
	bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found

```
I'm looking into it

kadel:
> There is something horribly wrong with our setup scripts.
> This is why tests are failing :-(
> 
> ```
> 	bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
> ```
> 
> I'm looking into it

fix in https://github.com/openshift/odo/pull/3984

dgolovin:
@kadel should kind value be different for s2i and devfile components? currently the only way to recognize devfile component is to check for devfile.yaml in local context.

metacosm:
Could I get an LGTM back, please? 😄

girishramnani:
@dgolovin we can do that in a follow up

Issue #3842, opened 2020-08-26
odo registry list causes panic if preference has not been setup


mohammedzee1000:
/kind bug

## What versions of software are you using?
Master

**Operating System:**
Fedora 31

**Output of `odo version`:**
```
❯ make bin && sudo cp -avrf ./odo /usr/bin/odo
go build -ldflags="-w -X github.com/openshift/odo/pkg/version.GITCOMMIT=68bc98ba8" cmd/odo/odo.go
'./odo' -> '/usr/bin/odo'
❯ odo version
odo v1.2.5 (68bc98ba8)

```

## How did you run odo exactly?

```
❯ rm -rf ~/.odo && odo registry list
```

## Actual behavior

```
❯ rm -rf ~/.odo && odo registry list
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x188db3d]

goroutine 1 [running]:
github.com/openshift/odo/pkg/odo/cli/registry.(*ListOptions).Run(0x316a9c8, 0x0, 0x0)
	/home/mzee1000/workbench/work/golang/redhat/src/github.com/openshift/odo/pkg/odo/cli/registry/list.go:60 +0x5d
github.com/openshift/odo/pkg/odo/genericclioptions.GenericRun(0x2062c80, 0x316a9c8, 0xc000524280, 0x316a9c8, 0x0, 0x0)
	/home/mzee1000/workbench/work/golang/redhat/src/github.com/openshift/odo/pkg/odo/genericclioptions/runnable.go:31 +0x13c
github.com/openshift/odo/pkg/odo/cli/registry.NewCmdList.func1(0xc000524280, 0x316a9c8, 0x0, 0x0)
	/home/mzee1000/workbench/work/golang/redhat/src/github.com/openshift/odo/pkg/odo/cli/registry/list.go:101 +0x5e
github.com/openshift/odo/vendor/github.com/spf13/cobra.(*Command).execute(0xc000524280, 0x316a9c8, 0x0, 0x0, 0xc000524280, 0x316a9c8)
	/home/mzee1000/workbench/work/golang/redhat/src/github.com/openshift/odo/vendor/github.com/spf13/cobra/command.go:830 +0x2aa
github.com/openshift/odo/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc000189180, 0x1e47658, 0xc000043920, 0x2)
[...]

Issue #2705, opened 2020-03-11
Determine how odo should handle multiple projects in a devfile


rajivnathan:
## User Story

As a user, I want to be able to use a devfile that has multiple projects defined because devfiles allow this. Need to determine how odo should handle multiple projects in a devfile.

This follows from the discussion here: https://github.com/openshift/odo/pull/2681#issuecomment-597783079

## Acceptance Criteria
- [ ] When a single project is defined the source should be synced to `/projects/<project name>`
- [ ] *TBD for multiple projects*

/kind user-story
/area devfile

johnmcollier:
@rajivnathan Maybe we want this to be a user story?

elsony:
This item is done as part of https://github.com/openshift/odo/issues/2582
/close

PR #3845, opened 2020-08-26
Add devfile registry on preferences init when ODO_EXPERIMENTAL=true

johnmcollier:
**What type of PR is this?**
/kind bug

This PR is mostly a backport of https://github.com/johnmcollier/odo/commit/274acb72db1fb3ac4ec06104dae639856d6602f2, which was delivered into the odov2 alpha release

This PR fixes a panic that can be seen when `odo registry list` is called but the preferences file does not yet exist (and experimental mode is set via `ODO_EXPERIMENTAL=true`). What was happening was that when retrieve a preferences file, we only add the default devfile registry if the preferences file already exists, if we have to create and initialize the preferences file, we don't add the registry.  So a simple solution is to make sure we add the devfile registries when we create and initialize the preferences file.

I also added an integration test case to verify this scenario. As this bug was specific to when `ODO_EXPERIMENTAL` was set, I had to add some wrapper functions that allow us to pass in env variables to commands in the test framework.

Fixes https://github.com/openshift/odo/issues/3842

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

1. Delete `~/.odo/preference.yaml`
2. Run `export ODO_EXPERIMENTAL=true`
3. Run `odo registry list`, verify the default registry is listed and noo panic occurs.

### DIFF
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -218,6 +218,15 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
 
 	// If the preference file doesn't exist then we return with default preference
 	if _, err = os.Stat(preferenceFile); os.IsNotExist(err) {
+		// If initializing a new preferences file, make sure we add the default devfile registry to it
+		defaultRegistryList := []Registry{
+			{
+				Name:   DefaultDevfileRegistryName,
+				URL:    DefaultDevfileRegistryURL,
+				Secure: false,
+			},
+		}
+		c.OdoSettings.RegistryList = &defaultRegistryList
 		return &c, nil
 	}
 
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -218,15 +218,20 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
 
 	// If the preference file doesn't exist then we return with default preference
 	if _, err = os.Stat(preferenceFile); os.IsNotExist(err) {
-		// If initializing a new preferences file, make sure we add the default devfile registry to it
-		defaultRegistryList := []Registry{
-			{
-				Name:   DefaultDevfileRegistryName,
-				URL:    DefaultDevfileRegistryURL,
-				Secure: false,
-			},
+		// Add the default devfile registry if "ODO_EXPERIMENTAL" env variable is true
+		// We specifically need to check for the experimental mode env, as `odo preferences set experimental true` already properly
+		// initializes the default devfile registries.
+		experimentalEnvStr, _ := os.LookupEnv("ODO_EXPERIMENTAL")
+		if experimentalEnvStr == "true" {
+			defaultRegistryList := []Registry{
+				{
+					Name:   DefaultDevfileRegistryName,
+					URL:    DefaultDevfileRegistryURL,
+					Secure: false,
+				},
+			}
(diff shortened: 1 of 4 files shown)


johnmcollier:
Infra flakes
```
 2020/08/26 19:49:21 error: unable to signal to artifacts container to terminate in pod integration-e2e-ipi-conf, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/08/26 19:49:21 error: artifacts container for pod integration-e2e-ipi-conf unready: timed out waiting for the condition
2020/08/26 19:49:27 Executing "integration-e2e-gather-aws-console"
2020/08/26 20:19:57 error: unable to signal to artifacts container to terminate in pod integration-e2e-gather-aws-console, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/08/26 20:19:57 error: artifacts container for pod integration-e2e-gather-aws-console unready: timed out waiting for the condition
2020/08/26 20:20:02 Executing "integration-e2e-gather-must-gather"
2020/08/26 20:20:02 The namespace in which this test is executing has been deleted, cancelling the test... 
```

/retest

adisky (changes requested):
@johnmcollier Thanks for the fix, please fix here too to avoid panic in future, dereferencing a pointer without checking for nil
https://github.com/openshift/odo/blob/master/pkg/odo/cli/registry/list.go#L60

johnmcollier:
@adisky Thanks! Good catch. I've updated this PR to include the nil check 👍

pkg/preference/preference.go
```diff
@@ -218,6 +218,20 @@ func NewPreferenceInfo() (*PreferenceInfo, error) {
 
 	// If the preference file doesn't exist then we return with default preference
 	if _, err = os.Stat(preferenceFile); os.IsNotExist(err) {
+		// Add the default devfile registry if "ODO_EXPERIMENTAL" env variable is true
+		// We specifically need to check for the experimental mode env, as `odo preferences set experimental true` already properly
+		// initializes the default devfile registries.
+		experimentalEnvStr, _ := os.LookupEnv("ODO_EXPERIMENTAL")
```
cdrage, line 224: Do we need this check? Should we even bother checking if we're in experimental mode? I have applied your patch to my PR here: https://github.com/openshift/odo/pull/3705 which doesn't include looking up experimental and I have all tests passing.

Just out of curiosity.

johnmcollier: @cdrage This test https://github.com/openshift/odo/blob/master/tests/integration/component.go#L293 was failing without this check, as devfile components were getting listed without experimental mode being on.

johnmcollier: Looks like your branch modifies the test to remove that check (which makes sense, as devfile is being moved out of experimental mode), but that explains why the test isn't failing on your branch

johnmcollier:
Closing this PR in favour of https://github.com/openshift/odo/pull/3705 which has the same changes

adisky:
@johnmcollier charlie's PR does not have this change https://github.com/openshift/odo/pull/3845/files#diff-0cd3824178637db7a863b63f797b00d4, you could reopen and update this PR OR include this changes in #3705

johnmcollier:
@adisky I'll add a note to #2705 to get the nil check in

PR #3854, opened 2020-08-27
fix: remove requirement for URL to end with .zip

### LINKED ISSUE #3853: zip starter projects should not require URLs ending with .zip
/kind bug

**Operating System:** macOS 10.15.6

**Output of `odo version`:** odo v1.2.5 (0780a6bbf)

`odo create --devfile ../devfile.yaml --starter=foo` with a devfile I created explicitly to have several starter zip projects

## Actual behavior
 ✗  Invalid zip url: <url that doesn't end with .zip>

## Expected behavior
If the URL points to a non-zip file, it should only fail when trying to unzip it, not prevent people from using URLs that don't end in `.zip`

See above.

metacosm:
**What type of PR is this?**
/kind bug

Fixes #3853

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -904,9 +904,6 @@ func GetAndExtractZip(zipURL string, destination string, pathToUnzip string) err
 	if zipURL == "" {
 		return errors.Errorf("Empty zip url: %s", zipURL)
 	}
-	if !strings.Contains(zipURL, ".zip") {
-		return errors.Errorf("Invalid zip url: %s", zipURL)
-	}
 
 	var pathToZip string
 	if strings.HasPrefix(zipURL, "file://") {


pkg/util/util.go
```diff
@@ -904,9 +904,6 @@ func GetAndExtractZip(zipURL string, destination string, pathToUnzip string) err
 	if zipURL == "" {
 		return errors.Errorf("Empty zip url: %s", zipURL)
 	}
-	if !strings.Contains(zipURL, ".zip") {
-		return errors.Errorf("Invalid zip url: %s", zipURL)
-	}
 
 	var pathToZip string
 	if strings.HasPrefix(zipURL, "file://") {
```
girishramnani, line 909: I am ok with this PR but do we want to also enforce that the url starts with `file://` a url could be a simple http or https one but could have a zip file returned?

metacosm: You're missing the rest of the code which deals with that… 😉

amitkrout:
platform issue
/test v4.5-integration-e2e

kadel:
small fix
/lgtm
/approve

PR #3857, opened 2020-08-27
move URL info to devfile

### LINKED ISSUE #3615: url list and describe need update to display url defined in devfile endpoint entry

### LINKED ISSUE #3718: Move URLs from env to devfile
/kind feature

- [ ] `odo url create` should create endpoint entries into the devfile instead of creating URLs in the env file. 
- [ ] https://github.com/openshift/odo/pull/3665 will remove the secure flag temporarily. It should be restored and the attribute should be stored against the endpoint parameter `secure`
- [ ] should provide option to select the path of the url
- [ ] `odo url list` and `odo url delete` should list endpoints (with proper states) and delete from the devfile respectively.

## Why is this needed?

This will help keep URLs in one place i.e the devfile and reduce confusion.

yangcao77:
**What type of PR is this?**

/area devfile
/area url

This PR moves most of URL info from env.yaml to devfile, for Route and Ingress URL type for devfile components

Upon url creation, it will save the URL info into devfile endpoints. Only `urlName`, `urlKind`, `host` and `tlsSecret`
(for ingress type only) will be saved in env.yaml. devfile endpoint will contain information of `Name`(should be as urlName), `TargetPort`, `Exposure`, `Secure`, `Protocol`, `Path`. 
created `--path`, `--exposure`, `--protocol` flags for `odo url create` to collect those specific settings. 
The settings will be validated. If not provided, default values will be assigned according to definition in devfile schema: https://devfile.github.io/devfile/_attachments/api-reference.html

 
url list and describe checks and treats devfile endpoints as local URLs
url delete would remove the specific endpoint with the provided name from devfile 

Fixes #3718 
Fixes #3615 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

**Automatim tests:**
run `make test-cmd-devfile-url` for integration tests and `make test` for unit tests

**Manual tests:**
`odo create java-springboot`
`odo url create url1 --port <portNumber> <any arguments you want to specify>`
should see endpoint with `url1` created in devfile.yaml, with all information provided within `odo url create`

`odo url describe url1` should see url1's info, and state is `not pushed`

`odo push` should see message that url1 is getting created. 

`odo url describe url1` should see url1's info, and state is `pushed`

`odo url list` should see url1 and the original url `8080/tcp` defined in springoot devfile.

`odo url delete url1 ` should successfully delete the url info saved in devfile.yaml and env.yaml


pkg/devfile/adapters/kubernetes/utils/utils_test.go
```diff
@@ -620,5 +622,152 @@ func TestUpdateContainersWithSupervisord(t *testing.T) {
...
+			wantMap: map[string]map[string]common.Endpoint{
+				"testcontainer1": map[string]common.Endpoint{
+					urlName: common.Endpoint{
+						Name:       urlName,
+						TargetPort: 8080,
+						Secure:     false,
+					},
+				},
+			},
+			devfileData: &v200.Devfile200{
```
mik-dass, line 647: use `testingutil.TestDevfileData()` instead `v200.Devfile200{}`

pkg/odo/cli/url/create.go
```diff
@@ -229,6 +276,27 @@ func (o *URLCreateOptions) Validate() (err error) {
...
+			}
+			if len(o.protocol) > 0 {
+				errorList = append(errorList, "endpoint protocol is not supported in docker kind")
+			}
+			if len(o.path) > 0 {
+				errorList = append(errorList, "endpoint path is not supported in docker kind")
+			}
+			if o.secureURL {
+				errorList = append(errorList, "secure endpoint is not supported in docker kind")
+			}
```
mik-dass, line 291: these flags are not present for the docker preference, no need to check

pkg/odo/cli/url/create.go
```diff
@@ -284,8 +351,73 @@ func (o *URLCreateOptions) Run() (err error) {
...
+				if _, exist := endpointMap[o.urlName]; exist {
+					if !o.forceFlag && !ui.Proceed(fmt.Sprintf("URL %v already exist in devfile endpoint entry under container %v. Do you want to override the endpoint", o.urlName, containerName)) {
+						log.Info("Aborted by the user")
+						return nil
+					}
+					if len(o.container) > 0 && o.container != containerName {
+						delete(endpointMap, o.urlName)
+					}
+					containerEndpointMap[containerName] = endpointMap
+				}
```
mik-dass, line 375: I tried to create a URL with the same name and this part was not executed. It failed in the validate() part. Also since other commands don't support overriding in create, I think we can remove it from here.

pkg/odo/cli/url/create.go
```diff
@@ -284,8 +351,73 @@ func (o *URLCreateOptions) Run() (err error) {
...
+							return fmt.Errorf("cannot set URL %v under container %v, TargetPort %v is being used for endpoint %v under container %v", o.urlName, o.container, o.componentPort, endpoint.Name, containerName)
+						} else {
+							o.container = containerName
+						}
+						break
+					}
+				}
+			}
+			if reflect.DeepEqual(containerEndpointMap, map[string]map[string]common.Endpoint{}) {
+				// devfile have no containers had endpoint entry
```
mik-dass, line 390: `have no containers and entrypoint`? or `with entrypoint`?

pkg/url/url.go
```diff
@@ -1086,3 +1125,19 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
+
+// UpdateEndpointsInDevfile writes the provided endpoint information into devfile
+func UpdateEndpointsInDevfile(devObj parser.DevfileObj, containerEndpointMap map[string]map[string]parsercommon.Endpoint) error {
+	components := devObj.Data.GetComponents()
+	for _, component := range components {
+		if component.Container != nil {
+			var endpoints []common.Endpoint
+			for _, endpoint := range containerEndpointMap[component.Container.Name] {
+				endpoints = append(endpoints, endpoint)
+			}
```
mik-dass, line 1137: Why not just pass the container name and endpoint and use it for matching and appending instead of looping? 

I know it won't work for delete but we can have a separate delete function which just takes the url name.

yangcao77: It's because initially I want to support overriding the existing endpoint name in devfile.yaml, then this function need to support both deletion and creation. 
Since we do not want to provide overwriting of existed endpoint name. it can be separated into two separate functions for add & remove respectively.

pkg/devfile/adapters/kubernetes/utils/utils_test.go
yangcao77: updated

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/url/url.go

pkg/devfile/parser/configurables.go
```diff
@@ -236,7 +236,8 @@ func addEndpoints(current []common.Endpoint, other []common.Endpoint) []common.E
 
 			protocol := presentep.Protocol
 			if protocol == "" {
-				protocol = "tcp"
+				// endpoint protocol efault value is http
```
mik-dass, line 239: default

pkg/odo/cli/url/create.go
```diff
@@ -123,6 +134,24 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
 				o.urlType = envinfo.ROUTE
 			}
+			if len(o.protocol) == 0 {
+				// endpoint protocol default value is http
+				o.protocol = string(common.HTTP)
+			}
+			if len(o.exposure) == 0 {
+				// endpoint exposure default value is public
+				o.exposure = string(common.Public)
+			}
```
mik-dass, line 144: set these defaults during flag initialization e.g
`			urlCreateCmd.Flags().StringVarP(&o.protocol, "protocol", "", string(common.HTTP), "protocol for this URL")
`

pkg/odo/cli/url/create.go
```diff
@@ -344,6 +448,11 @@ func NewCmdURLCreate(name, fullName string) *cobra.Command {
 			urlCreateCmd.Flags().StringVar(&o.tlsSecret, "tls-secret", "", "TLS secret name for the url of the component if the user bring their own TLS secret")
 			urlCreateCmd.Flags().StringVarP(&o.host, "host", "", "", "Cluster IP for this URL")
 			urlCreateCmd.Flags().BoolVar(&o.wantIngress, "ingress", false, "Create an Ingress instead of Route on OpenShift clusters")
+			urlCreateCmd.Flags().BoolVarP(&o.secureURL, "secure", "", false, "Create a secure URL")
+			urlCreateCmd.Flags().StringVarP(&o.exposure, "exposure", "", "", "URL exposure")
```
mik-dass, line 452: I feel we don't need to provide a exposure flag option to the use and instead make all URLs public.

pkg/odo/cli/url/create.go
```diff
@@ -284,8 +337,59 @@ func (o *URLCreateOptions) Run() (err error) {
...
+				}
+			}
+			if reflect.DeepEqual(containerEndpointMap, map[string]map[string]common.Endpoint{}) {
+				// devfile have no containers with endpoints
+				// pick the first container to store the new enpoint
+				o.container = o.devfileContainers[0].Name
+			}
+			if len(o.container) == 0 {
+				o.container = firstContainer
+			}
```
mik-dass, line 374: move this into Complete()

mik-dass, line 374: I think this operation can be moved into the Complete() function and we can use the loop in https://github.com/openshift/odo/pull/3857/files#diff-476c0d834fc648caad7c721d6b10694bR207 to perform the validation by using the container ports.

pkg/odo/cli/url/delete.go
```diff
@@ -117,6 +144,18 @@ func (o *URLDeleteOptions) Run() (err error) {
...
+			// for containerName, endpointMap := range containerEndpointMap {
+			// 	if _, exist := endpointMap[o.urlName]; exist {
+			// 		// delete(endpointMap, o.urlName)
+			// 		err = url.RemoveEndpointInDevfile(o.devObj, o.urlName, containerName)
+			// 		if err != nil {
+			// 			return errors.Wrapf(err, "failed to write endpoints information into devfile")
+			// 		}
+			// 		break
+			// 	}
+			// }
```
mik-dass, line 158: you forgot to remove these comments.

pkg/odo/cli/url/delete.go
```diff
@@ -77,15 +83,36 @@ func (o *URLDeleteOptions) Complete(name string, cmd *cobra.Command, args []stri
...
 	if experimental.IsExperimentalModeEnabled() {
-		urls := o.EnvSpecificInfo.GetURL()
+
 		componentName := o.EnvSpecificInfo.GetName()
-		for _, url := range urls {
-			if url.Name == o.urlName {
+
+		devObj, err := devfile.ParseAndValidate(o.DevfilePath)
+		if err != nil {
+			return fmt.Errorf("fail to parse the devfile %s, with error: %s", o.DevfilePath, err)
```
mik-dass, line 91: `failed`

pkg/odo/cli/url/describe.go
```diff
@@ -38,16 +42,21 @@ type URLDescribeOptions struct {
...
 // NewURLDescribeOptions creates a new NewURLDescribeOptions instance
 func NewURLDescribeOptions() *URLDescribeOptions {
-	return &URLDescribeOptions{&config.LocalConfigInfo{}, "", "", &genericclioptions.Context{}}
+	return &URLDescribeOptions{&config.LocalConfigInfo{}, "", "", &genericclioptions.Context{}, "", false}
 }
 
 // Complete completes URLDescribeOptions after they've been Listed
 func (o *URLDescribeOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	if experimental.IsExperimentalModeEnabled() {
+	o.devfilePath = clicomponent.DevfilePath
```
mik-dass, line 56: Join the filepath with o.componentContext otherwise context won't work
```
[mrinaldas@localhost project]$ odo url describe example-11
I0903 12:31:27.064141   17431 cert_rotation.go:137] Starting client certificate rotation controller
NAME           STATE          URL                       PORT     SECURE     KIND
example-11     Not Pushed     http://example-11.com     3000     false      ingress
[mrinaldas@localhost project]$ cd ..
[mrinaldas@localhost ~]$ odo url describe example-11 --context project
I0903 12:31:43.739120   17482 cert_rotation.go:137] Starting client certificate rotation controller
 ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directo...
```

pkg/odo/cli/url/list.go
```diff
@@ -55,8 +56,9 @@ func NewURLListOptions() *URLListOptions {
 
 // Complete completes URLListOptions after they've been Listed
 func (o *URLListOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
-	o.isDevfile = util.CheckPathExists(filepath.Join(o.componentContext, component.DevfilePath))
-	if o.isDevfile {
+	o.devfilePath = clicomponent.DevfilePath
```
mik-dass, line 59: join the filepath with o.componentContext here

pkg/url/url.go
```diff
@@ -110,33 +117,56 @@ func GetIngressOrRoute(client *occlient.Client, kClient *kclient.Client, envSpec
...
+			if !routeSupported && envURL.Kind == envinfo.ROUTE {
+				return URL{}, errors.New(fmt.Sprintf("the url %v is defined with type of Route, but Route is not support in current cluster", urlName))
+			}
+			devfileURL = envURL
+			devfileURL.Port = int(localEndpoint.TargetPort)
+			devfileURL.Secure = localEndpoint.Secure
+		}
+		if reflect.DeepEqual(devfileURL, envinfo.EnvInfoURL{}) {
+			// Devfile endpoint by default should create a route if no host information is provided in env.yaml
+			// If it is not openshify cluster, should ignore the endpoint entry when executing url describe/list
```
mik-dass, line 145: openshift

pkg/url/url.go
```diff
@@ -110,33 +117,56 @@ func GetIngressOrRoute(client *occlient.Client, kClient *kclient.Client, envSpec
...
-					clusterURL.Status.State = StateTypePushed
-					return clusterURL, nil
-				} else if route != nil {
-					clusterURL := getMachineReadableFormat(*route)
-					clusterURL.Status.State = StateTypePushed
-					return clusterURL, nil
-				}
-			} else {
-				localURL.Status.State = StateTypeNotPushed
+		for _, envURL := range envinfoURLs {
```
mik-dass, line 129: put these in a map and fetch from it

pkg/url/url.go
```diff
@@ -88,12 +90,17 @@ func Get(client *occlient.Client, localConfig *config.LocalConfigInfo, urlName s
 }
 
 // GetIngressOrRoute returns ingress/route spec for given URL name
-func GetIngressOrRoute(client *occlient.Client, kClient *kclient.Client, envSpecificInfo *envinfo.EnvSpecificInfo, urlName string, componentName string, routeSupported bool) (URL, error) {
+func GetIngressOrRoute(client *occlient.Client, kClient *kclient.Client, envSpecificInfo *envinfo.EnvSpecificInfo, urlName string, containerEndpointsMap map[string]map[string]parsercommon.Endpoint, componentName string, routeSupported bool) (URL, error) {
```
mik-dass, line 93: instead of `containerEndpointsMap` I think we can use []common.DevfileComponent, that way in a best/average case  we won't need to loop over the entire component data to form the map.

pkg/url/url.go
```diff
@@ -512,16 +542,46 @@ func ListIngressAndRoute(oclient *occlient.Client, client *kclient.Client, envSp
...
-		if !routeSupported && envinfoURL.Kind == envinfo.ROUTE {
-			continue
+	for _, endpointMap := range containerEndpointsMap {
+		for _, localEndpoint := range endpointMap {
+			// only exposed endpoint will be shown as a URL in `odo url list`
+			if localEndpoint.Exposure == common.None || localEndpoint.Exposure == common.Internal {
+				continue
+			}
+			var devfileURL envinfo.EnvInfoURL
+			for _, envinfoURL := range localEnvinfoURLs {
```
mik-dass, line 552: put these URLs in a map and fetch from it

pkg/url/url.go
```diff
@@ -512,16 +542,46 @@ func ListIngressAndRoute(oclient *occlient.Client, client *kclient.Client, envSp
...
+				}
+				devfileURL.Name = localEndpoint.Name
+				devfileURL.Port = int(localEndpoint.TargetPort)
+				devfileURL.Secure = localEndpoint.Secure
+				devfileURL.Kind = envinfo.ROUTE
+			}
+			localURL := ConvertEnvinfoURL(devfileURL, componentName)
+			// use the trimmed URL Name as the key since remote URLs' names are trimmed
+			trimmedURLName := strings.TrimSpace(util.GetDNS1123Name(strings.ToLower(localURL.Name)))
+			trimmedURLName = util.TruncateString(trimmedURLName, 15)
```
mik-dass, line 581: These steps are present in GetIngressOrRoute() too. Please put them in a function to maintain consistency.

pkg/url/url.go
```diff
@@ -904,15 +959,42 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
-						Secure:    secure,
-						TLSSecret: url.TLSSecret,
-						Kind:      url.Kind,
-						Path:      path,
-					},
+				name := strings.TrimSpace(util.GetDNS1123Name(strings.ToLower(endpoint.Name)))
+				name = util.TruncateString(name, 15)
+				envUrls := parameters.EnvURLS
+				existInEnv := false
+				for _, url := range envUrls {
```
mik-dass, line 966: put these in a map outside the loops and fetch from it

pkg/url/url.go
```diff
@@ -1086,3 +1125,38 @@ func Push(client *occlient.Client, kClient *kclient.Client, parameters PushParam
...
+			component.Container.Endpoints = append(component.Container.Endpoints, endpoint)
+			devObj.Data.UpdateComponent(component)
+			break
+		}
+	}
+	return devObj.WriteYamlDevfile()
+}
+
+// RemoveEndpointInDevfile deletes the provided endpoint information from devfile
+func RemoveEndpointInDevfile(devObj parser.DevfileObj, urlName string, container string) error {
```
mik-dass, line 1143: I think we don't need the containerName since to find it we are looping over the containers and their endpoints once in GetContainerEndpoints(). Instead perform the loop here and break once the first endpoint with the name is found.

pkg/odo/cli/url/create.go
yangcao77: Makes sense.

pkg/url/url.go
yangcao77: Why does it matter? Convert a slice into a map needs to iterating through the entire slice. 
We can break the loop once we find the matching name when looping through the slice inside.

pkg/url/url.go
yangcao77: Why does it matter? Convert a slice into a map needs to iterating through the entire slice. 
We can break the loop once we find the matching name when looping through the slice inside.

pkg/url/url.go
mik-dass: This is currently a O(n3) without any breaks. We can break after finding the URL, but if the endpoint, we are looking for, is in the last container of containerEndpointsMap and the last endpoint of the corresponding container, we will unnecessarily loop over localEnvinfoURLs multiple times. Thus looping over localEnvinfoURLs once and storing it once in a map is much more efficient and prevents the unnecessary iterations even in the worst case scenario.

pkg/url/url.go
mik-dass: same as https://github.com/openshift/odo/pull/3857#discussion_r483126733

cdrage:
Only comments I have were the same as Mrinal, but the overall implementation looks great. Will fix the problem I faced here: https://github.com/openshift/odo/issues/3893

pkg/devfile/parser/configurables.go
yangcao77: updated

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/odo/cli/url/delete.go
yangcao77: updated

pkg/odo/cli/url/delete.go
```diff
@@ -77,15 +83,36 @@ func (o *URLDeleteOptions) Complete(name string, cmd *cobra.Command, args []stri
...
+		}
+		o.devObj = devObj
+		containerEndpointMap, err := adapterutils.GetContainerEndpoints(devObj.Data)
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/odo/cli/url/delete.go
yangcao77: updated

pkg/odo/cli/url/describe.go
yangcao77: updated

pkg/odo/cli/url/list.go
yangcao77: updated

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go

pkg/url/url.go
```diff
@@ -512,16 +542,46 @@ func ListIngressAndRoute(oclient *occlient.Client, client *kclient.Client, envSp
...
 	}
-	for _, envinfoURL := range localEnvinfoURLs {
-		// only checks for Ingress and Route URLs
-		if envinfoURL.Kind == envinfo.DOCKER {
-			continue
-		}
-		if !routeSupported && envinfoURL.Kind == envinfo.ROUTE {
-			continue
+	for _, endpointMap := range containerEndpointsMap {
+		for _, localEndpoint := range endpointMap {
```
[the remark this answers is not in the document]
yangcao77: updated. also updated the loop in Push()

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -78,8 +78,14 @@ func GetContainers(devfileObj devfileParser.DevfileObj) ([]corev1.Container, err
 		for _, c := range containers {
 			for _, containerPort := range c.Ports {
 				for _, curPort := range container.Ports {
+					if curPort.Name == containerPort.Name {
+						// the name has to be unique across containers since it is considered as the URL name
+						return nil, fmt.Errorf("Devfile contains multiple endpoint entries with same name: %v", containerPort.Name)
+					}
```
mik-dass, line 84: Do we need this? ConvertPorts() seems to already do this check and it is called above before the loop.

pkg/url/url_test.go
```diff
@@ -1645,17 +1722,23 @@ func TestPush(t *testing.T) {
...
-					Name:       "example",
-					TargetPort: 8080,
-					Secure:     false,
-					Path:       "/testpath",
+			containerComponents: []versionsCommon.DevfileComponent{
+				{
+					Container: &versionsCommon.Container{
+						Name: "container1",
+						Endpoints: []versionsCommon.Endpoint{
+							versionsCommon.Endpoint{
```
mik-dass, line 1734: we don't need to type `versionsCommon.Endpoint` here and in the above cases

pkg/url/url_test.go
```diff
@@ -1490,19 +1522,28 @@ func TestPush(t *testing.T) {
...
 					Kind: envinfo.ROUTE,
 				},
 			},
-			endpintMap: map[int32]versionsCommon.Endpoint{
-				8080: versionsCommon.Endpoint{
-					Name:       "example",
-					TargetPort: 8080,
-					Secure:     false,
-					Exposure:   "internal",
+			containerComponents: []versionsCommon.DevfileComponent{
```
mik-dass, line 1528: this case seems to be the same case as above as suggested from the name of the case

pkg/url/url_test.go
```diff
@@ -2584,3 +2776,418 @@ func TestConvertEnvinfoURL(t *testing.T) {
...
+			jsonval, err := devfileCtx.YAMLToJSON(retval)
+			if err != nil {
+				t.Errorf("Unexpected err from YAMLToJSON: %v", err)
+			}
+			var newData testingutil.TestDevfileData
+
+			err = json.Unmarshal(jsonval, &newData)
+			if err != nil {
+				t.Errorf("Unexpected err from json.Unmarshal: %v", err)
+			}
```
mik-dass, line 3186: these seem to be the same steps as above. Let's put them in a function.

tests/integration/devfile/cmd_devfile_url_test.go
```diff
@@ -188,21 +181,34 @@ var _ = Describe("odo devfile url command tests", func() {
...
+			stdOut := helper.CmdShouldFail("odo", "url", "create", "--host", "https://127.0.0.1:60104", "--tls-secret", "foo", "--port", "3000", "--ingress")
 			Expect(stdOut).To(And(ContainSubstring("is not a valid host name"), ContainSubstring("TLS secret is only available for secure URLs of Ingress kind")))
 		})
+		It("should not allow creating under an invalid container", func() {
+			containerName := helper.RandString(5)
+			helper.CmdShouldPass("odo", "create", "nodejs", "--project", namespace)
+			stdOut := helper.CmdShouldFail("odo", "url", "create", "--host", "com", "--port", "3000", "--container", containerName, "--ingress")
+			Expect(stdOut).To(ContainSubstring(fmt.Sprintf("The container specified: %s does not exist in devfile", containerName)))
+		})
+		It("should not allow creating an endpoint with same name", func() {
```
mik-dass, line 203: Please add a new line between the `It`s

mik-dass:
After deleting a URL, the message is
```
[mrinaldas@localhost project]$ odo url delete http-3000 -f
 ✓  URL http-3000 removed from the env file

To delete the URL on the cluster, please use `odo push`
```
I think this should be `from the devfile`

mik-dass:
It seems protocol has no affect as the container ports are always `tcp`
```
                "ports": [
                    {
                        "containerPort": 3000,
                        "name": "example-2",
                        "protocol": "TCP"
                    },
                    {
                        "containerPort": 3000,
                        "name": "example-3",
                        "protocol": "TCP"
                    },
                    {
                        "containerPort": 3000,
                        "name": "example-4",
                        "protocol": "TCP"
                    },
                    {
                        "containerPort": 3000,
                        "name": "example-5",
                        "protocol": "TCP"
                    }
```
In the devfile.yaml
```
- container:
    endpoints:
    - exposure: public
      path: /
      secure: true
      name: example-2
      targetPort: 3000
      protocol: http
    - exposure: public
      path: /data
      name: example-3
      targetPort: 3000
      protocol: http
    - exposure: public
      path: /
      name: example-4
      targetPort: 3000
      protocol: tcp
    - exposure: public
      path: /
      name: example-5
      targetPort: 3000
      protocol: wss
    memoryLimit: 1024Mi
    mountSources: true
    name: runtime
    sourceMapping: /project
```

mik-dass:
Also it might be a good idea to display the protocol and path in the `list`, `describe` and the json output of both of them.
```
[mrinaldas@localhost project]$ odo url list 
Found the following URLs for component nodejs
NAME          STATE      URL                                                              PORT     SECURE     KIND
example-2     Pushed     https://example-2-nodejs-blah.b6ff.rh-idev.openshiftapps.com     3000     true       route
example-3     Pushed     http://example-3-nodejs-blah.b6ff.rh-idev.openshiftapps.com      3000     false      route
example-4     Pushed     http://example-4-nodejs-blah.b6ff.rh-idev.openshiftapps.com      3000     false      route
example-5     Pushed     https://example-5-nodejs-blah.b6ff.rh-idev.openshiftapps.com     3000     true       route
[mrinaldas@localhost project]$ odo url describe example-4
NAME          STATE      URL                                                             PORT     SECURE     KIND
example-4     Pushed     http://example-4-nodejs-blah.b6ff.rh-idev.openshiftapps.com     3000     false      route 
```

mik-dass:
One of the tests failed on Travis 
https://travis-ci.com/github/openshift/odo/jobs/381353550#L1147

kadel:
> It seems protocol has no affect as the container ports are always `tcp`

This is correct.
http, https and wss are application protocols that are using TCP as a transport layer. Problem is that devfile is mixing application and transport protocols into one field and kubernetes protocol is transport layer protocol (tcp/udp)

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: As described in the comments, ConvertPorts() only checks for duplicate endpoint name within same container. and the check here checks for duplication across containers.

mik-dass: Sorry didn't notice the comment

pkg/url/url_test.go
yangcao77: updated

pkg/url/url_test.go
yangcao77: Good catch. Deleted the duplicate test case

pkg/url/url_test.go
yangcao77: This part was for the old implementation which uses the containerEndpointMap. The latest implementation does not need this part. Deleted.

tests/integration/devfile/cmd_devfile_url_test.go
yangcao77: Updated

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -52,8 +51,9 @@ func ConvertPorts(endpoints []common.Endpoint) ([]corev1.ContainerPort, error) {
 		name := strings.TrimSpace(util.GetDNS1123Name(strings.ToLower(endpoint.Name)))
 		name = util.TruncateString(name, 15)
 		for _, c := range containerPorts {
-			if c.ContainerPort == endpoint.TargetPort {
-				return nil, fmt.Errorf("Devfile contains multiple identical ports: %v", endpoint.TargetPort)
+			if c.Name == endpoint.Name {
+				// the name has to be unique within a single container since it is considered as the URL name
+				return nil, fmt.Errorf("Devfile contains multiple endpoint entries with same name: %v", endpoint.Name)
```
cdrage, line 56: remove capitalization from error

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -78,8 +78,14 @@ func GetContainers(devfileObj devfileParser.DevfileObj) ([]corev1.Container, err
 		for _, c := range containers {
 			for _, containerPort := range c.Ports {
 				for _, curPort := range container.Ports {
+					if curPort.Name == containerPort.Name {
+						// the name has to be unique across containers since it is considered as the URL name
+						return nil, fmt.Errorf("Devfile contains multiple endpoint entries with same name: %v", containerPort.Name)
```
cdrage, line 83: remove capitalization from error

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -78,8 +78,14 @@ func GetContainers(devfileObj devfileParser.DevfileObj) ([]corev1.Container, err
...
+					}
 					if curPort.ContainerPort == containerPort.ContainerPort {
-						return nil, fmt.Errorf("Devfile contains multiple identical ports: %v", containerPort.ContainerPort)
+						// the same TargetPort present in different containers
+						// because containers in a single pod shares the network namespace
+						return nil, fmt.Errorf("Devfile contains multiple containers with same TargetPort: %v", containerPort.ContainerPort)
```
cdrage, line 88: remove capitalization from error

pkg/odo/cli/url/create.go
```diff
@@ -123,6 +131,16 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
 			} else {
 				o.urlType = envinfo.ROUTE
 			}
+			if len(o.path) > 0 && (strings.HasPrefix(o.path, "/") || strings.HasPrefix(o.path, "\\")) {
```
cdrage, line 134: Errrrrr... is there a better way to check for this? I think you should be using https://golang.org/pkg/path/filepath/#FromSlash so you can have a uniform return regardless of front or backslashes.

pkg/odo/cli/url/create.go
```diff
@@ -133,49 +151,90 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
+			compWithEndpoint := 0
+			for _, c := range containers {
+				if len(c.Ports) != 0 {
+					compWithEndpoint++
+					for _, port := range c.Ports {
+						portList = append(portList, strconv.FormatInt(int64(port.ContainerPort), 10))
+					}
+				}
+				if compWithEndpoint > 1 {
+					return fmt.Errorf("Devfile should only have one component containing endpoint")
```
cdrage, line 177: Remove capitalization from error output

pkg/odo/cli/url/create.go
```diff
@@ -133,49 +151,90 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
+				}
+			}
+			if compWithEndpoint == 0 {
+				return fmt.Errorf("No valid component with an endpoint found in the devfile")
```
cdrage, line 181: Remove capitalization from error output

pkg/odo/cli/url/create.go
```diff
@@ -133,49 +151,90 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
 			}
 			o.urlType = envinfo.DOCKER
-		}
-
-		if len(args) != 0 {
-			o.urlName = args[0]
 		} else {
-			o.urlName = url.GetURLName(componentName, o.componentPort)
+			if o.urlPort == -1 {
+				return fmt.Errorf("Port must be provided to create an endpoint entry in devfile")
```
cdrage, line 194: Remove capitalization from error output

pkg/odo/cli/url/create.go
```diff
@@ -133,49 +151,90 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
+			if len(args) != 0 {
+				o.urlName = args[0]
+			} else {
+				o.urlName = url.GetURLName(componentName, o.componentPort)
+			}
+
+			foundContainer := false
+			containerComponents := adaptersCommon.GetDevfileContainerComponents(devObj.Data)
+			if len(containerComponents) == 0 {
+				return fmt.Errorf("No valid components found in the devfile")
```
cdrage, line 206: Remove capitalization from error output

pkg/odo/cli/url/create.go
```diff
@@ -133,49 +151,90 @@ func (o *URLCreateOptions) Complete(_ string, cmd *cobra.Command, args []string)
...
+				}
+				for _, endpoint := range component.Container.Endpoints {
+					if endpoint.Name == o.urlName {
+						return fmt.Errorf("url %v already exist in devfile endpoint entry under container %v", o.urlName, component.Container.Name)
+					}
+					containerPortMap[int(endpoint.TargetPort)] = component.Container.Name
+				}
+			}
+			if len(o.container) > 0 && !foundContainer {
+				return fmt.Errorf("The container specified: %v does not exist in devfile", o.container)
```
cdrage, line 224: Remove capitalization from error output

cdrage (changes requested):
The code looks great! I do have some minor changes, but nothing fundamental to the actual code.

pkg/odo/cli/url/create.go
yangcao77: I think  `FromSlash` only checks for `/` and replace with the current operating system's file separator. This check here is to check and remove if there is a leading front or backslashes, in order to add `/` in front of the path in general. As a user can enter either `pathhewants` or `/pathhewants` as the path during url creation.

yangcao77:
I have to modify the integration test for s2i to devfile conversion due to the endpoint name is not converted properly. 
Opened an issue to track: https://github.com/openshift/odo/issues/3910

pkg/odo/cli/url/create.go

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: updated

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: updated

pkg/devfile/adapters/kubernetes/utils/utils.go
yangcao77: updated

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

pkg/odo/cli/url/create.go

adisky:
> I have to modify the integration test for s2i to devfile conversion due to the endpoint name is not converted properly.
> Opened an issue to track: #3910

It has to be reworked upon new style of devfile creation.
https://github.com/openshift/odo/blob/master/pkg/odo/cli/utils/convert.go#L239

mik-dass:
> This is correct.
http, https and wss are application protocols that are using TCP as a transport layer. Problem is that devfile is mixing application and transport protocols into one field and kubernetes protocol is transport layer protocol (tcp/udp)

OK let's keep track of UDP support in #3663

mik-dass (approved):
Works for me :+1: 
/lgtm

PR #3885, opened 2020-09-02
Sync operators on 4.* cluster

prietyc123:
**What type of PR is this?**

/kind bug
/kind feature

Operators are unstable with 4.6 cluster which causing our operator tests to get failed. However there is a common change in start csv version for service binding in all over 4.* cluster. This pr should handle those either by just removing the unstable operators or using workarounds like fetching the operators into the cluster.

Fixes NA

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

`make test-operator-hub` should pass.

### DIFF
--- a/scripts/setup-operators.sh
+++ b/scripts/setup-operators.sh
@@ -52,7 +52,7 @@ install_service_binding_operator(){
     name: service-binding-operator
     source: community-operators
     sourceNamespace: openshift-marketplace
-    startingCSV: service-binding-operator.v0.1.1-352
+    startingCSV: service-binding-operator.v0.1.1-364
 EOF
 }
 


dharmit:
@prietyc123 one thing I've observed in my local setup is that it's not essential to specify `startingCSV`. In absence of this field, Operator Hub automatically installs the one which is latest. Can you try if that works for you? If it does, we can remove `startingCSV` from the `setup-operators.sh` script.

prietyc123:
> Can you try if that works for you? If it does, we can remove startingCSV from the setup-operators.sh script.

@dharmit I have verified on 4.* cluster including 4.6 . So basically `startingCSV` is the mandatory requirement on 4.6 cluster. Although it works fine for 4.* except 4.6 .

I have used this subscription yaml with startingCSV
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: service-binding-operator
  namespace: openshift-operators
spec:
  channel: alpha
  installPlanApproval: Automatic
  name: service-binding-operator
  source: community-operators
  sourceNamespace: openshift-marketplace
```
Yes, it will create the subscription but not install the operator.
```
$ oc create -f service-binding-operator.yaml 
subscription.operators.coreos.com/service-binding-operator created

$ oc get subscription -n openshift-operators
NAME                                  PACKAGE                               SOURCE                          CHANNEL
service-binding-operator              service-binding-operator              community-operators             alpha

$ oc get csv -n openshift-operators
No resources found in openshift-operators namespace
```

PR #3900, opened 2020-09-04
Improve URL format validation

### LINKED ISSUE #3451: "odo registry add" adds registry for invalid url in devfileV2
/kind bug

**Operating System:**
All

**Output of `odo version`:**
master

Steps to reproduce:

1. Enable experimental mode `odo preference set Experimental true`
2. `odo registry add wRegistry "http://api.google.com/q?exp=a|b"` with invalid url
3. `odo registry list`

## Actual behavior

```
$ odo preference view
UpdateNotification    false
NamePrefix            
Timeout               
PushTimeout           
Experimental          true

$ odo registry add wRegistry "http://api.google.com/q?exp=a|b"
New registry successfully added

$ odo registry listNAME                       URL
CheDevfileRegistry         https://che-devfile-registry.openshift.io
demoRegistry               http://docker.io
wrongRegistry              http://www.xyz.com
wRegistry                  http://api.google.com/q?exp=a|b
```

## Expected behavior

`odo registry add` should throw error for invalid url

Manually verified.

GeekArthur:

**What type of PR is this?**
/kind bug
/area devfile

This PR adds more requirements to URL format validation so that gate more invalid registry URLs

Fixes #3451 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test  

Run `odo registry add test <invalid URL>`, for example run `odo registry add test http://$$$$,,,,,@hdjh.hcgcyd `

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1141,8 +1141,15 @@ func ValidateURL(sourceURL string) error {
 	if err != nil {
 		return err
 	}
+	scheme := u.Scheme
+	user := u.User.Username()
+	host := u.Host
 
-	if len(u.Host) == 0 || len(u.Scheme) == 0 {
+	// Valid URL needs to satisfy the following requirements:
+	// 1. URL has scheme and host components
+	// 2. Scheme, host and user (if applicable) components of the URL shouldn't contain reserved character
+	re := regexp.MustCompile(`[\/;?:@&=+$,]`)
+	if scheme == "" || host == "" || re.MatchString(scheme) || re.MatchString(host) || re.MatchString(user) {
 		return errors.New("URL is invalid")
 	}
 
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1137,19 +1137,17 @@ func CheckKubeConfigExist() bool {
 
 // ValidateURL validates the URL
 func ValidateURL(sourceURL string) error {
-	u, err := url.Parse(sourceURL)
+	// Valid URL needs to satisfy the following requirements:
+	// 1. URL has scheme and host components
+	// 2. Scheme, host of the URL shouldn't contain reserved character
+	url, err := url.ParseRequestURI(sourceURL)
 	if err != nil {
 		return err
 	}
-	scheme := u.Scheme
-	user := u.User.Username()
-	host := u.Host
+	host := url.Host
 
-	// Valid URL needs to satisfy the following requirements:
-	// 1. URL has scheme and host components
-	// 2. Scheme, host and user (if applicable) components of the URL shouldn't contain reserved character
-	re := regexp.MustCompile(`[\/;?:@&=+$,]`)
-	if scheme == "" || host == "" || re.MatchString(scheme) || re.MatchString(host) || re.MatchString(user) {
+	re := regexp.MustCompile(`[:\/\?#\[\]@]`)
+	if host == "" || re.MatchString(host) {
 		return errors.New("URL is invalid")
 	}
 
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -1766,17 +1766,12 @@ func TestValidateURL(t *testing.T) {
 		},
 		{
 			name:    "Case 4: Invalid URL - Host contains reserved character",
-			url:     "http://$$,,",
+			url:     "http://??##",
 			wantErr: true,
 		},
 		{
 			name:    "Case 5: Invalid URL - Scheme contains reserved character",
 			url:     "$$$,,://www.example.com",
-			wantErr: true,
-		},
-		{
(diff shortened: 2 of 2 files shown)


pkg/util/util.go
```diff
@@ -1141,8 +1141,15 @@ func ValidateURL(sourceURL string) error {
 	if err != nil {
 		return err
 	}
-
-	if len(u.Host) == 0 || len(u.Scheme) == 0 {
+	scheme := u.Scheme
```
adisky, line 1144: @GeekArthur can we use  `url.ParseRequestURI` in place of `url.Parse`?  This seems to check for the validation we are using
https://golang.org/pkg/net/url/#ParseRequestURI`

dharmit:
@GeekArthur I'm able to add invalid registries like below:
```sh
$ odo registry list
NAME                       URL                                                                SECURE
test3                      https://)(url                                                      No
test2                      noprotocol://noregistry                                            No
test                       https://nosuchurlblah                                              No
```

While the last one could be possible in an internal environment (and hence a valid URL for some environment), the other two seem far-fetched and I tried them just for the sake of breaking it. I don't claim to know the definition of 100% correct URL/URI but unless user uses their commonsense, there's going to be ways break this.

However, while looking around, I did see examples where `ParseRequestURI` is used as @adisky suggested in her review. If you could make the change or comment why it's not reasonable to do that, this PR looks good to approve to me. :+1:

mohammedzee1000:
> @GeekArthur I'm able to add invalid registries like below:
> 
> ```shell
> $ odo registry list
> NAME                       URL                                                                SECURE
> test3                      https://)(url                                                      No
> test2                      noprotocol://noregistry                                            No
> test                       https://nosuchurlblah                                              No
> ```
> 
> While the last one could be possible in an internal environment (and hence a valid URL for some environment), the other two seem far-fetched and I tried them just for the sake of breaking it. I don't claim to know the definition of 100% correct URL/URI but unless user uses their commonsense, there's going to be ways break this.

The only way you can validate the 3rd case, is to see if it is reachable or not, as technically, it is a valid url, just one that may not exist or even be reachable. Do we want to do this?
> 
> However, while looking around, I did see examples where `ParseRequestURI` is used as @adisky suggested in her review. If you could make the change or comment why it's not reasonable to do that, this PR looks good to approve to me.

+1

pkg/util/util.go
GeekArthur: Yes, we can switch to use that as it already covers the scheme validation, but we still need add the host validation by ourselves.

GeekArthur:
> @GeekArthur I'm able to add invalid registries like below:
> 
> ```shell
> $ odo registry list
> NAME                       URL                                                                SECURE
> test3                      https://)(url                                                      No
> test2                      noprotocol://noregistry                                            No
> test                       https://nosuchurlblah                                              No
> ```
> 
> While the last one could be possible in an internal environment (and hence a valid URL for some environment), the other two seem far-fetched and I tried them just for the sake of breaking it. I don't claim to know the definition of 100% correct URL/URI but unless user uses their commonsense, there's going to be ways break this.
> 
> However, while looking around, I did see examples where `ParseRequestURI` is used as @adisky suggested in her review. If you could make the change or comment why it's not reasonable to do that, this PR looks good to approve to me. 👍

We can use `url.ParseRequestURI` as `url.Parse` is loose version of URL parsing so that we can add customized URL parsing code, but given currently we only validate scheme and host components of the URL so we can use `url.ParseRequestURI` + customized host validation (`url.ParseRequestURI` doesn't fully cover host validation).

@dharmit Regarding the three cases you tested, `https://)(url` and `https://nosuchurlblah` should be valid URL as they follow the RFC 3986 (https://www.ietf.org/rfc/rfc3986.txt, FYI definition of 100% correct URL/URI is there but TL;DR), which is the latest global URL/URI standard. I know this one `https://)(url` is pretty tricky but if the user really wants to name the web host like that to use `(` or `)`, I believe it works as those characters are not reserved characters for host component of URL according to RFC 3986. For `noprotocol://noregistry`, I don't think currently there is a supported scheme list so we just use the scheme validation from `url.ParseRequestURI.

@mohammedzee1000 
> The only way you can validate the 3rd case, is to see if it is reachable or not, as technically, it is a valid url, just one that may not exist or even be reachable. Do we want to do this?

We decide to follow the design of `yum-config-manager`, which we don't validate the accessibility when adding the registry, we only validate the basic URL format when on adding URL stage.

GeekArthur:
@adisky @dharmit @mohammedzee1000 Comments addressed, please review again, thanks.

GeekArthur:
@dharmit All tests passed, can you review again?

GeekArthur:
This PR also needs an `approved` label, @kadel can you help add that?

Issue #3971, opened 2020-09-16
Update debugging page on how to debug via application + IDEs


cdrage:
/kind documentation

## What mistake did you find / what is missing in the documentation?

There are no instructions on how to debug via the IDE. We should add some material / instructions on how to do so.

## What is the relevance of it?

No IDE references / how to properly debug.

Issue #3840, opened 2020-08-26
odo debug port-forward uses wrong port


kadel:
/kind bug

## How did you run odo exactly?

Have devfile with debug command that defines env variable for debug port in debug command like this
```yaml
  - exec:
      env:
         - name: DEBUG_PORT
           value: "5859"
      commandLine: "pip install --user debugpy && python -m debugpy --listen 0.0.0.0:${DEBUG_PORT} app.py"
      id: py-debug
      workingDir: /projects
      component: py-web
      group:
        kind: debug   
```

Running
```
▶ odo debug port-forward
Started port forwarding at ports - 5858:5858
```

## Actual behavior

Using port 5858 for port forwarding

## Expected behavior
It should use 5959, because that is where the debugger is listening based on `DEBUG_PORT` env variable

## Any logs, error output, etc?

/priority medium
/area debug

prietyc123:
> It should use 5959, because that is where the debugger is listening based on DEBUG_PORT env variable

@kadel You mean 5859 according to

```
- exec:
      env:
         - name: DEBUG_PORT
           value: "5859"
      commandLine: "pip install --user debugpy && python -m debugpy --listen 0.0.0.0:${DEBUG_PORT} app.py"
      id: py-debug
      workingDir: /projects
      component: py-web
      group:
        kind: debug 
```
right?

sarveshtamba:
@prietyc123 
1.) Is this issue seen across all platforms including x86_64?
2.) Any timelines for resolving this, since this is kind of a blocker for ppc64le progress?

mik-dass:
@kadel Currently port-forward uses the DEBUG_PORT value from the env.yaml for devfile components. Should we remove  `odo env set DebugPort <value>`? Or in such a case, prioritize the devfile command defined one? Also what should `odo` do if the DEBUG_PORT is defined in the env field for the component container in the devfile?

kadel:
> @prietyc123
> 1.) Is this issue seen across all platforms including x86_64?
> 2.) Any timelines for resolving this, since this is kind of a blocker for ppc64le progress?

@sarveshtamba  problems described in  #3502 are not related to this.

kadel:
> @kadel Currently port-forward uses the DEBUG_PORT value from the env.yaml for devfile components. Should we remove `odo env set DebugPort <value>`? Or in such a case, prioritize the devfile command defined one? Also what should `odo` do if the DEBUG_PORT is defined in the env field for the component container in the devfile?

[...]

PR #3972, opened 2020-09-16
Update the debugging documentation

cdrage:
**What type of PR is this?**

/kind documentation
[skip ci]

Updates the debugging documentation with more information with example
output as well as a link to VSCode debugging tools

Closes https://github.com/openshift/odo/issues/3971

**PR acceptance criteria**:

- [ ] Unit test

- [ ] Integration test

- [X] Documentation

### DIFF
--- a/docs/public/debugging-using-devfile.adoc
+++ b/docs/public/debugging-using-devfile.adoc
@@ -1,29 +1,11 @@
-# Debugging components using devfiles
+= Debugging a component
 
-`odo` uses devfiles to build and deploy components. More information on devifles : https://redhat-developer.github.io/devfile/[Introduction to devfile]
+Debugging your component involves port forwarding with the Kubernetes pod. Before you start, it is required that you have a `kind: debug` step located within your `devfile.yaml`.
 
-In order to use `odo debug` your devfile is required to have a `debugrun` step. Example of a nodejs devfile with a debugrun step:
-
-```yaml
-schemaVersion: 2.0.0
-metadata:
-  name: nodejs
-  version: 1.0.0
-starterProjects:
-  - name: nodejs-starter
-    git:
-      remotes:
-        origin: "https://github.com/odo-devfiles/nodejs-ex.git"
-components:
-  - name: runtime
-    container:
-      image: registry.access.redhat.com/ubi8/nodejs-12:1-45
-      memoryLimit: 1024Mi
-      mountSources: true
-      sourceMapping: /project
-      endpoints:
-        - name: http-3000
-          targetPort: 3000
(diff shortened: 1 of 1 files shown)


docs/public/debugging-using-devfile.adoc
```diff
@@ -1,29 +1,11 @@
...
-      image: registry.access.redhat.com/ubi8/nodejs-12:1-45
-      memoryLimit: 1024Mi
-      mountSources: true
-      sourceMapping: /project
-      endpoints:
-        - name: http-3000
-          targetPort: 3000
+= Debugging a component
+
+Debugging your component involves port forwarding with the Kubernetes pod. Before you start, it is required that you have a `debugrun` step located within your `devfile.yaml`.
```
scottkurz, line 3: Wondering why you use `debugrun`  here instead of `debug`  ?   You need a `kind: debug` command (or did this change?).

scottkurz:
Might be helpful to add a section on debug with `odo watch`.   The fact that (IIUC) you have to do an `odo push --debug` first isn't obvious.

docs/public/debugging-using-devfile.adoc
```diff
@@ -49,18 +31,76 @@ commands:
...
+[source,sh]
+----
+$ odo debug info
 Debug is running for the component on the local port : 5858
-```
+----
+
+. Accessing the debugger:
++
+Accessing the debugger is accessible through an assortment of tools. An example of setting up a debug interface would be through link:https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging[VSCode's debugging interface].
```
maysunfaisal, line 106: Since we mentioned VSCode, can we give a small example on how to attach it to the Node debugger in this case with valid snippet? For example:
```
{
  "type": "node",
  "request": "attach",
  "name": "Attach to remote",
  "address": "TCP/IP address of process to be debugged",
  "port": 9229
}
```

maysunfaisal, line 106: `The debugger is accessible through`*

docs/public/debugging-using-devfile.adoc
mik-dass: Yes this should be any step with `kind: debug`

cdrage:
Thanks for the reviews @maysunfaisal + @scottkurz 

I have updated the documentation.

I am going to merge this for now as we need this out for 2.0.0. We can make further edits after.

scottkurz:
@cdrage I made a minor suggestion here: https://github.com/openshift/odo/pull/4039/files   though didn't get into 'watch'.   

 Noticed from https://github.com/openshift/odo/issues/3840 there is still some confusion around env so maybe better to clear that up before updating further.

PR #3964, opened 2020-09-15
odo list follow up implementation

girishramnani:
**What does does this PR do / why we need it**:
This PR implements 
- listing components without being in a context dir
- json output for devfile components
- --path for devfile components

**Which issue(s) this PR fixes**:

Fixes https://github.com/openshift/odo/issues/3521

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [X] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

PR #3990, opened 2020-09-20
updated the Kind in odo describe to be different for s2i and devfile …

girishramnani:
…component

**What type of PR is this?**

/kind api-change

Uses `DevfileComponent` for `odo describe` when a devfile component is present and `Component` for s2i 

No issue

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

test should pass

### DIFF
--- a/pkg/component/component_full_description.go
+++ b/pkg/component/component_full_description.go
@@ -3,12 +3,13 @@ package component
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
+
 	adaptersCommon "github.com/openshift/odo/pkg/devfile/adapters/common"
 	devfileParser "github.com/openshift/odo/pkg/devfile/parser"
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 	"github.com/openshift/odo/pkg/envinfo"
 	"github.com/openshift/odo/pkg/kclient"
-	"strings"
 
 	"github.com/openshift/odo/pkg/config"
 	"github.com/openshift/odo/pkg/log"
@@ -102,10 +103,6 @@ func (cfd *ComponentFullDescription) fillEmptyFields(componentDesc Component, co
 		cfd.Namespace = projectName
 	}
 
-	if len(cfd.Kind) <= 0 {
-		cfd.Kind = "Component"
-	}
-
 	if len(cfd.APIVersion) <= 0 {
 		cfd.APIVersion = apiVersion
 	}
@@ -125,8 +122,10 @@ func NewComponentFullDescriptionFromClientAndLocalConfig(client *occlient.Client
 	var err error
 	if envInfo != nil {
 		componentDesc, devfile, err = GetComponentFromDevfile(envInfo)
+		cfd.Kind = KindDevfileComponent
 	} else {
 		componentDesc, err = GetComponentFromConfig(localConfigInfo)
+		cfd.Kind = KindComponent
 	}
 	if err != nil {
 		return cfd, err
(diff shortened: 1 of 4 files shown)


dharmit:
> **Which issue(s) this PR fixes**:
> 
> No issue

Since there's no issue for this, where was this discussed? I remember we had a discussion in some call to not open PRs with no issues. And this one is an API change.



>  Uses `DevfileComponent` for `odo describe` when a devfile component is present and `Component` for s2i

Since Devfile based components are the future for odo and s2i components might get deprecated eventually, shouldn't we use special naming for s2i components instead of devfile components? Meaning, IMO, it would be more helpful to rename s2i components as `s2iComponent` and keep Devfile component as simply `Component`.

dharmit:
> > **Which issue(s) this PR fixes**:
> > No issue
> 
> Since there's no issue for this, where was this discussed? I remember we had a discussion in some call to not open PRs with no issues. And this one is an API change.
> 
> > Uses `DevfileComponent` for `odo describe` when a devfile component is present and `Component` for s2i
> 
> Since Devfile based components are the future for odo and s2i components might get deprecated eventually, shouldn't we use special naming for s2i components instead of devfile components? Meaning, IMO, it would be more helpful to rename s2i components as `s2iComponent` and keep Devfile component as simply `Component`.

Guess my points are invalidated by https://github.com/openshift/odo/pull/3964#issuecomment-695767848. Still don't prefer using `DevfileComponent` for what's going to be the default component with v2 but that's what the user wants, so...

/lgtm

girishramnani:
this was the reason https://github.com/openshift/odo/pull/3964#issuecomment-695081259

girishramnani:
tomas not around so I any one from the approvers like @mik-dass or @dharmit

mik-dass:
`odo describe -o json` doesn't show devfile component
```
[mrinaldas@localhost project]$ odo describe  -o json
{
        "kind": "Component",
        "metadata": {
                "name": "nodejs",
                "namespace": "default",
                "creationTimestamp": null
        },
        "spec": {
                "app": "app",
                "type": "nodejs",
                "urls": {
                        "kind": "List",
                        "metadata": {},
                        "items": [
                                {
                                        "kind": "url",
                                        "metadata": {
                                                "name": "example",
                                                "creationTimestamp": null
                                        },
                                        "spec": {
                                                "host": "example.com",
                                                "port": 3000,
                                                "secure": false,
                                                "kind": "ingress",
                                                "path": "/"
                                        },
                                        "status": {
                                                "state": "Pushed"
                                        }
                                }
                        ]
                },
                "storages": {
                        "kind": "List",
                        "metadata": {},
                        "items": null
                },
                "env": [
                        {
                                "name": "PROJECTS_ROOT",
                                "value": "/project"
                        },
                        {
                                "name": "DEBUG_PORT",
                                "value": "5858"
                        }
                ],
                "ports": [
                        "5858"
                ]
        },
        "status": {
                "state": "Pushed"
        }
}
```
/hold

pkg/component/component_full_description.go
```diff
@@ -125,8 +122,10 @@ func NewComponentFullDescriptionFromClientAndLocalConfig(client *occlient.Client
 	var err error
 	if envInfo != nil {
 		componentDesc, devfile, err = GetComponentFromDevfile(envInfo)
+		cfd.Kind = KindDevfileComponent
```
mik-dass, line 125: `copyFromComponentDesc` below overrides this assignment

girishramnani:
seems like, I need to refactor the code before to achieve this, hence closing it for now

Issue #3892, opened 2020-09-03
Specifying a name does not matter when using odo delete


cdrage:
/kind bug

<!--

Welcome! - We kindly ask you to:

  1. Fill out the issue template below 
  2. Use the Google group if you have a question rather than a bug or feature request.

The group is at: https://groups.google.com/forum/#!forum/odo-users

Thanks for understanding, and for contributing to the project!

-->

## What versions of software are you using?

**Operating System:** Linux

**Output of `odo version`:** master :+1: 

## How did you run odo exactly?


```sh
~/openshift/foobar                                                                                                                                                                                                                                                                                                                                                         ⍉
▶ odo list
Experimental mode is enabled, use at your own risk
Devfile Components: 
APP     NAME                PROJECT     TYPE                STATE
app     java-springboot     foo         java-springboot     Pushed
app     nodejs              foo         nodejs              Pushed
~/openshift/foobar                                                                                                                                                                                                                                                                                                                                                          
▶ odo delete kalsdjalksdjlakdjlasljkd -f
Gathering information for component nodejs
 ✓  Checking status for component [47ms]
Deleting component nodejs
 ✓  Deleting Kubernetes resources for component [46ms]
 ✓  Successfully deleted component
~/openshift/foobar   
```


## Actual behavior

It does not matter what you specify to `odo delete` it will delete whatever is in your current folder...


## Expected behavior

To fail when passing in something random such as `odo delete alksjdalksjdklasd -f`


## Any logs, error output, etc?

See above

johnmcollier:
This one's gonna be a little messy to fix 😕 - looking over the delete code, there's lots of assumptions that we have access to both the devfile and envinfo file when deleting a component

maysunfaisal:
@johnmcollier whats wrong with that? if devfile context is always defined by the current dir or `--context` flag

johnmcollier:
@maysunfaisal It's moreso that if we want to be able to delete a devfile component by name, we need to be able delete it without being in component's context folder.

maysunfaisal:
But

1. isnt that the point of having the `--context` flag? so that we can delete from anywhere using this flag
2. all odo commands require you to point to the component dir, so why should delete be any different?
3. I just tried your example for s2i, I cant do that from any dir. It requires a component in the current dir or requires the `--context` flag:
```
Maysuns-MacBook-Pro:resources maysun$ odo delete mys2ijava
[...]

PR #3998, opened 2020-09-21
Remove specifying a component when deleting

cdrage:
**What type of PR is this?**

/kind bug

This is a temporary fix by removing the current feature of being able to
pass in a component name to `odo delete`.

For example. `odo delete foobar` does not work. We remove that
functionality so you may only specify `odo delete` in the current
directory.

Does not fix any issues, but please see: https://github.com/openshift/odo/issues/3892

**PR acceptance criteria**:

- [X] Unit test

- [X] Integration test

- [X] Documentation

```sh
odo create nodejs --starter
odo push
odo delete
```

### DIFF
--- a/pkg/odo/cli/component/delete.go
+++ b/pkg/odo/cli/component/delete.go
@@ -29,9 +29,9 @@ import (
 // DeleteRecommendedCommandName is the recommended delete command name
 const DeleteRecommendedCommandName = "delete"
 
-var deleteExample = ktemplates.Examples(`  # Delete component named 'frontend'. 
-%[1]s frontend
-%[1]s frontend --all
+var deleteExample = ktemplates.Examples(`  # Delete the component in your current directory
+%[1]s
+%[1]s --all
   `)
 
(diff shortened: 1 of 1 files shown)


pkg/odo/cli/component/delete.go
```diff
@@ -295,11 +295,11 @@ func NewCmdDelete(name, fullName string) *cobra.Command {
 	do := NewDeleteOptions()
 
 	var componentDeleteCmd = &cobra.Command{
-		Use:         fmt.Sprintf("%s <component_name>", name),
+		Use:         fmt.Sprintf("%s", name),
 		Short:       "Delete component",
 		Long:        "Delete component.",
 		Example:     fmt.Sprintf(deleteExample, fullName),
-		Args:        cobra.MaximumNArgs(1),
+		Args:        cobra.MaximumNArgs(0),
```
johnmcollier, line 302: Should we still allow a name to be passed in when `--s2i` is used?

cdrage:
Closing for now. We should check this post-release as to not break any compatibility with current s2i components.

PR #4000, opened 2020-09-21
Update integration testcases and e2e testcases for ppc64le support

sarveshtamba:
**What type of PR is this?**
/kind feature

Update integration testcases and e2e testcases for ppc64le support after testing on Power system.

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/e2escenarios/e2e_source_test.go
+++ b/tests/e2escenarios/e2e_source_test.go
@@ -86,9 +86,9 @@ var _ = Describe("odo source e2e tests", func() {
 	Context("odo component creation", func() {
 
 		It("Should be able to deploy a python source application", func() {
-                        if runtime.GOARCH == "ppc64le" {
-                                Skip("Skipping test on Power because python is not fully supported by odo, and it is not guaranteed to work.")
-                        }
+			if runtime.GOARCH == "ppc64le" {
+				Skip("Skipping test on Power because python is not fully supported by odo, and it is not guaranteed to work.")
+			}
 			helper.CopyExample(filepath.Join("source", "python"), context)
 			helper.CmdShouldPass("odo", "create", "--s2i", "python", "python-app", "--project",
 				project, "--context", context)
--- a/tests/e2escenarios/e2e_source_test.go
+++ b/tests/e2escenarios/e2e_source_test.go
@@ -32,7 +32,7 @@ var _ = Describe("odo source e2e tests", func() {
 
 	Context("odo component creation from source", func() {
 		JustBeforeEach(func() {
-			if runtime.GOARCH == "s390x" {
+			if runtime.GOARCH == "s390x" || runtime.GOARCH == "ppc64le" {
 				Skip("Skipping test because there is no supported builder image.")
 			}
 		})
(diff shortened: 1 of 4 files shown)


tests/e2escenarios/e2e_source_test.go
```diff
@@ -86,6 +86,9 @@ var _ = Describe("odo source e2e tests", func() {
 	Context("odo component creation", func() {
 
 		It("Should be able to deploy a python source application", func() {
+			if runtime.GOARCH == "ppc64le" {
```
amitkrout, line 89: This is a pre test condition. Please add it in ```JustBeforeEach(func() {})``` of the same ```Context("", func() {})```

amitkrout (changes requested):

tests/e2escenarios/e2e_source_test.go
sarveshtamba: @amitkrout There are 2 other tests in this context. Adding the condition in `JustBeforeEach(func() {})` of the same `Context("", func() {})` would skip the other 2 tests as well. Hence added this check only to this one test which is to be skipped.
I have followed existing similar checks as below and few more:-
https://github.com/openshift/odo/blob/master/tests/integration/plugin_handler_test.go#L46
https://github.com/openshift/odo/blob/master/tests/integration/devfile/cmd_devfile_push_test.go#L965
https://github.com/openshift/odo/blob/master/tests/integration/operatorhub/cmd_service_test.go#L416

tests/e2escenarios/e2e_source_test.go
```diff
@@ -32,7 +32,7 @@ var _ = Describe("odo source e2e tests", func() {
 
 	Context("odo component creation from source", func() {
 		JustBeforeEach(func() {
-			if runtime.GOARCH == "s390x" {
+			if runtime.GOARCH == "s390x" || runtime.GOARCH == "ppc64le" {
 				Skip("Skipping test because there is no supported builder image.")
```
prietyc123, line 36: I would say just change the `Skip` comment as `Skipping test specs because there is no supported builder image or component is not fully supported by odo` and add the test spec https://github.com/openshift/odo/pull/4000/files#diff-e223a048840cec18f0eb80f10c780334R88 within this context itself.

prietyc123 (changes requested):

tests/e2escenarios/e2e_source_test.go
prietyc123, line 89: Lets have only one Context for Skip. Could you please change the tests according to https://github.com/openshift/odo/pull/4000#discussion_r494016243

prietyc123 (changes requested):

prietyc123:
@sarveshtamba Please address the requested change until then I am holding this pr

/hold

tests/e2escenarios/e2e_source_test.go
sarveshtamba: The skip comment is existing in the code since earlier and I have just added the additional check for `ppc64le`.

sarveshtamba: Moving the python spec from the current Context and adding that test to the Context at line number 33 would make python test to be skipped for s390x too which is not desired. Hence the changes as in original PR.

sarveshtamba (commented):
Provided respective comments explaining the changes.

prietyc123:
As discussed on slack, it seems this pr is blocker for the system P-Z team and it should go in before 2.0 release freeze. So I will do the clean up part later.

/unhold
/lgtm

Issue #3886, opened 2020-09-02
odo allows devfile commands with duplicate ids


maysunfaisal:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**

## How did you run odo exactly?
odo push

## Actual behavior
odo push proceeds

## Expected behavior
odo push should error out on command validation

## Any logs, error output, etc?

this is my devfile.yaml example:
```
components:
  - container:
      name: runtime
      image: registry.access.redhat.com/ubi8/nodejs-12:1-45
      memoryLimit: 1024Mi
      mountSources: true
      sourceMapping: /project
      endpoints:
        - name: http-3000
          targetPort: 3000
  - container:
      name: runtime2
      image: registry.access.redhat.com/ubi8/nodejs-12:1-45
      memoryLimit: 1024Mi
      mountSources: true
      sourceMapping: /project
      command: ['tail']
      args: [ '-f', '/dev/null']
commands:
  - exec:
      id: install
      component: runtime
      commandLine: npm install
      workingDir: /project
      group:
        kind: build
        isDefault: false
  - exec:
      id: install
      component: runtime2
      commandLine: npm install
      workingDir: /project
      group:
        kind: build
        isDefault: false
  - exec:
      id: run
      component: runtime
      commandLine: npm start
      workingDir: /project
      group:
        kind: run
        isDefault: false
  - exec:
      id: run2
      component: runtime
      commandLine: npm start
      workingDir: /project
      group:
        kind: run
        isDefault: false
```

validation does not catch this:

1. duplicate commands `install` reference different components
2. duplicate commands `install` belong to `build` and none are default 

/area devfile

johnmcollier:
I can do this

/assign

johnmcollier:
Blocked til https://github.com/openshift/odo/pull/3908 is merged

PR #4001, opened 2020-09-21
Don't allow duplicate command IDs

johnmcollier:
Signed-off-by: John Collier <jcollier@redhat.com>

**What type of PR is this?**
/kind bug

This PR updates the `GetCommands()` function to return an error if duplicate command IDs are found in a devfile, allowing the devfile validation to catch this and show a proper error to the user. Because a number of places also call `GetCommands()`, I had to add err checks to those places as well (as a side note, we really should look at refactoring our calls to that function, we're calling it way too much).

Fixes https://github.com/openshift/odo/issues/3886

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Sample devfile:
```
schemaVersion: 2.0.0
metadata:
  name: nodejs
  version: 1.0.0
starterProjects:
- name: nodejs-starter
  git:
    remotes:
components:
- name: runtime
  container:
    endpoints:
    - name: http-3000
      targetPort: 3000
    - exposure: public
      path: /
      name: nodejs-3000
      targetPort: 3000
      protocol: http
    memoryLimit: 1024Mi
    mountSources: true
    sourceMapping: /project
commands:
- exec:
    commandLine: npm install
    component: runtime
    group:
      isDefault: true
      kind: build
    workingDir: /project
  id: install
- exec:
    commandLine: npm install
    component: runtime
    group:
      isDefault: false
      kind: build
    workingDir: /project
  id: install
- exec:
    commandLine: npm start
    component: runtime
    group:
      isDefault: true
      kind: run
    workingDir: /project
  id: run
```

`odo create` / `odo push` should now show the following error:
```
 ✗  devfile has duplicate command IDs "install"
```


pkg/devfile/parser/data/2.0.0/components.go
```diff
@@ -131,24 +131,33 @@ func (d *Devfile200) UpdateComponent(component common.DevfileComponent) {
...
 	commands := make(map[string]common.DevfileCommand, len(d.Commands))
 
 	for _, command := range d.Commands {
 		// we convert devfile command id to lowercase so that we can handle
 		// cases efficiently without being error prone
 		// we also convert the odo push commands from build-command and run-command flags
-		commands[command.SetIDToLower()] = command
+		commandID := command.SetIDToLower()
+		if _, exists := commands[commandID]; exists {
+			return nil, fmt.Errorf("devfile has duplicate command IDs %q", commandID)
```
maysunfaisal, line 145: Interesting that you decided to do it here. Hmm, I was expecting it to be done in `validate` package and that wouldnt result in func signature changes 🤔

johnmcollier: Good question. It's because the most direct place to check for duplicate commands is when we're constructing the commands map.

If we do it elsewhere (like `validateCommands` https://github.com/openshift/odo/blob/master/pkg/devfile/validate/commands.go), the commands map has already been constructed and we won't know from that which commands (if any) are duplicated. We'd have to do a manual comparison between the list of all devfile commands and the commands Map.

pkg/devfile/adapters/common/generic.go
```diff
@@ -64,7 +64,11 @@ func (a GenericAdapter) ExecuteCommand(compInfo ComponentInfo, command []string,
 
 // ExecuteDevfileCommand executes the devfile init, build and test command actions synchronously
 func (a GenericAdapter) ExecuteDevfileCommand(command common.DevfileCommand, show bool) error {
-	c, err := New(command, a.Devfile.Data.GetCommands(), a)
+	commands, err := a.Devfile.Data.GetCommands()
```
maysunfaisal, line 67: can we rename the var to `commandsMap`/`devfileCommandsMap`? This file's command names are all over the place

pkg/devfile/adapters/common/generic.go
```diff
@@ -95,7 +99,10 @@ func convertGroupKindToString(exec *common.Exec) string {
 // Init only runs once when the component is created.
 func (a GenericAdapter) ExecDevfile(commandsMap PushCommandsMap, componentExists bool, params PushParameters) (err error) {
```
maysunfaisal, line 100: this `commandsMap` is different and we've to keep track, can you help rename it to something else, may prefix with push?

pkg/devfile/adapters/common/generic.go
maysunfaisal: Maybe `commandsMap` since all other files are calling it that

pkg/devfile/adapters/common/utils_test.go
```diff
@@ -431,7 +431,10 @@ func TestGetCommandsForGroup(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			commands := getCommandsByGroup(devObj.Data, tt.groupType)
+			commands, err := getCommandsByGroup(devObj.Data, tt.groupType)
```
maysunfaisal, line 434: Hmm do you want to maybe test the func now with some error test cases?

pkg/devfile/adapters/common/utils_test.go
```diff
@@ -533,7 +536,11 @@ func TestGetCommands(t *testing.T) {
 				},
 			}
 
-			commandsMap := devObj.Data.GetCommands()
+			commandsMap, err := devObj.Data.GetCommands()
```
maysunfaisal, line 539: not sure what this test is testing, this looks like redundant if it is already being tested at the parser level but now i feel we should test one err condition if this is going to stay here

pkg/devfile/parser/types.go
```diff
@@ -65,7 +65,12 @@ func (d DevfileObj) OverrideComponents(overridePatch []common.DevfileComponent)
 func (d DevfileObj) OverrideCommands(overridePatch []common.DevfileCommand) error {
 	for _, patchCommand := range overridePatch {
 		found := false
-		for _, originalCommand := range d.Data.GetCommands() {
+		commands, err := d.Data.GetCommands()
```
maysunfaisal, line 68: commandsMap perhaps

maysunfaisal (commented):
Looks good and works well. Just some comments on var name cleanup and adding a couple of test cases

pkg/devfile/parser/data/2.0.0/components.go
johnmcollier: Ended up finding an approach to do the validation in the validate package instead.

johnmcollier:
@maysunfaisal I ended up refactoring the validation somewhat, think this is a better / cleaner approach to what I initially suggested. Mind re-reviewing? Thanks!

pkg/devfile/validate/commands.go
```diff
@@ -1,16 +1,26 @@
...
 import (
+	"fmt"
 	"strings"
 
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
 )
 
-// validateCommands validates all the devfile commands
-func validateCommands(commands map[string]common.DevfileCommand, components []common.DevfileComponent) (err error) {
+// validateCommands validates all the devfile commands. If there are commands with duplicate IDs, an erroor is returned
```
mik-dass, line 10: error

johnmcollier: @mik-dass Fixed, thanks!

maysunfaisal (approved):
Thx for the changes, much smaller
/lgtm

mik-dass (approved):
/approve

PR #4029, opened 2020-09-23
make odo list's output for no components consistent

girishramnani:
**What type of PR is this?**

/kind bug

`odo list` returns an error when there are no components but `odo list -o json` returns empty lists. Both approach are subjective but we need to make things consistent. Hence we dont return an error now if there are no components

No issue

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [X] Documentation 

Everything should work as expected


mik-dass (approved):
/lgtm

Issue #2877, opened 2020-04-14
component push throws error of waited 4m0s but couldn't find running pod matching selector


prietyc123:
/kind flake

## What versions of software are you using?

**Operating System:**
All Supported

**Output of `odo version`:**
master

## How did you run odo exactly?
```odo push --context context``` on OpenShift ci.

## Actual behavior
Throwing error as: 
```
[odo] Please use `odo push` command to create the component with source deployed
Running odo with args [odo push --context /tmp/947847413]
[odo] Validation
[odo]  •  Checking component  ...
[odo] 
 ✓  Checking component [60ms]
[odo] 
[odo] Configuration changes
[odo]  ✓  Initializing component
[odo]  •  Creating component  ...
[odo] 
 ✓  Creating component [200ms]
[odo] 
[odo] Pushing to component dotnet-app of type local
[odo]  •  Checking files for pushing  ...
[odo] 
 ✓  Checking files for pushing [553410ns]
[odo]  •  Waiting for component to start  ...
[odo]  ✗  Waiting for component to start [4m]
[odo]  ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=dotnet-app-app'
Deleting project: bkbwthrcvq
Running odo with args [odo project delete bkbwthrcvq -f]
```

## Expected behavior
It should push the component successfully into the deployment.

## Any logs, error output, etc?

```
[odo] Please use `odo push` command to create the component with source deployed
Running odo with args [odo push --context /tmp/947847413]
[odo] Validation
[odo]  •  Checking component  ...
[odo] 
 ✓  Checking component [60ms]
[odo] 
[odo] Configuration changes
[odo]  ✓  Initializing component
[odo]  •  Creating component  ...
[odo] 
 ✓  Creating component [200ms]
[odo] 
[odo] Pushing to component dotnet-app of type local
[odo]  •  Checking files for pushing  ...
[odo] 
 ✓  Checking files for pushing [553410ns]
[odo]  •  Waiting for component to start  ...
[odo]  ✗  Waiting for component to start [4m]
[odo]  ✗  waited 4m0s but couldn't find running pod matching selector: 'deploymentconfig=dotnet-app-app'
Deleting project: bkbwthrcvq
Running odo with args [odo project delete bkbwthrcvq -f]
```
[...]

PR #4063, opened 2020-10-01
Setting up baseline POC of internal CI

### LINKED ISSUE #4056: Run PR tests on internal PSI resources
We are currently battling time lost in CI due to cluster provisioning / flakyness. This is happening while we have unused resources in PSI.

I propose we start using a combination of Public Ci and internal resources to run tests, using `Message Queues` as the mediator. 
Here is a mind map based on the plan and some research

image

mohammedzee1000:
fixes #4056 
CI firewall code is in https://github.com/mohammedzee1000/ci-firewall


mohammedzee1000:
Do not worry about so many retests, I am just perfecting the system :)

mohammedzee1000:
Please take a look before I move forward

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -5,58 +5,20 @@ set -e
 # show commands
 set -x
 
-ARCH=$(uname -m)
-export CI="openshift"
-if [ "${ARCH}" == "s390x" ]; then
-    make configure-installer-tests-cluster-s390x
-elif [ "${ARCH}" == "ppc64le" ]; then
-    make configure-installer-tests-cluster-ppc64le
+export AMQP_URI="amqp://mquser:password123@35.224.22.100/"
```
girishramnani, line 8: password directly exposed in an open repo?

mohammedzee1000: At the moment. That is where secrets come into play

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+  set -x
+}
+
+set -ex
+
+shout "Setting up"
+
+mkdir bin
+GOBIN="`pwd`/bin"
+KUBECONFIG="`pwd`/config"
```
girishramnani, line 15: can we check if kubeconfig exists or not? and error out if it isnt?

mohammedzee1000: No, the idea is to provide isolation. This would be the `global kubeconfig` for the test

mohammedzee1000: Using system level kubeconfig could conflict when multiple runs set different contexts

mohammedzee1000:
Note: unit and e2e are passing now, running for around 13 mins. The explicit `exit 1` is so you can see the logs coming through.
We can remove that before merging

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,61 +2,31 @@
...
-    make test-operator-hub
-    # E2e tests
-    make test-e2e-all
+    curl -kLO https://github.com/mohammedzee1000/ci-firewall/releases/download/${CI_FIREWALL_VERSION}/ci-firewall-linux-amd64.tar.gz
+    tar -xzf ci-firewall-linux-amd64.tar.gz
 fi
+./ci-firewall request --runscript scripts/run_all_tests.sh --timeout 45m
 
-odo logout
+exit 1
```
mohammedzee1000, line 32: This `exit 1` is to ensure logs are visible even with a successful run and needs to be removed before merging

mohammedzee1000:
Blocked on https://github.com/openshift/odo/issues/4278

dharmit:
https://github.com/CentOS/container-index/pull/412 is merged. Are we good to move further on this?

mohammedzee1000:
Waiting on container to become available on r.c.o. @bamachrn what would be the e.t.a for this?

mohammedzee1000:
nodejs-12 is now available in r.c.o. Unblocking

prietyc123:
Failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4063/pull-ci-openshift-odo-master-v4.6-integration-e2e/1336197397600014336#1:build-log.txt%3A466 falke and tracked in https://github.com/openshift/odo/issues/2877 . 

/retest

pkg/util/util.go
```diff
@@ -1432,3 +1432,11 @@ func GetCommandStringFromEnvs(envVars []v1alpha2.EnvVar) string {
 	}
 	return setEnvVariable
 }
+
+func GetEnvWithDefault(key string, defaultval string) string {
```
dharmit, line 1436: Can you add some comments to this function? Also, it's being used in tests script, IIUC. Why not put it in `helper` package under `tests`?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
 
 # fail if some commands fails
 set -e
+# Hide command that deals with secrets
+set +x
+if [[ -f $ODO_RABBITMQ_AMQP_URL ]]; then
+    export AMQP_URI=$(cat $ODO_RABBITMQ_AMQP_URL)
+fi
+if [[ -z $AMQP_URI ]]; then
```
dharmit, line 10: Shouldn't this be set as `else` instead? Something like:
```sh
if [[ -f $ODO_RABBITMQ_AMQP_URL ]]; then
    export AMQP_URI=$(cat $ODO_RABBITMQ_AMQP_URL)
else
    echo "Please set AMQP_URI env with amqp uri or provide path of file containing it as ODO_RABBITMQ_AMQP_URL env"
    exit 1
fi
```

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
...
+    echo "Please set AMQP_URI env with amqp uri or provide path of file containing it as ODO_RABBITMQ_AMQP_URL env"
+    exit 1
+fi
+export JOB_NAME="odo-pr-build"
+export REPO_URL="https://github.com/openshift/odo"
+export TARGET="$(jq .refs.pulls[0].number <<< $(echo $JOB_SPEC))"
+export CI_FIREWALL_VERSION="valpha"
```
dharmit, line 17: Would it make more sense to initialize these variables at the beginning of the file? Asking because that's how I have usually seen it done in shell scripts. 

Also, just for the sake of my understanding, do we really need them after the execution of this script is over? Asking because they're `export`ed.

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
...
+if [[ -f $ODO_RABBITMQ_AMQP_URL ]]; then
+    export AMQP_URI=$(cat $ODO_RABBITMQ_AMQP_URL)
+fi
+if [[ -z $AMQP_URI ]]; then
+    echo "Please set AMQP_URI env with amqp uri or provide path of file containing it as ODO_RABBITMQ_AMQP_URL env"
+    exit 1
+fi
+export JOB_NAME="odo-pr-build"
+export REPO_URL="https://github.com/openshift/odo"
+export TARGET="$(jq .refs.pulls[0].number <<< $(echo $JOB_SPEC))"
```
dharmit, line 16: Can you please add a comment here to explain what this is doing? I think it's fetching the PR number and storing it in variable `TARGET`. Also, can I request for a better variable name than `TARGET`, maybe?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
...
+export CI_FIREWALL_VERSION="valpha"
+echo "Initiating"
```
dharmit, line 18: Initiating what?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
...
 # show commands
 set -x
-
-ARCH=$(uname -m)
```
dharmit, line 8: Are these changes going to impact multi-arch team in any way? Have we communicated this to them?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,63 +2,31 @@
...
-    # E2e tests
-    make test-e2e-all
-elif  [ "${ARCH}" == "ppc64le" ]; then
-    # Integration tests
-    make test-integration
-    make test-cmd-login-logout
-    make test-cmd-project
-    # E2e tests
-    make test-e2e-all
+if [[ $CI_FIREWALL_VERSION == "master" ]]; then
```
dharmit, line 21: Just 3 lines before this check, we are setting `CI_FIRWALL_VERSION` to `valpha`. When is this condition going to satisfy when it's value is `master` instead?

scripts/openshiftci-presubmit-all-tests.sh
dharmit: Can you please address this TODO?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+  set +x
+  echo -e "\n!!!!!!!!!!!!!!!!!!!!\n${1}\n!!!!!!!!!!!!!!!!!!!!\n"
+  set -x
+}
+
+set -ex
+
+shout "Setting up"
+
+mkdir bin
```
dharmit, line 13: IIUC, this will create a `bin` directory under `cif` directory where we're cloning ci-firewall code, right? Same applies to next few steps where we're using `pwd`. This is by design, right?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+GOBIN="`pwd`/bin"
+KUBECONFIG="`pwd`/config"
+SCRIPT_IDENTITY=${SCRIPT_IDENTITY:-"def-id"}
```
dharmit, line 16: What's this step doing? Can we add a comment that would help us or someone who might look at this code few months later?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+export SKIP_USER_LOGIN_TESTS="true"
+export GINKGO_TEST_ARGS="--noColor"
```
dharmit, line 18: Colour is cool, no? Makes it easy to read logs, no? Why no colour? :sob: 

Seriously, why no colour? :thinking:

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+
+shout "Getting oc binary"
+if [[ $BASE_OS == "linux"  ]]; then
```
dharmit, line 21: Where/How is this variable being set to `linux` or anything else?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+    set +x
+    curl --connect-timeout 150 --max-time 2048 -k ${OCP4X_DOWNLOAD_URL}/${ARCH}/${BASE_OS}/oc.tar -o ./oc.tar
```
dharmit, line 23: Where are we setting the link to oc binary?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+PATH=$PATH:$GOBIN
+
+#-----------------------------------------------------------------------------
+
+shout "Testing"
+
+# Run unit tests
+GOFLAGS='-mod=vendor' make test
```
dharmit, line 49: Shouldn't we `shout` that we're doing unit testing?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# Prep for int
+shout "Building"
```
dharmit, line 52: Building what? :smile: 

Also, the comment above. Did you mean `Prep for init`? That is, `s/int/init/`? But `Prep for init` of what?

scripts/run_all_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+
+shout "Doing some presetup"
+
+for i in $(oc projects -q); do
+    if [[ $i == "${SCRIPT_IDENTITY}"* ]]; then
+        oc delete project $i
+    fi
+done
+
+export REDHAT_OPENJDK11_RHEL8_PROJECT="${SCRIPT_IDENTITY}$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 4 | head -n 1)"
```
dharmit, line 74: Do we really need to have `REDHAT` in the variable names? Can we do away without it please?

tests/e2escenarios/e2e_images_test.go
```diff
@@ -130,23 +131,27 @@ var _ = Describe("odo supported images e2e tests", func() {
 		})
 
 		It("Should be able to verify the openjdk-11-rhel8 image", func() {
-			oc.ImportImageFromRegistry("registry.redhat.io", "openjdk/openjdk-11-rhel8:latest", "java:8", "openjdk-11-rhel8")
-			verifySupportedImage("openjdk/openjdk-11-rhel8:latest", "openjdk", "java:8", "openjdk-11-rhel8", appName, commonVar.Context)
+			redhatOpenjdk12RHEL8Project := util.GetEnvWithDefault("REDHAT_OPENJDK11_RHEL8_PROJECT", "openjdk-11-rhel8")
+			oc.ImportImageFromRegistry("registry.redhat.io", "openjdk/openjdk-11-rhel8:latest", "java:8", redhatOpenjdk12RHEL8Project)
+			verifySupportedImage("openjdk/openjdk-11-rhel8:latest", "openjdk", "java:8", redhatOpenjdk12RHEL8Project, appName, commonVar.Context)
```
dharmit, line 136: Can we please do away with using `redhat` and `RHEL` in variable names? At other places in the file as well.

tests/helper/helper_oc.go
```diff
@@ -593,3 +593,31 @@ func (oc OcRunner) GetAllPVCNames(namespace string) []string {
 func (oc OcRunner) DeletePod(podName string, namespace string) {
 	CmdShouldPass(oc.path, "delete", "pod", "--namespace", namespace, podName)
 }
+
+func (oc OcRunner) GetAllPodsInNs(namespace string) string {
+	cmd := []string{"get", "pods", "-n", namespace}
```
dharmit, line 598: This is more like args and not cmd. Since we're passing these as arguments to the oc command a couple of lines later. We're calling it as `ocArgs` at numerous places in the tests.

tests/helper/helper_oc.go
```diff
@@ -593,3 +593,31 @@ func (oc OcRunner) GetAllPVCNames(namespace string) []string {
...
+	noResourcesMsg := fmt.Sprintf("No resources found in %s namespace", namespace)
+	WaitForCmdOut(oc.path, cmd, 1, true, func(output string) bool {
+		return !strings.Contains(output, noResourcesMsg)
+	}, true)
+	return CmdShouldPass(oc.path, cmd...)
+}
+
+func (oc OcRunner) StatFileInPod(cmpName, appName, project, filepath string) string {
```
dharmit, line 606: Some comments explaining what this function is doing?

dharmit (changes requested):

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: Updating

scripts/run_all_tests.sh
mohammedzee1000: Because jenkins craps itself with colored output.

scripts/run_all_tests.sh
mohammedzee1000: This value is injected on jenkins and is expected to be set already

scripts/run_all_tests.sh
dharmit: Haha. Okay.

scripts/run_all_tests.sh
dharmit: Ack, thanks.

dharmit:
From the failure logs:
```
 + ./ci-firewall request --runscript scripts/run_all_tests.sh --timeout 1h45m
2020/12/17 10:39:40 : timed out
```
Wow! Our tests take over 105 minutes and still don't finish? :scream:

mohammedzee1000:
> From the failure logs:
> 
> ```
>  + ./ci-firewall request --runscript scripts/run_all_tests.sh --timeout 1h45m
> 2020/12/17 10:39:40 : timed out
> ```
> 
> Wow! Our tests take over 105 minutes and still don't finish?

Well they do take a while

pkg/util/util.go
mohammedzee1000: Updated

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: Hmm makes sense

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: Export exports env to the current shell (which is basically the one running this script) and all child processes. 

If I don't export it, I will have to set it with every process that needs the env like 
```
ENV1=someval command1;
ENV1=someval command2;
```

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: yep, that is what it is doing. Adding comment.
I think target makes sense here as it couold mean PR no, branch name or tag to `ci-firewall`

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: /cc @sarveshtamba

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: Removed

scripts/run_all_tests.sh
mohammedzee1000: This has nothing to do with cloning of ci-firewall. This is the script that runs jenkins side.

It will create a directory called `bin` in whatever is the directory when the script runs. On jenkins this will be the workspace. 
I will add more comments to make this clear

scripts/run_all_tests.sh
mohammedzee1000: Comments added

scripts/run_all_tests.sh
mohammedzee1000: This is done on jenkins side using `--env` parameter of `ci-firewall work` which is run there. So as an example, this could end up looking like
`example.com/ppcle/linux/oc.tar`

scripts/run_all_tests.sh
mohammedzee1000: Updated

scripts/run_all_tests.sh
mohammedzee1000: I mean integration/e2e tests. Updating comment

tests/e2escenarios/e2e_images_test.go
mohammedzee1000: I see little point. The name is already there in code `registry.redhat.io` for example. I think as much as possible, envs names should make clear, the intent of use.
cc @girishramnani @kadel wdyt

tests/helper/helper_oc.go
mohammedzee1000: Updated

tests/helper/helper_oc.go
mohammedzee1000: Comments added

scripts/openshiftci-presubmit-all-tests.sh
sarveshtamba: These changes may impact P/Z. CC:- @zhengxiaomei123

mohammedzee1000: @sarveshtamba @zhengxiaomei123  letting you know so you are aware.
The idea here is to do the actual testing against internal resources like PSI. But exporting the logs here in a manner of speaking
It might be a good idea if we can hash this on a call. Let me know on slack

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,69 +2,30 @@
...
-if [ "${ARCH}" == "s390x" ]; then
-    make configure-installer-tests-cluster-s390x
-elif [ "${ARCH}" == "ppc64le" ]; then
-    make configure-installer-tests-cluster-ppc64le
-else
-    make configure-installer-tests-cluster
+# Hide command that deals with secrets
+set +x
+if [[ -f $ODO_RABBITMQ_AMQP_URL ]]; then
+    export AMQP_URI=$(cat $ODO_RABBITMQ_AMQP_URL)
```
anandrkskd, line 8: Does AMQP_URI contains the url to the instance of rabbitMQ? and why do we need to export AMQP_URI are we using this Env variable outside  `scripts/openshiftci-presubmit-all-tests.sh` ?

scripts/run_all_tests.sh
anandrkskd: Is this for latter retrive of artifacts, why are we not using the default gobin ?

mohammedzee1000:
It appears we are in a strange situation. The last error, as seen [here](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4063/pull-ci-openshift-odo-master-v4.6-integration-e2e/1348953368604708864) is actually related to https://github.com/openshift/odo/issues/4347. 

As things stand:
 1. Latest tag for nodejs-10 and nodejs-12 are broken on docker. Hence why we have temporary revert to a specific tag in master
 2. This specific tag is unavailable in r.c.o (rco only serves latest)
 3. It appears quay might be to only option after all.

Hence until the above issue is solved, this PR stands blocked
/status blocked

mohammedzee1000:
cc @prietyc123

dharmit:
> You must be a member of the [openshift/odo-maintainers](https://github.com/orgs/openshift/teams/odo-maintainers/members) GitHub team to add status labels. If you believe you should be able to issue the /status command, please contact your odo maintainers and have them propose you as an additional delegate for this responsibility.

What's this team that bot is asking to be a member of? The link throws a 404. Any clue @mohammedzee1000 @kadel ?

dharmit:
> The last error, as seen [here](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4063/pull-ci-openshift-odo-master-v4.6-integration-e2e/1348953368604708864) is actually related to #4347.

If you'd rebase on current master, wouldn't this be fixed since #4348 has already been merged? :thinking:

pkg/util/util.go
```diff
@@ -1436,3 +1436,12 @@ func GetCommandStringFromEnvs(envVars []v1alpha2.EnvVar) string {
 	}
 	return setEnvVariable
 }
+
+//GetEnvWithDefault gets value of spcified env if it is set, otherwise, it returns default value
```
prietyc123, line 1440: Typo `specified`

prietyc123 (changes requested):

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,69 +2,30 @@
...
-elif  [ "${ARCH}" == "ppc64le" ]; then
-    # Integration tests
-    make test-integration
-    make test-cmd-login-logout
-    make test-cmd-project
-    # E2e tests
-    make test-e2e-all
-else
-    # Integration tests
-    make test-integration || error=true
```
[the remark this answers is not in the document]
prietyc123: I am also worried, we won't be having such a single script to run the entire test target in one go locally. @mohammedzee1000 I think we can have a copy of it in another file, to avoid impacting P/Z, interop and running tests locally. WDYT?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,69 +2,30 @@
...
-    # Fail the build if there is any error while test execution
-    if [ $error ]; then 
-        exit -1
-    fi
-fi
+curl -kLO https://github.com/mohammedzee1000/ci-firewall/releases/download/valpha/ci-firewall-linux-amd64.tar.gz
+tar -xzf ci-firewall-linux-amd64.tar.gz
 
-cp -r reports $ARTIFACTS_DIR 
+./ci-firewall request --runscript scripts/run_all_tests.sh --timeout 2h15m
```
prietyc123, line 30: Timeout has been set on the basis of what?

scripts/run_all_tests.sh
prietyc123: > Can we do away without it please?

+1

prietyc123 (changes requested):

scripts/openshiftci-presubmit-all-tests.sh
mohammedzee1000: Tests can take an hour as far as I can see. So I am using a safe timeout. We will reduce it as we move forward.

scripts/openshiftci-presubmit-all-tests.sh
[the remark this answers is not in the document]
mohammedzee1000: The reason I am using openshift-presubmit-all-tests is that it is the entry point as defined on prow side. I can create a copy of the original script as-is and rename it to something else though if that is ok

pkg/util/util.go
mohammedzee1000: fixed

dharmit (commented):
/approve

scripts/run_all_tests.sh
mohammedzee1000: Mostly to prevent overlap between multiple scripts. East test instance gets its own GOBIN

mohammedzee1000:
There is ongoing PSI outage. Preventing reruns
/hold

prietyc123:
> There is ongoing PSI outage. Preventing reruns

@mohammedzee1000 How frequently we face this kind of outage?

mohammedzee1000:
PSI outage done, we have a new cluster and PR rebased
/hold cancel

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,69 +2,30 @@
...
-cp $KUBECONFIG $TMP_DIR/kubeconfig
-chmod 640 $TMP_DIR/kubeconfig
-export KUBECONFIG=$TMP_DIR/kubeconfig
 
-# Login as developer
-odo login -u developer -p password@123
+##### These are varialbes used by ci-firewall as one of the ways to get its parameters
+# If AMQP_URI is not set by the time we reach here, show error message and exit.
+export AMQP_URI=${AMQP_URI:?"Please set AMQP_URI env with amqp uri or provide path of file containing it as ODO_RABBITMQ_AMQP_URL env"}
+export JOB_NAME="odo-pr-build"
```
prietyc123, line 14: `odo-pr-build` seems so generic, isn't it? How do we differentiate minikube and openshift pr build. Can we say it as `odo-ocp-pr-build`?

prietyc123 (changes requested):

prietyc123:
> Running git with args [git clone https://github.com/openshift/nodejs-ex /tmp/003526642/nodejs-ex]
[ssh:Fedora 32] [git] Cloning into '/tmp/003526642/nodejs-ex'...
[ssh:Fedora 32] [git] remote: Not Found
[ssh:Fedora 32] [git] fatal: repository 'https://github.com/openshift/nodejs-ex/' not found

I don't find `https://github.com/openshift/nodejs-ex/` anymore. We need to replace it with `https://github.com/sclorg/nodejs-ex` I guess.

dharmit:
> I don't find `https://github.com/openshift/nodejs-ex/` anymore. We need to replace it with `https://github.com/sclorg/nodejs-ex` I guess.

It redirects automatically for me both in the browser as well as `git clone`:
```sh
$ git clone https://github.com/openshift/nodejs-ex /tmp/nodejs-ex
Cloning into '/tmp/nodejs-ex'...
remote: Enumerating objects: 660, done.
remote: Total 660 (delta 0), reused 0 (delta 0), pack-reused 660
Receiving objects: 100% (660/660), 261.48 KiB | 481.00 KiB/s, done.
Resolving deltas: 100% (257/257), done.
```

> I would say we should avoid cloning it externally.

Not sure what this means.

dharmit:
> > I would say we should avoid cloning it externally.
> 
> Not sure what this means.

Looking at the PR #4394, I think you meant to use `CopyExample` instead of doing `git clone` in the tests, right?

prietyc123:
> Looking at the PR #4394, I think you meant to use CopyExample instead of doing git clone in the tests, right?

Yes, I meant using `CopyExample`

mohammedzee1000:
Whoops, I had messed up on jenkins
/retest

anandrkskd:
@mohammedzee1000 can we try with a longer timeout? just to be sure if time out is not actually an issue
As the [log](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4063/pull-ci-openshift-odo-master-v4.6-integration-e2e/1358674634525904896) says it failed due to timeout.

Issue #4018, opened 2020-09-22
Add better message for component create when devfile does not exists


prietyc123:
/kind bug

## What versions of software are you using?

**Operating System:**
All supported

**Output of `odo version`:**
Master

## How did you run odo exactly?

Common steps on 4.5 and 3.11 cluster:
```
$ odo project create wcldizlwqg 
✓  Project 'wcldizlwqg' is ready for use
 ✓  New project created and now using project: wcldizlwqg

$ odo registry list
 ✗  No devfile registries added to the configuration. Refer `odo registry add -h` to add one

$ oc --request-timeout 5m import-image nodejs:latest --namespace=wcldizlwqg --from=docker.io/centos/nodejs-10-centos7:latest --confirm
imagestream.image.openshift.io/nodejs imported

Name:                   nodejs
Namespace:              wcldizlwqg
Created:                Less than a second ago
Labels:                 <none>
Annotations:            openshift.io/image.dockerRepositoryCheck=2020-09-22T15:08:26Z
Image Repository:       image-registry.openshift-image-registry.svc:5000/wcldizlwqg/nodejs
Image Lookup:           local=false
Unique Images:          1
Tags:                   1

latest
  tagged from docker.io/centos/nodejs-10-centos7:latest
[...]

$ oc annotate istag/nodejs:latest --namespace=wcldizlwqg tags=builder --overwrite
imagestreamtag.image.openshift.io/nodejs:latest annotated
```

On 4.5 cluster
```
$ odo create nodejs:latest nodejs-app --project wcldizlwqg --context . --app app
Devfile path: 
 ⚠  Registry is empty, please run `odo registry add <registry name> <registry URL>` to add a registry

Validation
 ✓  Validating component [483ms]

Please use `odo push` command to create the component with source deployed

$ odo push --context .
Validation
 ✓  Checking component [864ms]

Configuration changes
 ✓  Initializing component
 ✓  Creating component [3s]
[...]
✓  Changes successfully pushed to component
```

On 3.11 cluster running on travis
```
Running odo with args [odo create nodejs:latest nodejs-app --project wcldizlwqg --context /tmp/396049560 --app app]
[...]
[odo] Devfile path: 
[odo] I0920 18:33:45.143396   28128 util.go:425] path /tmp/396049560/.odo/config.yaml doesn't exist, skipping it
[odo] I0920 18:33:45.143411   28128 util.go:425] path /tmp/396049560/.odo/env/env.yaml doesn't exist, skipping it
[...]

PR #4075, opened 2020-10-03
Refactor: fix s2i flag for odo create

### LINKED ISSUE #4048: odo create: --s2i flag is broken
/kind bug

1. If a devfile component already exists in the working file directory, `odo create <cmpType> --s2i` does not error out.

    image

2. Another bug has been introduced at https://github.com/openshift/odo/blame/master/pkg/odo/cli/component/create.go#L349 .
The `if !co.forceS2i { }` block has logic inside like https://github.com/openshift/odo/blame/master/pkg/odo/cli/component/create.go#L377  & https://github.com/openshift/odo/blame/master/pkg/odo/cli/component/create.go#L351 .. which requires `co.forceS2i == true`. This code will never execute since the outer blog requires `co.forceS2i == false` .  Therefore, conflicting flags are not being checked properly either.

dev-gaur:
**What type of PR is this?**

/kind bug
/kind code-refactoring

Fixes #4048 , #4018 

Please try out the `odo create` command using all the possible combinations and flags.

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -346,42 +346,45 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 	// this populates the LocalConfigInfo as well
 	co.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
 
+	err = co.checkConflictingFlags()
+	if err != nil {
+		return
+	}
+
+	var catalogList catalog.ComponentTypeList
+	if co.forceS2i {
+		client := co.Client
+		catalogList, err = catalog.ListComponents(client)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Configure the context
+	if co.componentContext != "" {
+		DevfilePath = filepath.Join(co.componentContext, devFile)
+		log.Infof("Devfile path: %s", co.DevfilePath)
+		EnvFilePath = filepath.Join(co.componentContext, envFile)
+		ConfigFilePath = filepath.Join(co.componentContext, configFile)
+		co.PushOptions.componentContext = co.componentContext
+	}
+	co.DevfilePath = DevfilePath
+
+	if util.CheckPathExists(ConfigFilePath) || util.CheckPathExists(EnvFilePath) {
+		return errors.New("this directory already contains a component")
+	}
+
+	if util.CheckPathExists(co.DevfilePath) && co.devfileMetadata.devfilePath.value != "" && !util.PathEqual(co.DevfilePath, co.devfileMetadata.devfilePath.value) {
+		return errors.New("this directory already contains a devfile, you can't specify devfile via --devfile")
+	}
+
+	co.appName = genericclioptions.ResolveAppFlag(cmd)
+
+	isDevfileRegistryPresent := true // defaulted to true since odo ships with a default registry set
+	var catalogDevfileList catalog.DevfileComponentTypeList
+
 	// If not using --s2i
 	if !co.forceS2i {
-
-		err = co.checkConflictingFlags()
-		if err != nil {
-			return
-		}
-
-		// Configure the context
-		if co.componentContext != "" {
-			DevfilePath = filepath.Join(co.componentContext, devFile)
-			log.Infof("Devfile path: %s", co.DevfilePath)
-			EnvFilePath = filepath.Join(co.componentContext, envFile)
-			ConfigFilePath = filepath.Join(co.componentContext, configFile)
-			co.PushOptions.componentContext = co.componentContext
-		}
-		co.DevfilePath = DevfilePath
-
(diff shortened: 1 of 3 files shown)


dev-gaur:
image

@prietyc123 satisfied with the message, when devfile does not exist?

pkg/odo/cli/component/create.go
```diff
@@ -345,42 +345,45 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 	// this populates the LocalConfigInfo as well
 	co.Context = genericclioptions.NewContextCreatingAppIfNeeded(cmd)
 
-	// If not using --s2i
-	if !co.forceS2i {
+	err = co.checkConflictingFlags()
+	if err != nil {
+		return
```
cdrage, line 350: return... what? This should be returning `err`

pkg/odo/cli/component/create.go
```diff
@@ -542,7 +539,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 
 		if util.CheckPathExists(co.DevfilePath) || co.devfileMetadata.devfilePath.value != "" {
 			// Categorize the sections
-			log.Info("Validation")
+			log.Info("Validation for Devfile component")
```
cdrage, line 542: Don't change this IMO, it should just stay as `Validation`. No need to specify Devfile. We're trying to avoid using the word Devfile anyways in the UI, since it's actually what's being used "under the hood" / engine, rather than what the user will be directly interacting with.

pkg/odo/cli/component/create.go
```diff
@@ -567,70 +564,77 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
-			// supported by devfile, if it is supported we return and will download the corresponding devfile later,
-			// if it is not supported we still need to run all the codes related with s2i after devfile compatibility check
+		// Since we need to support both devfile and s2i, so we have to check if the component type is
+		// supported by devfile, if it is supported we return and will download the corresponding devfile later,
+		// if it is not supported we still need to run all the codes related with s2i after devfile compatibility check
 
-			hasComponent := false
-			devfileExistSpinner := log.Spinner("Checking devfile existence")
+		hasComponent := false
+		var devfileExistSpinner *log.Status
```
cdrage, line 579: As someone who wrote the `log` library and functionality, I'd suggest we avoid declaring a spinner like this.. 

You may run into issues in the future where the spinner will encounter a nil pointer.

pkg/odo/cli/component/create.go
```diff
@@ -849,7 +855,7 @@ func (co *CreateOptions) Validate() (err error) {
 		return nil
 	}
 
-	log.Info("Validation")
+	log.Info("Validation for S2I component")
```
cdrage, line 858: Just keep it as `Validation`, I don't think we should declare that it's S2I.. As that would confuse the user.

cdrage (changes requested):
Few changes, otherwise LGTM

pkg/odo/cli/component/create.go
dev-gaur: Hi @cdrage , If we keep it just as `Validation` then the logs come out as following if an S2I component is created by default.

image

IMO, some verbosity could actually help in this case. Also, this verbosity will actually fix #4018 

@girishramnani @kadel @adisky @mik-dass you thoughts are welcome as well

dev-gaur: Currently it looks like this.

image

dev-gaur: Maybe we can revert back to just `Validation` after we get rid of S2I totally ?

pkg/odo/cli/component/create.go
dev-gaur: Its defined here. https://github.com/openshift/odo/blob/ce72fb36681631b908ad7ca4731b2c4623db9823/pkg/odo/cli/component/create.go#L345 

That's a golang feature you can define the variable in the return block of function definition

pkg/odo/cli/component/create.go
adisky: I think using only `Validation` is ok, explicitly specifying `Validation for devfile component`/ `Validation for S2I component` does not seem very user friendly. To solve the problem in better way  we should use a common spinner ` Validating the component` if one of the s2i or devfile component found just continue with it.
```
$ odo create java
Validation 
 ✓  Validating the component 
```

if no s2i or devfile component found
```
$ odo create java
Validation 
 ✗ No devfile or s2i component found, please run `odo catalog list components` for a list of supported components.
```

cdrage: I agree with @adisky we should keep it as `Validation`. We should not be outputting both s2i and devfile in the output either... 

Having both would indeed confused the user

Issue #4073, opened 2020-10-02
Refactor devfile parser code to produce a common devfile parser can work in all tools


elsony:
/kind user-story

## User Story
As a tools developer, I want to use a common devfile library to read and write a devfile.

## Background
There are currently multiple instances of the parser implementation, e.g. the parser in odo repo and also the one in https://github.com/devfile/parser. This work will consolidate those implementations and create one single parser implementation and make it available to on https://github.com/devfile/parser.

- This work will be started from the parser code in the odo repo
- Restructure the code in odo repo to extract the generic parser library (refactor out anything that is odo specific)
- May include any extra functions that is currently available in the existing devfile/parser repo but not available in the odo repo
- Move the common library to the devfile/parser repo
- odo will switch to use the common library on the devfile parser repo and remove the corresponding code from the odo repo

## Acceptance Criteria 
- [ ] The common library parser should only contain support for the devfile 2.0 specification, any odo specific support will be move outside of the common library.
- [ ] Support both read and write of devfile

## Links 
- Related issue: This issue provides the ground work of the devfile libraryhttps://github.com/devfile/api/issues/144

Detailed design notes: https://docs.google.com/document/d/1ffLi37fM7LAjRK5SXAw-yVNwr9GchFYy_e_5VD-FOZc/

/kind user-story
/area devfile

maysunfaisal:
@girishramnani any concerns with the issue description, since there is a triage needs information tag? The issue OP outlines what is required to be done and the detailed work has been outlined in the attached document. Do we want to copy paste it out here?

maysunfaisal:
@yangcao77  validate pkg has been refactored and I guess there is nothing that needs to be refactored in the parser pkg. So we should be able to just copy paste any of the pkgs to devfile/parser or devfile/api..

I think this can be closed now?

yangcao77:
Yes. I think we can close this one. We can have separate issues for 
```

- Move the common library to the devfile/parser repo

- odo will switch to use the common library on the devfile parser repo and remove the corresponding code from the odo repo

```

PR #3959, opened 2020-09-15
add container name validation during creation and push

yangcao77:
Signed-off-by: Stephanie <yangcao@redhat.com>

**What type of PR is this?**
/kind bug
/area devfile

**What does does this PR do / why we need it**:
Add  validation for devfile container name during component creation, and push. Block the action if the container name is not a valid K8s resourece name. 

**Which issue(s) this PR fixes**:

Fixes #3698 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

test component creation:
create a devfile with a long container name, and run `odo create --devfile <devfile path>`
```
$ odo create --devfile ./devfile.yaml 
Validation
 ✓  Creating a devfile component from devfile path: /Users/stephanie/Documents/devfile-spring2/devfile.yaml [62590ns]
 ✓  Validating devfile component [277583ns]
 ✗  endpoint name "runtimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntime" is not valid, endpoint name should conform the following requirements: 
- Contain at most 63 characters
- Contain only lowercase alphanumeric characters or ‘-’
- Start with an alphanumeric character
- End with an alphanumeric character
- Must not contain all numeric values
```

test odo push:
`odo create nodejs --starter`
modify the devfile to use a long container name
run `odo push`
```
$ odo push
 ✗  Validating K8s Resources Name [132166ns]
 ✗  endpoint name "runtimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntime" is not valid, endpoint name should conform the following requirements: 
- Contain at most 63 characters
- Contain only lowercase alphanumeric characters or ‘-’
- Start with an alphanumeric character
- End with an alphanumeric character
- Must not contain all numeric values

```

PR #4085, opened 2020-10-06
Refactor validate pkg for generic and odo validation

maysunfaisal:
Signed-off-by: Maysun J Faisal <maysunaneek@gmail.com>

**What type of PR is this?**
/kind cleanup
/kind code-refactoring

We need this PR to refactor the validation pkg to odo specific validation and devfile generic validation, so that the generic validation can be exported out to the ~devfile/parser~ devfile/api repo

#4073

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Run the unit tests in the validate pkg


pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -59,6 +59,16 @@ func (dc DevfileCommand) GetExecWorkingDir() string {
 	return ""
 }
 
+// IsExec checks if the command is an exec command
+func (dc DevfileCommand) IsExec() bool {
```
cdrage, line 63: New function means new tests :)

Despite this test being extremely small haha, we should add a test function for this!

cdrage (approved):
Thanks for refactoring and moving the tests!

/approve

pkg/devfile/parser/data/common/command_helper.go
maysunfaisal: @cdrage yes, I'll add the new test in ;)

could you remove the lgtm label? it looks like it got added?

maysunfaisal:
@mik-dass @adisky could you pls review this PR? This PR refactors the validation pkg into odo specific and devfile specific logic; so that the devfile logic can be moved out to devfile/api repo in the future

yangcao77:
You might want to rebase due to merge of this pr: https://github.com/openshift/odo/pull/3959/files

pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -59,6 +59,16 @@ func (dc DevfileCommand) GetExecWorkingDir() string {
...
+	isExec := false
+	if dc.Exec != nil {
+		isExec = true
+	}
+
+	return isExec
```
yangcao77, line 69: can we just `return dc.Exec!=nil`?

yangcao77: Actually, `command.Composite != nil` and `command.Exec != nil` can be used. why we need an extra function?

adisky: yes

pkg/devfile/parser/data/common/command_helper_test.go
```diff
@@ -231,40 +231,52 @@ func TestGetExecWorkingDir(t *testing.T) {
 
 }
 
-func TestIsComposite(t *testing.T) {
+func TestIsCompositeAndIsExec(t *testing.T) {
```
adisky, line 234: IMO, we don't need a test case for these functions, they are very simple.

pkg/devfile/validate/commands.go
```diff
@@ -1,115 +1,52 @@
 package validate
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/openshift/odo/pkg/devfile/parser/data/common"
+	genericValidation "github.com/openshift/odo/pkg/devfile/validate/generic"
```
adisky, line 5: I think the alias also not needed, generic.ValidateCommands is pretty intuitive

adisky (commented):
@maysunfaisal If generic validation is extracted out for library, so the library is going to provide separate function to validate each resource? or it is only the first step for refactoring?  

i think odo does not need to call each function `isEventValid`, `ValidateCommands` etc, it should call a single function `generic.Validate()` and then if everything is good from generic validation, odo starts own validation, I think both the validation should be decoupled more.

pkg/devfile/validate/commands.go
```diff
@@ -1,115 +1,52 @@
...
-		processedCommands[commandID] = commandID
-
 		err = validateCommand(command, commandsMap, components)
 		if err != nil {
 			return err
 		}
 	}
 
-	return
+	err = genericValidation.ValidateCommands(commands, commandsMap, components)
```
mik-dass, line 21: we can return here

pkg/devfile/validate/generic/commands.go
```diff
@@ -0,0 +1,130 @@
...
+// validateCommand validates a given devfile command
+func validateCommand(command common.DevfileCommand, devfileCommands map[string]common.DevfileCommand, components []common.DevfileComponent) (err error) {
+
+	// If the command is a composite command, need to validate that it is valid
+	if command.IsComposite() {
+		parentCommands := make(map[string]string)
+		return ValidateCompositeCommand(&command, parentCommands, devfileCommands, components)
+	}
+
+	err = ValidateExecCommand(command, components)
```
mik-dass, line 42: we can return here

pkg/devfile/validate/components.go
```diff
@@ -1,81 +1,35 @@
...
-	for volumeMountName := range processedVolumeMounts {
-		if _, ok := processedVolumes[volumeMountName]; !ok {
-			invalidVolumeMounts = append(invalidVolumeMounts, volumeMountName)
-		}
-	}
-
-	if len(invalidVolumeMounts) > 0 {
-		return &MissingVolumeMountError{volumeName: strings.Join(invalidVolumeMounts, ",")}
-	}
+	err := genericValidation.ValidateComponents(components)
```
mik-dass, line 32: we can return here

pkg/devfile/validate/generic/events.go
```diff
@@ -0,0 +1,42 @@
...
+// IsEventValid checks if events belonging to a specific event type are valid ie; event should map to a valid devfile command
+func IsEventValid(eventNames []string, eventType string, commands map[string]common.DevfileCommand) error {
+	eventErrorMsg := make(map[string][]string)
+	eventErrors := ""
+
+	for _, eventName := range eventNames {
+		isEventPresent := false
+
+		if _, ok := commands[strings.ToLower(eventName)]; ok {
+			isEventPresent = true
```
mik-dass, line 20: I think we don't need this variable. We simplify it like
```
if _, ok := commands[strings.ToLower(eventName)]; !ok {
	klog.V(2).Infof("%s type event %s does not map to a valid devfile command", eventType, eventName)
	eventErrorMsg[strings.ToLower(eventName)] = append(eventErrorMsg[strings.ToLower(eventName)], fmt.Sprintf("event %s does not map to a valid devfile command", eventName))
}
```

pkg/devfile/validate/events.go
```diff
@@ -47,21 +61,15 @@ func validateEvents(events common.DevfileEvents, commands map[string]common.Devf
...
 	for _, eventName := range eventNames {
-		isEventPresent := false
 		for _, command := range commands {
-			// Check if event matches a valid devfile command
 			if command.GetID() == strings.ToLower(eventName) {
-				isEventPresent = true
-
-				// Check if the devfile command is valid
+				// Check if the devfile command is a valid odo devfile command
 				err := validateCommand(command, commands, components)
```
mik-dass, line 73: Since the `commands` map contain all the commands in the devfile and they are all validated before this call, can we just check if a `eventName` is in the `commands` map instead of validating each event again? What am I missing?

pkg/devfile/parser/data/common/command_helper.go
maysunfaisal: It looks prettier, if you're very opposed to it; I can remove it. These other funcs has been in the code, so I would have to update them all.

pkg/devfile/validate/events.go
maysunfaisal: good point

pkg/devfile/parser/data/common/command_helper.go
adisky: No problem,  I am fine with this.

girishramnani:
perfect 1000 lines added!!

maysunfaisal:
@girishramnani i would say 85-90% of those new lines are unit tests which will be moved to devfile/parser or devfile/api along with the generic pkg. I'm unit testing the generic func but since odo will consume them as well, i've added the same test for the odo repo. Lmk, if you think otherwise.

pkg/devfile/parser/data/common/command_helper.go
maysunfaisal: I've removed it. I think more ppl will question it and not use it. Its not worth explaining myself everytime 😄

pkg/devfile/validate/events.go
maysunfaisal: updated!, pls help review!

pkg/devfile/validate/commands.go
maysunfaisal: updated!, pls help review! 

I've also added a single point of validation call for generic validation like you suggested and it makes sense.

pkg/devfile/validate/generic/errors.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// ExecCommandMissingCommandLineError returns an error if the exec command does not have a command line
+type ExecCommandMissingCommandLineError struct {
+	commandId string
+}
+
+func (e *ExecCommandMissingCommandLineError) Error() string {
+	return fmt.Sprintf("exec command %q must have a command", e.commandId)
+}
```
adisky, line 40: should this all be covered in a single error like `MissingInformationError`?

pkg/devfile/validate/generic/errors.go
```diff
@@ -0,0 +1,136 @@
...
+
+// CompositeIndirectReferenceError returns an error if the composite command indirectly references itself
+type CompositeIndirectReferenceError struct {
+	commandId string
+}
+
+func (e *CompositeIndirectReferenceError) Error() string {
+	return fmt.Sprintf("the composite command %q cannot indirectly reference itself", e.commandId)
+}
+
```
adisky, line 78: I think the errors could me refactored more here, like instead of having separate error for each validation fail it could be `InvalidCommand`  `MissingInformation`

pkg/devfile/validate/validate.go
```diff
@@ -5,37 +5,36 @@ import (
...
 		}
 
 		// Validate all the devfile commands before validating events
-		if err := validateCommands(d.Commands, commandsMap, components); err != nil {
-			return err
-		}
-
-		// Validate all the events
-		if err := validateEvents(events, commandsMap, components); err != nil {
+		if err := validateCommands(commandsMap); err != nil {
```
adisky, line 37: I think `ValidateContainerName`, should also be part of generic validation, it could be added inside component validation

adisky (commented):
Mostly looks good to me, a nit here update the comment https://github.com/openshift/odo/blob/master/pkg/devfile/parse.go#L41

pkg/devfile/validate/validate.go
maysunfaisal: There are a few reasons why I did not do it:
1. validation pkg is generally for validating logic directly related to the devfile schema
2. having said point 1, this func should not have been in validate pkg in the first place and is relying on adapter util funcs which further adds reason behind my point no 1.

Odo is the only container based tool. Workspace may need it but I dont see Console using it? 🤔  
Lmk, what you think!

pkg/devfile/validate/generic/errors.go
adisky: This error refactoring could be done in separate PR, for now its ok.

pkg/devfile/validate/validate.go
adisky: @maysunfaisal It is ok to not to add in library, just wanted an opinion essentially it is the validation of component name specified in the devfile, the container component would be translated to containers to all other tools, no?  what console is translating devfile containers to if specified?

pkg/devfile/validate/generic/errors.go
maysunfaisal: I've already refactored, just need to push my changes

pkg/devfile/validate/validate.go
maysunfaisal: @adisky I've updated the pkg to validate the component names and made it more cleaner, pls have a look! Thx.

pkg/devfile/validate/generic/components.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	adaptersCommon "github.com/openshift/odo/pkg/devfile/adapters/common"
+	"github.com/openshift/odo/pkg/devfile/parser/data/common"
+	"github.com/openshift/odo/pkg/util"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
```
adisky, line 11: we should not import odo packages in library, move this function in generic package `util.ValidateK8sResourceName`

```
"github.com/openshift/odo/pkg/util"
```

not sure about this one, anyways the constants are not related with adapters, it is a devfile spec so can be moved in types package.
```
"github.com/openshift/odo/pkg/devfile/adapters/common
```

for below one i understand, it will get updated when we use types from devfile/api
```
github.com/openshift/odo/pkg/devfile/parser/data/common
```

maysunfaisal: 1. So devfile/parser already has a copy of the util func ValidateK8sResourceName https://github.com/devfile/parser/blob/master/pkg/util/util.go#L1044. I plan on using that when its exported out. It will just be changing the import.
2. `adaptersCommon "github.com/openshift/odo/pkg/devfile/adapters/common"` are just using constants. There will be new const when its moved to devfile/parser. Since this is still in odo, I dont want to maintain two separate const for the same thing.

Lmk, what you think @adisky

adisky: gotcha!

PR #4086, opened 2020-10-06
Update vendor-update makefile target

johnmcollier:
Signed-off-by: John Collier <jcollier@redhat.com>

**What type of PR is this?**

/kind bug

Ran in to this while trying to add a dependency to odo and realized that `make vendor-update` needs to be updated since odo uses go modules now.

This PR updates the `vendor-update` make target to use `go mod vendor` rather than using `glide up`.

N/A

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -402,7 +402,7 @@ upload-packages:
 # Update vendoring
 .PHONY: vendor-update
 vendor-update:
-	glide update --strip-vendor
+	go mod vendor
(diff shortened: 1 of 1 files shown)


cdrage:
Super small change..

so..

/approve
/lgtm

Issue #4027, opened 2020-09-23
New spring devfile does not work on default namespace


maysunfaisal:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**

## How did you run odo exactly?
odo push

## Actual behavior
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo push

Validation
 ✓  Validating the devfile [57999ns]

Creating Kubernetes resources for component myjava
 ✓  Waiting for component to start [22s]

Applying URL changes
 ✓  URL myingress successfully deleted

Syncing to component myjava
 ✓  Checking file changes for pushing [4ms]
 ✓  Syncing files to the component [3s]

Executing devfile commands for component myjava
 ✗  Executing build command "mvn clean -Dmaven.repo.local=/mypath/.m2/repository package -Dmaven.test.skip=true" [2s]
 ✗  Failed to start component with name myjava. Error: Failed to create the component: command execution failed: unable to execute the run command: unable to exec command [/bin/sh -c mvn clean -Dmaven.repo.local=/mypath/.m2/repository package -Dmaven.test.skip=true]: 
[ERROR] Could not create local repository at /mypath/.m2/repository -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/LocalRepositoryNotAccessibleException
: error while streaming command: command terminated with exit code 1
```

I manually tried it out inside the container, and looks like the container user does not have sufficient privilege to create dir:
```
I have no name!@myjava-556574d744-45hfg:/projects$ mvn clean -Dmaven.repo.local=/home/user/.m2/repository package -Dmaven.test.skip=true
[ERROR] Could not create local repository at /home/user/.m2/repository -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/LocalRepositoryNotAccessibleException
[...]

Issue #3947, opened 2020-09-14
rename odo env variable namespace to project


kadel:
We are using "project" everywhere else in odo. We even have `odo project` command.

`odo env` should follow the same terminology

metacosm:
I understand the will to make things consistent. However, doesn't that go against the idea of making odo more k8s-native as opposed to OpenShift-centric? If anything, wouldn't it make more sense (though, of course, that's a much bigger endeavor) to rename everything project-related to namespace?

kadel:
I view it similarly as what we do with url. We have `odo url` as an abstraction for Ingress and Routes. So here I look at the project not as OpenShift project, but as a generic project term that has nothing to do with OpenShift projects, but it is more abstraction on top of Kubernetes Namespaces and OpenShift projects.

Issue #3775, opened 2020-08-18
Allow users to toggle between ephemeral and persistent volumes for devfiles


johnmcollier:
/kind user-story
/area devfile

## User Story
Currently, odo uses an ephemeral emptyDir volume to store the synced project source code in the remote container of a devfile component. There's some use cases however where this is not ideal, such as when the pod needs to be restarted for whatever reason (env update, linking, etc), and then the source code needs to be re-synced. Additionally, to support scenarios with containers in their own dedicated pod (via the `dedicatedPod` field), a single shared emptyDir volume cannot be used.

As a user, I want to have the choice to use persistent storage for my project source if I desire, but also still have the freedom to still use emptyDir/ephemeral volumes if I want or if my cluster does not have storage configured. 

Eclipse Che uses the `persistVolumes` attribute in devfiles to toggle this functionality (default value is true when unset). This would be a good attribute to support, especially if we're still aiming for compatibility with Che devfiles:
<img width="948" alt="Screen Shot 2020-08-18 at 11 44 15 AM" src="https://user-images.githubusercontent.com/6880023/90534844-374e6180-e148-11ea-97d5-b1e052e858d6.png">

Additionally, we could provide a global-level preference in the `~/.odo/preferences.yaml` file if users didn't want to set it on a per-workspace basis.

## Acceptance Criteria 
- [ ] It should allow users to toggle between emptyDir and Persistent Volumes for storage. There should be a default value.
- [ ] It should either be a global setting, or component-level setting (via the `persistVolumes` attribute) or both.

## Links 
- Feature Request: 
- Related issue: https://github.com/openshift/odo/issues/3562

/kind user-story

adisky:
Using a pvc would also solve some of this problems
https://github.com/openshift/odo/issues/4050
https://github.com/openshift/odo/issues/4017

as discussed with @kadel we want to keep pvc as default option, and also allow users to set it to  `emptyDir`

adisky:
findings till now
-  [volume access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) is not a problem as long as we are using single pod with a component
- issue is divided in two parts
   - Adding support to use pvc to store source code
   - Updating the source code sync code.

Apart from this, do we want to add the source code pvc to devfile and do we want to show it in `odo storage list`?

adisky:
Things to be done to support pvc
[...]

Issue #2680, opened 2020-03-05
mkdir: cannot create directory '/mnt/app-root.setup-volume': Permission denied


jichenjc:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**
latest (build from master from last week)
## How did you run odo exactly?

this is similar to following 
 https://github.com/openshift/odo/issues/1623

but I used hostPath and NFS both got same error and I think it's not related to anyuid...
I added logs and showed the pv is mount as root:root and as normal user (1000580000), I can't create things there, 

from https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

the files or directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged Container or modify the file permissions on the host to be able to write to a hostPath volume

so at least hostPath doesn't work, NFS from my test also doesn't work
so the question is what we did on odo CI , which things are changed or PV used?

[root@infnod-1 ~]# crictl logs 15d0e7b81001e
2020/02/27 06:47:57 [go-init] Pre-start command launched : /opt/odo/bin/s2i-setup
1000580000
+ set -eo pipefail
+ whoami
+ id
uid=1000580000(1000580000) gid=0(root) groups=0(root),1000580000
+ ls -l /
total 8
lrwxrwxrwx.   1 root root    7 Jan 28 10:42 bin -> usr/bin
dr-xr-xr-x.   2 root root    6 Jan 28 10:42 boot
drwxr-xr-x.   5 root root  360 Feb 27 06:47 dev
drwxr-xr-x.   1 root root   44 Jan 29 14:29 etc
-rw-r--r--.   1 root root 7753 Jan 29 14:29 help.1
drwxr-xr-x.   2 root root    6 Jan 28 10:42 home
lrwxrwxrwx.   1 root root    7 Jan 28 10:42 lib -> usr/lib
lrwxrwxrwx.   1 root root    9 Jan 28 10:42 lib64 -> usr/lib64
drwxr-xr-x.   2 root root    6 Dec 14  2017 media
drwxr-xr-x.   2 root root    6 Dec 14  2017 mnt
drwxr-xr-x.   1 root root   17 Feb 27 06:47 opt
dr-xr-xr-x. 190 root root    0 Feb 27 06:47 proc
dr-xr-x---.   1 root root   23 Jan 28 10:54 root
drwxr-xr-x.   1 root root   18 Jan 29 14:29 run
[...]

PR #4092, opened 2020-10-07
Add warning for default namespace during odo create

### LINKED ISSUE #4053: Discussion about PVC usage for non root containers in OpenShift default namespace
When debugging issue https://github.com/openshift/odo/issues/4027, we noticed that any stack user is going to face a permission issue if they modify the pvc volume mount path outside of what is prescribed by the image in default namespace.

For example, if the image gave write permissions to a path `/mypath1` and the devfile pvc uses it in default namespace, it is fine. However if the user updates their devfile pvc volume mount path to `/mypath2`, then `odo push` is going to fail because `/mypath2` does not have the write permission but `/mypath1` did which was set by the image's dockerfile.

This however, is not an issue in non-default OpenShift namespaces. The pvc path has the correct ownership for the arbitrary user id and is able to read/write in the pvc path.

adisky:
**What type of PR is this?**

 /kind bug

Add explicit warning while using default namespace with odo, component fails due to volume permission issue
I have added warning for `odo push` only as in we want `odo create` to run without cluster access so avoided adding namespace warning in `odo create`

Fixes #4053 #4027 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -140,6 +140,10 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string) 
 		// The namespace was retrieved from the --project flag (or from the kube client if not set) and stored in kclient when initalizing the context
 		if !pushtarget.IsPushTargetDocker() {
 			po.namespace = po.KClient.Namespace
+			// log warning in case of default namespace
+			if po.namespace == "default" {
+				log.Warning("odo may not work as expected in default namespace, please run odo component in non default namespace")
+			}
 		}
 
 		return nil
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -140,10 +140,6 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string) 
 		// The namespace was retrieved from the --project flag (or from the kube client if not set) and stored in kclient when initalizing the context
 		if !pushtarget.IsPushTargetDocker() {
 			po.namespace = po.KClient.Namespace
-			// log warning in case of default namespace
-			if po.namespace == "default" {
-				log.Warning("odo may not work as expected in default project, please run odo component in non default project. You can use `odo project create` to create a new project")
-			}
 		}
 
 		return nil
--- a/pkg/odo/cli/component/push.go
+++ b/pkg/odo/cli/component/push.go
@@ -142,7 +142,7 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string) 
 			po.namespace = po.KClient.Namespace
 			// log warning in case of default namespace
 			if po.namespace == "default" {
-				log.Warning("odo may not work as expected in default namespace, please run odo component in non default namespace")
+				log.Warning("odo may not work as expected in default project, please run odo component in non default project. You can use `odo project create` to create a new project")
 			}
 		}
 
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -536,7 +536,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		}
 
 		if componentNamespace == "default" {
-			log.Warning("odo may not work as expected in default project, please run odo component in non default project. You can use `odo project create` to create a new project")
+			log.Warning("odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.")
 		}
 
 		// Set devfileMetadata struct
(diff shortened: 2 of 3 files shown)


pkg/odo/cli/component/push.go
```diff
@@ -140,6 +140,10 @@ func (po *PushOptions) Complete(name string, cmd *cobra.Command, args []string)
 		// The namespace was retrieved from the --project flag (or from the kube client if not set) and stored in kclient when initalizing the context
 		if !pushtarget.IsPushTargetDocker() {
 			po.namespace = po.KClient.Namespace
+			// log warning in case of default namespace
+			if po.namespace == "default" {
+				log.Warning("odo may not work as expected in default namespace, please run odo component in non default namespace")
```
dharmit, line 145: In v1, we used to create a `myproject` namespace for the user if they were working on a fresh cluster (which is when I think we will most likely face the issue we're trying to address here.)

If we don't want to do that, we should add something in the Warning that says that they can use `odo project create` to create a new namespace/project. Also, aren't we using the word "project" for odo users instead of "namespace"?

dharmit (changes requested):

pkg/odo/cli/component/push.go
kadel: > Warning that says that they can use odo project create to create a new namespace/project

+1 for mentioning that user should create a new project using `odo project create`



> Also, aren't we using the word "project" for odo users instead of "namespace"?

-1 we should be using "project" everywhere (https://github.com/openshift/odo/issues/3947#issuecomment-692789890)

adisky: @dharmit @kadel Please check, is it look fine now?

maysunfaisal (changes requested):
I would add this warning in `odo create` because the namespace info is stored in env.yaml 😄 

```
Maysuns-MacBook-Pro:nodejs-ex maysun$ odo create nodejs mynode1
Validation
 ✓  Checking devfile existence [65134ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [77025ns]
 ✓  Validating devfile component [372439ns]

Maysuns-MacBook-Pro:nodejs-ex maysun$ cat .odo/env/env.yaml 
ComponentSettings:
  Name: mynode1
  Project: mjf
  AppName: app
Maysuns-MacBook
```

otherwise the user would have to delete their component and re-create it again! 🤔

adisky:
@maysunfaisal you are right we should add it in `odo create`, i have added in `odo push` because in future we want to run odo create without having cluster access, but currently it is not the case so for now we can add it `odo create`, would update it.

kadel:
@maysunfaisal  has a good point.

We should do this in `odo create`.
@adisky we should not need a cluster to check what namespace is being used. The information about what namespace should be used is stored in kubeconfig.

adisky:
@kadel @maysunfaisal Updated

adisky:
> @adisky we should not need a cluster to check what namespace is being used. The information about what namespace should be used is stored in kubeconfig.

@kadel It means we expect the user to have valid kubeconfig while doing `odo create`? It saves the namespace from kubeconfig in env.yaml and later on during odo push we resolve the correct namespace?
I think we need to clarify our behavior for `odo create`, all namespace resolution currently we are doing in this step only.

dharmit:
Please pardon my lack of understanding but merging this PR would mean that whenever a user is trying to push to `default` namespace, this warning message will show up, right? Even if they're not using the maven and spring component types?

I admit that I don't fully understand what we're trying to fix here and so it feels like we're adding a warning message that might not be relevant if user is working with different component type than maven/spring.

adisky:
> Please pardon my lack of understanding but merging this PR would mean that whenever a user is trying to push to `default` namespace, this warning message will show up, right? Even if they're not using the maven and spring component types?
> 
> I admit that I don't fully understand what we're trying to fix here and so it feels like we're adding a warning message that might not be relevant if user is working with different component type than maven/spring.

@dharmit it is basically related when we are using pvc for non-root containers in default namespace, currently we are using pvc only in spring boot component for maven cache, but when we implement this https://github.com/openshift/odo/issues/3775 ie start using pvc for source code in all components this issue will be valid for all components.

In past we have faced similar issue for s2i components https://github.com/openshift/odo/issues/2680
https://github.com/openshift/odo/issues/1623

maysunfaisal (commented):
Just a nit, lmk what you think

odo may not work as expected in **_a_** default project, please run **_the_** odo component in **_a non-default_** project. **_To create a new project, please use `odo project create`_**

adisky:
> Just a nit, lmk what you think
> 
> odo may not work as expected in **_a_** default project, please run **_the_** odo component in **_a non-default_** project. **_To create a new project, please use `odo project create`_**

@maysunfaisal updated, PTAL.

maysunfaisal (approved):
/lgtm

PR #4121, opened 2020-10-19
Updating golang version to 1.15

girishramnani:
**What type of PR is this?**

> Documentation changes: Please include [skip ci] in your commit message as well
> /kind documentation
> [skip ci]

**What does does this PR do / why we need it**:

**Which issue(s) this PR fixes**:

Currently the prow is failing, seems like golang 1.13 was deprecated.

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

PR #4132, opened 2020-10-22
Revert "Updating golang version to 1.15"

girishramnani:
Reverts openshift/odo#4121


kadel:
/approve

/override ci/prow/unit
^ this would never pass, as tests are using Dockerfiles from master not form PR

Issue #3902, opened 2020-09-04
odo describe --context option is ignored when current directory is a devfile based component


dgolovin:
/kind bug

## What versions of software are you using?

**Operating System:** Linux

**Output of `odo version`:** 2.0.0 built from master with pr odo describe for devfile component applied

## How did you run odo exactly?
```
~/p/test-ws/nodejs-ex$ ls 
delete  devfile.yaml  helm  openshift  package.json  push  README.md  server.js  tests  views

~/p/test-ws/nodejs-ex$ odo describe
Component Name: nodejs-ex
Type: devfile

~/p/test-ws/nodejs-ex$ odo describe --context ../import1
 ✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component

~/p/test-ws/nodejs-ex$ cd ../import1

~/p/test-ws/import1$ ls -a .odo
.  ..  config.yaml

~/p/test-ws/import1$ odo describe
Component Name: component1
Type: nodejs
Source: file://./
Environment Variables:
 · DEBUG_PORT=5858

~/p/test-ws/import1$ odo describe --context ../nodeja-ex1
Component Name: nodejs-ex1
Type: nodejs:latest
Source: file://./

~/p/test-ws/import1$ cd ../nodeja-ex1/
~/p/test-ws/nodeja-ex1$ ls .odo -a
.  ..  config.yaml  odo-file-index.json
```

## Actual behavior

-- context option ignored when current component is devfile, --context works fine for s2i components

## Expected behavior

Should describe component in for --context option path

## Any logs, error output, etc?

N/A

girishramnani:
I can take a look at this issue

PR #4149, opened 2020-10-24
check if devfile exist in the context folder instead of working directory

girishramnani:
**What type of PR is this?**
/kind bug

Currently we were deciding the type of context ( devfile or s2i ) based on presence of devfile in the working directory which is a problem when user provides `--context`.  For e.g.
- `odo describe --context <s2i-component-dir>` <- this from a directory that has devfile directory would fail as it would initialise a `DevfileContext` that looks for `env.yaml`. 

Fixes https://github.com/openshift/odo/issues/3902

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- create a nodejs s2i component - `nodejs2i`
- create a nodejs devfile component - `nodejsdevfile` along side the s2i component so that both sit in the same dir
- now go to the `nodejsdevfile` component dir and run `odo describe --context ../nodejs2i`

Before this PR it should fail with
```
```

Now it should show the description.

### DIFF
--- a/tests/integration/devfile/cmd_devfile_create_test.go
+++ b/tests/integration/devfile/cmd_devfile_create_test.go
@@ -267,10 +267,10 @@ var _ = Describe("odo devfile create command tests", func() {
 		It("checks that odo describe works for s2i component from a devfile directory", func() {
 			helper.Chdir(newContext)
 			cmpName2 := helper.RandString(6)
-			output := helper.CmdShouldPass("odo", "create", "--starter", "nodejs")
+			helper.CmdShouldPass("odo", "create", "--starter", "nodejs")
 			context2 := helper.CreateNewContext()
 			helper.CmdShouldPass("odo", "create", "--s2i", "nodejs", "--context", context2, cmpName2)
-			output = helper.CmdShouldPass("odo", "describe", "--context", context2)
+			output := helper.CmdShouldPass("odo", "describe", "--context", context2)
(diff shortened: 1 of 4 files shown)


kadel:
/approve

Wouldn't this be a good candidate for an integration test?

girishramnani:
> /approve
> 
> Wouldn't this be a good candidate for an integration test?

agreed adding an integration test

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -206,12 +206,16 @@ var _ = Describe("odo devfile create command tests", func() {
...
 	Context("When executing odo create with --s2i flag", func() {
+		var newContext string
 		JustBeforeEach(func() {
-			newContext := path.Join(commonVar.Context, "newContext")
+			newContext = path.Join(commonVar.Context, "newContext")
 			devfilePath = filepath.Join(newContext, devfile)
 			helper.MakeDir(newContext)
 			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", devfile), devfilePath)
 		})
+		AfterEach(func() {
```
dharmit, line 216: Nit: should this be `JustAfterEach`? Also, I think, we have a common `BeforeEach` & `AfterEach`.

dharmit (changes requested):
One nit. Looks good otherwise.

dharmit (approved):
/lgtm

Issue #4144, opened 2020-10-24
odo list showing a duplicate entry of devfile component


girishramnani:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`: odo master

## How did you run odo exactly?
```
girish@MacBook-Pro rtr % odo project create girishn
 ✓  Project 'girishn' is ready for use
 ✓  New project created and now using project: girishn

(base) girish@MacBook-Pro rtr % odo project set girishn   
Already on project : girishn

(base) girish@MacBook-Pro rtr % odo create nodejs --starter --app new-app
...
Starter Project
 ✓  Downloading starter project nodejs-starter from https://github.com/odo-devfiles/nodejs-ex.git [2s]
Please use `odo push` command to create the component with source deployed

(base) girish@MacBook-Pro rtr % odo list
Devfile Components: 
APP         NAME       PROJECT     TYPE       STATE
new-app     nodejs     girishn     nodejs     Not Pushed

(base) girish@MacBook-Pro rtr % odo app list                             
There are no applications deployed in the project 'girishn'

(base) girish@MacBook-Pro rtr % odo push
...
Pushing devfile component nodejs
 ✓  Changes successfully pushed to component
```

## Actual behavior
```
(base) girish@MacBook-Pro rtr % odo list --project girishn --all-apps
Devfile Components: 
APP         NAME       PROJECT     TYPE       STATE
new-app     nodejs     girishn     nodejs     Pushed


Openshift Components: 
APP         NAME       PROJECT     TYPE       SOURCETYPE     STATE
new-app     nodejs     girishn     nodejs                    Pushed
```

as well as --app new-app
```
(base) girish@MacBook-Pro rtr % odo list --project girishn --app new-app
Devfile Components: 
APP         NAME       PROJECT     TYPE       STATE
new-app     nodejs     girishn     nodejs     Pushed


Openshift Components: 
APP         NAME       PROJECT     TYPE       SOURCETYPE     STATE
new-app     nodejs     girishn     nodejs                    Pushed
```

## Expected behavior
```
(base) girish@MacBook-Pro rtr % odo list --project girishn --app new-app
Devfile Components: 
APP         NAME       PROJECT     TYPE       STATE
new-app     nodejs     girishn     nodejs     Pushed
```

## Any logs, error output, etc?
This has been happening after the https://github.com/openshift/odo/pull/4007
And the reason is 
[...]

Issue #4023, opened 2020-09-23
Inconsistency in output of odo list


adisky:
/kind bug

## What versions of software are you using?

**Operating System:**

**Output of `odo version`:**
odo v2.0.0 (6fbb9d9bf)

## How did you run odo exactly?
```
$ odo list
Devfile Components: 
APP     NAME         PROJECT     TYPE                 STATE
app     nodejs     default     nodejs               Pushed
Openshift Components: 
APP     NAME      PROJECT     TYPE       SOURCETYPE     STATE
app     door2     default     nodejs     local          Pushed

$ odo list --path ~/
Devfile Components:
APP     NAME              PROJECT     STATE          CONTEXT
app     nodejs            default     Pushed         /home/adisky/check-odo-list/nodejs

S2I Components: 
APP     NAME                         PROJECT         TYPE       SOURCETYPE     STATE          CONTEXT
app     door2                        default         nodejs     local          Pushed         /home/adisky/check-nodejs-s2i/nodejs-ex
```
## Actual behavior
odo list displays as `Devfile components` & `Openshift Components` where as s2i component displays as `Devfile Components` and `S2I component`
odo list --path ~/ does not contain `type` column for devfiles.

## Expected behavior
consistent output for both commands.

PR #4177, opened 2020-10-30
remove duplicate list and refactor

girishramnani:
**What type of PR is this?**
/kind bug
/kind cleanup

- resolve `odo list` printing devfile components as s2i components
- refactor the code to not use `Kclient` directly in the cli
- refactor the code to not use `Deployment` struct ( structs from the lowest ) layer in the cli layer

Fixes https://github.com/openshift/odo/issues/4144

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 


girishramnani:
@dgolovin This might change the `odo list -o json` output, please confirm.

prietyc123:
@girishramnani unit test `TestList` for `List` function are failing due to component field mismatch. Could you please fix those.

girishramnani:
unit test should be resolved now

pkg/odo/cli/component/list.go
```diff
@@ -310,7 +304,7 @@ func (lo *ListOptions) Run() (err error) {
 			return
 		}
 	} else {
-		combinedComponents := component.GetMachineReadableFormatForCombinedCompList(components, devfileComponents)
+		combinedComponents := component.GetMachineReadableFormatForCombinedCompList(components, devfileComponents.Items)
```
dharmit, line 307: Variable names could be different to indicate s2i components and devfile components. Also, in one case we need the `.Items` whereas not in the other. Can we have uniformity?

pkg/component/types.go
```diff
@@ -48,22 +48,8 @@ type ComponentStatus struct {
...
-	Status            ComponentStatus      `json:"status,omitempty"`
-}
-
-type DevfileComponentSpec struct {
-	Name       string `json:"componentName,omitempty"`
-	App        string `json:"app,omitempty"`
-	Type       string `json:"type,omitempty"`
-	SourceType string `json:"sourceType,omitempty"`
+	S2IComponents     []Component `json:"s2iComponents"`
+	DevfileComponents []Component `json:"devfileComponents"`
```
dharmit, line 52: Thinking out loud. Can we have `Items` and make `S2IComponents` & `DevfileComponents` a part of it? I'm asking this because it'll help us have uniformity with the way k8s returns a list of items. This should also cascade to the way we return list of services, AFAIK.

pkg/component/component.go
```diff
@@ -937,6 +946,28 @@ func List(client *occlient.Client, applicationName string, localConfigInfo *conf
 	return compoList, nil
 }
 
+// List lists all components in active application
```
dharmit, line 949: `s/all/all s2i and devfile` would be more explanatory, IMO.

dharmit (changes requested):

pkg/component/component.go
```diff
@@ -856,8 +856,33 @@ func GetComponentNames(client *occlient.Client, applicationName string) ([]strin
 	return names, nil
 }
 
-// List lists components in active application
-func List(client *occlient.Client, applicationName string, localConfigInfo *config.LocalConfigInfo) (ComponentList, error) {
+func ListDevfileComponents(client *occlient.Client, selector string) (ComponentList, error) {
```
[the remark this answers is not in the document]
kadel: +1 especially description for selector. 
It looks a little bit weird to me as we don't use selectors for odo components.

pkg/component/component.go
```diff
@@ -937,6 +946,28 @@ func List(client *occlient.Client, applicationName string, localConfigInfo *conf
 	return compoList, nil
 }
 
+// List lists all components in active application
+func List(client *occlient.Client, applicationName string, localConfigInfo *config.LocalConfigInfo) (ComponentList, error) {
+	var applicationSelector string
+	if applicationName != "" {
+		applicationSelector = fmt.Sprintf("%s=%s", applabels.ApplicationLabel, applicationName)
```
kadel, line 953: Why not use  `applicationSelector = applabels.GetSelector(applicationName)` as is done in other places?

dharmit:
@girishramnani will this PR also fix https://github.com/openshift/odo/issues/4023? Not asking that it should, just wondering if it would.

girishramnani:
> @girishramnani will this PR also fix #4023? Not asking that it should, just wondering if it would.

actually that one has a bigger problem which we might not be able to solve right now

pkg/component/component.go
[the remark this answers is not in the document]
girishramnani: yes @kadel  - we use the project selector when we want to have components from all application, but in case of s2i components - we list all applications and then list per component application 
https://github.com/openshift/odo/blob/e5798c42c9ce191425e3ed18e52db985e1e2f920/pkg%2Fodo%2Fcli%2Fcomponent%2Flist.go#L254-L261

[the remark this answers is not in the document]
girishramnani: Now the S2I approach is not optimised as it Xn calls to the where n is the number of applications and X is the calls done per application.

pkg/component/component.go
```diff
@@ -937,6 +946,28 @@ func List(client *occlient.Client, applicationName string, localConfigInfo *conf
 	return compoList, nil
 }
 
+// List lists all components in active application
+func List(client *occlient.Client, applicationName string, localConfigInfo *config.LocalConfigInfo) (ComponentList, error) {
+	var applicationSelector string
+	if applicationName != "" {
```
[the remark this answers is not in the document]
girishramnani: I am not exactly sure if CLI prevents it but as `List` actually checks the local component as well - chances are that doesn't require application name

pkg/component/component.go
girishramnani: done

pkg/component/component.go
girishramnani: done

pkg/component/types.go
girishramnani: I am thinking as part of refactor we might consolidate these two list with a new label called `IsLegacy` once devfile matures further.

pkg/component/component.go
```diff
@@ -1541,6 +1571,12 @@ func GetLogs(client *occlient.Client, componentName string, applicationName stri
 }
 
 func getMachineReadableFormat(componentName, componentType string) Component {
+	cmp := NewComponent(componentName)
+	cmp.Spec.Type = componentType
+	return cmp
+}
+
+func NewComponent(componentName string) Component {
```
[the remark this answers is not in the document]
girishramnani: done

pkg/odo/cli/component/list.go
girishramnani: variable names are different to indicate s2i component and devfile components already, one is called `components` and other is `devfileComponents`. Am I missing something? also 

> Also, in one case we need the `.Items` whereas not in the other. Can we have uniformity?

How can I have uniformity? use the Items for everywhere?

girishramnani: resolved the `Also..` portion

dharmit: > variable names are different to indicate s2i component and devfile components already, one is called components and other is devfileComponents. Am I missing something?

My bad. Should have made it clear that I was requesting something like `s/components/s2iComponents` for the variable to store s2i components.

pkg/component/types.go
dharmit: So you mean that we can have just one list called `Items` and then in the individual items, there could be a bool label `IsLegacy`  which indicates that it's an s2i component, right? Sounds good to me. Or we could label it `IsS2i`. But that's a separate discussion for when we refactor it.

dharmit:
@girishramnani Multiple test cases failing in unit test

pkg/component/types.go
girishramnani: yeah

pkg/odo/cli/component/list.go
girishramnani: resolved

dharmit (approved):
/lgtm

PR #4207, opened 2020-11-10
Use example from the repo instead of downloading it

### LINKED ISSUE #3020: External dependencies in integrations test
The integrations test should not have external dependency unless it is absolutely unavoidable.
It creates unnecessary external dependencies that lead to flaky tests.
```
        util_test.go:1464: Error downloading zip: Get https://github.com/che-samples/web-nodejs-sample/archive/master.zip: dial tcp: lookup github.com on [::1]:53: read udp [::1]:52934->[::1]:53: read: connection refused
```

The example app should be stored with tests, and then copied into a temp directory using `helper.CopyExample`

These are some of the test scenarios that are unnecessary cloning GitHub repos.

https://github.com/openshift/odo/blob/10826cc3f43c74ebadd917c3c53e9b40d6853aa6/tests/integration/devfile/utils/utils.go#L38
https://github.com/openshift/odo/blob/10826cc3f43c74ebadd917c3c53e9b40d6853aa6/tests/integration/devfile/docker/cmd_docker_devfile_url_test.go#L48
https://github.com/openshift/odo/blob/2086305a0caaaf603b90a4fb9139ebb0453f3a10/tests/integration/devfile/cmd_devfile_url_test.go#L53
https://github.com/openshift/odo/blob/12801e01b47dc5b4a08e3b30b346417540e5b385/tests/integration/devfile/cmd_devfile_delete_test.go#L51

/assign @maysunfaisal 
/assign @yangcao77

/area testing

dharmit:
**What type of PR is this?**
/kind cleanup

We need this to remove external dependency of downloading the code from GitHub each time in the test

Fixes #3020

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Tests should pass.


girishramnani:
Sub modules are hard to maintain, are you sure about it?

girishramnani:
We would need to add a git submodule init & sync step in the readme

dharmit:
> Sub modules are hard to maintain, are you sure about it?

I'm not. That's why I am seeking thoughts from others. 

> We would need to add a git submodule init & sync step in the readme

I think it's more like we'll have to sync if there's any change in the submodule repo that we want to use on our side. But I get what you're suggesting. It's not as straightforward as, maybe, cloning the code somewhere and doing a `cp`.

adisky:
We are storing some external dependencies here https://github.com/openshift/odo/tree/master/tests/examples/source/devfiles/nodejs

We can add this one too, but i am not sure how maintainable it is store all dependencies in odo.

adisky:
This issue is reported before we started caching devfile registry, then it could be due to hitting github API rate limits.
have we seen this issue recently?

kadel:
> Sub modules are hard to maintain, are you sure about it?

I don't think that there is any reason why it needs to be a submodule. We can simply copy the samples we are already doing that it some examples.

dharmit:
It looks like we're unanimously sceptical of using submodule. :smile: 
I've removed the submodule thingy and simply copied the code for the repo.

tests/integration/devfile/cmd_devfile_log_test.go
```diff
@@ -56,8 +57,7 @@ var _ = Describe("odo devfile log command tests", func() {
 
 			projectDir := filepath.Join(commonVar.Context, "projectDir")
 
-			helper.CmdShouldPass("git", "clone", "https://github.com/che-samples/web-nodejs-sample.git", projectDir)
-
+			helper.CopyExample(filepath.Join("source", "web-nodejs-sample"), projectDir)
```
mik-dass, line 60: We can use the current nodejs sample directory here. It seems to have worked for me locally.

mohammedzee1000:
It might also be a good idea to add a script that updates the examples from its source, for when we need to.

tests/integration/devfile/cmd_devfile_log_test.go
dharmit: Correct me if I'm wrong but I think things would work with many nodejs projects. Besides, I don't mind testing a different example in our tests as compared to testing only one example project - for example, I'd prefer we test something beyond nodejs+etcd operator for Operator Hub tests at some point.

dharmit:
> It might also be a good idea to add a script that updates the examples from its source, for when we need to.

That would be over engineering things here, IMHO.

tests/integration/devfile/cmd_devfile_log_test.go
mik-dass: OK let's test against a new example

mik-dass (approved):
/lgtm

Issue #4093, opened 2020-10-07
Refactor occlient and kclient by unifying them into one


adisky:
/kind code-refactoring

## Acceptance Criteria 
- ~oc client and kclient should have the same interface and based on what odo encounters in the context, the client needs to be decided in the context.~ 
- [ ] Unify the `occlient` and `kclient` into one client called as `kclient`. Remove the duplications and do everything through `kclient`. Interfaces to be added later.

girishramnani:
We need to primarily bring back the original 3 layered architecture that odo had -
- cli packages
- odo specific packages
- cluster communication packages

Currently things are bleading into each other.

mik-dass:
It might be good idea to merge the two clients into a single client which handles both openshift and kubenetes clusters. We already are using the kclient inside the occlient
 https://github.com/openshift/odo/blob/49853dc69d4f45d2f54cd4bad80e601bd2b4e715/pkg/occlient/occlient.go#L200

This will help reduce duplicate code between the two clients. Also a single interface  for both the clients might be tricky as certain resources like routes etc are not in the kclient.

dharmit:
> It might be good idea to merge the two clients into a single client

I remember that we started `kclient` with the intention of being able to deprecate `occlient` at some point. So this *has* to be our end goal, IMO.

mik-dass:
> I remember that we started kclient with the intention of being able to deprecate occlient at some point. So this has to be our end goal, IMO.

We will just remove the unnecessary functions when that deprecation happens from the new merged client. Also creation of routes instead of ingress does need a occlient currently and we need to support s2i for some time.

metacosm:
The two clients should be indeed be merged. It shouldn't matter to the client code which kind of cluster odo is connected to. I don't think that this can be done without first isolating how components / etc. are realized on the cluster from code calling it: i.e. code calling the client shouldn't need to know how a component is implemented on the cluster. More precisely, the client should expose something like `GetComponent *Component` and not `GetDeployment *v1.Deployment`.

mik-dass:
> More precisely, the client should expose something like GetComponent *Component and not GetDeployment *v1.Deployment.

We were thinking of doing this on the odo specific packages. The current idea is to have interfaces for storage, URL etc and s2i/devfile  will have their own implementations of those functions. The client won't have any idea about odo specific things.

mik-dass:
[...]

PR #4211, opened 2020-11-11
Moves project related code from occlient.go to a new file

mik-dass:
**What type of PR is this?**

/kind cleanup
/kind code-refactoring

Moves project related code from occlient.go to projects.go

Fixes part of #4093 

**PR acceptance criteria**:

- [X] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Tests should pass

### DIFF
--- a/pkg/occlient/projects.go
+++ b/pkg/occlient/projects.go
@@ -15,8 +15,8 @@ import (
 	"k8s.io/klog"
 )
 
-// GetProject returns project based on the name of the project.Errors related to
-// project not being found or forbidden are translated to nil project for compatibility
+// GetProject returns project based on the name of the project
+// errors related to project not being found or forbidden are translated to nil project for compatibility
 func (c *Client) GetProject(projectName string) (*projectv1.Project, error) {
 	prj, err := c.projectClient.Projects().Get(projectName, metav1.GetOptions{})
 	if err != nil {
@@ -40,11 +40,11 @@ func (c *Client) ListProjects() (*projectv1.ProjectList, error) {
 	return c.projectClient.Projects().List(metav1.ListOptions{})
 }
 
-// GetProjectNames return list of existing project names that user has access to.
-func (c *Client) GetProjectNames() ([]string, error) {
+// ListProjectNames return list of existing project names that user has access to.
+func (c *Client) ListProjectNames() ([]string, error) {
 	projects, err := c.ListProjects()
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list projects")
+		return nil, err
 	}
 
 	var projectNames []string
@@ -156,7 +156,7 @@ func (c *Client) DeleteProject(name string, wait bool) error {
 // CreateNewProject creates project with given projectName
 func (c *Client) CreateNewProject(projectName string, wait bool) error {
 	// Instantiate watcher before requesting new project
-	// If watched is created after the project it can lead to situation when the project is created before the watcher.
+	// If watcher is created after the project it can lead to situation when the project is created before the watcher.
 	// When this happens, it gets stuck waiting for event that already happened.
 	var watcher watch.Interface
 	var err error
@@ -220,7 +220,7 @@ func (c *Client) GetCurrentProjectName() string {
 func (c *Client) SetCurrentProject(projectName string) error {
 	rawConfig, err := c.KubeConfig.RawConfig()
 	if err != nil {
-		return errors.Wrapf(err, "unable to switch to %s project", projectName)
+		return err
 	}
 
 	rawConfig.Contexts[rawConfig.CurrentContext].Namespace = projectName
--- a/pkg/occlient/projects.go
+++ b/pkg/occlient/projects.go
@@ -1,0 +1,236 @@
+package occlient
+
+import (
+	"fmt"
+	"time"
+
+	projectv1 "github.com/openshift/api/project/v1"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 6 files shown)


pkg/occlient/projects.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// IsProjectSupported checks if Project resource type is present on the cluster
+func (c *Client) IsProjectSupported() (bool, error) {
+	return c.isResourceSupported("project.openshift.io", "v1", "projects")
+}
+
+// GetCurrentProjectName returns the current project name
+func (c *Client) GetCurrentProjectName() string {
+	return c.Namespace
```
dharmit, line 216: Do we really need a function for this? Is there some design pattern we're following here?

pkg/occlient/projects.go
```diff
@@ -0,0 +1,236 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+)
+
+// GetProject returns project based on the name of the project.Errors related to
+// project not being found or forbidden are translated to nil project for compatibility
```
dharmit, line 19: Is it just me or is it really unclear/confusing as to what this function does based on the comment?

pkg/occlient/projects.go
```diff
@@ -0,0 +1,236 @@
...
+// ListProjects return list of existing projects that user has access to.
+func (c *Client) ListProjects() (*projectv1.ProjectList, error) {
+	return c.projectClient.Projects().List(metav1.ListOptions{})
+}
+
+// GetProjectNames return list of existing project names that user has access to.
+func (c *Client) GetProjectNames() ([]string, error) {
+	projects, err := c.ListProjects()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to list projects")
```
dharmit, line 47: Similar to what I requested on https://github.com/openshift/odo/pull/4213#discussion_r524912708. Can we not wrap and simply throw the error that `ListProjects` encountered? Also, is there a design pattern we're following here?

pkg/occlient/projects.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	// Return nil since we don't bother checking for the watcher..
+	return nil
+}
+
+// CreateNewProject creates project with given projectName
+func (c *Client) CreateNewProject(projectName string, wait bool) error {
+	// Instantiate watcher before requesting new project
+	// If watched is created after the project it can lead to situation when the project is created before the watcher.
```
dharmit, line 159: `s/If watched is created/If watcher is created/` maybe?

pkg/occlient/projects.go
```diff
@@ -0,0 +1,236 @@
...
+// GetCurrentProjectName returns the current project name
+func (c *Client) GetCurrentProjectName() string {
+	return c.Namespace
+}
+
+// SetCurrentProject sets the given projectName to current project
+func (c *Client) SetCurrentProject(projectName string) error {
+	rawConfig, err := c.KubeConfig.RawConfig()
+	if err != nil {
+		return errors.Wrapf(err, "unable to switch to %s project", projectName)
```
dharmit, line 223: We're not really trying to switch to a project specified by `projectName` in this step. It's kinda misleading to report so at this step, no? Can we change the error message to something that's more relevant for this step?

pkg/occlient/projects_test.go
```diff
@@ -0,0 +1,189 @@
...
+			name:     "Case 2: valid project name, waiting",
+			projName: "testing2",
+			wait:     true,
+			wantErr:  false,
+		},
+		// {
+		// 	name:     "Case 2: empty project name",
+		// 	projName: "",
+		// 	wantErr:  true,
+		// },
```
dharmit, line 40: Can we remove the commented code?

dharmit (changes requested):

pkg/occlient/projects.go
mik-dass: I copied it from the original code but I feel it helps in hiding the implementation details.

pkg/occlient/projects.go
mik-dass: fixed

pkg/occlient/projects.go
mik-dass: fixed

pkg/occlient/projects.go
mik-dass: Fixed and returning the complete error instead now.

pkg/occlient/projects_test.go
mik-dass: Fixed

pkg/occlient/projects.go
mik-dass: Fixed

pkg/occlient/projects.go
dharmit: Agree with your reasoning. It's just that I don't really find it very valuable. More so because it's being called at only two places in code and one place in tests:
```sh
$ grep -r "GetCurrentProjectName" --exclude-dir=vendor
Binary file odo matches
pkg/occlient/occlient.go:func (c *Client) GetCurrentProjectName() string {
pkg/occlient/occlient.go:	currentProjectName := c.GetCurrentProjectName()
pkg/occlient/occlient_test.go:				t.Errorf("GetImageStream() = %#v, want %#v and the current project name is %s\n\n", got, tt, fkclient.GetCurrentProjectName())
pkg/catalog/catalog.go:	currentNamespace := client.GetCurrentProjectName()
```

mik-dass: But it might a good idea not to leak the logic of fetching the current project in the layers above like catalog.go. The client should be responsible and should hide it away. Someday if the logic changes, we won't have to change it at the other places.

dharmit: Makes sense. :+1:

dharmit (approved):
@mik-dass thanks for answering/fixing the questions/suggestions in PR feedback. :+1: 

Going to approve & lgtm since it was my review that removed lgtm given by @adisky 

/approve
/lgtm

PR #4246, opened 2020-11-20
Update

boczkowska:
**What type of PR is this?**
/kind bug

Updates the script.

We changed the structure of downstream docs where some of the synced docs ended up in their own subdirectory. Tbh, I've never scripted before so I just added a loop to iterate through that subdirectory when converting files :D 

BUT, it still won't work bc the downstream subdirectory is missing a symlink (no idea how that happened and why the docs build). Fixed here: https://github.com/openshift/openshift-docs/pull/27551 I've pinged the OCP team to merge it. 
Should work when the PR above is merged otherwise you'll end up with a bunch of `unresolved directive` in docs.

Sync w/ upstream is fine.

The script doesn't work.
Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


scripts/sync.sh
```diff
@@ -100,8 +100,8 @@ git clone https://$openshift_docs_repo $tmp_openshift_docs_dir
 dir=$tmp_openshift_docs_dir$odo_doc_directory
 
 # Convert all the documentation to markdown from the OpenShift upstream repository
-for f in $dir/*.adoc; do
-  convert_to_markdown $f $dir
+for f in $dir/*.adoc $dir/creating_and_deploying_applications_with_odo/*.adoc; do
```
girishramnani, line 103: can we consider `$dir/**/*.adoc`?

girishramnani: actually 
```
find $dir -name '*.adoc' | xargs convert_to_markdown
```
should allow you to just remove the for loop?

mohammedzee1000: It should work

mohammedzee1000: Ah , but there is one problem, he is using an inbuilt function, not a command that is in the path. Another issue the function is not multi argh.

Basically, the only way this would work is by exporting the function. Infact, there will be more lines than now. Although you might get an advantage in can you can execute in parallel.

```bash
❯ cat readf.sh
readfile() {
	echo $1;
	cat $1;
}
export -f readfile
echo "finding files"
find dir -name '*.adoc' | xargs -d' ' -t -n1 -P2 bash -c 'readfile "$0"

❯ sh readf.sh
finding files
bash -c 'readfile "$0"' 'dir/f1.adoc'$'\n''dir/f3.adoc'$'\n''dir/f5.adoc'$'\n''dir/f4.adoc'$'\n''dir/f2.adoc'$'\n' 
dir/f1.adoc dir/f3.adoc dir/f5.adoc dir/f4.adoc dir/f2.adoc
hello f1.adoc

```

dharmit: @girishramnani 

> can we consider $dir/**/*.adoc?

I'll open a separate PR for it.

@mohammedzee1000 I'd request you to open a PR if the change you are proposing doesn't break anything with doc sync setup by Charlie.

dharmit (approved):
Played with this locally and it works.

/approve
/lgtm

Issue #3754, opened 2020-08-15
Push fails to sync files to the component


amitkrout:
/kind bug
/kind flake

## What versions of software are you using?

**Operating System:**
cluster version 
**Output of `odo version`:**
master
## How did you run odo exactly?
These are CI flakes coming from scenarios of debug test and cmp app

```
helper.CopyExample(filepath.Join("source", "nodejs"), context)
helper.CmdShouldPass("odo", "component", "create", "nodejs:latest", "nodejs-cmp-"+project, "--project", project, "--context", context)
helper.CmdShouldPass("odo", "push", "--context", context)
```

## Actual behavior
Failing

## Expected behavior
Push should go through

## Any logs, error output, etc?
```
 ------------------------------
Created dir: /tmp/038521211
Creating a new project: ftarejbpuw
Running odo with args [odo project create ftarejbpuw -w -v4]
[odo] I0815 13:12:24.345170   20768 util.go:734] HTTPGetRequest: https://raw.githubusercontent.com/openshift/odo/master/build/VERSION
[odo] I0815 13:12:24.345368   20768 util.go:755] Response will be cached in /tmp/odohttpcache for 1h0m0s
[odo] I0815 13:12:24.345683   20768 util.go:768] Cached response used.
[odo]  •  Waiting for project to come up  ...
[odo] I0815 13:12:26.259575   20768 occlient.go:542] Status of creation of project ftarejbpuw is Active
[odo] I0815 13:12:26.259598   20768 occlient.go:547] Project ftarejbpuw now exists
[odo] I0815 13:12:26.285186   20768 namespace.go:181] Status of creation of service account &ServiceAccount{ObjectMeta:{default  ftarejbpuw /api/v1/namespaces/ftarejbpuw/serviceaccounts/default b8923498-505e-4754-a97a-e42d027eb69d 45616 0 2020-08-15 13:12:24 +0000 UTC <nil> <nil> map[] map[] [] []  []},Secrets:[]ObjectReference{ObjectReference{Kind:,Namespace:,Name:default-dockercfg-lltz5,UID:,APIVersion:,ResourceVersion:,FieldPath:,},ObjectReference{Kind:,Namespace:,Name:default-token-hwj72,UID:,APIVersion:,ResourceVersion:,FieldPath:,},},ImagePullSecrets:[]LocalObjectReference{LocalObjectReference{Name:default-dockercfg-lltz5,},},AutomountServiceAccountToken:nil,} is ready
[odo] 
[...]

Issue #3256, opened 2020-05-28
Timeout while waiting for component deployment roll out


prietyc123:
/kind bug

## What versions of software are you using?

**Operating System:**
All Supported

**Output of `odo version`:**
master

## How did you run odo exactly?

Running test on OpenShift CI.

## Actual behavior

```
Creating a new project: ccupvnlhzv
Running odo with args [odo project create ccupvnlhzv -w -v4]
[odo]  •  Waiting for project to come up  ...
[odo] I0527 11:06:20.874584   26071 occlient.go:542] Status of creation of project ccupvnlhzv is Active
[odo] I0527 11:06:20.874675   26071 occlient.go:547] Project ccupvnlhzv now exists
[odo] 
[...]
Running odo with args [odo preference set Experimental true]
[odo] Global preference was successfully updated
[odo] I0527 11:06:21.164827   26107 preference.go:188] The path for preference file is /tmp/023406818/config.yaml
[odo] I0527 11:06:21.165321   26107 odo.go:72] Could not get the latest release information in time. Never mind, exiting gracefully :)
Running odo with args [odo create nodejs --project ccupvnlhzv veqzkx]
[odo] I0527 11:06:21.363048   26123 preference.go:188] The path for preference file is /tmp/023406818/config.yaml
[odo] I0527 11:06:21.363239   26123 util.go:399] path ./.odo/config.yaml doesn't exist, skipping it
[odo] I0527 11:06:21.363248   26123 preference.go:188] The path for preference file is /tmp/023406818/config.yaml
[odo] I0527 11:06:21.368382   26123 preference.go:188] The path for preference file is /tmp/023406818/config.yaml
[odo] Experimental mode is enabled, use at your own risk
[odo] 
[odo] I0527 11:06:21.834436   26123 util.go:399] path devfile.yaml doesn't exist, skipping it
[odo] I0527 11:06:21.834485   26123 util.go:399] path devfile.yaml doesn't exist, skipping it
[odo] Validation
[odo]  •  Checking devfile compatibility  ...
[odo]  •  Creating a devfile component from registry: DefaultDevfileRegistry  ...
[odo] 
 ✓  Checking devfile compatibility [68177ns]
[odo] 
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [60910ns]
[odo] I0527 11:06:21.834600   26123 preference.go:188] The path for preference file is /tmp/023406818/config.yaml
[...]

PR #4247, opened 2020-11-23
Handling absolute and relative source path file uri 

### LINKED ISSUE #3350: json match throws error due to source attribute mismatch on windows terminal
/kind bug

**Operating System:**
win10

**Output of `odo version`:**
master

Ran `make test-cmd-cmp` on windows.

## Actual behavior

```
SSSSSCreated dir: C:\Users\Admin\AppData\Local\Temp\209745756
Creating a new project: xgvocbsngm
Running odo.exe with args [odo project create xgvocbsngm -w -v4]
[odo]  -  Waiting for project to come up  ...
[odo] I0615 05:53:34.877438    8988 occlient.go:543] Status of creation of project xgvocbsngm is Active
[odo] I0615 05:53:34.923391    8988 occlient.go:548] Project xgvocbsngm now exists
[odo] I0615 05:53:34.953368    8988 occlient.go:583] Status of creation of service account &ServiceAccount{ObjectMeta:{default  xgvocbsngm /api/v1/namespaces/xgvocbsngm/serviceaccounts/default 493ec3a9-5db9-4eae-ab6f-794c313ad5ed 42582 0 2020-06-15 05:53:35 +0000 GMT <nil> <nil> map[] map[] [] []  []},Secrets:[]ObjectReference{ObjectReference{Kind:,Namespace:,Name:default-token-mv29v,UID:,APIVersion:,ResourceVersion:,FieldPath:,},ObjectReference{Kind:,Namespace:,Name:default-dockercfg-dddjj,UID:,APIVersion:,ResourceVersion:,FieldPath:,},},ImagePullSecrets:[]LocalObjectReference{LocalObjectReference{Name:default-dockercfg-dddjj,},},AutomountServiceAccountToken:nil,} is ready
 V  Waiting for project to come up [1s]
[odo]  V  Project 'xgvocbsngm' is ready for use
[odo]  V  New project created and now using project: xgvocbsngm
[odo] I0615 05:53:34.974973    8988 odo.go:72] Could not get the latest release information in time. Never mind, exiting gracefully :)
Current working dir: C:\Users\Admin\go\src\github.com\openshift\odo\tests\integration
Setting current dir to: C:\Users\Admin\AppData\Local\Temp\209745756
Running odo.exe with args [odo component create nodejs cmp-git --project xgvocbsngm --context C:\Users\Admin\AppData\Local\Temp\209745756 --app testing -o json]
[odo] {
[odo]     "kind": "Component",
[odo]     "apiVersion": "odo.dev/v1alpha1",
[odo]     "metadata": {
[odo]         "name": "cmp-git",
[odo]         "namespace": "xgvocbsngm",
[odo]         "creationTimestamp": null
[odo]     },
[odo]     "spec": {
[odo]         "app": "testing",
[odo]         "type": "nodejs",
[odo]         "source": "file:///./",
[odo]         "sourceType": "local",
[odo]         "ports": [
[odo]             "8080/TCP"
[odo]         ]
[odo]     },
[odo]     "status": {
[odo]         "state": "Not Pushed"
[odo]     }
[odo] }
Deleting project: xgvocbsngm
Running odo.exe with args [odo project delete xgvocbsngm -f]
[odo]  V  Deleted project : xgvocbsngm
[odo]  !  Warning! Projects are deleted from the cluster asynchronously. Odo does its best to delete the project. Due to multi-tenant clusters, the project 
may still exist on a different node.
Setting current dir to: C:\Users\Admin\go\src\github.com\openshift\odo\tests\integration
Deleting dir: C:\Users\Admin\AppData\Local\Temp\209745756

+ Failure [2.093 seconds]
odo sub component command tests
C:/Users/Admin/go/src/github.com/openshift/odo/tests/integration/cmd_cmp_sub_test.go:13
  Creating component
  C:/Users/Admin/go/src/github.com/openshift/odo/tests/integration/component.go:52
    should describe not pushed component when it is created with json output [It]
    C:/Users/Admin/go/src/github.com/openshift/odo/tests/integration/component.go:262

    Expected
        <string>: "...le:///./","..."
    to equal               |
        <string>: "...le://./","s..."

    C:/Users/Admin/go/src/github.com/openshift/odo/tests/integration/component.go:268
```

## Expected behavior

json output should match all the attribute

Manually verified on windows.

prietyc123:
**What type of PR is this?**

/kind bug
/kind failing-test

Fixing json mismatch on windows platform. 

Fixes #3350 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

`make test-cmd-cmp` and `make test-cmd-cmp-sub` should pass on windows


kadel:
I don't think that this is the correct solution. This i
This is a real bug in odo's `GenFileURL`  function. The file URI format should not depend on OS.
/hold

tests/helper/helper_generic.go
```diff
@@ -311,3 +312,14 @@ func CommonAfterEach(commonVar CommonVar) {
...
 }
+
+// GenFileURL Converts file path on windows to /C:/path/to/file to work in URL
+func GenFileURL() string {
+	var sourcePath string
+	if runtime.GOOS == "windows" {
+		sourcePath = "file:///./"
+		return sourcePath
+	}
+	sourcePath = "file://./"
```
kadel, line 323: this is not the correct local URI see https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

prietyc123: > this is not the correct local URI see https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

@kadel Yes, agreed. I have analysed this thoroughly and commented out all of my observation here https://github.com/openshift/odo/issues/3350#issuecomment-735639702 . Also I have modified the PR accordingly, so cancelling the hold.

/hold cancel

pkg/util/util.go
```diff
@@ -209,9 +209,15 @@ func GenFileURL(location string, os ...string) string {
 	}
 	urlPath := location
 	if currOS == WIN {
-		urlPath = "/" + strings.Replace(location, "\\", "/", -1)
+		urlPath = strings.Replace(location, "\\", "/", -1)
 	}
-	return "file://" + urlPath
+	if filepath.IsAbs(urlPath) || strings.Contains(urlPath, ":/") {
+		if currOS == WIN {
```
kadel, line 215: This still doesn't look right. 
There is no reason why URI should be different based on OS.

prietyc123: > There is no reason why URI should be different based on OS

TBH I didn't get your point, AFAIU we are using `file:///` URI across all the platform and yes we are adding `/` as a prefix on windows for a special case (when we have the absolute path from the root of different drive, I mean other than the current drive). In this special scenario the path would be like `C:\path\to\the\file`. However if we consider path from the root of current drive on windows or the absolute path on linux and macOS, it will always start with `/` for ex `/c/path/to/file` and hence combining `file://` with the path gives you the format `file:///`. Detailed absolute path description for windows has been described https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats . 
One more interesting fact for the special case is you won't be able to determine the path is absolute or not using `filepath.IsAbs` because it only checks the path has prefix `/`. 

> If we don't want to use URI format anymore, it is fine, but the output format needs to be consistent across all Operating Systems.

Well, I am not sure whether we should remove it or not because till now AFAIK we are using relative path for the json source field and I don't find any use case for the absolute. But if we have upcoming plans to use absolute in any case then keeping this would be helpful.

prietyc123:
The failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4247/pull-ci-openshift-odo-master-v4.6-integration-e2e/1334096274391044096#1:build-log.txt%3A1203 has been tracked in https://github.com/openshift/odo/issues/3256

/retest

tests/integration/component.go
```diff
@@ -256,7 +256,7 @@ func componentTests(args ...string) {
 			helper.CopyExample(filepath.Join("source", "nodejs"), commonVar.Context)
 			cmpDescribeJSON, err := helper.Unindented(helper.CmdShouldPass("odo", append(args, "create", "--s2i", "nodejs", "cmp-git", "--project", commonVar.Project, "--context", commonVar.Context, "--app", "testing", "-o", "json")...))
 			Expect(err).Should(BeNil())
-			expected, err := helper.Unindented(`{"kind": "Component","apiVersion": "odo.dev/v1alpha1","metadata": {"name": "cmp-git","namespace": "` + commonVar.Project + `","creationTimestamp": null},"spec":{"app": "testing","type":"nodejs","source": "file://./","sourceType": "local","ports": ["8080/TCP"]}, "status": {"state": "Not Pushed"}}`)
+			expected, err := helper.Unindented(`{"kind": "Component","apiVersion": "odo.dev/v1alpha1","metadata": {"name": "cmp-git","namespace": "` + commonVar.Project + `","creationTimestamp": null},"spec":{"app": "testing","type":"nodejs","source": "./","sourceType": "local","ports": ["8080/TCP"]}, "status": {"state": "Not Pushed"}}`)
```
kadel, line 259: Why is this changing `source` path format? Now it is not URI, but just path

pkg/util/util.go
kadel: You current code prints different output for win and Linux.
On windows path will be with `file://` on Linux it won't


All we need to to do fix the original issue is to make sure that the path aways starts with `file:///` (3 slashes)

kadel:
After discussing this with @prietyc123 we decided that we will go with just the plan path and stop using URL format for the local path.
There are multiple problems with using URL. For example, the URL can't have a relative path. So something like `file:///./` is actually invalid.

Please not that this will change JSON output a little bit.

prietyc123:
Added `release-note` label to be considered at the time of release as it might change the behaviour of JSON output.

prietyc123:
Failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4247/pull-ci-openshift-odo-master-v4.6-integration-e2e/1335931135380688896#1:build-log.txt%3A522 has been tracked in https://github.com/openshift/odo/issues/3754

/retest

tests/integration/component.go
mohammedzee1000: Agreed, this is a file path now

prietyc123: This will be a file path from now onwards as explained in comment https://github.com/openshift/odo/issues/3350#issuecomment-735639702 and discussed with Tomas. Me and Tomas has come up with an unanimous decision on this https://github.com/openshift/odo/pull/4247#issuecomment-738780434

mohammedzee1000 (approved):
looks good
/lgtm

PR #4202, opened 2020-11-06
List only installed Operators; some refactor

dharmit:
**What type of PR is this?**
/kind bug
/kind code-refactoring

**What does does this PR do / why we need it**:
This PR makes sure that odo lists only those Operators that are successfully installed. This success is determined on the basis of the `Phase` of Operator. If the phase evaluates to `Succeeded`, the Operator is considered available to use.

**Which issue(s) this PR fixes**:

Fixes #4155 & https://bugzilla.redhat.com/show_bug.cgi?id=1889961

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:
This can be better tested on a minikube setup.
```sh
$ minikube start
$ minikube addons enable olm

# make sure you're working on default namespace
$ kubectl apply -f https://gist.github.com/dharmit/89079ad3daa7df83b92cc39c499f670f/raw/ea8808fcc6ec876b1c7c1f0838295f21324f3baf/operators.yaml

# Check the PHASE column in output. Give it a few seconds or about a minute for etcd Operator to show up
$ kubectl get csv   
NAME                                  DISPLAY                    VERSION             REPLACES                              PHASE
etcdoperator.v0.9.4-clusterwide       etcd                       0.9.4-clusterwide                                         Succeeded
service-binding-operator.v0.1.1-364   Service Binding Operator   0.1.1-364           service-binding-operator.v0.1.1-354   Failed

# odo doesn't show service binding operator since it failed to install
$ odo catalog list services
Operators available in the cluster
NAME                                CRDs
etcdoperator.v0.9.4-clusterwide     EtcdCluster, EtcdBackup, EtcdRestore
```

PR #4248, opened 2020-11-23
Increase timeout for Operator Hub tests

dharmit:
**What type of PR is this?**

/kind flake

This increased timeout makes sure that `odo catalog list services` being
executed in `JustBeforeEach` gets enough time to check for existence of
two Operators that we're installing in CI to perform our checks.

We need to increase this timeout because it's taking a long time for
`odo catalog list services` to find "successfully installed" Operators
in the newly created namespace even if these Operators are installed at
cluster level, i.e., in the `openshift-operators` namespace.

The increase in the time taken by odo to find this info is mainly
attributed to https://github.com/openshift/odo/pull/4202. I am observing
that even `kubectl get csv` is taking more than usual time to list the
Operators or report their phase as "Succeeded" but there's not enough
data to back that up. So, in the interim, this seems like best way to
make sure that CI doesn't keep failing on `JustBeforeEach`.

Fixes #? NA

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/integration/operatorhub/cmd_service_test.go
+++ b/tests/integration/operatorhub/cmd_service_test.go
@@ -30,10 +30,10 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
 		helper.CmdShouldPass("odo", "project", "set", project)
 
 		// wait till oc can see the all operators installed by setup script in the namespace
-		ocArgs := []string{"catalog", "list", "services"}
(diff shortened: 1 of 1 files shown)


mohammedzee1000:
Hmm I was going to complain about this

kadel:
small fix 

/approve
/lgtm

PR #4252, opened 2020-11-24
add aditi as approver in top level OWNERS file

girishramnani:
**What type of PR is this?**

Adding aditi on top level OWNERS file because we need more approvers.

No issues


dharmit (approved):
/approve
/lgtm

PR #4292, opened 2020-12-07
Configure-installer script refactor part 2

mohammedzee1000:
/kind code-refactor
This is the second phase of the refactor of the configure the cluster. Again. Nothing the flow should be broken

### DIFF
--- a/scripts/configure-cluster/common/auth.sh
+++ b/scripts/configure-cluster/common/auth.sh
@@ -3,7 +3,6 @@
 SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 HTPASSWD_FILE="${SCRIPTPATH}/test-htpasswd"
 HTPASSWD_SECRET="htpasswd-secret"
-USERPASS="password@123"
 
 createclustersecret() {
     # Create secret in cluster, removing if it already exists
@@ -42,7 +41,7 @@ waitforstablelogin() {
     # Login as developer and check for stable server
     for i in {1..40}; do
         # Try logging in as developer
-        oc login -u developer -p $USERPASS &> /dev/null
+        oc login -u developer -p password@123 &> /dev/null
         if [ $? -eq 0 ]; then
             # If login succeeds, assume success
     	    OC_STABLE_LOGIN="true"
--- a/scripts/configure-cluster/common/auth.sh
+++ b/scripts/configure-cluster/common/auth.sh
@@ -1,29 +1,9 @@
 #!/usr/bin/env bash
 
-set -e
-
-HTPASSWD_FILE="./htpass"
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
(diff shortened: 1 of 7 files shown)


scripts/configure-cluster/common/kubeconfigreset.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+reset_kubeconfig() {
```
girishramnani, line 3: this can be in the kubeconfigandadmin.sh itself - setup and reset together in the same file

girishramnani:
small nit: rest all is good

scripts/configure-cluster/common/auth.sh
```diff
@@ -1,30 +1,10 @@
 #!/usr/bin/env bash
 
-set -e
-
-HTPASSWD_FILE="./htpass"
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+HTPASSWD_FILE="${SCRIPTPATH}/test-htpasswd"
 HTPASSWD_SECRET="htpasswd-secret"
```
girishramnani, line 5: are these used now - I see a file for passwords already created

scripts/configure-cluster/common/auth.sh
```diff
@@ -1,30 +1,10 @@
...
 USERPASS="password@123"
```
girishramnani, line 6: same here - is this needed?

scripts/configure-cluster/common/auth.sh
mohammedzee1000: Yes, they are used. It points to the pre-existing htpassword file.

scripts/configure-cluster/common/auth.sh
mohammedzee1000: Ok as its used only in one location, removing for now

scripts/configure-cluster/common/auth.sh
girishramnani: the `HTPASSWD_SECRET` is pointing to the file? because then this variable name is super confusing

scripts/configure-cluster/common/kubeconfigreset.sh
mohammedzee1000: Ideally yes, but since it exports envs, it needs to be called with . or source as in `. kubeconfigandadmin.sh`. This can result in the automatic execution of reset_kubeconfig as well. Hence why it is in a separate file.

scripts/configure-cluster/common/auth.sh
mohammedzee1000: HTPASSWD_SECRET is the name of the secret that is created on the cluster using the `HTPASSWD_FILE`. We then create a htpasswd auth provider that is linked to this secret i.e told to use it.

mohammedzee1000:
anything else @girishramnani ?

girishramnani:
Looks good 
/lgtm

girishramnani:
\lgtm

girishramnani:
What am I doing wrong - why isn’t it lgtming?

PR #4307, opened 2020-12-14
Fix odo hang when cluster is unreachable 

### LINKED ISSUE #4046: odo 'hangs' after using a cluster that's no longer accessible
/kind bug

**Operating System:**
MacBook Pro with Catalina

**Output of `odo version`:**
odo v2.0.0 (6fbb9d9bf)

I was using odo successfully, then installed CRC, used it for a bit, and then stopped the CRC cluster.

## Actual behavior
Tried to use odo again and every command (even ones I would have thought are 'local') takes >1min to respond. No errors, but every time you think it's hung and not going to return, then finally does. Even things that should fail fast (e.g. current dir is not a component) hang for a while.

## Expected behavior
I realize it's likely something underlying that is still trying to connect to my stopped cluster, but I didn't intentionally connect odo to it, nor is there any indication that's what the problem is: it just looks like odo performance is awful and it's my experience with odo that I associate with the problem.

There should be:
 - a shorter timeout
 - when there is a long delay, there should be some indication that the command is trying to connect (so I don't think it's hung)
 - when there's a timeout, there should be something that tells me how to stop trying to connect to that cluster

adisky:
**What type of PR is this?**
 /kind bug

Remove isResourceSupported call from command intialization

Fixes #4046 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Disconnect with the openshift/k8s cluster
- Run any odo command without changes 
- Run any odo command with  applying this PR
Observe the significant time difference between the output returned by odo

### DIFF
--- a/pkg/kclient/kclient.go
+++ b/pkg/kclient/kclient.go
@@ -200,7 +200,7 @@ func (c *Client) SetDiscoveryInterface(client discovery.DiscoveryInterface) {
 }
 
 func (c *Client) IsResourceSupported(apiGroup, apiVersion, resourceName string) (bool, error) {
-	klog.V(4).Infof("Checking is %s resource supported", resourceName)
+	klog.V(4).Infof("Checking if %q resource supported", resourceName)
 
 	if c.supportedResources == nil {
 		c.supportedResources = make(map[string]bool, 7)
--- a/pkg/kclient/kclient.go
+++ b/pkg/kclient/kclient.go
@@ -200,6 +200,8 @@ func (c *Client) SetDiscoveryInterface(client discovery.DiscoveryInterface) {
 }
 
 func (c *Client) IsResourceSupported(apiGroup, apiVersion, resourceName string) (bool, error) {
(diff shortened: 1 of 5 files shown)


pkg/kclient/kclient.go
```diff
@@ -200,6 +200,8 @@ func (c *Client) SetDiscoveryInterface(client discovery.DiscoveryInterface) {
 }
 
 func (c *Client) IsResourceSupported(apiGroup, apiVersion, resourceName string) (bool, error) {
+	klog.V(4).Infof("Checking is %s resource supported", resourceName)
```
dharmit, line 203: `Checking if %s resource is supported`

Also, I'd suggest `%q` instead of `%s`.

dharmit (changes requested):
One small change request. Looks good otherwise.

dharmit (approved):
/lgtm

PR #4316, opened 2020-12-17
Implement PVC for odo source volume

### LINKED ISSUE #3775: Allow users to toggle between ephemeral and persistent volumes for devfiles

adisky:
**What type of PR is this?**

 /kind feature

This PR proposes to create a PVC for odo source code volume, Uptil now we were using emptyDir to store source code.  
It makes PVC as a default option to store source code, If user wants to use `emptyDir` as source code, it can use
`odo preference set ephemeral true`

Fixes #3775

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- odo create nodejs --starter
- odo push
- check if it has created source pvc using `oc get pvc` or `kubectl get pvc`

### DIFF
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -85,8 +85,8 @@ const (
 	// DefaultRegistryCacheTime is time (in minutes) for how long odo will cache information from Devfile registry
 	DefaultRegistryCacheTime = 15
 
-	// EphemeralSourceVolume specifies if ephemeral volumes needs to be used as source volume.
-	EphemeralSourceVolume = "ephemeral"
+	// EphemeralSetting specifies if ephemeral volumes needs to be used as source volume.
+	EphemeralSetting = "Ephemeral"
 )
 
 // TimeoutSettingDescription is human-readable description for the timeout setting
@@ -101,8 +101,8 @@ var BuildTimeoutSettingDescription = fmt.Sprintf("BuildTimeout (in seconds) for 
 // RegistryCacheTimeDescription adds a description for RegistryCacheTime
 var RegistryCacheTimeDescription = fmt.Sprintf("For how long (in minutes) odo will cache information from Devfile registry (Default: %d)", DefaultRegistryCacheTime)
 
-// EphemeralSourceVolumeDescription adds a description for EphemeralSourceVolume
-var EphemeralSourceVolumeDescription = fmt.Sprintf("If true odo will create a emptyDir volume to store source code(Default: %t)", false)
+// EphemeralDescription adds a description for EphemeralSourceVolume
+var EphemeralDescription = fmt.Sprintf("If true odo will create a emptyDir volume to store source code(Default: %t)", false)
 
 // This value can be provided to set a seperate directory for users 'homedir' resolution
 // note for mocking purpose ONLY
@@ -119,7 +119,7 @@ var (
 		ExperimentalSetting:       ExperimentalDescription,
 		PushTargetSetting:         PushTargetDescription,
 		RegistryCacheTimeSetting:  RegistryCacheTimeDescription,
-		EphemeralSourceVolume:     EphemeralSourceVolumeDescription,
+		EphemeralSetting:          EphemeralDescription,
 	}
 
 	// set-like map to quickly check if a parameter is supported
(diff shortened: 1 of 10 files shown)


adisky:
not able to reproduce the failure locally
```
 Should ensure that files are correctly synced on pod redeploy, without force push specified [It]
    /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:715
    Expected
      
        <[]string | len:15, cap:16>: [
            "config",
            "devfile.yaml",
            "pom.xml",
            "src",
            "src/main",
            "src/main/java",
            "src/main/java/application",
            "src/main/java/application/Info.java",
            "src/main/java/application/SBApplication.java",
            "src/main/java/application/rest",
            "src/main/java/application/rest/v1",
            "src/main/java/application/rest/v1/Example.java",
            "src/main/resources",
            "src/main/resources/application-local.properties",
            "src/main/resources/application.properties",
        ]
    to equal
        <[]string | len:16, cap:16>: [
            "config",
            "devfile.yaml",
            "lost+found",
            "pom.xml",
            "src",
            "src/main",
            "src/main/java",
            "src/main/java/application",
            "src/main/java/application/Info.java",
            "src/main/java/application/SBApplication.java",
            "src/main/java/application/rest",
            "src/main/java/application/rest/v1",
            "src/main/java/application/rest/v1/Example.java",
            "src/main/resources",
            "src/main/resources/application-local.properties",
            "src/main/resources/application.properties",
        ]
```
don't know from where this directory `lost+found ` coming up

/retest

pkg/preference/preference.go
```diff
@@ -154,6 +161,8 @@ type OdoSettings struct {
 
 	// RegistryCacheTime how long odo should cache information from registry
 	RegistryCacheTime *int `yaml:"RegistryCacheTime,omitempty"`
+
+	EphemeralSourceVolume *bool `yaml:"ephemeral,omitempty"`
```
dharmit, line 165: I think this needs to be `yaml:"Ephemeral,omitempty"`. Capital `E` vs. `e`. Asking because this is how it shows up:
```sh
odo preference -h
Modifies odo specific configuration settings within the global preference file.

 
Available Global Parameters:
 BuildTimeout - BuildTimeout (in seconds) for waiting for a build of the git component to complete (Default: 300)
 Experimental - Set this value to true to expose features in development/experimental mode
 NamePrefix - Use this value to set a default name prefix (Default: current directory name)
 PushTarget - Set this value to 'kube' or 'docker' to tell odo where to push applications to. (Default: kube)
 PushTimeout - PushTimeout (in seconds) for waiting for a Pod to come up (Default: 240)
 RegistryCacheTime - For how long (in minutes) odo will cache information from Devfile registry (Default: 15)
 Timeout - Timeout (in seconds) for OpenShift server connection check (Default: 1)
 UpdateNotification - Flag to control if an update notification is shown or not (Default: true)
 ephemeral - If true odo will create a emptyDir volume to store source code(Default: false)
```
Everything except `ephemeral` is capitalized.

dharmit (changes requested):

dharmit (approved):
Overall the code does what it's meant for, but I'm facing some issues with setting and unsetting the value of `Ephemeral` in `~/.odo/preference.yaml` using `odo preference` commands. Opening a separate issue to address that.

/lgtm

dharmit:
> Opening a separate issue to address that.

https://github.com/openshift/odo/issues/4327

dharmit (approved):
Thanks for fixing `odo preference` related points. :+1: 

/hold cancel
/lgtm

PR #4206, opened 2020-11-09
Adding the latest java image ubi8/openjdk-11 in odo supported list

prietyc123:
**What type of PR is this?**
/kind failing-test
/kind feature

**What does does this PR do / why we need it**:

There is latest java image difference on 4.6 and 4.5 cluster. So, adding `ubi8/openjdk-11` image as supported will unblock enabling odo periodic job on 4.6 cluster https://github.com/openshift/odo/issues/4138

**Which issue(s) this PR fixes**:

Fixes - part of issue https://github.com/openshift/odo/issues/4138

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [x] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:
Latest java image on 4.6 should execute successfully for odo

Issue #4301, opened 2020-12-10
Failing "odo link" tests are blocking CI system


dharmit:
/kind bug
/area linking

odo doesn't fail if at the time of doing `odo link <cr-name>/<cr-instance-name>`, it fails to create a Secret on the cluster. Creation of this Secret is a task performed by Service Binding Operator. But if odo doesn't fail when such a Secret doesn't get created, it can cause user confusion as well as CI issues as we're seeing in our environment.

_Originally posted by @dharmit in https://github.com/openshift/odo/issues/3256#issuecomment-742480657_

dharmit:
Picking this up for Sprint 195 since I'm observing these show up often for periodic jobs.

dharmit:
What's interesting is that this test hasn't failed remotely as often on the integration tests as it has failed in periodic e2e tests. Before I make any change in odo code w.r.t this issue, I'd prefer we have more information about the cause for such erratic beahviour.

@prietyc123 @mohammedzee1000 can you folks shed any light on why we're seeing this mostly in periodic tests only?

prietyc123:
> What's interesting is that this test hasn't failed remotely as often on the integration tests as it has failed in periodic e2e tests. 

This is really a weird behaviour and more interestingly we are hitting it on PSI as well https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4063/pull-ci-openshift-odo-master-v4.6-integration-e2e/1344158957064687616#1:build-log.txt%3A1371 . Kind of blocker for POC pr https://github.com/openshift/odo/pull/4063

> Before I make any change in odo code w.r.t this issue, I'd prefer we have more information about the cause for such erratic beahviour.

I think if we want to get more info on this then the only way would be adding more debug logs. 

/priority-high

dharmit:
> I think if we want to get more info on this then the only way would be adding more debug logs.

Can we have access to the system that's hosting the tests when we hit this failure? That would be much more helpful for me to troubleshoot the problem. cc @mohammedzee1000 (for PSI).

dharmit:
> odo doesn't fail if at the time of doing `odo link <cr-name>/<cr-instance-name>`, it fails to create a Secret on the cluster. Creation of this Secret is a task performed by Service Binding Operator. But if odo doesn't fail when such a Secret doesn't get created, it can cause user confusion as well as CI issues as we're seeing in our environment.


[...]

PR #4317, opened 2020-12-18
Some fix for configure script of s390x.

zhengxiaomei123:
**What type of PR is this?**
 /kind bug
/kind failing-test

Add same fix with #4206.
Add some fix for importing java imagestream.

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/configure-installer-tests-cluster-s390x.sh
+++ b/scripts/configure-installer-tests-cluster-s390x.sh
@@ -12,7 +12,7 @@ KUBEADMIN_SCRIPT="$LIBCOMMON/kubeconfigandadmin.sh"
 # Exported to current env
 
 # list of namespace to create
-IMAGE_TEST_NAMESPACES="openjdk-11-rhel8 nodejs-12-rhel7 nodejs-12"
+IMAGE_TEST_NAMESPACES="openjdk-11-rhel8 nodejs-12-rhel7 nodejs-12 openjdk-11"
 
 . $KUBEADMIN_SCRIPT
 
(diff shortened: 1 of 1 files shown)


rnapoles-rh:
@zhengxiaomei123 integration tests failed.

dharmit:
@prietyc123 @mohammedzee1000 this PR is modifying the cluster configuration script for s390x architecture. But the 4.6 tests that are failing are being run on x86, right? What I'm observing in the test failures is that `odo catalog list services` command is failing in Operator Hub integration tests because it can't find any Operators installed on the cluster.

In this PR, I'm seeing that `KUBEADMIN_PASSWORD` value is being removed from configuration script of s390x: https://github.com/openshift/odo/pull/4317/files#diff-e24df2f816e33c2abe791ad856ac6f922a218705363d4f3176ec5e12050b5cb2L103-L106. But why would that cause failures in our integration tests that are running on x86? That should use `configure-installer-tests-cluster.sh` script, no?

mohammedzee1000:
Interesting, that should not affect x86 indeed,

mohammedzee1000 (approved):
Looks good to me codewise
/lgtm

scripts/configure-installer-tests-cluster-s390x.sh
```diff
@@ -73,9 +73,13 @@ sleep 5
 oc import-image nodejs:latest --from=registry.redhat.io/rhscl/nodejs-12-rhel7 --confirm -n openshift
 sleep 5
 oc annotate istag/nodejs:latest tags=builder -n openshift --overwrite
+oc delete istag java:8 -n openshift
+sleep 5
```
prietyc123, line 77: @zhengxiaomei123 Why do we have `sleep 5`? Wouldn't this slow down the test execution.

prietyc123 (changes requested):

rnapoles-rh:
@zhengxiaomei123 also wondering why the sleep 5 is needed.

scripts/configure-installer-tests-cluster-s390x.sh
zhengxiaomei123: @prietyc123 @rnapoles-rh with Sleep 5 to confirm the image is here after importing and to confirm the image tag is deleted . We met error before without sleep 5. The error is not always happening but happening many times.

dharmit: @zhengxiaomei123 I agree with you point but @prietyc123 has a valid point here as well. One way to address both of your concerns could be to use the `oc import-image` command the way it's used here: https://github.com/openshift/odo/blob/f463d096610b9c88d2507d8481d0dd09b47ba650/tests/helper/helper_oc.go#L312-L317

More specifically, use `oc import-image` command with `--request-timeout 5m` flag. That should give us way more time than `sleep 5` would to import the image.

@prietyc123 WDYT?

prietyc123: > More specifically, use oc import-image command with --request-timeout 5m flag.

+1, I would really prefer to use `--request-timeout` than `sleep`.

zhengxiaomei123: @dharmit Thanks very much. I updated the code following your good points. @prietyc123 @mohammedzee1000 @rnapoles-rh Please help review again. Thanks.

prietyc123:
Failing due to https://github.com/openshift/odo/issues/4301 .

/retest

rnapoles-rh (approved):

Issue #3924, opened 2020-09-09
Cleanup odo create validation headers and spinners for devfile


maysunfaisal:
Looks like `odo create` does two validations - 
- ValidateK8sResourceName in Validate()
- ParseFromDataAndValidate in Run()->devfileRun()

So if there is an actual devfile validation error when validating the components/commands/events in ParseFromDataAndValidate, it says validation passed from ValidateK8sResourceName and then outputs the ParseFromDataAndValidate error:
```
Maysuns-MacBook-Pro:springboot-ex maysun$ odo create myspr
Experimental mode is enabled, use at your own risk

Validation
 ✓  Creating a devfile component from devfile path: /Users/maysun/dev/redhat/resources/springboot-ex/devfile.yaml [51615ns]
 ✓  Validating devfile component [203508ns]
 ✗  the command "devbuild" does not map to a container component
```

We should be precise with the headers and spinners

/kind cleanup
/area devfile

johnmcollier:
As a side note, we should probably also show more than _just_ the validation header. For example, `  ✓  Creating a devfile component from devfile path` really doesn't belong under `Validation`

girishramnani:
we need to add spinner for storage related operations too as it was present for s2i components

niconosenzo:
I would like to have a look at this one if you don't mind :).

dharmit:
Closed by #4325

/close

PR #4325, opened 2020-12-22
Validation spinners refactor

niconosenzo:
**What type of PR is this?**

/kind code-refactoring

This PR redefine the "Validation" header into below different headers:

- Devfile Object Validation 
- Devfile Component Name Validation
- Supportability Validation

Also re-phrased ValidateK8sResourceName() spinner so it now reflects what kind of check it performs. 
`odo create` output now looks like:

<p align="center">

</p>

And,
<p align="center">

</p>

Address issue #3924 

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -837,7 +837,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 // Validate validates the create parameters
 func (co *CreateOptions) Validate() (err error) {
 
-	log.Info("Devfile Component Name Validation")
+	log.Info("Validation")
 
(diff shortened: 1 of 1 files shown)


girishramnani:
the `Devfile Component Name` validation would always have a single check so we should keep it under the `Validation` header only.

niconosenzo:
> the `Devfile Component Name` validation would always have a single check so we should keep it under the `Validation` header only.

@girishramnani so the headers would be like:

Devfile Object Validation
Validation (Merged Devfile Component Name Validation and Supportability Validation, both are single check validations).

Is that correct ?

girishramnani:
Yes

niconosenzo:
> Yes

Done.

Issue #4210, opened 2020-11-11
describe and create commands test for existing component in context folder is inconsistent


dgolovin:
/kind bug

## What versions of software are you using?

**Operating System:** Linux

**Output of `odo version`:** 2.0.0

## How did you run odo exactly?

Current folder has .odo/env left after previous component and no devfile.yaml

```
$ odo describe
✗  The current directory does not represent an odo component. Use 'odo create' to create component here or switch to directory with a component
```

## Actual behavior

```
$ odo create nodejs
✗  this directory already contains a component
```

## Expected behavior

```odo create nodejs```  creates component

## Any logs, error output, etc?

N/A

girishramnani:
I think as there is no devfile present - the user cannot move forward in anyway, one solution I feel is allowing the user to override the `env/env.yaml` when user does `odo create nodejs`.

kadel:
> I think as there is no devfile present - the user cannot move forward in anyway, one solution I feel is allowing the user to override the `env/env.yaml` when user does `odo create nodejs`.

This should work. 

It would be interesting to know how this situation can happen.


> Current folder has .odo/env left after previous component and no devfile.yaml

@dgolovin, how did you get it into this state? There should be no way to create this situation with odo.

dgolovin:
@kadel I probably did it outside of odo, by deleting devfile.yaml form file browser, but did not see `.odo` folder because it was hidden. Behaviour should be ether overwriting .odo/env/env.yaml in case of something missing to detect component or recognizing component with invalid configuration, but it should be consistent throughout all commands.

girishramnani:
Understood moving forward on the approach I have mentioned above

PR #4332, opened 2020-12-29
override an env file if found without a devfile on odo create

girishramnani:
**What type of PR is this?**

/kind bug

override an env file if found without a devfile on odo create

Fixes https://github.com/openshift/odo/issues/4210

**PR acceptance criteria**:

- [ ] Unit test 

- [X] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -685,7 +685,9 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			return err
 		}
 	} else {
-		co.SetComponentSettings(args)
+		if err := co.SetComponentSettings(args); err != nil {
+			return err
+		}
 	}
 
 	if co.now {
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -371,7 +371,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 	}
 
 	if util.CheckPathExists(EnvFilePath) && !util.CheckPathExists(co.DevfilePath) {
-		log.Warningf("Found a dangling env file without a devfile, overriding it")
+		log.Warningf("Found a dangling env file without a devfile, overwriting it")
 		if err := util.DeletePath(EnvFilePath); err != nil {
 			return err
 		}
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -10,7 +10,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/zalando/go-keyring"
-	"k8s.io/klog"
 
 	devfilev1 "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/pkg/devfile"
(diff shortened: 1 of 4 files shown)


girishramnani:
CI pass in one run 🥳

pkg/odo/cli/component/create.go
```diff
@@ -370,10 +366,17 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 	}
 	co.DevfilePath = DevfilePath
 
-	if util.CheckPathExists(ConfigFilePath) || util.CheckPathExists(EnvFilePath) {
+	if util.CheckPathExists(ConfigFilePath) {
 		return errors.New("this directory already contains a component")
 	}
 
+	if util.CheckPathExists(EnvFilePath) && !util.CheckPathExists(co.DevfilePath) {
+		log.Warningf("Found a dangling env file without a devfile, overriding it")
```
dharmit, line 374: `s/override/overwrite`

pkg/odo/cli/component/create.go
```diff
@@ -370,10 +366,17 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+		if err := util.DeletePath(EnvFilePath); err != nil {
+			return err
```
dharmit, line 376: Can we replace:
```go
if err := util.DeletePath(EnvFilePath); err != nil {
    return err
```
with
```go
return util.DeletePath(EnvFilePath)
```
Of course, only if it doesn't seem to hurt the readability. I asked this because I found a few occurrences of something like it in odo code.

tests/helper/helper_run.go
```diff
@@ -35,6 +35,15 @@ func CmdShouldPass(program string, args ...string) string {
 	return string(session.Wait().Out.Contents())
 }
 
+// CmdShouldPassIncludeErrStream returns stdout and stderr if command succeeds
+func CmdShouldPassIncludeErrStream(program string, args ...string) (string, string) {
```
dharmit, line 39: The only difference between this function and `CmdShouldPass` seems like this one returns `stderr` as well. Why not modify `CmdShouldPass` instead? 

One reason for not doing so could be it would require modification at tonnes of places in this very PR. But keeping a separate function would mean that someone might forget that something like this exists.

If there's no other reason than having to modify `CmdShouldPass` at many places, I'd request we modify `CmdShouldPass` instead of having `CmdShouldPassIncludeErrStream`.

dharmit (changes requested):

tests/helper/helper_run.go
girishramnani: currently - we cannot solve this problem other then rewriting the whole helper

pkg/odo/cli/component/create.go
girishramnani: we need to return the error **only if there is one i.e. its not nil**. What you are suggesting is to return irrespective of an error which is logically incorrect.

dharmit: Hmm. Second time in a week that I suggested something like this https://github.com/openshift/odo/pull/4311#discussion_r547119944. Something's wrong in the way I'm looking at these things.

dharmit (approved):
/lgtm

Issue #4305, opened 2020-12-13
devfile validation fails if folder name is not valid


gorkem:
/kind bug

## What versions of software are you using?

**Operating System:**
MacOs

**Output of `odo version`:**

odo 2.0.3

## How did you run odo exactly?

odo push

## Actual behavior
When I run odo push on a folder with devfile. devfile validation fails with the following. 
the _staticHTML_ is the name of the folder and I never set a component name. 
I think odo is setting the component name from folder name without converting it to be vaid.

```shell
❯ odo push

Validation
 ✗  Failed to start component with name staticHTML. Error: Failed to create the component: component name "staticHTML" is not valid, component name should conform the following requirements: 
- Contain at most 63 characters
- Contain only lowercase alphanumeric characters or ‘-’
- Start with an alphanumeric character
- End with an alphanumeric character
- Must not contain all numeric values
```

## Expected behavior
validation does not fail

kadel:
> I think odo is setting the component name from folder name without converting it to be vaid.

Yes, that is what is happening here :-(

odo needs to convert the directory name into the always valid component name.
odo needs to do the following  cleanup before using value as a component name:
 1. convert to lowercase
 2. remove all non-alphanumeric characters except `-`
 4. if it is all numeric values, prepend it with `x`
 3. remove all leading and trailing non-alphanumeric characters (because of the previous step, this can be only `-`)

girishramnani:
I can take a look at this

PR #4333, opened 2020-12-29
sanitize dir name before using it on odo push

girishramnani:
**What type of PR is this?**

 /kind bug

sanitize dir name before using it on odo push

Fixes https://github.com/openshift/odo/issues/4305

**PR acceptance criteria**:

- [X] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- create a devfile with no `name` metadata field
-  do `odo push` on it with some invalid directory names
   - 4564564
   - staticHTML
   - 54sldkff%%23

And it would santize them before using them as names


pkg/util/util.go
```diff
@@ -314,11 +314,16 @@ func GetDNS1123Name(str string) string {
 	withReplacedChars := strings.Replace(
 		nonAllowedCharsRegex.ReplaceAllString(str, "-"),
 		"--", "-", -1)
-	return removeNonAlphaSuffix(removeNonAlphaPrefix(withReplacedChars))
+	name := strings.ToLower(removeNonAlphaSuffix(removeNonAlphaPrefix(withReplacedChars)))
+	// if the name is all numeric
+	if len(str) != 0 && len(name) == 0 {
```
dharmit, line 319: How does checking the length of variables help us determine if the name is all numeric? What am I missing here?

dharmit (changes requested):
:+1: for removing an `else` block there. But I'm confused.

pkg/util/util.go
girishramnani: If the len of the original string is not 0 is the control condition just or make sure no one passed a blank directory,
Then after converting it to DNS1123 - if all are numeric the “removeNonAlphaprefix” would remove all characters as everything is non alpha hence returning a blank string.

dharmit:
@girishramnani can you PTAL at the failing integration test for this PR?

pkg/url/url.go
```diff
@@ -1117,7 +1119,7 @@ func RemoveEndpointInDevfile(devObj parser.DevfileObj, urlName string) error {
 
 // getValidURLName returns valid URL resource name for Kubernetes based cluster
 func getValidURLName(name string) string {
-	trimmedName := strings.TrimSpace(util.GetDNS1123Name(strings.ToLower(name)))
+	trimmedName := strings.TrimSpace(slug.Make(name))
```
girishramnani, line 1122: GetDNS1123Name now has a way stronger check - not needed for a url string

dharmit (commented):
LGTM if url tests are happy. I'll add the lgtm label when I see the CI green or at least not red because of failing `odo url` tests.

dharmit:
/retest

CI failing due to #4347

girishramnani:
This PR would break existing urls

dharmit:
> This PR would break existing urls

/hold

girishramnani:
we can say that this one was bot's fault

dharmit:
I hope it doesn't fire back. :crossed_fingers:

kadel:
So we know that it can break existing users and leaving it as it is?? 😱

dharmit:
> This PR would break existing urls

@girishramnani Can you elaborate a little as to where you think this could blow up?

Issue #4298, opened 2020-12-09
Refactoring odo in small chunks


dharmit:
## What are we doing?
For Sprint 194, we will work on refactoring odo code in small chunks. What this means is that we will try to:
1. remove functions or code that's repeating across packages/layers but doing (almost) the same thing
1. break large function that's doing multiple things into smaller functions that are easier to read and understand
1. add more comments to the code to make it simpler for other contributors to understand
1. unify repeating code into functions
1. clean the code iteratively with small and focussed PRs

### Why are we not doing what's suggested/discussed in #4057? What about the refactoring issues like #4081, #4093, #4115, #4145, and more?
We will still be addressing these issues when we're more confident. Yes, we're not enough confident right now to fully embrace Go interfaces and make our code adhere to Go best practices. If you have any ideas/suggestions, or feel like helping out, please do! The goal of odo project still remains the same - to achieve code architecture similar to what's discussed in #4057.

### What's the desired outcome of this issue?
Parts of our code were added/developed by contributors who are not active at the moment. While working on this issue we would like to better understand the flow of things across the code instead of limiting ourselves only to the packages we contributed to in the past.

### What's the Acceptance Criteria for this issue?
There's no predefined criteria except the points mentioned under "What are we doing?" As we open & merge PRs, we will mention them in the comments on this issue.

dharmit:
@kadel @mik-dass @adisky let me know if anything about issue description needs to be changed.

dharmit:
In Sprint 194, as a part of this issue, I started stepping through the code to understand where we can make improvements. I ended up creating:
* PR 4322
* PR 4323
* Issue #4313

And to the best of my knowledge, @mik-dass opened PR 4311.

In Sprint 195, I intend to continue stepping through the code as much as I can. AFAIK, @mik-dass is going to open a PR to refactor storage bits once PR 4311 is merged.

@mik-dass correct me if I'm wrong.

PR #4344, opened 2021-01-05
Refactors storage code for localConfigProvider.

mik-dass:
**What type of PR is this?**

/kind code-refactoring

It adds storage related functions to the localConfigProvider and also refactors the storage create and delete CLI code to use them.

Fixes part of #4298  

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Tests should pass.

### DIFF
--- a/pkg/config/storage.go
+++ b/pkg/config/storage.go
@@ -1,72 +1,84 @@
 package config
 
 import (
+	"fmt"
+
+	"github.com/openshift/odo/pkg/localConfigProvider"
 	"github.com/pkg/errors"
 )
 
-// StorageCreate sets the storage related information in the local configuration
-func (lci *LocalConfigInfo) StorageCreate(name, size, path string) (ComponentStorageSettings, error) {
-	storage := ComponentStorageSettings{
-		Name: name,
-		Size: size,
-		Path: path,
-	}
-	err := lci.SetConfiguration("storage", storage)
-	if err != nil {
-		return ComponentStorageSettings{}, err
-	}
-	return storage, err
-}
-
-func (lci *LocalConfigInfo) StorageExists(storageName string) bool {
-	for _, storage := range lci.GetStorage() {
+// GetStorage gets the storage with the given name
+func (lci *LocalConfigInfo) GetStorage(storageName string) *localConfigProvider.LocalStorage {
+	for _, storage := range lci.ListStorage() {
 		if storageName == storage.Name {
-			return true
-		}
-	}
-	return false
-}
-
-func (lci *LocalConfigInfo) StorageList() ([]ComponentStorageSettings, error) {
-	storageConfigList := lci.GetStorage()
-	var storageList []ComponentStorageSettings
-	for _, storage := range storageConfigList {
-		storageList = append(storageList, ComponentStorageSettings{
-			Name: storage.Name,
-			Path: storage.Path,
-			Size: storage.Size,
-		})
-	}
-	return storageList, nil
-}
-
-func (lci *LocalConfigInfo) ValidateStorage(storageName, storagePath string) error {
-	for _, storage := range lci.GetStorage() {
-		if storage.Name == storageName {
-			return errors.Errorf("there already is a storage with the name %s", storageName)
-		}
-		if storage.Path == storagePath {
-			return errors.Errorf("there already is a storage mounted at %s", storagePath)
+			return &storage
 		}
 	}
 	return nil
 }
 
-func (lci *LocalConfigInfo) StorageDelete(name string) error {
-	exists := lci.StorageExists(name)
-	if !exists {
+// CreateStorage sets the storage related information in the local configuration
+func (lci *LocalConfigInfo) CreateStorage(storage localConfigProvider.LocalStorage) error {
+	err := lci.SetConfiguration("storage", storage)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+// ListStorage gets all the storage from the config
+func (lci *LocalConfigInfo) ListStorage() []localConfigProvider.LocalStorage {
+	if lci.componentSettings.Storage == nil {
+		return []localConfigProvider.LocalStorage{}
+	}
+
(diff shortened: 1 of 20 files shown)


pkg/config/storage.go
```diff
@@ -1,72 +1,84 @@
...
 )
 
-// StorageCreate sets the storage related information in the local configuration
-func (lci *LocalConfigInfo) StorageCreate(name, size, path string) (ComponentStorageSettings, error) {
-	storage := ComponentStorageSettings{
-		Name: name,
-		Size: size,
-		Path: path,
+// GetStorage gets the storage with the given name
+func (lci *LocalConfigInfo) GetStorage(storageName string) *localConfigProvider.LocalStorage {
```
dharmit, line 11: Since there's no modification of the data, this would be better off as a function instead of being a method attached to the `LocalConfigInfo`, IMO.

pkg/config/storage.go
```diff
@@ -1,72 +1,84 @@
...
-func (lci *LocalConfigInfo) StorageDelete(name string) error {
-	exists := lci.StorageExists(name)
-	if !exists {
-		return errors.Errorf("storage named %s doesn't exists", name)
+// CompleteStorage completes the given storage
+func (lci *LocalConfigInfo) CompleteStorage(storage *localConfigProvider.LocalStorage) {}
+
+// ValidateStorage validates the given storage
+func (lci *LocalConfigInfo) ValidateStorage(storage localConfigProvider.LocalStorage) error {
+	if storage.Size == "" || storage.Path == "" {
```
dharmit, line 60: Can we add a comment here that this error is expected only when working with s2i components and not with devfile components?

pkg/odo/cli/storage/create.go
```diff
@@ -30,121 +25,69 @@ var (
...
-		}
-	} else {
-		_, err := o.LocalConfigInfo.StorageCreate(o.storageName, o.storageSize, o.storagePath)
-		if err != nil {
-			return err
-		}
-	}
 
-	storageResultMachineReadable := storage.GetMachineReadableFormat(o.storageName, o.storageSize, o.storagePath)
+	storageResultMachineReadable := storage.GetMachineReadableFormat(o.storage.Name, o.storage.Size, o.storage.Path)
```
dharmit, line 85: I understand that this is not something that's been touched by your PR but would it make sense to change the signature of `GetMachineReadableFormat` to just take the storage's `CreateOptions` instead of taking three `string`s?

I'm asking this because if we modify this JSON output in future to add/remove something from it, we will have to change the function signature. And that would break unit tests. Instead, if we modify it to just use one parameter, the storage `CreateOptions`, unit tests can help us point when something's not working as expected.

pkg/odo/cli/storage/create.go
```diff
@@ -30,121 +25,69 @@ var (
...
-func (o *StorageCreateOptions) Validate() (err error) {
-	if o.isDevfile {
-		return
-	}
-	// validate storage path
-	return o.LocalConfigInfo.ValidateStorage(o.storageName, o.storagePath)
+// Validate validates the CreateOptions based on completed values
+func (o *CreateOptions) Validate() (err error) {
+	// validate the storage
+	return o.LocalConfigInfo.ValidateStorage(o.storage)
```
dharmit, line 75: Why are we using the same function to validate storage for components of s2i as well as devfile? Or am I missing something?

pkg/odo/cli/storage/create.go
dharmit: Also, are we evaluating this if it's needed only in the case when the user has request for JSON output? Shouldn't this be in the block of `if log.IsJSON()` instead?

dharmit (changes requested):
Review based on playing with `odo storage create` for s2i & devfile component types

pkg/odo/cli/storage/delete.go
```diff
@@ -29,97 +24,61 @@ var (
...
-			return err
-		}
-		mPath, err = devFile.Data.GetVolumeMountPath(o.storageName)
-		if err != nil {
-			return err
-		}
-	} else {
-		mPath = o.LocalConfigInfo.GetMountPath(o.storageName)
+func (o *DeleteOptions) Run() (err error) {
+	mPath, err := o.Context.LocalConfigProvider.GetStorageMountPath(o.storageName)
```
dharmit, line 68: Thinking out loud - since this function's finding if the storage with the provided name exists, doesn't it make more sense to put it in `Validate`?

dharmit (changes requested):
Only one question from trying to do `odo storage delete`.

pkg/config/storage.go
mik-dass: We need it to call `ListStorage()`

pkg/config/storage.go
mik-dass: This is in the `localConfig` package used by s2i components and we should not mention devfile here.

pkg/odo/cli/storage/create.go
mik-dass: > I understand that this is not something that's been touched by your PR but would it make sense to change the signature of GetMachineReadableFormat to just take the storage's CreateOptions instead of taking three strings?

That would create a import cycle.

> Also, are we evaluating this if it's needed only in the case when the user has request for JSON output? Shouldn't this be in the block of if log.IsJSON() instead?

Fixed

pkg/odo/cli/storage/create.go
mik-dass: Fixed

pkg/odo/cli/storage/delete.go
mik-dass: > Thinking out loud - since this function's finding if the storage with the provided name exists, doesn't it make more sense to put it in Validate?

That function just gets the mount path for printing and doesn't perform any validation.

dharmit (approved):
/lgtm

PR #4345, opened 2021-01-05
Separates test into its own spec

### LINKED ISSUE #4301: Failing "odo link" tests are blocking CI system

dharmit:
**What type of PR is this?**
/kind failing-test

It separates an edge case of tests into its own spec.

Fixes #4301

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

Test should pass. Especially for #4063.


dharmit:
/retest

CI failing due to #4347

PR #4375, opened 2021-01-20
Elaborate on minikube, ingress, nip.io setup

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

**What type of PR is this?**
/kind documentation
[skip ci]

Though this is real basic, getting started info it doesn't include the most basic steps required for having odo leverage the ingress on a basic kubernetes setup like minikube.   

**PR acceptance criteria**:

- [ ] Documentation 

### DIFF
--- a/docs/public/deploying-a-devfile-using-odo.adoc
+++ b/docs/public/deploying-a-devfile-using-odo.adoc
@@ -483,7 +483,7 @@ Pushing devfile component myopenliberty
 
 ----
 
-. View your deployed application in a browser using the generated url
+. List the URLs of the component
 +
 [source,sh]
 ----
@@ -493,15 +493,21 @@ Pushing devfile component myopenliberty
   myopenliberty-9     Pushed     http://myopenliberty-9.192.168.99.100.nip.io        9080     false
 ----
 
-
-. Have odo watch for changes in the source code
+. View your deployed application using the generated URL (this example shows an ingress hostname URL, while an OpenShift route would look a bit different):
++
+[source,sh]
+----
+ $ curl http://myopenliberty-9.$(minikube ip).nip.io/api/resource
+----
++
+. Have odo watch for changes in the source code:
 +
 [source,sh]
 ----
  $ odo watch
 ----
 
-You can now continue developing your application. Just refreshing the browser will render the source code changes.
+You can now continue developing your application. Refreshing the browser or hitting the endpoint again will render the source code changes.
 
 Run `odo delete` to delete the application from cluster.
 
--- a/docs/public/deploying-a-devfile-using-odo.adoc
+++ b/docs/public/deploying-a-devfile-using-odo.adoc
@@ -29,29 +29,50 @@ An example deployment scenario:
 
 == Deploying your first devfile
 
-===== Prerequisites for a Kubernetes Cluster
+=== Ingress Setup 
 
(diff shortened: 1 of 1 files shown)


awisniew90:
Looks good to me. One nit would be to specify for each stack that the `odo url create --host $(minikube ip).nip.io` command is an example if you are using minikube and nip.io. Seems obvious, but since these commands are for generic kubernetes, it might be good to clarify.  Since you already mention this in the above "Ingress Setup" section, maybe just append to the note:

```
NOTE: If deploying on OpenShift, you can skip this step and a Route will be created for you automatically. On Kubernetes, you need to pass ingress domain name via `--host` flag. See "Ingress Setup".
```

scottkurz:
> 
> 
> Looks good to me. One nit would be to specify for each stack that the `odo url create --host $(minikube ip).nip.io` command is an example if you are using minikube and nip.io. Seems obvious, but since these commands are for generic kubernetes, it might be good to clarify. Since you already mention this in the above "Ingress Setup" section, maybe just append to the note:
> 
> ```
> NOTE: If deploying on OpenShift, you can skip this step and a Route will be created for you automatically. On Kubernetes, you need to pass ingress domain name via `--host` flag. See "Ingress Setup".
> ```

Thx @awisniew90 for the review.  I'm going to see what @dharmit says before making any other changes.

docs/public/deploying-a-devfile-using-odo.adoc
```diff
@@ -29,29 +29,50 @@ An example deployment scenario:
...
+  $ minikube addons enable ingress
+----
+
+With Minikube running in a virtual machine, the ingress controller IP address is obtained via `minikube ip` (the value is `192.168.99.100` in the sample output shown below).
+
+===== OpenShift Cluster Ingress Setup
 
-. Log in to an OpenShift cluster:
+For link:https://developers.redhat.com/products/codeready-containers/overview[CodeReady Containers] running in a virtual machine, for example, you can get it by `crc ip`.
+Note you might not need to use the ingress in OpenShift, where you can use routes instead.
```
dharmit, line 59: So far, I don't think we're suggesting the users to use ingress on OpenShift/CRC. That's because, as you rightly mentioned, routes make things easier on OpenShift. IMO, we should change the wording so that it's clear to the reader that OpenShift has routes, which are simpler to use, but they can _also_ use ingress if they would like to.

docs/public/deploying-a-devfile-using-odo.adoc
```diff
@@ -29,29 +29,50 @@ An example deployment scenario:
...
+
+===== Ingress Notes
+
+Of course there are other options but this approach avoids the need to edit the `/etc/hosts/` file with each created service URL.
+
+Checkout this https://kubernetes.io/docs/concepts/services-networking/ingress/[document] to know more about ingress.
+
+=== First Steps
+
+* Login to your cluster (unnecessary if you've used other standard methods, e.g. kubectl to establish the current context)
```
dharmit, line 69: `s/unnecessary/not required/`. Suggesting this because "unnecessary" gives quite different perspective to things, IMO.

dharmit (changes requested):
Required few minor tweaks.

dharmit:
@scottkurz that's OK. You can still open a quick PR to address the requests made by @awisniew90 and me.

In the meanwhile, I have pushed the changes made in this PR to [odo.dev](https://odo.dev) website.

PR #4396, opened 2021-01-29
adding project resource check before warning about default project usage

### LINKED ISSUE #4309: Confusing message about default project.
```
▶ odo create java-maven
 ⚠  odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.
Validation
 ✓  Checking devfile existence [23679ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [51231ns]
 ✓  Validating devfile component [95996ns]

```

> ⚠  odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.

This is a really confusing message for both Kubernetes and OpenShift users

### Kubernetes
For minikube or any other Kubernetes users this message provides no value. It doesn't matter if they are using the `default` project or not. 
#### Solution
Add a condition that will check if the cluster supports `Project` resource. If not, the warning should not be displayed even if the `default` project is being used.

### OpenShift
OpenShift users should not use the `default` project. 
But the problem is that if the user does what the warning says; run `odo project create` it won't help them, the component project name is already recorded in `env.yml`, so it will have no effect on the existing component.
#### Solution
Not sure what to do here. I would probably make this warning error and completely prevent users from using `default` project on OpenShift. 

(a little bit of background why we don't users to use `default` project is in https://github.com/openshift/odo/issues/4027#issuecomment-700179907)

niconosenzo:
**What type of PR is this?**
/kind bug

Adding condition to check if `project` is a supported resource during `odo create` command, if so, check non-default project is being used. 

Fixes #4309

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

`odo create` should only error (and exit) with _odo may not work as expected[...]_ when using the default project in an Openshift cluster.

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -547,11 +547,11 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		projectSupported, err := co.Client.IsProjectSupported()
 
 		if err != nil {
-			return errors.Wrap(err, "Resource project validation check failed.")
+			return errors.Wrap(err, "resource project validation check failed.")
 		}
 
 		if projectSupported && componentNamespace == "default" {
-			return errors.New("odo may not work as expected in a default project, please run the odo component in a non-default project")
+			return errors.New("odo may not work as expected in the default project, please run the odo component in a non-default project")
 		}
 
 		// Set devfileMetadata struct
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -543,8 +543,16 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
-		if componentNamespace == "default" {
-			log.Warning("odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.")
+		// Check whether resource "Project" is supported
+		projectSupported, err := co.Client.IsProjectSupported()
+
+		if err != nil {
+			// we don't need to return (fail) if this happens
+			log.Warning("Project resource supportability check failed.")
+		}
+
+		if projectSupported && componentNamespace == "default" {
+			return errors.New("odo may not work as expected in a default project, please run the odo component in a non-default project")
 		}
 
 		// Set devfileMetadata struct
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -547,8 +547,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 		projectSupported, err := co.Client.IsProjectSupported()
 
 		if err != nil {
-			// we don't need to return (fail) if this happens
-			log.Warning("Project resource supportability check failed.")
+			return errors.Wrap(err, "Resource project validation check failed.")
 		}
 
 		if projectSupported && componentNamespace == "default" {
--- a/tests/integration/devfile/cmd_devfile_create_test.go
+++ b/tests/integration/devfile/cmd_devfile_create_test.go
@@ -54,7 +54,7 @@ var _ = Describe("odo devfile create command tests", func() {
 			helper.CmdShouldPass("odo", "create", "java-openliberty", "--project", componentNamespace)
 		})
 
-		It("should fail tocreate the devfile component if --project value is 'default'", func() {
+		It("should fail to create the devfile component if --project value is 'default'", func() {
 			output := helper.CmdShouldFail("odo", "create", "java", "--project", "default")
 			expectedString := "odo may not work as expected in the default project, please run the odo component in a non-default project"
 			helper.MatchAllInOutput(output, []string{expectedString})
--- a/tests/integration/devfile/cmd_devfile_create_test.go
+++ b/tests/integration/devfile/cmd_devfile_create_test.go
@@ -53,6 +53,13 @@ var _ = Describe("odo devfile create command tests", func() {
 			componentNamespace := helper.RandString(6)
 			helper.CmdShouldPass("odo", "create", "java-openliberty", "--project", componentNamespace)
 		})
+
+		It("should fail tocreate the devfile component if --project value is 'default'", func() {
+			output := helper.CmdShouldFail("odo", "create", "java", "--project", "default")
+			expectedString := "odo may not work as expected in the default project, please run the odo component in a non-default project"
+			helper.MatchAllInOutput(output, []string{expectedString})
+		})
+
 	})
 
 	Context("When executing odo create with devfile component type argument and --registry flag", func() {


dharmit:
@niconosenzo Code looks good to me. But I'll wait for @kadel to take a look at the PR because he was of the opinion to make this warning an error if user's running OpenShift cluster thus not letting them create a component if they're running `odo create` against `default` project on OpenShift/CRC cluster.

pkg/odo/cli/component/create.go
```diff
@@ -543,8 +543,16 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
-		if componentNamespace == "default" {
-			log.Warning("odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.")
+		// Check whether resource "Project" is supported
+		projectSupported, err := co.Client.IsProjectSupported()
+
+		if err != nil {
+			// we don't need to return (fail) if this happens
```
mik-dass, line 550: Why are we ignoring the error?

niconosenzo: My thought was to avoid exiting due to external reasons during validation (i.e., temporary Openshift API unresponsiveness ). Anyhow, I understand this will skip the conditional right below and may cause some troubles if the default namespace is being used. 
I will change it to catch the error and return if you concur.

mik-dass: Yeah let's catch and return the error 👍

niconosenzo: Done :smiley:

pkg/odo/cli/component/create.go
```diff
@@ -543,8 +543,15 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 			}
 		}
 
-		if componentNamespace == "default" {
-			log.Warning("odo may not work as expected in a default project, please run the odo component in a non-default project. To create a new project, use `odo project create`.")
+		// Check whether resource "Project" is supported
+		projectSupported, err := co.Client.IsProjectSupported()
+
+		if err != nil {
+			return errors.Wrap(err, "Resource project validation check failed.")
```
mik-dass, line 550: please start the error message with a small letter

pkg/odo/cli/component/create.go
```diff
@@ -543,8 +543,15 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
+		}
+
+		if projectSupported && componentNamespace == "default" {
+			return errors.New("odo may not work as expected in a default project, please run the odo component in a non-default project")
```
mik-dass, line 554: Can we change this to `in the default project`?

mik-dass:
Can we add a integration test for this behavior?

kadel:
> Can we add a integration test for this behavior?

I don't think that needs to have its own integration test. It is more of a UX improvement.


/approve

pkg/odo/cli/component/create.go
niconosenzo: Done.

pkg/odo/cli/component/create.go
niconosenzo: Changed.

mik-dass (approved):
/lgtm

prietyc123:
> > Can we add a integration test for this behavior?
> 
> I don't think that needs to have its own integration test. It is more of a UX improvement.
> 
> /approve

AFAIU It would be more accurate, if I say we are improving our feature which results into UX improvement, correct me if I am wrong. So IMO we should add test to check the behaviour of the functionality. I mean odo create ... should error out when we are in default name space or using default namespace through flag. For ex-

```
cmdShouldFail(odo, component, create, java, --project, default)
```

/hold

niconosenzo:
> > > Can we add a integration test for this behavior?
> > 
> > 
> > I don't think that needs to have its own integration test. It is more of a UX improvement.
> > /approve
> 
> AFAIU It would be more accurate, if I say we are improving our feature which results into UX improvement, correct me if I am wrong. So IMO we should add test to check the behaviour of the functionality. I mean `odo create ...` should error out when we are in default name space or using default namespace through flag. For ex-
> 
> ```
> cmdShouldFail("odo", "component", "create", "java", "--project", "default")
> ```
> 
> /hold

@prietyc123 ,We can add the test [here](https://github.com/openshift/odo/blob/master/tests/integration/devfile/cmd_devfile_create_test.go) :

```go
	Context("When executing odo create with devfile component type argument and --project flag", func() {
		It("should successfully create the devfile component", func() {
			componentNamespace := helper.RandString(6)
			helper.CmdShouldPass("odo", "create", "java-openliberty", "--project", componentNamespace)
		})

		It("should fail to create the devfile component if --project value is 'default'", func() {
			helper.CmdShouldPass("odo", "create", "java", "--project", "default")
		})
		
	})
```

If you agree I'd open a separate PR.

dharmit:
> If you agree I'd open a separate PR.

Tests go in the same PR. Can you push a patch that adds test?

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -53,6 +53,13 @@ var _ = Describe("odo devfile create command tests", func() {
 			componentNamespace := helper.RandString(6)
 			helper.CmdShouldPass("odo", "create", "java-openliberty", "--project", componentNamespace)
 		})
+
+		It("should fail tocreate the devfile component if --project value is 'default'", func() {
```
prietyc123, line 57: @niconosenzo could you please give a single space in `tocreate`.

prietyc123 (changes requested):

tests/integration/devfile/cmd_devfile_create_test.go
niconosenzo: @prietyc123 thanks, fixed.

prietyc123:
Thanks for addressing the concerns 🙂 

/lgtm

dharmit:
```
✗  Waiting for component to start [5m] [WARNING x7: Failed]
[ssh:Fedora 32] [odo]  ✗  Failed to start component with name dvpwko. Error: Failed to create the component: error while waiting for deployment rollout: timeout while waiting for dvpwko deployment roll out\nFor more information to help determine the cause of the error, re-run with '-v'.
[ssh:Fedora 32] [odo] See below for a list of failed events that occured more than 5 times during deployment:
[ssh:Fedora 32] [odo] 
[ssh:Fedora 32] [odo]  NAME                                      COUNT  REASON  MESSAGE                        
[ssh:Fedora 32] [odo] 
[ssh:Fedora 32] [odo]  dvpwko-5d6987f8b8-gwmnq.16623f8146e9c6e5  7      Failed  Error: secret                  
[ssh:Fedora 32] [odo]                                                           "dvpwko-etcdcluster-example"   
[ssh:Fedora 32] [odo]                                                           not found                      
[ssh:Fedora 32] [odo] 
[ssh:Fedora 32] [odo] 
[ssh:Fedora 32] Deleting project: cxezhimxkp
[ssh:Fedora 32] Running odo with args [odo project delete cxezhimxkp -f]
```
#4301

PR #4381, opened 2021-01-22
Adds debug-command flag to odo watch

mik-dass:
**What type of PR is this?**

/kind feature

**What does does this PR do / why we need it**:

It takes the custom debug command from the CLI while trigerring odo watch.

**Which issue(s) this PR fixes**:

Fixes #4357 

**PR acceptance criteria**:

- [X] Unit test 

- [X] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

- `odo watch --debug-command <custom debug command>` should work.

PR #4405, opened 2021-02-03
 Disable preStart event using validations

### LINKED ISSUE #4187: Odo should use apply command for preStart(and postStop when its implemented)
There is a semantic difference between an exec and apply command - exec commands are to be executed only when a container is up and running and apply commands are applied when there is no running container.

However, odo only has exec and composite commands in its scope and preStart utilizes an exec command. But according to the schema definition and interpretation this is not right. See issue https://github.com/devfile/api/issues/204 for discussion.

Currently no odo devfiles have a preStart, so it would be good to update the implementation to use apply commands before a devfile starts using it.

For more information on which events use apply/exec, pls read devfile validation doc - https://github.com/devfile/api/blob/master/pkg/validation/validation-rule.md. Also note the condition for composite commands when it comes to events!

cc: @kadel

kadel:
**What type of PR is this?**

/kind bug

preStart even is still not clearly defined in Devfile and it is clear how it should be defined. This will make sure that no one is using PreStart event until we have a proper implementation.

Fixes #4187

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Create a devfile with the preStart event command

```
events:
  preStart:
     - some-command
```

### DIFF
--- a/pkg/devfile/validate/errors.go
+++ b/pkg/devfile/validate/errors.go
@@ -44,5 +44,5 @@ type UnsupportedFieldError struct {
 }
 
 func (e *UnsupportedFieldError) Error() string {
-	return fmt.Sprintf("%s is not supported in odo", e.fieldName)
+	return fmt.Sprintf("%q is not supported in odo", e.fieldName)
 }
--- a/pkg/devfile/validate/errors.go
+++ b/pkg/devfile/validate/errors.go
@@ -38,3 +38,11 @@ type CompositeRunKindError struct {
 func (e *CompositeRunKindError) Error() string {
 	return "composite commands of run kind are not supported currently"
 }
+
+type UnsupportedFieldError struct {
+	fieldName string
+}
+
+func (e *UnsupportedFieldError) Error() string {
+	return fmt.Sprintf("%s is not supported in odo", e.fieldName)
+}
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -368,7 +368,7 @@ var _ = Describe("odo devfile push command tests", func() {
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", commonVar.Project, cmpName)
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), commonVar.Context)
-			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile-with-valid-events.yaml"), filepath.Join(commonVar.Context, "devfile.yaml"))
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile-with-preStart.yaml"), filepath.Join(commonVar.Context, "devfile.yaml"))
 
 			output := helper.CmdShouldFail("odo", "push", "--project", commonVar.Project)
 			// This is expected to fail for now.
--- a/tests/integration/devfile/cmd_devfile_push_test.go
+++ b/tests/integration/devfile/cmd_devfile_push_test.go
@@ -363,61 +363,67 @@ var _ = Describe("odo devfile push command tests", func() {
 		})
 
 		It("should execute PreStart commands if present during pod startup", func() {
-			expectedInitContainers := []string{"tools-myprestart-1", "tools-myprestart-2", "runtime-secondprestart-3"}
+			// expectedInitContainers := []string{"tools-myprestart-1", "tools-myprestart-2", "runtime-secondprestart-3"}
 
 			helper.CmdShouldPass("odo", "create", "nodejs", "--project", commonVar.Project, cmpName)
 
 			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project"), commonVar.Context)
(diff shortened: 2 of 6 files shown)


pkg/devfile/validate/errors.go
```diff
@@ -38,3 +38,11 @@ type CompositeRunKindError struct {
 func (e *CompositeRunKindError) Error() string {
 	return "composite commands of run kind are not supported currently"
 }
+
+type UnsupportedFieldError struct {
+	fieldName string
+}
+
+func (e *UnsupportedFieldError) Error() string {
+	return fmt.Sprintf("%s is not supported in odo", e.fieldName)
```
dharmit, line 47: Can we use `%q` instead of `%s`?

dharmit (changes requested):
Something's off here. When I do `odo push`, I see:
```sh
$ odo push        
 ✗  devfile events validation error: 
preStart type events are invalid: 
some-command does not map to a valid devfile command
```

I expected to see something along the lines of:
```sh
$ odo push
✗  devfile events validation error: 
"preStart" type events are not supported in odo
```

dharmit:
Also, are the integration tests written with the assumption that `PreStart` events should work fine? :thinking: 
```
 • Failure [2.025 seconds]
odo devfile push command tests
/go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:20
  Verify devfile push works
  /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:64
    should execute PreStart commands if present during pod startup [It]
    /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:365
    No future change is possible.  Bailing out early after 0.071s.
      
    Running odo with args [odo push --project tmpjpmfnmc]
    Expected
        <int>: 1
    to match exit code:
        <int>: 0
```

kadel:
> Something's off here. When I do `odo push`, I see:
> 
> ```shell
> $ odo push        
>  ✗  devfile events validation error: 
> preStart type events are invalid: 
> some-command does not map to a valid devfile command
> ```

This is actually correct. This is a validation that comes from `devfile/library`.
It means that the command that you are using in `preStart` even doesn't exist in `commands` definition. You can't reference non-existing command in events.

dharmit:
> It means that the command that you are using in `preStart` even doesn't exist in `commands` definition. You can't reference non-existing command in events.

Second time in less than 24 hours that I misunderstood or didn't know what command means in devfile lingo. I did the same on https://github.com/openshift/odo/pull/4381#issuecomment-772493200. This indicates two things to me:
1. I obviously don't understand the devfile API as well as people who designed (you) it or integrated it with odo (@mik-dass).
2. We need to make this obvious/clear in devfile documentation so that people don't fumble like I did. I'll leave this for you to take up with devfile team if it makes sense.

dharmit:
Some rebase mess up here https://github.com/openshift/odo/pull/4405#commits-pushed-6dea337?

kadel:
fixed

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -363,61 +363,67 @@ var _ = Describe("odo devfile push command tests", func() {
...
-						if len(fileContents)-1 != 3 {
-							statErr = fmt.Errorf("prestart event action count error, expected: 3 strings, got %d strings: %s", len(fileContents), strings.Join(fileContents, ","))
-						} else if cmdOp != "hello test2\nhello test2\nhello test\n" {
-							statErr = fmt.Errorf("prestart event action error, expected: hello test2\nhello test2\nhello test\n, got: %s", cmdOp)
+			output := helper.CmdShouldFail("odo", "push", "--project", commonVar.Project)
+			// This is expected to fail for now.
+			// see https://github.com/openshift/odo/issues/4187 for more info
+			helper.MatchAllInOutput(output, []string{"\"preStart\" is not supported in odo"})
+
+			/*
```
dharmit, line 378: Any reason why this piece of code is commented instead of being removed?

dharmit (approved):
Just a question about comment vs. deleting. But it's not a show stopper.

/lgtm

tests/integration/devfile/cmd_devfile_push_test.go
kadel: My reasoning was that disabling `preStart` is a temporary measure. Once it is clear how it should be implemented this might be useful again.

kadel:
```
[odo] : error while streaming command: error sending request: Post https://api.ci-op-tl8sj0lj-de358.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/avacaawobu/pods/xwslrp-7c86484c96-wgkxk/exec?command=%2Fbin%2Fsh&command=-c&command=cd+%24%7BPROJECTS_ROOT%7D+%26%26+npm+install&container=runtime&stderr=true&stdout=true: dial tcp: lookup api.ci-op-tl8sj0lj-de358.origin-ci-int-aws.dev.rhcloud.com on 172.30.0.10:53: read udp 10.129.43.239:44972->172.30.0.10:53: i/o timeout
```

tests/integration/devfile/cmd_devfile_push_test.go
dharmit: Makes sense. :+1:

PR #4430, opened 2021-02-12
Fixes no default build command error.

### LINKED ISSUE #4363: build command should be optional
/kind bug

**Operating System:**

**Output of `odo version`:**

devfile without default build command

```
▶ odo push

Validation
 ✗  Validating the devfile [222784ns]
 ✗  Failed to start component with name java-jboss-eap-xp. Error: Failed to create the component: failed to validate devfile build and run commands:
there should be exactly one default command for command group build, currently, there is no default command
```

## Expected behavior
Odo should not fail in this case. default build command should be optional. 
In that case, should simply skip the build step. There are valid use cases for this. For example, some of the interpreted languages don't require a build step, or some frameworks with hot reload can handle rebuilds themself.

 
/bug 
/priority high

mik-dass:
**What type of PR is this?**

/kind bug

It introduces two error types called NoDefaultForGroup and MoreDefaultForGroup.

Fixes #4363 

**PR acceptance criteria**:

- [X] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Create a component and remove the `build` kind command from the devfile.
- `odo push` should pass.

### DIFF
--- a/pkg/devfile/adapters/common/command.go
+++ b/pkg/devfile/adapters/common/command.go
@@ -147,15 +147,15 @@ func validateCommandsForGroup(data data.DevfileData, groupType devfilev1.Command
 				defaultCommandCount++
 			}
 		}
-	} else {
+	} else if len(commands) == 1 {
 		// if there is only one command, it is the default command for the group
 		defaultCommandCount = 1
 	}
 
 	if defaultCommandCount == 0 {
-		return fmt.Errorf("there should be exactly one default command for command group %v, currently there is no default command", groupType)
+		return NoDefaultForGroup{Group: groupType}
 	} else if defaultCommandCount > 1 {
-		return fmt.Errorf("there should be exactly one default command for command group %v, currently there is more than one default command", groupType)
+		return MoreDefaultForGroup{Group: groupType}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/devfile/adapters/common/command.go
```diff
@@ -147,29 +147,42 @@ func validateCommandsForGroup(data data.DevfileData, groupType devfilev1.Command
...
 
 	return nil
 }
 
 // GetBuildCommand iterates through the components in the devfile and returns the build command
 func GetBuildCommand(data data.DevfileData, devfileBuildCmd string) (buildCommand devfilev1.Command, err error) {
+	supportedCommand, err := getCommand(data, devfileBuildCmd, devfilev1.BuildCommandGroupKind)
+	// since build command is not necessary
+	// check if we can ignore it
+	if _, ok := err.(NoDefaultForGroup); ok {
```
dharmit, line 169: Should there be an `else` block covering the case when error is something other than `NoDefaultForGroup`?

pkg/devfile/adapters/common/command.go
```diff
@@ -147,29 +147,42 @@ func validateCommandsForGroup(data data.DevfileData, groupType devfilev1.Command
...
+	return supportedCommand, err
 }
 
 // GetDebugCommand iterates through the components in the devfile and returns the debug command
 func GetDebugCommand(data data.DevfileData, devfileDebugCmd string) (debugCommand devfilev1.Command, err error) {
-	return getCommand(data, devfileDebugCmd, devfilev1.DebugCommandGroupKind)
+	supportedCommand, err := getCommand(data, devfileDebugCmd, devfilev1.DebugCommandGroupKind)
+	// since debug command is not necessary
+	// check if we can ignore it
+	if _, ok := err.(NoDefaultForGroup); ok {
```
dharmit, line 181: Same question - should there be an `else` block?

dharmit (commented):
Just a question. Looks good otherwise.

pkg/devfile/adapters/common/command.go
mik-dass: Updated the logic

pkg/devfile/adapters/common/command.go
mik-dass: same as https://github.com/openshift/odo/pull/4430#discussion_r576113341

dharmit (approved):
/approve
/lgtm

PR #4456, opened 2021-02-24
Add telemetry consent to preference

### LINKED ISSUE #1236: End-user telemetry data of odo features
[kind/epic]
[milestone/backlog]

As a developer or product manager on the `odo` team, I would like to get have some telemetry data to better understand how many users are using individual odo features.

This will help the team prioritize their efforts in the highest impact areas and also to figure out what some areas aren't used as much as expected. This will lead to the end-user having a better overall experience with the too. Of course, the end-user should be able to opt-out of this feature.

valaparthvi:
**What type of PR is this?**
 /kind feature

This PR adds telemetry consent to preference, so that user can set and unset their telemetry preference.

Partly Fixes #1236 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

1. `odo preference --help` and check if ConsentTelemetry appears in the help.
2. `odo preference set ConsentTelemetry true` and check if the preference was set correctly.
3. `odo preference unset ConsentTelemetry` and check if the preference was unset.
4. `odo preference view` and check if ConsentTelemetry is present in the output.

Note that this does not yet collect data, this just adds `ConsentTelemetry` to preferences.

### DIFF
--- a/pkg/odo/cli/preference/set.go
+++ b/pkg/odo/cli/preference/set.go
@@ -69,6 +69,7 @@ func (o *SetOptions) Run() (err error) {
 
 	if !o.configForceFlag {
 		if isSet := cfg.IsSet(o.paramName); isSet {
+			// TODO: could add a logic to check if the new value set by the user is not same as the current value
 			if !ui.Proceed(fmt.Sprintf("%v is already set. Do you want to override it in the config", o.paramName)) {
 				log.Info("Aborted by the user")
 				return nil
--- a/pkg/odo/cli/preference/set.go
+++ b/pkg/odo/cli/preference/set.go
@@ -30,6 +30,7 @@ var (
    %[1]s %[6]s 30
    %[1]s %[7]s true
    %[1]s %[8]s docker
+   %[1]s %[9]s true
 	`)
 )
 
@@ -94,7 +95,7 @@ func NewCmdSet(name, fullName string) *cobra.Command {
 		Example: fmt.Sprintf(fmt.Sprint("\n", setExample), fullName,
 			preference.UpdateNotificationSetting, preference.NamePrefixSetting,
 			preference.TimeoutSetting, preference.BuildTimeoutSetting, preference.PushTimeoutSetting,
-			preference.ExperimentalSetting, preference.PushTargetSetting),
+			preference.ExperimentalSetting, preference.PushTargetSetting, preference.ConsentTelemetrySetting),
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 2 {
 				return fmt.Errorf("please provide a parameter name and value")
--- a/pkg/odo/cli/preference/view.go
+++ b/pkg/odo/cli/preference/view.go
@@ -68,6 +68,7 @@ func (o *ViewOptions) Run() (err error) {
 	fmt.Fprintln(w, "Experimental", "\t", showBlankIfNil(cfg.OdoSettings.Experimental))
 	fmt.Fprintln(w, "PushTarget", "\t", showBlankIfNil(cfg.OdoSettings.PushTarget))
 	fmt.Fprintln(w, "Ephemeral", "\t", showBlankIfNil(cfg.OdoSettings.Ephemeral))
+	fmt.Fprintln(w, "ConsentTelemetry", "\t", showBlankIfNil(cfg.OdoSettings.ConsentTelemetry))
 
 	w.Flush()
 	return
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -114,7 +114,7 @@ var RegistryCacheTimeDescription = fmt.Sprintf("For how long (in minutes) odo wi
 var EphemeralDescription = fmt.Sprintf("If true odo will create a emptyDir volume to store source code (Default: %t)", DefaultEphemeralSettings)
 
 //TelemetryConsentDescription adds a description for TelemetryConsentSetting
-var ConsentTelemetryDescription = fmt.Sprintf("If true odo will collect telemetry for the user's odo usage (Default: %t)", DefaultConsentTelemetrySetting)
+var ConsentTelemetryDescription = fmt.Sprintf("If true odo will collect telemetry for the user's odo usage (Default: %t)\n\t\t    For more information: https://developers.redhat.com/article/tool-data-collection", DefaultConsentTelemetrySetting)
 
 // This value can be provided to set a seperate directory for users 'homedir' resolution
 // note for mocking purpose ONLY
@@ -424,6 +424,7 @@ func (c *PreferenceInfo) SetConfiguration(parameter string, value string) error 
 			}
 			c.OdoSettings.UpdateNotification = &val
 
+		//	TODO: should we add a validator here? What is the use of nameprefix?
 		case "nameprefix":
 			c.OdoSettings.NamePrefix = &value
 
--- a/pkg/preference/preference.go
+++ b/pkg/preference/preference.go
@@ -90,6 +90,12 @@ const (
 
 	// DefaultEphemeralSettings is a default value for Ephemeral preference
 	DefaultEphemeralSettings = true
+
+	// ConsentTelemetrySettings specifies if the user consents to telemetry
+	ConsentTelemetrySetting = "ConsentTelemetry"
+
+	// DefaultConsentTelemetry is a default value for ConsentTelemetry preference
+	DefaultConsentTelemetrySetting = false
 )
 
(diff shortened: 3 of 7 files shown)


pkg/odo/cli/preference/view.go
```diff
@@ -68,6 +68,7 @@ func (o *ViewOptions) Run() (err error) {
 	fmt.Fprintln(w, "Experimental", "\t", showBlankIfNil(cfg.OdoSettings.Experimental))
 	fmt.Fprintln(w, "PushTarget", "\t", showBlankIfNil(cfg.OdoSettings.PushTarget))
 	fmt.Fprintln(w, "Ephemeral", "\t", showBlankIfNil(cfg.OdoSettings.Ephemeral))
+	fmt.Fprintln(w, "ConsentTelemetry", "\t", showBlankIfNil(cfg.OdoSettings.ConsentTelemetry))
```
kadel, line 71: Could we change it from`ConsetTelemetry` to `SendTelemetry`? Maybe it is just me, but it sounds a little bit better 😇

pkg/preference/preference.go
```diff
@@ -107,6 +113,9 @@ var RegistryCacheTimeDescription = fmt.Sprintf("For how long (in minutes) odo wi
 // EphemeralDescription adds a description for EphemeralSourceVolume
 var EphemeralDescription = fmt.Sprintf("If true odo will create a emptyDir volume to store source code (Default: %t)", DefaultEphemeralSettings)
 
+//TelemetryConsentDescription adds a description for TelemetryConsentSetting
+var ConsentTelemetryDescription = fmt.Sprintf("If true odo will collect telemetry for the user's odo usage (Default: %t)", DefaultConsentTelemetrySetting)
```
kadel, line 117: Not sure how long the description can be. If it is not going to look weird in console,  it would be good to include a link to https://developers.redhat.com/article/tool-data-collection

kadel:
This is going in the right direction. I have just added a few comments that are only cosmetics.

pkg/odo/cli/preference/view.go
valaparthvi: `ConsentTelemetry` did seem intuitive to me at first, but I guess `SendTelemetry` makes sense and sounds good too. I'll change that!

kadel: Now I just realized that it might be good to use the same name as CRC.
I just checked that they have `ConsentTelemetry` 😇 So lets keep it :-)

pkg/odo/cli/preference/set.go
```diff
@@ -30,6 +30,7 @@ var (
    %[1]s %[6]s 30
    %[1]s %[7]s true
    %[1]s %[8]s docker
+   %[1]s %[9]s true
```
valaparthvi, line 33: I don't exactly like this kind of formatting because if the place of any preference changes [here](https://github.com/openshift/odo/pull/4456/files#diff-520872630a47e18df6266f58252a6fd10fe43eba71f49d23d346b2a76ec1b3baR99) without the change being made in the template, it can cause a misunderstanding in the help. I did not find an elegant enough solution but I would like if we can do something about it.

dharmit: You mean if someone changes the order of
```go
preference.ExperimentalSetting, preference.PushTargetSetting, preference.ConsentTelemetrySetting
```
to 
```go
preference.PushTargetSetting, preference.ConsentTelemetrySetting, preference.ExperimentalSetting
```
but doesn't make a change in this template, the `odo preference set -h` would mean something entirely different?

I think that's a valid point. But what's the solution? One way it could be made less confusing is by adding code comments.

valaparthvi: Yes, exactly that! I was hoping to find a solution more on the line where we could modify the template. I'll add a TODO for now.

girishramnani: I would suggest to use package text/template.
That way you can write something like below e.g.

```

{{ .PushTargetSetting }} {{ .PushTargetSettingDefault }}

```

and this is quite descriptive and non positional.

you can consider looping over these values to make it even more cooler?

girishramnani: there is ktemplate package which we use at some places? Does that work similarly? Check that too

valaparthvi: Currently, this piece of code is using ktemplates. But I do like the idea of text/template, it's exactly what I was looking for. I'll see if we can do something like that with ktemplate, if not, I'll go ahead with text/template.

valaparthvi: I raised another [PR](https://github.com/openshift/odo/pull/4471/commits/99b535e8cf7e461078646dd98e267f1d9ddf20b5) because this change seemed somewhat out of scope for this PR. I realized there might be other places requiring the same kind of change, so I plan on adding them later on to the same PR.

valaparthvi:
Seeing integration-e2e failure due to the following - 
```
Setting current dir to: 
• Failure in Spec Setup (JustBeforeEach) [307.297 seconds]
odo service command tests for OperatorHub
/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:17
  When linking devfile component with Operator backed service
/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:372
should fail if service name doesn't adhere to <service-type>/<service-name> format [JustBeforeEach]
/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:391

Timeout after 5 minutes

    /go/src/github.com/openshift/odo/tests/helper/helper_generic.go:44
```

Looks like other PRs are seeing it too.

dharmit:
> Seeing integration-e2e failure due to the following -
> 
> ```
> Setting current dir to: 
> • Failure in Spec Setup (JustBeforeEach) [307.297 seconds]
> odo service command tests for OperatorHub
> /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:17
>   When linking devfile component with Operator backed service
> /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:372
> should fail if service name doesn't adhere to <service-type>/<service-name> format [JustBeforeEach]
> /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:391
> 
> Timeout after 5 minutes
> 
>     /go/src/github.com/openshift/odo/tests/helper/helper_generic.go:44
> ```

Looking at the error trail, it seems like Service Binding Operator is not getting installed. We check for existence of `etcdoperator` and `service-binding-operator` in `preSetup()` function.

kadel:
> > Seeing integration-e2e failure due to the following -
> > ```
> > Setting current dir to: 
> > • Failure in Spec Setup (JustBeforeEach) [307.297 seconds]
> > odo service command tests for OperatorHub
> > /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:17
> >   When linking devfile component with Operator backed service
> > /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:372
> > should fail if service name doesn't adhere to <service-type>/<service-name> format [JustBeforeEach]
> > /go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:391
> > 
> > Timeout after 5 minutes
> > 
> >     /go/src/github.com/openshift/odo/tests/helper/helper_generic.go:44
> > ```
> 
> Looking at the error trail, it seems like Service Binding Operator is not getting installed. We check for existence of `etcdoperator` and `service-binding-operator` in `preSetup()` function.

Is someone looking into this?

dharmit:
> Is someone looking into this?

Not yet. Latest CRC is running 4.6.15. I know we have other places to spin up 4.7 cluster. I will do that as soon as I can.

girishramnani:
This timeout after 5 min error is blocking all PRs, please Prioritize this issue.
And if it’s hard to solve let’s comment it out for now

dharmit:
> This timeout after 5 min error is blocking all PRs, please Prioritize this issue.
> And if it’s hard to solve let’s comment it out for now

I'm not in favour of commenting tests because odo already doesn't have a good enough test coverage. What we're suggesting to comment here is the entire Operator Hub integration test suite! I'd rather prefer putting a `/hold` on the PRs so that they don't get triggered on CI and troubleshoot the failure.

Yes it blocks PRs from getting tested/merged, but it's not as bad as code going in without being tested. At the very least, we can afford to hold PRs for some time and figure out why something that was working fine in the past has started breaking! In this case, it's moving from 4.6 to 4.7 (which no developer is using as latest CRC is still based on 4.6.15) that has broke things on CI.

Issue #4160, opened 2020-10-27
odo service create should add operator CR to the devfile


girishramnani:
/kind user-story

## User Story

As a developer working with odo, when I add a new service to the application (like DB), I want to be able to easily share service definitions with my team so they can start the application with the service in the same way as me. 

## Acceptance Criteria 
- [x] `odo service create` should not create any services on cluster (#4650)
- [x] `odo service create` should store service definition in devfile as `kubernetes` component (`ServiceInstance` for SericeCatalog or CR for operator backed service) (#4465)
- [ ] `odo service delete` should delete the service from devfile (remove appropriate `kubernetes` component). Deletion from the cluster will happen once `odo push` is executed (similar to how we handle URLs).
- [ ] `odo service list` should list services and their status. Important is to differentiate between statues like DeletedLocally, DeletedRemotely etc... (Similarly to what odo does for URLs)
- [ ] ~`odo link` should add Service Binding information to devfile.yaml and `odo unlink` should remove it from the file. Subsequent `odo push` should create/delete the actual link (a Service Binding).~ (being tracked by #4208)

## Links 
- Feature Request: https://github.com/openshift/odo/issues/4159
- Related issue: 

/kind user-story

dharmit:
> It seems the Uri holds a link to the yaml or we can provide the yaml as a string in `Inlined`. both seem mutually exclusive.

I'm not sure what purpose does having both `Inlined` and `Uri` serve. When I read this, I think of `Inlined` as a `string` representation of the CRD used to spin up a CR. It would be the same as what we have right now when doing `odo service create <operator-name>/<cr-name> --dry-run`. Then what is `Uri`/`reference` in this case?

What is odo supposed to do with `name`? At the moment, we have `odo service create etcdoperator.v0.9.4/EtcdCluster newetcd` which creates a new service from the CR `EtcdCluster` provided by the Operator `etcdoperator.v0.9.4` and names it as `newetcd`. Should odo use the value in `name` for the newly created service?

> devfile/api#36

This is a proposal, IIUC. Do we have to implement things based on this proposal? Or do we use `Kubernetes`/`OpenShift` lingo instead? If we should be using `Kubernetes`/`OpenShift` based on current devfile schema, what happens when this change goes in? We make changes to work with the new terminology? 

Few other questions (more along the lines of "thinking out loud"):
* What's the timeline for this? It seems more like an epic than a user story to me.
* What about Service Catalog backed services? Are we planning to support those in the new approach?
[...]

Issue #4105, opened 2020-10-09
storage create should have the option to specify a container


maysunfaisal:
/kind user-story

## User Story
As a user when I create storage via `odo storage create mynewstorage --size 4Gi --path /some/path`; it creates the volume mount with the specified storage for all the containers if the devfile has more than one container. It should have a flag to specify a container name and should mount storage for only that container.

## Acceptance Criteria 
- [x] It should have a flag for container
- [x] It should validate to check if the container mentioned is valid
- [x] It should mount storage for only the container specified

## Links 
- Feature Request: 
- Related issue: 

/kind user-story

maysunfaisal:
Here is my odo command
```
Maysuns-MacBook-Pro:nodejs-ex maysun$ odo storage create test --path /test
 ✓  Added storage test to no

Please use `odo push` command to make the storage accessible to the component
```

and the devfile with more than one container:
```
components:
- name: runtime
  container:
    endpoints:
    - name: 3000/tcp
      targetPort: 3000
    env:
    - name: FOO
      value: bar
    image: quay.io/eclipse/che-nodejs10-ubi:nightly
    memoryLimit: 1024Mi
    mountSources: true
    volumeMounts:
    - name: firstvol
      path: /data
    - name: test
      path: /test
- name: runtime2
  container:
    image: quay.io/eclipse/che-nodejs10-ubi:nightly
    memoryLimit: 1024Mi
    volumeMounts:
    - name: firstvol
      path: /data
    - name: secondvol
      path: /data2
    - name: test
      path: /test
- name: firstvol
  volume: {}
- name: secondvol
  volume:
    size: 3Gi
- name: test
  volume:
    size: 1Gi
```

mohammedzee1000:
I am currently walking through `url create` and `storage create` and will get back with my findings shortly

mohammedzee1000:
Ok, from what I have read, this might be a little bit more complicated than anticipated.
It is easy enough to add a flag for the container with cobra, and even ensure it gets assigned to LocalStorage

Like so
```golang
	o.storage = localConfigProvider.LocalStorage{
		Name: o.storageName,
		Size: o.storageSize,
		Path: o.storagePath,
		Container: o.container,
	}
```
But doing so is not enough. The creation of the storage happens on 

https://github.com/openshift/odo/blob/ba7161940c735caffaa9bd49e731dd15b8e9a61f/pkg/odo/cli/storage/create.go#L80

The function is present in vendor here

[...]

PR #4465, opened 2021-03-01
Add and remove service info from devfile

dharmit:
**What type of PR is this?**
/kind feature
/kind code-refactoring

1. Implements the feature to add CRD and ServiceInstance information into `devfile.yaml`
2. Refactors `odo service create` and `odo service delete` CLI packages to use interface

Addresses acceptance criteria discussed in:
* https://github.com/openshift/odo/issues/4160#issuecomment-775838428
* https://github.com/openshift/odo/issues/4160#issuecomment-790577742

It also makes changes to `pkg/envinfo/storage.go` to be compatible with changes introduced in `devfile/library`.

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

- `odo service create` should add CRD and ServiceInstance information to `devfile.yaml`.
- `odo service delete` should remove the CRD information from `devfile.yaml`.
- `odo storage create` and `odo storage delete` should work as they work in current `master` branch. Code for these commands was changed to be compatible with changes introduced in `devfile/library`.


dharmit:
Unit test failed locally with following error:
```sh
?   	github.com/openshift/odo/cmd/cli-doc	[no test files]
?   	github.com/openshift/odo/cmd/odo	[no test files]
ok  	github.com/openshift/odo/pkg/application	(cached)
ok  	github.com/openshift/odo/pkg/application/labels	(cached)
?   	github.com/openshift/odo/pkg/auth	[no test files]
ok  	github.com/openshift/odo/pkg/catalog	(cached)
ok  	github.com/openshift/odo/pkg/component	(cached)
ok  	github.com/openshift/odo/pkg/component/labels	(cached)
ok  	github.com/openshift/odo/pkg/config	(cached)
ok  	github.com/openshift/odo/pkg/debug	(cached)
ok  	github.com/openshift/odo/pkg/devfile/adapters	(cached)
 ⚠  there should be exactly one default command for command group build, currently there is no default command
 ⚠  there should be exactly one default command for command group debug, currently there is no default command
--- FAIL: TestGetCommandFromFlag (0.00s)
    --- FAIL: TestGetCommandFromFlag/Case_9:_Valid_devfile_with_invalid_composite_commands (0.00s)
        command_test.go:803: TestGetCommand unexpected error for command: build wantErr: true err: <nil>
 •  Executing command1 command ""  ...
 ✓  Executing command1 command "" [1ms]
 •  Executing command2 command ""  ...
 ✓  Executing command2 command "" [3ms]
 •  Executing command1 command ""  ...
 ✗  Executing command1 command "" [734461ns]
 •  Executing command1 command ""  ...
 •  Executing command2 command ""  ...
 ✓  Executing command1 command "" [6ms]
 ✓  Executing command2 command "" [6ms]
 •  Executing command2 command ""  ...
 •  Executing command1 command ""  ...
 ✗  Executing command2 command "" [1ms]
 ✗  Executing command1 command "" [1ms]
 •  Executing command1 command ""  ...
 ✓  Executing command1 command "" [448100ns]
 •  Executing command1 command ""  ...
 ✓  Executing command1 command "" [478143ns]
 •  Executing command2 command ""  ...
 ✓  Executing command2 command "" [519094ns]
 •  Executing command1 command ""  ...
 •  Executing command1 command ""  ...
 ✓  Executing command1 command "" [5ms]
 ✓  Executing command1 command "" [1ms]
 •  Executing command2 command ""  ...
 ✓  Executing command2 command "" [577013ns]
FAIL
FAIL	github.com/openshift/odo/pkg/devfile/adapters/common	0.205s
?   	github.com/openshift/odo/pkg/devfile/adapters/docker	[no test files]
ok  	github.com/openshift/odo/pkg/devfile/adapters/docker/component	(cached)
ok  	github.com/openshift/odo/pkg/devfile/adapters/docker/storage	(cached)
ok  	github.com/openshift/odo/pkg/devfile/adapters/docker/utils	(cached)
?   	github.com/openshift/odo/pkg/devfile/adapters/kubernetes	[no test files]
ok  	github.com/openshift/odo/pkg/devfile/adapters/kubernetes/component	(cached)
ok  	github.com/openshift/odo/pkg/devfile/adapters/kubernetes/storage	(cached)
ok  	github.com/openshift/odo/pkg/devfile/adapters/kubernetes/utils	(cached)
?   	github.com/openshift/odo/pkg/devfile/convert	[no test files]
ok  	github.com/openshift/odo/pkg/devfile/validate	(cached)
--- FAIL: TestValidateCommands (0.00s)
    --- FAIL: TestValidateCommands/Case_8:_Duplicate_commands (0.00s)
        commands_test.go:197: TestValidateAction unexpected error: <nil>
FAIL
FAIL	github.com/openshift/odo/pkg/devfile/validate/generic	0.089s
ok  	github.com/openshift/odo/pkg/envinfo	(cached)
ok  	github.com/openshift/odo/pkg/kclient	(cached)
?   	github.com/openshift/odo/pkg/kclient/fake	[no test files]
ok  	github.com/openshift/odo/pkg/lclient	(cached)
?   	github.com/openshift/odo/pkg/localConfigProvider	[no test files]
?   	github.com/openshift/odo/pkg/log	[no test files]
?   	github.com/openshift/odo/pkg/log/fidget	[no test files]
?   	github.com/openshift/odo/pkg/machineoutput	[no test files]
ok  	github.com/openshift/odo/pkg/notify	(cached) [no tests to run]
ok  	github.com/openshift/odo/pkg/occlient	(cached)
?   	github.com/openshift/odo/pkg/odo/cli	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/application	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/catalog	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/catalog/describe	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/catalog/list	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/catalog/search	[no test files]
ok  	github.com/openshift/odo/pkg/odo/cli/catalog/util	(cached)
?   	github.com/openshift/odo/pkg/odo/cli/component	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/component/ui	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/config	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/debug	[no test files]
ok  	github.com/openshift/odo/pkg/odo/cli/env	(cached)
?   	github.com/openshift/odo/pkg/odo/cli/login	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/logout	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/plugins	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/preference	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/project	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/registry	[no test files]
ok  	github.com/openshift/odo/pkg/odo/cli/registry/util	(cached)
ok  	github.com/openshift/odo/pkg/odo/cli/service	(cached)
ok  	github.com/openshift/odo/pkg/odo/cli/service/ui	(cached)
ok  	github.com/openshift/odo/pkg/odo/cli/storage	(cached)
?   	github.com/openshift/odo/pkg/odo/cli/ui	[no test files]
?   	github.com/openshift/odo/pkg/odo/cli/url	[no test files]
ok  	github.com/openshift/odo/pkg/odo/cli/utils	(cached)
?   	github.com/openshift/odo/pkg/odo/cli/version	[no test files]
?   	github.com/openshift/odo/pkg/odo/genericclioptions	[no test files]
ok  	github.com/openshift/odo/pkg/odo/util	(cached)
ok  	github.com/openshift/odo/pkg/odo/util/completion	(cached)
ok  	github.com/openshift/odo/pkg/odo/util/experimental	(cached)
ok  	github.com/openshift/odo/pkg/odo/util/pushtarget	(cached)
ok  	github.com/openshift/odo/pkg/odo/util/validation	(cached)
ok  	github.com/openshift/odo/pkg/preference	(cached)
ok  	github.com/openshift/odo/pkg/project	(cached)
ok  	github.com/openshift/odo/pkg/secret	(cached)
ok  	github.com/openshift/odo/pkg/service	(cached)
ok  	github.com/openshift/odo/pkg/storage	(cached)
ok  	github.com/openshift/odo/pkg/storage/labels	(cached)
ok  	github.com/openshift/odo/pkg/sync	(cached)
?   	github.com/openshift/odo/pkg/sync/mock	[no test files]
ok  	github.com/openshift/odo/pkg/testingutil	(cached)
?   	github.com/openshift/odo/pkg/testingutil/filesystem	[no test files]
ok  	github.com/openshift/odo/pkg/url	(cached)
ok  	github.com/openshift/odo/pkg/url/labels	(cached)
ok  	github.com/openshift/odo/pkg/util	(cached)
?   	github.com/openshift/odo/pkg/version	[no test files]
ok  	github.com/openshift/odo/pkg/watch	(cached)
FAIL
make: *** [Makefile:156: test] Error 1
```

pkg/config/config_test.go
```diff
@@ -525,18 +525,9 @@ func TestSetDevfileConfiguration(t *testing.T) {
...
-							Id: "devbuild",
-							CommandUnion: devfilev1.CommandUnion{
-								Exec: &devfilev1.ExecCommand{
-									WorkingDir: "/projects/nodejs-starter",
-								},
-							},
-						},
-					},
-					Components: []devfilev1.Component{
+				Data: func() data.DevfileData {
```
dharmit, line 528: @mik-dass @kadel I have made this change based on what seemed like the best option to me. If you have any suggestions, let me know. There are about 60 to 70 occurrences of this change across the unit tests and code.

dharmit:
Holding since creating a service from outside the component directory is resulting in panic instead of a proper error message:
```sh
$ odo service create etcdoperator.v0.9.4-clusterwide/EtcdCluster 
Deploying service "example" of type: "EtcdCluster"
 ✗  Deploying service [4ms]
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb0 pc=0x1439e65]

goroutine 1 [running]:
github.com/openshift/odo/pkg/envinfo.(*EnvSpecificInfo).AddServiceToDevfile(0x0, 0xc0006c2000, 0xba, 0xc000677830, 0x7, 0xba, 0x0)
        /home/dshah/go/src/github.com/openshift/odo/pkg/envinfo/service.go:29 +0x125
github.com/openshift/odo/pkg/service.CreateOperatorService(0xc0005b99e0, 0x0, 0xc000677830, 0x7, 0xc000a30a2d, 0x18, 0xc000676d86, 0x7, 0xc000a30a20, 0xc, ...)
        /home/dshah/go/src/github.com/openshift/odo/pkg/service/service.go:117 +0x12e
github.com/openshift/odo/pkg/odo/cli/service.(*OperatorBackend).RunServiceCreate(0xc0004403c0, 0xc000835e40, 0x0, 0x0)
        /home/dshah/go/src/github.com/openshift/odo/pkg/odo/cli/service/backends.go:223 +0x2a3
github.com/openshift/odo/pkg/odo/cli/service.(*CreateOptions).Run(0xc000835e40, 0x0, 0x0)
        /home/dshah/go/src/github.com/openshift/odo/pkg/odo/cli/service/create.go:168 +0x43
github.com/openshift/odo/pkg/odo/genericclioptions.GenericRun(0x232a180, 0xc000835e40, 0xc00075f180, 0xc00076e950, 0x1, 0x1)
        /home/dshah/go/src/github.com/openshift/odo/pkg/odo/genericclioptions/runnable.go:31 +0x13c
github.com/openshift/odo/pkg/odo/cli/service.NewCmdServiceCreate.func1(0xc00075f180, 0xc00076e950, 0x1, 0x1)
        /home/dshah/go/src/github.com/openshift/odo/pkg/odo/cli/service/create.go:194 +0x5e
github.com/spf13/cobra.(*Command).execute(0xc00075f180, 0xc00076e930, 0x1, 0x1, 0xc00075f180, 0xc00076e930)
        /home/dshah/go/src/github.com/openshift/odo/vendor/github.com/spf13/cobra/command.go:830 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0xc000612c80, 0x20beed0, 0xc00005c360, 0x3)
        /home/dshah/go/src/github.com/openshift/odo/vendor/github.com/spf13/cobra/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
        /home/dshah/go/src/github.com/openshift/odo/vendor/github.com/spf13/cobra/command.go:864
main.main()
        /home/dshah/go/src/github.com/openshift/odo/cmd/odo/odo.go:67 +0x330
```

/hold

dharmit:
Unit test failure:
```
 --- FAIL: TestGetCommandFromFlag (0.00s)
    --- FAIL: TestGetCommandFromFlag/Case_9:_Valid_devfile_with_invalid_composite_commands (0.00s)
        command_test.go:803: TestGetCommand unexpected error for command: build wantErr: true err: <nil>
 •  Executing command1 command ""  ...

 ✓  Executing command1 command "" [1ms]
 •  Executing command2 command ""  ...

 ✓  Executing command2 command "" [814921ns]
 •  Executing command1 command ""  ...

 ✓  Executing command1 command "" [1ms]
 •  Executing command2 command ""  ...

 ✓  Executing command1 command "" [2ms]

 ✓  Executing command2 command "" [853597ns]
FAIL
FAIL	github.com/openshift/odo/pkg/devfile/adapters/common	0.101s 
```

pkg/config/config_test.go
mik-dass: I feel this might make debugging tests harder since we are ignoring the errors. I feel the best approach would be to mock the `DevfileData` interface using gomock. That way we can remove a lot of the unnecessary struct values from functions like these but that probably should be done as a separate PR.

dharmit: About ignoring the error, I'm doing this because I don't feel like it's worth doing that in unit tests. But I don't think of it as a good idea so not defending it here.

About mocking `DevfileData`, I think something like that should even exist in `devfile/library` repo. Let me see or ask around.

dharmit:
@kadel @mik-dass @girishramnani @valaparthvi 

This PR is ready for review now. TBH, there was a lot of tangential stuff I had to address while working on this, so apologies in advance if I missed something obvious. 😄 

Also, I have skipped adding integration tests but added unit tests. I'd prefer not to add more integration tests unless it's really, really required.

pkg/config/config_test.go
maysunfaisal: @dharmit @mik-dass 

1. yeah probably should not ignore err, makes it difficult to debug
2. to alleviate this PR, I can revert the changes to TestDevfileData and have a discussion around replacing it with gomock and whether this should stay in devfile/library

what do you guys think?

mik-dass: @maysunfaisal I feel that would be great :+1:

maysunfaisal: @mik-dass @dharmit I've not used gomock before for interfaces. So I can have a look and see the pain points that is being discussed here and what it means for devfile/library consumers.

In the mean time, I've put up a PR https://github.com/devfile/library/pull/72 to revert the test data deletion so that it doesn't affect odo as much. It has been approved and I will leave it up to @dharmit to merge(pls feel free) the PR and use the devfile/library latest in this PR if needed.

dharmit: @maysunfaisal I don't think that it's required to revert the changes just because we don't have mocking mechanism in place right now. My PR has gone ahead with making the changes that would make it compatible with changes introduced in https://github.com/devfile/library/pull/62.

Sure the changes I have made might not be the best approach, but we can improve that once mocking is available in `devfile/library` repo.

Merging https://github.com/devfile/library/pull/72 would be akin to taking a step backwards, and I'd like to avoid it unless others feel this is the right thing to do.

WDYT @mik-dass @kadel

dharmit:
[Test failures](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4465/pull-ci-openshift-odo-master-v4.7-integration-e2e/1372196041264205824) are showing up at places that have not been modified by this PR.

/test v4.7-integration-e2e

pkg/envinfo/service.go
```diff
@@ -0,0 +1,57 @@
+package envinfo
```
kadel, line 1: why is this in `envinfo`? THis package is handling `env.yaml` file, the function in this file are working on top of `devfile.yaml`

dharmit: That's because I was going through the way things are done for adding URL & storage information to devfile. It's being done from `envinfo` package right now.

dharmit: But since the code in this PR is not dependent on `env.yaml` in any way, I get why putting it here would be a bad idea. In that case, putting it under `pkg/service` seems like the way to me. WDYT?

Just looked at this again and currently "odo service create" calls `AddServiceToDevfile` from `pkg/service/service.go` whereas `odo service delete` calls `DeleteServiceFromDevfile` from the newly created `pkg/odo/cli/service/backends.go`. And it doesn't delete a Service Catalog based service from devfile.

dharmit:
> @kadel @mik-dass @girishramnani @valaparthvi
> 
> This PR is ready for review now. TBH, there was a lot of tangential stuff I had to address while working on this, so apologies in advance if I missed something obvious. smile
> 
> Also, I have skipped adding integration tests but added unit tests. I'd prefer not to add more integration tests unless it's really, really required.

Things that still need to be addressed:
- [ ] deletion of service catalog based service from devfile
- [ ] fix Operator Hub integration tests to execute from component directory
- [ ] move the functions to add and remove service info from devfile to a different package (https://github.com/openshift/odo/pull/4465#pullrequestreview-615053035)

pkg/config/config_test.go
mik-dass: > EDIT: Not sure how I missed this comment - #4465 (comment). I still feel like it would be a step backwards. Instead of ignoring the error, I could add a t.Error. Wouldn't that be OK? @mik-dass @maysunfaisal

I don't mind the `t.Error()` approach for now. But I also OK with https://github.com/devfile/library/pull/72 as it gives us more flexibility.

pkg/envinfo/service.go
mik-dass: I feel we should just rename the package name to something which indicates `env.yaml + devfile.yaml`as that might help with abstraction.

kadel: it would help if we separate abstraction, devfile and env.yaml are two completely different things they should not live in one package

mik-dass: But devfile already has the library. I don't feel we need a separate package to add `odo` concepts like service, url, storage etc to it. Since both are some sort of local data storage, combining them makes more sense to me. The outside packages like CLI don't need to know where the data comes from locally. That's how the URL create, delete etc for local are designed currently.

pkg/envinfo/storage.go
```diff
@@ -55,16 +55,12 @@ func (ei *EnvInfo) GetStorage(name string) (*localConfigProvider.LocalStorage, e
...
 func (ei *EnvInfo) CreateStorage(storage localConfigProvider.LocalStorage) error {
-	err := ei.devfileObj.Data.AddVolume(devfilev1.Component{
-		Name: storage.Name,
-		ComponentUnion: devfilev1.ComponentUnion{
-			Volume: &devfilev1.VolumeComponent{
-				Volume: devfilev1.Volume{
-					Size: storage.Size,
-				},
-			},
+	err := ei.devfileObj.Data.AddVolumeMounts(ei.componentSettings.Name, []devfilev1.VolumeMount{
```
maysunfaisal, line 58: this new func adds only volume mounts to the container component in the devfile.

you need to make a new call to add the volume component which was inherently done by the prev function. So just need to call `AddComponents(components []v1.Component) error`

pkg/envinfo/storage.go
```diff
@@ -119,7 +115,7 @@ func (ei *EnvInfo) ListStorage() ([]localConfigProvider.LocalStorage, error) {
 
 // DeleteStorage deletes the storage with the given name
 func (ei *EnvInfo) DeleteStorage(name string) error {
-	err := ei.devfileObj.Data.DeleteVolume(name)
+	err := ei.devfileObj.Data.DeleteVolumeMount(name)
```
maysunfaisal, line 118: This only deletes the volume mount, you need to delete the volume component which was inherently done by the prev function.

So just need to call `DeleteComponent(name string) error`

pkg/envinfo/storage.go
```diff
@@ -133,7 +129,7 @@ func (ei *EnvInfo) DeleteStorage(name string) error {
 
 // GetStorageMountPath gets the mount path of the storage with the given storage name
 func (ei *EnvInfo) GetStorageMountPath(storageName string) (string, error) {
-	return ei.devfileObj.Data.GetVolumeMountPath(storageName)
+	return ei.devfileObj.Data.GetVolumeMountPath(storageName, ei.componentSettings.Name)
```
maysunfaisal, line 132: So not sure if `ei.componentSettings.Name` is correct here. 

`GetVolumeMountPath()` takes a mount name and container name because essentially devfile can have many container components and each container component can have mounts with the same name, so we need to specify which container component in the devfile do you want the mount name from..

maysunfaisal: Thinking about this a bit more, the return value of `GetVolumeMountPath` should be an array because each container component can have multiple mounts on the same volume for a container component can have

```
components:
  - name: test2
    volume:
        size: 1Gi
  - name: test
    container:
        image: xyz
        volumeMounts:
           - name: test2
             path: /path1
           - name: test2
             path: /path2
```

so this is essentially a bug with devfile/library new func too

maysunfaisal (changes requested):
Some comments around new devfile/library funcs

pkg/envinfo/service.go
kadel: The function that are here in this file, don't have anything specific to odo, expect its name.
`AddServiceToDevfile` Could be named `AddK8sComponentToDevfile`, use the same parameters and it would make sense as well (maybe even more sense)

mik-dass: > The function that are here in this file, don't have anything specific to odo, expect its name.
> `AddServiceToDevfile` Could be named `AddK8sComponentToDevfile`, use the same parameters and it would make sense as well (maybe even more sense)

Since it has nothing specific to odo and doesn't fit into the local config provider interface, I would suggest making it a utils function or a private function. I don't think we need a seperate package for it.

dharmit: > `AddServiceToDevfile` Could be named `AddK8sComponentToDevfile`

I like this suggestion. Makes more sense since from devfile point of view we're actually adding a k8s component.

> I would suggest making it a utils function or a private function. I don't think we need a seperate package for it.

If we rename it to `AddK8sComponentToDevfile`, it sure doesn't make sense to move this to `pkg/service` either. It's more generic than being a part of `pkg/service`. Which utils package would this make more sense to live in - `pkg/util` or `pkg/odo/util`? And why (asking this for better understanding on my part)?

mik-dass: It can also be part of some utils function in the `pkg/service` package as crds are mostly handled by it.

dharmit: PTAL https://github.com/openshift/odo/pull/4465/commits/f4123527811c23c04fdf97799a6dca797c510dfc. I have kept it in `pkg/service` but not a part of utils yet since I made the changes before you made a comment.

mik-dass: Looks OK to me

pkg/envinfo/storage.go
dharmit: Is there a bug on `devfile/library` repo to track this? I don't suppose I can fix this correctly without this being fixed there. Or can I? Argh! It sucks that I don't have much idea about the devfile side of odo code. Every interaction is making me feel lost. 😕

pkg/envinfo/storage.go
dharmit: Addressed in c4bc9d033. PTAL.

pkg/envinfo/storage.go
dharmit: Addressed in c4bc9d033. PTAL.

pkg/envinfo/storage.go
mik-dass: Yeah as @maysunfaisal pointed out, it should return all the mount paths for a given storage name and return a map (with container names) or array of the paths. 

> So not sure if ei.componentSettings.Name is correct here.

It is incorrect. We need to first fetch all the containers, then fetch the mount path for each container and return an array  of the path or a map. Maybe for now, we can return the first path from the first container and handle the rest when it's fixed on the library's side.

maysunfaisal: @mik-dass right now the func `GetVolumeMountPath()` takes the mount name and the container name but returns the first path.. so essentially, to make this PR compatible with current odo master, @dharmit  would just need to pass the container name or know the container name to pass to this func.

@mik-dass do you want me to correct this on devfile/library master like you suggested? ie `GetVolumeMountPath` takes a mount name and returns a map of container -> mountPath or `GetVolumeMountPath` can take the mount name and container name and return an array of mountPaths for that container component

pkg/envinfo/storage.go
maysunfaisal: looks good

pkg/envinfo/storage.go
maysunfaisal: looks good

pkg/envinfo/storage.go
maysunfaisal: Correct me if I'm wrong but FWIW, I just checked the repo and it looks like no one is using this func `(ei *EnvInfo) GetStorageMountPath(storageName string)`

mik-dass: > current odo master, @dharmit would just need to pass the container name or know the container name to pass to this func.

Yes. In a later PR, we can update and fetch all the containers and call the function for all of them as it might be mounted at a different path in a different container. That would change the return signature of the function.

> @mik-dass do you want me to correct this on devfile/library master like you suggested? ie GetVolumeMountPath takes a mount name and returns a map of container -> mountPath or GetVolumeMountPath can take the mount name and container name and return an array of mountPaths for that container component

Yes that would be great :+1: The function can return an array of all the mountPaths for that given container component.

> Correct me if I'm wrong but FWIW, I just checked the repo and it looks like no one is using this func (ei *EnvInfo) GetStorageMountPath(storageName string)

It's being used here https://github.com/openshift/odo/blob/7e9ad7bda5c35d81890413a6524dff3ef73e1d94/pkg/odo/cli/storage/delete.go#L71

pkg/envinfo/storage.go
maysunfaisal: @dharmit sorry, I've unresolved this today. Just came back to realize that this has the same problem as https://github.com/openshift/odo/pull/4465#discussion_r596964320

devfile component name != odo component name, a devfile container is a component in devfile schema

pkg/envinfo/storage.go
```diff
@@ -56,16 +56,25 @@ func (ei *EnvInfo) GetStorage(name string) (*localConfigProvider.LocalStorage, e
...
-			Path: storage.Path,
-		},
-	})
+	// Get all the containers in the devfile
+	containers, err := ei.GetContainers()
 	if err != nil {
 		return err
 	}
 
+	// Add volumeMount to all containers in the devfile
```
dharmit, line 65: Doing this because the current behaviour of odo master branch is to add storage to all container components. This is up for discussion in #4105

pkg/envinfo/storage.go
```diff
@@ -149,7 +158,18 @@ func (ei *EnvInfo) DeleteStorage(name string) error {
...
+		return "", err
+	}
+
+	// since all container components have same volume mounts, we simply refer to the first container in the list
+	// refer https://github.com/openshift/odo/issues/4105 for addressing "all containers have same volume mounts"
+	paths, err := ei.devfileObj.Data.GetVolumeMountPaths(storageName, containers[0].Name)
+	if err != nil {
+		return "", err
+	}
+	return paths[0], nil
```
dharmit, line 172: @maysunfaisal @mik-dass please suggest if this should be changed.

pkg/envinfo/storage.go
dharmit: Ack. PTAL https://github.com/openshift/odo/pull/4465/commits/5b494c927dfecfd5fa0bea683436356576d1d503.

pkg/envinfo/storage.go
dharmit: PTAL https://github.com/openshift/odo/pull/4465/commits/5b494c927dfecfd5fa0bea683436356576d1d503

pkg/envinfo/storage.go
mik-dass: Looks OK to me for now. Please check the length of the containers and path before using them to avoid a panic.

dharmit: I was thinking about it just a few minutes back before I saw your comment. Ack, will change. 👍🏾

dharmit: Addressed. PTAL f8e2eb255.

dharmit:
Looks like an infra error:
```
 error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: while pulling "docker://registry.svc.ci.openshift.org/openshift/release:golang-1.13" as "registry.svc.ci.openshift.org/openshift/release:golang-1.13": Error initializing source docker://registry.svc.ci.openshift.org/openshift/release:golang-1.13: error pinging docker registry registry.svc.ci.openshift.org: invalid status code from registry 503 (Service Unavailable) 
```
/test v4.7-integration-e2e
/hold cancel

pkg/envinfo/storage.go
maysunfaisal: looks good!

pkg/envinfo/storage.go
maysunfaisal: looks good

pkg/envinfo/storage.go
dharmit: Is it a valid devfile if it doesn't have any container components?

mik-dass: > Is it a valid devfile if it doesn't have any container components?

Yes, that devfile is still valid.

dharmit: In that case, I don't think we should error with like I'm doing in https://github.com/openshift/odo/pull/4465/commits/ff5a4fb0f1e6c838316a2b220c5c579105a45a2c:
```go
"invalid devfile: components.container: required value"
```
What do you suggest?

mik-dass: `odo` needs a container component to work properly and thus the error is fine.

pkg/config/config_test.go
dharmit: PTAL https://github.com/openshift/odo/pull/4465/commits/2ad804242dfdda5411326bad8c66a170bac86a43.
Still missing at a few places, but I tried my best. Mocking would help iron things out, I hope.

dharmit:
All tests passed locally:
```
Ran 16 of 16 Specs in 429.686 seconds
SUCCESS! -- 16 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```
Going to force push the rebase. Let's see 🤞🏾

dharmit:
@mik-dass The `ci/prow/v4.7-integration-e2e` test has failed for the same test for two consecutive times.
* [Link 1](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4465/pull-ci-openshift-odo-master-v4.7-integration-e2e/1374996739852865536)
* [Link 2](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4465/pull-ci-openshift-odo-master-v4.7-integration-e2e/1376452603692453888)

In both the failures, the log is similar:
```sh
 odo devfile push command tests
/go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:20
  Verify devfile volume components work
  /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:719
    should throw a validation error for v1 devfiles [It]
    /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:779
    Expected
      
        <string>: I0325 09:53:49.340537   49860 context.go:163] absolute devfile path: '/tmp/210638443/devfile.yaml'
        I0325 09:53:49.340627   49860 context.go:90] absolute devfile path: '/tmp/210638443/devfile.yaml'
        I0325 09:53:49.340663   49860 util.go:727] HTTPGetRequest: https://raw.githubusercontent.com/openshift/odo/master/build/VERSION
        I0325 09:53:49.340791   49860 util.go:748] Response will be cached in /tmp/odohttpcache for 1h0m0s
        I0325 09:53:49.340972   49860 content.go:33] converted devfile YAML to JSON
        I0325 09:53:49.340990   49860 util.go:761] Cached response used.
         ✗  schemaVersion not present in devfile: /tmp/210638443/devfile.yaml
        
    to contain substring
        <string>: unable to find schema for version "1.0.0"
    /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_push_test.go:787 
```
That's why I have pushed 8e511a5.

pkg/odo/cli/service/backends.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+	}
+	return nil
+}
+
+func (b *ServiceCatalogBackend) ValidateServiceCreate(o *CreateOptions) (err error) {
+	// make sure the service type exists
+	classPtr, err := o.Client.GetKubeClient().GetClusterServiceClass(o.ServiceType)
+	if err != nil {
+		return errors.Wrap(err, "unable to create service because Service Catalog is not enabled in your cluster")
```
valaparthvi, line 344: Do we need to wrap the error here? Can't you just use errors.Error since you know why it failed to create the service?

pkg/devfile/adapters/common/generic_test.go
```diff
@@ -300,8 +301,11 @@ func TestExecuteDevfileCommand(t *testing.T) {
 }
 
 func adapter(fakeExecClient ExecClient, commands []devfilev1.Command, cif func(command devfilev1.Command) (ComponentInfo, error)) *GenericAdapter {
-	data := &testingutil.TestDevfileData{}
-	_ = data.AddCommands(commands...)
+	data := func() data.DevfileData {
+		devfileData, _ := data.NewDevfileData(string(data.APIVersion200))
+		return devfileData
+	}()
+	_ = data.AddCommands(commands)
```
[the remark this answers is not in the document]
dharmit: Ignoring it here because 1) the function in which it is called doesn't have an error return type 2) it's in unit test and not in actual code 3) once `devfile/library` repo provides mocking mechanism (https://github.com/devfile/api/issues/383), we won't have to make function calls to achieve what we're doing here.

pkg/odo/cli/service/backends.go
dharmit: Thanks. Fixed. I'm always confused when to use `Wrap` and when not to. 😕

pkg/kclient/deployments_test.go
```diff
@@ -26,11 +27,13 @@ func createFakeDeployment(fkclient *Client, fkclientset *FakeClientset, podName
 	fakeUID := types.UID("12345")
 
 	devObj := devfileParser.DevfileObj{
-		Data: &testingutil.TestDevfileData{
-			Components: []devfilev1.Component{
+		Data: func() data.DevfileData {
+			devfileData, _ := data.NewDevfileData(string(data.APIVersion200))
+			_ = devfileData.AddComponents([]devfilev1.Component{
```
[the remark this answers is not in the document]
dharmit: Fixed.

pkg/envinfo/storage_test.go
```diff
@@ -237,8 +252,12 @@ func TestEnvInfo_ListStorage(t *testing.T) {
...
 			fields: fields{
 				devfileObj: parser.DevfileObj{
-					Data: &testingutil.TestDevfileData{
-						Components: []devfilev1.Component{
+					Data: func() data.DevfileData {
+						devfileData, err := data.NewDevfileData(string(data.APIVersion200))
+						if err != nil {
+							t.Error(err)
+						}
+						_ = devfileData.AddComponents([]devfilev1.Component{
```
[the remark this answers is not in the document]
dharmit: Fixed.

pkg/envinfo/storage.go
```diff
@@ -133,7 +158,29 @@ func (ei *EnvInfo) DeleteStorage(name string) error {
...
+	}
+
+	// since all container components have same volume mounts, we simply refer to the first container in the list
+	// refer https://github.com/openshift/odo/issues/4105 for addressing "all containers have same volume mounts"
+	paths, err := ei.devfileObj.Data.GetVolumeMountPaths(storageName, containers[0].Name)
+	if err != nil {
+		return "", err
+	}
+
+	// Below "if" condition needs to go away when https://github.com/openshift/odo/issues/4105 is addressed.
```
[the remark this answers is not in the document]
dharmit: Added.

pkg/devfile/adapters/kubernetes/component/status_test.go
```diff
@@ -162,12 +164,30 @@ func TestGetDeploymentStatus(t *testing.T) {
...
+					if err != nil {
+						t.Error(err)
+					}
+					err = devfileData.AddCommands([]devfilev1.Command{getExecCommand("run", devfilev1.RunCommandGroupKind)})
+					if err != nil {
+						t.Error(err)
+					}
+					if err != nil {
+						t.Error(err)
+					}
```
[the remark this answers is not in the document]
dharmit: And at a bunch of places above that as well. Fixed.

pkg/odo/cli/service/backends.go
valaparthvi: I think if the cause of the error is unclear, it makes sense to use `errors.Wrap` because it wraps the error with a custom message, that gives out something like - `<custom error message>: <err>`.

pkg/devfile/adapters/kubernetes/component/status_test.go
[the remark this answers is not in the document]
dharmit: That happened because I used a few `sed` one-liners to add error check at a bunch of places to avoid having to do it all manually:
```sh
#!/bin/sh

sed -i 's/devfileData, _ \:\=/devfileData, err \:\=/g'  $1
sed -i '/devfileData, err \:\=/a if err!=nil \{\n t.Error(err) \n \}' $1
sed -i 's/_ \= devfileData\.Add/err = devfileData\.Add/g' $1
sed -i '/err \= devfileData.Add/a if err!=nil \{\n t.Error(err) \n \}' $1

go fmt $1
go test -i -race $1
echo $?
```

valaparthvi:
@dharmit I checked  https://github.com/openshift/odo/pull/4465/commits/d1d5f552c46fb9b4424d5311a914cc369d6aa463 and the changes you addressed look good to me. Although I do feel like there was a lot of repetition, for e.g. instantiating DevfileData and checking error like - 
```
devfileData, err := data.NewDevfileData(string(data.APIVersion200))
if err != nil {
	t.Error(err)
}
```

Perhaps we could put it in a helper function, but I am not sure if it makes much sense.

dharmit:
@valaparthvi I agree with your observation. However, the helper function would only be about that one call. And again it's going to require error check as well. I don't feel like it will be much helpful.

pkg/odo/cli/service/backends.go
```diff
@@ -0,0 +1,422 @@
...
+
+	err = svc.DeleteKubernetesComponentFromDevfile(instanceName, o.EnvSpecificInfo.GetDevfileObj())
+	if err != nil {
+		return errors.Wrap(err, "failed to delete service from the devfile")
+	}
+
+	return nil
+}
+
+func (b *ServiceCatalogBackend) CompleteServiceDelete(o *DeleteOptions, cmd *cobra.Command, args []string) (err error) {
```
mik-dass, line 266: Might be a good idea to separate the service catalog code into a separate file

dharmit: Makes sense. The file is becoming a mess.

pkg/odo/cli/service/backends.go
```diff
@@ -0,0 +1,422 @@
...
+	// if two args are given, first is service type and second one is service name
+	if len(args) == 2 {
+		o.ServiceName = args[1]
+	}
+
+	return nil
+}
+
+func (b *OperatorBackend) ValidateServiceCreate(o *CreateOptions) (err error) {
+	d := NewDynamicCRD()
```
mik-dass, line 59: Since the `DynamicCRD` struct and it's functions are specific to operator hub, please move them out of the create.go

pkg/odo/cli/service/backends.go
```diff
@@ -0,0 +1,422 @@
...
+			return fmt.Errorf("no plan was supplied for service %v.\nPlease select one of: %v\n", o.ServiceType, strings.Join(ui.GetServicePlanNames(plans), ","))
+		}
+	} else {
+		// when the plan has been supplied, we need to make sure it exists
+		if _, ok := plans[o.Plan]; !ok {
+			return fmt.Errorf("plan %s is invalid for service %v.\nPlease select one of: %v\n", o.Plan, o.ServiceType, strings.Join(ui.GetServicePlanNames(plans), ","))
+		}
+	}
+	//validate service name
+	return o.validateServiceName(o.ServiceName)
```
mik-dass, line 372: This functions is specific to service catalog and can be moved out of the create.go

pkg/odo/cli/service/interface.go
```diff
@@ -0,0 +1,15 @@
+package service
+
+import "github.com/spf13/cobra"
+
+// ServiceProviderBackend is implemented by the backends supported by odo
+// It is used in "odo service create"
```
mik-dass, line 6: It is also used by `odo service delete`

pkg/service/service.go
```diff
@@ -53,16 +59,16 @@ func (params servicePlanParameters) Swap(i, j int) {
 }
 
 // CreateService creates new service from serviceCatalog
-func CreateService(client *occlient.Client, serviceName string, serviceType string, servicePlan string, parameters map[string]string, applicationName string) error {
+// It returns string representation of service instance created on the cluster and error (if any).
+func CreateService(client *occlient.Client, esi *envinfo.EnvSpecificInfo, serviceName, serviceType, servicePlan string, parameters map[string]string, applicationName string) (string, error) {
```
mik-dass, line 63: The envinfo is not used in the function

pkg/service/service.go
```diff
@@ -97,7 +103,7 @@ func doesCRExist(kind string, csvs *olm.ClusterServiceVersionList) (olm.ClusterS
 }
 
 // CreateOperatorService creates new service (actually a Deployment) from OperatorHub
-func CreateOperatorService(client *kclient.Client, group, version, resource string, CustomResourceDefinition map[string]interface{}) error {
+func CreateOperatorService(client *kclient.Client, esi *envinfo.EnvSpecificInfo, serviceName, group, version, resource string, CustomResourceDefinition map[string]interface{}) error {
```
mik-dass, line 106: The envInfo is not used in the function

pkg/service/service_test.go
```diff
@@ -763,3 +772,125 @@ func TestServicePlanParameterMarshalling(t *testing.T) {
...
+			},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := AddKubernetesComponentToDevfile(tt.args.crd, tt.args.name, tt.args.devfileObj); (err != nil) != tt.wantErr {
+				t.Errorf("AddKubernetesComponentToDevfile() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			got, _ := tt.args.devfileObj.Data.GetComponents(common.DevfileOptions{})
```
mik-dass, line 828: Please handle the error using `t.Errorf()`

pkg/service/service_test.go
```diff
@@ -763,3 +772,125 @@ func TestServicePlanParameterMarshalling(t *testing.T) {
...
+			wantErr: false,
+			want:    []v1alpha2.Component{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := DeleteKubernetesComponentFromDevfile(tt.args.name, tt.args.devfileObj); (err != nil) != tt.wantErr {
+				t.Errorf("DeleteKubernetesComponentFromDevfile() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			got, _ := tt.args.devfileObj.Data.GetComponents(common.DevfileOptions{})
```
mik-dass, line 890: Please handle the error using `t.Errorf()`

pkg/odo/cli/service/create.go
```diff
@@ -117,127 +105,52 @@ func NewDynamicCRD() *DynamicCRD {
...
-			// otherwise select the plan interactively
-			o.Plan = ui.SelectPlanNameInteractively(plans, "Which service plan should we use ")
-			svcPlan = plans[o.Plan]
-		}
-
-		o.ParametersMap = ui.EnterServicePropertiesInteractively(svcPlan)
-		o.ServiceName = ui.EnterServiceNameInteractively(o.ServiceType, "How should we name your service ", o.validateServiceName)
-		o.outputCLI = commonui.Proceed("Output the non-interactive version of the selected options")
-		o.wait = commonui.Proceed("Wait for the service to be ready")
+	_, _, err = svc.SplitServiceKindName(args[0])
```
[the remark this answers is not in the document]
dharmit: > This causes a panic when argument is passed

You mean when the argument is "not" passed?

pkg/odo/cli/service/interface.go
```diff
@@ -0,0 +1,15 @@
...
+type ServiceProviderBackend interface {
+	CompleteServiceCreate(options *CreateOptions, cmd *cobra.Command, args []string) error
+	ValidateServiceCreate(options *CreateOptions) error
+	RunServiceCreate(options *CreateOptions) error
+
+	CompleteServiceDelete(options *DeleteOptions, cmd *cobra.Command, args []string) error
```
mik-dass, line 12: Both the implementation don't use any of the arguments passed and just return nil

pkg/odo/cli/service/backends.go
dharmit: Separated into different files.

pkg/odo/cli/service/backends.go
dharmit: Moved to `operator_backend.go`.

pkg/odo/cli/service/interface.go
dharmit: Updated.

pkg/service/service.go
dharmit: Removed

pkg/service/service_test.go
dharmit: Done.

pkg/service/service_test.go
dharmit: Done.

pkg/odo/cli/service/create.go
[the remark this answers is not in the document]
dharmit: Fixed. Modified the test as well.

pkg/odo/cli/service/interface.go
dharmit: Removed from interface and implementation.

pkg/odo/cli/service/backends.go
dharmit: Moved.

pkg/config/config_test.go
dharmit: Marking resolved since I've addressed the comments, IMO.

pkg/odo/cli/service/create.go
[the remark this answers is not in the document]
mik-dass: > You mean when the argument is "not" passed?

Yes. Sorry, I meant when no argument is not passed.

tests/integration/operatorhub/cmd_service_test.go
```diff
@@ -26,13 +25,10 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
 	})
 
 	AfterEach(func() {
-		helper.CommonAfterEach(commonVar)
+		//helper.CommonAfterEach(commonVar)
```
mik-dass, line 28: commented code

pkg/odo/cli/service/delete.go
```diff
@@ -26,89 +29,86 @@ var (
...
+	if err != nil {
+		// failure to split provided name into two; hence ServiceCatalogBackend
+		o.Backend = NewServiceCatalogBackend()
+
+		err = nil
 	} else {
-		o.Context, err = genericclioptions.NewContext(cmd)
+		// provided name adheres to the format <operator-type>/<crd-name>; hence OperatorBackend
+		o.Backend = NewOperatorBackend()
 	}
```
mik-dass, line 60: The logic is the same as create.go. Maybe put them in a single place like types.go in this package.

mik-dass:
Please add a devfile related check in the integration test

dharmit:
> Please add a devfile related check in the integration test

I have not added integration test for the functionality added in this PR because there's unit test coverage. Can you PTAL and suggest if something's really missing there? 

My reasoning for not adding an integration test is that:
1. I think the unit test is good enough since it checks the devfile and also uses a mock fs
2. Integration tests take longer and odo needs to reduce them as much as possible.

pkg/odo/cli/service/delete.go
dharmit: But the struct types are different in both cases. One is `DeleteOptions` and the other is `CreateOptions`. How do you suggest I create a "common" function that would set the backend type for both types?

mik-dass:
> I have not added integration test for the functionality added in this PR because there's unit test coverage. Can you PTAL and suggest if something's really missing there?

But we need integration test to ensure that the devfile is modified accordingly when the command is triggered and that the various units of the code work properly together. We can skip this for now, but when we implement the push functionality, please add a test that ensures `odo push` does create a service.

> Integration tests take longer and odo needs to reduce them as much as possible.

We can add these tests in between other tests.

pkg/odo/cli/service/delete.go
mik-dass: We can use the `args[0]` value and decide the backend type. We can also pass the various options like interactive, fromFile etc for deciding the backend type.

dharmit:
> But we need integration test to ensure that the devfile is modified accordingly when the command is triggered 

And hence I have added unit test for the function that's going to be called when the command is triggered. This is why I mentioned that it's already validating devfile and also using mock fs.

> and that the various units of the code work properly together.

IMO, integration test shouldn't be our only way to ensure this. But I'm more than willing to hear differing opinion. Please chime in @kadel @rnapoles-rh 

> We can skip this for now, but when we implement the push functionality, please add a test that ensures `odo push` does create a service.

Ack, this makes sense to me right now since that would be testing a complete functionality which ensures that requested resources were created on the cluster. Just to be clear, I don't intend to "skip" adding integration test. I am looking for a valid reason to add it.

kadel:
> I have not added integration test for the functionality added in this PR because there's unit test coverage. Can you PTAL and suggest if something's really missing there?
> 
> My reasoning for not adding an integration test is that:
> 
> 1. I think the unit test is good enough since it checks the devfile and also uses a mock fs
> 2. Integration tests take longer and odo needs to reduce them as much as possible.

there should be at least one simple integration test that validates that `odo service create` and `odo service delete` are working correctly.

pkg/odo/cli/service/delete.go
dharmit: > We can use the args[0] value and decide the backend type

We rely on this information at only one place and there's barely a few lines of code for it:
```go
_, _, err = svc.SplitServiceKindName(args[0])
if err != nil {
	// failure to split provided name into two; hence ServiceCatalogBackend
	o.Backend = NewServiceCatalogBackend()

	err = nil
}
```

> We can also pass the various options like interactive, fromFile etc for deciding the backend type.

These options are specific to service creation only. For deletion, these options are of no use.

prietyc123:
> there should be at least one simple integration test that validates that odo service create and odo service delete are working correctly.

+1, Not sure if I am missing something here. But these are user facing scenario and if we are enhancing some feature, we can just modify our existing integration test cases accordingly. Just for validating our current behaviour for such scenarios.

kadel:
Executing `odo service create` in a directory without devfile.yaml panics.

```
▶ odo service create --from-file database.yaml
Deploying service "mydb" of type: "Database"
 ✓  Deploying service [11ms]
 ✓  Service "mydb" was created
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a75228]

goroutine 1 [running]:
github.com/openshift/odo/pkg/odo/cli/service.(*OperatorBackend).RunServiceCreate(0xc00004ed20, 0xc00051e2c0, 0x0, 0x0)
        /home/tomas/Code/Work/odo/pkg/odo/cli/service/operator_backend.go:283 +0x4a8
github.com/openshift/odo/pkg/odo/cli/service.(*CreateOptions).Run(0xc00051e2c0, 0x0, 0x0)
        /home/tomas/Code/Work/odo/pkg/odo/cli/service/create.go:173 +0x43
github.com/openshift/odo/pkg/odo/genericclioptions.GenericRun(0x21b9d08, 0xc00051e2c0, 0xc000684a00, 0xc0007185c0, 0x0, 0x2)
        /home/tomas/Code/Work/odo/pkg/odo/genericclioptions/runnable.go:78 +0x202
github.com/openshift/odo/pkg/odo/cli/service.NewCmdServiceCreate.func1(0xc000684a00, 0xc0007185c0, 0x0, 0x2)
        /home/tomas/Code/Work/odo/pkg/odo/cli/service/create.go:201 +0x5e
github.com/spf13/cobra.(*Command).execute(0xc000684a00, 0xc0007185a0, 0x2, 0x2, 0xc000684a00, 0xc0007185a0)
        /home/tomas/Code/Work/odo/vendor/github.com/spf13/cobra/command.go:830 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc00057ac80, 0x1ff2bd0, 0xc000117860, 0x4)
        /home/tomas/Code/Work/odo/vendor/github.com/spf13/cobra/command.go:914 +0x30b
github.com/spf13/cobra.(*Command).Execute(...)
        /home/tomas/Code/Work/odo/vendor/github.com/spf13/cobra/command.go:864
main.main()
        /home/tomas/Code/Work/odo/cmd/odo/odo.go:68 +0x373
```

valaparthvi:
@dharmit Can you update your PR branch with the master? This branch is missing out on the new changes and hence sending false telemetry data.

dharmit:
> @dharmit Can you update your PR branch with the master? This branch is missing out on the new changes and hence sending false telemetry data.

Done.

dharmit:
> Executing `odo service create` in a directory without devfile.yaml panics.

PTAL https://github.com/openshift/odo/pull/4465/commits/794b5b904e6290700395fdad7778d675e47a0d76.

pkg/odo/cli/service/delete.go
mik-dass: > We rely on this information at only one place and there's barely a few lines of code for it:

The same logic is present in create.go. If the deciding logic changes due to some library update then we would need to fix it at both places.

> These options are specific to service creation only. For deletion, these options are of no use.

But it enables us to put the deciding logic in one place.

dharmit: > If the deciding logic changes due to some library update then we would need to fix it at both places.

Makes sense. I'll put this part in one function.

> But it enables us to put the deciding logic in one place. Anyways, we can leave this for now.

This doesn't make sense even for later because:
1. the options are specific only create and not delete
2. they are specific to operators and have nothing to do with service catalog backend.

dharmit: https://github.com/openshift/odo/pull/4465/commits/648eb24cae329b6f9c7ff58cc1192c7112bec25d

mik-dass: > the options are specific only create and not delete
they are specific to operators and have nothing to do with service catalog backend.

It puts the decision making in one place and can clean the create.go and delete.go code. Anyways we can skip the code for now.

kadel:
Did few tests and from a functionality point of view everything looks good. I haven't found any big issues

pkg/service/service.go
```diff
@@ -679,10 +683,58 @@ func isRequired(required []string, name string) bool {
...
+					},
+				},
+			},
+		},
+	}})
+	if err != nil {
+		return err
+	}
+
+	return devfileObj.WriteYamlDevfile()
```
kadel, line 713: Shouldn't this call be outside of this function?
This pattern seems a little bit weird to me. Based on the description and function name I would not expect to write the information directly to the file. What If I want to do more modification to the devfile structure and then write it all at once?

It might be better if the caller of this function can decide when to write devfile to file.

dharmit: > Shouldn't this call be outside of this function?

I don't have a strong opinion.

> This pattern seems a little bit weird to me.

I'm following the same approach as in:

* https://github.com/openshift/odo/blob/ba7161940c735caffaa9bd49e731dd15b8e9a61f/pkg/envinfo/url.go#L319-L333
* https://github.com/openshift/odo/blob/ba7161940c735caffaa9bd49e731dd15b8e9a61f/pkg/envinfo/url.go#L335-L360
* https://github.com/openshift/odo/blob/ba7161940c735caffaa9bd49e731dd15b8e9a61f/pkg/envinfo/storage.go#L120-L132
* https://github.com/openshift/odo/blob/ba7161940c735caffaa9bd49e731dd15b8e9a61f/pkg/envinfo/storage.go#L56-L78

> Based on the description and function name I would not expect to write the information directly to the file. 

Well the function comment does say that we put the info in devfile. I'm not sure how it's not saying what it's doing.

> What If I want to do more modification to the devfile structure and then write it all at once?

Would that modification be about service create and in the same odo command?

> It might be better if the caller of this function can decide when to write devfile to file.

Again, no strong opinion myself. Even if this is the right thing to do, I'm actually tired of accommodating so many changes that I have been making on this PR. I request to open a new issue for this.

dharmit:
> Did few tests and from a functionality point of view everything looks good. I haven't found any big issues

Thanks for trying it out!

pkg/service/service.go
kadel: It is ok, it looks like we have made this mistake already in a lot of places. This will need to be cleaned up. But obviously not in this PR

dharmit:
```
INFO[2021-04-08T16:45:23Z] Received signal.                              signal=interrupt
INFO[2021-04-08T16:45:23Z] error: Process interrupted with signal interrupt, cancelling execution... 
INFO[2021-04-08T16:45:23Z] cleanup: Deleting release pod release-latest 
INFO[2021-04-08T16:47:21Z] Step integration-e2e-ipi-deprovision-deprovision succeeded after 6m20.074798184s. 
INFO[2021-04-08T16:47:21Z] Releasing leases for test integration-e2e    
INFO[2021-04-08T16:47:21Z] Ran for 4h1m57s                              
ERRO[2021-04-08T16:47:21Z] Some steps failed:                           
ERRO[2021-04-08T16:47:21Z]   * could not run steps: execution cancelled 
```

Wonder if this means CI failed after executing the integration tests for 4+ hours.

/test v4.7-integration-e2e

dharmit:
```
error: error creating buildah builder: Error writing blob: error storing blob to file "/var/tmp/storage358802064/6": error happened during read: read tcp 10.130.41.37:41084->108.177.11.128:443: read: connection reset by peer 
```

pkg/service/service.go
dharmit: https://github.com/openshift/odo/issues/4607

mik-dass (approved):
/lgtm

Issue #3786, opened 2020-08-19
Add errors.Wrapf to portions of the code that using json.Unmarshal


cdrage:
/kind feature

## Which functionality do you think we should add?

I am running into issues like below:
```sh
~/openshift/test                                                                                                                                                                     
▶ odo create --devfile devfile.yaml --starter
Validation
 ✓  Creating a devfile component from devfile path: /home/wikus/openshift/test/devfile.yaml [34546ns]
 ✓  Validating devfile component [30625ns]
 ✗  yaml: line 7: could not find expected ':'
```

Where the error output is ambiguous. As a user, the error output is not very good. As a developer, its hard to find what part of the code is actually causing the issue.

Throughout our code we are using `json.Unmarshal` but with no error wrapping to indicate what part of the code is causing the error.

For example:

```go
  // OverrideComponents overrides the components of the parent devfile    
  // overridePatch contains the patches to be applied to the parent's components    
  func (d DevfileObj) OverrideComponents(overridePatch []common.DevfileComponent) error {    
    for _, patchComponent := range overridePatch {    
      found := false    
      for _, originalComponent := range d.Data.GetComponents() {    
        if strings.ToLower(patchComponent.Container.Name) == originalComponent.Container.Name {    
          found = true    
       
          var updatedComponent common.Container    
      
          merged, err := handleMerge(originalComponent.Container, patchComponent.Container, common.Container{})    
          if err != nil {    
            return err    
          }    
      
          err = json.Unmarshal(merged, &updatedComponent)    
          if err != nil {    
            return err            
          }    
      
          d.Data.UpdateComponent(common.DevfileComponent{Container: &updatedComponent})    
        }    
      }    
      if !found {    
        return fmt.Errorf("the component to override is not found in the parent")    
      }
    }  
    return nil
  }
```

has

```go
          err = json.Unmarshal(merged, &updatedComponent)    
          if err != nil {    
            return err            
          }    
```

Which should be wrapped.

## Why is this needed?

Doing error wrapping gives:
-  A clear indicator of where the issue is
- What we can do to fix it
- Where the unmarshaling went wrong.

PR #4482, opened 2021-03-04
Enter Telemetry 

### LINKED ISSUE #4462: Track telemetry implementation for odo
/kind feature
/priority high

## Acceptance Criteria
For first pass - 
     - - [x] 3.1. OS information
     - - [x] 3.3. Odo version

For later - 
Collect information such as -
- User preferences
- Proxy enabled

## Why is this needed?

valaparthvi:
**What type of PR is this?**
/kind feature

This PR implements segment integration into odo. With the changes introduced in this PR, we will be able to see telemetry data in segment.

Fixes parts of #4462 

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [x] Be able to see data(errors, success/failure) in segment

Run any command now and see if you can see data in segment.
**Note:** If you do not have access to the segment, you can do the following - 
1) Create a free account on https://segment.com/ with your work email ID
2) Create a Go source with whatever name you like. If prompted to create a destination, skip that for now.
3) Navigate to your sources' details page and copy the WriteKey.
4) Paste the writekey in `segment.go::WriteKey`
5) Give consent to telemetry so that data can be sent to segment with `odo set preference ConsentTelemetry true`
6) Run any command.
7) Go to the debugger section of your source on Segment. There must be some events in the list.

Note: Having the WriteKey here in the code will not cause a security concern.


valaparthvi:
@dharmit @girishramnani Can one of you tell me why this PR deletes more than 47k lines of code? I had to install analytics-go package so that I could interact with segment API for which I followed this [doc](https://github.com/openshift/odo/blob/master/docs/dev/development.adoc#dependency-management), so I assume `go mod tidy` removed the unused packages which is 47k lines of code.

girishramnani:
if odo is building correctly locally then you shouldn't worry @valaparthvi and yeah its a good thing that `go mod tidy` removed irrelavent packages

pkg/odo/genericclioptions/runnable.go
```diff
@@ -3,21 +3,39 @@ package genericclioptions
...
 
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/odo/util"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
+var (
+	segmentClient *segment.Client
+	//ctx context.Context
```
girishramnani, line 20: commented code?

pkg/odo/genericclioptions/runnable.go
```diff
@@ -3,21 +3,39 @@ package genericclioptions
...
+)
+
 type Runnable interface {
 	Complete(name string, cmd *cobra.Command, args []string) error
 	Validate() error
 	Run() error
 }
 
 func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
+	//ctx = telemetry.NewContext(context.Background())
```
girishramnani, line 30: same?

girishramnani: commented code?

pkg/odo/genericclioptions/runnable.go
```diff
@@ -3,21 +3,39 @@ package genericclioptions
...
+	startTime := time.Now()
+	cfg, _ := preference.New()
+	var err error
+	// Initiate the segment client
+	if segmentClient, err = segment.NewClient(cfg); err != nil {
+		klog.Fatal(err.Error())
```
girishramnani, line 36: As telemetry is an auxiliary system - we never fail on it. Even if there is an error - propogate it to upper layers or just log it as info. We never fail if segment is not available.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -26,9 +44,28 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
...
+}
+
+// uploadToSegmentAndLog uploads the data to segment
+func uploadToSegmentAndLog(cmd *cobra.Command, err error, startTime time.Time) {
+	if serr := segmentClient.Upload(cmd.CommandPath(), time.Since(startTime), err); serr != nil {
+		klog.Errorf("Cannot send data to telemetry: %v", serr)
+	}
+	// If the error is not nil, client will be closed so that data can be sent before the program exits.
+	if err != nil {
+		segmentClient.Close()
```
girishramnani, line 66: I saw a differ that closes the client already above, so do we need to close it here?

pkg/odo/genericclioptions/runnable.go
valaparthvi: Yes. I kept it for the future. I am not sure if we want to use context as they have done in CRC. In case we do, this would act as a starting point. I will remove it later in case we decide context is not required.

CRC is using context to pass some values to segment. For e.g. w.r.t to odo, if we decide we want to pass values such as `odo preference set buildtimeout 5`, values being `buildtimeout` and `5`, we might need to use context.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -3,21 +3,39 @@ package genericclioptions
...
 type Runnable interface {
 	Complete(name string, cmd *cobra.Command, args []string) error
 	Validate() error
 	Run() error
 }
 
 func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
+	//ctx = telemetry.NewContext(context.Background())
+	startTime := time.Now()
+	cfg, _ := preference.New()
```
girishramnani, line 32: I would like you to consider the testablity of you code as well, assume you had to test this function `GenericRun` while you mock this preference object, could you be able to do that?
I would suggest - only pass the preference value you care about and not the whole object.

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+type Client struct {
+	segmentClient     analytics.Client
+	config            *preference.PreferenceInfo
+	telemetryFilePath string
+}
+
+func NewClient(config *preference.PreferenceInfo) (*Client, error) {
+	homeDir, _ := os.UserHomeDir()
+	return newCustomClient(config,
+		filepath.Join(homeDir, ".redhat", "anonymousId"),
```
girishramnani, line 32: we already have `.odo` dir, canot we use that instead of `.redhat`? or is this something all other devtools projects do?

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+		analytics.DefaultEndpoint,
+	)
+}
+
+func newCustomClient(config *preference.PreferenceInfo, telemetryFilePath string, segmentEndpoint string) (*Client, error) {
+	client, err := analytics.NewWithConfig(WriteKey, analytics.Config{
+		Endpoint: segmentEndpoint,
+		DefaultContext: &analytics.Context{
+			IP: net.IPv4(0, 0, 0, 0),
```
girishramnani, line 41: why aren't we getting the actually IP address here? or we actually shouldn't?

pkg/odo/genericclioptions/runnable.go
valaparthvi: Yes, we do. The `segmentClient.Upload` function queues the data to be sent to the segment and sends it once we call the `segmentClient.Close` and close the connection.

`uploadToSegmentAndLog` uploads the data to segment and also calls `LogErrorAndExit` which would exit the program in case err!=nil, therefore not going back to the deferred close call and not sending the data to segment. So in case err!=nil, we want to close the segment connection and send the data before exiting the program.

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+
+// Close client connection and send the data
+func (c *Client) Close() error {
+	return c.segmentClient.Close()
+}
+
+// Upload prepares the data to be sent to segment and send it once the client connection closes
+func (c *Client) Upload(action string, duration time.Duration, err error) error {
+	// if the user has not consented for telemetry, return
+	if !c.config.GetConsentTelemetry() {
```
girishramnani, line 62: I would suggest we refactor this consent piece in a way that we dont even initialise the segment client if the consent is not set, and lets do that in the genericclioptions? would save us some initialisation time?

dharmit:
I built odo from this PR. Executed a few random odo commands, but odo doesn't prompt me to enable telemetry. Shouldn't odo prompt the user to enable it for the first time? If user opts-in, set the value to `true` in `~/.odo/preference.yaml`; if they opt-out, set the value to `false`. So basically if the value is not set in `preference.yaml`, we should prompt the user to opt-in/-out, no?

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := c.segmentClient.Enqueue(analytics.Identify{
+		AnonymousId: anonymousID,
+		Traits:      addConfigTraits(c.config, traits()),
+	}); err != nil {
+		return err
+	}
+
+	// add information to the data
+	properties := analytics.NewProperties()
+	// TODO: add other properties when required
```
girishramnani, line 82: TODO?

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+	properties = properties.Set("version", "odo "+version.VERSION+" ("+version.GITCOMMIT+")").
```
girishramnani, line 83: please use Sprintf or templates to make this readable

pkg/segment/segment.go
valaparthvi: You are correct, it is something that all the other devtools projects do. There are future plans to segregate telemetry data from multiple tools and it would help if the anonymousID is the same. And I think it also affects the cost of Segment. Segment charges based on the number of users tracked/month, so having this mechanism is cost effective.

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+		AnonymousId: anonymousID,
+		Event:       action,
+		Properties:  properties,
+	})
+}
+
+// addConfigTraits add more information to be sent to segment
+// Note: This currently acts as a placeholder
+func addConfigTraits(c *preference.PreferenceInfo, in analytics.Traits) analytics.Traits {
+	// TODO: add more traits later
```
girishramnani, line 102: TODO?

pkg/segment/segment.go
valaparthvi: Right, we shouldn't. We cannot collect any Personally Identifiable Information(PII).

pkg/segment/segment.go
valaparthvi: Right, that sounds like a good idea. But I think I would like to keep this check here, in case someone explicitly calls the Upload function, I understand the chances of that are slim, but I would still keep it.

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+	return in
+}
+
+// getUserIdentity returns the anonymous ID if it exists, else creates a new one
+func getUserIdentity(telemetryFilePath string) (string, error) {
+	var id []byte
+
+	if err := os.MkdirAll(filepath.Dir(telemetryFilePath), 0750); err != nil {
```
girishramnani, line 110: please check how other functions that use MkdirAll etc.. use the afero for testing 🙂 .

That would make writing unit test for this code easy

pkg/segment/segment.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return strings.TrimSpace(string(id)), nil
+}
+
+// SetError sanitizes any pii information from the error
+func SetError(err error) string {
+	// Sanitize user information
+	user1, err1 := user.Current()
+	if err1 != nil {
+		return err1.Error()
```
girishramnani, line 136: probably errors.Wrap here as we dont wanna miss the original `err`?

valaparthvi:
>I built odo from this PR. Executed a few random odo commands, but odo doesn't prompt me to enable telemetry. Shouldn't odo prompt the user to enable it for the first time? If user opts-in, set the value to true in ~/.odo/preference.yaml; if they opt-out, set the value to false. So basically if the value is not set in preference.yaml, we should prompt the user to opt-in/-out, no?

@dharmit 
Yes, it should and it will, once I implement it. The scope of issue #4462  for this sprint was just to get data to segment and report success/failures, but I will work on this opt-in/out thing next.

pkg/segment/segment_test.go
```diff
@@ -0,0 +1,191 @@
...
+			testName: "command failed",
+			err:      errors.New("some error occurred"),
+			success:  false,
+			errType:  "*errors.errorString",
+		},
+	}
+	for _, tt := range tests {
+		t.Log("Running test: ", tt.testName)
+		t.Run(tt.testName, func(t *testing.T) {
+			c, err := newCustomClient(cfg, filepath.Join(os.TempDir(), "telemetry"), server.URL)
```
girishramnani, line 125: A unit test shouldn't interact with external system that are not now controlled by you e.g. filesystems, API etc..
Please mock the filesystem using afero - you will find many such examples

pkg/segment/segment.go
girishramnani: yeah I dont mind if this code stays here - but we should only init the client if consent is true in the beginning itself

pkg/odo/genericclioptions/runnable.go
```diff
@@ -26,9 +44,28 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
...
+	err = o.Validate()
+	if err != nil {
+		uploadToSegmentAndLog(cmd, err, startTime)
+	}
+	uploadToSegmentAndLog(cmd, o.Run(), startTime)
+}
+
+// uploadToSegmentAndLog uploads the data to segment
+func uploadToSegmentAndLog(cmd *cobra.Command, err error, startTime time.Time) {
+	if serr := segmentClient.Upload(cmd.CommandPath(), time.Since(startTime), err); serr != nil {
```
girishramnani, line 61: is this async in nature?

pkg/odo/genericclioptions/runnable.go
girishramnani: We would need context to cancel a segment upload when someone `Sigkills` odo? just thinking out loud, if segment uploads are not async in nature then we would need to have a functionality to breakout of a stuck segment upload?

pkg/segment/segment.go
girishramnani: then why even set it?

pkg/segment/segment_test.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	}
+
+	c, err := newCustomClient(cfg, filepath.Join(os.TempDir(), "telemetry"), server.URL)
+	//	assert err != nil
+	if err != nil {
+		t.Fatal(err)
+	}
+	// run a command, odo preference view
+	if err := c.Upload("odo preference view", time.Second, errors.New("an error occured")); err != nil {
```
girishramnani, line 75: you need to mock the segment api responses as expected from segment and then validate the response instead of just checking for error?

dharmit:
```sh
$ odo preference set ConsentTelemetry 123
 ✗  strconv.ParseBool: parsing "123": invalid syntax
```

Is this PR not rebased on top of https://github.com/openshift/odo/pull/4498, or did we miss this check here? Can we have a common function for this kind of issues?

pkg/segment/segment_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+			default:
+				t.Error("Server should receive data")
+			}
+		})
+	}
+}
+
+func TestSetError(t *testing.T) {
+	err := errors.New("this is an error string")
```
girishramnani, line 176: not sure whats the purpose of this?

dharmit:
* odo built using `master` branch takes a total 0.053 seconds to show the result:
    ```sh
    $ time odo preference view
    UpdateNotification    
    NamePrefix            dharmit
    Timeout               
    BuildTimeout          3000
    PushTimeout           480
    Experimental          
    PushTarget            
    Ephemeral             true
    ConsentTelemetry      true
    odo preference view  0.07s user 0.01s system 142% cpu 0.053 total
    ```
* odo built from this PR takes 1.243 seconds to show the output and the slowness is very much real (in the sense, I already started typing another odo command and was wondering what's happening on my CLI):
    ```
    $ time odo preference view
    UpdateNotification    
    NamePrefix            dharmit
    Timeout               
    BuildTimeout          3000
    PushTimeout           480
    Experimental          
    PushTarget            
    Ephemeral             true
    ConsentTelemetry      true
    odo preference view  0.08s user 0.01s system 6% cpu 1.243 total
    ```
* And when the system is disconnected (considering disconnected use cases), the command fails since it can't connect to segment.io:
    ```sh
    $ time odo preference view
    UpdateNotification    
    NamePrefix            dharmit
    Timeout               
    BuildTimeout          3000
    PushTimeout           480
    Experimental          
    PushTarget            
    Ephemeral             true
    ConsentTelemetry      true
    segment 2021/03/18 13:07:19 ERROR: sending request - Post https://api.segment.io/v1/batch: dial tcp: lookup api.segment.io: no such host
    segment 2021/03/18 13:07:19 ERROR: 2 messages dropped because they failed to be sent and the client was closed
    odo preference view  0.06s user 0.02s system 135% cpu 0.058 total
    ```

I'm wondering if we really need to connect to segment.io for `odo preference view`.

pkg/segment/segment_test.go
valaparthvi: We are doing that in the other test `TestClientUploadWithConsent`. This test just makes sure that no data is received by the server if the user had not consented for telemetry.

pkg/segment/segment_test.go
valaparthvi: To check if the username, incase it appears in an error, is sanitized before being sent to Segment. I'll be adding more tests related to sanitization once we add more sanitization checks.

valaparthvi: The test first checks a proper error string that does not contain any PII and makes sure the error returned by `SetError` is the same as the error itself. Next, it passes an error with a username(PII) and makes sure the error returned by `SetError` is not the same as the error itself and that the username was sanitized.

pkg/odo/genericclioptions/runnable.go
valaparthvi: As far as I can tell, yes. The way it works is, calling the `Upload` method adds data to the queue and BTS sends it asynchronously in a batch once the connection closes.

girishramnani: Can we confirm that?

pkg/odo/genericclioptions/runnable.go
valaparthvi: I checked the source code for segment's go client and it is asynchronous as far as I can tell, but I will do some more research and make sure it is.

pkg/odo/genericclioptions/runnable.go
valaparthvi: >I would like you to consider the testablity of you code as well, assume you had to test this function GenericRun while you mock this preference object, could you be able to do that?

I am still not sure I understand what you mean here.

>I would suggest - only pass the preference value you care about and not the whole object.

I care about `ConsentTelemetry` in particular, but in case we want to pass other preference values to Segment like CRC has done [here](https://github.com/code-ready/crc/blob/master/pkg/crc/segment/segment.go#L104), it would be easier if we had the preference object.

pkg/segment/segment.go
valaparthvi: From what I read [here](https://segment.com/docs/connections/spec/common/#context), if we don't set the context with IP as we have done right now, it will take the current user's IP and we don't want that.
I will confirm this with PK and let you know.

pkg/segment/segment.go
valaparthvi: That TODO will come into action if we decide to add more properties like I mentioned here - https://github.com/openshift/odo/pull/4482#discussion_r596599713

pkg/segment/segment.go
valaparthvi: Add more information related to the user's system and preferences w.r.t odo.
Like here - https://github.com/code-ready/crc/blob/master/pkg/crc/segment/segment.go#L101

valaparthvi:
>Is this PR not rebased on top of #4498, or did we miss this check here? Can we have a common function for this kind of issues?

@dharmit 
It is, but I forgot to make the change for ConsentTelemetry that I did for other preferences. I'll add it to this PR.

The solution for this kind of issue is to replace Wrapf with Errorf, but also make sure the actual error does not go unnoticed.
But making this change kinda goes against the use of Wrapf as is described in issue #3786. We need to evaluate where we should use Wrapf and Errorf.

pkg/segment/segment.go
valaparthvi: I confirmed this with PK and he agrees with the reason I have given here.

pkg/odo/genericclioptions/runnable.go
kadel: that is a good point. failing to send or create telemetry should not effect odo functionality.
The error should be just logged `klog.V(4).Infof`

pkg/odo/genericclioptions/runnable.go
kadel: It is async.

pkg/odo/genericclioptions/runnable.go
kadel: > Can we confirm that?

It works as @valaparthvi describes

valaparthvi:
>And when the system is disconnected (considering disconnected use cases), the command fails since it can't connect to segment.io: 

@dharmit I checked this and it does not really block anything for mainly 2 reasons - 1) It is executed at the very end, 2) If it is unable to send the data, it prints the message and continues.

The klog suggestion will not work here, because it is analytics that sends the data via a channel and it all happens [internally](https://github.com/segmentio/analytics-go/blob/v3.2.0/analytics.go#L273). I could look into this later but I don't think it is a priority right now.

>odo built from this PR takes 1.243 seconds to show the output and the slowness is very much real (in the sense, I already started typing another odo command and was wondering what's happening on my CLI):

It takes the time it takes to send the data, but as a temporary solution, we could print a statement saying it is sending the data to Segment, please wait, or something like that, although I think it would be a bit annoying to see that message every time.

pkg/odo/genericclioptions/runnable.go
valaparthvi: > I would like you to consider the testablity of you code as well, assume you had to test this function GenericRun while you mock this preference object, could you be able to do that?

I think I would be able to write the test. `preference.New()` returns the preference object as set in the preference.yaml file. If I wanted to mock, I would first create a separate file, set the env var `GlobalConfigEnvName` to use this new file, and then run a command as I normally would. I think [this test](https://github.com/openshift/odo/blob/master/pkg/preference/preference_test.go#L16) is a good example of doing this.

Correct me if I misunderstood your concern.

kadel:
I'm running odo with telemetry for the first time and I don't want odo to ask me the telemetry question. I try to execute` odo preference set ConsentTelemetry false` 
```▶ odo preference set ConsentTelemetry false
? Help odo improve by allowing it to collect usage data. Read about our privacy statement: https://developers.redhat.com/article/tool-data-collection. You can change your preference later by changing the ConsentTelemetry preference. (y/N)
```

There has to be away to disable/enable telemetry without asking user interactively.

valaparthvi:
Since the last review, the following changes have been introduced - 
1) Do not prompt the user if the preference command or help was called.
2) Add integration tests
3) Add more comments
4) Make unit tests a little better.
5) Remove all the unnecessary `.Fatal` calls, use `.Error` or `.Info` instead.
6) Only instantiate the segment client if `ConsentTelemetry` was set to true.
7) Fix error when a wrong type of value is passed to `ConsentTelemetry`.
8) Add a unit test for `ConsentTelemetry`.

TODO:
1) Lessen the time it takes to upload data to Segment.
@girishramnani I think now I understand what you meant when you asked if the upload was async. I asked other projects if they were facing similar time issues and a suggestion came up that said this upload must be done asynchronously. I might need your help tackling this one but we can do it in the next sprint.

USAGE_DATA.adoc
```diff
@@ -0,0 +1,28 @@
+[[odo]]
+= `odo` - Developer-focused CLI for Kubernetes and OpenShift
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+[[usage-data]]
+== Usage Data
```
dharmit, line 8: Not sure what this file's purpose is. I mean I get it. But the points under this section look half-baked. It's not clear that x, y and z info will be collected when the preference is set to True. 

Are we collecting the OS type and odo version in use event= when preference is set to `False`?

dharmit (commented):
Very quick review of only a doc.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -4,20 +4,60 @@ import (
...
 }
 
 func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
+	var err error
+	var startTime time.Time
+	cfg, _ := preference.New()
+
+	// Prompt the user to consent for telemetry if a value is not set already
+	// Skip prompting if the preference command is called
+	// This prompt has been placed here so that it does not prompt the user when they call --help
```
dharmit, line 40: Great work here! It's not prompting for any `-h` commands - `odo -h`, `odo service -h`, etc.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -26,9 +66,30 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 	// LogErrorAndExit is used so that we get -o (jsonoutput) for cmds which have json output implemented
 	util.LogErrorAndExit(checkConflictingFlags(cmd), "")
 	// Run completion, validation and run.
-	util.LogErrorAndExit(o.Complete(cmd.Name(), cmd, args), "")
-	util.LogErrorAndExit(o.Validate(), "")
-	util.LogErrorAndExit(o.Run(), "")
+	// Only upload data to segment for completion and validation if a non-nil error is returned.
```
dharmit, line 69: +1 for clear comment. 👍🏾

USAGE_DATA.adoc
valaparthvi: >Are we collecting the OS type and odo version in use event= when preference is set to False?

```
* when `ConsentTelemetry` preference is set to `false`
* system information such as OS type
* `odo` version in use
```
All of the above are different events but I can see why it is confusing. If the user had set the preference to true earlier and then changed it to false, the setting false action will be sent to Segment. The last 2 points about OS type and odo version are unrelated to the setting preference to false point.

I'll work on fully baking it!

pkg/segment/traits.go
```diff
@@ -0,0 +1,11 @@
+package segment
+
+import (
+	"gopkg.in/segmentio/analytics-go.v3"
+	"runtime"
+)
+
+func traits() analytics.Traits {
```
dharmit, line 8: What's the purpose of this function? Are we creating a separate file under `pkg/segment` for just one function? Or do we plan to add more to it going further?

tests/helper/helper_generic.go
```diff
@@ -293,6 +295,9 @@ func CommonBeforeEach() CommonVar {
 	commonVar.Project = commonVar.CliRunner.CreateRandNamespaceProject()
 	commonVar.OriginalWorkingDirectory = Getwd()
 	os.Setenv("GLOBALODOCONFIG", filepath.Join(commonVar.Context, "preference.yaml"))
+	// Set ConsentTelemetry to false so that it does not prompt to set a preference value
+	cfg, _ := preference.New()
+	_ = cfg.SetConfiguration(preference.ConsentTelemetrySetting, "false")
```
dharmit, line 300: I think we should not ignore the error here and fail starting the CI job if there's failure doing this.

tests/integration/cmd_telemetry_test.go
```diff
@@ -0,0 +1,59 @@
+package integration
```
dharmit, line 1: I think tests for this belong to `tests/integration/cmd_pref_config_test.go`.

dharmit (changes requested):

pkg/preference/preference.go
```diff
@@ -452,7 +452,7 @@ func (c *PreferenceInfo) SetConfiguration(parameter string, value string) error
 		case "consenttelemetry":
 			val, err := strconv.ParseBool(strings.ToLower(value))
 			if err != nil {
-				return errors.Wrapf(err, "unable to set %s to %s", parameter, value)
+				return errors.Errorf("unable to set %s to %s, value must be a boolean", parameter, value)
```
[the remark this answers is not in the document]
valaparthvi: I raised another PR(#4540) to fix this and a few other similar issues, and some other typos in preference_test.go

pkg/segment/segment.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/openshift/odo/pkg/version"
+	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
+	"golang.org/x/term"
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+// Writekey will be the API key used to send data to the correct source on Segment
+// This is the dev key, it will be replaced by the prod key when odo is released
+var WriteKey = "4xGV1HV7K2FtUWaoAozSBD7SNCBCJ65U"
```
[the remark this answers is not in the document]
valaparthvi: As discussed on the channel, we're not going to keep 2 separate keys, we'll just use the one for all users.

pkg/odo/genericclioptions/runnable.go
dharmit: I think @kadel & @girishramnani had different views than mine.

pkg/segment/segment.go
```diff
@@ -0,0 +1,156 @@
...
+func addConfigTraits(c *preference.PreferenceInfo, in analytics.Traits) analytics.Traits {
+	// TODO: add more traits later
+	return in
+}
+
+// getUserIdentity returns the anonymous ID if it exists, else creates a new one
+func getUserIdentity(telemetryFilePath string) (string, error) {
+	var id []byte
+
+	if err := os.MkdirAll(filepath.Dir(telemetryFilePath), 0750); err != nil {
```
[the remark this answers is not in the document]
valaparthvi: `os.MkdirAll` checks if the path is a directory, if not, it creates one matching the path.

Once the path, in our case `$HOME/.redhat`, is found/created, we next check if it has the file we need, which is `anonymousID`, if it does not exist, it creates one.

Once the `anonymousID` file is found/created, we next check if the ID within it is a valid UUID with `uuid.Parse` function, if it is not, the function returns nil, and then we create a new valid UUID and write it to the file and return it.

pkg/segment/segment.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	if _, err := os.Stat(telemetryFilePath); !os.IsNotExist(err) {
+		id, err = ioutil.ReadFile(telemetryFilePath)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	// check if the id a valid uuid, if it is not, nil is returned
```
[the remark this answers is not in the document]
valaparthvi: https://github.com/openshift/odo/pull/4482#discussion_r599364067

pkg/segment/traits.go
valaparthvi: Moved it to `segment.go::addConfigTraits`.

dharmit:
> > Is this PR not rebased on top of #4498, or did we miss this check here? Can we have a common function for this kind of issues?
> 
> @dharmit
> It is, but I forgot to make the change for ConsentTelemetry that I did for other preferences. I'll add it to this PR.
> 
> The solution for this kind of issue is to replace Wrapf with Errorf, but also make sure the actual error does not go unnoticed.
> But making this change kinda goes against the use of Wrapf as is described in issue #3786. We need to evaluate where we should use Wrapf and Errorf.
> IMO, if it is possible to pinpoint the cause of the error, we should mention that with Errorf, but in case it is not possible, we go with Wrapf.
> For e.g. [here](https://github.com/openshift/odo/blob/master/pkg/preference/preference.go#L434), we exactly knew why we were getting errors, someone passed a value other than a bool, we could compose an error message that told the user exactly what to do to fix it, in which case Errorf is a good idea. But [here](https://github.com/openshift/odo/blob/master/pkg/preference/preference.go#L465), I could not pinpoint the reason, in which Wrapf might be the way to go.

But I think the piece of code for which I made the comment falls into the `Errorf` category - we know (or can know) what's causing the error. 🤔

USAGE_DATA.adoc
dharmit: Yeah, it's confusing. We could just document what we're doing if the telemetry is enabled. Let's make it simple to understand.

valaparthvi:
> But I think the piece of code for which I made the comment falls into the Errorf category - we know (or can know) what's causing the error. thinking

It does fall under the errorf category, and I fixed that here.

tests/integration/cmd_pref_config_test.go
```diff
@@ -325,4 +327,44 @@ var _ = Describe("odo preference and config command tests", func() {
 			Expect(ok).To(BeFalse())
 		})
 	})
+
+	FContext("When no ConsentTelemetry preference value is set", func() {
```
dharmit, line 331: Forgot to remove the Focus?

tests/integration/cmd_pref_config_test.go
```diff
@@ -325,4 +327,44 @@ var _ = Describe("odo preference and config command tests", func() {
...
+			output = helper.CmdShouldPass("odo", "preference", "unset", "buildtimeout", "-f")
+			Expect(output).ToNot(ContainSubstring(promtMessageSubString))
+		})
+		It("prompt should appear when non-preference command is run", func() {
+			output := helper.CmdShouldPass("odo", "create", "nodejs", "--context", commonVar.Context)
+			Expect(output).To(ContainSubstring(promtMessageSubString))
+		})
+	})
+
+	FContext("Prompt should not appear when", func() {
```
dharmit, line 357: Forgot to remove the Focus?

pkg/segment/segment.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Writekey will be the API key used to send data to the correct source on Segment
+var WriteKey = "R1Z79HadJIrphLoeONZy5uqOjusljSwN"
+
+type Client struct {
+	// SegmentClient helps interact with the segment API
+	SegmentClient analytics.Client
+	// Preference points to the global odo config
+	Preference *preference.PreferenceInfo
+	// TelemetryFilePath points to the file containing anonymousID used for tracking users
```
dharmit, line 30:
```suggestion
	// TelemetryFilePath points to the file containing anonymousID used for tracking odo commands executed by the user
```
Nitpick. But this changes the meaning to a good extent. If someone judges odo's compliance by reading this comment, it would be a bad/incorrect impression that they would gather.

dharmit (changes requested):

dharmit (approved):
/lgtm

pkg/preference/preference_test.go
```diff
@@ -491,6 +491,33 @@ func TestSetConfiguration(t *testing.T) {
...
+			parameter:      ConsentTelemetrySetting,
+			value:          "true",
+			existingConfig: Preference{},
+			wantErr:        false,
+			want:           "true",
+		},
+		{
+			name:      fmt.Sprintf("Case 29: set %s from true to false", ConsentTelemetrySetting),
+			parameter: ConsentTelemetrySetting,
+			value:     "false",
```
dharmit, line 512: We don't want an error in this test case. But the value is a string. Are we sure we don't want an error in this one?

valaparthvi: I think it will be fine, the string will be parsed to bool in  `SetConfiguration`. But I guess I can make it bool instead.

valaparthvi: https://github.com/openshift/odo/pull/4482/commits/b4d64b2865c1806d7431595c5bed22935c258f6f

dharmit:
/hold

@valaparthvi tests are failing with:
```sh
 • Failure [2.070 seconds]
odo preference and config command tests
/go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:15
  When viewing global config
  /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:57
    should get the default global config keys [It]
    /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:58
    Expected
      
        <string>: false
    to be empty
    /go/src/github.com/openshift/odo/tests/integration/cmd_pref_config_test.go:64 
```

I think it's because we're setting `ConsentTelemetry` to `false` in this PR to ensure that telemetry doesn't run in CI. But [in a previous commit](https://github.com/openshift/odo/commit/eed0087ba102eef5d2b868ceb4edc59ba5727724#diff-f55d1dcfa285f4e2e11df9fce169ec0b4256649e7a7c67ab11e33ed667d6dc8fR58.) we changed integration tests to expect its value to be not set to anything.

valaparthvi:
> I think it's because we're setting ConsentTelemetry to false in this PR to ensure that telemetry doesn't run in CI. But in a previous commit we changed integration tests to expect its value to be not set to anything.

Thank you for bringing my attention to this! I've pushed a fix that creates a new context and points to a new preference.yaml so that we can replicate the real-time behaviour in tests.

valaparthvi:
/retest

Prow is unable to fetch golang image from the registry.

>Pulling image registry.svc.ci.openshift.org/openshift/release:golang-1.13 ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: while pulling "docker://registry.svc.ci.openshift.org/openshift/release:golang-1.13" as "registry.svc.ci.openshift.org/openshift/release:golang-1.13": Error initializing source docker://registry.svc.ci.openshift.org/openshift/release:golang-1.13: error pinging docker registry registry.svc.ci.openshift.org: invalid status code from registry 503 (Service Unavailable)
2021/03/25 04:59:29 No custom metadata found and prow metadata already exists. Not updating the metadata.
2021/03/25 04:59:29 Ran for 31s
error: some steps failed:
>* could not run steps: step root failed: the build root failed after 28s with reason PullBuilderImageFailed: Failed pulling builder image.
Pulling image registry.svc.ci.openshift.org/openshift/release:golang-1.13 ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
Warning: Pull failed, retrying in 5s ...
error: build error: failed to pull image: After retrying 2...invalid status code from registry 503 (Service Unavailable)
time="2021-03-25T04:59:29Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:step_failed:building_image_from_source'"

dharmit:
Seeing a [bunch of failures](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4482/pull-ci-openshift-odo-master-v4.7-integration-e2e/1375066143043096576), but none related to the PR, AFAICS.

/hold cancel
/test v4.7-integration-e2e

dharmit:
Just to stop the bot from making too much noise and since below tests are anyway not mandatory atm.

/override  ci/prow/psi-kubernetes-integration-e2e
/override ci/prow/psi-minishift-integration-e2e
/override ci/prow/psi-openshift-integration-e2e

valaparthvi:
The following tests passed locally - 

1.
• Failure [1.398 seconds]
odo devfile create command tests
/go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_create_test.go:19
  When executing odo create with devfile component type argument and --context flag
  /go/src/github.com/openshift/odo/tests/integration/devfile/cmd_devfile_create_test.go:79
    should successfully create the devfile component and show json output for non connected cluster [It] 

I am going to run other tests locally.

dharmit:
I'm seeing `--from-file` scenario break in both test runs. And it's throwing the same error both times:
```
✗  error converting YAML to JSON: yaml: control characters are not allowed
```
I'm looking at this from Operator Hub point of view for both of the below tests:
```
 Summarizing 2 Failures:
[Fail] odo service command tests for OperatorHub When using from-file option [It] should be able to create a service 
[Fail] odo service command tests for OperatorHub When using from-file option [It] should be able to create a service with name passed on CLI 
Ran 16 of 16 Specs in 273.807 seconds
FAIL! -- 14 Passed | 2 Failed | 0 Pending | 0 Skipped
--- FAIL: TestOperatorhub (273.81s)
FAIL
```

dharmit:
> I'm seeing `--from-file` scenario break in both test runs. And it's throwing the same error both times:
> 
> ```
> ✗  error converting YAML to JSON: yaml: control characters are not allowed
> ```
> 
> I'm looking at this from Operator Hub point of view for both of the below tests:
> 
> ```
>  Summarizing 2 Failures:
> [Fail] odo service command tests for OperatorHub When using from-file option [It] should be able to create a service 
> /go/src/github.com/openshift/odo/tests/helper/helper_run.go:34
> [Fail] odo service command tests for OperatorHub When using from-file option [It] should be able to create a service with name passed on CLI 
> /go/src/github.com/openshift/odo/tests/helper/helper_run.go:34
> Ran 16 of 16 Specs in 273.807 seconds
> FAIL! -- 14 Passed | 2 Failed | 0 Pending | 0 Skipped
> --- FAIL: TestOperatorhub (273.81s)
> FAIL
> ```

Passes locally:
```sh
$ ginkgo tests/integration/operatorhub
Running Suite: Operatorhub Suite
================================
Random Seed: 1616742864
Will run 2 of 16 specs

SSSSSSS
• [SLOW TEST:15.834 seconds]
odo service command tests for OperatorHub
/home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:17
  When using from-file option
  /home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:195
    should be able to create a service
    /home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:205
• [SLOW TEST:8.321 seconds]
odo service command tests for OperatorHub
/home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:17
  When using from-file option
  /home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:195
    should be able to create a service with name passed on CLI
    /home/dshah/go/src/github.com/openshift/odo/tests/integration/operatorhub/cmd_service_test.go:235
SSSSSSS
Ran 2 of 16 Specs in 24.155 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 14 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 26.297301008s
Test Suite Passed
Detected Programmatic Focus - setting exit status to 197

$ odo version
odo v2.0.7 (892530570)

Server: https://api.crc.testing:6443
Kubernetes: v1.19.0+1833054
```

dharmit:
/hold

In the test results, I'm seeing 76 occurrences of "Help odo improve..." string. This is likely because few integrations tests don't use the `CommonBeforeEach` helper function. At least that's true for Operator Hub integration tests.

PR #4465 fixes this for Operator Hub, but I think it makes sense to push the change here since this PR is important right now. We need to find if the other tests are failing for the same reason and fix it in this very PR.

@valaparthvi unhold when you are done making changes.

dharmit (approved):
/lgtm

PR #4472, opened 2021-03-02
odo create -o json for devfile also work towards making odo describe work without cluster connection

girishramnani:
**What type of PR is this?**
/kind feature

**What does does this PR do / why we need it**:
json output for `odo create -o json` in devfile

**Which issue(s) this PR fixes**:

fixes https://github.com/openshift/odo/issues/4443
partly https://github.com/openshift/odo/issues/4302

**PR acceptance criteria**:

- [ ] Unit test 

- [X] Integration test 

- [ ] Documentation 

- [X] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:
```

odo create nodejs -o json <- should give json output
odo create nodejs -o json --now <- should also give json output but with more details from cluster

```

PR #4508, opened 2021-03-12
Fixes odo create when cluster access is limited

### LINKED ISSUE #4507: REGRESSION: odo create is not working without cluster
In the current master
```
▶ odo version
odo v2.0.7 (ef867f46f)
```

odo create doesn't work without a cluster connection
```
▶ odo create  java-quarkus
 ✗  You don't have permission to create or set project 'tkral-test' or the project doesn't exist. Please create or set a different project
        odo project create|set <project_name>

```

When using the latest release v2.0.7 it works as it should.
```
▶ odo version
odo v2.0.7 (f6a819c8a)
```

I suspect that it might because some of the changes in 
https://github.com/openshift/odo/pull/4472

/kind bug
/priority critical

mik-dass:
**What type of PR is this?**

/kind bug

It also removes unused parameters from some functions.

Fixes #4507 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- Open the kubeconfig and edit the `server: ` values to a invalid value.
- `odo create` should pass.

### DIFF
--- a/pkg/odo/genericclioptions/client.go
+++ b/pkg/odo/genericclioptions/client.go
@@ -3,16 +3,24 @@ package genericclioptions
 import (
 	"github.com/openshift/odo/pkg/kclient"
 	"github.com/openshift/odo/pkg/occlient"
-	"github.com/spf13/cobra"
 )
 
 // Client returns an oc client "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"d for this command's options
-func Client(command *cobra.Command) (*occlient.Client, error) {
-	return client(command)
+func Client() (*occlient.Client, error) {
+	ocClient, err := client()
+	if err != nil {
+		return nil, err
+	}
+	kClient, err := kClient()
+	if err != nil {
+		return nil, err
+	}
+	ocClient.SetKubeClient(kClient)
+	return ocClient, nil
 }
 
-// client creates an oc client based on the command flags
-func client(command *cobra.Command) (*occlient.Client, error) {
+// client creates an oc client
+func client() (*occlient.Client, error) {
 	client, err := occlient.New()
 	if err != nil {
 		return nil, err
@@ -21,8 +29,8 @@ func client(command *cobra.Command) (*occlient.Client, error) {
 	return client, nil
 }
 
-// kClient creates an kclient based on the command flags
-func kClient(command *cobra.Command) (*kclient.Client, error) {
+// kClient creates an kclient
+func kClient() (*kclient.Client, error) {
 	kClient, err := kclient.New()
 	if err != nil {
 		return nil, err
--- a/pkg/odo/genericclioptions/client.go
+++ b/pkg/odo/genericclioptions/client.go
@@ -5,9 +5,9 @@ import (
 	"github.com/openshift/odo/pkg/occlient"
 )
 
-// Client returns an oc client "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"d for this command's options
+// Client returns an oc client with the kClient set
 func Client() (*occlient.Client, error) {
-	ocClient, err := client()
+	ocClient, err := ocClient()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


mik-dass:
```
error: some steps failed:
  * could not run steps: step integration-e2e failed: could not find secret "integration-e2e-cluster-profile": Get "https://172.30.0.1:443/api/v1/namespaces/ci-op-6jxst21p/secrets/integration-e2e-cluster-profile": http2: client connection lost
```
/retest

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -124,6 +125,33 @@ var _ = Describe("odo devfile create command tests", func() {
...
+			kubeConfigContents = reg.ReplaceAllString(kubeConfigContents, "server: https://not-reachable.com:443")
+
+			// write to a new file which will be used as the new kubeconfig
+			newKubeConfigPath := filepath.Join(commonVar.Context, "newKUBECONFIG")
+			newKubeConfig, err := os.Create(newKubeConfigPath)
+			Expect(err).To(BeNil())
+			_, err = newKubeConfig.WriteString(kubeConfigContents)
+			Expect(err).To(BeNil())
+
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "nodejs", "devfile.yaml"), filepath.Join(devfilePath))
```
valaparthvi, line 146: Why do we copy the devfile here? Wouldn't running the cmd below would create a devfile?

mik-dass: This saves a network call to the registry and instead uses the provided devfile.

pkg/odo/genericclioptions/client.go
```diff
@@ -3,16 +3,24 @@ package genericclioptions
...
+		return nil, err
+	}
+	ocClient.SetKubeClient(kClient)
+	return ocClient, nil
 }
 
-// client creates an oc client based on the command flags
-func client(command *cobra.Command) (*occlient.Client, error) {
+// client creates an oc client
+func client() (*occlient.Client, error) {
```
dharmit, line 23: It's a little confusing that `client` function creates/returns an `occlient.Client` while for kclient we have a different function whose name represents what it does - `kClient()`.

My concern is that it's going to seem more confusing as time passes. Can we rename it to `ocClient()` instead?

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -124,6 +125,33 @@ var _ = Describe("odo devfile create command tests", func() {
...
+			path := os.Getenv("KUBECONFIG")
+
+			// read the contents from the kubeconfig and replace the server entries
+			reg := regexp.MustCompile(`server: .*`)
+			kubeConfigContents, err := helper.ReadFile(path)
+			Expect(err).To(BeNil())
+			kubeConfigContents = reg.ReplaceAllString(kubeConfigContents, "server: https://not-reachable.com:443")
+
+			// write to a new file which will be used as the new kubeconfig
+			newKubeConfigPath := filepath.Join(commonVar.Context, "newKUBECONFIG")
```
dharmit, line 140: Why not use the `WithEnv` approach that was introduced in #4472? https://github.com/openshift/odo/blob/ea12bdf96c4b99b256bcb1aea7d5477afb900683/tests/helper/helper_cmd_wrapper.go#L41

dharmit (changes requested):

tests/integration/devfile/cmd_devfile_create_test.go
mik-dass: Already using it at https://github.com/openshift/odo/pull/4508/files#diff-298ac2fa009dfd61614b93eb6f167f458a022631741170d39af9aa9b8aac16f9R148

pkg/odo/genericclioptions/client.go
mik-dass: Fixed

tests/integration/devfile/cmd_devfile_create_test.go
dharmit: OK. After our discussion on Slack, I think I get why you're doing this. It's about specifically setting an incorrect value in kubeconfig.

dharmit (approved):

Issue #4422, opened 2021-02-09
Two `validation` logs when using odo create --s2i


girishramnani:
/kind bug

## What versions of software are you using?

**Operating System:** Mac OS

**Output of `odo version`:**
```

odo version
odo v2.0.4 (e76643939)

Server: https://api.engint.openshift.com:443
OpenShift: v3.11.154
Kubernetes: v1.11.0+d4cacc0

```

## How did you run odo exactly?
```
odo create --s2i nodejs
```

## Actual behavior
```
do create --s2i nodejs
Validation
Validation
 ✓  Validating component [526ms]

Please use `odo push` command to create the component with source deployed
```

## Expected behavior
One Validation blocks
```
do create --s2i nodejs
Validation
 ✓  Validating component [526ms]

Please use `odo push` command to create the component with source deployed
```

Issue #4197, opened 2020-11-05
odo push does not fail if executed devfile command failed to start


dgolovin:
/kind bug

## What versions of software are you using?

**Operating System:** Linux

**Output of `odo version`:** 2.0.1

## How did you run odo exactly?
```
user@host:~$ git clone https://github.com/dgolovin/nodejs-ex ./nodejs-devfile-issue
Cloning into './nodejs-devfile-issue'...
remote: Enumerating objects: 637, done.
remote: Total 637 (delta 0), reused 0 (delta 0), pack-reused 637
Receiving objects: 100% (637/637), 253.59 KiB | 2.59 MiB/s, done.
Resolving deltas: 100% (246/246), done.

user@host:~$ cd nodejs-devfile-issue/

user@host:~/nodejs-devfile-issue$ odo create nodejs --app app1 --project p1
Validation
 ✓  Checking devfile existence [35224ns]
 ✓  Checking devfile compatibility [55663ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [55176ns]
 ✓  Validating devfile component [94611ns]

Please use `odo push` command to create the component with source deployed
```
## Actual behavior

```
user@host:~/nodejs-devfile-issue$  odo push --debug

Validation
 ✓  Validating the devfile [59256ns]

Creating Kubernetes resources for component nodejs
 ✓  Waiting for component to start [6s]

Applying URL changes
 ✓  URL http-3000: http://http-3000-nodejs-p1.apps-crc.testing/ created

Syncing to component nodejs
 ✓  Checking files for pushing [677080ns]
 ✓  Syncing files to the component [223ms]

Executing devfile commands for component nodejs
 ✓  Executing install command "npm install" [5s]
 ✓  Executing debug command "npm run debug" [1s]

Pushing devfile component nodejs
 ✓  Changes successfully pushed to component

user@host:~/nodejs-devfile-issue$ odo log
time="2020-11-05T09:06:56Z" level=info msg="create process:devrun" 
time="2020-11-05T09:06:56Z" level=info msg="create process:debugrun" 
time="2020-11-05T09:07:02Z" level=debug msg="no auth required" 
time="2020-11-05T09:07:02Z" level=debug msg="succeed to find process:debugrun" 
time="2020-11-05T09:07:02Z" level=info msg="try to start program" program=debugrun 
time="2020-11-05T09:07:02Z" level=info msg="success to start program" program=debugrun 
[...]

PR #4518, opened 2021-03-16
implement devfile conversion ( with --now support) on --s2i without --git flag

### LINKED ISSUE #4281: use Devfiles for s2i components
Use devfile.yml for s2i components instead of LocalConfig.yaml

# Motivation

Odo has two types of components, Devfile and s2i. Currently, each component type has its own separate code path. This means that each command needs to understand how to do the action for Devfile and also for s2i. Technically it is possible to create a Devfile that will mimic what odo is doing with s2i. We already have this in `odo utils convert`. Odo should leverage that logic and start treating s2i components as regular Devfile. This will reduce the odo code base and will make code maintenance a lot simpler.

## `odo create --s2i`

`odo create --s2i` command would no longer generate `LocalConfig` (`./odo/config.yml`). Instead, it should just generate `devfile.yml`, and optionally `./odo/env.yml`) depending on what will be required.

## all other commands

The rest of the commands (like `odo storage`, `odo url`, etc..) should not need to know anything about s2i. For them, it will be just another Devfile component. 

## Acceptance Criteria

- [x] `odo create --s2i` should use the s2i to devfile tool to build the devfile on the fly and follow devfile flow in the future.
- [x] if user has existing `LocalConfig` s2i component it still needs to work
- [ ]  We're going to need to make changes to odo create -h as well.
- [ ]  Finally, we're going to need to update docs that end on odo.dev site as well.

girishramnani:
**What type of PR is this?**
/kind feature

when user provides --s2i flag on odo create we convert the s2i component into devfile and use that for all future workflows

Fixes #4281
Fixes https://github.com/openshift/odo/issues/4527
Fixes https://github.com/openshift/odo/issues/4422

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 


tests/integration/cmd_push_test.go
```diff
@@ -32,13 +32,13 @@ var _ = Describe("odo push command tests", func() {
...
 	Context("Check pod timeout", func() {
 
-		It("Check that pod timeout works and we time out immediately..", func() {
-			helper.CopyExample(filepath.Join("source", "nodejs"), commonVar.Context)
-			helper.CmdShouldPass("odo", "component", "create", "--s2i", "nodejs", cmpName, "--project", commonVar.Project, "--context", commonVar.Context, "--app", appName)
-			helper.CmdShouldPass("odo", "preference", "set", "PushTimeout", "1")
-			output := helper.CmdShouldFail("odo", "push", "--context", commonVar.Context)
-			Expect(output).To(ContainSubstring("waited 1s but couldn't find running pod matching selector"))
-		})
+		// TODO: issue with PushTimeout
```
mohammedzee1000, line 35: I assume this has something to do with the PushTimeout ping on chat. Rest of the code looks good to me.

dharmit (changes requested):

dharmit:
I was trying to get nodejs-ex working with odo built from this PR. It seems like `DEBUG_PORT` value is not getting populated which is preventing the application from starting up.

Here's logs (`odo log`) from successful run when odo is built from current master:
```
+ run_node
++ node --version
+ echo -e 'Using Node.js version: v12.19.1'
Using Node.js version: v12.19.1
+ echo -e 'Environment: \n\tDEV_MODE=true\n\tNODE_ENV=development\n\tDEBUG_PORT=5858'
Environment: 
	DEV_MODE=true
	NODE_ENV=development
	DEBUG_PORT=5858
++ id
+ echo -e 'Running as user uid=1000590000(1000590000) gid=0(root) groups=0(root),1000590000'
Running as user uid=1000590000(1000590000) gid=0(root) groups=0(root),1000590000
+ '[' true == true ']'
+ echo 'Launching via nodemon...'
Launching via nodemon...
+ exec npx nodemon --inspect=5858
[nodemon] master: <none>
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `node --inspect=5858 server.js`
```

And here's from unsuccessful run:
```
Using Node.js version: v10.24.0
Environment: 
	DEV_MODE=true
	NODE_ENV=development
	DEBUG_PORT=
+ echo -e 'Using Node.js version: v10.24.0'
+ echo -e 'Environment: \n\tDEV_MODE=true\n\tNODE_ENV=development\n\tDEBUG_PORT='
++ id
Running as user uid=1000590000(1000590000) gid=0(root) groups=0(root),1000590000
Launching via nodemon...
+ echo -e 'Running as user uid=1000590000(1000590000) gid=0(root) groups=0(root),1000590000'
+ '[' true == true ']'
+ echo 'Launching via nodemon...'
+ exec npx nodemon --inspect=
[nodemon] master: <none>
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `node --inspect= server.js`
node: --inspect= requires an argument
[nodemon] app crashed - waiting for file changes before starting...
time="2021-03-19T13:31:35Z" level=debug msg="wait program exit" program=devrun
```

Since the `DEBUG_PORT` env var is not getting populated, the component is failing to start correctly.

dharmit:
@girishramnani @kadel would https://github.com/openshift/odo/issues/4197 help take care of the kind of issue I found here -  missing `DEBUG_PORT` causing failure in pod to start well?

dharmit:
@kadel @girishramnani Not sure if this is expected behaviour.

`odo create --s2i --port` works without an error. But the resulting devfile.yaml doesn't have any `endpoints` in it.
```sh
$ odo create --s2i --port 8080 nodejs
Validation
 ✓  Validating component [4ms]
Conversion
 ✓  Successfully generated devfile.yaml and env.yaml for provided S2I component

```
Should we track this in a separate issue or should this be fixed via this PR?

tests/e2escenarios/e2e_images_test.go
```diff
@@ -36,21 +36,17 @@ var _ = Describe("odo supported images e2e tests", func() {
...
+		helper.CmdShouldPass("odo", "create", "--s2i", cmpType, cmpName, "--project", project, "--context", context, "--app", appName)
 
 		// push component and validate
 		helper.CmdShouldPass("odo", "push", "--context", context)
 		cmpList := helper.CmdShouldPass("odo", "list", "--context", context)
 		Expect(cmpList).To(ContainSubstring(srcType + "-app"))
 
-		// create a url
-		helper.CmdShouldPass("odo", "url", "create", "--port", "8080", "--context", context)
-		helper.CmdShouldPass("odo", "push", "--context", context)
```
kadel, line 53: We should have a second push in the testing scenario. 
In past, we had a situation where the second `odo push` broke a component. 
This scenario should have a second push in it. For example, you should change something using `odo config` and do `odo push`  again and verify that the application is still running even after the second push.

tests/helper/odo_utils.go
```diff
@@ -59,6 +59,24 @@ func DetermineRouteURL(context string) string {
 	return routeURL(context)
 }
 
+// DetermineRouteURLs takes context path as argument and returns the URLs
+// where the current component exposes it's service, these URL can
+// then be used in order to interact with the deployed service running in Openshift
+func DetermineRouteURLs(context string) []string {
```
kadel, line 65: There already is an almost identical function `routeURL` .
This creates duplicated code.
Would it make sense to delete the old `routeURL` and use new `DetermineRouteURLs` everywhere?

tests/integration/cmd_pref_config_test.go
```diff
@@ -186,7 +184,7 @@ var _ = Describe("odo preference and config command tests", func() {
 					paramValue: "https://github.com/sclorg/nodejs-ex",
 				},
 			}
-			helper.CmdShouldPass("odo", "create", "--s2i", "nodejs", "--project", commonVar.Project)
+			helper.CmdShouldPass("odo", "create", "--s2i", "nodejs", "--project", commonVar.Project, "--git", "https://github.com/sclorg/nodejs-ex")
```
kadel, line 187: I have noticed that you are using 3 different repos in tests. Is there a reason for that?
Could we use this one everywhere?

tests/e2escenarios/e2e_images_test.go
girishramnani: added

tests/integration/cmd_debug_test.go
```diff
@@ -60,9 +59,10 @@ var _ = Describe("odo debug command tests", func() {
 
 		It("should expect a ws connection when tried to connect on different debug port locally and remotely", func() {
 			helper.CopyExample(filepath.Join("source", "nodejs"), commonVar.Context)
-			helper.CmdShouldPass("odo", "component", "create", "--s2i", "nodejs:latest", "--project", commonVar.Project, "--context", commonVar.Context)
-			helper.CmdShouldPass("odo", "config", "set", "--force", "DebugPort", "9292", "--context", commonVar.Context)
-			dbgPort := helper.GetConfigValueWithContext("DebugPort", commonVar.Context)
+			helper.CmdShouldPass("odo", "component", "create", "--s2i", "nodejs", "--project", commonVar.Project, "--context", commonVar.Context)
+			helper.CmdShouldPass("odo", "env", "set", "--force", "DebugPort", "9292", "--context", commonVar.Context)
```
kadel, line 63: why we are setting debug port twice?

tests/integration/cmd_debug_test.go
```diff
@@ -77,22 +77,6 @@ var _ = Describe("odo debug command tests", func() {
 			stopChannel <- true
 		})
 
-		It("should expect a ws connection when tried to connect on default debug port locally", func() {
```
kadel, line 80: Why was this removed? This is still a valid  scenario

tests/integration/cmd_debug_test.go
```diff
@@ -119,43 +103,5 @@ var _ = Describe("odo debug command tests", func() {
 			stopChannel <- true
 		})
 
-		It("should start a debug session and run debug info on a closed debug session", func() {
```
kadel, line 122: Similarly, as above, this should be a valid scenario.

tests/integration/cmd_push_test.go
```diff
@@ -270,86 +163,10 @@ var _ = Describe("odo push command tests", func() {
 			Expect(output).To(Not(ContainSubstring("No file changes detected, skipping build")))
 		})
 
-		It("should push only the modified files", func() {
```
kadel, line 273: this is a valid scenario that we should check.

dharmit:
IMO, we should update docs on odo.dev as a part of this PR. It's fundamentally changing the behaviour of `odo create --s2i`. Maybe that would mean renaming the [Migrating S2I (Source-to-Image) components to Devfile components](https://odo.dev/docs/s2i-to-devfile/) to "Creating components using s2i images" and changing the content accordingly.

@kadel wdyt?

tests/integration/cmd_debug_test.go
girishramnani: Tests that have been removed without a comment have a devfile counterpart already present hence can be deleted

tests/integration/cmd_debug_test.go
girishramnani: Tests that have been removed without a comment have a devfile counterpart already present hence can be deleted

tests/integration/cmd_push_test.go
girishramnani: Tests that have been removed without a comment have a devfile counterpart already present hence can be deleted

tests/integration/cmd_debug_test.go
girishramnani: because the `env set DebugPort 9292` doesn't work. I am debugging that and would file a bug report if confirmed

tests/integration/cmd_debug_test.go
kadel: ok makes sense

tests/integration/cmd_debug_test.go
girishramnani: opened an issue https://github.com/openshift/odo/issues/4615, confirming the setting of that value isn't working

tests/integration/cmd_push_test.go
kadel: I was not able to find a test that would verify that `odo push` syncs only modified files. 
Would it make sense to convert this test to Devfile scenario?

girishramnani: maybe

girishramnani: I can do that

tests/integration/cmd_debug_test.go
```diff
@@ -1,108 +0,0 @@
...
-	})
-
-	// Clean up after the test
-	// This is run after every Spec (It)
-	var _ = AfterEach(func() {
-		helper.CommonAfterEach(commonVar)
-	})
-
-	Context("odo debug on a nodejs:latest component", func() {
-		It("check that machine output debug information works", func() {
```
girishramnani, line 34: we have exact same test in devfile already

tests/integration/cmd_debug_test.go
```diff
@@ -1,108 +0,0 @@
...
-			// 400 response expected because the endpoint expects a websocket request and we are doing a HTTP GET
-			// We are just using this to validate if nodejs agent is listening on the other side
-			helper.HttpWaitForWithStatus("http://localhost:5050", "WebSockets request was expected", 12, 5, 400)
-			stopChannel <- true
-		})
-
-	})
-
-	Context("odo debug info should work on a odo component", func() {
-		It("should start a debug session and run debug info on a running debug session", func() {
```
girishramnani, line 99: we have exactly same test in devfile

tests/integration/cmd_push_test.go
girishramnani: Can we do this in a follow up https://github.com/openshift/odo/issues/4618

girishramnani:
> IMO, we should update docs on odo.dev as a part of this PR. It's fundamentally changing the behaviour of `odo create --s2i`. Maybe that would mean renaming the [Migrating S2I (Source-to-Image) components to Devfile components](https://odo.dev/docs/s2i-to-devfile/) to "Creating components using s2i images" and changing the content accordingly.
> 
> @kadel wdyt?

Yeah I haven't thought of how much of the doc should change

tests/integration/cmd_pref_config_test.go
girishramnani: I am not sure which one is latest, there is one in odo-devfiles repo too. Shouldn't we consider that?

tests/integration/cmd_push_test.go
kadel: you are removing the test case that tests one of the `odo push` primary functions :-/ 
I would prefer to have it in this PR. But I would say it is ok if you do it in follow up, but it needs to come right after this.

tests/helper/odo_utils.go
girishramnani: I will do this in a follow up

kadel: why? You are adding a duplicate function just so you can later delete it?

tests/integration/component.go
```diff
@@ -106,76 +94,53 @@ func componentTests(args ...string) {
...
-			expected, err := helper.Unindented(fmt.Sprintf(`{"kind":"Component","apiVersion":"odo.dev/v1alpha1","metadata":{"name":"nodejs","namespace":"%s","creationTimestamp":null},"spec":{"app":"app","type":"nodejs","sourceType": "local","ports":["8080/TCP"]},"status":{"context":"%s","state":"Pushed"}}`, commonVar.Project, contextPath))
-			Expect(err).Should(BeNil())
-			Expect(actual).Should(ContainSubstring(expected))
-			// this orders the json
-			expected, err = helper.Unindented(fmt.Sprintf(`{"kind":"Component","apiVersion":"odo.dev/v1alpha1","metadata":{"name":"python","namespace":"%s","creationTimestamp":null},"spec":{"app":"app","type":"python","sourceType": "local","ports":["8080/TCP"]},"status":{"context":"%s","state":"Pushed"}}`, project2, contextPath2))
-			Expect(err).Should(BeNil())
-			Expect(actual).Should(ContainSubstring(expected))
-
-		})
+		// TODO: Fix later
```
kadel, line 97: what is borken?

tests/e2escenarios/e2e_images_test.go
```diff
@@ -111,11 +99,6 @@ var _ = Describe("odo supported images e2e tests", func() {
 			verifySupportedImage("redhat-openjdk-18/openjdk18-openshift:latest", "openjdk", "java:8", commonVar.Project, appName, commonVar.Context)
 		})
 
-		It("Should be able to verify the openjdk-11-rhel7 image", func() {
```
kadel, line 114: why do you want to remove this?

girishramnani: because there is a openjdk-11 already present

tests/integration/component.go
girishramnani: json outputs are different has changed so the tests are failing, will fix in a follow up

tests/integration/cmd_pref_config_test.go
kadel: I would go with https://github.com/sclorg/nodejs-ex as these scenarios are testing --s2i so it would be good to use an example that was built for s2i.

kadel: or https://github.com/openshift/nodejs-ex depending on which one is newer.

tests/e2escenarios/e2e_images_test.go
girishramnani: added them back

girishramnani: or we run them in the periodic job or we fix them if someone complains about them.

kadel: > because there is a openjdk-11 already present

but that is a different image :-/ I know that this is a mess with all the different images. But this is why we need those tests because the images are build differently and we need to verify that odo works with them.

tests/integration/cmd_pref_config_test.go
girishramnani: both are same

girishramnani: https://github.com/openshift/nodejs-ex points to https://github.com/sclorg/nodejs-ex

girishramnani: now only two kinds of repos are being used - https://github.com/openshift/nodejs-ex and https://github.com/sclorg/nodejs-ex
But both are same

tests/e2escenarios/e2e_images_test.go
girishramnani: I added the images back. Removed a rhel-7 one as there is rhel-8 for same java version

tests/helper/odo_utils.go
girishramnani: This is not a duplicate function - this one returns list of URLs, `routeURL` returns just the first one, so if you want me to use `DetermineRouteURLs` then that would require work, its not a simple search and replace

girishramnani: And as the function signature is different I cannot breakout the code into a helper function as well.

prietyc123:
Failure on kuberenetes:

```
[ssh:Fedora 32] Command Error failed to wait for command completion failed to wait ssh command: <nil>, failing gracefully 
INFO[2021-04-15T08:51:05Z] [ssh:Fedora 32] !!!run failed, skipping!!!   
INFO[2021-04-15T08:51:05Z]                                              
INFO[2021-04-15T08:51:05Z] 2021/04/15 08:50:56 : failed due to err Failed the test, see logs above ^ 
INFO[2021-04-15T08:51:05Z] {"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-04-15T08:50:56Z"} 
INFO[2021-04-15T08:51:05Z] error: failed to execute wrapped command: exit status 1 
```
Its openstack issue. Sometimes VM creation takes more than 1 min and we are trying to iterate on `vm ssh` for 1 min. To fix this failure options are:
1. We iterate for more than 1 min but yes there should be a limit set.
2. We increase the default retryloop in ci-firewall.
 I am looking for suggestions to set the limit. cc @mohammedzee1000 @kadel 

girishramnani:
## Things that could break for vs code adapter

### All the below points only apply on s2i components. which would work as devfiles after this PR is merged.

- [ ] odo url create after `odo create --s2i` would fail, you shouldn't do that as odo would create url directly without asking https://github.com/openshift/odo/issues/4621
- [ ] wildfly and dotnet s2i components will not anymore https://github.com/openshift/odo/issues/4623
- [ ] now flag for `odo env` wont work https://github.com/openshift/odo/issues/4594
- [ ] odo delete --wait wouldn't work https://github.com/openshift/odo/issues/4593
- [ ] `oc` based checks that checks for image streams and deployment configs would fail - https://github.com/redhat-developer/vscode-openshift-tools/blob/3cd96beac9c6a27360947c49b34d7e33d022e3fe/test/unit/oc.test.ts#L87
- [ ] json output of odo list, describe and create would be different from before as s2i component creation would yield into a devfile component now.
- [ ] `odo env set DebugPort` wont work as there is no "debug" specific commands for s2i converted devfiles. But I have added a work around so basic debugging would work. and you can change the port via `odo config set --env DEBUG_PORT`

girishramnani:
🥳 Finally all tests are passing

girishramnani:
Some questions asked by adapters team and their replies

>  Why odo create URL would fail as it is allowed for devfile ?

it won’t fail in the sense that if you tried the conventional s2i approach and try to create odo url create it would fail with url for 8080 port already present  as there would already be one for you. issue - https://github.com/openshift/odo/issues/4621

> is odo env --now and odo delete --wait permanent or temporary as we already have the issue for “normal” devfile components

Each and every change listed if they have an issue at the end of it will be fixed as a follow up  e.g. https://github.com/openshift/odo/issues/4594 for odo env and https://github.com/openshift/odo/issues/4593 for odo delete

> Not sure to understand the status of the debug

odo env set DebugPort wont work, instead you would need to use odo config set --env DEBUG_PORT  - this is because the s2i to devfile converted devfiles dont have a debug type command defined in them. We would fix this too.

> So every component will now be reported as devfile component but what about existing S2I components

They should work as is mostly because I tried them and they work, maybe try odo list because that one is quite complex to work across the board.

> If oc based checks are not going to work, is there an alternative odo support around it ?

I wouldn't say you cannot use oc based checks but they would break because now s2i components being converted to devfile would generate a deployment but the oc would try to find a DeploymentConfig.

> wildfly and dotnet s2i components will not work anymore - This one is specific to the next release, can be fixed in the upcoming releases ?

I think dotnet was mostly fixed but yeah the goal is to get both working asap and its the most important one right now.

dgolovin:
Main question how we are going to deal with existing components pushed or not pushed after extension update installed with new odo version. If we cannot distinguish components using `odo list` how are we going to detect them? If odo cannot work with them transparently (so user would not see any problems), then we have a problem and need a way out of it.

jeffmaury:
My understanding is that newly created components will be reported as devfile whereas older ones will still be S2I

girishramnani:
> My understanding is that newly created components will be reported as devfile whereas older ones will still be S2I

Yes that is correct, all the s2i code is still there so existing s2i components would work as is.

dharmit:
> Main question how we are going to deal with existing components pushed or not pushed after extension update installed with new odo version. If we cannot distinguish components using `odo list` how are we going to detect them? If odo cannot work with them transparently (so user would not see any problems), then we have a problem and need a way out of it.

* `odo list` and `odo list -o json` show existing "Pushed" components under s2i components.
* A component that's not pushed, shows up as "Pushed" (this is incorrect behaviour).

However, in both cases, an existing component created with `odo create --s2i` shows up as s2i component.

jeffmaury:
So does that mean that newly S2I components will be reported as devfile only after they have been pushed ? That's a major change as our tooling does not handle type change after it has been discovered

dharmit:
> So does that mean that newly S2I components will be reported as devfile only after they have been pushed ? 

Yes.

> That's a major change as our tooling does not handle type change after it has been discovered.

We're pushing this change for 2.1.1 release. 2.1 release is already on the way without this change.

How can we help make this less painful? Showing components created with `odo create --s2i` under "S2I Component" in `odo list` and `odo list -o json` outputs would mean to track this somewhere other than devfile.yaml. It might be doable, but I'm not sure if that's the direction we want to take. From VS Code tooling point of view, do we differentiate between s2i & devfile components? I'm asking this because in odo we're trying to move towards having only devfile components.

tests/integration/cmd_pref_config_test.go
prietyc123: IIRC we were always trying not to use external dependent repo in test, no? I am saying this because earlier we got into trouble due to this. So why are we using `--git` again 😕

tests/integration/cmd_push_test.go
```diff
@@ -159,21 +73,6 @@ var _ = Describe("odo push command tests", func() {
 	})
 
 	Context("when push command is executed", func() {
-		It("should not build when no changes are detected in the directory and build when a file change is detected", func() {
```
prietyc123, line 162: Why are we removing this?

tests/integration/cmd_push_test.go
```diff
@@ -210,7 +106,7 @@ var _ = Describe("odo push command tests", func() {
 			helper.CmdShouldPass("odo", "push", "--context", commonVar.Context, "-v4")
 
 			// Then check to see if it's truly been deleted
-			stdOut = oc.ExecListDir(podName, commonVar.Project, dir)
+			stdOut = oc.ExecListDir(podName, commonVar.Project, "/tmp/projects")
```
prietyc123, line 109: Why are we hardcoding the dir part?

tests/integration/cmd_push_test.go
prietyc123: > But I would say it is ok if you do it in follow up, but it needs to come right after this.

TBH I don't understand how does it make sense deleting the test in a pr then adding it in follow up pr. Is there a lot work involved?

prietyc123 (changes requested):

tests/integration/cmd_storage_test.go
```diff
@@ -32,97 +30,6 @@ var _ = Describe("odo storage command tests", func() {
 		})
 	})
 
-	Context("when running storage command without required flag(s)", func() {
```
prietyc123, line 35: Is there any reason we are removing this test?

tests/integration/debug/cmd_debug_test.go
```diff
@@ -32,8 +32,12 @@ var _ = Describe("odo debug command serial tests", func() {
 	})
 
 	It("should auto-select a local debug port when the given local port is occupied", func() {
+		projName := helper.GetCliRunner().CreateRandNamespaceProjectOfLength(5)
```
prietyc123, line 35: Could you please add it in `JustBeforeEach`. Mostly `It` spec consists only the test steps.

prietyc123 (changes requested):

mohitsuman:
> > My understanding is that newly created components will be reported as devfile whereas older ones will still be S2I
> 
> Yes that is correct, all the s2i code is still there so existing s2i components would work as is.

Given a scenario, a user migrates to new version of extension support this latest odo, and the components are migrated as s2i component. Now if the user does undeploy and again pushes that component, the component changes to s2i or devfile ? 

If it changes to devfile (as per the new workflow being suggested in the PR), then the scenario of s2i and devfile components is going to be confusing.

mohitsuman:
> How can we help make this less painful? Showing components created with `odo create --s2i` under "S2I Component" in `odo list` and `odo list -o json` outputs would mean to track this somewhere other than devfile.yaml. It might be doable, but I'm not sure if that's the direction we want to take. From VS Code tooling point of view, do we differentiate between s2i & devfile components? I'm asking this because in odo we're trying to move towards having only devfile components.

For many developers, they want have the support both for s2i and devfile components using the extension. Is there any specific discussion or reason why odo want to support only devfile components ?

tests/integration/cmd_storage_test.go
girishramnani: Tests that have been removed without a comment have a devfile counterpart already present hence can be deleted

tests/integration/cmd_push_test.go
girishramnani: Because it doesn’t change

tests/integration/cmd_push_test.go
girishramnani: Tests that have been removed without a comment have a devfile counterpart already present hence can be deleted

tests/e2escenarios/e2e_images_test.go
```diff
@@ -36,21 +36,18 @@ var _ = Describe("odo supported images e2e tests", func() {
...
-		helper.CmdShouldPass("odo", "config", "set", "maxmemory", "700Mi", "--context", context)
+		helper.CmdShouldPass("odo", "create", "--s2i", cmpType, cmpName, "--project", project, "--context", context, "--app", appName)
 
 		// push component and validate
 		helper.CmdShouldPass("odo", "push", "--context", context)
 		cmpList := helper.CmdShouldPass("odo", "list", "--context", context)
 		Expect(cmpList).To(ContainSubstring(srcType + "-app"))
-
-		// create a url
-		helper.CmdShouldPass("odo", "url", "create", "--port", "8080", "--context", context)
```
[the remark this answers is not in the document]
girishramnani: https://github.com/openshift/odo/issues/4621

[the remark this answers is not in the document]
girishramnani: It has been documented multiple times on the issue as well as PR

dharmit:
> For many developers, they want have the support both for s2i and devfile components using the extension.

The support isn't going away. odo is only converting those components to devfile components.

> Is there any specific discussion or reason why odo want to support only devfile components ?

#4281

pkg/odo/cli/component/create_helpers.go
```diff
@@ -68,3 +71,22 @@ func decideAndDownloadStarterProject(devObj parser.DevfileObj, projectPassed str
 
 	return component.DownloadStarterProject(starterProject, token, contextDir)
 }
+
+func (co *CreateOptions) DevfileJSON() error {
```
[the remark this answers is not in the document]
girishramnani: It’s just an extraction from the main loop

girishramnani:
> * A component that's not pushed, shows up as "Pushed" (this is incorrect behaviour).

We can record this in an issue because changes in this PR shouldn’t have affected this behaviour

tests/e2escenarios/e2e_source_test.go
```diff
@@ -33,50 +33,51 @@ var _ = Describe("odo source e2e tests", func() {
...
-			helper.CmdShouldPass("odo", "push", "--context", commonVar.Context)
-			routeURL := helper.DetermineRouteURL(commonVar.Context)
-
-			// Ping said URL
-			helper.HttpWaitFor(routeURL, "dotnet", 30, 1)
-
-			// Delete the component
-			helper.CmdShouldPass("odo", "app", "delete", "app", "--project", commonVar.Project, "-f")
-		})
+		// issue https://github.com/openshift/odo/issues/4623
```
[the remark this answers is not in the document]
girishramnani: No, we have to merge this first

Makefile
```diff
@@ -9,7 +9,7 @@ COMMON_LDFLAGS := -X $(PROJECT)/pkg/version.GITCOMMIT=$(GITCOMMIT)
 BUILD_FLAGS := -mod=vendor -ldflags="$(COMMON_LDFLAGS)"
 CROSS_BUILD_FLAGS := -mod=vendor -ldflags="-s -w -X $(PROJECT)/pkg/segment.writeKey=R1Z79HadJIrphLoeONZy5uqOjusljSwN $(COMMON_LDFLAGS)"
 FILES := odo dist
-TIMEOUT ?= 7200s
+TIMEOUT ?= 14400s
```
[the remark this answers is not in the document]
girishramnani: Yes

tests/e2escenarios/e2e_images_test.go
girishramnani: > I added the images back. ~~ Removed a rhel-7 one as there is rhel-8 for same java version~~
> All images were added back

Please read carefully @prietyc123

tests/integration/cmd_push_test.go
girishramnani: actually it does require some work but I am just trying to fix this PR in smaller chunks so that work moves along and next PRs will address smaller issues. This would also dependent team to use this change soon while we get feedback and I would get some breathing room 🙂

tests/e2escenarios/e2e_images_test.go
girishramnani: Typo from my end that didn't render the markdown as cancelled line, I apologize.

pkg/odo/cli/component/create_helpers.go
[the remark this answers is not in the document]
girishramnani: added comments

tests/integration/debug/cmd_debug_test.go
girishramnani: done

dgolovin:
> From VS Code tooling point of view, do we differentiate between s2i & devfile components?

VSCode Ext detects S2I based on `spec.sourceType` attribute in 'odo list -o json' output.

girishramnani:
My suggestion would be to please use this branch ( should be merged to master soon too ) with vs code openshift plugin and if you can find some pain points, I can fix them before the release is around? Can we condsider trying this branch @dgolovin ?

tests/helper/helper_kubectl.go
```diff
@@ -141,6 +141,12 @@ func (kubectl KubectlRunner) GetServices(namespace string) string {
 // CreateRandNamespaceProject create new project
 func (kubectl KubectlRunner) CreateRandNamespaceProject() string {
 	projectName := SetProjectName()
+	kubectl.createRandNamespaceProject(projectName)
+	return projectName
+}
+
+// CreateRandNamespaceProject create new project
```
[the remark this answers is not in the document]
girishramnani: done

tests/helper/helper_kubectl.go
```diff
@@ -149,6 +155,13 @@ func (kubectl KubectlRunner) CreateRandNamespaceProject() string {
 	return projectName
 }
 
+// CreateRandNamespaceProject create new project
```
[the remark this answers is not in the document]
girishramnani: done

tests/helper/odo_utils.go
girishramnani: Both functions servered a different purpose but I still removed duplicate code.

tests/integration/cmd_push_test.go
girishramnani: done, added this function back and some helper functions that needed implementation

tests/e2escenarios/e2e_source_test.go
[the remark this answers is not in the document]
girishramnani: We cannot strive for perfection for such a huge breaking change or else the PR would never get merged. So please lets try to merge this PR 🙂 .

tests/helper/helper_oc.go
```diff
@@ -565,6 +565,19 @@ func (oc OcRunner) VerifyResourceDeleted(resourceType, resourceName, namespace s
 // CreateRandNamespaceProject create new project
 func (oc OcRunner) CreateRandNamespaceProject() string {
 	projectName := SetProjectName()
+	oc.createRandNamespaceProject(projectName)
+	return projectName
+}
+
+// CreateRandNamespaceProject create new project
```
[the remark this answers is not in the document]
girishramnani: done

tests/helper/helper_oc.go
```diff
@@ -565,6 +565,19 @@ func (oc OcRunner) VerifyResourceDeleted(resourceType, resourceName, namespace s
...
+func (oc OcRunner) CreateRandNamespaceProjectOfLength(i int) string {
+	projectName := RandString(i)
+	oc.createRandNamespaceProject(projectName)
+	return projectName
+}
+
+// CreateRandNamespaceProject create new project
```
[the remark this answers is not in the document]
girishramnani: done

tests/integration/cmd_push_test.go
girishramnani: still made that use the constant from the convert.go now

tests/integration/cmd_push_test.go
girishramnani: https://github.com/openshift/odo/pull/4518/commits/ec833148b4082d0e9baca1e70711634ce1d4fe1d#diff-959d7173037ca057614f19d5dec825f383d808739d9d930933a6d6203405d203R156

girishramnani:
Blocked on https://github.com/openshift/odo/pull/4637,
Test failing on unavailability of `kubectl` binary

```
odo push command tests                       
INFO[2021-04-17T20:24:45Z] /go/src/github.com/openshift/odo/tests/integration/cmd_push_test.go:15 
INFO[2021-04-17T20:24:45Z]   when push command is executed              
INFO[2021-04-17T20:24:45Z]   /go/src/github.com/openshift/odo/tests/integration/cmd_push_test.go:76 
INFO[2021-04-17T20:24:45Z]     should push only the modified files [It] 
INFO[2021-04-17T20:24:45Z]     /go/src/github.com/openshift/odo/tests/integration/cmd_push_test.go:156 
INFO[2021-04-17T20:24:45Z]                                              
INFO[2021-04-17T20:24:45Z]     Unexpected error:                   
INFO[2021-04-17T20:24:45Z]         <*exec.Error | 0xc0009dee00>: {      
INFO[2021-04-17T20:24:45Z]             Name: "kubectl",                 
INFO[2021-04-17T20:24:45Z]             Err: {                           
INFO[2021-04-17T20:24:45Z]                 s: "executable file not found in $PATH", 
INFO[2021-04-17T20:24:45Z]             },                               
INFO[2021-04-17T20:24:45Z]         }                                    
INFO[2021-04-17T20:24:45Z]         exec: "kubectl": executable file not found in $PATH 
INFO[2021-04-17T20:24:45Z]     occurred                             
INFO[2021-04-17T20:24:45Z]                                              
```
@anandrkskd @rnapoles-rh PTAL

dharmit (approved):

dharmit:
/hold 

remove hold once 2.1 release is out.

Issue #4523, opened 2021-03-17
ci-firewall --setupscript and --runscript should take script parameter arg


prietyc123:
/kind bug

## What versions of software are you using?

**Operating System:**
All supported

**Output of `odo version`:**
master

## How did you run odo exactly?

Executing ci-firewall request with argument in `--setupscript ` and `--runscript` doesn't accept the arg

```
./ci-firewall request --sendqueue amqp.ci.queue.minikube.send --sendtopic amqp.ci.topic.minikube.send --setupscript scripts/minikube-minishift-setup-env.sh minikube --runscript scripts/minikube-minishift-all-tests.sh minikube --timeout 2h15m
Following jenkins build 486
[ssh:Fedora 32] !!!running tests on node Fedora 32 via ssh!!!
[ssh:Fedora 32] Executing command [rm -rf amqp_ci_rcv_odo-minikube-pr-build_PR_4521]
[ssh:Fedora 32] Executing command [mkdir -p amqp_ci_rcv_odo-minikube-pr-build_PR_4521/repo]
[ssh:Fedora 32] Executing command [git clone https://github.com/openshift/odo amqp_ci_rcv_odo-minikube-pr-build_PR_4521/repo]
[ssh:Fedora 32] Cloning into 'amqp_ci_rcv_odo-minikube-pr-build_PR_4521/repo'...
[ssh:Fedora 32] Executing command [git fetch origin pull/4521/head:pr4521]
[ssh:Fedora 32] From https://github.com/openshift/odo
[ssh:Fedora 32]  * [new ref]             refs/pull/4521/head -> pr4521
[ssh:Fedora 32] Executing command [git checkout master]
[ssh:Fedora 32] Your branch is up to date with 'origin/master'.
[ssh:Fedora 32] Already on 'master'
[ssh:Fedora 32] Executing command [git merge pr4521 --no-edit]
[ssh:Fedora 32] Updating e04d99e86..25f8e7f78
[ssh:Fedora 32] Fast-forward
[ssh:Fedora 32]  Makefile                                           |  46 ----
[ssh:Fedora 32]  tests/helper/helper_cli.go                         |   2 +-
[ssh:Fedora 32]  tests/helper/helper_docker.go                      | 270 -------------------
[ssh:Fedora 32]  .../docker/cmd_docker_devfile_catalog_test.go      |  72 -----
[ssh:Fedora 32]  .../docker/cmd_docker_devfile_delete_test.go       | 168 ------------
[...]

PR #4530, opened 2021-03-18
Passing minikube and minishift parameter using env

prietyc123:
**What type of PR is this?**

/kind bug

Temporary fix for https://github.com/openshift/odo/issues/4523

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

kuberenetes job should not be blocked on parameter thing

### DIFF
--- a/scripts/minici-presubmit-all-tests.sh
+++ b/scripts/minici-presubmit-all-tests.sh
@@ -17,16 +17,16 @@ case $1 in
         export JOB_NAME="odo-minikube-pr-build"
         export SENDQUEUE="amqp.ci.queue.minikube.send"
         export SENDTOPIC="amqp.ci.topic.minikube.send"
-        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh minikube"
-        export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh minikube"
+        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh"
+        export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh"
         export TIMEOUT="2h15m"
         ;;
     minishift)
         export JOB_NAME="odo-minishift-pr-tests"
         export SENDQUEUE="amqp.ci.queue.minishift.send"
         export SENDTOPIC="amqp.ci.topic.minishift.send"
-        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh minishift"
-        export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh minishift"
+        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh"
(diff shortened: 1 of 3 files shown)


scripts/minici-presubmit-all-tests.sh
```diff
@@ -17,16 +17,16 @@ case $1 in
...
         export JOB_NAME="odo-minishift-pr-tests"
         export SENDQUEUE="amqp.ci.queue.minishift.send"
         export SENDTOPIC="amqp.ci.topic.minishift.send"
-        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh minishift"
-        export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh minishift"
+        export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh"
+        export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh"
         export TIMEOUT="4h00m"
         ;;
     *)
```
rnapoles-rh, line 32: @prietyc123  Where do you set the value for CLUSTER?
Should need to export CLUSTER="minikuke" and CLUSTER="minishift" within each case block as applicable.
Also there is no need to remove parameters passed in lines 20-21 and 28-29 as this does not have an impact. If the firewall will be updated then later on we do not need to change those lines again.

rnapoles-rh (changes requested):

scripts/minici-presubmit-all-tests.sh
prietyc123: > Where do you set the value for CLUSTER?

I directly set env inside jenkins and calling it into scripts as it works for `${OCP4X_DOWNLOAD_URL}` https://github.com/openshift/odo/blob/master/scripts/run_all_tests.sh#L30 

> Also there is no need to remove parameters passed in lines 20-21 and 28-29 as this does not have an impact. If the firewall will be updated then later on we do not need to change those lines again.

I think we should change it because anyway we will be doing changes with the `case` once ci-firewall support parameter and I would prefer current approach of testing rather talking about future 🙂 IMO Passing the parameter might confuse people.

> [ssh:Fedora 32] go: found github.com/onsi/ginkgo/ginkgo in github.com/onsi/ginkgo v1.14.0
> [ssh:Fedora 32] ++ case ${CLUSTER} in
> [ssh:Fedora 32] ++ echo '<<< Need parameter set to minikube or minishift >>>'
> [ssh:Fedora 32] ++ exit 1

Need to look into it, TBH.

rnapoles-rh: As per our chat you are setting the parameter value in the Jenkins script, therefore approving. If it does not work using that approach then you will need to set it up in scripts/minici-presubmit-all-tests.sh within the case structure to reduce complexity.

rnapoles-rh (approved):

rnapoles-rh:
psi tests are optional at this time

Issue #4534, opened 2021-03-19
[Discussion] Update version of the ServiceBinding operator from 0.3.0 to 0.5.0


wtrocki:
SBO as dependency is used to build Binding CR, but current version is 0.5.0. 
It will be good to update it to latest version to support bindAsFiles flag.

This blocks:

https://github.com/openshift/odo/issues/4533

wtrocki:
/assign @wtrocki 

This might be controversial but I will see if can bump package

dharmit:
Updated to 0.7.1 - https://github.com/openshift/odo/blob/main/go.mod#L41

/close

Issue #4533, opened 2021-03-19
Add ability to reflect service binding type (secret vs filesystem)


wtrocki:
Service binding (odo link) offers ability to connect using bindingAsFile (creates filesystem based on secret) or link as environment variables.

It is crucial for usability to reflect that in odo.

See also
https://github.com/openshift/odo/issues/4505

kadel:
I've created a user story that covers this issue. https://github.com/openshift/odo/issues/4693

dharmit:
> I've created a user story that covers this issue. #4693

`--bind-as-files` has been added as a part of above issue. If you meant something else, please feel free to re-open.

/close

Issue #4505, opened 2021-03-11
odo link should support providing name of the SBO object


wtrocki:
ServiceBindingOperator plans to use file-based mounted bindings. 
This will mean that name of the service binding objects should be easy to control.

This way odo can be used the same way as service binding objects - creating paths that are expected by devs.

CC @petemuir @sbose78

kadel:
@wtrocki I'm lost :/ Can you point me to some documentation or an example?

wtrocki:
Odo link will create SBO object with `metadata.name` that cannot be specified by end user.

See: https://github.com/redhat-developer/service-binding-operator/issues/893

kadel:
> Odo link will create SBO object with metadata.name that cannot be specified by end user.
Can we change that? Will also discuss the same with OpenShift console team.

That should not be a problem. We could allow users to optionally set a name. It could look like this `odo link Database/database --name  mybinding`

wtrocki:
@kadel no problem. I'm trying to aggregate all those requirements in the ecosystem first and prioritize them. Already checked the code and see that this will be trivial to add so I can either do it myself after some time or someone from odo core team can pick that up. Thank you.

kadel:
I've created user story to cover this issue https://github.com/openshift/odo/issues/4693

dharmit:
> I've created user story to cover this issue #4693

Done as a part of this issue.

/close

PR #4535, opened 2021-03-19
Support Service Binding Spec in  ODO LINK

wtrocki:
## Motivation

Service Binding spec evolved over the last months offering file based bindings. 
This type of bindings are now first class citizens in Quarkus, Spring or Node.js and should be supported by ODO.

Old Env based bindings are still enabled by default so this change is not introducing any breaking/side effects. 

## What was done

This PR adds the following improvements:

1) Updates Service Binding version to 0.5.0 (paving the way for supporting latest changes and features)
2) Adds 2 new flags to enable new features of SBO and alignment with binding specification

```
Flags:
      --bindAsFiles          If present will mount service binding as files in /bindings folder
      --bindingName			 Name of the service binding object
```

Bind as files is defaulting to false for backwards compatibility, but according to the binding spec this should be default to true - it can be added later once we get official documentation/support in all starter apps.

Complete spec can be found here:
https://github.com/k8s-service-bindings/spec

## Verification

1. Use new flags
```
odo link --bindAsFiles=true --bindingName=db  postgresDB
```
Inspect service binding object that was created

**What type of PR is this?**

/kind feature

General improvements to ODO link command making it work with latest service binding features

https://github.com/openshift/odo/issues/4505
https://github.com/openshift/odo/issues/4533
https://github.com/openshift/odo/issues/4534


wtrocki:
CC @kadel

pkg/odo/cli/component/common_link.go
```diff
@@ -83,6 +85,10 @@ func (o *commonLinkOptions) complete(name string, cmd *cobra.Command, args []str
 		return err
 	}
 
+	if o.bindAsFiles && o.bindingName == "" {
+		return errors.New(`--bindAsFiles option requires --bindingName to be specified`)
```
wtrocki, line 89: I'm not sure if that should be enforced or defaults will work as well

maxandersen:
any reason why bindfiles shouldn't be true by default ?

@kadel @elsony this one would be great to see in !

wtrocki:
@dharmit looking into this now

wtrocki:
```
[ssh:Fedora 32] ++ echo '<<< Need parameter set to minikube or minishift >>>'
[ssh:Fedora 32] ++ exit 1
```

Do I need to setup something

dharmit:
@wtrocki thanks for the PR! There's one problem with the functionality.

When I use `bindAsFiles=true`, `odo push` fails (we need to push after doing `odo link`.) Looking at the events in the pod, I'm seeing:
```sh
$ odo link --bindAsFiles --bindingName=etcd EtcdCluster/example

$ odo push
odo push
 ✓  Waiting for component to start [7ms]

Validation
 ✓  Validating the devfile [29675ns]

Creating Kubernetes resources for component node-todo
 ✗  Waiting for component to start [5m] [WARNING x7: Failed]
 ✗  Failed to start component with name node-todo. Error: Failed to create the component: error while waiting for deployment rollout: timeout while waiting for node-todo deployment roll out\nFor more information to help determine the cause of the error, re-run with '-v'.
See below for a list of failed events that occured more than 5 times during deployment:

 NAME                                         COUNT  REASON  MESSAGE                        

 node-todo-bfdf95cf-6rggw.166e97fc5450b232    7      Failed  Error: secret "etcd" not found 

```
The problem is `Error: secret "etcd" not found`. odo is trying to add a `envFrom` to the pod spec. And it's using the secret `etcd` because when you do `odo link`, it stores link info in `.odo/env/env.yaml`:
```sh
$ cat .odo/env/env.yaml                             
ComponentSettings:
  Name: node-todo
  Project: default
  AppName: app
  Link:
  - Name: etcd    <-----
    ServiceKind: EtcdCluster
    ServiceName: example
  RunMode: run
```
odo tries to find a secret by the same name `etcd` as we used in `odo link`. But it can't find it since SBO created a secret with a different name (Note: earlier SBO used to create a Secret with same name as we passed in the spec for the ServiceBinding):
```sh
$ kubectl get secrets 
NAME                                            TYPE                                  DATA   AGE
etcd-da28a7aa8fd5664bb3177e64c9e4f692b1bd6e1f   Opaque                                1      6m
```
Should SBO create a Secret when using `--bindAsFiles`? Maybe we need to change the spec that we're creating while doing `odo link`? (I'm not sure since I don't know how this feature works.)

Also, this is a problem only when using `bindAsFiles`. If we only provide `bindingName=etcd` or don't provide any arguments proposed in this PR, `odo push` works just fine. So I think it's the files thing that's not functioning as expected.

Also, other than this functionality issue, we will need a few more things to take this PR ahead:
1. Integration tests
2. Validation of the name provided to `--bindingName`.
3. Update `odo link` example in docs under [`docs/public/operator-hub`](https://github.com/openshift/odo/blob/master/docs/public/operator-hub.adoc)

dharmit:
> ```
> [ssh:Fedora 32] ++ echo '<<< Need parameter set to minikube or minishift >>>'
> [ssh:Fedora 32] ++ exit 1
> ```
> 
> Do I need to setup something

@wtrocki Please ignore all the failures for tests with PSI name in it. It's an issue on CI which we're trying to fix.

dharmit:
> Integration tests

Test for `odo link` live under `tests/integration/operatorhub/cmd_service_test.go`.

wtrocki:
I will look into push functionality.
Do you think I can abstract behaviour little bit so link and push will be more aligned?

> Should SBO create a Secret when using --bindAsFiles?

It creates volume in the pod so users would be able to see all values and files.

My take is that when bindAsFiles is provided ODO push needs different handling to detect different way of binding. I will also add integration tests and docs

dharmit:
>  Do you think I can abstract behaviour little bit so link and push will be more aligned?

If user experience and/or overall flow is improving, changing the behaviour should be OK. WDYT @kadel? 
But, what does "more aligned" mean here? 

> > Should SBO create a Secret when using --bindAsFiles?
> 
> It creates volume in the pod so users would be able to see all values and files. Volume is created from secret so this should not differ.

Ah, OK. So a Secret is anytime mandatory. Thanks for explaining ! 👍🏾 

> My take is that when bindAsFiles is provided ODO push needs different handling to detect different way of binding.

Yes. I won't be surprised if you hit some bug or usability issue that's been around since some time but something that didn't get noticed.

wtrocki:
> But, what does "more aligned" mean here?

Ignore me here. Checked push implementation.
Adding some small changes will make this work

wtrocki:
I have been thinking how to refactor this. 
There have been many places where binding is used with assumptions that names are generated and env vars are checked. This needs to be extracted and values from bindings (exact name in CR) instead of computed names should be used. Started doing this but there is obviously a lot of work across different commands for this.

wtrocki:
## Progresss
- [ ] Odo push support
- [ ] Odo unlink support

:)

wtrocki:
I put some thoughts into this. Maybe a form of simplification, we can remove name of the service binding as an option here (it will be actually less complex for end user) while still allowing users to bind things as files. 

When custom name is removed both odo push and unlink will work without any changes, but actual binding will be created with file based links - which is what we looking.

wtrocki:
I'm little bit confused because there are also other service bindings initializers that use kubernetes package instead of sbo: https://github.com/openshift/odo/blob/master/pkg/kclient/serviceCatalog.go#L193-L204

What is the story with this?

wtrocki:
Created https://github.com/openshift/odo/pull/4549 to simplify merge as this kinda becomes little bit priority for me - I would love to have quick chat about the whole approach to custom binding names so we can eventually get both of those options done

kadel:
@wtrocki This PR addresses the Secret name issue https://github.com/openshift/odo/pull/4554

mohammedzee1000:
Please rebase and update your PR to be against new default branch, `main`

kadel:
We can close this PR
I will open a PR to address this soon.
/close

PR #4538, opened 2021-03-22
[Do-Not-Merge]Checking jenkins setup env var value

prietyc123:
**What type of PR is this?**

/kind failing-test

Just a debugging pr

### DIFF
--- a/scripts/minici-presubmit-all-tests.sh
+++ b/scripts/minici-presubmit-all-tests.sh
@@ -20,6 +20,7 @@ case $1 in
         export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh"
         export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh"
         export TIMEOUT="2h15m"
+        export CI_INFO="minikube"
         ;;
     minishift)
         export JOB_NAME="odo-minishift-pr-tests"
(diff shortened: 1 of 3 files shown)


scripts/minici-presubmit-all-tests.sh
```diff
@@ -28,6 +29,7 @@ case $1 in
         export SETUP_SCRIPT="scripts/minikube-minishift-setup-env.sh"
         export RUN_SCRIPT="scripts/minikube-minishift-all-tests.sh"
         export TIMEOUT="4h00m"
+        export CI_INFO="minikube"
```
rnapoles-rh, line 32: @prietyc123 this should be CI_INFO="minishift"

rnapoles-rh (changes requested):

scripts/minici-presubmit-all-tests.sh
prietyc123: Defining it locally is also not helping much 🙁 . Looking more into it.

rnapoles-rh: @prietyc123  try using $CI_INFO instead of ${CI_INFO} in minikube-minishift-all-tests.sh and minikube-minishift-setup-env.sh

prietyc123:
As workarounds are behaving weirdly. We come up with unanimous decision on separating out minikube script. opened pr https://github.com/openshift/odo/pull/4539. So closing debugging one.

PR #4549, opened 2021-03-24
General improvements and updates to odo link

wtrocki:
This is alternative version of https://github.com/openshift/odo/pull/4535 that doesn't support custom names of the service bindings


kadel:
I don't think that this will work without a larger change.
Odo made some assumptions about the SBO Secret name, and it looks like last SBO version is generating Secret name with hash.
I'm currently working on fixing this so odo no longer depends on the Secret name.

mohammedzee1000:
Please rebase and update your pr to be against new default branch, `main`

wtrocki:
@mohammedzee1000 sbo 0.7.0 was released that has different capabilities. Feel free to close/reuse those if needed.

kadel:
We can close this PR
I will open a PR to address this soon.

Issue #4553, opened 2021-03-25
Implement server side apply


kadel:
since k8s 1.16 there is a server-side apply https://kubernetes.io/docs/reference/using-api/server-side-apply/

Odo currently needs complex logic to handle updates of existing resources when running `odo push`.
Taking advantage of server-side apply we could reduce a lot of logic that odo currently has that is quite error-prone. 

This problem is especially noticeable in the way how odo handles updates to existing k8s resources in  [createOrUpdateComponent](https://github.com/openshift/odo/blob/12cd795ea43b1d9cc6e15ba240ce81d65929c067/pkg/devfile/adapters/kubernetes/component/adapter.go#L345)
The current approach is to regenerate all resources based on the local information from devfile.yml and env.yaml and just apply it.
This has one big downside. If the operator modifies the existing Deployment the next time odo executes `odo push` the changes get overridden. This is a problem for Service Binding Operator as it adds Secret and other information to Deployment upon ServiceBinding creation. Currently, odo replicates Service Binding Operator logic for injecting envFrom to Deployments. This is error-prone, and every time SBO changes this logic odo will break. 

The current approach is a BIG problem for implementing `bindAsFiles` (https://github.com/openshift/odo/issues/4533). Without server-side apply odo will have to again replicate all modifications that SBO does to Deployments. For `bindAsFiles` this gets more complicated as not just one ~file~ field like in the case of `envFrom`, but now it is multiple of them, (new env variable, `volumeMount`, `volume`)

Without using server-side apply odo won't be able to benefit from new SBO features.

kadel:
Experimental implementation at https://github.com/kadel/odo/tree/ssa

girishramnani:
are you planning to work on this or just on the POC @kadel ?

kadel:
> are you planning to work on this or just on the POC @kadel ?

I'll work on this. POC that I did is already like  80% of the work needed

PR #4554, opened 2021-03-25
Get SB secret name from SB

kadel:
**What type of PR is this?**

/kind bug

Currently, odo assumes that  Secret created by SBO has the same name as ServiceBinding.
This is no longer valid. SBO Secrets have some kind of has as a suffix. 
This PR makes sure that odo detect the correct Secret name from SB definition. 

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -15,6 +15,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
@@ -391,8 +392,25 @@ func (a Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSpe
 		return err
 	}
 
+	// Get ServiceBinding Secret name for each Link
+	var sbSecrets []string
+	for _, link := range ei.GetLink() {
+		sb, err := a.Client.GetDynamicResource(kclient.ServiceBindingGroup, kclient.ServiceBindingVersion, kclient.ServiceBindingResource, link.Name)
+		if err != nil {
+			return err
+		}
+		secret, found, err := unstructured.NestedString(sb.Object, "status", "secret")
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("unable to find secret in ServiceBinding %s", link.Name)
+		}
+		sbSecrets = append(sbSecrets, secret)
+	}
+
 	// set EnvFrom to the container that's supposed to have link to the Operator backed service
-	containers, err = utils.UpdateContainerWithEnvFrom(containers, a.Devfile, a.devfileRunCmd, ei)
+	containers, err = utils.UpdateContainerWithEnvFromSecrets(containers, a.Devfile, a.devfileRunCmd, ei, sbSecrets)
 	if err != nil {
 		return err
 	}
--- a/pkg/odo/cli/component/common_link.go
+++ b/pkg/odo/cli/component/common_link.go
@@ -9,6 +9,7 @@ import (
 	"github.com/openshift/odo/pkg/component"
 	componentlabels "github.com/openshift/odo/pkg/component/labels"
 	"github.com/openshift/odo/pkg/envinfo"
+	"github.com/openshift/odo/pkg/kclient"
 	"github.com/openshift/odo/pkg/log"
 	"github.com/openshift/odo/pkg/occlient"
 	"github.com/openshift/odo/pkg/odo/genericclioptions"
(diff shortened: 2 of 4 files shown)


pkg/odo/cli/component/common_link.go
```diff
@@ -142,7 +134,12 @@ func (o *commonLinkOptions) complete(name string, cmd *cobra.Command, args []str
 				Group:    deploymentSelfLinkSplit[2], // "apps" in above example output
 				Version:  deploymentSelfLinkSplit[3], // "v1" in above example output
 				Resource: deploymentSelfLinkSplit[6], // "deployments" in above example output
-			},
+			}
```
dharmit, line 137: I'm getting build error due to missing `,` here.

dharmit (changes requested):

pkg/odo/cli/component/common_link.go
```diff
@@ -142,7 +134,12 @@ func (o *commonLinkOptions) complete(name string, cmd *cobra.Command, args []str
...
+		}
+
+		// Populate the application selector field in service binding request
+		o.serviceBinding.Spec.Application = &servicebinding.Application{
+			Ref: ref,
```
dharmit, line 142: If I add `,` as [pointed here](https://github.com/openshift/odo/pull/4554/files#r602136902) I get below build error:
```
pkg/odo/cli/component/common_link.go:142:4: unknown field 'Ref' in struct literal of type "github.com/redhat-developer/service-binding-operator/pkg/apis/operators/v1alpha1".Application
```

dharmit: If I replace `Ref` with `LocalObjectReference` (I used `Ctrl+Space` to see what the IDE suggests), it complains that `ref` is undefined:
```sh
pkg/odo/cli/component/common_link.go:142:26: undefined: ref
```

dharmit:
@kadel wondering if you were able to test this successfully on local setup because on my end, when I checkout this PR, I am not able to build odo binary. 😕

kadel:
> @kadel wondering if you were able to test this successfully on local setup because on my end, when I checkout this PR, I am not able to build odo binary. 😕

I picked this commit from a branch with other changes, I guess `git cherry-pick` had some problems merging it.

dharmit:
> I picked this commit from a branch with other changes, I guess `git cherry-pick` had some problems merging it.

OK. I was thinking maybe you used Notepad since you recently moved to Windows. 🤔

wtrocki:
I'm not part of the team but why we need to detect SBO secret? 
My understanding is that for ODO work should be finished once CR is created and `status.conditions.ready==="True"`

kadel:
> I'm not part of the team but why we need to detect SBO secret?
> My understanding is that for ODO work should be finished once CR is created and `status.conditions.ready==="True"`

The reason is this https://github.com/openshift/odo/issues/4553

Currently, odo can't handle if someone else modifies Deployment :/ once you run `odo push` it will override all changes that were not made by odo.
I bad design I know :( , but at the time when this was written there was no other way.

wtrocki:
@kadel would it be possible to jump on call some day for 20 minutes max?

kadel:
> I think supporting file based bindings might have been little premature step that was triggered by some enthusiast who did not know ODO internals :). In SBO 0.7.0 we will get option to run it as SDK - which will mean that ODO could apply bindings locally and avoid problem with overriding resources. When doing work to support from file binding I actually learned a lot about ODO internals and current state. Changing push to patch mode is risky business IMHO and should not be done only for supporting sbo

I did some experiments on Friday and it looks like moving to the patch might not be that risky, but need to do more testing and run a full test suite against the changes, so I'm not going to open PR yet.

My experimental branch with patch and vendored  SBO 0.6.0 and OpenShift libs updated to 4.7 is here https://github.com/kadel/odo/tree/ssa

wtrocki:
@kadel Amazing work!

kadel:
This PR is ready. This is not an ideal solution, but It fixes one of the immediate problems that we have right now.

I'm working on proper fix in https://github.com/kadel/odo/tree/ssa but this is a larger chance that will require more tests and time.

dharmit:
> This PR is ready. This is not an ideal solution, but It fixes one of the immediate problems that we have right now.

`odo link` functionality looks OK. Checking the code.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -391,8 +392,25 @@ func (a Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSpe
...
+		sb, err := a.Client.GetDynamicResource(kclient.ServiceBindingGroup, kclient.ServiceBindingVersion, kclient.ServiceBindingResource, link.Name)
+		if err != nil {
+			return err
+		}
+		secret, found, err := unstructured.NestedString(sb.Object, "status", "secret")
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("unable to find secret in ServiceBinding %s", link.Name)
```
dharmit, line 407:
```suggestion
			return fmt.Errorf("unable to find secret in ServiceBinding %q", link.Name)
```
In an attempt to make odo errors more readable on the CLI.

dharmit (approved):
One nitpick. LGTM otherwise.

/lgtm

PR #4565, opened 2021-03-30
Make Telemetry upload faster

### LINKED ISSUE #4462: Track telemetry implementation for odo

valaparthvi:
**What type of PR is this?**
/kind feature

This PR is an attempt at making Telemetry upload faster. For more reference on this implementation, see - https://github.com/openshift/odo/issues/4462#issuecomment-810152058

Fixes part of #4462 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Set `ConsentTelemetry` preference to true and run a command with a timer.


kadel:
Please check my comment at the original issue https://github.com/openshift/odo/issues/4462#issuecomment-810875108

valaparthvi:
@kadel What if we let the other actions be as they are, but send a `cmd.error` action when the command fails?

valaparthvi:
@kadel On second thought, sending 2 different events won't have much of an impact.
In most cases, we will only send one message to Segment, and if `BatchSize` is 1, it will take the same time as it is taking now. I am going to revert the changes now. I think the one option we have right now is doing something with the process.

pkg/odo/cli/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+type TelemetryOptions struct {
+	telemetryData segment.TelemetryData
+}
+
+func NewTelemetryOptions() *TelemetryOptions {
+	return &TelemetryOptions{}
+}
+
+func (o *TelemetryOptions) Complete(name string, cmd *cobra.Command, args []string) (err error) {
+	err = json.Unmarshal([]byte(args[0]), &o.telemetryData)
```
mik-dass, line 28: This will panic if the caller doesn't provide any args. We can add `Args:cobra.ExactArgs(1),` to telemetryCmd for preventing this panic.

pkg/segment/segment.go
```diff
@@ -172,3 +195,14 @@ func errorType(err error) string {
 func RunningInTerminal() bool {
 	return term.IsTerminal(int(os.Stdin.Fd()))
 }
+
+// IsTelemetryEnabled returns true if user has consented to telemetry
+func IsTelemetryEnabled(cfg *preference.PreferenceInfo) bool {
```
mik-dass, line 200: Please write a unit test for this function

pkg/segment/segment.go
```diff
@@ -172,3 +195,14 @@ func errorType(err error) string {
...
+	// The env variable gets precedence in this decision
+	if os.Getenv(DisableTelemetryEnv) == "true" {
```
mik-dass, line 202: Might be a good idea to convert the env value to lowercase first

valaparthvi: This variable is used by 2 other files. Also, why do you suggest that?

mik-dass: So that `export ODO_DISABLE_TELEMETRY=TRUE` works too. Do we want to support that?

pkg/odo/cli/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+				klog.V(4).Infof("Cannot create a segment client, will not send any data: %s", err.Error())
+			}
+			defer segmentClient.Close()
+		}
+	}
+	if segmentClient != nil {
+		if serr := segmentClient.Upload(o.telemetryData); serr != nil {
+			klog.V(4).Infof("Cannot send data to telemetry: %q", serr)
+		}
+		return segmentClient.Close()
```
mik-dass, line 57: We can put this in the `else` block above and avoid this check when GetConsentTelemetry() is false. We can also remove the `segmentClient` variable declaration above after this change.

pkg/odo/cli/telemetry/telemetry.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (o *TelemetryOptions) Run() (err error) {
+	var segmentClient *segment.Client
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Errorf("unable to update data, required preference.yaml file not found.")
+	}
+	// Initiate the segment client if ConsentTelemetry preference is set to true
+	if cfg.GetConsentTelemetry() {
+		if os.Getenv(segment.DisableTelemetryEnv) == "true" {
```
mik-dass, line 44: I think we can use `IsTelemetryEnabled` here. We can move the warning message below inside `IsTelemetryEnabled` and also convert it into a `klog` message instead of a warning type output.

pkg/segment/segment.go
dharmit: > So that `export ODO_DISABLE_TELEMETRY=TRUE` works too. Do we want to support that?

It would already be covered, no? If I execute an odo command with `export ODO_DISABLE_TELEMETRY=TRUE`, the `os.Getenv` call here will grab it, no? Just to be clear, you didn't miss this declaration earlier in the file, right?
```go
 const DisableTelemetryEnv = "ODO_DISABLE_TELEMETRY"
```

dharmit: > So that `export ODO_DISABLE_TELEMETRY=TRUE` works too. Do we want to support that?

Or is this about `true`, `True`, `TRUE`, etc.?

mik-dass: > Or is this about true, True, TRUE, etc.?

Yes, I meant the case of `TRUE`, `True` etc

pkg/segment/segment_test.go
```diff
@@ -241,6 +252,51 @@ func TestSetError(t *testing.T) {
...
+			errMesssage:     "Telemetry must be disabled.",
+			envVar:          "true",
+			preferenceValue: true,
+		},
+		{
+			want:            true,
+			errMesssage:     "Telemetry must be enabled.",
+			envVar:          "false",
+			preferenceValue: true,
+		},
```
mik-dass, line 283: One other case would be with a invalid value for env var such that `strconv.ParseBool()` throws an error.

pkg/odo/cli/telemetry/telemetry.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (o *TelemetryOptions) Validate() (err error) {
+	return err
+}
+
+func (o *TelemetryOptions) Run() (err error) {
+	cfg, err := preference.New()
+	if err != nil {
+		return errors.Wrapf(err, "Unable to upload telemetry data.")
```
mik-dass, line 37: The error message should start with a small letter

pkg/segment/segment_test.go
valaparthvi: I was planning on adding that, but then I figured we're anyway ignoring the error from `strconv.ParseBool()`, then why check for it. If there is an error, it will consider it False. If you still think it will be good to have, let me know.

mik-dass: If someone modifies the code to
```
disableTelemetry, err:= strconv.ParseBool(os.Getenv(DisableTelemetryEnv))
	if err!=nil {
		return true
	}
```
In case of invalid value like `some-value`, `ParseBool` would throw an error and the function would return true but the current test cases won't detect it.

pkg/segment/segment_test.go
```diff
@@ -281,6 +281,12 @@ func TestIsTelemetryEnabled(t *testing.T) {
 			envVar:          "false",
 			preferenceValue: true,
 		},
+		{
+			want:            true,
+			errMesssage:     "Telemetry must be enabled.",
+			envVar:          "foobar",
+			preferenceValue: true,
```
mik-dass, line 288: Might be a good idea to set the `preferenceValue` value to false so as to be sure that an invalid value of `envVar` returns false.

mik-dass (approved):
/lgtm

PR #4572, opened 2021-03-31
Disable test sending out telemetry

kadel:
Commenting out test that is accidentally sending out telemetry data to production stats.

### DIFF
--- a/tests/integration/cmd_pref_config_test.go
+++ b/tests/integration/cmd_pref_config_test.go
@@ -87,7 +87,8 @@ var _ = Describe("odo preference and config command tests", func() {
 			{"NamePrefix", "foo", "bar", ""},
 			{"BuildTimeout", "5", "7", "foo"},
 			{"Experimental", "false", "true", "foo"},
-			{"ConsentTelemetry", "false", "true", "foo"},
+			// !! Do not test ConsentTelemetry with true because it sends out the telemetry data and messes up the statistics !!
+			{"ConsentTelemetry", "false", "false", "foo"},
 			{"PushTimeout", "4", "6", "f00"},
 		}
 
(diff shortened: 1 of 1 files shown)


tests/integration/cmd_pref_config_test.go
```diff
@@ -366,11 +366,15 @@ var _ = Describe("odo preference and config command tests", func() {
...
-		})
+
+		// This test is disable because it sends out the telemetry data and messes up the statistics
+
+		// It("ConsentTelemetry is set to true", func() {
+		// 	helper.CmdShouldPass("odo", "preference", "set", "ConsentTelemetry", "true", "-f")
+		// 	output := helper.CmdShouldPass("odo", "create", "nodejs", "--context", commonVar.Context)
+		// 	Expect(output).ToNot(ContainSubstring(promtMessageSubString))
+		// })
+
```
valaparthvi, line 377: Can't believe I missed something like this :facepalm: 
Instead of commenting this test, I suggest we remove it and rename the other `It` to 'ConsentTelemetry is set", it makes more sense IMO.

valaparthvi (changes requested):

tests/integration/cmd_pref_config_test.go
kadel: I think that here it might make sense to keep it commented out. It might prevent from someone writing similar test again :)

valaparthvi: Comment - 
Do not test with `true` because it sends out the telemetry data and messes up the statistics.

kadel: updated

valaparthvi (approved):

PR #4576, opened 2021-04-01
Use UserId instead of AnonymousId while sending telemetry data

### LINKED ISSUE #4571: Telemetry should use userId instead of anonymousId
/kind bug

I haven't run anything

## Actual behavior
Telemetry events sent to segment currently use the `anonymousId` event property. 

## Expected behavior
Telemetry events should actually use `userId` (see https://github.com/redhat-developer/vscode-redhat-telemetry/blob/main/src/services/reporter.ts#L28) instead, so they can be properly tracked in Woopra.
This is the exact same data, but sent using a different property

valaparthvi:
**What type of PR is this?**
/kind code-refactoring

This PR changes AnonymousId to UserID so that odo data can be synchronized with data from other projects.

Fixes #4571 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Send some data to Segment and check if `UserId` appears in the response.

### DIFF
--- a/pkg/segment/segment.go
+++ b/pkg/segment/segment.go
@@ -79,8 +79,8 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error 
 
 	// queue the data that helps identify the user on segment
 	if err1 := c.SegmentClient.Enqueue(analytics.Identify{
-		AnonymousId: anonymousID,
-		Traits:      addConfigTraits(),
+		UserId: anonymousID,
+		Traits: addConfigTraits(),
 	}); err1 != nil {
 		return err1
 	}
@@ -99,9 +99,9 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error 
 
 	// queue the data that has telemetry information
 	return c.SegmentClient.Enqueue(analytics.Track{
-		AnonymousId: anonymousID,
-		Event:       action,
-		Properties:  properties,
+		UserId:     anonymousID,
+		Event:      action,
+		Properties: properties,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


pkg/segment/segment.go
```diff
@@ -79,8 +79,8 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 
 	// queue the data that helps identify the user on segment
 	if err1 := c.SegmentClient.Enqueue(analytics.Identify{
-		AnonymousId: anonymousID,
-		Traits:      addConfigTraits(),
+		UserId: anonymousID,
```
dharmit, line 82:
```suggestion
		UserId: userID,
```

pkg/segment/segment.go
```diff
@@ -99,9 +99,9 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 
 	// queue the data that has telemetry information
 	return c.SegmentClient.Enqueue(analytics.Track{
-		AnonymousId: anonymousID,
-		Event:       action,
-		Properties:  properties,
+		UserId:     anonymousID,
```
dharmit, line 102:
```suggestion
		UserId:     userID,
```

dharmit (changes requested):
Might seem like a nitpick right now but, renaming it would help avoid confusion when someone looks at the code later.

valaparthvi:
>Might seem like a nitpick right now but, renaming it would help avoid confusion when someone looks at the code later.

I can't say I disagree, but `anonymousId` as a variable name makes more sense in a way that it tells that the UserId has to be anonymous. I have half a mind of renaming it to `userId` though :sweat_smile:

kadel:
> I can't say I disagree, but `anonymousId` as a variable name makes more sense in a way that it tells that the UserId has to be anonymous. I have half a mind of renaming it to `userId` though 😅

but the function that retrieves it is called `getUserIdentity` ;

dharmit (approved):

PR #4580, opened 2021-04-01
introduce ODO_DISABLE_TELEMETRY and skip telemetry question when there is no tty

### LINKED ISSUE #4462: Track telemetry implementation for odo

kadel:
**What type of PR is this?**

/kind cleanup
/kind code-refactoring

- introduces `ODO_DISABLE_TELEMETRY ` env variable. When this is set to `true` odo will skip the telemetry question and it won't send any telemetry data out even if `odo preference set consentTelemetry true`.
- sets `ODO_DISABLE_TELEMETRY` in all ginkgo test suites to make sure that no data is sent to telemetry even by accident
- if odo is not executed in terminal (no tty) it won't ask user to consent with telemetry

Partly Fixes #4462 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/USAGE_DATA.adoc
+++ b/USAGE_DATA.adoc
@@ -29,3 +29,4 @@ Disable::
 Note: If earlier the `ConsentTelemetry` preference was enabled, then the data will be collected about the disabling of the preference.
 
 Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `true`.
+This environment can override `ConsentTelemetry` value set by `odo preference`. 
--- a/USAGE_DATA.adoc
+++ b/USAGE_DATA.adoc
@@ -28,4 +28,4 @@ Disable::
 
 Note: If earlier the `ConsentTelemetry` preference was enabled, then the data will be collected about the disabling of the preference.
 
-Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `false`.
+Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `true`.
--- a/USAGE_DATA.adoc
+++ b/USAGE_DATA.adoc
@@ -27,3 +27,5 @@ Disable::
 `odo preference set ConsentTelemetry false`
 
 Note: If earlier the `ConsentTelemetry` preference was enabled, then the data will be collected about the disabling of the preference.
+
+Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `false`.
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -44,7 +44,9 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 	// This prompt has been placed here so that it does not prompt the user when they call --help
 
 	if !cfg.IsSet(preference.ConsentTelemetrySetting) && cmd.Parent().Name() != "preference" {
-		if os.Getenv(disableTelemetryEnv) == "true" {
+		if !segment.RunningInTerminal() {
+			klog.V(4).Infof("Skipping telemetry question because there is no terminal (tty)\n")
+		} else if os.Getenv(disableTelemetryEnv) == "true" {
 			klog.V(4).Infof("Skipping telemetry question due to %s=%s\n", disableTelemetryEnv, os.Getenv(disableTelemetryEnv))
 		} else {
 			var consentTelemetry bool
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -24,6 +24,10 @@ var (
 	segmentClient *segment.Client
 )
 
+// disableTelemetryEnv is name of environment variable, if set to true it disables odo telemetry completely
+// hiding even the question
+const disableTelemetryEnv = "ODO_DISABLE_TELEMETRY"
+
 type Runnable interface {
 	Complete(name string, cmd *cobra.Command, args []string) error
 	Validate() error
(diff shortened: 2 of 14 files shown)


dharmit:
Should we remove below code added in #4482 as a part of this PR? https://github.com/openshift/odo/blob/81ddd194cc4326b053cff7ced5b8be7fd2a4a628/tests/helper/helper_generic.go#L299-L302

USAGE_DATA.adoc
```diff
@@ -27,3 +27,5 @@ Disable::
 `odo preference set ConsentTelemetry false`
 
 Note: If earlier the `ConsentTelemetry` preference was enabled, then the data will be collected about the disabling of the preference.
+
+Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `false`.
```
dharmit, line 31:
```suggestion
Alternatively you can disable telemetry by setting `ODO_DISABLE_TELEMETRY` environment variable to `true`.
```
Simply based on what you've mentioned in PR description:
> introduces ODO_DISABLE_TELEMETRY env variable. When this is set to true odo will skip the telemetry question and it won't send any telemetry data out even if odo preference set consentTelemetry true.

I haven't looked at the code yet.

kadel: oh, good catch.

kadel:
> Should we remove below code added in #4482 as a part of this PR?

I guess I can do that. But it si not required.

dharmit:
I hope this is not some git rebase mess - https://github.com/openshift/odo/pull/4580#commits-pushed-59133a1.

tests/integration/cmd_pref_config_test.go
```diff
@@ -359,10 +359,6 @@ var _ = Describe("odo preference and config command tests", func() {
 			output = helper.CmdShouldPass("odo", "preference", "unset", "buildtimeout", "-f")
 			Expect(output).ToNot(ContainSubstring(promtMessageSubString))
 		})
-		It("prompt should appear when non-preference command is run", func() {
-			output := helper.CmdShouldPass("odo", "create", "nodejs", "--context", commonVar.Context)
-			Expect(output).To(ContainSubstring(promtMessageSubString))
-		})
```
valaparthvi, line 365: Can we unset(in setup) and set(in teardown) the ODO_DISABLE_TELEMETRY env to test this scenario instead of removing it?

tests/template/template_suite_test.go
```diff
@@ -3,11 +3,10 @@ package template
...
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"github.com/openshift/odo/tests/helper"
 )
 
 func TestTemplate(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "TestTemplate Suite")
+	helper.RunTestSpecs(t, "TestTemplate Suite")
```
valaparthvi, line 10: This one doesn't require `RunSpecsWithDefaultAndCustomReporters`, yet calling `RunTestSpecs` will call it, is that alright?

tests/integration/operatorhub/operatorhub_suite_test.go
```diff
@@ -3,11 +3,9 @@ package integration_test
...
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"github.com/openshift/odo/tests/helper"
 )
 
 func TestOperatorhub(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Operatorhub Suite")
+	helper.RunTestSpecs(t, "Operatorhub Suite")
```
valaparthvi, line 10: This one requires `RunSpecs`, but `RunTestSpecs` calls `RunSpecsWithDefaultAndCustomReporters`, will that be alright?

valaparthvi (changes requested):
Apart from the weird magic that happened with the commits, and a few questions, everything else looks good to me.

tests/template/template_suite_test.go
kadel: Yes, it was a mistake. Someone forgot to update the template when we started using Reproters

tests/integration/operatorhub/operatorhub_suite_test.go
kadel: yes, I think this was also a mistake where someone forgot to update it to use Reporters.
This is exactly the reason why we should not have duplicated code like this, and instead have one common function that handles common setup.

dharmit:
@kadel PTAL if this is not some git rebase mess.

image

tests/integration/cmd_pref_config_test.go
kadel: Originally I deleted this to make sure that we don't send something out even by accident.
Now when we also have https://github.com/openshift/odo/pull/4585 I can put it back and try unsettling ODO_DISABLE_TELEMETRY

kadel: There is also problem that now the message is now shown if odo is not executed in terminal

valaparthvi: In that case, I'd rather we comment out the test with a reason than remove it.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -38,25 +42,35 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
...
-		var consentTelemetry bool
-		prompt := &survey.Confirm{Message: "Help odo improve by allowing it to collect usage data. Read about our privacy statement: https://developers.redhat.com/article/tool-data-collection. You can change your preference later by changing the ConsentTelemetry preference.", Default: false}
-		err = survey.AskOne(prompt, &consentTelemetry, nil)
-		ui.HandleError(err)
-		if err == nil {
-			if err1 := cfg.SetConfiguration(preference.ConsentTelemetrySetting, strconv.FormatBool(consentTelemetry)); err1 != nil {
-				klog.V(4).Info(err1.Error())
+		if !segment.RunningInTerminal() {
+			klog.V(4).Infof("Skipping telemetry question because there is no terminal (tty)\n")
+		} else if os.Getenv(disableTelemetryEnv) == "true" {
```
dharmit, line 49: There are in all four calls to `os.Getenv` in this function. Would it make sense to store the value in a variable and reduce it to just one call? TBH, I don't know how costly a call like `os.Getenv` would be from performance point of view.

pkg/segment/segment.go
```diff
@@ -20,7 +20,9 @@ import (
 )
 
 // Writekey will be the API key used to send data to the correct source on Segment
-var WriteKey = "R1Z79HadJIrphLoeONZy5uqOjusljSwN"
+// odo-prod R1Z79HadJIrphLoeONZy5uqOjusljSwN
+// odo-dev 4xGV1HV7K2FtUWaoAozSBD7SNCBCJ65U
+var writeKey = "R1Z79HadJIrphLoeONZy5uqOjusljSwN"
```
dharmit, line 25: Out of curiosity - this key is OK to be even put out in a public code base like odo's. Why are we changing it to a private variable in that case?

dharmit (changes requested):
More like a few questions than review. But would prefer getting an answer for better understanding. :)

pkg/segment/segment.go
valaparthvi: >this key is OK to be even put out in a public code base like odo's.

Yes, we confirmed this with a few people. We need those keys to be able to send data to Segment, hence they will be in our user's system. I understand you had a concern about us receiving borked data with these keys out in the open, but that's a risk we are willing to take it seems.

dharmit: @valaparthvi the question in my review is - "Why are we changing it to a private variable in that case?" I understand that it's fine to put that key in public.

mohammedzee1000:
@kadel plz rebase

pkg/segment/segment.go
valaparthvi: My bad. Anytime I read something about the key being open, my fingers tend to type those words :facepalm:

kadel: > @valaparthvi the question in my review is - "Why are we changing it to a private variable in that case?" I understand that it's fine to put that key in public.

Good practise, there is no need for this variable to be public, and it should not be used outside of the `segment` package

pkg/odo/genericclioptions/runnable.go
kadel: the call is cheap, but it might be more readable if it is stored in one call.

dharmit (approved):

Issue #4504, opened 2021-03-10
Use official devfile registry


kadel:
The New Official Devfile registry is running at https://registry.devfile.io/  
Devfiles served by the registry are at https://github.com/devfile/registry

Odo should deprecate and stop using https://github.com/odo-devfiles/registry and use over one instead.

The new registry has a different API, so all commands and function operating on the registry will have to be updated

/priority high
/area registry

GeekArthur:
Related issue from devfile service side: https://github.com/devfile/api/issues/367

girishramnani:
is this the OCI registry with devfile as an manifest type?

kadel:
> is this the OCI registry with devfile as an manifest type?

yes, but OCI registry is just an implementation detail. There is an HTTP REST interface. odo and other tools will interface just with that.

GeekArthur:
All the PRs for OCI-based registry migration are merged, close this issue.

PR #4636, opened 2021-04-16
OCI-based devfile reigstry support patch

GeekArthur:
Signed-off-by: jingfu wang <jingfwan@redhat.com>

**What type of PR is this?**
/kind bug

This PR fixes the issue that starter project overrides the devfile stack additional resources in the project root directory. The issue only happens on the case where the devfile stack has additional resources (eg. icon, outer loop configuration). 

Related issues:
1. https://github.com/openshift/odo/issues/4504
2. https://github.com/devfile/api/issues/367

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

1. Run `odo create nodejs --registry DefaultDevfileRegistry --starter` then verify the nodejs devfile and corresponding resources and starter project can be downloaded from OCI-based public devfile registry.

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -866,7 +866,7 @@ func (co *CreateOptions) devfileRun() (err error) {
 	if err != nil {
 		return errors.Wrapf(err, "unable to save devfile to %s", DevfilePath)
 	}
-	if co.devfileMetadata.devfilePath.value == "" && !devfileExist {
+	if co.devfileMetadata.devfilePath.value == "" && !devfileExist && !strings.Contains(co.devfileMetadata.devfileRegistry.URL, "github") {
 		err = registryLibrary.PullStackFromRegistry(co.devfileMetadata.devfileRegistry.URL, co.devfileMetadata.componentType, co.componentContext)
 		if err != nil {
 			return err
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -773,6 +773,7 @@ func (co *CreateOptions) s2iRun() (err error) {
 // Run has the logic to perform the required actions as part of command
 func (co *CreateOptions) devfileRun() (err error) {
 	var devfileData []byte
+	devfileExist := util.CheckPathExists(DevfilePath)
 	// Use existing devfile directly from --devfile flag
 	if co.devfileMetadata.devfilePath.value != "" {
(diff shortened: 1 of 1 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -859,12 +860,18 @@ func (co *CreateOptions) devfileRun() (err error) {
...
 	err = ioutil.WriteFile(DevfilePath, devfileData, 0644) // #nosec G306
 	if err != nil {
 		return errors.Wrapf(err, "unable to save devfile to %s", DevfilePath)
 	}
+	if co.devfileMetadata.devfilePath.value == "" && !devfileExist && !strings.Contains(co.devfileMetadata.devfileRegistry.URL, "github") {
+		err = registryLibrary.PullStackFromRegistry(co.devfileMetadata.devfileRegistry.URL, co.devfileMetadata.componentType, co.componentContext)
+		if err != nil {
+			return err
+		}
+	}
```
valaparthvi, line 874: @GeekArthur Why do you rewrite this piece of code when it is already well defined here at [L820](https://github.com/openshift/odo/pull/4636/files#diff-e9cbb4b32525048f3bf6ffce25bb7e24c02221819d5b9de9347fe2c835828058R820)?
```go
devfileExist := util.CheckPathExists(DevfilePath)
if co.devfileMetadata.devfilePath.value != "" {
...
} else {
    if devfileExist {
    ...
    } else {
        if strings.Contains(co.devfileMetadata.devfileRegistry.URL, "github") {
        ...
        } else {
            err = registryLibrary.PullStackFromRegistry(co.devfileMetadata.devfileRegistry.URL, co.devfileMetadata.componentType, co.componentContext)
            if err != nil {
		return err
	    }
	}
    }
}
```    
The above logic seems to be the same as the one you have written - 
```go
if co.devfileMetadata.devfilePath.value == "" && !devfileExist && !strings.Contains(co.devfileMetadata.devfileRegistry.URL, "github") {
		err = registryLibrary.PullStackFromRegistry(co.devfileMetadata.devfileRegistry.URL, co.devfileMetadata.componentType, co.componentContext)
		if err != nil {
			return err
		}
	}
```

Do we need to rewrite the devfile? This seems to be a problem because it rewrites the devfile that was already written at [L865](https://github.com/openshift/odo/pull/4636/files#diff-e9cbb4b32525048f3bf6ffce25bb7e24c02221819d5b9de9347fe2c835828058R865).

PR #4648, opened 2021-04-21
use server side apply

### LINKED ISSUE #4553: Implement server side apply

kadel:
**What type of PR is this?**

/kind cleanup
/kind code-refactoring

Fixes #4553 

** Notes ***
There is a fallback for  k8s < v1.16.0. On later versions, it falls back to not using server-side apply. This is ok, as the main reason for using SSA is SBO and SBO won't work on those k8s versions anyway.

I have executed `▶ make test-integration-devfile ` against ocp 3.11.
Two tests are failing.
```
Summarizing 2 Failures:

[Fail] odo devfile app command tests Testing URLs for OpenShift specific scenarios [It] it should list, describe and delete the apps 
/home/tomas/Code/Work/odo/tests/helper/helper_run.go:34

[Fail] odo devfile push command tests push with listing the devfile component [It] checks devfile and s2i components together 
/home/tomas/Code/Work/odo/tests/helper/helper_run.go:34

Ran 203 of 208 Specs in 1338.623 seconds
FAIL! -- 201 Passed | 2 Failed | 0 Pending | 5 Skipped
```
But the same two tests are failing on master as well. So this is not something that was introduced in this PR.

https://github.com/openshift/odo/issues/4678

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 


kadel:
```
error: error creating buildah builder: Error reading signatures: Error downloading signatures for sha256:7157e96d3be123ef879a2c7061271a97b75d58b2669c9afc5ca35cebaa1ba720 in registry.build01.ci.openshift.org/ci/managed-clonerefs: received unexpected HTTP status: 504 Gateway Time-out
```
/retest

pkg/kclient/kclient.go
```diff
@@ -228,3 +233,38 @@ func (c *Client) IsResourceSupported(apiGroup, apiVersion, resourceName string)
...
+// IsSSASupported checks if Server Side Apply is supported by cluster
+// SSA was introduced in Kubernetes 1.16
+// If there is an error while parsing versions, it assumes that SSA is supported by cluster.
+// Most of clusters these days are 1.16 and up
+func (c *Client) IsSSASupported() bool {
+	// check if this was done before so we don't query cluster multiple times for the same info
+	if c.isSSASupported == nil {
+		versionWithSSA, err := semver.Make("1.16.0")
+		if err != nil {
+			klog.Warningf("unable to parse version %q", err)
```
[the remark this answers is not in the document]
kadel: Yes, good catch, I will add return.
The reason I added this check was to make sure that if in the future someone updates the `1.16.0` string and makes a typo it will be counght.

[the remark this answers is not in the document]
feloy: In this case, you can perhaps make an assertion and panic, like it is done for `regexp.MustCompile` for example (https://golang.org/pkg/regexp/#MustCompile), so the developer will see the error quickly if it makes a typo

pkg/kclient/deployments.go
```diff
@@ -185,18 +190,40 @@ See below for a list of failed events that occured more than %d times during dep
...
+func (c *Client) ApplyDeployment(deploy appsv1.Deployment) (*appsv1.Deployment, error) {
+	data, err := json.Marshal(deploy)
+
+	klog.V(5).Infoln("Applying Deployment via server-side apply:")
+	klog.V(5).Infoln(resourceAsJson(deploy))
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to marshal deployment")
+	}
+	deployment, err := c.KubeClient.AppsV1().Deployments(c.Namespace).Patch(context.TODO(), deploy.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: "odo", Force: boolPtr(true)})
```
dharmit, line 226: One could just read the docs, but can you add a comment explaining what's going on here in context of odo?

pkg/kclient/deployments.go
```diff
@@ -185,18 +190,40 @@ See below for a list of failed events that occured more than %d times during dep
...
-// CreateDeployment creates a deployment based on the given deployment spec
+func resourceAsJson(resource interface{}) string {
+	data, _ := json.MarshalIndent(resource, " ", " ")
+	return string(data)
+}
+
+// ApplyDeployment updates a deployment based on the given deployment spec
 func (c *Client) CreateDeployment(deploy appsv1.Deployment) (*appsv1.Deployment, error) {
-	deployment, err := c.KubeClient.AppsV1().Deployments(c.Namespace).Create(&deploy)
+	deployment, err := c.KubeClient.AppsV1().Deployments(c.Namespace).Create(context.TODO(), &deploy, metav1.CreateOptions{FieldManager: "odo"})
```
dharmit, line 200: Would it add value to store `metav1.CreateOptions{FieldManager: "odo"}` in a variable? I'm asking because:

```sh
$ git diff main pkg/ | grep "metav1.CreateOptions{FieldManager: \"odo\"}" | wc -l
19
```

pkg/kclient/kclient.go
```diff
@@ -228,3 +233,38 @@ func (c *Client) IsResourceSupported(apiGroup, apiVersion, resourceName string)
 	}
 	return supported, nil
 }
+
+// IsSSASupported checks if Server Side Apply is supported by cluster
+// SSA was introduced in Kubernetes 1.16
+// If there is an error while parsing versions, it assumes that SSA is supported by cluster.
```
dharmit, line 239: I'm confused. If there's an error to parse k8s version, we assume that SSA is supported? Is this for real? Why would parsing the version info fail? And how does that lead to concluding the SSA is supported? I'm lost. 🤔

dharmit (changes requested):

dharmit:
@kadel, First off, I know that I myself don't entertain the "what about users who modify devfile by hand" discussions. So I'm not asking to tackle the scenario here. I'm just putting my thoughts, and we can very well handle it at a later stage.

With the disclaimer in mind, should something change with regard to our approach for `odo service create` here: https://github.com/openshift/odo/blob/a920cccb8513da7fc012e699be3c8dda77e3d3ba/pkg/service/service.go#L856-L867

When I wrote that, I simply made sure to ignore the error. But if a user modified the service info (Kubernetes inlined thing) in devfile by hand, `odo push` won't be able to make changes on the cluster. My understanding/assumption/expectation was the with SSA, it would be possible for odo to handle such changes when doing subsequent `odo push`.

For example, I'm spinning up an EtcdCluster using below manifest:
```yaml
- kubernetes:
    inlined: |
      apiVersion: etcd.database.coreos.com/v1beta2
      kind: EtcdCluster
      metadata:
        annotations:
          etcd.database.coreos.com/scope: clusterwide
        name: example
      spec:
        size: 3
        version: 3.2.13
  name: example
```

I then modified size from `size: 3` to `size: 1`. Should `odo push` (with support for SSA) be able to scale down the pods for EtcdCluster?

pkg/auth/login.go
```diff
@@ -135,7 +135,7 @@ func copyAndFilter(w io.Writer, r io.Reader) ([]byte, error) {
 func filteredInformation(s []byte) []byte {
 
 	// List of strings to correctly filter
-	s = bytes.Replace(s, []byte("new-project"), []byte("project create"), -1)
+	s = bytes.Replace(s, []byte("oc new-project"), []byte("odo project create"), -1)
```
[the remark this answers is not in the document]
kadel: `oc` changed the message a little bit. Without this change, the `odo login` outputs `oc project create` instead of `odo project create` 
```
▶ odo login -u developer -p developer
Connecting to the OpenShift cluster

Login successful.

You don't have any projects. You can try to create a new project, by running

    oc project create <project-name>
```

pkg/kclient/deployments.go
kadel: Do you have any specific part in mind?
This is just calling ServerSide to apply. it is equivalent of `kubectl apply --server-side --force-conflicts`.
I'm using force to make sure that if someone changed one of the values that odo manages, odo overrides it with the value it expects instead of failing due to conflict.

pkg/kclient/deployments.go
kadel: don't think it makes sense to put the whole `metav1.CreateOptions` into a variable. It would make the code less readable, as you would have to jump to different code sections to see create options. 

I could put `odo` string for `FieldManager` into a separate value.

But for some reason, the way it is written right now felt right to me.

pkg/kclient/kclient.go
kadel: Failing to parse an error would be an unexpected error, that should never happen.  
But that also means that it is bound to happen at least once future :-D 

Most of the k8s versions that are currently being used have support for SSA. 
So even if there is some weird version string that we can't parse odo can try to continue doing its work.
If the cluster with a weird version number supports SSA odo will be able to continue, otherwise, it would fail on a relatively meaningless version check. 
If the cluster doesn't support SSA it will fail while trying to use it.

kadel:
> I then modified size from `size: 3` to `size: 1`. Should `odo push` (with support for SSA) be able to scale down the pods for EtcdCluster?

yes

pkg/kclient/deployments.go
kadel: `Force` requires a pointer to `boolean` the `boolPtr` functions return point to the boolean with value defined as an argument.

pkg/kclient/kclient.go
dharmit: > Failing to parse an error would be an unexpected error, that should never happen.
> But that also means that it is bound to happen at least once future :-D

Is the k8s project going to change the versioning approach in future?

>  So even if there is some weird version string that we can't parse odo can try to continue doing its work.

Did you come across some weird version number that you could share to explain this?

My concern is that if it's an assumption then we're better off not bothering about it right now and addressing/fixing things when we hit them.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -409,29 +409,6 @@ func (a Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSpe
 		return err
 	}
 
-	// Get ServiceBinding Secret name for each Link
```
feloy, line 412: If the cluster does not support SSA, the link information in the container will be erased, right?

kadel: Yes, and No :-) If the cluster doesn't support SSA there will be no `ServiceBinding` becasue Service Binding Operator can't work without it.

pkg/kclient/kclient.go
kadel: > Is the k8s project going to change the versioning approach in future?

Don't know

> Did you come across some weird version number that you could share to explain this?

I did not.





> My concern is that if it's an assumption then we're better off not bothering about it right now and addressing/fixing things when we hit them.

Don't understand what you mean.

dharmit: Probably there's some misunderstanding, but I don't understand why we're expecting to eventually hit an error while parsing the server version.

dharmit: Resolved this over a chat with @kadel & @feloy. I was focussing more on the code part than on the SSA part.

kadel:
done, managed to a resolved failing test. It was due to conflicts with https://github.com/openshift/odo/pull/4671

dharmit:
Looks good to me.

/hold cancel

PR #4650, opened 2021-04-21
Creates service on cluster upon "odo push"

dharmit:
**What type of PR is this?**
/kind feature

This PR ensures that an Opreator backed service is created on the cluster only when user does `odo push` and not when they do `odo service create`.

Fixes part of the scope mentioned in https://github.com/openshift/odo/issues/4160#issuecomment-809366184

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

Notes:
* This PR is targeted only for `odo service create; odo push` workflow for Operator backed services. Adding labels to the service, deleting service upon doing `odo push` and Service Catalog related changes will come in subsequent PRs after this one is merged. Doing this to break big piece into smaller chunks.
* ~I need to add integration tests to the PR.~ Integration tests have been modified to use `odo push` to actually create service on the cluster. Can't think of a new scenario that needs to be added. Open to ideas.

How to test:
* For Operator backed services, `odo service create` and `odo push`. Subsequent `odo push` would silently pass without any service info (because the service already exists on cluster.)

### DIFF
--- a/tests/integration/operatorhub/cmd_service_test.go
+++ b/tests/integration/operatorhub/cmd_service_test.go
@@ -83,7 +83,7 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
 			operators := helper.CmdShouldPass("odo", "catalog", "list", "services")
 			etcdOperator := regexp.MustCompile(`etcdoperator\.*[a-z][0-9]\.[0-9]\.[0-9]-clusterwide`).FindString(operators)
 			stdOut := helper.CmdShouldPass("odo", "service", "create", fmt.Sprintf("%s/EtcdCluster", etcdOperator), "--project", commonVar.Project)
-			Expect(stdOut).To(ContainSubstring("Service \"example\" was created"))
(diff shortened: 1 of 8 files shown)


tests/integration/operatorhub/cmd_service_test.go
```diff
@@ -459,39 +465,5 @@ spec:
 			stdOut = helper.CmdShouldFail("odo", "unlink", "EtcdCluster/example")
 			Expect(stdOut).To(ContainSubstring("failed to unlink the service"))
 		})
-
-		It("should fail if we delete a link outside of odo (using oc)", func() {
```
dharmit, line 463: Removing this test spec because it's about a scenario where the user does something outside the odo workflow.

valaparthvi:
Codewise, it looks good to me. I haven't been able to test functionality though, I couldn't get `odo catalog list services` to work.

valaparthvi (approved):

PR #4662, opened 2021-04-27
Collect Component type for usage data

### LINKED ISSUE #4462: Track telemetry implementation for odo

valaparthvi:
**What type of PR is this?**
/kind feature
/kind code-refactoring

This PR adds the component type to the telemetry data when the user creates a component.

Fixes part of #4462

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

With telemetry enabled, run `odo create nodejs` and see if the data pops up on Segment.

A note: This PR is based off #4565.

### DIFF
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	segment "github.com/openshift/odo/pkg/segment/context"
+	scontext "github.com/openshift/odo/pkg/segment/context"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -908,7 +908,7 @@ func (co *CreateOptions) devfileRun() (err error) {
 
 // Run has the logic to perform the required actions as part of command
 func (co *CreateOptions) Run(cmd *cobra.Command) (err error) {
-	segment.SetComponentType(cmd.Context(), co.devfileMetadata.componentType)
+	scontext.SetComponentType(cmd.Context(), co.devfileMetadata.componentType)
 	// By default we run Devfile
 	if !co.forceS2i && co.devfileMetadata.devfileSupport {
 		err := co.devfileRun()
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/openshift/odo/pkg/segment"
+	segment "github.com/openshift/odo/pkg/segment/context"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/openshift/odo/pkg/segment"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/zalando/go-keyring"
@@ -905,8 +907,8 @@ func (co *CreateOptions) devfileRun() (err error) {
 }
 
 // Run has the logic to perform the required actions as part of command
-func (co *CreateOptions) Run() (err error) {
-
+func (co *CreateOptions) Run(cmd *cobra.Command) (err error) {
+	segment.SetComponentType(cmd.Context(), co.devfileMetadata.componentType)
 	// By default we run Devfile
 	if !co.forceS2i && co.devfileMetadata.devfileSupport {
 		err := co.devfileRun()
--- a/pkg/segment/context/context.go
+++ b/pkg/segment/context/context.go
@@ -1,0 +1,61 @@
+package context
+
+import (
+	"context"
+	"sync"
+)
+
+type contextKey struct{}
+
+var key = contextKey{}
+
+type properties struct {
+	lock    sync.Mutex
+	storage map[string]interface{}
+}
+
+func (p *properties) set(name string, value interface{}) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	p.storage[name] = value
+}
+
+func (p *properties) values() map[string]interface{} {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	ret := make(map[string]interface{})
+	for k, v := range p.storage {
+		ret[k] = v
+	}
+	return ret
+}
+
+// NewContext returns a context more specifically to be used for telemetry data collection
+func NewContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, key, &properties{storage: make(map[string]interface{})})
+}
+
+func propertiesFromContext(ctx context.Context) *properties {
+	value := ctx.Value(key)
+	if cast, ok := value.(*properties); ok {
+		return cast
+	}
+	return nil
+}
+
+// SetContextProperty sets the value of a key in given context for telemetry data
+func SetContextProperty(ctx context.Context, key string, value interface{}) {
+	properties := propertiesFromContext(ctx)
+	if properties != nil {
+		properties.set(key, value)
+	}
+}
+
+// GetContextProperties retrieves all the values set in a given context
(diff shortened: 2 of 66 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -905,8 +907,8 @@ func (co *CreateOptions) devfileRun() (err error) {
 }
 
 // Run has the logic to perform the required actions as part of command
-func (co *CreateOptions) Run() (err error) {
-
+func (co *CreateOptions) Run(cmd *cobra.Command) (err error) {
+	segment.SetComponentType(cmd.Context(), co.devfileMetadata.componentType)
```
girishramnani, line 911: why change the signature of run to get access to the `context.Context`? I would suggest maybe setting it as part of the `CreateOptions` struct
```
type CreateOptions struct {
cmdContext context.Context
}
```

then setting it inside the `Complete` function which gets the `cmd` as an argument.

valaparthvi: Yup, there is cmd in Complete(), but I don't want to upload information about a component type that might have been misspelled or some other error, I want to wait until it is ready for Run(). But I'll look into this https://github.com/openshift/odo/pull/4662#discussion_r626018142. See if it's possible.

girishramnani: You can validate the name in complete for this case only. We have made exceptions when it’s necessary

girishramnani: Yeah so I would suggest just set the context in Complete,
Then validate the type in Validate but send it in Run. 
as you would have the context already in the struct you don’t need to change the interface

valaparthvi: @girishramnani I tried what you suggested here but it doesn't make sense to me to have one more context added to co, plus,  you suggested we set `co.cmdContext` in `Complete()` and then send it in `Run()` but that is not possible as far as I can tell because the value is sent from `runnable.go::startTelemetry` which calls `segment.GetContextProperties(context.Context)`, we need to pass the same context in which we set a value before, in our case, `co.cmdContext`. I do not think there is a way to pass that value in `startTelemetry`.

Also, having `cmd` in `Run()` will help when we want to pass other values from other runnables.

feloy: > why change the signature of run to get access to the `context.Context`? I would suggest maybe setting it as part of the `CreateOptions` struct
> 
> ```
> type CreateOptions struct {
> cmdContext context.Context
> }
> ```

No, please don't do this, as this is not the recommended way to pass contexts: 

"Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:" from https://golang.org/pkg/context/

pkg/segment/segment.go
```diff
@@ -211,3 +218,59 @@ func IsTelemetryEnabled(cfg *preference.PreferenceInfo) bool {
 	}
 	return false
 }
+
+type contextKey struct{}
```
feloy, line 222: It could be interesting to have this code in a separate package. 

You are using a generic storage (map[string]interface{}), but you are restricting the possibilites to add values to this storage by exporting only `SetComponentType`, and not `set`.

- If you want to have a generic storage, you shouldn't have a `SetComponentType` but only a `Set` (exported),
- If you want to restrict the possibilites of values to store, you could continue to have `SetComponentType` and have a struct like:

```
type Properties struct {
        componentType string
}
```

valaparthvi: I do want to restrict the possibility of values stored in the context. I tried what you have suggested, i.e. having a Properties struct, but it seemed to get complicated, mainly for 2 reasons -
1. We are currently using maps, which are easier to deal with mainly because they allow easy iteration.
2. We are sending this data to segment.io along with some other information, if we use struct, we'll need to convert them to maps so that we can iterate and all information there is, or we will need to explicitly add every information.

Perhaps I am missing something, it might help if you can elaborate on your suggestion.

>It could be interesting to have this code in a separate package.

It makes sense to move to a separate package if we will use it somewhere other than here, else I can move the code to a different file within the same package for better readability.

feloy: > I do want to restrict the possibility of values stored in the context. I tried what you have suggested, i.e. having a Properties struct, but it seemed to get complicated, mainly for 2 reasons -
> 
> 1. We are currently using maps, which are easier to deal with mainly because they allow easy iteration.
> 2. We are sending this data to segment.io along with some other information, if we use struct, we'll need to convert them to maps so that we can iterate and all information there is, or we will need to explicitly add every information.
> 
> Perhaps I am missing something, it might help if you can elaborate on your suggestion.
> 

OK, storing the data in the desired form is a good idea. Let's keep it as is.

> > It could be interesting to have this code in a separate package.
> 
> It makes sense to move to a separate package if we will use it somewhere other than here, else I can move the code to a different file within the same package for better readability.

If you want to restrict the possible values and prevent the future developers to be tempted to use the `set` function directly, it would be better to place these functions in a separate package, so the `set` function is unaccessible. In a different file of  the same package, it will remain accessible.

valaparthvi: Done. I moved the context to another package within the segment package since I think it is more relevant to it and I don't expect people to use it for other packages.

pkg/segment/context/context.go
```diff
@@ -0,0 +1,61 @@
...
+	defer p.lock.Unlock()
+	p.storage[name] = value
+}
+
+func (p *properties) values() map[string]interface{} {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	ret := make(map[string]interface{})
+	for k, v := range p.storage {
+		ret[k] = v
```
feloy, line 28: You are doing this to return a deep copy of the map, right?
Are we sure all values of the map will be copied this way? 
Here, If v were a map, I think it would be referenced instead

pkg/segment/context/context.go
```diff
@@ -0,0 +1,61 @@
...
+type contextKey struct{}
+
+var key = contextKey{}
+
+type properties struct {
+	lock    sync.Mutex
+	storage map[string]interface{}
+}
+
+func (p *properties) set(name string, value interface{}) {
```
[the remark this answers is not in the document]
valaparthvi: To be honest, a lot of this code is copied from another project([crc](https://github.com/code-ready/crc/blob/master/pkg/crc/telemetry)), I modified it for our usage, most of the things are unchanged, so if you think something needs to be changed, let me know. I let these methods be private because I don't intend to use them outside of this file.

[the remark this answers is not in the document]
feloy: Yes, it is better to keep these methods private.

The only thing I would change would be to place the public methods/functions before the private ones in the file.

pkg/segment/context/context.go
valaparthvi: Yes, for a deep copy. The chances of v being a map are pretty slim in the case of telemetry, but I guess we can use json instead to fix this.

feloy: To be sure that we have only strings (and ints probably later) stored in the map, you can set as private the `setContextProperties` (as in the original source), and place `SetComponentType` in the segment/context package, so only this method is callable and is allowed to set values into the map, so we are sure values are strings, which will be copied correctly

pkg/segment/context/context.go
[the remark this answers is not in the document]
valaparthvi: Ah right, I misread your comment the first time :facepalm:

pkg/odo/cli/component/create.go
```diff
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	segment "github.com/openshift/odo/pkg/segment/context"
```
feloy, line 10: This has probably been set by your IDE, but is confusing. You should instead write `context.SetComponentType` in line 911

feloy:
/hold

Feel free to unhold when ready to merge

PR #4669, opened 2021-04-29
Adds a check to ignore pods which are marked for deletion

### LINKED ISSUE #4655: odo should not show PVC in terminating state
```

▶ odo delete
? Are you sure you want to delete the devfile component: java-springboot-springboot-iiun? Yes

Gathering information for component java-springboot-springboot-iiun
 ✓  Checking status for component [137ms]

Deleting component java-springboot-springboot-iiun
 ✓  Deleting Kubernetes resources for component [136ms]


▶ oc get pvc       
NAME                                                 STATUS        VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
first-java-springboot-springboot-iiun-gmrg           Terminating   pvc-3c08d919-baa9-4722-bf3f-66d05dd642bd   1Gi        RWO            standard       3m58s
springbootpvc-java-springboot-springboot-iiun-tugp   Terminating   pvc-d882e055-6cdf-4b82-bf78-48ddb8352f1f   3Gi        RWO            standard       3m58s

▶ odo storage list
The component 'java-springboot-springboot-iiun' has the following storage attached:
NAME              SIZE     PATH                CONTAINER     STATE
springbootpvc     3Gi      /data/cache/.m2     tools         Pushed
first             1Gi      /data1              tools         Pushed
springbootpvc     3Gi      /data/cache/.m2     runtime       Pushed

```

in this scenario odo should not show any storage, as they are all in terminating state.

/kind bug
/priority low

mik-dass:
**What type of PR is this?**

/kind bug

It adds a check to ignore pods which are marked for deletion.

Fixes #4655 

**PR acceptance criteria**:

- [X] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Create a storage and push the changes to the component `odo storage create && odo push`
- Delete the component and then list the storage `odo delete -f && odo storage list`.

### DIFF
--- a/pkg/kclient/pods.go
+++ b/pkg/kclient/pods.go
@@ -208,6 +208,11 @@ func (c *Client) GetOnePodFromSelector(selector string) (*corev1.Pod, error) {
 		return nil, fmt.Errorf("multiple Pods exist for the selector: %v. Only one must be present", selector)
 	}
 
+	// check if the pod is in the terminating state
+	if pods.Items[0].DeletionTimestamp != nil {
+		return nil, &PodNotFoundError{Selector: selector}
+	}
+
 	return &pods.Items[0], nil
 }
 
--- a/pkg/kclient/pods_test.go
+++ b/pkg/kclient/pods_test.go
@@ -5,6 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"reflect"
 	"testing"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -99,6 +100,11 @@ func TestGetOnePodFromSelector(t *testing.T) {
 	fakePod := FakePodStatus(corev1.PodRunning, "nodejs")
 	fakePod.Labels["component"] = "nodejs"
 
+	fakePodWithDeletionTimeStamp := FakePodStatus(corev1.PodRunning, "nodejs")
+	fakePodWithDeletionTimeStamp.Labels["component"] = "nodejs"
+	currentTime := metav1.NewTime(time.Now())
+	fakePodWithDeletionTimeStamp.DeletionTimestamp = &currentTime
+
 	type args struct {
 		selector string
 	}
@@ -136,6 +142,17 @@ func TestGetOnePodFromSelector(t *testing.T) {
 				Items: []corev1.Pod{
 					*fakePod,
 					*fakePod,
+				},
+			},
+			want:    &corev1.Pod{},
+			wantErr: true,
+		},
+		{
+			name: "pod is in the deletion state",
+			args: args{selector: fmt.Sprintf("component=%s", "nodejs")},
+			returnedPods: &corev1.PodList{
+				Items: []corev1.Pod{
+					*fakePodWithDeletionTimeStamp,
 				},
 			},
 			want:    &corev1.Pod{},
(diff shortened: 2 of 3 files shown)


dharmit (changes requested):
> **How to test changes / Special notes to the reviewer**:
> 
>   * Create a storage and push the changes to the component `odo storage create && odo push`
> 
>   * Delete the component and then list the storage `odo delete -f && odo storage list`.

Upon trying this, I see that the storage attached to the component is showing up as `Not Pushed` in `odo storage list`. The issue talks about not showing the storage at all, but `Not Pushed` seems like the right behaviour to me.

However, the unit test added here doesn't test the behaviour of `odo storage list` command that we're trying to address here. Can we please add an integration test for the same?

mik-dass:
> Upon trying this, I see that the storage attached to the component is showing up as Not Pushed in odo storage list. The issue talks about not showing the storage at all, but Not Pushed seems like the right behaviour to me.

Yes `Not Pushed` is the right behavior because the devfile.yaml and the env.yaml still exists for the component.

> However, the unit test added here doesn't test the behaviour of odo storage list command that we're trying to address here. Can we please add an integration test for the same?

Sure but there is no guarantee that the integration test would be able to test this properly as the pod needs to be marked for deletion but should exist on the cluster. If the pod is deleted quickly on the cluster, the test won't be able to catch it.

dharmit:
> Sure but there is no guarantee that the integration test would be able to test this properly as the pod needs to be marked for deletion but should exist on the cluster. If the pod is deleted quickly on the cluster, the test won't be able to catch it.

I was thinking more along the lines of checking `odo storage list` output. Kinda covering what's mentioned in the original issue.

mik-dass:
> I was thinking more along the lines of checking odo storage list output. Kinda covering what's mentioned in the original issue.

I meant even if I check the list output, there is no guarantee that the integration test would be able to test this properly.

dharmit:
> I meant even if I check the list output, there is no guarantee that the integration test would be able to test this properly.

Really? Why? It's a genuine question because when I played with the functionality locally, it worked just fine.
If you think it'll take a few seconds for the output to reflect correctly, we could use `helper.WaitForCmdOut`.

Just to be clear, I don't mean that I am not going to LGTM without an integration test. I'm just trying to understand why it would be unreliable.

mik-dass:
> Really? Why? It's a genuine question because when I played with the functionality locally, it worked just fine.
If you think it'll take a few seconds for the output to reflect correctly, we could use helper.WaitForCmdOut.

If the pod is deleted quickly before we execute list, the output would be `Not Pushed` even without this fix and `helper.WaitForCmdOut` would exit after the first check. We need the pod to be marked for deletion but existing on the cluster.

But since it worked for you locally, I guess we can ignore the above and add a integration test. Let me find a good place to keep the check.

dharmit:
> We need the pod to be marked for deletion but existing on the cluster.

I think this part is where my understanding is different from yours. In the integration test, I was only considering checking output of `odo storage list` without being concerned about the state of the pod. Pod is indeed getting deleted fast.

> But since it worked for you locally, I guess we can ignore the above and add a integration test.

I used springboot example - `odo create java-springboot --starter`. It showed the pod to be in `Terminating` state for some time.

mik-dass:
@dharmit I have added the integration test. Please have a look.

dharmit (approved):
Thanks for the conversation about integration test! 👍🏾

PR #4525, opened 2021-03-18
OCI-based devifle registry support

GeekArthur:
Signed-off-by: jingfu wang <jingfwan@redhat.com>

**What type of PR is this?**
/kind feature

**What does this PR do / why we need it**:
This PR implements the OCI-based devfile registry support, more specifically:
1. Consume devfile registry library or REST API (list devfile components + create devfile component)
2. Change the default devfile registry URL to OCI-based public devfile registry

**Which issue(s) this PR fixes**:

Related issues:
1. https://github.com/devfile/api/issues/367
2. https://github.com/openshift/odo/issues/4504

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

- [x] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:
1. Run `odo registry add OCI https://registry.devfile.io` to add OCI-based public devfile registry if it's not in the registry list
2. Run `odo catalog list components` then verify the devfiles from OCI-based public devfile registry can be listed
3. Run `odo create nodejs --registry OCI` then verify the nodejs devfile can be downloaded from OCI-based public devfile registry

PR #4679, opened 2021-04-30
Update devfile tests with OCI-based registry

josephca:
**What type of PR is this?**
 /kind cleanup
 
After moving to use OCI-based registry (https://github.com/openshift/odo/pull/4525), devfile tests should be using OCI-based registry as well. 

Related issues:
https://github.com/devfile/api/issues/367
https://github.com/openshift/odo/issues/4504
https://github.com/openshift/odo/pull/4525

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

This PR is for changing test codes.

### DIFF
--- a/tests/integration/devfile/cmd_devfile_catalog_test.go
+++ b/tests/integration/devfile/cmd_devfile_catalog_test.go
@@ -11,7 +11,7 @@ import (
 var _ = Describe("odo devfile catalog command tests", func() {
 	const registryName string = "RegistryName"
 	// Use staging OCI-based registry for tests to avoid overload
-	const addRegistryURL string = "https://registry.devfile.io"
+	const addRegistryURL string = "https://registry.stage.devfile.io"
 
 	var commonVar helper.CommonVar
 
--- a/tests/integration/devfile/cmd_devfile_catalog_test.go
+++ b/tests/integration/devfile/cmd_devfile_catalog_test.go
@@ -10,7 +10,9 @@ import (
 
 var _ = Describe("odo devfile catalog command tests", func() {
 	const registryName string = "RegistryName"
-	const addRegistryURL string = "https://github.com/odo-devfiles/registry"
+	// Use staging OCI-based registry for tests to avoid overload
+	const addRegistryURL string = "https://registry.devfile.io"
+
 	var commonVar helper.CommonVar
 
 	// This is run before every Spec (It)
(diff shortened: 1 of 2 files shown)


rnapoles-rh (commented):
@mohammedzee1000  Looks like both:

- https://github.com/odo-devfiles/registry

and 

- https://registry.devfile.io

have the same content, why use different URLs?

josephca:
> @mohammedzee1000 Looks like both:
> 
> * https://registry.stage.devfile.io
> 
> and
> 
> * https://registry.devfile.io
> 
> have the same content, why use different URLs?

@rnapoles-rh 
I intentionally used two registries to test out different test cases against the official OCI-based registry vs. staging registry (contains the latest contents) and also to ensure the availability of the registries. 

Here is the detail : 

- Official registry (https://registry.devfile.io) updated weekly vs. staging registry(https://registry.stage.devfile.io) updated with PR merges. 
- odo registry command test uses the official registry as it doesn't cause much overload on the public registry vs. odo catalog test uses the staging registry to avoid potential overload on the official registry.

josephca:
@GeekArthur - please review. Thanks,

girishramnani:
> > @mohammedzee1000 Looks like both:
> > 
> > * https://registry.stage.devfile.io
> > 
> > and
> > 
> > * https://registry.devfile.io
> > 
> > have the same content, why use different URLs?
> 
> @rnapoles-rh
> I intentionally used two registries to test out different test cases against the official OCI-based registry vs. staging registry (contains the latest contents) and also to ensure the availability of both registries.
> 
> Here is the detail :
> 
> * Official registry (https://registry.devfile.io) updated weekly vs. staging registry(https://registry.stage.devfile.io) updated with PR merges.
> * odo registry command test uses the official registry as it doesn't cause much overload on the public registry vs. odo catalog test uses the staging registry to avoid potential overload on the official registry.

So odo tests would fail if there is breaking changing in the staging registry? wouldn't that block us?

prietyc123:
> So odo tests would fail if there is breaking changing in the staging registry? wouldn't that block us?

Even I do feel the same. @josephca Currently the breaking changes in Official registry are bothering us more and now it seems like we need to deal with staging one as well, no?

josephca:
> > So odo tests would fail if there is breaking changing in the staging registry? wouldn't that block us?
> 
> Even I do feel the same. @josephca Currently the breaking changes in devfile registry are bothering us more and now it seems like we need to deal with staging one as well, no?

@girishramnani @prietyc123 Thanks for your inputs. 
After discussion with devfile service team, it would be sufficient to test only with staging server. In addition, they will try to get some telemetry data in the future to track usage patterns on the community registry. Therefore, it will be good to use the staging server for testing to avoid polluting the data.

That said, I will update the code to use the staging server only.

rnapoles-rh (approved):
/lgtm

PR #4732, opened 2021-05-20
Adding container based filtering for get ports

### LINKED ISSUE #4330: "odo url create" asks for port when devfile exposes two ports but a container exposes only one port

**Operating System:** all

**Output of `odo version`:**
```
$ odo version
odo v2.0.3 (96fc76e69)
```

1. Used below devfile.yaml from #4318:
    ```yaml
    schemaVersion: 2.0.0
    metadata:
      name: nodejs
      version: 1.0.0
      attributes:
        alpha.build-context: src
        alpha.build-dockerfile: Dockerfile
    components:
      - name: buildguidance
        container:
          image: buildguidanceimage-placeholder
          memoryLimit: 1024Mi
          endpoints:
            - name: http-3001
              targetPort: 3001
      - name: runtime
        container:
          memoryLimit: 1024Mi
          mountSources: true
          sourceMapping: /project
          endpoints:
            - name: http-3000
              targetPort: 3000
    commands:
      - id: install
        exec:
          component: runtime
          commandLine: npm install
          workingDir: /project/src
          group:
            kind: build
            isDefault: true
      - id: run
        exec:
          component: runtime
          commandLine: npm start
          workingDir: /project/src
          group:
            kind: run
            isDefault: true
    ```
1. Did `odo create nodejs --devfile devfile.yaml`
2. Tried to create URL with `--container` flag.

## Actual behavior
```sh
$ odo url create --host `minikube ip`.nip.io --container buildguidance
 ✗  port for the component nodejs is required as it exposes 2 ports: 3000,3001

$ odo url create --host `minikube ip`.nip.io --container runtime      
 ✗  port for the component nodejs is required as it exposes 2 ports: 3001,3000
```

## Expected behavior
Both `runtime` and `buildguidance` containers in the devfile expose one port each. odo should not ask the user to specify port number in such case and be smart enough to create URL for the port exposed by the specified container.

Kudos for odo being smart to prevent user from creating URL for invalid container-port combination. Below commands rightly fail:
```sh
$ odo url create --host `minikube ip`.nip.io --container runtime --port 3001
 ✗  cannot set URL nodejs-3001 under container runtime, TargetPort 3001 is being used under container buildguidance

$ odo url create --host `minikube ip`.nip.io --container buildguidance --port 3000
 ✗  url nodejs-3000 already exist in devfile endpoint entry under container runtime
```

/area url
/priority medium
/kind bug

mohammedzee1000:

**What type of PR is this?**
/kind bug

This PR adds the ability to filter ports by container name (interface change) thereby fixing the case where more than 1 ports are spread across containers resulting in need to specify a port

Fixes #4330 

**PR acceptance criteria**:

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] Update changelog

See linked issue for error case. As for checking how it is fixed

```bash
❯ ls
devfile.yaml  node-ex  package.json  sample  script-validator  server.js  test-repo
❯ cat devfile.yaml
schemaVersion: 2.0.0
metadata:
  name: nodejs
  version: 1.0.0
  attributes:
    alpha.build-context: src
    alpha.build-dockerfile: Dockerfile
components:
  - name: buildguidance
    container:
      image: buildguidanceimage-placeholder
      memoryLimit: 1024Mi
      endpoints:
        - name: http-3001
          targetPort: 3001
  - name: runtime
    container:
      memoryLimit: 1024Mi
      mountSources: true
      sourceMapping: /project
      endpoints:
        - name: http-3000
          targetPort: 3000
commands:
  - id: install
    exec:
      component: runtime
      commandLine: npm install
      workingDir: /project/src
      group:
        kind: build
        isDefault: true
  - id: run
    exec:
      component: runtime
      commandLine: npm start
      workingDir: /project/src
      group:
        kind: run
        isDefault: true
❯ odo url create --host com --container buildguidance --ingress
 ✓  URL nodejs-testbench-qjcp-3001 created for component: nodejs-testbench-qjcp

❯ odo url create --host com --container runtime --ingress
 ✓  URL nodejs-testbench-qjcp-3000 created for component: nodejs-testbench-qjcp

```


mohammedzee1000:
Unrelated failure in windows UT https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4732/pull-ci-openshift-odo-main-psi-unit-test-windows/1395286598446223360
/test psi-unit-test-windows

mohammedzee1000:
Again failed due to same issue. Not related to code of this pr
/retest

pkg/localConfigProvider/localConfigProvider.go
```diff
@@ -63,7 +63,7 @@ type LocalConfigProvider interface {
 	ValidateURL(url LocalURL) error
 	CreateURL(url LocalURL) error
 	DeleteURL(name string) error
-	GetPorts() ([]string, error)
+	GetPorts(container string) ([]string, error)
```
feloy, line 66: I think it would be easier to understand the behaviour of this function by splitting it into 2 functions:

- `GetContainerPorts(container string)`
- `GetComponentPorts()`

(I'm not sure about the naming of the second one)

The two functions could call a same private function handling the 2 cases if you want to not duplicate too much code.

mohammedzee1000: I guess something like that could be done. I just wanted to achieve this with minimal code change, but i guess refactoring stuff also makes sense.

pkg/localConfigProvider/mock_localConfigProvider.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by MockGen. DO NOT EDIT.
 // Source: localConfigProvider.go
 
-// Package localConfigProvider is a generated GoMock package.
+// Package mock_localConfigProvider is a generated GoMock package.
```
mik-dass, line 4: The package is `localConfigProvider`

mohammedzee1000: Updated tho i dont get why it is still showing this in diff

pkg/envinfo/url.go
```diff
@@ -15,31 +14,45 @@ import (
 	"github.com/pkg/errors"
 )
 
-// GetPorts returns the ports, returns default if nil
-func (ei *EnvInfo) GetPorts() ([]string, error) {
+//getPorts gets the ports from devfile
+func (ei *EnvInfo) getPorts(container string) ([]string, error) {
 	var portList []string
-
+	portMap := make(map[string]bool)
```
feloy, line 20: I think you can let the definition of portMap at its previous place, as it is not used before.

pkg/envinfo/url_test.go
```diff
@@ -343,6 +344,31 @@ func TestEnvInfo_CompleteURL(t *testing.T) {
 				Container: "runtime",
 			},
 		},
+		{
+			name: "case 15: Does not error out if no port is specified, but container will single port is specified in multi container devfile",
```
feloy, line 348:
```suggestion
			name: "case 15: Does not error out if no port is specified, but container with single port is specified in multi container devfile",
```

pkg/envinfo/url.go
```diff
@@ -15,31 +14,45 @@ import (
...
 	}
-
+	if !containerExists {
+		return portList, fmt.Errorf("container named %s does not exist", container)
+	}
 	for port := range portMap {
 		portList = append(portList, port)
 	}
-
-	sort.Strings(portList)
```
feloy, line 39: Why not sorting the ports anymore? That could break some tests and users scenarios

pkg/envinfo/url_test.go
```diff
@@ -364,7 +390,7 @@ func TestEnvInfo_CompleteURL(t *testing.T) {
 			if tt.wantErr && err != nil {
 				return
 			}
-
+			fmt.Printf("%#v\n", tt.args.url)
```
feloy, line 393: You can remove this debug

pkg/envinfo/url.go
mohammedzee1000: Yea i guess this can make tests flaky, adding it back

pkg/envinfo/url.go
mohammedzee1000: Moved to just before for loop

pkg/envinfo/url_test.go
mohammedzee1000: Fixed

pkg/localConfigProvider/mock_localConfigProvider.go
mik-dass: Guess it's gone now from the diff

pkg/config/url.go
```diff
@@ -10,20 +10,24 @@ import (
...
-func (lc *LocalConfig) GetPorts() ([]string, error) {
+func (lc *LocalConfig) GetComponentPorts() ([]string, error) {
 	if lc.componentSettings.Ports == nil {
 		return nil, nil
 	}
 	return *lc.componentSettings.Ports, nil
 }
 
+func (lc *LocalConfig) GetContainerPorts(container string) ([]string, error) {
+	return nil, fmt.Errorf("Not implemented for local config")
```
mik-dass, line 21: I think we can call GetComponentPorts instead of the error here as there is only one container in case of s2i components.

mohammedzee1000: I think it makes sense that function handles only what it is intended for. I would expect `GetContainerPorts` only to get ports for the specified containers and if no container being specified is an error state in my book

If you check the localconfig changes for s2i. It only calls `GetComponentPorts` directly and never calls `GetContainerPorts`

mik-dass: > If you check the localconfig changes for s2i. It only calls GetComponentPorts directly and never calls GetContainerPorts

OK. Please change the error message to start with a lower case.

mohammedzee1000: fixed

mohammedzee1000:
Recurrence of the same issue https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4732/pull-ci-openshift-odo-main-psi-kubernetes-integration-e2e/1395704113907372032
/retest

pkg/envinfo/url.go
```diff
@@ -62,9 +77,18 @@ func (ei *EnvInfo) CompleteURL(url *localConfigProvider.LocalURL) error {
...
-		return err
+	var ports []string
+	var err error
+	if url.Container == "" {
+		ports, err = ei.GetComponentPorts()
+		if err != nil {
+			return err
+		}
+	} else {
+		ports, err = ei.GetContainerPorts(url.Container)
```
mik-dass, line 88: When no port is found for the specified container, the error message is
```
[mrinaldas@localhost project]$ odo url create example --container runtime-1 --host com
 ✗  no port is exposed by the component nodejs-project-hron, please specify a port
```
It feel it should say `no port is exposed by the container <container-name> ....`

mohammedzee1000: Ok pushing some changes

pkg/envinfo/url.go
```diff
@@ -78,23 +78,26 @@ func (ei *EnvInfo) CompleteURL(url *localConfigProvider.LocalURL) error {
...
 		}
+		portsOf = fmt.Sprintf("container %s", url.Container)
 	}
 	if url.Port == -1 {
 		if len(ports) > 1 {
-			return fmt.Errorf("port for the component %s is required as it exposes %d ports: %s", ei.GetName(), len(ports), strings.Trim(strings.Replace(fmt.Sprint(ports), " ", ",", -1), "[]"))
+			return fmt.Errorf("port for the %s is required as it exposes %d ports: %s", portsOf, len(ports), strings.Trim(strings.Replace(fmt.Sprint(ports), " ", ",", -1), "[]"))
 		} else if len(ports) <= 0 {
-			return fmt.Errorf("no port is exposed by the component %s, please specify a port", ei.GetName())
+			return fmt.Errorf("no port is exposed by the %s, please specify a port", portsOf)
```
mik-dass, line 100: Instead of an extra variable, we can also check against the value of `url.Container` since that triggers the earlier condition.

mohammedzee1000: Hmm i guess that makes sense

mohammedzee1000: Although I think that would mean repetition of either the condition check(`if url.Container` check) or the entire error check blob twice in the if and else block, so I feel this is more efficient

mik-dass: But if `url.Port != -1`, the variable will be kept in memory unnecessarily throughout rest of the run.

mohammedzee1000: hmm ok, guess the better thing to do is to make a function out of this at this point

pkg/envinfo/url.go
```diff
@@ -55,6 +55,22 @@ func (ei *EnvInfo) GetComponentPorts() ([]string, error) {
 	return ei.getPorts("")
 }
 
+func (ei *EnvInfo) completeNoPortSpecified(url *localConfigProvider.LocalURL, portsOf string, ports []string) (err error) {
```
mik-dass, line 58: Maybe just `checkValidPort` as the function doesn't complete anything

pkg/envinfo/url.go
```diff
@@ -85,13 +101,19 @@ func (ei *EnvInfo) CompleteURL(url *localConfigProvider.LocalURL) error {
...
 		if err != nil {
 			return err
 		}
-		portsOf = fmt.Sprintf("container %s", url.Container)
+		err = ei.completeNoPortSpecified(url, fmt.Sprintf("container %s", ei.GetName()), ports)
+		if err != nil {
+			return err
+		}
 	}
 	if url.Port == -1 {
```
mik-dass, line 118: I think we can safely remove this as the check has been moved in the if-else condition above

mohammedzee1000: Ah damn missed that :(

pkg/envinfo/url.go
mohammedzee1000: Ok updating

mik-dass (approved):
/lgtm

PR #4742, opened 2021-05-24
List components from other sources

### LINKED ISSUE #4688: List components created by sources other than odo
/kind user-story

## User Story
As a user, I want to be able to see components created by other tools such as ODC(and using labels as stated by the guideline [here](https://github.com/redhat-developer/app-labels).) when I list components with odo.  So that it will be easier for me to move from a project that started on the developer console or any other tool to odo.

## Acceptance Criteria 
- [ ] It should list components when running `odo list`, considering the component's use labels mentioned [here](https://github.com/redhat-developer/app-labels). and show the source(?)
- [ ] `odo list` should be able to distinguish between local and remote components and remote components must be read-only.
- [ ] Output of `odo list` to look something like this(up for a debate) - 
      
```
➜  odo git:(listcom) ✗ odo list --all-apps
Devfile Components: 
APP      NAME      PROJECT       TYPE      STATE
app1     njs2i     myproject     njs2i     Pushed

Other Components running on the cluster(read-only): 
APP            NAME              PROJECT       TYPE
sample-app     nodejs-sample     myproject     nodejs-sample
httpd          example           myproject     example
```

## Links 
- Related Epic (mandatory): #4192 , [ODO-59](https://issues.redhat.com/browse/ODO-59)

/kind user-story

valaparthvi:
**What type of PR is this?**
/kind feature
/kind code-refactoring

This PR adds the support to see components created/managed by tools other than odo.

Fixes #4688

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] Update changelog

Manually create a deployment and deploymentconfig in openshift web console or any other tool. Make sure it is labelled with the following labels - 
```
        app: httpd
        app.kubernetes.io/instance: example
        app.kubernetes.io/managed-by: console
        app.kubernetes.io/managed-by-version: v4.8.0
        app.kubernetes.io/name: example
        app.kubernetes.io/part-of: httpd
```
Or better yet, it follows these labels - https://github.com/redhat-developer/app-labels

With OCP4, it will be easy to create a component of a deployment kind by using the samples provided by ocp, but to create a deploymentConfig kind component you can use the following yaml - 
```
apiVersion: apps.openshift.io/v1
kind: DeploymentConfig
metadata:
  name: example
  namespace: myproject
  labels:
    app: httpd
    app.kubernetes.io/instance: example
    app.kubernetes.io/managed-by: console
    app.kubernetes.io/managed-by-version: v4.8.0
    app.kubernetes.io/name: example
    app.kubernetes.io/part-of: httpd
spec:
  selector:
    app: httpd
  replicas: 3
  template:
    metadata:
      labels:
        app: httpd
        app.kubernetes.io/instance: example
        app.kubernetes.io/managed-by: console
        app.kubernetes.io/managed-by-version: v4.8.0
        app.kubernetes.io/name: example
        app.kubernetes.io/part-of: httpd
    spec:
      containers:
        - name: httpd
          image: >-
            image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest
          ports:
            - containerPort: 8080
```

Also create a few components with the help of odo.

Run `odo list` and you should be able to see the non-odo components in the output that looks something like this - 
```
➜  odo git:(listcom) ✗ odo list --all-apps
W0524 13:12:51.685576   13526 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0524 13:12:55.884893   13526 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0524 13:12:58.278964   13526 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
Devfile Components: 
APP      NAME      PROJECT       TYPE      STATE
app1     njs2i     myproject     njs2i     Pushed

Other Components running on the cluster(read-only): 
APP            NAME              PROJECT       TYPE
sample-app     nodejs-sample     myproject     nodejs-sample
httpd          example           myproject     example

```
Also, do test with `odo list --app sample-app`, `odo list --all-apps -o json`, `odo list`.

Note - This is work-in-progress and you might find some commented code and incomplete tests, I am yet to add unit and integrations tests as well, but I am waiting on it to make sure that everyone agrees on the implmentation.


pkg/component/component.go
```diff
@@ -929,11 +923,12 @@ func ListS2IComponents(client *occlient.Client, applicationName string, localCon
 
 		// create a list of object metadata based on the component and application name (extracted from DeploymentConfig labels)
 		for _, elem := range dcList {
-			component, err := GetComponent(client, elem.Labels[componentlabels.ComponentLabel], applicationName, client.Namespace)
+			component, err := GetComponent(client, elem.Labels[componentlabels.ComponentLabel], elem.Labels[applabels.ApplicationLabel], client.Namespace)
 			if err != nil {
 				return ComponentList{}, errors.Wrap(err, "Unable to get component")
 			}
```
valaparthvi, line 929: I have a question here. Why do we return if `err!=nil` even though there are some other checks pending in the function, for e.g. this check - https://github.com/openshift/odo/pull/4742/files#diff-a84453ec52dba47b60234fd3aba56849e52e2bae2577c082dcf1a79ef6b25855R946 ?
```
if client != nil {
```

Why do we avoid that and return? Shouldn't we check for `if err == nil { continue }` and execute rest of the code? I have seen this type of logic in almost all places. For some it makes sense, but for others I think we should continue executing the program and if it does not make sense to do so, then perhaps we should add a comment about why we are returning without executing the rest of the code.

Other examples are - 
1. pkg/odo/cli/component/list.go:264 - https://github.com/openshift/odo/pull/4742/files#diff-a5ee7fb0c76cd592580dce46cda117d8a483452979b0da40d644906f88cba343R264
2. pkg/odo/cli/component/list.go:274 - https://github.com/openshift/odo/pull/4742/files#diff-a5ee7fb0c76cd592580dce46cda117d8a483452979b0da40d644906f88cba343R274
3. ..and many more

tests/integration/component.go
```diff
@@ -159,6 +159,25 @@ func componentTests(args ...string) {
 			helper.MatchAllInOutput(cmpList, []string{"cmp-git", "Not Pushed"})
 			helper.CmdShouldPass("odo", append(args, "delete", "-f", "--all", "--context", commonVar.Context)...)
 		})
+		Context("when components are not created/managed by odo", func() {
+			var appName, compName string
+			JustBeforeEach(func() {
+				//CreateNonOdoComponent
```
valaparthvi, line 165: Can someone help me with this one? I plan on creating a deployment and deployment-config manually via API, but I need to prepare a lot of data before doing that. Do we already have helper functions that might help here?

kadel: you can just create `deployment.yaml` with the most simple deployment that you can come up with. Add the same labels to it as WebConsole does and just call `kubectl apply -f deplyomet.yaml`

pkg/component/component.go
```diff
@@ -1651,13 +1642,16 @@ func GetMachineReadableFormatForList(s2iComps []Component) ComponentList {
 }
 
 // GetMachineReadableFormatForCombinedCompList returns list of devfile and s2i components in machine readable format
```
girishramnani, line 1644: maybe update this comment?

pkg/util/util.go
```diff
@@ -82,6 +82,10 @@ type DownloadParams struct {
 }
 
 // ConvertLabelsToSelector converts the given labels to selector
+// To pass operands such as !=, append a ! prefix to the value.
+// For E.g. map[string]string{"app.kubernetes.io/managed-by": "!odo"}
+// Using != operators also means that resource will be filtered even if it doesn't have the key.
+// So a resource not labelled with key "app.kubernetes.io/managed-by" will also be returned.
 func ConvertLabelsToSelector(labels map[string]string) string {
```
girishramnani, line 89: why extend the labels here to be a list of structs here like 
```
struct {
Key string
Value string
Operand Operand
}

type Operand string

var Equal Operand = "="
var NotEqual Operand = "!="
```

might be more future proof?

valaparthvi: Can you elaborate?

kadel:
tested, it it looks good
/approve

valaparthvi:
/test v4.7-integration-e2e

The failing test passes locally. Retrying.

pkg/util/util.go
girishramnani: instead of this label map
```
labels map[string]string
```


```
struct {
Key string
Value string
Operand Operand
}

type Operand string

var Equal Operand = "="
var NotEqual Operand = "!="

```

So the function would look like this `ConvertLabelsToSelector(labels []Label)`.

So I can do something like this 
```
labels := []Label{
   {
   Key: "app.kubernetes.io/managed-by",
   Operand: NotEqual,
    Value: "odo"
     }
}

ConvertLabelsToSelector(labels)
```

The conversion of `[]Label` to `string` would also become simplified

girishramnani: This would allow for room for extension and also less mistakes

girishramnani:
/hold 
Just want to make sure that someone doesn't lgtm until all comments are resolved 🙂

pkg/util/util.go
valaparthvi: It seems to me as if this will require a significant amount of effort and it seems a little out of scope for this PR. Is it alright if I take this in another PR?

Also, can you give me an example of possible mistakes that could happen here?

girishramnani: e.g. someone did `!!odo` the logic of `=` and `!=` is dependent on `!` and also it doesn't come as natural to the reader of the code. they have to go to the function and read its comment ( which is fine ) but says that there is room for improvement.
Also the `Label` approach allows for simpler extending in future.

girishramnani: But about not being in scope, I am not sure about that as the logic of `!` was introduced in this PR 🙂 . So the refactor for the code would probably be part of this PR too?

dharmit: @valaparthvi I'm not sure why someone would try to set `"app.kubernetes.io/managed-by": "!odo"` as a label to the Kubernetes resources created by odo. At least in the odo code we don't do it AFAIK. What's the rationale here?

I have not looked at full code so maybe I'm missing the context here.

girishramnani: https://github.com/openshift/odo/pull/4742/files#diff-71f82b8ba7b20d4531fabf0bf878e138fa960894290d65d12e5804c162e39440R55-R61
https://github.com/openshift/odo/pull/4742/files#diff-0fa449830b5dc2b274b7c95a58b5071edc23f1e44af84a99c1fa9e600767829fR85-R88

girishramnani: the comment of the function explains this 
```
// To pass operands such as !=, append a ! prefix to the value.// For E.g. map[string]string{"app.kubernetes.io/managed-by": "!odo"}// Using != operators also means that resource will be filtered even if it doesn't have the key.// So a resource not labelled with key "app.kubernetes.io/managed-by" will also be returned.
```

valaparthvi: >@valaparthvi I'm not sure why someone would try to set "app.kubernetes.io/managed-by": "!odo" as a label to the Kubernetes resources created by odo. At least in the odo code we don't do it AFAIK. What's the rationale here?

I am using it to filter the resources(deployments/dc) not managed by odo. Hence `!odo` part.

valaparthvi: >e.g. someone did !!odo the logic of = and != is dependent on ! and also it doesn't come as natural to the reader of the code. they have to go to the function and read its comment ( which is fine ) but says that there is room for improvement.
Also the Label approach allows for simpler extending in future.

Okay, that makes sense. But implementing this would probably affect a lot of other tests, I would much rather work on that in a different PR.

valaparthvi:
/retest
Unit test windows failure.

pkg/util/util.go
kadel: if you change `strings.Replace(v, "!", "", -1)` to `strings.Replace(v, "!", "", 1)` wouldn't it the problem of doing `!!odo`?

valaparthvi: @kadel I am not sure I understand your problem correctly.

girishramnani: I am good from my end, please address @kadel's question and then remove the hold

kadel: Just realized something. `!` can't be present in a label or its value.
I think that we are ok with this. front. 

But I think that I'm really missing here is a unit test for this function :-( It became much more complicated than it used to be, and it definetly deserves unit test

valaparthvi: @kadel Unit tests! That can be easily fixed! I'll get to it.

@girishramnani I looked at how k8s.io/apimachinery does this label selector thing and it is a little complicated, they have special interface and methods for it. See here - https://github.com/openshift/odo/blob/89ac67e425bffe8c400645bca933d687b5ac8b19/vendor/k8s.io/apimachinery/pkg/fields/selector.go#L430

I will rely on Unit tests for now, and hopefully we will not run into corner cases, or catch them if we do.

valaparthvi:
/unhold
Addressed concern in https://github.com/openshift/odo/pull/4742#discussion_r650881978

valaparthvi:
/override ci/prow/psi-unit-test-mac
/override ci/prow/psi-unit-test-windows

Reason - These are optional checks.

pkg/util/util_test.go
```diff
@@ -2684,3 +2684,48 @@ func TestGetGitOriginPath(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertLabelsToSelector(t *testing.T) {
```
kadel, line 2688: :heart:

PR #4751, opened 2021-05-26
Revised Proposal to param based operator service creation

### LINKED ISSUE #4240: Proposal for improved and interactive operator backed services
/kind user-story

## User Story
As a developer the proposal should contain the workflow for improved interactive and non-interactive operator backed services workflow driven by schema present in the CRDs.

## Acceptance Criteria 
- [ ] It should consider both kuberenetes and openshift parts if they are different
- [ ] It should `odo service create` interactive and non-interactive flows
- [ ] It should consider validating parameters provided on non-interactive flows
- [ ] It should consider a route to still use the `alm-examples` of the CR
- [ ] It should consider providing useful information on `odo catalog describe services` command
- [ ] It should also provide json output that would be provided when these commands are executed
- [ ] These things should be documented in a markdown proposal which when merge would only then accomplish all the ACs of this issue

Note - `It` here is the proposal

### Not in scope
- This proposal **doesn't** scope out the storage of operator backed services in the devfile 
- This proposal **doesn't** scope out creation of operators only on `odo push` ( odo new workflow for operator services )

## Links 
- Feature Request: 
- Related issue: 
- https://github.com/openshift/odo/issues/2785
- https://github.com/openshift/odo/issues/2615

/kind user-story

girishramnani:
**What type of PR is this?**
/kind design
/kind documentation
[skip ci]

see $SUBJECT

Fixes #4240

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [X] Documentation 

- [ ] Update changelog


docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+
+This proposal is about improving the experience for a user of the operator backed services. Currently a user doesn't know exaustively which fields are available in a CR as we depend on optional metadata present in CRDs, to improve this we are considering using the metadata available from the cluster about the CRDs.
+Getting that metadata from the cluster is challenging because a normal cluster user ( plain vanilla ) doesn't have access to that metadata so we are gonna follow a 3 tiered approach
+
+- check if the user has access to the CRD, if so then use that.
+- get the `swagger.json` from the openapi endpoint provided by kubernetes. a sample can be found here https://gist.github.com/girishramnani/f29949d5cb8c6547102776437e05ac19
+- finally if we cannot get the openapi data as well then use the `ClusterServiceVersion` to get the parameters.
+
+This change would affect multiple service commands and the changes are described briefly below -
+- `odo catalog describe service` should include the metadata fields with description so the users can provide then when doing `odo service create`
```
kadel, line 12: To just make sure that we don't forget this:
Every command should have working and updated `-o json`.

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+Sample output - 
+https://gist.github.com/girishramnani/cbb4400e463efe89c13f1386e0788793
+
+We care about the `openAPIV3Schema` as that would be used to build the CRD struct. 
+
+### Kuberenetes cluster Swagger has the schema
+
+If the user doesn't have CRD access then we fetch try to fetch the same `openAPIV3Schema` from the cluster's `swagger.json` from the endpoint `<cluster-url>/api/kubernetes/openapi/v2`. This is a very large document as it holds all the definitions present on the cluster.
+
+So this needs to be cached and refreshed whenever a new operator is installed.
```
kadel, line 38: How we will detect that it needs to be refreshed? Storing the hash of all installed operators and comparing it if it didn't change?

We also need to keep in mind that users might work with multiple clusters, so the cache needs to be per cluster.

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+```
+
+## Allowing user to set the parameters 
+
+The current approach of sending map values via cli are as follows - 
+
+- we would add a `-p` cobra param as a list
+- each of the parameter would represent the key in a map and value in a map `e.g. -p "key"="value"`
+- we would allow json path in the key for the user to specific any field in the map that they want to set e.g. `services[0].namespace`.
+- we will pass all values as string.
```
kadel, line 123: I thought that we already verified that this can't be done.
We need to at least detect and recognize the basic types like int, float, bool,  string

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+  ]
+}
+```
+
+## Allowing user to set the parameters 
+
+The current approach of sending map values via cli are as follows - 
+
+- we would add a `-p` cobra param as a list
+- each of the parameter would represent the key in a map and value in a map `e.g. -p "key"="value"`
```
kadel, line 121: `key` needs to match "parameter name" as shown in `odo catalog component describe`

docs/proposals/improved-schema-for-services.md
girishramnani: https://github.com/openshift/odo/pull/4751#issuecomment-848601517

girishramnani: had a similar approach in my mind shared in the original comment

docs/proposals/improved-schema-for-services.md
girishramnani: yeah this part is older

docs/proposals/improved-schema-for-services.md
girishramnani: adding this

docs/proposals/improved-schema-for-services.md
girishramnani: sure, but we would have different json output structurally for when user has access to `CRD` vs just `CSV`

kadel: > sure, but we would have different json output structurally for when user has access to `CRD` vs just `CSV`

that is not good. We need to align  it

docs/proposals/improved-schema-for-services.md
girishramnani: hashing would reduce the amount of information, and it the cache meta can be stored in `.odo` folder?

kadel: sorry, did not notice that comment before. 

What about the multiple clusters problem? Directory for each cluster based on its api url? But than there might be issues with minikube/crc, where you can delete and create new cluster and url remains the same. It would be nice if we could get some kind of id/uuid of the cluster

docs/proposals/improved-schema-for-services.md
girishramnani: actually if you think about it - 
the openapi json output would look like this
https://gist.github.com/girishramnani/0d56b28108594657a25a002993840872
So there would be nested structures. 
on the other hand `x-descriptor` descriptions are flat as they use `json-path` like syntax to specify values.

From readability standpoint `x-descriptors` is better as a user can just get the `key` from the description and use it in `odo service create`

but the json output of `x-descriptor` is limiting.

girishramnani: @kadel I dont know if we can flatten this `json-schema` structure to something human readable.
there are some arrays in the schema as well so we cannot show arrays as a `json-schema`
e.g.

we have a `secret` list of objects defined in the json schema. if we try to flatten it - we cannot have it like this `zookeeper.secret[0].name` its not always gonna be`0` indexed? or we just make it `zookeeper.secret.name` and assume that its append only list? or `zookeeper.secret[-1].name` like how sjson prefer?

docs/proposals/improved-schema-for-services.md
girishramnani: we can use the `cluster: api-engint-openshift-com:443` i.e. url with port like how kube context does it.

girishramnani: so `{"api-engint-openshift-com:443": "<hash>",....}` would be how the operator-listing-cache.json ( we can decide on the name ) could look like?

docs/proposals/improved-schema-for-services.md
kadel: > @kadel I dont know if we can flatten this `json-schema` structure to something human readable.

you mean for example this form `zookeeper.secret.name`? I think that this is perfectly human readable



> we have a `secret` list of objects defined in the json schema. if we try to flatten it - we cannot have it like this `zookeeper.secret[0].name` its not always gonna be`0` indexed?

Why it wouldn't be 0 indexed?

> or we just make it zookeeper.secret.name and assume that its append only list?

I'm afraid that this wouldn't work. 
Ther will be problems if you have array of objects. Like for example `env`


`-p spec.env.name FOO -p spec.env.value BAR`

Would this resolve to 
```
spec:
  env:
    - name: FOO
      value: BAR
```

or 


```
spec:
  env:
    - name: FOO
    - value: BAR
```

?

kadel: I think that we will have to use indexes like `zookeeper.secret[0].name`

`-p spec.env[0].name FOO -p spec.env[0].value BAR`

```
spec:
  env:
    - name: FOO
      value: BAR
```

`-p spec.env[0].name FOO -p spec.env[1].value BAR`
```
spec:
  env:
    - name: FOO
    - value: BAR
 ```

But we need to figure out how to communicate to user that the given field is array and the index is needed in `odo catalog service describe` output

Comment (girishramnani) [reply docs/proposals/improved-schema-for-services.md:121]: Yeah as I think about it, append only won’t work.

and the reason I said. [0] won’t work I meant that the idea that key from describe cannot be just copied as is to odo service create.
The user has to think that this is an array so it would have a [index].
Maybe we can show it as `secret[*].name` and explain in describe that * needs to be an index 0,1….

File: docs/proposals/improved-schema-for-services.md (hunk 3 above)
Comment (girishramnani) [reply docs/proposals/improved-schema-for-services.md]: fixed

File: docs/proposals/improved-schema-for-services.md (hunk 2 above)
Comment (girishramnani) [reply docs/proposals/improved-schema-for-services.md:39]: added

File: docs/proposals/improved-schema-for-services.md (hunk 1 above)
Comment (girishramnani) [reply docs/proposals/improved-schema-for-services.md]: fixed

File: docs/proposals/improved-schema-for-services.md [hunk 5]
```diff
@@ -0,0 +1,442 @@
...
+
+### Kuberenetes cluster Swagger has the schema
+
+If the user doesn't have CRD access then we fetch try to fetch the same `openAPIV3Schema` from the cluster's `swagger.json` from the endpoint `<cluster-url>/api/kubernetes/openapi/v2`. This is a very large document as it holds all the definitions present on the cluster.
+
+So this needs to be cached and refreshed whenever a new operator is installed.
+
+The caching would be per cluster as the `swagger.json` would change for different clusters. The approach would be as follows -
+
+- when the users executes `odo catalog list services`, if there is no cache then we would fetch the `swagger.json`, odo wouldn't exit until download.
```
dharmit, line 43: > odo wouldn't exit until download

As you mentioned on contributor's call, we had a similar situation with Telemetry. I think this one is worth exploring options where we can do things in background without blocking an odo command. I don't mean to do it in the first pass, but eventually.

I think it's worth it because `odo catalog list services` is going to be used extensively as we pitch odo more and more as a tool that plays well with services. And every time a new Operator is installed, the process of downloading `swagger.json` is going to be repeated. Although these are presumptions and not backed by Telemetry.

docs/proposals/improved-schema-for-services.md
dharmit: > `-p spec.env[0].name FOO -p spec.env[0].value BAR`

Do you folks think this would be good from UX point of view? Taking a step back, why can't we rely only on swagger.json for params backed services? If the user doesn't want to pass parameters, we can use what's there in alm-examples.

Correct me if I am misinterpreting things here.

dharmit:
@girishramnani @kadel I'm probably missing something simple here. Where in the swagger AP (`http://<cluster-url>/openapi/v2`) that we're discussing here is the information about parameters available?

I'm referring to the JSON available at this endpoint and I can't find the parameters  that we see in the alm-example. I'm currently trying to find parameters for etcdcluster.

Besides that, in the JSON, is there one place where I can find all the Operators? I ask this because information for etcd Operator and service binding Operator are available at two different endpoints:
- /apis/etcd.database.coreos.com/v1beta2/etcdclusters
- /apis/binding.operators.coreos.com/v1alpha1/servicebindings

Am I looking at the right place?

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+# Implementation plan
+
+Note - Below I would mention `cluster` manytimes and that means both openshift and kuberenetes. If there is something specific to one or the other then I would mention it.
+
+Below is the step wise plan for implementing this feature - 
+
+## Getting the metadata
+
+### Admin Access/Access to CRD API path
+First we would consider the scenerio where the user has admin access or privileges to access the CRD API to the cluster.
```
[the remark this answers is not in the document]
girishramnani: We would implement all 3 scenarios. As it’s implemented in webconsole

docs/proposals/improved-schema-for-services.md
girishramnani: Well I don’t have a problem with either approach but what if the file is not downloaded and the user tries to create a service, it might be confusing

girishramnani:
> @girishramnani @kadel I'm probably missing something simple here. Where in the swagger AP (`http://<cluster-url>/openapi/v2`) that we're discussing here is the information about parameters available?
> 
> I'm referring to the JSON available at this endpoint and I can't find the parameters that we see in the alm-example. I'm currently trying to find parameters for etcdcluster.
> 
> Besides that, in the JSON, is there one place where I can find all the Operators? I ask this because information for etcd Operator and service binding Operator are available at two different endpoints:
> 
> * /apis/etcd.database.coreos.com/v1beta2/etcdclusters
> * /apis/binding.operators.coreos.com/v1alpha1/servicebindings
> 
> Am I looking at the right place?

The json that I would have pasted wouldn’t have etcd as I didn’t have that installed at the time.
Also that gist is half, will paste a full file soon

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
+
+# Improved schema for operator backed services
+
+This proposal is about improving the experience for a user of the operator backed services. Currently a user doesn't know exaustively which fields are available in a CR as we depend on optional metadata present in CRDs, to improve this we are considering using the metadata available from the cluster about the CRDs.
+Getting that metadata from the cluster is challenging because a normal cluster user ( plain vanilla ) doesn't have access to that metadata so we are gonna follow a 3 tiered approach
+
+- check if the user has access to the CRD, if so then use that.
+- get the `swagger.json` from the openapi endpoint provided by kubernetes. a sample can be found here https://gist.github.com/girishramnani/f29949d5cb8c6547102776437e05ac19
```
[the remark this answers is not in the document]
girishramnani: I don’t think you can read that api using kubectl

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
+
+# Improved schema for operator backed services
+
+This proposal is about improving the experience for a user of the operator backed services. Currently a user doesn't know exaustively which fields are available in a CR as we depend on optional metadata present in CRDs, to improve this we are considering using the metadata available from the cluster about the CRDs.
+Getting that metadata from the cluster is challenging because a normal cluster user ( plain vanilla ) doesn't have access to that metadata so we are gonna follow a 3 tiered approach
+
+- check if the user has access to the CRD, if so then use that.
```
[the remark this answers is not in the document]
girishramnani: If the user has permissions to that endpoint but is not admin user then they can use it

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
+
+# Improved schema for operator backed services
+
+This proposal is about improving the experience for a user of the operator backed services. Currently a user doesn't know exaustively which fields are available in a CR as we depend on optional metadata present in CRDs, to improve this we are considering using the metadata available from the cluster about the CRDs.
+Getting that metadata from the cluster is challenging because a normal cluster user ( plain vanilla ) doesn't have access to that metadata so we are gonna follow a 3 tiered approach
```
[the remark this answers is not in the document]
girishramnani: Both actually, a normal user or non admin user. From the metadata perspective both are same

docs/proposals/improved-schema-for-services.md
girishramnani: That is the case, if the user doesn’t provide any params we use alm-examples

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,409 @@
...
+
+- check if the user has access to the CRD, if so then use that.
+- get the `swagger.json` from the openapi endpoint provided by kubernetes. a sample can be found here https://gist.github.com/girishramnani/f29949d5cb8c6547102776437e05ac19
+- finally if we cannot get the openapi data as well then use the `ClusterServiceVersion` to get the parameters.
+
+This change would affect multiple service commands and the changes are described briefly below -
+- `odo catalog describe service` should include the metadata fields with description so the users can provide then when doing `odo service create`
+- `odo catalog list service` shouldn't change much other then listing CRDs per operator
```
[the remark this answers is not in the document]
girishramnani: ack, changing that

docs/proposals/improved-schema-for-services.md
girishramnani: @kadel can we conclude on the approach here?

dharmit:
> > 
> 
> The json that I would have pasted wouldn’t have etcd as I didn’t have that installed at the time.
> Also that gist is half, will paste a full file soon

I was referring to the Swagger data I got from my cluster which had etcd Opreator installed.

docs/proposals/improved-schema-for-services.md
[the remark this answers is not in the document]
dharmit: OK. Technically feasible. But doesn't seem like a scenario we would have too often. Again, I can't back up my comment with Telemetry. It's more of a hunch.

docs/proposals/improved-schema-for-services.md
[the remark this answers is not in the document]
dharmit: That seems overkill to me. If there's one approach that would work for admin as well as non-admin users, we should stick with that and improve upon that instead of trying to cover multiple scenarios. So far, it seems like swagger.json information is that thing which works for all types of users.

docs/proposals/improved-schema-for-services.md
dharmit: > Well I don’t have a problem with either approach but what if the file is not downloaded and the user tries to create a service, it might be confusing

We can print a message at the end of `odo catalog list services` saying that the information is being downloaded from the cluster and that the user should wait for a few seconds.

Besides `odo catalog list services`, don't you think we will need to validate our cache for `odo catalog describe service <operator>/<cr>` command as well?

docs/proposals/improved-schema-for-services.md
dharmit: But if the user does want to provide params, they will have to use the format like `-p spec.env[0].name FOO -p spec.env[0].value BAR`? This is a horrible UX, IMO.

girishramnani: This is only for lists and I am open to suggestions on the format

docs/proposals/improved-schema-for-services.md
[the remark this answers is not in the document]
girishramnani: I understand but this proposal is designed towards having similar experience in odo as web console.

swagger approach also has its dis-advantages and if the feedback after the swagger implementation is that it’s enough then we can drop it. But for the proposal as it is trying to provide feature parity with web console. I have kept it.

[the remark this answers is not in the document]
girishramnani: Also after the swagger approach is implemented. It’s gonna be comparatively straightforward to implement the admin approach

girishramnani:
> > The json that I would have pasted wouldn’t have etcd as I didn’t have that installed at the time.
> > Also that gist is half, will paste a full file soon
> 
> I was referring to the Swagger data I got from my cluster which had etcd Opreator installed.

Could you please share that file?

docs/proposals/improved-schema-for-services.md
girishramnani: The validation which comes to my mind is mostly about it being a valid json.
Because validating anything else might be equal to downloading the file from cluster everytime.

girishramnani: > We can print a message at the end of odo catalog list services saying that the information is being downloaded from the cluster and that the user should wait for a few seconds.

then why not just block on the command?

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,442 @@
...
+# Implementation plan
+
+Note - Below I would mention `cluster` manytimes and that means both openshift and kuberenetes. If there is something specific to one or the other then I would mention it.
+
+Below is the step wise plan for implementing this feature - 
+
+## Getting the metadata
+
+### Admin Access/Access to CRD API path
+First we would consider the scenerio where the user has admin access or privileges to access the CRD API to the cluster.
```
kadel, line 27: Just to be on the same page. When this says "**First**", It means in code flow/logic. 
As far as the implementation plan goes it will be actually done in reverse order. 
First, the `ClusterServiceVersion` approach will be implemented (already in progress)
Than Swagger approach
Then the "admin" approach will be implemented last.

docs/proposals/improved-schema-for-services.md
kadel: Let's start with simply blocking the command. Operators are not installed that often, so most of the time the cache will be used.
Let's just make sure that users know what is going on, and what they are waiting for. 
Something like "Downloading service information from the cluster ........"

docs/proposals/improved-schema-for-services.md
girishramnani: yes

docs/proposals/improved-schema-for-services.md
girishramnani: ack, adding that ask a comment

docs/proposals/improved-schema-for-services.md
girishramnani: made it clear in the proposal by adding an explicit order of implementation and changed the working at places to point out that "above steps might not be implemented yet"

docs/proposals/improved-schema-for-services.md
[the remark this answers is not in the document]
girishramnani: changed

docs/proposals/improved-schema-for-services.md
girishramnani: done

docs/proposals/improved-schema-for-services.md
girishramnani: It would be really amazing if we can confirm this approach 🙂

kadel: I can't currently think of any other solutions or this. @dharmit feel free to propose an alternative.

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,465 @@
...
+
+```
+
+
+
+## "odo service create"
+
+For any scenerio the user would provide the parameters in flat format i.e. `odo service create -p zookeeper.storage.type=ephemeral`. We would build the full parameters map using these flat key values then in case of `openapiv3schema`, validate it against the jsonschema. For `ClusterServiceVersion` we already have the flat path values so we validate them before building the parameters map.
+
+### --input-file flag
```
kadel, line 461: isn't this flag called `--from-file` now? Do you want to rename it?

girishramnani: ack? by mistake

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,465 @@
...
+```
+
+- odo also needs to have smart auto-complete which auto selects the version if the CR only has one version.
+
+
+## Using the metadata to validate the user input
+
+At this stage the user either has access to the `openAPIV3Schema` or `ClusterServiceVersion` and also the user has provided the service parameters they want to set as well. To hide the difference between these implementation from the user the `catalog describe` output would follow the same format. e.g.
+```
+- FieldPath: zookeeper.resources
```
kadel, line 169: how are arrays of objects going to be represented in this output?

lets say that there is that `first.second.envVars` is and array of env variable objects (`{name: "foo", value:"bar"}`)
How the `FieldPath` will look like, so users know that `envVars` is an array of objects and the format they need to use is `first.second.envVars[0].name=foo` `first.second.envVars[0].value=bar`)
and they don't try to do something like  `first.second.envVars.name=foo` `first.second.envVars.value=bar` or `first.second[0].envVars.name=foo` `first.second[0].envVars.value=bar`)

girishramnani:
So I worked on going through the swagger and realised some CRs dont have any information.

```
      "com.redhat.quay.v1.QuayRegistry": {
        "type": "object",
        "x-kubernetes-group-version-kind": [
          {
            "group": "quay.redhat.com",
            "kind": "QuayRegistry",
            "version": "v1"
          }
        ]
      }
```
in such cases we will as usual fallback to `x-descriptor`

docs/proposals/improved-schema-for-services.md
girishramnani: my thought was to Maybe we can show it as `secret[*].name` and explain in describe that * needs to be an index 0,1….?

kadel: > my thought was to Maybe we can show it as `first.second.envVars[*].name` and explain in describe that * needs to be an index 0,1….?

:+1: If there is enough information to do that that would be great!

kadel:
There is still a lot of details that need to be figured out. But those are going to be easier to figure out once we start implementing this.
From a high level, this looks good.
/approve

docs/proposals/improved-schema-for-services.md
girishramnani: yeah we have information on what is an array from swagger

girishramnani: added conversion detail in the proposal

docs/proposals/improved-schema-for-services.md
dharmit: An alternative could be to not use the array formatting and check in odo code if the parameter that the user's trying to set is indeed available. odo would have to do some heavy lifting in that case in terms of validating stuff.

But this is implementation detail, and it's not relevant for proposal.

docs/proposals/improved-schema-for-services.md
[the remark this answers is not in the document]
dharmit: Makes sense.

docs/proposals/improved-schema-for-services.md
```diff
@@ -0,0 +1,472 @@
...
+Sample output - 
+https://gist.github.com/girishramnani/cbb4400e463efe89c13f1386e0788793
+
+We care about the `openAPIV3Schema` as that would be used to build the CRD struct. 
+
+### Kuberenetes cluster Swagger has the schema
+
+If the user doesn't have CRD access or that portion is not implemented yet then we fetch try to fetch the same `openAPIV3Schema` from the cluster's `swagger.json` from the endpoint `<cluster-url>/api/kubernetes/openapi/v2`. This is a very large document as it holds all the definitions present on the cluster.
+
+So this needs to be cached and refreshed whenever a new operator is installed.
```
feloy, line 45: I'm not very comfortable with managing an automatic cache (which, without a serious protocol between the two parts, is doomed to be inefficient).

What about a specific and explicit odo command, that the user could run to fetch the param information for one or more services?

User story:

1. `odo catalog list services` indicates the params or number of params if known, or if not known, that the user can fetch them.
2. fetch the params for a specific service 

```
$ odo catalog list services
Services available through Operators
NAME                                             CRDs                                                          Params
etcdoperator.v0.9.4-clusterwide     EtcdCluster, EtcdBackup, EtcdRestore       (to fetch)
service-binding-operator.v0.7.1     ServiceBinding                                             3 parameters
To fetch params for a service, you can run `odo service fetch-params <service-name>


$ odo service fetch-params etcdoperator.v0.9.4-clusterwide/EtcdCluster
...
```


Also, only the params information could be stored locally (and not the complete swagger/CRD info), independently of the source of the params (ClusterServiceVersion, swagger, CRD), so the use of this data by `odo service create` is generic.

The user could detect which method for fetching the data is available, or give the user the choice to select one.

kadel: > fetch the params for a specific service

I'm not sure about this. As a user I would be pretty annoyed that I have to run extra command to get information that is essentially required for creating services.
Why `odo catalog list services` couldn't do that automatically?



> Also, only the params information could be stored locally (and not the complete swagger/CRD info), independently of the source of the params (ClusterServiceVersion, swagger, CRD), so the use of this data by `odo service create` is generic.

I like the idea of caching just params for all sources. This is good.

> The user could detect which method for fetching the data is available, or give the user the choice to select one.

I think that this is exposing implementation to users. The user doesn't care where the data comes from. Especially when it can be a fairly complicated decision.

Issue #4639, opened 2021-04-19
Odo integration tests should use staging devfile registry (registry.stage.devfile.io)


johnmcollier:
With https://github.com/openshift/odo/pull/4525 now merged, Odo now supports the new community devfile registry at https://registry.devfile.io, and uses it by default.

While most Odo integration tests don't directly use the registry for testing, there are a few other integration tests that do:
- `odo catalog` tests
- `odo create --registry` tests
- `odo registry` tests

A few of us on devfile services were chatting and think that for the integration tests in Odo that _do_ use the registry, it might be a good idea to use the staging devfile registry (https://registry.stage.devfile.io) instead of production for a couple reasons:
- It'll reduce unnecessary load on the production registry, especially when there's multiple test runs running concurrently
- The staging and production registries are largely identical, the only difference is _when_ they are updated
    - Staging is updated with each commit to master in [devfile/registry](https://github.com/devfile/registry), production is updated manually each Wednesday
- Telemetry will eventually be enabled on the production devfile registry and we do not want the integration tests polluting the data we get back from the telemetry.

Let me know if this sounds like a reasonable thing to do, or if there's any concerns about this approach.

johnmcollier:
I just want to clarify too, this is **not** something that necessarily needs to be done urgently or soon - we're still a ways away from implementing telemetry on the devfile registry, and the extra load that the tests are creating on the registry is useful for us as we monitor load in this transition period.

But it is something that will need to be done _once_ we implement telemetry

josephca:
Currently, the following tests are using the staging registry - https://github.com/openshift/odo/pull/4679
- odo catalog test 
- odo registry tests

Thus, `odo create --registry tests` is remained to use the staging registry for the test.

valaparthvi:
This issue is still in progress w.r.t. PR #4775.

PR #4752, opened 2021-05-26
Use staging devfile registry for integration tests

josephca:
**What type of PR is this?**
/kind cleanup

Use staging devfile registry for integration test with devfile. Using a production registry would result in unnecessary overload.

Fixes https://github.com/openshift/odo/issues/4639

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [ ] Update changelog

### DIFF
--- a/tests/integration/devfile/devfile_suite_test.go
+++ b/tests/integration/devfile/devfile_suite_test.go
@@ -3,9 +3,18 @@ package devfile
 import (
 	"testing"
 
+	. "github.com/onsi/ginkgo"
 	"github.com/openshift/odo/tests/helper"
 )
 
 func TestDevfiles(t *testing.T) {
 	helper.RunTestSpecs(t, "Devfile Suite")
 }
+
+var _ = BeforeSuite(func() {
+	const registryName string = "DefaultDevfileRegistry"
+	const addRegistryURL string = "https://registry.stage.devfile.io"
+
+	// Use staging OCI-based registry for tests to avoid a potential overload
+	helper.CmdShouldPass("odo", "registry", "update", registryName, addRegistryURL)
+})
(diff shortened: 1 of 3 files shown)


josephca:
@mohammedzee1000 and @rnapoles-rh: please review. Thanks,

tests/integration/devfile/devfile_suite_test.go
```diff
@@ -3,9 +3,18 @@ package devfile
...
 func TestDevfiles(t *testing.T) {
 	helper.RunTestSpecs(t, "Devfile Suite")
 }
+
+var _ = BeforeSuite(func() {
+	const registryName string = "DefaultDevfileRegistry"
+	const addRegistryURL string = "https://registry.stage.devfile.io"
+
+	// Use staging OCI-based registry for tests to avoid a potential overload
+	helper.CmdShouldPass("odo", "registry", "update", registryName, addRegistryURL)
```
girishramnani, line 19: @kadel Trying to understand if this would be a good idea,
On one side we would detect errors early before the registry changes land into production?
but then anytime a small incompatible change is pushed to stage, it would break odo's test.

josephca: @girishramnani Copying the second bullet point of the associated issue here: 

- The staging and production registries are largely identical, the only difference is when they are updated

    - Staging is updated with each commit to master in devfile/registry, production is updated manually each Wednesday
   
IMHO, we don't expect a big difference between staging vs. production registries. Early detection of a potential break would give more benefit than finding an error on the production registry that requires fixing anyway.

kadel: I totally agree with @josephca.

> but then anytime a small incompatible change is pushed to stage, it would break odo's test.

This is exactly the reason why we should test against staging.  To get an early warning. 
If breaking change gets merged into "stage" we want to know about it before it gets into "prod" and fix it before it gets to "prod"

girishramnani: Cool, merging

PR #4775, opened 2021-06-03
Use staging registry for devfile debug suite test

### LINKED ISSUE #4639: Odo integration tests should use staging devfile registry (registry.stage.devfile.io)

josephca:
**What type of PR is this?**
 /kind cleanup

Use staging devfile registry for devfile debug suite test similar to https://github.com/openshift/odo/pull/4752

Fixes #4639

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [ ] Update changelog

### DIFF
--- a/tests/integration/devfile/debug/debug_suite_test.go
+++ b/tests/integration/devfile/debug/debug_suite_test.go
@@ -3,9 +3,18 @@ package debug
 import (
 	"testing"
 
+	. "github.com/onsi/ginkgo"
 	"github.com/openshift/odo/tests/helper"
 )
 
 func TestDebug(t *testing.T) {
 	helper.RunTestSpecs(t, "Devfile Debug Suite")
 }
+
+var _ = BeforeSuite(func() {
+	const registryName string = "DefaultDevfileRegistry"
+	const addRegistryURL string = "https://registry.stage.devfile.io"
+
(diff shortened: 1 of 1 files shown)


valaparthvi (approved):
Based on the comments from PR #4752, these changes /lgtm

Issue #4615, opened 2021-04-14
odo env set DebugPort doesn't work for converted devfile 


girishramnani:
/kind bug

## What versions of software are you using?

**Operating System:** MacOS

**Output of `odo version`:**

## How did you run odo exactly?
The reason this is not working for an s2i converted devfile because there is not dedicated `debug` command present for s2i converted devfiles and hence the `DebugPort` in env is not respected.

girishramnani:
added a workaround for now https://github.com/openshift/odo/pull/4518#discussion_r613205859

girishramnani:
uncomment the workaround once this issue is fixed

dharmit:
Workaround for this is documented in #4683.

girishramnani:
The ideal fix for this should be to have `debug` command on converted devfile.

girishramnani:
One fix would be to set the `DEBUG_PORT` env variable when someone sets `DebugPort` in localEnvInfo?

girishramnani:
@kadel thoughts on the possible fix^

PR #4785, opened 2021-06-08
change debugport in container if its different from env_info

girishramnani:
**What type of PR is this?**
/kind bug

change debugport in container if its different from env_info, also added debug command for converted devfiles.

Fixes https://github.com/openshift/odo/issues/4615

**PR acceptance criteria**:

- [x] Unit test 

- [X] Integration test 

- [ ] Documentation 

- [x] Update changelog


pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -209,6 +209,17 @@ func UpdateContainersWithSupervisord(devfileObj devfileParser.DevfileObj, contai
...
 						Value: strconv.Itoa(devfileDebugPort),
 					})
+			} else {
+				// if the env has the debug port but if its value is different then we change it
+				// this way the value of env info is given more importance then whats there on the devfile
+				for i, envVar := range container.Env {
+					if envVar.Name == adaptersCommon.EnvDebugPort {
+						if envVar.ValueFrom == nil && (envVar.Value != strconv.Itoa(devfileDebugPort)) {
+							container.Env[i].Value = strconv.Itoa(devfileDebugPort)
+							break
```
mik-dass, line 219: Please add a test case here https://github.com/openshift/odo/blob/ca5ebd60331eaa32bc5667ca59ac2e010707c369/pkg/devfile/adapters/kubernetes/utils/utils_test.go#L161

girishramnani: added

mik-dass (approved):
/lgtm

girishramnani:
@mik-dass the lgtmer has to approve also, please can approve it as well.

mik-dass:
> @mik-dass the lgtmer has to approve also, please can approve it as well.

I think, in a recent discussion, it was decided to have two reviewers on each PR.

girishramnani:
> > @mik-dass the lgtmer has to approve also, please can approve it as well.
> 
> I think, in a recent discussion, it was decided to have two reviewers on each PR.

Ack, thank you for reminding

dharmit:
I followed the below steps:

```sh
$ git clone https://github.com/odo-devfiles/nodejs-ex
$ cd nodejs-ex
$ odo create --s2i nodejs node-s2i

$ odo env set DebugPort 8888
$ cat .odo/env/env.yaml
ComponentSettings:
  Name: node-s2i
  Project: myproject
  AppName: app
  DebugPort: 8888
  RunMode: run

$ odo push

$ odo exec -- env | grep DEBUG
DEBUG_PORT=5858

$ grep DEBUG_PORT devfile.yaml
    - name: DEBUG_PORT
      value: "5858"
```

There are two concerns I have:
- Pushed component's env var still shows 5858 as the `DebugPort`
- devfile contains 5858 as the value for `DEBUG_PORT`

`DebugPort` and `DEBUG_PORT` are the same things, right? Or am I looking at this the wrong way?

Am I testing this the right way?

dharmit:
> `DebugPort` and `DEBUG_PORT` are the same things, right? Or am I looking at this the wrong way?

https://github.com/openshift/odo/issues/4615#issuecomment-855879426 gives me an impression that these two are different things. 😕

girishramnani:
> @girishramnani I followed the below steps:
> 
> ```shell
> $ git clone https://github.com/odo-devfiles/nodejs-ex
> $ cd nodejs-ex
> $ odo create --s2i nodejs node-s2i
> 
> $ odo env set DebugPort 8888
> $ cat .odo/env/env.yaml
> ComponentSettings:
>   Name: node-s2i
>   Project: myproject
>   AppName: app
>   DebugPort: 8888
>   RunMode: run
> 
> $ odo push
> 
> $ odo exec -- env | grep DEBUG
> DEBUG_PORT=5858
> 
> $ grep DEBUG_PORT devfile.yaml
>     - name: DEBUG_PORT
>       value: "5858"
> ```
> 
> There are two concerns I have:
> 
> * Pushed component's env var still shows 5858 as the `DebugPort`
> * devfile contains 5858 as the value for `DEBUG_PORT`
> 
> `DebugPort` and `DEBUG_PORT` are the same things, right? Or am I looking at this the wrong way?
> 
> Am I testing this the right way?
> 
> cc @kadel

@dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.

You can still see the debug port in the env because its part of the original env of the s2i and gets passed to the devfile.

Please follow the same steps but with `odo push --debug` and you will see that the debug port gets changed in the pod.

dharmit:
> @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.
> You can still see the debug port in the env because its part of the original env of the s2i and gets passed to the devfile.
> Please follow the same steps but with `odo push --debug` and you will see that the debug port gets changed in the pod.

That works. But should we also change the `DEBUG_PORT` value in devfile.yaml when user does `odo env set DebugPort`? I am asking this because there are different values (and even different naming style - `DEBUG_PORT` vs. `DebugPort`) for debug port in devfile vs. that in env.yaml. 

I think env.yaml is for environment specific information and it can have values different from devfile. But I would like to be sure.

dharmit:
> @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.

From UX perspective, how do we convey this to the user that they must do `odo push --debug` for this env var value to actually take effect in the component? Or should the user be aware of it already since `odo push --debug` is the only way to start the component in debug mode?

girishramnani:
> > @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.
> 
> From UX perspective, how do we convey this to the user that they must do `odo push --debug` for this env var value to actually take effect in the component? Or should the user be aware of it already since `odo push --debug` is the only way to start the component in debug mode?

This been a standard set in devfile already since a while now

girishramnani:
> > @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.
> > You can still see the debug port in the env because its part of the original env of the s2i and gets passed to the devfile.
> > Please follow the same steps but with `odo push --debug` and you will see that the debug port gets changed in the pod.
> 
> That works. But should we also change the `DEBUG_PORT` value in devfile.yaml when user does `odo env set DebugPort`? I am asking this because there are different values (and even different naming style - `DEBUG_PORT` vs. `DebugPort`) for debug port in devfile vs. that in env.yaml.
> 
> I think env.yaml is for environment specific information and it can have values different from devfile. But I would like to be sure.

Technically you can but changing debug port from what it is using `odo env set DebugPort` probably feels like it should affect env.yml and doesn’t get persisted to devfile.

its feels more of a tool to override the value in devfile for your local setup only?

dharmit:
> @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.

Not sure if I'm nitpicking here. Let me know if so.

When I do `odo push --debug`, the debugport env does reflect correctly in the pod. But devfile still shows 5858. Based on above comment, I expected it to change in devfile as well. Or is that a wrong expectation?

dharmit:
> > > @dharmit the debug port only gets changed in devfile when you provide `odo push --debug`.
> > 
> > 
> > From UX perspective, how do we convey this to the user that they must do `odo push --debug` for this env var value to actually take effect in the component? Or should the user be aware of it already since `odo push --debug` is the only way to start the component in debug mode?
> 
> This been a standard set in devfile already since a while now

OK. I find it unintuitive that user isn't notified about using `odo push --debug`. I'm not sure if this is something we should cover in the docs or in the cli messaging.

Just to be clear, I don't mean we address it in this PR. We should do it separately.

dharmit:
> > 
> 
> Technically you can but changing debug port from what it is using `odo env set DebugPort` probably feels like it should affect env.yml and doesn’t get persisted to devfile.

Agree. `odo env` commands shouldn't touch anything in devfile. That would be an even bigger problem. 👍🏾

dharmit:
/approve
/lgtm

Doing both since lgtm was removed by rebase - https://github.com/openshift/odo/pull/4785#event-4882685309

/override ci/prow/psi-unit-test-mac

PR #4795, opened 2021-06-10
Fixing GOPATH and GOBIN in minikube and minishift scripts

mohammedzee1000:

### DIFF
--- a/scripts/minikube-minishift-setup-env.sh
+++ b/scripts/minikube-minishift-setup-env.sh
@@ -7,10 +7,14 @@ shout() {
 }
 
 # Create a bin directory whereever script runs. This will be where all binaries that need to be in PATH will reside.
-mkdir bin artifacts
+mkdir -p go/bin artifacts
 # Change the default location of go's bin directory (without affecting GOPATH). This is where compiled binaries will end up by default
 # for eg go get ginkgo later on will produce ginkgo binary in GOBIN
-export GOBIN="`pwd`/bin"
+export GOPATH="`pwd`/go"
+export GOBIN="$GOPATH/bin"
+
+# Set kubeconfig to current dir. This ensures no clashes with other test runs
+export KUBECONFIG="`pwd`/config"
 export ARTIFACTS_DIR="`pwd`/artifacts"
 export CUSTOM_HOMEDIR=$ARTIFACT_DIR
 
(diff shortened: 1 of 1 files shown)


scripts/minikube-minishift-setup-env.sh
```diff
@@ -7,10 +7,11 @@ shout() {
 }
 
 # Create a bin directory whereever script runs. This will be where all binaries that need to be in PATH will reside.
-mkdir bin artifacts
+mkdir -p go/bin artifacts
```
kadel, line 10: we should make sure that the script can be executed from any directory.
If `go/bin` and `artifacts` are supposed to create in home it should use `$HOME`

scripts/minikube-minishift-setup-env.sh
```diff
@@ -7,10 +7,11 @@ shout() {
...
 # Change the default location of go's bin directory (without affecting GOPATH). This is where compiled binaries will end up by default
 # for eg go get ginkgo later on will produce ginkgo binary in GOBIN
-export GOBIN="`pwd`/bin"
+export GOPATH="`pwd`/go"
```
kadel, line 13: similarly to before, this depends on the current directory, we should try to avoid that

mohammedzee1000: Uniqie dir will be ensured by ci-firewall with https://github.com/mohammedzee1000/ci-firewall/pull/24

scripts/minikube-minishift-setup-env.sh
mohammedzee1000: no, not in home. It will be same as not creating at all. We want unique GOPATH and GOBIn for each run

mohammedzee1000:
Prow issues
/retest

scripts/minikube-minishift-setup-env.sh
feloy: I think it would be safer to have a different $HOME for each script running, as lots of tools are relying on this variable to set files.
If necessary, we could have a bunch of users available for the different scripts, or create these users on the fly.

scripts/minikube-minishift-setup-env.sh
```diff
@@ -7,15 +7,16 @@ shout() {
...
+export GOBIN="$GOPATH/bin"
 
 # Set kubeconfig to current dir. This ensures no clashes with other test runs
 export KUBECONFIG="`pwd`/config"
 export ARTIFACTS_DIR="`pwd`/artifacts"
-export CUSTOM_HOMEDIR=$ARTIFACT_DIR
+export HOME="`pwd`/home"
```
feloy, line 19: Why not HOME=`pwd` ?

mohammedzee1000: Just want to make sure there is absolutely no chance of overlap. GOPATH, GOBIN and HOME

feloy (approved):

PR #4770, opened 2021-06-03
Refactor OperatorHub integration tests

feloy:
**What type of PR is this?**

/kind code-refactoring

**What does this PR do / why we need it**:

This PR refactors the code of the Operator Hub integration tests.

- `Context` describes a state of the cluster, and `JustBeforeEach` checks/wait that the state is ready
- `When` describes the execution of a command or other action **with side-effect** by the user, `JustBeforeEach` contains the execution of the command, `JustAfterEach` contains the counterpart of this command if necessary
- `It` checks the expectations after the last command executed. For readibility, `It` can contain commands executed by the user if these commands have no side-effect (commands failing, lists, describes, etc)

To add a test in this flow, a developer has to follow the Context/When tree to place its tests at the moment in the flow where (s)he wants to execute the test.

A plugin is available in VSCode to help browse the tree of Context/When/It (you will need ginkgo 1.15):

ginkgo-test-explorer

The output:

ginkgo-plugin

**Which issue(s) this PR fixes**:

Fixes #?

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] Update changelog

- [ ] I have read the [test guidelines](https://github.com/openshift/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

PR #4806, opened 2021-06-15
app tests refactored

feloy:
/kind code-refactoring

Demo to refactor app integration tests

### DIFF
--- a/tests/integration/cmd_app_test.go
+++ b/tests/integration/cmd_app_test.go
@@ -11,10 +11,6 @@ import (
 
 var _ = Describe("odo app command tests", func() {
 	var commonVar helper.CommonVar
-	appName := "app"
-	cmpName := "nodejs"
-	mountPath := "/data"
-	size := "1Gi"
 
 	// This is run before every Spec (It)
 	var _ = BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


tests/integration/cmd_app_test.go
```diff
@@ -27,102 +23,137 @@ var _ = Describe("odo app command tests", func() {
...
+
+			It("should fail", func() {
+				By("describing app with app parameter", func() {
+					helper.CmdShouldFail("odo", "app", "describe", "--project", commonVar.Project)
+				})
+
+				By("deleting an app without app parameter", func() {
+					helper.CmdShouldFail("odo", "app", "delete", "-f", "--project", commonVar.Project)
+				})
+			})
```
valaparthvi, line 81: Hi @feloy, this part does not make much sense to me. If I was to run a test, as far as I know, I would not see the By part, so the test output would be something like -  `odo app command tests on a fresh new project [It] should fail`, which would not exactly make sense. I understand this approach would make the tests faster, but then I also feel that it contradicts with the approach you implemented in #4770 which made tests easier to read and extend.

feloy: Good catch, you are right, I thank the By would be displayed.

I'm trying to find a compromise between the approach of #4770 and the duration of tests. Using By can reduce the execution time by running all the "BeforeEach"s only once, but it contradicts the spirit of the approach, to have only one Expectation per test.

I will make separate "It"'s instead of using "By"s, and see how longe are the tests. Running the tests on 4 nodes (or more) would however accelerate the tests

tests/integration/cmd_app_test.go
```diff
@@ -27,102 +23,128 @@ var _ = Describe("odo app command tests", func() {
...
+		appHelp := helper.Cmd("odo", "app", "-h").ShouldPass().Out()
+		// Trimmed the end of the message string to make it compatible across clusters
+		Expect(appHelp).To(ContainSubstring("Performs application operations related to"))
 	})
 
-	Context("when running app delete, describe and list command on fresh cluster", func() {
-		It("should error out display the help", func() {
+	Context("on a fresh new project", func() {
+
+		BeforeEach(func() {
```
prietyc123, line 34: Why not `JustBeforeEach`?

prietyc123 (changes requested):

tests/integration/cmd_app_test.go
```diff
@@ -27,102 +23,128 @@ var _ = Describe("odo app command tests", func() {
...
-			helper.Cmd("odo", "app", "describe", "--project", commonVar.Project).ShouldFail()
-			helper.Cmd("odo", "app", "delete", "-f", "--project", commonVar.Project).ShouldFail()
-		})
-	})
+		appName := "app"
+		cmpName := "nodejs"
 
-	Context("when running app command app parameter in directory that doesn't contain .odo config directory", func() {
-		It("should successfully execute list, describe and delete along with machine readable output", func() {
+		BeforeEach(func() {
```
prietyc123, line 60: https://github.com/openshift/odo/pull/4806/files#r657657242

tests/integration/cmd_app_test.go
feloy: What would be the advantage of using `JustBeforeEach` here?

We have to use `JustBeforeEach` only when really necessary, as the ginkgo doc explains:

```
You can have multiple JustBeforeEaches at different levels of nesting. Ginkgo will first run all the BeforeEaches from the outside in, then it will run the JustBeforeEaches from the outside in. While powerful, this can lead to confusing test suites – so use nested JustBeforeEaches judiciously.

Some parting words: JustBeforeEach is a powerful tool that can be easily abused. Use it well.
```

prietyc123 (changes requested):

tests/integration/cmd_app_test.go
```diff
@@ -27,102 +23,128 @@ var _ = Describe("odo app command tests", func() {
...
 
-		})
+			mountPath := "/data"
+			size := "1Gi"
+
+			BeforeEach(func() {
+				helper.Cmd("odo", "storage", "create", "storage-one", "--context", commonVar.Context, "--path", mountPath, "--size", size).ShouldPass()
+			})
 
+			When("odo push is executed", func() {
```
anandrkskd, line 139: @feloy Can we merge this two `when`? I guess it will decrease the level.

tests/integration/cmd_app_test.go
```diff
@@ -27,102 +23,128 @@ var _ = Describe("odo app command tests", func() {
...
+			It("should describe specific app in JSON format", func() {
+				actualDesAppJSON := helper.Cmd("odo", "app", "describe", appName, "--project", commonVar.Project, "-o", "json").ShouldPass().Out()
+				valuesDes := gjson.GetMany(actualDesAppJSON, "kind", "metadata.name", "metadata.namespace")
+				expectedDes := []string{"Application", appName, commonVar.Project}
+				Expect(helper.GjsonMatcher(valuesDes, expectedDes)).To(Equal(true))
+			})
+
+			When("changing to context directory", func() {
+
+				BeforeEach(func() {
```
prietyc123, line 104: We have defined context in `BeforeEach` https://github.com/openshift/odo/pull/4806/files#diff-ae5d2699e830fc0052b86751a2d5816e2f6381a8d27e71c28bb56e8c44651b26R17 and then modifying the content of it now. AFAIK in such scenario we go for `JustBeforeEach`. Correct me if I am wrong.

prietyc123 (changes requested):

tests/integration/cmd_app_test.go
feloy: Yes, done

anandrkskd: Thanks :)

tests/integration/cmd_app_test.go
feloy: This is not what I understand about `JustBeforeEach`.

`BeforeEach`es are exectued one after the other, for example:

```
When("...", func() {
  BeforeEach(func() {
     fmt.Printf("Executed #1")
  }

  When("...", func() {
    BeforeEach(func() {
      fmt.Printf("Executed #2")
    }
  })
})
```

When using `JustBeforeEach`, they will be executed after all `BeforeEach`es:

```
When("...", func() {
  BeforeEach(func() {
     fmt.Printf("Executed #1")
  }

  JustBeforeEach(func() {
     fmt.Printf("Executed #3")
  }

  When("...", func() {
    BeforeEach(func() {
      fmt.Printf("Executed #2")
    }
  })
})
```

tests/integration/cmd_app_test.go
prietyc123: AFAIU spec is the final level in the hierarchy of ginkgo test. For example Describe -> context -> when -> It (Spec). Similarly as part of teardown hierarchy is concerned it will be something like BeforeSuite -> BeforeEach (Describe, context and when level)-> JustBeforeEach (Spec level). The only advantage i can see here is a hierarchical way of applying teardown steps.

feloy: What you recommend is using `JustBeforeEach` for the last level of hierarchy before `It`, right?

```
Describe("odo app command tests", func() {
	var _ = BeforeEach(func() {
		// #1
	})
	When("creating a new component", func() {
		BeforeEach(func() {
			// #2
		})
		When("odo push is executed", func() {
			JustBeforeEach(func() {
				// #3
			})
			It("...", func() {
				// ...
			})

		})
	})
})
```

1. I cannot see an advantage of using JustBeforeEach instead of BeforeEach, as in both cases, it will be executed in #3.

2. The drawback I can see if if you want to add some tests later, like this, you will be in risk to finish with several `JustBeforeEach` at different levels, which is complex to manage (as explained in the ginkgo doc):

```
Describe("odo app command tests", func() {
	var _ = BeforeEach(func() {
		// #1
	})
	When("creating a new component", func() {
		BeforeEach(func() {
			// #2
		})
		When("odo push is executed", func() {
			JustBeforeEach(func() {
				// #3
			})
			It("...", func() {
				// ...
			})

			When("...", func() {
				JustBeforeEach(func() {
					// #4
				})
				It("...", func() {
					// ...
				})
			})
		})
	})
})
```

anandrkskd:
Windows test failed due to timeout. looking into it.

/test psi-unit-test-windows

tests/integration/cmd_app_test.go
prietyc123: Thanks for the details. We will document those in out test guidelines 👍

PR #4807, opened 2021-06-15
Remove nesting levels in operatorhub integration tests

feloy:
**What type of PR is this?**

/kind code-refactoring

- Remove two nesting levels in operatorhub integration tests
- Change all JustBeforeEach/JustAfterEach with BeforeEach/AfterEach

### DIFF
--- a/tests/integration/operatorhub/cmd_service_test.go
+++ b/tests/integration/operatorhub/cmd_service_test.go
@@ -32,7 +32,7 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
 
 	Context("Operators are installed in the cluster", func() {
 
-		JustBeforeEach(func() {
+		BeforeEach(func() {
 			// wait till odo can see that all operators installed by setup script in the namespace
 			odoArgs := []string{"catalog", "list", "services"}
 			operators := []string{"etcdoperator", "service-binding-operator"}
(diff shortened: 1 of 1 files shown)


tests/integration/operatorhub/cmd_service_test.go
```diff
@@ -68,33 +68,33 @@ var _ = Describe("odo service command tests for OperatorHub", func() {
...
 
-					JustAfterEach(func() {
+					AfterEach(func() {
 						helper.Cmd("odo", "service", "delete", fmt.Sprintf("Database/%s", operandName), "-f").ShouldPass().Out()
 						helper.Cmd("odo", "push").ShouldPass().Out()
 					})
 
 					When("odo push is executed", func() {
-						JustBeforeEach(func() {
+						BeforeEach(func() {
```
anandrkskd, line 97: @feloy I see multiple nested `when` (from L55-L115) but they are testing only two `IT` spec which are at the same level, are these in this way to avoid extra work to add new test case if required in future?

anandrkskd (commented):
Just one question, otherwise lgtm!

tests/integration/operatorhub/cmd_service_test.go
feloy: Girish have written this part, but I think it is interesting to keep the different steps as nested Whens, so we can see more clearly what are the different steps executed by the user, and also it will be easier to add more tests between steps in the future.

Generally, I put an odo command and the following `odo push` in the same When if I can't see any `It` that could be added between the odo command and `odo push`.

anandrkskd: understood, and make sense! 
Thanks @feloy

PR #4846, opened 2021-06-24
Run operator hub tests in parallel

feloy:
**What type of PR is this?**

/kind code-refactoring

Runs the Operator Hub integration tests in parallel and not serially, as it is not necessary anymore.


girishramnani:
Small enough change lgtming it
/lgtm

PR #4863, opened 2021-06-30
Refactor component tests

feloy:
**What type of PR is this?**
/kind code-refactoring

Refactoring of s2i component integration tests


dharmit (approved):
/approve
/lgtm

(doing both since PR is refactoring tests and CI's green)

Issue #4894, opened 2021-07-08
Add "Getting Started" docs to the docusaurus based docs site


dharmit:
## What mistake did you find / what is missing in the documentation?
N/A

## What is the relevance of it?
This is a child issue of spawned from https://github.com/openshift/odo/issues/4664#issuecomment-875511836

## Acceptance Criteria
"Getting Started" section gives users an overview about odo:
- [ ] Quickstart (@kadel you need to help us here with [Springboot example](https://github.com/spring-projects/spring-petclinic))(#4922)
- [x] Basics (#4914)
- [ ] Cluster Setup (#4973)
  - [ ] ~Development Cluster (this doc clarifies that odo docs help setup development clusters only.)~
  - [x] Kubernetes
  - [x] OpenShift
- [x] Installation (#4923)
- [ ] ~Import applications (talk about how existing git repo can be deployed to k8s using odo)~
- [x] Configuration (talk about odo preference file) (#4921)

/kind documentation
/kind task

dharmit:
Opened separate issue for Quickstart. Rest of the things are done by @valaparthvi 

/close

PR #4914, opened 2021-07-15
Add Getting Started > Basics documentation

valaparthvi:
**What type of PR is this?**
/kind documentation

This PR adds content to Getting Started > Basics.
_Note_: Some of the content might not be true and will need a lot of refining.

Fixes part of #4894

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [x] Documentation 

- [ ] Update changelog

```shell
$ cd website/docs
$ npm install # this command installs dependencies required to create the website
$ npm run start
```


website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
 
 ### Concepts
 
-#### Component
+* Component
+  A component is a microservice. Each component can be run and deployed separately. Examples of components: Warehouse API Backend, Inventory API, Web Frontend, Payment Backend
+  > A component in odo is an equivalent of a deployment or deploymentconfig in Kubernetes. Creating a new component in odo will create a deployment in Kubernetes or for git and binary based component creation, it will create deploymentconfig.
```
dharmit, line 13: 1. Why is this quoted with a `>`? Same question for everywhere else in the doc.
2. IMO, odo component is not an equivalent of Deployment or DeploymentConfig. By default, odo creates a bunch of Kubernetes resources for a component:
    - Deployment (which creates a ReplicaSet and Pods)
    - Service (a Kubernetes Service; not odo service)
    - Storage (emptyDir or PVC depending on user's preference setting)
    - probably more that I might have missed.
3. We don't need to go into this level of detail for git/binary components. Also, it doesn't create just DC but other things that I mentioned above.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
 * Application
+  An application consists of multiple components which may span over multiple projects, and work individually to build the entire application. Examples of applications: e-Shop, Hotel Reservation System, Online Booking
```
dharmit, line 15: > An application consists of multiple components which may span over multiple projects

This is not true for odo. odo doesn't cover things across multiple namespaces.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
+  > An application can be considered as an equivalent of labels in Kubernetes that help in grouping a set of resources.
 * Project
+  A project is your source code, tests, and libraries organized in a separate single unit.
```
dharmit, line 18: This sounds more like a definition of what a component directory (a.ka. context) is on the dev's local system.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
+  > A project in odo is an equivalent of a namespace in Kubernetes. Creating a project in odo will create a namespace in Kubernetes with the same name.
 * Context
+  A context is the directory where the source code, tests, and libraries for your component resides. A single context can only contain a single component.
 * URL
+  A URL exposes your component to be accessed outside the cluster.
+  > A URL is an equivalent of a service in Kubernetes.
```
dharmit, line 24: NO! It's not an equivalent of a Service in k8s! It's equivalent to Ingresses in k8s and Routes in OCP.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
 * Storage
+  A storage volume is [PVC](https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim) which is a way for you to "claim" persistent storage without knowing the details of the environment. Storage volume can persist data across restarts and rebuilds of a component.
```
dharmit, line 26: Using k8s terminology to explain odo concept is not a good idea. It just means we fail to explain (or even understand ourselves!) the abstractions we created to make k8s terminology not required for the user.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
 * Service
+  A service is another microservice, or a Kubernetes Custom Resource that your component connects to or depends on. Example of services: MariaDB, MySQL
+  > A service created with odo creates multiple Kubernetes resources to establish a proper linking between the source component and target microservice, or 
 * Devfile
+  A portable YAML file responsible for your entire reproducible development environment. See [Devfile](../architecture/devfile.md) to know more about devfile.
+  > A devfile can be considered similar to a manifest file in Kubernetes, but there is no equivalent of it in Kubernetes.
+  
 
 ### Component vs. Application
+A component is a microservice, while an application is a group of microservices. To make a microservice belong to an application, all the resources belonging to the microservice are assigned a label.
```
dharmit, line 37:
```suggestion
A component is akin to a microservice, while an application is a group of components.
```
> To make a microservice belong to an application, all the resources belonging to the microservice are assigned a label.

Removed this as it is more about how it's done under the hood, and we don't need to expose the implementation detail to the user/reader.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
 
-### odo services vs. Kubernetes services
\ No newline at end of file
+### odo services vs. Kubernetes services
+odo service helps in connecting one microservice to another, while a Kubernetes service helps in making a microservice accessible outside the cluster.
```
dharmit, line 40: NO! It's the link that helps in connecting stuff. 

odo service is a service created from an Operator. k8s service definition looks good.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
+Creating an odo service will create a set of resources to establish a successful connection between the two microservices.
```
dharmit, line 41: This is totally wrong. It's a link that helps setup connectivity between microservices. We don't need to have this under odo service vs k8s service.

dharmit (changes requested):

website/docs/getting-started/basics.md
feloy: A kubernetes service is mainly an abstraction that serves to expose a set of Pods **in a stable way** (the service will keep the same IP during its life, when Pods can be scaled and restarted, each instance having a new and separate IP).

Services can expose the microservice at different levels depending on its type:
- clusterIP type on the cluster for inter-microservices communication, 
- node type to access from outside the Kubernetes cluster through the nodes
- loadBalancerType to access from outside the provider cluster, through some load balancer

website/docs/getting-started/basics.md
feloy: I understand that the quoted part explain how the concepts are implemented in Kubernetes.

But it is probably not a good place to talk about these implementations. This concept page should help the reader make a mental representation of the objects handled by odo, without other distraction.

valaparthvi: >Why is this quoted with a >? Same question for everywhere else in the doc.

1. I figured we would put the Kubernetes part in a quoted text like [Philippe has done in his blog](https://feloy.medium.com/odo-a-tool-to-simplify-development-on-kubernetes-ce0df34313b7).
I can add a note about it in the beginning.

>IMO, odo component is not an equivalent of Deployment or DeploymentConfig. By default, odo creates a bunch of Kubernetes resources for a component:

    Deployment (which creates a ReplicaSet and Pods)
    Service (a Kubernetes Service; not odo service)
    Storage (emptyDir or PVC depending on user's preference setting)
    probably more that I might have missed.

2. Okay, I can change that.

>We don't need to go into this level of detail for git/binary components. Also, it doesn't create just DC but other things that I mentioned above.

3. Noted.

website/docs/getting-started/basics.md
feloy: Without naming the underlying Kubernetes `namespace` resource, you can explain the motivations of using separate namespaces:
- a scope for names
- divide cluster resources between users (resources quotas, RBAC are based on namespaces)

https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/

website/docs/getting-started/basics.md
valaparthvi: I meant that an application can be accessed over multiple namespaces. But I see how it might have come off to mean that a user can access the same component over multiple namespaces, is that what you meant?

Anyway, I'll try to fix that.

website/docs/getting-started/basics.md
valaparthvi: Yeah, I actually picked that up from one of the existing docs. I'll try to fix that.

website/docs/getting-started/basics.md
valaparthvi: :grimacing: Okay. Changing that.

website/docs/getting-started/basics.md
valaparthvi: Again, picked it up from existing docs, and it seemed to make sense to me, but I get your point.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
...
 * Context
+  A context is the directory where the source code, tests, and libraries for your component resides. A single context can only contain a single component.
 * URL
+  A URL exposes your component to be accessed outside the cluster.
+  > A URL is an equivalent of a service in Kubernetes.
 * Storage
+  A storage volume is [PVC](https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim) which is a way for you to "claim" persistent storage without knowing the details of the environment. Storage volume can persist data across restarts and rebuilds of a component.
+  > A storage volume is an equivalent of PVC in Kubernetes. Creating a storage with odo will create a PVC resource in Kubernetes with the same name.
 * Service
+  A service is another microservice, or a Kubernetes Custom Resource that your component connects to or depends on. Example of services: MariaDB, MySQL
```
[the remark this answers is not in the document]
valaparthvi: How would you define an operator-backed service? If I use the term, I will still need to explain what a service is. How do I define a service?
By "microservice" I meant component. Ah, I confused service with link :facepalm:

website/docs/getting-started/basics.md
valaparthvi: :grimacing: 

>How would you define an operator-backed service? If I use the term, I will still need to explain what a service is. How do I define a service?

Quoting the same comment as here https://github.com/openshift/odo/pull/4914#discussion_r670525665

website/docs/getting-started/basics.md
valaparthvi: >This is totally wrong. It's a link that helps setup connectivity between microservices

Right, my bad. I mixed up service with a link while writing this.

website/docs/getting-started/basics.md
```diff
@@ -8,15 +8,34 @@ odo abstracts Kubernetes concepts into developer focussed terminology. In this d
 
 ### Concepts
 
-#### Component
+* Component
+  A component is a microservice. Each component can be run and deployed separately. Examples of components: Warehouse API Backend, Inventory API, Web Frontend, Payment Backend
```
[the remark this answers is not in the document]
valaparthvi: Yeah, my bad, I was too focused on the quoted part. I'll fix that.

website/docs/getting-started/basics.md
valaparthvi: >But it is probably not a good place to talk about these implementations. This concept page should help the reader make a mental representation of the objects handled by odo, without other distraction.

Okay okay. I will remove the quoted texts. @feloy, let me know if you have something specific in mind about formatting this.

website/docs/getting-started/basics.md
feloy: A service in the odo sense is some external application that the user's application can interact with to gain some functionalities.

This wikipedia page can help find more inspiration: https://en.wikipedia.org/wiki/Service-oriented_architecture

website/docs/getting-started/basics.md
valaparthvi: > >I understand that the quoted part explain how the concepts are implemented in Kubernetes.
> >
> > But it is probably not a good place to talk about these implementations. This concept page should help the reader make a mental representation of the objects handled by odo, without other distraction.
> 
> Okay okay. I will remove the quoted texts. @feloy, let me know if you have something specific in mind about formatting this.

On a second thought, we mention early on in this page that we will be talking about these concepts and their Kubernetes equivalent, so it does make sense to me to put them here.

website/docs/getting-started/basics.md
valaparthvi: @dharmit Do we need this section? I think that we covered the differences in the concepts section and I cannot think of anything new to put in the component vs application section.

website/docs/getting-started/basics.md
valaparthvi: The OCP part was just for my fyi, right? Do we want to mention it in the docs?

website/docs/getting-started/basics.md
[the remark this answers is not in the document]
dharmit: > How would you define an operator-backed service? If I use the term, I will still need to explain what a service is. How do I define a service?

A loose definition of service is a long-running daemon (think a server process), and listening for incoming requests based on which it would do certain actions. But we don't need to explain this in our docs. It's a definition that almost everyone relates to when they read the word "service".

When you write "Operator backed service", add a link to the term Operator. Use https://github.com/operator-framework.

website/docs/getting-started/basics.md
dharmit: > I figured we would put the Kubernetes part in a quoted text like [Philippe has done in his blog](https://feloy.medium.com/odo-a-tool-to-simplify-development-on-kubernetes-ce0df34313b7).
>        I can add a note about it in the beginning.

We don't want to introduce k8s lingo here.

> On a second thought, we mention early on in this page that we will be talking about these concepts and their Kubernetes equivalent, so it does make sense to me to put them here.

My bad. We should not talk about k8s equivalents in this doc. Can you remove that statement as well?

website/docs/getting-started/basics.md
dharmit: What I understood is that the components can be distributed over different namespaces. Which is not true for odo. At least not yet.

website/docs/getting-started/basics.md
dharmit: Yeah. Even the Ingress thing was for reference. We need to explain URL without talking about Ingresses or Routes.

website/docs/getting-started/basics.md
dharmit: We can remove it from here.

website/docs/getting-started/basics.md
dharmit: We can get rid of the "odo service vs. k8s service" from this page as long as we explain odo services well. The comparison can be kept for Command Reference section.

website/docs/getting-started/basics.md
valaparthvi: Okay. Where else do we mention it then? This seemed like the right place to me to explain about the K8s equivalent of these concepts.

feloy: We could limit talking about Kubernetes in the Command Reference and Architecture pages

valaparthvi: But I do think that talking about the K8s equivalent here will help get a better understanding of odo in general. We do not need to talk about the deep implementation level details, just the high-level ones to make a relation between odo terminology and K8s terminology.

I understand that it might defeat the purpose of abstracting K8s out of the docs, but if I was reading this doc as odo user, having this relation would help me better understand the docs. Or I guess we can make a better effort at making these concepts better understandable.

dharmit: > I understand that it might defeat the purpose of abstracting K8s out of the docs,

This is why we don't want to introduce k8s lingo so early.

> but if I was reading this doc as odo user, having this relation would help me better understand the docs

We do not want application developers (our target audience) to deal with k8s lingo. It's for application architects and whoever _already_ has knowledge of k8s. Majority of odo users are not expected to have knowledge of k8s.

> Or I guess we can make a better effort at making these concepts better understandable.

And that we'll be doing in Command Reference and Architecture sections. Maybe at some places in tutorials where it really makes sense. But Getting Started is definitely not the place to introduce k8s lingo.

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+* **Application**
+  
+  An application is a group of one or more components that work individually to build the entire application. Examples of applications: Online Video Streaming, Hotel Reservation System, Online Booking
 
-### odo services vs. Kubernetes services
\ No newline at end of file
+
+* **Project**
+  
+  A project is a separate single unit that provides a scope for names and helps in dividing cluster resources between users. A resource name must be unique within the project but not across multiple projects.
```
dharmit, line 22: Wow! That sounds like a textbook definition. Where did you get that from?
```suggestion
  A project helps achieve multi-tenancy. Resource names must be unique within the project but not across multiple projects.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+
+* **Context**
+  
+  A context is the directory where the source code, tests, libraries and odo specific config files for your component resides. A single context can only contain one component.
```
dharmit, line 27: IMO, we should point that context "resides" on the user's system.
```suggestion
  A context is the directory on your system where the source code, tests, libraries and config files for a component reside. Loosely speaking, it is a locally cloned git repository and odo specific config files (created by `odo create`). A single context can only contain one component.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+* **URL**
+  
+  A URL exposes your component to be accessed outside the cluster.
+
+
+* **Storage**
+  
+  A storage is a way to claim persistent storage in the cluster. A storage can persist data across restarts and rebuilds of a component.
```
dharmit, line 36:
```suggestion
  A storage is persistent storage in the cluster. It helps persist data across restarts and rebuilds of a component.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+
+* **Service**
+  
+  A service is an external application that a component can connect to or depend on to gain some functionality. Example of services: MySQL, Redis
```
dharmit, line 41:
```suggestion
  A service is an external application that a component can connect to or depend on to gain some functionality. Example of services: MySQL, Redis.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+
+* **Devfile**
+  
+  A portable YAML file responsible for your entire reproducible development environment. See [Devfile](../architecture/devfile.md) to know more about devfile.
+  
+
+### odo services vs. Kubernetes services
+Service in terms of odo is an external application that provides an additional functionality, odo service can only be created with [Operators](https://operatorframework.io/what/).
```
dharmit, line 50:
```suggestion
Service, in terms of odo, is an external application that provides some extra functionality. odo services can only be created with [Operators](https://operatorframework.io/what/).
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+Service in terms of Kubernetes is an abstract way of exposing a microservice as a set of pods. Learn more about [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/).
```
dharmit, line 52:
```suggestion
Service, in terms of Kubernetes, is an abstract way of exposing a microservice. Learn more about [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/).
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,51 @@ sidebar_position: 3
...
+
+Deploying a service created with odo will create a set of resources in the Kubernetes cluster to establish a successful connection between the odo component and the external application.
```
dharmit, line 54: I strongly feel like we should have this differentiation in the Command Reference section where we talk about `odo service` commands.

Without creating a link, the below statement is incorrect:
> establish a successful connection between the odo component and the external application.
```suggestion
Deploying a service created with odo will create a set of resources in the Kubernetes cluster which can then be used from within an odo component after performing link operation.
```

dharmit (changes requested):

valaparthvi:
@dharmit I have removed the `component vs application` and `odo svc vs k8s svc` section and made some cosmetic changes to the site. I had a call with @feloy where we did a live review and edit session, we also discussed the changes you proposed to the docs and tried to make it simpler to understand while abstracting all the K8s references.
I agree with your point about the examples being boring(they reminded me of textbooks too :grimacing:), but I am having a little difficult time coming up with cool examples, let me know if you have some on your mind.

dharmit (approved):
/lgtm

dharmit:
@feloy leaving `/approve` for you to add to this PR.

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,34 @@ sidebar_position: 3
...
 
 #### Component
-* Application
-* Project
-* Context
-* URL
-* Storage
-* Service
-* Devfile
+  In the cloud-native architecture, an application is a collection of small, independent, and loosely coupled components; and an odo component is one of these components.
```
feloy, line 15:
```suggestion
  In the cloud-native architecture, an application is a collection of small, independent, and loosely coupled components; an odo component is one of these components.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,34 @@ sidebar_position: 3
...
 
-### Component vs. Application
+  Examples of components: Warehouse API Backend, Inventory API, Web Frontend, Payment Backend.
 
-### odo services vs. Kubernetes services
\ No newline at end of file
+#### Project
+  A project helps achieve multi-tenancy: several applications can be run in the same cluster by different teams.
```
feloy, line 20:
```suggestion
  A project helps achieve multi-tenancy: several applications can be run in the same cluster by different teams in different projects.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,34 @@ sidebar_position: 3
...
+
+#### Context
+  A context is the directory on the system that contains the source code, tests, libraries and odo specific config files for a single component.
+
+#### URL
+  A URL exposes a component to be accessed outside the cluster.
```
feloy, line 26:
```suggestion
  A URL exposes a component to be accessed from outside the cluster.
```

website/docs/getting-started/basics.md
```diff
@@ -4,19 +4,34 @@ sidebar_position: 3
...
+
+#### Storage
+  A storage is persistent storage in the cluster: it persists the data across restarts and rebuilds of a component.
```
feloy, line 29:
```suggestion
  A storage is a persistent storage in the cluster: it persists the data across restarts and rebuilds of a component.
```

PR #4919, opened 2021-07-16
Do not display "Waiting for component to start" twice

### LINKED ISSUE #4362: "Waiting for component to start" gets printed multiple times
/kind bug

**Operating System:** Fedora

**Output of `odo version`:** master

`odo push -f`

## Actual behavior

Due to #4358 

```
[mrinaldas@localhost project]$ odo push -f
 ✓  Waiting for component to start [5ms]

Validation
 ✓  Validating the devfile [33292ns]

 ⚠  Unable to create ingress, missing host information for Endpoint http-3000, please check instructions on URL creation (refer `odo url create --help`)


Applying URL changes

Syncing to component nodejs
 ✓  Checking file changes for pushing [419423ns]
 ✓  Syncing files to the component [207ms]

 ✓  Waiting for component to start [2ms]
 ✓  Waiting for component to start [2ms]
 ✓  Executing install command "npm install" [2s]
 ✓  Executing run command "npm start" [1s]

```
`Waiting for component to start` gets printed multiple times as both `WaitForDeploymentRollout` and `WaitAndGetPodWithEvents` start a new spinner with the same message in most cases.

## Expected behavior

`Waiting for component to start` gets printed once

feloy:
**What type of PR is this?**

/kind bug

`WaitAndGetPodWithEvents` now waits for an event with a Phase different from the expected one before to start the Spinner. This way, the spinner is only displayed when the pod is not in the expected phase.

Fixes #4362 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

- [x] Update changelog

### DIFF
--- a/pkg/kclient/pods.go
+++ b/pkg/kclient/pods.go
@@ -26,7 +26,6 @@ import (
 
 // WaitAndGetPod block and waits until pod matching selector is in in Running state
 // desiredPhase cannot be PodFailed or PodUnknown
-// hideSpinner hides the spinner
 func (c *Client) WaitAndGetPodWithEvents(selector string, desiredPhase corev1.PodPhase, waitMessage string) (*corev1.Pod, error) {
 
 	// Try to grab the preference in order to set a timeout.. but if not, we'll use the default.
@@ -40,8 +39,12 @@ func (c *Client) WaitAndGetPodWithEvents(selector string, desiredPhase corev1.Po
 
 	klog.V(3).Infof("Waiting for %s pod", selector)
 
-	spinner := log.Spinner(waitMessage)
-	defer spinner.End(false)
+	var spinner *log.Status
+	defer func() {
+		if spinner != nil {
+			spinner.End(false)
+		}
+	}()
 
 	w, err := c.KubeClient.CoreV1().Pods(c.Namespace).Watch(context.TODO(), metav1.ListOptions{
 		LabelSelector: selector,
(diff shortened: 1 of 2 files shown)


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -56,6 +57,22 @@ func (a *Adapter) getPod(refresh bool) (*corev1.Pod, error) {
 	if refresh || a.pod == nil {
 		podSelector := fmt.Sprintf("component=%s", a.ComponentName)
 
+		// First check if the pod is already in desired state
+		c := a.Client.GetKubeClient()
+		podList, err := c.KubeClient.CoreV1().Pods(c.Namespace).List(context.TODO(), metav1.ListOptions{
```
mik-dass, line 62: I think we can avoid listing the pods and use the condition `appsv1.DeploymentAvailable` in the `WaitAndGetPodWithEvents` function to start and stop the spinner

feloy: In fact, I cannot see any use case where this function getPod is really waiting for the pod to be started. @mik-dass do you have any example when this function is waiting?

feloy: I changed to start the spinner only when a phase different from the expected one is received

pkg/kclient/pods.go
```diff
@@ -40,8 +39,12 @@ func (c *Client) WaitAndGetPodWithEvents(selector string, desiredPhase corev1.Po
 
 	klog.V(3).Infof("Waiting for %s pod", selector)
 
-	spinner := log.Spinner(waitMessage)
-	defer spinner.End(false)
+	var spinner *log.Status
+	defer func() {
+		if spinner != nil {
```
dharmit, line 44: In what situation will this check be not satisfied? We're doing `var spinner *log.Status` right above this so, I'm confused as to when `spinner` will be `nil` and this check would be not satisfied.

feloy: In `var spinner *log.Status`, spinner is a non initialized pointer, so getting its zero value, nil. 
The pointer gets affected a value only when receiving a phase different from the expected one.

mik-dass (approved):
/lgtm

PR #4926, opened 2021-07-21
Removes Girish from OWNERS files

dharmit:
Used a one-liner while learning, so I'm not sure if it removes references from all places. Here's the one-liner I used:
```sh
$ find . -iname "*OWNERS*" -not -path "./vendor/*" | xargs sed -i '/girishramnani/d'
```

### DIFF
--- a/pkg/devfile/OWNERS
+++ b/pkg/devfile/OWNERS
@@ -1,4 +1,3 @@
 # See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
 
 approvers:
-- girishramnani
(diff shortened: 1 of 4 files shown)


pkg/devfile/OWNERS
```diff
@@ -1,4 +1,3 @@
 # See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
 
 approvers:
-- girishramnani
```
valaparthvi, line 4: This file has no approvers, I wonder if it makes sense to have this file at all.

valaparthvi (changes requested):

pkg/devfile/OWNERS
dharmit: Thanks for catching that. Removed the file.

valaparthvi (approved):

Issue #4741, opened 2021-05-24
Improve tests readability and spec clean up


prietyc123:
We have currently a messy tests spec which pretty hard to analyze multiple test spec description quickly in the same test file because the implementation of the spec happens in the same file. As a user, he or she will be pretty happy to see the test spec description in the very next line. For example 
```code
func ProjectTestScenario() {
    When("Machine readable output tests", func() {
        It("Help for odo project list should contain machine output", HelpCommand) // HelpCommand will be expanded in a separate package
        It("should be able to get project", GetProject)
    })
    When("when running help for project command", func() {
        It("should display the help", HelpCommandProject) // HelpCommandProject will be expanded in a separate package
    })
}
```
Advantages:
1. Our test will get an index page like a book having an index
2. Test Readability 
3. Easy to figure what test case we are missing
4. Scenario integrity 
5. Easy to understand

prietyc123:
Test needs to be refactored:
S2i Integration Test:
- [x] test-cmd-project
- [ ] test-cmd-app
- [ ] test-cmd-cmp
- [ ] test-cmd-cmp-sub
- [ ] test-cmd-pref-config
- [ ] test-cmd-push
- [ ] test-plugin-handler
- [ ] test-cmd-service
- [ ] test-cmd-link-unlink-311-cluster
- [ ] test-cmd-link-unlink-4-cluster
- [ ] test-cmd-login-logout
- [ ] test-generic
- [ ] test-cmd-storage
- [ ] test-cmd-url
- [ ] test-cmd-watch
- [ ] test-cmd-debug
Devfile:
- [ ] test-cmd-devfile-catalog
- [ ] test-cmd-devfile-create
- [ ] test-cmd-devfile-push
- [ ] test-cmd-devfile-exec
- [ ] test-cmd-devfile-status
- [ ] test-cmd-devfile-watch
- [ ] test-cmd-devfile-app
- [ ] test-cmd-devfile-delete
- [ ] test-cmd-devfile-registry
- [ ] test-cmd-devfile-test
- [ ] test-cmd-devfile-url
- [ ] test-cmd-devfile-debug
- [ ] test-cmd-devfile-storage
- [ ] test-cmd-devfile-log
- [ ] test-cmd-devfile-env
- [ ] test-cmd-devfile-config
E2E Tests:
- [ ] test-e2e-beta
- [ ] test-e2e-java
- [ ] test-e2e-source
- [ ] test-e2e-images
[...]

PR #4927, opened 2021-07-21
Set correct component name in the devfile metadata

### LINKED ISSUE #4815: `odo create <comp-type> <comp-name>` should use `<comp-name>` value in devfile's `.metadata.name`
/kind feature

As pointed out in https://github.com/openshift/odo/issues/4802#issuecomment-862297237, `odo create` did what it's doing right now because there was no other option at the time.

But now the behaviour should be changed so that it looks like the **Expected** section.
### Now
```sh
$ odo create nodejs nodejs-ex-123
Devfile Object Validation
 ✓  Checking devfile existence [37955ns]
 ✓  Creating a devfile component from registry: OCI [49465ns]
Validation
 ✓  Validating if devfile name is correct [36998ns]

$ grep -A3 metadata devfile.yaml
metadata:
  name: nodejs
  version: 1.0.1
  displayName: NodeJS Runtime
```

### Expected
```sh
$ odo create nodejs nodejs-ex-123
Devfile Object Validation
 ✓  Checking devfile existence [37955ns]
 ✓  Creating a devfile component from registry: OCI [49465ns]
Validation
 ✓  Validating if devfile name is correct [36998ns]

$ grep -A3 metadata devfile.yaml
metadata:
  name: nodejs-ex-123
  version: 1.0.1
  displayName: NodeJS Runtime
```

## Why is this needed?
More info on https://github.com/openshift/odo/issues/4802.

/area component

valaparthvi:
**What type of PR is this?**
 /kind bug
/kind code-refactoring

1. Set the correct component name in the devfile metadata.
2. In case a devfile is already present and user runs `odo create`, name the component to what is present in the devfile and not the dir name. If the `.metadata.name` is not present or if the devfile is not accessible, only then name the component to context dir name.
2. List and Describe to show the `.metadata.projectType` or `.metadata.Language` as the component type, if neither of them are available, it sets the type to `Not available`.
2. Add tests, and modify existing tests.
4. Move `GetComponentTypeFromMetadata` from `create_helpers.go` to `component.go`  so that the function can be used in both the files without any cyclic dependencies.
5. Add annotation for component-type to the component being pushed, and make `getType` to retrieve this annotation to check type, instead of label. For an existing pushed component, if an annotation is not present, but a label is, simply return Not available, and running with verbose suggests pushing the component again.
6. `getRemoteComponentMetadata` now returns annotations and labels.

Fixes #4815

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [x] Update changelog

1. `odo create java-maven mycomp`
2. Check if `mycomp` is set in devfile.yaml.
3. Run `odo describe`; also check for `odo describe -o json`.
4. Check if `mycomp` is displayed in `Name` and `Maven` in `Type`.
5. Run `odo list`, also check for `odo list -o json`.
6. Check if `mycomp` is displayed in `Name` and `Maven` in `Type`.
Experiment the above by removing projectType or, both projectType and Language.

Additionally,
1. Copy a devfile to the context dir.
2. `odo create`
3. Run `odo list` or `odo describe` and check its name, make sure it corresponds to that of `.metadata.name`, and in case the field is absent, it should name it to the context dir.
Experiment the above by removing the metadata.name along with `odo create custom_name`.

### DIFF
--- a/pkg/component/pushed_component.go
+++ b/pkg/component/pushed_component.go
@@ -247,7 +247,7 @@ func getType(component provider) (string, error) {
 		return componentType, nil
 	} else if _, ok = component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
 		klog.V(1).Info("No annotation assigned; retuning 'Not available' since labels are assigned. Annotations will be assigned when user pushes again.")
-		return NOTAVAILABLE, nil
+		return NotAvailable, nil
 	}
 	return "", fmt.Errorf("%s component doesn't provide a type annotation; consider pushing the component again", component.GetName())
 }
--- a/pkg/component/pushed_component.go
+++ b/pkg/component/pushed_component.go
@@ -243,10 +243,13 @@ func getSource(component provider) (string, string, error) {
 }
 
 func getType(component provider) (string, error) {
-	if componentType, ok := component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
+	if componentType, ok := component.GetAnnotations()[componentlabels.ComponentTypeAnnotation]; ok {
 		return componentType, nil
+	} else if _, ok = component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
+		klog.V(1).Info("No annotation assigned; retuning 'Not available' since labels are assigned. Annotations will be assigned when user pushes again.")
+		return NOTAVAILABLE, nil
 	}
-	return "", fmt.Errorf("%s component doesn't provide a type label", component.GetName())
+	return "", fmt.Errorf("%s component doesn't provide a type annotation; consider pushing the component again", component.GetName())
 }
 
 // GetPushedComponents retrieves a map of PushedComponents from the cluster, keyed by their name
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -500,12 +500,20 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 					componentName = args[0]
 
 				} else {
-					currentDirPath, err := os.Getwd()
-					if err != nil {
-						return err
+					// If there is an existing devfile, and no component name is passed, parse it from the devfile,
+					// and assign the value if the metadata name is set
+					devfileObj, err := devfile.ParseFromFile(DevfilePath)
+					if (err == nil) && (devfileObj.GetMetadataName() != "") {
+						componentName = devfileObj.GetMetadataName()
+					} else {
+						// If the metadata name is not available, then assign the current directory name to component
+						currentDirPath, err := os.Getwd()
+						if err != nil {
+							return err
+						}
+						currentDirName := filepath.Base(currentDirPath)
+						componentName = currentDirName
 					}
-					currentDirName := filepath.Base(currentDirPath)
-					componentName = currentDirName
 				}
 
 				co.devfileMetadata.devfileSupport = true
@@ -856,7 +864,7 @@ func (co *CreateOptions) devfileRun(cmd *cobra.Command) (err error) {
 	// Add component type in case it is not already added or is empty
 	if scontext.GetTelemetryStatus(cmd.Context()) {
 		if value, ok := scontext.GetContextProperties(cmd.Context())[scontext.ComponentType]; !ok || value == "" {
-			scontext.SetComponentType(cmd.Context(), GetComponentTypeFromDevfile(devObj.Data.GetMetadata()))
+			scontext.SetComponentType(cmd.Context(), component.GetComponentTypeFromDevfileMetadata(devObj.Data.GetMetadata()))
 		}
 	}
 	err = validate.ValidateDevfileData(devObj.Data)
@@ -887,6 +895,19 @@ func (co *CreateOptions) devfileRun(cmd *cobra.Command) (err error) {
 		err = registryLibrary.PullStackFromRegistry(co.devfileMetadata.devfileRegistry.URL, co.devfileMetadata.componentType, co.componentContext)
 		if err != nil {
 			return err
+		}
+	}
+
+	// set user provided component name in the devfile
+	if co.devfileMetadata.componentName != "" {
+		devObj, err = devfile.ParseFromFile(DevfilePath)
+		if err != nil {
+			return fmt.Errorf("failed to create devfile component: %w", err)
+		}
+
+		err = devObj.SetMetadataName(co.devfileMetadata.componentName)
+		if err != nil {
+			return fmt.Errorf("failed to create devfile component: %w", err)
 		}
 	}
 
--- a/pkg/odo/cli/component/create.go
+++ b/pkg/odo/cli/component/create.go
@@ -503,7 +503,7 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
 					// If there is an existing devfile, and no component name is passed, parse it from the devfile,
 					// and assign the value if the metadata name is set
 					devfileObj, err := devfile.ParseFromFile(DevfilePath)
-					if (err == nil) && (devfileObj.GetMetadataName() != "") {
+					if err == nil && devfileObj.GetMetadataName() != "" {
 						componentName = devfileObj.GetMetadataName()
 					} else {
 						// If the metadata name is not available, then assign the current directory name to component
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -356,17 +356,6 @@ func componentTests(args ...string) {
 				BeforeEach(func() {
 					helper.Cmd("odo", append(args, "push", "--context", commonVar.Context)...).ShouldPass()
 				})
-				When("odo describe is executed", func() {
-					var stdOut string
-					BeforeEach(func() {
-						stdOut = helper.Cmd("odo", "describe").ShouldPass().Out()
-					})
-					It("should have the correct component Name and Type in the describe output", func() {
-						metadata := helper.GetMetadataFromDevfile("devfile.yaml")
-						Expect(stdOut).To(ContainSubstring(metadata.ProjectType))
-						Expect(stdOut).To(ContainSubstring(cmpName))
-					})
-				})
 
 				It("should not list component even in new project with --project and --context at the same time", func() {
 					projectList := helper.Cmd("odo", "project", "list").ShouldPass().Out()
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -358,7 +358,7 @@ func componentTests(args ...string) {
 				})
 				When("odo describe is executed", func() {
 					var stdOut string
-					JustBeforeEach(func() {
+					BeforeEach(func() {
 						stdOut = helper.Cmd("odo", "describe").ShouldPass().Out()
 					})
 					It("should have the correct component Name and Type in the describe output", func() {
--- a/tests/integration/component.go
+++ b/tests/integration/component.go
@@ -9,8 +9,6 @@ import (
 	"runtime"
 	"time"
 
-	"github.com/openshift/odo/pkg/devfile"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/odo/tests/helper"
@@ -294,17 +292,17 @@ func componentTests(args ...string) {
 				Expect(info.GetApplication(), appName)
 				Expect(info.GetName(), cmpName)
 			})
+
 			It("should set the correct component name, language and projectType in devfile metadata", func() {
-				devObj, err := devfile.ParseFromFile(filepath.Join(commonVar.Context, "devfile.yaml"))
-				Expect(err).ToNot(HaveOccurred())
-				metadata := devObj.Data.GetMetadata()
+				metadata := helper.GetMetadataFromDevfile("devfile.yaml")
 				Expect(metadata.Name).To(BeEquivalentTo(cmpName))
 				Expect(metadata.Language).To(ContainSubstring("nodejs"))
 				Expect(metadata.ProjectType).To(ContainSubstring("nodejs"))
 			})
 			It("should list the component when it is not pushed", func() {
 				cmpList := helper.Cmd("odo", append(args, "list", "--context", commonVar.Context)...).ShouldPass().Out()
-				helper.MatchAllInOutput(cmpList, []string{cmpName, "Not Pushed"})
+				metadata := helper.GetMetadataFromDevfile("devfile.yaml")
+				helper.MatchAllInOutput(cmpList, []string{cmpName, "Not Pushed", metadata.ProjectType})
 				helper.Cmd("odo", append(args, "delete", "-f", "--all", "--context", commonVar.Context)...).ShouldPass()
 			})
 
@@ -337,9 +335,10 @@ func componentTests(args ...string) {
 				helper.Cmd("odo", "url", "create", "url-2", "--context", commonVar.Context).ShouldPass()
 				helper.Cmd("odo", "storage", "create", "storage-1", "--size", "1Gi", "--path", "/data1", "--context", commonVar.Context).ShouldPass()
 				cmpDescribe := helper.Cmd("odo", append(args, "describe", "--context", commonVar.Context)...).ShouldPass().Out()
+				metadata := helper.GetMetadataFromDevfile("devfile.yaml")
 				helper.MatchAllInOutput(cmpDescribe, []string{
 					cmpName,
-					"nodejs",
+					metadata.ProjectType,
 					"url-1",
 					"url-2",
 					"storage-1",
@@ -357,6 +356,17 @@ func componentTests(args ...string) {
 				BeforeEach(func() {
 					helper.Cmd("odo", append(args, "push", "--context", commonVar.Context)...).ShouldPass()
 				})
+				When("odo describe is executed", func() {
+					var stdOut string
+					JustBeforeEach(func() {
+						stdOut = helper.Cmd("odo", "describe").ShouldPass().Out()
+					})
+					It("should have the correct component Name and Type in the describe output", func() {
+						metadata := helper.GetMetadataFromDevfile("devfile.yaml")
+						Expect(stdOut).To(ContainSubstring(metadata.ProjectType))
+						Expect(stdOut).To(ContainSubstring(cmpName))
+					})
+				})
 
 				It("should not list component even in new project with --project and --context at the same time", func() {
 					projectList := helper.Cmd("odo", "project", "list").ShouldPass().Out()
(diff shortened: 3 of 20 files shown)


pkg/odo/cli/component/create.go
```diff
@@ -879,6 +873,19 @@ func (co *CreateOptions) devfileRun(cmd *cobra.Command) (err error) {
 		}
 	}
 
+	// set user provided component name in the devfile
+	if co.devfileMetadata.componentName != "" {
+		devObj, err = devfile.ParseFromFile(DevfilePath)
```
mik-dass, line 878: Why are we not setting the below changes to the devfile before it is written to the disk?

valaparthvi: Setting these changes before the devfile is written to the disk is not possible with the 2 approaches that we tried - 
1.  We could use `devObj.WriteYamlDevfile()` after modifying the devObj but that is not possible because it does not take the correct path into account.
2. We parse the devfileData to devObj to validate the YAML file, so another option is to convert the devObj back to YAML using the k8s YAML library and write that YAML but that seems to be overridden by the code on [L869](https://github.com/openshift/odo/pull/4927/files#diff-e9cbb4b32525048f3bf6ffce25bb7e24c02221819d5b9de9347fe2c835828058R869).


The second approach seems like a viable option, so I have dropped a comment on https://github.com/openshift/odo/pull/4636#discussion_r680913176 because the code seems redundant to me. If we can remove that code, we can go with the second option.

mik-dass (commented):
/lgtm

dharmit (changes requested):
While `odo describe` works as you said it should, `odo list` doesn't. It shows incorrect value for `TYPE` column.
```sh
$ odo describe
Component Name: mynode-123
Type: nodejs

$ odo list    
Devfile Components: 
APP     NAME           PROJECT     TYPE           STATE
app     mynode-123                 mynode-123     Not Pushed
```

valaparthvi:
/retest 

Reason - 
```
[must-gather      ] OUT Using must-gather plug-in image: registry.redhat.io/openshift4/ose-must-gather:latest
When opening a support case, bugzilla, or issue please include the following summary data along with any other requested information.
error getting cluster version: Get "https://api.ci-op-s21v700v-d028c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version": dial tcp 52.7.180.144:6443: i/o timeout
ClusterID: 
ClusterVersion: Installing "" for <unknown>: <unknown>
error getting cluster operators: Get "https://api.ci-op-s21v700v-d028c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp 52.7.180.144:6443: i/o timeout
ClusterOperators:
	clusteroperators are missing




When opening a support case, bugzilla, or issue please include the following summary data along with any other requested information.
error getting cluster version: Get "https://api.ci-op-s21v700v-d028c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version": dial tcp 44.196.163.44:6443: i/o timeout
ClusterID: 
ClusterVersion: Installing "" for <unknown>: <unknown>
error getting cluster operators: Get "https://api.ci-op-s21v700v-d028c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp 44.196.163.44:6443: i/o timeout
ClusterOperators:
	clusteroperators are missing
```

pkg/component/pushed_component.go
```diff
@@ -250,10 +250,13 @@ func getSource(component provider) (string, string, error) {
 }
 
 func getType(component provider) (string, error) {
-	if componentType, ok := component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
+	if componentType, ok := component.GetAnnotations()["odo.dev/project-type"]; ok {
 		return componentType, nil
+	} else if _, ok = component.GetLabels()[componentlabels.ComponentTypeLabel]; ok {
+		klog.V(1).Info("No annotation assigned; retuning 'Not available' since labels are assigned. Annotations will be assigned when user pushes again.")
+		return NOTAVAILABLE, nil
```
valaparthvi, line 257: Since we will be using Annotations to pass the actual component type, we use Annotations to retrieve it as well, but for an existing pushed component, this annotation is not added, it will be added when it is pushed again, until then, we can either return the component type retrieved from label or return `Not available`, which one should it be?

valaparthvi:
/retest

Reason - 
```
 [ssh:Fedora-32-minikube] • Failure [7.824 seconds]
[ssh:Fedora-32-minikube] odo devfile describe command tests
[ssh:Fedora-32-minikube] /home/fedora/amqp_ci_rcv_odo-minikube-pr-build_PR_4927/iTfKg/repo/tests/integration/devfile/cmd_devfile_describe_test.go:16
[ssh:Fedora-32-minikube]   When executing odo describe
[ssh:Fedora-32-minikube]   /home/fedora/amqp_ci_rcv_odo-minikube-pr-build_PR_4927/iTfKg/repo/tests/integration/devfile/cmd_devfile_describe_test.go:33
[ssh:Fedora-32-minikube]     should describe the component when it is not pushed [It]
[ssh:Fedora-32-minikube] 
[ssh:Fedora-32-minikube] [Fail] odo devfile describe command tests When executing odo describe [It] should describe the component when it is not pushed 
[ssh:Fedora-32-minikube] /home/fedora/amqp_ci_rcv_odo-minikube-pr-build_PR_4927/iTfKg/repo/tests/integration/devfile/cmd_devfile_describe_test.go:63
[ssh:Fedora-32-minikube] 
[ssh:Fedora-32-minikube] Ran 196 of 207 Specs in 1431.846 seconds
[ssh:Fedora-32-minikube] FAIL! -- 195 Passed | 1 Failed | 1 Pending | 10 Skipped
[ssh:Fedora-32-minikube] 
```

Passes locally.

dharmit:
> While `odo describe` works as you said it should, `odo list` doesn't. It shows incorrect value for `TYPE` column.
> 
> ```shell
> $ odo describe
> Component Name: mynode-123
> Type: nodejs
> 
> $ odo list    
> Devfile Components: 
> APP     NAME           PROJECT     TYPE           STATE
> app     mynode-123                 mynode-123     Not Pushed
> ```

Before I review this PR again, have you addressed above or is it going to be done in a different PR?

valaparthvi:
>Before I review this PR again, have you addressed above or is it going to be done in a different PR?

Of course, I have. This comment was the reason why I had to do a lot of rework.

pkg/odo/cli/component/create.go
```diff
@@ -499,12 +499,20 @@ func (co *CreateOptions) Complete(name string, cmd *cobra.Command, args []string
...
 
 				} else {
-					currentDirPath, err := os.Getwd()
-					if err != nil {
-						return err
+					// If there is an existing devfile, and no component name is passed, parse it from the devfile,
+					// and assign the value if the metadata name is set
+					devfileObj, err := devfile.ParseFromFile(DevfilePath)
+					if err == nil && devfileObj.GetMetadataName() != "" {
+						componentName = devfileObj.GetMetadataName()
```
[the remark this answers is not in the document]
dharmit: The `Complete` function in this case seems to be doing SO many things already. I don't see why we can't do more till the point it becomes a mess we need to absolutely fix. 😕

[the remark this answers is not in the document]
dharmit: Hopefully getting rid of s2i components will make it more readable.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -88,6 +95,19 @@ var _ = Describe("odo devfile describe command tests", func() {
 	})
 
 	Context("when running odo describe for machine readable output", func() {
+		When("a component is created with a custom name", func() {
```
dharmit, line 98: The `Context` above starts with `when`. Classic example of a test file that's not refactored, and we're adding tests as per our understanding based on #4770.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -107,4 +127,66 @@ var _ = Describe("odo devfile describe command tests", func() {
 		})
 	})
 
+	Context("devfile has missing metadata", func() {
+		// Reference: https://github.com/openshift/odo/issues/4815
```
dharmit, line 131: I don't think we need to keep referring an issue in the code unless it's something that needs to be fixed as a part of some TODO.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -107,4 +127,66 @@ var _ = Describe("odo devfile describe command tests", func() {
...
+		var metadata devfilepkg.DevfileMetadata
+
+		// copyAndCreate copies required source code and devfile, and creates a component
+		var copyAndCreate = func(path string) {
```
dharmit, line 135: But it takes only one argument `path`. How does it figure out both source code and devfile using it? 🤔

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -107,4 +127,66 @@ var _ = Describe("odo devfile describe command tests", func() {
...
+			// Using SpringBoot example here because it helps to distinguish between language and projectType.
+			// In terms of SpringBoot, spring in the projectType and java is the language
```
dharmit, line 137: Could you remove the hardcoding of Spring Boot example here? Try to make this more generic example by putting it into a helper file that makes sense? I'm asking this because we're copying source and devfile at a lot of places and I see this could be helpful in more generic cases than it would be if it were going to be a function limited to this `Context`.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -107,4 +127,66 @@ var _ = Describe("odo devfile describe command tests", func() {
...
+			helper.CopyExample(filepath.Join("source", "devfiles", "springboot", "project"), commonVar.Context)
+			helper.CopyExampleDevFile(filepath.Join("source", "devfiles", "springboot", path), filepath.Join(commonVar.Context, "devfile.yaml"))
+			helper.Cmd("odo", "create", "--context", commonVar.Context).ShouldPass()
+		}
+
+		// checkDescribe checks the describe output (both normal and json) to see if it contains the expected componentType
+		var checkDescribe = func(componentType string) {
+			By("checking the normal output", func() {
```
dharmit, line 145: What's the purpose of `By`? I'm asking because I expected something nested in it but I didn't find any `It` clause in it.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -107,4 +127,66 @@ var _ = Describe("odo devfile describe command tests", func() {
...
+				Expect(stdOut).To(ContainSubstring(componentType))
+			})
+			By("checking the json output", func() {
+				stdOut := helper.Cmd("odo", "describe", "--context", commonVar.Context, "-o", "json").ShouldPass().Out()
+				Expect(gjson.Get(stdOut, "spec.type").String()).To(Equal(componentType))
+			})
+		}
+
+		When("projectType is missing", func() {
+			JustBeforeEach(func() {
```
dharmit, line 156: Same thing about `JustBeforeEach`. We discussed somewhere to use `BeforeEach`.

pkg/component/pushed_component.go
valaparthvi: @dharmit @kadel ^

tests/integration/devfile/cmd_devfile_list_test.go
```diff
@@ -4,6 +4,10 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/tidwall/gjson"
```
dharmit, line 7: We decided to refactor the tests file whenever some PR modifies it because it is not possible to otherwise refactor our entire test suite as proposed in #4770. Do you want to do this in this PR or in a separate PR right after this PR?

dharmit (changes requested):

pkg/component/pushed_component.go
dharmit: To support backwards compatibility, I would suggest retrieving it from the label for cases where it's not available in the annotations.

tests/integration/component.go
```diff
@@ -357,6 +356,17 @@ func componentTests(args ...string) {
 				BeforeEach(func() {
 					helper.Cmd("odo", append(args, "push", "--context", commonVar.Context)...).ShouldPass()
 				})
+				When("odo describe is executed", func() {
+					var stdOut string
+					JustBeforeEach(func() {
+						stdOut = helper.Cmd("odo", "describe").ShouldPass().Out()
+					})
+					It("should have the correct component Name and Type in the describe output", func() {
```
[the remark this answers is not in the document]
valaparthvi: When I started work on this PR, the s2i removal hadn't begun, so I added them anyway, I guess I can remove it now that s2i removal is in progress.

tests/integration/devfile/cmd_devfile_create_test.go
```diff
@@ -49,6 +49,13 @@ var _ = Describe("odo devfile create command tests", func() {
 		It("should successfully create the devfile component with valid component name", func() {
 			helper.Cmd("odo", "create", "java-openliberty", cmpName).ShouldPass()
 		})
+		It("should set the correct devfile metadata component name and language on component creation", func() {
```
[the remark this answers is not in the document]
valaparthvi: I would much rather do the refactoring in a separate PR.

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -2,9 +2,13 @@ package devfile
 
 import (
 	"os"
+	"path/filepath"
```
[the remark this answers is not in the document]
valaparthvi: https://github.com/openshift/odo/pull/4927#discussion_r700774769

tests/integration/devfile/cmd_devfile_describe_test.go
```diff
@@ -36,14 +40,17 @@ var _ = Describe("odo devfile describe command tests", func() {
 		})
 
 		It("should describe the component when it is not pushed", func() {
-			helper.Cmd("odo", "create", "nodejs", "cmp-git", "--project", commonVar.Project, "--context", commonVar.Context, "--app", "testing").ShouldPass()
+			// Using Django example here because it helps to distinguish between language and projectType.
+			// With nodejs, both projectType and language is nodejs, but with python-django, django is the projectType and python is the language
+			// Reference: https://github.com/openshift/odo/issues/4815
+			helper.Cmd("odo", "create", "python-django", "cmp-git", "--project", commonVar.Project, "--context", commonVar.Context, "--app", "testing").ShouldPass()
```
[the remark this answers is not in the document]
valaparthvi: Name of the component.

tests/integration/devfile/cmd_devfile_describe_test.go
valaparthvi: How about we keep it as is for now, and I pick the refactoring in the next PR next sprint along with all the other files?

dharmit:
@valaparthvi 

> I would much rather do the refactoring in a separate PR.

Sure. Works for me. Please open separate issues for the test files that you plan to refactor in follow-up PRs. Also update #4741 like others have done so far.

tests/integration/devfile/cmd_devfile_describe_test.go
valaparthvi: Both the devfiles in use are of java-springboot, so the source code remains the same. `path` is just the file name. I guess I should change the variable name in this case to make more sense.

tests/integration/devfile/cmd_devfile_describe_test.go
valaparthvi: Quoting from https://onsi.github.io/ginkgo/#documenting-complex-its-by
>As a rule, you should try to keep your Its, BeforeEaches, etc. short and to the point. Sometimes this is not possible, particularly when testing complex workflows in integration-style tests. In these cases your test blocks begin to hide a narrative that is hard to glean by looking at code alone. Ginkgo provides by to help in these situations.

`By` is a way of structuring checks inside the `It`.

tests/integration/devfile/cmd_devfile_describe_test.go
valaparthvi: I figured since this is an unusual scenario, something that we haven't taken into account before, it might help to reference the issue in order to avoid confusion.

tests/integration/devfile/cmd_devfile_describe_test.go
valaparthvi: I would still like to make a note about why Spring Boot is used here, hence I've added a comment in the `Context("devfile has missing metadata", func(){})` block, I hope that is okay.

tests/integration/component.go
```diff
@@ -357,6 +356,17 @@ func componentTests(args ...string) {
 				BeforeEach(func() {
 					helper.Cmd("odo", append(args, "push", "--context", commonVar.Context)...).ShouldPass()
 				})
+				When("odo describe is executed", func() {
+					var stdOut string
+					JustBeforeEach(func() {
```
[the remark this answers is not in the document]
valaparthvi: I believe this(https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach) is what you are looking for.
You are correct in this case, it should be `BeforeEach`.

tests/integration/devfile/cmd_devfile_describe_test.go
dharmit: Thanks.

tests/integration/component.go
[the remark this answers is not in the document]
dharmit: Yeah, please don't add anything related to s2i any more. It'll only make removing things even more all over the place than it already is.

tests/integration/devfile/cmd_devfile_describe_test.go
dharmit: I agree with the unusual part. But I saw the issue getting referenced more than a few times. Besides, `git blame` would always bring someone to this PR and this PR can take them to issue 4815. So, IMO, it's OK to not mention it in the code unless it's a part of TODO.

tests/integration/devfile/cmd_devfile_describe_test.go
dharmit: Yeah. My request was more along the lines of making it generic so that it can be used in more places.

dharmit (approved):

valaparthvi:
My tests are failing because an old image of java-springboot is used, the tests are failing locally as well. 
@feloy Any thoughts on fixing this?

valaparthvi:
>My tests are failing because an old image of java-springboot is used, the tests are failing locally as well.
@feloy Any thoughts on fixing this?

Nevermind, I understand why this is happening. The copied devfile contains the old image, which is why it is failing.

valaparthvi:

reason - perhaps a flake
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/4927/pull-ci-openshift-odo-main-psi-kubernetes-integration-e2e/1434517392939028480

dharmit (approved):

PR #4963, opened 2021-08-03
Send telemetry data on user interrupt

### LINKED ISSUE #4929: Data must be sent even if user terminates the command execution.
/kind feature

Currently, telemetry data is only sent on the complete execution of a command, but we should also be able to send data in case a user terminates the command(Ctrl+C^).

## Why is this needed?
This will help us understand why the command was terminated, maybe the command took a long time and the user decided to terminate it, if that is the case we will be able to know which commands.

valaparthvi:
**What type of PR is this?**
/kind feature

This PR allows sending telemetry data for odo push on user interrupt.

Fixes #4929 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] Update changelog

```sh
odo preference set consenttelemetry true -f
odo create java-openliberty myproject --starter
odo push     <- interrupt the program
Ctrl^C
```

See if telemetry data was sent to Segment.
Sample telemetry data - 
```
map[string]interface{}{
    "clusterType": "kubernetes",
    "componentType": "docker",
    "duration(ms)": 995,
    "error": "User quit.",
    "error-type": "*errors.fundamental",
    "success": false,
    "tty": true,
    "version": "odo v2.2.3 (2e661bf71)",
  }
```

### DIFF
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -89,6 +89,7 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 	captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt, os.Kill}
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, captureSignals...)
+	defer signal.Stop(c)
 	go commonutil.StartSignalWatcher(captureSignals, func() {
 		receivedSignal := <-c
 		scontext.SetSignal(cmd.Context(), receivedSignal)
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -8,7 +8,12 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
+
+	"github.com/pkg/errors"
+
+	commonutil "github.com/openshift/odo/pkg/util"
 
 	"github.com/openshift/odo/pkg/version"
 
@@ -77,7 +82,12 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
-
+	if cmd.Name() == "push" {
+		// TODO: capture ctrl^z
+		go commonutil.StartSignalWatcher([]os.Signal{syscall.SIGTERM, syscall.SIGQUIT, syscall.SIGINT}, func() {
+			startTelemetry(cmd, errors.Errorf("user quit"), startTime)
+		})
+	}
 	err = o.Validate()
 	if err != nil {
 		startTelemetry(cmd, err, startTime)
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -11,11 +11,9 @@ import (
 	"syscall"
 	"time"
 
-	"gopkg.in/AlecAivazis/survey.v1/terminal"
-
 	commonutil "github.com/openshift/odo/pkg/util"
-
 	"github.com/pkg/errors"
+	"gopkg.in/AlecAivazis/survey.v1/terminal"
 
 	"github.com/openshift/odo/pkg/version"
 
@@ -41,7 +39,7 @@ type Runnable interface {
 
 func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 	var err error
-	var startTime time.Time
+	startTime := time.Now()
 	cfg, _ := preference.New()
 	disableTelemetry, _ := strconv.ParseBool(os.Getenv(segment.DisableTelemetryEnv))
 
@@ -69,7 +67,12 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 	// set value for telemetry status in context so that we do not need to call IsTelemetryEnabled every time to check its status
 	scontext.SetTelemetryStatus(cmd.Context(), segment.IsTelemetryEnabled(cfg))
 
-	startTime = time.Now()
+	// Send data to telemetry in case of user interrupt
+	captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt}
+	go commonutil.StartSignalWatcher(captureSignals, func(receivedSignals os.Signal) {
+		scontext.SetSignal(cmd.Context(), receivedSignals)
+		startTelemetry(cmd, errors.Wrapf(terminal.InterruptErr, "user interrupted the command execution"), startTime)
+	})
 
 	// CheckMachineReadableOutput
 	// fixes / checks all related machine readable output functions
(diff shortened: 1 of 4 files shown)


pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,7 +82,12 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
-
+	if cmd.Name() == "push" {
+		// TODO: capture ctrl^z
+		go commonutil.StartSignalWatcher([]os.Signal{syscall.SIGTERM, syscall.SIGQUIT, syscall.SIGINT}, func() {
+			startTelemetry(cmd, errors.Errorf("user quit"), startTime)
```
kadel, line 88: It might be better to actually send the information about signal odo received.

For example adding data like this to telemetry `receivedSignal: SIGTERM`

pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,7 +82,12 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
-
+	if cmd.Name() == "push" {
```
kadel, line 85: We don't need to filter this just for `odo push` this can be useful thing for all commands.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,6 +85,15 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
+	// TODO: capture ctrl^z
+	captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt, os.Kill}
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, captureSignals...)
```
valaparthvi, line 91: TODO: Add `defer signal.Stop()`

pkg/util/util.go
```diff
@@ -1407,14 +1407,10 @@ func copyDirWithFS(src string, dst string, fs filesystem.Filesystem) error {
 	return nil
 }
 
-// StartSignalWatcher watches for kill,interrupt etc signals and cleans up temp files and folders
-func StartSignalWatcher(handle func()) {
+// StartSignalWatcher watches for signals and handles the situation before exiting the program
+func StartSignalWatcher(watchSignals []os.Signal, handle func()) {
```
valaparthvi, line 1411: TODO: Maybe add a unit test for this function?

pkg/util/util.go
```diff
@@ -1465,7 +1461,7 @@ func GitSubDir(srcPath, destinationPath, subDir string) error {
 
 // gitSubDir handles subDir for git components
 func gitSubDir(srcPath, destinationPath, subDir string, fs filesystem.Filesystem) error {
-	go StartSignalWatcher(func() {
+	go StartSignalWatcher([]os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt}, func() {
```
dharmit, line 1464: Why aren't we capturing `os.Kill` here? I'm asking because that would help us reuse:
```go
captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt, os.Kill}
```
defined in `pkg/odo/genericclioptions/runnable.go`.

pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,6 +85,15 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
+	// TODO: capture ctrl^z
```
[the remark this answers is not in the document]
valaparthvi: I am not sure what signal is triggered on ctrl^z. Pressing it stops the operation and doesn't send telemetry data.

pkg/util/util.go
valaparthvi: I am not sure why it wasn't used, Mrinal wrote this function, so he must've had his reasons.

pkg/odo/genericclioptions/runnable.go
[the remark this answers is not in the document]
dharmit: @feloy any pointers?

pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,6 +85,16 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
...
+	captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt, os.Kill}
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, captureSignals...)
```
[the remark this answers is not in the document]
valaparthvi: I want to catch the signal so that it can be sent to telemetry, for which I need another notifier, I don't think I can use the StartSignalWatcher one.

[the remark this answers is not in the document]
feloy: You are catching the same signal twice, the first time in this line, and the second time in the body of StartSignalWatcher. A proper way would be to add a `signal` parameter to the `handle` function, and to pass the catched signal to the handle, that it could use (instead of getting the value in line 94 from another catch)

pkg/odo/genericclioptions/runnable.go
```diff
@@ -77,6 +85,16 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		startTelemetry(cmd, err, startTime)
 	}
 	util.LogErrorAndExit(err, "")
+	// TODO: capture ctrl^z
+	captureSignals := []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, os.Interrupt, os.Kill}
```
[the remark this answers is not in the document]
valaparthvi: No reason really, except lack of research on my end! I assume since we cannot catch it, it doesn't make sense to call it. I'll remove it.

pkg/util/util.go
```diff
@@ -1407,18 +1407,14 @@ func copyDirWithFS(src string, dst string, fs filesystem.Filesystem) error {
 	return nil
 }
 
-// StartSignalWatcher watches for kill,interrupt etc signals and cleans up temp files and folders
-func StartSignalWatcher(handle func()) {
+// StartSignalWatcher watches for signals and handles the situation before exiting the program
+func StartSignalWatcher(watchSignals []os.Signal, handle func(receivedSignals os.Signal)) {
```
feloy, line 1411: This should be `receivedSignal`, as we receive one signal at a time

valaparthvi:
@feloy @dharmit I would appreciate a review. All the changes have been addressed.

dharmit (commented):
/approve

PR #4964, opened 2021-08-03
Support creating k8s built-in resources

### LINKED ISSUE #4685: Support using built-in resource types for inlined Kubernetes-typed components
/kind bug

**Operating System:** Win10, Linux

**Output of `odo version`:** odo v2.1.0 (8479050f6)

Well, I'm not saying I had a compelling business or technical case for doing this, but just to understand the technology, I tried adding some buit-in type defs as inlined kubernetes-typed components, like this:

```yaml
components:
  - name: myk8deploy
    kubernetes:
      inlined: |
        apiVersion: v1
        kind: Pod
        metadata:
          name: nginx
          labels:
            name: nginx
        spec:
          containers:
          - name: nginx
            image: nginx
            ports:
            - containerPort: 80
```

(Note there's no reference to this component from any of the `commands`.)

  
## Actual behavior

In the VSCode debug console I see: 

>  Failed to start component with name "in1". Error: Failed to create the component: failed to create service(s) associated with the component: could not find specified service/custom resource: Pod; please check the "kind" field in the yaml (it's case-sensitive)

and nothing matching the nginx pod in the namespace via `kubectl get all`.

In the debugger I see we step through:

```go
func GetCSV(client *kclient.Client, crd map[string]interface{}) (string, olm.ClusterServiceVersion, error) {
	cr := crd["kind"].(string)
	csvs, err := client.ListClusterServiceVersions()
```

fail to find a CSV match for the Pod type, and exit.

## Expected behavior

I'd expect this definition to have been passed to "kubectl apply".   odo shouldn't only be creating operator-backed resources.

See above

## LINK

This could be grouped with this issue: https://github.com/openshift/odo/issues/4159

dharmit:
**What type of PR is this?**
/kind feature

Adds support for creating k8s built-in resource types when user has put them in devfile.

Fixes #4685

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test

- [ ] Documentation 

- [x] Update changelog

* Suggest using a fresh minikube cluster without OLM for testing this PR.
* Create a nodejs odo component using https://github.com/odo-devfiles/nodejs-ex
* Download this [devfile](https://github.com/dharmit/odo/blob/fix-4685-rebased/tests/examples/source/devfiles/nodejs/devfile-with-pod.yaml) which is added to this PR for tests.
* Run `odo service list`. It should show `EtcdCluster/etcdcluster` even if that Operator is not installed on the cluster. This wasn't working before this PR. Make sure you have not installed the Operator in the past on the cluster you're using. If you did, please get a fresh cluster. 
* Try doing `odo push`. It will show an error message asking user to install the Operator that provides the CRD in question. In this case, `EtcdCluster`.
* Try deleting this service using `odo service delete`. It will be deleted from devfile.
* Run `odo push` and it should create the nginx pod in the devfile.


pkg/service/service.go
```diff
@@ -356,51 +334,30 @@ func ListOperatorServices(client *kclient.Client) ([]unstructured.Unstructured,
...
 
 // GetGVRFromCR parses and returns the values for group, version and resource
 // for a given Custom Resource (CR).
-func GetGVRFromCR(cr *olm.CRDDescription) (group, version, resource string, err error) {
+func GetGVRFromCR(cr *olm.CRDDescription) (string, string, string) {
+	var group, version, resource string
 	version = cr.Version
 
 	gr := strings.SplitN(cr.Name, ".", 2)
-	if len(gr) != 2 {
```
mik-dass, line 396: Please add a check before using the indexes to avoid panic

dharmit: I removed it because a CR would always be of the form `<api>/<version>`. It's a check that didn't serve us anything.

mik-dass: If someone sets the CR name to a invalid one before the function call, this would panic and stop the execution of the program immediately. It will lead to a bad user experience.

dharmit: Can you think of a scenario or try a scenario in the odo workflow that would cause panic? I don't mind adding this check again if that's the case. It's basically just undoing what I removed from the code.

mik-dass: > Can you think of a scenario or try a scenario in the odo workflow that would cause panic?

Any human error and failure to detect the problem on the CI can cause it to panic.

dharmit: > Any human error and failure to detect the problem on the CI can cause it to panic.

Can you please point an odo command that could panic things here?

mik-dass: > Can you please point an odo command that could panic things here?

All the commands and code points which call this function are potential candidates.

dharmit: Before this code gets executed as a result of any odo command, we go through https://github.com/openshift/odo/blob/6b03998f5b6b4edb0835d796a90b229e790f7317/pkg/service/service.go#L574

which ensures that if something's not following the `<operator-name>/<cr-name>` convention, odo stops further execution. That's why I don't think odo would hit a panic at this point.

pkg/odo/cli/service/operator_backend.go
```diff
@@ -68,50 +74,26 @@ func (b *OperatorBackend) ValidateServiceCreate(o *CreateOptions) (err error) {
...
 		if err != nil {
 			return err
 		}
 
-		// all is well, let's populate the fields required for creating operator backed service
-		b.group, b.version, b.resource, err = svc.GetGVRFromOperator(csv, b.CustomResource)
-		if err != nil {
-			return err
-		}
+		gvk := svc.GetGVKFromUnstructured(u)
```
kadel, line 82: Why not use function that is already in k8s apimachinery https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured.GroupVersionKind ?

pkg/service/service.go
```diff
@@ -1120,59 +1029,79 @@ func UpdateServicesWithOwnerReferences(client *kclient.Client, k8sComponents []d
...
-	metadata, ok := crd["metadata"].(map[string]interface{})
-	if !ok {
-		return "", false
+func isLinkResource(kind string) bool {
+	return kind == "ServiceBinding"
+}
+
+// GetGVKFromUnstructured takes an unstructured struct and returns group, version and kind information as strings
+// It assumes group to be "core" for cases where metadata.apiVersion can't be split using forward slash, e.g. 'v1'
+func GetGVKFromUnstructured(u unstructured.Unstructured) schema.GroupVersionKind {
```
kadel, line 1048: this version already exists in https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured.GroupVersionKind

pkg/service/service.go
```diff
@@ -1120,59 +1029,79 @@ func UpdateServicesWithOwnerReferences(client *kclient.Client, k8sComponents []d
...
+		Kind:    kind,
 	}
-	return name, true
 }
 
-func isLinkResource(kind string) bool {
-	return kind == "ServiceBinding"
-}
+// GetGVRFromUnstructured returns GVR from unstructured data
+func GetGVRFromUnstructured(client *kclient.Client, u unstructured.Unstructured) (*meta.RESTMapping, error) {
```
kadel, line 1067: This function has misleading name and comment. Name suggests that it should return GVR, but it returns RestMapping

pkg/service/service.go
```diff
@@ -1120,59 +1029,79 @@ func UpdateServicesWithOwnerReferences(client *kclient.Client, k8sComponents []d
...
 	}
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(dc))
+
+	return mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+}
 
-	// add labels to the CRD before creation
-	d.AddComponentLabelsToCRD(labels)
+// getOperatorGVRList creates a slice of GVRs that are provided by Operators (CSV)
+func getOperatorGVRList(client *kclient.Client) ([]meta.RESTMapping, error) {
```
kadel, line 1085: similar to above function, this seems to be doing something else than what name suggests

kadel:
@dharmit Are you planning to add some integration tests?

pkg/odo/cli/service/operator_backend.go
dharmit: Because I didn't check that it exists already. 😕

pkg/service/service.go
mik-dass: > Before this code gets executed as a result of any odo command, we go through

If someone in future tries to call the function outside of this flow and doesn't check against `SplitServiceKindName()`, it might panic. It's always better to prevent such issues before hand. Anyways let's keep it for now.

dharmit:
> @dharmit Are you planning to add some integration tests?

No. It's not a user facing change, so I'm not going to add them.

pkg/odo/cli/service/operator_backend.go
dharmit: https://github.com/openshift/odo/pull/4964/commits/8270bb639dbd38589082f73b590955c489cdddd7

pkg/service/service.go
dharmit: https://github.com/openshift/odo/pull/4964/commits/8270bb639dbd38589082f73b590955c489cdddd7

pkg/service/service.go
dharmit: https://github.com/openshift/odo/pull/4964/commits/8270bb639dbd38589082f73b590955c489cdddd7

pkg/service/service.go
dharmit: https://github.com/openshift/odo/pull/4964/commits/8270bb639dbd38589082f73b590955c489cdddd7

kadel:
> No. It's not a user facing change, so I'm not going to add them.

:astonished: How is this not user facing change? This is new feature.
It needs tests!

dharmit:
> > No. It's not a user facing change, so I'm not going to add them.
> 
> astonished How is this not user facing change? This is new feature.
> It needs tests!

Isn't this something we're adding for the advanced users who want to modify devfile by hands? AFAIU, we don't want to talk about it to normal users who shouldn't have to deal with devfiles. That's why, IMO, it isn't a user facing change.

kadel:
> Isn't this something we're adding for the advanced users who want to modify devfile by hands? AFAIU, we don't want to talk about it to normal users who shouldn't have to deal with devfiles. That's why, IMO, it isn't a user facing change.


Unless you want to say that we don't care if is not working, than it needs tests.

pkg/service/service.go
```diff
@@ -1120,59 +1029,59 @@ func UpdateServicesWithOwnerReferences(client *kclient.Client, k8sComponents []d
...
-
 func isLinkResource(kind string) bool {
 	return kind == "ServiceBinding"
 }
 
-// createOperatorService creates the given operator on the cluster
-// it returns the CR,Kind and errors
-func createOperatorService(client *kclient.Client, d *DynamicCRD, labels map[string]string, ownerReferences []metav1.OwnerReference) (string, string, error) {
-	cr, csv, err := GetCSV(client, d.OriginalCRD)
+// GetRestMappingFromUnstructured returns GVR from unstructured data
```
kadel, line 1046: Comment still says something else then function name

pkg/service/service.go
kadel: Comment still says something else then function name

kadel:
I'm having issues create simple java-springboot  on minikube component using code from this branch

Full scenario:
```
▶ minikube start
😄  minikube v1.22.0 on Arch 
✨  Automatically selected the kvm2 driver. Other choices: virtualbox, ssh
👍  Starting control plane node minikube in cluster minikube
🔥  Creating kvm2 VM (CPUs=12, Memory=12288MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.21.2 on Docker 20.10.6 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

▶ minikube addons enable ingress
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
    ▪ Using image k8s.gcr.io/ingress-nginx/controller:v0.44.0
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
🔎  Verifying ingress addon...
🌟  The 'ingress' addon is enabled

▶ git clone https://github.com/spring-projects/spring-petclinic

Cloning into 'spring-petclinic'...
remote: Enumerating objects: 8598, done.
remote: Total 8598 (delta 0), reused 0 (delta 0), pack-reused 8598
Receiving objects: 100% (8598/8598), 7.28 MiB | 9.03 MiB/s, done.
Resolving deltas: 100% (3267/3267), done.

▶ cd spring-petclinic 

▶ odo project create petclinic

 ✓  Project 'petclinic' is ready for use
 ✓  New project created and now using project: petclinic

▶ odo create java-springboot petclinic

Devfile Object Validation
 ✓  Checking devfile existence [29400ns]
 ✓  Creating a devfile component from registry: DefaultDevfileRegistry [44740ns]
Validation
 ✓  Validating if devfile name is correct [25340ns]

▶ odo url create  --port 8080 --host `minikube ip`.nip.io

 ✓  URL 8080-tcp created for component: petclinic

▶ odo push

Validation
 ✓  Validating the devfile [39050ns]

Creating Services for component petclinic
 ✗  Failed to start component with name "petclinic". Error: Failed to create the component: failed to create service(s) associated with the component: could not find specified operator

```

The same scenario with odo build from main branch works without any problems.

dharmit:
> I was able to push it just fine on OpenShift cluster. I'll try on minikube and ping back.

Yep. It fails. I'll take a look. But I find it weird that it works fine on OpenShift. 😕

dharmit:
Digging for this issue led me to a couple of problems in our code.

Firstly, in the `adapter.go`, we parse Kubernetes components in the devfile and send them to service package for creation: https://github.com/openshift/odo/blob/7d206a9d61263e528beb0c2b0f20a8800ce50be8/pkg/devfile/adapters/kubernetes/component/adapter.go#L182-L192

There needs to be a check for size of `k8sComponents` before doing anything else in that function: https://github.com/openshift/odo/blob/7d206a9d61263e528beb0c2b0f20a8800ce50be8/pkg/service/service.go#L946-L950

Secondly, there could still be a situation where we hit this problem.
1. Devfile doesn't have an Operator backed service, but does have a Pod, Deployment, etc. and odo has to create it. In this case, odo will still fail with the same error at https://github.com/openshift/odo/blob/7d206a9d61263e528beb0c2b0f20a8800ce50be8/pkg/service/service.go#L954-L957
2. User A is working on a cluster without OLM setup; they clone a git repo with devfile which has an Operator backed service in it. odo would fail again at the same place.

We could either address this in the code such that odo works fine when a cluster doesn't have OLM installed. Or we could make error and docs more explicit about informing the users to set up OLM when working with odo. The latter still leaves those users dry who are working on a shared k8s cluster and can't get OLM installed easily/quickly.

dharmit:
> We could either address this in the code such that odo works fine when a cluster doesn't have OLM installed. Or we could make error and docs more explicit about informing the users to set up OLM when working with odo. The latter still leaves those users dry who are working on a shared k8s cluster and can't get OLM installed easily/quickly.

Highly recommend reading the following comments first:
- https://github.com/openshift/odo/pull/4964#issuecomment-895309193
- https://github.com/openshift/odo/pull/4964#issuecomment-896557897
- https://github.com/openshift/odo/pull/4964#issuecomment-897347311

At the moment, my PR is modifying the `PushServices` function in `service` package. So far, Kubernetes inlined components meant only Operator backed services till we came across this issue of creating a Pod.

At code level, should odo try to differentiate between Operator CRDs and those objects which aren't based on Operators? How? Could it be based on an api group? Or should we individually check for each Kubenetes inlined component if the GVK is supported by the underlying cluster? Basing our code on `isCSVSupported` might have been helpful in the past, but the rest of the code written with CSV in mind is likely being problematic here.

@kadel @feloy @mik-dass @valaparthvi what do you folks think?

kadel:
> Or we could make error and docs more explicit about informing the users to set up OLM when working with odo.

OLM can't be hard requirement for odo. If users are not doing anything with OLM (services) odo should not require it

kadel:
> At code level, should odo try to differentiate between Operator CRDs and those objects which aren't based on Operators? How? 

Why do you want to differentiate between those?

Odo should not care about what type the resource is. It is all just Kubernetes resources.


> Basing our code on `isCSVSupported` might have been helpful in the past, but the rest of the code written with CSV in mind is likely being problematic here.

Why do we need that function? For Resource creation odo doesn't need to check if the CSV is supported or not. If it really needs to do some sanity checks, it could check if the resource that it is trying to create is supported by cluster or not. 
(But if odo just errors out if the resource is not supported, without any extra logic, nice message or cleanup, it can just let `Create` call to error out. The result will be the same )

pkg/service/service.go
```diff
@@ -44,78 +51,28 @@ const ServiceKind = "app.kubernetes.io/service-kind"
...
-	}
-
-	csv, err := doesCRExist(cr, csvs)
-	if err != nil {
-		return cr, olm.ClusterServiceVersion{},
-			fmt.Errorf("could not find specified service/custom resource: %s; please check the \"kind\" field in the yaml (it's case-sensitive)", cr)
-	}
-	return cr, csv, nil
+func (params servicePlanParameters) Len() int {
+	return len(params)
```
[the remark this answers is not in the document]
dharmit: That's rebase leftover. We don't need this and similar functions.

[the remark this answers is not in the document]
dharmit: Or maybe not. I'm seeing bunch of Service Catalog code even in the main branch. 😕 😞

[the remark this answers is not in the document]
dharmit: I see some of it going back to `common_link.go`. I'll refrain from touching those since I think @mohammedzee1000 is working on it already.

dharmit:
>     1. > @dharmit is this still a work-in-progress? I was able to create a service from the YAML and push it, but I did not see any pod come up on the minikube cluster. Do I need to enable OLM on minikube for this to work? ([Support creating k8s built-in resources #4964 (comment)](https://github.com/openshift/odo/pull/4964#issuecomment-906312464))
>        Oh I see, you need to enable OLM in order to make this work. Do you think we should print a warning saying that it might not work if OLM is not installed?

No. I have made changes such that even if OLM is not installed, the devfile described in the PR description should work fine. 

>     2. I fail to understand what value will people get out of this feature.

Refer the issue it's trying to fix.

>     3. > The way I have gone about it in the code is to put only one label on Kubernetes inlined component - app.kubernetes.io/managed-by: odo.([Support using built-in resource types for inlined Kubernetes-typed components #4685 (comment)](https://github.com/openshift/odo/issues/4685#issuecomment-891788502))
>        
>        
>        This would mean that it removes any other labels that were provided by the user, I don't think that is the right way to go, we should preserve the user provided labels, in case they are required to be filtered.

No, it doesn't mean that it will remove any existing labels. We want to simply append odo specific labels to existing definitions. However, technically, it was indeed removing the labels because that's what `SetLabels` method I have used in this PR does. Anyway, I have fixed that in https://github.com/openshift/odo/pull/4964/commits/db5dcfd0f6cb5d3506b14563a637006dc7202804.

>     4. The services created from K8s components doesn't show up in `odo service list` and there's no other way except checking `kubectl get <resource_kind>`. Is this something you're going to work on later?

This is expected behaviour. We don't want to show kubernetes resources created this way to show up in odo outputs.

kadel:
I'm not able to make this work.
Even using the devfile mentioned in the PR descriptoin no Pod is created. `odo push` just says ` ✓  Services are in sync with the cluster, no changes are required`

valaparthvi:
> I'm not able to make this work.
> Even using the devfile mentioned in the PR descriptoin no Pod is created. `odo push` just says ` ✓ Services are in sync with the cluster, no changes are required`

It worked fine for me. Both the pods are running fine for me.

dharmit:
> I'm not able to make this work.
> Even using the devfile mentioned in the PR descriptoin no Pod is created. `odo push` just says ` ✓ Services are in sync with the cluster, no changes are required`

It's weird why this worked locally for me. Anyway, https://github.com/openshift/odo/pull/4964/commits/d640acb038867f23479227efa651a61829e32a24 should address it.

dharmit:
> > I'm not able to make this work.
> > Even using the devfile mentioned in the PR descriptoin no Pod is created. `odo push` just says ` ✓ Services are in sync with the cluster, no changes are required`
> 
> It worked fine for me. Both the pods are running fine for me.

On a fresh minikube, I got only the component pod working, not the nginx pod. It's probably because OLM was enabled on your cluster. PTAL https://github.com/openshift/odo/pull/4964/commits/d640acb038867f23479227efa651a61829e32a24.

dharmit:
> It's weird why this worked locally for me.

It's because yesterday I was developing this PR against CRC on which OLM is enabled by default.

valaparthvi:
@dharmit Can you resolve the merge conflicts here>?

pkg/service/service.go
dharmit: Addressed.

dharmit:
Gave up on rebase. Opened https://github.com/openshift/odo/pull/5107. Closing this in favour of it.

/close

PR #4985, opened 2021-08-12
Cleanup namespaces

### LINKED ISSUE #4467: tests are not cleaning up projects/namespaces.
Integrations and e2e tests are not properly cleaning up project/namespaces.
Currently, there are 77 projects with random 10 letter name (those are created by our integration tests)

```
kubectl get projects -o=custom-columns=NAME:.metadata.name | grep -E "^[a-z]{10}$" | wc -l
77
```
The above issue is tracked with [issue 4502](https://github.com/openshift/odo/issues/4502)

On top of that, we have 90 projects starting with `odo-pr-build`
```
kubectl get projects -o=custom-columns=NAME:.metadata.name | grep -E "odo-pr-build" | wc -l
90
````

Most of those project are older than 1 day (`kubectl get projects -o=custom-columns=NAME:.metadata.name,CREATED:.metadata.creationTimestamp`) so they are not from currently running tests .

### PR DESCRIPTION
**Pull request by anandrkskd**
**What type of PR is this?**

/kind cleanup

This PR adds a script to cleanup namespace that are `x` days old from the cluster, as well as adds helper functions to support the cleanup process

Fixes #4467

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] Update changelog

### CODE REVIEW

**Comment by dharmit** [comment]
Please update the Changelog.md.

**Comment by dharmit** [comment]
@anandrkskd is this still WIP?

File: scripts/CleanCluster.sh [hunk 1]
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env sh
+
```
dharmit, line 2: Can you please add a brief comment indicating the purpose of the code in this file? Also, should the name reflect that it's a script about cleaning namespaces on the cluster? `CleanCluster` kinda sounds like it's going to clean the cluster itself.

scripts/CleanCluster.sh
```diff
@@ -0,0 +1,38 @@
...
+if [[ $CLUSTER_TYPE == "PSI" ]]; then
+    ￼ #PSI cluster login
+    ￼ oc login -u kubeadmin -p ${OCP4X_KUBEADMIN_PASSWORD} --insecure-skip-tls-verify ${OCP4X_API_URL}
+else
+    ￼ # Login to IBM Cloud using service account API Key
+    ￼ ##ibmcloud login --apikey $IBMC_OCP47_APIKEY -a cloud.ibm.com -r eu-de -g "Developer CI and QE"
```
dharmit, line 8: Why have you commented this? For the time being, it's anyway going to run on PSI, right? When we shift to IBM Cloud, we would have to remember to uncomment this part so that things work on it. Hence, suggesting to not keep it commented.

scripts/CleanCluster.sh
```diff
@@ -0,0 +1,38 @@
...
+    ￼
+    ￼ # Login to cluster in IBM Cloud using cluster API key
+    ￼ ##oc login --token=$IBMC_OCLOGIN_APIKEY --server=$IBMC_OCP47_SERVER
```
dharmit, line 11: Same as above comment. Can we keep this uncommented?

scripts/CleanCluster.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+PROJECT_AND_TIME=$(kubectl get projects -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.creationTimestamp} {"\n"}{end}')
```
dharmit, line 14: What's the purpose of this variable? Do you think a brief comment would help?

dharmit, line 14:
```suggestion
PROJECT_AND_TIME=$(kubectl get projects -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.creationTimestamp} {"\n"}{end}') | grep -v '^openshift\|^kube\|^default'
```

scripts/CleanCluster.sh
```diff
@@ -0,0 +1,38 @@
...
+
+for PROJECT in ${PROJECT_AND_TIME}; do
+    IFS='|'
+    read -ra ADDR <<<"$PROJECT"
+    echo "INFO: Project="${ADDR[0]}, "Date=" ${ADDR[1]}
+    if [[ ${ADDR[0]} == "openshift"* || ${ADDR[0]} == "kube"* || ${ADDR[0]} == "default" ]]; then
+        echo "Skipped"
```
dharmit, line 21: If you use the `grep -v` suggestion above, you can get rid of this `if` block.

tests/helper/helper_oc.go
```diff
@@ -766,3 +767,68 @@ func (oc OcRunner) CreateSecret(secretName, secretPass, project string) {
 func (oc OcRunner) GetSecrets(project string) string {
 	return GetSecrets(oc.path, project)
 }
+
+func (oc OcRunner) AddSecret(comvar CommonVar) {
```
dharmit, line 771: Please add a comment describing what this function is about.

dharmit (changes requested):

scripts/CleanCluster.sh
anandrkskd: understood

tests/helper/helper_oc.go
anandrkskd: updated with a comment

scripts/CleanCluster.sh
anandrkskd: updated in the new script

scripts/CleanCluster.sh
anandrkskd: comment added

scripts/CleanCluster.sh
anandrkskd: Comment added, also updated the name of the script

scripts/clean-old-namespace-from-cluster.sh
```diff
@@ -0,0 +1,43 @@
...
+    ￼ oc login --token=$IBMC_OCLOGIN_APIKEY --server=$IBMC_OCP47_SERVER
+fi
+
+# PROJECT_AND_TIME var will contain namespace and date with time seperated with `|`
+# PROJECT_AND_TIME doesn't contain openshift/ibm/kube namespace
+# eg. cmd-push-test157kgb|2021-08-17T12:40:20Z
+PROJECT_AND_TIME=$(kubectl get projects -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.creationTimestamp} {"\n"}{end}' | grep -v '^openshift\|^kube\|^default\|^ibm\|^calico\|^tigera')
+
+for PROJECT in ${PROJECT_AND_TIME}; do
+    IFS='|'
```
dharmit, line 23: Where/how is this value being used?

scripts/clean-old-namespace-from-cluster.sh
```diff
@@ -0,0 +1,43 @@
...
+    read -ra ADDR <<<"$PROJECT"
```
dharmit, line 24: Thanks for adding `eg. cmd-push-test157kgb|2021-08-17T12:40:20Z` above. Can you add comments within this function that would help understand what the value of `ADDR` will be based on that example?

Similarly, for other variables below like `datetime`, `dtSec`, `taSec` as well? I'm asking this only because when someone other than you takes a look at this code a few weeks/months later, these comments would make it really simpler for them to understand what changes need to be made.

dharmit (changes requested):

dharmit:
@feloy I've requested for your review on this because we need to help @anandrkskd get this merged sooner rather than later. Moving to IBM Cloud is not going to be worth much if the problem addressed by this PR isn't fixed.

scripts/clean-old-namespace-from-cluster.sh
anandrkskd: understood will add more comments to the code.

scripts/clean-old-namespace-from-cluster.sh
anandrkskd: @dharmit  are you asking about IFS value?

feloy, line 23: I would prefer a one-line to avoid modifying the global IFS value (and you don't need anymore to reset it to a "space" value later):

```
IFS='|' read -ra ADDR <<<"$PROJECT"
```

Also, you could use more descriptive variables, without an array:

```
IFS='|' read -r PRJ TIME <<<"$PROJECT"
```

scripts/clean-old-namespace-from-cluster.sh
```diff
@@ -0,0 +1,43 @@
...
+    ￼ ibmcloud login --apikey $IBMC_OCP47_APIKEY -a cloud.ibm.com -r eu-de -g "Developer CI and QE"
+    ￼
+    ￼ # Login to cluster in IBM Cloud using cluster API key
+    ￼ oc login --token=$IBMC_OCLOGIN_APIKEY --server=$IBMC_OCP47_SERVER
+fi
+
+# PROJECT_AND_TIME var will contain namespace and date with time seperated with `|`
+# PROJECT_AND_TIME doesn't contain openshift/ibm/kube namespace
+# eg. cmd-push-test157kgb|2021-08-17T12:40:20Z
+PROJECT_AND_TIME=$(kubectl get projects -o jsonpath='{range .items[*]}{.metadata.name}{"|"}{.metadata.creationTimestamp} {"\n"}{end}' | grep -v '^openshift\|^kube\|^default\|^ibm\|^calico\|^tigera')
```
feloy, line 20: I understood we were running on a cluster in common with other teams. Won't we delete other teams namespaces this way?

tests/helper/helper_oc.go
```diff
@@ -766,3 +767,71 @@ func (oc OcRunner) CreateSecret(secretName, secretPass, project string) {
...
+		newYaml := strings.Replace(yaml, "openshift-config", comvar.Project, -1)
+		filename := fmt.Sprint(RandString(4), ".yaml")
+		f, err := os.Create(filename)
+		if err != nil {
+			fmt.Println(err)
+		}
+		_, err = f.WriteString(newYaml)
+		if err != nil {
+			fmt.Println(err)
+			f.Close()
```
feloy, line 789: Please use defer f.Close() instead

tests/helper/helper_oc.go
```diff
@@ -766,3 +767,71 @@ func (oc OcRunner) CreateSecret(secretName, secretPass, project string) {
...
+		}
+		err = f.Close()
```
feloy, line 791: You could use ioutil.WriteFile instead (https://gobyexample.com/writing-files)

scripts/clean-old-namespace-from-cluster.sh
dharmit: Yes.

anandrkskd: it's `Input Field Separators` do we want a comment for that too? I will be using what @feloy suggested in one of his comment https://github.com/openshift/odo/pull/4985#discussion_r691866953

dharmit: Is it some kind of built-in variable? I don't know. Might help to comment it even when updating as Philippe suggested. I'll leave the decisoin to you. :)

tests/helper/helper_oc.go
anandrkskd: Is this what you are suggesting? 
```
filename := fmt.Sprint(RandString(4), ".yaml")
d1 := []byte(newYaml)
err := ioutil.WriteFile(filename, d1, 0644)
if err != nil {
     panic(err)
}		
Cmd(oc.path, "apply", "-f", filename).ShouldPass()
```

feloy: Yes, right

anandrkskd: done

dharmit (commented):
/approve

scripts/clean-old-namespace-from-cluster.sh
anandrkskd: yes we will remove the namespace for other teams too, but this was discussed on the [issue](https://github.com/openshift/odo/issues/4467#issuecomment-893363050) as this will be a common solution for every team.

feloy: Cluster will be in trouble each time someone in a team will install something in a new namespace. It is not a very robust solution.

I would prefer to use a prefix for each team to prefix the namespaces to delete, and grep on this prefix, instead of invert-grepping on the namespaces to remove.

@rnapoles-rh @kadel WDYT?

anandrkskd: +1 feloy I suggested the same on this [comment](https://github.com/openshift/odo/issues/4467#issuecomment-891777443)

mohammedzee1000: I guess the deletion of projects based on time still stands, and this is more a discussion on which projects will go under the knife in the first place. 

I guess we could have a prefix on projects which can be deleted and only put those projects through the age check before removing them. 

This also means we have to communicate to every team beforehand about this when they are onboarded into the cluster and make it clear that if a prefix is not set to the tests, then it may not get cleaned up resulting in not only leaks in resources, but also usage costs

mohammedzee1000: Also, namespaces cannot be renamed, at least not easily, so the name with prefix should be set during the creation

kadel: > I would prefer to use a prefix for each team to prefix the namespaces to delete, and grep on this prefix, instead of invert-grepping on the namespaces to remove.

This could lead to other problems. For example, we don't have a mechanism to enforce prefixes. So if someone creates namespace without correct prefix, it will be there forever. 

I think that invert-grepping can be ok, but it needs to be clearly communicated to everyone who is using the cluster that this mechanism is there.

anandrkskd: > This could lead to other problems. For example, we don't have a mechanism to enforce prefixes. So if someone creates namespace without correct prefix, it will be there forever.

but those namespace will be created by someone manually, so it may have some importance to it for that time,
and we can add an additional layer of check for non prefix namespace which are more then 20/30 days old will be deleted?

feloy: The Hierarchical namespaces (https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/) could help resolve these multi-tenancy problems

mohammedzee1000: > The Hierarchical namespaces (https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/) could help resolve these multi-tenancy problems

But is this something that can be done in the background without significant changes to testing. Also, prevention of deletion of constant namespaces can be handled using an exclusion list akin to gitignore

Basically
```
for ns in namespaces
{
   if ns != kube* || ns != openshift* || ns not in exclusion-list {
     if ns is too old 
     {
         delete it
     }
  }
}
```

feloy: The exclusion list will be difficult to maintain, and subject to some errors: we will probably remember to add a namespace to the list only after the first time it is deleted, which can be painful.

mohammedzee1000: > The exclusion list will be difficult to maintain, and subject to some errors: we will probably remember to add a namespace to the list only after the first time it is deleted, which can be painful.

This is fine, the reverse of this is an improperly named ns being left behind(with all of its resources) which is going to increase operational costs. If a test user forgets to add an ns, then it is his fault, not ours.

I agree that communication should be clear and people should be aware cleanup happens and how to avoid it.
An exclusion list is not that hard to maintain. It should be a simple file with one project name per line. Also ignoring system namespaces like kube* and openshift* should be built into the script and not in the exclusion list

mohammedzee1000: As discussed in cabal, we are going for labels. I would recommend a label like 'testing' to make script more generic and reusable and generic where script label indicates project was created for testing.

Filter projects with testing label and delete the old ones.

mohammedzee1000: wdyt @feloy

feloy: I agree with this. Depending when what was discussed with other teams, perhaps also add a label defining the team, to filter on this label too, something similar to: `team=odo&app=testing`

mohammedzee1000: Yep, and team name can be an argument to script

anandrkskd: Do we want to add `app=testing` to the script as well?

mohammedzee1000: I guess @feloy is pointing to the testing label I mentioned. The form is irrelevant. The important thing is we only destroy projects that

1. Have a `testing` label. This can be `app=testing`
2. Have a `team=blah` label where `blah` can be passed to the script.
3. Older than X
This assumes that only projects created by tests have the `testing` label

anandrkskd: Understood, I have made the changes.

feloy:
Lots of tests are failing

/test v4.8-integration-e2e

mohammedzee1000:
Are we going to have namespace/project labelling in a separate PR?

Basically `kubectl label namespace <ns>` and `oc label namespace <ns>`

feloy:
> Are we going to have namespace/project labelling in a separate PR?
> 
> Basically `kubectl label namespace <ns>` and `oc label namespace <ns>`

I think it is already present in the PR:

tests/helper/helper_kubectl.go:163 (and same in oc helper)
```
Cmd("kubectl", "label", "namespace", projectName, "app=test", "team=odo").ShouldPass() // labels added for cleanup purpose
```

scripts/clean-old-namespace-from-cluster.sh
```diff
@@ -0,0 +1,44 @@
...
+    IFS='|' read -r PRJ TIME <<<"$PROJECT" # seperate the Namespace and time of creation using IFS(Input Field Seperators) value `|`
+
+    CONFIGMAP=$(kubectl get configmaps -n $PRJ -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}' | grep config-map-for-cleanup) # check if PROJECT contains configmap named 'config-map-for-cleanup'
+                                                                                                                                    # if not then the value of CONFIGMAP will be empty
+    if [[ ! -z $CONFIGMAP ]]; then
+
+        TEAMANDTYPE=$(kubectl get configmaps $CONFIGMAP -n $PRJ -o jsonpath='{.data.team}{"|"}{.data.type}') # fetch team and tyoe(testing) data from the configmap
+        IFS='|' read -r TEAM TYPE <<<"$TEAMANDTYPE"                                                          # seperate the TEAM and TYPE of creation using IFS(Input Field Seperators) value `|`
+
+        if [[ $TYPE -eq "testing" ]]; then # check if type if testing
```
feloy, line 29: I think we also want to filter on the team value

anandrkskd: understood, adding it ASAP

tests/helper/helper_kubectl.go
```diff
@@ -160,9 +160,11 @@ func (kubectl KubectlRunner) CreateRandNamespaceProject() string {
 func (kubectl KubectlRunner) createRandNamespaceProject(projectName string) string {
 	fmt.Fprintf(GinkgoWriter, "Creating a new project: %s\n", projectName)
 	Cmd("kubectl", "create", "namespace", projectName).ShouldPass()
+	Cmd("kubectl", "label", "namespace", projectName, "app=test", "team=odo").ShouldPass() // labels added for cleanup purpose
```
feloy, line 163: You don't need this anymore

feloy (commented):
Just an unused line to remove, and that seems ok to me

tests/helper/helper_kubectl.go
anandrkskd: ohh yes, Thanks @feloy

anandrkskd: Updated!

tests/helper/helper_oc.go
```diff
@@ -628,9 +630,9 @@ func (oc OcRunner) CreateRandNamespaceProjectOfLength(i int) string {
 
 func (oc OcRunner) createRandNamespaceProject(projectName string) string {
 	fmt.Fprintf(GinkgoWriter, "Creating a new project: %s\n", projectName)
-	session := Cmd("odo", "project", "create", projectName, "-w", "-v4").ShouldPass().Out()
-	Expect(session).To(ContainSubstring("New project created"))
+	session := Cmd("oc", "new-project", projectName).ShouldPass().Out()
```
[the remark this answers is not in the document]
anandrkskd: Updated!

PR #4986, opened 2021-08-12
Remove empty docs; add when ready

dharmit:
$subject

ping @valaparthvi can you help with this quick one?

/kind documentation


valaparthvi (changes requested):
What about the quickstart doc? It is not useful as such, are we keeping it?
Other changes look fine.

dharmit:
> What about the quickstart doc? It is not useful as such, are we keeping it?
> Other changes look fine.

I kept it because your PR's addressing it. But it makes sense to remove it. I'll remove that as well.

valaparthvi (approved):

valaparthvi:
>I kept it because your PR's addressing it. But it makes sense to remove it. I'll remove that as well.

Since you removed other files such as cluster-setup that were being addressed by open PRs, I suggested we remove quickstart as well.

valaparthvi:
/approve 
Minor change, hence approving.

PR #4987, opened 2021-08-12
Removes blog entry since it wasn't useful

dharmit:
**What type of PR is this?**

$subject


dharmit:
I'll just replace it with a relevant blog and then remove the current unhelpful one.

/close

PR #5032, opened 2021-09-01
Fix clusterwide permission issues with odo link and run K8s tests with a developer user

### LINKED ISSUE #4965: odo can't create link using  client-side service binding library if user doesn't have permissions to list cluster wide resources
Original epic: https://github.com/openshift/odo/issues/4543

It looks like service binding library is trying to get cluster-wide resources.
This is especially problematic on OpenShift as normal cluster users usually don't have permissions to do so.
If the code base is supposed to be useful as a library it should avoid calls that require special privileges.

Issue reported to SBO: https://github.com/redhat-developer/service-binding-operator/issues/1003

/priority high
/kind bug

valaparthvi:
**What type of PR is this?**
/kind bug

This PR attempts at fixing the permission issues caused while linking without the SBO.

Fixes #4965 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

1. Login as a non-admin user or someone who doesn't have permission to access CRD.
2. Create a comp and push it.
3. Create another comp and push it.
4. Link comp1 to comp2 and push it: the push should be done successfully.

TODO: Have to test this with service.


dharmit:
> * Login as a non-admin user or someone who doesn't have permission to access CRD.
> 
> * Create a comp and push it.
> 
> * Create another comp and push it.
> 
> * Link comp1 to comp2 and push it: the push should be done successfully.

I see the Secret getting created after `odo link` as well as the env vars with info about Service belonging to c2 in c1's pod. But I think odo unlink doesn't delete the Secret. Please verify that part.

feloy:
I have made some tests on a fresh new minikube, as a User1 bound to this role (so, without access to CRD resource):

```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: odo-user
rules:
- apiGroups: [""] # “” indicates the core API group
  resources: ["*"]
  verbs: ["*"]
- apiGroups: ["apps"]
  resources: ["*"]
  verbs: ["*"]
- apiGroups: ["operators.coreos.com"]
  resources: ["clusterserviceversions"]
  verbs: ["*"]
- apiGroups: ["redis.redis.opstreelabs.in"]
  resources: ["*"]
  verbs: ["*"]
- apiGroups: ["networking.k8s.io"]
  resources: ["ingresses"]
  verbs: ["*"]
- apiGroups: ["route.openshift.io"]
  resources: ["*"]
  verbs: ["*"]

---

kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: odo-user-binding
subjects:
- kind: User
  name: user1 # Name is case sensitive
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole #this must be Role or ClusterRole
  name: odo-user
  apiGroup: rbac.authorization.k8s.io
```

All seems to work correctly. I can create and push a Redis service, link to it and push, unlink and push.

The secret attached to the link is deleted when unlinked.

valaparthvi:
@baijum @pedjak If any of you have time, can  you take a look at this implementation? I think I could use an opinion on this one. I haven't implemented our own `Context` to override the `CustomResourcesDefinition()` method because I think it might be a little more work than it is worth. But if you think overriding the `CustomResourcesDefinition()` method  might be a better option, I would like to know more about it.

valaparthvi:
/retest-required

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_odo/5032/pull-ci-openshift-odo-main-psi-kubernetes-integration-e2e/1433687869737668608

valaparthvi:

Reason - it did not run the commands I expected it to run, which is creating a developer user.

valaparthvi:

For some reason, the k8s tests are not taking the latest changes into account.

scripts/minikube-minishift-setup-env.sh
```diff
@@ -39,6 +39,62 @@ setup_kubeconfig() {
...
+  verbs: ["*"]
+- apiGroups: ["networking.k8s.io", "extensions"]
+  resources: ["ingresses"]
+  verbs: ["*"]
+- apiGroups: ["route.openshift.io"]
+  resources: ["*"]
+  verbs: ["*"]
+- apiGroups: ["apps.openshift.io"]
+  resources: ["*"]
+  verbs: ["*"]
```
valaparthvi, line 78: Does it make sense to add OpenShift resources to a minikube user? In a real world scenario, I think no minikube user will be concerned about OCP resources, but some of our tests rely on the user having access to these resources, and they might show a different output depending on which resource the user has access to.

scripts/minikube-minishift-setup-env.sh
```diff
@@ -39,6 +39,62 @@ setup_kubeconfig() {
...
+    cd $certdir
+    shout "Starting to create minikube developer"
+    openssl genrsa -out developer.key 2048
+    openssl req -new -key developer.key -out developer.csr -subj "/CN=developer/O=minikube"
+    openssl x509 -req -in developer.csr -CA ~/.minikube/ca.crt -CAkey ~/.minikube/ca.key -CAcreateserial -out developer.crt -days 500
+    kubectl config set-credentials developer --client-certificate=developer.crt --client-key=developer.key
+    kubectl config set-context developer-minikube --cluster=minikube --user=developer
+    kubectl create -f - <<EOF
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
```
feloy, line 54: Please comment that the intent of this role is to remove access to the CRD resource

scripts/minikube-minishift-setup-env.sh
feloy: odo fails if it does not have access to these openshift resources, even in minikube. It could be interesting to open an issue, to make odo ignore resources that it cannot access, instead of failing

scripts/minikube-minishift-setup-env.sh
```diff
@@ -81,6 +138,13 @@ case ${1} in
 
         # Create Operators for Operator tests
         sh $SETUP_OPERATORS
+
+        # Create a developer user if it is not created already and change the context to use it after the setup is done
+        kubectl config get-contexts developer-minikube
```
feloy, line 143: You could write it like:
```
kubectl config get-contexts developer-minikube || setup_minikube_developer
```

pkg/service/service_binding_builder.go
```diff
@@ -0,0 +1,198 @@
+package service
```
feloy, line 1: Please indicate that this code is a fork from the Service Binding Operator, and what are the conditions to start using the upstream library again (an issue number?)

scripts/minikube-minishift-setup-env.sh
prietyc123: > Does it make sense to add OpenShift resources to a minikube user? In a real world scenario, I think no minikube user will be concerned about OCP resources, but some of our tests rely on the user having access to these resources, and they might show a different output depending on which resource the user has access to.

I think its fine if we add permission for user to access those resources, as tests are relying on that. However I would like here from Team on this. @kadel @feloy @mohammedzee1000 Could you please share your thoughts. I have just pointed out to few people, however it would be great if any one of our team can share their thoughts.

scripts/minikube-minishift-setup-env.sh
```diff
@@ -81,6 +138,13 @@ case ${1} in
...
+        if [ $? -ne 0 ]; then
+          setup_minikube_developer
+        fi
+        kubectl config use-context developer-minikube
```
prietyc123, line 147: What is it confirming `developer` user already exits?

prietyc123 (changes requested):

scripts/minikube-minishift-setup-env.sh
prietyc123: @valaparthvi Could you please confirm if its a problem with odo or just odo test?

scripts/minikube-minishift-setup-env.sh
valaparthvi: Yeah, in case there is a running minikube cluster, we check if the context is created, if not, we create the context, user, and everything else, and then use the context.

scripts/minikube-minishift-setup-env.sh
valaparthvi: It is a problem with odo. See #5042 for example.

scripts/minikube-minishift-setup-env.sh
```diff
@@ -39,6 +39,63 @@ setup_kubeconfig() {
...
+    pwd=`pwd`
+    certdir=`mktemp -d`
+    cd $certdir
+    shout "Creating a minikube developer user"
+    openssl genrsa -out developer.key 2048
+    openssl req -new -key developer.key -out developer.csr -subj "/CN=developer/O=minikube"
+    openssl x509 -req -in developer.csr -CA ~/.minikube/ca.crt -CAkey ~/.minikube/ca.key -CAcreateserial -out developer.crt -days 500
+    kubectl config set-credentials developer --client-certificate=developer.crt --client-key=developer.key
+    kubectl config set-context developer-minikube --cluster=minikube --user=developer
+    # Create role and rolebinding to allow the user admin access to the cluster; this does not include access to CRD
```
feloy, line 52:
```suggestion
    # Create role and rolebinding to allow the user necessary access to the cluster; this does not include access to CRD
```

feloy (approved):

valaparthvi:
@feloy Can you add the lgtm label again?
Reason - https://github.com/openshift/odo/pull/5032/commits/94184a21af5dd357153ee9135e7b5908e7571cbe

pedjak:
@valaparthvi the change looks good

scripts/minikube-minishift-setup-env.sh
valaparthvi: > It could be interesting to open an issue, to make odo ignore resources that it cannot access, instead of failing

https://github.com/openshift/odo/issues/5043

dharmit: I don't understand why we would want access to OpenShift specific apiGroups for a user being created on minikube. But, it's not going to hurt, so I won't request a change.

pkg/service/service_binding_builder.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+		} else {
+			crd, err := service.CustomResourceDefinition()
+			if err != nil {
+				// If the user does not have permission to access CRD, ignore
+				if !kerrors.IsForbidden(err) {
+					requestRetry(ctx, collect.ErrorReadingCRD, err)
+					return
+				} else {
+					klog.V(4).Infof("Skipping the check for CRD, user does not have access")
```
kadel, line 80: Can we make this user normal facing message? Ideally a warning?
odo user should know that something like this was skipped without running odo with increased log level.

kadel (changes requested):

kadel:
It looks like the warning is showing twice

```
Creating Kubernetes resources for component java-springboot
 ⚠  Skipping the check for CRD, user does not have access
 ⚠  Skipping the check for CRD, user does not have access
```

But we can fix this in a separate PR.  For now this will make sure that user really notices this :-D

scripts/minikube-minishift-setup-env.sh
valaparthvi: That is my concern as well. But if we don't allow access to these apiGroups, we see a different output from that of a kubeadmin user,

valaparthvi:
> It looks like the warning is showing twice
> 
> ```
> Creating Kubernetes resources for component java-springboot
>  ✓  Waiting for component to start [1ms]
>  ⚠  Skipping the check for CRD, user does not have access
>  ⚠  Skipping the check for CRD, user does not have access
> ```
> 
> But we can fix this in a separate PR. For now this will make sure that user really notices this :-D

That's not a bug per se. Since we added the warning in both the overridden handler functions, it prints the warning when they are called.

kadel:
> That's not a bug per se. Since we added the warning in both the overridden handler functions, it prints the warning when they are called.

From user point of view this is bug!

Issue #4994, opened 2021-08-16
Capture OS info in telemetry


deboer-tim:
/kind feature

## Which functionality do you think we should add?
Telemetry from other tools typically includes additional context information including OS and location or timezone, e.g.:
```
  "context": {
    "locale": "en-US",
    "location": {
      "country": "TH"
    },
    "os": {
      "name": "Mac",
      "version": "20.6.0"
    },
    "timezone": "Asia/Bangkok"
    ...
  }
```
I'm not suggesting we need to send all of this, but the OS and locale or timezone would be useful.

## Why is this needed?
This should help us get a better idea of what platforms our users are on and where in the world they are, which in turn should help us focus our efforts.

## Acceptance Criteria

- [ ] Docs to help new folks understand debugging with telemetry code
- [ ] Capturing the info related to "Locale", "Timezone", and "Location".

deboer-tim:
I'll admit I do see a graph showing OS breakdown, but I can't tell where the data for this is coming from...

@valaparthvi

kadel:
> I'll admit I do see a graph showing OS breakdown, but I can't tell where the data for this is coming from...


OS info gets send as part of the "identify" event. Those properties get assigned to "People". You can see it in properties in  "People" section

deboer-tim:
Ok, there are no identify events in Segment at the moment which explains why I couldn't see it there. I also clicked in the properties as suggested, and about half didn't have an OS listed. Most other components are including the OS in every event as per the first comment, so I think this request is still valid.

valaparthvi:
>I also clicked in the properties as suggested, and about half didn't have an OS listed.

If you checked the properties for TRACK event, that might explain why you didn't see any OS listed, we only show this information in IDENTIFY event, which is when a new user joins in.

valaparthvi:
> > I also clicked in the properties as suggested, and about half didn't have an OS listed.
> 
> If you checked the properties for TRACK event, that might explain why you didn't see any OS listed, we only show this information in IDENTIFY event, which is when a new user joins in.

Segment only seems to be keeping data from the past day, maybe that's the reason why you didn't find any IDENTIFY event. @deboer-tim  Do you think it adds value to add OS and location information in TRACK events as well?

PS: I've added this issue to be discussed in the next grooming call.

kadel:
@valaparthvi 
Let's add location data into identify event.

> I also clicked in the properties as suggested, and about half didn't have an OS listed.

[...]

PR #5040, opened 2021-09-03
Send Identify event with every Track event

kadel:
**What type of PR is this?**

/kind bug

This will make sure that we identify every user,
not just those for which we created anonymousId.
User can already have anonymousId assigned by other tools like crc or
vscode plugins.
Before this odo did not send identify event for those users. This created
a situation where we were missing some information about those users.

More info about this in https://github.com/openshift/odo/issues/4994#issuecomment-902606879

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/segment/segment.go
+++ b/pkg/segment/segment.go
@@ -98,7 +98,7 @@ func (c *Client) Upload(data TelemetryData) error {
 	}
 
 	// obtain the user ID
-	userId, uerr := getUserIdentity(c, c.TelemetryFilePath)
+	userId, uerr := getUserIdentity(c.TelemetryFilePath)
 	if uerr != nil {
 		return uerr
 	}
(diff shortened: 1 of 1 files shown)


pkg/segment/segment.go
```diff
@@ -120,6 +120,18 @@ func (c *Client) Upload(data TelemetryData) error {
 		properties = properties.Set("error", data.Properties.Error).Set("error-type", data.Properties.ErrorType)
 	}
 
+	// send the Identify message data that helps identify the user on segment
+	err := c.SegmentClient.Enqueue(analytics.Identify{
+		UserId: strings.TrimSpace(string(userId)),
```
valaparthvi, line 125: We don't need `strings.TrimSpace(string(userId))`, it is already trimmed when returned by `getUserIdentity`.

valaparthvi: You can also remove `client *Client` arg from `getUserIdentity`.

valaparthvi (changes requested):

valaparthvi:
The functionality looks good, though.

/approve

pkg/segment/segment.go
kadel: thx, updated

valaparthvi (approved):

Issue #5051, opened 2021-09-08
Refactor test-cmd-devfile-url tests


prietyc123:
Scope of the issue:

- [ ] test-cmd-devfile-url should use new test approach and run successfully.

anandrkskd:
Adding to the scope of the issue:
- [ ] test-cmd-devfile-debug

PR #5057, opened 2021-09-09
Set odo to use staging registry in `CommonBeforeEach`

### LINKED ISSUE #5056: Some integration tests still using the production devfile registry
A while back https://github.com/openshift/odo/pull/4752 was merged to move the odo integration tests from using the production devfile registry (https://registry.devfile.io) to staging (https://registry.stage.devfile.io) as we didn't want the telemetry for the devfile registry polluted.

We recently set up telemetry on the devfile registry service and saw a **ton** of pollution coming back from what appears to be an automated process, for both listing devfiles, and download nodejs & python stacks. Since odo's integration tests primarily test with those stacks (mainly nodejs, but also python for link/unlink tests), we suspected it would be these tests.

I looked through odo's integration tests and saw that the following test files call out to the registry but don't set the registry to staging:
- cmd_link_unlink_test.go: `odo create` without `--s2i` flag
- cmd_pref_config_test.go: `odo create` without `--s2i` flag
- operatorhub/cmd_link_test.go `odo create` without `--s2i` flag
- operatorhub/cmd_service_test.go `odo create` without `--s2i` flag
- component.go: `odo create` without `--s2i` flag and `odo catalog list components`
- generic_test.go: `odo catalog list components`

I wonder if would be more straightforward to call `helper.SetDefaultDevfileRegistryAsStaging()` from within `helper.CommonBeforeEach()` so that helper.SetDefaultDevfileRegistryAsStaging() doesn't have to be called from each individual test file?

dharmit:
**What type of PR is this?**

/kind tests

Sets up odo to use staging registry in the `CommonBeforeEach` so that we don't need to set it again each time.

Fixes #5056 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

CI should pass.

### DIFF
--- a/tests/e2escenarios/e2e_devfile_test.go
+++ b/tests/e2escenarios/e2e_devfile_test.go
@@ -33,7 +33,6 @@ var _ = Describe("odo devfile supported tests", func() {
 		projectDirPath = commonVar.Context + projectDir
 		helper.MakeDir(projectDirPath)
 		helper.Chdir(projectDirPath)
-		helper.SetDefaultDevfileRegistryAsStaging()
 	})
 
 	// This is run after every Spec (It)
--- a/tests/helper/helper_generic.go
+++ b/tests/helper/helper_generic.go
@@ -302,6 +302,7 @@ func CommonBeforeEach() CommonVar {
 	cfg, _ := preference.New()
 	err := cfg.SetConfiguration(preference.ConsentTelemetrySetting, "false")
 	Expect(err).To(BeNil())
+	SetDefaultDevfileRegistryAsStaging()
 	return commonVar
 }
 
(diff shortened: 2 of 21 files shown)


feloy:
The symptom pointed by https://github.com/openshift/odo/issues/5056 also shows that devfiles are downloaded from tests outside of the `devfile` directory, which is not expected.

dharmit:
@feloy the tests in, for example, `tests/integration/operatorhub` create devfile components, but they are out of the `tests/integration/devfile` directory. So it's not entirely incorrect/problematic to be using devfile registry outside the `tests/integration/devfile` directory.

Something else that I was thinking yesterday when we were discussing removal of s2i related code and also #5051 was that we probably don't need to put anything under `tests/integration/devfile` any more.

feloy:
@prietyc123 could you validate?

prietyc123:
> Something else that I was thinking yesterday when we were discussing removal of s2i related code and also #5051 was that we probably don't need to put anything under tests/integration/devfile any more.

@dharmit I am confused, why don't we need to put anything under `tests/integration/devfile` ?

dharmit:
@prietyc123 that's unrelated to this PR. But the point is to move those tests to `tests/integration` once we get rid of everything related to s2i. In this case the tests related to s2i.

prietyc123:
make sense to me 👍 

/approve

PR #5065, opened 2021-09-13
Use interface for kclient

feloy:
/kind cleanup

- Uses an interface instead of concrete kclient.Client to make easier to test functions using this package
- Added mock generated with mockgen
- Updated mocks for existing mocsk (url, storage, locaConfigProvider)
- Added  a script to regenerate the mocks
- Added a unit test to demo the use of kclient mock

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


dharmit:
@feloy I haven't reviewed the PR. My comment is based only on `pkg/kclient/interface.go`.

Do you think having smaller interfaces would be better than having one large interface? I am suggesting along the lines of creating an interface for each of the `.go` files in `pkg/kclient`. So the methods under `// deploymnet.go` could live in its own interface, same for `// events.go`, `// ingress.go`, so on and so forth.

The reason I'm thinking this way is it makes testing easier because we can focus on implementing only a set of methods instead of a large number. Maybe using gomock we can automate the generation of mocks, but I remember reading on some official document that it's preferrable to have many smaller interfaces than one large interface. And in the end we can always make one big interface which takes all these interfaces if we really want a big interface, no?

What are your thoughts?

feloy:
@dharmit I agree that interfaces should be small.
That was my first intent to break into smaller interfaces, one for each file (deployment, pod, service, etc), but that will increase the number of fileds in the `genericclioptions.genericCxt`, and probably the number of parameters for some functions needing methods from more than one interface.

Even with a large interface, Gomock will make easy to mock some methods only.

dharmit:
@feloy thinking out loud here. If we're going to do this, would it provide benefit or be painful with those increased fields and parameters? 
* If it's painful, sure, we can avoid it. 
* If we don't know whether it will provide benefit, _but_ it won't be painful either, I'd prefer smaller interfaces.

I admit that my comment is based more on the articles/docs I have read and not based on actual work done along these lines. :)

feloy:
@dharmit 
If we split the interfaces, we will get something similar to the ClientSet of the client-go library, containing a set of interfaces, one for each file in the kclient directory.

We could build a clientset similar to the client-go one: https://pkg.go.dev/k8s.io/client-go/kubernetes#Clientset

(let me have a try...)

feloy:
/test v4.8-integration-e2e

```
[Fail] odo devfile push command tests when creating a nodejs component [BeforeEach] when setting git config and running odo push checks that odo push works with a devfile 
/go/src/github.com/openshift/odo/tests/helper/helper_cmd_wrapper.go:99
```

dharmit (approved):
/approve
/lgtm

dharmit:
@feloy thanks for getting this done! Just one question. We discussed on some call to name the interface as `Interface` so that it could be imported/used as `kclient.Interface`. It's `ClientInterface` at the moment. Just trying to understand if you forgot to change it or you would rather prefer it to be `ClientInterface`.

feloy:
> @feloy thanks for getting this done! Just one question. We discussed on some call to name the interface as `Interface` so that it could be imported/used as `kclient.Interface`. It's `ClientInterface` at the moment. Just trying to understand if you forgot to change it or you would rather prefer it to be `ClientInterface`.

I didn't change for the moment, I prefer to change it in a next iteration of refactoring

PR #5066, opened 2021-09-13
Fetch locale information for Telemetry

dharmit:
**What type of PR is this?**

/kind feature

$subject

Fixes part of #4994

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Look at the data being captured in Segment. You should see "locale" information in the "Raw" data.


kadel:
Not sure why, but I don't see any locale data in segment console

dharmit:
> Not sure why, but I don't see any locale data in segment console.

Did you check in the "Raw" tab? Even the time zone data that's added via #5067 shows in the "Raw" tab in `context`. I'm curious to understand why we're using `traits` when `context` struct already provides keys of our interest:
```go
// This type provides the representation of the `context` object as defined in
// https://segment.com/docs/spec/common/#context
type Context struct {
	App       AppInfo      `json:"app,omitempty"`
	Campaign  CampaignInfo `json:"campaign,omitempty"`
	Device    DeviceInfo   `json:"device,omitempty"`
	Library   LibraryInfo  `json:"library,omitempty"`
	Location  LocationInfo `json:"location,omitempty"`
	Network   NetworkInfo  `json:"network,omitempty"`
	OS        OSInfo       `json:"os,omitempty"`    <------------------------------
	Page      PageInfo     `json:"page,omitempty"`
	Referrer  ReferrerInfo `json:"referrer,omitempty"`
	Screen    ScreenInfo   `json:"screen,omitempty"`
	IP        net.IP       `json:"ip,omitempty"`
	Locale    string       `json:"locale,omitempty"`    <------------------------------
	Timezone  string       `json:"timezone,omitempty"`    <------------------------------
	UserAgent string       `json:"userAgent,omitempty"`
	Traits    Traits       `json:"traits,omitempty"`

	// This map is used to allow extensions to the context specifications that
	// may not be documented or could be introduced in the future.
	// The fields of this map are inlined in the serialized context object,
	// there is no actual "extra" field in the JSON representation.
	Extra map[string]interface{} `json:"-"`
}
```

kadel:
> I'm curious to understand why we're using `traits` when `context` struct already provides keys of our interest:

Have you been able to get to the `context` data from Woopra?

dharmit:
> Have you been able to get to the `context` data from Woopra? I was not able to see it there.

Woopra is not making much sense to me, tbh. In the sense, I'm not good at using it, yet.

> If you check how vscode telemetry does it, they send locale and timezone info in both context and traits. I assume that the reason is that you can't see the all the context data in Woopra.

Yeah, I assumed the same that adding it to `traits` could be for Woopra. I can add it to `traits` as well in the same PR if it helps.

kadel:
> Yeah, I assumed the same that adding it to `traits` could be for Woopra. I can add it to `traits` as well in the same PR if it helps.

yes please

pkg/segment/segment.go
```diff
@@ -67,12 +69,18 @@ func NewClient(preference *preference.PreferenceInfo) (*Client, error) {
 
 // newCustomClient returns a Client created with custom args
 func newCustomClient(preference *preference.PreferenceInfo, telemetryFilePath string, segmentEndpoint string) (*Client, error) {
+	// get the locale information
+	tag, err := locale.Detect()
+	if err != nil {
+		return nil, err
```
kadel, line 75: If locale detection fails, the whole telemetry will fail, and nothing will be send, do we want that?

dharmit: Snap. I did it correctly when I added locale info in `traits`. Fixed now.

feloy (approved):

pkg/segment/segment.go
```diff
@@ -68,6 +70,11 @@ func NewClient(preference *preference.PreferenceInfo) (*Client, error) {
 
 // newCustomClient returns a Client created with custom args
 func newCustomClient(preference *preference.PreferenceInfo, telemetryFilePath string, segmentEndpoint string) (*Client, error) {
+	// get the locale information
+	tag, err := locale.Detect()
+	if err != nil {
+		klog.V(4).Infof("couldn't fetch locale info: %w", err.Error())
```
feloy, line 76:
```suggestion
		klog.V(4).Infof("couldn't fetch locale info: %w", err)
```

pkg/segment/segment.go
```diff
@@ -157,6 +165,13 @@ func (c *Client) Upload(data TelemetryData) error {
 func addConfigTraits() analytics.Traits {
 	traits := analytics.NewTraits().Set("os", runtime.GOOS)
 	traits.Set("timezone", getTimeZoneRelativeToUTC())
+	// get the locale information
+	tag, err := locale.Detect()
+	if err != nil {
+		klog.V(4).Infof("couldn't fetch locale info: %w", err.Error())
```
feloy, line 171:
```suggestion
		klog.V(4).Infof("couldn't fetch locale info: %w", err)
```

feloy:
@dharmit It seems that `%w` is not supported by `Infof`

```
pkg/segment/segment.go:171:3: Infof call has error-wrapping directive %w
```

dharmit:
> @dharmit It seems that `%w` is not supported by `Infof`
> 
> ```
> pkg/segment/segment.go:171:3: Infof call has error-wrapping directive %w
> ```

I think I should revert it to the way it was, that is, use `err.Error()` in that call. WDYT?

feloy:
> > @dharmit It seems that `%w` is not supported by `Infof`
> > ```
> > pkg/segment/segment.go:171:3: Infof call has error-wrapping directive %w
> > ```
> 
> I think I should revert it to the way it was, that is, use `err.Error()` in that call. WDYT?

Yes, and also using %s instead of %w I think

PR #5076, opened 2021-09-16
cleanup website node packages

kadel:
**What type of PR is this?**

/kind documentation

- lockfiles for both yarn and npm were present. As Netlify is configured to use npm, I kept only npm lock file
- updated docusaurus to latest version


feloy (approved):

dharmit (commented):
/approve

Issue #4461, opened 2021-02-25
Remove "experimental" from odo preferences


dharmit:
## What versions of software are you using?

**Operating System:** all

**Output of `odo version`:** master (2.0.5)

## How did you run odo exactly?
`odo preference view`
`odo preference set -h`

## Actual behavior
odo preferences mention `Experimental` mode that was added during the run up to v2 release

## Expected behavior
`Experimental` mode is not being used any more. So odo should remove it.

/good-first-issue
/kind bug

kadel:
/status blocked

Docker support is currently experimetnal, before we  can remove it from preference we need to decide on its fate. 
https://github.com/openshift/odo/issues/4500

kadel:
/reopen
/remove-lifecycle rotten
still a prolem

kadel:
```
▶ odo preference view
PARAMETER             CURRENT_VALUE
UpdateNotification    
NamePrefix            
Timeout               
BuildTimeout          
PushTimeout           
Experimental          
Ephemeral             true
ConsentTelemetry      false
```

There is still `Experimental` config option in odo preference. This option should not be used for anything.

dimitraz:
I'm happy to take this on for Hacktoberfest, if it's still available

valaparthvi:
@dimitraz Thank you for your interest, let me confirm it and get back to you.

@kadel I think we can move this forward, I don't see any reference to Experimental except in its code implementation and some help docs.

dharmit:
@valaparthvi  Should we be adding some hacktoberfest related label to this issue and its PR?

Issue #5116, opened 2021-10-04
Protocol error during tests


feloy:
/kind bug

## How did you run odo exactly?

The error happens during integration tests, on different clusters (IBM Cloud, PSI)

## Actual behavior

```
protocol error: received *http.http2GoAwayFrame before a SETTINGS frame
Get "https://registry.stage.devfile.io/index": connection error: PROTOCOL_ERROR
```

```
protocol error: received *http.http2GoAwayFrame before a SETTINGS frame
[odo]  ✗  Failed to pull stack nodejs from registry.stage.devfile.io/devfile-catalog/nodejs:latest with allowed media types [application/vnd.devfileio.devfile.layer.v1 image/png image/svg+xml application/vnd.devfileio.vsx.layer.v1.tar application/x-tar]: failed to do request: Head "https://registry.stage.devfile.io/v2/devfile-catalog/nodejs/manifests/latest": connection error: PROTOCOL_ERROR
```

kadel:
This looks it might be something with devfile registry instance or registry library.
Maybe @GeekArthur or @johnmcollier could know something about this.

feloy:
@kadel In your opinion, what would be the effort (installation + maintenance) of having our own devfile registry on the test cluster?

kadel:
> @kadel In your opinion, what would be the effort (installation + maintenance) of having our own devfile registry on the test cluster?

I don't think it is worth it to host our own instance. But we could avoid using devfile registry in most of the tests.

We can modify our test to not use any registry and use devfile.yaml files that are part of our test fixtures instead.

The only tests that are meant to test registry integration would be using remote devfile registry. Theoretically it could be only one test that does that.

PR #5117, opened 2021-10-04
Remove experimental mode flag

dimitraz:
Signed-off-by: dimitraz <dimitrazuccarelli@gmail.com>

**What type of PR is this?**
/kind cleanup

Remove the experimental flag that was added during the run up to v2 release

https://github.com/openshift/odo/issues/4461


dimitraz:
Just btw, I haven't been able to run the integretation/e2e tests locally yet due to some issues with my environment, so I'm not 100% sure if they will pass when the PR tests run 😅

feloy:
Hi @dimitraz It seems there is unused code that makes `make validate` fail.

Don't worry about errors `protocol error: received *http.http2GoAwayFrame before a SETTINGS frame` in the integration tests, they are relative to the infrastucture, we are on it (https://github.com/openshift/odo/issues/5116). Apart from this error, all integration tests seem to pass.

Thanks for this work :+1:

feloy:
@dimitraz, I can see some other places where code related to experimental mode remains:

- `pkg/log/status.go`

The function `Experimental` is not used anymore and was introduced for the experimental mode (https://github.com/feloy/odo/commit/f47c9c9f4c0e1200f6272232cc8919bd3283d0ed)

- `pkg/odo/cli/component/push.go`

The Example string is split in `pushCmdExample` and `pushCmdExampleExperimentalOnly`. They could be merged in a single string.

Also, the `pushCmd.Annotations` value can be written in a single step, like in `pkg/odo/cli/component/list.go` for example.

This PR can explain why these example and annotations are split: https://github.com/feloy/odo/commit/3fcd3670e8320d73fdc14b67823b46ce14c8459b

- `pkg/odo/cli/url/create.go`

Same for `urlCreateExampleExperimental`.

PR #5138, opened 2021-10-12
allow `.devfile.yaml`  

### LINKED ISSUE #3126: allow `.devfile.yaml`
/kind feature

similar to discussed in https://github.com/eclipse/che/issues/14548 it is good when `devfile.yaml` does not need to be shown by default. 

## Why is this needed?

to make it easier to add support for odo and che without "polluting" users view.

anandrkskd:

**What type of PR is this?**
 /kind feature

This PR adds a feature with which we can use `.devfile.yaml` without adding `--devfile` flag.
With this PR in, odo will prefer `devfile.yaml` and if not present will use `.devfile.yaml` if   `.devfile.yaml` exists.

Fixes #3126

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

Simplest way to check if `.devfile.yaml` is used without using `--devfile` flag
```
git clone https://github.com/openshift/nodejs-ex && cd nodejs-ex

odo create nodejs 

mv devfile.yaml .devfile.yaml

odo push
```

### DIFF
--- a/pkg/odo/cli/service/list.go
+++ b/pkg/odo/cli/service/list.go
@@ -53,7 +53,7 @@ func (o *ServiceListOptions) Complete(name string, cmd *cobra.Command, args []st
 	}
 	o.Context, err = genericclioptions.New(genericclioptions.CreateParameters{
 		Cmd:              cmd,
-		DevfilePath:      devfile.DevfileFilenamesProvider(o.componentContext),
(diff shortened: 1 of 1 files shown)


feloy:
@anandrkskd It would be nice to have a more complete test with a `.devfile.yaml` file, covering the most important odo commands

PR #5192, opened 2021-10-28
odo build-images command

### LINKED ISSUE #5145: odo should be able to parse Devfile with `image` components and build+push the image
This is part of a larger user story #5109.
Because we don't want partial implementation of `odo deploy` command we will create a new command that will just build and push all image components in devfile without the deploy step. 

Acceptance criteria:
- [ ] odo should be able to parse Devfile with `image` components using devfile/library 1.2.0 and trigger image build and push using docker or podman
- [ ] `odo build-images` sub-command builds all image components defined in the devfile
- [ ] `odo build-images --push` builds and push all image components defined in the devfile
- [ ] odo can detect if podman or docker is installed on local machine (if both, it should use podman)
- [ ] the image build and push actions are done via podman/docker using their cli (odo will execute podman/docker commands)
- [ ] Docs for using this new command
- [ ] Tests

Implementation Notes:
- autodetect and use container engine that is available on the local machine (we will start with podman or docker)
- there is no odo or devfile specific way to handle image registry credentials, it will reduce whatever is configured for local container engine (set by docker login or podman login)

/priority high
/kind feature

feloy:
/kind feature

This PR implements the `odo build-images` command:

- [x] odo should be able to parse Devfile with image components using devfile/library 1.2.0 and trigger image build and push using docker or podman
- [x] odo build-images sub-command builds all image components defined in the devfile
- [x] odo build-images --push builds and push all image components defined in the devfile
- [x] odo can detect if podman or docker is installed on local machine (if both, it should use podman)
- [x] the image build and push actions are done via podman/docker using their cli (odo will execute podman/docker commands)
- [x] Docs for using this new command
- [x] Tests

Fixes #5145

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

Your devfile should respect:

```
schemaVersion: 2.2.0
```


kadel:
> But `odo create` always creates a devfile with schemaVersion 2.0.0. IMO, we should fix things in odo create. Maybe even before going ahead with this feature. Or at least before releasing this feature in upcoming odo release. WDYT?

odo doesn't create any devfiles. It just downloads it from devile registry. If there is a devfile 2.0.0 it can't have outerloop fields. Once there are devfiles with outerloop support they will have version 2.2.0.

kadel:
The main functionality works great. :+1:

But I had similar issues as Dharmit. The lack of the output is weird. For example, I had a typo in `buildContext` and only output I got was `error running podman command: exit status 125`

It would be great if `odo build-images` command could stream live output from podman/docker, so users know that something is happening, and when there was an error what it was.

feloy:
@kadel @dharmit Thanks for the feedback. I was not sure at first if the Docker/Podman output should be displayed, I first made it visible with the -v 4 option. 
I changed to make it streamed on stdout/stderr.

docs/website/docs/command-reference/build-images.md
```diff
@@ -0,0 +1,28 @@
...
+    imageName: quay.io/myusername/myimage
+    dockerfile:
+      uri: ./Dockerfile
+      buildContext: ${PROJECT_ROOT}
+  name: component-built-from-dockerfile
+```
+
+The `uri` field indicates the relative path of the Dockerfile to use, relative to the directory containing the `devfile.yaml`. The devfile specification indicates that `uri` could also be an HTTP URL, but this case is not supported by odo yet.
+
+The `buildContext` indicates the directory used as build context. The default value is `${PROJECT_ROOT}`.
```
kadel, line 24: I think that this will have to be changed. I opened it here https://github.com/devfile/api/issues/668

Right now, we don't need to do anything about this, this is how it is in spec, so it is technically correct.

pkg/image/image.go
```diff
@@ -0,0 +1,95 @@
+// image package provides functions to work with Components of type Image declared in the devfile
+package image
```
kadel, line 2: wouldn't `pkg/devfile/image/image.go` be a better place for this?

pkg/devfile/image/image.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return nil
+}
+
+// selectBackend selects the container backend to use for building and pushing images
+// It will detect podman and docker CLIs (in this order),
+// or return an error if none are present locally
+func selectBackend() (Backend, error) {
+
+	podmanCmd := os.Getenv("PODMAN_CMD")
```
kadel, line 79: 👍 this is good. This will could be useful for more than testing.

kadel:
/approve
I've also checked the code, and it looks good to me.

PR #5194, opened 2021-10-28
Bump to v2.4.2

dharmit:
Bumping the version rather quickly to make a release with #5187.


kadel:
/hold

I don't think that for this one bug, on a feature that according to telemetry is not that widely used needs an extra release.
The 2.4.2 would be normally released next week. I think that in this case it is ok to have this normally in the next release

rnapoles-rh (commented):
providing my review for when the hold gets removed
/lgtm

kadel:
/close
we will create a new PR for

PR #5196, opened 2021-10-29
Scripts for IBM Pipelines

feloy:
**What type of PR is this?**

/kind tests

This PR adds scripts to be executed from IBM Pipelines to run unit tests and integration tests on IBM Cloud

### DIFF
--- a/.ibm/pipelines/kubernetes-tests.sh
+++ b/.ibm/pipelines/kubernetes-tests.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+LOGFILE="pr-${GIT_PR_NUMBER}-kubernetes-tests-${BUILD_NUMBER}"
+
+source .ibm/pipelines/functions.sh
+
+install_ibmcloud cloud-object-storage kubernetes-service
+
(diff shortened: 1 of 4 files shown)


.ibm/pipelines/kubernetes-tests.sh
```diff
@@ -0,0 +1,25 @@
...
+    set -e
+    make install
+    make test-integration-devfile
+    make test-e2e-devfile
+    make test-cmd-project
+) |& tee "/tmp/${LOGFILE}"
+RESULT=${PIPESTATUS[0]}
+
+install_gh
+save_logs "${LOGFILE}" "Kubernetes Tests"
```
anandrkskd, line 23: @feloy this save log is added as a comment on gihub by itself or do we need to enable anything on pipleline side

anandrkskd (approved):
/lgtm

.ibm/pipelines/kubernetes-tests.sh
feloy: All is done here, there is nothing else to be done (except getting a private key to connect to GitHub)

rnapoles-rh (commented):
/lgtm

PR #5198, opened 2021-10-29
Refactor genericclioptions.Context

### LINKED ISSUE #5176: pkg/odo/genericclioptions usability
/kind bug

the package `pkg/odo/genericclioptions` is difficult to use, for the following reasons:

- [ ] there are different `New*` functions to initialize the context (`New`, `NewContext`, `NewContextCreatingAppIfNeeded`, `NewContextCompletion`, `NewOfflineContext`) and it is not clear which one to use on which situation
- [ ] The context contains a series of values (component context, project, app, kclient, occlient, ...), it is not clear where these information come from and what we can do with.

feloy:
**What type of PR is this?**

/kind cleanup

Fixes #5176 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	// DefaultAppName is the default name of the application when an application name is not provided
-	DefaultAppName = "app"
+	// defaultAppName is the default name of the application when an application name is not provided
+	defaultAppName = "app"
 
 	// gitDirName is the git dir name in a project
 	gitDirName = ".git"
--- a/pkg/odo/genericclioptions/context.go
+++ b/pkg/odo/genericclioptions/context.go
@@ -39,7 +39,7 @@ type internalCxt struct {
 	Client              *occlient.Client
 	command             *cobra.Command
 	project             string
-	Application         string
+	application         string
 	cmp                 string
 	OutputFlag          string
 	KClient             kclient.ClientInterface
@@ -276,4 +276,8 @@ func (o *Context) ComponentAllowingEmpty(allowEmpty bool, optionalComponent ...s
 
 func (o *Context) GetProject() string {
 	return o.project
+}
+
+func (o *Context) GetApplication() string {
+	return o.application
 }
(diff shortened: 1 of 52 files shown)


valaparthvi (changes requested):

valaparthvi:
It tends to show old result.

pkg/odo/genericclioptions/context.go
```diff
@@ -35,218 +34,140 @@ type Context struct {
...
 }
 
 // NewContextCompletion disables checking for a local configuration since when we use autocompletion on the command line, we
 // couldn't care less if there was a configuration. We only need to check the parameters.
 func NewContextCompletion(command *cobra.Command) *Context {
-	ctx, err := newContext(command, false)
+	ctx, err := New(CreateParameters{cmd: command})
 	if err != nil {
 		util.LogErrorAndExit(err, "")
 	}
```
valaparthvi, line 167: Can we fix this function to return the error as well so that we can avoid abrupt exit calls?

pkg/odo/genericclioptions/context.go
```diff
@@ -35,218 +34,140 @@ type Context struct {
...
+	devfile           bool
+	offline           bool
+	appIfNeeded       bool
+}
+
+func NewCreateParameters(cmd *cobra.Command) CreateParameters {
+	return CreateParameters{cmd: cmd}
+}
+
+func (o CreateParameters) CheckRouteAvailability() CreateParameters {
```
valaparthvi, line 71: **Optional:** Can we rename this method to something like `RequireRouteAvailability`? I understand your reasoning behind naming it `CheckRouteAvailability`, but I still find it a little non-intuitive. Intuitively I think it would be a function that checks if the route is available to use.

valaparthvi (changes requested):

pkg/odo/genericclioptions/context.go
feloy: This function is called during completion, and I'm not sure if we can modify its signature easily. I would prefer te change is part of another PR concerning completion.

feloy:
/test v4.9-integration-e2e

```
Error: Provider produced inconsistent result after apply
```

valaparthvi (approved):
Thank you for this PR :)

PR #5199, opened 2021-10-30
Fix odo overwrites devfile with flattened/resolved devfile and refactor component creation

### LINKED ISSUE #5112: odo overwrites devfile with parent reference with flattened/resolved devfile
/kind bug

**Operating System:** Win10

**Output of `odo version`:**   v2.3.1

1. Start with a `devfile.yaml` with a parent reference like:
```yaml
schemaVersion: 2.1.0
metadata:
  name: new-ol-version
  description: Child devfile to test OL image updating
parent:
  uri: "https://github.com/OpenLiberty/application-stack/releases/download/maven-0.7.0/devfile.yaml"
  components:
    - name: dev
      container:
        image: 'maven:3.6-adoptopenjdk-11-openj9'
```
2. `odo create mycomp`

## Actual behavior

My original devfile.yaml is completely replaced with a flattened devfile merging in the parent contents.  This is nice to see, but it might not be I want to store in Git / version control.

## Expected behavior

I'd expect to have some kind of option at least to flatten or not flatten.. or possibly something like the flattening would happen in a generated side file.

valaparthvi:
**What type of PR is this?**
/kind cleanup

Refactors the component/create.go 

Fixes #5112 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 


pkg/odo/cli/component/create.go
```diff
@@ -121,324 +113,118 @@ func NewCreateOptions() *CreateOptions {
...
 		log.Warningf("Found a dangling env file without a devfile, overwriting it")
-		if err = util.DeletePath(EnvFilePath); err != nil {
+		// Note: if the IF condition seems to have a side-effect, it is better to do the condition check separately, like below
+		err = util.DeletePath(envFilePath)
+		if err != nil {
 			return err
 		}
 	}
-
+	//Check if the directory already contains a devfile when --devfile flag is passed
```
feloy, line 154: could we move these checks to the implementation level? (in `FetchDevfileAndCreateComponent` or in some `Check` method?)

feloy (changes requested):

pkg/odo/cli/component/create_methods.go
```diff
@@ -0,0 +1,393 @@
...
+
+	co.devfileMetadata.devfileLink, co.devfileMetadata.devfileRegistry, err = findDevfileFromRegistry(catalogDevfileList, co.devfileMetadata.devfileRegistry.Name, co.devfileMetadata.componentType)
+	if err != nil {
+		return err
+	}
+	return fetchDevfileFromRegistry(co.devfileMetadata.devfileRegistry, co.devfileMetadata.devfileLink, co.DevfilePath, co.devfileMetadata.componentType, co.componentContext)
+}
+
+func (icm InteractiveCreateMethod) Rollback(componentContext string) {
+	os.RemoveAll(componentContext)
```
feloy, line 84: Are you sure we want to remove all the files of the devfile context? The user could have source files in this place

feloy (changes requested):

pkg/odo/cli/component/create_methods.go
```diff
@@ -0,0 +1,393 @@
...
+	// Check if both --devfile and --registry flag are used, in which case raise an error
+	if registryName != "" {
+		return &DevfileFlagWithRegistryFlagError{}
+	}
+	return nil
+}
+
+// validateAndFetchRegistry validates if the provided registryName exists and returns the devfile listed in the registy;
+// if the registryName is "", then it returns devfiles of all the available registries
+func validateAndFetchRegistry(registryName string) (catalog.DevfileComponentTypeList, error) {
```
valaparthvi, line 246: Should we allow creating a component if no devfile registry is present for direct(`odo create <comp_name> <comp_type>`) and interactive mode?

@feloy @dharmit

pkg/odo/cli/component/create_methods.go
valaparthvi: Philippe, what was the library you mentioned which could check diff in the directory. I am thinking to use it to decide what files should be deleted.

feloy: I talked about this lib recently, but I'm not sure it is relevant: https://github.com/monochromegane/go-gitignore

But don't you think we just need to delete the devfile.yaml file here? The interface is about fetching this devfile.yaml only, I think it should only revert the creation of this file.

pkg/odo/cli/component/create_methods.go
feloy: Yes, I think you could detect early that there is no registry configured and stop at this point

pkg/odo/cli/component/create_methods.go
valaparthvi: On a second thought, I am not sure if deleting the starter project files is a good idea, we don't do it now. We can just delete the devfile and env files for now.

valaparthvi:
@feloy Did you push https://github.com/openshift/odo/pull/5199/commits/afaf7a3a69dc0503c87605367eba682d9159ae44 by mistake on this branch instead of #5198 ?

pkg/odo/cli/component/create_methods.go
valaparthvi: Looks like this check has already been implemented.

feloy:
/lgtm
/approve

Thanks for this refactoring, this will help for the maintenance of the create command

PR #5206, opened 2021-11-05
Install podman on container for OpenShift tests

feloy:
**What type of PR is this?**

/kind tests

Install podman on container used for integration tests on OpenShift


kadel:
Just for the record. This is related to tests for https://github.com/openshift/odo/pull/5192

/approve
/lgtm

PR #5230, opened 2021-11-17
Do not run service related tests on Kubernetes

feloy:
**What type of PR is this?**

/kind tests

This PR skips tests on Kubernetes that are expecting operator backed services to be installed: on IBM environment, the Kubernetes cluster does not provide any operator backed service.

### DIFF
--- a/tests/integration/devfile/cmd_dot_devfile_test.go
+++ b/tests/integration/devfile/cmd_dot_devfile_test.go
@@ -2,6 +2,7 @@ package devfile
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


dharmit:
> on IBM environment, the Kubernetes cluster does not provide any operator backed service.

Likely a gap in my understanding, but why doesn't IBM Cloud provide any Operator backed service?

feloy:
> > on IBM environment, the Kubernetes cluster does not provide any operator backed service.
> 
> Likely a gap in my understanding, but why doesn't IBM Cloud provide any Operator backed service?

When I deployed the Kubernetes cluster, I installed the only necessary things on it, and the operators backed services were not necessary, based on the existing tests.

I think it is an interesting configuration to stick with, as we can test that odo works well on a vanilla cluster without OLM installed.

dharmit:
But there's nothing really for which we would like to run tests without the OLM. If you are concerned about testing linking using the library, it can still be done without installing the SBO on k8s on IBM Cloud, right?

Operators and `odo service` in general form a pretty important aspect of what odo provides vs. other tools, and if we are to confidently brag that odo is not OCP specific but also support vanilla k8s, we should run these tests on k8s. WDYT?

feloy:
> But there's nothing really for which we would like to run tests without the OLM. If you are concerned about testing linking using the library, it can still be done without installing the SBO on k8s on IBM Cloud, right?
> 
> Operators and `odo service` in general form a pretty important aspect of what odo provides vs. other tools, and if we are to confidently brag that odo is not OCP specific but also support vanilla k8s, we should run these tests on k8s. WDYT?

Yes, that's a good point. Let's close this PR, I will create an issue to deploy OLM and operators on IBM Kubernetes cluster

PR #5266, opened 2021-12-01
default to Y when asking to opt in for telemetry

kadel:
This is still aligned with our Data Collection Policy (https://developers.redhat.com/article/tool-data-collection) 
Telemetry is not enabled by default. Just the preselected answer is now Y instead of N.

**What type of PR is this?**

/kind feature


feloy:
@kadel it seems that refresh and retest are no effect on these tests that are finished but not taken into account. I think you will need to push your commit again

kadel:
opening new PR (https://github.com/redhat-developer/odo/pull/5270) 
There are some strange things happening with openshift-ci. The new PR will verify if it is just some that the new config not getting properly applied to existing PR or some other issue

Issue #5247, opened 2021-11-24
Having separate layers


feloy:
/kind bug

As described in the comment https://github.com/openshift/odo/issues/4057, the architecture of the code should use several layers (bottom to top):

- low-level layer that interacts with the cluster (using `k8s.io` packages)
- business layer that defines abstractions of the domain (app, component, project, storage, url, etc)
- cli layer that defines the UI

## Packages in layers

The low-level layer is composed of packages that access the underlying platforms:
- pkg/occlient
- pkg/kclient
- pkg/devfile (?)

The business layer is composed of packages that manage the `odo` objects:
- [x] pkg/application (#5293)
- [x] #5345
- [ ] pkg/catalog
- [ ] https://github.com/redhat-developer/odo/issues/5346  (#5329)
- [ ] pkg/config
- [ ] pkg/debug
- [x] pkg/preference (#5296)
- [x] pkg/project (#5293)
- [ ] pkg/service
- [ ] pkg/storage
- [ ] pkg/url
- [ ] pkg/version

The CLI layer is composed of packages in:
- [ ] cmd
- [ ] pkg/odo/cli

Other packages:
- [ ] pkg/segment

## Import of CLI layer

Packages of the CLI layer should not be imported from outside this CLI layer. These imports should be fixed:

```
$ grep -r github.com/openshift/odo/pkg/odo/cli pkg/ | grep -v ^pkg/odo/cli

pkg/odo/genericclioptions/runnable.go:			"github.com/openshift/odo/pkg/odo/cli/ui"
pkg/preference/preference.go:				"github.com/openshift/odo/pkg/odo/cli/ui"
pkg/catalog/catalog.go:			registryConsts	"github.com/openshift/odo/pkg/odo/cli/registry/consts"
pkg/catalog/catalog.go:			registryUtil	"github.com/openshift/odo/pkg/odo/cli/registry/util"
pkg/component/starter_project.go:	registryUtil	"github.com/openshift/odo/pkg/odo/cli/registry/util"
```

## Interactions with the console

Interactions with user (and especially console) should be done only from the CLI layer.

```
$ grep -r fmt.Print pkg/ | grep -v ^pkg/odo/cli | grep -v _test.go:

pkg/testingutil/httputils.go:	fmt.Println("running")
pkg/log/fidget/spinner.go:	fmt.Print("\r")
pkg/auth/login.go:		fmt.Println("Login failed (401 Unauthorized)")
[...]

PR #5296, opened 2021-12-16
Separate layers : Preference

feloy:
**What type of PR is this?**

/kind cleanup

Part of #5247 

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/redhat-developer/odo/blob/master/docs/dev/test-architecture.adoc)


feloy:
Unit Tests finished with errors.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-203.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-203.html)

feloy:
Kubernetes Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-190.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-190.html)

feloy:
OpenShift Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-220.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-220.html)

feloy:
Kubernetes Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-191.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-191.html)

feloy:
Unit Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-204.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-204.html)

feloy:
OpenShift Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-221.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-221.html)

feloy:
Kubernetes Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-193.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-193.html)

feloy:
Unit Tests finished with errors.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-206.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-206.html)

feloy:
OpenShift Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-223.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-223.html)

pkg/preference/implem.go
```diff
@@ -0,0 +1,564 @@
...
+	URL    string `yaml:"URL,omitempty"`
+	Secure bool
+}
+
+// Preference stores all the preferences related to odo
+type Preference struct {
+	metav1.TypeMeta `yaml:",inline"`
+
+	// Odo settings holds the odo specific global settings
+	OdoSettings odoSettings `yaml:"OdoSettings,omitempty"`
```
valaparthvi, line 172: Why do we need this struct? Can we not simply implement the `OdoSettings` properties into `Preference`?

pkg/odo/cli/preference/unset_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	err = opts.Validate()
+	if err != nil {
+		t.Errorf("Expected nil error, got %s", err)
+		return
+	}
+
+	prefClient.EXPECT().DeleteConfiguration("arg1")
+	err = opts.Run()
```
valaparthvi, line 38: This looks incomplete.
What are your thoughts on merging both the unit tests? Any reason for keeping them separate?

pkg/preference/implem.go
```diff
@@ -0,0 +1,564 @@
...
+}
+
+func (c *preferenceInfo) EphemeralSourceVolume() *bool {
+	return c.OdoSettings.Ephemeral
+}
+
+func (c *preferenceInfo) ConsentTelemetry() *bool {
+	return c.OdoSettings.ConsentTelemetry
+}
+
```
valaparthvi, line 542: Why have these functions when you already have the other Getters on these properties? Why can't you directly access them? These additional methods are a little confusing to me.

valaparthvi (changes requested):

pkg/preference/implem.go
feloy: I think is has been done this way to have this format for the preference file, similar to the Kubernetes resources format (with `OdoSettings` insted of `Spec`):
```
kind: Preference
apiversion: odo.dev/v1alpha1
OdoSettings:
  RegistryList:
    secure: false
  Ephemeral: false
  ConsentTelemetry: false
```

If we inline the fields of OdoSettings into Preferences, all the fields would be at the same level as `kind` and `apiVersion`.

pkg/odo/cli/preference/unset_test.go
feloy: Good catch, I will rework these tests, thanks

pkg/preference/implem.go
feloy: The other ones are returning the default value if the pointer value is nil or the dereferenced value otherwise, and these ones are returning the **raw** value (the pointer) - both are used in different places in the code.

We cannot access the values directly anymore because we want to access them through an interface (to be able to mock the full `Preference` object)

feloy:
Unit Tests finished with errors.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-218.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-unit-tests-218.html)

feloy:
OpenShift Tests finished with errors.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-235.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-openshift-tests-235.html)

feloy:
Kubernetes Tests finished successfully.
View logs: [TXT](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-205.txt) [HTML](https://s3.eu-de.cloud-object-storage.appdomain.cloud/odo-tests-openshift-logs/pr-5296-kubernetes-tests-205.html)

dharmit (changes requested):
Reviewed the `implem.go` file.

cmd/odo/odo.go
```diff
@@ -55,7 +55,7 @@ func main() {
 		return
 	}
 
-	cfg, err := preference.New()
+	cfg, err := preference.NewClient()
```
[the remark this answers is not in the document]
dharmit: How about replacing two occurrences of below in `odo.go`:
```go
util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.Background())), "")
```
with below:
```go
util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.WithValue(context.Background(), "preference", cfg))), "")
```
This way, every CLI's `Complete` function can access it with `cmdline.Context().Value("preference")`, and we can convert it to `preference.preferenceInfo` using something like below within the `preference` package:
```go
func GeneratePreferenceInfo(p interface{}) preferenceInfo {
	return p.(preferenceInfo)
}
```
WDYT? My motivation behind trying this is to reduce a bunch of calls to `preference.NewClient()` across the code to a single call. Sort of "don't repeat yourself" thing. It could also potentially help remove a bunch of `// TODO(feloy) Get from DI`, I guess.

dharmit (changes requested):
Hopefully not getting too carried away with the `context` thing here. 😉

pkg/odo/cli/component/create.go
```diff
@@ -345,7 +346,11 @@ func (co *CreateOptions) Run() (err error) {
 
 // NewCmdCreate implements the create odo command
 func NewCmdCreate(name, fullName string) *cobra.Command {
-	co := NewCreateOptions()
+	prefClient, err := preference.NewClient()
+	if err != nil {
+		panic("unable to set preference, something is wrong with odo, kindly raise an issue at https://github.com/redhat-developer/odo/issues/new?template=Bug.md")
```
valaparthvi, line 351: I think it is better to use LogErrorAndExit here, instead of panicking. Same comment for other panics.

dharmit:
@feloy I like the interfaces approach you have proposed here. However, I see some discrepancies in the structure of `pkg/preference` and `pkg/application` & `pkg/project` in #5293.
* Why are we using `implem.go` file? It's not present in #5293.
* How about using `interface.go` or `types.go` to define an interface and have its implementation in `preference.go` (for this PR) and `application.go` & `project.go` for #5293?
Is there some structuring paradigm that you are following here which I should rather be aware of?

My intention is to setup a proper code contributing guideline for new contributors to get started with odo code.

cmd/odo/odo.go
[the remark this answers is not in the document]
feloy: I would prefer not to use Context.Value, as it obscures the program's flow (see https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39 - "Context.Value obscures your program’s flow" for detailed arguments)

feloy:
> @feloy I like the interfaces approach you have proposed here. However, I see some discrepancies in the structure of `pkg/preference` and `pkg/application` & `pkg/project` in #5293.
> 
> * Why are we using `implem.go` file? It's not present in [Separate layers - Application + Project #5293](https://github.com/redhat-developer/odo/pull/5293).

In application and project, I used kubernetes.go because the implementation is Kubernetes-related, and to make it similar to the `url` and `storage` packages.

The `preference` package is not related to Kubernetes, but to filesystem. I'm not sure which file name to use here.



> * How about using `interface.go` or `types.go` to define an interface and have its implementation in `preference.go` (for this PR) and `application.go` & `project.go` for [Separate layers - Application + Project #5293](https://github.com/redhat-developer/odo/pull/5293)?

It seems the standard is to declare the interface in a file named with the package name, for example in the go stdlib, the `FS` interface is declared in a `fs.go` file: https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/io/fs/fs.go


> * What's the purpose of `kubernetes.go` file in [Separate layers - Application + Project #5293](https://github.com/redhat-developer/odo/pull/5293)? odo has very limited functionality that's specific to OCP. And even those are just another kubernetes resources ([Having separate layers #5247 (comment)](https://github.com/redhat-developer/odo/issues/5247#issuecomment-978933036)).
> * Can we define below type at one place:
>   ```go
>   type kubernetesClient struct {
>   	client kclient.ClientInterface
>   }
>   ```

To implement an interface, we need a type. Incidentally, each implementation (for project, application) has the same underlying structure, but it is not sure it will always be the same. We will probably add more clients to some of these structures (for ex a project.ClientInterface, or a service.ClientInterface, etc)

>   
>       
>     
>   I genuinely don't know at the time of writing this as to how it can be done. But I have a feeling that defining `kubernetesClient` under various `pkg/*` directories will bite us at some point.
> 
> Is there some structuring paradigm that you are following here which I should rather be aware of?
> 
> My intention is to setup a proper code contributing guideline for new contributors to get started with odo code.

cmd/odo/odo.go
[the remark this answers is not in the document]
dharmit: Thanks for sharing that article. It was a great read. Sorry about long reply. Correct me wherever you find the need to.

That article was heavily based on REST APIs, which I personally have very less experience with. I was trying to draw parallels with odo which is a CLI without a daemon process (loosely comparable to DB from that article).

I really loved the "inform vs. control" argument in that article.

In case of odo, if we do use `Context.Value`, I think, it should stay limited to the CLI layer. Layers below it - business logic & clients - should not use it directly in any way. Like the blog you shared suggests, we should create struct that is then passed to the below layers. Currently, below three calls to `preference.NewClient()` are making me cringe a bit since they are in layers below the CLI layer:
* `pkg/devfile/adapters/kubernetes/storage/utils.go`
* `pkg/kclient/oc_server.go`
* `pkg/kclient/pods.go`

I'm probably overthinking a condition which is making me cringe. I'll try to explain.

In case of `pkg/devfile/adapters/kubernetes/storage/utils.go`, the call is happening as a part of https://github.com/feloy/odo/blob/277a8c3f2b36b5a808bed9a0edbb88a60dd749bc/pkg/devfile/adapters/kubernetes/component/adapter.go#L199 which is after the call to https://github.com/feloy/odo/blob/277a8c3f2b36b5a808bed9a0edbb88a60dd749bc/pkg/devfile/adapters/kubernetes/component/adapter.go#L192.

Now, a hypothetical/crazy scenario I have thought of is what if `service.PushKubernetesResources` takes quite long (which is quite possible when creating a service) and user changes something in the preference before `a.createOrUpdateComponent` is called.

Irrespective of us using `Context.Value`, I feel that we should not be reading the preferences from the disk so deep into the code. It should be just passed along from CLI layer, and your approach of using Dependency Injection is absolutely cool.

> I would prefer not to use Context.Value

TBH, I have no preference. I just learned about `Context.Value` while doing that review 10 days back because I remembered reading about context but not being able to understand how it could be used. While doing the review, I found about `Context.WithValue` and tried it out. I'm just trying to have an open conversation and decide what's best.

pkg/preference/implem.go
```diff
@@ -0,0 +1,561 @@
...
+		EphemeralSetting:          EphemeralDescription,
+		ConsentTelemetrySetting:   ConsentTelemetryDescription,
+	}
+
+	// set-like map to quickly check if a parameter is supported
+	lowerCaseParameters = util.GetLowerCaseParameters(GetSupportedParameters())
+)
+
+// odoSettings holds all odo specific configurations
+type odoSettings struct {
```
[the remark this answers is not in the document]
feloy: The content of the `$HOME/.odo/preference.yaml` file is:

```
kind: Preference
apiversion: odo.dev/v1alpha1
OdoSettings:
  RegistryList:
    secure: false
  Ephemeral: false
  ConsentTelemetry: false
```

Inside, we have an `OdoSettings` part, matching the `OdoSettings` structure. I would prefer to not break the format of the `preference.yaml` file, and have the same name in the file and in the code.

pkg/preference/implem.go
```diff
@@ -0,0 +1,561 @@
+package preference
```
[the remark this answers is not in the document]
feloy: I moved the constants and variables to a `variables.go` file.

The structs are parts of the implementation, I would prefer to keep them in the `implem.go` file

pkg/preference/implem.go
```diff
@@ -0,0 +1,561 @@
...
+	// Odo settings holds the odo specific global settings
+	OdoSettings odoSettings `yaml:"OdoSettings,omitempty"`
+}
+
+// preferenceInfo wraps the preference and provides helpers to
+// serialize it.
+type preferenceInfo struct {
+	Filename   string `yaml:"FileName,omitempty"`
+	Preference `yaml:",omitempty"`
+}
```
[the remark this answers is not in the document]
feloy: I find also this quite intricate, but I would prefer to change this in another refactoring

pkg/preference/implem.go
```diff
@@ -0,0 +1,561 @@
...
+				break
+			}
+		}
+	}
+
+	return &c, nil
+}
+
+// RegistryHandler handles registry add, update and delete operations
+func (c *preferenceInfo) RegistryHandler(operation string, registryName string, registryURL string, forceFlag bool, isSecure bool) error {
```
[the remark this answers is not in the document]
feloy: As this is part of the implementation of the interface, I don't think so.

pkg/preference/implem.go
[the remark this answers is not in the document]
feloy: See https://github.com/redhat-developer/odo/pull/5296#issuecomment-1004109663

pkg/odo/cli/component/create.go
feloy: You're right, thanks

valaparthvi (changes requested):

feloy:
/test v4.9-integration-e2e

```
  * could not run steps: step [release:latest] failed: release "release-latest" failed: the pod ci-op-1wjwi82z/release-latest was deleted without completing after 29s (failed containers: ) 
```

feloy:
/test unit

```
  * could not run steps: step unit failed: "unit" test steps failed: "unit" pod "unit-unit-steps" failed: the pod ci-op-7nl9cs8b/unit-unit-steps was deleted without completing after 7m20s (failed containers: )
```

pkg/devfile/adapters/kubernetes/storage/utils.go
```diff
@@ -188,7 +188,7 @@ func generateVolumeNameFromPVC(pvc string) (volumeName string, err error) {
 
 // HandleEphemeralStorage creates or deletes the ephemeral volume based on the preference setting
 func HandleEphemeralStorage(client kclient.ClientInterface, storageClient storage.Client, componentName string) error {
-	pref, err := preference.New()
+	pref, err := preference.NewClient()
```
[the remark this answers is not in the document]
dharmit: I suggested the same in https://github.com/redhat-developer/odo/pull/5296#discussion_r777858651. But it seems like it was a bit too long to read. 😃 

> Like the blog you shared suggests, we should create struct that is then passed to the below layers. Currently, below three calls to `preference.NewClient()` are making me cringe a bit since they are in layers below the CLI layer:
> 
> * `pkg/devfile/adapters/kubernetes/storage/utils.go`
> 
> * `pkg/kclient/oc_server.go`
> 
> * `pkg/kclient/pods.go`

pkg/segment/integrations.go
```diff
@@ -17,7 +17,8 @@ func getTelemetryForDevfileRegistry() (registryLibrary.TelemetryData, error) {
 		Client: registryConsts.TelemetryClient,
 	}
 
-	cfg, err := preference.New()
+	// TODO(feloy) Get from DI
+	cfg, err := preference.NewClient()
```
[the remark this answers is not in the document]
feloy: I would prefer to try to refactor the `segment` package the same way as the preference, application and project ones, to be able to inject the preference Client into the Segment module at creation

pkg/odo/cli/component/create.go
```diff
@@ -115,9 +119,10 @@ odo catalog list components
 %[1]s nodejs --app myapp --project myproject`)
 
 // NewCreateOptions returns new instance of CreateOptions
-func NewCreateOptions(prjClient project.Client) *CreateOptions {
+func NewCreateOptions(prjClient project.Client, prefClient preference.Client) *CreateOptions {
 	return &CreateOptions{
-		PushOptions: NewPushOptions(prjClient),
+		PushOptions: NewPushOptions(prjClient, prefClient),
+		prefClient:  prefClient,
```
valaparthvi, line 125: I don't think you will need a separate prefClient here. CreateOptions 'inherits' PushOptions, thereby getting access to prefClient.

pkg/odo/cli/preference/view_test.go
```diff
@@ -0,0 +1,43 @@
...
+		return
+	}
+
+	prefClient.EXPECT().UpdateNotification()
+	prefClient.EXPECT().NamePrefix()
+	prefClient.EXPECT().Timeout()
+	prefClient.EXPECT().BuildTimeout()
+	prefClient.EXPECT().PushTimeout()
+	prefClient.EXPECT().EphemeralSourceVolume()
+	prefClient.EXPECT().ConsentTelemetry()
```
valaparthvi, line 37: Don't you need to return something here?

valaparthvi (changes requested):

pkg/odo/cli/preference/unset_test.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+
+			err = opts.Validate()
+			if err != nil {
+				t.Errorf("Expected nil error, got %s", err)
+				return
+			}
+
+			if tt.exists || tt.forceFlag {
+				prefClient.EXPECT().DeleteConfiguration("arg1")
```
valaparthvi, line 58: Should this return any error, or if nothing is mentioned, it returns a default value?

valaparthvi (approved):
I have one more suggestion, but seems like things are working without it as well.

pkg/odo/cli/preference/unset_test.go
feloy: I cannot find any documentation on this, but I assume it will return the zero values for all the returned values if we don't specify any other value.

PR #5310, opened 2021-12-27
Start using SBO again

### LINKED ISSUE #5221: start using SBO again
new SBO  version hard codes binding definitions for some of the operators. This might make using SBO with odo much easier

- [ ] investigate if we can leverage this to fix https://github.com/openshift/odo/issues/5084 
- [ ] Investigate which Operators' annotations have been hard-coded by SBO and add tests for these in odo
  - [ ] Doesn't have to be all operators in a single sprint.
  - [ ] MVP would be to manually test it, but automate at least one in the first sprint

/triage needs-information

dharmit:
**What type of PR is this?**
/kind feature

(I really don't know what kind this could be. Since SBO version has changed between the time we stopped using it and this PR, I have put it as feature.)

$subject. We need it since we don't support SBO at all at the moment and use the SB library by default. However, SBO 1.0 is now GA, and we can expect some stability on that part.

Fixes #5221

**PR acceptance criteria**:

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- [x] I have read the [test guidelines](https://github.com/redhat-developer/odo/blob/master/docs/dev/test-architecture.adoc)

CI should pass.

### DIFF
--- a/pkg/kclient/operators.go
+++ b/pkg/kclient/operators.go
@@ -30,8 +30,8 @@ const (
 // IsServiceBindingSupported checks if resource of type service binding request present on the cluster
 func (c *Client) IsServiceBindingSupported() (bool, error) {
 	// Detection of SBO has been removed from issue https://github.com/redhat-developer/odo/issues/5084
-	return false, nil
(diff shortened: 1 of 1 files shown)


dharmit:
@feloy is looking into this.

/close

PR #5315, opened 2022-01-05
odo push recreates the pod if scaled to 0

### LINKED ISSUE #5041: odo push fails with idled components
/kind bug

**Operating System:**
MacOS 

**Output of `odo version`:**
odo v2.3.0 (089b8ea26)

odo push

## Actual behavior
After creating a component and doing `odo push` scale the pod count to 0 on the deployment. 
do a second `odo push` it fails  with 

```
Failed to start component with name "java-quarkus-test-wrhd". Error: Failed to create the component: unable to get pod for component java-quarkus-test-wrhd: error while waiting for pod component=java-quarkus-test-wrhd: waited 4m0s but couldn't find running pod matching selector: 'component=java-quarkus-test-wrhd'
```

`odo delete` in the same situation fails with
```
 pod not found for the selector: app.kubernetes.io/part-of=app,app.kubernetes.io/instance=java-quarkus-test-wrhd
```

## Expected behavior

Either `odo push` succeeds and scales the deployment 
or `odo delete` cleans the deployment so that `odo push` can follow

dharmit:
**What type of PR is this?**
/kind bug

Refer issue being fixed

Fixes #5041 

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [ ] I have read the [test guidelines](https://github.com/redhat-developer/odo/blob/master/docs/dev/test-architecture.adoc)

Refer issue description of original issue for steps to test changes.

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -263,8 +263,9 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	var podName string
 
 	// If the component already exists, retrieve the pod's name before it's potentially updated
+	var podErr error
 	if componentExists {
-		pod, podErr := a.getPod(true)
+		pod, podErr = a.getPod(true)
 		if podErr != nil {
 			return errors.Wrapf(podErr, "unable to get pod for component %s", a.ComponentName)
 		}
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -131,7 +131,7 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	// If the component already exists, retrieve the pod's name before it's potentially updated
 	if componentExists {
 		// First see if the component does have a pod. it could have been scaled down to zero
-		_, err := a.Client.GetOnePodFromSelector(fmt.Sprintf("component=%s", a.ComponentName))
+		_, err = a.Client.GetOnePodFromSelector(fmt.Sprintf("component=%s", a.ComponentName))
 		// If an error occurs, we don't call a.getPod (a blocking function that waits till it finds a pod in "Running" state.)
 		// We would rely on a call to a.createOrUpdateComponent to reset the pod count for the component to one.
 		if err == nil {
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/utils/pointer"
+
 	"github.com/devfile/library/pkg/devfile/generator"
 	componentlabels "github.com/redhat-developer/odo/pkg/component/labels"
 	"github.com/redhat-developer/odo/pkg/devfile"
@@ -533,7 +535,7 @@ func (a *Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSp
 	if err != nil {
 		return err
 	}
-
+	deployment.Spec.Replicas = pointer.Int32Ptr(1)
 	if deployment.Annotations == nil {
 		deployment.Annotations = make(map[string]string)
 	}
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -118,27 +118,12 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 		}
 	}
 
-	componentExists := false
-	if a.deployment != nil {
-		componentExists = true
-	}
+	componentExists := a.deployment != nil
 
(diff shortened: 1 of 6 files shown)


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -529,7 +529,7 @@ func (a *Adapter) createOrUpdateComponent(componentExists bool, ei envinfo.EnvSp
 	if err != nil {
 		return err
 	}
-
+	deployment.Spec.Replicas = pointer.Int32Ptr(1)
```
feloy, line 532: It would be nice to have this field set when the Deployment is first created, by the devfile library. Could we propose a path to the library to add the Replicas to the `generator.DeploymentParams` structure?

tests/integration/devfile/cmd_devfile_push_test.go
```diff
@@ -188,14 +188,23 @@ var _ = Describe("odo devfile push command tests", func() {
...
 
 				It("should execute run command on odo push", func() {
 					output = helper.Cmd("odo", "push", "--project", commonVar.Project).ShouldPass().Out()
 					Expect(output).To(ContainSubstring("Executing devrun command"))
 				})
 			})
+
+			When("the Deployment's Replica count (pods) is scaled to 0", func() {
+				BeforeEach(func() {
+					commonVar.CliRunner.ScalePodToZero(cmpName, "app")
```
feloy, line 201: I think you will need to wait for the pod to be terminated before to continue, or the scale command won't have the time to have any effect

pkg/devfile/adapters/kubernetes/component/adapter.go
dharmit: By setting it at library level, it becomes applicable to Che folks as well. I'm not sure if that's what they would want. Otherwise, I think it could be patched at library level.

IMO, we could track this separately (issue on devfile repo or a discussion in the Devfile cabal). WDYT?

feloy: Through the `generator.DeploymentParams` structure, it could be an optional parameter, to not impact other library's users.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -127,7 +126,7 @@ func (a Adapter) Push(parameters common.PushParameters) (err error) {
 	var podName string
 
 	// If the component already exists, retrieve the pod's name before it's potentially updated
-	if componentExists {
+	if componentExists && *a.deployment.Spec.Replicas == 1 {
```
feloy, line 129: This is very specific to this issue, and is not generic enough to cover all the cases when a pod is not present.

I made some more reading of this Push() function, and understand now that this call to getPod() is necessary to get the name of the previous pod (the one we synced files to previously), and to compare it later (https://github.com/redhat-developer/odo/pull/5315/files#diff-674dc676ed2f8aa2588ede2db4112fe412bfa0365f7b0f7303d5ae407582d2aaL271) with the pod we will sync the files to.

If the pod is restarted, all the files need to be synced; in the other case, only changed files need to be synced.

Instead of making the specific test on Deployment Replicas, you could react differently if `getPod` in line 131 fails. Instead of returning an error, you could continue and make sure that `podChanged`will be set to true (as it is sure we will need to sync all the files again), for example by keeping `podName` to the empty string.

Also, could you please change the variable name `componentExists` to `deploymentExists`, to make the flow easier to understand?

feloy (changes requested):

pkg/devfile/adapters/kubernetes/component/adapter.go
dharmit: > I made some more reading of this Push() function, and understand now that this call to getPod() is necessary to get the name of the previous pod (the one we synced files to previously), and to compare it later (https://github.com/redhat-developer/odo/pull/5315/files#diff-674dc676ed2f8aa2588ede2db4112fe412bfa0365f7b0f7303d5ae407582d2aaL271) with the pod we will sync the files to.

Yes, that's what I was saying while we were discussing this yesterday. But I was struggling to explain. :) 



> Instead of making the specific test on Deployment Replicas, you could react differently if `getPod` in line 131 fails. Instead of returning an error, you could continue and make sure that `podChanged`will be set to true (as it is sure we will need to sync all the files again), for example by keeping `podName` to the empty string.

But for that to happen, the user will have to wait for whatever the value of `PushTimeout` is in odo preferences. By default, it is 4 minutes. Also, `getPod()` function makes a call to `WaitAndGetPodWithEvents` in `pkg/kclient/pods.go` which mandates the desired phase of the pod to be `Running`: https://github.com/redhat-developer/odo/blob/724f16e689545dd4a81671da3e116a33df4832d3/pkg/kclient/pods.go#L25-L27

Now, I know that we can make change to the code to not mandate it. But I'm honestly having a hard time and would appreciate some help. Using the debugger, I can't seem to be able to get inside below go routine: https://github.com/redhat-developer/odo/blob/724f16e689545dd4a81671da3e116a33df4832d3/pkg/kclient/pods.go#L46-L95

PR #5341, opened 2022-01-12
[WIP] odo v3 cli proposal

kadel:


docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,305 @@
...
+#### `odo list binding`
+
+#### `odo list service`
+
+#### `odo list endpoint`
+
+#### `odo list storage`
+
+### `odo app`
+TODO
```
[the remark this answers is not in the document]
kadel: Yes, we might end up using flag instead of command

docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,305 @@
...
+
+
+
+#### Flags
+- `--name` - name of the component (required)
+- `--devfile` - name of the devfile in devfile registry (required if `--devfile-path` is not defined)
+- `--devfile-registry` - name of the devfile registry (as configured in `odo registry`). It can be used in combination with `--devfile`, but not with `--devfile-path` (optional)
+- `--starter` - name of the  starter project (optional)
+- `--devfile-path` - path to a devfile. This is alternative to using devfile from Devfile registry. It can be local filesystem path or http(s) URL (required if `--devfile` is not defined)
+
```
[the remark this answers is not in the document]
kadel: > What about --project and --app flag? I do not see how we'll be using them with odo dev. 

There won't be there in the initial implementation. I'm not sure yet, but I was also playing with the idea of completely removing "app"  from odo, in favor of generic label support. 

> Are we dropping the support for --context flag?

yes, I was thinking that we could remove it for sake of simplicity. 

> I also don't see --token flag that should be used while downloading devfile from ---devfile-path. 

`--token` was a mechanism for authenticating aginst GitHub-based Devfile registries. We are not going to support them anymore.

docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,305 @@
...
+### `odo exec`
+mostly as it is
+
+### `odo status`
+mostly as it is
+
+## Commands that will be added in v3.0.0-alpha2
+
+### `odo create`
```
feloy, line 254: What will happen to the v2 `odo create` command? Will it remain as is, or removed in favour of `odo init` / `odo dev`?

kadel: `odo create` on itself won't be a valid command. But it will have subcommand to create other "things" like `odo create service`

docs/proposals/odo-v3-cli/odo-init.md
```diff
@@ -0,0 +1,94 @@
...
+Command should use registries as configured in `odo registry` command. If there is multiple registries configured it should use all of them.
+
+## Flags
+
+- `--name` (string) - name of the component (required)
+- `--devfile` (string) - name of the devfile in devfile registry (required if `--devfile-path` is not defined)
+- `--registry` - (string) name of the devfile registry (as configured in `odo registry`). It can be used in combination with `--devfile`, but not with `--devfile-path` (optional)
+- `--starter`  (string) - name of the  starter project (optional)
+- `--devfile-path` (string) - path to a devfile. This is alternative to using devfile from Devfile registry. It can be local filesystem path or http(s) URL (required if `--devfile` is not defined)
+- `-o` (string) output information in a specified format (json).
```
feloy, line 24: What is the expected information in output? The devfile.yaml file in json format?

docs/proposals/odo-v3-cli/odo-login-logout.md
```diff
@@ -0,0 +1,9 @@
+# `odo login` and `odo logout`
+
+## `odo login`
+
+`odo login` should work exactly the same way as `oc login` command. It should have the same arguments and flags.
+
+## `odo logout`
+
+`odo login` should work exactly the same way as `oc login` command. It should have the same arguments and flags.
```
feloy, line 9:
```suggestion
`odo logout` should work exactly the same way as `oc logout` command. It should have the same arguments and flags.
```

docs/proposals/odo-v3-cli/odo-create-component.md
```diff
@@ -0,0 +1,66 @@
...
+  Delete environment variable "FOO1"
+  Add new environment variable
+
+⠏ Downloading "java-quarkus". DONE
+
+Your component is ready in the current directory.
+To deploy it to the cluster you can run `odo deploy`.
+```
+
+The questions for interactive mode are identical to questions in `odo dev` and `odo deploy` command.
```
dharmit, line 65: Just scrap the `odo create` command for the sake of making odo maintenance less of a nightmare than it already is.

docs/proposals/odo-v3-cli/odo-delete-component.md
```diff
@@ -0,0 +1,17 @@
+# `odo delete component`
+Delete component.
+By default it deletes component from cluster (deletes all resources from both `odo dev` and `odo deploy` that were crated on the cluster that belong to the component) that in the current directory.
+
+Similarly to `odo describe component` this commands works in two modes.
```
dharmit, line 5: Why do we want multiple modes with everything? What's the obsession here to provide more ways/modes with every command? Why can't we figure just ONE way of doing things, doing it well without breaking, and then if user feedback is to have more ways, we can have a discussion to add more!

docs/proposals/odo-v3-cli/odo-delete-component.md
```diff
@@ -0,0 +1,17 @@
...
+Without `--component` `--application` or `--namespace` it works with the component in the current directory.
+If user defines `--component` and `--application` than it works with the component that should exist on the cluster, and it should not touch local component even if the current directory is component directory.
```
dharmit, line 7: > it should not touch local component even if the current directory is component directory.

Again, why not make odo mimic the behaviour of git here?

Also, what kind of commands do we expect from users here?
```sh
$ odo delete component --component someTerribleComponent --application blah
$ odo delete component --component blah --application moreBlah --namespace yikes
$ odo delete component
```

Since when did odo start using the term namespace instead of project? Can we do more research/thinking before coming up with what we want the CLI UX to look like?

docs/proposals/odo-v3-cli/odo-delete-component.md
```diff
@@ -0,0 +1,17 @@
...
+`--namespace` flag is optional, but it makes sense only together with `--component` and `--application`, if not used than it uses the current namespace as defined in `KUBECONFIG`.
+
+
+## flags
+- `-o json` - show output in json format
+- `-f` `--force` - force deletion, don't ask for confirmation
+- `--component` - name of the existing component on the cluster
+- `--application` - name of the application that the component belongs to
+- `--namespace` - namespace
```
dharmit, line 16: I request removing all these flags and make a user `cd` into the directory of the component they want to delete. As for IDE plugins, they can use `odo list` to get the path of the component and `cd` into it before doing `odo delete`. So, for IDE plugins, deletion would be a two-step thing - first `cd` and then `odo delete`. However, it could be invisible to the user.

docs/proposals/odo-v3-cli/odo-describe-component.md
```diff
@@ -0,0 +1,35 @@
+
+# `odo describe component`
+
+Show detailed information about component.
+
+By default, it shows detail about component in the current directory.
+
+Alternatively users can use flags (`--component`, `--application`, `--namespace`) to specify existing component on the cluster and show information about it.
```
dharmit, line 8: Same request - please remove these flags. Please allow us to have a simpler odo v3 than its past iterations.

docs/proposals/odo-v3-cli/odo-describe-component.md
```diff
@@ -0,0 +1,35 @@
+
+# `odo describe component`
+
+Show detailed information about component.
+
+By default, it shows detail about component in the current directory.
```
dharmit, line 6: Define "detail". Define it clearly. Please.

dharmit (changes requested):
Few problems with the proposal from a developer and maintainer point of view:
1. We have started working on v3 (`odo init`) already before the proposal was approved.
2. We are continuing with providing more than one way of doing things at a bunch of places. Clearly, we haven't learned our lessons yet.
3. Unclear usage of flags like `--component` and `--application`.
4. No mention of `odo app`. We have not seen any benefit of `app` in the past. If it's needed for other tools, we could hard code app to be same as the value of project. But it makes no sense to continue with app in odo. If it does, please care to explain.

docs/proposals/odo-v3-cli/odo-describe-component.md
kadel: work in progress. Will do it later

docs/proposals/odo-v3-cli/odo-create-component.md
```diff
@@ -0,0 +1,66 @@
+
+# `odo create component`
+
+
+
+Create new Devfile component in the current directory.
+Some users might prefer creating components separately before running `odo dev` or `odo deploy`.
+But this command is mainly intended to be used by IDE plugins or scripts.
```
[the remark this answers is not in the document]
dharmit: By "they", I mean the IDE plugins.

docs/proposals/odo-v3-cli/odo-describe-component.md
kadel: can't how would user describe the component that exists on the cluster but for which there is no local devfile? This is critical functionality for IDE plugins.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: > I request removing all these flags and make a user cd into the directory of the component they want to delete. As for IDE plugins, they can use odo list to get the path of the component and cd into it before doing odo delete

That wouldn't work. The user needs to be able to delete components for which there is no local devfile.
What if I accidentally delete a component locally before deleting it from the cluster?

docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,89 @@
...
+## Changes from odo v2 command perspective
+
+
+- `odo component` **removed** every command from “odo component *” already exists as a root command.
+- `odo watch` replaced with [`odo dev`](odo-v3-cli/odo-dev.md)
+- `odo push`  replaced with [`odo dev`](odo-v3-cli/odo-dev.md)
+- `odo unlink`  replaced with [`odo delete binding`](odo-v3-cli/odo-delete-binding.md)
+- `odo link`  replaced with [`odo create binding`](odo-v3-cli/odo-create-binding.md)
+- `odo url`  replaced with [`odo create endpoint`](odo-v3-cli/odo-create-endpoint.md)
+- `odo test`  replaced `odo run test`
```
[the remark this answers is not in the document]
kadel: People are using it but not much. The problem is also that not every devfile has this command defined.

docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,89 @@
...
+- `odo env` integrated with [`odo config`](odo-v3-cli/odo-config.md)
+- `odo debug` replaced with [`odo run debug`](odo-v3-cli/odo-run-debug.md)
+- `odo registry` removed, but the functionality will be added to [`odo preference`](odo-v3-cli/odo-preference.md)
+- `odo preference` - mostly as it is, with additional incorporation of  `odo registry`
+- `odo login` - as it is
+- `odo logout` - as it is
+- `odo build-images` - replaced with `odo build images`
+- `odo deploy` - with new interactive mode
+- `odo storage` -  removed, we can consider adding it back, but users can still add storage manuals into `devfile.yaml`
+- `odo exec` -  removed. We will need to rework this command later
```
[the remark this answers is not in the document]
kadel: removing it only for alpha-1, because it is not essential for main use case, we will add back later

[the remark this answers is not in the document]
kadel: don't understand the question.

docs/proposals/odo-v3-cli.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+
+
+## odo v3 CLI
+
+### general rules for odo cli
+
+- Once even a single flag is provided it ruins in non-interactive mode. All required information needs to be provided via flags
```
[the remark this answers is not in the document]
kadel: > Does this apply to all commands? Are we moving towards having interactive mode for every odo command?

No, not every command will have interactive mode, for some commands it doesn't make sense like for `odo list *`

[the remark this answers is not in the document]
kadel: > Does this apply to all commands? Are we moving towards having interactive mode for every odo command?

No, not every command will have interactive mode, for some commands it doesn't make sense like for `odo list *`

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,83 @@
...
+Running the application on the cluster for **development** (inner loop)
+
+The result of successful execution of `odo dev` is Devfile component deployed in inner loop mode.
+**By default command deletes all resources that it created before it exits.**
+
+When users runs command with `--cleanup=false` there will be no cleanup before existing. The message "Press Ctrl+c to exit and clean up resources from cluster." will be only "Press Ctrl+c to exit."
+
+If user executed `odo dev --cleanup=false` and then run this command again. The first line of the output should display warning: "Reusing already existing resources".
+
+When there is no devfile in the current directory and no flags were provided command starts in interactive mode to guide user through the devfile selection.
```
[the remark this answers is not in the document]
kadel: yes

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,83 @@
...
+[#5299](https://github.com/redhat-developer/odo/issues/5298)
+
+Running the application on the cluster for **development** (inner loop)
+
+The result of successful execution of `odo dev` is Devfile component deployed in inner loop mode.
+**By default command deletes all resources that it created before it exits.**
+
+When users runs command with `--cleanup=false` there will be no cleanup before existing. The message "Press Ctrl+c to exit and clean up resources from cluster." will be only "Press Ctrl+c to exit."
+
+If user executed `odo dev --cleanup=false` and then run this command again. The first line of the output should display warning: "Reusing already existing resources".
```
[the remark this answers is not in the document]
kadel: If you think that it is too early to for implementing this flags, we can delay it to alpha-2 release

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,83 @@
...
+
+When there is no devfile in the current directory and no flags were provided command starts in interactive mode to guide user through the devfile selection.
+
+When some flags were provided and there is no devfile in the current directory command exits with error:
+```
+No devfile.yaml in the current directory. Use `odo create component` to get devfile.yaml for your application."
+```
```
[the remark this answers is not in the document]
kadel: I was talking about `--watch` and `--cleanup` flags

docs/proposals/odo-v3-cli.md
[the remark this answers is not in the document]
dharmit: Ah, okay. IMO, it's very helpful when showing demos or writing docs.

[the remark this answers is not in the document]
dharmit: What led to all these changes in the commands? What problem is odo going to focus on solving in short/medium/long term? For example, services were odo's USP. But in v1alpha1, we are removing them. So what's the roadmap here?

docs/proposals/odo-v3-cli/odo-dev.md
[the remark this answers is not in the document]
dharmit: I do strongly feel that we should put this in alpha2. `odo dev` should be just one thing and doing it well.

docs/proposals/odo-v3-cli/odo-deploy.md
```diff
@@ -0,0 +1,59 @@
+# `odo deploy`
+
+[#5298](https://github.com/redhat-developer/odo/issues/5298)
+
+The result of successful execution of `odo deploy` is Devfile component deployed in outer loop mode.
+
+When there is no devfile in the current directory yet and no flags provided, command runs in  interactive mode to guide user through the devfile selection.
+
+When some flags wre provided and there is no devfile in current directory, command exits with error:
```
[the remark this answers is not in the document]
kadel: I wrote this in case that we add more flags to this command

docs/proposals/odo-v3-cli/odo-preference.md
```diff
@@ -0,0 +1,32 @@
+
+# `odo preference`
+
+[#5402](https://github.com/redhat-developer/odo/issues/5402)
+
+Configures odo behavior like timeouts or registries that odo uses
+
+The behavior is mostly the same as in v2. We can just remove some fields from the preference.
+The `odo registry` command should be integrated into `odo preference` as a `odo preference registry` subcommand.
```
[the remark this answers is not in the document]
kadel: More consistent UX, and make it clear that the registry configuration is part of the odo instance configuration and it is not related to component configuration

docs/proposals/odo-v3-cli/odo-preference.md
```diff
@@ -0,0 +1,32 @@
...
+
+- NamePrefix          - remove
+- BuildTimeout        - remove
+- Timeout             - add explanations
+- PushTimeout         - add explanations
```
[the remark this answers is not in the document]
kadel: Whoever works on this will have to check code what this does. I don't remember it from top of my head.

docs/proposals/odo-v3-cli/odo-list-component.md
```diff
@@ -0,0 +1,36 @@
+
+# `odo list component`
+list devfile components deployed to the cluster in the current namespace.
+
+
+## Flags
+- `--namespace` - list components from the given namespace instead of current namespace.
+- `--path` - find and list all components that are in a given path or in its subdirectories.
```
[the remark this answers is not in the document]
kadel: this is critical functionality to make IDE plugins possible.

docs/proposals/odo-v3-cli/odo-delete-component.md
dharmit: > The user needs to be able to delete components for which there is no local devfile.
> What if I accidentally delete a component locally before deleting it from the cluster?

This, IMO, is an over-thinking scenario. Have we come across users who have faced this problem? As far as IDE users are concerned, they are interacting via IDE, so they are not going to fall into this issue. And for CLI, if a user does `rm -rf`, they can very well do `kubectl delete deployment` as well. If they don't know of this kubectl command, they can open an issue on repo, and we can guide them. Or a support ticket if that's what they would prefer.

IMO, we should focus on 80 part of the 80-20 principle - what matters to 80% of the users and do that.

docs/proposals/odo-v3-cli/odo-list-component.md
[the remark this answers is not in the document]
kadel: I checked the vscode openshift adapter source, it looks like they are not using it currently. Will remove it

docs/proposals/odo-v3-cli/odo-describe-component.md
dharmit: But why does it need to have these extra flags? How are IDE users going to know that there's a component on the cluster which they want to `odo describe`? That's going to happen with `odo list` - I don't know anything else. Then why can't we just `odo describe <component-name>` instead of taking these extra flags? The `--application` flag has really not served us any purpose. If they want to describe a component from a different project, what's stopping plugins from doing `odo project set <another-project>; odo describe <comp-name>` internally?

docs/proposals/odo-v3-cli/odo-list-component.md
```diff
@@ -0,0 +1,36 @@
...
+  NAME            APPLICATION    TYPE         MANAGED B   RUNNING IN  PATH
+  frontend        myapp          nodejs       odo         Dev         frontend
+  backend         myapp          springboot   odo         Deploy      backend
+  backend         myapp          springboot   odo         None        asdf
+
+```
+
+- row/component marked with `*` at the begging of the line is the one that is also in the current directory.
+- `TYPE` corresponds to the `langauge` field in `devfile.yaml` tools, this should also correspond to `odo.dev/project-type` label.
+- `RUNNING IN` indicates in what modes the component is running. `Dev` means the component is running in development mode (`odo dev`). `Deploy` indicates that the component is running in deploy mode (`odo deploy`), `None` means that the component is currently not running on cluster. `Unknown` indicates that odo can't detect in what mode is component running. `Unknown` will also be used for component that are running on the cluster but are not managed by odo.
```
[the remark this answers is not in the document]
kadel: `odo deploy` is not long-running command, and doesn't have `--cleanup` flag.
If user executed `odo dev --cleanup=false` than this column would contain "Dev"

docs/proposals/odo-v3-cli/odo-describe-component.md
dharmit: My point is to keep odo layer simple and slim. Let's not imagine the scenarios that IDE plugins or other tooling needs and create flags/commands for that in odo. If the same thing can be done by running 2 odo commands, why do we want to add an extra command in odo and make it a maintenance overhead on odo?

docs/proposals/odo-v3-cli/odo-list-component.md
```diff
@@ -0,0 +1,36 @@
...
+- `PATH` column is displayed only if the command was executed with `--path` flag. It shows the path in which the component "lives". This is relative path to a given `--path`.
```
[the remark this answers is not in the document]
kadel: No, because in "normal" component listing from cluster we don't know the path.

docs/proposals/odo-v3-cli/odo-describe-component.md
dharmit: OK. Please mark as resolved when it's done. Thanks.

docs/proposals/odo-v3-cli/odo-create-component.md
```diff
@@ -0,0 +1,66 @@
+
+# `odo create component`
+
+
+
+Create new Devfile component in the current directory.
+Some users might prefer creating components separately before running `odo dev` or `odo deploy`.
```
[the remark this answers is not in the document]
kadel: we don't need `odo create` we can use `odo init` instead

docs/proposals/odo-v3-cli/odo-create-component.md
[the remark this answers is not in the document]
kadel: if we integrate this to `odo init` it can be used for both

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: The idea is that we want to make sure that odo can be useful for users that don't want to use "Devfile" yet. 

This is something that we discussed for a long time. And If I remember you agreed with this.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: Odo needs to be able to delete component form the cluster even without having access to a devfile.yaml that was originally used to create that component.

If you have a better way how to achieve that I'm open to suggestions.

docs/proposals/odo-v3-cli.md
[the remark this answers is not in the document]
kadel: odo's goals is still the same. The short team target is to provide a better onboarding experience and better experience when working with odo components.
Medium-term goal is to figure out how to make easy for users to work (deploy and connect) with services.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: >This, IMO, is an over-thinking scenario. Have we come across users who have faced this problem?

Yes you are talking to one.

>. And for CLI, if a user does rm -rf, they can very well do kubectl delete deployment

It is not as easy as doing `kubectl delete deployment` odo component will typically create a lot more resources.

> IMO, we should focus on 80 part of the 80-20 principle - what matters to 80% of the users and do that.

This functionality is basic requirement that needs to be there.

docs/proposals/odo-v3-cli/odo-describe-component.md
kadel: >But why does it need to have these extra flags? How are IDE users going to know that there's a component on the cluster which they want to odo describe

`odo list components`


>I don't know anything else. Then why can't we just odo describe <component-name> instead of taking these extra flags? 

that is not enough, component with the same name can be present in multiple applications

> If they want to describe a component from a different project, what's stopping plugins from doing odo project set <another-project>; odo describe <comp-name> internally?

`odo project set <another-project>;` would modify the currently active namespace in KUBECONFIG and it would affect all other `odo` `kubectl` or `oc` commands running at the same time.

docs/proposals/odo-v3-cli/odo-describe-component.md
kadel: It would be nice if you could help me with that.

docs/proposals/odo-v3-cli/odo-delete-component.md
dharmit: > The idea is that we want to make sure that odo can be useful for users that don't want to use "Devfile" yet.

But if they already have a component that they want to delete, aren't they using devfile already?

> This is something that we discussed for a long time. And If I remember you agreed with this.

Yes, I remember agreeing to it. But my concern right now is to keep odo slim and less bloated, and add features only when "real" users ask for something. That makes development faster, and maintenance simpler.

Do you truly feel that odo is ready to support a scenario where devfile.yaml doesn't exist, and each "create" command would create a manifest file and then `odo push`/`odo dev` would create those resources on the cluster? I don't think we have discussed that sort of flow in enough detail to start working on it for v3alpha1.

Also, I'll repeat the same question/point that I have been annoying you with since a long time - did any "real" user(s) request for this or is it just our imagination that some users would like to have this?

docs/proposals/odo-v3-cli/odo-delete-component.md
dharmit: > Odo needs to be able to delete component form the cluster even without having access to a devfile.yaml that was originally used to create that component.

OK. I won't annoy you with "did a user ask for it", because I want to be able to do "constructive criticism" instead of "ranting".



> If you have a better way how to achieve that I'm open to suggestions.

For a component "odcComponent" created by ODC in the namespace "myproject", user can delete it by doing:
```sh
$ odo project set myproject
$ odo delete component odcComponent -f
```
No use of `--component` or `--application` or `--namespace` flags there. It becomes a two-step process, but it is succinct and clear, and at code level it's simpler to maintain.

docs/proposals/odo-v3-cli/odo-delete-component.md
dharmit: > Yes you are talking to one.

Ah, that's cool. I didn't know that you used odo for creating microservices using Spring/NodeJS/Postgres.

> It is not as easy as doing `kubectl delete deployment` odo component will typically create a lot more resources.

odo makes sure to tie everything to a deployment by using `ownerReferences`. So, unless I am mistaken, it is actually as simple as doing `kubectl delete deployment`.

> This functionality is basic requirement that needs to be there.

Disagree. But https://github.com/redhat-developer/odo/pull/5341#discussion_r796302382 should help with it. My request is to not bloat the commands with flags just because we can do it. Instead, add flags consciously when a functionality can't be achieved otherwise. I don't agree in doing this because the IDE plugins need it, either. They can easily execute a task by running two commands instead of one, no?

docs/proposals/odo-v3-cli/odo-describe-component.md
dharmit: > that is not enough, component with the same name can be present in multiple applications

And I'm yet to see a sensible use of application flag happen. It's pretty easy for users to create different projects/namespaces to keep their stuff separate. Why are we still sticking with application label?



> `odo project set <another-project>;` would modify the currently active namespace in KUBECONFIG and it would affect all other `odo` `kubectl` or `oc` commands running at the same time.

I thought that IDE plugins were to use odo exclusively to communicate with a k8s/ocp cluster. If they are using `kubectl`/`oc` for other things, what's even the purpose of having odo for them? Why can't they do all communication using `kubectl`/`oc`? Rant apart, I see why `--namespace` flag could be useful at times. But `--component` and `--application` flag don't make sense to me. I would rather have them scrapped.

docs/proposals/odo-v3-cli/odo-describe-component.md
dharmit: Sure.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: > But if they already have a component that they want to delete, aren't they using devfile already?

in case of `odo describe component`, the component will probably be always created based on the devfile, but that devfile might not be available to odo.


> Also, I'll repeat the same question/point that I have been annoying you with since a long time - did any "real" user(s) request for this or is it just our imagination that some users would like to have this?

Yes, this is based on the feedback from users.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: >OK. I won't annoy you with "did a user ask for it", because I want to be able to do "constructive criticism" instead of "ranting".

yes, the same as for `odo describe` this is based on user feedback.


I've remove `--aplication` flag from all commands, for now. But `--name` and `--namespace` needs to be there.

docs/proposals/odo-v3-cli/odo-delete-component.md
kadel: >> It is not as easy as doing kubectl delete deployment odo component will typically create a lot more resources.

> odo makes sure to tie everything to a deployment by using ownerReferences. So, unless I am mistaken, it is actually as simple as doing kubectl delete deployment.

Yes right now it does. But as we discussed many times this is not a good design that needs to change, for example, we want to support "service-only" components in the future.

docs/proposals/odo-v3-cli/odo-describe-component.md
kadel: > And I'm yet to see a sensible use of application flag happen. It's pretty easy for users to create different projects/namespaces to keep their stuff separate. Why are we still sticking with application label?

I've completely removed application support from alpha1


>I thought that IDE plugins were to use odo exclusively to communicate with a k8s/ocp cluster. If they are using kubectl/oc for other things, what's even the purpose of having odo for them? Why can't they do all communication using kubectl/oc? Rant apart, I see why --namespace flag could be useful at times. But --component and --application flag don't make sense to me. I would rather have them scrapped.

removed `--application` flag for now. Component needs to be there (I just renamed it to `--name`)

docs/proposals/odo-v3-cli/odo-delete-component.md
feloy: The flags `--namespace`, `--context`, and similar are flags that modify the "context" of the application. 
One solution would be to let the user do these changes of context itself (cd $CONTEXT for `--context`, kubectl ns $NAMESPACE for `--namespace`, etc).

If we want to efficiently integrate these flags in odo, we should consider them just for this change of context: instead of polluting the code with passing the values of these flags, we should instead change the context as soon as possible in the flow of the code (probably long before we enter the code of a specific command), and forget them for the code we are working on in a daily basis.

As an example, most of the code can consider odo works on local directory, and you can change the current working directory at the very beginning (and don't forget to change it back before to quit).

dharmit: > for example, we want to support "service-only" components in the future.

I wager it'll require doing a v4, because in v3 our main focus is components. For example, `odo init` bootstraps a component. Component, based on what I understand, is the most basic unit of odo.

kadel: `odo init` doesn't need to know how to create "service-only" components. `odo dev` will just need to understand how to create resources out of it,  `odo list/describe/delete` will also have to understand it.

We know about this requirement and we are building v3, so we need to consider this so it can be incorporated in the future without requiring v4.

dharmit: If we were to make `odo dev` ready for this, it should create cluster resources for the components that are not of type `image`. And then it should execute the commands for these components. But there has to be some order in which the commands are run, right? Does API define that? 

What else should I be thinking from devfile point of view?

kadel: >If we were to make odo dev ready for this, it should create cluster resources for the components that are not of type image. And then it should execute the commands for these components. But there has to be some order in which the commands are run, right? Does API define that?

yes

dharmit: Anything else you think we should be keeping in mind? I'm asking this as someone who finds it sorta difficult to figure things out from devfile API.

docs/proposals/odo-v3-cli/odo-list.md
```diff
@@ -0,0 +1,29 @@
...
+
+## example
+```
+$ odo list
+Components in the "mynamspace" namespace:
+
+  NAME              TYPE         MANAGED BY       RUNNING IN
+* frontend          nodejs       odo              Dev,Deploy
+  backend           springboot   odo              Deploy
+  created-by-odc    python       Unknown          Unknown
```
dharmit, line 16: I think that `TYPE` here should be component, endpoint, service, etc. because `odo list`, as per the discussion we had on call during the past couple of days, is supposed to list all the kinds of resource.

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,91 @@
+# `odo dev`
+
+[#5299](https://github.com/redhat-developer/odo/issues/5298)
+
+Running the application on the cluster for **development** (inner loop)
+
+The result of successful execution of `odo dev` is Devfile component deployed in inner loop mode.
+**By default command deletes all resources that it created before it exits.**
+
+When users runs command with `--cleanup=false` there will be no cleanup before existing. The message "Press Ctrl+c to exit and clean up resources from cluster." will be only "Press Ctrl+c to exit."
```
valaparthvi, line 10:
```suggestion
When users runs command with `--cleanup=false` there will be no cleanup before exiting. The message "Press Ctrl+c to exit and clean up resources from cluster." will be only "Press Ctrl+c to exit."
```

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,91 @@
+# `odo dev`
+
+[#5299](https://github.com/redhat-developer/odo/issues/5298)
+
+Running the application on the cluster for **development** (inner loop)
+
+The result of successful execution of `odo dev` is Devfile component deployed in inner loop mode.
+**By default command deletes all resources that it created before it exits.**
```
valaparthvi, line 8: Why? And what resources exactly? Does it also delete the deployments that were created for this devfile component, or just the URL, storages, services that were created while `odo dev` was executing?
And what happens to URL, storages, and services that were present when `odo dev` began executing? Will those be deleted as well?
And if we also delete deployments, then what is the point of having `odo list` for component's pushed via `odo dev`?
Or will all of this make sense when we implement `--cleanup` flag?

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,91 @@
...
+If user executed `odo dev --cleanup=false` and then run this command again. The first line of the output should display warning: "Reusing already existing resources".
+
+When there is no devfile in the current directory and no flags were provided command starts in interactive mode to guide user through the devfile selection.
+
+When some flags were provided and there is no devfile in the current directory command exits with error:
+```
+No devfile.yaml in the current directory. Use `odo init component` to get devfile.yaml for your application."
+```
+
+When devfile exists in the current directory deploy application in inner loop mode using the information form `devfile.yaml` in the current directory.
```
valaparthvi, line 21:
```suggestion
When devfile exists in the current directory deploy application in inner loop mode using the information from `devfile.yaml` in the current directory.
```

docs/proposals/odo-v3-cli/odo-dev.md
```diff
@@ -0,0 +1,91 @@
...
+
+`odo dev` commands sets up port-forwarding 
+localhost.
+
+
+
+## Flags
+
+- `-o` (string) output information in a specified format (json).
+- `--watch` (boolean) Run command in watch mode. In this mode command is watching the current directory for file changes and automatically sync change to the container where it rebuilds and reload the application.
```
valaparthvi, line 32: If `--watch=false`, then why do we need to continuously run `odo dev`? Wouldn't it just function like `odo push`?

docs/proposals/odo-v3-cli/odo-list-namespace.md
```diff
@@ -0,0 +1,9 @@
+
+# `odo list namespace`
+Lists all namespaces that user has access to.
+This is the same as `kubectl get namespace` or `oc projects`
+The output should indicate the active namespace.
+
```
valaparthvi, line 6:
```suggestion

## example

```sh
$ odo project list
ACTIVE     NAME
           cmd-devfile-app-test123ane
           cmd-devfile-app-test123thr
*          default
           kube-node-lease
           kube-public
           kube-system
```

docs/proposals/odo-v3-cli/odo-list.md
```diff
@@ -0,0 +1,29 @@
+# `odo list`
+
+list all odo related information from the current namespace
+
+## Flags
+- `--namespace` - list components from the given namespace instead of current namespace.
+
+## example
+```
+$ odo list
```
valaparthvi, line 10: I know we have discussed this before, but I keep circling back to thinking that this is an inconsistent behavior :/

docs/proposals/odo-v3-cli/odo-init.md
```diff
@@ -0,0 +1,138 @@
...
+    At this point, the previous answers should be enough to uniquely select one Devfile from registry.
+    List of all starter projects defined in selected devfile.
+
+4. **"Enter component name:"**
+    Name of the component. This should be saved in the local `devfile.yaml` as a value for `metadata.name` field.
+
+### When executed in non-empty directory
+
+```
+$ odo create
```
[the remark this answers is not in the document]
kadel: yes, typo

docs/proposals/odo-v3-cli/odo-init.md
```diff
@@ -0,0 +1,138 @@
...
+- FOO1 = bar1
+
+? What configuration do you want change?  [Use arrows to move, type to filter]
+> NOTHING - configuration is correct
+  Delete port "8080"
+  Delete port "8084"
+  Add new port
+  Delete environment variable "FOO"
+  Delete environment variable "FOO1"
+  Add new environment variable
```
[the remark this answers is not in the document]
kadel: we have merged `odo create` and `odo init` into one command `odo init`.
Now `odo init` can be used to create/initialize component with the existing code base. For this the configuration is required.

docs/proposals/odo-v3-cli/odo-list-component.md
```diff
@@ -0,0 +1,25 @@
...
+list devfile components deployed to the cluster in the current namespace.
+
+
+## Flags
+- `--namespace` - list components from the given namespace instead of current namespace.
+
+## example
+```
+$ odo list components
+Components in the "mynamspace" namespace:
```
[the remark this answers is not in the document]
kadel: yes

docs/proposals/odo-v3-cli/odo-list-component.md
```diff
@@ -0,0 +1,25 @@
...
+
+  NAME            APPLICATION    TYPE         MANAGED BY       RUNNING IN
+* frontend        myapp          nodejs       odo              Dev,Deploy
+  backend         myapp          springboot   odo              Deploy
+  created-by-odc  asdf           python       Unknown          Unknown
```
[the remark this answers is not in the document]
kadel: Running `None` can be only for component in the current directory marked with `*`

docs/proposals/odo-v3-cli/odo-dev.md
kadel: yes

docs/proposals/odo-v3-cli/odo-list.md
kadel: > I know we have discussed this before, but I keep circling back to thinking that this is an inconsistent behavior :/

Feel free to suggest better alternative

kadel:
I'm moving this to https://docs.google.com/document/d/1I0ncXurBat81PB99jiNGhWXMIJ8NgaDXgNxODOfNSbw/edit# as we are starting alpha2 discussions it is becoming hard to easily collaborate on this PR

PR #5342, opened 2022-01-12
Move envvar to specific package

### LINKED ISSUE #4783: Support multi-line values for environment variables.
/kind feature

`odo config set --env ...` should support multi-line values.

## Why is this needed?

Sometimes users need to pass in values with newline characters to environment variables, for example, TLS certificates. It would be great if we could use something like:

```
odo config set --env SMTP_TLS_CERT="$(cat client.cert)"
```

Rather than this ugly work-around:

```
sed -i devfile.yaml -e "/- name: SMTP_TLS_CERT/{n;s|value: .*|value: $(jq -Rs . < client.cert | sed 's/\\/\\\\/g')|}"
```

Currently using the first approach sets the value to only the first line of the file and the rest of the file is mysteriously ignored.

feloy:
**What type of PR is this?**

/kind cleanup

- Move Envvar functions from config package to own package.
- Make multi-line env variables possible (#4783)
- Removes the possibility to use --env a=b,c=d instead of `--env a=b --env c=d` for `odo config set`
- Checks the name of env variables with the regexp `^[a-zA-Z_][a-zA-Z0-9_]*$`

By replacing `cobra.StringSliceVar` by `cobra.StringArrayVar`, we can pass multi-line values, but cannot use `--env a=b,c=d` anymore

Part of #5247 
Fixes #4783 

**PR acceptance criteria**:

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [x] I have read the [test guidelines](https://github.com/redhat-developer/odo/blob/master/docs/dev/test-architecture.adoc)


pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+			Name:  ev.Name,
+			Value: ev.Value,
+		})
+	}
+	return envList
+}
+
+// newFromString takes a string of format "name=value" and returns an Env
+// variable struct
+func newFromString(envStr string) (EnvVar, error) {
```
valaparthvi, line 34: Can you move this private method to the bottom?

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
+// envvar package helps converting name/value pairs to devfile EnvVars
```
valaparthvi, line 1:
```suggestion
// envvar package helps converting name/value pairs to devfile EnvVar
```

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+// EnvVar represents an environment variable
+type EnvVar struct {
+	Name  string `yaml:"Name"`
+	Value string `yaml:"Value"`
+}
+
+// List represents a list of environment variables
+type List []EnvVar
+
+// ToDevfileEnv converts the EnvVarList to the list of Envs supported by devfile
```
valaparthvi, line 20:
```suggestion
// ToDevfileEnv converts the List to the list of devfile EnvVar
```

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+func (evl List) ToDevfileEnv() []devfilev1.EnvVar {
```
valaparthvi, line 21:
```suggestion
func (evl List) ToDevfileEnvVar() []devfilev1.EnvVar {
```

I wonder if it makes more sense to rename the method to ToDevfileEnvVar since it is returning devfile EnvVar.

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+			Value: ev.Value,
+		})
+	}
+	return envList
+}
+
+// newFromString takes a string of format "name=value" and returns an Env
+// variable struct
+func newFromString(envStr string) (EnvVar, error) {
+	envList := strings.SplitN(envStr, "=", 2)
```
valaparthvi, line 35: :100:
I hadn't understood the use of `SplitN` until now.

valaparthvi: But using this also means that env vars such as `a==b` will split into [a, =b] or `a!=b` will split into [a!, b]. I don't think these can be considered as valid cases, can they? I understand these are corner cases and not likely to be encountered, but do you think we should cover them?

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+
+	devfilev1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+// EnvVar represents an environment variable
+type EnvVar struct {
+	Name  string `yaml:"Name"`
+	Value string `yaml:"Value"`
+}
```
valaparthvi, line 15: Why did you define a separate struct if we already have devfile EnvVar? Is it because of the yaml keys?

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil, err
+		}
+		envVarList = append(envVarList, envVar)
+	}
+
+	return envVarList, nil
+
+}
+
+// NewListFromDevfileEnv creates an EnvVarList from the list of envs present in a devfile.
```
valaparthvi, line 63:
```suggestion
// NewListFromDevfileEnv creates an List from the list of envs present in a devfile.
```
Using `List` so that it is referenced correctly.

valaparthvi (changes requested):
I understand that you've simply moved the files into a separate pkg, but it would be nice if you can include some more changes.

pkg/envvar/env_var.go
feloy: For the first one, a = '=b', it is a valid case for me. We don't have to limit the possible values to pass to the container, the value '=b' seems completely legit to me.
For the second one, a! = b, I'm not sure that a! is a valid name for an environment variable. Looking more precisely to it, it is not valid: https://www.tutorialspoint.com/unix/unix-using-variables.htm

pkg/envvar/env_var.go
feloy: Yes, it seems it has been done to have `Name` and `Value` written capitalized in the output of `odo config view`.

feloy:
@kadel What do you think would be more important (because we cannot have both)?

- being able to have multi line values for env variables (to fix #4783)
- being able to continue to use the short syntax `odo config set --env a=b,c=d` instead of `odo config set --env a=b --env c=d` (which is a non documented feature)

kadel:
> * being able to continue to use the short syntax `odo config set --env a=b,c=d` instead of `odo config set --env a=b --env c=d` (which is a non documented feature)

We don't need short syntax. I think that `--env a=b --env c=d` is completely fine.

pkg/envvar/env_var.go
```diff
@@ -0,0 +1,81 @@
...
+
+// newFromString takes a string of format "name=value" and returns an EnvVar
+func newFromString(envStr string) (EnvVar, error) {
+	envList := strings.SplitN(envStr, "=", 2)
+	// if there is not = in the string
+	if len(envList) < 2 {
+		return EnvVar{}, errors.New("invalid environment variable format")
+	}
+	name := strings.TrimSpace(envList[0])
+	match, err := regexp.MatchString("^[a-zA-Z_][a-zA-Z0-9_]*$", name)
```
valaparthvi, line 70: Can you add unit tests to TestNewEnvVarFromString to cover this new addition?

pkg/odo/cli/config/set.go
```diff
@@ -205,7 +206,7 @@ func NewCmdSet(name, fullName string) *cobra.Command {
 		},
 	}
 	configurationSetCmd.Flags().BoolVarP(&o.forceFlag, "force", "f", false, "Don't ask for confirmation, set the config directly")
-	configurationSetCmd.Flags().StringSliceVarP(&o.envArrayFlag, "env", "e", nil, "Set the environment variables in config")
+	configurationSetCmd.Flags().StringArrayVarP(&o.envArrayFlag, "env", "e", nil, "Set the environment variables in config")
```
valaparthvi, line 209: Thoughts on adding an integration test testing multi-line value support for this one?

valaparthvi (changes requested):
The changes look good, I just have 2 questions about tests.

feloy:
@valaparthvi, thanks for your review. I've added the missing tests

PR #5363, opened 2022-01-19
Add server flag to odo login

### LINKED ISSUE #5189: suggestion: make login command accept the same arguments as 'oc'
This is a UX suggestion.

The OpenShift Web UI provides the option to 'Copy login command'.

This provides a user with something copy-pastable like:
```
oc login --token=deadbeef --server=https://myserver:6443
```

It would be nice if the `oc` command from the above line can be simply substituted by `odo`, so the that the following would log into the cluster:

```
odo login --token=deadbeef --server=https://myserver:6443
```

Currently, `odo` requires to rewrite these arguments as:
```
odo login https://myserver:6443 --token=deadbeef
```

cc @kadel

### LINKED ISSUE #5345: cleanup pkg/auth
- [ ] refactor based on the https://github.com/redhat-developer/odo/issues/5247
- [ ] fix https://github.com/redhat-developer/odo/issues/5189

dharmit:
**What type of PR is this:**
/kind bug

Adds `--server` flag to `odo login`. We need it to be able to use `oc login` commands as-is by replacing `oc` with `odo`.

Fixes #5345 
Fixes #5189 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Try logging into an OpenShift server using different ways.
* Using token and server with `--server` flag
* Using token and passing server as a parameter (do not use `--server` flag)
* Any other login combination that you can think of. If it works with `oc` but fails with `odo`, please flag it.


dharmit:
@feloy @valaparthvi I couldn't find anything that required changing from "separating layers" point of view. If you can think of anything, please let me know.

pkg/odo/cli/login/login.go
```diff
@@ -49,19 +50,31 @@ func NewLoginOptions() *LoginOptions {
...
 	return
 }
 
 // Validate validates the LoginOptions based on completed values
 func (o *LoginOptions) Validate() (err error) {
+	if o.server != "" && o.serverFlag != "" && o.server != o.serverFlag {
+		// if user has passed server value as parameter as well as used --server flag:
+		// * odo errors *if* the values are different
+		// * odo silently continues if the values are same
+		return fmt.Errorf("either use --server flag or pass server link as a paremeter, don't use both")
```
valaparthvi, line 67:
```suggestion
		return fmt.Errorf("either use --server flag or pass server link as a parameter, don't use both")
```

pkg/odo/cli/login/login.go
```diff
@@ -86,5 +99,6 @@ func NewCmdLogin(name, fullName string) *cobra.Command {
 	loginCmd.Flags().StringVarP(&o.tokenFlag, "token", "t", "", "token, will prompt if not provided")
 	loginCmd.Flags().BoolVar(&o.skipTlsFlag, "insecure-skip-tls-verify", false, "If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure")
 	loginCmd.Flags().StringVar(&o.caAuthFlag, "certificate-authority", "", "Path to a cert file for the certificate authority")
+	loginCmd.Flags().StringVar(&o.serverFlag, "server", "", "OpenShift server to log into")
```
valaparthvi, line 102:
```suggestion
	loginCmd.Flags().StringVar(&o.serverFlag, "server", "", "OpenShift/Kubernetes Server host to log into")
```

valaparthvi (changes requested):

pkg/odo/cli/login/login.go
dharmit: I didn't put Kubernetes because I don't remember logging into vanilla Kubernetes using any kind of login command. If you see, `kubectl` doesn't have a `login` command to it. This is OpenShift specific thingy, IMO.

valaparthvi (commented):
The code lgtm, but do you think we can a test for this?

dharmit:
> The code lgtm, but do you think we can a test for this?

I don't think so. That's why I haven't added any. It was discussed in the past as well when we were working on login, and we decided to not have tests for it.

feloy:
> @feloy @valaparthvi I couldn't find anything that required changing from "separating layers" point of view. If you can think of anything, please let me know.

You could create a Client interface in the `pkg/auth` package containing the method `Login`, so you could create a mock implementation of it for testing the logic inside `odo/cli/login`: you could test the value of the `server` parameter passed to the Login method depending on the different possible values for server flag and parameter.

dharmit:
> @dharmit You could create a Client interface in the `pkg/auth` package containing the method `Login`

But `Login` is a package level function. It's not a method of any type. How do you suggest creating a client interface and adding test for it? Genuine question because I'm lost here. 😕

dharmit:
@feloy Added the tests after our discussion. I don't see a possibility to add tests for the `Run` method for `LoginOptions`. Correct me if I am wrong.

dharmit:
```
[odo] Pushing devfile component "wzivdw"
[odo]  ✓  Changes successfully pushed to component
[odo] I0124 05:57:19.296524   35404 odo.go:75] Could not get the latest release information in time. Never mind, exiting gracefully :)
Running oc with args [oc get pods --namespace cmd-devfile-push-test815tpr --selector=component=wzivdw -o jsonpath={.items[*].metadata.name}]
[oc] wzivdw-app-68f6b9b745-84zrjRunning oc with args [oc exec wzivdw-app-68f6b9b745-84zrj -c runtime --namespace cmd-devfile-push-test815tpr -- ps -ef]
[oc] Signal 23 (URG) caught by ps (3.3.15).
[oc] ps:ps/display.c:66: please report this bug
[oc] command terminated with exit code 1
Deleting project: cmd-devfile-push-test815tpr
Running odo with args [odo project delete cmd-devfile-push-test815tpr -f] 
```
Probably a flake.
/test v4.9-integration-e2e

feloy:
> @feloy Added the tests after our discussion. I don't see a possibility to add tests for the `Run` method for `LoginOptions`. Correct me if I am wrong.

Yes, it could be possible to create a test for thr Run method. Not very useful, as the method is very simple, but that could be a good first example with mocking methods

dharmit:
> but that could be a good first example with mocking methods

I honestly can't imagine a scenario to test for it. Could you shed some light?

pkg/auth/login.go
```diff
@@ -13,9 +13,10 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-// Login takes of authentication part and returns error if there any
-func Login(server, username, password, token, caAuth string, skipTLS bool) error {
+type KubernetesClient struct{}
```
valaparthvi, line 16: Can you define a factory function to return this object? [Example](https://github.com/redhat-developer/odo/blob/main/pkg/application/kubernetes.go#L20).

valaparthvi:
@kadel https://github.com/redhat-developer/odo/runs/4980351617?check_suite_focus=true is missing SONAR_TOKEN.

PR #5364, opened 2022-01-19
odo.dev: reorder getting-started section

kadel:
**What type of PR is this:**

/kind documentation

Installation should be listed before quickstart


cdrage:
Makes sense! I agree with the new order

/lgtm
/approve

PR #5378, opened 2022-01-25
Fix sonarcloud

kadel:
fixes SonarCloud integration

/kind tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
     types: [opened, synchronize, reopened]
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
+name: Build
+on:
+  push:
+    branches:
+      - master
```
valaparthvi, line 5: Shouldn't this be main?

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
+name: Build
+on:
+  push:
+    branches:
```
valaparthvi, line 4: When would this be triggered?

kadel: yes, my bad

.github/workflows/build.yml
kadel: yes, my bad

valaparthvi (approved):

kadel:
no point in testing this
/override v4.9-integration-e2e

PR #5318, opened 2022-01-06
Preserving devfile provided by starter project

mohammedzee1000:
Signed-off-by: Mohammed Zeeshan Ahmed <mzee1000@work.mzee1000.io>

**What type of PR is this?**

/kind feature

**What does this PR do / why we need it**:

This PR ensures that if a specified starter project provides
its own devfile, then the same is preserved during create
instead of being overridden.

**Which issue(s) this PR fixes**:

Fixes #5200

**PR acceptance criteria**:

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation 

- [x] I have read the [test guidelines](https://github.com/redhat-developer/odo/blob/master/docs/dev/test-architecture.adoc)

**How to test changes / Special notes to the reviewer**:

####  When --devfile and --starter are provided:

```bash
 $ odo create nodejs --devfile ../../odo/tests/examples/source/devfiles/nodejs/devfile-with-starter-with-devfile.yaml --starter nodejs-starter
 ⚠  Found a dangling env file without a devfile, overwriting it
Devfile Object Creation
 ✓  Creating a devfile component from devfile path: /home/mzee1000/workbench/work/odo/odo/tests/examples/source/devfiles/nodejs/devfile-with-starter-with-devfile.yaml [130659ns]
Validation
 ✓  Validating if devfile name is correct [22475ns]
 ✓  Validating the devfile for odo [6ms]

Starter Project
 ✓  Downloading starter project nodejs-starter from https://github.com/nodeshift-starters/devfile-sample.git [671ms]

Please use `odo push` command to create the component with source deployed

$ ls       
devfile.yaml  Dockerfile  LICENSE  outerloop-deploy.yaml  package.json  package-lock.json  README.md  server.js  test

$ cat devfile.yaml  
schemaVersion: 2.2.0
metadata:
  name: nodejs
  version: 1.0.1
  displayName: Node.js Runtime
  description: Stack with Node.js 14
  tags: ["NodeJS", "Express", "ubi8"]
  projectType: "nodejs"
  language: "nodejs"
  attributes:
    alpha.dockerimage-port: 3001
  provider: Red Hat
  supportUrl: https://github.com/devfile-samples/devfile-support#support-information
components:
  - name: outerloop-build
    image:
      imageName: nodejs-image:latest
      dockerfile:
        uri: Dockerfile
        buildContext: .
        rootRequired: false
  - name: outerloop-deploy
    kubernetes:
      uri: outerloop-deploy.yaml
  - name: runtime
    container:
[...]

PR #5414, opened 2022-02-02
More inline help for odo init

feloy:
**What type of PR is this:**

/kind bug

Based on feedback from sprint 213 demo, added clearer error messages and more examples

### DIFF
--- a/pkg/odo/cli/init/init.go
+++ b/pkg/odo/cli/init/init.go
@@ -35,7 +35,22 @@ const RecommendedCommandName = "init"
 var initExample = templates.Examples(`
   # Boostrap a new component in interactive mode
   %[1]s
-`)
+
+  # Bootstrap a new component with a specific devfile from registry
+  %[1]s --name my-app --devfile nodejs
+  
+  # Bootstrap a new component with a specific devfile from a specific registry
+  %[1]s --name my-app --devfile nodejs --devfile-registry MyRegistry
+  
+  # Bootstrap a new component with a specific devfile from the local filesystem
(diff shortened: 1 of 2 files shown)


pkg/odo/cli/init/init.go
```diff
@@ -35,7 +35,22 @@ const RecommendedCommandName = "init"
...
+  %[1]s --name my-app --devfile nodejs
+  
+  # Bootstrap a new component with a specific devfile from a specific registry
+  %[1]s --name my-app --devfile nodejs --devfile-registry MyRegistry
+  
+  # Bootstrap a new component with a specific devfile from the local filesystem
+  %[1]s --name my-app --devfile-path $HOME/devfiles/nodejs/devfile.yaml
+  
+  # Bootstrap a new component with a specific devfile from the web
+  %[1]s --name my-app --devfile-path https://devfiles.example.com/nodejs/devfile.yaml
```
dharmit, line 49: Not directly related to the changes in this PR, but a question popped up in my head. Is it possible to preserve the starter project provided by devfile, a la #5318, when using `odo init`? Should it be possible? If yes, is it relevant for v3alpha1? IMO, no. The command I am talking about is:

```sh
$ odo init --name my-app --devfile-path https://devfiles.example.com/nodejs/devfile.yaml --starter nodejs-starter
```

feloy: Yes, the implementation takes care of the case #5318.

https://github.com/redhat-developer/odo/blob/main/pkg/odo/cli/init/init.go#L158-L180

PR #5427, opened 2022-02-02
WIP: Dependency injection using Cobra command annotation

feloy:
**What type of PR is this:**

/kind cleanup

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/odo/genericclioptions/dependencies/dependencies.go
```diff
@@ -0,0 +1,111 @@
...
+		dep.KubernetesClient, err = kclient.New()
+		if err != nil && isDefined(command, KUBERNETES) {
+			return nil, err
+		}
+	}
+	if isDefined(command, APPLICATION) {
+		dep.ApplicationClient = application.NewClient(dep.KubernetesClient)
+	}
+
+	/* Instantiate new dependencies here. Take care to instantiate after all sub-dependencies */
```
dharmit, line 109: But we aren't instantiating new dependencies here, are we? IIUC, it happens at the beginning of this function. I can't understand what this comment means. 🤔

pkg/odo/genericclioptions/runnable.go
```diff
@@ -81,6 +83,13 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 
 	// LogErrorAndExit is used so that we get -o (jsonoutput) for cmds which have json output implemented
 	util.LogErrorAndExit(checkConflictingFlags(cmd, args), "")
+
+	deps, err := dependencies.Fetch(cmd)
+	if err != nil {
+		util.LogErrorAndExit(err, "")
+	}
+	o.SetDependencies(deps)
```
dharmit, line 91: How about using "clientset" instead of "dependencies"? Suggesting this because I have come across it in other code bases.

dharmit (changes requested):

pkg/odo/genericclioptions/dependencies/dependencies.go
feloy: But yes, talking about clients and clientset would be more understandable I agree

pkg/odo/cli/init/init.go
```diff
@@ -74,12 +67,15 @@ type InitOptions struct {
...
+	return &InitOptions{}
+}
+
+func (o *InitOptions) SetDependencies(dependencies *dependencies.Dependency) {
+	o.dependencies = dependencies
+	o.backends = []params.ParamsBuilder{
+		params.NewFlagsBuilder(),
+		params.NewInteractiveBuilder(asker.NewSurveyAsker(), catalog.NewCatalogClient(dependencies.FS, dependencies.PreferenceClient)),
 	}
 }
```
valaparthvi, line 80: If `SetDependencies` does everything, then do we need `New<CMD>Options` method?

pkg/odo/genericclioptions/dependencies/dependencies.go
dharmit: > I mean "Instantiate new _clients_ here": if a developer creates a new package with a client and wants to add it to this package, he has to instantiate the client in this function

Ah, I just realized that this is more like "to add more stuff, do it here".

pkg/odo/cli/init/init.go
dharmit: I think `SetClientset`/`SetDependencies` is only for initializing the clients needed by a command. `New<CMD>Options` do much more than that. There could/would be command specific fields in `<CMD>Options` struct which can't be initialized with `SetClientset`. It's out of its scope.

@feloy correct me if I am wrong.

feloy: Yes, right, we are removing the initialization of the clients here, but there could be other initializations (not depending on clients).

But I also agree that in most cases, this initialization will be empty.

For consistency, let's use the New<Cmd>Options() form, event it is doing nothing more than creating an empty struct.

dharmit: > For consistency, let's use the NewOptions() form, event it is doing nothing more than creating an empty struct.

+1. Similarly, we are going to have empty `SetClientset` for many commands as well.

pkg/odo/genericclioptions/clientset/clientset.go
```diff
@@ -0,0 +1,111 @@
+// clientset package is used to inject clients inside commands
```
dharmit, line 1: This is probably not that important. But when I was browsing through the code on GoLand, it suggested some fixes. Attaching a screenshot. The changes can be found on https://github.com/dharmit/odo/commit/c3e5429304f2e72e0f0e50e8505801a95b3186d7
Screenshot from 2022-02-04 09-29-14

PR #5434, opened 2022-02-04
Refactor odo init to make parts reusable by odo dev/deploy

### LINKED ISSUE #5420: Fetch starter project
/kind feature

Create a package to fetch a starter project, given:
- a devfile content (devfileObj)
- a starter project name
- a destination directory

The code already exists in `odo init` (in `pkg/odo/cli/init` package), but needs to be reusable from other commands

Refactor odo init code to use this package

### LINKED ISSUE #5421: Select and pull a devfile in interactive mode or with parameters
/kind feature

Create a package to select a devfile and pull it, either:
- interactively, or
- from parameters
Given:
- parameters to define devfile (or no parameters to enter interactive mode),
- a destination directory

The code already exists in odo init (in pkg/odo/cli/init package), but needs to be reusable from other commands

Refactor odo init code to use this package

feloy:
**What type of PR is this:**

/kind cleanup

Fixes #5421 
Fixes #5420

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 


dharmit:
@feloy, there's no change in the functionality of `odo init` here, right? It's just that you moved things from CLI layer to business layer, right?

feloy:
> @feloy, there's no change in the functionality of `odo init` here, right? It's just that you moved things from CLI layer to business layer, right?

Yes, from the user point of view it should be the same, except that with this new version, the user can see the devfile downloaded before to select a starter project and the component name. Not sure if it is ok for UI?

```
? Select language: go
? Select project type: Go Runtime
 ✓  Downloading devfile "go" from registry "DefaultDevfileRegistry" [945ms]
? Which starter project do you want to use? go-starter
 ✓  Downloading starter project "go-starter" [403ms]
? Enter component name: my-go-app

Your new component "my-go-app" is ready in the current directory.
To start editing your component, use "odo dev" and open this folder in your favorite IDE.
Changes will be directly reflected on the cluster.
```

dharmit:
> Not sure if it is ok for UX?

IMO it's better since the user gets feedback sooner than they did earlier. I'm wondering why we removed the "GO BACK" from starter project question.

feloy:
> > Not sure if it is ok for UX?
> 
> IMO it's better since the user gets feedback sooner than they did earlier. I'm wondering why we removed the "GO BACK" from starter project question.

Because with thhis 'GO BACK', we came back to the selection of the devfile, which is now already selected and downloaded: it is too late to come back IMO

dharmit (changes requested):
Don't have anything else at the moment. TBH, it took me a good amount of time to figure things out because of https://github.com/redhat-developer/odo/pull/5434#issuecomment-1033924762.

feloy:
> @feloy why are we moving away from deciding the "backend" upfront? While reading the code, I found it easier to follow things when we figured the backend in "Complete" step before this PR. With this change, I feel a bit uncomfortable that we iterate through all backends that implement the interface.
> 
> In this case, it's guaranteed that the two backends are mutually exclusive because one takes no flags and the other doesn't work without a flag. So I think it's OK. But it can quickly become a pain if a third backend were to be introduced which was going to require, say, a couple of particular flags.

Yes, I agree that it is more difficult to find which backend will be called using this method. 

I propose to change to make the decision upfront in the `pkg/init` package instead (but not in the Complete/Validate/Run functions).

feloy:
> > @feloy why are we moving away from deciding the "backend" upfront? While reading the code, I found it easier to follow things when we figured the backend in "Complete" step before this PR. With this change, I feel a bit uncomfortable that we iterate through all backends that implement the interface.
> > In this case, it's guaranteed that the two backends are mutually exclusive because one takes no flags and the other doesn't work without a flag. So I think it's OK. But it can quickly become a pain if a third backend were to be introduced which was going to require, say, a couple of particular flags.
> 
> Yes, I agree that it is more difficult to find which backend will be called using this method.
> 
> I propose to change to make the decision upfront in the `pkg/init` package instead (but not in the Complete/Validate/Run functions).

@dharmit Please see my last commit.

One could think that we could factorize the detection of the backend, because the condition to determine which backend to use is always the same (`len(flags)==0`), but this will change with the use of Alizer and the personalization of the Devfile.

dharmit:
> Please see my last commit.

What's your rationale behind not making `backend` a part of `InitOptions` at CLI layer and checking for `len(flags)` in each of the interface methods? Genuine question, because I remember the time when I implemented an interface to use between ServiceCatalog and OperatorHub for `odo service` command, and there I determined backend in `Complete` method at CLI layer.

I'm not saying what I did was right or what you did here is wrong. I'm simply trying to understand the thought process. One way I see what you did is - it lets the business logic decide the backend instead of CLI layer. And that's a pretty good reason to offload it to business logic.


> One could think that we could factorize the detection of the backend, because the condition to determine which backend to use is always the same (`len(flags)==0`)

I do find it discomforting because now we have moved the check to each method instead of looping the backends. It's still happening at the start of every method. I was suggesting more along the lines of determining it only once, and then reusing it. But again, I'm curious to know the rationale.

> but this will change with the use of Alizer and the personalization of the Devfile

I can imagine Alizer becoming a third scenario/backend when doing `odo init`. Not sure how personalization of Devfile would affect. I think it happens after determining the devfile (which we do using the backends). Anyway, it doesn't matter here.

----
One final thing. Instead of having `len(flags)==0` check in four methods, can you extract it into an internal function and reuse it so that we have to make changes at only one place if something changes w.r.t determining backend?

pkg/init/interface.go
```diff
@@ -1,9 +1,40 @@
...
 
-import "github.com/devfile/library/pkg/devfile/parser"
+import (
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/library/pkg/devfile/parser"
+
+	"github.com/redhat-developer/odo/pkg/init/backend"
+)
 
 type Client interface {
```
valaparthvi, line 17: This `Client` interface it seems implements the `InitBackend`. This seems confusing to me. For `InitClient`, all the methods in common check if the backend is a flags or interactive, and then call it's method. Why can't we assign a backend when initializing `InitClient` and remove these methods from `Client` interface?

Point being, can we make this work without defining the same method in both `Client` and `InitBackend`?

valaparthvi: I see @dharmit had raised the same question. IIUC, Dharmit is suggesting that we assign the backend while initializing the `InitClient`, so in the `NewInitClient` factory function?

feloy:
> > Please see my last commit.
> 
> What's your rationale behind not making `backend` a part of `InitOptions` at CLI layer and checking for `len(flags)` in each of the interface methods? Genuine question, because I remember the time when I implemented an interface to use between ServiceCatalog and OperatorHub for `odo service` command, and there I determined backend in `Complete` method at CLI layer.
> 

Because this routine will be used in odo init, dev and deploy, we want to factorize the logic in a common package (pkg/init).

> I'm not saying what I did was right or what you did here is wrong. I'm simply trying to understand the thought process. One way I see what you did is - it lets the business logic decide the backend instead of CLI layer. And that's a pretty good reason to offload it to business logic.
> 
> > One could think that we could factorize the detection of the backend, because the condition to determine which backend to use is always the same (`len(flags)==0`)
> 
> I do find it discomforting because now we have moved the check to each method instead of looping the backends. It's still happening at the start of every method. I was suggesting more along the lines of determining it only once, and then reusing it. But again, I'm curious to know the rationale.
> 

Beacuse it won't be the same backend for each method with Alizer and personalizing the devfile.

- with Alizer, SelectDevfile will call Alizer when `len(flags)==0 && sourceFilesExist()`, Interactive when `len(flags)==0 && !sourceFilesExist()` and backend if len(flags) != 0`. 
- The method PersonalizeDevfile will call the InteractiveBackend in any case

> > but this will change with the use of Alizer and the personalization of the Devfile
> 
> I can imagine Alizer becoming a third scenario/backend when doing `odo init`. Not sure how personalization of Devfile would affect. I think it happens after determining the devfile (which we do using the backends). Anyway, it doesn't matter here.
> 
> One final thing. Instead of having `len(flags)==0` check in four methods, can you extract it into an internal function and reuse it so that we have to make changes at only one place if something changes w.r.t determining backend?

pkg/init/interface.go
feloy: > This `Client` interface it seems implements the `InitBackend`. This seems confusing to me. For `InitClient`, all the methods in common check if the backend is a flags or interactive, and then call it's method. Why can't we assign a backend when initializing `InitClient` and remove these methods from `Client` interface?
> 

See https://github.com/redhat-developer/odo/pull/5434#issuecomment-1034560129

> Point being, can we make this work without defining the same method in both `Client` and `InitBackend`?

The `Client` and `InitBackend` interfaces are very similar, but this is by "chance", not "by design". If you look at the other implementation (before my last commit), the signatures of the methods were different. Coupling the interface of `Client` and `InitBackend` would limit the possibilities to adapt one without affecting the other.

dharmit:
> Because this routine will be used in odo init, dev and deploy, we want to factorize the logic in a common package (pkg/init).

👍🏾 

> with Alizer, SelectDevfile will call Alizer when `len(flags)==0 && sourceFilesExist()`, Interactive when `len(flags)==0 && !sourceFilesExist()` and backend if len(flags) != 0`.

👍🏾 

> The method PersonalizeDevfile will call the InteractiveBackend in any case

I think PersonalizeDevfile will happen after the backend is done doing its work. Without pulling a devfile from registry (be it stored on disk or be in-memory), how can PersonalizeDevfile know what the user can possibly personalize. Note that my understanding of PersonalizeDevfile is the user being able to modify env vars and/or endpoints.

feloy:
> > The method PersonalizeDevfile will call the InteractiveBackend in any case
> 
> I think PersonalizeDevfile will happen after the backend is done doing its work. Without pulling a devfile from registry (be it stored on disk or be in-memory), how can PersonalizeDevfile know what the user can possibly personalize. Note that my understanding of PersonalizeDevfile is the user being able to modify env vars and/or endpoints.

PersonalizeDevfile will be called by CLI after SelectDevfile, ..., PersonalizeName. So at this point the devfile will be already pulled. What I'm saying is that (in this state of the specs), PersonalizeDevfile will always use the Interactive backend, because the specs only indicates that we want to personalize interactively. But in the future, we could be able to personalize the devfile with another backend (by passing flags to the command, to run an Alizer-like routine that will "guess" the env vas and ports, etc)

pkg/init/interface.go
dharmit: > IIUC, Dharmit is suggesting that we assign the backend while initializing the `InitClient`, so in the `NewInitClient` factory function?

Nope, not in the `NewInitClient` factory function. More like in `Complete` or `Validate` phase of CLI. Factory method is, in my understanding, definitely not the place to do this. But I agree with what @feloy said:

> Because this routine will be used in odo init, dev and deploy, we want to factorize the logic in a common package (pkg/init).

dharmit:
> PersonalizeDevfile will be called by CLI after SelectDevfile, ..., PersonalizeName. So at this point the devfile will be already pulled. What I'm saying is that (in this state of the specs), PersonalizeDevfile will always use the Interactive backend, because the specs only indicates that we want to personalize interactively. But in the future, we could be able to personalize the devfile with another backend (by passing flags to the command, to run an Alizer-like routine that will "guess" the env vas and ports, etc)

☮️ 

I am afraid that I'm digressing into the bits that shouldn't be of much concern simply because I can have an opinion about it. I'll stop that and review the PR.

pkg/init/backend/interactive_test.go
```diff
@@ -0,0 +1,264 @@
...
+func TestInteractiveBackend_SelectDevfile(t *testing.T) {
+	type fields struct {
+		buildAsker         func(ctrl *gomock.Controller) asker.Asker
+		buildCatalogClient func(ctrl *gomock.Controller) catalog.Client
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *DevfileLocation
+		wantErr bool
```
dharmit, line 28: This is not being used/set in any tests. Is that intentional?

dharmit (commented):
Not major changes. Can do away without them as well.

/lgtm

dharmit:
@feloy @kadel are we going to need json output for `odo init`?

pkg/init/backend/interactive_test.go
valaparthvi: I think it's just from the template that your IDE provides for tests.

pkg/init/init.go
```diff
@@ -18,21 +21,59 @@ import (
...
+
+// Validate calls Validate method of the adequate backend
+func (o *InitClient) Validate(flags map[string]string) error {
+	var backend backend.InitBackend
+	if len(flags) == 0 {
+		backend = o.interactiveBackend
+	} else {
+		backend = o.flagsBackend
+	}
+	return backend.Validate(flags)
```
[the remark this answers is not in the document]
feloy: As these conditions will change very soon with Alizer and PersonalizeDevfile, I would prefer to wait and not make premature optimizations

pkg/init/backend/interface.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// SelectDevfile selects a devfile and returns its location information, depending on the flags
+	SelectDevfile(flags map[string]string) (location *DevfileLocation, err error)
+
+	// SelectStarterProject selects a starter project from the devfile and returns information about the starter project,
+	// depending on the flags. If not starter project is selected, a nil starter is returned
+	SelectStarterProject(devfile parser.DevfileObj, flags map[string]string) (starter *v1alpha2.StarterProject, err error)
+
+	// PersonalizeName updates a devfile name, depending on the flags
+	PersonalizeName(devfile parser.DevfileObj, flags map[string]string) error
```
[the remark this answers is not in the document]
feloy: I would prefer to concentrate the knowledge of the `flags` content in only one place, here in the backend (across all the methods of `pkg/init).

valaparthvi:
> @feloy @kadel are we going to need json output for `odo init`?

I think we can consider the JSON output we have for `odo create`.
```sh
➜  odo create nodejs -o json                               
{
        "kind": "Component",
        "metadata": {
                "name": "nodejs-x101-aisy",
                "creationTimestamp": null
        },
        "spec": {
                "app": "app",
                "type": "nodejs",
                "ports": [
                        "5858"
                ]
        },
        "status": {
                "state": "Not Pushed"
        }
}
➜  odo create nodejs -o json
{
        "kind": "Error",
        "metadata": {
                "creationTimestamp": null
        },
        "message": "this directory already contains a component"
}
```

valaparthvi:
```sh
➜   odo init --name my-app --devfile-path https://devfiles.example.com/nodejs/devfile.yaml

 ✗  Downloading devfile from "https://devfiles.example.com/nodejs/devfile.yaml" [253ms]
 ✗  Unable to download devfile: Get "https://devfiles.example.com/nodejs/devfile.yaml": dial tcp: lookup devfiles.example.com: no such host
```
I think we can put up a link here that actually works, instead of a mock, I think it will provide a better user experience. For example - https://github.com/OpenLiberty/devfile-stack/releases/download/maven-0.7.0/devfile.yaml, although to be consistent with other nodejs examples in `odo init --help`, we could use https://raw.githubusercontent.com/redhat-developer/odo/main/tests/examples/source/devfiles/nodejs/devfile-registry.yaml.

feloy:
@valaparthvi I agree concerning the error messages we can do better. Could you create an issue to track the subject, I would prefer not to treat it as part of this PR

valaparthvi:
/approve

/hold 
Until https://github.com/redhat-developer/odo/pull/5434#issuecomment-1034642311 is resolved.

valaparthvi:
/unhold
postponing JSON output until after alpha1.

PR #5469, opened 2022-02-19
Modify periodic test script for v2 branch

### LINKED ISSUE #5411: ODO V2 testing during V3 development
/kind tests
Identify tests that are required to be tested during when supporting V2 during V3 development.  

- [x] Identify tests that can be run periodically (low risk) - V2 Branch
- [x] Identify tests that need to be run on a regular basis - V2 Branch

## Why is this needed?
Avoid bottlenecks do to PRs submitted against the two branches and avoid unnecessary costs.

anandrkskd:

**What type of PR is this:**

/kind tests

This PR will update the script used for periodic testing on openshift-ci.

Fixes #5411 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/openshiftci-periodic-tests.sh
+++ b/scripts/openshiftci-periodic-tests.sh
@@ -24,9 +24,10 @@ check_and_run_devfileTest() {
         Devfile_path=$TEMPDIR/devfile.yaml
         set +e
         # check if devfiles differ then the one in examples dir
-        diff $Devfile_path $Example_devfile_path || NOTEQUAL="true" 
+        diff $Devfile_path $Example_devfile_path
         set -e 
-        if NOTEQUAL="true"; then
+        if [[ $? == 1 ]]; then
+            NOTEQUAL="true"
             echo "Devfile for $LANGUAGE is not equal to the one in examples dir"
             cp $Devfile_path $Example_devfile_path
         fi
--- a/scripts/openshiftci-periodic-tests.sh
+++ b/scripts/openshiftci-periodic-tests.sh
@@ -6,15 +6,15 @@ set -e
 set -x
 
 # check if devfiles are updated and if so, update and test them
-check_devfile() {
-    NOTEQUAL= false
+check_and_run_devfileTest() {
+    NOTEQUAL=false
     # Languages for which devfiles preset in examples dir
-    LANGUAGES=('nodejs' 'python' 'springboot')
+    LANGUAGES=('python' 'nodejs' 'springboot')
 
     for LANGUAGE in "${LANGUAGES[@]}"; do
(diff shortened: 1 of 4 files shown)


anandrkskd:
all so require PR https://github.com/openshift/release/pull/26125 to go in.

scripts/openshiftci-periodic-tests.sh
```diff
@@ -5,6 +5,41 @@ set -e
 # show commands
 set -x
 
+# check if devfiles are updated and if so, update and test them
+check_devfile() {
+    NOTEQUAL= false
```
feloy, line 10: This is not an affectation, but this is running the `false` command by defining the env variable NOTEQUAL being equals to the empty string (like you would run `DEBUG=true mycommand`)

(shell is very sensitive to spaces :( )

feloy (changes requested):

scripts/openshiftci-periodic-tests.sh
anandrkskd: Updated!

scripts/openshiftci-periodic-tests.sh
```diff
@@ -5,13 +5,46 @@ set -e
...
+        TEMPDIR=$(mktemp -d)
+        # download devfiles with odo
+        if [[ $LANGUAGE == "springboot" ]]; then
+            odo create java-$LANGUAGE language --context $TEMPDIR
+        else
+            odo create $LANGUAGE language --context $TEMPDIR
+        fi
+
+        Devfile_path=$TEMPDIR/devfile.yaml
+        set +e
```
feloy, line 25: Do we need this?
`false || NOTEQUAL="true"` is not a failing command

You would need it to do smth similar to:
```
set +e
diff $Devfile_path $Example_devfile_path 
set -e
if [[ $? == 1 ]]; then
 ...
fi
```

but with the `||`, you are "catching" the error and this is the result of `NOTEQUAL="true" which applies, which is always OK

scripts/openshiftci-periodic-tests.sh
```diff
@@ -5,13 +5,46 @@ set -e
...
+        # check if devfiles differ then the one in examples dir
+        diff $Devfile_path $Example_devfile_path || NOTEQUAL="true" 
+        set -e 
+        if NOTEQUAL="true"; then
```
feloy, line 29:
```
if [ "$NOTEQUAL" == "true" ]; then
```

feloy (changes requested):

scripts/openshiftci-periodic-tests.sh
anandrkskd: Updated!

Issue #5484, opened 2022-02-23
odo dev part 1 (basics)


kadel:
## Acceptance Criteria 
- [ ] The result of successful execution of `odo dev` is Devfile component deployed in inner loop mode.
- [ ] The command is "blocking" the terminal. It keeps running until the user hits "ctrl+c".
- [ ] The first line of the output should indicate if odo found devfile yaml in the current directory or not
- [ ] The command should inform the user about how it is progressing (downloading etc...) (see example for details)
- [ ] The last line of the output needs to always inform user that ctrl+c will exit the command

ideally this should be merged by Wednesday 2.3.2022

## blocked issues
- [ ] #5485 
- [ ] #5476 
- [ ] #5477

PR #5493, opened 2022-02-24
Cleanup Kubernetes resources when `odo dev` exits

### LINKED ISSUE #5485: odo dev part 2 (cleanup)
## Acceptance Criteria
- [x] Before the command ends it should delete all **inner loop** resources that it created on the cluster.
   - [x] There are different ways how the command can end (for example SIGQUIT, SIGHUP, SIGINT etc...) Program should catch all those signals and do cleanup before exiting.
   - [x] if the command is interrupted in the middle of some other operation (for example file syncing, k8s resource creation etc...) it should complete what is is doing, and when it gets to the point when `odo dev` is waiting or files to change start the cleanup procedure
- [x] The last line of the output needs to always inform user that ctrl+c will end the command and that it will result in deleting all resources from the cluster
- [ ] at least one test with devfile that creates storage, k8s. networking/Service (endpoint), k8s component.  (moved to separate issue https://github.com/redhat-developer/odo/issues/5621)
- [ ] verify that this work on Windows (moved to separate issue https://github.com/redhat-developer/odo/issues/5621)
- [x] for now this won't wait for resources to get deleted, it can reuse current `odo delete component` logic. It should show warning that it might take some time for resources to get deleted

## blocked on
- https://github.com/redhat-developer/odo/issues/5484

dharmit:
**What type of PR is this:**
/kind feature

Cleans up the Kubernetes Deployment when user hits `Ctrl+C`. Since we have `ownerRerences` set for resources created by odo in the inner loop, deleting a Deployment will delete everything else that belongs to the component.

Fixes #5485

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


dharmit:
Closing this. Will open a separate PR once we have #5484 addressed.

PR #5500, opened 2022-02-28
Delete with devfile.yaml

### LINKED ISSUE #5478: `odo delete component` -  in the directory with `devfile.yaml`
/kind task

## Acceptance Criteria 
- [ ] If `odo delete component` is executed in a directory with `devile.yaml` it should delete all cluster resources related to that devfile (both inner loop and outer loop)
- [ ] Command to delete component should print what will be deleted first and ask the user to confirm.
- [ ] Command should have a flag to "force" deletion without confirmation
- [ ] No local files are deleted or modified.

valaparthvi:

**What type of PR is this:**
/kind feature

This PR adds feature to delete component when devfile.yaml is accessible.

Fixes #5478 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

- [ ] Demo

1. Have a directory with devfile and source files
2. `odo push` or `odo dev`
3. `export PODMAN_CMD=echo` (optional)
4. `odo deploy`
5. `odo delete component`
Sample Output - 
```sh
➜  odo delete component 
The following components will be deleted: 
 •  Devfile Component: nodejs-prj1-api-abhz
 •  Kubernetes Resources:
         Deployment : my-component
         Deployment : my-component-2
? Are you sure you want to delete these resources? Yes
 ✓  Deleting Devfile component [22ms]
 ✓  Deleting Kubernetes resources [49ms]
```
TODO:
- [x] Currently no error is printed if the resource is not found on the cluster, it can be seen when run with debug level set between 1-4.
- [ ] Cleanup v2delete
- [x] Unit tests
- [ ] Documentation


pkg/component/delete/delete.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (o *DeleteComponentClient) UnDeploy(devfileObj parser.DevfileObj, path string) error {
+	undeployHandler := newUndeployHandler(path, o.kubeClient)
+	return libdevfile.Deploy(devfileObj, undeployHandler)
+}
+
+// DevfileComponentDelete deletes the devfile component
+func (do *DeleteComponentClient) DeleteComponent(devfileObj parser.DevfileObj, componentName string) error {
+	labels := componentlabels.GetLabels(componentName, "app", false)
+	return component.Delete(do.kubeClient, devfileObj, componentName, "app", labels, false, false)
```
feloy, line 29: Do you want to move this package-level function `component.Delete` to this `delete` package? It would be the occasion to cleanup the `component` package, and to see what to do with the `spinner`: do we want to rely on spinner into this layer, or only on the CLI one?

feloy (changes requested):

pkg/component/delete/delete.go
valaparthvi: CLI does seem to be the ideal place to put spinner logic because it prints things to the terminal and is user facing, but then it is much more easier to place the spinner at the right place in the business logic.
But then we also discussed that business layer should just return an object and the CLI layer will print that object in JSON or human readable format.
I think we should stick to spinners in the CLI layer (write the methods in business logic in small pieces such that spinners at the CLI layer make sense), and use klog in the business layer for better understandably.

What do you think?

pkg/component/delete/delete.go
```diff
@@ -24,6 +27,24 @@ func NewDeleteComponentClient(kubeClient kclient.ClientInterface) *DeleteCompone
 	}
 }
 
+func (o *DeleteComponentClient) ListKubernetesComponents(devfileObj parser.DevfileObj, path string) (list []unstructured.Unstructured, err error) {
+	components, err := devfileObj.Data.GetComponents(common.DevfileOptions{})
```
feloy, line 31: You can restrict the result with some devfile option: 
```
common.DevfileOptions{
		ComponentOptions: parsercommon.ComponentOptions{ComponentType: devfile.KubernetesComponentType},
}
```

pkg/component/delete/delete.go
```diff
@@ -24,6 +27,24 @@ func NewDeleteComponentClient(kubeClient kclient.ClientInterface) *DeleteCompone
...
+	if err != nil {
+		return
+	}
+	var u unstructured.Unstructured
+	for _, component := range components {
+		if component.Kubernetes != nil {
+			u, err = service.GetK8sComponentAsUnstructured(component.Kubernetes, path, devfilefs.DefaultFs{})
```
feloy, line 38: You could move the `GetK8sComponentAsUnstructured` function to some devfile-related package (libdevfile?) as it is 100% related to devfile

pkg/odo/cli/delete/component.go
```diff
@@ -67,25 +67,44 @@ func (o *ComponentOptions) deleteNamedComponent() error {
...
+	}
+	for _, resource := range resources {
+		fmt.Println("\t", resource.GetKind(), ":", resource.GetName())
+	}
 
-	componentName := o.EnvSpecificInfo.GetName()
+	if o.forceFlag || ui.Proceed("Are you sure you want to delete these resources?") {
+		// delete innerloop resources
+		spinner := log.Spinner("Deleting Devfile component")
+		defer spinner.End(false)
```
feloy, line 88: As there is no return statement before the `spinner.End(true)`, you don't need this `defer`

pkg/odo/cli/delete/component.go
```diff
@@ -67,25 +67,44 @@ func (o *ComponentOptions) deleteNamedComponent() error {
...
+			log.Errorf("error occurred while deleting component, cause: %v", err)
+		}
+		spinner.End(true)
 
-	if o.forceFlag || ui.Proceed(fmt.Sprintf("Are you sure you want to delete the devfile component: %s?", componentName)) {
 		// delete outerloop resources
-		devfileObj := o.EnvSpecificInfo.GetDevfileObj()
-		err := o.clientset.DeleteClient.UnDeploy(devfileObj, filepath.Dir(o.EnvSpecificInfo.GetDevfilePath()))
+		spinner = log.Spinner("Deleting Kubernetes resources")
+		defer spinner.End(false)
```
feloy, line 97: Same here

feloy (changes requested):

pkg/odo/cli/delete/component.go
```diff
@@ -22,15 +37,16 @@ func NewComponentOptions() *ComponentOptions {
 }
 
 func (o *ComponentOptions) SetClientset(clientset *clientset.Clientset) {
+	o.clientset = clientset
 }
 
 func (o *ComponentOptions) Complete(cmdline cmdline.Cmdline, args []string) (err error) {
-	return nil
+	o.Context, err = genericclioptions.New(genericclioptions.NewCreateParameters(cmdline).NeedDevfile(""))
```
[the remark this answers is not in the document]
feloy: sorry, name == ""

pkg/component/delete/delete.go
```diff
@@ -97,10 +73,10 @@ func (o DeleteComponentClient) getPod(componentName, appName string) (pod *corev
 		klog.V(1).Info("Component not found on the cluster.")
 		if kerrors.IsForbidden(err) {
 			klog.V(3).Infof("Resource for %s forbidden", componentName)
-			return pod, nil
+			return pod, err
```
feloy, line 76: Thsi changes the behaviour described in the comment line 91/67. Are you sure you want to change this behaviour?

pkg/component/delete/undeploy_handler.go
```diff
@@ -40,14 +39,7 @@ func (o *undeployHandler) ApplyKubernetes(kubernetes v1alpha2.Component) error {
...
 	// Un-deploy the K8s manifest
-	err = o.kubeClient.DeleteDynamicResource(u.GetName(), gvr.Resource.Group, gvr.Resource.Version, gvr.Resource.Resource)
-	if err != nil {
-		if strings.Contains(err.Error(), "not found") {
-			klog.V(3).Infof("Resource for %s not found; cause: %w", u.GetName(), err)
-			return nil
-		}
-	}
-	return err
+	return o.kubeClient.DeleteDynamicResource(u.GetName(), gvr.Resource.Group, gvr.Resource.Version, gvr.Resource.Resource)
```
feloy, line 42: This is dangerous to change this behaviour. Because of `ownerReferences` and the garbage collection triggered by the deletion of a previous resource, this resource could have been deleted in the meantime, and we don't want to fail the command at this point in this case. Or the user could have deleted this resource by hand, and wants to delete the other ones with `odo delete`, etc

valaparthvi: Are you suggesting that in the case that the resource is not found for whatsoever reason, and this method returns a *NotFoundError, we return a `nil`?

feloy: Yes, because we don't want to fail the complete command if the deletion of a single resource fails

pkg/component/delete/delete.go
valaparthvi: Can't believe I missed that. I was focused on removing log.Status out of this pkg, but also wanted to ensure the correct error messages are printed, I forgot about the effect of this one. Thanks for catching this. :100:

valaparthvi:
/retest
Failure not related to tests.

pkg/component/delete/delete_test.go
```diff
@@ -0,0 +1,461 @@
...
+				},
+			},
+			args: args{
+				devfileObj:    odoTestingUtil.GetTestDevfileObjWithPreStopEvents(fs, "runtime", "echo \"Hello World!\""),
+				componentName: componentName,
+			},
+			wantErr: true,
+		},
+		{
+			name: "failed to delete component where a devfile that has PreStopEvents because it failed to execute the command inside the container",
```
feloy, line 150: @kadel do we want this behaviour, or do we want to delete the component regardless of the reult of the preStopEvents commands?

pkg/component/delete/delete_test.go
```diff
@@ -0,0 +1,461 @@
+package delete
```
feloy, line 1: Great coverage, thanks

pkg/odo/cli/delete/component.go
```diff
@@ -22,15 +40,20 @@ func NewComponentOptions() *ComponentOptions {
...
+	o.clientset = clientset
 }
 
 func (o *ComponentOptions) Complete(cmdline cmdline.Cmdline, args []string) (err error) {
-	return nil
+	createParameters := genericclioptions.NewCreateParameters(cmdline)
+	if o.name == "" {
+		createParameters = createParameters.NeedDevfile("")
+	}
+	o.Context, err = genericclioptions.New(createParameters)
```
feloy, line 51: Could you create the context only when name is ""?
AS a metter of fact, it is not possible to use `genericclioptions.New` outside an odo directory, as it will fail due to the missing env file, and the context is not necessary anymore in this case ([see https://github.com/redhat-developer/odo/pull/5511/files](https://github.com/redhat-developer/odo/pull/5511/files#diff-1a997d7c41a88e0f0001418a841d488fe47beffdbfc7e2616dc52925568e5913R42).

tests/integration/devfile/cmd_delete_test.go
```diff
@@ -0,0 +1,65 @@
...
+		})
+		When("the component is deployed in DEV mode", func() {
+			BeforeEach(func() {
+				session := helper.CmdRunner("odo", "dev")
+				defer session.Kill()
+				helper.WaitForOutputToContain("Waiting for something to change", 180, 10, session)
+				list := commonVar.CliRunner.Run("get", "deployment", "-n", commonVar.Project).Out.Contents()
+				Expect(list).To(ContainSubstring("mynodejs"))
+			})
+			It("should not fail when odo delete component is run", func() {
```
feloy, line 42: could you indicate that the deployment (or component?) will be deleted?

tests/integration/devfile/cmd_delete_test.go
```diff
@@ -0,0 +1,65 @@
...
+		})
+		When("the component is deployed in DEPLOY mode", func() {
+			BeforeEach(func() {
+				helper.Cmd("odo", "deploy").AddEnv("PODMAN_CMD=echo").ShouldPass()
+				deployList := commonVar.CliRunner.Run("get", "deployment", "-n", commonVar.Project).Out.Contents()
+				Expect(deployList).To(ContainSubstring("my-component"))
+				serviceList := commonVar.CliRunner.Run("get", "svc", "-n", commonVar.Project).Out.Contents()
+				Expect(serviceList).To(ContainSubstring("my-cs"))
+			})
+			It("should not fail when odo delete component is run", func() {
```
feloy, line 56: could you indicate the component is deleted?

feloy (changes requested):

pkg/odo/cli/delete/component.go
```diff
@@ -22,15 +40,20 @@ func NewComponentOptions() *ComponentOptions {
...
 
 func (o *ComponentOptions) SetClientset(clientset *clientset.Clientset) {
+	o.clientset = clientset
 }
 
 func (o *ComponentOptions) Complete(cmdline cmdline.Cmdline, args []string) (err error) {
+	if o.name == "" {
+		createParameters := genericclioptions.NewCreateParameters(cmdline).NeedDevfile("")
+		o.Context, err = genericclioptions.New(createParameters)
+		return
```
feloy, line 50: Even if the language supports this, it is a better practice to explicitely return the value, to make it easier to read:
```
return err
```

pkg/odo/cli/delete/component.go
```diff
@@ -91,11 +99,85 @@ func (o *ComponentOptions) deleteNamedComponent() error {
...
+	}
+	return
+}
+
+// deleteDevfileComponent deletes all the components defined by the devfile in the current directory
 func (o *ComponentOptions) deleteDevfileComponent() error {
+	devfileObj := o.EnvSpecificInfo.GetDevfileObj()
+
+	// Print all the resources that will be deleted.
+	performUndeploy, err := o.printDevfileComponents(devfileObj)
```
feloy, line 126: Having this function `printDevfileComponents` to also compute `undeploy` is confusing, as these 2 things are non related from the caller point of view.

I would instead have a separate function to check if there are Kubernetes components in a devfile, that would be more eloquent

feloy (commented):
Just a little remark, and a latest test is missing because of the capitalization of an error message:
```
✗  The current directory does not represent an odo component.
          To start editing your component, use "odo dev" and open this folder in your favorite IDE. Changes will be directly reflected on the cluster.
          Or switch to directory with a component.
          
      to contain substring
          <string>: the current directory does not represent an odo component
```

pkg/odo/cli/delete/component.go
valaparthvi: Yeah, that makes more sense. On it.

pkg/component/delete/delete_test.go
```diff
@@ -0,0 +1,461 @@
...
+				},
+			},
+			args: args{
+				devfileObj:    odoTestingUtil.GetTestDevfileObjWithPreStopEvents(fs, "runtime", "echo \"Hello World!\""),
+				componentName: componentName,
+			},
+			wantErr: false,
+		},
+		{
+			name: "failed to delete component where a devfile that has PreStopEvents because pod is not in the running state",
```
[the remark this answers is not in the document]
kadel: Delete should happen even if the event fails. Otherwise, it would be easy to get into a situation where it is impossible to delete the component due to a bug/type/mistake in PreStop event. 

If this turns out to be a problem, we can consider changing this in the future, to not delete if it fails, and add flag `--ignore-even-failures`.
To keep things simple for now, lets just ignore prestop events failures by default .

pkg/component/delete/delete_test.go
kadel: we definitely want to delete the component even if pre stop even fails. more info in https://github.com/redhat-developer/odo/pull/5500#discussion_r824572155

kadel:
I would update output a little bit, in its current form
```
The following resources will be deleted: 
 •  Devfile Component: nodejs-prj1-api-abhz
 •  Kubernetes Resources:
         Deployment : my-component
         Deployment : my-component-2
? Are you sure you want to delete these resources? (y/N)
```
it looks like The devfile component and Kubernetes resources are two separate things. But in fact, the Devfile component is represented by those resources.


Something like this might be better:
```
This will delete nodejs-prj1-api-abhz from "namespace".
The component contains the following resources that will get deleted:
 - Deployment : my-component
 - Deployment : my-component-2

? Do you want to delete "component-from-devfile" and all its resources? (y/N):
```

valaparthvi:
The output for delete with devfile is a little different than without devfile. Without devfile displays deployments matching the name and its associated endpoints that will be deleted, while this one doesn't show endpoints, is that alright @kadel ?

kadel:
> The output for delete with devfile is a little different than without devfile. Without devfile displays deployments matching the name and its associated endpoints that will be deleted, while this one doesn't show endpoints, is that alright @kadel ?

I haven't checked that to be honest. We need to make sure that the outputs are the same.

kadel:
> > The output for delete with devfile is a little different than without devfile. Without devfile displays deployments matching the name and its associated endpoints that will be deleted, while this one doesn't show endpoints, is that alright @kadel ?
> 
> I haven't checked that to be honest. We need to make sure that the outputs are the same.

please try to make the output look like what I suggested. We will have to update the output for deletion scenario.

valaparthvi:
>please try to make the output look like what I suggested. We will have to update the output for deletion scenario.

Yup, I'm already on it.

pkg/odo/cli/delete/component.go
```diff
@@ -91,11 +99,79 @@ func (o *ComponentOptions) deleteNamedComponent() error {
 	return nil
 }
 
-// deleteDevfileComponent deletes a component defined by the devfile in the current directory
+// deleteDevfileComponent deletes all the components defined by the devfile in the current directory
 func (o *ComponentOptions) deleteDevfileComponent() error {
+	devfileObj := o.EnvSpecificInfo.GetDevfileObj()
+	k8sComponents, err := libdevfile.ListKubernetesComponents(devfileObj, filepath.Dir(o.EnvSpecificInfo.GetDevfilePath()))
```
kadel, line 105: Why are you not resuing `ListResourcesToDelete` (https://github.com/redhat-developer/odo/blob/daa9c61a4108dbef439842a17f993ebda74a8fcd/pkg/component/delete/delete.go#L26)?

kadel (changes requested):
I think that this needs to be refactored to fit it with https://github.com/redhat-developer/odo/pull/5511 There is a lot of overlapping logic, that is duplicity implemented. For example all output and user confirmation should be shared between those two scenarios.

pkg/odo/cli/delete/component.go
feloy: I'm not sure this is what we want to do here.

- With `ListResourcesToDelete`, we will find in the cluster all the resources belonging to a component, based on their labels.
- When deleting using a devfile, I understood that we want to delete the resources found in the devfile.

kadel: @feloy that is a good point. I forgot about that. It makes sense to have two separate logics to handle this

valaparthvi:
@kadel I've tried to ensure similar outputs for both delete operations, but there is some UI improvements that can be done, e.g. adding spinners here and there. I intend to leave that to @cdrage, because he's good with that. For now, everything seems to be functional.

I haven't reused the functions from #5511 because I would like to treat both the delete operations as separate, and also as Philippe mentioned in his https://github.com/redhat-developer/odo/pull/5500#discussion_r824778966.

kadel:
I've a simple devfile with outer loop and inner loop config (https://github.com/kadel/devfile-nodejs-deploy)

I do `odo dev` than `odo deploy` and then `odo delete component`
The output shows only one Deployment and one Service.
```
$ odo delete component
This will delete "devfile-nodejs-deploy" from the namespace "myproj".
	- Deployment : devfile-nodejs-deploy
	- Service : devfile-nodejs-deploy
? Are you sure you want to delete "devfile-nodejs-deploy" and all its resources? Yes
```

But there are two of each.
```
$ k get deployment,svc
NAME                                        READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/devfile-nodejs-deploy       1/1     1            1           42s
deployment.apps/devfile-nodejs-deploy-app   1/1     1            1           70s

NAME                                TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)          AGE
service/devfile-nodejs-deploy       LoadBalancer   10.217.5.93    <pending>     3000:32466/TCP   41s
service/devfile-nodejs-deploy-app   ClusterIP      10.217.5.180   <none>        3000/TCP         70s
```

I would expect all 4 resources listed in the output of what will be deleted.

kadel:
I also just noticed that the name of the Deployment that is being shown is not correct. I haven't checked the code yet, but it looks like it is showing the name of the component instead of the name of the Deployment.

valaparthvi:
> I've a simple devfile with outer loop and inner loop config (https://github.com/kadel/devfile-nodejs-deploy)
> 
> I do `odo dev` than `odo deploy` and then `odo delete component` The output shows only one Deployment and one Service.
> 
> ```
> $ odo delete component
> This will delete "devfile-nodejs-deploy" from the namespace "myproj".
>  •  The component contains the following resources that will get deleted:
> 	- Deployment : devfile-nodejs-deploy
> 	- Service : devfile-nodejs-deploy
> ? Are you sure you want to delete "devfile-nodejs-deploy" and all its resources? Yes
> ```
> 
> But there are two of each.
> 
> ```
> $ k get deployment,svc
> NAME                                        READY   UP-TO-DATE   AVAILABLE   AGE
> deployment.apps/devfile-nodejs-deploy       1/1     1            1           42s
> deployment.apps/devfile-nodejs-deploy-app   1/1     1            1           70s
> 
> NAME                                TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)          AGE
> service/devfile-nodejs-deploy       LoadBalancer   10.217.5.93    <pending>     3000:32466/TCP   41s
> service/devfile-nodejs-deploy-app   ClusterIP      10.217.5.180   <none>        3000/TCP         70s
> ```
> 
> I would expect all 4 resources listed in the output of what will be deleted. Or at least both Deployments to show. `devfile-nodejs-deploy-app` might be tricker to show.

The resources `deployment.apps/devfile-nodejs-deploy-app` and `service/devfile-nodejs-deploy-app` belong to the innerloop I assume, that were created when you ran `odo dev` for "devfile-nodejs-deploy` component.

The above output is expected, since for `odo delete` with access to devfile.yaml, we only examine the devfile, we do not touch the cluster until the actual deletion operation.

>I also just noticed that the name of the Deployment that is being shown is not correct. I haven't checked the code yet, but it looks like it is showing the name of the component instead of the name of the Deployment.

Correct, it shows the devfile component name.

kadel:
I've discussed https://github.com/redhat-developer/odo/pull/5500#issuecomment-1066626285 with @valaparthvi The agreement is that it will be updated so it shows the list of the resources that odo is going to delete (what Kubernetes resources odo is going to api `Delete` action)

For https://github.com/kadel/devfile-nodejs-deploy it should end up looking like this

```
$ odo delete component
This will delete "devfile-nodejs-deploy" from the namespace "myproj".
	- Deployment : devfile-nodejs-deploy
	- Deployment : devfile-nodejs-deploy-app
	- Service : devfile-nodejs-deploy
? Are you sure you want to delete "devfile-nodejs-deploy" and all its resources? Yes
```

tests/integration/devfile/cmd_delete_test.go
```diff
@@ -9,15 +10,19 @@ import (
 	"github.com/redhat-developer/odo/tests/helper"
 )
 
-var _ = Describe("odo delete command tests", func() {
-	var cmpName string
+var _ = FDescribe("odo delete command tests", func() {
```
anandrkskd, line 13: Shouldn't this be `Describe` instead?

valaparthvi: Yup, that was left by mistake.

valaparthvi:
@feloy @kadel 
This is up for review again. I haven't fixed unit tests yet because I feel as if I've over worked them with this PR, I'll fix them once you both approve the changes.

pkg/component/delete/delete.go
```diff
@@ -74,3 +85,85 @@ func references(list []unstructured.Unstructured, ownerRef metav1.OwnerReference
 	}
 	return false
 }
+func (do DeleteComponentClient) ListResourcesToDeleteFromDevfile(devfileObj parser.DevfileObj, appName string) (map[string]string, []unstructured.Unstructured, error) {
+	var resources []unstructured.Unstructured
+	var isInnerLoopLabels map[string]string
+	componentName := devfileObj.GetMetadataName()
+
+	deployment, err := do.kubeClient.GetDeploymentByName(fmt.Sprintf("%s-%s", componentName, appName))
```
feloy, line 100: It would be important to use a centralized function wrapping the `fmt.Sprintf("%s-%s", componentName, appName)`, that is also used during deployment creation, so we can change the name of the deployment easily.

As far as I can see, it is `pkg.util.NamespaceKubernetesObject` that is used during deployment creation. Please use this function, and feel free to change its name to make it more expressive. You could also place it in the `pkg/component` package

pkg/component/delete/delete.go
```diff
@@ -74,3 +85,95 @@ func references(list []unstructured.Unstructured, ownerRef metav1.OwnerReference
...
+	var resources []unstructured.Unstructured
+	// innerLoopLabels will be passed on to DeleteComponentClient.DeleteDevfileComponent() to fetch the deployment for deletion;
+	// an empty list is returned if the deployment is not found on the cluster
+	var innerLoopLabels map[string]string
+
+	// Fetch the deployment of the devfile component
+	componentName := devfileObj.GetMetadataName()
+	deployment, err := do.kubeClient.GetDeploymentByName(fmt.Sprintf("%s-%s", componentName, appName))
+	if err == nil {
+		innerLoopLabels = deployment.Labels
```
feloy, line 102: Instead of returning labels to be used to delete the deployment by labels, you should return the deployment as part of the resources list, so it is identified by its kind/name, and not by its labels anymore.

I think it is dangerous to identify the deployment to delete by its labels, as you are not 100% sure there won't have other deployment with the same labels.

Going this way, you would have to change `DeleteDevfileComponent` to be some kind of `PreDeleteDevfileComponent` which will run the hooks before to delete the component, but won't delete the deployment itself (as it will be deleted as part of the call to `DeleteResources`)

feloy (changes requested):

pkg/component/delete/delete.go
valaparthvi: >Instead of returning labels to be used to delete the deployment by labels, you should return the deployment as part of the resources list, so it is identified by its kind/name, and not by its labels anymore.

I initially had that thought, but I thought we didn't want to return a low layer k8s object to the CLI layer. Also, I didn't see a way to delete the deployment by using the deployment object itself.

>I think it is dangerous to identify the deployment to delete by its labels, as you are not 100% sure there won't have other deployment with the same labels.

I don't disagree to this, but I thought you and Tomas were specific about deleting the Deployment, which I assume to mean using the `kubeclient.Delete` function that I am currently using. So, I went ahead with fetching resource from the cluster and using it's label.

>Going this way, you would have to change DeleteDevfileComponent to be some kind of PreDeleteDevfileComponent which will run the hooks before to delete the component, but won't delete the deployment itself (as it will be deleted as part of the call to DeleteResources)

To use Delete Resources, I'd have to 1) fetch the deployment by name - this returns an `appsv1.Deployment` object,2)Convert the Deployment object into unstructured.Unstructured object with marshaling and un-marshaling - this seemed like an unnecessary task which I can't remember why.
But, if you are okay with using `DeleteResources` instead of `kubeclient.Delete(labels)`, I can make that work.

pkg/component/delete/delete.go
valaparthvi: I saw `pkg.util.NamespaceKubernetesObject` yday, but it seemed to be doing more than just appending the two names, so I decided to not use it, but sure I can use it.

pkg/component/delete/delete.go
```diff
@@ -126,23 +151,19 @@ func (do *DeleteComponentClient) DeleteDevfileComponent(devfileObj parser.Devfil
...
-		} else {
-			klog.V(4).Infof("Executing %q event commands for component %q", libdevfile.PreStop, componentName)
-			// ignore the failures if any; delete should not fail because preStop events failed to execute
-			err = libdevfile.ExecPreStopEvents(devfileObj, componentName, component.NewExecHandler(do.kubeClient, pod.Name, false))
-			if err != nil {
-				klog.V(4).Infof("Failed to execute %q event commands for component %q, cause: %v", libdevfile.PreStop, componentName, err.Error())
-			}
+	// do not fail Delete operation if if the pod is not running or if the event execution fails
+	if pod.Status.Phase != corev1.PodRunning {
+		klog.V(4).Infof("unable to execute preStop events, pod for component %q is not running", componentName)
```
feloy, line 156: To avoid nested blocks, you can return at this point and remove the following `else`

pkg/odo/cli/delete/component.go
```diff
@@ -105,30 +105,28 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
 	appName := "app"
 
 	log.Info("Searching resources to delete, please wait...")
-	innerLoopLabels, resources, err := o.clientset.DeleteClient.ListResourcesToDeleteFromDevfile(devfileObj, appName)
+	isInnerLoop, resources, err := o.clientset.DeleteClient.ListResourcesToDeleteFromDevfile(devfileObj, appName)
```
feloy, line 108: should be `isInnerLoopDeployed`

feloy (commented):
I find the flow much clearer with these changes. Thanks
Just some more nit-picking ;)

kadel:
/approve
the code needs some cleanup and tests. But functionality and output looks good to me

pkg/component/delete/delete.go
```diff
@@ -74,3 +86,105 @@ func references(list []unstructured.Unstructured, ownerRef metav1.OwnerReference
...
+	// else if it is not found, then continue to outer loop
+	if err == nil {
+		isInnerLoopDeployed = true
+		var unstructuredDeploy unstructured.Unstructured
+		unstructuredDeploy, err = kclient.ConvertK8sResourceToUnstructured(deployment)
+		if err != nil {
+			return isInnerLoopDeployed, nil, fmt.Errorf("Failed to parse the resource %q: %q; cause: %w", kclient.DeploymentKind, deploymentName, err)
+		}
+		resources = append(resources, unstructuredDeploy)
+	}
```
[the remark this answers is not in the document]
feloy: You could write a sub-function for this block, to be able to return ASAP in particular cases, for example the case where `IsNotFound(err) is true`.

The `err == nil` is difficult to interpret here, it should be better to have `deployment != nil` instead.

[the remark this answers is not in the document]
valaparthvi: I cannot check `deployment != nil` because it's never the case, I have settled for `deployment.Name != ""` instead.
I tried moving the block to a sub-function, but that made things worse with `isInnerLoopDeployed` involved, so I've kept things as they are.

PR #5510, opened 2022-03-02
GetAllResourcesFromSelector to retrieve resources of all kinds 

feloy:
with specific labels (to be used by "odo list" and "odo delete component")

**What type of PR is this:**

/kind feature

This PR adds a GetAllResourcesFromSelector, function thats searches all recources of any kind in the cluster

This function is intended to be used at least in `odo list` and `odo delete component`

The code comes 99% from https://github.com/ahmetb/kubectl-tree

### DIFF
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -1,0 +1,182 @@
+package kclient
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog"
+)
+
+// Code into this file is heavily inspired from https://github.com/ahmetb/kubectl-tree
+
+func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unstructured.Unstructured, error) {
+	apis, err := findAPIs(c.discoveryClient)
+	if err != nil {
+		return nil, err
+	}
+	return getAllResources(c.DynamicClient, apis.list, ns, selector)
+}
+
+func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	var out []unstructured.Unstructured
+
+	start := time.Now()
+	klog.V(2).Infof("starting to query %d APIs in concurrently", len(apis))
+
+	var errResult error
+	for _, api := range apis {
+		if !api.r.Namespaced {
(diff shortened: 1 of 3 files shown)


pkg/kclient/all.go
```diff
@@ -0,0 +1,182 @@
...
+type resourceMap struct {
+	list []apiResource
+	m    resourceNameLookup
+}
+
+func findAPIs(client discovery.DiscoveryInterface) (*resourceMap, error) {
+	start := time.Now()
+	resList, err := client.ServerPreferredResources()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch api groups from kubernetes: %w", err)
```
dharmit, line 116: IMO, we should return a zero value instead of `nil` when returning a pointer because the program would panic if the caller were to make use of it. I know it's not the case here, and that most of the time if `err != nil`, we don't care about what else was returned. I'm saying it so that we don't get into a problem if the returned value is indeed used.

Correct me if I am wrong.

feloy: IMO, absolutely not. The accepted rule is that if `err` is not nil, the other values are not usable. 
If the caller tries to use this value, it is a programming error and a potential bug/security hole, and should by caught as soon as possible. The panic is a very good way to catch it.

pkg/kclient/all.go
```diff
@@ -0,0 +1,182 @@
...
+		klog.V(5).Infof("iterating over group %s/%s (%d apis)", group.GroupVersion, group.APIVersion, len(group.APIResources))
+		gv, err := schema.ParseGroupVersion(group.GroupVersion)
+		if err != nil {
+			return nil, fmt.Errorf("%q cannot be parsed into groupversion: %w", group.GroupVersion, err)
+		}
+
+		for _, apiRes := range group.APIResources {
+			klog.V(5).Infof("  api=%s namespaced=%v", apiRes.Name, apiRes.Namespaced)
+			if !contains(apiRes.Verbs, "list") {
+				klog.V(4).Infof("    api (%s) doesn't have required verb, skipping: %v", apiRes.Name, apiRes.Verbs)
```
dharmit, line 134: Being pedantic/nitpicky here. Apologies in advance.

> `doesn't have required verb, skipping`

If something is "required" but it doesn't exist, we should fail. In this case, "list" doesn't fall into the definition of required, IMO.

pkg/kclient/all.go
dharmit: Hmm. Makes sense. 👍🏾

pkg/kclient/all.go
```diff
@@ -0,0 +1,182 @@
...
+	var errResult error
+	for _, api := range apis {
+		if !api.r.Namespaced {
+			klog.V(4).Infof("[query api] api (%s) is non-namespaced, skipping", api.r.Name)
+			continue
+		}
+		wg.Add(1)
+		go func(a apiResource) {
+			defer wg.Done()
+			klog.V(4).Infof("[query api] start: %s", a.GroupVersionResource())
```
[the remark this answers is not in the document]
feloy: I'm not the author of these lines, but I'm pretty sure that the author wanted, in `apiNames`, to not modify the `a` structure, and particularly the `a.SingularName` field (the immutability of variables should be considered as much as possible, even if not enforced by the language).

PR #5466, opened 2022-02-17
POC for odo interactive testing

anandrkskd:
Signed-off-by: anandrkskd <anandrkskd@gmail.com>

**What type of PR is this:**

/kind tests

**What does this PR do / why we need it:**
This PR contains a POC to write integration test for interactive mode of odo.

**Which issue(s) this PR fixes:**
Fixes #5320 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

**How to test changes / Special notes to the reviewer:**
ginkgo ./tests/integration/interactive/

PR #5514, opened 2022-03-03
Add welcoming messages to `odo init`

### LINKED ISSUE #5494: `odo init` - welcoming messages
/kind task

I had issues with my connection and when I run `odo init` it took a few seconds before it showed any output.
It looks like it is trying to do `Alizer` analysis as a first thing. 

There should be a message displayed first. So user is aware that odo is doing something

`odo init` can run in two modes. In an empty directory. Or with directory with existing code.

In empty directory the welcoming message should be something like:

```
The current directory is empty.
Odo will help with you starting a new project.
```

With directory with source code:
```
The current directory already contains source code.
Odo will try to autodetect language and project type to select best suited Devfile for your project.
```

## links
one of the acceptance criteria in 
- #5470 
- #5408

rm3l:
**What type of PR is this:**
/kind feature

This adds welcoming messages to the interactive mode of `init` CLI subcommand, so users do not wait few seconds before being showed any output.

Fixes #5494

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

This builds upon #5466 for testing interactive commands. You can run the dedicated integration tests with:
```
ginkgo -v -focus 'Displaying welcoming messages' tests/integration/interactive
```


tests/helper/helper_interactive.go
```diff
@@ -37,6 +37,9 @@ func RunInteractive(command []string, test func(*expect.Console, *bytes.Buffer))
 	cmd.Stdin = c.Tty()
 	cmd.Stdout = c.Tty()
 	cmd.Stderr = c.Tty()
+	if env != nil {
+		cmd.Env = env
```
feloy, line 41: Are you sure the Env is empty? I would use `cmd.Env = append(cmd.Env, env)` instead

tests/integration/interactive/cmd_init_test.go
```diff
@@ -59,4 +61,126 @@ var _ = Describe("odo init interactive command tests", func() {
 		Expect(output).To(ContainSubstring("Your new component \"my-go-app\" is ready in the current directory."))
 		Expect(helper.ListFilesInDir(commonVar.Context)).To(ContainElements("devfile.yaml"))
 	})
+
+	Describe("Displaying welcoming messages", func() {
+
+		type tester func(*expect.Console, *bytes.Buffer)
```
feloy, line 67: Let's describe this type in `helper_interactive.go` and use it in the definition of `func RunInteractive`, so it can be reused

tests/integration/interactive/cmd_init_test.go
```diff
@@ -59,4 +61,126 @@ var _ = Describe("odo init interactive command tests", func() {
...
+
+		testFunc := func(language string, welcomingMsgs []string, additionalTester tester) tester {
```
feloy, line 69: Nice. Can you document what's the behaviour of this function so it could be reused or we can use it at a template for other tests?
Something similar to:
```
// this function returns a tester (usable for `RunInteractive`), that first expects welcomingMsgs, then run additionalTesters and finally expects the asking of a component name
```

tests/integration/interactive/cmd_init_test.go
```diff
@@ -59,4 +61,126 @@ var _ = Describe("odo init interactive command tests", func() {
...
+			It("should display appropriate welcoming messages", func() {
+				command := []string{"odo", "init"}
+				language := "java"
+				welcomingMsgs := []string{"The current directory is empty.", "Odo will help with you starting a new project."}
+				output, err := helper.RunInteractive(command,
+					// Setting verbosity level to 0, because we would be asserting the welcoming message is the first
+					// message displayed to the end user. So we do not want any potential debug lines to be printed first.
+					// Using envvars here (and not via the -v flag), because of https://github.com/redhat-developer/odo/issues/5513
+					[]string{"ODO_LOG_LEVEL=0"},
+					testFunc(language, welcomingMsgs, func(c *expect.Console, output *bytes.Buffer) {
```
feloy, line 113: As this block of test is already used in the previous tests, it would be nice to define this function at a higher level so it can be reused in different tests

feloy (changes requested):

feloy:
/test v4.9-integration-e2e

Infra error

tests/helper/helper_interactive.go
rm3l: Yes, I checked, and `Env` is empty by default.

Actually, per [the doc](https://pkg.go.dev/os/exec#Cmd):

> If Env is nil, the new process uses the current process's environment.

I assumed that, if a non-`nil` `env` is passed, it means that callers may want to have full control over the sub-process environment, and not inherit the current process one.
Unless we want to also allow to include the current process environment; in this case, we would need to append the slice returned by `os.Environ()`, I guess..
What do you think?

feloy: Yes, you are tight, I remember I used os.Environ() here (https://github.com/redhat-developer/odo/blob/main/pkg/devfile/image/docker_compatible.go#L36) for this purpose. 
If tests are passing without getting the original environment, let's keep it as is (`cmd.Env = env`).
Thanks

tests/integration/interactive/cmd_init_test.go
rm3l: Done. Please take another look.

rm3l:
> @rm3l can you take a look why test are facing due to timeout?

@anandrkskd Sure. Looking into it..
To help understand and debug the issue in CI, it would be helpful to make the `helper.RunInteractive` function log the output of the command passed in case errors happen. I will try to make those changes and see how it goes.

pkg/odo/cli/init/init.go
```diff
@@ -125,6 +125,16 @@ func (o *InitOptions) Run() (err error) {
...
 	}()
 
+	isEmptyDir, err := location.DirIsEmpty(o.clientset.FS, o.contextDir)
+	if err != nil {
+		return err
+	}
+	if isEmptyDir && len(o.flags) == 0 {
+		fmt.Println("The current directory is empty.\nOdo will help with you starting a new project.")
+	} else if len(o.flags) == 0 {
+		fmt.Println("The current directory already contains source code.\nOdo will try to autodetect language and project type to select best suited Devfile for your project.")
```
cdrage, line 135: odo should always be lower case :) never capitalized

cdrage (changes requested):
Awesome PR! I've proposed some changes.

rm3l:
> > @rm3l can you take a look why test are facing due to timeout?
> 
> @anandrkskd Sure. Looking into it.. To help understand and debug the issue in CI, it would be helpful to make the `helper.RunInteractive` function log the output of the command passed in case errors happen. I will try to make those changes and see how it goes.

@anandrkskd and @feloy: [9bd8c0c](https://github.com/redhat-developer/odo/pull/5514/commits/9bd8c0c008e6542bad0953a1391971b96fe638b3) makes it possible to log the command output read so far, in case of errors while calling `helper.ExpectString` or `helper.SendLine`. Do let me know what you think.
This commit allowed to narrow down the timeout issue to the interactive command testing environment being different.

In fact, I had a telemetry prompt being displayed first when calling `odo init` in a sub-process:
```
running interactive command [odo init] with env [ODO_LOG_LEVEL=0]
? Help odo improve by allowing it to collect usage data. Read about our privacy statement: https://developers.redhat.com/article/tool-data-collection. You can change your preference later by changing the ConsentTelemetry preference. (Y/n) [?25l78[?25h
```

And the tests were not passing because they expect the new welcoming messages to be displayed first.

This has been fixed with [ae75270](https://github.com/redhat-developer/odo/pull/5514/commits/ae75270bedec78d42fc6a2f20248d456760bcf03), which initializes the sub-process environment with the environment of the current process. This should hopefully fix the timeout issue in CI :)

feloy:
/lgtm
Great wotk, thanks @rm3l

pkg/odo/cli/init/init.go
```diff
@@ -125,6 +125,16 @@ func (o *InitOptions) Run() (err error) {
 		}
 	}()
 
+	isEmptyDir, err := location.DirIsEmpty(o.clientset.FS, o.contextDir)
+	if err != nil {
+		return err
+	}
+	if isEmptyDir && len(o.flags) == 0 {
+		fmt.Println("The current directory is empty.\nOdo will help with you starting a new project.")
```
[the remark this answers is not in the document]
rm3l: Thanks for your review, @cdrage.

> If the bolded output doesn't look good, just use fmt.Println.

Indeed, I used this function in the first place, but found the bold output not to be that readable when followed by other bold interactive questions ^^

> `odo` should also never be capitalized

Per the [issue acceptance criteria](https://github.com/redhat-developer/odo/issues/5494), I read "Odo", but you are certainly right.
```
The current directory is empty.
Odo will help with you starting a new project.
```

@kadel Any thoughts? Capitalized or not?

pkg/odo/cli/init/init.go
rm3l: Comment below : https://github.com/redhat-developer/odo/pull/5514#discussion_r820491142

pkg/odo/cli/init/init.go
```diff
@@ -125,6 +125,18 @@ func (o *InitOptions) Run() (err error) {
 		}
 	}()
 
+	isEmptyDir, err := location.DirIsEmpty(o.clientset.FS, o.contextDir)
+	if err != nil {
+		return err
+	}
+	if isEmptyDir && len(o.flags) == 0 {
+		fmt.Println("The current directory is empty.")
+		fmt.Println("Odo will help with you starting a new project.")
```
cdrage, line 134: `odo` should not be capitalized

Please avoid using newline / merge these messages into one line :)

pkg/odo/cli/init/init.go
```diff
@@ -125,6 +125,18 @@ func (o *InitOptions) Run() (err error) {
...
+	} else if len(o.flags) == 0 {
+		fmt.Println("The current directory already contains source code.")
+		fmt.Println("Odo will try to autodetect language and project type to select best suited Devfile for your project.")
```
cdrage, line 137: `odo` should not be capitalized

Please avoid using newline / merge these messages into one line :)

cdrage (changes requested):

pkg/odo/cli/init/init.go
cdrage: Correct grammar would be` odo will help you start a new project`

pkg/odo/cli/init/init.go
cdrage: Correct grammar would be: `odo will try to autodetect the language and project type in order to select the best suited Devfile for your project`

pkg/odo/cli/init/init.go
rm3l: Ok, I'll change those lines. I copied the messages provided in the parent issue #5494.
Thanks.

pkg/odo/cli/init/init.go
rm3l: Ok, I'll change those lines. I copied the messages provided in the parent issue #5494.
Thanks.

cdrage (approved):
/lgtm

PR #5538, opened 2022-03-10
Make sure Makefile targets for both integration and E2E tests depend on the 'install' target

rm3l:
**What type of PR is this:**
/kind tests

Integration and E2E tests targets rely on the `odo` binary to execute subcommands in separate processes.

While iterating on the code, we could often end up with situations where  the `odo` binary is no longer up-to-date, should we forget to `make install` before.
This fixes such behavior by ensuring the installed binary is always up-to-date prior to running the actual test target.

I am not sure yet of the implications (e.g., in terms of performance in CI), but I guess there should be a minimal impact thanks to the Go build cache.

Let me know if you think this should be done differently.

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test
- [x] E2E test 

- [ ] Documentation 

Run any of the modified targets in verbose mode, and make sure the `install` target is executed first. For instance, with a dry run:
```shell
❯ make -n --debug='j,v' test-integration-interactive
GNU Make 4.3
Built for x86_64-redhat-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Reading makefile 'Makefile'...
Updating makefiles....
Updating goal targets....
Considering target file 'test-integration-interactive'.
 File 'test-integration-interactive' does not exist.
  Considering target file 'install'.
   File 'install' does not exist.
   Finished prerequisites of target file 'install'.
  Must remake target 'install'.
go install -mod=vendor -ldflags="-X github.com/redhat-developer/odo/pkg/version.GITCOMMIT=29f7e659f" ./cmd/odo/
  Successfully remade target file 'install'.
 Finished prerequisites of target file 'test-integration-interactive'.
Must remake target 'test-integration-interactive'.
GOFLAGS='-mod=vendor' go run -mod=vendor github.com/onsi/ginkgo/ginkgo  -randomizeAllSpecs -slowSpecThreshold=120 -timeout 14400s -nodes=4 tests/integration/interactive/
Successfully remade target file 'test-integration-interactive'.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ default: bin
 
 .PHONY: help
 help: ## Show this help.
-	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {sub("\\\\n",sprintf("\n%22c"," "), $$2);printf "\033[36m%-26s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {sub("\\\\n",sprintf("\n%22c"," "), $$2);printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
 .PHONY: bin
 bin: ## build the odo binary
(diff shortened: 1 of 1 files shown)


feloy (approved):

anandrkskd:
Looks good to me 
/approve

kadel:
I think that this is going to cause some problems. But it is possible that this is just me :-) 
I often run tests against different versions of odo binaries and quite often they don't match the version of the code that I have checked out from git. 
I don't have `GOPATH` in my `PATH` so luckily this won't affect my workflow. But Someone might be actually surprised that executing tests builds odo as well. 

But if everyone is ok with that it is OK :-)

rm3l:
> I think that this is going to cause some problems. But it is possible that this is just me :-) I often run tests against different versions of odo binaries and quite often they don't match the version of the code that I have checked out from git.
> 

Thanks for raising this.
But then, how do you make sure the tests you ran correspond to the features of your `odo` binary? Or is this something you first check manually in the tests?

> But Someone might be actually surprised that executing tests builds odo as well.
>
Maybe I am a bit influenced by my background with other build tools like Gradle / Maven (where test-related tasks make sure to first compile the artifact and place it in a dedicated path so as to be consumed in the actual tests), but why do you find this surprising? I can't see why :)
Because this was precisely the issue I had: tests using an `odo` binary that does not match the current version of the code.

I saw a similar strategy used for writing integration tests for CLI applications written in Rust, where the test itself compiles the main binary before using it: https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

> I don't have `GOPATH` in my `PATH` so luckily this won't affect my workflow. 
>
This makes me think that instead of `install`ing the binary in `GOPATH`, a better approach could be to just build the artifact instead (i.e., depend on `make bin` instead), and resolve the binary path in the tests. This would ideally allow us to avoid potential collisions.
I will try to submit another PR when I get a chance. This would allow overriding the `odo` binary path if needed (thus skipping the build). Would that make sense for you?

kadel:
Don't worry about me :-) 
If this works for everyone it is great.

My setup is fairly specific because I do a lot of weird tests with odo.

rm3l:
:D Okay - thanks for your input.

PR #5572, opened 2022-03-18
Use .gitignore file for odo dev

### LINKED ISSUE #5581: odo dev does not use .gitignore to sync files to container
/kind bug

**Operating System:**

Linux Fedora

**Output of `odo version`:**

odo v2.5.0 (5aa1dc1d0)

```
$ odo init --name myapp --devfile go --starter go-starter
$ mkdir -p a/b b
$ echo b/ >> .gitignore
$ odo dev
```

```
$ kubectl exec  deployment/myapp-app -- ls /projects
README.md
a
b
devfile.yaml
go.mod
main
main.go
```

## Actual behavior

`b` directory is present in the container

## Expected behavior

`b` directory should not be synchronized to container.

feloy:
**What type of PR is this:**

/kind bug

This PR makes `odo dev` use the content of `.gitignore` file

Fixes #5581 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

Blocked by #5569 to write integration test with `odo dev`

### DIFF
--- a/pkg/dev/dev.go
+++ b/pkg/dev/dev.go
@@ -6,6 +6,7 @@ import (
 	"github.com/redhat-developer/odo/pkg/envinfo"
 
 	"github.com/devfile/library/pkg/devfile/parser"
+	"github.com/devfile/library/pkg/util"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters/common"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters/kubernetes"
@@ -26,7 +27,7 @@ func NewDevClient(watchClient watch.Client) *DevClient {
 	}
 }
 
-func (o *DevClient) Start(devfileObj parser.DevfileObj, platformContext kubernetes.KubernetesContext, path string) error {
+func (o *DevClient) Start(devfileObj parser.DevfileObj, platformContext kubernetes.KubernetesContext, ignorePaths []string, path string) error {
 	klog.V(4).Infoln("Creating new adapter")
 	adapter, err := adapters.NewComponentAdapter(devfileObj.GetMetadataName(), path, "app", devfileObj, platformContext)
 	if err != nil {
@@ -42,6 +43,7 @@ func (o *DevClient) Start(devfileObj parser.DevfileObj, platformContext kubernet
 		EnvSpecificInfo: *envSpecificInfo,
 		DebugPort:       envSpecificInfo.GetDebugPort(),
 		Path:            path,
+		IgnoredFiles:    ignorePaths,
 	}
 
 	klog.V(4).Infoln("Creating inner-loop resources for the component")
@@ -64,6 +66,8 @@ func (o *DevClient) Watch(devfileObj parser.DevfileObj, path string, ignorePaths
 		return err
 	}
 
+	absIgnorePaths := util.GetAbsGlobExps(path, ignorePaths)
+
 	watchParameters := watch.WatchParameters{
 		Path:                path,
 		ComponentName:       devfileObj.GetMetadataName(),
@@ -71,7 +75,7 @@ func (o *DevClient) Watch(devfileObj parser.DevfileObj, path string, ignorePaths
 		ExtChan:             make(chan bool),
 		DevfileWatchHandler: h.RegenerateAdapterAndPush,
 		EnvSpecificInfo:     envSpecificInfo,
-		FileIgnores:         ignorePaths,
+		FileIgnores:         absIgnorePaths,
 		InitialDevfileObj:   devfileObj,
 	}
 
--- a/pkg/dev/interface.go
+++ b/pkg/dev/interface.go
@@ -13,7 +13,7 @@ import (
 
 type Client interface {
 	// Start the resources in devfileObj on the platformContext. It then pushes the files in path to the container.
-	Start(devfileObj parser.DevfileObj, platformContext kubernetes.KubernetesContext, path string) error
+	Start(devfileObj parser.DevfileObj, platformContext kubernetes.KubernetesContext, ignorePaths []string, path string) error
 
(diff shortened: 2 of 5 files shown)


rm3l (approved):
LGTM from the code perspective.

One question I was wondering out of curiosity after testing this. Maybe this is a larger question and not necessarily related to your changes, but what is expected if the `.gitignore` file itself is modified during an `odo dev` session, e.g. to ignore something that was already sync'ed but was not supposed to?
I was thinking of quite frequent scenarios where a file could be created locally, but not added yet to `.gitignore`. So this file would get sync'ed into the container, but will still be in the container if `git-ignore`'d later on, unless the `odo dev` session is restarted.
I think we should handle this case and remove the files/directories from the containers.

valaparthvi:
I am not sure if this works.
1. `mkdir /tmp/102; cd /tmp/102`
2. `odo init --name mynode --devfile nodejs --starter-project nodejs-starter`
3. echo `foobar.txt >> .gitignore`
4. `odo dev`
5. `touch foobar.txt`
6. `k exec -it <pod_name> -- ls /project`
I expect foobar.txt to not be pushed, but it's still pushed. Am I doing understanding this correctly?

feloy:
> I am not sure if this works.
> 
> 1. `mkdir /tmp/102; cd /tmp/102`
> 2. `odo init --name mynode --devfile nodejs --starter-project nodejs-starter`
> 3. echo `foobar.txt >> .gitignore`
> 4. `odo dev`
> 5. `touch foobar.txt`
> 6. `k exec -it <pod_name> -- ls /project`
> 
> I expect foobar.txt to not be pushed, but it's still pushed. Am I understanding this correctly?

This PR mostly fixes the point 4. After 4, foobar.txt should not be in the container. Is it?

If it is not present after 4. but pushed after 5., it is another bug :(

valaparthvi:
> case

I am not sure I understand this behavior. So if I create a file, add it to .gitignore, and then do `odo dev`, it doesn't push file, but if I add a new file to .gitignore, create it, and then do `odo dev`, it pushes both the files to the cluster.

feloy:
> > > I am not sure if this works.
> > > 
> > > 1. `mkdir /tmp/102; cd /tmp/102`
> > > 2. `odo init --name mynode --devfile nodejs --starter-project nodejs-starter`
> > > 3. echo `foobar.txt >> .gitignore`
> > > 4. `odo dev`
> > > 5. `touch foobar.txt`
> > > 6. `k exec -it <pod_name> -- ls /project`
> > > 
> > > I expect foobar.txt to not be pushed, but it's still pushed. Am I understanding this correctly?
> > 
> > 
> > This PR mostly fixes the point 4. After 4, foobar.txt should not be in the container. In this case, it won't as it is created after 4., but if you create it before 4., it should not be pushed during 4.
> > If it is not present after 4. but pushed after 5., it is another bug :(
> 
> I am not sure I understand this behavior. So if I create a file, add it to .gitignore, and then do `odo dev`, it doesn't push file, but if I add a new file to .gitignore, create it, and then do `odo dev`, it pushes both the files to the cluster.

Because there are 2 different code that push the code to the container.
- one that pushes the code when you run `odo dev`
- one that synchronize the files between local and the container when some files are modified locally.

This PR fixes only the first point, the seoncd one is more complex (you can modify both files and .gitignore at the same time...)

rm3l:
> LGTM from the code perspective.
> 
> One question I was wondering out of curiosity after testing this. Maybe this is a larger question and not necessarily related to your changes, but what is expected if the `.gitignore` file itself is modified during an `odo dev` session, e.g. to ignore something that was already sync'ed but was not supposed to? I was thinking of quite frequent scenarios where a file could be created locally, but not added yet to `.gitignore`. So this file would get sync'ed into the container, but will still be in the container if `git-ignore`'d later on, unless the `odo dev` session is restarted. I think we should handle this case and remove the files/directories from the containers.

As I understand from the previous comments, this specific point of syncing files while `odo dev` is running is a separate issue then, which should be able to handle updates to `.gitignore` as well.

feloy:
> > LGTM from the code perspective.
> > One question I was wondering out of curiosity after testing this. Maybe this is a larger question and not necessarily related to your changes, but what is expected if the `.gitignore` file itself is modified during an `odo dev` session, e.g. to ignore something that was already sync'ed but was not supposed to? I was thinking of quite frequent scenarios where a file could be created locally, but not added yet to `.gitignore`. So this file would get sync'ed into the container, but will still be in the container if `git-ignore`'d later on, unless the `odo dev` session is restarted. I think we should handle this case and remove the files/directories from the containers.
> 
> As I understand from the previous comments, this specific point of syncing files while `odo dev` is running is a separate issue then, which should be able to handle updates to `.gitignore` as well.

Yes, this problem is not related to this PR, which only cares about the first push of the files.

But you are right, deleting a file that has been added to .gitignore should occur, but I don't think it is done for the moment

rm3l:
> > > LGTM from the code perspective.
> > > One question I was wondering out of curiosity after testing this. Maybe this is a larger question and not necessarily related to your changes, but what is expected if the `.gitignore` file itself is modified during an `odo dev` session, e.g. to ignore something that was already sync'ed but was not supposed to? I was thinking of quite frequent scenarios where a file could be created locally, but not added yet to `.gitignore`. So this file would get sync'ed into the container, but will still be in the container if `git-ignore`'d later on, unless the `odo dev` session is restarted. I think we should handle this case and remove the files/directories from the containers.
> > 
> > 
> > As I understand from the previous comments, this specific point of syncing files while `odo dev` is running is a separate issue then, which should be able to handle updates to `.gitignore` as well.
> 
> Yes, this problem is not related to this PR, which only cares about the first push of the files.
> 
> But you are right, deleting a file that has been added to .gitignore should occur, but I don't think it is done for the moment

Okay - let's create a separate issue once this PR is merged.
LGTM then, after testing your branch.
I'll let @valaparthvi approve this if it's ok for her.

PR #5596, opened 2022-03-28
Cleans up innerloop resources upon Ctrl+c

### LINKED ISSUE #5532: When odo dev detects changes it pushes files twice
/kind bug

**Operating System:**
MacOS
**Output of `odo version`:**
-> % odo version
odo v2.5.0 (b39534823)
Happens when using either Kubernetes or openshift clusters

## Actual behavior
When odo dev detects changes in code it pushes files twice

## Expected behavior
It should push changes just once

I made a change in one file, then saved, and odo dev reported detecting changes twice, and pushed twice:
```
-> % odo dev                                 
Using devfile.yaml from the current directory.
Starting your application on cluster in developer mode
 ✓  Waiting for Kubernetes resources [4s]
 ✓  Syncing files into the container [2ms]

Your application is running on cluster.
 
Waiting for something to change in /Users/rnapoles/temp/t1

Press Ctrl+c to exit and clean up resources from cluster.


File /Users/rnapoles/temp/t1/main.go changed
Pushing files...
 ✓  Waiting for Kubernetes resources [4s]
 ✓  Syncing files into the container [2s]
 ✓  Building your application in container on cluster [2s]
 ✓  Executing the application [2s]

Waiting for something to change in /Users/rnapoles/temp/t1

Press Ctrl+c to exit and clean up resources from cluster.


File /Users/rnapoles/temp/t1/main.go changed
Pushing files...
 ✓  Waiting for Kubernetes resources [5s]
 ✓  Syncing files into the container [2s]
 ✓  Building your application in container on cluster [2s]
 ✓  Executing the application [3s]

Waiting for something to change in /Users/rnapoles/temp/t1

Press Ctrl+c to exit and clean up resources from cluster.
```

### LINKED ISSUE #5595: Stop catching `os.Interrupt` signal for Telemetry in case of `odo dev`
In the case of `odo dev`, interrupt is not really interrupt that we want to catch as an error. Perhaps you can add a check for `odo dev` before the signal watcher starts.
https://github.com/redhat-developer/odo/blob/main/pkg/odo/genericclioptions/runnable.go#L73

Check if `cmd.Name() == "dev"`, and if it is, don't add the `os.Interrupt` signal to the list. @feloy what do you think?

_Originally posted by @valaparthvi in https://github.com/redhat-developer/odo/pull/5576#discussion_r836294874_

dharmit:
**What type of PR is this:**

$subject

Fixes part of #5485
Fixes #5532 
Fixes #5595

**PR acceptance criteria:**

- [ ] Unit test

- [ ] Integration test 

- [ ] Documentation 

Run odo dev and hit `Ctrl+c`. Innerloop resources should get deleted.

**Things from acceptance criteria addressed in this PR:**
IMHO, Done
- [x] Before the command ends it should delete all **inner loop** resources that it created on the cluster.
  - [x] There are different ways how the command can end (for example SIGQUIT, SIGHUP, SIGINT etc...) Program should catch all those signals and do cleanup before exiting.
  - [x] if the command is interrupted in the middle of some other operation (for example file syncing, k8s resource creation etc...) it should complete what is is doing, and when it gets to the point when `odo dev` is waiting or files to change start the cleanup procedure
- [x] The last line of the output needs to always inform user that ctrl+c will end the command and that it will result in deleting all resources from the cluster

Prefer separate PR:
- [ ] at least one test with devfile that creates storage, k8s. networking/Service (endpoint), k8s component. 
- [ ] verify that this work on Windows


pkg/odo/cli/dev/dev.go
```diff
@@ -267,6 +281,13 @@ func printPortForwardingInfo(portPairs map[string][]string, out io.Writer) {
 	fmt.Fprintf(out, "\n%s", portForwardURLs.String())
 }
 
+func (o *DevOptions) HandleSignal() error {
+	o.cancel()
+	<-o.cleanupDone
+	time.Sleep(10 * time.Second)
```
dharmit: @kadel @feloy I have removed this sleep in a subsequent commit, but you could add this to test that when cleanup logic is being executed, it doesn't re-trigger the watch logic.

pkg/watch/watch.go
```diff
@@ -178,34 +187,43 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters) er
...
-				}
-			case event := <-watcher.Events:
-				changeLock.Lock()
+	wg.Add(1)
+	for {
+		select {
+		case event := <-watcher.Events:
+			// check if the delay between events is more than a second; if not don't push again
+			if time.Now().After(lastChange.Add(delay)) {
+				lastChange = time.Now()
```
dharmit, line 225: @feloy one of the problems I am facing after removing that infinite for loop and doing everything in this single go routine is that https://github.com/redhat-developer/odo/issues/5532 pops up frequently. 

I added the above logic to ensure that we don't do odo dev too often, but only if the gap between two successive events is more than 1 second. Even though it fixed the problem initially, it came back, and I am seeing odo dev push things not just twice but sometimes as much as 4 times!

pkg/watch/watch.go
```diff
@@ -178,34 +187,43 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters) er
...
-					changeLock.Unlock()
-				}
-			case event := <-watcher.Events:
-				changeLock.Lock()
+	wg.Add(1)
+	for {
+		select {
+		case event := <-watcher.Events:
+			// check if the delay between events is more than a second; if not don't push again
+			if time.Now().After(lastChange.Add(delay)) {
```
feloy, line 224: I don't think it will work this way. Instead, you should create a timer of 'delay' duration that you start when an event is received, and you add a case to catch the end of the timer. If you receive another event before the end of the timer, you reinitialize the timer to delay, and if you receive a timer end, you can proceed

feloy (changes requested):

pkg/watch/watch.go
dharmit: PTAL at 86c8e04.

pkg/watch/watch.go
```diff
@@ -178,34 +187,47 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters) er
...
 			select {
-			case extMsg := <-parameters.ExtChan:
-				if extMsg {
-					changeLock.Lock()
-					watchError = ErrUserRequestedWatchExit
-					changeLock.Unlock()
-				}
 			case event := <-watcher.Events:
-				changeLock.Lock()
+				lastEvent = event
```
feloy, line 227: You can have several `watcher.Events` before to have `timer.C`, so you need to store your events into an array of events: `lastEvents = append(lastEvents, event)`

feloy (changes requested):

pkg/watch/watch.go
dharmit: But wouldn't that cause the same problem because `lastEvents` slice/array would have multiple events triggered for a single change?

feloy: Probably you could filter it to remove duplicates?

pkg/odo/cli/dev/dev.go
```diff
@@ -267,6 +281,12 @@ func printPortForwardingInfo(portPairs map[string][]string, out io.Writer) {
 	fmt.Fprintf(out, "\n%s", portForwardURLs.String())
 }
 
+func (o *DevOptions) HandleSignal() error {
+	o.cancel()
+	<-o.cleanupDone
```
feloy, line 286:
```suggestion
	// we are waiting for the cleanup to finish, before to continue the signal handler processing
	<-o.cleanupDone
```

pkg/watch/watch.go
dharmit: Probably. But what would be the benefit compared to adding the logic for filtering? In the interval of 300 ms (the duration of the timer), isn't it safe to assume there won't be too many events that we would need to filter them?

PS: I might be horribly misunderstanding `timer.Timer`, so correct me if I am wrong.

pkg/watch/watch.go
```diff
@@ -193,205 +199,145 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters) er
...
+		case <-parameters.ExtChan:
+			watchError = ErrUserRequestedWatchExit
 			return watchError
-		}
-		if showWaitingMessage {
-			if parameters.EnvSpecificInfo != nil && parameters.EnvSpecificInfo.GetRunMode() == envinfo.Debug {
-				log.Finfof(out, "Component is running in debug mode\nPlease start port-forwarding in a different terminal.")
+		case event := <-watcher.Events:
+			lastEvent = event
+			timer.Reset(100 * time.Millisecond)
```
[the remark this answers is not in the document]
dharmit: I don't know, I still don't understand timer well. But I tried what you suggested.

Replacing:
```go
timer.Reset(100 * time.Millisecond)
``` 
with
```go
if !timer.Stop(){
    <-timer.C
}
timer.Reset(100 * time.Millisecond)
```
causes `odo dev` to not respond to a change in the file. That is, it doesn't trigger a push when we make a change in component directory.



What I think (my understanding of timers is still evolving) is happening is:
1. when there's an event, it gets stored in `lastEvent` and the timer is reset, which prevents the `case <-timer.C` from getting triggered.
2. this keeps happening a few times because a single file save triggers multiple fsnotify events.
3. finally, when all events triggered as a result of that single save are reported by fsnotify and overridden in `lastEvent` variable, the `case <- timer.C` gets triggered a 100ms after step 2 stops looping.

pkg/watch/watch.go
```diff
@@ -193,205 +199,145 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters) er
...
+			}
+		case watchErr := <-watcher.Errors:
+			return watchErr
+		case <-ctx.Done():
+			devfileObj := parameters.InitialDevfileObj
+			err = o.kubernetesClient.DeleteDeployment(componentlabels.GetLabels(devfileObj.GetMetadataName(), "app", false))
+			if err != nil {
+				fmt.Fprintf(out, "failed to delete inner loop resources: %v", err)
+			}
+			cleanupDone <- true
```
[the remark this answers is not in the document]
dharmit: Not sure what you mean. Why would we want to sleep here? Wouldn't we want to continue with the next steps?

pkg/watch/watch.go
dharmit: Makes sense. I'll append events and keep the unique ones. 👍🏾

pkg/odo/cli/dev/dev.go
```diff
@@ -256,6 +267,14 @@ func regenerateComponentAdapterFromWatchParams(parameters watch.WatchParameters)
 	return adapters.NewComponentAdapter(parameters.ComponentName, parameters.Path, parameters.ApplicationName, devObj, platformContext)
 }
 
+func (o *DevOptions) HandleSignal() error {
+	o.cancel()
+	// At this point, `ctx.Done()` will be raised, and the cleanup will be done
+	// we are waiting for the cleanup to finish, before to continue the signal handler processing
+	for {
```
feloy, line 274: Finally, I'm not sure it is a good practice to have a long running signal handler. Feel free to look at this patch for another handling solution: https://github.com/feloy/odo/commit/4b9b49029795704afec57186ec6b51ab64d39c4c

feloy (changes requested):

pkg/odo/cli/dev/dev.go
dharmit: Yeah, I am not sure either. The onus is on the handler to make sure that the program doesn't go into a state of waiting forever that would require something like a `SIGKILL`. That could be a good thing or a bad thing depending on the situation we are in. As far as `odo dev` in the current PR is concerned, it's handling it correctly, so it shouldn't be a problem.

How about we keep it this way for the time being and see if it becomes a problem? I could very well just modify the implementation to what you have in your commit, but if we are not sure that it's definitely bad practice, why not keep it instead...

feloy: ok, let's keep it as this and see

pkg/odo/cli/dev/dev.go
```diff
@@ -268,11 +268,11 @@ func regenerateComponentAdapterFromWatchParams(parameters watch.WatchParameters)
 }
 
 func (o *DevOptions) HandleSignal() error {
+	fmt.Fprintf(o.out, "\n\nCancelling deployment. Press Ctrl-c again to terminate immediately\n\n")
```
feloy, line 271: The message is not accurate. If you press Ctrl-C again, nothing happens (precisely because we didn't return from the signal handler)

feloy (changes requested):

feloy:
You will need to adapt this test with the new behaviour of odo when it terminates (related to #5595):

```
• Failure [47.236 seconds]
odo dev command tests
/workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:23
  when a component is bootstrapped and pushed
  /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:53
    when odo dev is executed
    /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:168
      when recording telemetry data
      /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:208
        should record the telemetry data correctly [It]
        /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:217

        Expected
            <bool>: true
        to be false

        /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:220
```

pkg/odo/cli/dev/dev.go
dharmit: Removed. It wasn't valid to add this without adding the other parts of your example commit.

dharmit:
> You will need to adapt this test with the new behaviour of odo when it terminates (related to #5595):
> 
> ```
> • Failure [47.236 seconds]
> odo dev command tests
> /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:23
>   when a component is bootstrapped and pushed
>   /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:53
>     when odo dev is executed
>     /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:168
>       when recording telemetry data
>       /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:208
>         should record the telemetry data correctly [It]
>         /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:217
> 
>         Expected
>             <bool>: true
>         to be false
> 
>         /workspace/6c1f4ab8-732d-4c15-bdb7-b2c40e1ce749/tests/integration/devfile/cmd_dev_test.go:220
> ```

Thanks for pointing. 
Removed a few checks because they weren't valid anymore for `odo dev` scenario.

pkg/odo/cli/dev/dev.go
feloy: You could keep the first part of the message. It is disappoointing for the user to press Ctrl-C and see nothing happening during the deployment time

dharmit: WDYT about a message like below:
```go
fmt.Fprintf(o.out, "\n\nCancelling deployment.\n\nThis is non-preemptive operation, it will wait for other tasks to finish first")
```

feloy:
/lgtm

Thanks @dharmit for this PR 

IBM Cloud tests pass

PR #5598, opened 2022-03-29
odo delete component - show warning that something was not deleted

### LINKED ISSUE #5555: `odo delete component` - show warning that something was not deleted
Show warning when the component was deleted but there are still resources matching labels.
This situation can happen when the devfile was modified and something was removed from it.

It could look like this

```
$ odo delete component
Searching resources to delete, please wait...
This will delete "devfile-nodejs-deploy" from the namespace "myproj".
	- Deployment: devfile-nodejs-deploy-app
? Are you sure you want to delete "devfile-nodejs-deploy" and all its resources? (y/N) y
The component "devfile-nodejs-deploy" is successfully deleted from namespace "myproj"

There are still resources left in the cluster that might be belonging to the deleted component.
 - Service: something

If you want to delete those, execute `odo delete component --name devfile-nodejs-deploy`
```

valaparthvi:

**What type of PR is this:**
/kind feature

This PR adds information about not deleted component resources because devfile.yaml was changed.

Fixes #5555 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

1. `mkdir /tmp/101; cd /tmp/101`
2. `cp $ODOPATH/tests/examples/source/devfiles/nodejs/devfile-deploy-with-multiple-resources.yaml devfile.yaml`
3. `cp -r $ODOPATH/tests/examples/source/nodejs/* . `
4. `export PODMAN_CMD=echo`
5. `odo deploy`
6. `odo dev`
7. Replace `my-cs` service name to `my-changed-cs` in devfile.yaml
8. `odo delete component`

Output will be similar to:
```sh
➜  odo delete component                                                                                                                                                  
Searching resources to delete, please wait...                                                                                                                                
This will delete "mynodejs" from the namespace "default".                                                                                                                    
        - Deployment: mynodejs-app                                                                                                                                           
? Are you sure you want to delete "mynodejs" and all its resources? Yes                                                                                                      
The component "mynodejs" is successfully deleted from namespace "default"                                                                                                    
 •  There are still resources left in the cluster that might be belonging to the deleted component.
        - Service: my-cs
        - Deployment: my-component
        - Endpoints: my-cs
If you want to delete those, execute `odo delete component --name mynodejs --namespace default`
```

### DIFF
--- a/pkg/odo/cli/delete/component.go
+++ b/pkg/odo/cli/delete/component.go
@@ -129,26 +129,15 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
 		log.Infof("No resource found for component %q in namespace %q\n", componentName, namespace)
 		return nil
 	}
-	var remainingResources []unstructured.Unstructured
-	k8sResources, _ := o.clientset.DeleteClient.ListClusterResourcesToDelete(componentName, namespace)
-	// get resources present in k8sResources(present on the cluster) but not in devfileResources(not present in the devfile)
-	for _, k8sresource := range k8sResources {
-		var present bool
-		for _, dresource := range devfileResources {
-			//  skip if the cluster and devfile resource are same OR if the cluster resource is the component's Endpoints resource
-			if reflect.DeepEqual(dresource, k8sresource) || (k8sresource.GetKind() == "Endpoints" && strings.Contains(k8sresource.GetName(), componentName)) {
-				present = true
-				break
-			}
-		}
-		if !present {
-			remainingResources = append(remainingResources, k8sresource)
-		}
-	}
 	// Print all the resources that odo will attempt to delete
 	printDevfileComponents(componentName, namespace, devfileResources)
 
 	if o.forceFlag || ui.Proceed(fmt.Sprintf("Are you sure you want to delete %q and all its resources?", componentName)) {
+		// Get a list of component's resources present on the cluster
+		clusterResources, _ := o.clientset.DeleteClient.ListClusterResourcesToDelete(componentName, namespace)
+		// Get a list of component's resources absent from the devfile, but present on the cluster
+		remainingResources := listResourcesMissingFromDevfilePresentOnCluster(componentName, devfileResources, clusterResources)
+
 		// if innerloop deployment resource is present, then execute preStop events
 		if isInnerLoopDeployed {
 			err = o.clientset.DeleteClient.ExecutePreStopEvents(devfileObj, appName)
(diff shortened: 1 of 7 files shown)


pkg/odo/cli/delete/component.go
```diff
@@ -119,17 +121,33 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
...
-	if len(resources) == 0 {
+	if len(devfileResources) == 0 {
 		log.Infof("No resource found for component %q in namespace %q\n", componentName, namespace)
 		return nil
 	}
-
+	var remainingResources []unstructured.Unstructured
+	k8sResources, _ := o.clientset.DeleteClient.ListResourcesToDelete(componentName, namespace)
+	// get resources present in k8sResources(present on the cluster) but not in devfileResources(not present in the devfile)
+	if len(k8sResources) != 0 {
```
dharmit, line 135: Is this check redundant? I wonder so because we are already returning when `if len(devfileResources) == 0`, and I think it's safe to assume that if that condition is not satisfied, this condition will never be satisfied. WDYT?

pkg/odo/cli/delete/component.go
```diff
@@ -119,17 +121,33 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
...
+		for _, k8sresource := range k8sResources {
+			var present bool
+			for _, dresource := range devfileResources {
+				//                                              skip the component's endpoints resource
```
dharmit, line 139:
```suggestion
				// skip the component's endpoints resource
```

dharmit: Also, there's more happening below than just skipping an endpoint resource. Can you summarize that briefly in the code comment?

pkg/odo/cli/delete/component.go
```diff
@@ -119,17 +121,33 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
...
+				if reflect.DeepEqual(dresource, k8sresource) || (k8sresource.GetKind() == "Endpoints" && strings.Contains(k8sresource.GetName(), componentName)) {
+					present = true
```
dharmit, line 141: Should we add a `break` statement here so that it doesn't check for the remaining resources in the slice?

pkg/odo/cli/delete/component.go
```diff
@@ -119,17 +121,33 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
...
 		return err
 	}
-	if len(resources) == 0 {
+	if len(devfileResources) == 0 {
 		log.Infof("No resource found for component %q in namespace %q\n", componentName, namespace)
 		return nil
 	}
-
+	var remainingResources []unstructured.Unstructured
+	k8sResources, _ := o.clientset.DeleteClient.ListResourcesToDelete(componentName, namespace)
```
dharmit, line 133: This is not a part of this PR, but, IMO, this method should have been named `ListKubernetesResourceToDelete` instead. The other method in this interface, `ListResourcesToDeleteFromDevfile`, is clear from its name itself as to which resources it will delete.

pkg/odo/cli/delete/component.go
```diff
@@ -141,11 +159,19 @@ func (o *ComponentOptions) deleteDevfileComponent() error {
...
 		for _, fail := range failed {
 			log.Warningf("Failed to delete the %q resource: %s\n", fail.GetKind(), fail.GetName())
 		}
 		log.Infof("The component %q is successfully deleted from namespace %q", componentName, namespace)
+
+		if len(remainingResources) != 0 {
+			log.Printf("There are still resources left in the cluster that might be belonging to the deleted component.")
+			for _, resource := range remainingResources {
+				fmt.Printf("\t- %s: %s\n", resource.GetKind(), resource.GetName())
+			}
```
dharmit, line 172: Can we use the tabwriter here instead? It takes care of formatting. We have used it in the odo code in the past.

dharmit (changes requested):

feloy:
Please update the mock so the unit tests and validation pass
/approve

PR #5604, opened 2022-03-30
Wait deletion

### LINKED ISSUE #5565: `odo delete component --wait`
## User Story
- As an odo user 
- I want to `odo delete component` command to exit only after a component was deleted from the cluster
- So that I have a confirmation when it is deleted, and I know when I can safely run `odo dev` or `odo deploy` again.

## Acceptance Criteria
applies for both `odo delete component` and `odo delete component --name`

- [ ] the command with `--wait` flag should wait for all resources to be deleted before it exits
- [ ]  there should be some sensible timeout so it doesn't end in an infinite loop. 
- [ ] If it is not possible to wait for resources to get deleted, odo should display a warning message stating that it might take some time before all resources are actually deleted from the cluster. 

### notes
some useful information about waiting for deletion and finalizers - https://github.com/redhat-developer/odo/issues/4733#issuecomment-847792991

feloy:
**What type of PR is this:**

/kind feature

- During odo dev, when creating resources (services, pvcs) attached to the main deployment, odo first tries to set `BlockOwnerDeletion` on the ownerReference, so the call to `Delete` with `PropagationPolicy` set to `DeletePropagationForeground` is effective. Because this operation needs specific rights, if the execution fails, it is not fatal and the ownerReference is still set but without `BlockOwnerDeletion`,
- a flag `--wait` is added to the `odo delete component` command, that calls the `Delete` functions with the `wait` parameter,
- When executing `Delete`  calls on resources with `wait` set to true, set  `PropagationPolicy` is set to `DeletePropagationForeground`, so the `Delete` call returns only when owned resources with `BlockOwnerDeletion` set are completely removed
- When executing `Delete`  calls on resources with `wait` set to true, the owned without `BlockOwnerDeletion` set are collected, and odo waits for their termination
- in `pkg/kclient` package, `*Dynamic*` methods are moved to the `dynamic.go` file, and `(group, version, kind string)` parameters are changed to a single parameter `schema.GroupVersionResource`.

Fixes #5565 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/component/delete/delete.go
+++ b/pkg/component/delete/delete.go
@@ -69,7 +69,7 @@ func (do *DeleteComponentClient) DeleteResources(resources []unstructured.Unstru
 			failed = append(failed, resource)
 			continue
 		}
-		err = do.kubeClient.DeleteDynamicResource(resource.GetName(), gvr.Resource.Group, gvr.Resource.Version, gvr.Resource.Resource, wait)
+		err = do.kubeClient.DeleteDynamicResource(resource.GetName(), gvr.Resource, wait)
 		if err != nil {
 			klog.V(3).Infof("failed to delete resource %q (%s.%s.%s): %v", resource.GetName(), gvr.Resource.Group, gvr.Resource.Version, gvr.Resource.Resource, err)
 			failed = append(failed, resource)
@@ -131,7 +131,7 @@ func (do DeleteComponentClient) ListResourcesToDeleteFromDevfile(devfileObj pars
 		}
 		// Try to fetch the resource from the cluster; if it exists, append it to the resources list
 		var cr *unstructured.Unstructured
-		cr, err = do.kubeClient.GetDynamicResource(gvr.Resource.Group, gvr.Resource.Version, gvr.Resource.Resource, lr.GetName())
+		cr, err = do.kubeClient.GetDynamicResource(gvr.Resource, lr.GetName())
 		if err != nil {
 			continue
 		}
--- a/pkg/component/delete/delete.go
+++ b/pkg/component/delete/delete.go
@@ -61,7 +61,7 @@ func (do *DeleteComponentClient) ListClusterResourcesToDelete(componentName stri
 	return result, nil
 }
 
-func (do *DeleteComponentClient) DeleteResources(resources []unstructured.Unstructured) []unstructured.Unstructured {
+func (do *DeleteComponentClient) DeleteResources(resources []unstructured.Unstructured, wait bool) []unstructured.Unstructured {
 	var failed []unstructured.Unstructured
 	for _, resource := range resources {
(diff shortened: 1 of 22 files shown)


pkg/kclient/owner_reference_test.go
```diff
@@ -0,0 +1,110 @@
...
+						default:
+							t.Errorf("only two calls should happen")
+							return nil
+						}
+					}
+				},
+			},
+			wantErr: true,
+		},
+		// TODO: Add test cases.
```
rm3l, line 100: To me, it looks like all cases in `TryWithBlockOwnerDeletion` are covered, no? So we could remove this `TODO` comment, I think.

tests/integration/devfile/cmd_dev_test.go
```diff
@@ -226,6 +226,65 @@ var _ = Describe("odo dev command tests", func() {
...
+					))
+				})
+			})
+
+			When("running odo delete component --wait", func() {
+				BeforeEach(func() {
+					helper.Cmd("odo", "delete", "component", "--wait", "-f").ShouldPass()
+				})
+
+				It("should have deleted all resources before to return", func() {
```
rm3l, line 272: Small suggestion ;)

```suggestion
				It("should have deleted all resources before returning", func() {
```

pkg/component/delete/delete.go
```diff
@@ -61,15 +61,15 @@ func (do *DeleteComponentClient) ListClusterResourcesToDelete(componentName stri
 	return result, nil
 }
 
-func (do *DeleteComponentClient) DeleteResources(resources []unstructured.Unstructured) []unstructured.Unstructured {
+func (do *DeleteComponentClient) DeleteResources(resources []unstructured.Unstructured, wait bool) []unstructured.Unstructured {
```
rm3l, line 64: Maybe I missed it, but per the acceptance criteria in #5565, I didn't see any sensible timeout defined when `wait` is `true`:

> there should be some sensible timeout so it doesn't end in an infinite loop.

rm3l (changes requested):

pkg/component/delete/delete.go
rm3l: Oh wait, I just saw a timeout defined in `WaitDynamicResourceDeleted`. So please ignore this comment :)

rm3l (approved):
LGTM. I tested your branch, and this new flag `--wait` worked like a charm.

One thing to note from the user perspective. No progress/output is displayed while waiting. I guess this might be problematic in the future if resource deletion takes long, until the timeout elapses. But this behavior can be fixed later.

rm3l (approved):

cdrage:
/approve

Tested and it works well here too! I didn't find any issues with the code at all.

PR #5610, opened 2022-03-30
Updates the error output for an empty directory

cdrage:
**What type of PR is this:**

Updates the error output for running `odo dev` or `odo deploy` in an
empty directory. Telling the user to run the same command in a non-empty
dir, OR use `odo init`.

N/A

**PR acceptance criteria:**

- [X] Unit test

- [X] Integration test

- [X] Documentation

### DIFF
--- a/pkg/odo/cli/deploy/deploy.go
+++ b/pkg/odo/cli/deploy/deploy.go
@@ -65,7 +65,7 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
 		return err
 	}
 	if isEmptyDir {
-		return errors.New("this command cannot run in an empty directory, you need to run it in a directory containing source code")
+		return errors.New("this command cannot run in an empty directory, run the command in a directory containing source code or initialize using 'odo init'")
 	}
 
 	initFlags := o.clientset.InitClient.GetFlags(cmdline.GetFlags())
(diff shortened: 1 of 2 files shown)


pkg/odo/cli/deploy/deploy.go
```diff
@@ -65,7 +65,7 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
 		return err
 	}
 	if isEmptyDir {
-		return errors.New("this command cannot run in an empty directory, you need to run it in a directory containing source code")
+		return errors.New("this command cannot run in an empty directory, run the command in a directory containing source code or initialize using 'odo init'")
```
valaparthvi, line 68: Can't we make define a sentinel error for this one in the component pkg? Something like the following - 
```go
var EmptyDirError := errors.New("this command cannot run in an empty directory, run the command in a directory containing source code or initialize using 'odo init'")
```

cdrage: Hey @valaparthvi I want to try and keep the error logging consistent across `odo` so I prefer not creating a sentinal error despite it being copy and pasted in two places...

Preferably because using those types of errors are unconventional and I want to make this PR more of a UI/UX change rather than adding to pkg/component instead of pkg/cli

valaparthvi: In that case,
/lgtm
/approve

PR #5613, opened 2022-03-30
Updates odo init output, fixes colorized output for tests.

cdrage:
**What type of PR is this:**

/kind feature

We have the logo appearing in `odo dev` and `odo deploy`.

Those 3 commands are the most used with `odo`, and the ones that are
most used. All three should have similar and consistent output that
contains:
  * What the component is doing (first line)
  * What's detected (in odo init, it's the files, in odo deploy and dev
    it's the component name)
  * odo version that is being used / outputted.

We also
* Make sure that colorized output is DISABLED when running tests
* Update the vendoring to enable `NO_COLOR` support
* Fixes tests when comparing colorized output vs non-colorized.

```sh
$ odo init
  __
 \__/  \    Files: No source code detected, a starter project will be created in the current directory
 /  \__/    odo version: v2.5.0
 \__/

? Select language: javascript
? Select project type: Next.js
 ✓  Downloading devfile "nodejs-nextjs" from registry "DefaultDevfileRegistry" [450ms]
? Which starter project do you want to use? nodejs-nextjs-starter
? Enter component name: my-nodejs-nextjs-app
 ✓  Downloading starter project "nodejs-nextjs-starter" [516ms]

Your new component 'my-nodejs-nextjs-app' is ready in the current directory.
To start editing your component, use 'odo dev' and open this folder in your favorite IDE.
Changes will be directly reflected on the cluster.
```

N/A

**PR acceptance criteria:**

- [X] Unit test

- [X] Integration test

- [X] Documentation

N/A


dharmit:
> ```shell
> Files: No source code detected, a starter project will be created in the current directory
> ```

This assumes that the user will always download a starter project, while they can choose not to do so. I am likely nitpicking, but felt like pointing out nevertheless.

rm3l:
@cdrage Indeed, it looks like the tests for this PR are not passing because of the welcoming messages added as part of #5514
Since the whole point of the issue reported (#5494) was to have such messages displayed first, I added tests that make sure such behavior is honored.
But I guess it should be ok to check that the first lines of the output contain the expected messages.

cdrage:
/hold

Trying to fix an issue with tests not passing...

pkg/component/component.go
```diff
@@ -20,6 +20,12 @@ import (
 
 const NotAvailable = "Not available"
 
+// Below are various "outputs" that we use in both the CLI and test cases, so we must
+// put them in a common place
+const InitializingNewComponent = "Initializing new component"
+const NoSourceCodeDetected = "Files: Source code detected, a Devfile will be determined based upon source code autodetection"
+const SourceCodeDetected = "Files: No source code detected, a starter project will be created in the current directory"
```
valaparthvi, line 27: The names are misplaced.

valaparthvi (changes requested):

pkg/log/status.go
```diff
@@ -294,6 +294,16 @@ func Title(firstLine, secondLine, thirdLine string) {
  \__/%s`, blue(firstLine), secondLine, thirdLine, "\n")
 }
 
+// Stitle is the same as Title but returns the string instead
+func Stitle(firstLine, secondLine, thirdLine string) string {
```
rm3l, line 298: If `Stitle` is the same as `Title` but just returning the string, why not make `Title` use the result of `Stitle` then? This would help reduce code duplication in this particular case, I think.

```go
func Title(firstLine, secondLine, thirdLine string) {
	fmt.Fprintf(GetStdout(), Stitle(firstLine, secondLine, thirdLine))
}
```

tests/helper/helper_interactive.go
```diff
@@ -57,7 +57,8 @@ func RunInteractive(command []string, env []string, tester Tester) (string, erro
 
 	term := vt10x.New(vt10x.WithWriter(pts))
 
-	c, err := expect.NewConsole(expect.WithStdin(ptm), expect.WithStdout(term), expect.WithCloser(pts, ptm), expect.WithDefaultTimeout(3*time.Minute))
+	// Set to 1 MINUTE timeout, since the command may take a while to start
+	c, err := expect.NewConsole(expect.WithStdin(ptm), expect.WithStdout(term), expect.WithCloser(pts, ptm), expect.WithDefaultTimeout(1*time.Minute))
```
rm3l, line 61: Wouldn't this be too short as a timeout, because `helper.RunInteractive` is called in few other tests that run `dev`, `deploy`. Those commands might take longer, no?
#5518 previously set this to 3 minutes, as a default timeout. But I am not sure how reducing this to 1 minute would impact the tests in CI.

rm3l (changes requested):
I added few comments. Let me know what you think.

pkg/component/component.go
```diff
@@ -20,6 +20,12 @@ import (
 
 const NotAvailable = "Not available"
 
+// Below are various "outputs" that we use in both the CLI and test cases, so we must
+// put them in a common place
+const InitializingNewComponent = "Initializing new component"
+const SourceCodeDetected = "Files: Source code detected, a Devfile will be determined based upon source code autodetection"
+const NoSourceCodeDetected = "Files: No source code detected, a starter project will be created in the current directory"
```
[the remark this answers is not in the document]
cdrage: It's to reduce copy-and-pasting in the tests :) So that's why I put them as constants.

pkg/log/status.go
cdrage: You're right! Thank you.

tests/helper/helper_interactive.go
cdrage: I've changed it back to 3 minutes now

pkg/log/status.go
cdrage: So unfortunately if I change it to the above suggestion, it results in a lint error:

```sh
pkg/log/status.go:289:2: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
```

We'll have to keep it to how it was before.

pkg/component/component.go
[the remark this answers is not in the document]
rm3l: I agree with defining constants, but my point was more about moving those constants to `pkg/odo/cli/init/init.go` instead of  `pkg/component/component.go`.
I feel like those constants should belong to the `init` CLI layer.
But maybe I missed something about the codebase structure :D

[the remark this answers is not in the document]
cdrage: Ah, but if you try to import `pkg/odo/cli/init` you'll run into issues in the testing suite! So into `component.go` it goes.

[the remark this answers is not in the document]
feloy: In this case, let's create a new package on `pkg/odo/cli`, for example `messages`, intended to declare all the messages strings that the tests could use. This way, you should not have cycled imports.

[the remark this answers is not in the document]
rm3l: > Ah, but if you try to import `pkg/odo/cli/init` you'll run into issues in the testing suite! So into `component.go` it goes.

I tried doing this yesterday, and Go was complaining that `init` needed to be a func. So I tried using an import alias instead, and the tests passed (no import cycle reported at that time).

> In this case, let's create a new package on `pkg/odo/cli`, for example `messages`, intended to declare all the messages strings that the tests could use. This way, you should not have cycled imports.

That would be a better alternative, I think.

pkg/log/status.go
rm3l: Ah, I see. As suggested in the [`Staticcheck` rule](https://staticcheck.io/docs/checks#SA1006), how about fixing this with `Fprint` instead, since `Stitle` returns the string and not a format string?
```go
func Title(firstLine, secondLine, thirdLine string) {
	fmt.Fprint(GetStdout(), Stitle(firstLine, secondLine, thirdLine))
}
```

rm3l (approved):
Thanks for the changes.

PR #5617, opened 2022-04-01
Fix potential issue with odo binary being outdated when running tests directly with Ginkgo

rm3l:
**What type of PR is this:**
/kind tests
/area testing

#5538 made both integration and E2E tests `Makefile` targets depend on the `install` one. This is fine when running the tests with `make`, but I noticed we could still run into the same issue when running them using the `ginkgo` CLI (as I often do when iterating on the code): the `odo` binary might be outdated and not match the current version of the code being tested, if one forgot to run `make install` before or make sure to add the relevant binary for `odo` to their `PATH`.

The approach I'm suggesting here is to:

1. Let Ginkgo build `odo` using the `make bin` target, prior to running the tests specs. To avoid building in parallel because of parallel tests specs, we could leverage Ginkgo's [`SynchronizedBeforeSuite/SynchronizedAfterSuite`](https://onsi.github.io/ginkgo/#parallel-suite-setup-and-cleanup-synchronizedbeforesuite-and-synchronizedaftersuite) to build the project only once. And this should not have an impact on performance thanks to the Go build cache.
2. Use the path to that built binary everywhere a test tries to run `odo`
3. Let developers/testers choose a different binary path if they wish to do so, e.g., via an environment variable (aptly named `ODO_BINARY_PATH`). If this environment variable is set, we would not build `odo`, but we would use that path instead everywhere a test runs `odo`. I'm thinking of @kadel 's comment here: https://github.com/redhat-developer/odo/pull/5538#issuecomment-1064989123

This is somehow inspired by how integration tests are written in Rust CLI apps: https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/.ibm/pipelines/kubernetes-tests.sh
+++ b/.ibm/pipelines/kubernetes-tests.sh
@@ -13,6 +13,7 @@ cleanup_namespaces
 (
     set -e
     make install
+    export ODO_BINARY_PATH="$(go env GOPATH)/bin/odo"
     make test-integration-devfile
     make test-interactive
     make test-e2e-devfile
--- a/.ibm/pipelines/openshift-tests.sh
+++ b/.ibm/pipelines/openshift-tests.sh
@@ -14,6 +14,7 @@ cleanup_namespaces
 (
     set -e
     make install
+    export ODO_BINARY_PATH="$(go env GOPATH)/bin/odo"
     make test-integration
     make test-interactive
     make test-integration-devfile
--- a/Makefile
+++ b/Makefile
@@ -160,119 +160,119 @@ test-windows:
 	go test $(UNIT_TEST_ARGS)  $(PKGS)
 
 .PHONY: test-generic
-test-generic: install ## Run generic integration tests
+test-generic: ## Run generic integration tests
 	$(RUN_GINKGO) $(GINKGO_FLAGS) -focus="odo generic" tests/integration/
 
 .PHONY: test-cmd-login-logout
-test-cmd-login-logout: install ## Run odo login and logout tests
+test-cmd-login-logout: ## Run odo login and logout tests
 	$(RUN_GINKGO) $(GINKGO_FLAGS_SERIAL) -focus="odo login and logout command tests" tests/integration/loginlogout/
 
 .PHONY: test-cmd-link-unlink-4-cluster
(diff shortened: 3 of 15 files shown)


feloy:
It is regularly necessary to run integration tests on several versions of odo to debug tests. With this entngled dependency, it will be more difficult to do this.

rm3l:
> It is regularly necessary to run integration tests on several versions of odo to debug tests. With this entngled dependency, it will be more difficult to do this.

The ability to specify the path to the `odo` binary via an environment variable, as proposed here, would solve this, no?

While I understand the need to test different versions of odo, I still don't get how we can be sure the binary tested is actually the right one (in line with the test code), unless you set up your environment accordingly.

Anyways, no problem if this issue impacts only me. I'll try to find a different way to work around this locally.

dharmit:
> we could still run into the same issue when running them using the `ginkgo` CLI

I used to run tests using `ginkgo` cli in the past when I was using vim to write Go. But since moving to GoLand, I use the `FIt`/`FDescribe` constructs of ginkgo and use GoLand to run the tests. The first thing it does when running tests is to build a temporary odo binary from the current code and use it for the tests. I imagine VS Code would do something similar.

rm3l:
> > we could still run into the same issue when running them using the `ginkgo` CLI
> 
> I used to run tests using `ginkgo` cli in the past when I was using vim to write Go. But since moving to GoLand, I use the `FIt`/`FDescribe` constructs of ginkgo and use GoLand to run the tests.

Are the `FIt`/`FDescribe` constructs the same as running `ginkgo --focus="test name"` ?

> The first thing it does when running tests is to build a temporary odo binary from the current code and use it for the tests. I imagine VS Code would do something similar.

I agree with you, but I think this works this way only if you call the project code like a library in the tests.
I was thinking of our integration tests that execute `odo` using `exec.Commmand`. The executable gets resolved from the `PATH`, so I think the issue of ensuring the right binary would still persist regardless of the IDE.
I'm also using GoLand, and for example, if you remove `odo` from your `PATH`, you will notice that the tests do not pass at all even if the IDE built the project first:

image

rm3l:
Closing. Found a workaround locally with a hook into the shell.

PR #5618, opened 2022-04-01
Refactor labels

feloy:
**What type of PR is this:**

/kind cleanup

This PR refactors the labels packages to completely hide the keys and instead expose the odo-related concepts: component name, app name, etc.

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


feloy:
@kadel Could you please help review this PR? 
I'm still confused about the use of the "app" and "component" labels. Are they really useful?

pkg/labels/selector_builder.go
```diff
@@ -0,0 +1,38 @@
...
+func SelectorBuilder() selectorBuilder {
+	return selectorBuilder{
+		selector: labels.NewSelector(),
+	}
+}
+
+func (o selectorBuilder) WithComponent(name string) selectorBuilder {
+	req, err := labels.NewRequirement(componentLabel, selection.Equals, []string{name})
+	if err != nil {
+		panic(err)
```
valaparthvi, line 21: Is it okay to panic ?

valaparthvi (approved):

pkg/labels/selector_builder.go
feloy: I think it is. The only error that can happen is when the number of values in the last param's array does not match with the operation in the second argument. Here it will never fail because we need one value with the `Eq` operation, and we want to fail fast if a developer changes the operation or the last param in a wrong way.

feloy:
> @feloy since there is no issue linked/related to this PR, how should a reviewer go about checking what changes to expect as a result of merging this PR? Is this PR changing the labels applied to the Kubernetes resources created by odo, or is it changing the way we create the labels that are eventually applied to the Kubernetes resources created by odo?

This is only a code refactoring, the resulting labels in the resources and the selectors used for filtering should be the same before and after this PR

dharmit:
> This is only a code refactoring, the resulting labels in the resources and the selectors used for filtering should be the same before and after this PR

👍🏾 

/approve

feloy:

```
	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
	|                                                                   |
	 -------------------------------------------------------------------
 devfile timed out. path: ./tests/integration/devfile
```

Tests pass on IBM Cloud

PR #5623, opened 2022-04-05
Set list command in main help section

feloy:
**What type of PR is this:**

/kind bug

This PR moves the `list` command into the `main` help section, so it is displayed when running `odo help`


dharmit:
Small fix

/approve
/lgtm

PR #5629, opened 2022-04-05
Adds warning for running podman on Apple Silicon

cdrage:
**What type of PR is this:**

/kind feature

Warns the user if we are trying to run Podman on Mac M1. As we are
unable to build x86 images natively unless using the workaround.

Fixes https://github.com/redhat-developer/odo/issues/5597

**PR acceptance criteria:**

- [X] Unit test

- [X] Integration test

- [X] Documentation

Run it on a M1 mac :)

### DIFF
--- a/pkg/devfile/image/image.go
+++ b/pkg/devfile/image/image.go
@@ -95,6 +95,21 @@ func selectBackend() (Backend, error) {
 		podmanCmd = "podman"
 	}
 	if _, err := lookPathCmd(podmanCmd); err == nil {
+
+		// Podman does NOT build x86 images on Apple Silicon / M1 and we must *WARN* the user that this will not work.
+		// There is a temporary workaround in order to build x86 images on Apple Silicon / M1 by running the following commands:
+		// podman machine ssh sudo rpm-ostree install qemu-user-static
+		// podman machine ssh sudo systemctl reboot
+		//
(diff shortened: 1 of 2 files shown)


pkg/devfile/image/image.go
```diff
@@ -95,6 +95,21 @@ func selectBackend() (Backend, error) {
...
+		// podman machine ssh sudo systemctl reboot
+		//
+		// The problem is that Fedora CoreOS does not have qemu-user-static installed by default,
+		// and the workaround is to install it manually as the dependencies need to be integrated into the Fedora ecosystem
+		// The open discussion is here: https://github.com/containers/podman/discussions/12899
+		//
+		// TODO: Remove this warning when Podman natively supports x86 images on Apple Silicon / M1.
+		if log.IsAppleSilicon() {
+			log.Warning("WARNING: Building images on Apple Silicon / M1 is not (yet) supported natively on Podman")
+			log.Warning("There is however a temporary workaround: https://github.com/containers/podman/discussions/12899")
```
valaparthvi, line 111: Can you also add a message asking the user to use Docker instead? or perhaps use Docker as a fallback option?

cdrage: Ideally I'd like to not suggest to the user `Docker` as it's not free for business use (which are the people we are targeting...) and to instead try to use Podman with the workaround.

PR #5633, opened 2022-04-06
Tests cleanup after odo dev terminates

### LINKED ISSUE #5621: followup on "odo dev part 2 (cleanup) #5485"
- [ ] at least one test with devfile that creates storage, k8s. networking/Service (endpoint), k8s component. 
- [ ] verify that this work on Windows

feloy:
**What type of PR is this:**

/kind tests

Fixes partially #5621 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/integration/devfile/cmd_dev_test.go
+++ b/tests/integration/devfile/cmd_dev_test.go
@@ -240,12 +240,12 @@ var _ = Describe("odo dev command tests", func() {
 
 		When("odo dev is executed and Ephemeral is set to false", func() {
 
+			var devSession helper.DevSession
 			BeforeEach(func() {
 				helper.Cmd("odo", "preference", "set", "Ephemeral", "false").ShouldPass()
-
-				devSession, _, _, _, err := helper.StartDevMode()
+				var err error
+				devSession, _, _, _, err = helper.StartDevMode()
 				Expect(err).ToNot(HaveOccurred())
-				defer devSession.Kill()
 				// An ENV file should have been created indicating current namespace
 				Expect(helper.VerifyFileExists(".odo/env/env.yaml")).To(BeTrue())
 				helper.FileShouldContainSubstring(".odo/env/env.yaml", "Project: "+commonVar.Project)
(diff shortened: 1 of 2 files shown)


rm3l (approved):

tests/integration/devfile/cmd_dev_test.go
```diff
@@ -240,17 +240,21 @@ var _ = Describe("odo dev command tests", func() {
...
+				devSession, _, _, _, err = helper.StartDevMode()
 				Expect(err).ToNot(HaveOccurred())
-				defer devSession.Kill()
 				// An ENV file should have been created indicating current namespace
 				Expect(helper.VerifyFileExists(".odo/env/env.yaml")).To(BeTrue())
 				helper.FileShouldContainSubstring(".odo/env/env.yaml", "Project: "+commonVar.Project)
 			})
 
+			AfterEach(func() {
+				devSession.Kill()
```
valaparthvi, line 255: session.Kill() will not cleanup the resources, will it? We'll need to run session.Stop() for this, no?

tests/integration/devfile/cmd_dev_test.go
```diff
@@ -275,8 +279,9 @@ var _ = Describe("odo dev command tests", func() {
 				})
 			})
 
-			When("running odo delete component --wait", func() {
+			When("killing odo dev and running odo delete component --wait", func() {
```
valaparthvi, line 282: Don't you think this test is more appropriate for cmd_delete_test.go?
Or, to keep it more inline with the following test, check instead that it does not delete the resources.?

tests/integration/devfile/cmd_dev_test.go
feloy: We don't particularly need to cleanup resources at this point (this will be done when deleting the namespace), but to be consistent with the other tests, let's use `Stop()`

tests/integration/devfile/cmd_dev_test.go
feloy: I find interesting to place it here as we can see the similarity with the test `when stopping odo dev normally`

tests/integration/devfile/cmd_dev_test.go
feloy: Reverting to `Kill` as `Stop` would need a `WaitEnd()` but it is not necessary, and adding some comment on the reason why killing the process is necessary

feloy:

IBM tests are passing

PR #5641, opened 2022-04-08
Create binding between devfile component and existing service

### LINKED ISSUE #5627: Create binding between devfile component and existing service
This issue covers the situation when `odo add binding` is executed in a directory with `devfile.yaml`.

#### Select the service instance you want to bind to
List of all objects (CRs) that are bindable. 
This list can be obtained by querying  `bindablekinds`.
`bindable-kinds` gives you just all resources (kinds) that are bindable.
To get the actual object (instances) you need to list objects of a given type

For example using kubectl it is 
```
for r in `kubectl get bindablekinds bindable-kinds -o go-template='{{range .status}}{{.kind}}.{{.group}}{{"\n"}}{{end}}'`; 
do
  kubectl get $r;
done

```

example of how this can be done in go using client-go is in https://github.com/kadel/odo-v3-mockup/blob/c5f12da1c54856b82c431f5173e0086544c015da/cmd/binding.go#L147-L197

#### What will be the ServiceBinding's name?
The name of a new ServiceBinding resource.
The default value should be automatically generated `<devfile name>-<service name>`

#### Bind as files:
Y/N question
sets `bindAsFiles` based on users input
default is Y

### Example:
[asciicast](https://asciinema.org/a/484558)

## non-interactive mode
- [ ] `--service` bindable service to bind to format is `<name>/<kind>.<apigroup>` or just `<name>`.  The command should be smart enough that the user can specify just the name of the object. If there are multiple bindable objects with the same name it errors out and tells user that he/she needs to include the kind and apigroup. Command should validate that the object exists, if not the error message should list what are available options (list all bindable object in a format that user can use with `--service` flag. 
- [ ] `--name` name of newly created service binding object
- [ ] `--bind-as-files` boolean, default is `true` (`--bind-as-files=false` to turn this off)

valaparthvi:

**What type of PR is this:**
/kind feature

Fixes #5627 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test [Ignore the IBM cluster tests for now, we're working on installing operators on it.]

- [ ] Documentation 

1.  `odo init --devfile nodejs --name my-node-app --starter nodejs-starter`
2. Run `./scripts/configure-cluster/common/setup-operators.sh` on your cluster.
3. Run `kubectl get csv -w` and wait until you see a result similar to the following:
```sh
➜  odo git:(main) ✗ k get csv           
NAME                              DISPLAY                    VERSION   REPLACES                          PHASE
cloud-native-postgresql.v1.15.0   Cloud Native PostgreSQL    1.15.0    cloud-native-postgresql.v1.14.0   Succeeded
service-binding-operator.v1.0.1   Service Binding Operator   1.0.1     service-binding-operator.v1.0.0   Succeeded
```
4. Run `kubectl apply -f https://raw.githubusercontent.com/valaparthvi/odo/create-binding/tests/examples/manifests/bindablekind-instance.yaml`
5. `odo add binding` to run in the interactive mode or `odo add binding --service cluster-sample --name my-binding` to run in the non-interactive mode.
6. Check the devfile to see if the Service Binding instance was added.
7. Run `odo dev` and check if the ServiceBinding instance is created successfully on the cluster.


valaparthvi:
@feloy Can you please review the skeleton for this command?

pkg/odo/cli/create/binding.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/spf13/cobra"
+	ktemplates "k8s.io/kubectl/pkg/util/templates"
+
+	"github.com/redhat-developer/odo/pkg/odo/cmdline"
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions/clientset"
+)
+
+// BindingRecommendedCommandName is the recommended binding sub-command name
+const BindingRecommendedCommandName = "binding"
```
dharmit, line 17: It's `RecommendedCommandName` pretty much everywhere in the code. Any reason for naming it differently in this file?

pkg/odo/cli/create/binding.go
```diff
@@ -0,0 +1,119 @@
...
+	o.flags = o.clientset.BindingClient.GetFlags(cmdline.GetFlags())
+
+	return nil
+}
+
+func (o *CreateBindingOptions) Validate() (err error) {
+	return o.clientset.BindingClient.Validate(o.flags)
+}
+
+func (o *CreateBindingOptions) Run(ctx context.Context) error {
```
dharmit, line 74: `ctx` is not used, maybe make it `_`?

dharmit (changes requested):
Reviewed the CLI layer.

kadel:
When SBO is not installed it errours out with confusing message. 

```
odo create binding                                                      
 ✗  bindablekinds.binding.operators.coreos.com "bindable-kinds" is forbidden: User "developer" cannot get resource "bindablekinds" in API group "binding.operators.coreos.com" at the cluster scope
```

The error should instruct the user that this functionality requires SBO and it needs to be installed in order to use this command .

kadel:
/hold

I haven't noticed any e2e tests or unit tests :-/

dharmit (changes requested):
That's an incomplete review of the business layer. Incomplete because I wasn't able to start a Redis Operator backed service on minikube. 😕

dharmit:
How should I create a binding in non-interactive mode? What am I doing wrong here?
```sh
# this doesn't work
$ odo create binding --service redis-standalone/Redis.redis.redis.opstreelab.in --name bind1
 ✗  "redis-standalone/Redis.redis.redis.opstreelab.in" service not found

# this works
$ odo create binding
? Select service instance you want to bind to: redis-standalone (redis.redis.opstreelabs.in/v1beta1, Kind=Redis)
? Enter the ServiceBinding's name: my-mynodejs-app
? Bind as files? Yes
```
Also, I think it would be helpful to print the equivalent non-interactive command at the end of a successful interactive session of creating binding. @kadel wdyt?

pkg/odo/cli/create/binding.go
```diff
@@ -0,0 +1,119 @@
+package create
```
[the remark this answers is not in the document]
valaparthvi: The current file structure does not allow using a different package. When we do add more things, we can move this file to a different folder.

pkg/odo/cli/create/binding.go
valaparthvi: Because there's already a constant with that name in the package.

kadel:
> How should I create a binding in non-interactive mode? What am I doing wrong here?
> 
> ```shell
> # this doesn't work
> $ odo create binding --service redis-standalone/Redis.redis.redis.opstreelab.in --name bind1
>  ✗  "redis-standalone/Redis.redis.redis.opstreelab.in" service not found
> 
> # this doesn't work either
> $ odo create binding --service redis-standalone/redis.redis.opstreelabs.in --name bind
>  ✗  "redis-standalone/redis.redis.opstreelabs.in" service not found
> 
> # doesn't work
> $ odo create binding --service redis.redis.opstreelabs.in/redis-standalone --name bind
>  ✗  "redis.redis.opstreelabs.in/redis-standalone" service not found
> 
> # doesn't work
> $ odo create binding --service Redis.redis.redis.opstreelab.in/redis-standalone --name bind 
>  ✗  "Redis.redis.redis.opstreelab.in/redis-standalone" service not found
> 
> # obviously I'm feeling dumb by now, so I go with interactive mode!
> 
> # this works
> $ odo create binding
> ? Select service instance you want to bind to: redis-standalone (redis.redis.opstreelabs.in/v1beta1, Kind=Redis)
> ? Enter the ServiceBinding's name: my-mynodejs-app
> ? Bind as files? Yes
> ```
> 
> Also, I think it would be helpful to print the equivalent non-interactive command at the end of a successful interactive session of creating binding. @kadel wdyt?

```
odo create binding --service redis-standalone.Redis.redis.redis.opstreelabs.in/v1beta1 --name bind
```
should work


also 
```
odo create binding --service redis-standalone --name bind
```

needs to work as well, if there is only one bindable resource called `redis-standalone`

pkg/odo/cli/create/binding.go
```diff
@@ -0,0 +1,119 @@
...
+
+# Create binding interactively
+%[1]s
+`)
+
+type CreateBindingOptions struct {
+	// name of the service to bind
+	name string
+	// service is name of the service to be bound to the component
+	service string
```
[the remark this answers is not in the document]
dharmit: But from the comments, it's not super clear. It's confusing even.

pkg/odo/cli/create/binding.go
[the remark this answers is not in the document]
dharmit: > The current file structure does not allow using a different package.

Why so?

dharmit:
> ```
> odo create binding --service redis-standalone.Redis.redis.redis.opstreelabs.in/v1beta1 --name bind
> ```
> 
> should work

Yikes. That's inception level deep! Dot inside a dot inside a dot (dream inside a dream inside a dream.) This is a UX nightmare, IMO.

> also
> 
> ```
> odo create binding --service redis-standalone --name bind
> ```
> 
> needs to work as well, if there is only one bindable resource called `redis-standalone`

+1. Also, interactive mode should print the equivalent non-interactive mode command so that the user can create a binding with a one-liner next time.

kadel:
> Yikes. That's inception level deep! Dot inside a dot inside a dot (dream inside a dream inside a dream.) This is a UX nightmare, IMO.
> 
> >

not much we can do, this is specificum of this Redis CRD. The API group of that CRD is `redis.redis.opstreelabs.in`, and Kind is `Redis` so this leads to those 3x `redis`

valaparthvi:
> `odo create binding --service redis-standalone.Redis.redis.redis.opstreelabs.in/v1beta1 --name bind`

This is not mentioned in the issue, should this also be included? I am currently working with <name>/<kind>.<apigroup> format.

What all formats do we want to support?
- [ ] <name>/<kind>.<apigroup>
- [ ] <name>.<kind>.<apigroup>
- [ ] <name>

pkg/odo/cli/create/binding.go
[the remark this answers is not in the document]
valaparthvi:
```sh
➜  tree pkg/odo/cli/create  
pkg/odo/cli/create
├── binding.go
└── create.go

0 directories, 2 files
```

2 files in the same directory cannot have different packages.

dharmit:
> The command should be `odo add binding` not `odo create binding`, can you please change that?

Shouldn't `pkg/odo/cli/create/binding.go` be moved to `pkg/odo/cli/add/binding.go`? It's still under `odo create` command group.

valaparthvi:
> > The command should be `odo add binding` not `odo create binding`, can you please change that?
> 
> Shouldn't `pkg/odo/cli/create/binding.go` be moved to `pkg/odo/cli/add/binding.go`? It's still under `odo create` command group.

It should. I am working on it.

valaparthvi:
>What will be the ServiceBinding's name?
>
>The name of a new ServiceBinding resource.
The default value should be automatically generated <devfile name>-<service name>

I've forgotten to add this. Working on it.

pkg/odo/cli/create/binding.go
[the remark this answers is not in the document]
dharmit: > 2 files in the same directory cannot have different packages.

I meant more along the lines of creating a directory called `binding` and having a `package binding` inside it. I do know that files within the same directory can't have different packages.

feloy (commented):
I have made some first tests of `odo add binding`. It seems to work correctly.

The only disturbing thing is that the command does not explain what it has done before exiting. It got me some time to understand the servicebinding resource has been added to the devfile, but the resource has not been created on the cluster

Also, (out of scope) I would love to have an ouput of the interactive command explaining me how to run the same command with parameters, for example:

```
You can automate this command by executing:
  odo add binding --service myservice.Redis.redis.redis.opstreelab.in --name myRedisService
```

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+	// We create the asker client and the backends here and not at the CLI level, as we want to hide these details to the CLI
+	askerClient := asker.NewSurveyAsker()
+	return &BindingClient{
+		flagsBackend:       backend.NewFlagsBackend(),
+		interactiveBackend: backend.NewInteractiveBackend(askerClient),
+		kubernetesClient:   kubernetesClient,
+	}
+}
+
+// GetFlags gets the flag specific to init operation so that it can correctly decide on the backend to be used
```
feloy, line 38:
```suggestion
// GetFlags gets the flag specific to add binding operation so that it can correctly decide on the backend to be used
```

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+// It ignores all the flags except the ones specific to init operation, for e.g. verbosity flag
```
feloy, line 39:
```suggestion
// It ignores all the flags except the ones specific to add binding operation, for e.g. verbosity flag
```

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+	if len(flags) == 0 {
+		backend = o.interactiveBackend
+	} else {
+		backend = o.flagsBackend
+	}
+	return backend.AskBindAsFiles(flags)
+}
+
+func (o *BindingClient) CreateBinding(serviceName string, bindingName string, bindAsFiles bool, obj parser.DevfileObj, serviceMap map[string]servicebinding.Ref, componentContext string) error {
+	// serviceName format is <name> (<kind>.<apigroup>)
```
feloy, line 93: Could we find a structure in https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/schema/group_version.go to pass this service name in a more structured way?

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	deploymentGVR, err := o.kubernetesClient.GetDeploymentAPIVersion()
+	if err != nil {
+		return err
+	}
+
+	serviceBinding := &servicebinding.ServiceBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: strings.Join([]string{kclient.ServiceBindingGroup, kclient.ServiceBindingVersion}, "/"),
```
feloy, line 124: Please use `func (gv GroupVersion) String() string` from `apimachinery/schema` package

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: bindingName,
+		},
+		Spec: servicebinding.ServiceBindingSpec{
+			DetectBindingResources: true,
+			BindAsFiles:            bindAsFiles,
+			Application: servicebinding.Application{
+				Ref: servicebinding.Ref{
+					Name:     fmt.Sprintf("%s-app", obj.GetMetadataName()),
```
feloy, line 135: I would prefer to see that name computed at a higher level, so you don't have to pass devfileObj at this level

pkg/binding/interface.go
```diff
@@ -0,0 +1,25 @@
...
+	// SelectServiceInstance returns the service to bind to the component
+	SelectServiceInstance(flags map[string]string, options []string, serviceMap map[string]servicebinding.Ref) (string, error)
+	// AskBindingName returns the name to be set for the binding
+	AskBindingName(serviceName, componentName string, flags map[string]string) (string, error)
+	// AskBindAsFiles asks if the service should be binded as files
+	AskBindAsFiles(flags map[string]string) (bool, error)
+
+	// CreateBinding adds the ServiceBinding manifest to the devfile
+	CreateBinding(service string, bindingName string, bindAsFiles bool, obj parser.DevfileObj, serviceMap map[string]servicebinding.Ref, componentContext string) error
+	// GetServiceInstances gets a list of all the bindable instance names, an error, and a map of bindable instance name with it's servicebinding.Ref;
```
feloy, line 22:
```suggestion
	// GetServiceInstances gets a list of all the bindable instance names, an error, and a map of bindable instance name with its servicebinding.Ref;
```

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+	yamlDesc, err := yaml.Marshal(serviceBindingUnstructured.UnstructuredContent())
+	if err != nil {
+		return err
+	}
+
+	err = libdevfile.AddKubernetesComponentToDevfile(string(yamlDesc), serviceBinding.Name, obj)
+	return err
+}
+
+func (o *BindingClient) GetServiceInstances() ([]string, map[string]servicebinding.Ref, error) {
```
feloy, line 160: It seems this function could be part of a `kclient/binding.go` file

pkg/kclient/operators.go
```diff
@@ -191,6 +220,23 @@ func (c *Client) GetRestMappingFromUnstructured(u unstructured.Unstructured) (*m
 	return mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
 }
 
+func (c *Client) GetRestMappingFromGVK(gvk schema.GroupVersionKind) (*meta.RESTMapping, error) {
```
feloy, line 223: You could factorize this function and GetRestMappingFromUnstructured which are very similar.

pkg/odo/cli/add/binding/binding.go
```diff
@@ -0,0 +1,129 @@
...
+		Use:     name,
+		Short:   "Add Binding",
+		Long:    "Add a binding between a service and the component in the devfile",
+		Args:    cobra.NoArgs,
+		Example: fmt.Sprintf(addBindingExample, fullName),
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(o, cmd, args)
+		},
+	}
+	bindingCmd.Flags().StringVar(&o.name, "name", "", "Name of the Binding to create")
```
feloy, line 123: Please use the `Flag` suffix to be consistent with other commands

feloy (changes requested):
A few comments, but that is a big piece of work :)

pkg/kclient/operators.go
```diff
@@ -78,6 +87,26 @@ func (c *Client) GetResourceSpecDefinition(group, version, kind string) (*spec.S
 	return getResourceSpecDefinitionFromSwagger(data, group, version, kind)
 }
 
+func (c *Client) GetBindableKinds() (sboApi.BindableKinds, error) {
```
feloy, line 90: The Binding specific code could be placed in a specific `binding.go` file

pkg/binding/backend/flags.go
```diff
@@ -0,0 +1,125 @@
...
+	return bindAsFiles, nil
+}
+
+// parseServiceName parses various service name formats. It supports the following formats:
+// - <name>
+// - <name>.<kind>
+// - <name>.<kind>.<apigroup>
+// - <name>/<kind>
+// - <name>/<kind>.<apigroup>
+func parseServiceName(service string) (name, kind, group string) {
```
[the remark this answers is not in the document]
valaparthvi: I intend to write unit tests in a separate PR, this PR is already long as is.

pkg/binding/binding.go
valaparthvi: I could, but I would still need the obj to add the binding to devfile, unless you think it would be better to move it to the CLI layer.

feloy: OK, so you could split this function in 2: the first that builds the serviceBinding (up to L144), and another one that adds it to the devfile. So we could do something else with the binding later (I'm confident we will want to do something with it even if we don't have a devfile).
You could keep these 2 functions in this package IMO

pkg/kclient/operators.go
valaparthvi: I have added a TODO for it, but again, this PR is long as so I'll take it up in a separate PR.

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+
+	serviceBindingUnstructured, err := kclient.ConvertK8sResourceToUnstructured(serviceBinding)
+	if err != nil {
+		return err
+	}
+
+	yamlDesc, err := yaml.Marshal(serviceBindingUnstructured.UnstructuredContent())
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
valaparthvi: Screenshot from 2022-05-06 20-12-42

Difference.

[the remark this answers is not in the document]
feloy: ok, thanks

valaparthvi:
> Also, (out of scope) I would love to have an ouput of the interactive command explaining me how to run the same command with parameters, for example:
> 
> ```
> You can automate this command by executing:
>   odo add binding --service myservice.Redis.redis.redis.opstreelab.in --name myRedisService
> ```

When do you suggest we print this? After asking all the questions or right before the command exits?

feloy:
> > Also, (out of scope) I would love to have an ouput of the interactive command explaining me how to run the same command with parameters, for example:
> > ```
> > You can automate this command by executing:
> >   odo add binding --service myservice.Redis.redis.redis.opstreelab.in --name myRedisService
> > ```
> 
> I like this idea. When do you suggest we print this? After asking all the questions or right before the command exits?

Just before the command exits, after all other output

pkg/binding/binding.go
valaparthvi: I have addressed this change by moving the servicebinding.Service and serviceBinding.ServiceBinding objects creation to kclient/binding.go, and keeping other things as is.

pkg/binding/binding.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (o *BindingClient) GetServiceInstances() ([]string, map[string]servicebinding.Ref, error) {
+	var bindableObjectMap = map[string]servicebinding.Ref{}
+	bindableKind, err := o.kubernetesClient.GetBindableKinds()
+	if err != nil {
+		return nil, bindableObjectMap, err
+	}
+	var bindableObjects []servicebinding.Ref
+
+	for _, bks := range bindableKind.Status {
```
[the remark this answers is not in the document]
valaparthvi: Moved parts of it to kclient/binding.go. Hopefully it is simpler than before. Divided into 3 parts:
1. kclient/binding.go: GetBindableKinds - this returns the BindableKinds/bindable-kinds object
2. kclient/binding.go: GetBindableKindStatusRestMapping this returns a list of restMapping objects of all the GVK from BindableKinds/bindable-kinds' Status list
3. binding/binding.go: GetServiceInstances - calls 1,and 2 in that order, loops throw the restMapping list, lists all the instances of a restMapping in the form of unstructured objects, creates a map with it and returns.

pkg/binding/binding.go
valaparthvi: https://github.com/redhat-developer/odo/pull/5641#discussion_r867064683

pkg/binding/backend/flags.go
```diff
@@ -0,0 +1,125 @@
...
+
+	err := dfutil.ValidateK8sResourceName("name", flags[FLAG_NAME])
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// SelectServiceInstance returns the service name in the form of '<name> (<kind>.<apigroup>)'
```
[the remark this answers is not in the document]
valaparthvi: Removed the options. serviceMap is now a map of service names (format: `<name> (<kind>.<group>)`) and it's unstructured object.

pkg/kclient/binding.go
```diff
@@ -0,0 +1,142 @@
...
+		return sboApi.BindableKinds{}, nil
+	}
+
+	var (
+		unstructuredBK *unstructured.Unstructured
+		bindableKind   sboApi.BindableKinds
+		err            error
+	)
+
+	gvr := schema.GroupVersionResource{Group: ServiceBindingGroup, Version: ServiceBindingVersion, Resource: BindableKindsResource}
```
feloy, line 42: You could define gvr as a package constant, and also gvk.

pkg/binding/binding.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+
+	serviceBinding := o.kubernetesClient.NewServiceBindingObject(bindingName, bindAsFiles, deploymentName, deploymentGVR, []sboApi.Mapping{}, []sboApi.Service{service})
+
+	// Note: we cannot directly marshal the serviceBinding object to yaml because it doesn't do that in the correct k8s manifest format
+	serviceBindingUnstructured, err := kclient.ConvertK8sResourceToUnstructured(serviceBinding)
+	if err != nil {
+		return err
+	}
```
feloy, line 108: You could use `ToUnstructuredAsGVK` defined [here](https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/converter/unstructured.go) so you don't need to build the `TypeMeta` in `NewServiceBindingObject`. And you could use the gvk of ServiceBinding defined in kclient/binding

pkg/kclient/binding.go
```diff
@@ -0,0 +1,142 @@
...
+	err = c.ConvertUnstructuredToResource(unstructuredBK.UnstructuredContent(), &bindableKind)
+	if err != nil {
+		return bindableKind, err
+	}
+	return bindableKind, nil
+}
+
+// GetBindableKindStatusRestMapping retuns a list of *meta.RESTMapping of all the bindable kind operator CRD
+func (c Client) GetBindableKindStatusRestMapping(bindableKindStatuses []sboApi.BindableKindsStatus) ([]*meta.RESTMapping, error) {
+	var bindableObjectRESTMappings []*meta.RESTMapping
```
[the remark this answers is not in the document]
valaparthvi: That is a good suggestion, thanks!

pkg/kclient/binding.go
valaparthvi: > `Const initializer 'schema.GroupVersionResource{ Group: "", Version: "", Kind: "", }' is not a constant`

Cannot do that. I could define a global variable, but I am not sure very sure about it, since it is vulnerable to change.

pkg/kclient/binding.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	ServiceBindingGroup    = "binding.operators.coreos.com"
+	ServiceBindingVersion  = "v1alpha1"
+	ServiceBindingKind     = "ServiceBinding"
+	ServiceBindingResource = "servicebindings"
+	BindableKindsResource  = "bindablekinds"
+)
```
feloy, line 21: You can use the constants defined in https://github.com/redhat-developer/service-binding-operator/blob/master/apis/binding/v1alpha1/groupversion_info.go

pkg/binding/binding.go
feloy: Or you can use the GroupVersionResource and GroupVersionKind defined in https://github.com/redhat-developer/service-binding-operator/blob/master/apis/binding/v1alpha1/groupversion_info.go

pkg/kclient/binding.go
feloy: Or you can use the value GroupVersionResource defined in https://github.com/redhat-developer/service-binding-operator/blob/master/apis/binding/v1alpha1/groupversion_info.go

pkg/binding/binding.go
valaparthvi: I am not very sure about this. The reason I added NewServiceBindingObject request to is to ensure I return a complete SB object. If I remove the `TypeMeta`, and if this method is called from some place, chances are it will not be called with `ToUnstructuredAsGVK`. Don't you think it will fail in the case where it's called with some other method?

feloy: Yes, the problem is that the Kind and APIVersion won't be added without `ToUnstructuredAsGVK`. Let's keep as this

feloy (approved):
I think it would be nice to have a clientset for the servicebinding resource, (but I canoot find one built into the servicebindingoperator repository), so we would not have to use Unstructured to work with serviceBindings.

With this clientset, we could run
```
c.sbClient.ServiceBindings().Get(context.TODO(), "bindable-kinds", metav1.GetOptions{})
```

instead of 
```
c.DynamicClient.Resource(sboApi.GroupVersion.WithResource(BindableKindsResource)).Get(context.TODO(), "bindable-kinds", v1.GetOptions{})
```

Let's keep this for a later refactoring

valaparthvi:
@kadel Can you give this a review?

kadel:
The error message for cases where there are no services available is not helpful 
```
▶ odo add binding
 ✗  please provide options to select from

```
This is when there is at least one bindablekind but it doesn't have any instances.


The error should be more descriptive, maybe something like `No bindable instances found` or something along those lines. The current message `please provide options to select from`  doesn't say much about what is happening

feloy (approved):

PR #5649, opened 2022-04-08
Bumping odo version to v2.5.1

mohammedzee1000:


rm3l:
@mohammedzee1000 Out of curiosity, how do you usually bump versions?
- Manually? If so, is it documented somewhere?
- Or I guess from the changes here that you ran the `bump-version.sh` script?

mohammedzee1000:
> @mohammedzee1000 Out of curiosity, how do you usually bump versions?
> 
> * Manually? If so, is it documented somewhere?
> * Or I guess from the changes here that you ran the [`bump-version.sh`](https://github.com/redhat-developer/odo/blob/v2/scripts/bump-version.sh) script?

Either works so long as all the correct files are bumped as socumented in https://github.com/redhat-developer/odo/wiki/Release:-New-version-of-odo

PR #5655, opened 2022-04-12
fix artifact collection

anandrkskd:

**What type of PR is this:**

/kind tests

This PR fix artifact collection for periodic job on v2

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/openshiftci-periodic-tests.sh
+++ b/scripts/openshiftci-periodic-tests.sh
@@ -64,6 +64,8 @@ if [ "$error" == "true" ]; then
     exit -1
 fi
 
+# copy generated artifacts/reports to artifacts dir
+set +e
 cp -r reports tests/reports $ARTIFACT_DIR
 
 oc logout


cdrage:
May have to rebase + push to get this to pass again.. SonarCloud seems to be stuck

scripts/openshiftci-periodic-tests.sh
```diff
@@ -64,6 +64,8 @@ if [ "$error" == "true" ]; then
     exit -1
 fi
 
+# copy generated artifacts/reports to artifacts dir
+set +e
```
dharmit, line 68: I'm struggling a bit to understand what's going on here and since there's no corresponding issue, I am asking my doubts here.

First off, I didn't know what `set +e` does, so I looked it up and found only references to `set -e`. But I understand now that if `set +e` is specified, a command that fails in a shell script wouldn't cause the script execution to end. It would let the next commands in the script to get executed.

So in this case, we have a `set -e` at the top of the file, which ensures that if any of the commands in the script fails the script execution is terminated. But before doing the `cp`, we are now setting `set +e` which, IIUC, wouldn't stop the execution of `oc logout` if the `cp -r` command failed.

However, if the `cp` command failed, that would still cause artifact collection to fail, no? Note that I don't know why artifact collection is failing in the first place or which artifact we are talking about here, since there's no issue - I'm kinda jumping onto a running train. 

Bear with me if I'm misunderstanding what's happening here, I'm just trying to understand and learn.

anandrkskd: > I'm kinda jumping onto a running train.

I can understand, as we don't have issue, it's confusing.
Currently, the flow of the test first checks if devfile are updated on the registry or not, if there is a change only then we run the devfile tests, otherwise we only run operator hub tests.
- devfile tests creates reports under  `$odo/reports` dir and if we don't run the tests these reports are not generated and hence the copy commands fails.
- You're right that even if we use set +e the command will fail, but it should not cause the prow tests to show it as failure.

dharmit: Makes a lot more sense now. Thanks for explaining @anandrkskd 👍🏾 

/lgtm

Issue #5524, opened 2022-03-07
odo registry


serenamarie125:
description to be updated by @kadel 

/kind user-story

## User Story
- As an odo user
- I want to be able to list the content of devfile registries that odo is using
- So that I know what stacks are available for me to use

## Acceptance Criteria
- [ ] `odo registry` should be the command for listing the content of devfile registries
- [ ] the output should be sorted alphabetically, if there are more stacks in the same name they should be ordered by the registry priority (registry that is configured first in odo preference registry has a priority)
- [ ] the description field should be shortened to some sensible length
- [ ] `--filter` flags allows filtering the output, it searches for matches in NAME, DESCRIPTION (not shortened) fields
- [ ] `--details` flag allows printing detailed information about devfile stacks, to see fields in detailed output see an example. Most of the information is from the metadata section in devfile. "Supported odo features" is based on the command group kinds defined in devfile. If there is at least one `group.kind: run` command `dev: true` if `group.kind: debug` then `debug: true` if `group.kind: deploy` `deploy:true`
- [ ] `--details` and `--filter` can be combined
- [ ] `--devfile` and `--devfile-registry` can be used point to the single devfile stack. This will be mainly use together with `--detail`,( see example) 

## Example
```
$ odo registry                                                                                                                       
Listing all Devfile stacks in registries configured in odo.
NAME                             REGISTRY                  DESCRIPTION                                                         
dotnet50                         DefaultDevfileRegistry    Stack with .NET 5.0                                                 
dotnet50                         StagingRegistry           Stack with .NET 5.0                                                 
dotnet60                         DefaultDevfileRegistry    Stack with .NET 6.0                                                 
dotnet60                         StagingRegistry           Stack with .NET 6.0                                                 
dotnetcore31                     DefaultDevfileRegistry    Stack with .NET Core 3.1                                            
dotnetcore31                     StagingRegistry           Stack with .NET Core 3.1                                            
go                               DefaultDevfileRegistry    Stack with the latest Go version                                    
go                               StagingRegistry           Stack with the latest Go version                                    
java-maven                       DefaultDevfileRegistry    Upstream Maven and OpenJDK 11                                       
java-maven                       StagingRegistry           Upstream Maven and OpenJDK 11                                       
java-openliberty                 DefaultDevfileRegistry    Java application Maven-built stack using the Open Liberty ru...     
java-openliberty                 StagingRegistry           Java application Maven-built stack using the Open Liberty ru...     
[...]

PR #5667, opened 2022-04-14
Adds odo registry command

cdrage:
**What type of PR is this:**

/kind feature

This adds the `odo registry` command, to output the details of what's
in the registry.

It also implements the flags:
- `--filter`
- `--devfile`
- `--details`

This can be tested by typing in the `odo registry` command after adding
a registry (you should already have one by default)

part of https://github.com/redhat-developer/odo/issues/5524

**PR acceptance criteria:**

- [X] Unit test

- [X] Integration test

- [X] Documentation

`odo registry`

### DIFF
--- a/pkg/odo/cli/registry/registry.go
+++ b/pkg/odo/cli/registry/registry.go
@@ -29,14 +29,17 @@ var Example = `  # Get all devfile components
 %[1]s --filter nodejs --devfile-registry DefaultDevfileRegistry
 
 # Show more details
-%[1]s --details`
+%[1]s --details
+
+# Show more details from a specific devfile and registry
+%[1]s --details --devfile nodejs --devfile-registry DefaultDevfileRegistry`
 
 // ListOptions encapsulates the options for the odo registry command
 type ListOptions struct {
 	clientset *clientset.Clientset
 
 	// List of known devfiles
-	DevfileList registry.DevfileStackList
+	devfileList registry.DevfileStackList
 
 	// Flags
 	filterFlag   string
@@ -57,12 +60,12 @@ func (o *ListOptions) SetClientset(clientset *clientset.Clientset) {
 // Complete completes ListOptions after they've been created
 func (o *ListOptions) Complete(cmdline cmdline.Cmdline, args []string) (err error) {
 
-	o.DevfileList, err = o.clientset.RegistryClient.ListDevfileStacks("")
+	o.devfileList, err = o.clientset.RegistryClient.ListDevfileStacks("")
 	if err != nil {
 		return err
 	}
 
-	if o.DevfileList.DevfileRegistries == nil {
+	if o.devfileList.DevfileRegistries == nil {
 		log.Warning("Please run 'odo preference registry add <registry name> <registry URL>' to add registry for listing devfile components\n")
 	}
 
@@ -71,7 +74,7 @@ func (o *ListOptions) Complete(cmdline cmdline.Cmdline, args []string) (err erro
 
 // Validate validates the ListOptions based on completed values
 func (o *ListOptions) Validate() error {
-	if len(o.DevfileList.Items) == 0 {
+	if len(o.devfileList.Items) == 0 {
 		return fmt.Errorf("no deployable components found")
 	}
 	return nil
@@ -79,7 +82,7 @@ func (o *ListOptions) Validate() error {
 
 // Run contains the logic for the command associated with ListOptions
 func (o *ListOptions) Run(ctx context.Context) (err error) {
-	o.printDevfileList(o.DevfileList.Items)
+	o.printDevfileList(o.devfileList.Items)
 	return nil
 }
 
@@ -140,10 +143,6 @@ func (o *ListOptions) printDevfileList(DevfileList []registry.DevfileStack) {
 	t.SetOutputMirror(log.GetStdout())
 
 	t.AppendHeader(table.Row{"NAME", "REGISTRY", "DESCRIPTION"})
-	t.SortBy([]table.SortBy{
-		{Name: "NAME", Mode: table.Asc},
-		{Name: "REGISTRY", Mode: table.Asc},
-	})
 
 	devfiles := []registry.DevfileStack{}
 	// Filter through all the devfile components per the filters / parameters passed in.
--- a/pkg/odo/cli/registry/registry.go
+++ b/pkg/odo/cli/registry/registry.go
@@ -1,0 +1,211 @@
+package registry
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/redhat-developer/odo/pkg/log"
+	"github.com/redhat-developer/odo/pkg/odo/cmdline"
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
+	"github.com/redhat-developer/odo/pkg/odo/genericclioptions/clientset"
+	odoutil "github.com/redhat-developer/odo/pkg/odo/util"
+	"github.com/redhat-developer/odo/pkg/registry"
+	"github.com/redhat-developer/odo/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+const RecommendedCommandName = "registry"
+
+var Example = `  # Get all devfile components
+  %[1]s
+
+# Filter by name
+%[1]s --filter nodejs
+
+# Show more details
+%[1]s --details`
+
+// ListOptions encapsulates the options for the odo registry command
+type ListOptions struct {
+	clientset *clientset.Clientset
+
+	// List of known devfiles
+	DevfileList registry.DevfileStackList
+
+	// Flags
+	filterFlag          string
+	devfileFlag         string
+	devfileRegistryFlag string
+	detailsFlag         bool
+}
+
+// NewListOptions creates a new ListOptions instance
+func NewListOptions() *ListOptions {
+	return &ListOptions{}
+}
+
+func (o *ListOptions) SetClientset(clientset *clientset.Clientset) {
+	o.clientset = clientset
+}
+
+// Complete completes ListOptions after they've been created
+func (o *ListOptions) Complete(cmdline cmdline.Cmdline, args []string) (err error) {
+
+	o.DevfileList, err = o.clientset.RegistryClient.ListDevfileStacks("")
+	if err != nil {
+		return err
+	}
+
+	if o.DevfileList.DevfileRegistries == nil {
+		log.Warning("Please run 'odo preference registry add <registry name> <registry URL>' to add registry for listing devfile components\n")
+	}
+
+	return nil
+}
+
+// Validate validates the ListOptions based on completed values
+func (o *ListOptions) Validate() error {
+	if len(o.DevfileList.Items) == 0 {
+		return fmt.Errorf("no deployable components found")
+	}
+	return nil
+}
+
+// Run contains the logic for the command associated with ListOptions
+func (o *ListOptions) Run(ctx context.Context) (err error) {
+	o.printDevfileList(o.DevfileList.Items)
+	return nil
+}
+
+func NewCmdRegistry(name, fullName string) *cobra.Command {
+	o := NewListOptions()
+
+	var listCmd = &cobra.Command{
+		Use:     name,
+		Short:   "List all components from the Devfile registry",
+		Long:    "List all components from the Devfile registry",
+		Example: fmt.Sprintf(Example, fullName),
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(o, cmd, args)
+		},
+	}
+
+	clientset.Add(listCmd, clientset.REGISTRY)
+
+	// Flags
+	listCmd.Flags().StringVar(&o.filterFlag, "filter", "", "Filter based on the name of the component")
+	listCmd.Flags().StringVar(&o.devfileFlag, "devfile", "", "Only the specific Devfile component")
+	listCmd.Flags().StringVar(&o.devfileRegistryFlag, "devfile-registry", "", "Only show components from the specific Devfile registry")
+	listCmd.Flags().BoolVar(&o.detailsFlag, "details", false, "Show details of each component")
+
+	// Add a defined annotation in order to appear in the help menu
+	listCmd.Annotations["command"] = "main"
+	listCmd.SetUsageTemplate(odoutil.CmdUsageTemplate)
+
+	return listCmd
+}
+
+func (o *ListOptions) printDevfileList(DevfileList []registry.DevfileStack) {
+
+	// Create the table and use our own style
+	t := table.NewWriter()
+
+	t.SetStyle(table.Style{
+		Box: table.BoxStyle{
+			PaddingLeft:  " ",
+			PaddingRight: " ",
+		},
+		Color: table.ColorOptions{
+			Header: text.Colors{text.FgHiGreen, text.Underline},
+		},
+		Format: table.FormatOptions{
+			Footer: text.FormatUpper,
+			Header: text.FormatUpper,
+			Row:    text.FormatDefault,
+		},
+		Options: table.Options{
+			DrawBorder:      false,
+			SeparateColumns: false,
+			SeparateFooter:  false,
+			SeparateHeader:  false,
+			SeparateRows:    false,
+		},
+	})
+	t.SetOutputMirror(log.GetStdout())
+
+	t.AppendHeader(table.Row{"NAME", "REGISTRY", "DESCRIPTION"})
+	t.SortBy([]table.SortBy{
+		{Name: "NAME", Mode: table.Asc},
+		{Name: "REGISTRY", Mode: table.Asc},
+	})
+
+	for _, devfileComponent := range DevfileList {
+
+		// Mark the name as yellow in the index to it's easier to see.
+		name := text.Colors{text.FgHiYellow}.Sprint(devfileComponent.Name)
+
+		// If the user has specified a filter with variable o.filterFlag, then only show the components
+		// containing that specific string.
+		if o.filterFlag != "" {
+			if !strings.Contains(devfileComponent.Name, o.filterFlag) {
+				continue
+			}
+		}
+
+		// If the user passed in --devfile-registry <REGISTRY-NAME>, then only show the components from that Devfile stack
+		if o.devfileRegistryFlag != "" {
+			if !strings.Contains(devfileComponent.Registry.Name, o.devfileRegistryFlag) {
+				continue
+			}
+		}
+
+		// If the user passed in --devfile <NAME> only show that specific component matching that name
+		if o.devfileFlag != "" {
+			if devfileComponent.Name != o.devfileFlag {
+				continue
+			}
+		}
+
+		if o.detailsFlag {
+
+			// Output the details of the component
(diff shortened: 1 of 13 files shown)


pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,213 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			genericclioptions.GenericRun(o, cmd, args)
+		},
+	}
+
+	clientset.Add(listCmd, clientset.REGISTRY)
+
+	// Flags
+	listCmd.Flags().StringVar(&o.filterFlag, "filter", "", "Filter based on the name of the component")
+	listCmd.Flags().StringVar(&o.devfileFlag, "devfile", "", "Only show components from the specific Devfile stack")
```
dharmit, line 99: Playing with this, I typed individual Devfile stack name like `python` and `python-django`, but that failed. Passed the name of the registry and it worked. So, IMO, this should be reworded to below.:
```suggestion
	listCmd.Flags().StringVar(&o.devfileFlag, "devfile", "", "Only show components from the specific Devfile registry")
```

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,213 @@
...
+	return listCmd
+}
+
+func (o *ListOptions) printDevfileList(DevfileList []registry.DevfileStack) {
+
+	// Create the table and use our own style
+	t := table.NewWriter()
+	rowConfigAutoMerge := table.RowConfig{AutoMerge: true}
+
+	// If passed int he details flag, we'll be using a different style of table output.
```
dharmit, line 115:
```suggestion
	// If passed in the details flag, we'll be using a different style of table output.
```

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,213 @@
...
+				SeparateHeader:  false,
+				SeparateRows:    false,
+			},
+		})
+	}
+	t.SetOutputMirror(log.GetStdout())
+
+	// If we pass in the details flag, we will manipulate the table as well as create some
+	if o.detailsFlag {
+		t.AppendHeader(table.Row{"NAME / VERSION", "REGISTRY / URL", "LANGUAGE / TAGS", "DEV", "DEPLOY", "DEBUG", "DESCRIPTION", "STARTER PROJ."}, rowConfigAutoMerge)
```
dharmit, line 149: IMO, keeping the `REGISTRY/URL` column a bit later might be better. Saying this because in a real registry, there are going to be a bunch of stacks and the registry info might not even be immediately useful to users not working with multiple registries. 

In the screenshot, you can see that I have only one registry and that is occupying a good amount of space which causes info about starter projects difficult to get.
Screenshot from 2022-04-18 15-45-47

dharmit (changes requested):

kadel:
That tabular output doesn't work at all :-/ 
image

kadel:
> Fixes #5524
`--devfile` flag doesn't seems to work, and there is no `--devfile-registry` flag.

cdrage:
> That tabular output doesn't work at all :-/ image

Argh! That's unfortunate. I'll play around and probably switch to the "list" output, as much as the tabbed output looks better.

cdrage:
/hold

Have to wait for https://github.com/redhat-developer/odo/pull/5664 to be merged in so we have support for detecting if a Devfile is using DEBUG / DEPLOY / DEV commands.

cdrage:
Updated PR to now list components with `--details` as a list.

I've also added `--devfile` and `--devfile-registry` so you can specify 1 specific Devfile, as well as what registry you wish to query.

```sh
Name: python-django
Display Name: Django
Registry: DefaultDevfileRegistry
Registry URL: https://registry.devfile.io
Version: 1.0.0
Description: Python3.7 with Django
Tags: Python, pip, Django
Project Type: django
Language: python
Starter Projects:
  - django-example
Supported odo features:
  - dev: Y
  - deploy: Y
  - debug: Y
```

cdrage:
Example:

```sh
go/src/odo  add-registry-command ✗                                                                                                                                                                                                                                                                                                                                28m ⚑
▶ odo registry --devfile-registry Test --devfile python
 NAME    REGISTRY  DESCRIPTION
 python  Test      Python Stack with Python 3.7
 ```

File: pkg/odo/cli/registry/registry.go [hunk 4]
```diff
@@ -0,0 +1,214 @@
...
+				devfileComponent.DisplayName,
+				devfileComponent.Registry.Name,
+				devfileComponent.Registry.URL,
+				devfileComponent.Version,
+				devfileComponent.Description,
+				strings.Join(devfileComponent.Tags[:], ", "),
+				devfileComponent.ProjectType,
+				devfileComponent.Language,
+				strings.Join(devfileComponent.StarterProjects, "\n  - "),
+				// TODO, showing dev / deploy / debug NOT yet implemented
```
kadel, line 200: If there are things that are not implemented yet, and you plan to add it in this PR, please mark PR as work in progress.

kadel:
I haven't noticed any e2e or unit tests :-/

cdrage:
Unit and integration tests have now been added 👍

cdrage:
Ready for review!

pkg/odo/cli/registry/registry.go
dharmit: @cdrage this part is still commented. Is it going to be handled in this PR or a subsequent one?

kadel:
if I specify non-existing`--devfile` it silently exits with no error :/ even the exit code is `0`

```
▶ odo registry --details --devfile non

▶ echo $?
0
```

kadel:
`--filter` doesn't seem to search in Description.

```
▶ odo registry --filter OpenJDK
 NAME  REGISTRY  DESCRIPTION

▶ odo registry | grep  OpenJDK
 java-maven                    DefaultDevfileRegistry  Upstream Maven and OpenJDK 11
 java-maven                    StageRegistry           Upstream Maven and OpenJDK 11
```

cdrage:
Hey @kadel !

Thanks for the review, I've updated it so it now exists with an error if there is nothing found, as well as updated the `--filter ` so that it includes searching through the description:

```sh
go/src/odo  add-registry-command ✗                                                                                                                                                                                                                                                                                                                               1[14/152]
▶ odo registry --filter kalsjdfklasjdf
 ✗  There are no Devfiles available to show

go/src/odo  add-registry-command ✗                                                                                                                                                                                                                                                                                                                              1d0h ⚑  ⍉
▶ odo registry --filter Java application --devfile-registry Test
 NAME                          REGISTRY  DESCRIPTION
 java-openliberty              Test      Java application Maven-built stack using...
 java-openliberty-gradle       Test      Java application Gradle-built stack usin...
 java-quarkus                  Test      Quarkus with Java
 java-springboot               Test      Spring Boot® using Java
 java-vertx                    Test      Upstream Vert.x using Java
 java-websphereliberty         Test      Java application Maven-built stack using...
 java-websphereliberty-gradle  Test      Java application Gradle-built stack usin...
 java-wildfly-bootable-jar     Test      Java stack with WildFly in bootable Jar ...
```

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+			t.AppendRow(table.Row{name, devfileComponent.Registry.Name, util.TruncateString(devfileComponent.Description, 40, "...")})
+		}
+
+	}
+
+	// Exit with an error if there are no components to show, so we don't render the table / continue
+	if len(devfiles) == 0 {
+		log.Error("There are no Devfiles available to show")
+		return
+	}
```
valaparthvi, line 219: Can you move this before the for loop? I know for loop is not going to be bothered by 0 length, but it just makes sense to me to check the length and return early on if required.

pkg/registry/registry_test.go
```diff
@@ -87,6 +88,153 @@ OdoSettings:
...
+				// So we simply just say 'nil'
+				DevfileRegistries: nil,
+				Items:             nil,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			prefClient, _ := preference.NewClient()
```
valaparthvi, line 222: Can you use Mock client instead? See https://github.com/redhat-developer/odo/blob/main/pkg/init/init_test.go#L65 for reference.

valaparthvi (changes requested):
Everything worked great for me, but I just have a few comments.

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (o *ListOptions) printDevfileList(DevfileList []registry.DevfileStack) {
+
+	// Create the table and use our own style
+	t := table.NewWriter()
+
+	t.SetStyle(table.Style{
+		Box: table.BoxStyle{
+			PaddingLeft:  " ",
+			PaddingRight: " ",
```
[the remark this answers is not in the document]
cdrage: Unfortunately I've tried this with more padding, but found that it would "wrap" the terminal if there's not enough space :( so we'll have to keep it as small as possible.

pkg/odo/cli/registry/registry.go
cdrage: I put it after the for loop because the for loop ADDS to `devfiles` variable.

pkg/registry/registry_test.go
cdrage: Didn't need to since we're not modifying the clients (clients are both "blank"), so it was cleaner to implement it this way, instead of via fields.

What do you think @feloy ?

feloy: By mocking ths Preference client, you could mock the methods on this client the tested function is using only, without needing to create the underlying file. This way, you should not rely on the implementation of the preference package

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+  %[1]s
+
+# Filter by name
+%[1]s --filter nodejs
+
+# Filter by name and devfile registry
+%[1]s --filter nodejs --devfile-registry DefaultDevfileRegistry
+
+# Show more details
+%[1]s --details`
```
feloy, line 32: Could you add the example with  `--details --devfile <name> --devfile-registry <registry>` to get the details of a named devfile?

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+
+// ListOptions encapsulates the options for the odo registry command
+type ListOptions struct {
+	clientset *clientset.Clientset
+
+	// List of known devfiles
+	DevfileList registry.DevfileStackList
```
feloy, line 39: You can make it non visible (`devfileList`)

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+			SeparateColumns: false,
+			SeparateFooter:  false,
+			SeparateHeader:  false,
+			SeparateRows:    false,
+		},
+	})
+	t.SetOutputMirror(log.GetStdout())
+
+	t.AppendHeader(table.Row{"NAME", "REGISTRY", "DESCRIPTION"})
+	t.SortBy([]table.SortBy{
```
feloy, line 143: I think we want to order by registry in order of appareance in the config (the last in the list appearing first):

```
$ cat ~/.odo/preference.yaml
kind: Preference
apiversion: odo.dev/v1alpha1
OdoSettings:
  RegistryList:
    secure: false
  - Name: MyRegistry
    URL: https://registry.stage.devfile.io
    secure: false
  Ephemeral: false
  ConsentTelemetry: true


$ odo registry
 NAME                          REGISTRY                DESCRIPTION                                 
 dotnet50                      DefaultDevfileRegistry  Stack with .NET 5.0                         
 dotnet50                      MyRegistry              Stack with .NET 5.0                         
 dotnet60                      DefaultDevfileRegistry  Stack with .NET 6.0                         
 dotnet60                      MyRegistry              Stack with .NET 6.0                         
 dotnetcore31                  DefaultDevfileRegistry  Stack with .NET Core 3.1                    
 dotnetcore31                  MyRegistry              Stack with .NET Core 3.1                    
```

should be

```
$ odo registry
 NAME                          REGISTRY                DESCRIPTION                                 
 dotnet50                      MyRegistry              Stack with .NET 5.0                         
 dotnet50                      DefaultDevfileRegistry  Stack with .NET 5.0                         
 dotnet60                      MyRegistry              Stack with .NET 6.0                         
 dotnet60                      DefaultDevfileRegistry  Stack with .NET 6.0                         
 dotnetcore31                  MyRegistry              Stack with .NET Core 3.1                    
 dotnetcore31                  DefaultDevfileRegistry  Stack with .NET Core 3.1                    
```

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+		{Name: "REGISTRY", Mode: table.Asc},
+	})
+
+	devfiles := []registry.DevfileStack{}
+	// Filter through all the devfile components per the filters / parameters passed in.
+	for _, devfileComponent := range DevfileList {
+
+		// If the user has specified a filter with variable o.filterFlag, then only show the components
+		// containing that specific string.
+		if o.filterFlag != "" {
```
feloy, line 154: Filtering should happen in business level IMO

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+			if !strings.Contains(devfileComponent.Name, o.filterFlag) && !strings.Contains(devfileComponent.Description, o.filterFlag) {
+				continue
+			}
+		}
+
+		// If the user passed in --devfile-registry <REGISTRY-NAME>, then only show the components from that Devfile stack
+		if o.registryFlag != "" {
```
feloy, line 161: idem as filtering

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,226 @@
...
+			if !strings.Contains(devfileComponent.Registry.Name, o.registryFlag) {
+				continue
+			}
+		}
+
+		// If the user passed in --devfile <NAME> only show that specific component matching that name
+		if o.devfileFlag != "" {
```
feloy, line 168: idem as filtering

feloy (changes requested):

feloy:
> Well.. that's unfortunate. With the registry schema, we can't get the commands used within the devfile, and in order to implement the "Odo supported features:" part of odo registry, we have to make an upstream change.
> 
> @feloy implementation is based upon parsing devfile.yaml / object.
> 
> This means that in order to show it within odo registry would have to download each devfile.yaml, parse it, and then show it...
> 
> We can either:
> 
> 1. Implement odo registry and open up an issue upstream (I've opened it up here: [Implement a commands variable in the registry schema for easier group ID retrieval. devfile/api#831](https://github.com/devfile/api/issues/831)). Merge this PR as-is without showing "Odo supported features"
> 2. Download and parse each devfile.yaml whenever you run --details for each devfile component listed in order to show "Odo supported features".

I think we discussed and selected the second solution: Download and parse each devfile.yaml whenever you run --details for each devfile component listed in order to show "Odo supported features".

cdrage:
> > Well.. that's unfortunate. With the registry schema, we can't get the commands used within the devfile, and in order to implement the "Odo supported features:" part of odo registry, we have to make an upstream change.
> > @feloy implementation is based upon parsing devfile.yaml / object.
> > This means that in order to show it within odo registry would have to download each devfile.yaml, parse it, and then show it...
> > We can either:
> > 
> > 1. Implement odo registry and open up an issue upstream (I've opened it up here: [Implement a commands variable in the registry schema for easier group ID retrieval. devfile/api#831](https://github.com/devfile/api/issues/831)). Merge this PR as-is without showing "Odo supported features"
> > 2. Download and parse each devfile.yaml whenever you run --details for each devfile component listed in order to show "Odo supported features".
> 
> I think we discussed and selected the second solution: Download and parse each devfile.yaml whenever you run --details for each devfile component listed in order to show "Odo supported features".

Okay, that'll probably add a few days to the implementation then. I've opened up the upstream feature request anyways, but i'll implement it in this PR.

pkg/registry/registry_test.go
cdrage: @feloy Unfortunately https://github.com/redhat-developer/odo/blob/main/pkg/registry/registry.go#L175 has way too many HTTP requests and because it's not part of the preference or registry client, I'd be unable to mock it without refactoring that side of code (which is outside the scope of this PR).

Could we keep these tests (which are based on the `odo dev` unit tests) and revisit it at a later date to refactor?

At the moment, they pass, they work well.

feloy: @cdrage I don't think it is related. You can see on this commit how to mock the preference client to avoid using a preference file:
https://github.com/feloy/odo/commit/6551ebe2d74281b87b0a4b50f76ba78168da1713

feloy (changes requested):

pkg/registry/registry_test.go
cdrage: Ohhhh! I thought you were asking me to also somehow mock the http server... which would of been a lot.

I'll add the preference mock

cdrage:
I/O timeout

Retesting

/retest

cdrage:
Tests added! Ready for a review 👍

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+			if devfileComponent.Name != o.devfileFlag {
+				continue
+			}
+		}
+		devfiles = append(devfiles, devfileComponent)
+	}
```
feloy, line 167: I think you can remove this code now the filtering is done elsewhere

pkg/odo/cli/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+
+	for _, devfileComponent := range devfiles {
+		// Mark the name as yellow in the index so it's easier to see.
+		name := text.Colors{text.FgHiYellow}.Sprint(devfileComponent.Name)
+
+		if o.detailsFlag {
+
+			// If details are being shown, we must download the Devfile from the registry
+			// parse it, in order to retrieve command information such as if debug, dev, deploy and test is enabled.
+			devfileData, err := o.clientset.RegistryClient.RetrieveDevfileDataFromRegistry(devfileComponent.Registry.Name, devfileComponent.Name)
```
feloy, line 177: We will also need this data for the JSON output, please move this code to the business layer

feloy (changes requested):

pkg/odo/cli/registry/registry.go
cdrage: Have to keep it in this layer.. because this is only triggered with `--details`. If we had this in the business layer, we would http get every single devfile which is what we don't want.

Or I could add it as a boolean of true/false for the function if we require details or not? What do you think?

feloy: Yes, you could pass an option to the business layer function to indicate if you want or not the details

cdrage:
@feloy I've gone ahead and moved that functionality down to the business layer as well as improved tests.

cdrage:
@feloy Updated the code to add the sorting logic 👍

feloy:
@cdrage the OpenShift test is failing on one test. Could you check if your PR is completely rebased on main, and re-push the latest commit to restart the tests?

cdrage:
> @cdrage the OpenShift test is failing on one test. Could you check if your PR is completely rebased on main, and re-push the latest commit to restart the tests?

Yup, repushing now

PR #5687, opened 2022-04-22
Add additional checks to golangci-lint (#5567)

### LINKED ISSUE #5567: add additional checks to golangci-lint
/kind task

configure golangci-lint (https://github.com/redhat-developer/odo/blob/main/.golangci.yaml) to perform two following additional checks:

##  check 1

This follows a discussion in #5557, related to the removal of the deprecated `github.com/pkg/errors` package.

> To prevent any future use of this deprecated package, I was wondering if we shouldn't have a check that makes sure this package is not introduced again?
> 
> I don't know to what extent this is do-able with what we already have at this time, but this could be a check in CI, a Git hook, a custom linter, etc., which would simply grep for such occurrence..

_Originally posted by @rm3l in https://github.com/redhat-developer/odo/pull/5557#discussion_r828428625_

## check 2

/kind documentation

## What mistake did you find / what is missing in the documentation?

See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

All error messages should not be capitalize or end with punctuation. 

Unless it's the most *top* level error message. 

## What is the relevance of it?

A lot of our error output is inconsistent throughout `odo`

https://github.com/redhat-developer/odo/issues/5611

- [x] fix all errors, to make check pass

rm3l:
**What type of PR is this:**
/kind task

This configures [golangci-lint](https://golangci-lint.run/) with the following additional checks:
- [`depguard`](https://golangci-lint.run/usage/linters/#depguard), to check whether packages imported are allowed or denied. For example, this allows to check that the deprecated `github.com/pkg/errors` is not introduced again.
- [`revive`](https://golangci-lint.run/usage/linters/#revive) (successor of the deprecated golint), configured with the following rules:
  - [`error-strings`](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#error-strings), to check common conventions around error messages. This PR also fixes errors reported by this rule.

Additional checks might be added later on as needed.

Fixes #5567 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

`make validate` and all tests should pass.

### DIFF
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -189,7 +189,9 @@ func CheckMachineReadableOutputCommand(cmd *cobra.Command) error {
 
 	// Check the valid output
 	if hasFlagChanged && outputFlag.Value.String() != "json" {
+		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
 		return fmt.Errorf("Please input a valid output format for -o, available format: json")
+		//revive:enable:error-strings
 	}
 
 	// Check that if -o json has been passed, that the command actually USES json.. if not, error out.
@@ -199,7 +201,9 @@ func CheckMachineReadableOutputCommand(cmd *cobra.Command) error {
 		_ = flag.Set("o", "")
 
 		// Return the error
+		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
 		return fmt.Errorf("Machine readable output is not yet implemented for this command")
+		//revive:enable:error-strings
 	}
 
 	// Before running anything, we will make sure that no verbose output is made
--- a/pkg/odo/genericclioptions/runnable.go
+++ b/pkg/odo/genericclioptions/runnable.go
@@ -3,6 +3,7 @@ package genericclioptions
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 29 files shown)


pkg/odo/genericclioptions/runnable.go
```diff
@@ -189,7 +189,9 @@ func CheckMachineReadableOutputCommand(cmd *cobra.Command) error {
 
 	// Check the valid output
 	if hasFlagChanged && outputFlag.Value.String() != "json" {
+		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
 		return fmt.Errorf("Please input a valid output format for -o, available format: json")
+		//revive:enable:error-strings
```
valaparthvi, line 194: I find it a little weird that you need to enable the lint after disabling it. Will adding the disabling comment inline not require enabling it again?

rm3l: I had the same feeling, but this is how `revive` recommends using comment directives to configure it: https://github.com/mgechev/revive#comment-directives.
If we don't enable it again, the `error-strings` rule will be disabled for the rest of the file.

rm3l: By the way, this makes me think that the `fmt.Errorf` here should be replaced by `errors.New` instead. I'll update it.

feloy (approved):

PR #5703, opened 2022-04-28
odo dev store information about currently forwarded ports

### LINKED ISSUE #5676: `odo dev` - should store information about currently forwarded ports.
## User Story
- As an odo developer
- I want to be able to get information about what ports are currently forwarded
- So that I can display that information to user

Currently, the only way to get information about currently forwarded ports is from `odo dev` output. 
This can be problematic, as there are situation where we want to display this information to users like in `odo describe component` output (https://github.com/redhat-developer/odo/issues/5661)

## Acceptance Criteria
- [ ] `odo dev` should store information about currently forwarded ports in `.odo/devstate.json`. If the file doesn't exist it should create it.
- [ ] see example for how the `devstate.json` structure should look like
- [ ] as part of the "shutdown" procedure `odo dev` should remove `forwardedPorts` field from `.odo/devstate.json`
- [ ] documentation

## example 
```
$ cat .odo/devstate.json
```
```json
{
    "forwardedPorts": [
        {
        "containerName": "runtime",
        "containerPort": 8080,
        "localPort": 40001,
        "localAddress": "127.0.0.1"
        }
    ]

}
```

/kind user-story

feloy:
**What type of PR is this:**

/kind feature

See issue.

The PR updates the image used to run tests on IBM Cloud to be able to run tests as non-root user, to make a new test relying on user rights pass.

Fixes #5676 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation

The values for the forwarded ports are obtained through the `PortWriter`, from the text written by the low level port forwarder. This is because we need to get the information of the local port when we are using random ports (with --random-ports flag).

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -231,8 +231,7 @@ components:
 
 When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/devstate.json`. 
 
-This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp 
-at which the state has been last saved, and the currently forwarded ports.
+This state file contains the currently forwarded ports.
 
 ```
 {
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -229,7 +229,7 @@ components:
 
 ### State file
 
-When the command `odo dev` is executed, the state of the command isa saved in the file `./.odo/state.json`. 
+When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/state.json`. 
 
 This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp 
 at which the state has been last saved, and the currently forwarded ports.
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -229,7 +229,7 @@ components:
 
 ### State file
 
-When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/state.json`. 
+When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/devstate.json`. 
 
 This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp 
 at which the state has been last saved, and the currently forwarded ports.
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -226,3 +226,25 @@ components:
     memoryLimit: 1024Mi
     mountSources: true
 ```
+
+### State file
+
+When the command `odo dev` is executed, the state of the command isa saved in the file `./.odo/state.json`. 
+
+This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp 
+at which the state has been last saved, and the currently forwarded ports.
+
+```
+{
+ "timestamp": 1651247240,
+ "pid": 154474,
+ "forwardedPorts": [
+  {
+   "containerName": "runtime",
+   "localAddress": "127.0.0.1",
+   "localPort": 40001,
+   "containerPort": 3000
+  }
+ ]
+}
+```
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -236,8 +236,6 @@ at which the state has been last saved, and the currently forwarded ports.
 
 ```
 {
- "timestamp": 1651247240,
- "pid": 154474,
  "forwardedPorts": [
   {
    "containerName": "runtime",
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -241,7 +241,7 @@ func (o *DevOptions) Run(ctx context.Context) (err error) {
 	}()
 
 	portsBuf.Wait()
-	err = o.clientset.StateClient.SetForwardedPorts(portsBuf.GetForwaredPorts())
+	err = o.clientset.StateClient.SetForwardedPorts(portsBuf.GetForwardedPorts())
 	if err != nil {
 		return fmt.Errorf("unable to save forwarded ports to state file: %v", err)
 	}
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -224,7 +224,7 @@ func (o *DevOptions) Run(ctx context.Context) error {
 	// Output that the application is running, and then show the port-forwarding information
 	log.Info("\nYour application is now running on the cluster")
 
-	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice))
+	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
 	go func() {
 		err = o.clientset.KubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, o.errOut)
 		if err != nil {
@@ -233,6 +233,10 @@ func (o *DevOptions) Run(ctx context.Context) error {
 	}()
 
 	portsBuf.Wait()
+	err = o.clientset.StateClient.SetForwardedPorts(portsBuf.GetForwaredPorts())
+	if err != nil {
+		fmt.Printf("unable to save forwarded ports to state file: %v", err)
+	}
 
 	devFileObj := o.Context.EnvSpecificInfo.GetDevfileObj()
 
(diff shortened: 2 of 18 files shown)


docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -226,3 +226,25 @@ components:
     memoryLimit: 1024Mi
     mountSources: true
 ```
+
+### State file
+
+When the command `odo dev` is executed, the state of the command isa saved in the file `./.odo/state.json`.
```
rm3l, line 232:
```suggestion
When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/state.json`. 
```

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -226,3 +226,25 @@ components:
...
+This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp 
+at which the state has been last saved, and the currently forwarded ports.
+
+```
+{
+ "timestamp": 1651247240,
+ "pid": 154474,
+ "forwardedPorts": [
+  {
+   "containerName": "runtime",
```
rm3l, line 243: Maybe we should update the example in the description of the original issue: the field is named `container` there..

rm3l (changes requested):
Just a very first quick look. I am yet to review the other changes in this PR :)

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
feloy: done, thanks

pkg/odo/cli/dev/dev.go
```diff
@@ -233,6 +233,10 @@ func (o *DevOptions) Run(ctx context.Context) error {
 	}()
 
 	portsBuf.Wait()
+	err = o.clientset.StateClient.SetForwardedPorts(portsBuf.GetForwaredPorts())
+	if err != nil {
+		fmt.Printf("unable to save forwarded ports to state file: %v", err)
```
rm3l, line 238: Shouldn't we exit right away if we cannot write to the state file? As I see it from the issue description, not exiting when such errors occur might create issues for other commands like `odo describe component`.
Of if an error here is not problematic, maybe we could print a warning instead (using `log.Warningf`).

pkg/odo/cli/dev/writer.go
```diff
@@ -45,3 +64,38 @@ func (o *PortWriter) Write(buf []byte) (n int, err error) {
 func (o *PortWriter) Wait() {
 	<-o.end
 }
+
+func (o *PortWriter) GetForwaredPorts() []api.ForwardedPort {
```
rm3l, line 68:
```suggestion
func (o *PortWriter) GetForwardedPorts() []api.ForwardedPort {
```

pkg/state/state.go
```diff
@@ -0,0 +1,59 @@
...
+		fs:                  fs,
+		getSecondsFromEpoch: getSecondsFromEpoch,
+		getpid:              os.Getpid,
+	}
+}
+
+func (o *State) SetForwardedPorts(fwPorts []api.ForwardedPort) error {
+	// TODO(feloy) When other data is persisted into the state file, it will be needed to read the file first
+	o.content.ForwardedPorts = fwPorts
+	o.setMetadata()
```
rm3l, line 30: I like the idea of persisting such metadata as well. But wouldn't it make sense to persist the `PID` earlier, as soon as the command starts? I am thinking of scenarios where it could be interesting to get access to information like the PID as soon as possible, for example if the command is hanging before this `SetForwardedPorts` method is called.

rm3l: Also, I suggest changing the `timestamp` field with two different fields, to make the intent clearer. Because I think the meaning was not clear without looking at the PR description. Maybe fields like `createdAt` and `lastUpdatedAt`.

tests/integration/devfile/cmd_dev_test.go
```diff
@@ -1398,4 +1399,50 @@ var _ = Describe("odo dev command tests", func() {
...
+			Expect(helper.VerifyFileExists(stateFile)).To(BeTrue())
+			contentJSON, err := ioutil.ReadFile(stateFile)
+			Expect(err).ToNot(HaveOccurred())
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.containerName", "runtime")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.containerName", "runtime")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.localAddress", "127.0.0.1")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.localAddress", "127.0.0.1")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.containerPort", "3000")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.containerPort", "4567")
+		})
```
rm3l, line 1432: I think we should also test that the `localPort` field is set (for both forwarded ports) to a value > 0

rm3l (changes requested):

pkg/state/state.go
feloy: That makes sense. I would prefer to discuss this during cabal to have a consensus before going further.

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -226,3 +226,23 @@ components:
     memoryLimit: 1024Mi
     mountSources: true
 ```
+
+### State file
+
+When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/state.json`.
```
rm3l, line 232: As discussed and as you changed it in the other files, the file name should now be `./.odo/devstate.json` ^^

```suggestion
When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/devstate.json`. 
```

tests/integration/devfile/cmd_dev_test.go
```diff
@@ -1398,4 +1415,52 @@ var _ = Describe("odo dev command tests", func() {
...
+			Expect(helper.VerifyFileExists(stateFile)).To(BeTrue())
+			contentJSON, err := ioutil.ReadFile(stateFile)
+			Expect(err).ToNot(HaveOccurred())
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.containerName", "runtime")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.containerName", "runtime")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.localAddress", "127.0.0.1")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.localAddress", "127.0.0.1")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.0.containerPort", "3000")
+			helper.JsonPathContentIs(string(contentJSON), "forwardedPorts.1.containerPort", "4567")
+			helper.JsonPathContentMatch(string(contentJSON), "forwardedPorts.0.localPort", "[0-9]+")
```
rm3l, line 1448: `[0-9]+` could potentially match `0` or even negative values, I think. I think we should really test for a port strictly greater than 0.
Maybe using a regexp like `^[1-9]+[0-9]*$`.

We could even go further by also testing that the number written for the port is less than 65,535.

rm3l (changes requested):

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -226,3 +226,23 @@ components:
     memoryLimit: 1024Mi
     mountSources: true
 ```
+
+### State file
+
+When the command `odo dev` is executed, the state of the command is saved in the file `./.odo/devstate.json`. 
+
+This state file contains the PID of the `odo dev` running process (`0` if no command is running), the timestamp
```
rm3l, line 234: Whoops - I just noticed that this statement about storing the PID and the timestamp is no longer true. So this sentence needs to be updated accordingly.

tests/helper/helper_generic.go
```diff
@@ -258,17 +258,25 @@ func CommonAfterEach(commonVar CommonVar) {
...
-func JsonPathContentContain(json string, path string, value string) bool {
+func JsonPathContentContain(json string, path string, value string) {
 	result := gjson.Get(json, path)
 	Expect(result.String()).To(ContainSubstring(value), fmt.Sprintf("content of path %q should contain %q but is %q", path, value, result.String()))
-	return true
+}
+
+func JsonPathContentIsValidUserPort(json string, path string) {
+	result := gjson.Get(json, path)
+	intVal, err := strconv.Atoi(result.String())
```
rm3l, line 274: Even better than using a regular expression 👍🏿

rm3l (changes requested):

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
feloy: oh, my bad, removing it

pkg/state/state.go
rm3l: Just for reference, it has been decided to remove both the timestamp and the PID for now. They may be added later on if needed.

rm3l (approved):

pkg/state/interface.go
```diff
@@ -0,0 +1,11 @@
+package state
+
+import "github.com/redhat-developer/odo/pkg/api"
+
+type Client interface {
+	// SetForwardedPorts sets the forwarded ports in the state file and saves it to the file, updating the metadata
+	SetForwardedPorts(fwPorts []api.ForwardedPort) error
```
dharmit, line 7: Thinking out loud - why do we want business layer to get involved with storing the state? Isn't this something that odo "CLI" is interested in doing? Maybe a different implementation using odo as library would not want to store the state.

feloy: In any case, we want to create an abstraction for the state, hence this package. 
For me, it is a business concern, as it is a way to be able to give to the CLI the state of the Dev session at any time. Any other CLI would be interested in getting this state.

PR #5705, opened 2022-04-29
Fix dev integration tests not finishing

feloy:
**What type of PR is this:**

/kind tests

### DIFF
--- a/tests/helper/helper_dev.go
+++ b/tests/helper/helper_dev.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 )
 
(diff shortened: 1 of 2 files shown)


tests/helper/helper_dev.go
```diff
@@ -169,6 +170,12 @@ func (o DevSession) WaitSync() ([]byte, []byte, error) {
 	return outContents, errContents, nil
 }
 
+func (o DevSession) CheckNotSynced(timeout time.Duration) {
+	time.Sleep(timeout)
```
rm3l, line 174: How about using [`Consistently`](https://onsi.github.io/gomega/#consistently) to repeatedly check the assertion during this period of time? This way, I think it will basically fail fast if the matcher does not pass, rather than just waiting until the end.

rm3l (approved):
LGTM - thanks.

feloy:
/override ci/prow/unit

Infra error

Tests pass on IBM Cloud

Issue #5704, opened 2022-04-29
Support `openshift` Devfile components (the same way as `kubernetes` components)


rm3l:
/kind feature

Stumbled upon this while working on #5451 

## Which functionality do you think we should add?

`openshift` components should have been deprecated and removed from the [Devfile specification](https://devfile.io/docs/devfile/2.0.0/user-guide/api-reference/), but this is currently not the case. So `odo` should support them.

## Why is this needed?

At this time, running `odo deploy` with an `openshift` component does not output any error/warning messages, and does not deploy the component either, which seems quite confusing.
For example, running `odo deploy` with the following [`devfile.yaml`](https://github.com/redhat-developer/odo/files/8589653/devfile.yaml.txt) succeeds with the following message: `Your Devfile has been successfully deployed`, but nothing was actually deployed.

After a discussion with @kadel:

> So there is a history behind openshfit component in devfile. It was there in devfile v1, and it got carried over to devfile v2, but the decision was made that it doesn't make much sense to have both openshift and kubernetes components when they are doing the same thing and that having just kubernetes component is ok. 
> But it looks like it was never deprecated from the spec.
> That is why it was never implemented. 
> But because it looks like it was never deprecated, and removing it would mean breaking change in spec, we will have to support it now.

rm3l:
**Grooming (2023-01-19)**

- ~Idea to make odo convert OpenShift components into K8s components right after parsing the Devfile. Beware of side effects~

rm3l:
After starting pair programming on this issue today, we noticed that running "odo dev" with a Devfile containing an OpenShift component would create this resource on the cluster, but immediately delete it afterward. This seems due to https://github.com/redhat-developer/odo/pull/6189, because only Kubernetes components are being loaded from the Devfile.
So we need to fix this behavior.

After digging further, we'll also need to check all places where Devfile K8s components are being loaded and handle OpenShift components accordingly. So this might affect the following commands as well:
- [ ] `describe component`: we need also to show OpenShift components in the human-readable output
- [ ] `delete component`: resources created for OpenShift components are currently not detected as candidates for deletion
- [ ] `dev`: K8s components not referenced by any apply command are automatically created by `odo dev`. This should also be the case for OpenShift components.

Issue #5637, opened 2022-04-06
odo delete fails for kubernetes components


gorkem:
/kind bug

## What versions of software are you using?

**Operating System:**
MacOS 

**Output of `odo version`:**
odo v2.5.0 (5626b5d8a)

## How did you run odo exactly?
odo delete component 

## Actual behavior

I have a kubernetes component that is defined as follows in my devfile
```YAML  
- name: outerloop-deploy
    kubernetes:
      uri: deploy.yaml
```

It gets deployed because I have an associated deploy command. When I invoke odo delete it fails with the following error. 

```
odo delete component
Searching resources to delete, please wait...
 ✗  Failed to gather resources for deletion: open /Users/***/****/backstage-starter/devfile.yaml/deploy.yaml: not a directory
```

## Expected behavior
component is deleted.

PR #5711, opened 2022-05-02
Fix issue with variables not substituted in `uri`-referenced manifests (#5451)

### LINKED ISSUE #5451: odo deploy - devfile variable substitution does not occur in uri referenced deployment manifest yaml file
/kind bug

**Operating System:**
```
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal
```
**Output of `odo version`:**
```
$ odo version
odo v2.5.0 (724f16e68)
```
`$ odo deploy`

with the following devfile variables defined:

```
variables:
  container_image: 'docker.io/ajymau/asi-outer-loop:app-provided'
  appname: 'test-inline'
```

and the following component (and related commands) defined:

```
  - name: myk8sdeploy
    kubernetes:
      uri: ./app-deploy.yaml
```

app-deploy.yaml contains substitution placeholders for the vars defined in the devfile:

```
apiVersion: apps.openliberty.io/v1beta2
kind: OpenLibertyApplication
metadata:
  name: {{appname}}
spec:
  # Add fields here.
  applicationVersion: 1.0.0
  applicationImage: {{container_image}}
  service:
    type: ClusterIP
    port: 9080
    annotations:
      prometheus.io/scrape: 'true'
  probes:
    readiness:
      failureThreshold: 12
      httpGet:
        path: /health/ready
        port: 9080
      initialDelaySeconds: 5
      periodSeconds: 2
      timeoutSeconds: 1
    liveness:
      failureThreshold: 12
      httpGet:
        path: /health/live
        port: 9080
      initialDelaySeconds: 5
      periodSeconds: 2
  expose: true
  route:
    # Ingress required entries.
    pathType: 'Prefix'
    path: '/'
```

## Actual behavior
the app-deploy.yaml is used as is to do the deploy - no substitution occurs and the deploy fails because the names cannot be parsed:

```
$ odo deploy
.
.
..
173357979467: Layer already exists
app-provided: digest: sha256:40a963755d2202edf123e5868227dc64490a95302855ad20a9a607ced80d92f8 size: 4706
 ✗  command execution failed: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"appname":interface {}(nil)}
 $
```

## Expected behavior
variable substitution occurs first and the deploy occurs next, successfully

Note: if i inline the contents of app-deploy.yaml in the devfile the substitutions occur as expected. This should happen as well for a uri specified deployment manifest yaml.

rm3l:
**What type of PR is this:**
/kind bug

This fixes the issue reported in #5451, where running `odo deploy` (actually any commands calling `libdevfile.GetK8sComponentAsUnstructured`) using a Devfile containing `uri`-referenced manifests containing variables would  not succeed, either with a YAML validation error message similar to the line below or with an error (e.g., unresolvable image name) in the manifest once deployed in the cluster:

```
error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"appname":interface {}(nil)}
```

Fixes #5451 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

Refer to the reproduction steps depicted in the original issue for testing this: #5451 

From the code perspective, as discussed during one of the Sprint Plannings, this PR is enriching the `libdevfile` package with a new `GetK8sManifestWithVariablesSubstituted` function that is responsible for returning the resource manifest from the Kubernetes/OpenShift Devfile component with all variables substituted, rather than having `libdevfile.GetK8sComponentAsUnstructured` (and few other places in the codebase) fetch the raw manifest from the URI and handle it as is.
Also, the implementation aims at leaving room for supporting `OpenShift` components in a near future (scope of #5704).

~Since this seems a much more realistic scenario of how outer-loop manifests are referenced in Devfiles, this PR also switches to `uri` references in all test Devfiles. This is the purpose of df1f19ed959b8c82aa17ac3cf12c343305789149
Note that tests for `odo deploy` still test both `inlined` (as before) and `uri` manifests (scope of this issue).~ [Will be covered in a separate PR]


pkg/libdevfile/libdevfile.go
```diff
@@ -193,3 +199,83 @@ func GetEndpointsFromDevfile(devfileObj parser.DevfileObj, ignoreExposures []v1a
 	}
 	return endpoints, nil
 }
+
+// GetComponentResourceManifestContentWithVariablesResolved returns the full content of either a Kubernetes or an Openshift
+// Devfile component, either Inlined or referenced via a URI.
+// No matter how the component is defined, it returns
+// the content with all variables substituted using the global variables map defined in `devfileObj`.
+// An error is returned if the content references an invalid variable key not defined in the Devfile object.
+func GetComponentResourceManifestContentWithVariablesResolved(devfileObj parser.DevfileObj, devfileCmp interface{},
```
feloy, line 208: Now we get the entire devfileObj as a parameter, you could pass the name of the component instead of its content

tests/integration/devfile/cmd_devfile_build_images_test.go
```diff
@@ -79,6 +79,9 @@ var _ = Describe("odo devfile build-images command tests", func() {
 
 		BeforeEach(func() {
 			helper.CopyExample(filepath.Join("source", "nodejs"), commonVar.Context)
+			helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project", "kubernetes",
```
feloy, line 82: Should odo copy these dependencies when running `odo init --devfile-path`? @kadel WDYT?

feloy (changes requested):

tests/integration/devfile/cmd_devfile_build_images_test.go
rm3l: +1. Indeed, that would be a nice feature, I think.

pkg/libdevfile/libdevfile.go
rm3l: Is there a simple way to retrieve a Component by its name in `devfileObj`? It is not clear in the docs, but are we sure component names are unique in the Devfile? (Maybe by both name and type?)

Or maybe I could avoid passing the `devfileObj`, and use the variables map as an argument instead. This is what I actually used `devfileObj` for, to substitute the variables using: `devfileObj.Data.GetDevfileWorkspaceSpec().Variables`

WDYT?

rm3l: Ah, I found a `Filter` field in `DevfileOptions` that could allow to retrieve Components by their name. But still not sure about the unicity.

rm3l: Just found our `ComponentsWithSameNameError` error type we already have in `libdevfile`. This seems to answer my previous question :)
So I will make the changes you suggested, using the component name instead.

feloy (commented):
You have moved devfile from using `Inlined` to using `URI` as discussed during issue preparation. Are there some remaining devfile using `Inlined`? I would not like we make regressions on this now :(

pkg/libdevfile/errors.go
```diff
@@ -64,14 +64,18 @@ func (e ComponentNotExistError) Error() string {
...
-	return ComponentsWithSameNameError{}
+func NewComponentsWithSameNameError(name string) ComponentsWithSameNameError {
+	return ComponentsWithSameNameError{
+		name: name,
+	}
 }
 
 func (e ComponentsWithSameNameError) Error() string {
-	return "more than one component with the same name, should not happen"
+	return fmt.Sprintf("more than one component with the same name (%q), should not happen", e.name)
```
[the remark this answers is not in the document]
rm3l: Good catch. I'll change it - thanks.

rm3l:
> You have moved devfile from using `Inlined` to using `URI` as discussed during issue preparation. Are there some remaining devfile using `Inlined`? I would not like we make regressions on this now :(

I identified the following places:
- [`devfile-deploy-with-k8s-inlined.yaml`](https://github.com/redhat-developer/odo/pull/5711/files#diff-1e678e962ad0dcb3c1e54400840d6e085daba732288da5692d2464adc40d0a89) (actually, it is just the old `devfile-deploy.yaml` file that I renamed). Renaming it was intentional to make sure 'Inlined' manifests could also be tested with `odo deploy` as before.
- [`devfile-deploy-with-multiple-resources.yaml`](https://github.com/redhat-developer/odo/pull/5711/files#diff-81ecd14331e084235ba14a08e423561e6aedded41486f39f137385f065a92405) : this one can't use URI references at this time because of #5637. I added `TODO` comments in this file.

I was also hesitant on changing this in this PR. That's why the commit is isolated if we want to revert or do this change in a separate issue/PR.

Ideally, I think we should come up with a way to easily reuse the same tests against both Inlined and URI-referenced manifests. This is what I tried to accomplish with this loop in [`cmd_devfile_deploy_test.go`](https://github.com/redhat-developer/odo/pull/5711/files#diff-ffa8db34cac925d833d18af25ef37c7e7466ea55253fd810215ce600c3d9c01b).

feloy (approved):

rm3l:
> > You have moved devfile from using `Inlined` to using `URI` as discussed during issue preparation. Are there some remaining devfile using `Inlined`? I would not like we make regressions on this now :(
> 
> I identified the following places:
> 
> * [`devfile-deploy-with-k8s-inlined.yaml`](https://github.com/redhat-developer/odo/pull/5711/files#diff-1e678e962ad0dcb3c1e54400840d6e085daba732288da5692d2464adc40d0a89) (actually, it is just the old `devfile-deploy.yaml` file that I renamed). Renaming it was intentional to make sure 'Inlined' manifests could also be tested with `odo deploy` as before.
> * [`devfile-deploy-with-multiple-resources.yaml`](https://github.com/redhat-developer/odo/pull/5711/files#diff-81ecd14331e084235ba14a08e423561e6aedded41486f39f137385f065a92405) : this one can't use URI references at this time because of [odo delete fails for kubernetes components #5637](https://github.com/redhat-developer/odo/issues/5637). I added `TODO` comments in this file.
> 
> I was also hesitant on changing this in this PR. That's why the commit is isolated if we want to revert or do this change in a separate issue/PR.
> 
> Ideally, I think we should come up with a way to easily reuse the same tests against both Inlined and URI-referenced manifests. This is what I tried to accomplish with this loop in [`cmd_devfile_deploy_test.go`](https://github.com/redhat-developer/odo/pull/5711/files#diff-ffa8db34cac925d833d18af25ef37c7e7466ea55253fd810215ce600c3d9c01b).

To be safe, I just reverted that commit from this PR, but saved it to a separate branch, and will try to include it in a dedicated PR later on.

rm3l:
Just rebased onto `main` to benefit from #5712, which fixed the tests.

tests/integration/devfile/cmd_devfile_build_images_test.go
rm3l: For reference, #5713 has been created for this point.

rm3l:

Looks like a problem not related to this PR, but related to the deprovisioning job that tears down the cluster on Prow: 

>   * could not run steps: step integration-e2e failed: "integration-e2e" post steps failed: "integration-e2e" pod "integration-e2e-ipi-deprovision-deprovision" failed: the pod ci-op-kzni8cn5/integration-e2e-ipi-deprovision-deprovision was deleted without completing after 4m50s (failed containers: )

All tests passed on IBM Cloud by the way.

PR #5731, opened 2022-05-11
odo list -o json

### LINKED ISSUE #5643: Provide an `odo list -o json` command
`odo list` command should provide a `-o json` flag to return the list of components in json format.

The list should be prepared to contain several kind of objects: components,  bindings, etc.

```
$ odo list -o json
{
  componentInDevfile: "a-component",
  components: [ 
  {
    "name": "a-component",
    "projectType": "nodejs",
    "runningIn": {
        "Dev": true, 
        "Deploy": true
    },
    "managedBy": "odo"
  }
  ]
}
```

The caller will use the exit status of the command to get the error status. 0=> OK and  (not 0) => Error

In case of error, the json output will be:
```
{
  message: "..."
}
```

feloy:
**What type of PR is this:**

/kind feature

Fixes #5643 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 


rm3l (commented):
Using a Devfile that does not declare any (optional) `projectType` in its metadata (like [`devfile-no-endpoints.yaml`](https://github.com/redhat-developer/odo/blob/main/tests/examples/source/devfiles/nodejs/devfile-no-endpoints.yaml) in our tests), I can see that this command returns `"projectType": "Not available"`, which makes think that this is the actual project type value.
I think it makes sense to display "Not available" for human-readable output, but for JSON, shouldn't we return a `null` JSON field or an empty string instead ?

```json
{
        "componentsInDevfile": [
                "nodejs"
        ],
        "components": [
                {
                        "name": "my-component-with-command-and-args",
                        "managedBy": "odo",
                        "runningIn": [
                                "Dev"
                        ],
                        "projectType": "quarkus"
                },
                {
                        "name": "nodejs",
                        "managedBy": "",
                        "runningIn": [],
                        "projectType": "Not available"
                }
        ]
}

```

pkg/component/component.go
```diff
@@ -175,34 +175,40 @@ func ListAllClusterComponents(client kclient.ClientInterface, namespace string)
...
+					for _, m := range components[v].RunningIn {
+						if m == api.RunningMode(mode) {
+							modeFound = true
+							break
+						}
+					}
+					if !modeFound {
+						components[v].RunningIn = append(components[v].RunningIn, api.RunningMode(mode))
+						sort.Sort(components[v].RunningIn)
+					}
```
rm3l, line 199: I couldn't determine if integration tests covered this. Would it be possible to update the unit test cases to make sure the changes here are covered?

feloy:
> Using a Devfile that does not declare any (optional) `projectType` in its metadata (like [`devfile-no-endpoints.yaml`](https://github.com/redhat-developer/odo/blob/main/tests/examples/source/devfiles/nodejs/devfile-no-endpoints.yaml) in our tests), I can see that this command returns `"projectType": "Not available"`, which makes think that this is the actual project type value. I think it makes sense to display "Not available" for human-readable output, but for JSON, shouldn't we return a `null` JSON field or an empty string instead ?
> 
> ```json
> {
>         "componentsInDevfile": [
>                 "nodejs"
>         ],
>         "components": [
>                 {
>                         "name": "my-component-with-command-and-args",
>                         "managedBy": "odo",
>                         "runningIn": [
>                                 "Dev"
>                         ],
>                         "projectType": "quarkus"
>                 },
>                 {
>                         "name": "nodejs",
>                         "managedBy": "",
>                         "runningIn": [],
>                         "projectType": "Not available"
>                 }
>         ]
> }
> ```

Yes, that's a good point. I would like to think about this in a more global way, to have something homogeneous over the entire API. I'll add this subject to  the next cabal

pkg/component/component_test.go
```diff
@@ -71,15 +81,38 @@ func TestListAllClusterComponents(t *testing.T) {
...
 				Name:      "svc1",
 				ManagedBy: "odo",
-				Modes:     map[string]bool{},
+				RunningIn: nil,
+				Type:      "nodejs",
+			}},
+			wantErr: false,
+		},
+		{
+			name: "TODO",
```
rm3l, line 98: Thanks for adding this case. A more meaningful name would be relevant here. ;-)

rm3l (changes requested):
LGTM from the code perspective. Just a small comment on a test case name ;)

PR #5749, opened 2022-05-18
Substituting variables into the devfile from the CLI

### LINKED ISSUE #5489: Substituting variables into the devfile from the CLI
/kind feature

Actually, it is possible to use variable substitution on devfile, and the value of the variables comes from the devfile itself (from the `.variables` part).

It would be helpful to be able to pass values for variables from the CLI, without modifying the devfile.

For consistency, the same behaviour as the `--env` and `--env-file` options from `docker` could be used, by replacing `env` by `var` as these are not *environment* variables. For example:

```
$ odo dev --var CONTAINER_IMAGE=quay.io/myuser/myimage --var DEBUG=true --var USER
$ odo dev --var-file my-variables.var
$ cat my-variables.var
CONTAINER_IMAGE=quay.io/myuser/myimage
DEBUG=true
USER # this gets the value from the CLI current environment, if defined. If the env var is not defined, the variable is not overriden
``` 

AC:
- [ ] for both `odo dev` and `odo deploy` commands
- [ ] it should not modify devfile.yaml

## Why is this needed?

It seems to be a common use case to create templates to be personalized at run time, similarly to how it is done with helm.

Initial discussion: https://kubernetes.slack.com/archives/C01D6L2NUAG/p1645604581682759

feloy:
**What type of PR is this:**

/kind feature

Fixes #5489

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -48,11 +48,49 @@ will first delete all resources deployed into the cluster for this session befor
 The Devfile can define variables to make the Devfile parameterizable. The Devfile can define values for these variables, and you 
 can override the values for variables from the command line when running `odo dev`, using the `--var` and `--var-file` options.
 
-The `--var` option takes a comma-separated list of `variable=value` pairs, where `=value` is optional. If the `=value` is omitted, the value is extracted from the environment variable named `variable`. In this case, if the environment variable with this name is not defined, the value defined into the Devfile will be used.
+The `--var` option is a repeatable option that takes a `variable=value` pair, where `=value` is optional. If the `=value` is omitted, the value is extracted from the environment variable named `variable`. In this case, if the environment variable with this name is not defined, the value defined into the Devfile will be used.
 
 The `--var-file` option takes a filename as argument. The file contains a line-separated list of `variable=value` pairs, with the same behaviour as before.  
 
 Note that the values passed with the `--var` option overrides the values obtained with the `--var-file` option.
+
+#### Examples
+
+Considering the Devfile contains this `variables` field:
+
+```
+variables:
+  USER: anonymous
+  DEBUG: false
+```
+
+
+This command will override the `USER` Devfile variable with the value of the `USER` environment variable, if it is defined.
+It will also override the value of the `DEBUG` Devfile variable with the `true` value.
+
+```shell
+$ odo dev --var USER --var DEBUG=true
+```
+
+If you create a file `config.vars` containing:
+
+```
+USER
+DEBUG=true
+```
+
+The following command will have the same behaviour as the previous one:
+
+```shell
+$ odo dev --var-file config.vars
+```
+
+The following command will override the `USER` Devfile variable with the `john` value:
+
+
+```shell
+$ odo dev --var USER=john --var-file config.vars
+```
 
 ## Devfile (Advanced Usage)
 
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
@@ -43,6 +43,17 @@ In the above example, three things have happened:
 You can press Ctrl-c at any time to terminate the development session. The command can take a few moment to terminate, as it
 will first delete all resources deployed into the cluster for this session before terminating.
 
+### Sustituting variables
+
+The Devfile can define variables to make the Devfile parameterizable. The Devfile can define values for these variables, and you 
+can override the values for variables from the command line when running `odo dev`, using the `--var` and `--var-file` options.
+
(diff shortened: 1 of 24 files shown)


feloy:
Blocked by https://github.com/devfile/library/pull/136

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -43,6 +43,17 @@ In the above example, three things have happened:
 You can press Ctrl-c at any time to terminate the development session. The command can take a few moment to terminate, as it
 will first delete all resources deployed into the cluster for this session before terminating.
 
+### Sustituting variables
+
+The Devfile can define variables to make the Devfile parameterizable. The Devfile can define values for these variables, and you 
+can override the values for variables from the command line when running `odo dev`, using the `--var` and `--var-file` options.
+
+The `--var` option takes a comma-separated list of `variable=value` pairs, where `=value` is optional. If the `=value` is omitted, the value is extracted from the environment variable named `variable`. In this case, if the environment variable with this name is not defined, the value defined into the Devfile will be used.
```
rm3l, line 51: Is the `--var` option really a comma-separated list of pairs?

I tried using a comma-separated list of `variable=value` pairs, but my `odo deploy` errored out with the message below, suggesting that it did not handle the comma separator:
```
❯ odo deploy --var RESOURCE_NAME=rm3l-dotnet,CONTAINER_PORT=1234
...
 ✗  failed to create service(s) associated with the component: Deployment.apps "rm3l-dotnet,CONTAINER_PORT=1234" is invalid
...

```

From the `cobra` doc, if we want to support both repeatable arguments and comma-separated values, we should use [`StringSliceVar`](https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar) instead of [`StringArrayVar`](https://pkg.go.dev/github.com/spf13/pflag#StringArrayVar).
But I am not sure having a comma-separated list is the expected behavior, is it?

Also, I think providing an example of usage in this document could be helpful.

feloy: You are right, the expected behaviour is to have only repeatable args, to be able to have commas in values. I fixed the doc and added some examples.

pkg/vars/vars.go
```diff
@@ -0,0 +1,112 @@
...
+	"bufio"
+	"fmt"
+	"strings"
+	"unicode"
+
+	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
+)
+
+// GetVariables returns a map of key/value from pairs defined in the file and in the list of strings in the format KEY=VALUE or KEY
+// For a KEY entry, the value wbe obtained from the environment, and if the value is not defined in the environment, the entry KEY will be ignored
```
valaparthvi, line 13:
```suggestion
// For a KEY entry, the value will be obtained from the environment, and if the value is not defined in the environment, the entry KEY will be ignored
```

pkg/vars/vars.go
```diff
@@ -0,0 +1,112 @@
...
+
+	for k, v := range overrideVars {
+		result[k] = v
+	}
+
+	return result, nil
+}
+
+// parseKeyValueFile parses a file for "KEY=VALUE" lines and returns a map of keys/values
+// If a key is defined without value "KEY", the value is searched into the environment with lookupEnv function
```
valaparthvi, line 38:
```suggestion
// If a key is defined without a value as "KEY", the value is searched into the environment with lookupEnv function
```

pkg/vars/vars.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		result[key] = value
+	}
+	return result, nil
+}
+
+// parseKeyValueString parses a string to extract a key and its associated value
+// if a line is empty or a comment, a nil error and an empty key are returned
+// if a key does not define a value, the value will be obtained from the environment
+// in this case, if the environment does not define the valriable, the entrey will be ignored
```
valaparthvi, line 86:
```suggestion
// in this case, if the environment does not define the variable, the entry will be ignored
```

docs/website/versioned_docs/version-3.0.0/command-reference/dev.md
```diff
@@ -43,6 +43,55 @@ In the above example, three things have happened:
...
+$ odo dev --var-file config.vars
+```
+
+The following command will override the `USER` Devfile variable with the `john` value:
+
+
+```shell
+$ odo dev --var USER=john --var-file config.vars
+```
+
```
valaparthvi, line 94: What do you think about adding a note stating that passing a KEY=VALUE for a key that does not exist in the devfile.yaml will be ignored?

valaparthvi (commented):
Code LGTM apart from a few nitpicks.

rm3l (changes requested):
Added one last comment on the code..

rm3l (approved):

feloy:
Infra error

IBM tests pass

Issue #5622, opened 2022-04-04
`odo logs` part 1


kadel:
Part of the https://github.com/redhat-developer/odo/issues/5533 user story

## users should be able to use odo logs command to display stdout and stderr logs of components deployed in dev and deploy mode

- [x] `odo logs` show ALL logs (both dev mode and deploy mode, from all containers)
- [x] `odo logs --dev` shows logs of all containers running in dev mode
- [x] `odo logs --deploy` show logs of all containers from deploy mode
- [x] each log line should be prefixed with name of the container
- [x] command need to be able to handle multiple deployments, pods, and containers (For example devfile can contain multiple Deployments in outer-loop or there can be multiple containers in inner loop etc... )
- [x] Documentation

dharmit:
@kadel this might have been discussed on some call, but it's not noted on this or the parent issue - do we want to follow the logs by default or do we want to print the logs available at the time of running the command? This might have been taken into consideration when pointing the issue as well, but I was probably not around on that call (this Monday.)

dharmit:
In #5760, I have worked upon displaying logs for pods created by odo during the inner loop, a.k.a. Dev mode. It doesn't have follow/tail ability yet. Deploy mode, flags for `--dev` and `--deploy` will follow in a separate PR later. However, from the acceptance criteria, I am ensuring that below two items are covered (IMO, they need to be covered for all scenarios):
* each log line should be prefixed with name of the container
* command need to be able to handle multiple deployments, pods, and containers

kadel:
remaining at least 5 points of work

feloy:
I'm reopening this issue as an integration test covering this feature is failing from time to time:

```
• Failure [2.493 seconds]
odo logs command tests
/home/phmartin/Documents/gits/odo/tests/integration/devfile/cmd_logs_test.go:11
  when component is created and odo logs is executed
  /home/phmartin/Documents/gits/odo/tests/integration/devfile/cmd_logs_test.go:38
    when running in Deploy mode
    /home/phmartin/Documents/gits/odo/tests/integration/devfile/cmd_logs_test.go:71
      should successfully show logs of the running component [It]
      /home/phmartin/Documents/gits/odo/tests/integration/devfile/cmd_logs_test.go:78

      No future change is possible.  Bailing out early after 1.058s.
      Running odo with args [odo logs]
      Expected
          <int>: 1
      to match exit code:
          <int>: 0

```

```
[...]

PR #5760, opened 2022-05-23
Add `odo logs`

dharmit:
**What type of PR is this:**

$subject

Fixes part of #5622

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation  (https://deploy-preview-5760--odo-docusaurus-preview.netlify.app/docs/3.0.0/command-reference/logs)

* In simplest case, create a nodejs component and run `odo logs` in a separate window after running `odo dev`.
* A bit complicated case. Use the below devfile.yaml, create another pod (using pod.yaml) outside the odo flow, and run `odo logs`. odo should show logs for all containers except the one created using `pod.yaml` because it wasn't created using odo.
    <details>
  commands:
  - exec:
      commandLine: npm install
      component: runtime
      group:
        isDefault: true
        kind: build
      workingDir: ${PROJECT_SOURCE}
    id: install
  - exec:
      commandLine: npm start
      component: runtime
      group:
        isDefault: true
        kind: run
      workingDir: ${PROJECT_SOURCE}
    id: run
  - exec:
      commandLine: npm run debug
      component: runtime
      group:
        isDefault: true
        kind: debug
      workingDir: ${PROJECT_SOURCE}
    id: debug
  - exec:
      commandLine: npm test
      component: runtime
      group:
        isDefault: true
        kind: test
      workingDir: ${PROJECT_SOURCE}
    id: test
  components:
  - container:
      endpoints:
      - name: http-3000
        targetPort: 3000
      memoryLimit: 1024Mi
      mountSources: true
    name: runtime
  - name: infinitepodone
    kubernetes:  
      inlined: |
        apiVersion: v1
        kind: Pod
        metadata:
          name: infinitepodone
        spec:
          containers:
            - name: blah
              image: docker.io/dharmit/infiniteloop
  - name: infinitedeployment
    kubernetes:
      inlined: |
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: infinitedeployment
        spec:
          replicas: 1
          selector:
            matchLabels:
              app: infinite
          template:
            metadata:
              labels:
                app: infinite
            spec:
              containers:
              - name: podone
                image: docker.io/dharmit/infiniteloop
              - name: podtwo
                image: docker.io/dharmit/infiniteloop
  metadata:
    description: Stack with Node.js 14
    displayName: Node.js Runtime
    icon: https://nodejs.org/static/images/logos/nodejs-new-pantone-black.svg
    language: javascript
    name: node
    projectType: nodejs
    tags:
    - NodeJS
    - Express
    - ubi8
    version: 1.0.1
  schemaVersion: 2.0.0
  starterProjects:
  - git:
      remotes:
    name: nodejs-starter
    </details>
    <details>
    <summary>pod.yaml</summary>
  apiVersion: v1
  kind: Pod
  metadata:
    name: infinitepodtwo
  spec:
    containers:
      - name: blah
        image: docker.io/dharmit/infiniteloop
    </details>


pkg/logs/interface.go
```diff
@@ -0,0 +1,9 @@
+package logs
+
+import "io"
+
+type Client interface {
+	// DevModeLogs gets logs for the provided component name and namespace. A component could have multiple pods and
+	// containers running on the cluster. It returns a map where container name is the key and its logs are the value
```
feloy, line 7: You will probably have several containers from different pods with same name

pkg/logs/logs.go
```diff
@@ -0,0 +1,111 @@
...
+	// first, check if ownerReference belongs to any of the resources
+	for _, resource := range resources {
+		if resource.GetUID() == owner.UID {
+			return true, nil
+		}
+	}
+	// second, get the resource indicated by ownerReference and check its ownerReferences field
+	var ownerReferences []metav1.OwnerReference
+	switch owner.Kind {
+	case "ReplicaSet":
```
feloy, line 93: You could use GetDynamicResource to get any resource kind, not only ReplicaSet and Deployment (we will want to work with knative workloads, for example)

feloy (changes requested):

dharmit:
@rm3l i'm feeling pretty dumb about asking this, but in spite of adding below to `.golangci.yaml`, I keep seeing the error:
```diff
+  nolintlint:
+    # Disable to ensure that all nolint directives actually have an effect.
+    # Default: false
+    allow-unused: false
```
To me, the message is not very clear, so had `allow-unusde: true` as well. Still it fails. I have already added `//nolint` (without space between `//` and `/nolint`) to the line in question.

Here's the error I'm getting on CI and locally:
```
go run -mod=vendor github.com/securego/gosec/v2/cmd/gosec -severity medium -confidence medium -exclude G304,G204 -quiet  ./...
Results:


[/home/dshah/src/odo/pkg/odo/cli/logs/logs.go:135] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    134: 	rand.Seed(time.Now().UnixNano())
  > 135: 	i := rand.Intn(len(colors)) //nolint
    136: 	return colors[i]

```
What am I missing here?

pkg/logs/interface.go
dharmit: Yep. How do we want to handle such a situation? 

We didn't discuss/imagine this scenario earlier. Do you think it would be rare enough to fix when some user opens an issue about it? 

Or, we could make pod name + container name as the key. However, odo users might not really know what a pod is, no? They are generally enterprise devs good at some framework/language, and not at k8s. It would also be ugly because odo creates a deployment which creates pod. For example, I have below pods created using the devfile I have mentioned in PR:
```sh
$ kubectl get pods
NAME                                  READY   STATUS    RESTARTS       AGE
infinitedeployment-67899ff5c5-jppvz   2/2     Running   0              31s
infinitepodone                        1/1     Running   0              31s
infinitepodtwo                        1/1     Running   4 (126m ago)   6d4h
node-app-64c7b9c8fc-kglhz             1/1     Running   0              31s
```

Again, pod is the most basic unit of deployment for k8s, so one could argue that someone dealing with k8s in any form would know what a pod is.

cc @kadel

pkg/kclient/replicaset.go
```diff
@@ -0,0 +1,17 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// GetReplicaSetByName gets a replicaset matching the given name
+func (c *Client) GetReplicaSetByName(name string) (*appsv1.ReplicaSet, error) {
+	replicaSet, err := c.KubeClient.AppsV1().ReplicaSets(c.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return &appsv1.ReplicaSet{}, err
```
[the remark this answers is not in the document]
dharmit: Removed this file itself because we could use `GetDynamicResource` as you suggested. :)

pkg/logs/logs.go
```diff
@@ -0,0 +1,127 @@
...
+			logs[container.Name] = containerLogs
+		}
+	}
+
+	return logs, nil
+}
+
+// matchOwnerReferenceWithResources recursively checks if the owner reference passed to it matches any of the resources
+// This is useful when trying to find if a pod is owned by any of the ReplicaSet or Deployment in the cluster.
+func (o *LogsClient) matchOwnerReferenceWithResources(owner metav1.OwnerReference, resources []unstructured.Unstructured) (bool, error) {
```
feloy, line 86: Could you add some unit tests for this function?

pkg/logs/logs.go
```diff
@@ -0,0 +1,127 @@
...
+	// first, check if ownerReference belongs to any of the resources
+	for _, resource := range resources {
+		if resource.GetUID() == owner.UID {
+			return true, nil
+		}
+	}
+	// second, get the resource indicated by ownerReference and check its ownerReferences field
+	group, version := getGroupVersion(owner.APIVersion)
```
feloy, line 94: You could use this function from `https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/schema/`:

```
FromAPIVersionAndKind(apiVersion, kind string) GroupVersionKind
```

```
restMapping, err := o.kubernetesClient.GetRestMappingFromGVK(schema.FromAPIVersionAndKind(owner.APIVersion, owner.Kind))
```

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+	return nil
+}
+
+// printLogs prints the logs of the containers with container name prefixed to the log message
+func printLogs(containerName string, rd io.ReadCloser, out io.Writer) error {
+	reader := bufio.NewReader(rd)
+	var lines []string
+	for {
```
feloy, line 106: Could we use a scanner instead?

https://pkg.go.dev/go/scanner

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,156 @@
...
+		line, err := reader.ReadString('\n')
+		if err != nil {
+			if err != io.EOF {
+				return err
+			} else {
+				break
+			}
+		}
+		line = strings.Join([]string{containerName, line}, ": ")
+		lines = append(lines, line)
```
feloy, line 116: Why don't you print the line at this point?

You are storing all the lines in memory, this can be very memory consuming if the logs are very long (this regularly happens when there is an error on the container, it's important to not cascade the problems)

rm3l:
> @rm3l i'm feeling pretty dumb about asking this, but in spite of adding below to `.golangci.yaml`, I keep seeing the error locally and on CI:
> 
> ```
> go run -mod=vendor github.com/securego/gosec/v2/cmd/gosec -severity medium -confidence medium -exclude G304,G204 -quiet  ./...
> Results:
> 
> 
> [/home/dshah/src/odo/pkg/odo/cli/logs/logs.go:135] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
>     134: 	rand.Seed(time.Now().UnixNano())
>   > 135: 	i := rand.Intn(len(colors)) //nolint
>     136: 	return colors[i]
> ```
> 
> ```diff
> +  nolintlint:
> +    # Disable to ensure that all nolint directives actually have an effect.
> +    # Default: false
> +    allow-unused: false
> ```
> 
> To me, the message is not very clear, so had `allow-unusde: true` as well. Still it fails. I have already added `//nolint` (without space between `//` and `/nolint`) to the line in question.
> 
> What am I missing here?

@dharmit The error here is being reported by `gosec` (run by the `sec` Make target). At this point, `golangci-lint` has not started yet - so changing `.golangci.yaml` will have no effect.
How about applying the recommendation displayed, by using `crypto/rand`  instead of `math/rand`?
But if you think this is a false positive, you can annotate the code wit a `#nosec [RuleList] [-- Justification]` comment, as documented here: https://github.com/securego/gosec#annotating-code

feloy (changes requested):

pkg/logs/interface.go
kadel: > We didn't discuss/imagine this scenario earlier. Do you think it would be rare enough to fix when some user opens an issue about it?

We need to handle this scenario. Sooner or later it will be a problem and we should have a solution from the beginning. I can see situations where for example the container for the inner loop and the outer loop would be named the same way. 

> Again, pod is the most basic unit of deployment for k8s, so one could argue that someone dealing with k8s in any form would know what a pod is.

We should not try to hide the fact that there are Kubernetes resources underneath.  We are already exposing users to Pod, Deployments terms in other parts like `ode delete` and it is completely fine. 



For the inner loop we should be using name of the container from the Devfile container component (`components.container.name`). Devfile spec guarantees that there can't be two components with the same name.

With the outer loop resources, it is more complicated. I wouldn't be afraid to show Deployment name / Pod name / container name. But that could get quite long. @feloy proposal could solve this problem here.

dharmit:
@rm3l thanks for your input. I need to do a better job at reading the errors. It was quite clear that gosec complained about it, not golangci-lint.

pkg/logs/logs.go
```diff
@@ -0,0 +1,127 @@
...
+
+func NewLogsClient(kubernetesClient kclient.ClientInterface) *LogsClient {
+	return &LogsClient{
+		kubernetesClient: kubernetesClient,
+	}
+}
+
+var _ Client = (*LogsClient)(nil)
+
+func (o *LogsClient) DevModeLogs(componentName string, namespace string) (map[string]io.ReadCloser, error) {
```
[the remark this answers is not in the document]
dharmit: Indeed. But I'm separating things into `DevModeLogs`, `DeployModeLogs` and `AllModeLogs`. This is because we are going to support flags that support showing logs for different modes. You don't see the other two methods because this PR is targeting only dev mode.

pkg/odo/cli/logs/logs.go
dharmit: What are the advantages of using `scan.Scanner` over `bufio.Reader`?

feloy: Scanner handles different end of lines for you

dharmit: And as far as logs are concerned, what could those different end of lines be? Even in general, what other end of lines does one need to consider? Genuinely curious.

pkg/kclient/pods.go
```diff
@@ -242,3 +242,13 @@ func (c *Client) GetPodLogs(podName, containerName string, followLog bool) (io.R
 
 	return rd, err
 }
+
+func (c *Client) GetAllPodsInNamespace() (*corev1.PodList, error) {
+	pods, err := c.KubeClient.CoreV1().Pods(c.Namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		// Don't wrap error since we want to know if it's a forbidden error
+		// if we wrap, we lose the err status reason and callers of this api rely on it
```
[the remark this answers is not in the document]
dharmit: You're right. The comment you're seeing is a copy-paste from another method in the same file.

pkg/odo/cli/logs/logs.go
feloy: Unix systems use `\n` when Windows ones use `\r\n`

I'm not sure what could be returned as part of the logs, probably `\n` only, but not sure. `Scanner` helps not bother with this.

kadel:
Each time I run `odo logs` the output is in a different color. Is that a feature?

image

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,156 @@
...
+	for i := 0; i < len(lines); i++ {
+		_, err := fmt.Fprintf(out, lines[i])
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func selectRandomColor() color.Attribute {
```
cdrage, line 132: This all should be in the logging pkg/log/status.go file..

These colours won't work correctly on terminals that don't support colour such as cmd.exe on Windows. Make sure we use `log.GetStdout()` for any output as well, have a look at the `pkg/log/status.go` file and implement the colourization there please!

cdrage (changes requested):

dharmit:
> Each time I run `odo logs` the output is in a different color. Is that a feature?

@kadel we wanted different colours for the logs of different containers. So I went with the approach of randomly picking a colour. I like @cdrage's idea better and will switch to sequentially pick the colour.

pkg/odo/cli/logs/logs.go
dharmit: I have created a function called "ColorPicker" in the file you have mentioned and it's picking colour sequentially instead of a random manner. PTAL.

cdrage:
> > Each time I run `odo logs` the output is in a different color. Is that a feature?
> 
> @kadel we wanted different colours for the logs of different containers. So I went with the approach of randomly picking a colour. I like @cdrage's idea better and will switch to sequentially pick the colour.

I didn't mention anything about sequentially, but rather to make sure we use `GetStdout()` for the output since it'll automatically detect the correct terminal for colourization output. We've had problems in the past for colours not outputting correctly and instead showing ANSI codes.

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// variables
+	componentName string
+	contextDir    string
+	out           io.Writer
+}
+
+func NewLogsOptions() *LogsOptions {
+	return &LogsOptions{
+		out: log.GetStdout(),
```
cdrage, line 42: 👍  this is what I meant, thanks for implementing this!

.golangci.yaml
```diff
@@ -67,7 +67,6 @@ linters-settings:
     rules:
       # Conventions around error strings. See https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#error-strings
       - name: error-strings
-
```
cdrage, line 70: ?

cdrage (changes requested):

pkg/logs/logs.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	ownerReferences := resource.GetOwnerReferences()
+	// recursively check if ownerReference matches any of the resources' UID
+	for _, ownerReference := range ownerReferences {
+		return o.matchOwnerReferenceWithResources(ownerReference, resources)
+	}
+	return false, nil
+}
+
+func getGroupVersion(apiVersion string) (string, string) {
```
feloy, line 110: This function is not used anymore

feloy (changes requested):

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+		line = strings.Join([]string{containerName, line}, ": ")
+		_, err = fmt.Fprintf(out, line)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
dharmit: All the confusion was because you mentioned `scanner.Scanner` in your original comment. https://github.com/redhat-developer/odo/pull/5760#discussion_r883359977

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,192 @@
+---
+title: odo logs
+---
+
+`odo logs` displays the logs for all the containers odo created for the
```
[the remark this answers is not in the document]
dharmit: Thanks for pointing. Thanks a lot for creating a common pattern/template. 👍🏾 
I have made the changes, but not added 'Advanced Usage' since I couldn't think of putting anything in there. PTAL.

feloy:
@dharmit the architecture of the code and the behaviour lgtm

/approve

pkg/odo/cli/logs/logs.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func NewCmdLogs(name, fullname string) *cobra.Command {
+	o := NewLogsOptions()
+	logsCmd := &cobra.Command{
+		Use:   name,
+		Short: "Show logs of all containers of the component",
+		Long: `odo logs shows logs of all containers of the component, whether they are running in Dev mode or Deploy mode.
```
kadel, line 145: This description is misleading. Currently, this command only shows logs from containers running in Dev mode.

dharmit: I purposefully kept it this way at this particular place, because I didn't want to have to remember to modify it when support for showing Deploy mode containers' logs was added to `odo logs`. Kinda taking liberty of being in the alpha/beta stage. If you look at the [documentation added in this PR](https://deploy-preview-5760--odo-docusaurus-preview.netlify.app/docs/3.0.0/command-reference/logs/), it doesn't mention Deploy mode anywhere.

If you insist on rewording it to be accurate, I can do so, but I would prefer not to.

kadel: > I purposefully kept it this way at this particular place, because I didn't want to have to remember to modify it when support for showing Deploy mode containers' logs was added to odo logs.

We could release the next odo alpha/beta before the PR that will complete the implementation is merged, and then even more people would see that misleading help message. 

The odo users should be always on the first place. We should not cut corners to make it easier for ourself and create it harder for users to use odo.

If you are afraid that you will forget to update it, you can keep checklist of the things that you need to update in the original issue. 


>If you look at the [documentation added in this PR](https://deploy-preview-5760--odo-docusaurus-preview.netlify.app/docs/3.0.0/command-reference/logs/), it doesn't mention Deploy mode anywhere.

And that is part of the problem. You correctly omitted that from the docs but kept it in the help message, this will be confusing for users. 

>  Kinda taking liberty of being in the alpha/beta stage. 

We already have a good amount of people experimenting with odo v3, yes we are in alpha/beta stage we can still afford to break stuff, but saying in help message that it works in one way, but in reality not working is not good even for beta stage.


We don't have to merge complete implementation in one PR, but help messages and docs should always match what is actually implemented.

dharmit: Updated.

valaparthvi (changes requested):
I am yet to review the CLI layer code, but things lgtm otherwise, except the documentation changes.

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,189 @@
+---
+title: odo logs
+---
+
+## Description
+
```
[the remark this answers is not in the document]
dharmit: What's astray here? I checked `dev.md`, `list.md` and `list-namespaces.md`. I don't follow what you're trying to say.

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,189 @@
...
+2. Kubernetes component named `infinitepodone`
+3. Kubernetes component named `infinitedeployment`. As can be seen under `.
+   spec.template.spec.containers` for this particular component, it creates 
+   two containers for it.
+
+When you do `odo logs`, you should logs from all these containers. Each line 
+is prefixed with `<container-name>:` to easily distinguish which the 
+container the logs belong to. Since we named multiple containers in the 
+`devfile.yaml` as `container`, `odo logs` has distinguished these containers 
+as `container`, `container1` and `container2`:
```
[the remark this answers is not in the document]
feloy: To disambiguate a little bit more, you could use container[1] instead

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,188 @@
...
+            app: infinite
+        template:
+          metadata:
+            labels:
+              app: infinite
+          spec:
+            containers:
+            - name: container
+              image: docker.io/dharmit/infiniteloop
+            - name: container
```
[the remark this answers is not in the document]
dharmit: Thanks for the catch. It was a wrong example. I have fixed it and named the containers `main`.

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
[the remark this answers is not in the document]
dharmit: Makes sense. 97b28d3.

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,170 @@
...
+          spec:
+            containers:
+            - name: main
+              image: docker.io/dharmit/infiniteloop
+ ```
+When you do `odo dev`, odo creates pods for:
+1. The component named `node` itself. Containers for this are created using `.components.container`.
+2. Kubernetes component named `infinitepodone`
+3. Kubernetes component named `infinitedeployment`. As can be seen under `.spec.template.spec.containers` for this 
+   particular component, it creates two containers for it.
```
feloy, line 115:
```suggestion
   particular component, it creates one container for it.
```

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,170 @@
...
+
+When you run `odo logs`, you should see logs from all these containers. Each line is prefixed with 
+`<container-name>:` to easily distinguish which the container the logs belong to. Since we named multiple 
+containers in the `devfile.yaml` as `container`, `odo logs` has distinguished these containers as `container`,
```
feloy, line 119:
```suggestion
containers in the `devfile.yaml` as `main`, `odo logs` has distinguished these containers as `main`
```

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -0,0 +1,170 @@
...
+When you do `odo dev`, odo creates pods for:
+1. The component named `node` itself. Containers for this are created using `.components.container`.
+2. Kubernetes component named `infinitepodone`
+3. Kubernetes component named `infinitedeployment`. As can be seen under `.spec.template.spec.containers` for this 
+   particular component, it creates two containers for it.
+
+When you run `odo logs`, you should see logs from all these containers. Each line is prefixed with 
+`<container-name>:` to easily distinguish which the container the logs belong to. Since we named multiple 
+containers in the `devfile.yaml` as `container`, `odo logs` has distinguished these containers as `container`, 
+`container1` and `container2`:
```
feloy, line 120:
```suggestion
and `main1`:
```

feloy (approved):
Just a few fixes in the doc, else lgtm

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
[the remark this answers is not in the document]
valaparthvi: Yes that! My GoLand does soft/hard wraps and it annoys me. I assumed that's what happened with you but looks like it was intended.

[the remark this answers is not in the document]
dharmit: It wasn't intended, but in the past I used to hard wrap at 80 chars when I was using vim. So it went unnoticed to me that GoLand was doing the same.

PR #5773, opened 2022-05-30
odo describe binding

### LINKED ISSUE #5659: user should be able to display details about SB
## User Story
- As an odo user
- I want to be able to get details of odo binding
- So that I know what binding information is available in my application

## Acceptance Criteria
- [ ] `odo describe binding` should display details about all binding in the current devfile.yaml and in the cluster that belong to the current component
- [ ] `odo describe binding --name <binding name>` should display details about given binding only in current namespace
- [ ] json output
- [ ] support for both SB resources
  - servicebindings.binding.operators.coreos.com
  - servicebindings.servicebinding.io 

/kind user-story

## Example 
```
$ odo describe binding
ServiceBinding used by the current component:

Service Binding Name: myapp-mymongo
Services:
 - mymongo (PerconaServerMongoDB.psmdb.percona.com)
Bind as files: true
Detect binding resources: false
Available biding information: 
- /bindings/myapp-mymongo/username
- /bindings/myapp-mymongo/password

Service Binding Name: myapp-mymongo2
Service: 
 - mymongo (PerconaServerMongoDB.psmdb.percona.com)
Bind as files: false
Detect binding resources: false
Available biding information:
- PERCONASERVERMONGODB_PASSWORD
- PERCONASERVERMONGODB_USERNAME

# if service binding exists in devfile but not on the cluster
Service Binding Name: myapp-mymongo3
Services:
 - mymongo (PerconaServerMongoDB.psmdb.percona.com)
Bind as files: false
Detect binding resources: false
Available biding information: unknown
```

```
$ odo describe binding --name myapp-mymongo2
Service Binding Name: myapp-mymongo2
Services:
 -  mymongo (PerconaServerMongoDB.psmdb.percona.com)
Bind as files: false
Detect binding resources: false
Available biding information:
- PERCONASERVERMONGODB_PASSWORD
- PERCONASERVERMONGODB_USERNAME
```

feloy:
**What type of PR is this:**

/kind feature

Fixes #5659 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation 

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
@@ -2,21 +2,21 @@
 title: odo describe binding
 ---
 
-`odo describe binding` command is useful for getting information about a service binding. 
+## Description
+
+`odo describe binding` command is useful for getting information about service bindings.
 
 This command supports the service bindings added with the command `odo add binding` and bindings added manually to the Devfile, using a `ServiceBinding` resource from one of these apiVersion:
 - `binding.operators.coreos.com/v1alpha1`
 - `servicebinding.io/v1alpha3`
 
+## Running the Command
+
 There are 2 ways to describe a service binding:
 - [Describe with access to Devfile](#describe-with-access-to-devfile)
 - [Describe without access to Devfile](#describe-without-access-to-devfile)
 
-## Describe with access to Devfile
-
-```shell
-odo describe binding
-```
+### Describe with access to Devfile
 
 This command returns information extracted from the Devfile and, if possible, from the cluster.
 
@@ -31,9 +31,68 @@ For each of these resources, the following information is displayed:
 - if the variables are bound as files or as environment variables,
 - if the binding information is auto-detected.
 
-If the resources have been deployed to the cluster, the command also extracts information from the status of the resources to display information about the variables that can be used from the component.
+When the service binding are not deployed yet to the cluster:
 
-## Describe without access to Devfile
+```shell
+$ odo describe binding
+ServiceBinding used by the current component:
+
+Service Binding Name: my-nodejs-app-cluster-sample
+Services:
+ •  cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
+Bind as files: false
+Detect binding resources: true
+Available binding information: unknown
+
+Service Binding Name: my-nodejs-app-redis-standalone
+Services:
+ •  redis-standalone (Redis.redis.redis.opstreelabs.in)
+Bind as files: false
+Detect binding resources: true
+Available binding information: unknown
+
+Binding information for one or more ServiceBinding is not available because they don't exist on the cluster yet.
+Start "odo dev" first to see binding information.
+```
+
+When the resources have been deployed to the cluster, the command also extracts information from the status of the resources to display information about the variables that can be used from the component.
+
+```shell
+$ odo describe binding 
+ServiceBinding used by the current component:
+
+Service Binding Name: my-nodejs-app-cluster-sample-2
+Services:
+ •  cluster-sample-2 (Cluster.postgresql.k8s.enterprisedb.io)
+Bind as files: false
+Detect binding resources: true
+Available binding information:
+ •  CLUSTER_PASSWORD
+ •  CLUSTER_PROVIDER
+ •  CLUSTER_TLS.CRT
+ •  CLUSTER_TLS.KEY
+ •  CLUSTER_USERNAME
+ •  CLUSTER_CA.KEY
+ •  CLUSTER_CLUSTERIP
+ •  CLUSTER_HOST
+ •  CLUSTER_PGPASS
+ •  CLUSTER_TYPE
+ •  CLUSTER_CA.CRT
+ •  CLUSTER_DATABASE
+
+Service Binding Name: my-nodejs-app-redis-standalone
+Services:
+ •  redis-standalone (Redis.redis.redis.opstreelabs.in)
+Bind as files: false
+Detect binding resources: true
+Available binding information:
+ •  REDIS_CLUSTERIP
+ •  REDIS_HOST
+ •  REDIS_PASSWORD
+ •  REDIS_TYPE
+```
+
+### Describe without access to Devfile
 
 ```shell
 odo describe binding --name <component_name>
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
@@ -1,0 +1,48 @@
+---
+title: odo describe binding
+---
+
+`odo describe binding` command is useful for getting information about a service binding. 
+
+This command supports the service bindings added with the command `odo add binding` and bindings added manually to the Devfile, using a `ServiceBinding` resource from one of these apiVersion:
+- `binding.operators.coreos.com/v1alpha1`
+- `servicebinding.io/v1alpha3`
+
+There are 2 ways to describe a service binding:
+- [Describe with access to Devfile](#describe-with-access-to-devfile)
+- [Describe without access to Devfile](#describe-without-access-to-devfile)
+
+## Describe with access to Devfile
+
+```shell
+odo describe binding
+```
+
+This command returns information extracted from the Devfile and, if possible, from the cluster.
+
+The command lists the Kubernetes resources declared in the Devfile as a Kubernetes component,
+with the kind `ServiceBinding` and one of these apiVersion:
+- `binding.operators.coreos.com/v1alpha1`
+- `servicebinding.io/v1alpha3`
+
+For each of these resources, the following information is displayed:
+- the resource name,
+- the list of the services to which the component is bound using this service binding,
+- if the variables are bound as files or as environment variables,
+- if the binding information is auto-detected.
+
+If the resources have been deployed to the cluster, the command also extracts information from the status of the resources to display information about the variables that can be used from the component.
+
+## Describe without access to Devfile
+
+```shell
+odo describe binding --name <component_name>
+```
+
+The command extracts information from the cluster.
+
+The command searches for a resource in the current namespace with the given name, the kind `ServiceBinding` and one of these apiVersion:
+- `binding.operators.coreos.com/v1alpha1`
+- `servicebinding.io/v1alpha3`
+
+If a resource is found, it displays information about the service binding and the variables that can be used from the component.
--- a/pkg/api/binding.go
+++ b/pkg/api/binding.go
@@ -26,7 +26,7 @@ type ServiceBindingStatus struct {
 
 // ServiceBindingFromBinding returns a common api.ServiceBinding structure
 // from a ServiceBinding.binding.operators.coreos.com/v1alpha1
-func ServiceBindingFromBinding(binding bindingApi.ServiceBinding) (ServiceBinding, error) {
+func ServiceBindingFromBinding(binding bindingApi.ServiceBinding) ServiceBinding {
 
 	var dstSvcs []specApi.ServiceBindingServiceReference
 	for _, srcSvc := range binding.Spec.Services {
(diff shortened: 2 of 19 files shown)


pkg/binding/binding.go
```diff
@@ -146,3 +155,184 @@ func (o *BindingClient) GetServiceInstances() (map[string]unstructured.Unstructu
...
+// from group binding.operators.coreos.com/v1alpha1
+func (o *BindingClient) getStatusFromBinding(name string) (*api.ServiceBindingStatus, error) {
+	bindingSB, err := o.kubernetesClient.GetBindingServiceBinding(name)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	secretName := bindingSB.Status.Secret
```
rm3l, line 281: Putting it here for the record...

As discussed, `secretName` might be empty if the ServiceBinding instance is in a non-ready state. As seen together, the ServiceBinding status needs to be checked before reading the secret.

Otherwise, this might cause issues like this when reading the secret afterwards:

```
❯ odo describe binding
 ✗  unable to get the secret &Secret{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Data:map[string][]byte{},Type:,StringData:map[string]string{},Immutable:nil,}: resource name may not be empty
```

This happened after binding to a Redis instance, added via the [Redis Operator](https://operatorhub.io/operator/redis-operator) and the Redis CRD:
```
❯ kubectl get servicebindings my-miniflux-app-redis-standalone
NAME                               READY   REASON            AGE
my-miniflux-app-redis-standalone   False   ProcessingError   149m

❯ kubectl get servicebindings my-miniflux-app-redis-standalone -o=jsonpath='{.status}' | jq   
{
  "conditions": [
    {
      "lastTransitionTime": "2022-05-31T11:36:09Z",
      "message": "redisSecret is not found",
      "reason": "ErrorReadingBinding",
      "status": "False",
      "type": "CollectionReady"
    },
    {
      "lastTransitionTime": "2022-05-31T11:36:09Z",
      "message": "redisSecret is not found",
      "reason": "ProcessingError",
      "status": "False",
      "type": "Ready"
    }
  ],
  "secret": ""
}
```

rm3l (changes requested):

pkg/api/binding.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type ServiceBindingStatus struct {
+	BindingFiles   []string `json:"bindingsFiles,omitempty"`
+	BindingEnvVars []string `json:"bindingEnvVars,omitempty"`
+}
+
+// ServiceBindingFromBinding returns a common api.ServiceBinding structure
+// from a ServiceBinding.binding.operators.coreos.com/v1alpha1
+func ServiceBindingFromBinding(binding bindingApi.ServiceBinding) (ServiceBinding, error) {
```
rm3l, line 29: Do we really need the `error` return type in the signature? We are always returning `nil` as `error` here. Removing it would simplify callers' code by removing an error handling branch.

pkg/api/binding.go
```diff
@@ -0,0 +1,63 @@
...
+			Services:               dstSvcs,
+			DetectBindingResources: binding.Spec.DetectBindingResources,
+			BindAsFiles:            binding.Spec.BindAsFiles,
+		},
+	}, nil
+}
+
+// ServiceBindingFromSpec returns a common api.ServiceBinding structure
+// from a ServiceBinding.servicebinding.io/v1alpha3
+func ServiceBindingFromSpec(spec specApi.ServiceBinding) (ServiceBinding, error) {
```
rm3l, line 54: Same comment about the need to have an `error` return type in the signature, since we are always returning `nil` as `error`.

pkg/binding/binding.go
rm3l: With hindsight, shouldn't we check that the secret does exist as well (since we are getting the binding information from the secret)? I just manually deleted the secret in my local cluster (thinking it would maybe get recreated, but it was not). Then I hit the same issue again:

```
❯ odo describe binding
 ✗  unable to get the secret &Secret{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Data:map[string][]byte{},Type:,StringData:map[string]string{},Immutable:nil,}: secrets "my-miniflux-postgresql-db-28d34d55" not found
```

I think this should allow us to write a test case for this, if the secret is missing, regardless of the service binding status.

rm3l (changes requested):

rm3l:
> > With hindsight, shouldn't we check that the secret does exist as well (since we are getting the binding information from the secret)? I just manually deleted the secret in my local cluster (thinking it would maybe get recreated, but it was not)
> 
> I'm not sure. If the `InjectionReady` condition of the SB is true but the secret is not present, that seems an operational problem to me, and the user should be aware of it. Also, if the user has deleted the secret, odo should not behave normally IMO, but return an error.

Okay - makes sense. Thanks.

rm3l (approved):

docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
```diff
@@ -0,0 +1,48 @@
...
+
+This command supports the service bindings added with the command `odo add binding` and bindings added manually to the Devfile, using a `ServiceBinding` resource from one of these apiVersion:
+- `binding.operators.coreos.com/v1alpha1`
+- `servicebinding.io/v1alpha3`
+
+There are 2 ways to describe a service binding:
+- [Describe with access to Devfile](#describe-with-access-to-devfile)
+- [Describe without access to Devfile](#describe-without-access-to-devfile)
+
+## Describe with access to Devfile
```
cdrage, line 15: For command reference, we are trying to have a consistent format between all documentation. Can you split these sections up similar to: https://odo.dev/docs/3.0.0/command-reference/dev ?

Ex. Section for Description, Running the Command and Devfile usage.

cdrage (changes requested):

docs/website/versioned_docs/version-3.0.0/command-reference/describe-binding.md
feloy: Hi made the changes, thanks

pkg/binding/binding.go
```diff
@@ -146,3 +157,178 @@ func (o *BindingClient) GetServiceInstances() (map[string]unstructured.Unstructu
...
+	}
+
+	secretName := bindingSB.Status.Secret
+	secret, err := o.kubernetesClient.GetSecret(secretName, o.kubernetesClient.GetCurrentNamespace())
+	if err != nil {
+		return nil, err
+	}
+
+	if bindingSB.Spec.BindAsFiles {
+		bindings := make([]string, 0, len(secret.Data))
```
dharmit, line 279: This declaration could be made outside the `if` block so that it doesn't need to be done twice.

pkg/kclient/binding.go
```diff
@@ -121,7 +122,45 @@ func NewServiceBindingObject(bindingName string, bindAsFiles bool, deploymentNam
 	}
 }
 
-func mappingContainsBKS(bindableObjects []*meta.RESTMapping, bks sboApi.BindableKindsStatus) bool {
+// GetBindingServiceBinding returns a ServiceBinding from group binding.operators.coreos.com/v1alpha1
+func (c Client) GetBindingServiceBinding(name string) (bindingApi.ServiceBinding, error) {
```
dharmit, line 126: Since it's returning a ServiceBinding from Binding, do you think it would be better to rename this to `GetServiceBindingFromBinding`? And then the `name` parameter is the name of the Binding using which the method tries to find the Binding.

pkg/kclient/binding.go
```diff
@@ -121,7 +122,45 @@ func NewServiceBindingObject(bindingName string, bindAsFiles bool, deploymentNam
...
+	if c.DynamicClient == nil {
+		return bindingApi.ServiceBinding{}, nil
+	}
+
+	u, err := c.GetDynamicResource(bindingApi.GroupVersionResource, name)
+	if err != nil {
+		return bindingApi.ServiceBinding{}, err
+	}
+
+	var result bindingApi.ServiceBinding
```
dharmit, line 136: Since this would create a zero value `binding.ServiceBinding`, would declaring it at the beginning of the method and then using it in the above two cases when `DynamicClient` is `nil` or when there's error in `c.GetDynamicResource` call be simpler?

pkg/kclient/binding.go
```diff
@@ -121,7 +122,45 @@ func NewServiceBindingObject(bindingName string, bindAsFiles bool, deploymentNam
...
+	if c.DynamicClient == nil {
+		return specApi.ServiceBinding{}, nil
+	}
+
+	u, err := c.GetDynamicResource(specApi.GroupVersionResource, name)
+	if err != nil {
+		return specApi.ServiceBinding{}, err
+	}
+
+	var result specApi.ServiceBinding
```
dharmit, line 155: Similar question as in `GetBindingServiceBinding` method for `var result` declaration.

pkg/odo/cli/describe/binding.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+// BindingRecommendedCommandName is the recommended binding sub-command name
+const BindingRecommendedCommandName = "binding"
+
+var describeBindingExample = ktemplates.Examples(`
+# Describe the bindings in the current devfile
+%[1]s
+
+# Describe a binding in the cluster
```
dharmit, line 28: Nitpick
```suggestion
# Describe a binding on the cluster
```

pkg/odo/cli/describe/binding.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return nil
+}
+
+func (o *BindingOptions) Validate() (err error) {
+	return nil
+}
+
+func (o *BindingOptions) Run(ctx context.Context) error {
+	if len(o.nameFlag) == 0 {
```
dharmit, line 78: Nothing wrong, but in the same file we are once checking using `o.nameFlag=='"` and next time using `len(o.nameFlag)==0`. I understand they are the same thing.

dharmit (changes requested):
Few comments/questions, few change requests.

pkg/api/binding.go
```diff
@@ -0,0 +1,63 @@
...
+type ServiceBindingSpec struct {
+	Services               []specApi.ServiceBindingServiceReference `json:"services"`
+	DetectBindingResources bool                                     `json:"detectBindingResources"`
+	BindAsFiles            bool                                     `json:"bindAsFiles"`
+}
+
+type ServiceBindingStatus struct {
+	BindingFiles   []string `json:"bindingsFiles,omitempty"`
+	BindingEnvVars []string `json:"bindingEnvVars,omitempty"`
+}
```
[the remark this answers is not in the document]
feloy: `Status` is the term used for the part of the Kubernetes resources containing this kind of information (as opposed to `Spec`).

pkg/binding/binding.go
```diff
@@ -146,3 +157,178 @@ func (o *BindingClient) GetServiceInstances() (map[string]unstructured.Unstructu
...
+		sb := api.ServiceBindingFromBinding(bindingSB)
+		sb.Status, err = o.getStatusFromBinding(bindingSB.Name)
+		if err != nil {
+			return api.ServiceBinding{}, err
+		}
+		return sb, nil
+	}
+	if err != nil && !kerrors.IsNotFound(err) {
+		return api.ServiceBinding{}, err
+	}
```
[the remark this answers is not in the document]
feloy: I wanted to make it different, to highlight that there is a missing case when `err != nil && kerrors.IsNotFound(err)`.

I don't know how to express it clearly.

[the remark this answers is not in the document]
dharmit: > I don't know how to express it clearly.

TBH, I found your original code more readable after I put my suggested code in the above comment. I think there's no case that only checks if error is not `nil`, so we can live with this. It's obviously more readable than my suggestion.

pkg/kclient/binding.go
feloy: The prefix `Binding` here (as opposed to `Spec` in the following method), indicates the group of the ServiceBinding (binding.operators.coreos.com/v1alpha1 as opposed to servicebinding.io/v1alpha3 alias Spec)

I find this order more readable as the method
GetBindingServiceBinding
returns a struct
bindingApi.ServiceBinding

and
GetSpecServiceBinding
returns 
specApi.ServiceBinding

pkg/kclient/binding.go
feloy: I'm not sure. For the return with err, if I read, 
```
return result, err
```

I have to pass several seconds to understand what is the value of result.

With `return bindingApi.ServiceBinding{}, err`, I'm sure instantly it is the zero value

feloy:
Thanks @dharmit for your review. I made changes based on your feedback, or replied to some comments

dharmit:
Since the request made by @cdrage is addressed already, and my change requests have also been addressed, LGTM.

/lgtm

PR #5790, opened 2022-06-06
Add unit test for odo add binding

### LINKED ISSUE #5627: Create binding between devfile component and existing service

valaparthvi:
**What type of PR is this:**
/kind tests

This PR adds unit test cases for pkg/binding/binding.go

Fixes #5627 

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/binding/binding_test.go
+++ b/pkg/binding/binding_test.go
@@ -1,0 +1,308 @@
+package binding
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/library/pkg/devfile/parser"
+	"github.com/devfile/library/pkg/testingutil/filesystem"
+	"github.com/golang/mock/gomock"
+	servicebinding "github.com/redhat-developer/service-binding-operator/apis/binding/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/redhat-developer/odo/pkg/kclient"
+	odoTestingUtil "github.com/redhat-developer/odo/pkg/testingutil"
+)
+
+func TestBindingClient_GetFlags(t *testing.T) {
+	type args struct {
+		flags map[string]string
+	}
+	tests := []struct {
+		name string
+		args args
+		want map[string]string
+	}{
+		{
+			name: "service and name flags are present",
+			args: args{flags: map[string]string{"service": "redisService", "name": "mybinding", "v": "9"}},
+			want: map[string]string{"service": "redisService", "name": "mybinding"},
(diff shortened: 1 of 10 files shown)


valaparthvi:
/retest-required
IBM tests pass.

pkg/binding/binding_test.go
```diff
@@ -0,0 +1,308 @@
...
+	Version: "v1",
+}
+
+var clusterGVR = schema.GroupVersionResource{
+	Group:    clusterGVK.Group,
+	Version:  clusterGVK.Version,
+	Resource: "clusters",
+}
+
+var deploymentGVR = metav1.GroupVersionResource{
```
feloy, line 264: You could use

```
var deploymentGVR = appsv1.SchemeGroupVersion.WithResource("deployments")
```

pkg/binding/binding_test.go
```diff
@@ -0,0 +1,308 @@
...
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("AddBinding() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+var clusterGVK = schema.GroupVersionKind{
```
feloy, line 252: You can first define 
```
var clusterGV = schema.GroupVersion{
	Group:   "postgresql.k8s.enterprisedb.io",
	Version: "v1",
}
```
Then
```
var clusterGVK = clusterGV.WithKind("Cluster")
var clusterGVR = clusterGV.WithResource("clusters")
```

feloy: Could you place these variables at the beginning of the file?

PR #5795, opened 2022-06-06
Change ephemeral default to false

kadel:
**What type of PR is this:**

/kind feature

Feel free to use other [labels](https://github.com/redhat-developer/odo/labels) as needed. However one of the above labels must be present or the PR will not be reviewed. This instruction is for reviewers as well.
-->

fixes https://github.com/redhat-developer/odo/issues/5241

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


feloy:
> All the integration tests are now running with non-ephemeral volumes for component's project directory.
> 
> IMO, this has some effect on the limits of resources on the cluster, and could slow down the tests (with 32 tests running in parallel, we can easily find Pod in Pending state for lack of memory)
> 
> Perhaps do we want to switch the preference to ephemeral during the tests?

@kadel, what's your opinion?

kadel:
> All the integration tests are now running with non-ephemeral volumes for component's project directory.
> 
> IMO, this has some effect on the limits of resources on the cluster, and could slow down the tests (with 32 tests running in parallel, we can easily find Pod in Pending state for lack of memory)
> 
> Perhaps do we want to switch the preference to ephemeral during the tests?

This is a good point. 
I'll change our tests to use ephemeral by default.
There already is at least one test that verifies that `ephemeral=false` works correctly (https://github.com/redhat-developer/odo/blob/main/tests/integration/devfile/cmd_dev_test.go#L269)

feloy:
> There already is at least one test that verifies that `ephemeral=false` works correctly (https://github.com/redhat-developer/odo/blob/main/tests/integration/devfile/cmd_dev_test.go#L269)

You will need to update this test to now force the `odo set preference` command, or it will refuse because the value is now already set:

```
[odo] ? ephemeral is already set. Do you want to override it in the config (y/N)
I0616 07:53:42.259922   22593 common.go:17] Encountered an error processing prompt: EOF
[odo] Aborted by the user
```

feloy:
/override ci/prow/unit

Tests pass on IBM Cloud

PR #5802, opened 2022-06-07
bump version to v3.0.0-alpha3

kadel:


feloy:
Flaky test, we are not checking if the namespace already exists when getting a random name
```
[odo]  X  unable to create new project: unable to create new project cmd-namespace-test74cnd: project.project.openshift.io "cmd-namespace-test74cnd" already exists
```

PR #5809, opened 2022-06-09
Migrate to ginkgo v2 

### LINKED ISSUE #5791: Upgrade ginkgo to version 2.0
Ginkgo 2.0 is now available and offers additional reporting functionality

## Why is this needed?
Upgrading to Ginkgo v2.0 offers: 
- new GinkgoWriter methods to write output that is only made visible if a test fails 
- improved reporting (machine-readable reports, generation of custom reports) ability to control output

An [upgrade guide and new feature list ](https://github.com/onsi/ginkgo/blob/ver2/docs/MIGRATING_TO_V2.md) is available.

anandrkskd:
**What type of PR is this:**

/kind tests

Fixes #5791

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


tests/integration/cmd_namespace_test.go
```diff
@@ -178,7 +178,7 @@ ComponentSettings:
 			})
 		})
 
-		Describe("list "+commandName, func() {
+		Describe("list "+commandName, FlakeAttempts(3), func() {
```
feloy, line 181: We should avoid using this, as it will result in more flaky tests, and so much longer execution times
I proposed another solution in https://github.com/redhat-developer/odo/pull/5837

tests/e2escenarios/e2e_devfile_test.go
```diff
@@ -64,7 +64,7 @@ var _ = Describe("odo devfile supported tests", func() {
 		//		stopChannel <- true
 	}
 
-	Context("odo debug support for devfile components", func() {
+	Context("odo debug support for devfile components", FlakeAttempts(3), func() {
```
feloy, line 67: We should avoid using this IMO

feloy (changes requested):

tests/integration/cmd_namespace_test.go
anandrkskd: Understood! Will revert these changes.

Makefile
```diff
@@ -116,8 +116,7 @@ goget-tools:
 
 .PHONY: goget-ginkgo
 goget-ginkgo:
-	@echo "This is no longer used."
-	@echo "Ginkgo can be executed directly from this repository using command '$(RUN_GINKGO)'"
+	 go install github.com/onsi/ginkgo/v2/ginkgo@latest
```
rm3l, line 119: Just curious, using the vendored version no longer works with this upgrade?
While I understand the previous behavior of running Ginkgo from the vendor folder (making sure the right version of Ginkgo is running), I think installing the `latest` version might be dangerous.

anandrkskd: When we are trying to use it the old way using vendor, it was not recognizing the new flags, and was not honoring the old flags.
I think we can use vendor after we merge [PR](https://github.com/redhat-developer/odo/pull/5833) in.

anandrkskd: I will update that with a specific version, thanks.

anandrkskd: I think it will make this PR huge. I can create a new PR replacing your #5833. We can merge this PR after the cleanup PR if merged?
WDYT @feloy

anandrkskd:
@feloy the validate test is failing with 
```go run -mod=vendor github.com/frapposelli/wwhrd check -q
ERRO[0001] Found Non-Approved license      license=UNKNOWN package=github.com/segmentio/backo-go
FATA[0001] Exiting: Non-Approved license found          
exit status 1
```
And this used inside segmentio package, Not sure what should be my next step.

feloy:
> @feloy the validate test is failing with
> 
> ```go
> ERRO[0001] Found Non-Approved license      license=UNKNOWN package=github.com/segmentio/backo-go
> FATA[0001] Exiting: Non-Approved license found          
> exit status 1
> ```
> 
> And this used inside segmentio package, Not sure what should be my next step.

This is because we are now including the v1.0.0 instead of at a specific commit:

```

File: scripts/generate-mod.sh [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+# for odo
+go mod edit -require github.com/devfile/api/v2@v2.0.0-20220309195345-48ebbf1e51cf
+go mod edit -require github.com/devfile/library@v1.2.1-0.20220602130922-85a4805bd59c
+go mod edit -require github.com/devfile/registry-support/index/generator@v0.0.0-20220222194908-7a90a4214f3e
+go mod edit -require github.com/devfile/registry-support/registry-library@v0.0.0-20220504150710-21de53798172
+go mod edit -require github.com/kubernetes-sigs/service-catalog@v0.3.1
+go mod edit -require github.com/redhat-developer/alizer/go@v0.0.0-20220530162645-f27a60be88a1
+go mod edit -require github.com/redhat-developer/service-binding-operator@v1.0.1-0.20211222115357-5b7bbba3bfb3
+go mod edit -require github.com/onsi/ginkgo/v2@v2.1.4
+go mod edit -require github.com/segmentio/backo-go@v1.0.1-0.20200129164019-23eae7c10bd3
```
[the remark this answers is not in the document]
anandrkskd: added

.ibm/pipelines/kubernetes-tests.sh
```diff
@@ -12,6 +12,7 @@ cleanup_namespaces
 
 (
     set -e
+    make goget-ginkgo
```
rm3l, line 15: Since we are still running Ginkgo from the `vendor` folder, maybe this is no longer necessary.

.ibm/pipelines/openshift-tests.sh
```diff
@@ -13,6 +13,7 @@ cleanup_namespaces
 
 (
     set -e
+    make goget-ginkgo
```
rm3l, line 16: Same comment: this is no longer necessary.

scripts/generate-mod.sh
```diff
@@ -0,0 +1,30 @@
...
+go mod edit -require github.com/onsi/ginkgo/v2@v2.1.4
+go mod edit -require github.com/segmentio/backo-go@v1.0.1-0.20200129164019-23eae7c10bd3 # fixing this revision because of the missing license in the latest released version
+go mod edit -replace gopkg.in/segmentio/analytics-go.v3=github.com/segmentio/analytics-go/v3@v3.2.1
+go mod edit -replace github.com/apcera/gssapi=github.com/openshift/gssapi@v0.0.0-20161010215902-5fb4217df13b # for oc
+
+go get -u github.com/openshift/oc@openshift-clients-4.11.0-202204020828
+
+
+
+go mod tidy -compat=1.17 # why?
```
rm3l, line 28: Trying to understand the `why?`.. @feloy Can we elaborate a little bit on this? A comment in this file would be helpful for all of us..

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -16,7 +16,7 @@ else
 fi
 make bin
 mkdir -p $GOPATH/bin
-make goget-ginkgo
+GOFLAGS= make goget-ginkgo
```
rm3l, line 19: Same comment: Running `make goget-ginkgo` is no longer necessary, I think.

scripts/generate-mod.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
rm3l, line 1: @feloy Unless I've missed something about the usage of this script, I'm wondering if it wouldn't make sense to run this as part of CI, to detect cases where the `go.mod` file would be edited directly without updating this script (e.g, Dependabot PRs or someone updating/adding a new dependency). This way, we would be aware of editing this file if there is a diff.
But I guess it might not be straightforward especially where this script might potentially update dependencies with `go get -u`..

.ibm/pipelines/kubernetes-tests.sh
anandrkskd: removed

scripts/generate-mod.sh
feloy: @dharmit I really don't know, that's why I added a why? here meaning "TODO: understand why `go mod tidy` fails and the option `-compat=1.17` is necessary".

scripts/generate-mod.sh
```diff
@@ -0,0 +1,30 @@
...
+go mod edit -require github.com/devfile/library@v1.2.1-0.20220602130922-85a4805bd59c
+go mod edit -require github.com/devfile/registry-support/index/generator@v0.0.0-20220222194908-7a90a4214f3e
+go mod edit -require github.com/devfile/registry-support/registry-library@v0.0.0-20220504150710-21de53798172
+go mod edit -require github.com/kubernetes-sigs/service-catalog@v0.3.1
+go mod edit -require github.com/redhat-developer/alizer/go@v0.0.0-20220530162645-f27a60be88a1
+go mod edit -require github.com/redhat-developer/service-binding-operator@v1.0.1-0.20211222115357-5b7bbba3bfb3
+go mod edit -require github.com/onsi/ginkgo/v2@v2.1.4
+go mod edit -require github.com/segmentio/backo-go@v1.0.1-0.20200129164019-23eae7c10bd3 # fixing this revision because of the missing license in the latest released version
+go mod edit -replace gopkg.in/segmentio/analytics-go.v3=github.com/segmentio/analytics-go/v3@v3.2.1
+go mod edit -replace github.com/apcera/gssapi=github.com/openshift/gssapi@v0.0.0-20161010215902-5fb4217df13b # for oc
```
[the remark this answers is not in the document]
feloy: IMO we cannot, because this one is not used directly in the odo sources, but in one of the dependencies

scripts/generate-mod.sh
feloy: I agree that I would make sense. I remember we discussd to put in place a script to check these kind of generated code (mocks at least); that would be part of this work

scripts/generate-mod.sh
dharmit: Ah, ok. I thought it was Anand who added the comment. My bad.

I was doing some reading around this after noticing the code comment and my response to Armel's comment was based on what I understood. I mainly referred to this [stackoverflow answer](https://stackoverflow.com/a/71974997/395670) and below excerpt from [Go 1.17 release notes](https://go.dev/doc/go1.17#go-command):
> The -compat flag allows that version to be overridden to support older (or only newer) versions, up to the version specified by the go directive in the go.mod file. To tidy a go 1.17 module for Go 1.17 only, without saving checksums for (or checking for consistency with) Go 1.16:

scripts/generate-mod.sh
rm3l: Indeed, I also remember a discussion around such checks. I'll try to find a related issue (or create a new one) to track this.

rm3l (approved):

PR #5815, opened 2022-06-10
ignore dynamic resource when not found

vinny-sabatini:
**What type of PR is this:**
/kind bug

If there are cluster scoped resources that are reported in the
`BindableKinds`, they will be included when trying to retrieve Service
Instances. When this happens and we try to search for those resources
in a namespace, Kubernetes will return the error that they are not found.

We can not assume all developers would have access to these
cluster scoped resources, so we can not safely run additional checks to
verify they are cluster scoped resources. However, if nothing is found,
it is safe to assume they can not be bound to as the developer does not
have access to them.

N/A

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- Add service binding annotations to a cluster scoped CRD
- Run `odo add binding`
- Ensure you can see the Service Bindings you would expect

### DIFF
--- a/pkg/kclient/dynamic.go
+++ b/pkg/kclient/dynamic.go
@@ -40,7 +40,7 @@ func (c *Client) CreateDynamicResource(resource unstructured.Unstructured) error
 }
 
 // ListDynamicResource returns an unstructured list of instances of a Custom
-// Resource currently deployed in the active namespace of the cluster
+// Resource currently deployed in the active namespace of the cluster.
 func (c *Client) ListDynamicResources(gvr schema.GroupVersionResource) (*unstructured.UnstructuredList, error) {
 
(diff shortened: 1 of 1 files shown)


vinny-sabatini:
One thing I considered doing instead of just returning the empty list upon the error was actually trying to see if there were cluster scoped resources for the given `BindableKind`, and then either return the list if they are accessible or return an empty list if they are still not found.

I decided to take the simpler route because I was not sure if you wanted to/are able to support "cluster scoped" BindableKinds. Please let me know if that assumption is wrong and I can try to make that change.

feloy:
/lgtm

Thanks Vincent for this fix. This is greatly appreciated

PR #5817, opened 2022-06-13
odo add binding - Bind as files UI update

### LINKED ISSUE #5770: `odo add binding` - Bind as files UI update
In interactive flow odo ask users 

"Bind as files?" Yes/no

For users this might not be obvious what this actually means.

It would be better to have an option (select) with two options.

How to expose binding information?
- As files.
- As environment variables.

valaparthvi:

**What type of PR is this:**
/kind feature

This PR adds a more understandable question when asking how to service should be bound to the component.

Fixes #5770 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation (https://deploy-preview-5817--odo-docusaurus-preview.netlify.app/docs/3.0.0/command-reference/add-binding)

1. `mkdir /tmp/101; cd /tmp/101`
2. Setup a bindable operator and spawn a service. For example - https://operatorhub.io/operator/cloud-native-postgresql
3. `odo init --name restapi --devfile go --starter go-starter`
4. `odo add binding`

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -5,7 +5,7 @@ title: odo add binding
 ## Description
 The `odo add binding` command can add a link between an operator-backed service and a component. odo uses the Service Binding Operator to create this link. Running this command will make the necessary changes to the Devfile, and once pushed to the cluster, it creates an instance of the `ServiceBinding` resource.
 
-Currently, it only allows connecting to the operator-backed services which support binding via the Service Binding Operator.
+Currently, it only allows connecting to the Operator-backed services which support binding via the Service Binding Operator.
 To know about the Operators supported by the Service Binding Operator, read its [README](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators).
 
 ## Running the Command
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -39,11 +39,11 @@ odo add binding --name mybinding --service myRedisService.Redis
 ```
 
 #### Understanding Bind as Files
-To connect your component with a service, you need to store some data(e.g. username, password, host address) on your component's container.
-If the service is bound as files, this data will be written to a file and stored on the container, else it will be exported to Environment Variables inside the container.
+To connect your component with a service, you need to store some data (e.g. username, password, host address) on your component's container.
+If the service is bound as files, this data will be written to a file and stored on the container, else it will be injected as Environment Variables inside the container.
 
 Note that every piece of data is stored in its own individual file or environment variable.
-For example, if your data includes a username, and password, then 2 separate files, or 2 environment variables will be created to store them both.
+For example, if your data includes a username and password, then 2 separate files, or 2 environment variables will be created to store them both.
 
 #### Formats supported by the `--service` flag
 The `--service` flag supports the following formats to specify the service name:
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -3,7 +3,7 @@ title: odo add binding
 ---
 
 ## Description
-The `odo add binding` command can add a link between an operator-backed service and a component. odo uses the Service Binding Operator to create this link. Running this command will make the necessary changes to the Devfile, and once pushed to the cluster, it creates an instance of the `ServiceBinding` resource.
+The `odo add binding` command adds a link between an Operator-backed service and a component. odo uses the [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator/) to create this link. Running this command will modify the Devfile, and once pushed (using `odo dev`) to the cluster, it creates an instance of the `ServiceBinding` resource.
 
 Currently, it only allows connecting to the Operator-backed services which support binding via the Service Binding Operator.
 To know about the Operators supported by the Service Binding Operator, read its [README](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators).
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -17,7 +17,7 @@ To know about the operators supported by the Service Binding Operator, read thei
 ### Interactive Mode
 In the interactive mode, you will be guided to choose:
 * a service from the list of bindable service instances as supported by the Service Binding Operator,
-* option to bind the service as a file; see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this.
+* option to bind the service as a file (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this),
 * a name for the binding.
 
 ```shell
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -28,7 +28,7 @@ odo add binding
 ### Non-interactive mode
 In the non-interactive mode, you will have to specify the following required information through the command-line:
 * `--service` flag to specify the service you want to bind to,
-* `--name` flag to specify a name for the binding; see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this.
+* `--name` flag to specify a name for the binding (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this)
 * `--bind-as-files` flag to specify if the service should be bound as a file; this flag is set to true by default.
 
 
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -2,41 +2,50 @@
 title: odo add binding
 ---
 
+## Description
 The `odo add binding` command can add a link between an operator-backed service and a component. odo uses the Service Binding Operator to create this link. Running this command will make the necessary changes to the Devfile, and once pushed to the cluster, it creates an instance of the `ServiceBinding` resource.
 
 Currently, it only allows connecting to the operator-backed services which support binding via the Service Binding Operator.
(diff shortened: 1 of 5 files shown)


pkg/binding/asker/survey_asker.go
```diff
@@ -40,14 +40,19 @@ func (s *Survey) AskServiceBindingName(defaultName string) (string, error) {
...
+	var answer string
 	err := survey.AskOne(question, &answer)
 	if err != nil {
-		return false, err
+		return true, err
 	}
-	return answer, nil
+
+	var bindAsFiles bool
+	if answer == "Bind as Files" {
```
feloy, line 54: IMO, you should avoid testing the literal value directly. You could pass an int as AskOne to get the index of the result, and test on the index.

valaparthvi: Right, I keep forgetting about using the index.

valaparthvi: I do not want to return the int value here, I want to evaluate the `bindAsFiles`'s bool here. So, even if I use int to store the value of user's answer, I would rely on the options to maintain. Something like below - 
```go
func (o *Survey) AskBindAsFiles() (bool, error) {
	options := []string{"Bind as Files", "Bind as Environment Variables"}
	question := &survey.Select{
		Message: "How do you want to bind the service?",
		Options: options,
	}
	var answer int
	err := survey.AskOne(question, &answer)
	if err != nil {
		return true, err
	}

	var bindAsFiles bool
	if answer == 0 {
		bindAsFiles = true
	}
	return bindAsFiles, nil
}
```
This doesn't seem correct to me.
Is this what you were suggesting?

valaparthvi: Const sounds like a better idea, will do that.

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
 title: odo add binding
 ---
 
+## Description
 The `odo add binding` command can add a link between an operator-backed service and a component. odo uses the Service Binding Operator to create this link. Running this command will make the necessary changes to the Devfile, and once pushed to the cluster, it creates an instance of the `ServiceBinding` resource.
```
dharmit, line 6: I know it's not a part of this PR, but I was just reading the rendered doc and found it relevant. It's your call to change it here or separately.
```suggestion
The `odo add binding` command adds a link between an Operator-backed service and a component. odo uses the [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator/) to create this link. Running this command will modify the Devfile, and once pushed (using `odo dev`) to the cluster, it creates an instance of the `ServiceBinding` resource.
```
Specifying `odo dev` in the bracket because we don't have `odo push` anymore.

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
 
 Currently, it only allows connecting to the operator-backed services which support binding via the Service Binding Operator.
```
dharmit, line 8:
```suggestion
Currently, it only allows connecting to the Operator-backed services which support binding via the Service Binding Operator.
```

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
 To know about the operators supported by the Service Binding Operator, read their [README](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators).
```
dharmit, line 9:
```suggestion
To know about the Operators supported by the Service Binding Operator, read its [README](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators).
```

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
 
-## Pre-requisites
+## Running the Command
+
+### Pre-requisites
 * A directory containing a Devfile; if you don't have one, see [odo init](init.md) on obtaining a devfile.
 * A cluster with the Service Binding operator installed, along with the operator whose service you need to bind to
```
dharmit, line 15:
```suggestion
* A cluster with the Service Binding Operator installed, along with the Operator whose service you want to bind to
```

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
+# Add binding between a service, and the component present in the working directory in the interactive mode
 odo add binding
 ```
 
-## Non-interactive mode
+### Non-interactive mode
 In the non-interactive mode, you will have to specify the following required information through the command-line:
 * `--service` flag to specify the service you want to bind to,
-* `--name` flag to specify a name for the binding,
+* `--name` flag to specify a name for the binding; see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this.
```
dharmit, line 31: IMO, it makes more sense this way.
```suggestion
* `--name` flag to specify a name for the binding (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this)
```

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
+### Pre-requisites
 * A directory containing a Devfile; if you don't have one, see [odo init](init.md) on obtaining a devfile.
 * A cluster with the Service Binding operator installed, along with the operator whose service you need to bind to
 
-## Interactive Mode
+### Interactive Mode
 In the interactive mode, you will be guided to choose:
 * a service from the list of bindable service instances as supported by the Service Binding Operator,
-* option to bind the service as a file,
+* option to bind the service as a file; see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this.
```
dharmit, line 20: IMO, it makes more sense this way.
```suggestion
* option to bind the service as a file (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this),
```

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -2,41 +2,50 @@
...
 odo add binding --name mybinding --service myRedisService.Redis
 ```
 
-### Formats supported by the `--service` flag
+#### Understanding Bind as Files
+To connect your component with a service, you need to store some data(e.g. username, password, host address) on your component's container.
+If the service is bound as files, this data will be written to a file and stored on the container, else it will be exported to Environment Variables inside the container.
+
+Note that every piece of data is stored in its own individual file or environment variable.
+For example, if your data includes a username, and password, then 2 separate files, or 2 environment variables will be created to store them both.
```
dharmit, line 46:
```suggestion
#### Understanding Bind as Files
To connect your component with a service, you need to store some data (e.g. username, password, host address) on your component's container.
If the service is bound as files, this data will be written to a file and stored on the container, else it will be injected as Environment Variables inside the container.

Note that every piece of data is stored in its own individual file or environment variable.
For example, if your data includes a username and password, then 2 separate files, or 2 environment variables will be created to store them both.
```

dharmit (changes requested):
Only doc review. Besides the requested/suggested changes, I feel that we should document the need to either have a running `odo dev` in a separate terminal, or execute `odo dev` after adding a binding if it's not running in a separate terminal.

pkg/binding/asker/survey_asker.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/AlecAivazis/survey/v2"
 )
 
+const (
+	BindAsFiles  = "Bind As Files"
+	BindAsEnvVar = "Bind As Environment Variables"
+)
```
dharmit, line 12: I find creating a type before `const` declaration to be more helpful since it enables us to use `switch..case` statement unlike `if..else`. It's not an issue here, but when we have more items/possibilities, it comes in handy. Something like:

```go
type BindAs string

const (
	BindAsFiles  BindAs = "Bind As Files"
	BindAsEnvVar BindAs = "Bind As Environment Variables"
)
```
(Not a request to change anything; just sharing my understanding. I'm using it in my upcoming `odo logs` to switch between `--dev`, `--deploy`, none)

pkg/odo/cli/add/binding/binding.go
```diff
@@ -19,7 +19,7 @@ import (
 const BindingRecommendedCommandName = "binding"
 
 var addBindingExample = ktemplates.Examples(`
-# Add binding between service named 'myservice' and the component present in the working directory in the interactive mode
+# Add binding between a service, and the component present in the working directory in the interactive mode
```
dharmit, line 22: IMO (and I am not an English expert), a `,` before "and" is used when there are a list of items. Here you are categorically talking about "service" and "component". Hence, no need of a `,` here. Happy to be corrected.
```suggestion
# Add binding between a service and the component present in the working directory in the interactive mode
```

tests/interactive/cmd_add_binding_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+var _ = Describe("odo init interactive command tests", func() {
+
+	var commonVar helper.CommonVar
+	var serviceName string
+
+	// This is run before every Spec (It)
+	var _ = BeforeEach(func() {
+		if helper.IsKubernetesCluster() {
+			Skip("Operators have not been setup on Kubernetes cluster yet. Remove this once the issue has been fixed.")
```
dharmit, line 25: Really? Why is that the case? I mean, what's blocking us from setting up Operators on k8s cluster?

dharmit (changes requested):
Code review. Looks good, only minor suggestions.

pkg/binding/asker/survey_asker.go
valaparthvi: I'll keep this in mind, but for now, I do not think we need to follow this convention.

valaparthvi:
>I feel that we should document the need to either have a running odo dev in a separate terminal, or execute odo dev after adding a binding if it's not running in a separate terminal.

I don't think that has to be the case. `odo add binding` does not require for the component to be pushed on the cluster, it simply adds data to the devfile, you can run `odo dev` later on to create it on the cluster. I think the suggestion you made in https://github.com/redhat-developer/odo/pull/5817#discussion_r896426197 should suffice.

Issue #5535, opened 2022-03-09
odo preference - cleanup


kadel:
there should be the following set of subcommands for `odo preference` command

- [x] `odo preference set` - sets value for single value item (`UpdateNotification`, `Timeout`, `PushTimeout`, `RegistryCacheTime`, `Ephemeral`, `ConsentTelemetry`) (#5822 )
- [x] `odo preference unset` - unsets value (restores it to default) for single value item (`UpdateNotification`, `Timeout`, `PushTimeout`, `RegistryCacheTime`, `Ephemeral`, `ConsentTelemetry`) (#5822 )
- [x] `odo preference add` - adds value to array items (`registry`) (#5873 )
- [x] `odo preference remove` - removes value from array items (`registry`) (#5873 )
- [x] `odo preference view` -  view all preferences, and move `registry list` to `view` (#5850)
- [x] review name of the preference items, if needed rename it  and add description (#5822 )
- [x] Drop odo preference registry update; if needed, user can always remove and add registry (#5853)
## examples 
```
$ odo preference set ConsentTelemetry true

Value of 'ConsentTelemetry' preference was set to 'true'
```

```
$ odo preference unset ConsentTelemetry

Value of 'ConsentTelemetry' was removed from preferences. Its default value will be used.
```

```
$ odo preference add registry StagingRegistry https://registry.stage.devfile.io```

New Devfile registry was added to odo.
```

```
$ odo preference remove registry StagingRegistry

Devfile registry 'StagingRegistry' was removed from odo.
```

```
$ odo preference view

Preference parameters:
PARAMETER             VALUE
UpdateNotification    true (default)
Timeout               
PushTimeout           
RegistryCacheTime     
Ephemeral             
ConsentTelemetry      false


Devfile registries:
NAME                       URL                                   
StagingRegistry            https://registry.stage.devfile.io     
DefaultDevfileRegistry     https://registry.devfile.io           
``

dharmit:
> [ ] `odo preference set` - sets value for single value item (`UpdateNotification`, `Timeout`, `PushTimeout`, `RegistryCacheTime`, `Ephemeral`, `ConsentTelemetry`)

We have been using `int` for all timeouts so far. I suggest using `time.Duration` for it. We can use `DurationVar` instead of `IntVar` from cobra point of view. Two benefits that I can think of by doing this:
* Not having to convert `int` to `time.Duration` as the values are now in `time.Duration` by default. 
[...]

PR #5822, opened 2022-06-14
Preference cleanup (1/n)

valaparthvi:
**What type of PR is this:**
/kind cleanup

This PR cleans the preference pkg with the following changes:
1. Use `time.Duration` instead of `int` for time related preferences.
2. Modify the success messages for `set`, and `unset`.
3. 

Fixes part of #5535 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

1. `odo preference set pushtimeout 4m`
2. `odo preference unset pushtimeout`


pkg/odo/cli/preference/view_test.go
```diff
@@ -36,10 +37,11 @@ func TestView(t *testing.T) {
 		return
 	}
 
+	timePointer := 10 * time.Second
```
feloy, line 40: Please name it `timeValue` instead

pkg/util/defaults.go
```diff
@@ -1,13 +1,17 @@
...
+
+func GetBoolOrDefault(ptr *bool, defaultValue bool) bool {
 	if ptr == nil {
 		return defaultValue
 	}
 	return *ptr
 }
 
-func GetBoolOrDefault(ptr *bool, defaultValue bool) bool {
+func GetTimeDefault(ptr *time.Duration, defaultValue time.Duration) time.Duration {
```
feloy, line 14: Note that this package provides the same functions as the ones defined here:
https://github.com/kubernetes/utils/blob/master/pointer/pointer.go

`DurationDeref`

valaparthvi:
/override ci/prow/unit

IBM tests pass.

feloy:
When executing:
```
$ odo preference set PushTimeout 1
 ✗  unable to set "pushtimeout" to "1", value must be an integer
```

the message doesn't seem correct. It should be instead "value should be a duration"

feloy:
I can see this pattern in several places, which gives the wrong result:

```
time.Duration(o.clientset.PreferenceClient.GetTimeout()) * time.Second
```

Is should be: 

```
o.clientset.PreferenceClient.GetTimeout()
```

pkg/odo/cli/version/version.go
```diff
@@ -63,7 +64,7 @@ func (o *VersionOptions) Complete(cmdline cmdline.Cmdline, args []string) (err e
 			// checking the value of timeout in preference
 			var timeout time.Duration
 			if o.clientset.PreferenceClient != nil {
-				timeout = time.Duration(o.clientset.PreferenceClient.GetTimeout()) * time.Second
+				timeout = o.clientset.PreferenceClient.GetTimeout() * time.Second
```
feloy, line 67: I don't think you need to multiply by 1s (same in line 70)

valaparthvi: You're right, I missed this one.

rm3l (changes requested):

pkg/preference/implem.go
```diff
@@ -253,32 +254,23 @@ func (c *preferenceInfo) SetConfiguration(parameter string, value string) error
...
 			}
 			c.OdoSettings.Timeout = &typedval
 
 		case "pushtimeout":
-			typedval, err := strconv.Atoi(value)
+			typedval, err := time.ParseDuration(value)
 			if err != nil {
-				return fmt.Errorf("unable to set %q to %q, value must be an integer", parameter, value)
-			}
-			if typedval < 0 {
```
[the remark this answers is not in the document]
valaparthvi: Looks like `time.ParseDuration` does support all kinds of values.
```go
// ParseDuration parses a duration string.
// A duration string is a possibly signed sequence of
// decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "-1.5h" or "2h45m".
// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
func ParseDuration(s string) (Duration, error) {
```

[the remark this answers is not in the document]
rm3l: > I was not aware of the --. 
>
Yes, it is a way to tell the CLI parser to stop further flag processing. Not sure, but I think this is a POSIX requirement, respected by most CLI parsers out there (at least the POSIX-compliant ones).

> I am surprised time.ParseDuration allows negative values,
>
Since a duration measures the difference between two time points, it can be negative if the end is before the start. To me, this is a valid case of allowing negative values.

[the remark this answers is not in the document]
rm3l: > Looks like `time.ParseDuration` does support all kinds of values.
> 
> ```go
> // ParseDuration parses a duration string.
> // A duration string is a possibly signed sequence of
> // decimal numbers, each with optional fraction and a unit suffix,
> // such as "300ms", "-1.5h" or "2h45m".
> // Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
> func ParseDuration(s string) (Duration, error) {
> ```

I don't know yet how we use all those Durations in `odo`, but maybe we could set a minimum acceptable value for each Duration. I don't think it makes sense to accept PushTimeout in microseconds for example :-) What do you think?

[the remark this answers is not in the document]
valaparthvi: What do you think about  `1 * time.Second` as a minimum acceptable value for preferences that expect time.Duration?
@feloy @rm3l

valaparthvi:
I intend on adding unit tests if https://github.com/redhat-developer/odo/pull/5822/commits/397ae32838424921cc6ee07897049605779c8d76 is acceptable.

pkg/preference/implem.go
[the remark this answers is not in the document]
rm3l: > What do you think about `1 * time.Second` as a minimum acceptable value for preferences that expect time.Duration? 
>
That's fine to me. Setting a too-high min value would give less flexibility to users, I think.

feloy (changes requested):

pkg/preference/implem.go
```diff
@@ -253,32 +254,23 @@ func (c *preferenceInfo) SetConfiguration(parameter string, value string) error
...
 		case "timeout":
-			typedval, err := strconv.Atoi(value)
-			if err != nil {
-				return fmt.Errorf("unable to set %q to %q", parameter, value)
-			}
-			if typedval < 0 {
-				return errors.New("cannot set timeout to less than 0")
+			typedval, err := time.ParseDuration(value)
+			if err != nil || typedval < minimumDurationValue {
+				return fmt.Errorf("unable to set %q to %q, value must be a positive Duration (e.g. 4s, 5m, 1h); minimum value: 1s", parameter, value)
```
rm3l, line 259: Wondering if the error we return here should not also wrap the original `err` returned by `time.ParseDuration`, if any. This way, it would be reported to the user. What do you think?

Because if I enter a duration but with a unit suffix (like [days](https://github.com/golang/go/issues/17767)) not supported by `time.ParseDuration`, I find our error message not to be much helpful in understanding the issue:
```
❯ odo preference set PushTimeout 1d 
 ✗  unable to set "pushtimeout" to "1d", value must be a positive Duration (e.g. 4s, 5m, 1h); minimum value: 1s
```

pkg/preference/variables.go
```diff
@@ -60,13 +64,13 @@ const (
 )
 
 // TimeoutSettingDescription is human-readable description for the timeout setting
-var TimeoutSettingDescription = fmt.Sprintf("Timeout (in seconds) for OpenShift server connection check (Default: %d)", DefaultTimeout)
+var TimeoutSettingDescription = fmt.Sprintf("Timeout (in Duration) for cluster server connection check (Default: %s)", DefaultTimeout)
 
 // PushTimeoutSettingDescription adds a description for PushTimeout
-var PushTimeoutSettingDescription = fmt.Sprintf("PushTimeout (in seconds) for waiting for a Pod to come up (Default: %d)", DefaultPushTimeout)
+var PushTimeoutSettingDescription = fmt.Sprintf("PushTimeout (in Duration) for waiting for a Pod to come up (Default: %s)", DefaultPushTimeout)
```
rm3l, line 70: Surely nitpicking ^^, but would it be possible to change the format string? I found it a bit weird that this shows up as `4m0s` in the help message. Same thing for the default values of the other durations, except for `Timeout`:

```
❯ odo help preference                                                                                                                                                                                                                                                            
Modifies odo specific configuration settings within the global preference file.                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                 
Available Global Parameters:                                                                                                                                                                                                                                                     
 ConsentTelemetry - If true, odo will collect telemetry for the user's odo usage (Default: false)                                                                                                                                                                                
                    For more information: https://developers.redhat.com/article/tool-data-collection                                                                                                                                                                             
 Ephemeral - If true, odo will create an emptyDir volume to store source code (Default: true)                                                                                                                                                                                    
 PushTimeout - PushTimeout (in Duration) for waiting for a Pod to come up (Default: 4m0s)                                                                                                                                                                                        
 RegistryCacheTime - For how long (in Duration) odo will cache information from the Devfile registry (Default: 15m0s)                                                                                                                                                            
 Timeout - Timeout (in Duration) for cluster server connection check (Default: 1s)                                                                                                                                                                                               
 UpdateNotification - Flag to control if an update notification is shown or not (Default: true)                                                                                                                                                                     

```

valaparthvi: That format is how golang parses duration. I'll have to see if that can be modified, but another way would be to parse everything into `second`.

pkg/preference/implem.go
```diff
@@ -253,32 +254,23 @@ func (c *preferenceInfo) SetConfiguration(parameter string, value string) error
 		switch p {
 
 		case "timeout":
-			typedval, err := strconv.Atoi(value)
-			if err != nil {
-				return fmt.Errorf("unable to set %q to %q", parameter, value)
-			}
-			if typedval < 0 {
-				return errors.New("cannot set timeout to less than 0")
+			typedval, err := time.ParseDuration(value)
```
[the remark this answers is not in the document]
feloy: Yes, that's about the migration plan I was talking about

valaparthvi:
@rm3l @feloy
https://github.com/redhat-developer/odo/pull/5822/commits/3f53cdfc748a49b2ef17a5e016554453e8ffa5e2 includes
- a warning as a migration plan from v2 to v3, see the integration test on how to reproduce it.
- a better error message when setting the preference value fails due to an incompatible format or any other cause

rm3l (approved):
Thanks for the changes!

PR #5826, opened 2022-06-14
Fix parametrized integration tests

### LINKED ISSUE #5814: Dynamically generated specs with for loop do not work with ginkgo
We recently started using `for` loops for test cases that share the same test logic but require different parameters. But it does not seem to work as we would have hoped.
For example, consider the following test:
https://github.com/redhat-developer/odo/blob/4ba943986dd85e773626659567c73ce69db55213/tests/integration/cmd_namespace_test.go#L26-L49

The parameters `[]string{"namespace", "project"}` works well for the `Describe` and `It` blocks' text, but not with the logic of the test. For some reason, it always takes "project" into account while testing.

Following is the output of a test run(tests have been explicitly set to fail in order to see the output) - https://gist.github.com/valaparthvi/3d178d3e495a4c42746501cc51c33080. Notice that for both the test runs, it uses `odo create project`.

A solution to writing parametrized test cases would be to use `DescribeTable`. See https://onsi.github.io/ginkgo/#table-specs.

I was unable to find `DescribeTable` with the ginkgo pkg that we have at the moment.
I noticed that the ginkgo pkg we use is actually github.com/openshift/onsi-ginkgo which happens to be a very very old copy of github.com/onsi/ginkgo. And, the version that we are using also happens to be quite old, the latest that I see of openshift/onsi-ginkgo is 4.9, while we're still on 4.7. I am not sure why we decided to use it in the first place, but perhaps we can move to using the onsi/ginkgo pkg for our test cases.

valaparthvi:
**What type of PR is this:**
/kind tests

This PR makes an attempt at fixing dynamically generated specs. We have been using for loops with our integration tests, but it doesn't work as expected. The fix introduced with this PR attempts at fixing that.

Fixes #5814 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

1. Debug tests, make sure correct data is taken into account.

### DIFF
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -22,13 +22,13 @@ var _ = Describe("odo create/delete/list/set namespace/project tests", func() {
 	AfterEach(func() {
 		helper.CommonAfterEach(commonVar)
 	})
-
 	for _, commandName := range []string{"namespace", "project"} {
-
+		// this is a workaround to ensure that the for loop works with `It` blocks
+		commandName := commandName
 		Describe("create "+commandName, func() {
+
 			namespace := fmt.Sprintf("%s-%s", helper.RandString(4), commandName)
 
-			// TODO: Remove for loop; find a way to use DescribeTable
 			It(fmt.Sprintf("should successfully create the %s", commandName), func() {
 				helper.Cmd("odo", "create", commandName, namespace, "--wait").ShouldPass()
 				defer func(ns string) {
--- a/tests/integration/devfile/cmd_delete_test.go
+++ b/tests/integration/devfile/cmd_delete_test.go
@@ -36,7 +36,7 @@ var _ = Describe("odo delete command tests", func() {
 	}{
 		{
 			title:       "a component is bootstrapped",
-			devfileName: "devfile-deploy-with-multiple-resources",
+			devfileName: "devfile-deploy-with-multiple-resources.yaml",
 		},
 		{
 			title:       "a component is bootstrapped using a devfile.yaml with URI-referenced Kubernetes components",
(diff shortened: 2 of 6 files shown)


tests/integration/cmd_namespace_test.go
```diff
@@ -22,13 +22,13 @@ var _ = Describe("odo create/delete/list/set namespace/project tests", func() {
 	AfterEach(func() {
 		helper.CommonAfterEach(commonVar)
 	})
-
 	for _, commandName := range []string{"namespace", "project"} {
-
+		// this is a workaround to ensure that the for loop works with `It` blocks
+		commandName := commandName
```
feloy, line 27: You are creating shadowed variables this way. It would be better to use different names for the two variables

valaparthvi: > This will generate several Its - one for each category. Note that you must assign a copy of the loop variable to a local variable (that's what category := category is doing) - otherwise the It closure will capture the mutating loop variable and all the specs will run against the last element in the loop. It is idiomatic to give the local copy the same name as the loop variable.(https://onsi.github.io/ginkgo/#dynamically-generating-specs)

It is by design if I am understanding it correctly.

rm3l: I was also surprised to see that this was not reported by `golangci-lint` (and `govet`) as a potential issue.
I guess this is particularly fine for `govet` as we are shadowing the variable with itself.
It would be detected by `golangci-lint` if we assigned a different value, I think.

tests/integration/devfile/cmd_delete_test.go
```diff
@@ -36,7 +36,7 @@ var _ = Describe("odo delete command tests", func() {
 	}{
 		{
 			title:       "a component is bootstrapped",
-			devfileName: "devfile-deploy-with-multiple-resources",
+			devfileName: "devfile-deploy-with-multiple-resources.yaml",
```
rm3l, line 39: Thanks for fixing this, and adding the missing Devfile 👍🏿

rm3l (commented):
/approve

Thanks for fixing these tests, @valaparthvi!

PR #5841, opened 2022-06-20
[WIP] Add support for composite run/debug commands

### LINKED ISSUE #5054: Add support for composite run commands
Odo currently doesn't support run commands that are composite.

- [ ] this needs to support `parallel` flag as well

simple example with composite run command for testing
```yaml
schemaVersion: 2.1.0
metadata:
  name: nodejs

components:
  - container:
      endpoints:
        - name: http-3000
          targetPort: 3000
    name: runtime

commands:
  - id: install
    exec:
      commandLine: npm install
      component: runtime
      workingDir: /projects
  - id: start
    exec:
      workingDir: /projects
      commandLine: npm start
      component: runtime
  - id: run
    composite:
      commands:
        - install
        - start
      group:
        isDefault: true
        kind: run
```

/kind feature
/priority high

rm3l:
**What type of PR is this:**
/kind feature
/area feature
/area devfile-spec

Fixes #5054

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 


rm3l:
> Following discussion during Backlog Grooming call, it could be interesting to have integration tests with several containers, where some steps of run/debug are executed in container1, others in container2

Okay, I'll add that. I was actually manually testing this way. But it makes sense to have this as a test case.

feloy:
When the composite commands are executed, are they with redirection to /proc/1/fd? I cannot see the output of the composite commands in logs

rm3l:
> When the composite commands are executed, are they with redirection to /proc/1/fd? I cannot see the output of the composite commands in logs

🤔 Yes, exec commands part of the composite run/debug should have their output redirected to `/proc/1/fd/*`. For example, this is what I got with the Devfile I am using:

```
❯ odo logs
runtime: npm WARN nodejs-starter@1.0.0 No repository field.
runtime: 
runtime: audited 64 packages in 0.691s
runtime: found 0 vulnerabilities
runtime: 
runtime: 
runtime: > nodejs-starter@1.0.0 start /projects
runtime: > node server.js
runtime: 
runtime: App started on PORT 3000
say-hello-container: intentional-error
```

Can you show the Devfile you are using?

feloy:
In fact, I see that the logs are missing for the *build* commands

feloy:
Or, if you want to test building in a container and running in anorther one, with a shared volume:

```

**Comment by rm3l** [comment]
> In fact, I see that the logs are missing for the _build_ commands

That was a behavior I also noticed when working on Supervisord removal.. Build command logs are only displayed locally, when running `odo dev -v 2`.
It makes sense to include them too as part of the logs, especially if they take long to run (case of Maven projects builds). With #5846, it will make it easier to follow the progress as well.. I'll add this while I am on it.

**Comment by rm3l** [comment]
> Or, if you want to test building in a container and running in anorther one, with a shared volume:
> 
Thanks!

> It seems that we need to set explicitely the `tail -f /dev/null` command to some containers
>
I'll test with this Devfile, but you had to do that with the containers used by the build command, right?

**Comment by feloy** [comment]
> > Or, if you want to test building in a container and running in anorther one, with a shared volume:
> 
> Thanks!
> 
> > It seems that we need to set explicitely the `tail -f /dev/null` command to some containers
> 
> I'll test with this Devfile, but you had to do that with the containers used by the build command, right?

Yes, you are right, only for build containers. Here is a more appropriate devfile to illustrate (with 2 sleeperbuild and sleeperrun containers instead of one common one):

```
commands:
- id: build
  composite:
    parallel: true
    commands:
    - build1
    - build2
    group:
      isDefault: true
      kind: build
- exec:
    commandLine: (echo sleeping for 2 seconds before build; sleep 2; echo done) > /proc/1/fd/1
    component: sleeperbuild
    workingDir: ${PROJECT_SOURCE}
  id: build1
- exec:
    commandLine: GOCACHE=${PROJECT_SOURCE}/.cache go build main.go; mv main /artifacts/; echo build done 
    component: builder
    workingDir: ${PROJECT_SOURCE}
  id: build2
- composite:
    parallel: true
    commands:
    - run1
    - run2
    group:
      isDefault: true
      kind: run
  id: run
- exec:
    commandLine: (echo sleeping for 2 seconds before run; sleep 2; echo done) 
    component: sleeperrun
    workingDir: ${PROJECT_SOURCE}
  id: run1
- exec:
    commandLine: ./main 
    component: runtime
    workingDir: /artifacts
  id: run2
components:
- container:
    endpoints:
    - name: http
      targetPort: 8080
    memoryLimit: 1024Mi
    mountSources: false
    volumeMounts:
    - name: artifacts
      path: /artifacts
  name: runtime
- container:
    memoryLimit: 1024Mi
    mountSources: true
    volumeMounts:
    - name: artifacts
      path: /artifacts
    command: ["tail"]
    args: ["-f", "/dev/null"]
  name: builder
- container:
    memoryLimit: 256Mi
    mountSources: false
    command: ["tail"]
    args: ["-f", "/dev/null"]
  name: sleeperbuild
- container:
    memoryLimit: 256Mi
    mountSources: false
  name: sleeperrun
- name: artifacts
  volume:
    ephemeral: true
    size: 1Gi
metadata:
  description: Stack with the latest Go version
  displayName: Go Runtime
  language: go
  name: my-go-app
  projectType: go
  tags:
  - Go
  version: 1.0.0
schemaVersion: 2.1.0
starterProjects:
- git:
    checkoutFrom:
      revision: main
    remotes:
  name: go-starter
```

**Comment by feloy** [comment]
> Following the grooming discussion about supporting exec commands for Deploy, I think we should actually do the same thing for all container components, regardless of the kind of the command. What do you think? (But this would be done in the scope of #5701)

Yes, I think so also

File: pkg/devfile/adapters/kubernetes/component/commandhandler.go [hunk 1]
```diff
@@ -44,7 +44,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 		switch status {
 		case remotecmd.Starting:
 			_ = log.SpinnerNoSpin(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id))
-		case remotecmd.Stopped:
+		case remotecmd.Stopped, remotecmd.Errored:
 			if err != nil {
 				klog.V(2).Infof("error while running background command: %v", err)
 			}
```
feloy, line 50: I propose this, so we can see when an application is terminated:

```
	var spin *log.Status
	startHandler := func(status remotecmd.RemoteProcessStatus, stdout []string, stderr []string, err error) {
		switch status {
		case remotecmd.Starting:
			spin = log.SpinnerNoSpin(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id))
		case remotecmd.Stopped, remotecmd.Errored:
			if spin != nil {
				spin.End(status == remotecmd.Stopped)
				spin = nil
			}
			if err != nil {
				klog.V(2).Infof("error while running background command: %v", err)
			}
		}
	}
```

output:
```
 ✓  Waiting for Kubernetes resources [8s]
 ✓  Syncing files into the container [256ms]
 ✓  Building your application in container on cluster (command: build1) [2s]
 ✓  Building your application in container on cluster (command: build2) [983ms]
 •  Executing the application (command: run1)  ...
 ✓  Executing the application (command: run1) [2s]
 •  Executing the application (command: run2)  ...

 - Forwarding from 127.0.0.1:40001 -> 8080

```

feloy (commented):
/approve

Just a comment about the spinner for the app

pkg/devfile/adapters/kubernetes/component/commandhandler.go
rm3l: That's a good point, from the user perspective. I'll make the changes you suggested.

feloy:
/override ci/prow/unit

Tests pass on IBM cloud

PR #5846, opened 2022-06-20
Adds support to follow/tail/stream odo logs

### LINKED ISSUE #5715: `odo logs` part 2 - command should have `--follow` flag
`--follow` flag will enable automatic tailing of stderr and stdout logs from all containers

- [ ] should stream logs from all the containers together into one output, lines will be ordered by time, no container grouping

```
$ odo logs --follow
```

dharmit:
**What type of PR is this:**

/kind feature

It allows the users to follow/stream the logs for the components.

Fixes #5715

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation 

Use the devfile modified in this PR and run following set of commands:
1. `odo logs` / `odo logs --follow`
2. `odo logs --dev` / `odo logs --dev --follow`
3. `odo logs --deploy` / `odo logs --deploy --follow`
4. `odo logs --dev --deploy` / `odo logs --dev --deploy --follow` (this one should fail.)

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
@@ -23,6 +23,7 @@ container name to help differentiate between the containers. In the output, you 
 
 It also supports `--follow` flag which allows you to follow/tail/stream the logs of the containers. It works by using 
 the same commands as above albeit, with a `--follow` flag:
-* Use `odo logs --dev` to follow the logs for the containers created by `odo dev` command.
-* Use `odo logs --deploy` to follow the logs for the containers created by `odo deploy` command.
-* Use `odo logs` (without any flag) to follow the logs of all the containers created by both `odo dev` and `odo deploy`.
+* Use `odo logs --dev --follow` to follow the logs for the containers created by `odo dev` command.
+* Use `odo logs --deploy --follow` to follow the logs for the containers created by `odo deploy` command.
+* Use `odo logs --follow` (without `--dev` or `--deploy`) to follow the logs of all the containers created by both `odo 
+  dev` and `odo deploy`.
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
@@ -20,3 +20,9 @@ init` command.
 Note that if multiple containers are named the same (for example, `main`), the `odo logs` output appends a number to 
 container name to help differentiate between the containers. In the output, you will see containers named as `main`, 
 `main[1]`, `main[2]`, so on and so forth.
+
+It also supports `--follow` flag which allows you to follow/tail/stream the logs of the containers. It works by using 
+the same commands as above albeit, with a `--follow` flag:
+* Use `odo logs --dev` to follow the logs for the containers created by `odo dev` command.
+* Use `odo logs --deploy` to follow the logs for the containers created by `odo deploy` command.
+* Use `odo logs` (without any flag) to follow the logs of all the containers created by both `odo dev` and `odo deploy`.
--- a/pkg/odo/cli/logs/logs.go
+++ b/pkg/odo/cli/logs/logs.go
@@ -207,13 +207,14 @@ func printLogs(containerName string, rd io.ReadCloser, out io.Writer, colour col
 	for scanner.Scan() {
 		line := scanner.Text()
 		mu.Lock()
+		defer mu.Unlock()
 		color.Set(colour)
+		defer color.Unset()
+
 		_, err := fmt.Fprintln(out, containerName+": "+line)
 		if err != nil {
 			return err
 		}
-		color.Unset()
-		mu.Unlock()
 	}
 
 	return nil
--- a/pkg/odo/cli/logs/logs.go
+++ b/pkg/odo/cli/logs/logs.go
@@ -137,7 +137,6 @@ func (o *LogsOptions) Run(ctx context.Context) error {
 	uniqueContainerNames := map[string]struct{}{}
 	errChan := make(chan error)
 	wg := sync.WaitGroup{}
-	var mu sync.Mutex
 	for _, entry := range containersLogs {
 		for container, logs := range entry {
 			uniqueName := getUniqueContainerName(container, uniqueContainerNames)
@@ -146,22 +145,13 @@ func (o *LogsOptions) Run(ctx context.Context) error {
 			if o.follow {
 				l := logs
 				wg.Add(1)
-				go func(out io.Writer, follow bool) {
+				go func(out io.Writer) {
 					defer wg.Done()
-					if !follow {
-						// ensure that only one go routine does printLogs at a time; this is helpful when logs are longer
-						// than just a few lines in that logs for each container are printed before starting to print those
-						// of the next container
-						mu.Lock()
-					}
 					err = printLogs(uniqueName, l, out, colour)
 					if err != nil {
 						errChan <- err
 					}
-					if !follow {
-						mu.Unlock()
-					}
-				}(o.out, o.follow)
+				}(o.out)
 			} else {
 				err = printLogs(uniqueName, logs, o.out, colour)
 				if err != nil {
--- a/pkg/odo/cli/logs/logs.go
+++ b/pkg/odo/cli/logs/logs.go
@@ -206,12 +206,15 @@ func printLogs(containerName string, rd io.ReadCloser, out io.Writer, colour col
 
 	for scanner.Scan() {
 		line := scanner.Text()
-		mu.Lock()
-		defer mu.Unlock()
-		color.Set(colour)
-		defer color.Unset()
+		err := func() error {
+			mu.Lock()
+			defer mu.Unlock()
+			color.Set(colour)
+			defer color.Unset()
 
-		_, err := fmt.Fprintln(out, containerName+": "+line)
+			_, err := fmt.Fprintln(out, containerName+": "+line)
+			return err
+		}()
 		if err != nil {
 			return err
 		}
--- a/pkg/odo/cli/logs/logs.go
+++ b/pkg/odo/cli/logs/logs.go
@@ -130,7 +130,7 @@ func (o *LogsOptions) Run(ctx context.Context) error {
 		// 1. user specifies --dev flag, but the component's running in Deploy mode
 		// 2. user specified --deploy flag, but the component's running in Dev mode
 		// 3. user passes no flag, but component is running in neither Dev nor Deploy mode
-		fmt.Fprintf(o.out, "no containers running in the specified mode for the component %q", o.componentName)
+		fmt.Fprintf(o.out, "no containers running in the specified mode for the component %q\n", o.componentName)
 		return nil
 	}
 
--- a/pkg/odo/cli/logs/logs.go
+++ b/pkg/odo/cli/logs/logs.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 
 	odolabels "github.com/redhat-developer/odo/pkg/labels"
 
@@ -42,6 +43,7 @@ type LogsOptions struct {
 	// flags
 	devMode    bool
 	deployMode bool
+	follow     bool
 }
 
 type logsMode string
@@ -119,7 +121,7 @@ func (o *LogsOptions) Run(ctx context.Context) error {
 	default:
 		mode = odolabels.ComponentAnyMode
 	}
-	containersLogs, err = o.clientset.LogsClient.GetLogsForMode(mode, o.componentName, o.Context.GetProject())
+	containersLogs, err = o.clientset.LogsClient.GetLogsForMode(mode, o.componentName, o.Context.GetProject(), o.follow)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


pkg/odo/cli/logs/logs.go
```diff
@@ -133,17 +135,40 @@ func (o *LogsOptions) Run(ctx context.Context) error {
...
+				}
+			}(o.follow)
 		}
 	}
-
+	select {
+	case err := <-errChan:
+		return err
+	default:
+		// do nothing
```
dharmit, line 169: Without this `default` block, `odo logs` (without follow) will be stuck/blocked on getting something from `errChan` channel and the command won't exit.

tests/integration/devfile/cmd_logs_test.go
```diff
@@ -66,6 +67,29 @@ var _ = Describe("odo logs command tests", func() {
...
+				It("should successfully follow logs of running component", func() {
+					var linesOfLogs int
+					Consistently(func() bool {
+						logs := logsSession.OutContents()
+						if len(string(logs)) < linesOfLogs {
+							return false
+						}
+						linesOfLogs = len(logs)
+						return true
+					}, 20*time.Second, 5).Should(BeTrue())
```
dharmit, line 90: The way this test and subsequent `--follow` tests are written is that we aren't really bothered about what's being printed on stdout, but we care if the number of lines between successive runs is increasing.

The container image being used is `docker.io/dharmit/infiniteloop` (I am perfectly OK to use a more centralized container image). All it does is print a message on stdout every 5 seconds. Its Dockerfile is:
```Dockerfile
FROM fedora

CMD while true; do echo "`date` - this is infinite for loop"; sleep 1; done
```

So between the checks, we check if there are more lines in the current logs than previous one.

dharmit: > ```dockerfile
> FROM fedora
> 
> CMD while true; do echo "`date` - this is infinite for loop"; sleep 1; done
> ```

I've modified this to:
```dockerfile
FROM alpine

CMD while true; do echo "`date` - this is infinite while loop"; sleep 5; done
```

pkg/odo/cli/logs/logs.go
```diff
@@ -133,17 +135,41 @@ func (o *LogsOptions) Run(ctx context.Context) error {
...
-				return err
-			}
+			l := logs
+			wg.Add(1)
+			go func(out io.Writer, follow bool) {
+				defer wg.Done()
+				colour := log.ColorPicker()
+				if !follow {
+					// ensure that only one go routine does printLogs at a time; this is helpful when logs are longer
+					// than just a few lines in that logs for each container are printed before starting printing those
```
valaparthvi, line 152:
```suggestion
					// than just a few lines in that logs for each container are printed before starting to print those
```

pkg/odo/cli/logs/logs.go
valaparthvi: > Without this `default` block, `odo logs` (without follow) will be stuck/blocked on getting something from `errChan` channel and the command won't exit.

Can you add this comment in the code?

dharmit: Sure. Added.

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
```diff
@@ -19,4 +19,10 @@ init` command.
 
 Note that if multiple containers are named the same (for example, `main`), the `odo logs` output appends a number to 
 container name to help differentiate between the containers. In the output, you will see containers named as `main`, 
-`main[1]`, `main[2]`, so on and so forth.
\ No newline at end of file
+`main[1]`, `main[2]`, so on and so forth.
+
+It also supports `--follow` flag which allows you to follow/tail/stream the logs of the containers. It works by using 
+the same commands as above albeit, with a `--follow` flag:
+* Use `odo logs --dev` to follow the logs for the containers created by `odo dev` command.
```
feloy, line 26: Could you add the `--follow` flag to the commands?

pkg/odo/cli/logs/logs.go
```diff
@@ -133,17 +135,44 @@ func (o *LogsOptions) Run(ctx context.Context) error {
...
 		for container, logs := range entry {
 			uniqueName := getUniqueContainerName(container, uniqueContainerNames)
 			uniqueContainerNames[uniqueName] = struct{}{}
-			err = printLogs(uniqueName, logs, o.out)
-			if err != nil {
-				return err
-			}
+			l := logs
+			wg.Add(1)
+			go func(out io.Writer, follow bool) {
```
feloy, line 147: Without the `--follow` flag, the logs will be displayed in a random order (the one acquiring the lock first, etc).

IMO, you could run the function as non go routine for non-follow, and run it as go routine for follow, so you don't need the use of mutex and you respect the order of the logs.

tests/examples/source/devfiles/nodejs/devfile-deploy-functional-pods.yaml
```diff
@@ -113,7 +114,7 @@ components:
         spec:
           containers:
           - name: main
-            image: "{{CONTAINER_IMAGE}}"
+            image: "{{LOGS_IMAGE}}"
```
feloy, line 117: Instead of using a specific image, it could be better to use a standard one, and to make appear what the image does in the command/args of the container:

```
          containers:
          - name: main
            image: fedora # (or lighter one)
            command: bash
            args: [ "-c", "while true; do echo \"`date` - this is infinite while loop\"; sleep 5; done" ]
```

feloy (changes requested):

docs/website/versioned_docs/version-3.0.0/command-reference/logs.md
dharmit: Oops. 😆

pkg/odo/cli/logs/logs.go
dharmit: Yes. That's something that crossed my mind as well when I noticed the ordering. But I didn't make the change to be able to reuse the same implementation. Anyway, I'll make the change.

tests/examples/source/devfiles/nodejs/devfile-deploy-functional-pods.yaml
dharmit: Makes sense. I'd prefer `alpine` vs. `fedora` because of the difference in size and also because we always pull an image.

dharmit: Addressed

feloy:
When executing `odo logs` with no running pods, I get the error message with no end of line:

```
[phmartin@fedora go]$ odo logs
no containers running in the specified mode for the component "my-go-app"[phmartin@fedora go]$ 
```

feloy:
The execution ofo `odo logs` takes a long time for me, I think we will need to optimize all the requests we are doing to get to interesting pods:

```
$ time odo logs 
runtime: got resquest
runtime: got resquest
runtime: got resquest
runtime: got resquest
runtime: got resquest
runtime: got resquest

real	0m5,045s
user	0m0,236s
sys	0m0,054s
```

feloy:
With the follow option, we are only getting the last line of actual logs. Is this intentional?

```
$ odo logs 
runtime: got resquest
runtime: got resquest
runtime: got resquest
runtime: got resquest
nginx: /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
nginx: /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
nginx: /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
nginx: 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
nginx: 10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
nginx: /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
nginx: /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
nginx: /docker-entrypoint.sh: Configuration complete; ready for start up
nginx: 2022/06/22 08:09:03 [notice] 1#1: using the "epoll" event method
nginx: 2022/06/22 08:09:03 [notice] 1#1: nginx/1.21.6
nginx: 2022/06/22 08:09:03 [notice] 1#1: built by gcc 10.2.1 20210110 (Debian 10.2.1-6) 
nginx: 2022/06/22 08:09:03 [notice] 1#1: OS: Linux 5.17.11-200.fc35.x86_64
nginx: 2022/06/22 08:09:03 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker processes
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 31
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 32
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 33
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 34
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 35
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 36
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 37
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 38
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 39
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 40
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 41
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 42
```

```
$ odo logs --follow
nginx: 2022/06/22 08:09:03 [notice] 1#1: start worker process 42
runtime: got resquest
```

dharmit:
> I haven't modified it, but I feel that should show logs since the container started.

b0abf07

> Doesn't happen to me, but I have added \n so that it's covered on all shells.

f30357b

feloy:
> I am a bit lost about what's happening here. It looks like you have only one Pod in the output you have shared. I have been testing `odo logs` locally against the devfile modified in this PR, and it's not taking long for me.

It's also working fine for me on a local cluster (CRC), but it taking 5s with a cluster on my local network.

feloy:
odo-logs-follow

The logs of containers (the ones displayed at the beginning of the command, not the one coming later) are mixed up, and sometimes some color disappears, probably due to the mix up

dharmit:
> As discussed with @feloy, I'll take a look at pprof to figure out what's taking the most time and how we can possibly minimize it. But Philippe's observation is that `GetAllResourcesFromSelector` and `GetAllPodsInNamespace` are taking maximum time. What in these methods is taking so long is what we need to figure out.

I was looking into [`pprof`](https://go.dev/blog/pprof) today, but it's helpful for CPU and memory profiling. What we are facing is slowness over the network. The main culprits seem to the below functions:
* `kclient.GetAllResourcesFromSelector`
* `matchOwnerReferenceWithResources`

These are being called from `getLogsWithSelector` in `pkg/logs/logs.go` file added in this PR.

As discussed, we will track the performance issue for `odo logs` in a separate issue of its own.

dharmit:
> The logs of containers (the ones displayed at the beginning of the command, not the one coming later) are mixed up, and sometimes some color disappears, probably due to the mix up

I tried using `sync.Mutex` for this, but the order is still garbled. The disappearance of colour is also random.

pkg/odo/cli/logs/logs.go
```diff
@@ -173,18 +200,21 @@ func getUniqueContainerName(name string, uniqueNames map[string]struct{}) string
...
-	color.Set(log.ColorPicker())
-	defer color.Unset()
+func printLogs(containerName string, rd io.ReadCloser, out io.Writer, colour color.Attribute, mu *sync.Mutex) error {
 	scanner := bufio.NewScanner(rd)
 	scanner.Split(bufio.ScanLines)
 
 	for scanner.Scan() {
 		line := scanner.Text()
-		_, err := fmt.Fprintln(out, containerName+": "+line)
+		c := fmt.Sprintf("%d", colour)
```
feloy, line 209: You can remove this debug info

feloy (changes requested):

pkg/odo/cli/logs/logs.go
```diff
@@ -173,18 +200,20 @@ func getUniqueContainerName(name string, uniqueNames map[string]struct{}) string
...
 	scanner := bufio.NewScanner(rd)
 	scanner.Split(bufio.ScanLines)
 
 	for scanner.Scan() {
 		line := scanner.Text()
+		mu.Lock()
+		color.Set(colour)
 		_, err := fmt.Fprintln(out, containerName+": "+line)
 		if err != nil {
 			return err
```
rm3l, line 213: If there is an error, the mutex will never be unlocked, no? Wouldn't that be an issue?

rm3l (changes requested):

pkg/odo/cli/logs/logs.go
dharmit: Probably not, because if we hit error, the execution stops anyway. But it's a bad design, so I'm making the change.

pkg/odo/cli/logs/logs.go
```diff
@@ -173,14 +200,17 @@ func getUniqueContainerName(name string, uniqueNames map[string]struct{}) string
...
-	color.Set(log.ColorPicker())
-	defer color.Unset()
+func printLogs(containerName string, rd io.ReadCloser, out io.Writer, colour color.Attribute, mu *sync.Mutex) error {
 	scanner := bufio.NewScanner(rd)
 	scanner.Split(bufio.ScanLines)
 
 	for scanner.Scan() {
 		line := scanner.Text()
+		mu.Lock()
+		defer mu.Unlock()
```
rm3l, line 210: But with this, you are locking for each iteration of the `for` loop, but unlocking when the function returns. Is a `sync.Mutex` reentrant? If not, I guess this call would deadlock or error.

rm3l: Maybe this could be done as you did, but in a separate function handling each line..

feloy: or using a lambda:

```
line := scanner.Text()
func() {
  mu.Lock()
  defer mu.Unlock()
  ...
/* exactly what you want to lock here */
}()

Comment (dharmit) [reply pkg/odo/cli/logs/logs.go]: > But with this, you are locking for each iteration of the `for` loop, but unlocking when the function returns. Is a `sync.Mutex` reentrant? If not, I guess the call to `mu.Lock()` at the next iteration would deadlock or error.

Totally missed that I'm doing this inside a `for` loop. Friday effect, maybe. 😉 
@feloy I like the anonymous function idea. PTAL 084d0f4.

**Review by rm3l — APPROVED** [review]
Thanks for the changes!

**Comment by dharmit** [comment]
Timeout issue in the sole test that failed on IBM Cloud:
```
• Failure in Spec Setup (BeforeEach) [185.505 seconds]
odo delete command tests
/go/odo_1/tests/integration/devfile/cmd_delete_test.go:13
  when deleting a component containing preStop event that is deployed with DEV
  /go/odo_1/tests/integration/devfile/cmd_delete_test.go:282
    should contain preStop events list [BeforeEach]
    /go/odo_1/tests/integration/devfile/cmd_delete_test.go:297

    Timed out after 180.004s.
    Expected
        <string>:   __
         /  \__     Developing using the nodejs Devfile
         \__/  \    Namespace: cmd-delete-test297nzf
         \__/
        
    to contain substring
        <string>: Press Ctrl+c to exit
```

dharmit:
Overriding below tests since they are passing on IBM Cloud

/override ci/prow/unit

PR #5852, opened 2022-06-21
Do not save component name in env.yaml

cdrage:
**What type of PR is this:**

/kind bug

By default we should be saving the namespace / project name, not the app name /
component name.

Fixes https://github.com/redhat-developer/odo/issues/5780
Fixes https://github.com/redhat-developer/odo/issues/5886

**PR acceptance criteria:**

- [X] Unit test

- [X] Integration test

- [X] Documentation

### DIFF
--- a/pkg/component/pushed_component.go
+++ b/pkg/component/pushed_component.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	odolabels "github.com/redhat-developer/odo/pkg/labels"
 	"github.com/redhat-developer/odo/pkg/storage"
--- a/pkg/odo/cli/deploy/deploy.go
+++ b/pkg/odo/cli/deploy/deploy.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
+	"k8s.io/klog"
 	"k8s.io/kubectl/pkg/util/templates"
 )
 
@@ -105,27 +106,33 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
 		return err
 	}
 
-	envFileInfo, err := envinfo.NewEnvSpecificInfo(o.contextDir)
+	// ENV.YAML
+	//
+	// Set the appropriate variables in the env.yaml file
+	//
+	// TODO: Eventually refactor with code in dev/dev.go
+
+	envfileinfo, err := envinfo.NewEnvSpecificInfo("")
 	if err != nil {
 		return fmt.Errorf("unable to retrieve configuration information: %w", err)
 	}
-	if !envFileInfo.Exists() {
-		var cmpName string
-		cmpName, err = component.GatherName(o.EnvSpecificInfo.GetDevfileObj(), o.GetDevfilePath())
-		if err != nil {
-			return fmt.Errorf("unable to retrieve component name: %w", err)
-		}
-		err = envFileInfo.SetComponentSettings(envinfo.ComponentSettings{Name: cmpName, Project: o.GetProject(), AppName: "app"})
+
+	// If the env.yaml does not exist, we will save the project name
+	if !envfileinfo.Exists() {
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
-
-	} else if envFileInfo.GetComponentSettings().Project != o.GetProject() {
-		err = envFileInfo.SetConfiguration("project", o.GetProject())
+	} else if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
+		// If the env.yaml exists and the project is set incorrectly, we'll override it.
+		klog.V(4).Info("Overriding project name in env.yaml as it's set incorrectly, new project name: ", o.GetProject())
+		err = envfileinfo.SetConfiguration("project", o.GetProject())
 		if err != nil {
 			return fmt.Errorf("failed to update project in env.yaml file: %w", err)
 		}
 	}
+
+	// END ENV.YAML
 
 	// this ensures that odo deploy uses the namespace set in env.yaml
 	o.clientset.KubernetesClient.SetNamespace(o.GetProject())
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/devfile/library/pkg/devfile/parser"
+	"k8s.io/klog"
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
 
 	"github.com/redhat-developer/odo/pkg/component"
@@ -148,33 +149,35 @@ func (o *DevOptions) Complete(cmdline cmdline.Cmdline, args []string) error {
 		return fmt.Errorf("unable to create context: %v", err)
 	}
 
+	o.initialDevfileObj = o.Context.EnvSpecificInfo.GetDevfileObj()
+
+	// ENV.YAML
+	//
+	// Set the appropriate variables in the env.yaml file
+	//
+	// TODO: Eventually refactor with code in deploy/deploy.go
 	envfileinfo, err := envinfo.NewEnvSpecificInfo("")
 	if err != nil {
 		return fmt.Errorf("unable to retrieve configuration information: %v", err)
 	}
 
-	o.initialDevfileObj = o.Context.EnvSpecificInfo.GetDevfileObj()
-
+	// If the env.yaml does not exist, we will save the project name
 	if !envfileinfo.Exists() {
-		// if env.yaml doesn't exist, get component name from the devfile.yaml
-		var cmpName string
-		cmpName, err = component.GatherName(o.EnvSpecificInfo.GetDevfileObj(), o.GetDevfilePath())
-		if err != nil {
-			return fmt.Errorf("unable to retrieve component name: %w", err)
-		}
-		// create env.yaml file with component, project/namespace and application info
-		// TODO - store only namespace into env.yaml, we don't want to track component or application name via env.yaml
-		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Name: cmpName, Project: o.GetProject(), AppName: "app"})
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
-	} else if envfileinfo.GetComponentSettings().Project != o.GetProject() {
-		// set namespace if the evn.yaml exists; that's the only piece we care about in env.yaml
+	} else if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
+		// If the env.yaml exists and the project is set incorrectly, we'll override it.
+		klog.V(4).Info("Overriding project name in env.yaml as it's set incorrectly, new project name: ", o.GetProject())
 		err = envfileinfo.SetConfiguration("project", o.GetProject())
 		if err != nil {
 			return fmt.Errorf("failed to update project in env.yaml file: %w", err)
 		}
 	}
+
+	// END ENV.YAML
+
 	o.clientset.KubernetesClient.SetNamespace(o.GetProject())
 
 	// 3 steps to evaluate the paths to be ignored when "watching" the pwd/cwd for changes
--- a/tests/integration/cmd_dev_test.go
+++ b/tests/integration/cmd_dev_test.go
@@ -273,6 +273,7 @@ var _ = Describe("odo dev command tests", func() {
 				// An ENV file should have been created indicating current namespace
 				Expect(helper.VerifyFileExists(".odo/env/env.yaml")).To(BeTrue())
 				helper.FileShouldContainSubstring(".odo/env/env.yaml", "Project: "+commonVar.Project)
+				helper.FileShouldNotContainSubstring(".odo/env/env.yaml", "Name: "+cmpName)
 			})
 
 			AfterEach(func() {
@@ -302,6 +303,12 @@ var _ = Describe("odo dev command tests", func() {
 						Not(BeEmpty()),
 						ContainSubstring(fmt.Sprintf("%s-app-", cmpName)),
 					))
+				})
+
+				// Returned pvc yaml contains ownerreference
+				By("creating a pvc with ownerreference", func() {
+					output := commonVar.CliRunner.Run("get", "pvc", "--namespace", commonVar.Project, "-o", `jsonpath='{range .items[*].metadata.ownerReferences[*]}{@..kind}{"/"}{@..name}{"\n"}{end}'`).Out.Contents()
+					Expect(string(output)).To(ContainSubstring(fmt.Sprintf("Deployment/%s-app", cmpName)))
 				})
 			})
 
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -414,7 +414,7 @@ func (a *Adapter) createOrUpdateComponent(
 	ei.SetDevfileObj(a.Devfile)
 	componentName := a.ComponentName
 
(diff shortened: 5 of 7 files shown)


feloy:
I think you will have to make the change on deploy command also (pkg/odo/cli/deploy/deploy.go)

cdrage:
Thanks @valaparthvi I'll cover a test.

The problem is that I am going on PTO shortly so I won't be able to get back to this for another week.

I'm also running into this issue: https://github.com/redhat-developer/odo/issues/5927 which is not helping with regards to diagnosing what integration test is failing. So I've been working on that so I can test these changes locally.

Unfortunately the IBM test cloud isn't showing the logging (see K8s slack) for which test is failing as the log (I presume) is too large to show. 

I'll let you know what progress I get with this, but this is my current update!

cdrage:
Ready for review! Tests pass, env no longer passes the name, and storage has been fixed with ephemeral.

I've also updated the tests to check that the name is NOT passed into env.yaml whatsoever.

pkg/odo/cli/deploy/deploy.go
```diff
@@ -105,28 +105,33 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
...
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
+	}
 
-	} else if envFileInfo.GetComponentSettings().Project != o.GetProject() {
-		err = envFileInfo.SetConfiguration("project", o.GetProject())
+	// If the env.yaml exists and the project is set incorrectly, we'll override it.
+	if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
+		err = envfileinfo.SetConfiguration("project", o.GetProject())
```
valaparthvi, line 127: Can you add a klog warning about this?

pkg/odo/cli/deploy/deploy.go
```diff
@@ -105,28 +105,33 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
...
 
-	envFileInfo, err := envinfo.NewEnvSpecificInfo(o.contextDir)
+	// ENV.YAML
+	//
+	// Set the appropriate variables in the env.yaml file
+	//
+	envfileinfo, err := envinfo.NewEnvSpecificInfo("")
 	if err != nil {
-		return fmt.Errorf("unable to retrieve configuration information: %w", err)
+		return fmt.Errorf("unable to retrieve configuration information: %v", err)
```
valaparthvi, line 114: Why change to %v? Shouldn't %w be used for errors?

valaparthvi (changes requested):
Thank you for this PR!
The code lgtm, except a few things.
Do you think you can add a regression test to check if the owner reference is set for the PVC, addressing issue 5886?
Also, this might be unrelated to your PR changes, I couldn't help but notice the similarities between dev and deploy commands during the setup. Do you think you can add a TODO to abstract the commonalities between the two to a separate function so that we don't need to change both the places? I like that they're defined explicitly, but if we want the behavior to be same for both of them, it makes sense to abstract things to a single function.

cdrage:
> Thank you for this PR! The code lgtm, except a few things. Do you think you can add a regression test to check if the owner reference is set for the PVC, addressing issue 5886? Also, this might be unrelated to your PR changes, I couldn't help but notice the similarities between dev and deploy commands during the setup. Do you think you can add a TODO to abstract the commonalities between the two to a separate function so that we don't need to change both the places? I like that they're defined explicitly, but if we want the behavior to be same for both of them, it makes sense to abstract things to a single function.

I've made sure to add a test to remove the `metadata.name` by checking to see if it's invoked with `env.yaml`.

I've updated the PR and I'll add the regression test shortly!

cdrage:
Thanks @valaparthvi 

This is ready for review! I've combined the commits so it's cleaner, as well as added a regression test for making sure that "ownerReferences" is created with a PVC.

tests/integration/cmd_dev_test.go
```diff
@@ -303,6 +304,12 @@ var _ = Describe("odo dev command tests", func() {
 						ContainSubstring(fmt.Sprintf("%s-app-", cmpName)),
 					))
 				})
+
+				// Returned pvc yaml contains ownerreference
+				By("creating a pvc with ownerreference", func() {
+					output := commonVar.CliRunner.Run("get", "pvc", "--namespace", commonVar.Project, "-o", "yaml").Out.Contents()
+					Expect(string(output)).To(ContainSubstring("ownerReferences"))
```
rm3l, line 311: Can you be more specific here, by checking that this list of owner references does contain the Deployment? Here we are sure that there is an owner reference, but we can't tell for sure that it contains the one we want.
I think that was the actual issue: the PVC was not owned by the Deployment. 

Maybe using a JSONPath template expression like this can help:
```sh
kubectl get pvc --namespace <ns> \
  -o jsonpath='{range .items[*].metadata.ownerReferences[*]}{@..kind}{"/"}{@..name}{end}'
```
It will return both the kind and name of each owner reference separated by a `/`, e.g.: `Deployment/${componentName}-app`

pkg/odo/cli/deploy/deploy.go
```diff
@@ -105,28 +106,36 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
...
 	}
 
-	envFileInfo, err := envinfo.NewEnvSpecificInfo(o.contextDir)
+	// ENV.YAML
+	//
+	// Set the appropriate variables in the env.yaml file
+	//
+	// TODO: Eventually refactor with code in dev/dev.go
+
+	envfileinfo, err := envinfo.NewEnvSpecificInfo("")
```
rm3l, line 115: Just curious as to why you replaced `o.contextDir` by `""`.. I think it makes sense to leave it explicit here, as it represents the current working directory, no?

pkg/odo/cli/deploy/deploy.go
```diff
@@ -105,28 +106,36 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
...
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
+	}
 
-	} else if envFileInfo.GetComponentSettings().Project != o.GetProject() {
-		err = envFileInfo.SetConfiguration("project", o.GetProject())
+	// If the env.yaml exists and the project is set incorrectly, we'll override it.
+	if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
```
rm3l, line 129: nit: The previous `else if envfileinfo.GetComponentSettings().Project != o.GetProject()` makes the condition clearer, I think.

pkg/odo/cli/dev/dev.go
```diff
@@ -148,33 +149,37 @@ func (o *DevOptions) Complete(cmdline cmdline.Cmdline, args []string) error {
...
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
-	} else if envfileinfo.GetComponentSettings().Project != o.GetProject() {
-		// set namespace if the evn.yaml exists; that's the only piece we care about in env.yaml
+	}
+
+	// If the env.yaml exists and the project is set incorrectly, we'll override it.
+	if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
```
rm3l, line 174: nit: The previous `else if envfileinfo.GetComponentSettings().Project != o.GetProject()` makes the condition clearer, I think.

pkg/component/pushed_component.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
rm3l, line 7: To reduce the number of changes in this PR, do we need this import alias?

From the coding conventions page ([`ImportAliasing`](https://github.com/redhat-developer/odo/wiki/Dev:-Coding-Conventions#import-aliasing)):

> Import aliasing must be used if the package name does not match the last element of the import path.
> In all other scenarios, import aliases should be avoided unless there is a direct conflict between imports.

rm3l (changes requested):
Added a few comments.

Also, I see that the `component.GatherName` function, which was used before, is no longer being used here. Can you remove it then?

pkg/odo/cli/deploy/deploy.go
cdrage: To prevent using context anymore and since it's the same code for both `dev.go` and `deploy.go`, that's why it's "" :)

pkg/odo/cli/deploy/deploy.go
cdrage: Yes, but we also need envfileinfo.Exists() for nil check as well, so that's why it's added

pkg/odo/cli/dev/dev.go
cdrage: Need envfileinfo.Exists for nil check.

pkg/odo/cli/deploy/deploy.go
rm3l: Not sure I understand.. Which nil check are you referring to? 
With the `else if` statement, we are sure that `envfileinfo.Exists()` evaluates to `true`, so we don't need to check it again.
I just find the code to be more readable with the `else if` that was there before.

pkg/odo/cli/dev/dev.go
```diff
@@ -148,33 +149,37 @@ func (o *DevOptions) Complete(cmdline cmdline.Cmdline, args []string) error {
...
-		// if env.yaml doesn't exist, get component name from the devfile.yaml
-		var cmpName string
-		cmpName, err = component.GatherName(o.EnvSpecificInfo.GetDevfileObj(), o.GetDevfilePath())
-		if err != nil {
-			return fmt.Errorf("unable to retrieve component name: %w", err)
-		}
-		// create env.yaml file with component, project/namespace and application info
-		// TODO - store only namespace into env.yaml, we don't want to track component or application name via env.yaml
-		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Name: cmpName, Project: o.GetProject(), AppName: "app"})
+		klog.V(4).Info("Overriding project name in env.yaml as it's set incorrectly, new project name: ", o.GetProject())
```
rm3l, line 166: The message logged here is a bit misleading, I think, as the `env.yaml` file did not exist. Looks like this should go in the other condition (`envfileinfo.GetComponentSettings().Project != o.GetProject()`) instead.

tests/integration/cmd_dev_test.go
```diff
@@ -303,6 +304,12 @@ var _ = Describe("odo dev command tests", func() {
 						ContainSubstring(fmt.Sprintf("%s-app-", cmpName)),
 					))
 				})
+
+				// Returned pvc yaml contains ownerreference
+				By("creating a pvc with ownerreference", func() {
+					output := commonVar.CliRunner.Run("get", "pvc", "--namespace", commonVar.Project, "-o", `jsonpath='{range .items[*].metadata.ownerReferences[*]}{@..kind}{"/"}{@..name}{end}'`).Out.Contents()
```
rm3l, line 310: I just tested this command, and without a break line after each element, if there are multiple PVCs, we might everything printed on a single line and with no separator.

```suggestion
					output := commonVar.CliRunner.Run("get", "pvc", "--namespace", commonVar.Project, "-o", `jsonpath='{range .items[*].metadata.ownerReferences[*]}{@..kind}{"/"}{@..name}{"\n"}{end}'`).Out.Contents()
```

rm3l (changes requested):

pkg/odo/cli/deploy/deploy.go
```diff
@@ -105,28 +106,34 @@ func (o *DeployOptions) Complete(cmdline cmdline.Cmdline, args []string) (err er
...
+	// If the env.yaml does not exist, we will save the project name
+	if !envfileinfo.Exists() {
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
-
-	} else if envFileInfo.GetComponentSettings().Project != o.GetProject() {
-		err = envFileInfo.SetConfiguration("project", o.GetProject())
+	} else if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
```
rm3l, line 126: Thanks for the changes, but then we don´t need to check for `envfileinfo.Exists()`..

```suggestion
	} else if envfileinfo.GetComponentSettings().Project != o.GetProject() {
```

pkg/odo/cli/dev/dev.go
```diff
@@ -148,33 +149,35 @@ func (o *DevOptions) Complete(cmdline cmdline.Cmdline, args []string) error {
...
-		// create env.yaml file with component, project/namespace and application info
-		// TODO - store only namespace into env.yaml, we don't want to track component or application name via env.yaml
-		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Name: cmpName, Project: o.GetProject(), AppName: "app"})
+		err = envfileinfo.SetComponentSettings(envinfo.ComponentSettings{Project: o.GetProject()})
 		if err != nil {
 			return fmt.Errorf("failed to write new env.yaml file: %w", err)
 		}
-	} else if envfileinfo.GetComponentSettings().Project != o.GetProject() {
-		// set namespace if the evn.yaml exists; that's the only piece we care about in env.yaml
+	} else if envfileinfo.Exists() && envfileinfo.GetComponentSettings().Project != o.GetProject() {
```
rm3l, line 170: Thanks for the changes, but then we don´t need to check for `envfileinfo.Exists()`..

```suggestion
	} else if envfileinfo.GetComponentSettings().Project != o.GetProject() {
```

rm3l (changes requested):

cdrage:
Rebasing and re-pushing, but up for another review! 👍

valaparthvi (approved):

rm3l (commented):
LGTM feature-wise, but would it be possible to clean up the code by removing both `Name` and `AppName` fields from the [`ComponentSettings`](https://github.com/redhat-developer/odo/blob/main/pkg/envinfo/types.go) struct?
I guess this will require cleaning up further things, like removing the `GetApplication` and `GetName` methods from the [`LocalConfigProvider`](https://github.com/redhat-developer/odo/blob/main/pkg/localConfigProvider/localConfigProvider.go#L25-L26) interface as well..

Maybe I've missed something, but I think that this way, we would be sure no one can use that information (I've tested this PR against an existing project that already has component and app names in its `env.yaml` files, but then from the code perspective, there is currently nothing in this PR preventing from loading and using such information).

cdrage:
> LGTM feature-wise, but would it be possible to clean up the code by removing both `Name` and `AppName` fields from the [`ComponentSettings`](https://github.com/redhat-developer/odo/blob/main/pkg/envinfo/types.go) struct? I guess this will require cleaning up further things, like removing the `GetApplication` and `GetName` methods from the [`LocalConfigProvider`](https://github.com/redhat-developer/odo/blob/main/pkg/localConfigProvider/localConfigProvider.go#L25-L26) interface as well..
> 
> Maybe I've missed something, but I think that this way, we would be sure no one can use that information (I've tested this PR against an existing project that already has component and app names in its `env.yaml` files, but then from the code perspective, there is currently nothing in this PR preventing from loading and using such information).

Let's leave this for not for a later refactor?

rm3l:
> > LGTM feature-wise, but would it be possible to clean up the code by removing both `Name` and `AppName` fields from the [`ComponentSettings`](https://github.com/redhat-developer/odo/blob/main/pkg/envinfo/types.go?rgh-link-date=2022-07-27T18%3A32%3A18Z) struct? I guess this will require cleaning up further things, like removing the `GetApplication` and `GetName` methods from the [`LocalConfigProvider`](https://github.com/redhat-developer/odo/blob/main/pkg/localConfigProvider/localConfigProvider.go?rgh-link-date=2022-07-27T18%3A32%3A18Z#L25-L26) interface as well..
> > Maybe I've missed something, but I think that this way, we would be sure no one can use that information (I've tested this PR against an existing project that already has component and app names in its `env.yaml` files, but then from the code perspective, there is currently nothing in this PR preventing from loading and using such information).
> 
> Let's leave this for now for a future refactor as that'd have to include a lot of changes to LocalConfigProvider and ComponentSEttings.

Alright - let's track this in a separate issue then.

/approve

cdrage:
Re-added LGTM due to rebase

PR #5858, opened 2022-06-22
odo add binding without devfile.yaml

### LINKED ISSUE #5772: odo add binding - executed in directory without `devfile.yaml`
This is extension of https://github.com/redhat-developer/odo/issues/5627

But for situations where it is executed in a directory without devfile.yaml.

## User Story
- As an user
- I want to be able to interactively generate ServiceBinding
- So that I don't have to run multiple `kubectl`/`oc` command to retrieve all required information

## Acceptance Criteria
- [ ] 

/kind user-story

### Select service instance you want to bind to:
The same as in  https://github.com/redhat-developer/odo/issues/5627

### Select Kubernetes resource kind of your application:
Select, Currently we will list just core resources that have PodSpec (Deploymet,DeploymentConfig, StatefullSet, DaemonSet .....)

### ? What is "&lt;selected option from the previous step>" name?: 
Lists all resource instances of a given kind in the current namespace + on extra option "DOES NOT EXISTS YET"

If the user selects an existing resource it moves to the next question.
If the user selects "DOES NOT EXISTS YET" then it asks the extra question "What is the  "&lt;selected option from the previous step>" name.

### Enter the Binding's name:
The same as in  https://github.com/redhat-developer/odo/issues/5627

### How to expose binding information?
see https://github.com/redhat-developer/odo/issues/5770

### What to do with generated ServiceBinding?
Multi select option
```
> [ ]  create it on cluster
  [x]  display it
  [ ]  save to file
```

if `seve to file` is selected it asks user to input the path to the file. The default value  for this is `kubernetes/<service binding name>`

if `create it on cluster` is selected it directly applies the resulting SB resource to cluster

if `display it` is selected it show resulting SB resource on stdout

feloy:
**What type of PR is this:**

/kind feature

- ServiceBinding service is now referenced by its Kind, not its Resource, to be consistent with the value added as flag (with Kind)
- Ability to run `odo add binding` withtout devfile.yaml
  - in interactive mode, output as:
    - file
    - stdout
    - creat on cluster
  - in non-interactive mode, with `--workload` flag, output as:
    - stdout

Fixes #5772 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation 

### DIFF
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -3,7 +3,17 @@ title: odo add binding
 ---
 
 ## Description
-The `odo add binding` command adds a link between an Operator-backed service and a component. odo uses the [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator/) to create this link. Running this command will modify the Devfile, and once pushed (using `odo dev`) to the cluster, it creates an instance of the `ServiceBinding` resource.
+The `odo add binding` command adds a link between an Operator-backed service and a component. odo uses the [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator/) to create this link. 
+
+Running this command from a directory containing a Devfile will modify the Devfile, and once pushed (using `odo dev`) to the cluster, it creates an instance of the `ServiceBinding` resource.
+
+Running this command from a directory without a Devfile in the interactive mode will perform one or several operations,
+depending on your choice:
+- display the YAML definition of the Service binding in the output,
+- save the YAML definition of the ServiceBinding on a file,
+- create an instance of the ServiceBinding resource on the cluster.
+
+In non-interactive mode from a directory without a Devfile, the only possible operation is to display the YAML definition of the Service binding in the output.
 
 Currently, it only allows connecting to the Operator-backed services which support binding via the Service Binding Operator.
 To know about the Operators supported by the Service Binding Operator, read its [README](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators).
@@ -11,13 +21,14 @@ To know about the Operators supported by the Service Binding Operator, read its 
 ## Running the Command
 
 ### Pre-requisites
-* A directory containing a Devfile; if you don't have one, see [odo init](init.md) on obtaining a devfile.
 * A cluster with the Service Binding Operator installed (see installation instructions for [Kubernetes](../overview/cluster-setup/kubernetes.md#installing-the-service-binding-operator) and [OpenShift](../overview/cluster-setup/openshift.md#installing-the-service-binding-operator) cluster)
 * Operator-backed services or resources you want to bind your application to
+* Optional, a directory containing a Devfile; if you don't have one, see [odo init](init.md) on obtaining a devfile.
 
 ### Interactive Mode
 In the interactive mode, you will be guided to choose:
 * a service from the list of bindable service instances as supported by the Service Binding Operator,
+* if a Devfile is not present in the directory, a workload resource,
 * option to bind the service as a file (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this),
 * a name for the binding.
 
@@ -29,6 +40,7 @@ odo add binding
 ### Non-interactive mode
 In the non-interactive mode, you will have to specify the following required information through the command-line:
 * `--service` flag to specify the service you want to bind to,
+* `--workload` flag to specify the workload resource, if a Devfile is not present in the directory,
 * `--name` flag to specify a name for the binding (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this)
 * `--bind-as-files` flag to specify if the service should be bound as a file; this flag is set to true by default.
 
@@ -46,8 +58,8 @@ If the service is bound as files, this data will be written to a file and stored
 Note that every piece of data is stored in its own individual file or environment variable.
 For example, if your data includes a username and password, then 2 separate files, or 2 environment variables will be created to store them both.
 
-#### Formats supported by the `--service` flag
-The `--service` flag supports the following formats to specify the service name:
+#### Formats supported by the `--service` and `--workload` flags
+The `--service` and `--workload` flags support the following formats to specify the service or workload name:
 * `<name>`
 * `<name>.<kind>`
 * `<name>.<kind>.<apigroup>`
--- a/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
+++ b/docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
@@ -58,12 +58,17 @@ If the service is bound as files, this data will be written to a file and stored
 Note that every piece of data is stored in its own individual file or environment variable.
 For example, if your data includes a username and password, then 2 separate files, or 2 environment variables will be created to store them both.
 
-#### Formats supported by the `--service` and `--workload` flags
-The `--service` and `--workload` flags support the following formats to specify the service or workload name:
+#### Formats supported by the `--service` flag
+The `--service` flag supports the following formats to specify the service name:
 * `<name>`
 * `<name>.<kind>`
 * `<name>.<kind>.<apigroup>`
 * `<name>/<kind>`
+* `<name>/<kind>.<apigroup>`
+
+#### Formats supported by the `--workload` flag
+The `--workload` flag supports the following formats to specify the workload name:
+* `<name>.<kind>.<apigroup>`
 * `<name>/<kind>.<apigroup>`
 
 The above formats are helpful when multiple services with the same name exist on the cluster.
--- a/pkg/binding/add.go
+++ b/pkg/binding/add.go
@@ -128,9 +128,12 @@ func (o *BindingClient) AddBinding(
 		backend = o.flagsBackend
 	}
 
-	options, err := backend.SelectCreationOptions(flags)
-	if err != nil {
-		return err
+	var options []asker.CreationOption
+	for len(options) == 0 {
+		options, err = backend.SelectCreationOptions(flags)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Note: we cannot directly marshal the serviceBinding object to yaml because it doesn't do that in the correct k8s manifest format
--- a/pkg/binding/add.go
+++ b/pkg/binding/add.go
@@ -2,15 +2,20 @@ package binding
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/devfile/library/pkg/devfile/parser"
 	sboApi "github.com/redhat-developer/service-binding-operator/apis/binding/v1alpha1"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"github.com/redhat-developer/odo/pkg/binding/asker"
 	backendpkg "github.com/redhat-developer/odo/pkg/binding/backend"
 	"github.com/redhat-developer/odo/pkg/kclient"
 	"github.com/redhat-developer/odo/pkg/libdevfile"
+	"github.com/redhat-developer/odo/pkg/log"
 )
 
 // ValidateAddBinding calls Validate method of the adequate backend
@@ -34,6 +39,20 @@ func (o *BindingClient) SelectServiceInstance(flags map[string]string, serviceMa
 	return backend.SelectServiceInstance(flags[backendpkg.FLAG_SERVICE], serviceMap)
 }
 
+func (o *BindingClient) SelectWorkloadInstance(flags map[string]string) (string, schema.GroupVersionKind, error) {
+	var backend backendpkg.AddBindingBackend
+	if len(flags) == 0 {
+		backend = o.interactiveBackend
+	} else {
+		backend = o.flagsBackend
+	}
+	resource, gvk, err := backend.SelectWorkloadInstance(flags[backendpkg.FLAG_WORKLOAD])
+	if err != nil {
+		return "", schema.GroupVersionKind{}, err
+	}
+	return resource, gvk, nil
+}
+
 func (o *BindingClient) AskBindingName(serviceName, componentName string, flags map[string]string) (string, error) {
 	var backend backendpkg.AddBindingBackend
 	if len(flags) == 0 {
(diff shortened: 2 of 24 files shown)


valaparthvi:
I ran this command from an empty directory, and these are my observations.
1. Interactive mode UI is functional, but it does not seem to be doing anything, it exits just like that.
```sh
➜  odo add binding                                                                                       
? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
? Select workload resource you want to bind: Deployment
? Select workload resource name you want to bind: my-node-app-app
? Enter the Binding's name: my-node-app-app-cluster-sample
? How do you want to bind the service? Bind As Files
? What do you want to do with the ServiceBinding: 
```
2. There is no option to go back in case you select the wrong workload resource for binding. It would also be nice to go back if the user changes their mind while selecting resource name and wants to go back to a different workload resource.
```sh
➜  odo add binding
? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
? Select workload resource you want to bind: DaemonSet
? Select workload resource name you want to bind:  [Use arrows to move, type to filter]
> DOES NOT EXIST YET
```

The work you've done on UI is good, but it looks like you're missing some function calls, I haven't checked the code yet, so I'm not sure which.

feloy:
> I ran this command from an empty directory, and these are my observations.
> 
> 1. Interactive mode UI is functional, but it does not seem to be doing anything, it exits just like that.
> 
> ```shell
> ➜  odo add binding                                                                                       
> ? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
> ? Select workload resource you want to bind: Deployment
> ? Select workload resource name you want to bind: my-node-app-app
> ? Enter the Binding's name: my-node-app-app-cluster-sample
> ? How do you want to bind the service? Bind As Files
> ? What do you want to do with the ServiceBinding: 
> ```
> 

Don't you get the list of options after `What do you want to do with the ServiceBinding:`?

```
 odo add binding
? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
? Select workload resource you want to bind: Deployment
? Select workload resource name you want to bind: nginx
? Enter the Binding's name: nginx-cluster-sample
? How do you want to bind the service? Bind As Files
? What do you want to do with the ServiceBinding:  [Use arrows to move, enter to select, type to filter]
> [ ]  create it on cluster
  [ ]  display it
  [ ]  save to file
```

valaparthvi:
> > I ran this command from an empty directory, and these are my observations.
> > 
> > 1. Interactive mode UI is functional, but it does not seem to be doing anything, it exits just like that.
> > 
> > ```shell
> > ➜  odo add binding                                                                                       
> > ? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
> > ? Select workload resource you want to bind: Deployment
> > ? Select workload resource name you want to bind: my-node-app-app
> > ? Enter the Binding's name: my-node-app-app-cluster-sample
> > ? How do you want to bind the service? Bind As Files
> > ? What do you want to do with the ServiceBinding: 
> > ```
> 
> Don't you get the list of options after `What do you want to do with the ServiceBinding:`?
> 
> ```
>  odo add binding
> ? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
> ? Select workload resource you want to bind: Deployment
> ? Select workload resource name you want to bind: nginx
> ? Enter the Binding's name: nginx-cluster-sample
> ? How do you want to bind the service? Bind As Files
> ? What do you want to do with the ServiceBinding:  [Use arrows to move, enter to select, type to filter]
> > [ ]  create it on cluster
>   [ ]  display it
>   [ ]  save to file
> ```

I do, but selecting either of the options does not do anything. It ends abruptly.

feloy:
> > > I ran this command from an empty directory, and these are my observations.
> > > 
> > > 1. Interactive mode UI is functional, but it does not seem to be doing anything, it exits just like that.
> > > 
> > > ```shell
> > > ➜  odo add binding                                                                                       
> > > ? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
> > > ? Select workload resource you want to bind: Deployment
> > > ? Select workload resource name you want to bind: my-node-app-app
> > > ? Enter the Binding's name: my-node-app-app-cluster-sample
> > > ? How do you want to bind the service? Bind As Files
> > > ? What do you want to do with the ServiceBinding: 
> > > ```
> > 
> > 
> > Don't you get the list of options after `What do you want to do with the ServiceBinding:`?
> > ```
> >  odo add binding
> > ? Select service instance you want to bind to: cluster-sample (Cluster.postgresql.k8s.enterprisedb.io)
> > ? Select workload resource you want to bind: Deployment
> > ? Select workload resource name you want to bind: nginx
> > ? Enter the Binding's name: nginx-cluster-sample
> > ? How do you want to bind the service? Bind As Files
> > ? What do you want to do with the ServiceBinding:  [Use arrows to move, enter to select, type to filter]
> > > [ ]  create it on cluster
> >   [ ]  display it
> >   [ ]  save to file
> > ```
> 
> I do, but selecting either of the options does not do anything. It ends abruptly.

As discussed, I'll change to make it clear the user needs to select options with space bar

valaparthvi (changes requested):

docs/website/versioned_docs/version-3.0.0/command-reference/add-binding.md
```diff
@@ -54,6 +66,11 @@ The `--service` flag supports the following formats to specify the service name:
 * `<name>/<kind>`
 * `<name>/<kind>.<apigroup>`
 
+#### Formats supported by the `--workload` flag
+The `--workload` flag supports the following formats to specify the workload name:
+* `<name>.<kind>.<apigroup>`
+* `<name>/<kind>.<apigroup>`
```
valaparthvi, line 72: :100:

pkg/odo/cli/add/binding/binding.go
```diff
@@ -34,6 +35,8 @@ var addBindingExample = ktemplates.Examples(`
 
 %[1]s --service myservice.Redis --name myRedisService
 
+# Add binding between service named 'myservice' of kind 'Redis' and the deployment app (without Devfile)
+%[1]s --service myservice/Redis --name myRedisService --workload app/Deployment.apps
```
valaparthvi, line 39:
```suggestion
%[1]s --service myservice/Redis --name myRedisService --workload app/Deployment.apps
%[1]s --service myservice/Redis --name myRedisService --workload app.Deployment.apps
```

pkg/binding/backend/flags.go
```diff
@@ -23,17 +27,34 @@ func NewFlagsBackend() *FlagsBackend {
...
 }
 
+func (o *FlagsBackend) SelectWorkloadInstance(workloadName string) (string, schema.GroupVersionKind, error) {
+	selectedName, selectedKind, selectedGroup := parseServiceName(workloadName)
+	for _, gvk := range append(kclient.NativeWorkloadKinds, kclient.CustomWorkloadKinds...) {
+		if gvk.Group == selectedGroup && gvk.Kind == selectedKind {
+			return selectedName, gvk, nil
+		}
+	}
+	return "", schema.GroupVersionKind{}, fmt.Errorf("group/kind %q not found on the cluster", selectedKind+"/"+selectedGroup)
```
valaparthvi, line 55:
```suggestion
	return "", schema.GroupVersionKind{}, fmt.Errorf("group/kind %q not found on the cluster", selectedGroup+"/"+selectedKind)
```

pkg/kclient/binding.go
```diff
@@ -280,6 +299,32 @@ func (c Client) APIServiceBindingFromSpec(spec specApi.ServiceBinding) api.Servi
 	}
 }
 
+// GetWorkloadKinds returns all the workload kinds present in the cluster
+// It considers that all native resources are present and tests only for custom resources
+func (c Client) GetWorkloadKinds() ([]string, []schema.GroupVersionKind, error) {
```
valaparthvi, line 304: Can you document about the objects that are returned?

pkg/binding/backend/interactive.go
```diff
@@ -2,25 +2,100 @@ package backend
...
+		kubernetesClient: kubernetesClient,
 	}
 }
 
-func (o *InteractiveBackend) Validate(_ map[string]string) error {
+func (o *InteractiveBackend) Validate(_ map[string]string, _ bool) error {
 	return nil
 }
 
+func (o *InteractiveBackend) SelectWorkloadInstance(workloadName string) (string, schema.GroupVersionKind, error) {
```
valaparthvi, line 36:
```suggestion
func (o *InteractiveBackend) SelectWorkloadInstance(_ string) (string, schema.GroupVersionKind, error) {
```

pkg/binding/asker/survey_asker.go
```diff
@@ -61,3 +110,33 @@ func (o *Survey) AskBindAsFiles() (bool, error) {
 	}
 	return bindAsFile, nil
 }
+
+func (o *Survey) SelectCreationOption() ([]CreationOption, error) {
+	options := []int{}
+	prompt := &survey.MultiSelect{
+		Message: "Check with spacebar the operations to perform with the ServiceBinding:",
+		Options: []string{"create it on cluster", "display it", "save it to file"}, // respect order of CreateOption constants
+	}
```
valaparthvi, line 119:
```suggestion
	prompt := &survey.MultiSelect{
		Message: "Check(with Space Bar) one or more operations to perform with the ServiceBinding:",
		Options: []string{"create it on cluster", "display it", "save it to file"}, // respect order of CreateOption constants
		Help: "Use the Space Bar to select one or more operations to perform with the ServiceBinding"
	}
```

valaparthvi (changes requested):
Code lgtm, I just have a few nitpick comments about comments.

pkg/binding/add.go
```diff
@@ -81,3 +105,90 @@ func (o *BindingClient) AddBinding(bindingName string, bindAsFiles bool, unstruc
...
+	var options []asker.CreationOption
+	for len(options) == 0 {
+		options, err = backend.SelectCreationOptions(flags)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Note: we cannot directly marshal the serviceBinding object to yaml because it doesn't do that in the correct k8s manifest format
+	serviceBindingUnstructured, err := kclient.ConvertK8sResourceToUnstructured(serviceBinding)
```
dharmit, line 140: Have you explored `runtime.DefaultUnstructuredConverter.ToUnstructured(serviceBinding)`? I used its counterpart in `odo logs` https://github.com/redhat-developer/odo/blob/b3517f285e9e1a3a19996401f7e7d5e494986d2e/pkg/logs/logs.go#L76

pkg/binding/add.go
```diff
@@ -81,3 +105,90 @@ func (o *BindingClient) AddBinding(bindingName string, bindAsFiles bool, unstruc
...
+
+		case asker.OutputToStdout:
+			fmt.Println(string(yamlDesc))
+
+		case asker.CreateOnCluster:
+			_, err = o.kubernetesClient.PatchDynamicResource(serviceBindingUnstructured)
+			if err != nil {
+				return err
+			}
+		}
```
dharmit, line 180: IMO, we are doing a set of things in the business layer that ought to be done by CLI, or perhaps by a different method in `kclient` package. For example, creating a file and writing to it, and printing something using `fmt.Println` should not be happening in business layer. And the call to `o.kubernetesClient.PatchDynamicResource(serviceBindingUnstructured)` should be happening from the CLI layer.

This method should only help generate a binding and not get into the business of doing anything else. In that sense, it should probably be called `CreateBindingWithoutDevfile` and return the created binding.

Probably as you said on https://github.com/redhat-developer/odo/pull/5858/files#r905903094, this was done by mistake.

pkg/binding/asker/survey_asker.go
```diff
@@ -61,3 +110,34 @@ func (o *Survey) AskBindAsFiles() (bool, error) {
 	}
 	return bindAsFile, nil
 }
+
+func (o *Survey) SelectCreationOption() ([]CreationOption, error) {
+	options := []int{}
+	prompt := &survey.MultiSelect{
+		Message: "Check(with Space Bar) one or more operations to perform with the ServiceBinding:",
+		Options: []string{"create it on cluster", "display it", "save it to file"}, // respect order of CreateOption constants
```
dharmit, line 118: Which/where are these `CreateOption` constants that you talk about?

pkg/kclient/operators.go
```diff
@@ -211,6 +211,21 @@ func (c *Client) GetGVKFromGVR(gvr schema.GroupVersionResource) (schema.GroupVer
 	return mapper.KindFor(gvr)
 }
 
+func (c *Client) GetGVRFromGVK(gvk schema.GroupVersionKind) (schema.GroupVersionResource, error) {
+	cfg := c.GetClientConfig()
+
+	dc, err := discovery.NewDiscoveryClientForConfig(cfg)
```
dharmit, line 217: Do you think we could reuse `c.discoverClient` here instead of creating a new client? Should save some API calls, I guess.

dharmit (changes requested):

pkg/binding/add.go
feloy: There is  a TODO in `ConvertK8sResourceToUnstructured` to use this method instead

I prefer if it is done as part of another PR, to do this in a consistent manner across all the code

pkg/binding/add.go
feloy: I agree for the log.Print and fmt.Print.

For the creation of file and resource, I'm not sure

pkg/binding/asker/survey_asker.go
feloy: I fixed the typo: it was CreationOption

feloy: I fixed the typo: it was CreationOption (constants defined in the interface.go file of the same package, and of this type CreationOptions)

pkg/kclient/operators.go
feloy: Oh, you are right. I also added the restmapper to the client so it is created only once.

pkg/binding/add.go
dharmit: > For the creation of file and resource, I'm not sure

IMO, business layer should help do things that CLI layer absolutely can't/shouldn't. Also, if I imagine this blurb of code from a unit test perspective, it's not a good candidate because of the sheer amount of things that this function does due to the different cases it handles. 

OTOH, there are places in `pkg/binding/` where we do things based on certain backend. But if we think of the business layer as the API, it's difficult to use this function for something helpful without being tied tightly to what it does and how it does it.

@feloy maybe we should discuss `pkg/binding/` in a separate issue and refactor it at some point?

pkg/binding/add.go
dharmit: Sure, makes sense. I didn't notice the TODO.

pkg/binding/add.go
feloy: > > For the creation of file and resource, I'm not sure
> 
> IMO, business layer should help do things that CLI layer absolutely can't/shouldn't. Also, if I imagine this blurb of code from a unit test perspective, it's not a good candidate because of the sheer amount of things that this function does due to the different cases it handles.
> 

For me, the business layer does what is common to all frontends (CLI, plugin, HTML interface?, etc). 

To get the plugin as an example, it would use the flags (and so the flag backend) and the JSON output, and in this case, the file would be created the same way, and the resource would be created in the cluster the same way. The only difference would be how the result is displayed. For JSON output, it would be displayed in the JSON, not directly to stdout as raw text.

pkg/binding/asker/survey_asker.go
```diff
@@ -61,3 +110,34 @@ func (o *Survey) AskBindAsFiles() (bool, error) {
 	}
 	return bindAsFile, nil
 }
+
+func (o *Survey) SelectCreationOptions() ([]CreationOption, error) {
+	options := []int{}
+	prompt := &survey.MultiSelect{
+		Message: "Check(with Space Bar) one or more operations to perform with the ServiceBinding:",
+		Options: []string{"create it on cluster", "display it", "save it to file"}, // respect order of CreationOption constants
```
dharmit, line 118: > // respect order of CreationOption constants

Wouldn't using string constants instead of `iota` make it simpler? That way we could reuse them in these `Options` slice as well, IMO.  WDYT?

dharmit: This is not really a request for change. In case you feel it would be better to use `string`, just remove the lgtm flag, and I'll re-apply it when you push the change. Otherwise, LGTM.

/lgtm

feloy: I generally prefer using codes (here iota representing the index in the array of proposed options), instead of passing literal texts around. Mainly because literal text is intended to be translated and that would need more manipulation, and also because if is simpler to compare codes instead of literal text.

feloy:
Tests pass on IBM Cloud

PR #5884, opened 2022-06-27
Remove "unmaintained" banner for 2.5.0

cdrage:
**What type of PR is this:**

/kind documentation

Removes the "unmaintained" part of website when it refers to 2.5.0

N/A

### DIFF
--- a/docs/website/docusaurus.config.js
+++ b/docs/website/docusaurus.config.js
@@ -129,10 +129,13 @@ module.exports = {
             current: {
               label: '3.0.0 (Alpha 2) 🚧',
               badge: true,
+              banner: 'none',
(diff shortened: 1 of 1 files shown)


cdrage:
> Why are we removing this banner? I thought it was the way to alert the user they had landed in the "wrong" documentation section. Without this banner, it is not straightforward to distinguish between v2.5 and v3 pages

Unfortunately Docusaurus automatically adds a "unmaintained" banner if the version number is lower than the most current one.

So instead, I have disabled the banner and enabled the "badge" so that the version will always show (with a label for stable)

Screen Shot 2022-06-28 at 8 05 11 AM

Issue #5872, opened 2022-06-24
`odo logs` too slow for non-local clusters


dharmit:
## What versions of software are you using?

**Operating System:** all

**Output of `odo version`:** `main`

## How did you run odo exactly?
`odo logs` against a cluster that's not minikube/CRC. For example, one spun up using ClusterBot

## Actual behavior
Takes many seconds to show the logs

## Expected behavior
Take lesser time

## Any logs, error output, etc.?
Originally discussed at https://github.com/redhat-developer/odo/pull/5846#issuecomment-1164423486.

/kind bug
/area log

rm3l:
> ## Expected behavior
> Take lesser time
>
To ensure we won't have regressions in the future, I think it could be great to include some kind of performance testing for this, with an acceptable threshold for what we mean by "lesser time". Relates to #5830, but for `odo logs`.

dharmit:
Self-note:
* Look at client-go caching mentioned by Tomas
* By threshold, Armel referred to how long odo should wait till `odo logs` exits if it doesn't get a response it's looking for.

dharmit:
Tasks pending on this issue:
* Address `odo logs --follow` issue explained in https://github.com/redhat-developer/odo/pull/5942#issuecomment-1187685258.
* In a further PR, modify the code to first fetch the resources most likely to have Pods of interest for odo. This would include Deployments, ReplicaSets and Pods itself. Check other resources after that.
* Use channels instead of sharing a slice+mutex in the go routines, as is the case here: https://github.com/redhat-developer/odo/blob/e7588e33ef412177cf050efe68fc2af0e16ac79b/pkg/kclient/all.go#L29-L57
* Anything else that we might be able to think of while working on the above.

We have discussed adding tests that would help us track the performance of `odo logs` so that we can know if it slows down in the future. This is to avoid regression, or at least finding it ourselves before it becomes a problem for the users. I prefer it be done in a separate issue than this.

PR #5908, opened 2022-07-01
Use cached discovery client

dharmit:

~This PR changes odo to use a cached discovery client.~
This PR adds a cached discovery client along with the existing discovery 
client. Functions can use the client appropriate for their use.

A cached discovery client refers its cache under `~/.kube/cache` instead
of checking with the cluster every time.

Fixes part of #5872

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

**How to test changes:**
* Spin up a k8s/ocp cluster other than one on the local box. For example, use ClusterBot.
* Execute `odo list -v2` and look for the line that reads: `all.go:119] queried api discovery in`. Also, worth noting is the line that reads: `all.go:61] all goroutines have returned in`.
* Note the time it takes and compare the same with odo built from current `main` branch.

Here are two comparative runs that I did. `odo` is the binary built using code in this PR, while `odo-main` is the binary built using `main` branch.

```sh
$ for i in `seq 1 10`; do time odo list 1>/dev/null; done     
odo list > /dev/null  0.60s user 0.15s system 14% cpu 5.357 total
odo list > /dev/null  0.46s user 0.07s system 13% cpu 4.009 total
odo list > /dev/null  0.48s user 0.08s system 13% cpu 4.190 total
odo list > /dev/null  0.46s user 0.07s system 13% cpu 3.815 total
odo list > /dev/null  0.51s user 0.08s system 12% cpu 4.686 total
odo list > /dev/null  0.47s user 0.07s system 15% cpu 3.646 total
odo list > /dev/null  0.46s user 0.08s system 13% cpu 4.120 total
odo list > /dev/null  0.55s user 0.11s system 9% cpu 6.706 total
odo list > /dev/null  0.49s user 0.09s system 14% cpu 3.853 total
odo list > /dev/null  0.49s user 0.07s system 15% cpu 3.558 total


$ for i in `seq 1 10`; do time odo-main list 1>/dev/null; done
odo-main list > /dev/null  0.65s user 0.13s system 13% cpu 5.633 total
odo-main list > /dev/null  0.55s user 0.09s system 5% cpu 11.381 total
odo-main list > /dev/null  0.48s user 0.09s system 11% cpu 4.915 total
odo-main list > /dev/null  0.58s user 0.09s system 10% cpu 6.531 total
odo-main list > /dev/null  0.61s user 0.12s system 10% cpu 6.999 total
odo-main list > /dev/null  0.50s user 0.08s system 11% cpu 4.917 total
odo-main list > /dev/null  0.58s user 0.12s system 13% cpu 5.420 total
odo-main list > /dev/null  0.51s user 0.08s system 12% cpu 4.603 total
odo-main list > /dev/null  0.58s user 0.08s system 16% cpu 3.965 total
odo-main list > /dev/null  0.59s user 0.11s system 8% cpu 8.230 total
```

### DIFF
--- a/pkg/kclient/kclient.go
+++ b/pkg/kclient/kclient.go
@@ -53,9 +53,10 @@ type Client struct {
 	// DynamicClient interacts with client-go's `dynamic` package. It is used
 	// to dynamically create service from an operator. It can take an arbitrary
 	// yaml and create k8s/OpenShift resource from it.
-	DynamicClient   dynamic.Interface
-	discoveryClient discovery.DiscoveryInterface
-	restmapper      *restmapper.DeferredDiscoveryRESTMapper
+	DynamicClient         dynamic.Interface
+	discoveryClient       discovery.DiscoveryInterface
+	cachedDiscoveryClient discovery.CachedDiscoveryInterface
+	restmapper            *restmapper.DeferredDiscoveryRESTMapper
 
 	supportedResources map[string]bool
 	// Is server side apply supported by cluster
@@ -155,8 +156,13 @@ func NewForConfig(config clientcmd.ClientConfig) (client *Client, err error) {
 		return nil, err
 	}
 
+	client.discoveryClient, err = discovery.NewDiscoveryClientForConfig(client.KubeClientConfig)
+	if err != nil {
+		return nil, err
+	}
+
 	config_flags := genericclioptions.NewConfigFlags(true)
-	client.discoveryClient, err = config_flags.ToDiscoveryClient()
+	client.cachedDiscoveryClient, err = config_flags.ToDiscoveryClient()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/kclient/kclient.go
+++ b/pkg/kclient/kclient.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
 	"github.com/blang/semver"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -153,7 +155,8 @@ func NewForConfig(config clientcmd.ClientConfig) (client *Client, err error) {
 		return nil, err
 	}
 
-	client.discoveryClient, err = discovery.NewDiscoveryClientForConfig(client.KubeClientConfig)
+	config_flags := genericclioptions.NewConfigFlags(true)
+	client.discoveryClient, err = config_flags.ToDiscoveryClient()
 	if err != nil {
 		return nil, err
 	}
--- a/docs/website/docs/command-reference/list.md
+++ b/docs/website/docs/command-reference/list.md
@@ -18,3 +18,18 @@ defined in the local Devfile,
 
 * `--namespace` - Namespace to list the components from (optional). By default, the current namespace defined in kubeconfig is used
 * `-o json` - Outputs the list in JSON format. See [JSON output](json-output.md) for more information
+
+:::tip use of cache
+
+`odo list` makes use of cache for performance reasons. This is the same cache that is referred by `kubectl` command 
+when you do `kubectl api-resources --cached=true`. As a result, if you were to install an Operator/CRD on the 
+Kubernetes cluster, and create a resource from it using odo, you might not see it in the `odo list` output. This 
+would be the case for 10 minutes timeframe for which the cache is considered valid. Beyond this 10 minutes, the 
+cache is updated anyway.
+
+If you would like to invalidate the cache before the 10 minutes timeframe, you could manually delete it by doing:
+```shell
+rm -rf ~/.kube/cache/discovery/api.crc.testing_6443/
+```
+Above example shows how to invalidate the cache for a CRC cluster. Note that you will have to modify the `api.crc.
+testing_6443` part based on the cluster you are working against.
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -18,7 +18,7 @@ import (
 // Code into this file is heavily inspired from https://github.com/ahmetb/kubectl-tree
 
 func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unstructured.Unstructured, error) {
-	apis, err := findAPIs(c.discoveryClient)
+	apis, err := findAPIs(c.cachedDiscoveryClient)
 	if err != nil {
 		return nil, err
 	}


dharmit:
Test failures are real. We are getting the output `The project something does not exist` instead of the expected output `Unauthorized to access the cluster` because of the usage of cached discovery client introduced in this PR.

<details>

<summary> Logs from failing tests </summary>

```
•••
• Failure [6.168 seconds]
odo project command tests
/go/odo_1/tests/integration/project/cmd_project_test.go:12
  when user is logged out
  /go/odo_1/tests/integration/project/cmd_project_test.go:56
    should show login message when setting project and not login [It]
    /go/odo_1/tests/integration/project/cmd_project_test.go:69

    Expected
        <string>: I0701 12:54:47.677616   28023 util.go:767] Cached response used.
        I0701 12:54:47.685740   28023 util.go:188] path /go/odo_1/tests/integration/project/devfile.yaml doesn't exist, skipping it
        I0701 12:54:47.685815   28023 util.go:188] path /go/odo_1/tests/integration/project/.devfile.yaml doesn't exist, skipping it
        I0701 12:54:47.803845   28023 kclient.go:201] Checking if "projects" resource is supported
         ✗  The project something does not exist
        
    to contain substring
        <string>: Unauthorized to access the cluster

    /go/odo_1/tests/integration/project/cmd_project_test.go:71
• Failure [6.612 seconds]
odo project command tests
/go/odo_1/tests/integration/project/cmd_project_test.go:12
  when user is logged out
  /go/odo_1/tests/integration/project/cmd_project_test.go:56
    should show login message when deleting project and not login [It]
    /go/odo_1/tests/integration/project/cmd_project_test.go:73

    Expected
        <string>: I0701 12:54:47.435941   28006 util.go:754] Response will be cached in /tmp/odohttpcache for 1h0m0s
        I0701 12:54:47.438116   28006 util.go:767] Cached response used.
        I0701 12:54:47.447973   28006 util.go:188] path /go/odo_1/tests/integration/project/devfile.yaml doesn't exist, skipping it
        I0701 12:54:47.448063   28006 util.go:188] path /go/odo_1/tests/integration/project/.devfile.yaml doesn't exist, skipping it
        I0701 12:54:47.596967   28006 kclient.go:201] Checking if "projects" resource is supported
         ✗  The project "something" does not exist. Please check the list of projects using `odo project list`
        
    to contain substring
        <string>: Unauthorized to access the cluster
```

</details>

Currently, this PR replaces `DiscoveryClient` with `CachedDiscoveryClient`, which causes the entire odo code to use the replaced client. I'll change the behaviour such that both clients are available in odo code, so that functions can choose the client that they would prefer to use instead of being forced to use cached data each time.

pkg/kclient/kclient.go
```diff
@@ -158,6 +161,12 @@ func NewForConfig(config clientcmd.ClientConfig) (client *Client, err error) {
 		return nil, err
 	}
 
+	config_flags := genericclioptions.NewConfigFlags(true)
+	client.cachedDiscoveryClient, err = config_flags.ToDiscoveryClient()
```
feloy, line 165: Do we want to cache between calls to `odo`, or only during an odo execution.

If we want to cache only during an odo execution, it seems that we could call `Invalidate` just after creating the client

dharmit: > If we want to cache only during an odo execution, it seems that we could call `Invalidate` just after creating the client

If we were to invalidate the cache each time we create a client, I think the gain we see in `odo list` performance in this PR would be gone. But I need to check that to be sure.

dharmit: > I think the gain we see in `odo list` performance in this PR would be gone. But I need to check that to be sure.

12s without invalidating the cache, 16s if we invalidate it right after creation.
```sh
$ odo list
 ✓  Listing components from namespace 'myproject' [12s]
 NAME                     PROJECT TYPE  RUNNING IN  MANAGED 
 * devfile-nodejs-deploy  nodejs        Deploy      odo     

$ odo list
 ✓  Listing components from namespace 'myproject' [16s]
 NAME                     PROJECT TYPE  RUNNING IN  MANAGED 
 * devfile-nodejs-deploy  nodejs        Deploy      odo
```

feloy: What is the client caching? the discovered resources (CRD level), or also the resources themselves (CR level)?

How is the cache consistent: Do we always get the latest versions of what we query (at the price of some back-and-forths to know if the cached one is the last one), of is there some period during which we get old versions?

dharmit: > What is the client caching? the discovered resources (CRD level), or also the resources themselves (CR level)?

You can look at the cache under `~/.kube/cache/discovery`. I took the cue from `kubectl api-resources cached=true` command.



> How is the cache consistent: Do we always get the latest versions of what we query (at the price of some back-and-forths to know if the cached one is the last one), of is there some period during which we get old versions?

There's a ttl, beyond which it falls back to using regular discovery client:
https://github.com/kubernetes/client-go/blob/master/discovery/cached/disk/cached_discovery.go#L155-L157

dharmit: > What is the TTL used?

Screenshot from 2022-07-06 12-59-26

I think it's the amount of nanoseconds you see in that screenshot.

> I think some documentation on this behaviour will be necessary, for the user to understand why `odo list` (and other commands) does not return the same values in the same circumstances

OK. But I don't follow you. Can you share an example of the `odo list` behaviour you're describing? As far as this PR is concerned, it's only caching the API resources. It's still making calls to the server to know if there are any instances of those resources. So, unless a user's adding an API resource within the TTL, I don't understand why odo would return differing values in any circumstances. However, as I said, I don't follow you entirely, so an example would be helpful.

feloy: 10 minutes, right? 
The output could differ after installing a new operator / CRD and deploying resources of this kind. These new resources should not be visible during 10 minutes

feloy: Yes, and if possible, what are the solutions if the user doesn't want to wait for 10 minutes

feloy: > > what are the solutions if the user doesn't want to wait for 10 minutes
> 
> We could document that the user could remove the cache directory from their system using `rm -rf` if they don't want to wait. However, this would be relevant for users using minikubs/CRC/kinD. Folks in the more realistic environment might not even have the permissions to install an Operator/CRD.
> 
> For the time being, I will add this info in `odo list` docs in this PR.

We discussed this topic when being in Paris, with Serena and the team. odo is dedicated to two different kind of people:

- the "lead" developer, who will experiment odo, create and update the Devfile, test different operators, bindings, etc. He will probably have admin access to the cluster to make experiments (on its own CRC cluster for example)
- the "normal" developer, who will have a tailor-made environment (prepared by the first "lead" developer"), and with minimal access to cluster.

The conclusion, as I remember, were that we need mostly to document for the "lead" developer, with the maximum of information for him to understand how to create and update the Devfile efficiently. The doc for the "normal" developer would be very simple, and would be prepared by the "lead" developer.

feloy: Yes, I think so. Don't forget to say it is a cache common to the kubectl command

dharmit: OK.

dharmit: https://deploy-preview-5908--odo-docusaurus-preview.netlify.app/docs/command-reference/list

PR #5912, opened 2022-07-04
Allow to set `namingStrategy` when using `odo add binding`

### LINKED ISSUE #5816: Ability to set `namingStrategy` when using interactive `odo add binding`
/kind feature

I would like the ability when using the interactive `odo add binding` command to set the `.spec.namingStrategy` for the `ServiceBinding`.

## Why is this needed?

I would like to be able to set the naming strategy to `none` to make sure the default naming strategy is not applied and the Service Binding name will not be appended to the environment variable name. I am guessing other people may prefer the other strategies so I think it would be a good idea to have the other options available in the prompts.

rm3l:
**What type of PR is this:**
/kind feature

This PR makes it possible to set the `namingStrategy` field when using `odo add binding`.
For consistency, this supports both interactive and non-interactive modes.

Fixes #5816

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] [Documentation](https://deploy-preview-5912--odo-docusaurus-preview.netlify.app/)

- Interactive mode: new questions added when running `odo add binding`
- Non-Interactive mode: new `--naming-strategy` flag added to `odo add binding`

### DIFF
--- a/docs/website/docs/command-reference/add-binding.md
+++ b/docs/website/docs/command-reference/add-binding.md
@@ -43,12 +43,15 @@ In the non-interactive mode, you will have to specify the following required inf
 * `--workload` flag to specify the workload resource, if a Devfile is not present in the directory,
 * `--name` flag to specify a name for the binding (see [Understanding Bind as Files](#understanding-bind-as-files) for more information on this)
 * `--bind-as-files` flag to specify if the service should be bound as a file; this flag is set to true by default.
-
+* `--naming-strategy` flag to specify the naming strategy to use for binding names. This flag is empty by default, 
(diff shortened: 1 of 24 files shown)


feloy:
I think you could add the information to `odo describe binding` (it is already present in the JSON output), and update the docs to add this field to the outputs

rm3l:
> I think you could add the information to `odo describe binding` (it is already present in the JSON output), and update the docs to add this field to the outputs

I just updated the output of `odo describe naming` to include such information. This led me to also update the related docs, as well as the integration tests for both `odo describe binding` and `odo list binding`.
Also had to rebase onto `main` to fix a conflict.
Please take another look when you get a chance.

feloy (approved):

PR #5915, opened 2022-07-04
Blog: Connecting to a service without SBO

### LINKED ISSUE #5771: Docs:  add documentation on how to connect to a service without using SBO
This can be in form of guide or example and it could even use helm instead of operators.

The importance of this is to remind users that they can use their regular flow with "hardcoding" connection information to env variables in devfile.yaml.

valaparthvi:

**What type of PR is this:**
/kind documentation

Fixes #5771 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [x] Documentation 

https://deploy-preview-5915--odo-docusaurus-preview.netlify.app/blog/binding-database-service-without-sbo

### DIFF
--- a/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
+++ b/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
@@ -20,7 +20,7 @@ We have a simple CRUD application built in Go that can create/list/update/delete
 ## Prerequisites:
 This blog assumes:
 - [odo v3.0.0-beta1](https://github.com/redhat-developer/odo/releases/tag/v3.0.0-beta1)
-- you have access to a Kubernetes or OpenShift cluster to be able to install the MongoDB operator.
+- you have access to a Kubernetes or OpenShift cluster.
 - you have _Helm_ installed on your system. See https://helm.sh/docs/intro/install/ for installation instructions.
 
 ## (Optional) Setting up the namespace
@@ -39,11 +39,11 @@ helm repo add bitnami https://charts.bitnami.com/bitnami && helm repo update
 
 2. Export the necessary environment variables:
 ```sh
-export MY_MONGODB_ROOT_USERNAME=root
-export MY_MONGODB_ROOT_PASSWORD=my-super-secret-root-password
-export MY_MONGODB_USERNAME=my-app-username
-export MY_MONGODB_PASSWORD=my-app-super-secret-password
-export MY_MONGODB_DATABASE=my-app
+MY_MONGODB_ROOT_USERNAME=root
+MY_MONGODB_ROOT_PASSWORD=my-super-secret-root-password
+MY_MONGODB_USERNAME=my-app-username
+MY_MONGODB_PASSWORD=my-app-super-secret-password
+MY_MONGODB_DATABASE=my-app
 ```
 
 3. Create the MongoDB service.
--- a/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
+++ b/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
@@ -15,11 +15,11 @@ There are a few ways of binding your application to a database service with the 
 
 
 ## Architecture
-We have a simple CRUD application built in Go that can create/list/update/delete a place. This application requires connecting to a mongodb database in order to function correctly, which will be deployed as a microservice on the cluster.
+We have a simple CRUD application built in Go that can create/list/update/delete a place. This application requires connecting to a MongoDB database in order to function correctly, which will be deployed as a microservice on the cluster.
 
 ## Prerequisites:
 This blog assumes:
-- you have access to a Kubernetes or OpenShift cluster.
+- you have admin access to a Kubernetes or OpenShift cluster to be able to install the MongoDB operator.
 - you have _Helm_ installed on your system. See https://helm.sh/docs/intro/install/ for installation instructions.
 
 ## Setting up the application
@@ -35,8 +35,8 @@ odo create namespace restapi-mongodb
 ```
 
 
-## Setting up the mongodb microservice
-We are going to use the Percona's operator for creating our mongodb database. For the sake of simplicity, we will use Helm to deploy our mongodb operator and the service.
+## Setting up the MongoDB microservice
+We are going to use the Percona's operator for creating our MongoDB database. For the sake of simplicity, we will use Helm to deploy our MongoDB operator and the service.
 
 3. Add the Percona’s Helm charts repository and make your Helm client up to date with it:
 ```sh
@@ -97,7 +97,7 @@ my-op-psmdb-operator-69d88f479c-7hj5m   1/1     Running   0          3m34s
 odo init --devfile go --name places
 ```
 
-7. Edit the 'runtime' component container in devfile to add information such as username, password, and host required to connect to the mongodb service.
+7. Edit the 'runtime' container component in devfile to add information such as username, password, and host required to connect to the MongoDB service.
 ```yaml
 components:
 - container:
@@ -113,8 +113,10 @@ components:
   name: runtime
 ```
 
-The _username_, and _password_ values are hard-coded here as a part of the helm deployment, but their value can be obtained from the secret resource called `minimal-cluster` that was deployed via our local helm chart `psmdb-db`.
-The value for _host_ is name of the service that belongs to our database application.
+The _username_, and _password_ values are hard-coded here as a part of the helm deployment, but their value can be obtained from the secret resource called "minimal-cluster" that was deployed via our local helm chart "psmdb-db".
+
+The value for _host_ is name of the service that belongs to our database application, in this case it is a service resource called "minimal-cluster-mongos".
+Optionally, you can run `kubectl get psmdb/minimal-cluster -ojsonpath='{.status.host}'` to obtain the host's value.
 
 ## Deploy the application
 8. Run `odo dev` to deploy the application on the cluster.
--- a/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
+++ b/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
@@ -136,7 +136,7 @@ There are 3 changes that we will need to make to our devfile:
 ```yaml
 schemaVersion: 2.2.0
 ```
-Please note that this change is only necessary because we are using [devfile variable substitution](../docs/command-reference/dev.md#substituting-variables).
+Please note that this change is only necessary because we are using [devfile variable substitution](/docs/command-reference/dev#substituting-variables).
 
 6.2 Add a `variables` field in the devfile.
 ```yaml
@@ -298,4 +298,4 @@ To conclude this blog, it is possible to connect your application with another m
 
 ### Related articles on binding:
 * [Binding an external service with odo v3](./2022-06-14-binding-external-service.md)
-* [odo add binding](../docs/command-reference/add-binding.md)
+* [odo add binding](/docs/command-reference/add-binding)
--- a/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
+++ b/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
@@ -269,7 +269,10 @@ starterProjects:
 ## Deploy the application
 7. Run `odo dev` to deploy the application on the cluster.
 ```sh
-odo dev --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD --var USERNAME=$MY_MONGODB_ROOT_USERNAME  --var HOST="mongodb"
+odo dev \
+  --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD \
+  --var USERNAME=$MY_MONGODB_ROOT_USERNAME \
+  --var HOST="mongodb"
 ```
 
 The value for _host_ is name of the service that belongs to our database application, in this case it is a service resource called "mongodb", you might have noticed it when we deployed the helm chart.
--- a/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
+++ b/docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
@@ -1,0 +1,170 @@
+---
+title: Binding to a database service without SBO
+author: Parthvi Vala
+author_url: https://github.com/valaparthvi
+author_image_url: https://github.com/valaparthvi.png
+tags: ["binding"]
+slug: binding-database-service-without-sbo
+---
+
+How to bind to your application to a database service without SBO.
+
+<!--truncate-->
+
+There are a few ways of binding your application to a database service with the help of odo. The recommended way is with the help of Service Binding Operator, but it is also possible to bind without it, and this blog will show you how.
+
+
+## Architecture
+We have a simple CRUD application built in Go that can create/list/update/delete a place. This application requires connecting to a mongodb database in order to function correctly, which will be deployed as a microservice on the cluster.
+
+## Prerequisites:
+This blog assumes:
+- you have access to a Kubernetes or OpenShift cluster.
+- you have _Helm_ installed on your system. See https://helm.sh/docs/intro/install/ for installation instructions.
+
+## Setting up the application
+1. Clone the repository, and cd into it.
+```sh
+git clone https://github.com/valaparthvi/restapi-mongodb-odo.git && cd restapi-mongodb-odo
+```
+
+## Setting up the namespace
+2. We will create a new namespace to deploy our application in, with the help of odo.
+```sh
+odo create namespace restapi-mongodb
+```
+
+
+## Setting up the mongodb microservice
+We are going to use the Percona's operator for creating our mongodb database. For the sake of simplicity, we will use Helm to deploy our mongodb operator and the service.
+
+3. Add the Percona’s Helm charts repository and make your Helm client up to date with it:
+```sh
+helm repo add percona https://percona.github.io/percona-helm-charts/ && helm repo update
+```
+
+4. Install Percona Operator for MongoDB:
+```sh
+helm install my-op percona/psmdb-operator
+```
+Wait for the pod to come up:
+```sh
+$ kubectl get pods
+NAME                                    READY   STATUS    RESTARTS   AGE
+my-op-psmdb-operator-69d88f479c-7hj5m   1/1     Running   0          3m34s
+```
+
+5. Install Percona server for MongoDB from our local `psmdb-db` Helm chart.
+```sh
+helm install my-db ./psmdb-db
+```
+
+```sh
+$ helm install my-db ./psmdb-db
+
+NAME: my-db
+LAST DEPLOYED: Mon Jul  4 13:49:26 2022
+NAMESPACE: restapi-mongodb
+STATUS: deployed
+REVISION: 1
+TEST SUITE: None
+NOTES:
+To get a MongoDB prompt inside your new cluster you can run:
+
+  ADMIN_USER=$(kubectl -n restapi-mongodb get secrets minimal-cluster -o jsonpath="{.data.MONGODB_USER_ADMIN_USER}" | base64 --decode)
+  ADMIN_PASSWORD=$(kubectl -n restapi-mongodb get secrets minimal-cluster -o jsonpath="{.data.MONGODB_USER_ADMIN_PASSWORD}" | base64 --decode)
+
+And then for replica set:
+  $ kubectl run -i --rm --tty percona-client --image=percona/percona-server-mongodb:4.4 --restart=Never -- mongo "mongodb+srv://${ADMIN_USER}:${ADMIN_PASSWORD}@minimal-cluster-rs0.restapi-mongodb.svc.cluster.local/admin?replicaSet=rs0&ssl=false"
+
+Or for sharding setup:
+  $ kubectl run -i --rm --tty percona-client --image=percona/percona-server-mongodb:4.4 --restart=Never -- mongo "mongodb://${ADMIN_USER}:${ADMIN_PASSWORD}@minimal-cluster-mongos.restapi-mongodb.svc.cluster.local/admin?ssl=false"
+```
+
+Wait for the pods to come up, this might take a few minutes:
+```sh
+$ kubectl get pods
+NAME                                    READY   STATUS    RESTARTS   AGE
+minimal-cluster-cfg-0                   1/1     Running   0          3m6s
+minimal-cluster-mongos-0                1/1     Running   0          3m5s
+minimal-cluster-rs0-0                   1/1     Running   0          3m5s
+my-op-psmdb-operator-69d88f479c-7hj5m   1/1     Running   0          3m34s
+```
+
+## Download the devfile.yaml
+6. Run `odo init` to fetch the necessary devfile.
+```sh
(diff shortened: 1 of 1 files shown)


dharmit:
The blog's not showing a single instance of `odo add binding`. How is it justified as a blog that uses odo to connect to a service without SBO? It's using the service discovery mechanism used since a long time - manually populating the environment variables.

valaparthvi:
> The blog's not showing a single instance of `odo add binding`. How is it justified as a blog that uses odo to connect to a service without SBO? It's using the service discovery mechanism used since a long time - manually populating the environment variables.

It is not supposed to use SBO to connect, hence there is no `odo add binding`. The purpose of this blog is to show that you can connect to a service without the help of SBO. You can read the attached issue for more details.

feloy (changes requested):

dharmit:
> It is not supposed to use SBO to connect, hence there is no `odo add binding`. The purpose of this blog is to show that you can connect to a service without the help of SBO. You can read the attached issue #5771 for more details.

Yeah, I read that issue right after commenting. It seems like we only want to show that odo can create a Deployment, and a Pod of that Deployment can talk to a Service using env vars. Anyway, it seems like that's what @kadel wants to showcase.

However, the flow here is not the same as:
```sh
$ odo init
$ odo dev

# Now create a service (using helm or operators)
# Next modify the existing devfile to hardcodde service info, and let odo dev do the magic
```

dharmit (changes requested):

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,170 @@
...
+```yaml
+components:
+- container:
+    ...
+    ...
+    env:
+    - name: username
+      value: userAdmin
+    - name: password
+      value: userAdmin123456
```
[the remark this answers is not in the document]
valaparthvi: Does devfile support dynamically obtaining values? I mean if we can put `kubectl get secrets minimal-secret -ojsonpath='{.data.username}' | base64 -d` inside the devfile, we can avoid having to hard code password.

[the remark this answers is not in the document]
valaparthvi: I could add a note acknowledging that it's a bad practice, but this is for example purpose only, or I could find a way to fetch password from the code itself to avoid this practice altogether.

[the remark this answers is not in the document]
feloy: Yes, you could start with this solution, by indicating from the very beginning this is not a secure solution, and then continue with more secure solutions (and probaly more difficult ones). That would show the interest of using SBO.

- The simplest solution would be to have an `envFrom` field in the Devfile's container, but this field does not exist. Is it worth asking Devfile team to add it? I'm not sure
- You could pass the secret name and the secret keys as environment variables, and the program would be responsible for getting the values from the secret. That means the program needs access to the Kubernetes API, which opens security breaches

dharmit:
> However, the flow here is not the same as:
> 
> ```shell
> $ odo init
> $ odo dev
> 
> # Now create a service (using helm or operators)
> # Next modify the existing devfile to hardcodde service info, and let odo dev do the magic
> ```

@valaparthvi please make sure that your blog complies with this. At the moment, `odo init` is on Step 6. I would prefer it be Step 1 right after cloning the repo and creating a namespace (which should be mentioned as "optional", IMO.)

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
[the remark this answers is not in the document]
valaparthvi: >The simplest solution would be to have an envFrom field in the Devfile's container, but this field does not exist. Is it worth asking Devfile team to add it? I'm not sure

When you say `envFrom`, do you mean devfile can evaluate the value from the environment variable exposed in your local system?
```yaml
env:
- name: password
  value:
    envFrom: $MONGODB_PASSWORD
```
Something like this? Or more like this https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/pods/inject/pod-secret-envFrom.yaml?


>You could pass the secret name and the secret keys as environment variables, and the program would be responsible for getting the values from the secret. That means the program needs access to the Kubernetes API, which opens security breaches

I think asking the program to be responsible for fetching everything would defeat the purpose of this blog. I think I'll go with a note for now, and it's like you said, it might encourage users to try SBO.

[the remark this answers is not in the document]
feloy:
```
> 
> Something like this? Or more like this https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/pods/inject/pod-secret-envFrom.yaml?
> 

Yes, like this one

**Comment by dharmit** [comment]
@valaparthvi thanks for the fixes.

/approve

**Review by rm3l — COMMENTED** [review]
@valaparthvi Great blog post 👍🏿 

I added a few comments below.

Another small comment. The SBO acronym is mentioned all over the place, but I saw nowhere where it was defined. It is clear to us, but might not for readers.
Maybe it could be worth changing the title to something more explicit, like "Binding to a database service without the Service Binding Operator (SBO)".

My 2 cents...

File: docs/website/blog/2022-06-30-binding-database-service-without-sbo.md (hunk 1 above)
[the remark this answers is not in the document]
Comment (valaparthvi) [reply docs/website/blog/2022-06-30-binding-database-service-without-sbo.md:146]: But then devfile would have to interact with the k8s API and that does not make a lot of sense to me from the design standpoint. I don't think devfile is dependent on the K8s API right now, and I'm not sure if they'd be willing to create a dependency now. Can we take this up in cabal?

File: docs/website/blog/2022-06-30-binding-database-service-without-sbo.md [hunk 2]
```diff
@@ -0,0 +1,326 @@
...
+3. Run `odo init` to fetch the necessary devfile.
+```sh
+odo init --devfile go --name places
+```
+
+
+## Deploy the application
+4. Run `odo dev` to deploy the application on the cluster.
+```sh
+odo dev --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD --var USERNAME=$MY_MONGODB_ROOT_USERNAME
```
feloy, line 60: Same here, I think it is premature to pass these variables, as they are not existing in the Devfile yet and we are not using them. I think it would be better to wait for steps 7.x before to do this

feloy, line 60: Same here, I think it is premature to pass these variables, as they are not existing in the Devfile yet and we are not using them. I think it would be better to wait for steps 7.x before to do this

cdrage:
Why blog? Shouldn't this be also under Advanced Usage in the user guides?

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Setting up the application
+1. Clone the repository, and cd into it.
+```sh
+git clone https://github.com/valaparthvi/restapi-mongodb-odo.git && cd restapi-mongodb-odo
+```
+
+
+## Export environment variables
+2. Export the necessary environment variables:
```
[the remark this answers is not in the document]
valaparthvi: I agree, but then if we want to maintain the order of `odo init` > `odo dev` > deploy charts > wait for odo dev to detect changes, it becomes necessary to export these variables at this point

valaparthvi:
> Why blog? Shouldn't this be also under Advanced Usage in the user guides?

I don't think it's an advanced usage (wrt odo) use case. It's more like a workaround blog.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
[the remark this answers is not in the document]
feloy: The order 
- odo init
- odo dev
- deploy chart
is just an example provided by Tomas, I'm not sure it must be respected at all price.

By first deploying the chart (including exportint variables), then running odo init and odo dev, the flow seems more understandable to me

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
+---
+title: Binding to a database service without SBO
+author: Parthvi Vala
+author_url: https://github.com/valaparthvi
+author_image_url: https://github.com/valaparthvi.png
+tags: ["binding"]
+slug: binding-database-service-without-sbo
+---
+
+How to bind to your application to a database service without the Service Binding Operator.
```
rm3l, line 10:
```suggestion
How to bind your application to a database service without the Service Binding Operator.
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
+---
+title: Binding to a database service without SBO
```
rm3l, line 2: How about making the SBO acronym explicit in the page title?

```suggestion
title: Binding to a database service without the Service Binding Operator
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
...
+export MY_MONGODB_USERNAME=my-app-username
+export MY_MONGODB_PASSWORD=my-app-super-secret-password
+export MY_MONGODB_DATABASE=my-app
+```
+
+
+## Download the devfile.yaml
+3. Run `odo init` to fetch the necessary devfile.
+```sh
+odo init --devfile go --name places
```
rm3l, line 53: There is already a `devfile.yaml` file after cloning your repo. So running `odo init` complains:
```
❯ odo init --devfile go --name places
 ✗  a devfile already exists in the current directory
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
...
+- container:
+    ...
+    ...
+    env:
+    - name: username
+      value: "{{USERNAME}}"
+    - name: password
+      value: "{{PASSWORD}}"
+    - name: host
+      value: "mongodb"
```
rm3l, line 201: The database host could be a variable too, I think. Its default value could still be `mongodb` though.

rm3l (changes requested):
One last comment. I think it would also be great to include `odo list` after deploying the Bitnami Helm Chart, just to show that odo can list Helm-managed resources as well. What do you think?

```
❯ odo list
 ✓  Listing components from namespace 'blog-helm-wo-sbo-5915' [2s]
 NAME       PROJECT TYPE  RUNNING IN  MANAGED 
 * restapi  go            Dev         odo     
 mongodb    Unknown       None        Helm    
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,326 @@
...
+There are a few ways of binding your application to a database service with the help of odo. The recommended way is with the help of Service Binding Operator(SBO), but it is also possible to bind without it, and this blog will show you how.
+
+
+## Architecture
+We have a simple CRUD application built in Go that can create/list/update/delete a place. This application requires connecting to a MongoDB database in order to function correctly, which will be deployed as a microservice on the cluster.
+
+## Prerequisites:
+This blog assumes:
+- [odo v3.0.0-beta1](https://github.com/redhat-developer/odo/releases/tag/v3.0.0-beta1)
+- you have admin access to a Kubernetes or OpenShift cluster to be able to install the MongoDB operator.
```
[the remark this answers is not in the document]
rm3l: Indeed, this should not be necessary.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,301 @@
...
+There are a few ways of binding your application to a database service with the help of odo. The recommended way is with the help of Service Binding Operator(SBO), but it is also possible to bind without it, and this blog will show you how.
+
+
+## Architecture
+We have a simple CRUD application built in Go that can create/list/update/delete a place. This application requires connecting to a MongoDB database in order to function correctly, which will be deployed as a microservice on the cluster.
+
+## Prerequisites:
+This blog assumes:
+- [odo v3.0.0-beta1](https://github.com/redhat-developer/odo/releases/tag/v3.0.0-beta1)
+- you have access to a Kubernetes or OpenShift cluster to be able to install the MongoDB operator.
```
feloy, line 23:
```suggestion
- you have access to a Kubernetes or OpenShift cluster to be able to install the MongoDB database.
```

But I'm not sure this sentence makes sense, now you've removed the "admin" access necessity: installing the MongoDB database does not need more rights than deploying the app with `odo`

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,301 @@
...
+We are going to use the Bitnami's helm charts for creating our MongoDB database.
+
+1. Add the Bitnami's Helm charts repository and make your Helm client up to date with it:
+```sh
+helm repo add bitnami https://charts.bitnami.com/bitnami && helm repo update
+```
+
+2. Export the necessary environment variables:
+```sh
+export MY_MONGODB_ROOT_USERNAME=root
```
feloy, line 42: I'm pretty sure the exports are not necessary.

They would be necessary if you used `odo dev --var MY_MONGODB_ROOT_USERNAME`, but using `odo dev --var PASSWORD=$MY_MONGODB_ROOT_USERNAME `, you just need to declare the variables into the shell session, without exporting them.

Note that exporting a variable (especially a password) into a shell gives access to it from *all* the processes you start from your shell. This can be a security breach, it must be used cautiously.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,301 @@
...
+
+** Please be patient while the chart is being deployed **
+
+MongoDB&reg; can be accessed on the following DNS name(s) and ports from within your cluster:
+
+    mongodb.restapi-mongodb.svc.cluster.local
+
+To get the root password run:
+
+    export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace restapi-mongodb mongodb -o jsonpath="{.data.mongodb-root-password}" | base64 -d)
```
feloy, line 86: Is it different from MY_MONGODB_ROOT_PASSWORD?

valaparthvi: No, it's not.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
valaparthvi: It should be fixed now, I'd accidentally pushed devfile.yaml

valaparthvi:
> One last comment. I think it would also be great to include `odo list` after deploying the Bitnami Helm Chart, just to show that odo can list Helm-managed resources as well. What do you think?
> 
> ```
> ❯ odo list
>  ✓  Listing components from namespace 'blog-helm-wo-sbo-5915' [2s]
>  NAME       PROJECT TYPE  RUNNING IN  MANAGED 
>  * restapi  go            Dev         odo     
>  mongodb    Unknown       None        Helm    
> ```

I think it might be stirring a little away from the topic at hand. I would rather do that in a separate blog if necessary.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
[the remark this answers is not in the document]
dharmit: > is just an example provided by Tomas, I'm not sure it must be respected at all price.
> 
> 
> By first deploying the chart (including exportint variables), then running odo init and odo dev, the flow seems more understandable to me

I like the current approach of enabling Helm repo first. But I don't agree with the idea of spinning up Helm chart using it before doing `odo init`. If it were me, I would first show that your Go application is deployed using `odo dev`, but `curl` request made to it is throwing 404 or some error. Thereafter, we spin up the MongoDB database required by the application.

Or maybe, do an `odo dev` without all the variables which would help one connect to the MongoDB database. Make a `curl` call that returns 404 or some error. Now add information about connection to MongoDB and see how the application returns expected output.

I'm saying this because that's how I think an application developer's flow would be like. But then I have not a lot of application development myself, so correct me if I'm mistaken.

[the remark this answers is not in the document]
valaparthvi: > Or maybe, do an `odo dev` without all the variables which would help one connect to the MongoDB database. Make a `curl` call that returns 404 or some error. Now add information about connection to MongoDB and see how the application returns expected output.
> 
> I'm saying this because that's how I think an application developer's flow would be like. But then I have not a lot of application development myself, so correct me if I'm mistaken.

I agree with you, but I can't find a flow that works for everyone. For now, I have added a note stating that `odo dev` will fail if run before adding connection information.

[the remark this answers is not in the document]
dharmit: Works for me. I discussed this with @feloy and agree with the flow we have now.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,335 @@
...
+    checkoutFrom:
+      revision: main
+    remotes:
+      origin: https://github.com/devfile-samples/devfile-stack-go.git
+  name: go-starter
+```
+</details>
+
+
+## Deploy the application
```
feloy, line 269: You would need to execute this command at some point before:

```
odo set namespace restapi-mongodb
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,335 @@
...
+    workingDir: ${PROJECT_SOURCE}
+  id: run
+components:
+- container:
+    dedicatedPod: false
+    endpoints:
+    - name: port-3000-tcp
+      protocol: tcp
+      secure: false
+      targetPort: 3000
```
feloy, line 231: The default port is 8080 in the Devfile I downloaded with `odo init`. It owuld be helpful to indicate to the reader to change it

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,335 @@
...
+
+### List of available API endpoints
+- GET `/api/places` - List all places
+- POST `/api/places` - Add a new place
+- PUT `/api/places` - Update a place
+- GET `/api/places/<id>` - Fetch place with id `<id>`
+- DELETE `/api/places/<id>` - Delete place with id `<id>`
+
+## Conclusion
+To conclude this blog, it is possible to connect your application with another microservice without the Service Binding Operator if you have the correct connection information. Using the Service Binding Operator with a [Bindable Operator](https://github.com/redhat-developer/service-binding-operator#known-bindable-operators) makes it easy for you to not care about finding the connection information and ease the binding.
```
feloy, line 331: You could add to the conclusion that it is up to the reader to manage the secrets stored in the environment variables MY_MONGODB_ROOT_PASSWORD and MY_MONGODB_ROOT_USERNAME, so they are accessible every time the following command is executed:

```
odo dev --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD --var USERNAME=$MY_MONGODB_ROOT_USERNAME --var HOST="mongodb"
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,335 @@
...
+MY_MONGODB_DATABASE=my-app
+```
+
+3. Create the MongoDB service.
+```sh
+helm install mongodb bitnami/mongodb \
+  --set auth.rootPassword=$MY_MONGODB_ROOT_PASSWORD \
+  --set auth.username=$MY_MONGODB_USERNAME \
+  --set auth.password=$MY_MONGODB_PASSWORD \
+  --set auth.database=$MY_MONGODB_DATABASE
```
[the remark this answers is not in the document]
feloy: My bad, I used `kubectl create namespace ...` instead, you can ignore most of my comments :(

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,335 @@
...
+      origin: https://github.com/devfile-samples/devfile-stack-go.git
+  name: go-starter
+```
+</details>
+
+
+## Deploy the application
+7. Run `odo dev` to deploy the application on the cluster.
+```sh
+odo dev --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD --var USERNAME=$MY_MONGODB_ROOT_USERNAME  --var HOST="mongodb"
```
rm3l, line 272: Surely nitpicking, but would it be possible to display this command under multiple lines? Otherwise, we need to scroll horizontally to see the full command line:

```suggestion
odo dev \
  --var PASSWORD=$MY_MONGODB_ROOT_PASSWORD \
  --var USERNAME=$MY_MONGODB_ROOT_USERNAME \
  --var HOST="mongodb"
```

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
valaparthvi: I've added this note in the section where we introduce these variables for the first time, I think it makes more sense to put them there than in the conclusion.

feloy:
/lgtm

/hold

Feel free to unhold if other reviewers are ok

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
rm3l: Instead of changing the port in the Devfile, wouldn't it be more straightforward to change the default port in the [application source code](https://github.com/valaparthvi/restapi-mongodb-odo/blob/main/config.yml#L2) to use 8080 as default port instead? This would mean one less manual operation for the readers.

docs/website/blog/2022-06-30-binding-database-service-without-sbo.md
```diff
@@ -0,0 +1,395 @@
...
+    - name: password
+      value: "{{PASSWORD}}"
+    - name: host
+      value: "{{HOST}}"
+  name: runtime
+```
+
+The values for _username_, _password_, and _host_ will be passed to devfile.yaml with the `--var` flag when we run the `odo dev` command.
+
+6.4 If you have not already set the 'runtime' container component port to 3000, now is the time to do so.
```
rm3l, line 246: Instead of changing the port in the Devfile, wouldn't it be more straightforward to change the default port in the [application source code](https://github.com/valaparthvi/restapi-mongodb-odo/blob/main/config.yml#L2) to use 8080 as default port instead? This would mean one less manual operation for the readers.

valaparthvi: You're right.

rm3l (approved):
/hold cancel

valaparthvi:

feloy:

PR #5930, opened 2022-07-07
[WIP] Fix M1 bug with tmp files

cdrage:
**What type of PR is this:**

/kind bug

Fixes the odd bug with /private/var/folders being added to tmp files (it's an alias
for /var/folders with M1)

Fixes https://github.com/redhat-developer/odo/issues/5927

### DIFF
--- a/pkg/envinfo/envinfo.go
+++ b/pkg/envinfo/envinfo.go
@@ -64,7 +64,8 @@ type EnvSpecificInfo struct {
 var _ localConfigProvider.LocalConfigProvider = (*EnvSpecificInfo)(nil)
 
 func (esi EnvSpecificInfo) GetDevfilePath() string {
-	return esi.devfilePath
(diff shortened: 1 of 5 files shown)


cdrage:
This is WIP just so I can work against the IBM cluster with this.

pkg/envinfo/envinfo.go
```diff
@@ -64,7 +64,8 @@ type EnvSpecificInfo struct {
 var _ localConfigProvider.LocalConfigProvider = (*EnvSpecificInfo)(nil)
 
 func (esi EnvSpecificInfo) GetDevfilePath() string {
-	return esi.devfilePath
+	// Return esi.devfilePath without /private as the base folder
+	return strings.Replace(esi.devfilePath, "/private", "", 1)
```
kadel, line 68: This looks really really hacky. We should try to investigate what is happening here and why is this there.

Issue #5879, opened 2022-06-27
Unit tests failing on PRs


anandrkskd:
/area testing
/priority High

Unit tests on PR are failing with error

```make goget-tools
make[1]: Entering directory `/go/src/github.com/redhat-developer/odo'
mkdir -p /go/bin
curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /go/bin v1.37.0
golangci/golangci-lint info checking GitHub for tag 'v1.37.0'
golangci/golangci-lint crit unable to find 'v1.37.0' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
make[1]: *** [goget-tools] Error 1
make[1]: Leaving directory `/go/src/github.com/redhat-developer/odo'
make: *** [openshiftci-presubmit-unittests] Error 2```

anandrkskd:
will get fixed by:  https://github.com/openshift/release/pull/30052

kadel:
> will get fixed by: [openshift/release#30052](https://github.com/openshift/release/pull/30052)

Do you know what is that error happening. It is strange as the same command always works correctly on my local machines (tested on mac and linux)

anandrkskd:
> Do you know what is that error happening. It is strange as the same command always works correctly on my local machines (tested on mac and linux)

I tried to debug it, but found no useful info. 
so I just tried to do the fixing instead.

PR #5934, opened 2022-07-08
download golangci-lint binary with test script

anandrkskd:

**What type of PR is this:**

/area testing

Removing golangci-lint from the script, as the installation is added as [pre step](https://github.com/anandrkskd/release/blob/0692f2611817db9766dd338317b5627e4c4b52cf/ci-operator/config/redhat-developer/odo/redhat-developer-odo-main.yaml#L14) in unit test set on prow, it downloads the binary directly for the test.

Fixes [#5879](https://github.com/redhat-developer/odo/issues/5879)

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/openshiftci-presubmit-unittests.sh
+++ b/scripts/openshiftci-presubmit-unittests.sh
@@ -11,6 +11,18 @@ export PATH=$PATH:$GOPATH/bin
 # otherwise /.cache is used, and it fails on permission denied
 export GOLANGCI_LINT_CACHE="/tmp/.cache"
 
+ARCH=`uname -i`
+if [ "$ARCH" = "x86_64" ]; then
+  export GOARCH=amd64
+else
+  export GOARCH=$ARCH
+fi
+
+wget https://github.com/golangci/golangci-lint/releases/download/v1.37.0/golangci-lint-1.37.0-linux-$GOARCH.tar.gz
+tar --no-same-owner -xzf golangci-lint-1.37.0-linux-amd64.tar.gz
+mv golangci-lint-1.37.0-linux-amd64/golangci-lint $(go env GOPATH)/bin/golangci-lint
+chmod +x $(go env GOPATH)/bin/golangci-lint
+
 make validate
 make test
 
--- a/scripts/openshiftci-presubmit-unittests.sh
+++ b/scripts/openshiftci-presubmit-unittests.sh
@@ -11,7 +11,7 @@ export PATH=$PATH:$GOPATH/bin
 # otherwise /.cache is used, and it fails on permission denied
 export GOLANGCI_LINT_CACHE="/tmp/.cache"
 
-make goget-tools 
+make goget-tools
 make validate
 make test
 
--- a/scripts/openshiftci-presubmit-unittests.sh
+++ b/scripts/openshiftci-presubmit-unittests.sh
@@ -11,18 +11,7 @@ export PATH=$PATH:$GOPATH/bin
 # otherwise /.cache is used, and it fails on permission denied
 export GOLANGCI_LINT_CACHE="/tmp/.cache"
 
-ARCH=`uname -i`
-if [ "$ARCH" = "x86_64" ]; then
-  export GOARCH=amd64
-else
-  export GOARCH=$ARCH
-fi
-
-wget https://github.com/golangci/golangci-lint/releases/download/v1.37.0/golangci-lint-1.37.0-linux-$GOARCH.tar.gz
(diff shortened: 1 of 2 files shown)


scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -11,7 +11,18 @@ export PATH=$PATH:$GOPATH/bin
...
+if [ "$ARCH" = "x86_64" ]; then
+  export GOARCH=amd64
+else
+  export GOARCH=$ARCH
+fi
+
+wget https://github.com/golangci/golangci-lint/releases/download/v1.37.0/golangci-lint-1.37.0-linux-$GOARCH.tar.gz
+tar --no-same-owner -xzf golangci-lint-1.37.0-linux-amd64.tar.gz
+mv golangci-lint-1.37.0-linux-amd64/golangci-lint $(go env GOPATH)/bin/golangci-lint
+chmod +x $(go env GOPATH)/bin/golangci-lint
```
rm3l, line 24: Wouldn't it make more sense to change the `goget-tools` target instead to install `golangci-lint` via `wget` + `tar`? This way, we would have a single place to look at if we have to change the version of `golangci-lint` in the future. And we would be sure to have the exact version locally and in all our test platforms.

rm3l: Also, I think it would be nice to extract the version of `golangci-lint` as a local variable, so it can be easily updated in the future.

rm3l (changes requested):

dharmit:
/approve

@rm3l could you PTAL for the LGTM flag?

scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -11,7 +11,7 @@ export PATH=$PATH:$GOPATH/bin
 # otherwise /.cache is used, and it fails on permission denied
 export GOLANGCI_LINT_CACHE="/tmp/.cache"
 
-make goget-tools
+make goget-tools
```
rm3l, line 14: This extra space may not be needed anymore, to reduce the scope of changes.

rm3l (changes requested):
Added a few comments.

I was also wondering: have we tried any of the other installation methods recommended in the [installation docs](https://golangci-lint.run/usage/install/)? Like:
```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.37.0
```

This might be slower than binary installation, but maybe this would solve the issue with determining the best binary to use without using their `install.sh` script.

rm3l:
> Added a few comments.
> 
> I was also wondering: have we tried any of the other installation methods recommended in the [installation docs](https://golangci-lint.run/usage/install/)? Like:
> 
> ```
> go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.37.0
> ```
> 
> This might be slower than binary installation, but maybe this would solve the issue of determining the best binary to use without using their `install.sh` script.

We could even use this `go install` way as a fallback mechanism if the `install.sh` fails, e.g.:
```shell
(curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.37.0) || \
  go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.37.0
```

anandrkskd:
I can give a try but, I am not sure if using go install is a good step as the doc states it `we've encountered a lot of issues with Go modules hashes.` 

I will use the command you suggested.

rm3l:
> I can give a try but, I am not sure if using go install is a good step as the doc states it `we've encountered a lot of issues with Go modules hashes.`
> 
Ah, I see. In this case, I think it makes even more sense to use this as a last resort. 

> I will use the command you suggested.

Great - let's see how it goes now.. BTW, I think we should also use the tagged version of the `install.sh` script, so we don't have any future surprises in the future. So the command should be this instead, IMO. Extracting the version as a local variable would be even better.

```shell
(curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.37.0/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.37.0) || \
  go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.37.0
```

scripts/openshiftci-presubmit-unittests.sh
rm3l: LGTM, but could you also remove the extra space added here?

rm3l (approved):

valaparthvi:

Test failure is unrelated.

PR #5936, opened 2022-07-11
Allow binding to services from other namespaces

### LINKED ISSUE #5820: `odo add binding` should be able to list service instances from other namespaces
## User Story
- As an odo user
- I want to be able to bind services that are in a different namespace than my application
- So that I can for example share services that are shared between multiple applications or developers.

## Acceptance Criteria
- [x] first question in `odo add binding` is:
  "Do you want to list services from":
  - current namespace
  - all accessible namespaces  
- [x] When `odo add binding` lists service instances it should look beyond just the current namespace. It should check for bindable instances in all namespaces where users can list resources from.
- [x] If a service from other than the current namespace is selected the generated `ServiceBinding` should specify the namespace in the service section
```yaml
        services:
        - group: postgres-operator.crunchydata.com
          id: my-java-springboot-app-hippo
          kind: PostgresCluster
          name: hippo
          namespace: db
          resource: postgresclusters
          version: v1beta1
```
- [x] if service instance from the current namespace is selected it `namespace` should not be specified in `ServiceBinding` definition

rm3l:
**What type of PR is this:**
/kind feature

This PR makes it possible to select or specify any other namespace when listing the service instances to use for adding a new binding.
For consistency, this supports both interactive and non-interactive modes. And output / docs for `odo describe binding` and `odo list binding` (and their JSON counterparts) have been updated to include information about the service namespace, if available.

Fixes #5820

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] [Documentation](https://deploy-preview-5936--odo-docusaurus-preview.netlify.app/) 

- Non-Interactive mode: new `--service-namespace` flag added to `odo add binding`
- Interactive mode: new questions added when running `odo add binding`. They allow picking the namespace from which services instances will be listed
  - Current namespace
[asciicast](https://asciinema.org/a/507965)

  - All accessible namespaces: note that if the user does not have the permissions to list namespaces/projects, a simple warning is displayed, and they are prompted to enter the namespace manually instead.
[asciicast](https://asciinema.org/a/507966)

### DIFF
--- a/pkg/binding/asker/interface.go
+++ b/pkg/binding/asker/interface.go
@@ -8,7 +8,24 @@ const (
 	OutputToFile
 )
 
+type ServiceInstancesNamespaceListOption int
+
+const (
+	CurrentNamespace ServiceInstancesNamespaceListOption = iota + 1
+	AllAccessibleNamespaces
+)
+
 type Asker interface {
+	// SelectNamespaceListOption asks the user to select how service instances should be listed,
+	// i.e. from the current namespace or from all accessible namespaces.
+	// If user selects to list from all accessible namespaces, callers should fetch the list of namespaces
+	// and prompt users to either pick or enter one.
+	// See SelectNamespace and AskNamespace methods.
+	SelectNamespaceListOption() (ServiceInstancesNamespaceListOption, error)
+	// AskNamespace asks the user to enter a namespace
+	AskNamespace() (string, error)
+	// SelectNamespace takes a list of namespaces and asks the user to pick one of them
+	SelectNamespace(options []string) (string, error)
 	// SelectWorkloadResource takes a list of workloads resources and asks user to select one
 	SelectWorkloadResource(options []string) (int, error)
 	// SelectWorkloadResourceName takes a list of workloads resources names and asks user to select one
--- a/pkg/binding/backend/flags.go
+++ b/pkg/binding/backend/flags.go
@@ -15,11 +15,12 @@ import (
 )
 
 const (
-	FLAG_WORKLOAD        = "workload"
-	FLAG_SERVICE         = "service"
-	FLAG_NAME            = "name"
-	FLAG_BIND_AS_FILES   = "bind-as-files"
-	FLAG_NAMING_STRATEGY = "naming-strategy"
+	FLAG_WORKLOAD          = "workload"
+	FLAG_SERVICE           = "service"
+	FLAG_SERVICE_NAMESPACE = "service-namespace"
+	FLAG_NAME              = "name"
+	FLAG_BIND_AS_FILES     = "bind-as-files"
+	FLAG_NAMING_STRATEGY   = "naming-strategy"
 )
 
 // FlagsBackend is a backend that will extract all needed information from flags passed to the command
@@ -57,6 +58,10 @@ func (o *FlagsBackend) SelectWorkloadInstance(workloadName string) (string, sche
 		}
 	}
 	return "", schema.GroupVersionKind{}, fmt.Errorf("group/kind %q not found on the cluster", selectedGroup+"/"+selectedKind)
+}
+
+func (o *FlagsBackend) SelectNamespace(flags map[string]string) (string, error) {
+	return flags[FLAG_SERVICE_NAMESPACE], nil
 }
 
 // SelectServiceInstance parses the service's name, kind, and group from arg:serviceName,
--- a/pkg/binding/backend/interactive.go
+++ b/pkg/binding/backend/interactive.go
@@ -1,11 +1,18 @@
 package backend
 
 import (
+	"fmt"
+	"sort"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog"
 
 	"github.com/redhat-developer/odo/pkg/binding/asker"
 	"github.com/redhat-developer/odo/pkg/kclient"
+	"github.com/redhat-developer/odo/pkg/log"
+	"github.com/redhat-developer/odo/pkg/project"
 )
 
 type selectWorkloadStep int
@@ -19,14 +26,16 @@ const (
 // InteractiveBackend is a backend that will ask information interactively using the `asker` package
 type InteractiveBackend struct {
 	askerClient      asker.Asker
+	projectClient    project.Client
 	kubernetesClient kclient.ClientInterface
 }
 
 var _ AddBindingBackend = (*InteractiveBackend)(nil)
 
-func NewInteractiveBackend(askerClient asker.Asker, kubernetesClient kclient.ClientInterface) *InteractiveBackend {
+func NewInteractiveBackend(askerClient asker.Asker, projectClient project.Client, kubernetesClient kclient.ClientInterface) *InteractiveBackend {
 	return &InteractiveBackend{
 		askerClient:      askerClient,
+		projectClient:    projectClient,
 		kubernetesClient: kubernetesClient,
 	}
 }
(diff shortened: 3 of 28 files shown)


dharmit:
```sh
$ odo add binding
? Do you want to list services from: all accessible namespaces
? Select the namespace containing the service instances: newproject
 ✗  Service Binding Operator is not installed on the cluster, please ensure it is installed before proceeding. See installation instructions: https://odo.dev/docs/overview/cluster-setup/
```
IMO, the check for SBO's availability should happen before we ask the user to answer these questions.

pkg/binding/backend/interactive.go
```diff
@@ -98,6 +102,38 @@ loop:
 	return selectedName, selectedGVK, nil
 }
 
+// SelectNamespaceForServiceInstances returns the namespace which services instances should be listed from.
+// An empty return value means that service instances will be listed from the current namespace.
+// Otherwise, they will be listed from the specified namespace.
```
dharmit, line 107: The comment seems copy-pasted from `interface.go` file. IMO, having it at one place is OK. In an earlier discussion (on some PR review) with Philippe, he suggested putting it in `interface.go` file and only adding implementation specific comment at the places where it is implemented. It made sense to me.

pkg/binding/backend/interactive.go
```diff
@@ -98,6 +102,38 @@ loop:
...
+		return "", err
+	}
+
+	switch option {
+	case asker.CurrentNamespace:
+		return "", nil
+	case asker.AllAccessibleNamespaces:
+		klog.V(2).Infof("Listing all projects/namespaces...")
+		var nsList []string
+		nsList, err = nsListProvider()
```
dharmit, line 120: IMO, the logic to list namespaces should be in the business layer because it's not something specific to the CLI layer. It's the stuff we want to get from the k8s cluster. Such handler function coming from CLI layer should be used when we want to execute some CLI specific logic, which is not the case here. WDYT?

pkg/binding/backend/interactive.go
```diff
@@ -98,6 +102,38 @@ loop:
...
+		if err != nil {
+			return "", err
+		}
+		sort.Strings(nsList)
+		klog.V(4).Infof("all accessible namespaces: %v", nsList)
+		if len(nsList) == 0 {
```
dharmit, line 126: Out of curiosity, when is this condition going to be satisfied? The user would always have the current working namespace in the `nsList`, no?

pkg/binding/backend/flags.go
```diff
@@ -59,6 +60,13 @@ func (o *FlagsBackend) SelectWorkloadInstance(workloadName string) (string, sche
 	return "", schema.GroupVersionKind{}, fmt.Errorf("group/kind %q not found on the cluster", selectedGroup+"/"+selectedKind)
 }
 
+// SelectNamespaceForServiceInstances returns the namespace which services instances should be listed from.
+// An empty return value means that service instances will be listed from the current namespace.
+// Otherwise, they will be listed from the specified namespace.
```
dharmit, line 65: Same request as for the file `interactive.go` to keep comment at one place and add only implementation specific comment where an interface method is getting implemented.

dharmit (changes requested):

pkg/binding/backend/interactive.go
valaparthvi: If the user does not have the permission to list the namespace, this condition will become true.

pkg/binding/backend/interactive.go
valaparthvi: I agree with Dharmit here. Also, if we only need to list namespaces for the interactive mode, why not just do it in `InteractiveBackend.SelectNamespaceForServiceInstances`?

pkg/binding/backend/interactive.go
dharmit: Thanks, @valaparthvi. That didn't occur to me.

pkg/odo/cli/add/binding/binding.go
```diff
@@ -201,12 +232,13 @@ func NewCmdBinding(name, fullName string) *cobra.Command {
 	bindingCmd.Flags().String(backend.FLAG_NAME, "", "Name of the Binding to create")
 	bindingCmd.Flags().String(backend.FLAG_WORKLOAD, "", "Name of the workload to bind, only when no devfile is present in current directory")
 	bindingCmd.Flags().String(backend.FLAG_SERVICE, "", "Name of the service to bind")
+	bindingCmd.Flags().String(backend.FLAG_SERVICE_NAMESPACE, "", "Namespace of the service to bind to bind to. Default is the component namespace.")
```
valaparthvi, line 235:
```suggestion
	bindingCmd.Flags().String(backend.FLAG_SERVICE_NAMESPACE, "", "Namespace of the service to bind to. Default is the component namespace.")
```

pkg/binding/asker/interface.go
```diff
@@ -8,7 +8,23 @@ const (
...
 type Asker interface {
+	// SelectNamespaceListOption asks the user to select how service instances should be listed,
+	// i.e. from the current namespace or from all accessible namespaces.
+	// If user selects to list from all accessible namespaces, callers should fetch the list of namespaces and prompt them to either pick or enter one.
+	// See SelectNamespaceForServiceInstancesFrom and AskNamespaceForServiceInstances methods.
+	SelectNamespaceListOption() (ServiceInstancesNamespaceListOption, error)
+	// AskNamespaceForServiceInstances asks the user to enter a namespace
+	AskNamespaceForServiceInstances() (string, error)
+	// SelectNamespaceForServiceInstancesFrom takes a list of namespaces and asks the user to pick one of them
+	SelectNamespaceForServiceInstancesFrom(options []string) (string, error)
```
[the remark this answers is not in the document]
rm3l: I hoped the longer names would make them more understandable, but I understand your point :)
I'll shorten the names.

pkg/binding/backend/interactive.go
rm3l: Makes sense. I updated the comments in the interface, but also in each different implementation.

pkg/binding/backend/interactive.go
rm3l: I think I did it this way mainly because of the warning message I wanted to display in the CLI if odo could not list all the namespaces/projects (e.g., due to RBAC issues in Kubernetes clusters). But you're right - I'll handle this differently.

pkg/binding/backend/interactive.go
rm3l: Indeed, this can happen because of RBAC issues. In this case, the user will be prompted to manually enter a namespace (or press Enter to select the current one).
I wanted to test this scenario after reading a difference in behavior between listing all projects in OpenShift and listing all namespaces in Kubernetes. In OpenShift, it returns the list of all projects the user really has access to. But in Kubernetes, by design, it returns an error if the user does not have permission to list namespaces.

See message of commit [88e6902a0819256a9508c4b7f30beba6b04dacef](https://github.com/redhat-developer/odo/pull/5936/commits/88e6902a0819256a9508c4b7f30beba6b04dacef) for more details.

pkg/binding/backend/interactive.go
valaparthvi: > I think I did it this way mainly because of the warning message I wanted to display in the CLI if odo could not list all the namespaces/projects (e.g., due to RBAC issues in Kubernetes clusters). But you're right - I'll handle this differently.

Ah that makes sense, but we have been using log/status on business layer for quite a few packages, adding one more to the list won't make much difference :sweat_smile:

pkg/binding/binding.go
```diff
@@ -64,15 +66,11 @@ func (o *BindingClient) GetFlags(flags map[string]string) map[string]string {
 }
 
 func (o *BindingClient) GetServiceInstances(namespace string) (map[string]unstructured.Unstructured, error) {
-	isServiceBindingInstalled, err := o.kubernetesClient.IsServiceBindingSupported()
+	err := o.checkServiceBindingOperatorInstalled()
```
dharmit, line 69: Putting this over here means this method is also doing some validation that's not about "getting service instances". How about checking this in the `Validate` method at CLI layer?

rm3l: I had also added the check in the `ValidateAddBinding` method (which is indeed called by the `Validate` method in the CLI layer). See `pkg/binding/add.go` in the last commit ([64bd15fc733d04b1592e0aab5a66250d864f681c](https://github.com/redhat-developer/odo/pull/5936/commits/64bd15fc733d04b1592e0aab5a66250d864f681c))

I chose to leave the check here in `GetServiceInstances` (I think this was already there in the original implementation of `odo add binding`), because I think it still makes sense as it relies on the ServiceBinding operator presence to get this list. I know we might not be there yet, but it also made sense to me especially if this business package could potentially be imported and used by external tools.

Does that make sense?

dharmit: > because I think it still makes sense as it relies on the ServiceBinding operator presence to get this list. 

Ah, I forgot that we are trying to get the bindable service instances. OK, now it makes sense to put this in here. 👍🏾 

> I know we might not be there yet, but it also made sense to me especially if this business package could potentially be imported and used by external tools.

That's the (indirect) goal. Some day, for sure. :)

/lgtm

dharmit:
@rm3l it's a bit too many failing tests to override. Can you PTAL?

rm3l:
> @rm3l it's a bit too many failing tests to override. Can you PTAL?

Sure - I've fixed the failing Validate jobs, related to some changes I pushed.
I've also rebased onto the main branch to benefit from the recent fixes for tests. Can you please lgtm it again when it looks ok to you?

rm3l:
```
secret/pull-secret created
++ oc adm policy add-role-to-user edit developer
Warning: User 'developer' not found
clusterrole.rbac.authorization.k8s.io/edit added: "developer"
++ oc apply -n openshift -f https://raw.githubusercontent.com/openshift/library/master/arch/x86_64/community/wildfly/imagestreams/wildfly-centos7.json
error: unable to read URL "https://raw.githubusercontent.com/openshift/library/master/arch/x86_64/community/wildfly/imagestreams/wildfly-centos7.json", server reported 404 Not Found, status code=404
```

/test v4.10-integration-e2e

PR #5942, opened 2022-07-12
Grab pod logs concurrently

dharmit:

**What type of PR is this:**
/kind cleanup

Uses go routines to concurrently grab the logs of the Pods matching an odo component

Fixes part of #5872

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

* Create a component using [this devfile](https://github.com/redhat-developer/odo/blob/main/tests/examples/source/devfiles/nodejs/devfile-deploy-functional-pods.yaml)
* Place `start := time.Now()` at the beginning of `Run` method in `pkg/odo/cli/logs/logs.go` and print the time since `start` using `time.Since(start)` right before the call to `printLogs` function.
* Do the same for code from `main` branch.
* Compare the timings.


pkg/logs/logs.go
```diff
@@ -29,103 +29,114 @@ func NewLogsClient(kubernetesClient kclient.ClientInterface) *LogsClient {
...
+) error {
 	var selector string
-	logs := []map[string]io.ReadCloser{}
-	unknownMode := true
+	podChan := make(chan corev1.Pod) // grab the logs of the pod put on this channel
+	done := make(chan struct{})      // because populating doneChan directly would cause odo logs to exit prematurely.
+
+	go func() {
+		// this go routine gets the logs of the pods put on the podChan
+		for {
```
feloy, line 47: When is this infinite loop finishing? I cannot see any return inside

pkg/odo/cli/logs/logs.go
```diff
@@ -123,34 +122,41 @@ func (o *LogsOptions) Run(ctx context.Context) error {
...
-		// 2. user specified --deploy flag, but the component's running in Dev mode
-		// 3. user passes no flag, but component is running in neither Dev nor Deploy mode
-		fmt.Fprintf(o.out, "no containers running in the specified mode for the component %q\n", o.componentName)
-		return nil
-	}
+
+	errChan := make(chan error) // errors are put on this channel
+	done := make(chan struct{}) // a struct{}{} on done channel indicates that logs for all pods are fetched
+
+	go func() {
```
feloy, line 129: The content of GetLogsForMode is already executed in g a go routine. Do you really need another one here?

pkg/odo/cli/logs/logs.go
```diff
@@ -36,9 +34,10 @@ type LogsOptions struct {
 	clientset *clientset.Clientset
 
 	// variables
-	componentName string
-	contextDir    string
-	out           io.Writer
+	componentName    string
+	contextDir       string
+	out              io.Writer
+	containerLogChan chan map[string]interface{}
```
feloy, line 40: It seems this variable can be defined as a local varibale in the Run method

feloy (changes requested):

pkg/logs/logs.go
dharmit: When the parent function populates `done` channel. It's implicit in nature rather than being explicit.

feloy:
It seems the architecture could be simplified. 

If you look to other watcher methods (fsnotify, kubernetes), the business level function should return to the CLI a channel, and the CLI should loop over this channel to get the information and display it, until the channel is closed.

pkg/odo/cli/logs/logs.go
dharmit: Yeah, can do. I don't have a preference for either.

feloy: Varibales should always have the smallest possible scope. Having it in the Receiver makes it accessible from all the methods of the receiver, when it is only needed in the Run method.

pkg/logs/logs.go
dharmit: This go routine has been removed. We now return a struct of channels from the business layer akin to what you recommended in https://github.com/redhat-developer/odo/pull/5942#issuecomment-1183394909.

pkg/odo/cli/logs/logs.go
dharmit: This go routine has been removed. We now return a struct of channels from the business layer akin to what you recommended in https://github.com/redhat-developer/odo/pull/5942#issuecomment-1183394909.

pkg/odo/cli/logs/logs.go
dharmit: Variable removed in favour of a struct of channels being returned by business layer.

pkg/logs/logs.go
```diff
@@ -29,103 +35,130 @@ func NewLogsClient(kubernetesClient kclient.ClientInterface) *LogsClient {
...
+			select {
+			case pod := <-podChan:
+				for _, container := range pod.Spec.Containers {
+					containerLogs, err := o.kubernetesClient.GetPodLogs(pod.Name, container.Name, follow)
+					if err != nil {
+						fmt.Fprintf(out, "failed to get logs for container %s; error: %v", container.Name, err)
```
rm3l, line 76: Just curious as to why we don't write this error to `events.Err`..

pkg/logs/logs.go
```diff
@@ -29,103 +35,130 @@ func NewLogsClient(kubernetesClient kclient.ClientInterface) *LogsClient {
...
+					}
+					events.Logs <- map[string]interface{}{"name": container.Name, "logs": containerLogs}
```
rm3l, line 78: This will get executed even if `err != nil`, potentially causing `containerLogs` to added as `nil` in the map; and later in the CLI, the value would be type-asserted as an `io.ReadCloser`, which might cause this to panic, I think, no?

rm3l (changes requested):
Added a few comments.

Also, while testing this, I noticed a strange behavior (well, I had observed a different behavior prior to checking out this branch). In my case, when I had issues with `fsnotify` (e.g.: `too many open files`), `odo logs --follow` used to exit right away. But now, it no longer exits, but subsequent container logs are no longer displayed either.
To me, not exiting right away is a bit odd, because it makes users think that the error message is not an issue, so it might make them expect subsequent logs to be returned, which is not the case here.

```
❯ odo logs --follow
runtime: 
runtime: added 32 packages, and audited 33 packages in 2s
runtime: 
runtime: 3 packages are looking for funding
runtime:   run `npm fund` for details
runtime: 
runtime: found 0 vulnerabilities
runtime: npm notice 
runtime: npm notice New minor version of npm available! 8.3.1 -> 8.14.0
runtime: npm notice Changelog: <https://github.com/npm/cli/releases/tag/v8.14.0>
runtime: npm notice Run `npm install -g npm@8.14.0` to update!
runtime: npm notice 
runtime: 
runtime: > node-echo@0.0.1 start
runtime: > node server.js
runtime: 
runtime: Server running on 0.0.0.0:8080
runtime: failed to create fsnotify watcher: too many open files
main: Server running on 0.0.0.0:8080
main: failed to create fsnotify watcher: too many open files


```

pkg/logs/logs.go
dharmit: Mainly because I was dicey about stopping the flow if we failed to grab logs of one container among the possibly more than one container that a user could have running. But I have modified it to return an error to `events.Err`. 

We can think if there's any user feedback around what was cooking in my head. 😉

pkg/logs/logs.go
dharmit: Ah, nice catch! 👍🏾

pkg/logs/logs.go
```diff
@@ -19,6 +19,12 @@ type LogsClient struct {
 	kubernetesClient kclient.ClientInterface
 }
 
+type Events struct {
+	Logs chan map[string]interface{}
```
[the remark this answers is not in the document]
dharmit: Addressed.

pkg/odo/cli/logs/logs.go
```diff
@@ -123,34 +120,35 @@ func (o *LogsOptions) Run(ctx context.Context) error {
...
 	wg := sync.WaitGroup{}
+	errChan := make(chan error) // errors are put on this channel
 	var mu sync.Mutex
-	for _, entry := range containersLogs {
-		for container, logs := range entry {
-			uniqueName := getUniqueContainerName(container, uniqueContainerNames)
+	for {
+		select {
+		case containerLogs := <-events.Logs:
+			uniqueName := getUniqueContainerName(containerLogs["name"].(string), uniqueContainerNames)
```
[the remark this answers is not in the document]
dharmit: Makes sense. Addressed.

pkg/logs/logs.go
dharmit: Addressed.

pkg/logs/logs.go
dharmit: Addressed.

pkg/logs/logs.go
rm3l: > Mainly because I was dicey about stopping the flow if we failed to grab logs of one container among the possibly more than one container that a user could have running. But I have modified it to return an error to `events.Err`.
> 
> We can think if there's any user feedback around what was cooking in my head. wink

It made sense to me in the first place because of the other "too many open files" error I mentioned above, and then no new logs were displayed. But okay, let's see if there is any other user feedback on this.

dharmit:
> Also, while testing this, I noticed a strange behavior (well, I had observed a different behavior prior to checking out this branch). In my case, when I had issues with `fsnotify` (e.g.: `too many open files`), `odo logs --follow` used to exit right away. But now, it no longer exits, but subsequent container logs are no longer displayed either.
> To me, not exiting right away is a bit odd, because it makes users think that the error message is not an issue, so it might make them expect subsequent logs to be returned, which is not the case here.

I'm at loss as to what exactly is happening here. The `too many open files` seems to be an error caused because the kernel won't let you open any more files as you have many opened already. But I don't understand what behaviour you are seeing with `odo logs --follow` when you face this error. 😕

rm3l:
> > Also, while testing this, I noticed a strange behavior (well, I had observed a different behavior prior to checking out this branch). In my case, when I had issues with `fsnotify` (e.g.: `too many open files`), `odo logs --follow` used to exit right away. But now, it no longer exits, but subsequent container logs are no longer displayed either.
> > To me, not exiting right away is a bit odd, because it makes users think that the error message is not an issue, so it might make them expect subsequent logs to be returned, which is not the case here.
> 
> I'm at loss as to what exactly is happening here. The `too many open files` seems to be an error caused because the kernel won't let you open any more files as you have many opened already. But I don't understand what behaviour you are seeing with `odo logs --follow` when you face this error. confused

I wanted to see how this would behave with the default limit on open files (I had this error in the past and increased the limit).

On `main`, `odo logs --folllow` exits immediately whenever I face this error (perhaps it is the same behavior for other errors). But now it no longer exits, but what I found odd is that it looks like it is stuck, because it does not display new logs either.
Maybe it's me, but I thought exiting immediately would make more sense.

dharmit:
> I wanted to see how this would behave with the default limit on open files (I had this error in the past and increased the limit).
> 
> On `main`, `odo logs --follow` exits immediately whenever I face this error (perhaps it is the same behavior for other errors). But now it no longer exits, but what I found odd is that it looks like it is stuck, because it does not display new logs either. Maybe it's me, but I thought exiting immediately would make more sense.

I still don't understand this entirely. Can you share the steps to reproduce? Are you using `ulimit` to reduce the number of files that could be opened? Also, is it still the case after addressing https://github.com/redhat-developer/odo/pull/5942#discussion_r922305744?

rm3l (changes requested):

rm3l:
> > I wanted to see how this would behave with the default limit on open files (I had this error in the past and increased the limit).
> > On `main`, `odo logs --follow` exits immediately whenever I face this error (perhaps it is the same behavior for other errors). But now it no longer exits, but what I found odd is that it looks like it is stuck, because it does not display new logs either. Maybe it's me, but I thought exiting immediately would make more sense.
> 
> I still don't understand this entirely. Can you share the steps to reproduce? Are you using `ulimit` to reduce the number of files that could be opened? 
> 
Yes, I was using `ulimit` and I was testing against this fork of the Spring Boot guide: https://github.com/rm3l/gs-spring-boot/tree/main/complete
I don't have clear reproduction steps for now, as sometimes it works regardless of the hard/soft limit set in the system.

Ah actually, I noticed that I had the same issue when running `odo logs --follow` from a separate laptop, but targeting the same cluster. After digging a little bit, I noticed that the `failed to create fsnotify watcher: too many open files` message seems to actually be a line in the container logs..
But again, I still don't get why `odo logs --follow` used to exit in the past, and now it looks like it is blocked.

Even `kubectl logs --follow` exits right away past this error message:

```
☸ kind-local-k8s-cluster in ~/w/t/s/gs-spring-boot/complete on  main via ☕ v17.0.3 on ☁️   
❯ kubectl logs my-java-springboot-app-app-c69bfd669-4fn72 --follow

-- Truncated --
viewResolver
webEndpointDiscoverer
webEndpointPathMapper
webEndpointServletHandlerMapping
webExposeExcludePropertyEndpointFilter
webMvcMetricsFilter
webMvcTagsProvider
webServerFactoryCustomizerBeanPostProcessor
websocketServletWebServerCustomizer
welcomePageHandlerMapping
failed to create fsnotify watcher: too many open files%                                                                                                                                                                                                                          

☸ kind-local-k8s-cluster in ~/w/t/s/gs-spring-boot/complete on  main via ☕ v17.0.3 on ☁️   
❯ echo $?
0

☸ kind-local-k8s-cluster in ~/w/t/s/gs-spring-boot/complete on  main via ☕ v17.0.3 on ☁️   
❯ 
```
 
> Also, is it still the case after addressing [#5942 (comment)](https://github.com/redhat-developer/odo/pull/5942#discussion_r922305744)?
>
Unfortunately, it is still the case with your new changes.

I can try to show you tomorrow if that's ok for you.

pkg/logs/logs.go
```diff
@@ -29,103 +43,130 @@ func NewLogsClient(kubernetesClient kclient.ClientInterface) *LogsClient {
...
+	errChan := make(chan error)
+	doneChan := make(chan struct{}) // because populating doneChan directly would cause odo logs to exit prematurely.
+
+	go func() {
+		// this go routine gets the logs of the pods put on the podChan
+		for {
+			select {
+			case pod := <-podChan:
+				for _, container := range pod.Spec.Containers {
+					containerLogs, err := o.kubernetesClient.GetPodLogs(pod.Name, container.Name, follow)
```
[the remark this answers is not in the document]
dharmit: It's doable, but this piece of code is already inside a go routine. Plus, there's nothing else to be done till we get the logs. So I don't think it's going to add any value.

cdrage:
Darn good job. No changes needed, it works wonders!

/lgtm

PR #5946, opened 2022-07-14
Sbo nightly test

### LINKED ISSUE #5880: odo-v2-operatorhub-integration-nightly tests failed to notify breaking change
/kind bug
/area testing

odo-v2-operatorhub-integration-nightly were set up to notify any changes that can break odo functionality. 
Incompatibility of odo with SBO 1.1.0 should have been notified by the failure of tests, but the tests didn't fail on nightly tests.

## Accepted behavior
SBO nightly tests should fail in case of incompatibility.
odo-v2-operatorhub-integration-nightly tests should be fixed to prevent similar failure in the future.

anandrkskd:
**What type of PR is this:**

/kind tests

This PR will add scripts required for periodic tests that will run against master build of SBO.

Fixes #5880

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/configure-cluster/common/setup-operators.sh
+++ b/scripts/configure-cluster/common/setup-operators.sh
@@ -31,7 +31,43 @@ install_service_binding_operator() {
 EOF
 }
 
-if [ "$KUBERNETES" == "true" ]; then
+install_service_binding_operator_master() {
+  oc create -f - <<EOF
+  apiVersion: operators.coreos.com/v1alpha1
+  kind: CatalogSource
+  metadata:
+    name: service-binding-master
+    namespace: openshift-marketplace
+  spec:
+    displayName: Service Binding Operator build from master
+    image: quay.io/redhat-developer/servicebinding-operator:index
+    priority: 500
+    publisher: Red Hat
+    sourceType: grpc
+    updateStrategy:
+      registryPoll:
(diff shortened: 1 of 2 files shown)


scripts/configure-cluster/common/setup-operators.sh
```diff
@@ -31,7 +31,43 @@ install_service_binding_operator() {
...
+  metadata:
+    name: service-binding-operator
+    namespace: openshift-operators
+  spec:
+    channel: candidate
+    installPlanApproval: Automatic
+    name: service-binding-operator
+    source: service-binding-master
+    sourceNamespace: openshift-marketplace
+    startingCSV: service-binding-operator.v1.1.1
```
dharmit, line 63: IMO, we should skip setting this. My understanding is that if v1.1.2 is released, this directive here would cause OLM to install v1.1.1, which is not what we want.

dharmit (changes requested):

scripts/configure-cluster/common/setup-operators.sh
anandrkskd: understood I will update it ASAP

dharmit (approved):

PR #5958, opened 2022-07-20
Remove green bold from port forwarding

cdrage:
**What type of PR is this:**

/kind bug

Fixes the problem of port information not showing on light terminals.
This changes the ports being shown as bolded rather than green (and
bolded).

Confirmed to work on both light and dark terminal colours

Fixes https://github.com/redhat-developer/odo/issues/5843

**PR acceptance criteria:**

- [X] Unit test
- [X] Integration test
- [X] Documentation

N/A

### DIFF
--- a/pkg/portForward/writer.go
+++ b/pkg/portForward/writer.go
@@ -8,9 +8,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/fatih/color"
-
 	"github.com/redhat-developer/odo/pkg/api"
+	"github.com/redhat-developer/odo/pkg/log"
 
 	"k8s.io/klog"
 )
@@ -37,11 +36,6 @@ func NewPortWriter(buffer io.Writer, len int, mapping map[string][]int) *PortWri
 
 func (o *PortWriter) Write(buf []byte) (n int, err error) {
 
-	// Set the colours to green (to indicate that the port is OPEN)
-	// as well as bold. So it stands our that the application is currently
-	// being port forwarded.
-	color.Set(color.FgGreen, color.Bold)
-	defer color.Unset() // Use it in your function
 	s := string(buf)
 	if strings.HasPrefix(s, "Forwarding from 127.0.0.1") {
 
@@ -52,7 +46,7 @@ func (o *PortWriter) Write(buf []byte) (n int, err error) {
 			klog.V(4).Infof("unable to get forwarded port: %v", err)
 		}
 
-		fmt.Fprintf(o.buffer, " - %s", s)
+		fmt.Fprintf(o.buffer, " -  %s", log.Sbold(s))
 		o.len--
 		if o.len == 0 {
 			o.end <- true


valaparthvi:
Wasn't the problem with white font on white background? Why did you change the green font? I liked it when it was highlighted.

cdrage:
> Wasn't the problem with white font on white background? Why did you change the green font? I liked it when it was highlighted, and set apart.

Whoops! I forgot to clarify, the reason is that the `color` library tries to set it as green, but in some terminals it doesn't work correctly with colours. Instead, it defaults to white. 

The green unfortunately causes too many issues, it it's better to use "bold" which will try and bold the current colour, rather than turn it white.

valaparthvi (approved):

rm3l:
> > Wasn't the problem with white font on white background? Why did you change the green font? I liked it when it was highlighted, and set apart.
> 
> Whoops! I forgot to clarify, the reason is that the `color` library tries to set it as green, but in some terminals it doesn't work correctly with colours. Instead, it defaults to white.
> 
> The green unfortunately causes too many issues, it it's better to use "bold" which will try and bold the current colour, rather than turn it white.

I found other occurrences in the code where we are using green as well:

```sh
❯ git grep "color.FgGreen" -- ':!vendor'

pkg/init/backend/interactive.go:        color.New(color.Bold, color.FgGreen).Println("Current component configuration:")
pkg/log/status.go:var colors = []color.Attribute{color.FgRed, color.FgGreen, color.FgYellow, color.FgBlue, color.FgMagenta, color.FgCyan, color.FgWhite}
pkg/log/status.go:                      green := color.New(color.FgGreen).SprintFunc()
pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
```

For example, it is used after personalizing the Devfile configuration when running `odo dev` or `odo deploy`.

Does it mean we should handle those as well?

/hold
(But feel free to remove the hold if needed)

rm3l (commented):
> > Wasn't the problem with white font on white background? Why did you change the green font? I liked it when it was highlighted, and set apart.
> 
> Whoops! I forgot to clarify, the reason is that the `color` library tries to set it as green, but in some terminals it doesn't work correctly with colours. Instead, it defaults to white.
> 
> The green unfortunately causes too many issues, it it's better to use "bold" which will try and bold the current colour, rather than turn it white.

I found other occurrences in the code where we are using green as well:

```sh
❯ git grep "color.FgGreen" -- ':!vendor'

pkg/init/backend/interactive.go:        color.New(color.Bold, color.FgGreen).Println("Current component configuration:")
pkg/log/status.go:var colors = []color.Attribute{color.FgRed, color.FgGreen, color.FgYellow, color.FgBlue, color.FgMagenta, color.FgCyan, color.FgWhite}
pkg/log/status.go:                      green := color.New(color.FgGreen).SprintFunc()
pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
```

For example, it is used after personalizing the Devfile configuration when running `odo dev` or `odo deploy`.

Does it mean we should handle those as well? Or at least the case in `pkg/init/backend/interactive.go`, which uses green and bold?

/hold
(But feel free to remove the hold if needed)

cdrage:
> > > Wasn't the problem with white font on white background? Why did you change the green font? I liked it when it was highlighted, and set apart.
> > 
> > 
> > Whoops! I forgot to clarify, the reason is that the `color` library tries to set it as green, but in some terminals it doesn't work correctly with colours. Instead, it defaults to white.
> > The green unfortunately causes too many issues, it it's better to use "bold" which will try and bold the current colour, rather than turn it white.
> 
> I found other occurrences in the code where we are using green as well:
> 
> ```shell
> ❯ git grep "FgGreen" -- ':!vendor'
> 
> pkg/init/backend/interactive.go:        color.New(color.Bold, color.FgGreen).Println("Current component configuration:")
> pkg/log/status.go:var colors = []color.Attribute{color.FgRed, color.FgGreen, color.FgYellow, color.FgBlue, color.FgMagenta, color.FgCyan, color.FgWhite}
> pkg/log/status.go:                      green := color.New(color.FgGreen).SprintFunc()
> pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
> pkg/log/status.go:              green := color.New(color.FgGreen).SprintFunc()
> ```
> 
> For example, it is used after personalizing the Devfile configuration when running `odo dev` or `odo deploy`.
> 
> Does it mean we should handle those as well? Or at least the case in `pkg/init/backend/interactive.go`, which uses green and bold?
> 
> /hold (But feel free to remove the hold if needed)

Ah thanks! You bring up a good point as to what's happening. 

I totally forgot we check the terminal in status.go for color compatibility, hence why green is working in log/status.go, but not in the other sections.

I'm going to close this PR and propose something different.

PR #5963, opened 2022-07-21
Refactor color througout interactive mode and dev

cdrage:
**What type of PR is this:**

/kind cleanup

This PR fixes the following:
- Updates the interactive mode so it's better looking when outputting
  configuration so it's similar to the rest of the output:

```sh
? Is this correct? Yes
 ✓  Downloading devfile "nodejs" from registry "DefaultDevfileRegistry" [1s]

↪ Container Configuration "runtime":
  OPEN PORTS:
    - 3000
  ENVIRONMENT VARIABLES:

? Select container for which you want to change configuration? runtime
? What configuration do you want change? Delete port "3000"
? What configuration do you want change? NOTHING - configuration is correct

↪ Container Configuration "runtime":
  OPEN PORTS:
  ENVIRONMENT VARIABLES:

? Select container for which you want to change configuration?  [Use arrows to move, type to filter]
  runtime
> NONE - configuration is correct
```

- Removes issue of fgGreen not appearing correctly because of direct use
  of `color` instead of log/status.go. This is due to color not
  detecting if the terminal supports the color, which is done in
  `log/status.go` instead.

Fixes https://github.com/redhat-developer/odo/issues/5843

### DIFF
--- a/pkg/init/backend/interactive.go
+++ b/pkg/init/backend/interactive.go
@@ -8,7 +8,6 @@ import (
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/pkg/devfile/parser"
 	parsercommon "github.com/devfile/library/pkg/devfile/parser/data/v2/common"
-	"github.com/fatih/color"
 
 	"github.com/redhat-developer/odo/pkg/api"
 	"github.com/redhat-developer/odo/pkg/init/asker"
@@ -210,7 +209,6 @@ func (o *InteractiveBackend) PersonalizeDevfileConfig(devfileobj parser.DevfileO
 }
 
 func PrintConfiguration(config asker.DevfileConfiguration) {
-	color.New(color.Bold, color.FgGreen).Println("Current component configuration:")
 
 	var keys []string
 	for key := range config {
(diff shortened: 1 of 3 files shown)


pkg/init/backend/interactive.go
```diff
@@ -220,18 +218,24 @@ func PrintConfiguration(config asker.DevfileConfiguration) {
...
+
 		for key, value := range container.Envs {
-			color.New(color.Bold, color.FgWhite).Printf("   - %s = %s\n", key, value)
+			fmt.Printf("    - %s = %s\n", key, value)
 		}
+
 	}
+
+	// Make sure we add a newline at the end
+	fmt.Println()
```
dharmit, line 238: Please don't use `fmt.Printf` or `fmt.Println` in the business layer. It should ideally be used only in CLI layer. Use `fmt.Fprintf` with stdout as `io.Writer`.

dharmit (changes requested):

cdrage:
Thanks! Updated @dharmit

rm3l (commented):
/approve

cdrage:
Re-added

/lgtm

Because I forgot to sign the commit

rm3l:

Tests passing in IBM Cloud.

PR #5964, opened 2022-07-21
Add spinner ending for running command

cdrage:
**What type of PR is this:**

/kind bug

When running `odo dev`, it will never show that the command has
completed.

```sh
 ✓  Building your application in container on cluster (command: install) [3s]
 -  Forwarding from 127.0.0.1:40001 -> 3000
```

`spinner.End` hadn't been added so it never shows at it ending.

This PR fixes that:

```sh
 ✓  Building your application in container on cluster (command: install) [3s]
 ✓  Executing the application (command: run) [5s]
 ```

N/A

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/commandhandler.go
+++ b/pkg/devfile/adapters/kubernetes/component/commandhandler.go
@@ -62,6 +62,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 	// Spinner created but not started yet.
 	// It will be displayed when the statusHandlerFunc function is called with the "Starting" state.
 	spinner := log.NewStatus(log.GetStdout())
+	defer spinner.End(false)
 
 	// if we need to restart, issue the remote process handler command to stop all running commands first.
 	// We do not need to restart Hot reload capable commands.
@@ -122,6 +123,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 		return err
 	}
 
+	spinner.End(true)
 	return a.checkRemoteCommandStatus(devfileCmd, a.podName,
 		fmt.Sprintf("Devfile command %q exited with an error status in %.0f second(s)", devfileCmd.Id, totalWaitTime))
 }

### CODE REVIEW

**Review by rm3l — CHANGES_REQUESTED** [review]
> When running `odo dev`, it will never show that the command has completed.
> 
> ```shell
>  ✓  Building your application in container on cluster (command: install) [3s]
>  •  Executing the application (command: run)  ...
>  -  Forwarding from 127.0.0.1:40001 -> 3000
> ```
> 
> `spinner.End` hadn't been added so it never shows at it ending.
> 
> This PR fixes that:
> 
> ```shell
>  ✓  Building your application in container on cluster (command: install) [3s]
>  •  Executing the application (command: run)  ...
>  ✓  Executing the application (command: run) [5s]
> ```
> 
@cdrage With the removal of Supervisord, `odo dev` now executes the `run` or `debug` command in a separate goroutine. So to me, we cannot know for sure when the command is done (until the corresponding goroutine reports so). This is why we used a spinner that does not end.
So adding a spinner that ends here can be a bit misleading to the users, no?

Let's say the command exits a few seconds/minutes later (simulated by the command below in my Devfile):

```yaml
- exec:
    commandLine: 'sleep 10 && echo Exiting... && exit 1'
    component: runtime
    group:
      isDefault: true
      kind: run
    workingDir: $PROJECT_SOURCE
  id: run
```

With the changes in this PR, the spinner has already ended, so it will never be updated again, and the user would never know that the command failed:
```
❯ odo dev      
  __
 /  \__     Developing using the my-nodesjs-with-k8s-uri-with-var Devfile
 \__/  \    Namespace: default
 /  \__/    odo version: v3.0.0-beta2
 \__/

 ✓  Waiting for Kubernetes resources [6s]
 ✓  Syncing files into the container [151ms]
 ✓  Building your application in container on cluster (command: install) [4s]
 ✓  Executing the application (command: run) [5s]
 - Forwarding from 127.0.0.1:40003 -> 8080

Watching for changes in the current directory /home/asoro/work/tmp/5451-deploy-k8s-uri-var-substitution
Press Ctrl+c to exit `odo dev` and delete resources from the cluster
```

Before, users had such information at the end:
```
❯ odo dev
  __
 /  \__     Developing using the my-nodesjs-with-k8s-uri-with-var Devfile
 \__/  \    Namespace: default
 /  \__/    odo version: v3.0.0-beta2
 \__/

 ✓  Waiting for Kubernetes resources [8s]
 ✓  Syncing files into the container [160ms]
 ✓  Building your application in container on cluster (command: install) [5s]
 - Forwarding from 127.0.0.1:40003 -> 8080

Watching for changes in the current directory /home/asoro/work/tmp/5451-deploy-k8s-uri-var-substitution
Press Ctrl+c to exit `odo dev` and delete resources from the cluster

 ✗  Executing the application (command: run) [10s]

```

cdrage:
> •  Executing the application (command: run)  ...

Ah, I get your point.

The problem is.. it looks really out of place with the spinner.

We should remove the spinner entirely and just have a prompt saying that the command would be running in the background? I don't like how it's out of place with all the other stuff in the workflow when running `odo dev`. 

We should just have it as:

```sh
 ✓  Waiting for Kubernetes resources [8s]
 ✓  Syncing files into the container [160ms]
 ✓  Building your application in container on cluster (command: install) [5s]
 ✓  Executing the application (command: run) in the background
 - Forwarding from 127.0.0.1:40003 -> 8080
 ```

Instead of the spinner.

What do you think?

rm3l:
> > •  Executing the application (command: run)  ...
> 
> Ah, I get your point.
> 
> The problem is.. it looks really out of place with the spinner.
> 
> We should remove the spinner entirely and just have a prompt saying that the command would be running in the background? I don't like how it's out of place with all the other stuff in the workflow when running `odo dev`.
> 
> We should just have it as:
> 
> ```shell
>  ↪ Deploying to the cluster in developer mode
>  ✓  Waiting for Kubernetes resources [8s]
>  ✓  Syncing files into the container [160ms]
>  ✓  Building your application in container on cluster (command: install) [5s]
>  ✓  Executing the application (command: run) in the background
>  - Forwarding from 127.0.0.1:40003 -> 8080
> ```
> 
> Instead of the spinner.
> 
> What do you think?

That would be better indeed, as long as we are able to report the command status later (as shown in my previous example), if the command exits at any later point in time (either successfully or not).

cdrage:
Closing this in favour of https://github.com/redhat-developer/odo/pull/5972

PR #5972, opened 2022-07-25
Allow specifying message to end the spinner with

### LINKED ISSUE #5881: `odo dev` prints `✗  Executing the application` after cleaning up the resources
/kind bug
/area dev

**Operating System:** all

* Built odo from the `main` branch and ran `odo dev`
* Hit `Ctrl+C` to stop the `odo dev` session

## Actual behavior
odo prints ` ✗  Executing the application (command: run) [1m]` after cleaning up the resources. But it doesn't error out - `echo $?` output is 0.
```sh
...
Cancelling deployment.
This is non-preemptive operation, it will wait for other tasks to finish first

 ✗  Executing the application (command: run) [1m]
```

## Expected behavior
odo shouldn't print such a message

dharmit:

**What type of PR is this:**
/kind bug

It fixes the UX that would lead a user to think that there was some error in executing the `run` command

Fixes #5881

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

* Run `odo dev`
* Change a code file in the directory; odo should not exhibit the erroneous UX
* Stop `odo dev` with `Ctrl+C`; odo should not exhibit the erroneous UX

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/commandhandler.go
+++ b/pkg/devfile/adapters/kubernetes/component/commandhandler.go
@@ -55,7 +55,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 				// Creating with no spin because the command could be long-running, and we cannot determine when it will end.
 				s.Start(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id), true)
 			case remotecmd.Stopped, remotecmd.Errored:
-				s.EndWithStatus(fmt.Sprintf("Finished executing the long running application (command: %s)", devfileCmd.Id), status == remotecmd.Stopped)
+				s.EndWithStatus(fmt.Sprintf("Finished executing the application (command: %s)", devfileCmd.Id), status == remotecmd.Stopped)
 				if err != nil {
 					klog.V(2).Infof("error while running background command: %v", err)
 				}
--- a/pkg/devfile/adapters/kubernetes/component/commandhandler.go
+++ b/pkg/devfile/adapters/kubernetes/component/commandhandler.go
@@ -55,7 +55,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 				// Creating with no spin because the command could be long-running, and we cannot determine when it will end.
 				s.Start(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id), true)
 			case remotecmd.Stopped, remotecmd.Errored:
-				s.End(status == remotecmd.Stopped)
+				s.EndWithStatus(fmt.Sprintf("Finished executing the long running application (command: %s)", devfileCmd.Id), status == remotecmd.Stopped)
 				if err != nil {
 					klog.V(2).Infof("error while running background command: %v", err)
 				}
--- a/pkg/log/status.go
+++ b/pkg/log/status.go
@@ -221,6 +221,15 @@ func (s *Status) End(success bool) {
 	s.status = ""
 }
 
+// EndWithStatus is similar to End, but lets the user specify a custom message/status while ending
+func (s *Status) EndWithStatus(status string, success bool) {
+	if status == "" {
+		return
+	}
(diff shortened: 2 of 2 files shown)


pkg/devfile/adapters/kubernetes/component/commandhandler.go
```diff
@@ -49,9 +49,8 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 			switch status {
 			case remotecmd.Starting:
 				// Creating with no spin because the command could be long-running, and we cannot determine when it will end.
-				s.Start(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id), true)
+				log.SpinnerNoSpin(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id))
 			case remotecmd.Stopped, remotecmd.Errored:
-				s.End(status == remotecmd.Stopped)
```
rm3l, line 54: If there is an error a few seconds/minutes after the command has started, users would never be aware of this, I think. See my comment in https://github.com/redhat-developer/odo/pull/5964#pullrequestreview-1047724907

rm3l (changes requested):
Added a comment. I think #5964 is also trying to change how the "Executing the application..." message is displayed to the users.

But regarding the original issue addressed here, would it make sense to capture the interrupt signal in `kubeexec.go` (where the command goroutine is spawned), and then, in the goroutine itself, detect that signal and call the handler in a way that allows it to handle this case differently (maybe with a status different from Stopped/Errored)?
This way, I think, it would solve the issue of notifying users if the command errors out anytime after it has been started, and it would not display any message if the dev session stops normally.
What do you think?

dharmit:
> But regarding the original issue addressed here, would it make sense to capture the interrupt signal in `kubeexec.go` (where the command goroutine is spawned), and then, in the goroutine itself, detect that signal and call the handler in a way that allows it to handle this case differently (maybe with a status different from Stopped/Errored)? 
> This way, I think, it would solve the issue of notifying users if the command errors out anytime after it has been started, and it would not display any message if the dev session stops normally. What do you think?

@rm3l I had a chat about this with @feloy. His point was that the status indicated by odo is technically correct. So we concluded to change the wording so that it's less confusing for the user. I have pushed a change accordingly. PTAL and let me know what you think about it.

pkg/devfile/adapters/kubernetes/component/commandhandler.go
```diff
@@ -55,7 +55,7 @@ func (a *adapterHandler) Execute(devfileCmd devfilev1.Command) error {
 				// Creating with no spin because the command could be long-running, and we cannot determine when it will end.
 				s.Start(fmt.Sprintf("Executing the application (command: %s)", devfileCmd.Id), true)
 			case remotecmd.Stopped, remotecmd.Errored:
-				s.End(status == remotecmd.Stopped)
+				s.EndWithStatus(fmt.Sprintf("Finished executing the long running application (command: %s)", devfileCmd.Id), status == remotecmd.Stopped)
```
rm3l, line 58: Okay, after a quick chat with Philippe as well, I'm ok with the changes here. Let's hope it helps users understand what is being done here.

One comment though on the message. The command might not necessarily be a long-running one (case of short commands part of a composite run/debug command). So I would suggest changing the message to something like:

```suggestion
				s.EndWithStatus(fmt.Sprintf("Finished executing the application (command: %s)", devfileCmd.Id), status == remotecmd.Stopped)
```

rm3l: Another point discussed was, if possible, having the "Cleaning resources, please wait" message displayed last, after all the errors have been captured and handled. This would also help avoid waiting a few minutes until resources are deleted before displaying errors.
But this can be addressed in a separate issue.

rm3l (changes requested):

feloy (approved):

PR #5973, opened 2022-07-25
Fetch pods first; use channel instead of slice

### LINKED ISSUE #5872: `odo logs` too slow for non-local clusters

dharmit:
/kind bug

We first fetch the pods matching the selector so that we can quickly start showing the logs to the user. For a component created using [this devfile in test suite](https://github.com/redhat-developer/odo/blob/main/tests/examples/source/devfiles/nodejs/devfile-deploy-functional-pods.yaml) and running in both Dev and Deploy modes, performance improvement is as below:

```sh
# odo built using the main branch
$ odo-main logs | grep "===="
============================= 2.322569471s =============================
============================= 2.614284155s =============================
============================= 6.419786304s =============================
============================= 6.727155955s =============================

# odo built using this PR's branch
$ odo-pods-first logs | grep "====" 
============================= 1.475623872s =============================
============================= 1.781928233s =============================
============================= 6.903285785s =============================
============================= 7.182508784s =============================

```

I added logic to grab the deployments from the cluster and then match pods to those to see if there could be further improvement, but it's worsening:
```sh
$ odo-get-deployments logs | grep "===="
============================= 1.575954862s =============================
============================= 1.873502308s =============================
============================= 8.719334078s =============================
============================= 9.051154327s =============================
```

Fixes #5872

Marking the issue fixed since there aren't any other ideas that we have to improve the performance at the moment.

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

* Create a component using the devfile mentioned above
* Use below diff to benchmark the performance:
  ```diff
  $ git diff --unified=0 
  diff --git a/pkg/odo/cli/logs/logs.go b/pkg/odo/cli/logs/logs.go
  index 6779c5173..04741027a 100644
  --- a/pkg/odo/cli/logs/logs.go
  +++ b/pkg/odo/cli/logs/logs.go
  @@ -15,0 +16 @@ import (
  +       "time"
  @@ -40,0 +42 @@ type LogsOptions struct {
  +       start         time.Time
  @@ -68,0 +71 @@ func (o *LogsOptions) SetClientset(clientset *clientset.Clientset) {
  +       o.start = time.Now()
  @@ -160,0 +164 @@ func (o *LogsOptions) Run(_ context.Context) error {
  +                               fmt.Println("=============================", time.Since(o.start), "=============================")
  ```

### DIFF
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -28,7 +28,9 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
 	var out []unstructured.Unstructured
 	outChan := make(chan []unstructured.Unstructured)
+	defer close(outChan)
 	errChan := make(chan struct{}) // use this to ensure that go routine doesn't keep waiting on outChan
+	defer close(errChan)
 
 	var apisOfInterest []apiResource
 	for _, api := range apis {
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -43,6 +43,7 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 	klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
 
 	for _, api := range apisOfInterest {
+		api := api // shadowing because go vet complains "loop variable api captured by func literal"
 		group.Go(func() error {
 			klog.V(4).Infof("[query api] start: %s", api.GroupVersionResource())
 			v, err := queryAPI(client, api, ns, selector)
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -3,8 +3,8 @@ package kclient
 import (
 	"context"
 	"fmt"
+	"golang.org/x/sync/errgroup"
 	"strings"
-	"sync"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,7 +26,6 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 }
 
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
-	var wg sync.WaitGroup
 	var out []unstructured.Unstructured
 	outChan := make(chan []unstructured.Unstructured)
 
@@ -40,34 +39,33 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 	}
 
 	start := time.Now()
+	group := new(errgroup.Group) // an error group errors when any of the go routines encounters an error
 	klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
 
-	var errResult error
 	for _, api := range apisOfInterest {
-		wg.Add(1)
-		go func(a apiResource) {
-			defer wg.Done()
-			klog.V(4).Infof("[query api] start: %s", a.GroupVersionResource())
-			v, err := queryAPI(client, a, ns, selector)
+		group.Go(func() error {
+			klog.V(4).Infof("[query api] start: %s", api.GroupVersionResource())
+			v, err := queryAPI(client, api, ns, selector)
 			if err != nil {
-				klog.V(4).Infof("[query api] error querying: %s, error=%v", a.GroupVersionResource(), err)
-				errResult = err
-				return
+				klog.V(4).Infof("[query api] error querying: %s, error=%v", api.GroupVersionResource(), err)
+				return err
 			}
 			outChan <- v
-			klog.V(4).Infof("[query api]  done: %s, found %d apis", a.GroupVersionResource(), len(v))
-		}(api)
+			klog.V(4).Infof("[query api]  done: %s, found %d apis", api.GroupVersionResource(), len(v))
+			return nil
+		})
 	}
 
-	for i := 0; i < len(apisOfInterest); i++ {
-		out = append(out, <-outChan...)
-	}
+	go func() {
+		for val := range outChan {
+			out = append(out, val...)
+		}
+	}()
 
 	klog.V(2).Infof("fired up all goroutines to query APIs")
-	wg.Wait()
 	klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
-	klog.V(2).Infof("query result: error=%v, objects=%d", errResult, len(out))
-	return out, errResult
+	klog.V(2).Infof("query result: objects=%d", len(out))
+	return out, group.Wait()
 }
 
 func queryAPI(client dynamic.Interface, api apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -26,19 +26,24 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 }
 
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
-	var mu sync.Mutex
 	var wg sync.WaitGroup
 	var out []unstructured.Unstructured
+	outChan := make(chan []unstructured.Unstructured)
 
-	start := time.Now()
-	klog.V(2).Infof("starting to query %d APIs in concurrently", len(apis))
-
-	var errResult error
+	var apisOfInterest []apiResource
 	for _, api := range apis {
 		if !api.r.Namespaced {
 			klog.V(4).Infof("[query api] api (%s) is non-namespaced, skipping", api.r.Name)
 			continue
 		}
+		apisOfInterest = append(apisOfInterest, api)
+	}
+
+	start := time.Now()
+	klog.V(2).Infof("starting to query %d APIs in concurrently", len(apis))
+
+	var errResult error
+	for _, api := range apisOfInterest {
 		wg.Add(1)
 		go func(a apiResource) {
 			defer wg.Done()
@@ -49,11 +54,13 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 				errResult = err
 				return
 			}
-			mu.Lock()
-			out = append(out, v...)
-			mu.Unlock()
+			outChan <- v
 			klog.V(4).Infof("[query api]  done: %s, found %d apis", a.GroupVersionResource(), len(v))
 		}(api)
+	}
+
+	for i := 0; i < len(apisOfInterest); i++ {
+		out = append(out, <-outChan...)
 	}
 
 	klog.V(2).Infof("fired up all goroutines to query APIs")
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -56,6 +56,7 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 			return nil
 		})
 	}
+	klog.V(2).Infof("fired up all goroutines to query APIs")
 
 	go func() {
 		for val := range outChan {
@@ -63,10 +64,12 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 		}
 	}()
 
-	klog.V(2).Infof("fired up all goroutines to query APIs")
+	err := group.Wait()
 	klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
+	close(outChan) // because we are ranging over the channel in a goroutine
+
 	klog.V(2).Infof("query result: objects=%d", len(out))
-	return out, group.Wait()
+	return out, err
 }
 
 func queryAPI(client dynamic.Interface, api apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -28,9 +28,6 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
 	var out []unstructured.Unstructured
 	outChan := make(chan []unstructured.Unstructured)
-	defer close(outChan)
-	errChan := make(chan struct{}) // use this to ensure that go routine doesn't keep waiting on outChan
-	defer close(errChan)
 
 	var apisOfInterest []apiResource
 	for _, api := range apis {
@@ -52,7 +49,6 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 			v, err := queryAPI(client, api, ns, selector)
 			if err != nil {
 				klog.V(4).Infof("[query api] error querying: %s, error=%v", api.GroupVersionResource(), err)
-				errChan <- struct{}{}
 				return err
 			}
 			outChan <- v
@@ -62,21 +58,20 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 	}
 	klog.V(2).Infof("fired up all goroutines to query APIs")
 
+	errChan := make(chan error)
 	go func() {
-		for {
-			select {
-			case <-errChan:
-				return
-			case val := <-outChan:
-				out = append(out, val...)
-			}
-		}
+		err := group.Wait()
+		klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
+		close(outChan)
+		errChan <- err
 	}()
 
-	err := group.Wait()
-	klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
+	for v := range outChan {
+		out = append(out, v...)
+	}
+
 	klog.V(2).Infof("query result: objects=%d", len(out))
-	return out, err
+	return out, <-errChan
 }
 
 func queryAPI(client dynamic.Interface, api apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -28,6 +28,7 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
 	var out []unstructured.Unstructured
 	outChan := make(chan []unstructured.Unstructured)
+	errChan := make(chan struct{}) // use this to ensure that go routine doesn't keep waiting on outChan
 
 	var apisOfInterest []apiResource
 	for _, api := range apis {
@@ -49,6 +50,7 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 			v, err := queryAPI(client, api, ns, selector)
 			if err != nil {
 				klog.V(4).Infof("[query api] error querying: %s, error=%v", api.GroupVersionResource(), err)
+				errChan <- struct{}{}
 				return err
 			}
 			outChan <- v
@@ -59,15 +61,18 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 	klog.V(2).Infof("fired up all goroutines to query APIs")
 
 	go func() {
-		for val := range outChan {
-			out = append(out, val...)
+		for {
+			select {
+			case <-errChan:
+				return
+			case val := <-outChan:
+				out = append(out, val...)
+			}
 		}
 	}()
 
 	err := group.Wait()
 	klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
-	close(outChan) // because we are ranging over the channel in a goroutine
-
 	klog.V(2).Infof("query result: objects=%d", len(out))
 	return out, err
 }
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -40,7 +40,7 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
 	}
 
 	start := time.Now()
-	klog.V(2).Infof("starting to query %d APIs in concurrently", len(apis))
+	klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
 
 	var errResult error
 	for _, api := range apisOfInterest {
--- a/pkg/logs/logs.go
+++ b/pkg/logs/logs.go
@@ -3,8 +3,6 @@ package logs
 import (
 	"fmt"
 	"io"
-	"k8s.io/apimachinery/pkg/runtime"
-
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -116,46 +114,41 @@ func (o *LogsClient) getPodsForSelector(
 	namespace string,
 	podChan chan corev1.Pod,
 ) error {
+	// set of unique Pods with Pod name as key; these are the Pods whose logs we want to get from the cluster
+	pods := map[string]struct{}{}
+
+	podList, err := o.kubernetesClient.GetPodsMatchingSelector(selector)
+	if err != nil {
+		return err
+	}
+	for _, pod := range podList.Items {
+		podChan <- pod
+		pods[pod.GetName()] = struct{}{}
+	}
+
 	resources, err := o.kubernetesClient.GetAllResourcesFromSelector(selector, namespace)
 	if err != nil {
 		return err
 	}
-	// set of unique Pods with Pod name as key; these are the Pods whose logs we want to get from the cluster
-	pods := map[string]struct{}{}
-
-	// if there's a Pod in the resources, we add it to the set of Pods whose logs we are interested in
-	for _, r := range resources {
-		if r.GetKind() == "Pod" {
-			var pod corev1.Pod
-			err = runtime.DefaultUnstructuredConverter.FromUnstructured(r.Object, &pod)
-			if err != nil {
-				return err
-			}
-			pods[pod.GetName()] = struct{}{}
-			if podChan != nil {
-				podChan <- pod
-			}
-		}
-	}
 
 	// get all pods in the namespace
-	podList, err := o.kubernetesClient.GetAllPodsInNamespace()
+	podList, err = o.kubernetesClient.GetAllPodsInNamespace()
 	if err != nil {
 		return err
 	}
 
 	// match pod ownerReference (if any) with resources matching the selector
 	for _, pod := range podList.Items {
+		if _, ok := pods[pod.GetName()]; ok {
+			// Pod's logs have already been displayed to user
+			continue
+		}
 		match := false
 		for _, owner := range pod.GetOwnerReferences() {
 			match, err = o.matchOwnerReferenceWithResources(owner, resources)
 			if err != nil {
 				return err
 			} else if match {
-				if _, ok := pods[pod.GetName()]; ok {
-					// Pod's logs have already been displayed to user
-					continue
-				}
 				pods[pod.GetName()] = struct{}{}
 				podChan <- pod
 				break // because we don't need to check other owner references of the pod anymore
--- a/pkg/kclient/deployments.go
+++ b/pkg/kclient/deployments.go
@@ -69,8 +69,7 @@ func (c *Client) GetOneDeploymentFromSelector(selector string) (*appsv1.Deployme
 	return &deployments[0], nil
 }
 
-// GetDeploymentFromSelector returns an array of Deployment resources which
-// match the given selector
+// GetDeploymentFromSelector returns an array of Deployment resources which match the given selector
 func (c *Client) GetDeploymentFromSelector(selector string) ([]appsv1.Deployment, error) {
 	var deploymentList *appsv1.DeploymentList
 	var err error
--- a/pkg/kclient/interface.go
+++ b/pkg/kclient/interface.go
@@ -113,6 +113,7 @@ type ClientInterface interface {
 	GetOnePodFromSelector(selector string) (*corev1.Pod, error)
 	GetPodLogs(podName, containerName string, followLog bool) (io.ReadCloser, error)
 	GetAllPodsInNamespace() (*corev1.PodList, error)
+	GetPodsMatchingSelector(selector string) (*corev1.PodList, error)
 
 	// port_forwarding.go
 	// SetupPortForwarding creates port-forwarding for the pod on the port pairs provided in the
--- a/pkg/kclient/mock_Client.go
+++ b/pkg/kclient/mock_Client.go
@@ -828,6 +828,21 @@ func (mr *MockClientInterfaceMockRecorder) GetPodUsingComponentName(componentNam
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetPodUsingComponentName", reflect.TypeOf((*MockClientInterface)(nil).GetPodUsingComponentName), componentName)
 }
 
+// GetPodsMatchingSelector mocks base method.
+func (m *MockClientInterface) GetPodsMatchingSelector(selector string) (*v11.PodList, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetPodsMatchingSelector", selector)
+	ret0, _ := ret[0].(*v11.PodList)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// GetPodsMatchingSelector indicates an expected call of GetPodsMatchingSelector.
+func (mr *MockClientInterfaceMockRecorder) GetPodsMatchingSelector(selector interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetPodsMatchingSelector", reflect.TypeOf((*MockClientInterface)(nil).GetPodsMatchingSelector), selector)
+}
+
 // GetProject mocks base method.
 func (m *MockClientInterface) GetProject(projectName string) (*v1.Project, error) {
 	m.ctrl.T.Helper()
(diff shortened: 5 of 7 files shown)


pkg/kclient/all.go
```diff
@@ -26,19 +26,24 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
 }
 
 func getAllResources(client dynamic.Interface, apis []apiResource, ns string, selector string) ([]unstructured.Unstructured, error) {
-	var mu sync.Mutex
 	var wg sync.WaitGroup
 	var out []unstructured.Unstructured
+	outChan := make(chan []unstructured.Unstructured)
```
feloy, line 31: I would probably create a new function instead of modifying this generic one: this optimization is valid for logs, probably not for other routines using it.

dharmit: > I would probably create a new function instead of modifying this generic one: this optimization is valid for logs, probably not for other routines using it.


I made this optimization with two things in mind:
1. mutex vs channel
2. all go routines trying to append to a single slice
 
Instead of locking and unlocking every time, I am putting the received slice on the channel and using it in `append` call later. We can do without a mutex because it's an unbuffered channel, and hence only one go routine can put `[]unstructured.Unstructured` to it at a time. So, I think this optimization should be generalized because it's made purely from Go perspective without a thought about which functionality of odo could benefit from it.

Do you see a downside to this optimization being generalized? If yes, I would surely like to understand what it is, and try to optimize this only from `odo logs` perspective instead of generalizing it.

feloy: You are right, I misread your changes in this function

pkg/logs/logs.go
```diff
@@ -116,46 +114,41 @@ func (o *LogsClient) getPodsForSelector(
...
-			}
-		}
+	resources, err := o.kubernetesClient.GetAllResourcesFromSelector(selector, namespace)
+	if err != nil {
+		return err
 	}
 
 	// get all pods in the namespace
-	podList, err := o.kubernetesClient.GetAllPodsInNamespace()
+	podList, err = o.kubernetesClient.GetAllPodsInNamespace()
```
valaparthvi, line 135: I am not sure I understand why you want to get a list of all the pods in the namespace when you already have a list of the pods that match a given selector. Shouldn't that be enough?

pkg/kclient/all.go
```diff
@@ -49,13 +54,15 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
...
-			mu.Lock()
-			out = append(out, v...)
-			mu.Unlock()
+			outChan <- v
 			klog.V(4).Infof("[query api]  done: %s, found %d apis", a.GroupVersionResource(), len(v))
 		}(api)
 	}
 
+	for i := 0; i < len(apisOfInterest); i++ {
+		out = append(out, <-outChan...)
```
valaparthvi, line 63: This might be silly, but can you not nest this line to the for loop above that also iterates over apisOfInterest, or does it have something to do with go routines and channels?

pkg/kclient/all.go
```diff
@@ -26,19 +26,24 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
...
 	for _, api := range apis {
 		if !api.r.Namespaced {
 			klog.V(4).Infof("[query api] api (%s) is non-namespaced, skipping", api.r.Name)
 			continue
 		}
+		apisOfInterest = append(apisOfInterest, api)
+	}
+
+	start := time.Now()
+	klog.V(2).Infof("starting to query %d APIs in concurrently", len(apis))
```
valaparthvi, line 43:
```suggestion
	klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
```

pkg/logs/logs.go
dharmit: > I am not sure I understand why you want to get a list of all the pods in the namespace when you already have a list of the pods that match a given selector. Shouldn't that be enough?

If a Pod is created for a Deployment in Deploy mode, it won't have the selector, and hence won't show up in the list of Pods matching the given selector.

pkg/kclient/all.go
dharmit: > This might be silly,

It's not silly. In fact, last time I made this exact change in the code (but didn't push it), it took me probably around an hour to get it right without introducing a deadlock.

And you are not wrong. It "just works" if I make the change you described, but...

> can you not nest this line to the for loop above that also iterates over apisOfInterest, or does it have something to do with go routines and channels?

But that would defeat the purpose of spinning up the go routines. If I move the line `out = append(out, <-outChan...)` just before the end of the previous `for` loop, it will be a blocking call that won't increment the `for` loop, and hence won't spin up the next go routine, till the current go routine has put something on `outChan`.

If you test against a local minikube/CRC instance, the performance issue won't be noticeable, but take a look at the below benchmark for a CRC VM running in North America to which I am connected via VPN:

```sh
# with code as-is in the PR
$ odo-pods-first logs | grep "===="
============================= 9.079955189s =============================
============================= 9.339786147s =============================

# after making the change you are wondering about
# basically remove the extra for loop and move the append statement to just after calling the go routine
$ odo logs | grep "===="
============================= 1m1.383878082s =============================
============================= 1m1.650332361s =============================

# I hear you say "make it a buffered channel instead of an unbuffered one"
# ok, here's the result after that change where I make outChan buffered
# by doing outChan := make(chan []unstructured.Unstructured, len(apisOfInterest))
# it's still a blocking call
$ odo logs | grep "===="
============================= 58.996780932s =============================
============================= 59.300285635s =============================
```

So an extra `for` loop is well worth it, IMO.

rm3l, line 63: If one of the goroutines fired returns an error, it will not write to `outChan`. So attempting to read from `outChan` `len(apisOfInterest)` times could block and hang up the `odo logs` command, I think.

rm3l (changes requested):

pkg/kclient/all.go
rm3l: > If one of the goroutines fired returns an error, it will not write to `outChan`. So attempting to read from `outChan` `len(apisOfInterest)` times could block the `odo logs` command, I think.

Confirmed by simulating an error from time to time with the following diff:
```diff
diff --git a/pkg/kclient/all.go b/pkg/kclient/all.go
index f1a0abe52..85bdbe9ff 100644
--- a/pkg/kclient/all.go
+++ b/pkg/kclient/all.go
@@ -43,12 +43,16 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
        klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
 
        var errResult error
+       var i int
        for _, api := range apisOfInterest {
                wg.Add(1)
-               go func(a apiResource) {
+               go func(a apiResource, j int) {
                        defer wg.Done()
                        klog.V(4).Infof("[query api] start: %s", a.GroupVersionResource())
                        v, err := queryAPI(client, a, ns, selector)
+                       if err == nil && j%10 == 5 {
+                               err = fmt.Errorf("XXX an error for j=%d", j)
+                       }
                        if err != nil {
                                klog.V(4).Infof("[query api] error querying: %s, error=%v", a.GroupVersionResource(), err)
                                errResult = err
@@ -56,7 +60,8 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
                        }
                        outChan <- v
                        klog.V(4).Infof("[query api]  done: %s, found %d apis", a.GroupVersionResource(), len(v))
-               }(api)
+               }(api, i)
+               i++
        }
 
        for i := 0; i < len(apisOfInterest); i++ {

```

pkg/kclient/all.go
```diff
@@ -26,19 +26,24 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
...
 			klog.V(4).Infof("[query api] api (%s) is non-namespaced, skipping", api.r.Name)
 			continue
 		}
+		apisOfInterest = append(apisOfInterest, api)
+	}
+
+	start := time.Now()
+	klog.V(2).Infof("starting to concurrently query %d APIs", len(apis))
+
+	var errResult error
```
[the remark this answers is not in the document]
dharmit: f3649ef

pkg/kclient/all.go
dharmit: f3649ef

pkg/kclient/all.go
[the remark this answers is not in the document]
dharmit: What you brought up on this is, without a doubt, technically correct. But I think that it is something that would be hugely beneficial for APIs. It's very subjective, and I'm not debating it. `errgroup` was new to me, and I had fun learning and implementing it. 😆

pkg/kclient/all.go
dharmit: Good catch! Initially I tried by `close`'ing the channel when a go routine encountered an error. But it was prone to panic and race conditions. Eventually, I ended up implementing the pattern of `range`'ing over a channel in a go routine. Take a look.

pkg/kclient/all.go
```diff
@@ -26,41 +26,47 @@ func (c *Client) GetAllResourcesFromSelector(selector string, ns string) ([]unst
...
-			klog.V(4).Infof("[query api]  done: %s, found %d apis", a.GroupVersionResource(), len(v))
-		}(api)
+			outChan <- v
+			klog.V(4).Infof("[query api]  done: %s, found %d apis", api.GroupVersionResource(), len(v))
+			return nil
+		})
 	}
 
+	go func() {
+		for val := range outChan {
```
rm3l, line 61: I think this goroutine will leak and keep running indefinitely.
Maybe I'm wrong, but from my understanding, this `for-range` loop will finish only when the channel is closed.
So if we run a long-running process like `odo logs --follow`, this goroutine will stay here forever, waiting for `outChan` to be closed.
Also, I noticed that `getAllResources` might be called several times (tested with a component in both Dev and Deploy modes), and this led to multiple goroutines leaked and waiting forever for `outChan` to be closed.
Maybe it makes more sense to have a dedicated goroutine that calls `group.Wait()` and closes `outChan`? Then outside, this `for-range` loop. This way, we are sure that `outChan` is closed properly after all goroutines fired have returned (successfully or not), so no risk of panic'ing, IMO. I guess this might require to capture the error returned by `group.Wait()` in a dedicated error channel..

Maybe it is more clear with some code :)
```go
klog.V(2).Infof("fired up all goroutines to query APIs")

// Buffered channel of size 1 so as not to block when writing the error,
// but `group.Wait()` waits for the goroutines in the group to return (or any of them returning an error).
errChan := make(chan error, 1)
go func() {
  errChan <- group.Wait()
  close(outChan)
})

for v := range outChan {
  out = append(out, v...)
}

//At this point, we are sure that outChan is closed and we have everything,
// even if any of the goroutines fired returned an error.

klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
klog.V(2).Infof("query result: objects=%d", len(out))
return out, <-errChan
```
 
What do you think?

rm3l (changes requested):

pkg/kclient/all.go
dharmit: > Maybe I'm wrong, but from my understanding, this `for-range` loop will finish only when the channel is closed.

Nope, you are right. I wasn't aware of this till I read your comment and looked it up. 



> ```go
> // Buffered channel of size 1 so as not to block when writing the error,
> // but `group.Wait()` waits for the goroutines in the group to return
> // (or any of them returning an error).
> ```

I don't understand the rationale behind buffer size of 1. Based on the [documentation of `Wait()`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait), it's going to return the first non-nil error if any of the go routines had any. I don't think the channel's going to be written to again by `group.Wait()`.

Otherwise, what you have suggested, makes sense overall.

dharmit: [b3001cb](https://github.com/redhat-developer/odo/pull/5973/commits/b3001cbdedbcb1aa354f166ebdc7bd0a7af262d0)

rm3l: > > Maybe I'm wrong, but from my understanding, this `for-range` loop will finish only when the channel is closed.
> 
> Nope, you are right. I wasn't aware of this till I read your comment and looked it up.
> 
> > ```go
> > // Buffered channel of size 1 so as not to block when writing the error,
> > // but `group.Wait()` waits for the goroutines in the group to return
> > // (or any of them returning an error).
> > ```
> 
> I don't understand the rationale behind buffer size of 1. Based on the [documentation of `Wait()`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait), it's going to return the first non-nil error if any of the go routines had any. I don't think the channel's going to be written to again by `group.Wait()`.
> 

This was mainly to have a non-blocking write to `errChan`. Otherwise, this write operation waits for a read, which actually happens after the `for-range` loop (which also blocks until `outChan` is closed).

pkg/kclient/all.go
```diff
@@ -56,17 +56,23 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
...
 		for val := range outChan {
 			out = append(out, val...)
 		}
 	}()
 
-	klog.V(2).Infof("fired up all goroutines to query APIs")
-	klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
 	klog.V(2).Infof("query result: objects=%d", len(out))
-	return out, group.Wait()
+	return out, <-errChan
```
rm3l, line 75: Okay, with the `for-range` in a separate goroutine, we avoid making `errChan` a buffered channel. This function will return as soon as a value is read from `errChan`, but I think we still cannot guarantee that the `for-range` has finished iterating at this point. So we cannot be sure that we have read everything from `outChan`, no?

dharmit: Why would we want to read everything from `outChan` if there was an error?

rm3l: I meant that even if there was no error, `out` would contain everything only after `outChan` is closed, which, in your changes, happens after the result of `group.Wait()` has been written to `errChan`. But at the same time, `<-errChan` would no longer be blocking, but we are not sure `outChan` has been closed yet when returning.
In other words, `return out, <-errChan` could return before we are done iterating over `outChan`, I think.

rm3l: If you are iterating over `outChan` in a separate goroutine, then maybe the calls to `group.Wait()` and `close(outChan)` could be done right in the function itself, not another goroutine.. I think this would be useful if we were `for-range`'ing in the main function itself.

```diff
-       errChan := make(chan error)
-       go func() {
-               errChan <- group.Wait()
-               klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
-               close(outChan) // because we are ranging over the channel below
-       }()
-
        go func() {
                for val := range outChan {
                        out = append(out, val...)
                }
        }()
 
+       err := group.Wait()
+       klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
+       close(outChan)
+
        klog.V(2).Infof("query result: objects=%d", len(out))
-       return out, <-errChan
+       return out, err

```

pkg/kclient/all.go
dharmit: > Otherwise, this write operation waits for a read, which only happens after the `for-range` loop (which also blocks until `outChan` is closed).

Your concern is valid. But when a go routine that is waiting for something for longer time is encountered, the go scheduler takes care of being fair and not letting it run all the time. This is something I learned from [this talk](https://www.youtube.com/watch?v=wQpC99Xu1U4). I watched it many months back when I had not even started writing much concurrent code. So if you find something different from my understanding, let me know!

dharmit: > > Otherwise, this write operation waits for a read, which only happens after the `for-range` loop (which also blocks until `outChan` is closed).
> 
> Your concern is valid. But when a go routine that is waiting for something for longer time is encountered, the go scheduler takes care of being fair and not letting it run all the time. This is something I learned from [this talk](https://www.youtube.com/watch?v=wQpC99Xu1U4). I watched it many months back when I had not even started writing much concurrent code. So if you find something different from my understanding, let me know!

> This was mainly to have a non-blocking write to `errChan`.

Even with the above concern, I can't understand how a buffer size of 1 would have helped. If the `range` block was going to block, what difference was making the `errChan` buffered going to make? 🤔

rm3l: > > > Otherwise, this write operation waits for a read, which only happens after the `for-range` loop (which also blocks until `outChan` is closed).
> 
> Your concern is valid. But when a go routine that is waiting for something for longer time is encountered, the go scheduler takes care of being fair and not letting it run all the time. This is something I learned from [this talk](https://www.youtube.com/watch?v=wQpC99Xu1U4). I watched it many months back when I had not even started writing much concurrent code. So if you find something different from my understanding, let me know!
> 
Interesting - thanks for the link to this talk. Will watch it soon.

> > This was mainly to have a non-blocking write to `errChan`.
> 
> Even with the above concern, I can't understand how a buffer size of 1 would have helped. If the `range` block was going to block, what difference was making the `errChan` buffered going to make? thinking
>
In my code sample, the buffer size of 1 was there mainly to be able to call `close(outChan)`. Otherwise, `errChan <- group.Wait()` would block waiting for a reader; but in my example, we read from `errChan` only after the `for-range` loop (which itself blocks until `close(outChan)` is called). So deadlock without the buffer size of 1.

pkg/kclient/all.go
dharmit: > I meant that even if there was no error, `out` would contain everything only after `outChan` is closed, which, in your changes, happens after the result of `group.Wait()` has been written to `errChan`. But at the same time, `<-errChan` would no longer be blocking, but we are not sure `outChan` has been closed yet when returning.
> In other words, `return out, <-errChan` could return before we are done iterating over `outChan`, I think.

I still don't find the answer to - Why would we want to read everything from `outChan` if there was an error?

> I meant that even if there was no error, `out` would contain everything only after `outChan` is closed, which, in your changes, happens after the result of `group.Wait()` has been written to `errChan`.

`out` already has it. It's just that the goroutine ranging over `outChan` keeps ranging over the channel. That's how, in spite of https://github.com/redhat-developer/odo/pull/5973#discussion_r933315813, we get the expected result with `odo logs` and `odo logs --follow`. Besides that, there's no other time when we could close the `outChan` - we need to wait til `group.Wait()` is done.

> But at the same time, `<-errChan` would no longer be blocking, but we are not sure `outChan` has been closed yet when returning.

And why do we want to be sure to close the `outChan`? We encountered an error, and we want that to pass on to the user as soon as doable. Are you concerned about the leaky aspect of it? Since we are stopping things as soon as we encounter an error, it's not a problem for us. If we were running an API server, we would have to be bothered about closing the channel.

rm3l: > > I meant that even if there was no error, `out` would contain everything only after `outChan` is closed, which, in your changes, happens after the result of `group.Wait()` has been written to `errChan`. But at the same time, `<-errChan` would no longer be blocking, but we are not sure `outChan` has been closed yet when returning.
> > In other words, `return out, <-errChan` could return before we are done iterating over `outChan`, I think.
> 
> I still don't find the answer to - Why would we want to read everything from `outChan` if there was an error?
> 
> > I meant that even if there was no error, `out` would contain everything only after `outChan` is closed, which, in your changes, happens after the result of `group.Wait()` has been written to `errChan`.
> 
> `out` already has it. It's just that the goroutine ranging over `outChan` keeps ranging over the channel. That's how, in spite of [#5973 (comment)](https://github.com/redhat-developer/odo/pull/5973#discussion_r933315813), we get the expected result with `odo logs` and `odo logs --follow`. Besides that, there's no other time when we could close the `outChan` - we need to wait til `group.Wait()` is done.
> 
> > But at the same time, `<-errChan` would no longer be blocking, but we are not sure `outChan` has been closed yet when returning.
> 
> And why do we want to be sure to close the `outChan`? We encountered an error, and we want that to pass on to the user as soon as doable. Are you concerned about the leaky aspect of it? Since we are stopping things as soon as we encounter an error, it's not a problem for us. If we were running an API server, we would have to be bothered about closing the channel.

I agree with you about the case where we encounter an error. I was mostly thinking about the other case, where `group.Wait()` would not return any error (`nil` error written to `errChan` and read from it): my point was that with the previous changes, `out` could have been filled partially if `outChan` was not closed before the function returns.
But overall yes, my concern is about not leaking goroutines, but with long-running processes in mind, like `odo logs --follow` (or any other long-running command that could use this function).

Anyways, thanks for the discussion. Your new changes now look good to me.

rm3l (commented):
I can see some IBM tests not passing for Kubernetes, like:
```
    [odo] This will delete "mynodejs" from the namespace "cmd-delete-test186nfv".
    [odo]  •  The component contains the following resources that will get deleted:
    [odo] 	- Endpoints: my-cs
    [odo] 	- Deployment: my-component
    [odo] 	- EndpointSlice: my-cs-k5b78
    [odo] The component "mynodejs" is successfully deleted from namespace "cmd-delete-test186nfv"
    [odo] I0801 11:07:18.434601   21846 odo.go:75] Could not get the latest release information in time. Never mind, exiting gracefully :)
    STEP: listing the resource to delete 08/01/22 11:07:18.441
    Deleting project: cmd-delete-test186nfv
    Running kubectl with args [kubectl delete namespaces cmd-delete-test186nfv --wait=false]
    [kubectl] namespace "cmd-delete-test186nfv" deleted
    Deleting dir: /tmp/930020196
    Deleting dir: /tmp/1703044102
  << End Captured GinkgoWriter Output

  Expected
      <string>: Searching resources to delete, please wait...
      This will delete "mynodejs" from the namespace "cmd-delete-test186nfv".
      	- Endpoints: my-cs
      	- Deployment: my-component
      	- EndpointSlice: my-cs-k5b78
      The component "mynodejs" is successfully deleted from namespace "cmd-delete-test186nfv"
      
  to contain substring
      <string>: Service: my-cs
  In [It] at: /go/odo_1/tests/integration/cmd_delete_test.go:189
```

Could that be related to the changes here?

dharmit:
> Could that be related to the changes here?

Nope.

Even the OpenShift failures on IBM Cloud are because the Pod fails to start. I'll rebase and force push to re-trigger a test run. Let's see.

dharmit:
IBM Cloud k8s test failure:
```
Running kubectl with args [kubectl create namespace cmd-add-binding-test253tnt]
```

rm3l (approved):

dharmit:
IBM Cloud kubernetes tests:
```
odo dev command tests
/go/odo_1/tests/integration/cmd_dev_test.go:27
  when a component is bootstrapped and pushed
  /go/odo_1/tests/integration/cmd_dev_test.go:57
    when odo dev is executed
    /go/odo_1/tests/integration/cmd_dev_test.go:188
      when deleting previous deployment and switching kubeconfig to another namespace [BeforeEach]
      /go/odo_1/tests/integration/cmd_dev_test.go:204
        should run odo dev on initial namespace
        /go/odo_1/tests/integration/cmd_dev_test.go:220

...
...
[odo] The component "hdchtb" is successfully deleted from namespace "cmd-dev-test220avm"
    [odo] I0802 10:00:00.215112   12893 odo.go:75] Could not get the latest release information in time. Never mind, exiting gracefully :)
    Running kubectl with args [kubectl get deployment -n cmd-dev-test220avm]
    [kubectl] NAME         READY   UP-TO-DATE   AVAILABLE   AGE
    [kubectl] hdchtb-app   1/1     1            1           26s
    Deleting project: 
    Running kubectl with args [kubectl delete namespaces  --wait=false]
    [kubectl] error: resource name may not be empty
    Deleting project: cmd-dev-test220avm
    Running kubectl with args [kubectl delete namespaces cmd-dev-test220avm --wait=false]
    [kubectl] namespace "cmd-dev-test220avm" deleted
    Deleting dir: /tmp/3662520053
    Deleting dir: /tmp/2113309568
  << End Captured GinkgoWriter Output

  Expected
      <string>: 
  to contain substring
      <string>: No resources found in cmd-dev-test220avm namespace.
  In [BeforeEach] at: /go/odo_1/tests/integration/cmd_dev_test.go:207
```

Somehow, the name of the namespace didn't make it to the `kubectl delete namespaces` command. Passes locally.

dharmit:
Since I closed the PR and reopened it, and also rebased it against `main`, two builds were triggered. Both the builds are failing with different errors on IBM Cloud. And running things locally seems to be working just fine.

dharmit:
Two occurrences of below error on OCP tests on prow:
```
   Unexpected error:
      <*fs.PathError | 0xc000bbc030>: {
          Op: "read",
          Path: "|0",
          Err: <*poll.DeadlineExceededError | 0x3d844e0>{},
      }
      read |0: i/o timeout
  occurred
  In [It] at: /go/src/github.com/redhat-developer/odo/tests/helper/helper_interactive.go:121 
```

pkg/kclient/all.go
```diff
@@ -59,15 +61,18 @@ func getAllResources(client dynamic.Interface, apis []apiResource, ns string, se
...
 
 	go func() {
-		for val := range outChan {
-			out = append(out, val...)
+		for {
+			select {
+			case <-errChan:
+				return
+			case val := <-outChan:
+				out = append(out, val...)
```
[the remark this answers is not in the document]
dharmit: > I certainly miss something about goroutines, but does that mean that this goroutine will keep running indefinitely if there was nothing written to `errChan`? I don't see how it would return from this infinite loop otherwise. Again, as we previously discussed, I am thinking about long-running processes like `odo logs --follow`..

I'm struggling really badly with imagining things from `odo logs --follow` perspective. 

> But it seems to me that the channel was closed after all the goroutines returned. So I am not sure I understand why the channel would get closed too soon.

I initially ignored few test failures because they didn't happen in `odo logs` or `odo list`. But one failure in `odo delete component` piqued my interest. In the logs, it shows that 1 resource of kind `Deployment` was found, but the message printed when we do `odo delete component -f` didn't include the `Deployment: <name of the deployment>`. My understanding is that this was because of the channel getting closed just a little too early. Also, this failure was intermittent and did not happen 100% of the times. For example, this morning when I cancelled a few builds of Charlie's PR, the CI passed just fine with the previous commit.

> Or would the approach[1] of iterating over `outChan` in the `getAllResources` function itself (i.e., not in a goroutine) be helpful, in the sense that the `for-range` loop would block until `outChan` is closed? So we would be sure that we are done iterating over `outChan` before the function returns, while not leaking any goroutines in this function.

I did think about iterating over `outChan` without a goroutine. But where would I put the `group.Wait()` in that case? I think that:
*  if we put it after the loop that's iterating over `outChan`, it won't let `group.Wait()` return an error as soon as one of the goroutines returns an error. 
* if we put `group.Wait()` before the iteration over `outChan`, we would have to make `outChan` buffered to the size of `apisOfInterest`.

So I'm considering doing `defer close(errChan)` and `defer close(outChan)` right after creating them. So, when `getAllResources` is done, it would close the channels and the goroutine would no longer be waiting on the two channels. WDYT?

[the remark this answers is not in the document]
dharmit: > > I certainly miss something about goroutines, but does that mean that this goroutine will keep running indefinitely if there was nothing written to `errChan`? I don't see how it would return from this infinite loop otherwise. Again, as we previously discussed, I am thinking about long-running processes like `odo logs --follow`..
> 
> I'm struggling really badly with imagining things from `odo logs --follow` perspective.

I mean, I am just not able to keep `odo logs --follow` in my head while coding on this PR.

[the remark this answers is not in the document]
dharmit: > But one failure in `odo delete component` piqued my interest. In the logs, it shows that 1 resource of kind `Deployment` was found, but the message printed when we do `odo delete component -f` didn't include the `Deployment: <name of the deployment>`. My understanding is that this was because of the channel getting closed just a little too early. Also, this failure was intermittent and did not happen 100% of the times.

Just faced this in the latest failure. Here's the excerpt:
```
<< End Captured GinkgoWriter Output

  Expected
      <string>: Searching resources to delete, please wait...
      This will delete "mynodejs" from the namespace "cmd-delete-test186eqw".
      	- Service: my-cs
      	- Endpoints: my-cs
      The component "mynodejs" is successfully deleted from namespace "cmd-delete-test186eqw"
      
  to contain substring
      <string>: Deployment: my-component
  In [It] at: /go/odo_1/tests/integration/cmd_delete_test.go:188
```
I'm probably going to get rid of goroutines, or maybe just revert the `sync.Mutex` that was there earlier.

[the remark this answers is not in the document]
rm3l: > > I certainly miss something about goroutines, but does that mean that this goroutine will keep running indefinitely if there was nothing written to `errChan`? I don't see how it would return from this infinite loop otherwise. Again, as we previously discussed, I am thinking about long-running processes like `odo logs --follow`..
> 
> I'm struggling really badly with imagining things from `odo logs --follow` perspective.
> 
> > But it seems to me that the channel was closed after all the goroutines returned. So I am not sure I understand why the channel would get closed too soon.
> 
> I initially ignored few test failures because they didn't happen in `odo logs` or `odo list`. But one failure in `odo delete component` piqued my interest. In the logs, it shows that 1 resource of kind `Deployment` was found, but the message printed when we do `odo delete component -f` didn't include the `Deployment: <name of the deployment>`. My understanding is that this was because of the channel getting closed just a little too early. Also, this failure was intermittent and did not happen 100% of the times. For example, this morning when I cancelled a few builds of Charlie's PR, the CI passed just fine with the previous commit.
> 
> > Or would the approach[1] of iterating over `outChan` in the `getAllResources` function itself (i.e., not in a goroutine) be helpful, in the sense that the `for-range` loop would block until `outChan` is closed? So we would be sure that we are done iterating over `outChan` before the function returns, while not leaking any goroutines in this function.
> 
> I did think about iterating over `outChan` without a goroutine. But where would I put the `group.Wait()` in that case? I think that:
> 
> * if we put it after the loop that's iterating over `outChan`, it won't let `group.Wait()` return an error as soon as one of the goroutines returns an error.
> * if we put `group.Wait()` before the iteration over `outChan`, we would have to make `outChan` buffered to the size of `apisOfInterest`.
> 
>

Indeed, I agree with these points; that's why I suggested we put both `group.Wait()` and `close(outChan)` in the same "monitoring" goroutine? Well, that's a pattern I saw used with `sync.WaitGroup` (see for example the realistic example depicted in the "Using WaitGroups" section of the "Learning Go" book - pages 221 through 222).

And if we don't want to make `errChan` buffered, I think writing to it after closing `outChan` (rather than before) should fix the blocking issue. The write to `errChan` would block until a read happens, which will actually happen because the for-range loop was done (because `outChan`  was closed). Like so:

```go
    klog.V(2).Infof("fired up all goroutines to query APIs")

	errChan := make(chan error)
	go func() {
		err := group.Wait()
		klog.V(2).Infof("all goroutines have returned in %v", time.Since(start))
		close(outChan)
		errChan <- err
	}()

	for v := range outChan {
		out = append(out, v...)
	}

	klog.V(2).Infof("query result: objects=%d", len(out))
	return out, <-errChan
```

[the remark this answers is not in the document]
rm3l: > > But one failure in `odo delete component` piqued my interest. In the logs, it shows that 1 resource of kind `Deployment` was found, but the message printed when we do `odo delete component -f` didn't include the `Deployment: <name of the deployment>`. My understanding is that this was because of the channel getting closed just a little too early. Also, this failure was intermittent and did not happen 100% of the times.
> 
> Just faced this in the latest failure. Here's the excerpt:
> 
> ```
> << End Captured GinkgoWriter Output
> 
>   Expected
>       <string>: Searching resources to delete, please wait...
>       This will delete "mynodejs" from the namespace "cmd-delete-test186eqw".
>        •  The component contains the following resources that will get deleted:
>       	- Service: my-cs
>       	- Endpoints: my-cs
>       The component "mynodejs" is successfully deleted from namespace "cmd-delete-test186eqw"
>       
>   to contain substring
>       <string>: Deployment: my-component
>   In [It] at: /go/odo_1/tests/integration/cmd_delete_test.go:188
> ```
> 
> I'm probably going to get rid of goroutines, or maybe just revert the `sync.Mutex` that was there earlier.
> 
Can we try the approach I suggested above in https://github.com/redhat-developer/odo/pull/5973#discussion_r937164777 and see how it goes?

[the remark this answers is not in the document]
dharmit: > Can we try the approach I suggested above in [#5973 (comment)](https://github.com/redhat-developer/odo/pull/5973#discussion_r937164777) and see how it goes?

Sure. By just reading the code, I think it should solve things for us.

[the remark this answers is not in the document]
dharmit: Done. Let's see if the CI gods like it. As a relief to me, the k8s IBM Cloud tests and OCP prow tests both had the same failure. OCP IBM Cloud was showing different errors, but I'm not bothered about those at the moment. Let's see.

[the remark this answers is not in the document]
rm3l: Thanks Dharmit.

rm3l (approved):
Looks like all tests are now passing with the latest changes.

@dharmit From the performance perspective (which is the end goal of this PR), I didn't retest it, but I hope we are still good with those changes. Can you please confirm?

dharmit:
> @dharmit From the performance perspective (which is the end goal of this PR), I didn't retest it, but I hope we are still good with those changes. Can you please confirm? (I'll also test this against a remote cluster soon)

I didn't retest it either till the CI gods seemed OK.

Here are the results for the devfile mentioned in the PR description:
```sh
# built from the code in this PR
$ odo logs | grep "===="                                                                                                                    
============================= 1.451086565s =============================                                                                    
============================= 6.720378476s =============================                                                                    
============================= 7.08179099s =============================                                                                     
============================= 8.566956054s =============================                                                                    
============================= 8.848807171s =============================

# built from the current main branch
$ odo-main logs | grep "===="
============================= 2.579176905s =============================
============================= 6.295186235s =============================
============================= 7.158730706s =============================
============================= 8.008279701s =============================
============================= 8.315148635s =============================
```

dharmit:
Looks like infra error:
```
error: default cluster has no server defined
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"k8s.io/test-infra/prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2022-08-05T09:37:45Z"}
error: failed to execute wrapped command: exit status 1 
INFO[2022-08-05T09:37:54Z] Step integration-e2e-gather-extra failed after 20s. 
```

/test v4.10-integration-e2e

PR #6006, opened 2022-08-04
Enhance error message when user runs from a non-component directory

### LINKED ISSUE #5987: Strange message when deleting a non existent component
/kind bug

**Operating System:**
Win10Pro

**Output of `odo version`:**

```
odo v3.0.0-beta2 (c8e4458f4)

Server: https://api.sandbox-m2.ll9k.p1.openshiftapps.com:6443
Kubernetes: v1.23.5+9ce5071
```

`odo delete component -f `

in a folder with no `.odo` and no `devfile.yaml`

## Actual behavior

```
C:\work\src\tmp\quarkus\maven\rest>odo delete component -f
 X  The current directory does not represent an odo component.
To start editing your component, use "odo dev" and open this folder in your favorite IDE. Changes will be directly reflected on the cluster.
Or switch to directory with a component.
```

## Expected behavior

We should have the error message that there is no component in this folder and maybe some info how to create it

valaparthvi:
**What type of PR is this:**
/kind bug

Enhance error message when user runs from a non-component directory

Fixes #5987

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

1. Run `odo delete component` or any other command from an empty directory.

The error message is long and could use an improvement, so I am looking for suggestions.

### DIFF
--- a/pkg/odo/genericclioptions/localprovider.go
+++ b/pkg/odo/genericclioptions/localprovider.go
@@ -2,9 +2,12 @@ package genericclioptions
 
 import (
 	"errors"
+	"fmt"
 
+	"github.com/redhat-developer/odo/pkg/devfile/location"
 	"github.com/redhat-developer/odo/pkg/envinfo"
 	"github.com/redhat-developer/odo/pkg/odo/cmdline"
+	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
 )
 
 // GetValidEnvInfo accesses the environment file
@@ -33,15 +36,24 @@ func GetValidEnvInfo(cmdline cmdline.Cmdline) (*envinfo.EnvSpecificInfo, error) 
 
 	// Check to see if the environment file exists
 	if !envInfo.Exists() {
+		exitMessage := `The current directory does not represent an odo component. 
+
+To get started,%s
+  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
+
+Visit https://odo.dev for more information.`
+
+		if isEmpty, _ := location.DirIsEmpty(filesystem.DefaultFs{}, componentContext); isEmpty {
+			exitMessage = fmt.Sprintf(exitMessage, `
+  * Create and move to a new directory
+  * Use "odo init" to initialize an odo component in the folder.
+  * Use "odo dev" to deploy it on cluster.`)
+		} else {
+			exitMessage = fmt.Sprintf(exitMessage, `
+  * Use "odo dev" to initialize an odo component for this folder and deploy it on cluster.`)
+		}
 		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
-		return nil, errors.New(`The current directory does not represent an odo component. 
-
-To get started,
-  * Create and move to a new directory, or use an existing one.
-  * Run "odo init" from the directory to initialize an odo component.
-  * Start editing the component in an IDE and run "odo dev" to see your changes get reflected on the cluster.
-
-Visit https://odo.dev for more information.`)
+		return nil, errors.New(exitMessage)
 		//revive:enable:error-strings
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/odo/genericclioptions/localprovider.go
```diff
@@ -35,7 +35,7 @@ func GetValidEnvInfo(cmdline cmdline.Cmdline) (*envinfo.EnvSpecificInfo, error)
 	if !envInfo.Exists() {
 		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
 		return nil, errors.New(`The current directory does not represent an odo component.
-To start editing your component, use "odo dev" and open this folder in your favorite IDE. Changes will be directly reflected on the cluster.
+If you are working from an empty directory, use "odo init" to get started, else start editing your component, use "odo dev" and open this folder in your favorite IDE. Changes will be directly reflected on the cluster.
 To deploy your component to a cluster use "odo deploy".
 Or switch to directory with a component.`)
```
dharmit, line 40: How about rethinking this? I would go with bullets/list, and not go beyond 120 characters in a line. This one is a lot of content to read and make sense of.

Something like:

```
The current directory does not represent an odo component. 
* To get started with creating an odo component, create a new directory and cd into it.
* From that empty directory, use "odo init" to get started.
* Once you have initialized an odo component, you can start editing it in an IDE, and use `odo dev` to see your changes get reflected on the cluster.
```

I find mentioning `odo deploy` a bit pointless when the user is operating from an empty/incorrect directory. We might even the odo.dev docs if we want to be **SO** verbose!

dharmit (changes requested):

pkg/odo/genericclioptions/localprovider.go
valaparthvi: I like the idea about listing.

>I find mentioning odo deploy a bit pointless when the user is operating from an empty/incorrect directory.

I agree, I'll remove that.

>Let's even dump the entire odo.dev docs if we want to be SO verbose!

What do you think about pointing at odo.dev for more information then?

pkg/odo/genericclioptions/localprovider.go
```diff
@@ -34,10 +34,14 @@ func GetValidEnvInfo(cmdline cmdline.Cmdline) (*envinfo.EnvSpecificInfo, error)
...
-To deploy your component to a cluster use "odo deploy".
-Or switch to directory with a component.`)
+		return nil, errors.New(`The current directory does not represent an odo component. 
+
+To get started,
+  * Create and move to a new directory, or use an existing one.
+  * Run "odo init" from the directory to initialize an odo component.
+  * Start editing the component in an IDE and run "odo dev" to see your changes get reflected on the cluster.
+
+Visit https://odo.dev for more information.`)
```
dharmit, line 44: This looks good, but the issue was opened by the user who was trying to delete from a non-component dir. The entire messaging here is targeted for a user who hasn't created a component. 

Now one could say that the first point does tell the user to move to a directory. But it's not accurate.

Maybe we shouldn't display a generic message for `odo init` and `odo delete`. We could tailor a message specific to each one of them.

dharmit (changes requested):

pkg/odo/genericclioptions/localprovider.go
valaparthvi: >Maybe we shouldn't display a generic message for odo init and odo delete. We could tailor a message specific to each one of them.

That was my initial idea. But then I had to figure out how to determine if the directory was empty, and changing the error message seemed like simplest thing to do. I'll get to it.

valaparthvi: [39758f0](https://github.com/redhat-developer/odo/pull/6006/commits/39758f0e45d3b2544a136d3b2aab01cc29b7d8e6) feels like over engineering, but eh :shrug:

rm3l (commented):
/approve

valaparthvi:

Reason: prow job passed.

PR #6013, opened 2022-08-08
Restart port forwarding on failure

### LINKED ISSUE #5877: odo dev port forward timeouts if app is not ready
/kind bug

**Operating System:**
MacOS

**Output of `odo version`:**
```
$ odo version
odo v3.0.0-alpha3 (77ff2b89d)

Server: https://my.openshift.cluster:6443
Kubernetes: v1.23.5+3afdacb
```

I started a new project using `odo init`. I used the Go starter project and added a sleep before starting the web server. I ran the `odo dev` command, and when the `odo` said it was running, I tried to access the endpoint before the duration of the sleep. I also tried to access the endpoint multiple times after the sleep.

The intention of the sleep is to simulate if the container takes some time to initialize and reach a ready state, which could be handled by a readiness probe if it were running in Kubernetes. I originally experienced this behavior with a Java Spring Boot application that took about 10 seconds to reach a ready state.

## Actual behavior

The first request returned immediately with an empty reply from server. After the duration of the sleep, some requests were successful, but others would occasionally timeout with a connection reset by peer.

## Expected behavior

- I think odo should wait until the container is listening on the intended port before stating the application was ready
- I would not expect intermittent connection timeouts when the application is running

Here is the odo output, note the first error is from when the first request came through (before the sleep duration completed), the second error is from an intermittent error:
```
$ odo dev
  __
 /  \__     Developing using the app Devfile
 \__/  \    Namespace: vinny-test-odo
 \__/

 ✓  Waiting for Kubernetes resources [16s]
 ✓  Syncing files into the container [726ms]
 ✓  Building your application in container on cluster (command: build) [3s]

 - Forwarding from 127.0.0.1:40001 -> 8080

Watching for changes in the current directory /Users/vinny/projects/odo-testing
Press Ctrl+c to exit `odo dev` and delete resources from the cluster

E0624 16:53:03.016599    9294 portforward.go:400] an error occurred forwarding 40001 -> 8080: error forwarding port 8080 to pod 3b77b1f3d378f912ec93eda422f7bd5fe99f63dd63466249ae41364edb59effe, uid : port forward into network namespace "/var/run/netns/ea2bbe59-8ea2-40f2-aca4-01228c6072f7": failed to connect to localhost:8080 inside namespace 3b77b1f3d378f912ec93eda422f7bd5fe99f63dd63466249ae41364edb59effe: dial tcp [::1]:8080: connect: connection refused
E0624 16:54:20.148247    9294 portforward.go:340] error creating error stream for port 40001 -> 8080: Timeout occurred
```

Here is the output from the failed curl request:
```
$ # First request before sleep
$ curl -vvv 127.0.0.1:40001
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 40001 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:40001
> User-Agent: curl/7.64.1
> Accept: */*
>
* Empty reply from server
* Connection #0 to host 127.0.0.1 left intact
curl: (52) Empty reply from server
* Closing connection 0

$ # Intermittent timeout request
$ curl -vvv http://127.0.0.1:40001
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 40001 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:40001
> User-Agent: curl/7.64.1
> Accept: */*
>
* Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```

feloy:
**What type of PR is this:**

/kind bug

Fixes #5877 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -82,7 +82,8 @@ func (o *PFClient) StartPortForwarding(
 		return err
 	}
 
-	o.originalErrorHandlers = runtime.ErrorHandlers
+	o.originalErrorHandlers = append([]func(error){}, runtime.ErrorHandlers...)
+
 	runtime.ErrorHandlers = append(runtime.ErrorHandlers, func(err error) {
 		if err.Error() == "lost connection to pod" {
 			// Stop the low-level port forwarding
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -138,7 +138,7 @@ func (o *PFClient) StopPortForwarding() {
 	o.stopChan <- struct{}{}
 	o.stopChan = nil
 
-	// Ask the inifinte loop to stop
+	// Ask the infinite loop to stop
 	o.isRunning = false
 
 	// Wait for low level port forward to be finished
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -4,16 +4,19 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"time"
 
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/pkg/devfile/parser"
 	parsercommon "github.com/devfile/library/pkg/devfile/parser/data/v2/common"
+	"k8s.io/apimachinery/pkg/util/runtime"
 
 	"github.com/redhat-developer/odo/pkg/kclient"
 	"github.com/redhat-developer/odo/pkg/libdevfile"
 	"github.com/redhat-developer/odo/pkg/log"
 	"github.com/redhat-developer/odo/pkg/state"
 	"github.com/redhat-developer/odo/pkg/util"
+	"github.com/redhat-developer/odo/pkg/watch"
 )
 
 var _ Client = (*PFClient)(nil)
@@ -28,6 +31,11 @@ type PFClient struct {
 	stopChan chan struct{}
 	// finishedChan is written when the port forwarding is finished
 	finishedChan chan struct{}
+
+	originalErrorHandlers []func(error)
+
+	// indicates that the port forwarding is started, and not stopped
+	isRunning bool
 }
 
 func NewPFClient(kubernetesClient kclient.ClientInterface, stateClient state.Client) *PFClient {
@@ -58,7 +66,6 @@ func (o *PFClient) StartPortForwarding(
 	o.StopPortForwarding()
 
 	o.stopChan = make(chan struct{}, 1)
-	o.finishedChan = make(chan struct{}, 1)
 
 	var portPairs map[string][]string
 	if randomPorts {
@@ -76,10 +83,33 @@ func (o *PFClient) StartPortForwarding(
 	}
 
 	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
+
+	o.originalErrorHandlers = runtime.ErrorHandlers
+	runtime.ErrorHandlers = append(runtime.ErrorHandlers, func(err error) {
+		if err.Error() == "lost connection to pod" {
+			// Stop the low-level port forwarding
+			// the infinite loop will restart it
+			if o.stopChan == nil {
+				return
+			}
+			o.stopChan <- struct{}{}
+			o.stopChan = make(chan struct{}, 1)
+		}
+	})
+
+	o.isRunning = true
 	go func() {
-		err = o.kubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, errOut, o.stopChan)
-		if err != nil {
-			fmt.Printf("failed to setup port-forwarding: %v\n", err)
+		backo := watch.NewExpBackoff()
+		for {
+			o.finishedChan = make(chan struct{}, 1)
+			err = o.kubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, errOut, o.stopChan)
+			if err != nil {
+				fmt.Printf("failed to setup port-forwarding: %v\n", err)
+				time.Sleep(backo.Delay())
+			}
+			if !o.isRunning {
+				break
+			}
 		}
 		o.finishedChan <- struct{}{}
 	}()
@@ -97,10 +127,18 @@ func (o *PFClient) StopPortForwarding() {
 	if o.stopChan == nil {
 		return
 	}
+	// Ask the low-level port forward to stop
 	o.stopChan <- struct{}{}
 	o.stopChan = nil
+
+	// Ask the inifinte loop to stop
+	o.isRunning = false
+
+	// Wait for low level port forward to be finished
+	// and the infinite loop to exit
 	<-o.finishedChan
 	o.finishedChan = nil
+	runtime.ErrorHandlers = o.originalErrorHandlers
 }
 
 func (o *PFClient) GetForwardedPorts() map[string][]int {
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -1,6 +1,7 @@
 package portForward
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"reflect"
@@ -117,7 +118,10 @@ func (o *PFClient) StartPortForwarding(
 			err = o.kubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, errOut, o.stopChan)
 			if err != nil {
 				fmt.Fprintf(errOut, "Failed to setup port-forwarding: %v\n", err)
-				time.Sleep(backo.Delay())
+				d := backo.Delay()
+				time.Sleep(d)
+			} else {
+				backo.Reset()
 			}
 			if !o.isRunning {
 				break
@@ -127,8 +131,13 @@ func (o *PFClient) StartPortForwarding(
 	}()
 
 	// Wait the first time the devstate file is written
-	err = <-devstateChan
-	return err
+	timeout := 1 * time.Minute
+	select {
+	case err = <-devstateChan:
+		return err
+	case <-time.After(timeout):
+		return errors.New("unable to setup port forwarding")
+	}
 }
 
 func (o *PFClient) StopPortForwarding() {
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -82,8 +82,6 @@ func (o *PFClient) StartPortForwarding(
 		return err
 	}
 
-	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
-
 	o.originalErrorHandlers = runtime.ErrorHandlers
 	runtime.ErrorHandlers = append(runtime.ErrorHandlers, func(err error) {
 		if err.Error() == "lost connection to pod" {
@@ -98,13 +96,26 @@ func (o *PFClient) StartPortForwarding(
 	})
 
 	o.isRunning = true
+
+	devstateChan := make(chan error)
 	go func() {
 		backo := watch.NewExpBackoff()
 		for {
 			o.finishedChan = make(chan struct{}, 1)
+			portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
+
+			go func() {
+				portsBuf.Wait()
+				err = o.stateClient.SetForwardedPorts(portsBuf.GetForwardedPorts())
+				if err != nil {
+					err = fmt.Errorf("unable to save forwarded ports to state file: %v", err)
+				}
+				devstateChan <- err
+			}()
+
 			err = o.kubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, errOut, o.stopChan)
 			if err != nil {
-				fmt.Printf("failed to setup port-forwarding: %v\n", err)
+				fmt.Fprintf(errOut, "Failed to setup port-forwarding: %v\n", err)
 				time.Sleep(backo.Delay())
 			}
 			if !o.isRunning {
@@ -114,13 +125,9 @@ func (o *PFClient) StartPortForwarding(
 		o.finishedChan <- struct{}{}
 	}()
 
-	portsBuf.Wait()
-	err = o.stateClient.SetForwardedPorts(portsBuf.GetForwardedPorts())
-	if err != nil {
-		return fmt.Errorf("unable to save forwarded ports to state file: %v", err)
-	}
-
-	return nil
+	// Wait the first time the devstate file is written
+	err = <-devstateChan
+	return err
 }
 
 func (o *PFClient) StopPortForwarding() {
--- a/tests/integration/cmd_dev_test.go
+++ b/tests/integration/cmd_dev_test.go
@@ -398,7 +398,7 @@ var _ = Describe("odo dev command tests", func() {
 				helper.ReplaceString(filepath.Join(commonVar.Context, "devfile.yaml"), "npm start", "sleep 20 ; npm start")
 
 				var err error
-				devSession, _, _, ports, err = helper.StartDevMode()
+				devSession, _, _, ports, err = helper.StartDevMode(nil)
 				Expect(err).ToNot(HaveOccurred())
 			})
 
--- a/tests/integration/cmd_dev_test.go
+++ b/tests/integration/cmd_dev_test.go
@@ -388,6 +388,47 @@ var _ = Describe("odo dev command tests", func() {
 				})
 			})
 		})
+
+		When("a delay is necessary for the component to start and running odo dev", func() {
+
+			var devSession helper.DevSession
+			var ports map[string]string
+
+			BeforeEach(func() {
+				helper.ReplaceString(filepath.Join(commonVar.Context, "devfile.yaml"), "npm start", "sleep 20 ; npm start")
+
+				var err error
+				devSession, _, _, ports, err = helper.StartDevMode()
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			AfterEach(func() {
+				devSession.Kill()
+				devSession.WaitEnd()
+			})
+
+			It("should first fail then succeed querying endpoint", func() {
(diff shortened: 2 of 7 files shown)


dharmit:
I was having a conversation about this with Tomas on Friday. The way I patched the PR #5933 was by adding one more check to the `if` condition:
```diff
diff --git a/pkg/portForward/portForward.go b/pkg/portForward/portForward.go
index 7101fc2fb..fc900d573 100644
--- a/pkg/portForward/portForward.go
+++ b/pkg/portForward/portForward.go
@@ -49,7 +49,7 @@ func (o *PFClient) StartPortForwarding(
                return err
        }
 
-       if o.stopChan != nil && reflect.DeepEqual(ceMapping, o.appliedEndpoints) {
+       if o.stopChan != nil && o.finishedChan == nil && reflect.DeepEqual(ceMapping, o.appliedEndpoints) {
                return nil
        }
```
The problem I was then trying to solve was user doing `curl` while `odo dev` is syncing a change made by the user to any of the source files. Not the first `odo dev`, but a subsequent run of it. I figured that if the user `curl`'s while `odo dev` is still syncing, the goroutine that's handling port-forwarding is exited. And since the `stopChan` is not `nil` and endpoints haven't changed, the `if` condition would result in `return`'ing before it could start port-forwarding again.

feloy:
> I was having a conversation about this with Tomas on Friday. The way I patched the PR #5933 was by adding one more check to the `if` condition:
> 
> ```diff
> diff --git a/pkg/portForward/portForward.go b/pkg/portForward/portForward.go
> index 7101fc2fb..fc900d573 100644
> --- a/pkg/portForward/portForward.go
> +++ b/pkg/portForward/portForward.go
> @@ -49,7 +49,7 @@ func (o *PFClient) StartPortForwarding(
>                 return err
>         }
>  
> -       if o.stopChan != nil && reflect.DeepEqual(ceMapping, o.appliedEndpoints) {
> +       if o.stopChan != nil && o.finishedChan == nil && reflect.DeepEqual(ceMapping, o.appliedEndpoints) {
>                 return nil
>         }
> ```
> 
> The problem I was then trying to solve was user doing `curl` while `odo dev` is syncing a change made by the user to any of the source files. Not the first `odo dev`, but a subsequent run of it. I figured that if the user `curl`'s while `odo dev` is still syncing, the goroutine that's handling port-forwarding is exited. And since the `stopChan` is not `nil` and endpoints haven't changed, the `if` condition would result in `return`'ing before it could start port-forwarding again.

I don't think this change is necessary with the changes I have made on my side.

pkg/portForward/portForward.go
```diff
@@ -75,32 +82,70 @@ func (o *PFClient) StartPortForwarding(
...
+	o.originalErrorHandlers = runtime.ErrorHandlers
+	runtime.ErrorHandlers = append(runtime.ErrorHandlers, func(err error) {
+		if err.Error() == "lost connection to pod" {
+			// Stop the low-level port forwarding
+			// the infinite loop will restart it
+			if o.stopChan == nil {
+				return
+			}
+			o.stopChan <- struct{}{}
+			o.stopChan = make(chan struct{}, 1)
```
dharmit, line 94: Why are we doing this? I understand the part about `o.stopChan <- struct{}{}` is to stop the port-forwarding started using client-go library. But why are we doing `o.stopChan = make(chan struct{}, 1)` after that?

pkg/portForward/portForward.go
```diff
@@ -58,7 +66,6 @@ func (o *PFClient) StartPortForwarding(
 	o.StopPortForwarding()
 
 	o.stopChan = make(chan struct{}, 1)
```
dharmit, line 68: Why is this a buffered channel?

tests/integration/cmd_dev_test.go
```diff
@@ -386,6 +386,47 @@ var _ = Describe("odo dev command tests", func() {
...
 			})
 		})
+
+		When("a delay is necessary for the component to start and running odo dev", func() {
+
+			var devSession helper.DevSession
+			var ports map[string]string
+
+			BeforeEach(func() {
+				helper.ReplaceString(filepath.Join(commonVar.Context, "devfile.yaml"), "npm start", "sleep 20 ; npm start")
```
dharmit, line 396: Would 10s suffice? Asking because this could set a precedent and subsequent tests like this would usually reuse `sleep 20`, thus slowing down the tests.

tests/integration/cmd_dev_test.go
```diff
@@ -386,6 +386,47 @@ var _ = Describe("odo dev command tests", func() {
...
+			})
+
+			It("should first fail then succeed querying endpoint", func() {
+				url := fmt.Sprintf("http://%s", ports["3000"])
+				_, err := http.Get(url)
+				Expect(err).To(HaveOccurred())
+
+				podName := commonVar.CliRunner.GetRunningPodNameByComponent(cmpName, commonVar.Project)
+				Eventually(func() bool {
+					logs := helper.GetCliRunner().GetLogs(podName)
```
dharmit, line 415: Can we use `odo logs` here instead of `kubectl`/`oc`?

dharmit (commented):
Mostly questions where I didn't understand why we are doing what we are doing.

A BIG THANK YOU for adding comments. It made reading code between the goroutines and `StopPortForwarding` easier.

pkg/portForward/portForward.go
```diff
@@ -75,32 +82,70 @@ func (o *PFClient) StartPortForwarding(
 		return err
 	}
 
-	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
+	o.originalErrorHandlers = runtime.ErrorHandlers
+	runtime.ErrorHandlers = append(runtime.ErrorHandlers, func(err error) {
```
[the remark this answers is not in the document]
feloy: runtime.ErrorHandlers is a generic error handler defined in the API Machinery library (https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go).

It is used by the client-go library, specifically in the `forward` function, through the call to runtime.HandleError:

```
// forward dials the remote host specific in req, upgrades the request, starts
// listeners for each port specified in ports, and forwards local connections
// to the remote host via streams.
func (pf *PortForwarder) forward() error {
[...]
	// wait for interrupt or conn closure
	select {
	case <-pf.stopChan:
	case <-pf.streamConn.CloseChan():
		runtime.HandleError(errors.New("lost connection to pod"))
	}

	return nil
}
```

By default, the ErrorHandlers contains two handlers: 1 for logging the error using klog, and 1 for backing off the errors:

```
var ErrorHandlers = []func(error){
	logError,
	(&rudimentaryErrorBackoff{
		lastErrorTime: time.Now(),
		// 1ms was the number folks were able to stomach as a global rate limit.
		// If you need to log errors more than 1000 times a second you
		// should probably consider fixing your code instead. :)
		minPeriod: time.Millisecond,
	}).OnError,
}
```

Here, we are adding a new handler to the list, so our handler can be called when errors occur, and specifically when the "list connection" error happens. 

We need to reset the handlers to their original value when we are stopping the port forwarding, or the same handler will be added twice, 3x, 4x, etc when we are restartting the port forward again after a failure.

pkg/portForward/portForward.go
feloy: I prefer to start with a fresh new channel. The old one will be cleaned by the GC.

tests/integration/cmd_dev_test.go
feloy: I tried with 15s and it failed in IBM cloud tests from time to time.

tests/integration/cmd_dev_test.go
feloy: No, we discussed to not rely on other odo commands during integration tests. This way, odo logs can be buggy, or be removed, and this won't affect this test

pkg/portForward/portForward.go
feloy: Using a non-buffered one, writing into this channel will block until the someone reads the channel.

Usin a buffered one, the first write into the channel won't block, even if no one reads it.

This way, we can write on the channel and continue, even if the `forward` function is not yet started and so wion't read the channel. 

That's also why I'm recreating a fresh new channel after, so if the channel has not been read, we won't be blocked during the second write.

pkg/portForward/portForward.go
dharmit: OK. But what's the rationale behind it? If it's about unblocking, as you mentioned in another comment, I think we are doing some kind of "trick".

pkg/portForward/portForward.go
feloy: But of which size? If you create it of size `n`, it will happen once in a while that it will be written `n+1` times.

Particularly, I tested the changes by flooding the app with the following command, while modifying source code so the pod restarts:

```
while : ; do curl localhost:40001; done
```

With a buffered channel of size 100, I'm pretty sure I would have filled the buffer

pkg/portForward/portForward.go
feloy: It's about restarting with fresh new state, because we are restarting a new port forwarding, and don't care about the state of the previous one.

feloy:
Windows IBM cloud tests:

```
odo dev command tests
/go/odo_1/tests/integration/cmd_dev_test.go:29
  devfile contains composite apply command
  /go/odo_1/tests/integration/cmd_dev_test.go:1197
    when odo dev is running
    /go/odo_1/tests/integration/cmd_dev_test.go:1209
      [It] should execute the composite apply commands successfully
      /go/odo_1/tests/integration/cmd_dev_test.go:1219
```

feloy:
> Openshift IBM Cloud tests:
> 
> ```
> odo dev command tests
> /go/odo_1/tests/integration/cmd_dev_test.go:29
>   devfile contains composite apply command
>   /go/odo_1/tests/integration/cmd_dev_test.go:1197
>     when odo dev is running
>     /go/odo_1/tests/integration/cmd_dev_test.go:1209
>       [It] should execute the composite apply commands successfully
>       /go/odo_1/tests/integration/cmd_dev_test.go:1219
> 
>     [odo] Error: Cannot find module 'prom-client'
>     [odo] Require stack:
>     [odo] - /projects/server.js
>     [odo]     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
>     [odo]     at Function.Module._load (internal/modules/cjs/loader.js:746:27)
>     [odo]     at Module.require (internal/modules/cjs/loader.js:974:19)
>     [odo]     at require (internal/modules/cjs/helpers.js:93:18)
>     [odo]     at Object.<anonymous> (/projects/server.js:1:20)
>     [odo]     at Module._compile (internal/modules/cjs/loader.js:1085:14)
>     [odo]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
>     [odo]     at Module.load (internal/modules/cjs/loader.js:950:32)
>     [odo]     at Function.Module._load (internal/modules/cjs/loader.js:790:12)
>     [odo]     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12) {
>     [odo]   code: 'MODULE_NOT_FOUND',
>     [odo]   requireStack: [ '/projects/server.js' ]
>     [odo] }
> ```

@valaparthvi
I'm seeing this error regularly in Openshift IBM Cloud tests (in this PR but also other PRs). I've made this commit to try and fix it: https://github.com/redhat-developer/odo/pull/6013/commits/4d2396826540f7a8b892e262a9b1056985e298b4

WDYT?

feloy:
Restart tests to check flakiness of fixed test

pkg/portForward/portForward.go
```diff
@@ -75,32 +82,70 @@ func (o *PFClient) StartPortForwarding(
...
+				if err != nil {
+					err = fmt.Errorf("unable to save forwarded ports to state file: %v", err)
+				}
+				devstateChan <- err
+			}()
+
+			err = o.kubernetesClient.SetupPortForwarding(pod, portPairsSlice, portsBuf, errOut, o.stopChan)
+			if err != nil {
+				fmt.Fprintf(errOut, "Failed to setup port-forwarding: %v\n", err)
+				time.Sleep(backo.Delay())
```
rm3l, line 119: In my understanding, in case of an error, we are now retrying with an exponential backoff strategy (capped at 10s by default), which is nice. But would it make sense to give up after a reasonable number of retries rather than keep retrying?
Because I noticed that it is no longer possible to stop the `odo dev` process as long we enter this condition. I simulated this by simulating a Dial error that might be returned by the `kclient.ClientInterface#SetupPortForwarding` implementation in `port_forwarding.go`.

rm3l (commented):
Added a few comments.

You already mentioned that, but the more I used the `--random-ports` flag, the more I found it confusing that users get assigned a new random port when the port forwarding is restarted (in the same Dev session). If we could store the previous ports in memory and try to reuse them first, I think this would improve the user experience. But this can be addressed in a separate issue.

LGTM overall otherwise.

pkg/portForward/portForward.go
rm3l: For channels sizing, see this recommendation (size of 0 or 1) from the coding conventions doc: https://github.com/redhat-developer/odo/wiki/Dev:-Coding-Conventions#channel-size-is-one-or-none

pkg/portForward/portForward.go
```diff
@@ -75,32 +82,70 @@ func (o *PFClient) StartPortForwarding(
 		return err
 	}
 
-	portsBuf := NewPortWriter(log.GetStdout(), len(portPairsSlice), ceMapping)
+	o.originalErrorHandlers = runtime.ErrorHandlers
```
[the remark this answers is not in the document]
feloy: My bad, saving the pointer to the slice does not help, effectively. Thanks

pkg/portForward/portForward.go
feloy: I was also wondering. I added some timeout while waiting for the devstate to be written

feloy:
> You already mentioned that, but the more I used the `--random-ports` flag, the more I found it confusing that users get assigned a new random port when the port forwarding is restarted (in the same Dev session). If we could store the previous ports in memory and try to reuse them first, I think this would improve the user experience. But this can be addressed in a separate issue.
> 

I was also inclined to do this, but this would break the integration tests, as we are not guaranteed that the port is not taken by another process during the restart of the port forwarding.

pkg/portForward/portForward.go
feloy: This gives a different behaviour for the first port forwarding and the next restarts: the first one will be timed out, the following ones won't.

But only the first one is blocking (we are waiting for the first port forwarding to be effective before to return the function); the other ones are done in the go routine after the function returned.

rm3l: > This gives a different behaviour for the first port forwarding and the next restarts: the first one will be timed out, the following ones won't.
> 
> But only the first one is blocking (we are waiting for the first port forwarding to be effective before to return the function); the other ones are done in the go routine after the function returned.

This is fine, I think. Once the dev session is running (and ports are forwarded), it is understandable to me that we try to recover from potential errors happening. Thanks.

rm3l:
> > You already mentioned that, but the more I used the `--random-ports` flag, the more I found it confusing that users get assigned a new random port when the port forwarding is restarted (in the same Dev session). If we could store the previous ports in memory and try to reuse them first, I think this would improve the user experience. But this can be addressed in a separate issue.
> 
> I was also inclined to do this, but this would break the integration tests, as we are not guaranteed that the port is not taken by another process during the restart of the port forwarding.

I see. I think we can find a trade-off by letting users decide (via a yet-another flag) which behavior to adopt w.r.t random ports. Default behavior could be to try to reuse the ports, but in integration tests, it would be acceptable to have new ports assigned.

rm3l (commented):
One last comment - do you think unit-testing the methods in `portForward.go` could be do-able? I understand it might not be that easy to do, especially with the different goroutines we have and the global `runtime.ErrorHandlers` variable we use..

feloy:
> I see. I think we can find a trade-off by letting users decide (via a yet-another flag for example) which behavior to adopt w.r.t random ports. Default behavior could be to try to reuse the ports, but in integration tests, it would be acceptable to have new ports assigned.

I would prefer to wait and see if people are using this `random-ports` flag. It has been made for integration tests, I'm not sure there is a real use case for developers. And I would not like to add complexity to this part of the code which is already quite imbricated, if not necessary.

feloy:
> One last comment - do you think unit-testing the methods in `portForward.go` could be do-able? I understand it might not be that easy to do, especially with the different goroutines we have and the global `runtime.ErrorHandlers` variable we use..

That's an interesting challenge. I would like to have a try. Either in this PR or another one

rm3l (approved):
Thanks for your work on this.

dharmit:
> > I see. I think we can find a trade-off by letting users decide (via a yet-another flag for example) which behavior to adopt w.r.t random ports. Default behavior could be to try to reuse the ports, but in integration tests, it would be acceptable to have new ports assigned.
> 
> I would prefer to wait and see if people are using this `random-ports` flag. It has been made for integration tests, I'm not sure there is a real use case for developers. And I would not like to add complexity to this part of the code which is already quite imbricated, if not necessary.

I would rather not expose that port to the users at all. Opinionated. 😉

rm3l:
> > > I see. I think we can find a trade-off by letting users decide (via a yet-another flag for example) which behavior to adopt w.r.t random ports. Default behavior could be to try to reuse the ports, but in integration tests, it would be acceptable to have new ports assigned.
> > 
> > 
> > I would prefer to wait and see if people are using this `random-ports` flag. It has been made for integration tests, I'm not sure there is a real use case for developers. And I would not like to add complexity to this part of the code which is already quite imbricated, if not necessary.
> 
> I would rather not expose that ~port~ flag to the users at all. Opinionated. wink

Indeed. If possible to hide it (mostly from the user-facing Help output), it would make more sense not to expose it 👍🏿

PR #6015, opened 2022-08-09
Make odo work if optional `metadata.name` field is missing in Devfile

### LINKED ISSUE #5821: odo should not require metadata.name in devfile
## User Story
- As an odo user I have created a devfile without `metadata` section (devfile spec defines that whole `metadata` section as optional)
- I want to be able to use such devfile with odo
- This makes sense as if I have devfile inside my project repo metadata is not much use for me

## Acceptance Criteria
- [ ] if there is no `metadata.name` defined in devfile, use alizer to evaluate the component name
- [ ] do all necessary sanitation to directory name before using it as a devfile name (length, allowed characters etc...)

/kind user-story
## example

```yaml
commands:
- exec:
    commandLine: npm install
    component: runtime
    group:
      isDefault: true
      kind: build
    workingDir: ${PROJECT_SOURCE}
  id: install
- exec:
    commandLine: npm start
    component: runtime
    group:
      isDefault: true
      kind: run
    workingDir: ${PROJECT_SOURCE}
  id: run
- exec:
    commandLine: npm run debug
    component: runtime
    group:
      isDefault: true
      kind: debug
    workingDir: ${PROJECT_SOURCE}
  id: debug
- exec:
    commandLine: npm test
    component: runtime
    group:
      isDefault: true
      kind: test
    workingDir: ${PROJECT_SOURCE}
  id: test
components:
- container:
    endpoints:
    - name: http-3000
      targetPort: 3000
    memoryLimit: 1024Mi
    mountSources: true
  name: runtime
schemaVersion: 2.0.0

```

current behavior
```
 odo dev   
  __
 /  \__     Developing using the  Devfile
 \__/  \    Namespace: asdf
 \__/

 ✗  Waiting for Kubernetes resources 
failed to delete inner loop resources: failed to get the resource "Deployment" name for component ""; cause: namespacing: component name cannot be blank ✗  failed to create the component: component name "" is not valid, component name should conform the following requirements: 
- Contain at most 63 characters
- Contain only lowercase alphanumeric characters or ‘-’
- Start with an alphanumeric character
- End with an alphanumeric character
- Must not contain all numeric values
```

rm3l:
**What type of PR is this:**
/kind feature

Fixes #5821

**PR acceptance criteria:**

- [x] Unit test

- [x] Integration test

- [ ] Documentation

### DIFF
--- a/pkg/binding/add.go
+++ b/pkg/binding/add.go
@@ -100,6 +100,7 @@ func (o *BindingClient) AskNamingStrategy(flags map[string]string) (string, erro
 }
 
 func (o *BindingClient) AddBindingToDevfile(
+	componentName string,
 	bindingName string,
 	bindAsFiles bool,
 	serviceNs string,
@@ -112,7 +113,7 @@ func (o *BindingClient) AddBindingToDevfile(
 		return obj, err
 	}
 
-	deploymentName := fmt.Sprintf("%s-app", obj.GetMetadataName())
+	deploymentName := fmt.Sprintf("%s-app", componentName)
 	deploymentGVK, err := o.kubernetesClient.GetDeploymentAPIVersion()
 	if err != nil {
 		return obj, err
--- a/pkg/binding/binding_test.go
+++ b/pkg/binding/binding_test.go
@@ -395,7 +395,15 @@ func TestBindingClient_AddBindingToDevfile(t *testing.T) {
 			o := &BindingClient{
 				kubernetesClient: tt.fields.kubernetesClient(ctrl),
 			}
-			got, err := o.AddBindingToDevfile(tt.args.bindingName, tt.args.bindAsFiles, tt.args.namespace, tt.args.namingStrategy, tt.args.unstructuredService, tt.args.obj)
+			got, err := o.AddBindingToDevfile(
+				tt.args.obj.GetMetadataName(),
+				tt.args.bindingName,
+				tt.args.bindAsFiles,
+				tt.args.namespace,
+				tt.args.namingStrategy,
+				tt.args.unstructuredService,
+				tt.args.obj,
+			)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("AddBindingToDevfile() error = %v, wantErr %v", err, tt.wantErr)
 				return
--- a/pkg/binding/interface.go
+++ b/pkg/binding/interface.go
@@ -39,6 +39,7 @@ type Client interface {
 	AskNamingStrategy(flags map[string]string) (string, error)
 	// AddBindingToDevfile adds the ServiceBinding manifest to the devfile
 	AddBindingToDevfile(
+		componentName string,
 		bindingName string,
 		bindAsFiles bool,
 		serviceNs string,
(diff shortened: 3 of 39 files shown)


rm3l:
/hold

Builds upon 5989, so waiting for it to be merged in first.

rm3l:
```
The devfile "a-devfile-name" from the registry "a-registry" will be downloaded.
Language: 
Project type: 
The devfile "" from the registry "" will be downloaded.
--- FAIL: TestInteractiveBackend_PersonalizeName (0.22s)
    --- FAIL: TestInteractiveBackend_PersonalizeName/no_flag (0.22s)
        controller.go:137: missing call(s) to *asker.MockAsker.AskName(is anything) /go/src/github.com/redhat-developer/odo/pkg/init/backend/interactive_test.go:223
        controller.go:137: aborting test due to missing call(s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x258d712]
```

/test unit-and-validate-test

pkg/binding/add.go
```diff
@@ -112,7 +113,12 @@ func (o *BindingClient) AddBindingToDevfile(
 		return obj, err
 	}
 
-	deploymentName := fmt.Sprintf("%s-app", obj.GetMetadataName())
+	componentName, err := component.GatherName(obj)
```
feloy, line 116: I would prefer this function to be called during the context retrieval on genericclioptions.New(), and the component name stored in the context.

rm3l: Indeed, makes sense from the CLI perspective. I'll try to call this during the context retrieval. I see there is already a notion of component name in such context. However, I guess this would require passing this information down to the other functions/methods that might need it (instead of recomputing it from the Devfile object)..

feloy: Yes, the value will need to be passed to business packages.

At some point, I would like the context to be built at the very beginning, and passed in the business client structures alongside the dependent clients, so we don't have to bother passing these values all around.

rm3l: I've changed this so as to call this function during context retrieval.
I'm leveraging the already existing `component` field in the Context (which, per my understanding, relied on the `--component` flag, but I am not sure this flag is still being used).
Please take a look at [718a6ad975b58f03f6b2fe5a99797a59e2975539](https://github.com/redhat-developer/odo/pull/6015/commits/718a6ad975b58f03f6b2fe5a99797a59e2975539) ([3f7ef4d58fd19805f033b34744caabdb58ba2e4f](https://github.com/redhat-developer/odo/pull/6015/commits/3f7ef4d58fd19805f033b34744caabdb58ba2e4f) amends that commit, following an issue I ran into when testing `odo list`).

rm3l: The CI failures should be fixed now. I had to rebase onto `main` to fix a few conflicts.

feloy (approved):

feloy:
/override ci/prow/unit-and-validate-test
Tests pass on IBM Cloud

PR #6020, opened 2022-08-11
Watch for devfile dependencies

### LINKED ISSUE #6017: odo dev does not react on devfile parent change
/kind bug

The devfile references a parent devfile, and the devfile.yaml and parent.yaml are added to the gitignore file, so these files are not synced to the container.

```
// parent.yaml
commands:
- exec:
    commandLine: GOCACHE=${PROJECT_SOURCE}/.cache go build main.go
    component: runtime
    group:
      isDefault: true
      kind: build
    workingDir: ${PROJECT_SOURCE}
  id: build
- exec:
    commandLine: ./main
    component: runtime
    group:
      isDefault: true
      kind: run
    workingDir: ${PROJECT_SOURCE}
  id: run
components:
- container:
    endpoints:
    - name: http
      targetPort: 8080
    memoryLimit: 1024Mi
    mountSources: true
  name: runtime
metadata:
  description: Stack with the latest Go version
  displayName: Go Runtime
  language: go
  name: my-go-app
  projectType: go
  tags:
  - Go
  version: 1.0.0
schemaVersion: 2.1.0
starterProjects:
- git:
    checkoutFrom:
      revision: main
    remotes:
  name: go-starter
```

```
// devfile.yaml
schemaVersion: 2.1.0
metadata:
  name: child
parent:
  uri: "parent.yaml"
```

```
// .gitignore
.odo/env
.odo/odo-file-index.json
.odo/devstate.json
main

/devfile.yaml
/parent.yaml
```

## Actual behavior

When the parent.yaml file is modified, odo dev does not react to the change.

## Expected behavior

odo dev should update the component with the changes done to the parent devfile.

feloy:
**What type of PR is this:**

/kind bug

Fixes #6017

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/libdevfile/uris.go
+++ b/pkg/libdevfile/uris.go
@@ -1,0 +1,123 @@
+package libdevfile
+
+import (
+	"errors"
+	"net/url"
+	"sort"
+	"strings"
+
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/api/v2/pkg/attributes"
+	"github.com/devfile/api/v2/pkg/validation"
+	"github.com/devfile/library/pkg/devfile/parser"
+	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
+)
+
+const _importSourceAttributeUriPrefix = "uri: "
+
+// GetReferencedLocalFiles returns the local files referenced by the Devfile. This includes:
+// - the non-inlined Kubernetes and Openshift components
+// - the Dockerfiles of Image components
+// - the parent devfile
+// - resursively, the local files referenced by the parent Devfile
+// The passed Devfile must be flattened
+func GetReferencedLocalFiles(devfileObj parser.DevfileObj) (result []string, err error) {
+
+	setResult := map[string]struct{}{}
+
+	parent := devfileObj.Data.GetParent()
+	if parent != nil {
+		return nil, errors.New("devfile must be flattened")
+	}
+
+	components, err := devfileObj.Data.GetComponents(common.DevfileOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, component := range components {
+		var componentType v1alpha2.ComponentType
+		componentType, err = common.GetComponentType(component)
+		if err != nil {
+			return nil, err
+		}
+
+		switch componentType {
+		case v1alpha2.KubernetesComponentType:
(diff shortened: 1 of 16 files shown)


pkg/libdevfile/uris.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// appendUriIfFile appends uri to the result if the uri is a local path
+func appendUriIfFile(result map[string]struct{}, uri string) (map[string]struct{}, error) {
+	if uri != "" {
+		u, err := url.Parse(uri)
+		if err != nil {
+			return nil, err
+		}
+		if u.Scheme == "" {
```
rm3l, line 99: Just curious - is the `file` URI scheme something possible here to reference local files? Or should we always expect files relative to the devfile.yaml? The Devfile spec is pretty vague about this:
> Location in a file fetched from a uri.

pkg/watch/watch.go
```diff
@@ -242,16 +242,23 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters, ct
...
+		var devfileFiles []string
+		devfileFiles, err = libdevfile.GetReferencedLocalFiles(parameters.InitialDevfileObj)
 		if err != nil {
 			return err
 		}
+		devfileFiles = append(devfileFiles, parameters.DevfilePath)
+		for _, f := range devfileFiles {
+			err = devfileWatcher.Add(f)
+			if err != nil {
+				klog.V(4).Infof("error adding watcher for path %s: %v", f, err)
```
rm3l, line 259: Shouldn't we return the error right away instead? Or maybe make the error more visible to the user by logging it as a warning?

rm3l (changes requested):

pkg/libdevfile/uris.go
feloy: I'm pretty sure it is related to the directory containing the devfile. For example, the doc for parent URI is:

```
URI Reference of a parent devfile YAML file. It can be a full URL or a relative URI with the current devfile as the base URI.
```

pkg/watch/watch.go
feloy: I first started with returning the error, but lots of integration tests fail, because they are using Devfiles containing URIs but withtout the corresponding file. 

I would prefer the error to be handled by the devfile validation. I consider that if the devfile validation does not complain, we should not here.

rm3l (approved):

PR #6025, opened 2022-08-16
Do not get project from env.yaml file

### LINKED ISSUE #6018: `odo dev` New way of storing `namespace`/`projects` information
/kind user-story

## User Story
As an odo user, I expect that `odo dev` will deploy my application to the current namespace.

current situation:
`odo dev` deploys the application to the current namespace only the first time. Then the namespace is saved into the `env.yaml` and the next time you run `odo dev`, even if the namespace was changed, the resources are created in the original namespace.

## Acceptance Criteria 
- [ ] every time `odo dev` is started, it should always use the current namespace/project as set in kubeconfig
- [ ] `odo dev` should remove "project" information from the `env.yaml` when it ends (as part of the cleanup procedure)

/kind user-story

### LINKED ISSUE #6027: `odo deploy` doesn't immediately check for namespace existence
/kind bug

**Operating System:** all

* Fresh minikube cluster
* The namespace `myproject`, which was mentioned in `env.yaml`, didn't exist on the cluster
* `odo deploy` for the [devfile with outer loop components](https://github.com/redhat-developer/odo/blob/main/tests/examples/source/devfiles/nodejs/devfile-deploy-functional-pods.yaml) and [devfile-nodejs-deploy](https://github.com/kadel/devfile-nodejs-deploy) example project

## Actual behavior
`odo deploy` builds the image, but fails when creating Kubernetes components

## Expected behavior
Should odo spend time building the image if the namespace doesn't exist? 

```sh
$ odo deploy
  __
 /  \__     Deploying the application using devfile-nodejs-deploy Devfile
 \__/  \    Namespace: myproject
 /  \__/    odo version: v3.0.0-beta3
 \__/

↪ Building & Pushing Container: quay.io/tkral/devfile-nodejs-deploy:latest
 •  Building image locally  ...
build -t quay.io/tkral/devfile-nodejs-deploy:latest -f /home/dshah/src/devfile-nodejs-deploy/Dockerfile /home/dshah/src/devfile-nodejs-deploy
 ✓  Building image locally [2ms]
 •  Pushing image to container registry  ...
push quay.io/tkral/devfile-nodejs-deploy:latest
 ✓  Pushing image to container registry [1ms]

↪ Deploying Kubernetes Component: devfile-nodejs-deploy
 ✗  failed to create service(s) associated with the component: namespaces "myproject" not found

```

feloy:
**What type of PR is this:**

/kind feature

- set current namespace in env.yaml file when starting `odo dev`, remove it when cleanup
- do not read project from env.yaml file when running `odo dev` (or any other command)
- env.yaml file is not created when running `odo deploy`
- the existence of env.yaml file is not checked when running `odo delete component`, `odo build-images` (or any other command)

Fixes #6018
Fixes #6027

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/docs/website/docs/command-reference/dev.md
+++ b/docs/website/docs/command-reference/dev.md
@@ -417,13 +417,9 @@ This state file contains the forwarded ports:
 
 ## Env File
 
-When `odo dev` is executed, a `.odo/env/env.yaml` file is created if it does not exist. It stores the namespace/project that will be used by the component, and its default value is set to currently active namespace/project.
+When `odo dev` is executed, a `.odo/env/env.yaml` file is created if it does not exist. It stores the namespace/project that will be used by the component during the dev session. The value is removed from the file at the end of the session.
 
 ```yaml
 ComponentSettings:
   Project: myproject
 ```
-
-:::note
-Creating, deleting or setting a namespace/project will not modify this value. To use a different namespace/project, user will have to manually change `.ComponentSettings.Project` value, or delete the `.odo` directory.
-:::
--- a/pkg/odo/genericclioptions/errors.go
+++ b/pkg/odo/genericclioptions/errors.go
@@ -1,11 +1,39 @@
 package genericclioptions
 
+import (
+	"fmt"
+
+	"github.com/redhat-developer/odo/pkg/devfile/location"
+	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
+)
+
 var _ error = NoDevfileError{}
 
-type NoDevfileError struct{}
+type NoDevfileError struct {
+	context string
+}
+
+func NewNoDevfileError(context string) NoDevfileError {
+	return NoDevfileError{
+		context: context,
+	}
+}
 
 func (o NoDevfileError) Error() string {
-	return "no devfile found"
+	message := `The current directory does not represent an odo component. 
+To get started:%s
+  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
+Visit https://odo.dev for more information.`
+
+	if isEmpty, _ := location.DirIsEmpty(filesystem.DefaultFs{}, o.context); isEmpty {
+		message = fmt.Sprintf(message, `
+  * Use "odo init" to initialize an odo component in the folder.
+  * Use "odo dev" to deploy it on cluster.`)
+	} else {
+		message = fmt.Sprintf(message, `
+  * Use "odo dev" to initialize an odo component for this folder and deploy it on cluster.`)
+	}
+	return message
 }
 
 func IsNoDevfileError(err error) bool {
--- a/pkg/odo/genericclioptions/localprovider.go
+++ b/pkg/odo/genericclioptions/localprovider.go
@@ -1,13 +1,8 @@
 package genericclioptions
 
 import (
-	"errors"
-	"fmt"
-
-	"github.com/redhat-developer/odo/pkg/devfile/location"
 	"github.com/redhat-developer/odo/pkg/envinfo"
 	"github.com/redhat-developer/odo/pkg/odo/cmdline"
-	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
 )
 
 // GetValidEnvInfo accesses the environment file
@@ -22,39 +17,6 @@ func GetValidEnvInfo(cmdline cmdline.Cmdline) (*envinfo.EnvSpecificInfo, error) 
 	envInfo, err := envinfo.NewEnvSpecificInfo(componentContext)
 	if err != nil {
 		return nil, err
-	}
-
-	// Now we check to see if we can skip gathering the information.
-	// Return if we can skip gathering configuration information
-	configIsNeeded, err := cmdline.CheckIfConfigurationNeeded()
-	if err != nil {
-		return nil, err
-	}
-	if !configIsNeeded {
-		return envInfo, nil
-	}
-
-	// Check to see if the environment file exists
-	if !envInfo.Exists() {
-		exitMessage := `The current directory does not represent an odo component. 
-
-To get started,%s
-  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
-
-Visit https://odo.dev for more information.`
-
-		if isEmpty, _ := location.DirIsEmpty(filesystem.DefaultFs{}, componentContext); isEmpty {
-			exitMessage = fmt.Sprintf(exitMessage, `
-  * Create and move to a new directory
-  * Use "odo init" to initialize an odo component in the folder.
-  * Use "odo dev" to deploy it on cluster.`)
-		} else {
-			exitMessage = fmt.Sprintf(exitMessage, `
-  * Use "odo dev" to initialize an odo component for this folder and deploy it on cluster.`)
-		}
-		//revive:disable:error-strings This is a top-level error message displayed as is to the end user
-		return nil, errors.New(exitMessage)
-		//revive:enable:error-strings
 	}
 
 	return envInfo, nil
--- a/pkg/odo/genericclioptions/resolve.go
+++ b/pkg/odo/genericclioptions/resolve.go
@@ -27,15 +27,12 @@ func (o *internalCxt) resolveProjectAndNamespace(cmdline cmdline.Cmdline, config
 		}
 		namespace = projectFlag
 	} else {
-		namespace = configProvider.GetNamespace()
-		if namespace == "" {
-			namespace = o.KClient.GetCurrentNamespace()
-			if len(namespace) <= 0 {
-				errFormat := "Could not get current namespace. Please create or set a namespace\n"
-				err := checkProjectCreateOrDeleteOnlyOnInvalidNamespace(cmdline, errFormat)
-				if err != nil {
-					return err
-				}
+		namespace = o.KClient.GetCurrentNamespace()
+		if len(namespace) <= 0 {
+			errFormat := "Could not get current namespace. Please create or set a namespace\n"
+			err := checkProjectCreateOrDeleteOnlyOnInvalidNamespace(cmdline, errFormat)
+			if err != nil {
+				return err
 			}
 		}
 
@@ -91,11 +88,7 @@ func resolveComponent(cmdline cmdline.Cmdline, localConfiguration localConfigPro
 }
 
 func resolveProject(cmdline cmdline.Cmdline, localConfiguration localConfigProvider.LocalConfigProvider) string {
-	projectFlag := cmdline.FlagValueIfSet(util.ProjectFlagName)
-	if projectFlag != "" {
-		return projectFlag
-	}
-	return localConfiguration.GetNamespace()
+	return cmdline.FlagValueIfSet(util.ProjectFlagName)
 }
 
 // checkComponentExistsOrFail checks if the specified component exists with the given context and returns error if not.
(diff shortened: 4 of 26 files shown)


pkg/odo/genericclioptions/localprovider.go
```diff
@@ -24,38 +19,5 @@ func GetValidEnvInfo(cmdline cmdline.Cmdline) (*envinfo.EnvSpecificInfo, error)
...
-	}
-
-	// Check to see if the environment file exists
-	if !envInfo.Exists() {
-		exitMessage := `The current directory does not represent an odo component. 
-
-To get started,%s
-  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
-
-Visit https://odo.dev for more information.`
```
valaparthvi, line 44: With this gone, we solely rely on the existence of devfile.yaml to determine if the directory represents an odo component.
```sh
$ odo delete component                                      
 ✗  no devfile found
```
Can you add a message about next steps when no devfile is present?

pkg/odo/genericclioptions/resolve.go
```diff
@@ -27,15 +27,12 @@ func (o *internalCxt) resolveProjectAndNamespace(cmdline cmdline.Cmdline, config
 		}
 		namespace = projectFlag
 	} else {
-		namespace = configProvider.GetNamespace()
```
valaparthvi, line 30: You can remove configProvider argument from the function signature. Can you also remove the check on L24?
```go
// do not error out when the user is running `odo project`
		if err != nil {
			if cmdline.GetParentName() != "project" {
				return err
			}
		}
```

pkg/odo/genericclioptions/resolve.go
```diff
@@ -27,15 +27,12 @@ func (o *internalCxt) resolveProjectAndNamespace(cmdline cmdline.Cmdline, config
...
-			namespace = o.KClient.GetCurrentNamespace()
-			if len(namespace) <= 0 {
-				errFormat := "Could not get current namespace. Please create or set a namespace\n"
-				err := checkProjectCreateOrDeleteOnlyOnInvalidNamespace(cmdline, errFormat)
-				if err != nil {
-					return err
-				}
+		namespace = o.KClient.GetCurrentNamespace()
+		if len(namespace) <= 0 {
+			errFormat := "Could not get current namespace. Please create or set a namespace\n"
```
valaparthvi, line 32: This might be out of scope for this PR, but we need to streamline namespace vs project throughout all the UX messages. We should use project if it's OCP and namespace if K8s.

pkg/odo/genericclioptions/resolve.go
```diff
@@ -91,11 +88,7 @@ func resolveComponent(cmdline cmdline.Cmdline, localConfiguration localConfigPro
 }
 
 func resolveProject(cmdline cmdline.Cmdline, localConfiguration localConfigProvider.LocalConfigProvider) string {
-	projectFlag := cmdline.FlagValueIfSet(util.ProjectFlagName)
-	if projectFlag != "" {
-		return projectFlag
-	}
-	return localConfiguration.GetNamespace()
+	return cmdline.FlagValueIfSet(util.ProjectFlagName)
```
valaparthvi, line 91: You might as well just call Cmdline.FlagValueIfSet method where you're calling resolveProject. If not, you can consider removing localConfiguration arg from the function signature.

tests/integration/cmd_devfile_deploy_test.go
```diff
@@ -108,34 +103,6 @@ var _ = Describe("odo devfile deploy command tests", func() {
 					session.WaitEnd()
 				})
 
-				When("deleting previous deployment and switching kubeconfig to another namespace", func() {
```
valaparthvi, line 111: This might be an overkill, but I think you can add a test case where:
```go
When "env.yaml containing project is left behind":
    When "current namespace is different from env.yaml project":
        When "odo deploy is run":
            It "should take the current namespace into account"
```

A similar test for `odo delete component`, perhaps.

valaparthvi (changes requested):

pkg/odo/genericclioptions/resolve.go
feloy: I prefer to keep this for a UX-relative PR

pkg/odo/genericclioptions/resolve.go
feloy: I removed the unnecessary param. I prefer to keep the function, to hide the details to the caller

pkg/odo/genericclioptions/errors.go
```diff
@@ -1,11 +1,39 @@
...
+	message := `The current directory does not represent an odo component. 
+To get started:%s
+  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
+Visit https://odo.dev for more information.`
+
+	if isEmpty, _ := location.DirIsEmpty(filesystem.DefaultFs{}, o.context); isEmpty {
+		message = fmt.Sprintf(message, `
+  * Use "odo init" to initialize an odo component in the folder.
+  * Use "odo dev" to deploy it on cluster.`)
+	} else {
```
valaparthvi, line 32: For some reason, this always prints the second message, it does not seem to be detecting empty directory correctly.

<details>
<summary>O/p from odo delete component</summary>

```sh
➜  mkcd /tmp/101       
➜  odo delete component
 ✗  The current directory does not represent an odo component. 
To get started:
  * Use "odo dev" to initialize an odo component for this folder and deploy it on cluster.
  * Open this folder in your favorite IDE and start editing, your changes will be reflected directly on the cluster.
Visit https://odo.dev for more information.

```
</details

feloy: Fixed, thanks

dharmit:
@feloy running `odo dev` from this PR doesn't store the port information in the state file. I think that's a regression, right?

dharmit:
We also need to update the [env file](https://odo.dev/docs/command-reference/dev#env-file) section of the docs to reflect the changes being here, no?

dharmit:
@feloy I think this warning message shouldn't be printed anymore. If I wanted to deploy my component in a particular namespace, this is _the_ way to do it, since the namespace won't be stored in `env.yaml` after this PR is merged. WDYT?
```sh
$ odo set namespace default  
 ⚠  This is being executed inside a component directory. This will not update the namespace of the existing component
```

dharmit:
What I am describing below was purposefully done to break stuff.

I first created an Operator backed service in the namespace `myproject`.
1. `odo dev` in dir1. Component gets pushed to the namespace `myproject`. Keep this running.
2. `cd` to dir2 and set the namespace using `odo set namespace default`. `odo dev` pushes to the namespace `default`. Keep it running or stop it, doesn't matter.
3. `cd` to dir1 and do `odo add binding` (interactive mode) and choose the option "current namespace".

At step 3, odo says it can't find any bindable services in the current namespace. I am not suggesting that this is incorrect/invalid behaviour. I'm simply wondering what should the behaviour be:
1. Should it, as is currently the case, fail to list bindable services since step 2 above changed the kubeconfig to use `default` namespace?
2. Or, should odo smartly figure that since the `env.yaml` file does have a namespace set, it shouldn't use the namespace pointed to by kubeconfig but use the one in `env.yaml` instead?

feloy:
> @feloy running `odo dev` from this PR doesn't store the port information in the state file. I think that's a regression, right?

I tested it but cannot reproduce. Can you share the steps to reproduce?

```

**Comment by dharmit** [comment]
> I tested it but cannot reproduce. Can you share the steps to reproduce?

@feloy my bad. I was looking into `env.yaml` instead of `devstate.json`. me--. 😞

**Comment by feloy** [comment]
> What I am describing below was purposefully done to break stuff.
> 
> I first created an Operator backed service in the namespace `myproject`.
> 
> 1. `odo dev` in dir1. Component gets pushed to the namespace `myproject`. Keep this running.
> 2. `cd` to dir2 and set the namespace using `odo set namespace default`. `odo dev` pushes to the namespace `default`. Keep it running or stop it, doesn't matter.
> 3. `cd` to dir1 and do `odo add binding` (interactive mode) and choose the option "current namespace".
> 
> At step 3, odo says it can't find any bindable services in the current namespace. I am not suggesting that this is incorrect/invalid behaviour. I'm simply wondering what should the behaviour be:
> 
> 1. Should it, as is currently the case, fail to list bindable services since step 2 above changed the kubeconfig to use `default` namespace?
> 2. Or, should odo smartly figure that since the `env.yaml` file does have a namespace set, it shouldn't use the namespace pointed to by kubeconfig but use the one in `env.yaml` instead?

I'm pretty sure we don't want anymore odo to be *smart* to detect the right namespace for the user, as we have seen that it is counter-productive: users generally knows very well on which namespace they are operating (lots of people are setting the current Kubernetes namespace in the PS1 of their shell!), and making odo operate in another namespace is disturbing for the user.

**Comment by kadel** [comment]
> > What I am describing below was purposefully done to break stuff.
> > I first created an Operator backed service in the namespace `myproject`.
> > 
> > 1. `odo dev` in dir1. Component gets pushed to the namespace `myproject`. Keep this running.
> > 2. `cd` to dir2 and set the namespace using `odo set namespace default`. `odo dev` pushes to the namespace `default`. Keep it running or stop it, doesn't matter.
> > 3. `cd` to dir1 and do `odo add binding` (interactive mode) and choose the option "current namespace".
> > 
> > At step 3, odo says it can't find any bindable services in the current namespace. I am not suggesting that this is incorrect/invalid behaviour. I'm simply wondering what should the behaviour be:
> > 
> > 1. Should it, as is currently the case, fail to list bindable services since step 2 above changed the kubeconfig to use `default` namespace?
> > 2. Or, should odo smartly figure that since the `env.yaml` file does have a namespace set, it shouldn't use the namespace pointed to by kubeconfig but use the one in `env.yaml` instead?
> 
> I'm pretty sure we don't want anymore odo to be _smart_ to detect the right namespace for the user, as we have seen that it is counter-productive: users generally knows very well on which namespace they are operating (lots of people are setting the current Kubernetes namespace in the PS1 of their shell!), and making odo operate in another namespace is disturbing for the user.


Originally I was thinking about the scenario that @dharmit mentioned as well so I thought that we would still have namespace in env.yaml for the duration of the `odo dev` run.

@feloy has a point, this could still be misleading for some users. Let's keep it simple and just use the current namespace.

/approve

**Comment by dharmit** [comment]
> Originally I was thinking about the scenario that @dharmit mentioned as well so I thought that we would still have namespace in env.yaml for the duration of the `odo dev` run.
> 
> @feloy has a point, this could still be misleading for some users. Let's keep it simple and just use the current namespace.

Yep. In fact, when I was breaking odo for specifically that scenario, it felt like going a bit over the board to just break odo. We can wait till some users share feedback around it.

File: docs/website/docs/command-reference/dev.md [hunk 7]
```diff
@@ -417,13 +417,9 @@ This state file contains the forwarded ports:
 
 ## Env File
 
-When `odo dev` is executed, a `.odo/env/env.yaml` file is created if it does not exist. It stores the namespace/project that will be used by the component, and its default value is set to currently active namespace/project.
+When `odo dev` is executed, a `.odo/env/env.yaml` file is created if it does not exist. It stores the namespace/project that will be used by the component during the dev session. The value is removed from the file at the end of the session.
```
dharmit, line 420: Not meaning to make this change here, but this seems like a good candidate to be put into `devstate.json` instead. WDYT?

dharmit (commented):
/lgtm 

@valaparthvi, IMO your change requests have been addressed.

docs/website/docs/command-reference/dev.md
feloy: Yes, I think so. @kadel what's your opinion?

feloy:
Tests pass on IBM Cloud

PR #6045, opened 2022-08-22
Adds option to trigger manual sync, updates UI of odo dev

cdrage:
**What type of PR is this:**

/kind feature

This PR:
* Adds the ability to press "p" to "push" changes manually / syncing
* Updates the general UI / flow of `odo dev`

Fixes https://github.com/redhat-developer/odo/issues/5634

**PR acceptance criteria:**

- [ ] Unit test

- [ ] Integration test

- [ ] Documentation

### DIFF
--- a/pkg/watch/pod_phases.go
+++ b/pkg/watch/pod_phases.go
@@ -1,6 +1,7 @@
 package watch
 
 import (
+	"fmt"
 	"io"
 	"sort"
 	"strings"
@@ -42,7 +43,7 @@ func (o *PodPhases) Delete(out io.Writer, pod *corev1.Pod) {
 func (o PodPhases) Display(out io.Writer) {
 
 	if len(o) == 0 {
-		log.Fwarning(out, "No pod exists")
+		log.Fwarning(out, "No pod exists [Background task]")
 		return
 	}
 
@@ -54,10 +55,10 @@ func (o PodPhases) Display(out io.Writer) {
 	if len(keys) == 1 {
 		phase := o[keys[0]]
 		if phase == corev1.PodRunning {
-			log.Fsuccess(out, "Pod is "+phase)
+			log.Fsuccess(out, fmt.Sprintf("Pod is %s [Background task]", phase))
 			return
 		}
-		log.Fwarning(out, "Pod is "+phase)
+		log.Fwarning(out, fmt.Sprintf("Pod is %s [Background task]", phase))
 		return
 	}
 
@@ -69,5 +70,5 @@ func (o PodPhases) Display(out io.Writer) {
 	for _, k := range keys {
 		values = append(values, string(o[k]))
 	}
-	log.Fwarning(out, "Pods are "+strings.Join(values, ", "))
+	log.Fwarning(out, fmt.Sprintf("Pods are %s [Background task]", strings.Join(values, ", ")))
 }
--- a/pkg/watch/watch.go
+++ b/pkg/watch/watch.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/devfile/library/pkg/devfile/parser"
+	tty "github.com/mattn/go-tty"
 
 	_delete "github.com/redhat-developer/odo/pkg/component/delete"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters"
@@ -38,7 +39,10 @@ import (
 const (
 	// PushErrorString is the string that is printed when an error occurs during watch's Push operation
 	PushErrorString = "Error occurred on Push"
-	CtrlCMessage    = "Press Ctrl+c to exit `odo dev` and delete resources from the cluster"
+	PromptMessages  = `
+ [Ctrl+c] - Exit and delete resources from the cluster
+ [p]      - Manually sync / push files to the cluster
+`
 )
 
 type WatchClient struct {
@@ -271,7 +275,9 @@ func (o *WatchClient) WatchAndPush(out io.Writer, parameters WatchParameters, ct
 		log.Fwarning(out, "Unable to watch Events resource, warning Events won't be displayed")
 	}
 
-	return o.eventWatcher(ctx, sourcesWatcher, deploymentWatcher, devfileWatcher, podWatcher, warningsWatcher, parameters, out, evaluateFileChanges, processEvents, componentStatus)
+	keyboardInput := createKeyWatcher(ctx)
+
+	return o.eventWatcher(ctx, sourcesWatcher, deploymentWatcher, devfileWatcher, podWatcher, warningsWatcher, parameters, out, evaluateFileChanges, processEvents, componentStatus, keyboardInput)
 }
 
 func getFullSourcesWatcher(path string, fileIgnores []string) (*fsnotify.Watcher, error) {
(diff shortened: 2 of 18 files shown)


pkg/watch/watch.go
```diff
@@ -320,25 +324,81 @@ func (o *WatchClient) eventWatcher(ctx context.Context, sourcesWatcher *fsnotify
 
 	podsPhases := NewPodPhases()
 
+	// Open up a tty to read in character keys from the user, this will be done in the background
```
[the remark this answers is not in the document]
cdrage: Wouldn't we have to keep it in the eventWatcher function because that's where the main select loop is? I had it incorporated in WatchAndPush before until I realized that we have to have the main loop in order to coordinate full pushing / updating.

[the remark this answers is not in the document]
feloy: To stick with how it is done for other watchers, you would need to have a `createKeyWatcher` function similar to :

```
func createKeyWatcher() (chan string {
        manualSync := make(chan string)
        // In order to capture the keyboard events, we'll use the "tty" library.
	// Open up a tty to read in character keys from the user, this will be done in the background
	tty, err := tty.Open()
	if err != nil {
		klog.Fatal(err)
	}
	defer tty.Close()

	// Create a go routine that will read a character from the tty in the background without interruption
	go func() {
		for {
			char, err := tty.ReadRune()
			if err != nil {
				klog.Fatal(err)
			}
			// If we received the "r" character, we will push to the open channel for manual syncing.
			if string(char) == "p" {
				manualSync <- string(char)
			}
		}
	}()
        return manualSync
}
```

and call this function from `watchAndPush`, then pass the returned `manualSync` to `eventWatcher`

This has been done this way (creating the channels outside of the `eventWatcher` function and passing them to `eventWatcher`), to be able to unit-test the `eventWatcher` by faking the channels.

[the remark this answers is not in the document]
cdrage: Ahhhh, I see. I've gone ahead and implemented that. Working on tests now.

pkg/watch/pod_phases.go
```diff
@@ -42,7 +43,7 @@ func (o *PodPhases) Delete(out io.Writer, pod *corev1.Pod) {
 func (o PodPhases) Display(out io.Writer) {
 
 	if len(o) == 0 {
-		log.Fwarning(out, "No pod exists")
+		log.Fwarning(out, "No pod exists [Background task]")
```
feloy, line 46: What means "Background task"? Is this information necessary?

pkg/watch/watch.go
```diff
@@ -318,27 +324,64 @@ func (o *WatchClient) eventWatcher(ctx context.Context, sourcesWatcher *fsnotify
...
+				events = []fsnotify.Event{} // empty the events slice to capture new events
+			}
+
+			// Reset the retry timer
+			if retry != nil {
+				retryTimer.Reset(*retry)
+			} else {
+				retryTimer.Reset(time.Millisecond)
+				<-retryTimer.C
+			}
```
feloy, line 368: Ths code is very similar to the one in `case <-sourcesTimer.C:`

We should try to create a function with this code.

pkg/watch/watch.go
```diff
@@ -572,7 +615,9 @@ func processEvents(
 }
 
 func (o *WatchClient) CleanupDevResources(devfileObj parser.DevfileObj, out io.Writer) error {
-	fmt.Fprintln(out, "Cleaning resources, please wait")
+	spinner := log.Spinner("Cleaning resources, please wait")
```
feloy, line 618: When `odo dev` exits, my terminal is in a bad state, the keys are not displayed on my command line (like in password mode), I need to run the `reset` command to make it work correctly again.

feloy: Not sure if this is this spinner responsible.

pkg/watch/watch.go
```diff
@@ -636,9 +682,43 @@ func removeDuplicates(input []string) []string {
...
+	log.Sectionf("Dev mode")
+	fmt.Fprintf(out, " %s\n Watching for changes in the current directory %s\n\n %s%s", log.Sbold("Status:"), path, log.Sbold("Keyboard Commands:"), PromptMessages)
 }
 
 func isFatal(err error) bool {
 	return errors.As(err, &adapters.ErrPortForward{})
 }
+
+// createKeyWatcher will create a go routine that will run in the backgrond in order to
+// watch for changes. In this paticular case it'll
```
feloy, line 694: Please document what is the return value

feloy (changes requested):

rm3l (changes requested):
Added some comments.

Concerning the tests, I think there is a test in [`cmd_dev_test.go`](https://github.com/redhat-developer/odo/blob/main/tests/integration/cmd_dev_test.go#L381-L409) that deals with the `--no-watch` flag..
Can you also write a test case that ensures that we can manually trigger sync when `odo` is executed with the `--no-watch` flag?
To me, the "--no-watch" case is more important to have tested.

cdrage:
Working on passing the tests! 💯

tests/integration/interactive_dev_test.go
```diff
@@ -93,7 +93,42 @@ var _ = Describe("odo dev interactive command tests", func() {
...
+					ctx.StopCommand()
+				})
+
+			Expect(helper.ListFilesInDir(commonVar.Context)).To(ContainElements("devfile.yaml"))
+		})
+
+		It("should run alizer and then be able to run manual syncing", func() {
+
+			language := "python"
+			_, _ = helper.RunInteractive([]string{"odo", "dev", "--random-ports"},
```
rm3l, line 106: As I commented out in https://github.com/redhat-developer/odo/pull/6045#pullrequestreview-1087171779, would it  be possible to add a test (or enrich the existing one) that makes use of the "--no-watch" flag (see [cmd_dev_test.go](https://github.com/redhat-developer/odo/blob/main/tests/integration/cmd_dev_test.go#L369-L394))?

The [existing test](https://github.com/redhat-developer/odo/blob/main/tests/integration/cmd_dev_test.go#L369-L394) currently makes sure that no push is triggered after changing a file while running `odo dev --no-watch`. Per the acceptance criteria, I think we could enrich it to make sure that updating a file and pressing the "p" key triggers a sync.. Same point without the "--no-watch" flag.
To me, this behavior is important to test to ensure no future regression.

rm3l (changes requested):

pkg/watch/watch.go
```diff
@@ -636,9 +682,40 @@ func removeDuplicates(input []string) []string {
...
+	fmt.Fprintf(out, " %s\n Watching for changes in the current directory %s\n\n %s%s", log.Sbold("Status:"), path, log.Sbold("Keyboard Commands:"), PromptMessages)
 }
 
 func isFatal(err error) bool {
 	return errors.As(err, &adapters.ErrPortForward{})
 }
+
+// createKeyWatcher will create a go routine that will run in the background in order to
+// watch user key inputs
+func createKeyWatcher() chan string {
```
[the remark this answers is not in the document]
rm3l: Looks like the change was reverted by a Git force-push.. Was it causing issues?

pkg/watch/watch.go
```diff
@@ -294,7 +300,7 @@ func getFullSourcesWatcher(path string, fileIgnores []string) (*fsnotify.Watcher
 // eventWatcher loops till the context's Done channel indicates it to stop looping, at which point it performs cleanup.
 // While looping, it listens for filesystem events and processes these events using the WatchParameters to push to the remote pod.
 // It outputs any logs to the out io Writer
-func (o *WatchClient) eventWatcher(ctx context.Context, sourcesWatcher *fsnotify.Watcher, deploymentWatcher watch.Interface, devfileWatcher *fsnotify.Watcher, podWatcher watch.Interface, eventsWatcher watch.Interface, parameters WatchParameters, out io.Writer, evaluateChangesHandler evaluateChangesFunc, processEventsHandler processEventsFunc, componentStatus ComponentStatus) error {
+func (o *WatchClient) eventWatcher(ctx context.Context, sourcesWatcher *fsnotify.Watcher, deploymentWatcher watch.Interface, devfileWatcher *fsnotify.Watcher, podWatcher watch.Interface, eventsWatcher watch.Interface, parameters WatchParameters, out io.Writer, evaluateChangesHandler evaluateChangesFunc, processEventsHandler processEventsFunc, componentStatus ComponentStatus, keyboardInput chan string) error {
```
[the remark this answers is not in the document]
rm3l: Looks like the change was reverted by a Git force-push.. Was it causing issues?

Issue #5984, opened 2022-07-28
Cleanup `ComponentSettings` struct and `LocalConfigProvider` interface


rm3l:
/kind task
/area refactoring

This is a follow-up to a discussion in #5852

> Would it be possible to clean up the code by removing both `Name` and `AppName` fields from the [`ComponentSettings`](https://github.com/redhat-developer/odo/blob/main/pkg/envinfo/types.go) struct?
> I guess this will require cleaning up further things, like removing the `GetApplication` and `GetName` methods from the [`LocalConfigProvider`](https://github.com/redhat-developer/odo/blob/main/pkg/localConfigProvider/localConfigProvider.go#L25-L26) interface as well..
> 
> Maybe I've missed something, but I think that this way, we would be sure no one can use that information (I've tested this PR against an existing project that already has component and app names in its `env.yaml` files, but then from the code perspective, there is currently nothing in this PR preventing from loading and using such information).

_Originally posted by @rm3l in https://github.com/redhat-developer/odo/pull/5852#pullrequestreview-1052758059_

PR #6057, opened 2022-08-25
Do not use env.yaml file anymore

### LINKED ISSUE #6046: Remove env.yaml
Get rid of the env.yaml file. We do not need to store namespaces anymore, and there is no other info that we're storing in this file, so it makes sense to get rid of it.
Doing this might also help with v2-v3 transition since v2 can handle cases where there is no env.yaml as compared to half-baked env.yaml

### Acceptance Criteria

- [ ] Make sure `odo dev` only takes namespace at the beginning of the command and does not change if namespace changes [Move this to a separate issue of it's own.]
- [ ] Remove any references to `env.yaml`

/kind user-story

### LINKED ISSUE #6056: DEBUG_PORT variable is injected into container by `odo dev`
/kind bug

**Operating System:**

Fedora Linux

**Output of `odo version`:**

odo v3.0.0-beta3 (644d94cde)

```
$ odo init --devfile nodejs --starter nodejs-starter --name mycomponent
$ odo dev
```

```
$ kubectl exec deploy/mycomponent-app -- bash -c 'echo $DEBUG_PORT'
5858
```

## Actual behavior

An environment variable is injected into the container, by `odo dev`, when this env var is not defined in the devfile.

## Expected behavior

No variable should be injected automatically by `odo dev`. The DEBUG_PORT variable should be described in the devfile. 
See for an example https://registry.devfile.io/viewer/devfiles/Community+java-maven

feloy:
**What type of PR is this:**

/kind bug

This PR removes the concept of env.yaml file for `odo`. 

> The `envinfo.EnvSpecificInfo` structure now only contains information about devfile: devfilePath and devfileObj. 
>
> The `pkg/storage` package is still using this structure to get the devfileObj. A refactoring of this package would be useful to move the devfilePath and devfileObj to the Context structure directly and get rid of the EnvSpecificInfo structure, but is not part of this PR.

Fixes #6046 
Fixes #6056 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/envinfo/envinfo.go
```diff
@@ -93,225 +35,17 @@ func NewEnvSpecificInfo(envDir string) (*EnvSpecificInfo, error) {
 // newEnvSpecificInfo retrieves the env.yaml file, if it does not exist, we return a *BLANK* environment file.
 func newEnvSpecificInfo(envDir string, fs filesystem.Filesystem) (*EnvSpecificInfo, error) {
```
valaparthvi, line 36: Nit: You can move the contents of this function to NewEnvSpecificInfo.

pkg/odo/cli/dev/dev.go
```diff
@@ -213,13 +184,7 @@ func (o *DevOptions) Run(ctx context.Context) (err error) {
 	}
 
 	// add odo-file-index.json path to .gitignore
```
valaparthvi, line 186:
```suggestion
	// add .odo dir path to .gitignore
```

pkg/odo/genericclioptions/resolve.go
```diff
@@ -59,43 +58,9 @@ func (o *internalCxt) resolveProjectAndNamespace(cmdline cmdline.Cmdline) error
 // Or If app is set in envfile, return its value
 // Or if createAppIfNeeded, returns the default app name
 func resolveApp(cmdline cmdline.Cmdline, localConfiguration localConfigProvider.LocalConfigProvider, createAppIfNeeded bool) string {
```
valaparthvi, line 60: Why don't we get rid of this function? If not, at least the unused arguments.

valaparthvi (changes requested):
Code lgtm, have a few suggestions.

pkg/devfile/adapters/kubernetes/utils/utils.go
```diff
@@ -16,9 +14,6 @@ import (
 const (
 	// _envProjectsRoot is the env defined for project mount in a component container when component's mountSources=true
 	_envProjectsRoot = "PROJECTS_ROOT"
-
-	// _envDebugPort is the env defined in the runtime component container which holds the debug port for remote debugging
-	_envDebugPort = "DEBUG_PORT"
```
rm3l, line 21: Wouldn't it make more sense to handle #6056 in a separate PR (with a more specific test case, which I did not see here)? I don't really see the link between `env.yaml` and the Debug port env var here ;-)

rm3l:
I've also linked this PR as closing #5984, as I think it is covered here.

pkg/devfile/adapters/kubernetes/utils/utils.go
rm3l: Ah, I see: the debug port used to be stored in `env.yaml`, right? In this case, I think what's missing (unless I missed it) is a test case for https://github.com/redhat-developer/odo/issues/6056 to ensure `odo dev` does not inject a `DEBUG_PORT` env var into the container.

feloy: Yes, this is the reason why it is done here. Let me check if a test exists for this case

tests/integration/cmd_dev_debug_test.go
```diff
@@ -53,6 +53,13 @@ var _ = Describe("odo dev debug command tests", func() {
 				// We are just using this to validate if nodejs agent is listening on the other side
 				helper.HttpWaitForWithStatus("http://"+ports["5858"], "WebSockets request was expected", 12, 5, 400)
 			})
+
+			// #6056
+			It("should not add a DEBUG_PORT variable to the container", func() {
+				podName := commonVar.CliRunner.GetRunningPodNameByComponent(cmpName, commonVar.Project)
+				stdout := commonVar.CliRunner.Exec(podName, commonVar.Project, "--", "sh", "-c", "echo ${DEBUG_PORT}")
```
rm3l, line 60:
```suggestion
				stdout := commonVar.CliRunner.Exec(podName, commonVar.Project, "--", "sh", "-c", "echo -n ${DEBUG_PORT}")
```

I guess the trailing newline added by `echo` might be captured into `stdout`..

rm3l (changes requested):

rm3l (approved):

PR #6086, opened 2022-09-01
Reorganizes installation document

cdrage:
Reorganizes installation document

**What type of PR is this:**

/kind documentation

- Separates Homebrew and Binary sections for MAC
- Separates installing from source / maven plugins / etc.

Fixes https://github.com/redhat-developer/odo/issues/6082

### DIFF
--- a/docs/website/docs/overview/installation.md
+++ b/docs/website/docs/overview/installation.md
@@ -5,13 +5,11 @@ sidebar_position: 4
 
 `odo` can be used as either a [CLI tool](./installation#cli-binary-installation) or an [IDE plugin](./installation#ide-installation) on Mac, Windows or Linux.
 
-## CLI installation
-
 Each release is *signed*, *checksummed*, *verified*, and then pushed to our [binary mirror](https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/odo/). 
 
 For more information on the changes of each release, they can be viewed either on [GitHub](https://github.com/redhat-developer/odo/releases) or the [blog](/blog).
 
-### Linux
+## Linux
 
 import Tabs from '@theme/Tabs';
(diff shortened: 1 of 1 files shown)


docs/website/docs/overview/installation.md
```diff
@@ -270,8 +267,24 @@ type odo.exe.sha256
 
 4. Add the binary to your `PATH`
 
+## Visual Studio Code Plugin
+
+Installs `odo` capabilities Visual Studio Code (VSCode).
```
rm3l, line 272:
```suggestion
Installs `odo` capabilities in Visual Studio Code (VSCode).
```

docs/website/docs/overview/installation.md
```diff
@@ -270,8 +267,24 @@ type odo.exe.sha256
 
 4. Add the binary to your `PATH`
 
+## Visual Studio Code Plugin
```
rm3l, line 270: Since the VSCode Plugin currently installs odo v2, I was wondering if we shouldn't hide this section for now. When the VSCode Plugin migrates to v3, then we would consider adding this section back. Would even make more sense as part of the work to be done for https://github.com/redhat-developer/odo/issues/5674.
@kadel Any thoughts?

rm3l (changes requested):

docs/website/docs/overview/installation.md
valaparthvi: Considering this doc points to v3, I agree with Armel. It makes more sense to hide this section for now.

cdrage: Brew also installs v2 of odo Mac, so 🤷 

Let's get @kadel 's opinion on this

cdrage:
@rm3l @valaparthvi Updated, ready for another review 💯

PR #6089, opened 2022-09-02
Manual sync of files pressing p

### LINKED ISSUE #5634: manually trigger file sync in `odo dev`
## User Story
- As an odo user running `odo dev --no-watch`
- I want to be able to trigger push and restart on demand
- So that I can trigger push only after I made multiple edits, because I don't need to see every single small change and on a complex application it can be expensive to trigger rebuild

## Acceptance Criteria

1. This issue
- [ ] when `odo dev` is running user should be able to trigger push by pressing a key (p).
- [ ] `odo dev` output should inform user about this option
- [ ] triggering push on-demand should be possible regardless if `odo dev` is running with or without `--no-watch`

2. How to trigger the file sync
 https://github.com/redhat-developer/odo/issues/5986

/kind user-story

feloy:
**What type of PR is this:**

/kind feature

This PR configures the terminal to work in Character mode, instead of the default Line mode. Read() will return as soon as a character is sent to the input stream, instead of waiting for a newline character.

code refactoring:
- moved getFullSourcesWatcher and related functions to a specific file file_watcher.go
- moved long list of watchers variables into receiver instead of passing as parameters

Fixes #5634

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

- `odo dev` will fail to configure terminal on Windows/cygwin terminal.

  You can use `$ winpty odo dev` to fix the problem

  https://bryantson.medium.com/troubleshooting-make-raw-terminal-failed-the-handle-is-invalid-57b7e1d9b12b

- Windows terminal seetings
  https://docs.microsoft.com/en-us/windows/console/setconsolemode

- Unix terminal settings:
  https://man7.org/linux/man-pages/man3/termios.3.html

## Refactoring integration tests

Some tests were calling devSession.Kill() (or Stop()) twice, which makes ginkgo stop abrupty during the second call on Windows.

In the following example, the Kill in AfterEach is necessary for the cleanup after test1, but is problematic for test2, as Stop has been called before.

As the AfterEach/BeforeEach are evaluated at the beginning, it is not possible to rely on a field in devSession to record if the method has already been called.

```
When("odo dev is executed", func() {

			var devSession helper.DevSession

			BeforeEach(func() {
				var err error
				devSession, _, _, _, err = helper.StartDevMode(nil)
				Expect(err).ToNot(HaveOccurred())
			})

			AfterEach(func() {
				devSession.Kill()
				devSession.WaitEnd()
			})

			It("test 1", func() { ... })

			When("odo dev is stopped", func() {
				BeforeEach(func() {
					devSession.Stop()
					devSession.WaitEnd()
				})

				It("test 2", func() { ... })
			})
		})
```

These tests have been split in two trees, separating test1 and test2, to be able to call Kill/Stop only once.


feloy:
Blocked by #6091

feloy:
Thanks @rm3l for your review

pkg/watch/watch.go
```diff
@@ -38,13 +35,26 @@ import (
 const (
 	// PushErrorString is the string that is printed when an error occurs during watch's Push operation
 	PushErrorString = "Error occurred on Push"
-	CtrlCMessage    = "Press Ctrl+c to exit `odo dev` and delete resources from the cluster"
+	PromptMessage   = `
+[Ctrl+c] - Exit and delete resources from the cluster
+     [p] - Manually sync / push files to the cluster
```
rm3l, line 40: One last thing I was thinking of: maybe we could also say that pressing 'p' might also restart/reload the running application.
This could be mentioned here and/or in the documentation for `odo dev`.
Other than sync'ing source files, I didn't see anything related to the running application mentioned (unless that's what we mean by "sync").

docs/website/docs/command-reference/dev.md
```diff
@@ -46,6 +48,22 @@ In the above example, three things have happened:
...
+
+By default, the changes made by the user to the Devfile and source files are applied directly.
+
+The flag `--no-watch` can be used to change this behaviour: when the user changes the devfile or any source file, the changes
+won't be applied immediately, but the next time the user presses the `p` key.
+
+Depending on the local changes, different events can occur on the cluster:
+
+- if source files are modified, they are pushed to the container running the application, and:
+  - if component is marked as `HotReloadCapable`, the application is responsible for applying the new changes
```
rm3l, line 61: `HotReloadCapable` is a field at the command level, not the component.

rm3l (approved):
Thanks for the changes.
LGTM. I'll let other reviewers take another look.

dharmit:
Unrelated to the code changes in this PR. I noticed that the `watch.go` file has many timers now. 

```go
sourcesTimer := time.NewTimer(time.Millisecond)
<-sourcesTimer.C

devfileTimer := time.NewTimer(time.Millisecond)
<-devfileTimer.C

deployTimer := time.NewTimer(time.Millisecond)
<-deployTimer.C

retryTimer := time.NewTimer(time.Millisecond)
<-retryTimer.C
```
IMO, it would help if we can add a comment that points which watcher it corresponds to. I'm OK if we want to open a new issue and mark it as good first issue (Hacktoberfest is just around the corner.)

pkg/watch/key_watcher_windows.go
```diff
@@ -0,0 +1,20 @@
+package watch
+
+import (
+	"golang.org/x/sys/windows"
+)
+
+// makeRaw is forked from golang.org/x/term
```
dharmit, line 7: What is the `makeRaw` that you refer to here? I don't find it anywhere except this comment:
```shell
$  grep -r makeRaw pkg/      
pkg/watch/key_watcher_windows.go:// makeRaw is forked from golang.org/x/term

```

pkg/watch/term_unix_other.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
dharmit, line 3: Is it OK to have this when the odo code is licensed as Apache Licence 2.0?

pkg/watch/term_unix_other.go
```diff
@@ -0,0 +1,13 @@
...
+
+//go:build aix || linux || solaris || zos
+// +build aix linux solaris zos
+
+package watch
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TCGETS
+const ioctlWriteTermios = unix.TCSETS
```
dharmit, line 13: What are these constants for exactly? 
What's the significance of this file and the file `term_unix_bsd.go`?

pkg/watch/key_watcher_unix.go
```diff
@@ -0,0 +1,21 @@
...
+	if err != nil {
+		return err
+	}
+	termios.Lflag &^= unix.ICANON
+	if err := unix.IoctlSetTermios(fd, ioctlWriteTermios, termios); err != nil {
+		return err
+	}
+
+	return nil
+}
```
dharmit, line 21: I find this to be magic. Can you explain in code comment as to what this is happening here?

feloy:
> Unrelated to the code changes in this PR. I noticed that the `watch.go` file has many timers now.
> 
> ```go
> sourcesTimer := time.NewTimer(time.Millisecond)
> <-sourcesTimer.C
> 
> devfileTimer := time.NewTimer(time.Millisecond)
> <-devfileTimer.C
> 
> deployTimer := time.NewTimer(time.Millisecond)
> <-deployTimer.C
> 
> retryTimer := time.NewTimer(time.Millisecond)
> <-retryTimer.C
> ```
> 
> IMO, it would help if we can add a comment that points which watcher it corresponds to. I'm OK if we want to open a new issue and mark it as good first issue (Hacktoberfest is just around the corner.)

I've added some comments

pkg/watch/file_watcher.go
```diff
@@ -0,0 +1,137 @@
+package watch
```
dharmit, line 1: @feloy, there's a bunch of things being added/moved under the `pkg/watch` directory. I humbly request you to please add comments to the code. Especially functions, even if they are internal. It feels like magic right now, and could be really difficult for others to comprehend what's going on.

dharmit (changes requested):

pkg/watch/key_watcher_windows.go
feloy: `enableCharInput` just below. Sorry I renamed the function name, but didn't update the comment

pkg/watch/term_unix_other.go
feloy: It is the license for the Go library and the code coming from `golang.org/x/*` repos.
I can imagine it is ok

This code specifically comes from the `golang.org/x/term` repository (https://cs.opensource.google/go/x/term/+/a9ba230a:term_unix_other.go)

pkg/watch/key_watcher_unix.go
feloy: I've put links on the PRs for you to understand what it is about:

https://man7.org/linux/man-pages/man3/termios.3.html

This is not magic, but terminal settings.

pkg/watch/file_watcher.go
feloy: I've added some more comments to the functions. But note that I added into the PR comment links to Unix and Windows terminal settings reference, that will help you understand what it is about.

pkg/watch/term_unix_other.go
feloy: They are the constants for Terminal I/Os on Unix. Unfortunately, constants values are different on BSD and on other Unixes, thus the 2 different files with different constants.

pkg/watch/file_watcher.go
dharmit: I'm referring to below functions:
* `getKeyWatcher`
* `getKey`
specifically, what are the go routines doing?

dharmit: Thanks. I just read them and it makes a huge difference for me personally. The explanations are very helpful!

dharmit:
/lgtm
/approve

Adding both flags because @rm3l already lgtm'd, but my requests removed the flag.
Thanks @feloy ! 🎉

feloy:
Tests pass on IBM Cloud

feloy:
@anandrkskd I've added a PR to fix the problem of failing integration tests on Windows

rm3l (approved):

feloy:
/ci/prow/v4.10-integration-e2e
Tests pass on IBM cloud

feloy:
Tests pass on IBM cloud

PR #6108, opened 2022-09-09
check if namespace is created instead of project

anandrkskd:

**What type of PR is this:**

/kind tests

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/helper/helper_cli.go
+++ b/tests/helper/helper_cli.go
@@ -26,6 +26,7 @@ type CliRunner interface {
 
 	// HasNamespaceProject returns whether the specified namespace or project exists in the cluster
 	HasNamespaceProject(name string) bool
+	// ListNamespaceProject checks if the namespace is present in the list of namespaces
 	ListNamespaceProject(name string)
 
 	GetActiveNamespace() string
--- a/tests/helper/helper_cli.go
+++ b/tests/helper/helper_cli.go
@@ -26,6 +26,7 @@ type CliRunner interface {
 
 	// HasNamespaceProject returns whether the specified namespace or project exists in the cluster
 	HasNamespaceProject(name string) bool
+	ListNamespaceProject(name string)
 
 	GetActiveNamespace() string
 	GetEnvsDevFileDeployment(componentName, appName, projectName string) map[string]string
--- a/tests/helper/helper_oc.go
+++ b/tests/helper/helper_oc.go
@@ -582,6 +582,12 @@ func (oc OcRunner) HasNamespaceProject(name string) bool {
 	return strings.Contains(out, name)
 }
 
+func (oc OcRunner) ListNamespaceProject(name string) {
+	Eventually(func() string {
+		return Cmd(oc.path, "get", "project").ShouldRun().Out()
+	}, 30, 1).Should(ContainSubstring(name))
+}
+
 func (oc OcRunner) GetActiveNamespace() string {
 	return Cmd(oc.path, "config", "view", "--minify", "-ojsonpath={..namespace}").ShouldPass().Out()
 }
--- a/tests/integration/interactive_add_binding_test.go
+++ b/tests/integration/interactive_add_binding_test.go
@@ -226,8 +226,8 @@ var _ = Describe("odo add binding interactive command tests", func() {
 					helper.GetExamplePath("manifests", "bindablekind-instance.yaml"))
 				Expect(addBindableKindInOtherNs.ExitCode()).To(BeEquivalentTo(0))
 				nsWithNoService = commonVar.CliRunner.CreateAndSetRandNamespaceProject()
-				Expect(commonVar.CliRunner.HasNamespaceProject(nsWithNoService)).Should(Equal(true))
-				Expect(commonVar.CliRunner.HasNamespaceProject(otherNS)).Should(Equal(true))
+				commonVar.CliRunner.ListNamespaceProject(nsWithNoService)
+				commonVar.CliRunner.ListNamespaceProject(otherNS)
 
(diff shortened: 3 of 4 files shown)


valaparthvi (commented):
Why do we need this change? Isn't `HasNamespaceProject` enough?

anandrkskd:
> Why do we need this change? Isn't HasNamespaceProject enough?
We are still experiencing test flakes for these test cases ([flake chart](https://github.com/anandrkskd/odo-tools/blob/master/output/flake-stats.md)), even if we are checking for project. @feloy suggested it would be better to check if the `namespace` is present instead of `project`. This PR implements the same.

tests/helper/helper_oc.go
```diff
@@ -577,7 +577,7 @@ func (oc OcRunner) GetNamespaceProject() string {
 }
 
 func (oc OcRunner) HasNamespaceProject(name string) bool {
-	out := Cmd(oc.path, "get", "project", name, "-o", "jsonpath={.metadata.name}").
+	out := Cmd(oc.path, "get", "project", "-o", "jsonpath={.metadata.name}").
```
kadel, line 580: This is not going to work, as you removed the `name` parameter the command will return all project on the cluster so the `jsonpath` is no longer valid as not the command returns the array

tests/helper/helper_oc.go
```diff
@@ -577,7 +577,7 @@ func (oc OcRunner) GetNamespaceProject() string {
 }
 
 func (oc OcRunner) HasNamespaceProject(name string) bool {
```
feloy, line 579: You are changing the function `HasNamespaceProject` to use the List operation instead of Get, it will have an effect for all the tests calling this function. But we see that the change is not transparent, and could introduce other flaky tests.

It would be safer to create another function using the List operation, for example a `ListNamespacesProjects` that would return the list of namespaces/projects. And you could check for the existence of requested name on the value returned by the function.

feloy (changes requested):

tests/integration/interactive_add_binding_test.go
```diff
@@ -226,8 +226,8 @@ var _ = Describe("odo add binding interactive command tests", func() {
 					helper.GetExamplePath("manifests", "bindablekind-instance.yaml"))
 				Expect(addBindableKindInOtherNs.ExitCode()).To(BeEquivalentTo(0))
 				nsWithNoService = commonVar.CliRunner.CreateAndSetRandNamespaceProject()
-				Expect(commonVar.CliRunner.HasNamespaceProject(nsWithNoService)).Should(Equal(true))
-				Expect(commonVar.CliRunner.HasNamespaceProject(otherNS)).Should(Equal(true))
+				commonVar.CliRunner.ListNamespaceProject(nsWithNoService)
```
feloy, line 229: You need to use Eventually

feloy (approved):

tests/helper/helper_cli.go
```diff
@@ -26,6 +26,7 @@ type CliRunner interface {
 
 	// HasNamespaceProject returns whether the specified namespace or project exists in the cluster
 	HasNamespaceProject(name string) bool
+	ListNamespaceProject(name string)
```
dharmit, line 29: Can you add a comment here that would help understand when to use `HasNamespaceProject` vs. `ListNamespaceProject`?

dharmit (changes requested):

rm3l:
@anandrkskd Can you please rebase your branch onto `main`?

I have the feeling that the potential issue (about Windows tests) I mentioned yesterday during standup is occurring here too. When looking at [the logs](https://cloud.ibm.com/devops/pipelines/929da3ed-6cad-4026-a847-7a8485f20fa4/582614a3-b910-4659-ac27-6288668c5fe6/a1d74ffa-3d39-4dbb-8583-b0a34f87f3bd/1971d1b6-a8b0-445b-afc2-ec1118bd5d46/d365c569-6e4b-4a51-b7c8-8ba994c1892a?env_id=ibm:yp:eu-de), I can see that this test (`should expose all endpoints on localhost regardless of exposure`) is not passing:
```
+ [FAILED] [240.286 seconds]
odo dev command tests
C:/Users/Administrator.ANSIBLE-TEST-VS/2011/tests/integration/cmd_dev_test.go:31
  C:/Users/Administrator.ANSIBLE-TEST-VS/2011/tests/integration/cmd_dev_test.go:471
    when devfile has multiple endpoints
    C:/Users/Administrator.ANSIBLE-TEST-VS/2011/tests/integration/cmd_dev_test.go:548
      when running odo dev
      C:/Users/Administrator.ANSIBLE-TEST-VS/2011/tests/integration/cmd_dev_test.go:555
        [It] should expose all endpoints on localhost regardless of exposure
        C:/Users/Administrator.ANSIBLE-TEST-VS/2011/tests/integration/cmd_dev_test.go:573
```

The problem is that this test was introduced recently in 16c6d1589f364ac81b679a48c9d86af6f76601b5, but your branch does not contain this commit yet; so I'm confused `¯\_(ツ)_/¯`

```sh
❯ git remote add anandrkskd git@github.com:anandrkskd/odo.git
❯ git fetch anandrkskd
❯ git branch -r --contains=16c6d1589f364ac81b679a48c9d86af6f76601b5
  origin/6f77dependabot/npm_and_yarn/docs/website/mdx-js/react-2.1.3
  origin/HEAD -> origin/main
  origin/fix-links-to-devfile.io
  origin/main
  upstream/dependabot/npm_and_yarn/docs/website/mdx-js/react-2.1.3
  upstream/main
```

PR #6111, opened 2022-09-09
WIP set service account for dev container via component attributes

### LINKED ISSUE #5977: Ability to specify ServiceAccount for `odo dev`
/kind feature

When running the `odo dev` command, I would like to be able to specify the Kubernetes ServiceAccount the pod should run as.

## Why is this needed?

Currently, if you use `odo dev` the pod will start as the default service account. Some workloads running on Kubernetes need to run as a specific service account in order to have the proper permissions to work properly. One common use case is a Kubernetes Operator.

vinny-sabatini:
Signed-off-by: Vinny Sabatini <vincent.sabatini@kohls.com>

**What type of PR is this:**

/kind feature

Allow users to specify what Kubernetes service account the pod should run as for an `odo dev` session.
This is generally useful if you want your workload to have additional permissions within a Kubernetes cluster
and you do not want to grant additional access to the default service account.

If the attribute is not set, the `default` service account will be used.

Fixes #5977

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

- In your Kubernetes cluster, create a service account
- In your devfile, set `.components.attributes.serviceAccountName` for your container component to the name of the service account you created
  ```yaml
  components:
  - attributes:
      serviceAccountName: my-service-account
    container:
      dedicatedPod: false
      endpoints:
      - name: http
        secure: false
        targetPort: 8080
      memoryLimit: 1024Mi
      mountSources: true
    name: runtime
  ```
- Start `odo dev`
- Check `.spec.template.spec.serviceAccountName` on the deployment, or `.spec.serviceAccountName` of the running pod

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -458,6 +458,27 @@ func (a *Adapter) createOrUpdateComponent(
 	if err != nil {
 		return nil, false, err
 	}
+
+	// Set service account if component has attribute `serviceAccountName`
+	var serviceAccountName string
+	components, _ := a.Devfile.Data.GetComponents(parsercommon.DevfileOptions{})
+	for _, component := range components {
+		for _, container := range containers {
+			if component.Name == container.Name {
+				var err *error
+				if component.Attributes.Exists("serviceAccountName") {
+					serviceAccountName = component.Attributes.GetString("serviceAccountName", err)
+					if err != nil {
+						return nil, false, *err
+					}
+				} else {
+					serviceAccountName = "default"
+				}
+			}
+		}
+	}
+	deployment.Spec.Template.Spec.ServiceAccountName = serviceAccountName
+
 	klog.V(2).Infof("Creating deployment %v", deployment.Spec.Template.GetName())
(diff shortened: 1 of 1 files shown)


kadel:
Hi @vinny-sabatini, earlier this week, there was a discussion on one of the devfile issues. And the solution that was proposed there would also address the "ServiceAccount" issue as well but in a more generic way. 

You can see what is being proposed here https://github.com/devfile/api/issues/920#issuecomment-1244059075

vinny-sabatini:
@kadel thanks for the info! That solution looks much more flexible than trying to key every single customization and overloading the devfile API spec. I can work on updating this PR to follow the convention mentioned in that issue

kadel:
> @kadel thanks for the info! That solution looks much more flexible than trying to key every single customization and overloading the devfile API spec. I can work on updating this PR to follow the convention mentioned in that issue

odo uses [devfile/library](https://github.com/devfile/library) for generating k8s resources from devfile. I think that this logic should be implemented in the devfile/library. There is a set of functions in the `generator` package that odo uses. Those functions should be updated to use the information from the annotations.

vinny-sabatini:
This was done in https://github.com/redhat-developer/odo/pull/6512
Closing this PR

PR #6112, opened 2022-09-12
Add runtime label based on metadata:language/projectType

### LINKED ISSUE #6044: add `app.openshift.io/runtime` label to resources created by odo
/kind user-story

## User Story
As an odo and ODC user, I want to see language/framework icons in ODC topology view So that I can quickly recognize what is running and so that I have a consistent view with components created by ODC.

Screenshot 2022-08-22 at 17 12 39

The left is the component with the correct labels deployed with ODC, the right is nodejs component deployed using `odo dev`

## Acceptance Criteria 
- [ ] `odo dev` should add `app.openshift.io/runtime` label  to all resources created with value of `metadata.language` field in devfile.yaml
- [ ] `odo deploy` should add `app.openshift.io/runtime` label  to all resources created with value of `metadata.language` field in devfile.yaml

/kind user-story
/priority medium

feloy:
**What type of PR is this:**

/kind feature

Fixes #6044

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -42,13 +42,13 @@ func GetComponentTypeFromDevfileMetadata(metadata devfile.DevfileMetadata) strin
 	return componentType
 }
 
-// GetComponentRuntimeFromDevfileMetadata returns the Language defined in the Devfile metadata
-// or if not set, the Project Type
+// GetComponentRuntimeFromDevfileMetadata returns the Project Type defined in the Devfile metadata
+// or if not set, the Language
 func GetComponentRuntimeFromDevfileMetadata(metadata devfile.DevfileMetadata) string {
-	if metadata.Language != "" {
-		return metadata.Language
+	if metadata.ProjectType != "" {
+		return metadata.ProjectType
 	}
-	return metadata.ProjectType
+	return metadata.Language
 }
 
 // GatherName computes and returns what should be used as name for the Devfile object specified.
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -42,6 +42,15 @@ func GetComponentTypeFromDevfileMetadata(metadata devfile.DevfileMetadata) strin
 	return componentType
 }
 
+// GetComponentRuntimeFromDevfileMetadata returns the Language defined in the Devfile metadata
+// or if not set, the Project Type
+func GetComponentRuntimeFromDevfileMetadata(metadata devfile.DevfileMetadata) string {
+	if metadata.Language != "" {
+		return metadata.Language
+	}
+	return metadata.ProjectType
+}
+
 // GatherName computes and returns what should be used as name for the Devfile object specified.
 //
 // If a non-blank name is available in the Devfile metadata (which is optional), it is sanitized and returned.
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -12,8 +12,11 @@ import (
 // if you need labels to filter component then use GetSelector instead
 // Note: isPartOfComponent denotes if the label is required for a core resource(deployment, svc, pvc, pv) of a given component deployed with `odo dev`;
 // it is the only thing that sets it apart from the resources created via other ways (`odo deploy`, deploying resource with apply command during `odo dev`)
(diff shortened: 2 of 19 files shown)


pkg/labels/labels.go
```diff
@@ -12,8 +12,11 @@ import (
 // if you need labels to filter component then use GetSelector instead
 // Note: isPartOfComponent denotes if the label is required for a core resource(deployment, svc, pvc, pv) of a given component deployed with `odo dev`;
 // it is the only thing that sets it apart from the resources created via other ways (`odo deploy`, deploying resource with apply command during `odo dev`)
-func GetLabels(componentName string, applicationName string, mode string, isPartOfComponent bool) map[string]string {
+func GetLabels(componentName string, applicationName string, runtime string, mode string, isPartOfComponent bool) map[string]string {
 	labels := getLabels(componentName, applicationName, mode, true, isPartOfComponent)
+	if runtime != "" {
+		labels[openshiftRunTimeLabel] = runtime
+	}
```
valaparthvi, line 19: This is an OpenShift specific label, isn't it? It does not make sense to add it to resources that might be deployed in a K8s cluster.

feloy: Do you suggest to add this label only when deploying to an OpenShift cluster?

feloy: I'm not sure it is what has been discussed. @kadel WDYT?

kadel: This is just a label. Yes, it is OpenShift specific, but there is no harm in always applying this. It might be even better, as there might be another tooling that could eventually leverage that.
And the only OpenShift specific is it's name, functionality wise there is nothing that should tie this only to openshift

pkg/component/component.go
```diff
@@ -42,6 +42,15 @@ func GetComponentTypeFromDevfileMetadata(metadata devfile.DevfileMetadata) strin
 	return componentType
 }
 
+// GetComponentRuntimeFromDevfileMetadata returns the Language defined in the Devfile metadata
+// or if not set, the Project Type
+func GetComponentRuntimeFromDevfileMetadata(metadata devfile.DevfileMetadata) string {
+	if metadata.Language != "" {
+		return metadata.Language
+	}
+	return metadata.ProjectType
```
rm3l, line 51: I know we already discussed about using the language first and then the project type, but after quickly testing on Dev Sandbox, I think we should use the project type first.

image

The Deployment I created manually in Dev Sandbox has the following label: `app.openshift.io/runtime=nodejs`, but the one created by `odo` has `app.openshift.io/runtime=javascript`, and the right icon is not showing up in Topology View.

feloy: What is the behavioue with Java projects?

rm3l: With a Quarkus app, we have the right icon for the language (Java):

image

If I manually edit the label to use the project type (`quarkus`) defined in the Devfile, we have a more accurate icon for this:

image

rm3l: Same point for a SpringBoot app: we currently have the same icon as Quarkus for the language (Java). But if I set the label to the project type (`spring`) defined in the Devfile, then we have a more specific icon for this component:

image

This is why I was thinking about using the project type first instead.

kadel: I guess we never will be able to reliably cover all the cases. If we use projectType then there will be a problem with `java-maven`. I guess we just need to check what approach works for most of the cases

rm3l:
Thanks for the new changes.

/approve

cdrage:
Tested and works well 👍 

Thanks @feloy 

/lgtm

feloy:
Flaky test
```
odo add binding interactive command tests
/go/src/github.com/redhat-developer/odo/tests/integration/interactive_add_binding_test.go:16
  when the component is bootstrapped
  /go/src/github.com/redhat-developer/odo/tests/integration/interactive_add_binding_test.go:48
    when binding to a service in a different namespace
    /go/src/github.com/redhat-developer/odo/tests/integration/interactive_add_binding_test.go:219
      [It] should error out if service is not found in the namespace selected
      /go/src/github.com/redhat-developer/odo/tests/integration/interactive_add_binding_test.go:240
```

PR #6118, opened 2022-09-13
Adds support for odo list services

### LINKED ISSUE #5996: `odo list services` - list bindable services
/kind user-story

## User Story
As an odo user (developer working on IDE plugin based on odo) I want to be able to list all bindable services So that I can run `odo add binding` command non-interactively

## Acceptance Criteria 
- [x] `odo list services` should list all bindable services in the current namespace
- [x] `odo list services --namespace <foo>` should list all bindable services in the `<foo>` namespace
- [x] `odo list services --all-namespaces` should list all bindable services in all the namespace that user has access to.
- [x] `odo list services -o json` should print the output in the json format, applies when any of the flag is used as well.

```
$ odo list services

 ✓  Listing bindable services from namespace 'asdf' [5s]
 NAME                                                        NAMESPACE
 hippo/PerconaXtraDBCluster.pxc.percona.com                  asdf
 cluster1/PostgresCluster.postgres-operator.crunchydata.com  asdf

$ odo list services -o json

{
  bindableServices: [
    name: hippo/PerconaXtraDBCluster.pxc.percona.com,
    namespace: asdf
  ],
  [
    cluster1: hippo/PostgresCluster.postgres-operator.crunchydata.com,
    namespace: asdf
  ]
}
```
```
$ odo list services --all-namespaces

 ✓  Listing bindable services from namespace 'asdf' [5s]
 NAME                                                        NAMESPACE
 hippo/PerconaXtraDBCluster.pxc.percona.com                  asdf
 cluster1/PostgresCluster.postgres-operator.crunchydata.com  asdf
 cluster2/PostgresCluster.postgres-operator.crunchydata.com  ffff


$ odo list services -o json

{
  bindableServices: [
    name: hippo/PerconaXtraDBCluster.pxc.percona.com,
    namespace: asdf
  ],
  [
    cluster1: hippo/PostgresCluster.postgres-operator.crunchydata.com,
    namespace: asdf
  ],
  [
    cluster2: cluster2/PostgresCluster.postgres-operator.crunchydata.com,
    namespace: ffff
  ]
}
```

## Links 
- Related Epic (mandatory): #5994 

/kind user-story

dharmit:
**What type of PR is this:**

/kind feature

Besides $subject, it also adds a few aliases to simplify things for users.

Fixes #5996

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [x] Documentation 

Refer the original issue


pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+	services, err := o.run()
+	if err != nil {
+		return err
+	}
+	s.End(true)
+	HumanReadable(services)
+	return nil
+}
+
+func (o *ServiceListOptions) run() ([]map[string]unstructured.Unstructured, error) {
```
feloy, line 107: IIUC, you could reduce the list of services as a []Unstructured inside this function, so we don't need to pass this complex structure to other functions, which don't need it.

feloy (changes requested):

pkg/odo/cli/list/services/services.go
dharmit: Makes sense. Done.

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+func (o *ServiceListOptions) run() ([]unstructured.Unstructured, error) {
+	services := []unstructured.Unstructured{}
+
+	if o.allNamespacesFlag {
+		projects, err := o.clientset.ProjectClient.List()
+		if err != nil {
+			return services, err
+		}
+		for _, project := range projects.Items {
+			svcs, err := o.clientset.BindingClient.GetServiceInstances(project.Name)
```
feloy, line 116: This will make (# of projects x # of CRDs) calls to the API.

I think it would be more efficient to make cluster-wide calls to the API to reduce the number of calls to (# of CRDs).

feloy (changes requested):

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+
+func isServiceSliceEmpty(services []unstructured.Unstructured) bool {
+	if len(services) == 0 {
+		return true
+	}
+	if len(services) == 1 {
+		if reflect.DeepEqual(services[0], unstructured.Unstructured{}) {
+			return true
+		}
+	}
```
feloy, line 172: In which circumstances can we have this case?

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+	services, err := o.run()
+	if err != nil {
+		return nil, err
+	}
+
+	var servicesList []api.BindableService
+	for _, svc := range services {
+		servicesList = append(servicesList, api.BindableService{Name: svc.GetName(), Namespace: svc.GetNamespace()})
+	}
+	return map[string]interface{}{"bindableServices": servicesList}, nil
```
feloy, line 147: I would prefer to use a Struct here.

You could add the list of services into the api.ResourcesList struct.

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+	return nil
+}
+
+func (o *ServiceListOptions) run() ([]unstructured.Unstructured, error) {
+	services := []unstructured.Unstructured{}
+
+	if o.allNamespacesFlag {
+		projects, err := o.clientset.ProjectClient.List()
+		if err != nil {
+			return services, err
```
feloy, line 113: Why returning services and not nil as usual? It would be more readable

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (o *ServiceListOptions) RunForJsonOutput(_ context.Context) (out interface{}, err error) {
+	services, err := o.run()
+	if err != nil {
+		return nil, err
+	}
+
+	var servicesList []api.BindableService
+	for _, svc := range services {
+		servicesList = append(servicesList, api.BindableService{Name: svc.GetName(), Namespace: svc.GetNamespace()})
```
feloy, line 145: Also this could be done into `run`.

To go further, you could also do the embedding into the `bindableServices` field.

The "perfect" `RunForJsonOutput` function being, IMO, this:

```
func (lo *ListOptions) RunForJsonOutput(ctx context.Context) (out interface{}, err error) {
	return lo.run(ctx)
}
```

feloy (changes requested):

pkg/odo/cli/list/services/services.go
dharmit: Are you suggesting `oc get redis -A` instead of `oc get redis -n somens`?

pkg/odo/cli/list/services/services.go
feloy: I still don't understand how the list could contain a single "zero" Unstructured element. When would it be added?

feloy (changes requested):

pkg/odo/cli/list/services/services.go
feloy: yes. But finally, I'm not sure the user will get the rights to run such request :( On Openshift, it won't. So let's keep it as it is

dharmit: Yes, that's the problem I saw. But I wanted to be sure what you were suggesting before saying that it doesn't work on CRC.

docs/website/docs/command-reference/list-services.md
```diff
@@ -0,0 +1,52 @@
...
+```
+```shell
+$ odo list services -n newproject
+ ✓  Listing bindable services from namespace "newproject" [45ms]
+
+ NAME                                      NAMESPACE  
+ hello-world/RabbitmqCluster.rabbitmq.com  newproject 
+```
+
+To get the JSON formatted output for any of the above commands, add `-o json` to the commands shown above. That
```
feloy, line 48: Also please add content in the `json-output.md` file, with an example of output, so the reader can see the structure of the output from the doc.

feloy (changes requested):

pkg/odo/cli/list/services/services.go
dharmit: I take it back. TBH, I found it weird too. Anyway, it should be fixed now.

pkg/odo/cli/list/services/services.go
dharmit: Makes sense. Done.

pkg/odo/cli/list/services/services.go
dharmit: Makes sense. But now it's different, still not `nil`. PTAL.

pkg/odo/cli/list/services/services.go
dharmit: I liked this idea. Done.

docs/website/docs/command-reference/list-services.md
dharmit: Added.

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,183 @@
...
+			return api.ResourcesList{}, err
+		}
+		for k := range svcs {
+			services = append(services, svcs[k])
+		}
+	}
+
+	var servicesList []api.BindableService
+	for _, svc := range services {
+		servicesList = append(servicesList, api.BindableService{
```
feloy, line 135: Why return this compacted name? Is it imposed?

It would be more open for user to have separated info (namespace, name, kind, group), or at least (namespace, name, kind+group).

dharmit: Yeah, that's from the acceptance criteria.

feloy: IIUC, this compacted name is the value that should be passed as service to `odo add binding --service myservice/Redis.redis.redis.opstreelab.in --name myRedisService`. 


If it is the case, it would be interesting to have specific a function for building this string near the definition of `parseServiceName` in `pkg/binding/backend`

Also, it would be more understandable if the field is named `service` instead of `name`, to indicate it is the value of the `--service` flag.

To sum up, I think having those fields would be interesting:
- service: the value to pass to `add binding`
- name: the name of the instance
- namespace: 
- kind: the kind of the resource
- group/version: the group/version of the resource

So the plugin can display any of the name / namespace / kind / groupVersion to the user for them to choose one, and use service for `add binding`

dharmit: @kadel PTAL above comment and suggest.

kadel: > Also, it would be more understandable if the field is named service instead of name, to indicate it is the value of the --service flag.

This is good point. `service` makes more sense here.


> To sum up, I think having those fields would be interesting:
> 
> service: the value to pass to add binding
> name: the name of the instance
> namespace:
> kind: the kind of the resource
> group/version: the group/version of the resource


We can definitely do this for JSON  output.
```json
[ {
    "name": "hippo",
    "service": "hippo/PerconaXtraDBCluster.pxc.percona.com",
    "kind": "PerconaXtraDBCluster",
    "group": "pxc.percona.com",
    "namespace": "asdf"
},{

    "name": "cluster1",
    "service": "cluster1.PostgresCluster.postgres-operator.crunchydata.com",
    "kind": "PostgresCluster",
    "group": "postgres-operator.crunchydata.com",
    "namespace": "asdf"
}]
```

but for normal output, I'm afraid that it will be too many columns.

```
 NAME      SERVICE                                                     KIND                  GROUP                               NAMESPACE
 hippo     hippo/PerconaXtraDBCluster.pxc.percona.com                  PerconaXtraDBCluster  postgres-operator.crunchydata.com   asdf
 cluster1  cluster1/PostgresCluster.postgres-operator.crunchydata.com  PostgresCluster       postgres-operator.crunchydata.com   asdf
```

For normal output I would prefer keeping as it is, only renaming name to service:

```
 SERVICE                                                     NAMESPACE
 hippo/PerconaXtraDBCluster.pxc.percona.com                  asdf
 cluster1/PostgresCluster.postgres-operator.crunchydata.com  asdf
```

Or we could consider both NAME, and SERVICE, 
```
 NAME      SERVICE                                                    NAMESPACE
 hippo     hippo/PerconaXtraDBCluster.pxc.percona.com                 asdf
 cluster1  cluster1/PostgresCluster.postgres-operator.crunchydata.com asdf
```

feloy: For me, it is also interesting to have this `service` value in the JSON output, so the plugin does not have to compute this value when it wants to call `add binding` (I consider this as an "opaque token")

dharmit: I haven't added the `service` field to the commit I pushed. IMO the CLI layer and plugin should handle it by themselves instead. Here's the output:
```sh
$ odo list services -A
 ✓  Listing bindable services from all namespaces [57ms]

 NAME                                                   NAMESPACE  
 cluster-sample/Cluster.postgresql.k8s.enterprisedb.io  myproject  
 hello-world/RabbitmqCluster.rabbitmq.com               newproject 

$ odo list services -A -o json
{
	"bindableServices": [
		{
			"name": "cluster-sample",
			"namespace": "myproject",
			"kind": "Cluster",
			"group": "postgresql.k8s.enterprisedb.io"
		},
		{
			"name": "hello-world",
			"namespace": "newproject",
			"kind": "RabbitmqCluster",
			"group": "rabbitmq.com"
		}
	]
}
```

feloy: I don't agree, at it is error prone for the plugin to build this string by themselves (and not really their responsability).

dharmit: > For me, it is also interesting to have this `service` value in the JSON output, so the plugin does not have to compute this value when it wants to call `add binding` (I consider this as an "opaque token")

For me, this is precisely the reason we should not do it. I know that `run` function is at CLI layer, not at business layer, but producing JSON outputs that require consumers (here plugins) to not have to do anything except use a specific field is a bad idea. It would set a precedent for further requests.

If we would still like to do it, fine by me.

dharmit: > I don't agree, at it is error prone for the plugin to build this string by themselves (and not really their responsability).

I disagree. It's only string concatenation that needs to be done here. How could that be error-prone?

About responsibility, I think that JSON outputs should help the user have everything they need, which is surely the case here. By putting the service name in JSON output, we are spoon-feeding the users to not have to do anything at all. And that, in my opinion, is a bad precedent which could easily lead to more such requests from the users.

feloy: I would like to understand your argument. It is not clear for me.

Mine is that a data structure generally contains a unique (and sometimes opaque) ID (in databases, in JSON/YAML structures like devfile, etc). Other data is here for display usage or for other usage, but not for refeencing the element. The unique ID is the only field used for this identification.

Here, the service would play the role of the unique ID.

kadel: >By putting the service name in JSON output, we are spoon-feeding the users to not have to do anything at all.

We actually should spoon-feed them as much as we can (where it makes sense).


For me, the biggest argument is that if we show the full service name (service identifier) as it is required by `odo link` command, it will allow us to change the way how it is generated without breaking any integrations build on top of that.

Another argument is that the way how the service identifier is constructed is odo specific thing, as such it needs to be handled by odo, if we let plugins construct it on their side, then we are leaking odo internal logic.

If the service identifier string would be some kind of standard that is well defined and used somewhere else. Then it would make sense to just provide all required info and let others construct it on their own. But that is not the case here.

dharmit: TL;DR - 501e5a0

Responses below.

> if we show the full service name (service identifier) as it is required by `odo link` command, it will allow us to change the way how it is generated without breaking any integrations build on top of that.
> 
> Another argument is that the way how the service identifier is constructed is odo specific thing, as such it needs to be handled by odo, if we let plugins construct it on their side, then we are leaking odo internal logic.

I was thinking exactly this after my last response about concatenation and having every piece of info already there in the JSON output. And it inclined me to put the service field in there, as @feloy suggested.

> We actually should spoon-feed them as much as we can (where it makes sense).

I look at users as the odo CLI users and the plugin devs. For the first, I agree to spoon-feed and beyond! But for plugin devs, I think we should draw a line. However, in this case, it makes sense to me to put the service field.

> Mine is that a data structure generally contains a unique (and sometimes opaque) ID (in databases, in JSON/YAML structures like devfile, etc). Other data is here for display usage or for other usage, but not for refeencing the element. The unique ID is the only field used for this identification.
> 
> Here, the service would play the role of the unique ID.

I think you are saying this based on your experience and/or a paradigm you have come across. I wasn't looking at the idea of putting the service field in this way. Thanks for this info.

pkg/odo/cli/list/services/services.go
```diff
@@ -0,0 +1,188 @@
...
+	%[1]s --all-namespaces
+	%[1]s -A
+
+	# List the bindable Operator backed services in JSON format
+	%[1]s -o json
+	%[1]s --all-namespaces -o json
+	%[1]s -A -o json`)
+
+	listLongDesc = ktemplates.LongDesc(`
+	List the bindable Operator backed services that could bound to the odo component
```
rm3l, line 39:
```suggestion
	List the bindable Operator backed services that could be bound to the odo component 
```

tests/integration/cmd_list_services.go
```diff
@@ -0,0 +1,67 @@
+package integration
```
rm3l, line 1: Can you rename this file into `cmd_list_services_test.go`, just to be consistent with how the other files are named?

tests/integration/cmd_list_services.go
```diff
@@ -0,0 +1,67 @@
...
+		// from current namespace
+		out := helper.Cmd("odo", "list", "services", "-n", commonVar.Project).ShouldPass().Out()
+		helper.MatchAllInOutput(out, []string{"cluster-sample", commonVar.Project, "Listing bindable services from namespace"})
+
+		// from all namespaces
+		out = helper.Cmd("odo", "list", "services", "-A").ShouldPass().Out()
+		helper.MatchAllInOutput(out, []string{"cluster-sample", commonVar.Project, randomProject, "Listing bindable services from all namespaces"})
+
+		// fail if -A and -n flags are used together
+		out = helper.Cmd("odo", "list", "services", "-A", "-n", commonVar.Project).ShouldFail().Err()
```
rm3l, line 47: Can this be a separate `It` spec instead?

tests/integration/cmd_list_services.go
```diff
@@ -0,0 +1,67 @@
...
+		Expect(gjson.Get(out, "bindableServices.0.namespace").String()).To(Equal(commonVar.Project))
+		Expect(gjson.Get(out, "bindableServices.0.kind").String()).To(Equal("Cluster"))
+		Expect(gjson.Get(out, "bindableServices.0.group").String()).To(Equal("postgresql.k8s.enterprisedb.io"))
+
+		// from all namespaces
+		out = helper.Cmd("odo", "list", "services", "-A", "-o", "json").ShouldPass().Out()
+		helper.MatchAllInOutput(out, []string{"cluster-sample", commonVar.Project, randomProject, "Cluster", "postgresql.k8s.enterprisedb.io"})
+
+		// fail if -A and -n flags are used together
+		out = helper.Cmd("odo", "list", "services", "-o", "json", "-A", "-n", commonVar.Project).ShouldFail().Err()
```
rm3l, line 64: Same point about extracting this to a separate `It` spec..

tests/integration/cmd_list_services.go
```diff
@@ -0,0 +1,67 @@
...
+	It("should list bindable services in JSON format", func() {
+		// from current namespace
+		out := helper.Cmd("odo", "list", "services", "-o", "json", "-n", commonVar.Project).ShouldPass().Out()
+		Expect(gjson.Get(out, "bindableServices.0.name").String()).To(ContainSubstring("cluster-sample"))
+		Expect(gjson.Get(out, "bindableServices.0.namespace").String()).To(Equal(commonVar.Project))
+		Expect(gjson.Get(out, "bindableServices.0.kind").String()).To(Equal("Cluster"))
+		Expect(gjson.Get(out, "bindableServices.0.group").String()).To(Equal("postgresql.k8s.enterprisedb.io"))
+
+		// from all namespaces
+		out = helper.Cmd("odo", "list", "services", "-A", "-o", "json").ShouldPass().Out()
```
rm3l, line 60: I think it would be also help to ensure that `out` is a valid JSON string with: `Expect(helper.IsJSON(out)).To(BeTrue())`

tests/integration/cmd_list_services.go
```diff
@@ -0,0 +1,67 @@
...
+
+		// fail if -A and -n flags are used together
+		out = helper.Cmd("odo", "list", "services", "-A", "-n", commonVar.Project).ShouldFail().Err()
+		Expect(out).To(ContainSubstring("cannot use --all-namespaces and --namespace flags together"))
+	})
+
+	It("should list bindable services in JSON format", func() {
+		// from current namespace
+		out := helper.Cmd("odo", "list", "services", "-o", "json", "-n", commonVar.Project).ShouldPass().Out()
+		Expect(gjson.Get(out, "bindableServices.0.name").String()).To(ContainSubstring("cluster-sample"))
```
rm3l, line 54: It is okay as it is, but I think you could have used the various `JsonPath*` helper functions that are already there, like `helper.JsonPathContentContain` or `helper.JsonPathContentIs`

rm3l (changes requested):
LGTM feature-wise. Added a few small comments.

docs/website/docs/command-reference/list-services.md
```diff
@@ -0,0 +1,52 @@
...
+odo list services -A 
+```
+```shell
+odo list services -A
+ ✓  Listing bindable services from all namespaces [182ms]
+
+ NAME                                               NAMESPACE  
+ redis-standalone/Redis.redis.redis.opstreelabs.in  myproject  
+ hello-world/RabbitmqCluster.rabbitmq.com           newproject 
+```
```
valaparthvi, line 34: Do we want to display the output for commands? Do you find it helpful? I ask this because I am working on making the docs consistent, and I can't decide if having this output is helpful or not. If we keep this, we'll need to ensure the output is up to date.

tests/integration/cmd_list_services.go
dharmit: I'd prefer to keep it here because a separate `It` spec would mean more time to spin up and tear things down. I wouldn't mind doing it if there was something where we were testing more stuff. Here, it's quite small to put in a separate `It`, IMO.

tests/integration/cmd_list_services.go
dharmit: Same reasoning as https://github.com/redhat-developer/odo/pull/6118#discussion_r971889106.

docs/website/docs/command-reference/json-output.md
```diff
@@ -656,3 +656,58 @@ odo preference view -o json
 	]
 }
 ```
+
+## odo list services -o json
+
+The `odo list services` command lists all the bindable Opeartor backed services available in the current
```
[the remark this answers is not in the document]
dharmit: 503614f

pkg/odo/cli/list/services/services.go
dharmit: 503614f

tests/integration/cmd_list_services.go
dharmit: 503614f

tests/integration/cmd_list_services.go
dharmit: Thanks for the tip. I'll keep it in mind. 👍🏾

docs/website/docs/command-reference/list-services.md
dharmit: I think they are helpful and would rather have the output here. But if it's a universal thing we are doing across the docs, I am OK to remove it.

rm3l (commented):
LGTM feature-wise.
/approve

Will re-add the lgtm label Philippe had put previously.

rm3l:
Following test failed on Windows because of `error looking up service account cmd-logs-test189bph/default: serviceaccount "default" not found`. Will need to relaunch them to see if it happens again:

```
+ [FAILED] [497.821 seconds]
odo logs command tests
C:/Users/Administrator.ANSIBLE-TEST-VS/2007/tests/integration/cmd_logs_test.go:14
  when component is created and odo logs is executed
  C:/Users/Administrator.ANSIBLE-TEST-VS/2007/tests/integration/cmd_logs_test.go:69
    when running in both Dev and Deploy mode [BeforeEach]
    C:/Users/Administrator.ANSIBLE-TEST-VS/2007/tests/integration/cmd_logs_test.go:177
      should successfully show logs of the running component
      C:/Users/Administrator.ANSIBLE-TEST-VS/2007/tests/integration/cmd_logs_test.go:189
...
[odo]  X  failed to create Kubernetes resources associated with the component: pods "myapp-inner" is forbidden: error looking up service account cmd-logs-test189bph/default: serviceaccount "default" not found
    [odo] Cleaning resources, please wait
...
Timed out after 360.012s.
  Expected
      <string>:   __
       /  \__     Developing using the mbysyt Devfile
       \__/  \    Namespace: cmd-logs-test189bph
       /  \__/    odo version: v3.0.0-rc1
       \__/
      
      - Deploying to the cluster in developer mode
       -  Waiting for Kubernetes resources  ...
      
  to contain substring
      <string>: [Ctrl+c] - Exit
  In [BeforeEach] at: C:/Users/Administrator.ANSIBLE-TEST-VS/2007/tests/helper/helper_run.go:40
```

rm3l:
However, on Prow, there is an interesting error related to the new tests added in this PR, because the user does not have the required permissions:

```
• [FAILED] [4.620 seconds]
odo list services tests
/go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:10
  [It] should list bindable services in human readable format
  /go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:37
...
Running odo with args [odo list services -A]
...
[odo]  •  Listing bindable services from all namespaces  ...
[odo]  ✗  Listing bindable services from all namespaces [98ms]
    [odo]  ✗  clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"
    Deleting project: cmd-list-services-test37xgo
    Running odo with args [odo delete project cmd-list-services-test37xgo -f --wait=false]
Running odo with args [odo list services -A]
  Expected
      <int>: 1
  to match exit code:
      <int>: 0
  In [It] at: /go/src/github.com/redhat-developer/odo/tests/helper/helper_cmd_wrapper.go:101
```

tests/integration/cmd_list_services.go
dharmit: > Yes, I was thinking about having this test in its own container node, which would not have all the setup/teardown code.
> This would help in terms of reporting and also for test parallelization, because I think we are testing independent things here: testing that "-A" and "-n" cannot be used together should be run even if the "odo list services -A" command above is returning the wrong data.

Sounds like a good thing. Is it the `By` directive of ginkgo that helps do this? Can you point me to an example if there's already one in the odo code?

dharmit:
> [odo]  •  Listing bindable services from all namespaces  ...
> [odo]  ✗  Listing bindable services from all namespaces [98ms]
> [odo]  ✗  clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"

This one needs to be looked into. I had faced this on and off on my CRC, but not always. So I pushed the code anyway. Removing lgtm label for the time being.

dharmit:
> However, on Prow, there is an interesting error related to the new tests added in this PR, because the user does not have the required permissions:
> 
> ```
> • [FAILED] [4.620 seconds]
> odo list services tests
> /go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:10
>   [It] should list bindable services in human readable format
>   /go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:37
> ...
> Running odo with args [odo list services -A]
> ...
> [odo]  •  Listing bindable services from all namespaces  ...
> [odo]  ✗  Listing bindable services from all namespaces [98ms]
> [odo]  ✗  clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"
> Deleting project: cmd-list-services-test37xgo
> Running odo with args [odo delete project cmd-list-services-test37xgo -f --wait=false]
> Running odo with args [odo list services -A]
>   Expected
>       <int>: 1
>   to match exit code:
>       <int>: 0
>   In [It] at: /go/src/github.com/redhat-developer/odo/tests/helper/helper_cmd_wrapper.go:101
> ```
> 
> What can we do about this?

While debugging this one, I found that the namespace for which we get the error `✗  clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"` is the namespace that's `Terminating` and belongs to a different test spec, not the one which is being executed. So basically a race condition, IMO.

As a fix, I have pushed e0a7254 which checks the project/namespace's `.status.phase` before appending it to the slice. PTAL @rm3l @feloy.

feloy:
> > However, on Prow, there is an interesting error related to the new tests added in this PR, because the user does not have the required permissions:
> > ```
> > • [FAILED] [4.620 seconds]
> > odo list services tests
> > /go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:10
> >   [It] should list bindable services in human readable format
> >   /go/src/github.com/redhat-developer/odo/tests/integration/cmd_list_services_test.go:37
> > ...
> > Running odo with args [odo list services -A]
> > ...
> > [odo]  •  Listing bindable services from all namespaces  ...
> > [odo]  ✗  Listing bindable services from all namespaces [98ms]
> > [odo]  ✗  clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"
> > Deleting project: cmd-list-services-test37xgo
> > Running odo with args [odo delete project cmd-list-services-test37xgo -f --wait=false]
> > Running odo with args [odo list services -A]
> >   Expected
> >       <int>: 1
> >   to match exit code:
> >       <int>: 0
> >   In [It] at: /go/src/github.com/redhat-developer/odo/tests/helper/helper_cmd_wrapper.go:101
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What can we do about this?
> 
> While debugging this one, I found that the namespace for which we get the error `✗ clusters.postgresql.k8s.enterprisedb.io is forbidden: User "developer" cannot list resource "clusters" in API group "postgresql.k8s.enterprisedb.io" in the namespace "cmd-dev-test183uvq"` is the namespace that's `Terminating` and belongs to a different test spec, not the one which is being executed. So basically a race condition, IMO.
> 
> As a fix, I have pushed [e0a7254](https://github.com/redhat-developer/odo/commit/e0a72541205796e77c04549c6a35e55c9838c5d6) which checks the project/namespace's `.status.phase` before appending it to the slice. PTAL @rm3l @feloy.

That won't fix the problem, as the namespace can still be deleted between the time you execute the "List namespaces" and you execute the "List Clusters" in this namespace (and because we are executing a lot of List Clusters, one for all namespaces, the interval can be quite long).

I think you will need to ignore some errors when running "List Clusters" in a namespace. To double check, you could test afterward that the namespace is not in Active mode.

dharmit:
> That won't fix the problem, as the namespace can still be deleted between the time you execute the "List namespaces" and you execute the "List Clusters" in this namespace

I imagine this being the case with our integration tests in particular. Or are you referring to some other scenario as well?

> I think you will need to ignore some errors when running "List Clusters" in a namespace. To double check, you could test afterward that the namespace is not in Active mode.

OK. Let me see if I can try ignoring the error we are seeing here.

feloy:
> > That won't fix the problem, as the namespace can still be deleted between the time you execute the "List namespaces" and you execute the "List Clusters" in this namespace
> 
> I imagine this being the case with our integration tests in particular. Or are you referring to some other scenario as well?
> 

Yes, this will happen during integration tests. And so, could happen in other circumstances

pkg/binding/binding.go
```diff
@@ -91,6 +91,10 @@ func (o *BindingClient) GetServiceInstances(namespace string) (map[string]unstru
 		// List all the instances of the restMapping object
 		resources, err := o.kubernetesClient.ListDynamicResources(namespace, restMapping.Resource)
 		if err != nil {
+			if kerrors.IsForbidden(err) {
```
feloy, line 94: You probably also need to check if IsNotFound, as the namespace could be completely deleted

tests/integration/cmd_list_services.go
rm3l: [`By`](https://onsi.github.io/ginkgo/#documenting-complex-specs-by) is syntactic sugar that generates "STEP: " in reports, to better document complex specs. I think this could have been used for example to separate the checks against the current namespace and all namespaces.

I was thinking of having those tests under a completely different `Context` or `Describe` container node, like:
```go
var _ = Describe("odo list services tests", func() {

	Context("passing the wrong arguments", func() {
		var commonVar helper.CommonVar

		var _ = BeforeEach(func() {
            // no need to create a project
			commonVar = helper.CommonBeforeEach(helper.SetupClusterFalse)
			helper.Chdir(commonVar.Context)
		})

		var _ = AfterEach(func() {
			helper.CommonAfterEach(commonVar)
		})

		It("should fail if -A and -n flags are used together", func() {
			By("human-readable output", func() {
				out := helper.Cmd("odo", "list", "services", "-A", "-n", "any-namespace").ShouldFail().Err()
				Expect(out).To(ContainSubstring("cannot use --all-namespaces and --namespace flags together"))
			})
			By("JSON output", func() {
				out := helper.Cmd("odo", "list", "services", "-A", "-n", "any-namespace", "-o", "json").ShouldFail().Err()
				Expect(out).To(ContainSubstring("cannot use --all-namespaces and --namespace flags together"))
			})
		})
	})

	Context("passing the right arguments", func() {
      // -- Truncated : the other tests --
```

There should be some examples in odo (like in `cmd_dev_test.go`), but I think they are also nested in other container nodes.

dharmit:
> Yes, this will happen during integration tests. And so, could happen in other circumstances

Interested in knowing what other circumstances this could be. Just curious. I think a user is rarely (if ever) going to face this unless they use odo for some automation.

Anyhow, PTAL f901fc7.

feloy:
> > Yes, this will happen during integration tests. And so, could happen in other circumstances
> 
> Interested in knowing what other circumstances this could be. Just curious. I think a user is rarely (if ever) going to face this unless they use odo for some automation.
> 
> Anyhow, PTAL [f901fc7](https://github.com/redhat-developer/odo/commit/f901fc70661c39930519ed9b34db7232b332de7e).

On a cluster with multiple users, you cannot know is someone else or some other automation is creating and deleting namespaces, and at which rate. In this circumstance, you will easily hit the problem.

pkg/binding/binding.go
rm3l: Would it be possible to check this differently? Or maybe display the error to the user..

With this, if I run `odo add binding` with a user that does not have the required permissions to list the corresponding resources, I have this error:

```
❯ odo add binding     
? Do you want to list services from: current namespace
 ✗  No bindable service instances found in current namespace
```

On `main`, the error returned was much more explicit to me:
```
❯ odo add binding
? Do you want to list services from: current namespace
 ✗  perconaservermongodbs.psmdb.percona.com is forbidden: User "user" cannot list resource "perconaservermongodbs" in API group "psmdb.percona.com" in the namespace "user-ns"
```

rm3l: Or logging the error with `klog` could be sufficient to help with troubleshooting.

dharmit: `klog` sounds more appropriate to me. I have added it.

dharmit:
OpenShift test failure on IBM Cloud:
```
 [FAILED] [264.176 seconds]
odo dev command tests
/go/odo_1/tests/integration/cmd_dev_test.go:29
  /go/odo_1/tests/integration/cmd_dev_test.go:469
    when devfile has single endpoint
    /go/odo_1/tests/integration/cmd_dev_test.go:470
      when running odo dev
      /go/odo_1/tests/integration/cmd_dev_test.go:477
        when modifying memoryLimit for container in Devfile [BeforeEach]
        /go/odo_1/tests/integration/cmd_dev_test.go:507
          should expose the endpoint on localhost
          /go/odo_1/tests/integration/cmd_dev_test.go:523

...
...
[odo] I0916 12:16:20.293554   20045 remotecmd.go:46] ExecuteCommand returned an an err: error while streaming command: command terminated with exit code 137. for command '[/bin/sh -c echo $$ > /opt/odo/.odo_cmd_devrun.pid && cd ${PROJECTS_ROOT} &&  (npm start) 1>>/proc/1/fd/1 2>>/proc/1/fd/2; echo $? >> /opt/odo/.odo_cmd_devrun.pid]'
    [odo] stdout: []
    [odo] stderr: []
    [odo] I0916 12:16:20.293606   20045 kubeexec.go:103] error while running background command: unable to exec command [/bin/sh -c echo $$ > /opt/odo/.odo_cmd_devrun.pid && cd ${PROJECTS_ROOT} &&  (npm start) 1>>/proc/1/fd/1 2>>/proc/1/fd/2; echo $? >> /opt/odo/.odo_cmd_devrun.pid]: error while streaming command: command terminated with exit code 137
    [odo] I0916 12:16:20.293625   20045 kubeexec.go:40] GetProcessInfoForCommand for "devrun"
    [odo] I0916 12:16:20.293666   20045 remotecmd.go:30] Executing command [/bin/sh -c cat /opt/odo/.odo_cmd_devrun.pid || true] for pod: uxhnpk-app-6c79d857b7-zp4zr in container: runtime
    [odo]  ✗  Finished executing the application (command: devrun) [3m]
    [odo] I0916 12:16:20.644393   20045 remotecmd.go:46] ExecuteCommand returned an an err: error while streaming command: unable to upgrade connection: container not found ("runtime"). for command '[/bin/sh -c cat /opt/odo/.odo_cmd_devrun.pid || true]'
    [odo] stdout: []
    [odo] stderr: []
    [odo] I0916 12:16:20.644519   20045 commandhandler.go:60] error while running background command: unable to exec command [/bin/sh -c echo $$ > /opt/odo/.odo_cmd_devrun.pid && cd ${PROJECTS_ROOT} &&  (npm start) 1>>/proc/1/fd/1 2>>/proc/1/fd/2; echo $? >> /opt/odo/.odo_cmd_devrun.pid]: error while streaming command: command terminated with exit code 137
    Deleting project: cmd-dev-test523yim
    Running odo with args [odo delete project cmd-dev-test523yim -f --wait=false]
    [odo] I0916 12:16:30.769003   29141 util.go:754] Response will be cached in /tmp/odohttpcache for 1h0m0s
    [odo] I0916 12:16:30.769691   29141 util.go:767] Cached response used.
    [odo] I0916 12:16:30.934617   29141 component.go:73] name of component is "1378735227", and sanitized name is "x1378735227"
    [odo] I0916 12:16:30.934656   29141 kclient.go:207] Checking if "projects" resource is supported
    [odo]  ✓  Project "cmd-dev-test523yim" deleted
    [odo] I0916 12:16:30.984391   29141 odo.go:75] Could not get the latest release information in time. Never mind, exiting gracefully :)
    Deleting dir: /tmp/1378735227
    Deleting dir: /tmp/928477368
  << End Captured GinkgoWriter Output

  Timed out after 180.000s.
  Expected
      <string>:  ⚠  Pod is Terminating
       ✗  Finished executing the application (command: devrun) [41s]
       ⚠  No pod exists
       ⚠  Pod is Pending
       ✓  Pod is Running
       •  Syncing files into the container  ...
      
 ✓  Syncing files into the container [668ms]
      cat: /opt/odo/.odo_cmd_devrun.pid: No such file or directory
       •  Building your application in container on cluster (command: devbuild)  ...
      
 ✓  Building your application in container on cluster (command: devbuild) [5s]
       •  Executing the application (command: devrun)  ...
      46
      0
      46
      0
       -  Forwarding from 127.0.0.1:42375 -> 3000
      
      
      ↪ Dev mode
       Keyboard Commands:
      
  To satisfy at least one of these matchers: [%!s(*matchers.ContainSubstringMatcher=&{Pushing files... []}) %!s(*matchers.ContainSubstringMatcher=&{Updating Component... []})]
  In [BeforeEach] at: /go/odo_1/tests/helper/helper_run.go:53
```
I think this is unrelated to this PR. Probably a flake.

rm3l:
```
• [FAILED] [120.321 seconds]
odo init interactive command tests
/go/src/github.com/redhat-developer/odo/tests/integration/interactive_init_test.go:20
  [It] should download correct devfile
  /go/src/github.com/redhat-developer/odo/tests/integration/interactive_init_test.go:95

...
✗  Downloading devfile "go" from registry "DefaultDevfileRegistry" [102ms]
   ✗  unable to download devfile: failed to pull stack go from registry.stage.devfile.io/devfile-catalog/go:1.0.0 with allowed media types [application/vnd.devfileio.devfile.layer.v1 image/png image/svg+xml application/vnd.devfileio.vsx.layer.v1.tar application/x-tar]: unexpected status code [manifests 1.0.0]: 502 Bad Gateway
  the command failed, the devfile has been removed from current directory
```
Looks like a temporary network error with this test. Overriding since the tests for "odo list services" are now passing on Prow and IBM Cloud.

PR #5556, opened 2022-03-15
Make `odo dev` and `odo deploy` display welcoming messages too before running Alizer detection, just like `odo init`

rm3l:
**What type of PR is this:**
/kind feature
/area ux

**What does this PR do / why we need it:**
This relates to #5494, because users might have similar symptoms (few seconds before any output) when running either `odo dev` or `odo deploy` in a source code directory with no Devfile, and those commands try to do an `Alizer` analysis first.

**Which issue(s) this PR fixes:**
Relates to #5494

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

**How to test changes / Special notes to the reviewer:**
Run either `odo dev` or `odo deploy` from a source code directory with no Devfile. A welcoming message should be displayed before the Alizer detection messages are showed.

PR #6127, opened 2022-09-15
 Add extra output when init project using odo dev or deploy

cdrage:
Add extra output when init project using odo dev

**What type of PR is this:**

/kind cleanup

Updates the messaging when you run `odo dev` and there is no
`devfile.yaml` available. Outputs more information that makes it similar
to running `odo dev` or `odo init`

Fixes https://github.com/redhat-developer/odo/issues/5679

### DIFF
--- a/pkg/odo/cli/messages/messages.go
+++ b/pkg/odo/cli/messages/messages.go
@@ -5,3 +5,4 @@ const InteractiveModeEnabled = "Interactive mode enabled, please answer the foll
 const InitializingNewComponent = "Initializing new component"
 const SourceCodeDetected = "Files: Source code detected, a Devfile will be determined based upon source code autodetection"
 const NoSourceCodeDetected = "Files: No source code detected, a starter project will be created in the current directory"
+const InitializeExistingComponent = "Dev mode ran, but no Devfile was found. Initializing a Devfile in the current directory"
--- a/pkg/odo/cli/messages/messages.go
+++ b/pkg/odo/cli/messages/messages.go
@@ -2,8 +2,8 @@
 package messages
 
 const InteractiveModeEnabled = "Interactive mode enabled, please answer the following questions:"
-const InitializingNewComponent = "Initializing new component"
+const InitializingNewComponent = "Initializing a new component"
 const SourceCodeDetected = "Files: Source code detected, a Devfile will be determined based upon source code autodetection"
 const NoSourceCodeDetected = "Files: No source code detected, a starter project will be created in the current directory"
-const DevInitializeExistingComponent = "Dev mode ran, but no Devfile was found. Initializing a Devfile in the current directory"
-const DeployInitializeExistingComponent = "Deploy mode ran, but no Devfile was found. Initializing a Devfile in the current directory"
+const DevInitializeExistingComponent = "Dev mode ran, but no Devfile was found. Initializing a component in the current directory"
+const DeployInitializeExistingComponent = "Deploy mode ran, but no Devfile was found. Initializing a component in the current directory"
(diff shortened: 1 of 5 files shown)


pkg/odo/cli/messages/messages.go
```diff
@@ -5,3 +5,4 @@ const InteractiveModeEnabled = "Interactive mode enabled, please answer the foll
 const InitializingNewComponent = "Initializing new component"
 const SourceCodeDetected = "Files: Source code detected, a Devfile will be determined based upon source code autodetection"
 const NoSourceCodeDetected = "Files: No source code detected, a starter project will be created in the current directory"
+const InitializeExistingComponent = "Dev mode ran, but no Devfile was found. Initializing a Devfile in the current directory"
```
rm3l, line 8: This constant is not used and could be removed.

rm3l (changes requested):
I think we currently have the same kind of issue when running `odo deploy` in a dir with no devfile.yaml:
```
❯ odo deploy                                                                                
The current directory already contains source code. odo will try to autodetect the language and project type in order to select the best suited Devfile for your project.
Language: go
Project type: go
The devfile "go" from the registry "DefaultDevfileRegistry" will be downloaded.
? Is this correct? (Y/n)
```

To be consistent, can you do the same thing for `odo deploy` as well?

rm3l:
Also, I think there are some tests (introduced in #5556)  that would need to be fixed/rewritten: they were testing that we had the previous messages displayed first in the output.

cdrage:
> I think we currently have the same kind of issue when running `odo deploy` in a dir with no devfile.yaml:
> 
> ```
> ❯ odo deploy                                                                                
> The current directory already contains source code. odo will try to autodetect the language and project type in order to select the best suited Devfile for your project.
> Based on the files in the current directory odo detected
> Language: go
> Project type: go
> The devfile "go" from the registry "DefaultDevfileRegistry" will be downloaded.
> ? Is this correct? (Y/n)
> ```
> 
> To be consistent, can you do the same thing for `odo deploy` as well?

Good point, I'll do the same for `odo deploy`

tests/integration/interactive_dev_test.go
```diff
@@ -142,8 +142,7 @@ var _ = Describe("odo dev interactive command tests", func() {
 			lines, err := helper.ExtractLines(output)
 			Expect(err).To(BeNil())
 			Expect(lines).To(Not(BeEmpty()))
-			Expect(lines[0]).To(Equal("The current directory already contains source code. " +
-				"odo will try to autodetect the language and project type in order to select the best suited Devfile for your project."))
+			Expect(lines[1]).To(ContainSubstring("Dev mode ran, but no Devfile was found. Initializing a Devfile in the current directory"))
```
valaparthvi, line 145: `odo init` says "Initializing new component" when downloading devfile, while `odo dev` and `odo deploy` say "Initializing a Devfile". Can we make these messages more consistent?

valaparthvi (changes requested):
While you're at this, can we make an additional change to `odo dev` and `odo deploy` messages?
When there is a devfile present and you run either of the commands, it says "Developing using the devfile-nodejs-deploy Devfile", can we double-quote the devfile name? I always find it very confusing to determine the component/devfile name when these commands are run.

cdrage:
Thanks @valaparthvi I've updated it based on your review and also made changes to the Devfile output / added double quotes.

valaparthvi:
/lgtm
Thanks for patiently working on all the changes, Charlie. :)

rm3l (commented):
Thanks for the changes.

/approve

feloy:
Pass on IBM Cloud

Issue #5865, opened 2022-06-22
[POC] odo without kubernetes cluster / odo on podman


kadel:
/kind user-story

## User Story
As an odo user, I want to be able to run my development environment locally without requiring kubernetes/openshift cluster 
So that 
- I don't need to run the cluster locally and save resources
- I get a quick response (quicker than when using cluster)

## Acceptance Criteria 
- [ ] It should be possible to run `odo dev` on a simple devfile that  doesn't use `kubernetes` components

## more info
It should be possible to leverage podman for local development.
Podman already understands some of the kubernetes resources https://docs.podman.io/en/latest/markdown/podman-play-kube.1.html

/triage needs-more-information

/kind user-story

kadel:
The idea is that odo will try to use existing k8s resources that it generates and just pass it down to the `podman play kube `  command.


- deployments generated by odo can be almost used as it is with two exceptions
	- `hostPort` needs to be added to tell podman to expose port
	- `emptyDir` is not supported by podman and needs to be replaced with PVC
- `Service` is not supported by `podman play kube`. But it is not necessary required to make everything work.


- The most work will be to re-write file syncing logic to work with podman container instead of the container running on the cluster.
- `podman cp` command can be used for transferring files between local filesystem and container

rm3l:
Closing this issue as completed.
Issues created via #6135 and added to the parent epic: https://github.com/redhat-developer/odo/issues/5871

/close

PR #6141, opened 2022-09-20
Refacto odo dev

### LINKED ISSUE #6159: Make pkg/dev platform agnostic
/kind feature

The pkg/dev package is for the moment designed for working on Kubernetes platform only.

Make the package platform agnostic.

## Why is this needed?

So we can reuse it for other platforms (Podman)

feloy:
**What type of PR is this:**

/kind cleanup

Part of #5865 
Fixes #6159 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


valaparthvi (approved):
Nice refactor!

dharmit:
> Fixes #6159

How do we know that the PR fixes this issue? Have we tried this with podman?

feloy:
> > Fixes #6159
> 
> How do we know that the PR fixes this issue? Have we tried this with podman?

If you look at the code into pkg/odo/cli/dev, you should not see any code related to Kubernetes (1).

Also, into the interface definition in pkg/dev, no parameter of the methods should be related to Kubernetes.

(1) There is still a problem about calling:
```
o.clientset.KubernetesClient.SetNamespace(o.GetProject())
```
but this is a generic problem how we are managing the context, not specific to the `odo dev` command.

pkg/dev/dev.go
```diff
@@ -2,20 +2,24 @@ package dev
...
+	"fmt"
 	"io"
 
 	"github.com/redhat-developer/odo/pkg/binding"
 	"github.com/redhat-developer/odo/pkg/envinfo"
 	"github.com/redhat-developer/odo/pkg/kclient"
 	"github.com/redhat-developer/odo/pkg/portForward"
 	"github.com/redhat-developer/odo/pkg/preference"
-	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
+	filesystem "github.com/redhat-developer/odo/pkg/testingutil/filesystem"
```
rm3l, line 13: nit: is this import alias needed? I find it a bit redundant.
https://github.com/redhat-developer/odo/wiki/Dev:-Coding-Conventions#import-aliasing

pkg/dev/interface.go
```diff
@@ -17,44 +16,29 @@ type Client interface {
...
-
-	// Watch watches for any changes to the files under path while ignoring the files/directories in ignorePaths.
+	//
+	// It then watches for any changes to the files under path while ignoring the files/directories in ignorePaths.
 	// It logs messages to out and uses the Handler h to perform push operation when anything changes in path.
 	// It uses devfileObj to notify user to restart odo dev if they change endpoint information in the devfile.
 	// If debug is true, the debug command will be started after a sync, or the run command by default.
 	// If buildCommand is set, this will look up the specified build command in the Devfile. Otherwise, it uses the default one.
 	// If runCommand is set, this will look up the specified run command in the Devfile and execute it. Otherwise, it uses the default one.
 	// componentStatus is the status returned from the call to the Start Method
```
rm3l, line 26: Can you remove the doc about `componentStatus` since it is not a parameter of `Start`?

rm3l (commented):
LGTM - thanks for this!

Added a few nitpicky comments.

pkg/dev/interface.go
```diff
@@ -17,44 +16,29 @@ type Client interface {
...
 		buildCommand string,
 		runCommand string,
-		variables map[string]string,
 		randomPorts bool,
 		watchFiles bool,
+		variables map[string]string,
+		out io.Writer,
 		errOut io.Writer,
-		componentStatus watch.ComponentStatus,
 	) error
```
dharmit, line 42: IMO, we should separate optional parameters into a struct. This is too large a number of arguments for the function, no?

pkg/dev/dev.go
```diff
@@ -25,6 +29,7 @@ type DevClient struct {
 	portForwardClient portForward.Client
 	watchClient       watch.Client
 	bindingClient     binding.Client
+	filesystem        filesystem.Filesystem
 }
 
 var _ Client = (*DevClient)(nil)
```
dharmit, line 35: Could we ensure here that the `Handler` interface is implemented since it's no longer on the CLI but on the business layer?

dharmit (changes requested):

pkg/dev/interface.go
feloy: I have also rewritten the doc

pkg/dev/dev.go
feloy: This Handler is not necessary anymore, removing it.

dharmit:
Thanks for the discussion and the work @feloy.

/lgtm

feloy:
Tests pass on IBM Cloud

PR #6172, opened 2022-09-26
fix new version notification

anandrkskd:

**What type of PR is this:**
/kind bug

This PR fixes the notification for newer version of odo, if available.
 
Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


cmd/odo/odo.go
```diff
@@ -68,11 +68,9 @@ func main() {
 		go version.GetLatestReleaseInfo(updateInfo)
 
 		util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.Background())), "")
-		select {
-		case message := <-updateInfo:
+		message := <-updateInfo
```
rm3l, line 71: This will block until the previous goroutine (`version.GetLatestReleaseInfo`) writes to the `updateInfo` channel. And this depends on some network call.
On slow networks, this will likely block short-running commands, like a simple `odo version --client`. I emulated this by adding a few seconds to my network latency, and now running `odo version --client` is taking some time to return.

I think we should instead give this read operation a timeout (like 1 or 2 seconds) to return the information.

rm3l (changes requested):

build/VERSION
```diff
@@ -1 +1 @@
-v2.5.0
\ No newline at end of file
+v2.5.1
```
kadel, line 1: 👍

cmd/odo/odo.go
anandrkskd: > Version check is optional, odo should never wait for the result. If odo is not able to get information in time before the main functionality ends, it is ok.

Does that mean, no changes required on this PR?

rm3l: > By design, the version check should be on a best-effort basis.
> 
> Essentially if the command is something really quick like `odo version` it is ok that odo won't be able to perform a version check. Version check is optional, odo should never wait for the result. If odo is not able to get information in time before the main functionality ends, it is ok.

Makes sense that it should be a best-effort version check.
After a quick test, I have the feeling that even for long-running operations like `odo dev`, version check is currently not working either..

rm3l: > > Version check is optional, odo should never wait for the result. If odo is not able to get information in time before the main functionality ends, it is ok.
> 
> Does that mean, no changes required on this PR?

Change in build/VERSION file is still needed until v3 is out. Concerning the version check itself, maybe what you did could be run in a separate goroutine, hence not blocking the odo command from returning. If this goroutine has time to display something, great. Otherwise, it will end along with the process if the command is very quick to run.

anandrkskd: > maybe what you did could be run in a separate goroutine,

I am not able to understand which part needs to run in separate goroutine? We are already running the version check in the go routine.

rm3l: > > maybe what you did could be run in a separate goroutine,
> 
> I am not able to understand which part needs to run in separate goroutine? We are already running the version check in the go routine.

The block you updated, which is potentially blocking:

```go
message := <-updateInfo
if message != "" {
    log.Info(message)
}
```

But this should run before the call to `util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.Background())), "")`, which, in my understanding, executes the `odo` command (and can be potentially long-running or very quick to run).

anandrkskd: Won't it delay the `util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.Background())), "")` execution if we move it before?

rm3l: Or you could even simplify everything by making `version.GetLatestReleaseInfo` just return the new tag or an empty string (instead of writing to a channel). Then it would be up to callers to decide how they want to handle the result.
In `cmd/odo/odo.go`, you would spin up a single goroutine, in which you would call `version.GetLatestReleaseInfo` and print the update notification message if needed.
Just how I see it, but feel free to find a different approach.

rm3l: > Won't it delay the `util.LogErrorAndExit(root.ExecuteContext(segment.NewContext(context.Background())), "")` execution if we move it before?

If it is run in a separate goroutine, nope.

rm3l: After quick discussion, the current version of the code with the `select ... case ... default...` looks ok for a best-effort version check.
This might be a separate issue, but currently `odo dev` never checks for a newer version when it exits (probably because of the error returned when hitting `Ctrl+C`).

anandrkskd:
As per discussion with @rm3l, I will close this PR, and for BUILD/version update, we will perform it during the release Process.
/close

PR #6176, opened 2022-09-28
Additional Languages Tests support with ODO

HarshithaMS005:
**What type of PR is this:**
/kind feature

Test odo init and odo dev commands for different languages like Python, Python-Django, Java-Maven, dotnet60, GO, 
php-laravel, springboot.

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


kadel:
I'm afraid that those test doesn't make sense for odo repository. Odo tests should test odo features. Test added in this PR are testing the same odo features that are already tested elsewhere in odo test suite.


If the goal is to validate that various devfiles are correctly working with odo than this should happen on devfile/registry not here.

valaparthvi:
/close
As discussed in the cabal call, this is not the correct place to test support for all the Devfiles, https://github.com/devfile/registry probably is. I'm sorry that we cannot move forward with this.

Issue #6150, opened 2022-09-21
Experimental mode


feloy:
/kind user-story

## User Story
As a user I want to define ODO_EXPERIMENTAL So that I can use experimental commands and flags

## Acceptance Criteria 

When ODO_EXPERIMENTAL=true
- [ ] It should display in the header the command is running in Experimental mode
- [ ] It should enable experimental commands, which are not visible otherwise
- [ ] It should enable experimental flags, which are not visible otherwise
- [ ] If odo is executed without `ODO_EXPERIMENTAL=true`, experimental commands should appear non-existing (not just hidden) In other words, if `--run-on` is experimental flag `odo --run-on podman` should error out complaining about invalid flag. 

## Links 
- Related Epic (mandatory): #5865

/kind user-story

dharmit:
I'm speaking from an unpleasant experience in the past.

Having gone through one instance of moving things from experimental mode to normal mode, I humbly request not to go down this route again and find a different way of pushing new features. The myriad of `if..else` conditions it leads to and then removing those conditions when we want to take the feature(s) out of the experimental mode is not worth it, IMHO.

A few ideas that come to my mind:
1. Don't cut a release till we are sure that something is baked well.
2. Have a separate v3 branch, like we have the v2 branch. But v2 was slated for deprecation, v3 isn't/won't.
3. Keep doing releases every sprint like we do now, but have a warning on the website and repo that informs the users about the WIP nature of things and the unreliability that it entails.

feloy:
> I'm speaking from an unpleasant experience in the past.
> 
> Having gone through one instance of moving things from experimental mode to normal mode, I humbly request not to go down this route again and find a different way of pushing new features. The myriad of `if..else` conditions it leads to and then removing those conditions when we want to take the feature(s) out of the experimental mode is not worth it, IMHO.
> 

I can imagine some solutions to limit the problems for code maintenance.

For example, by defining Features, which are themselves defined as Experimental or not, and use conditions on these Features

dharmit:
> For example, by defining Features, which are themselves defined as Experimental or not, and use conditions on these Features

I don't understand. :(

kadel:
> I'm speaking from an unpleasant experience in the past.
> 
[...]

PR #6187, opened 2022-10-03
Support a generic --run-on flag

### LINKED ISSUE #6185: Accept a `--run-on` flag
/kind feature

`odo` should accept a `run-on` flag generically, in the same way it accepts a `-o` flag.

The value of the flag can be either `cluster` or `podman`

The value of the flag should be passed in the `ctx` received by Run (and make Complete and Validate receive this `ctx` too).

## Why is this needed?

The generic code (`genericlioptions.GenericRun`) needs to know the targeted platform to inject the correct package (podman of kubernetes) during dependency injection.

Part of #6151

feloy:
**What type of PR is this:**

/kind feature

- adds a new generic option `--run-on`.
- ctx is passed to Complete and Validate methods
- the presence of the -o json flag is saved in the ctx passed to Complete, Validate and Run
- the value of the --run-on flag is saved in the ctx passed to Complete, Validate and Run

Fixes #6185 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/odo/commonflags/run_on.go
```diff
@@ -0,0 +1,66 @@
...
+func CheckRunOnCommand(cmd *cobra.Command) error {
+
+	// Get the needed values
+	runOnFlag := pflag.Lookup(RunOnFlagName)
+	hasFlagChanged := runOnFlag != nil && runOnFlag.Changed
+	runOn := cmd.Annotations["runOn"]
+
+	// Check the valid output
+	if hasFlagChanged && runOnFlag.Value.String() != RunOnPodman && runOnFlag.Value.String() != RunOnCluster {
+		return fmt.Errorf("%s is not a valid target platform for --run-on, please select either cluster (default) or podman", runOnFlag.Value.String())
```
rm3l, line 48: Just to make it clear what the expected values are.
Also, even though #6150 will be adding an experimental mode, meanwhile, I think we could make it clear that `podman` support is experimental:

```suggestion
		return fmt.Errorf("%s is not a valid target platform for --run-on, please select either 'cluster' (default) or 'podman' (experimental)", runOnFlag.Value.String())
```

pkg/odo/commonflags/run_on.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	cmd.Annotations["runOn"] = "true"
+}
+
+// AddRunOnFlag adds the --run-on output flag to all commands
+// We use "flag" in order to make this accessible throughtout ALL of odo, rather than the
+// above traditional "persistentflags" usage that does not make it a pointer within the 'pflag'
+// package
+func AddRunOnFlag() {
+	flag.CommandLine.String(RunOnFlagName, "", "Specify target platform, supported platforms: cluster(default), podman")
```
rm3l, line 34: Same suggestion as above:

```suggestion
	flag.CommandLine.String(RunOnFlagName, "", "Specify target platform, supported platforms: 'cluster' (default), 'podman' (experimental)")
```

pkg/odo/cli/init/init.go
```diff
@@ -113,7 +114,7 @@ func (o *InitOptions) Validate() error {
 		return err
 	}
 
-	if len(o.flags) == 0 && log.IsJSON() {
+	if len(o.flags) == 0 && fcontext.GetOutput(ctx) {
```
rm3l, line 117: Would it make sense to have this function named `fcontext.IsJsonOutput`? Or make `GetOutput` return the actual output value as a constant?
(I am thinking of cases where we might need to support other output types in the future.)

pkg/odo/commonflags/context/context.go
```diff
@@ -0,0 +1,49 @@
...
+	runOnKey  runOnKeyType
+)
+
+// WithOutput sets the value for the output flag (-o) in ctx
+func WithOutput(ctx context.Context, val bool) context.Context {
+	return context.WithValue(ctx, outputKey, val)
+}
+
+// GetOutput gets value of output flag (-o) in ctx
+func GetOutput(ctx context.Context) bool {
```
rm3l, line 29: Same comment as above:

Would it make sense to have this function named `IsJsonOutput`? Or make it return the actual output value as a constant?
(I am thinking of cases where we might need to support other output types in the future.)

pkg/odo/commonflags/context/context.go
```diff
@@ -0,0 +1,49 @@
...
+	runOnKeyType  struct{}
+)
+
+var (
+	outputKey outputKeyType
+	runOnKey  runOnKeyType
+)
+
+// WithOutput sets the value for the output flag (-o) in ctx
+func WithOutput(ctx context.Context, val bool) context.Context {
```
rm3l, line 24: Same question: to make the intent clearer, would it make sense to rename this into something like `WithJsonOutput`?

pkg/odo/commonflags/output.go
```diff
@@ -0,0 +1,84 @@
...
+		v := flag.CommandLine.Lookup("v").Value.String()
+		if level, ok := os.LookupEnv("ODO_LOG_LEVEL"); ok && v == "0" {
+			_ = flag.CommandLine.Set("v", level)
+		}
+	}
+	return nil
+}
+
+// GetOutputValue returns true if -o flag is used and value is "json"
+func GetOutputValue(cmd cmdline.Cmdline) bool {
```
rm3l, line 82: Same question: to make the intent clearer, would it make sense to rename this into something like `GetJsonOutputValue`?

rm3l (changes requested):

pkg/odo/commonflags/output.go
feloy: I was not really happy with these function names, thanks for the suggestion

rm3l (approved):

PR #6189, opened 2022-10-04
`odo dev` deletes remote resources not present in the Devfile

### LINKED ISSUE #5943: odo dev cannot delete from cluster servicebindings from group servicebinding.io
/kind bug

 When a Devfile contains a Kubernetes resource declaring a servicebinding from group the servicebinding.io

and when removing the Kubernetes resource from the Devfile

`odo  dev` is unable to delete the servicebinding from the cluster, as it tries to remove a servicebinding from the group binding.operators.coreos.com and not servicebinding.io.

### LINKED ISSUE #6101: `odo dev` not handling changes to the Devfile after removing a Binding with `odo remove binding`
/kind bug
/area binding

**Operating System:**
Fedora 36

**Output of `odo version`:**
odo v3.0.0-rc1 (897f5f3e0)

Reproduction steps:
1. New project, but you can use any other project: `odo init --name my-sample-go --devfile go --starter go-starter`
2. Run `odo dev`, and wait until the Dev session is up and running
3. Add a new binding, e.g.: `odo add binding --name my-sample-go-postgresql-binding --service postgresql` => Changes to the Devfile are automatically handled, and the pod is recreated and binding information injected, which is fine.
4. Wait until the Dev session is ok again. Now, if I remove the previous binding from the Devfile with `odo remove binding --name my-sample-go-postgresql-binding`, the running `odo dev` session will just display "Updating component...", but it looks like nothing is actually happening in the cluster.

## Actual behavior
The pod is not recreated and the binding information is still there in the pod.

## Expected behavior
Similar to how `odo add binding` affects the currently running Dev pod, I would expect `odo remove binding` to cause the Dev pod to be recreated with no binding data injected in it.

valaparthvi:
**What type of PR is this:**
/kind bug
/area dev

With this PR, `odo dev` now handles K8s component related changes to the Devfile. It will remove resources from the cluster that are not present in the local Devfile.

Fixes #5943
Fixes #6101

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

1.
```sh
odo init --devfile nodejs --name my-node-app --starter nodejs-starter
```
2. Replace the devfile.yaml with content below.
<details>

```yaml
commands:
- exec:
    commandLine: npm install
    component: runtime
    group:
      isDefault: true
      kind: build
    hotReloadCapable: false
    workingDir: ${PROJECT_SOURCE}
  id: install
- exec:
    commandLine: npm start
    component: runtime
    group:
      isDefault: true
      kind: run
    hotReloadCapable: false
    workingDir: ${PROJECT_SOURCE}
  id: run
- exec:
    commandLine: npm run debug
    component: runtime
    group:
      isDefault: true
      kind: debug
    hotReloadCapable: false
    workingDir: ${PROJECT_SOURCE}
  id: debug
- exec:
    commandLine: npm test
    component: runtime
    group:
      isDefault: true
      kind: test
    hotReloadCapable: false
    workingDir: ${PROJECT_SOURCE}
  id: test
components:
- container:
    args:
    - tail
    - -f
    - /dev/null
    dedicatedPod: false
    endpoints:
    - name: http-node
      secure: false
      targetPort: 3000
    image: registry.access.redhat.com/ubi8/nodejs-16:latest
    memoryLimit: 1024Mi
    mountSources: true
  name: runtime
- name: my-node-app-app-cluster-sample
  kubernetes:
    inlined: |
      apiVersion: servicebinding.io/v1beta1
      kind: ServiceBinding
      metadata:
        name: my-node-app-app-cluster-sample
      spec:
        workload:
          apiVersion: apps/v1
          kind: Deployment
          name: my-node-app-app
        service:
          apiVersion: postgresql.k8s.enterprisedb.io/v1
          kind: Cluster
          name: cluster-sample
- name: deploy-k8s-resource
  kubernetes:
    inlined: |
        kind: Deployment
        apiVersion: apps/v1
        metadata:
          name: my-component
        spec:
          replicas: 1
          selector:
            matchLabels:
              app: node-app
          template:
            metadata:
              labels:
                app: node-app
            spec:
              containers:
                - name: main
                  image: registry.access.redhat.com/ubi8/nodejs-12:1-36
                  resources:
                    limits:
                      memory: "128Mi"
                      cpu: "500m"
metadata:
  description: Stack with Node.js 16
  displayName: Node.js Runtime
  icon: https://nodejs.org/static/images/logos/nodejs-new-pantone-black.svg
  language: javascript
  name: my-node-app
  projectType: nodejs
  tags:
  - NodeJS
  - Express
  - ubi8
  version: 2.0.0
schemaVersion: 2.1.0
starterProjects:
- git:
    remotes:
  name: nodejs-starter
```
</details>

3. After you run the command below, ensure that `odo dev` detects the change, and removes ServiceBinding.v1beta1.servicebinding.io from the cluster.
```shell
odo remove binding --name my-node-app-cluster-sample
```
4. Remove the deployment K8s component from devfile.yaml, and ensure that `odo dev` detects it and removes it from the cluster.

### DIFF
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -132,10 +132,22 @@ func (a Adapter) Push(ctx context.Context, parameters adapters.PushParameters, c
 	}
 	ownerReference := generator.GetOwnerReference(deployment)
 
+	// Delete remote resources that are not present in the Devfile
 	selector := odolabels.GetSelector(a.ComponentName, a.AppName, odolabels.ComponentDevMode, false)
-	_, _, err = a.deleteRemoteResourcesNotPresentInDevfile(selector, deployment)
+
+	objectsToRemove, serviceBindingSecretsToRemove, err := a.getRemoteResourcesNotPresentInDevfile(selector)
+	if err != nil {
+		return fmt.Errorf("unable to determine resources to delete: %w", err)
+	}
+
+	err = a.deleteRemoteResources(objectsToRemove)
 	if err != nil {
 		return fmt.Errorf("unable to delete remote resources: %w", err)
+	}
+
+	err = a.deleteServiceBindingSecretWhenSBOIsNotInstalled(serviceBindingSecretsToRemove, deployment)
+	if err != nil {
+		return fmt.Errorf("unable to delete service binding secrets: %w", err)
 	}
 
 	// Create all the K8s components defined in the devfile
@@ -582,10 +594,10 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
 	}
 }
 
-// deleteRemoteResourcesNotPresentInDevfile compares the list of Devfile K8s component and remote K8s resources
-// and removes the remote resources not present in the Devfile;
+// getRemoteResourcesNotPresentInDevfile compares the list of Devfile K8s component and remote K8s resources
+// and returns a list of the remote resources not present in the Devfile;
 // it ignores the core components (such as deployments, svc, pods; all resources with `component:<something>` label)
-func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(selector string, deployment *appsv1.Deployment) (objectsToRemove []unstructured.Unstructured, serviceBindingSecretsToRemove []unstructured.Unstructured, _ error) {
+func (a Adapter) getRemoteResourcesNotPresentInDevfile(selector string) (objectsToRemove, serviceBindingSecretsToRemove []unstructured.Unstructured, err error) {
 	currentNamespace := a.kubeClient.GetCurrentNamespace()
 	allRemoteK8sResources, err := a.kubeClient.GetAllResourcesFromSelector(selector, currentNamespace)
 	if err != nil {
@@ -621,9 +633,7 @@ func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(selector string, deplo
 	if err != nil {
 		return nil, nil, fmt.Errorf("error in determining if the Service Binding Opertor is supported, cause: %w", err)
 	}
-	isServiceBindingSecretFound := func(u unstructured.Unstructured, k8sName string) bool {
-		return u.GroupVersionKind() == kclient.SecretGVK && service.IsLinkSecret(u.GetLabels()) && u.GetLabels()[service.LinkLabel] == k8sName && u.GetNamespace() == currentNamespace
-	}
+
 	for _, remoteK := range remoteK8sResources {
 		matchFound := false
 		for _, devfileK := range devfileK8sResourcesUnstructured {
@@ -634,11 +644,16 @@ func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(selector string, deplo
 				break
 			}
 
-			// if SBO is supported, check if it's owner reference is a servicebinding present in the devfile;
-			// if the owner matches, consider it a match found
-			if isSBOSupported {
-				if remoteK.GroupVersionKind() == kclient.SecretGVK {
-					secretHasServiceBindingOwner := func() bool {
+			// check for secrets related to ServiceBinding
+			// if the SBO is installed, we can mark the secret as match if its ServiceBinding resource is found locally;
+			// we do this because such secrets will be owned by the ServiceBinding resource and will be handled by it
+			// if the SBO is not installed, we check if the secret's labels matches those created by odo for binding
+			if remoteK.GroupVersionKind() == kclient.SecretGVK {
+				var remoteSecretHasLocalServiceBindingOwner bool
+				if isSBOSupported {
+					// if SBO is installed, check if its owner reference is a servicebinding present in the devfile;
+					// if the owner matches, consider it a match found
+					remoteSecretHasLocalServiceBindingOwner = func() bool {
 						for _, owner := range remoteK.GetOwnerReferences() {
 							if owner.Kind == kclient.ServiceBindingKind && owner.Name == devfileK.GetName() {
 								return true
@@ -646,14 +661,14 @@ func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(selector string, deplo
 						}
 						return false
 					}()
-					if secretHasServiceBindingOwner {
-						matchFound = true
-						break
-					}
+
+				} else {
+					// if SBO is not installed, check if it is a servicebinding secret
+					remoteSecretHasLocalServiceBindingOwner = (service.IsLinkSecret(remoteK.GetLabels()) &&
+						remoteK.GetLabels()[service.LinkLabel] == devfileK.GetName() &&
+						remoteK.GetNamespace() == currentNamespace)
 				}
-			} else {
-				// if SBO is not supported, check if it is a servicebinding secret
-				if isServiceBindingSecretFound(remoteK, devfileK.GetName()) {
+				if remoteSecretHasLocalServiceBindingOwner {
 					matchFound = true
 					break
 				}
@@ -661,53 +676,58 @@ func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(selector string, deplo
 		}
 		if !matchFound {
 			// only remove servicebinding secrets if the SBO is not installed
-			if remoteK.GroupVersionKind() == kclient.SecretGVK && !isSBOSupported && service.IsLinkSecret(remoteK.GetLabels()) {
+			if !isSBOSupported && remoteK.GroupVersionKind() == kclient.SecretGVK && service.IsLinkSecret(remoteK.GetLabels()) {
 				serviceBindingSecretsToRemove = append(serviceBindingSecretsToRemove, remoteK)
 			} else {
 				objectsToRemove = append(objectsToRemove, remoteK)
 			}
 		}
 	}
+	return objectsToRemove, serviceBindingSecretsToRemove, nil
+}
 
+// deleteRemoteResources takes a list of remote resources to be deleted
+func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructured) error {
 	// Delete the resources present on the cluster but not in the Devfile
 	for _, objectToRemove := range objectsToRemove {
 		spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
 
-		gvr, err1 := a.kubeClient.GetGVRFromGVK(objectToRemove.GroupVersionKind())
-		if err1 != nil {
-			return objectsToRemove, serviceBindingSecretsToRemove, fmt.Errorf("unable to get information about Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err1.Error())
+		gvr, err := a.kubeClient.GetGVRFromGVK(objectToRemove.GroupVersionKind())
+		if err != nil {
+			return fmt.Errorf("unable to get information about Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
 		}
 
 		err = a.kubeClient.DeleteDynamicResource(objectToRemove.GetName(), gvr, true)
 		if err != nil {
 			if !kerrors.IsNotFound(err) {
 				spinner.End(false)
-				return objectsToRemove, serviceBindingSecretsToRemove, fmt.Errorf("unable to delete Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
+				return fmt.Errorf("unable to delete Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
 			}
 
 			klog.V(4).Infof("Failed to delete Kubernetes resource: %s/%s; resource not found", objectToRemove.GetKind(), objectToRemove.GetName())
 		}
 		spinner.End(true)
 	}
-	if isSBOSupported {
-		return objectsToRemove, serviceBindingSecretsToRemove, nil
-	}
-	// this part will only be executed with SBO is not installed
+	return nil
+}
+
+// deleteServiceBindingSecretWhenSBOIsNotInstalled takes a list of Service Binding secrets that should be deleted when the SBO is not installed.
+func (a Adapter) deleteServiceBindingSecretWhenSBOIsNotInstalled(serviceBindingSecretsToRemove []unstructured.Unstructured, deployment *appsv1.Deployment) error {
 	for _, secretToRemove := range serviceBindingSecretsToRemove {
 		spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", secretToRemove.GetKind(), secretToRemove.GetName())
 
-		err = service.DeleteLinkWithoutOperator(a.kubeClient, secretToRemove, deployment)
+		err := service.DeleteLinkWithoutOperator(a.kubeClient, secretToRemove, deployment)
 		if err != nil {
 			if !kerrors.IsNotFound(err) {
 				spinner.End(false)
-				return objectsToRemove, serviceBindingSecretsToRemove, fmt.Errorf("unable to delete Kubernetes resource: %s/%s: %s", secretToRemove.GetKind(), secretToRemove.GetName(), err.Error())
+				return fmt.Errorf("unable to delete Kubernetes resource: %s/%s: %s", secretToRemove.GetKind(), secretToRemove.GetName(), err.Error())
 			}
 			klog.V(4).Infof("Failed to delete Kubernetes resource: %s/%s; resource not found", secretToRemove.GetKind(), secretToRemove.GetName())
 		}
 		spinner.End(true)
 
 	}
-	return objectsToRemove, serviceBindingSecretsToRemove, nil
+	return nil
 }
 
 // getFirstContainerWithSourceVolume returns the first container that set mountSources: true as well
--- a/pkg/devfile/adapters/kubernetes/component/adapter.go
+++ b/pkg/devfile/adapters/kubernetes/component/adapter.go
@@ -701,17 +701,12 @@ func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructur
 		return nil
 	}
 
-	// stopCh := make(chan struct{})
-	// errCh := make(chan error)
-	// var goroutineCount int64
 	g := new(errgroup.Group)
 	// Delete the resources present on the cluster but not in the Devfile
 	for _, objectToRemove := range objectsToRemove {
-		spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
-		defer spinner.End(false)
-		// wg.Add(1)
-		// atomic.AddInt64(&goroutineCount, 1)
 		g.Go(func() error {
+			spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
+			defer spinner.End(false)
 			gvr, err := a.kubeClient.GetGVRFromGVK(objectToRemove.GroupVersionKind())
 			if err != nil {
 				err = fmt.Errorf("unable to get information about Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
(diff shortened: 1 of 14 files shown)


pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +579,65 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
 	}
 }
 
+// deleteRemoteResourcesNotPresentInDevfile compares the list of Devfile K8s component and remote K8s resources
+// and removes the remote resources not present in the Devfile;
+// it ignores the core components (such as deployments, svc, pods; all resources with `component:<something>` label)
+func (a Adapter) deleteRemoteResourcesNotPresentInDevfile(labels map[string]string) error {
+	allRemoteK8sResources, err := a.kubeClient.GetAllResourcesFromSelector(util.ConvertLabelsToSelector(labels), a.kubeClient.GetCurrentNamespace())
```
feloy, line 586: `labels` has been built with `odolabels.GetLabels`, which includes informative labels like `managed-by-version` one.

You will need to use `odolabels.GetSelector` instead to have a wider selection.

feloy: Also, you can use `selector` instead of `labels` as param name, to differentiate label / selector.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +579,65 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+			return fmt.Errorf("unable to obtain unstructured data: %w", err)
+		}
+		devfileK8sResourcesUnstructured = append(devfileK8sResourcesUnstructured, devfileKUnstructured)
+	}
+
+	var objectsToRemove []unstructured.Unstructured
+	for _, remoteK := range remoteK8sResources {
+		matchFound := false
+		for _, devfileK := range devfileK8sResourcesUnstructured {
+			if devfileK.GetKind() == remoteK.GetKind() && devfileK.GetName() == remoteK.GetName() && remoteK.GetNamespace() == a.kubeClient.GetCurrentNamespace() {
```
feloy, line 616: You will also need to test the group, or you could mix resources of same kind in different groups.

feloy (changes requested):

feloy (changes requested):

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +579,65 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+	// Delete the resources present on the cluster but not in the Devfile
+	for _, objectToRemove := range objectsToRemove {
+		log.Infof("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
+		gvr, err := a.kubeClient.GetGVRFromGVK(objectToRemove.GroupVersionKind())
+		if err != nil {
+			log.Warningf("Unable to get information about Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
+		}
+		err = a.kubeClient.DeleteDynamicResource(objectToRemove.GetName(), gvr, true)
+		if err != nil {
+			log.Warningf("Unable to delete Kubernetes resource: %s/%s: %s", objectToRemove.GetKind(), objectToRemove.GetName(), err.Error())
```
valaparthvi, line 635: @feloy I am confused if we should just warn the error here or return the error? Wdyt?

feloy: You could check for the response: if the error is NotFound, you could ignore it, elsewhere you could return the error

pkg/service/link.go
```diff
@@ -124,6 +124,49 @@ func pushLinksWithoutOperator(client kclient.ClientInterface, u unstructured.Uns
 	return nil
 }
 
+// DeleteLinkWithoutOperator deletes the ServiceBinding secret when Service Binding Operator is not installed
+func DeleteLinkWithoutOperator(kubeClient kclient.ClientInterface, secretToRemove unstructured.Unstructured, deployment *appsv1.Deployment) error {
+	currentNamespace := kubeClient.GetCurrentNamespace()
```
valaparthvi, line 129: The logic of this function has been ported from https://github.com/redhat-developer/odo/blob/224069cd68f4a03f4b5b17009fb4ca42ef8c5d89/pkg/service/link.go.

valaparthvi:
<details>
<summary>OCP failure seems flaky:</summary>

```sh
Summarizing 1 Failure:
  [FAIL] odo dev command tests port-forwarding for the component when devfile has single endpoint when running odo dev when modifying memoryLimit for container in Devfile [BeforeEach] should expose the endpoint on localhost
  Timed out after 180.000s.
  Expected
      <string>:  âš   Pod is Terminating
       âœ—  Finished executing the application (command: devrun) [38s]
       âš   No pod exists
       âš   Pod is Pending
       âœ“  Pod is Running
       â€¢  Syncing files into the container  ...
      
 âœ“  Syncing files into the container [586ms]
      cat: /opt/odo/.odo_cmd_devrun.pid: No such file or directory
       â€¢  Building your application in container on cluster (command: devbuild)  ...
      
 âœ“  Building your application in container on cluster (command: devbuild) [4s]
       â€¢  Executing the application (command: devrun)  ...
      44
      0
      44
      0
       -  Forwarding from 127.0.0.1:40247 -> 3000
      
      
      â†ª Dev mode
       Keyboard Commands:
      
  To satisfy at least one of these matchers: [%!s(*matchers.ContainSubstringMatcher=&{Pushing files... []}) %!s(*matchers.ContainSubstringMatcher=&{Updating Component... []})]
  In [BeforeEach] at: /go/odo_1/tests/helper/helper_run.go:53
```
</details>

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +594,142 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+					// if SBO is installed, check if its owner reference is a servicebinding present in the devfile;
+					// if the owner matches, consider it a match found
+					remoteSecretHasLocalServiceBindingOwner = func() bool {
+						for _, owner := range remoteK.GetOwnerReferences() {
+							if owner.Kind == kclient.ServiceBindingKind && owner.Name == devfileK.GetName() {
+								return true
+							}
+						}
+						return false
+					}()
```
feloy, line 663: 💯

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +594,142 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+
+				} else {
+					// if SBO is not installed, check if it is a servicebinding secret
+					remoteSecretHasLocalServiceBindingOwner = (service.IsLinkSecret(remoteK.GetLabels()) &&
+						remoteK.GetLabels()[service.LinkLabel] == devfileK.GetName() &&
+						remoteK.GetNamespace() == currentNamespace)
```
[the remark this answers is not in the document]
feloy: I don't think these brackets are necessary

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +594,142 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+			}
+
+			klog.V(4).Infof("Failed to delete Kubernetes resource: %s/%s; resource not found", objectToRemove.GetKind(), objectToRemove.GetName())
+		}
+		spinner.End(true)
+	}
+	return nil
+}
+
+// deleteServiceBindingSecretWhenSBOIsNotInstalled takes a list of Service Binding secrets that should be deleted when the SBO is not installed.
```
feloy, line 714: I cannot see the test if SBO is installed in this function

pkg/devfile/components.go
```diff
@@ -31,7 +32,7 @@ func GetKubernetesComponentsToPush(devfileObj parser.DevfileObj) ([]devfilev1.Co
 		componentName := ""
 		if command.Exec != nil {
 			componentName = command.Exec.Component
-		} else if command.Apply != nil {
+		} else if !allowApply && command.Apply != nil {
```
feloy, line 35: I think it should be the opposite, or I don't understand the meaning of this option:

```suggestion
		} else if allowApply && command.Apply != nil {
```

feloy (changes requested):

pkg/devfile/components.go
valaparthvi: By default, this function ignores the K8s components referenced by any command.
1. It collects all the K8s component and puts it in a map of name:component.
2. Filter out the components referenced by any command(apply or exec) and delete them from the map. Here, if allowApply is true, `!allowApply & command.Apply!=nil` becomes false and the components referenced by Apply command will not be deleted from the map.
3. Return the list of remaining components, essentially the ones not referenced by any command.

valaparthvi:
/retest
Flaky test.

tests/examples/source/devfiles/nodejs/devfile-with-k8s-resource.yaml
```diff
@@ -0,0 +1,75 @@
...
+        kind: build
+  - id: devrun
+    exec:
+      component: runtime
+      commandLine: npm start
+      workingDir: ${PROJECTS_ROOT}
+      group:
+        kind: run
+        isDefault: true
+  - id: run
```
rm3l, line 69: To simplify, this non-default `run` command does not seem necessary..

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +596,145 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+		}
+	}
+	return objectsToRemove, serviceBindingSecretsToRemove, nil
+}
+
+// deleteRemoteResources takes a list of remote resources to be deleted
+func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructured) error {
+	// Delete the resources present on the cluster but not in the Devfile
+	for _, objectToRemove := range objectsToRemove {
+		spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
```
rm3l, line 697: Safer to call `defer spinner.End(false)` right after this. Otherwise, spinner will never end if `gvr, err := a.kubeClient.GetGVRFromGVK(objectToRemove.GroupVersionKind())` line below returns a non-nil error.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +596,145 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+		spinner.End(true)
+	}
+	return nil
+}
+
+// deleteServiceBindingSecret takes a list of Service Binding secrets that should be deleted;
+// this is mainly helpful when Service Binding Operator is not installed on the cluster
+func (a Adapter) deleteServiceBindingSecret(serviceBindingSecretsToRemove []unstructured.Unstructured, deployment *appsv1.Deployment) error {
+	for _, secretToRemove := range serviceBindingSecretsToRemove {
+		spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", secretToRemove.GetKind(), secretToRemove.GetName())
```
rm3l, line 722: Same point about calling `defer spinner.End(false)` right after creating it.

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +596,145 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+			klog.V(4).Infof("Failed to delete Kubernetes resource: %s/%s; resource not found", objectToRemove.GetKind(), objectToRemove.GetName())
+		}
+		spinner.End(true)
+	}
+	return nil
+}
+
+// deleteServiceBindingSecret takes a list of Service Binding secrets that should be deleted;
+// this is mainly helpful when Service Binding Operator is not installed on the cluster
+func (a Adapter) deleteServiceBindingSecret(serviceBindingSecretsToRemove []unstructured.Unstructured, deployment *appsv1.Deployment) error {
```
rm3l, line 720: I would rename this method into `deleteServiceBindingSecrets` since it is handling multiple items ^^

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +596,145 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+			}
+		}
+	}
+	return objectsToRemove, serviceBindingSecretsToRemove, nil
+}
+
+// deleteRemoteResources takes a list of remote resources to be deleted
+func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructured) error {
+	// Delete the resources present on the cluster but not in the Devfile
+	for _, objectToRemove := range objectsToRemove {
```
rm3l, line 696: Open for discussion: we are iterating and sending out requests sequentially to delete possibly independent resources. We are also waiting for each `DeleteDynamicResource` to finish, which might take some time..
I think we should improve this by performing this in parallel with goroutines, waiting for them to return and returning any error that might occur.
I am thinking of `odo logs` which had similar issues with requests in a loop..

What do you think?

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +596,145 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+		}
+		spinner.End(true)
+	}
+	return nil
+}
+
+// deleteServiceBindingSecret takes a list of Service Binding secrets that should be deleted;
+// this is mainly helpful when Service Binding Operator is not installed on the cluster
+func (a Adapter) deleteServiceBindingSecret(serviceBindingSecretsToRemove []unstructured.Unstructured, deployment *appsv1.Deployment) error {
+	for _, secretToRemove := range serviceBindingSecretsToRemove {
```
rm3l, line 721: Same point about sending out requests in parallel.

rm3l (changes requested):

pkg/devfile/adapters/kubernetes/component/adapter.go
valaparthvi: You're right. I missed that.

pkg/devfile/adapters/kubernetes/component/adapter.go
valaparthvi: Hmm, that's interesting, and you're possibly right. Although, currently, we wait for the resources to be deleted, I could simply not wait for it to be deleted, that would also make things faster, don't you think?

pkg/devfile/adapters/kubernetes/component/adapter.go
valaparthvi: I am not sure doing that here is a good idea. There are possible independent secrets(only one secret most of the time), but only one deployment that is linked to all the secrets. When the unbinding happens, it modifies the deployment by removing any reference to the secret, and then proceeds to delete the secret.

rm3l: Alright, fine then to keep it as is here.

valaparthvi:
@rm3l I've attempted to implement parallel resource deletion, but I cannot say for sure if it is working as expected. I'll leave it to you to test the functionality.

feloy:
For parallel deletion, you can use WaitGroups: https://gobyexample.com/waitgroups

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +598,167 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructured) error {
+	if len(objectsToRemove) == 0 {
+		return nil
+	}
+
+	g := new(errgroup.Group)
+	// Delete the resources present on the cluster but not in the Devfile
+	for _, objectToRemove := range objectsToRemove {
+		g.Go(func() error {
+			spinner := log.Spinnerf("Deleting Kubernetes resource: %s/%s", objectToRemove.GetKind(), objectToRemove.GetName())
```
feloy, line 708: I didn't test, but I think the display of the spinners will be mixed up between go-routines, as they will execute in parallel.

We will probably need to have only one spinner to indicate the deletion of all the resources (to be started before the for loop, and terminated after Wait())

pkg/devfile/adapters/kubernetes/component/adapter.go
```diff
@@ -571,6 +598,174 @@ func (a Adapter) generateDeploymentObjectMeta(deployment *appsv1.Deployment, lab
...
+
+// deleteRemoteResources takes a list of remote resources to be deleted
+func (a Adapter) deleteRemoteResources(objectsToRemove []unstructured.Unstructured) error {
+	if len(objectsToRemove) == 0 {
+		return nil
+	}
+
+	var resources string
+	for _, u := range objectsToRemove {
+		resources += fmt.Sprintf("%s/%s, ", u.GetKind(), u.GetName())
```
rm3l, line 706: This will leave a trailing comma in the spinner message, e.g:

```
 ✓  Deleting Kubernetes resources not present in the Devfile: Deployment/my-component, Deployment/my-component-2, Deployment/my-component-3, Deployment/my-component-4, Deployment/my-component-5, Deployment/my-component10, Deployment/my-component6, Deployment/my-component7, Deployment/my-component8, Deployment/my-component9,  
```

Maybe using `strings.Join` against a slice of strings would be better.

rm3l (commented):
Nice - thanks for the changes.

I just tested it, and it works better for me with parallel deletion.
On a remote ClusterBot cluster, I went from 21 seconds waiting for deletion, to "just" 9 seconds (because some of the parallel requests got throttled by a few seconds, but I think it should still be ok).

```
File /tmp/go/devfile.yaml changed                                                                                                                                                                                                              
 ◐  Deleting Kubernetes resources not present in the Devfile: Deployment/my-component, Deployment/my-component-2, Deployment/my-component-3, Deployment/my-component-4, Deployment/my-component-5, Deployment/my-component10, Deploy...I1021 13:21:27.226557 3338741 request.go:601] Waited for 1.000571871s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-ln-s5tfy4k-76ef8.aws-2.ci.openshift.org:6443/apis/apps/v1/namespaces/rrr/deployments?limit=250
 ✓  Deleting Kubernetes resources not present in the Devfile: Deployment/my-component, Deployment/my-component-2, Deployment/my-component-3, Deployment/my-component-4, Deployment/my-component-5, Deployment/my-component10, Deployment/my-component6, Deployment/my-component7, Deployment/my-component8, Deployment/my-component9,  [9s]
 ✓  Syncing files into the container [15ms]
```

rm3l:
> @rm3l I've attempted to implement parallel resource deletion, but I cannot say for sure if it is working as expected. I'll leave it to you to test the functionality.
> 
> Also, locally I am seeing the following error in my IDE
> 
> ```go
> Cannot use '(*Adapter)(nil)' (type *Adapter) as the type ComponentAdapter Type does not implement 'ComponentAdapter' as some methods are missing: Push(parameters adapters.PushParameters, componentStatus *watch.ComponentStatus) error
> ```
> 
> Can you confirm if you too are seeing it ? I am using Go 1.18 locally.

I didn't see this error. Are you still seeing it?

valaparthvi:
> > @rm3l I've attempted to implement parallel resource deletion, but I cannot say for sure if it is working as expected. I'll leave it to you to test the functionality.
> > Also, locally I am seeing the following error in my IDE
> > ```go
> > Cannot use '(*Adapter)(nil)' (type *Adapter) as the type ComponentAdapter Type does not implement 'ComponentAdapter' as some methods are missing: Push(parameters adapters.PushParameters, componentStatus *watch.ComponentStatus) error
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can you confirm if you too are seeing it ? I am using Go 1.18 locally.
> 
> I didn't see this error. Are you still seeing it?

Not on Go1.17

pkg/devfile/adapters/kubernetes/component/adapter.go
valaparthvi: It's definitely better!

PR #6201, opened 2022-10-06
Move --var-file and --var as generic flags

### LINKED ISSUE #6200: Move --var-file and --var as generic flags
/kind feature

Move `--var-file` and `--var` as generic flags, so it is possible to read the Devfile in a generic way from the `genericclioptions.GenericRun` function, and pass the Devfile in the global context

feloy:
**What type of PR is this:**

/kind feature

This PR moves the --var-files and --var flags to generic flags, accessible before any command is considered. This will help reading the Devfile in a generic way. 

For the moment, the Devfile is still read from the commands' Complete methods, but this will be moved to `genericclioptions.GenericRun` in a future PR.

Fixes #6200

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/odo/commonflags/context/context.go
```diff
@@ -47,3 +49,17 @@ func GetRunOn(ctx context.Context) string {
 	}
 	return commonflags.RunOnDefault
 }
+
+// WithRunOn sets the value for the run-on flag in ctx
+func WithVariables(ctx context.Context, val map[string]string) context.Context {
```
dharmit, line 54: Comment and function name don't match.

pkg/odo/commonflags/context/context.go
```diff
@@ -47,3 +49,17 @@ func GetRunOn(ctx context.Context) string {
...
+	return context.WithValue(ctx, variablesKey, val)
+}
+
+// GetRunOn gets value of run-on flag in ctx
+func GetVariables(ctx context.Context) map[string]string {
```
dharmit, line 59: Same - comment and function name don't match.

pkg/odo/commonflags/variables.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd.Annotations["variables"] = "true"
+}
+
+// AddVariablesFlags adds the --var-file and --var flags to all commands
+// We use "flag" in order to make this accessible throughtout ALL of odo, rather than the
+// above traditional "persistentflags" usage that does not make it a pointer within the 'pflag'
+// package
+func AddVariablesFlags() {
+	pflag.CommandLine.StringArray(VarFlagName, []string{}, "Variable to override Devfile variable and variables in var-file")
+	pflag.CommandLine.String(VarFileFlagName, "", "File containing variables to override Devfile variables")
```
rm3l, line 33: Just a discussion (also applies to the `run-on` flag):

From the user perspective, now that these two flags are visible globally, can the `odo help` output provide more guidance by telling users which commands these flags can be used with? I understand this would mean maintaining these descriptions, but I think it would be helpful.
So I'd suggest something like:

```suggestion
	pflag.CommandLine.String(VarFileFlagName, "", "File containing variables to override Devfile variables. Usable with \"odo dev\" and \"odo deploy\" commands.")
```

What do you think?

Because currently, users have to explicitly run the command to see if the flag can be used or not, e.g.:

```
> odo init --var USER                                                             
 ✗  --var-file and --var flags are not supported for this command
```

pkg/odo/cli/dev/dev.go
```diff
@@ -126,10 +125,7 @@ func (o *DevOptions) Complete(ctx context.Context, cmdline cmdline.Cmdline, args
 		return err
 	}
 
-	o.variables, err = vars.GetVariables(o.clientset.FS, o.varFileFlag, o.varsFlag, os.LookupEnv)
-	if err != nil {
-		return err
-	}
+	o.variables = fcontext.GetVariables(ctx)
```
rm3l, line 128: Just curious as to why `pkg/odo/cli/deploy/deploy.go` was not refactored too.. Was it forgotten or is something preventing us from having the same logic in `odo deploy`? It also supports `--var` and `--var-file` flags.

rm3l (changes requested):

pkg/odo/cmdline/cobra.go
```diff
@@ -87,6 +87,12 @@ func (o *Cobra) FlagValueIfSet(flagName string) string {
 	return flag
 }
 
+// FlagValueIfSet retrieves the value of the specified flag if it is set for the given command
+func (o *Cobra) FlagValuesIfSet(flagName string) []string {
+	flag, _ := o.cmd.Flags().GetStringArray(flagName)
```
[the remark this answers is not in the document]
feloy: Right, I changed it to FlagValues instead, returnning the error

pkg/odo/cli/dev/dev.go
feloy: I forgot it. Adding it now

pkg/odo/commonflags/variables.go
feloy: I would like to try and do it in an automatic way, to save maintenance

rm3l: Yup, it would be even better if we could do it automatically 👍🏿 
If it is not too much work, we can do it in a separate issue/PR though.

feloy: ok, let's work on this in another PR

rm3l (commented):
Thanks for the changes.

/approve

rm3l:
LGTM feature-wise.

@dharmit Since you added the initial lgtm label, can you please take another look and re-add it if ok for you?
Thanks.

Issue #5517, opened 2022-03-03
remove `--context` flag from cli


kadel:
/kind task
/priority high

in v3 there will be no `--context` flag in any of the odo commands.
The flag should be removed from odo.

## Acceptance criteria
- [x] check all commands and remove `--context` flag

## links
parent task https://github.com/redhat-developer/odo/issues/5403

kadel:
`--context` flag is there for commands that will be removed in #5521

feloy:
The `build-image` command still accepts the `--context` flag

feloy:
Fixed #6202 
/close

PR #6202, opened 2022-10-07
Build context as part of GenericRun

### LINKED ISSUE #6186: Build the Context as part of the GenericRun
/kind feature

Get the context as part of the `genericclioptions.GenericRun` function, instead of calling `o.Context, err = genericclioptions.New(params)` in every command code, and pass the context as part of the ctx received by Run (and make Complete and Validate also receive this context).

Related to #6185 

Context:
- application: "app"
- component name
- directory
- devfile name (.../devfile.yaml or .../.devfile.yaml)
- devfile object
- output flag

feloy:
**What type of PR is this:**

/kind cleanup

Fixes #6186 , #5517

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/odo/cli/dev/dev.go
```diff
@@ -214,11 +168,11 @@ func (o *DevOptions) HandleSignal() error {
...
 
-func (o *DevOptions) Cleanup(commandError error) {
+func (o *DevOptions) Cleanup(ctx context.Context, commandError error) {
 	if commandError != nil {
-		devFileObj := o.Context.EnvSpecificInfo.GetDevfileObj()
-		componentName := o.GetComponentName()
-		_ = o.clientset.WatchClient.CleanupDevResources(devFileObj, componentName, log.GetStdout())
+		devFileObj := odocontext.GetDevfileObj(ctx)
+		componentName := odocontext.GetComponentName(ctx)
+		_ = o.clientset.WatchClient.CleanupDevResources(ctx, *devFileObj, componentName, log.GetStdout())
```
rm3l, line 175: Would it make sense to change the `watch` interface and implementation, so clients could just pass a Context to `CleanupDevResources`, e.g.:
```go
_ = o.clientset.WatchClient.CleanupDevResources(ctx, log.GetStdout())
```

rm3l (changes requested):
Added a few comments I had so far. I am yet to review the other commits ;-)

pkg/odo/cli/dev/dev.go
feloy: There are many places where this kind of refactoring needs to be done, I would prefer to do this in specific PRs, to not add more to this one.

We have to take care in which packages we want to propagate this context (for example, the odocontext.Get* functions should not be used in low-level packages like kclient, exec, sync, etc but only on odo-specific packages.

rm3l: Makes sense - thanks.

rm3l (approved):
LGTM - it was quite interesting to follow the narrative of this quite large PR by reviewing it commit-by-commit. It didn't require too much context-switch. Thanks.

dharmit:

because the tests are passing on IBM Cloud.

PR #6209, opened 2022-10-11
Add GitHub Workflow for issue management

rm3l:
**What type of PR is this:**
/kind feature

To ease project and issue management, this PR adds a GitHub Workflow that automates some manual tasks:
- automatically labels new or reopened issues with the "needs-triage" label
- automatically adds new or reopened issues to the odo Project.

&mdash;

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

Once merged, new or reopened issues should be added to the Project and labeled accordingly.

### DIFF
--- a/.github/workflows/issue-management.yaml
+++ b/.github/workflows/issue-management.yaml
@@ -1,0 +1,39 @@
+name: Issue Management
+
+on:
+  issues:
+    types:
+    - opened
+    - reopened
+
+env:
+  ORGANIZATION: redhat-developer
+  # See https://github.com/redhat-developer/odo/projects?query=is%3Aopen
+  PROJECT_NUMBER: 16
+
+jobs:
+  label_issue:
+    name: Label issue
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dharmit (commented):
I see two jobs with one step each. Since both have `runs-on` field, I assume they would each need/spin up a container. Is it possible to combine them into one job to reduce the resources it would need?

My reasoning is to only be resourceful.

rm3l:
> I see two jobs with one step each. Since both have `runs-on` field, I assume they would each need/spin up a container. Is it possible to combine them into one job to reduce the resources it would need?
> 
> My reasoning is to only be resourceful.

I understand your point.
For now, we have one step in each job, but I wanted to leave room for other steps we might think of later.
Also, the jobs being quite independent, it made sense to me to have different jobs that would run in parallel.

rm3l:

Not related to code, and tests are passing on IBM Cloud.

PR #6223, opened 2022-10-13
Allow passing `caller` to telemetry

### LINKED ISSUE #6147: allow passing `caller` to telemetry
/kind user-story

IDE plugins should be able to send some kind of id, that will get passed down to the 
telemetry so we know if the command was executed via plugin or directly by user from cli

IDE plugins will set env variable with every odo command they execute

## User Story
As an odo developer/PM I want to create metrics that will track odo usage by different plugin users

## Acceptance Criteria 
- [ ] In the event that is send to telemetry there should be a `caller` field
- [ ] if `caller` is empty it means that odo was invoked directly via command line
- [ ] each IDE plugin using odo will have its own caller id `vscode`, `intellij`, `jboss`
- [ ] we should enforce a list of allowed values (`vscode`, `intellij`, `jboss`). Do not display this list if value is invalid.
- [ ] there should be an environment variable `TELEMETRY_CALLER` that sets the caller field in telemetry

## Links 
- Related Epic (mandatory): 

/kind user-story

rm3l:
**What type of PR is this:**
/kind feature
/area telemetry

This sets a `caller` property in telemetry based upon the `TELEMETRY_CALLER` environment variable.
Per the acceptance criteria, this is enforced against a list of allowed values.

Fixes #6147

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

Set `ODO_DEBUG_TELEMETRY_FILE` and `TELEMETRY_CALLER` environment variables prior to calling `odo` and check the resulting telemetry data file. Or check telemetry servers.


pkg/odo/genericclioptions/runnable.go
```diff
@@ -93,6 +93,11 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		klog.V(4).Infof("WARNING: debug telemetry, if enabled, will be logged in %s", debugTelemetry)
 	}
 
+	err = scontext.SetCaller(cmd.Context(), scontext.CallerType(os.Getenv(segment.TelemetryCaller)))
+	if err != nil {
+		log.Warningf("error handling caller property for telemetry: %v", err)
```
rm3l, line 98: @kadel Not sure what the behavior should be for example if an invalid / unknown value is set as `caller`. I chose to just display a warning (and not block the `odo` command from running), but telemetry will not contain the `caller` property. Is that ok?

pkg/segment/context/context.go
```diff
@@ -130,6 +140,33 @@ func SetFlags(ctx context.Context, flags *pflag.FlagSet) {
 	setContextProperty(ctx, Flags, strings.Join(changedFlags, " "))
 }
 
+// SetCaller sets the caller property for telemetry to record the tool used to call odo.
+// Passing an empty caller is not considered invalid, but means that odo was invoked directly from the command line.
+// In all other cases, the value is verified against a set of allowed values.
+func SetCaller(ctx context.Context, caller CallerType) error {
+	isValid := func(caller CallerType) bool {
+		if caller == "" {
```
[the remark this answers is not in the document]
rm3l: No specific reason.
Initially the condition `if len(caller) == 0`, but I changed to `if caller == ""` (which reads better to me), but didn't think about including it in the switch.
Thanks - I'll change it.

pkg/odo/genericclioptions/runnable.go
kadel: My concern about this approach is that it will create bad data in our telemetry. 

In case of the wrong `TELEMETRY_CALLER` value, I would either error out (block odo from running) or show a warning, but in that case, I think we should set `caller` property.

The reason for this is that we can filter out wrong `caller` values in telemetry, but if `caller` is omited, then it will be automatically assigned as a regular cli call, it will affect statistics, and we won't be able to filter those out.

valaparthvi: I think it's right to not block odo from running and simply display warning, but I would prefer if we do not log it as a level 0, user-facing warning, given that we do not want users to know of telemetry. IMO, if the value is not acceptable, we can return "caller: unknown".

rm3l: I also agree with not blocking `odo`.
But maybe we want to know which value was set, instead of overriding it with `unknown`?
Given the explicit list of allowed values, we could still be able to filter.

rm3l: > but I would prefer if we do not log it as a level 0, user-facing warning, given that we do not want users to know of telemetry.

If `TELEMETRY_CALLER` is explicitly set, it may mean to me that the user is "telemetry-aware".. In this case, I think it might make sense to tell them that the value is not acceptable, no?

docs/website/docs/overview/configure.md
```diff
@@ -179,12 +179,13 @@ Options here are mostly used for debugging and testing `odo` behavior.
...
+| `DOCKER_CMD`                | The command executed to run the local docker binary. `docker` by default                                            |
+| `ODO_LOG_LEVEL`             | Useful for setting a log level to be used by odo commands.                                                          |
+| `ODO_DISABLE_TELEMETRY`     | Useful for disabling telemetry collection.                                                                          |
+| `GLOBALODOCONFIG`           | Useful for setting a different location of global preference file preference.yaml.                                  |
+| `ODO_DEBUG_TELEMETRY_FILE`  | Useful for debugging telemetry. When set it will save telemetry data to a file instead of sending it to the server. |
+| `DEVFILE_PROXY`             | Integration tests will use this address as Devfile registry instead of `https://registry.stage.devfile.io`          |
+| `TELEMETRY_CALLER`          | Caller identifier passed to telemetry.                                                                              |
```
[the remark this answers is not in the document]
rm3l: Okay, I'll add the acceptable values here.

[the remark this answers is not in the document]
rm3l: Updated.

pkg/segment/context/context.go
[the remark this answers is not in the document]
rm3l: Updated in [5d731b67810eea8b82baf518906768f013340f8c](https://github.com/redhat-developer/odo/pull/6223/commits/5d731b67810eea8b82baf518906768f013340f8c)

pkg/odo/genericclioptions/runnable.go
rm3l: Okay, so I've changed the behavior so that the `caller` property is set in telemetry with the value of `TELEMETRY_CALLER`, no matter what this value is. We are still displaying a user-facing warning if `TELEMETRY_CALLER` is invalid, but the `odo` command does not error out.

See [5d731b67810eea8b82baf518906768f013340f8c](https://github.com/redhat-developer/odo/pull/6223/commits/5d731b67810eea8b82baf518906768f013340f8c)

I've updated the integration tests accordingly in [515b8a3e4a01f9fe565cc5217d79e3c823c90890](https://github.com/redhat-developer/odo/pull/6223/commits/515b8a3e4a01f9fe565cc5217d79e3c823c90890)

pkg/odo/genericclioptions/runnable.go
```diff
@@ -93,6 +93,11 @@ func GenericRun(o Runnable, cmd *cobra.Command, args []string) {
 		klog.V(4).Infof("WARNING: debug telemetry, if enabled, will be logged in %s", debugTelemetry)
 	}
 
+	err = scontext.SetCaller(cmd.Context(), scontext.CallerType(os.Getenv(segment.TelemetryCaller)))
+	if err != nil {
+		log.Warningf("Error handling caller property for telemetry: %v", err)
```
valaparthvi, line 98: Are we sure we want to print this as a user facing warning? Shouldn't we hide it to log level 3 or 4?

rm3l: It made sense to me when invalid values were not inserted to telemetry properties. Now that even invalid values are being sent out, yes, I think we can remove this warning.
I'll change it.

rm3l: Changes pushed - please take another look.

PR #6230, opened 2022-10-18
Minor fixes to documentation

valaparthvi:
**What type of PR is this:**
/kind documentation

**What does this PR do / why we need it:**
This PR makes some things that we missed out on that I observed while reviewing downstream.
**Which issue(s) this PR fixes:**

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [x] Documentation 

**How to test changes / Special notes to the reviewer:**

PR #6238, opened 2022-10-20
Update message about v3 on documentation home page

rm3l:
**What type of PR is this:**
/kind documentation
/area documentation

Current message `Version 3 of odo is arriving`. This PR changes that message, now that v3 is available.

&mdash;

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [x] Documentation 


valaparthvi:
Perhaps we should also consider removing "Stable" from the dropdown for both the versions?
Screenshot from 2022-10-20 14-34-32

rm3l:
> Perhaps we should also consider removing "Stable" from the dropdown for both the versions? Screenshot from 2022-10-20 14-34-32

Yes, good point. I didn't think about this when I changed the version.
Can I push that change to #6230 instead ? This would make more sense to me :)

rm3l:
> > Perhaps we should also consider removing "Stable" from the dropdown for both the versions? Screenshot from 2022-10-20 14-34-32
> 
> Yes, good point. I didn't think about this when I changed the version. Can I push that change to #6230 instead ? This would make more sense to me :)

Whoops - too late - already merged. I'll update it here then..

valaparthvi (approved):

valaparthvi:

Issue #6237, opened 2022-10-20
Integration tests without real Kubernetes/OpenShift clusters


feloy:
/kind user-story

## User Story
As a developer I want to run Integration tests without real Kubernetes/openShift clusters So that tests are more reliable

## Acceptance Criteria 
- [ ] It should test commands which are not using cluster (odo preference, etc) without cluster connection
- [ ] It should test on an environment without cluster that commands needing a cluster are failing correctly with an appropriate message
- [ ] It should use a very simple cluster (https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) and test the resources created  by odo into the cluster, without expecting any action from controllers

## Links 
- Related Epic (mandatory): 

/kind user-story

Issue #5840, opened 2022-06-20
Fix integration and e2e test organization structure


kadel:
I've just noticed that we still have integration tests separated from devfile tests (#5839).
This organization doesn't make sense anymore.

## User Story
- As an odo developer
- I want to have a straightforward test structure, and when I'm executing integration tests I want to be sure that all tests are executed
- So that don't introduce new bugs into existing code that is already covered by tests

/kind user-story

## Acceptance criteria
- [ ] To make executing tests more straightforward forward there should be only 3 ways to start tests.
  ```
  # unit test
  make test
  
  # integration tests
  make test-integrationn
  
  # e2e tess
  make test-e2e
  ```
- [ ] no other `make test-..` target should exist
- [ ] There should be no `tests/integration/devile/ directory, all integration tests should be in `tests/integration`.  Or in another words `go run github.com/onsi/ginkgo/ginkgo tests/integration/` should execute all integration tests.

/area user-story
/priority hight

rm3l:
Also, there currently is a separate `test-interactive` target because interactive tests do not run on Windows (cf. https://github.com/redhat-developer/odo/issues/5573). But we could unify all those targets by skipping interactive tests in the test code (as we already do with an environment variable for Kubernetes vs Openshift tests).

anandrkskd:
Do we want to remove the tree structure of directories under `./test/integration/` and have only one directory `/test/integration` that contains all the go test files?
Or do we just want to have a simplified `make targets` for tests execution?

rm3l:
> Do we want to remove the tree structure of directories under `./test/integration/` and have only one directory `/test/integration` that contains all the go test files? Or do we just want to have a simplified `make targets` for tests execution?
> 
> WDYT? @rm3l @feloy @dharmit
>

I think @kadel's point - and I agree with him - was to have both :-)
- all integration test files should live in the same `tests/integration` folder, and all e2e test files stay (as they currently are) in the  `tests/e2escenarios` folder. If we need to skip some tests, we can do it in the test code, rather than creating different Make targets.
- there is way too many (32) `test*` targets in the current Makefile. I am not even sure all of these targets are used; we should simplify this by having just 3 targets for running the tests: `test`, `test-integration`, and `test-e2e`.
```shell
❯ grep "^test" Makefile | awk -F ':' '{print $1}' | sort
test
test-cmd-debug
test-cmd-delete
test-cmd-devfile-app
test-cmd-devfile-config
test-cmd-devfile-debug
[...]

PR #6242, opened 2022-10-20
Isolate no cluster tests

feloy:
**What type of PR is this:**

/kind tests

This PR moves integration tests for commands not needing a cluster to a specific directory
so they can be executed separately, in an environment without any cluster.

Fixes partly #6237 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,8 @@ export ARTIFACT_DIR ?= .
 
 GINKGO_FLAGS_ALL = $(GINKGO_TEST_ARGS) --randomize-all --slow-spec-threshold=$(SLOW_SPEC_THRESHOLD) -timeout $(TIMEOUT) --no-color
 
-# Flags for tests that must not be run in parallel.
-GINKGO_FLAGS_SERIAL = $(GINKGO_FLAGS_ALL) -nodes=1
+# Flags to run one test per core.
+GINKGO_FLAGS_AUTO = $(GINKGO_FLAGS_ALL) -p
 # Flags for tests that may be run in parallel
 GINKGO_FLAGS=$(GINKGO_FLAGS_ALL) -nodes=$(TEST_EXEC_NODES)
 # GolangCi version for unit-validate test
@@ -184,9 +184,16 @@ vendor-update: ## Update vendoring
 openshiftci-presubmit-unittests:
 	./scripts/openshiftci-presubmit-unittests.sh
 
+.PHONY: test-integration-cluster
+test-integration-cluster:
+	$(RUN_GINKGO) $(GINKGO_FLAGS) --label-filter="!nocluster" tests/integration
+
+.PHONY: test-integration-no-cluster
+test-integration-no-cluster:
+	$(RUN_GINKGO) $(GINKGO_FLAGS_AUTO) --label-filter=nocluster tests/integration
+
 .PHONY: test-integration
-test-integration:
-	$(RUN_GINKGO) $(GINKGO_FLAGS) tests/integration
+test-integration: test-integration-no-cluster test-integration-cluster
 
 .PHONY: test-e2e
 test-e2e:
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -188,6 +188,10 @@ openshiftci-presubmit-unittests:
 test-integration:
 	$(RUN_GINKGO) $(GINKGO_FLAGS) tests/integration
 
+.PHONY: test-integration-no-cluster
+test-integration-no-cluster:
+	$(RUN_GINKGO) $(GINKGO_FLAGS_AUTO) tests/integration/no-cluster
+
```
valaparthvi, line 194:
```sh
Ran 58 of 58 Specs in 21.788 seconds
SUCCESS! -- 58 Passed | 0 Failed | 0 Pending | 0 Skipped


Ginkgo ran 1 suite in 37.69106764s
Test Suite Passed
```
:100:

valaparthvi (commented):
/approve

Makefile
```diff
@@ -188,6 +188,10 @@ openshiftci-presubmit-unittests:
 test-integration:
 	$(RUN_GINKGO) $(GINKGO_FLAGS) tests/integration
```
valaparthvi, line 189: Wouldn't this run the no-cluster tests as well?

Makefile
```diff
@@ -188,6 +188,10 @@ openshiftci-presubmit-unittests:
...
 
+.PHONY: test-integration-no-cluster
+test-integration-no-cluster:
```
rm3l, line 192: Open question I was wondering: while I understand the value of this, in #5840, we decided to simplify the targets and structure, with a single `test-integration` target that would execute all tests.
So I was wondering if we could still keep a single `test-integration` target, but perhaps instruct Ginkgo to pick the `no-cluster` tests differently. I see something really interesting with [Spec Labels](https://onsi.github.io/ginkgo/#spec-labels) that could be used as a more structured mechanism for organizing and filtering specs.

What do you think?

feloy: I would prefer to start working on a cluster-less environment from the beginning.

If we have a single test-integration mixing the cluster-full and cluster-less environments, we still need to provide an environment with a cluster, and connect get experience from running a cluster-less env.

rm3l: Okay, that makes sense from an environment setup perspective.

But regarding the Makefile, I'm still wondering, since this means that we need to not forget about running both `test-integration-no-cluster` and `test-integration` targets if we want to run all tests.
So maybe we could keep the `test-integration` target as a single target that would run all tests, but introduce a `test-integration-cluster` that would run the ones requiring a cluster?

feloy: Yes, that makes sense

feloy:
> LGTM, but I was wondering if the `setupCluster` argument we have in `helper#CommonBeforeEach` now does not seem kind of redundant with the `nocluster` label.
> 
> Thinking about this because I noticed that tests that are labelled with `nocluster` actually need to also call `helper.CommonBeforeEach(false)`, but there is nothing guaranteeing that this is the case.
> 
> At least, I think we can maybe programmatically determine the presence (or absence) of the `nocluster` label and based on that information determine the value of `setupCluster`. I can see that Ginkgo has an interesting [`CurrentSpecReport().Labels()`](https://pkg.go.dev/github.com/onsi/ginkgo/v2#CurrentSpecReport) function that could return such information during the spec run phase.
> 
> What do you think?

I'm not very comfortable with this kind of hidden side effect.

I find meaningful to have the `nocluster` label confirmed by `helper.CommonBeforeEach(false)`: every time I added the label, I checked that this declaration was confirmed by something in the code.

rm3l (approved):
> > LGTM, but I was wondering if the `setupCluster` argument we have in `helper#CommonBeforeEach` now does not seem kind of redundant with the `nocluster` label.
> > Thinking about this because I noticed that tests that are labelled with `nocluster` actually need to also call `helper.CommonBeforeEach(false)`, but there is nothing guaranteeing that this is the case.
> > At least, I think we can maybe programmatically determine the presence (or absence) of the `nocluster` label and based on that information determine the value of `setupCluster`. I can see that Ginkgo has an interesting [`CurrentSpecReport().Labels()`](https://pkg.go.dev/github.com/onsi/ginkgo/v2#CurrentSpecReport) function that could return such information during the spec run phase.
> > What do you think?
> 
> I'm not very comfortable with this kind of hidden side effect.
> 
> I find meaningful to have the `nocluster` label confirmed by `helper.CommonBeforeEach(false)`: every time I added the label, I checked that this declaration was confirmed by something in the code.

Okay - makes sense. I was mostly thinking of a case where someone has the `nocluster` label set to a spec, but is calling `helper.CommonBeforeEach(true)`, but okay, it also makes sense to have a way to confirm in the code.

PR #6243, opened 2022-10-20
Fix e2e test after registry changes

feloy:
**What type of PR is this:**

/kind tests

### DIFF
--- a/tests/e2escenarios/e2e_test.go
+++ b/tests/e2escenarios/e2e_test.go
@@ -93,7 +93,7 @@ var _ = Describe("E2E Test", func() {
 
 			// "running odo list"
 			stdout := helper.Cmd("odo", "list", "component").ShouldPass().Out()
-			helper.MatchAllInOutput(stdout, []string{componentName, "nodejs", "Dev"})
+			helper.MatchAllInOutput(stdout, []string{componentName, "Node.js", "Dev"})
 
 			// "exit dev mode and run odo deploy"
 			devSession.Stop()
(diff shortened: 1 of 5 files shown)


tests/e2escenarios/e2e_test.go
```diff
@@ -93,7 +93,7 @@ var _ = Describe("E2E Test", func() {
 
 			// "running odo list"
 			stdout := helper.Cmd("odo", "list", "component").ShouldPass().Out()
-			helper.MatchAllInOutput(stdout, []string{componentName, "nodejs", "Dev"})
+			helper.MatchAllInOutput(stdout, []string{componentName, "Node.js", "Dev"})
```
valaparthvi, line 96: It looks like this should be left as is. Same for the change on L119.
```
  Expected
      <string>:  â€¢  Listing components from namespace 'e2e-test47grk'  ...
      
 âœ“  Listing components from namespace 'e2e-test47grk' [455ms]
       NAME      PROJECT TYPE  RUNNING IN  MANAGED      
       * nlbvnt  nodejs        Deploy      odo (v3.1.0) 
      
  to contain substring
      <string>: Node.js
  In [It] at: /go/odo_1/tests/helper/helper_generic.go:76
```

tests/e2escenarios/e2e_test.go
```diff
@@ -169,7 +169,7 @@ var _ = Describe("E2E Test", func() {
 			_, err := helper.RunInteractive(command, nil, func(ctx helper.InteractiveContext) {
 
 				// helper.ExpectString(ctx, "Based on the files in the current directory odo detected")
-				helper.ExpectString(ctx, "Language: javascript")
+				helper.ExpectString(ctx, "Language: javaScript")
```
valaparthvi, line 172:
```suggestion
				helper.ExpectString(ctx, "Language: JavaScript")
```

tests/e2escenarios/e2e_test.go
feloy: Thanks. It shows there is an inconsistency for the PROJECT value between Dev and Deploy'ed components. I'll create an issue to follow this.

valaparthvi:
I wonder if we should define these projectType and componentName values as constants. We have been going back and forth a lot lately, as a temporary solution, it might help to define constants.

feloy:
PVC problem

/NoCluster-Integration-tests/NoCluster-Integration-tests
Non applicable

feloy:
PVC problem

/override NoCluster-Integration-tests/NoCluster-Integration-tests
Non applicable

rm3l (approved):
/approve

PR #6245, opened 2022-10-21
add make release-bin target

kadel:
**What type of PR is this:**

Adds new `release-bin` make target.

This is for the homebrew package, and it could be also used by another packaging system.

The problem is that currently, [odo-dev formulae ](https://github.com/Homebrew/homebrew-core/blob/master/Formula/odo-dev.rb) uses `make bin`, this is a problem because this build odo with dev Segment key. This means that MacOS users installing odo via homebrew don't show up in our production telemetry. 

new `make release-bin` uses prod Segment key and it should be used only in packaging systems.

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ endif
 COMMON_GOFLAGS := -mod=vendor
 COMMON_LDFLAGS := -X $(PROJECT)/pkg/version.GITCOMMIT=$(GITCOMMIT)
 BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="$(COMMON_LDFLAGS)"
-CROSS_BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="-s -w -X $(PROJECT)/pkg/segment.writeKey=R1Z79HadJIrphLoeONZy5uqOjusljSwN $(COMMON_LDFLAGS)"
+RELEASE_BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="-s -w -X $(PROJECT)/pkg/segment.writeKey=R1Z79HadJIrphLoeONZy5uqOjusljSwN $(COMMON_LDFLAGS)"
 PKGS := $(shell go list $(COMMON_GOFLAGS)  ./... | grep -v $(PROJECT)/vendor | grep -v $(PROJECT)/tests)
 FILES := odo dist
 TIMEOUT ?= 14400s
@@ -67,6 +67,10 @@ help: ## Show this help.
 .PHONY: bin
 bin: ## build the odo binary
 	go build ${BUILD_FLAGS} cmd/odo/odo.go
+
+.PHONY: release-bin
+release-bin: ## build the odo binary
+	go build ${RELEASE_BUILD_FLAGS} cmd/odo/odo.go
 
 .PHONY: install
 install:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ endif
 COMMON_GOFLAGS := -mod=vendor
 COMMON_LDFLAGS := -X $(PROJECT)/pkg/version.GITCOMMIT=$(GITCOMMIT)
 BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="$(COMMON_LDFLAGS)"
-CROSS_BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="-s -w -X $(PROJECT)/pkg/segment.writeKey=R1Z79HadJIrphLoeONZy5uqOjusljSwN $(COMMON_LDFLAGS)"
+RELEASE_BUILD_FLAGS := $(COMMON_GOFLAGS) -ldflags="-s -w -X $(PROJECT)/pkg/segment.writeKey=R1Z79HadJIrphLoeONZy5uqOjusljSwN $(COMMON_LDFLAGS)"
```
rm3l, line 11: This is a different issue from the one addressed here, but I was wondering if it is safe to have this key public.

I didn't see anything on Segment website, but saw [this article](https://help.klaviyo.com/hc/en-us/articles/115005252988-How-to-Integrate-with-Segment) saying:

> Your write key is another private API key. Treat it like a password; keep it in a safe place and never expose it to the public.

rm3l:
> **What type of PR is this:**
> 
> Adds new `release-bin` make target.
> 
> This is for the homebrew package, and it could be also used by another packaging system.
> 
> The problem is that currently, [odo-dev formulae ](https://github.com/Homebrew/homebrew-core/blob/master/Formula/odo-dev.rb?rgh-link-date=2022-10-21T08%3A15%3A31Z) uses `make bin`, this is a problem because this build odo with dev Segment key. This means that MacOS users installing odo via homebrew don't show up in our production telemetry.
> 
Just curious, because I am not familiar with Homebrew :-)

Could the `odo-dev` formula download the right binaries (and checksums) directly from https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/odo (at least when a build from sources is not required)?

Makefile
kadel: If we want people to build the production `odo` binary from our public source code there is no way to make this not public.

I don't think that this key should be treated as a secret, the most that users can do with this is to send fake data to our telemetry, but no-one will be able to access any data with this key.

rm3l (approved):
> binary packages can't go to `homebrew-core`, we would have to move odo to `homebrew-cask` ([docs.brew.sh/Acceptable-Formulae#we-dont-like-binary-formulae](https://docs.brew.sh/Acceptable-Formulae#we-dont-like-binary-formulae))
> 
> homebrew-cask is primarily for GUI apps. I think that I have seen CLI apps there as well, but it is not common, and I think that, in general, it is not what homebrew users would expect.
> 
> But this is going to be a common thing for most package managers. As far as I know, every decent packaging system prefers if the packaged software is built from the source instead of downloading binary blobs from somewhere

Thanks for the clarifications.

/approve

feloy:
Non code-related changes
The failing tests are fixed by another PR

PR #6260, opened 2022-10-27
Fix potential issue preventing application from being restarted correctly by `odo dev`

rm3l:
**What type of PR is this:**
/kind bug
/area dev

For some reason, this happened with NodeJS starter projects for example,
where killing the parent pid for the application ('npm start')
would not stop the underlying 'node' process, causing the application
port to still being used.

Given the following process tree for the Node.JS application:
```
default       49  0.0  0.0  11924  2828 ?        Ss   20:23   0:00 /bin/sh -c echo $$ > /opt/odo/.odo_cmd_run.pid && cd ${PROJECT_SOURCE} &&  (npm start) 1>>/proc/1/fd/1 2>>/proc/1/fd/2; echo $? >> /opt/odo/.odo_cmd_run.pid
default       55  1.0  0.0 735940 62956 ?        Sl   20:23   0:00  \_npm start
default       66  0.0  0.0  11928  2760 ?        S    20:23   0:00    \_ sh /tmp/start-c40b5918.sh
default       67  0.5  0.0 624368 49772 ?        Sl   20:23   0:00      \_ node server.js
```
We were previously just killing the direct child process (pid 55 here),
and it used to also stop all its descendants. This appears to no longer
be the case.

So to make sure the application is stopped correctly,
we are now making sure to traverse the full parent-child tree
using a post-order strategy, to start killing children processes first.
So in the example above, we would start killing pids 67, 66 and 55 in
this order.

Issues with E2E tests e.g.: https://github.com/redhat-developer/odo/blob/main/tests/e2escenarios/e2e_test.go#L80-L85

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test (already tested here: https://github.com/redhat-developer/odo/blob/main/tests/e2escenarios/e2e_test.go#L80-L85)

- [ ] Documentation 

```shell
odo init --devfile nodejs --name my-node-app --starter nodejs-starter

odo dev

# When Dev Session is running, try to access the application
curl -i http://localhost:40001

# Make changes to HTTP response returned by `server.js`
# and make sure they are taken into account.
```

### DIFF
--- a/pkg/remotecmd/kubeexec.go
+++ b/pkg/remotecmd/kubeexec.go
@@ -179,6 +179,8 @@ func (k *kubeExecProcessHandler) StopProcessForCommand(
 		return err
 	}
 
+	klog.V(3).Infof("Found %d children (either direct and indirect) for parent process %d: %v", len(children), ppid, children)
+
 	if len(children) == 0 {
 		//TODO(rm3l): A length of 0 might indicate that there is no children file, which might happen if the host kernel
 		//was not built with the CONFIG_PROC_CHILDREN config.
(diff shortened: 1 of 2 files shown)


feloy:
I can see the problem fixed with nodejs devfile.

Thanks for the fix

/approve

PR #6266, opened 2022-10-28
resolve inconsistent order of registries in preference view

### LINKED ISSUE #6205: Inconsistent order of registries in `odo preference view` JSON output
/kind bug

**Operating System:**
Fedora 36

**Output of `odo version`:**
```
odo v3.0.0-rc2 (a450cec6e)

Server: https://10.10.10.105:6443
Kubernetes: v1.24.6
```

```shell
export GLOBALODOCONFIG=/tmp/my-odo-preferences.yaml

odo preference add registry StagingRegistry https://registry.stage.devfile.io
```

## Actual behavior

- Order in preferences file: `odo preference add registry` adds new entries to the end of the list.

```yaml
$ cat $GLOBALODOCONFIG

kind: Preference
apiversion: odo.dev/v1alpha1
OdoSettings:
  RegistryList:
    secure: false
  - Name: StagingRegistry
    URL: https://registry.stage.devfile.io
    secure: false
  ConsentTelemetry: true
```

- `odo preference view`: [explicit order](https://github.com/redhat-developer/odo/blob/main/pkg/odo/cli/preference/view.go#L90-L98) from last to first in human-readable output. The logic of iterating from last to first also seems to be duplicated in [`pkg/registry/registry.go`](https://github.com/redhat-developer/odo/blob/main/pkg/registry/registry.go#L64-L65)

```shell
$ odo preference view

Preference parameters:
 PARAMETER           VALUE 
 ConsentTelemetry    true  
 Ephemeral                 
 PushTimeout               
 RegistryCacheTime         
 Timeout                   
 UpdateNotification        

Devfile registries:
 NAME                    URL                                SECURE 
 StagingRegistry         https://registry.stage.devfile.io  No     
 DefaultDevfileRegistry  https://registry.devfile.io        No     
```
- `odo preference view -o json`: no order. We rely on the order of registries in the preferences file.

```json
{
  --- Truncated --

  "registries": [
    {
      "name": "DefaultDevfileRegistry",
      "url": "https://registry.devfile.io",
      "secure": false
    },
    {
      "name": "StagingRegistry",
      "url": "https://registry.stage.devfile.io",
      "secure": false
    }
  ]
}
```

## Expected behavior

Unless I missed something, I remember a discussion a few months ago where it was mentioned that the list of registries should be handled as last-in, first-out: the last registry added with `odo preference add registry` should have priority.
To me, this should be reflected consistently in the code base and in the related commands outputs.

I was also wondering if it wouldn't be safer that `odo preference add registry` simply adds the new entry to the beginning of the list, and have all other commands just rely on the order of the elements in the preferences file, without having to duplicate the iteration logic.
This would also make it clearer if users happen to manually edit the preferences file.

&mdash;

anandrkskd:

**What type of PR is this:**

/kind bug

This PR resolves the inconsistency in output of registry list in command `odo preference view` and `odo preference view -o json`

Fixes #6205 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

```
odo preference add registry CheRegistry https://che-devfile-registry.openshift.io

odo preference view 
odo preference view -o json
```
Order of registry list should be same in `odo preference view` `odo preference view -o json` with latest added registry on the top.

### DIFF
--- a/pkg/odo/cli/preference/view.go
+++ b/pkg/odo/cli/preference/view.go
@@ -64,18 +64,9 @@ func (o *ViewOptions) RunForJsonOutput(ctx context.Context) (result interface{},
 	preferenceList := o.clientset.PreferenceClient.NewPreferenceList()
 	registryList := o.clientset.PreferenceClient.RegistryList()
 
-	var regList, newRegistryList []preference.Registry
-	if registryList != nil {
-		regList = *registryList
-	}
-	// Loop backwards here to ensure the registry display order is correct (display latest newly added registry firstly)
-	for i := len(regList) - 1; i >= 0; i-- {
-		newRegistryList = append(newRegistryList, regList[i])
-	}
-
 	return api.PreferenceView{
 		Preferences: preferenceList.Items,
-		Registries:  newRegistryList,
+		Registries:  *registryList,
 	}, nil
 }
 
@@ -98,7 +89,7 @@ func HumanReadableOutput(preferenceList preference.PreferenceList, registryList 
 		regList = *registryList
 	}
 	// Loop backwards here to ensure the registry display order is correct (display latest newly added registry firstly)
-	for i := len(regList) - 1; i >= 0; i-- {
+	for i := range regList {
 		registry := regList[i]
 		secure := "No"
 		if registry.Secure {
--- a/pkg/odo/cli/preference/view.go
+++ b/pkg/odo/cli/preference/view.go
@@ -63,9 +63,19 @@ func (o *ViewOptions) Run(ctx context.Context) (err error) {
 func (o *ViewOptions) RunForJsonOutput(ctx context.Context) (result interface{}, err error) {
 	preferenceList := o.clientset.PreferenceClient.NewPreferenceList()
 	registryList := o.clientset.PreferenceClient.RegistryList()
+
+	var regList, newRegistryList []preference.Registry
+	if registryList != nil {
+		regList = *registryList
+	}
+	// Loop backwards here to ensure the registry display order is correct (display latest newly added registry firstly)
+	for i := len(regList) - 1; i >= 0; i-- {
+		newRegistryList = append(newRegistryList, regList[i])
+	}
+
 	return api.PreferenceView{
 		Preferences: preferenceList.Items,
-		Registries:  *registryList,
+		Registries:  newRegistryList,
 	}, nil
 }
 
--- a/pkg/odo/cli/preference/view.go
+++ b/pkg/odo/cli/preference/view.go
@@ -84,13 +84,9 @@ func HumanReadableOutput(preferenceList preference.PreferenceList, registryList 
 	registryT := ui.NewTable()
 	registryT.AppendHeader(table.Row{"NAME", "URL", "SECURE"})
 
-	var regList []preference.Registry
-	if registryList != nil {
-		regList = *registryList
-	}
 	// Loop backwards here to ensure the registry display order is correct (display latest newly added registry firstly)
-	for i := range regList {
-		registry := regList[i]
+	for i := range *registryList {
+		registry := (*registryList)[i]
 		secure := "No"
 		if registry.Secure {
 			secure = "Yes"
--- a/pkg/preference/implem.go
+++ b/pkg/preference/implem.go
@@ -133,6 +133,19 @@ func newPreferenceInfo() (*preferenceInfo, error) {
 		return nil, err
 	}
 
+	regList := []Registry{}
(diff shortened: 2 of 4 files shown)


rm3l (changes requested):
@anandrkskd Can you add a test case for this (to catch any potential future regressions on this)?

pkg/odo/cli/preference/view.go
```diff
@@ -63,9 +63,19 @@ func (o *ViewOptions) Run(ctx context.Context) (err error) {
...
 	registryList := o.clientset.PreferenceClient.RegistryList()
+
+	var regList, newRegistryList []preference.Registry
+	if registryList != nil {
+		regList = *registryList
+	}
+	// Loop backwards here to ensure the registry display order is correct (display latest newly added registry firstly)
+	for i := len(regList) - 1; i >= 0; i-- {
+		newRegistryList = append(newRegistryList, regList[i])
+	}
```
feloy, line 74: The CLI level should not have to handle this, it is the responsability of the business level to return the correct information. This way, we should not have to duplicate this code here and in `HumanReadableOutput`.

Also, there is a nice feature in go to exchange two values: `a, b = b, a`. You could use it to reverse the order of the array in place (an elegant solution here: https://stackoverflow.com/a/53736310).

feloy (changes requested):

anandrkskd:
> @anandrkskd Can you add a test case for this (to catch any potential future regressions on this)?
Thanks, @rm3l, added the test.

anandrkskd:
Thanks Philippe, rebased the PR.

feloy:
/override v4.11-integration-e2e
Tests pass on IBM Cloud

pkg/preference/implem.go
```diff
@@ -133,6 +133,19 @@ func newPreferenceInfo() (*preferenceInfo, error) {
...
+	}
+
+	i := 0
+	j := len(regList) - 1
+	for i < j {
+		regList[i], regList[j] = regList[j], regList[i]
+		i++
+		j--
+	}
+
```
rm3l, line 148: @anandrkskd Now that the Preference Client implementation does the reverse operation (which is a good thing), I think there is one side effect, which we should consider, IMO.
In [`pkg/registry/registry.go`](https://github.com/redhat-developer/odo/blob/main/pkg/registry/registry.go#L64-L65), the registry list is also iterated from the end, which again returns the list in the wrong order, for example when running `odo registry` or even when fetching a new Devfile with `odo init` :)

<details>
<summary>Examples</summary>

```sh
# Correct order: StagingRegistry comes first. Same with JSON output.
$ odo preference view

...
Devfile registries:
 NAME                    URL                                SECURE 
 StagingRegistry         https://registry.stage.devfile.io  No     
 DefaultDevfileRegistry  https://registry.devfile.io        No     
```

```sh
# Issue 1: For each Devfile name, StagingRegistry should be listed first
$ odo registry

...
NAME                          REGISTRY                DESCRIPTION                                 
 dotnet50                      DefaultDevfileRegistry  Stack with .NET 5.0                         
 dotnet50                      StagingRegistry         Stack with .NET 5.0                         
 dotnet60                      DefaultDevfileRegistry  Stack with .NET 6.0                         
 dotnet60                      StagingRegistry         Stack with .NET 6.0                         
 dotnetcore31                  DefaultDevfileRegistry  Stack with .NET Core 3.1                    
 dotnetcore31                  StagingRegistry         Stack with .NET Core 3.1                    
 go                            DefaultDevfileRegistry  Go is an open source programming languag... 
 go                            StagingRegistry         Go is an open source programming language... 
```

```sh
# Issue 2: Devfile should have been selected from "StagingRegistry"
$ odo init           
  __
 \__/  \    Files: Source code detected, a Devfile will be determined based upon source code autodetection
 /  \__/    odo version: v3.2.0
 \__/

Language: Java
Project type: Maven
The devfile "java-maven" from the registry "DefaultDevfileRegistry" will be downloaded.
? Is this correct? (Y/n) 
```

</details>


So I think we need to change [`pkg/registry/registry.go`](https://github.com/redhat-developer/odo/blob/main/pkg/registry/registry.go#L64-L65) to rely on the order returned by the Preference Client.

rm3l: And we should also make sure that there are no other places where the list returned by the Preference Client is iterated in the reverse order ^^

rm3l: Since this PR is already merged in, I've called this out in a separate issue (https://github.com/redhat-developer/odo/issues/6283). That would be great if you could take a look. Thanks.

PR #6276, opened 2022-10-31
Bump odo to v3.2.0

feloy:


build/VERSION
```diff
@@ -1 +1 @@
-v3.0.0
\ No newline at end of file
+v3.2.0
```
rm3l, line 1: I think we should update this file only after all binaries are published to the Content Gateway, no?
Otherwise, I guess users would get notified before the actual binaries are available.

feloy: My bad, I reverted it. Thanks

rm3l (approved):
/approve

rm3l:
Overriding Prow - All tests are passing in IBM Cloud.

PR #6281, opened 2022-11-02
Single command is default

kadel:
**What type of PR is this:**

/kind bug

Fixes https://github.com/redhat-developer/odo/issues/6280

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 


pkg/libdevfile/libdevfile.go
```diff
@@ -114,6 +114,10 @@ func getDefaultCommand(devfileObj parser.DevfileObj, groupType v1alpha2.CommandG
 		}
 	}
 
+	// if there is only one command of a given group kind, use it as default
+	if len(commands) == 1 {
+		return commands[0], true, nil
+	}
```
rm3l, line 120: Can we move this block before the `for` loop (even before L104, where `defaultCmds` is declared and initialized`)? It would avoid iterating for nothing if there is only one command.

rm3l (commented):
Feature-wise, this means that we would also be executing the `build` command if there is only one command of that kind (regardless of the `isDefault` field). Is that okay?

PR #6284, opened 2022-11-04
Sanitize value for 'app.openshift.io/runtime' label

### LINKED ISSUE #6234: `odo dev` fails if
/kind bug
```
▶ odo init --devfile-path https://raw.githubusercontent.com/devfile/registry/269dc7a2f0e8b19fbd53250e858a895f6aa93c62/stacks/dotnet60/devfile.yaml --starter dotnet60-example --name dotnet60-example


▶ odo dev
  __
 /  \__     Developing using the "dotnet60-example" Devfile
 \__/  \    Namespace: default
 /  \__/    odo version: v3.1.0
 \__/

 ✗  unable to create or update component: unable to create PVC: unable to create PVC: PersistentVolumeClaim "odo-projects-dotnet60-example-app" is invalid: metadata.labels: Invalid value: ".NET": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

the devfile.yaml has `projectType: .NET` and odo is generating k8s rousrces with ` app.openshift.io/runtime` label with value that is that same as  `projectType`

odo should clean `projectType` value first to make sure that it can be used as in labels.

/priority high

rm3l:
/kind bug

This sanitizes the `runtime` value used for setting the `app.openshift.io/runtime` label value (added in https://github.com/redhat-developer/odo/issues/6044) so that it is compliant with [what is expected by Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). This was based upon the Devfile `metadata.projectType` or `metadata.language` fields, which are free-form strings.

Sanitization is done as follows:
- Leading and trailing non-alphanumeric characters are either replaced or removed:
  - if the value starts or ends with well-known characters (`.` or `#` for now), it replaces them with strings (like `dot` or `sharp`, respectively), and the case depends on the case of the remaining characters of the `runtime` value. For example, this would replace `.NET` with `DOTNET` and `.net` with `dotnet`.
  - If the value starts with other non-alphanumeric characters, they are removed
- Dashes (`-`), underscores (`_`), dots (`.`), and alphanumerics between are kept as is. Any other character is replaced with a dash (`-`)
- it truncates the resulting value if it is more than 63 characters

See the unit tests added for a better understanding.

This PR only addresses the issue with the `app.openshift.io/runtime` label value, as I noticed that the other labels added by us are either hard-coded or already checked for compliance (like the Devfile `metadata.name` field). 

Fixes #6234

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

See the unit tests added and the Devfiles added for the integration tests.

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -32,7 +32,8 @@ var (
 func GetLabels(componentName string, applicationName string, runtime string, mode string, isPartOfComponent bool) map[string]string {
 	labels := getLabels(componentName, applicationName, mode, true, isPartOfComponent)
 	if runtime != "" {
-		labels[openshiftRunTimeLabel] = sanitizeLabelValue(openshiftRunTimeLabel, runtime)
+		// 'app.openshift.io/runtime' label added by OpenShift console is always lowercase
+		labels[openshiftRunTimeLabel] = strings.ToLower(sanitizeLabelValue(openshiftRunTimeLabel, runtime))
 	}
 	return labels
 }
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -2,10 +2,27 @@ package labels
 
 import (
 	"errors"
+	"regexp"
+	"strings"
+	"unicode"
 
-	"k8s.io/apimachinery/pkg/labels"
+	dfutil "github.com/devfile/library/pkg/util"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/klog"
 
 	"github.com/redhat-developer/odo/pkg/version"
+)
+
+var _replacementMap = map[string]string{
+	".": "dot",
+	"#": "sharp",
+}
+
+var (
+	_regexpInvalidCharacters        = regexp.MustCompile(`[^a-zA-Z0-9._-]`)
+	_regexpStartingWithAlphanumeric = regexp.MustCompile(`^[a-z0-9A-Z]`)
+	_regexpEndingWithAlphanumeric   = regexp.MustCompile(`[a-z0-9A-Z]$`)
 )
 
 // GetLabels return labels that should be applied to every object for given component in active application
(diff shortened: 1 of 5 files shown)


pkg/labels/labels.go
```diff
@@ -2,24 +2,146 @@ package labels
...
+//
+// - must be 63 characters or fewer (can be empty),
+//
+// - unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
+//
+// - could contain dashes (-), underscores (_), dots (.), and alphanumerics between.
+//
+// As such, sanitizeLabelValue might:
+//
+// - truncate the label value so that it is less than 63 characters,
```
feloy, line 53: You can move this point after the replcments, as it is done after. Also, you could add a test that checks that a long string with . or # is truncated to 63 after replacement (and not before).

feloy (changes requested):

pkg/labels/labels.go
rm3l: Done in [`d6941ee` (#6284)](https://github.com/redhat-developer/odo/pull/6284/commits/d6941eeb792bf30a338ead7f76de61ac701dbd41) - thanks for pointing this out.

pkg/labels/labels.go
```diff
@@ -2,24 +2,139 @@ package labels
...
+
 // GetLabels return labels that should be applied to every object for given component in active application
 // if you need labels to filter component then use GetSelector instead
 // Note: isPartOfComponent denotes if the label is required for a core resource(deployment, svc, pvc, pv) of a given component deployed with `odo dev`;
 // it is the only thing that sets it apart from the resources created via other ways (`odo deploy`, deploying resource with apply command during `odo dev`)
 func GetLabels(componentName string, applicationName string, runtime string, mode string, isPartOfComponent bool) map[string]string {
 	labels := getLabels(componentName, applicationName, mode, true, isPartOfComponent)
 	if runtime != "" {
-		labels[openshiftRunTimeLabel] = runtime
+		labels[openshiftRunTimeLabel] = sanitizeLabelValue(openshiftRunTimeLabel, runtime)
```
kadel, line 35: I've noticed that `app.openshift.io/runtime` added by OpenShift console is always lowercase. we probably should lovercase the result of sanitizeLabelValue here

rm3l: Okay, I'll lowercase the result then. Thanks for the information.

rm3l: Done in [`7257a27` (#6284)](https://github.com/redhat-developer/odo/pull/6284/commits/7257a27a2d3e80cf4abd024592b82c767cafd45c)

rm3l:

Infra errors on Prow, and all tests passing on IBM Cloud.

feloy (approved):

PR #6295, opened 2022-11-08
Adds ability to list namespaces in JSON format

### LINKED ISSUE #6257: '-o json' support for 'odo list namespace'
/kind feature

$subj

## Why is this needed?

For IDEs integration.

dharmit:

/kind feature

**What type of PR is this:**

Fixes #6257 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/cli/list/namespace/namespace.go
+++ b/pkg/odo/cli/list/namespace/namespace.go
@@ -93,6 +93,16 @@ func (plo *NamespaceListOptions) RunForJsonOutput(_ context.Context) (out interf
 // NewCmdNamespaceList implements the odo list project command.
 func NewCmdNamespaceList(name, fullName string) *cobra.Command {
 	o := NewNamespaceListOptions()
+	// To help the UI messages deal better with namespace vs project
+	o.commandName = name
+	if len(os.Args) > 2 {
+		o.commandName = os.Args[2]
+	}
+	// trim commandName if user passed a plural form
+	lastElement := len(o.commandName) - 1
+	if o.commandName[lastElement] == byte('s') {
+		o.commandName = o.commandName[:lastElement]
+	}
 	projectListCmd := &cobra.Command{
 		Use:     name,
 		Short:   listLongDesc,
--- a/pkg/odo/cli/list/namespace/namespace.go
+++ b/pkg/odo/cli/list/namespace/namespace.go
@@ -7,13 +7,14 @@ import (
 	"os"
 	"text/tabwriter"
 
+	"github.com/redhat-developer/odo/pkg/api"
+	"github.com/redhat-developer/odo/pkg/odo/commonflags"
 	"github.com/spf13/cobra"
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
 
 	"github.com/redhat-developer/odo/pkg/odo/cmdline"
 	"github.com/redhat-developer/odo/pkg/odo/genericclioptions"
 	"github.com/redhat-developer/odo/pkg/odo/genericclioptions/clientset"
-	"github.com/redhat-developer/odo/pkg/project"
 )
 
 const RecommendedCommandName = "namespace"
@@ -57,13 +58,36 @@ func (plo *NamespaceListOptions) Validate(ctx context.Context) (err error) {
 }
 
 // Run contains the logic for the odo list project command
-func (plo *NamespaceListOptions) Run(ctx context.Context) error {
-	namespaces, err := plo.clientset.ProjectClient.List()
+func (plo *NamespaceListOptions) Run(_ context.Context) error {
+	namespaces, err := plo.run()
 	if err != nil {
 		return err
 	}
 
 	return HumanReadableOutput(os.Stdout, namespaces, plo.commandName)
+}
+
+func (plo *NamespaceListOptions) run() (api.ResourcesList, error) {
+	namespaces, err := plo.clientset.ProjectClient.List()
+	if err != nil {
+		return api.ResourcesList{}, err
+	}
+
+	if len(namespaces.Items) == 0 {
+		return api.ResourcesList{}, fmt.Errorf("you are not a member of any %[1]ss. You can request a %[1]s to be created using the `odo create %[1]s <%[1]s_name>` command", plo.commandName)
+	}
+
+	var projects []api.Project
+	for _, proj := range namespaces.Items {
+		project := api.Project{Name: proj.Name, Active: proj.Status.Active}
+		projects = append(projects, project)
+	}
+
+	return api.ResourcesList{Namespaces: projects}, nil
+}
+
+func (plo *NamespaceListOptions) RunForJsonOutput(_ context.Context) (out interface{}, err error) {
+	return plo.run()
 }
 
 // NewCmdNamespaceList implements the odo list project command.
@@ -81,19 +105,21 @@ func NewCmdNamespaceList(name, fullName string) *cobra.Command {
 		Aliases: []string{"namespaces", "project", "projects"},
 	}
 	clientset.Add(projectListCmd, clientset.PROJECT)
+	commonflags.UseOutputFlag(projectListCmd)
+
 	return projectListCmd
 }
 
 // HumanReadableOutput outputs the list of namespaces in a human readable format
-func HumanReadableOutput(w io.Writer, o project.ProjectList, commandName string) error {
-	if len(o.Items) == 0 {
+func HumanReadableOutput(w io.Writer, o api.ResourcesList, commandName string) error {
+	if len(o.Namespaces) == 0 {
 		return fmt.Errorf("you are not a member of any %[1]ss. You can request a %[1]s to be created using the `odo create %[1]s <%[1]s_name>` command", commandName)
 	}
 	wr := tabwriter.NewWriter(w, 5, 2, 3, ' ', tabwriter.TabIndent)
 	fmt.Fprintln(wr, "ACTIVE", "\t", "NAME")
-	for _, project := range o.Items {
+	for _, project := range o.Namespaces {
 		activeMark := " "
-		if project.Status.Active {
+		if project.Active {
 			activeMark = "*"
 		}
 		fmt.Fprintln(wr, activeMark, "\t", project.Name)
--- a/pkg/odo/cli/list/namespace/namespace.go
+++ b/pkg/odo/cli/list/namespace/namespace.go
@@ -59,12 +59,12 @@ func (plo *NamespaceListOptions) Validate(ctx context.Context) (err error) {
 
 // Run contains the logic for the odo list project command
 func (plo *NamespaceListOptions) Run(_ context.Context) error {
-	namespaces, err := plo.run()
+	resourcesList, err := plo.run()
 	if err != nil {
 		return err
 	}
 
-	return HumanReadableOutput(os.Stdout, namespaces, plo.commandName)
+	return HumanReadableOutput(os.Stdout, resourcesList, plo.commandName)
 }
 
 func (plo *NamespaceListOptions) run() (api.ResourcesList, error) {
@@ -74,7 +74,7 @@ func (plo *NamespaceListOptions) run() (api.ResourcesList, error) {
 	}
 
 	if len(namespaces.Items) == 0 {
-		return api.ResourcesList{}, fmt.Errorf("you are not a member of any %[1]ss. You can request a %[1]s to be created using the `odo create %[1]s <%[1]s_name>` command", plo.commandName)
+		return api.ResourcesList{}, nil
 	}
 
 	var projects []api.Project
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -183,6 +183,13 @@ ComponentSettings:
 					return out
 				}, 10*time.Second, 1*time.Second).Should(ContainSubstring(commonVar.Project))
 			})
+			It(fmt.Sprintf("should successfully list all the %ss in JSON format", commandName), func() {
+				Eventually(func() bool {
+					out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
+					helper.MatchAllInOutput(out, []string{commonVar.Project, "\"active\": true", "\"active\": false"})
+					return true
+				}, 10*time.Second, 1*time.Second).Should(Equal(true))
+			})
 		})
 	}
 })
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"golang.org/x/text/cases"
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/tidwall/gjson"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 
@@ -191,7 +192,10 @@ ComponentSettings:
 				}).WithTimeout(10 * time.Second).Should(BeTrue())
 
 				out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
-				helper.MatchAllInOutput(out, []string{commonVar.Project, "\"active\": true", "\"active\": false"})
+				Expect(helper.IsJSON(out)).To(BeTrue())
+				// check if the namespace/project created for this test is marked as active in the JSON output
+				gjsonStr := fmt.Sprintf("namespaces.#[name==%s].active", commonVar.Project)
+				Expect(gjson.Get(out, gjsonStr).String()).To(Equal("true"))
 			})
 		})
 	}
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -196,6 +196,8 @@ ComponentSettings:
 				// check if the namespace/project created for this test is marked as active in the JSON output
 				gjsonStr := fmt.Sprintf("namespaces.#[name==%s].active", commonVar.Project)
 				Expect(gjson.Get(out, gjsonStr).String()).To(Equal("true"))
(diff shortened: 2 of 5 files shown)


dharmit:
Real failure in the tests added in this PR. Looking into it:

<details>

<summary> Log </summary>

```
   Expected
      <string>: {
      	"namespaces": [
      		{
      			"name": "cmd-add-binding-test128cfn",
      			"active": false
      		},
      		{
      			"name": "cmd-describe-list-binding-test626yvx",
      			"active": false
      		},
      		{
      			"name": "cmd-dev-test1992ynr",
      			"active": false
      		},
      		{
      			"name": "cmd-dev-test831yfw",
      			"active": false
      		},
      		{
      			"name": "cmd-dev-test89dva",
      			"active": false
      		},
      		{
      			"name": "default",
      			"active": false
      		},
      		{
      			"name": "myproject",
      			"active": false
      		},
      		{
      			"name": "nodejs-12",
      			"active": false
      		},
      		{
      			"name": "nodejs-12-rhel7",
      			"active": false
      		},
      		{
      			"name": "nodejs-14",
      			"active": false
      		},
      		{
      			"name": "openjdk-11",
      			"active": false
      		},
      		{
      			"name": "openjdk-11-rhel8",
      			"active": false
      		}
      	]
      }
      
  to contain substring
      <string>: cmd-namespace-test183lov 
```

</details>

dharmit:
Wondering if this could be the race condition wherein `odo list namespaces -o json` is being executed too fast. Thinking this way because tests don't fail every time when run locally, and also observed this that new project created by `oc` command doesn't show up in the output:

```sh
$ oc new-project test && odo list namespaces -o json
Already on project "test" on server "https://api.crc.testing:6443".

You can add applications to this project with the 'new-app' command. For example, try:

    oc new-app rails-postgresql-example

to build a new example application in Ruby. Or use kubectl to deploy a simple Kubernetes application:

    kubectl create deployment hello-node --image=k8s.gcr.io/e2e-test-images/agnhost:2.33 -- /agnhost serve-hostname

{
	"namespaces": [
		{
			"name": "myproject",
			"active": false
		},
		{
			"name": "newproject",
			"active": false
		}
	]
}
```

tests/integration/cmd_namespace_test.go
```diff
@@ -180,6 +180,13 @@ ComponentSettings:
 					return out
 				}, 10*time.Second, 1*time.Second).Should(ContainSubstring(commonVar.Project))
 			})
+			It(fmt.Sprintf("should successfully list all the %ss in JSON format", commandName), func() {
+				Eventually(func() bool {
+					out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
+					helper.MatchAllInOutput(out, []string{commonVar.Project, "\"active\": true", "\"active\": false"})
```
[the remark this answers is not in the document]
feloy: As the output is in JSON, it is possible (and preferable) to use the JSON related functions (helper.JsonPathContentIs, ...)

If it is too difficult using gomega operators, you can do the check into the function (without any gomega.Expect), and return true when the result is ok, and just expect a True in Eventually(...).Should()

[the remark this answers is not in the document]
dharmit: @feloy it seems like even with `helper.JsonPathContentContain`, test immediately exit instead of retrying (as we would expect with `Eventually`). For below diff, the test immediately exits with failure:

```diff
diff --git a/tests/integration/cmd_namespace_test.go b/tests/integration/cmd_namespace_test.go
index 3439c6aa1..bf8f7f4c8 100644
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -176 +176 @@ ComponentSettings:
-               Describe("list "+commandName, func() {
+               FDescribe("list "+commandName, func() {
@@ -186 +186 @@ ComponentSettings:
-                                       helper.MatchAllInOutput(out, []string{commonVar.Project, "\"active\": true", "\"active\": false"})
+                                       helper.JsonPathContentContain(out, "namespaces", commonVar.Project)
```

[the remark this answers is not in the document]
feloy: @dharmit You will need to get inspiration from the implementation of these functions, but without any Expect, as this is what makes it fail immediately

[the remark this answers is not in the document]
dharmit: So I managed to make progress by nesting `Eventually` like this:
```go
It(fmt.Sprintf("should successfully list all the %ss in JSON format", commandName), func() {
	Eventually(func() bool {
		Eventually(func() bool {
			return commonVar.CliRunner.HasNamespaceProject(commonVar.Project)
		}, 10*time.Second, 1*time.Second).Should(Equal(true))

		out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
		helper.JsonPathContentContain(out, "namespaces", commonVar.Project)
		return true
	}, 10*time.Second, 1*time.Second).Should(Equal(true))
})
```
but it's still failing. I don't think the `Eventually` aspect is being respected, in that it just fails if it can't find the namespace. It doesn't seem to be trying a bunch of times till `10*time.Second` is reached. 😕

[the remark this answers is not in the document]
feloy: > @dharmit You will need to get inspiration from the implementation of these functions, but without any Expect, as this is what makes it fail immediately

As said before, you cannot use Expect inside the Eventually function, or it will raise the error immediately. Here, you are still using it (inside `JsonPathContentContain`).

You will need to pass a Gomega-free function to Eventually, only returning true/false, and Expect'ing on this true/false value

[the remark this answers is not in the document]
dharmit: > If it is too difficult using gomega operators, you can do the check into the function (without any gomega.Expect), and return true when the result is ok, and just expect a True in Eventually(...).Should()

Wish I had read your above command properly. @kadel helped me understand it better. 😞 
Fixed the test issue (at least locally.) 😕

dharmit:
Weird to see this error in golangci-lint. Passes locally with same version of the tool:
```
golangci-lint run ./... --timeout 15m
tests/integration/cmd_namespace_test.go:190:14: undeclared name: `strings` (typecheck)
						return strings.Contains(ns, commonVar.Project) 
```

/test unit-and-validate-test

Looks like infra error for 4.11 tests
/test v4.11-integration-e2e

pkg/odo/cli/list/namespace/namespace.go
```diff
@@ -57,15 +58,38 @@ func (plo *NamespaceListOptions) Validate(ctx context.Context) (err error) {
...
 }
 
+func (plo *NamespaceListOptions) run() (api.ResourcesList, error) {
+	namespaces, err := plo.clientset.ProjectClient.List()
+	if err != nil {
+		return api.ResourcesList{}, err
+	}
+
+	if len(namespaces.Items) == 0 {
+		return api.ResourcesList{}, fmt.Errorf("you are not a member of any %[1]ss. You can request a %[1]s to be created using the `odo create %[1]s <%[1]s_name>` command", plo.commandName)
```
rm3l, line 77: I was testing this, and noticed that the error message displayed here can be a bit confusing:

```
$ odo list namespaces
 ✗  you are not a member of any s. You can request a  to be created using the `odo create  <_name>` command

```

It looks like `plo.commandName` is not being set. I guess we'll need to do something similar to what is done in `pkg/odo/cli/create/namespace/namespace.go`:  https://github.com/redhat-developer/odo/blob/f96c35ac0ee2970eb873bacd2b8af097f4be2110/pkg/odo/cli/create/namespace/namespace.go#L116-L120

rm3l (changes requested):

pkg/odo/cli/list/namespace/namespace.go
rm3l: Also, even with the changes to set `plo.commandName`, I think we might still have an issue with the wording of the error message if users make use of the plural form of the alias, given the string format (`%[1]ss`):

```
x odo list projects
 ✗  you are not a member of any projectss. You can request a projects to be created using the `odo create projects <projects_name>` command
```

dharmit: @rm3l thanks for catching this and for the tip to fix it. I have pushed the fix.

dharmit:
I'm seeing this error often lately. @feloy @anandrkskd WDYT about it?
```
Unexpected error:
      <*url.Error | 0xc0002664b0>: {
          Op: "Get",
          URL: "http://127.0.0.1:39501/",
          Err: <*errors.errorString | 0xc0001180f0>{s: "EOF"},
      }
      Get "http://127.0.0.1:39501/": EOF
  occurred
  In [It] at: /go/src/github.com/redhat-developer/odo/tests/e2escenarios/e2e_test.go:32
Summarizing 1 Failure:
  [FAIL] E2E Test starting with non-empty Directory [It] should verify developer workflow from non-empty Directory
  /go/src/github.com/redhat-developer/odo/tests/e2escenarios/e2e_test.go:32
```

Unit tests failing on CI with below error:
```
 golangci-lint run ./... --timeout 15m
tests/integration/cmd_namespace_test.go:189:13: undeclared name: `strings` (typecheck)
					return strings.Contains(ns, commonVar.Project)
					       ^ 
```
Locally, `make validate` fails with a whole different bunch of errors in spite of the same version of `golangci-lint`. Also, the unit tests pass on IBM Cloud:
```
golangci-lint run ./... --timeout 15m
pkg/storage/storage.go:73:2: ineffectual assignment to storageClusterList (ineffassign)
	storageClusterList := StorageList{}
	^
pkg/log/status.go:249: File is not `gofmt`-ed with `-s` (gofmt)
//  ✓  <message>
tests/integration/cmd_namespace_test.go:72:53: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
						ContainSubstring(fmt.Sprintf("%s %q deleted", strings.Title(commandName), namespace)))
						                                              ^
pkg/devfile/image/docker_compatible.go:118: File is not `gofmt`-ed with `-s` (gofmt)
//getShellCommand creates the docker compatible build command from detected backend,
//container image and devfile path
pkg/libdevfile/libdevfile.go:447: File is not `gofmt`-ed with `-s` (gofmt)
//Inspired from variables.validateAndReplaceDataWithVariable, which is unfortunately not exported
pkg/odo/cli/create/namespace/namespace.go:97:58: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
	successMessage := fmt.Sprintf(`%s %q is ready for use`, strings.Title(nco.commandName), nco.namespaceName)
	                                                        ^
pkg/api/component.go:32:24: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
			strs = append(strs, strings.Title(string(s)))
			                    ^
pkg/odo/util/cmdutils.go:29:40: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
				printstring := fmt.Sprintf("%s%s", strings.Title(context), "\nError: %v")
				                                   ^
pkg/kclient/fakeclient.go:61: File is not `gofmt`-ed with `-s` (gofmt)
//FakePodStatus returns a pod with the status
pkg/util/util_test.go:1418:39: SA9002: file mode '644' evaluates to 01204; did you mean '0644'? (staticcheck)
				if e := os.Mkdir(dir, os.FileMode(644)); e != nil {
				                                  ^
pkg/odo/cli/delete/namespace/namespace.go:97:4: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
			strings.Title(do.commandName), do.namespaceName)
			^
pkg/devfile/location/location.go:11: File is not `gofmt`-ed with `-s` (gofmt)
//  possibleDevfileNames contains possivle devfile name that should be checked in the context dir.
tests/helper/helper_telemetry.go:19: File is not `gofmt`-ed with `-s` (gofmt)
//EnableTelemetryDebug creates a temp file to use for debugging telemetry.
//it also sets up envs and cfg for the same
make: *** [Makefile:100: golint] Error 1
```
/test unit-and-validate-test

pkg/odo/cli/list/namespace/namespace.go
```diff
@@ -57,18 +58,51 @@ func (plo *NamespaceListOptions) Validate(ctx context.Context) (err error) {
...
```
rm3l, line 77: Just noticed that `<` and `>` characters are escaped when marshaling the JSON output.

```
$ odo list namespaces -o json
{
        "message": "you are not a member of any namespaces. You can request a namespace to be created using the `odo create namespace \u003cnamespace_name\u003e` command"
}
```

From [this post on SO](https://stackoverflow.com/questions/28595664/how-to-stop-json-marshal-from-escaping-and), it seems this is the behavior of `json.Marshal`, but might be customized using a custom encoder.
I'd suggest removing those two characters here:

```go
return api.ResourcesList{}, fmt.Errorf("you are not a member of any %[1]ss. You can request a %[1]s to be created using the 'odo create %[1]s $%[1]s' command", plo.commandName)
```

What do you think?

tests/integration/cmd_namespace_test.go
```diff
@@ -183,6 +184,15 @@ ComponentSettings:
...
 				}, 10*time.Second, 1*time.Second).Should(ContainSubstring(commonVar.Project))
 			})
+			It(fmt.Sprintf("should successfully list all the %ss in JSON format", commandName), func() {
+				Eventually(func() bool {
+					ns := helper.Cmd("odo", "list", commandName).ShouldPass().Out()
+					return strings.Contains(ns, commonVar.Project)
+				}).WithTimeout(10 * time.Second).Should(BeTrue())
+
+				out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
+				helper.MatchAllInOutput(out, []string{commonVar.Project, "\"active\": true", "\"active\": false"})
```
rm3l, line 194: This check is fine but, to me, it does not guarantee that the output is valid JSON. Maybe we could also check that using the `helper.IsJSON` function?

Also, to check that the current active namespace is marked as 'active' in the output, maybe we could rely on some [advanced filtering of `gjson`](https://github.com/tidwall/gjson#path-syntax), by extracting the `active` field of the current project using `namespaces.#[name=="$commonVar.Project"].active` and check that the result is `true`?

Using this would also guarantee that the output is also valid JSON.

pkg/odo/cli/list/namespace/namespace.go
dharmit: > What do you think?

This comment made me look at other `odo list` commands' JSON outputs more consciously, and below is what I see:
```sh
$ odo list components -o json
{}

$ odo list bindings -o json  
{}

$ odo list services -o json
{}

$ odo list namespaces -o json
{
	"message": "you are not a member of any namespaces. You can request a namespace to be created using the `odo create namespace \u003cnamespace_name\u003e` command"
}

```
I'd prefer consistency in JSON outputs, and hence shows just a `{}` as output if the user doesn't have access to any namespaces on the cluster.

WDYT?

dharmit: PTAL 61b8479.

tests/integration/cmd_namespace_test.go
dharmit: Makes sense. PTAL 23cc5b2.

tests/integration/cmd_namespace_test.go
```diff
@@ -183,6 +185,18 @@ ComponentSettings:
...
+				Eventually(func() bool {
+					ns := helper.Cmd("odo", "list", commandName).ShouldPass().Out()
+					return strings.Contains(ns, commonVar.Project)
+				}).WithTimeout(10 * time.Second).Should(BeTrue())
+
+				out := helper.Cmd("odo", "list", commandName, "-o", "json").ShouldPass().Out()
+				Expect(helper.IsJSON(out)).To(BeTrue())
+				// check if the namespace/project created for this test is marked as active in the JSON output
+				gjsonStr := fmt.Sprintf("namespaces.#[name==%s].active", commonVar.Project)
+				Expect(gjson.Get(out, gjsonStr).String()).To(Equal("true"))
```
rm3l, line 198: Besides that, how about also making sure the current namespace/project is not listed in the non-active ones?

```go
Expect(gjson.Get(out, "namespaces.#[active==false]#.name").String()).ShouldNot(ContainSubstring(commonVar.Project))
```

rm3l (commented):
Thanks for the changes. LGTM feature-wise. Just added one last comment in the tests.

/approve

tests/integration/cmd_namespace_test.go
feloy: IMO, if you check that there is a `{ "name": "ns", "active": true }` (the first check), you cannot have a `{ "name": "ns", "active": false }` (the second check)

rm3l: I agree. I actually wanted us to make sure that any other projects/namespaces in the output do have the `active` field set to `false`.

dharmit: Depending upon when the tests are running, the output could contain multiple projects that are marked `active: false`. So while it might make sense, I don't think it will be accurate, unless we check every namespace marked as `active: false`.

feloy: ok, I misunderstood the query, that would be interesting to test, effectively

dharmit: When I replied, I could see only the top comment. GH doing tricks. 
OK, I have added a check for that. On CI, we don't expect to be in a scenario where there won't be any other namespaces, so this wouldn't be a problem.

rm3l (approved):

feloy:
Tests pass on IBM Cloud

PR #6341, opened 2022-11-23
Disable flaky test on Windows

feloy:
**What type of PR is this:**

/kind tests

**What does this PR do / why we need it:**

This test fails on Windows because of the terminal emulator behaviour.

**Which issue(s) this PR fixes:**

Fixes #6327

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

**How to test changes / Special notes to the reviewer:**

PR #6343, opened 2022-11-23
Use go-cmp instead of reflect.DeepEqual to compare unit test results

### LINKED ISSUE #5932: Use `go-cmp` instead of `reflect.DeepEqual` to compare unit test results
/kind user-story
/area refactoring
/area testing

## User Story
As an **odo developer** writing unit tests or inspecting non-passing tests, I want to be able to **compare test results**, so that I can **iterate and fix tests faster**.

This was discussed during one of our Cabal meetings.

We currently have a lot of places in the unit tests where we write:
```go
if !reflect.DeepEqual(tt.want, got) {
  t.Errorf("expected: %v, got %v", tt.want, got)
}
```

When there is a mismatch causing a test failure, it is quite difficult to figure out what the diff is by reading the output, e.g:
```
--- FAIL: TestGetInputEnvVarsFromStrings/Test_case_1:_with_valid_two_key_value_pairs (0.00s)
utils_test.go:99: expected: [{key value nil} {key1 value12 nil}], got [{key value nil} {key1 value1 nil}]
```

In some other places, we rely on `reflect.DeepEqual` to compare objects and on `pretty.Compare` to display the diff:
```go
if !reflect.DeepEqual(got, tt.want) {
  t.Errorf("Got: %+v \n\n Want: %+v", got, tt.want)
  t.Errorf("Comparison: %v", pretty.Compare(got, tt.want))
  t.Logf("Error message is: %v", err)
}
```

[go-cmp](https://github.com/google/go-cmp) allows to write the condition in a simpler (and apparently more common) way and returns a detailed diff of what does not match, making it quicker to investigate. For example, the condition below would generate the output that follows:
```go
if diff := cmp.Diff(tt.want, got); diff != "" {
  t.Error(diff)
}
```

```
--- FAIL: TestGetInputEnvVarsFromStrings/Test_case_1:_with_valid_two_key_value_pairs (0.00s)
utils_test.go:96:   []v1.EnvVar{
                {Name: "key", Value: "value"},
                {
                        Name:      "key1",
            -           Value:     "value12",
            +           Value:     "value1",
                        ValueFrom: nil,
                },
              }
```

## Acceptance Criteria 
- [ ] Replace occurrences of all our `reflect.DeepEqual` and `pretty.Compare` usage with `cmp.Diff`. Replacement should be done only in test sources, as `go-cmp` is recommended in tests only.
- [ ] Replacement should be backwards-compatible, i.e., no test should be broken or skipped. Bear in mind that `go-cmp` allows [flexible options](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts) (or even defining custom [Comparers](https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer)) such as:
  - the ability to change its behavior w.r.t. unexported fields: [`IgnoreUnexported(typs ...interface{})`](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreUnexported)
  - the ability to compare floating-point numbers with an acceptable margin: [`EquateApprox(fraction, margin float64)`](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#EquateApprox)
  - the ability to sort maps prior to comparing them: [`SortMaps(lessFunc interface{})`](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortMaps)
- [ ] If not used anymore, remove the dependency to the `github.com/kylelemons/godebug` package (which `pretty.Compare` comes from)
- [ ] Document this in the [Coding Conventions wiki page](https://github.com/redhat-developer/odo/wiki/Dev:-Coding-Conventions)

## Links 
- Related Epic (mandatory): 
- go-cmp: https://github.com/google/go-cmp

/kind user-story

rm3l:
**What type of PR is this:**
/kind code-refactoring
/area testing
/area refactoring

This PR makes use of [github.com/google/go-cmp](https://github.com/google/go-cmp) everywhere in the unit tests.

I remembered this while debugging some tests, and thought it would make it easier to analyze unit tests.

More details about this in #5932 

Fixes #5932

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

`make test` should pass.


rm3l:
Rebased onto `main` and replaced a few other places where `reflect.DeepEqual` was being used in the unit tests.

rm3l:
Rebased to get #6341

valaparthvi (approved):

feloy:
The fact that unexported fields are not compared by default seems very error prone to me. 

We need to be sure that we check for every tests that we add the necessary structures containing unexported fields; but we won't get an error if we don't. I generally prefer the other way: that tests fail until we have defined all the necessary stuff.

rm3l:
> The fact that unexported fields are not compared by default seems very error prone to me.
> 
> We need to be sure that we check for every tests that we add the necessary structures containing unexported fields; but we won't get an error if we don't. I generally prefer the other way: that tests fail until we have defined all the necessary stuff.

Actually, the test would panic if `cmp.Diff` encounters unexported fields, requiring us to be more explicit about the behavior.
For example, if we remove `cmp.AllowUnexported(devfileCtx.DevfileCtx{})` from `TestBindingClient_RemoveBinding`, we would get the error below:

```
=== RUN   TestBindingClient_RemoveBinding/removed_the_binding_successfully_when_bound_as_files
    --- FAIL: TestBindingClient_RemoveBinding/removed_the_binding_successfully_when_bound_as_files (0.00s)
panic: cannot handle unexported field at {parser.DevfileObj}.Ctx.apiVersion:
	"github.com/devfile/library/pkg/devfile/parser/context".DevfileCtx
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported [recovered]
	panic: cannot handle unexported field at {parser.DevfileObj}.Ctx.apiVersion:
	"github.com/devfile/library/pkg/devfile/parser/context".DevfileCtx
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported

```

So yes, I agree with your point (tests failing until we have defined all the necessary stuff) and I think this can be helpful here. I found it even easier to achieve with the various options we can pass to `cmp.Diff`.

feloy:
> > The fact that unexported fields are not compared by default seems very error prone to me.
> > We need to be sure that we check for every tests that we add the necessary structures containing unexported fields; but we won't get an error if we don't. I generally prefer the other way: that tests fail until we have defined all the necessary stuff.
> 
> Actually, the test would panic if `cmp.Diff` encounters unexported fields, requiring us to be more explicit about the behavior. For example, if we remove `cmp.AllowUnexported(devfileCtx.DevfileCtx{})` from `TestBindingClient_RemoveBinding`, we would get the error below:
> 
> ```
> === RUN   TestBindingClient_RemoveBinding/removed_the_binding_successfully_when_bound_as_files
>     --- FAIL: TestBindingClient_RemoveBinding/removed_the_binding_successfully_when_bound_as_files (0.00s)
> panic: cannot handle unexported field at {parser.DevfileObj}.Ctx.apiVersion:
> 	"github.com/devfile/library/pkg/devfile/parser/context".DevfileCtx
> consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported [recovered]
> 	panic: cannot handle unexported field at {parser.DevfileObj}.Ctx.apiVersion:
> 	"github.com/devfile/library/pkg/devfile/parser/context".DevfileCtx
> consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported
> ```
> 
> So yes, I agree with your point (tests failing until we have defined all the necessary stuff) and I think this can be helpful here. I found it even easier to achieve with the various options we can pass to `cmp.Diff`.

OK, knowing this, I'm completely reassured. Thanks

rm3l:
```
Summarizing 1 Failure:
  [FAIL] odo add binding interactive command tests when running a deployment when binding to a service in a different namespace [It] should error out if service is not found in the namespace selected
  /go/src/github.com/redhat-developer/odo/tests/helper/helper_interactive.go:111
Ran 333 of 427 Specs in 2901.749 seconds
FAIL! -- 332 Passed | 1 Failed | 0 Pending | 94 Skipped
```

Overriding Prow tests because all tests are passing on IBM Cloud.

PR #6344, opened 2022-11-23
Fix potential unit tests failures caused by envconfig relying on the system environment

rm3l:
**What type of PR is this:**
/kind bug
/area testing

I just ran into the issue below, where our unit tests might fail depending on the system environment variables:

```
$ GLOBALODOCONFIG=/dev/null PODMAN_CMD=docker go test -race github.com/redhat-developer/odo/pkg/config/
--- FAIL: TestDefaultValues (0.00s)
    config_test.go:30: default value for "PodmanCmd" should be "podman" but is "docker"
    config_test.go:44: value for non specified env var "Globalodoconfig" should be nil but is "/dev/null"
    config_test.go:44: value for non specified env var "Globalodoconfig" should be nil but is "/dev/null"
FAIL
FAIL    github.com/redhat-developer/odo/pkg/config      0.008s
FAIL
```

I think unit tests should not depend on the system environment, as this can cause troubles.
`envconfig` supports this use case by providing a `Lookuper`: https://github.com/sethvargo/go-envconfig#testing

&mdash;

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,8 +20,12 @@ type Configuration struct {
 }
 
 func GetConfiguration() (*Configuration, error) {
+	return GetConfigurationWith(envconfig.OsLookuper())
+}
+
+func GetConfigurationWith(lookuper envconfig.Lookuper) (*Configuration, error) {
 	var s Configuration
-	err := envconfig.Process(context.Background(), &s)
+	err := envconfig.ProcessWith(context.Background(), &s, lookuper)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


rm3l:
```
...

  Expected
      <string>: 
  to equal
      <string>: true
  In [It] at: /go/src/github.com/redhat-developer/odo/tests/integration/cmd_namespace_test.go:198

...

Summarizing 1 Failure:
  [FAIL] odo create/delete/list/set namespace/project tests list namespace [It] should successfully list all the namespaces in JSON format
  /go/src/github.com/redhat-developer/odo/tests/integration/cmd_namespace_test.go:198
Ran 333 of 427 Specs in 2804.466 seconds
FAIL! -- 332 Passed | 1 Failed | 0 Pending | 94 Skipped
```

Overriding this because all tests are passing on IBM Cloud.

PR #6355, opened 2022-11-28
Fix flaky behavior with "list namespaces/projects -o json" integration tests

rm3l:
**What type of PR is this:**
/kind bug
/area testing

I noticed repeating issues with this test on a few PRs.

For some reason, `odo list namespaces/projects` would contain the current test namespace, but the subsequent `odo list namespaces/projects -o json` would no longer contain this namespace.
After taking a look at [Gomega doc](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode), this PR changes the test so that it only relies on what is actually needed for the test and uses expectations on the JSON output only.

&mdash;
**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/tests/integration/cmd_namespace_test.go
+++ b/tests/integration/cmd_namespace_test.go
@@ -3,7 +3,6 @@ package integration
 import (
 	"fmt"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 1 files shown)


feloy:
Thanks for this PR.
This function signature for Eventually is interesting

PR #6377, opened 2022-12-05
Get next free port when forwarding ports on podman

### LINKED ISSUE #6333: [podman] Expose the ports using the hostPort flag in the Ports of the container
As an odo user, I want `odo dev` to be resilient when a port is not available, so that I can have different components running on Podman.

Expose the ports (in a port range different from 40001-), using the hostPort flag in the Ports of the container

NOTES:
- Support `--random-ports` as well
  - with the cluster mode, a port is randomly selected at the kernel level; we need to check if something similar is possible with Podman

feloy:
**What type of PR is this:**

/kind feature

Fixes #6333

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/dev/podmandev/pod_test.go
+++ b/pkg/dev/podmandev/pod_test.go
@@ -331,7 +331,7 @@ func Test_createPodFromComponent(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, gotFwPorts, err := createPodFromComponent(tt.args.devfileObj(), tt.args.componentName, tt.args.appName, tt.args.buildCommand, tt.args.runCommand, tt.args.debugCommand, nil)
+			got, gotFwPorts, err := createPodFromComponent(tt.args.devfileObj(), tt.args.componentName, tt.args.appName, tt.args.buildCommand, tt.args.runCommand, tt.args.debugCommand, []int{40001, 40002})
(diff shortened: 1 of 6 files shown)


pkg/dev/podmandev/pod_test.go
```diff
@@ -221,15 +221,15 @@ func Test_createPodFromComponent(t *testing.T) {
 					Name:          "http",
 					ContainerPort: 8080,
 					Protocol:      "TCP",
-					HostPort:      39001,
+					HostPort:      40001,
```
rm3l, line 224: Wondering if checking the values this way might not be error-prone if the port is already taken locally. I had an unrelated local component running in Dev mode (and listening on port 40001), and this test was no longer passing because a different port was picked by `createPodFromComponent`.

Maybe we could make sure to explicitly pass a non-nil slice of ports as `usedPorts`  when calling `createPodFromComponent`, so that a port can be picked from this list.

Or, better to me, we should ideally check that the returned port is within the expected range and that the values for `LocalPort` (in `api.ForwardedPort`) and the `HostPort` (in the Pod Container Spec) are the same.

feloy: I would prefer the solution to pass ports as usedPorts, so we don't have to rely on the system

rm3l (approved):

Issue #6379, opened 2022-12-06
odo describe component does not work when logged out of cluster


dgolovin:
/kind bug

## What versions of software are you using?

**Operating System: linux**

**Output of `odo version`**

```
~/p/vscode-openshift-tools/out/tools/linux/odo version
odo v3.3.0 (1d315866b)

Server: https://api.sandbox-m2.ll9k.p1.openshiftapps.com:6443
Kubernetes: v1.24.6+5157800
```

## How did you run odo exactly?

```
 ~/p/vscode-openshift-tools/out/tools/linux/oc whoami
Error from server (Forbidden): users.user.openshift.io "~" is forbidden: User "system:anonymous" cannot get resource "users" in API group "user.openshift.io" at the cluster scope
```

```odo describe component```

## Actual behavior

```
 ~/p/vscode-openshift-tools/out/tools/linux/odo describe component
 ✗  failed to get ingresses/routes: ingresses.networking.k8s.io is forbidden: User "system:anonymous" cannot list resource "ingresses" in API group "networking.k8s.io" in the namespace "nivologd-dev"
```

## Expected behavior

odo prints out human readable output

## Any logs, error output, etc?

See actual behavior section.

**Comment by sbouchet** [comment]
@dgolovin : 
cannot reproduce on RHEL. 
```
sbouchet@sbouchet:Empty$ /home/sbouchet/.odo/cache/3.3.0/odo version
odo v3.3.0 (1d315866b)
sbouchet@sbouchet:Empty$ /home/sbouchet/.odo/cache/3.3.0/odo describe component
 ⚠  No connection to cluster defined
Name: node
Display Name: Node.js Runtime
Project Type: Node.js
Language: JavaScript
Version: 2.1.1
Description: Stack with Node.js 16
Tags: Node.js, Express, ubi8

Running in: None

Supported odo features:
 •  Dev: true
 •  Deploy: false
 •  Debug: true

Container components:
 •  runtime
```

The output of odo version still shows you're connected to a cluster.

dgolovin:
@sbouchet Thank you for verifying that. Do you actually have current-context set in ~/.kube/config? I have the same issue on MacOS. If I have current-context pointed to crc instance that is not running at the moment, I have the same issue. 

In my case that cleans Components View, because every 'odo describe component' command fails and extension cannot detect components inside workspace. 

We can - of course - workaround this and just look for devfile.yaml/.devfile.yaml to detect odo components in workspace.

[...]

PR #6395, opened 2022-12-08
Test with OpenShift cluster non authenticated

feloy:
**What type of PR is this:**

/kind tests

Helps test #6379

Run the same tests with NoCluster and with Unauth label.

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/.ibm/pipelines/openshift-unauth-tests.sh
+++ b/.ibm/pipelines/openshift-unauth-tests.sh
@@ -11,7 +11,7 @@ ibmcloud oc cluster config -c "${CLUSTER_ID}"
 (
     set -e
     make install
-    make test-integration-cluster-unauth
+    make test-integration-openshift-unauth
 ) |& tee "/tmp/${LOGFILE}"
 
 RESULT=${PIPESTATUS[0]}
--- a/.ibm/pipelines/openshift-unauth-tests.sh
+++ b/.ibm/pipelines/openshift-unauth-tests.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+LOGFILE="pr-${GIT_PR_NUMBER}-openshift-tests-${BUILD_NUMBER}"
+
+source .ibm/pipelines/functions.sh
+
+ibmcloud login --apikey "${API_KEY_QE}"
+ibmcloud target -r eu-de
+ibmcloud oc cluster config -c "${CLUSTER_ID}"
+
+cleanup_namespaces
+
+(
+    set -e
+    make install
+    make test-integration-cluster-unauth
+) |& tee "/tmp/${LOGFILE}"
+
+RESULT=${PIPESTATUS[0]}
+
+save_logs "${LOGFILE}" "OpenShift Unauthenticated Tests" ${RESULT}
+
+exit ${RESULT}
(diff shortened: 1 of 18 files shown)


tests/helper/helper_generic.go
```diff
@@ -197,7 +197,9 @@ func CommonBeforeEach() CommonVar {
 	commonVar.OriginalKubeconfig = os.Getenv("KUBECONFIG")
 	if NeedsCluster(CurrentSpecReport().Labels()) {
 		LocalKubeconfigSet(commonVar.ConfigDir)
-		commonVar.Project = commonVar.CliRunner.CreateAndSetRandNamespaceProject()
+		if IsAuth(CurrentSpecReport().Labels()) {
```
rm3l, line 200: Would it be possible to check the current login status (using for example `oc whoami`), so as to make the test fail right away if the test is labeled as `unauth` but the current user is already logged in?
This way, the test could just fail sooner because the preconditions are not met.
I was testing this, and noticed that `make test-integration-cluster-unauth` was taking a lot of time to finish because I was already logged into the cluster.
(But that can be handled in a separate issue/PR if you prefer (and if you think that makes sense :-)).

Makefile
```diff
@@ -192,7 +192,11 @@ openshiftci-presubmit-unittests:
 
 .PHONY: test-integration-cluster
 test-integration-cluster:
-	$(RUN_GINKGO) $(GINKGO_FLAGS) --junit-report="test-integration.xml" --label-filter="!nocluster && !podman" tests/integration
+	$(RUN_GINKGO) $(GINKGO_FLAGS) --junit-report="test-integration.xml" --label-filter="!unauth && !nocluster && !podman" tests/integration
+
+.PHONY: test-integration-cluster-unauth
+test-integration-cluster-unauth:
```
rm3l, line 198: Can we rename this target to include `openshit` in the name? This will make it clear that it is specific to OpenShift (and also why it is not run as part of the `test-integration` target), I think.

.ibm/pipelines/openshift-unauth-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+
+source .ibm/pipelines/functions.sh
+
+ibmcloud login --apikey "${API_KEY_QE}"
+ibmcloud target -r eu-de
+ibmcloud oc cluster config -c "${CLUSTER_ID}"
+
+(
+    set -e
+    make install
```
rm3l, line 13: For consistency with the other scripts, do we want to also export the `DEVFILE_PROXY` env var here?

rm3l (commented):
LGTM overall, but left a few comments.

tests/helper/helper_generic.go
feloy: Good idea, added

.ibm/pipelines/openshift-unauth-tests.sh
feloy: The proxy running in the cluster, it is not accessible, I had to remove it or it contains an error message

rm3l (approved):
Thanks Philippe.

PR #6405, opened 2022-12-13
updating-test-suit-commands

HarshithaMS005:
**What type of PR is this:**
kind code-refactoring

Execute clustered test cases for s390x and power architecture 

**PR acceptance criteria:**
- [ ] Integration test

### DIFF
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -33,18 +33,29 @@ oc login -u developer -p password@123 --insecure-skip-tls-verify
 oc whoami
 
 if [ "${ARCH}" == "s390x" ]; then
-    # Integration tests
+    # Integration tests for no cluster
     make test-integration
+
+    # Integration tests for cluster
+    make test-integration-cluster
+
     # E2e tests
     make test-e2e
 elif [ "${ARCH}" == "ppc64le" ]; then
-    # Integration tests
+    # Integration tests for no cluster
     make test-integration
+
+    # Integration tests for cluster
+    make test-integration-cluster
+
     # E2e tests
     make test-e2e
 else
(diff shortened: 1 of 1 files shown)


scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -33,19 +33,30 @@ oc login -u developer -p password@123 --insecure-skip-tls-verify
 oc whoami
 
 if [ "${ARCH}" == "s390x" ]; then
-    # Integration tests
+    # Integration tests for no cluster
     make test-integration
```
feloy, line 37: Note that `make test-integration` covers both cluster and no-cluster related tests.

Do you want to use `make test-integration-no-cluster`instead? (here and in different other places)

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -33,19 +33,30 @@ oc login -u developer -p password@123 --insecure-skip-tls-verify
...
     # E2e tests
     make test-e2e
 else
-    # Integration tests
+    # Integration tests for no cluster
     make test-integration || error=true
 
+    # Integration tests for cluster
+    make test-integration-cluster || error=true
+
```
anandrkskd, line 59: Same as above [comment](https://github.com/redhat-developer/odo/pull/6405/files#r1047131601) 
This won't be required.

anandrkskd (changes requested):

scripts/openshiftci-presubmit-all-tests.sh
HarshithaMS005: No. In our CI no-cluster test cases are failing and test suit exiting without running cluster test cases. We will recheck on this. Thank you

PR #6415, opened 2022-12-15
Fix preference not taken into account for `odo version`

### LINKED ISSUE #6417: Timeout not taken into account during `odo version`
/kind bug

When running `odo version`, the Timeout preference is not considered, as it should be.

feloy:
**What type of PR is this:**

/kind bug

This PR adds the PREFERNECE dependency to `odo version`, and creates a CommandGroup abstraction to avoid this error occurs again.

`odo version` can use the Timeout value defined in the odo preferences, but this value were never taken into account, because the PREFERENCE dependency were overwritten when setting the command group.

Fixes #6417

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/cmd/odo/help_test.go
+++ b/cmd/odo/help_test.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/redhat-developer/odo/pkg/config"
+	envcontext "github.com/redhat-developer/odo/pkg/config/context"
+	"github.com/redhat-developer/odo/pkg/odo/cli"
+	"k8s.io/klog"
+)
+
+var (
+	intro = `Usage:
+  odo [flags]
+  odo [command]
+
+Examples:
+  Initializing your component by taking your pick from multiple languages or frameworks:
+  odo init
+  
+  After creating your initial component, start development with:
+  odo dev
+  
+  Want to deploy after development? See it live with:
+  odo deploy`
+
+	mainCommands = `Main Commands:
+  build-images Build images
+  deploy       Deploy components
+  dev          Deploy component to development cluster
+  init         Init bootstraps a new project
+  logs         Show logs of all containers of the component
+  registry     List all components from the Devfile registry
+
+`
+
+	managementCommands = `Management Commands:
+  add          Add resources to devfile (binding)
+  create       Perform create operation (namespace)
+  delete       Delete resources (component, namespace)
+  describe     Describe resource (binding, component)
+  list         List all components in the current namespace (binding, component, namespace, services)
+  remove       Remove resources from devfile (binding)
+  set          Perform set operation (namespace)
+
+`
+
+	openshiftCommands = `OpenShift Commands:
+  login        Login to cluster 
+  logout       Logout of the cluster`
+
+	utilityCommands = `Utility Commands:
+  analyze      Detect devfile to use based on files present in current directory
+  completion   Add odo completion support to your development environment
+  preference   Modifies preference settings (add, remove, set, unset, view)
+  version      Print the client version information
+
+`
+)
+
+func TestOdoHelp(t *testing.T) {
+	ctx := context.Background()
+	envConfig, err := config.GetConfiguration()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx = envcontext.WithEnvConfig(ctx, *envConfig)
+	klog.InitFlags(nil)
+
+	root := cli.NewCmdOdo(ctx, cli.OdoRecommendedName, cli.OdoRecommendedName)
+
+	var stdoutB, stderrB bytes.Buffer
+	root.SetOut(&stdoutB)
+	root.SetErr(&stderrB)
+
+	root.SetArgs([]string{"help"})
+
+	err = root.ExecuteContext(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	stdout := stdoutB.String()
+	stderr := stderrB.String()
+
+	if stderr != "" {
+		t.Fatal("stderr should be empty")
+	}
+
+	if !strings.Contains(stdout, intro) {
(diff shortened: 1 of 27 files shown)


rm3l (commented):
Good catch - thanks!

I'm wondering in this case if we shouldn't make the preference client a mandatory dependency, and explicitly fail if this is not the case.
I can see that `preferenceInfo#GetTimeout()` already falls back to `DefaultTimeout` if there is no timeout setting defined. So we could be more strict by expecting the Preference client dependency to be there in `VersionOptions#Complete`.
To be clear, I am just thinking of the following diff:
```diff
diff --git a/pkg/odo/cli/version/version.go b/pkg/odo/cli/version/version.go
index 180c8b58a..e31226808 100644
--- a/pkg/odo/cli/version/version.go
+++ b/pkg/odo/cli/version/version.go
@@ -5,13 +5,11 @@ import (
        "fmt"
        "os"
        "strings"
-       "time"
 
        "github.com/redhat-developer/odo/pkg/kclient"
        "github.com/redhat-developer/odo/pkg/odo/cmdline"
        "github.com/redhat-developer/odo/pkg/odo/genericclioptions"
        "github.com/redhat-developer/odo/pkg/odo/genericclioptions/clientset"
-       "github.com/redhat-developer/odo/pkg/preference"
        odoversion "github.com/redhat-developer/odo/pkg/version"
 
        "github.com/spf13/cobra"
@@ -63,16 +61,7 @@ func (o *VersionOptions) Complete(ctx context.Context, cmdline cmdline.Cmdline,
                client, err := kclient.New()
 
                if err == nil {
-                       // checking the value of timeout in preference
-                       var timeout time.Duration
-                       if o.clientset.PreferenceClient != nil {
-                               timeout = o.clientset.PreferenceClient.GetTimeout()
-                       } else {
-                               // the default timeout will be used
-                               // when the value is not readable from preference
-                               timeout = preference.DefaultTimeout
-                       }
-                       o.serverInfo, err = client.GetServerVersion(timeout)
+                       o.serverInfo, err = client.GetServerVersion(o.clientset.PreferenceClient.GetTimeout())
                        if err != nil {
                                klog.V(4).Info("unable to fetch the server version: ", err)
                        }
```

At least, this would explicitly panic if the Preference Client is not there and allow to catch this issue if it occurs again, IMO.

What do you think?

feloy:
> Good catch - thanks!
> 
> I'm wondering in this case if we shouldn't make the preference client a mandatory dependency, and explicitly fail if this is not the case. I can see that `preferenceInfo#GetTimeout()` already falls back to `DefaultTimeout` if there is no timeout setting defined. So we could be more strict by expecting the Preference client dependency to be there in `VersionOptions#Complete`. To be clear, I am just thinking of the following diff:
> 
> ```diff
> diff --git a/pkg/odo/cli/version/version.go b/pkg/odo/cli/version/version.go
> index 180c8b58a..e31226808 100644
> --- a/pkg/odo/cli/version/version.go
> +++ b/pkg/odo/cli/version/version.go
> @@ -5,13 +5,11 @@ import (
>         "fmt"
>         "os"
>         "strings"
> -       "time"
>  
>         "github.com/redhat-developer/odo/pkg/kclient"
>         "github.com/redhat-developer/odo/pkg/odo/cmdline"
>         "github.com/redhat-developer/odo/pkg/odo/genericclioptions"
>         "github.com/redhat-developer/odo/pkg/odo/genericclioptions/clientset"
> -       "github.com/redhat-developer/odo/pkg/preference"
>         odoversion "github.com/redhat-developer/odo/pkg/version"
>  
>         "github.com/spf13/cobra"
> @@ -63,16 +61,7 @@ func (o *VersionOptions) Complete(ctx context.Context, cmdline cmdline.Cmdline,
>                 client, err := kclient.New()
>  
>                 if err == nil {
> -                       // checking the value of timeout in preference
> -                       var timeout time.Duration
> -                       if o.clientset.PreferenceClient != nil {
> -                               timeout = o.clientset.PreferenceClient.GetTimeout()
> -                       } else {
> -                               // the default timeout will be used
> -                               // when the value is not readable from preference
> -                               timeout = preference.DefaultTimeout
> -                       }
> -                       o.serverInfo, err = client.GetServerVersion(timeout)
> +                       o.serverInfo, err = client.GetServerVersion(o.clientset.PreferenceClient.GetTimeout())
>                         if err != nil {
>                                 klog.V(4).Info("unable to fetch the server version: ", err)
>                         }
> ```
> 
> At least, this would explicitly panic if the Preference Client is not there and allow to catch this issue if it occurs again, IMO.
> 
> What do you think?

You are right, I'm also annoyed by this test on the client, which was hiding the error.

I'll apply your patch.
Thanks

cmd/odo/help_test.go
```diff
@@ -0,0 +1,108 @@
...
+	utilityCommands = `Utility Commands:
+  analyze      Detect devfile to use based on files present in current directory
+  completion   Add odo completion support to your development environment
+  preference   Modifies preference settings (add, remove, set, unset, view)
+  version      Print the client version information
+
+`
+)
+
+func TestOdoHelp(t *testing.T) {
```
rm3l, line 64: Thanks for this 😍 💯

rm3l (approved):

feloy:
Unrelated flaky test
```
[FAILED] [238.740 seconds]
odo dev command tests
/go/odo_1/tests/integration/cmd_dev_test.go:31
  /go/odo_1/tests/integration/cmd_dev_test.go:595
    when devfile has single endpoint
    /go/odo_1/tests/integration/cmd_dev_test.go:596
      when running odo dev
      /go/odo_1/tests/integration/cmd_dev_test.go:605
        when modifying memoryLimit for container in Devfile [BeforeEach]
        /go/odo_1/tests/integration/cmd_dev_test.go:639
          should expose the endpoint on localhost
          /go/odo_1/tests/integration/cmd_dev_test.go:655
```

Issue #6416, opened 2022-12-15
Add more tests for odo dev on podman


feloy:
/kind feature

## Which functionality do you think we should add?

The tests are currently getting the created Deployment/Pod on cluster, we need to write some abstraction to be able to get the Deployment/Pod on podman or cluster, depending on the platform the test is running on.

PR #6418, opened 2022-12-15
odo delete component deletes components running on podman

### LINKED ISSUE #6296: `odo delete component` should be able to delete components running on podman
Acceptance Criteria:
- [ ] Default behavior: delete everything from all accessible platforms
- [ ] If `platform` is specified, limit deletion to that platform
- [ ] Do not fail if K8s cluster or podman are not available

NOTES:
- Depends on https://github.com/redhat-developer/odo/issues/6317

feloy:
**What type of PR is this:**

/kind feature

`odo delete component` support for podman when devfile is present

(support for `--name` will be done in a separate PR)

Fixes partially #6296

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/component/delete/delete.go
+++ b/pkg/component/delete/delete.go
@@ -20,6 +20,7 @@ import (
 	odolabels "github.com/redhat-developer/odo/pkg/labels"
 	"github.com/redhat-developer/odo/pkg/libdevfile"
 	"github.com/redhat-developer/odo/pkg/log"
+	clierrors "github.com/redhat-developer/odo/pkg/odo/cli/errors"
 	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
 	"github.com/redhat-developer/odo/pkg/util"
 )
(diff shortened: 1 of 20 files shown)


feloy:
Blocked by #6416 
This PR needs more abstraction for component to add integration tests

pkg/podman/podman.go
```diff
@@ -89,6 +89,33 @@ func (o *PodmanCli) PlayKube(pod *corev1.Pod) error {
...
+	serializer := jsonserializer.NewSerializerWithOptions(
+		jsonserializer.SimpleMetaFactory{},
+		scheme.Scheme,
+		scheme.Scheme,
+		jsonserializer.SerializerOptions{
+			Yaml: true,
+		},
+	)
+
+	cmd := exec.Command(o.podmanCmd, "kube", "generate", name)
```
valaparthvi, line 102: I am running into the following error while running `ODO_EXPERIMENTAL_MODE=t odo delete component`, it seems it has something to do with this command:
```sh
odo delete component                                                                                                         

Searching resources to delete, please wait...
 ⚠  failed to get deployment "my-go-app-app": Get "https://192.168.49.2:8443/apis/apps/v1/namespaces/default/deployments/my-go-app-app": dial tcp 192.168.49.2:8443: i/o timeout
 ✗  exit status 125: Error: unrecognized command `podman kube`
Try 'podman --help' for more information.

```

But when I tried to use `podman play kube ...`, I ran into another error: " ✗  exit status 125: Error: accepts 1 arg(s), received 2"

feloy: Which version of podman are you using?

valaparthvi:
```sh
$ podman version
Version:      3.4.7
API Version:  3.4.7
Go Version:   go1.16.15
Built:        Mon Jul 11 04:05:25 2022
OS/Arch:      linux/amd64
```

valaparthvi: Is there a minimum version of podman that we need?

feloy: The devs are done based on podman v4 (v4.3.0 exactly)

valaparthvi: Okay. There's one more error (might be out of context for this PR) that I am facing while using `odo dev`.

I am not connected to any cluster atm, I usually use minikube, which is in Stopped state.
When I run `ODO_EXPERIMENTAL_MODE=t odo dev --run-on=podman`, it panics because of https://github.com/redhat-developer/odo/blob/0c008c6be9666bcd61d6d3874fb6b60fe5b6f917/pkg/component/component.go#L94 because `GetRunningPodFromSelector` has not been implemented yet for podmanClient.

Are you able to reproduce this as well?

feloy: Yes, this is something that needs to be done. I just created an issue to follow it: https://github.com/redhat-developer/odo/issues/6434

pkg/podman/podman.go
```diff
@@ -144,6 +184,30 @@ func (o *PodmanCli) VolumeLs() (map[string]bool, error) {
 	return SplitLinesAsSet(string(out)), nil
 }
 
+func (o *PodmanCli) CleanupPodResources(pod *corev1.Pod) error {
```
valaparthvi, line 187: This fails if `odo dev` was stopped because of an error.
```sh
$ ODO_EXPERIMENTAL_MODE=true odo delete component   

 •  The following pods and associated volumes will be deleted from podman:
        - my-nodejs-app-app
? Are you sure you want to delete "my-nodejs-app" and all its resources? Yes
 ⚠  Failed to delete the pod "my-nodejs-app-app" from podman: exit status 125: Error: error freeing lock for container dfb4ce8ee86a29b5c8845a2dd8818d801990a40e2413ef6b7876b7c8240d939a: no such file or directory
```

To reproduce:
1. `mkcd /tmp/nodejs`
2. `odo init --devfile nodejs --name my-node-app`; skip downloading the starter project
3. `ODO_EXPERIMENTAL_MODE=t odo dev --run-on podman`; this will fail during the build because there will be no project files.
4. `ODO_EXPERIMENTAL_MODE=t odo delete componen`.

valaparthvi (changes requested):

pkg/component/delete/delete.go
```diff
@@ -107,6 +116,8 @@ func (do DeleteComponentClient) ListResourcesToDeleteFromDevfile(devfileObj pars
 
 		deployment, err = do.kubeClient.GetDeploymentByName(deploymentName)
 		if err != nil && !kerrors.IsNotFound(err) {
+			// Kubernetes cluster access fails, return with a warning only
+			err = clierrors.NewWarning(fmt.Sprintf("failed to get deployment %q", deploymentName), err)
```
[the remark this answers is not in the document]
feloy: 1. I would also like to, but I'm not sure we can find an exhaustive and definitive lit of such error(s).

2. As the function where it happens (ListClusterResourcesToDeleteFromDevfile) is a Devfile-related function, I would consider cluster information optional on the result, and never return an error when an error happens on the cluster, whatever the error.

pkg/component/delete/delete.go
```diff
@@ -197,3 +208,27 @@ func (do *DeleteComponentClient) ExecutePreStopEvents(devfileObj parser.DevfileO
...
+		return false, nil, fmt.Errorf("failed to get the resource %q name for component %q; cause: %w", kclient.DeploymentKind, componentName, err)
+	}
+
+	allPods, err := do.podmanClient.PodLs()
+	if err != nil {
+		err = clierrors.NewWarning("failed to get pods on podman", err)
+		return false, nil, err
+	}
+
+	if _, isInnerLoopDeployed = allPods[podName]; isInnerLoopDeployed {
```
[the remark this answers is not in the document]
feloy: I would prefer to keep the innerloop information, as we really want to return that the inner loop pod has been deployed (and not some pod from odo deploy), even if we only create pods dor inner loop.

This information is used by the caller to execute the prestop handler.

pkg/odo/cli/delete/component/component.go
```diff
@@ -127,27 +152,75 @@ func (o *ComponentOptions) deleteNamedComponent(ctx context.Context) error {
...
 }
 
+func messageWithPlatforms(cluster, podman bool, name, namespace string) string {
+	details := []string{}
+	if cluster {
+		details = append(details, fmt.Sprintf(" in namespace %q", namespace))
+	}
+	if podman {
+		details = append(details, " on podman")
+	}
```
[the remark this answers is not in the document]
feloy: I prefer the original solution, as it does not gives to the function the details on how the cluster/podman platforms are used (the fact that we are setting the clients to nil to describe that the platform is not used is an implementation detail, and could change).

pkg/podman/podman.go
valaparthvi: I was unable to reproduce this again, it is a glitch with podman, we can ignore it.

pkg/component/delete/delete.go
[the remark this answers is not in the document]
valaparthvi: >I would also like to, but I'm not sure we can find an exhaustive and definitive lit of such error(s).

What error is returned rn?

> As the function where it happens (ListClusterResourcesToDeleteFromDevfile) is a Devfile-related function, I would consider cluster information optional on the result, and never return an error when an error happens on the cluster, whatever the error.

Yeah, I guess you're right about that.

pkg/odo/cli/delete/component/component.go
```diff
@@ -85,6 +90,18 @@ func (o *ComponentOptions) Complete(ctx context.Context, cmdline cmdline.Cmdline
...
+	if !feature.IsEnabled(ctx, feature.GenericRunOnFlag) {
+		o.clientset.PodmanClient = nil
+	}
+	switch fcontext.GetRunOn(ctx, "") {
+	case commonflags.RunOnCluster:
+		o.clientset.PodmanClient = nil
+	case commonflags.RunOnPodman:
+		o.clientset.KubernetesClient = nil
+	}
+
```
valaparthvi, line 104: I think it should go on L79, because if `o.name == "" ` is satisfied, it will return early.

valaparthvi (changes requested):

pkg/component/delete/delete.go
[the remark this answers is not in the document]
feloy: > > I would also like to, but I'm not sure we can find an exhaustive and definitive lit of such error(s).
> 
> What error is returned rn?
> 

for the case I can think of, it is a "Forbidden" error

valaparthvi:
/lgtm

Thank you for working on this!

PR #6471, opened 2023-01-06
Ignore and show warning when handling Kubernetes components on Podman

### LINKED ISSUE #6299: [odo on Podman] Ignore and show warning when handling Kubernetes components
See https://github.com/redhat-developer/odo/issues/6299#issuecomment-1317187243

valaparthvi:

**What type of PR is this:**
/kind feature

Fixes #6299 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

1. `odo init --devfile go --starter go-starter --name my-go-app`
2. Use the following Devfile:
<details>

```yaml
commands:
- exec:
    commandLine: go build main.go
    component: runtime
    env:
    - name: GOPATH
      value: ${PROJECT_SOURCE}/.go
    - name: GOCACHE
      value: ${PROJECT_SOURCE}/.cache
    group:
      isDefault: true
      kind: build
    workingDir: ${PROJECT_SOURCE}
  id: build
- exec:
    commandLine: ./main
    component: runtime
    group:
      isDefault: true
      kind: run
    workingDir: ${PROJECT_SOURCE}
  id: run

components:
- container:
    args:
    - tail
    - -f
    - /dev/null
    endpoints:
    - name: http-go
      targetPort: 8080
    image: registry.access.redhat.com/ubi9/go-toolset:latest
    memoryLimit: 1024Mi
    mountSources: true
  name: runtime
# This will create a Deployment in order to run your container image across
# the cluster.
- name: outerloop-deployment
  kubernetes:
    inlined: |
      kind: Deployment
      apiVersion: apps/v1
      metadata:
        name: {{RESOURCE_NAME}}
      spec:
        replicas: 1
        selector:
          matchLabels:
            app: {{RESOURCE_NAME}}
        template:
          metadata:
            labels:
              app: {{RESOURCE_NAME}}
          spec:
            containers:
              - name: {{RESOURCE_NAME}}
                image: registry.access.redhat.com/ubi9/go-toolset:latest
                ports:
                  - name: http
                    containerPort: {{CONTAINER_PORT}}
                    protocol: TCP
                resources:
                  limits:
                    memory: "1024Mi"
                    cpu: "500m"
metadata:
  description: Go is an open source programming language that makes it easy to build
    simple, reliable, and efficient software.
  displayName: Go Runtime
  language: Go
  name: my-go-app
  projectType: Go
  provider: Red Hat
  tags:
  - Go
  version: 1.0.2
# Deploy "kind" ID's use schema 2.2.0+
schemaVersion: 2.2.0
starterProjects:
- description: A Go project with a simple HTTP server
  git:
    checkoutFrom:
      revision: main
    remotes:
  name: go-starter
# Add the following variables code anywhere in devfile.yaml
# This MUST be a container registry you are able to access
variables:
  # CONTAINER_IMAGE: quay.io/MYUSERNAME/go-odo-example
  RESOURCE_NAME: my-go-app
  CONTAINER_PORT: "8080"
  DOMAIN_NAME: go.example.com
```
</details>

3. `ODO_EXPERIMENTAL_MODE=true odo dev --platform=podman`


pkg/dev/podmandev/reconcile.go
```diff
@@ -118,6 +123,24 @@ func (o *DevClient) reconcile(
...
+	if err != nil {
+		return
+	}
+	for _, comp := range k8sComponents {
+		uList, err := libdevfile.GetK8sComponentAsUnstructuredList(devfileObj, comp.Name, "", devfilefs.DefaultFs{})
+		if err != nil {
+			continue
+		}
+		for _, u := range uList {
+			log.Warningf("Skipping Kubernetes component \"%v/%v\". Kubernetes components are not supported on Podman.", u.GetKind(), u.GetName())
```
rm3l, line 138: This message might be repeated if there are multiple resources.
I'd suggest we display a single warning with all the resource names as a comma-separated string (or a bullet list). What do you think?

rm3l: Also, I'm wondering if we need to go so far with the K8s component (via `libdevfile.GetK8sComponentAsUnstructuredList`). I think it is not needed as we skipping it anyways. Worse, the warning will not be displayed if `libdevfile.GetK8sComponentAsUnstructuredList` returns an error (if the YAML content for the Kubernetes component is invalid for example).
So I think we could just display the component name as defined in the Devfile (`comp.Name`)). What do you think?

rm3l (changes requested):

pkg/dev/podmandev/reconcile.go
```diff
@@ -118,6 +123,20 @@ func (o *DevClient) reconcile(
...
+func (o *DevClient) warnAboutK8sResources(devfileObj parser.DevfileObj) {
+	k8sComponents, err := devfile.GetKubernetesComponentsToPush(devfileObj, false)
+	if err != nil {
+		return
+	}
+	var compNames []string
+	for _, comp := range k8sComponents {
+		compNames = append(compNames, comp.Name)
+	}
+	log.Warningf("Skipping Kubernetes component %q. Kubernetes components are not supported on Podman.", strings.Join(compNames, ", "))
```
rm3l, line 136: This should only be displayed if `compNames` is not empty.

rm3l: Small suggestion: I'd also reword the warning message like so, to have the reason displayed first:
```
 ⚠  Kubernetes components are not supported on Podman. Skipping "$comp1", "$comp2".
```

What do you think?

pkg/dev/podmandev/reconcile.go
```diff
@@ -118,6 +123,20 @@ func (o *DevClient) reconcile(
 	return nil
 }
 
+// warnAboutK8sResources prints a warning if the Devfile contains a K8s resource that it needs to create on Podman.
+func (o *DevClient) warnAboutK8sResources(devfileObj parser.DevfileObj) {
+	k8sComponents, err := devfile.GetKubernetesComponentsToPush(devfileObj, false)
```
rm3l, line 128: One more thing I just thought about: if the run command is a composite referencing an `apply` command with a K8s component, I think no warning will be displayed for this component.
I think you also need to consider this case by also displaying a warning in the command handler.

rm3l (changes requested):

pkg/dev/podmandev/reconcile.go
```diff
@@ -118,6 +123,20 @@ func (o *DevClient) reconcile(
 	return nil
 }
 
+// warnAboutK8sResources prints a warning if the Devfile contains a K8s resource that it needs to create on Podman.
+func (o *DevClient) warnAboutK8sResources(devfileObj parser.DevfileObj) {
+	k8sComponents, _ := devfile.GetKubernetesComponentsToPush(devfileObj, true)
```
rm3l, line 128: As I understand it, this would include all Devfile components referenced by an `apply` command, regardless of whether `odo dev` uses them or not, right? So if my Devfile contains a `deploy` command with some Kubernetes resources, this will show warnings for such components, even if they will never be used by `odo dev`, no?

rm3l:
```
$ odo init --name java-springboot-starter --devfile java-springboot --devfile-registry DefaultDevfileRegistry --devfile-version 2.0.0 --starter springbootproject                                                   

$ ODO_EXPERIMENTAL_MODE=t odo dev --platform=podman

  __
 /  \__     Developing using the "java-springboot-starter" Devfile
 \__/  \    Platform: podman
 /  \__/    odo version: v3.5.0
 \__/

↪ Deploying to podman in developer mode
 ✓  Deploying pod [4s]
 ⚠  Kubernetes components are not supported on Podman. Skipping: deploy.
 ✓  Building your application in container (command: build) [46s]
 -  Forwarding from 127.0.0.1:40001 -> 8080
 -  Forwarding from 127.0.0.1:40002 -> 5858                                                         
```

I can confirm the behavior, and find that the warning here can be confusing, as, IMO, it makes users think that `odo dev` was about to run the "deploy" component.

valaparthvi: We also don't support building images for podman as of now, I add adding a warning for those components as well.

pkg/devfile/components.go
```diff
@@ -50,3 +50,56 @@ func GetKubernetesComponentsToPush(devfileObj parser.DevfileObj, allowApply bool
...
+			CommandOptions: parsercommon.CommandOptions{CommandType: devfilev1.ApplyCommandType},
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		if len(cmd) == 0 {
+			continue
+		}
+		// we assume there will be only one command that matches the subCommand ID
+		if applyCmd := cmd[0]; applyCmd.Apply != nil {
```
rm3l, line 84: This assumes that the `apply` component will be a direct child of the composite command. But we might have several levels, I think, e.g., a composite command containing another composite command in turn containing another composite command referencing an `apply` Kubernetes component. And in such cases, no warning is displayed.

What do you think about displaying the warning messages in the [Podman command handler](https://github.com/redhat-developer/odo/blob/main/pkg/dev/podmandev/command.go) instead, while the handler is walking through each component when running the command? You would still need to handle the case of K8s components not referenced by any command.

valaparthvi: >What do you think about displaying the warning messages in the [Podman command handler](https://github.com/redhat-developer/odo/blob/main/pkg/dev/podmandev/command.go) instead, while the handler is walking through each component when running the command? You would still need to handle the case of K8s components not referenced by any command.

:facepalm:
I can't believe I didn't think of that.

pkg/dev/podmandev/reconcile.go
```diff
@@ -118,6 +123,23 @@ func (o *DevClient) reconcile(
 	return nil
 }
 
+// warnAboutApplyComponents prints a warning if the Devfile contains a K8s resource that it needs to create or build an image referenced by an apply command on Podman for a given command name and groupKind.
```
rm3l, line 126: Suggestion to update the doc of this function.

```suggestion
// warnAboutApplyComponents prints a warning if the Devfile contains standalone K8s components (not referenced by any Apply commands). These resources are currently applied when running in the cluster mode, but not on Podman.
```

rm3l (commented):
Thanks for the changes. Just a small suggestion to update a function doc.

rm3l (approved):

PR #6473, opened 2023-01-06
Do not output log to stdout

feloy:
**What type of PR is this:**

/kind bug

When running `odo dev -v2`, output of commands are displayed on stdout, and makes some tests fail because of these unexpected data in output. 

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -79,11 +79,13 @@ func startReaderGoroutine(reader io.Reader, show bool, cmdOutput *[]string, cons
 		for scanner.Scan() {
 			line := scanner.Text()
 
-			if show || log.IsDebug() {
+			if show {
 				_, err := fmt.Fprintln(os.Stdout, line)
 				if err != nil {
 					log.Errorf("Unable to print to stdout: %s", err.Error())
 				}
+			} else {
+				klog.V(2).Infof(line)
 			}
 
 			if cmdOutput != nil {


valaparthvi:
IBM passes.

valaparthvi:
I couldn't reproduce the error, and I couldn't try to fake the error either, do you know if there is a way to test that this will work as expected?

feloy:
The lines:
```
      + 	cat: /opt/odo/.odo_cmd_run.pid: No such file or directory
        	✓  Building your application in container on cluster (command: build) [1s]
      + 	196
      + 	0
      + 	196
      + 	0
```

are displayed when you run odo dev with `-v2`.

With this patch, they are displayed as log, and not on stdout directly.

valaparthvi:
> The lines:
> 
> ```
>       + 	cat: /opt/odo/.odo_cmd_run.pid: No such file or directory
>         	✓  Building your application in container on cluster (command: build) [1s]
>         	•  Executing the application (command: run)  ...
>       + 	196
>       + 	0
>       + 	196
>       + 	0
> ```
> 
> are displayed when you run odo dev with `-v2`.
> 
> With this patch, they are displayed as log, and not on stdout directly.

I understand that, but I wasn't able to reproduce the error when I ran doc tests. So I tried to fake my way into getting an output like that, but couldn't. I'll try running the tests again.

But, if you're able to reproduce the error and this solution works for you, I'll take your word for it.

/lgtm

PR #6477, opened 2023-01-09
Display a warning that "odo dev" on Podman needs to be restarted if the Devfile is changed

### LINKED ISSUE #6336: Warn users that odo dev needs to be restarted when related information is changed in Devfile
Inform user for the moment with a warning that odo dev needs to be restarted

rm3l:
**What type of PR is this:**
/area odo-on-podman

Fixes #6336 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/dev/podmandev/podmandev.go
+++ b/pkg/dev/podmandev/podmandev.go
@@ -1,15 +1,22 @@
 package podmandev
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"path/filepath"
 	"strings"
 
+	"github.com/devfile/library/pkg/devfile/parser"
+	"k8s.io/klog"
+
 	"github.com/redhat-developer/odo/pkg/dev"
 	"github.com/redhat-developer/odo/pkg/dev/common"
+	"github.com/redhat-developer/odo/pkg/devfile"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters"
+	"github.com/redhat-developer/odo/pkg/devfile/location"
 	"github.com/redhat-developer/odo/pkg/exec"
 	"github.com/redhat-developer/odo/pkg/log"
 	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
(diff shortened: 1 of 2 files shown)


pkg/dev/podmandev/podmandev.go
```diff
@@ -167,5 +168,8 @@ func (o *DevClient) watchHandler(ctx context.Context, pushParams adapters.PushPa
 		WatchFiles:   watchParams.WatchFiles,
 		Variables:    watchParams.Variables,
 	}
+
+	fmt.Fprintln(watchParams.Out)
+	log.Fwarning(watchParams.Out, "Changes to the Devfile not supported yet on Podman. Please restart 'odo dev' for such changes to be applied.\n")
```
feloy, line 173: This will be displayed every time, even if the user only modifies sources files.

I would instead add an `else` part to the `if equality.Semantic.DeepEqual(o.deployedPod, pod)` condition (in the reconcile.go file), and print the warning there

rm3l: Ah yes, thanks for the pointer. Initially, I had changed the part that looks for changes in the Devfile watcher (in watch.go), but then I had an issue when manually sync'ing with `--no-watch`.
This looks like a better approach. I'll make the change - thanks.

rm3l: One thing I just noticed: if we display the warning as part of `if !equality.Semantic.DeepEqual(o.deployedPod, pod)`, the problem is that this would work only if the changes in the Devfile affect the resulting Pod spec.

I noticed that if I change for example the `run` command only in the Devfile, the resulting Pod spec is still semantically the same as the previous one, so no warning is displayed. But, the changes were not taken into account (I think because we are still using the same Devfile Object from the context).

rm3l: I am thinking of just re-parsing the Devfile when the watch handler is called and comparing it to the currently loaded one. And display the warning if there are any differences.
@feloy What do you think?

The Kubedev implementation actually re-parses the Devfile when the watch handler is called: 
https://github.com/redhat-developer/odo/blob/020009b33cbef46c143075f8d830bb0284f872c6/pkg/dev/kubedev/kubedev.go#L156-L160

tests/integration/cmd_dev_test.go
```diff
@@ -692,45 +692,53 @@ ComponentSettings:
...
+									devSession.PressKey('p')
+								}
+								var err error
+								var stdoutBytes []byte
+								stdoutBytes, _, ports, err = devSession.WaitSync()
+								Expect(err).Should(Succeed())
+								stdout = string(stdoutBytes)
+							})
+
+							It("should expose the endpoint on localhost", func() {
```
feloy, line 715:
```suggestion
							It("should react on the Devfile modification", func() {
```

feloy (commented):
Also, could you add a test to check that no warning is displayed when the Devfile is not changed?
As the code to read the original and new Devfiles is not in the same place, I'm afraid it could drift over time

feloy (approved):

feloy:
Flaky e2e test

PR #6481, opened 2023-01-10
[podman] show logs when command fails

### LINKED ISSUE #6434: Implement Logs for podman platform
/kind feature

For the moment, displaying the logs of a container running in Podman is not implemented. For example, if the build or run command fails, the logs should be displayed. For the moment, odo panics.

feloy:
**What type of PR is this:**

/kind feature

Fixes #6434 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/podman/logs.go
+++ b/pkg/podman/logs.go
@@ -1,10 +1,35 @@
 package podman
 
-import "io"
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os/exec"
+
+	"k8s.io/klog"
+)
 
 // GetPodLogs returns the logs of the specified pod container.
 // All logs for all containers part of the pod are returned if an empty string is provided as container name.
 func (o *PodmanCli) GetPodLogs(podName, containerName string, followLog bool) (io.ReadCloser, error) {
-	// TODO(feloy)
-	return nil, nil
+	// TODO(feloy) implement followLog = true
+	args := []string{"pod", "logs"}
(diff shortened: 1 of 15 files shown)


pkg/podman/logs.go
```diff
@@ -1,10 +1,35 @@
...
+	args := []string{"pod", "logs"}
+	if containerName != "" {
+		args = append(args, "--container", podName+"-"+containerName)
+	}
+	args = append(args, podName)
+
+	cmd := exec.Command(o.podmanCmd, args...)
+	klog.V(3).Infof("executing %v", cmd.Args)
+	// We get the commbined output as podman logs outputs logs in stdout && stderr (when kubectl outputs all logs on stdout)
+	resultBytes, err := cmd.CombinedOutput()
```
rm3l, line 25: I was thinking of the case where "podman pod logs" returns a lot of lines, and this might increase memory usage.
I thought Go would propose a `CombinedOutputPipe` (similar to `Std{out,err}Pipe`, but it seems it is not the case. Instead, maybe we can use the solution depicted in https://github.com/golang/go/issues/27611#issuecomment-423525751, to return both stdout and stderr pipes as an `io.ReadCloser`?

```go
out, err := cmd.StdoutPipe()
// ... handle err
cmd.Stderr = cmd.Stdout
// ...
```

feloy:
Flaky e2e test

PR #6497, opened 2023-01-17
odo describe component: display remote source code location for each container component

valaparthvi:
**What type of PR is this:**
/kind feature

This PR shows remote source code location for each container component.
Currently it is only implemented for components with access to Devfile.

Fixes # 6453

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 


pkg/odo/cli/describe/component.go
```diff
@@ -443,7 +463,11 @@ func listComponentsNames(title string, devfileObj *parser.DevfileObj, typ v1alph
 	}
 	log.Info(title)
 	for _, container := range containers {
-		log.Printf("%s", container.Name)
+		printmsg := container.Name
+		if container.Container != nil {
+			printmsg += fmt.Sprintf("\n    ProjectSource: %s", container.Container.SourceMapping)
```
rm3l, line 468: Per the acceptance criteria, I think it should not report this at all if `mountSources` if `false`. Currently, it is displaying an empty value for a container component where `mountSources` is `false`:

```
$ odo describe component                    
Name: a-test-nodejs
Display Name: Node.js Runtime
Project Type: Node.js
Language: JavaScript
Version: 2.2.5
Description: Stack with Node.js 16
Tags: Node.js, Express, ubi8

Running in: None

Supported odo features:
 •  Dev: true
 •  Deploy: false
 •  Debug: true

Container components:
 •  runtime
    ProjectSource: 
```

rm3l: For consistency with how the field is named in the JSON output, I'd suggest renaming this into `Source Mapping`. What do you think?

rm3l (changes requested):

pkg/odo/cli/describe/component.go
```diff
@@ -306,9 +308,27 @@ func (o *ComponentOptions) describeDevfileComponent(ctx context.Context) (result
...
+		ComponentOptions: common.ComponentOptions{ComponentType: v1alpha2.ContainerComponentType},
+	})
+	if err != nil {
+		return
+	}
+	for _, comp := range components {
+		if *comp.Container.MountSources {
+			if comp.Container.SourceMapping == "" {
+				comp.Container.SourceMapping = generator.DevfileSourceVolumeMount
+				_ = cmp.DevfileData.Devfile.UpdateComponent(comp)
```
rm3l, line 326: Could we also `klog` the error if any instead of ignoring it? You never know! :)

pkg/odo/cli/describe/component.go
valaparthvi: >For consistency with how the field is named in the JSON output, I'd suggest renaming this into Source Mapping. What do you think?

I do like the idea of it being consistent, but I am not sure if `Source Mapping`  makes sense on UI. I guess I'll change that.

valaparthvi: >Per the acceptance criteria, I think it should not report this at all if mountSources if false. Currently, it is displaying an empty value for a container component where mountSources is false:

I thought it would be covered by the `updateWithRemoteSourceLocation` function, but perhaps you're right, I've added it.

pkg/odo/cli/describe/component.go
```diff
@@ -306,9 +309,30 @@ func (o *ComponentOptions) describeDevfileComponent(ctx context.Context) (result
...
 
+func updateWithRemoteSourceLocation(cmp *api.Component) {
+	components, err := cmp.DevfileData.Devfile.GetComponents(common.DevfileOptions{
+		ComponentOptions: common.ComponentOptions{ComponentType: v1alpha2.ContainerComponentType},
+	})
+	if err != nil {
+		return
+	}
+	for _, comp := range components {
+		if *comp.Container.MountSources {
```
rm3l, line 324: Can `comp.Container.MountSources` be nil?
I saw a `GetMountSources()` method which is supposed to handle this case and also return the appropriate boolean if `dedicatedPod` is set.

```suggestion
		if comp.Container.GetMountSources() {
```

pkg/odo/cli/describe/component.go
```diff
@@ -443,7 +467,11 @@ func listComponentsNames(title string, devfileObj *parser.DevfileObj, typ v1alph
 	}
 	log.Info(title)
 	for _, container := range containers {
-		log.Printf("%s", container.Name)
+		printmsg := container.Name
+		if container.Container != nil && *container.Container.MountSources {
```
rm3l, line 471: Same point here: can `container.Container.MountSources` be nil?

tests/integration/cmd_describe_component_test.go
```diff
@@ -597,4 +597,93 @@ var _ = Describe("odo describe component command tests", func() {
...
+						helper.DontMatchAllInOutput(output, []string{"sleeper-run\n    Source Mapping: /projects", "sleeper-build\n    Source Mapping:", "echo-er\n    Source Mapping:", "build-checker\n    Source Mapping:"})
+					},
+				},
+			} {
+				ctx := ctx
+				When(fmt.Sprintf("using %s and starting an odo dev session", ctx.title), helper.LabelPodmanIf(podman, func() {
+					var devSession helper.DevSession
+					BeforeEach(func() {
+						helper.CopyExample(filepath.Join("source", "nodejs"), commonVar.Context)
+						helper.Cmd("odo", "init", "--name", "aname", "--devfile-path", helper.GetExamplePath("source", "devfiles", "nodejs", ctx.devfile)).ShouldPass()
```
rm3l, line 654: One last thing: can you use the random `cmpName` as the component name instead of the hardcoded `aname`? This will help when running Podman tests in parallel (https://github.com/redhat-developer/odo/pull/6499).

rm3l (approved):
Thanks for the changes!

PR #6505, opened 2023-01-18
Port-forward Debug endpoints only when running `odo dev` with `--debug`

### LINKED ISSUE #6254: Port-forward `debug` endpoints only when running `odo dev` with `--debug`
/kind feature
/area dev

I think we should port-forward `debug` endpoints only when running `odo dev --debug`.
As discussed in https://github.com/redhat-developer/odo/issues/5988#issuecomment-1239497488, we should have a way to identify Debug endpoints.

## Why is this needed?
At this time, when running `odo dev` against a Devfile with a Debug endpoint, all endpoints are port-forwarded, including the debug endpoint.
But when trying (intentionally or not) to reach the corresponding local port, we might get an error because the container may not be listening on the debug port (because we are not running the debug command); as a consequence, port-forwarding is restarted.
Worse, if running `odo dev --random-ports`, new local random ports get generated.

```shell
# Sample project
odo init \
  --devfile-path https://raw.githubusercontent.com/redhat-developer/odo/main/tests/examples/source/devfiles/nodejs/devfile-with-debugrun.yaml \
  --starter nodejs-starter \
  --name nodejs-starter

# Dev Session
odo dev

# Once Dev Session is running, try to access the port-forwarded for the debug endpoint (5858), say: 40002
# This local port will never be reachable unless we stop `odo dev` and run `odo dev --debug` instead.
```

Related: https://github.com/devfile/registry/pull/122#issuecomment-1289039375

rm3l:
**What type of PR is this:**
/kind feature

Fixes #6254 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/dev/podmandev/pod.go
+++ b/pkg/dev/podmandev/pod.go
@@ -11,6 +11,7 @@ import (
 	"github.com/redhat-developer/odo/pkg/component"
 	"github.com/redhat-developer/odo/pkg/devfile/adapters/kubernetes/utils"
 	"github.com/redhat-developer/odo/pkg/labels"
+	"github.com/redhat-developer/odo/pkg/libdevfile"
 	"github.com/redhat-developer/odo/pkg/odo/commonflags"
 	"github.com/redhat-developer/odo/pkg/storage"
 	"github.com/redhat-developer/odo/pkg/util"
@@ -23,6 +24,7 @@ func createPodFromComponent(
 	devfileObj parser.DevfileObj,
 	componentName string,
 	appName string,
+	debug bool,
 	buildCommand string,
 	runCommand string,
 	debugCommand string,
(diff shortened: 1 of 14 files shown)


rm3l:
I added a test on Podman to check the access to the forwarded debug port, but it [does not pass for some reason](https://github.com/redhat-developer/odo/actions/runs/3950317394/jobs/6762689716#step:5:909). See https://github.com/redhat-developer/odo/issues/6510
We can work on fixing this issue on Podman in a subsequent PR.

feloy (approved):

feloy:

Flaky test
```
  [FAIL] odo dev command tests port-forwarding for the component when devfile has single endpoint when running odo dev when modifying memoryLimit for container in Devfile [BeforeEach] should react on the Devfile modification
```

PR #6519, opened 2023-01-20
Change port range used for port-forwarding endpoints to start at 20001

### LINKED ISSUE #6483: Change port range used for port-forwarding endpoints
/kind user-story

## User Story
As an odo user running odo inside DevSpaces, I want to use automatic port exposure done by the Che port plugin  So that I can access the application.

Currently this is not possible as Che plugin detect only opened ports for port numbers <32000

## Acceptance Criteria 
- [ ] port forwarding done by `odo dev` should use port numbers starting from 20000 instead of current 40000

## Links 
- Related bug: https://github.com/redhat-developer/odo/issues/6457

/kind user-story

feloy:
**What type of PR is this:**

/kind feature

Fixes #6483 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/dev/podmandev/pod.go
+++ b/pkg/dev/podmandev/pod.go
@@ -116,7 +116,7 @@ func getVolumeName(volume string, componentName string, appName string) string {
 
 func addHostPorts(containers []corev1.Container, debug bool, randomPorts bool, usedPorts []int) []api.ForwardedPort {
 	var result []api.ForwardedPort
-	startPort := 40001
+	startPort := 20001
 	endPort := startPort + 10000
 	usedPortsCopy := make([]int, len(usedPorts))
 	copy(usedPortsCopy, usedPorts)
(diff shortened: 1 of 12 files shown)


pkg/dev/podmandev/pod.go
```diff
@@ -113,7 +113,7 @@ func getVolumeName(volume string, componentName string, appName string) string {
 
 func addHostPorts(containers []corev1.Container, debug bool, usedPorts []int) []api.ForwardedPort {
 	var result []api.ForwardedPort
-	startPort := 40001
+	startPort := 20001
```
rm3l, line 116: What do you think of defining both the start and end ports as constants that could be used from the Podman and Kube implementations? To have a single place in case we need to change the range later..

feloy: I'm not fan of adding coupling between modules if not really coupled. Here, we may want to differentiate podman / cluster ports.

rm3l: Okay - that makes sense.

rm3l (approved):
LGTM - I just tested that it (Che's automatic exposure of ports) worked as expected when using odo from within Dev Spaces:

image

PR #6522, opened 2023-01-20
Display error message when odo dev fails on podman and clean resources

### LINKED ISSUE #6487: `odo dev` on podman exits in case of failure without an error message
/kind bug

**Operating System:**
Fedora 35
**Output of `odo version`:**
odo v3.5.0 (da9a9efb8f)
1. `odo init --devfile nodejs --starter nodejs-starter --name my-nodejs-app`
2. `sed -i "s/registry.access.redhat.com\/ubi8\/nodejs-16:latest/adsffdfe/" devfile.yaml`
3. `ODO_EXPERIMENTAL_MODE=true odo dev --platform=podman`

## Actual behavior
```sh
$ ODO_EXPERIMENTAL_MODE=true odo dev --platform=podman

  __
 /  \__     Developing using the "my-nodejs-app" Devfile
 \__/  \    Platform: podman
 /  \__/    odo version: v3.5.0
 \__/

 ✗  Deploying pod [2s]
Cleaning up resources
 ✗  exit status 125: 
```
`odo dev` fails because it did not find the image, but it exits without an error message.

## Expected behavior
`odo dev` should exit with an appropriate error message.

valaparthvi:

**What type of PR is this:**
/kind bug

This PR ensures `odo dev` on podman exits with actual failure from podman and also cleans up pod if it was created.

Fixes #6487 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

1. `odo init --devfile nodejs --starter nodejs-starter --name my-nodejs-app`
2. `sed -i "s/registry.access.redhat.com\/ubi8\/nodejs-16:latest/adsffdfe/" devfile.yaml`
3. `ODO_EXPERIMENTAL_MODE=true odo dev --platform=podman`

For 6501
1. `export PODMAN_CMD=echo`
2. `ODO_EXPERIMENTAL_MODE=true odo dev --platform=podman`

### DIFF
--- a/pkg/dev/podmandev/cleanup.go
+++ b/pkg/dev/podmandev/cleanup.go
@@ -4,12 +4,24 @@ import (
 	"context"
 	"fmt"
 	"io"
+
+	corev1 "k8s.io/api/core/v1"
+
+	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
+	"github.com/redhat-developer/odo/pkg/util"
 )
 
 func (o *DevClient) CleanupResources(ctx context.Context, out io.Writer) error {
 	fmt.Printf("Cleaning up resources\n")
 	if o.deployedPod == nil {
-		return nil
+		compName := odocontext.GetComponentName(ctx)
+		appName := odocontext.GetApplication(ctx)
+		name, err := util.NamespaceKubernetesObject(compName, appName)
+		if err != nil {
+			return nil
+		}
+		o.deployedPod = &corev1.Pod{}
+		o.deployedPod.SetName(name)
 	}
 	return o.podmanClient.CleanupPodResources(o.deployedPod)
 }
--- a/pkg/dev/podmandev/cleanup.go
+++ b/pkg/dev/podmandev/cleanup.go
@@ -4,24 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
-
-	corev1 "k8s.io/api/core/v1"
-
-	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
-	"github.com/redhat-developer/odo/pkg/util"
 )
 
 func (o *DevClient) CleanupResources(ctx context.Context, out io.Writer) error {
 	fmt.Printf("Cleaning up resources\n")
 	if o.deployedPod == nil {
-		compName := odocontext.GetComponentName(ctx)
-		appName := odocontext.GetApplication(ctx)
-		name, err := util.NamespaceKubernetesObject(compName, appName)
-		if err != nil {
-			return nil
-		}
-		o.deployedPod = &corev1.Pod{}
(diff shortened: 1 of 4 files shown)


pkg/podman/podman.go
```diff
@@ -81,7 +82,10 @@ func (o *PodmanCli) PlayKube(pod *corev1.Pod) error {
 	}()
 	if err = cmd.Wait(); err != nil {
 		if exiterr, ok := err.(*exec.ExitError); ok {
-			err = fmt.Errorf("%s: %s", err, string(exiterr.Stderr))
+			out := strings.Split(podmanOut, "\n")
+			// the last line is an empty new line; so we revert to the second last line for error
+			errLine := out[len(out)-2]
+			err = fmt.Errorf("%s: %s\n%s", err, string(exiterr.Stderr), errLine)
```
feloy, line 88: Why do we want to display only the last non-blank line, and not the entire content?
This seems very specific to a use-case where the error message would be in the last non-blank line.

pkg/dev/podmandev/cleanup.go
```diff
@@ -4,12 +4,24 @@ import (
...
 	if o.deployedPod == nil {
-		return nil
+		compName := odocontext.GetComponentName(ctx)
+		appName := odocontext.GetApplication(ctx)
+		name, err := util.NamespaceKubernetesObject(compName, appName)
+		if err != nil {
+			return nil
+		}
+		o.deployedPod = &corev1.Pod{}
+		o.deployedPod.SetName(name)
```
feloy, line 24: Is this related to this PR? What is the problem fixed by this change?

feloy (changes requested):

pkg/dev/podmandev/cleanup.go
valaparthvi: In case of some failures, pod is still created, this change attempts at fixing this problem.

pkg/podman/podman.go
valaparthvi: On a second thought, the last line (err line) usually sums up the core issue, so showing just that should be fine. WDYT?

feloy: I would prefer for the first versions to display the complete output, to have the maximum of information. If we consider it is too much, we could reduce it

pkg/dev/podmandev/cleanup.go
feloy: I would prefer to fix the problem why deployedPod is null when a Pod is actually deployed

valaparthvi: Podman creates the pod even if something is incorrect with the manifest, for e.g. non-existent image.

feloy (approved):

feloy:
Flaky e2e test

PR #6548, opened 2023-01-30
Add support for OpenShift Devfile components

### LINKED ISSUE #5704: Support `openshift` Devfile components (the same way as `kubernetes` components)

rm3l:
**What type of PR is this:**

Fixes #5704 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 


valaparthvi:
IBM K8s failure is related to because there was an abrupt logout from the cluster.

While the following tests failed on OC:
```
Summarizing 3 Failures:
  [FAIL] odo dev command tests 1. devfile contains composite apply command when odo dev is running [It] should execute the composite apply commands successfully
  /go/odo_1/tests/integration/cmd_dev_test.go:1684
  [FAIL] odo dev debug command tests when a composite apply command is used as debug command [It] should execute the composite apply commands successfully
  /go/odo_1/tests/helper/helper_oc.go:34
  [FAIL] odo dev command tests checking if odo dev matches local Devfile K8s resources and remote resources when odo dev is executed to run a devfile containing a k8s resource with apply command [BeforeEach] should have deleted the old resource and created the new resource
  /go/odo_1/tests/integration/cmd_dev_test.go:562
```

1. 
```
  Expected
      <[]uint8 | len:124, cap:124>: [78, 65, 77, 69, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 82, 69, 65, 68, 89, 32, 32, 32, 85, 80, 45, 84, 79, 45, 68, 65, 84, 69, 32, 32, 32, 65, 86, 65, 73, 76, 65, 66, 76, 69, 32, 32, 32, 65, 71, 69, 10, 109, 121, 45, 111, 112, 101, 110, 115, 104, 105, 102, 116, 45, 99, 111, 109, 112, 111, 110, 101, 110, 116, 32, 32, 32, 48, 47, 49, 32, 32, 32, 32, 32, 49, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 48, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 55, 49, 115, 10]
  not to contain substring
      <string>: my-openshift-component
```

feloy (approved):

PR #6553, opened 2023-01-31
Run podman tests on 10 nodes

feloy:
**What type of PR is this:**

/kind tests

### DIFF
--- a/.github/workflows/podman-test.yaml
+++ b/.github/workflows/podman-test.yaml
@@ -23,6 +23,10 @@ jobs:
 
     - name: Run Integration tests
       env:
+        # This should ideally be a configuration in the GH repo (secret or variable).
+        # This is currently hard-coded because GH won't expose secrets or variables to PRs created from forks.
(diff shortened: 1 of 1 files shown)


.github/workflows/podman-test.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
 
     - name: Run Integration tests
       env:
-        PODMAN_EXEC_NODES: ${{ secrets.PODMAN_TEST_EXEC_NODES }}
+        PODMAN_EXEC_NODES: 10
```
rm3l, line 26: Suggestion to add a comment.

```suggestion
        # This should ideally be a configuration in the GH repo (secret or variable).
        # This is currently hard-coded because GH won't expose secrets or variables to PRs created from forks.
        # Hopefully, variables (which are non-sensible by definition) will be passed to workflows triggered by PRs from forks.
        # See https://github.com/community/community/discussions/44322
        PODMAN_EXEC_NODES: 10
```

valaparthvi:
13 min, that's a huge up from 1h!
:100:

rm3l (approved):

PR #6604, opened 2023-02-17
Add mockgen installation to the script

valaparthvi:
**What type of PR is this:**

Fixes #

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

1. Remove `mockgen` binary.
```sh
rm $(which mockgen)
```
2. Run the script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,8 @@ clean:
 
 .PHONY: goget-tools
 goget-tools:  ## Install binaries of the tools used by odo (current list: golangci-lint, mockgen)
-	(cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION) && go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION))
+	cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION)
+	cd / && go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION)
 
 .PHONY: goget-ginkgo
 goget-ginkgo:
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,9 @@ GOLANGCI_LINT_VERSION=1.49.0
 
 RUN_GINKGO = go run -mod=vendor github.com/onsi/ginkgo/v2/ginkgo
 
+# making sure that mockgen binary version is the same as github.com/golang/mock library version
+GO_MOCK_LIBRARY_VERSION = $(shell go list -mod=readonly -m -f '{{.Version}}' github.com/golang/mock)
+
 default: bin
 
 .PHONY: help
@@ -127,8 +130,8 @@ clean:
 	@rm -rf $(FILES)
 
 .PHONY: goget-tools
-goget-tools:
-	(cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION))
+goget-tools:  ## Install binaries of the tools used by odo (current list: golangci-lint, mockgen)
+	(cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION) && go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION))
 
 .PHONY: goget-ginkgo
 goget-ginkgo:
--- a/scripts/mockgen.sh
+++ b/scripts/mockgen.sh
@@ -1,7 +1,18 @@
 #!/usr/bin/env bash
 
-# Use this script to regererate generated mock files
+# Use this script to regenerate generated mock files
 # after changing signatures of interfaces in these packages
+
+# check if mockgen is installed
+mockgen --version
+if [ $? -eq 0 ]; then
+  echo "✓ mockgen is installed"
+else
+  echo "mockgen not found; installing it..."
+  go install github.com/golang/mock/mockgen@v1.6.0
+  echo "✓ mockgen is installed"
+fi
+
 
 mockgen -source=pkg/kclient/interface.go \
     -package kclient \
--- a/scripts/mockgen.sh
+++ b/scripts/mockgen.sh
@@ -3,101 +3,95 @@
 # Use this script to regenerate generated mock files
 # after changing signatures of interfaces in these packages
 
-# check if mockgen is installed
-mockgen --version
-if [ $? -eq 0 ]; then
-  echo "✓ mockgen is installed"
-else
-  echo "mockgen not found; installing it..."
-  go install github.com/golang/mock/mockgen@v1.6.0
-  echo "✓ mockgen is installed"
-fi
+gomockVersion=$(go list -mod=readonly -m -f '{{.Version}}' github.com/golang/mock)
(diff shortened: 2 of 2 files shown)


scripts/mockgen.sh
```diff
@@ -1,8 +1,19 @@
 #!/usr/bin/env bash
 
-# Use this script to regererate generated mock files
+# Use this script to regenerate generated mock files
 # after changing signatures of interfaces in these packages
 
+# check if mockgen is installed
+mockgen --version
+if [ $? -eq 0 ]; then
```
feloy, line 8: You could also test the version installed, as the output differs depending on the version

scripts/mockgen.sh
```diff
@@ -1,8 +1,19 @@
...
+  echo "✓ mockgen is installed"
+else
+  echo "mockgen not found; installing it..."
+  go install github.com/golang/mock/mockgen@v1.6.0
```
rm3l, line 12: Could we include this in the `goget-tools` `Make` target? I find it a bit confusing that the script also installs `mockgen`.. At least, we could display a message instructing to install `mockgen` with `make goget-tools` if it is not installed or if it does not have the right version.
Also, I see that we now have 2 versions that will need to be kept in sync (the library in `go.mod` and the version of the `mockgen` binary).
Would it be possible to have a single source (for future maintenance)?

rm3l: With hindsight, I think it would be more maintainable if we could make the script detect the appropriate version from `go.mod` using `go mod graph` and use it with `go run -mod=readonly` for example; this way, we won't even need to `go install` it, and our source of truth is `go.mod` (well, how the dependency graph is actually resolved by `go`).

Quick example of how we could do this:

```bash
set -x

gomockCoordinates=$(go mod graph | grep '^github.com/redhat-developer/odo\sgithub.com/golang/mock@' | awk '{print $2}')

mockgen="go run -mod=readonly $(echo "$gomockCoordinates" | sed 's/\/mock@/\/mock\/mockgen@/g')"

# Now we can use $mockgen as our command, e.g.:
$mockgen -source=pkg/kclient/interface.go \
    -package kclient \
    -destination pkg/kclient/mock_Client.go

...
```

@feloy @valaparthvi What do you think?

feloy: This solution introduces many dependencies (every time I see greps and seds I know something will break if a space changes somewhere :( )

I would prefer to duplicate the version in go.mod and Makefile (by installing the binary) instead

rm3l: > This solution introduces many dependencies (every time I see greps and seds I know something will break if a space changes somewhere :( )
> 
:) Okay, I agree using `grep` and `sed` is a bit fragile, but I wanted to discuss the overall idea.

How about that one using `go list`. I stumbled upon this, to extract the version in a much simpler way:
```
$ go list -mod=readonly -m -f '{{.Version}}' github.com/golang/mock
v1.6.0
```

So we could simply use:
```
gomockVersion=$(go list -mod=readonly -m -f '{{.Version}}' github.com/golang/mock)
mockgen="go run -mod=readonly github.com/golang/mock/mockgen@$gomockVersion"

echo Using mockgen "$($mockgen --version)"
```

> I would prefer to duplicate the version in go.mod and Makefile (by installing the binary) instead
>
If we go that way, I'd suggest documenting both in the Makefile and `go.mod` that we need to update both places.
Even there, I think we might still be out of sync in case another dependency requires a higher version. For example, if I downgrade `github.com/golang/mock` to `v1.5.0` in `go.mod`, `go mod` will automatically set it to `v1.6.0` because one of our dependencies requires a higher version (`v1.6.0`).

feloy: I'm not sure that the important thing is that we use the same version between binary and library.

The problem we have currently is that different developers have different version of the binary, and the mocks files are modified endlessly because of version change depending on who re-runs the script.

rm3l: > I'm not sure the important thing is to use the same version between binary and library.
> 
> The problem we have currently is that different developers have different version of the binary, and the mocks files are modified endlessly because of version change depending on who re-runs the script.
>

In [their Readme](https://github.com/golang/mock#readme), they state that "You should try to keep the library in sync with the version of mockgen used to generate your mocks."; that's why I thought ensuring everyone is using the right version while not needing to install mockgen would probably make sense.

But okay, no problem, it's fine if we duplicate the version in both `go.mod` and `Makefile`.

valaparthvi: > ```
> gomockVersion=$(go list -mod=readonly -m -f '{{.Version}}' github.com/golang/mock)
> mockgen="go run -mod=readonly github.com/golang/mock/mockgen@$gomockVersion"
> 
> echo Using mockgen "$($mockgen --version)"
> ```

I liked the solution Armel suggested in https://github.com/redhat-developer/odo/pull/6604#discussion_r1114053184, so I have modified the mockgen script to use the library version and `go run -mod=readonly` to run the scripts.

However, I have also added instructions to install mockgen script as a part of the goget-tools target in case someone wants to run mockgen on a single file. Here as well, we make sure we fetch the library version and use it.

feloy (approved):

Makefile
```diff
@@ -127,8 +130,8 @@ clean:
 	@rm -rf $(FILES)
 
 .PHONY: goget-tools
-goget-tools:
-	(cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION))
+goget-tools:  ## Install binaries of the tools used by odo (current list: golangci-lint, mockgen)
+	(cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION) && go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION))
```
rm3l, line 134: nit: suggestion to improve readability (and also if we need to do additional things later):

```suggestion
	( \
		cd / && \
		go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION) && \
		go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION) \
	)
```

rm3l: or we can simply execute all those commands in separate shells:

```suggestion
	cd / && go install -mod=mod github.com/golangci/golangci-lint/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION)
	cd / && go install -mod=mod github.com/golang/mock/mockgen@$(GO_MOCK_LIBRARY_VERSION)
```

rm3l (approved):

rm3l:
```
  user-guides/quickstart/docs-mdx/java/java_odo_dev_output.mdx
  Expected
      <string>: Â Â (
      Â Â 	"""
      Â Â 	... // 10 identical lines
      Â Â 	âœ“  Added storage m2 to component
      Â Â 	âš   Pod is Pending
      -Â 	âœ“  Pod is Running
      Â Â 	âœ“  Syncing files into the container [1s]
      Â Â 	âœ“  Building your application in container (command: build) [1s]
      Â Â 	... // 9 identical lines
      Â Â 	[Ctrl+c] - Exit and delete resources from the cluster
      Â Â 	[p] - Manually apply local changes to the application on the cluster
      +Â 	âœ“  Pod is Running
      Â Â 	```
      Â Â 	"""
      Â Â )
      
  to be empty
  In [It] at: /go/odo_1/tests/documentation/user-guides/doc_user_guides_quickstart_test.go:256
```

Flaky test - addressed by https://github.com/redhat-developer/odo/pull/6545

feloy:
Non related tests

Issue #6510, opened 2023-01-18
On Podman, unable to access local ports forwarded to container applications listening on the loopback interface (e.g. debug endpoints)


rm3l:
/kind bug
/area dev
/area odo-on-podman

## What versions of software are you using?

**Operating System:**
Fedora 37, kernel 6.1.6-200.fc37.x86_64

**Output of `odo version`:**
`odo v3.5.0` (8dbf42e5e)

## How did you run odo exactly?

```
$ mkdir /tmp/nodejs && cd /tmp/nodejs
$ odo init --name debug-nodejs --devfile nodejs --starter nodejs-starter
$ ODO_EXPERIMENTAL_MODE=t odo dev --platform=podman --debug

...
 •  Executing the application (command: debug)  ...
 -  Forwarding from 127.0.0.1:40001 -> 3000
 -  Forwarding from 127.0.0.1:40002 -> 5858

...
```

As expected, trying to access the forwarded application port (40001 here) works correctly:
```
$ curl -i http://localhost:40001
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 39
ETag: W/"27-Jw3xdDamjf5jthPp3a/zd95HG0Y"
Date: Wed, 18 Jan 2023 15:04:06 GMT
Connection: keep-alive
Keep-Alive: timeout=5

Hello from Node.js Starter Application!%
```

But trying to access the application on the forwarded debug port (40002 here) returns an error:
```
$ curl -i http://localhost:40002
curl: (56) Recv failure: Connection reset by peer
```

## Actual behavior
Accessing the application on the forwarded debug port (40002 here) returns an error:
```
$ curl -i http://localhost:40002
curl: (56) Recv failure: Connection reset by peer
```

## Expected behavior
When running `odo dev` against a cluster, we get the expected HTTP response from the forwarded application debug port:
```
$ curl -i http://localhost:40002
HTTP/1.0 400 Bad Request
Content-Type: text/html; charset=UTF-8

WebSockets request was expected
```

I was expecting the same behavior on Podman.

## Any logs, error output, etc?

<details>
<summary>podman kube generate</summary>

```yaml
# Save the output of this file and use kubectl create -f to import
# it into Kubernetes.
#
# Created with podman-4.3.1
apiVersion: v1
kind: Pod
metadata:
  annotations:
    io.kubernetes.cri-o.ContainerType/debug-nodejs-app-runtime: container
[...]

PR #6620, opened 2023-02-23
On Podman, detect if application is listening on the loopback interface, and either error out or not depending on `--ignore-localhost`

rm3l:
**What type of PR is this:**
/kind feature
/area dev
/area odo-on-podman

The way `odo` currently does port-forwarding on Podman is via a `HostPort` in the relevant container in the pod spec. But as reported in https://github.com/redhat-developer/odo/issues/6510 and https://github.com/containers/podman/issues/17353, Podman won't forward traffic to the container if the container port is bound to the container loopback interface.
As discussed in https://github.com/redhat-developer/odo/issues/6510#issuecomment-1439986558 (and similar to how things work on DevSpaces), this PR tries to detect if ports that need to be forwarded are bound to the container loopback interface. If this is the case:
- by default, `odo dev` on Podman will error out with an error message indicating the issue, along with a recommendation to either change the application to listen on 0.0.0.0, or to run `odo dev` with  `--forward-localhost` (PR to follow up soon).
- if `odo dev` is run with `--ignore-localhost`, this message will be displayed as a warning, but this won't prevent `odo dev` from running. However, any request sent out to the local port forwarded to the container on its loopback interface might not work with Podman.

I'll create a subsequent PR that adds a new `--forward-localhost` flag, the goal of which is to make port-forwarding work in such cases via a side container. It'll be inspired by what had been done in https://github.com/redhat-developer/odo/pull/6589

This relates to #6510 

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

This behavior should *not* affect the default cluster mode.

To test the changes on Podman, either use a project with an application listening on some port on localhost, or leverage some existing stacks like Node.JS which have debuggers listening on localhost.

```
$ mkdir /tmp/nodejs && cd /tmp/nodejs
$ odo init --name debug-nodejs --devfile nodejs --starter nodejs-starter
```

- Without `--ignore-localhost`, `odo dev` on Podman should start but error out after detecting that the port is bound to the container loopback interface:
```
$ ODO_EXPERIMENTAL_MODE=t odo dev --platform=podman --debug

...
 ✓  Deploying pod [3s]
 ✓  Building your application in container (command: install) [2s]
 •  Executing the application (command: debug)  ...
 ✗  Detected that the following port(s) can be reached only via the container loopback interface: debug (5858).
Port forwarding on Podman currently does not work with applications listening on the loopback interface.
Either change the application to make those port(s) reachable on all interfaces (0.0.0.0), or rerun 'odo dev' with any of the following options:
- --ignore-localhost: no error will be returned by odo, but port-forwarding on those ports might not work on Podman.
- --forward-localhost: odo will inject a dedicated side container to redirect traffic to such ports.
Cleaning up resources
 ✗  Finished executing the application (command: debug) [16s]
 ✗  cannot make port forwarding work with ports bound to the loopback interface only

```

NOTE: only ports that are expected to be forwarded are checked. In the example above, if I run `ODO_EXPERIMENTAL_MODE=t odo dev --platform=podman` (without `--debug`), `odo dev` should start and work normally.

- With `--ignore-localhost`, `odo dev` on Podman should start but display a warning message after detecting that the port is bound to the container loopback interface:
```
$ ODO_EXPERIMENTAL_MODE=t odo dev --platform=podman --debug --ignore-localhost

...
 ✓  Deploying pod [3s]
 ✓  Building your application in container (command: install) [4s]
 •  Executing the application (command: debug)  ...
 ⚠  Detected that the following port(s) can be reached only via the container loopback interface: debug (5858).
Port forwarding on Podman currently does not work with applications listening on the loopback interface.
Either change the application to make those port(s) reachable on all interfaces (0.0.0.0), or rerun 'odo dev' with '--forward-localhost' to make port-forwarding work with such ports.
 -  Forwarding from 127.0.0.1:20001 -> 3000
 -  Forwarding from 127.0.0.1:20002 -> 5858

↪ Dev mode
 Status:
 Watching for changes in the current directory /tmp/test/nodejs

 Keyboard Commands:
```

### DIFF
--- a/pkg/dev/podmandev/reconcile.go
+++ b/pkg/dev/podmandev/reconcile.go
@@ -118,37 +118,9 @@ func (o *DevClient) reconcile(
 	// By default, Podman will not forward to container applications listening on the loopback interface.
 	// So we are trying to detect such cases and act accordingly.
 	// See https://github.com/redhat-developer/odo/issues/6510#issuecomment-1439986558
-	var loopbackPorts []api.ForwardedPort
-	if len(pod.Spec.Containers) != 0 {
-		loopbackPorts, err = port.DetectRemotePortsBoundOnLoopback(o.execClient, pod.Name, pod.Spec.Containers[0].Name, fwPorts)
-		if err != nil {
-			return fmt.Errorf("unable to detect container ports bound on the loopback interface: %w", err)
-		}
-		klog.V(4).Infof("detected %d ports bound on the loopback interface in the pod", len(loopbackPorts))
-	}
-	if len(loopbackPorts) != 0 {
-		list := make([]string, 0, len(loopbackPorts))
-		for _, p := range loopbackPorts {
-			list = append(list, fmt.Sprintf("%s (%d)", p.PortName, p.ContainerPort))
-		}
-		msg := fmt.Sprintf(`Detected that the following port(s) can be reached only via the container loopback interface: %s.
-Port forwarding on Podman currently does not work with applications listening on the loopback interface.
-Either change the application to make those port(s) reachable on all interfaces (0.0.0.0), or rerun 'odo dev' with `, strings.Join(list, ", "))
-		if options.IgnoreLocalhost {
-			msg += "'--forward-localhost' to make port-forwarding work with such ports."
-		} else {
-			msg += `any of the following options:
-- --ignore-localhost: no error will be returned by odo, but port-forwarding on those ports might not work on Podman.
-- --forward-localhost: odo will inject a dedicated side container to redirect traffic to such ports.`
-		}
-
-		if !options.IgnoreLocalhost {
-			log.Errorf(msg)
-			return errors.New("cannot make port forwarding work with ports bound to the loopback interface only")
-		}
-
-		// Only a warning if using --ignore-localhost
-		log.Warningf(msg)
+	err = o.handleLoopbackPorts(options, pod, fwPorts)
+	if err != nil {
+		return err
 	}
 
 	for _, fwPort := range fwPorts {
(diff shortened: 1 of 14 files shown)


pkg/dev/podmandev/reconcile.go
```diff
@@ -113,6 +115,42 @@ func (o *DevClient) reconcile(
 		componentStatus.RunExecuted = true
 	}
 
+	// By default, Podman will not forward to container applications listening on the loopback interface.
```
feloy, line 118: Could we make some function with this new code, to try and reduce the lebgth of the reconcile function?

rm3l: Extract this logic into a separate method. See [`17e0292` (#6620)](https://github.com/redhat-developer/odo/pull/6620/commits/17e0292a8210b744532473cb6eeb734090ea0793)

rm3l:
```
  Expected
      <*url.Error | 0xc0007d3170>: {
          Op: "Post",
          URL: "http://127.0.0.1:65530/api/newuser",
          Err: <*errors.errorString | 0xc00007c130>{s: "EOF"},
      }
  to be nil
  In [It] at: C:/Users/Administrator.ANSIBLE-TEST-VS/3329/tests/e2escenarios/e2e_test.go:306

  There were additional failures detected after the initial failure.  Here's a summary - for full details run Ginkgo in verbose mode:
    [FAILED] in [AfterEach] at C:/Users/Administrator.ANSIBLE-TEST-VS/3329/tests/helper/helper_filesystem.go:48


Summarizing 1 Failure:
  [FAIL] E2E Test starting with non-empty Directory add Binding [It] should verify developer workflow of using binding as env in innerloop
  C:/Users/Administrator.ANSIBLE-TEST-VS/3329/tests/e2escenarios/e2e_test.go:306
```

Flaky E2E test - reported in https://github.com/redhat-developer/odo/issues/6582

PR #6634, opened 2023-03-06
Bump to v3.8.0

feloy:


rm3l (approved):

feloy:
Flaky test

rm3l:
@feloy Can we wait for tagging until https://github.com/redhat-developer/odo/issues/6636 is fixed? Otherwise, we might end up with annoying situations similar to https://github.com/redhat-developer/odo/issues/6379, and people might not be able to use `odo init` / `odo registry` / `odo preference view` at all if the cluster is not reachable.
I think we should also wait until https://github.com/redhat-developer/odo/issues/6637 is fixed too.

feloy:
> @feloy Can we wait for tagging until #6636 is fixed? Otherwise, we might end up with annoying situations similar to #6379, and people might not be able to use `odo init` / `odo registry` / `odo preference view` at all if the cluster is not reachable.
> I think we should also wait until #6637 is fixed too.

Yes, sure, good catch, we will wait for these fixes

feloy:
Flaky e2e test

PR #6639, opened 2023-03-07
Ignore errors that might happen when getting Devfile Registries from the cluster

### LINKED ISSUE #6636: odo fails to fetch devfiles when disconnected from the cluster
/kind bug

**Operating System:** 
MacOS

**Output of `odo version`:**
odo v3.7.0 (5c12434b4)

1. If working with OC cluster, logout of the cluster.
2. Run `odo registry` or `odo init`.

## Actual behavior
In an attempt to fetch in-cluster devfile registry, `odo` fails to fetch offline registries when disconnected from the cluster.

## Expected behavior
`odo` should fetch whatever registries it can and should not fail if it fails to fetch in-cluster registries.

```sh
$ odo registry
 ✗  Get "https://a586844e14ae14dd38a5b322458498a0-d4a4d9f009fe7434.elb.us-east-1.amazonaws.com:6443/apis/registry.devfile.io/v1alpha1/namespaces/default/devfileregistrieslists": dial tcp: lookup a586844e14ae14dd38a5b322458498a0-d4a4d9f009fe7434.elb.us-east-1.amazonaws.com: no such host
```

```sh
$ odo init

  __
 \__/  \    Files: No source code detected, a starter project will be created in the current directory
 /  \__/    odo version: v3.7.0
 \__/

 ✗  please provide options to select from
the command failed, the devfile has been removed from current directory
```

rm3l:
**What type of PR is this:**
/kind bug
/area registry

Fixes #6636 

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -67,9 +67,11 @@ func (o RegistryClient) GetDevfileRegistries(registryName string) ([]api.Registr
 	if o.kubeClient != nil {
 		clusterRegistries, err := o.kubeClient.GetRegistryList()
 		if err != nil {
-			return nil, err
+			// #6636 : errors should not be blocking
+			klog.V(3).Infof("failed to get Devfile registries from the cluster: %v", err)
+		} else {
+			allRegistries = append(allRegistries, clusterRegistries...)
 		}
-		allRegistries = append(allRegistries, clusterRegistries...)
 	}
 	allRegistries = append(allRegistries, o.preferenceClient.RegistryList()...)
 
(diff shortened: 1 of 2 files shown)


valaparthvi (approved):
Thank you for fixing this. I was going to suggest adding a test case for this, but I see we already have the test cases but I am surprised they did not catch the error.

rm3l:
> Thank you for fixing this. I was going to suggest adding a test case for this, but I see we already have the test cases but I am surprised they did not catch the error.

The code actually works (and is tested) when the user is not logged into the cluster or when not using any cluster at all..
I don't think we have test cases for clusters that might not be reachable at all (as was the case in #6636). We have test cases for either no-cluster or unauthenticated (but still reachable).
That's why this PR is ignoring any errors returned while trying to get the list of registries from the cluster, and this should be covered by unit tests.

PR #6642, opened 2023-03-08
Fix potential issue with `odo describe component` integration tests when Podman is installed

rm3l:
**What type of PR is this:**
/kind bug
/area testing

```
content of path "message" should contain "no component found with name \"unknown-name\" in the namespace \"cmd-describe-component-test70ezo\"" but is "no component found with name \"unknown-name\""
Expected
    <string>: no component found with name "unknown-name"
to contain substring
    <string>: no component found with name "unknown-name" in the namespace "cmd-describe-component-test70ezo"
```

I noticed this error message when running the following integration tests locally:
- `odo describe component command tests when creating a component [It] should not describe the component from another directory`
-  `odo describe component command tests [It] should fail, with cluster`

 `odo describe component` tries to describe the component on both Podman (if available) and cluster. Because I have Podman installed locally, the error message is not actually the one expected in the tests.
When running locally with Podman installed, the error message no longer contains the reference to the namespace.

&mdash;

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

With the Podman binary installed locally, the tests mentioned above should now pass with the changes here.

### DIFF
--- a/tests/integration/cmd_describe_component_test.go
+++ b/tests/integration/cmd_describe_component_test.go
@@ -67,9 +67,26 @@ var _ = Describe("odo describe component command tests", func() {
 		})
 	}
 
-	It("should fail, with cluster", func() {
+	It("should fail, with default cluster mode", func() {
 		By("running odo describe component -o json with an unknown name", func() {
 			res := helper.Cmd("odo", "describe", "component", "--name", "unknown-name", "-o", "json").ShouldFail()
+			stdout, stderr := res.Out(), res.Err()
+			Expect(helper.IsJSON(stderr)).To(BeTrue())
+			Expect(stdout).To(BeEmpty())
+			helper.JsonPathContentContain(stderr, "message", "no component found with name \"unknown-name\"")
+		})
+
+		By("running odo describe component with an unknown name", func() {
+			res := helper.Cmd("odo", "describe", "component", "--name", "unknown-name").ShouldFail()
+			stdout, stderr := res.Out(), res.Err()
+			Expect(stdout).To(BeEmpty())
+			Expect(stderr).To(ContainSubstring("no component found with name \"unknown-name\""))
+		})
+	})
+
+	It("should fail, with cluster", func() {
+		By("running odo describe component -o json with an unknown name", func() {
+			res := helper.Cmd("odo", "describe", "component", "--name", "unknown-name", "--platform", "cluster", "-o", "json").ShouldFail()
 			stdout, stderr := res.Out(), res.Err()
 			Expect(helper.IsJSON(stderr)).To(BeTrue())
 			Expect(stdout).To(BeEmpty())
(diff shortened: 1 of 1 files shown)


tests/integration/cmd_describe_component_test.go
```diff
@@ -67,19 +67,36 @@ var _ = Describe("odo describe component command tests", func() {
...
 
-	It("should fail, with cluster", func() {
+	It("should fail, with default cluster mode", func() {
 		By("running odo describe component -o json with an unknown name", func() {
 			res := helper.Cmd("odo", "describe", "component", "--name", "unknown-name", "-o", "json").ShouldFail()
 			stdout, stderr := res.Out(), res.Err()
 			Expect(helper.IsJSON(stderr)).To(BeTrue())
 			Expect(stdout).To(BeEmpty())
-			helper.JsonPathContentContain(stderr, "message", "no component found with name \"unknown-name\" in the namespace \""+commonVar.Project+"\"")
+			helper.JsonPathContentContain(stderr, "message", "no component found with name \"unknown-name\"")
```
valaparthvi, line 76: I understand the behavior of this message, but I find it a little weird. If the default mode is cluster, shouldn't the namespace part be printed? It would make sense to not print it if we were printing output for both cluster and podman like we do for `odo delete component` command. WDYT? Although this is a case that can be discussed in a different issue.

valaparthvi (approved):

tests/integration/cmd_describe_component_test.go
rm3l: The default mode for `odo describe component` is to try to get info from both platforms, unless specified otherwise.
So yes, the current behavior looks fine to me if Podman is detected when running in the default mode.

rm3l:
```
Summarizing 1 Failure:
  [FAIL] odo dev command tests port-forwarding for the component when devfile has single endpoint when running odo dev when modifying memoryLimit for container in Devfile [BeforeEach] should react on the Devfile modification
  /go/odo_1/tests/helper/helper_run.go:64

Ran 429 of 779 Specs in 1223.965 seconds
FAIL! -- 428 Passed | 1 Failed | 0 Pending | 350 Skipped
```

Flaky test - reported in https://github.com/redhat-developer/odo/issues/6601

rm3l:
```
  Expected
      <*url.Error | 0xc000450000>: {
          Op: "Post",
          URL: "http://127.0.0.1:59998/api/newuser",
          Err: <*errors.errorString | 0xc00007c130>{s: "EOF"},
      }
  to be nil
  In [It] at: C:/Users/Administrator.ANSIBLE-TEST-VS/3384/tests/e2escenarios/e2e_test.go:306

  There were additional failures detected after the initial failure.  Here's a summary - for full details run Ginkgo in verbose mode:
    [FAILED] in [AfterEach] at C:/Users/Administrator.ANSIBLE-TEST-VS/3384/tests/helper/helper_filesystem.go:48

Summarizing 1 Failure:
  [FAIL] E2E Test starting with non-empty Directory add Binding [It] should verify developer workflow of using binding as env in innerloop
  C:/Users/Administrator.ANSIBLE-TEST-VS/3384/tests/e2escenarios/e2e_test.go:306

Ran 7 of 7 Specs in 208.687 seconds
FAIL! -- 6 Passed | 1 Failed | 0 Pending | 0 Skipped
```

Flaky E2E test - reported in https://github.com/redhat-developer/odo/issues/6582

PR #6660, opened 2023-03-14
Fix issue with conflicting ports when starting multiple Dev sessions on Podman

### LINKED ISSUE #6612: Cannot start 2 different Dev sessions on Podman due to conflicting host ports
/kind bug

**Operating System:**
Fedora 37, kernel 6.1.12-200.fc37.x86_64
MacOS Ventura 13, Darwin Kernel Version 22.3.0

**Output of `odo version`:**
odo v3.7.0 (e32024f87)

**Output of `podman version`:**
```
Client:       Podman Engine
Version:      4.4.1
API Version:  4.4.1
Go Version:   go1.19.5
Built:        Thu Feb  9 11:58:53 2023
OS/Arch:      linux/amd64
```

In one terminal:
```
$ mkdir /tmp/nodejs1 && cd /tmp/nodejs1
$ odo init --name my-nodejs --devfile nodejs --starter nodejs-starter
$ ODO_EXPERIMENTAL_MODE=t odo dev --debug --platform=podman
```

Wait until the Dev session is started.

In a separate terminal:
```
$ mkdir /tmp/nodejs2 && cd /tmp/nodejs2
$ odo init --name my-nodejs-2 --devfile nodejs --starter nodejs-starter
$ ODO_EXPERIMENTAL_MODE=t odo dev --debug --platform=podman
```

## Actual behavior

The dev session for `my-nodejs-2` (second terminal) wouldn't start, because we are trying to bind to port 20001, which is already used by the dev session started by `my-nodejs` (first terminal).
The error is explicit: `[... rootlessport listen tcp 0.0.0.0:20001: bind: address already in use]`

```
$ ODO_EXPERIMENTAL_MODE=t odo dev --debug --platform=podman

  __
 /  \__     Developing using the "my-nodejs-2" Devfile
 \__/  \    Platform: podman
 /  \__/    odo version: v3.7.0
 \__/

 ✗  Deploying pod [6s]
Cleaning up resources
 ✗  exit status 125: 
Complete Podman output:
Trying to pull registry.access.redhat.com/ubi8/nodejs-16:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob sha256:f6020ef1f0c813cb6e68b95ffe2ba1eb76a61b5083e624cb8a72a51626191bf7
Copying blob sha256:1eb1cc491a1bd8e2827ef1c90d2253d5aca4348cc5642ec2206c85dfd20a22e1
Copying blob sha256:ea0a20a2c44861ba5e4cd80a32868b6f8265cd3ee81310d4d493a1b50d1c0d8c
Copying config sha256:bfc987a83b9e6ffa90480235b015d1eb969bb78adec0e78714140787b091efaf
Writing manifest to image destination
Storing signatures
[starting container c2b8f4c5cef80875ea6e4987bbb81e21cdceaf3894a382c5f1178a1fc2c0c337: rootlessport listen tcp 0.0.0.0:20001: bind: address already in use]
[starting container c2b8f4c5cef80875ea6e4987bbb81e21cdceaf3894a382c5f1178a1fc2c0c337: rootlessport listen tcp 0.0.0.0:20001: bind: address already in use starting container 5c71090911cb59380f68223bf1ef033b5edbff0543d53b5c4a736e3c86465810: a dependency of container 5c71090911cb59380f68223bf1ef033b5edbff0543d53b5c4a736e3c86465810 failed to start: container state improper]
Pod:
f673cb2681aa98a0f976f50f4acb6e08f89342f698004654d45bda5742270f84
Container:
5c71090911cb59380f68223bf1ef033b5edbff0543d53b5c4a736e3c86465810

starting container c2b8f4c5cef80875ea6e4987bbb81e21cdceaf3894a382c5f1178a1fc2c0c337: rootlessport listen tcp 0.0.0.0:20001: bind: address already in use
starting container 5c71090911cb59380f68223bf1ef033b5edbff0543d53b5c4a736e3c86465810: a dependency of container 5c71090911cb59380f68223bf1ef033b5edbff0543d53b5c4a736e3c86465810 failed to start: container state improper

Error: failed to start 2 containers
```

## Expected behavior
The dev session started in the second tab should listen on different ports not being used. So it looks like there is an issue with the way we detect if a port is used or not ([`util.IsPortFree(int)`](https://github.com/redhat-developer/odo/blob/main/pkg/util/util.go#L786-L796)).

Initially, I suspected the [kernel issue](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/7VPNMC77YC3SI5LFYKUA4B5MTFPLTLVB/) we already hit a few weeks ago. But running https://go.dev/play/p/ijrIFcB3ayo?v=goprev works as expected. And this is reportedly fixed in recent kernels. So this is not a kernel issue to me.

rm3l:
**What type of PR is this:**
/kind bug
/area odo-on-podman

Due to regression on 6.1+ kernels [1], binding on 127.0.0.1:$port after some process listens on 0.0.0.0:$port does pass, which is not expected. Until this issue is fixed [2], a possible workaround is to try to bind on 0.0.0.0:$port. This works correctly on Podman, and also on Kubernetes (because we do port-forwarding on Kubernetes on 127.0.0.1:$port).

The unit test added should allow us to detect similar issues faster in the future, should it happen again.

[1] https://lore.kernel.org/stable/e21bf153-80b0-9ec0-15ba-e04a4ad42c34@redhat.com/T/
[2] https://lore.kernel.org/netdev/20230312031904.4674-3-kuniyu@amazon.com/T/

Fixes #6612 

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

See the repro steps provided in #6612

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -782,12 +782,11 @@ func SafeGetBool(b *bool) bool {
 
 // IsPortFree checks if the port on localhost is free to use
 func IsPortFree(port int) bool {
-	address := fmt.Sprintf("localhost:%d", port)
+	address := fmt.Sprintf("0.0.0.0:%d", port)
 	listener, err := net.Listen("tcp", address)
 	if err != nil {
 		return false
 	}
-	_ = listener.Addr().(*net.TCPAddr).Port
 	err = listener.Close()
 	return err == nil
 }
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -2735,3 +2735,129 @@ bar
 		})
 	}
 }
+
+func TestIsPortFree(t *testing.T) {
+	type serverCloser interface {
+		Close()
+	}
+	type args struct {
+		port         int
+		portProvider func() (int, serverCloser, error)
+	}
+	type test struct {
+		name string
+		args args
+		want bool
+	}
+	tests := []test{
+		{
+			name: "negative port should return an error, handles as false",
+			args: args{port: -10},
+			want: false,
+		},
+		{
+			name: "0 should always be free",
+			args: args{port: 0},
+			want: true,
+		},
+		{
+			name: "random port bound on 127.0.0.1",
+			args: args{
+				portProvider: func() (int, serverCloser, error) {
+					s := httptest.NewServer(nil)
+					_, p, err := net.SplitHostPort(strings.TrimPrefix(s.URL, "http://"))
+					if err != nil {
+						return 0, s, err
+					}
+					port, err := strconv.Atoi(p)
+					if err != nil {
+						return 0, s, err
+					}
+					return port, s, nil
+				},
+			},
+			want: false,
+		},
+		{
+			name: "random port bound on 127.0.0.1 and checking 0 as input",
+			args: args{
+				portProvider: func() (int, serverCloser, error) {
+					s := httptest.NewServer(nil)
+					return 0, s, nil
+				},
+			},
+			want: true,
+		},
+		{
+			name: "random port bound on 0.0.0.0 and checking 0 as input",
+			args: args{
+				portProvider: func() (int, serverCloser, error) {
+					// Intentionally not using httptest.Server, which listens to 127.0.0.1
+					l, err := net.Listen("tcp", "0.0.0.0:0")
+					if err != nil {
+						return 0, nil, err
+					}
+					s := &httptest.Server{
+						Listener: l,
+						Config:   &http.Server{},
+					}
+					s.Start()
+
+					return 0, s, nil
+				},
+			},
+			want: true,
+		},
+		{
+			name: "random port bound on 0.0.0.0",
+			args: args{
+				portProvider: func() (int, serverCloser, error) {
+					// Intentionally not using httptest.Server, which listens to 127.0.0.1
+					l, err := net.Listen("tcp", "0.0.0.0:0")
+					if err != nil {
+						return 0, nil, err
+					}
+					s := &httptest.Server{
+						Listener: l,
+						Config:   &http.Server{},
+					}
+					s.Start()
+
+					_, p, err := net.SplitHostPort(strings.TrimPrefix(s.URL, "http://"))
+					if err != nil {
+						return 0, s, err
+					}
+					port, err := strconv.Atoi(p)
+					if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/util/util.go
```diff
@@ -782,12 +782,11 @@ func SafeGetBool(b *bool) bool {
 
 // IsPortFree checks if the port on localhost is free to use
 func IsPortFree(port int) bool {
-	address := fmt.Sprintf("localhost:%d", port)
+	address := fmt.Sprintf("0.0.0.0:%d", port)
 	listener, err := net.Listen("tcp", address)
 	if err != nil {
 		return false
 	}
-	_ = listener.Addr().(*net.TCPAddr).Port
```
valaparthvi, line 790: What did this do?

pkg/util/util.go
```diff
@@ -782,12 +782,11 @@ func SafeGetBool(b *bool) bool {
 
 // IsPortFree checks if the port on localhost is free to use
 func IsPortFree(port int) bool {
-	address := fmt.Sprintf("localhost:%d", port)
+	address := fmt.Sprintf("0.0.0.0:%d", port)
```
valaparthvi, line 785: Should this change be reverted once the regression has been fixed?  If so, can you add a todo about it?

pkg/util/util.go
rm3l: It can be used to determine the port value on which we are listening (by casting `listener.Addr()` into `*net.TCPAddr`). But the return value is not used at all. And it is useless to me since we are already passing the input `port`. It would have been useful for example if we were letting the kernel bind to a random port and we wanted to get that random port value.

pkg/util/util.go
rm3l: To me, this is still valid even after the regression is fixed in the kernel. Podman listens by default on `0.0.0.0`, so it makes sense to check on all addresses. As for Kubernetes, port-forwarding currently binds to  `127.0.0.1` by default, but it makes sense to me to still check on all addresses.
When https://github.com/redhat-developer/odo/issues/6479 is implemented, we would need to properly support the address on which to bind and pass the right address to this function.

feloy (approved):
Thanks for this fix

PR #6678, opened 2023-03-23
Speedup odo registry --details

### LINKED ISSUE #6401: Speedup `odo registry --details`
## User Story
- As an odo user
- I don't want to wait 30+ seconds for the output of `odo registry --details`

(in general, any registry with `--details` can take longer than without it)

## Acceptance Criteria
- [ ] using `--details` flag command should not need to download devfiles and get all required information just from the devfile registry index. 

/kind user-story

/blocked
on https://github.com/devfile/api/issues/996

feloy:
**What type of PR is this:**

/kind feature

- `odo registry --details -o json` continues to make new calls to get Devfile content for all stacks of registries (to not break plugin compatibility). **This command will be deprecated**, as `--details` flag was not intended to be used without `--devfile` flag.
- `odo registry --details --devfile "name" -o json` continues to make a new call to get the "name" Devfile content (to not break plugin compatibility).
- `odo registry --details` do not make anymore new calls to get Devfile content for all stacks of registries. **This command will be deprecated**, as `--details` flag was not intended to be used without `--devfile` flag.
- `odo registry --details --devfile "name"` do not make anymore a new call to get the "name" Devfile content.

Fixes #6401 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/docs/website/docs/command-reference/json-output.md
+++ b/docs/website/docs/command-reference/json-output.md
@@ -370,7 +370,7 @@ odo registry -o json
 ]
 ```
 
-Using the `--details` flag, you will also get information about the Devfile:
+Using the `--details` flag with `--devfile <name>`, you will also get information about the Devfile:
 
 ```shell
 odo registry --devfile java-springboot --details -o json
--- a/docs/website/docs/command-reference/json-output.md
+++ b/docs/website/docs/command-reference/json-output.md
@@ -351,7 +351,14 @@ odo registry -o json
         "schemaVersion": "2.1.0",
         "starterProjects": [
           "flask-example"
-        ]
+        ],
+        "commandGroups": {
+            "build": true,
+            "debug": true,
+            "deploy": false,
+            "run": true,
+            "test": false
+        }
       },
       {
         "version": "3.0.0",
@@ -359,7 +366,14 @@ odo registry -o json
         "schemaVersion": "2.2.0",
         "starterProjects": [
           "flask-example"
-        ]
+        ],
+        "commandGroups": {
+            "build": true,
+            "debug": true,
+            "deploy": false,
+            "run": true,
+            "test": false
+        }
       }
     ],
     "starterProjects": [
@@ -400,7 +414,14 @@ odo registry --devfile java-springboot --details -o json
         "schemaVersion": "2.1.0",
         "starterProjects": [
           "springbootproject"
-        ]
+        ],
+        "commandGroups": {
+            "build": true,
+            "debug": true,
+            "deploy": false,
+            "run": true,
+            "test": false
+        }
       },
       {
         "version": "2.0.0",
@@ -408,7 +429,14 @@ odo registry --devfile java-springboot --details -o json
         "schemaVersion": "2.2.0",
         "starterProjects": [
           "springbootproject"
-        ]
+        ],
+        "commandGroups": {
+            "build": true,
+            "debug": true,
+            "deploy": true,
+            "run": true,
+            "test": false
+        }
       }
     ],
     "starterProjects": [
--- a/pkg/api/registry.go
+++ b/pkg/api/registry.go
@@ -36,9 +36,9 @@ type DevfileStack struct {
 }
 
 type DevfileStackVersion struct {
-	Version         string   `json:"version,omitempty"`
-	IsDefault       bool     `json:"isDefault"`
-	SchemaVersion   string   `json:"schemaVersion,omitempty"`
-	StarterProjects []string `json:"starterProjects"`
-	CommandGroups   map[schema.CommandGroupKind]bool
+	Version         string                           `json:"version,omitempty"`
+	IsDefault       bool                             `json:"isDefault"`
+	SchemaVersion   string                           `json:"schemaVersion,omitempty"`
+	StarterProjects []string                         `json:"starterProjects"`
+	CommandGroups   map[schema.CommandGroupKind]bool `json:"commandGroups"`
 }
--- a/pkg/api/registry.go
+++ b/pkg/api/registry.go
@@ -1,5 +1,9 @@
 package api
 
+import (
+	schema "github.com/devfile/registry-support/index/generator/schema"
+)
+
 // Registry is the main struct of devfile registry
 type Registry struct {
 	Name   string `json:"name"`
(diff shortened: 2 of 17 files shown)


valaparthvi (changes requested):
Changes lgtm, and the speed is visible. One small change is to remove reference to `devfileData` from [L132](https://github.com/redhat-developer/odo/pull/6678/files#diff-76ab00d8197655b21bc58d22d4ab8fd3d46a44a00fc14fc4607e3afba0f35bc7R132).

rm3l (changes requested):
/hold

> A side-effet is that the stack content is not available anymore in the JSON output of the command. This removal will break the IDE plugins.

Indeed, I think we'll need to check with the IDE plugins or find a way not to break the output structure..

After a quick [search](https://github.com/search?q=repo%3Aredhat-developer%2Fvscode-openshift-tools+devfileData&type=code) into the source code, the VSCode plugin seems to rely on the `.devfileData.devfile` field from the `odo registry --details` output. 
See:
- https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/registriesView.ts#L141-L151
- https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/odo/command.ts#L285-L294

feloy:
> /hold
> 
> > A side-effet is that the stack content is not available anymore in the JSON output of the command. This removal will break the IDE plugins.
> 
> Indeed, I think we'll need to check with the IDE plugins or find a way not to break the output structure..
> 
> After a quick [search](https://github.com/search?q=repo%3Aredhat-developer%2Fvscode-openshift-tools+devfileData&type=code) into the source code, the VSCode plugin seems to rely on the `.devfileData.devfile` field from the `odo registry --details` output. See:
> 
> * https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/registriesView.ts#L141-L151
> * https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/odo/command.ts#L285-L294

Yes, right. These lines particularly are using the `supportedOdoFeatures` which are removed by the current patch.

https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/openshift/component.ts#L108-L110

Also the devfileData.devfile.metadata.name used in many places, for example 
https://github.com/redhat-developer/vscode-openshift-tools/blob/main/src/openshift/component.ts#L386

feloy:
I've made changes according to what has been discussed during cabal. See https://github.com/redhat-developer/odo/pull/6678#issue-1637723214 for more details.

pkg/api/registry.go
```diff
@@ -36,4 +40,5 @@ type DevfileStackVersion struct {
 	IsDefault       bool     `json:"isDefault"`
 	SchemaVersion   string   `json:"schemaVersion,omitempty"`
 	StarterProjects []string `json:"starterProjects"`
+	CommandGroups   map[schema.CommandGroupKind]bool
```
rm3l, line 43: To follow the field naming convention in the JSON output:

```suggestion
	CommandGroups   map[schema.CommandGroupKind]bool `json:"commandGroups,omitempty"`
```

pkg/odo/cli/registry/registry.go
```diff
@@ -257,3 +265,12 @@ func boolToYesNo(b bool) string {
...
 	return "N"
 }
+
+func getVersion(stack api.DevfileStack, v string) (api.DevfileStackVersion, error) {
+	for _, version := range stack.Versions {
+		if version.Version == v {
+			return version, nil
+		}
+	}
+	return api.DevfileStackVersion{}, errors.New("version not found")
```
rm3l, line 275: To help understand the error:

```suggestion
	return api.DevfileStackVersion{}, fmt.Errorf("version %q not found in Devfile stack for %q", v, stack.Name)
```

rm3l (changes requested):

rm3l (commented):
LGTM, but maybe you can also add a note to the `odo registry` command reference page, just to mention that `--details` is intended to be used with `--devfile`? Currently, it reads "`--details` to display details about the Devfile stacks", which is not clear enough..

https://odo.dev/docs/command-reference/registry

feloy:

Non related flaky test
```
[FAILED] [266.696 seconds]
odo dev command tests port-forwarding for the component when devfile has single endpoint when running odo dev when modifying memoryLimit for container in Devfile [BeforeEach] should react on the Devfile modification
  [BeforeEach] /go/odo_1/tests/integration/cmd_dev_test.go:800
  [It] /go/odo_1/tests/integration/cmd_dev_test.go:820
```

rm3l (approved):

feloy:
Flaky test

docs/website/docs/command-reference/json-output.md
```diff
@@ -370,7 +370,7 @@ odo registry -o json
 ]
 ```
 
-Using the `--details` flag, you will also get information about the Devfile:
+Using the `--details` flag with `--devfile <name>`, you will also get information about the Devfile:
 
 ```shell
 odo registry --devfile java-springboot --details -o json
```
rm3l, line 376: Can you also update the sample output below to include the new `versions.commandGroups` fields?

rm3l: Whoops, I think we'll also need the same thing for the sample output of `odo registry -o json` above.

rm3l (approved):

feloy:
> Sounds good to me to remove the hold on this PR, now that we are pretty confident in not breaking the IDE plugins, no? What do you think? I added a reminder to document the deprecated `odo registry --details` (without `--devfile`) in the [v3.9.0](https://github.com/redhat-developer/odo/issues/6643) release notes too.

Yes, the JSON output is not changed by this patch, it should not break the IDE plugins

rm3l:
> > Sounds good to me to remove the hold on this PR, now that we are pretty confident in not breaking the IDE plugins, no? What do you think? I added a reminder to document the deprecated `odo registry --details` (without `--devfile`) in the [v3.9.0](https://github.com/redhat-developer/odo/issues/6643) release notes too.
> 
> Yes, the JSON output is not changed by this patch, it should not break the IDE plugins

Thanks for confirming.

/hold cancel

feloy:

Flaky e2e test

```
 [FAILED] [212.271 seconds]
/go/odo_1/tests/e2escenarios/e2e_test.go:337
```

PR #6704, opened 2023-03-31
Allow users to define custom port-mappings for port forwarding [Kubernetes]

### LINKED ISSUE #6479: Allow users define custom port-mappings for port forwarding
/kind user-story

## User Story
As an odo user, I want to be able to define a mapping for port-forwarding So that I can have predictable port numbers; this is useful when developing multi-component applications.

example:

`cd frontend`
`odo dev`

`cd ../backend`
`odo dev`

configure the `frontend` to use port 20002 (`odo dev` for backed was started second)

I stop everything, next day I want to continue working but this time I start `odo dev` session for `backend` before `frontend`. Now I have to keep in mind to reconfigure the frontend to use a different port.

A slightly better experience would be to have 

`cd frontend`
`odo dev`

`cd ../backend`
`odo dev --port-forward 8888:8000`

Now no matter what order I run commands, I will get backed on localhost:8888.

An even better experience would be for odo to remember custom port mappings so next time I can just do `odo dev`. This is something that we can look into as second stage. 

## Acceptance Criteria 
- [x] Users should be able to define custom port mapping using flags provided to `odo dev` command using `--port-forward` flag. Flag accepts argument in `<localPort>:<containerPort>`
- [x] `port-forward` can be defined as multiple times (for multiple ports)
- [x] if multiple containers have the same port and the command should fail and guide the user to specify container name `--port-forward <localPort>:<containerName>:<containerPort>`
- [ ] `--address <ipaddress>` flag should allow specifying on which local address the port will be listening 
- [x] Validate if the containerPort is a valid endpoint(?)
- [ ] Handle `dedicatedPod` (is it handled currently?)
- [x] Cannot use `--random-port` and `--port-forward` together.
- [x] Validate that a local port is not defined more than once

allowed formats for `--port-forward`

`8080:80` - local port 8080 listening only on `localhost`, container port 80
`8080:mycontainer:80` - local port 8080 listening only on `localhost`, container port 80 on container called mycontainer

/kind user-story

valaparthvi:
**What type of PR is this:**
/kind feature

Fixes part of #6479 

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test [WIP]

- [ ] Documentation [TBD]

1. `odo dev --port-forward <localPort>:<containerPort> --port-forward <localPort>:<containerName>:<containerPort>`

### DIFF
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -46,9 +46,10 @@ type DevOptions struct {
 	clientset *clientset.Clientset
 
 	// Variables
-	ignorePaths []string
-	out         io.Writer
-	errOut      io.Writer
+	ignorePaths    []string
+	out            io.Writer
+	errOut         io.Writer
+	forwardedPorts []api.ForwardedPort
 
 	// ctx is used to communicate with WatchAndPush to stop watching and start cleaning up
 	ctx context.Context
@@ -223,15 +224,16 @@ func (o *DevOptions) Run(ctx context.Context) (err error) {
 		o.out,
 		o.errOut,
 		dev.StartOptions{
-			IgnorePaths:      o.ignorePaths,
-			Debug:            o.debugFlag,
-			BuildCommand:     o.buildCommandFlag,
-			RunCommand:       o.runCommandFlag,
-			RandomPorts:      o.randomPortsFlag,
-			WatchFiles:       !o.noWatchFlag,
-			IgnoreLocalhost:  o.ignoreLocalhostFlag,
-			ForwardLocalhost: o.forwardLocalhostFlag,
-			Variables:        variables,
+			IgnorePaths:          o.ignorePaths,
+			Debug:                o.debugFlag,
+			BuildCommand:         o.buildCommandFlag,
+			RunCommand:           o.runCommandFlag,
+			RandomPorts:          o.randomPortsFlag,
+			WatchFiles:           !o.noWatchFlag,
+			IgnoreLocalhost:      o.ignoreLocalhostFlag,
+			ForwardLocalhost:     o.forwardLocalhostFlag,
+			Variables:            variables,
+			CustomForwardedPorts: o.forwardedPorts,
 		},
 	)
 }
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -160,7 +160,7 @@ func (o *DevOptions) Validate(ctx context.Context) error {
 		if len(errStrings) != 0 {
 			log.Error("There are following issues with values provided by --port-forward flag:")
 			for _, errStr := range errStrings {
-				fmt.Printf("\t- %s\n", errStr)
+				fmt.Fprintf(log.GetStderr(), "\t- %s\n", errStr)
 			}
 		}
 		return err
--- a/pkg/odo/cli/dev/dev.go
+++ b/pkg/odo/cli/dev/dev.go
@@ -4,8 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/redhat-developer/odo/pkg/api"
 	"io"
+	"k8s.io/klog"
 	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 	ktemplates "k8s.io/kubectl/pkg/util/templates"
@@ -58,6 +64,9 @@ type DevOptions struct {
 	runCommandFlag       string
 	ignoreLocalhostFlag  bool
 	forwardLocalhostFlag bool
+	// TODO: Add doc
+	portForwardFlag []string
+	addressFlag     string
 }
 
 var _ genericclioptions.Runnable = (*DevOptions)(nil)
@@ -127,6 +136,26 @@ func (o *DevOptions) Validate(ctx context.Context) error {
 		}
 		scontext.SetPlatform(ctx, o.clientset.PodmanClient)
 	}
+
+	if o.randomPortsFlag && o.portForwardFlag != nil {
+		return errors.New("--random-ports and --port-forward cannot be used together")
+	}
+
+	if o.portForwardFlag != nil {
+		containerEndpointMapping, err := libdevfile.GetDevfileContainerEndpointMapping(devfileObj, true)
+		if err != nil {
+			return fmt.Errorf("failed to obtain container endpoints to validate --port-forward ports; cause:%w", err)
+		}
+
+		forwardedPorts, err := parsePortForwardFlag(o.portForwardFlag)
+		if err != nil {
+			return err
+		}
+		o.forwardedPorts = forwardedPorts
+
+		return validatePortForwardFlagData(forwardedPorts, containerEndpointMapping)
+	}
+
 	return nil
 }
 
@@ -246,6 +275,12 @@ It forwards endpoints with any exposure values ('public', 'internal' or 'none') 
 		"Whether to ignore errors related to port-forwarding apps listening on the container loopback interface. Applicable only if platform is podman.")
 	devCmd.Flags().BoolVar(&o.forwardLocalhostFlag, "forward-localhost", false,
 		"Whether to enable port-forwarding if app is listening on the container loopback interface. Applicable only if platform is podman.")
+	// TODO: Fix usage
+	//   kubectl port-forward TYPE/NAME [options] [LOCAL_PORT:]REMOTE_PORT [...[LOCAL_PORT_N:]REMOTE_PORT_N]
+	devCmd.Flags().StringArrayVar(&o.portForwardFlag, "port-forward", nil,
+		"Define custom port forwards")
+	devCmd.Flags().StringVar(&o.addressFlag, "address", "",
+		"Which local address the port will be listening")
 
 	clientset.Add(devCmd,
 		clientset.BINDING,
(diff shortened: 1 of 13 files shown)


tests/integration/cmd_dev_test.go
```diff
@@ -712,6 +712,216 @@ ComponentSettings:
...
+				helper.Cmd("odo", "set", "project", commonVar.Project).ShouldPass()
+				helper.CopyExample(filepath.Join("source", "devfiles", "nodejs", "project-with-multiple-endpoints"), commonVar.Context)
+				helper.Cmd("odo", "init", "--name", cmpName, "--devfile-path", helper.GetExamplePath("source", "devfiles", "nodejs", "devfile-with-multiple-endpoints.yaml")).ShouldPass()
+			})
+
+			When("running odo dev", func() {
+				var devSession helper.DevSession
+				var ports map[string]string
+				BeforeEach(func() {
+					opts := []string{fmt.Sprintf("--port-forward=%s:%s", LocalPort1, ContainerPort1), fmt.Sprintf("--port-forward=%s:%s", LocalPort2, ContainerPort2), fmt.Sprintf("--port-forward=%s:%s", LocalPort3, ContainerPort3)}
```
feloy, line 847: It could be interesting to test with 3 endpoints on Devfile and only 2 custom forwarded ports, to be sure the other one is forwarded from 20001.

feloy: Yes, you are right. Thanks

valaparthvi:
> > windows-integration-test/Windows-test failure is due to same random port returned twice, although I am not sure why. The [seed](https://github.com/redhat-developer/odo/pull/6704/files#diff-307f9917f1ad315bfc5bd954f3bfedc51f87105e838f3d89d7533218db174867R57) used should've made sure we get a unique port, but it did not here.
> > ```shell
> >   Running odo.exe with args [odo dev --debug --port-forward=35507:3000 --port-forward=35507:5858] and odo env: [ODO_LOG_LEVEL=4 ODO_TRACKING_CONSENT=no]
> > 
> > [odo]  X  There are following issues with values provided by --port-forward flag:
> >   [odo]  X  values for --port-forward flag are invalid
> > 
> > 
> >   [FAILED] Timed out after 360.000s.
> >   Expected
> >       <string>: 	- local port 35507 is used more than once, please use unique local ports
> >       
> >   to contain substring
> >       <string>: [Ctrl+c] - Exit
> > ```
> 
> This failed again.
> 
> ```shell
>   Running odo.exe with args [odo dev --port-forward=53120:3000 --port-forward=53120:4567 --port-forward=53120:7890] and odo env: [ODO_LOG_LEVEL=4 ODO_TRACKING_CONSENT=no]
> 
>   [odo] I0404 05:25:14.793715   12260 dev.go:371] container port 7890 matches http-7890 endpoints of container 
>   [odo]  X  There are following issues with values provided by --port-forward flag:
>   [odo] 	- local port 53120 is used more than once, please use unique local ports
>   [odo] 	- local port 53120 is used more than once, please use unique local ports
>   [odo]  X  values for --port-forward flag are invalid
> 
> 
>   [FAILED] session exited, but substring not found
>   Expected
>       <string>: 
>   to contain substring
>       <string>: [Ctrl+c] - Exit
> ```
> 
> My guess is we're using unix time as a seed, but since this is a windows machine, it always takes a similar port into account. I have changed to seed to use a normal `NanoSeconds` to see if it makes any difference.

Changing the seed to `NanoSeconds` did not make any difference.

tests/helper/helper_http.go
```diff
@@ -47,3 +50,27 @@ func HttpWaitForWithStatus(url string, match string, maxRetry int, interval int,
 	fmt.Fprintf(GinkgoWriter, "Last output from %s: %s\n", url, string(body))
 	Fail(fmt.Sprintf("Failed after %d retries. Content in %s doesn't include '%s'.", maxRetry, url, match))
 }
+
+// GetRandomFreePort returns a random free port(string) between 1024 and 65535, or within a given portRange if provided
+// WARN: If length of portRange is anything other than 2 and first index is greater-than-or-equal-to second index, it will use the default range.
+func GetRandomFreePort(portRange ...int) string {
+	// rand.Seed() has been deprecated from Go 1.20, and it has been advised to use the following method instead
+	randomizer := rand.New(rand.NewSource(int64(time.Now().Nanosecond()))) // #nosec  // cannot use crypto/rand library here
```
feloy, line 58: Adding a seed is useful to not have the same series of number every time. But in this case, we don't care about the reproducibility of this series, we just want to have different ports between subsequent calls to rand.Intn. So, I would try bby using rand.Intn without any previous call to Seed or equivalent.

Another thing, Seed is normally called only once, to start a new sequence of pseudo-random numbers. Here, you are re-initializing it every time you pick a new number, so you are getting every time the first number of the series.

feloy: To illustrate what I mean, it could be better to use a global counter starting from 30000 across all the tests and it should work. The series would always be 30000, 30001, 30002, ..., but you are sure you never get the same port during the series.

valaparthvi: > To illustrate what I mean, it could be better to use a global counter starting from 30000 across all the tests and it should work. The series would always be 30000, 30001, 30002, ..., but you are sure you never get the same port during the series.

Do you mean I should simply try to use the way `odo dev`? Start from a counter(30000), set a range(40000), and check for the next free port?

If that is true, then it won't work in case we're calling the function back to back without calling `odo dev`, it will return the same port number because it is still free, and will be until `odo dev` starts. This was the reason I wanted to use random port numbers.

>Adding a seed is useful to not have the same series of number every time. But in this case, we don't care about the reproducibility of this series, we just want to have different ports between subsequent calls to rand.Intn. So, I would try bby using rand.Intn without any previous call to Seed or equivalent.

You're right, we can do without the Seed.

feloy: > Do you mean I should simply try to use the way odo dev? Start from a counter(30000), set a range(40000), and check for the next free port?

No, I mean a global value that would increment itself each time you get its value, in an atomic way.

valaparthvi: > > Do you mean I should simply try to use the way odo dev? Start from a counter(30000), set a range(40000), and check for the next free port?
> 
> No, I mean a global value that would increment itself each time you get its value, in an atomic way.

Okay, I'll try that.

feloy: This was to illustrate my meaning about seed. If it works by removing the seed, I think it's ok

tests/helper/helper_http.go
```diff
@@ -51,24 +51,14 @@ func HttpWaitForWithStatus(url string, match string, maxRetry int, interval int,
...
-	return strconv.Itoa(freePort)
+var startPort int64 = 30000
 
+// GetRandomFreePort increases the counter of global variable startPort, checks if the counter value is a free port, and returns.
+// If the port is not free, it calls itself again. Checking the port is an additional validation to ensure we don't run into random failures because of port unavailability.
+func GetRandomFreePort() string {
+	atomic.AddInt64(&startPort, 1)
+	if !util.IsPortFree(int(startPort)) {
+		return GetRandomFreePort()
+	}
```
feloy, line 62: I'm not sure IsPortFree is atomic here. You could just expect the port is free.

The random ports are in the range of the Ephemeral ports, in my machine it is:

```
$ sysctl net.ipv4.ip_local_port_range
net.ipv4.ip_local_port_range = 32768	60999
```

So we can consider ports below 32768 will be free.

pkg/odo/cli/dev/dev.go
```diff
@@ -127,6 +139,33 @@ func (o *DevOptions) Validate(ctx context.Context) error {
 		}
 		scontext.SetPlatform(ctx, o.clientset.PodmanClient)
 	}
+
+	if o.randomPortsFlag && o.portForwardFlag != nil {
+		return errors.New("--random-ports and --port-forward cannot be used together")
+	}
+
+	if o.portForwardFlag != nil {
```
rm3l, line 147: I know that Podman support for this will come after this PR, but until this is done, I think we should either error out if `platform` is `podman` and `--port-forward` is set. Or at least display a warning.
Otherwise, the output is a bit confusing IMO.

```
$ odo dev --debug --port-forward 8089:3000 --port-forward 5005:runtime:5858 --platform podman
  __
 /  \__     Developing using the "my-nodejs" Devfile
 \__/  \    Platform: podman
 /  \__/    odo version: v3.8.0
 \__/

 ✓  Deploying pod [16s]
 ✓  Building your application in container (command: install) [3s]
 •  Executing the application (command: debug)  ...
 -  Forwarding from 127.0.0.1:20001 -> 3000
 -  Forwarding from 127.0.0.1:20002 -> 5858

↪ Dev mode
 Status:
 Watching for changes in the current directory /home/asoro/work/tmp/6479-allow-users-define-custom-port-mappings-for-port-forwarding

 Keyboard Commands:

```

pkg/odo/cli/dev/dev.go
```diff
@@ -268,3 +310,105 @@ It forwards endpoints with any exposure values ('public', 'internal' or 'none')
...
+// parsePortForwardFlag parses custom port forwarding configuration; acceptable patterns: <localPort>:<containerPort>, <localPort>:<containerName>:<containerPort>
+func parsePortForwardFlag(portForwardFlag []string) (forwardedPorts []api.ForwardedPort, err error) {
+	// acceptable examples: 8000:runtime_123:8080, 8000:9000, 8000:runtime:8080, 20001:20000
+	// unacceptable examples: :8000, 80000:
+	pattern := `^(\d{1,5})(:\w*)?:(\d{1,5})$`
+	portReg := regexp.MustCompile(pattern)
+	const largestPortValue = 65535
+	for _, portData := range portForwardFlag {
+		if !portReg.MatchString(portData) {
+			return nil, errors.New("ports are not defined properly, acceptable formats are: <localPort>:<containerPort>, <localPort>:<containerName>:<containerPort>")
```
rm3l, line 394: Suggestion to update the error message. In the example below, I intentionally passed a high port respecting the format, but the error message does not help identify the cause of the error:

```
$ odo dev --debug --port-forward 8089:3000 --port-forward 100001:runtime:5858
 ✗  ports are not defined properly, acceptable formats are: <localPort>:<containerPort>, <localPort>:<containerName>:<containerPort>
```

```suggestion
			return nil, errors.New("ports are not defined properly, acceptable formats are: <localPort>:<containerPort>, <localPort>:<containerName>:<containerPort>", where ports must be numbers in the range [1, 65535])
```

pkg/odo/cli/dev/dev.go
```diff
@@ -268,3 +310,105 @@ It forwards endpoints with any exposure values ('public', 'internal' or 'none')
...
+		case 2:
+			portF.LocalPort, _ = strconv.Atoi(portDataArr[0])
+			portF.ContainerPort, _ = strconv.Atoi(portDataArr[1])
+		case 3:
+			portF.LocalPort, _ = strconv.Atoi(portDataArr[0])
+			portF.ContainerName = portDataArr[1]
+			portF.ContainerPort, _ = strconv.Atoi(portDataArr[2])
+		}
+		if !(portF.LocalPort > 0 && portF.LocalPort <= largestPortValue) || !(portF.ContainerPort > 0 && portF.ContainerPort <= largestPortValue) {
+			return nil, fmt.Errorf("%s is invalid; port number must not be greater than 0 and less than %d", portData, largestPortValue)
```
rm3l, line 409: Port numbers must actually be > 0 and <= 65535:

```suggestion
			return nil, fmt.Errorf("%s is invalid; port numbers must be greater than 0 and less than %d", portData, largestPortValue)
```

pkg/portForward/kubeportforward/portForward_test.go
```diff
@@ -0,0 +1,62 @@
...
+			},
+			wantPortPairs: map[string][]string{
+				"runtime": {"8080:8000", "20001:9000"},
+				"tools":   {"5000:5000"},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if gotPortPairs := getCustomPortPairs(tt.args.definedPorts, tt.args.ceMapping); !reflect.DeepEqual(gotPortPairs, tt.wantPortPairs) {
```
rm3l, line 57: Can you use `cmp.Diff` for comparing the results instead? It will help analyze the output in case of failures (https://github.com/redhat-developer/odo/issues/5932).

pkg/odo/cli/dev/dev.go
valaparthvi: I am planning to implement this for podman before our next release, but you're right, it is better to add a warning until them.

valaparthvi:
```sh
  [FAILED] in [AfterEach] - /go/odo_1/tests/helper/helper_dev.go:211 @ 04/06/23 11:22:15.369
  Deleting project: cmd-dev-test1485ipw
  Running oc with args [oc delete project cmd-dev-test1485ipw --wait=false] and odo env: []
  [oc] project.project.openshift.io "cmd-dev-test1485ipw" deleted
  Deleting dir: /tmp/2482955252
  Deleting dir: /tmp/1006440071
  << Timeline

  [FAILED] Timed out after 180.001s.

  Expected process to exit.  It did not.
  In [AfterEach] at: /go/odo_1/tests/helper/helper_dev.go:211 @ 04/06/23 11:22:15.369

  [FAIL] odo dev command tests when Starting a PostgreSQL service when creating local files and dir and running odo dev - with metadata.name [AfterEach] should correctly propagate changes to the container
```

OpenShift Test failure.

rm3l (approved):

feloy:

Flaky e2e tests

```
 [FAILED] [113.660 seconds]
odo dev command tests port-forwarding for the component with custom port mapping when devfile has single endpoint when running odo dev [BeforeEach] should expose the endpoint on localhost
  [BeforeEach] /go/odo_1/tests/integration/cmd_dev_test.go:746
  [It] /go/odo_1/tests/integration/cmd_dev_test.go:761
```

```
[FAILED] [211.978 seconds]
/go/odo_1/tests/e2escenarios/e2e_test.go:337
```

PR #6709, opened 2023-04-03
Add DEVFILE_REGISTRY env var for periodic tests

### LINKED ISSUE #6702: Use In-cluster Devfile registries for integration tests
/kind user-story

## User Story
As an odo developer, I want to use stable Devfile registries (but not the default Production one) for the integration tests So that any changes to Staging Registry (what we're currently using) do not break or block us from continuing odo development.

## Acceptance Criteria 
- [ ] It should setup an in-cluster Devfile registry. See https://odo.dev/blog/deploying-and-using-in-cluster-devfile-registry#deploying-a-devfile-registry-in-the-cluster
- [ ] Tests should be configured such that they use the staging registry by default, because it is still great that `odo` can detect and report such issues to the Devfile Services team.
- [ ] But if the tests fail due to the Staging Registry, it should fallback to using in-cluster registry. This can be done quite easily now that `odo` can automatically detect and use registries declared in the cluster. See https://odo.dev/blog/deploying-and-using-in-cluster-devfile-registry#declaring-the-devfile-registry

## Links 
- Related Epic (mandatory): 

/kind user-story

feloy:
**What type of PR is this:**

/kind tests

Fixes #6702

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/openshiftci-periodic-tests.sh
+++ b/scripts/openshiftci-periodic-tests.sh
@@ -26,6 +26,8 @@ oc login -u developer -p password@123 --insecure-skip-tls-verify
 # Check login user name for debugging purpose
 oc whoami
 
+source ./scripts/openshiftci-config.sh
+
 # Integration tests
 make test-integration || error=true
 
--- a/scripts/openshiftci-config.sh
+++ b/scripts/openshiftci-config.sh
@@ -1,0 +1,1 @@
+export DEVFILE_REGISTRY=https://devfile-registry-ci-devfile-registry.odo-test-kubernetes-clust-49529fc6e6a4a9fe7ebba9a3db5b55c4-0000.eu-de.containers.appdomain.cloud/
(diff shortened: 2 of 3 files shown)


valaparthvi (changes requested):
Thank you for fixing this. 
There's one more change: You'll also need to fix scripts/openshiftci-Nightly-SBO-tests.sh.

scripts/openshiftci-periodic-tests.sh
```diff
@@ -26,6 +26,8 @@ oc login -u developer -p password@123 --insecure-skip-tls-verify
 # Check login user name for debugging purpose
 oc whoami
 
+source ./scripts/openshiftci-config.sh
```
rm3l, line 29: This assumes that this script is always run using `./scripts/openshiftci-periodic-tests.sh` and will fail if we run it from elsewhere, no? Is it always the case?
How about determining the location of `openshiftci-periodic-tests.sh` and using it instead?

rm3l: Okay, with the different `make` invocations in these scripts, it should be okay this way. Never mind then.

feloy: As there are some `make ...` just after, I assume the script is always started from the root directory

rm3l (approved):

feloy:
Unrelated tests

PR #6713, opened 2023-04-04
Use devstats.PID.json as devstate

### LINKED ISSUE #6494: odo dev on podman and cluster cannot run simultaneously
/kind bug

**Operating System:**
Fedora 35
**Output of `odo version`:**
odo v3.5.0 (da9a9efb8f)
1. `odo init --devfile go --name my-go-app --starter go-starter`
2. `ODO_EXPERIMENTAL_MODE=true odo dev --platform podman` wait for it to finish
3. In another tab `odo dev`.

## Actual behavior
Once `odo dev` starts, both the dev sessions continue to sync because of the changes in .odo/devstate.json and never reach a stable state.

## Expected behavior
Both sessions should be able to run and reach a stable state.

I am not sure if this is a valid use case and if something like this is an ideal workflow but it still a nuisance as an odo developer when I want to build and test new features for both podman and cluster.

## Acceptance criteria

- [x] PID is added to the content of the devstate file
- [x] Platform is added to the content of the devstate file
- [x] Forwarded Ports are added to the content of the devstate file
- [x] each `odo dev` session creates its own devstate.$PID.json file
- [x] the first `odo dev` session also creates a `devstate.json` file, for compatibility.
- [x] each `odo dev` session should check if `devstate.json` file exists and references a PID not existing, it deletes it; this helps in cleaning any file lying around when the previous process was kill(9)-ed. Same for all `odo` commands reading this file (`odo describe component`, ...)
- [x] `odo describe component` displays forwarded ports from all devstate.$PID.json files
- [x] `odo describe component --platform <platform>` displays forwarded ports from  devstate.$PID.json file with platform=&lt;platform>
- [x] `odo dev` can run only once per platform, exit if a devstate.$PID.json file with the platform already exists.

feloy:
**What type of PR is this:**

/kind feature

State of `odo dev` if now saved in a file `.odo/devstate.${PID}.json`, to be able to run multiple instances and state for each of them.

For compatibility, the file `.odo/devstate.json` is also used for the first instance of `odo dev`.

One drawback is that if an instance of `odo dev` is killed, the `devstate.json` file won't be overwritten. One possibility would be to delete it when running `odo delete component` and the process with registered PID into devstate.json is not existing.

Fixes #6494 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/state/errors.go
+++ b/pkg/state/errors.go
@@ -4,12 +4,13 @@ import "fmt"
 
 type ErrAlreadyRunningOnPlatform struct {
 	platform string
+	pid      int
 }
 
-func NewErrAlreadyRunningOnPlatform(platform string) ErrAlreadyRunningOnPlatform {
+func NewErrAlreadyRunningOnPlatform(platform string, pid int) ErrAlreadyRunningOnPlatform {
 	return ErrAlreadyRunningOnPlatform{platform: platform}
 }
 
 func (e ErrAlreadyRunningOnPlatform) Error() string {
-	return fmt.Sprintf("a session is already running on platform %q", e.platform)
+	return fmt.Sprintf("a session with PID %d is already running on platform %q", e.pid, e.platform)
 }
--- a/pkg/state/errors.go
+++ b/pkg/state/errors.go
@@ -8,7 +8,10 @@ type ErrAlreadyRunningOnPlatform struct {
 }
 
 func NewErrAlreadyRunningOnPlatform(platform string, pid int) ErrAlreadyRunningOnPlatform {
-	return ErrAlreadyRunningOnPlatform{platform: platform}
+	return ErrAlreadyRunningOnPlatform{
+		platform: platform,
+		pid:      pid,
+	}
 }
 
 func (e ErrAlreadyRunningOnPlatform) Error() string {
--- a/pkg/state/errors.go
+++ b/pkg/state/errors.go
@@ -1,0 +1,15 @@
+package state
+
+import "fmt"
+
+type ErrAlreadyRunningOnPlatform struct {
+	platform string
+}
+
+func NewErrAlreadyRunningOnPlatform(platform string) ErrAlreadyRunningOnPlatform {
+	return ErrAlreadyRunningOnPlatform{platform: platform}
+}
+
+func (e ErrAlreadyRunningOnPlatform) Error() string {
+	return fmt.Sprintf("a session is already running on platform %q", e.platform)
+}
--- a/pkg/state/process_unix.go
+++ b/pkg/state/process_unix.go
@@ -1,0 +1,38 @@
+//go:build aix || darwin || dragonfly || freebsd || linux || netbsd || openbsd || solaris || zos
+// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris zos
+
+package state
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func pidExists(pid int) (bool, error) {
+	if pid <= 0 {
+		return false, fmt.Errorf("invalid pid %v", pid)
+	}
+	proc, err := os.FindProcess(pid)
+	if err != nil {
+		return false, nil
+	}
+	err = proc.Signal(syscall.Signal(0))
+	if err == nil {
+		return true, nil
+	}
+	if err.Error() == "os: process already finished" {
+		return false, nil
+	}
+	errno, ok := err.(syscall.Errno)
+	if !ok {
+		return false, err
+	}
+	switch errno {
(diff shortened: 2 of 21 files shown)


pkg/state/state.go
```diff
@@ -61,10 +182,75 @@ func (o *State) save() error {
...
 
-func (o *State) read() error {
+func (o *State) isFreeOrOwnedBy(pid int) (bool, error) {
 	jsonContent, err := o.fs.ReadFile(_filepath)
 	if err != nil {
-		return err
+		if errors.Is(err, os.ErrNotExist) {
+			return true, nil
+		}
+		// File not found, it is free
```
rm3l, line 191: This comment applies to the `if errors.Is(err, os.ErrNotExist) {` above...

pkg/state/state.go
```diff
@@ -24,30 +30,145 @@ func NewStateClient(fs filesystem.Filesystem) *State {
...
+	if err != nil {
+		return err
+	}
+	// .odo directory is supposed to exist, don't create it
+	dir := filepath.Dir(getFilename(pid))
+	err = os.MkdirAll(dir, 0750)
+	if err != nil {
+		return err
+	}
+	return o.fs.WriteFile(getFilename(pid), jsonContent, 0644)
```
rm3l, line 118: Could we extract a separate method from L108 through L118? I saw this repeated, and the only difference was the target state file path we are writing to.

pkg/state/process_unix.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func pidExists(pid int) (bool, error) {
+	if pid <= 0 {
+		return false, fmt.Errorf("invalid pid %v", pid)
+	}
+	proc, err := os.FindProcess(pid)
```
rm3l, line 16: Wondering why we have a different implementation on Unix? `os.FindProcess` does not already check and return the right info about the process?

rm3l: Ah, got it. From the [doc](https://pkg.go.dev/os#FindProcess), it looks like this is expected:

> On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.

feloy: Yes, this is the reason

feloy:
Thanks @rm3l for your review

pkg/state/errors.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "fmt"
+
+type ErrAlreadyRunningOnPlatform struct {
+	platform string
+	pid      int
+}
+
+func NewErrAlreadyRunningOnPlatform(platform string, pid int) ErrAlreadyRunningOnPlatform {
+	return ErrAlreadyRunningOnPlatform{platform: platform}
```
rm3l, line 11: Missing `pid`:

```suggestion
	return ErrAlreadyRunningOnPlatform{
		platform: platform, 
		pid: pid,
	}
```

rm3l (changes requested):
One last suggestion, LGTM otherwise.

feloy:
Investigating this error:

On Windows, it seems that the second process detects correctly that a first process exists.
But at the time of stopping the first process at the end of the test, an error occurs.

```
[FAILED] [66.184 seconds]
odo dev command tests when a component with multiple endpoints is run [It] should fail running a second session on the same platform
C:/Users/Administrator.ANSIBLE-TEST-VS/3707/tests/integration/cmd_dev_test.go:3081

[...]

 [odo]   __
  [odo]  /  \__     Developing using the "ijoreq" Devfile
  [odo]  \__/  \    Namespace: cmd-dev-test3081fac
  [odo]  /  \__/    odo version: v3.9.0
  [odo]  \__/
  [odo] 
  [odo] - Running on the cluster in Dev mode
  [odo]  X  unable to save state file: a session with PID 11436 is already running on platform "cluster"

[...]

[FAILED] Unexpected error:
      <*fmt.wrapError | 0xc001fd6260>: {
          msg: "calling GenerateConsoleCtrlEvent: The parameter is incorrect.",
          err: <syscall.Errno>0x57,
      }
      calling GenerateConsoleCtrlEvent: The parameter is incorrect.
  occurred
  In [It] at: C:/Users/Administrator.ANSIBLE-TEST-VS/3707/tests/helper/helper_dev.go:195 @ 04/17/23 03:58:31.703
```

rm3l:
@feloy Can you rebase your branch onto `main`? I was about to re-lgtm this PR, but noticed it has a conflict.

rm3l (approved):

feloy:
Flaky tests

PR #6737, opened 2023-04-13
Fix flaky unit Test_getCompleteCustomPortPairs

### LINKED ISSUE #6741: Flaky unit test
/kind bug

```
make test

[...]
--- FAIL: Test_getCompleteCustomPortPairs (0.00s)
    --- FAIL: Test_getCompleteCustomPortPairs/ports_are_provided_with_container_name (0.00s)
        portForward_test.go:77: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{
              	"runtime": {
              		"8080:8000",
            - 		"20002:9000",
            + 		"20001:9000",
              	},
            - 	"tools": {"20001:5000"},
            + 	"tools": {"20002:5000"},
              }
```

I can see in the implementation of `getCompleteCustomPortPairs` that maps are used. To have reproducibility, we need to loop over **ordered** keys of the map, or the map will be looped over in a random order, and the results of the tests will be randomly ordered too.

valaparthvi:
**What type of PR is this:**
/area testing
/kind flake

This fixes flaky test.

Fixes part of #6479
Fixes #6741

**PR acceptance criteria:**

- [x] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/portForward/kubeportforward/portForward_test.go
+++ b/pkg/portForward/kubeportforward/portForward_test.go
@@ -23,7 +23,6 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			args: args{
 				definedPorts: []api.ForwardedPort{
 					{ContainerName: "runtime", LocalPort: 8080, ContainerPort: 8000},
-					{ContainerName: "tools", LocalPort: 5000, ContainerPort: 5000},
 				},
 				ceMapping: map[string][]v1alpha2.Endpoint{
 					"runtime": {{TargetPort: 8000}, {TargetPort: 9000}},
@@ -32,7 +31,7 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			},
 			wantPortPairs: map[string][]string{
 				"runtime": {"8080:8000", "20001:9000"},
-				"tools":   {"5000:5000"},
+				"tools":   {"20002:5000"},
 			},
 		},
 		{
--- a/pkg/portForward/kubeportforward/portForward_test.go
+++ b/pkg/portForward/kubeportforward/portForward_test.go
@@ -23,6 +23,7 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			args: args{
 				definedPorts: []api.ForwardedPort{
 					{ContainerName: "runtime", LocalPort: 8080, ContainerPort: 8000},
+					{ContainerName: "tools", LocalPort: 5000, ContainerPort: 5000},
 				},
 				ceMapping: map[string][]v1alpha2.Endpoint{
 					"runtime": {{TargetPort: 8000}, {TargetPort: 9000}},
@@ -31,7 +32,7 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			},
 			wantPortPairs: map[string][]string{
 				"runtime": {"8080:8000", "20001:9000"},
-				"tools":   {"20002:5000"},
+				"tools":   {"5000:5000"},
 			},
 		},
 		{
--- a/pkg/portForward/kubeportforward/portForward_test.go
+++ b/pkg/portForward/kubeportforward/portForward_test.go
@@ -1,21 +1,29 @@
 package kubeportforward
 
 import (
+	"fmt"
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
-	"github.com/google/go-cmp/cmp"
 	"github.com/redhat-developer/odo/pkg/api"
+	"strconv"
+	"strings"
 	"testing"
 )
 
 func Test_getCompleteCustomPortPairs(t *testing.T) {
+	const (
+		acceptablePortRange = "[20001-30001]"
+		acceptableMinPort   = 20001
+		acceptableMaxPort   = 30001
+	)
 	type args struct {
 		definedPorts []api.ForwardedPort
 		ceMapping    map[string][]v1alpha2.Endpoint
 	}
 	tests := []struct {
-		name          string
-		args          args
-		wantPortPairs map[string][]string
+		name string
+		args args
+		// wantPortPairs format: {containerName: {containerPort: localPort}}
+		wantPortPairs map[string]map[string]string
 	}{
 		// TODO: Add test cases.
 		{
@@ -23,16 +31,15 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			args: args{
 				definedPorts: []api.ForwardedPort{
 					{ContainerName: "runtime", LocalPort: 8080, ContainerPort: 8000},
-					{ContainerName: "tools", LocalPort: 5000, ContainerPort: 5000},
 				},
 				ceMapping: map[string][]v1alpha2.Endpoint{
 					"runtime": {{TargetPort: 8000}, {TargetPort: 9000}},
 					"tools":   {{TargetPort: 5000}},
 				},
 			},
-			wantPortPairs: map[string][]string{
-				"runtime": {"8080:8000", "20001:9000"},
-				"tools":   {"5000:5000"},
+			wantPortPairs: map[string]map[string]string{
+				"runtime": {"8000": "8080", "9000": acceptablePortRange},
+				"tools":   {"5000": acceptablePortRange},
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


pkg/portForward/kubeportforward/portForward_test.go
```diff
@@ -23,6 +23,7 @@ func Test_getCompleteCustomPortPairs(t *testing.T) {
 			args: args{
 				definedPorts: []api.ForwardedPort{
 					{ContainerName: "runtime", LocalPort: 8080, ContainerPort: 8000},
+					{ContainerName: "tools", LocalPort: 5000, ContainerPort: 5000},
```
valaparthvi, line 26: Previously, there were 2 non-customized ports, and a random value was assigned to 5000 and 9000 from time to time, sometimes 20001 would be assigned to 5000 and 20002 to 9000 and vice versa.

Adding 5000 to the definedPorts will help avoid this flake while also maintaining the test integrity.

valaparthvi:
OC Failure:
```sh

  [FAILED] Timed out after 180.000s.
  Expected process to exit.  It did not.
  In [AfterEach] at: /go/odo_1/tests/helper/helper_dev.go:211 @ 04/14/23 05:41:57.43

Summarizing 1 Failure:
  [FAIL] odo dev command tests when Starting a PostgreSQL service when creating local files and dir and running odo dev - without metadata.name [AfterEach] when deleting local files and dir and waiting for sync should not list deleted dir and file in container
  /go/odo_1/tests/helper/helper_dev.go:211
```

Windows Failure:
```sh
  [FAILED] No future change is possible.  Bailing out early after 1.332s.
  Running oc.exe with args [oc get pods -n interactive-add-binding-test577amu -o jsonpath='{range .items[*]}{.metadata.name}'] and odo env: []
  Expected
      <int>: 1
  to match exit code:
      <int>: 0

  [FAILED] No future change is possible.  Bailing out early after 1.192s.
  Running oc.exe with args [oc exec nxbmst-app-7b69cff74b-9m7wl --namespace cmd-dev-test1738lfs -- ls -lai /projects/webapp] and odo env: []
  Expected
      <int>: 1
  to match exit code:
      <int>: 0

  [FAILED] No future change is possible.  Bailing out early after 21.166s.
  Running oc.exe with args [oc create configmap config-map-for-cleanup --from-literal type=testing --from-literal team=odo -n cmd-delete-test625lsx] and odo env: []
  Expected
      <int>: 1
  to match exit code:
      <int>: 0
  In [BeforeEach] at: C:/Users/Administrator.ANSIBLE-TEST-VS/3699/tests/helper/helper_cmd_wrapper.go:101 @ 04/14/23 00:19:16.65


Summarizing 3 Failures:
  [FAIL] odo delete command tests [BeforeEach] when a component is bootstrapped when the component is running in both DEV and DEPLOY mode and dev mode is killed when the component is deleted while having access to the devfile.yaml should delete the appropriate resources
  C:/Users/Administrator.ANSIBLE-TEST-VS/3699/tests/helper/helper_cmd_wrapper.go:101
  [FAIL] odo add binding interactive command tests [BeforeEach] when running a deployment when binding to a service in a different namespace should error out if service is not found in the namespace selected
  C:/Users/Administrator.ANSIBLE-TEST-VS/3699/tests/helper/helper_generic.go:58
  [FAIL] odo dev command tests when multiple projects are present - with metadata.name [It] should sync to the correct dir in container
  C:/Users/Administrator.ANSIBLE-TEST-VS/3699/tests/helper/helper_cmd_wrapper.go:101

Ran 437 of 805 Specs in 1189.646 seconds
FAIL! -- 434 Passed | 3 Failed | 0 Pending | 368 Skipped

File: pkg/portForward/kubeportforward/portForward_test.go
Comment (rm3l) [inline pkg/portForward/kubeportforward/portForward_test.go:1]: I noticed that all the test cases here are relying on opening actual ports and expecting exact ports. But if I already have some other apps listening on those ports, then the test would fail.
I ran into this issue because I had unrelated `odo dev` sessions running and taking ports 20001 through 20003, and the test is now failing:

```
--- FAIL: Test_getCompleteCustomPortPairs (0.00s)                                                                                                                                                                   
    --- FAIL: Test_getCompleteCustomPortPairs/ports_are_provided_with_container_name (0.00s)                                                                                                                        
        portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{                                                                                                            
                "runtime": {                                                                                                                                                                                        
                        "8080:8000",                                                                                                                                                                                
            -           "20004:9000",                                                                                                                                                                               
            +           "20001:9000",                                                                                                                                                                               
                },                                                                                                                                                                                                  
                "tools": {"5000:5000"},                                                                                                                                                                             
              }                                                                                                                                                                                                     
    --- FAIL: Test_getCompleteCustomPortPairs/ports_are_provided_without_container_name (0.00s)                                                                                                                     
        portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{                                                                                                            
                "runtime": {                                                                                                                                                                                        
                        "8080:8000",                                                                                                                                                                                
            -           "20004:9000",                                                                                                                                                                               
            +           "20001:9000",
                },
                "tools": {"5000:5000"},
              }
    --- FAIL: Test_getCompleteCustomPortPairs/local_ports_in_range_[20001-30001]_are_provided_as_custom_forward_ports (0.00s)
        portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{
                "runtime": {"20001:8000", "20002:9000"},
                "tools": {
                        "5000:5000",
            -           "20004:8080",
            +           "20003:8080",
                },
              }
```

Ideally, the unit test should not have to open any actual port, but I understand this is exactly what we are trying to test here. So I'd suggest relaxing the checks, by just checking that the port pairs we are getting in response are just within the range we expect. What do you think?

rm3l (changes requested):

pkg/portForward/kubeportforward/portForward_test.go
feloy: > I noticed that all the test cases here are relying on opening actual ports and expecting exact ports. But if I already have some other apps listening on those ports, then the test would fail. I ran into this issue because I had unrelated `odo dev` sessions running and taking ports 20001 through 20003, and the test is now failing:
> 
> ```
> --- FAIL: Test_getCompleteCustomPortPairs (0.00s)                                                                                                                                                                   
>     --- FAIL: Test_getCompleteCustomPortPairs/ports_are_provided_with_container_name (0.00s)                                                                                                                        
>         portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{                                                                                                            
>                 "runtime": {                                                                                                                                                                                        
>                         "8080:8000",                                                                                                                                                                                
>             -           "20004:9000",                                                                                                                                                                               
>             +           "20001:9000",                                                                                                                                                                               
>                 },                                                                                                                                                                                                  
>                 "tools": {"5000:5000"},                                                                                                                                                                             
>               }                                                                                                                                                                                                     
>     --- FAIL: Test_getCompleteCustomPortPairs/ports_are_provided_without_container_name (0.00s)                                                                                                                     
>         portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{                                                                                                            
>                 "runtime": {                                                                                                                                                                                        
>                         "8080:8000",                                                                                                                                                                                
>             -           "20004:9000",                                                                                                                                                                               
>             +           "20001:9000",
>                 },
>                 "tools": {"5000:5000"},
>               }
>     --- FAIL: Test_getCompleteCustomPortPairs/local_ports_in_range_[20001-30001]_are_provided_as_custom_forward_ports (0.00s)
>         portForward_test.go:78: getCompleteCustomPortPairs() (got vs want) diff =   map[string][]string{
>                 "runtime": {"20001:8000", "20002:9000"},
>                 "tools": {
>                         "5000:5000",
>             -           "20004:8080",
>             +           "20003:8080",
>                 },
>               }
> ```
> 
> Ideally, the unit test should not have to open any actual port, but I understand this is exactly what we are trying to test here. So I'd suggest relaxing the checks, by just checking that the port pairs we are getting in response are just within the range we expect. What do you think?

In the other hand, this solution will hide the problem raised in https://github.com/redhat-developer/odo/issues/6741: I think we want to have the same local ports every time we work on a Devfile with several ports to forward. Without fixing 
#6741, local ports will differ from one run to another.

valaparthvi: > I think we want to have the same local ports every time we work on a Devfile with several ports to forward.

We do, but that is why we are defining custom port mapping. For ports that we do not define a custom mapping, there will always be a chance that you get a different port number every time, so perhaps it does make sense that we only check the port range instead of the exact value. Although we still want to ensure it does not assign the same port number more than once, I can add an additional check for that.

Regardless of how we test, I agree that we can still loop over the containers in an orderly manner.

rm3l: > In the other hand, this solution will hide the problem raised in https://github.com/redhat-developer/odo/issues/6741: I think we want to have the same local ports every time we work on a Devfile with several ports to forward. Without fixing
https://github.com/redhat-developer/odo/issues/6741, local ports will differ from one run to another.
>
I agree with ensuring we have the same local ports. As Parthvi said, that should be okay, I think, for custom ports.
Overall, my main concern is more on how those unit tests are depending on the ports opened on the system.. So I'd need to make sure I don't have any other unrelated application listening on the local ports expected..

valaparthvi: FYI, I reverted the entire commit (https://github.com/redhat-developer/odo/pull/6737/commits/fb31c33af01308ce14df8a6d1d2768b20c6117ca), which means current files do not contain the relaxed checks. I shall revert it again once we have a consensus.

rm3l: Whoops, just noticed this PR got merged already. Alright, let's see if this becomes annoying on a daily basis..

PR #6774, opened 2023-04-26
PODMAN_USE_NAMESPACES to disable using namespaces on podman integration tests for podman v4.5

### LINKED ISSUE #6773: Unable to run podman integration tests on MacOS
/kind bug
/area testing

It is not possible to run podman integration tests on MacOS due to the custom `containers.conf` file our tests are using.

We use a very simple `containers.conf` file with just the `namespace`. MacOS requires the network configuration which is missing from the file since podman on MacOS runs inside a VM.

Reproduction:
```sh
$ cat containers.conf
[engine]
namespace="somethign"

$ CONTAINERS_CONF=containers.conf podman pod ls
Cannot connect to Podman. Please verify your connection to the Linux system using `podman system connection list`, or try `podman machine init` and `podman machine start` to manage a new Linux VM
Error: unable to connect to Podman socket: Get "http://d/v4.3.1/libpod/_ping": dial unix ///var/folders/_r/3xp61_nn34sgqn4bkt8cddrc0000gn/T/podman-run--1/podman/podman.sock: connect: no such file or directory
```

We could use an existing conf file and append the `namespace` to it to fix this issue.

feloy:
**What type of PR is this:**

This PR introduces a new env var for integration tests: PODMAN_USE_NAMESPACES

This is useful when running podman tests on a local system with podman v4.5.0, as namespace feature has been removed from podman v4.5

(tests on github actions are not impacted because running on podman v3)

When using this option, you need to take care of not running tests in parallel, or some tests could fail.

Ex:
```
PODMAN_USE_NAMESPACES=false make test-integration-podman
```

/area testing

Fixes #6773 
6773 was not the original issue, but this PR still fixes it./=

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


tests/helper/helper_podman.go
```diff
@@ -14,6 +14,9 @@ import (
 )
 
 func GenerateAndSetContainersConf(dir string) {
+	if os.Getenv("PODMAN_USE_NAMESPACES") == "false" {
```
valaparthvi, line 17: I believe this would fix #6773 as well.

valaparthvi, line 17:
```suggestion
    useNamespace, _ := strconv.ParseBool(os.Getenv("PODMAN_USE_NAMESPACES"))
    
	if !useNamespace {
```

tests/helper/helper_podman.go
```diff
@@ -7,13 +7,30 @@ import (
...
 	. "github.com/onsi/gomega"
 	"github.com/redhat-developer/odo/pkg/podman"
 )
 
+func getBooleanValueFromEnvVar(envvar string, defaultValue bool) bool {
+	strVal := os.Getenv("PODMAN_USE_NAMESPACES")
+	if strVal == "" {
+		return defaultValue
+	}
+	boolValue, err := strconv.ParseBool(strVal)
```
valaparthvi, line 22:
```suggestion
	boolValue, err := strconv.ParseBool(strVal)
```

If the string is empty, it will return the same error as if the string were invalid.
```sh
false strconv.ParseBool: parsing "": invalid syntax
false strconv.ParseBool: parsing "somthing": invalid syntax
```

valaparthvi (approved):

PR #6801, opened 2023-05-09
Updating the oudated release script

### LINKED ISSUE #6730: Fix outdated release verification script
/kind user-story
/area release-eng
/area testing

## User Story
As an odo developer or QE, I want to test `odo` binaries candidates for release, So that I can be confident to mark the corresponding Errata as `REL_PREP`.

The [release-bit-verification](https://github.com/redhat-developer/odo/blob/main/scripts/release-bit-verification.sh) script is mentioned in the QE release checklist, as part of the QE process to test `odo` release candidates attached as RPMs to the Errata. But it is currently outdated; for example, it is running Make targets that no longer exist.

From my quick analysis of the script, it currently:
- checks the checksums from the RPM
- checks that the version reported by `odo version` is the same as the one recorded in the passed RPM file
-  checks out the `odo` repository at the specified version tag and runs integration and E2E tests using the appropriate `odo` binary (extracted from the passed RPM file)

## Acceptance Criteria
- [ ] Make sure the script errors out immediately if there is an error
- [ ] The release verification script should only run E2E tests
- [ ] To avoid manual (and tedious) work, it should ideally work on all supported Operating Systems (Linux, Windows, Mac)

## Links 
- Related Epic (mandatory): 

/kind user-story

ritudes:
**What type of PR is this:**

Fixes #6730 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/scripts/release-bit-verification.sh
+++ b/scripts/release-bit-verification.sh
@@ -21,7 +21,7 @@ shout() {
 
 
 
-# Create a Temp directory
+# Checking for no or invaild arguments
 if [ "$#" -lt 1 ]
 then
   echo "No arguments supplied"
@@ -32,11 +32,12 @@ if [ ! -f ${1} ];
 then
     echo "Please enter a valid filepath";
     exit 1
-else
-    WORKING_DIR=$(mktemp -d)
-    shout "WORKING_DIR=$WORKING_DIR"
-    export REPO_URL=${REPO_URL:-"https://github.com/redhat-developer/odo.git"}
 fi
+#Creating an Temp directory    
+WORKING_DIR=$(mktemp -d)
+shout "WORKING_DIR=$WORKING_DIR"
+export REPO_URL=${REPO_URL:-"https://github.com/redhat-developer/odo.git"}
+
 
 # Extract from rpm file
 rpm2cpio ${1} | cpio -idmvD $WORKING_DIR
--- a/scripts/release-bit-verification.sh
+++ b/scripts/release-bit-verification.sh
@@ -18,7 +18,7 @@ shout() {
 # Check SHASUM for all the binary files and there should be no difference
 
 #For erroring out in case of error
-set -e
+set -eo pipefail
 
 # Create a Temp directory
 WORKING_DIR=$(mktemp -d)
--- a/scripts/release-bit-verification.sh
+++ b/scripts/release-bit-verification.sh
@@ -22,9 +22,21 @@ shout() {
 
 
 # Create a Temp directory
+if [ "$#" -lt 1 ]
+then
+  echo "No arguments supplied"
+  exit 1
+fi
+
 WORKING_DIR=$(mktemp -d)
-shout "WORKING_DIR=$WORKING_DIR"
-export REPO_URL=${REPO_URL:-"https://github.com/redhat-developer/odo.git"}
+if [ -f ${1} ]; 
+then
+    shout "WORKING_DIR=$WORKING_DIR"
+    export REPO_URL=${REPO_URL:-"https://github.com/redhat-developer/odo.git"}
+else
+    echo "Please enter a valid filepath";
+    exit 1
+fi
 
 # Extract from rpm file
 rpm2cpio ${1} | cpio -idmvD $WORKING_DIR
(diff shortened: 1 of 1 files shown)


scripts/release-bit-verification.sh
```diff
@@ -17,6 +17,9 @@ shout() {
 }
 # Check SHASUM for all the binary files and there should be no difference
 
+#For erroring out in case of error
+set -e
```
rm3l, line 21: Can you move this as the first instruction (i.e., right after the shebang)? Just so it handles all subsequent commands below it.

rm3l: Also, because the script contains some commands that are piped, like `rpm2cpio ${1} | cpio -idmvD $WORKING_DIR` in L29, I would suggest setting `-o pipefail` too. Otherwise, the exit code of a piped command would be that of the last command in the pipe.

```suggestion
set -eo pipefail
```

rm3l: Also, because the script contains some commands that are piped, like `rpm2cpio ${1} | cpio -idmvD $WORKING_DIR` in L29, I would suggest setting `-o pipefail` too. Otherwise, the exit code of a piped command would be that of the last command in the pipe.

```suggestion
set -eo pipefail
```

scripts/release-bit-verification.sh
```diff
@@ -67,10 +70,9 @@ fi
...
 
 #Run tests
-make test-integration-devfile
-make test-integration
-make test-e2e-all
-make test-cmd-project
+make test-e2e
 
 # Cleanup
 rm -rf /tmp/odo /tmp/usr
```
rm3l, line 76: I think we can just remove `"$WORKING_DIR"`:
```suggestion
rm -rf "$WORKING_DIR"
```

scripts/release-bit-verification.sh
```diff
@@ -17,6 +17,9 @@ shout() {
 }
 # Check SHASUM for all the binary files and there should be no difference
 
+#For erroring out in case of error
+set -e
+
 # Create a Temp directory
```
rm3l, line 23: Regarding error management, can you check that the argument (`$1`) passed to this script is a valid file that exists even before creating the temporary working dir. I think we should not even go further if the path to the RPM file passed is a not valid.

rm3l (changes requested):

scripts/release-bit-verification.sh
ritudes: Yeah I'll check that

ritudes: I checked this and it looks like its working fine and without any errors

scripts/release-bit-verification.sh
```diff
@@ -17,6 +17,9 @@ shout() {
 }
 # Check SHASUM for all the binary files and there should be no difference
 
+#For erroring out in case of error
+set -eo pipefail
```
rm3l, line 21: Can you move this as the first instruction (i.e., right after the shebang)? Just so it handles all subsequent commands below it.

rm3l (changes requested):

scripts/release-bit-verification.sh
ritudes: Ok I'm looking into that

scripts/release-bit-verification.sh
```diff
@@ -11,16 +11,32 @@
...
+if [ "$#" -lt 1 ]
+then
+  echo "No arguments supplied"
+  exit 1
+fi
+
 WORKING_DIR=$(mktemp -d)
-shout "WORKING_DIR=$WORKING_DIR"
-export REPO_URL=${REPO_URL:-"https://github.com/redhat-developer/odo.git"}
+if [ -f ${1} ];
```
rm3l, line 32: This would still create the temporary directory if the file is not valid.
I'd suggest inverting the condition: exit if the file does not exist, without even creating the temporary directory.

Once all checks are done on the first argument, we can proceed with creating the temporary directory.

rm3l (approved):
Thanks for the changes.

PR #6840, opened 2023-05-24
Document how to fix file permission errors while using odo with GKE/AKS/EKS

### LINKED ISSUE #6821: Document troubleshooting storage permission issues with GKE/AKS/EKS
/area documentation

/kind feature

Elaborate https://github.com/redhat-developer/odo/issues/6706#issuecomment-1549323644

valaparthvi:
**What type of PR is this:**
/area documentation

Fixes #6821 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

https://deploy-preview-6840--odo-docusaurus-preview.netlify.app/docs/user-guides/advanced/using-odo-with-other-clusters


docs/website/docs/user-guides/advanced/using-odo-with-other-clusters.md
```diff
@@ -0,0 +1,322 @@
...
+This guide will discuss some workarounds that can be used to fix these issues.
+
+### Using Ephemeral Volumes
+This is the simplest way to overcome this issue. There are 2 parts to this solution:
+1. Set `odo` preference `Ephemeral` to _true_.
+
+   ```shell
+    odo preference set Ephemeral true -f
+    ```
+2. If the Devfile contains a `volume` component, then set its `ephemeral` property to `true`.
```
rm3l, line 138: Can you also add an example of how this would look like in a Devfile with a `volume` component?

docs/website/docs/user-guides/advanced/using-odo-with-other-clusters.md
```diff
@@ -0,0 +1,322 @@
...
+   The above configuration will use the [`emptyDir` Ephemeral volumes](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) instead of creating Persistent Volumes to mount the source files; it also ensures the current user can read/write to the directories.
```
rm3l, line 139: I guess this sentence applies to both 1. and 2., and could be separated from the 2nd bullet point.
I would also suggest linking to this [other page](https://odo.dev/docs/development/architecture/how-odo-works#volumes) in case readers want to learn more about how `odo` works with Volumes?

```suggestion

   The above configuration will use the [`emptyDir` Ephemeral volumes](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) instead of creating Persistent Volumes to mount the source files; it also ensures the current user can read/write to the directories.
```

docs/website/docs/user-guides/advanced/using-odo-with-other-clusters.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Define custom location to mount volumes and project source
+All the Devfiles provided by the [Devfile Registry](https://registry.devfile.io) use container image set a non-root user, which is why we see these permission related issues; we usually do not see this for a root user.
```
rm3l, line 142: Are we sure all the images used in the default registry are configured this way?

```suggestion
Most of the Devfiles provided by the [Devfile Registry](https://registry.devfile.io) use container images with a non-root user, which is why we see these permission related issues; we usually do not see this for a root user.
```

docs/website/docs/user-guides/advanced/using-odo-with-other-clusters.md
```diff
@@ -0,0 +1,322 @@
...
+
+When a user does not have access to all the locations in a system, we need to provide a way to use a location where the user can read/write. There are 2 parts to this solution as well:
+1. Set `sourceMapping` of the Devfile `container` component to a location where the user has read/write access.
+2. If a Devfile `volume` component is defined, then ensure that the `.volumeMounts.path` of the Devfile container component where the volume will be mount is in a location where the user has read/write access.
```
rm3l, line 146:
```suggestion
2. If a Devfile `volume` component is defined, then ensure that the `.volumeMounts.path` of the Devfile container component where the volume will be mounted is in a location where the user has read/write access.
```

rm3l (changes requested):

docs/website/docs/user-guides/advanced/using-odo-with-other-clusters.md
```diff
@@ -0,0 +1,322 @@
...
+- git:
+    remotes:
+      origin: https://github.com/odo-devfiles/springboot-ex.git
+  name: springbootproject
+```
+</details>
+
+
+But this solution may not always be feasible, especially while dealing with large filesystems.
+>Be cautious with the use of fsGroup. The changing of group ownership of an entire volume can cause pod startup delays for slow and/or large filesystems. It can also be detrimental to other processes that share the same volume if their processes do not have access permissions to the new GID. For this reason, some providers for shared file systems such as NFS do not implement this functionality. These settings also do not affect ephemeral volumes.
```
feloy, line 320: The second part of the warning (sharing the volume with other processes) is not related to how odo uses the volume, as odo is creating a dedicated volume for the component, which won't be shared with others.

For the first part of the warning, we can explain that this will have an effect only at the "restart" of the pod (when the user modifies the container's information in the Devfile). In a normal use (without modifying the container in Devfile), there won't be any effect, as the Volume is empty when it is mounted the first time, and only populated with the sync.

These warnings are valid for containers that mount volumes with persistent data (imagine a Postgres image mounting a 1Tb volume with its data), but I think we can mitigate them in the case of odo.

feloy:
LGTM. Thanks @valaparthvi for this work

@rm3l feel free to approve it when you are ok with the changes

Issue #6568, opened 2023-02-03
`odo run <devfile-command>` when `odo dev` is running


kadel:
/kind user-story

## User Story
As an odo user, I want to be able to start commands defined in the devfile when I need to So that I can run actions defined in the devfile, like, for example, running commands that test things, or nondefault build commands that do some optional extra work.

This will allow users to follow the same workflow as in IDEs and DevSpaces where the command are usually executed only upon user request.

## Acceptance Criteria 
- [ ] there should be a command (`odo run` for example) that takes one argument that is the command name and executes the command defined in devfile
- [ ] `odo run` assumes that `odo dev` is running in a separate terminal, if `odo dev` is not running (resources not present on the cluster) `odo run` errors out and tells the user to run `odo dev` first. 
- [ ] if the command is `exec` command and the container where the program needs to be started is not running, the command should be executed as a `Job`; for podman this can be running inside a pod/container.
- [ ] if the command is `exec` and the container is already running, it should be just executed inside the running container
- [ ] if command is `apply` it just applies what it needs to :-) `kubernetes`/`openshift` components get created, `container` component gets turned into `Deployment` is `dedicatedPod: true`
- [ ] if command is `composite` that it performs all action based on the above rules
- [ ] `odo run` streams logs into the terminal and waits for the command to finish.
- [ ] user can interrupt the command by pressing  control-c; cleanup the resources. 
- [ ] the exit code of `odo run` command reflect the exit code of the last devfile command executed. If the last command was `apply` and it was successful it returns 0, if failed returns 1
- [ ] Because odo dev is running, if Kubernetes/OpenShift resources are created with Apply commands, they should be deployed in "Dev" mode (to be cleanup when odo dev terminates) - make it clear it will be cleaned up.

## Links 

This will enable the implementation of  https://github.com/redhat-developer/odo/issues/6070

related to https://github.com/redhat-developer/odo/issues/6478 epic

/kind user-story
/priority medium

rm3l:
/triage needs-information
/assign @kadel 

Add more details

valaparthvi:
**[Grooming Call April 6, 2023]**
- Implement this with command-first approach.
- `odo run` running a long running command should terminate if the `odo dev` pod restarts.
- Refactor the command handlers to be more generic. Design a new architecture. [To be handled in a separate issue].
- Implement `odo run` with the new architecture.
- Interactive mode to be added at a later stage.
[...]

PR #6857, opened 2023-06-01
Add an `odo run` command to manually execute command during `odo dev`

feloy:
**What type of PR is this:**

/kind feature

Fixes partially #6568 

The points covered by this PR are:

- there should be a command (odo run for example) that takes one argument that is the command name and executes the command defined in devfile
- odo run assumes that odo dev is running in a separate terminal, if odo dev is not running (resources not present on the cluster) odo run errors out and tells the user to run odo dev first.
- if the command is exec command and the container where the program needs to be started is not running, the command should be executed as a Job; for podman this can be running inside a pod/container (this case is not possible currently: odo dev includes **all** the containers defined in the Devfile).
- if the command is exec and the container is already running, it should be just executed inside the running container
- if command is apply it just applies what it needs to :-) kubernetes/openshift components get created, container component gets turned into Deployment is dedicatedPod: true
- if command is composite that it performs all action based on the above rules

This PR does not completely covers #6568. Points not implemented/tested yet are:

- odo run streams logs into the terminal and waits for the command to finish.
- user can interrupt the command by pressing  control-c; cleanup the resources.
- the exit code of odo run command reflect the exit code of the last devfile command executed. If the last command was apply and it was successful it returns 0, if failed returns 1
- Because odo dev is running, if Kubernetes/OpenShift resources are created with Apply commands, they should be deployed in "Dev" mode (to be cleanup when odo dev terminates) - make it clear it will be cleaned up.

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/dev/common/run.go
+++ b/pkg/dev/common/run.go
@@ -30,7 +30,7 @@ func Run(
 
 	pod, err := platformClient.GetPodUsingComponentName(componentName)
 	if err != nil {
-		return fmt.Errorf("unable to get pod for component %s: %w", componentName, err)
+		return fmt.Errorf("unable to get pod for component %s: %w. Please check the command odo dev is running", componentName, err)
 	}
 
 	handler := component.NewRunHandler(
--- a/pkg/dev/common/run.go
+++ b/pkg/dev/common/run.go
@@ -1,0 +1,53 @@
+package common
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/redhat-developer/odo/pkg/component"
+	"github.com/redhat-developer/odo/pkg/configAutomount"
+	"github.com/redhat-developer/odo/pkg/devfile/image"
+	"github.com/redhat-developer/odo/pkg/exec"
+	"github.com/redhat-developer/odo/pkg/libdevfile"
+	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
+	"github.com/redhat-developer/odo/pkg/platform"
+	"github.com/redhat-developer/odo/pkg/testingutil/filesystem"
+)
+
+func Run(
+	ctx context.Context,
+	commandName string,
+	platformClient platform.Client,
+	execClient exec.Client,
+	configAutomountClient configAutomount.Client,
+	filesystem filesystem.Filesystem,
+) error {
+	var (
+		componentName = odocontext.GetComponentName(ctx)
+		devfileObj    = odocontext.GetEffectiveDevfileObj(ctx)
+		devfilePath   = odocontext.GetDevfilePath(ctx)
+	)
+
+	pod, err := platformClient.GetPodUsingComponentName(componentName)
+	if err != nil {
+		return fmt.Errorf("unable to get pod for component %s: %w", componentName, err)
+	}
+
+	handler := component.NewRunHandler(
+		ctx,
+		platformClient,
+		execClient,
+		configAutomountClient,
+		pod.Name,
+		false,
+		component.GetContainersNames(pod),
+		"Executing command in container",
+
+		filesystem,
+		image.SelectBackend(ctx),
+		*devfileObj,
+		devfilePath,
+	)
+
+	return libdevfile.ExecuteCommandByName(ctx, *devfileObj, commandName, handler, false)
+}
--- a/pkg/dev/common/run.go
+++ b/pkg/dev/common/run.go
@@ -30,7 +30,7 @@ func Run(
 
 	pod, err := platformClient.GetPodUsingComponentName(componentName)
 	if err != nil {
-		return fmt.Errorf("unable to get pod for component %s: %w. Please check the command odo dev is running", componentName, err)
+		return fmt.Errorf("unable to get pod for component %s: %w. Please check the command 'odo dev' is running", componentName, err)
 	}
 
 	handler := component.NewRunHandler(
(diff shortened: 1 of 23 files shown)


pkg/odo/cli/run/run.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if devfileObj == nil {
+		return genericclioptions.NewNoDevfileError(odocontext.GetWorkingDirectory(ctx))
+	}
+
+	commands, err := devfileObj.Data.GetCommands(common.DevfileOptions{
+		FilterByName: o.commandName,
+	})
+	if err != nil || len(commands) != 1 {
+		return errors.NewNoCommandNameInDevfileError(o.commandName)
```
rm3l, line 72: Can we separate the cases where `err != nil` and `len(commands)!= 1`? In the first case, I think it would make more sense to return `err` as it might provide the user with additional details about the error (other than just a command not found error)..

pkg/dev/podmandev/run.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/redhat-developer/odo/pkg/dev/common"
+	"k8s.io/klog"
+)
+
+func (o *DevClient) Run(
+	ctx context.Context,
+	commandName string,
+) error {
+	klog.V(4).Infof("running command %q on cluster", commandName)
```
rm3l, line 14:
```suggestion
	klog.V(4).Infof("running command %q on podman", commandName)
```

pkg/libdevfile/libdevfile.go
```diff
@@ -66,6 +66,26 @@ func ExecuteCommandByNameAndKind(ctx context.Context, devfileObj parser.DevfileO
 	return executeCommand(ctx, devfileObj, cmd, handler)
 }
 
+// ExecuteCommandByNameAndKind executes the specified command cmdName of the given kind in the Devfile.
```
rm3l, line 69:
```suggestion
// ExecuteCommandByName executes the specified command cmdName in the Devfile.
```

pkg/podman/pods.go
rm3l, line 1: One comment on the implementation of `GetRunningPodFromSelector` for Podman. 
I think we should return an error if the pod state is not "Running".. Otherwise, when running `odo run` on Podman and the pod is not completely running (for example when the image is being pulled), we try to run the command, which fails immediately:
```
$ odo run exec-11 --platform podman
 ✗  Executing command in container (command: exec-11) [37ms]
 ✗  unable to log error unable to exec command [/bin/sh -c (echo Hey there from exec-11) 1>>/proc/1/fd/1 2>>/proc/1/fd/2]: exit status 255: unable to show logs, component is not in running st
```

On cluster, since we are filtering running pods only, an appropriate error is returned if the pod is not running yet:
```
$ odo run exec-11 --platform cluster
 ✗  unable to get pod for component odo-run-devfile-command-when-odo-dev-is-running: pod not found for the selector: component=odo-run-devfile-command-when-odo-dev-is-running. Please check the command odo dev is running
```

pkg/dev/common/run.go
```diff
@@ -0,0 +1,53 @@
...
+) error {
+	var (
+		componentName = odocontext.GetComponentName(ctx)
+		devfileObj    = odocontext.GetEffectiveDevfileObj(ctx)
+		devfilePath   = odocontext.GetDevfilePath(ctx)
+	)
+
+	pod, err := platformClient.GetPodUsingComponentName(componentName)
+	if err != nil {
+		return fmt.Errorf("unable to get pod for component %s: %w. Please check the command odo dev is running", componentName, err)
```
rm3l, line 33:
```suggestion
		return fmt.Errorf("unable to get pod for component %s: %w. Please check the command 'odo dev' is running", componentName, err)
```

tests/examples/source/devfiles/nodejs/devfile-for-run.yaml
```diff
@@ -0,0 +1,75 @@
...
+        origin: "https://github.com/odo-devfiles/nodejs-ex.git"
+components:
+  - name: runtime
+    container:
+      image: registry.access.redhat.com/ubi8/nodejs-12:1-36
+      memoryLimit: 1024Mi
+      endpoints:
+        - name: "3000-tcp"
+          targetPort: 3000
+      mountSources: true
```
rm3l, line 19: Can we also add a case with one more side container component in which we want to run commands? I think this can be interesting to cover..

rm3l (changes requested):
Added a few comments..

pkg/podman/pods.go
feloy: This change will have some side effects on other calls of GetRunningPodFromSelector. I'll look at them

rm3l (approved):
Thanks for the changes - it works great so far!

feloy:
```
 [FAILED] [206.381 seconds]
C:/Users/Administrator.ANSIBLE-TEST-VS/4064/tests/e2escenarios/e2e_test.go:462
[...]
[FAILED] Expected
      <*url.Error | 0xc0001090e0>: {
          Op: "Post",
          URL: "http://127.0.0.1:62941/api/newuser",
          Err: <*errors.errorString | 0xc00008a130>{s: "EOF"},
      }
  to be nil
```

feloy:
/override SonarCloud Code Analysis
Duplicate code on mocks due to Platform interface embedded into Podman interface

PR #6915, opened 2023-06-19
Implement API endpoints

### LINKED ISSUE #6302: Implement API Endpoints
Here is my first draft of how the HTTP API could look like.

### GET /instance
Get information about the this 'odo dev' instance.

return data example
```json
{
  "componentDirectory": "/Users/user/Documents/myproject",
  "pid": 42
}
```

### DELETE /instance
Stop this 'odo dev' instance. The same as 'ctrl+c' on terminal

return data example
``` json
{
  "message": "'odo dev' instance with pid: 42 is shuting down."
}
```

### GET /component
Information about the component. This should return the same data as `odo describe component -o json`

### POST /component/command
Instruct 'odo dev' to perform the given command on the component

This expects the name of the command to be passed with the request.

request data:
```json
{
  "name": "push"
}
```

return data example

```json
{
  "message": "push was successfully executed"
}
```

It might look more complex than it needs to be, but I wanted to ensure that we can easily extend it in the future without breaking backward compatibility.

I've created OpenAPI we can use it to have automatically generated documentation, and consumers of this API can use it to create mock servers for development testing. When we agreed on the first design, we should keep the OpenAPI definition with our source code.

OpenAPI spec: https://ododev.docs.apiary.io/api-description-document
Rendered Documentation: https://ododev.docs.apiary.io/

There is also an automatically generated mock server at `https://private-anon-13f9ca22d0-ododev.apiary-mock.com`

```
curl https://private-anon-13f9ca22d0-ododev.apiary-mock.com/instance
curl -X DELETE https://private-anon-13f9ca22d0-ododev.apiary-mock.com/instance
curl -X POST -d '{"name": "push"}'  https://private-anon-13f9ca22d0-ododev.apiary-mock.com/component/command
curl https://private-anon-13f9ca22d0-ododev.apiary-mock.com/component
```

_Originally posted by @kadel in https://github.com/redhat-developer/odo/issues/5986#issuecomment-1302383935_

feloy:
**What type of PR is this:**

/kind feature

Fixes #6302 

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/apiserver-impl/api_default_service.go
+++ b/pkg/apiserver-impl/api_default_service.go
@@ -7,20 +7,34 @@ import (
 	"net/http"
 
 	openapi "github.com/redhat-developer/odo/pkg/apiserver-gen/go"
+	"github.com/redhat-developer/odo/pkg/component/describe"
+	"github.com/redhat-developer/odo/pkg/kclient"
 	odocontext "github.com/redhat-developer/odo/pkg/odo/context"
+	"github.com/redhat-developer/odo/pkg/podman"
+	"github.com/redhat-developer/odo/pkg/state"
 )
 
 // DefaultApiService is a service that implements the logic for the DefaultApiServicer
 // This service should implement the business logic for every endpoint for the DefaultApi API.
 // Include any external packages or services that will be required by this service.
 type DefaultApiService struct {
-	cancel context.CancelFunc
+	cancel       context.CancelFunc
+	kubeClient   kclient.ClientInterface
+	podmanClient podman.Client
+	stateClient  state.Client
 }
 
 // NewDefaultApiService creates a default api service
-func NewDefaultApiService(cancel context.CancelFunc) openapi.DefaultApiServicer {
+func NewDefaultApiService(cancel context.CancelFunc,
+	kubeClient kclient.ClientInterface,
+	podmanClient podman.Client,
+	stateClient state.Client,
+) openapi.DefaultApiServicer {
 	return &DefaultApiService{
-		cancel: cancel,
+		cancel:       cancel,
+		kubeClient:   kubeClient,
+		podmanClient: podmanClient,
+		stateClient:  stateClient,
 	}
 }
 
@@ -37,20 +51,18 @@ func (s *DefaultApiService) ComponentCommandPost(ctx context.Context, componentC
 
 // ComponentGet -
 func (s *DefaultApiService) ComponentGet(ctx context.Context) (openapi.ImplResponse, error) {
-	// TODO - update ComponentGet with the required logic for this service method.
-	// Add api_default_service.go to the .openapi-generator-ignore to avoid overwriting this service implementation when updating open api generation.
-
-	// TODO: Uncomment the next line to return response Response(200, ComponentGet200Response{}) or use other options such as http.Ok ...
-	// return Response(200, ComponentGet200Response{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("ComponentGet method not implemented")
+	value, _, err := describe.DescribeDevfileComponent(ctx, s.kubeClient, s.podmanClient, s.stateClient)
+	if err != nil {
+		return openapi.Response(http.StatusInternalServerError, ""), fmt.Errorf("error getting the description of the component: %w", err)
+	}
+	return openapi.Response(http.StatusOK, value), nil
 }
 
 // InstanceDelete -
 func (s *DefaultApiService) InstanceDelete(ctx context.Context) (openapi.ImplResponse, error) {
 	s.cancel()
 	return openapi.Response(http.StatusOK, openapi.GeneralSuccess{
-		Message: fmt.Sprintf("'odo dev' instance with pid: %d is shuting down.", odocontext.GetPID(ctx)),
+		Message: fmt.Sprintf("'odo dev' instance with pid: %d is shutting down.", odocontext.GetPID(ctx)),
 	}), nil
 }
 
--- a/pkg/apiserver-impl/api_default_service.go
+++ b/pkg/apiserver-impl/api_default_service.go
@@ -45,8 +45,17 @@ func NewDefaultApiService(
 func (s *DefaultApiService) ComponentCommandPost(ctx context.Context, componentCommandPostRequest openapi.ComponentCommandPostRequest) (openapi.ImplResponse, error) {
 	switch componentCommandPostRequest.Name {
 	case "push":
-		s.pushWatcher <- struct{}{}
-		return openapi.Response(http.StatusOK, nil), nil
+		select {
+		case s.pushWatcher <- struct{}{}:
+			return openapi.Response(http.StatusOK, openapi.GeneralSuccess{
+				Message: "push was successfully executed",
+			}), nil
+		default:
+			return openapi.Response(http.StatusTooManyRequests, openapi.GeneralError{
+				Message: "a push operation is not possible at this time. Please retry later",
+			}), nil
+		}
+
 	default:
 		return openapi.Response(http.StatusBadRequest, nil), fmt.Errorf("command name %q not supported. Supported values are: %q", componentCommandPostRequest.Name, "push")
 	}
--- a/pkg/apiserver-impl/api_default_service.go
+++ b/pkg/apiserver-impl/api_default_service.go
@@ -3,6 +3,7 @@ package apiserver_impl
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 
 	openapi "github.com/redhat-developer/odo/pkg/apiserver-gen/go"
@@ -13,11 +14,14 @@ import (
 // This service should implement the business logic for every endpoint for the DefaultApi API.
 // Include any external packages or services that will be required by this service.
 type DefaultApiService struct {
+	cancel context.CancelFunc
 }
 
 // NewDefaultApiService creates a default api service
-func NewDefaultApiService() openapi.DefaultApiServicer {
-	return &DefaultApiService{}
+func NewDefaultApiService(cancel context.CancelFunc) openapi.DefaultApiServicer {
+	return &DefaultApiService{
+		cancel: cancel,
+	}
 }
 
 // ComponentCommandPost -
@@ -44,13 +48,10 @@ func (s *DefaultApiService) ComponentGet(ctx context.Context) (openapi.ImplRespo
 
 // InstanceDelete -
 func (s *DefaultApiService) InstanceDelete(ctx context.Context) (openapi.ImplResponse, error) {
-	// TODO - update InstanceDelete with the required logic for this service method.
-	// Add api_default_service.go to the .openapi-generator-ignore to avoid overwriting this service implementation when updating open api generation.
-
-	// TODO: Uncomment the next line to return response Response(200, GeneralSuccess{}) or use other options such as http.Ok ...
-	// return Response(200, GeneralSuccess{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("InstanceDelete method not implemented")
+	s.cancel()
+	return openapi.Response(http.StatusOK, openapi.GeneralSuccess{
+		Message: fmt.Sprintf("'odo dev' instance with pid: %d is shuting down.", odocontext.GetPID(ctx)),
+	}), nil
 }
 
 // InstanceGet -
--- a/pkg/apiserver-impl/api_default_service.go
+++ b/pkg/apiserver-impl/api_default_service.go
@@ -3,8 +3,10 @@ package apiserver_impl
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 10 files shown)


tests/integration/cmd_dev_api_server_test.go
```diff
@@ -63,11 +69,157 @@ var _ = Describe("odo dev command with api server tests", func() {
...
+					When("/component/command endpoint is POSTed", func() {
+
+						BeforeEach(func() {
+							url := fmt.Sprintf("http://%s/component/command", devSession.APIServerEndpoint)
+							resp, err := http.Post(url, "application/json", bytes.NewBuffer([]byte(`{"name": "push"}`)))
+							Expect(err).ToNot(HaveOccurred())
+							Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusOK))
+						})
+
+						It("should trigger a push", func() {
```
valaparthvi, line 181: Do you think it makes sense to add this test when `--no-watcher` is _not_ used? Functionally it makes sense to only test it when there is no watcher, but I would still expect it to work just like manual push by pressing `p`.

valaparthvi (changes requested):
The normal workflow works well for me, but I am unable to run parallel sessions on mac:
```sh
$ ODO_EXPERIMENTAL_MODE=t odo dev --api-server

  __
 /  \__     Developing using the "my-go-app" Devfile
 \__/  \    Namespace: default
 /  \__/    odo version: v3.11.0
 \__/

 ⚠  You are using "default" namespace, odo may not work as expected in the default namespace.
 ⚠  You may set a new namespace by running `odo create namespace <name>`, or set an existing one by running `odo set namespace <name>`

↪ Running on the cluster in Dev mode
I0621 16:41:04.909656   22390 starterserver.go:58] API Server started at localhost:20000/api/v1
I0621 16:41:04.909831   22390 starterserver.go:78] Stopping the API server; encountered error: listen tcp :20000: bind: address already in use
 ✗  error watching deployment: context canceled

$ ODO_EXPERIMENTAL_MODE=t odo dev --api-server  --platform podman

  __
 /  \__     Developing using the "my-go-app" Devfile
 \__/  \    Platform: podman
 /  \__/    odo version: v3.11.0
 \__/

I0621 16:44:09.646006   22887 starterserver.go:58] API Server started at localhost:20000/api/v1
I0621 16:44:09.646188   22887 starterserver.go:78] Stopping the API server; encountered error: listen tcp :20000: bind: address already in use
 ✓  Deploying pod [8s]
 ✗  Syncing files into the container [4ms]
Error occurred on Push - failed to sync to component with name my-go-app: unable to exec command [rm -rf /projects/*]: context canceled


↪ Dev mode
 Status:
 Watching for changes in the current directory /tmp/nodejs-debug

 Keyboard Commands:
Cleaning up resources
 ✗  Dev mode interrupted by user
```

I think it again has to do with the fact that mac believes a port is free even if it is in use.

Perhaps we can add an `--api-address` flag in the future iterations.

Also, cluster was very quick in terminating the session, but podman took a while to terminate the session when context was cancelled.

pkg/apiserver-impl/api_default_service.go
```diff
@@ -2,62 +2,78 @@ package apiserver_impl
...
-
-	// TODO: Uncomment the next line to return response Response(200, ComponentGet200Response{}) or use other options such as http.Ok ...
-	// return Response(200, ComponentGet200Response{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("ComponentGet method not implemented")
+	value, _, err := describe.DescribeDevfileComponent(ctx, s.kubeClient, s.podmanClient, s.stateClient)
+	if err != nil {
+		return openapi.Response(http.StatusInternalServerError, ""), fmt.Errorf("error getting the description of the component: %w", err)
+	}
+	return openapi.Response(http.StatusOK, value), nil
```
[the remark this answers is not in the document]
feloy: I'm not sure it is necessary, as you can pipe the output of the request with a tool similar to json_reformat to get an indented output.

Furthermore, this change would be part of the apiserver-gen code, wnd I would prefer not to maintain a kind of fork of this generated code.

pkg/dev/interface.go
```diff
@@ -36,7 +36,8 @@ type StartOptions struct {
 	// Applicable to Podman only.
 	ForwardLocalhost bool
 	// Variables to override in the Devfile
-	Variables map[string]string
+	Variables   map[string]string
+	PushWatcher <-chan struct{}
```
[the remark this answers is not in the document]
feloy: Adding a comment.

For the moment, it is used by the API server only, but is intended to be used by any part of the code. I would prefer not to specialize it

tests/integration/cmd_dev_api_server_test.go
```diff
@@ -63,11 +69,157 @@ var _ = Describe("odo dev command with api server tests", func() {
...
+						url := fmt.Sprintf("http://%s/instance", devSession.APIServerEndpoint)
+						resp, err := http.Get(url)
+						Expect(err).ToNot(HaveOccurred())
+						Expect(resp.StatusCode).To(BeEquivalentTo(http.StatusOK))
+						defer resp.Body.Close()
+						body, err := io.ReadAll(resp.Body)
+						Expect(err).ToNot(HaveOccurred())
+						strBody := string(body)
+						helper.JsonPathExist(strBody, "pid")
+						helper.JsonPathContentIs(strBody, "componentDirectory", commonVar.Context)
```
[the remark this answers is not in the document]
feloy: Why this difference on macOS?

tests/integration/cmd_dev_api_server_test.go
```diff
@@ -1,12 +1,18 @@
 package integration
```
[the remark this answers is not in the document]
feloy: For the case when DELETE /instance is called, the dev session will be terminated twice (one with DELETE, one during AfterEach).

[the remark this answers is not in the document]
valaparthvi: That's right. But is that a bad thing? I debugged through the steps a couple of times, it seems it simply runs through the checks, the first time `WaitEnd()` is called from `It` block, it waits until the session terminates, but the second time `WaitEnd`()` is called from `AfterEach` it runs through the steps and returns immediately. I am assuming there's not much to be done for `Stop()` after DELETE/instance is called.

tests/integration/cmd_dev_api_server_test.go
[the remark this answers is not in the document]
valaparthvi: I am not sure, it seems macOS appends `/private` to all the paths, I think it is more like a temp directory on macOS. All the directories I create with `mkdir /tmp/something`, always return `/private/tmp/something` when I `pwd` on it.

But I think this can be ignored, it might be specific to my system alone.

Some more info is available here: https://apple.stackexchange.com/a/227869

tests/integration/cmd_dev_api_server_test.go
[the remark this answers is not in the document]
feloy: I remember changing these kind of tests because calling Stop() when the dev session is already terminated is causing problems (on Windows?).

pkg/component/describe/describe.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	if !isPlatformFeatureEnabled {
+		// Display RunningOn field only if the feature is enabled
+		cmp.RunningOn = nil
+	}
+	updateWithRemoteSourceLocation(&cmp)
+	return cmp, devfileObj, err
+}
+
+// describeNamedComponent describes a component given its name
```
rm3l, line 134: nit

```suggestion
// DescribeNamedComponent describes a component given its name
```

pkg/component/describe/describe.go
```diff
@@ -0,0 +1,250 @@
...
+func DescribeNamedComponent(
+	ctx context.Context,
+	name string,
+	kubeClient kclient.ClientInterface,
+	podmanClient podman.Client,
+	// stateClient state.Client,
```
rm3l, line 140:
```suggestion
```

pkg/apiserver-impl/api_default_service.go
```diff
@@ -2,62 +2,78 @@ package apiserver_impl
...
-	// Add api_default_service.go to the .openapi-generator-ignore to avoid overwriting this service implementation when updating open api generation.
-
-	// TODO: Uncomment the next line to return response Response(200, GeneralSuccess{}) or use other options such as http.Ok ...
-	// return Response(200, GeneralSuccess{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("ComponentCommandPost method not implemented")
+	switch componentCommandPostRequest.Name {
+	case "push":
+		s.pushWatcher <- struct{}{}
+		return openapi.Response(http.StatusOK, nil), nil
```
rm3l, line 49: I think the response body should contain a message like when the operation has been submitted successfully. At least, that's what I can see from the response example provided in the example:
```json
{
  "message": "push was successfully executed"
}
```

pkg/apiserver-impl/api_default_service.go
```diff
@@ -2,62 +2,78 @@ package apiserver_impl
...
+	default:
+		return openapi.Response(http.StatusBadRequest, nil), fmt.Errorf("command name %q not supported. Supported values are: %q", componentCommandPostRequest.Name, "push")
```
rm3l, line 51: Wondering if the response body in case of error should not be serialized as a real JSON string too. Currently, I get a simple string, which does not respect the response content type returned in the response headers:

```
$ http POST http://127.0.0.1:20002/api/v1/component/command name=fake-cmd
HTTP/1.1 400 Bad Request
Content-Length: 74
Content-Type: application/json; charset=UTF-8
Date: Thu, 22 Jun 2023 15:45:12 GMT

"command name \"fake-cmd\" not supported. Supported values are: \"push\""
```

I'd expect something like below. What do you think?

```json
{
  "message": "command name \"fake-cmd\" not supported. Supported values are: \"push\""
}
```

pkg/apiserver-impl/api_default_service.go
```diff
@@ -2,62 +2,78 @@ package apiserver_impl
...
-	// TODO - update ComponentCommandPost with the required logic for this service method.
-	// Add api_default_service.go to the .openapi-generator-ignore to avoid overwriting this service implementation when updating open api generation.
-
-	// TODO: Uncomment the next line to return response Response(200, GeneralSuccess{}) or use other options such as http.Ok ...
-	// return Response(200, GeneralSuccess{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("ComponentCommandPost method not implemented")
+	switch componentCommandPostRequest.Name {
+	case "push":
+		s.pushWatcher <- struct{}{}
```
rm3l, line 48: Also, one thing I just noticed with this: when there is already a 'push' operation in progress (for example, I sent a first `POST /component/command` request, immediately followed by the same request again), the second request would be blocked waiting for the in-progress push operation to finish (which in my case took a lot of seconds).
Wouldn't it make more sense to return immediately if there is already a 'push' operation?

pkg/apiserver-impl/api_default_service.go
feloy: I think we will need to complete the openapi spec, to point to the correct `GeneralError` instead of using the default. WDYT?

pkg/apiserver-impl/api_default_service.go
feloy: Yes, that makes sense. In this case, which status code do we want to return?

rm3l: Maybe a `429 Too Many Requests` would be relevant, as the user can retry later when there is no other push in progress?

pkg/apiserver-impl/api_default_service.go
rm3l: Yes, that makes sense.

pkg/apiserver-impl/api_default_service.go
rm3l: @feloy Not sure if you saw my previous comment.. What do you think?

feloy: I changed from `StatusServiceUnavailable` to `StatusTooManyRequests`.

I think we will need to update the openapi spec for the non 2xx responses, but I would prefer we discuss it before to write it. That could be part of another PR?

rm3l: Yes, sure, that can be a separate PR indeed.

rm3l (approved):

feloy:
```
 [FAILED] [10.576 seconds]
odo devfile deploy command tests deploying devfile with exec when using devfile where the exec command is bound to fail [It] should print the last 100 lines of the log to the output
/go/odo_1/tests/integration/cmd_devfile_deploy_test.go:648
```

rm3l (approved):

PR #6920, opened 2023-06-21
Disable service binding related integration tests

feloy:
**What type of PR is this:**

/area testing

The purpose of this PR is to test the impact of removing the integration tests related to service binding.

It introduces an `SKIP_SERVICE_BINDING_TESTS` environment variable which indidate to skip tests related to service binding.

This variable has been set to `true` for *Kubernetes*, *OpenShift* and *Windows* tests on IBM Cloud, to disable these tests for PRs.

The tests will still run during the nightly tests.

### DIFF
--- a/.ibm/pipelines/windows-test-script.ps1
+++ b/.ibm/pipelines/windows-test-script.ps1
@@ -57,6 +57,7 @@ function Run-Test {
     [Environment]::SetEnvironmentVariable("TEST_EXEC_NODES", "$TEST_EXEC_NODES") 
     [Environment]::SetEnvironmentVariable("SKIP_USER_LOGIN_TESTS","true")
     [Environment]::SetEnvironmentVariable("SKIP_WELCOMING_MESSAGES","true")
+    [Environment]::SetEnvironmentVariable("SKIP_SERVICE_BINDING_TESTS","$SKIP_SERVICE_BINDING_TESTS" )
     # Integration tests detecting key press when running DevSession are not working on Windows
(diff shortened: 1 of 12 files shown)


valaparthvi (changes requested):
e2e_test.go:`Context("starting with non-empty Directory add Binding", func() {`
 can use skipping.

feloy:
> e2e_test.go:`Context("starting with non-empty Directory add Binding", func() {` can use skipping.

No, I would like to keep the tests for e2e tests.

valaparthvi:
>Also, wouldn't you need to modify IBM pipelines to export the env var? I believe the tests are still run on OpenShift, since I do not see any difference between the number of k8s and oc tests.

In case you missed.

feloy:
> > Also, wouldn't you need to modify IBM pipelines to export the env var? I believe the tests are still run on OpenShift, since I do not see any difference between the number of k8s and oc tests.
> 
> In case you missed.

I defined the env var in the pipelines, but after the job has been created. It seems the env variables are considered at the time the job is created (not the time it is started).

Let me try again

rm3l (approved):

feloy:
```
[FAILED] [74.869 seconds]
User guides: Quickstart test Java quickstart guide [It] should test the complete java springboot quickstart output in order
/go/odo_1/tests/documentation/user-guides/doc_user_guides_quickstart_test.go:217

  [odo]  âš   0/3 nodes are available: 3 pod has unbound immediate PersistentVolumeClaims. preemption: 0/3 nodes are available: 3 Preemption is not helpful for scheduling.

```

PR #6922, opened 2023-06-22
List namespace right after namespace has been created

### LINKED ISSUE #6827: `-w` flag for `odo create namespace` doesn't work as expected on `crc` 2.18 and 2.19
/kind bug

**Operating System:** RHEL 8.7

**Output of `odo version`:**

```
odo v3.10.0 (039ef0732)

Server: https://api.crc.testing:6443
Kubernetes: v1.25.8+27e744f
```

I currently have namespaces `asdf1` and `asdf2` created on the cluster, then I run:
`odo create namespace asdf3 -w; odo list namespace -o json`

## Actual behavior
```
 ✓  Waiting for namespace to come up [639ms]
 ✓  Namespace "asdf3" is ready for use
 ✓  New namespace created and now using namespace: asdf3
{
        "namespaces": [
                {
                        "name": "asdf1",
                        "active": false
                },
                {
                        "name": "asdf2",
                        "active": false
                }
        ]
}
```

## Expected behavior
```
 ✓  Waiting for namespace to come up [639ms]
 ✓  Namespace "asdf3" is ready for use
 ✓  New namespace created and now using namespace: asdf3
{
        "namespaces": [
                {
                        "name": "asdf1",
                        "active": false
                },
                {
                        "name": "asdf2",
                        "active": false
                },
                {
                        "name": "asdf3",
                        "active": true
                },
        ]
}
```

valaparthvi:
**What type of PR is this:**
/kind bug
/flake

This PR fixes the bug where `odo create namespace -w` does not work as expected.

Fixes #6827

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/odo/cli/create/namespace/namespace.go
+++ b/pkg/odo/cli/create/namespace/namespace.go
@@ -82,29 +82,21 @@ func (nco *NamespaceCreateOptions) Validate(ctx context.Context) error {
 
 // Run runs the namespace create command
 func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
-	// Create the "spinner"
-	s := &log.Status{}
-
-	// If the --wait parameter has been passed, we add a spinner..
-	if nco.waitFlag {
-		s = log.Spinnerf("Waiting for %s to come up", nco.commandName)
-		defer s.End(false)
-	}
+	createSpinner := log.Spinnerf("Creating the %s %q", nco.commandName, nco.namespaceName)
+	defer createSpinner.End(false)
 
 	// Create the namespace & end the spinner (if there is any..)
 	err = nco.clientset.ProjectClient.Create(nco.namespaceName, nco.waitFlag)
 	if err != nil {
 		return err
 	}
-	s.End(true)
+	createSpinner.End(true)
 
-	// Set the current namespace when created
-	err = nco.clientset.ProjectClient.SetCurrent(nco.namespaceName)
-	if err != nil {
-		return err
-	}
+	// If the --wait parameter has been passed, we add a spinner..
 	if nco.waitFlag {
-		var timeOut = time.After(nco.clientset.PreferenceClient.GetTimeout())
+		waitSpinner := log.Spinnerf("Waiting for the %s to come up", nco.commandName)
+		defer waitSpinner.End(false)
+		timeOut := time.After(nco.clientset.PreferenceClient.GetTimeout())
 	L:
 		for {
 			select {
@@ -124,6 +116,13 @@ func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
 				time.Sleep(50 * time.Millisecond)
 			}
 		}
+		waitSpinner.End(true)
+	}
+
+	// Set the current namespace when created
+	err = nco.clientset.ProjectClient.SetCurrent(nco.namespaceName)
+	if err != nil {
+		return err
 	}
 	caser := cases.Title(language.Und)
 	successMessage := fmt.Sprintf(`%s %q is ready for use`, caser.String(nco.commandName), nco.namespaceName)
--- a/pkg/odo/cli/create/namespace/namespace.go
+++ b/pkg/odo/cli/create/namespace/namespace.go
@@ -4,12 +4,11 @@ import (
 	"context"
 	"fmt"
 	"github.com/redhat-developer/odo/pkg/project"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"k8s.io/klog"
 	"os"
 	"time"
-
-	"golang.org/x/text/cases"
-	"golang.org/x/text/language"
 
 	dfutil "github.com/devfile/library/v2/pkg/util"
 	"github.com/spf13/cobra"
@@ -99,10 +98,6 @@ func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
 	}
 	s.End(true)
 
-	caser := cases.Title(language.Und)
-	successMessage := fmt.Sprintf(`%s %q is ready for use`, caser.String(nco.commandName), nco.namespaceName)
-	log.Successf(successMessage)
-
 	// Set the current namespace when created
 	err = nco.clientset.ProjectClient.SetCurrent(nco.namespaceName)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/odo/cli/create/namespace/namespace.go
```diff
@@ -105,7 +108,28 @@ func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
...
+				var nsList project.ProjectList
+				nsList, err = nco.clientset.ProjectClient.List()
+				if err != nil {
+					klog.V(4).Infof("Failed to list %ss", nco.commandName)
+				}
+				for _, ns := range nsList.Items {
+					if ns.Name == nco.namespaceName {
+						break L
+					}
+				}
```
feloy, line 129: I would add a sleep for let's say 50ms at this point, to not respawn the List operation too fast

feloy (changes requested):

pkg/odo/cli/create/namespace/namespace.go
valaparthvi: That's a good idea.

feloy (approved):
Thanks for this fix

pkg/odo/cli/create/namespace/namespace.go
```diff
@@ -105,7 +108,29 @@ func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
-
+	if nco.waitFlag {
+		var timeOut = time.After(nco.clientset.PreferenceClient.GetTimeout())
```
rm3l, line 112: The default value for `nco.clientset.PreferenceClient.GetTimeout()` is currently [1 second](https://odo.dev/docs/overview/configure#preference-key-table). Isn't that too low?

rm3l: Also, to me, if the user sets `--wait`, it means that they purposely are willing to wait until they get confirmation that the namespace/project is returned correctly. So IMO, we should not even have a timeout.
The user would need to interrupt the process to cancel the wait.
I'm thinking for example a script using `odo create namespace --wait` before doing other things..

feloy: For me, it's a matter of 10's of milliseconds. Having to break manually can be dangerous if executed from a script: is the user expecting the command will never return in case of problem? I would instead return an error if the namespace is not created after a *reasonable* time. As the timeout is configurable, I would keep as is, and users in heavy loaded systems could increase it.

rm3l: Yup, that makes sense. Personally, I found this `Timeout` preference confusing and not used consistently. I think every command with a `--wait` flag should also provide an additional `--timeout` flag to set a timeout only for this command, but that's a different story.
Alright, in this case, returning an error if the timeout is reached would indeed make more sense. The error should also clearly indicate how the user can adjust this timeout, I think.

cc @valaparthvi

rm3l (commented):
One last comment: I think the success message (`Namespace "$ns" is ready for use`) should be displayed only after the new checks are done. Otherwise, the output might be confusing if the namespace still does not appear in the list after the timeout is reached, e.g.:

```
$ odo create namespace test-ns11 --wait
 ✓  Waiting for namespace to come up [11ms]
 ✓  Namespace "test-ns11" is ready for use
 ⚠  Timeout while waiting for namespace "test-ns11"; it may take a while to appear
```

I'd suggest displaying ` ✓  Namespace "test-ns11" is ready for use` only if the namespace does appear in the list. What do you think?

pkg/odo/cli/create/namespace/namespace.go
```diff
@@ -99,10 +98,6 @@ func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
 	}
 	s.End(true)
```
rm3l, line 99: Can you move this `s.End(true)` line after the `for` loop? Otherwise, it looks like `odo` is blocked if the `for` loop  takes too long to return.

[Screencast from 2023-06-27 09-43-07.webm](https://github.com/redhat-developer/odo/assets/593208/763f8cd2-79fe-4df4-834e-19f433250823)

rm3l (changes requested):

rm3l (approved):

pkg/odo/cli/create/namespace/namespace.go
```diff
@@ -80,32 +82,51 @@ func (nco *NamespaceCreateOptions) Validate(ctx context.Context) error {
...
 func (nco *NamespaceCreateOptions) Run(ctx context.Context) (err error) {
-	// Create the "spinner"
-	s := &log.Status{}
-
-	// If the --wait parameter has been passed, we add a spinner..
-	if nco.waitFlag {
-		s = log.Spinnerf("Waiting for %s to come up", nco.commandName)
-		defer s.End(false)
-	}
+	createSpinner := log.Spinnerf("Creating the %s %q", nco.commandName, nco.namespaceName)
```
rm3l, line 85: I guess you also need to update the doc automation tests as well:

```
  << Timeline                                                                                                                                                                                  
                                                                                                                                                                                               
  [FAILED] create_namespace.mdx                                                                                                                                                                
  Expected                                                                                                                                                                                     
      <string>:   (                                                                                                                                                                            
                """                                                                                                                                                                            
                ```console                                                                                                                                                                     
                $ odo create namespace odo-dev                                                                                                                                                 
      +         •  Creating the namespace "odo-dev"  ...
      +         ✓  Creating the namespace "odo-dev" [1s]
                ✓  Namespace "odo-dev" is ready for use
                ✓  New namespace created and now using namespace: odo-dev
                ... // 2 identical lines
                """
        )
       
  to be empty
  In [It] at: /home/asoro/work/projects/odo/tests/documentation/command-reference/doc_command_reference_createnamespace_test.go:42 @ 06/27/23 12:03:09.82
•

Summarizing 2 Failures:
  [FAIL] doc command reference odo create namespace To create a namespace resource [It] Creates a project resource for a kubernetes cluster
  /home/asoro/work/projects/odo/tests/documentation/command-reference/doc_command_reference_createnamespace_test.go:52
  [FAIL] doc command reference odo create namespace To create a namespace resource [It] Creates a namespace resource for a kubernetes cluster
  /home/asoro/work/projects/odo/tests/documentation/command-reference/doc_command_reference_createnamespace_test.go:42
```

Also, can you rebase your branch onto `main`?

rm3l (approved):

PR #6938, opened 2023-06-28
Document and return JSON data for API errors

feloy:
**What type of PR is this:**

/kind bug

Fixes the API response data for the error cases, and documents these error responses in the openapi spec

### DIFF
--- a/ododevapispec.yaml
+++ b/ododevapispec.yaml
@@ -172,6 +172,14 @@ paths:
                   },
                   "managedBy": "odo"
                }
+        '500':
+          description: error getting the description of the component
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GeneralError'
+              example:
+                message: "error getting the description of the component"
 
   /component/command:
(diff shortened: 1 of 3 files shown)


rm3l (approved):

feloy:
```
[FAILED] [137.366 seconds]
odo dev command with api server tests when the component is bootstrapped when odo is executed with --no-watch and --api-server flags [BeforeEach] when a file in component directory is modified when /component/command endpoint is POSTed should trigger a push
  [BeforeEach] C:/Users/Administrator.ANSIBLE-TEST-VS/4280/tests/integration/cmd_dev_api_server_test.go:180
  [It] C:/Users/Administrator.ANSIBLE-TEST-VS/4280/tests/integration/cmd_dev_api_server_test.go:213
[...]
starterserver.go:78] Stopping the API server; encountered error: listen tcp :20000: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.
```

PR #6970, opened 2023-07-17
Propagate local Devfile changes to the UI

### LINKED ISSUE #6902: Propagate changes from the local Devfile to the `odo dev` web UI
/kind user-story

## User Story
As an odo user, I want my changes to be propagated automatically from the local Devfile to the `odo dev` web UI and the other way around, So that my changes are reflected regardless of where those changes are done.

## Acceptance Criteria 
- [x] When the local Devfile is modified and saved using some other editor, changes should be reflected in the `odo dev` web UI (done in https://github.com/redhat-developer/odo/pull/6970)
- [x] It should warn UI users if the devfile has been changed on the filesystem => UI users should choose whether to overwrite their changes with the new Devfile (addressed in https://github.com/redhat-developer/odo/pull/6996). Example of how Gedit displays warnings:

Image

## Links 
- Related Epic (mandatory): https://github.com/redhat-developer/odo/issues/6895

/kind user-story

rm3l:
**What type of PR is this:**
/kind feature
/area dev

This PR introduces a new `/notifications` API endpoint that is able to send SSE events to connected clients. Currently, if sends `DevfileUpdated` events only (along with a Heartbeat comment) whenever the local Devfile is updated on the local filesystem.
The UI listens on this `DevfileUpdated` event and automatically overwrites the Devfile in the YAML view (and in other related tabs).
This is the first part for propagating changes from the local filesystem to the UI. Another PR will follow after to warn users whenever a `DevfileUpdated` event is received and allow them to either overwrite or discard the new Devfile.

Fixes acceptance criteria 1 of #6902

**PR acceptance criteria:**

- [ ] Unit test 

- [x] Integration test 

- [ ] Documentation 


feloy:
> Add '/notifications' endpoint for subscribing to server-sent events

I would prefer to add a `watch` parameter to the `GET /devfile` endpoint, to have an interface similar to the Kubernetes API's one.

I had created this PoC a while ago to get events from a watch K8s API's endpoint: https://github.com/feloy/kubernetes-explorer/blob/master/src/app/services/kubernetes.service.ts

ui/src/app/services/sse.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    constructor(private _zone: NgZone) {
+        this.evtSource = new EventSource(this.base + "/notifications");
+    }
+
+    subscribeTo(eventTypes: string[]): Observable<any> {
+        return new Observable( (subscriber) => {
+            eventTypes.forEach(eventType => {
+                this.evtSource.addEventListener(eventType,  (event) => {
+                    this._zone.run(() => {
```
feloy, line 19: I tried on my side, it seems that running into _zone is not necessary: the event is propaagted and the new Devfile is correctly displayed in the UI

pkg/apiserver-impl/sse/event.go
```diff
@@ -0,0 +1,46 @@
...
+	jsonBytes, err := json.Marshal(e.data)
+	if err != nil {
+		return "", err
+	}
+	var dataStr bytes.Buffer
+	err = json.Compact(&dataStr, jsonBytes)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("event: %s\ndata: %s\n\n", eventName, dataStr.String()), nil
```
feloy, line 45: To limit the round-trips, I would send the new content of the devfile as part of the event. WDYT?

ui/src/app/services/sse.service.ts
rm3l: Ah, interesting..
I took inspiration from various sources out there (like [here (in French)](https://blog.octo.com/angular-2-sse-et-la-detection-de-changements/),  [here](https://medium.com/@chrisbautistaaa/server-sent-events-in-angular-node-908830cc29aa), or [here](https://javascript.plainenglish.io/angular-and-server-sent-events-sse-54c29b4f174d)), and all of them relied on an Angular Zone. The previous version used `eventsource.onmessage`, but I changed it to `eventsource.addEventListener` when I wanted subscribers to be more specific on the event types they are interested in.
I'll remove it then if it works without this.
 Thanks.

pkg/apiserver-impl/sse/event.go
rm3l: Yes, that could be useful, indeed.

rm3l:
> > Add '/notifications' endpoint for subscribing to server-sent events
> 
> I would prefer to add a `watch` parameter to the `GET /devfile` endpoint, to have an interface similar to the Kubernetes API's one.
> 
> I had created this PoC a while ago to get events from a watch K8s API's endpoint: [feloy/kubernetes-explorer@`master`/src/app/services/kubernetes.service.ts](https://github.com/feloy/kubernetes-explorer/blob/master/src/app/services/kubernetes.service.ts?rgh-link-date=2023-07-17T07%3A18%3A45Z)

For the record, and as discussed together earlier today, the main reason for having this `/notifications` endpoint is to have a single EventSource connection from the client side that will receive any event. This is because browsers enforce a pretty-low limit on the number of persistent connections per domain. We might be quickly limited if we were to keep a permanent connection per type of event.

feloy (approved):

Issue #6966, opened 2023-07-12
`odo list namespace` tests seem flaky in periodic test infra


rm3l:
/kind bug
/area testing
/area flake

## Any logs, error output, etc?

- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-redhat-developer-odo-main-v4.11-integration-e2e-periodic/1679279741862940672
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-redhat-developer-odo-main-v4.10-integration-e2e-periodic/1678917250859405312
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-redhat-developer-odo-main-v4.11-integration-e2e-periodic/1678554931772002304
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-redhat-developer-odo-main-v4.13-integration-e2e-periodic/1678554934296973312
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-redhat-developer-odo-main-odo-ocp4.14-lp-interop-odo-scenario-aws/1678283073944293376

```
• [FAILED] [1.270 seconds]
odo create/delete/list/set namespace/project tests when namespace is created with -w [It] should list the new namespace when listing namespace
/go/src/github.com/redhat-developer/odo/tests/integration/cmd_namespace_test.go:36
  Timeline >>
  Created dir: /tmp/2077704388
  Created dir: /tmp/1854115374
  Setting KUBECONFIG=/tmp/1854115374/config
  Creating a new project: cmd-namespace-test36cyp
  Running oc with args [oc new-project cmd-namespace-test36cyp] and odo env: []
  [oc] Now using project "cmd-namespace-test36cyp" on server "https://api.ci-op-pbirj3v6-894cb.origin-ci-int-aws.dev.rhcloud.com:6443".
  [oc] 
  [oc] You can add applications to this project with the 'new-app' command. For example, try:
  [oc] 
  [oc]     oc new-app rails-postgresql-example
  [oc] 
  [oc] to build a new example application in Ruby. Or use kubectl to deploy a simple Kubernetes application:
  [oc] 
  [oc]     kubectl create deployment hello-node --image=k8s.gcr.io/e2e-test-images/agnhost:2.33 -- /agnhost serve-hostname
  [oc] 
[...]

PR #6989, opened 2023-07-20
Serve SwaggerUI

### LINKED ISSUE #6986: It should be documented (later). We can even publish the OpenAPI spec (Swagger UI?) (later)

feloy:
**What type of PR is this:**

/kind feature

Serve a swagger ui from `odo dev` / `odo api-server`

Instructions from:
https://ribice.medium.com/serve-swaggerui-within-your-golang-application-5486748a5ed4

Fixes #6986 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 


Makefile
rm3l, line 1: I know we discussed adding a comment in the code indicating where to find the `swagger-ui` files, but I'm wondering if we should not actually add an additional target in the `Makefile` to automate the whole process: download the swagger-ui distribution from GitHub and override the whole `pkg/swagger-imp/swagger-ui` folder?
This would serve as our reproducible source of truth. What do you think?

Makefile
```diff
@@ -86,10 +86,12 @@ ui-static: ## build static files for UI to be served by embedded API server
 
 .PHONY: bin
 bin: ## build the odo binary
+	cp ododevapispec.yaml pkg/apiserver-impl/swagger-ui/swagger.yaml
```
rm3l, line 89: guess we should improve the [`check-generated-files` workflow](https://github.com/redhat-developer/odo/blob/main/.github/workflows/check-generated-files.yaml) to make sure `pkg/apiserver-impl/swagger-ui/swagger.yaml` in kept in sync with `ododevapispec.yaml`.

rm3l (commented):
As discussed also, we'll need to add a simple page on odo.dev (under "Advanced Usage") mentioning the API Server and the Swagger UI.

Makefile
feloy: As this file is copied every time we build odo, I would say no (it is different from the ui files which are much more longer to generate/copy and which we need to copy manually)

rm3l: Yes, but currently, it is copied every time you build with `make bin`, but not every time you run `make install` (which I often do instead of `make bin`)..

rm3l: And even if it is copied, if the developer chooses not to commit/push the `pkg/apiserver-impl/swagger-ui/swagger.yaml` file, we might end up with a binary not serving the right spec file. So we need to take care of this when reviewing PRs changing the OpenAPI spec file..

rm3l (approved):

rm3l:
```

Summarizing 2 Failures:
  [FAIL] odo create/delete/list/set namespace/project tests when namespace is created with -w [It] should list the new namespace when listing namespace
  /go/odo_1/tests/integration/cmd_namespace_test.go:38
  [FAIL] odo devfile deploy command tests when starting with Devfile with autoBuild or deployByDefault components when running odo deploy with some components not referenced in the Devfile [BeforeEach] should create the appropriate resources
  /go/odo_1/tests/helper/helper_cmd_wrapper.go:101

Ran 406 of 956 Specs in 1463.962 seconds
FAIL! -- 404 Passed | 2 Failed | 0 Pending | 550 Skipped
```

```
 [odo]  âœ—  failed to create service(s) associated with the component: pods "k8s-deploybydefault-true-and-referenced" is forbidden: error looking up service account cmd-devfile-deploy-test723cyz/default: serviceaccount "default" not found
```

Unrelated issue (hopefully temporary) - but let's keep an eye on this..

```
odo create/delete/list/set namespace/project tests when namespace is created with -w [It] should list the new namespace when listing namespace
/go/odo_1/tests/integration/cmd_namespace_test.go:36
...
  to contain substring
      <string>: cmd-namespace-test36csz
  In [It] at: /go/odo_1/tests/integration/cmd_namespace_test.go:38 @ 07/21/23 07:21:41.626
```

Unrelated issue with `odo list namespace` tests (already reported in #6966)

PR #7004, opened 2023-07-27
odo init filters devfile stacks by supported architectures

### LINKED ISSUE #6803: Filtering devfile stacks based on the architecture field in metadata when running odo init
## User Story
- As an odo user using a cluster running on IBM Z
- I want `odo init` Stack selection to list only stacks supporting IBM Z
- So that I don't create a component that uses devfile that won't work on my cluster.

Parent user-story: #6669

## Acceptance Criteria
- [ ] Stacks without any architectures in the index are compatible with **all** of them
- [ ] odo init interactive mode should ask for architectures (multi-choice) before to ask to select a Devfile stack
- [ ] x86 architecture is checked by default
- [ ] make it work with `index` and `v2index`
- [ ] Possibly add human readable architecture names for the user selection
- [ ] On flag mode, add a flag --architecture=x64 --architecture=ppc64le. Default value: amd64 (to be sure it works for amd64 users)
Get available values at https://github.com/devfile/api/blob/main/pkg/devfile/header.go#L24-L29 to display it on help/output mesg, and error out if passed value not in the list

- No automatic 
- In Alizer mode, display the architectures supported by the chosen Devfile, so the user can reply No and select manually a new one 
- 
registry-library documentation on how to filter based on architecture https://github.com/devfile/registry-support/tree/main/registry-library#specify-registry-options

/kind user-story
/priority high

feloy:
**What type of PR is this:**

/kind feature

Fixes #6803 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

### DIFF
--- a/pkg/api/analyze.go
+++ b/pkg/api/analyze.go
@@ -20,5 +20,5 @@ type DetectionResult struct {
 	// Name represents the project/application name as detected by alizer
 	Name string `json:"name,omitempty"`
 	// Architectures represent the architectures with which the Devfile must be compatible with.
-	Architectures []string
+	Architectures []string `json:"architectures,omitempty"`
 }
--- a/pkg/api/analyze.go
+++ b/pkg/api/analyze.go
@@ -19,4 +19,6 @@ type DetectionResult struct {
 	DevfileVersion   string `json:"devfileVersion,omitempty"`
 	// Name represents the project/application name as detected by alizer
 	Name string `json:"name,omitempty"`
+	// Architectures represent the architectures with which the Devfile must be compatible with.
+	Architectures []string
 }
--- a/pkg/init/backend/interactive.go
+++ b/pkg/init/backend/interactive.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/v2/pkg/devfile/parser"
@@ -22,7 +23,8 @@ import (
 )
 
 const (
-	STATE_ASK_LANG = iota
+	STATE_ASK_ARCHITECTURES = iota
+	STATE_ASK_LANG
 	STATE_ASK_TYPE
 	STATE_ASK_VERSION
 	STATE_END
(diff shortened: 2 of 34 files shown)


pkg/api/analyze.go
```diff
@@ -19,4 +19,6 @@ type DetectionResult struct {
 	DevfileVersion   string `json:"devfileVersion,omitempty"`
 	// Name represents the project/application name as detected by alizer
 	Name string `json:"name,omitempty"`
+	// Architectures represent the architectures with which the Devfile must be compatible with.
+	Architectures []string
```
rm3l, line 23: This will consequently be exposed in the output of `odo analyze -o json`:
```json
[
  {
    "devfile": "dotnet50",
    "devfileRegistry": "StagingRegistry",
    "devfileVersion": "1.0.3",
    "name": "my-dotnet",
    "Architectures": null
  }
]
```

I think it is okay, but we should consistently lower the field case and omit the field if it is empty.

pkg/odo/cli/registry/registry.go
```diff
@@ -120,7 +120,7 @@ func NewCmdRegistry(name, fullName string, testClientset clientset.Clientset) *c
 	clientset.Add(listCmd, clientset.REGISTRY)
 
 	// Flags
-	listCmd.Flags().StringVar(&o.filterFlag, "filter", "", "Filter based on the name or description or supported architecture of the component")
+	listCmd.Flags().StringVar(&o.filterFlag, "filter", "", "Filter based on the name or description or supported architecture of the component. Comma-separated list of search terms")
```
rm3l, line 123: Maybe we can also explicitly state that the search is done using a logical AND. Otherwise, it can be surprising at first sight IMO.

```suggestion
	listCmd.Flags().StringVar(&o.filterFlag, "filter", "", "Comma-separated list of terms for filtering. Search is done using a logical AND against the name or description or supported architectures of the component.")
```

pkg/init/backend/interactive.go
```diff
@@ -51,22 +53,36 @@ func (o *InteractiveBackend) Validate(flags map[string]string, fs filesystem.Fil
...
 loop:
 	for {
 		switch state {
 
+		case STATE_ASK_ARCHITECTURES:
+			archs, err = o.askerClient.AskArchitectures(knownArchitectures, archs)
+			if err != nil {
+				return nil, err
+			}
+			state = STATE_ASK_LANG
```
rm3l, line 71:
```
[...]
? Select architectures to support: 
? Select language: .NET
? Select project type:  [Use arrows to move, type to filter]
> .NET 5.0 (dotnet50, registry: StagingRegistry)
  .NET 5.0 (dotnet50, registry: DefaultDevfileRegistry)
  .NET 6.0 (dotnet60, registry: StagingRegistry)
  .NET 6.0 (dotnet60, registry: DefaultDevfileRegistry)
  .NET Core 3.1 (dotnetcore31, registry: StagingRegistry)
  .NET Core 3.1 (dotnetcore31, registry: DefaultDevfileRegistry)
  ** GO BACK **
```

In the `Select architectures to support` prompt, what does it mean to select nothing? I would assume no Devfile should be proposed if the user selects nothing, because it means to me "select a Devfile not compatible with any of the supported architectures", which basically should not be allowed.
So would it be possible to disable the "none" selection, so the user selects at least one architecture? Or prompt again the user to select an architecture if the list is empty?
What do you think?

rm3l (changes requested):

pkg/init/backend/interactive.go
feloy: In reality, it is more a "filter" than a "support". When no one is selected, the operation done is that no filter is applied.

rm3l: Ah, that makes sense then. In this case, maybe we can rename the prompt into something like: `Filter by supported architectures`?

feloy: What about `Select architectures to filter on:`?

rm3l: Sounds good too, as long as it is clear that we are filtering, I think ;-)

feloy:
```
[FAILED] [2.933 seconds]
odo init interactive command tests label nocluster [It] should start downloading starter project only after all interactive questions have been asked [nocluster]
```

PR #7012, opened 2023-08-01
Move UI out of experimental mode

### LINKED ISSUE #6963: Move `odo dev` UI out of the experimental mode
/kind task
/area dev

`odo dev` UI and API are currently hidden under the experimental mode. We should promote them - maybe after all the other issues in [the epic](https://github.com/redhat-developer/odo/issues/6895) are done.

## Acceptance Criteria 
- [ ] It should not be hidden in experimental mode
- [ ] Web UI endpoint should be exposed by default in `odo dev`
- [ ] The output of the `odo dev` should show on which host and port the web interface and API are served
- [ ] `odo describe component` should display the URL to the web UI in both human-readable and JSON formats. Currently, it is only available when running under the experimental mode
- [ ] Document in https://github.com/redhat-developer/odo/blob/main/USAGE_DATA.md what kind of data is captured from the UI
- [ ] Document the use cases and limitations (e.g., parent support) of the UI
- [ ] UI should include a link to the documentation about limitations
- [ ] Remove link to previous repo on "Work in progress"

## Links 
- Related Epic (mandatory): https://github.com/redhat-developer/odo/issues/6895

feloy:
**What type of PR is this:**

/kind feature

Fixes #6963

**PR acceptance criteria:**

- [x] Unit test 

- [x] Integration test 

- [x] Documentation 

Preview: https://odo-dev-pr-7012.odo-test-kubernete-449701-49529fc6e6a4a9fe7ebba9a3db5b55c4-0000.eu-de.containers.appdomain.cloud/docs/user-guides/advanced/using-gui/


pkg/apiserver-impl/starterserver.go
```diff
@@ -129,7 +130,11 @@ func StartServer(
 		cancelFunc()
 	}
 
-	klog.V(0).Infof("API Server started at localhost:%d/api/v1", listeningPort)
+	if feature.IsEnabled(ctx, feature.UIServer) {
+		log.Spinner(fmt.Sprintf("Web console accessible at http://localhost:%d/", listeningPort)).End(true)
```
rm3l, line 134: Can we make all these messages (or just the URLs) stand out, by printing them in bold green for example (just like the info about ports forwarded)?

ui/src/app/app.component.html
```diff
@@ -1,8 +1,7 @@
 <mat-toolbar color="primary">
     <span>Devfile Builder</span>
```
rm3l, line 2: Those were not part of the original acceptance criteria (so can be done later), but I was wondering if we could also change the following things before moving the UI out of the experimental mode:
- changing the favicon from the Angular logo to the `odo` logo.
- changing this title from "Devfile Builder" to something like "Devfile Builder (Beta)". A basic chip near the title, like below, would be even better but that can be done later:
image

rm3l (changes requested):

pkg/apiserver-impl/starterserver.go
feloy: I'm not sure. URLs are already clickable and highlighted in some way by terminals, as they start with http://

ui/src/app/app.component.html
feloy: - ok for the favicon, I'll change it as part of this PR
- For the Beta info, there is a "Work In Progress" on the right side of the toolbar

pkg/apiserver-impl/starterserver.go
rm3l: I guess it would depend on the terminal emulator, because mine only highlights the link when I hover over it. But I was mostly concerned about having a way to quickly visually catch the information, in case it is lost in the middle of the logs (for example if I run `odo dev -v 2` or even just `odo dev --logs`).
Actually, what do you think of having all this displayed under the "Dev mode" status info?

```
[...]
↪ Dev mode
 Status:
 Watching for changes in the current directory /home/asoro/work/tmp/6963-move-odo-dev-ui-out-of-the-experimental-mode
 - Web console accessible at http://localhost:20000/
 - API Server started at http://localhost:20000/api/v1
 - API documentation accessible at http://localhost:20000/swagger-ui/

 Keyboard Commands:

```

feloy: The info can be difficult to find with `-v 2`, but is it a standard way to run `odo dev`? Without this flag, it appears at the very beginning of the output, it seems to me very easy to find (even when `--logs` is used)

Why not for the web console, but for the API server/doc, these URLs will be used only by plugins developers, I don't think we want to spam the user with this info.

feloy: I have made the change to display the Web UI URL with the Status

rm3l (approved):
Great - thanks!

feloy:
```
[FAILED] [3.000 seconds]
odo create/delete/list/set namespace/project tests delete namespace when force-deleting a valid namespace [BeforeEach] should successfully delete the namespace asynchronously
  [BeforeEach] C:/Users/Administrator.ANSIBLE-TEST-VS/4469/tests/integration/cmd_namespace_test.go:74
  [It] C:/Users/Administrator.ANSIBLE-TEST-VS/4469/tests/integration/cmd_namespace_test.go:96

**Comment by feloy** [comment]
```
[FAILED] [3.000 seconds]
odo create/delete/list/set namespace/project tests delete namespace when force-deleting a valid namespace [BeforeEach] should successfully delete the namespace asynchronously
  [BeforeEach] C:/Users/Administrator.ANSIBLE-TEST-VS/4469/tests/integration/cmd_namespace_test.go:74
  [It] C:/Users/Administrator.ANSIBLE-TEST-VS/4469/tests/integration/cmd_namespace_test.go:96
```

PR #7041, opened 2023-08-25
Fix API Server port detection issue by making sure the same local address is used for listening and checking if the port is available

### LINKED ISSUE #6721: Port in use detection on Windows
/kind bug

**Operating System:**

Windows 10

**Output of `odo version`:**

Running 2 sessions of odo dev, one on podman and one on cluster:

```
odo dev --platform podman
[...]
- Forwarding from 127.0.0.1:20001 -> 8080
```

```
odo dev
[...]
Unable to listen on port 20001: [...] bind: Only one usage of each socket address is normally permitted [...]
```

## Actual behavior

The second session tries to reuse the port 20001, wh it is already in use.

## Expected behavior

The second session should detect the port 20001 is already in use, and use the next one (20002)

rm3l:
**What type of PR is this:**
/kind bug
/area dev

This ensures the same local address is used for listening and checking if a given port is free.
Otherwise, `net.listen("tcp", ":$port")` would listen on `0.0.0.0:$port` (see output below),
and, on some operating systems like Windows 11, `127.0.0.1:$port` is surprisingly considered free.

```
PS C:\Users\asoro> netstat -aon | grep 2000
  TCP    0.0.0.0:20000          0.0.0.0:0              LISTENING       11044
  TCP    127.0.0.1:20001        0.0.0.0:0              LISTENING       11044
  TCP    [::]:20000             [::]:0                 LISTENING       11044
```

This, as a consequence, made it impossible to run multiple Dev Sessions on Windows.

Using the same local address for listening and checking if the port is free would be safer.
If we decide to support passing a custom address to the API Server, we will use that address instead.

Fixes #6721 

**PR acceptance criteria:**

- [ ] Unit test 

- [ ] Integration test 

- [ ] Documentation 

It should be possible to start several Dev Sessions on Windows. See https://github.com/redhat-developer/odo/issues/6721#issuecomment-1693354838

### DIFF
--- a/pkg/apiserver-impl/starterserver.go
+++ b/pkg/apiserver-impl/starterserver.go
@@ -96,15 +96,16 @@ func StartServer(
 		router.PathPrefix("/").Handler(staticServer)
 	}
 
+	addr := "127.0.0.1"
 	if port == 0 && !randomPort {
-		port, err = util.NextFreePort(20000, 30001, nil, "")
+		port, err = util.NextFreePort(20000, 30001, nil, addr)
 		if err != nil {
 			klog.V(0).Infof("Unable to start the API server; encountered error: %v", err)
 			cancelFunc()
 		}
 	}
 
-	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", addr, port))
 	if err != nil {
 		return ApiServer{}, fmt.Errorf("unable to start API Server listener on port %d: %w", port, err)
 	}


feloy (approved):
Thanks for this fix

rm3l:
```
  [FAIL] odo create/delete/list/set namespace/project tests create namespace [It] should successfully create the namespace
  /go/odo_1/tests/helper/helper_cmd_wrapper.go:101
```

Test failing on OpenShift for similar reasons as in #7044 
Let's run them again to see if the cleanup operation fixes the issue..

rm3l:
```
Summarizing 1 Failure:
  [FAIL] odo create/delete/list/set namespace/project tests when namespace is created with -w [It] should list the new namespace when listing namespace
  /go/odo_1/tests/integration/cmd_namespace_test.go:38

Ran 404 of 954 Specs in 1333.451 seconds
FAIL! -- 403 Passed | 1 Failed | 0 Pending | 550 Skipped

```

Failure to be addressed in #7044 

